PR #3192, opened 2019-07-11
Fix link and add introduction to Kubernetes Client doc

geoand:


docs/src/main/asciidoc/kubernetes-client.adoc
```diff
@@ -8,9 +8,14 @@ https://github.com/quarkusio/quarkus/tree/master/docs/src/main/asciidoc
 include::./attributes.adoc[]
 
 
-Quarkus includes the `kubernetes-client` extension which enables the use of the [Fabric8 Kubernetes Client](https://github.com/fabric8io/kubernetes-client)
+Quarkus includes the `kubernetes-client` extension which enables the use of the link:https://github.com/fabric8io/kubernetes-client[Fabric8 Kubernetes Client]
```
gsmet: No need to fix it but for the future, you only need link if it's not obvious it's a URL (so things starting with `https://` or `http://` don't need it).

docs/src/main/asciidoc/kubernetes-client.adoc
```diff
@@ -8,9 +8,14 @@ https://github.com/quarkusio/quarkus/tree/master/docs/src/main/asciidoc
...
 in native mode while also making it easier to work with.
 
+Having a Kubernetes Client extension in Quarkus is very useful in order to unlock the power of Kubernetes Operators. Kubernetes Operators are quickly
+emerging as a new class of Cloud Native applications. These applications essentially watch the Kubernetes API and react to changes on various resources
+and can be used to manage the lifecycle of all kinds of complex systems like databases, messaging systems and much much more.
+Being able to write such operators in Java with a the very low footprint that native images provide, is a great match.
```
gsmet:
```suggestion
Being able to write such operators in Java with a the very low footprint that native images provide is a great match.
```
the comma doesn't make sense

Could you have one sentence per line. It would be far easier to read and better for future diffs. Asciidoc will append them anyway.

docs/src/main/asciidoc/kubernetes-client.adoc
geoand: Good to know!

gsmet (approved):

PR #3193, opened 2019-07-11
introduction of quarkus-deployment-bom

### LINKED ISSUE #3190: introduce quarkus-deployment-bom
Currently we have `quarkus-bom` which includes all the Quarkus extension runtime dependency versions. Which is then imported by `quarkus-build-parent` and user applications.

We are not offering an equivalent BOM for externally developed extension projects. A workaround would be to import `quarkus-build-parent` as a BOM. That's not its primary purpose though.

I suggest we introduce `quarkus-deployment-bom` (following the naming convention we established for extensions, i.e. extension-name -> extension-name-deployment) that would include all the Quarkus (Core) extension deployment dependency versions, which would also import `quarkus-bom` and then be imported by `quarkus-build-parent`.

aloubyansky:
Fixes #3190 

This change applies the extension artifact naming convention to the BOM(s) by moving `quarkus-bom` from `bom` dir to `bom/runtime` and introducing `quarkus-deployment-bom` under `bom/deployment`.


rsvoboda (commented):
Should we have a doc related to BOMs in Quarkus?
Having multiple BOMs can make some people confused

aloubyansky:
Definitely, I was going to update the docs in a separate PR. This change doesn't "break" the existing docs.

build-parent/pom.xml
```diff
@@ -22,19 +22,10 @@
...
         <!-- These properties are needed in order for them to be resolvable by the documentation -->
         <!-- The Graal version we suggest using in documentation - as that's
            what we work with by self downloading it: -->
         <graal-sdk.version-for-documentation>19.0.2</graal-sdk.version-for-documentation>
-        <rest-assured.version>3.3.0</rest-assured.version>
-        <axle-client.version>0.0.7</axle-client.version>
-        <vertx.version>3.7.1</vertx.version>
-
-        <!-- Enable APT by default for Eclipse -->
-        <m2e.apt.activation>jdt_apt</m2e.apt.activation>
```
gsmet: Mmmmh, where did you move these properties? I don't see them anymore (and they are needed)?

gsmet (changes requested):
I added a question. I might have missed something but it seems wrong.

build-parent/pom.xml
aloubyansky: Ugh, my bad. I had another pom which I decided not to include here. Thanks for catching it @gsmet !

bom/runtime/pom.xml
```diff
@@ -588,11 +606,13 @@
                 <groupId>io.quarkus</groupId>
                 <artifactId>quarkus-test-h2</artifactId>
                 <version>${project.version}</version>
+                <scope>test</scope>
             </dependency>
             <dependency>
                 <groupId>io.quarkus</groupId>
                 <artifactId>quarkus-test-kubernetes-client</artifactId>
                 <version>${project.version}</version>
+                <scope>test</scope>
```
gsmet: We decided against adding the `scope` in the BOM and we explicitly removed them. It's considered bad practice.

aloubyansky: Ok, removed scope test.

gsmet (approved):
Looks good to me now.

PR #3205, opened 2019-07-12
feat(#3204): Autoconfigure H2 Test Resource

### LINKED ISSUE #3204: Auto-configure H2 test resource
**Description**
Configuration of H2 could be automatically provided in the H2DatabaseTestResource as we know exactly what are the parameters we are configuring in the H2 instance.

Properties set using these method has a high preference than the ones provided when using %test profile.

Discussion here: https://groups.google.com/forum/#!topic/quarkus-dev/XL-sET6HFfs

lordofthejars:
Fixes #3204

### DIFF
--- a/extensions/agroal/deployment/src/test/resources/application-default-datasource.properties
+++ b/extensions/agroal/deployment/src/test/resources/application-default-datasource.properties
@@ -1,5 +1,5 @@
 #tag::basic[]
-quarkus.datasource.url=jdbc:h2:tcp://localhost/mem:default
+quarkus.datasource.url=jdbc:h2:mem:default
 quarkus.datasource.driver=org.h2.Driver
 quarkus.datasource.username=username-default
 quarkus.datasource.min-size=3
--- a/extensions/agroal/deployment/src/test/java/io/quarkus/agroal/test/DefaultDataSourceConfigTest.java
+++ b/extensions/agroal/deployment/src/test/java/io/quarkus/agroal/test/DefaultDataSourceConfigTest.java
@@ -45,7 +45,7 @@ public class DefaultDataSourceConfigTest {
         AgroalConnectionFactoryConfiguration agroalConnectionFactoryConfiguration = configuration
                 .connectionFactoryConfiguration();
 
-        assertEquals("jdbc:h2:tcp://localhost/mem:default", agroalConnectionFactoryConfiguration.jdbcUrl());
+        assertEquals("jdbc:h2:mem:default", agroalConnectionFactoryConfiguration.jdbcUrl());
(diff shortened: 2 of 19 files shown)


lordofthejars:
Ok, there is a problem when running the pattern of extending a test and annotate it with `@SubtrateVM`. The problem is that since `H2DatabaseTestResource` returned properties have more priority than the ones set in `application.properties` it makes the test fail because the native tests require a PostgreSQL, the configuration is not overridden to the PostgeSQL instance so the test fails.

So maybe we should close this PR as it is too restrictive or we stop using the `extend` pattern in IT tests.

extensions/agroal/deployment/src/test/resources/application-default-datasource.properties
```diff
@@ -1,5 +1,5 @@
 #tag::basic[]
-quarkus.datasource.url=jdbc:h2:tcp://localhost/mem:default
+quarkus.datasource.url=jdbc:h2:mem:default
```
gsmet: What's the rationale of this change? I don't see how it is an improvement?

gsmet (changes requested):
Blocking the PR for now, we need some clarification about the changes applied.

lordofthejars:
@gsmet https://groups.google.com/forum/#!topic/quarkus-dev/XL-sET6HFfs

But as I said now after running the IT tests, I don't know if it is not too restrictive.

extensions/agroal/deployment/src/test/resources/application-default-datasource.properties
gsmet: I have read the thread.

Still don't see why we change all these URLs everywhere?

lordofthejars: Ah ok sorry, you are changing because in these tests you need more than one different DataSource so since now `H2DatabaseTestResource` can only provide one H2 (because the config is hardcoded) I cannot use it in these tests and I need to create an embedded memory database. This comes because of this limitation of hardcoding properties, then you cannot use it in these scenarions

gsmet: Yeah, frankly, considering this limitation, I don't see this as an improvement.

lordofthejars: Exactly this was what I have noticed after seeing the IT failing tests, but since I have already sent the PR I wanted to just discuss. So I think that for now, we can close it.

aloubyansky:
Ok, closing this PR. Thanks for researching it @lordofthejars

PR #3207, opened 2019-07-12
feat(#3206): Adds a method to sort Quarkus Test Resources

### LINKED ISSUE #3206: Quarkus Test Resources should be sortable
**Description**
Quarkus Test Resources should be sortable so the developer can have a way to define in which order they should be started.

lordofthejars:
Fixes #3206

### DIFF
--- a/test-framework/common/src/main/java/io/quarkus/test/common/QuarkusTestResourceLifecycleManager.java
+++ b/test-framework/common/src/main/java/io/quarkus/test/common/QuarkusTestResourceLifecycleManager.java
@@ -32,4 +32,14 @@ public interface QuarkusTestResourceLifecycleManager {
      */
     default void inject(Object testInstance) {
     }
+
+    /**
+     * If multiple Test Resources are specified,
+     * this control the order of which they will be executed.
+     * 
+     * @return The order to be executed. The larger the number, the later the Resource is invoked.
+     */
+    default int order() {
+        return 0;
+    }
 }
--- a/test-framework/common/pom.xml
+++ b/test-framework/common/pom.xml
@@ -27,6 +27,11 @@
             <groupId>org.jboss</groupId>
             <artifactId>jandex</artifactId>
         </dependency>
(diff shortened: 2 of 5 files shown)


geoand:
Can you please squash the change into a single commit @lordofthejars ?

test-framework/common/src/main/java/io/quarkus/test/common/QuarkusTestResourceLifecycleManager.java
```diff
@@ -32,4 +32,14 @@
      */
     default void inject(Object testInstance) {
     }
+
+    /**
+     * If multiple Test Resources are specified,
+     * this control the order of which they will be executed.
+     * 
+     * @return The order to be executed. The bigger the number, it is executed later.
```
geoand: Would probably be better phrased like so: `The larger the number, the later the Resource is invoked`

lordofthejars:
@geoand It can be squashed when merged, there is a Squash and Merge button. Or you said for any other reason?

geoand:
@lordofthejars , when merging PRs Quarkus has elected to always create a merge commit and the Github UI doesn't allow you to both create a merge commit and squash commits to a single one (unless I have missed that).

lordofthejars:
@geoand Maybe you've got configured in this way in the Quarkus repo, but in FMP we are doing this way, merge + squash button.

geoand:
In any case I can't change things, so it's probably best to do what we have been doing for all other PRs.

Thanks

gsmet:
I confirm we can't do Squash + merge commit with the GitHub UI. That's why we ask for the authors to squash things themselves.

Squash + merge is a fast-forward merge. See https://help.github.com/en/articles/about-pull-request-merges#squash-and-merge-your-pull-request-commits for more info.

lordofthejars:
@gsmet @geoand Squashed.

geoand (approved):

PR #3110, opened 2019-07-05
Reinitialize Reactive Pg Client pool on restart

michalszynkiewicz:
Fixes #3037

PR #3208, opened 2019-07-12
mailer client not taking account configuration update in hot reload

### LINKED ISSUE #3038: Mail Client not taking account configuration update in hot reload
Everything is in the title.

machi1990:
Fixes #3038

/cc @cescoffier

### DIFF
--- a/extensions/mailer/runtime/src/main/java/io/quarkus/mailer/impl/MailConfigRecorder.java
+++ b/extensions/mailer/runtime/src/main/java/io/quarkus/mailer/impl/MailConfigRecorder.java
@@ -14,6 +14,7 @@ import io.vertx.ext.mail.StartTLSOptions;
 public class MailConfigRecorder {
 
     private static volatile MailClient client;
+    private static volatile MailConfig config;
 
     public RuntimeValue<MailClient> configureTheClient(RuntimeValue<Vertx> vertx, BeanContainer container,
             MailConfig config, LaunchMode launchMode, ShutdownContext shutdown) {
@@ -41,10 +42,13 @@ public class MailConfigRecorder {
     }
 
     void initialize(Vertx vertx, MailConfig config) {
-        if (client != null) {
+        if (client != null && config.equals(MailConfigRecorder.config)) {
             // Already configured
             return;
         }
+
+        this.close();
+        MailConfigRecorder.config = config;
         io.vertx.ext.mail.MailConfig cfg = toVertxMailConfig(config);
         client = MailClient.createNonShared(vertx, cfg);
     }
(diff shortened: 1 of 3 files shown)


cescoffier (approved):

cescoffier:
Waiting for CI.

extensions/mailer/runtime/src/main/java/io/quarkus/mailer/impl/MailConfigRecorder.java
```diff
@@ -41,10 +42,13 @@
     }
 
     void initialize(Vertx vertx, MailConfig config) {
-        if (client != null) {
+        if (client != null && config.equals(MailConfigRecorder.config)) {
```
gsmet: @cescoffier @machi1990 I'm not totally convinced by this change.

Shouldn't we properly close the client on shutdown and start a new one each time?

gsmet (changes requested):
Marking it as "Request changes" to get an answer to my question before it being merged :).

extensions/mailer/runtime/src/main/java/io/quarkus/mailer/impl/MailConfigRecorder.java
machi1990: > Shouldn't we properly close the client on shutdown and start a new one each time?

This was my other option but creating a client on each reload even though unnecessary was not right to me.

gsmet: Well, we usually don't do fancy things like that in other extensions and TBH I'm not sure I like it.

The contract we have for dev mode was: let's restart your Quarkus fully. And if we begin having all sorts of edge cases like that, I'm not sure it won't be a Pandora box of future bugs.

gsmet: But I'll let @cescoffier decide, I just wanted to raise my concerns.

machi1990: > The contract we have for dev mode was: let's restart your Quarkus fully. 

Sounds reasonable enough. @cescoffier WDYT?

aloubyansky:
@cescoffier i removed waiting-for-ci set by you earlier, since your opinion is needed to clear some concerns.

gsmet:
@cescoffier ping? I think we can converge quickly on this one, could you take a look at my comments?

machi1990:
I have added a second commit as par Guillaume suggestion on completely re-initialising the client during restart. The same approach was applied here: https://github.com/quarkusio/quarkus/pull/3110.

Before merging we can squash the two commits into one atomic commit.

gsmet (approved):

gsmet:
I'm merging this. @cescoffier can still revert the second commit if he disagrees with my approach.

gsmet:
Thanks @machi1990 !

PR #3211, opened 2019-07-12
Update to MP Metrics 2.0 and a compatible version of SmallRye Fault Tolerance

### LINKED ISSUE #3195: Update SmallRye Metrics to 2.0
**Description**
Update to latest SmallRye Metrics implementation

**Implementation ideas**
Update the version

machi1990:
Fixes #3195. 

Supersides #2695

@jmartisk @kenfinnigan @gsmet 

Thanks to @jmartisk for the initial work.

### DIFF
--- a/extensions/smallrye-metrics/deployment/src/main/java/io/quarkus/smallrye/metrics/deployment/SmallRyeMetricsProcessor.java
+++ b/extensions/smallrye-metrics/deployment/src/main/java/io/quarkus/smallrye/metrics/deployment/SmallRyeMetricsProcessor.java
@@ -3,9 +3,11 @@ package io.quarkus.smallrye.metrics.deployment;
 import static io.quarkus.deployment.annotations.ExecutionTime.STATIC_INIT;
 
 import java.util.Arrays;
+import java.util.Collection;
 import java.util.HashSet;
 import java.util.Set;
 
+import org.eclipse.microprofile.metrics.annotation.ConcurrentGauge;
 import org.eclipse.microprofile.metrics.annotation.Counted;
 import org.eclipse.microprofile.metrics.annotation.Gauge;
 import org.eclipse.microprofile.metrics.annotation.Metered;
@@ -13,7 +15,9 @@ import org.eclipse.microprofile.metrics.annotation.Metric;
 import org.eclipse.microprofile.metrics.annotation.Timed;
 import org.jboss.jandex.AnnotationInstance;
 import org.jboss.jandex.AnnotationValue;
+import org.jboss.jandex.ClassInfo;
 import org.jboss.jandex.DotName;
+import org.jboss.jandex.MethodInfo;
 
 import io.quarkus.arc.deployment.AdditionalBeanBuildItem;
 import io.quarkus.arc.deployment.AnnotationsTransformerBuildItem;
@@ -29,12 +33,16 @@ import io.quarkus.deployment.builditem.substrate.ReflectiveClassBuildItem;
 import io.quarkus.deployment.logging.LogCleanupFilterBuildItem;
 import io.quarkus.runtime.annotations.ConfigItem;
 import io.quarkus.runtime.annotations.ConfigRoot;
+import io.quarkus.smallrye.metrics.deployment.jandex.JandexBeanInfoAdapter;
+import io.quarkus.smallrye.metrics.deployment.jandex.JandexMemberInfoAdapter;
 import io.quarkus.smallrye.metrics.runtime.SmallRyeMetricsRecorder;
 import io.quarkus.smallrye.metrics.runtime.SmallRyeMetricsServlet;
 import io.quarkus.undertow.deployment.ServletBuildItem;
 import io.smallrye.metrics.MetricProducer;
 import io.smallrye.metrics.MetricRegistries;
 import io.smallrye.metrics.MetricsRequestHandler;
+import io.smallrye.metrics.elementdesc.BeanInfo;
+import io.smallrye.metrics.interceptors.ConcurrentGaugeInterceptor;
 import io.smallrye.metrics.interceptors.CountedInterceptor;
 import io.smallrye.metrics.interceptors.MeteredInterceptor;
 import io.smallrye.metrics.interceptors.MetricNameFactory;
@@ -44,11 +52,19 @@ import io.smallrye.metrics.interceptors.TimedInterceptor;
 
 public class SmallRyeMetricsProcessor {
 
-    private static final Set<DotName> metricsAnnotations = new HashSet<>(Arrays.asList(
-            DotName.createSimple(Gauge.class.getName()),
-            DotName.createSimple(Counted.class.getName()),
-            DotName.createSimple(Timed.class.getName()),
-            DotName.createSimple(Metered.class.getName())));
+    public static final DotName GAUGE = DotName.createSimple(Gauge.class.getName());
+    public static final DotName COUNTED = DotName.createSimple(Counted.class.getName());
+    public static final DotName TIMED = DotName.createSimple(Timed.class.getName());
+    public static final DotName METERED = DotName.createSimple(Metered.class.getName());
+    public static final DotName CONCURRENT_GAUGE = DotName.createSimple(ConcurrentGauge.class.getName());
+    public static final DotName METRICS_BINDING = DotName.createSimple(MetricsBinding.class.getName());
+
+    public static final Set<DotName> metricsAnnotations = new HashSet<>(Arrays.asList(
+            GAUGE,
+            COUNTED,
+            TIMED,
+            METERED,
+            CONCURRENT_GAUGE));
 
     @ConfigRoot(name = "smallrye-metrics")
     static final class SmallRyeMetricsConfig {
@@ -77,6 +93,7 @@ public class SmallRyeMetricsProcessor {
                 MetricRegistries.class,
                 MetricsInterceptor.class,
                 MeteredInterceptor.class,
+                ConcurrentGaugeInterceptor.class,
                 CountedInterceptor.class,
                 TimedInterceptor.class,
                 MetricsRequestHandler.class,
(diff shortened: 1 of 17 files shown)


gsmet:
So I know I usually ask for squashing but, in this case, we should really have 2 commits: the one from Jan and yours.

Considering each of you contributed significantly to this PR, it's probably better. And this way, I will be able to see your changes.

If it's too much work, let's leave it at that but if not, it would be better.

extensions/smallrye-metrics/deployment/src/main/java/io/quarkus/smallrye/metrics/deployment/SmallRyeMetricsProcessor.java
```diff
@@ -141,4 +142,44 @@ public void logCleanup(BuildProducer<LogCleanupFilterBuildItem> logCleanupFilter
...
+                switch (metricAnnotationTarget.kind()) {
+                    case CLASS: {
+                        ClassInfo clazz = metricAnnotationTarget.asClass();
+                        BeanInfo beanInfo = beanInfoAdapter.convert(clazz);
+                        for (MethodInfo method : clazz.methods()) {
+                            metrics.registerMetrics(beanInfo, memberInfoAdapter.convert(method));
+                        }
+                        break;
+                    }
+                    case METHOD: {
```
kenfinnigan: Does this result in redefining metrics if there multiple different Metric annotations on the same method or class?

machi1990: I am not sure I follow.

By `..result in redefining..` you mean `..result in the metric being registered more than once`?

extensions/smallrye-metrics/deployment/src/main/java/io/quarkus/smallrye/metrics/deployment/SmallRyeMetricsProcessor.java
```diff
@@ -165,28 +150,39 @@ public void logCleanup(BuildProducer<LogCleanupFilterBuildItem> logCleanupFilter
...
+                    case CLASS: {
+                        ClassInfo clazz = metricAnnotationTarget.asClass();
+                        collectedMetricsClasses.put(clazz.name(), clazz);
+                        break;
+                    }
+                }
             }
+        }
+
+        for (ClassInfo clazz : collectedMetricsClasses.values()) {
```
gsmet: There's something weird here: if the type is not registered but only some methods are registered, then we should not register metrics for all the methods of the classes.

Or did I miss something?

gsmet (changes requested):
Added a couple of comments inline.

extensions/smallrye-metrics/deployment/src/main/java/io/quarkus/smallrye/metrics/deployment/SmallRyeMetricsAnnotationsUtils.java
```diff
@@ -0,0 +1,37 @@
...
+import org.eclipse.microprofile.metrics.annotation.Gauge;
+import org.eclipse.microprofile.metrics.annotation.Metered;
+import org.eclipse.microprofile.metrics.annotation.Metric;
+import org.eclipse.microprofile.metrics.annotation.Timed;
+import org.jboss.jandex.AnnotationInstance;
+import org.jboss.jandex.DotName;
+
+import io.smallrye.metrics.interceptors.MetricsBinding;
+
+public class SmallRyeMetricsAnnotationsUtils {
```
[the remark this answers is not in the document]
machi1990: Done

extensions/smallrye-metrics/deployment/src/main/java/io/quarkus/smallrye/metrics/deployment/SmallRyeMetricsProcessor.java
machi1990: > ..if the type is not registered but only some methods are registered, then we should not register metrics for all the methods of the classes.

I think this is done in `MetricResolver.class` 
```
 private <T extends Annotation> MetricResolver.Of<T> beanResolverOf(MemberInfo element, Class<T> metric, BeanInfo bean) {
        if (bean.isAnnotationPresent(metric)) {
            AnnotationInfo annotation = bean.getAnnotation(metric);
            String name = this.metricName(bean, element, metric, annotation.name(), annotation.absolute());
            Tag[] tags = this.metricTags(annotation);
            return new MetricResolver.DoesHaveMetric(annotation, name, tags);
        } else {
            return (MetricResolver.Of)(bean.getSuperclass() != null ? this.beanResolverOf(element, metric, bean.getSuperclass()) : new MetricResolver.DoesNotHaveMetric());
        }
    }
```

gsmet: I'm not sure I have been clear here.

Right now, we do:
- if method annotated, add the class to a set (or whatever structure)
- if class annotated, add the class to a set
- for each class, register metrics for all the methods

Wouldn't it be better to do:
- if method annotated, add the method to a set
- if class annotated, add the class to a set
- for each method which is not part of a registered class, register metrics for the method
- for each class, register metrics for all the methods of the class

The code you point out makes me think we should also consider all the subclasses of the annotated classes (only when the annotation is defined on the class AFAICS).

machi1990: Thanks for the clarification. 

My understanding was that the above logic was handled in the resolver code I pointed above and in `Metadata.registerMetrics()` which checks if any metric is present before registering a particular metric. 

But maybe it is better to explicit this logic in the processor as you suggested. 
 
> The code you point out makes me think we should also consider all the subclasses of the annotated classes (only when the annotation is defined on the class AFAICS).

+1 good catch.

machi1990: Suggestion taken into account. WDYT @gsmet ?

extensions/smallrye-metrics/deployment/src/main/java/io/quarkus/smallrye/metrics/deployment/SmallRyeMetricsProcessor.java
```diff
@@ -165,26 +152,56 @@ public void logCleanup(BuildProducer<LogCleanupFilterBuildItem> logCleanupFilter
...
+        Map<DotName, ClassInfo> collectedMetricsClasses = new HashMap<>();
+
+        for (DotName metricAnnotation : METRICS_ANNOTATIONS) {
+            Collection<AnnotationInstance> metricAnnotationInstances = index.getAnnotations(metricAnnotation);
+            for (AnnotationInstance metricAnnotationInstance : metricAnnotationInstances) {
+                AnnotationTarget metricAnnotationTarget = metricAnnotationInstance.target();
+                switch (metricAnnotationTarget.kind()) {
+                    case METHOD: {
+                        MethodInfo method = metricAnnotationTarget.asMethod();
+                        if (!collectedMetricsMethods.contains(method)) {
```
gsmet: Maybe a `HashSet` would be better than a list?

extensions/smallrye-metrics/deployment/src/main/java/io/quarkus/smallrye/metrics/deployment/SmallRyeMetricsProcessor.java
```diff
@@ -165,26 +152,56 @@ public void logCleanup(BuildProducer<LogCleanupFilterBuildItem> logCleanupFilter
...
         }
     }
 
-    private boolean containsMetricsAnnotations(Collection<AnnotationInstance> annotations) {
-        return annotations.stream().anyMatch(annotation -> metricsAnnotations.contains(annotation.name()));
+    private void collectMetricsClassAndSubClasses(IndexView index, Map<DotName, ClassInfo> collectedMetricsClasses,
+            ClassInfo clazz) {
+        collectedMetricsClasses.put(clazz.name(), clazz);
+        for (ClassInfo subClass : index.getAllKnownSubclasses(clazz.name())) {
+            collectMetricsClassAndSubClasses(index, collectedMetricsClasses, subClass);
```
gsmet: I think you should just do:
```
collectedMetricsClasses.put(subClass.name(), subClass);
```
You don't need to go recursive as `index.getAllKnownSubclasses(clazz.name())` will return all the subclasses, not just the direct ones.

gsmet (changes requested):
@machi1990 I think we're nearly there. I added two comments, one minor, the other one not so much.

extensions/smallrye-metrics/deployment/src/main/java/io/quarkus/smallrye/metrics/deployment/SmallRyeMetricsProcessor.java
machi1990: MethodInfo has no `hashCode` method

extensions/smallrye-metrics/deployment/src/main/java/io/quarkus/smallrye/metrics/deployment/SmallRyeMetricsProcessor.java
machi1990: Good to know. Thanks.

extensions/smallrye-metrics/deployment/src/main/java/io/quarkus/smallrye/metrics/deployment/SmallRyeMetricsProcessor.java
machi1990: Sure, thanks. I thought `MethodInfo` has no `hashCode` method just like `ClassInfo`. Will change.

gsmet (approved):
Looks good to me now.

I will let @jmartisk do a review when he's back but we can already commit that.

Thanks @machi1990  and @jmartisk .

Let's wait for CI before merging.

jmartisk:
Just got back from PTO. Looks nice! Thanks @machi1990 for finishing this.

PR #3213, opened 2019-07-12
Prevent custom profile from being applied during tests execution

### LINKED ISSUE #3146: Document that running the test suite under a custom test profile is done via
Background info https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/quarkus-dev/NFpDtpRVZ7Y/3G7WmRMJDAAJ

We want to maintain an intuitive experience around using profile configuration:

1. `mvn clean install package -Dquarkus.profile=foo` should build the jar or native image with the `foo` profile but still use the `test` profile when running test

In other words, the profile passed to maven as arguments will not be propagated to the tests, only to the build process. Not doing so leads to confusing experiences and harder to write test suites. This is the current behavior AFAIK.

However, it is useful in some situations to have custom test profiles for example to ensure (part of) the test suite runs for different databases.

To achieve that, we will use the build tool native mechanism (Maven surefire configuration via maven profile, and the equivalent Gradle way). A user would write a surefire config driven by a custom Maven profile that will

1. set the user specific test specificities (maybe even filter the test suite)
2. pass as input parameter to surefire the Quarkus custom (test) profile expected.

We need documentation showing people how to do this.

### LINKED ISSUE #3149: Building with custom profile should run tests with the test profile by default
`mvn install -Dquarkus.profile=foo` should run tests under the `test` profile and build the jar / native executable under the `foo` profile.
See https://github.com/quarkusio/quarkus/issues/3146

gwenneg:
Fixes #3149

I think this PR fixes the issue expressed in #3149 but this code might need to change to fix #3146. Let's discuss that in the #3146 issue.

### DIFF
--- a/core/runtime/src/main/java/io/quarkus/runtime/configuration/ProfileManager.java
+++ b/core/runtime/src/main/java/io/quarkus/runtime/configuration/ProfileManager.java
@@ -33,22 +33,26 @@ public class ProfileManager {
     }
 
     public static String getActiveProfile() {
+        if (launchMode == LaunchMode.TEST) {
+            return launchMode.getDefaultProfile();
+        }
+
         String profile = System.getProperty(QUARKUS_PROFILE_PROP);
         if (profile != null) {
             return profile;
         }
+
         profile = System.getProperty(BACKWARD_COMPATIBLE_QUARKUS_PROFILE_PROP);
         if (profile != null) {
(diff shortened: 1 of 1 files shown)


gsmet (approved):

gsmet:
Let's merge this one and build on it. Thanks!

Issue #2125, opened 2019-04-17
Native Image SSL Problem


mertmr:
Although I have followed all the directions in the documentations(https://quarkus.io/guides/native-and-ssl-guide) I am getting an annoying warning and error at the end of my microprofile rest client to the service with SSL(only in native image, not in development):

WARNING: The sunec native library, required by the SunEC provider, could not be loaded.

Caused by: java.lang.UnsatisfiedLinkError: sun.security.ec.ECKeyPairGenerator.generateECKeyPair(I[B[B)[Ljava/lang/Object; [symbol: Java_sun_security_ec_ECKeyPairGenerator_generateECKeyPair or Java_sun_security_ec_ECKeyPairGenerator_generateECKeyPair__I_3B_3B]

I had followed these steps to produce the native image:

mvn package -Pnative -Dnative-image.docker-build=true
docker build -f src/main/docker/Dockerfile.native -t dockerrepo.company.com/appTodo
docker run -i --rm -p 8080:8080 dockerrepo.company.com/appTodo

And I put all the SSL configuration that I can find from the documentations in the maven plugin:

```
                    <plugin>
                        <groupId>io.quarkus</groupId>
                        <artifactId>quarkus-maven-plugin</artifactId>
                        <version>${quarkus.version}</version>
                        <executions>
                            <execution>
                                <goals>
                                    <goal>native-image</goal>
                                </goals>
                                <configuration>
                                    <enableHttpUrlHandler>true</enableHttpUrlHandler>
                                    <enableJni>true</enableJni>
                                    <enableAllSecurityServices>true</enableAllSecurityServices>
                                </configuration>
                            </execution>
                        </executions>
                    </plugin>
```
Any additional steps did I miss?

gsmet:
Hi Mert,

Did you see the section talking about the SunEC library? You need to include it in your Docker image and pass system properties.

Unfortunately, it's not yet automatized.

mertmr:
Thanks for the quick response, can you elaborate your suggestions a little bit more? I am not really good at Docker and I don't know how to add the library do the image and properties. I am using the default Dockerfile.native

```
FROM registry.fedoraproject.org/fedora-minimal
ENV ENV dev
WORKDIR /work/
COPY target/*-runner /work/application
RUN chmod 775 /work
EXPOSE 8080
CMD ["./application", "-Dquarkus.http.host=0.0.0.0"]
```

gsmet:
[...]

PR #3222, opened 2019-07-14
feat: added sample Dockerfile for the SunEC lib

mmornati:
# SSL connection problem in native image

## Why?
As described in the SunEC libraries and friends section, the default final docker images is not containing the GraalVM libraries to allow the usage of the SSL connection. Which means all the HTTPS client connection will fail to initialize the SSL protocol.
This part is actually quite cryptical and some information can actually be found into the Issue #2125.

## What?
An example showing how you can fix the SSL problem directly within your Docker build (nothing more necessary in your local/build machine).

### DIFF
--- a/docs/src/main/asciidoc/native-and-ssl-guide.adoc
+++ b/docs/src/main/asciidoc/native-and-ssl-guide.adoc
@@ -180,6 +180,24 @@ Now, the application should work as expected:
 
 When working with containers, the idea is to bundle both the SunEC library and the certificates in the container and to point your binary to them using the system properties mentioned above.
 
+You can for example modify your `Dockerfile.native` as following to allow copying the required lib in your final image:
+
+[source, subs=attributes+]
+----
+FROM quay.io/quarkus/ubi-quarkus-native-image:{graalvm-version} as nativebuilder
+RUN mkdir -p /tmp/ssl-libs/lib \
+  && cp /opt/graalvm/jre/lib/security/cacerts /tmp/ssl-libs \
+  && cp /opt/graalvm/jre/lib/amd64/libsunec.so /tmp/ssl-libs/lib/ 
+
+FROM registry.access.redhat.com/ubi8/ubi-minimal
+WORKDIR /work/
+COPY target/*-runner /work/application
+COPY --from=nativebuilder /tmp/ssl-libs/ /work/
+RUN chmod 775 /work
+EXPOSE 8080
+CMD ["./application", "-Dquarkus.http.host=0.0.0.0", "-Djava.library.path=/work/lib", "-Djavax.net.ssl.trustStore=/work/cacerts"]
+----
+
 [TIP]
 ====
 The root certificates file of GraalVM might not be totally up to date.


docs/src/main/asciidoc/native-and-ssl-guide.adoc
```diff
@@ -180,6 +180,23 @@ Now, the application should work as expected:
 
 When working with containers, the idea is to bundle both the SunEC library and the certificates in the container and to point your binary to them using the system properties mentioned above.
 
+You can for example modify your `Dockerfile.native` as following to allow copying the required lib in your final image:
+
+```
+FROM quay.io/quarkus/ubi-quarkus-native-image:19.0.2 as nativebuilder
```
gsmet: `19.0.2` should be replaced by `{graalvm-version}`

docs/src/main/asciidoc/native-and-ssl-guide.adoc
```diff
@@ -180,6 +180,23 @@ Now, the application should work as expected:
 
 When working with containers, the idea is to bundle both the SunEC library and the certificates in the container and to point your binary to them using the system properties mentioned above.
 
+You can for example modify your `Dockerfile.native` as following to allow copying the required lib in your final image:
+
+```
```
gsmet: This should be replaced by:
```
[source, subs=attributes+]
----
```

docs/src/main/asciidoc/native-and-ssl-guide.adoc
```diff
@@ -180,6 +180,23 @@ Now, the application should work as expected:
...
+  && cp /opt/graalvm/jre/lib/amd64/libsunec.so /tmp/ssl-libs/lib/ 
+
+FROM registry.access.redhat.com/ubi8/ubi-minimal
+WORKDIR /work/
+COPY target/*-runner /work/application
+COPY --from=nativebuilder /tmp/ssl-libs/ /work/
+RUN chmod 775 /work
+EXPOSE 8080
+CMD ["./application", "-Dquarkus.http.host=0.0.0.0", "-Djava.library.path=/work/lib", "-Djavax.net.ssl.trustStore=/work/cacerts"]
+```
```
gsmet: This should be replaced by `----`.

gsmet (changes requested):
Nice addition, I added a couple of comments.

docs/src/main/asciidoc/native-and-ssl-guide.adoc
mmornati: Ah yeah sorry... I used to make Markdown file and not adoc. Chaning it. Thanks

mmornati:
Changes applied to the PR. Thanks mate.

gsmet:
@mmornati perfect, thanks! Could you squash into one commit and I'll merge. Thanks!

gsmet (approved):

gsmet:
Merged, thanks!

gsmet:
@geoand @cescoffier could you take a look at this PR and see if it would make sense to provide an additional Dockerfile for SSL? (or always include the library + cacerts in the existing one)

geoand:
@gsmet I'll take a look later on today

mmornati:
Personally, as we were talking on Zulip with @gsmet, I think your first decision to put this kind of file outside the quarkus final image is the good one: getting on your "repo" files you are not managing it is only exposing you to an extra (useless?) job.
For this reason, maybe a better idea can be to retrieve them directly, at the build time, from the builder/graalvm image: when new files are available on the graalvm image, your image will get it to at the next build.

geoand:
I am a hesitant to add these to the Dockerfile for the reasons that @mmornati mentioned and also because of the fact that currently we don't force multi-stage docker builds

mmornati:
What about, as proposed, if we add it to a different docker file `Dockerfile.native-ssl`. You can use it only if needed but you don't have to spent time to know how you can fix your SSL problem.

geoand:
Yeah, that sounds like a good compromise to me. @cescoffier WDYT?

PR #3229, opened 2019-07-15
[mongodb-client] Avoid to fail the tests when the database cannot be stopped

loicmathieu:
There is some intermitent failures with the mongodb-client extension tests.

Some are due to the inhability to stop the launched database (I have it multiple times).

As the Database is stopped at the end of the test and the build pipeline uses containers, it's safe to ignore those errors as we cannot do anything about it anyway and the next pipeline will start with a fresh container.


aloubyansky (approved):
Thanks!

PR #3216, opened 2019-07-13
fixes #3215 CORS bug when preflight request is made to REST resource

cristhiank:
Please review if this approach is right to allow CORS preflight request pass through the Filter chain.

see #3215

PR #3230, opened 2019-07-15
Avoid using deprecated VertxOptions methods

rsvoboda:
Avoid using deprecated VertxOptions methods


aloubyansky:
Same failure as in #3216

aloubyansky (approved):
Looks good. Thanks!

PR #3236, opened 2019-07-15
Do not rely on initialization of log manager via property in native image

dmlloyd:
Required for correct operation under Java 11


dmlloyd:
Hang on, it's still racy.  V2 coming

dmlloyd:
OK better version pushed!

aloubyansky:
I'm not familiar with this, it looks like this is about not giving a choice of alternative impl, forcing our own one, is that correct?

dmlloyd:
Correct.  We've never offered a choice of alternative implementation, but the difference is that under Java 11 with native images there's some odd class initialization race condition that causes a problem with the old approach.  This approach solves the race completely by avoiding the racy code.

aloubyansky (approved):
Thanks!

PR #3242, opened 2019-07-16
Update to smallrye_jwt_1.1_2.0.1

sberyozkin:
It supports the file based key assets, avoids NPE for claims whose array values may contain null as well as does a correct mapping from a `scope` to `groups` claim when required.


gsmet (approved):
Let's wait for CI but it's a minor update so it looks good to go once CI is OK.

PR #3243, opened 2019-07-16
Arc - propose a new API for context propagation

mkouba:
Note that `RequestContext` is not thread-safe (intentionally). We use `ConcurrentMap` just to make sure the changes in the map are visible from different threads.

### DIFF
--- a/extensions/arc/runtime/src/main/java/io/quarkus/arc/runtime/context/ArcContextProvider.java
+++ b/extensions/arc/runtime/src/main/java/io/quarkus/arc/runtime/context/ArcContextProvider.java
@@ -1,6 +1,5 @@
 package io.quarkus.arc.runtime.context;
 
-import java.util.Collection;
 import java.util.Map;
 
 import org.eclipse.microprofile.context.ThreadContext;
@@ -10,7 +9,7 @@ import org.eclipse.microprofile.context.spi.ThreadContextSnapshot;
 
 import io.quarkus.arc.Arc;
 import io.quarkus.arc.ArcContainer;
-import io.quarkus.arc.ContextInstanceHandle;
+import io.quarkus.arc.InjectableContext;
 import io.quarkus.arc.ManagedContext;
 
 /**
@@ -35,8 +34,8 @@ public class ArcContextProvider implements ThreadContextProvider {
             return NOOP_SNAPSHOT;
         }
 
-        // capture all instances
-        Collection<ContextInstanceHandle<?>> instancesToPropagate = arc.requestContext().getAll();
+        // capture the state
+        InjectableContext.ContextState state = arc.requestContext().getState();
         return () -> {
             // can be called later on, we should retrieve the container again
             ArcContainer arcContainer = Arc.container();
@@ -48,17 +47,17 @@ public class ArcContextProvider implements ThreadContextProvider {
             // this is executed on another thread, context can but doesn't need to be active here
             if (isContextActiveOnThisThread(arcContainer)) {
                 // context active, store current state, feed it new one and restore state afterwards
-                Collection<ContextInstanceHandle<?>> instancesToRestore = requestContext.getAll();
+                InjectableContext.ContextState stateToRestore = requestContext.getState();
                 requestContext.deactivate();
-                requestContext.activate(instancesToPropagate);
+                requestContext.activate(state);
                 controller = () -> {
                     // clean up, reactivate context with previous values
                     requestContext.deactivate();
-                    requestContext.activate(instancesToRestore);
+                    requestContext.activate(stateToRestore);
                 };
             } else {
                 // context not active, activate and pass it new instance, deactivate afterwards
-                requestContext.activate(instancesToPropagate);
+                requestContext.activate(state);
                 controller = () -> {
                     requestContext.deactivate();
                 };
@@ -92,13 +91,13 @@ public class ArcContextProvider implements ThreadContextProvider {
             // this is executed on another thread, context can but doesn't need to be active here
             if (isContextActiveOnThisThread(arcContainer)) {
                 // context active, store current state, start blank context anew and restore state afterwards
-                Collection<ContextInstanceHandle<?>> instancesToRestore = requestContext.getAll();
+                InjectableContext.ContextState stateToRestore = requestContext.getState();
                 requestContext.deactivate();
                 requestContext.activate();
                 controller = () -> {
                     // clean up, reactivate context with previous values
                     requestContext.deactivate();
-                    requestContext.activate(instancesToRestore);
+                    requestContext.activate(stateToRestore);
                 };
             } else {
                 // context not active, activate blank one, deactivate afterwards
--- a/extensions/arc/runtime/src/main/java/io/quarkus/arc/runtime/context/ArcContextProvider.java
+++ b/extensions/arc/runtime/src/main/java/io/quarkus/arc/runtime/context/ArcContextProvider.java
@@ -18,9 +18,6 @@ import io.quarkus.arc.ManagedContext;
  */
 public class ArcContextProvider implements ThreadContextProvider {
 
-    private static ThreadContextSnapshot NOOP_SNAPSHOT = () -> () -> {
-    };
-
     @Override
     public ThreadContextSnapshot currentContext(Map<String, String> map) {
         ArcContainer arc = Arc.container();
@@ -29,13 +26,8 @@ public class ArcContextProvider implements ThreadContextProvider {
             return null;
         }
 
-        if (!isContextActiveOnThisThread(arc)) {
-            // request context not active, nothing to propagate, return no-op
-            return NOOP_SNAPSHOT;
-        }
-
-        // capture the state
-        InjectableContext.ContextState state = arc.requestContext().getState();
+        // capture the state, null indicates no active context while capturing snapshot
+        InjectableContext.ContextState state = isContextActiveOnThisThread(arc) ? arc.requestContext().getState() : null;
         return () -> {
             // can be called later on, we should retrieve the container again
             ArcContainer arcContainer = Arc.container();
@@ -49,17 +41,30 @@ public class ArcContextProvider implements ThreadContextProvider {
                 // context active, store current state, feed it new one and restore state afterwards
                 InjectableContext.ContextState stateToRestore = requestContext.getState();
                 requestContext.deactivate();
-                requestContext.activate(state);
+                if (state != null) {
+                    // only activate if previous thread had it active
+                    requestContext.activate(state);
+                }
                 controller = () -> {
                     // clean up, reactivate context with previous values
-                    requestContext.deactivate();
+                    if (state != null) {
+                        // only deactivate if previous thread had it active
+                        requestContext.deactivate();
+                    }
                     requestContext.activate(stateToRestore);
                 };
             } else {
                 // context not active, activate and pass it new instance, deactivate afterwards
-                requestContext.activate(state);
+                if (state != null) {
+                    // only activate if previous thread had it active
+                    requestContext.activate(state);
+
+                }
                 controller = () -> {
-                    requestContext.deactivate();
+                    if (state != null) {
+                        // only deactivate if previous thread had it active
+                        requestContext.deactivate();
+                    }
                 };
             }
             return controller;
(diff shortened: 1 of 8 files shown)


manovotn (approved):

independent-projects/arc/runtime/src/main/java/io/quarkus/arc/RequestContext.java
```diff
@@ -101,32 +86,34 @@ public void destroy(Contextual<?> contextual) {
...
     }
 
     @Override
-    public void deactivate() {
-        currentContext.remove();
+    public InjectableContext.ContextState getState() {
+        ConcurrentMap<Contextual<?>, ContextInstanceHandle<?>> ctx = currentContext.get();
+        if (ctx == null) {
+            // Thread local not set - context is not active!
+            throw new ContextNotActiveException();
```
FroMage: I'm not sure this is the best way to do this. For users it would be easier to have a `null` state or a singleton marker state that could be used in `activate` to call `currentContext.remove()`. Also you currently accept `null` in `activate` but it doesn't do `deactivate` which is confusing.

Otherwise callers will have to capture/restore the state by looking first at whether the request scope is active or not and that complexifies everything. If we treat inactive request contexts as a special state it makes things simpler.

independent-projects/arc/runtime/src/main/java/io/quarkus/arc/RequestContext.java
```diff
@@ -101,32 +86,34 @@ public void destroy(Contextual<?> contextual) {
...
+            currentContext.set(new ConcurrentHashMap<>());
+        } else {
+            if (initialState instanceof RequestContextState) {
+                currentContext.set(((RequestContextState) initialState).value);
+            } else {
+                throw new IllegalArgumentException("Invalid inital state: " + initialState);
+            }
         }
+        // Fire an event with qualifier @Initialized(RequestScoped.class) if there are any observers for it
+        fireIfNotEmpty(initializedNotifier);
```
FroMage: I did wonder about this: is it good that we're firing notifiers for every context restore? I've never used `@Initialized` but are they expected to be called more than once for a single request context?

extensions/arc/runtime/src/main/java/io/quarkus/arc/runtime/context/ArcContextProvider.java
```diff
@@ -35,8 +34,8 @@ public ThreadContextSnapshot currentContext(Map<String, String> map) {
             return NOOP_SNAPSHOT;
```
FroMage: @manovotn I wonder if this is wrong: supposedly we may want to capture the fact that there's no request context, so that if we ever restore that context somewhere where this is another request context we would temporarily deactivate it?

extensions/undertow/runtime/src/main/java/io/quarkus/undertow/runtime/HttpSessionContext.java
```diff
@@ -65,12 +65,19 @@ public boolean isActive() {
...
-        if (httpServletRequest != null) {
-            return new ArrayList<>(getContextualInstances(httpServletRequest).getPresentValues());
-        }
-        return Collections.emptyList();
+    public ContextState getState() {
+        return new ContextState() {
+
+            @Override
+            public Map<InjectableBean<?>, Object> getContextualInstances() {
+                HttpServletRequest httpServletRequest = servletRequest();
```
FroMage: So ATM this is not a `ManagedContext` which means we can't restore it, with a `ThreadContextProvider`? But we don't need to because it gets its values from the servlet request, which _is_ restored by CP. That's lucky because we don't support `ThreadContextProvider`s which depend on one another.

FroMage (changes requested):

independent-projects/arc/runtime/src/main/java/io/quarkus/arc/RequestContext.java
mkouba: That's a good point. I think we should only fire the event in case of `initialState == null`. The spec says that we should fire the event _"when the request context is initialized"_. But `initialized` is not well-defined. I'll change this...

independent-projects/arc/runtime/src/main/java/io/quarkus/arc/RequestContext.java
mkouba: > For users it would be easier to have a null...

I disagree. Throwing a `ContextNotActiveException` is very common in the CDI API (see for example `BeanManager.getContext()`). A user/integrator can always check `Context.isActive()` and Arc has even `ArcContainer.getActiveContext(Class<? extends Annotation>)` which returns the active context or null.

>  Also you currently accept null in activate but it doesn't do deactivate which is confusing.

I don't understand.

extensions/undertow/runtime/src/main/java/io/quarkus/undertow/runtime/HttpSessionContext.java
mkouba: Yes, it would be more complicated to make this a managed context. Honestly, this context is not supposed to be used in production ;-).

extensions/arc/runtime/src/main/java/io/quarkus/arc/runtime/context/ArcContextProvider.java
manovotn: You are right in theory but the practice is that we are doing the propagation and we are responsible for cleaning up afterwards (deactivate context). Following that principle, there cannot really be an active req. context on the "new" thread. Hence we have this `NOOP` snapshot here.

independent-projects/arc/runtime/src/main/java/io/quarkus/arc/RequestContext.java
FroMage: Well, look at the `ArcContextProvider` class: it needs to deal with deactivated contexts on the capturing side and on the restoring side. If an inactive context was a special `ContextState` it would make the API a lot simpler to use. It's cosmetic, though, we can make the `ArcContextProvider` as complex as we need to if required.

manovotn: > Throwing a ContextNotActiveException is very common in the CDI API

That's true.

> Also you currently accept null in activate but it doesn't do deactivate which is confusing.

Null just means that you make a new backing map, the deactivation is then done as usual.

extensions/undertow/runtime/src/main/java/io/quarkus/undertow/runtime/HttpSessionContext.java
FroMage: No, let's not change it because it should just work currently, the way it's implemented :)

manovotn: This isn't really for propagation, it's just a quick grab for all instance in that context.
But yes, it should incidentally be kept up to date.

extensions/arc/runtime/src/main/java/io/quarkus/arc/runtime/context/ArcContextProvider.java
FroMage: That's not true: 

```java
Runnable hasNoContext = threadContext.withContextCapture(() -> { /* no request context */});

requestContext.activate();

hasNoContext.run(); // has a request context because capture/restore is NOOP

requestContext.deactivate();
```

Now, this is academic at this point because it requires an action to be created in a thread with no request context active (perhaps the `@OnInit` beans?), receiving a notification from a request thread, which will be rare, but not impossible.

manovotn: Doing that just breaks everything, let me show you a vice-versa example of capturing a valid context a then shutting it down manually. The logic is equally skewed and it doesn't make sense to do that:

```
Runnable hasContext = threadContext.withContextCapture(() -> { /* active req. context exists here */});

requestContext.ternminate(); // destroys the whole map, even for propagated context that didnt run yet

hasContext.run(); // now this expects propagated context but there will be none now
```
E.g. we are now not capturing instances but at certain point in time (== when snapshot is created) we have to capture whatever is in the thread local Map field. If there is no active req. context, then there is no map and nothing can be put into the snapshot - simple as that.

Or maybe I still don't understand what you mean?

FroMage: Well, I'm only simulating request activation to show my point. `terminate` is done by the container, so that's another story, but the point is that if you have a contextual "thing" created outside a request, and it ends up being invoked withing a request it will inherit the request, which is not the expected outcome.

manovotn: Aah I think i finally undestand what you mean - that NOOP doesn't suspend&restore which theoretically might cause problems if the contextualized action runs on the same thread. Well, you are right there. Though from user perspective if you have an action that is able to run w/o req. context, having it won't making a difference. But we can definitely fix that. Lemme add a commit.

manovotn: @FroMage added a commit that should resolve that unless you meant something else ;)

extensions/arc/runtime/src/main/java/io/quarkus/arc/runtime/context/ArcContextProvider.java
```diff
@@ -49,17 +41,31 @@ public ThreadContextSnapshot currentContext(Map<String, String> map) {
                 // context active, store current state, feed it new one and restore state afterwards
                 InjectableContext.ContextState stateToRestore = requestContext.getState();
```
mkouba: This could throw `ContextNotActiveException`...

manovotn: This only gets called if the above condition holds true which should make sure you don't get that exception.

extensions/arc/runtime/src/main/java/io/quarkus/arc/runtime/context/ArcContextProvider.java
```diff
@@ -49,17 +41,31 @@ public ThreadContextSnapshot currentContext(Map<String, String> map) {
...
                 requestContext.deactivate();
-                requestContext.activate(state);
+                if (state != null) {
```
FroMage: OK, that's funny: this is _precisely_ what I said would be easier for API users: if inactive contexts return a `null` state and if passing `null` to `activate` would deactivate the context ;) See, I'm not _that_ crazy after all.

extensions/arc/runtime/src/main/java/io/quarkus/arc/runtime/context/ArcContextProvider.java
```diff
@@ -49,17 +41,31 @@ public ThreadContextSnapshot currentContext(Map<String, String> map) {
...
+                    // only activate if previous thread had it active
+                    requestContext.activate(state);
+                }
                 controller = () -> {
                     // clean up, reactivate context with previous values
+                    if (state != null) {
+                        // only deactivate if previous thread had it active
+                        requestContext.deactivate();
+                    }
                     requestContext.deactivate();
```
FroMage: Doesn't that deactivate it twice?

FroMage (changes requested):

extensions/arc/runtime/src/main/java/io/quarkus/arc/runtime/context/ArcContextProvider.java
FroMage: > Though from user perspective if you have an action that is able to run w/o req. context, having it won't making a difference

Well, maybe it would behave differently accidentally if it had a request context?

FroMage: But yeah, thanks.

extensions/arc/runtime/src/main/java/io/quarkus/arc/runtime/context/ArcContextProvider.java
manovotn: That's an oopsie, the second one shouldn't be there, thanks!

extensions/arc/runtime/src/main/java/io/quarkus/arc/runtime/context/ArcContextProvider.java
manovotn: I don't think passing `null` into `activate()` to deactivate is intuitive. Especially not when we need to differentiate between deactivate and destroy.

FroMage: Nevermind.

FroMage (approved):

PR #3246, opened 2019-07-16
Java 11: Disable Graphics for JAXB

dmlloyd:
This allows native images to be built on Java 11 including JAXB.

### DIFF
--- a/core/runtime/src/main/java/io/quarkus/runtime/graal/Java2DSubstitutions.java
+++ b/core/runtime/src/main/java/io/quarkus/runtime/graal/Java2DSubstitutions.java
@@ -1,0 +1,20 @@
+package io.quarkus.runtime.graal;
+
+import java.awt.Graphics;
+import java.awt.GraphicsEnvironment;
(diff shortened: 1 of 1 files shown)


gsmet (changes requested):
I'm not a big fan of having global JDK substitutions in particular extensions as, as soon as you add the extension, you change the behavior of a global component.

I mean, if it doesn't work, it should be disabled globally in the same way as ImageIO for instance. Or is there a specific reason why it's disabled only for JAXB?

I also think it would be better to throw something like `throw new UnsupportedOperationException("Not implemented yet for GraalVM native images");` so that the error has a bit of context.

dmlloyd:
The problem came up under JAXB.  But I don't have a problem moving it either.

gsmet:
I would prefer having it together with the `ImageIO` substitution and have all the JDK-level substitution global.

gsmet (approved):

gsmet:
Merged, thanks!

PR #3247, opened 2019-07-16
Allow to use the Sort for something else than ORDER BY

loicmathieu:
With the future implementation of Panache with something else than Hibernate, we need to decouple Panache Sort from SQL so providing a way to use it outside of Panache common. So I add a `getColumns()` method and opens the Column class to be use outside of the Sort class.

I deprecate the `Sort.toOrderBy()` method in case someone use it outside of hibernate-orm-panache but as I refactor hibernate-orm-panache we should be able to delete it safely.

Ping @FroMage


FroMage (approved):
LGTM

PR #3248, opened 2019-07-16
MongoDB client enhancements

loicmathieu:
Three MongoDB Client enhancements:
- add the deployment dependency to the deployment bom
- add a 'database' property, this is needed for the future Panache MongoDB extension, I also upgrade de credential management code to use it as default auth source in case no authSource is provided (as it is the case when a database is set in the connection URL)
- add the PojoCodec in automatic mode with support for Bson annotations: this is needed for the current implementation of Panache Mongo that will use this features to automatically map an entity to a Bson Document. If didn't put it behind a flag as it cannot be disabled, I can if needed

Ping @cescoffier @FroMage

### DIFF
--- a/extensions/mongodb-client/runtime/src/main/java/io/quarkus/mongodb/runtime/MongoClientConfig.java
+++ b/extensions/mongodb-client/runtime/src/main/java/io/quarkus/mongodb/runtime/MongoClientConfig.java
@@ -69,6 +69,12 @@ public class MongoClientConfig {
     public List<String> hosts;
 
     /**
+     * Configure the database name.
+     */
+    @ConfigItem
+    public Optional<String> database;
+
+    /**
      * Configures the application name.
      */
     @ConfigItem
--- a/extensions/mongodb-client/runtime/src/main/java/io/quarkus/mongodb/runtime/MongoClientRecorder.java
+++ b/extensions/mongodb-client/runtime/src/main/java/io/quarkus/mongodb/runtime/MongoClientRecorder.java
@@ -16,6 +16,8 @@ import java.util.stream.Collectors;
 import org.bson.codecs.configuration.CodecProvider;
 import org.bson.codecs.configuration.CodecRegistries;
 import org.bson.codecs.configuration.CodecRegistry;
+import org.bson.codecs.pojo.Conventions;
+import org.bson.codecs.pojo.PojoCodecProvider;
 import org.jboss.logging.Logger;
 
 import com.mongodb.AuthenticationMechanism;
@@ -86,11 +88,19 @@ public class MongoClientRecorder {
             settings.applyConnectionString(connectionString);
         }
 
-        CodecRegistry registry = defaultCodecRegistry;
+        List<CodecProvider> providers = new ArrayList<>();
         if (!codecProviders.isEmpty()) {
-            registry = CodecRegistries.fromRegistries(defaultCodecRegistry,
-                    CodecRegistries.fromProviders(getCodecProviders(codecProviders)));
+            providers.addAll(getCodecProviders(codecProviders));
         }
+        // add pojo codec provider with automatic capabilities
+        // it always needs to be the last codec provided
+        CodecProvider pojoCodecProvider = PojoCodecProvider.builder()
+                .automatic(true)
+                .conventions(Conventions.DEFAULT_CONVENTIONS)
+                .build();
+        providers.add(pojoCodecProvider);
+        CodecRegistry registry = CodecRegistries.fromRegistries(defaultCodecRegistry,
+                CodecRegistries.fromProviders(providers));
         settings.codecRegistry(registry);
 
         config.applicationName.ifPresent(settings::applicationName);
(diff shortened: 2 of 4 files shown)


extensions/mongodb-client/runtime/src/main/java/io/quarkus/mongodb/runtime/MongoClientRecorder.java
```diff
@@ -189,18 +200,19 @@ private AuthenticationMechanism getAuthenticationMechanism(String authMechanism)
...
+        String username = config.credentials.username.orElse(null);
         if (username == null) {
             return null;
         }
 
-        char[] password = config.password.map(String::toCharArray).orElse(null);
-        //admin is the default auth source in mongo and null is not allowed
-        String authSource = config.authSource.orElse("admin");
+        char[] password = config.credentials.password.map(String::toCharArray).orElse(null);
+        // get the auhtsource, or the database from the config or admin as it is the default auth source in mongo
```
machi1990: get ‘authsource’ or the database config.
Defaults to ‘admin’. 

Can we make ‘admin’ default value for ‘database’ config item?

loicmathieu: I prefer not, it's a bit dangerous as the database config will be used in something else (a future Panache Mongo implementation) and it will implies that if a user don't configure explicitly the database config it will default to admin and the data will be store in the admin database !

machi1990: Also there is a typo - should read ‘authsource’. 

I get it now. Thanks for clarifying.

loicmathieu: @machi1990 I fixed the typo, thanks

gsmet:
@loicmathieu can you please squash the typo fix in the original commit introducing it? Thanks!

loicmathieu:
@gsmet commit squashed

FroMage:
I don't know the first thing about MongoDB, but this looks OK to me.

loicmathieu:
I think it needs to be validated by @cescoffier as he's the contributor of the mongodb-client extension.

extensions/mongodb-client/runtime/src/main/java/io/quarkus/mongodb/runtime/MongoClientRecorder.java
```diff
@@ -189,18 +200,19 @@ private AuthenticationMechanism getAuthenticationMechanism(String authMechanism)
...
             return null;
         }
 
-        char[] password = config.password.map(String::toCharArray).orElse(null);
-        //admin is the default auth source in mongo and null is not allowed
-        String authSource = config.authSource.orElse("admin");
+        char[] password = config.credentials.password.map(String::toCharArray).orElse(null);
+        // get the authsource, or the database from the config, or 'admin' as it is the default auth source in mongo
+        // and null is not allowed
+        String authSource = config.credentials.authSource.orElse(config.database.orElse("admin"));
```
FroMage: Well, actuall the docs about `authsource` say:

> The value defaults to the database specified in the path portion of the connection string

But it's a lie, because I don't see that being used here. Not your fault though, but still looks like a bug to me.

FroMage (approved):

extensions/mongodb-client/runtime/src/main/java/io/quarkus/mongodb/runtime/MongoClientRecorder.java
loicmathieu: IIt was in the original code, I keep it this way but maybe this should be revisited. 
In fact it's a little complex as if you set credential inforamtion in the URL, it's true, but if you set credential information in the CredentialConfig properties an algorithm is used that mimics what the behaviour with URL does ...

So this should be rephrased by
```
This value default to the one specified in the admin property or admin if not set.
``` 

The guide clearly states that everything can be added to the URL and this is the more easier way to configure the connection to MongoDB.

extensions/mongodb-client/runtime/src/main/java/io/quarkus/mongodb/runtime/MongoClientConfig.java
```diff
@@ -68,6 +68,12 @@
     @ConfigItem
     public List<String> hosts;
 
+    /**
+     * Configure the database name.
+     */
+    @ConfigItem
+    public Optional<String> database;
```
cescoffier: Going this way, do we want to produce the MonoData and reactive equivalent directly?

cescoffier (approved):

cescoffier:
I've made a small comment but more as a follow up to ease the usage.

extensions/mongodb-client/runtime/src/main/java/io/quarkus/mongodb/runtime/MongoClientConfig.java
loicmathieu: You propose to produce a MongoDatabase and equivalent reactive objects inside the bean container if the property is set in order the user to directly use it ?

Yes, this can be convenient, as it will always be the first thing a user will do : get the database, so we can do it for him.

PR #3252, opened 2019-07-16
Provide configurable Neo4j pool settings.

michael-simons:
This provides access to configure the drivers internal pool settings under the new property group `pool`.

I have added those to the guide too.

In addition, while experimenting with moving part of the config to build phase, I moved the ssl indication out of the runtime init.

### DIFF
--- a/docs/src/main/asciidoc/neo4j-guide.adoc
+++ b/docs/src/main/asciidoc/neo4j-guide.adoc
@@ -172,7 +172,7 @@ quarkus.neo4j.authentication.password = secret
 
 You'll recognize the authentication here that you passed on to the docker command above.
 
-Having done that, the driver is ready to use.
+Having done that, the driver is ready to use, there are however other configuration options, detailed below.
 
 == Using the driver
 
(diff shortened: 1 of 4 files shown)


extensions/neo4j/runtime/src/main/java/io/quarkus/neo4j/runtime/Neo4jConfiguration.java
```diff
@@ -46,4 +53,57 @@
...
+        @ConfigItem(defaultValue = "100")
+        public int maxConnectionPoolSize;
+
+        /**
+         * Pooled connections that have been idle in the pool for longer than this timeout will be tested before they are used
+         * again. The value {@literal 0} means connections will always be tested for validity and negative values mean
+         * connections
+         * will never be tested.
+         */
+        @ConfigItem(defaultValue = "PT-0.001S")
```
machi1990: With a custom duration converter we have in place the prefix `PT` is not needed.

michael-simons: I noticed it, but put this in the standard format on purpose.

It's nice to be able to do this, but using the ISO format makes it easier to copy properties from one project to another, where maybe such a converter is not in place.

docs/src/main/asciidoc/neo4j-guide.adoc
```diff
@@ -472,3 +472,61 @@ public class ReactiveFruitResource {
...
+|Pooled connections older than this threshold will be closed and removed from the pool.
+
+|quarkus.neo4j.pool.connection-acquisition-timeout
+|java.time.Duration
+|`PT1M`
+|Acquisition of new connections will be attempted for at most configured timeout.
+
+|quarkus.neo4j.pool.idle-time-before-connection-test
+|java.time.Duration
+|`PT-0.001S`
```
gsmet: Couldn't you simply use `-1`? Our formatter should support it if it doesn't already and translate that to -1 second.

That being said, I'm not sure it's a safe default value.

extensions/neo4j/runtime/src/main/java/io/quarkus/neo4j/runtime/Neo4jConfiguration.java
```diff
@@ -46,4 +53,57 @@
...
         public boolean disabled = false;
     }
+
+    @ConfigGroup
+    static class Pool {
+
+        /**
+         * Flag, if metrics are enabled.
+         */
+        public boolean metricsEnabled = false;
```
gsmet: Why no `@ConfigItem` here?

extensions/neo4j/runtime/src/main/java/io/quarkus/neo4j/runtime/Neo4jConfiguration.java
```diff
@@ -46,4 +53,57 @@
...
+
+        /**
+         * Flag, if leaked sessions logging is enabled.
```
gsmet: Same here.

extensions/neo4j/runtime/src/main/java/io/quarkus/neo4j/runtime/Neo4jConfiguration.java
gsmet: Well, the Quarkus .properties is specific and you always have this converter in place.

As mentioned above, `-1` should work and would be much more readable.

gsmet (changes requested):
I added a few comments, thanks for the follow up!

extensions/neo4j/runtime/src/main/java/io/quarkus/neo4j/runtime/Neo4jConfiguration.java
michael-simons: Habit. Not used to the fact that I _have_ to add the annotation, was expecting plain bean mapping in such cases. Maybe it should be added to https://quarkus.io/guides/extension-authors-guide#configuration like "Leaf configuration keys must use `@ConfigItem`…"

I'll fixe it.

docs/src/main/asciidoc/neo4j-guide.adoc
michael-simons: It's the same default value the driver uses internally, so that is safe. 
I cannot use `-1` because that is a second and not a millisecond and I prefer having the same defaults as the pure driver.

extensions/neo4j/runtime/src/main/java/io/quarkus/neo4j/runtime/Neo4jConfiguration.java
michael-simons: See above. 

I removed the prefix now and hope I found a good compromise for the guide, that allows sticking with the same default as the driver.

extensions/neo4j/runtime/src/main/java/io/quarkus/neo4j/runtime/Neo4jDriverRecorder.java
```diff
@@ -1,5 +1,7 @@
 package io.quarkus.neo4j.runtime;
 
+import static java.util.concurrent.TimeUnit.*;
```
gsmet: We don't do star imports, and AFAICS, you're using only one.

gsmet (changes requested):
Added one minor comment. Otherwise it looks good to me.

gsmet (approved):

gsmet:
Merged, thanks!

michael-simons:
Awesome, thank you!

PR #3254, opened 2019-07-17
MP Metrics 2.0 no longer supports monotonic

rsvoboda:
Docs leftover removal, MP Metrics 2.0 no longer supports monotonic

MP Metrics 2.0 introduced in https://github.com/quarkusio/quarkus/commit/db171b5d7772af6a1a661b8bb86db950cc1b0ea0


gsmet (approved):
Good catch!

PR #3259, opened 2019-07-17
Arc - Default scope for additional bean defining annotations is no lo…

### LINKED ISSUE #2905: Reimplement BeanDefiningAnnotation default scope
A bean defining annotation produced by `BeanDefiningAnnotationBuildItem` with a default scope value set is currently recognized as a stereotype. For example:

```java
@Path("foo")
class MyResource {
}
```
`@Path` is by default recognized as a stereotype with `@Singleton`. This works fine even if a user adds a custom scope:

```java
@RequestScoped
@Path("foo") // -> stereotype is ignored
class MyResource {
}
```
However, the app would fail if a user adds a strereotype with a different scope:

```java
@Model // -> stereotype - declares @RequestScoped
@Path("foo") // -> stereotype - declares @Singleton
class MyResource {
}
```
We could try to add the scope annotation directly if the bean class does not declare any scope annotation. But the problem is that we can't register an `AnnotationTransformer` because the custom scopes are not known during registration and so we may end up with unpredictable results. I'm not sure what the right approach would be.

manovotn:
…nger contesting with user defined stereotypes.

Fixes #2905 

I tested this using altered quickstart sample.

### DIFF
--- a/independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/Beans.java
+++ b/independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/Beans.java
@@ -324,10 +324,16 @@ final class Beans {
             return null;
         }
         final Set<ScopeInfo> stereotypeScopes = new HashSet<>();
+        final Set<ScopeInfo> additionalBDAScopes = new HashSet<>();
         for (StereotypeInfo stereotype : stereotypes) {
-            stereotypeScopes.add(stereotype.getDefaultScope());
+            if (!stereotype.isAdditionalBeanDefiningAnnotation()) {
+                stereotypeScopes.add(stereotype.getDefaultScope());
+            } else {
+                additionalBDAScopes.add(stereotype.getDefaultScope());
+            }
         }
-        return BeanDeployment.getValidScope(stereotypeScopes, target);
+        // if the stereotypeScopes set is empty, operate on additional BDA stereotypes instead
+        return BeanDeployment.getValidScope(stereotypeScopes.isEmpty() ? additionalBDAScopes : stereotypeScopes, target);
     }
 
     private static boolean initStereotypeAlternative(List<StereotypeInfo> stereotypes) {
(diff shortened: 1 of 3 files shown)


independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/Beans.java
```diff
@@ -323,9 +323,15 @@ private static ScopeInfo initStereotypeScope(List<StereotypeInfo> stereotypes, A
         if (stereotypes.isEmpty()) {
             return null;
         }
-        final Set<ScopeInfo> stereotypeScopes = new HashSet<>();
-        for (StereotypeInfo stereotype : stereotypes) {
-            stereotypeScopes.add(stereotype.getDefaultScope());
+        // first go through all standard stereotypes
+        Set<ScopeInfo> stereotypeScopes = stereotypes.stream().filter(s -> !s.isAdditionalBeanDefiningAnnotation())
```
mkouba: What's wrong with adding one `if` statement to the original for loop? I mean the stream and two more lambdas don't bring any benefit here, except that it's harder to read and few more objects are created ;-).

independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/Beans.java
```diff
@@ -323,9 +323,15 @@ private static ScopeInfo initStereotypeScope(List<StereotypeInfo> stereotypes, A
...
+                .map(stereotypeInfo -> stereotypeInfo.getDefaultScope())
+                .collect(Collectors.toSet());
+        // if the set is empty, try adding all stereotypes originating from additional bean def. ann.
+        if (stereotypeScopes.isEmpty()) {
+            stereotypeScopes.addAll(stereotypes.stream().filter(s -> s.isAdditionalBeanDefiningAnnotation())
```
mkouba: There's no need to `filter` the stream here, right?

manovotn: You either filter it or you attempt to re-add those that you already went over, which is harmless anyway.

mkouba: If it's empty then there are no non-bean defining annotations stereotypes, or?

independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/Beans.java
manovotn: @mkouba better? :)

mkouba (approved):

PR #3265, opened 2019-07-17
Document custom test profile usage with Maven and Gradle

### LINKED ISSUE #3146: Document that running the test suite under a custom test profile is done via

gwenneg:
Fixes #3146

### DIFF
--- a/docs/src/main/asciidoc/maven-tooling.adoc
+++ b/docs/src/main/asciidoc/maven-tooling.adoc
@@ -485,3 +485,51 @@ Uber-Jar's final name is configurable via a Maven's build settings `finalName` o
 
 <1> This configuration will output a file `uber-jar-name-runner.jar` inside the `target` directory. When `finalName` option is not set, project's build final name is used.
 
+[[custom-test-configuration-profile]]
+=== Custom test configuration profile
+
+By default, Quarkus tests are run using the `test` configuration profile. If you are not familiar with Quarkus configuration
+profiles, everything you need to know is explained in the
+link:application-configuration-guide.html#configuration-profiles[Configuration Profiles Documentation].
+
+It is however possible to use a custom configuration profile for your tests with the Maven Surefire and Maven Failsafe
+configurations shown below. This can be useful if you need for example to run some tests using a specific database which is not
+your default testing database.
+
+[source,xml,subs=attributes+]
+----
+<project>
+  [...]
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-surefire-plugin</artifactId>
+        <version>${surefire-plugin.version}</version>
+        <configuration>
+          <systemPropertyVariables>
+            <quarkus.test.profile>foo</quarkus.test.profile> <1>
+            <buildDirectory>${project.build.directory}</buildDirectory>
+            [...]
+          </systemPropertyVariables>
+        </configuration>
+      </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-failsafe-plugin</artifactId>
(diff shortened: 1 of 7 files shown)


gwenneg:
@gsmet: I need to confirm the `NativeImageLauncher` change isn't breaking anything, I'll try to test that ASAP. In the meantime, it would be great if you could start reviewing the PR (especially the doc part).

gwenneg:
The commit message could be changed if you don't want to draw too much attention on the system property.

geoand:
Tests are failing

gwenneg:
@geoand Sorry about that, I forgot to reset the ProfileManager state between tests. It should be better now!

docs/src/main/asciidoc/maven-tooling.adoc
```diff
@@ -485,3 +485,39 @@ Uber-Jar's final name is configurable via a Maven's build settings `finalName` o
...
+  [...]
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-surefire-plugin</artifactId>
+        <version>3.0.0-M3</version>
+        <configuration>
+          <systemPropertyVariables>
+            <quarkus.test.profile>foo</quarkus.test.profile> <1>
```
gsmet: We should have an example for failsafe too.

gsmet (commented):
Nice work (should I say "as usual"?).

Could you add a Failsafe example? Otherwise, I think it looks good to go.

docs/src/main/asciidoc/maven-tooling.adoc
gwenneg: You're right, I forgot about it. It's fixed.

docs/src/main/asciidoc/maven-tooling.adoc
```diff
@@ -485,3 +485,51 @@ Uber-Jar's final name is configurable via a Maven's build settings `finalName` o
...
+[source,xml,subs=attributes+]
+----
+<project>
+  [...]
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-surefire-plugin</artifactId>
+        <version>3.0.0-M3</version>
```
gsmet: Last minute thought: I'm torn here. I think having the versions is not a very good idea: they will become outdated quickly. So, either we consider people have a plugin management section or we insert the version we have in Quarkus or (and it has my preference) we follow the pattern we have in our generated projects and we use `${surefire-plugin.version}` (we use it also for Failsafe)?

WDYT?

gwenneg: I vote for staying consistent with the generated projects. Versions removed :)

gsmet (approved):
Looks good to me. Thanks!

PR #3270, opened 2019-07-18
Improve Hibernate Validator integration with CDI

### LINKED ISSUE #2900: Not able to perform @Inject on class which implements ConstraintValidator
Say if I have a custom validator annotation:

```
@Target( { ElementType.TYPE, ElementType.FIELD, ElementType.METHOD, ElementType.PARAMETER, ElementType.ANNOTATION_TYPE } )
@Retention( RetentionPolicy.RUNTIME )
@Constraint( validatedBy = PayloadValidator.class )
@Documented
public @interface PayloadCheck {
    String message( ) default "Not a valid blah blah";
    Class< ? >[] groups( ) default { };
    Class< ? extends Payload >[] payload( ) default { };
}
```

and custom payload class:
```
@PayloadCheck
public class CustomPayload {
    // all fields, setters and getters
}
```
and a custom implemented validator class:
```
@ApplicationScoped
public class PayloadValidator implements ConstraintValidator< PayloadCheck, Payload > {

    @Inject
    SomeQueryService someQueryServ;    // This will be "null"

    @Override
    public void initialize( PayloadCheck ca ) {
    }

    @Override
    public Class< ? extends Annotation > annotationType( ) {
        return PayloadCheck.class;
    }

    @Override
    public boolean isValid( CustomPayload payload, ConstraintValidatorContext ctx ) {
        someQueryServ.doSomething();    // someQueryServ will not be injected and will throw a NullPointerException
        return false;
    }
}
```

For any other class annotated with @ApplicationScope that has:
```
@Inject
Validator validator;
```
and later
```
CustomPayload customPayload = someWaysOfGettingThis();
Set< ConstraintViolation< CustomPayload > > violations = validator.validate( customPayload );
```

this will throw a NullPointerException because "someQueryServ" in PayloadValidator is not injected.

There are some issues and solution from other framework on this, which you may reference: 
[https://github.com/micronaut-projects/micronaut-core/issues/1143](https://github.com/micronaut-projects/micronaut-core/issues/1143)

[https://github.com/micronaut-projects/micronaut-core/blob/master/configurations/hibernate-validator/src/main/java/io/micronaut/configuration/hibernate/validator/ValidatorFactoryProvider.java#L74](https://github.com/micronaut-projects/micronaut-core/blob/master/configurations/hibernate-validator/src/main/java/io/micronaut/configuration/hibernate/validator/ValidatorFactoryProvider.java#L74)

Quarkus Version: 0.17.0
JDK Version: openjdk version "1.8.0_212"
OpenJDK Runtime Environment (Zulu 8.38.0.13-CA-macosx) (build 1.8.0_212-b04)
OpenJDK 64-Bit Server VM (Zulu 8.38.0.13-CA-macosx) (build 25.212-b04, mixed mode)

Many thanks!

### LINKED ISSUE #2974: quarkus-hibernate-validator module ignore custom constraint validators defined in SPI file
**Describe the bug**
quarkus-hibernate-validator module ignores ConstraintValidators defined in 

/META-INF/services/javax.validation.ConstraintValidator file

**Expected behavior**
to be able to register custom constraint validators  for third party annotations in /META-INF/services/javax.validation.ConstraintValidator

**Actual behavior**
Custom constraint validators defined in SPI file not registered and ignored.

**To Reproduce**
1. Create new project with quarkus-hibernate-validator module
2. Create new constraint validator for one of existing constraint (For example create MinCustomValidator for @Min constraint)

```java
public class MinCustomValidator implements ConstraintValidator<Min, Long> {

  @Override
  public boolean isValid(Long value, ConstraintValidatorContext context) {
    return false;
  }
  
}
```

3. put the full class name in to the /META-INF/services/javax.validation.ConstraintValidator SPI file 

` com.acme.MinCustomValidator`

4. Add a simple rest resource with parameter annotated with selected annotation for the custom constraint validator.

```java
@GET
@Path("/{value}")
public Response validate(@Min(1) @PathParam("value") long value) {
}
```

5. Call endpoint with long value greater than 1

You will see that the custom constraint validator not taken into consideration

gsmet:
This is a work in progress as it will require a Hibernate Validator release.

Fixes #2900
Fixes #2974

### DIFF
--- a/extensions/hibernate-validator/deployment/src/main/java/io/quarkus/hibernate/validator/deployment/HibernateValidatorProcessor.java
+++ b/extensions/hibernate-validator/deployment/src/main/java/io/quarkus/hibernate/validator/deployment/HibernateValidatorProcessor.java
@@ -9,8 +9,13 @@ import java.util.HashSet;
 import java.util.Set;
 import java.util.function.Predicate;
 
+import javax.validation.ClockProvider;
 import javax.validation.Constraint;
 import javax.validation.ConstraintValidator;
+import javax.validation.ConstraintValidatorFactory;
+import javax.validation.MessageInterpolator;
+import javax.validation.ParameterNameProvider;
+import javax.validation.TraversableResolver;
 import javax.validation.Valid;
 import javax.validation.executable.ValidateOnExecution;
 
@@ -46,6 +51,13 @@ import io.quarkus.hibernate.validator.runtime.interceptor.MethodValidationInterc
 
 class HibernateValidatorProcessor {
 
+    private static final DotName CONSTRAINT_VALIDATOR_FACTORY = DotName
+            .createSimple(ConstraintValidatorFactory.class.getName());
+    private static final DotName MESSAGE_INTERPOLATOR = DotName.createSimple(MessageInterpolator.class.getName());
+    private static final DotName TRAVERSABLE_RESOLVER = DotName.createSimple(TraversableResolver.class.getName());
+    private static final DotName PARAMETER_NAME_PROVIDER = DotName.createSimple(ParameterNameProvider.class.getName());
+    private static final DotName CLOCK_PROVIDER = DotName.createSimple(ClockProvider.class.getName());
+
     private static final DotName CONSTRAINT_VALIDATOR = DotName.createSimple(ConstraintValidator.class.getName());
 
     private static final DotName VALIDATE_ON_EXECUTION = DotName.createSimple(ValidateOnExecution.class.getName());
@@ -77,11 +89,13 @@ class HibernateValidatorProcessor {
                     "io.quarkus.hibernate.validator.runtime.jaxrs.JaxrsEndPointValidationInterceptor"));
         }
 
-        // Do not remove the ConstraintValidator beans
+        // Do not remove the Bean Validation beans
         unremovableBean.produce(new UnremovableBeanBuildItem(new Predicate<BeanInfo>() {
             @Override
             public boolean test(BeanInfo beanInfo) {
-                return beanInfo.hasType(CONSTRAINT_VALIDATOR);
+                return beanInfo.hasType(CONSTRAINT_VALIDATOR) || beanInfo.hasType(CONSTRAINT_VALIDATOR_FACTORY)
+                        || beanInfo.hasType(MESSAGE_INTERPOLATOR) || beanInfo.hasType(TRAVERSABLE_RESOLVER)
+                        || beanInfo.hasType(PARAMETER_NAME_PROVIDER) || beanInfo.hasType(CLOCK_PROVIDER);
             }
         }));
     }
--- a/extensions/hibernate-validator/deployment/src/main/java/io/quarkus/hibernate/validator/deployment/HibernateValidatorProcessor.java
+++ b/extensions/hibernate-validator/deployment/src/main/java/io/quarkus/hibernate/validator/deployment/HibernateValidatorProcessor.java
@@ -22,6 +22,8 @@ import javax.validation.valueextraction.ValueExtractor;
 
 import org.hibernate.validator.internal.metadata.core.ConstraintHelper;
 import org.hibernate.validator.messageinterpolation.AbstractMessageInterpolator;
+import org.hibernate.validator.spi.properties.GetterPropertySelectionStrategy;
+import org.hibernate.validator.spi.scripting.ScriptEvaluatorFactory;
 import org.jboss.jandex.AnnotationInstance;
 import org.jboss.jandex.AnnotationTarget;
 import org.jboss.jandex.ClassInfo;
@@ -58,6 +60,9 @@ class HibernateValidatorProcessor {
     private static final DotName TRAVERSABLE_RESOLVER = DotName.createSimple(TraversableResolver.class.getName());
     private static final DotName PARAMETER_NAME_PROVIDER = DotName.createSimple(ParameterNameProvider.class.getName());
     private static final DotName CLOCK_PROVIDER = DotName.createSimple(ClockProvider.class.getName());
+    private static final DotName SCRIPT_EVALUATOR_FACTORY = DotName.createSimple(ScriptEvaluatorFactory.class.getName());
+    private static final DotName GETTER_PROPERTY_SELECTION_STRATEGY = DotName
+            .createSimple(GetterPropertySelectionStrategy.class.getName());
 
     private static final DotName CONSTRAINT_VALIDATOR = DotName.createSimple(ConstraintValidator.class.getName());
     private static final DotName VALUE_EXTRACTOR = DotName.createSimple(ValueExtractor.class.getName());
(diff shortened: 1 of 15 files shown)


extensions/hibernate-validator/runtime/src/main/java/io/quarkus/hibernate/validator/runtime/ArcConstraintValidatorFactoryImpl.java
```diff
@@ -0,0 +1,40 @@
...
+        if (instanceHandle.isAvailable()) {
+            return instanceHandle.get();
+        } else {
+            return run(NewInstance.action(key, "ConstraintValidator"));
+        }
+    }
+
+    @Override
+    public void releaseInstance(ConstraintValidator<?, ?> instance) {
+        // we let ArC release the bean instances on shutdown as they were instantiated by ArC.
```
atzoum: This is OK for scoped validators, but if a validator is of `@Dependent` scope then ARC will produce a new validator instance for every call of `ArcConstraintValidatorFactoryImpl#getInstance` leading to a potential memory leak. Please consider this solution

```
    private final Map<Object, InstanceHandle<?>> destroyableValidators = new ConcurrentHashMap<>();
    
    @Override
    public <T extends ConstraintValidator<?, ?>> T getInstance(Class<T> key) {
        InstanceHandle<T> handle = Arc.container().instance(key);
        if (handle.isAvailable()) {
            T instance = handle.get();
            if (handle.getBean().getScope().equals(Dependent.class)) {
                destroyableValidators.put(instance, handle);
            }
            return instance;
        }
        return run( NewInstance.action( key, "ConstraintValidator" ) );
    }

    @Override
    public void releaseInstance(ConstraintValidator<?, ?> instance) {
        ofNullable(destroyableValidators.remove(instance)).ifPresent(InstanceHandle::destroy);
    }
    
    private <T> T run(PrivilegedAction<T> action) {
        return System.getSecurityManager() != null ? AccessController.doPrivileged( action ) : action.run();
    }
```

atzoum (changes requested):
By coincidence, I also started to work on #2900 because it affects our migration. Please consider the proposed change :)

extensions/hibernate-validator/runtime/src/main/java/io/quarkus/hibernate/validator/runtime/ArcConstraintValidatorFactoryImpl.java
gsmet: I must admit that I only considered application scoped and singleton beans.

I added a commit along the lines of what you suggested.

gsmet:
@aloubyansky this PR is now ready for review.

extensions/hibernate-validator/runtime/src/main/java/io/quarkus/hibernate/validator/runtime/ArcConstraintValidatorFactoryImpl.java
```diff
@@ -13,19 +16,27 @@
...
-        if (instanceHandle.isAvailable()) {
-            return instanceHandle.get();
-        } else {
-            return run(NewInstance.action(key, "ConstraintValidator"));
+    public <T extends ConstraintValidator<?, ?>> T getInstance(Class<T> key) {
+        InstanceHandle<T> handle = Arc.container().instance(key);
+        if (handle.isAvailable()) {
+            T instance = handle.get();
+            if (handle.getBean().getScope().equals(Dependent.class)) {
+                destroyableConstraintValidators.put(instance, handle);
```
aloubyansky: Is it safe to use `instance` as a key?

gsmet: Which issues do you think it could cause?

I don't see a lot of other solutions.

gsmet: @mkouba we could probably use your wisdom here.

On this particular commit: https://github.com/quarkusio/quarkus/pull/3270/commits/e743b017b88a1f3c9e931c6aa94ae4e1221d671e , added following this comment https://github.com/quarkusio/quarkus/pull/3270#discussion_r304915534 .

Does it look OK to you?

gsmet: FWIW, I was wondering if I would only allow `@Singleton` and `@ApplicationScoped` beans here and throw an error if `ConstraintValidator` beans have another scope.

Thoughts?

gsmet: @aloubyansky hmm, and you're right, even if we keep this behavior, the map should be an identity based map.

gsmet: I switched to an `IdentityHashMap` (it's safe as in the case of Quarkus, constraint validators are all created upfront).

Still interested in your feedback on the approach though.

extensions/hibernate-validator/deployment/src/main/java/io/quarkus/hibernate/validator/deployment/HibernateValidatorProcessor.java
```diff
@@ -55,7 +82,9 @@ LogCleanupFilterBuildItem logCleanup() {
     }
 
     @BuildStep
-    void registerAdditionalBeans(BuildProducer<AdditionalBeanBuildItem> additionalBeans) {
+    void registerAdditionalBeans(BuildProducer<AdditionalBeanBuildItem> additionalBeans,
+            BuildProducer<AnnotationsTransformerBuildItem> annotationsTransformers,
```
aloubyansky: BTW, this arg isn't used in this method.

aloubyansky (approved):
One minor remark. Otherwise, it looks good to me.

extensions/hibernate-validator/deployment/src/main/java/io/quarkus/hibernate/validator/deployment/HibernateValidatorProcessor.java
gsmet: Ah, yes, good catch, I removed the annotation transformers I initially added. Will clean it up.

gsmet: @aloubyansky fixed and rebased.

extensions/hibernate-validator/runtime/src/main/java/io/quarkus/hibernate/validator/runtime/ArcConstraintValidatorFactoryImpl.java
mkouba: @gsmet So if you produce a `@Dependent` bean instance via prorammatic lookup you should always destroy the bean instance correctly once it is not needed. And `InstanceHandle.destroy()` is a non-portable convenient way to do this. In other words, if you produce `@Dependent` validators you should destroy them afterwards. HTH

PR #3272, opened 2019-07-18
inheriting @Observers, fixes #3260

### LINKED ISSUE #3260: CDI observer methods should be inherited if not overriden
See https://docs.jboss.org/cdi/spec/2.0/cdi-spec.html#member_level_inheritance for more information.

michalszynkiewicz:


independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/BeanDeployment.java
```diff
@@ -605,29 +608,53 @@ static ScopeInfo getValidScope(Collection<ScopeInfo> stereotypeScopes, Annotatio
...
+                            beanClasses.add(beanClass);
+                            if (!hasBeanDefiningAnnotation) {
+                                LOGGER.debugf("Observer method found but %s has no bean defining annotation - using @Dependent",
+                                        beanClass);
+                            }
                         }
                     }
                 }
+                Type superType = aClass.superClassType();
+                aClass = superType != null && superType.kind() == Type.Kind.CLASS
```
mkouba: 1. End iteration if `superType.name()` is `java.lang.Object`
2. I don't think that `superType.kind() == Type.Kind.CLASS` is valid -> if `Foo extends Bar<String>` then you would not scan `Bar` as it has `Type.Kind.PARAMETERIZED_TYPE`
3. `index.getClassByName(superType.name())` may return null

independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/BeanDeployment.java
```diff
@@ -681,31 +708,37 @@ static ScopeInfo getValidScope(Collection<ScopeInfo> stereotypeScopes, Annotatio
...
         }
 
-        for (MethodInfo observerMethod : syncObserverMethods) {
-            BeanInfo declaringBean = beanClassToBean.get(observerMethod.declaringClass());
-            if (declaringBean != null) {
-                Injection injection = Injection.forObserver(observerMethod, declaringBean.getImplClazz(), this,
-                        injectionPointTransformer);
-                observers.add(new ObserverInfo(declaringBean, observerMethod, injection, false));
-                injectionPoints.addAll(injection.injectionPoints);
+        syncObserverMethods.forEach(
```
mkouba: Pls don't use lambdas if not necessary. A simple `for` cycle is much more readable.

independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/BeanDeployment.java
michalszynkiewicz: if `index.getClassByName(...)` returns `null`, the loop will not start the next iteration

independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/BeanDeployment.java
michalszynkiewicz: thought it's slightly more readable this way than iterating through `Map.Entry<...>` from `syncObserverMethods.entrySet` but will change it if you prefer it this way.

manovotn (commented):
The only problem I can think of are (as usual) generics and specifically bridge methods which might interfere with the overriden filtering you do.
Which brings me to another point - it would be good to add some tests which use CDI beans and inherit observers. Though I have to say I checked CDI TCK and found only tests without generics (such as [this one](https://github.com/cdi-spec/cdi-tck/blob/master/impl/src/main/java/org/jboss/cdi/tck/tests/event/observer/inheritance/ObserverInheritanceTest.java)), so perhaps we don't need to go overboard with generics here ;-)

michalszynkiewicz:
@manovotn yeah, I think I overdid the check for overriding. The current version does it in a much simpler way, IIUC, consistent with the way Weld does it.

Thank you both for the review :)

manovotn:
Oki, then I'll re-review it.
BTW `AssignabilityWithGenericsTest` in Arc seems to be failing in CI now.

michalszynkiewicz:
@manovotn should be fixed, there was an inherited observer there with an injected bean that wasn't triggered before my change (because it wasn't supported :) )

mkouba (approved):

manovotn (approved):

PR #3275, opened 2019-07-18
test url wrong in 'Configuring Your Application' guide

chaensel:
The test url for section 'Create the Configuration' is wrong. It should be ```http://localhost:8080/hello``` instead of ```http://localhost:8080/greeting```.

### DIFF
--- a/docs/src/main/asciidoc/application-configuration-guide.adoc
+++ b/docs/src/main/asciidoc/application-configuration-guide.adoc
@@ -109,7 +109,7 @@ Once set, check the application with:
 
 [source,shell]
 ----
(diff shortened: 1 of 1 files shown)


aloubyansky:
Are you sure? I don't see any `/hello` path in the text. It looks like `/greeting` is correct.

aloubyansky:
If you think it should be `/hello` please reference the relevant configuration indicating that. Until then I am closing the issue. Thanks.

chaensel:
You're right. I've reused the source code from the "Creating Your First Application" guide and have overseen that the Path changed for the GreetingResource.
Thanks

PR #3283, opened 2019-07-19
Upgrade to Kogito 0.2.0

mswiderski:


gsmet (approved):

gsmet:
@mswiderski I let you check that the quickstart still works, once this is applied? Thanks!

mswiderski:
@gsmet I tested it before (few days ago) and now but found a problem with dev mode that was not there few days ago. 
using-kogito QuickStart works fine with jvm mode and native image mode but does not work with dev mode. It simply does not see the REST endpoints that are generated... any idea what that might be?

I verified that the classes are generated and even delivered (via index) to `ResteasyServerCommonProcessor` but somehow they are not available on running application

mswiderski:
and forgot to add, I downgraded kogito locally to be 0.1.3 (which is in the released version) but that has same effect, dev mode not working as expected.

gsmet:
Could you create an issue with the details. It seems very weird.

mswiderski:
@gsmet fully agree ... it is weird and it did work smoothly till like two or three days ago (As I didn't sync with master for a few days) ... anyway

here is the issue https://github.com/quarkusio/quarkus/issues/3288

PR #2930, opened 2019-06-21
Add an ActiveMQ Artemis extension

middagj:
I have added an extension to support ActiveMQ Artemis via Core or JMS. Using AMQP via reactive messaging does not suit our needs as we want to control the acknowledgements and commits.

I have kept the CDI part small and only produce a ServerLocater (Core) and ActiveMQJMSConnectionFactory (JMS). These components should not be created again even if the connection dies for some reason. I don't know if building a framework around this (using an annotation for the listener method, create a proper producing template with serialization and transactions) is needed. It would take considerably more time, but would definitely add developer joy.

I also have chosen to support a single url property and not all the options separately. In Artemis you can specify the options there (like JDBC).

The changes for the Netty native libraries I have put in the NettyProcessor. They postpone initialization of those classes to runtime. I have disable epoll and kqueue in SubstrateVM for Artemis because of an UnsatisfiedLinkError. Others also seem to fail, e.g. [Ratpack](https://e.printstacktrace.blog/ratpack-graalvm-how-to-start/), [Vert.x](https://github.com/oracle/graal/issues/442).

PR #3286, opened 2019-07-19
Revert JniBuildItem in Netty and add in Amazon DynamoDB

middagj:
I was a little quick in #3267. Artemis uses Netty which requires JNI, but only if the epoll and kqueue are explicitly added as dependency. This is not the case with Infinispan and neo4j. They only require JNI only if SSL is used. So hereby a revert of that change and I missed DynamoDB.

I will fix Netty in #2930 since there is already code that checks if epoll or kqueue are added as dependency.

Sorry for the mess.


gsmet (approved):

gsmet:
Thanks for the follow-up!

PR #3302, opened 2019-07-22
Dev mode does not work with kogito quickstart

mswiderski:
Delegates classes to source path lookup to CompilerProvider as there might be different relation shift between class and source. That is especially valid for kogito that generates classes based on processes and rules files. Similar might be for Kotlin and that's why I just added simple implementation for Kotlin as well to avoid same situation as it was in kogito that generated class files where removed due to not found corresponding java source file.

main implementation is in this commit https://github.com/quarkusio/quarkus/commit/2d68d37f6a78883d45f7a9e15289f2ac5c86fcf3 while the other commits improve the hot reload for kogito to make use of 0.2.0 changes.


devtools/maven/src/test/resources/projects/simple-kogito/pom.xml
```diff
@@ -0,0 +1,88 @@
...
+      <version>${quarkus.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>@project.groupId@</groupId>
+      <artifactId>quarkus-jsonp</artifactId>
+      <version>${quarkus.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>@project.groupId@</groupId>
+      <artifactId>quarkus-kogito</artifactId>
```
stuartwdouglas: You will need to also add a test scoped dependency to quarkus-kogito to the pom for the maven plugin. Otherwise it is possible for maven to decide to build the plugin before the kogito module and the build will fail.

Longer term we want to add a more generic way of doing hot reload tests for extensions via some kind of dev mode test framework for extensions, and when we have this we will likely try and move these tests to there.

stuartwdouglas (changes requested):

devtools/maven/src/test/resources/projects/simple-kogito/pom.xml
mswiderski: added the dependency the maven plugin pom as requested

mswiderski:
@stuartwdouglas any further comments or is this good to go?

stuartwdouglas (approved):

mswiderski:
thanks @stuartwdouglas

Issue #3253, opened 2019-07-17
ExceptionMapper doesn't work for NotFoundException


barata0:
**Describe the bug**
ExceptionMapper works fine for Exceptions, WebApplicationExceptions and whatever exception you create FooException, BarException, etc...

But it does not work for NotFoundException. Instead the default Resource Not Found page is displayed 

**Expected behavior**
When a NotFoundException occurs, the mapper should take precedence over the default Resource Not Found

**Actual behavior**
Does not matter if there are any Mapper registered for NotFoundExecption, the html page is displayed

**To Reproduce**
Create a Mapper for every exception 

```
public class NotFoundMapper implements ExceptionMapper<Exception> {
    @Override
    public Response toResponse(Exception exception) {
        int code = 500;
        if (exception instanceof WebApplicationException) {
            code = ((WebApplicationException) exception).getResponse().getStatus();
        }
        return Response.status(code)
                .entity(Json.createObjectBuilder().add("error", exception.getMessage()).add("code", code).build())
                .build();
    }
}
```

Then, throw an NotFoundException on an endpoint

```
    @GET
    @Path("/{id}")
    public Transf findById(@PathParam("id") Long id) {
        Transf t = Transf.findById(id);
        if (t == null) {
            throw new NotFoundException();
        }
        return t;
    }

```

The result is an HTML page.
Changing the NotFoundException for any other exception (even WebApplicationException) works fine and the json with the error status is returned.

**Configuration**
```
21:40:32 INFO  [io.quarkus]] (executor-thread-1) Installed features: [agroal, cdi, hibernate-orm, jdbc-h2, narayana-jta, resteasy, resteasy-jsonb]

```

**Environment:**
```
Microsoft Windows [Version 10.0.17134.885]
java version "1.8.0_151"
INFO  [io.quarkus]] (executor-thread-1) Quarkus 0.19.1 started in 0.835s. Listening on: http://[::]:8080

```
** Response page snapshot**
As it returns a Content-Type:application/json + the html content the browser does not render the html page

![image](https://user-images.githubusercontent.com/461378/61339324-ae157e80-a813-11e9-8bbe-bc235f2c0efc.png)

gwenneg:
Hi @barata0!

[...]

PR #3309, opened 2019-07-23
fix: add a lesser priority than the default one

loicmathieu:
Adding a lesser priority than the default one for the NotFoundExceptionMapper allow a user to easily override it.

Be careful that for this to work the exception mapper needs to be of the same parameter type: `ExceptionMapper<NotFoundException>` if a broader parameter type is used like `ExceptionMapper<RuntimeException` it will not take precedence from the default one.

Tested on a local project but no test provided.

Refers to #3253

### DIFF
--- a/extensions/resteasy/runtime/src/main/java/io/quarkus/resteasy/runtime/NotFoundExceptionMapper.java
+++ b/extensions/resteasy/runtime/src/main/java/io/quarkus/resteasy/runtime/NotFoundExceptionMapper.java
@@ -13,8 +13,10 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
+import javax.annotation.Priority;
 import javax.ws.rs.NotFoundException;
 import javax.ws.rs.Path;
+import javax.ws.rs.Priorities;
 import javax.ws.rs.core.Context;
 import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
@@ -33,6 +35,7 @@ import io.quarkus.runtime.TemplateHtmlBuilder;
  * 
  */
 @Provider
+@Priority(Priorities.USER + 1)
 public class NotFoundExceptionMapper implements ExceptionMapper<NotFoundException> {
 
     @Context


kenfinnigan:
Is it possible to add an integration test for this?

It would be important to ensure this doesn't regress with other changes

loicmathieu:
@kenfinnigan I can add a testcase in the resteasy extension with the other test cases but I don't know how to create a test case that runs in dev mode ...

My current test case is as following but it didn't run as dev mode so it didn't implies any regression protection

```
public class CustomExceptionMapperTestCase {
    @RegisterExtension
    static QuarkusUnitTest runner = new QuarkusUnitTest()
            .setArchiveProducer(() -> ShrinkWrap.create(JavaArchive.class)
                    .addClasses(RootResource.class, CustomExceptionMapper.class));

    @Test
    public void testResourceNotFound() {
        RestAssured.when().get("/not_found")
                .then()
                    .statusCode(200)
                    .body(Matchers.is("not found but OK"));
    }
}
```

kenfinnigan:
@stuartwdouglas do we have a way to test DEV mode during the build?

How do we prevent regressions in DEV mode?

stuartwdouglas:
We have tests in the maven plugin integration tests. They are slow and don't really scale so the plan is to write a dev mode test framework however it is just something that nobody has got to yet.

This is the second time this has come up today so maybe I need to move it up my list.

loicmathieu:
@stuartwdouglas it's the one from the maven plugin like `io.quarkus.maven.it.DevMojoIT` ?

It's very complicate to setup (needs a whole project) and I don't think it is really needed to just test this kind of small functionalities.

I propose to skip the test untill we have an easy way to setup test in dev mode via `QuarkusUnitTest` ...

stuartwdouglas:
I agree

stuartwdouglas:
https://github.com/quarkusio/quarkus/pull/3317

kenfinnigan (approved):

PR #3317, opened 2019-07-24
Add QuarkusDevModeTest for testing hot reload

stuartwdouglas:
This allows extensions to write unit test cases for hot reload that
do not require a new maven project


stuartwdouglas:
Looks like there is a race here, I will investigate.

machi1990:
Can't wait to play with this. Thanks @stuartwdouglas. 

It should fix #3317

mkouba:
Would it make sense to add custom build chain customizers to `QuarkusDevModeTest`; i.e. something like https://github.com/quarkusio/quarkus/blob/master/test-framework/junit5-internal/src/main/java/io/quarkus/test/QuarkusUnitTest.java#L107-L110. This would basically allow an extension author to test a hot deployment feature based on a build item and not a concrete component (such as servlet).

stuartwdouglas:
I don't know, hot deployment is really a user facing feature, I think it makes sense to test it the same way a user would. Do you have a concrete example of where this would be useful?

mkouba:
The only example I have is a test for `HotDeploymentWatchedFileBuildItem.restartNeeded` - there is currently no built-in extension using this flag and so we can't verify the functionality inside the quarkus project.

stuartwdouglas:
I was thinking about creating a hot-deployment-tests module and moving some of the tests that are currently in DevMojoIT into this module to improve the build time. If we were going to do this we could likely just include some @BuildSteps in this module to test this functionality. 

Dev mode is not really setup for chain customisers, it would need to be passed down through a few different layers.

stuartwdouglas:
@emmanuelbernard can you review as you opened the original issue?

docs/src/main/asciidoc/extension-authors-guide.adoc
```diff
@@ -798,6 +798,109 @@ public class PersistenceAndQuarkusConfigTest {
 <1> This tells JUnit that the Quarkus deployment should fail with a specific exception
 
 
+=== Testing hot reload
+
+It is also possible to write tests that verify you extension works correctly in development mode and can correctly
```
kenfinnigan: "verify you" -> "verify an"

docs/src/main/asciidoc/extension-authors-guide.adoc
```diff
@@ -798,6 +798,109 @@ public class PersistenceAndQuarkusConfigTest {
...
+handle updates.
+
+For most extensions this will just work 'out of the box' however it is still a good idea to have a smoke test to
```
kenfinnigan: Add comma after 'out of the box'

docs/src/main/asciidoc/extension-authors-guide.adoc
```diff
@@ -798,6 +798,109 @@ public class PersistenceAndQuarkusConfigTest {
...
+
+        RestAssured.when().get("/new.txt").then()
+                .statusCode(200)
+                .body(is("New File"));
+
+    }
+}
+----
+
+<1> This is the starting deployment, your test can modify it as part of the test suite. Quarkus will be restarted between
```
kenfinnigan: "This starts the deployment"

docs/src/main/asciidoc/extension-authors-guide.adoc
```diff
@@ -798,6 +798,109 @@ public class PersistenceAndQuarkusConfigTest {
...
+each test method so every method starts with a clean deployment.
+
+<2> This method allows you to modify the source of a class file. The old source is passed into the function, and the source
```
kenfinnigan: "and the updated source is returned"

kenfinnigan (changes requested):
Just some minor verbiage suggestions

geoand (approved):

geoand:
@machi1990 would you like to check and see sometime if some the current maven dev mode tests could be replaced with the new test framework?

gwenneg:
@stuartwdouglas I tried to use `QuarkusDevModeTest` in a slightly different way from the example you added in the guide, but I got a NPE while using an injected bean:
```
public class DevModeCachedServiceChangeTest {

    @RegisterExtension
    final static QuarkusDevModeTest TEST = new QuarkusDevModeTest()
            .setArchiveProducer(() -> ShrinkWrap.create(JavaArchive.class).addClass(DevModeCachedService.class));

    @Inject
    DevModeCachedService cachedService; // @ApplicationScoped bean

    @Test
    public void testServiceChange() throws InterruptedException {
        cachedService.doSomething(); // NPE thrown here
    }
}
```
Is this supposed to work or do I have to put my service behind a servlet or a JAX-RS resource?

stuartwdouglas:
You can't inject into dev mode tests, it would be complicated to do as dev
mode creates its own class loader for the beans (to allow restart) and
every restart the test would need to be re-injected.

gwenneg:
I'll adapt my test then, thanks for your answer!

gwenneg:
I replaced the bean injection with a JAX-RS resource and now my hot reload test works like a charm. It could be helpful to add a few lines in the guide about the need to add the following dependency for extensions that don't already have `quarkus-resteasy` in their `compile` scope:

```
<dependency>
    <groupId>io.quarkus</groupId>
    <artifactId>quarkus-resteasy-deployment</artifactId>
    <scope>test</scope>
</dependency>
```

Or the servlet equivalent.

PR #3323, opened 2019-07-24
Introduce the ability to configure Jsonb and generate serializers for JAX-RS return types

geoand:


extensions/resteasy-jsonb/deployment/src/main/java/io/quarkus/resteasy/jsonb/deployment/JsonbConfig.java
```diff
@@ -0,0 +1,72 @@
...
+            new HashSet<>(Arrays.asList(PropertyOrderStrategy.LEXICOGRAPHICAL, PropertyOrderStrategy.ANY,
+                    PropertyOrderStrategy.REVERSE)));
+
+    /**
+     * If enabled, Quarkus will a create a JAX-RS resolves that configures JSON-B with the properties
+     * specified here
+     * It will also attempt to generate serializers for JAX-RS return types
+     */
+    @ConfigItem(defaultValue = "false")
+    boolean enabled;
```
geoand: I have elected to have the feature off by default - I'm not sure it's the correct choice so I am open to suggestions

extensions/resteasy-jsonb/deployment/src/main/java/io/quarkus/resteasy/jsonb/deployment/JsonbConfig.java
```diff
@@ -0,0 +1,61 @@
...
+     * encoding to use when deserializing data
+     */
+    @ConfigItem(defaultValue = "")
+    String encoding;
+
+    /**
+     * specified whether unknown properties will cause deserialization to fail
+     */
+    @ConfigItem(defaultValue = "true")
+    boolean failOnUnknownProperties;
```
loicmathieu: Today Yasson don't fail on unknown properties so `false` should be the default. IMHO it's a good default for microservices because it allow a service to "remove" an attribute without changing the client.

extensions/resteasy-jsonb/deployment/src/main/java/io/quarkus/resteasy/jsonb/deployment/JsonbConfig.java
loicmathieu: IMHO the configuration of JSON-B is a long waiting functionality and cannot cause any regression if the default are the Yasson one so it should always apply.
The boolean should only apply to the generation of serializers, this could be disabled by default (features in BETA) and enabled after some feedback from early adopters

extensions/resteasy-jsonb/deployment/src/main/java/io/quarkus/resteasy/jsonb/deployment/JsonbConfig.java
geoand: According to [this](https://github.com/eclipse-ee4j/yasson/blob/master/src/main/java/org/eclipse/yasson/YassonProperties.java#L23), the default value is true.

I agree with you that `false` is probably a better default and I  am inclined to change it. Let's see what others think as well

extensions/resteasy-jsonb/deployment/src/main/java/io/quarkus/resteasy/jsonb/deployment/JsonbConfig.java
geoand: I pondered that option myself as well. Honestly I think what you say makes a lot of sense and I wil end up doing it unless others object

extensions/resteasy-jsonb/deployment/src/main/java/io/quarkus/resteasy/jsonb/deployment/JsonbConfig.java
loicmathieu: It's strange because I tested it before commenting and I can send unknown properties to a restendpoint and yasson don't generate any error.

geoand: In any case I agree that `false` should be the default,  but I would like to see what others think about changing the default value of well known library

geoand:
A fairly obvious improvement that can come after this is to provide a Jsonb default bean.
That bean could be used globally wherever it's needed - of course the generated ContextResolver would need to be changed to accommodate this and not create jsonb from scratch

geoand:
I will also be adding JMH results comparing the generated serializer to default Yasson soon

geoand:
The latest push adds `Jsonb` bean registration as well

geoand:
Here is what I used as a JMH benchmark:

```
@BenchmarkMode(Mode.Throughput)
@Timeout(time = 20)
@State(Scope.Benchmark)
@Warmup(iterations = 5)
@OutputTimeUnit(TimeUnit.MILLISECONDS)
public class CatTest {

    private Jsonb jsonbStock;
    private Jsonb jsonbOptimized;

    private Cat cat;

    @Setup(Level.Trial)
    public void setUp() {
        jackson = new ObjectMapper();
        jsonbStock = JsonbBuilder.create();
        jsonbOptimized = TestUtil.getConfiguredJsonb();
        cat = new Cat("grey", 1, "Persian");
    }

    @Benchmark
    public void testOptimized(Blackhole blackhole) {
        blackhole.consume(jsonbOptimized.toJson(cat));
    }

    @Benchmark
    public void testStock(Blackhole blackhole) {
        blackhole.consume(jsonbStock.toJson(cat));
    }
}
```

`Cat` simply has 3 fields, nothing special going on.

And here is what are the results I got:

```
Result "org.acme.CatTest.testOptimized":
  2968.344 ±(99.9%) 37.999 ops/ms [Average]
  (min, avg, max) = (2878.414, 2968.344, 3057.704), stdev = 50.728
  CI (99.9%): [2930.345, 3006.343] (assumes normal distribution)
```

and

```
Result "org.acme.CatTest.testStock":
  2507.857 ±(99.9%) 33.402 ops/ms [Average]
  (min, avg, max) = (2403.747, 2507.857, 2587.101), stdev = 44.591
  CI (99.9%): [2474.455, 2541.259] (assumes normal distribution)
```

The real performance benefit however can be seen when using a complex object that contains lists, maps, arrays, various fields etc. etc..

A run JMH run I got from testing such an object was:

```
Result "org.acme.CoffeeTest.testOptimized":
  131.295 ±(99.9%) 2.551 ops/ms [Average]
  (min, avg, max) = (125.866, 131.295, 139.503), stdev = 3.406
  CI (99.9%): [128.744, 133.846] (assumes normal distribution)
```

and 

```
Result "org.acme.CoffeeTest.testStock":
  73.423 ±(99.9%) 1.280 ops/ms [Average]
  (min, avg, max) = (69.618, 73.423, 76.052), stdev = 1.708
  CI (99.9%): [72.143, 74.702] (assumes normal distribution)
```

geoand:
I forced pushed the PR to resolve conflicts with master

gsmet:
@geoand removed his branch by mistake, he will create another PR soon.

geoand:
Followed up by #3553

PR #3325, opened 2019-07-25
Document gmail smtp configuration

### LINKED ISSUE #3039: Extend mail client to support Gmail SMTP server
This issue covers:

* the test of the mail client with Gmail SMTP server
* the update of the guide to explain the required configuration

Requires Vert.x 3.8

Dufgui:
fix #3039

### DIFF
--- a/docs/src/main/asciidoc/sending-emails.adoc
+++ b/docs/src/main/asciidoc/sending-emails.adoc
@@ -274,7 +274,12 @@ quarkus.mailer.port=465
 quarkus.mailer.ssl=true
 quarkus.mailer.username=YOUREMAIL@gmail.com
 quarkus.mailer.password=YOURGENERATEDAPPLICATIONPASSWORD
----- 
+----
+
+== Using SSL with native executables
+
+Note that if you enable SSL for the mailer and you want to build a native executable, you will need to enable the SSL support.
+Please refer to the native-and-ssl-guide.html[Using SSL With Native Executables] guide for more information.
 
 == Using the underlying Vert.x Mail Client
 
@@ -301,6 +306,3 @@ You can also create your own instance, and pass your own configuration.
 
 This guide has shown how you can send emails from a Quarkus application.
 The _mailer_ extension works in JVM and native mode.
-
-
-
--- a/docs/src/main/asciidoc/sending-emails.adoc
+++ b/docs/src/main/asciidoc/sending-emails.adoc
@@ -246,6 +246,36 @@ class MailTest {
 }
 ----
 
+== Gmail specific configuration
+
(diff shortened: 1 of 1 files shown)


docs/src/main/asciidoc/sending-emails.adoc
```diff
@@ -246,6 +246,36 @@ class MailTest {
 }
 ----
 
+== Gmail specific configuration
+
+If you want to use the gmail smtp server, first, please create a dedicated password in `google account > security > application password` or go to https://myaccount.google.com/apppasswords
```
gsmet:
```suggestion
If you want to use the Gmail SMTP server, first create a dedicated password in `Google Account > Security > App passwords` or go to https://myaccount.google.com/apppasswords.
```

docs/src/main/asciidoc/sending-emails.adoc
```diff
@@ -246,6 +246,36 @@ class MailTest {
 }
 ----
 
+== Gmail specific configuration
+
+If you want to use the gmail smtp server, first, please create a dedicated password in `google account > security > application password` or go to https://myaccount.google.com/apppasswords 
+
+When it done you can use it in your configuration.
```
gsmet:
```suggestion
When done, you can configure your Quarkus application by adding the following properties to your `application.properties`: 
```

gsmet (changes requested):
Thanks for taking the time to test that.

I added some wording suggestions, could you include them and squash your commits in one?

Thanks!

docs/src/main/asciidoc/sending-emails.adoc
```diff
@@ -246,6 +246,36 @@ class MailTest {
...
+----
+
+Or with SSL:
+
+[source]
+----
+quarkus.mailer.from=YOUREMAIL@gmail.com
+quarkus.mailer.host=smtp.gmail.com
+quarkus.mailer.port=465
+quarkus.mailer.ssl=true
```
machi1990: this will work only in combination with `quarkus.ssl.native=true` .

Dufgui: I can add it in the doc but I propose to change `SslNativeConfigBuildItem` as a `MultiBuildItem` and produce them in the `MailerProcessor` if the `MailConfig.ssl` is enable ? Are you ok @gsmet @machi1990  ?
Else I just document it....

gsmet: I added a note about the SSL support and will merge.

I wonder if we should enable SSL by default when the Mailer extension is used. @cescoffier WDYT?

gsmet (approved):

PR #3333, opened 2019-07-25
Conditional build step inclusion

### LINKED ISSUE #2962: @IncludeWhen(LaunchMode.NORMAL)
**Description**
I think there's a few different places where we have @BuildSteps that we want to include or exclude only for a specific launch mode.  It bothers me that I have to include a LaunchModeBuildItem as a parameter to my method and then do an if statement to determine whether or not to run the build step.  THis is mostly a cosmetic suggestion, but it would be nice to have this encapsulated within an annotation:

@BuildStep
@IncludeWhen(LaunchMode.NORMAL)

or

@BuildStep
@ExcludeWhen(LaunchMode.DEV)

### LINKED ISSUE #3233: Allow extensions and/or build steps to be disabled by config
We should allow build steps to be skipped or excluded based on configuration.  Right now the build time configuration is being set up during the build process, but there is really no reason to do it this way; we could possibly have a utility to set it up the configuration ahead of time which would make this possible.

We could add annotations that allow some kind of predicate class to be specified which determines whether the step should be included.  This would allow steps to be included on other bases as well.

dmlloyd:
Fixes #3233, fixes #2962

### DIFF
--- a/core/builder/src/main/java/io/quarkus/builder/BuildStepBuilder.java
+++ b/core/builder/src/main/java/io/quarkus/builder/BuildStepBuilder.java
@@ -3,6 +3,7 @@ package io.quarkus.builder;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Set;
+import java.util.function.BooleanSupplier;
 
 import org.wildfly.common.Assert;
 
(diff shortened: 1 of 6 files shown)


kenfinnigan:
@stuartwdouglas are you able to review this?

stuartwdouglas (approved):

PR #3358, opened 2019-07-27
Add kotlin to search path for gradle tests

oztimpower:
Gradle seperates java and kotlin classes.

Test classes search path needs to add kotlin for tests created using kotlin.


geoand:
Thanks for the PR! Could you please squash into a single commit?

PR #3359, opened 2019-07-27
Add kotlin to search path for test classes using gradle

oztimpower:
Gradle seperates classes from java and kotlin (and scala et al).

Test class search path must explicitly add kotlin, in addition to java which is already set.

### DIFF
--- a/devtools/gradle/src/main/java/io/quarkus/gradle/QuarkusPluginExtension.java
+++ b/devtools/gradle/src/main/java/io/quarkus/gradle/QuarkusPluginExtension.java
@@ -98,7 +98,7 @@ public class QuarkusPluginExtension {
 
     public String finalName() {
         if (finalName == null || finalName.length() == 0)
-            return project.getName();
+            return project.getName() + "-" + project.getVersion();
         else
             return finalName;
     }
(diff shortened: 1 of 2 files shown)


gsmet:
@geoand I think this one all makes sense but I let you review it?

geoand:
@gsmet sure thing!

test-framework/common/src/main/java/io/quarkus/test/common/PathTestHelper.java
```diff
@@ -24,6 +24,10 @@
         TEST_TO_MAIN_DIR_FRAGMENTS.put(
                 "classes" + File.separator + "java" + File.separator + "test",
                 "classes" + File.separator + "java" + File.separator + "main");
+        TEST_TO_MAIN_DIR_FRAGMENTS.put(
```
geoand: Looks great, :+1:

devtools/gradle/src/main/java/io/quarkus/gradle/QuarkusPluginExtension.java
```diff
@@ -98,7 +98,7 @@ public void setLibDir(String libDir) {
 
     public String finalName() {
         if (finalName == null || finalName.length() == 0)
-            return project.getName();
+            return project.getName() + "-" + project.getVersion();
```
geoand: Is there any specific reason this was changed?

gsmet: Yeah, see the full commit message of this commit.

geoand: Duh, I should have checked that :)

geoand (approved):

gsmet:
Merged, thanks!

PR #3367, opened 2019-07-29
Remove static driver instance.

michael-simons:
I just noticed that the Neo4j recorder could have been more simplified during the initial PR. As the framework makes sure, the recorder is never invoked twice after recording, we removed all checks whether the driver is initialized, but not the static instance variable in the recorder.
This can go away,  too.


gsmet (approved):
Thanks. Let's wait for CI.

PR #3368, opened 2019-07-29
Offer `no-file` option to override `import.sql`

emmanuelbernard:
This is to address https://github.com/quarkusio/quarkusio.github.io/pull/254#discussion_r306181911 to make it clearer.

Right now the user would use `quarkus.hibernate-orm.sql-load-script = no-file`

Another option would be to have different defaults between `DEV` and `TEST` mode using `import.sql` by default and `PROD` use `no-file`. This is not implemented but I was wondering about this idea. @FroMage seems to like it

### DIFF
--- a/docs/src/main/asciidoc/hibernate-orm-guide.adoc
+++ b/docs/src/main/asciidoc/hibernate-orm-guide.adoc
@@ -137,6 +137,7 @@ The storage engine to use when the dialect supports multiple storage engines.
 `quarkus.hibernate-orm.sql-load-script`::
 (defaults to `/import.sql`) Name of the file containing the SQL statements to execute when Hibernate ORM starts.
 By default, simply add `import.sql` in the root of your resources directory and it will be picked up without having to set this property.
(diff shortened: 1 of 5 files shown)


gsmet:
Build failed due to an invalid import order.

Just asking: you don't have a warning saying that the import file does not exist when you're pointing to a non-existing file?

emmanuelbernard:
How comes the build does not fail for me?
`mvn clean install`

No I did not see any warning. I also looked at the code path of ORM but did not see a warning.

mkouba:
AFAIK `mvn clean install` should not fail but format the code correctly instead. Are you sure you did run the build before you've changed the `NoFileOptionTestCase` class?

emmanuelbernard:
you were right @mkouba I committed before running clean install

mkouba:
Looks good to me but it would be nice to have a review from a "data" specialist ;-). @Sanne @gsmet @FroMage?

gsmet:
Yeah, it's on my list for this morning. That's why I added the milestone.

gsmet (approved):

PR #3373, opened 2019-07-29
Use host string instead of address in SyslogHandler

asoldano:
in case only hostname is setup, this prevents test failures (CC @gsmet )


dmlloyd (approved):

gsmet:
For history, the issue is that we can end up with an unresolved `InetSocketAddress` and, in this very case, `getAddress()` returns null. Thus Alessio had a NPE.

gsmet (approved):

asoldano:
Thanks @dmlloyd and @gsmet

PR #3375, opened 2019-07-30
Add clarification to the testing part of the kubernetes client doc

geoand:
This was done in response to the question raised [here](https://quarkusio.zulipchat.com/#narrow/stream/187038-dev/topic/Kubernetes.20Client)

### DIFF
--- a/docs/src/main/asciidoc/kubernetes-client.adoc
+++ b/docs/src/main/asciidoc/kubernetes-client.adoc
@@ -73,7 +73,30 @@ To make testing against a mock Kubernetes API extremely simple, Quarkus provides
 a mock of the Kubernetes API server and sets the proper environment variables needed so that the Kubernetes Client configures itself to use said mock.
 Tests can inject the mock and set it up in any way necessary for the particular testing using the `@MockServer` annotation.
 
-Here is an example of such test:
+Lets assume we have a REST endpoint defined like so:
+
+[source%nowrap,java]
+----
+
+@Path("/pod")
+public class Pods {
+
+    private final KubernetesClient kubernetesClient;
+
+    public Pods(KubernetesClient kubernetesClient) {
+        this.kubernetesClient = kubernetesClient;
+    }
+
+    @GET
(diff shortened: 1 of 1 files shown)


docs/src/main/asciidoc/kubernetes-client.adoc
```diff
@@ -73,7 +73,30 @@ To make testing against a mock Kubernetes API extremely simple, Quarkus provides
 a mock of the Kubernetes API server and sets the proper environment variables needed so that the Kubernetes Client configures itself to use said mock.
 Tests can inject the mock and set it up in any way necessary for the particular testing using the `@MockServer` annotation.
 
-Here is an example of such test:
+Lets assume we have a REST endpoint defined like so:
```
gsmet:
```suggestion
Let's assume we have a REST endpoint defined like so:
```

docs/src/main/asciidoc/kubernetes-client.adoc
```diff
@@ -73,7 +73,30 @@ To make testing against a mock Kubernetes API extremely simple, Quarkus provides
...
+    @GET
+    @Produces(MediaType.APPLICATION_JSON)
+    @Path("/{namespace}")
+    public List<Pod> pods(@PathParam("namespace") String namespace) {
+        return kubernetesClient.pods().inNamespace(namespace).list().getItems();
+    }
+}
+----
+
+We could write a test for this endpoint very easily like so:
```
gsmet: Does the URL used in the test matches the URL in the resource?

It doesn't look like it but maybe I'm missing something?

gsmet (changes requested):
Added one typo fix and one question.

docs/src/main/asciidoc/kubernetes-client.adoc
geoand: The RestAssured test and the JAX-RS endpoint do seem to match. The other URLs are the Kubernetes API URLs that are mocked

gsmet (approved):

PR #3379, opened 2019-07-30
Upgrade to Hibernate ORM 5.4.4

Sanne:


gsmet:
I'm wondering if we should merge this one just after the release so that we have a bit more hindsight.

Merging this just the day before the release makes me a bit nervous.

WDYT?

Sanne:
I'd merge it, it's awesomely better. Also .. which one is the experimental project ;)

gsmet (approved):
Waiting for CI.

PR #3393, opened 2019-07-31
Make RegisterForReflection annotation inheritable

gastaldi:
This allows setting this annotation in interfaces and superclasses. Useful for annotation-based generators like http://immutables.github.io/


gastaldi:
Doesn't work, closing

gsmet:
@gastaldi I think it's not a good idea to register all the implementations/subclasses of interfaces/classes marked with `@RegisterForReflection`. It probably should be done via Jandex lookups.

PR #3154, opened 2019-07-09
Native executable tests should be run under the test profile

machi1990:
Fixes #3148

PR #3398, opened 2019-08-01
Change default value of `sql-load-script` to `no-file` in PROD mode

### LINKED ISSUE #3389: Change default value of `sql-load-script` to `no-file` in PROD mode
While having `import.sql` imported for dev and test makes sense, I have the feeling that making it an explicit choice in PROD is likely worthwile.
What do you think @gsmet @Sanne ?

It's a restart on the question asked https://github.com/quarkusio/quarkus/pull/3368 that got unaddressed.

gwenneg:
Fixes #3389 

WDYT of the doc changes @gsmet?

### DIFF
--- a/extensions/hibernate-orm/deployment/src/main/java/io/quarkus/hibernate/orm/deployment/HibernateOrmProcessor.java
+++ b/extensions/hibernate-orm/deployment/src/main/java/io/quarkus/hibernate/orm/deployment/HibernateOrmProcessor.java
@@ -60,6 +60,7 @@ import io.quarkus.deployment.builditem.FeatureBuildItem;
 import io.quarkus.deployment.builditem.GeneratedClassBuildItem;
 import io.quarkus.deployment.builditem.GeneratedResourceBuildItem;
 import io.quarkus.deployment.builditem.HotDeploymentWatchedFileBuildItem;
+import io.quarkus.deployment.builditem.LaunchModeBuildItem;
 import io.quarkus.deployment.builditem.SystemPropertyBuildItem;
 import io.quarkus.deployment.builditem.substrate.ReflectiveClassBuildItem;
 import io.quarkus.deployment.builditem.substrate.SubstrateResourceBuildItem;
@@ -80,6 +81,7 @@ import io.quarkus.hibernate.orm.runtime.TransactionEntityManagers;
 import io.quarkus.hibernate.orm.runtime.boot.scan.QuarkusScanner;
 import io.quarkus.hibernate.orm.runtime.dialect.QuarkusH2Dialect;
 import io.quarkus.hibernate.orm.runtime.dialect.QuarkusPostgreSQL95Dialect;
+import io.quarkus.runtime.LaunchMode;
 
 /**
  * Simulacrum of JPA bootstrap.
@@ -105,10 +107,10 @@ public final class HibernateOrmProcessor {
     HibernateOrmConfig hibernateConfig;
 
     @BuildStep
-    List<HotDeploymentWatchedFileBuildItem> hotDeploymentWatchedFiles() {
+    List<HotDeploymentWatchedFileBuildItem> hotDeploymentWatchedFiles(LaunchModeBuildItem launchMode) {
         List<HotDeploymentWatchedFileBuildItem> watchedFiles = new ArrayList<>();
         watchedFiles.add(new HotDeploymentWatchedFileBuildItem("META-INF/persistence.xml"));
-        getSqlLoadScript().ifPresent(script -> {
+        getSqlLoadScript(launchMode.getLaunchMode()).ifPresent(script -> {
             watchedFiles.add(new HotDeploymentWatchedFileBuildItem(script));
         });
         return watchedFiles;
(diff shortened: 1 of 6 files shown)


gwenneg:
Sorry, I didn't expect the native build to fail on this while the JDK build is successful. I'll fix the error ASAP.

integration-tests/jpa/src/main/java/io/quarkus/it/jpa/configurationless/CRUDResource.java
```diff
@@ -76,10 +79,10 @@
     @Transactional
     @Path("/import")
     public Map<String, String> get() {
+        boolean isProdMode = ProfileManager.getActiveProfile().equals(LaunchMode.NORMAL.getDefaultProfile());
```
gwenneg: I'm not very satisfied with this but I didn't find a cleaner way to fix this test...

machi1990: Is it because the native tests are running under the `prod` profile?

gwenneg: Now that you asked the question @machi1990, this might be linked to the #3154 reversal in #3265... (even if I tested #3265 in native mode and everything seemed fine at that time)... This needs to be investigated.

gwenneg: I'm pretty sure reverting #3154 was a bad idea so I put it back (a modified version since there's now the `quarkus.test.profile` property to consider) and I got positive results locally on tests done from a quickstarts project, but CI is still not happy (the native test is running under the `prod` profile)...

Sorry about reverting your commit @machi1990 :)

machi1990: > I'm pretty sure reverting #3154 was a bad idea so I put it back (a modified version since there's now the `quarkus.test.profile` property to consider) and I got positive results locally on tests done from a quickstarts project, but CI is still not happy (the native test is running under the `prod` profile)...

I think this is because the image is generated with a `prod` profile. Reverting the commit in question will only help with properties that are available at runtime. 

I think we'll need investigating a bit more: @stuartwdouglas detailed the problem better here https://github.com/quarkusio/quarkus/pull/3154#issuecomment-509878882.

gwenneg: @gsmet I have a few questions for you. I've been dealing with a native tests profile issue here which will require further investigations.

- I think we should limit this PR content to the original issue only and accept the `CRUDResource` workaround. WDYT?
- With the current `master` code, native tests are always run under the `prod` profile because I reverted #3154 in #3265. If we put back #3154 (the current version of `NativeImageLauncher` in this PR), we will have limited support for `test` profile in native tests, depending on when the profile is used (build time or runtime). Should we put it back or wait for the investigations and offer a full `test` profile support only if it's possible and nothing otherwise? In both cases, I think we need to put a warning in the `maven-tooling` and `gradle-tooling` guides about this native tests issue.

gwenneg: Here's the `CRUDResource` workaround since it's not clear in the current PR content:
```
    public Map<String, String> get() {
        boolean isProdMode = ProfileManager.getActiveProfile().equals(LaunchMode.NORMAL.getDefaultProfile());
        Gift gift = em.find(Gift.class, 100000L);
        Map<String, String> map = new HashMap<>();
        map.put("jpa", gift != null || isProdMode ? "OK" : "Boooo");
        return map;
    }
```
I had to add `isProdMode` to make the native test pass.

gwenneg:
@emmanuelbernard Since it seems @gsmet is away, could you take a look at the comments above and tell me what you'd like me to do about this PR?

gwenneg:
@gsmet I still don't know what to do about this PR :)

gwenneg:
@gsmet Following #3680 merge, I rebased this PR and separated the profile issue from the initial content which was written to fix #3389. If you could review the current version of the PR, we could fix #3389 and possibly create another issue/PR to discuss/fix the test profile issue in native mode.

gwenneg:
I didn't see there were new `spring-data-jpa` tests that use an `import.sql` file. The same prod mode check as in `CRUDResource` will have to be used in all these tests (if it's OK to use it).

extensions/hibernate-orm/deployment/src/main/java/io/quarkus/hibernate/orm/deployment/HibernateOrmProcessor.java
```diff
@@ -328,9 +332,10 @@ public void startPersistenceUnits(HibernateOrmRecorder recorder, BeanContainerBu
         recorder.startAllPersistenceUnits(beanContainer.getValue());
     }
 
-    private Optional<String> getSqlLoadScript() {
+    private Optional<String> getSqlLoadScript(LaunchMode launchMode) {
         // Explicit file or default Hibernate ORM file.
-        String script = hibernateConfig.sqlLoadScript.orElse("import.sql");
+        String script = hibernateConfig.sqlLoadScript
+                .orElseGet(() -> launchMode == LaunchMode.NORMAL ? NO_SQL_LOAD_SCRIPT_FILE : "import.sql");
```
gsmet: Hmmm, why do you need a lambda here?

gsmet (changes requested):

extensions/hibernate-orm/deployment/src/main/java/io/quarkus/hibernate/orm/deployment/HibernateOrmProcessor.java
gwenneg: I wanted to avoid a lot of `if else` blocks and I thought lambdas were OK in deployment code, but it's gone now :)

gwenneg:
@gsmet In case you didn't see my previous message, this PR should be ready for a merge now.

gsmet (approved):

gsmet:
Merged, thanks, sorry for the delay.

Issue #1413, opened 2019-03-11
Scala language support


vertexclique:
In addition to Kotlin, maybe it's also good to have Scala supported. Adding Scala source discovery and having it to compile an Akka application would be a nice way to provide a starting ground for Scala.

gsmet:
@vertexclique would you be interested in taking a stab at it?

The PR for the Kotlin addition is here: https://github.com/quarkusio/quarkus/pull/1059 .

marcobattaglia:
@geoand @vertexclique
You can find an example of using quarkus, scala and kotlin together here:
[https://github.com/marcobattaglia/quarkus-scala](https://github.com/marcobattaglia/quarkus-scala)

We are working on testing other quarkus features: jpa, reactiveX and so on.

If you think that it needs some work in order to provide a better integration with quarkus we can work on it.

geoand:
@marcobattaglia thanks for the pointer!

I think what @vertexclique was asking for was support for the dev mode with Scala resources

vertexclique:
> I think what @vertexclique was asking for was support for the dev mode with Scala resources

Yes, also the support dependency like how kotlin has been done.

gabfssilva:
Scala is a bit different from Kotlin since it uses its own Option<T>, List<T>, Future<T> and so on, so maybe it needs more efford than Kotlin to implement expressive Scala with Quarkus.

wiradikusuma:
I use Grade 5.2.1 and followed https://quarkus.io/guides/gradle-tooling but I still can't make `gradlew quarkusDev` to run on my Java/Scala project.

First, it complains when the source dir is separate (`src/main/java` and `src/main/scala`). When I merged them together (`src/main/java`), it complains that it can't find some stuff that the Java classes rely on (basically, `gradlew quarkusDev` completely ignores Scala files).

Is https://github.com/quarkusio/quarkus/pull/3399 supposed to fix this? If yes, which release will have that?

geoand:
The Scala PR has been merged but there is no release that contains it yet.
0.22.0 which is the next release will have this support

forevergenin:
Looks like 0.22 release doesn't include scala support.

geoand:
That is true.

0.22 turned out to be an intentionally small release because we only wanted to get some of the stuff out.
Stay tuned for newer Quarkus releases soon 😉

aesteve:
Tried it with 0.23.1 seems to be working really fine in dev/jar mode!

[...]

PR #3399, opened 2019-08-01
Scala language support

miteshaghera:
Initial support for Scala.

Fix for #1413


gsmet:
Hi @miteshaghera ,

Thanks for this. Looks like we have some test failures:
```

**Comment by miteshaghera** [comment]
> Hi @miteshaghera ,
> 
> Thanks for this. Looks like we have some test failures:
> 
> ```
> [INFO] Quarkus - Maven Plugin ............................. FAILURE [ 28.296 s]
> ```

I fixed for `Quarkus - Maven Plugin` but now it's failing for `Quarkus - Gradle Plugin`. By looking into [logs](url) I found following exception:

    Caused by: org.gradle.internal.resolve.ModuleVersionNotFoundException: Could not find io.quarkus:quarkus-bootstrap-parent:999-SNAPSHOT.
    Searched in the following locations:
      - file:/D:/a/1/s/devtools/gradle/target/dependencies/io/quarkus/quarkus-bootstrap-parent/999-SNAPSHOT/maven-metadata.xml
      - file:/D:/a/1/s/devtools/gradle/target/dependencies/io/quarkus/quarkus-bootstrap-parent/999-SNAPSHOT/quarkus-bootstrap-parent-999-SNAPSHOT.pom
      - file:/D:/a/1/s/devtools/gradle/target/dependencies/io/quarkus/quarkus-bootstrap-parent/999-SNAPSHOT/quarkus-bootstrap-parent-999-SNAPSHOT.jar
      - https://repo.maven.apache.org/maven2/io/quarkus/quarkus-bootstrap-parent/999-SNAPSHOT/maven-metadata.xml
      - https://repo.maven.apache.org/maven2/io/quarkus/quarkus-bootstrap-parent/999-SNAPSHOT/quarkus-bootstrap-parent-999-SNAPSHOT.pom
      - https://repo.maven.apache.org/maven2/io/quarkus/quarkus-bootstrap-parent/999-SNAPSHOT/quarkus-bootstrap-parent-999-SNAPSHOT.jar

Do I need to change anything to support scala?

File: devtools/common/src/main/filtered/extensions.json [hunk 1]
```diff
@@ -601,5 +601,14 @@
...
     "guide": "https://quarkus.io/guides/using-vertx"
+  },
+  {
+    "name": "Scala",
+    "labels": [
+      "scala"
+    ],
+    "groupId": "io.quarkus",
+    "artifactId": "quarkus-scala",
+    "guide": "https://quarkus.io/guides/scala"
```
gsmet: Let's remove this last entry as we don't have a guide yet.

gsmet (changes requested):
@miteshaghera looks good to me except for a small comment.

Could you fix it? If you can squash everything into one commit, that would be appreciated too.

In the meantime, I'll try to create a Scala project and see how it goes.

core/deployment/src/main/java/io/quarkus/deployment/builditem/FeatureBuildItem.java
```diff
@@ -60,6 +60,7 @@
     public static final String UNDERTOW_WEBSOCKETS = "undertow-websockets";
     public static final String VERTX = "vertx";
     public static final String VERTX_WEB = "vertx-web";
+    public static final String SCALA = "scala";
```
gsmet: Ah and another minor additional comment: please respect the alphabetical order.

miteshaghera: Fixed it

devtools/common/src/main/filtered/extensions.json
miteshaghera: Removed the entry and squashed them.

devtools/common/src/main/java/io/quarkus/generators/SourceType.java
```diff
@@ -68,4 +73,8 @@ private String computeTemplateFile(String genericTemplate, String templateName)
     private String computeTemplateFile(String genericTemplate, String templateName, String fileName) {
         return String.format(genericTemplate, templateName, getPathDiscriminator(), fileName);
     }
+
+    public static Optional<SourceType> parse(String extension) {
+        return Arrays.stream(values()).filter(e -> e.extension.contains(extension)).findAny();
```
gsmet: This won't work. The extension we're talking about is the file extension so in the case of Kotlin, it's `.kt` so it doesn't contain `kotlin`.

I think I would have a method in the enum matching the extension list itself so that we can take a decision based on the list of extensions (it won't be useful for now but we could imagine it could).

devtools/common/src/main/resources/templates/basic-rest/scala/pom-template.ftl
```diff
@@ -0,0 +1,171 @@
+<project xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://maven.apache.org/POM/4.0.0"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <groupId>${project_groupId}</groupId>
+    <artifactId>${project_artifactId}</artifactId>
+    <version>${project_version}</version>
```
gsmet: AFAICS, this file should be `pom.xml-template.ftl`.

Otherwise, when doing:
```
mvn io.quarkus:quarkus-maven-plugin:999-SNAPSHOT:create \
    -DprojectGroupId=org.acme \
    -DprojectArtifactId=rest-scala \
    -DclassName="org.acme.rest.GreetingResource" \
    -Dpath="/greeting" \
    -Dextensions="scala,resteasy-jsonb"
```

I have the following error:

> [ERROR] Failed to execute goal io.quarkus:quarkus-maven-plugin:999-SNAPSHOT:create (default-cli) on project standalone-pom: Template resource is missing: /templates/basic-rest/scala/pom.xml-template.ftl -> [Help 1]

I wonder if the kotlin template is currently broken by the way as you copied the structure from the Kotlin one. We should check that, once you have fixed the extension/template detection.

gsmet (changes requested):
OK I tried to create a project and found a couple more issues.

Could you check what's going on?

Thanks!

devtools/common/src/main/java/io/quarkus/generators/SourceType.java
miteshaghera: I have fixed it by comparing it with name.

devtools/common/src/main/resources/templates/basic-rest/scala/pom-template.ftl
miteshaghera: I have renamed the file for Kotlin as well.

miteshaghera:
> @miteshaghera looks good to me except for a small comment.
> 
> Could you fix it? If you can squash everything into one commit, that would be appreciated too.
> 
> In the meantime, I'll try to create a Scala project and see how it goes.

I have to add entry in `extensions.json` to make build successful for `quarkus:create` . However I omitted `guide` key as we don't have guide yet.

devtools/common/src/main/filtered/extensions.json
```diff
@@ -230,6 +230,14 @@
...
     "guide": "https://quarkus.io/guides/kotlin"
   },
+  {
+    "name": "Scala",
+    "labels": [
+      "scala"
+    ],
+    "groupId": "io.quarkus",
+    "artifactId": "quarkus-scala"
+  },
```
gsmet: Please let's respect the alphabetical order.

devtools/common/src/main/java/io/quarkus/generators/SourceType.java
```diff
@@ -68,4 +73,8 @@ private String computeTemplateFile(String genericTemplate, String templateName)
     private String computeTemplateFile(String genericTemplate, String templateName, String fileName) {
         return String.format(genericTemplate, templateName, getPathDiscriminator(), fileName);
     }
+
+    public static Optional<SourceType> parse(String extension) {
+        return Arrays.stream(values()).filter(e -> extension.contains(e.name().toLowerCase())).findAny();
```
gsmet: I really would prefer us having a match method for each `SourceType`. Using the enum name is really fragile as we could change it without thinking it would break the behavior.

gsmet (changes requested):
I added a couple more comments.

Could you also add a test of project creation? Considering it was broken, we really need one. See my comment here: https://github.com/quarkusio/quarkus/pull/3542#issuecomment-521933968 .

Thanks!

devtools/common/src/main/filtered/extensions.json
miteshaghera: Fixed it

miteshaghera:
> I added a couple more comments.
> 
> Could you also add a test of project creation? Considering it was broken, we really need one. See my comment here: [#3542 (comment)](https://github.com/quarkusio/quarkus/pull/3542#issuecomment-521933968) .
> 
> Thanks!

I have added a test project creation test in`CreateProjectMojoIT` for scala.

devtools/common/src/main/java/io/quarkus/generators/SourceType.java
miteshaghera: I have add an another field `name` in `SourceType` to maintain name of the extension.

devtools/maven/src/test/java/io/quarkus/maven/it/DevMojoIT.java
```diff
@@ -189,6 +189,34 @@ public void testThatTheApplicationIsReloadedOnKotlinChange() throws MavenInvocat
                 .atMost(1, TimeUnit.MINUTES).until(() -> getHttpResponse("/app/hello").contains("carambar"));
     }
 
+    @Test
+    public void testThatTheApplicationIsReloadedOnScalaChange() throws MavenInvocationException, IOException {
```
machi1990: By nature, these tests IT take some minutes  to run,  I wonder if this test should be a unit test using `QuarkusDevModeTest` since https://github.com/quarkusio/quarkus/pull/3317 has already landed. 

@gsmet WDYT?

machi1990 (changes requested):

devtools/maven/src/test/java/io/quarkus/maven/it/DevMojoIT.java
geoand: If `QuarkusDevModeTest` can cover this case, it would be ideal to use it. 
And if it does, then replacing the existing tests with something that uses `QuarkusDevModeTest` would be even better :)

machi1990: Sure,  I think tests that check if files have been correctly modified, deleted etc can leverage `QuarkusDevModeTest`. We have an issue for that?

geoand: Not that I know of, but I haven't really checked

gsmet (approved):

gsmet:
Merged, thanks! Very nice addition.

Issue #3400, opened 2019-08-01
Reactive Messaging - bug injecting @Stream in a class with @Observes StartupEvent


famartinrh:
**Describe the bug**
Trying the last quarkus version `0.20.0` I found an issue with `quarkus-smallrye-reactive-messaging-amqp` that actually looks like could affect to `smallrye-reactive-messaging-kafka` and `smallrye-reactive-messaging-mqtt` too.
The issue happens at startup time (I only tried hotspot, nothing of native-image) when you inject a `@Stream` and in the same class you `@Observes StartupEvent` you get an error saying the stream can't be found eventhough having a correct configuration in the application.properties file.

My investigation makes me think the problem is in `io.quarkus.smallrye.reactivemessaging.runtime.SmallRyeReactiveMessagingLifecycle` which observes `StartupEvent` too, so then what could be happening is that `io.smallrye.reactive.messaging.extension.StreamProducer` which depends on `io.smallrye.reactive.messaging.ChannelRegistry` doesn't know of any registered publisher or subscriber because the `MediatorManager` isn't initializated because my StartupEvent listener executes before the one in `SmallRyeReactiveMessagingLifecycle`.

The first solution that comes to my mind is to change the event that `SmallRyeReactiveMessagingLifecycle` listen's to another one but I don't actually know what event could be used. Another solution could be to prioritize somehow the execution of the events in listeners from quarkus extensions

**Expected behavior**
Application successfully starts

**Actual behavior**
This error appears when trying to run the application
```java.lang.IllegalStateException: Unable to find a stream with the name prices, available streams are: []
	at io.smallrye.reactive.messaging.extension.StreamProducer.getPublisher(StreamProducer.java:68)
	at io.smallrye.reactive.messaging.extension.StreamProducer.producePublisher(StreamProducer.java:38)
	at io.smallrye.reactive.messaging.extension.StreamProducer_ProducerMethod_producePublisher_78a82f199fc91074cdfbd65c72662644b3754ad7_Bean.create(StreamProducer_ProducerMethod_producePublisher_78a82f199fc91074cdfbd65c72662644b3754ad7_Bean.zig:110)
	at io.smallrye.reactive.messaging.extension.StreamProducer_ProducerMethod_producePublisher_78a82f199fc91074cdfbd65c72662644b3754ad7_Bean.get(StreamProducer_ProducerMethod_producePublisher_78a82f199fc91074cdfbd65c72662644b3754ad7_Bean.zig:333)
[...]

PR #3402, opened 2019-08-01
Reactive Messaging - initialize MediatorManager earlier giving priority to StartupEvent observer

famarting:
related to #3400 
Because this extension observes StartupEvent which is a common event to be observed by user code I think it's necessary add priority to this observer in order to ensure reactive-messaging extension is fully initialized prior any user code execution


mkouba (approved):

mkouba:
@famartinrh Thanks for the contribution!

PR #3406, opened 2019-08-02
Arc - add built in beans for BM, Event, Instance. Add automated tests.

### LINKED ISSUE #2459: Make it possible to inject built-in beans through BeanManager.getInjectableReference()
This is a follow-up of https://github.com/quarkusio/quarkus/issues/2368. Currently, we don't provide the real beans for built-in beans (`BeanManager`, `Instance`, `Event`, etc.) but resolve them directly during build. This works fine for beans but not for `BeanManager.getInjectableReference()` and `@QuarkusTest`.

manovotn:
Fixes #2459

### DIFF
--- a/independent-projects/arc/runtime/src/main/java/io/quarkus/arc/BuiltInBean.java
+++ b/independent-projects/arc/runtime/src/main/java/io/quarkus/arc/BuiltInBean.java
@@ -1,0 +1,20 @@
+package io.quarkus.arc;
+
+import javax.enterprise.context.spi.CreationalContext;
+
+/**
+ * Common class for all built-in beans.
+ *
+ */
+public abstract class BuiltInBean<T> implements InjectableBean<T> {
+
+    @Override
+    public String getIdentifier() {
+        return "builtin_bean_" + this.getClass().getSimpleName();
+    }
+
+    @Override
+    public T create(CreationalContext<T> creationalContext) {
+        return get(creationalContext);
+    }
+}
--- a/independent-projects/arc/runtime/src/main/java/io/quarkus/arc/EventBean.java
+++ b/independent-projects/arc/runtime/src/main/java/io/quarkus/arc/EventBean.java
@@ -1,0 +1,33 @@
+package io.quarkus.arc;
+
+import java.lang.reflect.Type;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Set;
+import javax.enterprise.context.spi.CreationalContext;
+import javax.enterprise.event.Event;
+import javax.enterprise.inject.spi.InjectionPoint;
+
+public class EventBean extends BuiltInBean<Event<?>> {
+
+    public static final Set<Type> EVENT_TYPES = Collections
+            .unmodifiableSet(new HashSet<>(Arrays.asList(Event.class, Object.class)));
+
+    @Override
+    public Set<Type> getTypes() {
+        return EVENT_TYPES;
+    }
+
+    @Override
+    public Event<?> get(CreationalContext<Event<?>> creationalContext) {
+        // Obtain current IP to get the required type and qualifiers
+        InjectionPoint ip = InjectionPointProvider.get();
+        return new EventImpl<>(ip.getType(), ip.getQualifiers());
+    }
+
+    @Override
+    public Class<?> getBeanClass() {
+        return EventImpl.class;
+    }
+}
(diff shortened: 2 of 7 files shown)


independent-projects/arc/runtime/src/main/java/io/quarkus/arc/BuiltInBean.java
```diff
@@ -0,0 +1,45 @@
...
+
+/**
+ * Common class for all built-in beans.
+ *
+ */
+public abstract class BuiltInBean<T> implements InjectableBean<T> {
+
+    @Override
+    public String getIdentifier() {
+        return sha1(toString());
```
mkouba: I don't think we need such a complicated id for built-in beans. I would use some prefix and bean class name instead. Something like `builtin_BeanManagerBean`...

independent-projects/arc/runtime/src/main/java/io/quarkus/arc/EventBean.java
```diff
@@ -0,0 +1,37 @@
...
+    public Set<Type> getTypes() {
+        return EVENT_TYPES;
+    }
+
+    @Override
+    public Event<?> get(CreationalContext<Event<?>> creationalContext) {
+        // Obtain current IP to get the required type and qualifiers
+        InjectionPoint ip = InjectionPointProvider.get();
+        EventImpl<?> instance = new EventImpl<>(ip.getType(), ip.getQualifiers());
+        CreationalContextImpl.addDependencyToParent((InjectableBean<Event<?>>) ip.getBean(), instance, creationalContext);
```
mkouba: See also my comment for `BeanManagerBean`...

mkouba (changes requested):

independent-projects/arc/runtime/src/main/java/io/quarkus/arc/EventBean.java
```diff
@@ -0,0 +1,35 @@
...
+import java.util.Set;
+import javax.enterprise.context.spi.CreationalContext;
+import javax.enterprise.event.Event;
+import javax.enterprise.inject.spi.InjectionPoint;
+import javax.enterprise.util.TypeLiteral;
+
+public class EventBean extends BuiltInBean<Event<?>> {
+
+    public static final Set<Type> EVENT_TYPES = Collections
+            .unmodifiableSet(new HashSet<>(Arrays.asList(Event.class, new TypeLiteral<Event<Object>>() {
```
mkouba: Why the `TypeLiteral` when we're only checking the raw type?

independent-projects/arc/runtime/src/main/java/io/quarkus/arc/InstanceBean.java
```diff
@@ -0,0 +1,36 @@
...
+    @Override
+    public Class<?> getBeanClass() {
+        return InstanceImpl.class;
+    }
+
+    @Override
+    public Instance<?> get(CreationalContext<Instance<?>> creationalContext) {
+        // Obtain current IP to get the required type and qualifiers
+        InjectionPoint ip = InjectionPointProvider.get();
+        return new InstanceImpl<Instance<?>>((InjectableBean<?>) ip.getBean(), ip.getType(),
```
mkouba: I think we do need to add this instance to the parent CC in this case because all `@Dependent` bean instances produced by `Instance` belong to the CC of the bean that injects the `Instance` instance.

independent-projects/arc/runtime/src/main/java/io/quarkus/arc/EventBean.java
manovotn: Not sure, TBF I went after what Weld had and this was there as well.
It was only later when I realized that's not really used, want me to remove that?

independent-projects/arc/runtime/src/main/java/io/quarkus/arc/InstanceBean.java
manovotn: Oops, you are right, I didn't mean to remove that one.
BTW is it safe to assume that `InjectionPoint.getBean()` always returns instance of `InjectableBean<Instance<?>>`? Because we need to cast it there anyway...

mkouba: It should be safe...

manovotn:
@mkouba I addressed all the comments

mkouba (approved):

PR #3421, opened 2019-08-05
fix: possible infinite loop on Oauth2 Identity Manager

### LINKED ISSUE #3419: OAuth2IdentityManager.verify(String id, Credential credential) - infinite loop
OAuth2IdentityManager.verify(String id, Credential credential) has infinite loop / recursive call.

link to code: https://github.com/quarkusio/quarkus/blob/master/extensions/elytron-security-oauth2/runtime/src/main/java/io/quarkus/elytron/security/oauth2/runtime/auth/OAuth2IdentityManager.java#L31

@loicmathieu should be current `return verify(null, credential);` changed to `return verify(credential);` ?

loicmathieu:
Fixes #3419

### DIFF
--- a/extensions/elytron-security-oauth2/runtime/src/main/java/io/quarkus/elytron/security/oauth2/runtime/auth/OAuth2IdentityManager.java
+++ b/extensions/elytron-security-oauth2/runtime/src/main/java/io/quarkus/elytron/security/oauth2/runtime/auth/OAuth2IdentityManager.java
@@ -29,7 +29,7 @@ public class OAuth2IdentityManager implements IdentityManager {
 
     @Override
     public Account verify(String id, Credential credential) {
-        return verify(null, credential);
+        return null;
     }
 
     @Override


machi1990 (approved):

sberyozkin (commented):
Should this one throw UnsupportedOperationException ?

loicmathieu:
@sberyozkin it could be but none of the other IndentityManager on the other security extension throws UnsupportedOperationException for unused methods, they return `null` so I can change my code to return `null` if you prefere as a safeguard to align with JwtIdentityManager and ElytronIdentityManager.

sberyozkin:
Hi @loicmathieu - IMHO returning `null` would be better, looks like this method is never called, so even the current delegation would do - but I'm not sure this method's contact will be satisfied this way if it is ever called...

loicmathieu:
@sberyozkin done

n1hility (approved):

n1hility:
Thanks 🍺!

Issue #1594, opened 2019-03-20
Propose a parent pom for external extensions


danielpetisme:
Currently, the extensions are embedded into the main Quarkus repository thus use the internal `build-parent` pom.xml.
To permits 3rd parties to develop their own extensions, it would be great to propose a public `quarkus-extension-parent` pom.

https://quarkusio.zulipchat.com/#narrow/stream/187030-users/topic/Extensions/near/161188290

ppalaga:
I proposed a basic parent for third party extensions in https://github.com/quarkusio/quarkus/pull/3427 the commit: https://github.com/ppalaga/quarkus/commit/512022706309a0de6925c9a9ef5fd10a067e8ae9

However, when we start refactoring the hierarchy, @aloubyansky would like to solve more than this single problem https://github.com/quarkusio/quarkus/pull/3427#issuecomment-518580389

> This is basically the same approach I was investigating. I was just thinking it could be organized slightly differently, e.g.
> 
> ```
> build-parent/
> |- application/
> |  `- pom.xml (parent for apps built on Quarkus core, flattened)
> |- extension/
> |  `- pom.xml (parent for extensions built on Quarkus core, probably has application/pom.xml as its parent, flattened)
> `- internal/
>    `- pom.xml (Quarkus core build parent pom, has extension/pom.xml as its parent).
> ```
> 
> However, before actually applying this kind of change, it'd be a good idea to announce/discuss it on the dev list.

danielpetisme:
Fixed by https://github.com/quarkusio/quarkus-platform fix

PR #3407, opened 2019-08-02
 CreateExtensionMojo should optionally add the newly created modules to BOMs

ppalaga:
A followup of https://github.com/quarkusio/quarkus/pull/3390

PR #3427, opened 2019-08-05
 Fix #3321 create-extension Maven mojo should allow creating from scratch

### LINKED ISSUE #3321: create-extension Maven mojo should allow creating from scratch
A followup of #2113 .

create-extension Maven mojo added via #3191 allows creating new extension modules inside an existing source tree. It should be enhanced to be able to create new extension projects from scratch.

cc @aloubyansky @Dufgui

ppalaga:
The first two commits are covered by https://github.com/quarkusio/quarkus/pull/3407 so they do not need to get reviewed here.

### DIFF
--- a/core/builder/pom.xml
+++ b/core/builder/pom.xml
@@ -11,7 +11,7 @@
         <groupId>io.quarkus</groupId>
         <artifactId>quarkus-build-parent</artifactId>
         <version>999-SNAPSHOT</version>
-        <relativePath>../../build-parent/pom.xml</relativePath>
+        <relativePath>../../parents/build-parent/pom.xml</relativePath>
     </parent>
 
     <properties>
--- a/core/creator/pom.xml
+++ b/core/creator/pom.xml
@@ -6,7 +6,7 @@
         <artifactId>quarkus-build-parent</artifactId>
         <groupId>io.quarkus</groupId>
         <version>999-SNAPSHOT</version>
-        <relativePath>../../build-parent/pom.xml</relativePath>
+        <relativePath>../../parents/build-parent/pom.xml</relativePath>
     </parent>
     <modelVersion>4.0.0</modelVersion>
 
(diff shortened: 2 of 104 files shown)


aloubyansky:
This is basically the same approach I was investigating. I was just thinking it could be organized slightly differently, e.g.
````
build-parent/
|- application/
|  `- pom.xml (parent for apps built on Quarkus core, flattened)
|- extension/
|  `- pom.xml (parent for extensions built on Quarkus core, probably has application/pom.xml as its parent, flattened)
`- internal/
   `- pom.xml (Quarkus core build parent pom, has extension/pom.xml as its parent).
````
However, before actually applying this kind of change, it'd be a good idea to announce/discuss it on the dev list.

ppalaga:
Yes, I think your proposal could work for me too.

+1 for opening the topic on quarkus-dev. Could you please do that because I think I am not acquainted with all problems your tri-section tries to solve?

aloubyansky:
Here it is https://groups.google.com/forum/#!topic/quarkus-dev/UoDBpG4PCJQ

ppalaga:
Blocked by https://github.com/quarkusio/quarkus/issues/1594 

@aloubyansky may I expect that you are going to fix https://github.com/quarkusio/quarkus/issues/1594 ?

aloubyansky:
Yes, I can look into this next week.

gsmet:
@aloubyansky do you still plan to work on this one?

aloubyansky:
Yes, I am working on the split. Not a trivial task. In addition, getting destructed by other issues. We'll most probably need to discuss the details when I am more or less satisfied with my prototype.

gsmet:
@aloubyansky OK, thanks for the feedback!

cescoffier:
Any update on this one? Would be great to have it!

ppalaga:
@cescoffier fully forgot it, sorry. I hope I can get back to it when I have a couple of free cycles.

gsmet:
@aloubyansky I think we should define a proper plan to have things in line with your other parent pom PR. Otherwise it will conflict.

I'm putting this one on ice until we have a clear plan.

ppalaga:
Too old to be useful for anything

Issue #2489, opened 2019-05-17
Add validation parameters to ConfigItem


dmlloyd:
Add something like the following to `ConfigItem`:

```java
    String NOT_SPECIFIED = "<<not specified>>";

    String matching() default NOT_SPECIFIED;

    String minValue() default NOT_SPECIFIED;
    String maxValue() default NOT_SPECIFIED;
    boolean minInclusive() default true;
    boolean maxInclusive() default false;
````

The behavior of the `minValue` and `maxValue` parameters should be that for each parameter, if it is given and the value is `Comparable`, then the value should be `convert`ed using the defined converters and compared to the given value, registering an error if it is not within range (which will lead to Quarkus not starting up).  Specifying `minValue` and `maxValue` for values which are not either `Comparable` or `OptionalInt`/`OptionalLong`/`OptionalDouble` would be an error.

The `minInclusive` and `maxInclusive` parameters would define whether the minimum and maximum values are inclusive.

The behavior of `matching` is that the value is compiled to a regular expression `Pattern` which is matched against the given value (which must have a meaningful `toString` method implementation).  Failing to match the pattern would result in an error being registered.

Having both `minValue`/`maxValue` and `matching` would be allowed though probably only useful in a very small set of cases.

**Comment by dmlloyd** [comment]
Additional ideas:

```java
    boolean allowEmpty() default true;
    boolean required() default false;
```

The distinctions between these values and their interactions with `defaultValue` would not be clear until the MP config issues and discussions are resolved:

* eclipse/microprofile-config#407
* smallrye/smallrye-config#83

dmlloyd:
Another idea: print a deprecation warning if the property is populated but is marked `@Deprecated`.  If we wanted to get very fancy we could print out the summary string from the `@deprecated` doc tag.

svenhaag:
Hi, what's the state of this? I would love to see these validation annotations in SmallRye-Config (the MP-Config Spec). Cheers

dmlloyd:
I'm piloting a rough draft of this in https://github.com/smallrye/smallrye-config/pull/310; @radcortez is trying a different approach (reusing Bean Validation annotations) in https://github.com/smallrye/smallrye-config/pull/317.

svenhaag:
Both issues are simply awesome, but I would prefer Bean Validation too over custom annotations. Cheers.

geoand:
[...]

PR #3428, opened 2019-08-05
Generate Asciidoc configuration file for each extension

machi1990:
This PR proposes generation of config properties for each config root. 
- the generated `.adoc`  file resembles in form this [mongo-db configuration](https://quarkus.io/guides/mongo-guide#connection-properties)

The docs are generated to `docs/src/main/asciidoc/generated/`. 
I have modified the datasource documentation and sending emails guide so that we can see the overall output. 

To see more examples of generate docs just run `mvn clean install -DskipTests -DskipITs`.

or clone this repository https://github.com/machi1990/quarkus-generated-extension-config-docs

Opening as draft to gather early feedback (thanks @gsmet for the suggestion)

## TODO 
# 1st Iteration
Several points needs to be addressed
- [x] ~~Is the rendering okay? What needs improving? We can also determine if a property is runtime configurable, how do we present this information?~~ @emmanuelbernard come with a suggestion of using a descriptor list https://asciidoctor.org/docs/user-manual/#description-list. 
- [x] Output location of the generated docs. In `{root}/docs/src/main/asciidoc/generated/` 
- [x] ~~What should be the final file name, and how will it be defined. Annotation based? In the meantime, the implementation reads package segments e.g `io.quarkus.agroal.runtime` will result in a file name `quarkus-agroal-runtime.adoc` :-)~~ . Final name is in `kebab-case` and is generated from package segments prefixed with `quarkus` e.g `quarkus-agroal.adoc`, `quarkus-hibernate-orm.adoc`, `quarkus-hibernate-search-elasticsearch.adoc`. There are exceptions like core configs and in this case we have `quarkus-core-log-config`, `quarkus-core-application-config.adoc` etc
- [x] ~~Fix an issue related to a configuration group class defined in another module and then used in configuration root of another module. e.g `ServerSslConfig.java` which is defined in the core module and used in `HttpConfig.java` in undertow module.~~ Descoped by moving the config group in related module. 
- [x] Add `Default Value`  and `Visibility` infos
- [x] ~~Indicate if a config is optional or not.~~ This information should be part of the javadoc as in https://github.com/quarkusio/quarkus/blob/ef684647754391c75d282ee499984c27890eba92/extensions/keycloak/deployment/src/main/java/io/quarkus/keycloak/KeycloakConfig.java#L71-L74. 
- [x] Fix generated config ordering issue 
- [x] Generate output in a summary table format with each row linking to a long descriptive format (the current state of the PR) 
- [x] Add extension JavaDoc site link to`type` value or a link to Github source code e.g for agroal, values in the `type` columns will point to this link https://jar-download.com/artifacts/io.agroal/agroal-api/1.5/documentation while the mailer extension we'll point to GH link https://github.com/quarkusio/quarkus/blob/master/extensions/mailer/runtime/src/main/java/io/quarkus/mailer/impl/MailConfig.java

# 2nd Iteration
We have an issue tracking this iteration here - https://github.com/quarkusio/quarkus/issues/3803 

/cc @emmanuelbernard @dmlloyd @gsmet @stuartwdouglas


emmanuelbernard:
Great work! Here are a few ideas to improve and integrate. 

I think “default value: <default>” in each entry will read better than the (<default>) you currently have. 

What I love about the generation of one file per extension is that we can then add an include in the asciidoc configuration for each. We might need a convention for the generated file name. 

I don’t know if the type is that important (genuine question). It does take a lot of space. 

It would be useful to express if a property is optional or not. 

I recommend also we express if the property is build time only. Readable also at runtime but as read only or if that’s overridable at runtime. It’s part of the config group annotation. I think the second column is good for that.

emmanuelbernard:
About how to present the info on the visibility of the property. I recommend the second column with the following text:
- visible at build time only
- visible at build and runtime time, read only at runtime 
- overridable at runtime

emmanuelbernard:
Generated in `docs/src/main/asciidoc/generated/`

emmanuelbernard:
Ah so you have different files for deployment and runtime for each extension. I wonder if we should try and merge them.
To me the most stable name would be either the quarkus property name space. Or the extension GAV. Probably the extension GAV is the safest.

emmanuelbernard:
@stuartwdouglas what do we do about `ServerSslConfig` being shared between core and undertow? It feels like `ServerSslConfig` should not generate doc as it's not a `@ConfigRoot`?

emmanuelbernard:
CC @dmlloyd on ^

emmanuelbernard:
@machi1990 Here is a challenge for you. I improved the ORM config documentation in the Javadoc itself. Let me know if that still works for you and if you think the asciidoclet model would work. https://github.com/quarkusio/quarkus/pull/3430

emmanuelbernard:
@machi1990 what does `Unresolved directive in hibernate-orm.adoc - include::duration-format-note.adoc[]` in hibernate-orm.adoc means? Where does that one comes from?

emmanuelbernard:
I did the following replacements using backticks to render as code in Asciidoc
```
        ret = JAVA_DOC_CODE_PATTERN.matcher(ret).replaceAll("`$1`");
//and
            ret = ret.replace(matcher.group(0), "`" + configify(matcher.group(1)) + "`");
```

machi1990:
> @machi1990 what does `Unresolved directive in hibernate-orm.adoc - include::duration-format-note.adoc[]` in hibernate-orm.adoc means? Where does that one comes from?

@emmanuelbernard this is a generic note that is appended to give precision on the custom `DurationConverter` that we use. It is only available if one of the config property is of type  `Duration`.

machi1990:
> @machi1990 Here is a challenge for you. I improved the ORM config documentation in the Javadoc itself. Let me know if that still works for you and if you think the asciidoclet model would work. #3430

It should be handled by the current implementation. If not, then it is definitely an are for improvement that I'll be looking to work on.

machi1990:
> @stuartwdouglas what do we do about `ServerSslConfig` being shared between core and undertow? It feels like `ServerSslConfig` should not generate doc as it's not a `@ConfigRoot`?

Yes `ServerSslConfig` is just a config group so it does not generate a doc on it is own but we should generate a doc of its usage in any config root. It appears that it is used in `HttpConfig` and my current implementation does not handle that at the moment. So I am wondering if it only used in `HttpConfig` maybe we can move it in the same package.

machi1990:
> Ah so you have different files for deployment and runtime for each extension. I wonder if we should try and merge them.
> To me the most stable name would be either the quarkus property name space. Or the extension GAV. Probably the extension GAV is the safest.

+1. My idea was to read runtime props and serialise them somewhere and when processing the corresponding extension `deployment` module, deserialise these props and create one file par extension. Maybe there is a better approach? @dmlloyd @stuartwdouglas

emmanuelbernard:
> > @machi1990 Here is a challenge for you. I improved the ORM config documentation in the Javadoc itself. Let me know if that still works for you and if you think the asciidoclet model would work. #3430
> 
> It should be handled by the current implementation. If not, then it is definitely an are for improvement that I'll be looking to work on.

Well I tested a quite a few constructs are not supported. Rebase my branch off your work and you will see. But see also my email on quarkus-dev about using Asciidoc to write the Javadoc.

machi1990:
> Generated in `{root}/docs/src/main/asciidoc/generated/` 

+1 . I will add this a bit later today.

machi1990:
> About how to present the info on the visibility of the property. I recommend the second column with the following text:
> 
> * visible at build time only
> * visible at build and runtime time, read only at runtime
> * overridable at runtime

Maybe we can drop `Java Type` column with a `Visibility - default value` column. ?

machi1990:
> Great work! Here are a few ideas to improve and integrate.
> 
> I think “default value: ” in each entry will read better than the () you currently have.
> 
> What I love about the generation of one file per extension is that we can then add an include in the asciidoc configuration for each. We might need a convention for the generated file name.
> 
> I don’t know if the type is that important (genuine question). It does take a lot of space.
> 
> It would be useful to express if a property is optional or not.
> 
> I recommend also we express if the property is build time only. Readable also at runtime but as read only or if that’s overridable at runtime. It’s part of the config group annotation. I think the second column is good for that.

@emmanuelbernard thanks for the early feedback. I am in phase with most of them. I'll also confirm a question regarding #3430 tonight once I have run some tests. Either way, the rendering part can be improved with a bit more parsing logics. I used the same parser as the one used to generate an example `application.properties` file.

emmanuelbernard:
I've been thinking about it and I don't like the table format. If only because it makes mobile miserable and it creates a separation between your guide and that inserted piece of info.
I think it's better if we use the descriptor list in asciidoc
https://asciidoctor.org/docs/user-manual/#description-list

```
`quarkus.hibernate-orm.database.generation`::
default value: `none` ; 
type: `java.lang.String`
+
Select whether the database schema is generated or not. Options are `none`, `create`, `drop-and-create`, `drop`, `update`. The default is `none`. `drop-and-create` is awesome in development mode.
+
Same as JPA’s `javax.persistence.schema-generation.database.action`.

`quarkus.hibernate-orm.database.generation.halt-on-error`::
default value: `false` ;
type: `boolean`
+
Whether we should stop schema application at the first error or continue.
+
[NOTE]
--
Foo is


Bar
--
+
Go
```

The one concern I have is that to do proper indenting, one need to up the + on empty lines to link the paragraphs. But that should not happen on nested blocks (like the note above).

Finally, i wonder if we should have headers for each sub section.

```
//general section not in sub section
`quarkus.hibernate-orm.batch-fetch-size`:: [...]

`quarkus.hibernate-orm.dialect`:: [...]

== Database

`quarkus.hibernate-orm.database.generation`:: [...]

`quarkus.hibernate-orm.database.generation.halt-on-error`:: [...]

== Cache

[...]
```


Also let's shortcut java.lang in the types. `String` instead of `java.lang.String`

emmanuelbernard:
Another thing, since we will group properties per subgroup (like orm general, orm database, orm cache), I think you should generate the following files:

- all-properties.qdoc
- hibernate-orm-all-properties.adoc
- hibernate-orm-general-properties.adoc (top level ones)
- hibernate-orm-cache-properties.adoc
- hibernate-orm-database-properties.adoc

That way a guide can split properties per section or give n app properties section as best fit

machi1990:
> I've been thinking about it and I don't like the table format. If only because it makes mobile miserable and it creates a separation between your guide and that inserted piece of info.
> I think it's better if we use the descriptor list in asciidoc
> https://asciidoctor.org/docs/user-manual/#description-list
> 
> ```
> `quarkus.hibernate-orm.database.generation`::
> default value: `none` ; 
> type: `java.lang.String`
> +
> Select whether the database schema is generated or not. Options are `none`, `create`, `drop-and-create`, `drop`, `update`. The default is `none`. `drop-and-create` is awesome in development mode.
> +
> Same as JPA’s `javax.persistence.schema-generation.database.action`.
> 
> `quarkus.hibernate-orm.database.generation.halt-on-error`::
> default value: `false` ;
> type: `boolean`
> +
> Whether we should stop schema application at the first error or continue.
> +
> [NOTE]
> --
> Foo is
> 
> 
> Bar
> --
> +
> Go
> ```
> 
> The one concern I have is that to do proper indenting, one need to up the + on empty lines to link the paragraphs. But that should not happen on nested blocks (like the note above).
> 
> Finally, i wonder if we should have headers for each sub section.
> 
> ```
> //general section not in sub section
> `quarkus.hibernate-orm.batch-fetch-size`:: [...]
> 
> `quarkus.hibernate-orm.dialect`:: [...]
> 
> == Database
> 
> `quarkus.hibernate-orm.database.generation`:: [...]
> 
> `quarkus.hibernate-orm.database.generation.halt-on-error`:: [...]
> 
> == Cache
> 
> [...]
> ```
> 
> Also let's shortcut java.lang in the types. `String` instead of `java.lang.String`

Nice. It was another option, I think we have a guide with a similar format - https://quarkus.io/guides/mongo-guide#connection-properties which was also an option.

dmlloyd:
> @stuartwdouglas what do we do about `ServerSslConfig` being shared between core and undertow? It feels like `ServerSslConfig` should not generate doc as it's not a `@ConfigRoot`?

It would be nice if we only generate one copy of documentation for a given class, especially in cases like this where there is reuse, because it makes it clearer that there is consistency (i.e. I know I can configure B just like I configured A because it's the same).  Doing so in an intuitive way that also accounts for config class inheritance is an exercise for someone not currently on PTO though. :)

dmlloyd:
> > Ah so you have different files for deployment and runtime for each extension. I wonder if we should try and merge them.
> > To me the most stable name would be either the quarkus property name space. Or the extension GAV. Probably the extension GAV is the safest.
> 
> +1. My idea was to read runtime props and serialise them somewhere and when processing the corresponding extension `deployment` module, deserialise these props and create one file par extension. Maybe there is a better approach? @dmlloyd @stuartwdouglas

The distinction between runtime and deployment is not important to users.  What matters far more is the distinction between run time properties (which will always be in the runtime artifact) and build time properties (which may be in the runtime or deployment artifact, and might be moved from one to the other over time).

core/processor/src/main/java/io/quarkus/annotation/processor/Constants.java
```diff
@@ -0,0 +1,66 @@
...
+import java.util.OptionalLong;
+import java.util.Set;
+import java.util.regex.Pattern;
+
+public class Constants {
+    public static final String PARENT = "<<parent>>";
+    public static final String NO_DEFAULT = "<<no default>>";
+    public static final String HYPHENATED_ELEMENT_NAME = "<<hyphenated element name>>";
+
+    public static final Pattern JAVA_DOC_CODE_PATTERN = Pattern.compile("\\{@code (.*?)\\}");
```
dmlloyd: Here's what I sent to @gsmet before I left on PTO:

> The way it works is, the Quarkus extension
`quarkus-extension-processor` runs on each extension module which
contains build steps (obviously deployment modules only) or
configuration (both deployment and runtime).  It generates a file
`META-INF/quarkus-javadoc.properties` that contains key-value pairs
where the key is `package.name.ClassName.methodName` and the value is
the complete JavaDoc text including tags.
>
> The configuration classes can be found by tracing the path from
classes marked with `@ConfigRoot` annotations.  We do not presently
Jandex our extensions.  To find the config classes in use, you have to
iterate the classes named in `META-INF/quarkus-build-steps.list` and
then iterate all the constructor parameters, non-final fields, and
parameters of any method marked with the `@BuildStep` annotation and
locate any class named there that has `@ConfigRoot` on it.
>
> This is obviously not ideal and I'd like to simplify it.  I want to
move to putting a Jandex index on every extension deployment and
runtime module.  I also think we should be putting the config member
javadoc in the Jandex index just so we can have one place to look;
however this is a bit tricky to do in a way which isn't overly complex
or which doesn't include unwanted JavaDoc.  Then I'd like to locate
all configuration classes simply by examining the Jandex index, rather
than traversing outwards from the build steps.
>
> The correct approach for generating documentation for these things is
to start by parsing the JavaDoc, in two stages.  In the first stage
you expand inline doc tags and isolate the block tags into separate
units - this gives us the chance to introduce our own doc tags, if we
want.  In the second stage you parse them as HTML using something like
jsoup.  Once the JavaDoc is parsed into some kind of structure, you
can then render it in the desired format (plain text, adoc, md, HTML5,
etc.).
>
> Stuart bypassed a lot of this by just slamming it with regex
substitutions.  That only works up to a point though.
>
> Hope this helps.

The output is always going to be janky if we're going to be using regex to "parse" the docs.  If we want clean formatting, the parser needs semantic awareness of what it has parsed.  Would be nice to do this right.

gsmet:
> It would be nice if we only generate one copy of documentation for a given class, especially in cases like this where there is reuse, because it makes it clearer that there is consistency (i.e. I know I can configure B just like I configured A because it's the same).

I was thinking about doing that at first but the issue with it is that we have an artificial separation for build time config and runtime config.

### File name

I would like to have one file per extension named after the extension artifact id if possible.

### Build time/runtime config

We really need this info in the doc.

### Parsing the javadoc

I found this project https://github.com/javaparser/javaparser and especially the work presented here https://github.com/javaparser/javaparser/pull/699 .

It does a lot more that what we want as it parses the whole Java files but I wonder if we could reuse just the Javadoc part somehow (might now be easy though so maybe it's not a good idea).

If we have to reimplement the parser, taking inspiration from it and maybe use https://github.com/sirthias/parboiled instead of antlr might be a possibility (we use parboiled in Hibernate OGM and it was pretty handy for simple stuff).

I agree with @dmlloyd that using Asciidoc comments, even if easier for us, is not very IDE friendly.

### Making progress

@machi1990 I think it could be handy if you created a GitHub project containing the output so we can discuss the progress with the Asciidoc easily visible.

It shouldn't be too hard to write a script to copy all the files in a repo and commit them.

machi1990:
Hi @gsmet, welcome back :-)

> ### File name
> I would like to have one file per extension named after the extension artifact id if possible.

I'll look at this effectively over the weekend. I am yet to figure out a better way to achieve this. 

> ### Build time/runtime config
> We really need this info in the doc.

One of @emmanuelbernard suggestions was to add a visibility note. You can se it from the generated output file. 
 
> ### Parsing the javadoc
> I found this project https://github.com/javaparser/javaparser and especially the work presented here [javaparser/javaparser#699](https://github.com/javaparser/javaparser/pull/699) .
> 
> It does a lot more that what we want as it parses the whole Java files but I wonder if we could reuse just the Javadoc part somehow (might now be easy though so maybe it's not a good idea).
> 
> If we have to reimplement the parser, taking inspiration from it and maybe use https://github.com/sirthias/parboiled instead of antlr might be a possibility (we use parboiled in Hibernate OGM and it was pretty handy for simple stuff).

Great, thanks for the many options. I'll look into them after I have made progress on the other points. I have used `jsoup` before and I think it could also be a viable option. 

> ### Making progress
> @machi1990 I think it could be handy if you created a GitHub project containing the output so we can discuss the progress with the Asciidoc easily visible.
> 
> It shouldn't be too hard to write a script to copy all the files in a repo and commit them.

Sure thing, I like the idea. I have created a repo here https://github.com/machi1990/quarkus-generated-extension-config-docs. I'll be updating the output as I progress on this.

core/processor/src/main/java/io/quarkus/annotation/processor/Constants.java
machi1990: So this have moved a bit since the last implementation. I am now using a combination of `jsoup` and `javaparser` libs. WDYT?

machi1990:
@gsmet @emmanuelbernard @dmlloyd I have created this repository https://github.com/machi1990/quarkus-generated-extension-config-docs. so that we can track progress on the output.  In doing so, I thought maybe we can git-ignore the generated files. WDYT of the output so far?

gsmet:
@machi1990 I took a look at the output. It's quite good already.

There's still one big issue though: when you have config maps, you end up with the config being all random. See here for instance: https://github.com/machi1990/quarkus-generated-extension-config-docs/blob/master/src/main/asciidoc/generated/quarkus-agroal.adoc .

I don't know if it's possible to generate the documentation in the order of the source code (not sure how easy it would be when merging runtime and build time config tbh) but I think we really need to improve on that.

gsmet:
As for the output, I would try something like:
```
`quarkus.datasource.idle-removal-interval`:: The interval at which we try to remove idle connections. +
Type: `java.time.Duration` _see note below_ +
Defaults to: `5M` +
This configuration property is overridable at runtime.
```
(I did a lot of small adjustments, they are all on purpose)

And see if it's too big or more readable. Note the `+` at the end of the sentence to keep the indentation (at least, I recall it worked like this at some point :)).

I would also only say if a property is overridable at runtime or not. I don't think it's useful for the user to know that the configuration is used at build time only and then not visible at runtime or build time only but visible at runtime. These are really implementation details.

FroMage:
First: this is a really great PR.

Ideally, I'd want the generated config list to look like this:

Screenshot from 2019-08-19 18-01-59

In other words: like javadoc. This may require html/css, but it's worth it because it makes the docs so much more readable.

FroMage:
I'd make the runtime/build time info be a 16x16 color icon: saves a lot of space and easy to visualise: red for build, green for runtime (whatever color, doesn't matter).

I'd make the type (String, Duration) use non-qualified type, but link to its javadoc.

gsmet:
Emmanuel already decided against the table format as it’s not mobile friendly. I tend to agree. 

I would like to keep it Asciidoc only as it will be better to generate a PDF if we need one.

gastaldi:
@machi1990 not sure if it's too late for the party, but you can also parse the Java source files and extract the Javadoc directly using Roaster:  https://github.com/forge/roaster#javadoc-creation-and-parsing

machi1990:
> There's still one big issue though: when you have config maps, you end up with the config being all random. See here for instance: https://github.com/machi1990/quarkus-generated-extension-config-docs/blob/master/src/main/asciidoc/generated/quarkus-agroal.adoc . 
> I don't know if it's possible to generate the documentation in the order of the source code (not sure how easy it would be when merging runtime and build time config tbh) but I think we really need to improve on that.

@gsmet good eye. Current I have not tried to order the config property par produced output.  I am thinking of defining a strategy that will order generated the properties in alphabetical order of config key. If it is a map, it will be put at the end of the ordered list. 

For a start this could be for a ConfigRoot and then I'll try to preserve the order when merging build and runtime configs.

machi1990:
> As for the output, I would try something like:
> 
> ```
> `quarkus.datasource.idle-removal-interval`:: The interval at which we try to remove idle connections. +
> Type: `java.time.Duration` _see note below_ +
> Defaults to: `5M` +
> This configuration property is overridable at runtime.
> ```
> 
> (I did a lot of small adjustments, they are all on purpose)
> 
> And see if it's too big or more readable. Note the `+` at the end of the sentence to keep the indentation (at least, I recall it worked like this at some point :)).

Thanks for the example. Makes the comment way clearer. I'll integrate the proposed adjustments. 
 
> I would also only say if a property is overridable at runtime or not. I don't think it's useful for the user to know that the configuration is used at build time only and then not visible at runtime or build time only but visible at runtime. These are really implementation details.

Okay, the easier it is to users  to understand for the better. So let's sum up things:
- BUILD_TIME or BUILD_AND_RUN_TIME_FIXED ==> `This configuration property is not overridable at runtime.`
- RUNTIME ==> `This configuration property is overridable at runtime.`

machi1990:
> First: this is a really great PR.

> Ideally, I'd want the generated config list to look like this:
> 
> Screenshot from 2019-08-19 18-01-59
> 
> In other words: like javadoc. This may require html/css, but it's worth it because it makes the docs so much more readable.

@FroMage thanks. About the table format, it was an early option that we finally decided against because of responsiveness in mobile devices.

gsmet:
I checked the output: you really need a `+` at the end of each new line so that Asciidoc associate the other sentences to the item.

All the sentences in a given block has to end with `+` - except the last one before the following block.

machi1990:
> I checked the output: you really need a `+` at the end of each new line so that Asciidoc associate the other sentences to the item.
> 
> All the sentences in a given block has to end with `+` - except the last one before the following block.

I see what you mean. I initially read the comment on my mobile and misunderstood that a `+` should be only be applied to `Type` and `Default value`.

gsmet:
You're too fast :). I added a few additional comments in my comment above :).

gsmet:
And maybe you could add a title to the format notes. That would help if we drop the `Duration` text from the `see` part.

Something like:
```
[NOTE]
.About the Duration format
====
Your note...
====
```

machi1990:
> I checked the output: you really need a `+` at the end of each new line so that Asciidoc associate the other sentences to the item.
> 
> All the sentences in a given block has to end with `+` - except the last one before the following block.
> 
> The idea is that all the lines of a block should be indented a bit.
> 
> Try to check the following asciidoc for instance:
> 
> ```
> `quarkus.datasource.driver`:: The datasource driver class name +
> +
> Type: `java.lang.String` +
> The configuration is not overridable at runtime. 
> ```
> 
> I would drop the `.` and the end of `Defaults to: 5` and use `- _see note below_` with just a dash, no uppercase and nothing about `Duration`. It would be a bit lighter and I think it's clear enough. We will maybe have 2-3 notes max at the bottom and I'm pretty sure it will be clear enough.

@gsmet thanks. New commit pushed.

FroMage:
> Emmanuel already decided against the table format as it’s not mobile friendly. I tend to agree.

Sorry but @emmanuelbernard is wrong on this one: we're writing docs for devs. We can't advocate for a bad layout for the 90+% of devs who code on their computers just to make it look mobile-friendly for the 10% of non-coder visitors who got lost on the site.

This is what a non-table doc looks like ATM:

Screenshot from 2019-08-20 11-16-10

I don't mean to criticise any single person for this particular example, I'm guilty of using the same layout for docs I wrote for Quarkus. But it's just horrible: it's the opposite of compact and it's so ad-hoc that there's no structure the eye can use to determine quickly where each config starts, ends, where the default value is documented, where the type is located… It's the opposite of user-friendly. If we have to choose to make it look good for desktop and mobile (and I don't think that's true, it's very possible to write table-like HTML that renders well in mobile) we absolutely _have_ to favour desktop usability.

> I would like to keep it Asciidoc only as it will be better to generate a PDF if we need one.

I don't mind writing it in Asciidoc as long as the output looks as good and structured and compact and predictable.

gsmet:
Well, you **think** that Emmanuel is wrong. I personally agree with him.

We are in the process of improving the Asciidoc output and personally, I don't find tables more readable when there are a lot of items and information. It's harder to spot what you want.

So, yeah, it's a matter of taste, not absolute wisdom.

FroMage:
In fact, I gave some specific reasons for why the current haphazard layout is bad and why a table (or any layout that has the same advantages) are better:

- compact
- standard (every config key doc has the same layout)
- predictable (due to the previous)
- easier to find where keys start and end
- easier to distinguish what is a config doc and what is the rest of the docs
- coherent with decades of developer docs for API (devs know how to scan javadoc tables super fast)
- basing information on columns (and/or standard icons for trivial on/off switches) makes looking up keys, types and defaults super fast

This is an opinion substantiated by facts, not taste at all. Any other structure that can provide the same advantages will work for me. The current layout has no advantage besides "not a table" and being mobile-friendly.

gsmet:
Well, as mentioned, I find the current layout more readable than a table especially when you want to find a specific item. They are in bold and easily identifiable. And they won't be cut if too long, we have some long ones in quite a lot of extensions.

When having 20 configuration knobs, I don't find a table particularly easier to read, more the opposite.

A table is also often an issue with a PDF output where space is scarce, especially with the long keys we have in certain extensions.

I know for a fact we went from a table to this format in a few extensions documentation due to readability issues.

I don't have any issue with you stating an opinion. Just that it's what it is, your opinion, not the absolute truth.

FroMage:
> They are in bold and easily identifiable. 

Obviously we can make the key bold in a table.

> And they won't be cut if too long, we have some long ones in quite a lot of extensions

This should be solved by having an on-click toggle. That's much better than the two-table layout javadoc uses.

> When having 20 configuration knobs, I don't find a table particularly easier to read, more the opposite.

OK, but then why?

> I don't have any issue with you stating an opinion. Just that it's what it is, your opinion, not the absolute truth.

It's not an opinion, it's an observation justified with facts. I do see that you're taking this personally but I don't understand why. Obviously you don't like tables, but I'd much prefer you give reasons why you find them less usable than non-structured layout. 

I don't care _at all_ about tables. If you find me a layout that has as many advantages as the one I proposed, or different/better advantages, I'll take it for sure. Right now, the current state has so many disadvantages that it's impossible to justify. Or it has advantages that haven't been articulated here yet, but then, given that they're not obvious, I doubt that they're really so good.

FroMage:
> So, yeah, it's a matter of taste, not absolute wisdom.

Also please, stop with personal attacks like this: I am not giving any opinion because I have the delusion that I'm "wise". I've found the current config docs layout horrible in terms of UX for months and given a list of reasons why they are so. Please debate the virtue of non-structured layout in terms of its merit compared to structured layout. Not in terms of how wise you imagine I think myself.

gsmet:
> Obviously we can make the key bold in a table.

Obviously, that's not what I was saying. I meant that the key are easily identifiable in this layout because they are in bold and not the rest. In a table, it's far less easy to scan for information (see below).

This type of Asciidoc layout has been created exactly for this purpose.

The fact that we have very long keys will make the table hardly readable because either you will have one very large column with the key and no place for the description. Or your key will be on several lines and that won't help identifying the one you're looking for by quickly scanning the doc.

Having the keys being easily identifiable is very important when dealing with a configuration documentation.

Also keep in mind that for some of the configuration properties, the description is really long and has several paragraphs which won't be table-friendly. And no, I don't want javascript buttons everywhere to show/hide the descriptions. And, yes, I want the documentation, especially the reference one to be exportable as a PDF (see my comment about that but you apparently discarded it).

What I meant is that the advantages you see in a table format are not the ones that everybody sees. There are a lot of drawbacks too and I obviously don't have the same priorities as you when choosing the format. That doesn't make someone wrong. And yes, that's exactly with these words you started this discussion, thus my comments about that posture.

FroMage:
See https://modules.ceylon-lang.org/repo/1/ceylon/language/1.3.3/module-doc/api/index.html#byItem for an example of how to make longer-than-one-line descriptions collapsible. For HTML that's really the best option because when scanning for config keys, the second lines won't be useful until you've found one key that you suspect is useful, in which case you can click on it.

Having a variable height for each config line is a big UX problem because it makes it impossible to scan since you're left to guess where the next line will start. It's also something that does not scale past ten entries, which we tend to exceed many times.

> The fact that we have very long keys will make the table hardly readable because either you will have one very large column with the key and no place for the description

The link above shows a layout where the description is below the key, which I'm not especially fond of, but it would alleviate your concern.

> Having the keys being easily identifiable is very important when dealing with a configuration documentation.

I agree, and just putting them in bold is enough for that:

Screenshot from 2019-08-20 14-45-13

Now, we can combine a few columns too:

Screenshot from 2019-08-20 14-51-20

And we can even put the description below the key like we did in ceylondoc:

Screenshot from 2019-08-20 14-55-30

There are many options, all of them solve the issues we currently have: they're compact, easy to scan, structured and regular.

We can certainly switch the order of columns, or even make the default value float on the top right or whatever.

FroMage:
> There are a lot of drawbacks too

You've listed two.

About PDF: sure the thing should be exportable to PDF, but that's much less important than HTML in terms of consumption, and I'm sure we can make it work for PDF without any issue.

FroMage:
Examples of structured docs: https://docs.micronaut.io/latest/guide/configurationreference.html (though it's quite poorly documented) and yeah, even https://docs.spring.io/spring-boot/docs/current/reference/html/common-application-properties.html can be called that, even though that's pretty horrible and I hope they have something better than this that I just didn't find.

In both cases the docs can be scanned easily because everything has a standard placement and every line has the same height. Of course we can do better than those two.

gsmet:
> Having a variable height for each config line is a big UX problem because it makes it impossible to scan since you're left to guess where the next line will start. It's also something that does not scale past ten entries, which we tend to exceed many times.

Exactly.

And as said, I would prefer avoiding Javascript and dynamic behaviors as I think having a purely Asciidoc documentation is better in the long term and far easier to maintain. You won't be able to solve the long description with standard Asciidoctor.

All your screenshots use a short one because, yes, I totally agree with you, unequal height rows are a pain. Except for the Ceylon doc which is using Javascript to show/hide things.

I want to generate pure Asciidoc from this so if you want Javascript, that means having some sort of Javascript enhancer on the website to tweak the output, then we would have to maintain it when the Asciidoc output changes for whatever reason, typically on upgrade.

Also having text hidden by default is not a good thing when you do a text search in your page using your browser.

That won't solve the PDF output issue and the mobile/tablet issue either, which all come for free with standard output. And sure, you don't code on your tablet/phone but having a quick look at the doc on your phone/tablet doesn't look far fetched to me.

For me, the current output is a good compromise on readability and long term maintainability.

We will make progress on the current output for now as we still have a couple of technical problems to solve - the order being one of them and I want to be able to consume the first version of it soon. I think it's a good compromise and we're making progress on it to improve it.

I'll let you discuss the merits of each approach with Emmanuel when he's back. But if we go the Javascript enhancer way, I want to see a commitment that it will be properly maintained because I won't be the one maintaining it.

And it will probably be easier once this PR is in as you will be able to play with the output for ALL the extensions and check the approaches globally on ALL the extensions and in ALL the formats we want.

BTW, if I had to go the Javascript way, I would make it a totally separate Javascript app called "The Quarkus Configurator (TM)", based on some JSON output with a search feature and so on and the full list of all the properties for all extensions. I would be able to quickly search for a configuration property, click on it to add it to my "virtual config file" and then get the whole content once I'm done adding all the properties I want and paste it in my `application.properties` to set the values there. But that's not the purpose of this work for now, even if part of it can be used for building that at some point. The purpose of this work is to be able to easily include these things in an Asciidoc documentation.

FroMage:
Asciidoctor can generate tables, and can conditionally add HTML headers only for HTML output, so we can make it include a JS and CSS files which will collapse entries by default and expand on click. That's about 10 lines of JS. I can totally maintain that. The only thing is we have to not generate label text such as "default value: ", because that's costly to do in JS (it will flicker).

This way PDF output will be expanded, which is the only sane thing to do anyway.

dmlloyd:
I agree with @FroMage - and I'd go farther to say that I think the mobile experience is waaay less important than making the docs efficient to read and understand.  How many people are coding on their phone?

gsmet:
I wouldn't reduce the discussion to mobile support...

Typically, I already mentioned you won't be able to search the collapsed text with your browser search. And IMHO, that's a usability issue.

But I give up on this discussion.

Feel free to experiment with this once we have a more finalized version. But please check that **all** the separate documentations renders well when integrated on the website.

Because I know the current output will scale easily with long keys/long descriptions and I want the final output to scale properly and be usable for all the extensions.

Also check the output with all the common browsers. And be sure to check regularly that it's not broken.

And the PDF output should be checked too for all the generated configuration.

Thanks.

dmlloyd:
JavaDoc uses an abbreviated table (no collapsing) and then expands the documentation down below.  I think that's a smart approach and might achieve the best of all worlds: searchability, usability on mobile (fwiw), and completeness of information.  The amount of metadata that we add to config properties is only going to increase (for example #2489 will add min and max values, plus format info; at some point perhaps converters will be able to add customized metainformation as well).  Perhaps the "type" is less important in this light.

gsmet:
@dmlloyd yeah, having both might be a good idea.

machi1990:
Thanks all for your worth suggestions. I have integrated some of your suggestions especially the solution suggested by @dmlloyd on having a table summary format and a descriptive part. 

The summary table has four columns: 
-  *Configuration property* the name of the configuration property. It is also link permitting navigation to its corresponding description. 
-  *Type*  whenever possible this permits navigation to the JavaDoc, 
-  *Default* indicating the default value and
-  *A lifecycle icon* depicting if the config is build time only or overridable at runtime.
  
As you might have noticed the table has no description as I opted to not clutter it with a lot of information.  

Just below the table we have a detailed documentation, containing detailed description as written in the ConfigItem JavaDoc.   

Example generated documentations are available here: https://github.com/machi1990/quarkus-generated-extension-config-docs/blob/master/src/main/asciidoc/generated/ 

You can check the following docs for example:
-  [quarkus-agroal.adoc](https://github.com/machi1990/quarkus-generated-extension-config-docs/blob/master/src/main/asciidoc/generated/quarkus-agroal.adoc)
-  [quarkus-arc.adoc](https://github.com/machi1990/quarkus-generated-extension-config-docs/blob/master/src/main/asciidoc/generated/quarkus-arc.adoc)
- [quarkus-core-log-config.adoc](https://github.com/machi1990/quarkus-generated-extension-config-docs/blob/master/src/main/asciidoc/generated/quarkus-core-log-config.adoc)
- [quarkus-vertx-web.adoc](https://github.com/machi1990/quarkus-generated-extension-config-docs/blob/master/src/main/asciidoc/generated/quarkus-vertx-web.adoc) 
- [quarkus-vertx.adoc](https://github.com/machi1990/quarkus-generated-extension-config-docs/blob/master/src/main/asciidoc/generated/quarkus-vertx.adoc) 

It looks like a good first iteration so I'll promote this PR. Other eventual improvements will be done iteratively.  Thanks.

FroMage:
> It looks like a good first iteration

No, it is a _very_ good first iteration. Thanks a lot!

FroMage (approved):

gsmet:
@FroMage did you make a code review on this one?

Because I had plans to do so before merging.

machi1990:
> Because I had plans to do so before merging.

@gsmet We can still review the code and I'll look into integrating the feedback before the second iteration.

FroMage:
> Because I had plans to do so before merging.

Yeah, sorry, my fault. I thought you guys already reviewed it many times. I did look at the code and it seemed fine. I saw the hump camel iterator copied in there from quarkus-runtime but I figured it had to be copied to avoid the dependency, so it seemed acceptable. Nothing else looked bad to me.

emmanuelbernard:
Follow up work captured at https://github.com/quarkusio/quarkus/issues/3803

PR #3432, opened 2019-08-06
Introduce support for Spring Web

geoand:
Spring Web REST annotations can now be used instead of JAX-RS ones for defining REST endpoints


geoand:
Some dependency issue, will fix ASAP

geoand:
@gsmet could you take a look when you have the time?

patriot1burke:
Great work!  I don't have any suggestions for improvement.

geoand:
Thanks a lot @patriot1burke!

n1hility (approved):
This looks great. I only have two comments that are really future tasks

1. We should probably change the project creation generation to be influenceable so that when the user picks spring they get a spring greeting 
2. We should probably think about whether its worth creating a spring API only dependency, that way we don't end up with issues caused by spring implementation classes conflicting with our code or inadvertently getting picked up by the native-image generation process. It also would help establish what APIs we do and don't support.

n1hility:
https://github.com/quarkusio/quarkus/issues/3436

geoand:
@n1hility added #3437 for your other point. I'll probably have someone from my team have a look since I think it's a good first issue

PR #3446, opened 2019-08-07
convert Javadoc of configuration properties to Asciidoc

### LINKED ISSUE #3433: Convert Javadoc of configuration properties to Asciidoc
Replace the Javadoc for `@ConfigItem` and friends to Javadoc format.

Make sure the documentation is useful for people and complete. It will be what the guide will show.

From 

```
        /**
         * Schema genration config
         * The default is {@code none}.
         * <p>
         * Same as JPA's javax.persistence.schema-generation.database.action.
         */
        @ConfigItem
        public Optional<String> generation;
```

Go for 

```
        /**
         * Select whether the database schema is generated or not.
         * Options are `none`, `create`, `drop-and-create`, `drop`, `update`}.
         * The default is `none`.
         *
         * `drop-and-create` is particularly useful in Quarkus Lice Coding mode
         * 
         * Same as JPA's javax.persistence.schema-generation.database.action.
         */
        @ConfigItem
        public Optional<String> generation;
```

All extensions should use that model.

machi1990:
Fixes #3433

/cc @emmanuelbernard @gsmet

### DIFF
--- a/core/runtime/src/main/java/io/quarkus/runtime/ThreadPoolConfig.java
+++ b/core/runtime/src/main/java/io/quarkus/runtime/ThreadPoolConfig.java
@@ -38,8 +38,8 @@ public class ThreadPoolConfig {
      * The executor growth resistance.
      *
      * A resistance factor applied after the core pool is full; values applied here will cause that fraction
-     * of submissions to create new threads when no idle thread is available. A value of {@code 0.0f} implies that
-     * threads beyond the core size should be created as aggressively as threads within it; a value of {@code 1.0f}
+     * of submissions to create new threads when no idle thread is available. A value of `0.0f` implies that
+     * threads beyond the core size should be created as aggressively as threads within it; a value of `1.0f`
      * implies that threads beyond the core size should never be created.
      */
     @ConfigItem(defaultValue = "0")
@@ -55,7 +55,7 @@ public class ThreadPoolConfig {
 
     /**
      * The amount of time to wait for thread pool shutdown before tasks should be interrupted. If this value is
-     * greater than or equal to the value for {@link #shutdownTimeout}, then tasks will not be interrupted before
+     * greater than or equal to the value for `shutdown-timeout`, then tasks will not be interrupted before
      * the shutdown timeout occurs.
      */
     @ConfigItem(defaultValue = "10")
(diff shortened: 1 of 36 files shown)


n1hility:
@machi1990 sorry for the delay on this one. Not sure if @emmanuelbernard mentioned this, but we were thinking about using AsciiDoclet for processing the configuration classes, and then coming up with a summarization strategy for the configuration file. What do you think about this approach?

machi1990:
> @machi1990 sorry for the delay on this one. Not sure if @emmanuelbernard mentioned this, but we were thinking about using AsciiDoclet for processing the configuration classes, and then coming up with a summarization strategy for the configuration file. What do you think about this approach?

@n1hility thanks for the feedback, I saw that in the mailing list. We can add AsciiDoclet, also if there is a summarization strategy defined (was it reading only the first paragraph when generating an `example.properties` file?) I can work on it this weekend to make this PR ready for review. Once this PR lands, it will be a huge step for the doc generation part.

dmlloyd:
This seems ill-advised to me: all existing tooling (IDEs) will not know what to do with asciidoc in javadoc blocks and expect doc tags, taglets, and HTML.  Why again do we not parse javadoc correctly and then emit the needed result on the back end?  Seems very, very unclear to me and also does not really match the config design either to do it this way.  Augmentation could probably produce these docs much more simply since all the inputs are there, just like for example config file generation.  Is it just because nobody is able to correctly parse the doc inputs?

n1hility:
@dmlloyd so the idea was to allow for rich docs under a config value instead of just a basic summary and parameter description, by just using asciidoclet on those config based java files (we'd need to figure out the selection mechanism).  The drawbacks are pretty small in the case of just config java files since it's not common (at least for users) to directly refer to them from java and see the IDE popups.  The advantage though is its much easier to write docs for content like this, since the javadoc grammar is quite cumbersome and forces you into raw html.

gsmet:
FWIW, it's not possible right now to have only part of the doc using Asciidoclet. We created https://github.com/asciidoctor/asciidoclet/pull/91 a while ago with Steve for Hibernate ORM 6. Looks like it's time to get this one in, independently from the decision we take for Quarkus. Will do.

PR #3459, opened 2019-08-08
Tika extension configuration support

### LINKED ISSUE #3167: Make Tika extension configurable
**Description**
At the moment the Tika extension is very basic - it does not let users customize any part of the internal Tika flow while Tika itself can be configured deeply with [TikaConfig](https://tika.apache.org/1.21/configuring.html),

**Implementation ideas**
Introduce `TikaConfiguration` for loading the native Tika `TikaConfig`, but also support the most mainstream configuration options directly and generate `TikaConfig` internally for users not having to type `tika-config.xml` (though this kind of simplification may not be immediately done as some user tries will be needed to check what TikaConfig properties are used most often).

A user has also reported earlier that with the Tika extension an extra step removing the empty lines was needed compared to the direct use of PDFBox - it will need to be confirmed if configuring a Tika PDFParser at the TikaConfig level could address this issue, but in any case letting the users register the custom ContentHandlers will be needed

sberyozkin:
It allows to load `tika-config.xml` resources which can be used to customize the parsing. Initially I also wanted to configure the custom `ContenHandler` names but it found out the recording of the `ContentHandler` instances has the side effects since they can track some state.
The configuration will be expanded, there are so many options that can be enabled in `tika-config.xml` that I'm not sure yet which options to support directly as the Quarkus properties, I think few typical options for the most popular formats like PDF and ODT, etc can be directly supported but I need some time to see what the users try most often.
Fixes #3167, #3460


sberyozkin:
@gsmet This PR is not ready for the review yet, 

I've realized, while looking at #3460, that adding ContentHandler makes the API a bit ambiguous (#3460 requires a special type of ContentHandler and as such it may not be clear what is the user provided ContentTyoe will do in that case). I'm coming to the conclusion that this extension should not try to be ultimately flexible and instead focus on getting the content converted to the text in the plain format only (which is done now) and make sure it can be tuned for the mainstream cases.

sberyozkin:
Having another thought about the custom ContentHandlers. It can be quite restrictive after all if the extension does not allow for it. I'm thinking throwing an ISE instead if a user chooses to do a per-embedded document parsing as part of #3460 and passes a custom ConetntHandler which should be fine, and the restriction will be doc-ed.

extensions/tika/runtime/src/main/java/io/quarkus/tika/runtime/TikaRecorder.java
```diff
@@ -0,0 +1,42 @@
...
+        initialize(config);
+        TikaParserProducer producer = container.instance(TikaParserProducer.class);
+        producer.initialize(parser);
+    }
+
+    private void initialize(TikaConfiguration config) {
+        if (parser == null) {
+            // tika-config.xml resource
+            TikaConfig tikaConfig = null;
+            InputStream is = getClass().getResourceAsStream(
```
gsmet: Are you sure the `InputStream` is properly closed? This seems a bit fragile to me.

extensions/tika/runtime/src/main/java/io/quarkus/tika/runtime/TikaRecorder.java
```diff
@@ -0,0 +1,42 @@
...
+    private static volatile TikaParser parser;
+
+    public void initTikaParser(BeanContainer container, TikaConfiguration config) {
+        initialize(config);
+        TikaParserProducer producer = container.instance(TikaParserProducer.class);
+        producer.initialize(parser);
+    }
+
+    private void initialize(TikaConfiguration config) {
+        if (parser == null) {
```
gsmet: You don't need that, your method will be called once and only once. Or did I miss something?

extensions/tika/runtime/src/main/java/io/quarkus/tika/TikaParser.java
```diff
@@ -68,8 +87,11 @@ protected TikaContent parseStream(InputStream entityStream, String contentType,
     }
 
     private Parser createParser() {
-        // TODO Use the properties to support loading the format-specific and other generic parsers
-        return new AutoDetectParser();
+        return new AutoDetectParser(tikaConfig);
+    }
+
+    private ContentHandler createContentHandler() {
+        return new ToTextContentHandler();
```
gsmet: So what you could do here is make the default content handler a bean marked with `@DefaultBean`. Then the user would be able to easily tweak the default one.

gsmet (commented):
I added a couple of comments and suggestions.

extensions/tika/runtime/src/main/java/io/quarkus/tika/TikaParser.java
sberyozkin: That one, as well as other often used ones are inside the Tika `tiika-parsers` module, this is why I thought adding few overrides to `TikaParser` can close the gap, and ContentHandler has to be instantiated per request, so I'm not sure yet if `@DefaultBean` can help. 
The other point is that the content per embedded document approach complicates things further, so I'm adding some content handler validation and some specific instantiation if that strategy is chosen. Hoping to update the PR shortly (with the above 2 comments being addressed) and then I'll ping you to do the review :-)

extensions/tika/runtime/src/main/java/io/quarkus/tika/runtime/TikaRecorder.java
sberyozkin: I  copied the null check from the mail client recorder :-), but I'm happy to remove the check

sberyozkin:
@gsmet I think it should be enough now as far as the next iteration is concerned, and it took me awhile to complete it :-). 
So the users will be able to configure the parsers via `tika-config.xml` and choose to access the embedded documents individually (by default it is appended to the master text which is OK for indexing etc but some users may want to access the embedded content individually). I have added few comments and some Javadocs. Will do the Tika extension docs once this PR passes the review.
There is much more that can be improved but I'd rather go step by step, Perhaps the next high level feature would be add the tests for the image text recognition with Tesseract OCR parser, add the limit on the file sizes, deal with encrypted PDFs, the list goes on :-) Hope the users will start driving it further as well. 
Cheers

extensions/tika/runtime/src/main/java/io/quarkus/tika/TikaParser.java
```diff
@@ -67,9 +115,28 @@ protected TikaContent parseStream(InputStream entityStream, String contentType,
...
+
+    private ContentHandler createContentHandler() {
+        return this.appendEmbeddedContent ? new ToTextContentHandler()
+                : new RecursiveParserWrapperHandler(
+                        new BasicContentHandlerFactory(BasicContentHandlerFactory.HANDLER_TYPE.TEXT, -1));
+    }
+
+    private ContentHandler createContentHandlerForMetadataOnly(String contentType) {
+        if (this.appendEmbeddedContent) {
+            return contentType != null && contentType.contains("pdf") ? null : new DefaultHandler();
```
gsmet: I'm not sure I understand why pdf is special cased?

extensions/tika/runtime/src/main/java/io/quarkus/tika/runtime/TikaConfiguration.java
```diff
@@ -0,0 +1,26 @@
...
+@ConfigRoot(phase = ConfigPhase.BUILD_AND_RUN_TIME_FIXED)
+public class TikaConfiguration {
+    /**
+     * The path to the tika-config.xml
+     */
+    @ConfigItem(defaultValue = "tika-config.xml")
+    public String tikaConfigPath;
+
+    /**
+     * Controls how the the content of the embedded documents is parsed.
```
gsmet:
```suggestion
     * Controls how the content of the embedded documents is parsed.
```

extensions/tika/runtime/src/main/java/io/quarkus/tika/runtime/TikaRecorder.java
```diff
@@ -0,0 +1,50 @@
...
+import org.apache.tika.parser.RecursiveParserWrapper;
+
+import io.quarkus.arc.runtime.BeanContainer;
+import io.quarkus.runtime.annotations.Recorder;
+import io.quarkus.tika.TikaParseException;
+import io.quarkus.tika.TikaParser;
+
+@Recorder
+public class TikaRecorder {
+    private static volatile TikaParser parser;
```
gsmet: You don't need that anymore and it will make the code cleaner.

gsmet (changes requested):
I added a few minor suggestions/questions. Could you have a look?

extensions/tika/runtime/src/main/java/io/quarkus/tika/TikaParser.java
sberyozkin: I'm fond of this check because it is my only contribution to the PDF parsing story in Tika :-). Typically, when one needs the metadata only, the parsers are still processing and reporting the text which is not needed, ideally. But as I was explained, some formats have the metadata mixed in with the text, etc, and it is impossible to avoid parsing the text. Luckily, for the PDF format, it is possible, by providing a null handler, completely avoid the text processing work.

sberyozkin: I'll add the comment to the code

sberyozkin:
@gsmet Updated the `TikaRecorder`, and the docs/comments

gsmet (approved):
OK, let's wait for CI. Thanks!

sberyozkin:
@gsmet just noticed a test excel file name I copied it from Tika needs a fix, will sort it out later

PR #3464, opened 2019-08-08
Fix Gradle plugin --docker-build flag

### LINKED ISSUE #2786: --docker-build flag seems to have no effect with Gradle plugin
Since 0.15.0 version, when buildNative gradle task is invoked it asks for GRAALVM_HOME environment variable regardless of whether it is being run with --docker-build=true or --docker-build=false

**Expected behavior**
I would expect the build to not require GRAALVM_HOME set when using docker image as a build environment

 - Output of `uname -a` or `ver`: **Linux localhost.localdomain 5.1.7-300.fc30.x86_64 #1 SMP Wed Jun 5 12:32:28 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux**
 - Output of `java -version`: **openjdk version "1.8.0_212"**
 - GraalVM version (if different from Java): **graalvm-ce-1.0.0-rc16**
 - Quarkus version or git rev: **0.16.1**

gwenneg:
Fixes #2786

### DIFF
--- a/devtools/gradle/src/main/java/io/quarkus/gradle/tasks/QuarkusNative.java
+++ b/devtools/gradle/src/main/java/io/quarkus/gradle/tasks/QuarkusNative.java
@@ -217,9 +217,6 @@ public class QuarkusNative extends QuarkusTask {
     @Optional
     @Input
     public String getGraalvmHome() {
-        if (graalvmHome == null || graalvmHome.length() < 1)
-            throw new GradleException(
-                    "The GRAALVM_HOME environment variable need to be set to your GraalVM root directory to use native mode");
         return graalvmHome;
     }
 


devtools/gradle/src/main/java/io/quarkus/gradle/tasks/QuarkusNative.java
```diff
@@ -217,7 +217,8 @@ public void setEnableCodeSizeReporting(boolean enableCodeSizeReporting) {
     @Optional
     @Input
     public String getGraalvmHome() {
-        if (graalvmHome == null || graalvmHome.length() < 1)
+        if ((dockerBuild == null || "false".equals(dockerBuild.toLowerCase()))
```
gwenneg: This is the same test as in [NativeImagePhase.java#L229](https://github.com/quarkusio/quarkus/blob/master/core/creator/src/main/java/io/quarkus/creator/phase/nativeimage/NativeImagePhase.java#L229).

geoand: Haven't looked at this stuff extensively, but why can we not just remove the part that checks if `graalVmHome` was provided?
The Maven plugin doesn't perform this kind of check

gwenneg: You're right, NativeImagePhase already has a GRAALVM_HOME lookup logic that seems fine, there's no reason to do that in the Gradle plugin. It also seems better to have the same checks in both plugins. Thanks for suggesting this!

geoand: You are welcome!

geoand (approved):
Thanks for looking into this!

PR #3471, opened 2019-08-09
Read sources using the UTF_8 charset during hot reload compilation

### LINKED ISSUE #3467: quarkus:dev hot reload changes default charset?
Running a project with quarkus:dev, everything works fine. When triggering hot reload (changing some source), the behavior changes and Strings that exists in the code seems to be interpreted in a different charset.

I created a small project to reproduce this error: https://github.com/seseso/quarkus-hot-reload
Steps: 
1. mvn clean install -DskipTests quarkus:dev
2. curl http://localhost:8085/hello/S%C3%A9rgio
3. Excpected output: hello Boss!
4. Change a source and save, triggering hot reload
5. curl http://localhost:8085/hello/S%C3%A9rgio
6. Excpected output: hello Boss! Actually, it returns Hello Sérgio

The rest endpoint is this one:

    @GET
    @Path("/{name}")
    public String hello(@PathParam("name") String name) {
        
        if("Sérgio".equals(name)) {
            return "hello Boss!";
        }
        
        return "hello " + name;
    }

So, after hot reload, the if does not matches.

**Environment**
 - Output of `uname -a` or `ver`: Windows 10 PRO
 - Output of `java -version`: openjdk version "1.8.0_212"
 - GraalVM version (if different from Java): Not using
 - Quarkus version or git rev: 0.20.0

If more information is needed, please let me know as I'll be glad to provide...

gwenneg:
Works around #3467

### DIFF
--- a/core/devmode/src/main/java/io/quarkus/dev/JavaCompilationProvider.java
+++ b/core/devmode/src/main/java/io/quarkus/dev/JavaCompilationProvider.java
@@ -3,6 +3,7 @@ package io.quarkus.dev;
 import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
+import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
@@ -42,7 +43,8 @@ public class JavaCompilationProvider implements CompilationProvider {
             throw new RuntimeException("No system java compiler provided");
         }
         DiagnosticCollector<JavaFileObject> diagnostics = new DiagnosticCollector<>();
-        try (StandardJavaFileManager fileManager = compiler.getStandardFileManager(diagnostics, null, null);) {
+        try (StandardJavaFileManager fileManager = compiler.getStandardFileManager(diagnostics, null,
+                StandardCharsets.UTF_8)) {
 
             fileManager.setLocation(StandardLocation.CLASS_PATH, context.getClasspath());
             fileManager.setLocation(StandardLocation.CLASS_OUTPUT, Collections.singleton(context.getOutputDirectory()));


geoand:
I think that this is fine for now, but I would like to know what @gsmet thinks as well

gwenneg:
Thanks @geoand!

I took some time to explore the dynamic encoding option. It's pretty simple with the Maven plugin (I've got some code ready for a review if needed) and not that simple with Gradle (I didn't find a clean way to retrieve the encoding so far). I'll put this on hold for now and continue the exploration later if needed.

geoand:
If you could get that working for Gradle too, it would be awesome.

I am of the opinion that we should fix the most common case now and improve later on

core/devmode/src/main/java/io/quarkus/dev/JavaCompilationProvider.java
```diff
@@ -42,7 +43,8 @@ public void compile(Set<File> filesToCompile, Context context) {
             throw new RuntimeException("No system java compiler provided");
         }
         DiagnosticCollector<JavaFileObject> diagnostics = new DiagnosticCollector<>();
-        try (StandardJavaFileManager fileManager = compiler.getStandardFileManager(diagnostics, null, null);) {
+        try (StandardJavaFileManager fileManager = compiler.getStandardFileManager(diagnostics, null,
+                StandardCharsets.UTF_8)) {
```
gsmet: I agree we should have a fix in Wednesday's release.

Could you promote this PR and also create a proper GH issue. And I think it would be interesting to create a draft PR with your Maven work so that we don't lose it.

gsmet (approved):

core/devmode/src/main/java/io/quarkus/dev/JavaCompilationProvider.java
gwenneg: PR promoted. Regarding the GH issue, there's already #3467.

gsmet:
Merged, thanks!

gwenneg:
I created #3487 for the dynamic encoding issue. I'll continue investigating the Gradle case this week.

PR #3475, opened 2019-08-11
Align command and gradle to use ListExtensions as Maven

### LINKED ISSUE #3473: Align QuarkusListExtensions and ListExtensionsCommand to ListExtensionsMojo
**Description**
List extension don't display the same format, because it's not using the same code

**Implementation ideas**
Align QuarkusListExtensions and ListExtensionsCommand to ListExtensionsMojo implementation

Dufgui:
fix #3473

### DIFF
--- a/devtools/aesh/src/main/java/io/quarkus/cli/commands/ListExtensionsCommand.java
+++ b/devtools/aesh/src/main/java/io/quarkus/cli/commands/ListExtensionsCommand.java
@@ -1,14 +1,18 @@
 package io.quarkus.cli.commands;
 
+import java.io.File;
+import java.io.IOException;
+
 import org.aesh.command.Command;
 import org.aesh.command.CommandDefinition;
 import org.aesh.command.CommandException;
 import org.aesh.command.CommandResult;
 import org.aesh.command.invocation.CommandInvocation;
 import org.aesh.command.option.Option;
+import org.aesh.io.Resource;
 
-import io.quarkus.dependencies.Extension;
-import io.quarkus.maven.utilities.MojoUtils;
+import io.quarkus.cli.commands.writer.FileProjectWriter;
+import io.quarkus.generators.BuildTool;
 
 /**
  * @author <a href="mailto:stalep@gmail.com">Ståle Pedersen</a>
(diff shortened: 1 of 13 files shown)


aloubyansky:
@Dufgui this one needs a rebase. Thanks!

Dufgui:
@aloubyansky ready for review

devtools/common/src/main/java/io/quarkus/cli/commands/file/GradleBuildFile.java
```diff
@@ -194,4 +217,51 @@ protected boolean containsBOM() {
...
+        private List<Dependency> result;
+        private boolean inConstraints = false;
+        private boolean inImplementation = false;
+
+        public GetConstraints(List<Dependency> result) {
+            this.result = result;
+        }
+
+        @Override
+        public void accept(String currentLine) {
```
aloubyansky: This does not appear to catch `enforcedPlatform` or `platform`. This is inconsistent with Maven's `model.getDependencyManagement()`. Note sure whether it's a big deal though.

devtools/aesh/src/main/java/io/quarkus/cli/commands/ListExtensionsCommand.java
```diff
@@ -19,24 +23,36 @@
...
-
     @Option(shortName = 'a', hasValue = false, description = "Display name, group-id, artifact-id and version (default behaviour)")
-    private boolean all;
+    private boolean all = false;
+
+    @Option(shortName = 'f', hasValue = true, description = "Display in simplified format, only display the extension names or full format")
+    private String format = "simple";
+
+    @Option(shortName = 's', hasValue = true, description = "Search filter on extension list. The format is based on Java Pattern.")
+    private String search;
```
aloubyansky: To align completely, shouldn't be it be `searchPattern`?

devtools/aesh/src/main/java/io/quarkus/cli/commands/ListExtensionsCommand.java
```diff
@@ -19,24 +23,36 @@
...
+
+    @Option(shortName = 'p', description = "path for the project")
```
aloubyansky: "Path to the project."

devtools/common/src/main/java/io/quarkus/cli/commands/file/BuildFile.java
```diff
@@ -52,7 +59,7 @@ public boolean addExtensionAsGAV(String query) {
             addDependencyInBuildFile(parsed);
             return true;
         } else {
-            PRINTER.noop(" Dependency " + parsed.getManagementKey() + " already in the pom.xml file - skipping");
+            PRINTER.noop(" Skipping dependency " + parsed.getManagementKey() + " already present");
```
aloubyansky: " Skipping already present dependency " + ...

aloubyansky (changes requested):

devtools/common/src/main/java/io/quarkus/cli/commands/file/GradleBuildFile.java
Dufgui: done.

I am just dig a little more to not use string and delegate parse of model to gradle lib.
I see how eclipse do it:
https://github.com/eclipse/buildship/blob/master/org.eclipse.buildship.kotlin/build.gradle
https://github.com/eclipse/buildship/blob/master/org.eclipse.buildship.kotlin/src/main/java/org/eclipse/buildship/kotlin/internal/GradleKotlinScriptTemplateProvider.java#L80

And with ProjectConnection we can load the groovy model  too, but I don't know the limitation and we need all gradle core as dependency. Not sur we want this....
Do you want I create an issue for this @aloubyansky ?

aloubyansky: Ideally, yes, we want to operate on a representation of the config model, as in Maven equivalent. Relying on string parsing is kind of fragile, since e.g. the formatting or composition of the config could vary. So, yes, I think it's worth creating an issue for that. Thanks @Dufgui

aloubyansky (approved):

PR #3490, opened 2019-08-12
Fix NPE on build project without src

### LINKED ISSUE #3225: NPE when trying to build a maven module that has no src folder
```
[ERROR] Failed to execute goal io.quarkus:quarkus-maven-plugin:0.19.1:build (default) on project behavior-definition: Failed to build a runnable JAR: Failed to build a runner jar: NullPointerException -> [Help 1]
```

Error handling should be improved here.

Cause: the module had no src folder structure defined

Dufgui:
fix #3225


aloubyansky:
@Dufgui have you actually tried whether it works? This may fail in QuarkusAugmentor. If it does fail, a simple fix could be to create that empty directory. But this needs actual verification.

Dufgui:
Yes I have another error after but It was late and I think it's the same as fix here:
https://github.com/quarkusio/quarkus/pull/3491/files

In fact no 👎

Dufgui:
I try both approach. fix each error or create empty directory. 
The code is clearer and works with create empty directory so i push this approach.

aloubyansky (approved):

PR #3493, opened 2019-08-13
Move Undertow substitutions to their own module so that camel-undertow can make use of them

stuartwdouglas:
…w can make use of them


lburgazzoli:
+1

gsmet (approved):

Issue #3443, opened 2019-08-07
CDI tries to inject variables passed to lambda function


gjeraj:
The problem lies with the `quarkus-arc` extension where it tries to inject parameters passed to lambda expression. To illustrate this I've created a simple project https://github.com/gjeraj/quarkus-cdi-bug

```
@ApplicationScoped
public class EventHandler {
	@Inject
	public LambdaHandler lambdaHandler;

	public void event(@Observes String event) {
		// Fails
		lambdaHandler.process("First", (first, second) -> event + " " + second);
	}

	public void direct(String event) {
		// Works
		lambdaHandler.process("First", (first, second) -> event + " " + second);
	}
}
```
As you can see the event observer function fails to build with the following error:

```
[ERROR] Failed to execute goal io.quarkus:quarkus-maven-plugin:0.20.0:build (default) on project quark: Failed to build a runnable JAR: Failed to build a runner jar: Failed to augment application classes: Build failure: Build failed due to errors
[ERROR]         [error]: Build step io.quarkus.arc.deployment.ArcProcessor#validate threw an exception: javax.enterprise.inject.spi.DeploymentException: Found 2 deployment problems:
[ERROR] [1] Unsatisfied dependency for type java.lang.String and qualifiers [@Default]
[ERROR]         - java member: com.test.EventHandler#lambda$event$0()
[ERROR]         - declared on io.quarkus.arc.processor.ObserverInfo@169f2f1b
[ERROR] [2] Unsatisfied dependency for type java.lang.String and qualifiers [@Default]
[ERROR]         - java member: com.test.EventHandler#lambda$event$0()
[ERROR]         - declared on io.quarkus.arc.processor.ObserverInfo@169f2f1b
```
Whats interesting if `event` parameter is first assigned to another variable, it also works:

```
	public void event(@Observes String event) {
		// Works
                String working = event;
		lambdaHandler.process("First", (first, second) -> working + " " + second);
	}
```

If the line is commented out, the second method `direct` builds just fine.

Using 
- `quarkus <= 0.20.0` 
- `maven 3.6.1 `

gwenneg:
Hi @gjeraj and thanks for reporting this.

@mkouba, @manovotn: Could you take a look at this please?

mkouba:
Hi @gjeraj, I've just tried your reproducer and it builds and runs just fine. What JDK version are you using?

gjeraj:
[...]

PR #3495, opened 2019-08-13
Arc - ignore synthetic methods when discovering observers and producers

mkouba:
- this should fix problems with lambdas on some old JDKs, see #3443
- also merge MethodUtils into the Methods class


manovotn (approved):
Looks good!

independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/Methods.java
```diff
@@ -20,12 +20,19 @@
 import org.jboss.jandex.WildcardType;
 import org.jboss.logging.Logger;
 
+/**
+ * 
+ * @author Martin Kouba
+ * @author Michal Szynkiewicz, michal.l.szynkiewicz@gmail.com
+ */
```
gsmet: Didn't we decide to not add new `@author` tags?

manovotn: The deleted file (`MethodUtils`) had Michal as author and the functionality is now merged into this class, so the authors are just preserved.
Or is there a policy not to add `@author` at all to any files?

mkouba: Arc is still developed as an "independent" project and I'd like to keep this tag ;-).

PR #3503, opened 2019-08-13
Quarkus Test Resources with same order are all ignored except one

### LINKED ISSUE #3501: Quarkus Test Resources with same order are considered equal
**Describe the bug**
I have implemented a QuarkusTestResourceLifecycleManager with default order (0). When running my tests, the LambdaResourceManager is never started.

**Expected behavior**
All Test Resources should be started.

**Actual behavior**
When debugging into TestResourceManager I noticed that Test Resources are gathered in a TreeSet.
`Set<QuarkusTestResourceLifecycleManager> testResourceRunners = new TreeSet(new QuarkusTestResourceLifecycleManagerComparator());`
So two Test Resources with the same order are considered equal and one replaces the other.

**To Reproduce**
1. Implement a QuarkusTestResourceLifecycleManager with default order.
2. Run tests for quarkus-amazon-lambda.

 - Output of `uname -a` or `ver`: Linux sku-ubuntu 5.0.0-23-generic #24~18.04.1-Ubuntu SMP Mon Jul 29 16:12:28 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux
 - Output of `java -version`: openjdk version "1.8.0_222"
OpenJDK Runtime Environment (build 1.8.0_222-20190711120915.buildslave.jdk8u-src-tar--b08)
OpenJDK 64-Bit GraalVM CE 19.1.1 (build 25.222-b08-jvmci-19.1-b01, mixed mode)
 - Quarkus version or git rev: 0.20.0

**Additional context**
Introduced with #3206.

machi1990:
Fixes #3501

### DIFF
--- a/test-framework/common/src/main/java/io/quarkus/test/common/TestResourceManager.java
+++ b/test-framework/common/src/main/java/io/quarkus/test/common/TestResourceManager.java
@@ -13,12 +13,14 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.StandardOpenOption;
 import java.nio.file.attribute.BasicFileAttributes;
+import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.LinkedHashSet;
+import java.util.List;
 import java.util.Map;
 import java.util.ServiceLoader;
 import java.util.Set;
-import java.util.TreeSet;
 
 import org.jboss.jandex.AnnotationInstance;
 import org.jboss.jandex.DotName;
@@ -27,7 +29,7 @@ import org.jboss.jandex.Indexer;
 
 public class TestResourceManager {
 
-    private final Set<QuarkusTestResourceLifecycleManager> testResources;
+    private final List<QuarkusTestResourceLifecycleManager> testResources;
     private Map<String, String> oldSystemProps;
 
     public TestResourceManager(Class<?> testClass) {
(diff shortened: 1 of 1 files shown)


test-framework/common/src/main/java/io/quarkus/test/common/TestResourceManager.java
```diff
@@ -105,11 +107,14 @@ public void stop() {
...
         }
 
-        for (QuarkusTestResourceLifecycleManager i : ServiceLoader.load(QuarkusTestResourceLifecycleManager.class)) {
-            testResourceRunners.add(i);
+        Collections.sort(testResourceRunners, new QuarkusTestResourceLifecycleManagerComparator());
+
+        for (QuarkusTestResourceLifecycleManager quarkusTestResourceLifecycleManager : ServiceLoader
+                .load(QuarkusTestResourceLifecycleManager.class)) {
+            testResourceRunners.add(quarkusTestResourceLifecycleManager);
         }
```
gsmet: Shouldn't you sort the list when they are all added?

test-framework/common/src/main/java/io/quarkus/test/common/TestResourceManager.java
```diff
@@ -105,11 +107,14 @@ public void stop() {
...
 
-        return testResourceRunners;
+        return new HashSet<>(testResourceRunners);
```
gsmet: You will lose the ordering with this call. I think this method should return a `List` not a `Set`.

gsmet (changes requested):
Thanks! Added some comments inline.

test-framework/common/src/main/java/io/quarkus/test/common/TestResourceManager.java
machi1990: Good catch. A fixup commit incoming.

gsmet (approved):

PR #3509, opened 2019-08-14
feat: test custom exception mapper with resteasy

loicmathieu:
Now that #3317 is merged, we can add devmode specific tests.

This PR provides tests for the two already merged PR:
- Custom NotFoundExceptionMapper : #3309
- JSON representation in the default NotFoundExceptionMapper : #3401

In order to be able to test the JSON representation I had to add the jsonb provider as test dependency. If it's not desirable I can move this test to the resteady-jsonb dependency (but there is no existing test there) or remove it ...

### DIFF
--- a/extensions/resteasy/deployment/pom.xml
+++ b/extensions/resteasy/deployment/pom.xml
@@ -40,6 +40,17 @@
             <artifactId>rest-assured</artifactId>
             <scope>test</scope>
         </dependency>
+        <!-- These are needed to test the JSON representation of the NotFoundExceptionMapper-->
+        <dependency>
+            <groupId>org.jboss.resteasy</groupId>
+            <artifactId>resteasy-json-binding-provider</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>commons-io</groupId>
+            <artifactId>commons-io</artifactId>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 
     <build>
--- a/extensions/resteasy/deployment/src/test/java/io/quarkus/resteasy/test/NotFoundExceptionMapperTestCase.java
+++ b/extensions/resteasy/deployment/src/test/java/io/quarkus/resteasy/test/NotFoundExceptionMapperTestCase.java
@@ -1,0 +1,41 @@
+package io.quarkus.resteasy.test;
+
+import static org.hamcrest.Matchers.is;
+
+import org.hamcrest.Matchers;
+import org.jboss.shrinkwrap.api.ShrinkWrap;
+import org.jboss.shrinkwrap.api.spec.JavaArchive;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.RegisterExtension;
+
+import io.quarkus.test.QuarkusDevModeTest;
+import io.restassured.RestAssured;
+import io.restassured.http.ContentType;
+
+public class NotFoundExceptionMapperTestCase {
+    @RegisterExtension
+    static QuarkusDevModeTest test = new QuarkusDevModeTest()
+            .setArchiveProducer(() -> ShrinkWrap.create(JavaArchive.class)
+                    .addClasses(RootResource.class));
+
+    @Test
+    public void testHtmlResourceNotFound() {
+        // test the default exception mapper provided in dev mode : displays HTML by default
+        RestAssured.when().get("/not_found")
+                .then()
+                .statusCode(404)
+                .body(Matchers.containsString("<div class=\"component-name\"><h1>Resource Not Found</h1>"));
+    }
+
+    @Test
+    public void testJsonResourceNotFound() {
+        // test the default exception mapper provided in dev mode : displays json when accept is application/json
+        RestAssured.given().accept(ContentType.JSON)
+                .when().get("/not_found")
+                .then()
+                .statusCode(404)
+                .body("errorMessage", is("Resource Not Found"))
+                .body("existingResourcesDetails[0].basePath", is("/"));
+    }
+
+}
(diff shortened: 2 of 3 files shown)


extensions/resteasy/deployment/src/test/java/io/quarkus/resteasy/test/NotFoundExceptionMapperTestCase.java
```diff
@@ -0,0 +1,40 @@
...
+import org.jboss.shrinkwrap.api.ShrinkWrap;
+import org.jboss.shrinkwrap.api.spec.JavaArchive;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.RegisterExtension;
+
+import io.quarkus.test.QuarkusDevModeTest;
+import io.restassured.RestAssured;
+import io.restassured.http.ContentType;
+
+public class NotFoundExceptionMapperTestCase {
```
gwenneg: I don't see any source change that would trigger a hot reload in this class. Is there a point running a `QuarkusDevModeTest` without reloading?

gwenneg (commented):
I got curious about this PR since I'm also adding `QuarkusDevModeTest` to another one.

extensions/resteasy/deployment/src/test/java/io/quarkus/resteasy/test/NotFoundExceptionMapperTestCase.java
machi1990: I'll let @loicmathieu  give a detail answer but I think it is for running under the `dev` profile.

gwenneg: I just realized the `NotFoundExceptionMapper` PRs were submitted to fix an issue in dev mode, so you're right @machi1990! It makes sense to use `QuarkusDevModeTest` without reloading here :)

loicmathieu: Yes the exception mapper is only used in dev mode that what I use the QuarkusDevModeTest

extensions/resteasy/deployment/src/test/java/io/quarkus/resteasy/test/NotFoundExceptionMapperTestCase.java
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    @Test
+    public void testJsonResourceNotFound() {
+        // test the default exception mapper provided in dev mode : displays json when accept is application/json
+        RestAssured.given().accept(ContentType.JSON)
+                .when().get("/not_found")
+                .then()
+                .statusCode(404)
+                .body(Matchers.is("{\"errorMessage\":\"Resource Not Found\",\"existingResourcesDetails\":" +
```
stuartwdouglas: I think this should be testing the json content rather than an exact JSON string. For some examples see: https://github.com/rest-assured/rest-assured/wiki/usage .

I think it should be more like:

.body("errorMessage", is("Resource Not Found"))
.body("existingResourcesDetails.basePath",is("/"))  etc

loicmathieu: @stuartwdouglas done

extensions/resteasy/deployment/pom.xml
```diff
@@ -40,6 +40,17 @@
             <artifactId>rest-assured</artifactId>
             <scope>test</scope>
         </dependency>
+        <!-- These are needed to test the JSON representation of the NotFoundExeptionMapper-->
```
gsmet:
```suggestion
        <!-- These are needed to test the JSON representation of the NotFoundExceptionMapper-->
```

gsmet (changes requested):
I'm a bit ashamed to create a review just for that but I can't help it sorry ;)

Can you adjust and amend your commit? I'll merge right away.

extensions/resteasy/deployment/src/test/java/io/quarkus/resteasy/test/NotFoundExceptionMapperTestCase.java
```diff
@@ -0,0 +1,42 @@
...
+                    .addClasses(RootResource.class));
+
+    @Test
+    public void testHtmlResourceNotFound() {
+        // test the default exception mapper provided in dev mode : displays HTML by default
+        RestAssured.when().get("/not_found")
+                .then()
+                .statusCode(404)
+                .body(Matchers.containsString("<div class=\"component-name\"><h1>Resource Not Found</h1>" +
+                        "<h2>REST interface overview</h2></div><h2>/</h2>"));
```
gsmet: Oh wait, another comment here. Let's just test that it contains `<h1>Resource Not Found</h1>`.

It will be more resistant to potential future changes to the HTML output.

extensions/resteasy/deployment/pom.xml
loicmathieu: Done.
@gsmet don't worry, it's your job to verify the quality of Quarkus codebase ... and I should have been more careful about this (I really need to re-read twice my code before pushing it) ;)

extensions/resteasy/deployment/src/test/java/io/quarkus/resteasy/test/NotFoundExceptionMapperTestCase.java
loicmathieu: done

gsmet (approved):

PR #3513, opened 2019-08-14
Vertx 3.8 and new sql client

FroMage:
This is a draft PR because I want to wait for opinions about the new mysql async client.

Right now I've just updated the `quarkus-reactive-pg-client` module to use the new `vertx-sql` module, but I wonder if it should support both pg and mysql, by discriminating on the URL string. It would make it depend on both clients, though.

An alternative would be to make a new `quarkus-reactive-mysql-client` module, but it would largely be similar.

Opinions @vietj and @tsegismont ?

This may be related to the ongoing work of @gsmet about supporting multiple datasources, but I don't know what state it is in.

### DIFF
--- a/bom/runtime/pom.xml
+++ b/bom/runtime/pom.xml
@@ -100,7 +100,7 @@
         <wildfly-elytron.version>2.0.0.Alpha4</wildfly-elytron.version>
         <jboss-modules.version>1.8.7.Final</jboss-modules.version>
         <jboss-threads.version>3.0.0.Beta5</jboss-threads.version>
-        <vertx.version>3.7.1</vertx.version>
+        <vertx.version>3.8.0</vertx.version>
         <httpclient.version>4.5.9</httpclient.version>
         <httpcore.version>4.4.11</httpcore.version>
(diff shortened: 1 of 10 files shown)


gsmet:
We already support multiple datasources. Are you talking about supporting multiple persistence units?

FroMage:
> We already support multiple datasources. Are you talking about supporting multiple persistence units?

Mmmm, that's an interesting subtle difference I did not realise. So, since the reactive drivers don't need persistence units, we could already support multiple data sources at the same time with one JPA and two reactive sources (pg and mysql). We should definitely look into that.

gsmet:
@FroMage could you rebase and squash everything?

I think we should promote this PR because, the MySQL discussion is IMHO unrelated and could be started on the mailing list.

FroMage:
Yeah, that was my plan.

FroMage:
OK, rebased and squashed.

cescoffier (approved):

gsmet:
Let's merge then, thanks!

PR #3514, opened 2019-08-14
Provides a JSON representation for the default quarkus error page

loicmathieu:
The default Quarkus error page is served by Undertow in HTML representation only.

I propose to add a JSON representation (if Accept header with application/json is set) with the same information on it.

As it's on the undertow layer, I don't use any json specific library and rely on string manipulation for the implementation. Not sure it's a good idea ...

### DIFF
--- a/extensions/undertow/runtime/src/main/java/io/quarkus/undertow/runtime/QuarkusErrorServlet.java
+++ b/extensions/undertow/runtime/src/main/java/io/quarkus/undertow/runtime/QuarkusErrorServlet.java
@@ -35,9 +35,20 @@ public class QuarkusErrorServlet extends HttpServlet {
             details += "Error id " + uuid;
         }
 
-        resp.setContentType("text/html");
-        resp.setCharacterEncoding(StandardCharsets.UTF_8.name());
-        resp.getWriter().write(new TemplateHtmlBuilder("Internal Server Error", details, details).stack(stack).toString());
+        String accept = req.getHeader("Accept");
+        if (accept != null && accept.contains("application/json")) {
+            resp.setContentType("application/json");
+            resp.setCharacterEncoding(StandardCharsets.UTF_8.name());
+            String escapedStack = stack.replace(System.lineSeparator(), "\\n").replace("\"", "\\\"");
+            StringBuilder jsonPayload = new StringBuilder("{\"details\":\"").append(details).append("\",\"stack\":\"")
+                    .append(escapedStack).append("\"}");
+            resp.getWriter().write(jsonPayload.toString());
+        } else {
+            //We default to HTML representation
+            resp.setContentType("text/html");
+            resp.setCharacterEncoding(StandardCharsets.UTF_8.name());
+            resp.getWriter().write(new TemplateHtmlBuilder("Internal Server Error", details, details).stack(stack).toString());
+        }
     }
 
     private static String generateStackTrace(final Throwable exception) {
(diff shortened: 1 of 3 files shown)


loicmathieu:
As a switch to Vert.x is on the way, maybe the same should be done on the Vert.x default error page

gsmet:
Yeah, I would hold on this one because the HTTP support is in flux right now.

Let's wait for things to settle as there's a good chance we won't have undertow for simple REST services. At least, that would be an option.

extensions/undertow/runtime/src/main/java/io/quarkus/undertow/runtime/QuarkusErrorServlet.java
```diff
@@ -35,9 +35,20 @@ protected void service(HttpServletRequest req, HttpServletResponse resp) throws
             details += "Error id " + uuid;
         }
 
-        resp.setContentType("text/html");
-        resp.setCharacterEncoding(StandardCharsets.UTF_8.name());
-        resp.getWriter().write(new TemplateHtmlBuilder().error(details).stack(stack).toString());
+        String accept = req.getHeader("Accept");
+        if (accept != null && accept.contains("application/json")) {
```
gsmet: It would be nice to see if we can use some content negotiation code in Undertow. I don't know if there is any though.

loicmathieu: I cannot find any content negociation in undertow so if needed I will need to build my own. And content negociation is complex (need to leverage the weight of each part of the Accept header). 

Here we favor json each time json is part of the Accept header, else we generates HTML message. This is a very _opiniated_ content negociation but a server can choose it's own algorithm and use the Accept header _as a hint_ only it doesn't break the HTTP spec.

I can implement a better alogithm if needed. But will this PR be merged some days or need it to be build on top of Vert.x ?

Maybe a broader PR should be created to cover all error pages (resteasy, Vert.x, undertow, security, validation) for multiple accept header (HTML, json, text) ?

extensions/undertow/runtime/src/main/java/io/quarkus/undertow/runtime/QuarkusErrorServlet.java
```diff
@@ -35,9 +35,20 @@ protected void service(HttpServletRequest req, HttpServletResponse resp) throws
...
+            resp.setCharacterEncoding(StandardCharsets.UTF_8.name());
+            String escapedStack = stack.replace("\n", "\\n").replace("\"", "\\\"");
+            StringBuilder jsonPayload = new StringBuilder("{\"details\":\"").append(details).append("\",\"stack\":\"")
+                    .append(escapedStack).append("\"}");
+            resp.getWriter().write(jsonPayload.toString());
+        } else {
+            //We default to HTML representation
+            resp.setContentType("text/html");
+            resp.setCharacterEncoding(StandardCharsets.UTF_8.name());
         resp.getWriter().write(new TemplateHtmlBuilder("Internal Server Error", details, details).stack(stack).toString());
```
gsmet: I think there is an indentation issue here.

loicmathieu: fixed

extensions/undertow/runtime/src/main/java/io/quarkus/undertow/runtime/QuarkusErrorServlet.java
loicmathieu: done

gsmet (changes requested):
Failing on Windows. I pointed the issues in the review.

gsmet (approved):

gsmet:
Merged, thanks!

PR #3527, opened 2019-08-15
Add VertX BOM to Quarkus

cristianonicolai:
Adding vertx-web-graphql to bom so versions are aligned when using this library.

### DIFF
--- a/bom/runtime/pom.xml
+++ b/bom/runtime/pom.xml
@@ -213,6 +213,15 @@
                 <scope>import</scope>
             </dependency>
 
+            <!-- Vert.X dependencies, imported as a BOM -->
+            <dependency>
+                <groupId>io.vertx</groupId>
+                <artifactId>vertx-stack-depchain</artifactId>
+                <version>${vertx.version}</version>
+                <scope>import</scope>
+                <type>pom</type>
+            </dependency>
+
             <!-- Quarkus core -->
 
             <dependency>
@@ -1669,21 +1678,6 @@
                 <groupId>org.neo4j.driver</groupId>
                 <artifactId>neo4j-java-driver</artifactId>
                 <version>${neo4j-java-driver.version}</version>
-            </dependency>
-            <dependency>
-                <groupId>io.vertx</groupId>
-                <artifactId>vertx-core</artifactId>
-                <version>${vertx.version}</version>
-            </dependency>
-            <dependency>
-                <groupId>io.vertx</groupId>
-                <artifactId>vertx-web</artifactId>
-                <version>${vertx.version}</version>
-            </dependency>
-            <dependency>
-                <groupId>io.vertx</groupId>
-                <artifactId>vertx-web-client</artifactId>
-                <version>${vertx.version}</version>
             </dependency>
             <dependency>
                 <groupId>io.smallrye.reactive</groupId>


bom/runtime/pom.xml
```diff
@@ -1665,6 +1665,11 @@
                 <artifactId>vertx-web-client</artifactId>
                 <version>${vertx.version}</version>
             </dependency>
+            <dependency>
+                <groupId>io.vertx</groupId>
+                <artifactId>vertx-web-graphql</artifactId>
+                <version>${vertx.version}</version>
```
gsmet: @cescoffier @vietj I wonder if we should include the Vert.x bom in Quarkus?

cristianonicolai: +1 very likely any other artifact that is used would need to be added anyway, using the bom definitely solve for more use cases.

cescoffier: Yes, that would avoid many additional dependencies. 

However, we don't have a `bom` named `bom`. It's called `dep-chain`. The dependency to add is:

```
<dependency>
    <groupId>io.vertx</groupId>
    <artifactId>vertx-stack-depchain</artifactId>
    <version>3.8.0</version> <!-- must be replaced with the Vert.x version -->
    <type>pom</type>
</dependency>
```

gsmet: @cristianonicolai I let you update your PR? We have a few boms already defined there: https://github.com/quarkusio/quarkus/blob/master/bom/runtime/pom.xml#L171

cristianonicolai: sure, I can update my PR.

bom/runtime/pom.xml
```diff
@@ -213,6 +213,15 @@
...
             </dependency>
 
+            <!-- Vert.X dependencies, imported as a BOM -->
+            <dependency>
+                <groupId>io.vertx</groupId>
+                <artifactId>vertx-stack-depchain</artifactId>
+                <version>${vertx.version}</version>
+                <scope>import</scope>
+                <type>pom</type>
+            </dependency>
```
gsmet: @rsvoboda could you check that adding this BOM doesn't create dependency issues? Especially for Netty and Jackson as the Vert.x BOM also includes them.

Thanks!

rsvoboda: will do ;) later today or tomorrow

rsvoboda (commented):
I created few small projects, one depending on quarkus-jackson, second on quarkus-netty, third on quarkus-vertx and last one depending on all deps from https://github.com/vert-x3/vertx-stack/blob/3.7.1/stack-depchain/pom.xml.

Netty was resolved to version 4.1.34.Final in all relevant cases, Jackson databind resolved to 2.9.9.3.

vertx-stack-depchain BOM is added after netty-bom and jackson-bom in bom/runtime/pom.xml so I assume "correct" version should be resolved.

To check this idea I updated netty to  4.1.35.Final, built Quarkus and checked my sample apps. Updated version was present in all relevant dependency trees.

Looking at stack-depchain BOM makes me feel a bit uncertain as the **BOM defines 80 dependencies**.
 * Do we really want to import all of those 80 deps via Quarkus bom?
 * Will they work in native mode ?
 * What will be the support policy for these artifacts - e.g. CVE rating / fixes 

Other BOM imports like Jackson, RESTEasy, Netty, kubernetes-client-bom and JUnit are pulling a lot of stuff too, some of that stuff can be unwanted for Quarkus. Maybe a topic for wider discussion, especially if we think about 1.0 and post-1.0 era.

cescoffier:
> Do we really want to import all of those 80 deps via Quarkus bom?

These are mainly the modules from the Vert.x ecosystem. 

> Will they work in native mode?

Not all of them, as for the other dependencies included in the BOM file. 
To be honest, I don't know which one will work OOTB in native mode (for sure config and discovery won't)

> What will be the support policy for these artifacts - e.g. CVE rating / fixes

The Vert.x team does a great job to update in a timely fashion.

gsmet:
Well, in any case, we don't include the dependencies, we just ensure that they are all consistent.

Let's merge this, thanks @rsvoboda for checking the dependencies and thanks @cristianonicolai for the PR.

gsmet (approved):

PR #3529, opened 2019-08-15
remove redundant config item name and default value from KeycloakConfig

machi1990:


extensions/keycloak/deployment/src/main/java/io/quarkus/keycloak/KeycloakConfig.java
```diff
@@ -11,53 +11,53 @@
...
-    @ConfigItem(name = "realm")
+    @ConfigItem
     String realm;
 
     /**
-     * Name of the realm.
+     * Name of the realm key.
      */
     @ConfigItem(name = "realm-public-key")
     Optional<String> realmKey;
```
gsmet: Let's also try to be consistent here. Let's rename the property and remove the `name`.

extensions/keycloak/deployment/src/main/java/io/quarkus/keycloak/KeycloakConfig.java
```diff
@@ -11,53 +11,53 @@
...
 
     /**
      * The client-id of the application. Each application has a client-id that is used to identify the application
      */
```
gsmet: The docs seems a bit off here. Or the name of the property. Or maybe it's just me :). Could you check what this one is used for?

extensions/keycloak/deployment/src/main/java/io/quarkus/keycloak/KeycloakConfig.java
```diff
@@ -262,34 +262,34 @@
...
+    @ConfigItem
     boolean verifyTokenAudience;
 
     /**
      * If set to true will turn off processing of the access_token query parameter for bearer token processing.
      * Users will not be able to authenticate if they only pass in an access_token
      */
-    @ConfigItem(name = "ignore-oauth-query-parameter", defaultValue = "false")
+    @ConfigItem(name = "ignore-oauth-query-parameter")
     boolean ignoreOAuthQueryParameter;
```
gsmet: This one doesn't work and the property ends up being `o-auth`? I thought David had handled this case?

gsmet (changes requested):
I added a few questions and suggestions.

extensions/keycloak/deployment/src/main/java/io/quarkus/keycloak/KeycloakConfig.java
machi1990: Yes exactly. It ends up being `o-auth`. I think only `JBoss` is handled differently.

extensions/keycloak/deployment/src/main/java/io/quarkus/keycloak/KeycloakConfig.java
machi1990: Seems to be good. The doc is lifted from https://www.keycloak.org/docs/latest/securing_apps/index.html#_java_adapter_config

gsmet: OK. Not excited about it but I suppose it's better to be consistent with the JSON config file.

extensions/keycloak/deployment/src/main/java/io/quarkus/keycloak/KeycloakConfig.java
gsmet: BTW, I'm talking about renaming the *Java* property.

machi1990: Yes. While I was at it I renamed several others config items.

gsmet (approved):

gsmet:
Merged, thanks!

PR #3536, opened 2019-08-15
Added Resteasy integration tests

gastaldi:


integration-tests/pom.xml
```diff
@@ -52,6 +52,7 @@
         <module>neo4j</module>
         <module>mongodb-client</module>
         <module>resteasy-jackson</module>
+        <module>resteasy</module>
```
gsmet: Could we add these tests in the main ITs instead? We already have some RESTEasy tests there.

gsmet:
BTW, this comes a bit out of the blue. I suppose it's related to the `ServletContext` injection but having a description to the issue with some context would help.

I don't see a fix so I suppose the test is supposed to fail for now?

gastaldi:
Got it, I'll close this PR for now

PR #3540, opened 2019-08-16
Reinitialize ThreadLocalRandom class to ensure seed is reset at runtime

wburns:
The ThreadLocalRandom seeder and probeGenerator fields are never reinitialized, which causes the values generated to always be the same from native run to the next. See https://github.com/oracle/graal/issues/1614 for more details and hopefully the final fix. This PR will fix the issue until it is fixed at the GraalVM level.


gsmet (approved):

gsmet:
Thanks!

Issue #3202, opened 2019-07-12
Kafka Streams Extension can't find Serdes.StringSerde


muirandy:
**Describe the bug**
Native image generated cannot find the StringSerde. This causes the app to fail on startup.

**Expected behavior**
The app should start up and run without issues.

**Actual behavior**
I have a Java Kafka Streams app. I configure the Serdes.StringSerde as the default Serdes in the Kafka properties.
I then build a native image (either on OSX or docker).
When I run the native image, it fails with:
```
Caused by: org.apache.kafka.common.config.ConfigException: Invalid value org.apache.kafka.common.serialization.Serdes$StringSerde for configuration default.key.serde: Class org.apache.kafka.common.serialization.Serdes$StringSerde could not be found.
	at org.apache.kafka.common.config.ConfigDef.parseType(ConfigDef.java:718)
	at org.apache.kafka.common.config.ConfigDef.parseValue(ConfigDef.java:471)
	at org.apache.kafka.common.config.ConfigDef.parse(ConfigDef.java:464)
	at org.apache.kafka.common.config.AbstractConfig.<init>(AbstractConfig.java:62)
	at org.apache.kafka.streams.StreamsConfig.<init>(StreamsConfig.java:849)
	at org.apache.kafka.streams.StreamsConfig.<init>(StreamsConfig.java:844)
	at org.apache.kafka.streams.KafkaStreams.<init>(KafkaStreams.java:544)
	at com.aimyourtechnology.quarkus.kafka.streams.ConverterStream.runTopology(ConverterStream.java:52)
	at com.aimyourtechnology.quarkus.kafka.streams.StreamingApp.onStart(StreamingApp.java:54)
	at com.aimyourtechnology.quarkus.kafka.streams.StreamingApp_Observer_onStart_fd71b5e0b207b7d1ef838b94eaeff75e52b8f463.notify(StreamingApp_Observer_onStart_fd71b5e0b207b7d1ef838b94eaeff75e52b8f463.zig:63)
	at io.quarkus.arc.EventImpl$Notifier.notify(EventImpl.java:228)
	at io.quarkus.arc.EventImpl.fire(EventImpl.java:69)
	at io.quarkus.arc.runtime.LifecycleEventRunner.fireStartupEvent(LifecycleEventRunner.java:23)
	at io.quarkus.arc.runtime.ArcDeploymentTemplate.handleLifecycleEvents(ArcDeploymentTemplate.java:99)
[...]

PR #3547, opened 2019-08-16
Kafka Streams - pick up default Serdes and ExceptionHandler from props

muirandy:


extensions/kafka-streams/deployment/src/main/java/io/quarkus/kafka/streams/deployment/KafkaStreamsProcessor.java
```diff
@@ -66,20 +62,89 @@ void build(RecorderContext recorder,
...
+        if (allDefaultSerdesAreNotDefinedInProperties(defaultKeySerdeClass, defaultValueSerdeClass)) {
+            registerDefaultSerde(reflectiveClasses);
+        }
+    }
+
+    private void registerClassName(BuildProducer<ReflectiveClassBuildItem> reflectiveClasses, String defaultKeySerdeClass) {
+        reflectiveClasses.produce(new ReflectiveClassBuildItem(true, false, false, defaultKeySerdeClass));
+    }
+
+    private boolean allDefaultSerdesAreNotDefinedInProperties(String defaultKeySerdeClass, String defaultValueSerdeClass) {
```
gunnarmorling: Minor nit: I'd avoid the negated method name and instead invert it, so it's used like `if (!allDefaultSerdesDefinedInProperties) { ... }`. "Positive" method names make things more comprehensible IMO.

extensions/kafka-streams/deployment/src/main/java/io/quarkus/kafka/streams/deployment/KafkaStreamsProcessor.java
```diff
@@ -66,20 +62,89 @@ void build(RecorderContext recorder,
...
     @BuildStep
     @Record(ExecutionTime.STATIC_INIT)
     BeanContainerListenerBuildItem processBuildTimeConfig(KafkaStreamsRecorder recorder) {
-        Config config = ConfigProvider.getConfig();
+        Properties kafkaStreamsProperties = buildKafkaStreamsProperties();
+        return new BeanContainerListenerBuildItem(recorder.configure(kafkaStreamsProperties));
+    }
 
-        Properties properties = new Properties();
+    private Properties buildKafkaStreamsProperties() {
```
gunnarmorling: This one is called three times overall during the build phase. This might be avoided?

extensions/kafka-streams/deployment/src/main/java/io/quarkus/kafka/streams/deployment/KafkaStreamsProcessor.java
```diff
@@ -66,20 +62,89 @@ void build(RecorderContext recorder,
...
+
+    private void registerClassName(BuildProducer<ReflectiveClassBuildItem> reflectiveClasses, String defaultKeySerdeClass) {
+        reflectiveClasses.produce(new ReflectiveClassBuildItem(true, false, false, defaultKeySerdeClass));
+    }
+
+    private boolean allDefaultSerdesAreNotDefinedInProperties(String defaultKeySerdeClass, String defaultValueSerdeClass) {
+        return defaultKeySerdeClass == null || defaultValueSerdeClass == null;
+    }
+
+    private void registerDefaultSerde(BuildProducer<ReflectiveClassBuildItem> reflectiveClasses) {
```
gunnarmorling: Perhaps have a more generic `register()` method to be used everywhere:

    register(reflectiveClasses, ByteArraySerde.class);

gunnarmorling (commented):
LGTM, @muirandy! Some minor stylistic comments inline, if you'd like to apply them. Thanks!

extensions/kafka-streams/deployment/src/main/java/io/quarkus/kafka/streams/deployment/KafkaStreamsProcessor.java
```diff
@@ -66,20 +62,89 @@ void build(RecorderContext recorder,
...
+
+    private void registerDefaultSerdes(BuildProducer<ReflectiveClassBuildItem> reflectiveClasses) {
+        Properties properties = buildKafkaStreamsProperties();
+        String defaultKeySerdeClass = properties.getProperty(StreamsConfig.DEFAULT_KEY_SERDE_CLASS_CONFIG);
+        String defaultValueSerdeClass = properties.getProperty(StreamsConfig.DEFAULT_VALUE_SERDE_CLASS_CONFIG);
+
+        if (defaultKeySerdeClass != null)
+            registerClassName(reflectiveClasses, defaultKeySerdeClass);
+        if (defaultValueSerdeClass != null)
+            registerClassName(reflectiveClasses, defaultValueSerdeClass);
```
gsmet: Please use curly braces.

extensions/kafka-streams/deployment/src/main/java/io/quarkus/kafka/streams/deployment/KafkaStreamsProcessor.java
```diff
@@ -66,20 +62,89 @@ void build(RecorderContext recorder,
...
+    }
+
+    private void registerDefaultExceptionHandler(BuildProducer<ReflectiveClassBuildItem> reflectiveClasses) {
+        reflectiveClasses.produce(new ReflectiveClassBuildItem(true, false, false, LogAndFailExceptionHandler.class));
+    }
+
+    private void registerDefaultSerdes(BuildProducer<ReflectiveClassBuildItem> reflectiveClasses) {
+        Properties properties = buildKafkaStreamsProperties();
+        String defaultKeySerdeClass = properties.getProperty(StreamsConfig.DEFAULT_KEY_SERDE_CLASS_CONFIG);
+        String defaultValueSerdeClass = properties.getProperty(StreamsConfig.DEFAULT_VALUE_SERDE_CLASS_CONFIG);
```
gsmet: We all agree that this configuration is fixed at build time and cannot be changed later? Otherwise, it won't work that well.

gsmet (dismissed):
Added one minor comment and one question.

extensions/kafka-streams/deployment/src/main/java/io/quarkus/kafka/streams/deployment/KafkaStreamsProcessor.java
gunnarmorling: Yes, at least that's how it currently is built -- the majority of properties (everything in this `Properties` is fixed at build time.

extensions/kafka-streams/deployment/src/main/java/io/quarkus/kafka/streams/deployment/KafkaStreamsProcessor.java
muirandy: Happy to change, though I prefer to avoid the "!" as much as possible.

extensions/kafka-streams/deployment/src/main/java/io/quarkus/kafka/streams/deployment/KafkaStreamsProcessor.java
muirandy: Does this have a performance impact? 
I imagine the alternative would be to cache the properties in a field. Is this approach preferred?

extensions/kafka-streams/deployment/src/main/java/io/quarkus/kafka/streams/deployment/KafkaStreamsProcessor.java
muirandy: Sure!

extensions/kafka-streams/deployment/src/main/java/io/quarkus/kafka/streams/deployment/KafkaStreamsProcessor.java
muirandy: Sure. 
I was looking for some coding style settings for IntelliJ. Would it be possible to have a file in the `ide-config` directory?

extensions/kafka-streams/deployment/src/main/java/io/quarkus/kafka/streams/deployment/KafkaStreamsProcessor.java
muirandy: I'm getting to grips with things still. I defer to @gunnarmorling and @gsmet !

muirandy:
We don't seem to have any tests for the Kafka Streams extension. Am I missing something? Is it feasible to add some?

gunnarmorling:
> We don't seem to have any tests for the Kafka Streams extension

There's an integration test [here](https://github.com/quarkusio/quarkus/blob/master/integration-tests/kafka/src/test/java/io/quarkus/it/kafka/streams/KafkaStreamsTest.java).

extensions/kafka-streams/deployment/src/main/java/io/quarkus/kafka/streams/deployment/KafkaStreamsProcessor.java
gunnarmorling: I'd suggest to just build the properties once and pass them down as a parameter to all the methods needing it. This will require to merge the two static init methods that currently require it into one.

muirandy: Ok.... for that, I need to understand where the `KafkaStreamsRecorder` comes from which is passed in to `processBuildTimeConfig`. I tried adding it in as an argument to the `build` method, and it worked. I guess there is some DI involved somewhere?

The other side effect is, obviously, that we're adding another argument, which IMHO is never good.   Would there be any harm, in your opinion, to using fields? The `reflectiveClasses` variable could also benefit from being a field. I'm not familiar with how the class is used.

gunnarmorling: > I tried adding it in as an argument to the build method, and it worked. I guess there is some DI involved somewhere?

Yes, Quarkus will pass all the required parameters for the build step methods automatically.

> Would there be any harm, in your opinion, to using fields?

I don't know tbh., I'm not sure about the lifecycle of these classes (perhaps @gsmet can comment); in any case just a parameter seemed saver and the easiest to me.

gsmet: yes, use a parameter.

muirandy: Thanks. OK - so unless I hear that fields are safe to use, I'll use arguments.
Merging the 2 static init methods nicely is now leading me to extract some methods to indicate whats going on. Can you tell me what these 2 lines from the `build` method do?:
`feature.produce(new FeatureBuildItem(FeatureBuildItem.KAFKA_STREAMS));`
`jni.produce(new JniBuildItem());`

gunnarmorling: This enable JNI for the native build. This is needed for RocksDB which loads native libraries via JNI.

muirandy: Thanks!

muirandy:
I've pushed the changes I've made following the above comments.

extensions/kafka-streams/deployment/src/main/java/io/quarkus/kafka/streams/deployment/KafkaStreamsProcessor.java
muirandy: @gunnarmorling, @gsmet  So, the build failed. After digging about a bit it appears that (at least for the tests) I can't add the `KafkaStreamsRecorder` in as an extra parameter to `build` (even though it worked in practice!):
```
[ERROR] Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 8.403 s <<< FAILURE! - in io.quarkus.it.kafka.streams.KafkaStreamsTest
[ERROR] testKafkaStreams  Time elapsed: 0.007 s  <<< ERROR!
org.junit.jupiter.api.extension.TestInstantiationException: TestInstanceFactory [io.quarkus.test.junit.QuarkusTestExtension] failed to instantiate test class [io.quarkus.it.kafka.streams.KafkaStreamsTest]: Failed to start quarkus

...

        at java.util.Hashtable.putAll(Hashtable.java:523)
        at io.quarkus.kafka.streams.runtime.KafkaStreamsTopologyManager.getStreamsProperties(KafkaStreamsTopologyManager.java:80)
        at io.quarkus.kafka.streams.runtime.KafkaStreamsTopologyManager.onStart(KafkaStreamsTopologyManager.java:106)
...
```

I can't find anything in the test that seems related, and I can't run it from the IDE as there are a bunch of compilation issues related to `org.apache.maven.wagon.Wagon`.

The implications are that the `processBuildTimeConfig` method can't be changed from being a static init method. From there, the kafkaStreamProperties can't be shared. 

Do you have any suggestions? Do we stick to calling `buildKafkaStreamsProperties` 3 times, or am I missing something? Apologies for taking so much of your time folks - its taking me longer than I anticipated too!

muirandy: I've pushed code that passes the test, but calls `buildKafkaStreamsProperties ` 3 times. Hope its good.

muirandy:
I can't see why the build is still red. I suspect that re-running it will let it go green. All good on my laptop. I can't seem to kick of a re-run of the build though!

gsmet:
The Windows failure is unrelated: we have an instability right now.

Could you squash all your commits into one? Thanks!

muirandy:
OK - I've pushed a squashed commit.

cescoffier:
Changes have been made.

cescoffier (approved):

machi1990:
Fixes #3202

PR #3549, opened 2019-08-16
Infinispan client updates

wburns:
1. Update Infinispan Version
2. Add in automatic SSL native support


wburns:
This is first step. I have some enhancements I need to do the extension that should be coming next week. And after that I should be able to add in the first version of the embedded extension of Infinispan embedded.

wburns:
SSL was working just fine without the pom.xml changes to force JNI or security.

The SASL mechs were working fine without elytron dependencies as well when I tested manually. I need to talk with Tristan at some point to verify what needs to be done there if anything.

gsmet (approved):

gsmet:
Merged, thanks @wburns !

Issue #3552, opened 2019-08-17
[gradle] gradle test still not work


pipinet:
### console error log
```bash
$./gradlew test

> Task :test FAILED

quarkus.modern.gradle.example.UserResourceTest > testHelloEndpoint() FAILED
    org.junit.jupiter.api.extension.TestInstantiationException
        Caused by: java.lang.IllegalStateException
            Caused by: io.quarkus.bootstrap.BootstrapException

1 test completed, 1 failed

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':test'.
> There were failing tests. See the report at: file:///Users/pipinet/qwlabs/quarkus-issues/2019-06-21-21-00/build/reports/tests/test/index.html

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Get more help at https://help.gradle.org

BUILD FAILED in 2s
4 actionable tasks: 1 executed, 3 up-to-date
```
### test report error log
```bash
org.junit.jupiter.api.extension.TestInstantiationException: TestInstanceFactory [io.quarkus.test.junit.QuarkusTestExtension] failed to instantiate test class [quarkus.modern.gradle.example.UserResourceTest]: Failed to create the boostrap class loader
	at org.junit.jupiter.engine.descriptor.ClassBasedTestDescriptor.invokeTestInstanceFactory(ClassBasedTestDescriptor.java:298)
	at org.junit.jupiter.engine.descriptor.ClassBasedTestDescriptor.instantiateTestClass(ClassBasedTestDescriptor.java:273)
	at org.junit.jupiter.engine.descriptor.ClassTestDescriptor.instantiateTestClass(ClassTestDescriptor.java:70)
	at org.junit.jupiter.engine.descriptor.ClassBasedTestDescriptor.instantiateAndPostProcessTestInstance(ClassBasedTestDescriptor.java:256)
	at org.junit.jupiter.engine.descriptor.ClassBasedTestDescriptor.lambda$testInstancesProvider$2(ClassBasedTestDescriptor.java:250)
	at java.util.Optional.orElseGet(Optional.java:267)
	at org.junit.jupiter.engine.descriptor.ClassBasedTestDescriptor.lambda$testInstancesProvider$3(ClassBasedTestDescriptor.java:249)
	at org.junit.jupiter.engine.execution.TestInstancesProvider.getTestInstances(TestInstancesProvider.java:29)
[...]

PR #3555, opened 2019-08-17
Start using Gradle Tooling API

### LINKED ISSUE #3497: Use gradle model instead of string parsing for manipulating gradle files
**Description**
Actually we use string parsing and manipulation to add extension, complete existing gradle file, or simply read gradle files. It's fragil, On maven side, we use an external model and parser. 

**Implementation ideas**
I am just dig a little more to not use string and delegate parse of model to gradle lib.
I see how eclipse do it:
https://github.com/eclipse/buildship/blob/master/org.eclipse.buildship.kotlin/build.gradle
https://github.com/eclipse/buildship/blob/master/org.eclipse.buildship.kotlin/src/main/java/org/eclipse/buildship/kotlin/internal/GradleKotlinScriptTemplateProvider.java#L80

And with ProjectConnection we can load the groovy model too, but I don't know the limitation and we need all gradle core as dependency. 

see https://github.com/quarkusio/quarkus/pull/3475#discussion_r313254631

Dufgui:
fix #3497


Dufgui:
normally ready for review @geoand @aloubyansky

geoand:
Tests seem to be failing:

```
[ERROR] Tests run: 8, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 2.19 s <<< FAILURE! - in io.quarkus.cli.commands.CreateProjectTest
[ERROR] createGradle  Time elapsed: 1.861 s  <<< ERROR!
org.gradle.tooling.BuildException: Could not fetch model of type 'EclipseProject' using Gradle distribution 'https://services.gradle.org/distributions/gradle-5.5.1-bin.zip'.
```

Dufgui:
I think it's now fixed, the remaining error seems to be an instability from the test job (tomeout) @geoand @aloubyansky

gsmet:
It might be a stupid question but why do we have Gradle specific classes in `devtools/common`?

Not exactly related to your patch but adding a dependency to Gradle + the Gradle Maven repo there seems a bit of a problem to me.

Dufgui:
We already have a dependence on model from maven. And I want to produce the
same feature in gradle and maven.


Le lun. 19 août 2019 à 11:56, Guillaume Smet <notifications@github.com> a
écrit :

Dufgui:
But I can do a refactoring to remove maven and gradle dep from their ?

Le lun. 19 août 2019 à 12:26, Guillaume Dufour <guillaume.duff@gmail.com> a
écrit :

Dufgui:
Or you want to keep maven as core. And overwrite only gradle version ?

Dufgui:
I have the same problem in bootstrap core. With #3552   What do you prefere @gsmet

Dufgui:
so, what could anyone, relaunch the tests and review please ?
@gsmet @aloubyansky @geoand  ?

devtools/common/src/test/resources/gradle-project/gradle.properties
```diff
@@ -1 +1 @@
-quarkusVersion = 999-SNAPSHOOT
\ No newline at end of file
+quarkusVersion = 0.21.1
```
geoand: This is probably not a good question and I am assuming that I am missing something fundamental here, but how would this work with the current code if the test uses a specific version?

Dufgui:
You are right. I must migrate this test to gradle plugin.

Dufgui:
gradle dep removed from common and put in gradle project.
So the maven mojo cannot list extension version of a gradle project now.

Dufgui:
@aloubyansky it's works !!!!! you can review

devtools/gradle/src/test/java/io/quarkus/gradle/GradleBuildFileTest.java
```diff
@@ -0,0 +1,88 @@
...
+
+    private static GradleBuildFileFromConnector buildFile;
+
+    @BeforeAll
+    public static void beforeAll() throws IOException, URISyntaxException {
+        URL url = GradleBuildFileTest.class.getClassLoader().getResource("gradle-project");
+        URI uri = url.toURI();
+        Path gradleProjectPath;
+        if (uri.getScheme().equals("jar")) {
+            FileSystem fileSystem = FileSystems.newFileSystem(uri, Collections.<String, Object> emptyMap());
```
aloubyansky: Normally this should be a directory, right? When would that be found in a JAR? FileSystem should be closed then.

aloubyansky (commented):
Well done, @Dufgui! I added a question there.

devtools/gradle/src/test/java/io/quarkus/gradle/GradleBuildFileTest.java
Dufgui: you are right

aloubyansky (approved):
I'm approving this PR but I also see a few things that need to be fixed as a follow up (they are not introduced by this specific PR). It also doesn't necessarily mean they all fall on your shoulders @Dufgui :) If you like to continue improving this area - great, if not, that's fine too, your contributions are always appreciated.
Here is what I see needs to be improved:
* proper logger (there are `System.out.println` calls in various places, there is also BuildFile.PRINTER, if we can't use jboss-logging directly then we should use BuildFile.PRINTER or another abstraction everywhere);
* we should look into replacing mentions of `999-SNAPSHOT` in gradle build scripts with `${version}` that's set in one place like `gradle.properties`.

Actually the bigger task should be to move the gradle plugin into a separate repo and be a pure Gradle project. I think that's the requirement to be accepted into the Gradle plugin repo.

kenfinnigan:
Is this good to be merged?

PR #3556, opened 2019-08-18
Add CORS Access-Control-Max-Age response header to cache preflight request

machi1990:
This PR adds a cors `Access-Control-Max-Age` response header indicating the cache duration  of a preflight request.

### DIFF
--- a/extensions/vertx-web/runtime/src/main/java/io/quarkus/vertx/web/runtime/cors/CORSConfig.java
+++ b/extensions/vertx-web/runtime/src/main/java/io/quarkus/vertx/web/runtime/cors/CORSConfig.java
@@ -1,9 +1,12 @@
 package io.quarkus.vertx.web.runtime.cors;
 
+import java.time.Duration;
+import java.util.List;
 import java.util.Optional;
 
 import io.quarkus.runtime.annotations.ConfigGroup;
 import io.quarkus.runtime.annotations.ConfigItem;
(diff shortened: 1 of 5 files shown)


extensions/vertx-web/runtime/src/main/java/io/quarkus/vertx/web/runtime/cors/CORSConfig.java
```diff
@@ -37,19 +42,32 @@
...
     @ConfigItem
-    public Optional<String> exposedHeaders;
+    public List<String> exposedHeaders;
+
+    /**
+     * The `Access-Control-Max-Age` response header value indicating
+     * how long the results of a pre-flight request can be cached.
+     */
+    @ConfigItem
+    public Optional<Duration> accessControlMaxAge;
```
gsmet: You don't want to include it in the `toString()`?

gsmet (changes requested):
Looks good to me. I added a very minor comment, could you check it out?

extensions/vertx-web/runtime/src/main/java/io/quarkus/vertx/web/runtime/cors/CORSConfig.java
machi1990: Well spotted. I have added it.

gsmet (approved):

PR #3557, opened 2019-08-18
Register enclosing class for reflection when used as JAX-RS return type

### LINKED ISSUE #3538: quarkus-resteasy-jsonb Marshaling Generic Type results in WARNINGS on each request
Using quarkus-resteasy-jsonb for my RESTful service

and with this class
```
public class MyEnvelopeClass<T> {
    private T content;

    public MyEnvelopeClass(T content) {
        this.content = content;
    }

    public T getContent() {
        return content;
    }
}
```

Returned from this resource method:
```
    @GET
    @Path("/{id}")
    public MyEnvelopeClass<MyPayloadClass> getById(@PathParam("id") String id){
        return new MyEnvelopeClass<MyPayloadClass>(new MyPayloadClass());
    }

```
The result from my GET request returns the correct payload, however,  I get this warning in the log each time the request is made.

```
2019-08-15 12:25:36,477 WARNING [org.ecl.yas.int.ReflectionUtils] (executor-thread-1) Generic bound not found for type T declared in class MyEnvelopeClass.
```

If I switch to quarkus-resteasy-jackson, the warning is not logged. ( however I have to include other dependencies to get "Instant" to serialize properly, and configure the ObjectMapper to use it... Then I end up with Maven warnings too... so I would prefer to stick with the jsonb dependency )

If I change my code to use an anonymous inner class, extending MyEnvelopeClass with the Generic type... the warning is not logged:
```
    @GET
    @Path("/{id}")
    public MyEnvelopeClass<MyPayloadClass> getById(@PathParam("id") String id){
        return new MyEnvelopeClass<MyPayloadClass>(new MyPayloadClass()){};
    }

```
I can use this as a workaround, but I would rather fix the issue holistically rather than within every single Resource method we write.

Thanks!

-sean

Darwin USMACEA013SSCOT 18.7.0 Darwin Kernel Version 18.7.0: Thu Jun 20 18:42:21 PDT 2019; root:xnu-4903.270.47~4/RELEASE_X86_64 x86_64

 - Output of `java -version`: 
openjdk version "11.0.3" 2019-04-16
OpenJDK Runtime Environment AdoptOpenJDK (build 11.0.3+7)
OpenJDK 64-Bit Server VM AdoptOpenJDK (build 11.0.3+7, mixed mode)

openjdk version "1.8.0_222"
OpenJDK Runtime Environment (build 1.8.0_222-20190711112007.graal.jdk8u-src-tar-gz-b08)
OpenJDK 64-Bit GraalVM CE 19.1.1 (build 25.222-b08-jvmci-19.1-b01, mixed mode)

 - Quarkus version or git rev: 
    <quarkus.version>0.20.0</quarkus.version>

geoand:
This is needed when a JAX-RS resource returns a parameterized type that
is meant to be an enclosing class

Fixes: #3538

### DIFF
--- a/core/deployment/src/main/java/io/quarkus/deployment/steps/ReflectiveHierarchyStep.java
+++ b/core/deployment/src/main/java/io/quarkus/deployment/steps/ReflectiveHierarchyStep.java
@@ -1,6 +1,7 @@
 package io.quarkus.deployment.steps;
 
 import java.lang.reflect.Modifier;
+import java.util.Arrays;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
@@ -32,6 +33,9 @@ import io.quarkus.deployment.builditem.substrate.ReflectiveHierarchyBuildItem;
 public class ReflectiveHierarchyStep {
 
     private static final Logger log = Logger.getLogger(ReflectiveHierarchyStep.class);
+
+    private static final Set<String> IGNORED_PARAMETERIZED_TYPE_PACKAGE_NAMES = new HashSet<>(
+            Arrays.asList("java.util", "io.reactivex"));
 
     @Inject
     List<ReflectiveHierarchyBuildItem> hierarchy;
@@ -83,7 +87,9 @@ public class ReflectiveHierarchyStep {
             addReflectiveHierarchy(i, type.asArrayType().component(), processedReflectiveHierarchies, unindexedClasses);
         } else if (type instanceof ParameterizedType) {
             ParameterizedType p = (ParameterizedType) type;
-            addReflectiveHierarchy(i, p.owner(), processedReflectiveHierarchies, unindexedClasses);
+            if (!IGNORED_PARAMETERIZED_TYPE_PACKAGE_NAMES.contains(p.name().toString())) {
+                addClassTypeHierarchy(i, p.name(), processedReflectiveHierarchies, unindexedClasses);
+            }
             for (Type arg : p.arguments()) {
                 addReflectiveHierarchy(i, arg, processedReflectiveHierarchies, unindexedClasses);
             }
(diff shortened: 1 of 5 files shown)


geoand:
@gsmet This probably isn't the absolute cleanest way to handle this case, but it's in line with what we already have.
One this I am not clear about is whether the `addReflectiveHierarchy(i, p.owner(), processedReflectiveHierarchies, unindexedClasses);` should be completely removed because `p.owner` always seems to be `null`...

core/deployment/src/main/java/io/quarkus/deployment/steps/ReflectiveHierarchyStep.java
```diff
@@ -84,6 +88,9 @@ private void addReflectiveHierarchy(ReflectiveHierarchyBuildItem i, Type type, S
         } else if (type instanceof ParameterizedType) {
             ParameterizedType p = (ParameterizedType) type;
             addReflectiveHierarchy(i, p.owner(), processedReflectiveHierarchies, unindexedClasses);
+            if (!IGNORED_PARAMETERIZED_TYPE_PACKAGE_NAMES.contains(p.name().toString())) {
+                addClassTypeHierarchy(i, p.name(), processedReflectiveHierarchies, unindexedClasses);
```
gsmet: I think that `p.owner()` was supposed to do what you just added. So if it doesn't do it, we can drop it.

geoand: That's what I thought as well but wanted to check. I'll update the PR later on

geoand:
@gsmet PR updated

gsmet (approved):

gsmet:
Merged, thanks!

geoand:
Sure thing!

PR #3564, opened 2019-08-19
Fix possible timeouts on devtools project generation tests

ia3andy:
- Change timeout to 2s
- Use only 4 threads to reduce and make the duration stabler
- Local tested duration is under 250ms

### DIFF
--- a/devtools/common/src/test/java/io/quarkus/cli/commands/CreateProjectTest.java
+++ b/devtools/common/src/test/java/io/quarkus/cli/commands/CreateProjectTest.java
@@ -219,10 +219,10 @@ public class CreateProjectTest {
     }
 
     @Test
-    @Timeout(1)
+    @Timeout(2)
     @DisplayName("Should create correctly multiple times in parallel with multiple threads")
     void createMultipleTimes() throws InterruptedException {
-        final ExecutorService executorService = Executors.newFixedThreadPool(10);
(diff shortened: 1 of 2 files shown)


FroMage (approved):
LGTM

gsmet:
The Windows failure is unrelated, merging.

devtools/common/src/test/java/io/quarkus/cli/commands/CreateProjectTest.java
```diff
@@ -243,6 +246,7 @@ void createMultipleTimes() throws InterruptedException {
         }).collect(Collectors.toList());
         executorService.invokeAll(collect);
         latch.await();
+        System.out.println(watch.toString());
```
gsmet: Hmmm, what's the purpose of this?

ia3andy: It logs the test duration. I thought it was not a problem to leave it here. should I remove it?

gsmet: I would say that I wouldn't import a Google class just for that. So I would remove it for now and revisit if the changes you made are not stable enough.

gsmet:
@ia3andy you broke the formatting again. Can you fix? Thanks!

PR #3580, opened 2019-08-19
Make java.time.Duration serializable to bytecode

gwenneg:
I used a workaround in #3394 to use `java.time.Duration` objects from a `Recorder` which is usually impossible because `Duration` doesn't have a zero-arg constructor and therefore can't be serialized to bytecode.

@gsmet [suggested](https://github.com/quarkusio/quarkus/pull/3394#discussion_r315190695) to try and make `Duration` serializable to bytecode by modifying `BytecodeRecorderImpl`, so here I am with a PR that gave positive results with the #3394 tests involving the `Recorder` mentioned above.

@stuartwdouglas Would you please review this PR?


gsmet (approved):
Stuart is on PTO so you will have to do with me :).

It looks good to me. Let's wait for CI.

PR #3587, opened 2019-08-19
Virtual HTTP and Azure Function Support

patriot1burke:
* Added virtual connection capabilities to Vertx web so that netty messages can be pumped in intra-VM.  This allows adapters like azure functions to bridge the azure runtime to resteasy, undertow, and vertx
* Added Azure Functions http extension.


extensions/vertx-web/runtime/src/main/java/io/quarkus/vertx/web/runtime/HttpConfiguration.java
```diff
@@ -65,6 +65,13 @@
     @ConfigItem
     public OptionalInt ioThreads;
 
+    /**
+     * If this is true then only a virtual channel will be set up for vertx web.
+     * We have this switch for testing purposes.
+     */
+    @ConfigItem(defaultValue = "false")
```
machi1990: defaultValue not needed. It will be automatically default to false.

docs/src/main/asciidoc/azure-functions-http-guide.adoc
```diff
@@ -0,0 +1,120 @@
+////
+This guide is maintained in the main Quarkus repository
+and pull requests should be submitted there:
+https://github.com/quarkusio/quarkus/tree/master/docs/src/main/asciidoc
+////
+
+include::./attributes.adoc[]
+= {project-name} - Using Resteasy, Undertow, or Vertx Web with Azure Functions (Serverless)
```
machi1990: use `Quarkus` instead of `{project-name}`

machi1990 (commented):
Really great PR, I just have two minor comments.

FroMage:
@gsmet is the windows failure really related? Aren't we getting a lot of unrelated Windows failures on CI these days?

docs/src/main/asciidoc/azure-functions-http-guide.adoc
```diff
@@ -0,0 +1,120 @@
...
+
+The `quarkus-azure-functions-http` extension allows you to write microservices with Resteasy (JAX-RS),
```
FroMage: You of all people should be able to settle this: all the resteasy docs say it's `RESTEasy`, yet you use `Resteasy` here. How should we call it in Quarkus?

docs/src/main/asciidoc/azure-functions-http-guide.adoc
```diff
@@ -0,0 +1,120 @@
...
+Undertow (servlet), or Vert.x Web and make these microservices deployable to the Azure Functions runtime.
+
+One azure function can represent any number of JAX-RS, servlet, or Vertx Web endpoints.
```
FroMage: Similarly, it should be `Vert.x`.

FroMage (approved):
The strategy looks good to me. The PR too. I'm not qualified to review the netty virtual low-level code, though. Perhaps @vietj or @stuartwdouglas would be better for that part?

We can always merge this as-is and make corrections later.

gsmet:
The common Windows failure we have is related to the new HTTP layer but not to this specific PR.

docs/src/main/asciidoc/azure-functions-http-guide.adoc
gsmet: It's a bit more than that: the title needs to be just after the end of the comment and the includes after the title. Otherwise, the website doesn't take it. Thus the variable cannot be used as it's declared after the title itself.

docs/src/main/asciidoc/azure-functions-http-guide.adoc
patriot1burke: I ended up never liking RESTEasy.  Looks like a typo.

geoand:
The PR now needs to be rebased.

Issue #3585, opened 2019-08-19
Create the test structure when using `create-extension` Maven Mojo.


danielpetisme:
**Description**
When initializing an extension with the `create-extension` Maven Mojo it would be neat to create the `src/main/java` folders, plus create a blank test.

**Implementation ideas**
(If you have any implementation ideas, they can go here, however please note that all design change proposals should be posted to [the Quarkus developer mailing list](mailto:quarkus-dev@googlegroups.com) (or [the corresponding Google Group](https://groups.google.com/group/quarkus-dev); see [the decisions process document](https://github.com/quarkusio/quarkus/blob/master/DECISIONS.adoc) for more information).

The folder and code generation works already so it's not a big deal to create the structure. The point is what test should be generate. IMHO, we should only add a test in the extensions's `deployment` module.

When generating an extension `my-extension` for instance , the mojo creates a `MyExtensionProcessor`. We could generate a `MyExtensionTest` test class with a basic

```
@RegisterExtension
    static final QuarkusUnitTest config = new QuarkusUnitTest()
            .setArchiveProducer(() -> ShrinkWrap.create(JavaArchive.class));
```

Given the content of the generated `Processor` the only thing we could test out of the box would be the presence of the extension in `QuarkusUnitTest`

gwenneg:
That's a great idea! I would also add a blank dev mode / hot reload test:

```
@RegisterExtension
    final static QuarkusDevModeTest test = new QuarkusDevModeTest()
            .setArchiveProducer(() -> ShrinkWrap.create(JavaArchive.class));
```

gsmet:
@machi1990 @gwenneg this one looks like a quick win? Would one of you be interested in tackling this so we can close the issue?

machi1990:
Let me have a look.

PR #3595, opened 2019-08-20
Add default tests when creating an extension with Maven Mojo

danielpetisme:
Here is a proposal for #3585 

The only basic task we could do is ensure the extension is properly loaded (by scanning the startup output or programatically.

WDYT?

@gwenneg taked also to test hot reloading, I'm not sure to understand the goal.

### DIFF
--- a/devtools/maven/src/test/resources/expected/create-extension-pom-minimal/minimal-extension/deployment/src/test/java/org/acme/my/project/minimal/extension/deployment/MinimalExtensionProcessorTest.java
+++ b/devtools/maven/src/test/resources/expected/create-extension-pom-minimal/minimal-extension/deployment/src/test/java/org/acme/my/project/minimal/extension/deployment/MinimalExtensionProcessorTest.java
@@ -1,0 +1,20 @@
+package org.acme.my.project.minimal.extension.deployment;
+
+import org.jboss.shrinkwrap.api.ShrinkWrap;
+import org.jboss.shrinkwrap.api.spec.JavaArchive;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.RegisterExtension;
+
+import io.quarkus.test.QuarkusUnitTest;
+
+class MinimalExtensionProcessorTest {
+
+    @RegisterExtension
+    static final QuarkusUnitTest config = new QuarkusUnitTest()
+        .setArchiveProducer(() -> ShrinkWrap.create(JavaArchive.class));
+
+    @Test
+    public void should() {
+
+    }
+}
--- a/devtools/maven/src/main/java/io/quarkus/maven/CreateExtensionMojo.java
+++ b/devtools/maven/src/main/java/io/quarkus/maven/CreateExtensionMojo.java
@@ -444,6 +444,12 @@ public class CreateExtensionMojo extends AbstractMojo {
                         + "/deployment/" + model.artifactIdBaseCamelCase + "Processor.java");
         evalTemplate(cfg, "Processor.java", processorPath, charset, model);
 
+        final Path processorTestPath = basedir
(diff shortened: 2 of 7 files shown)


gwenneg:
You'll find more info about hot reload tests here: https://quarkus.io/guides/extension-authors-guide#testing-hot-reload

This was introducer in Quarkus `0.21.0` and should be used in all extensions.

danielpetisme:
Thanks for the quick reply. I missed that point.

Open question, can we consider that the majority of the extensions will require HTTP and generate an opinionated  `Processor` which produce a Servlet + the corresponding `ProcessorTest` which will start the app in dev mode, verify the HTTP endpoint and verify hot reload.

It implies an opinionated vision but have the benefits to drive new extension development. That means **undertow-extension** and **Restassured** added by default in all the created extension.

Thoughts?

gsmet:
I don't know, I wouldn't clutter the template too much, tbh.

danielpetisme:
Me too...
I would like to provide a consistent experience between Quarkus app and extension generation:
- When generating an extension I should be able to run a smoke test right away

-  Enforce best practices (Like testing hot reloading for instance)

- Be self-documented/intuitive, the `FeatureBuildItem` is easy to understand and a simple test would be also helpful. 

I admit these objectives are to simple, there is a trade off to find.

Shall we only limit the current PR to the structure creation as is and complete the generation later on with more developer feedbacks?

devtools/maven/src/test/resources/expected/create-extension-pom-minimal/minimal-extension/deployment/src/test/java/org/acme/my/project/minimal/extension/deployment/MinimalExtensionProcessorTest.java
```diff
@@ -0,0 +1,20 @@
...
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.RegisterExtension;
+
+import io.quarkus.test.QuarkusUnitTest;
+
+class MinimalExtensionProcessorTest {
+
+    @RegisterExtension
+    static final QuarkusUnitTest config = new QuarkusUnitTest()
+        .setArchiveProducer(() -> ShrinkWrap.create(JavaArchive.class));
```
cescoffier: Can you add comments to explain what the user should do in this code?

cescoffier (changes requested):
The templates should be kept minimal. So they are good right now.
However, generating comments so the user knows where hacking is required may significantly improve the experience.

cescoffier:
@danielpetisme also, you would need to rebase your PR.

danielpetisme:
WDYT of [these comments].(https://github.com/quarkusio/quarkus/blob/3e08ec268c23bbe68555aaed252c0462c2b9672b/devtools/maven/src/main/resources/create-extension-templates/ProcessorTest.java)

danielpetisme:
@cescoffier rebase done.

machi1990:
@danielpetisme hi, we have tests failing, since the Azure artifacts expired I cannot tell if it is related or not. Do you mind having a look?

cescoffier:
@danielpetisme ping? ^^^

danielpetisme:
Looks like @machi1990 has been faster cf. #5438
Feel free to close it if it makes sense.
Sorry for the delay

machi1990:
Ah @danielpetisme I did not see this initial PR at all, otherwise I would not have opened my PR. Sorry about that. You feedbacks/comments are however welcomed on the PR #5438 I opened.

danielpetisme:
No problem. I prefer to see the issue to be tackled !!!

machi1990:
Superseded by #5438

PR #3597, opened 2019-08-20
Remove containerRuntime default value from Gradle plugin

### LINKED ISSUE #2063: QuarkusNative now depends on Docker while building from Gradle?
This is kinda of a question more than a bug report or something.
I was able to successfully try to use quarkus `0.13.1` to build a sample application, but what I found strange is that even though I have set (as I did in `0.12.0`) the `GRAALVM_HOME` variable, now it is using docker to build the application.

Is there an option to disable the need of using docker to build the native binary?

gwenneg:
Fixes #2063 

The `containerRuntime = "docker"` default value was forcing `NativeImagePhase` to contact the Docker daemon at the beginning of the build even when the `--docker-build` Gradle option was not set (and therefore `false`).

That default value was also absent from `quarkus-maven-plugin`. Now, both plugins have the same behavior.

### DIFF
--- a/devtools/gradle/src/main/java/io/quarkus/gradle/tasks/QuarkusNative.java
+++ b/devtools/gradle/src/main/java/io/quarkus/gradle/tasks/QuarkusNative.java
@@ -56,7 +56,7 @@ public class QuarkusNative extends QuarkusTask {
 
     private String nativeImageXmx;
 
-    private String containerRuntime = "docker";
+    private String containerRuntime;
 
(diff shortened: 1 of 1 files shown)


gsmet (approved):
Looks good, thanks. Let's wait for CI - well, we don't have ITs for the Gradle support but let's do as there were some ;).

gsmet:
OMG, a CI build not failing \o/

PR #3603, opened 2019-08-20
Support IndexDependencyBuildItem

wburns:
Adds a new BuildItem that can be used by extensions to define what artifact should be also indexed that currently doesn't have a Jandex index included or doesn't have a marker to find. This also consolidates the `application.properties` `index-dependency` setting to create the build items.


gsmet (approved):
Looks good to me. 

Can’t say I’m a big fan of those three letters variables though :)

Let’s wait for CI

wburns:
> Can’t say I’m a big fan of those three letters variables though :)

Haha, I was kinda debating about that as well. I can change the names to make it more friendly :)

wburns:
Renamed the variables.

Issue #3605, opened 2019-08-20
Vert.x - No ReactiveStreamsFactory implementation found!


p-zalejko:
I started playing with Reactive Streams and ran into a problem with a missing implementation of the `ReactiveStreamsFactory`.

**To Reproduce**
Follow this tutorial https://quarkus.io/guides/using-vertx :
1. Create a hello world application
2. Add `vertx` extension
3. Implement a simple endpoint (using the Axle provider)
4. Build a native app

My pom.xml dependencies looked as follows:

```
<dependency>
  <groupId>io.quarkus</groupId>
  <artifactId>quarkus-resteasy</artifactId>
</dependency>
<dependency>
  <groupId>org.jetbrains.kotlin</groupId>
  <artifactId>kotlin-stdlib-jdk8</artifactId>
</dependency>
<dependency>
  <groupId>io.quarkus</groupId>
  <artifactId>quarkus-junit5</artifactId>
<scope>test</scope>
</dependency>
<dependency>
  <groupId>io.rest-assured</groupId>
  <artifactId>rest-assured</artifactId>
  <scope>test</scope>
</dependency>
<dependency>
  <groupId>io.quarkus</groupId>
  <artifactId>quarkus-vertx</artifactId>
</dependency>
```
The endpoint as this:
```

import io.vertx.axle.core.Vertx
import org.reactivestreams.Publisher
import javax.inject.Inject
import javax.ws.rs.GET
import javax.ws.rs.Path;
import javax.ws.rs.Produces
import javax.ws.rs.core.MediaType


@Path("/events")
class EventController {

  @Inject
  lateinit var vertx: Vertx

  @GET
  @Produces(MediaType.SERVER_SENT_EVENTS)
  fun getAll(): Publisher<String> {
    val toPublisherBuilder = vertx.periodicStream(200).toPublisherBuilder()
    return toPublisherBuilder
           .map { "hello" }
           .buildRs()
  }
}
```
**Issue**
After launching the application I got the following exception:
```
Caused by: java.lang.IllegalStateException: No ReactiveStreamsFactory implementation found!
at org.eclipse.microprofile.reactive.streams.operators.spi.ReactiveStreamsFactoryResolver.instance(ReactiveStreamsFactoryResolver.java:64)
at org.eclipse.microprofile.reactive.streams.operators.ReactiveStreams.fromPublisher(ReactiveStreams.java:60)
[...]

PR #3611, opened 2019-08-21
Add SmallRye Reactive Streams Operators dependency to Vert.x

p-zalejko:
Fix for #3605

### DIFF
--- a/extensions/vertx/deployment/pom.xml
+++ b/extensions/vertx/deployment/pom.xml
@@ -30,7 +30,10 @@
             <groupId>io.quarkus</groupId>
             <artifactId>quarkus-vertx</artifactId>
         </dependency>
-
+	<dependency>
+            <groupId>io.quarkus</groupId>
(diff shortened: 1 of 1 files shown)


cescoffier (approved):
It would be a temporary fix until we change the Axle model.

extensions/vertx/deployment/pom.xml
```diff
@@ -30,7 +30,10 @@
             <groupId>io.quarkus</groupId>
             <artifactId>quarkus-vertx</artifactId>
         </dependency>
-
+	<dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-smallrye-reactive-streams-operators</artifactId>
```
gsmet: @cescoffier is it normal that we only add the runtime dependency to the deployment artifact? I.e. we don't add the deployment one here AND we don't add the runtime dependency to runtime.

Just checking.

cescoffier: it should be in both.

PR #3616, opened 2019-08-21
Test framework: only add the test jar as additional archive if it's different from the app jar

### LINKED ISSUE #3615: Test framework: only add the test jar as additional archive if it's different from the app jar
Currently, when initializing the RuntimeRunner in the QuarkusTestExtension we assume the test location and the app location are different (in case the tests are provided in a JAR). This could not be the case. In which case, we shouldn't be adding the JAR containing the tests as an additional archive to be indexed. Otherwise the same location gets indexed twice which leads to build errors.

aloubyansky:
Fixes #3615

### DIFF
--- a/test-framework/junit5/src/main/java/io/quarkus/test/junit/QuarkusTestExtension.java
+++ b/test-framework/junit5/src/main/java/io/quarkus/test/junit/QuarkusTestExtension.java
@@ -86,7 +86,9 @@ public class QuarkusTestExtension
         if (Files.isDirectory(testClassLocation)) {
             testWiringClassesDir = testClassLocation;
         } else {
-            runnerBuilder.addAdditionalArchive(testClassLocation);
+            if (!appClassLocation.equals(testClassLocation)) {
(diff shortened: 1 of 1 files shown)


geoand (approved):
:+1:

devtools/gradle/gradle.properties
```diff
@@ -0,0 +1 @@
+version = 999-SNAPSHOT
```
gsmet: Hmm, is it related to the issue at hand?

This has consequences on the release process if we add a version in the middle of things.

gsmet (dismissed):
Please see my question inline.

devtools/gradle/gradle.properties
geoand: Oh, I didn't know that, sorry

aloubyansky: Ah, sorry, good catch. It's not related. I keep adding this file locally when I'm working on gradle issues.

aloubyansky: Removed it.

PR #3618, opened 2019-08-21
VirtualAddress fix to remove io.netty.channel.local.LocalAddress

rsvoboda:
VirtualAddress fix to remove io.netty.channel.local.LocalAddress 

This was introduced by copy error when forking from https://github.com/netty/netty/blob/4.1/transport/src/main/java/io/netty/channel/local/LocalAddress.java


patriot1burke (approved):

gsmet:
@rsvoboda good catch!

PR #3623, opened 2019-08-21
Arc - fix bean generic injection for then required and bean type parameters are type variable, but bean one has no bounds

### LINKED ISSUE #3619: Issue with generic bean injection
When there's a generic bean with an unbounded type parameter T [(example)](https://github.com/eclipse/microprofile-metrics/blob/2.0.1/tck/api/src/main/java/io/astefanutti/metrics/cdi/se/ConcurrentGaugedMethodBean.java#L21) and an injection point that declares a particular type for T [(example)](https://github.com/eclipse/microprofile-metrics/blob/2.0.1/tck/api/src/main/java/io/astefanutti/metrics/cdi/se/ConcurrentGaugedMethodBeanTest.java#L93), ARC blows up with `UnsatisfiedResolutionException`. I assume this setup should be valid, because the CDI [spec](https://docs.jboss.org/cdi/spec/2.0/cdi-spec.html#assignable_parameters) says

> A parameterized bean type is considered assignable to a parameterized required type if they have identical raw type and for each parameter:
> (...)
> the required type parameter is an actual type, the bean type parameter is a type variable and the actual type is assignable to the upper bound, if any, of the type variable, or
> (...)

in our case,
- _required type parameter is an actual type_ - `java.lang.Long`
- _the actual type is assignable to the upper bound of the type variable_ - `Long` is assignable to `java.lang.Object` 

The Metrics TCKs, from which I link examples, work without changes with Weld. To make this test work with Quarkus, I had to add an upper bound for the bean (`ConcurrentGaugedMethodBean<T extends Long>`)

manovotn:
…meters are type variable, but bean one has no bounds.

Fixes #3619

### DIFF
--- a/independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/BeanResolver.java
+++ b/independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/BeanResolver.java
@@ -111,7 +111,14 @@ class BeanResolver {
                     throw new IllegalArgumentException("Invalid argument combination " + requiredType + "; " + beanType);
                 }
                 for (int i = 0; i < requiredTypeArguments.size(); i++) {
-                    if (!parametersMatch(requiredTypeArguments.get(i), beanTypeArguments.get(i))) {
+                    Type requiredTypeArg = requiredTypeArguments.get(i);
+                    Type beanTypeArg = beanTypeArguments.get(i);
+                    if (isActualType(requiredTypeArg) && isActualType(beanTypeArg)
+                            && beanTypeArg.name().equals(DotNames.OBJECT)) {
+                        // required type = FooTyped<Long> and bean type is FooTyped<Object>, that is a match
+                        return true;
+                    }
+                    if (!parametersMatch(requiredTypeArg, beanTypeArg)) {
                         return false;
                     }
                 }
--- a/independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/Types.java
+++ b/independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/Types.java
@@ -145,9 +145,9 @@ final class Types {
                         "Producer method return type not found in index: " + producerMethod.returnType().name());
             }
             if (Kind.CLASS.equals(returnType.kind())) {
-                types = getTypeClosure(returnTypeClassInfo, Collections.emptyMap(), beanDeployment, null);
+                types = getTypeClosure(returnTypeClassInfo, true, Collections.emptyMap(), beanDeployment, null);
             } else if (Kind.PARAMETERIZED_TYPE.equals(returnType.kind())) {
-                types = getTypeClosure(returnTypeClassInfo,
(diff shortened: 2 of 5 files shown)


manovotn:
@jmartisk can you give this a go? It should fix the problem you had in the TCKs...

jmartisk:
This caused ambiguous dependencies in some places, can't compile :/

manovotn:
Yea, can see that. I've only ran it through arc tests before sending this, sorry.
I'll check that.

jmartisk (approved):
Works for me!

independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/BeanResolver.java
```diff
@@ -111,7 +111,14 @@ boolean matchesNoBoxing(Type requiredType, Type beanType) {
                     throw new IllegalArgumentException("Invalid argument combination " + requiredType + "; " + beanType);
                 }
                 for (int i = 0; i < requiredTypeArguments.size(); i++) {
-                    if (!parametersMatch(requiredTypeArguments.get(i), beanTypeArguments.get(i))) {
+                    Type requiredTypeArg = requiredTypeArguments.get(i);
+                    Type beanTypeArg = beanTypeArguments.get(i);
+                    if (isActualType(requiredTypeArg) && isActualType(beanTypeArg)
+                            && beanTypeArg.name().equals(DotNames.OBJECT)) {
+                        // required type = FooTyped<Long> and bean type is FooTyped<Object>, that is a match
```
gsmet: We don't support `FooType<Number>` though?

gsmet (commented):
Added a question out of curiosity. Looks good but I would prefer Martin having a look.

independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/BeanResolver.java
manovotn: That should already work I think (should better re-check this).
The problem in this PR was mainly raw types which arc automagically converted to from `FooTyped` to `FooTyped<Object>`

manovotn: Wrong...that actually shouldn't work and it doesn't. Neither on master nor on this PR.
Nonetheless, this PR, I'll put that in a separate comment.

manovotn:
So this PR is partially wrong because (according to specification), when you have an injection point (a required type) of `FooTyped<Long>` and then you have beans:
a) `FooTyped<T>`
b) `FooTyped<Object>`

Then a) should be assignable because that is effectively a parameterized type `T extends Object`,  whereas b) should not as then you are comparing simply `Long` and `Object` and those are not equal.
This PR fixes some problems but also allows b) to pass, which is wrong.

mkouba:
Superseded by https://github.com/quarkusio/quarkus/pull/3707

PR #3625, opened 2019-08-21
Podman requires --userns=keep-id to avoid AccessDeniedExceptions while building the native image

gastaldi:
According to the podman documentation:
--userns=keep-id: creates a user namespace where the current rootless user's UID:GID are mapped to the same values in the container.

Updating to `container-selinux-2.115.0-1.gitfddfbbb` is also required to avoid disabling SELinux (Cannot restore segment prot after reloc: Permission denied).
More information: https://github.com/containers/libpod/issues/3859

### DIFF
--- a/core/creator/src/main/java/io/quarkus/creator/phase/nativeimage/NativeImagePhase.java
+++ b/core/creator/src/main/java/io/quarkus/creator/phase/nativeimage/NativeImagePhase.java
@@ -335,12 +335,16 @@ public class NativeImagePhase implements AppCreationPhase<NativeImagePhase>, Nat
             // E.g. "/usr/bin/docker run -v {{PROJECT_DIR}}:/project --rm quarkus/graalvm-native-image"
             nativeImage = new ArrayList<>();
             Collections.addAll(nativeImage, containerRuntime, "run", "-v", outputDir.toAbsolutePath() + ":/project:z", "--rm");
-
-            if (IS_LINUX & "docker".equals(containerRuntime)) {
-                String uid = getLinuxID("-ur");
-                String gid = getLinuxID("-gr");
-                if (uid != null & gid != null & !"".equals(uid) & !"".equals(gid)) {
-                    Collections.addAll(nativeImage, "--user", uid.concat(":").concat(gid));
+            if (IS_LINUX) {
+                if ("docker".equals(containerRuntime)) {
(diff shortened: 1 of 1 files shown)


gsmet:
@gastaldi do you see someone with Podman experience who could review it? I can approve it but I have no idea if it's the right fix or not :).

gastaldi:
@rhatdan perhaps you could review this? Thanks

rhatdan:
Basically this would mean the container will run as the UID of the user who runs it, meaning any content created will be owned by that user when run as rootless podman.  
Docker running this command will switch to the from root.
The issue is, if the command is running as root, you really want podman to act like docker does.

I would change the check, to check if the command is being run by non root then if doing podman execute the --keep option.  Otherwise use the --user option.

gastaldi:
Hm, even if the user is root, using the `keep-id` option will reuse the same UID from the root user, no? I am not sure I understand why the root check is needed in this case

rhatdan:
The following code seems to be attempting to run the container with a specifc User
```
                String uid = getLinuxID("-ur");
                String gid = getLinuxID("-gr");
                if (uid != null & gid != null & !"".equals(uid) & !"".equals(gid)) {
                    Collections.addAll(nativeImage, "--user", uid.concat(":").concat(gid));
```
Docker and Podman started as root would want to run with this user correct?

Podman run by that user would stay as  that user.

gastaldi:
AFAIK that code uses the effective user and group IDs from the process that runs the Maven plugin (it runs a `id -gr` and `id -ur`), so unless I am missing something (which I probably am) `keep-id` should do that too.

rhatdan:
Ok, so if this is always running as the user, then yes this Patch LGTM.  
Docker is a client server operation, so talking to the `docker daemon` via the docker.sock would require you to launch the container with the user.  

With the change to Podman you no longer require access to the docker.sock, which would hugely increase security.  Since I believe access to the docker.sock is the most dangerous thing you can do on Linux. More dangerous then giving the user sudo root without password.

https://www.projectatomic.io/blog/2014/09/granting-rights-to-users-to-use-docker-in-fedora/

gastaldi:
@rhatdan thank you very much for your comments. @gsmet this PR is good to go :)

cescoffier (approved):

PR #3628, opened 2019-08-21
Optionally replace netty JdkSslContext

wburns:
In an extra extension, we have a custom implementation of ApplicationProtocolNegotiator. Unfortunately I cannot replace the values for it as the method and object reference is already used. This will allow an extension to disable the netty replacement of these classes and the extension that depends on netty can replace them instead.

I originally had the system property start with `quarkus`, but these are special and are not passed to the `native-image` command. If there is a different approach preferred, let me know.

### DIFF
--- a/extensions/netty/runtime/src/main/java/io/quarkus/netty/runtime/graal/NettySubstitutions.java
+++ b/extensions/netty/runtime/src/main/java/io/quarkus/netty/runtime/graal/NettySubstitutions.java
@@ -3,6 +3,7 @@ package io.quarkus.netty.runtime.graal;
 import java.security.PrivateKey;
 import java.security.Provider;
 import java.security.cert.X509Certificate;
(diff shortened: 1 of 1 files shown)


gsmet:
@wburns I saw you discuss it with @dmlloyd on Zulip, what's the status of this exactly?

wburns:
@gsmet I don't think @dmlloyd had anything in particular to say for or against. Unfortunately there is no way I know of to disable a substitution from another extension without doing something like this. And in Infinispan we need to replace a different ALPN provider; this was the least intrusive way I could think of.

To be honest this brings up an issue with Graal substitution in regards to having multiple extensions, that I am not sure if it is a problem long term or not.

wburns:
It is probably easier to just not support ALPN for now in our app. This will be a non issue if Graal ever supports JDK 11 as well.

Issue #3530, opened 2019-08-15
Intercepted subclass does not share interceptor instances


mkouba:
"The lifecycle of an interceptor instance is the same as that of the target class instance with which it is associated."

PR #3639, opened 2019-08-22
Fix @Intercepted metadata injection

### LINKED ISSUE #3594: Wrong bean metadata injected into interceptor using @Intercepted when using dependent scope
I'm trying to run MP Metrics TCK, in particular `ConcurrentGaugedClassBeanTest` [1]. I am facing an issue that when a `ConcurrentGaugeInterceptor` is being constructed during the test to intercept calls to methods of `ConcurrentGaugedClassBean` [2], wrong bean is injected into its `@Intercepted` injection point [3] - the `bean` parameter should contain the `ConcurrentGaugedClassBean` that is being called, but instead it contains the test that is being run. Therefore the interceptor can't work properly because it can't find the right metric instance which should be updated during the call.
This has something to do with the fact that `ConcurrentGaugedClassBean` has the dependent CDI scope. This issue doesn't occur if I attach any other CDI scope to `ConcurrentGaugedClassBean` in the TCK code. 

So I assume the problem is that when bean A injects a bean B that has a dependent scope, and that bean B has got an interceptor applied, and the interceptor's constructor injects the metadata of the intercepted bean using `@Intercepted`, the bean that gets injected into the interceptor's constructor is not B (the bean being injected), but A - the bean that is injecting B. This differs from the behavior I observe with Thorntail, where `@Intercepted` correctly injects the metadata of B.

How to reproduce:
- Use my branch [4]
- from `quarkus/tcks/microprofile-metrics/api`: `mvn test -Dtest=ConcurrentGaugedClassBeanTest#callCountedMethodsOnce`
You will see the test fail with `java.lang.UnsupportedOperationException` because the concurrent gauge interceptor will receive a `DoesNotHaveMetric` metric resolver, because it will try to locate the metric metadata on the test class instead of the `ConcurrentGaugedClassBean` bean

[1] https://github.com/eclipse/microprofile-metrics/blob/2.0.1/tck/api/src/main/java/io/astefanutti/metrics/cdi/se/ConcurrentGaugedClassBeanTest.java#L120
[2] https://github.com/eclipse/microprofile-metrics/blob/2.0.1/tck/api/src/main/java/io/astefanutti/metrics/cdi/se/ConcurrentGaugedClassBean.java
[3] https://github.com/smallrye/smallrye-metrics/blob/2.1.3/implementation/src/main/java/io/smallrye/metrics/interceptors/ConcurrentGaugeInterceptor.java#L56
[4] https://github.com/jmartisk/quarkus/commits/metrics-tck-interceptor-reproducer

manovotn:
Fixes #3594 and is also additional fix for #3530 

I am not all too confident about changes to `InterceptedBeanMetadataProvider` but this state is what gives me green on all tests locally, so it stand a chance of being right I suppose :)

### DIFF
--- a/independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/BeanGenerator.java
+++ b/independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/BeanGenerator.java
@@ -823,6 +823,11 @@ public class BeanGenerator extends AbstractGenerator {
                 Set<InterceptorInfo> wraps = new HashSet<>();
                 wraps.addAll(aroundConstructs.interceptors);
                 wraps.addAll(postConstructs.interceptors);
+
+                // instances of around/post construct interceptors also need to be shared
+                // build a map that links InterceptorInfo to ResultHandle and reuse that when creating wrappers
+                Map<InterceptorInfo, ResultHandle> interceptorToResultHandle = new HashMap<>();
+
                 for (InterceptorInfo interceptor : wraps) {
                     ResultHandle interceptorProvider = create.readInstanceField(
                             FieldDescriptor.of(beanCreator.getClassName(), interceptorToProviderField.get(interceptor),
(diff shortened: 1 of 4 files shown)


manovotn:
@jmartisk FYI, this should fix yet another of those TCK issues you are fighting

jmartisk (approved):
I confirm this resolves the Metrics TCK issues mentioned in https://github.com/quarkusio/quarkus/issues/3594

cescoffier (approved):

PR #3640, opened 2019-08-22
Correct calculation of remaining time in millis within the AmazonLambdaContext

debae:
…daContext

### DIFF
--- a/bom/runtime/pom.xml
+++ b/bom/runtime/pom.xml
@@ -165,6 +165,7 @@
         <wiremock.version>2.24.1</wiremock.version>
         <spring.version>5.1.8.RELEASE</spring.version>
         <mvel2.version>2.4.4.Final</mvel2.version>
+        <mockito.version>3.0.0</mockito.version>
     </properties>
 
     <dependencyManagement>
@@ -1711,6 +1712,11 @@
                 <groupId>org.neo4j.driver</groupId>
                 <artifactId>neo4j-java-driver</artifactId>
                 <version>${neo4j-java-driver.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>org.mockito</groupId>
+                <artifactId>mockito-core</artifactId>
+                <version>${mockito.version}</version>
             </dependency>
             <dependency>
                 <groupId>io.smallrye.reactive</groupId>
--- a/extensions/amazon-lambda/runtime/pom.xml
+++ b/extensions/amazon-lambda/runtime/pom.xml
@@ -35,6 +35,16 @@
             <groupId>com.fasterxml.jackson.core</groupId>
             <artifactId>jackson-databind</artifactId>
         </dependency>
+        <dependency>
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-core</artifactId>
+            <scope>test</scope>
+        </dependency>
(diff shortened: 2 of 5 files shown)


extensions/amazon-lambda/runtime/src/test/java/io/quarkus/amazon/lambda/runtime/AmazonLambdaContextTest.java
```diff
@@ -0,0 +1,26 @@
...
+    public void testGetRemainingTimeInMillis() throws Exception {
+        HttpURLConnection request = mock(HttpURLConnection.class);
+
+        // We add 5 seconds to the current date and set this as lambda deadline, so we know there is still some remaining time left
+        when(request.getHeaderField(AmazonLambdaApi.LAMBDA_RUNTIME_DEADLINE_MS))
+                .thenReturn(String.valueOf(System.currentTimeMillis() + 5000L));
+
+        AmazonLambdaContext amazonLambdaContext = new AmazonLambdaContext(request, null, null);
+
+        System.out.println(amazonLambdaContext.getRemainingTimeInMillis());
```
gsmet: Please remove the output, it's useless.

extensions/amazon-lambda/runtime/pom.xml
```diff
@@ -35,6 +35,17 @@
...
+        <dependency>
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-core</artifactId>
+            <version>3.0.0</version>
+            <scope>test</scope>
```
gsmet: I think we already have other mocking library declared or maybe this one. Could you check the bom/build-parent and see if we can avoid having 3 different mocking library?

gsmet (changes requested):
The fix looks good. I added some comment about the test.

TBH, I'm wondering if we really need a test for that. Granted it was incorrect but now that it's correct I don't see it having regressions.

So depending on the amount of effort you want to put in it, feel free to remove the test altogether.

extensions/amazon-lambda/runtime/pom.xml
debae: I could only find mockito as dependency in devtools. Didn't find any other mocking frameworks to use inside unit tests, except for mocking http / ... but that's a different purpose.

gsmet: OK, looks like I misremember. I would have sworn having seen at least Powermock. Sorry about that!

The version should go in the bom though.

bom/runtime/pom.xml
```diff
@@ -1712,6 +1713,12 @@
                 <artifactId>neo4j-java-driver</artifactId>
                 <version>${neo4j-java-driver.version}</version>
             </dependency>
+            <dependency>
+                <groupId>org.mockito</groupId>
+                <artifactId>mockito-core</artifactId>
+                <version>${mockito.version}</version>
+                <scope>test</scope>
```
gsmet: One last minor comment: we don't want to define the scope in the bom. We did that at the beginning and we had people complaining about it.

Once it's fixed, this can go in.

Thanks!

gsmet (changes requested):

bom/runtime/pom.xml
debae: No problem, it's fixed.

gsmet (approved):
Perfect, thanks!

PR #3658, opened 2019-08-23
reactive-pg-client.version removal from BOM

rsvoboda:
reactive-pg-client.version removal from BOM, this property is not used

Leftover from Vertx and new clients upgrade


FroMage (approved):

FroMage:
Well spotted, thanks.

Issue #3107, opened 2019-07-05
Make sure each relevant extension offer a liveliness and readiness data


emmanuelbernard:
Of course first checking if MP Health is an enlisted extension

loicmathieu:
+1 I miss this functionality from other framework :
- When Agroal is in the classpath, we should automatically check the DS and add a health check about it
- When Kafka is in the classpath we should automatically check the broker and add a health check
- ...

Of course we need to be able to disable those health checks if needed.

As Quarkus is build time and prevent the usage of reflection, I don't know how to implement this as we need from one extension (agroal) to test if another is here (health check) and if yes provide the needed build items ...

I'm interested in this functionality so I'm happy to help if needed ;)

loicmathieu:
As I understand liveness : it's about restarting the application so the extension should not provide liveness because I don't want to restart my apps if my DB is down because it will be useless (it's the DB that needs to be restarted).

**Liveness**: global UP for the apps without checking any dependecies (so any extension)
**Readiness** : UP for my apps and all it's dependecies (DB, broker, LDAP, external rest, ...)

So I think only health check needs to be provided by extentions, they map to readiness checks (in k8s).

emmanuelbernard:
@loicmathieu on the build time discovery, the idea is to have a `LivenessBuildItem` and a `ReadinessBuildItem` probably in a shared extension like `health-producer-extension` or something like that.
Every extension wanting to produce a health check would add a dependency to this shared extension and produce the relevant build item.
The healthcheck extensions proper would depend on `health-producer-extension` and consume the build items.

There is I think a trick to make sure to push these health check classes in the runtime path if and only if the build item is consumed.
Only if there is a build item consumer would this build step be called. That solves that problem.
The second problem is to have this healthcheck code pushed at runtime.
@stuartwdouglas  can advise but I imagine either or a combination of the SR health check extension registering the class for reflection or generating this class via the gizmo based recorder?

stuartwdouglas:
I have made a start on this here: https://github.com/stuartwdouglas/quarkus/pull/new/build-in-health-checks

No tests but it should work, it just needs some actual health check implementations.

kenfinnigan:
Though I think it's a good idea to offer built in health checks for our extensions, we need to do it in a way that's very easy to turn off.

My concern is that users will want very specific to their application ways for different extensions, so we need a way to turn them off so it's an "all or nothing" type approach

emmanuelbernard:
[...]

PR #3671, opened 2019-08-23
Provides extension health via builditem

loicmathieu:
This PR implements health check inside the Agroal extension so a user will have an health check that validate the database if it includes the agroal extension.

I create a `HealthBuildItem` that an extension needs to emmit to inlcude it's health check. Smallrye Heath extension will use these `HealthBuildItem`s to add the health check to ArC if extension health check is enabled (true by default).

It can be disabled with the following properties:
```
# disable the datasource health check
quarkus.datasource.health.enabled=false
# disable all extension health check
quarkus.health.extensions.enabled=false
```

This is part of #3107

Ping @emmanuelbernard @kenfinnigan @stuartwdouglas

### DIFF
--- a/docs/src/main/asciidoc/datasource-guide.adoc
+++ b/docs/src/main/asciidoc/datasource-guide.adoc
@@ -244,7 +244,7 @@ to validate the datasource.
 So when you access the `/health/ready` endpoint of your application you will have information about the datasource validation status.
 If you have multiple datasources, all datasources will be checked and the status will be `DOWN` as soon as there is one datasource validation failure.
 
-This behavior can be disabled via the property `quarkus.datasource.health.disabled`.
+This behavior can be disabled via the property `quarkus.datasource.health.enabled`.
 
 
 [[configuration-reference]]
--- a/docs/src/main/asciidoc/datasource-guide.adoc
+++ b/docs/src/main/asciidoc/datasource-guide.adoc
@@ -236,6 +236,16 @@ AgroalDataSource dataSource1;
 AgroalDataSource dataSource2;
 --
 
+== Datasource Health Check
+
+If you are using the `quarkus-smallrye-health` extension, `quarkus-agroal` will automatically add a readiness health check
+to validate the datasource.
+
+So when you access the `/health/ready` endpoint of your application you will have information about the datasource validation status.
+If you have multiple datasources, all datasources will be checked and the status will be `DOWN` as soon as there is one datasource validation failure.
+
+This behavior can be disabled via the property `quarkus.datasource.health.disabled`.
+
 
 [[configuration-reference]]
 == Agroal Configuration Reference
--- a/docs/src/main/asciidoc/extension-authors-guide.adoc
+++ b/docs/src/main/asciidoc/extension-authors-guide.adoc
@@ -1206,6 +1206,36 @@ As a CDI based runtime, Quarkus extensions often make CDI beans available as par
 However, Quarkus DI solution does not support CDI Portable Extensions.
 Instead, Quarkus extensions can make use of various link:cdi-reference.html[Build Time Extension Points].
 
+=== Extension Health Check
+
+Health checks are provided via the `quarkus-smallrye-health` extension. It provides both liveness and readiness checks capabilities.
+
+When writing an extension, it's beneficial to provide health checks for the extension, that can be automatically included without the developer needing to write their own.
+
+In order to provide a health check, you should do the following:
+
+- Import the `quarkus-smallrye-health` extension as an **optional** dependency in your runtime module so it will not impact the size of the application if
+health check is not included.
+- Create your health check following the link:health-guide.html[Quarkus - MicroProfile Health] guide. We advise providing only
+readiness check for an extension (liveness check is designed to express the fact that an application is up and needs to be lightweight).
+- Import the `quarkus-smallrye-health-spi` library in your deployment module.
+- Add a build step in your deployment module that produces a `HealthBuildItem`.
+- Add a way to disable the extension health check via a config item `quarkus.<extension>.health.enabled` that should be enabled by default.
+
+Following is an example from the Agroal extension that provides a `DataSourceHealthCheck` to validate the readiness of a datasource.
+
+[source%nowrap,java]
+----
+@BuildStep
+HealthBuildItem addHealthCheck(AgroalBuildTimeConfig agroalBuildTimeConfig) {
+    if (agroalBuildTimeConfig.healthEnabled) {
+        //produce the health bean
+        return new HealthBuildItem(DataSourceHealthCheck.class);
+    }
+    return null;
+}
+----
+
 === Testing Extensions
 
 Testing of extensions should be done with the `io.quarkus.test.QuarkusUnitTest` runner. This runner allows
--- a/docs/src/main/asciidoc/extension-authors-guide.adoc
+++ b/docs/src/main/asciidoc/extension-authors-guide.adoc
@@ -1228,11 +1228,7 @@ Following is an example from the Agroal extension that provides a `DataSourceHea
 ----
 @BuildStep
 HealthBuildItem addHealthCheck(AgroalBuildTimeConfig agroalBuildTimeConfig) {
-    if (agroalBuildTimeConfig.healthEnabled) {
-        //produce the health bean
-        return new HealthBuildItem(DataSourceHealthCheck.class);
-    }
-    return null;
+    return new HealthBuildItem(DataSourceHealthCheck.class, agroalBuildTimeConfig.healthEnabled);
 }
 ----
 
--- a/docs/src/main/asciidoc/extension-authors-guide.adoc
+++ b/docs/src/main/asciidoc/extension-authors-guide.adoc
@@ -1228,7 +1228,8 @@ Following is an example from the Agroal extension that provides a `DataSourceHea
 ----
 @BuildStep
 HealthBuildItem addHealthCheck(AgroalBuildTimeConfig agroalBuildTimeConfig) {
-    return new HealthBuildItem(DataSourceHealthCheck.class, agroalBuildTimeConfig.healthEnabled);
+    return new HealthBuildItem("io.quarkus.agroal.runtime.health.DataSourceHealthCheck",
+            agroalBuildTimeConfig.healthEnabled, "datasource");
 }
 ----
 
--- a/docs/src/main/asciidoc/health-guide.adoc
+++ b/docs/src/main/asciidoc/health-guide.adoc
@@ -296,6 +296,15 @@ error along with the health check response.
         }
 ----
 
+== Extension health checks
+
+Some extension may provide default health checks, including the extension will automatically register its health checks.
+
+For example, `quarkus-agroal` that is used to manage Quarkus datasource(s) automatically register a readiness health check
+that will validate each datasources: link:datasource-guide.html#datasource-health-check[Datasource Health Check].
+
+You can disable extension health check via the property `quarkus.health.extension.enabled` so none will be automatically registered.
+
 == Conclusion
 
 MicroProfile Health provides a way for your application to distribute information 
--- a/docs/src/main/asciidoc/health-guide.adoc
+++ b/docs/src/main/asciidoc/health-guide.adoc
@@ -303,7 +303,7 @@ Some extension may provide default health checks, including the extension will a
 For example, `quarkus-agroal` that is used to manage Quarkus datasource(s) automatically register a readiness health check
 that will validate each datasources: link:datasource-guide.html#datasource-health-check[Datasource Health Check].
 
-You can disable extension health check via the property `quarkus.health.extension.enabled` so none will be automatically registered.
+You can disable extension health check via the property `quarkus.health.extensions.enabled` so none will be automatically registered.
 
 == Conclusion
 
--- a/extensions/agroal/deployment/pom.xml
+++ b/extensions/agroal/deployment/pom.xml
@@ -30,12 +30,9 @@
             <groupId>io.quarkus</groupId>
             <artifactId>quarkus-narayana-jta-deployment</artifactId>
         </dependency>
-
-        <!-- Add the health extension as optional as we will produce the health check only if it's included -->
         <dependency>
             <groupId>io.quarkus</groupId>
-            <artifactId>quarkus-smallrye-health</artifactId>
-            <optional>true</optional>
+            <artifactId>quarkus-smallrye-health-spi</artifactId>
         </dependency>
 
         <!-- test dependencies -->
--- a/extensions/agroal/deployment/pom.xml
+++ b/extensions/agroal/deployment/pom.xml
@@ -31,6 +31,13 @@
             <artifactId>quarkus-narayana-jta-deployment</artifactId>
         </dependency>
 
+        <!-- Add the health extension as optional as we will produce the health check only if it's included -->
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-smallrye-health</artifactId>
+            <optional>true</optional>
+        </dependency>
+
         <!-- test dependencies -->
         <dependency>
             <groupId>io.quarkus</groupId>
--- a/extensions/agroal/runtime/src/main/java/io/quarkus/agroal/runtime/AgroalBuildTimeConfig.java
+++ b/extensions/agroal/runtime/src/main/java/io/quarkus/agroal/runtime/AgroalBuildTimeConfig.java
@@ -22,7 +22,7 @@ public class AgroalBuildTimeConfig {
     public Map<String, DataSourceBuildTimeConfig> namedDataSources;
 
     /**
-     * Whether or not an healtcheck is published in case the smallrye-health extionsion is present (default to true).
+     * Whether or not an healtcheck is published in case the smallrye-health extension is present (default to true).
      */
     @ConfigItem(name = "health.enabled", defaultValue = "true")
     public boolean healthEnabled;
--- a/extensions/agroal/runtime/src/main/java/io/quarkus/agroal/runtime/AgroalBuildTimeConfig.java
+++ b/extensions/agroal/runtime/src/main/java/io/quarkus/agroal/runtime/AgroalBuildTimeConfig.java
@@ -20,4 +20,10 @@ public class AgroalBuildTimeConfig {
      */
     @ConfigItem(name = ConfigItem.PARENT)
     public Map<String, DataSourceBuildTimeConfig> namedDataSources;
+
+    /**
+     * Whether or not an healtcheck is published in case the smallrye-health extionsion is present (default to true).
+     */
+    @ConfigItem(name = "health.enabled", defaultValue = "true")
+    public boolean healthEnabled;
 }
--- a/extensions/agroal/runtime/src/main/java/io/quarkus/agroal/runtime/health/DataSourceHealthCheck.java
+++ b/extensions/agroal/runtime/src/main/java/io/quarkus/agroal/runtime/health/DataSourceHealthCheck.java
@@ -53,8 +53,7 @@ public class DataSourceHealthCheck implements HealthCheck {
                     builder.down().withData(dsName, data);
                 }
             } catch (SQLException e) {
-                String data = DEFAULT_DS.equals(dataSource.getKey())
-                        ? "Unable to execute the validation check for the default DataSource: "
+                String data = isDefault ? "Unable to execute the validation check for the default DataSource: "
                         : "Unable to execute the validation check for DataSource '" + dataSource.getKey() + "': ";
                 String dsName = isDefault ? "quarkus-default-ds" : dataSource.getKey();
                 builder.down().withData(dsName, data + e.getMessage());
--- a/extensions/agroal/runtime/src/main/java/io/quarkus/agroal/runtime/health/DataSourceHealthCheck.java
+++ b/extensions/agroal/runtime/src/main/java/io/quarkus/agroal/runtime/health/DataSourceHealthCheck.java
@@ -1,0 +1,65 @@
+package io.quarkus.agroal.runtime.health;
+
+import java.sql.Connection;
+import java.sql.SQLException;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+
+import javax.annotation.PostConstruct;
+import javax.enterprise.context.ApplicationScoped;
+import javax.enterprise.inject.spi.Bean;
+import javax.sql.DataSource;
+
+import org.eclipse.microprofile.health.HealthCheck;
+import org.eclipse.microprofile.health.HealthCheckResponse;
+import org.eclipse.microprofile.health.HealthCheckResponseBuilder;
+import org.eclipse.microprofile.health.Readiness;
+
+import io.quarkus.arc.Arc;
+
+@Readiness
+@ApplicationScoped
+public class DataSourceHealthCheck implements HealthCheck {
+    private static final String DEFAULT_DS = "__default__";
+    private Map<String, DataSource> dataSources = new HashMap<>();
+
+    @PostConstruct
+    protected void init() {
+        Set<Bean<?>> beans = Arc.container().beanManager().getBeans(DataSource.class);
+        for (Bean<?> bean : beans) {
+            if (bean.getName() == null) {
+                // this is the default DataSource: retrieve it by type
+                DataSource defaultDs = Arc.container().instance(DataSource.class).get();
+                dataSources.put(DEFAULT_DS, defaultDs);
+            } else {
+                DataSource ds = (DataSource) Arc.container().instance(bean.getName()).get();
+                dataSources.put(bean.getName(), ds);
+            }
+        }
+    }
+
+    @Override
+    public HealthCheckResponse call() {
+        HealthCheckResponseBuilder builder = HealthCheckResponse.named("Database connection(s) health check").up();
+        for (Map.Entry<String, DataSource> dataSource : dataSources.entrySet()) {
+            boolean isDefault = DEFAULT_DS.equals(dataSource.getKey());
+            try (Connection con = dataSource.getValue().getConnection()) {
+                boolean valid = con.isValid(0);
+                if (!valid) {
+                    String data = isDefault ? "validation check failed for the default DataSource"
+                            : "validation check failed for DataSource '" + dataSource.getKey() + "'";
+                    String dsName = isDefault ? "quarkus-default-ds" : dataSource.getKey();
+                    builder.down().withData(dsName, data);
+                }
+            } catch (SQLException e) {
+                String data = DEFAULT_DS.equals(dataSource.getKey())
+                        ? "Unable to execute the validation check for the default DataSource: "
+                        : "Unable to execute the validation check for DataSource '" + dataSource.getKey() + "': ";
+                String dsName = isDefault ? "quarkus-default-ds" : dataSource.getKey();
+                builder.down().withData(dsName, data + e.getMessage());
+            }
+        }
+        return builder.build();
+    }
+}
--- a/extensions/smallrye-health/spi/src/main/java/io/quarkus/smallrye/health/deployment/spi/HealthBuildItem.java
+++ b/extensions/smallrye-health/spi/src/main/java/io/quarkus/smallrye/health/deployment/spi/HealthBuildItem.java
@@ -4,12 +4,18 @@ import io.quarkus.builder.item.MultiBuildItem;
 
 public final class HealthBuildItem extends MultiBuildItem {
     private Class<?> healthCheckClass;
+    private boolean enabled;
 
-    public HealthBuildItem(Class<?> healthCheckClass) {
+    public HealthBuildItem(Class<?> healthCheckClass, boolean enabled) {
         this.healthCheckClass = healthCheckClass;
+        this.enabled = enabled;
     }
 
     public Class<?> getHealthCheckClass() {
         return this.healthCheckClass;
+    }
+
+    public boolean isEnabled() {
+        return enabled;
     }
 }
--- a/extensions/smallrye-health/spi/src/main/java/io/quarkus/smallrye/health/deployment/spi/HealthBuildItem.java
+++ b/extensions/smallrye-health/spi/src/main/java/io/quarkus/smallrye/health/deployment/spi/HealthBuildItem.java
@@ -1,0 +1,15 @@
+package io.quarkus.smallrye.health.deployment.spi;
+
+import io.quarkus.builder.item.MultiBuildItem;
+
+public final class HealthBuildItem extends MultiBuildItem {
+    private Class<?> healthCheckClass;
+
+    public HealthBuildItem(Class<?> healthCheckClass) {
+        this.healthCheckClass = healthCheckClass;
+    }
+
+    public Class<?> getHealthCheckClass() {
+        return this.healthCheckClass;
+    }
+}
--- a/extensions/smallrye-health/spi/src/main/java/io/quarkus/smallrye/health/deployment/spi/HealthBuildItem.java
+++ b/extensions/smallrye-health/spi/src/main/java/io/quarkus/smallrye/health/deployment/spi/HealthBuildItem.java
@@ -3,19 +3,32 @@ package io.quarkus.smallrye.health.deployment.spi;
 import io.quarkus.builder.item.MultiBuildItem;
 
 public final class HealthBuildItem extends MultiBuildItem {
-    private Class<?> healthCheckClass;
+    private String healthCheckClass;
     private boolean enabled;
+    private String configRootName;
 
-    public HealthBuildItem(Class<?> healthCheckClass, boolean enabled) {
+    /**
+     * @param healthCheckClass the name of the health check class, needs to implements
+     *        org.eclipse.microprofile.health.HealthCheck
+     * @param enabled whether or not the check is enabled
+     * @param configRootName the name of the root configuration of the extension as defined by the <code>@ConfigRoot</code>
+     *        annotation.
+     */
+    public HealthBuildItem(String healthCheckClass, boolean enabled, String configRootName) {
         this.healthCheckClass = healthCheckClass;
         this.enabled = enabled;
+        this.configRootName = configRootName;
     }
 
-    public Class<?> getHealthCheckClass() {
+    public String getHealthCheckClass() {
         return this.healthCheckClass;
     }
 
     public boolean isEnabled() {
         return enabled;
+    }
+
+    public String getConfigRootName() {
+        return configRootName;
     }
 }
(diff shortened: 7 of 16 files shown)


extensions/agroal/deployment/pom.xml
```diff
@@ -31,6 +31,13 @@
             <artifactId>quarkus-narayana-jta-deployment</artifactId>
         </dependency>
 
+        <!-- Add the health extension as optional as we will produce the health check only if it's included -->
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-smallrye-health</artifactId>
```
machi1990: is `-deployment` missing from the artifactId?

extensions/agroal/runtime/src/main/java/io/quarkus/agroal/runtime/health/DataSourceHealthCheck.java
```diff
@@ -0,0 +1,36 @@
...
+    Instance<DataSource> dataSources;
+
+    @Override
+    public HealthCheckResponse call() {
+        HealthCheckResponseBuilder builder = HealthCheckResponse.named("Database connection(s) health check").up();
+        for (DataSource dataSource : dataSources) {
+            try {
+                boolean valid = dataSource.getConnection().isValid(0);
+                if (!valid) {
+                    builder.down().withData("reason", "validation check failed");
```
machi1990: I think this will be overridden if two datasources are down? 
Since we support multiple datasources, will it be possible to return an object having this structure in case of failures? 
``` 
{
 "reason": {
    "validation check failed",
     "datasources": ["data-source-name-1", "data-source-name-2", ..., "data-source-name-N"]
  }
} 
```

extensions/agroal/runtime/src/main/java/io/quarkus/agroal/runtime/AgroalBuildTimeConfig.java
```diff
@@ -20,4 +20,10 @@
      */
     @ConfigItem(name = ConfigItem.PARENT)
     public Map<String, DataSourceBuildTimeConfig> namedDataSources;
+
+    /**
+     * Whether or not an healtcheck is published in case the smallrye-health extionsion is present (default to true).
+     */
+    @ConfigItem(name = "health.enabled", defaultValue = "true")
```
machi1990: I wonder if leaving out this naming the config item and having a default name `health-enabled` would have been a viable alternative? 

Also would one of `enable-healthcheck` or `healthcheck-enabled` would be better name for this config item?

Should the default value be controlled by whatever is set in `HealthBuildTimeConfig#extensionsEnabled`?

extensions/smallrye-health/runtime/src/main/java/io/quarkus/smallrye/health/runtime/HealthBuildTimeConfig.java
```diff
@@ -0,0 +1,15 @@
...
+import io.quarkus.runtime.annotations.ConfigRoot;
+
+@ConfigRoot(name = "health")
+public class HealthBuildTimeConfig {
+    /**
+     * Whether or not extensions published health check should be enabled.
+     */
+    // This config item is used by extensions to avoid to publish the health check but not directly used in the health
+    // extension. It must exists to avoid a log WARN of 'Unrecognized configuration key'.
+    @ConfigItem(name = "extensions.enabled", defaultValue = "true")
```
machi1990: How does renaming this to `enableExtensionCheck` with the config key being `enable-extension-check` sound? 

Will defaulting to hyphenated config key name  be a viable naming? e.g taking for example the present variable naming i.e `extensionsEnabled ` the configuration key will be `extensions-enabled`?

machi1990 (changes requested):
@loicmathieu I just have a couple of questions and suggestions. Great PR otherwise.

extensions/agroal/deployment/pom.xml
loicmathieu: Agroal only needs the runtime dependency to be able to compile the health check. Moreover the dependency is optional to avoid adding it in case agroal is used without smallrye-health.

In fact we can witch to io.smallrye:smallrye-health dependency instead of Quarkus one if we want as there is no dependency to the extention ...

extensions/agroal/runtime/src/main/java/io/quarkus/agroal/runtime/health/DataSourceHealthCheck.java
loicmathieu: In case multiple datasources down, there will be multiple data with the same content as the `withData` method adds the data to a HashMap.

First I want to implement exactly what you ask ... then I realize that I don't know how to retrieve the beans with there name in Arc :) 

In case multiple beans of the same type I assume they will have different qualifier. But in CDI to retrieve multiple beans of the same type you use `Instance<Datasource>` that provides the list of the beans and not there names ... If someone can explain me how to retrieve the list of beans from ArC with there qualifier (name) I'll update my implementation.

extensions/agroal/runtime/src/main/java/io/quarkus/agroal/runtime/AgroalBuildTimeConfig.java
loicmathieu: This is a design choice still opens to discussion ;) I know that @emmanuelbernard want that as much as possible works will be included inside the smallrye-health extension but I think that it's the agroal extension that needs to control it's configuration and the integration of the health check: this will be less works to do during the augmenation phase ...

Your proposal to add `HealthBuildTimeConfig#extensionsEnabled` will not facilitate the usage for the agroal extension a lot as it will still need to support the configuration item ...

extensions/smallrye-health/runtime/src/main/java/io/quarkus/smallrye/health/runtime/HealthBuildTimeConfig.java
loicmathieu: These are the configuration names proposed in the original issue ... I'm open to suggestion but I found them easy to use and honestly I'm not a big fan of hyphenated names ...

kenfinnigan (changes requested):
Would be good if @xstefank and @antoinesd also looked at this

extensions/smallrye-health/runtime/src/main/java/io/quarkus/smallrye/health/runtime/HealthBuildTimeConfig.java
```diff
@@ -0,0 +1,15 @@
+package io.quarkus.smallrye.health.runtime;
+
+import io.quarkus.runtime.annotations.ConfigItem;
+import io.quarkus.runtime.annotations.ConfigRoot;
+
+@ConfigRoot(name = "health")
+public class HealthBuildTimeConfig {
```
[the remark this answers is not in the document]
loicmathieu: It confuses me also but I copied it from other extensions that separate runtime and buildtime config in two classes, both in the runtime artifact.

IMHO, `SmallRyeHealthProcessor.SmallRyeHealthConfig` is smallrye related (config root is `quarkus.smallrye-health`) and here it's Quarkus specific stuff (config root is `quarkus.health`).

extensions/agroal/runtime/src/main/java/io/quarkus/agroal/runtime/health/DataSourceHealthCheck.java
loicmathieu: Totaly agree but as I reply to @machi1990 just above, I don't know how to retrieve the bean names from ArC, I use `@Inject Instance<DataSource> dataSources;` that gives me all datasources but not the bean name/qualifier ...

extensions/agroal/runtime/src/main/java/io/quarkus/agroal/runtime/health/DataSourceHealthCheck.java
```diff
@@ -0,0 +1,36 @@
...
+                }
+            } catch (SQLException e) {
+                builder.down().withData("reason", "SQLException trying to validate the connection: " + e.getMessage());
```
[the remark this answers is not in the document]
loicmathieu: Maybe `"Unable to execute the health check: " + e.getMessage` ?

extensions/agroal/runtime/src/main/java/io/quarkus/agroal/runtime/health/DataSourceHealthCheck.java
machi1990: > ... If someone can explain me how to retrieve the list of beans from ArC with there qualifier (name) I'll update my implementation.

@loicmathieu I am certainly not a CDI expert, but have you tried something like this? 
```
 final Set<Bean<?>> beans = Arc.container().beanManager().getBeans(javax.sql.DataSource.class);
        for (Bean<?> bean : beans) {
            if (bean.getName() == null) {
                //  default DataSource has not name can be retrieved via
                //  @Inject
                //  javax.sql.DataSource defaultDataSource;

                //TODO do something with defaultDataSource
            } else {
                InstanceHandle<javax.sql.DataSource> instanceHandle = Arc.container().instance(bean.getName());
                if (instanceHandle.isAvailable()) {
                    javax.sql.DataSource dataSource = instanceHandle.get();
                    // TODO do something with dataSource
                }
            }
        }
```
Also pinging @mkouba @manovotn, may be there is a better way to achieve this  :-)

manovotn: Your solution is viable, otherwise you could use `@Any Instance<DataSource>` which would bring you all beans of type `DataSource` regardless of their qualifiers. You can then iterate over all of them and inspect them.

If you are after bean names (and not `DataSource` instances), then going via `Arc.container().beanManager().getBeans(...)` is better.

loicmathieu: Thanks @machi1990 and @manovotn I used `Arc.container().beanManager().getBeans(...)` in a `  @PostConstruct` and it works fine!

loicmathieu: @kenfinnigan the first argument of `withData` should be a valid json attribute, for the moment I ended up generating `reason1`, `reason2`, ... but maybe `datasourceName-check-status` will be better but the default datasource has no name and if I put `default` it can collide with someone that use a named datasource of name `default`...

loicmathieu: No, it surprises me so I check but you can have a default datasource (with no name) and a nammed datasource. See [Multiple Datasource](https://quarkus.io/guides/datasource-guide#multiple-datasources).

So I can configure something like this:
```
# default un-named datasource : can be injected with @Inject with no other annotation
quarkus.datasource.driver=org.h2.Driver
quarkus.datasource.url=jdbc:h2:tcp://localhost/mem:default
quarkus.datasource.username=username-default
quarkus.datasource.min-size=3
quarkus.datasource.max-size=13

# Named datasource with the name default, must be injected with @Datasource annotation
quarkus.datasource.default.driver=org.h2.Driver
quarkus.datasource.default.url=jdbc:h2:tcp://localhost/mem:users
quarkus.datasource.default.username=username1
quarkus.datasource.default.min-size=1
quarkus.datasource.default.max-size=11
```

loicmathieu: It's just a bit misleading for the developer to talk about the datasource named default in this case: do it refers to the default unamed datasource or the named default datasource ? My proposal clearly talk about the default datasource or the datasource named 'default' in this case.

loicmathieu: You can, the default datasource has no name (name = null)

loicmathieu:
@kenfinnigan @machi1990 I add some documentation.
I didn't add documentation to the extension guide as there is no extension specific section in the extension guide but maybe it worth it ?

docs/src/main/asciidoc/health-guide.adoc
```diff
@@ -296,6 +296,15 @@ error along with the health check response.
         }
 ----
 
+== Extension health checks
+
+Some extension may provides default health checks, including the extension will automatically register its health checks.
+
+For example, `quarkus-agroal` that is use to manage Quarkus datasource(s) automatically register a readiness health check
```
machi1990: There is a typo .. `s/that is use/that is used`

loicmathieu:
@kenfinnigan I add a section in the extension guide. 
I found this guide very complex and I have a hard time decided where to include this new section so I'm not sure it was the best place to put it ...

loicmathieu:
@kenfinnigan I didn't provide any tests, should I create an health check on the test-extension and validate of the test-extension integration test that the health check is successfully integrated ?

dmlloyd:
Let's follow the general approach of putting shared build items into an extension-specific SPI module.  This way the build items can be shared without introducing hard dependencies on the extension deployment modules.

loicmathieu:
@dmlloyd I implemented the SPI approach, can you check if it's correclty implemented ?

docs/src/main/asciidoc/datasource-guide.adoc
```diff
@@ -223,6 +223,16 @@ AgroalDataSource dataSource1;
 AgroalDataSource dataSource2;
 --
 
+== Datasource Health Check
+
+If you are using the `quarkus-smallrye-health` extension, `quarkus-agroal` will automatically add a readiness health check
+to validate the datasource.
+
+So when you access the `/health/ready` endpoint of your application you will have information about the datasource validation status.
+If you have multiple datasources, all datasources will be checked and the status will be `DOWN` as soon as there is one datasource validation failed.
```
kenfinnigan: Change "validation failed" to "validation failure"

docs/src/main/asciidoc/extension-authors-guide.adoc
```diff
@@ -874,6 +874,37 @@ As a CDI based runtime, Quarkus extensions often make CDI beans available as par
 However, Quarkus DI solution does not support CDI Portable Extensions.
 Instead, Quarkus extensions can make use of various link:cdi-reference.html[Build Time Extension Points].
 
+=== Extension Health Check
+
+Health check is provided via the `quarkus-smallrye-health` extension. It provides both liveness and readiness checks capabilities.
```
kenfinnigan: "Health check is" to "Health checks are"

docs/src/main/asciidoc/extension-authors-guide.adoc
```diff
@@ -874,6 +874,37 @@ As a CDI based runtime, Quarkus extensions often make CDI beans available as par
...
+
+When you write an extension, it can be very handy to provides a default health check for your extension to prevent the developer to do it
```
kenfinnigan: Maybe reword to "When writing an extension, it's beneficial to provide health checks for the extension that can be automatically included without the developer needing to write their own". Or something like that

docs/src/main/asciidoc/extension-authors-guide.adoc
```diff
@@ -874,6 +874,37 @@ As a CDI based runtime, Quarkus extensions often make CDI beans available as par
...
+by itself.
+
+In order to provide a health check, you should do the following:
+
+- Import the `quarkus-smallrye-health` extension as an **optional** dependency in your runtime module so it will not impact the size of the application if
+health check is not needed.
```
kenfinnigan: Should that be "not needed" or "not included"?

docs/src/main/asciidoc/health-guide.adoc
```diff
@@ -296,6 +296,15 @@ error along with the health check response.
         }
 ----
 
+== Extension health checks
+
+Some extension may provides default health checks, including the extension will automatically register its health checks.
```
kenfinnigan: Change "extension may provides" to "extensions may provide".

extensions/agroal/runtime/src/main/java/io/quarkus/agroal/runtime/health/DataSourceHealthCheck.java
```diff
@@ -0,0 +1,62 @@
...
+                if (!valid) {
+                    String data = dataSource.getKey() == DEFAULT_DS ? "validation check failed for the default DataSource"
+                            : "validation check failed for DataSource '" + dataSource.getKey() + "'";
+                    builder.down().withData("reason" + cpt++, data);
+                }
+            } catch (SQLException e) {
+                String data = dataSource.getKey() == DEFAULT_DS
+                        ? "Unable to execute the validation check for the default DataSource: "
+                        : "Unable to execute the validation check for DataSource '" + dataSource.getKey() + "': ";
+                builder.down().withData("reason" + cpt++, data + e.getMessage());
```
kenfinnigan: I still think we should use a key other than "reason" to indicate an exception occurred.

kenfinnigan (changes requested):

dmlloyd (commented):
Seems OK to me: in terms of the SPI structure that is.  I'll let the other reviews stand.

extensions/agroal/runtime/src/main/java/io/quarkus/agroal/runtime/health/DataSourceHealthCheck.java
loicmathieu: OK, I propose to use the name of the datasource or 'quarkus-default-ds' for the default one. 
It's very unlinkly that a user will create a named datasource with 'quarkus-default-ds' as its name.

docs/src/main/asciidoc/datasource-guide.adoc
loicmathieu: done

extensions/agroal/runtime/src/main/java/io/quarkus/agroal/runtime/health/DataSourceHealthCheck.java
```diff
@@ -0,0 +1,62 @@
...
+
+    @Override
+    public HealthCheckResponse call() {
+        HealthCheckResponseBuilder builder = HealthCheckResponse.named("Database connection(s) health check").up();
+        int cpt = 1;
+        for (Map.Entry<String, DataSource> dataSource : dataSources.entrySet()) {
+            try {
+                boolean valid = dataSource.getValue().getConnection().isValid(0);
+                if (!valid) {
+                    String data = dataSource.getKey() == DEFAULT_DS ? "validation check failed for the default DataSource"
```
dmlloyd: String compare should always use `equals`.

docs/src/main/asciidoc/extension-authors-guide.adoc
loicmathieu: done

docs/src/main/asciidoc/extension-authors-guide.adoc
loicmathieu: done

extensions/agroal/runtime/src/main/java/io/quarkus/agroal/runtime/health/DataSourceHealthCheck.java
```diff
@@ -0,0 +1,62 @@
...
+                            : "validation check failed for DataSource '" + dataSource.getKey() + "'";
+                    builder.down().withData("reason" + cpt++, data);
+                }
+            } catch (SQLException e) {
+                String data = dataSource.getKey() == DEFAULT_DS
```
dmlloyd: Another `equals`.

docs/src/main/asciidoc/extension-authors-guide.adoc
loicmathieu: done

docs/src/main/asciidoc/health-guide.adoc
loicmathieu: done

extensions/smallrye-health/runtime/src/main/java/io/quarkus/smallrye/health/runtime/HealthBuildTimeConfig.java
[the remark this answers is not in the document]
dmlloyd: The build time config should normally be in the `deployment` artifact.  Here's the breakdown:

* `BUILD_TIME` (the default) - `deployment` artifact (or possibly `spi` in some future cases)
* `BUILD_AND_RUN_TIME_FIXED` - `runtime` artifact always
* `RUN_TIME` - `runtime` artifact always

extensions/agroal/runtime/src/main/java/io/quarkus/agroal/runtime/health/DataSourceHealthCheck.java
loicmathieu: Arrrhgggg such a newbie error! I'm really ashame as I think I didn't do this error since at least 10 yers !

extensions/smallrye-health/runtime/src/main/java/io/quarkus/smallrye/health/runtime/HealthBuildTimeConfig.java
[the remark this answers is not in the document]
dmlloyd: Also it's 100% OK to have more than one config root use the same name and overlap arbitrarily, whether in the same or in different config phases, as long as no *leaf* configuration values end up with conflicting names.

extensions/agroal/runtime/src/main/java/io/quarkus/agroal/runtime/health/DataSourceHealthCheck.java
loicmathieu: Yep, I saw all of them :=

extensions/smallrye-health/runtime/src/main/java/io/quarkus/smallrye/health/runtime/HealthBuildTimeConfig.java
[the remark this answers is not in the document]
loicmathieu: @kenfinnigan, @dmlloyd will move it to the deployment artifact next week (I'm leaving now : week-end :) ).
As for the name, it's part of a global naming discussed during the design of the feature, I'll wait for @emmanuelbernard review on the design.

extensions/agroal/runtime/src/main/java/io/quarkus/agroal/runtime/health/DataSourceHealthCheck.java
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    @Override
+    public HealthCheckResponse call() {
+        HealthCheckResponseBuilder builder = HealthCheckResponse.named("Database connection(s) health check").up();
+        for (Map.Entry<String, DataSource> dataSource : dataSources.entrySet()) {
+            try {
+                boolean valid = dataSource.getValue().getConnection().isValid(0);
+                if (!valid) {
+                    String data = DEFAULT_DS.equals(dataSource.getKey())  ? "validation check failed for the default DataSource"
```
machi1990: Maybe we can declare a variable for `DEFAULT_DS.equals(dataSource.getKey())`, seems to be done on multiple places.

extensions/smallrye-health/runtime/src/main/java/io/quarkus/smallrye/health/runtime/HealthBuildTimeConfig.java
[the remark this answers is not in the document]
loicmathieu: @kenfinnigan @dmlloyd the build time config has been moved to the deployment module.

extensions/agroal/runtime/src/main/java/io/quarkus/agroal/runtime/health/DataSourceHealthCheck.java
loicmathieu: fixed

extensions/agroal/runtime/src/main/java/io/quarkus/agroal/runtime/health/DataSourceHealthCheck.java
loicmathieu: fixed

extensions/agroal/runtime/src/main/java/io/quarkus/agroal/runtime/health/DataSourceHealthCheck.java
loicmathieu: done

extensions/agroal/runtime/src/main/java/io/quarkus/agroal/runtime/health/DataSourceHealthCheck.java
loicmathieu: For the record, it worked because it's the same string constant that is defined on the class and put as hashmap key so it's valid to use `==` but I'm agree that this should not be used!

loicmathieu:
@kenfinnigan is the last batch of changes OK for you ? Can you advise on what should be done for testing it ? (adding an health check in the test-extension extension ?)

@emmanuelbernard can you validate the design regarding what was discussed for the issue #3107

emmanuelbernard:
I’m swamped man. So I can trust your work or have somebody quickly look it
up. If you have the doc side of this PR i could definitely read it quick
and see if any alarm bell rings :)

kenfinnigan:
@loicmathieu I think it looks good. In terms of a test, I *think* there's an integration-test that uses a database so we could have a test that ensures the Agroal health check is present?

loicmathieu:
@emmanuelbernard I understand that :) that's why I ask to check the design, @kenfinnigan already works with me on the implementation side! 

You can have a look at the following documentation parts:
- Datasource guide: https://github.com/quarkusio/quarkus/pull/3671/files#diff-eb196d770f1a1f55ee5d373bdd05646a
- Health guide: https://github.com/quarkusio/quarkus/pull/3671/files#diff-4e46b3e4681926690297d644df6344fb
- Extension author guide: https://github.com/quarkusio/quarkus/pull/3671/files#diff-38cd571a6836972129eb3e2eb8528382

 @kenfinnigan the main integration test has both datasource and health extension so it will be as easy as adding an assert to it ;). I also integrate an health check to the test-extension extesion, it's up to you wether I push it or not !

loicmathieu:
@kenfinnigan I add a test on the main integration test. I also rebase on master.

FroMage:
@kenfinnigan ball is in your camp :)

kenfinnigan (approved):

extensions/agroal/runtime/src/main/java/io/quarkus/agroal/runtime/health/DataSourceHealthCheck.java
```diff
@@ -0,0 +1,63 @@
...
+        }
+    }
+
+    @Override
+    public HealthCheckResponse call() {
+        HealthCheckResponseBuilder builder = HealthCheckResponse.named("Database connection(s) health check").up();
+        for (Map.Entry<String, DataSource> dataSource : dataSources.entrySet()) {
+            boolean isDefault = DEFAULT_DS.equals(dataSource.getKey());
+            try {
+                boolean valid = dataSource.getValue().getConnection().isValid(0);
```
stuartwdouglas: Don't we need to close the connection here to release it back to the pool?

stuartwdouglas (changes requested):

extensions/agroal/runtime/src/main/java/io/quarkus/agroal/runtime/health/DataSourceHealthCheck.java
loicmathieu: Yes, fixed

docs/src/main/asciidoc/extension-authors-guide.adoc
```diff
@@ -874,6 +874,36 @@ As a CDI based runtime, Quarkus extensions often make CDI beans available as par
...
+- Add a build step in your deployment module that produces a `HealthBuildItem`.
+- Add a way to disable the extension health check via a config item `quarkus.<extension>.health.enabled` that should be enabled by default.
+
+Following is an example from the Agroal extension that provides a `DataSourceHealthCheck` to validate the readiness of a datasource.
+
+[source%nowrap,java]
+----
+@BuildStep
+HealthBuildItem addHealthCheck(AgroalBuildTimeConfig agroalBuildTimeConfig) {
+    if (agroalBuildTimeConfig.healthEnabled) {
```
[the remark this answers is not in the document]
emmanuelbernard: My main point is that all extensions would need to do that : read the flag and do the if. And when we introduce filtering with wildcard and all, it will be even more conplicated.

extensions/smallrye-health/spi/src/main/java/io/quarkus/smallrye/health/deployment/spi/HealthBuildItem.java
```diff
@@ -0,0 +1,15 @@
+package io.quarkus.smallrye.health.deployment.spi;
+
+import io.quarkus.builder.item.MultiBuildItem;
+
+public final class HealthBuildItem extends MultiBuildItem {
+    private Class<?> healthCheckClass;
+
+    public HealthBuildItem(Class<?> healthCheckClass) {
```
emmanuelbernard: I think a health builditem could have a name: e.g. `quarkus.agroal.database_connection` (or something else). this is how we could do selective filtering down the line

emmanuelbernard (changes requested):
Looks good except my remark on reuse of the filtering code

docs/src/main/asciidoc/extension-authors-guide.adoc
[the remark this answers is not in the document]
loicmathieu: I add a boolean on the `HealthBuildItem` so that each extension will only needs to pass it's configuration item to it.

At the begining I wanted to retrieve the configuration item in smallrye-health via `ConfigProvider.getConfig()` but it didn't works, I was not able to retrieve the configuration item during the build step ...

Another solution will be to change the way all this is configured from :
```
quarkus.health.extensions.enabled=true # provided by smallrye-health
quarkus.datasource.health.enabled=false # provided by **agroal**
```

To
```
quarkus.health.extensions.enabled=true # provided by smallrye-health
quarkus.health.extensions.datasource.enabled=false # provided by **smallrye-health**
```

extensions/smallrye-health/spi/src/main/java/io/quarkus/smallrye/health/deployment/spi/HealthBuildItem.java
loicmathieu: As the name is not currently used I didn't incude it already so we will nee to ask for it when filtering will be implemented.

extensions/agroal/runtime/src/main/java/io/quarkus/agroal/runtime/health/DataSourceHealthCheck.java
loicmathieu: @stuartwdouglas is the fix OK for you ?

loicmathieu:
@emmanuelbernard is it OK for you ?

machi1990:
@loicmathieu just a minor comment, can we squash the commits? thanks.

loicmathieu:
@machi1990 I prefere to wait when all is OK before squashing the commits. Still waiting for some feedback ...

docs/src/main/asciidoc/extension-authors-guide.adoc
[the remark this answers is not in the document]
emmanuelbernard: > I add a boolean on the HealthBuildItem so that each extension will only needs to pass it's configuration item to it.

What do you mean by that? You lost me I think.

extensions/smallrye-health/spi/src/main/java/io/quarkus/smallrye/health/deployment/spi/HealthBuildItem.java
emmanuelbernard: I was about to have it now as this will be a breaking change for extensions and wanted to avoid that but if we are unstable on the name concept, that's premature.

docs/src/main/asciidoc/extension-authors-guide.adoc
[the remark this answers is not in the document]
emmanuelbernard: Ah I think I see now, you mean this https://github.com/quarkusio/quarkus/pull/3671/files#diff-38cd571a6836972129eb3e2eb8528382R901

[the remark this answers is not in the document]
emmanuelbernard: Sounds ok, if reading properties from the SR extension is not an option.

extensions/smallrye-health/spi/src/main/java/io/quarkus/smallrye/health/deployment/spi/HealthBuildItem.java
loicmathieu: I was asking myself the same: why not to add it now. 
But I don't think breaking change on extension is really an issue as we can revisit all extension that uses it at the same time. It's an internal breaking change that didn't impact users so it's OK.

loicmathieu:
@machi1990 I squashed some commits but I keep the last one as there was too many conflit.
@kenfinnigan I think that all points are covered by the current implementation so it is good to go.

kenfinnigan:
I think we need @emmanuelbernard and @stuartwdouglas to review the changes they requested

emmanuelbernard:
On 13 Sep 2019, at 15:56, Loïc Mathieu wrote:

> loicmathieu commented on this pull request.
>
>
>
>> @@ -0,0 +1,15 @@
> +package io.quarkus.smallrye.health.deployment.spi;
> +
> +import io.quarkus.builder.item.MultiBuildItem;
> +
> +public final class HealthBuildItem extends MultiBuildItem {
> +    private Class<?> healthCheckClass;
> +
> +    public HealthBuildItem(Class<?> healthCheckClass) {
>
> I was asking myself the same: why not to add it now.
> But I don't think breaking change on extension is really an issue as 
> we can revisit all extension that uses it at the same time. It's an 
> internal breaking change that didn't impact users so it's OK.
>

That is not true, we have external extensions, Camel moved out of the 
Quarkus repo for example and that trend will only accelerate.

loicmathieu:
I didn't thought about external extension.  So do you want to add the name now to avoid a future breaking change later ?

loicmathieu:
> I didn't thought about external extension. So do you want to add the name now to avoid a future breaking change later ?

@emmanuelbernard do you want me to add the name of the extension in the `HealthBuildItem for future use?

loicmathieu:
@kenfinnigan I discuss with @emmanuelbernard on zulip and we agree to add the `configRootName` for future usage.

I also switch to use class name instead of class object because it is the prefered usage across Quarkus extension and, as I understand it, avoid possible unecessary class loading.

I also rebase onto master.

I think that everything is OK so for me it's ready to be merged. If we agree on this I will squash on two commits (one for agroal health check and one for smallrye-heath).

emmanuelbernard (approved):

stuartwdouglas (approved):

PR #3672, opened 2019-08-23
Allow * expression for quarkus:add-extension

### LINKED ISSUE #1234: Allow * expression for quarkus:add-extension
Allow * expression so users can add multiple extensions in one command

For example  adding all MP/SmallRye related extensions - `mvn quarkus:add-extension -Dextensions="SmallRye*"`

Dufgui:
fix #1234

### DIFF
--- a/devtools/common/src/test/java/io/quarkus/cli/commands/AddExtensionsTest.java
+++ b/devtools/common/src/test/java/io/quarkus/cli/commands/AddExtensionsTest.java
@@ -65,6 +65,38 @@ class AddExtensionsTest {
     }
 
     @Test
+    void testRegexpMatches() throws IOException {
+        File pom = new File("target/extensions-test", "pom.xml");
+
+        CreateProjectTest.delete(pom.getParentFile());
+        new CreateProject(new FileProjectWriter(pom.getParentFile()))
+                .groupId("org.acme")
+                .artifactId("add-extension-test")
+                .version("0.0.1-SNAPSHOT")
+                .doCreateProject(new HashMap<>());
+
+        File pomFile = new File(pom.getAbsolutePath());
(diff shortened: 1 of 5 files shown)


devtools/common/src/test/java/io/quarkus/cli/commands/AddExtensionsTest.java
```diff
@@ -64,6 +64,38 @@ void testPartialMatches() throws IOException {
...
+        hasDependency(model, "quarkus-smallrye-metrics");
+        hasDependency(model, "quarkus-smallrye-reactive-messaging-kafka");
+        hasDependency(model, "quarkus-smallrye-health");
+        hasDependency(model, "quarkus-smallrye-openapi");
+        hasDependency(model, "quarkus-smallrye-jwt");
+        hasDependency(model, "quarkus-smallrye-context-propagation");
+        hasDependency(model, "quarkus-smallrye-reactive-type-converters");
+        hasDependency(model, "quarkus-smallrye-reactive-messaging-amqp");
+        hasDependency(model, "quarkus-smallrye-fault-tolerance");
+    }
```
cescoffier: We need to test with more complicated regexps as all the escapes are uncovered.
Also, we need to check with non matching regexp

cescoffier (changes requested):
I've made a comment about the tests.
This should also be documented somewhere.

devtools/common/src/test/java/io/quarkus/cli/commands/AddExtensionsTest.java
cescoffier: What does happen in the case of `**`?

Dufgui:
I add doc, and I fix the conf about the plugin in gradle and missing doc of addExtension
@cescoffier @aloubyansky

cescoffier (approved):

PR #3679, opened 2019-08-23
Add Live Coding support for Kafka Streams

gunnarmorling:


gunnarmorling:
@gsmet, @cescoffier, could either of you review this one? Thanks!

cescoffier (approved):
Looks good to me. Waiting for @stuartwdouglas feedback.

stuartwdouglas:
This looks good to me. Any chance you could add a test using QuarkusDevModeTest?

cescoffier:
@gunnarmorling can you look into implementing a QuarkusDevModeTest?

gunnarmorling:
Yes, I can look into it, but I'm at a F2F this week, so might take a bit.
If there's a release this week, could we get it in then also without that
test and I'll add it later? Would be nice to have it merged.

>

stuartwdouglas:
Ok, we can add the test later.

gsmet:
FYI, if we have a release this week, it will be a pure bugfix release.

This will be in the next one.

gunnarmorling:
Stuart, one question btw. on the dev mode for Undertow: how does it work
that you don't "lose" the current HTTP request that's re-triggering the
reload?

>

stuartwdouglas:
This happens very first thing, so the hot-reload basically happens before it has hit any application logic. When the app is restarted the request continues but into the new application rather than the old one.

gsmet:
@gunnarmorling FYI, change of plan: we will release master tomorrow. So if you want to update the doc, you have until tomorrow morning.

gunnarmorling:
Thanks for the heads-up. Still in the F2F, but I'll see what I can do; in
general, the live reload will work with default settings, but it requires
some option tuning on the client and broker to make it instant and give the
true snappy feedback feeling, as otherwise there'll be some default delay
for reconnecting to the broker.

>

gunnarmorling:
Ok, here's the PR for the doc update:
https://github.com/quarkusio/quarkus/pull/3756

>

Issue #3624, opened 2019-08-21
Request parameters not injected in POST/PUT REST methods


gunnarmorling:
**Describe the bug**
When running the rest-json and hibernate-orm-resteasy quickstarts in dev mode, the `Fruit` param isn't injected, causing an NPE in the ORM example, causing null to be appended to the fruit list in the former. This only happens during dev mode, but not when running the actually built JAR. As per @gsmet, things work as expected on 0.21.1.

**Expected behavior**
The fruit parameter is injected also in dev mode.

**Actual behavior**
The fruit parameter is *not* injected  in dev mode.

**To Reproduce**
check out either example from master/develop, then run `mvn compile quarkus:dev`. Open localhost:8080/fruits.html (index.html for the ORM example) and observe how adding a fruit fails.

**Configuration**
See the repo.

**Screenshots**
n/a

**Additional context**
n/a

gsmet:
@stuartwdouglas as it's working properly with 0.21.1, I suspect it's an issue with the new HTTP layer.

Tagging the issue appropriately.

Reproducing it is easy:
- get the quickstarts
- cd rest-json
- mvn clean compile quarkus:dev
- go to http://0.0.0.0:8080/fruits.html
- try to add a fruit
- you end up with the following error in the log:
```
2019-08-21 17:29:07,450 SEVERE [org.ecl.yas.int.Unmarshaller] (executor-thread-1) Invalid token=EOF at (line no=1, column no=0, offset=-1). Expected tokens are: [CURLYOPEN, SQUAREOPEN, STRING, NUMBER, TRUE, FALSE, NULL]
```
Looks like somehow the correct JSON input is not pushed to the unmarshaller.

PR #3682, opened 2019-08-24
upgrade of kogito (0.3.0) that supports persistence based on infinispan

mswiderski:


mswiderski:
Windows test failure does not seem to be related to his changes.

extensions/kogito/deployment/src/main/java/io/quarkus/kogito/deployment/KogitoAssetsProcessor.java
```diff
@@ -150,10 +247,12 @@ private ApplicationGenerator createApplicationGenerator(Path projectPath, Launch
             throws IOException {
 
         Path srcPath = projectPath.resolve("src");
-        String appPackageName = "org.kie.kogito.app";
+        boolean usePersistence = combinedIndexBuildItem.getIndex()
+                .getClassByName(createDotName("org.kie.kogito.persistence.KogitoProcessInstancesFactory")) != null;
```
cescoffier: You may want to create a constant for this.

extensions/kogito/deployment/src/main/java/io/quarkus/kogito/deployment/KogitoAssetsProcessor.java
```diff
@@ -69,16 +131,50 @@ public void generateModel(ArchiveRootBuildItem root,
...
+            },
+                    launchMode.getLaunchMode(), result, targetClassesPath);
+
+            Index index = kogitoIndexer.complete();
+
+            generatePersistenceInfo(root, generatedBeans, CompositeIndex.create(combinedIndexBuildItem.getIndex(), index),
+                    launchMode, resource);
+
+            reflectiveClass.produce(
+                    new ReflectiveClassBuildItem(true, true, "org.kie.kogito.services.event.ProcessInstanceDataEvent"));
```
cescoffier: Do you need method and field access?

cescoffier (changes requested):
Mostly cosmetic.

We would need to update the quickstart also. Was the QS updated (and tested in native)?

extensions/kogito/deployment/src/main/java/io/quarkus/kogito/deployment/KogitoAssetsProcessor.java
mswiderski: It is used by the jsonb Marshallers I I believe it is needed.

mswiderski:
QuickStart was tested in all modes - there is an issue with dev mode but that is a general issue with json marshalling https://github.com/quarkusio/quarkus/issues/3624

In general QuickStart does not have to be updated

extensions/kogito/deployment/src/main/java/io/quarkus/kogito/deployment/KogitoAssetsProcessor.java
mswiderski: updated

mswiderski:
@cescoffier comments addressed, thanks for review

cescoffier (approved):

PR #3683, opened 2019-08-24
Make CORS Allowed-Headers check case insensitive

### LINKED ISSUE #3681: CORS Allowed-Headers shouldn't be case sensitive
**Describe the bug**
CORS config `quarkus.http.cors.headers` shouldn't be case sensitive, since [HTTP headers are case-insensitive](https://tools.ietf.org/html/rfc2616#section-4.2).

**Expected behavior**
Setting  `quarkus.http.cors.headers=Content-Type` on `application.properties` should cause [pre-flight `OPTIONS` http requests](https://developer.mozilla.org/en-US/docs/Glossary/Preflight_request) with `Access-Control-Request-Headers: content-type` to return a response with header `Access-Control-Allow-Headers: content-type`.

**Actual behavior**
Lower case allow-headers `content-type` is not returned when client asks for camel-case `Content-Type`.

**To Reproduce**
1. [Configure the Quarkus CORS filter](https://quarkus.io/guides/undertow-reference#cors-filter) with `quarkus.http.cors.headers=Content-Typet`
2. Make a pre-flight `OPTIONS` request with the header `Access-Control-Request-Headers: content-type`
3. Notice the missing `Access-Control-Allow-Headers: content-type` in the response headers.

**Configuration**
```properties
#CORS
quarkus.http.cors=true
quarkus.http.cors.headers=Content-Type
```
 - Quarkus version or git rev: 0.21.1

**Additional context**
HTTP headers are case-insensitive, from [RFC 2616 -   Hypertext Transfer Protocol -- HTTP/1.1, section 4.2 - Message Headers](https://tools.ietf.org/html/rfc2616#section-4.2),.

gwenneg:
Fixes #3681

### DIFF
--- a/extensions/vertx-web/runtime/src/main/java/io/quarkus/vertx/web/runtime/cors/CORSFilter.java
+++ b/extensions/vertx-web/runtime/src/main/java/io/quarkus/vertx/web/runtime/cors/CORSFilter.java
@@ -1,6 +1,9 @@
 package io.quarkus.vertx.web.runtime.cors;
 
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Objects;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
 import io.vertx.core.Handler;
@@ -12,6 +15,8 @@ import io.vertx.ext.web.RoutingContext;
 
 public class CORSFilter implements Handler<RoutingContext> {
 
+    private static final Pattern COMMA_SEPARATED_SPLIT_REGEX = Pattern.compile("\\s*,\\s*");
+
     // This is set in the recorder at runtime.
     // Must be static because the filter is created(deployed) at build time and runtime config is still not available
     final CORSConfig corsConfig;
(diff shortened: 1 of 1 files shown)


extensions/vertx-web/runtime/src/main/java/io/quarkus/vertx/web/runtime/cors/CORSFilter.java
```diff
@@ -24,9 +24,10 @@ private void processRequestedHeaders(HttpServerResponse response, String request
         if (corsConfig.headers.isEmpty()) {
             response.headers().set(HttpHeaders.ACCESS_CONTROL_ALLOW_HEADERS, requestedHeaders);
         } else {
+            String lcRequestedHeaders = requestedHeaders.toLowerCase();
             final String validRequestHeaders = corsConfig.headers
                     .stream()
-                    .filter(requestedHeaders::contains)
+                    .filter(configHeader -> lcRequestedHeaders.contains(configHeader.toLowerCase()))
                     .collect(Collectors.joining(","));
```
gwenneg: @gsmet Is it OK to keep this lambda here?

machi1990: Should we lower case requested headers too?

gwenneg: We should and it's already done at line 27 :)

machi1990: Oops, my bad. Please Ignore my comment, I missed that part. Thanks :-)

gsmet: I think I would have lowercased the config headers at the init of the filter once and for all.

As it's runtime code and the lambda is not that useful, I would have removed it.

And, if I'm not mistaken, the current code is fragile and unsafe. I think we should split the requested headers before doing any comparison. Otherwise, you could have a partial match.

And I would probably have done the same for the methods.

gsmet (changes requested):
Added some comments. Not directly related to your change but while we're at it...

extensions/vertx-web/runtime/src/main/java/io/quarkus/vertx/web/runtime/cors/CORSFilter.java
gwenneg: I did lowercase the headers in the filter constructor at first and then saw that `CORSConfig` was part of `HttpConfiguration` which is annotated with `@ConfigRoot(phase = ConfigPhase.RUN_TIME)`. Knowing that, is it really OK to lowercase once at the filter init?

gsmet: Hmmm, you’re right.

gsmet: Yup, that's exactly why I'm not happy with the current code.

gwenneg: And it's about to be changed, I'll update the PR as soon as I find some time :)

gwenneg:
The lambdas and unsafe lists comparisons are gone.

geoand:
@gsmet I think we longer need the `backport` label, correct?

extensions/vertx-web/runtime/src/main/java/io/quarkus/vertx/web/runtime/cors/CORSFilter.java
```diff
@@ -12,6 +15,8 @@
 
 public class CORSFilter implements Handler<RoutingContext> {
 
+    private static final String COMMA_SEPARATED_SPLIT_REGEX = "\\s*,\\s*";
```
gsmet: You need to compile the regexp, otherwise it will be compiled each time you use it.

extensions/vertx-web/runtime/src/main/java/io/quarkus/vertx/web/runtime/cors/CORSFilter.java
```diff
@@ -20,34 +25,43 @@ public CORSFilter(CORSConfig corsConfig) {
...
 
-            final String validRequestedMethods = corsConfig.methods
-                    .stream()
-                    .filter(method -> requestedMethods.contains(method.name()))
-                    .map(HttpMethod::name)
-                    .collect(Collectors.joining(","));
+            List<String> validRequestedMethods = new ArrayList<>();
+            for (HttpMethod configMethod : corsConfig.methods) {
+                if (requestedMethods.contains(configMethod.name())) {
+                    validRequestedMethods.add(configMethod.name());
```
gsmet: Shouldn't we be case insensitive too?

gsmet (changes requested):
Added 2 comments inline.

extensions/vertx-web/runtime/src/main/java/io/quarkus/vertx/web/runtime/cors/CORSFilter.java
gwenneg: Not according to the RFC 2616: https://tools.ietf.org/html/rfc2616#section-5.1.1

> The method is case-sensitive.

gsmet: OK, I would have been more lenient but I can live with it :).

gwenneg: I'm fine with being lenient :) I'll change this while adding the regexp compilation.

gwenneg:
There's an author issue with my last commit, I'll fix that in a couple hours.

gsmet (approved):

gsmet:
OK, please post a message when fixed!

gwenneg:
Author issue solved!

PR #3613, opened 2019-08-21
Do not delete generated files

machi1990:
Fixes #3448 - generated files that are not within module source paths returned a null sourceFilePath. We should not delete the corresponding generated .class file in this case.

/cc @antoniomacri @masini this should fix the issue you encountered. 

/cc @gsmet

PR #3686, opened 2019-08-25
fixes devmode with generated sources

### LINKED ISSUE #3450: Generated-sources not resolved in dev mode
hi there,

**Describe the bug**
Using annotation processor e.g. mapstruct, the generated sources end up in `target/generated-sources` folder. Compilation works fine, however during dev mode, these generated classes are not present in classpath, causing ClassNotFoundError.
This seems to be a regression from version 0.19.0 as it works fine there.

**Expected behavior**
Generated classes should be present in classpath in dev mode.

**Actual behavior**
Classes from `generated-sources` produced by annotation processor are not present in classpath in dev mode.

**To Reproduce**
You can use the Mapstruct-quarkus example project: https://github.com/mapstruct/mapstruct-examples/tree/master/mapstruct-quarkus and test it. With version >=`0.19.0` it works fine, bumping the version up to `0.20.0` breaks the dev mode.

matus-m:
Hi,
this is my first MR in Quarkus, so be gentle :)

 **what does this MR fix?** 

Version 0.20 caused a regression in devmode(#3450, due to changes in class/source change detection) when annotation processors are used. 

 **How does it fix it?** 

This MR fixes it by including `generatedSourcesDirectory` folder to the `ModuleInfo#sourcePaths` if that directory exists. Location of the folder is determined from maven compiler plugin config, or a default location as [per maven docs ](https://maven.apache.org/plugins/maven-compiler-plugin/compile-mojo.html#generatedSourcesDirectory)

A new IT test case has been added, that uses Mapstruct annotation processor.

Fixes #3450


geoand:
Thanks a lot for the contribution!

I'll check it out soon

geoand:
One question, you have tested this with Gradle?

yntelectual:
Hey, no, I have not tested it with Gradle, looking at the sources, it seems like it will be affected as well, but the way annotation processors are handled in Gradle seems dependent on Gradle version. Will try to look into that.

geoand:
Thanks!

geoand:
Also, it looks like the PR needs a rebase now

yntelectual:
Ok , as it turns out, there has been a parallel MR dealing with change detection #3613 - that just ignores that classes for which source is not found. I guess this MR can be closed then... 
Tested it with gradle and maven on current Master build and classes from generated sources were preserved.

geoand:
OK, thanks for the effort regardless!

PR #3700, opened 2019-08-26
Arc - SubclassGenerator needlessly uses field instead of local variable.

manovotn:


independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/SubclassGenerator.java
```diff
@@ -140,14 +141,9 @@ protected FieldDescriptor createConstructor(ClassOutput classOutput, BeanInfo be
 
         // we build a map for each interceptor instance created, so that they are shared
         // Map<String, Object> where InjectableInterceptor.getIdentifier() is key and Object is instance of the interceptor for this bean
-        FieldCreator interceptorInstanceMap = subclass.getFieldCreator("interceptorInstanceMap", Map.class.getName())
-                .setModifiers(ACC_PRIVATE | ACC_FINAL);
+        AssignableResultHandle interceptorInstanceMap = constructor.createVariable(Map.class);
```
mkouba: I don't think you need an `AssignableResultHandle` here, `ResultHandle interceptorInstanceMap = constructor.newInstance(MethodDescriptor.ofConstructor(HashMap.class)` should work.

manovotn: Isn't that the same thing? According to Gizmo doc, `constructor.createVariable()` will give me local variable. Or am I missing something?

mkouba: Yes, the output is almost the same. But when using the regular `ResultHandle` a final variable is created and it's just one line - there is no need to invoke `constructor.assign()` afterwards ;-).

manovotn: Ah. got it! Will do :)

mkouba (approved):

PR #3703, opened 2019-08-26
Fix issue with runtime config being consumed at build time

### LINKED ISSUE #3677: Run time configuration consumed from VertxWebProcessor.kubernetes()
**Describe the bug**
A build step (`VertxWebProcessor#kubernetes`) is consuming a run time configuration object during build time to determine the HTTP listen port.  This is not allowed because run time values cannot be resolved (#3516, #3637) or accessed during build time.  The run time configuration objects can only correctly be used to pass as proxies to recorders for run time actions (though in this case, it is quite hard to detect invalid accesses to config fields because of the decision to opt for field injection instead of using e.g. a proxy interface for configuration values, so I might still look into directly injecting the config roots into the recorder proxies and drop support for any injection of run time config objects into build steps).

**Additional context**
Blocks #3516/#3637.

geoand:
The Kubernetes resource generation only occurs at build time so it suffices
to read the http port value from MP Config

Fixes: #3677

### DIFF
--- a/extensions/vertx-web/deployment/src/main/java/io/quarkus/vertx/web/deployment/VertxWebProcessor.java
+++ b/extensions/vertx-web/deployment/src/main/java/io/quarkus/vertx/web/deployment/VertxWebProcessor.java
@@ -11,6 +11,7 @@ import java.util.stream.Collectors;
 
 import javax.inject.Singleton;
 
+import org.eclipse.microprofile.config.ConfigProvider;
 import org.jboss.jandex.AnnotationInstance;
 import org.jboss.jandex.DotName;
 import org.jboss.jandex.MethodInfo;
@@ -91,8 +92,12 @@ class VertxWebProcessor {
     }
 
     @BuildStep
-    public void kubernetes(HttpConfiguration config, BuildProducer<KubernetesPortBuildItem> portProducer) {
-        portProducer.produce(new KubernetesPortBuildItem(config.port, "http"));
+    public void kubernetes(BuildProducer<KubernetesPortBuildItem> portProducer) {
+        // we can't use HttpConfiguration since the port is a runtime configuration value
+        // therefore we just read the value from MP Config - this is fine since the value is only ever used to create
+        // Kubernetes resources at build time
+        portProducer.produce(new KubernetesPortBuildItem(
+                ConfigProvider.getConfig().getOptionalValue("quarkus.http.port", Integer.class).orElse(8080), "http"));
     }
 
     @BuildStep
--- a/extensions/kubernetes/deployment/pom.xml
+++ b/extensions/kubernetes/deployment/pom.xml
@@ -18,6 +18,10 @@
             <artifactId>quarkus-kubernetes-spi</artifactId>
         </dependency>
         <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-arc-deployment</artifactId>
+        </dependency>
+        <dependency>
             <groupId>io.dekorate</groupId>
             <artifactId>kubernetes-annotations</artifactId>
             <classifier>noapt</classifier>
--- a/extensions/kubernetes/deployment/src/main/java/io/quarkus/kubernetes/deployment/KubernetesProcessor.java
+++ b/extensions/kubernetes/deployment/src/main/java/io/quarkus/kubernetes/deployment/KubernetesProcessor.java
@@ -1,5 +1,7 @@
 package io.quarkus.kubernetes.deployment;
 
+import static io.quarkus.gizmo.MethodDescriptor.ofMethod;
+
 import java.io.IOException;
 import java.io.UnsupportedEncodingException;
 import java.nio.file.Files;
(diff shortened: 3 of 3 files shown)


stuartwdouglas (approved):

extensions/vertx-web/deployment/src/main/java/io/quarkus/vertx/web/deployment/VertxWebProcessor.java
```diff
@@ -91,8 +92,12 @@ FeatureBuildItem feature() {
...
 
     @BuildStep
-    public void kubernetes(HttpConfiguration config, BuildProducer<KubernetesPortBuildItem> portProducer) {
-        portProducer.produce(new KubernetesPortBuildItem(config.port, "http"));
+    public void kubernetes(BuildProducer<KubernetesPortBuildItem> portProducer) {
+        // we can't use HttpConfiguration since the port is a runtime configuration value
+        // therefore we just read the value from MP Config - this is fine since the value is only ever used to create
+        // Kubernetes resources at build time
+        portProducer.produce(new KubernetesPortBuildItem(
+                ConfigProvider.getConfig().getOptionalValue("quarkus.http.port", Integer.class).orElse(8080), "http"));
```
dmlloyd: There's no guarantee that the value will be present at build time.

geoand: Doesn't the use of `Optional` cover that? Am I missing something?

dmlloyd: If I built without `quarkus.http.port` and then run with it set to 9999, then it will fail.  If I know that I need one value for build and another for run then maybe it's better to have a separate property that can be set.  You could also add in a run time check that prints a log message if the ports don't match up, if you wanted to.

geoand: That is certainly true, however it won't exactly be a failure - basically the Kubernetes resource file will not match what is being used at runtime, but there is no way to match that anyway. 

I would assume that users already know this since they know that there Kubernetes files are being generated at build time.

Adding a kubernetes port build item (and configuration to go with it) won't really get users much IMHO, since it won't be able to change anything at runtime either - meaning they are in the same spot as if they didn't have it.

Am I missing something?

geoand: Absolutely fair point. 
However wouldn't the same logic apply if there were 2 config properties, that need to change in tandem (and one only being configurable at build time)? Wouldn't the user need to infer that `quarkus.http.port` and `quarkus.kubernetes.port` need to point  to the same value?
Now of course like you said you could have a runtime check (and probably a build time check as well), but I am not convinced it's a better experience for users.

I'll leave it up to you however, if you think it should be 2 different config items, then I'll do that :)

geoand: Another thing if we add 2 different config items, is what would be the name of the config? Asking because `quarkus.kubernetes` is already being used by `KubernetesConfig` in the `quarkus-kubernetes` extension

geoand: It just seems weird to add a kubernetes port property (which is exactly the same as the regular http port) in the vertx configuration.
I know that I myself as a user would question why I have to do:

```
quarkus.http.port=9090
quarkus.http.kubernetes-port=9090
```

instead of just 

```
quarkus.http.port=9090
```

which would cover both cases. And of course it could be documented that changing the port at runtime would have no effect on the generated configuration.

Furthemore I think users will really not like having the port property being a build time only property. I personally don't mind, but I think a lot of users would.

So, do you want me to go ahead with the 2 properties?

geoand: Just thought of another solution which I think would give the best of both worlds:

We keep the single port, but at build time we also generate a class that is run at application startup and checks whether the runtime port is the different than the build time one and if so log a warning about the generates Kubernetes resources not being in sync with the runtime config.

WDYT @dmlloyd ?

geoand: OK, I'll proceed to code it then

geoand: Need to step out for a few hours. I'll continue when and push the updates to the PR when I get back

geoand: @dmlloyd PR updated

stuartwdouglas:
You can just do it like this: https://github.com/quarkusio/quarkus/pull/3715

As @Record is optional this method will only be invoked if the kubernetes extension is actually present.

geoand:
Didn't know that record can be used in the absence of a extension in that fashion, cool!

PR #3707, opened 2019-08-26
 ArC - fix generics processing when collecting a bean type closure

### LINKED ISSUE #3619: Issue with generic bean injection

mkouba:
- resolves #3619

### DIFF
--- a/independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/Beans.java
+++ b/independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/Beans.java
@@ -41,7 +41,7 @@ final class Beans {
     static BeanInfo createClassBean(ClassInfo beanClass, BeanDeployment beanDeployment, InjectionPointModifier transformer) {
         Set<AnnotationInstance> qualifiers = new HashSet<>();
         List<ScopeInfo> scopes = new ArrayList<>();
-        Set<Type> types = Types.getClassBeanTypeClosure(beanClass, Collections.emptyMap(), beanDeployment);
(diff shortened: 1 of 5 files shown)


manovotn (approved):
This looks much better then my previous attempt!

manovotn:
@jmartisk shouldn't have troubles with this PR as one of my added tests (which is cherry-picked here) copies the scenario from TCK.

PR #3708, opened 2019-08-26
fixed typo in rest client guide - javax.inject.Singleton

mswiderski:
While working on RestClient based sample I run into this warning being issued during augmentation

```
[WARNING] [io.quarkus.restclient.deployment.RestClientProcessor] Unsupported default scope java.inject.Singleton provided for rest client org.acme.travels.rest.UsersRemoteService. Defaulting to @Dependent.
```

looks like this has changed so I made the change in the guide so people do not get this warning while using RestClient.

### DIFF
--- a/docs/src/main/asciidoc/rest-client-guide.adoc
+++ b/docs/src/main/asciidoc/rest-client-guide.adoc
@@ -185,7 +185,7 @@ The name of the property needs to follow a certain convention which is best disp
 ----
 # Your configuration properties
 org.acme.restclient.CountriesService/mp-rest/url=https://restcountries.eu/rest # // <1>
-org.acme.restclient.CountriesService/mp-rest/scope=java.inject.Singleton # // <2>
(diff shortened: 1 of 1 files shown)


stuartwdouglas:
There is a typo is that it is java.inject instead of javax.inject. Singleton scope is supported, can you modify the PR to fix the typo?

mswiderski:
@stuartwdouglas thanks for spotting that ... updated

stuartwdouglas (approved):

viniciusfcf:
The site https://quarkus.io/guides/rest-client still shows "Supported scope values are @Singleton". It should be ApplicationScope, right @mswiderski

mswiderski:
Singleton is a valid scope and the fix made here was to adjust the package name used

juangon:
FYI I am still reproducing this warning, and I have `javax.inject.Singleton` as configured scope in application.properties

PR #3731, opened 2019-08-27
Arc - toString() method invocation should be delegated to contextual …

### LINKED ISSUE #3722: Arc - client proxy should delegate toString() to contextual instance
According to CDI spec, methods on proxies inherited from `Object` are undefined except for `toString()`. See [this](https://docs.jboss.org/cdi/spec/2.0/cdi-spec.html#client_proxy_invocation) chapter. In combination with chapter 5.4 this indirectly requires `toString()` to be delegated to contextual instances.

ATM Arc doesn't do that (in comparison, Weld does).

@mkouba @jmartisk FYI

manovotn:
…instance.

Fixes #3722

### DIFF
--- a/independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/ClientProxyGenerator.java
+++ b/independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/ClientProxyGenerator.java
@@ -134,7 +134,10 @@ public class ClientProxyGenerator extends AbstractGenerator {
              * as it just works, and the reflection case cannot be true since it's not possible to have
              * non-public default interface methods.
              */
-            if (isInterface) {
+            if (Methods.isObjectToString(method)) {
+                // Always use invokevirtual and the original descriptor for java.lang.Object#toString()
+                ret = forward.invokeVirtualMethod(originalMethodDescriptor, delegate, params);
+            } else if (isInterface) {
                 ret = forward.invokeInterfaceMethod(method, delegate, params);
             } else if (isReflectionFallbackNeeded(method, targetPackage)) {
                 // Reflection fallback
--- a/independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/Methods.java
+++ b/independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/Methods.java
@@ -107,6 +107,10 @@ final class Methods {
         return false;
     }
 
+    static boolean isObjectToString(MethodInfo method) {
+        return method.declaringClass().name().equals(DotNames.OBJECT) && method.name().equals(TO_STRING);
+    }
+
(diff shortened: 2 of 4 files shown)


mkouba (approved):

manovotn:
@stuartwdouglas @dmlloyd I can see some native failures here and I have no idea what to do about that. Could you please help?

The stack from CI goes like this:
> Caused by: com.oracle.graal.pointsto.constraints.UnresolvedElementException: Discovered unresolved method during parsing: java.lang.Object.toString(). To diagnose the issue you can use the --allow-incomplete-classpath option. The missing method is then reported at run time when it is accessed the first time.

mkouba:
Note that the problem occurs when we override `toString()` on a client proxy for a producer method that returns an interface. The method looks like:
```java
public String toString() {
   return this.arc$delegate().toString();
}
```
Where `arc$delegate()` returns an interface such as `io.vertx.ext.web.Router`.

mkouba:
So it turns out we use `invokeinterface` to invoke the `toString()` upon an interface which seems to work in JVM but is probably not legal in SVM. I have a fix so I'll try to add a commit to this PR.

mkouba:
Does anybody have a clue what this means?
```
[INFO] [io.quarkus.creator.phase.nativeimage.NativeImagePhase] docker run -v /home/vsts/work/1/s/integration-tests/hibernate-orm-panache/target:/project:z --rm --user 1001:117 quay.io/quarkus/ubi-quarkus-native-image:19.2.0 -J-Djava.util.logging.manager=org.jboss.logmanager.LogManager -J-Dio.netty.leakDetection.level=DISABLED -J-Dcom.sun.xml.bind.v2.bytecode.ClassTailor.noOptimize=true -J-Dvertx.disableDnsResolver=true -J-Dio.netty.noUnsafe=true --initialize-at-build-time= -H:InitialCollectionPolicy=com.oracle.svm.core.genscavenge.CollectionPolicy$BySpaceAndTime -jar quarkus-integration-test-hibernate-orm-panache-999-SNAPSHOT-runner.jar -J-Djava.util.concurrent.ForkJoinPool.common.parallelism=1 -H:FallbackThreshold=0 -H:+ReportExceptionStackTraces -H:+PrintAnalysisCallTree -J-Xmx6g -H:-AddAllCharsets -H:EnableURLProtocols=http,https --enable-all-security-services -H:-SpawnIsolates -H:+JNI --no-server -H:-UseServiceLoaderFeature -H:+StackTrace
Unable to find image 'quay.io/quarkus/ubi-quarkus-native-image:19.2.0' locally
docker: Error response from daemon: Get https://quay.io/v2/: EOF.
See 'docker run --help'.
```

gsmet:
@mkouba I would say it's just a transient error of Quay not responding correctly.

PR #3734, opened 2019-08-27
Make better use of the maven repo cache in CI

stuartwdouglas:


geoand (approved):

Sparow199:
I tried it on Ubuntu: mvn - B -f path/to/pom.xml dependency:go-offline It works without any issue.

stuartwdouglas:
Did you try it with a clean maven repo, and the same settings.xml we are using in CI?

For some reason I think it is trying to download a plugin that is only available from the jboss nexus from central, even though our settings should ignore central.

Issue #1673, opened 2019-03-25
Native compilation doesn't like space in project directory


jcordes73:
When creating a project directory with a space in it, native compilation fails, see steps to reproduce below:

mkdir "dir with space"
cd dir\ with\ space/
mvn io.quarkus:quarkus-maven-plugin:0.12.0:create \
    -DprojectGroupId=quarkus.io \
    -DprojectArtifactId=space-quickstart \
    -DprojectVersion=1.0-SNAPSHOT \
    -DclassName="io.quarkus.SpaceResource"
mvn compile quarkus:dev
mvn package -Pnative

The following fixes the issue:

mv dir\ with\ space/ dir_with_space/

dmlloyd:
This seems related to #1402/#1607/#1438/#1030/etc.

rsvoboda:
The first error is `Error: Class initialization failed: io.quarkus.undertow.runtime.graal.URLResourceSubstitution$ResourceInfo`
Caused by: `java.io.FileNotFoundException: /Users/rsvoboda/tmp/dir%20with%20space/target/lib/io.smallrye.smallrye-config-1.3.5.jar (No such file or directory)` in my case.

I see same error when the root directory is named:
 * app dir ect ory
 * app#dir
 * žluťoučký-kůň
 * app!@#%dir ect ory
 
For the following cases I have successful run of compilation + dev mode + native build
 * app@dir 
 * app(dir)ectory

With `app?dir` case I don't get through package phase
```
14:55:46 [INFO] 
14:55:46 [INFO] --- quarkus-maven-plugin:999-SNAPSHOT:build (default) @ space-quickstart ---
14:55:46 [INFO] ------------------------------------------------------------------------
14:55:46 [INFO] BUILD FAILURE
14:55:46 [INFO] ------------------------------------------------------------------------
14:55:46 [INFO] Total time: 7.579 s
14:55:46 [INFO] Finished at: 2019-03-27T14:55:46+01:00
14:55:46 [INFO] ------------------------------------------------------------------------
14:55:46 [ERROR] Failed to execute goal io.quarkus:quarkus-maven-plugin:999-SNAPSHOT:
  build (default) on project space-quickstart: Failed to build a runnable JAR: 
  Failed to build a runner jar: URI has a query component -> [Help 1]
```

dmlloyd:
Again it's going to be wrong URI handling somewhere.

Here's the rules for class paths:

[...]

PR #3735, opened 2019-08-27
Correctly add class path entries

### LINKED ISSUE #3592: Hot reload fails with InvalidPathException on Windows
Starting with Quarkus 0.21.0 hot reload failes with InvalidPathException (at least) on Windows.

When a Quarkus application runs in dev mode, i. e. using mvn quarkus:dev, a REST call will trigger a hot reload, if one of the source files has changed before. That worked like a charm up to version 0.20.0. With 0.21.0 the REST response contains an error message:

`<header>
    <h1 class="container">Error restarting Quarkus</h1>
    <div class="exception-message">
        <h2 class="container">java.nio.file.InvalidPathException: Illegal char &lt;:&gt; at index 2: /C:/Users/dw/.m2/repository/io/quarkus/arc/arc/0.21.0/arc-0.21.0.jar</h2>
    </div>
</header>`

As the message complains about the ':' in the file path, I guess, the problem arises on Windows only.

The bug can easily be reproduced:
- Use any of the applications from the guide, e. g getting-started
- Make sure that quarkus.version in pom.xml is 0.21.0
- Build the application and start dev mode (mvn clean package quarkus:dev)
- Make some code change
- Call a REST endpoint (curl http://localhost:8080/hello)

My environment:
- Windows 10 Pro
- AdoptOpenJDK 8 (jdk8u202-b08) and 11 (jdk-11.0.3+7)

dmlloyd:
Fixes #3592.  Windows URI paths are not correctly extracted.

### DIFF
--- a/devtools/maven/src/main/java/io/quarkus/maven/DevMojo.java
+++ b/devtools/maven/src/main/java/io/quarkus/maven/DevMojo.java
@@ -11,8 +11,6 @@ import java.net.MalformedURLException;
 import java.net.Socket;
 import java.net.URI;
 import java.net.URL;
-import java.net.URLDecoder;
-import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
@@ -48,6 +46,7 @@ import io.quarkus.bootstrap.model.AppModel;
 import io.quarkus.bootstrap.resolver.BootstrapAppModelResolver;
 import io.quarkus.bootstrap.resolver.maven.MavenArtifactResolver;
 import io.quarkus.bootstrap.resolver.maven.workspace.LocalProject;
+import io.quarkus.bootstrap.util.PropertyUtils;
 import io.quarkus.dev.DevModeContext;
 import io.quarkus.dev.DevModeMain;
 import io.quarkus.maven.components.MavenVersionEnforcer;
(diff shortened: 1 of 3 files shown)


geoand:
Seems like imports are not sorted

geoand:
If you are using `IntelliJ`, it might happing on commit where the IDE tries to optimize imports before doing the actual commit. It's happened to me plenty of times when I forgot the relevant setting enabled

cescoffier (approved):

stuartwdouglas (dismissed):
Failures appear related

gsmet:
@dmlloyd any chance you could have another look? It would be nice to have this one in tomorrow's release! Thanks!

dmlloyd:
Updated the comment here with relevant info: https://github.com/quarkusio/quarkus/issues/1673#issuecomment-477205320

integration-tests/kogito-maven/src/test/java/io/quarkus/kogito/maven/it/KogitoDevModeIT.java
```diff
@@ -15,7 +15,7 @@
     @Test
     public void testThatTheKogitoApplicationRuns() throws MavenInvocationException, IOException {
         testDir = initProject("projects/simple-kogito", "projects/project-classic-run-kogito");
-        run();
+        run("-e");
```
gsmet: Do we want to keep that?

devtools/maven/src/main/java/io/quarkus/maven/DevMojo.java
```diff
@@ -430,6 +433,11 @@ private void addToClassPaths(StringBuilder classPathManifest, DevModeContext cla
             throw new RuntimeException(e);
         }
         String path = uri.getRawPath();
+        if (PropertyUtils.isWindows()) {
+            if (path.length() > 2 && Character.isLetter(path.charAt(0)) && path.charAt(1) == ':') {
+                path = "/" + path;
+            }
+        }
```
gsmet: Wondering if we should have an util for that as we have the exact same dark magic code in 2 places? Not a blocker, just a though.

gsmet (changes requested):
Added a couple of questions/comments.

integration-tests/kogito-maven/src/test/java/io/quarkus/kogito/maven/it/KogitoDevModeIT.java
dmlloyd: Yes IMO.  It doesn't do any harm, and you always want to add it when things go wrong.

devtools/maven/src/main/java/io/quarkus/maven/DevMojo.java
dmlloyd: Maybe at some point.  There are probably a lot of things that can be cleaned up in this code.

dmlloyd:
@gsmet, are we good to go?

gsmet (approved):

PR #3744, opened 2019-08-27
ci(Pipeline): Simplify maven cache cmd

aoudiamoncef:


stuartwdouglas:
I already tried this, I am not sure why it does not work. 

I have actually changed this in https://github.com/quarkusio/quarkus/pull/3734 so that this step only runs if the bom.xml file has been modified, which is a big time saving.

Sparow199:
I think that the complete command will be :

mvn -f bom/runtime/pom.xml dependency:go-offline

I tried it on my Ubuntu with the same configuration as mvn package and the both gave me the same result.

stuartwdouglas:
That won't download all the plugins needed for the build/testing though, as the bom does not have any of them?

Sparow199:
I'm in the same situation as you. dependency:go-offline doesn't work for the whole project. 
Thez combo of:
dependency:resolve-plugins -> got an issue but with -DexcludeTransitive=true, it works
+
dependency:tree -> It downloads the majority of dependencies but not 100%

I concluded that package is the the best solution because Apache Maven Dependency Plugin have some issues and it doesn't feet Quarkus use cases.
Feel free to close this PR.
Thanks

PR #3746, opened 2019-08-28
 Update to vertx 3.8.1

cescoffier:
This PR updates the Vert.x version to 3.8.1 as well as dependencies using Vert.x.
It also updates the smallrye artifacts Ids that have been renamed.
Finally, it provides support for the back pressure control on emitter in Reactive Messaging.

It supersedes #3711.

### DIFF
--- a/bom/runtime/pom.xml
+++ b/bom/runtime/pom.xml
@@ -38,9 +38,9 @@
         <smallrye-fault-tolerance.version>1.0.1</smallrye-fault-tolerance.version>
         <smallrye-jwt.version>2.0.1</smallrye-jwt.version>
         <smallrye-context-propagation.version>1.0.11</smallrye-context-propagation.version>
-        <smallrye-reactive-streams-operators.version>1.0.6</smallrye-reactive-streams-operators.version>
-        <smallrye-converter-api.version>1.0.4</smallrye-converter-api.version>
-        <smallrye-reactive-messaging.version>0.0.10</smallrye-reactive-messaging.version>
+        <smallrye-reactive-streams-operators.version>1.0.9</smallrye-reactive-streams-operators.version>
+        <smallrye-converter-api.version>1.0.9</smallrye-converter-api.version>
+        <smallrye-reactive-messaging.version>1.0.3</smallrye-reactive-messaging.version>
         <swagger-ui.version>3.20.9</swagger-ui.version>
         <javax.activation.version>1.1.1</javax.activation.version>
         <javax.inject.version>1</javax.inject.version>
@@ -100,7 +100,7 @@
         <wildfly-elytron.version>2.0.0.Alpha4</wildfly-elytron.version>
         <jboss-modules.version>1.8.7.Final</jboss-modules.version>
         <jboss-threads.version>3.0.0.Beta5</jboss-threads.version>
-        <vertx.version>3.8.0</vertx.version>
+        <vertx.version>3.8.1</vertx.version>
         <httpclient.version>4.5.9</httpclient.version>
         <httpcore.version>4.4.11</httpcore.version>
         <httpasync.version>4.1.4</httpasync.version>
@@ -123,7 +123,7 @@
         <reactive-streams.version>1.0.2</reactive-streams.version>
         <test-containers.version>1.12.0</test-containers.version>
         <jboss-logging.version>3.3.2.Final</jboss-logging.version>
-        <axle-client.version>0.0.8</axle-client.version>
+        <axle-client.version>0.0.9</axle-client.version>
         <kafka-clients.version>2.2.1</kafka-clients.version>
         <kafka2.version>2.2.1</kafka2.version>
         <debezium.version>0.9.5.Final</debezium.version>
@@ -412,7 +412,7 @@
             </dependency>
             <dependency>
                 <groupId>io.quarkus</groupId>
-                <artifactId>quarkus-smallrye-reactive-messaging-amqp</artifactId>
+                <artifactId>quarkus-smallrye-reactive-amqp</artifactId>
                 <version>${project.version}</version>
             </dependency>
             <dependency>
@@ -1829,17 +1829,17 @@
 
             <dependency>
                 <groupId>io.smallrye.reactive</groupId>
-                <artifactId>smallrye-reactive-messaging-provider-1.0</artifactId>
+                <artifactId>smallrye-reactive-messaging-provider</artifactId>
                 <version>${smallrye-reactive-messaging.version}</version>
             </dependency>
             <dependency>
                 <groupId>io.smallrye.reactive</groupId>
-                <artifactId>smallrye-reactive-messaging-mqtt-1.0</artifactId>
+                <artifactId>smallrye-reactive-messaging-mqtt</artifactId>
                 <version>${smallrye-reactive-messaging.version}</version>
             </dependency>
             <dependency>
                 <groupId>io.smallrye.reactive</groupId>
-                <artifactId>smallrye-reactive-messaging-kafka-1.0</artifactId>
+                <artifactId>smallrye-reactive-messaging-kafka</artifactId>
                 <version>${smallrye-reactive-messaging.version}</version>
                 <exclusions>
                     <exclusion>
@@ -1866,7 +1866,7 @@
             </dependency>
             <dependency>
                 <groupId>io.smallrye.reactive</groupId>
-                <artifactId>smallrye-reactive-messaging-amqp-1.0</artifactId>
+                <artifactId>smallrye-reactive-messaging-amqp</artifactId>
                 <version>${smallrye-reactive-messaging.version}</version>
                 <exclusions>
                     <exclusion>
@@ -1950,7 +1950,7 @@
             </dependency>
             <dependency>
                 <groupId>io.smallrye.reactive</groupId>
-                <artifactId>smallrye-reactive-streams-operators-1.0</artifactId>
+                <artifactId>smallrye-reactive-streams-operators</artifactId>
                 <version>${smallrye-reactive-streams-operators.version}</version>
             </dependency>
             <dependency>
--- a/bom/runtime/pom.xml
+++ b/bom/runtime/pom.xml
@@ -133,7 +133,7 @@
         <aws-lambda-java.version>1.1.0</aws-lambda-java.version>
         <aws-lambda-java-events.version>2.2.5</aws-lambda-java-events.version>
         <aws-lambda-serverless-java-container.version>1.3.1</aws-lambda-serverless-java-container.version>
-        <awssdk.version>2.7.0</awssdk.version>
+        <awssdk.version>2.7.35</awssdk.version>
         <azure-functions-java-library.version>1.3.0</azure-functions-java-library.version>
         <kotlin.version>1.3.41</kotlin.version>
         <dekorate.version>0.6.1</dekorate.version>
--- a/bom/runtime/pom.xml
+++ b/bom/runtime/pom.xml
@@ -23,7 +23,7 @@
         <opentracing-tracerresolver.version>0.1.5</opentracing-tracerresolver.version>
         <opentracing-concurrent.version>0.2.0</opentracing-concurrent.version>
         <jaeger.version>0.34.0</jaeger.version>
-        <quarkus-http.version>3.0.0.Alpha4</quarkus-http.version>
+        <quarkus-http.version>3.0.0.Alpha5</quarkus-http.version>
         <jboss-servlet-api_4.0_spec.version>1.0.0.Final</jboss-servlet-api_4.0_spec.version>
         <microprofile-config-api.version>1.3</microprofile-config-api.version>
         <microprofile-context-propagation.version>1.0-RC1</microprofile-context-propagation.version>
--- a/bom/runtime/pom.xml
+++ b/bom/runtime/pom.xml
@@ -119,7 +119,7 @@
         <infinispan.version>10.0.0.Beta5</infinispan.version>
         <infinispan.protostream.version>4.3.0.Alpha9</infinispan.protostream.version>
         <caffeine.version>2.6.2</caffeine.version>
-        <netty.version>4.1.34.Final</netty.version>
+        <netty.version>4.1.39.Final</netty.version>
         <reactive-streams.version>1.0.2</reactive-streams.version>
         <test-containers.version>1.12.0</test-containers.version>
         <jboss-logging.version>3.3.2.Final</jboss-logging.version>
--- a/extensions/smallrye-reactive-messaging/deployment/src/main/java/io/quarkus/smallrye/reactivemessaging/deployment/EmitterBuildItem.java
+++ b/extensions/smallrye-reactive-messaging/deployment/src/main/java/io/quarkus/smallrye/reactivemessaging/deployment/EmitterBuildItem.java
@@ -5,16 +5,48 @@ import io.quarkus.builder.item.MultiBuildItem;
 public final class EmitterBuildItem extends MultiBuildItem {
 
     /**
+     * Creates a new instance of {@link EmitterBuildItem} setting the overflow strategy.
+     *
+     * @param name the name of the stream
+     * @param overflow the overflow strategy
+     * @param bufferSize the buffer size, if overflow is set to {@code BUFFER}
+     * @return the new {@link EmitterBuildItem}
+     */
+    static EmitterBuildItem of(String name, String overflow, int bufferSize) {
+        return new EmitterBuildItem(name, overflow, bufferSize);
+    }
+
+    /**
+     * Creates a new instance of {@link EmitterBuildItem} using the default overflow strategy.
+     *
+     * @param name the name of the stream
+     * @return the new {@link EmitterBuildItem}
+     */
+    static EmitterBuildItem of(String name) {
+        return new EmitterBuildItem(name, null, -1);
+    }
+
+    /**
      * The name of the stream the emitter is connected to.
      */
     private final String name;
 
-    private String overflow;
+    /**
+     * The name of the overflow strategy. Valid values are {@code BUFFER, DROP, FAIL, LATEST, NONE}.
+     * If not set, it uses {@code BUFFER} with a default buffer size.
+     */
+    private final String overflow;
 
-    private int bufferSize;
+    /**
+     * The buffer size, used when {@code overflow} is set to {@code BUFFER}. Not that if {@code overflow} is set to
+     * {@code BUFFER} and {@code bufferSize} is not set, an unbounded buffer is used.
+     */
+    private final int bufferSize;
 
-    public EmitterBuildItem(String name) {
+    public EmitterBuildItem(String name, String overflow, int bufferSize) {
         this.name = name;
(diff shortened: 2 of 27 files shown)


geoand:
JDK 11 tests seems to be failing

extensions/smallrye-reactive-messaging/deployment/src/main/java/io/quarkus/smallrye/reactivemessaging/deployment/EmitterBuildItem.java
```diff
@@ -9,6 +9,10 @@
      */
     private final String name;
 
+    private String overflow;
```
mkouba: AFAIK build items should be immutable. What's the reason to make these fields mutable?

extensions/smallrye-reactive-messaging/deployment/src/main/java/io/quarkus/smallrye/reactivemessaging/deployment/SmallRyeReactiveMessagingProcessor.java
```diff
@@ -82,15 +84,24 @@ void validateBeanDeployment(
...
                     String name = stream.value().asString();
+                    AnnotationInstance annotation = injectionPoint.getTarget().asField()
+                            .annotation(ON_OVERFLOW);
+
                     LOGGER.debugf("Emitter injection point '%s' detected, stream name: '%s'",
                             injectionPoint.getTargetInfo(), name);
-                    emitters.produce(new EmitterBuildItem(name));
+                    EmitterBuildItem item = new EmitterBuildItem(name);
+                    if (annotation != null) {
+                        item.setOverflow(annotation.value().toString());
```
mkouba: Ok now I see why those setters. Well, I'd personally prefer a static factory method for this particular use case, ie. something like `EmitterBuildItem.from(name, annotation)` and final fields.

cescoffier:
The CI fail was kind of expected as right now we have both Vert.x 3.80. (quarkus-http) and Vert.x 3.8.1. Working on it.

extensions/smallrye-reactive-messaging/deployment/src/main/java/io/quarkus/smallrye/reactivemessaging/deployment/SmallRyeReactiveMessagingProcessor.java
cescoffier: +1

stuartwdouglas:
The version in the bom should override the version in quarkus-http, so unless there are some incompatibilities this should not be a factor.

cescoffier:
@stuartwdouglas in this release there is one incompatibility regarding the Cookie class moved from Vert.x web to Vert.x core.

gsmet:
@cescoffier in any case, as we import the Vert.x BOM, the versions should be consistent. If they aren't, it needs fixing in the BOM as we need them to be.

cescoffier:
@gsmet we need to exclude the Vert.x version from quarkus-http. Will do this.
Almost done with the quarkus-http release.

cescoffier:
@mkouba I've made the `EmitterBuildItem` immutable.
@gsmet I've made sure the Vert.x version is the one from Quarkus
@stuartwdouglas I've released and bumped Quarkus HTTP

So... now the build should pass.... Fingers crossed. 


@tsegismont FYI

bom/runtime/pom.xml
```diff
@@ -904,6 +904,11 @@
                         <groupId>io.netty</groupId>
                         <artifactId>netty-all</artifactId>
                     </exclusion>
+                    <exclusion>
+                        <!-- explicitly exclude the vert.x dependency to use the version define by Quarkus -->
```
FroMage: s/define/defined/

extensions/amazon-lambda/runtime/pom.xml
```diff
@@ -45,6 +49,7 @@
             <artifactId>mockito-core</artifactId>
             <scope>test</scope>
         </dependency>
+
```
FroMage: How's that related? Did this really pass the automated formatter?

extensions/amazon-lambda/runtime/pom.xml
```diff
@@ -35,6 +35,10 @@
             <groupId>io.quarkus</groupId>
             <artifactId>quarkus-jackson</artifactId>
         </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-vertx</artifactId>
+        </dependency>
```
FroMage: Why the new dep?

extensions/vertx/deployment/pom.xml
```diff
@@ -34,7 +34,7 @@
             <groupId>io.quarkus</groupId>
             <artifactId>quarkus-vertx</artifactId>
         </dependency>
-	<dependency>
+	    <dependency>
```
FroMage: Doubt this passes the automated formatter…

extensions/vertx/runtime/pom.xml
```diff
@@ -47,6 +47,10 @@
             <groupId>io.vertx</groupId>
             <artifactId>vertx-rx-java2</artifactId>
         </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-smallrye-reactive-streams-operators</artifactId>
+        </dependency>
```
FroMage: Why does the vert.x extension depend on SR-RSO now? I don't see any new code in that extension.

test-framework/amazon-lambda/pom.xml
```diff
@@ -31,5 +31,9 @@
             <groupId>io.quarkus.http</groupId>
             <artifactId>quarkus-http-vertx-backend</artifactId>
         </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-vertx</artifactId>
+        </dependency>
```
FroMage: Same question as for the lambda ext.

FroMage (changes requested):

extensions/amazon-lambda/runtime/pom.xml
cescoffier: We exclude Vert.x from `quarkus-http`, so we need to have access to it.

extensions/smallrye-reactive-messaging/deployment/src/main/java/io/quarkus/smallrye/reactivemessaging/deployment/SmallRyeReactiveMessagingProcessor.java
```diff
@@ -82,15 +84,27 @@ void validateBeanDeployment(
...
 
-        for (InjectionPointInfo injectionPoint : validationPhase.getContext().get(BuildExtension.Key.INJECTION_POINTS)) {
+        for (InjectionPointInfo injectionPoint : validationPhase.getContext()
+                .get(BuildExtension.Key.INJECTION_POINTS)) {
             if (injectionPoint.getRequiredType().name().equals(NAME_EMITTER)) {
                 AnnotationInstance stream = injectionPoint.getRequiredQualifier(NAME_STREAM);
                 if (stream != null) {
                     // Stream.value() is mandatory
                     String name = stream.value().asString();
+                    AnnotationInstance annotation = injectionPoint.getTarget().asField()
```
mkouba: We should also support constructor and initializer/setter injection...

extensions/amazon-lambda/runtime/pom.xml
cescoffier: It actually did... weird.

cescoffier: Removed the extra line.

extensions/vertx/deployment/pom.xml
cescoffier: Hum, I don't have this issue.
I had to fix it because it was using a tab instead of spaces.

extensions/smallrye-reactive-messaging/deployment/src/main/java/io/quarkus/smallrye/reactivemessaging/deployment/SmallRyeReactiveMessagingProcessor.java
```diff
@@ -82,15 +84,27 @@ void validateBeanDeployment(
...
+
                     LOGGER.debugf("Emitter injection point '%s' detected, stream name: '%s'",
                             injectionPoint.getTargetInfo(), name);
-                    emitters.produce(new EmitterBuildItem(name));
+
+                    if (annotation != null) {
+                        emitters.produce(
+                                EmitterBuildItem.of(name,
+                                        annotation.value().toString(),
+                                        annotation.value("bufferSize").asInt()));
```
mkouba: `annotation.value("bufferSize")` may return `null` if the value is not specified. You should either use `org.jboss.jandex.AnnotationInstance.valueWithDefault(IndexView, String)` or handle `null` value, e.g. with a hardcoded default value.

extensions/vertx/runtime/pom.xml
cescoffier: It's to have access to the Axle classes. It's was merged in another PR. It should be fixed with the new Axle stuff.

test-framework/amazon-lambda/pom.xml
cescoffier: Same answer, you need to give access to the Vert.x classes as they are not resolved by transitivity anymore.

extensions/smallrye-reactive-messaging/deployment/src/main/java/io/quarkus/smallrye/reactivemessaging/deployment/SmallRyeReactiveMessagingProcessor.java
mkouba: By the way, the `OnOverflow`'s javadoc is using wrong annotation name in the example: `OnBackPressure`.

mkouba (changes requested):

extensions/amazon-lambda/runtime/pom.xml
FroMage: OK.

extensions/vertx/deployment/pom.xml
FroMage: Well, if you ran it, and it's fine…

bom/runtime/pom.xml
```diff
@@ -904,6 +904,11 @@
                         <groupId>io.netty</groupId>
                         <artifactId>netty-all</artifactId>
                     </exclusion>
+                    <exclusion>
+                        <!-- explicitly exclude the vert.x dependency to use the version defined by Quarkus -->
+                        <groupId>io.vertx</groupId>
+                        <artifactId>vertx-core</artifactId>
```
gsmet: What I was saying is that this shouldn't be necessary as we include the Vert.x bom at the beginning of this bom.

If it is an issue, then there's something going wrong.

extensions/vertx/runtime/pom.xml
FroMage: I don't understand how SR-RSO is related to Axle.

extensions/amazon-lambda/runtime/pom.xml
gsmet: Don't we need to add the deployment dependency in the deployment module?

extensions/vertx/deployment/pom.xml
gsmet: The automated formatter doesn't check XML files, only Java.

But, yes, there is still a tab there.

bom/runtime/pom.xml
cescoffier: Your BOM won't change the version imported by a dependency.

extensions/smallrye-reactive-messaging/deployment/src/main/java/io/quarkus/smallrye/reactivemessaging/deployment/SmallRyeReactiveMessagingProcessor.java
cescoffier: > We should also support constructor and initializer/setter injection...

what do you mean?

extensions/vertx/runtime/pom.xml
cescoffier: Axle provides methods returning `PublisherBuilder` defined in RSO.

FroMage: Well, OK, sorry, but does the vert.x extension use `PublisherBuilder` now?

bom/runtime/pom.xml
cescoffier: I may have a solution, but that may end up ugly.

extensions/vertx/runtime/pom.xml
cescoffier: It exposes the Axle Vert.x instance so somehow yes...
This dependency was added to be more convenient (in another PR, this is just a fix as the dependency for not declared in the runtime).

extensions/smallrye-reactive-messaging/deployment/src/main/java/io/quarkus/smallrye/reactivemessaging/deployment/SmallRyeReactiveMessagingProcessor.java
mkouba: An injection point does not have to be a field ;-). You can also use a bean constructor or initializer methods (e.g. a setter) -> in that case, `InjectionPointInfo.getTarget()` returns a `MethodInfo`; `InjectionPointInfo.getPosition()` can be used to obtain the parameter annotations. `io.quarkus.arc.processor.Annotations.getParameterAnnotations(BeanDeployment, MethodInfo, int)` is a convenient util method to get the annotations including those added by means of `AnnotationsTransformer`.

extensions/vertx/runtime/pom.xml
FroMage: If it's exposed by the Axle Vert.x instance, it should be imported by the vertx axle module, no? If it's for convenience, that's another thing.

bom/runtime/pom.xml
cescoffier: Actually nope.

What could be done is to have an _empty_ module in quarkus that depends on Vert.x core and quarkus-http (in this order), and all other modules would depend on this module.

As there is no easy solution, we probably should remove the `exclusion` and manage the alignment by hand for now. That would remove the dependencies added to the lambda modules.

extensions/smallrye-reactive-messaging/deployment/src/main/java/io/quarkus/smallrye/reactivemessaging/deployment/SmallRyeReactiveMessagingProcessor.java
cescoffier: Oh I see...

bom/runtime/pom.xml
gsmet: > Your BOM won't change the version imported by a dependency.

Hu? Are we talking about the same thing? A BOM is part of dependency management so it will apply a consistent version across all your dependencies.

cescoffier: Here, quarkus-http was depending on vert.x 3.8.0.
Even if the bom says that the Vert.x version is 3.8.1, if a module depends on quarkus-http, and does not explicitly depends on vert.x, it will resolve to 3.8.0.

cescoffier: That was actually the first build issue. Even with the BOM, it was resolving to Vert.x 3.8.0.

extensions/vertx/runtime/pom.xml
cescoffier: Vert.x Axle depends on the spec only, not on the implementation. Here we explicitly depend on the implementation.

extensions/smallrye-reactive-messaging/deployment/src/main/java/io/quarkus/smallrye/reactivemessaging/deployment/SmallRyeReactiveMessagingProcessor.java
cescoffier: Ok should be fixed now.

extensions/smallrye-reactive-messaging/deployment/src/main/java/io/quarkus/smallrye/reactivemessaging/deployment/SmallRyeReactiveMessagingProcessor.java
cescoffier: Done.

bom/runtime/pom.xml
cescoffier: I've removed the exclusion.

extensions/smallrye-reactive-messaging/deployment/src/main/java/io/quarkus/smallrye/reactivemessaging/deployment/SmallRyeReactiveMessagingProcessor.java
cescoffier: so it was NOT fixed.

bom/runtime/pom.xml
gsmet: Then we have an issue with the BOM because it should enforce consistently the Vert.x version across all the codebase, including in your case. That's what dependency management is for. So if it doesn't work, we have something to fix with the Vert.x BOM inclusion.

cescoffier: Then we need to merge quarkus-http in the main codebase or implement the _empty_ module trick.

gsmet: No, it should work as is. That's what I'm trying to say. Dependency management will enforce consistency across all your dependencies and the dependencies of your dependencies (and so on).

I trust you that there was an issue, I'll check later what's going on with the BOM.

cescoffier: So right now, it enforce the vert.x version in the quarkus codebase which is what we want.

extensions/smallrye-reactive-messaging/deployment/src/main/java/io/quarkus/smallrye/reactivemessaging/deployment/SmallRyeReactiveMessagingProcessor.java
cescoffier: now it's fixed, and I've added a test to ensure it's working.

mkouba (approved):

FroMage (approved):

cescoffier:
It will likely not work, as the updated netty contains substitutions and so we have during duplicates. I will investigate tomorrow.

cescoffier:
ok, the missing dependency has finally been cached by the Google Maven mirror.

cescoffier:
Update:

* Another issue with the Vert.x web client has been fixed
* Due to AWS SDK incompatibility with the new netty, I've disabled the async support of the dynamodb extension. An issue has been filled on the SDK (https://github.com/aws/aws-sdk-java-v2/issues/1399), and the @marcinczeczko has been made aware (and is ok with it)

PR #3752, opened 2019-08-28
Removal of unused org.glassfish.web:el-impl from BOM

rsvoboda:
Removal of unused org.glassfish.web:el-impl from BOM

I noticed that due to undefined `el-impl.version` property used by `<version>${el-impl.version}</version>`. When looking a bit more I realized that org.glassfish.web:el-impl is not used anywhere in Quarkus code.

### DIFF
--- a/bom/runtime/pom.xml
+++ b/bom/runtime/pom.xml
@@ -1678,11 +1678,6 @@
                 </exclusions>
             </dependency>
             <dependency>
-                <groupId>org.glassfish.web</groupId>
-                <artifactId>el-impl</artifactId>
-                <version>${el-impl.version}</version>
-            </dependency>
-            <dependency>
                 <groupId>org.postgresql</groupId>
                 <artifactId>postgresql</artifactId>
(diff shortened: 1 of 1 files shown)


geoand:
Perhaps it was there to ensure that it doesn't break anything in Quarkus?

rsvoboda:
Nope, the `el-impl.version` property is not defined at all

you would get
```
[ERROR] Some problems were encountered while processing the POMs:
  [ERROR] 'dependencies.dependency.version' for org.glassfish.web:el-impl:jar must be
  a valid version but is '${el-impl.version}'. @ io.quarkus:quarkus-bom:999-SNAPSHOT
...
```

geoand:
OK thanks. Seems good to me but I'll let someone else comment if they have a better idea of why it might be there.

stuartwdouglas:
It would have originally been there to prevent dependency convergence errors, which may not be an issue any more. I think it was added due to conflicts between HV and Weld, but I could be wrong.

stuartwdouglas (approved):

PR #3759, opened 2019-08-28
Property el-impl.version used but not defined

rverma-nikiai:
This is causing failure when running a bazel based migration to consume pom and generate bazel artifacts. Bazel has strict definitions for pom structures.


gsmet:
If so, that would be the right fix because this EL artifact is clearly outdated.

stuartwdouglas:
Closing, as the correct fix is to remove the unused definition

PR #3775, opened 2019-08-29
MP Metrics TCK

jmartisk:
I tried squashing the commits into reasonable chunks, if more squashing is required just tell me

Hmm also I just noticed that @kenfinnigan 's original commit also contains some basic scaffolding for openapi TCK, but it's not active. If requested I will remove it.


kenfinnigan (approved):

kenfinnigan:
Sorry about the OpenAPI TCK pom! Didn't realize that was in my original commit.

It's one of the TCKs we need to add, so don't mind whether it's there and not used or removed from PR

PR #3778, opened 2019-08-30
Fixes for JAX-RS and reactive return types

stuartwdouglas:


extensions/vertx-web/runtime/src/main/java/io/quarkus/vertx/web/runtime/VertxWebRecorder.java
```diff
@@ -87,19 +87,19 @@ public static void shutDownDevMode() {
...
         ConfigInstantiator.handleObject(vertxConfiguration);
-        VertxRecorder.initialize(vertxConfiguration);
+        VertxRecorder.setWebVertx(VertxRecorder.initialize(vertxConfiguration));
 
         try {
             HttpConfiguration config = new HttpConfiguration();
             ConfigInstantiator.handleObject(config);
 
-            router = Router.router(VertxRecorder.getVertx());
+            router = Router.router(VertxRecorder.getWebVertx());
```
cescoffier: not sure about the name as it's a bit confusing with Vert.x Web. 
should we qualified is as "RestEasy" or "internal"

integration-tests/reactive-pg-client/src/main/java/io/quarkus/it/reactive/pg/client/FruitResource.java
```diff
@@ -34,7 +34,7 @@ void setupDb() {
     @GET
     @Produces(MediaType.APPLICATION_JSON)
     public CompletionStage<JsonArray> listFruits() {
-        return client.query("SELECT * FROM fruits").thenApply(pgRowSet -> {
+        return client.query("SELECT * FROM fruits").thenApplyAsync(pgRowSet -> {
```
cescoffier: We don't recommend using async without executor as it may behave very differently in containers. But as it's an IT, should not matter too much.

cescoffier (approved):
It fixes the SSE issue we have seen last week.

extensions/vertx/deployment/src/main/java/io/quarkus/vertx/deployment/WebVertxBuildItem.java
```diff
@@ -0,0 +1,24 @@
+package io.quarkus.vertx.deployment;
+
+import io.quarkus.builder.item.SimpleBuildItem;
+import io.quarkus.runtime.RuntimeValue;
+import io.vertx.core.Vertx;
+
+/**
+ * TODO: DELETE THIS CLASS
```
mkouba: Do we have a follow-up issue? I'm afraid a TODO is not enough and it should be really just a temporary workaround...

FroMage: Well, the issue is well known and being discussed in many places, I doubt it will get lost.

integration-tests/reactive-pg-client/src/main/java/io/quarkus/it/reactive/pg/client/FruitResource.java
FroMage: Errr, didn't we add `Async` to work around the fact that we're on the IO thread? I thought this PR fixed that issue by getting around it. @stuartwdouglas are you sure about this line?

stuartwdouglas: That came from the revert

FroMage: Well, ok, but that was a workaround that the code reverted fixed. Your new dual-vertx also fixes the issue so that workaround is also not needed anymore.

gsmet:
Superseded by https://github.com/quarkusio/quarkus/pull/3896 .

PR #3780, opened 2019-08-30
Fix example snippet typo containing double ']'

tarilabs:
Taking reference default for example https://github.com/quarkusio/quarkus/blob/master/docs/src/main/asciidoc/logging-guide.adoc#run-time-configuration

Change (what to me looks indeed a) typo in the example logging configuration snippet

```
from: [%c{2.}]]
  to: [%c{2.}]
```


geoand (approved):
Thanks!

PR #3794, opened 2019-09-01
Fix maven cache on windows

stuartwdouglas:


azure-pipelines.yml
```diff
@@ -56,7 +56,33 @@ stages:
...
+          vmImage: 'vs2017-win2016'
+
+        variables:
+          imageName: 'quarkus-windows:$(build.buildId)'
+
+        steps:
+
+          - task: CacheBeta@0
+            inputs:
+              key: mavenWindows | bom/runtime/pom.xml   #if we attempt to use all poms then when they get copied to target everything breaks. This should be good enough, it does not need to be perfect
```
geoand: Just out of curiosity, where does this `mavenWindows` come from? Is it documented in Azure Pipeline docs?

stuartwdouglas: Its an arbitrary string. The cache key is this string appended to the sha256 of the contents of the bom. Because the line endings are different on windows the cache key was not matching on windows, because it thought the bom had changed. I could have just left is as 'maven' and everything would have still worked but I think it is slightly cleaner this way. ('maven' is also an arbitrary string)

geoand: Thanks, great explanation! 
I thought `maven` was a binary or some azure command or something which is why I asked.

geoand (approved):
:+1:

PR #3800, opened 2019-09-02
Rename Subclass method

stuartwdouglas:
The current name has a high chance of conflicting, e.g. HttpFilter.destroy()

### DIFF
--- a/independent-projects/arc/runtime/src/main/java/io/quarkus/arc/Subclass.java
+++ b/independent-projects/arc/runtime/src/main/java/io/quarkus/arc/Subclass.java
@@ -7,8 +7,4 @@ package io.quarkus.arc;
  */
 public interface Subclass {
 
-    default void destroy$subclass() {
-        // Noop
-    }
-
 }
--- a/independent-projects/arc/runtime/src/main/java/io/quarkus/arc/Subclass.java
+++ b/independent-projects/arc/runtime/src/main/java/io/quarkus/arc/Subclass.java
@@ -7,7 +7,7 @@ package io.quarkus.arc;
  */
 public interface Subclass {
 
-    default void destroy() {
+    default void destroy$subclass() {
         // Noop
     }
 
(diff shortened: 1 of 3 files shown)


independent-projects/arc/runtime/src/main/java/io/quarkus/arc/Subclass.java
```diff
@@ -7,7 +7,7 @@
  */
 public interface Subclass {
 
-    default void destroy() {
+    default void destroy$subclass() {
```
mkouba: +1 for renaming but when I look at it, this method should not be part of the public API at all. We should remove this method from the `Subclass` interface and follow the naming convention we use for client proxies where an internal generated method is called `arc$delegate()`. So it should be `arc$destroy()` in this case.

mkouba (changes requested):

independent-projects/arc/runtime/src/main/java/io/quarkus/arc/Subclass.java
```diff
@@ -7,8 +7,4 @@
  */
 public interface Subclass {
```
manovotn: With the method removed from here, this interface is now empty - does it make sense to keep it just as a marker interface?

mkouba: I believe it could be useful ;-)

manovotn: Hmm, I suppose you are right, after all, similar thing is in Weld for subclasses and for proxies...

manovotn (approved):

mkouba (approved):

independent-projects/arc/runtime/src/main/java/io/quarkus/arc/Subclass.java
gsmet: Yes, please keep it, I use it :).

PR #3809, opened 2019-09-02
Tweak the configuration reference generation

FroMage:


core/processor/src/main/java/io/quarkus/annotation/processor/generate_doc/DocGeneratorUtil.java
```diff
@@ -195,4 +217,8 @@ static String join(String delim, Iterator<String> it) {
     static String hyphenate(String orig) {
         return join("-", lowerCase(camelHumpsIterator(orig)));
     }
+
+    static String hyphenateEnumValue(String orig) {
```
machi1990: We should probably skip these cases: 
- an enum with a `@DefaultConverter`  e.g a`Protocol` enum here https://github.com/quarkusio/quarkus/blob/master/extensions/vertx-web/runtime/src/main/java/io/quarkus/vertx/web/runtime/ServerSslConfig.java#L28-L30 and its converter here  
https://github.com/quarkusio/quarkus/blob/master/core/runtime/src/main/java/io/quarkus/runtime/configuration/ssl/ProtocolConverter.java 
- an config item with a custom `@ConvertWith()`, I do not think we have any example from an extension but this is possible. Example usage of `@ConvertWith()` can be found here https://github.com/quarkusio/quarkus/blob/master/core/test-extension/runtime/src/main/java/io/quarkus/extest/runtime/config/TestRunTimeConfig.java#L72

gsmet: And we need to check if it's OK with the Hibernate Search enums as we have a MP config `parse` method there.

core/processor/src/main/java/io/quarkus/annotation/processor/generate_doc/SummaryTableDocFormatter.java
```diff
@@ -7,8 +7,8 @@
 
 class SummaryTableDocFormatter implements DocFormatter {
     private static final String TABLE_CLOSING_TAG = "\n|===";
-    private static final String TABLE_ROW_FORMAT = "\n\n|<<%s, %s>>\n|%s %s\n|%s\n| %s";
-    private static final String TABLE_HEADER_FORMAT = "== Summary\n%s|===\n|Configuration property|Type|Default|Lifecycle";
+    private static final String TABLE_ROW_FORMAT = "\n\n|<<%s, %s>>\n\n%s|%s %s\n|%s\n| %s";
+    private static final String TABLE_HEADER_FORMAT = "== Summary\n%s\n[cols=\"50,10,10,5\"]\n|===\n|Configuration property|Type|Default|Lifecycle";
```
gsmet: Is it normal that the sum is not 100%?

core/processor/src/main/java/io/quarkus/annotation/processor/generate_doc/GenerateExtensionConfigurationDoc.java
```diff
@@ -356,13 +358,12 @@ private void recordConfigItems(List<ConfigItem> configItems, Element element, St
                                 configItem.setWithinAMap(true);
                             }
                         } else {
-                            type = typeArguments.get(0).toString();
+                            // FIXME: I assume this is for Optional<T>
```
machi1990: Thanks for spotting this, it is for `Optional<T>` and `List<T>`.

core/processor/src/main/java/io/quarkus/annotation/processor/Constants.java
```diff
@@ -90,9 +90,10 @@
             "====\n";
 
     static {
-        OPTIONAL_NUMBER_TYPES.put(OptionalLong.class.getName(), Long.class.getName());
-        OPTIONAL_NUMBER_TYPES.put(OptionalInt.class.getName(), Integer.class.getName());
-        OPTIONAL_NUMBER_TYPES.put(OptionalDouble.class.getName(), Double.class.getName());
+        ALIASED_TYPES.put(OptionalLong.class.getName(), Long.class.getName());
+        ALIASED_TYPES.put(OptionalInt.class.getName(), Integer.class.getName());
+        ALIASED_TYPES.put(OptionalDouble.class.getName(), Double.class.getName());
+        ALIASED_TYPES.put("java.lang.Class<?>", "class name");
```
gsmet: +1, this one was bugging me.

core/processor/src/main/java/io/quarkus/annotation/processor/generate_doc/GenerateExtensionConfigurationDoc.java
```diff
@@ -342,6 +343,7 @@ private void recordConfigItems(List<ConfigItem> configItems, Element element, St
                     List<? extends TypeMirror> typeArguments = declaredType.getTypeArguments();
 
                     if (!typeArguments.isEmpty()) {
+                        // FIXME: this is super dodgy: we should check the type!!
                         if (typeArguments.size() == 2) {
```
machi1990: About the FIXME: may be we have to add a check for `java.util.Map` type.

machi1990:
We have a CI failure on this test https://github.com/quarkusio/quarkus/blob/master/core/processor/src/test/java/io/quarkus/annotation/processor/generate_doc/DocGeneratorUtilTest.java#L52 since primitive wrappers are unboxed. We can delete this line :-)

core/processor/src/main/java/io/quarkus/annotation/processor/generate_doc/DocGeneratorUtil.java
FroMage: > And we need to check if it's OK with the Hibernate Search enums as we have a MP config parse method there.

Well, they have an `of(String)` method, as I mentioned, but it just maps `FOO_BAR` to `foo-bar`. So we're good, _except_ that we can't assume all custom converters will do the same thing. I suggest we presume enums with custom converters map that way, rather than defaulting to the more useless type of `String`, no?

Otherwise we need to special-case known enums, but that's sad.

FroMage: > an enum with a @DefaultConverter e.g aProtocol enum here https://github.com/quarkusio/quarkus/blob/master/extensions/vertx-web/runtime/src/main/java/io/quarkus/vertx/web/runtime/ServerSslConfig.java#L28-L30 and its converter here
https://github.com/quarkusio/quarkus/blob/master/core/runtime/src/main/java/io/quarkus/runtime/configuration/ssl/ProtocolConverter.java

I don't get it. How is `Protocol` even involved? It's a `List<String>`.

core/processor/src/main/java/io/quarkus/annotation/processor/generate_doc/SummaryTableDocFormatter.java
FroMage: Yeah, OCD does not apply to percentage sums in HTML columns :)

FroMage:
> Thanks for this first improvement of the doc generation @FroMage. We have a CI failure on this test https://github.com/quarkusio/quarkus/blob/master/core/processor/src/test/java/io/quarkus/annotation/processor/generate_doc/DocGeneratorUtilTest.java#L52 since primitive wrappers are unboxed. We can delete this line :-)

Good point!

core/processor/src/main/java/io/quarkus/annotation/processor/generate_doc/DocGeneratorUtil.java
machi1990: `Protocol` was just an example of cases where you'd want to bail out (hence the  `@DefaultConverter`) from the usage of hyphenated enum converter and use a `custom MP config converter`.

FroMage:
@gsmet can we merge this as improvement to the current situation ? this PR is not intended to fix all issues with the docs and I likely won't have time to fix all remaining issues that are logged on other issues.

core/processor/src/main/java/io/quarkus/annotation/processor/generate_doc/GenerateExtensionConfigurationDoc.java
```diff
@@ -380,6 +381,34 @@ private void recordConfigItems(List<ConfigItem> configItems, Element element, St
...
+    }
+
+    private String extractEnumValues(TypeMirror realTypeMirror) {
+        Element declaredTypeElement = ((DeclaredType) realTypeMirror).asElement();
+        StringBuilder sb = new StringBuilder();
+        for (Element field : declaredTypeElement.getEnclosedElements()) {
+            if (field.getKind() == ElementKind.ENUM_CONSTANT) {
+                if (sb.length() > 0)
+                    sb.append(", ");
+                sb.append(DocGeneratorUtil.hyphenateEnumValue(field.getSimpleName().toString()));
```
gsmet: Should we add backticks around each value?

FroMage: Huh, probably, yes

FroMage: Done.

gsmet:
I'll review the output tomorrow to be sure it's OK.

gsmet (approved):

gsmet:
Ah missed an issue with the backticks, will fix.

PR #3811, opened 2019-09-02
Initializing the default config at runtime

sberyozkin:


gsmet:
What's the rationale of this change? Might help for the review.

sberyozkin:
Hi @gsmet, 

There is a long [discussion](https://groups.google.com/forum/#!topic/quarkus-dev/OmcRAFO6Bkk) at the forum: the problem is that the native `TikaConfig`, in the default form, attempts to load all the parsers supported by Tika which causes a failure for the static init recording, few parsers which I try to exclude by not registering them with `ServiceProviderBuildItem` are still loaded, as Martin explained to me. 

I've only discovered it after I started working on the quickstart demo which does not yet have its own `tika-config.xml`.

So the current compromise as I said in that forum is to do the best effort at recording at the static init time and only defaulting to the runtime init time if no custom config resource is available.

The only other solution I can think of is to revert to the old code where the TikaParser was created on the 1st request, and add some synchronization to avoid creating it every time, but it is not perfect.

It is possible in theory that someone will use a tika-config.xml to configure one of those blocked parsers and the static init will fail again, but it won't be a problem of this branching, but the fact some parsers simply won't work in the native mode and I really don't expect users using them any time soon. 

I think once the tika-parsers modularization work is completed (a Tika colleague of mine did a lot of work earlier and I've committed to completing that work - but it will take few months) it will be better and more optimal.

Let me know please what you think, may be there could be a cleaner solution

FroMage:
CI failure seems unrelated.

sberyozkin:
I've been thinking more all these blocked parsers and just could not come up with the more optimal solution - delaying the check till runtime will not make them operational anyway. 

Out of those 6 blocked parsers, as far as I recall, 3 or 4 of them can be reenabled for the native image (with the runtime init time) once the `allow an incomplete classpath` option status is made visible to the extension (though I'd rather wait for the users actually ask for those rare parsers), while 2 of them I'd keep blocked (Grob one which depends on the remote call and the CXF client code and MAT one - its underlying MAT reader was last updated in 2015, and it is all sync-ed, with the runtime created method handle, with a lot of file mapping, not sure it will be needed).
I do hope that with the tika-parsers modularization done, I'd just have most of those blocked parsers removed and let users add the needed dependencies if they need them, delay the class path check till runtime, etc

FroMage:
That looks OK to me, but @gsmet do you want to review yourself?

sberyozkin:
Hi @gsmet and @FroMage, the thought that something is not quite perfect with this PR keeps nagging me :-). Especially the fact I've changed the default value of the configuration. Another idea came to me yesterday while I was off, which is to generate the default configuration if no custom resource exists, while continue blocking those parsers. It will let the extension do the static init recording only. I'll experiment next week when I can get to back to my laptop full time. 

Sorry I've rushed with this PR, but I hope I will be able to close it if the other solution works, will mark this one as coding in progress for now.
Thanks for the review so far.

sberyozkin:
@gsmet, @FroMage, the alternative approach I mentioned in the previous comment works really well, with the static init recording only, so I will close this PR. 
One thing I'd like to retain from this PR though, despite being concerned about it earlier, is to remove the default value for the `tika-config-path` property, I've realized again today the extension can't react otherwise to the case where the custom resource path is wrong.
thanks.

PR #3688, opened 2019-08-25
Add support for Jackson's JsonPOJOBuilder

misl:
Added support to Jackson extension for immutable POJO models using @JsonPOJOBuilder.

- Both model and builder are added as reflective classes
- Also supports model and builder inheritance (add superclasses as well)
- Added exclusion for @JsonDeserialize without builder field
- Includes integration tests.

PR #3814, opened 2019-09-03
Ensure the Kubernetes extension doesn't register too much reflection for Jackson

geoand:
Addresses the concern raised [here](https://github.com/quarkusio/quarkus/pull/3688#issuecomment-527289195).


geoand:
For posterity, the problem was that without the change a very large part of the Kubernetes model becomes reachable (during native image build). Due to the large number of classes in the model (500+) this resulted in much longer native image creation time and a much larger native image.

geoand:
@stuartwdouglas, @gsmet, @n1hility can someone review and merge please so we can get CI times down?

Thanks

stuartwdouglas (approved):

PR #3821, opened 2019-09-03
Fix missing qualifiers in reactive messaging

cescoffier:


mkouba (approved):

gsmet:
@cescoffier I suppose this should be backported for 0.21.2 or is it a new bug?

PR #3831, opened 2019-09-03
Normal scoped producer method that returns null value at runtime should throw IllegalProductException

### LINKED ISSUE #3829: Normal scoped producer method that returns null value at runtime should throw IllegalProductException
See https://docs.jboss.org/cdi/spec/2.0/cdi-spec.html#producer_method.

gastaldi:
Fixes #3829

@mkouba please review

### DIFF
--- a/extensions/arc/deployment/src/test/java/io/quarkus/arc/test/producer/NullProducerMethodTest.java
+++ b/extensions/arc/deployment/src/test/java/io/quarkus/arc/test/producer/NullProducerMethodTest.java
@@ -1,0 +1,43 @@
+package io.quarkus.arc.test.producer;
+
+import javax.enterprise.context.ApplicationScoped;
+import javax.enterprise.inject.IllegalProductException;
+import javax.enterprise.inject.Instance;
+import javax.enterprise.inject.Produces;
+import javax.inject.Inject;
+
+import io.quarkus.test.QuarkusUnitTest;
+import org.jboss.shrinkwrap.api.ShrinkWrap;
+import org.jboss.shrinkwrap.api.spec.JavaArchive;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.RegisterExtension;
+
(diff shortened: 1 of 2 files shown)


extensions/arc/deployment/src/test/java/io/quarkus/arc/test/producer/NullProducerMethodTest.java
```diff
@@ -0,0 +1,43 @@
...
+        assertThrows(IllegalProductException.class, instance::get);
+    }
+
+    class MyBean {
+    }
+
+    @ApplicationScoped
+    static class MyBeanProducer {
+        @Produces
+        MyBean produceNull() {
```
mkouba: In fact, the producer must be annotated with a normal scope. In this case, the producer bean is `@Dependent` and so returning `null` is legal.

independent-projects/arc/runtime/src/main/java/io/quarkus/arc/InstanceImpl.java
```diff
@@ -80,7 +81,13 @@ public T get() {
         } else if (beans.size() > 1) {
             throw new AmbiguousResolutionException("Beans: " + beans.toString());
         }
-        return getBeanInstance((InjectableBean<T>) beans.iterator().next());
+        InjectableBean<T> bean = (InjectableBean<T>) beans.iterator().next();
```
mkouba: This is not correct. `IllegalProductException` should only be thrown for producer beans with a normal scope. This would throw an IPE even for `@Dependent` producers that are allowed to return `null`.

mkouba (commented):
This PR attempts to solve problem with the programmatic lookup. But we need to modify the bytecode generator as well, to throw an IPE when a producer method is called for normal injection. I'm going to send a new PR to address all the mentioned issues.

mkouba:
This PR is superseded by https://github.com/quarkusio/quarkus/pull/3841.

PR #3836, opened 2019-09-03
Add an option for create-extension mojo to create an integration test module

ppalaga:
The first commit adding `additionalRuntimeBomEntries` is motivated by our need in Camel Quarkus, where the newly added extensions typically depend on the Camel component they add support for. That component needs be added to the runtime BOM and to the extension's runtime module.

The second commit adds the option to create also an integration test module when scaffolding a new extension.

### DIFF
--- a/devtools/common/src/main/java/io/quarkus/maven/utilities/PomTransformer.java
+++ b/devtools/common/src/main/java/io/quarkus/maven/utilities/PomTransformer.java
@@ -359,15 +359,10 @@ public class PomTransformer {
         }
 
         public static Transformation addManagedDependency(String groupId, String artifactId, String version) {
-            return addManagedDependency(groupId, artifactId, version, null, null);
+            return addManagedDependency(new Gavtcs(groupId, artifactId, version, null, null, null));
         }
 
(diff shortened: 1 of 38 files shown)


FroMage:
Mmmm, not sure who should review this. @cescoffier ?

ppalaga:
> Mmmm, not sure who should review this. @cescoffier ?

@aloubyansky ?

ppalaga:
Any chance to review, @aloubyansky ?

aloubyansky (approved):
I went over it and didn't see any immediate issue. Haven't tried it though. I'd merge it and improve it as people start using it. Thanks @ppalaga

emmanuelbernard:
One comment @ppalaga , why did you decide to add a `fail()` test case in the generated code. My concern is that it would be a bit counter intuitive for a generated project to fail and my through people off. But you probably have a reason in mind.

emmanuelbernard (commented):
see my comment on the `fail()` The rest looks good to me

ppalaga:
> why did you decide to add a `fail()`

It is just a way to mark the test as unfinished. I think the test should not pass as long as the user has not bothered to add some meaningful assertions.

PR #3841, opened 2019-09-04
Null normal scoped producers throw IllegalProductException

### LINKED ISSUE #3829: Normal scoped producer method that returns null value at runtime should throw IllegalProductException

mkouba:
- resolves #3829

### DIFF
--- a/independent-projects/arc/tests/src/test/java/io/quarkus/arc/test/producer/illegal/IllegalProducerTest.java
+++ b/independent-projects/arc/tests/src/test/java/io/quarkus/arc/test/producer/illegal/IllegalProducerTest.java
@@ -1,0 +1,56 @@
+package io.quarkus.arc.test.producer.illegal;
+
+import static org.junit.Assert.fail;
+
+import io.quarkus.arc.Arc;
+import io.quarkus.arc.test.ArcTestContainer;
+import java.io.Serializable;
+import java.time.temporal.Temporal;
+import javax.enterprise.context.ApplicationScoped;
+import javax.enterprise.context.Dependent;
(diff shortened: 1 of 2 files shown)


independent-projects/arc/tests/src/test/java/io/quarkus/arc/test/producer/illegal/IllegalProducerTest.java
```diff
@@ -0,0 +1,59 @@
...
+import io.quarkus.arc.test.ArcTestContainer;
+
+public class IllegalProducerTest {
+
+    @Rule
+    public ArcTestContainer container = new ArcTestContainer(IllegalProducer.class);
+
+    @Test
+    public void testNormalScopedProducerMethodReturnsNull() {
+        Serializable serializable = Arc.container().instance(Serializable.class).get();
```
mkouba: @gastaldi Note that the invocation of `Instance.get()` does not result in `IllegalProductException` because the producers are called lazily. In other words, a proxy is returned and the producer is invoked when a proxy method is called -> a proxy is required to lookup the current bean instance per each invocation. This behavior is the same as in Weld.

gastaldi:
Thanks for picking this ;) LGTM

manovotn (approved):

manovotn:
@mkouba apparently you have commited the crime of not sorting the imports ;)

mkouba:
Oops. I need to stop using auto-imports in my IDE ;-)

PR #3847, opened 2019-09-04
[DO NOT MERGE] 0.21 backports for CI testing

gsmet:
(we don't have the list of all of them below because some of them have already been pushed but it's just for CI testing anyway)


gsmet:
Hmm, looks like CI is not triggered.

PR #3854, opened 2019-09-04
More fixes to the error pages

gsmet:
@FroMage here it is!

### DIFF
--- a/extensions/resteasy/runtime/src/main/java/io/quarkus/resteasy/runtime/NotFoundExceptionMapper.java
+++ b/extensions/resteasy/runtime/src/main/java/io/quarkus/resteasy/runtime/NotFoundExceptionMapper.java
@@ -23,7 +23,6 @@ import javax.ws.rs.core.Variant;
 import javax.ws.rs.ext.ExceptionMapper;
 import javax.ws.rs.ext.Provider;
 
-import org.jboss.resteasy.core.ResourceLocatorInvoker;
 import org.jboss.resteasy.core.ResourceMethodInvoker;
 import org.jboss.resteasy.core.ResourceMethodRegistry;
 import org.jboss.resteasy.spi.Registry;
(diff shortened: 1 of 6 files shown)


FroMage (approved):
LGTM

gsmet:
I had a test failure locally so there's a good chance CI will fail. I'll have another look tomorrow.

gsmet:
@FroMage I finally was able to fix all the test issues introduced by the auto-negotiation. The last one being https://github.com/quarkusio/quarkus/pull/3854/commits/d0f6e2c3b4a68ba7c1cb720b4d7a3dc7fdc81a74 (not totally sure if RESTEasy is too strict or if the default Java Accept header does not support the RFC).

FroMage:
Huh, the spec says RESTEasy is right and the JDK is wrong: https://www.w3.org/Protocols/rfc2616/rfc2616-sec3.html#sec3.9 That's super weird. We can't be the first ones to notice this.

extensions/resteasy/runtime/src/main/java/io/quarkus/resteasy/runtime/NotFoundExceptionMapper.java
```diff
@@ -145,50 +157,58 @@ public Response toResponse(NotFoundException exception) {
...
 
-        return Response.status(Status.NOT_FOUND).entity(sb.toString()).build();
+        return Response.status(Status.NOT_FOUND).build();
     }
 
-    public static class ErrorMessage {
-        public String errorMessage;
-        public List<ResourceDescription> existingResourcesDetails;
+    private static Variant selectVariant(HttpHeaders headers) {
+        ServerDrivenNegotiation negotiation = new ServerDrivenNegotiation();
```
FroMage: Why did you use this instead of the regular jax-rs method? Did it give you different results?

FroMage (approved):

extensions/resteasy/runtime/src/main/java/io/quarkus/resteasy/runtime/NotFoundExceptionMapper.java
gsmet: TBH, I did that while trying to understand what was going wrong and then I thought it was better to keep it this way as the JAX-RS requests do more things than strictly necessary.

If this class is not stable enough, we can change it later to the version you had.

FroMage: No, I just wanted to know if it behaved differently. I expect the JAX-RS idiomatic way probably uses this underneath.

gsmet: It behaves exactly the same, it uses this class underneath but it does a bit more things. I trimmed it down to what we really wanted.

PR #3866, opened 2019-09-05
Remove the JaxbEnabledBuildItem from the resteasy-server-common exten…

### LINKED ISSUE #3860: Provides an opt-in for jaxb support in resteasy-server-common extension
**Description**
The resteasy-server-common extension provides a `@BuildStep JaxbEnabledBuildItem` which forces the support and inclusion of jaxb.  However, JAXB is not mandatory and when compiling to native, it forces the whole jaxp+jaxb stack to be included while it may not be used.

**Implementation ideas**
Make the build item to be provided only when opting-in with a configuration property somehow.  Also, I'm not even sure where this is needed as the resteasy-jaxb-provider is not even in the dependency tree, so maybe remove the JaxbEnabledBuildItem completely.

gnodet:
…sion as it's not required, fixes #3860

### DIFF
--- a/extensions/resteasy-server-common/deployment/src/main/java/io/quarkus/resteasy/server/common/deployment/ResteasyServerCommonProcessor.java
+++ b/extensions/resteasy-server-common/deployment/src/main/java/io/quarkus/resteasy/server/common/deployment/ResteasyServerCommonProcessor.java
@@ -57,7 +57,6 @@ import io.quarkus.deployment.builditem.substrate.RuntimeInitializedClassBuildIte
 import io.quarkus.deployment.builditem.substrate.SubstrateConfigBuildItem;
 import io.quarkus.deployment.builditem.substrate.SubstrateProxyDefinitionBuildItem;
 import io.quarkus.deployment.builditem.substrate.SubstrateResourceBuildItem;
-import io.quarkus.jaxb.deployment.JaxbEnabledBuildItem;
 import io.quarkus.resteasy.common.deployment.JaxrsProvidersToRegisterBuildItem;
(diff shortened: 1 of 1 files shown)


gsmet:
I think it's a bit more complicated that that :). I'll think a bit about it and come up with a plan soon.

geoand:
If it's removed, how would it re-enabled? I assume that is what you mean @gsmet by thinking about it and coming up with a plan :)

gnodet:
It could be re-enabled explicitely when actually used, or implicitely by creating a new extension which would provide the necessary `JaxbEnabledBuildItem`.

gsmet:
So, the issue is the following RESTEasy might need it, be it with the client or the server *if* the JAXB provider is used.

There are two possibilities:
- create a specific extension for JAXB which would include the dependency to the JAXB provider and enable the JAXB support
- trigger the build item if one of the JAXB providers is enabled when we scan the list of providers, we would test if one of the JAXB providers is part of the enabled ones and enable JAXB in this case

My inclination would be to do the former because I think JAXB is sufficiently widely used that an extension similar to the RESTEasy JSON extensions would be nice.

But... we would also need an extension for the REST client I think or maybe the extension could serve both purposes.

@geoand any opinion?

geoand:
@gsmet I like the idea of resteasy-jaxb extension. That approach is consistent with where things have been going with regards to other areas

gsmet:
@gnodet interested in giving it a try?

gsmet:
Superseded in favor of https://github.com/quarkusio/quarkus/pull/4405 .

Issue #3857, opened 2019-09-04
Openapi not working when quarkus.swagger-ui.path is "/"


lbernardomaia:
**Describe the bug**
Hi, not sure if is a Quarkus or a MicroProfile issue, but when I have the config bellow in the application.properties, the OpenApi endpoint doesn't work. I've attached a PrintScreen and there isn't error in the logs

application.properties:
`
quarkus.swagger-ui.path=/
`

**Expected behaviour**
The OpenApi endpoint should work independently of quarkus.swagger-ui.path config

![Screenshot from 2019-09-04 15-05-11](https://user-images.githubusercontent.com/11513050/64264348-0bcb6a80-cf29-11e9-8d41-fd1f45b4b447.png)

stuartwdouglas:
I don't think we should allow a path of / for swagger UI. Swagger takes complete control of quarkus.swagger-ui.path so if it is set of / then nothing else can be served.

lbernardomaia:
Ok, closing the issue!

machi1990:
> I don't think we should allow a path of / for swagger UI. Swagger takes complete control of quarkus.swagger-ui.path so if it is set of / then nothing else can be served.

Good candidate of https://github.com/quarkusio/quarkus/issues/2489.

PR #3872, opened 2019-09-05
fix(swagger-ui): throw an error when "/" is set as swagger-ui path

machi1990:
The value blocks the application from serving anything else - see
https://github.com/quarkusio/quarkus/issues/3857#issuecomment-528149421

### DIFF
--- a/extensions/swagger-ui/deployment/src/main/java/io/quarkus/swaggerui/deployment/SwaggerUiProcessor.java
+++ b/extensions/swagger-ui/deployment/src/main/java/io/quarkus/swaggerui/deployment/SwaggerUiProcessor.java
@@ -27,6 +27,7 @@ import io.quarkus.deployment.annotations.Record;
 import io.quarkus.deployment.builditem.FeatureBuildItem;
 import io.quarkus.deployment.builditem.LaunchModeBuildItem;
 import io.quarkus.deployment.builditem.LiveReloadBuildItem;
+import io.quarkus.deployment.configuration.ConfigurationError;
 import io.quarkus.deployment.index.ClassPathArtifactResolver;
 import io.quarkus.deployment.index.ResolvedArtifact;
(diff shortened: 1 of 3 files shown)


extensions/swagger-ui/deployment/src/main/java/io/quarkus/swaggerui/deployment/SwaggerUiProcessor.java
```diff
@@ -181,6 +187,7 @@ public String updateApiUrl(String original) {
     static final class SwaggerUiConfig {
         /**
          * The path of the swagger-ui servlet.
+         * The value `/` is not allowed as it blocks the application from serving anything else.
```
gsmet: A `<p>` just before this sentence would be welcome.

gsmet (changes requested):
Looks good apart from some very minor comment.

extensions/swagger-ui/deployment/src/main/java/io/quarkus/swaggerui/deployment/SwaggerUiProcessor.java
```diff
@@ -181,6 +187,9 @@ public String updateApiUrl(String original) {
     static final class SwaggerUiConfig {
         /**
          * The path of the swagger-ui servlet.
+         * <p>
+         * The value `/` is not allowed as it blocks the application from serving anything else.
+         * </p>
```
gsmet: Ah ah, that's why I was precise about just adding a `<p>` before :). You don't need an ending p and Javadoc throws a warning for that.

machi1990: :-) Let me change that.

gsmet (approved):

Issue #1987, opened 2019-04-10
MINOR Getting Started Guide example fix


nmcl:
In https://quarkus.io/guides/getting-started-guide the example shows the output of curl on example application which returns the string 'hello'. However, on some environments that string ends up being placed on the same line as the command line prompt which can make it harder to see, so suggest changing ...

 @Produces(MediaType.TEXT_PLAIN)
    public String hello() {
        return "hello";

... to

 @Produces(MediaType.TEXT_PLAIN)
    public String hello() {
        return "hello\n";

Sanne:
ouch, the boss is back writing in all-capitals :)

nmcl:
;)

Suggest the same for the example modification later in the section:

@ApplicationScoped
public class GreetingService {

    public String greeting(String name) {
        return "hello " + name + "\n";
    }

Sanne:
@nmcl thanks, sent a PR.

A suggestion: for any such small detail it might be easier to just edit the file on the web UI.
For example in this case, navigate to  https://github.com/quarkusio/quarkus/edit/master/docs/src/main/asciidoc/getting-started-guide.adoc you can then create a git commit and send a PR directly.

nmcl:
Though the tests also need to be changed:

@QuarkusTest
public class GreetingResourceTest {

    @Test
    public void testHelloEndpoint() {
        given()
          .when().get("/hello")
          .then()
             .statusCode(200)
             .body(is("hello\n"));
    }

    @Test
    public void testGreetingEndpoint() {
        String uuid = UUID.randomUUID().toString();
        given()
          .pathParam("name", uuid)
          .when().get("/hello/greeting/{name}")
          .then()
            .statusCode(200)
            .body(is("hello " + uuid + "\n"));
    }

}

PR #3873, opened 2019-09-05
Fixes after going through getting started guide

maxandersen:
While going through getting started guide I got "caught" doing
a few stupid things - mainly due to the docs not being explicit 
enough. Thus here are each issue I found in one PR with separate commits
for each found issue.

### DIFF
--- a/docs/src/main/asciidoc/getting-started-guide.adoc
+++ b/docs/src/main/asciidoc/getting-started-guide.adoc
@@ -367,7 +367,14 @@ public class GreetingResourceTest {
 
 These tests use http://rest-assured.io/[RestAssured], but feel free to use your favorite library.
 
-You can run the test from your IDE directly (be sure you stopped the application first), or from Maven using: `./mvnw test`.
+You can run these using Maven:
+
+[source,shell,subs=attributes+]
+----
+./mvnw test
+----
+
+You can also run the test from your IDE directly (be sure you stopped the application first).
 
 By default, tests will run on port `8081` so as not to conflict with the running application. We automatically
 configure RestAssured to use this port. If you want to use a different client you should use the `@TestHTTPResource`
--- a/docs/src/main/asciidoc/getting-started-guide.adoc
+++ b/docs/src/main/asciidoc/getting-started-guide.adoc
@@ -36,6 +36,15 @@ To complete this guide, you need:
 * JDK 1.8+ installed with `JAVA_HOME` configured appropriately
 * Apache Maven 3.5.3+
 
+[TIP]
+.Verify Maven is using the Java you expect
+====
+If you have multiple JDK's installed it is not certain Maven will pick up the expected java
+and you could end up with unexpected results.
+You can verify which JDK Maven uses by running `mvn --version`. 
+====
+
+
 == Architecture
 
 In this guide, we create a straightforward application serving a `hello` endpoint. To demonstrate
@@ -168,6 +177,7 @@ Use: `./mvnw compile quarkus:dev`:
 
 [source,shell]
 ----
+$ ./mvnw compile quarkus:dev
 [INFO] --------------------< org.acme:getting-started >---------------------
 [INFO] Building getting-started 1.0-SNAPSHOT
 [INFO] --------------------------------[ jar ]---------------------------------
--- a/docs/src/main/asciidoc/getting-started-guide.adoc
+++ b/docs/src/main/asciidoc/getting-started-guide.adoc
@@ -73,13 +73,13 @@ mvn io.quarkus:quarkus-maven-plugin:{quarkus-version}:create \
     -Dpath="/hello"
 ----
 
-It generates:
+It generates the following in  `./getting-started`:
 
 * the Maven structure
 * an `org.acme.quickstart.GreetingResource` resource exposed on `/hello`
 * an associated unit test
 * a landing page that is accessible on `http://localhost:8080` after starting the application
-* example `Dockerfile` files for both `native` and `jvm` modes
+* example `Dockerfile` files for both `native` and `jvm` modes in `src/main/docker`
 * the application configuration file
 
 Once generated, look at the `pom.xml`.
--- a/docs/src/main/asciidoc/getting-started-guide.adoc
+++ b/docs/src/main/asciidoc/getting-started-guide.adoc
@@ -201,19 +201,17 @@ Listening for transport dt_socket at address: 5005
 Once started, you can request the provided endpoint:
 
 ```
-$ curl http://localhost:8080/hello
+$ curl -w "\n" http://localhost:8080/hello
 hello
 ```
 
 Hit `CTRL+C` to stop the application, but you can also keep it running and enjoy the blazing fast hot-reload.
 
 [TIP]
-Automatically add newline with curl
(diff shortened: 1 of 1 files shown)


docs/src/main/asciidoc/getting-started-guide.adoc
```diff
@@ -141,7 +156,7 @@ public class GreetingResource {
     @GET
     @Produces(MediaType.TEXT_PLAIN)
     public String hello() {
-        return "hello\n";
+        return "hello";
```
gsmet: I'm not sure I agree with this chance: the `\n` was added by a contributor because when doing simple command line calls, your shell didn't properly start on a new line without that.

maxandersen: That user in https://github.com/quarkusio/quarkus/issues/1987 most likely used quite old style shell setup ;) 

Anyways that’s why I added the curl TIP for those unfortunate.

docs/src/main/asciidoc/getting-started-guide.adoc
```diff
@@ -47,10 +47,16 @@ This guide also covers the testing of the endpoint.
...
+
 However, you can go right to the completed example.
 
-Clone the Git repository: `git clone {quickstarts-clone-url}`, or download an {quickstarts-archive-url}[archive].
+Download an {quickstarts-archive-url}[archive] or clone the git repository:
+
+[source,shell,subs=attributes+]
+----
+git clone {quickstarts-clone-url}
+----
```
gsmet: This section is present in most of the guides? Do you plan to upgrade all of them? Or we just make the getting started one smoother and suppose people looking at the other guides know how to git clone?

docs/src/main/asciidoc/getting-started-guide.adoc
```diff
@@ -262,13 +270,20 @@ public class GreetingResource {
...
     }
 }
 ----
 
 If you stopped the application, restart the application with `./mvnw compile quarkus:dev`.
-Then check that http://localhost:8080/hello/greeting/quarkus returns `hello quarkus`.
+Then check that the endpoint returns `hello quarkus` as expected:
+
+```
+$ curl http://localhost:8080/hello/greeting/quarkus
```
gsmet: I would use `-w "\n"` in this example.

docs/src/main/asciidoc/getting-started-guide.adoc
```diff
@@ -227,7 +235,7 @@ import javax.enterprise.context.ApplicationScoped;
 public class GreetingService {
 
     public String greeting(String name) {
-        return "hello " + name + "\n";
+        return "hello " + name;
```
gsmet: We need the corresponding quickstart to be aligned with this if it's not already the case.

gsmet (commented):
Added a couple of questions/comments. Looks good otherwise.

docs/src/main/asciidoc/getting-started-guide.adoc
maxandersen: i've just made it here so far - I can make updates for the others too; but I suggest we do that in a follow up PR.

docs/src/main/asciidoc/getting-started-guide.adoc
maxandersen: all the quickstarts I could find does not have \n. That is something only added to the getting started guide and one of the reasons I suggested to roll it back since it makes things "different" and quirky as it breaks the tests until you have updated them in the future steps.

docs/src/main/asciidoc/getting-started-guide.adoc
maxandersen: I kept it out to avoid having to repeat it everywhere - but I guess I could add it and then remove the [tip] section about curl ?

docs/src/main/asciidoc/getting-started-guide.adoc
gsmet: I was thinking about this code https://github.com/quarkusio/quarkus-quickstarts/blob/master/getting-started/src/main/java/org/acme/quickstart/GreetingResource.java#L27 but apparently it was not in sync. Looks good then.

docs/src/main/asciidoc/getting-started-guide.adoc
gsmet: I don't think it's an issue. I would keep the tip and add it to the 2 `curl` instances.

If it was a problem big enough for someone to create a PR to fix it, I think it's worth having a clean behavior.

docs/src/main/asciidoc/getting-started-guide.adoc
```diff
@@ -191,6 +207,14 @@ hello
 
 Hit `CTRL+C` to stop the application, but you can also keep it running and enjoy the blazing fast hot-reload.
 
+[TIP]
+Automatically add newline with curl
```
gsmet: This one misses the `.` at the beginning, breaking the TIP layout.

docs/src/main/asciidoc/getting-started-guide.adoc
```diff
@@ -246,13 +270,20 @@ public class GreetingResource {
...
+        return "hello";
     }
 }
 ----
 
 If you stopped the application, restart the application with `./mvnw compile quarkus:dev`.
-Then check that http://localhost:8080/hello/greeting/quarkus returns `hello quarkus`.
+Then check that the endpoint returns `hello quarkus` as expected:
+
+```
```
gsmet: This is Markdown, please use a `[source]` block.

gsmet:
I found a couple of formatting errors while reviewing the output, could you fix them?

docs/src/main/asciidoc/getting-started-guide.adoc
gsmet: I think we can be very descriptive in the very first guide for beginners and suppose they are able to do a git clone in the others. I don't think it's worth changing all the guides.

docs/src/main/asciidoc/getting-started-guide.adoc
maxandersen: fixing.

docs/src/main/asciidoc/getting-started-guide.adoc
maxandersen: whoops - sorry, bad habit :) fixed.

docs/src/main/asciidoc/getting-started-guide.adoc
maxandersen: okey - fixed in latest commit.

docs/src/main/asciidoc/getting-started-guide.adoc
maxandersen: fixed in latest commit.

docs/src/main/asciidoc/getting-started-guide.adoc
maxandersen: fixed in latest commit.

maxandersen:
> I found a couple of formatting errors while reviewing the output, could you fix them?

pushed a commit with those fixes.

gsmet (approved):

gsmet:
Merged, thanks!

PR #3879, opened 2019-09-05
Autocompile new projects before starting dev mode

evanchooly:
silently invoke the compile target if the application has not been built yet

### DIFF
--- a/integration-tests/maven/src/test/java/io/quarkus/maven/it/DevMojoIT.java
+++ b/integration-tests/maven/src/test/java/io/quarkus/maven/it/DevMojoIT.java
@@ -460,7 +460,7 @@ public class DevMojoIT extends RunAndCheckMojoTestBase {
     }
 
     @Test
-    public void testErrorMessageWhenNoTargetClasses() throws IOException, MavenInvocationException {
+    public void testNoErrorMessageWhenNoTargetClasses() throws IOException, MavenInvocationException {
         testDir = initProject("projects/classic", "projects/project-no-classes");
         new File(testDir, "target").mkdirs();
         // Be sure we don't have classes.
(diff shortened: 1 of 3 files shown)


kenfinnigan (approved):

aloubyansky:
I'm curious, have you tried compiling the project then changing some source file and then `mvn quarkus:dev`? Does it pick up the latest change?

evanchooly:
it does.  everything else works as usual.  it just doesn't fail on that initial start anymore.

aloubyansky:
Great. Another question, have you tried injecting an instance of the invoker into the mojo? i.e.
````
    @Component
    private Invoker invoker;
````
to make sure it is properly initialized.

**Comment by evanchooly** [comment]
I have not!  I didn't realize it was an option but it makes sense to try.  I'll let you know.

**Comment by evanchooly** [comment]
Worked like a champ.  Good call.

**Review by aloubyansky — APPROVED** [review]

**Comment by gsmet** [comment]
@cescoffier I know you had reservations about how we should do that when we talk about it several months ago. Is it OK for you?

(CI doesn't seem to like it but I'm not sure it's related)

**Comment by evanchooly** [comment]
That failure seemed to be a timeout issue on another test.

**Comment by evanchooly** [comment]
found and fixed the errant tests.

**Comment by evanchooly** [comment]
@cescoffier if there are no objections, i'll merge...

File: integration-tests/maven/src/test/java/io/quarkus/maven/it/DevMojoIT.java [hunk 1]
```diff
@@ -453,25 +452,20 @@ public void testThatNewBeanAreDiscovered() throws IOException, MavenInvocationEx
     }
 
     @Test
-    public void testErrorMessageWhenNoTarget() throws IOException, MavenInvocationException {
+    public void testNoErrorMessageWhenNoTarget() throws IOException, MavenInvocationException {
```
machi1990: Is this test still needed with auto-compile?

evanchooly: probably because it's testing missing folders at two distinct levels.  one could probably make an argument for removing it but the extra coverage is probably worth it.

PR #3880, opened 2019-09-05
Add support for Spring's @Value in a constructor

geoand:
The current codebase only support using `@Value` for field injection, but Spring allows it to be used for constructor injection as well

### DIFF
--- a/integration-tests/spring-di/src/main/java/io/quarkus/it/spring/GreeterBean.java
+++ b/integration-tests/spring-di/src/main/java/io/quarkus/it/spring/GreeterBean.java
@@ -18,11 +18,11 @@ public class GreeterBean {
     @Qualifier("cap")
     StringFunction capitalizerStringFunction;
 
-    @Value("${greeting.suffix:!}")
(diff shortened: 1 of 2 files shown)


integration-tests/spring-di/src/main/java/io/quarkus/it/spring/GreeterBean.java
```diff
@@ -18,11 +18,11 @@
     @Qualifier("cap")
     StringFunction capitalizerStringFunction;
 
-    @Value("${greeting.suffix:!}")
-    String suffix;
```
gsmet: Do we still have a test for that elsewhere?

geoand: It sure does: https://github.com/quarkusio/quarkus/blob/master/integration-tests/spring-di/src/main/java/io/quarkus/it/spring/MessageProducer.java#L8

gsmet (approved):

PR #3885, opened 2019-09-05
Set timeout for each native build to detect unexpected duration increase

cescoffier:
Set timeout for each native build to detect unexpected duration increase.
The times set in the PR are roughly computed based on the last 3 duration + 20%.


geoand (approved):
:+1: 

The timeouts seem reasonable

PR #3887, opened 2019-09-05
Add support for Asciidoc in configuration reference documentation

gsmet:
/cc @emmanuelbernard @FroMage @machi1990 

Here it is. It's a bit more than a one liner as I had to restructure the parser and the output a bit. It fixes a logic issue anyway so it's all for good.

I haven't written the extension guide documentation yet but let's agree on this first.

One thing that is a bit less nice compared to the handcrafted output is that we lost the nice titled sections but I think having everything generated is worth it.

At some point, we could imagine a new tag to generate proper sections but let's do that another day.

Once Asciidoclet has the related patch, we will have a proper output in the generated Javadoc too. 

@machi1990 could you upgrade your repository with this PR and point to it so that Emmanuel can have a look?

Note that GitHub does not support the new Font Awesome icons, I will attach a few screenshots.

### Summary table

picture1

### Details with Asciidoc

picture2


machi1990:
> @machi1990 could you upgrade your repository with this PR and point to it so that Emmanuel can have a look?

Repository update - https://github.com/machi1990/quarkus-generated-extension-config-docs/blob/master/src/main/asciidoc/generated/quarkus-hibernate-orm.adoc 

> Note that GitHub does not support the new Font Awesome icons, I will attach a few screenshots. 

Good thing you attached screenshots. GH did not like the icons :-P

extensions/hibernate-orm/deployment/src/main/java/io/quarkus/hibernate/orm/deployment/HibernateOrmConfig.java
```diff
@@ -117,27 +151,29 @@ public boolean isAnyPropertySet() {
...
-         * Control how schema generation is happening in Hibernate ORM.
-         * <p>
-         * Same as JPA's javax.persistence.schema-generation.database.action.
+         * Select whether the database schema is generated or not.
+         *
+         * `drop-and-create` is awesome in development mode.
+         *
+         * Accepted values: `none`, `create`, `drop-and-create`, `drop`, `update`. The default is `none`.
          */
         @ConfigItem
```
machi1990: Maybe add a `defaultValue="none"` to the annotation so that it is nicely added to the summary table :-)

gsmet: Yes, I have to tweak things a bit regarding that. And wondering if we could use the enum directly now but I have to test it. It's on my list of things to improve.

machi1990: An enum would be even better.

gsmet: Yup, I'm just not sure it will fly, I need to check it and wanted this PR out of the door.

I will fix the default value though. Small improvements are better than later improvements :).

gsmet: Done for the default value.

machi1990 (approved):

extensions/hibernate-orm/deployment/src/main/java/io/quarkus/hibernate/orm/deployment/HibernateOrmConfig.java
```diff
@@ -95,15 +127,17 @@ public boolean isAnyPropertySet() {
...
         public Optional<String> queryPlanCacheMaxSize;
 
         /**
-         * The default ordering of nulls specific in the ORDER BY clause.
-         * <p>
-         * Valid values are: none, first, last.
+         * Default precedence of null values in `ORDER BY` clauses.
+         *
+         * Valid values are: `none`, `first`, `last`.
```
FroMage: Shouldn't we replace this `String` type with an enum?

extensions/hibernate-orm/deployment/src/main/java/io/quarkus/hibernate/orm/deployment/HibernateOrmConfig.java
```diff
@@ -117,27 +151,29 @@ public boolean isAnyPropertySet() {
...
 
         /**
-         * Control how schema generation is happening in Hibernate ORM.
-         * <p>
-         * Same as JPA's javax.persistence.schema-generation.database.action.
+         * Select whether the database schema is generated or not.
+         *
+         * `drop-and-create` is awesome in development mode.
+         *
+         * Accepted values: `none`, `create`, `drop-and-create`, `drop`, `update`. The default is `none`.
```
FroMage: Same here, this custom tag makes it impossible to extract the allowed values in the table, so probably should use an enum, no?

extensions/hibernate-orm/deployment/src/main/java/io/quarkus/hibernate/orm/deployment/HibernateOrmConfig.java
gsmet: Sure, see the discussion with Manyanda above. But I need to check if we can use an ORM one or if we have to create a new one so I consider it outside of the scope of this PR whose main purpose is to validate the Asciidoc support.

FroMage: Ah sorry I didn't see it. Sure it can come in another PR.

machi1990:
> I improved the output a bit by adding a light separator between each entry of the detailed output. Things are far more readable this way.

tracking progress repo updated

emmanuelbernard (approved):

emmanuelbernard:
Boom, its in!

PR #3893, opened 2019-09-05
Make RestAssured and TestHTTPResourceManager aware of configured servlet context path

### LINKED ISSUE #3876: RestAssured unaware of quarkus.servlet.context-path, revisited
**Describe the bug**

This is a follow-up to #3551, which I believe was closed without adequate consideration.

**Expected behavior**

If I have a Quarkus app with passing `RestAssured` tests and `@SubstrateTest` variants, I can add

```properties
quarkus.servlet.context-path=/something
```

and tests will still pass.

**Actual behavior**

All tests fail.

**To Reproduce**

I do not currently have a public demo app reproducing the issue, though I think it would be easy to make one based on the above description.

**Configuration**

See above.

**Screenshots**

N/A

 - Output of `uname -a` or `ver`: `Linux … 5.0.0-25-generic #26-Ubuntu SMP Thu Aug 1 12:04:58 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux`
 - Output of `java -version`:
```
openjdk version "1.8.0_222"
OpenJDK Runtime Environment (build 1.8.0_222-8u222-b10-1ubuntu1~19.04.1-b10)
OpenJDK 64-Bit Server VM (build 25.222-b10, mixed mode)
```
 - GraalVM version (if different from Java): `quay.io/quarkus/centos-quarkus-maven:19.2.0`
 - Quarkus version or git rev: 0.21.2

**Additional context**

https://github.com/quarkusio/quarkus/issues/3551#issuecomment-522217871 suggests a workaround of simply not running tests with any context path, though I find this unacceptable since there are parts of the application that I _know_ behave differently when a context path is specified and I need these to be tested.

 It is possible to fix the JVM-mode tests by injecting

```java
@ConfigProperty(name = "quarkus.servlet.context-path") String contextPath;
```

and prepending `contextPath +` to any requested URIs (for example `RequestSenderOptions.get(String, Object...)`), but this does not work in native mode as the field will be null. The workaround I have found is to prepend `getContextPath() +` to requested URIs given

```java
@ConfigProperty(name = "quarkus.servlet.context-path") String contextPath;
/** must match {@code application.properties#quarkus.servlet.context-path} */
private static final String CONTEXT_PATH = "/something";
protected final String getContextPath() {
    if (contextPath != null) {
        assertEquals(CONTEXT_PATH, contextPath);
    } // else native mode
    return CONTEXT_PATH;
}
```

which allows tests to pass in native mode, by copying the context path, yet ensures that tests will fail if the copy gets out of synch with the real value.

While I have not yet tried it, it may also work to inject a `@TestHTTPResource URL` and parse out the `URL.path`.

The purported resolution was a documentation update #3684 which is worded rather vaguely and includes no code samples, so it is hard to evaluate whether it describes adequate workarounds. The ideal resolution from my perspective would be that `RestAssured` is somehow preconfigured by Quarkus to automatically prepend the context path to all requests, so that no other code changes need to be made when newly configuring a context path.

gwenneg:
Fixes #3876

### DIFF
--- a/docs/src/main/asciidoc/getting-started-testing.adoc
+++ b/docs/src/main/asciidoc/getting-started-testing.adoc
@@ -152,7 +152,7 @@ quarkus.http.test-port=8083
 ----
 ====
 
-Quarkus also provides Restassured integration that updates the default port used by Restassured before the tests are run,
+Quarkus also provides RestAssured integration that updates the default port used by RestAssured before the tests are run,
 so no additional configuration should be required.
 
 === Injecting a URI
@@ -223,11 +223,6 @@ public class StaticContentTest {
 <1> This annotation allows you to directly inject the URL of the Quarkus instance, the value of the annotation will be the path component of the URL
 
 For now `@TestHTTPResource` allows you to inject `URI`, `URL` and `String` representations of the URL.
-
-[TIP]
-=====
-If `quarkus.servlet.context-path` is configured, make sure to prefix the endpoint with the provided value acting as a base path during injection of a `URI` through `TestHTTPResource` or when calling a rest endpoint with `RestAssured`
-=====
 
 == Injection into tests
 
--- a/test-framework/common/src/main/java/io/quarkus/test/common/RestAssuredURLManager.java
+++ b/test-framework/common/src/main/java/io/quarkus/test/common/RestAssuredURLManager.java
@@ -1,6 +1,7 @@
 package io.quarkus.test.common;
 
 import java.lang.reflect.Field;
+import java.util.Optional;
 
 import org.eclipse.microprofile.config.ConfigProvider;
 
@@ -18,26 +19,33 @@ public class RestAssuredURLManager {
 
     private static final Field portField;
     private static final Field baseURIField;
+    private static final Field basePathField;
     private int oldPort;
     private String oldBaseURI;
+    private String oldBasePath;
 
     private final boolean useSecureConnection;
 
     static {
         Field p;
-        Field base;
+        Field baseURI;
+        Field basePath;
         try {
             Class<?> restAssured = Class.forName("io.restassured.RestAssured");
             p = restAssured.getField("port");
             p.setAccessible(true);
-            base = restAssured.getField("baseURI");
-            base.setAccessible(true);
+            baseURI = restAssured.getField("baseURI");
+            baseURI.setAccessible(true);
+            basePath = restAssured.getField("basePath");
+            basePath.setAccessible(true);
         } catch (Exception e) {
             p = null;
-            base = null;
+            baseURI = null;
+            basePath = null;
         }
         portField = p;
-        baseURIField = base;
+        baseURIField = baseURI;
+        basePathField = basePath;
     }
 
     public RestAssuredURLManager(boolean useSecureConnection) {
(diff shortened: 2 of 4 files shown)


test-framework/common/src/main/java/io/quarkus/test/common/RestAssuredURLManager.java
```diff
@@ -70,6 +78,18 @@ public void setURL() {
                 e.printStackTrace();
             }
         }
+        if (basePathField != null) {
+            try {
+                oldBasePath = (String) basePathField.get(null);
+                Optional<String> basePath = ConfigProvider.getConfig().getOptionalValue("quarkus.servlet.context-path",
```
machi1990: Do we need to handle `%test.quarkus.servlet.context-path` too?

docs/src/main/asciidoc/getting-started-testing.adoc
```diff
@@ -226,7 +226,8 @@ For now `@TestHTTPResource` allows you to inject `URI`, `URL` and `String` repre
 
 [TIP]
 =====
-If `quarkus.servlet.context-path` is configured, make sure to prefix the endpoint with the provided value acting as a base path during injection of a `URI` through `TestHTTPResource` or when calling a rest endpoint with `RestAssured`
+If `quarkus.servlet.context-path` is configured, make sure to prefix the endpoint with the provided value acting as a base path
+during injection of a `URI` through `TestHTTPResource`.
```
machi1990: Can we automatically configure this too?

test-framework/common/src/main/java/io/quarkus/test/common/RestAssuredURLManager.java
gwenneg: Well, if I'm right the configuration profile resolution is already done at that point, so I'd answer no.

docs/src/main/asciidoc/getting-started-testing.adoc
gwenneg: Good point, I'll take a look at it tomorrow.

gwenneg: It turns out the current `TestHTTPResourceManager` implementation has a bug which prevents the context path from being added to the URL. When several `QuarkusUnitTest` are run from the same project, the `TestHTTPResourceManager` static intializer is only called once when the first `QuarkusUnitTest` is run. Therefore, none of the configuration values (host and port in the current version) are correctly loaded.

@gsmet Is it OK if I try fixing the `TestHTTPResourceManager` issue in this PR? I don't know how far we are from `0.22.0`.

gwenneg:
It might actually be quicker to fix than what I expected. Let's see if CI is happy with this.

test-framework/common/src/main/java/io/quarkus/test/common/http/TestHTTPResourceManager.java
```diff
@@ -12,43 +12,36 @@
...
-        for (TestHTTPResourceProvider i : ServiceLoader.load(TestHTTPResourceProvider.class)) {
-            map.put(i.getProvidedType(), i);
-        }
-        providers = Collections.unmodifiableMap(map);
-        Config config = ConfigProvider.getConfig();
-        String host = config.getOptionalValue("quarkus.http.host", String.class).orElse("localhost");
-        String port = config.getOptionalValue("quarkus.http.test-port", String.class).orElse("8081");
-        uri = "http://" + host + ":" + port;
-        System.setProperty("test.url", uri);
+        System.setProperty("test.url", getUri());
```
gwenneg: I'm not sure about this line. It could still be subject to the bug observed while injecting the URL...

emmanuelbernard:
@machi1990 let me know if we are good to go?

test-framework/common/src/main/java/io/quarkus/test/common/RestAssuredURLManager.java
dmlloyd: Correct.

machi1990 (approved):
@emmanuelbernard this looks good to me.

@gwenneg thanks for taking into account my review.

geoand (approved):
Thanks!

PR #3899, opened 2019-09-06
fix the link to the distroless image documentation

cescoffier:
The instructions to build a distroless image are mostly a documentation page with some script. It does not depend on specific GraalVM versions. This explains why it target `master`.


cescoffier:
Fix https://github.com/quarkusio/quarkusio.github.io/issues/288

gsmet (approved):

PR #3901, opened 2019-09-06
list static resources

### LINKED ISSUE #3851: Improvements for 404
I got confused as I tried to reach a html page and got a Resource Not found page which showed me REST endpoint options
took me a while to figure that out. Here is what I propose to improve it:

- ~clearly show 404 in the page~ (done here: https://github.com/quarkusio/quarkus/pull/3854)
- ideally display the list of static resources if we know them
- ~give an intro before the list of rest endpoints "If you're looking for your REST resources, here are the ones we discovered:" soemthing like that~ (done here: https://github.com/quarkusio/quarkus/pull/3854)

Dufgui:
fix #3851

### DIFF
--- a/core/runtime/src/main/java/io/quarkus/runtime/TemplateHtmlBuilder.java
+++ b/core/runtime/src/main/java/io/quarkus/runtime/TemplateHtmlBuilder.java
@@ -23,7 +23,7 @@ public class TemplateHtmlBuilder {
             "</header>\n" +
             "<div class=\"container content\">\n";
 
-    private static final String RESOURCES_START = "<div class=\"intro\">If you are looking for your REST resources, here are the ones we discovered:</div><div class=\"resources\">";
+    private static final String RESOURCES_START = "<div class=\"intro\">%1$s</div><div class=\"resources\">";
 
     private static final String RESOURCE_TEMPLATE = "<h3>%1$s</h3>\n";
 
@@ -148,8 +148,8 @@ public class TemplateHtmlBuilder {
         return this;
     }
 
-    public TemplateHtmlBuilder resourcesStart() {
-        result.append(RESOURCES_START);
+    public TemplateHtmlBuilder resourcesStart(String title) {
+        result.append(String.format(RESOURCES_START, title));
         return this;
     }
 
@@ -164,8 +164,18 @@ public class TemplateHtmlBuilder {
     }
 
     public TemplateHtmlBuilder resourcePath(String title) {
+        return resourcePath(title, true);
+    }
+
+    public TemplateHtmlBuilder staticResourcePath(String title) {
+        return resourcePath(title, false);
+    }
+
+    private TemplateHtmlBuilder resourcePath(String title, boolean withListStart) {
         result.append(String.format(RESOURCE_TEMPLATE, escapeHtml(title)));
-        result.append(LIST_START);
+        if (withListStart) {
+            result.append(LIST_START);
+        }
         return this;
     }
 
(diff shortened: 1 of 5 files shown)


geoand:
@emmanuelbernard would you like to verify this since you opened the original issue?

Dufgui:
ready for review @emmanuelbernard

gsmet (changes requested):
Please wait for this PR https://github.com/quarkusio/quarkus/pull/3854/files to be in, before merging these changes.

core/runtime/src/main/java/io/quarkus/runtime/TemplateHtmlBuilder.java
```diff
@@ -12,17 +12,28 @@
...
             "<div class=\"container content\">\n";
 
+    private static final String HEADER_END = "</div>\n";
+
+    private static final String HEADER_TEMPLATE = "<header>\n" +
+            "    <div class=\"exception-message\">\n" +
+            "        <h2 class=\"container\">%1$s</h2>\n" +
+            "    </div>\n" +
+            "</header>\n" +
+            "<div class=\"container content\">\n";
```
gsmet: These changes don't look related to the issue at hand and don't seem like an improvement to me.

core/runtime/src/main/java/io/quarkus/runtime/TemplateHtmlBuilder.java
```diff
@@ -12,17 +12,28 @@
             "</head>\n" +
             "<body>\n";
 
-    private static final String HTML_TEMPLATE_END = "</div></body>\n" +
+    private static final String HTML_TEMPLATE_END = "</body>\n" +
```
gsmet: Again, looks unrelated. Moreover, I think it's wrong because the header explicitly open a div for the content of the page.

Can you get rid of it?

core/runtime/src/main/java/io/quarkus/runtime/TemplateHtmlBuilder.java
```diff
@@ -66,7 +77,7 @@
             "    padding: 1em 0 1em 0;\n" +
             "}\n" +
             "\n" +
-            "header, .component-name {\n" +
+            ".component-name {\n" +
```
gsmet: This change should be reverted too.

extensions/resteasy/runtime/src/main/java/io/quarkus/resteasy/runtime/NotFoundExceptionMapper.java
```diff
@@ -179,16 +182,33 @@ private Response respond(List<ResourceDescription> descriptions) {
...
             sb.resourceEnd();
         }
+
         if (descriptions.isEmpty()) {
             sb.noResourcesFound();
         }
+        sb.headerEnd();
+
+        if (!staticResources.isEmpty()) {
+            sb.header("Static Resources");
```
gsmet: You're using the header for something it shouldn't be used. That's why you have all those unrelated changes.

Dufgui:
I remove the header, but I dislike the page layout now:
image

stuartwdouglas:
IMHO we should only be listing .htm and .html files, otherwise this will potentially list a massive amount of files that the developer is not interested in.

emmanuelbernard:
So a few things "Resource Not Found" is not enough, we need to add back the 404 too "404 - Resource Not Found".

I agree with Stuart that we should list the htm / html files as priority and if there are 0 of them, list the other ones @Dufgui

Dufgui:
Changed done ;o)
@gsmet @emmanuelbernard @stuartwdouglas

Dufgui:
@gsmet  any time to review ?

gsmet:
@Dufgui could you close this one and merge everything in the other one? (or the other way around)

It's really a bad idea to have 2 different PRs for more or less the same thing.

Dufgui:
replace by https://github.com/quarkusio/quarkus/pull/3921

PR #3902, opened 2019-09-06
Log a warning if a non-singleton bean declares a Gauge

jmartisk:
This might save some confusion and trouble detected too late.
Other option would be to just disallow this completely (MP Metrics spec allows throwing a validation error), but the argument against this is that there might be cases where having a Gauge on a `@Dependent` bean is valid, if the developer makes sure there will be just one instance of the bean for the whole application lifecycle. We don't know for sure.

### DIFF
--- a/extensions/smallrye-metrics/deployment/src/main/java/io/quarkus/smallrye/metrics/deployment/SmallRyeMetricsProcessor.java
+++ b/extensions/smallrye-metrics/deployment/src/main/java/io/quarkus/smallrye/metrics/deployment/SmallRyeMetricsProcessor.java
@@ -146,17 +146,25 @@ public class SmallRyeMetricsProcessor {
                     return;
                 }
 
-                for (DotName annotationName : clazz.annotations().keySet()) {
-                    if (GAUGE.equals(annotationName)) {
-                        ctx.transform().add(AnnotationInstance.create(METRICS_BINDING,
-                                ctx.getTarget(), new AnnotationValue[0]))
-                                .done();
-                        return;
+                if (clazz.annotations().keySet().contains(GAUGE)) {
+                    BuiltinScope beanScope = BuiltinScope.from(clazz);
+                    if (!isJaxRsEndpoint(clazz) && beanScope != null &&
+                            !beanScope.equals(BuiltinScope.APPLICATION) &&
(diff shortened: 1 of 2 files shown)


geoand:
@mkouba could you take a quick look please?
Thanks

extensions/smallrye-metrics/deployment/src/main/java/io/quarkus/smallrye/metrics/deployment/SmallRyeMetricsProcessor.java
```diff
@@ -148,6 +148,18 @@ void annotationTransformers(BuildProducer<AnnotationsTransformerBuildItem> trans
 
                 for (DotName annotationName : clazz.annotations().keySet()) {
```
mkouba: Minor impl. detail - `clazz.annotations().keySet().contains(GAUGE)` would be probaly more performant than iterating over all keys...

mkouba (commented):
Looks good. I've added two minor comments.

extensions/smallrye-metrics/deployment/src/main/java/io/quarkus/smallrye/metrics/deployment/SmallRyeMetricsProcessor.java
```diff
@@ -148,6 +148,18 @@ void annotationTransformers(BuildProducer<AnnotationsTransformerBuildItem> trans
...
                     if (GAUGE.equals(annotationName)) {
+                        BuiltinScope beanScope = BuiltinScope.from(clazz);
+                        if (!isJaxRsEndpoint(clazz) && beanScope != null &&
```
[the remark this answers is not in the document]
jmartisk: Yeah. It does check for `beanScope != null` and if it is null, then we don't assume anything about the scope and don't log the warning.

extensions/smallrye-metrics/deployment/src/main/java/io/quarkus/smallrye/metrics/deployment/SmallRyeMetricsProcessor.java
jmartisk: Good point, changed that

mkouba (approved):

PR #3915, opened 2019-09-07
Remove dependency on junit4 JAR in quarkus-hibernate-orm-deployment tests

gastaldi:
The test is already running on JUnit 5, but using the `Assert`class from Junit4

### DIFF
--- a/extensions/hibernate-orm/deployment/src/test/java/io/quarkus/hibernate/orm/enhancer/HibernateEntityEnhancerTest.java
+++ b/extensions/hibernate-orm/deployment/src/test/java/io/quarkus/hibernate/orm/enhancer/HibernateEntityEnhancerTest.java
@@ -1,5 +1,8 @@
 package io.quarkus.hibernate.orm.enhancer;
 
+import static org.junit.jupiter.api.Assertions.assertFalse;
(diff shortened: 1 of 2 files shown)


extensions/hibernate-orm/deployment/src/test/java/io/quarkus/hibernate/orm/enhancer/HibernateEntityEnhancerTest.java
```diff
@@ -1,5 +1,7 @@
 package io.quarkus.hibernate.orm.enhancer;
 
+import static org.junit.jupiter.api.Assertions.*;
```
gsmet: Can you avoid the star imports? We are trying to avoid them.

Other than that, it looks good, nice catch.

gsmet (changes requested):

extensions/hibernate-orm/deployment/src/test/java/io/quarkus/hibernate/orm/enhancer/HibernateEntityEnhancerTest.java
gastaldi: Good catch. Done

gsmet (approved):

PR #3923, opened 2019-09-09
Split up the native jobs further to take advantage of the extra agents

stuartwdouglas:


stuartwdouglas:
This new approach uses an extra 10m of total agent time per build, but if the CI is not backed up it can complete a job around 25m faster, as the native runs start after 10m rather than 35m

stuartwdouglas:
If all 20 agents are free this can now do the build in under 40m

gsmet (approved):

PR #3912, opened 2019-09-06
Add Jackson Databind substitutions to vert.x

geoand:
This is needed in order to be able to build native images without Jackson databind on the classpath.

This work was done using in progress Vert.x work that will end up in version `3.8.2` the release of which is of course a prerequisite to merging this.
It goes without saying that CI will fail until we have such a release

PR #3924, opened 2019-09-09
Fixes #3907, move Vert.x producers to Resteasy

### LINKED ISSUE #3907: Check why Vert.x extension is depending on jboss-jaxrs-api_2.1_spec
In `extensions/vertx/runtime/pom.xml`, we have a dependency to `jboss-jaxrs-api_2.1_spec`. We should check if it's really necessary and remove it if we can.

stuartwdouglas:
Resteasy will always run on top of vert.x in the new model
so this dependency arrangement makes more sense

### DIFF
--- a/extensions/resteasy-server-common/runtime/pom.xml
+++ b/extensions/resteasy-server-common/runtime/pom.xml
@@ -34,6 +34,10 @@
             <groupId>io.quarkus</groupId>
             <artifactId>quarkus-jaxb</artifactId>
         </dependency>
+        <dependency>
+            <groupId>io.vertx</groupId>
+            <artifactId>vertx-core</artifactId>
+        </dependency>
     </dependencies>
 
     <build>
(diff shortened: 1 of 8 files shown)


gsmet:
I'm not totally sure about that. I refrained from doing it when I saw the issue because I was wondering if it would bring a hard JSON dependency to the RESTEasy extension, which we want to avoid.

@geoand thoughts?

geoand:
Yeah, what we are aiming for is when Vert.x `3.8.2` comes out, the `quarkus-resteasy` (and related resteasy stuff) will be rid of Jackson (see the draft #3912).
However, I don't see this PR changing anything WRT that aim. @gsmet did you have something specific in mind?

gsmet:
I suspect `JsonObjectWriter` is using Jackson under the hood and will create a potential dependency.

And I think they will be registered as soon as we produce/consume JSON.

geoand:
I think I have missread the dependencies then. 
From my PoV, the final dependencies and json related classes that end up in the `quarkus-resteasy`  are the same. Whatever we had before we would continue to have if the PR were to be applied. Is that not the case?

gsmet:
Mmmh, maybe I'm missing something but the moved classes are in the user-facing Vert.x extension and not in the `vertx-core` extension.

So right now, they are not part of the `quarkus-resteasy` dependencies.

geoand:
Sorry, you're right! I got mixed up with the various vert.x extensions.

Let me sleep on it and let you know what I think tomorrow 😊

geoand:
@gsmet I see your point now. So yeah, I would say this kind of goes contratry to where we are trying to get which is to have the `quarkus-resteasy` not depend on any Jackson stuff (and which #3912 will allow once Vert.x `3.8.2` lands)

PR #3927, opened 2019-09-09
Add noDeps to the integration tests

stuartwdouglas:
This allows dev mode to just work for some quick and dirty testing

### DIFF
--- a/integration-tests/pom.xml
+++ b/integration-tests/pom.xml
@@ -76,7 +76,20 @@
                 </configuration>
             </plugin>
         </plugins>
+        <pluginManagement>
+            <plugins>
+                <plugin>
+                    <groupId>io.quarkus</groupId>
+                    <artifactId>quarkus-maven-plugin</artifactId>
+                    <version>${project.version}</version>
+                    <configuration>
+                        <noDeps>true</noDeps>
+                    </configuration>
+                </plugin>
+            </plugins>
+        </pluginManagement>
     </build>
+
 
     <dependencyManagement>
         <dependencies>


geoand (approved):

geoand:
Oops, I just noticed that I had added the wrong label. Thanks for fixing it @stuartwdouglas

emmanuelbernard:
What's the reason for that code change?

geoand:
@emmanuelbernard essentially both myself and @dmlloyd at different times tried to run `mvn quarkus:dev` on the integration tests in order to test stuff we were doing locally that could have an effect on the dev mode.
In both cases we ran into weird ClassLoader issues that we didn't realize were due to the fact that quarkus:dev was including the Quarkus classes as well as the "application" classes and wasted time trying to debug things we thought were related to our changes.
So the @stuartwdouglas proposed we add this configuration so `quarkus:dev` will work OOTB in those applications as well :)

stuartwdouglas:
It makes dev mode work for the integration test modules. The multi module support does not work when you have extensions in the same project

geoand:
@emmanuelbernard do you have any objections merging this one?

emmanuelbernard:
I'll merge but @geoand let's please add a documentation of when it's supposed to be used somewhere. I have zero chance of remembering this in 5 days.

geoand:
It seems like it's already there: https://github.com/quarkusio/quarkus/blob/57c537ff3a262aca2fa6bf868b81c967476154a2/docs/src/main/asciidoc/extension-authors-guide.adoc#L1155

Issue #3928, opened 2019-09-09
Exceptions in quarkus Maven augmentation output mangled


galderz:
Even with @gastaldi's [PR](https://github.com/quarkusio/quarkus/pull/3761) to make mojo logger synchronized, exceptions logger were being mangled, e.g.

```
[DEBUG] [io.quarkus.builder] Starting step "io.quarkus.jsonb.deployment.JsonbProcessor#build"
[DEBUG] [io.quarkus.builder] End step completed; 4 remaining
[DEBUG] [io.quarkus.builder] Dependency of "io.quarkus.deployment.steps.ConfigurationSetup#writeDefaults" finished; 2 remaining
[DEBUG] [io.quarkus.builder] Dependency of "io.quarkus.deploeployment.index.ency of "imNative Method)uildStep#addConfiguredIndexedDependencies"etty.
    o.quarkuat java.lang.Class.forNames.vertx.deployment.VertxProcessor (Class.jav#build" finished; 4 rea:315maning remaining
[m)  EBatG] [io.quarkus.builder] Dependencmy of "io.quarkus.vertx.web.deployment.VertxWeb io.netty.util.inProcessor#build" finished; 5 remaining
[    [1DEBUG] [at;3 io.quarku6s.netty.deployment.NettyProcessor.buildmio.quarkus.builder] Dependency of "io.quarkus.deplo (NettyProcessor.java:47ment.steps.SubstrateAutoFm)
   eatureStep#generateFeature" finished;  22 remainingProcessor] Register JCA class: sun.security.rsa.RSAKeyFactory$Legacyn 25 ms
DEBUG] [io.quarkus.elytron.secmaturity.deployment.SecurityDeploymentProcessor] jdk.internal.refle[m Regict.Nativster JCA claeMethodAccss: sun.security.rsa.RSAKeyPairGenerator$Legacy
essorImpl.invoke0[DEBUG] [io.quarkus.elytron.security.deployment.SecurityDeploymentProcessor] Register JCA class: sun.security.rsa.RSASignature$MD2withRSA
 (Native Method)y.buffer.UnpooledByteBufAllocat.quaror.<clinkuit>s.builderUnpooledByteBufAllocator.java:37)jboss.threads.LoggingUncaughtExceptionHandler@7262f848 to "org.jboss.threads.errors"
[...]

PR #3929, opened 2019-09-09
Force Netty to use an internal logger based on JBoss logging

galderz:
With this, exceptions appear ok:

```
[DEBUG] [io.quarkus.builder] Dependency of "io.quarkus.elytron.security.deployment.SecurityDeploymentProcessor#registerAdditionalBeans" finished; 0 remaining
[DEBUG] [io.quarkus.builder] Starting step "io.quarkus.elytron.security.deployment.SecurityDeploymentProcessor#registerAdditionalBeans"
[DEBUG] [io.netty.util.internal.PlatformDependent0] direct buffer constructor: unavailable
java.lang.UnsupportedOperationException: Reflective setAccessible(true) disabled
    at io.netty.util.internal.ReflectionUtil.trySetAccessible (ReflectionUtil.java:31)
    at io.netty.util.internal.PlatformDependent0$4.run (PlatformDependent0.java:224)
    at java.security.AccessController.doPrivileged (Native Method)
    at io.netty.util.internal.PlatformDependent0.<clinit> (PlatformDependent0.java:218)
    at io.netty.util.internal.PlatformDependent.isAndroid (PlatformDependent.java:213)
    at io.netty.util.internal.PlatformDependent.<clinit> (PlatformDependent.java:81)
    at io.netty.buffer.UnpooledByteBufAllocator.<clinit> (UnpooledByteBufAllocator.java:37)
    at io.netty.buffer.Unpooled.<clinit> (Unpooled.java:73)
    at io.netty.handler.codec.http.HttpObjectEncoder.<clinit> (HttpObjectEncoder.java:53)
    at java.lang.Class.forName0 (Native Method)
    at java.lang.Class.forName (Class.java:315)
    at io.quarkus.netty.deployment.NettyProcessor.build (NettyProcessor.java:50)
    at jdk.internal.reflect.NativeMethodAccessorImpl.invoke0 (Native Method)
    at jdk.internal.reflect.NativeMethodAccessorImpl.invoke (NativeMethodAccessorImpl.java:62)
    at jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke (DelegatingMethodAccessorImpl.java:43)
    at java.lang.reflect.Method.invoke (Method.java:566)
    at io.quarkus.deployment.ExtensionLoader$1.execute (ExtensionLoader.java:768)
    at io.quarkus.builder.BuildContext.run (BuildContext.java:415)
    at org.jboss.threads.ContextClassLoaderSavingRunnable.run (ContextClassLoaderSavingRunnable.java:35)
    at org.jboss.threads.EnhancedQueueExecutor.safeRun (EnhancedQueueExecutor.java:2011)
    at org.jboss.threads.EnhancedQueueExecutor$ThreadBody.doRunTask (EnhancedQueueExecutor.java:1535)
    at org.jboss.threads.EnhancedQueueExecutor$ThreadBody.run (EnhancedQueueExecutor.java:1426)
    at java.lang.Thread.run (Thread.java:834)
    at org.jboss.threads.JBossThread.run (JBossThread.java:479)
[DEBUG] [io.quarkus.builder] Dependency of "io.quarkus.arc.deployment.BeanArchiveProcessor#build" finished; 8 remaining
```

As mentioned on #3928, there might be easier ways to get this fixed. This is one way.

### DIFF
--- a/extensions/netty/deployment/src/main/java/io/quarkus/netty/deployment/NettyProcessor.java
+++ b/extensions/netty/deployment/src/main/java/io/quarkus/netty/deployment/NettyProcessor.java
@@ -9,6 +9,7 @@ import javax.inject.Inject;
 import org.jboss.logging.Logger;
 
 import io.netty.channel.EventLoopGroup;
+import io.netty.util.internal.logging.InternalLoggerFactory;
 import io.quarkus.arc.deployment.RuntimeBeanBuildItem;
 import io.quarkus.deployment.annotations.BuildProducer;
 import io.quarkus.deployment.annotations.BuildStep;
@@ -27,6 +28,10 @@ class NettyProcessor {
     BuildProducer<ReflectiveClassBuildItem> reflectiveClass;
 
     private static final Logger log = Logger.getLogger(NettyProcessor.class);
+
+    static {
+        InternalLoggerFactory.setDefaultFactory(new JBossNettyLoggerFactory());
+    }
 
     @BuildStep
     SubstrateConfigBuildItem build(BuildProducer<JniBuildItem> jni) {
--- a/devtools/maven/src/main/java/io/quarkus/maven/MojoLogger.java
+++ b/devtools/maven/src/main/java/io/quarkus/maven/MojoLogger.java
@@ -5,7 +5,6 @@ import java.util.Collections;
 import java.util.Map;
 import java.util.function.Supplier;
 
-import org.apache.maven.plugin.Mojo;
 import org.apache.maven.plugin.logging.Log;
 import org.apache.maven.shared.utils.logging.MessageBuilder;
 import org.apache.maven.shared.utils.logging.MessageUtils;
--- a/devtools/maven/src/main/java/io/quarkus/maven/MojoLogger.java
+++ b/devtools/maven/src/main/java/io/quarkus/maven/MojoLogger.java
@@ -5,6 +5,7 @@ import java.util.Collections;
 import java.util.Map;
 import java.util.function.Supplier;
 
+import org.apache.maven.plugin.Mojo;
 import org.apache.maven.plugin.logging.Log;
 import org.apache.maven.shared.utils.logging.MessageBuilder;
 import org.apache.maven.shared.utils.logging.MessageUtils;
--- a/extensions/netty/deployment/src/main/java/io/quarkus/netty/deployment/JBossNettyLoggerFactory.java
+++ b/extensions/netty/deployment/src/main/java/io/quarkus/netty/deployment/JBossNettyLoggerFactory.java
@@ -1,0 +1,177 @@
+package io.quarkus.netty.deployment;
+
+import org.jboss.logging.Logger;
+
+import io.netty.util.internal.logging.AbstractInternalLogger;
+import io.netty.util.internal.logging.InternalLogger;
+import io.netty.util.internal.logging.InternalLoggerFactory;
+
+public class JBossNettyLoggerFactory extends InternalLoggerFactory {
+
+    @Override
+    protected InternalLogger newInstance(String name) {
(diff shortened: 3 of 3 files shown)


emmanuelbernard:
Since there is rework at these layers, asking @stuartwdouglas for a quick review.

gsmet:
@galderz Is it me or the exceptions are not consistent with our usual JBoss Logging formatter? Or did you tweak it?

galderz:
> @galderz Is it me or the exceptions are not consistent with our usual JBoss Logging formatter? Or did you tweak it?

What do you mean exactly? The exception is as logged to the output. If you remove this patch and set augmentor thread pool to 1, you'll see the same.

dmlloyd:
The Mojo logger passes things through to Maven, so it's going to be using that format, not our default JBoss LogManager format.

dmlloyd (approved):

stuartwdouglas:
There is no guarantee when that build step will be run, so its possible that these changes may only take effect half way through the build. I think it would be better to install it in a static initialiser block, as AFAIK the BuildStep classes are all initialised early enough to make this guaranteed to work.

Also this is done at augment time, do we need something similar for runtime?

galderz:
> There is no guarantee when that build step will be run, so its possible that these changes may only take effect half way through the build. I think it would be better to install it in a static initialiser block, as AFAIK the BuildStep classes are all initialised early enough to make this guaranteed to work.

Makes sense, I'll try to address that.

> Also this is done at augment time, do we need something similar for runtime?

I'm not sure if it's such an issue at runtime, because I guess that can run outside of Maven? Image augmentation for native image happens for sure at Maven build time, which is what I was trying to debug.

gsmet (dismissed):
This one shouldn't be merged for now. See Stuart's comment.

galderz:
@stuartwdouglas @gsmet updated PR

extensions/netty/deployment/src/main/java/io/quarkus/netty/deployment/NettyProcessor.java
```diff
@@ -137,4 +138,10 @@ void createExecutors(BuildProducer<RuntimeBeanBuildItem> runtimeBeanBuildItemBui
                 .build());
     }
 
+    @BuildStep
+    @Record(ExecutionTime.STATIC_INIT)
+    void setJBossInternalLoggerFactory() {
```
stuartwdouglas: This was not what I meant (and should fail as it is not valid to have a @Record method that does not inject a recorder).

I just meant a static block to make sure it is run when this class is loaded:

  static {
    InternalLoggerFactory.setDefaultFactory(new JBossNettyLoggerFactory());
  }

Anything in a build step has the possibility that other build steps will be run first.

galderz: > This was not what I meant (and should fail as it is not valid to have a @record method that does not inject a recorder).

Ok... 🙈

There's [an example](https://github.com/quarkusio/quarkus/blob/master/core/test-extension/deployment/src/main/java/io/quarkus/extest/deployment/TestProcessor.java#L231) in the codebase of this.

galderz:
@stuartwdouglas Updated

stuartwdouglas (approved):
Hmm, the validation must have gone missing somewhere. It's not a big deal, it will just end up generating an empty method so it is kinda pointless.

geoand:
Should we merge this? Just asking because `waiting-for-ci` was added a few days ago and tests are passing

dmlloyd:
@gsmet ?

gsmet:
My review was just there to prevent merging it accidentally before addressing Stuart's concerns. Everyone seems to be happy now so let's merge.

PR #3930, opened 2019-09-09
Mailer extension: Fix broken test example

mejlholm:
Example doesn't work without @QuarkusTest. 
All other examples in the doc suggests sending Response.accepted, but the test case expects Response.ok. Lets align that.


emmanuelbernard (approved):

emmanuelbernard:
Thanks @mejlholm

PR #3932, opened 2019-09-09
docs(openapi): add section on how to publish custom OpenAPI design document

### LINKED ISSUE #3914: Document how to publish custom OpenAPI design document (instead of driven by the code)
Currently the OpenAPI generate schema from JAX-RS codes. This is code first way.

But some API description is not clear enough, we have to use the Microprofile OpenAPI annotations to enrich it.  Thus a lots of codes in Jax-rs are MP OpenAPI annotations.

In my experience, some projects require Design APIs firstly. There are some tools can do this, such as Swagger Editor. We can a well defined schema there.

Is it possible let OpenAPI load a predefined schema file(such as specify the file in classpath in the application.properties ) instead of generating from the JAX-RS codes?

machi1990:
Fixes #3914

/cc @emmanuelbernard

### DIFF
--- a/docs/src/main/asciidoc/openapi-swaggerui-guide.adoc
+++ b/docs/src/main/asciidoc/openapi-swaggerui-guide.adoc
@@ -284,6 +284,88 @@ quarkus.smallrye-openapi.path=/swagger
 
 Hit `CTRL+C` to stop the application.
 
+== Loading OpenAPI Schama From Static Files
+
+Instead of dynamically creating OpenAPI schemas from annotation scanning, Quarkus also supports serving static OpenAPI documents.
+The static file to serve must be a valid document conforming to the https://swagger.io/docs/specification[OpenAPI specification].
+An OpenAPI document that conforms to the  OpenAPI Specification is itself a valid JSON object, that can be represented in `yaml` or `json` formats.
+
+To see this in action, we'll put OpenAPI documentation under `META-INF/openapi.yaml` for our `/fruits` endpoints.
+Quarkus also supports alternative <<open-document-paths, OpenAPI document paths>> if you prefer.
+
+[source, shell]
+----
+openapi: 3.0.1
+info:
+  title: Static OpenAPI document of fruits resource
+  description: Fruit resources Open API documentation
+  version: "1.0"
+
+servers:
+  - url: http://localhost:8080/openapi
+    description: Optional dev mode server description
+
+paths:
+  /fruits:
+    get:
+      responses:
+        200:
+          description: OK - fruits list
+          content:
+            application/json: {}
+    post:
+      requestBody:
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/Fruit'
+      responses:
+        200:
+          description: new fruit resource created
+          content:
+            application/json: {}
+    delete:
+      requestBody:
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/Fruit'
+      responses:
+        200:
+          description: OK - fruit resource deleted
+          content:
+            application/json: {}
+components:
+  schemas:
+    Fruit:
+      properties:
+        description:
+          type: string
+        name:
+          type: string
+----
+By default, a request to `/openapi` will serve the combined OpenAPI document from the static file and the model generated from application endpoints code.
+We can however change this to only serve the static OpenAPI document by adding `mp.openapi.scan.disable=true` configuration into `application.properties`.
+
+Now, a request to `/openapi` endpoint will serve the static OpenAPI document instead of the generated one.
+
+[[open-document-paths]]
+[TIP]
+.About OpenAPI document paths
+====
+Quarkus supports various paths to store your OpenAPI document under. We recommend you place it under `META-INF/openapi.yml`.
+Alternative paths are:
+
+* `META-INF/openapi.yaml`
+* `META-INF/openapi.yml`
+* `META-INF/openapi.json`
+* `WEB-INF/classes/META-INF/openapi.yml`
+* `WEB-INF/classes/META-INF/openapi.yaml`
+* `WEB-INF/classes/META-INF/openapi.json`
+
+Live reload of static OpenAPI document is supported during development. A modification to your OpenAPI document will be picked up on fly by Quarkus.
+====
+
 == Use Swagger UI for development
 
 When building APIs, developers want to test them quickly. https://swagger.io/tools/swagger-ui/[Swagger UI] is a great tool


machi1990:
I can request for a review, so I am assigning this one to you @emmanuelbernard.

Also cc @stuartwdouglas and maybe @gsmet :-)

docs/src/main/asciidoc/openapi-swaggerui-guide.adoc
```diff
@@ -284,6 +284,93 @@ quarkus.smallrye-openapi.path=/swagger
...
+
+Now, a request to `/openapi` endpoint will serve the static OpenAPI document instead of the generated one.
+
+[[open-document-paths]]
+[TIP]
+.About various OpenAPI document paths
+====
+Quarkus supports various file paths for the OpenAPI document. These are:
+
+* `META-INF/openapi.yaml`
```
emmanuelbernard: Can you add a note that we recommend the META-INF location. It does not matter much which one but I think we should push for one canonical version and the short path is better.

docs/src/main/asciidoc/openapi-swaggerui-guide.adoc
```diff
@@ -284,6 +284,93 @@ quarkus.smallrye-openapi.path=/swagger
...
+* `WEB-INF/classes/META-INF/openapi.json`
+
+Live reload of static OpenAPI document is supported during development. A modification to your OpenAPI document will be picked up on fly by Quarkus.
+====
+
+[TIP]
+.Disabling annotation scanning
+====
+When serving OpenAPI document from a static file, it is interesting to disable the dynamic generation of OpenAPI document through annotation scanning which will otherwise ovverride your static OpenAPI document.
+To do this, add the `mp.openapi.scan.disable=true` configuration to your `application.properties` file.
```
emmanuelbernard: Interesting, I wonder if the presence of a static file should just force the scanning to be disabled? Thoughts?

emmanuelbernard (changes requested):
Minor change requested and a question

docs/src/main/asciidoc/openapi-swaggerui-guide.adoc
emmanuelbernard: Also asked on the quarkus-dev mailing list

machi1990: Then this will force the check of the presence of "any" of the static files. Having a configuration that's document is enough IMO. 

I wonder what others think? @kenfinnigan @gsmet

docs/src/main/asciidoc/openapi-swaggerui-guide.adoc
machi1990: I'll say the first one `META-INF/openapi.yaml` hence my usage of it. But how do we tune that into a proper recommendation phrase? :-)

docs/src/main/asciidoc/openapi-swaggerui-guide.adoc
machi1990: > Also asked on the quarkus-dev mailing list

My page was not in sync so obvious I did not see this. 

Great. Let's wait for response.

kenfinnigan: Scanning isn't disabled because it combines the two models.

Currently you need to manually disable scanning with config

docs/src/main/asciidoc/openapi-swaggerui-guide.adoc
```diff
@@ -284,6 +284,93 @@ quarkus.smallrye-openapi.path=/swagger
 
 Hit `CTRL+C` to stop the application.
 
+== Loading OpenAPI Schama From Static Files
+
+Instead of dynamically creating OpenAPI schemas from annotation scanning, Quarkus also proposes serving static OpenAPI documents.
```
kenfinnigan: "proposes" -> "supports"?

docs/src/main/asciidoc/openapi-swaggerui-guide.adoc
```diff
@@ -284,6 +284,93 @@ quarkus.smallrye-openapi.path=/swagger
...
+The static file to serve must be a valid document conforming to the https://swagger.io/docs/specification[OpenAPI specification].
+An OpenAPI document that conforms to the  OpenAPI Specification is itself a valid JSON object, that can be represented in
+a `yaml` or `json` formats.
```
kenfinnigan: Remove "a"

docs/src/main/asciidoc/openapi-swaggerui-guide.adoc
```diff
@@ -284,6 +284,93 @@ quarkus.smallrye-openapi.path=/swagger
...
+
+To see this in action, we'll create a static OpenAPI document `META-INF/openapi.yaml` for our `/fruits` path.
+You can choose a different document path as indicating in the <<open-document-paths, document paths tip below>>
```
kenfinnigan: "indicating" -> "indicated"

docs/src/main/asciidoc/openapi-swaggerui-guide.adoc
```diff
@@ -284,6 +284,93 @@ quarkus.smallrye-openapi.path=/swagger
...
+  schemas:
+    Fruit:
+      properties:
+        description:
+          type: string
+        name:
+          type: string
+----
+
+Now, a request to `/openapi` endpoint will serve the static OpenAPI document instead of the generated one.
```
kenfinnigan: For this to be true I think scanning needs to be disabled, otherwise it merges the two models

machi1990: Yes, I have added a TIP just below.

kenfinnigan: I'm wondering if it needs to be made explicit here, as currently the example will serve the combined model and not just the OpenAPI static file

machi1990: Good point. So I have remove the tip and added a few sentences based on your suggestion. WDYT?

kenfinnigan: I think that looks good, thanks

docs/src/main/asciidoc/openapi-swaggerui-guide.adoc
machi1990: Added a recommendation sentence just above the list. Is that okay with you? @emmanuelbernard

emmanuelbernard: I don't see it but yes something like Put your OpenAPI document under `META-INF/openapi.yaml`. Quarkus support alternative paths if you prefer:

* ...

docs/src/main/asciidoc/openapi-swaggerui-guide.adoc
```diff
@@ -284,6 +284,88 @@ quarkus.smallrye-openapi.path=/swagger
...
+  schemas:
+    Fruit:
+      properties:
+        description:
+          type: string
+        name:
+          type: string
+----
+
+We'll also have to add the `mp.openapi.scan.disable=true` configuration to your `application.properties` file to disable the dynamic generation of OpenAPI document through annotation scanning.
```
emmanuelbernard: By default, it will combine in the OpenAPI document the static file and the model generated from your application endpoints code. You can however force Quarkus to only take your static openAPI document by using `mp.openapi.scan.disable=true`.

emmanuelbernard: The reason for the rephrase is that your sentences are extremely negative and feel like you are limiting options. Also We will have to seems like forced at gun point :)

machi1990: +1 with the rephrase. Should be good now. Tell me what you think.

docs/src/main/asciidoc/openapi-swaggerui-guide.adoc
```diff
@@ -284,6 +284,87 @@ quarkus.smallrye-openapi.path=/swagger
...
+  schemas:
+    Fruit:
+      properties:
+        description:
+          type: string
+        name:
+          type: string
+----
+By default, a request to `/openapi` will serve the combined OpenAPI document from the static file and the model generated from application endpoints code.
+We can however change this to only serve static OpenAPI document by adding `mp.openapi.scan.disable=true` configuration into `application.properties`.
```
emmanuelbernard: serve *the* static... and we are good to go.

docs/src/main/asciidoc/openapi-swaggerui-guide.adoc
```diff
@@ -284,6 +284,87 @@ quarkus.smallrye-openapi.path=/swagger
...
+
+Now, a request to `/openapi` endpoint will serve the static OpenAPI document instead of the generated one.
+
+[[open-document-paths]]
+[TIP]
+.About OpenAPI document paths
+====
+Quarkus supports various file paths for the static OpenAPI document as listed below.
```
emmanuelbernard: hum I don't see the recommended location, is that a regression? or GitHub weirdness?

emmanuelbernard (changes requested):

docs/src/main/asciidoc/openapi-swaggerui-guide.adoc
machi1990: Good catch. Done.

docs/src/main/asciidoc/openapi-swaggerui-guide.adoc
machi1990: Switching work stations "without-syncing-everything" weirdness. :-) Should be good now.

machi1990: Done.

emmanuelbernard (approved):

emmanuelbernard:
@machi1990 I tend to prefer `.yaml`? Any reason you chose `.yml`?

PR #3936, opened 2019-09-09
Generate default Tika config with the list of supported parsers

sberyozkin:
This PR ensures that when no custom configuration resource is provided, the default one is generated listing explicitly all the supported parsers (with the exception of the unsupported parsers), thus ensuring that the static init recording can work with or without the custom configuration resource


FroMage (approved):
LGTM

sberyozkin:
Hmm, was pretty sure it was green for me, let me check...

sberyozkin:
@FroMage, sorry, forgot it was only for the quickstart, 1 sec, application.properties are needed...

sberyozkin:
@gsmet, since Stephane has approved, I'm merging now. If there is anything in this PR that might concern still you then please ping me or open an issue and will do the follow up updates :-)

PR #3940, opened 2019-09-09
Retire quarkus-junit4

gastaldi:
This PR removes the `quarkus-junit4` and any references to the JUnit 4 JAR in the BOM. 

It also migrates the JUnit 4 tests to JUnit 5

Discussion in https://groups.google.com/forum/#!topic/quarkus-dev/2YFVhwYv2N4


bom/runtime/pom.xml
```diff
@@ -1096,11 +1089,6 @@
                 <artifactId>javax.ws.rs-api</artifactId>
                 <version>${javax.ws.rs-api.version}</version>
             </dependency>
-            <dependency>
```
kenfinnigan: Have you run this with TCKs enabled?

I suspect removing the junit dependency will break them

gastaldi: I browsed the POMs, and noticed they depend on `quarkus-arquillian`, which transitively resolves JUnit 4 through `arquillian-junit-container`, but I'll double check

kenfinnigan: The `common` module has a dependency on junit.

It might come in through Arq BOM

gastaldi: Not only that, but the tests in other modules (including the core) depend on JUnit4. I'll make sure to migrate them as well

gastaldi:
The PR is now ready to be reviewed.

stuartwdouglas (approved):

stuartwdouglas:
This LGTM, not sure if @kenfinnigan or @gsmet also want to look it over?

gsmet (approved):

rsvoboda:
@gsmet should we have `noteworthy-removal` label to track things like this for release notes ?

Issue #3943, opened 2019-09-10
Support injection of arguments into Arquillian test methods


jmartisk:
Some tests from the MP Metrics TCK are using this and for now we've had to exclude them.
CC @kenfinnigan @mkouba @manovotn

PR #3944, opened 2019-09-10
Injection into test method parameters in Arquillian tests

jmartisk:
#3943
@mkouba @kenfinnigan @manovotn please have a look. I don't feel very comfortable with the `convertToTCCL` thingy but for some reason it was necessary (Arquillian is passing the method metadata to `TestEnricher` and `TestMethodExecutor` loaded by system class loader). Any better solution to go with?

### DIFF
--- a/test-framework/arquillian/src/main/java/io/quarkus/arquillian/CreationalContextDestroyer.java
+++ b/test-framework/arquillian/src/main/java/io/quarkus/arquillian/CreationalContextDestroyer.java
@@ -1,0 +1,27 @@
+package io.quarkus.arquillian;
+
+import javax.enterprise.context.spi.CreationalContext;
+
+import org.jboss.arquillian.core.api.Instance;
+import org.jboss.arquillian.core.api.annotation.Inject;
+import org.jboss.arquillian.core.api.annotation.Observes;
+import org.jboss.arquillian.core.spi.EventContext;
+import org.jboss.arquillian.test.spi.event.suite.After;
+
+public class CreationalContextDestroyer {
+
+    @Inject
+    private Instance<CreationalContext> creationalContext;
+
+    public void destroy(@Observes EventContext<After> event) {
+        try {
+            event.proceed();
+        } finally {
+            CreationalContext<Object> cc = creationalContext.get();
+            if (cc != null) {
+                cc.release();
+            }
+        }
+    }
+
+}
--- a/test-framework/arquillian/src/main/java/io/quarkus/arquillian/InjectionEnricher.java
+++ b/test-framework/arquillian/src/main/java/io/quarkus/arquillian/InjectionEnricher.java
@@ -1,0 +1,83 @@
+package io.quarkus.arquillian;
+
+import static io.quarkus.arquillian.QuarkusProtocol.convertToTCCL;
+
+import java.lang.reflect.Method;
+
+import javax.enterprise.context.spi.CreationalContext;
+import javax.enterprise.inject.spi.BeanManager;
+
+import org.jboss.arquillian.core.api.InstanceProducer;
+import org.jboss.arquillian.core.api.annotation.Inject;
+import org.jboss.arquillian.test.spi.TestEnricher;
+import org.jboss.arquillian.test.spi.annotation.TestScoped;
+import org.jboss.logging.Logger;
+
+import io.quarkus.arc.Arc;
+
+/**
+ * Enricher that provides method argument injection.
+ */
+public class InjectionEnricher implements TestEnricher {
+
+    private static final Logger log = Logger.getLogger(TestEnricher.class.getName());
+
+    @Inject
+    @TestScoped
+    private InstanceProducer<CreationalContext> creationalContextProducer;
+
+    public BeanManager getBeanManager() {
+        return Arc.container().beanManager();
+    }
+
+    @SuppressWarnings("unchecked")
+    public CreationalContext<Object> getCreationalContext() {
+        CreationalContext<Object> cc = creationalContextProducer.get();
+        if (cc == null) {
+            cc = getBeanManager().createCreationalContext(null);
+            creationalContextProducer.set(cc);
+        }
+        return cc;
+    }
+
+    @Override
+    public void enrich(Object testCase) {
+
+    }
+
+    @Override
+    public Object[] resolve(Method method) {
+        Object[] values = new Object[method.getParameterTypes().length];
+        if (values.length > 0) {
+            BeanManager beanManager = getBeanManager();
+            if (beanManager == null) {
+                return values;
+            }
+            try {
+                // obtain the same method definition but from the TCCL
+                method = Thread.currentThread().getContextClassLoader()
+                        .loadClass(method.getDeclaringClass().getName())
+                        .getMethod(method.getName(), convertToTCCL(method.getParameterTypes()));
+            } catch (Throwable t) {
+                throw new RuntimeException(t);
+            }
+            Class<?>[] parameterTypes = method.getParameterTypes();
+            for (int i = 0; i < parameterTypes.length; i++) {
+                try {
+                    values[i] = getInstanceByType(beanManager, i, method);
+                } catch (Exception e) {
+                    log.warn("InjectionEnricher tried to lookup method parameter of type "
+                            + parameterTypes[i] + " but caught exception", e);
+                }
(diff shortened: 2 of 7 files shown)


test-framework/arquillian/src/main/java/io/quarkus/arquillian/InjectionEnricher.java
```diff
@@ -0,0 +1,93 @@
...
+        return values;
+    }
+
+    @SuppressWarnings("unchecked")
+    private <T> T getInstanceByType(BeanManager manager, final int position, final Method method) {
+        CreationalContext<?> cc = getCreationalContext();
+        return (T) manager.getInjectableReference(new MethodParameterInjectionPoint<T>(method, position, manager), cc);
+    }
+
+    private boolean isInjectClassPresent() {
```
mkouba: There is no need to do this check.

test-framework/arquillian/src/main/java/io/quarkus/arquillian/InjectionEnricher.java
```diff
@@ -0,0 +1,93 @@
...
+ */
+public class InjectionEnricher implements TestEnricher {
+    private static final String ANNOTATION_NAME = "javax.inject.Inject";
+
+    private static final Logger log = Logger.getLogger(TestEnricher.class.getName());
+
+    @SuppressWarnings("rawtypes")
+    @Inject
+    @TestScoped
+    private InstanceProducer<CreationalContext> creationalContextProducer;
```
mkouba: What's the purpose of this?

test-framework/arquillian/src/main/java/io/quarkus/arquillian/MethodParameterInjectionPoint.java
```diff
@@ -0,0 +1,114 @@
...
+        return null;
+    }
+
+    public Member getMember() {
+        return method;
+    }
+
+    public Set<Annotation> getQualifiers() {
+        Set<Annotation> qualifiers = new HashSet<>();
+        // TODO - how to properly determine which annotations are qualifiers? beanManager.isQualifier() can't be used
```
mkouba: Filtering annotations annotated with `@Qualifier` should work in 90% use cases...

test-framework/arquillian/src/main/java/io/quarkus/arquillian/InjectionEnricher.java
```diff
@@ -0,0 +1,93 @@
...
+                        .getMethod(method.getName(), convertToTCCL(method.getParameterTypes()));
+            } catch (Throwable t) {
+                throw new RuntimeException(t);
+            }
+            Class<?>[] parameterTypes = method.getParameterTypes();
+            for (int i = 0; i < parameterTypes.length; i++) {
+                try {
+                    values[i] = getInstanceByType(beanManager, i, method);
+                } catch (Exception e) {
+                    log.warn("CDIEnricher tried to lookup method parameter of type "
```
mkouba: `CDIEnricher`? :-)

mkouba (changes requested):
> I don't feel very comfortable with the convertToTCCL thingy but for some reason it was necessary (Arquillian is passing the method metadata to TestEnricher and TestMethodExecutor loaded by system class loader). Any better solution to go with?

It is needed. We use a different test instance due to quarkus-specific class loading issues.

test-framework/arquillian/src/main/java/io/quarkus/arquillian/InjectionEnricher.java
jmartisk: It's needed for getting injectable references, I basically took this from https://github.com/arquillian/arquillian-core/blob/1.4.1.Final/testenrichers/cdi/src/main/java/org/jboss/arquillian/testenricher/cdi/CDIInjectionEnricher.java#L103

manovotn: I think Arquillian needs it for this - https://github.com/arquillian/arquillian-core/blob/1.4.1.Final/testenrichers/cdi/src/main/java/org/jboss/arquillian/testenricher/cdi/CreationalContextDestroyer.java#L36
Which looks like automagical CC cleanup :)

test-framework/arquillian/src/main/java/io/quarkus/arquillian/MethodParameterInjectionPoint.java
```diff
@@ -0,0 +1,118 @@
...
+                qualifiers.add(potentialQualifier);
+            }
+        }
+        if (qualifiers.size() == 0) {
+            qualifiers.add(new DefaultLiteral());
+        }
+        return qualifiers;
+    }
+
+    private static class DefaultLiteral extends AnnotationLiteral<Default> implements Default {
```
manovotn: Shouldn't be needed from CDI 2.0
It is now built in - https://github.com/eclipse-ee4j/cdi/blob/master/api/src/main/java/javax/enterprise/inject/Default.java#L104

test-framework/arquillian/src/main/java/io/quarkus/arquillian/InjectionEnricher.java
jmartisk: Does that mean we should have a similar class in Quarkus too? We aren't importing that one

test-framework/arquillian/src/main/java/io/quarkus/arquillian/MethodParameterInjectionPoint.java
jmartisk: If I remove that I am getting ambiguous dependencies on injecting the application `MetricRegistry` from https://github.com/smallrye/smallrye-metrics/blob/2.1.4/implementation/src/main/java/io/smallrye/metrics/MetricRegistries.java#L40

mkouba: You should use `Default.Literal.INSTANCE` instead...

jmartisk: Ok will use that one

jmartisk: Ah, ok

test-framework/arquillian/src/main/java/io/quarkus/arquillian/InjectionEnricher.java
manovotn: Aren't we? In that case why exactly did you need `InstanceProducer`? What consumers were there?
And yes, we should have a place where we invoke `CreationalContext.release()` to properly destroy all dependent instances linked to this `CreationalContext`.

jmartisk: I guess you're right, in that case it wasn't quite necessary :) I added the `CreationalContextDestroyer` now

test-framework/arquillian/src/main/java/io/quarkus/arquillian/CreationalContextDestroyer.java
```diff
@@ -0,0 +1,27 @@
...
+
+import javax.enterprise.context.spi.CreationalContext;
+
+import org.jboss.arquillian.core.api.Instance;
+import org.jboss.arquillian.core.api.annotation.Inject;
+import org.jboss.arquillian.core.api.annotation.Observes;
+import org.jboss.arquillian.core.spi.EventContext;
+import org.jboss.arquillian.test.spi.event.suite.After;
+
+public class CreationalContextDestroyer {
```
manovotn: Quarkus already has one Arq. observer registered - `QuarkusBeforeAfterLifecycle`.
We could just add the logic there.

manovotn: But we might want to rename that class, now that I think of it.
So it's just a matter of preference I think :)

jmartisk: I don't know, it feels that they are doing quite different things, so the split into two classes makes sense to me :)

manovotn (approved):

mkouba (approved):

PR #3956, opened 2019-09-10
Quickstart ci test

stuartwdouglas:


cescoffier:
I'm surprised by the status of the build, as, at the moment, the quickstarts build does not pass:
https://dev.azure.com/quarkus-ci/quarkus/_build?definitionId=7&_a=summary

azure-pipelines.yml
```diff
@@ -297,3 +297,27 @@ stages:
...
+          - checkout: none
+          - task: DownloadPipelineArtifact@2
+            inputs:
+              artifact: BuiltMavenRepo
+              path: $(Pipeline.Workspace)/.m2/repository/
+          - script: git clone https://github.com/quarkusio/quarkus-quickstarts $(Agent.BuildDirectory)/s && cd $(Agent.BuildDirectory)/s && git checkout development
+          - task: Maven@3
+            displayName: 'Maven Install'
+            inputs:
+              goals: 'install'
```
cescoffier: should test the native build as it's where we got most regression.

azure-pipelines.yml
```diff
@@ -297,3 +297,27 @@ stages:
...
           name: hibernate_search_elasticsearch
+
+
+  - stage: test_quickstarts_stage
+    displayName: 'Quickstart Tests'
+    dependsOn: build_artifact_for_native_stage
+    jobs:
+      - job: Test_Quickstarts
+        displayName: 'Test Quickstarts'
+        timeoutInMinutes: 45
```
cescoffier: 45 is definitely not enough... closer to 380 minutes.

cescoffier (changes requested):

azure-pipelines.yml
gsmet: It won’t scale and will be a pain. 

Lately we had more compilation failures than anything else on the quickstarts. 

So the JVM build would be useful. Just not sure about the chicken and egg issue.

azure-pipelines.yml
gsmet: It’s enough in JVM mode :)

It doesn’t change the fact that we would need CI in the quickstarts themselves for native.

stuartwdouglas: If we were going to do the native builds we need to split it up like we do for our current native builds, and I don't think we have enough agents. 

I still don't like the chicken and egg problem it introduces though

cescoffier:
should we close or merge this one?

stuartwdouglas:
close for now

PR #3960, opened 2019-09-11
Remote is API so it shouldn't be in the runtime package but in the root

gsmet:
API annotations should be in the root package and not in the `runtime` package which is for the implementations.

Per the naming rules we decided a while ago. I deprecated the old
annotation but it should still work.

### DIFF
--- a/extensions/infinispan-client/runtime/src/main/java/io/quarkus/infinispan/client/runtime/InfinispanClientProducer.java
+++ b/extensions/infinispan-client/runtime/src/main/java/io/quarkus/infinispan/client/runtime/InfinispanClientProducer.java
@@ -245,15 +245,27 @@ public class InfinispanClientProducer {
         }
     }
 
-    @Remote
+    @SuppressWarnings("deprecation")
+    @io.quarkus.infinispan.client.Remote
+    @io.quarkus.infinispan.client.runtime.Remote
     @Produces
     public <K, V> RemoteCache<K, V> getRemoteCache(InjectionPoint injectionPoint, RemoteCacheManager cacheManager) {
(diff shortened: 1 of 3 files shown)


extensions/infinispan-client/runtime/src/main/java/io/quarkus/infinispan/client/runtime/InfinispanClientProducer.java
```diff
@@ -245,15 +245,27 @@ public void destroy() {
         }
     }
 
-    @Remote
+    @SuppressWarnings("deprecation")
+    @io.quarkus.infinispan.client.Remote
+    @io.quarkus.infinispan.client.runtime.Remote
```
dmlloyd: Why put both here?

integration-tests/infinispan-client/src/main/java/io/quarkus/it/infinispan/client/TestServlet.java
```diff
@@ -45,15 +45,16 @@
...
 import io.quarkus.runtime.StartupEvent;
 
 @Path("/test")
 public class TestServlet {
     private static final Log log = LogFactory.getLog(TestServlet.class);
 
+    @SuppressWarnings("deprecation")
     @Inject
-    @Remote("default")
+    @io.quarkus.infinispan.client.runtime.Remote("default")
```
dmlloyd: Is this qualification necessary?

gsmet:
Yeah, the idea is to not break existing applications. It was easy so I decided to not break things.

dmlloyd (approved):

PR #3961, opened 2019-09-11
Fix mailer extension runtime package

### LINKED ISSUE #3950: Improve consistency of the Mailer extension packages
Right now, we end up with the config reference doc being named `quarkus-mailer-impl.adoc` due to these inconsistencies.

We could work around it in the config reference doc generation but having consistency is better.

gsmet:
For consistency with the naming rules.

Fixes #3950


extensions/mailer/deployment/src/main/java/io/quarkus/mailer/deployment/MailerProcessor.java
```diff
@@ -11,7 +11,7 @@
 import io.quarkus.deployment.builditem.LaunchModeBuildItem;
 import io.quarkus.deployment.builditem.ShutdownContextBuildItem;
 import io.quarkus.deployment.builditem.substrate.ReflectiveClassBuildItem;
-import io.quarkus.mailer.impl.*;
+import io.quarkus.mailer.runtime.*;
```
machi1990: Does not look related but I think  we can avoid using start imports here :-)

machi1990 (approved):
LGTM - except for that one minor comment.

extensions/mailer/deployment/src/main/java/io/quarkus/mailer/deployment/MailerProcessor.java
gsmet: Yeah, it was preexisting but I pushed an update to fix it.

geoand (approved):

PR #3971, opened 2019-09-11
Reduce the number of FixedValueSupplier instances by creating them lazily

mthmulders:
I was toying around with a snapshot build of Quarkus today and noticed that the generated `ComponentsProvider` class creates a few "useless" (because: unused) `FixedValueSupplier` instances.

The `ComponentGenerator` always creates these instances, and puts them in a `Map` (at compile time). Some of these are looked up at compile time, and then it makes sense. But some of them aren't. In that case, the instance is created, but never used.

This patch addresses that. Instead of creating the `FixedValueSupplier` instance directly and putting it in the compile-time `Map`, it creates the instance _lazily_, only when it is needed. In the  `DefaultClassBeanTest`, this reduced the number of useless `FixedValueSupplier` instances from 6 to 0.

### DIFF
--- a/independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/ComponentsProviderGenerator.java
+++ b/independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/ComponentsProviderGenerator.java
@@ -107,7 +107,7 @@ public class ComponentsProviderGenerator extends AbstractGenerator {
             }
         }
 
-        Map<BeanInfo, ResultHandle> beanToResultSupplierHandle = new HashMap<>();
+        Map<BeanInfo, Supplier<ResultHandle>> beanToResultSupplierHandle = new HashMap<>();
         List<BeanInfo> processed = new ArrayList<>();
 
         // At this point we are going to fill the beanIdToBeanSupplier map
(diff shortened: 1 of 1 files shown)


mthmulders:
@mkouba  and @manovotn, since you have reviewed the original PR (#3559), could you please take a look?

independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/ComponentsProviderGenerator.java
```diff
@@ -282,9 +282,9 @@ private void addBean(MethodCreator getComponents, ResultHandle beanIdToBeanSuppl
                 beanInstance);
 
         // Create a Supplier that will return the bean instance at runtime.
-        ResultHandle beanInstanceSupplier = getComponents.newInstance(MethodDescriptors.FIXED_VALUE_SUPPLIER_CONSTRUCTOR,
-                beanInstance);
-        beanToResultSupplierHandle.put(bean, beanInstanceSupplier);
+
+        beanToResultSupplierHandle.put(bean,
+                () -> getComponents.newInstance(MethodDescriptors.FIXED_VALUE_SUPPLIER_CONSTRUCTOR, beanInstance));
```
stuartwdouglas: Shouldn't this cache the value in case .get() is called multiple times?

mthmulders: I don't think it *needs* to, but sure, it could be an optimisation. It will reduce the number of instances of `FixedValueSupplier` at runtime.

manovotn:
@mthmulders we're going to need a rebase here as another fix into was merged in the meantime.

After that I'll approve this, already reviewed it and tried it locally to see the generated code.

mthmulders:
Thanks @manovotn - rebase is done.

mkouba (approved):

manovotn (approved):

PR #3982, opened 2019-09-12
Change SR CP integration to make discovery based on TCCL.

### LINKED ISSUE #3908: MP CP integration doesn't allow for user defined custom contexts
I came across this while investigating execution of MP CP TCK on Quarkus.
What the TCK does is define custom contexts (`ThreadContextProvider`s) that are registered via service loader mechanism.
Here is a [link](https://github.com/eclipse/microprofile-context-propagation/blob/master/tck/src/main/java/org/eclipse/microprofile/context/tck/ThreadContextTest.java#L113) to those TCK tests.

However, our current SR CP integration does an early initialization of `SmallRyeContextManager` with set of discovered contexts that is determined before user application is inspected.
The integration happens [here](https://github.com/quarkusio/quarkus/blob/master/extensions/smallrye-context-propagation/deployment/src/main/java/io/quarkus/smallrye/context/deployment/SmallRyeContextPropagationProcessor.java#L41-L49) and is executed as `STATIC_INIT` which is too early, at least for TCK case.
I am not sure how this would affect actual Quarkus app where users would attempt to do similar thing - I suspect it would be the same.

We are going to need some way to make a second round of `ThreadContextProvider` discovery, or perhaps feed that information directly into `SmallRyeContextManager`?

Ccing @FroMage

manovotn:
Fixes #3908 

Assuming no tests crash here (tried non-native only locally), this fixes the problem of context provider discovery for MP CP TCK.

### DIFF
--- a/extensions/smallrye-context-propagation/deployment/src/main/java/io/quarkus/smallrye/context/deployment/SmallRyeContextPropagationProcessor.java
+++ b/extensions/smallrye-context-propagation/deployment/src/main/java/io/quarkus/smallrye/context/deployment/SmallRyeContextPropagationProcessor.java
@@ -38,7 +38,7 @@ class SmallRyeContextPropagationProcessor {
             throws ClassNotFoundException, IOException {
         List<ThreadContextProvider> discoveredProviders = new ArrayList<>();
         List<ContextManagerExtension> discoveredExtensions = new ArrayList<>();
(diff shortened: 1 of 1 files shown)


FroMage (approved):
LGTM but I still don't know why this makes any difference.

PR #3984, opened 2019-09-12
WIP: docs: docker only getting started guide

maxandersen:
Don't merge - just experiment.

I added a getting started guide that ONLY requires docker installed.
And probably only works on linux/mac systems as it uses alias to trick the system.

Also using the default image builder that can do java, maven and even native builds
but it is a whopping ~500 MB...but it shows the idea. Could probably be reduced.

Also would like to see if I could get it to work with pure podman but didn't find
good info on how to use that on mac.

Opening to get feedback if you think this idea has legs ?

### DIFF
--- a/docs/src/main/asciidoc/getting-started-guide-only-docker.adoc
+++ b/docs/src/main/asciidoc/getting-started-guide-only-docker.adoc
@@ -1,0 +1,447 @@
+////
+This guide is maintained in the main Quarkus repository
+and pull requests should be submitted there:
+https://github.com/quarkusio/quarkus/tree/master/docs/src/main/asciidoc
+////
+= Quarkus - Creating Your First Application
+
+include::./attributes.adoc[]
+
+:toc: macro
+:toclevels: 4
+:doctype: book
+:icons: font
+:docinfo1:
+
+:numbered:
+:sectnums:
+:sectnumlevels: 4
+
+
+Learn how to create a Hello World Quarkus app.
+This guide covers:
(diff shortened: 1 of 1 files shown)


gsmet (commented):
Can't say I'm excited about this but I'll let others decide :). Added one small comment.

cescoffier:
@maxandersen what do we do about this one?

maxandersen:
I'm biased :) I think it would be nice to have this documented where you can try out quarkus by only having docker installed - nothing else. Just not sure adding another getting started guide to the mix will get noticed/used...

gsmet:
I'm in favor of closing this one and see if our users complain about it.

Frankly, I think our situation is good enough and our requirements are pretty much OK for a Java developper.

geoand:
> I'm in favor of closing this one and see if our users complain about it.
> 
> Frankly, I think our situation is good enough and our requirements are pretty much OK for a Java developper.

+1

One reason for me +1 is that in the lab we did at Devoxx with @emmanuelbernard and @dandreadis most of the problems we had were docker related.
Had we never mentioned docker as an alternative for users, we would have had better results

maxandersen:
> Frankly, I think our situation is good enough and our requirements are pretty much OK for a Java developper.

this is to actually get those who are NON-active java users (like gophers). 

and major of users will have to use docker to build the native images anyway (windows is 85%++)

but im fine closing it just because its hard to find a place to put it ;)

PR #3989, opened 2019-09-12
Limit the number of verticle HTTP event loops based on availble memory

stuartwdouglas:


extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/VertxHttpRecorder.java
```diff
@@ -233,6 +233,19 @@ public void handle(AsyncResult<Void> event) {
                 "Listening on: http://%s:%s", httpServerOptions.getHost(), httpServerOptions.getPort()));
     }
 
+    private static int calculateDefaultIOThreads() {
+        //we only allow one event loop per 4mb of ram at the most
+        //its hard to say what this number should be, but it is also obvious
+        //that for constrained environments we don't want a lot of event loops
+        //lets start with 4mb and adjust as needed
+        int recommended = Runtime.getRuntime().availableProcessors() * 2;
```
dmlloyd: This should probably use `org.wildfly.common.cpu.ProcessorInfo#availableProcessors` at least for now.

gsmet: @stuartwdouglas do you plan to address this comment?

extensions/vertx-core/runtime/src/main/java/io/quarkus/vertx/core/runtime/VertxCoreRecorder.java
```diff
@@ -12,6 +12,8 @@
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
+import org.wildfly.common.cpu.ProcessorInfo;
```
cescoffier: Vert.x has something very similar. As this one does not add a dependency, we can keep this one. If one day the dependency is removed, we have a fallback.

extensions/vertx-core/runtime/src/main/java/io/quarkus/vertx/core/runtime/VertxCoreRecorder.java
```diff
@@ -149,6 +153,19 @@ private static VertxOptions convertToVertxOptions(VertxConfiguration conf) {
         return options;
     }
 
+    private static int calculateDefaultIOThreads() {
```
cescoffier: We also need to reduce the number of verticle deployed.

Event loops are created lazily (but it's always good to cap). when deploying a verticle, an event loop is assigned to it, which trigger the creation if the max is not reached. If you have more verticle than event loops, some verticle would share the event loop.

cescoffier (dismissed):
Asking about the number of verticle as it's related to the number of event loop.

gsmet (approved):
I checked with Stuart and @cescoffier had a look at this PR and said it was fine.

I'm approving and merging it so that we can have it in 0.23.0.

PR #4004, opened 2019-09-13
Fix NPE when handling circular dependencies

### LINKED ISSUE #3990: NPE when handling circular dependencies
If there is a normal scoped bean that declares a normal scoped producer and the bean depends on the producer method a NPE is thrown. See for example [MPConfigBean from the MP CP TCK](https://github.com/eclipse/microprofile-context-propagation/blob/master/tck/src/main/java/org/eclipse/microprofile/context/tck/MPConfigBean.java)

The producer bean can be added before the declaring bean and so a null `ResultHandle` is passed to Gizmo: https://github.com/quarkusio/quarkus/blob/master/independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/ComponentsProviderGenerator.java#L248.

mkouba:
- resolves #3990


mkouba:
@gsmet This is probably another candidate for backport to 0.22.0.

independent-projects/arc/runtime/src/main/java/io/quarkus/arc/MapBeanSupplier.java
```diff
@@ -8,17 +8,18 @@
  *
  * @author Maarten Mulders
  */
-public class MapBeanSupplier implements Supplier {
-    private final Map map;
+public class MapBeanSupplier implements Supplier<Object> {
```
mthmulders: Why not make it a `Supplier<Supplier<InjectableBean<?>>`, so the type parameter corresponds with the `beanIdToBeanSupplier` map?

mkouba: Good point. It has no importance but I can change it... ;-)

independent-projects/arc/runtime/src/main/java/io/quarkus/arc/FixedValueSupplier.java
```diff
@@ -7,7 +7,8 @@
  *
  * @author Maarten Mulders
  */
-public class FixedValueSupplier implements Supplier {
+public class FixedValueSupplier implements Supplier<Object> {
```
mthmulders: Maybe we should give this `Supplier` a more specific type parameter, too. Although it doesn't really add much value at runtime ;-)

mkouba: I'd keep this universal.

independent-projects/arc/runtime/src/main/java/io/quarkus/arc/MapBeanSupplier.java
mkouba: In fact, we could change it to `Map<String,T>` so that it could be theoretically reused.

mthmulders: Even better :-)

independent-projects/arc/runtime/src/main/java/io/quarkus/arc/FixedValueSupplier.java
mthmulders: Then maybe make this one a `FixedValueSupplier<T>`, just like the `MapBeanSupplier`. Still universal, but cleaner when it comes to using generics, I'd say?

manovotn (approved):
Went over some bits with Martin on Zulip and I think it looks good.

stuartwdouglas (approved):

PR #4012, opened 2019-09-13
Let users select the parsers in TikaConfiguration

sberyozkin:
This PR introduces a way for the users to select specific Tika parsers in the configuration, using a new `quarkus.tika.parsers` property.

At the moment the only way to select the specific parsers that will be loaded is to list them in `tika-config.xml` - however it does not affect the native image size which, by default, in a simple demo with resteasy and tika, is `96MB`.
Using the new property saves, when only PDF is required, `42MB`.

I've spent a lot of time trying to ensure that the same is achievable for `tika-config.xml` but if I load `TikaConfig` in the recorder then I can not access it at the build step point where the parser providers are added for the introspection and vice versa, if I load `TikaConfig` in the processor then the loaded instance can not be passed to the recorder.  May be it is achievable but after a while I've decided that trying to make a `tika-config.xml` way of configuring this extension a 1st class citizen is wrong anyway, and using the new property is the way to go as it paves the way for configuring the parsers directly in the application properties which is preferred in Quarkus and as have been advised by few colleagues already.

For example:
```
quarkus.tika.parsers=pdf
# This is not possible yet because TikaConfig can't load the parameters without their type specified and it is only possible at the moment with tika-config.xml.
I've opened a new Tika issue and once it is fixed the following will work 
quarkus.tika.parsers.pdf.property1=value
```

### DIFF
--- a/extensions/tika/runtime/src/main/java/io/quarkus/tika/runtime/TikaConfiguration.java
+++ b/extensions/tika/runtime/src/main/java/io/quarkus/tika/runtime/TikaConfiguration.java
@@ -12,10 +12,36 @@ import io.quarkus.runtime.annotations.ConfigRoot;
 @ConfigRoot(phase = ConfigPhase.BUILD_AND_RUN_TIME_FIXED)
 public class TikaConfiguration {
     /**
-     * The path to the tika-config.xml
+     * The resource path within the application artifact to the {@code tika-config.xml} file.
      */
     @ConfigItem
     public Optional<String> tikaConfigPath;
+
+    /**
+     * Comma separated list of the parsers which must be supported.
+     * <p>
+     * Most of the document formats recognized by Apache Tika are supported by default but it affects
+     * the application memory and native executable sizes. One can list only the required parsers in
+     * {@code tika-config.xml} to minimize a number of parsers loaded into the memory, but using this
+     * property is recommended to achieve both optimizations.
+     * <p>
+     * Either the abbreviated or full parser class names can be used.
+     * At the moment only PDF parser can be listed using a reserved 'pdf' abbreviation.
+     * Custom class name abbreviations have to be used for all other parsers.
+     * For example:
+     * 
+     * <pre>
+     * // Only PDF parser is required:
+     * tika-parsers = pdf
+     * // Only PDF and Java class parsers are required:
+     * tika-parsers = pdf,classparser
+     * classparser = org.apache.tika.parser.asm.ClassParser
+     * </pre>
+     * 
+     * This property will have no effect if the `tikaConfigPath' property has been set.
+     */
+    @ConfigItem
+    public Optional<String> parsers;
 
     /**
      * Controls how the content of the embedded documents is parsed.
--- a/docs/src/main/asciidoc/tika-guide.adoc
+++ b/docs/src/main/asciidoc/tika-guide.adoc
@@ -16,7 +16,8 @@ https://tika.apache.org/[Apache Tika] is a content analysis toolkit which is use
 [cols="<m,<m,<2",options="header"]
 |===
 |Property Name|Default|Description
-|quarkus.tika.tika-config-path||Path to the https://tika.apache.org/1.22/configuring.html[Tika configuration] resource. Typically a file named `tika-config.xml` is added to the root of the application resources path. The default configuration will be used if no configuration resource is available. 
+|quarkus.tika.tika-config-path||The resource path within the application artifact to the https://tika.apache.org/1.22/configuring.html[Tika configuration] resource. Typically a file named `tika-config.xml` is added to the root of the application resources path. The default configuration will be used if no configuration resource is available.
+|quarkus.tika.parsers||Comma-separated list of the abbreviated or full parser class names which have to be loaded by the extension. A `pdf` abbreviation can be used to refer to the PDF parser and custom abbreviations must be used for all other parsers. Note that this property is mutually exclusive with the `tika-config-path` property. 
 |quarkus.tika.append-embedded-content|true|The document may have other embedded documents, for example, an Excel document may include a PDF content. If such an embedded content is available then, by default, it will be appended to the content of the master document, thus, in this example, the text extracted from PDF file will be appended to the text extracted from the Excel file. This property has to be set to `false` if one needs to access the content of the master and each of the embedded documents individually.  
 |===
 
--- a/extensions/tika/deployment/pom.xml
+++ b/extensions/tika/deployment/pom.xml
@@ -30,11 +30,6 @@
             <artifactId>quarkus-junit5-internal</artifactId>
             <scope>test</scope>
         </dependency>
-        <dependency>
-            <groupId>io.rest-assured</groupId>
-            <artifactId>rest-assured</artifactId>
-            <scope>test</scope>
-        </dependency>
     </dependencies>
 
     <build>
--- a/extensions/tika/deployment/src/main/java/io/quarkus/tika/deployment/TikaProcessor.java
+++ b/extensions/tika/deployment/src/main/java/io/quarkus/tika/deployment/TikaProcessor.java
@@ -3,6 +3,9 @@ package io.quarkus.tika.deployment;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
+import java.util.Map;
+import java.util.NoSuchElementException;
+import java.util.Optional;
 import java.util.Set;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
@@ -10,6 +13,7 @@ import java.util.stream.Collectors;
 import org.apache.tika.detect.Detector;
 import org.apache.tika.detect.EncodingDetector;
 import org.apache.tika.parser.Parser;
+import org.eclipse.microprofile.config.ConfigProvider;
 
 import io.quarkus.arc.deployment.BeanContainerBuildItem;
 import io.quarkus.deployment.annotations.BuildProducer;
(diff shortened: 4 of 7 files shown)


sberyozkin:
In fact it is not even worth trying to do something like 
```
quarkus.tika.parsers.pdf.property1=value
```
and expecting the `property1` match precisely the underlying class setter property, the plan has always been to support the most useful properties at the Quarkus level only to support the typical cases, and if some users need some sophisticated parser configuration then they'll need to do it `tika-config.xml` (going this path will not give the same benefits as far as the native image size is concerned but in the longer term the tika-parser modularization will help as well)

extensions/tika/runtime/src/main/java/io/quarkus/tika/runtime/TikaConfiguration.java
```diff
@@ -12,11 +12,36 @@
...
 public class TikaConfiguration {
     /**
-     * The path to the tika-config.xml
+     * Path to the tika-config.xml
      */
     @ConfigItem
     public Optional<String> tikaConfigPath;
 
+    /**
+     * Comma separated list of the parsers which must be supported.
```
dmlloyd: Paragraph tags are missing between paragraphs.

extensions/tika/runtime/src/main/java/io/quarkus/tika/runtime/TikaConfiguration.java
```diff
@@ -12,11 +12,36 @@
...
+     * 
+     * Most of the document formats recognized by Apache Tika are supported by default but it affects
+     * the application memory and native executable sizes. One can list only the required parsers in
+     * tika-config.xml to minimize a number of parsers loaded into the memory, but using this
```
dmlloyd: `{@code tika-config.xml}`

extensions/tika/runtime/src/main/java/io/quarkus/tika/runtime/TikaConfiguration.java
```diff
@@ -12,11 +12,36 @@
...
+     * Either the abbreviated or full parser class names can be used.
+     * At the moment only PDF parser can be listed using a reserved 'pdf' abbreviation.
+     * Using the custom abbreviations is recommended for all other parsers.
+     * 
+     * <pre>
+     * // Only PDF parser is required:
+     * tika-parsers = pdf
+     * // Only PDF and Java class parsers are required:
+     * tika-parsers = pdf,classparser
+     * classparser=org.apache.tika.parser.asm.ClassParser
```
dmlloyd: Are these examples correct?  It doesn't seem to match the properties in this class.

extensions/tika/runtime/src/main/java/io/quarkus/tika/runtime/TikaConfiguration.java
```diff
@@ -12,11 +12,36 @@
 @ConfigRoot(phase = ConfigPhase.BUILD_AND_RUN_TIME_FIXED)
 public class TikaConfiguration {
     /**
-     * The path to the tika-config.xml
+     * Path to the tika-config.xml
```
dmlloyd: This should be wrapped in a taglet like this: `{@code tika-config.xml}`.

extensions/tika/runtime/src/main/java/io/quarkus/tika/runtime/TikaConfiguration.java
```diff
@@ -12,11 +12,36 @@
...
+     * tika-parsers = pdf
+     * // Only PDF and Java class parsers are required:
+     * tika-parsers = pdf,classparser
+     * classparser=org.apache.tika.parser.asm.ClassParser
+     * </pre>
+     * 
+     * This property will have no effect if the `tikaConfigPath' property has been set.
+     */
+    @ConfigItem
+    public Optional<String> parsers;
```
dmlloyd: If this is a list, shouldn't it be a `List<String>`? Also, if more parsers are supported by default, shouldn't there be a default value listing the supported parsers?

If the set of possible parsers is fixed, then rather than giving class names, it would be better to make an `enum` with the exact available choices.

extensions/tika/runtime/src/main/java/io/quarkus/tika/runtime/TikaConfiguration.java
```diff
@@ -12,11 +12,36 @@
 @ConfigRoot(phase = ConfigPhase.BUILD_AND_RUN_TIME_FIXED)
 public class TikaConfiguration {
     /**
-     * The path to the tika-config.xml
+     * Path to the tika-config.xml
      */
     @ConfigItem
     public Optional<String> tikaConfigPath;
```
dmlloyd: I know this isn't actually part of the PR, but it's better for paths to be of type `Path`; this adds a bit of extra validation.

sberyozkin:
Hi @dmlloyd, thanks for the review, let me work on it, please see my comment about the enums

extensions/tika/runtime/src/main/java/io/quarkus/tika/runtime/TikaConfiguration.java
sberyozkin: @dmlloyd 
This property instructs the extension which parser service providers should be loaded/added for the reflection, so I'd not like the extension 'committing' to some default values because if the users may need for ex PDF, but only Microsoft Office formats. 

The other problem is that a `tika-parsers` dependency brings 68 parsers with it, so it is difficult to decide which parsers will be supported OOB at the dependency level until after a [modularization](https://issues.apache.org/jira/browse/TIKA-2943) task is complete.
My plan has been to enhance this list of abbreviations such as `pdf` based on the demand. We start from `pdf`. If users say, why do we need to type a class name for this parser, then we make a new abbreviation recognized OOB. 
Keeping the list of parsers open-ended and use String may work better for it, given that the custom abbreviations are supported, see my comment about it above where you spotted the docs typo :-).

extensions/tika/runtime/src/main/java/io/quarkus/tika/runtime/TikaConfiguration.java
sberyozkin: @dmlloyd Thanks for spotting it the `tika-parsers` typo, should be just `parsers`.
The the `classparser` - this is a custom abbreviation and I'm going to restrict the `parsers` property containing only the abbreviations, either the ones recognized by the extension such as `pdf` or the custom one such as `classparser` (which may become promoted and become recognizable OOB) and not allow listing the class names in the `parsers` property directly:
the reason for it is, as I've mentioned above, I'd like to let users configure all the parsers (i.e the properties of the individual parsers) at the `application.properties` level without having to resort to typing `tika-config.xml`, 
So with the `parsers` property containing only the abbreviations (custom or built in ones), one would do:
```
quarkus.tika.parsers.pdf.sort-by-position=true
quarkus.tika.parsers.classparser.prop1=true
```

gsmet: Shouldn't this be a `Map` properly declared in the config object?

sberyozkin: @gsmet I have a different PR in the works to support this parser specific configuration, but as I've mentioned above, it requires [this issue](https://issues.apache.org/jira/browse/TIKA-2944) be resolved first, as under the hood the tika-config.xml will be generated but `TikaConfig` is unable at the moment to parse the properties without the XML type attributes and I'd like avoid users have to type that in themselves.

All this parser specific config is possible at the tika-config.xml level, [for example](https://github.com/apache/tika/blob/master/tika-parsers/src/test/resources/org/apache/tika/parser/pdf/tika-config.xml#L22). But this PR is a step toward letting users not to to use that XML config...

For the moment, for that another PR, I'm not adding a Map, and work with the `Config` provider directly to find out what is available for a given parser. If we add `Map` in `TikaConfiguration` then there will both the new `parsers` and this Map property will will be both optional/ignored if `tika-config-path` is available, so I have a minor preference to have only `parsers` here and doc how the parser-specific property can be set alongside the `parsers`.
In that PR I start with a `pdf.sort-by-position` - even though [TIKA-2944](https://issues.apache.org/jira/browse/TIKA-2944), effectively removing that tika-config.xml from the tests, because this property seems to be one of the most practical for the simple PDF invoices and I saw it being recommended few times. As such I reckon we can start with it with a minor property-specific check while keeping the code as generic as possible. I'll ask you and David to review that PR :-)

extensions/tika/runtime/src/main/java/io/quarkus/tika/runtime/TikaConfiguration.java
sberyozkin: @dmlloyd For the moment it is expected to be on the class path only and the extension does validates this value and throws the exception if the resource does not exist. No file resources are supported. So having `Path` would move this validation to the custom converter (which would to override the implicit one shipped with Quarkus :-) ). For the moment I'm keeping it as String - but would have no problems if you will still prefer `Path`.

extensions/tika/runtime/src/main/java/io/quarkus/tika/runtime/TikaConfiguration.java
```diff
@@ -12,11 +13,37 @@
...
+     * Most of the document formats recognized by Apache Tika are supported by default but it affects
+     * the application memory and native executable sizes. One can list only the required parsers in
+     * {@code tika-config.xml} to minimize a number of parsers loaded into the memory, but using this
+     * property is recommended to achieve both optimizations.
+     * <p>
+     * Either the abbreviated or full parser class names can be used.
+     * At the moment only PDF parser can be listed using a reserved 'pdf' abbreviation.
+     * Custom class name abbreviations have to be used for all other parsers.
+     * For example:
+     *
```
sberyozkin: The formatter does not let me add the paragraph tags around the `pre` ones, it removes them, I guess they are not needed with the `pre` tags

extensions/tika/runtime/src/main/java/io/quarkus/tika/runtime/TikaConfiguration.java
dmlloyd: Ah okay.  Using `String` is fine then, but you'd better clarify that in the doc, something like this:

```java
The resource path within the application artifact to the {@code tika-config.xml} file.
```

sberyozkin:
Hi @dmlloyd @gsmet 
I've reverted for the moment to having `String` instead of `List<String>` and will get back to the List once it is supported, also improved the unit tests. 
Have a look please, I'll follow up with another PR I've mentioned earlier a little bit later after this PR makes it in

extensions/tika/runtime/src/main/java/io/quarkus/tika/runtime/TikaConfiguration.java
sberyozkin: @dmlloyd done, I'd never be able to come up myself with such a compact yet complete doc sentence :-)

extensions/tika/runtime/src/main/java/io/quarkus/tika/runtime/TikaConfiguration.java
dmlloyd: Sounds good.

dmlloyd (approved):

PR #4025, opened 2019-09-15
Ensure default constructor for RESTEasy Resources also works for Kotlin code

geoand:
This is needed because Kotlin adds `@kotlin.Metadata` annotation to the generated bytecode.

Based on [this](https://quarkusio.zulipchat.com/#narrow/stream/187030-users/topic/JAX-RS.2C.20Kotlin.20and.20constructor.20injection/near/175741440) conversation.

### DIFF
--- a/extensions/resteasy-server-common/deployment/src/main/java/io/quarkus/resteasy/server/common/deployment/ResteasyServerCommonProcessor.java
+++ b/extensions/resteasy-server-common/deployment/src/main/java/io/quarkus/resteasy/server/common/deployment/ResteasyServerCommonProcessor.java
@@ -406,6 +406,7 @@ public class ResteasyServerCommonProcessor {
 
         final Set<String> allowedAnnotationPrefixes = new HashSet<>(1 + additionalJaxRsResourceDefiningAnnotations.size());
         allowedAnnotationPrefixes.add(packageName(ResteasyDotNames.PATH));
+        allowedAnnotationPrefixes.add("kotlin"); // make sure the annotation that the Kotlin compiler adds don't interfere with creating a default constructor
         for (AdditionalJaxRsResourceDefiningAnnotationBuildItem additionalJaxRsResourceDefiningAnnotation : additionalJaxRsResourceDefiningAnnotations) {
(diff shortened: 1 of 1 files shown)


extensions/resteasy-server-common/deployment/src/main/java/io/quarkus/resteasy/server/common/deployment/ResteasyServerCommonProcessor.java
```diff
@@ -406,6 +406,7 @@ private static void generateDefaultConstructors(BuildProducer<BytecodeTransforme
 
         final Set<String> allowedAnnotationPrefixes = new HashSet<>(1 + additionalJaxRsResourceDefiningAnnotations.size());
         allowedAnnotationPrefixes.add(packageName(ResteasyDotNames.PATH));
+        allowedAnnotationPrefixes.add("kotlin"); // make sure the annotation that the Kotlin compiler adds don't interfere with creating a default constructor
```
machi1990: Do we need to do the same for `scala`?

machi1990 (commented):
This LGTM. Just one minor question

extensions/resteasy-server-common/deployment/src/main/java/io/quarkus/resteasy/server/common/deployment/ResteasyServerCommonProcessor.java
geoand: Good question! 

I have :zero: knowledge of Scala TBH so I really don't know what the Scala compiler produces. I would say that if you or anyone else wants to check it out, then you can follow up with a new PR if needed :)

renannprado:
@machi1990 @geoand can't it be part of `0.22.0` that's about to be released?

geoand:
I think it could. WDYT @gsmet?

gsmet (approved):

gsmet:
It will have to wait a bit. 0.22 is ready to be released and it's too late to include any new commits that I would have to test.

In any case, 0.23 is not far away.

geoand:
Thanks for explaining @gsmet!

PR #4030, opened 2019-09-16
Make core vert.x lazy

stuartwdouglas:


gsmet:
What’s the purpose of this one? Isn’t the instance needed as soon as we have a REST resource?

stuartwdouglas:
WebVertx is the one that is needed as soon as we have REST. If there is no reactive stuff in the app then this one will not be used.

gsmet (approved):

PR #4046, opened 2019-09-16
Support HttpServletRequest/Response in Spring exception handlers

### LINKED ISSUE #4042: @ExceptionHandler method does not accept the WebRequest / HttpRequest as a parameter
Quarkus 0.22.0

`@ExceptionHandler` only accept one parameter, but it should allow more params in the method param list.

https://github.com/hantsy/quarkus-sample/tree/master/spring-post-service

```java
@RestControllerAdvice
public class PostExceptionHandler {

    @ExceptionHandler(PostNotFoundException.class)
    public ResponseEntity notFound(PostNotFoundException ex/*, WebRequest req*/) {
        Map<String, String> errors = new HashMap<>();
        errors.put("entity", "POST");
        errors.put("id", "" + ex.getSlug());
        errors.put("code", "not_found");
        errors.put("message", ex.getMessage());

        return status(HttpStatus.NOT_FOUND).body(errors);
    }

}
```

geoand:
Fixes: #4042

### DIFF
--- a/extensions/spring-web/deployment/src/main/java/io/quarkus/spring/web/deployment/ControllerAdviceAbstractExceptionMapperGenerator.java
+++ b/extensions/spring-web/deployment/src/main/java/io/quarkus/spring/web/deployment/ControllerAdviceAbstractExceptionMapperGenerator.java
@@ -1,19 +1,28 @@
 package io.quarkus.spring.web.deployment;
 
 import java.lang.annotation.Annotation;
+import java.lang.reflect.Modifier;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
 
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import javax.ws.rs.core.Context;
 import javax.ws.rs.core.Response;
 
 import org.jboss.jandex.AnnotationInstance;
 import org.jboss.jandex.DotName;
-import org.jboss.jandex.IndexView;
 import org.jboss.jandex.MethodInfo;
 import org.jboss.jandex.Type;
 
 import io.quarkus.arc.Arc;
 import io.quarkus.arc.ArcContainer;
 import io.quarkus.arc.InstanceHandle;
+import io.quarkus.gizmo.ClassCreator;
 import io.quarkus.gizmo.ClassOutput;
+import io.quarkus.gizmo.FieldCreator;
+import io.quarkus.gizmo.FieldDescriptor;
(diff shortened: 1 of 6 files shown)


extensions/spring-web/deployment/src/main/java/io/quarkus/spring/web/deployment/ControllerAdviceAbstractExceptionMapperGenerator.java
```diff
@@ -24,23 +33,77 @@
...
+                if (parameterTypeToField.containsKey(parameterType)) {
+                    throw new IllegalArgumentException("Parameter type " + parameterTypes.get(notAllowedParameterIndex).name()
+                            + " is being used multiple times in method" + controllerAdviceMethod.name() + " of class"
+                            + controllerAdviceMethod.declaringClass().name());
+                }
+
+                // we need to generate a field that injects the HttpServletRequest into the class
+                FieldCreator httpRequestFieldCreator = cc.getFieldCreator("httpServletRequest", HttpServletRequest.class)
+                        .setModifiers(Modifier.PRIVATE);
+                httpRequestFieldCreator.addAnnotation(Context.class);
```
gsmet: It might be a stupid question but is the scope of the `ExceptionMapper` restricted enough so that we can inject context values in it? I have a vague recollection of discussion about that but I might be mistaken.

geoand: I am sure it's a very valid question butI am not sure I follow. What could the potential problem be?

gsmet: If the exception mapper is scoped singleton or application, I'm not sure it's safe to inject context values in its fields. Or maybe RESTEasy takes care of that.

geoand: I didn't encounter any problems TBH and looking around it seemed like people were using this solution when writing their own ExceptionMappers

geoand: I think we should be OK for the time being. We can always go deeper if users report errors. WDYT  @gsmet ?

gsmet (approved):

Issue #4047, opened 2019-09-16
Native image start timeout when testing vertx-web on Quarkus 0.22.0


ppalaga:
An integration test for `vertx-web` extension (see [this branch](https://github.com/ppalaga/quarkus/commits/190916-vertx-web-itest) )  fails with the following message:

```
org.junit.jupiter.api.extension.TestInstantiationException: TestInstanceFactory [io.quarkus.test.junit.QuarkusTestExtension] failed to instantiate test class [io.quarkus.it.vertx.web.VertxWebIT]: Unable to start native image in 60s
        at org.junit.jupiter.engine.descriptor.ClassBasedTestDescriptor.invokeTestInstanceFactory(ClassBasedTestDescriptor.java:298)
        at org.junit.jupiter.engine.descriptor.ClassBasedTestDescriptor.instantiateTestClass(ClassBasedTestDescriptor.java:273)
        at org.junit.jupiter.engine.descriptor.ClassTestDescriptor.instantiateTestClass(ClassTestDescriptor.java:70)
        at org.junit.jupiter.engine.descriptor.ClassBasedTestDescriptor.instantiateAndPostProcessTestInstance(ClassBasedTestDescriptor.java:256)
        at org.junit.jupiter.engine.descriptor.ClassBasedTestDescriptor.lambda$testInstancesProvider$2(ClassBasedTestDescriptor.java:250)
        at java.util.Optional.orElseGet(Optional.java:267)
        at org.junit.jupiter.engine.descriptor.ClassBasedTestDescriptor.lambda$testInstancesProvider$3(ClassBasedTestDescriptor.java:249)
        at org.junit.jupiter.engine.execution.TestInstancesProvider.getTestInstances(TestInstancesProvider.java:29)
        at org.junit.jupiter.engine.descriptor.TestMethodTestDescriptor.lambda$prepare$0(TestMethodTestDescriptor.java:106)
        at org.junit.platform.engine.support.hierarchical.ThrowableCollector.execute(ThrowableCollector.java:73)
        at org.junit.jupiter.engine.descriptor.TestMethodTestDescriptor.prepare(TestMethodTestDescriptor.java:105)
        at org.junit.jupiter.engine.descriptor.TestMethodTestDescriptor.prepare(TestMethodTestDescriptor.java:69)
        at org.junit.platform.engine.support.hierarchical.NodeTestTask.lambda$prepare$1(NodeTestTask.java:107)
[...]

PR #4061, opened 2019-09-17
Add a vertx-web integration test with minimal dependencies

ppalaga:


gsmet:
What's the rationale of this? I know @cescoffier had plans to add more Vert.x tests so we should coordinate with him.

ppalaga:
My motivation is to make sure that issues like https://github.com/quarkusio/quarkus/issues/4047 won't happen.

gsmet:
And the issue you have is solved in master? Because the test passes and I don't see any fix.

ppalaga:
Yes, #4047 does not occur in master.

stuartwdouglas:
Vert.x is now used for all HTTP, so every integration test makes sure it works correctly.

This is a -1 from me, native image tests are very resource intensive and I don't really know if this one adds much.

ppalaga:
The minimal dependency set was supposed to be the main added value of this test. But I can understand your concerns. Feel free to close. In the worst case we'll catch any issues in Camel Quarkus.

gsmet:
OK, let's close it then.

PR #4074, opened 2019-09-17
Init HexUtil at runtime

stuartwdouglas:
It allocates some largish arrays in static init


gsmet:
It might be a stupid question but how is it better to allocate things at startup?

I suppose doing this reduces the size of the image?

stuartwdouglas:
If the class is not actually used then the allocation may never happen

gsmet:
Ah, that makes sense :). I thought it was a core class used for everything. Let's merge then.

gsmet (approved):

PR #4075, opened 2019-09-17
Implement @ConfigProperties as a way to group related properties and inject with minimum boilerplate

geoand:
See discussion at: https://groups.google.com/forum/#!topic/quarkus-dev/YTRo-NCf-p0

### DIFF
--- a/core/deployment/src/main/java/io/quarkus/deployment/bean/JavaBeanUtil.java
+++ b/core/deployment/src/main/java/io/quarkus/deployment/bean/JavaBeanUtil.java
@@ -1,9 +1,12 @@
-package io.quarkus.panache.common.deployment;
+package io.quarkus.deployment.bean;
 
 public class JavaBeanUtil {
 
+    private static final String GET = "get";
+    private static final String IS = "is";
+
     public static String getGetterName(String name, String type) {
-        String prefix = type.equals("Z") ? "is" : "get";
+        String prefix = type.equals("Z") ? IS : GET;
         return prefix + capitalize(name);
     }
 
@@ -38,6 +41,21 @@ public class JavaBeanUtil {
             }
         } else {
             return name;
+        }
+    }
+
+    /**
+     * Returns the corresponding property name for a getter method name
+     * 
+     * @throws IllegalArgumentException if the method name does not follow the getter name convention
+     */
+    public static String getPropertyNameFromGetter(String methodName) {
+        if (methodName.startsWith(GET)) {
+            return decapitalize(methodName.substring(GET.length()));
+        } else if (methodName.startsWith(IS)) {
+            return decapitalize(methodName.substring(IS.length()));
+        } else {
+            throw new IllegalArgumentException(methodName + " is not a getter");
         }
     }
 }
--- a/core/runtime/src/main/java/io/quarkus/runtime/configuration/HashSetFactory.java
+++ b/core/runtime/src/main/java/io/quarkus/runtime/configuration/HashSetFactory.java
@@ -1,0 +1,38 @@
+package io.quarkus.runtime.configuration;
+
+import java.util.HashSet;
+import java.util.function.IntFunction;
+
+/**
+ * A helper class to produce a right-sized hash set.
+ */
+public final class HashSetFactory<T> implements IntFunction<HashSet<T>> {
+    private static final HashSetFactory<?> INSTANCE = new HashSetFactory<>();
+
+    private HashSetFactory() {
+    }
+
+    public HashSet<T> apply(final int value) {
+        return new HashSet<>(getInitialCapacityFromExpectedSize(value));
+    }
+
+    /**
+     * As the default loadFactor is of 0.75, we need to calculate the initial capacity from the expected size to avoid
+     * resizing the collection when we populate the collection with all the initial elements. We use a calculation
+     * similar to what is done in {@link java.util.HashMap#putAll(Map)}.
+     *
+     * @param expectedSize the expected size of the collection
+     * @return the initial capacity of the collection
+     */
+    private int getInitialCapacityFromExpectedSize(int expectedSize) {
+        if (expectedSize < 3) {
+            return expectedSize + 1;
+        }
+        return (int) ((float) expectedSize / 0.75f + 1.0f);
+    }
+
+    @SuppressWarnings("unchecked")
+    public static <T> HashSetFactory<T> getInstance() {
+        return (HashSetFactory<T>) INSTANCE;
+    }
+}
--- a/docs/src/main/asciidoc/application-configuration-guide.adoc
+++ b/docs/src/main/asciidoc/application-configuration-guide.adoc
@@ -158,6 +158,144 @@ You can add the `greeting.suffix`, remove the other properties, change the value
 As usual, the application can be packaged using `./mvnw clean package` and executed using the `-runner.jar` file.
 You can also generate the native executable with `./mvnw clean package -Pnative`.
 
+== Using @ConfigProperties
+
+As an alternative to injecting multiple related configuration values in the way that was shown in the previous example,
+users can also use the `@io.quarkus.arc.config.ConfigProperties` annotation to group these properties together.
+
+For the greeting properties above, a `GreetingConfiguration` class could be created like so:
+
+[source,java]
+----
+package org.acme.config;
+
+import io.quarkus.arc.config.ConfigProperties;
+import java.util.Optional;
+
+@ConfigProperties(prefix = "greeting") <1>
+public class GreetingConfiguration {
+
+    private String message;
+    private String suffix = "!"; <2>
+    private Optional<String> name;
+
+    public String getMessage() {
+        return message;
+    }
+
+    public void setMessage(String message) {
+        this.message = message;
+    }
+
+    public String getSuffix() {
+        return suffix;
+    }
+
+    public void setSuffix(String suffix) {
+        this.suffix = suffix;
+    }
+
+    public Optional<String> getName() {
+        return name;
+    }
+
+    public void setName(Optional<String> name) {
+        this.name = name;
+    }
+}
+----
+<1> `prefix` is optional. If not set then the prefix to be used will be determined by the class name. In this case it would still be `greeting` (since the `Properties` suffix is removed). If the class were named `GreetingExtraProperties` then the resulting default prefix would be `greeting-extra`.
+<2> `!` will be the default value if `greeting.suffix` is not set
+
+This class could then be injected into the `GreetingResource` using the familiar CDI `@Inject` annotation like so:
+
+[source,java]
+----
+@Inject
+GreetingConfiguation greetingConfigation;
+----
+
+Another alternative style provided by Quarkus is to create `GreetingConfiguration` as an interface like so:
+
+[source,java]
+----
+package org.acme.config;
+
+import io.quarkus.arc.config.ConfigProperties;
+import org.eclipse.microprofile.config.inject.ConfigProperty;
+import java.util.Optional;
+
+@ConfigProperties(prefix = "greeting")
+public interface GreetingConfiguration {
+
+    @ConfigProperty(name = "message") <1>
+    String message();
+
+    @ConfigProperty(defaultValue = "!")
+    String getSuffix(); <2>
+
+    Optional<String> getName(); <3>
+}
+----
+<1> The `@ConfigProperty` annotation is needed because the name of the configuration property that the method corresponds to doesn't follow the getter method naming conventions
+<2> In this case since `name` was not set, the corresponding property will be `greeting.suffix`.
+<3> It is unnecessary to specify the `@ConfigProperty` annotation because the method name follows the getter method naming conventions (`greeting.name` being the corresponding property) and no default value is needed.
+
+=== Additional notes on @ConfigProperties
+
+When using a regular class annotated with `@ConfigProperties` the class doesn't necessarily have to declare getters and setters.
+Having simple public non-final fields is valid as well.
+
+Furthermore the configuration classes support nested object configuration. Suppose there was a need to have an extra layer
+of greeting configuration named `hidden` that would contain a few fields. This could be achieved like so:
+
+[source,java]
+----
+@ConfigProperties(prefix = "greeting")
+public class GreetingConfiguration {
+
+    public String message;
+    public String suffix = "!";
+    public Optional<String> name;
+    public HiddenConfig hidden; <1>
+
+    public static class HiddenConfig {
+        public Integer prizeAmount;
+        public List<String> recipients;
+    }
+}
+----
+<1> The name of the field (not the class name) will determine the name of the properties that are bound to the object.
+
+Setting the properties would occur in the normal manner, for example in `application.properties` one could have:
+
+[source,properties]
+----
+greeting.message = hello
+greeting.name = quarkus
+greeting.hidden.prizeAmount=10
+greeting.hidden.recipients=Jane,John
+----
+
+Furthermore, classes annotated with `@ConfigProperties` can be annotated with Bean Validation annotations similar to the following example:
+
+[source,java]
+----
+@ConfigProperties(prefix = "greeting")
+public class GreetingConfiguration {
+
+    @Size(min = 20)
+    public String message;
+    public String suffix = "!";
+
+}
+----
+
+If the validation fails with the given configuration, the application will fail to start and indicate the corresponding validation errors in the log.
+
+In the case of an interface being annotated with `@ConfigProperties`, the interface is allowed to extend other interfaces and methods from
+the entire interface hierarchy are used to bind properties.
+
 == Configuring Quarkus
 
 Quarkus itself is configured via the same mechanism as your application. Quarkus reserves the `quarkus.` namespace
--- a/extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/configproperties/ClassConfigPropertiesUtil.java
+++ b/extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/configproperties/ClassConfigPropertiesUtil.java
@@ -1,0 +1,372 @@
+package io.quarkus.arc.deployment.configproperties;
+
+import static io.quarkus.arc.deployment.configproperties.ConfigPropertiesUtil.ReadOptionalResponse;
+import static io.quarkus.arc.deployment.configproperties.ConfigPropertiesUtil.createReadMandatoryValueAndConvertIfNeeded;
+import static io.quarkus.arc.deployment.configproperties.ConfigPropertiesUtil.createReadOptionalValueAndConvertIfNeeded;
+import static io.quarkus.arc.deployment.configproperties.ConfigPropertiesUtil.determineSingleGenericType;
+
+import java.lang.reflect.Modifier;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.NoSuchElementException;
+import java.util.Optional;
+import java.util.Set;
+
+import javax.enterprise.context.Dependent;
+import javax.enterprise.event.Observes;
+import javax.enterprise.inject.Produces;
+import javax.inject.Inject;
+
+import org.eclipse.microprofile.config.Config;
+import org.jboss.jandex.ClassInfo;
+import org.jboss.jandex.DotName;
+import org.jboss.jandex.FieldInfo;
+import org.jboss.jandex.IndexView;
+import org.jboss.jandex.MethodInfo;
+import org.jboss.jandex.Type;
+
+import io.quarkus.deployment.bean.JavaBeanUtil;
+import io.quarkus.deployment.util.HashUtil;
+import io.quarkus.gizmo.BranchResult;
+import io.quarkus.gizmo.BytecodeCreator;
+import io.quarkus.gizmo.ClassCreator;
+import io.quarkus.gizmo.ClassOutput;
+import io.quarkus.gizmo.FieldCreator;
+import io.quarkus.gizmo.FieldDescriptor;
+import io.quarkus.gizmo.MethodCreator;
+import io.quarkus.gizmo.MethodDescriptor;
+import io.quarkus.gizmo.ResultHandle;
+import io.quarkus.runtime.StartupEvent;
+
+final class ClassConfigPropertiesUtil {
+
+    private static final String VALIDATOR_CLASS = "javax.validation.Validator";
+    private static final String HIBERNATE_VALIDATOR_IMPL_CLASS = "org.hibernate.validator.HibernateValidator";
+    private static final String CONSTRAINT_VIOLATION_EXCEPTION_CLASS = "javax.validation.ConstraintViolationException";
+
+    private ClassConfigPropertiesUtil() {
+    }
+
+    /**
+     * Generates a class like the following:
+     *
+     * <pre>
+     * &#64;ApplicationScoped
+     * public class EnsureValidation {
+     *
+     *     &#64;Inject
+     *     MyConfig myConfig;
+     *
+     *     &#64;Inject
+     *     OtherProperties other;
+     *
+     *     public void onStartup(@Observes StartupEvent ev) {
+     *         myConfig.toString();
+     *         other.toString();
+     *     }
+     * }
+     * </pre>
+     *
+     * This class is useful in order to ensure that validation errors will prevent application startup
+     */
+    static void generateStartupObserverThatInjectsConfigClass(ClassOutput classOutput, Set<DotName> configClasses) {
+        try (ClassCreator classCreator = ClassCreator.builder().classOutput(classOutput)
+                .className(ConfigPropertiesUtil.PACKAGE_TO_PLACE_GENERATED_CLASSES + ".ConfigPropertiesObserver")
+                .build()) {
+            classCreator.addAnnotation(Dependent.class);
+
+            Map<DotName, FieldDescriptor> configClassToFieldDescriptor = new HashMap<>(configClasses.size());
+
+            for (DotName configClass : configClasses) {
+                String configClassStr = configClass.toString();
+                FieldCreator fieldCreator = classCreator
+                        .getFieldCreator(
+                                configClass.isInner() ? configClass.local()
+                                        : configClass.withoutPackagePrefix() + "_" + HashUtil.sha1(configClassStr),
+                                configClassStr)
+                        .setModifiers(Modifier.PUBLIC); // done to prevent warning during the build
+                fieldCreator.addAnnotation(Inject.class);
+
+                configClassToFieldDescriptor.put(configClass, fieldCreator.getFieldDescriptor());
+            }
+
+            try (MethodCreator methodCreator = classCreator.getMethodCreator("onStartup", void.class, StartupEvent.class)) {
+                methodCreator.getParameterAnnotations(0).addAnnotation(Observes.class);
+                for (DotName configClass : configClasses) {
+                    /*
+                     * We call toString on the bean which ensure that bean is created thus ensuring
+                     * validation is actually performed
+                     */
+                    ResultHandle field = methodCreator.readInstanceField(configClassToFieldDescriptor.get(configClass),
+                            methodCreator.getThis());
+                    methodCreator.invokeVirtualMethod(
+                            MethodDescriptor.ofMethod(configClass.toString(), "toString", String.class.getName()),
+                            field);
+                }
+                methodCreator.returnValue(null); // the method doesn't need to do anything
+            }
+        }
+    }
+
+    /**
+     * @return true if the configuration class needs validation
+     */
+    static boolean addProducerMethodForClassConfigProperties(ClassInfo configPropertiesClassInfo,
+            ClassCreator producerClassCreator, String prefixStr, IndexView applicationIndex) {
+
+        if (!DotNames.OBJECT.equals(configPropertiesClassInfo.superName())) {
+            throw new IllegalArgumentException(
+                    "Classes annotated with @" + DotNames.CONFIG_PROPERTIES
+                            + " cannot extend other classes. Offending class is "
+                            + configPropertiesClassInfo);
+        }
+
+        if (!configPropertiesClassInfo.hasNoArgsConstructor()) {
+            throw new IllegalArgumentException(
+                    "Class " + configPropertiesClassInfo + " which is annotated with " + DotNames.CONFIG_PROPERTIES
+                            + " must contain a no-arg constructor");
+        }
+
+        if (!Modifier.isPublic(configPropertiesClassInfo.flags())) {
+            throw new IllegalArgumentException(
+                    "Class " + configPropertiesClassInfo + " which is annotated with " + DotNames.CONFIG_PROPERTIES
+                            + " must be public");
+        }
+
+        String configObjectClassStr = configPropertiesClassInfo.name().toString();
+
+        boolean needsValidation = needsValidation();
+        String[] produceMethodParameterTypes = new String[needsValidation ? 2 : 1];
+        produceMethodParameterTypes[0] = Config.class.getName();
+        if (needsValidation) {
+            produceMethodParameterTypes[1] = VALIDATOR_CLASS;
+        }
+
+        /*
+         * Add a method like this:
+         *
+         * @Produces
+         * public SomeClass produceSomeClass(Config config) {
+         *
+         * }
+         *
+         * or
+         *
+         * @Produces
+         * public SomeClass produceSomeClass(Config config, Validator validator) {
+         *
+         * }
+         */
+
+        try (MethodCreator methodCreator = producerClassCreator.getMethodCreator(
+                "produce" + configPropertiesClassInfo.name().withoutPackagePrefix(),
+                configObjectClassStr, produceMethodParameterTypes)) {
+            methodCreator.addAnnotation(Produces.class);
+
+            ResultHandle configObject = populateConfigObject(configPropertiesClassInfo,
+                    prefixStr,
+                    methodCreator, applicationIndex);
+
+            if (needsValidation) {
+                createValidationCodePath(methodCreator, configObject, prefixStr);
+            } else {
+                methodCreator.returnValue(configObject);
+            }
+        }
+
+        return needsValidation;
+    }
+
+    private static boolean needsValidation() {
+        /*
+         * Hibernate Validator has minimum overhead if the class is unconstrained,
+         * so we'll just pass all config classes to it if it's present
+         */
+        return isHibernateValidatorInClasspath();
+    }
+
+    private static boolean isHibernateValidatorInClasspath() {
+        try {
+            Class.forName(HIBERNATE_VALIDATOR_IMPL_CLASS);
+            return true;
+        } catch (ClassNotFoundException e) {
+            return false;
+        }
+    }
+
+    private static ResultHandle populateConfigObject(ClassInfo configClassInfo,
+            String prefixStr,
+            MethodCreator methodCreator, IndexView applicationIndex) {
+        String configObjectClassStr = configClassInfo.name().toString();
+        ResultHandle configObject = methodCreator.newInstance(MethodDescriptor.ofConstructor(configObjectClassStr));
+
+        // For each field of the class try to pull it out of MP Config and call the corresponding setter
+        List<FieldInfo> fields = configClassInfo.fields();
+        for (FieldInfo field : fields) {
+            boolean useFieldAccess = false;
+
+            String setterName = JavaBeanUtil.getSetterName(field.name());
+            Type fieldType = field.type();
+            MethodInfo setter = configClassInfo.method(setterName, fieldType);
+            if (setter == null) {
+                if (!Modifier.isPublic(field.flags()) || Modifier.isFinal(field.flags())) {
+                    throw new IllegalArgumentException(
+                            "Configuration properties class " + configClassInfo + " does not have a setter for field "
+                                    + field + " nor is the field a public non-final field");
+                }
+                useFieldAccess = true;
+            }
+            if (!useFieldAccess && !Modifier.isPublic(setter.flags())) {
+                throw new IllegalArgumentException(
+                        "Setter " + setterName + " of class " + configClassInfo + " must be public");
+            }
+
+            /*
+             * If the object is part of the application we are dealing with a nested object
+             * What we do is simply recursively build it up based by adding the field name to the config name prefix
+             */
+            DotName fieldTypeDotName = fieldType.name();
+            String fieldTypeStr = fieldTypeDotName.toString();
+            ClassInfo fieldTypeClassInfo = applicationIndex.getClassByName(fieldType.name());
+            if (fieldTypeClassInfo != null) {
+                if (!fieldTypeClassInfo.hasNoArgsConstructor()) {
+                    throw new IllegalArgumentException(
+                            "Nested configuration class " + fieldTypeClassInfo + " must contain a no-args constructor ");
+                }
+
+                if (!Modifier.isPublic(fieldTypeClassInfo.flags())) {
+                    throw new IllegalArgumentException(
+                            "Nested configuration class " + fieldTypeClassInfo + " must be public ");
+                }
+
+                ResultHandle nestedConfigObject = populateConfigObject(fieldTypeClassInfo,
+                        prefixStr + "." + field.name(),
+                        methodCreator, applicationIndex);
+                createWriteValue(methodCreator, configObject, field, setter, useFieldAccess, nestedConfigObject);
+
+            } else {
+                String fullConfigName = prefixStr + "." + field.name();
+                ResultHandle config = methodCreator.getMethodParam(0);
+                if (DotNames.OPTIONAL.equals(fieldTypeDotName)) {
+                    Type genericType = determineSingleGenericType(field.type(),
+                            field.declaringClass().name());
+
+                    // config.getOptionalValue
+                    ResultHandle setterValue = methodCreator.invokeInterfaceMethod(
+                            MethodDescriptor.ofMethod(Config.class, "getOptionalValue", Optional.class, String.class,
+                                    Class.class),
+                            config, methodCreator.load(fullConfigName), methodCreator.loadClass(genericType.name().toString()));
+                    createWriteValue(methodCreator, configObject, field, setter, useFieldAccess, setterValue);
+                } else {
+                    /*
+                     * We want to support cases where the Config class defines a default value for fields
+                     * by simply specifying the default value in its constructor
+                     * For such cases the strategy we follow is that when a requested property does not exist
+                     * we check the value from the corresponding getter (or read the field value if possible)
+                     * and if the value is not null we don't fail
+                     */
+                    if (shouldCheckForDefaultValue(configClassInfo, field)) {
+                        String getterName = JavaBeanUtil.getGetterName(field.name(), fieldTypeDotName.toString());
+
+                        ReadOptionalResponse readOptionalResponse = createReadOptionalValueAndConvertIfNeeded(fullConfigName,
+                                fieldType, field.declaringClass().name(), methodCreator, config);
+
+                        // call the setter if the optional contained data
+                        createWriteValue(readOptionalResponse.getIsPresentTrue(), configObject, field, setter, useFieldAccess,
+                                readOptionalResponse.getValue());
+
+                        // if optional did not contain data, check the getter and see if there is a value
+                        BytecodeCreator isPresentFalse = readOptionalResponse.getIsPresentFalse();
+                        ResultHandle defaultValue;
+                        if (useFieldAccess) {
+                            defaultValue = isPresentFalse.readInstanceField(FieldDescriptor.of(field), configObject);
+                        } else {
+                            defaultValue = isPresentFalse.invokeVirtualMethod(
+                                    MethodDescriptor.ofMethod(configObjectClassStr, getterName, fieldTypeStr),
+                                    configObject);
+                        }
+
+                        /*
+                         * If the getter value is null (meaning that no default was set)
+                         * throw the exception that MP Config would.
+                         * If the value is not null, then we don't need to do anything since the class will already
+                         * contain the default value the user intended to use
+                         */
+                        isPresentFalse.ifNull(defaultValue).trueBranch()
+                                .throwException(NoSuchElementException.class,
+                                        "Property " + fullConfigName + " not found");
+                    } else {
+                        /*
+                         * In this case we want a missing property to cause an exception that we don't handle
+                         * So we call config.getValue making sure to handle collection values
+                         */
+                        ResultHandle setterValue = createReadMandatoryValueAndConvertIfNeeded(
+                                fullConfigName, fieldType,
+                                field.declaringClass().name(), methodCreator, config);
+                        createWriteValue(methodCreator, configObject, field, setter, useFieldAccess, setterValue);
+
+                    }
+                }
+            }
+        }
+        return configObject;
+    }
+
+    private static void createWriteValue(BytecodeCreator bytecodeCreator, ResultHandle configObject, FieldInfo field,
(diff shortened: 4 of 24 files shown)


docs/src/main/asciidoc/application-configuration-guide.adoc
```diff
@@ -158,6 +158,127 @@ You can add the `greeting.suffix`, remove the other properties, change the value
...
+    Optional<String> getName(); <3>
+}
+----
+<1> The `@ConfigProperty` annotation is needed because the name of the configuration property that the method corresponds to doesn't follow getter method conventions
+<2> In this case since `name` was not set, the corresponding property will be `greeting.suffix`.
+<3> It is unnecessary to specify the `@ConfigProperty` annotation because the method name follows the getter method conventions (`greeting.name` being the corresponding property) and no default value is needed.
+
+=== Additional notes on @ConfigProperties
+
+When using a regular class annotated with @ConfigProperties the class doesn't necessarily have to getters and setters.
```
mkouba: @ConfigProperties -> `@ConfigProperties`
...the class doesn't necessarily have to DECLARE? getters...

extensions/arc/runtime/src/main/java/io/quarkus/arc/runtime/config/ConfigProperties.java
```diff
@@ -0,0 +1,17 @@
...
+import java.lang.annotation.Target;
+
+/**
+ * Allow configuration properties with a common prefix to be grouped into a single class
+ */
+@Target({ TYPE })
+@Retention(RUNTIME)
+public @interface ConfigProperties {
+
+    String prefix();
```
mkouba: I think that it should be possible to extract the prefix from the class name for simple use cases.

extensions/arc/runtime/src/main/java/io/quarkus/arc/runtime/config/ConfigUtils.java
```diff
@@ -0,0 +1,34 @@
+package io.quarkus.arc.runtime.config;
+
+import java.util.Collection;
+
+import org.eclipse.microprofile.config.Config;
+
+import io.smallrye.config.SmallRyeConfig;
+import io.smallrye.config.StringUtil;
+
+public final class ConfigUtils {
```
mkouba: This class shouldn't be probably public...

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/configproperties/ConfigPropertiesBuildStep.java
```diff
@@ -0,0 +1,97 @@
...
+            }
+        };
+
+        /*
+         * We generate CDI producer bean containing one method for each of the @ConfigProperties
+         * instances we encounter
+         */
+
+        ClassCreator producerClassCreator = ClassCreator.builder().classOutput(beansClassOutput)
+                .className("io.quarkus.config.ConfigPropertiesProducer")
```
mkouba: So we generate a bean that declares a producer for each `@ConfigProperties`, right? In theory, we could register custom beans instead but it really depends on the logic you're adding to the producer methods. I'll take a look at the generated class...

mkouba (changes requested):
I really like this feature ;-). I added few comments and I'm going to checkout the branch and look at the tests and generated stuff.

geoand:
> I really like this feature ;-). I added few comments and I'm going to checkout the branch and look at the tests and generated stuff.

Please do :)

extensions/arc/runtime/src/main/java/io/quarkus/arc/runtime/config/ConfigUtils.java
geoand: It is being called from the generated bytecode. I could of couse move all of it's functionality into generated code. Should I do that?

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/configproperties/ConfigPropertiesBuildStep.java
geoand: Yes exactly. The producer instantiates the class by pulling data from MP Config and setting the proper fields or calling setters

docs/src/main/asciidoc/application-configuration-guide.adoc
geoand: I always forget verbs... Thanks!

extensions/arc/runtime/src/main/java/io/quarkus/arc/runtime/config/ConfigProperties.java
geoand: Certainly. 

I am thinking of the following simple algorithm: 

* Remove any `Properties`, `Configuration`, `Config` suffix
* Make the remainder lowercase and use tha result.

WDYT?

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/configproperties/ConfigPropertiesBuildStep.java
geoand: Also keep in mind that works for both classes and interfaces as config properties holders

extensions/arc/runtime/src/main/java/io/quarkus/arc/runtime/config/ConfigUtils.java
mkouba: We could introduce an abstract super type (`AbstractConfigProperties`) and move the functionality into a protected method?

geoand: Nice, I like that solution a lot! I'll implement it once I'm done trying to integrate validation into this work

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/configproperties/ConfigPropertiesBuildStep.java
mkouba: > Also keep in mind that works for both classes and interfaces as config properties holders

Hm, what do you mean? A generated producer looks like this:

```java
@Produces
public TypicalInterfaceConfigPropertiesTest$DummyProperties produceTypicalInterfaceConfigPropertiesTest$DummyProperties(
			final Config config) {
		return (TypicalInterfaceConfigPropertiesTest$DummyProperties) new TypicalInterfaceConfigPropertiesTest$DummyPropertiesQuarkusImpl(
				config);
}
```

The injected `Config` could be easily obtained using `Arc.container().instance(Config.class).get()`.

On the other hand, having a generated `ConfigPropertiesProducer` is fine. Maybe just use `@Singleton` instead of `@ApplicationScoped` as we don't need client proxies here.

geoand: I see what you mean yes.

TBH, that is what I tried initially for interfaces, but then I changed it because I had to make `Config` and unremovable bean. So the current solution seemed cleaner and more consisent with class handling case.

WDYT?

mkouba: Yes, makes sense. I'd keep the current solution but change the scope of `ConfigPropertiesProducer`.

geoand: :+1: . What is the proper scope for this case?

geoand: :+1: . What is the proper scope for this case?

geoand: Oops, never mind, I see you mentioned `@Singleton` earlier :)

extensions/arc/runtime/src/main/java/io/quarkus/arc/runtime/config/ConfigProperties.java
Ladicek: "Make the remainder lowercase" should be something like "Make the remainder lowercase and insert `-` before the positions that previously contained a first uppercase character" or something like that. I think this already happens somewhere in Quarkus, so code for this should already exist.

geoand: Thanks for pointing it out. I'll check

geoand:
@mkouba Is there anyway I can make the production of Config class "eager"? 
I would like to have the build fail when no property is found. Currently that doesn't happen because the bean produced lazily.

extensions/arc/runtime/src/main/java/io/quarkus/arc/runtime/config/ConfigUtils.java
geoand: I am getting illegal access error all over the place even when I place the generated classes in the same package. 
The only way I don't get the illegal access errors is if the class and method is public - which is exactly what we don't want.

So I'm going to just push the functionality of this method into the generated bytecode.

mkouba: Hm, that's odd - `protected` should not care about the package at all?

geoand: Agreed... But I got fed up trying  to fight it (only public worked) so I will just generate it and all problems will go away (plus it will be totally hidden)

geoand: I ended up generating the a utility class not an abstract superclass because I was getting a warning from `BeanArchives.index`.

So now everything should be nicely hidden from the user and the functionality is back to what it was

extensions/arc/runtime/src/main/java/io/quarkus/arc/runtime/config/ConfigProperties.java
geoand: @IIUC you are proposing that `GreetingSubConfiguration` would translate into `greeting.sub` being used as the prefix, correct?

Ladicek: To `greeting-sub`; I think that's what Quarkus does elsewhere? But `greeting.sub` would also work.

geoand: Quarkus does the former, so let's go with it :)

geoand: Here is the link: https://github.com/quarkusio/quarkus/blob/master/core/deployment/src/main/java/io/quarkus/deployment/configuration/ConfigDefinition.java#L199

Ladicek: Yea, that's it! :-)

geoand: Fixed

geoand:
@mkouba the PR now addresses the concerns you raised. @gsmet it now also includes some rudimentary support for Bean Validation.

geoand:
Marking as work in progress since I'm going to make sure that when validation is present, the application will not startup

mkouba:
> Is there anyway I can make the production of Config class "eager"?

Just add `void onStart(@Observes StartupEvent event) { }` to the bean class. Note that this event is fired when the app starts (RUNTIME_INIT) and not when the container is started (STATIC_INIT).

> I would like to have the build fail when no property is found. 

This is not possible. MP Config takes system properties and ENV variables into account and so you can't fail during the build because it's perfectly valid to consume a property that is only available at runtime.

geoand:
@mkouba Thanks!

I had the same revelation last night and have started to implement it :). Should be done soon

geoand:
@mkouba should be good to go now

mkouba (approved):
Looks good!

geoand:
> Looks good!

Thanks!

@gsmet would you like to take a quick look at the bean validation to make sure I didn't do anything stupid?

geoand:
I am wondering... Should I perhaps rename the annotation to `@ConfigProperties` (from `@ConfigurationProperties`) to be more in line with the names that Microprofile Config uses?

geoand:
The force pushes were done to fix commit messages some of which contained the wrong annotation name

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/configproperties/ConfigPropertiesUtil.java
```diff
@@ -0,0 +1,301 @@
...
+     * Generates a class like:
+     * 
+     * <pre>
+     * final class AbstractConfigPropertiesHelper {
+     *
+     *     void collectionConfigProperty(Config config, String propertyValue,
+     *             Class clazz, Collection collection) {
+     *
+     *         SmallRyeConfig smallRyeConfig = (SmallRyeConfig) config;
+     *         String[] split = StringUtil.split(propertyValue);
```
dmlloyd: This is much more complex than necessary.  You can instead just use [`newCollectionConverter`](https://github.com/smallrye/smallrye-config/blob/master/implementation/src/main/java/io/smallrye/config/Converters.java#L205-L217) and save yourself some work.

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/configproperties/ConfigPropertiesUtil.java
```diff
@@ -0,0 +1,301 @@
...
+                config, bytecodeCreator.load(propertyName),
+                bytecodeCreator.loadClass(effectiveTypeResponse.getEffectiveType().toString()));
+
+        return createConversionToFinalValue(generateConfigPropertiesUtilClassName, resultType.name(),
+                effectiveTypeResponse.getEffectiveType(),
+                effectiveTypeResponse.getGenericType(), bytecodeCreator, value, config);
+    }
+
+    /**
+     * Generates code that uses config.getOptional value to obtain the value of the property.
```
dmlloyd: See also [`newOptionalConverter`](https://github.com/smallrye/smallrye-config/blob/master/implementation/src/main/java/io/smallrye/config/Converters.java#L235-L245)

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/configproperties/ConfigPropertiesUtil.java
geoand: Thanks for pointing it out. I'll have a look

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/configproperties/InterfaceConfigPropertiesUtil.java
```diff
@@ -0,0 +1,255 @@
...
+                                ReadOptionalResponse readOptionalResponse = createReadOptionalValueAndConvertIfNeeded(
+                                        generatedAbstractConfigPropertiesUtilClassName,
+                                        fullConfigName, returnType,
+                                        ifaceDotName, methodCreator, config);
+
+                                // just return the data if the Optional has data
+                                readOptionalResponse.getIsPresentTrue().returnValue(readOptionalResponse.getValue());
+
+                                // if the Optional does not contain data, we need to use the default value making sure we convert it first
+                                BytecodeCreator isPresentFalse = readOptionalResponse.getIsPresentFalse();
```
dmlloyd: Once we sort out some spec issues with MP-config (hopefully tomorrow) then the way to get the default value will be different from this: essentially what you're doing here will not work correctly because if a user forces a property to "empty" then you'll get the default value instead of the empty value.

A more correct approach that works today is to produce a config source which contains your default values at a low priority.  There's also a build item for this purpose: `io.quarkus.deployment.builditem.RunTimeConfigurationDefaultBuildItem`.

geoand: That's great to know! I'll give it a shot

gsmet (changes requested):
As discussed on Zulip, the BV support needs some tweaking. Let's simplify all this before merging.

geoand:
@gsmet @dmlloyd all of your suggestions have been applied (I believe)

core/deployment/src/main/java/io/quarkus/deployment/bean/JavaBeanUtil.java
```diff
@@ -40,4 +40,21 @@ public static String decapitalize(String name) {
...
     }
+
+    /**
+     * Returns the corresponding property name for a getter method name
+     * 
+     * @throws IllegalArgumentException if the method name does not follow the getter name convention
+     */
+    public static String getPropertyNameFromGetter(String methodName) {
+        final String get = "get";
+        final String is = "is";
```
gsmet: Maybe this should be proper constants?

docs/src/main/asciidoc/application-configuration-guide.adoc
```diff
@@ -158,6 +158,144 @@ You can add the `greeting.suffix`, remove the other properties, change the value
 As usual, the application can be packaged using `./mvnw clean package` and executed using the `-runner.jar` file.
 You can also generate the native executable with `./mvnw clean package -Pnative`.
 
+== Using @ConfigProperties
+
+As an alternative to injecting multiple related configuration values in the way that was shown in the previous example,
+users can also use the `@io.quarkus.arc.runtime.config.ConfigProperties` annotation to group these properties together.
```
gsmet: I don't know about Arc but in the other extensions, everything that is API is at the root of the extension package and the `runtime` package is for the internal classes.

@mkouba ?

docs/src/main/asciidoc/application-configuration-guide.adoc
```diff
@@ -158,6 +158,144 @@ You can add the `greeting.suffix`, remove the other properties, change the value
...
+    public Optional<String> getName() {
+        return name;
+    }
+
+    public void setName(Optional<String> name) {
+        this.name = name;
+    }
+}
+----
+<1> `prefix` is optional. If not set then the prefix to be used will be determined by the class name. In this case it would still be `greeting` (since the `Properties` suffix is removed). If the class were named `GreetingExtraProperties` then the resulting prefix would be `greeting-extra`.
```
gsmet:
```suggestion
<1> `prefix` is optional. If not set then the prefix to be used will be determined by the class name. In this case it would still be `greeting` (since the `Properties` suffix is removed). If the class were named `GreetingExtraProperties` then the resulting default prefix would be `greeting-extra`.
```

docs/src/main/asciidoc/application-configuration-guide.adoc
```diff
@@ -158,6 +158,144 @@ You can add the `greeting.suffix`, remove the other properties, change the value
...
+<2> `!` will be the default default value if `greeting.suffix` is not set
```
gsmet:
```suggestion
<2> `!` will be the default value if `greeting.suffix` is not set
```

docs/src/main/asciidoc/application-configuration-guide.adoc
```diff
@@ -158,6 +158,144 @@ You can add the `greeting.suffix`, remove the other properties, change the value
...
+
+This class could then be injected into the `GreetingResource` using the familiar CDI `@Inject` annotation like so.
```
gsmet:
```suggestion
This class could then be injected into the `GreetingResource` using the familiar CDI `@Inject` annotation like so:
```

docs/src/main/asciidoc/application-configuration-guide.adoc
```diff
@@ -158,6 +158,144 @@ You can add the `greeting.suffix`, remove the other properties, change the value
...
+    @ConfigProperty(name = "message") <1>
+    String message();
+
+    @ConfigProperty(defaultValue = "!")
+    String getSuffix(); <2>
+
+    Optional<String> getName(); <3>
+}
+----
+<1> The `@ConfigProperty` annotation is needed because the name of the configuration property that the method corresponds to doesn't follow getter method conventions
```
gsmet:
```suggestion
<1> The `@ConfigProperty` annotation is needed because the name of the configuration property that the method corresponds to doesn't follow the getter method naming conventions.
```

docs/src/main/asciidoc/application-configuration-guide.adoc
```diff
@@ -158,6 +158,144 @@ You can add the `greeting.suffix`, remove the other properties, change the value
...
+<2> In this case since `name` was not set, the corresponding property will be `greeting.suffix`.
+<3> It is unnecessary to specify the `@ConfigProperty` annotation because the method name follows the getter method conventions (`greeting.name` being the corresponding property) and no default value is needed.
```
gsmet:
```suggestion
<3> It is unnecessary to specify the `@ConfigProperty` annotation because the method name follows the getter method naming conventions (`greeting.name` being the corresponding property) and no default value is needed.
```

docs/src/main/asciidoc/application-configuration-guide.adoc
```diff
@@ -158,6 +158,144 @@ You can add the `greeting.suffix`, remove the other properties, change the value
...
+
+=== Additional notes on @ConfigProperties
+
+When using a regular class annotated with @ConfigProperties the class doesn't necessarily have to declare getters and setters.
```
gsmet:
```suggestion
When using a regular class annotated with `@ConfigProperties`, the class doesn't necessarily have to declare getters and setters.
```

docs/src/main/asciidoc/application-configuration-guide.adoc
```diff
@@ -158,6 +158,144 @@ You can add the `greeting.suffix`, remove the other properties, change the value
...
+Having simple public non-final fields is valid as well.
+
+Furthermore the configuration classes support nested object configuration. Suppose there was a need to have an extra layer
+of greeting configuration named hidden that would contain a few fields. This could be achieved like so:
```
gsmet:
```suggestion
of greeting configuration named `hidden` that would contain a few fields. This could be achieved like so:
```

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/configproperties/ConfigPropertiesUtil.java
geoand: Actually this doesn't appear to be in version `1.3.9` of Smallrye Config. How about we leave it as is for now and then I'll update the code when a new version lands?

docs/src/main/asciidoc/application-configuration-guide.adoc
```diff
@@ -158,6 +158,144 @@ You can add the `greeting.suffix`, remove the other properties, change the value
...
+
+[source,properties]
+----
+greeting.message = hello
+greeting.name = quarkus
+greeting.hidden.prizeAmount=10
+greeting.hidden.recipients=Jane,John
+----
+
+Furthermore, classes annotated with `@ConfigProperties` can be annotated with Bean Validation annotations similar the following example:
```
gsmet:
```suggestion
Furthermore, classes annotated with `@ConfigProperties` can be annotated with Bean Validation annotations similar to the following example:
```

geoand:
Actually the one point I missed can't be implemented with the current version of Smallrye Config.
So I propose to consider the comments addressed for the time being :)

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/configproperties/ConfigPropertiesUtil.java
geoand: Done

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/configproperties/ClassConfigPropertiesUtil.java
```diff
@@ -0,0 +1,359 @@
...
+import io.quarkus.gizmo.ResultHandle;
+import io.quarkus.runtime.StartupEvent;
+
+final class ClassConfigPropertiesUtil {
+
+    private static final String JAVAX_VALIDATION_PREFIX = "javax.validation";
+    private static final String VALIDATOR_CLASS = "javax.validation.Validator";
+    private static final String PATH = "javax.validation.Path";
+    private static final String CONSTRAINT_VIOLATION_CLASS = "javax.validation.ConstraintViolation";
+    private static final String HIBERNATE_VALIDATOR_IMPL_CLASS = "org.hibernate.validator.internal.engine.ValidatorImpl";
```
gsmet: Please don't use an internal class, better use `org.hibernate.validator.HibernateValidator` which is part of the public API.

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/configproperties/InterfaceConfigPropertiesUtil.java
geoand: Fixed

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/configproperties/ClassConfigPropertiesUtil.java
```diff
@@ -0,0 +1,359 @@
...
+import io.quarkus.gizmo.FieldCreator;
+import io.quarkus.gizmo.FieldDescriptor;
+import io.quarkus.gizmo.MethodCreator;
+import io.quarkus.gizmo.MethodDescriptor;
+import io.quarkus.gizmo.ResultHandle;
+import io.quarkus.runtime.StartupEvent;
+
+final class ClassConfigPropertiesUtil {
+
+    private static final String JAVAX_VALIDATION_PREFIX = "javax.validation";
```
gsmet: This shouldn't be useful anymore.

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/configproperties/ClassConfigPropertiesUtil.java
```diff
@@ -0,0 +1,359 @@
...
+     * &#64;ApplicationScoped
+     * public class EnsureValidation {
+     *
+     *     &#64;Inject
+     *     MyConfig myConfig;
+     *
+     *     &#64;Inject
+     *     OtherProperties other;
+     *
+     *     public void onStartup(@Observes StartupEvent ev) {
```
gsmet: Why do you need that if it's empty?

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/configproperties/ClassConfigPropertiesUtil.java
```diff
@@ -0,0 +1,359 @@
...
+            throw new IllegalArgumentException(
+                    "Classes annotated with @" + DotNames.CONFIG_PROPERTIES
+                            + " cannot extend other classes. Offending class is "
+                            + configPropertiesClassInfo);
+        }
+
+        if (!configPropertiesClassInfo.hasNoArgsConstructor()) {
+            throw new IllegalArgumentException(
+                    "Class " + configPropertiesClassInfo + " which is annotated with " + DotNames.CONFIG_PROPERTIES
+                            + " must contain a no-arg constructor");
```
gsmet: What? You don't generate the no-args constructor automatically? Are you Georgios or what? :)

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/configproperties/ClassConfigPropertiesUtil.java
```diff
@@ -0,0 +1,359 @@
...
+            } else {
+                methodCreator.returnValue(configObject);
+            }
+        }
+
+        return needsValidation;
+    }
+
+    private static boolean needsValidation() {
+        // Hibernate Validator has minimum overhead, so we'll just pass all config classes to it if it's present
```
gsmet:
```suggestion
        // Hibernate Validator has minimum overhead if the class is unconstrained,
        // so we'll just pass all config classes to it if it's present
```

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/configproperties/ClassConfigPropertiesUtil.java
```diff
@@ -0,0 +1,359 @@
...
+            boolean useFieldAccess = false;
+
+            String setterName = JavaBeanUtil.getSetterName(field.name());
+            Type fieldType = field.type();
+            MethodInfo setter = configClassInfo.method(setterName, fieldType);
+            if (setter == null) {
+                if (!Modifier.isPublic(field.flags()) || Modifier.isFinal(field.flags())) {
+                    throw new IllegalArgumentException(
+                            "Configuration properties class " + configClassInfo + " does not have a setter for field "
+                                    + field + " nor is the field public non-final field");
```
gsmet:
```suggestion
                                    + field + " nor is the field a public non-final field");
```

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/configproperties/ClassConfigPropertiesUtil.java
```diff
@@ -0,0 +1,359 @@
...
+             * If the object is part of the application we are dealing with a nested object
+             * What we do is simply recursively build it up based by adding the field name to the config name prefix
+             */
+            DotName fieldTypeDotName = fieldType.name();
+            String fieldTypeStr = fieldTypeDotName.toString();
+            ClassInfo fieldTypeClassInfo = applicationIndex.getClassByName(fieldType.name());
+            if (fieldTypeClassInfo != null) {
+                if (!fieldTypeClassInfo.hasNoArgsConstructor()) {
+                    throw new IllegalArgumentException(
+                            "Nested configuration class " + fieldTypeClassInfo + " must must contain a no-arg constructor ");
```
gsmet:
```suggestion
                            "Nested configuration class " + fieldTypeClassInfo + " must contain a no-args constructor ");
```

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/configproperties/ClassConfigPropertiesUtil.java
```diff
@@ -0,0 +1,359 @@
...
+                }
+
+                if (!Modifier.isPublic(fieldTypeClassInfo.flags())) {
+                    throw new IllegalArgumentException(
+                            "Nested configuration class " + fieldTypeClassInfo + " must must be public ");
```
gsmet:
```suggestion
                            "Nested configuration class " + fieldTypeClassInfo + " must be public ");
```

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/configproperties/ClassConfigPropertiesUtil.java
```diff
@@ -0,0 +1,359 @@
...
+                    // config.getOptionalValue
+                    ResultHandle setterValue = methodCreator.invokeInterfaceMethod(
+                            MethodDescriptor.ofMethod(Config.class, "getOptionalValue", Optional.class, String.class,
+                                    Class.class),
+                            config, methodCreator.load(fullConfigName), methodCreator.loadClass(genericType.name().toString()));
+                    createWriteValue(methodCreator, configObject, field, setter, useFieldAccess, setterValue);
+                } else {
+                    /*
+                     * We want to support cases where the Config class defines a default value for fields
+                     * by simply specifying the default value in it's constructor
```
gsmet:
```suggestion
                     * by simply specifying the default value in its constructor
```

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/configproperties/ClassConfigPropertiesUtil.java
```diff
@@ -0,0 +1,359 @@
...
+                     * For such cases the strategy we follow is that when a requested property does not exist
+                     * we check the value from the corresponding getter (or read the field value if possible) and if not the
```
gsmet: there's something weird here: "and if not the value is not null"

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/configproperties/ClassConfigPropertiesUtil.java
```diff
@@ -0,0 +1,359 @@
...
+                                    MethodDescriptor.ofMethod(configObjectClassStr, getterName, fieldTypeStr),
+                                    configObject);
+                        }
+
+                        /*
+                         * Throw the exception that MP Config would if the getter value is null (meaning that no default was
+                         * set)
+                         * If the value was not null, then we don't need to do anything since the class will already contain the
+                         * default
+                         * value the user intended to use
```
gsmet: The formatting is a bit chaotic here.

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/configproperties/ClassConfigPropertiesUtil.java
```diff
@@ -0,0 +1,359 @@
...
+        constraintSetIsEmptyBranch.trueBranch().returnValue(configObject);
+
+        BytecodeCreator constraintSetIsEmptyFalse = constraintSetIsEmptyBranch.falseBranch();
+
+        ResultHandle exception = constraintSetIsEmptyFalse.newInstance(
+                MethodDescriptor.ofConstructor(CONSTRAINT_VIOLATION_EXCEPTION_CLASS, Set.class.getName()), validationResult);
+        constraintSetIsEmptyFalse.throwException(exception);
+
+        constraintSetIsEmptyBranch.falseBranch().throwException(IllegalArgumentException.class,
+                "Configuration group " + configPrefix + " contains invalid values. The errors are: ");
```
gsmet: Hmmm, this should go away, shouldn't it?

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/configproperties/DotNames.java
```diff
@@ -0,0 +1,26 @@
...
+final class DotNames {
+
+    private DotNames() {
+    }
+
+    static final DotName OBJECT = DotName.createSimple(Object.class.getName());
+    static final DotName STRING = DotName.createSimple(String.class.getName());
+    static final DotName OPTIONAL = DotName.createSimple(Optional.class.getName());
+    static final DotName LIST = DotName.createSimple(List.class.getName());
+    static final DotName SET = DotName.createSimple(Set.class.getName());
```
gsmet: We don't support maps, here? (not that it's strictly necessary to get this PR in, it can be a later addition)

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/configproperties/InterfaceConfigPropertiesUtil.java
```diff
@@ -0,0 +1,237 @@
...
+                                    MethodDescriptor.ofMethod(Config.class, "getOptionalValue", Optional.class, String.class,
+                                            Class.class),
+                                    config, methodCreator.load(fullConfigName),
+                                    methodCreator.loadClass(genericType.name().toString()));
+                            methodCreator.returnValue(result);
+                        } else {
+                            if (defaultValueStr != null) {
+                                /*
+                                 * The effect this will have is to add a Config Source with a lower priority thus ensuring
+                                 * that when pull a value of using config.getValue(fullConfigName) the default value
```
gsmet: This sentence is weird but I'm not sure what needs fixing.

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/configproperties/InterfaceConfigPropertiesUtil.java
```diff
@@ -0,0 +1,237 @@
...
+                                 * will be returned if none is set
+                                 */
+                                defaultConfigValues
+                                        .produce(new RunTimeConfigurationDefaultBuildItem(fullConfigName, defaultValueStr));
+                            }
+                            // use config.getValue to obtain and return the result taking conversion into a collection if needed
```
gsmet:
```suggestion
                            // use config.getValue to obtain and return the result converting it to a collection if needed
```
?

gsmet (dismissed):
Nice work. I added a bunch of comments inline.

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/configproperties/DotNames.java
geoand: Nope. I didn't see anything in MP Config for maps, but I could very well have missed it. 
A very good addition for the next iteration :)

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/configproperties/ClassConfigPropertiesUtil.java
geoand: Awesome catch!

geoand:
@gsmet Thanks for the thorough review! All comments should now be addressed

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/configproperties/InterfaceConfigPropertiesUtil.java
geoand: Hopefully it's better now

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/configproperties/ClassConfigPropertiesUtil.java
geoand: The whole trick here is to force the beans to be produced eagerly. By injecting the config classes into this bean and adding an `@Observes StartupEvent` we ensure that Arc runs the producer thus ensuring validation happens at startup

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/configproperties/ClassConfigPropertiesUtil.java
geoand: Lololololo!

I actually hadn't thought of what, but I don't think it's possible because we would miss out on field initialization which in this case matters

geoand:
@gsmet Now everything should be properly addressed.

core/runtime/src/main/java/io/quarkus/runtime/configuration/HashSetFactory.java
```diff
@@ -0,0 +1,23 @@
...
+ * A helper class to produce a right-sized hash set.
+ */
+public final class HashSetFactory<T> implements IntFunction<HashSet<T>> {
+    private static final HashSetFactory<?> INSTANCE = new HashSetFactory<>();
+
+    private HashSetFactory() {
+    }
+
+    public HashSet<T> apply(final int value) {
+        return new HashSet<>(value);
```
gsmet: I don't think it works as you think it is: the number of initial elements is dependent of the fill factor. Right now, you will always create a slightly undersized map.

docs/src/main/asciidoc/application-configuration-guide.adoc
gsmet: @geoand we need to sort that out before merging as this annotation will be API.

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/configproperties/ClassConfigPropertiesUtil.java
gsmet: Wouldn't using `@Singleton` beans have the same consequence and avoid some boilerplate? I might be wrong, better check with @mkouba

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/configproperties/ConfigPropertiesUtil.java
gsmet: Maybe we should create an issue? Pretty sure we will forget about it.

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/configproperties/DotNames.java
gsmet: Yeah, I think the maps are an extension specifically developed by @dmlloyd .

gsmet:
@geoand there are still a few things we need to sort out before merging. /cc @mkouba for advice.

core/runtime/src/main/java/io/quarkus/runtime/configuration/HashSetFactory.java
geoand: Not sure, but we use the ArrayList one in Quarkus Build time Config. 
@dmlloyd care to comment here?

geoand:
@gsmet sure thing

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/configproperties/DotNames.java
geoand: I'll check and see if there is a converter. If so, we can use it

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/configproperties/ConfigPropertiesUtil.java
geoand: I was planning on it yes once the PR is merged

dmlloyd (approved):

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/configproperties/ConfigPropertiesUtil.java
geoand: Not a problem at all! I'll remove it when the new version lands

docs/src/main/asciidoc/application-configuration-guide.adoc
mkouba: The problem is that the package `io.quarkus.arc` is used in the `io.quarkus.arc:arc` artifact and we would like to avoid a package clash to prevent problems with Java 9+. So we could either rename the package in `io.quarkus.arc:arc` or use a different package in the runtime module. E.g. `io.quarkus.arc.api`...

geoand: Whatever you folks think is best. It's obviously no trouble for me to move the annotation

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/configproperties/ClassConfigPropertiesUtil.java
geoand: Instead of @ApplicationScoped you mean? That sounds reasonable to me.

I just used whatever I found in some existing code.

@mkouba should I change to `@Singleton`?

core/runtime/src/main/java/io/quarkus/runtime/configuration/HashSetFactory.java
geoand: Thanks for the explanation!

geoand:
So it seems like there are 2 things left on this:

* What the proper package name should be
* If the bean that defines `@Observes` StartupEven should become a `@Singleton` or can be left as `@ApplicationScoped`.

@mkouba I think the decisions rest in your hands :)

geoand:
@gsmet After consulting with @mkouba the issues that you raised have been addressed

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/configproperties/ClassConfigPropertiesUtil.java
gsmet: From what I understood, singletons are always initialized eagerly. So that might solve your issue and you might be able to drop the `onStartup` method. But I might be wrong on that :).

core/runtime/src/main/java/io/quarkus/runtime/configuration/HashSetFactory.java
gsmet: @geoand see this comment in HV where we try to size properly all the collections that are in the metadata: https://github.com/hibernate/hibernate-validator/blob/master/engine/src/main/java/org/hibernate/validator/internal/util/CollectionHelper.java#L130

It's not an issue with `ArrayList` as it sizes the array exactly with the size you pass. In any case, it's a detail.

gsmet:
Yeah, can't say I'm excited about the `api` sub package as it's not consistent with the rest of the code and the rest of our APIs. That's why I wanted us to have a more future proof package organization but this ship has sailed, unfortunately.

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/configproperties/ClassConfigPropertiesUtil.java
geoand: We discussed with @mkouba and the current implementation is the probably the best way to guarantee it always works

core/runtime/src/main/java/io/quarkus/runtime/configuration/HashSetFactory.java
geoand: I can add same method if you want, it's not a problem obviously :)

geoand: Added :)

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/configproperties/ClassConfigPropertiesUtil.java
mkouba: > From what I understood, singletons are always initialized eagerly.

Not really. A singleton is initialized eagerly when injected. This bean is used to initialize other beans when the app starts. That's why we use `StartupEvent` and inject other beans and invoke `toString()` upon them because a lazy proxy would be injected for a normal scoped bean.

gsmet: Ok, thanks for the explanation, Martin :).

gsmet:
@mkouba @geoand what about `io.quarkus.arc.config`? Unless it's used in Arc too?

I'm a bit reluctant to introduce an artificial `api` package whereas we don't have one anywhere else. Sorry about my OCD :).

geoand:
> @mkouba @geoand what about `io.quarkus.arc.config`? Unless it's used in Arc too?
> 
> I'm a bit reluctant to introduce an artificial `api` package whereas we don't have one anywhere else. Sorry about my OCD :).

I'm fine with whatever you think :)

mkouba:
I like it. Let's use `io.quarkus.arc.config`!

geoand:
PR updated with the proper package name

geoand:
@gsmet are we ok to merge this?

gsmet (approved):

geoand:
Commits are now squashed

geoand:
There is a CI infra related failure. I'll restart once the remaining jobs complete

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/configproperties/ConfigPropertiesUtil.java
geoand: Opened #4166 to clean this up when Smallrye Config is released

Issue #1029, opened 2019-02-23
Windows - console (cmd/powershell) polluted with escape characters for color output


rsvoboda:
Windows - console (cmd/powershell) polluted with escape characters for color output

~~I don't think cmd supports colors.~~ Cmd supports colors too, e.g. maven output is colored
PowerShell supports colors, but probably needs its own way.

Workaround on Windows could be to use consoles with color support - e.g. https://cmder.net/

Example:
```
java -jar .\target\demo-1.0-SNAPSHOT-runner.jar
[38;5;145m2019-02-23 09:43:45,028[39m[38;5;188m [39m[38;5;83mdesktop-agm44t8[39m[38;5;188m [39m[38;5;186mdemo-1.0-SNAPSHOT-runner.jar[39m[38;5;188m[[39m[38;5;180m8068[39m[38;5;188m] [39m[38;5;107mINFO [39m[38;5;188m [[39m[38;5;69mi.s.f.HystrixInitializer[39m[38;5;188m] ([39m[38;5;71mmain[39m[38;5;188m) [39m[38;5;151m[38;5;231m### Init Hystrix ###[39m[39m[38;5;203m[39m[38;5;227m
[39m[38;5;145m2019-02-23 09:43:45,060[39m[38;5;188m [39m[38;5;83mdesktop-agm44t8[39m[38;5;188m [39m[38;5;186mdemo-1.0-SNAPSHOT-runner.jar[39m[38;5;188m[[39m[38;5;180m8068[39m[38;5;188m] [39m[38;5;107mINFO [39m[38;5;188m [[39m[38;5;69mi.s.f.DefaultHystrixConcurrencyStrategy[39m[38;5;188m] ([39m[38;5;71mmain[39m[38;5;188m) [39m[38;5;151m[38;5;231m### Privilleged Thread Factory used ###[39m[39m[38;5;203m[39m[38;5;227m
[...]

PR #4076, opened 2019-09-17
fix: stop color mode on windows cmd/powershell

maxandersen:
Why:

 * by default ANSI colors are not enabled for cmd/powershell
   on Windows 10 machines so users will by default get garbled output.

 * Detect when running on Windows and NOT in known working windows
   emulators and there not enable colors.

Solves #1029

### DIFF
--- a/core/runtime/src/main/java/io/quarkus/runtime/logging/LoggingSetupRecorder.java
+++ b/core/runtime/src/main/java/io/quarkus/runtime/logging/LoggingSetupRecorder.java
@@ -7,6 +7,7 @@ import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.logging.ErrorManager;
@@ -35,6 +36,25 @@ import io.quarkus.runtime.annotations.Recorder;
  */
 @Recorder
 public class LoggingSetupRecorder {
+
+    static final boolean IS_WINDOWS = System.getProperty("os.name").toLowerCase(Locale.ENGLISH).contains("win");
+
+    /**
+     * <a href="https://conemu.github.io">ConEmu</a> ANSI X3.64 support enabled,
+     * used by <a href="https://cmder.net/">cmder</a>
+     */
+    static final boolean IS_CON_EMU_ANSI = "ON".equals(System.getenv("ConEmuANSI"));
+
+    static final boolean IS_CYGWIN = IS_WINDOWS
+            && System.getenv("PWD") != null
+            && System.getenv("PWD").startsWith("/")
+            && !"cygwin".equals(System.getenv("TERM"));
+
+    static final boolean IS_MINGW_XTERM = IS_WINDOWS
+            && System.getenv("MSYSTEM") != null
+            && System.getenv("MSYSTEM").startsWith("MINGW")
+            && "xterm".equals(System.getenv("TERM"));
+
     public LoggingSetupRecorder() {
     }
 
@@ -73,10 +93,29 @@ public class LoggingSetupRecorder {
         InitialConfigurator.DELAYED_HANDLER.setHandlers(handlers.toArray(EmbeddedConfigurator.NO_HANDLERS));
     }
 
+    private boolean workingConsole() {
+        if (System.console() != null) {
+            if (IS_WINDOWS && !(IS_CON_EMU_ANSI || IS_CYGWIN || IS_MINGW_XTERM)) {
+                // On Windows without a known good emulator
+                // TODO: optimally we would check if Win32 getConsoleMode has
+                // ENABLE_VIRTUAL_TERMINAL_PROCESSING enabled or enable it via 
+                // setConsoleMode.
+                // For now we turn it off to not generate noisy output for most
+                // users.
+                return false;
+            } else {
+                // Must be on some Unix variant or ANSI-enabled windows terminal...
+                return true;
+            }
+        } else {
+            return false;
+        }
+    }
+
     private ErrorManager configureConsoleHandler(ConsoleConfig config, ErrorManager errorManager,
             List<LogCleanupFilterElement> filterElements, ArrayList<Handler> handlers) {
         final PatternFormatter formatter;
-        if (config.color && System.console() != null) {
+        if (config.color && workingConsole()) {
             formatter = new ColorPatternFormatter(config.darken, config.format);
         } else {
             formatter = new PatternFormatter(config.format);
--- a/core/runtime/src/main/java/io/quarkus/runtime/logging/LoggingSetupRecorder.java
+++ b/core/runtime/src/main/java/io/quarkus/runtime/logging/LoggingSetupRecorder.java
@@ -93,7 +93,8 @@ public class LoggingSetupRecorder {
         InitialConfigurator.DELAYED_HANDLER.setHandlers(handlers.toArray(EmbeddedConfigurator.NO_HANDLERS));
     }
 
-    private boolean workingConsole() {
+    private boolean hasColorSupport() {
+
         if (System.console() != null) {
             if (IS_WINDOWS && !(IS_CON_EMU_ANSI || IS_CYGWIN || IS_MINGW_XTERM)) {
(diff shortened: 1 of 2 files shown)


maxandersen:
Work in progress - just opened PR to get feedback on general idea.

My windows virtual machine box been building quarkus master (without tests) for the last 30 minutes and its over midnight so i'm going to sleep while my laptop spends some cpu cycles.

Tomorrow i'll test and verify it works as expected.

dmlloyd:
You can create a pull request as a "draft" PR.  It's a relatively recent feature.  When you go to open the pull request, the "create pull request" button has a dropdown menu on it; choose "open as draft" from there.

dmlloyd (approved):

core/runtime/src/main/java/io/quarkus/runtime/logging/LoggingSetupRecorder.java
```diff
@@ -35,6 +36,25 @@
...
+    /**
+     * <a href="https://conemu.github.io">ConEmu</a> ANSI X3.64 support enabled,
+     * used by <a href="https://cmder.net/">cmder</a>
+     */
+    static final boolean IS_CON_EMU_ANSI = "ON".equals(System.getenv("ConEmuANSI"));
+
+    static final boolean IS_CYGWIN = IS_WINDOWS
+            && System.getenv("PWD") != null
+            && System.getenv("PWD").startsWith("/")
+            && !"cygwin".equals(System.getenv("TERM"));
```
gsmet: if this is correct, maybe it requires a comment. It feels weird to check the `TERM` is not `cygwin` for Cygwin.

core/runtime/src/main/java/io/quarkus/runtime/logging/LoggingSetupRecorder.java
```diff
@@ -73,10 +93,29 @@ public void initializeLogging(LogConfig config) {
         InitialConfigurator.DELAYED_HANDLER.setHandlers(handlers.toArray(EmbeddedConfigurator.NO_HANDLERS));
     }
 
+    private boolean workingConsole() {
```
gsmet: Maybe it's not the right name? It's just about color support, not if the console is working or not.

core/runtime/src/main/java/io/quarkus/runtime/logging/LoggingSetupRecorder.java
maxandersen: this is taken literally as is from jansi assuming they were right :)

core/runtime/src/main/java/io/quarkus/runtime/logging/LoggingSetupRecorder.java
gsmet: I suggest: `hasColorSupport` or `supportsColors`

core/runtime/src/main/java/io/quarkus/runtime/logging/LoggingSetupRecorder.java
maxandersen: for what it is worth - TERM=xterm on default cygwin.

core/runtime/src/main/java/io/quarkus/runtime/logging/LoggingSetupRecorder.java
maxandersen: fixed.

core/runtime/src/main/java/io/quarkus/runtime/logging/LoggingSetupRecorder.java
maxandersen: https://github.com/fusesource/jansi/commit/bb3d538315c44f799d34fd3426f6c91c8e8dfc55 is when it was added. 

Googling around a bit seems `TERM=cygwin` used mainly to trick some other tools to think you are running cygwin when you aren't :)

maxandersen: added comment and consider it correct as from jansi used by mvn and others.

maxandersen:
Okey, removing the WIP moniker as i finally got this tested.

Had to remove System.console() test under windows as when running on cygwin console is null.



screenshot above shows same app running in cygwin (left) and plain cmd.exe (right).

Optimally I would like to have a way for users to force color on in case they have set the registry flag. Suggestions welcome :)

core/runtime/src/main/java/io/quarkus/runtime/logging/LoggingSetupRecorder.java
```diff
@@ -35,6 +36,29 @@
  */
 @Recorder
 public class LoggingSetupRecorder {
+
+    static final boolean IS_WINDOWS = System.getProperty("os.name").toLowerCase(Locale.ENGLISH).contains("win");
```
gastaldi: Just for the sake of avoiding possible NPE, it would be nice to use `getProperty("os.name", "Linux")`, WDYT?

maxandersen: if that happens you are on a non-java vm :)

core/runtime/src/main/java/io/quarkus/runtime/logging/LoggingSetupRecorder.java
```diff
@@ -35,6 +36,29 @@
...
+
+    /**
+     * <a href="https://conemu.github.io">ConEmu</a> ANSI X3.64 support enabled,
+     * used by <a href="https://cmder.net/">cmder</a>
+     */
+    static final boolean IS_CON_EMU_ANSI = "ON".equals(System.getenv("ConEmuANSI"));
```
dmlloyd: This should be `IS_WINDOWS && ...` to avoid the extra `getenv`.

core/runtime/src/main/java/io/quarkus/runtime/logging/LoggingSetupRecorder.java
```diff
@@ -73,10 +97,32 @@ public void initializeLogging(LogConfig config) {
...
+                // On Windows without a known good emulator
+                // TODO: optimally we would check if Win32 getConsoleMode has
+                // ENABLE_VIRTUAL_TERMINAL_PROCESSING enabled or enable it via 
+                // setConsoleMode.
+                // For now we turn it off to not generate noisy output for most
+                // users.
+                return false;
+            } else {
+                // Must be on some Unix variant or ANSI-enabled windows terminal...
+                return true;
```
dmlloyd: This should also return `System.console() != null` IMO.  Otherwise if you pipe through e.g. `more`, you'll get lots of garbage.

dmlloyd (changes requested):

core/runtime/src/main/java/io/quarkus/runtime/logging/LoggingSetupRecorder.java
maxandersen: then cygwin won't work.

maxandersen: but yeah - just tested and piping goes bad...dammit.

dmlloyd: It's OK if cygwin doesn't work IMO.  They can enable colors manually.

maxandersen: note: `mvn compile | less` on cygwin results in color dumps too.

dmlloyd: cygwin will likely soon stop returning `null` for console: https://sourceware.org/git/?p=newlib-cygwin.git;a=commit;h=169d65a5774acc76ce3f3feeedcbae7405aa9b57

dmlloyd (approved):
Close enough, we can improve in a subsequent PR!

core/runtime/src/main/java/io/quarkus/runtime/logging/LoggingSetupRecorder.java
maxandersen: fixed in latest force push :)

PR #4083, opened 2019-09-18
TCK for MP Reactive Messaging

jmartisk:
@mkouba @cescoffier @kenfinnigan

### DIFF
--- a/extensions/smallrye-reactive-messaging/deployment/src/main/java/io/quarkus/smallrye/reactivemessaging/deployment/SmallRyeReactiveMessagingProcessor.java
+++ b/extensions/smallrye-reactive-messaging/deployment/src/main/java/io/quarkus/smallrye/reactivemessaging/deployment/SmallRyeReactiveMessagingProcessor.java
@@ -8,6 +8,9 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 
+import javax.enterprise.context.Dependent;
+import javax.enterprise.inject.spi.DeploymentException;
+
 import org.eclipse.microprofile.config.spi.ConfigProviderResolver;
 import org.eclipse.microprofile.reactive.messaging.Incoming;
 import org.eclipse.microprofile.reactive.messaging.Outgoing;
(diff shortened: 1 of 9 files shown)


extensions/smallrye-reactive-messaging/deployment/src/main/java/io/quarkus/smallrye/reactivemessaging/deployment/SmallRyeReactiveMessagingProcessor.java
```diff
@@ -79,8 +120,17 @@ void validateBeanDeployment(
             if (bean.isClassBean()) {
                 // TODO: add support for inherited business methods
```
mkouba: The `AnnotationsTransformer` above does support inherited business methods but we ignore these methods here. I think we should be consistent and improve this part as well ;-).

mkouba: Otherwise LGTM.

kenfinnigan (approved):

extensions/smallrye-reactive-messaging/deployment/src/main/java/io/quarkus/smallrye/reactivemessaging/deployment/SmallRyeReactiveMessagingProcessor.java
jmartisk: Hm I'm looking at the spec and its API, and AFAICS the `@Incoming` and `@Outgoing` annotations are not inherited, so I guess we don't need to care about inheritance here.
In that case it would rather make sense to remove the inheritance tree walking in `AnnotationsTransformer`? Is that right?

mkouba: Well, it seems the spec does not define this. So I would take into account the behavior defined by Commons Annotations: _"Members inherited from a superclass and which are not hidden or overridden maintain the annotations they had in the class that declared them..."_. In other words, those methods should be probably inherited.

However, the algorithm wouldn't be that simple, for example if we want to handle generics correctly.

And so I'd suggest to remove the inheritance processing from the annotation transformer and file a new issue to address the inheritance problem properly.

jmartisk: Ok, I removed the inheritance processing  and filed https://github.com/quarkusio/quarkus/issues/4106

mkouba (approved):

PR #4091, opened 2019-09-18
Upgrade to jboss-parent pom v.36

Sanne:
This combined with PR #4084 should significantly reduce the errors logged by IDEA on project import.

Mostly related to https://github.com/jboss/jboss-parent-pom/pull/68


gsmet (approved):
We need to wait for it being able on Central before running CI again.

Issue #4056, opened 2019-09-17
Spring Data's Page cannot be returned from a Spring Web controller


geoand:
See [this](https://github.com/quarkusio/quarkus/issues/4041#issuecomment-532081752) for details

geoand:
This was fixed in #4098

PR #4098, opened 2019-09-18
Replace JSON-B with Jackson for Spring Web

geoand:
This is done because Jackson is what Spring users are used. It will
also pave the way to implement #4056 and related features which rely
on Jackson


gsmet (approved):
Makes sense.

Issue #4099, opened 2019-09-18
Write down exception stack in reverse order in 500 error page


emmanuelbernard:
Play had a nice feature where they showed the most nested cause exception text and stack first and then moved up the chain.
A lot of time, you want to see the nested one which has the most context.

FroMage:
Didn't know of this feature, but yeah, it's nice.

jaikiran:
@emmanuelbernard Does this look like what you had in mind https://github.com/quarkusio/quarkus/pull/4109?

emmanuelbernard:
Turns out it wasn’t from play. They do something different. But I still
think it’s a good feature.

emmanuelbernard:
gentle ping :)

emmanuelbernard:
Got the same comment while driving a lab. Getting the first exception first is preferablr

jaikiran:
Hello @emmanuelbernard, I will get back to this, this week.

jaikiran:
@emmanuelbernard Here's a version which I think achieves what you wanted https://github.com/quarkusio/quarkus/pull/4904

jaikiran:
FWIW, here's how it actually looks like, now that this has made into the release - this image https://user-images.githubusercontent.com/846671/68887904-afac7000-073b-11ea-8b7b-34bdad16f71b.png from https://github.com/quarkusio/quarkus/issues/5488

PR #4109, opened 2019-09-19
issue#4099 - Show the root cause first in the 500 error page

jaikiran:
The commit here implements the idea suggested in https://github.com/quarkusio/quarkus/issues/4099. With this change, the generated HTML will show the root cause first (if there is any nested cause), before showin the complete stacktrace.

One such sample of the HTML (generated from the testcase included in this PR) will look like https://gist.github.com/jaikiran/67264f3c22964c02b0c34e47a3692739. The root cause is displayed first https://gist.github.com/jaikiran/67264f3c22964c02b0c34e47a3692739#file-error-page-with-root-cause-html-L105 followed by the complete nested stacktrace https://gist.github.com/jaikiran/67264f3c22964c02b0c34e47a3692739#file-error-page-with-root-cause-html-L172

FWIW - I didn't make this configurable since I don't think the configuration will add much value, since what we are doing here is just changing the way we display/present content on (the default) error page.

### DIFF
--- a/core/deployment/src/main/java/io/quarkus/deployment/devmode/ReplacementDebugPage.java
+++ b/core/deployment/src/main/java/io/quarkus/deployment/devmode/ReplacementDebugPage.java
@@ -20,8 +20,21 @@ public class ReplacementDebugPage {
     }
 
     private static String generateStackTrace(final Throwable exception) {
-        StringWriter stringWriter = new StringWriter();
-        exception.printStackTrace(new PrintWriter(stringWriter));
+        final StringWriter stringWriter = new StringWriter();
+        final PrintWriter pw = new PrintWriter(stringWriter);
+        // if we have a nested cause, we go all the way to the root cause
+        // and print it first, before printing the whole stacktrace
+        if (exception.getCause() != null) {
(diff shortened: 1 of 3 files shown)


geoand:
Thanks for taking this on @jaikiran!

@emmanuelbernard would you like to verify if this fits you expectations?

emmanuelbernard:
So what I had in mind is to reverse the whole stack. Not just showing the most nested one first. Actually we could also add a simple JS driven button to flip the stack ordering (using most inner first as default).

emmanuelbernard:
BTW play framework was doing something else. It shows the code that triggered the exception and a few lines before and after. But I don’t know if that’s what we want.

aloubyansky:
@jaikiran what do you think of @emmanuelbernard comments? Are you leaving it as it is? Thanks!

jaikiran:
Hi @aloubyansky, I forgot about this. 
> So what I had in mind is to reverse the whole stack.

I think for this it's probably better to do it in javascript. I'm not an expert in that but I can get to it sometime later. For now, this PR is no longer valid. So I'm going to close it.

PR #4112, opened 2019-09-19
Fully convert Arc to JUnit 5.

### LINKED ISSUE #4105: Arc - fully move testing to Junit5
As mentioned in #3607, we should drop Junit4 and fully replace that with Junit5.
This entails rewriting Junit4 `@Rule` `ArcTestContainer` into a Junit5 extension and then rewriting all tests.

manovotn:
There is loads of files since effectively all tests needed some import changes in the least.
The important parts are around `POM`s and then inside `ArcTestContainer` which was transferred from Junit4 rule to Junit5 extension while keeping the functionality.

Fixes #4105

### DIFF
--- a/independent-projects/arc/pom.xml
+++ b/independent-projects/arc/pom.xml
@@ -89,14 +89,7 @@
 
             <dependency>
                 <groupId>org.junit.jupiter</groupId>
-                <artifactId>junit-jupiter-api</artifactId>
-                <version>${version.junit5}</version>
-                <scope>test</scope>
-            </dependency>
-
-            <dependency>
-                <groupId>org.junit.jupiter</groupId>
-                <artifactId>junit-jupiter-engine</artifactId>
+                <artifactId>junit-jupiter</artifactId>
                 <version>${version.junit5}</version>
                 <scope>test</scope>
             </dependency>
--- a/independent-projects/arc/pom.xml
+++ b/independent-projects/arc/pom.xml
@@ -37,7 +37,7 @@
         <!-- Versions -->
         <version.cdi>2.0.SP1</version.cdi>
         <version.jandex>2.1.1.Final</version.jandex>
-        <version.junit4>4.12</version.junit4>
+        <version.junit5>5.5.2</version.junit5>
         <version.maven>3.5.2</version.maven>
         <version.assertj>3.12.2</version.assertj>
         <version.jboss-logging>3.3.2.Final</version.jboss-logging>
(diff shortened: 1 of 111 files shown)


manovotn:
As a side note, I am not sure that the functionality now done via `ArcTestContainer.getFailure()` (i.e. getting the cause of failed deployment) is performed correctly. It works like this, but perhaps we should rather use the `ExtensionContext.Store` for the `Throwable` cause as well?

independent-projects/arc/processor/src/test/java/io/quarkus/arc/processor/DotNamesTest.java
```diff
@@ -1,20 +1,22 @@
...
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
 
 public class DotNamesTest {
 
     @Test
     public void testSimpleName() throws IOException {
         Index index = index(Nested.class, NestedNested.class, DotNamesTest.class);
-        assertEquals("Nested", DotNames.simpleName(index.getClassByName(DotName.createSimple(Nested.class.getName()))));
+        Assertions.assertEquals("Nested",
```
mkouba: Could we use static imports everywhere?

manovotn: Some tests used them, some didn't. I merely converted the imports into Junit5 style.
Is there any benefit to doing this? I mean, next time I write a test, I am pretty sure I will write it without static imports yet again ;)

mkouba: Aesthetics and uniformity ;-). Ok, just ignore my comment...

mkouba:
> It works like this, but perhaps we should rather use the ExtensionContext.Store for the Throwable cause as well?

How would you assert the throwable if it's set in the store?

manovotn:
> > It works like this, but perhaps we should rather use the ExtensionContext.Store for the Throwable cause as well?
> 
> How would you assert the throwable if it's set in the store?

I had that solution for a bit in the code.
Basically, `ArcTestContainer.getFailure()` would query the `Store`. But to achieve that, you need to keep a reference to either the store, or the extension context somewhere, perhaps in a static variable. Then you would leverage the fact that the root store is shared for those tests, so what you put there, you can always retrieve, no matter if your extension context is slightly different (for instance any method gets new extension context).

independent-projects/arc/pom.xml
```diff
@@ -88,9 +88,16 @@
             </dependency>
 
             <dependency>
-                <groupId>junit</groupId>
-                <artifactId>junit</artifactId>
-                <version>${version.junit4}</version>
+                <groupId>org.junit.jupiter</groupId>
+                <artifactId>junit-jupiter-api</artifactId>
```
gastaldi: You can depend on `org.junit.jupiter:junit-jupiter` only (which will bring the necessary dependencies)

manovotn: Good point, done that!

mkouba (approved):

PR #4116, opened 2019-09-19
Add native-image.docker-build=true in the native profile

gastaldi:
That will save developers from having to install GraalVM locally and provide a much nicer user experience when compiling to native (specially because developers wouldn't need to care about upgrading GraalVM everytime a new version is available)

@emmanuelbernard @cescoffier @gsmet WDYT?


gsmet:
You asked for my opinion: I don't like it :).

We tried hard to have the default build not depending on Docker.

gastaldi:
@gsmet that's fine, I was looking for this kind of feedback :smiley: 

Now that I think about it it may not work in Windows environments. I'll close it for now

PR #4117, opened 2019-09-19
Implement initial-sql configuration for Agroal

chaminnk:


docs/src/main/asciidoc/datasource-guide.adoc
```diff
@@ -230,5 +230,12 @@ AgroalDataSource dataSource2;
 [[configuration-reference]]
 == Agroal Configuration Reference
 
+|===
```
machi1990: This doc change is not needed since the config documentation will be generated automatically from the javadoc: (thanks to the include directive below)

machi1990 (changes requested):

extensions/agroal/deployment/src/test/java/io/quarkus/agroal/test/DefaultDataSourceConfigTest.java
```diff
@@ -35,12 +35,13 @@
...
         testDataSource(defaultDataSource, "username-default", 3, 13, 7, Duration.ofSeconds(53), Duration.ofSeconds(54),
-                Duration.ofSeconds(55), Duration.ofSeconds(56), Duration.ofSeconds(57));
+                Duration.ofSeconds(55), Duration.ofSeconds(56), Duration.ofSeconds(57),
+                "create schema if not exists schema_default");
     }
 
     private static void testDataSource(AgroalDataSource dataSource, String username, int minSize, int maxSize,
             int initialSize, Duration backgroundValidationInterval, Duration acquisitionTimeout, Duration leakDetectionInterval,
-            Duration idleRemovalInterval, Duration maxLifetime) throws SQLException {
+            Duration idleRemovalInterval, Duration maxLifetime, String initialSQL) throws SQLException {
```
gsmet: It's very minor but let's be consistent and call the variable `initialSql`.

extensions/agroal/runtime/src/main/java/io/quarkus/agroal/runtime/DataSourceRuntimeConfig.java
```diff
@@ -87,4 +87,10 @@
      */
     @ConfigItem
     public boolean enableMetrics;
+
+    /**
+     * Set up initial sql configuration of a connection.
+     */
+    @ConfigItem
+    public Optional<String> initialSQL;
```
gsmet: Same here: let's be consistent with Agroal naming and call it `initialSql`.

docs/src/main/asciidoc/datasource-guide.adoc
chaminnk: Thanks for notifying me that. I'll change it.

extensions/agroal/runtime/src/main/java/io/quarkus/agroal/runtime/DataSourceRuntimeConfig.java
```diff
@@ -87,4 +87,10 @@
      */
     @ConfigItem
     public boolean enableMetrics;
+
+    /**
+     * Set up initial sql configuration of a connection.
```
gsmet:
```suggestion
     * Initial SQL query executed when first using a connection.
```
would be better :)

gsmet (changes requested):
Thanks for this. Generally, it looks good!

I added a couple of minor comments inline. Could you take care of them and amend your commit?

Thanks!

docs/src/main/asciidoc/datasource-guide.adoc
gsmet: Yup and CI is failing because of this so better get rid of it!

extensions/agroal/deployment/src/test/java/io/quarkus/agroal/test/DefaultDataSourceConfigTest.java
chaminnk: Sure

gsmet:
@chaminnk looks good to me. Could you squash all your commits into one? Then we will wait for CI.

chaminnk:
Sorry about not amending in the first place. I'm still learning Git commands. I squashed the individual commits into the commit hash c5d6d5a. Is that okay?

gsmet:
Well, we have 8 commits now so I would say no :).

The history is a bit of a mess, I will fix it.

machi1990 (approved):
LGTM. The test failure seems unrelated.  
Let's wait for @gsmet go ahead.

chaminnk:
Thank you for the reviews and helping me out @gsmet @machi1990 :) . Sorry about messing up the commits. Hopefully it will not happen again. Looking forward to contributing again :)

machi1990:
> Thank you for the reviews and helping me out @gsmet @machi1990 :) . 

Been a pleasure reviewing your work.

> Sorry about messing up the commits. Hopefully it will not happen again. 

Worry out, it can happen and we are here to help each ones out. I have a few `git` pointers
that may help out, you can reach me out. 

> Looking forward to contributing again :)

Cool, looking forward to your next contribution. You can start with [good first issues](https://github.com/quarkusio/quarkus/issues?q=is%3Aopen+is%3Aissue+label%3A%22good+first+issue%22), hope there is something of interest for you there. Of course, you can also take a stab at any other issue you might be interested in.

Sanne:
Hey, just making sure there's no confusion here: are you all aware that the Agroal `initialSql` configuration is meant to be run _on each connection_ ?

This feature is meant to help with connection setup; for example setting a property on the connection after it's opened. 

It's great that we expose the configuration option, but I don't think we should suggest / encourage people to use it to create schemas, or otherwise run "setup" scripts as that would be confusing.

Could the example be changed to something more suitable?

gsmet:
@Sanne yeah, I fixed the Javadoc to make it clear. I had an hesitation about the example but as it only creates the schema if it doesn't exist, it doesn't seem that bad (obviously if the DDL are transactional).

Maybe a less controversial query could be used? @chaminnk WDYT?

chaminnk:
> @Sanne yeah, I fixed the Javadoc to make it clear. I had an hesitation about the example but as it only creates the schema if it doesn't exist, it doesn't seem that bad (obviously if the DDL are transactional).
> 
> Maybe a less controversial query could be used? @chaminnk WDYT?
Sure. We can use a different query for that. What do you suggest?

gsmet:
Maybe something setting the environment like `SET CLIENT_ENCODING TO 'UTF8'`?

Sanne:
> Maybe something setting the environment like `SET CLIENT_ENCODING TO 'UTF8'`?

yeah that's more appropriate. Also it's actually a useful suggestion.

The name of the property might also need a change? `initialSql` is not terribly ambiguous in the scope of Agroal, but in the context of application properties I suspect people might expect it doing something different.

(As a general rule, we don't _necessarily_ reuse the same property names )

Sanne:
I would suggest `on-new-connection-sql` , or something along those lines.

chaminnk:
> I would suggest `on-new-connection-sql` , or something along those lines.

@gsmet WDYT?

chaminnk:
Should I change the configuration key to ```on-new-connection-sql```?

gsmet:
Hey,

Sorry for the delay. I think I would go with `new-connection-sql` without the `on-`.

Sanne:
hi, just a heads up I'll need to do some rather large refactoring on this component in the following days.

I could wait on my side for this to be merged first, if it's going to be done soon? If not I'm happy to help finishing this - just let me know please :)

chaminnk:
Hi @Sanne , I'll do the changes and commit today if that is fine with you :)

Sanne:
@chaminnk that would be great, thanks!

chaminnk:
> Maybe something setting the environment like `SET CLIENT_ENCODING TO 'UTF8'`?

`SET CLIENT_ENCODING TO 'UTF8'` cannot be used with the H2 datasource used in the tests. As H2 uses `'UNICODE'` as the default client encoding, I think it cannot be used. Shall we use a different query for that?

gsmet:
It can't be used because the syntax is not supported? If not, can't we find an encoding that makes sense? Or something similar?

Sanne:
Maybe have it tested with a database for which it makes sense?

Or just have the documentation test a different use case than the integration test:

when I asked to change the example from a "schema generation" use case my comment was specifically regarding the documentation, just so to not have people misunderstand the use case. Yet of course to run the integration test a different use case could be acceptable.

chaminnk:
> It can't be used because the syntax is not supported? If not, can't we find an encoding that makes sense? Or something similar?
It seems like H2 doesn't allow changing the client encoding. Are there any suggestions other than setting client encoding that we can execute on each new connection?

chaminnk:
> when I asked to change the example from a "schema generation" use case my comment was specifically regarding the documentation, just so to not have people misunderstand the use case. Yet of course to run the integration test a different use case could be acceptable.

@Sanne shall I commit other changes without changing the current query used for testing? (which is `create schema if not exists schema_default`)

Sanne:
@chaminnk sure that sounds good

chaminnk:
@Sanne I've commited the changes

Sanne:
@gsmet I started a local build and rebase as there's several conflicts. Mind if I take over and send a new PR for you?

Sanne:
there's some issues with git history in this PR, I'll rebased it and send a new PR.

Sanne:
see #4425 as follow up.

PR #4118, opened 2019-09-19
Fix #4115 Make vertx-web handleFileUploads, uploadsDirectory, bodyLim…

### LINKED ISSUE #4115: Make vertx-web handleFileUploads, uploadsDirectory, bodyLimit. etc. configurable
The proposed options translate to the fields of `io.vertx.ext.web.handler.impl.BodyHandlerImpl`. I am primarily interested in the `uploadsDirectory` because when running integration tests, the default directory `file-uploads` is created under the root directory of the current Maven project instead of under `target`.

**Implementation ideas**

It seems to be easy to implement in `io.quarkus.vertx.web.runtime.VertxWebRecorder.addRoute(Router, Handler<RoutingContext>, Route, List<Handler<RoutingContext>>, HttpConfiguration)` where the `BodyHandler` is created. 
The only thing unclear to me is whether the new options should be added to a new `VertxWebRuntimeConfiguration` or if it is OK to add them to the existing `io.quarkus.vertx.http.runtime.HttpConfiguration`?

ppalaga:
…it. etc. configurable

@gastaldi @machi1990 you may want to review.

This solves only the vert.x-web use case. I am not sure whether anything needs to be done about Undertow?

### DIFF
--- a/extensions/vertx-web/runtime/src/main/java/io/quarkus/vertx/web/runtime/VertxWebRecorder.java
+++ b/extensions/vertx-web/runtime/src/main/java/io/quarkus/vertx/web/runtime/VertxWebRecorder.java
@@ -6,6 +6,7 @@ import java.util.function.Function;
 
 import io.quarkus.runtime.annotations.Recorder;
 import io.quarkus.runtime.configuration.MemorySize;
+import io.quarkus.vertx.http.runtime.BodyConfig;
 import io.quarkus.vertx.http.runtime.HttpConfiguration;
 import io.quarkus.vertx.http.runtime.RouterProducer;
 import io.quarkus.vertx.web.Route;
@@ -71,6 +72,12 @@ public class VertxWebRecorder {
                 if (maxBodySize.isPresent()) {
                     bodyHandler.setBodyLimit(maxBodySize.get().asLongValue());
                 }
+                final BodyConfig bodyConfig = httpConfiguration.body;
+                bodyHandler.setHandleFileUploads(bodyConfig.handleFileUploads);
+                bodyHandler.setUploadsDirectory(bodyConfig.uploadsDirectory);
+                bodyHandler.setDeleteUploadedFilesOnEnd(bodyConfig.deleteUploadedFilesOnEnd);
+                bodyHandler.setMergeFormAttributes(bodyConfig.mergeFormAttributes);
+                bodyHandler.setPreallocateBodyBuffer(bodyConfig.preallocateBodyBuffer);
 
                 route.handler(bodyHandler);
                 return route;
--- a/extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/BodyConfig.java
+++ b/extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/BodyConfig.java
@@ -1,0 +1,58 @@
+package io.quarkus.vertx.http.runtime;
+
+import io.quarkus.runtime.annotations.ConfigGroup;
+import io.quarkus.runtime.annotations.ConfigItem;
+
+/**
+ * Request body related settings
+ */
+@ConfigGroup
+public class BodyConfig {
+
+    /**
+     * Whether the files sent using {@code multipart/form-data} will be stored locally.
+     * <p>
+     * If {@code true}, they will be stored in {@code quarkus.http.body-handler.uploads-directory} and will be made
+     * available via {@code io.vertx.ext.web.RoutingContext.fileUploads()}. Otherwise, the the files sent using
+     * {@code multipart/form-data} will not be stored locally, and {@code io.vertx.ext.web.RoutingContext.fileUploads()}
+     * will always return an empty collection. Note that even with this option being set to {@code false}, the
+     * {@code multipart/form-data} requests will be accepted.
+     */
+    @ConfigItem(defaultValue = "true")
+    public boolean handleFileUploads;
+
+    /**
+     * The directory where the files sent using {@code multipart/form-data} should be stored.
+     * <p>
+     * Either an absolute path or a path relative to the current directory of the application process.
+     */
+    @ConfigItem(defaultValue = "file-uploads")
+    public String uploadsDirectory;
+
+    /**
+     * Whether the form attributes should be added to the request parameters.
+     * <p>
+     * If {@code true}, the form attributes will be added to the request parameters; otherwise the form parameters will
+     * not be added to the request parameters
+     */
+    @ConfigItem(defaultValue = "true")
+    public boolean mergeFormAttributes;
+
+    /**
+     * Whether the uploaded files should be removed after serving the request.
+     * <p>
+     * If {@code true} the uploaded files stored in {@code quarkus.http.body-handler.uploads-directory} will be removed
+     * after handling the request. Otherwise the files will be left there forever.
+     */
+    @ConfigItem(defaultValue = "false")
+    public boolean deleteUploadedFilesOnEnd;
+
+    /**
+     * Whether the body buffer should pre-allocated based on the {@code Content-Length} header value.
+     * <p>
+     * If {@code true} the body buffer is pre-allocated according to the size read from the {@code Content-Length}
+     * header. Otherwise the body buffer is pre-allocated to 1KB, and is resized dynamically
+     */
+    @ConfigItem(defaultValue = "false")
+    public boolean preallocateBodyBuffer;
+}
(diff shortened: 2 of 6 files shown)


gastaldi (approved):
LGTM, I'd just have used static imports in the test class to avoid `RestAssured.given()` and `Assertions.assertTrue`, but that's ok ;)

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/BodyHandlerConfig.java
```diff
@@ -0,0 +1,43 @@
...
+
+import io.quarkus.runtime.annotations.ConfigGroup;
+import io.quarkus.runtime.annotations.ConfigItem;
+
+@ConfigGroup
+public class BodyHandlerConfig {
+
+    /**
+     * If `true` the files sent using `multipart/form-data` will be stored locally under `uploadsDirectory` and will be
+     * made available via `io.vertx.ext.web.RoutingContext.fileUploads()`
```
gsmet: It would be nice if you could have a short sentence first (ending with a dot) then the details.

It will be better for automatic doc generation.

ppalaga: @gsmet would you please propose a solution? I am a bit unsure about your definition of short and about what you see as a good split into a 'short sentence' and 'details'. I'd be ready to add a period at the end to make the parser happy, but that's probably not what you'd like.

gsmet: Something like:
```
Whether the files sent using {@code multipart/form-data} will be stored locally.
<p>
If true, they will be stored in {@code quarkus..<insert the right property here>} and will be made available via {@code io.vertx.ext.web.RoutingContext.fileUploads()}.
```
This should be done consistently for all the doc here.

BTW, be careful: either you use plain old Javadoc or Asciidoc but if you use Asciidoc in the doc, you need to add the `@asciidoclet` tag.

ppalaga: Thanks, polished the JavaDoc in 0309b80 following your advice.

machi1990 (approved):
LGTM.

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/BodyHandlerConfig.java
```diff
@@ -0,0 +1,55 @@
...
+    public boolean mergeFormAttributes;
+
+    /**
+     * Whether the uploaded files should be removed after serving the request.
+     * <p>
+     * If {@code true} the uploaded files stored in {@code quarkus.http.body-handler.uploads-directory} will be removed
+     * after handling the request. Otherwise the files will be left there forever.
+     */
+    @ConfigItem(defaultValue = "false")
+    public boolean deleteUploadedFilesOnEnd;
```
cescoffier: This one needs to be tested carefully... 
I had issues in a container where it was not behavior properly (not a big deal as the container was getting shutdown)

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/BodyHandlerConfig.java
```diff
@@ -0,0 +1,55 @@
...
+public class BodyHandlerConfig {
+
+    /**
+     * Whether the files sent using {@code multipart/form-data} will be stored locally.
+     * <p>
+     * If {@code true}, they will be stored in {@code quarkus.http.body-handler.uploads-directory} and will be made
+     * available via {@code io.vertx.ext.web.RoutingContext.fileUploads()}.
+     */
+    @ConfigItem(defaultValue = "true")
+    public boolean handleFileUploads;
```
cescoffier: Several config properties uses "enable", should this one follow this convention. Like `enableFileUploads`?

extensions/vertx-web/runtime/src/main/java/io/quarkus/vertx/web/runtime/VertxWebRecorder.java
```diff
@@ -84,6 +85,12 @@ public void addAdditionalRoutes(RuntimeValue<Router> routerRuntimeValue, Map<Str
         if (maxBodySize.isPresent()) {
             bodyHandler.setBodyLimit(maxBodySize.get().asLongValue());
         }
+        final BodyHandlerConfig bodyHandlerConfig = httpConfiguration.bodyHandler;
```
cescoffier: Should it be just "body"?
For example, when we will have the equivalent for the `cacheHandler`, or for `staticFileHandler`, we will _hide_ the handler part. WDYT?

cescoffier (changes requested):
Not really requesting changes, just asking questions about our favorite topics: naming...

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/BodyHandlerConfig.java
ppalaga: The DisabledUploadsTest https://github.com/quarkusio/quarkus/pull/4118/files#diff-c251801edc089c62bf58421345eefb5dR36 is enough, isn't it?

ppalaga: `DisabledUploadsTest` does something else. Sorry for the confusion.

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/BodyHandlerConfig.java
ppalaga: Not sure. `enableFileUploads = false` might sound like it is not possible to upload files to the given server. But even with `enableFileUploads = false`, the multipart requests are accepted. It only causes that the files are not stored locally and the `RoutingContext.fileUploads()` returns an empty collection.

extensions/vertx-web/runtime/src/main/java/io/quarkus/vertx/web/runtime/VertxWebRecorder.java
ppalaga: +1 let's remove the handler and also the reference to Wert.x web in the JavaDoc.

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/BodyHandlerConfig.java
ppalaga: Let me add a sentence about the negative behavior to the JavaDoc

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/BodyHandlerConfig.java
ppalaga: Added `DeleteUploadedFilesOnEndTest` in 950b2ad4e552f342f0a4adba99fe99a4ac5bc8de

extensions/vertx-web/runtime/src/main/java/io/quarkus/vertx/web/runtime/VertxWebRecorder.java
ppalaga: Addressed in 950b2ad4e552f342f0a4adba99fe99a4ac5bc8de

ppalaga:
Accepted all proposals of @cescoffier in 950b2ad4e552f342f0a4adba99fe99a4ac5bc8de except for the one to rename `handleFileUploads` to `enableFileUploads`.

gsmet:
This PR needs a rebase.

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/BodyHandlerConfig.java
cescoffier: +1

cescoffier:
@ppalaga do you want me to fix the conflict?

cescoffier (approved):

ppalaga:
86a72af rebased on top of the current master.

ppalaga:
The windows build fails in the MongoDB extension which seems to be unrelated to the current PR.

ppalaga:
3e9bc1a rebased again to see whether the MongoDB failure on windows persists.

ppalaga:
Is this good to merge now?

PR #4129, opened 2019-09-20
Unify how Arc attemps to load classes from Index, add logging for missing classes

### LINKED ISSUE #4080: Arc - add better handling of non-indexed classes we encounter
Currently, Arc in many places tries to search a class from index and while we check for `null`, we don't really log anything and we mostly silently proceed. This can lead to some puzzling user output, as was pointed out on another PR https://github.com/quarkusio/quarkus/pull/4060#discussion_r325540679.

The likelihood of this happening is small as `io.quarkus.arc.processor.BeanArchives.IndexWrapper` tries to index additional classes, so in most cases, all classes should be there.
It would still be worth it to add a single util class which would be used throughout Arc - this class would then attempt loading from Jandex index. Should it fail, we can choose to either log something, or straight up fail with an exception.

manovotn:
…sing classes.

Fixes #4080 

@mkouba looking at what we had in Arc now, we sometimes blow up, sometimes silently continue and at other times we logged something. So in the end I added a util class that does _info_ logging but still allows to return `null` as in some cases blowing up might be a tad bit too harsh?

### DIFF
--- a/independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/BeanDeployment.java
+++ b/independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/BeanDeployment.java
@@ -1,5 +1,7 @@
 package io.quarkus.arc.processor;
 
+import static io.quarkus.arc.processor.IndexClassLookupUtils.getClassByName;
+
 import io.quarkus.arc.processor.BeanDeploymentValidator.ValidationContext;
 import io.quarkus.arc.processor.BeanProcessor.BuildContextImpl;
 import io.quarkus.arc.processor.BeanRegistrar.RegistrationContext;
@@ -123,12 +125,9 @@ public class BeanDeployment {
         this.interceptorBindings = findInterceptorBindings(index);
         for (InterceptorBindingRegistrar registrar : bindingRegistrars) {
             for (DotName dotName : registrar.registerAdditionalBindings()) {
-                ClassInfo classInfo = index.getClassByName(dotName);
+                ClassInfo classInfo = getClassByName(index, dotName);
                 if (classInfo != null) {
                     interceptorBindings.put(dotName, classInfo);
-                } else {
-                    throw new RuntimeException("An attempt was made to turn class " + dotName + " into an interceptor " +
-                            "binding but the class was not found in Jandex index.");
                 }
             }
         }
@@ -453,7 +452,7 @@ public class BeanDeployment {
         }
         for (AnnotationInstance stereotype : stereotypeAnnotations) {
             final DotName stereotypeName = stereotype.target().asClass().name();
-            ClassInfo stereotypeClass = index.getClassByName(stereotypeName);
+            ClassInfo stereotypeClass = getClassByName(index, stereotypeName);
             if (stereotypeClass != null) {
 
                 boolean isAlternative = false;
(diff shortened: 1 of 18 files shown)


independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/IndexClassLookupUtils.java
```diff
@@ -0,0 +1,41 @@
+package io.quarkus.arc.processor;
+
+import org.jboss.jandex.ClassInfo;
+import org.jboss.jandex.DotName;
+import org.jboss.jandex.IndexView;
+import org.jboss.logging.Logger;
+
+public class IndexClassLookupUtils {
```
mkouba: This class should be probably package-private and final.

independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/IndexClassLookupUtils.java
```diff
@@ -0,0 +1,41 @@
...
+
+    private static final Logger LOGGER = Logger.getLogger(IndexClassLookupUtils.class);
+
+    private IndexClassLookupUtils() {
+    }
+
+    public static ClassInfo getClassByName(IndexView index, DotName dotName) {
```
mkouba: Package-private as well...

independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/BeanDeployment.java
```diff
@@ -123,12 +125,9 @@
         this.interceptorBindings = findInterceptorBindings(index);
         for (InterceptorBindingRegistrar registrar : bindingRegistrars) {
             for (DotName dotName : registrar.registerAdditionalBindings()) {
-                ClassInfo classInfo = index.getClassByName(dotName);
+                ClassInfo classInfo = getClassByName(index, dotName);
```
mkouba: Shouldn't we use the logging variant here? I mean we threw runtime exception before and now silently ignore the problem?

independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/IndexClassLookupUtils.java
```diff
@@ -0,0 +1,41 @@
...
+        if (dotName == null) {
+            throw new IllegalArgumentException("Cannot lookup class, provided DotName was null.");
+        }
+        if (index == null) {
+            throw new IllegalArgumentException("Cannot lookup class, provided Jandex Index was null.");
+        }
+        ClassInfo info = index.getClassByName(dotName);
+        if (info == null && withLogging) {
+            // class not in index, log warning as this may cause the application to blow up or behave weirdly
+            LOGGER.info("Class for name: " + dotName + " was not found in Jandex index. Please ensure the class " +
```
mkouba: Shouldn't we use the WARN level instead?

mkouba (changes requested):

independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/IndexClassLookupUtils.java
manovotn: That's what I had there originally, but since some pieces of code just choose to "silently" proceed now, WARN felt a bit too strong. But  I wouldn't mind...

independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/BeanDeployment.java
manovotn: This is the logging variant.

But yes, we threw exception before, but since we will now have logging, in this case I reconsidered and thought that should be enough instead of hard fail. WDYT?

mkouba: Ah sorry, these Fridays...

independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/IndexClassLookupUtils.java
mkouba: Hm, that's a tricky question. Let's keep INFO for now...

manovotn: I mean, even `INFO` was noticeable enough for me to fix all those `arc/processor` tests ;)

manovotn: I mean, even INFO made me notice and fix those `arc/processor` tests which were having unindexed classes ;-)

manovotn:
@mkouba addressed all your comments

mkouba (approved):

mkouba:
Looks good now. Let's wait for CI ;-).

PR #4132, opened 2019-09-20
issue #4073 - Display profile and whether Live Coding is enabled or not at startup

### LINKED ISSUE #4073: Display profile (and mode?) we run with in the logs at startup time
Is there any disadvantage to show the profile Quarkus has been started with in the log?
It's something I miss when I demo profiles.

CC @FroMage for advice

emmanuelbernard:
Fixes: #4073

### DIFF
--- a/core/runtime/src/main/java/io/quarkus/runtime/Timing.java
+++ b/core/runtime/src/main/java/io/quarkus/runtime/Timing.java
@@ -50,12 +50,13 @@ public class Timing {
         bootStartTime = System.nanoTime();
     }
 
-    public static void printStartupTime(String version, String features) {
+    public static void printStartupTime(String version, String features, String profile, boolean liveCoding) {
         final long bootTimeNanoSeconds = System.nanoTime() - bootStartTime;
         final Logger logger = Logger.getLogger("io.quarkus");
         //Use a BigDecimal so we can render in seconds with 3 digits precision, as requested:
         final BigDecimal secondsRepresentation = convertToBigDecimalSeconds(bootTimeNanoSeconds);
         logger.infof("Quarkus %s started in %ss. %s", version, secondsRepresentation, httpServerInfo);
+        logger.infof("Profile %s activated. %s", profile, liveCoding ? "Live Coding activated." : "");
         logger.infof("Installed features: [%s]", features);
         bootStartTime = -1;
     }
--- a/core/deployment/src/main/java/io/quarkus/deployment/steps/MainClassBuildStep.java
+++ b/core/deployment/src/main/java/io/quarkus/deployment/steps/MainClassBuildStep.java
@@ -21,6 +21,7 @@ import io.quarkus.deployment.builditem.BytecodeRecorderObjectLoaderBuildItem;
 import io.quarkus.deployment.builditem.ClassOutputBuildItem;
 import io.quarkus.deployment.builditem.FeatureBuildItem;
 import io.quarkus.deployment.builditem.JavaLibraryPathAdditionalPathBuildItem;
+import io.quarkus.deployment.builditem.LaunchModeBuildItem;
 import io.quarkus.deployment.builditem.MainBytecodeRecorderBuildItem;
 import io.quarkus.deployment.builditem.MainClassBuildItem;
 import io.quarkus.deployment.builditem.ObjectSubstitutionBuildItem;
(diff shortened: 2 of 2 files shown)


machi1990:
Does this issue fix https://github.com/quarkusio/quarkus/issues/4073 ?

core/runtime/src/main/java/io/quarkus/runtime/Timing.java
```diff
@@ -50,12 +50,13 @@ public static void restart() {
...
     }
 
-    public static void printStartupTime(String version, String features) {
+    public static void printStartupTime(String version, String features, String profile, boolean liveCoding) {
         final long bootTimeNanoSeconds = System.nanoTime() - bootStartTime;
         final Logger logger = Logger.getLogger("io.quarkus");
         //Use a BigDecimal so we can render in seconds with 3 digits precision, as requested:
         final BigDecimal secondsRepresentation = convertToBigDecimalSeconds(bootTimeNanoSeconds);
         logger.infof("Quarkus %s started in %ss. %s", version, secondsRepresentation, httpServerInfo);
+        logger.infof("Profile %s activated. %s", profile, liveCoding ? "Live Coding activated." : "");
```
gastaldi: What is "Live Coding"? Is that the dev mode when `quarkus:dev` is used?

loicmathieu: It should be "Live Reload" : https://github.com/quarkusio/quarkus/blob/master/docs/src/main/asciidoc/0-glossary.adoc

gastaldi:
```suggestion
        logger.infof("Profile %s activated. %s", profile, liveReload ? "Live Reload activated." : "");
```

core/runtime/src/main/java/io/quarkus/runtime/Timing.java
```diff
@@ -50,12 +50,13 @@ public static void restart() {
         bootStartTime = System.nanoTime();
     }
 
-    public static void printStartupTime(String version, String features) {
+    public static void printStartupTime(String version, String features, String profile, boolean liveCoding) {
```
gastaldi:
```suggestion
    public static void printStartupTime(String version, String features, String profile, boolean liveReload) {
```

core/runtime/src/main/java/io/quarkus/runtime/Timing.java
emmanuelbernard: We switched to Live Coding as a term a while back. I guess we have not propagated it. But it was for the general experience and maybe live reload is better in this case as it's specifically the live reload option. Not sure.

loicmathieu: IMHO "Live coding" refers to performing programmation in front of people (like in conferences).
https://en.wikipedia.org/wiki/Live_coding

Here we have a automatic reload (or hot reload like in hot code replacement)

gastaldi: >Here we have a automatic reload (or hot reload like in hot code replacement)

Right, my only concern is that it may confuse users since the reload happens when you actually perform a request to the application and not automatically.

emmanuelbernard: For the sake of not going over for ever. I'd propose to push that PR in and if people really hate Live Coding, we can always easily rename later. It's not like it's an API.

loicmathieu: > Well @loicmathieu The Wikipedia Live Coding definition is pretty spot on for us
> 
> > Live coding, [...], makes programming an integral part of the running program.

Yeah, I add it on purpose, it says both usage, I particularly love the reference to "performing art"
> It is most prominent as a performing arts form

gsmet:
@gastaldi @loicmathieu now that we agree to keep the name as is, one dumb question: has any of you tested it?

It would be nice to have someone test it.

emmanuelbernard:
I did test in JVM dev mode, in JVM prod mode and in native image. But yes would be good for someone else to test.

gsmet:
I'll do it if noone beats me to it (please, please, do!). I'd like to have this one in 0.23.0.

loicmathieu:
@gsmet better use your time to check my PR :)
I'll test this one now

gsmet:
Have you properly cleaned up your project? I wouldn't expect an issue here.

You can try rebasing Emmanuel's PR and see how it goes. If you still have
the issue, we might have a regression.

loicmathieu:
@gsmet after rebasing on master it works.
Tested both dev and prod mode.

gsmet (approved):

PR #4135, opened 2019-09-20
fix(dev-mode): live reload failing with Zip archives generated by code.quarkus.io due to time zone issue.

### LINKED ISSUE #4064: Live reload failing with Zip archives generated by code.quarkus.io due to time zone issues
on 0.22

1. Start new project with `code.quarkus.io` with resteasy,resteasy-jsonb,panache,postgresql,openapi,validator
2. go in live reload
3. add TodoResource

```
@Path("/api")
@Consumes(MediaType.APPLICATION_JSON)
public class TodoResource {
    @GET
    public List<Todo> getAll() {
        return Todo.listAll();
    }
}
```

4. add Todo

```
@Entity
public class Todo extends PanacheEntity {
    public String title;
}
```

5. refresh http://localhost:8080/api
6. see Hibernate not configured error
7. add in application.properties

```
# set DB
quarkus.datasource.url=jdbc:postgresql:quarkus-demo
quarkus.datasource.driver=org.postgresql.Driver
quarkus.datasource.username=quarkus-demo
quarkus.datasource.password=quarkus-demo

# Hibernate details
quarkus.hibernate-orm.database.generation=drop-and-create
quarkus.hibernate-orm.log.sql=true
```

8. refresh http://localhost:8080/api
9. You still see the error where it should not and the logs show that the hot reload detects the wrong file change (I did not update `ExampleResource.java` but `application.properties`

```
2019-09-17 07:07:00,034 INFO  [io.qua.dev] (executor-thread-1) Changed source files detected, recompiling [/Users/emmanuel/Code/quarkus/demos/CodeOne/hibernate-app/src/main/java/io/quarkus/hibernate/ExampleResource.java]
2019-09-17 07:07:00,247 INFO  [io.qua.dep.QuarkusAugmentor] (executor-thread-1) Beginning quarkus augmentation
```

I would love for this exact scenario to be tested in unit tests because it has broken numnerous times in the past.

machi1990:
This fixes the issue by implementing a solution proposed by this comment
https://github.com/quarkusio/quarkus/issues/4064#issuecomment-533133783 and just check if last
recorded time and current modification time are different.

Fixes #4064 

/cc @dmlloyd

### DIFF
--- a/core/devmode/src/main/java/io/quarkus/dev/RuntimeUpdatesProcessor.java
+++ b/core/devmode/src/main/java/io/quarkus/dev/RuntimeUpdatesProcessor.java
@@ -31,22 +31,33 @@ import io.quarkus.runtime.Timing;
 
 public class RuntimeUpdatesProcessor implements HotReplacementContext {
     private static final String CLASS_EXTENSION = ".class";
+    private static final Logger log = Logger.getLogger(RuntimeUpdatesProcessor.class.getPackage().getName());
+
     private final DevModeContext context;
     private final ClassLoaderCompiler compiler;
-    private volatile long lastChange = System.currentTimeMillis();
 
     // file path -> isRestartNeeded
     private volatile Map<String, Boolean> watchedFilePaths = Collections.emptyMap();
-    private final Map<Path, Long> watchedFileTimestamps = new ConcurrentHashMap<>();
-    private final Map<Path, Path> classFilePathToSourceFilePath = new ConcurrentHashMap<>();
+
     /**
-     * resources that appear in both src and target, these will be removed if the src resource subsequently disappears
-     *
+     * A first scan is considered done when we have visited all modules at least once.
+     * This is useful in two ways.
+     * - To make sure that source time stamps have been recorded at least once
+     * - To avoid re-compiling on first run by ignoring all first time changes detected by
+     * {@link RuntimeUpdatesProcessor#checkIfFileModified(Path, Map, boolean)} during the first scan.
+     */
+    private volatile boolean firstScanDone = false;
+
+    private final Map<Path, Long> sourceFileTimestamps = new ConcurrentHashMap<>();
+    private final Map<Path, Long> watchedFileTimestamps = new ConcurrentHashMap<>();
+    private final Map<Path, Long> classFileChangeTimeStamps = new ConcurrentHashMap<>();
+    private final Map<Path, Path> classFilePathToSourceFilePath = new ConcurrentHashMap<>();
+
+    /**
+     * Resources that appear in both src and target, these will be removed if the src resource subsequently disappears.
      * This set contains the paths in the target dir
      */
     private final Set<Path> correspondingResources = Collections.newSetFromMap(new ConcurrentHashMap<>());
-
-    private static final Logger log = Logger.getLogger(RuntimeUpdatesProcessor.class.getPackage().getName());
     private final List<Runnable> preScanSteps = new CopyOnWriteArrayList<>();
     private final List<Consumer<Set<String>>> noRestartChangesConsumers = new CopyOnWriteArrayList<>();
     private final List<HotReplacementSetup> hotReplacementSetup = new ArrayList<>();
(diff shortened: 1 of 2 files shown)


core/devmode/src/main/java/io/quarkus/dev/RuntimeUpdatesProcessor.java
```diff
@@ -31,22 +31,33 @@
...
+     * This is useful in two ways.
+     * - To make sure that source time stamps have been recorded at least once
+     * - To avoid re-compiling on first run by ignoring all first time changes detected by
+     * {@link RuntimeUpdatesProcessor#checkIfFileModified(Path, Map)} during the first scan.
+     */
+    private volatile boolean firstScanDone = false;
+
+    private final Map<Path, Long> sourceFileTimestamps = new ConcurrentHashMap<>();
     private final Map<Path, Long> watchedFileTimestamps = new ConcurrentHashMap<>();
+    private final Map<Path, Long> classFileChangeTimeStamps = new ConcurrentHashMap<>();
```
gsmet: It's minor but it should be `Timestamps` not `TimeStamps`.

gsmet (changes requested):

core/devmode/src/main/java/io/quarkus/dev/RuntimeUpdatesProcessor.java
machi1990: Good catch.

core/devmode/src/main/java/io/quarkus/dev/RuntimeUpdatesProcessor.java
```diff
@@ -356,10 +372,34 @@ public void accept(Path path) {
...
+            final long lastModificationTime = Files.getLastModifiedTime(path).toMillis();
+            final Long lastRecordedChange = pathModificationTimes.get(path);
+
+            if (lastRecordedChange == null) {
+                pathModificationTimes.put(path, lastModificationTime);
+                if (firstScanDone) {
+                    return true;
+                }
+
+                return false; // ignore changes to avoid recompilation and re-start
```
stuartwdouglas: I think this break the workflow of: Start Quarkus -> Make Change -> Make first request. This change will not be picked up because of the 'first scan' logic.

I don't really have a good solution for this, but I think that the above scenario will be much more common that the timestamps being wrong (once the issue is fixed in the code.quarkus.io generator).

machi1990: > I think this break the workflow of: Start Quarkus -> Make Change -> Make first request. This change will not be picked up because of the 'first scan' logic.
> 

Thanks Stuart for the review. Sure, we'll need two requests to make the changes in this workflow being taken into account. 

> I don't really have a good solution for this, but I think that the above scenario will be much more common that the timestamps being wrong (once the issue is fixed in the code.quarkus.io generator).

I was thinking of recording initial files timestamp in constructor. Will this work? WDYT?

gsmet: But we would need to wait for the thread to end before serving the first request, I suppose?

machi1990: I took a second look and this by trying to create a test case https://github.com/quarkusio/quarkus/pull/4135/files#diff-c0bdb27a09deae0802040076476bad74R250 for the scenario described above and this case seems already handled by the current code. 

I suppose with this line https://github.com/quarkusio/quarkus/blob/master/core/devmode/src/main/java/io/quarkus/dev/DevModeMain.java#L90 we'll record all the timestamps once a Quarkus app is started. All the modifications picked since the start and before the first call will be detected thus causing a re-compile since at this point we have done a first full scan and we have the timestamps recorded.

stuartwdouglas: Nice, in that case I think this is good to go.

stuartwdouglas (approved):

gsmet (approved):

gsmet:
Even if it's a bugfix, I'm pushing this to 0.24.0 only as I prefer we have some feedback before pushing it into a release.

gsmet:
Thanks @machi1990 , merging!

machi1990:
> Even if it's a bugfix, I'm pushing this to 0.24.0 only as I prefer we have some feedback before pushing it into a release.

Yeah, works for me. Thanks.

PR #4138, opened 2019-09-20
Bump kubernetes-client to 4.6.0

gastaldi:


geoand (approved):
Thanks @gastaldi!

What new does 4.6.x bring?

gastaldi:
@geoand according to https://github.com/fabric8io/kubernetes-client/releases/tag/v4.6.0, support for ControllerRevision,Template Instance and SharedInformer

geoand:
Thanks @gastaldi.

Seems like it adds things that we all use everyday 😂

gastaldi:
Ah and the Kubernetes Model was updated, I think that's why it's worth upgrading :wink:

geoand:
I just now realized what the SharedInformer is. That is actually super interesting!

Issue #3539, opened 2019-08-15
QuarkusTestExtension does not support -ntp in Maven 3.6.1


jglick:
**Describe the bug**

If I run Maven commands using `-ntp`, an option added in Maven 3.6.1, I get an error during Quarkus tests.

**Expected behavior**

Just runs.

**Actual behavior**

The test crashes with

```
org.junit.jupiter.api.extension.TestInstantiationException: TestInstanceFactory [io.quarkus.test.junit.QuarkusTestExtension] failed to instantiate test class […]
	at org.junit.jupiter.engine.descriptor.ClassBasedTestDescriptor.invokeTestInstanceFactory(ClassBasedTestDescriptor.java:298)
	at org.junit.jupiter.engine.descriptor.ClassBasedTestDescriptor.instantiateTestClass(ClassBasedTestDescriptor.java:273)
	at org.junit.jupiter.engine.descriptor.ClassTestDescriptor.instantiateTestClass(ClassTestDescriptor.java:70)
	at org.junit.jupiter.engine.descriptor.ClassBasedTestDescriptor.instantiateAndPostProcessTestInstance(ClassBasedTestDescriptor.java:256)
	at org.junit.jupiter.engine.descriptor.ClassBasedTestDescriptor.lambda$testInstancesProvider$2(ClassBasedTestDescriptor.java:250)
	at java.util.Optional.orElseGet(Optional.java:267)
	at org.junit.jupiter.engine.descriptor.ClassBasedTestDescriptor.lambda$testInstancesProvider$3(ClassBasedTestDescriptor.java:249)
	at org.junit.jupiter.engine.execution.TestInstancesProvider.getTestInstances(TestInstancesProvider.java:29)
	at org.junit.jupiter.engine.descriptor.TestMethodTestDescriptor.lambda$prepare$0(TestMethodTestDescriptor.java:106)
	at org.junit.platform.engine.support.hierarchical.ThrowableCollector.execute(ThrowableCollector.java:73)
	at org.junit.jupiter.engine.descriptor.TestMethodTestDescriptor.prepare(TestMethodTestDescriptor.java:105)
	at org.junit.jupiter.engine.descriptor.TestMethodTestDescriptor.prepare(TestMethodTestDescriptor.java:69)
	at org.junit.platform.engine.support.hierarchical.NodeTestTask.lambda$prepare$1(NodeTestTask.java:107)
[...]

PR #4139, opened 2019-09-20
Proposed Maven version is 3.6.2

gastaldi:
This is the proposed Maven version used by the Maven Wrapper when a new Quarkus project is created using the Maven plugin

### DIFF
--- a/build-parent/pom.xml
+++ b/build-parent/pom.xml
@@ -48,7 +48,7 @@
         <supported-maven-versions>[3.5.3,)</supported-maven-versions>
 
         <!-- These 2 properties are used by CreateProjectMojo to add the Maven Wrapper -->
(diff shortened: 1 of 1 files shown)


gsmet:
IIRC, we had some issues with this upgrade and our Maven plugin. @aloubyansky have the problems been solved?

aloubyansky (approved):
Well, the one we had with 3.6.2 was using `${project.groupId}` instead of `io.quarkus`, which was a good thing to fix. There is however #3539, which basically means there could be arguments that break testing. But that issue was reported for 3.6.1, which we apparently have been recommending. From that perspective, it's kind of not new. Whether it qualifies for a "recommended" version though, remains a question, I guess.
I'd probably merge it though.

geoand:
Should we merge this?

PR #2324, opened 2019-05-02
Adds test random port

lordofthejars:
Fixes #2269

PR #4128, opened 2019-09-20
Fail deployment when unable to create the vertx httpServer

geoand:
Fixes: #4124

Issue #3938, opened 2019-09-09
KubernetesMockServer tests sensitive to runOrder/reuseForks


jglick:
**Describe the bug**

The `KubernetesMockServer` test resource seems to carry over mocks between `@QuarkusTest` classes in the same JVM. Since Surefire defaults to `reuseForks=true` and `runOrder=filesystem`, this means that in a project with multiple test suites using the mock server, even if individual tests (`mvn test -Dtest=SomeTest`) reliably pass, the whole suite (`mvn test`) may fail with cryptic errors.

**Expected behavior**

Each `@Test` should get an independent mock context.

**Actual behavior**

Sometimes `expect()` calls seem to carry over between test classes, resulting in random failures.

**To Reproduce**

Do not currently have a public demo project, but could probably put one together. Noticed after introducing a new `@QuarkusTest` which caused an existing one to start failing in a CI environment, but not locally. Added

```xml
<runOrder>alphabetical</runOrder>
```

and tests passed (locally); with

```xml
<runOrder>reversealphabetical</runOrder>
```

they failed as on CI. Adding

```xml
<reuseForks>false</reuseForks>
```

seems to have fixed the issue.

**Configuration**

N/A

**Screenshots**

N/A

**Environment (please complete the following information):**
 - Java etc.: `quay.io/quarkus/centos-quarkus-maven:19.2.0`
 - Quarkus version or git rev: 0.21.2

**Additional context**

If the test framework itself cannot easily be fixed, at least the Maven archetype should preconfigure

```xml
<forkCount>1</forkCount>
<reuseForks>false</reuseForks>
<runOrder>alphabetical</runOrder>
```

for both `maven-surefire-plugin` and `maven-failsafe-plugin`. (`forkCount` is advisable due to #2269: Quarkus does not support running tests in parallel.)

geoand:
@iocanel how can one reset the Mock server? I searched a little but I didn't come up with anything useful

jglick:
Even with `reuseForks=false`, test methods within a class seem to reuse the mock server, which is pretty awkward: to create a genuinely independent test you need to create a fresh class. (Or maybe a nested class?) Of course there is a bit of overhead to start the app separately for each test class, but I think this is worth the reduced chance of wasting time tracking down interference between tests.

geoand:
I agree 100%.

@iocanel do we have a way to reset the mock server? If so, then I can patch the Quarkus parts to call whatever is needed to reset it between tests

[...]

PR #4140, opened 2019-09-20
Support random test port

### LINKED ISSUE #2269: Random port configuration of port in tests
Currently, there is one property called `quarkus.http.test-port` where you can set the port number where to start the Quarkus server, but we should provide a special char for setting a random private port (49152 through 65535)

I purpose something like `quarkus.http.test-port=$`

This is important for allowing the system to execute tests in parallel.

If agrees I can send a PR.

gastaldi:
By supporting random ports during test execution, this PR should allow running HTTP tests in parallel (as long as they run in different JVMs)

Fixes #2269 and closes #2324

### DIFF
--- a/test-framework/junit5/src/main/java/io/quarkus/test/junit/QuarkusTestExtension.java
+++ b/test-framework/junit5/src/main/java/io/quarkus/test/junit/QuarkusTestExtension.java
@@ -4,7 +4,6 @@ import static io.quarkus.test.common.PathTestHelper.getAppClassLocation;
 import static io.quarkus.test.common.PathTestHelper.getTestClassesLocation;
 
 import java.io.Closeable;
-import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.net.MalformedURLException;
@@ -112,9 +111,7 @@ public class QuarkusTestExtension
                     public void writeClass(boolean applicationClass, String className, byte[] data) throws IOException {
                         Path location = testWiringClassesDir.resolve(className.replace('.', '/') + ".class");
                         Files.createDirectories(location.getParent());
-                        try (FileOutputStream out = new FileOutputStream(location.toFile())) {
-                            out.write(data);
-                        }
+                        Files.write(location, data);
                         shutdownTasks.add(new DeleteRunnable(location));
                     }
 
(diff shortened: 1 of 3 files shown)


gastaldi:
@gsmet it would be nice if this PR could make it to 0.23.0. I'd appreciate if you (or someone else) could review

geoand:
This is going to conflict with #4128 (which I have already had to rebase to fix a different conflict 🤪)

gsmet:
Is there a specific reason? It looks like a new feature to me so I would keep it for 0.24.0.

gastaldi:
@gsmet no strong reason except early feedback from users, I'm fine with that in 0.24.0 ;)

gastaldi:
Rebased, thanks for the warning @geoand :+1:

jglick:
> as long as they run in different JVMs

FYI: #3938

test-framework/junit5/src/main/java/io/quarkus/test/junit/QuarkusTestExtension.java
```diff
@@ -192,9 +187,7 @@ protected ClassLoader getClassLoader() {
 
                                 Path location = testWiringClassesDir.resolve(resourceName);
                                 Files.createDirectories(location.getParent());
-                                try (FileOutputStream out = new FileOutputStream(location.toFile())) {
-                                    out.write(cw.toByteArray());
-                                }
+                                Files.write(location, cw.toByteArray());
```
jglick: Are the changes to this file related?

gastaldi: No, not at all, it is just a minor cleanup

jmartisk (commented):
I just tried running the MP TCKs with it and noticed that setting a `-Dquarkus.http.test-port` breaks a lot of tests in MP OpenApi tests. I'll have a look what can be done about that (it's not caused by this PR).
Perhaps an additional CI check with a random port enabled would be a good thing to introduce.
Anyway very nice, this will be very useful

geoand:
I like this! Let's see what other think

stuartwdouglas (approved):
LGTM, the only caveat is that this will not work with native mode tests.

The only way I can see to make it work is to parse the log output to read the http port, but that kinda sucks.

Issue #4126, opened 2019-09-20
"Failed to inject extension deployment dependencies: zip END header not found" error should mention the corrupt jar in my local maven repo


ge0ffrey:
I managed to sync maven in intellij at the same time as running "mvn" on the command line.
So one of my jars got corrupted. No idea which one (I am still trying to figure out which one).

When I run "mvn quarkus:dev" now, it doesn't tell me which one:

```
[INFO] --- quarkus-maven-plugin:0.22.0:dev (default-cli) @ optaplanner-quarkus-example ---
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
...
[ERROR] Failed to execute goal io.quarkus:quarkus-maven-plugin:0.22.0:dev (default-cli) on project optaplanner-quarkus-example: Failed to run: Failed to resolve Quarkus application model: Failed to inject extension deployment dependencies for org.optaplanner:optaplanner-quarkus-example:jar:7.28.0-SNAPSHOT: zip END header not found -> [Help 1]
...
```

**The Quarkus exception should include the name of the broken zip file (= corrupted jar file in my maven repo).** This messes up the getting started experience.

Stacktraces (mvn -e):
https://gist.github.com/ge0ffrey/1c1483b8508790aefeeda29580eeb20c

ge0ffrey:
**WORKAROUND**

1. Run "mvn quarkus:dev -e" and put a breakpoint on the ZipFileSsystem line that throws the error.
2. Run `mvnDebug quarkus:dev -Ddebug` (note that `mvn quarkus:dev -Ddebug` is not enough).
3. Connect with a remote debugger on port 8000 (=non standard port)
4. Wait for breakpoint, in Frames, look at the "this" var of 1 call up in the stacktrace. It will mention the corrupted jar name.

gsmet:
Looks like a good opportunity for a first contribution :).

jaikiran:
I have created a PR with a potential fix for this https://github.com/quarkusio/quarkus/pull/4145

PR #4145, opened 2019-09-22
[issue#4126] - include the path of the file which triggered an IOException

jaikiran:
The commit here improves the `IOException` message to include the URI/Path which caused a zip operation to fail, through `ZipUtils`. This should address the issue noted in https://github.com/quarkusio/quarkus/issues/4126.

The commit also includes a test case to reproduce and verify the fix.

P.S: This currently catches a `java.util.zip.ZipError` which shouldn't have been thrown in first place by Java (since it's an undeclared error). As noted in the code comments, that's now been fixed in Java 9 and we can remove that `ZipError` catch once the minimal Java runtime version for Quarkus is Java 9.

### DIFF
--- a/independent-projects/bootstrap/core/src/main/java/io/quarkus/bootstrap/util/ZipUtils.java
+++ b/independent-projects/bootstrap/core/src/main/java/io/quarkus/bootstrap/util/ZipUtils.java
@@ -17,6 +17,7 @@ import java.nio.file.attribute.BasicFileAttributes;
 import java.util.Collections;
 import java.util.EnumSet;
 import java.util.Map;
+import java.util.zip.ZipError;
 
 /**
  *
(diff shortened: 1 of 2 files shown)


geoand:
Hi @jaikaran,

Thanks a lot for the contribution!

Unfortunately the test you added fails in Windows as can be seen in CI. Could you please take a look?

jaikiran:
I'll take a look. I don't have a Windows OS locally to test this, so have to rely on the PR runs to verify the test. I've pushed some change and will wait for the jobs to complete, to see if it helps.

geoand:
It might help to look at the commit log and what changes have been made in other commits to address other Windows issues (there have been a few similar issues in the past)

geoand (approved):
LGTM, thanks @jaikiran! 
I'll let @aloubyansky have the final say in this :)

aloubyansky (approved):
Thanks @jaikiran!

PR #4146, opened 2019-09-22
Add a priority attribute to FilterBuildItem to sort the filters

### LINKED ISSUE #4113: Properly manage the order of Vert.x filters
Right now, the Vert.x filters are added in a totally unordered way.

This definitely won't work: we need to be able to order them properly.

@cescoffier I'm assigning this one to you for 0.24.0. It will definitely be needed for security + cors.

cescoffier:
Fix #4113.

### DIFF
--- a/extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/VertxHttpRecorder.java
+++ b/extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/VertxHttpRecorder.java
@@ -143,7 +143,7 @@ public class VertxHttpRecorder {
     }
 
     public void finalizeRouter(BeanContainer container, Handler<HttpServerRequest> defaultRouteHandler,
-            List<Handler<RoutingContext>> filters, LaunchMode launchMode, ShutdownContext shutdown,
+            List<Handler<RoutingContext>> orderedFilters, LaunchMode launchMode, ShutdownContext shutdown,
             RuntimeValue<Router> runtimeValue) {
         // install the default route at the end
         Router router = runtimeValue.getValue();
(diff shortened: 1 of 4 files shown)


extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/VertxHttpRecorder.java
```diff
@@ -132,9 +132,9 @@ public static void startServerAfterFailedStart() {
     }
 
     public void finalizeRouter(BeanContainer container, Handler<HttpServerRequest> defaultRouteHandler,
-            List<Handler<RoutingContext>> filters, LaunchMode launchMode, ShutdownContext shutdown) {
+            List<Handler<RoutingContext>> orderedFilters, LaunchMode launchMode, ShutdownContext shutdown) {
 
-        for (Handler<RoutingContext> filter : filters) {
+        for (Handler<RoutingContext> filter : orderedFilters) {
```
gsmet: But aren’t you still adding them in reverse order below?

gsmet (changes requested):
Not totally sure but let’s sort out my comment before merging.

geoand:
Imports need sorting :)

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/VertxHttpRecorder.java
cescoffier: does not need to be reversed, the sorting is done in the processor.

gsmet: But by doing:
```
router.route().order(-1).handler(filter);
```
aren't you inserting them in reverse order? I mean the last one will be first, won't it? Or did I miss something?

cescoffier: If security has ordinal 100, and cors ordinal set to 20, it will add:

security (because of 200 > 20), and then cors. 

why do you need to reverse them?

cescoffier: so on a request, the security filter will be call first, and then the cors filter and then the user route.

stuartwdouglas (approved):

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/VertxHttpRecorder.java
gsmet: Thanks for the comment, it makes things far clearer.

Issue #4081, opened 2019-09-18
AdminClient is not getting all the config properties.


hguerrero:
**Describe the bug**
AdminClient is not able to connect to SSL secured Kafka cluster

**Expected behavior**
AdminClient is configured with the same properties as KafkaStreams 

**Actual behavior**
AdminClient config is only using the bootstrap server

**To Reproduce**
1. Configure SSL connection 
2. Add passthrough config
3. Try to start KafkaStreams application

**Configuration**
```properties
# Add your application.properties here, if applicable.
kafka-streams.security.protocol=SSL
kafka-streams.ssl.keystore.location=src/main/resources/keystore.jks
kafka-streams.ssl.keystore.password=password
kafka-streams.ssl.truststore.location=src/main/resources/keystore.jks
kafka-streams.ssl.truststore.password=password
```

**Screenshots**
(If applicable, add screenshots to help explain your problem.)

**Environment (please complete the following information):**
 - Output of `uname -a` or `ver`: 
 - Output of `java -version`: 
 - GraalVM version (if different from Java): 
 - Quarkus version or git rev: 

**Additional context**
(Add any other context about the problem here.)

gsmet:
Maybe you could provide the error you have? That will be helpful.

Thanks.

hguerrero:
This is the error:

```
2019-09-18 18:38:17,808 ERROR [io.qua.kaf.str.run.KafkaStreamsTopologyManager] (pool-2-thread-1) Failed to get topic names from broker: java.util.concurrent.TimeoutException
        at org.apache.kafka.common.internals.KafkaFutureImpl$SingleWaiter.await(KafkaFutureImpl.java:108)
        at org.apache.kafka.common.internals.KafkaFutureImpl.get(KafkaFutureImpl.java:272)
        at io.quarkus.kafka.streams.runtime.KafkaStreamsTopologyManager.waitForTopicsToBeCreated(KafkaStreamsTopologyManager.java:151)
        at io.quarkus.kafka.streams.runtime.KafkaStreamsTopologyManager.lambda$onStart$1(KafkaStreamsTopologyManager.java:119)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
        at java.lang.Thread.run(Thread.java:748)```

geoand:
Hi,

[...]

PR #4149, opened 2019-09-23
Use all relevant configured properties for Kafka AdminClient 

jaikiran:
The commit here is a potential fix for https://github.com/quarkusio/quarkus/issues/4081. This change makes sure that any relevant `AdminClientConfig`(s) that may have been configured for the application are taken into account when creating a `AdminClient`.

I don't have an easy way to test this or add a test for this (still new to the quarkus testing infrastructure), so I haven't added any tests for this.

### DIFF
--- a/extensions/kafka-streams/runtime/src/main/java/io/quarkus/kafka/streams/runtime/KafkaStreamsTopologyManager.java
+++ b/extensions/kafka-streams/runtime/src/main/java/io/quarkus/kafka/streams/runtime/KafkaStreamsTopologyManager.java
@@ -24,8 +24,8 @@ import javax.inject.Inject;
 import javax.inject.Singleton;
 
 import org.apache.kafka.clients.admin.AdminClient;
+import org.apache.kafka.clients.admin.AdminClientConfig;
 import org.apache.kafka.clients.admin.ListTopicsResult;
-import org.apache.kafka.clients.consumer.ConsumerConfig;
 import org.apache.kafka.streams.KafkaStreams;
 import org.apache.kafka.streams.StreamsConfig;
 import org.apache.kafka.streams.Topology;
@@ -141,8 +141,17 @@ public class KafkaStreamsTopologyManager {
 
     private void waitForTopicsToBeCreated(Set<String> topicsToAwait, String bootstrapServersConfig)
             throws InterruptedException {
-        Map<String, Object> config = new HashMap<>();
-        config.put(ConsumerConfig.BOOTSTRAP_SERVERS_CONFIG, bootstrapServersConfig);
+        final Map<String, Object> config = new HashMap<>();
+        config.put(AdminClientConfig.BOOTSTRAP_SERVERS_CONFIG, bootstrapServersConfig);
+        // include other AdminClientConfig(s) that have been configured
+        for (final String knownAdminClientConfig : AdminClientConfig.configNames()) {
+            // give preference to admin.<propname> first
+            if (properties.containsKey(StreamsConfig.ADMIN_CLIENT_PREFIX + knownAdminClientConfig)) {
+                config.put(knownAdminClientConfig, properties.get(StreamsConfig.ADMIN_CLIENT_PREFIX + knownAdminClientConfig));
+            } else if (properties.containsKey(knownAdminClientConfig)) {
+                config.put(knownAdminClientConfig, properties.get(knownAdminClientConfig));
+            }
+        }
 
         try (AdminClient adminClient = AdminClient.create(config)) {
             while (true) {


extensions/kafka-streams/runtime/src/main/java/io/quarkus/kafka/streams/runtime/KafkaStreamsTopologyManager.java
```diff
@@ -141,8 +142,14 @@ public KafkaStreams getStreams() {
 
     private void waitForTopicsToBeCreated(Set<String> topicsToAwait, String bootstrapServersConfig)
             throws InterruptedException {
-        Map<String, Object> config = new HashMap<>();
+        final Map<String, Object> config = new HashMap<>();
         config.put(ConsumerConfig.BOOTSTRAP_SERVERS_CONFIG, bootstrapServersConfig);
```
gunnarmorling: Now that I'm thinking about it, this should be this instead:
```suggestion
        config.put(AdminClientConfig.BOOTSTRAP_SERVERS_CONFIG, bootstrapServersConfig);
```

extensions/kafka-streams/runtime/src/main/java/io/quarkus/kafka/streams/runtime/KafkaStreamsTopologyManager.java
```diff
@@ -141,8 +142,14 @@ public KafkaStreams getStreams() {
...
+        // include other AdminClientConfig(s) that have been configured
+        for (final String knownAdminClientConfig : AdminClientConfig.configNames()) {
```
gunnarmorling: Ok, I think that goes into the right direction. Only that it should pull the properties from the `kafka-streams` namespace, or better yet `kafka-streams.admin`; i.e. it should work to provide the properties like so:

    kafka-streams.admin.security.protocol=SSL
    ...

gunnarmorling (commented):
Thanks a lot, @jaikiran. Two comments inline.

extensions/kafka-streams/runtime/src/main/java/io/quarkus/kafka/streams/runtime/KafkaStreamsTopologyManager.java
jaikiran: Hi @gunnarmorling, at least in the context of this specific issue which prompted me to look into it, the configurations that are relevant are certain "common" configurations (SSL etc) which are (typically) applicable to consumers, producers and admin client(s) of Kafka streams.
Do you expect the values for these configs to be different in context of different uses (consumer vs producer vs admin client)? In other words, at least for the current configs, do we need an additional kafka-streams.admin.. prefix? As it is, this current change in PR makes it a point that only those properties that are known/recognized by AdminClient will be passed to it.
I'm open to making a new namespace as you suggest, but just want to be sure we need it.

extensions/kafka-streams/runtime/src/main/java/io/quarkus/kafka/streams/runtime/KafkaStreamsTopologyManager.java
jaikiran: You are right - that makes sense. I have updated the PR to use this suggestion.

gsmet:
@gunnarmorling @jaikiran if we can get this in by tomorrow, I'll try to push it to the upcoming 0.23.0 too.

extensions/kafka-streams/runtime/src/main/java/io/quarkus/kafka/streams/runtime/KafkaStreamsTopologyManager.java
jaikiran: Just noticed one part in your comment:
> Ok, I think that goes into the right direction. Only that it should pull the properties from the kafka-streams namespace

The `properties` member variable that you see in there, in the `KafkaStreamsTopologyManager` is actually built off the "kafka-streams" namespace. That part is already handled here  https://github.com/quarkusio/quarkus/blob/master/extensions/kafka-streams/deployment/src/main/java/io/quarkus/kafka/streams/deployment/KafkaStreamsProcessor.java#L142. So the contents of the `properties` member are those properties that begin with `kafka-streams.` prefix, with their prefixes removed.

extensions/kafka-streams/runtime/src/main/java/io/quarkus/kafka/streams/runtime/KafkaStreamsTopologyManager.java
```diff
@@ -141,8 +142,14 @@ public KafkaStreams getStreams() {
 
     private void waitForTopicsToBeCreated(Set<String> topicsToAwait, String bootstrapServersConfig)
             throws InterruptedException {
-        Map<String, Object> config = new HashMap<>();
-        config.put(ConsumerConfig.BOOTSTRAP_SERVERS_CONFIG, bootstrapServersConfig);
```
gsmet: Is it normal that we don't have this one anymore? (Note that I'm not familiar at all with the extension)

jaikiran: Hi @gsmet, both `AdminClientConfig.BOOTSTRAP_SERVERS_CONFIG` and `ConsumerConfig.BOOTSTRAP_SERVERS_CONFIG` constants translate back to the exact same string `bootstrap.servers` property. Gunnar's suggestion was to use `AdminClientConfig` for this constant since it made more sense to use that since we are dealing with an AdminClient and not a consumer.

gsmet: OK, perfect, then.

gsmet:
@gunnarmorling I let you approve this one and I'll merge once it got your approval. Please ping me then :).

extensions/kafka-streams/runtime/src/main/java/io/quarkus/kafka/streams/runtime/KafkaStreamsTopologyManager.java
gunnarmorling: @jaikiran, ok, so it seems we agree there should be the "kafka-streams" prefix, right? Unless I'm mistaken, that's not yet part of the PR?

About the "admin" sub-namespace, the idea was to resemble what you'd two when directly configuing Kafka Streams, where you'd also use `StreamsConfig#adminClientPrefix()` and related methods to prefix these values (see [here](https://docs.confluent.io/current/streams/developer-guide/config-streams.html#naming)). So to specify any value for the KStreams *internal*  admin client, the `kafka-streams.admin.` prefix would have to be used anyways (and `kafka-streams.` would be removed before passing on the value).

Now the question is is there the need for specifying different values for our own admin client here than for the KStreams internal one. If not, I'd argue, the `kafka-streams.admin.` prefix should be used for both.

jaikiran: Hi @gunnarmorling, I looked at the link you pasted in your comment and I see what you mean. So what I have now done is updated this PR to change the way I pass the configs to the `AdminClient`. What it does now (snippet):
```
final Map<String, Object> config = new HashMap<>();
        config.put(AdminClientConfig.BOOTSTRAP_SERVERS_CONFIG, bootstrapServersConfig);
        // include other AdminClientConfig(s) that have been configured
        for (final String knownAdminClientConfig : AdminClientConfig.configNames()) {
            // give preference to admin.<propname> first
            if (properties.containsKey("admin." + knownAdminClientConfig)) {
                config.put(knownAdminClientConfig, properties.get("admin." + knownAdminClientConfig));
            } else if (properties.containsKey(knownAdminClientConfig)) {
                config.put(knownAdminClientConfig, properties.get(knownAdminClientConfig));
            }
        }
```
So it first checks if the `properties` has any `admin.<propname>` (note the `kafka-streams.` prefix is already stripped so no need to check for that) and if it finds such a property, it uses it. Else it checks if any property that's recognized by `AdminClientConfig` exists and if it finds it then uses it. This way, users can now use either (an example):
```
kafka-streams.admin.security.protocol=SSL
```
or if they want to share the config value for all (consumers, producer and admin client), then:

```
kafka-streams.security.protocol=SSL
```
and the `AdminClient` that's constructed in this `KafkaTopologyManager` will honour it correctly.

Do you think this is fine?

gunnarmorling: Yes, I think that makes sense. I'd only suggest to reference `org.apache.kafka.streams.StreamsConfig.ADMIN_CLIENT_PREFIX` instead of specifying "admin." literally. And it would be nice to have a test, but I'm not sure how it could look like, other than changing the existing integration tests to make use of SSL. Would be nice if you're up for it, but no blocker IMHO.

jaikiran: >> I'd only suggest to reference org.apache.kafka.streams.StreamsConfig.ADMIN_CLIENT_PREFIX instead of specifying "admin." literally.

That makes sense. Done - I have updated this PR to make that change.

>> And it would be nice to have a test, but I'm not sure how it could look like, other than changing the existing integration tests to make use of SSL. Would be nice if you're up for it, but no blocker IMHO.

Agreed. I have been adding tests wherever possible to my PRs to this project. But for this one, I still need to understand and read up on the test infrastructure a bit more to be able to add anything useful. I probably will get to it at some point, but I don't think I will be able to do it in the next couple of days.

gunnarmorling: > but I don't think I will be able to do it in the next couple of days.

Yes, I think that's fine. To make the change eventually, you'd have to look into https://github.com/quarkusio/quarkus/blob/master/integration-tests/kafka/src/test/java/io/quarkus/it/kafka/KafkaTestResource.java, which is where the Kafka broker used for testing is started. This would have to be changed to use SSL; then the tests using that broker would have to be altered to specify the SSL options when connecting.

jaikiran: Thank you for that input @gunnarmorling. That definitely helps. I'll go through it and come up with a test case in the next few days.

gunnarmorling:
@hguerrero, any chance you build this PR in the course of the day and give it a test in your environment? It seems you'd have Kafka with SSL already running.

@gsmet, so the change looks good to me, but we don't have ultimate verification as per the discussion above. I think the risk of merging it is rather low (@jaikiran, perhaps you can confirm that passing *unknown*  options would simply be ignored, in case we did something wrong), so I'd go for it.

jaikiran:
>> (@jaikiran, perhaps you can confirm that passing unknown options would simply be ignored, in case we did something wrong)

Yes, Kafka just logs a WARN about unknown configs that gets passed to its consumer/producer/admin client configs and doesn't break at runtime. So if we did in fact end up passing unknown (from a Kafka perspective) configs to this AdminClient, they will end up seeing WARN logs (at worst a spam of such log messages) but won't break the runtime.

gunnarmorling (approved):

gunnarmorling:
@gsmet, so it's good to go from my PoV. Thanks a lot, @jaikiran! If you run into issues with the test in a subsequent PR, let us know, and we can see how to help. Thanks again!

gsmet (approved):
Approving on behalf of Gunnar.

gsmet:
Thanks to both of you.

PR #4154, opened 2019-09-23
Align Kogito components versions in Quarkus

cristianonicolai:


cristianonicolai:
@wburns @tristantarrant @gsmet please take a look, I'm locking the Infinispan to ensure the expected versions, as some of the Kogito dependencies also use Infinispan.

@mswiderski fyi

bom/runtime/pom.xml
```diff
@@ -224,6 +224,15 @@
...
             </dependency>
 
+            <!-- Kogito dependencies, imported as a BOM -->
+            <dependency>
+                <groupId>org.kie.kogito</groupId>
+                <artifactId>kogito-bom</artifactId>
+                <version>${kogito.version}</version>
+                <scope>import</scope>
+                <type>pom</type>
+            </dependency>
```
gsmet: I'm not sure it's worth it to import the Kogito BOM. Can't the users import both the Quarkus BOM and the Kogito BOM in their apps?

/cc @aloubyansky

wburns: That is my sentiment as well, however I am not a maven black wizard :)

bom/runtime/pom.xml
```diff
@@ -1976,11 +1985,36 @@
                 <artifactId>infinispan-commons</artifactId>
                 <version>${infinispan.version}</version>
             </dependency>
+            <dependency>
+                <groupId>org.infinispan</groupId>
+                <artifactId>infinispan-lucene-directory</artifactId>
```
wburns: Is there a reason we are including lucene and directory providers here? These are server artifacts that aren't going to be supported in Quarkus native mode since lucene is not supported there.

bom/runtime/pom.xml
mswiderski: Main reason is to avoid version mismatch different kogito version coming with quarkus and then different one selected by users

bom/runtime/pom.xml
cristianonicolai: @wburns added it here so that the correct versions are enforced in _quarkus-integration-test-infinispan-client_. I believe these deps get pulled in order to run the _InfinispanServerTestResource_.

bom/runtime/pom.xml
gsmet: Yeah, I see the point of it. My main interrogation is: should we add the BOM of all the technologies we use in our BOM. Kogito being a very specific component, I was thinking that maybe the user could include the BOM themselves.

I know @aloubyansky was a bit reluctant to add too many BOMs in ours.

aloubyansky: There is a supported extension for Kogito, so in this regard it's not different from other supported extensions that added their BOMs to the Quarkus one. So, based on that, I'd say lets add it.
However, this is something to think about, as our BOM will keep expanding a lot and rapidly. We may want to consider having a pure Quarkus-core BOM and then let users (and/or our tolling) import the rest separately. There will be different issues with that approach though.

bom/runtime/pom.xml
gsmet: @wburns are the answers OK for you?

wburns: Seems okay to me.

gsmet (approved):
OK, looks like everyone is OK with it, let's merge then.

PR #4155, opened 2019-09-23
Arc - add logging for interceptors/alternatives without @Priority. Fi…

### LINKED ISSUE #3767: Clarify behavior of interceptors/alternatives without @Priority
We do state that the contents of a `beans.xml` file is ignored. However, we don't say what happens if your app contains a interceptor/alternative without `@Priority`.

The current behavior is:
- an interceptor without `@Priority` is enabled and has a default priority of value `0` - `javax.interceptor.Interceptor.Priority.PLATFORM_BEFORE`
- an alternative bean without `@Priority` is not an alternative, but is considered a regular bean

I'm not quite sure whether this behavior makes sense though. From compatibility POV we should probably ignore such interceptors and disable such alternatives. From usability POV it could make sense to keep the interceptors behavior but disable such alternatives.

@manovotn ideas?

In any case, we should add some tests to cover this use case.

manovotn:
…x producer fields with @Alternative.

Fixes #3767

### DIFF
--- a/independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/BeanGenerator.java
+++ b/independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/BeanGenerator.java
@@ -385,6 +385,9 @@ public class BeanGenerator extends AbstractGenerator {
         if (qualifiers != null) {
             implementGetQualifiers(bean, beanCreator, qualifiers.getFieldDescriptor());
         }
+        if (bean.isAlternative()) {
+            implementGetAlternativePriority(bean, beanCreator);
+        }
         implementGetDeclaringBean(beanCreator);
         if (stereotypes != null) {
             implementGetStereotypes(bean, beanCreator, stereotypes.getFieldDescriptor());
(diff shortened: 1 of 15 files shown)


manovotn:
@mkouba this does what we agreed on via Zulip - logging only, keeping the current functionality as is.

I've discovered that we didn't check for `@Alternative` on producer fields at all, so this PR also has a bugfix in it.
Do you think we should cover this behaviour in docs? I was thinking we don't want to promote it, so it may be better to just keep it 'invisible' and any user running into it will now see some logs.

manovotn:
In the current state of this PR, we do logging for interceptors w/o priority and we throw exceptions for alternatives w/o priority.
Since `@Priority` cannot be placed on methods/fields, I've introduced a new annotation, `@AlternativePriority`, that is a combination of `@Alternative` and `@Priority` and can be placed everywhere needed.
ATM, if there are both annotations present, nothing blows up, but `@AlternativePriority` wins (over priority in case it is a class). In order for alternative producer, the declaring bean needs to be an alternative (this is spec-defined) but also needs to be an enabled alternative (spec is dubious in this regards but we already disallow alternatives w/o priority, so this is consistent behaviour).

mkouba (approved):
LGTM.

Issue #4110, opened 2019-09-19
Clean up dependency version overrides in MP Metrics TCK


jmartisk:
https://github.com/quarkusio/quarkus/blob/a79287cef3424777009edc8c98db91134411459d/tcks/microprofile-metrics/api/pom.xml#L57 
https://github.com/quarkusio/quarkus/blob/a79287cef3424777009edc8c98db91134411459d/tcks/microprofile-metrics/rest/pom.xml#L61

If possible we should get rid of them

kenfinnigan:
I added those, and I'm 99% sure they're still needed due to what's brought in through other versions.

By all means, give it a shot though.

jmartisk:
Removing the hamcrest-core override in API TCK will require a fix on the MP Metrics side: https://github.com/eclipse/microprofile-metrics/issues/443

I managed to remove the `shrinkwrap-resolver-api-maven` override and if we just exclude it (from the REST tck dependency) the test suite works because it's not needed. I'll send a PR with this in a bit. I'm not yet sure but it is possible that we might be able to remove it from the TCK's pom.xml as well.

PR #4158, opened 2019-09-23
[#4110] Remove unnecessary dependency in MP Metrics REST TCK

jmartisk:
First part of the fix for #4110
The other part will need a new MP Metrics spec micro-release


jmartisk:
CC @kenfinnigan

kenfinnigan (approved):

PR #4161, opened 2019-09-23
Fixes #4160: Do not register HUP and QUIT signals when in Windows

### LINKED ISSUE #4160: Windows native-image Signal handling
Running a Graal native-image build of the getting-started application on Windows fails with the following error:

`Exception in thread "main" java.lang.IllegalArgumentException: Unknown signal: HUP
        at sun.misc.Signal.<init>(Signal.java:143)
        at io.quarkus.runtime.Application.run(Application.java:185)
        at io.quarkus.runner.GeneratedMain.main(GeneratedMain.zig:34)`

The issue occurs because the HUP signal is not defined for the Windows OpenJDK implementation:
https://hg.openjdk.java.net/jdk8u/jdk8u-dev/hotspot/file/7d5c800dae75/src/os/windows/vm/jvm_windows.cpp

I know that GraalVM does not currently support Signal Handler registration for Windows. However, I suspect this will become an issue once they do.

Quarkus version: 0.21.1
Graal version: 19.3.0-dev (hacked in order to add signal handling from jvm_windows.cpp to jvmFuncs.c in Graal).

gastaldi:
@osmanrafiqss can you confirm if this PR fixes your issue?

Fixes #4160

### DIFF
--- a/core/runtime/src/main/java/io/quarkus/runtime/Application.java
+++ b/core/runtime/src/main/java/io/quarkus/runtime/Application.java
@@ -9,6 +9,8 @@ import org.graalvm.nativeimage.ImageInfo;
 import org.wildfly.common.Assert;
 import org.wildfly.common.lock.Locks;
 
+import com.oracle.svm.core.OS;
+
 import io.quarkus.runtime.graal.DiagnosticPrinter;
 import sun.misc.Signal;
 import sun.misc.SignalHandler;
@@ -182,21 +184,18 @@ public abstract class Application {
     public final void run(String[] args) {
         try {
             if (ImageInfo.inImageRuntimeCode() && System.getenv(DISABLE_SIGNAL_HANDLERS) == null) {
-                final SignalHandler handler = new SignalHandler() {
-                    @Override
-                    public void handle(final Signal signal) {
-                        System.exit(signal.getNumber() + 0x80);
-                    }
-                };
-                Signal.handle(new Signal("INT"), handler);
-                Signal.handle(new Signal("TERM"), handler);
-                Signal.handle(new Signal("HUP"), handler);
-                Signal.handle(new Signal("QUIT"), new SignalHandler() {
-                    @Override
-                    public void handle(final Signal signal) {
-                        DiagnosticPrinter.printDiagnostics(System.out);
-                    }
-                });
+                final SignalHandler handler = signal -> System.exit(signal.getNumber() + 0x80);
+                final SignalHandler quitHandler = signal -> DiagnosticPrinter.printDiagnostics(System.out);
+                handleSignal("INT", handler);
+                handleSignal("TERM", handler);
+                // the HUP and QUIT signals are not defined for the Windows OpenJDK implementation:
+                // https://hg.openjdk.java.net/jdk8u/jdk8u-dev/hotspot/file/7d5c800dae75/src/os/windows/vm/jvm_windows.cpp
+                if (OS.getCurrent() == OS.WINDOWS) {
+                    handleSignal("BREAK", quitHandler);
+                } else {
+                    handleSignal("HUP", handler);
+                    handleSignal("QUIT", quitHandler);
+                }
             }
 
             final ShutdownHookThread shutdownHookThread = new ShutdownHookThread(Thread.currentThread());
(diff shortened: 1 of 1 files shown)


core/runtime/src/main/java/io/quarkus/runtime/Application.java
```diff
@@ -190,13 +190,17 @@ public void handle(final Signal signal) {
...
-                    @Override
-                    public void handle(final Signal signal) {
-                        DiagnosticPrinter.printDiagnostics(System.out);
-                    }
-                });
+                if (!System.getProperty("os.name").contains("Windows")) {
+                    // the HUP and QUIT signals are not defined for the Windows OpenJDK implementation:
+                    // https://hg.openjdk.java.net/jdk8u/jdk8u-dev/hotspot/file/7d5c800dae75/src/os/windows/vm/jvm_windows.cpp
+                    Signal.handle(new Signal("HUP"), handler);
+                    Signal.handle(new Signal("QUIT"), new SignalHandler() {
```
dmlloyd: On Windows, we should look for `SIGBREAK` instead of `SIGQUIT`.

dmlloyd (changes requested):
See comments.

dmlloyd (approved):

osmanrafiqss:
@gastaldi this PR fixes the issue :)

However, as I more or less just copied the signal registration from jvm_windows.cpp to Graal jvmFuncs.c - I did run into a minor issue with the OpenJDK code not allowing registration of a signal handler for the "BREAK" signal unless the "-Xrs" flag is specified.

`Exception in thread "main" java.lang.IllegalArgumentException: Signal already used by VM or OS: SIGBREAK
        at sun.misc.Signal.handle(Signal.java:166)
        at io.quarkus.runtime.Application.run(Application.java:192)
        at io.quarkus.runner.GeneratedMain.main(GeneratedMain.zig:34)`

But looking at the jvm_linux.cpp from OpenJDK (which has the same constraint) and comparing this with the Graal teams implementation in SunMiscSubstituions for posix I can see they are working around this - so probably they will do the same for Windows and we should be good :)

gsmet:
@dmlloyd can you check above comment and confirm if we should amend the PR in some way or if we are good?

dmlloyd:
Honestly these should probably be doing what I do in the command mode branch, and wrap everything in `try`/`catch`: https://github.com/quarkusio/quarkus/pull/2866/files#diff-b884ac8878a228ba03046ea045b2d9cfR93-R110

gsmet:
Couldn't get GH to display the right portion of the code in this huge diff. Here is a more focused link: https://github.com/quarkusio/quarkus/blob/0f074935c6f85cb190bf4fa5a7a9c50a4612eabf/core/runtime/src/main/java/io/quarkus/runtime/execution/SignalHandlerExecutionHandler.java#L101-L110 .

gastaldi:
>wrap everything in try/catch

@dmlloyd so ignoring even the underlying OS completely or simply silently ignore the exception?

dmlloyd:
I guess if you probe the OS you can eliminate some "known" exceptions, which could shave off a few microseconds.  Otherwise just ignore exceptions.  If you can't construct the signal object, ignore the failure and don't register a handler for it.  If registering a handler fails, ignore the failure.

gastaldi:
Right, I force-pushed my changes. I think it matches with what you said. Thanks

dmlloyd (approved):

core/runtime/src/main/java/io/quarkus/runtime/Application.java
```diff
@@ -182,21 +184,18 @@ public static Application currentApplication() {
...
-                Signal.handle(new Signal("TERM"), handler);
-                Signal.handle(new Signal("HUP"), handler);
-                Signal.handle(new Signal("QUIT"), new SignalHandler() {
-                    @Override
-                    public void handle(final Signal signal) {
-                        DiagnosticPrinter.printDiagnostics(System.out);
-                    }
-                });
+                final SignalHandler handler = signal -> System.exit(signal.getNumber() + 0x80);
+                final SignalHandler quitHandler = signal -> DiagnosticPrinter.printDiagnostics(System.out);
```
gsmet: @gastaldi I will backport this as is but could you go back to not using lambdas? We are trying to avoid them in runtime code and AFAICS, this is runtime code.

Thanks!

gastaldi: Sure, I'll provide a new PR without them, thank you!

Issue #2176, opened 2019-04-22
Kafka Guide does not work


gastaldi:
I followed the steps in the Kafka Guide in https://quarkus.io/guides/kafka-guide and when I open the page, all I get is 
```
Last price

The last price is N/A €.
```

You can see that the connection closes prematurely if you use HTTPie to connect to `/prices/stream` as in:
 
```http --stream -f localhost:8080/prices/stream -v```
 
By debugging and comparing to the quickstart in https://github.com/quarkusio/quarkus-quickstarts/tree/master/kafka-quickstart (which works btw), the only missing part (and what makes the guide work) is to add the following code to the `PriceResource`:

```java
    @GET
    @Produces(MediaType.TEXT_PLAIN)
    public String hello() {
        return "hello";
    }
```

See: https://github.com/quarkusio/quarkus-quickstarts/blob/master/kafka-quickstart/src/main/java/org/acme/quarkus/sample/PriceResource.java#L21-L25

gsmet:
We merged a work around in the guide but we need to get to the bottom of this as it definitely doesn't look like normal.

jaikiran:
There appear to be a couple of issues going on here.

The root cause of this issue seems to be that in the absence of that above workaround (where a `@GET` `@Produces(TEXT_PLAIN)` is added), the `io.quarkus.resteasy.server.common.deployment.ResteasyServerCommonProcessor` isn't registering some of the (MessageWriter) `Provider`(s) (one of them being `org.jboss.resteasy.plugins.providers.DefaultNumberWriter`). Because of that, at runtime, when the streaming URL resource is requested, it runs into this exception:
```
2019-09-23 19:40:03,939 DEBUG [org.jbo.res.res.i18n] (vert.x-eventloop-thread-1) RESTEASY002030: Failed to write event org.jboss.resteasy.plugins.providers.sse.OutboundSseEventImpl@6ab5d8af: javax.ws.rs.ServerErrorException: RESTEASY004082: No suitable message body writer for class : java.lang.Double
	at org.jboss.resteasy.plugins.providers.sse.SseEventProvider.writeTo(SseEventProvider.java:111)
	at org.jboss.resteasy.plugins.providers.sse.SseEventProvider.writeTo(SseEventProvider.java:28)
	at org.jboss.resteasy.plugins.providers.sse.SseEventOutputImpl.writeEvent(SseEventOutputImpl.java:262)
[...]

PR #4163, opened 2019-09-23
[issue#2176] register the commonly needed text/plain MediaType MessageWriter by default

jaikiran:
The commit here by default registers the `MessageWriter` which deals with the commonly needed `text/plain` `MediaType`. This should solve issues like the one noted in https://github.com/quarkusio/quarkus/issues/2176


geoand:
@jaikiran can you please rebase onto the latest master?

jaikiran:
Hi @geoand, Actually let me close this. I think there's a better way to deal with this than blindly adding the provider (although it's a commonly used one). I will open a separate PR with the proposal

PR #4165, opened 2019-09-23
feat(rest-json): emits a warning during build time if we detect the need of a json extension and none is provided

### LINKED ISSUE #4157: Helpfull error message when there is no JSON extension in the classpath and @Produces(MediaType.APPLICATION_JSON) is used
To reproduce:
- Do getting started guide (= hello world example)
- Change @Produces(MediaType.TEXT-PLAIN) with @Produces(MediaType.APPLICATION_JSON)
- Change return type to LocalDate (or Duration or UUID)
- Browse to that url.

Output:

`Could not find MessageBodyWriter for response object of type: java.time.Duration of media type: application/json`

That's not the problem! There is no json extension (for example jackson) in the classpath. Once the error is detected, before the error is thrown, it could check if any json extension is available and if none is, mention that in the error message.

machi1990:
Fixes #4157

/cc @geoand


geoand (approved):
LGTM. Thanks @machi1990!

PR #4167, opened 2019-09-24
Register any additional necessary providers for server sent event @Produce(s)

### LINKED ISSUE #2176: Kafka Guide does not work

jaikiran:
The commit here introduces a way where any additional necessary `providers` are registered for any resource that `@Produces` `MediaType.SERVER_SENT_EVENTS`. This should allow for fixing the issue reported in https://github.com/quarkusio/quarkus/issues/2176 such that the `PriceResouce` in that example can now use:

```
@GET
    @Path("/stream")
    @Produces("text/event-stream;element-type=text/plain")
    public Publisher<Double> stream() {
        return prices;
    }
```
and will no longer require a (dummy) resource endpoint to register the basic `text/plain` type providers.

Fixes #2176

### DIFF
--- a/extensions/resteasy-common/deployment/src/main/java/io/quarkus/resteasy/common/deployment/ResteasyCommonProcessor.java
+++ b/extensions/resteasy-common/deployment/src/main/java/io/quarkus/resteasy/common/deployment/ResteasyCommonProcessor.java
@@ -5,6 +5,7 @@ import static io.quarkus.deployment.annotations.ExecutionTime.STATIC_INIT;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
@@ -25,11 +26,13 @@ import org.jboss.jandex.DotName;
 import org.jboss.jandex.IndexView;
 import org.jboss.jandex.MethodInfo;
 import org.jboss.logging.Logger;
+import org.jboss.resteasy.annotations.SseElementType;
 import org.jboss.resteasy.core.MediaTypeMap;
 import org.jboss.resteasy.plugins.interceptors.AcceptEncodingGZIPFilter;
 import org.jboss.resteasy.plugins.interceptors.GZIPDecodingInterceptor;
 import org.jboss.resteasy.plugins.interceptors.GZIPEncodingInterceptor;
 import org.jboss.resteasy.plugins.providers.StringTextStar;
+import org.jboss.resteasy.plugins.providers.sse.SseConstants;
 import org.jboss.resteasy.spi.InjectorFactory;
 
 import io.quarkus.arc.deployment.BeanArchiveIndexBuildItem;
@@ -52,6 +55,8 @@ import io.quarkus.runtime.configuration.MemorySize;
 
 public class ResteasyCommonProcessor {
     private static final Logger LOGGER = Logger.getLogger(ResteasyCommonProcessor.class.getName());
+
+    private static final DotName SSE_ELEMENT_TYPE = DotName.createSimple(SseElementType.class.getName());
 
     private static final ProviderDiscoverer[] PROVIDER_DISCOVERERS = {
             new ProviderDiscoverer(ResteasyDotNames.GET, false, true),
@@ -283,7 +288,7 @@ public class ResteasyCommonProcessor {
             for (AnnotationInstance classAnnotation : classAnnotations) {
                 if (mediaTypeAnnotation.equals(classAnnotation.name())) {
                     if (collectDeclaredProvidersForMediaTypeAnnotationInstance(providersToRegister, categorizedProviders,
-                            classAnnotation)) {
+                            classAnnotation, methodTarget)) {
                         return true;
                     }
                     return false;
@@ -292,7 +297,7 @@ public class ResteasyCommonProcessor {
             return defaultsToAll;
         }
         if (collectDeclaredProvidersForMediaTypeAnnotationInstance(providersToRegister, categorizedProviders,
-                mediaTypeAnnotationInstance)) {
+                mediaTypeAnnotationInstance, methodTarget)) {
             return true;
         }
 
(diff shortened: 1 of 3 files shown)


gsmet:
This is more correct but I doubt people will get it right.

Another option would be to include the text plain media type as soon as we have a `MediaType.SERVER_SENT_EVENTS`. Maybe we could have a generic method contributing additional media types based on a media type.

Typically, we would call `contributeAdditionalMediaTypes()` at these places:
- https://github.com/quarkusio/quarkus/blob/master/extensions/resteasy-common/deployment/src/main/java/io/quarkus/resteasy/common/deployment/ResteasyCommonProcessor.java#L188
- https://github.com/quarkusio/quarkus/blob/master/extensions/resteasy-common/deployment/src/main/java/io/quarkus/resteasy/common/deployment/ResteasyCommonProcessor.java#L198
- https://github.com/quarkusio/quarkus/blob/master/extensions/resteasy-common/deployment/src/main/java/io/quarkus/resteasy/common/deployment/ResteasyCommonProcessor.java#L208

and it would return an augmented list of elements in some cases (for now, only the `MediaType.SERVER_SENT_EVENTS` case).

Does it make sense to you?

jaikiran:
> This is more correct but I doubt people will get it right.

@gsmet, I understand what you mean.

> Another option would be to include the text plain media type as soon as we have a MediaType.SERVER_SENT_EVENTS. Maybe we could have a generic method contributing additional media types based on a media type.

Okay, so instead of expecting users to use an `element-type` `MediaType` parameter, we by default add the `text/plain` provider if and when we see the use of `MediaType.SERVER_SENT_EVENTS` in the application. That sounds fine to me.

> Typically, we would call contributeAdditionalMediaTypes() at these places: 

I have looked at the code that you pointed to (the 3 places) and do understand what you mean. However, those 3 places won't be the right place to do this logic, since those 3 places are processing the providers that have been found in the classpath, rather than the user's application resources which use those providers. That latter part happens later and I have updated this PR to use this common logic in that central place where it happens for `MessageWriter`, `MessageReader` and `ContextResolver`. 

The commit in this PR now adds the `text/plain` provider if it finds the usage of server sent events, plus also supports any additional `element-type` parameter if at all the user application has used such a construct.

Furthermore, I have included a test case to test the server sent events. Do note that this test case is ineffective in _reproducing_ the original reported issue because of the way we run our integration tests, where multiple other JaxRS endpoints are scanned and because of their usage of `text/plain`, the providers will anyway be registered and this issue won't be triggered. However, this test at least adds some amount of testing around server sent events.

asoldano (approved):
I think this proposal from Jaikiran is fine; see section 5 at https://docs.jboss.org/resteasy/docs/4.3.0.Final/userguide/html/Reactive.html#d4e2322

geoand:
If @asoldano agrees, then it I think we should merge. @gsmet ?

gsmet:
Please wait.

AFAICS there is also `@SseElementType` which seems to be used in this case - see example 1 here: https://docs.jboss.org/resteasy/docs/4.3.0.Final/userguide/html/Reactive.html#d4e2483

I think I would add `TEXT_PLAIN_TYPE` only if the element type is not defined one way or another.

What I'm unclear about is the precedence between the `;element-type` parameter and the `@SseElementType` annotation.

@asoldano could you shed some light on this?

geoand:
@gsmet up to you :)

asoldano:
@gsmet uh, thanks, I would need to run some tests, but you might have made me notice a "bug". So `@SeeElementType` is the RESTEasy annotation that controls the `;element-type` parameter, however it looks like a `@Produce("text/event-stream; element-type=foo/bar")` would result in a response with no parameter is the `@SseElementType` annotation is not there too. Anyway, for the Quarkus case, it's probably safe to check if the `element-type` parameter is defined in any way (directly or with the dedicated annotation), with the `@SseElementType` overriding the value from `@Produces`.

gsmet:
@asoldano OK, cool, thanks!

@jaikiran could you implement the above? (and defaults to text/plain if we don't have any meaningful information). I think it would greatly improve things in Quarkus.

jaikiran:
Thank you @asoldano for all these details and the review and thank you @gsmet for the patience with this review. I have now pushed a change to this PR which takes into account what we have discussed here. The comments in the code should summarize how we deal with it now (essentially the same as what we decided here).

geoand:
There are some formatting issues.

You can do `mvn process-resources` to fix them and then update the PR

geoand:
Clean install should do the trick. It has so often however that we commit before doing it 😅

extensions/resteasy-common/deployment/src/main/java/io/quarkus/resteasy/common/deployment/ResteasyCommonProcessor.java
```diff
@@ -269,26 +274,95 @@ private static boolean collectDeclaredProvidersForMethodAndMediaTypeAnnotation(S
...
+        if (matches(MediaType.SERVER_SENT_EVENTS_TYPE, mediaType)) {
+            final Set<String> additionalProvidersToRegister = new HashSet<>();
+            // first check for @SseElementType
+            final AnnotationInstance sseElementTypeAnnInst = targetMethod.annotation(SSE_ELEMENT_TYPE);
+            String elementType = null;
+            if (sseElementTypeAnnInst != null) {
+                elementType = sseElementTypeAnnInst.value().asString();
+            } else {
+                // fallback on the MediaType parameter
+                if (mediaType.getParameters() != null
```
gsmet: It's really very very minor but can it be a `else if` instead?

gsmet (changes requested):
@jaikiran I added a very minor comment but the main thing is that it needs a rebase (I would have merged as is if not).

Could you take care of the rebase and fix my minor remark along the way?

Other than that, it's perfect, thanks a lot for this.

Thanks!

jaikiran:
Hi @gsmet, I've updated this PR after rebasing with latest and also incorporated the (valid) change that you mentioned in the latest comment.

kenfinnigan (approved):

gsmet (approved):

gsmet:
Merged, very nice improvement, thanks!

PR #4168, opened 2019-09-24
Ensure maven integration tests don't fail due to clashing random file name generation

geoand:
Based on the failure I saw [here](https://dev.azure.com/quarkus-ci/quarkus/_build/results?buildId=8344&view=logs)

### DIFF
--- a/integration-tests/maven/src/test/java/io/quarkus/maven/CreateExtensionMojoTest.java
+++ b/integration-tests/maven/src/test/java/io/quarkus/maven/CreateExtensionMojoTest.java
@@ -8,6 +8,7 @@ import java.nio.file.Paths;
 import java.util.Arrays;
 import java.util.LinkedHashSet;
 import java.util.Set;
+import java.util.UUID;
 import java.util.stream.Collectors;
 
 import org.apache.maven.plugin.MojoExecutionException;
@@ -24,8 +25,7 @@ public class CreateExtensionMojoTest {
          * We want to run on the same project multiple times with different args so let's create a copy with a random
          * suffix
          */
-        final Path copyDir = Paths
-                .get("target/test-classes/projects/" + testProjectName + "-" + ((int) (Math.random() * 1000)));
+        final Path copyDir = getCopyDir(testProjectName);
         Files.walk(srcDir).forEach(source -> {
             try {
                 final Path dest = copyDir.resolve(srcDir.relativize(source));
@@ -43,6 +43,20 @@ public class CreateExtensionMojoTest {
         mojo.assumeManaged = true;
         mojo.nameSegmentDelimiter = CreateExtensionMojo.DEFAULT_NAME_SEGMENT_DELIMITER;
         return mojo;
+    }
+
+    private static Path getCopyDir(String testProjectName) {
+        int count = 0;
+        while (count < 100) {
+            Path path = Paths.get("target/test-classes/projects/" + testProjectName + "-" + UUID.randomUUID());
+            if (!Files.exists(path)) {
+                return path;
+            }
+            count++;
+        }
+
+        // if we have tried too many times we just give up instead of looping forever which could cause the test to never end
+        throw new RuntimeException("Unable to create a directory for copying the test application into");
     }
 
     @Test


integration-tests/maven/src/test/java/io/quarkus/maven/CreateExtensionMojoTest.java
```diff
@@ -45,6 +44,15 @@ static CreateExtensionMojo createMojo(String testProjectName) throws IllegalArgu
         return mojo;
     }
 
+    private static Path getCopyDir(String testProjectName) {
+        while (true) {
+            Path path = Paths.get("target/test-classes/projects/" + testProjectName + "-" + ((int) (Math.random() * 1000)));
```
gsmet: I wonder if using `Random` and `nextInt` would be better than this.

geoand: It probably would look aesthetically better, but there shouldn't be any real difference.
Would you like me to change it?

gsmet: Not mandatory but it would make things a bit cleaner.

Also, I'm not a big fan of having `while (true)`. Could we maybe limit it to 100 iterations and throw an exception if we can't find one valid path in 100 iterations?

geoand: Good idea. I'll do that ASAP

geoand: Fixed

integration-tests/maven/src/test/java/io/quarkus/maven/CreateExtensionMojoTest.java
```diff
@@ -45,6 +45,22 @@ static CreateExtensionMojo createMojo(String testProjectName) throws IllegalArgu
         return mojo;
     }
 
+    private static Path getCopyDir(String testProjectName) {
+        Random random = new Random();
+
+        int count = 0;
+        while (count < 100) {
+            Path path = Paths.get("target/test-classes/projects/" + testProjectName + "-" + random.nextInt(1000));
```
machi1990: Just a minor question, could we have used an `UUID` instead?

geoand: Yeah, but I don't think we should be overthinking such a small issue :)

integration-tests/maven/src/test/java/io/quarkus/maven/CreateExtensionMojoTest.java
```diff
@@ -45,6 +45,22 @@ static CreateExtensionMojo createMojo(String testProjectName) throws IllegalArgu
         return mojo;
     }
 
+    private static Path getCopyDir(String testProjectName) {
+        Random random = new Random();
```
gsmet: I think this would be better as static, don't you think?

gsmet:
There is a formatting issue.

geoand:
Of course there is...

geoand:
Should be fixed now

integration-tests/maven/src/test/java/io/quarkus/maven/CreateExtensionMojoTest.java
geoand: I've always been a little hazy about whether Random should be `static` or not. Let me read up

geoand: It looks like having it as an argument will cause it to use a new seed which should theoretically make it more random. 
In any case it's a super small issue, I don't mind doing it either way

mkouba: Shouldn't we use `UUID.randomUUID()` instead?

geoand: Since this is the second time this has come up, I'll go ahead and do it :P

gsmet: @geoand you don't want it to be super random because you don't want conflicts :). Using only one `Random` is recommended.

@mkouba why not? @machi1990 was suggesting this too.

geoand: `UUID.randomUUID()` it is :)

integration-tests/maven/src/test/java/io/quarkus/maven/CreateExtensionMojoTest.java
geoand: Done

integration-tests/maven/src/test/java/io/quarkus/maven/CreateExtensionMojoTest.java
```diff
@@ -45,6 +45,20 @@ static CreateExtensionMojo createMojo(String testProjectName) throws IllegalArgu
         return mojo;
     }
 
+    private static Path getCopyDir(String testProjectName) {
+        int count = 0;
+        while (count < 100) {
+            Path path = Paths.get("target/test-classes/projects/" + testProjectName + "-" + UUID.randomUUID());
```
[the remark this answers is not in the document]
geoand: I propose we leave it there and never have to deal with this issue again :P

gsmet (approved):

PR #4171, opened 2019-09-24
Log the H2 DB server status to provide more details on its state and URL

jaikiran:
A very minor change to print out the "status" of the H2 DB when it's started and stopped, so that details like the following gets printed when running integration tests:
```
[INFO] H2 database started in TCP server mode; server status: TCP server running at tcp://localhost:9092 (only local connections)
```
This additional info would have helped me when I ran into an issue trying to run the integration tests locally (more details on quarkus-dev mailing list)


gsmet (approved):
Looks like a useful addition. Thanks!

Let's wait for CI.

PR #4176, opened 2019-09-24
Remove an outdated Jackson version

gsmet:
This version has a CVE: we should use the one from the bom.

Note: all the versions in this pom should probably go away. Not totally sure if they should be in the bom though as they are really internal dependencies. Maybe let's wait for @aloubyansky 's work on the platform as there's a good chance we will split the internal bom from the outside world one.

### DIFF
--- a/core/processor/pom.xml
+++ b/core/processor/pom.xml
@@ -36,8 +36,8 @@
         <dependency>
             <groupId>com.fasterxml.jackson.core</groupId>
             <artifactId>jackson-databind</artifactId>
-            <version>2.9.9.3</version>
         </dependency>
+
         <!-- Test -->
         <dependency>
             <groupId>org.junit.jupiter</groupId>


machi1990 (approved):

geoand (approved):

aloubyansky:
We don't have to wait for the platform for that. There is a PR that splits the parent into the user parents and the internal one waiting for approval (and probably rebase now). I am not going to be keeping it up-to-date everyday unless we want to merge it though, it's pretty large. But the issue could be addressed there.

gsmet:
Yeah, I was thinking about this very PR. I'll try to have a look after the release, then.

geoand:
So we merge this then or not?

gsmet:
Yes, we merge it and the rest of the cleanup will wait (we don't want versions defined in deeply nested pom files).

PR #4179, opened 2019-09-24
Account for more complex throws exception types on proxies.

### LINKED ISSUE #4178: ClientProxyGenerator - handle thrown exception types correctly
Currently, we only expect a `ClassType`: https://github.com/quarkusio/quarkus/blob/0.22.0/independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/ClientProxyGenerator.java#L100

manovotn:
Fixes #4178

### DIFF
--- a/independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/ClientProxyGenerator.java
+++ b/independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/ClientProxyGenerator.java
@@ -98,13 +98,7 @@ public class ClientProxyGenerator extends AbstractGenerator {
 
             // Exceptions
             for (Type exception : method.exceptions()) {
-                if (exception.kind() == Type.Kind.CLASS) {
-                    forward.addException(exception.asClassType().toString());
-                } else if (exception.kind() == Type.Kind.TYPE_VARIABLE) {
-                    forward.addException(exception.asTypeVariable().bounds().get(0).asClassType().toString());
-                } else {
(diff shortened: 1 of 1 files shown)


kenfinnigan (approved):

independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/ClientProxyGenerator.java
```diff
@@ -98,7 +98,14 @@ public ClientProxyGenerator(Predicate<DotName> applicationClassPredicate) {
 
             // Exceptions
             for (Type exception : method.exceptions()) {
-                forward.addException(exception.asClassType().toString());
+                Type.Kind kind = exception.kind();
+                if (Type.Kind.CLASS.equals(kind)) {
```
mkouba: Hm, so I'm wondering:
1. Do we actually need to check the specific type or just always use `toString()`?
2. Should we add some tests? ;-)

manovotn: 1. that's a good point, we can just use `toString()`
2. I thought this change a bit too trivial to write a test for it, feel free to add one ;)

manovotn:
Apparently, Stuart came across the same thing here and just pushed that as part of another fix - https://github.com/quarkusio/quarkus/commit/22a354c7a3ca569c601aa5bc7218655e9cc53c20

I'll rebase it and keep just the plain `toString()` as that seems to be universal for all cases...

PR #4181, opened 2019-09-24
Improve the generated code for Spring Data JPA fragments

geoand:
This is done by actually creating an injection point for the bean
that implements each fragment instead of pulling the bean out of
Arc at runtime


geoand:
@gsmet I came across the improvement while looking into the CI issues. It's not critical by any means (since I couldn't reproduce the CI issue with the previous or this code), but it's nonetheless nice to have.

gsmet (approved):
I don't have a strong opinion on this but we have tests so I trust you on this :).

geoand:
Good idea 😉

Issue #4193, opened 2019-09-25
Followup on MongoDB with Panache


loicmathieu:
**Description**
This is an umbrella issue to track some enhancements and followup works to do to the just-landed MongoDB with Panache extension (#3650).

The list of tasks are:

- [ ] 	Investiguate the support of transactions
- [ ] 	Avoid reflection when building PanacheQL queries from `io.quarkus.mongodb.panache.runtime.PanacheQlQueryBinder`
- [x] 	Implements query projection
- [x] Avoid re-creation of the `Jsonb` object on each request inside `io.quarkus.mongodb.panache.jsonb.PanacheMongoJsonbContextResolver`
- [ ] Optimize the way the ID of the entity is retrieved inside `io.quarkus.mongodb.panache.runtime.MongoOperations`
- [x] Improve the way Jackson `ObjectId` serializer/deserializer are provided
- [x] Improve the way Jsonb `ObjectId` serializer/deserializer are provided
- [ ] Improve the way PanacheQL queries are parsed. Today it is based on a lot of String concatenation, we could do better.
- [x] Provides a quickstart application

loicmathieu:
@gsmet @emmanuelbernard @FroMage I will provide separate PR for each tasks listed, you can assign to me this issue.

Don't hesitate to give feedback or add missing tasks, I tried to be as exhaustif as possible regarding what we discussed on the current implementation.

loicmathieu:
I close this one as:

- Transaction support is curently prototyped in two PRs: #6365 and #7222 
- A lot of works has been done to avoid reflection at runtime
- String concatenation is as powerful as using a `StringBuilder` since Java 9 (new string concat factory). I've done tests and modifying the `MongoParserVisitor` in various way to avoid String concatenation without substential performance improvements.

PR #4196, opened 2019-09-25
Avoid creating a Jsonb each time in Mongo Panache

loicmathieu:
This is part of #4193

I'm not sure it's the best way to do this. Maybe someone from Jsonb / resteasy can give me feedback ?

### DIFF
--- a/extensions/panache/mongodb-panache/runtime/src/main/java/io/quarkus/mongodb/panache/jsonb/PanacheMongoJsonbContextResolver.java
+++ b/extensions/panache/mongodb-panache/runtime/src/main/java/io/quarkus/mongodb/panache/jsonb/PanacheMongoJsonbContextResolver.java
@@ -16,10 +16,16 @@ import javax.ws.rs.ext.Provider;
 @Provider
 public class PanacheMongoJsonbContextResolver implements ContextResolver<Jsonb> {
 
-    public Jsonb getContext(Class clazz) {
+    private final Jsonb jsonb;
+
+    public PanacheMongoJsonbContextResolver() {
         JsonbConfig config = new JsonbConfig();
         config.withSerializers(new ObjectIdSerializer()).withDeserializers(new ObjectIdDeserializer());
-        return JsonbBuilder.create(config);
+        this.jsonb = JsonbBuilder.create(config);
+    }
+
+    public Jsonb getContext(Class clazz) {
+        return jsonb;
     }
 
 }


geoand:
Not an expert in either, but I do think this is the right thing to do.  It is essentially what I have done here: https://github.com/quarkusio/quarkus/pull/3553/files#diff-9a1fe9d42ef5018e9ebc5cda3e45b9f9R22

gsmet:
But how does it work when we have multiple extensions doing that? Or the user willing to add their own serializers?

I don't think this approach will fly.

geoand:
We could do the following:

Provide a CDI producer for `Jsonb` (which I also did in my PR) and just inject the Jsonb bean into the `ContextResolver`.
We could also provide "hooks" for users to customize the the default jsonb bean instead of replacing it entirely.

If you like, I can create a PR showing what I have in mind

geoand:
Another thing we should do is align the handling between Jackson and JSON-B in RESTEasy

gsmet:
> Provide a CDI producer for Jsonb (which I also did in my PR) and just inject the Jsonb bean into the ContextResolver.

I agree with that.

> We could also provide "hooks" for users to customize the the default jsonb bean instead of replacing it entirely.

I'm thinking of 2 things:
* discover serializers automatically via Jandex (and other things if they make sense)
* have build items for extensions to push serializers
* then I don't know how we can allow the user to really tweak the JsonbConfig because in some cases, they might need to. Should we allow to entirely overwrite the jsonb bean via `@DefaultBean` (but that could break some extensions or the user would have to figure out what the extensions are doing)

geoand:
> > Provide a CDI producer for Jsonb (which I also did in my PR) and just inject the Jsonb bean into the ContextResolver.
> 
> I agree with that.
> 
> > We could also provide "hooks" for users to customize the the default jsonb bean instead of replacing it entirely.
> 
> I'm thinking of 2 things:
> 
>     * discover serializers automatically via Jandex (and other things if they make sense)
> 

Great idea!

>     * have build items for extensions to push serializers
> 

Makes perfect sense

>     * then I don't know how we can allow the user to really tweak the JsonbConfig because in some cases, they might need to. Should we allow to entirely overwrite the jsonb bean via `@DefaultBean` (but that could break some extensions or the user would have to figure out what the extensions are doing)

That's a good question. IMHO, user's shouldn't be prevented from doing such things if they really know what they are doing.
Ideally they would only configure the Jsonb and not override it. If we make it easy to configure it and make it explicit in the docs that overriding could break things depending on what extensions are being used, then I think it's a good compromise. 
WDYT?

gsmet:
> If we make it easy to configure it

yeah, for me, that's the missing piece. Have you something in mind?

geoand:
> > If we make it easy to configure it
> 
> yeah, for me, that's the missing piece. Have you something in mind?

Yes. I was thinking of having something like the following:

```
public interface JsonbCustomizer {
   
    void customize(Jsonb jsonb);
}
```

Users (or extensions for that matter)  would then simply add their own `JsonbCustomizer` beans which woould be applied with something like:

```
@Singleton
public class JsonbProducer {

     @Produces
     public Jsonb jsonb(List<JsonbCustomizer> customizers) {
         Jsonb jsonb = // create a default;
         for(JsonbCustomizer customizer: customizers) {
            customizer.customize(jsonb);
         }
         return jsonb;
     }
}
```

Does that sounds reasonable?

gsmet:
You would pass `JsonbConfig` not `Jsonb` to your `JsonbCustomizer`, right?

Other than that, it looks good. And yes, we would need something similar for Jackson.

geoand:
> You would pass `JsonbConfig` not `Jsonb` to your `JsonbCustomizer`, right?

Ah yes, you're right!

geoand:
@loicmathieu WDYT? Does the proposed plan sound good to you? 

If we all agree I can cook something up (probably tomorrow)

gsmet:
BTW, I think we should merge this PR anyway. It fixes a bad behavior and, this way, we can let the other solution mature.

gsmet (approved):

geoand:
> BTW, I think we should merge this PR anyway. It fixes a bad behavior and, this way, we can let the other solution mature.

Makes sense

loicmathieu:
@gsmet @geoand there is a task on my TODO list #4193 to implements a better solution for providing Serializer/Deserializer for both JsonB and Jackson.

My first idea was to provide a CDI bean (like it is done on Jackson) that can be overriden by a user ... but this is not very convenient as what we want is to provides a default bean that the user can use to a**dd more configuration** and not **replace**, because if a user replace it, he will need to register the serializer that we register on the default one ...

Anyway, @geoand if you have some ideas around this please go ahead and keep me posted.

geoand:
@loicmathieu the solution I added above should take care of the customization vs replacement problem.

I'll go ahead and start something soon, then we can discuss on something concrete.

PR #4202, opened 2019-09-25
Add query projection for MongoDB with Panache

loicmathieu:
This PR provides query projection for MongoDB with Panache.
It offers a `project(Class)` operation on `PanacheQuery` to implements query projection based on a projection class (public fields or getters/setters).

Ping @FroMage @emmanuelbernard 

This is part of #4193

### DIFF
--- a/docs/src/main/asciidoc/mongodb-panache.adoc
+++ b/docs/src/main/asciidoc/mongodb-panache.adoc
@@ -391,27 +391,31 @@ will use the entity field name: `name`, and native query will use the MongoDB fi
 
 Query projection can be done with the `project(Class)` method on the `PanacheQuery` object that is returned by the `find()` methods.
 
-You can use it to restrict which fields will be returned by the database, the ID field will always be returned.
+You can use it to restrict which fields will be returned by the database,
+the ID field will always be returned but it's not mandatory to include it inside the projection class.
 
-For this, you need to create a class (a Pojo) that will only contains the projected fields.
-The field names, or the getters, of the projection class will be used to restrict which properties will be loaded from the database.
-If your entity uses `@BsonProperty` to map a property to a database column, the same mapping must be used inside the projection class.
+For this, you need to create a class (a Pojo) that will only contain the projected fields.
+This pojo needs to be annotated with `@ProjectionFor(Entity.class)` where `Entity` is the name of your entity class.
+The field names, or  getters, of the projection class will be used to restrict which properties will be loaded from the database.
 
 Projection can be done for both PanacheQL and native queries.
 
 [source,java]
 ----
+import io.quarkus.mongodb.panache.ProjectionFor;
+import org.bson.codecs.pojo.annotations.BsonProperty;
+
 // using public fields
+@ProjectionFor(Person.class)
 public class PersonName {
-    @BsonProperty("lastname")
     public String name;
 }
 
 // using getters
+@ProjectionFor(Person.class)
 public class PersonNameWithGetter {
     private String name;
 
-    @BsonProperty("lastname")
     public String getName(){
         return name;
     }
@@ -427,7 +431,9 @@ PanacheQuery<PersonName> query = Person.find("'status': ?1", Status.Alive).proje
 PanacheQuery<PersonName> nativeQuery = Person.find("{'status': 'ALIVE'}", Status.Alive).project(PersonName.class);
 ----
 
-TIP: Inherited fields and methods from a super type will also be used.
+TIP: Using @BsonProperty is not needed to define custom column mappings, as the mappings from the entity class will be used.
+
+TIP: You can have your projection class extends from another class. In this case, the parent class also needs to have use `@ProjectionFor` annotation.
 
 == The DAO/Repository option
 
@@ -561,4 +567,3 @@ If you define your entities in the same project where you build your Quarkus app
 If the entities come from external projects or jars, you can make sure that your jar is treated like a Quarkus application library
 by indexing it via Jandex, see link:cdi-reference#how-to-generate-a-jandex-index[How to Generate a Jandex Index] in the CDI guide.
 This will allow Quarkus to index and enhance your entities as if they were inside the current project.
-
--- a/docs/src/main/asciidoc/mongodb-panache.adoc
+++ b/docs/src/main/asciidoc/mongodb-panache.adoc
@@ -387,6 +387,48 @@ public class Person extends PanacheMongoEntity {
 Both `findByNameWithPanacheQLQuery()` and `findByNameWithNativeQuery()` methods will return the same result but query written in PanacheQL
 will use the entity field name: `name`, and native query will use the MongoDB field name: `lastname`.
 
+== Query projection
+
+Query projection can be done with the `project(Class)` method on the `PanacheQuery` object that is returned by the `find()` methods.
+
+You can use it to restrict which fields will be returned by the database, the ID field will always be returned.
+
+For this, you need to create a class (a Pojo) that will only contains the projected fields.
+The field names, or the getters, of the projection class will be used to restrict which properties will be loaded from the database.
+If your entity uses `@BsonProperty` to map a property to a database column, the same mapping must be used inside the projection class.
+
+Projection can be done for both PanacheQL and native queries.
+
+[source,java]
+----
+// using public fields
+public class PersonName {
+    @BsonProperty("lastname")
+    public String name;
+}
+
+// using getters
+public class PersonNameWithGetter {
+    private String name;
+
+    @BsonProperty("lastname")
+    public String getName(){
+        return name;
+    }
+
+    public void setName(String name){
+        this.name = name;
+    }
+}
+
+// only 'name' will be loaded from the database
+PanacheQuery<PersonName> shortQuery = Person.find("status ", Status.Alive).project(PersonName.class);
+PanacheQuery<PersonName> query = Person.find("'status': ?1", Status.Alive).project(PersonNameWithGetter.class);
+PanacheQuery<PersonName> nativeQuery = Person.find("{'status': 'ALIVE'}", Status.Alive).project(PersonName.class);
+----
+
+TIP: Inherited fields and methods from a super type will also be used.
+
 == The DAO/Repository option
 
 Repository is a very popular pattern and can be very accurate for some use case, depending on
--- a/extensions/panache/mongodb-panache/deployment/src/main/java/io/quarkus/mongodb/panache/deployment/PanacheResourceProcessor.java
+++ b/extensions/panache/mongodb-panache/deployment/src/main/java/io/quarkus/mongodb/panache/deployment/PanacheResourceProcessor.java
@@ -1,13 +1,19 @@
 package io.quarkus.mongodb.panache.deployment;
 
+import java.util.HashMap;
 import java.util.HashSet;
+import java.util.Map;
 import java.util.Set;
 
+import org.bson.codecs.pojo.annotations.BsonProperty;
 import org.bson.types.ObjectId;
+import org.jboss.jandex.AnnotationInstance;
 import org.jboss.jandex.ClassInfo;
 import org.jboss.jandex.CompositeIndex;
 import org.jboss.jandex.DotName;
+import org.jboss.jandex.FieldInfo;
 import org.jboss.jandex.Indexer;
+import org.jboss.jandex.MethodInfo;
 import org.jboss.jandex.Type;
 
 import io.quarkus.deployment.Capabilities;
@@ -27,6 +33,7 @@ import io.quarkus.mongodb.panache.PanacheMongoEntity;
 import io.quarkus.mongodb.panache.PanacheMongoEntityBase;
 import io.quarkus.mongodb.panache.PanacheMongoRepository;
 import io.quarkus.mongodb.panache.PanacheMongoRepositoryBase;
+import io.quarkus.mongodb.panache.ProjectionFor;
 import io.quarkus.panache.common.deployment.PanacheFieldAccessEnhancer;
 import io.quarkus.panache.common.deployment.PanacheRepositoryEnhancer;
 
@@ -36,7 +43,12 @@ public class PanacheResourceProcessor {
     static final DotName DOTNAME_PANACHE_ENTITY_BASE = DotName.createSimple(PanacheMongoEntityBase.class.getName());
     private static final DotName DOTNAME_PANACHE_ENTITY = DotName.createSimple(PanacheMongoEntity.class.getName());
 
+    private static final DotName DOTNAME_PROJECTION_FOR = DotName.createSimple(ProjectionFor.class.getName());
+    private static final DotName DOTNAME_BSON_PROPERTY = DotName.createSimple(BsonProperty.class.getName());
+
     private static final DotName DOTNAME_OBJECT_ID = DotName.createSimple(ObjectId.class.getName());
+
+    private static final DotName DOTNAME_OBJECT = DotName.createSimple(Object.class.getName());
 
     @BuildStep
     CapabilityBuildItem capability() {
@@ -129,6 +141,48 @@ public class PanacheResourceProcessor {
                     transformers.produce(new BytecodeTransformerBuildItem(className, panacheFieldAccessEnhancer));
                 }
             }
+        }
+
+        // manage @BsonProperty for the @ProjectionFor annotation
+        Map<ClassInfo, Map<String, String>> propertyMapping = new HashMap<>();
+        for (AnnotationInstance annotationInstance : index.getIndex().getAnnotations(DOTNAME_PROJECTION_FOR)) {
+            Type targetClass = annotationInstance.value().asClass();
+            ClassInfo target = index.getIndex().getClassByName(targetClass.name());
+            Map<String, String> classPropertyMapping = new HashMap<>();
+            extractMappings(classPropertyMapping, target, index);
+            propertyMapping.put(target, classPropertyMapping);
+        }
+        for (AnnotationInstance annotationInstance : index.getIndex().getAnnotations(DOTNAME_PROJECTION_FOR)) {
+            Type targetClass = annotationInstance.value().asClass();
+            ClassInfo target = index.getIndex().getClassByName(targetClass.name());
+            Map<String, String> targetPropertyMapping = propertyMapping.get(target);
+            if (targetPropertyMapping != null && !targetPropertyMapping.isEmpty()) {
+                ClassInfo info = annotationInstance.target().asClass();
+                ProjectionForEnhancer fieldEnhancer = new ProjectionForEnhancer(targetPropertyMapping);
+                transformers.produce(new BytecodeTransformerBuildItem(info.name().toString(), fieldEnhancer));
+            }
+        }
+    }
+
+    private void extractMappings(Map<String, String> classPropertyMapping, ClassInfo target, CombinedIndexBuildItem index) {
+        for (FieldInfo fieldInfo : target.fields()) {
+            if (fieldInfo.hasAnnotation(DOTNAME_BSON_PROPERTY)) {
+                AnnotationInstance bsonProperty = fieldInfo.annotation(DOTNAME_BSON_PROPERTY);
+                classPropertyMapping.put(fieldInfo.name(), bsonProperty.value().asString());
+            }
+        }
+        for (MethodInfo methodInfo : target.methods()) {
+            if (methodInfo.hasAnnotation(DOTNAME_BSON_PROPERTY)) {
+                AnnotationInstance bsonProperty = methodInfo.annotation(DOTNAME_BSON_PROPERTY);
+                classPropertyMapping.put(methodInfo.name(), bsonProperty.value().asString());
+            }
+        }
+
+        // climb up the hierarchy of types
+        if (!target.superClassType().name().equals(DOTNAME_OBJECT)) {
+            Type superType = target.superClassType();
+            ClassInfo superClass = index.getIndex().getClassByName(superType.name());
+            extractMappings(classPropertyMapping, superClass, index);
         }
     }
 }
--- a/extensions/panache/mongodb-panache/runtime/src/main/java/io/quarkus/mongodb/panache/runtime/MongoPropertyUtil.java
+++ b/extensions/panache/mongodb-panache/runtime/src/main/java/io/quarkus/mongodb/panache/runtime/MongoPropertyUtil.java
@@ -1,0 +1,53 @@
+package io.quarkus.mongodb.panache.runtime;
+
+import java.lang.reflect.Field;
+import java.lang.reflect.Method;
+import java.util.HashMap;
+import java.util.Map;
+
+import org.bson.codecs.pojo.annotations.BsonProperty;
+
+final class MongoPropertyUtil {
+
+    private MongoPropertyUtil() {
+        //prevent initialization
+    }
+
+    static Map<String, String> extractReplacementMap(Class<?> clazz) {
+        //TODO cache the replacement map or pre-compute it during build (using reflection or jandex)
+        Map<String, String> replacementMap = new HashMap<>();
+        for (Field field : clazz.getDeclaredFields()) {
+            BsonProperty bsonProperty = field.getAnnotation(BsonProperty.class);
+            if (bsonProperty != null) {
+                replacementMap.put(field.getName(), bsonProperty.value());
+            }
+        }
+        for (Method method : clazz.getDeclaredMethods()) {
+            if (method.getName().startsWith("get")) {
+                // we try to replace also for getter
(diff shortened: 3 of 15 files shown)


extensions/panache/mongodb-panache/runtime/src/main/java/io/quarkus/mongodb/panache/runtime/PanacheQueryImpl.java
```diff
@@ -1,10 +1,13 @@
 package io.quarkus.mongodb.panache.runtime;
 
+import java.beans.Introspector;
```
gsmet: I think we have something for that in some utils but I don't have my IDE handy, better not use `java.beans`.

FroMage:
Thanks @loicmathieu.

I wonder what we're really trying to solve here, when it comes down to the basics. If I understand correctly, projections in HQL allow us to do two things:

- Return a type which is not an entity, because we have different columns entirely. For example `select new Foo(item.bar, other.gee) from Item item left join other on item.other = other.property where item.name is like '%fubar%'`.
- Physically transmit fewer columns from the DB because we only care about one. For example `select new ShortItem(name, fabric) from Item`.

The proposed PR doesn't solve the first case, and I don't know if it solves the second, because it looks like we're not telling Mongo to return a smaller document (can we even?).

Don't take this as criticism @loicmathieu, I just want to make sure we all understand the use-cases of projection and whether we can/do solve them.

@emmanuelbernard is there a projection use-case I've missed?

loicmathieu:
> The proposed PR doesn't solve the first case, and I don't know if it solves the second, because it looks like we're not telling Mongo to return a smaller document (can we even?).

@FroMage it does send the list of fields to load to the database as a projection on the MongoDB query so yes, the second point is covered. It is on this [line](https://github.com/quarkusio/quarkus/pull/4202/files#diff-07c6705d4a6b8e340d0e01d717f8f488R134)

FroMage:
@loicmathieu ok, my mistake, you're right. So the second case is indeed covered.

loicmathieu:
@emmanuelbernard is this query projection implementation OK for you for MongoDB with Panache ?

docs/src/main/asciidoc/mongodb-panache-guide.adoc
```diff
@@ -381,6 +381,30 @@ public class Person extends PanacheMongoEntity {
 Both `findByNameWithPanacheQLQuery()` and `findByNameWithNativeQuery()` methods will return the same result but query written in PanacheQL
 will use the entity field name: `name`, and native query will use the MongoDB field name: `lastname`.
 
+== Query projection
+
+Query projection can be done with the `project(Class)` method on the `PanacheQuery` object that is returned by the `find()` method.
+You can use it to restrict which fields will be returned by the database, for this you need to create a class (a Pojo) that will only contains the projected fields.
```
jmartisk: The implementation checks that there is a getter for that field, so for clarity I would add "with the corresponding getter" or something. 
Also, contains -> contain

extensions/panache/mongodb-panache/runtime/src/main/java/io/quarkus/mongodb/panache/runtime/PanacheQueryImpl.java
```diff
@@ -37,6 +41,24 @@
 
     // Builder
 
+    @Override
+    public <T> PanacheQuery<T> project(Class<T> type) {
+        this.projections = new Document();
+        for (Method method : type.getDeclaredMethods()) {
+            if (method.getName().startsWith("get")) {
+                BsonProperty bsonProperty = method.getAnnotation(BsonProperty.class);
```
jmartisk: I think the `@BsonProperty` can be on the (Java) field as well, not just the getter

jmartisk: Ah, I see the description says this is not covered, disregard then

loicmathieu: @jmartisk I can easily support both, this is already done for the __main pojo__ so it's will be easily done for the _projected pojo__.
For the moment this is just a draft PR, if the design is accepted I will update it to also support `@BsonProperty` on fields.

docs/src/main/asciidoc/mongodb-panache-guide.adoc
loicmathieu: As public entity fields are used a lot accross Quarkus examples, maybe I also need to implements the retieval of these fields. The Pojo Codec that is used supports them so it should be done.

cescoffier:
@FroMage @jmartisk What's the status of this PR?

loicmathieu:
@cescoffier we're waiting feedback from @emmanuelbernard for the design of the projection mechanism. He was too busy making 1.0 happens :) Now that it's out maybe he will have some time to validate the design and we could go further on this ?

loicmathieu:
@cescoffier we're waiting feedback from @emmanuelbernard for the design of the projection mechanism. He was too busy making 1.0 happens :) Now that it's out maybe he will have some time to validate the design and we could go further on this ?

loicmathieu:
@emmanuelbernard can you give feedback on the design of the project method? Is a projection via interface a good choice ?

emmanuelbernard:
yes hoping to come for a review relatively soon. Bare with me.

loicmathieu:
@emmanuelbernard ping ? Will you have time soon to validate the design of this PR ?
I know that I need to review the impl but first I need to be sure that the design will be accepted ...

docs/src/main/asciidoc/mongodb-panache-guide.adoc
```diff
@@ -381,6 +381,30 @@ public class Person extends PanacheMongoEntity {
...
+    }
+
+    public void setName(String name){
+        this.name = name;
+    }
+}
+
+// only 'name' will be returned
+PanacheQuery<PersonName> query = Person.find("status ", Status.Alive).project(PersonName.class);
+PanacheQuery<PersonName> query = Person.find("{'status': ?1}", Status.Alive).project(PersonName.class);
```
emmanuelbernard: So my main question on that is 

```
public PersonView {
    public String name;
    public LocalDate birthDate;
}

@MongoEntity(collection="ThePerson")
public class Person extends PanacheMongoEntity {
    public String name;

    // will be persisted as a 'birth' field in MongoDB
    @BsonProperty("birth")
    public LocalDate birthDate;

    public Status status;
}

PanacheQuery<PersonName> query = Person.find("status ", Status.Alive).project(PersonName.class);
```

will it return `birthDate` from the `birth` column or will it crash?

emmanuelbernard (commented):
Here is my review with one main question.

Speaking of, I saw in the guide (MongoDB Panache) that we say

```


If your query does not start with {, we will consider it a PanacheQL query:

    <singleColumnName> (and single parameter) which will expand to {'singleColumnName': '?'}
```

but later

```
When you use query parameters, be careful that PanacheQL queries will refer to the Object parameters name but native queries will refer to MongoDB field names.
```

I suppose the first one is an error, right and should be `<singlePropertyName> which will expand to {'singleColumnName': '?'}` correct?
```

File: docs/src/main/asciidoc/mongodb-panache-guide.adoc (hunk 4 above)
Comment (loicmathieu) [reply docs/src/main/asciidoc/mongodb-panache-guide.adoc]: With the current implementation, it will ask mongodb to project on the `birthDate` field, as it didn't exist, it will return null (no crash).

We should be explicit on the documentation that the attributse of the projection class needs to reflect the attributes on the MongoDB document, if not `@BsonProperty` should be used.

According to me, it's logical, the same kind of mapping (field, getter/setter, annotation) is done on both the projection class or the entity class.

Comment (loicmathieu) [reply docs/src/main/asciidoc/mongodb-panache-guide.adoc]: With the current implementation, it will ask mongodb to project on the `birthDate` field, as it didn't exist, it will return null (no crash).

We should be explicit on the documentation that the attributse of the projection class needs to reflect the attributes on the MongoDB document, if not `@BsonProperty` should be used.

According to me, it's logical, the same kind of mapping (field, getter/setter, annotation) is done on both the projection class or the entity class.

**Comment by loicmathieu** [comment]
@emmanuelbernard regarding the documentation for shorthand query, you're right, they are PanacheQL queries so field replacement is done via annotation.
I'll send a small doc PR to fix this

File: extensions/panache/mongodb-panache/runtime/src/main/java/io/quarkus/mongodb/panache/runtime/PanacheQueryImpl.java (hunk 1 above)
Comment (loicmathieu) [reply extensions/panache/mongodb-panache/runtime/src/main/java/io/quarkus/mongodb/panache/runtime/PanacheQueryImpl.java]: `io.quarkus.deployment.bean.JavaBeanUtil` but as it's inside the deployment module I cannot access it at runtime.
Anyway I will review the implementation next week

File: docs/src/main/asciidoc/mongodb-panache-guide.adoc (hunk 4 above)
Comment (emmanuelbernard) [reply docs/src/main/asciidoc/mongodb-panache-guide.adoc]: Ah, ok so hear my thinking. With Panache I define entities on which a mapping is defined (with BsonPorperty and friends). And the API especially PanacheQL operate on the properties. (Not the columns unless you go for native query). 
So I anticipated project to be operating on properties too. It is essentially a short hand for return type property projection. 

In that model we would find the property metadata from entity and do the conversion to column.

**Comment by emmanuelbernard** [comment]
Back to the first comment Steph put in this PR, this one does not address the projection between different parts of the query result, it’s just a projection of the main entity type and the casting of a return type. 

There are two levels I suppose for a general solution :
1. express that you only want to return a subset of the tulle and that might include nested elements I suppose. 
2. express on which object you want this shortened and transformed tuple to be projected onto. 
3. Also make it work on a native query that returns columns

For 2, I’m feel it will need to be merged with the current exploration of a DTO binding facility that we discussed and iced. But ideally we would not need to wait for its conclusion. 
For 1 my gut feeling is that a projection query language would be the most powerful model. But verbose and redundant in the simple cases where column names and property names are 1-1. 

Your proposal has the benefit of remaining very simple (with the open question of property vs column in the other comment). Is that powerful enough that it won’t frustrate people for the 95% projection use case?

File: docs/src/main/asciidoc/mongodb-panache-guide.adoc (hunk 4 above)
Comment (loicmathieu) [reply docs/src/main/asciidoc/mongodb-panache-guide.adoc]: So you would want to use `birthDate` inside the projection class without adding any `@BsonProperty` and having it mapped to the `birth` column because the `project` operator will lookup mapping customization inside the entity and not inside the projection class ?

Thinking about it, it's more logical as the entity defines the mapping related to the database and the projection class (we could have go to a simple interface with only getters instead of a class) only serve to restrict fields. This is a more clearer separation of concerns so I'm OK with it.

I will update my implementation and the documentation accordingly.

**Comment by loicmathieu** [comment]
@emmanuelbernard projection via classes does not covers all projection cases but provides an easy way to make projection and a compile type way so I really think it adds value and can easily match 80% use case of projection (this should be the primary target of a framework : 80% use case).

We have `find(Document)` to match complex use case.

Native query should also works with projection, I will add some more tests/doc example.

File: docs/src/main/asciidoc/mongodb-panache-guide.adoc (hunk 4 above)
Comment (FroMage) [reply docs/src/main/asciidoc/mongodb-panache-guide.adoc]: +1

Comment (loicmathieu) [reply docs/src/main/asciidoc/mongodb-panache-guide.adoc]: In fact, reviewing my implementation it was the case :)

Using `@BsonProperty` inside the projection class never worked as the query is mapped to a MongoDB `Document` prior to the construction of the `PanacheQueryImpl` and this also means that property/column mapping from the entity was always used :)

**Comment by loicmathieu** [comment]
@emmanuelbernard @FroMage I updated my PR to covers your concern about the implementation and to add more documentation and tests.

File: extensions/panache/mongodb-panache/runtime/src/main/java/io/quarkus/mongodb/panache/runtime/PanacheQueryImpl.java (hunk 1 above)
Comment (loicmathieu) [reply extensions/panache/mongodb-panache/runtime/src/main/java/io/quarkus/mongodb/panache/runtime/PanacheQueryImpl.java]: I copied the code from `io.quarkus.deployment.bean.JavaBeanUtil`.
Mabe we need to move this to the core runtime module.

File: docs/src/main/asciidoc/mongodb-panache-guide.adoc (hunk 2 above)
Comment (loicmathieu) [reply docs/src/main/asciidoc/mongodb-panache-guide.adoc]: Public fields and getters/setters are now supported.

File: docs/src/main/asciidoc/mongodb-panache-guide.adoc (hunk 4 above)
Comment (loicmathieu) [reply docs/src/main/asciidoc/mongodb-panache-guide.adoc]: @emmanuelbernard @FroMage sadly this is not possible as to return a `PanacheQuery<ProjectionClass>` we need to use the `ProjectionClass` to fetch the results from the MongoDB Collection so it must include `@BsonProperty` to map the class property to the database column if needed.

So if field mapping is used, it needs to be placed in the two classes: the entity and the projection one.

I cannot think of a way to avoid this.
My implementation and the documentation now reflect this.

Comment (emmanuelbernard) [reply docs/src/main/asciidoc/mongodb-panache-guide.adoc]: Hum hold one, at runtime we don't know that this PanacheQuery object comes from a EntityA repo or active record entity?

Comment (loicmathieu) [reply docs/src/main/asciidoc/mongodb-panache-guide.adoc]: No, and we don't care as for both Entity usage (active records or repository) mapping is defined the same way: using `@BsonProperty`.
It's the same for JPA where we use `@Column` to customize the property to column mapping for both active records and repository pattern.

Comment (emmanuelbernard) [reply docs/src/main/asciidoc/mongodb-panache-guide.adoc]: you lost me. You last message seems to tell me there is no problem. in JPA you project on property name and don't need to replicate your `@Column` annotations

Comment (loicmathieu) [reply docs/src/main/asciidoc/mongodb-panache-guide.adoc]: You're right ! (I think I lost myself in projection ...). JPA has a lot of different projection capabilites, one is projection DTO and very close to what we want to achieve here and it didn't need any column mapping ...

So if we want to offer the same, I can see no other solution but to use the Entity class when loading data from the database and using a reflection mapper to map from the Entity class to the Projection class.

I don't know if we have such reflection mapper somewhere in the Quarkus code base ?

Comment (FroMage) [reply docs/src/main/asciidoc/mongodb-panache-guide.adoc]: If we know a class is a projection (with an annotation pointing to the origin class) we can copy the column mapping annotations from the origin class to the projection class during transformation.

**Comment by loicmathieu** [comment]
@FroMage I implements `@ProjectionFor` as you suggested. Can you review this part closely (the last commit), ASM is tricky and I'm not an expert of it ...

I know that some part could be optimized but for the moment I just want it to works, maybe we can have a round of optimization later.

I didn't add a check to the project method to verify that the annotation is present. 
In fact I document it inside the guide as mandatory but if no `@BsonProperty` annotation is used this is not needed (and if a user duplicate the `@BsonProperty` inside the projection class this is also not needed).

File: docs/src/main/asciidoc/mongodb-panache.adoc [hunk 5]
```diff
@@ -387,6 +387,53 @@ public class Person extends PanacheMongoEntity {
...
 will use the entity field name: `name`, and native query will use the MongoDB field name: `lastname`.
 
+== Query projection
+
+Query projection can be done with the `project(Class)` method on the `PanacheQuery` object that is returned by the `find()` methods.
+
+You can use it to restrict which fields will be returned by the database,
+the ID field will always be returned but it's not mandatory to include it inside the projection class.
+
+For this, you need to create a class (a Pojo) that will only contains the projected fields.
```
machi1990:
```suggestion
For this, you need to create a class (a Pojo) that will only contain the projected fields.
```

docs/src/main/asciidoc/mongodb-panache.adoc
```diff
@@ -387,6 +387,53 @@ public class Person extends PanacheMongoEntity {
...
+This pojo needs to be annotated with `@ProjectionFor(Entity.class)` where `Entity` is the name of your entity class.
+The field names, or the getters, of the projection class will be used to restrict which properties will be loaded from the database.
```
machi1990:
```suggestion
The field names, or  getters, of the projection class will be used to restrict which properties will be loaded from the database.
```

docs/src/main/asciidoc/mongodb-panache.adoc
```diff
@@ -387,6 +387,53 @@ public class Person extends PanacheMongoEntity {
...
+    }
+}
+
+// only 'name' will be loaded from the database
+PanacheQuery<PersonName> shortQuery = Person.find("status ", Status.Alive).project(PersonName.class);
+PanacheQuery<PersonName> query = Person.find("'status': ?1", Status.Alive).project(PersonNameWithGetter.class);
+PanacheQuery<PersonName> nativeQuery = Person.find("{'status': 'ALIVE'}", Status.Alive).project(PersonName.class);
+----
+
+TIP: You can see that there is no need to use `@BsonProperty` annotations to define custom column mappings.
```
machi1990: I have re-read it a couple of times this tip and I find it hard to follow. Is there some ay we could re-phrase it? 

Otherwise, really great stuff. Thanks.

extensions/panache/mongodb-panache/runtime/src/main/java/io/quarkus/mongodb/panache/runtime/MongoPropertyUtil.java
```diff
@@ -0,0 +1,53 @@
...
+import org.bson.codecs.pojo.annotations.BsonProperty;
+
+final class MongoPropertyUtil {
+
+    private MongoPropertyUtil() {
+        //prevent initialization
+    }
+
+    static Map<String, String> extractReplacementMap(Class<?> clazz) {
+        //TODO cache the replacement map or pre-compute it during build (using reflection or jandex)
```
machi1990: +1

docs/src/main/asciidoc/mongodb-panache.adoc
loicmathieu: > Using `@BsonProperty` is not needed to define custom column mappings, as the mappings from the entity class will be used.

This is clearer and more detailed

machi1990: Go for it.

loicmathieu: updated

extensions/panache/mongodb-panache/runtime/src/main/java/io/quarkus/mongodb/panache/runtime/PanacheQueryImpl.java
loicmathieu: This is now supported on both fields and methods

loicmathieu:
@FroMage would you have time to check this PR (and with special attention to the last commit as I'm not an ASM expert) ?

extensions/panache/mongodb-panache/deployment/src/main/java/io/quarkus/mongodb/panache/deployment/PanacheResourceProcessor.java
```diff
@@ -130,5 +140,39 @@ void build(CombinedIndexBuildItem index,
                 }
             }
         }
+
+        // manage @BsonProperty for the @ProjectionFor annotation
+        Map<String, String> propertyMapping = new HashMap<>();
+        for (AnnotationInstance annotationInstance : index.getIndex().getAnnotations(DOTNAME_PROJECTION_FOR)) {
+            Type targetClass = annotationInstance.value().asClass();
+            ClassInfo target = index.getIndex().getClassByName(targetClass.name());
+            for (FieldInfo fieldInfo : target.fields()) {
```
FroMage: Aren't we already collecting all model info? Why do you have to rescan fields and methods? Won't this miss superclass properties?

extensions/panache/mongodb-panache/deployment/src/main/java/io/quarkus/mongodb/panache/deployment/PanacheResourceProcessor.java
```diff
@@ -130,5 +140,39 @@ void build(CombinedIndexBuildItem index,
...
+                if (fieldInfo.hasAnnotation(DOTNAME_BSON_PROPERTY)) {
+                    AnnotationInstance bsonProperty = fieldInfo.annotation(DOTNAME_BSON_PROPERTY);
+                    propertyMapping.put(fieldInfo.name(), bsonProperty.value().asString());
```
FroMage: This appears to map properties by name, but this can't work if we have more than one entity with the same field names, no?

extensions/panache/mongodb-panache/deployment/src/main/java/io/quarkus/mongodb/panache/deployment/PanacheResourceProcessor.java
```diff
@@ -130,5 +140,39 @@ void build(CombinedIndexBuildItem index,
...
+            }
+            for (MethodInfo methodInfo : target.methods()) {
+                if (methodInfo.hasAnnotation(DOTNAME_BSON_PROPERTY)) {
+                    AnnotationInstance bsonProperty = methodInfo.annotation(DOTNAME_BSON_PROPERTY);
+                    propertyMapping.put(methodInfo.name(), bsonProperty.value().asString());
+                }
+            }
+        }
+        for (AnnotationInstance annotationInstance : index.getIndex().getAnnotations(DOTNAME_PROJECTION_FOR)) {
+            ClassInfo info = annotationInstance.target().asClass();
```
FroMage: I don't think you need to register a transformer per field/method, especially given that each transformer works for the same sets of mappings. Also your transformer can do the transformation for both fields and methods in one go.

extensions/panache/mongodb-panache/deployment/src/main/java/io/quarkus/mongodb/panache/deployment/PanacheResourceProcessor.java
```diff
@@ -130,5 +140,39 @@ void build(CombinedIndexBuildItem index,
...
+            for (FieldInfo fieldInfo : info.fields()) {
```
FroMage: Does this support inheritance? Should it? If not, it should be documented.

FroMage (changes requested):

extensions/panache/mongodb-panache/deployment/src/main/java/io/quarkus/mongodb/panache/deployment/PanacheResourceProcessor.java
loicmathieu: Yes, the model info is available inside the `PanacheMongoRepositoryEnhancer` or the `PanacheMongoEntityEnhancer` but here we only get the fields annotated by `@BsonProperty` from the class denoted by the `@ProjectionFor` annotation. 
This is simpler to do it this way than searching on the two enhancers, then filter the fields.

It uses the same method from Jandex to gather the fields than the enhancer uses, but yes I think it will miss the fields from the super classes if any.

extensions/panache/mongodb-panache/deployment/src/main/java/io/quarkus/mongodb/panache/deployment/PanacheResourceProcessor.java
loicmathieu: Yes, silly bug :(

extensions/panache/mongodb-panache/deployment/src/main/java/io/quarkus/mongodb/panache/deployment/PanacheResourceProcessor.java
loicmathieu: I think inheritance will works as long as the user also annotate the parent class with `@ProjectionFor`. This should be tested ...

loicmathieu: I added a line inside the guide about inheritance

extensions/panache/mongodb-panache/deployment/src/main/java/io/quarkus/mongodb/panache/deployment/PanacheResourceProcessor.java
loicmathieu: fixed

extensions/panache/mongodb-panache/deployment/src/main/java/io/quarkus/mongodb/panache/deployment/PanacheResourceProcessor.java
loicmathieu: You're right, the transformer can be reuse for all fields/methods.
I also merge the two transformers. Code is more readable :)

loicmathieu:
@FroMage I think I cover all your concerns. Is it OK for you now ?

FroMage (approved):
Yup, LGTM, thanks.

PR #4210, opened 2019-09-26
Tests that static resources are served from a sub directory

stuartwdouglas:
Tests that static resources are served from a sub directory


cescoffier:
Should we add the same test for when the static files are served by the Vert.x body handler?

mkouba:
@cescoffier Feel to add one in a separate PR. I'm going to merge this one.

mkouba (approved):

stuartwdouglas:
Ideally we would have tests for both

PR #4212, opened 2019-09-26
Allow easy customization of Jackson’s ObjectMapper and JsonbConfig

geoand:


geoand:
@loicmathieu @gsmet This is the (initial) result of our discussion on #4196. 

I currently haven't replaced the Mongo Panache stuff (although I did test a patch version of it) because that deployment part of that extension does not depend on either jackson or jsonb meaning that I can't really generate the build items.
How would you like to handle this case (and similar ones in the future I guess). Should we have some kind of generic json-spi which would be used to register serializers / deserializers (and I guess the concepto of module which Jackson to group them has but Jsonb doesn't)?

loicmathieu:
@geoand I don't understand why you have both a JsonbCustomizer and a JsonbSerializerBuildItem ? Both allow to register a serializer right ?

For the integration with mongodb-panache, it depends of what will be used. 
- For customizer : using an optional dependency and providing a CDI bean if the capability is present would works (or even producing the bean always and depending on Arc bean removal ?).
- For the JsonbSerializerBuildItem it must be placed on an SPI module.

We can't have a generic API for both JsonB and Jackson as the serializer API are different, for this we will need to create an abstraction layer and I don't think it is worth it. As a framework, we should provide serializer for both !

For JsonB, you can easily test if it works, as the existing mongodb-panache integration test will not works if the serializer are not included. For Jackson I have a local project that uses it so I can test it.

I can take care of the mongodb-panache stuff if you want, just tell me :)

geoand:
> @geoand I don't understand why you have both a JsonbCustomizer and a JsonbSerializerBuildItem ? Both allow to register a serializer right ?
> 

The `JsonbCustomizer` is a CDI bean that allows for arbitrary customization.
`JsonbSerializerBuildItem` is aimed exclusively at extensions that want to register and JsonbSerializer - all they would do is create it and everything else would be taken care of by Quarkus.

> For the integration with mongodb-panache, it depends of what will be used.
> 
> * For customizer : using an optional dependency and providing a CDI bean if the capability is present would works (or even producing the bean always and depending on Arc bean removal ?).
> * For the JsonbSerializerBuildItem it must be placed on an SPI module.
> 
> We can't have a generic API for both JsonB and Jackson as the serializer API are different, for this we will need to create an abstraction layer and I don't think it is worth it. As a framework, we should provide serializer for both !
> 

I wasn't proposing a general API for serializers / deserializers. What I am trying to figure out is how to make it super easy for extensions (like Mongo Panache currently) to register there necessary serializers / deserializers.
The problem is that for example Mongo Panache does not depend on either Jackson or Jsonb so we can't just produce a specific build item for that.

So what I am thinking is this:

* Create a json-spi module that will contain a `JsonSerializerDeserializerBuildItem`.
This will contain the names of serializers / deserializers, a module name (which will only be used in Jackson) as well as a type enum - whose values will be jsonb and jackson.

Does that sounds reasoable @loicmathieu, @gsmet ?

> For JsonB, you can easily test if it works, as the existing mongodb-panache integration test will not works if the serializer are not included. For Jackson I have a local project that uses it so I can test it.
> 

Yeah, I already did it and it worked :).

> I can take care of the mongodb-panache stuff if you want, just tell me :)

No need, I have it covered. But thanks :)

loicmathieu:
@geoand mongodb-panache do have dependencies to jsonb and jackson via optional dependencies for `quarkus-resteasy-jsonb` and `quarkus-resteasy-jackson` inside its runtime module.

If you want both customizer and SerializerBuildItem to facilite the work of extensions that only want to provide serializer, I think the best is to create specific BuildItem for JsonB and Jackson not a generic one. And those needs to be inside SPI modules to minimize the dependencies.

There is currently no such things as **generic json** support in Quarkus, only JsonB or Jackson support.

For Jackson modules, I think it should be a separate BuildItem. I thinks specific BuildItems are better because it is compile-checked and more expressive when using them.

Speaking of Jackson modules, maybe we should register some of them by default (and provides a boolean to backoff from it) as almost everybody will enabled the ones related to Java 8.

> This will contain the names of serializers / deserializers

Why not using the class instead of the name ? Again compile-checked :) This is what I did from the BuildItem to register HealthCheck [here](https://github.com/quarkusio/quarkus/pull/3671/files#diff-2001e390234e7756606d18b873d135fbR5)

geoand:
> @geoand mongodb-panache do have dependencies to jsonb and jackson via optional dependencies for `quarkus-resteasy-jsonb` and `quarkus-resteasy-jackson` inside its runtime module.
> 

But no such dependency exists in the deployment module so generating specific BuildItems isn't possible.

> If you want both customizer and SerializerBuildItem to facilite the work of extensions that only want to provide serializer, I think the best is to create specific BuildItem for JsonB and Jackson not a generic one. And those needs to be inside SPI modules to minimize the dependencies.
> 
> There is currently no such things as **generic json** support in Quarkus, only JsonB or Jackson support.
> 
> For Jackson modules, I think it should be a separate BuildItem. I thinks specific BuildItems are better because it is compile-checked and more expressive when using them.
> 
> Speaking of Jackson modules, maybe we should register some of them by default (and provides a boolean to backoff from it) as almost everybody will enabled the ones related to Java 8.
> 
> > This will contain the names of serializers / deserializers
> 
> Why not using the class instead of the name ? Again compile-checked :) This is what I did from the BuildItem to register HealthCheck [here](https://github.com/quarkusio/quarkus/pull/3671/files#diff-2001e390234e7756606d18b873d135fbR5)

I understand your points. I don't necessarily share the concerns however :).

In any case in a few minutes I'll update the PR with what I mean

gsmet:
Does having a generic `json-spi` really works?

I would have thought that we would need a `jsonb-spi` and a `jackson-spi`.

geoand:
> Does having a generic `json-spi` really works?
> 
> I would have thought that we would need a `jsonb-spi` and a `jackson-spi`.

Let me finish up what I have in mind and show you. If all comes down to whether we think the deployment modules of extensions should depend on jsonb and jackson.

loicmathieu:
> But no such dependency exists in the deployment module so generating specific BuildItems isn't possible.

Maven optional dependencies are magical :) 

As deployement depends on runtime, even if it's optional on runtime it works as the serializer will be inside the runtime artifacts and produced as a bean inside the deployement one (if capability is present) as it is done today with `PanacheMongoJsonbContextResolver` and `ObjectMapperProducer` that resides inside the runtime module and used inside the deployment one.

geoand:
I pushed what I had in mind.

If you both think that having an optional dependency on both `jsonb-deployment` and `jackson-deployment` is the way to go, I'm cool with it and I'll oblidge.

The rest of the PR should be agnostic of this fact - the ideas should be applicable regardless of what spi strategy we use.

extensions/resteasy-jackson/runtime/src/main/java/io/quarkus/resteasy/jackson/runtime/QuarkusObjectMapperContextResolver.java
```diff
@@ -0,0 +1,25 @@
...
+
+@Provider
+@Priority(Priorities.USER + 10) // give it a priority that ensures that user supplied ContextResolver classes override this one
+public class QuarkusObjectMapperContextResolver implements ContextResolver<ObjectMapper> {
+
+    private ObjectMapper objectMapper;
+
+    @Override
+    public ObjectMapper getContext(Class<?> type) {
+        if (objectMapper == null) {
```
loicmathieu: I think this is not thread-safe, a begning race condition can occurs here.
Maybe getting the bean from Arc at postConstruct will be better ?

extensions/resteasy-jsonb/runtime/src/main/java/io/quarkus/resteasy/jsonb/runtime/QuarkusJsonbContextResolver.java
```diff
@@ -0,0 +1,24 @@
...
+
+@Provider
+@Priority(Priorities.USER + 10) // give it a priority that ensures that user supplied ContextResolver classes override this one
+public class QuarkusJsonbContextResolver implements ContextResolver<Jsonb> {
+
+    private Jsonb jsonb;
+
+    @Override
+    public Jsonb getContext(Class<?> type) {
+        if (null == jsonb) {
```
loicmathieu: I think this is not thread-safe, a begning race condition can occurs here.
Maybe getting the bean from Arc at postConstruct will be better ?

loicmathieu (changes requested):
As I said, I will favor separating Jackson and JsonB in two spi modules.
Some extensions relies on Jackson only, and they could takes advantage of this someday.

I would also provides separate BuildItem for ser/der and Jackson modules.

Regarding using  `Class<?>` or the name of the class as a `String`, I saw that the `String` flavor is more used accross Quarkus. Is there any reason why ? I'm just curious as I prefere compile safety but you guys should have a good reason to prefere `String` :)

loicmathieu:
@geoand I didn't find where your are creating a `ObjectMapperCustomizer` that uses the registered serializer to include them in the Jackson `ObjectMapper`. Is already covered ?

geoand:
> @geoand I didn't find where your are creating a `ObjectMapperCustomizer` that uses the registered serializer to include them in the Jackson `ObjectMapper`. Is already covered ?

Nope, I just wanted to validate the approach with you folks before moving onto that

extensions/resteasy-jsonb/runtime/src/main/java/io/quarkus/resteasy/jsonb/runtime/QuarkusJsonbContextResolver.java
geoand: I don't think it really matters to be honest. The returned object will always be the same, so even if multiple threads access it, at works they will both go to Arc and return the same object.

extensions/resteasy-jackson/runtime/src/main/java/io/quarkus/resteasy/jackson/runtime/QuarkusObjectMapperContextResolver.java
geoand: I don't think it really matters to be honest. The returned object will always be the same, so even if multiple threads access it, at works they will both go to Arc and return the same object.

geoand:
> As I said, I will favor separating Jackson and JsonB in two spi modules.
> Some extensions relies on Jackson only, and they could takes advantage of this someday.
> 
> I would also provides separate BuildItem for ser/der and Jackson modules.
> 

Let's let @gsmet have the final say in this :)

> Regarding using `Class<?>` or the name of the class as a `String`, I saw that the `String` flavor is more used accross Quarkus. Is there any reason why ? I'm just curious as I prefere compile safety but you guys should have a good reason to prefere `String` :)

I would assume the reason is that it keeps the build process clean of loaded classes and classloader hell.

extensions/resteasy-jsonb/runtime/src/main/java/io/quarkus/resteasy/jsonb/runtime/QuarkusJsonbContextResolver.java
gsmet: I'm not totally sure it will be safe. Standard injection does not work in `ContextResolver`? I'm wondering if you could just inject the instance via `@Inject`?

geoand: That was the first I tried and it didn't work

geoand: But let me give it another shot

geoand: It doesn't work. Do we have some way of making providers beans?

geoand: I'll dig into this some more and see if we can do something

gsmet: I thought they were to be honest.

Another option would be to initialize it in the constructor if it's not too early.

geoand: It shouldn't be too early. The only reason I wanted to do it was to avoid any unecessary initialization at startup.

If I can't figure out why the Providers aren't actually injecting (which TBH before doing this excersise I also thought were capable of injection), that's what I'll end up doing

loicmathieu:
> I would assume the reason is that it keeps the build process clean of loaded classes and classloader hell.

And this is a good reason :)

extensions/resteasy-jackson/runtime/src/main/java/io/quarkus/resteasy/jackson/runtime/QuarkusObjectMapperContextResolver.java
loicmathieu: If I remember well there is a slight chance of double-null reads/writes that may ends up on some NullPointerException (like one times in a million ...).
Long story: https://shipilev.net/blog/2014/safe-public-construction/ :)

extensions/resteasy-jsonb/runtime/src/main/java/io/quarkus/resteasy/jsonb/runtime/QuarkusJsonbContextResolver.java
loicmathieu: I'm surprised that using `@PostConstruct` didn't works. It should allow to add init blocks after the bean factory was initalized ... maybe a bug in Arc ?

geoand:
It seems like there is something funky going on with the `ContextResolver` classes I have introduced. They definitely **can** be beans (I checked in another project) so I just need to pinpoint why they are not being included here. 
But I think it should be easy to overcome the problem and just use `@Inject` and be done with it

geoand:
I found that `@Providers` added in an application are registered as beans. However then ones in an extensions aren't registered for some reason - I still haven't figured out why

gsmet:
My guess would be that they are not indexed so we don’t make them beans?

geoand:
> My guess would be that they are not indexed so we don’t make them beans?

That's what I'm thinking as well so we'll just have to register them manually

geoand:
So I had forgot to register it as an AdditionalBean :man_facepalming:

geoand:
`ContextResolver` injections are resolved.

Now all that remains is to decide on the the maven module structure @gsmet :)

geoand:
The tests are now failing due to #2773. 
Not really sure what we should to proceed since the way things are currently done, the providers are common between rest client and server. 

@gsmet WDYT?

mkouba:
@gsmet I don't think that provider instances are shared between rest client and server. At least it seems that each one is using a different `InjectorFactory`. The set of provider classes is probably shared the instances are not.

geoand:
So I guess the proper solution would be to make the client use the same `InjectorFactory`?

mkouba:
Yep, I think so but I'm no expert in this area ;-).

geoand:
OK, then let's wait to see what @gsmet thinks

gsmet:
@geoand two things:
- I think I would be more comfortable with having 2 different SPIs, especially since there is a good chance we will want to push different things at some point;
- I'm not an expert in RESTEasy either but, yeah, I think they should use the same injector factory e.g. the Quarkus one. Not using it in the client case looks like a bug to me.

geoand:
> @geoand two things:
> 
> * I think I would be more comfortable with having 2 different SPI modules (one for JSON-B and one for Jackson), especially since there is a good chance we will want to push different things at some point;

So you are OK with having optional dependencies `jackson` and `json` in extensions like mongo-panache?
Just checking to make sure we are on the same page

> * I'm not an expert in RESTEasy either but, yeah, I think they should use the same injector factory e.g. the Quarkus one. Not using it in the client case looks like a bug to me.

I can take a look at it but I don't know much about the rest client part.

gsmet:
> So you are OK with having optional dependencies jackson and json in extensions like mongo-panache?

Well, if we want to support both, there's no other option. Having a common one is a bit too risky IMHO.

> I can take a look at it but I don't know much about the rest client part.

@mkouba provided stacktraces in the other issue so we just need to find out where this injector factory comes from and figure out how to inject our own and ping Alessio if it's an issue in RESTEasy. Please ping me if you need some help with that or you want to discuss your findings!

geoand:
I commented on #2773 about the rest-client issue. I found the root cause and have a dirty solution. Next stop a proper solution :)

geoand:
I ran into what I was afraid would happen... 

By making the `quarkus-json-deployment` and `quarkus-jackson-deployment` dependencies `optional`, their BuildSteps always run, whether the application including `quarkus-mongodb-panache` includes jackson/jsonb or not... This of course creates a ton of problems

loicmathieu:
@geoand 
> By making the quarkus-json-deployment and quarkus-jackson-deployment dependencies optional in quarkus-mongodb-panache, their BuildSteps always run, whether the application including quarkus-mongodb-panache includes jackson/jsonb or not... This of course creates a ton of problems

This is not what I have in mind, optional dependencies should be on the runtime dependencies (like today) and you should put the new BuildItem on new SPI modules.
The trick is to emit those build items only if the capability is present.

I remember spending a lot of time to make it works in mongo-panach ...

geoand:
OK, so I think we are going in circles here....

Are we actually saying that we want the SPI module that this PR adds but instead of having a common BuildItem, to have two separate BuildItems one for Jackson and one for Jsonb?

geoand:
So this was my dumb mistake.
As pointed out by @gsmet all that needs to be done in addition to what I described above is to move the new build items to an spi module and have `quarkus-mongodb-panache-deployment` depend on those

geoand:
I now have this working the way it's supposed to :).
Once #4234 is merged (since it blocks this), I'll update the PR and make it ready for review

extensions/resteasy-jackson/runtime/src/main/java/io/quarkus/resteasy/jackson/runtime/QuarkusObjectMapperContextResolver.java
geoand: Cool link, thanks!

extensions/resteasy-jsonb/runtime/src/main/java/io/quarkus/resteasy/jsonb/runtime/QuarkusJsonbContextResolver.java
geoand: So now everything should work as expected. I had missed adding the providers as `AdditionalBean`s

extensions/jackson/deployment/src/main/java/io/quarkus/jackson/deployment/JacksonProcessor.java
```diff
@@ -81,4 +103,80 @@ private void addReflectiveHierarchyClass(DotName className) {
...
+        try (ClassCreator classCreator = ClassCreator.builder().classOutput(classOutput)
+                .className("io.quarkus.jackson.customizer.RegisterSerializersAndDeserializersCustomizer")
+                .interfaces(ObjectMapperCustomizer.class.getName())
+                .build()) {
+            classCreator.addAnnotation(Singleton.class);
+
+            try (MethodCreator customize = classCreator.getMethodCreator("customize", void.class, ObjectMapper.class)) {
+                ResultHandle objectMapper = customize.getMethodParam(0);
+
+                for (CustomJacksonSerDeserBuildItem customSerDeser : customSerDesers) {
```
gsmet: I don't think I would have the `Custom` in the name.

I wonder if it's such a good idea to couple the serializer and deserializer entirely. Maybe I would have had two different build items.

BTW, if we do that, I would name them `JacksonSerializerBuildItem` and `JacksonDeserializerBuildItem` which is consistent with the Jackson naming.

extensions/jackson/deployment/src/main/java/io/quarkus/jackson/deployment/JacksonProcessor.java
```diff
@@ -81,4 +103,80 @@ private void addReflectiveHierarchyClass(DotName className) {
...
+                    if (customSerDeser.getItems().isEmpty()) {
```
gsmet: I also wonder if we should use Jandex to discover the ones declared in the application automatically?

What I did in HV is that I detect them automatically if they are beans.

extensions/jackson/spi/pom.xml
```diff
@@ -0,0 +1,22 @@
...
+    <parent>
+        <artifactId>quarkus-jackson-parent</artifactId>
+        <groupId>io.quarkus</groupId>
+        <version>999-SNAPSHOT</version>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>quarkus-jackson-spi</artifactId>
+    <name>Quarkus - Jackson - SPI</name>
+    <description>Extension that provided Build Items specific to Jackson</description>
```
gsmet:
```suggestion
    <description>Extension that provides BuildItems specific to Jackson</description>
```

extensions/jackson/spi/src/main/java/io/quarkus/jackson/spi/CustomJacksonSerDeserBuildItem.java
```diff
@@ -0,0 +1,67 @@
...
+
+import io.quarkus.builder.item.MultiBuildItem;
+
+/**
+ * BuildItem that will be used to customize the ObjectMapper (produced by io.quarkus.jackson.ObjectMapperProducer)
+ * with custom serializers and/or deserializers
+ *
+ * Serializers and deserializers MUST contain a public no-args constructor
+ */
+public final class CustomJacksonSerDeserBuildItem extends MultiBuildItem {
```
gsmet: So reading that, I wonder if it should be `JacksonModuleBuildItem` instead and have proper methods to add serializer/deserializers to the module.

gsmet (dismissed):
So... I'm a bit torn of this one, I had mixed feelings while reading it: it does the job but I wonder if we should simplify that. My comments are not all consistent as I thought of a few different things while reading it. I kept them as they might have value if we decide to go one way or another.

I wonder if we should just have `JsonbCustomizerBuildItem`, `JacksonCustomizerBuildItem` that would push customizers to the build and create beans for them.

It might sound more low level but I *think* in the end, it's better for the user/extension developer because they end up with a well-known API, either Jackson or JSON-B. Might be better than have a Quarkus-specific one, that is different for each so you would have to understand the specifics anyway.

It's really open to discussion as both approaches have advantages and drawbacks so let's discuss it :).

extensions/jackson/deployment/src/main/java/io/quarkus/jackson/deployment/JacksonProcessor.java
geoand: You don't have to set both, you can easily just define one.

The "problem" with Jackson is that the serializers and deserializers need to be part of a module so having one build item makes it easier to define such a module.

I'm of course open to do whatever we think is best.

extensions/jackson/deployment/src/main/java/io/quarkus/jackson/deployment/JacksonProcessor.java
geoand: That's a very interesting point.

It would work fine for jsonb but probably be tricky for Jackson since as I mentioned earlier the serializers and deserializers need to be part of a module.
We could or course add a one with our own name.

I'm not sure what is best... Having BuildItems or Beans as the starting point...

extensions/jackson/spi/src/main/java/io/quarkus/jackson/spi/CustomJacksonSerDeserBuildItem.java
geoand: That sounds like an excellent idea!

geoand:
@gsmet you raised some excellent points!

I propose we sleep on this over the weekend and take it up further on Monday

geoand:
@gsmet can you perhaps expand on your CustomizerBuildItem idea?
Are you thinking that the extension would provide or create a class then the Jackson extension would just register it as a bean?
If so, that does sound interesting.
I still think however that serializer/deserializer build item is a more common need.

extensions/jackson/spi/src/main/java/io/quarkus/jackson/spi/CustomJacksonSerDeserBuildItem.java
geoand: Actually, do we have any examples of making the BuildItem mutable? I don't recall seeing that being done before.

extensions/jackson/deployment/src/main/java/io/quarkus/jackson/deployment/JacksonProcessor.java
geoand: I dropped `Custom` from the name

geoand:
Bases on a conversation with @gsmet we decided to do the following:

* Provide JsonbCustomizerBuildItem / JacksonCustomizerBuildItem - people will have the usual JSON-B/Jackson API
* Make existing build items better by having: `JacksonModuleBuildItem` for Jackson `JsonbSerializerBuildItem`/`JsonbDeserializerBuildItem` for Jsonb.

geoand:
Still missing documentation on the various available options. I'll add that soon

geoand:
Documentation about configuration options added/

extensions/jackson/deployment/src/main/java/io/quarkus/jackson/deployment/JacksonProcessor.java
```diff
@@ -81,4 +103,80 @@ private void addReflectiveHierarchyClass(DotName className) {
...
+                     * SimpleModule module = new SimpleModule("somename");
+                     * module.addSerializer(Foo.class, new FooSerializer());
+                     * module.addDeserializer(Foo.class, new FooDeserializer());
+                     * objectMapper.registerModule(module);
+                     */
+                    ResultHandle module = customize.newInstance(
+                            MethodDescriptor.ofConstructor(SimpleModule.class, String.class),
+                            customize.load(jacksonModule.getName()));
+
+                    for (JacksonModuleBuildItem.Item item : jacksonModule.getItems()) {
```
gsmet: Why don't you have a serializer list and then a deserializer list in your module?

That would simplify the code a lot and be more readable. I might be missing something obvious though.

extensions/jackson/runtime/src/main/java/io/quarkus/jackson/ObjectMapperCustomizer.java
```diff
@@ -0,0 +1,8 @@
+package io.quarkus.jackson;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+public interface ObjectMapperCustomizer {
```
gsmet: This is API, it needs some javadoc.

extensions/jackson/spi/src/main/java/io/quarkus/jackson/spi/JacksonModuleBuildItem.java
```diff
@@ -0,0 +1,101 @@
...
+            }
+            return new JacksonModuleBuildItem(this.name, items);
+        }
+    }
+
+    // This is needed because Jackson can't register a deserializer without the target type
+    public static class Item {
+        private final String targetClassName;
+        private final String serializerClassName;
+        private final String deserializerClassNames;
```
gsmet: I don't think I would do that. I would separate entirely serializers and deserializers.

extensions/jsonb/deployment/src/main/java/io/quarkus/jsonb/deployment/JsonbProcessor.java
```diff
@@ -47,4 +69,75 @@ public boolean test(BeanInfo bean) {
...
+        if (serializers.isEmpty()) {
+            return;
+        }
+
+        final Set<String> customSerializerClasses = new HashSet<>();
+        final Set<String> customDeserializerClasses = new HashSet<>();
+        for (JsonbSerializerBuildItem serializer : serializers) {
+            customSerializerClasses.addAll(serializer.getSerializerClassNames());
+        }
+        for (JsonbDeserializerBuildItem deserializer : desrializers) {
```
gsmet:
```suggestion
        for (JsonbDeserializerBuildItem deserializer : deserializers) {
```
(obviously, better do that with a refactor ;))

extensions/jsonb/spi/pom.xml
```diff
@@ -0,0 +1,22 @@
...
+    <parent>
+        <artifactId>quarkus-jsonb-parent</artifactId>
+        <groupId>io.quarkus</groupId>
+        <version>999-SNAPSHOT</version>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>quarkus-jsonb-spi</artifactId>
+    <name>Quarkus - JSON-B - SPI</name>
+    <description>Extension that provided Build Items specific to JSONB</description>
```
gsmet:
```suggestion
    <description>Artifact that provides BuildItems specific to JSON-B</description>
```

extensions/jackson/spi/pom.xml
```diff
@@ -0,0 +1,22 @@
...
+    <parent>
+        <artifactId>quarkus-jackson-parent</artifactId>
+        <groupId>io.quarkus</groupId>
+        <version>999-SNAPSHOT</version>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>quarkus-jackson-spi</artifactId>
+    <name>Quarkus - Jackson - SPI</name>
+    <description>Extension that provides BuildItems specific to Jackson</description>
```
gsmet:
```suggestion
    <description>Artifact that provides BuildItems specific to Jackson</description>
```

extensions/jsonb/runtime/src/main/java/io/quarkus/jsonb/JsonbConfigCustomizer.java
```diff
@@ -0,0 +1,8 @@
+package io.quarkus.jsonb;
+
+import javax.json.bind.JsonbConfig;
+
+public interface JsonbConfigCustomizer {
```
gsmet: This is API, it needs some documentation.

gsmet (changes requested):
Nice work but I shouldn't be surprised :).

I added a few comments.

geoand:
> Nice work but I shouldn't be surprised :).
> 
 
It has a ton of your input so it was bound to turn out well :)

I'll address the comments ASAP

extensions/jackson/spi/src/main/java/io/quarkus/jackson/spi/JacksonModuleBuildItem.java
geoand: That was my initial thought too. 
However it's very common that you would have the same `targetClass` and have both a serializer and a deserializer. So I think this modeling fits the use pattern best. WDYT?

extensions/jackson/deployment/src/main/java/io/quarkus/jackson/deployment/JacksonProcessor.java
geoand: In Jackson you need to specify the class that the serializer and deserializer are "targeting". So with `JacksonModuleBuildItem.Item` the targetClass and serializer/deserializer are grouped together.

Basically what the loop generates is something like:

```
SimpleModule module = new SimpleModule("somename");
module.addSerializer(Foo.class, new FooSerializer());
module.addDeserializer(Foo.class, new FooDeserializer());
objectMapper.registerModule(module);
```

Does it make more sense now?

extensions/jsonb/deployment/src/main/java/io/quarkus/jsonb/deployment/JsonbProcessor.java
geoand: I'm very surprised my IDE didn't warn me... I want my money back :)

geoand:
@gsmet I pushed pushed for most of the comment and responded with my view on a couple other ones.

extensions/jsonb/runtime/src/main/java/io/quarkus/jsonb/JsonbConfigCustomizer.java
```diff
@@ -0,0 +1,15 @@
+package io.quarkus.jsonb;
+
+import javax.json.bind.JsonbConfig;
+
+/**
+ * Meant to be implemented by a CDI bean that provided arbitrary customization for the default {@link JsonbConfig}
```
gsmet:
```suggestion
 * Meant to be implemented by a CDI bean that provides arbitrary customization for the default {@link JsonbConfig}.
 * <p>
```

extensions/jsonb/runtime/src/main/java/io/quarkus/jsonb/JsonbConfigCustomizer.java
```diff
@@ -0,0 +1,15 @@
...
+ * All implementations (that are registered as CDI beans) are taken into account when producing the default {@link JsonbConfig}.
+ * The {@link JsonbConfig} is in turn used to produce the default {@link javax.json.bind.Jsonb}
```
gsmet:
```suggestion
 * The {@link JsonbConfig} is in turn used to produce the default {@link javax.json.bind.Jsonb}.
```

extensions/jsonb/runtime/src/main/java/io/quarkus/jsonb/JsonbConfigCustomizer.java
```diff
@@ -0,0 +1,15 @@
...
+ *
+ * See also {@link JsonbProducer#jsonbConfig}
```
gsmet:
```suggestion
 * See also {@link JsonbProducer#jsonbConfig}.
```

extensions/jsonb/runtime/src/main/java/io/quarkus/jsonb/JsonbConfigCustomizer.java
```diff
@@ -0,0 +1,15 @@
+package io.quarkus.jsonb;
+
+import javax.json.bind.JsonbConfig;
+
+/**
+ * Meant to be implemented by a CDI bean that provided arbitrary customization for the default {@link JsonbConfig}
+ * All implementations (that are registered as CDI beans) are taken into account when producing the default {@link JsonbConfig}.
+ * The {@link JsonbConfig} is in turn used to produce the default {@link javax.json.bind.Jsonb}
+ *
```
gsmet:
```suggestion
 * <p>
```

extensions/jackson/spi/src/main/java/io/quarkus/jackson/spi/JacksonModuleBuildItem.java
```diff
@@ -0,0 +1,101 @@
...
+                throw new IllegalArgumentException("targetClassName cannot be null or empty");
+            }
+            this.targetClassName = targetClassName;
+        }
+
+        public String getSerializerClassName() {
+            return serializerClassName;
+        }
+
+        public String getDeserializerClassNames() {
```
gsmet: This should probably be singular now?

extensions/jackson/spi/src/main/java/io/quarkus/jackson/spi/JacksonModuleBuildItem.java
```diff
@@ -0,0 +1,101 @@
...
+            return deserializerClassNames;
```
gsmet: Same here.

extensions/jackson/spi/src/main/java/io/quarkus/jackson/spi/JacksonModuleBuildItem.java
```diff
@@ -0,0 +1,101 @@
...
+        }
+    }
+
+    // This is needed because Jackson can't register a deserializer without the target type
+    public static class Item {
+        private final String targetClassName;
+        private final String serializerClassName;
+        private final String deserializerClassNames;
+
+        public Item(String targetClassName, String serializerClassName, String deserializerClassNames) {
```
gsmet: The last one should probably be singular?

extensions/jackson/spi/src/main/java/io/quarkus/jackson/spi/JacksonModuleBuildItem.java
gsmet: OK, let's leave at that, both have values and it's purely internal.

extensions/jackson/runtime/src/main/java/io/quarkus/jackson/ObjectMapperCustomizer.java
```diff
@@ -0,0 +1,15 @@
+package io.quarkus.jackson;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+/**
+ * Meant to be implemented by a CDI bean that provided arbitrary customization for the default {@link ObjectMapper}
```
gsmet:
```suggestion
 * Meant to be implemented by a CDI bean that provides arbitrary customization for the default {@link ObjectMapper}.
 * <p>
```

extensions/jackson/runtime/src/main/java/io/quarkus/jackson/ObjectMapperCustomizer.java
```diff
@@ -0,0 +1,15 @@
...
+ * All implementations (that are registered as CDI beans) are taken into account when producing the default
+ * {@link ObjectMapper}.
+ *
```
gsmet:
```suggestion
 * <p>
```

extensions/jackson/runtime/src/main/java/io/quarkus/jackson/ObjectMapperCustomizer.java
```diff
@@ -0,0 +1,15 @@
...
+ * See also {@link ObjectMapperProducer#objectMapper}
```
gsmet:
```suggestion
 * See also {@link ObjectMapperProducer#objectMapper}.
```

docs/src/main/asciidoc/rest-json-guide.adoc
```diff
@@ -186,28 +186,61 @@ They are analyzed at build time and Quarkus restricts the number of JAX-RS provi
...
+==== JSON-B
+
+Quarkus makes it very easy to configure various JSON-B settings via CDI beans. The simplest (and suggested) approach is to define a CDI bean of type `io.quarkus.jsonb.JsonbConfigCustomizer`
+inside of which any JSON-B configuration can be applied.
+
+If for example a custom serializer named `FooSerializer` for type `com.example.Foo` needs to be registered with JSON-B, the addition of a bean like the following would suffice:
+
+[source,java]
+----
+
```
gsmet: You can remove this empty line.

gsmet (changes requested):
I added a couple very minor additional comments. Let's fix them and it will be good to go IMHO.

Nice to have this one tackled.

extensions/jackson/spi/src/main/java/io/quarkus/jackson/spi/JacksonModuleBuildItem.java
geoand: Definitely!

geoand:
@gsmet Thanks for spotting the mistakes! They should all be addressed now.

gsmet (approved):

gsmet:
I asked for a rerun of the failed check.

geoand:
> I asked for a rerun of the failed check.

OK thanks!

extensions/jackson/deployment/src/main/java/io/quarkus/jackson/deployment/JacksonProcessor.java
```diff
@@ -81,4 +103,80 @@ private void addReflectiveHierarchyClass(DotName className) {
...
+        ClassOutput classOutput = new ClassOutput() {
+            @Override
+            public void write(String name, byte[] data) {
+                generatedBeans.produce(new GeneratedBeanBuildItem(name, data));
+            }
+        };
+
+        try (ClassCreator classCreator = ClassCreator.builder().classOutput(classOutput)
+                .className("io.quarkus.jackson.customizer.RegisterSerializersAndDeserializersCustomizer")
+                .interfaces(ObjectMapperCustomizer.class.getName())
```
loicmathieu: Why do you use full String for `className` and the class for `interfaces` (in which you call `getName`).
Maybe use String for both will be more concistent.
As I understood, it is favored in build steps to avoid unecessary class loading.

extensions/jsonb/deployment/src/main/java/io/quarkus/jsonb/deployment/JsonbProcessor.java
```diff
@@ -47,4 +69,75 @@ public boolean test(BeanInfo bean) {
...
+        ClassOutput classOutput = new ClassOutput() {
+            @Override
+            public void write(String name, byte[] data) {
+                generatedBeans.produce(new GeneratedBeanBuildItem(name, data));
+            }
+        };
+
+        try (ClassCreator classCreator = ClassCreator.builder().classOutput(classOutput)
+                .className("io.quarkus.jsonb.customizer.RegisterSerializersAndDeserializersCustomizer")
+                .interfaces(JsonbConfigCustomizer.class.getName())
```
loicmathieu: Same note here, why use string and JsonbConfigCustomizer.class.gerName() ?

extensions/panache/mongodb-panache/deployment/src/main/java/io/quarkus/mongodb/panache/deployment/PanacheResourceProcessor.java
```diff
@@ -43,19 +41,22 @@ FeatureBuildItem featureBuildItem() {
...
+            BuildProducer<JsonbDeserializerBuildItem> jsonbDeserializers) {
+        jsonbSerializers
+                .produce(new JsonbSerializerBuildItem(io.quarkus.mongodb.panache.jsonb.ObjectIdSerializer.class.getName()));
+        jsonbDeserializers
+                .produce(new JsonbDeserializerBuildItem(io.quarkus.mongodb.panache.jsonb.ObjectIdDeserializer.class.getName()));
     }
 
     @BuildStep
-    ResteasyJaxrsProviderBuildItem registerJsonbSerDer(Capabilities capabilities) {
-        if (capabilities.isCapabilityPresent("io.quarkus.resteasy.jsonb")) {
```
loicmathieu: You remove the check to the capability presence, did you check that an apps without jsonb and jackson still works ?
I remind having classpath issue without this check

loicmathieu (approved):
I have a look and this looks very good :)
I add some questions but nothing important.
Maybe a adding a section on the extension author guide explaining how to add Jsonb/Jackson serializers will be good. Maybe this can be used by other extensions to provides serializer/deserializer (like for reactive-messaging to automatically serialize to json)

extensions/jackson/deployment/src/main/java/io/quarkus/jackson/deployment/JacksonProcessor.java
geoand: Good question. In this setting (strictly inside a build step) using the class doesn't cause problems. Moving classes around via Build Items can however.
This sort of pattern is used throughout Quarkus

extensions/jsonb/deployment/src/main/java/io/quarkus/jsonb/deployment/JsonbProcessor.java
geoand: Same answer :)

extensions/panache/mongodb-panache/deployment/src/main/java/io/quarkus/mongodb/panache/deployment/PanacheResourceProcessor.java
geoand: Yup. This now works because of the build items being in the SPI module (which has no dependency on the actual jackson/jsonb module). So even if Jsob and / Jackson are not present, the SPI build items will just be dropped on the floor. 
This sort of pattern is very flexible and has been used in other parts of Quarkus as well (see the Kubernetes or Undertow extensions) for example.

geoand:
> I have a look and this looks very good :)
> I add some questions but nothing important.
> Maybe a adding a section on the extension author guide explaining how to add Jsonb/Jackson serializers will be good. Maybe this can be used by other extensions to provides serializer/deserializer (like for reactive-messaging to automatically serialize to json)

Can I interest you in adding such documentation? :)
You had to see that one coming :)

gsmet:
Really glad we have this sorted out.

geoand:
Yeah, things should now make sense :). Thanks for your invaluable help @gsmet @loicmathieu !

extensions/jackson/deployment/src/main/java/io/quarkus/jackson/deployment/JacksonProcessor.java
loicmathieu: Thanks for the explaintation. A new day, a new learning :)

extensions/panache/mongodb-panache/deployment/src/main/java/io/quarkus/mongodb/panache/deployment/PanacheResourceProcessor.java
loicmathieu: OK, thanks for the explaintation

extensions/jackson/deployment/src/main/java/io/quarkus/jackson/deployment/JacksonProcessor.java
geoand: Glad to help anytime :)

loicmathieu:
> > I have a look and this looks very good :)
> > I add some questions but nothing important.
> > Maybe a adding a section on the extension author guide explaining how to add Jsonb/Jackson serializers will be good. Maybe this can be used by other extensions to provides serializer/deserializer (like for reactive-messaging to automatically serialize to json)
> 
> Can I interest you in adding such documentation?
> You had to see that one coming :)

@geoand of course ;)
Can you create an issue and assign it to me (I cannot assign issue), I will take car of it but not soon (I'll be on PTO for a week then busy on other stuff).

geoand:
> > > I have a look and this looks very good :)
> > > I add some questions but nothing important.
> > > Maybe a adding a section on the extension author guide explaining how to add Jsonb/Jackson serializers will be good. Maybe this can be used by other extensions to provides serializer/deserializer (like for reactive-messaging to automatically serialize to json)
> > 
> > 
> > Can I interest you in adding such documentation?
> > You had to see that one coming :)
> 
> @geoand of course ;)
> Can you create an issue and assign it to me (I cannot assign issue), I will take car of it but not soon (I'll be on PTO for a week then busy on other stuff).

Cool, thanks! [Here](https://github.com/quarkusio/quarkus/issues/4339) is the issue.

PR #4222, opened 2019-09-26
The SmallRye JWT extension depends on Undertow.

cescoffier:
It fixes the JWT quickstart.

### DIFF
--- a/extensions/smallrye-jwt/deployment/pom.xml
+++ b/extensions/smallrye-jwt/deployment/pom.xml
@@ -24,10 +24,6 @@
         </dependency>
         <dependency>
             <groupId>io.quarkus</groupId>
-            <artifactId>quarkus-vertx-web-deployment</artifactId>
-        </dependency>
-        <dependency>
-            <groupId>io.quarkus</groupId>
             <artifactId>quarkus-arc-deployment</artifactId>
         </dependency>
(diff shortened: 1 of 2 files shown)


cescoffier:
The dependency comes from SmallRye JWT.

stuartwdouglas (approved):

sberyozkin:
@stuartwdouglas @cescoffier this will disconnect smallrye-jwt from the vertx-based OIDC ? Or is it done until smallrye-jwt extension is rebuilt not to use Undertow ?

sberyozkin:
I've referred to it at #4226

gsmet:
Same question as Sergey. Not very excited about this change. 

We can’t make it work with vert.x?

cescoffier:
The new security architecture won't use SmallRye JWT. JWT is going to be handled by the quarkus security components. 

If you want to make SmallRye JWT work with the new architecture (which was not the goal) you would need to rewrite it.

stuartwdouglas:
Note that you can still inject stuff as per the MP JWT spec at the moment when using Vert.x Keycloak, it just does not use Smallrye to verify the token.

cescoffier:
Will try to find another way to make the quickstart pass without these changes. 
Might need a complete rewrite.

PR #4223, opened 2019-09-26
Allow additional classes to be easily registered for reflection

stuartwdouglas:


mkouba (approved):
Looks very similar to: https://github.com/quarkusio/qute/blob/master/core/src/main/java/io/quarkus/qute/TemplateData.java#L28 ;-)

PR #4234, opened 2019-09-27
Ensure that resteasy rest-client uses QuarkusInjectorFactory as well

### LINKED ISSUE #2773: NullPointerException using @Provider in MP Rest Client
**Describe the bug**
@Inject doesn't work in @Provider when the Provider is used by MP Rest Client

**Expected behavior**
The Bean injected

**Actual behavior**
The Bean field is null

**To Reproduce**
1. git clone https://github.com/masini/quarkus-rest-client-npe
2. mvn package
3. Server Test Works, Client Test Fails

geoand:
This is done in order to be able to use `@Provider` annotated `ContextResolver` classes as beans - same as is the case for RESTEasy server extensions.

Fixes #2773


geoand:
The only thing I am a little concerned about is setting `QuarkusInjectorFactory.CONTAINER` twice (if the rest client extension is in play). I don't think it will ever be a problem but if you think we should add some synchronization, then I can of course do that

extensions/rest-client/runtime/src/main/java/io/quarkus/restclient/runtime/RestClientRecorder.java
```diff
@@ -24,8 +30,12 @@ public void setSslEnabled(boolean sslEnabled) {
         RestClientBuilderImpl.setSslEnabled(sslEnabled);
     }
 
-    public void initializeResteasyProviderFactory(boolean useBuiltIn, Set<String> providersToRegister,
+    public void initializeResteasyProviderFactory(BeanContainer beanContainer,
+            List<Function<Object, Object>> propertyUnwrappers, boolean useBuiltIn,
+            Set<String> providersToRegister,
             Set<String> contributedProviders) {
+        QuarkusInjectorFactory.CONTAINER = beanContainer;
```
mkouba: We should probably move this (and the code from `ResteasyServerCommonRecorder#setupIntegration()`) to the `ResteasyCommonProcessor`. Actually, we should probably move the `ResteasyServerCommonProcessor#setupInjection()` method and this build step should depend on `ResteasyInjectionReadyBuildItem`, or?

geoand: Interesting idea, I like it. @gsmet do you agree?

mkouba: In fact, `ResteasyInjectionReadyBuildItem` could hold a `RuntimeValue<org.jboss.resteasy.spi.InjectorFactory>` reference and we could keep the `QuarkusInjectorFactory` fields package-private.

geoand: Should be fixed now

gsmet (changes requested):
It looks good to me, nice work.

I just had a comment about a comment where you totally changed the meaning of the initial wording, and the new meaning is IMHO incorrect. Can you check?

extensions/resteasy-common/deployment/src/main/java/io/quarkus/resteasy/common/deployment/ResteasyInjectionReadyBuildItem.java
```diff
@@ -0,0 +1,23 @@
+package io.quarkus.resteasy.common.deployment;
+
+import org.jboss.resteasy.spi.InjectorFactory;
+
+import io.quarkus.builder.item.SimpleBuildItem;
+import io.quarkus.runtime.RuntimeValue;
+
+/**
+ * Gives access to the configured {@link InjectorFactory}
+ * Can also be used as a marker that RESTEasy has been properly setup and is ready for use
```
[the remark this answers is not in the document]
geoand: Of course yes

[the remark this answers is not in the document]
geoand: Fixed

gsmet (approved):

gsmet:
It's OK for me. @mkouba I let you review the latest changes and approve too?

mkouba (approved):
@geoand That's exactly what I meant. Good job!

geoand:
Thanks for your help @mkouba and @gsmet !

PR #4236, opened 2019-09-27
Attempt to initialize the buffer pool in a lazy manner

stuartwdouglas:


stuartwdouglas:
Merging, as this was merged into 0.23

PR #4240, opened 2019-09-27
Avoid contention on lookup of UserTransaction and TransactionManager

### LINKED ISSUE #4239: High contention on @Transactional annotated methods
The interceptor we use to implement the `@Transactional` does a lookup of the TransactionManager on each need.

Narayana implements this lookup with a synchronized method; however it should be totally fine to cache this lookup even in a constant.

Sanne:
I'm experiencing contention on the underlying lookup of the TXM when using the `@Transactional` annottion on business methods.

This removes the contention, and simplify the code too.

Fixes #4239

### DIFF
--- a/extensions/narayana-jta/runtime/src/main/java/io/quarkus/narayana/jta/runtime/NarayanaJtaProducers.java
+++ b/extensions/narayana-jta/runtime/src/main/java/io/quarkus/narayana/jta/runtime/NarayanaJtaProducers.java
@@ -18,6 +18,9 @@ import com.arjuna.ats.jta.UserTransaction;
 @Dependent
 public class NarayanaJtaProducers {
 
+    private static final javax.transaction.UserTransaction USER_TRANSACTION = UserTransaction.userTransaction();
+    private static final javax.transaction.TransactionManager TRANSACTION_MANAGER = TransactionManager.transactionManager();
+
     @Produces
     @ApplicationScoped
     public UserTransactionRegistry userTransactionRegistry() {
@@ -27,14 +30,13 @@ public class NarayanaJtaProducers {
     @Produces
     @ApplicationScoped
     public javax.transaction.UserTransaction userTransaction() {
-        return UserTransaction.userTransaction();
+        return USER_TRANSACTION;
     }
 
     @Produces
     @ApplicationScoped
     public XAResourceRecoveryRegistry xaResourceRecoveryRegistry() {
         return new RecoveryManagerService();
-
     }
 
     @Produces
@@ -46,7 +48,7 @@ public class NarayanaJtaProducers {
     @Produces
     @ApplicationScoped
     public javax.transaction.TransactionManager transactionManager() {
-        return TransactionManager.transactionManager();
+        return TRANSACTION_MANAGER;
     }
 
     @Produces


stuartwdouglas (approved):

manovotn:
@Sanne have you debugged how often were those producers called? They are `@ApplicationScoped`, so in theory, it should only be invoked once and CDI should effectively do the "caching" by keeping the scope alive and passing around proxies.

Sanne:
good point @manovotn - and thanks for checking I'm not that handy with DI

I'll need to revert it locally and re-run my stressor, pretty sure I was having contention on the lookup but I don't remember the callstack. I'll paste it here later.

Sanne:
@manovotn , @gsmet sorry for the delay, I was away for 2 days and after it actually tooke a while to be able to reproduce this again.

So this is the stacktrace I could reproduce - when going back to the commit *just before this PR*:

```
Stack Trace	Count	Duration
com.arjuna.ats.jta.TransactionManager.transactionManager()	3	10,536,900
   io.quarkus.narayana.jta.runtime.interceptor.TransactionalInterceptorBase.invokeInOurTx(InvocationContext, TransactionManager, RunnableWithException)	3	10,536,900
      io.quarkus.narayana.jta.runtime.interceptor.TransactionalInterceptorBase.invokeInOurTx(InvocationContext, TransactionManager)	3	10,536,900
         io.quarkus.narayana.jta.runtime.interceptor.TransactionalInterceptorRequired.doIntercept(TransactionManager, Transaction, InvocationContext)	3	10,536,900
            io.quarkus.narayana.jta.runtime.interceptor.TransactionalInterceptorBase.intercept(InvocationContext)	3	10,536,900
               io.quarkus.narayana.jta.runtime.interceptor.TransactionalInterceptorRequired.intercept(InvocationContext)	3	10,536,900
                  io.quarkus.narayana.jta.runtime.interceptor.TransactionalInterceptorRequired_Bean.intercept(InterceptionType, Object, InvocationContext)	3	10,536,900
                     io.quarkus.arc.InvocationContextImpl$InterceptorInvocation.invoke(InvocationContext)	3	10,536,900
                        io.quarkus.arc.InvocationContextImpl.invokeNext()	3	10,536,900
                           io.quarkus.arc.InvocationContextImpl.proceed()	3	10,536,900
                              io.quarkus.benchmark.resource.DbResource_Subclass.updates(String)	3	10,536,900
                                 io.quarkus.benchmark.resource.DbResource_ClientProxy.updates(String)	3	10,536,900
```

### However

if I simply revert my PR on top of current master, this contention is no longer visible.

My guess is that there was indeed some other problem and it was fixed in the meantime?
Maybe @mkouba knows?

Either way, looks like we're good now. And I'd not revert this PR as it simplifies the code a bit anyway.

manovotn:
Interesting, I know of no change that would magically fix that.

> Either way, looks like we're good now. And I'd not revert this PR as it simplifies the code a bit anyway.

Agreed, certainly does no harm.

mkouba:
> They are @ApplicationScoped, so in theory, it should only be invoked once...

@manovotn exactly!

> So this is the stacktrace I could reproduce...

Hm, I'd guess it's this line instead:
https://github.com/quarkusio/quarkus/blob/master/extensions/narayana-jta/runtime/src/main/java/io/quarkus/narayana/jta/runtime/interceptor/TransactionalInterceptorBase.java#L100

Anyway, both producers are `@ApplicationScoped` and thus client proxies are injected. Each invocation upon a client proxy has a tiny overhead (we need to perform a bean lookup). If you're sure the undelying instance is not going to change I'd recommend to unwrap the contextual instance in the `TransactionalInterceptorBase`, ie. something like `this.transactionManager = ((ClientProxy) transactionManager).arc_contextualInstance()`.

manovotn:
> Hm, I'd guess it's this line instead: ...

That shouldn't cause the contention though.

> If you're sure the undelying instance is not going to change I'd recommend to unwrap the contextual instance

Or perhaps use `@Singleton` to avoid having proxies but I don't think that is a bottleneck.

mkouba:
> That shouldn't cause the contention though.

I'm just saying that this is the one from Sanne's stacktrace...

> Or perhaps use @Singleton...

I'd rather keep `@ApplicationScoped` for compatibility with user apps etc.

PR #4242, opened 2019-09-27
Fix 100% CPU in epoll_wait

### LINKED ISSUE #4241: 100% CPU in epoll_wait
**Describe the bug**
In some environments, using native image, all threads reading I/O can spend 100% CPU time polling `epoll_wait(..., timeout=0)`. Turns out this is an effect of `static final io.netty.util.concurrent.ScheduledFutureTask.START_TIME` being set to system-dependent `System.nanoTime()` during compilation and the NioEventLoop behaving in an unexpected way.

Note that this behaviour cannot be reproduced when compiling and running on the same machine, due to the nature of `System.nanoTime()` base value.

rvansa:
Fixes #4241

### DIFF
--- a/extensions/netty/runtime/src/main/java/io/quarkus/netty/runtime/graal/NettySubstitutions.java
+++ b/extensions/netty/runtime/src/main/java/io/quarkus/netty/runtime/graal/NettySubstitutions.java
@@ -12,6 +12,7 @@ import javax.net.ssl.SSLException;
 import javax.net.ssl.TrustManagerFactory;
 
 import com.oracle.svm.core.annotate.Alias;
+import com.oracle.svm.core.annotate.Delete;
 import com.oracle.svm.core.annotate.RecomputeFieldValue;
 import com.oracle.svm.core.annotate.Substitute;
 import com.oracle.svm.core.annotate.TargetClass;
@@ -331,6 +332,33 @@ final class Target_io_netty_util_AbstractReferenceCounted {
     @Alias
     @RecomputeFieldValue(kind = RecomputeFieldValue.Kind.FieldOffset, name = "refCnt")
     private static long REFCNT_FIELD_OFFSET;
+}
+
+// This class is runtime-initialized by NettyProcessor
+final class Holder_io_netty_util_concurrent_ScheduledFutureTask {
+    static final long START_TIME = System.nanoTime();
+}
+
+@TargetClass(className = "io.netty.util.concurrent.ScheduledFutureTask")
+final class Target_io_netty_util_concurrent_ScheduledFutureTask {
+    @Delete
+    public static long START_TIME = 0;
+
+    @Substitute
+    static long nanoTime() {
+        return System.nanoTime() - Holder_io_netty_util_concurrent_ScheduledFutureTask.START_TIME;
+    }
+
+    @Alias
+    public long deadlineNanos() {
+        return 0;
+    }
+
+    @Substitute
+    public long delayNanos(long currentTimeNanos) {
+        return Math.max(0,
+                deadlineNanos() - (currentTimeNanos - Holder_io_netty_util_concurrent_ScheduledFutureTask.START_TIME));
+    }
 }
 
 class NettySubstitutions {
--- a/extensions/netty/deployment/src/main/java/io/quarkus/netty/deployment/NettyProcessor.java
+++ b/extensions/netty/deployment/src/main/java/io/quarkus/netty/deployment/NettyProcessor.java
@@ -18,6 +18,7 @@ import io.quarkus.deployment.annotations.Record;
 import io.quarkus.deployment.builditem.JniBuildItem;
 import io.quarkus.deployment.builditem.SystemPropertyBuildItem;
 import io.quarkus.deployment.builditem.substrate.ReflectiveClassBuildItem;
+import io.quarkus.deployment.builditem.substrate.RuntimeReinitializedClassBuildItem;
 import io.quarkus.deployment.builditem.substrate.SubstrateConfigBuildItem;
 import io.quarkus.deployment.builditem.substrate.SubstrateSystemPropertyBuildItem;
 import io.quarkus.netty.BossEventLoopGroup;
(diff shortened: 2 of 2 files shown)


extensions/netty/runtime/src/main/java/io/quarkus/netty/runtime/graal/NettySubstitutions.java
```diff
@@ -284,6 +284,37 @@ final ChannelFuture initAndRegister() {
     }
 }
 
+final class Holder_io_netty_util_concurrent_ScheduledFutureTask {
+    private static long START_TIME = Long.MIN_VALUE;
```
dmlloyd: This `MIN_VALUE` trick is not correct.  In fact there are no special sentinel values possible for `System.nanoTime()`.  The field should be initialized directly in its declaration.  To handle the native image case, the field should be reinitialized in the substitution or the corresponding class should be marked as a runtime-initialization class.

dmlloyd: Additionally, the field should be `final`.

rvansa: @dmlloyd Yes, there are no sentinel values. However if the first invocation (that should initialize it) gets called when `System.nanoTime()` returns exactly the `Long.MIN_VALUE`, this only means that it gets initialized to a bit higher value sometimes later. Double initialization in this case causes no harm. Another way to get double initialization is when this gets called concurrently - one of the writes eventually wins but generally we don't care.

About the `final` - so if I declare `static final` field here next to the substitution class, will that get really called by the first time when we access the class? Because otherwise we would be back at the original problem when the `System.nanoTime()` value was compiled as constant into the native image.

dmlloyd: If you declare it final, it will be called when the class is initialized.  In JVM mode, that will be on first access.  In native mode, it's either build time or run time depending on config.  Alternatively, you can create a substitution just for a field that causes just that field to be recomputed at native image start, which overall is a much simpler solution.

rvansa: @dmlloyd Okay, I may have misunderstood `RecomputeFieldValue` javadoc, thinking that it is still recomputed when the image is built - so this computation happens when the native image starts?

dmlloyd: Correct.  The simplest mode is `FromAlias` where it just reruns the same initialization.  I believe this works by analyzing the compiled `<clinit>` method and tracing the data dependencies of the assignment, possibly using similar logic to the logic that ensures that every static final field is assigned exactly once.

(As you likely know, all static field initializations and static init blocks are rolled into a single, hidden `<clinit>` method by `javac`.)

rvansa: So I've tried to use just this simple substitution:
```java
@TargetClass(className = "io.netty.util.concurrent.ScheduledFutureTask")
final class Target_io_netty_util_concurrent_ScheduledFutureTask {
    @Alias
    @RecomputeFieldValue(kind = RecomputeFieldValue.Kind.FromAlias, isFinal = true)
    @Substitute
    private static long START_TIME = System.nanoTime();
}
```
and it didn't do the trick - CPU got to 100%. I'll try to remove `@Substitute` if that helps, but my undestanding is probably still incorrect. I should really [RTFM](https://quarkus.io/guides/extension-authors-guide).

rvansa: Didn't help.

stuartwdouglas: I don't think this will work, the aliased field will still be initialized at build time, so will have the same value. If you want to use RecomputeFieldValue I think you need to use 'Custom' and then provide a CustomFieldValueComputer that returns the time.

rvansa: Ok, I've tried the `CustomFieldValueComputer` and besides the nasty system-path dependency to `GRAALVM_HOME/jre/lib/jvmci/jvmci-api.jar` it doesn't work. It seemed to me that actually the cleanest way would be
```java
    @BuildStep
    public RuntimeReinitializedClassBuildItem reinitScheduledFutureTask() {
        return new RuntimeReinitializedClassBuildItem(io.netty.util.concurrent.ScheduledFutureTask.class.getName());
    }
```
but for some reason this doesn't do the trick either - is there a way to verify that this is being applied?

dmlloyd: The `FromAlias` won't work if the original field value was constant, so both the original *and* the recompute alias have to be assigned to `nanoTime()`.  In fact if you have access to the original, you can just put the annotation right on there IIRC.

rvansa: @dmlloyd Do you mean constant as `static final` or in the sense of compile-time-constant? Because it's the former, set to `System.nanoTime()`:
https://github.com/netty/netty/blob/4.1/common/src/main/java/io/netty/util/concurrent/ScheduledFutureTask.java#L29
and the code excerpt using `@FromAlias` above is setting that as well.

dmlloyd: I meant compile time constant.  But if that's already the case then I guess that wasn't the issue.

I am very curious to work out why this didn't work in this case.  Maybe because it's a primitive value?

rvansa: @dmlloyd I've updated the PR as per your suggestion.

dmlloyd (approved):

rvansa:
Rebased due to trivial conflict.

Issue #4009, opened 2019-09-13
native-image - static resources in sub-directories


haraldatbmw:
**Describe the bug**
My static HTML web-resources are located inside the directory META-INF/resources. My JavaScript, CSS files and images are in sub-directories (e.g. META-INF/resources/images, META-INF/resources/scripts, META-INF/resources/styles). 

If I create a native image only the files directly inside the META-INF/resources directory are included not the sub-directories and its files.

**Expected behavior**
sub-directories of META-INF/resources are included in the native image too.

**Actual behavior**
only files directly in META-INF/resources directory are included.

**To Reproduce**
1. Create a project with static web-content located in META-INF/resources and in sub-directories
2. Build and run a native image
3. Try to access the static resource inside the sub-directory

**Workaround**
Add additional build args to the native-image build.

```xml
<plugin>
  <groupId>io.quarkus</groupId>
  <artifactId>quarkus-maven-plugin</artifactId>
  <version>${quarkus.version}</version>
  <executions>
    <execution>
      <goals>
        <goal>native-image</goal>
      </goals>
      <configuration>
        <additionalBuildArgs>
          <additionalBuildArg>-H:IncludeResources=META-INF/resources/.*</additionalBuildArg>
        </additionalBuildArgs>
...
```

geoand:
Hi,

Can you please try:

```
<additionalBuildArg>-H:IncludeResources=META-INF/resources/**</additionalBuildArg>
```

I haven't tried it, but it might work.

In any case this isn't a Quarkus bug. This parameter is passed straight to GraalVM native-image.

haraldatbmw:
Hi,

it already works with this argument:
`<additionalBuildArg>-H:IncludeResources=META-INF/resources/.*</additionalBuildArg>`

I think it should be the default behaviour of the native image builder that sub-directories are part of the image.

geoand:
Oh, sorry I misread your issue :)

I think this makes sense

emmanuelbernard:
CC @stuartwdouglas but feel free to delegate that one

stuartwdouglas:
Can you provide a reproducer? I just tested this locally (see linked PR) and it worked for me.

haraldatbmw:
> Can you provide a reproducer? I just tested this locally (see linked PR) and it worked for me.

Did you test against a native image?

stuartwdouglas:
Yes, those tests are run in native mode

haraldatbmw:
[...]

PR #4244, opened 2019-09-27
Consider the filesystem file name separator while trying to resolve META-INF/resources directory

jaikiran:
An attempt to reproduce/verify/fix the issue reported in https://github.com/quarkusio/quarkus/issues/4009

### DIFF
--- a/extensions/undertow/deployment/src/main/java/io/quarkus/undertow/deployment/devmode/UndertowHotReplacementSetup.java
+++ b/extensions/undertow/deployment/src/main/java/io/quarkus/undertow/deployment/devmode/UndertowHotReplacementSetup.java
@@ -10,14 +10,16 @@ import io.quarkus.deployment.devmode.HotReplacementSetup;
 import io.quarkus.undertow.runtime.UndertowDeploymentRecorder;
 
 public class UndertowHotReplacementSetup implements HotReplacementSetup {
-
+    // TODO: This is marked as protected but I don't see anyone else (other than this class) using it.
+    // get rid of this if not used anywhere else or at least make it private
     protected static final String META_INF_SERVICES = "META-INF/resources";
 
     @Override
     public void setupHotDeployment(HotReplacementContext context) {
         List<Path> resources = new ArrayList<>();
         for (Path i : context.getResourcesDir()) {
-            Path resolved = i.resolve(META_INF_SERVICES);
+            final String nameSeparator = i.getFileSystem().getSeparator();
+            Path resolved = i.resolve("META-INF" + nameSeparator + "resources");
             if (Files.exists(resolved)) {
                 resources.add(resolved);
             }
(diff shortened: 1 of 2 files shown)


extensions/undertow/deployment/src/main/java/io/quarkus/undertow/deployment/devmode/UndertowHotReplacementSetup.java
```diff
@@ -10,14 +10,16 @@
...
+    // TODO: This is marked as protected but I don't see anyone else (other than this class) using it.
+    // get rid of this if not used anywhere else or at least make it private
     protected static final String META_INF_SERVICES = "META-INF/resources";
 
     @Override
     public void setupHotDeployment(HotReplacementContext context) {
         List<Path> resources = new ArrayList<>();
         for (Path i : context.getResourcesDir()) {
-            Path resolved = i.resolve(META_INF_SERVICES);
+            final String nameSeparator = i.getFileSystem().getSeparator();
```
gsmet: I agree this is theoretically more correct but we use `File.separator` everywhere in Quarkus so we could use that in the constant instead.

jaikiran: That (existing) constant actually is getting used in multiple different semantical ways - one for filesystem resource resolution and the other is to load resources from a classloader (which mandates the usage of `/` irrespective of the runtime). I can create a new constant solely for this purpose and then set its value to use `File.separator`. However, the reason I decided to go this approach is, I'm not yet 100% sure how the existing tests (one of which Stuart added recently exactly for this issue) are passing on Windows OS. Perhaps those test setup have mounted filesystem whose name separator is `/` ? That's why I decided to use the name separator that is more relevant/dynamic based on the `Path`'s `FileSystem` rather than the runtime OS. I'm open to changing this, if you think we should use `File.separator`.

stuartwdouglas: The existing tests are not run on windows. The native image tests are only run on Linux

jaikiran: OK, so that makes sense then. @stuartwdouglas, @gsmet do you want me to use `File.separator` instead of this current change? Let me know, I will update this PR.

stuartwdouglas:
I don't think this will fix the issue. The 'hot deployment' part only affects dev mode, so it will not be called for native compilation, and the application archives will be ZipFileSystem, which afaik always use "/" as the seperator. 

Its possible that for windows we may need to replace the '/' with a '\' in SubstrateResourceBuildItem, but if that is the case it will affect all substrate resources (and I would ideally like confirmation that this is actually the problem before merging the change)

jaikiran:
> and the application archives will be ZipFileSystem, which afaik always use "/" as the seperator.

You are right indeed https://github.com/openjdk/jdk/blob/master/src/jdk.zipfs/share/classes/jdk/nio/zipfs/ZipFileSystem.java#L280

stuartwdouglas:
@jaikiran are you planning on trying a different approach or should this be closed?

jaikiran:
> are you planning on trying a different approach or should this be closed?
I'm experimenting but I don't yet have a Windows OS at hand to be sure that whatever I do is the right thing. For now, I'm going to close this, since based on what you noted, the theory on which I had made this change no longer holds good. I'll either reopen or create a new one if/when I find something relevant.

PR #4246, opened 2019-09-27
Use a substitution to fix nanotime issue

### LINKED ISSUE #4241: 100% CPU in epoll_wait

stuartwdouglas:
Fixes #4241


stuartwdouglas:
Oops, I did not notice that #4242 had already been filed

PR #4252, opened 2019-09-28
Fix order based test failures in hibernate-orm/deployment module

jaikiran:
This is a potential fix for the issue reported by Bill in the quarkus-dev mailing list. The complete and lengthy details of this change is available in the thread here https://groups.google.com/d/msg/quarkus-dev/9mZDyCgrQO0/ndrBXgeeCAAJ


jaikiran:
After seeing a failure in `HotReloadTest` in the previous version of this PR, I have now updated the PR to do this in a different manner.
I now have a separate cleanup task which specifically is meant to cleanup the router during shutdown. This task gets registered as soon as a new router gets created in `initializeRouter` instead of waiting for `startServer` to be called and registering it there. This way, if the `startServer` never gets called (when application deployment fails), the cleanup start will still be run for the router

stuartwdouglas (approved):

stuartwdouglas:
LGTM, Thanks

PR #4258, opened 2019-09-29
Add a possibility to declare local development config file

### LINKED ISSUE #3670: Quarkus dev Mode loading a Env file
**Description**
Hi, Just an idea, would be nice to have the quarkus dev:mode accepting an env.file as an argument and all variables that eventually match one of the ${ENV_VARIABLE} in the application.properties, would be replaced.

Example:
`
mvn quarkus:dev  -Denv-file="/path/.env"
`

machi1990:
Fixes #3670 

/cc @dmlloyd

### DIFF
--- a/core/runtime/src/main/java/io/quarkus/runtime/configuration/ApplicationPropertiesConfigSource.java
+++ b/core/runtime/src/main/java/io/quarkus/runtime/configuration/ApplicationPropertiesConfigSource.java
@@ -14,7 +14,10 @@ import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.Collections;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Properties;
+
+import org.jboss.logging.Logger;
 
 import io.smallrye.config.PropertiesConfigSource;
 
@@ -23,7 +26,6 @@ import io.smallrye.config.PropertiesConfigSource;
  */
 public abstract class ApplicationPropertiesConfigSource extends PropertiesConfigSource {
     private static final long serialVersionUID = -4694780118527396798L;
-
     static final String APPLICATION_PROPERTIES = "application.properties";
 
     ApplicationPropertiesConfigSource(InputStream is, int ordinal) {
@@ -85,6 +87,46 @@ public abstract class ApplicationPropertiesConfigSource extends PropertiesConfig
             } else {
                 return null;
             }
+        }
+    }
+
+    public static final class LocalConfigFile extends ApplicationPropertiesConfigSource {
+        private static final Logger LOGGER = Logger.getLogger(LocalConfigFile.class.getName());
+
+        public LocalConfigFile() {
+            super(openStream(), 290);
+        }
+
+        private static InputStream openStream() {
+            final Optional<String> localConfigFile = getLocalConfigFilename();
+            if (localConfigFile.isPresent()) {
+
+                final Path localConfigFilePath = Paths.get(localConfigFile.get());
+
+                if (Files.notExists(localConfigFilePath)) {
+                    LOGGER.warnf("Supplied local configuration file %s does not exist.", localConfigFilePath);
+                    return null;
+                }
+
+                try {
+                    return Files.newInputStream(localConfigFilePath);
+                } catch (NoSuchFileException | FileNotFoundException e) {
+                    return null;
+                } catch (IOException e) {
+                    throw new IOError(e);
+                }
+            }
+
+            return null;
+        }
+
+        public static Optional<String> getLocalConfigFilename() {
+            final String envFile = System.getProperty("quarkus.local-config");
+            if (envFile == null) {
+                return Optional.empty();
+            }
+
+            return Optional.of(envFile);
         }
     }
 }
--- a/core/deployment/src/main/java/io/quarkus/deployment/ExtensionLoader.java
+++ b/core/deployment/src/main/java/io/quarkus/deployment/ExtensionLoader.java
@@ -207,8 +207,8 @@ public final class ExtensionLoader {
         final DefaultValuesConfigurationSource defaultSource = new DefaultValuesConfigurationSource(
                 buildTimeConfig.getLeafPatterns());
         final PropertiesConfigSource pcs = new PropertiesConfigSource(buildSystemProps, "Build system");
-
-        builder.withSources(inJar, defaultSource, pcs);
+        final ApplicationPropertiesConfigSource.LocalConfigFile localConfigSource = new ApplicationPropertiesConfigSource.LocalConfigFile();
+        builder.withSources(inJar, defaultSource, pcs, localConfigSource);
 
         // populate builder with all converters loaded from ServiceLoader
         ConverterSupport.populateConverters(builder);
--- a/core/deployment/src/main/java/io/quarkus/deployment/configuration/ConfigDefinition.java
+++ b/core/deployment/src/main/java/io/quarkus/deployment/configuration/ConfigDefinition.java
@@ -70,10 +70,9 @@ public class ConfigDefinition extends CompoundConfigType {
     private static final List<String> FALSE_POSITIVE_QUARKUS_CONFIG_MISSES = Arrays
             .asList(QUARKUS_NAMESPACE + ".live-reload.password", QUARKUS_NAMESPACE + ".live-reload.url",
                     QUARKUS_NAMESPACE + ".debug.generated-classes-dir", QUARKUS_NAMESPACE + ".debug.reflection",
-                    QUARKUS_NAMESPACE + ".build.skip",
-                    QUARKUS_NAMESPACE + ".version", QUARKUS_NAMESPACE + ".profile", QUARKUS_NAMESPACE + ".test.profile",
-                    QUARKUS_NAMESPACE + ".test.native-image-wait-time",
-                    QUARKUS_NAMESPACE + ".test.native-image-profile");
+                    QUARKUS_NAMESPACE + ".build.skip", QUARKUS_NAMESPACE + ".version", QUARKUS_NAMESPACE + ".profile",
+                    QUARKUS_NAMESPACE + ".test.profile", QUARKUS_NAMESPACE + ".test.native-image-wait-time",
+                    QUARKUS_NAMESPACE + ".test.native-image-profile", QUARKUS_NAMESPACE + ".local-config");
 
     private final TreeMap<String, Object> rootObjectsByContainingName = new TreeMap<>();
     private final HashMap<Class<?>, Object> rootObjectsByClass = new HashMap<>();
--- a/core/deployment/src/main/java/io/quarkus/deployment/steps/ConfigurationSetup.java
+++ b/core/deployment/src/main/java/io/quarkus/deployment/steps/ConfigurationSetup.java
@@ -258,7 +258,7 @@ public class ConfigurationSetup {
     }
 
     /**
-     * Add a config sources for {@code application.properties}.
+     * Add a config sources for {@code application.properties} and {@code local-config}.
      */
     @BuildStep
     void setUpConfigFile(BuildProducer<RunTimeConfigurationSourceBuildItem> configSourceConsumer) {
@@ -266,6 +266,8 @@ public class ConfigurationSetup {
                 ApplicationPropertiesConfigSource.InJar.class.getName(), OptionalInt.empty()));
         configSourceConsumer.produce(new RunTimeConfigurationSourceBuildItem(
                 ApplicationPropertiesConfigSource.InFileSystem.class.getName(), OptionalInt.empty()));
+        configSourceConsumer.produce(new RunTimeConfigurationSourceBuildItem(
+                ApplicationPropertiesConfigSource.LocalConfigFile.class.getName(), OptionalInt.empty()));
     }
 
     /**
--- a/core/deployment/src/main/java/io/quarkus/deployment/steps/DevModeBuildStep.java
+++ b/core/deployment/src/main/java/io/quarkus/deployment/steps/DevModeBuildStep.java
@@ -5,12 +5,24 @@ import java.util.List;
 
 import io.quarkus.deployment.annotations.BuildStep;
 import io.quarkus.deployment.builditem.HotDeploymentWatchedFileBuildItem;
+import io.quarkus.runtime.configuration.ApplicationPropertiesConfigSource;
 
 class DevModeBuildStep {
 
     @BuildStep
     List<HotDeploymentWatchedFileBuildItem> config() {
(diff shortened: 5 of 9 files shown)


docs/src/main/asciidoc/application-configuration-guide.adoc
```diff
@@ -360,6 +360,42 @@ You can override these runtime properties with the following mechanisms (in decr
...
+./mvnw clean compile quarkus:dev -Dquarkus.env.file=.env <1>
+--
+
+1. We start Quarkus application with a path to `.env` file. Absolute path are also accepted.
+
+Open your browser to `http://localhost:8080/greeting` to see the greeting message. Changing the `.env` file will be picked by Quarkus.
+
+NOTE: Variables names in the `.env` file follows the conversion rules of link:https://github.com/eclipse/microprofile-config/blob/master/spec/src/main/asciidoc/configsources.asciidoc#default-configsources[Eclipse MicroProfile]
+
+WARNING: DotEnv files are not to be committed in your git repository or to be used in `production` environments (prefer host/global environment variables in this case).
```
geoand: I am not sure this is absolutely correct. I can easily imagine a scenario where a project would commit an example `.env` file so new developers can easily so which env vars they can override

machi1990: What I have seen in most projects is to ignore a .env file and commit a example.env file.

geoand: So maybe the documentation should be adapted to take this suggestion into account?

machi1990: +1. Actually thought about it when I was writing my comment. In this case I’ll add it as a tip.

machi1990: @geoand Tip added. WDYT?

geoand: Looks good!

docs/src/main/asciidoc/application-configuration-guide.adoc
```diff
@@ -360,6 +360,45 @@ You can override these runtime properties with the following mechanisms (in decr
 
 NOTE: Environment variables names are following the conversion rules of link:https://github.com/eclipse/microprofile-config/blob/master/spec/src/main/asciidoc/configsources.asciidoc#default-configsources[Eclipse MicroProfile]
 
+== Using DotEnv files during Development
+As an alternative to defining global environment variables, you can define environment variables in a `.env` file.
+This is practical as it allows each member of your development team to customize and manage the development environment via a file.
+Let's see this in action.
```
dmlloyd: I think that we need to be clear here that this doesn't really define "real" process environment variables (which are immutable in Java).  Instead of using the phrase "environment variables", we should use the phrase "local configuration properties" or something like that.  In fact there is no relationship between environment variables and this implementation of `.env` other than the fact that both become config sources.

machi1990: Thanks @dmlloyd this is a good point. I'll change the wording.

maxandersen: shouldn't we call these `.properties` or similar instead ? 
`.env` are used by other tools to setup environment variables, i.e. docker-compose.

maxandersen: never mind - misread the PR - these aren't properties; they are actually env variables rather than overriding application.properties

dmlloyd: You were right the first time.  Really this is just a config source in `.properties` format, it's just called `.env` for convention.  But if your process calls `System.getenv()`, none of these values will show up.

machi1990: @dmlloyd would you say we change just the wording in the documentation, and the `-Dquarkus.env.file` know to something like `-Dquarkus.local-config` ?

dmlloyd: Possibly.  I wonder if @maxandersen's concern about overlap with docker-compose and similar tools would cause an actual problem for users.

Having a base name of `application.properties` is a bit awkward in this regard because there's no real logical shortening of this.

Maybe the file ought to be called e.g. `.quarkus` or something instead, and the local config property ought to be called `quarkus.local-config` as you suggest, or maybe `quarkus.config.local`.  Or if we want to try to match up with the original name, `.application` or even `.application.properties`?  I don't know.

machi1990: > Possibly. I wonder if @maxandersen's concern about overlap with docker-compose and similar tools would cause an actual problem for users.

Won't these configuration be part of host/container environment variables? In this case, they will take precedence over .env properties. 

> Maybe the file ought to be called e.g. `.quarkus` or something instead, and the local config property ought to be called `quarkus.local-config` as you suggest, or maybe `quarkus.config.local`. Or if we want to try to match up with the original name, `.application` or even `.application.properties`? I don't know.

I like the `.application.properties` suggestion, I'll go with this one. 
Also users may call the file however they see fit as long as it is a valid `.properties` file when using the `quarkus.local-config` config property.

dmlloyd: > > Possibly. I wonder if @maxandersen's concern about overlap with docker-compose and similar tools would cause an actual problem for users.
> 
> Won't these configuration be part of host/container environment variables? In this case, they will take precedence over .env properties.

Sure, if that's how you code it. :slightly_smiling_face:

> > Maybe the file ought to be called e.g. `.quarkus` or something instead, and the local config property ought to be called `quarkus.local-config` as you suggest, or maybe `quarkus.config.local`. Or if we want to try to match up with the original name, `.application` or even `.application.properties`? I don't know.
> 
> I like the `.application.properties` suggestion, I'll go with this one.
> Also users may call the file however they see fit as long as it is a valid `.properties` file when using the `quarkus.local-config` config property.

My only comment then is that the priority should not be equal to the priority of any other pre-defined configuration source.  You'll want a little bit of space in between so that users can do whatever crazy hacks they want to do.  Maybe we should have an overall configuration source priority list somewhere...

machi1990: > My only comment then is that the priority should not be equal to the priority of any other pre-defined configuration source. You'll want a little bit of space in between so that users can do whatever crazy hacks they want to do. Maybe we should have an overall configuration source priority list somewhere...

I see your point, I chose an arbitrary value of `299`, I can modify that too to `290` maybe :-)

dmlloyd: Our ordering is constrained at least in part by the MP config specification.  But I think it'd be good to get it all documented in one place regardless.

machi1990: @dmlloyd I have force pushed new changes. WDYT?

core/runtime/src/main/java/io/quarkus/runtime/configuration/LocalConfigSource.java
```diff
@@ -0,0 +1,107 @@
...
+
+import org.eclipse.microprofile.config.spi.ConfigSource;
+import org.jboss.logging.Logger;
+
+import io.smallrye.config.EnvConfigSource;
+
+/**
+ * A configuration source which supports local config file.
+ */
+final public class LocalConfigSource implements ConfigSource {
```
dmlloyd: Since it's no longer a `.env` situation, maybe it makes more sense to make another nested class in `ApplicationPropertiesConfigSource`.  All the rewriting code down on https://github.com/quarkusio/quarkus/pull/4258/files#diff-6ba9ba821829318906022ab5610f5f7cR55 could be dropped really.

docs/src/main/asciidoc/application-configuration-guide.adoc
```diff
@@ -360,6 +360,48 @@ You can override these runtime properties with the following mechanisms (in decr
...
+[source,shell]
+--
+GREETING_MESSAGE=hello world
+--
+
+Edit the main configuration `src/main/resources/application.properties` file so that it looks like:
+
+[source,properties]
+----
+greeting.message = ${GREETING_MESSAGE}
```
dmlloyd: This example is a little weird because `GREETING_MESSAGE` is the environment-style name of the `greeting.message` property.   So it's really unclear from the example whether the `GREETING_MESSAGE` in the `.local-config.properties` is meant to (or will) override the value or to provide the resolution for the expression.

docs/src/main/asciidoc/application-configuration-guide.adoc
```diff
@@ -360,6 +360,48 @@ You can override these runtime properties with the following mechanisms (in decr
...
 NOTE: Environment variables names are following the conversion rules of link:https://github.com/eclipse/microprofile-config/blob/master/spec/src/main/asciidoc/configsources.asciidoc#default-configsources[Eclipse MicroProfile]
 
+== Using local configuration file during Development
+As an alternative to defining some configuration variables via host environment variables during development, you can define configuration variables in a local configuration file via a `quarkus.local-config` property.
+This is practical as it allows each member of your development team to customize and manage the development environment via a file.
+Doing so, Quarkus will load configuration variables having higher priority than variable defined in a `application.properties` file but with a lower priority than host environment variables.
+
+See the example below.
+
+Let's create a `src/main/resources/.local-config.properties` file and add the following variable.
```
dmlloyd: Isn't the point of a local config that it would be stored in your home directory, e.g. `${user.home}/.application.properties`?  That was my understanding anyway, maybe I missed something though.

Is there a use case for putting it within the project?

dmlloyd (changes requested):

docs/src/main/asciidoc/application-configuration-guide.adoc
machi1990: I should change the variable name a bit.

docs/src/main/asciidoc/application-configuration-guide.adoc
machi1990: The `quarkus.local-config` permits us to point to any file within the file system. 
It is also easier to manage config file (local or not) that resides within the project rather than outside hence the choice.

dmlloyd: Nevertheless, `src/main/resources` is generally not on the file system.  The files within the project are generally included as resources.  So is the local config a resource or a file?  If it's a file then putting it in the resources directory is confusing if not outright wrong.

machi1990: > If it's a file then putting it in the resources directory is confusing if not outright wrong.

You are right and I have adjusted the example taking into account this feedback.

core/runtime/src/main/java/io/quarkus/runtime/configuration/ApplicationPropertiesConfigSource.java
```diff
@@ -87,4 +89,44 @@ private static InputStream openStream() {
...
+                    return null;
+                } catch (IOException e) {
+                    throw new IOError(e);
+                }
+            }
+
+            return null;
+        }
+
+        public static Optional<String> getLocalConfigFilename() {
```
dmlloyd: Shouldn't there be a default to `${user.home}/.application.properties`?  @lbernardomaia did I miss the purpose of the change?

machi1990: Hey @dmlloyd I missed updating this part. After giving it a bit of thoughts maybe we should drop the default value and just use the value that will be defined by the user by using the `-Dquakurs.local-config` property?

dmlloyd: I guess then the config source would not be included if there was no name defined by the property?

machi1990: Yes, it will be just like the FileSystem app config source which returns an empty Properties object.

dmlloyd: OK.  I'm just not 100% sure that we're really solving a real problem, or at least the real problem that the user originally encountered.  It sounds like what we're really doing here is the same as specifying a path for the existing filesystem-backed config source.

machi1990: > OK. I'm just not 100% sure that we're really solving a real problem, or at least the real problem that the user originally encountered. 

I'd be interested to hear what @lbernardomaia thinks about this and if it covers the use case described in the PR.

machi1990:
> What I'm thinking is maybe we should implements an env file instead of a java properties file. This is very common in a lot of framework/tools to have such file to override the default.
> So I will favor a `.env` file in the root of the project that will override the values from `application.properties`

Hi @loicmathieu a `.env` file is basically a `.properties` file structure wise. I think we preferred to not use the name as we don't want to override `System.env()` which is meant to stay immutable.

cescoffier:
@dmlloyd mind having another look?

dmlloyd:
The basic questions still haven't been addressed AFAICT.  It seems to me that the real problem being solved here is to support a per-user `application.properties` override file.  If that is the case, then it should be found in e.g. `${user.home}/.quarkus/application.properties` or perhaps `${user.home}/.quarkus/${application.name}/application.properties` or maybe both, and should have a priority higher than that of our in-JAR and on-filesystem files (but lower than, say, system properties).  All the talk about looking in `src/main/resources` or anyplace else really doesn't fit with this problem description.

If that use case is *not* what is being solved then we need to identify what the actual use case is.

machi1990:
> The basic questions still haven't been addressed AFAICT. It seems to me that the real problem being solved here is to support a per-user `application.properties` override file. If that is the case, then it should be found in e.g. `${user.home}/.quarkus/application.properties` or perhaps `${user.home}/.quarkus/${application.name}/application.properties` or maybe both, and should have a priority higher than that of our in-JAR and on-filesystem files (but lower than, say, system properties). All the talk about looking in `src/main/resources` or anyplace else really doesn't fit with this problem description.
> 
> If that use case is _not_ what is being solved then we need to identify what the actual use case is.

Thanks @dmlloyd @cescoffier I'll have another look at this. David, will it conflict with the config PR https://github.com/quarkusio/quarkus/pull/5387?

dmlloyd:
It is possible, yes.

machi1990:
Good to know, I am thinking of  holding this till #5387 is merged. Thoughts?

gsmet:
For now, I'm closing this one.

I think we need to understand what the user is trying to do and then we can come up with a solution. If this PR is the direction we should go, then we can reopen it.

I added a comment there: https://github.com/quarkusio/quarkus/issues/3670#issuecomment-559234558 .

Issue #2686, opened 2019-05-31
Native image generation fails with 'Invalid signature file digest' if 'uberJar' is enabled


sberyozkin:
**Describe the bug**
If the `quarkus-maven-plugin` enables a `uberJar` property and signed jars such as BouncyCastle ones are pulled in then the native image generation fails with the `java.lang.SecurityException: Invalid signature file digest for Manifest main attributes`.
Excluding `META-INF/*.SF`, `META-INF/*.DSA` and `META-INF/*.RSA` has no effect.

sberyozkin:
This can be reproduced with the `integration-tests/keycloak`

sberyozkin:
I've assigned it to myself but if someone already knows how to approach this issue please reassign otherwise I'll try to investigate later on, thanks

emmanuelbernard:
Discussing with some others, the best would be for us to filter out the signature files for the moment. If anyone wants to take that action, please open a PR.

jaikiran:
> If anyone wants to take that action, please open a PR.

@emmanuelbernard Here's a PR for this change https://github.com/quarkusio/quarkus/pull/4259

machi1990:
@sberyozkin I am assigning @jaikiran to the issue since he proposed a PR to fix it.

jaikiran:
The PR has been merged, so I think this enhancement can now be closed.

machi1990:
Fixed by #4259

PR #4259, opened 2019-09-29
Skip adding signature files to the generated uber jar

jaikiran:
As suggested in https://github.com/quarkusio/quarkus/issues/2686, the commit in this PR skips adding signature files to the generated uber jar.


machi1990:
@sberyozkin would you like to take a look if this fixes the issue you observed? 

Also /cc @emmanuelbernard

jaikiran:
I'll include a test case for this tomorrow.

sberyozkin:
@jaikiran @machi1990 I've confirmed a fix after re-enabling a uberjar option in the `integration-tests/tika`, perhaps it would suffice for an initial test, thanks

machi1990:
> @jaikiran @machi1990 I've confirmed a fix after re-enabling a uberjar option in the `integration-tests/tika`, perhaps it would suffice for an initial test, thanks

> I'll include a test case for this tomorrow.

Nice, thanks for confirming @sberyozkin . Looks like this is good to be merged once @jaikiran has updated the test case.

jaikiran:
I've now updated this PR to include a test case for this change.

stuartwdouglas (approved):

Issue #4267, opened 2019-09-30
Support WebSocket sub-protocols setting for Vert.x web server


cristianonicolai:
**Description**
In order to allow using GraqhQL WebSocket subscription implementation, Quarkus needs to allow setting the Vert.X webserver WebSocket sub-protocols.
See: https://vertx.io/docs/vertx-web-graphql/java/#_apollo_websocketlink

**Implementation ideas**
Add new configuration parameter to application.properties which allows to set the sub-protocols string. Ex: quarkus.http.web-socket-sub-protocols=graphql-ws

cescoffier:
@tsegismont FYI

cescoffier:
@tsegismont if we add a sub-protocol config would that be enough?

tsegismont:
@cescoffier yes that should be enough

cescoffier:
Ok will work on it.

cescoffier:
@stuartwdouglas Actually, how would that work as the WebSocket is managed in quarkus-http?

cescoffier:
Actually, Stuart already commented on the PR:
https://github.com/quarkusio/quarkus/pull/4268#issuecomment-536398930

gsmet:
I think this has been merged here: https://github.com/quarkusio/quarkus/pull/4268 .

PR #4268, opened 2019-09-30
Support WebSocket sub-protocols

cristianonicolai:
Proposed fix for https://github.com/quarkusio/quarkus/issues/4267

### DIFF
--- a/extensions/vertx-http/deployment/src/main/java/io/quarkus/vertx/http/deployment/WebsocketSubProtocolsBuildItem.java
+++ b/extensions/vertx-http/deployment/src/main/java/io/quarkus/vertx/http/deployment/WebsocketSubProtocolsBuildItem.java
@@ -1,0 +1,16 @@
+package io.quarkus.vertx.http.deployment;
+
+import io.quarkus.builder.item.MultiBuildItem;
+
+public final class WebsocketSubProtocolsBuildItem extends MultiBuildItem {
+
+    private final String websocketSubProtocols;
+
+    public WebsocketSubProtocolsBuildItem(String websocketSubProtocols) {
+        this.websocketSubProtocols = websocketSubProtocols;
+    }
+
+    public String getWebsocketSubProtocols() {
+        return websocketSubProtocols;
+    }
+}
--- a/extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/VertxHttpRecorder.java
+++ b/extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/VertxHttpRecorder.java
@@ -126,7 +126,7 @@ public class VertxHttpRecorder {
                 public Integer get() {
                     return ProcessorInfo.availableProcessors() * 2; //this is dev mode, so the number of IO threads not always being 100% correct does not really matter in this case
                 }
-            });
+            }, null);
 
         } catch (Exception e) {
             throw new RuntimeException(e);
@@ -147,7 +147,8 @@ public class VertxHttpRecorder {
 
     public void startServer(RuntimeValue<Vertx> vertxRuntimeValue, ShutdownContext shutdown,
             HttpConfiguration httpConfiguration, LaunchMode launchMode,
-            boolean startVirtual, boolean startSocket, Supplier<Integer> ioThreads) throws IOException {
+            boolean startVirtual, boolean startSocket, Supplier<Integer> ioThreads, String websocketSubProtocols)
+            throws IOException {
 
         Vertx vertx = vertxRuntimeValue.getValue();
         if (startVirtual) {
@@ -156,7 +157,7 @@ public class VertxHttpRecorder {
         if (startSocket) {
             // Start the server
             if (closeTask == null) {
-                doServerStart(vertx, httpConfiguration, launchMode, ioThreads);
+                doServerStart(vertx, httpConfiguration, launchMode, ioThreads, websocketSubProtocols);
                 if (launchMode != LaunchMode.DEVELOPMENT) {
                     shutdown.addShutdownTask(closeTask);
                 }
@@ -213,10 +214,9 @@ public class VertxHttpRecorder {
     }
 
     private static void doServerStart(Vertx vertx, HttpConfiguration httpConfiguration, LaunchMode launchMode,
-            Supplier<Integer> eventLoops)
-            throws IOException {
+            Supplier<Integer> eventLoops, String websocketSubProtocols) throws IOException {
         // Http server configuration
-        HttpServerOptions httpServerOptions = createHttpServerOptions(httpConfiguration, launchMode);
+        HttpServerOptions httpServerOptions = createHttpServerOptions(httpConfiguration, launchMode, websocketSubProtocols);
         HttpServerOptions sslConfig = createSslOptions(httpConfiguration, launchMode);
 
         int eventLoopCount = eventLoops.get();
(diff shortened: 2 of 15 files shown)


stuartwdouglas:
I don't think this config option makes sense here. This will not affect the JSR-256 websockets implementation as this does not use the Vert.x handshake, so it only really affects vertx-web clients, so it would make sense for it to be in vertx-web rather than vertx-http.

That said for your use case I think you might want a build item rather than a configuration option. It sounds like the graphql extension will always want this enabled? If so there should be a build item that can provide this config, and then the GraphQL extension can produce this build item to make sure the sub protocols are automatically set correctly.

tsegismont:
@stuartwdouglas do you mean that users of the raw Vert.x HTTP server API in Quarkus can't manage Websockets?

cristianonicolai:
@stuartwdouglas thanks for the feedback, correct GraphQL will always want this enabled, will take a look in creating a build item as suggested.

stuartwdouglas:
No, I mean that JSR-256 uses a different handshake, and has its own standard way of managing this. The option only affects users of the vertx-web extension so a user configurable option should be there, as otherwise it is confusing to have an option that is present but does not have any effect most of the time.

tsegismont:
@stuartwdouglas makes sense. Thanks for clarifying.

aloubyansky:
@cristianonicolai were you going to update this PR or open a new one? Thanks!

cristianonicolai:
@aloubyansky will update this one soon, thanks.

cristianonicolai:
@stuartwdouglas please take another look, is this what you had in mind? I'm enabling the subprotocol based on having the ApolloWSHandler on the classpath as this is an extra dependency from vert.x ( vertx-web-graphql ).

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/VertxHttpRecorder.java
```diff
@@ -67,6 +67,8 @@
 
     private static volatile Handler<RoutingContext> hotReplacementHandler;
 
+    private static volatile String websocketSubProtocols;
```
stuartwdouglas: This should not be a field, it should just be passed in as a parameter to the method that requires it.

extensions/vertx-web/deployment/src/main/java/io/quarkus/vertx/web/deployment/VertxWebProcessor.java
```diff
@@ -194,6 +196,20 @@ public void transform(TransformationContext context) {
         });
     }
 
+    @BuildStep
+    WebsocketSubProtocolsBuildItem websocketSubProtocols(CombinedIndexBuildItem combinedIndexBuildItem) {
+        return isApolloInClasspath() ? new WebsocketSubProtocolsBuildItem("graphql-ws") : null;
```
stuartwdouglas: This should be done in the Graph QL extension itself.

stuartwdouglas (changes requested):

extensions/vertx-web/deployment/src/main/java/io/quarkus/vertx/web/deployment/VertxWebProcessor.java
cristianonicolai: vertx-web-graphql is part of vert-web, do we need another quarkus extension for that? just for graphql.

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/VertxHttpRecorder.java
cristianonicolai: the reason I end up adding an attribute is because of the restart on fail. If I just send this to the start server part, It wouldnt be able to have the proper state when trying to recover for the restart.

extensions/vertx-web/deployment/src/main/java/io/quarkus/vertx/web/deployment/VertxWebProcessor.java
stuartwdouglas: So there is not going to be a seperate graphql extension? In that case I think this is ok.

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/VertxHttpRecorder.java
stuartwdouglas: Hmm, either way this is not going to work 100% reliably in the restart on failure case, as if that other method gets called will depend on when it fails.

extensions/vertx-web/deployment/src/main/java/io/quarkus/vertx/web/deployment/VertxWebProcessor.java
cristianonicolai: I dont think it is needed.

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/VertxHttpRecorder.java
cristianonicolai: @stuartwdouglas ok, I removed the variable.

cristianonicolai:
@stuartwdouglas when you have a chance, please take another look.

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/VertxHttpRecorder.java
```diff
@@ -116,98 +114,18 @@ public static void startServerAfterFailedStart() {
...
                 }
-            });
+            }, null);
 
         } catch (Exception e) {
             throw new RuntimeException(e);
         }
     }
 
-    public RuntimeValue<Router> initializeRouter(final RuntimeValue<Vertx> vertxRuntimeValue,
```
stuartwdouglas: Why has all this code been moved around? It makes it much harder to tell what has actually changed when reviewing.

cristianonicolai: sure, fixed now, endup applying the reformat for the entire class.

cristianonicolai:
@stuartwdouglas any chance we could get this merged in time for 0.27? :)

extensions/vertx-web/deployment/src/main/java/io/quarkus/vertx/web/deployment/VertxWebProcessor.java
stuartwdouglas: I had a look at the vert.x module and I think this definitely needs its own extension, if only to provide config. I will also be very surprised if this compiles to native without an extension to register stuff for reflection..

extensions/vertx-http/deployment/src/main/java/io/quarkus/vertx/http/deployment/WebsocketSubProtocolsBuildItem.java
```diff
@@ -0,0 +1,16 @@
+package io.quarkus.vertx.http.deployment;
+
+import io.quarkus.builder.item.SimpleBuildItem;
+
+public final class WebsocketSubProtocolsBuildItem extends SimpleBuildItem {
```
stuartwdouglas: This needs to be a MultiBuildItem, so different extensions can register different sub protocols. Otherwise if two extensions attempt to register sub protocols it will blow up.

extensions/vertx-web/deployment/src/main/java/io/quarkus/vertx/web/deployment/VertxWebProcessor.java
```diff
@@ -69,6 +71,15 @@
 
     HttpConfiguration httpConfiguration;
 
+    private static boolean isApolloInClasspath() {
+        try {
+            Class.forName("io.vertx.ext.web.handler.graphql.ApolloWSHandler");
```
stuartwdouglas: In general in Quarkus we use an additive approach where we define extensions for additional features, rather than the Spring Boot approach of checking what is on the CP to decide what to enable. It looks like GraphQL will require config and reflection so I really think this should be its own extension.

stuartwdouglas (changes requested):

cristianonicolai:
@stuartwdouglas refactored into a new extension.

cristianonicolai:
@stuartwdouglas mind taking another look? Thanks

stuartwdouglas (approved):

PR #4273, opened 2019-09-30
Add Elytron security JDBC extension

stuartwdouglas:


extensions/elytron-security-jdbc/deployment/src/main/java/io/quarkus/elytron/security/jdbc/deployment/ElytronSecurityJdbcProcessor.java
```diff
@@ -0,0 +1,53 @@
...
+    /**
+     * Check to see if a JdbcRealmConfig was specified and enabled and create a
+     * {@linkplain org.wildfly.security.auth.realm.JdbcSecurityRealmConfig}
+     * runtime value to process the user/roles properties files. This also registers the names of the user/roles properties
+     * files
+     * to include the build artifact.
+     *
+     * @param recorder - runtime security recorder
+     * @param securityRealm - the producer factory for the SecurityRealmBuildItem
+     *        // * @param beanContainer - ensure CDI bean container is ready
```
cescoffier: Can be dropped

extensions/elytron-security-jdbc/deployment/src/test/java/io/quarkus/elytron/security/jdbc/ParametrizedPathsResource.java
```diff
@@ -0,0 +1,28 @@
+package io.quarkus.elytron.security.jdbc;
+
+import javax.annotation.security.RolesAllowed;
+import javax.ws.rs.GET;
+import javax.ws.rs.Path;
+import javax.ws.rs.PathParam;
+
+/**
+ * @author Michal Szynkiewicz, michal.l.szynkiewicz@gmail.com
+ *         <br>
```
cescoffier: Not sure about the format

danielpetisme: Same formatting than [here](https://github.com/quarkusio/quarkus/blob/master/extensions/elytron-security/deployment/src/test/java/io/quarkus/security/test/ParametrizedPathsResource.java) Actually it's a copy/paste...

geoand:
The JVM are failing it looks like.

stuartwdouglas:
It passes on my machine, I have pushed a change that should give more info

geoand:
Cool

sberyozkin:
@stuartwdouglas looks like it is good to go ?

stuartwdouglas:
It has been ready for a while, it just needs someone to review it.

extensions/elytron-security-jdbc/runtime/src/main/java/io/quarkus/elytron/security/jdbc/AttributeMappingConfig.java
```diff
@@ -0,0 +1,31 @@
+package io.quarkus.elytron.security.jdbc;
+
+import io.quarkus.runtime.annotations.ConfigGroup;
+import io.quarkus.runtime.annotations.ConfigItem;
+
+/**
```
sberyozkin: It could be useful to expand a bit about what these properties mean here and remove a link to a WildFly class

extensions/elytron-security-jdbc/runtime/src/main/java/io/quarkus/elytron/security/jdbc/BcryptPasswordKeyMapperConfig.java
```diff
@@ -0,0 +1,76 @@
+package io.quarkus.elytron.security.jdbc;
+
+import org.wildfly.security.auth.realm.jdbc.mapper.PasswordKeyMapper;
+
+import io.quarkus.runtime.annotations.ConfigGroup;
+import io.quarkus.runtime.annotations.ConfigItem;
+
+/**
+ * Configuration information used to populate a "bcrypt"
+ * {@linkplain org.wildfly.security.auth.realm.jdbc.mapper.PasswordKeyMapper}
```
sberyozkin: Same here, remove the link to a WildFly class, I guess the users should not really expect that the strong link with WildFly classes will remain

sberyozkin: Never mind, they are used directly in the code...

extensions/elytron-security-jdbc/runtime/src/main/java/io/quarkus/elytron/security/jdbc/JdbcSecurityRealmConfig.java
```diff
@@ -0,0 +1,48 @@
...
+     * The authentication mechanism
+     */
+    @ConfigItem(defaultValue = "Quarkus")
+    public String realmName;
+
+    /**
+     * If the properties store is enabled.
+     */
+    @ConfigItem
+    public boolean enabled;
```
sberyozkin: Should it be set to true by default ? Otherwise the users who need it have to 1) add the module 2) enable the feature while the users who don't need the feature don't have to do anything :-)

sberyozkin (approved):
@stuartwdouglas my only preference is not to require the users enable it explicitly, but only disable it if they need to disable it; but if it is a common pattern then it is fine - up to you whether to change it or not, thanks.
@danielpetisme - thanks for your effort as well :-)

danielpetisme:
@stuartwdouglas  Did you had time to have a look to the doc+quickstart I proposed?

PR #4277, opened 2019-09-30
TCK for MicroProfile OpenTracing

jmartisk:


jmartisk:
I welcome suggestions how to simplify the dependency hell, getting them to converge was quite a challenge.

jmartisk:
cc @pavolloffay

tcks/microprofile-opentracing/pom.xml
```diff
@@ -0,0 +1,165 @@
...
+    <artifactId>quarkus-tck-microprofile-opentracing</artifactId>
+    <name>Quarkus - TCK - MicroProfile OpenTracing</name>
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-surefire-plugin</artifactId>
+                <configuration>
+                    <suiteXmlFiles>
```
kenfinnigan: If there's no exclusions in the tck suite file, can it be removed and the tests are just scanned from the dependencies?

pavolloffay:
There are two TCK modules. `microprofile-opentracing-tck-rest-client` seems to be missing https://github.com/eclipse/microprofile-opentracing/blob/master/tck/rest-client/pom.xml#L25

jmartisk:
> There are two TCK modules. `microprofile-opentracing-tck-rest-client` seems to be missing https://github.com/eclipse/microprofile-opentracing/blob/master/tck/rest-client/pom.xml#L25

No, it's included in the scanned dependencies here https://github.com/quarkusio/quarkus/pull/4277/files#diff-a5ec914784036e3b240f9e1e7e85c184R33

jmartisk:
Hold on, looks like you may be right and I forgot to actually add the dependency :)

tcks/microprofile-opentracing/pom.xml
jmartisk: After I split this into two modules, I can remove the suite file from the `rest-client` module, but not from the `base` module (in that case I'm getting some exceptions from Arc). Not sure exactly why, apparently some initialization code is skipped if I do that

jmartisk:
Should be now ready for another review

tcks/microprofile-opentracing/base/pom.xml
```diff
@@ -0,0 +1,167 @@
...
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-arquillian</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-smallrye-opentracing</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-resteasy-jackson</artifactId>
```
kenfinnigan: Any reason it's `jackson` instead of `jsonb`?

tcks/microprofile-opentracing/base/pom.xml
```diff
@@ -0,0 +1,167 @@
...
+            <groupId>io.smallrye</groupId>
+            <artifactId>smallrye-config</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.opentracing</groupId>
+            <artifactId>opentracing-mock</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.jboss.resteasy</groupId>
+            <artifactId>resteasy-client-microprofile</artifactId>
```
kenfinnigan: Should this be `quarkus-rest-client` instead?

tcks/microprofile-opentracing/base/src/main/java/io/quarkus/tck/opentracing/DeploymentProcessor.java
```diff
@@ -0,0 +1,36 @@
+package io.quarkus.tck.opentracing;
+
+import org.eclipse.microprofile.opentracing.ClientTracingRegistrarProvider;
+import org.jboss.arquillian.container.test.spi.client.deployment.ApplicationArchiveProcessor;
+import org.jboss.arquillian.test.spi.TestClass;
+import org.jboss.shrinkwrap.api.Archive;
+import org.jboss.shrinkwrap.api.spec.WebArchive;
+
+import sun.security.jca.Providers;
```
kenfinnigan: Wrong import, need the JAX-RS one

kenfinnigan (changes requested):

tcks/microprofile-opentracing/base/src/main/java/io/quarkus/tck/opentracing/DeploymentProcessor.java
jmartisk: Haha, good catch. This proves that we don't need this at all and the `ExceptionMapper` is detected automatically.

tcks/microprofile-opentracing/base/pom.xml
jmartisk: I'm seeing assertion errors from comparing equality of JSON documents after doing the switch from `jackson` to `jsonb`, is `jsonb` preferred and should I dig deeper to see if that can be fixed?

kenfinnigan: It's that MP has a dependency on JSONB and not Jackson, so ideally it should work with JSONB.

@pavolloffay, any reason the TCK shouldn't work with JSONB?

jmartisk: The TCK itself depends on jackson https://github.com/eclipse/microprofile-opentracing/blob/1.3.1/tck/base/pom.xml#L75

jmartisk: Ok so I dug deeper into it and found the root cause why switching to `jsonb` causes assertion errors (because normally it should still work nevertheless) - it's because with JSONB the tag `http.status_code` is decoded on the client as a `BigDecimal`, with Jackson it is an `Integer`, and the logic in the TCK (https://github.com/eclipse/microprofile-opentracing/blob/1.3.1/tck/base/src/main/java/org/eclipse/microprofile/opentracing/tck/tracer/TestSpan.java#L382) simply determines equality of tags using `Map.equals()`, so it fails with JSONB because one tag has a value of an unexpected type. So if we want to use JSONB we would need to implement a more lenient equality check in the TCK

tcks/microprofile-opentracing/base/pom.xml
jmartisk: Yes, good point

tcks/microprofile-opentracing/base/pom.xml
kenfinnigan: @jmartisk can you raise an issue with MP OT to fix that, as it shouldn't be relying on the presence of Jackson?

kenfinnigan (approved):

tcks/microprofile-opentracing/base/pom.xml
jmartisk: Yeah, filed https://github.com/eclipse/microprofile-opentracing/issues/159

PR #4284, opened 2019-09-30
Add example service account etc for example.

mejlholm:
I ended up spending some time getting access to the API working, so i thought it would be nice to share the solution. 

If the example is too much, perhaps just include some links to get people going faster.

### DIFF
--- a/docs/src/main/asciidoc/kubernetes-client.adoc
+++ b/docs/src/main/asciidoc/kubernetes-client.adoc
@@ -220,3 +220,47 @@ public class ResourceWatcher<T extends HasMetadata> implements Watcher<T> {
 client.pods().watch(new ResourceWatcher<Pod>());
 ----
 
+== Access to the Kubernetes API
+
+You will most likely need access to the kubernetes API using a ServiceAccount, Role and RoleBinding to run the new application.
+An example would look something like this:
+
+[source,yaml]
+----
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: <applicationName>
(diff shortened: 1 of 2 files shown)


gsmet:
@gastaldi @geoand could you have a look.

@mejlholm could you squash the commits a bit? 5 commits are a bit too much for this change, 2 would be better. Thanks!

geoand:
Looks good to me!

Let's squash however as @gsmet said

docs/src/main/asciidoc/kubernetes-client.adoc
```diff
@@ -220,3 +220,47 @@ public class ResourceWatcher<T extends HasMetadata> implements Watcher<T> {
 client.pods().watch(new ResourceWatcher<Pod>());
 ----
 
+== Access to the Kubernetes API
+
+You will most likely need access to the kubernetes API using a ServiceAccount, Role and RoleBinding to run the new application.
```
geoand: Perhaps we should say: `In many cases in order to access the Kubernetes API server a ServiceAccount, Role and RoleBinding will be necessary`

docs/src/main/asciidoc/kubernetes-client.adoc
```diff
@@ -220,3 +220,47 @@ public class ResourceWatcher<T extends HasMetadata> implements Watcher<T> {
...
+An example would look something like this:
```
geoand: `An example that allows listing all pods could look something like this`

mejlholm:
I'll just close this pull request, re-fork and request again.

PR #4286, opened 2019-10-01
Add a quarkus.http.root-path property that controls where everything is mounted

stuartwdouglas:


docs/src/main/asciidoc/http-reference.adoc
```diff
@@ -170,3 +168,16 @@ The following config options would recognize strings in this format (shown as a
 * `quarkus.http.limits.max-body-size`,
 * `quarkus.http.limits.max-header-size`
 ====
+
+== Servlet Config
```
kenfinnigan: Should this section mention the required Quarkus Undertow dependency specifically?

kenfinnigan (changes requested):

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/VertxHttpRecorder.java
```diff
@@ -180,27 +175,26 @@ public void finalizeRouter(BeanContainer container, Consumer<Route> defaultRoute
...
+        router.route().last().failureHandler(new QuarkusErrorHandler(launchMode.isDevOrTest()));
+
+        if (rootPath.equals("/")) {
+            if (hotReplacementHandler != null) {
+                router.route().order(-1).handler(hotReplacementHandler);
+            }
+            rootHandler = router;
+        } else {
+            Router mainRouter = Router.router(vertx.getValue());
+            mainRouter.mountSubRouter(rootPath, router);
```
cescoffier: shouldn't we add the hotReplaceHandler is this case also?

cescoffier (changes requested):

docs/src/main/asciidoc/http-reference.adoc
```diff
@@ -16,28 +16,27 @@ include::./attributes.adoc[]
 :sectnumlevels: 4
 
 
-This document explains various Undertow features that you can use in Quarkus.
+This document explains various HTTP features that you can use in Quarkus.
+
+HTTP is provided using Eclipse Vert.x as the base HTTP layer. Servlet's are supported using a modified version of Undertow that
+runs on top of Vert.x, and RESTeasy is used to provide JAX-RS support. If Undertow is present RESTeasy will run as a
```
FroMage: I believe it's `RESTEasy` :(

docs/src/main/asciidoc/http-reference.adoc
```diff
@@ -26,18 +30,12 @@ jar files as defined by the Servlet spec.
```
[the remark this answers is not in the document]
FroMage: Pfff, I would not even justify this choice, we can say that's the convention we chose, and we can even make it overridable in the future if someone asks. Who cares why we chose it: it's not a far-fetched default by any measure.

docs/src/main/asciidoc/http-reference.adoc
```diff
@@ -16,28 +16,27 @@ include::./attributes.adoc[]
...
-link:http://undertow.io/undertow-docs/undertow-docs-2.0.0/index.html#predicates-attributes-and-handlers[Undertow predicate language].
+By default Quarkus will serve content from under the root context. If you want to change this you can use the
+`quarkus.http.root-path` config key to set the context path.
 
-== web.xml
-
-If you are using a `web.xml` file as your configuration file, you can place it in the `src/main/resources/META-INF` directory.
+If you are using Servlet you can control the Servlet context path via `quarkus.servlet.context-path`. This item is relative
+to the http root above, and will only affect Servlet and things that run on top of Servlet. Most applications will
+want to use the HTTP root as this affects everything that Quarkus serves.
```
FroMage: I am confused here about how the two combine together. What happens when someone specifies both?

extensions/smallrye-openapi/deployment/src/test/java/io/quarkus/smallrye/openapi/test/OpenApiHttpRootDefaultPathTestCase.java
```diff
@@ -0,0 +1,41 @@
...
+    @RegisterExtension
+    static QuarkusUnitTest runner = new QuarkusUnitTest()
+            .setArchiveProducer(() -> ShrinkWrap.create(JavaArchive.class)
+                    .addClasses(OpenApiResource.class)
+                    .addAsResource(new StringAsset("quarkus.http.root-path=/foo"), "application.properties"));
+
+    @Test
+    public void testOpenApiPathAccessResource() {
+        RestAssured.given().header("Accept", "application/yaml")
+                .when().get(OPEN_API_PATH)
```
FroMage: But… it's at `/openapi`? Shouldn't it be at `/foo/openapi`?

stuartwdouglas:
On Wed, 2 Oct 2019 at 22:41, Stéphane Épardaud <notifications@github.com>
wrote:

> *@FroMage* commented on this pull request.
> ------------------------------
>
> In
> extensions/smallrye-openapi/deployment/src/test/java/io/quarkus/smallrye/openapi/test/OpenApiHttpRootDefaultPathTestCase.java
> <https://github.com/quarkusio/quarkus/pull/4286#discussion_r330525224>:
>
> > +import io.quarkus.test.QuarkusUnitTest;
>
> +import io.restassured.RestAssured;
>
> +
>
> +public class OpenApiHttpRootDefaultPathTestCase {
>
> +    private static final String OPEN_API_PATH = "/openapi";
>
> +
>
> +    @RegisterExtension
>
> +    static QuarkusUnitTest runner = new QuarkusUnitTest()
>
> +            .setArchiveProducer(() -> ShrinkWrap.create(JavaArchive.class)
>
> +                    .addClasses(OpenApiResource.class)
>
> +                    .addAsResource(new StringAsset("quarkus.http.root-path=/foo"), "application.properties"));
>
> +
>
> +    @Test
>
> +    public void testOpenApiPathAccessResource() {
>
> +        RestAssured.given().header("Accept", "application/yaml")
>
> +                .when().get(OPEN_API_PATH)
>
>
> But… it's at /openapi? Shouldn't it be at /foo/openapi?
>

We automatically update the Restassured base URI so you don't have to worry
about it in tests.

extensions/vertx-http/deployment/src/main/java/io/quarkus/vertx/http/deployment/HttpRootPathBuildItem.java
```diff
@@ -0,0 +1,26 @@
...
+    public HttpRootPathBuildItem(String rootPath) {
+        this.rootPath = rootPath;
+    }
+
+    public String getRootPath() {
+        return rootPath;
+    }
+
+    public String adjustPath(String path) {
+        if (!path.startsWith("/")) {
```
FroMage: What's the point here? Can't we add the initial `/` if missing? It's pretty obvious the root path must have a `/` because that's just how URIs are built, so it's not like this could be mistaken for a slashless root path…

FroMage (approved):

extensions/smallrye-openapi/deployment/src/test/java/io/quarkus/smallrye/openapi/test/OpenApiHttpRootDefaultPathTestCase.java
FroMage: Oh, I see. I wonder if that's nice or error-prone. I guess it's nice. Is it documented?

Won't we face people wondering why their browser says `/foo/openapi` exists and they write their test with `/foo/openapi` and get a 404?

docs/src/main/asciidoc/http-reference.adoc
FroMage: It's actually explained in the javadoc of the config itself, but should be clarified here.

cescoffier: it's explained in the next paragraph.

stuartwdouglas: I added that after @FroMage complained :-)

PR #4297, opened 2019-10-01
Add ability to produce overridable build items

### LINKED ISSUE #4270: Enhanced conditional BuildStep
The `BuildStep` annotation supports simple conditions `onlyIn/onlyIfNot` that one can implement using a `BooleanSupplier` but in some cases it would be nice to have support for more enhanced construct.

In my specific case I have a core extension with a `BuildStep` that should provide the default instance for a given interface and a number of extensions that can provide specialized implementations for that.

I've actually solved the issue by using an `MultiBuildItem` and depending on number of instances found, I do instantiate the default implementation, get the first one that is found or throw an exception in case of multiple instances being found.

This solution works fine but as it seems to me a common pattern it would be nice if it could be expressed through annotations, i.e. spring has a `ConditionalOnMissingBean` that I've used extensively to implement the same logic at runtime. 

Maybe related to: https://github.com/quarkusio/quarkus/issues/4114

dmlloyd:
Fixes #4270

### DIFF
--- a/core/builder/src/main/java/io/quarkus/builder/BuildChainBuilder.java
+++ b/core/builder/src/main/java/io/quarkus/builder/BuildChainBuilder.java
@@ -191,29 +191,33 @@ public final class BuildChainBuilder {
             for (Map.Entry<ItemId, Produce> entry : stepProduces.entrySet()) {
                 final ItemId id = entry.getKey();
                 final List<Produce> list = allProduces.computeIfAbsent(id, x -> new ArrayList<>(2));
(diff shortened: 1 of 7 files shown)


stuartwdouglas:
@lburgazzoli Can you verify that this works for your use case?

lburgazzoli:
I will, not very sure if I can do it today as I'm in the middle of a huge refactor but I definitively will

lburgazzoli:
will test tomorrow

lburgazzoli:
LGTM

lburgazzoli (approved):

stuartwdouglas (approved):

PR #4300, opened 2019-10-01
Add support for @MockBean

geoand:
With this annotation, CDI beans will automatically be mocked using mockito.

**TODO**:

- [ ] Add support for qualifiers
- [ ] Add documentation

### DIFF
--- a/integration-tests/mockbean/src/test/java/io/quarkus/it/mockbean/GreetingResourceTest.java
+++ b/integration-tests/mockbean/src/test/java/io/quarkus/it/mockbean/GreetingResourceTest.java
@@ -1,0 +1,44 @@
+package io.quarkus.it.mockbean;
+
+import static io.restassured.RestAssured.given;
+import static org.hamcrest.Matchers.is;
+
+import org.junit.jupiter.api.Test;
+import org.mockito.Mockito;
+
+import io.quarkus.test.junit.QuarkusTest;
+import io.quarkus.test.junit.mockito.MockBean;
+
+@QuarkusTest
+class GreetingResourceTest {
+
+    @MockBean
+    MessageService messageService;
+
+    @MockBean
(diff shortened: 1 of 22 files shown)


geoand:
Adding `coding-in-progress` in order to account for feedback and the fact that if people think it's a good idea I also need to document it in the testing guide.

integration-tests/mockbean/src/test/java/io/quarkus/it/mockbean/GreetingResourceTest.java
```diff
@@ -0,0 +1,44 @@
...
+import org.mockito.Mockito;
+
+import io.quarkus.test.junit.QuarkusTest;
+import io.quarkus.test.junit.mockito.MockBean;
+
+@QuarkusTest
+class GreetingResourceTest {
+
+    @MockBean
+    MessageService messageService;
```
mkouba: Currently, you can do:
```java
@Produces
@Mock
MessageService mockMessageService() {
   return Mockito.mock(MessageService.class);
}

@Inject
Message
```

geoand: Yes absolutely!

The PR doesn't bring in fundamental capability that didn't already exist. It does however lighten the burden for users if they have multiple mocks they want to use in their tests (whether that is good or bad is another question :) )

cescoffier:
@geoand @mkouba what's the status of this PR?

geoand:
On hold for the time being

gsmet:
I think a first good step would be to document what @mkouba is saying. Then we could see if it needs improvements.

Keep in mind that most users are not accustomed to CDI so this is not obvious.

geoand:
I'm going to close this since it's very out of date. If needed in the future, we can revisit

PR #4304, opened 2019-10-02
Prevent Vert.x/RESTeasy hang if HTTP URL handler is not installed

stuartwdouglas:
Also add Vert.x error handler

### DIFF
--- a/extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/VertxHttpRecorder.java
+++ b/extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/VertxHttpRecorder.java
@@ -37,7 +37,13 @@ import io.quarkus.vertx.core.runtime.VertxCoreRecorder;
 import io.quarkus.vertx.core.runtime.config.VertxConfiguration;
 import io.quarkus.vertx.http.runtime.filters.Filter;
 import io.quarkus.vertx.http.runtime.filters.Filters;
-import io.vertx.core.*;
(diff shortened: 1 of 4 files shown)


extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/VertxHttpRecorder.java
```diff
@@ -194,6 +200,7 @@ private static void doServerStart(Vertx vertx, HttpConfiguration httpConfigurati
             Supplier<Integer> eventLoops)
             throws IOException {
         // Http server configuration
+        router.route().failureHandler(new QuarkusErrorHandler(launchMode.isDevOrTest()));
```
cescoffier: It might be better to register it as "last" explicitly.

cescoffier (approved):
I made a small comment, but the rest looks good.

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/VertxHttpRecorder.java
stuartwdouglas: Won't this be last by default as the router has already been setup?

stuartwdouglas: I have updated it anyway

geoand:
The CI failure seems to be a `jcenter` issue. I propose we ignore it, it's been happening on various PRs

PR #4309, opened 2019-10-02
Update Security docs and properties files

stuartwdouglas:
This moves the properties file config out of the core Elytron module, and also adds support for hashed passwords to remove the need for plaintext passwords in the files.

@gsmet this has a high chance to conflict so a prompt review would be appreciated.


geoand:
I went through the doc update quickly and it looks good. I would personally merge this quickly assuming the tests pass so people can start looking into what is changing with security.

geoand:
The gradle plugin test failure seems totally unrelated. I'll restart it when the rest of the tests finish.

core/deployment/src/main/java/io/quarkus/deployment/builditem/FeatureBuildItem.java
```diff
@@ -48,6 +48,7 @@
     public static final String SCALA = "scala";
     public static final String SCHEDULER = "scheduler";
     public static final String SECURITY = "security";
+    public static final String SECURITY_PROPERTIES_FILE = "elytron-security-properties-file";
```
gsmet:
```suggestion
    public static final String SECURITY_PROPERTIES_FILE = "security-properties-file";
```
would be better I think. We totally removed the `elytron` part from the other extension names.

bom/deployment/pom.xml
```diff
@@ -346,6 +346,11 @@
                 <artifactId>quarkus-elytron-security-deployment</artifactId>
                 <version>${project.version}</version>
             </dependency>
+            <dependency>
+                <groupId>io.quarkus</groupId>
+                <artifactId>quarkus-elytron-security-properties-file-deployment</artifactId>
```
gsmet: I'm wondering if we should introduce `elytron` now? We named everything `security` before to not expose it.

docs/src/main/asciidoc/elytron-properties-guide.adoc
```diff
@@ -0,0 +1,163 @@
...
+include::./attributes.adoc[]
+
+Quarkus provides support for properties file based authentication that is intended for
+development and testing purposes. It is not recommended that this be used in production as at present only
+plaintext passwords are supported.
+
+Add the following to your `pom.xml`:
+
+[source,xml]
+--
```
gsmet: It should be `----` for sources.

docs/src/main/asciidoc/elytron-properties-guide.adoc
```diff
@@ -0,0 +1,168 @@
...
+scott=jb0ss <1>
+jdoe=p4ssw0rd <2>
+stuart=test
+noadmin=n0Adm1n
+----
+<1> User `scott` has password defined as `jb0ss`
+<2> User `jdoe` has password defined as `p4ssw0rd`
+
+This file has the usernames and passwords stored in plain text, which is not recommended. If plain-text is set to false
+(or omitted) in the config then passwords must be stored in the form `MD5 ( username : realm : password )`. This can
```
gsmet: Is there any other hash available? `md5` is really not considered secured these days.

docs/src/main/asciidoc/elytron-properties-guide.adoc
```diff
@@ -0,0 +1,168 @@
...
+and pull requests should be submitted there:
+https://github.com/quarkusio/quarkus/tree/master/docs/src/main/asciidoc
+////
+= Quarkus - Elytron Properties File Config
+
+include::./attributes.adoc[]
+
+Quarkus provides support for properties file based authentication that is intended for
+development and testing purposes. It is not recommended that this be used in production as at present only
+plaintext passwords are supported.
```
gsmet: This is not true anymore?

docs/src/main/asciidoc/elytron-properties-guide.adoc
```diff
@@ -0,0 +1,168 @@
...
+scott=Admin,admin,Tester,user <1>
+jdoe=NoRolesUser <2>
+stuart=admin,user <3>
+noadmin=user
+----
+<1> User `scott` has been assigned the roles `Admin`, `admin`, `Tester` and `user`
+<2> User `jdoe` has been assigned the role `NoRolesUser`
+<3> User `stuart` has been assigned the roles `admin` and `user`.
+
+Given these role mappings, only user `scott` would be allowed to access the `/subject/secured` endpoint from the <<subject-example>>.
```
gsmet: The `<<subject-example>>` cross reference points to nothing existing in this document.

docs/src/main/asciidoc/elytron-properties-guide.adoc
```diff
@@ -0,0 +1,168 @@
+////
+This guide is maintained in the main Quarkus repository
+and pull requests should be submitted there:
+https://github.com/quarkusio/quarkus/tree/master/docs/src/main/asciidoc
+////
+= Quarkus - Elytron Properties File Config
```
gsmet: I pushed a small update with some formatting improvements.

docs/src/main/asciidoc/keycloak-guide.adoc
```diff
@@ -86,28 +83,28 @@ import javax.ws.rs.Produces;
...
+        return new User(securityIdentity);
     }
 
     public class User {
 
         private final String userName;
 
-        User(KeycloakSecurityContext securityContext) {
-            this.userName = securityContext.getToken().getPreferredUsername();
+        User(SecurityIdentity securityContext) {
```
gsmet: should the variable be renamed too?

docs/src/main/asciidoc/keycloak-guide.adoc
```diff
@@ -371,48 +304,6 @@ export access_token=$(\
 )
 ```
 
-== Multi-Tenant Deployments
```
gsmet: We don't support that anymore?

docs/src/main/asciidoc/security-guide.adoc
```diff
@@ -3,12 +3,15 @@ This guide is maintained in the main Quarkus repository
...
 https://github.com/quarkusio/quarkus/tree/master/docs/src/main/asciidoc
 ////
-= Quarkus - Using Security
+= Quarkus - Security Guide
 
 include::./attributes.adoc[]
 
-Quarkus comes with integration with the Elytron security subsystem to allow for RBAC based on the
-common security annotations `@RolesAllowed`, `@DenyAll`, `@PermitAll` on REST endpoints. An example of an endpoint that makes use of both JAX-RS and Common Security annotations to describe and secure its endpoints is given in <<subject-example>>.
+Quarkus comes with build in securuty to allow for RBAC based on thecommon security annotations `@RolesAllowed`,
```
gsmet:
```suggestion
Quarkus comes with build in security to allow for RBAC based on the common security annotations `@RolesAllowed`,
```

docs/src/main/asciidoc/security-guide.adoc
```diff
@@ -61,264 +64,42 @@ public class SubjectExposingResource {
...
---
-<1> User `scott` has password defined as `jb0ss`
-<2> User `jdoe` has password defined as `p4ssw0rd`
+|Extension |Description
 
+|link:elytron-properties-guide.html[quarkus-elytron-security-properties-file]
+|Provides support for simples properties files that can be used for testing security. This supports both embedding user info in `application.properties` and standalone properties files.
 
-==== Roles.properties
+|link:oauth2-guide.html[quarkus-elytron-security-oath2]
```
gsmet:
```suggestion
|link:oauth2-guide.html[quarkus-elytron-security-oauth2]
```

docs/src/main/asciidoc/security-guide.adoc
```diff
@@ -61,264 +64,42 @@ public class SubjectExposingResource {
...
+|Provides support for OAth2 flows using Elytron. This extension will likely be deprecated soon and replaced by a reactive Vert.x version.
```
gsmet:
```suggestion
|Provides support for OAuth2 flows using Elytron. This extension will likely be deprecated soon and replaced by a reactive Vert.x version.
```

docs/src/main/asciidoc/security-guide.adoc
```diff
@@ -61,264 +64,42 @@ public class SubjectExposingResource {
...
-stuart=admin,user <3>
-noadmin=user
---
-<1> User `scott` has been assigned the roles `Admin`, `admin`, `Tester` and `user`
-<2> User `jdoe` has been assigned the role `NoRolesUser`
-<3> User `stuart` has been assigned the roles `admin` and `user`.
-
-Given these role mappings, only user `scott` would be allowed to access the `/subject/secured` endpoint from the <<subject-example>>.
+|link:jwt-guide.html[quarkus-smallrye-jwt]
+|A Microprofile JWT implementation that provides support for authenticating use Json Web Tokens. This also allows you to inject the token and claims into the application as per the MP JWT spec.
```
gsmet:
```suggestion
|A Microprofile JWT implementation that provides support for authenticating using Json Web Tokens. This also allows you to inject the token and claims into the application as per the MP JWT spec.
```

docs/src/main/asciidoc/elytron-properties-guide.adoc
```diff
@@ -0,0 +1,168 @@
...
+The elytron-security-properties-file extension currently supports two different realms for the storage of authentication
+and authorization information. Both support storage of this information in properties type files. The next two sections detail the specific configuration properties.
+
+=== Property Files Realm Configuration
+
+The property files realm supports mapping of users to password and users to roles with a combination of properties files. To enable and configure it, the following configuration properties are used:
+
+|===
+|Property Name|Default|Description
+|quarkus.users.file.enabled|false|Determine whether security via the file realm is enabled
```
gsmet: I wonder if it would be better to have a `security` prefix (so `quarkus.security.users`). For users, it might be self explanatory but it's not a given it will be for everything.

docs/src/main/asciidoc/security-guide.adoc
```diff
@@ -61,264 +64,42 @@ public class SubjectExposingResource {
...
 
-### Embedded Realm Configuration
-The embedded realm also supports mapping of users to password and users to roles. It uses the main application.properties Quarkus configuration file to embed this information. To enable and configure it, the following configuration properties are used:
+|link:keycloak-guide.html[quarkus-vertx-keycloak]
```
gsmet: why did we rename it `vertx-keycloak`? It doesn't work at all in a servlet environment?

gsmet (changes requested):
I added a commit for doc formatting and a couple of comments.

My main concern is about the names of the extensions:
- do we really want to expose `elytron`? Until now we refrained to do it.
- is the Vert.x component of `vertx-keycloak` important for the users? My main concern is that in the vast majority of cases, people won't know they are using Vert.x as they will just have included the RESTEasy extension. Thus they might think that `vertx-keycloak` is not for them.

bom/deployment/pom.xml
stuartwdouglas: The artifacts were still called Elytron. I am ok with renaming it.

docs/src/main/asciidoc/elytron-properties-guide.adoc
stuartwdouglas: LegacyProperyRealm in Elytron only supports this, and it's not supposed to be used in production anyway. The other advantage is that lots of systems have md5 available so you can generate the password by with echo -n user:realm:password | md5.

We could change it but it would be a bit of work as it would mean forking the Elytron implementation, and we would also need to come up with some kind of tool to enable the users to generate the passwords ( I want to do this eventually anyway, but not as an ad-hoc thing).

docs/src/main/asciidoc/elytron-properties-guide.adoc
stuartwdouglas: Oops, I wrote this before I decided I was going to go ahead and implement the hashed passwords

docs/src/main/asciidoc/keycloak-guide.adoc
stuartwdouglas: Nope. That was specific to the Keycloak Undertow adaptor.

docs/src/main/asciidoc/security-guide.adoc
stuartwdouglas: Because our naming format is productname-feature. In this case vertx is providing Keycloak, so according to our standard naming rules this should be the name. I am happy to change it to just keycloak, but it might even need to be something more generic (as it may just be the open id connect implementation)

bom/deployment/pom.xml
gsmet: Yeah, just saw that. I would say let's not block this PR for that. We can get it in and then decide to rename all of them consistently if we want to do so.

docs/src/main/asciidoc/elytron-properties-guide.adoc
gsmet: Yeah, probably not worth it. It's just that having `md5` in a security related documentation could look like a joke.

docs/src/main/asciidoc/security-guide.adoc
gsmet: Can we have a discussion about the naming on the mailing list? In respect to the `elytron` thing and `vertx-keycloak`? Let's not block this PR for that.

gsmet (approved):
Let's get this in and then we can discuss the naming globally.

PR #4314, opened 2019-10-02
Vertx web - use shared BodyHandler instance

### LINKED ISSUE #4294: Vertx web - shared BodyHandler instance
@stuartwdouglas @cescoffier Is there a reason why we create a new `BodyHandler` for each `@Route` method?

https://github.com/quarkusio/quarkus/blob/master/extensions/vertx-web/runtime/src/main/java/io/quarkus/vertx/web/runtime/VertxWebRecorder.java#L70-L82

If not I can modify the code to share one instance instead.

mkouba:
- resolves #4294

### DIFF
--- a/extensions/vertx-web/deployment/src/main/java/io/quarkus/vertx/web/deployment/VertxWebProcessor.java
+++ b/extensions/vertx-web/deployment/src/main/java/io/quarkus/vertx/web/deployment/VertxWebProcessor.java
@@ -131,13 +131,15 @@ class VertxWebProcessor {
                 generatedClass.produce(new GeneratedClassBuildItem(true, name, data));
             }
         };
+        Handler<RoutingContext> bodyHandler = recorder.createBodyHandler(httpConfiguration);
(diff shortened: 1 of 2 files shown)


cescoffier (approved):

mkouba:
**Native Tests main** fails with many occurrences of:

```
Caused by: com.oracle.svm.core.jdk.UnsupportedFeatureError: Proxy class defined by interfaces [interface javax.servlet.http.HttpServletRequest] not found. Generating proxy classes at runtime is not supported. Proxy classes need to be defined at image build time by specifying the list of interfaces that they implement. To define proxy classes use -H:DynamicProxyConfigurationFiles=<comma-separated-config-files> and -H:DynamicProxyConfigurationResources=<comma-separated-config-resources> options.
```

Does anybody have an idea?

cescoffier:
can you rebase, it may have been fixed.

mkouba:
Rebased...

PR #4327, opened 2019-10-02
Force java.io.tmpdir = /tmp for native executable 

oztimpower:
The only writeable folder for Amazon Lambda is /tmp.

vert.x upon startup attempts to create [tmpdir]/vertx-cache, which for the native executable is pointing to a readonly folder (/var/tmp) and not /tmp.

By modifying the bootstrap script it will force the native executable to use the correct java.io.tmpdir=/tmp, and the vertx-cache folder can be created.


stuartwdouglas (approved):

stuartwdouglas:
Should this be done for the JVM part of the script as well?

geoand:
@cescofier you were looking into this IIRC?

cescoffier:
@geoand yes it's a more general issue where we may have to pass the tmp directory in a lots of deployment.

PR #4328, opened 2019-10-02
Add the implementation for the RestClient configKey parameter

### LINKED ISSUE #4301: Add support for the configKey parameter for RegisterRestClient annotation
**Description**
The configKey parameter is new and add in the MP 1.3

The configKey parameter description:
Associates the annotated rest client interface with this configuration key. By specifying a non-empty value, this interface can be configured more simply using the configuration key rather than the fully-qualified class name of the interface.

**Implementation ideas**
Update the handling in the method getBaseUrl of the class RestClientBase
https://github.com/quarkusio/quarkus/blob/b11c1261626376c59744389551061174abae5d3d/extensions/rest-client/runtime/src/main/java/io/quarkus/restclient/runtime/RestClientBase.java#L44

andrejpetras:
Fixes: #4301

### DIFF
--- a/extensions/rest-client/deployment/src/main/java/io/quarkus/restclient/deployment/RestClientProcessor.java
+++ b/extensions/rest-client/deployment/src/main/java/io/quarkus/restclient/deployment/RestClientProcessor.java
@@ -212,10 +212,11 @@ class RestClientProcessor {
                     configurator.creator(m -> {
                         // return new RestClientBase(proxyType, baseUri).create();
                         ResultHandle interfaceHandle = m.loadClass(restClientName.toString());
-                        ResultHandle baseUriHandle = m.load(getBaseUri(entry.getValue()));
+                        ResultHandle baseUriHandle = m.load(getAnnotationParameter(entry.getValue(), "baseUri"));
+                        ResultHandle configKeyHandle = m.load(getAnnotationParameter(entry.getValue(), "configKey"));
                         ResultHandle baseHandle = m.newInstance(
-                                MethodDescriptor.ofConstructor(RestClientBase.class, Class.class, String.class),
-                                interfaceHandle, baseUriHandle);
+                                MethodDescriptor.ofConstructor(RestClientBase.class, Class.class, String.class, String.class),
+                                interfaceHandle, baseUriHandle, configKeyHandle);
                         ResultHandle ret = m.invokeVirtualMethod(
                                 MethodDescriptor.ofMethod(RestClientBase.class, "create", Object.class), baseHandle);
                         m.returnValue(ret);
@@ -262,13 +263,13 @@ class RestClientProcessor {
         return scopeInfo;
     }
 
-    private String getBaseUri(ClassInfo classInfo) {
+    private String getAnnotationParameter(ClassInfo classInfo, String parameterName) {
         AnnotationInstance instance = classInfo.classAnnotation(REGISTER_REST_CLIENT);
         if (instance == null) {
             return "";
         }
 
-        AnnotationValue value = instance.value("baseUri");
+        AnnotationValue value = instance.value(parameterName);
         if (value == null) {
             return "";
         }
--- a/extensions/rest-client/runtime/src/main/java/io/quarkus/restclient/runtime/RestClientBase.java
+++ b/extensions/rest-client/runtime/src/main/java/io/quarkus/restclient/runtime/RestClientBase.java
@@ -16,12 +16,14 @@ public class RestClientBase {
 
     private final Class<?> proxyType;
     private final String baseUriFromAnnotation;
+    private final String propertyPrefixFromAnnotation;
 
     private final Config config;
 
-    public RestClientBase(Class<?> proxyType, String baseUriFromAnnotation) {
+    public RestClientBase(Class<?> proxyType, String baseUriFromAnnotation, String propertyPrefixFromAnnotation) {
         this.proxyType = proxyType;
         this.baseUriFromAnnotation = baseUriFromAnnotation;
+        this.propertyPrefixFromAnnotation = propertyPrefixFromAnnotation;
         this.config = ConfigProvider.getConfig();
     }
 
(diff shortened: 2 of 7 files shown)


extensions/rest-client/deployment/src/test/java/io/quarkus/restclient/ft/RestClientAnnotationTest.java
```diff
@@ -0,0 +1,97 @@
...
+import io.quarkus.test.QuarkusUnitTest;
+
+public class RestClientAnnotationTest {
+
+    @RegisterExtension
+    static final QuarkusUnitTest config = new QuarkusUnitTest()
+            .setArchiveProducer(() -> ShrinkWrap.create(JavaArchive.class)
+                    .addClasses(RestClientAnnotationTest.class, TestEndpoint.class, Client.class));
+
+    static final String URL = "http://localhost:8081";
```
kenfinnigan: We don't want to hardcode this URL, as I know there's work towards supporting random test ports.

There might be a way to get the URL

kenfinnigan: Actually, I wonder if this wouldn't be better to start the test on an entirely separate port?

Then it's actually testing against a port that the app wouldn't be running on normally.

kenfinnigan (changes requested):

extensions/rest-client/deployment/src/test/java/io/quarkus/restclient/ft/RestClientAnnotationTest.java
```diff
@@ -0,0 +1,97 @@
+package io.quarkus.restclient.ft;
```
machi1990: Just a minor comment - It seems like this package is there for `fault tolerance` specific tests. Can we move this test on its own package?

andrejpetras: yes, sure. For example io.quarkus.restclient.rc would it be ok?

kenfinnigan: I would put it in restclient, no need for rc

extensions/rest-client/deployment/src/test/java/io/quarkus/restclient/ft/RestClientAnnotationTest.java
andrejpetras: Thanks for checking it.
Yes, you are right. The best solution would dynamically get the configuration.

For the configKey I need to create dynamically config property based on the test running URL (I can also delete the src/test/resources/application.properties).
Do we have any example for it?

The baseUri in the annotation is string constant. I have no idea how to fix that.

extensions/rest-client/deployment/src/test/java/io/quarkus/restclient/ft/RestClientAnnotationTest.java
andrejpetras: ok

machi1990: +1 with @kenfinnigan

extensions/rest-client/deployment/src/test/java/io/quarkus/restclient/ft/RestClientAnnotationTest.java
kenfinnigan: It should be possible to set system variables for the test, to circumvent the need to have them in a static file.

andrejpetras: I removed the hardcoded URL from the tests

extensions/rest-client/deployment/src/test/java/io/quarkus/restclient/ft/RestClientAnnotationTest.java
andrejpetras: Done!

extensions/rest-client/deployment/src/test/java/io/quarkus/restclient/ft/RestClientAnnotationTest.java
geoand: Just as an FYI, the random test port support has landed in #4140

extensions/rest-client/deployment/src/test/java/io/quarkus/restclient/RestClientAnnotationTest.java
```diff
@@ -0,0 +1,89 @@
...
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.RegisterExtension;
+
+import io.quarkus.test.QuarkusUnitTest;
+import io.quarkus.test.common.http.TestHTTPResourceManager;
+
+public class RestClientAnnotationTest {
+
+    static {
+        System.setProperty("io.quarkus.restclient.RestClientAnnotationTest$Client/mp-rest/url",
```
geoand: It would be great to avoid using System properties for such tests. Using something like [this](https://github.com/quarkusio/quarkus/blob/master/extensions/arc/deployment/src/test/java/io/quarkus/arc/test/config/ConfigArrayConverterTest.java#L23) would be best

andrejpetras: @geoand thanks for the hint. That's exactly what I was looking for. I removed the System properties.

geoand: Cool!

kenfinnigan:
Just had a thought while re-reviewing this. Would it be better if the test was actually in https://github.com/quarkusio/quarkus/tree/master/integration-tests/main/src/main/java/io/quarkus/it/rest and then there wasn't a need for `quarkus-rest-client-deployment` to have a test scope dependency to `quarkus-resteasy-deployment`?

andrejpetras:
That makes sense to me.
Should I move the test to integration-tests module and package: io.quarkus.it.rest?

kenfinnigan:
Sure

andrejpetras:
> Sure

Done!

kenfinnigan (approved):

PR #4337, opened 2019-10-03
Fix Hot reload when mixing filters and routes

cescoffier:
The solution is simple: make sure that the hot replacement handler is first.

A test has been added.


extensions/vertx-http/deployment/src/test/java/io/quarkus/vertx/http/hotreload/DevFilter.java
```diff
@@ -0,0 +1,15 @@
+package io.quarkus.vertx.http.hotreload;
+
+import javax.enterprise.event.Observes;
+
+import io.quarkus.vertx.http.runtime.filters.Filters;
+
+public class DevFilter {
+
+    public void init(@Observes Filters filters) {
+        filters.register(rc -> {
```
gsmet: Is this ability to add filters documented somewhere?

gsmet (approved):

PR #4340, opened 2019-10-03
Made the tests pass for me

FroMage:
Now go figure why you guys can get your tests to run despite not having anyone set the `test.url` system property in that damn test.

### DIFF
--- a/integration-tests/main/src/test/java/io/quarkus/it/main/RestClientTestCase.java
+++ b/integration-tests/main/src/test/java/io/quarkus/it/main/RestClientTestCase.java
@@ -9,6 +9,7 @@ import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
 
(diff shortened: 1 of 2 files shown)


integration-tests/main/src/test/java/io/quarkus/it/main/RestClientTestCase.java
```diff
@@ -18,6 +19,10 @@
 
     public static final String HEADER_NAME = "header-name";
 
+    // this is required, otherwise the test.url system property is never set
+    @TestHTTPResource
```
stuartwdouglas: I think it would be better to change the code to always set the property.

stuartwdouglas (changes requested):

integration-tests/main/src/test/java/io/quarkus/it/main/RestClientTestCase.java
FroMage: Fine by me, but then where/how?

FroMage:
Closing in favour of #4390

Issue #3316, opened 2019-07-23
Allow configuring debug port in quarkus:dev without suspending JVM


devauxbr:
Hello

**Description**
Here is my problem : I am often running 2 quarkus projects in maven quarkus:dev mode at the same time. So in order to start them both with the JVM debugger enabled, I need to change the default debug mode of at least one of them by specifying a port different that the default one (5005), for instance `-Ddebug=1234` to get the debugger listening on port 1234

It all works well, except that this `-Ddebug=1234` parameter also **forces the debugger to suspend** ([see here](https://github.com/quarkusio/quarkus/blob/767bff502cfa7f91c348097fc272e4c8fb699f33/devtools/maven/src/main/java/io/quarkus/maven/DevMojo.java#L203)) and prevents quarkus to start until any debugger is attached. 
I often find my self waiting several seconds, thinking that quarkus is slowly booting, before realizing that I need to toggle the debugger in my IDE to actually let quarkus start, and this happens to me several times a day...

=> To improve the developer experience, it would be awesome to allow configuring debug port in quarkus:dev **without suspending JVM**

**Implementation ideas**
I was thinking one possible solution would be to introduce another `@Parameter` in the DevMojo, for instance `suspend`, so that if the developer provides `-Dsuspend=false` (or maybe `-Dsuspend=n` to be consistent with the JVM argument) then we force this JVM argument to be of value `n` ([here](https://github.com/quarkusio/quarkus/blob/767bff502cfa7f91c348097fc272e4c8fb699f33/devtools/maven/src/main/java/io/quarkus/maven/DevMojo.java#L188), [here](https://github.com/quarkusio/quarkus/blob/767bff502cfa7f91c348097fc272e4c8fb699f33/devtools/maven/src/main/java/io/quarkus/maven/DevMojo.java#L192), [here](https://github.com/quarkusio/quarkus/blob/767bff502cfa7f91c348097fc272e4c8fb699f33/devtools/maven/src/main/java/io/quarkus/maven/DevMojo.java#L195) and [here](https://github.com/quarkusio/quarkus/blob/767bff502cfa7f91c348097fc272e4c8fb699f33/devtools/maven/src/main/java/io/quarkus/maven/DevMojo.java#L203)). Otherwise we stick to the existing behaviour ?

If you agree with my proposal I may submit a PR implementing this idea in the days to come

In any case thanks for the awesomeness of Quarkus ! I am converting all my colleagues at work to use it :rocket:

[...]

PR #4353, opened 2019-10-03
Don't suspend the JVM launch when debug port is non-default

jaikiran:
As discussed here https://github.com/quarkusio/quarkus/issues/3316#issuecomment-537764463 the commit in this PR changes the launch configuration to not suspend the JVM, when a custom debug port is specified.

P.S: The `debug` also takes a `true` value and that configuration will continue to suspend by default. I didn't change it in this PR. If it feels right to make that consistent with the rest of these values, I  can change it too. That way, if someone really wants to suspend the JVM debug launch, they can do so by `debug=false` and `jvmArgs` using the exact arguments needed to launch and suspend the JVM in debug mode.

### DIFF
--- a/devtools/maven/src/main/java/io/quarkus/maven/DevMojo.java
+++ b/devtools/maven/src/main/java/io/quarkus/maven/DevMojo.java
@@ -96,7 +96,7 @@ public class DevMojo extends AbstractMojo {
      * </tr>
      * <tr>
      * <td><b>{port}</b></td>
-     * <td>The JVM is started in debug mode and suspends until a debugger is attached to {port}</td>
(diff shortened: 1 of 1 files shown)


gsmet (commented):
Open question: I wonder if we should add a `suspend` parameter.

jaikiran:
I had given that a thought, but wanted to avoid too many configuration options (`debug`, `jvmArgs` and now this potential `suspend`) and then their permutations to launch the JVM in debug mode. Having said that, if it is the right way to go, I'm willing to go ahead and add that enhancement. @gsmet, should I?

gsmet:
I think it would be worth it. It's not hard to understand.

jaikiran:
Ok. Closing this one and will open a new one with the suspend param.

PR #4356, opened 2019-10-03
Safely propagate the state of InvocationContextImpl accross threads

### LINKED ISSUE #4355: Safely propagate the state of InvocationContextImpl accross threads
The object does not have to be thread-safe though. Use case - SmallRye FT has an interceptor that calls `InvocationContext#process()` on a different thread.

mkouba:
- resolves #4355

### DIFF
--- a/independent-projects/arc/runtime/src/main/java/io/quarkus/arc/InvocationContextImpl.java
+++ b/independent-projects/arc/runtime/src/main/java/io/quarkus/arc/InvocationContextImpl.java
@@ -4,10 +4,13 @@ import java.lang.annotation.Annotation;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Function;
 import java.util.function.Supplier;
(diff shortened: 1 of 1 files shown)


kenfinnigan (approved):

mkouba:
There are some weird issues I need to investigate... :-(

mkouba:
It was actually a PEBKAC error... time to finish today ;-).

independent-projects/arc/runtime/src/main/java/io/quarkus/arc/InvocationContextImpl.java
```diff
@@ -82,9 +87,9 @@ public static InvocationContextImpl aroundConstruct(Constructor<?> constructor,
 
     private final Constructor<?> constructor;
 
-    private Object[] args;
+    private final List<Object> args;
 
-    private int position;
+    private final AtomicInteger position;
```
stuartwdouglas: Can you change this to a volatile integer with an AtomicIntegerUpdater instead? This object can be created a lot so it is better to save the allocation.

independent-projects/arc/runtime/src/main/java/io/quarkus/arc/InvocationContextImpl.java
```diff
@@ -82,9 +87,9 @@ public static InvocationContextImpl aroundConstruct(Constructor<?> constructor,
 
     private final Constructor<?> constructor;
 
-    private Object[] args;
+    private final List<Object> args;
```
stuartwdouglas: Why not just use a volatile field for the array?

stuartwdouglas (changes requested):

independent-projects/arc/runtime/src/main/java/io/quarkus/arc/InvocationContextImpl.java
stuartwdouglas: TBH I don't think you even need the updater, if you have multiple threads calling next() concurrently then you are into undefined territory anyway. Just making the field volatile should be enough to make sure the state is visible to a new thread.

mkouba:
Will be superseded by a PR for https://github.com/quarkusio/quarkus/issues/4552.

PR #4360, opened 2019-10-03
Support for the parser configuration in the properties

sberyozkin:
This PR does the following:
- support for the parser specific properties directly in the application.properties
- it temporarily uses the reflection code in the build phase to find the type of the parameter due to `TikaConfig` in Apache Tika not currently able to load the configuration with out the type attributes; it will be gone soon enough I hope

This is the final PR I'd like to merge to the Tika extension code before 0.24.0/1.0 and would only follow up with a simple user guide demo which is mostly ready


sberyozkin:
@dmlloyd Hi David, this is [the code](https://github.com/quarkusio/quarkus/pull/4360/files#diff-cefda81cecd3360b76139d0044ecb53aR162) I referred to earlier. If you think it can be moved to a common utility code, then I'll be happy to do it (and can replace some of that code with a more compact stream api version if preferred)

dmlloyd:
I don't think it needs to be a common utility at this time.  It looks fine though!

sberyozkin:
@dmlloyd sounds good, thanks

sberyozkin:
Hi Guillaume @gsmet If this code can make it to the master before 0.25.0 then it would be nice, as my talk is in 2 weeks on 23rd Oct :-)

docs/src/main/asciidoc/tika-guide.adoc
```diff
@@ -48,11 +48,22 @@ This will add the following to your pom.xml:
...
+
+[source]
+----
+quarkus.tika.parsers=pdf
+# Report the content in the order it occurs in the document
+# which can be useful for processing the invoices in PDF format 
+quarkus.tika.parser.pdf.sort-by-position=true
+----
+
+Please engage with Quarkus or Apache Tika communities to find out about a complete set of supported propertie for a given parser.
```
gsmet: I don't think this sentence makes sense? You expect users to ask the community what are the supported properties?

I think you need to provide some useful pointers.

extensions/tika/deployment/src/main/java/io/quarkus/tika/deployment/ParsersConfigBuildItem.java
```diff
@@ -0,0 +1,21 @@
+package io.quarkus.tika.deployment;
+
+import java.util.List;
+import java.util.Map;
+
+import io.quarkus.builder.item.SimpleBuildItem;
+import io.quarkus.tika.runtime.TikaParserParameter;
+
+public final class ParsersConfigBuildItem extends SimpleBuildItem {
```
gsmet: Any change we could name it `TikaParsersConfigBuildItem`?

gsmet (changes requested):
Added 2 minor comments inline, otherwise I suppose we could merge it soon.

gsmet:
@sberyozkin if you can fix these 2 issues early tomorrow morning and CI has the time to run, we can have this in 0.24.0 already.

extensions/tika/deployment/src/main/java/io/quarkus/tika/deployment/TikaProcessor.java
```diff
@@ -102,22 +113,41 @@ public void registerTikaProviders(BuildProducer<ServiceProviderBuildItem> servic
...
-            return providerNames.stream().filter(pred).filter(p -> requiredParsersFullNames.contains(p))
-                    .collect(Collectors.toList());
+            return providerNames.stream().filter(pred).filter(p -> fullNamesAndAbbreviations.containsKey(p))
+                    .collect(Collectors.toMap(Function.identity(),
+                            p -> getParserConfig(p, fullNamesAndAbbreviations.get(p))));
         }
     }
 
+    static List<TikaParserParameter> getParserConfig(String parserName, String abbreviation) {
+        final String prefix = "quarkus.tika.parsers." + abbreviation + ".";
```
gsmet: Just to be sure, when you use these properties, you don't have a bunch of warning saying the properties are unconsumed?

sberyozkin:
@gsmet, @aloubyansky sorry, totally missed it, myself to blame only :-). Let me address the comments and also prepare an accompanying quickstart demo for 0.25.0. I'll still have 1 week to build a talk around 0.25.0
Thanks

extensions/tika/deployment/src/main/java/io/quarkus/tika/deployment/TikaProcessor.java
sberyozkin: @gsmet thanks for asking, actually it does:
```
[WARNING] [io.quarkus.config] Unrecognized configuration key "quarkus.tika.parsers.pdf.sort-by-position" provided
```
Hmm...The problem is the list of parsers can be nearly open ended, and if I hide it back to `tika-config.xml` then there is no warning :-), but I'd like to 'break away' from `tika-config.xml`. 

I propose to just ignore such Warnings but also consider a quarkus config enhancement to tell it not to warn for a property whose name starts from, in this case, `quarkus.tika.parsers.pdf`. @dmlloyd FYI as well.
Does it sound reasonable ?

extensions/tika/deployment/src/main/java/io/quarkus/tika/deployment/ParsersConfigBuildItem.java
sberyozkin: Done

docs/src/main/asciidoc/tika-guide.adoc
sberyozkin: @gsmet I've given it hopefully a bit better wording. It is a bit tricky since I believe in most cases the users will be happy with just getting a text, customizing the parser requires some knowledge of how for example PDFBox works which Tika PDFParser delegates to and there is not that much info available specifically at the Tika site. And there could be so many parsers. 
So I've linked to the Tika source, alongside the suggestion to check with the community if required.
What do you think ?

extensions/tika/deployment/src/main/java/io/quarkus/tika/deployment/TikaProcessor.java
gsmet: So the idea when we have this type of pass-through properties is to not put them in the `quarkus.` namespace.

So you would have the `quarkus.tika.` properties for everything handled by Quarkus and the `tika.` prefix for everything that is pass-through.

We already do that for a few other extensions.

dmlloyd: Maybe it would be better to use e.g. a `Map<String,Map<String,String>>` property than a pass-through.  You all know how I feel about pass-through properties. :)

sberyozkin: @gsmet, OK, thanks, the parser property configuration is now done in the custom namespace.

docs/src/main/asciidoc/tika-guide.adoc
```diff
@@ -48,11 +48,23 @@ This will add the following to your pom.xml:
...
 |===
 
+== Parser configuration example
+
+[source]
+----
+quarkus.tika.parsers=pdf
+# Report the content in the order it occurs in the document
+# which can be useful for processing the invoices in PDF format 
+quarkus.tika.parser.pdf.sort-by-position=true
```
gsmet: This should be updated and the `quarkus.` prefix should be removed.

gsmet (changes requested):
I added one small comment about the doc.

@sberyozkin have you tried @dmlloyd 's suggestion with the `Map`? It might be another way to solve this issue and keep the `quarkus` namespace. It's probably better if we can make it work appropriately.

If you don't have time to figure it out, we can merge this for now for your talk and revisit it after your talk.

sberyozkin:
@gsmet, I'm looking at the Map approach and I'm not sure I can fix it quickly enough. Thanks for noticing this namespace issue, I was actually not entirely happy about it too, but could not see an alternative (see below, the map can't have a composite property name as a key)
Example:
```
quarkus.tika.parsers=pdf
``` 
That works. Now lets try to configure PDF and capture it into a Map (quarkus.tika.parser)
```
quarkus.tika.parsers=pdf
# This does not work:
# quarkus.tika.parser.pdf.sort-by-position=true
# But this works which is ok though the parser abbreviation and the parser property have to become a single property:
quarkus.tika.parser.pdf-sort-by-position=true
``` 
May be it is OK, but I'd like to discuss a bit more an option of suppressing the warnings (or may be only tracing it) when the properties not typed in Configuration are in the same namespace, example, `**quarkus.tika.parsers**=pdf` (typed in Configuration) and `**quarkus.tika.parsers**.pdf.sort-by-position=true` (not typed in Configuration but in the `quarkus.tika.parsers` namespace).
Lets indeed keep this issue for 0.26.0 but I'll very quickly open a PR just to add an `odf` shortcut to support the quickstart demo

sberyozkin:
@gsmet I finally rebased yesterday and updated `TikaConfiguration` to use the Maps (key the parser abbreviation, value - another Map of the parameter name to value pairs, etc, as per @dmlloyd suggetsion) to keep the parser configurations in the Quarkus namespace

extensions/tika/runtime/src/main/java/io/quarkus/tika/runtime/TikaConfiguration.java
```diff
@@ -26,23 +27,44 @@
...
     @ConfigItem
     public Optional<String> parsers;
 
+    /**
+     * Configuration of the individual parsers.
+     * For example:
+     *
+     * <pre>
+     * quarkus.tika.parsers = pdf,odf
+     * quarkus.tika.parser.pdf.sort-by-position = true
```
gsmet: I think `pre` needs closing in Javadoc. Could you adjust?

extensions/tika/runtime/src/main/java/io/quarkus/tika/runtime/TikaConfiguration.java
```diff
@@ -26,23 +27,44 @@
...
+    @ConfigItem
+    public Map<String, Map<String, String>> parser;
+
+    /**
+     * Full parser class name for a given parser abbreviation.
+     * For example:
+     *
+     * <pre>
+     * quarkus.tika.parsers = classparser
+     * quarkus.tika.parser-name.classparser = org.apache.tika.parser.asm.ClassParser
```
gsmet: I think `pre` needs closing in Javadoc. Could you adjust?

extensions/tika/runtime/src/main/java/io/quarkus/tika/runtime/TikaConfiguration.java
```diff
@@ -26,23 +27,44 @@
...
+     */
+    @ConfigItem
+    public Map<String, String> parserName;
```
gsmet: I would have named this one `parser` and the other one `parserOptions`.

That would be far more logical.

gsmet (changes requested):
I added a few minor comments and it needs a rebase.

@sberyozkin if you have 30mn, it would be nice to tackle this one and get it merged. That would be one less PR to manage.

Thanks!

extensions/tika/runtime/src/main/java/io/quarkus/tika/runtime/TikaConfiguration.java
sberyozkin: @gsmet, yes, I updated the code, makes perfect sense, thanks, however it looks like there is a minor regression on the config path...

sberyozkin:
@dmlloyd Hi David, the following stopped working for me:
```
/**
     * Configuration of the individual parsers.
     * For example:
     *
     * <pre>
     * quarkus.tika.parsers = pdf
     * quarkus.tika.parser-options.pdf.sort-by-position = true
     */
    @ConfigItem
    public Map<String, Map<String, String>> parserOptions;
```

Before the major config update I was seeing the map initialized as follows, given the above config fragment:

`pdf` (key):  `sort-by-position` = `true` (single entry map value, `sort-by-position` - key, `true` - value)

Now I'm seeing:

`sort-by-position` (key):  `sort-by-position` = `true` (single entry map value, `sort-by-position` - key, `true` - value)

so the `pdf` bit is lost.

What do you think, should an issue be opened ?
Thanks

dmlloyd:
Yes, that looks like a bug unfortunately, so that will need an issue.

dmlloyd:
The issue is #5982.

sberyozkin:
@gsmet I've rebased this PR

gsmet (approved):

gsmet:
I rebased and force-pushed. Let's wait for CI.

sberyozkin:
@gsmet thanks

sberyozkin:
@gsmet merge is done, thanks for the review :-)

PR #4363, opened 2019-10-03
support deployment of additional static resources beyond src/main/resources/META-INF/resources

### LINKED ISSUE #3886: Support deployment and hot deployment of additional web content beyond src/main/resources/META-INF/resources
**Description**
I am developing a web application on Quarkus like this [example](https://github.com/aaronanderson/quarkus-oidc/tree/master/example) that uses React and webpack. I have configured my JavaScript build process to publish resources to the supported src/main/resources/META-INF/resources directory and deployment and hot deployment work. However this location is intended to store version controlled sources and is not an ideal location for generated resources.

In addition to the standard web resource location above I would like for the Quarkus undertow extension to support an additional build directory for both packaging and hot deployment of dynamically generated content.  target/generated-resources is my suggestion.

**Implementation ideas**
Ideally both src/main/resources/META-INF/resources and target/generated-resources could be merged together for both packaging and hot deployment purposes. It appears knownDirectories is already built into the undertow deployment extension so perhaps this enhancement can be accomplished by simply modifying the maven/gradle plugins to support configuring it.

machi1990:
Fixes #3886

### DIFF
--- a/docs/src/main/asciidoc/http-reference.adoc
+++ b/docs/src/main/asciidoc/http-reference.adoc
@@ -24,10 +24,25 @@ Servlet filter, otherwise it will run directly on top of Vert.x with no Servlet 
 
 == Serving Static Resources
 
-To serve static resources you must place them in the `META-INF/resources` directory of your application. This location
+By default Quarkus will serve static files in the `META-INF/resources` directory of your application. This location
 was chosen as it is the standard location for resources in `jar` files as defined by the Servlet spec. Even though
 Quarkus can be used without Servlet following this convention allows existing code that places its resources in this
 location to function correctly.
+
+Additionally, Quarkus supports deployment of static files defined in a directory other than `META-INF/resources`.
+This is particularly useful when you want to leverage the power of a Quarkus backend alongside a frontend application (outside your Quarkus backend repository).
+In this workflow, you can configure your Quarkus backend using the configuration properties below to serve static files from the directory that contains your web application contents or the result of a frontend framework build process.
+
+[source,properties]
+----
+quarkus.http.static.web-content.endpoint=/web <1>
+quarkus.http.static.web-content.index=index.html <2>
+quarkus.http.static.web-content.path=/path/to/directory <3>
+----
+
+1. Configure a `/web` endpoint named `web-content` that will serve static web resources.
+2. Initialise the directory path containing static content for the `/web` endpoint.
+3. Set `index.html` as the default landing page for the `/web` endpoint.
 
 == Configuring the Context path
 
--- a/extensions/vertx-http/deployment/src/main/java/io/quarkus/vertx/http/deployment/VertxHttpProcessor.java
+++ b/extensions/vertx-http/deployment/src/main/java/io/quarkus/vertx/http/deployment/VertxHttpProcessor.java
@@ -3,7 +3,11 @@ package io.quarkus.vertx.http.deployment;
 import java.io.IOException;
 import java.util.Collections;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
+import java.util.TreeMap;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
 import org.eclipse.microprofile.config.ConfigProvider;
@@ -26,13 +30,17 @@ import io.quarkus.kubernetes.spi.KubernetesPortBuildItem;
 import io.quarkus.netty.runtime.virtual.VirtualServerChannel;
 import io.quarkus.runtime.LaunchMode;
 import io.quarkus.runtime.RuntimeValue;
+import io.quarkus.runtime.configuration.ConfigurationException;
 import io.quarkus.vertx.core.deployment.CoreVertxBuildItem;
 import io.quarkus.vertx.core.deployment.EventLoopCountBuildItem;
 import io.quarkus.vertx.core.deployment.InternalWebVertxBuildItem;
+import io.quarkus.vertx.http.deployment.devmode.NotFoundPageDisplayableEndpointBuildItem;
 import io.quarkus.vertx.http.runtime.CurrentVertxRequest;
 import io.quarkus.vertx.http.runtime.HttpBuildTimeConfig;
 import io.quarkus.vertx.http.runtime.HttpConfiguration;
 import io.quarkus.vertx.http.runtime.RouterProducer;
+import io.quarkus.vertx.http.runtime.StaticResourceConfig;
+import io.quarkus.vertx.http.runtime.StaticResourceRecorder;
 import io.quarkus.vertx.http.runtime.VertxHttpRecorder;
 import io.quarkus.vertx.http.runtime.cors.CORSRecorder;
 import io.quarkus.vertx.http.runtime.filters.Filter;
@@ -42,6 +50,7 @@ import io.vertx.ext.web.Router;
 import io.vertx.ext.web.RoutingContext;
 
 class VertxHttpProcessor {
+    private static Pattern MULTIPLE_TRAILING_SLASHES = Pattern.compile("(/+)$");
 
     @BuildStep
     HttpRootPathBuildItem httpRoot(HttpBuildTimeConfig httpBuildTimeConfig) {
@@ -106,6 +115,38 @@ class VertxHttpProcessor {
 
     @BuildStep
     @Record(ExecutionTime.RUNTIME_INIT)
+    void initializeAdditionalStaticResource(StaticResourceRecorder recorder, BuildProducer<RouteBuildItem> routes,
+            BuildProducer<NotFoundPageDisplayableEndpointBuildItem> displayableEndpoints,
+            HttpBuildTimeConfig httpBuildTimeConfig,
+            LaunchModeBuildItem launchModeBuildItem) {
+
+        Map<String, StaticResourceConfig.StaticResourceGeneralConfig> config = new TreeMap<>(
+                httpBuildTimeConfig.staticConfig.config);
+        for (Map.Entry<String, StaticResourceConfig.StaticResourceGeneralConfig> entry : config.entrySet()) {
+            StaticResourceConfig.StaticResourceGeneralConfig value = entry.getValue();
+            final String sanitizedEndpoint = sanitizeStaticResourceEndpoint(value.endpoint);
+            if (sanitizedEndpoint.isEmpty()) {
+                throw new ConfigurationException(
+                        String.format("The given static resource endpoint '%s' is invalid", value.endpoint));
+            }
+
+            Handler<RoutingContext> handler = recorder.handler(httpBuildTimeConfig, entry.getKey(), sanitizedEndpoint);
+            routes.produce(new RouteBuildItem(sanitizedEndpoint, handler));
+            routes.produce(new RouteBuildItem(sanitizedEndpoint + "/*", handler));
+            if (launchModeBuildItem.getLaunchMode().isDevOrTest()) {
+                displayableEndpoints.produce(new NotFoundPageDisplayableEndpointBuildItem(sanitizedEndpoint + "/"));
+            }
+        }
+    }
+
+    private String sanitizeStaticResourceEndpoint(String endpoint) {
+        final String staticResourceEndpoint = endpoint.startsWith("/") ? endpoint : "/" + endpoint;
+        Matcher matcher = MULTIPLE_TRAILING_SLASHES.matcher(staticResourceEndpoint);
+        return matcher.replaceAll("");
+    }
+
+    @BuildStep
+    @Record(ExecutionTime.RUNTIME_INIT)
     ServiceStartBuildItem finalizeRouter(
             VertxHttpRecorder recorder, BeanContainerBuildItem beanContainer,
             Optional<RequireVirtualHttpBuildItem> requireVirtual, InternalWebVertxBuildItem vertx,
@@ -117,9 +158,8 @@ class VertxHttpProcessor {
             List<WebsocketSubProtocolsBuildItem> websocketSubProtocols,
             List<RequireBodyHandlerBuildItem> requireBodyHandlerBuildItems,
             BodyHandlerBuildItem bodyHandlerBuildItem,
-            CoreVertxBuildItem core // Injected to be sure that Vert.x has been produced before calling this method.
-    )
-            throws BuildException, IOException {
+            CoreVertxBuildItem core, // Injected to be sure that Vert.x has been produced before calling this method.
+            EventLoopCountBuildItem eventLoopCountBuildItem) throws BuildException, IOException {
         Optional<DefaultRouteBuildItem> defaultRoute;
         if (defaultRoutes == null || defaultRoutes.isEmpty()) {
             defaultRoute = Optional.empty();
--- a/extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/StaticResourceConfig.java
+++ b/extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/StaticResourceConfig.java
@@ -1,0 +1,59 @@
+package io.quarkus.vertx.http.runtime;
+
+import java.util.Map;
+import java.util.Optional;
+import java.util.OptionalInt;
+
+import io.quarkus.runtime.annotations.ConfigDocMapKey;
+import io.quarkus.runtime.annotations.ConfigGroup;
+import io.quarkus.runtime.annotations.ConfigItem;
+
+@ConfigGroup
+public class StaticResourceConfig {
+
+    /**
+     * General static resource configuration
+     */
+    @ConfigDocMapKey("static-resource-name")
+    @ConfigItem(name = ConfigItem.PARENT)
+    public Map<String, StaticResourceGeneralConfig> config;
+
+    @ConfigGroup
+    public static class StaticResourceGeneralConfig {
+        /**
+         * The endpoint from which the static resources will be served.
+         */
+        @ConfigItem
+        public String endpoint;
+
+        /**
+         * Enable caching of static resources.
+         */
+        @ConfigItem(defaultValue = "true")
+        public boolean enableCache;
+
+        /**
+         * The maximum cache size.
+         */
+        @ConfigItem(defaultValue = "10000")
+        public OptionalInt cacheSize;
+
+        /**
+         * The default landing page of static resource.
+         */
+        @ConfigItem(defaultValue = "index.html")
+        public Optional<String> index;
+
+        /**
+         * Allow directory listing.
+         */
+        @ConfigItem
+        public boolean allowDirectoryListing;
+
+        /**
+         * The path to a directory from which the static resources will be served.
+         */
+        @ConfigItem
+        public String path;
+    }
+}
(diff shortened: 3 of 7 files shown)


geoand:
@machi1990 Nice one! However the windows tests seem to be failing consistently and you'll also have to rebase I'm afraid

machi1990:
> @machi1990 Nice one! However the windows tests seem to be failing consistently and you'll also have to rebase I'm afraid

Hi @geoand thanks, I noticed that I had an old branch and rebased like three times hence the force push. I'll deal with the conflicts late on, looks like I have to do another one either way :-).

The window problem looks like to be a Path problem, I do not have a window machine to deal with it correctly but I have some ideas that I'll try later today.

patriot1burke:
If you are adding another directory mounted on the root path, or on a path that will conflict with servlet or jaxrs endpoints, this new handler you're writing will have to come first as servlet and resteasy handle 404s themselves.

But the thing with that though is the StaticHandler in Vertx is not very performant on path misses.  Please look at ResteasyStandaloneRecorder and ResteasyStandaloneBuildStep.staticInit.  You'll see that I build up a set of valid paths and pass it to a wrapper around StaticResource to check for cache hits so that the static handler can be completely bypassed.

machi1990:
> If you are adding another directory mounted on the root path, or on a path that will conflict with servlet or jaxrs endpoints, this new handler you're writing will have to come first as servlet and resteasy handle 404s themselves.
> 
> But the thing with that though is the StaticHandler in Vertx is not very performant on path misses. Please look at ResteasyStandaloneRecorder and ResteasyStandaloneBuildStep.staticInit. You'll see that I build up a set of valid paths and pass it to a wrapper around StaticResource to check for cache hits so that the static handler can be completely bypassed.

Hi @patriot1burke, thanks for this insight. Do you think the best way is to check if the file can be served by the StaticHandler, if not call the next routing context? With something similar to this pseudo code

```
if (Files.exists(rootPath.resolves(ctx.normalisedPath())))) 
{
     staticResourceHandler.handle(ctx);
} else {
      ctx.next();
  }

``` 
?

machi1990:
@patriot1burke ping :-)

extensions/vertx-http/deployment/src/main/java/io/quarkus/vertx/http/deployment/VertxHttpProcessor.java
```diff
@@ -96,6 +102,28 @@ VertxWebRouterBuildItem initializeRouter(VertxHttpRecorder recorder,
...
+            routes.produce(new RouteBuildItem(staticResourceEndpoint, handler));
+            routes.produce(new RouteBuildItem(staticResourceEndpoint + "/*", handler));
+            displayableEndpoints.produce(new NotFoundPageDisplayableEndpointBuildItem(staticResourceEndpoint + "/"));
+        }
+    }
+
+    private String getStaticResourceEndpoint(StaticResourceBuildTimeConfig staticResourceBuildTimeConfig) {
+        String endpoint = staticResourceBuildTimeConfig.endpoint;
+        final String staticResourceEndpoint = endpoint.startsWith("/") ? endpoint : "/" + endpoint;
+        return staticResourceEndpoint.replaceAll("(/+)$", "");
```
gsmet: This regexp should be compiled.

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/static_resource/StaticResourceRecorder.java
```diff
@@ -0,0 +1,54 @@
...
+                if (routingContext.normalisedPath().equals(endpoint)) {
+                    routingContext.response().setStatusCode(302);
+                    routingContext.response().headers().set(HttpHeaders.LOCATION, endpoint + "/");
+                    routingContext.response().end();
+                    return;
+                }
+
+                final Path staticFilePath = staticResourceDir.resolve(
+                        routingContext.normalisedPath()
+                                .replaceAll("^" + endpoint + "/?", ""));
```
gsmet: Why a regexp and a replaceAll to remove the start of a string?

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/static_resource/StaticResourceRecorder.java
```diff
@@ -0,0 +1,54 @@
...
+        final StaticHandler staticHandler = StaticHandler.create()
+                .setAllowRootFileSystemAccess(true)
+                .setWebRoot(runtimeConfig.path)
+                .setDefaultContentEncoding("UTF-8")
+                .setIndexPage(buildTimeConfig.index);
+
+        return new Handler<RoutingContext>() {
+            @Override
+            public void handle(RoutingContext routingContext) {
+                if (routingContext.normalisedPath().equals(endpoint)) {
```
gsmet: Just to be sure `normalisedPath` exclude any query string parameter?

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/static_resource/StaticResourceRuntimeConfig.java
```diff
@@ -0,0 +1,14 @@
...
+import io.quarkus.runtime.annotations.ConfigPhase;
+import io.quarkus.runtime.annotations.ConfigRoot;
+
+@ConfigRoot(name = "static-resource", phase = ConfigPhase.RUN_TIME)
+public class StaticResourceRuntimeConfig {
+    /**
+     * The directory path from which the static resources will be served.
+     */
+    @ConfigItem
+    public String path;
```
gsmet: Shouldn't this be Optional?

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/static_resource/StaticResourceBuildTimeConfig.java
```diff
@@ -0,0 +1,24 @@
...
+import io.quarkus.runtime.annotations.ConfigPhase;
+import io.quarkus.runtime.annotations.ConfigRoot;
+
+@ConfigRoot(name = "static-resource", phase = ConfigPhase.BUILD_AND_RUN_TIME_FIXED)
+public class StaticResourceBuildTimeConfig {
+    /**
+     * The endpoint from which the static resources will be served.
+     */
+    @ConfigItem(name = ConfigItem.PARENT)
+    public String endpoint;
```
gsmet: Optional?

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/static_resource/StaticResourceBuildTimeConfig.java
```diff
@@ -0,0 +1,24 @@
...
+
+    /**
+     * The default landing page of static resource.
+     */
+    @ConfigItem(defaultValue = "index.html")
+    public String index;
```
gsmet: Optional?

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/static_resource/StaticResourceBuildTimeConfig.java
```diff
@@ -0,0 +1,24 @@
+package io.quarkus.vertx.http.runtime.static_resource;
+
+import io.quarkus.runtime.annotations.ConfigItem;
+import io.quarkus.runtime.annotations.ConfigPhase;
+import io.quarkus.runtime.annotations.ConfigRoot;
+
+@ConfigRoot(name = "static-resource", phase = ConfigPhase.BUILD_AND_RUN_TIME_FIXED)
+public class StaticResourceBuildTimeConfig {
+    /**
```
gsmet: I think I would make this a Map of potential endpoints to support several. But then I don't know if the Map is properly ordered. We need to make sure things are ordered so I would use a `SortedMap` and check the config code handles that properly.

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/static_resource/StaticResourceBuildTimeConfig.java
```diff
@@ -0,0 +1,24 @@
...
+     * The endpoint from which the static resources will be served.
+     */
+    @ConfigItem(name = ConfigItem.PARENT)
```
gsmet: I would avoid having this weird parent thing.

extensions/vertx-http/deployment/src/main/java/io/quarkus/vertx/http/deployment/VertxHttpProcessor.java
```diff
@@ -96,6 +102,28 @@ VertxWebRouterBuildItem initializeRouter(VertxHttpRecorder recorder,
...
+            Handler<RoutingContext> handler = recorder.handler(staticResourceBuildTimeConfig, staticResourceRuntimeConfig,
+                    staticResourceEndpoint);
+            routes.produce(new RouteBuildItem(staticResourceEndpoint, handler));
+            routes.produce(new RouteBuildItem(staticResourceEndpoint + "/*", handler));
+            displayableEndpoints.produce(new NotFoundPageDisplayableEndpointBuildItem(staticResourceEndpoint + "/"));
+        }
+    }
+
+    private String getStaticResourceEndpoint(StaticResourceBuildTimeConfig staticResourceBuildTimeConfig) {
+        String endpoint = staticResourceBuildTimeConfig.endpoint;
```
gsmet: Right now, I think if you define `/`as the path, it will work and it definitely should be absolutely avoided and throw a `ConfigurationException`.

gsmet (changes requested):
I added a bunch of comments.

extensions/vertx-http/deployment/src/main/java/io/quarkus/vertx/http/deployment/VertxHttpProcessor.java
machi1990: Good point.

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/static_resource/StaticResourceBuildTimeConfig.java
machi1990: Yes, I hadn't rebased with the new config PR.

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/static_resource/StaticResourceBuildTimeConfig.java
machi1990: Yes

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/static_resource/StaticResourceBuildTimeConfig.java
machi1990: I am thinking of having something like this for build time configuration. 

```
    /**
     * The endpoint from which the static resources will be served.
     */
    @ConfigDocMapKey("endpoint")
    public SortedMap<String, StaticResourceGeneralConfig> config;

    @ConfigGroup
    public static class StaticResourceGeneralConfig {
         @ConfigItem(name=ConfigItem.PARENT)
        public boolean enabled; 

        @ConfigItem
        public boolean enableCache; 
        
        @ConfigItem
        public Optional<MemorySize> cacheSize;
        
        @ConfigItem
        public Optional<String> index;
        
        @ConfigItem
        public boolean allowDirectoryListing;
    }
```

WDYT? 

The directory path for each endpoint can be made a runtime config to be overridable through env variable - thanks to @rsvoboda for the initial idea.

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/static_resource/StaticResourceBuildTimeConfig.java
```diff
@@ -0,0 +1,54 @@
...
+         * Enable caching of static resource
+         */
+        @ConfigItem(defaultValue = "true")
+        public boolean enableCache;
+
+        /**
+         * The default cache size.
+         */
+        @ConfigItem(defaultValue = "10K")
+        public Optional<MemorySize> cacheSize;
```
gsmet: I think this is a number of objects, not a memory size. Could you check?

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/static_resource/StaticResourceBuildTimeConfig.java
```diff
@@ -0,0 +1,54 @@
...
+    @ConfigGroup
+    public static class StaticResourceGeneralConfig {
+        /**
+         * Indicate if the static resource endpoint is enabled
+         */
+        @ConfigItem
+        public boolean enabled;
+
+        /**
+         * Enable caching of static resource
```
gsmet:
```suggestion
         * Enable caching of static resources.
```

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/static_resource/StaticResourceRecorder.java
```diff
@@ -0,0 +1,63 @@
...
+                    routingContext.response().setStatusCode(302);
+                    routingContext.response().headers().set(HttpHeaders.LOCATION, endpoint + "/");
+                    routingContext.response().end();
+                    return;
+                }
+
+                final Path staticFilePath = staticResourceDir
+                        .resolve(routingContext.normalisedPath().replaceFirst(endpoint + "/", ""));
```
gsmet: I suspect this won't work very well on Windows?

gsmet (changes requested):
I added a few more comments.

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/static_resource/StaticResourceBuildTimeConfig.java
```diff
@@ -0,0 +1,54 @@
...
+    @ConfigItem(name = ConfigItem.PARENT)
+    public Map<String, StaticResourceGeneralConfig> config;
+
+    @ConfigGroup
+    public static class StaticResourceGeneralConfig {
+        /**
+         * Indicate if the static resource endpoint is enabled
+         */
+        @ConfigItem
+        public boolean enabled;
```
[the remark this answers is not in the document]
machi1990: I was thinking more of enabling or disabling a resource depending on the environment. 

But let's make `endpoint` part of the config object too.

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/static_resource/StaticResourceBuildTimeConfig.java
machi1990: > I added a few more comments.

Thanks @gsmet let me check them.

machi1990:
> I added a few more comments.

Thanks @gsmet, let me check them.

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/static_resource/StaticResourceBuildTimeConfig.java
machi1990: It is indeed what you think it is. Good catch, thanks.

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/static_resource/StaticResourceRecorder.java
machi1990: The `StaticResourceTest` test did not fail in the CI. I guess we should be okay for Windows too.

docs/src/main/asciidoc/http-reference.adoc
```diff
@@ -24,11 +24,26 @@ Servlet filter, otherwise it will run directly on top of Vert.x with no Servlet
...
 Quarkus can be used without Servlet following this convention allows existing code that places its resources in this
 location to function correctly.
 
+Additionally, Quarkus supports deployment of static files defined in a directory other than `META-INF/resources`.
+This is particularly useful when you want to leverage the power of a Quarkus backend alongside a frontend application (outside your Quarkus backend repository).
+In this workflow, you can configure your Quarkus backend using the configuration properties below to serve static files from the directory that contains your web application contents or the result of a frontend framework build process.
+
+[source,properties]
+----
+quarkus.static-resource.web.endpoint=/web <1>
```
FroMage: The main prefix is called `quarkus.http.root-path` so it's a bit unfortunate that we're not using the same term to designate both paths. I guess I would have used `prefix` instead of `endpoint` or `path`, myself.

Should we not use the quarkus.http.static` config prefix instead of `quarkus.static-resource`?

machi1990: Yes, I'll review the config knobs and update the change.

sebthom:
@machi1990 why did you close your PR?

tassadar81:
Hello,

the properties mentioned are supported? I can't find a clue in the documentation

quarkus.http.static.web-content.endpoint=/web
quarkus.http.static.web-content.index=index.html
quarkus.http.static.web-content.path=/path/to/directory

machi1990:
> @machi1990 why did you close your PR?

Hi @sebthom the PR was a bit old, however I have left the branch in my fork and I am happy to mentor / guide if you want to take over the work.

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/StaticResourceConfig.java
```diff
@@ -0,0 +1,59 @@
...
+import java.util.Map;
+import java.util.Optional;
+import java.util.OptionalInt;
+
+import io.quarkus.runtime.annotations.ConfigDocMapKey;
+import io.quarkus.runtime.annotations.ConfigGroup;
+import io.quarkus.runtime.annotations.ConfigItem;
+
+@ConfigGroup
+public class StaticResourceConfig {
```
machi1990, line 12: @gsmet this is the configuration group evoked in https://github.com/quarkusio/quarkus/pull/4363/files#r483076835

machi1990:
> What is the status of this feature? I need to have additional static resource folders outside of the Quarkus jar (e.g. META-INF/resources). Was this implemented? I see its closed but I don't see this in the code. Is there another way to accomplish this?

Hi @dhoffer the issue is under heavy consideration from the team. It should not be long before this work is revived. 

As for a workaround @DaveO-Home might have merged to come up with something custom (application based setup ) in this https://github.com/quarkusio/quarkus/issues/3886#issuecomment-643839426. You might want to look at that for the time being

PR #4364, opened 2019-10-03
Fix formatting typo in build section

aguibert:


geoand (approved):
Thanks for catching these!

Issue #1793, opened 2019-04-01
publish gradle plugin to the gradle plugin portal


cniweb:
Publish gradle plugin to the gradle plugin portal.

How it works, see:
https://plugins.gradle.org/docs/submit

stalep:
It is also in the todo :)

Sanne:
Added it to the list of issues we should have for 1.0

Dufgui:
I create a merge request to have better information display in gradle portal.
So now you must create a quarkus release account in gradle portal : 
https://plugins.gradle.org/user/register

Get the API Keys :
```gradle.publish.key=<key>
gradle.publish.secret=<secret>
```

So to do it, during the release process, you must call in `devtools/gradle` (with key and secret replace by the previous API Keys values) :
```
gradle publishPlugins -Pgradle.publish.key=<key> -Pgradle.publish.secret=<secret>
```
No more

Sanne:
Reopening as it's not published yet.

gsmet:
Raaah, it was closed automatically by the PR. Thanks @Sanne

patoessy:
Why is this always been moved to another milestone? Is it always hard to publish?

patoessy:
How can I assist?

maxandersen:
duplicate of https://github.com/quarkusio/quarkus/issues/4311 which has a bit more context of why its relevant to push.

PR #4373, opened 2019-10-04
fix templates to use gradle

maxandersen:


devtools/common/src/main/resources/templates/basic-rest/kotlin/build.gradle-template.ftl
```diff
@@ -1,20 +1,8 @@
...
-        mavenLocal()
-    }
-    dependencies {
-        classpath "io.quarkus:quarkus-gradle-plugin:${quarkusVersion}"
-    }
-}
-
 plugins {
     id 'kotlin'
+    id 'io.quarkus'
```
Sanne: cool! Does it mean the build plugin was finally deployed on the plugins portal?

devtools/common/src/main/java/io/quarkus/cli/commands/file/GradleBuildFile.java
```diff
@@ -62,21 +62,25 @@ public void completeFile(String groupId, String artifactId, String version) thro
...
-            res.append("buildscript {").append(System.lineSeparator());
-            res.append("    repositories {").append(System.lineSeparator());
-            res.append("        mavenLocal()").append(System.lineSeparator());
-            res.append("    }").append(System.lineSeparator());
-            res.append("    dependencies {").append(System.lineSeparator());
-            res.append("        classpath \"io.quarkus:quarkus-gradle-plugin:").append(getPluginVersion()).append("\"")
-                    .append(System.lineSeparator());
-            res.append("    }").append(System.lineSeparator());
+            res.append("repositories {").append(System.lineSeparator());
+            res.append("    mavenLocal()").append(System.lineSeparator());
```
Sanne: it's NOT considered a good practice in the Gradle world to rely on the maven local repo. Could we avoid this one?

geoand: I think we need this however in order to test snapshot versions of the plugin

geoand: The alternative I see would be to move the plugin out of the main repo like @aloubyansky has mentioned. That way the tests would never have to use snapshot version.

geoand:
@maxandersen there are formatting errors. You can fix them by doing `mvn process-resources`

devtools/common/src/main/resources/templates/basic-rest/kotlin/build.gradle-template.ftl
gsmet: No it's not. I need to include that in the release process. I'll try to work on it early next week for the next release.

devtools/common/src/main/java/io/quarkus/cli/commands/file/GradleBuildFile.java
Sanne: > I think we need this however in order to test snapshot versions of the plugin

No, to test snapshot versions of the plugin you should override your local work settings.

+1 to not have the Gradle plugin in a Maven repo.. there's no reason for that.

devtools/common/src/main/resources/templates/basic-rest/kotlin/build.gradle-template.ftl
Sanne: Ok. So I guess the template is crafted like this so that it _will_ work after that?

aloubyansky: It should work. I tested it. One of the reasons we had the legacy way of applying the plugin was due to a bug in Gradle which was fixed in 5.5.1.

maxandersen:
> @maxandersen there are formatting errors. You can fix them by doing `mvn process-resources`

reformatted and rebased with master.

devtools/common/src/main/java/io/quarkus/cli/commands/file/GradleBuildFile.java
maxandersen: to get that we would need #1793 fixed, right ?

Sanne: Maybe I misunderstood - but I thought this section wasn't about where to download plugins from, but where to download the actual application dependencies from?

If it's not relating to the plugins, then it's not depending on #1793 : Gradle has a different local cache which is more efficient and ensures correctnes (e.g. it won't cache http error codes as it were valid jars, and it can better track differences from different sources), so you don't want it to enforce using the local Maven repositories.

maxandersen: i'm okey removing it if can explain how we can get quarkus core builds locally by mvn seen by that gradle cache ? 

that’s really the only reason the setup is there - to be able to build quarkus core and have them pick up by the gradle based examples.

maxandersen: btw. this was already there before so i I wouldn't block this PR on it as this PR makes the generated gradle projects actulally work where they failed before.

Sanne: > i'm okey removing it if can explain how we can get quarkus core builds locally by mvn seen by that gradle cache ?

When testing snapshot you would indeed need to add this line. It just doesn't belong there for end users though:

- they are supposed to use a tag, not a snapshot
- they are supposed to download such tagged dependencies from somewhere else

Also a problem: listing "local maven" as a valid source implies snapshots might not get refreshed; when you list it as a source the tool takes your word for it to be authoritative - not just a cache.

> btw. this was already there before so i I wouldn't block this PR on it as this PR makes the generated gradle projects actulally work where they failed before.

sure, didn't mean to block the PR. I was just having a quick look and hoping this was sorted - my comment was only meant to be a reminder.

Just please all keep in mind that's a horrible thing to do in terms of Gradle sane practices - expect some outrage if it's not sorted :)

Sanne: > When testing snapshot you would indeed need to add this line.

Actually let me qualift that. You'd only need this line if you expect to test Quarkus by "deploying" a snapshot in your local repository - so this requirements depend on the assumption of a typical Maven workflow, but that's not the only way to do things.

There are some alternatives, such as deploying to a "local staging" (nice and clean as you can wipe it out w/o destroying your local cache), or having Gradle pointed to the sources: pointing the gradle integration tests to the Quarkus sources you don't have to rebuild the snapshots at all, it will check for source code changes and only rebuild the classes as necessary; this would skip jar creations and possibly skip all project modules which aren't necessary to run the selected integration tests - when you apply such solutions at scale with larger projects, you can save massive amounts of time. It requires some extra setup though, so to have it consume the Maven project as sources so I wouldn't suggest doing this - would be easy if the consumed projects were also built by Gradle.

maxandersen:
many other things done to the templates since then thus closing this as out-of-date.

Issue #4349, opened 2019-10-03
Add HTTP Permission Layer


stuartwdouglas:
There should be a generic interface that can be implemented to managed HTTP path permissions, that decide if a given SecurityIdentity can access a provided path+method combo.

By default this should support:
- Servlet web.xml mappings, so even if a security constraint is defined in web.xml and the target of the request is not a Servlet the constraint still takes effect. 
- application.properties mappings, a simple way to restrict access to paths based on application.properties.

This will also support user defined constraints by allowing the user to implement an interface with a CDI bean.

sberyozkin:
@stuartwdouglas and as you suggested, lets have it pluggable so that the OIDC adapter can provide its own UMA/etc mechanism, @pedroigor FYI

dfranssen:
Currently we have different microservices with similar `application.properties` as below. Authorization is externalised in Keycloak with policies, permissions and scopes ([see Keycloak doc](https://www.keycloak.org/docs/7.0/authorization_services/)).
With this approach there are no explicit role checks in the different rest endpoints and fine-grained authorization is externalized. 

With the latest changes in Quarkus (v0.24) it seems that the `keycloak` extension is renamed to `oidc` and the specific Keycloak functionality as policyEnforcer has been removed. **This is a major issue as our security approach is now completely obsolete once we update the quarkus version.**

Is it possible to keep the keycloak extension next to the oidc one until it has been clarified how to move forward with this issue as it is labeled version 1.0? 
FYI @sberyozkin @pedroigor 

```
quarkus.keycloak.realm=my-realm
quarkus.keycloak.auth-server-url=http://localhost:8081/auth
quarkus.keycloak.resource=assets-resource-server
quarkus.keycloak.credentials.secret=my-secret

quarkus.keycloak.policy-enforcer.claim-information-point.claims.request-uri={request.relativePath}
quarkus.keycloak.policy-enforcer.claim-information-point.claims.request-method={request.method}
quarkus.keycloak.policy-enforcer.lazy-load-paths=true

quarkus.keycloak.policy-enforcer.paths.1.path=/asset-service/resources/tenants
quarkus.keycloak.policy-enforcer.paths.1.methods.1.method=POST
quarkus.keycloak.policy-enforcer.paths.1.methods.1.scopes=tenants:create

quarkus.keycloak.policy-enforcer.paths.10.path=/asset-service/resources/tenants/{tenantId}
[...]

PR #4376, opened 2019-10-04
Rename Keycloak extension to quarkus-oidc

sberyozkin:


sberyozkin:
Hi @stuartwdouglas, @gsmet, @pedroigor  I have renamed the module, the configuration root from 'keycloak' to 'oidc', and did some initial doc updates, but @pedroigor would be the best person to complete the docs once the adapter review is complete.

sberyozkin:
I'll need to rename 'integration-tests/oidc' back to 'integration-tests/keycloak' to get the tests passing with the pipeline expecting the tests at 'integration-tests/keycloak', it is OK I guess to leave the integration test module unchanged

gsmet:
You can change it here: https://github.com/quarkusio/quarkus/blob/master/azure-pipelines.yml#L245 .

sberyozkin:
@gsmet sounds good, sorry for a 'force-push' noise :-).

stuartwdouglas (approved):

stuartwdouglas:
@sberyozkin @pedroigor  we should also rename the properties in the config to the OpenID standard names rather than Keycloak ones as was mentioned on the call. It would be great if this could happen before the 0.24 release, will one of you have time to look at this?

sberyozkin:
@stuartwdouglas thanks for the approval; re the property names - they look neutral enough to me at the moment, they are just preparing the adapter to verify the token, and each adapter would usually have some similar adapter specific properties, no standard names exists in the OIDC spec for properties like the IDP address, etc; 
I can definitely check which properties which are currently used in the tests are actually needed for the tests to pass, like I'm not sure about the bearer-only properties, etc.  @pedroigor have a look please as well and see if some of the properties which are currently being used can be renamed somehow...

dfranssen:
What about the previous integration of Keycloak Authorization via the policy-enforcer? Is this not supported anymore as of v0.24? We have plenty of methods and scopes in the application.properties

gsmet:
@dfranssen please create a new issue with your requirements with as many details as possible. Thanks!

We are still in the process of addressing this.

stuartwdouglas:
I think https://github.com/quarkusio/quarkus/issues/4349 will be what you are after, can you leave a comment there with exactly what you need?

dfranssen:
I commented on #4349, bottom line: With the latest changes in Quarkus (v0.24) it seems that the keycloak extension is renamed to oidc and the specific Keycloak functionality as policyEnforcer has been removed. This is a major issue as our security approach is now completely obsolete once we update the quarkus version. Is it possible to keep the keycloak extension next to the oidc one?

stuartwdouglas:
Is there anything else missing other than what is mentioned on #4349? There is a good chance we can basically replace the functionality in the next release (so a weeks time). Initial work for 'simple' permissions just went in to https://github.com/quarkusio/quarkus/pull/4487, however it is designed to be 100% pluggable.

stuartwdouglas:
Also it looks like you are mapping scope based permissions to JAX-RS resources? If so would an interceptor/annotation based model be a better fit?

dfranssen:
Yes, an annotation on the rest resource where multiple policies can be listed (with or functionality) would be less verbose. 

Also mind those settings are required in order to use the path in a Keycloak javascript permission:
‘’’
quarkus.keycloak.policy-enforcer.claim-information-point.claims.request-uri={request.relativePath}
quarkus.keycloak.policy-enforcer.claim-information-point.claims.request-method={request.method}
‘’’

sberyozkin:
@dfranssen Please see https://github.com/quarkusio/quarkus/issues/4495

PR #4382, opened 2019-10-04
Add io pause

patriot1burke:
* When adding a vertx filter like elytron, request bodies were being thrown away before resteasy had a chance to register request handlers.
* Also, elytron had a dependency on resteasy-deployment so I removed that.

-- pause code pulled in from stuarts branch I just added tests and remove the resteasy dependency.


gsmet:
@stuartwdouglas what's the status of this one?

PR #4384, opened 2019-10-04
Fix minor wording issues in integration tests README.md

gsmet:
Very minor but posted my review too late.

### DIFF
--- a/integration-tests/reactive-mysql-client/README.md
+++ b/integration-tests/reactive-mysql-client/README.md
@@ -4,7 +4,7 @@
 
 By default, the tests of this module are disabled.
 
-To run the tests in a standard JVM with MySQL started as a Docker container, you can run the following command:
+To run the tests in a standard JVM with MariaDB started as a Docker container, you can run the following command:
 
 ```
 mvn clean install -Dtest-mariadb -Ddocker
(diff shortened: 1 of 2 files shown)


kenfinnigan (approved):

integration-tests/reactive-mysql-client/README.md
```diff
@@ -4,7 +4,7 @@
 
 By default, the tests of this module are disabled.
 
-To run the tests in a standard JVM with MySQL started as a Docker container, you can run the following command:
+To run the tests in a standard JVM with MariaDB started as a Docker container, you can run the following command:
```
Sanne: This module is called `reactive-mysql-client`: why is it using MariaDB for testing?

gsmet: Because I suppose it supports both? I haven't changed it, just made the README state what the code does. @tsegismont wrote this IIRC.

FroMage:
Yeah, it's confusing that the module name is one thing and the README talks about something else.

gsmet:
Well, the README says what we're doing, which is less confusing than before. You can change the ITs to use MySQL but my understanding was that we were favoring MariaDB over MySQL.

PR #4388, opened 2019-10-04
Workaround backtick and tick issue

mejlholm:
Seems that asciidoc cannot handle backtics followed by tics. In this case the formatting continues until `Pod` is reached. 
Commit works around the problem by reformulating slightly, I don't know ascii doc well enough to know if there is a better way to escape the tick.

### DIFF
--- a/docs/src/main/asciidoc/kubernetes-client.adoc
+++ b/docs/src/main/asciidoc/kubernetes-client.adoc
@@ -142,7 +142,7 @@ Note that to take advantage of these features, the `quarkus-test-kubernetes-clie
 == Note on implementing the Watcher interface
 
 Due to the restrictions imposed by GraalVM, extra care needs to be taken when implementing a `io.fabric8.kubernetes.client.Watcher` if the application is intended to work in native mode.
-Essentially every `Watcher` implementation needs to specify the Kubernetes model class that it handles via the `Watcher`'s generic type at class definition time.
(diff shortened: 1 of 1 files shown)


mejlholm:
Looks like other alternative is to escape tick with &#8217;

stuartwdouglas (approved):

docs/src/main/asciidoc/kubernetes-client.adoc
```diff
@@ -142,7 +142,7 @@ Note that to take advantage of these features, the `quarkus-test-kubernetes-clie
 == Note on implementing the Watcher interface
 
 Due to the restrictions imposed by GraalVM, extra care needs to be taken when implementing a `io.fabric8.kubernetes.client.Watcher` if the application is intended to work in native mode.
-Essentially every `Watcher` implementation needs to specify the Kubernetes model class that it handles via the `Watcher`'s generic type at class definition time.
+Essentially every `Watcher` implementation needs to specify the Kubernetes model class that it handles via the `Watcher`&#8217;s generic type at class definition time.
```
gsmet: IIRC, in this case, you should use:
```
``Watcher``'s
```
with 2 backticks.

mejlholm: I thought i had tried that, guess not. Oh well, I've changed the doc and created new pull request.

PR #4395, opened 2019-10-06
Add "suspend" param to control the debug mode launch

jaikiran:
As recommened in this PR https://github.com/quarkusio/quarkus/pull/4353, the commit here introduces a `suspend` param to the `DevModeMojo` to allow control over how the JVM is launched in debug mode. This param is defaulted to `n` to allow the usecase noted here https://github.com/quarkusio/quarkus/issues/3316#issuecomment-537764463

### DIFF
--- a/devtools/maven/src/main/java/io/quarkus/maven/DevMojo.java
+++ b/devtools/maven/src/main/java/io/quarkus/maven/DevMojo.java
@@ -18,6 +18,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 import java.util.Set;
 import java.util.jar.Attributes;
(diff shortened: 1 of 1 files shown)


devtools/maven/src/main/java/io/quarkus/maven/DevMojo.java
```diff
@@ -103,6 +104,36 @@
...
+     * Whether or not the JVM launch, in debug mode, should be suspended. This parameter is only
+     * relevant when the JVM is launched in {@link #debug debug mode}. This parameter supports the
+     * following values (all the allowed values are case insensitive):
+     * <table>
+     * <th>
+     * <td>Value</td>
+     * <td>Effect</td>
+     * </th>
+     * <tr>
+     * <td>true</td>
```
gsmet: Shouldn't we just have `true or y` and `false or n` and thus only 2 lines? I think it would be a bit better.

gsmet (changes requested):
Looks good thanks. I added one small comment about the doc. Can you have a look?

devtools/maven/src/main/java/io/quarkus/maven/DevMojo.java
jaikiran: Done. I've updated this PR with this change.

gsmet (approved):

jaikiran:
> @jaikiran I think it makes sense to do the same modification and doc update Gradle. Would you like to take care of that?

@geoand I just submitted a PR for this https://github.com/quarkusio/quarkus/pull/4631. I haven't yet had a chance to test it out but I can do some manual testing later tonight when I find some time.

PR #4397, opened 2019-10-06
Maven plugin: Support <compilerArgs/>, <source/>, <target/> in dev mode

benkard:
Adds support for the `<compilerArgs/>`, `<source/>`, and `<target/>`
settings to the Maven plugin, which are then used to configure javac
in development mode.

This enables users to set lint flags and enable preview language
features, among other things.

### DIFF
--- a/core/devmode/src/main/java/io/quarkus/dev/CompilerFlags.java
+++ b/core/devmode/src/main/java/io/quarkus/dev/CompilerFlags.java
@@ -1,0 +1,75 @@
+package io.quarkus.dev;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+import io.quarkus.runtime.util.StringUtil;
+
+/**
+ * A set of compiler flags for javac.
+ *
+ * Can combine system-provided default flags with user-supplied flags and <code>-source</code>
+ * and <code>-target</code> settings.
+ */
+public class CompilerFlags {
+
+    private final Set<String> defaultFlags;
+    private final List<String> userFlags;
+    private final String sourceJavaVersion; //can be null
+    private final String targetJavaVersion; //can be null
+
+    public CompilerFlags(
+            Set<String> defaultFlags,
+            Collection<String> userFlags,
+            String sourceJavaVersion,
+            String targetJavaVersion) {
+
+        this.defaultFlags = defaultFlags == null ? new HashSet<>() : new HashSet<>(defaultFlags);
+        this.userFlags = userFlags == null ? new ArrayList<>() : new ArrayList<>(userFlags);
+        this.sourceJavaVersion = sourceJavaVersion;
+        this.targetJavaVersion = targetJavaVersion;
+    }
+
+    public List<String> toList() {
+        List<String> flagList = new ArrayList<>();
+
+        // The set of effective default flags is the set of default flags except the ones also
+        // set by the user.  This ensures that we do not needlessly pass the default flags twice.
+        Set<String> effectiveDefaultFlags = new HashSet<>(this.defaultFlags);
+        effectiveDefaultFlags.removeAll(userFlags);
+
+        flagList.addAll(effectiveDefaultFlags);
+
+        // Add -source and -target flags.
+        if (sourceJavaVersion != null) {
+            flagList.add("-source");
+            flagList.add(sourceJavaVersion);
+        }
+        if (targetJavaVersion != null) {
+            flagList.add("-target");
+            flagList.add(targetJavaVersion);
+        }
+
+        flagList.addAll(userFlags);
+
+        return flagList;
+    }
+
+    @Override
+    public int hashCode() {
+        return toList().hashCode();
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+        return obj instanceof CompilerFlags && toList().equals(((CompilerFlags) obj).toList());
+    }
+
+    @Override
+    public String toString() {
+        return "CompilerFlags@{" + StringUtil.join(", ", toList().iterator()) + "}";
+    }
+}
--- a/core/devmode/src/main/java/io/quarkus/dev/JavaCompilationProvider.java
+++ b/core/devmode/src/main/java/io/quarkus/dev/JavaCompilationProvider.java
@@ -8,7 +8,7 @@ import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.Arrays;
 import java.util.Collections;
-import java.util.List;
+import java.util.HashSet;
 import java.util.Set;
 
 import javax.tools.Diagnostic;
@@ -28,7 +28,7 @@ public class JavaCompilationProvider implements CompilationProvider {
     // -g is used to make the java compiler generate all debugging info
     // -parameters is used to generate metadata for reflection on method parameters
     // this is useful when people using debuggers against their hot-reloaded app
-    private static final List<String> COMPILER_OPTIONS = Arrays.asList("-g", "-parameters");
+    private static final Set<String> COMPILER_OPTIONS = new HashSet<>(Arrays.asList("-g", "-parameters"));
 
     @Override
     public Set<String> handledExtensions() {
@@ -48,9 +48,12 @@ public class JavaCompilationProvider implements CompilationProvider {
             fileManager.setLocation(StandardLocation.CLASS_PATH, context.getClasspath());
             fileManager.setLocation(StandardLocation.CLASS_OUTPUT, Collections.singleton(context.getOutputDirectory()));
 
+            CompilerFlags compilerFlags = new CompilerFlags(COMPILER_OPTIONS, context.getCompilerOptions(),
+                    context.getSourceJavaVersion(), context.getTargetJvmVersion());
+
             Iterable<? extends JavaFileObject> sources = fileManager.getJavaFileObjectsFromFiles(filesToCompile);
             JavaCompiler.CompilationTask task = compiler.getTask(null, fileManager, diagnostics,
-                    COMPILER_OPTIONS, null, sources);
+                    compilerFlags.toList(), null, sources);
 
             if (!task.call()) {
                 throw new RuntimeException("Compilation failed" + diagnostics.getDiagnostics());
--- a/core/devmode/src/test/java/io/quarkus/dev/CompilerFlagsTest.java
+++ b/core/devmode/src/test/java/io/quarkus/dev/CompilerFlagsTest.java
@@ -1,0 +1,91 @@
+package io.quarkus.dev;
+
+import static org.junit.jupiter.api.Assertions.assertAll;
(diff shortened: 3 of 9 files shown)


geoand:
Thanks for the PR @benkard!

Actually I am not sure if this is the way we want to go. Ideally the dev mode would just pick up the settings that the compiler has been configured with without having to provide such configuration again the quarkus plugins's configuration.

@aloubyansky @stuartwdouglas WDYT?

gsmet:
Yes, I agree we would need to be consistent with what is configured in the Maven compiler plugin (in the Maven case).

benkard:
`source` and `target` already default to `${maven.compiler.source}` and `${maven.compiler.target}`, respectively, so those should be covered.

As for the `compilerArgs`, I agree it would be best if they defaulted to whatever the `maven-compiler-plugin` is configured to use.  It would certainly cover my own use case.

On the other hand, I imagine it could still be useful to be able to configure something else, as in theory you might be using a different compiler plugin (the Takari lifecycle plugin comes to mind, but there may be others).  Or you may want to leave out flags that make compilation slower.  Not sure how often that's relevant.

I'll see if I can take a stab at pulling in the `maven-compiler-plugin`'s configuration.

benkard:
With the latest patch, the plugin now defaults to the Maven Compiler Plugin's configuration.  The code to do this isn't very pretty, but it seems to work.

gsmet:
@geoand do you think you would have the time to drive this puppy home? It looks like a very valuable addition.

geoand:
@gsmet of course. I actually took a quick look at it earlier and it looked nice.

I'll take a closer look tomorrow

core/devmode/src/main/java/io/quarkus/dev/JavaCompilationProvider.java
```diff
@@ -48,9 +49,21 @@ public void compile(Set<File> filesToCompile, Context context) {
             fileManager.setLocation(StandardLocation.CLASS_PATH, context.getClasspath());
             fileManager.setLocation(StandardLocation.CLASS_OUTPUT, Collections.singleton(context.getOutputDirectory()));
 
+            List<String> compilerOptions = new ArrayList<String>();
```
geoand: Shouldn't we make this a `Set` now that arbitrary options are coming in?

benkard: The order of arguments might matter in some cases (`-source` would a good example were it not handled in a special way here: its argument has to come right after the option; it can also share the same argument as `-target`, so even an ordered set would break it).  I see what you mean, though.  Maybe it would make sense to do something smarter than a simple `addAll` of the fixed `COMPILER_OPTIONS`.

geoand: How about a `LinkedHashSet` which preserved order?

dmlloyd: Still no good: `-target` `8` `-target` `9` yields `-target` `8` `9`.

geoand: Great point. So we'll need some more sophisticated handling here.

benkard: I put the more sophisticated flag handling in a separate class (`CompilerFlags`) and made a unit test for it.

core/devmode/src/main/java/io/quarkus/dev/CompilerFlags.java
```diff
@@ -0,0 +1,85 @@
...
+
+/**
+ * A set of compiler flags for javac.
+ *
+ * Can combine system-provided default flags with user-supplied flags and <code>-source</code>
+ * and <code>-target</code> settings.
+ */
+public class CompilerFlags {
+
+    @NotNull
```
geoand: I don't think these annotations provide any benefit. Or am I missing something?

core/devmode/src/test/java/io/quarkus/dev/CompilerFlagsTest.java
```diff
@@ -0,0 +1,86 @@
+package io.quarkus.dev;
+
+import static org.junit.jupiter.api.Assertions.*;
```
geoand: Can you please replace with the regular imports? We try to not use start imports

core/devmode/src/main/java/io/quarkus/dev/CompilerFlags.java
benkard: They document the invariants of the class.  But they don’t do anything, so if you prefer, I’ll remove them.

core/devmode/src/test/java/io/quarkus/dev/CompilerFlagsTest.java
benkard: Done.

core/devmode/src/main/java/io/quarkus/dev/CompilerFlags.java
benkard: Done.

geoand: That's what I assumed, but since we aren't using them elsewhere, best be uniform.

core/devmode/src/test/java/io/quarkus/dev/CompilerFlagsTest.java
```diff
@@ -0,0 +1,87 @@
...
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+import org.junit.jupiter.api.Test;
+
+public class CompilerFlagsTest {
```
geoand: :+1:

core/devmode/src/test/java/io/quarkus/dev/CompilerFlagsTest.java
```diff
@@ -0,0 +1,87 @@
...
+                () -> assertEquals(
+                        new CompilerFlags(setOf("-a", "-b"), listOf(), null, null),
+                        new CompilerFlags(setOf(), listOf("-a", "-b"), null, null)),
+                () -> assertEquals(
+                        new CompilerFlags(setOf("-a", "-b", "-c"), listOf("-a", "-b"), null, null),
+                        new CompilerFlags(setOf("-c"), listOf("-a", "-b"), null, null)));
+    }
+
+    @Test
+    void sourceAndTarget() {
```
geoand: What would happen if `--source` is set but the source flag is also set in maven's `compilerArgs`?

devtools/maven/src/main/java/io/quarkus/maven/DevMojo.java
```diff
@@ -239,6 +260,34 @@ public void execute() throws MojoFailureException, MojoExecutionException {
             }
             devModeContext.getBuildSystemProperties().putAll((Map) project.getProperties());
             devModeContext.setSourceEncoding(getSourceEncoding());
+            devModeContext.setSourceJavaVersion(source);
```
geoand: Actually I don't think it should be called `sourceJavaVersion` since it should (in a future PR) be able to handle Kotlin and Scala as well, right?

geoand:
@benkard when we have everything resolved, it would also be good to have the commits squashed to a single one for this feature

devtools/maven/src/main/java/io/quarkus/maven/DevMojo.java
benkard: I don't know whether the Kotlin and Scala compilers have equivalent flags and whether their Maven plugins have equivalent settings.  If not, then I would argue that the current name is more accurate.  Do you know whether Scala and Kotlin users would expect `source` and `target` settings to exist, and what they would mean?  I'm pretty sure that for Scala, using a different source version would mean switching to a different version of the compiler.  I'm also not sure what the `target` setting would mean in that context.

geoand: I don't know anything about Scala. However with Kotlin you can definitely set the `target` version, see https://kotlinlang.org/docs/reference/using-maven.html#attributes-specific-for-jvm

core/devmode/src/test/java/io/quarkus/dev/CompilerFlagsTest.java
benkard: The way it is implemented right now, it would keep both.

I think there are two reasonable choices:

1. prefer the explicit compiler flags and ignore `source`/`target` if they conflict (rationale: if you are setting compiler flags explicitly, you probably expect them to be applied; the `source` and `target` settings might just have been taken from their defaults and you may not even be aware of them)
 2. pass both

I wonder what the Maven compiler plugin does here.  It would probably be best to try and be consistent with it.  My guess is (2), but I'll see if I can find out more so I can say with more certainty.

We should definitely add a unit test that at least documents the behavior in either case, though.

benkard: OK, looking at `org.codehaus.plexus.compiler.javac.JavacCompiler` (`plexus-compiler-javac-2.8.5-sources.jar`) it does look like it simply keeps both.  It adds any custom flags _after_ the `-source` and `-target` settings derived from the configuration, so we should probably make sure that we do that, too.

geoand: Makes sense

devtools/maven/src/main/java/io/quarkus/maven/DevMojo.java
benkard: OK, renamed `targetJavaVersion` to `targetJvmVersion`.

core/devmode/src/test/java/io/quarkus/dev/CompilerFlagsTest.java
benkard: Done.

devtools/maven/src/main/java/io/quarkus/maven/DevMojo.java
benkard: As for the source version, do we really want a single variable that covers both languages?  That might cause trouble when you're mixing several languages in the same project.  It might make more sense to rename the `<source/>` element to `<javaSource>` in the plugin configuration instead.  What do you think?

geoand: Currently mixing several languages will (most likely) not work in Dev mode because there is way to setup the proper order of compilation.
We should probably just keep source for now I think.

geoand:
@benkard Thanks a lot! I think we are good now so could you please squash the commits and force push?

benkard:
Squashed and rebased.

geoand (approved):
Thanks for this!

Let's wait for CI and merge

geoand:
@benkard would be interested in doing the same for Kotlin and/or Scala?

PR #4402, opened 2019-10-07
Remove useless capability strings

geoand:
Found by @gsmet 

With #4212 in, these capability strings are now longer used anywhere

### DIFF
--- a/extensions/resteasy-common/deployment/src/main/java/io/quarkus/resteasy/common/deployment/ResteasyCommonProcessor.java
+++ b/extensions/resteasy-common/deployment/src/main/java/io/quarkus/resteasy/common/deployment/ResteasyCommonProcessor.java
@@ -191,7 +191,7 @@ public class ResteasyCommonProcessor {
     private boolean restJsonSupportNeeded(CombinedIndexBuildItem indexBuildItem, DotName mediaTypeAnnotation) {
         for (AnnotationInstance annotationInstance : indexBuildItem.getIndex().getAnnotations(mediaTypeAnnotation)) {
             final AnnotationValue annotationValue = annotationInstance.value();
-            if (annotationInstance == null) {
+            if (annotationValue == null) {
                 continue;
             }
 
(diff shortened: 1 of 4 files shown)


gsmet:
And we need the first one?

If so, I would clean up:
```
public static final String RESTEASY_JSON_EXTENSION = "io.quarkus.resteasy-json-extension";
```
and make it `io.quarkus.resteasy-json`. Right now, it's not consistent with the others.

gsmet:
And we have the exact same pattern in `ResteasyJacksonProcessor`.

I suppose the first one is saying that we have some JSON capability, be it Jackson or JSON-B.

geoand:
Yes the first one is being used in `ResteasyCommonProcess`. 
Let me dig in a little more to see why it's actually being used. Maybe it can be removed completely

geoand:
The capability is used to provide a good warning message so it makes sense to keep it. I'll just update the names.

geoand:
@gsmet PR updated

geoand:
> I suppose the first one is saying that we have some JSON capability, be it Jackson or JSON-B.

Yes, that' exactly what is going on

extensions/resteasy-common/deployment/src/main/java/io/quarkus/resteasy/common/deployment/ResteasyCommonProcessor.java
```diff
@@ -191,7 +191,7 @@ JaxrsProvidersToRegisterBuildItem setupProviders(BuildProducer<ReflectiveClassBu
     private boolean restJsonSupportNeeded(CombinedIndexBuildItem indexBuildItem, DotName mediaTypeAnnotation) {
         for (AnnotationInstance annotationInstance : indexBuildItem.getIndex().getAnnotations(mediaTypeAnnotation)) {
             final AnnotationValue annotationValue = annotationInstance.value();
-            if (annotationInstance == null) {
+            if (annotationValue == null) {
```
gsmet: Good catch.

gsmet (approved):
I think it's good but not having CI running makes me a bit nervous.

geoand:
I think it's safe as well, but we have all seen stranger things happen :P

extensions/resteasy-common/deployment/src/main/java/io/quarkus/resteasy/common/deployment/ResteasyCommonProcessor.java
geoand: Gotta hand it to IntelliJ for pointing it out :)

machi1990: 👍

geoand:
I think perhaps the CI events are being delivered very very slowly

Issue #4393, opened 2019-10-05
NPE when using nested JUnit5 classes


acoburn:
When using the `quarkus-junit5` module (Quarkus version 0.23.2) with the `@QuarkusTest` annotation, any `@Nested` JUnit5 classes consistently lead to null pointer exceptions.

The relevant portion of the stack trace is:

```
org.junit.jupiter.api.extension.TestInstantiationException: TestInstanceFactory [io.quarkus.test.junit.QuarkusTestExtension] failed to instantiate test class [org.acme.quickstart.GreetingResourceTest$NestedClass]: Bean is null
	at org.junit.jupiter.engine.descriptor.ClassBasedTestDescriptor.invokeTestInstanceFactory(ClassBasedTestDescriptor.java:300)
	at org.junit.jupiter.engine.descriptor.ClassBasedTestDescriptor.instantiateTestClass(ClassBasedTestDescriptor.java:275)
	[ ... ]
Caused by: java.lang.NullPointerException: Bean is null
	at java.base/java.util.Objects.requireNonNull(Objects.java:246)
	at io.quarkus.arc.BeanManagerImpl.getReference(BeanManagerImpl.java:54)
	at io.quarkus.test.common.TestInstantiator.instantiateTest(TestInstantiator.java:25)
	at io.quarkus.test.junit.QuarkusTestExtension.createTestInstance(QuarkusTestExtension.java:368)
	at org.junit.jupiter.engine.descriptor.ClassBasedTestDescriptor.invokeTestInstanceFactory(ClassBasedTestDescriptor.java:285)
	... 82 more
```

I can reproduce this consistently in JDK8 and JDK11, with both Gradle and Maven build environments.

A minimal example of this error can be found at https://github.com/acoburn/quarkus-junit5-test, which is built from the `getting-started` archetype. The only [change](https://github.com/acoburn/quarkus-junit5-test/commit/5e6caeb240314c18fe5661fd1657ecc4e4fdc732) to the generated code is that the JUnit `@Test` is [wrapped](https://github.com/acoburn/quarkus-junit5-test/blob/master/src/test/java/org/acme/quickstart/GreetingResourceTest.java#L13-L23) in a `@Nested` class.

geoand:
[...]

PR #4409, opened 2019-10-07
Junit5 - inner classes are not allowed as they cannot become managed …

manovotn:
…beans.

I tried to write an automated test too, but I don't think we have a way to assert failure during junit extension processing, right @geoand ?


manovotn:
Fixes #4393

geoand:
@manovotn not that I know of anyway :)

geoand (approved):

geoand:
Thanks for taking care of this @manovotn !

PR #4413, opened 2019-10-07
Security annotations for cdi

### LINKED ISSUE #4334: CDI security interceptor for @RolesAllowed, etc
> There are a few things to bear in mind here, one is that those annotations are not actually interceptor bindings, so you will need to use an ArC build time extension to do the bindings, the other is to figure out how it works with RESTEasy. At the moment there is a RESTeasy Filter thing that handles this, but it would be better to just have the interceptor handle everything and the filter just catch the exception and handle it appropriatle

The source of roles is `io.quarkus.security.identity.SecurityIdentity`

### LINKED ISSUE #4344: Handle non-binding fields for interceptor bindings from non-binding annotations
`InterceptorBindingRegistrar` makes it possible to make annotations binding even if they do not have `@InterceptorBinding`.

I'd like to use it `@RolesAllowed`. The problem is this annotation requires a `value`. 
An interceptor for it should be triggered regardless of the value. I'd like to have an option to specify that the `value` field is non-binding in this situation.

I have briefly discussed it with @mkouba and one option to achieve that would be to extend `InterceptorBindingRegistrar` to provide such information.

I'd like to implement it, creating the issue to provide a space to discuss the problem.

cc @manovotn

michalszynkiewicz:
fixes #4334 
fixes #4344


extensions/resteasy/runtime/src/main/java/io/quarkus/resteasy/runtime/RequestFailer.java
```diff
@@ -1,78 +0,0 @@
...
-        } catch (Exception e) {
-
-        }
-        AUTHENTICATE = auth;
-        HTTP_SERVLET_REQUEST = httpServletReq;
-        HTTP_SERVLET_RESPONSE = httpServletResp;
-    }
-
-    public static void fail(ContainerRequestContext requestContext) {
-        if (requestContext.getSecurityContext().getUserPrincipal() == null) {
```
michalszynkiewicz: @stuartwdouglas what was the case in which this was needed? I moved the logic to https://github.com/quarkusio/quarkus/pull/4413/files#diff-cd55a2a1e7a26ebe9e639f503a059ef1R26 and it is working in the tests without it but I'd like to be sure I'm not missing some scenario in which this is still needed.

extensions/security/runtime/src/main/java/io/quarkus/security/runtime/interceptor/SecurityHandler.java
```diff
@@ -0,0 +1,34 @@
...
+public class SecurityHandler {
+
+    private static final String HANDLER_NAME = SecurityHandler.class.getName();
+    private static final String EXECUTED = "executed";
+
+    @Inject
+    SecurityConstrainer constrainer;
+
+    public Object handle(InvocationContext ic) throws Exception {
+        if (alreadyHandled(ic)) {
```
geoand: This might be a stupid question, but why is this check needed?

extensions/security/runtime/src/main/java/io/quarkus/security/runtime/interceptor/SecurityConstrainer.java
```diff
@@ -0,0 +1,164 @@
...
+        if (securityAnnotation == null) {
+            Class<?> declaringClass = method.getDeclaringClass();
+            securityAnnotation = determineSecurityAnnotation(declaringClass.getDeclaredAnnotations(),
+                    declaringClass::getCanonicalName);
+        }
+        if (securityAnnotation == null) {
+            securityAnnotation = determineSecurityAnnotationFromBindings(interceptorBindings, method::toString);
+        }
+        return checkForAnnotation(securityAnnotation);
+    }
```
geoand: So far the codebase has tried to not use lambdas in runtime code unless there is a very good reason to (for performance reasons). Would it be possible to change these instances to non-lambda code?

extensions/security/runtime/src/main/java/io/quarkus/security/runtime/SecurityBuildTimeConfig.java
```diff
@@ -0,0 +1,39 @@
...
+     * </pre>
+     */
+    @ConfigItem(name = "deny.unannotated", defaultValue = "false")
+    public boolean denyUnannotated;
+
+    /**
+     * if set to true, access to all JAX-RS resources will be denied by default
+     */
+    @ConfigItem(name = "deny.jaxrs", defaultValue = "false")
+    public boolean denyJaxRs;
```
geoand: I don't really see the usefulness of these TBH. Is this something that Thorntail had and proved useful for users?

stuartwdouglas: It means that you cannot accidentally expose something because an annotation was forgotten. IMHO this is a useful feature as the 'public by default' approach can be a security risk.

extensions/security/runtime/src/main/java/io/quarkus/security/runtime/interceptor/SecurityHandler.java
michalszynkiewicz: it's only for preventing the check to be triggered multiple times, e.g. if one annotation is on the class, another on the method

extensions/security/runtime/src/main/java/io/quarkus/security/runtime/interceptor/SecurityConstrainer.java
```diff
@@ -0,0 +1,164 @@
...
+    public void checkRoles(Method method, Collection<Annotation> interceptorBindings) {
+        getCheck(method, interceptorBindings)
+                .ifPresent(check -> check.apply(identity));
+    }
+
+    private Optional<Check> getCheck(Method method, Collection<Annotation> interceptorBindings) {
+        return checkForMethod.computeIfAbsent(method, m -> determineSecurityCheck(m, interceptorBindings));
+    }
+
+    private Optional<Check> determineSecurityCheck(Method method, Collection<Annotation> interceptorBindings) {
```
geoand: Is there any thought to leveraging build time information for the types of checks that are being performed here?

michalszynkiewicz: that would definitely be good. 
OTOH, determining the check is done only once for method.

extensions/security/runtime/src/main/java/io/quarkus/security/runtime/interceptor/SecurityConstrainer.java
michalszynkiewicz: These are method references, are they forbidden too?
They can be translated without problem to classes but we'd lose on readability.
And it's executed only on failure, and only once for a method so not sure if it's worth it.

geoand: `determineSecurityAnnotationFromBindings` however is full of lambdas :)

extensions/security/runtime/src/main/java/io/quarkus/security/runtime/interceptor/SecurityConstrainer.java
stuartwdouglas: +1 to doing all this at build time, but I am not sure if our current API's really support it. @mkouba any ideas about this? We really want method level interceptors to be able to override class level interceptors, and also some way of passing in information to the interceptors (i.e. the roles allowed). At the moment this is not really possible with CDI but it would be great if we could support it somehow.

extensions/security/runtime/src/main/java/io/quarkus/security/runtime/interceptor/SecurityConstrainer.java
stuartwdouglas: They use a lot more memory at runtime, and are slower to load which impacts startup time. It is best to avoid them.

extensions/security/runtime/src/main/java/io/quarkus/security/runtime/interceptor/SecurityConstrainer.java
```diff
@@ -0,0 +1,164 @@
...
+    private static class RolesAllowedCheck implements Check {
+        private final String[] allowedRoles;
+
+        private RolesAllowedCheck(String[] allowedRoles) {
+            this.allowedRoles = allowedRoles;
+        }
+
+        @Override
+        public void apply(SecurityIdentity identity) {
+            if (identity.isAnonymous()) {
```
stuartwdouglas: I would skip this check. Potentially you can apply roles to unauthenticated identities (e.g. if a request comes from inside the company network you might want to give it slightly higher permissions). By default the role set will be empty anyway.

extensions/smallrye-openapi/runtime/pom.xml
```diff
@@ -40,6 +40,10 @@
       <groupId>io.quarkus</groupId>
       <artifactId>quarkus-swagger-ui</artifactId>
     </dependency>
+    <dependency>
+      <groupId>io.quarkus</groupId>
+      <artifactId>quarkus-resteasy</artifactId>
```
stuartwdouglas: Why was adding this dep nessesary?

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/security/BasicAuthenticationMechanism.java
```diff
@@ -172,8 +172,18 @@ private static void clear(final char[] array) {
                 return CompletableFuture.completedFuture(false);
             }
         }
-        context.response().headers().set(HttpHeaderNames.WWW_AUTHENTICATE, challenge);
-        context.response().setStatusCode(HttpResponseStatus.UNAUTHORIZED.code());
+        context.response().headers().set(challengeHeader(), challengeContent());
```
stuartwdouglas: I think all these changes to the mechanisms will need to be changed to support multiple authentication mechanisms. 

Also the API needs to be changed to allow returning an AuthResult with all the info, as this is much more flexible

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/security/HttpAuthenticationMechanism.java
```diff
@@ -15,4 +16,11 @@
 
     CompletionStage<Boolean> sendChallenge(RoutingContext context);
```
stuartwdouglas: Rather than adding these extra methods there should be a ComplectionStage<ChallengeResult> getChallenge(RoutingContext context). The result should contain this extra information. sendChallenge should have a default implementation added that delegates to the results of this method.

independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/BeanDeployment.java
```diff
@@ -57,6 +57,8 @@
 
     private final Map<DotName, ClassInfo> interceptorBindings;
 
+    private final Map<DotName, Set<String>> nonBindingFields;
```
stuartwdouglas: @mkouba can you look over these ArC changes?

integration-tests/cdi-security-deny-all-jaxrs/pom.xml
```diff
@@ -0,0 +1,117 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
```
stuartwdouglas: Does this really need its own integration test modules? Most of this testing should be done as unit testing using QuarkusUnitTest, and native image testing should be possible in one of the existing security modules that already use this functionality.

It is definitely not appropriate to have one integration test module per interceptor, that is way too fine grained an approach for integration tests.

test-framework/maven/src/main/java/io/quarkus/maven/it/MojoTestBase.java
```diff
@@ -293,7 +293,7 @@ public static void assertThatOutputWorksCorrectly(String logs) {
         assertThat(logs.contains(infoLogLevel)).isTrue();
         Predicate<String> datePattern = Pattern.compile("\\d{4}-\\d{2}-\\d{2}\\s\\d{2}:\\d{2}:\\d{2},\\d{3}\\s").asPredicate();
         assertThat(datePattern.test(logs)).isTrue();
-        assertThat(logs.contains("features: [cdi, resteasy, servlet, undertow-websockets]")).isTrue();
+        assertThat(logs.contains("features: [cdi, resteasy, security, servlet, undertow-websockets]")).isTrue();
```
stuartwdouglas: I don't really like that JAX-RS now includes security by default, we should look at fixing this, usually we work around this by adding some kind of SPI module.

stuartwdouglas (changes requested):

integration-tests/cdi-security-deny-all-jaxrs/pom.xml
michalszynkiewicz: Config switches for denying all JAX-RS or denying unnanotated, they have to be different for the tests. Is there a way to have one test module that would test different configuration combinations?
Or do you mean that the switches should not be tested in integration tests?

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/security/HttpAuthenticationMechanism.java
michalszynkiewicz: and then translate the exception in a vertx filter/handler instead of JAX-RS exception mapper?

integration-tests/cdi-security-deny-all-jaxrs/pom.xml
stuartwdouglas: Have a look at QuarkusUnitTest. These tests are in the deployment modules, and each test can start Quarkus with a different config. There are some security related ones in the vertx-http module.

extensions/smallrye-openapi/runtime/pom.xml
michalszynkiewicz: there's a dependency between deployments already, without it the TCK failed with missing `SecurityIdentityAssociation` class. That's probably because of the dependency between RestEasy and security extensions. I check if it's still needed after fixing all the issues in the PR

extensions/security/runtime/src/main/java/io/quarkus/security/runtime/interceptor/SecurityConstrainer.java
michalszynkiewicz: don't we have to differentiate between 401 and 403?
Or should 401 be returned for a logged in user without roles too?

test-framework/maven/src/main/java/io/quarkus/maven/it/MojoTestBase.java
michalszynkiewicz: could you write a bit more on that, or maybe give me an example?

extensions/security/runtime/src/main/java/io/quarkus/security/runtime/interceptor/SecurityConstrainer.java
stuartwdouglas: Maybe do the check at the point you throw the exception.

michalszynkiewicz:
so far having trouble reproducing the test failure locally, this is what happens there: https://gist.github.com/michalszynkiewicz/b69ab0f4b42dc08a4ecd733e80985bf9

test-framework/maven/src/main/java/io/quarkus/maven/it/MojoTestBase.java
stuartwdouglas: Let's look at it when the other issues have been addressed, I am about to board a flight back to Australia so will be out of contact for a while.

extensions/resteasy/runtime/src/main/java/io/quarkus/resteasy/runtime/ForbiddenExceptionMapper.java
```diff
@@ -0,0 +1,23 @@
...
+ * @author Michal Szynkiewicz, michal.l.szynkiewicz@gmail.com
+ *         <br>
+ *         Date: 05/10/2019
+ */
+@Provider
+@Priority(Priorities.USER + 1)
+public class ForbiddenExceptionMapper implements ExceptionMapper<ForbiddenException> {
+    @Override
+    public Response toResponse(ForbiddenException exception) {
+        return Response.status(403).entity("Access forbidden: role not allowed").build();
```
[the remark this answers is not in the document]
michalszynkiewicz: done

extensions/security/deployment/src/main/java/io/quarkus/security/deployment/SecurityTransformerUtils.java
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * @author Michal Szynkiewicz, michal.l.szynkiewicz@gmail.com
+ *         <br>
+ *         Date: 11/10/2019
+ */
+public class SecurityTransformerUtils {
+    public static final DotName DENY_ALL = DotName.createSimple(DenyAll.class.getName());
+
+    public static boolean hasSecurityAnnotation(MethodInfo methodInfo) {
```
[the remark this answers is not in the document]
michalszynkiewicz: done

extensions/security/runtime/src/main/java/io/quarkus/security/runtime/ForbiddenException.java
```diff
@@ -0,0 +1,9 @@
+package io.quarkus.security.runtime;
+
+/**
+ * @author Michal Szynkiewicz, michal.l.szynkiewicz@gmail.com
+ *         <br>
+ *         Date: 07/10/2019
```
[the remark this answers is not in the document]
michalszynkiewicz: done

extensions/security/runtime/src/main/java/io/quarkus/security/runtime/interceptor/SecurityConstrainer.java
```diff
@@ -0,0 +1,164 @@
...
+ *         <br>
+ *         Date: 03/10/2019
+ */
+@Singleton
+public class SecurityConstrainer {
+
+    private static final List<Class<? extends Annotation>> SECURITY_ANNOTATIONS = asList(Authenticated.class, DenyAll.class,
+            PermitAll.class, RolesAllowed.class);
+
+    private final Map<Method, Optional<Check>> checkForMethod = new HashMap<>();
```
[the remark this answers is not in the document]
michalszynkiewicz: done

extensions/security/runtime/src/main/java/io/quarkus/security/runtime/interceptor/SecurityConstrainer.java
michalszynkiewicz: true about `determineSecurityAnnotationFromBindings` :) I will de-lambdify it

extensions/security/runtime/src/main/java/io/quarkus/security/runtime/interceptor/SecurityConstrainer.java
michalszynkiewicz: hmm... the exception is thrown here. Do you mean the point of translating the error to response?

extensions/security/deployment/src/main/java/io/quarkus/security/deployment/SecurityTransformerUtils.java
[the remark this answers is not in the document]
michalszynkiewicz: done

extensions/security/runtime/src/main/java/io/quarkus/security/runtime/interceptor/SecurityConstrainer.java
```diff
@@ -0,0 +1,164 @@
...
+
+    @Inject
+    SecurityIdentity identity;
+
+    public void checkRoles(Method method, Collection<Annotation> interceptorBindings) {
+        getCheck(method, interceptorBindings)
+                .ifPresent(check -> check.apply(identity));
```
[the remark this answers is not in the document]
michalszynkiewicz: The next commit will remove the lambdas from here but I'm wondering about the `::toString` and `::getCanonicalName` method references for printing out the location of the error. Is it okay to keep them?

extensions/security/runtime/src/main/java/io/quarkus/security/runtime/interceptor/SecurityConstrainer.java
```diff
@@ -0,0 +1,164 @@
...
+        if (securityAnnotation instanceof PermitAll) {
+            return Optional.of(new PermitAllCheck());
+        }
+        if (securityAnnotation instanceof Authenticated) {
+            return Optional.of(new AuthenticatedCheck());
+        }
+        return Optional.empty();
+    }
+
+    private Annotation determineSecurityAnnotationFromBindings(Collection<Annotation> interceptorBindings,
```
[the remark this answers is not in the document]
michalszynkiewicz: done

extensions/security/runtime/src/main/java/io/quarkus/security/runtime/interceptor/SecurityConstrainer.java
```diff
@@ -0,0 +1,164 @@
...
+            Supplier<String> annotationPlacement) {
+        List<Annotation> securityAnnotations = interceptorBindings.stream()
+                .filter(anno -> SECURITY_ANNOTATIONS.stream().anyMatch(c -> c.isInstance(anno)))
+                .collect(Collectors.toList());
+        return getExactlyOne(securityAnnotations, annotationPlacement);
+    }
+
+    private Annotation determineSecurityAnnotation(Annotation[] annotations, Supplier<String> annotationPlacement) {
```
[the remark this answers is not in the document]
michalszynkiewicz: done

extensions/resteasy/deployment/src/main/java/io/quarkus/resteasy/deployment/DenyJaxRsTransformer.java
```diff
@@ -0,0 +1,56 @@
...
+    @Override
+    public void transform(TransformationContext transformationContext) {
+        ClassInfo classInfo = transformationContext.getTarget().asClass();
+        if (!hasSecurityAnnotation(classInfo) && isJaxRsResource(classInfo)) {
+            transformationContext.transform().add(DENY_ALL).done();
+        }
+    }
+
+    private boolean isJaxRsResource(ClassInfo classInfo) {
+        return !Modifier.isInterface(classInfo.flags())
```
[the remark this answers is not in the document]
michalszynkiewicz: works and simplifies things a lot :)

extensions/security/runtime/src/main/java/io/quarkus/security/runtime/interceptor/PermitAllInterceptor.java
```diff
@@ -0,0 +1,27 @@
...
+import javax.interceptor.InvocationContext;
+
+/**
+ * @author Michal Szynkiewicz, michal.l.szynkiewicz@gmail.com
+ *         <br>
+ *         Date: 05/10/2019
+ */
+@Interceptor
+@PermitAll
+@Priority(1)
```
manovotn: Instead of magic numbers, you should probably use a combination of a predefined constant plus/minus a number (applies to all interceptors added here).
See the values here - https://javaee.github.io/javaee-spec/javadocs/javax/interceptor/Interceptor.Priority.html
I am not sure what would be appropriate for security interceptors though; I guess you want them to trigger very soon.

extensions/security/runtime/src/main/java/io/quarkus/security/runtime/interceptor/SecurityConstrainer.java
manovotn: Normally you solve this simply by nature of interceptor resolution - by bindings. But the combinations here don't really allow that - there are 4 interceptors and they all share the same logic which only then decides what should be done.

> some way of passing in information to the interceptors (i.e. the roles allowed)

Not sure I follow? You can store anything inside `InvocationContext` and pass it around between the interceptors if that's what you mean.

As for method and class bindings, we could perhaps store them separately in arc and expose them as we now do the combined version? So at least user can see what was defined where. But otherwise I think interceptors spec already defined that method ones have a priority over class ones, so I suppose if you place the same annotation on a class and a method, then the method one will be taken into account. I would have to poke around the spec to verify that.

independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/BeanDeployment.java
manovotn: I already did (in Michal's first version of the code and this bit doesn't seem to have changed). In general this is correct and will work. The question is whether we somehow want to weave in annotations transformer directly into the registrar so that we avoid hardcoding non-binding fields and allow others to view that change via annotation store. But I am not quite sure how what would work, WDYT @mkouba ?

extensions/security/runtime/src/main/java/io/quarkus/security/runtime/interceptor/PermitAllInterceptor.java
michalszynkiewicz: `LIBRARY_BEFORE` seems most appropriate, wdyt?

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/security/HttpAuthenticationMechanism.java
michalszynkiewicz: tried that, unfortunately doesn't play well with the undertow exception mapping.
Willadd the `getChallenge` and stay on resteasy exception mapper and read the challenge in a blocking manner for now.

extensions/security/runtime/src/main/java/io/quarkus/security/runtime/interceptor/SecurityConstrainer.java
mkouba: > At the moment this is not really possible with CDI but it would be great if we could support it somehow.

@stuartwdouglas @michalszynkiewicz an interceptor can inject any bean so if we could collect/serialize the info (roles allowed) during build then we could inject/use it at runtime.

manovotn: @mkouba you mean collect them and expose as a bean? That sounds interesting.

ci-templates/native-build-steps.yaml
```diff
@@ -44,4 +44,4 @@ jobs:
       inputs:
         goals: 'install'
         mavenOptions: $(MAVEN_OPTS)
-        options: '-pl integration-tests/${{ join('',integration-tests/'', parameters.modules) }} -B --settings azure-mvn-settings.xml -Dnative-image.docker-build -Dtest-postgresql -Dtest-elasticsearch -Dtest-keycloak -Ddocker-keycloak -Dtest-dynamodb -Dtest-mysql -Dnative-image.xmx=6g -Dnative -Dno-format'
+        options: '-pl integration-tests/${{ join('',integration-tests/'', parameters.modules) }} -B --settings azure-mvn-settings.xml -Dnative-image.docker-build -Dtest-postgresql -Dtest-elasticsearch -Dtest-keycloak -Pdocker-keycloak -Dtest-dynamodb -Dtest-mysql -Dnative-image.xmx=6g -Dnative -Dno-format'
```
michalszynkiewicz: the oidc tests use `docker-keycloak` as the profile name, the property to control it is `docker`. It seems to me that `-Ddocker-keycloak` was a typo here but maybe there's more magic here...

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/security/HttpAuthenticationMechanism.java
michalszynkiewicz: I added the first part of the rework and tried to address it. Is this what you meant @stuartwdouglas ?

integration-tests/cdi-security-deny-all-jaxrs/pom.xml
michalszynkiewicz: Anyway, will do it tomorrow

extensions/resteasy/runtime/src/main/java/io/quarkus/resteasy/runtime/UnauthorizedExceptionMapper.java
```diff
@@ -0,0 +1,73 @@
...
+                    AUTHENTICATE.invoke(httpServletRequest, httpServletResponse);
+                } catch (Exception e) {
+                    throw new RuntimeException(e);
+                }
+            }
+        }
+        HttpAuthenticator authenticator = identity.getAttribute(HttpAuthenticator.class.getName());
+        RoutingContext context = ResteasyContext.getContextData(RoutingContext.class);
+        if (authenticator != null && context != null) {
+            return Response.status(authenticator.challengeStatus())
```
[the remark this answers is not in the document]
michalszynkiewicz: I think the changed one should be easier to adjust to multiple auth mechanisms, wdyt?

integration-tests/cdi-security-deny-all-jaxrs/pom.xml
stuartwdouglas: yea, so you should add a smoke test into one of the existing security integration tests. We can't add a new module for each discrete feature because otherwise our CI will take hours to run.

extensions/elytron-security-properties-file/deployment/src/test/java/io/quarkus/security/test/CustomAuth.java
```diff
@@ -76,8 +76,18 @@
 
     @Override
     public CompletionStage<Boolean> sendChallenge(RoutingContext context) {
-        context.response().headers().set(HttpHeaderNames.WWW_AUTHENTICATE, "BASIC realm=CUSTOM");
```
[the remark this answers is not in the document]
michalszynkiewicz: sendChallenge is not overriden here anymore

extensions/security/runtime/src/main/java/io/quarkus/security/runtime/SecurityBuildTimeConfig.java
michalszynkiewicz: moved

extensions/security/runtime/src/main/java/io/quarkus/security/runtime/interceptor/SecurityConstrainer.java
michalszynkiewicz: I chatted with @mkouba and @manovotn about it: while it's doable to prepare the data for the check in the build time, it's not that straighforward and the perf impact of doing it as it is is not big.
For this PR I'd rather keep it in the runtime (it's already quite big now) and, if needed, create an issue to move it to build time later on/in a separate PR

extensions/oidc/runtime/src/main/java/io/quarkus/oidc/VertxOAuth2AuthenticationMechanism.java
```diff
@@ -80,11 +80,26 @@ public VertxOAuth2AuthenticationMechanism setAuth(OAuth2Auth auth) {
 
     @Override
     public CompletionStage<Boolean> sendChallenge(RoutingContext context) {
-        context.response().setStatusCode(302);
-        context.response().headers().set(HttpHeaders.LOCATION, authURI(authServerURI));
+        context.response().setStatusCode(challengeStatus());
```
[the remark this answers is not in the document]
michalszynkiewicz: the code change a lot there, I'm marking the comment as resolved

integration-tests/cdi-security-deny-all-jaxrs/pom.xml
michalszynkiewicz: done. One test added to the main integration tests, the rest is moved to deployment modules.

test-framework/maven/src/main/java/io/quarkus/maven/it/MojoTestBase.java
michalszynkiewicz: Just to add some background: I added it to the resteasy extension because before the changes resteasy had RBAC for JAX-RS implemented.

docs/src/main/asciidoc/security-guide.adoc
```diff
@@ -64,6 +64,15 @@ public class SubjectExposingResource {
 <4> This call to obtain the user principal will return null if the caller is unauthenticated, non-null if the caller is authenticated.
 <5> The `/subject/denied` endpoint disallows any access regardless of whether the call is authenticated by specifying the `@DenyAll` annotation.
 
+There are two configuration settings that alter the RBAC behavior:
+
+- `quarkus.security.deny.jaxrs=true|false` - if set to true, the access will be denied for all JAX-RS endpoints will be
```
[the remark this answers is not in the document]
michalszynkiewicz: The original text here was only about JAX-RS and the code example is about JAX-RS. Do you think the example should be changed to a pure CDI bean too? Or are you referring to the `deny.jaxrs` switch only?

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/security/BasicAuthenticationMechanism.java
michalszynkiewicz: I think, with the current changes, it should be easier to support multiple authentication mechanisms, but I'm not entirely sure how would that work so I'm not certain that's what you meant.
By AuthResult did you mean something like the added `ChallengeData`?

extensions/security/runtime/src/main/java/io/quarkus/security/runtime/interceptor/SecurityConstrainer.java
stuartwdouglas: I mean change the code so it first checks the roles, as it is possible to have an anonymous identity that still has some roles (e.g. an anonymous identity from inside the company network might have more permissions than an external one). If the role check fails then decide which exception to throw.

michalszynkiewicz: got it now, thanks.

michalszynkiewicz: will be fixed in the next commit

extensions/security/runtime/src/main/java/io/quarkus/security/runtime/interceptor/SecurityConstrainer.java
michalszynkiewicz: I created https://github.com/quarkusio/quarkus/issues/4695 to dicuss it further/implement it later

extensions/smallrye-openapi/runtime/pom.xml
michalszynkiewicz: still needed at this point.

docs/src/main/asciidoc/security-guide.adoc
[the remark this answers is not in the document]
michalszynkiewicz: Turns out we don't have a Rest guid at the moment. There is one for REST+JSON, one for MP JWT that also touches REST security. 
Did you mean using one of these or creating a separate REST Security guide?
I'd be happy to do it (when I understand what :) ) but if you find it acceptable I'd rather do this change in a separate PR.

independent-projects/bootstrap/pom.xml
```diff
@@ -85,8 +86,18 @@
                         <groupId>org.sonatype.plexus</groupId>
                         <artifactId>plexus-cipher</artifactId>
                     </exclusion>
+                    <exclusion>
```
michalszynkiewicz: needed because jsr250 brought in old security annotations (e.g. `@DenyAll` that couldn't be placed on classes)

michalszynkiewicz:
I squashed the commits not to pollute the changelog. Hope it doesn't make reviewing more difficult

docs/src/main/asciidoc/security-guide.adoc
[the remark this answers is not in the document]
stuartwdouglas: It's probably fine to keep it here, but maybe the property should be something like quarkus.jaxrs.deny-uncovered.

extensions/resteasy/deployment/pom.xml
```diff
@@ -38,6 +38,10 @@
             <groupId>com.oracle.substratevm</groupId>
             <artifactId>svm</artifactId>
         </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-security-deployment</artifactId>
```
stuartwdouglas: RESTeasy should not have a hard dependency on the security extension. I think the exceptions should probably be moved to the quarkus-security project to avoid the runtime dependency, and just copy the trivial utility method used in the -deployment module.

extensions/smallrye-openapi/runtime/pom.xml
stuartwdouglas: Can you re-check after making the dependency changes I mention above?

stuartwdouglas (changes requested):
Once the dependency issues are addressed we should be good to go.

docs/src/main/asciidoc/security-guide.adoc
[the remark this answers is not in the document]
michalszynkiewicz: ok, I will switch deny.unannotated to deny-unannotated to be consistent

extensions/resteasy/deployment/pom.xml
michalszynkiewicz: done, now it's only a test dependency

michalszynkiewicz: done

michalszynkiewicz:
Requires a new release of Quarkus Security that would include https://github.com/quarkusio/quarkus-security/pull/2

stuartwdouglas (approved):

PR #4417, opened 2019-10-07
Make working directory configurable for Gradle's and Maven's dev tasks

### LINKED ISSUE #4229: Working Directory for mvnw compile quarkus:dev
**Description**
The current system changes the working directory when run using "mvnw compile quarkus:dev" to the target/classes directory.  This basically comes down to an unexpected behaviour. 

A basic tenant I normally work to is "when I run something from a directory I expect it to be running in that directory rather than changing to another directory."

As mentioned in the bug report that I believe this will be confusing for anyone new to the Quarkus framework as it is an unexpected behaviour.

Screen Shot 2019-09-27 at 08 30 12

[QuarkusExample.zip](https://github.com/quarkusio/quarkus/files/3659773/QuarkusExample.zip)

The above code has been modified (from that provided in the bug report) slightly to better explain the issue.

See related Issue https://github.com/quarkusio/quarkus/issues/4205.

I believe this will save people new to the framework a significant time, effort.  I am concerned that it will put people off this fantastic framework.  I is very important make sure that the framework behaviour is what you would expect.  

**Implementation ideas**
This is the example of what I mean.

The internal application.properties has the quarkus.http.port set to 9090 and the external my_application.properties file has the quarkus.http.port set to 9292.

When I start up the application either through
mvnw compile quarkus:dev

or
java -jar target/quarkus-example-runner.jar

I am expecting it to listening on 9292.

Actual behavior
When started with the mvnw compile quarkus:dev it listens on 9090 (and can't find the my_application.properties) and java -jar target/quarkus-example-runner.jar listens on 9292.

As pointed out in the bug to me, I could get it to work by putting an absolute path to the external my_application.properties and it would work.  But that fact that it has changed directory is confusing.

paulolieuthier:
Please, show me what I have missed, first time around! :grimacing: 

Fixes #4229.

### DIFF
--- a/devtools/gradle/src/main/java/io/quarkus/gradle/QuarkusPluginExtension.java
+++ b/devtools/gradle/src/main/java/io/quarkus/gradle/QuarkusPluginExtension.java
@@ -30,6 +30,8 @@ public class QuarkusPluginExtension {
 
     private String sourceDir;
 
+    private String workingDir;
+
     private String outputConfigDirectory;
 
     public QuarkusPluginExtension(Project project) {
@@ -78,6 +80,18 @@ public class QuarkusPluginExtension {
 
     public void setSourceDir(String sourceDir) {
         this.sourceDir = sourceDir;
+    }
+
+    public File workingDir() {
+        if (workingDir == null) {
+            workingDir = outputDirectory().getPath();
+        }
+
+        return new File(workingDir);
+    }
+
+    public void setWorkingDir(String workingDir) {
+        this.workingDir = workingDir;
     }
 
     public File wiringClassesDirectory() {
--- a/devtools/gradle/src/main/java/io/quarkus/gradle/tasks/QuarkusDev.java
+++ b/devtools/gradle/src/main/java/io/quarkus/gradle/tasks/QuarkusDev.java
@@ -70,6 +70,8 @@ public class QuarkusDev extends QuarkusTask {
 
     private String sourceDir;
 
+    private String workingDir;
+
     private String jvmArgs;
 
     private boolean preventnoverify = false;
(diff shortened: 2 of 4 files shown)


paulolieuthier:
Just tested it with gradle:
```
quarkusDev {
    workingDir = rootProject.projectDir
}
```

aloubyansky (approved):

paulolieuthier:
@aloubyansky, just a note: I didn't test the task option in a maven project. It would also be a good idea to add documentation somewhere. Could you point it me the direction if so?

aloubyansky:
The docs would be
https://github.com/quarkusio/quarkus/blob/master/docs/src/main/asciidoc/maven-tooling.adoc
https://github.com/quarkusio/quarkus/blob/master/docs/src/main/asciidoc/gradle-tooling.adoc
Thanks!

stuartwdouglas (approved):

brentcrammond:
Is there any documentation for how to use the working dir in Maven?

paulolieuthier:
I have never used Maven directly (only Gradle), so I didn't get to test it, and so I added no documentation. Have you tried adding something like `<working-dir>${project.basedir}</working-dir>` to the plugin's configuration?

brentcrammond:
Thanks for the clue, after a bit of testing the answer is to add a workingDir under the configuration.  Where the $project.basedir will run it from the base directory.

           <plugin>
                <groupId>io.quarkus</groupId>
                <configuration>
                    <uberJar>true</uberJar>
                    **<workingDir>${project.basedir}</workingDir>**
                </configuration>
                <executions>
                    <execution>
                        <goals>
                            <goal>build</goal>
                        </goals>
                    </execution>
                </executions>
            </plugin>

Issue #4317, opened 2019-10-02
Generate extensions.json based on `quarkus-extension.properties`


maxandersen:
Once we have #833 and all extensions own its own default metadata we should generate `extensions.json` based on the properties. 

There seem to exist https://github.com/quarkusio/quarkus/pull/663 that does something similar - need to check if we can just use that or not.

kenfinnigan:
Why generate a file from a file?

For Thorntail we had the metadata for an extension in the `pom.xml` for that extension, and then we generated a file with all that information

maxandersen:
that is what is proposed here - each extension add the metadata into the pom.xml where a plugin generates quarkus-extension.properties with the values.

That is then picked up to generate the extensions.json which will be used by the platform which  mvn/gradle/cli and code.quarkus.io will use. Later one we'll have to have some kind of registry that provide an api to look these up for things in broader ecosystem.

This allows extension owners to "own" the initial metadata and reduces the need for maintanence. 

make sense now @kenfinnigan  ?

maxandersen:
and this is similar to what https://github.com/quarkusio/quarkus/pull/663 does afaics - thus wondering what happened to that ;)

kenfinnigan:
#663 I think is still there, but it's not getting information from anywhere at present, so it just generates a skeleton.

I guess I'm wondering why pom.xml -> quarkus-extension.properties -> extensions.json and why not skip the middle man?

maxandersen:
because when you access a jar you don't necessarily have the pom.xml. Some artifact in the extension is where you would have this. And what if you use gradle or gawd forbid manually wrote that file :)

Idea is that quarkus-extension.properties becomes the place to declare the metadata used for registries/platforms. extension.json is just our current mechanism to "aggregate" it if that makes sense ?

maxandersen:
the difference from #663 (unless I misread it) is that we don't have all the extensions in the same reactor/pom.xml to go generate from or am I missing your point ? :)

kenfinnigan:
The original plan with the extension generator piece I wrote is that it runs as part of the build, so, therefore, has access to all the pom.xml content.

Is that no longer the case?

kenfinnigan:
So `extension.json` is now the aggregated version?

Why not have our one, and then each extension from outside our repo generates it's own into the artifact, and we just "collate" them?

maxandersen:
[...]

PR #4423, opened 2019-10-07
Extension metadata and generation of extensions.json

### LINKED ISSUE #833: Add extension metadata to all the extensions
Following the integration of https://github.com/jbossas/protean-shamrock/pull/663/files#diff-2cfc93ecb828363d0be95549c3abbbfeR31 , we need to add the appropriate metadata to all the existing extensions.

See our `extensions.json` for some label ideas.

aloubyansky:
Fixes #833, #4317 

* Enhanced the quarkus-bootstrap:extension-descriptor mojo to also generate META-INF/quarkus-extension.json with the extension descr/props;
* Added quarkus:generate-extensions-json mojo that generates extensions.json file (and attaches it as a Maven artifact to the project) for a BOM by collecting all the META-INF/quarkus-extension.json files from its managed deps;
* Added devtools/extensions-json module that generates extensions.json file for Quarkus Core (which for now isn't used yet as it needs refinement).

I decided not to re-use `META-INF/quarkus-extension.properties` for this, I think we should deprecate it and have a simpler `META-INF/quarkus-extension-deployment-gav` or something to make it easier/faster to resolve the deployment artifact at build time.

FYI @maxandersen, @paulrobinson, @quintesse, @gastaldi

### DIFF
--- a/devtools/maven/pom.xml
+++ b/devtools/maven/pom.xml
@@ -110,6 +110,11 @@
             <artifactId>freemarker</artifactId>
         </dependency>
 
+        <dependency>
+            <groupId>org.glassfish</groupId>
+            <artifactId>javax.json</artifactId>
+        </dependency>
+
         <!-- extensions reader -->
         <dependency>
             <groupId>com.fasterxml.jackson.core</groupId>
--- a/devtools/maven/src/main/java/io/quarkus/maven/GenerateExtensionsJsonMojo.java
+++ b/devtools/maven/src/main/java/io/quarkus/maven/GenerateExtensionsJsonMojo.java
@@ -1,0 +1,197 @@
+package io.quarkus.maven;
+
+import java.io.File;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.file.FileSystem;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.Collections;
+import java.util.List;
+
+import javax.json.Json;
+import javax.json.JsonArrayBuilder;
+import javax.json.JsonObject;
+import javax.json.JsonObjectBuilder;
+import javax.json.JsonReader;
+import javax.json.JsonWriter;
+import javax.json.JsonWriterFactory;
+import javax.json.stream.JsonGenerator;
+
+import org.apache.maven.plugin.AbstractMojo;
+import org.apache.maven.plugin.MojoExecutionException;
+import org.apache.maven.plugin.MojoFailureException;
+import org.apache.maven.plugins.annotations.Component;
+import org.apache.maven.plugins.annotations.Mojo;
+import org.apache.maven.plugins.annotations.Parameter;
+import org.apache.maven.project.MavenProject;
+import org.apache.maven.project.MavenProjectHelper;
+import org.eclipse.aether.RepositorySystem;
+import org.eclipse.aether.RepositorySystemSession;
+import org.eclipse.aether.artifact.Artifact;
+import org.eclipse.aether.artifact.DefaultArtifact;
+import org.eclipse.aether.graph.Dependency;
+import org.eclipse.aether.repository.RemoteRepository;
+import org.eclipse.aether.resolution.ArtifactDescriptorException;
+import org.eclipse.aether.resolution.ArtifactDescriptorRequest;
+import org.eclipse.aether.resolution.ArtifactRequest;
+import org.eclipse.aether.resolution.ArtifactResolutionException;
+import org.eclipse.aether.resolution.ArtifactResult;
+
+import io.quarkus.bootstrap.BootstrapConstants;
+import io.quarkus.bootstrap.util.ZipUtils;
+
+/**
+ * This goal generates a list of extensions for a given BOM
+ * and stores it in a JSON format file that is later used by the tools
+ * as the catalog of available extensions.
+ */
+@Mojo(name = "generate-extensions-json")
+public class GenerateExtensionsJsonMojo extends AbstractMojo {
+
+    @Parameter(property = "bomGroupId", defaultValue = "${project.groupId}")
+    private String bomGroupId;
+
+    @Parameter(property = "bomArtifactId", defaultValue = "${project.artifactId}")
+    private String bomArtifactId;
+
+    @Parameter(property = "bomVersion", defaultValue = "${project.version}")
+    private String bomVersion;
+
+    @Parameter(property = "outputFile", defaultValue = "${project.build.directory}/extensions.json")
+    private File outputFile;
+
+    @Component
+    private RepositorySystem repoSystem;
+
+    @Parameter(defaultValue = "${repositorySystemSession}", readonly = true)
+    private RepositorySystemSession repoSession;
+
+    @Parameter(defaultValue = "${project.remoteProjectRepositories}", readonly = true, required = true)
+    private List<RemoteRepository> repos;
+
+    @Component
+    private MavenProject project;
+    @Component
+    private MavenProjectHelper projectHelper;
+
+    public GenerateExtensionsJsonMojo() {
+        MojoLogger.logSupplier = this::getLog;
+    }
+
+    @Override
+    public void execute() throws MojoExecutionException, MojoFailureException {
+
+        final DefaultArtifact bomArtifact = new DefaultArtifact(bomGroupId, bomArtifactId, "", "pom", bomVersion);
+        debug("Generating catalog of extensions for %s", bomArtifact);
+
+        final List<Dependency> deps;
+        try {
+            deps = repoSystem
+                    .readArtifactDescriptor(repoSession,
+                            new ArtifactDescriptorRequest().setRepositories(repos).setArtifact(bomArtifact))
+                    .getManagedDependencies();
+        } catch (ArtifactDescriptorException e) {
+            throw new MojoExecutionException("Failed to read descriptor of " + bomArtifact, e);
+        }
+        if (deps.isEmpty()) {
+            getLog().warn("BOM " + bomArtifact + " does not include any dependency");
+            return;
+        }
+
+        final JsonArrayBuilder extListJson = Json.createArrayBuilder();
+        for (Dependency dep : deps) {
+            final Artifact artifact = dep.getArtifact();
+            if (!artifact.getExtension().equals("jar")
+                    || "javadoc".equals(artifact.getClassifier())
+                    || "tests".equals(artifact.getClassifier())
+                    || "sources".equals(artifact.getClassifier())) {
+                continue;
+            }
+            ArtifactResult resolved = null;
+            try {
+                resolved = repoSystem.resolveArtifact(repoSession,
+                        new ArtifactRequest().setRepositories(repos).setArtifact(artifact));
+                processDependency(resolved.getArtifact(), extListJson);
+            } catch (ArtifactResolutionException e) {
+                // there are some parent poms that appear as jars for some reason
+                debug("Failed to resolve dependency %s defined in %s", artifact, bomArtifact);
+            } catch (IOException e) {
+                throw new MojoExecutionException("Failed to process dependency " + artifact, e);
+            }
+        }
+
+        final JsonObjectBuilder platformJson = Json.createObjectBuilder();
+        final JsonObjectBuilder bomJson = Json.createObjectBuilder();
+        bomJson.add("groupId", bomGroupId);
+        bomJson.add("artifactId", bomArtifactId);
+        bomJson.add("version", bomVersion);
+        platformJson.add("bom", bomJson.build());
+        platformJson.add("extensions", extListJson.build());
+
+        final File outputDir = outputFile.getParentFile();
+        if (!outputDir.exists()) {
+            if (!outputDir.mkdirs()) {
+                throw new MojoExecutionException("Failed to create output directory " + outputDir);
+            }
+        }
+        final JsonWriterFactory jsonWriterFactory = Json
+                .createWriterFactory(Collections.singletonMap(JsonGenerator.PRETTY_PRINTING, true));
+        try (JsonWriter jsonWriter = jsonWriterFactory.createWriter(new FileOutputStream(outputFile))) {
+            jsonWriter.writeObject(platformJson.build());
+        } catch (IOException e) {
+            throw new MojoExecutionException("Failed to persist " + outputFile, e);
+        }
+
+        projectHelper.attachArtifact(project, "json", null, outputFile);
+    }
+
+    private void processDependency(Artifact artifact, JsonArrayBuilder extJsonBuilder) throws IOException {
+        final Path path = artifact.getFile().toPath();
+        if (Files.isDirectory(path)) {
+            processMetaInfDir(artifact, path.resolve(BootstrapConstants.META_INF), extJsonBuilder);
+        } else {
+            try (FileSystem artifactFs = ZipUtils.newFileSystem(path)) {
+                processMetaInfDir(artifact, artifactFs.getPath(BootstrapConstants.META_INF), extJsonBuilder);
+            }
+        }
+    }
+
+    private boolean processMetaInfDir(Artifact artifact, Path metaInfDir, JsonArrayBuilder extJsonBuilder)
+            throws IOException {
+        if (!Files.exists(metaInfDir)) {
+            return false;
+        }
+        final Path p = metaInfDir.resolve(BootstrapConstants.EXTENSION_PROPS_JSON_FILE_NAME);
+        if (!Files.exists(p)) {
+            return false;
+        }
+        processPlatformArtifact(artifact, p, extJsonBuilder);
+        return true;
+    }
+
+    private void processPlatformArtifact(Artifact artifact, Path descriptor, JsonArrayBuilder extJsonBuilder)
+            throws IOException {
+        try (InputStream is = Files.newInputStream(descriptor)) {
+            try (JsonReader reader = Json.createReader(is)) {
+                final JsonObject object = reader.readObject();
+                debug("Adding Quarkus extension %s:%s", object.get("groupId"), object.get("artifactId"));
+                extJsonBuilder.add(object);
+            }
+        } catch (IOException e) {
+            throw new IOException("Failed to parse " + descriptor, e);
+        }
+    }
+
+    private void debug(String msg, Object... args) {
+        if (!getLog().isDebugEnabled()) {
+            return;
+        }
+        if (args.length == 0) {
+            getLog().debug(msg);
+            return;
+        }
+        getLog().debug(String.format(msg, args));
+    }
+}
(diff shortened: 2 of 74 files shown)


devtools/maven/src/main/java/io/quarkus/maven/GenerateExtensionsJsonMojo.java
```diff
@@ -0,0 +1,258 @@
...
+            }
+        } finally {
+            try {
+                is.close();
+            } catch (IOException e) {
+                e.printStackTrace();
+            }
+        }
+        @formatter:on
+        */
```
gastaldi: Remove this commented out code

independent-projects/bootstrap/pom.xml
```diff
@@ -24,6 +24,7 @@
         <maven.compiler.source>1.8</maven.compiler.source>
         <maven.compiler.target>1.8</maven.compiler.target>
         <!-- Versions -->
+        <javax.json.version>1.1.4</javax.json.version>
```
gastaldi: Not required, but if you want a JSON Writer with less dependencies (specially on the javax.json API), there is https://github.com/ralfstx/minimal-json

gastaldi (approved):
LGTM, added a few minor comments

maxandersen:
Weren't we going to put these in quarkus-extension.properties rather than introduce another marker file ?

aloubyansky:
@maxandersen i mentioned my reasons in the summary of the PR.

aloubyansky:
I added a few changed since @gastaldi has approved it, specifically I added the BOM GAV from which the JSON was generated to the resulting JSON. So now it's an object with attributes: `bom` and `extensions` which is an array of extensions with the usual attributes.
So now this PR needs a new approval to get merged.

extensions/agroal/runtime/src/main/resources/META-INF/quarkus-extension.json
```diff
@@ -0,0 +1,8 @@
+
+{
+    "name": "Agroal - Database connection pool",
+    "labels": [
```
gsmet: I'm wondering if we should include at least part of the info we have on the `extensions.yaml` file from the website?

See https://github.com/quarkusio/quarkusio.github.io/blob/develop/_data/extensions.yaml .

I'm thinking about:
- the description
- a key defining the category (I would keep the list of category key <-> name mapping and the order in the website - or on `code.quarkus.io`).

If we do that, I wouldn't have to add the extensions on the website.

Note that I haven't followed all the discussions so maybe it's a stupid idea :).

aloubyansky: Sure. Here I'm taking what's in the current `extensions.json` as a start.

aloubyansky:
Please, don't merge it yet. I'm looking into a build issue in Eclipse.

maxandersen:
> I decided not to re-use `META-INF/quarkus-extension.properties` for this, I think we should deprecate it and have a simpler `META-INF/quarkus-extension-deployment-gav` or something to make it easier/faster to resolve the deployment artifact at build time.

I'm not following why we wouldn't want to reuse the existing properites, nor why there should be `-gav` info at the end. Can you elaborate ?

aloubyansky:
The properties file has been used so far to store the GAV of the corresponding deployment artifact. While I did have in mind to possibly extend this file with more info (primarily provisioning related), the way it is evolving now is not looking to be effective to me.
1. We need to discover what the corresponding deployment artifact coords are during the build and test setup in the most efficient manner. Checking for a specific JAR entry containing only what we need it is a lot more efficient then loading a potentially large properties file looking for a single key.
2. It seems like what we need to have at the end is content in a JSON format, it is easier to provide it in this format in the first place and avoid transforming it into properties syntax and then back.

aloubyansky:
> Please, don't merge it yet. I'm looking into a build issue in Eclipse.

That's related to mojo parameter initialization. E.g. we've had a parameter `deployment` with some `defaultValue` properly initialized. If I add another parameter initialized the same way with the same value, this parameter will remain `null` when Eclipse is building the workspace. Yet, the code that I change in the plugin is executed, so it's not like it's picking the wrong impl of the plugin. Perhaps, the wrong metadata somehow. Perhaps, it's also related to the fact that I'm testing a SNAPSHOT version. This is too weird, I'm done with it.
The conclusion is building from Eclipse is not reliable. Not a surprise, I guess.

maxandersen:
It is intentional we are committing the marker files for each extension or do we plan to have our own extension using the mojo's ? 

also if we are introducing `META-INF/quarkus-extension.json` as the  official marker file I do think it would be better to either do `META-INF/quarkus-extension.yaml` or at least support both formats.

This was one of the reasons I was liking more `META-INF/quarkus-extension.properties` to start :)

aloubyansky:
> It is intentional we are committing the marker files for each extension or do we plan to have our own extension using the mojo's ?

Not sure I understand your question. Every extension is supposed to include the description. Whether it's generated or written maintained manually.

> also if we are introducing `META-INF/quarkus-extension.json` as the official marker file I do think it would be better to either do `META-INF/quarkus-extension.yaml` or at least support both formats.

Why?

> This was one of the reasons I was liking more `META-INF/quarkus-extension.properties` to start :)

To later support 3 formats?

maxandersen:
> Not sure I understand your question. Every extension is supposed to include the description. Whether it's generated or written maintained manually.

just curious as thought you said you wanted the extensions in core to have the maven plugin generate it.

maxandersen:
> > also if we are introducing `META-INF/quarkus-extension.json` as the official marker file I do think it would be better to either do `META-INF/quarkus-extension.yaml` or at least support both formats.
> 
> Why?

yaml is better for humans  - json is for machines ;)

maxandersen:
> > This was one of the reasons I was liking more `META-INF/quarkus-extension.properties` to start :)
> 
> To later support 3 formats?

no, to avoid having to deal with  anything but built-in to java parsing/formats :)

devtools/maven/src/main/java/io/quarkus/maven/GenerateExtensionsJsonMojo.java
```diff
@@ -0,0 +1,203 @@
...
+        final Path path = artifact.getFile().toPath();
+        try {
+            if (Files.isDirectory(path)) {
+                processMetaInfDir(artifact, path.resolve(BootstrapConstants.META_INF), extJsonBuilder);
+            } else {
+                try (FileSystem artifactFs = ZipUtils.newFileSystem(path)) {
+                    processMetaInfDir(artifact, artifactFs.getPath(BootstrapConstants.META_INF), extJsonBuilder);
+                }
+            }
+        } catch (Throwable t) {
```
maxandersen: shouldn't this not just catch (Exception e) ?

devtools/maven/pom.xml
```diff
@@ -100,6 +100,11 @@
             <artifactId>freemarker</artifactId>
         </dependency>
 
+        <dependency>
+            <groupId>org.glassfish</groupId>
```
maxandersen: would this not be better/sufficient:

```
<groupId>javax.json</groupId>
<artifactId>javax.json-api<artifactId>
```

devtools/maven/src/main/java/io/quarkus/maven/GenerateExtensionsJsonMojo.java
```diff
@@ -0,0 +1,203 @@
...
+            throw new DeploymentInjectionException("Failed to inject extension deplpyment dependencies", t);
+        }
+    }
+
+    private boolean processMetaInfDir(Artifact artifact, Path metaInfDir, JsonArrayBuilder extJsonBuilder)
+            throws BootstrapDependencyProcessingException {
+        if (!Files.exists(metaInfDir)) {
+            return false;
+        }
+        final Path p = metaInfDir.resolve(BootstrapConstants.EXTENSION_PROPS_JSON_FILE_NAME);
```
maxandersen: I'm getting a compile error on this:

```
ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.8.0-jboss-2:compile (default-compile) on project quarkus-maven-plugin: Compilation failure
[ERROR] /Users/max/code/quarkus/quarkus/devtools/maven/src/main/java/io/quarkus/maven/GenerateExtensionsJsonMojo.java:[172,61] cannot find symbol
[ERROR]   symbol:   variable EXTENSION_PROPS_JSON_FILE_NAME
[ERROR]   location: interface io.quarkus.bootstrap.BootstrapConstants
```

missing a commit/add of `BootstrapConstants.java` ?

maxandersen: never mind - mistakenly only compiled devtools ;)

extensions/agroal/runtime/src/main/resources/META-INF/quarkus-extension.json
maxandersen: +1 on getting the minimal version in first.
Next step there efter is to add the additional features such as description, category, etc.

independent-projects/bootstrap/maven-plugin/src/main/java/io/quarkus/maven/ExtensionDescriptorMojo.java
```diff
@@ -18,14 +19,22 @@
...
  * Generates Quarkus extension descriptor for the runtime artifact.
  *
+ * <p/>Also generates META-INF/quarkus-extension.json which includes properties of the extension
+ * such as name, labels, maven coordinates, etc that are used by the tools.
+ *
  * @author Alexey Loubyansky
  */
-@Mojo(name = "extension-descriptor", defaultPhase = LifecyclePhase.COMPILE, requiresDependencyResolution = ResolutionScope.COMPILE_PLUS_RUNTIME)
+@Mojo(name = "extension-descriptor", defaultPhase = LifecyclePhase.PROCESS_RESOURCES, requiresDependencyResolution = ResolutionScope.COMPILE_PLUS_RUNTIME)
 public class ExtensionDescriptorMojo extends AbstractMojo {
```
maxandersen: is it expected that on first build you get errors like:

```
ERROR] Could not find goal 'generate-extensions-json' in plugin io.quarkus:quarkus-maven-plugin:999-SNAPSHOT among available goals add-extension, add-extensions, analyze-call-tree, build, create, create-extension, dev, generate-config, help, list-extensions, native-image, remote-dev -> [Help 1]
``` ?

second build works since then the plugin is available to use.

its as if the sequence of builds are not right.

not a biggie but just worth mentioning in case it can be avoided.

**Review by maxandersen — APPROVED** [review]
There are few minor things in this but I approve this as its generally what we want to get to and I would rather have generation of extensions.json in core now than wait.

The other things we can fix in follow-up PR's as we haven't yet removed core extensions.json thus not breaking anything.

**Comment by maxandersen** [comment]
just noticed that the generated quarkus-extension.json in some cases ends up with:

```
{
  "groupId": null,
  "artifactId": null
}
```
i.e.  `./vertx-core/runtime/target/classes/META-INF/quarkus-extension.json`

others have no GAV at all:
`./jgit/runtime/src/main/resources/META-INF/quarkus-extension.json`:

```
{
    "name": "JGit",
    "labels": [
        "git"
    ]
}
```

and finally also when existing data you get:

`./flyway/runtime/target/classes/META-INF/quarkus-extension.json`

```
{
  "name": "Flyway",
  "labels": [
    "flyway",
    "database",
    "data"
  ],
  "guide": "https://quarkus.io/guides/flyway-guide",
  "groupId": null,
  "artifactId": null
}
```

Shouldn't there at least be a groupid and artifactid in here ?

**Comment by gsmet** [comment]
@maxandersen I wasn't able to reproduce your last issue with groupId and artifactId being null. Everything is OK here.

**Review by gsmet — COMMENTED** [review]
I added a few questions.

It needs a rebase and once it's done, if you could run a full build with an empty repo and check that:
* it works
* we don't have these null groupIds and artifactIds issues

Once you have checked that, I would say let's commit it and make incremental changes on top of it.

**Comment by maxandersen** [comment]
> I wasn't able to reproduce your last issue with groupId and artifactId being null. Everything is OK here.

shouldn't `mvn package` be sufficient ?

**Comment by gsmet** [comment]
Ah, that's a good question for Alexey. With Quarkus, I usually do `mvn clean install`

**Comment by maxandersen** [comment]
I did a new run with `mvn clean package` and now I have groupId/artifactId filled out where they before was `null`.

Still have neither in ` ./jgit/runtime/src/main/resources/META-INF/quarkus-extension.json`:
```
{
    "name": "JGit",
    "labels": [
        "git"
    ]
}
```

**Comment by aloubyansky** [comment]
> > > also if we are introducing `META-INF/quarkus-extension.json` as the official marker file I do think it would be better to either do `META-INF/quarkus-extension.yaml` or at least support both formats.
> > 
> > 
> > Why?
> yaml is better for humans - json is for machines ;)

I can do that. But I'll keep only one format. The aggregated extensions list will still be in JSON though, I guess.

File: devtools/maven/src/main/java/io/quarkus/maven/GenerateExtensionsJsonMojo.java (hunk 4 above)
Comment (aloubyansky) [reply devtools/maven/src/main/java/io/quarkus/maven/GenerateExtensionsJsonMojo.java]: Thanks, I am actually throwing a wrong exception here. Copy-pasted from another class.

File: independent-projects/bootstrap/maven-plugin/pom.xml [hunk 8]
```diff
@@ -61,6 +61,10 @@
             <groupId>org.apache.maven</groupId>
             <artifactId>maven-core</artifactId>
         </dependency>
+        <dependency>
+            <groupId>com.eclipsesource.minimal-json</groupId>
+            <artifactId>minimal-json</artifactId>
```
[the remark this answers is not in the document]
aloubyansky: I was originally using the glassfish one but then George suggested me to use this impl as a smaller and faster one. Given that it's an independent project, I didn't mind changing. I won't mind changing back neither.

independent-projects/bootstrap/maven-plugin/src/main/java/io/quarkus/maven/ExtensionDescriptorMojo.java
aloubyansky: I couldn't reproduce it neither. It might have used a locally available version of the plugin that didn't include this mojo. If you remove quarkus artifacts from your local maven repo and rebuild you shouldn't see this issue.

aloubyansky:
> just noticed that the generated quarkus-extension.json in some cases ends up with:
> 
> ```
> {
>   "groupId": null,
>   "artifactId": null
> }
> ```
> 
> i.e. `./vertx-core/runtime/target/classes/META-INF/quarkus-extension.json`
> 
> others have no GAV at all:
> `./jgit/runtime/src/main/resources/META-INF/quarkus-extension.json`:
> 
> ```
> {
>     "name": "JGit",
>     "labels": [
>         "git"
>     ]
> }
> ```
> 
> and finally also when existing data you get:
> 
> `./flyway/runtime/target/classes/META-INF/quarkus-extension.json`
> 
> ```
> {
>   "name": "Flyway",
>   "labels": [
>     "flyway",
>     "database",
>     "data"
>   ],
>   "guide": "https://quarkus.io/guides/flyway-guide",
>   "groupId": null,
>   "artifactId": null
> }
> ```
> 
> Shouldn't there at least be a groupid and artifactid in here ?

That's probably your IDE building it? There is a comment somewhere above where I summarized my struggling figuring out why Eclipse (in my case) didn't properly initialize Mojo parameters.

aloubyansky:
> I did a new run with `mvn clean package` and now I have groupId/artifactId filled out where they before was `null`.
> 
> Still have neither in ` ./jgit/runtime/src/main/resources/META-INF/quarkus-extension.json`:
> 
> ```
> {
>     "name": "JGit",
>     "labels": [
>         "git"
>     ]
> }
> ```

Yes, but that's `src` dir. It's a template. You should be checking the `target` one.

gsmet:
The `target` one is perfectly OK for me.

aloubyansky:
> > I wasn't able to reproduce your last issue with groupId and artifactId being null. Everything is OK here.
> 
> shouldn't `mvn package` be sufficient ?

Worked for me.

devtools/maven/pom.xml
maxandersen: ah - gdammit - keep forgetting that detail ;)

independent-projects/bootstrap/maven-plugin/src/main/java/io/quarkus/maven/ExtensionDescriptorMojo.java
maxandersen: i'll assume I had a bad state as I can't reproduce it neither now :)

maxandersen:
> That's probably your IDE building it? There is a comment somewhere above where I summarized my struggling figuring out why Eclipse (in my case) didn't properly initialize Mojo parameters.

could be - I had code open that uses m2e behind the scenes.
Lets not block this on that issue but I opened https://github.com/quarkusio/quarkus/issues/4542 as it would affect extension writers so would be good to have a way to handle/workaround this.

FroMage:
Damnit, I only noticed this issue now. I would prefer to generate that `quarkus-extension.json` file from our existing maven extension plugin if the following properties are defined (and the file does not exist already):

```
<project>
 ...
 <properties>
  <quarkus.extension.name>RESTEasy - Jackson</quarkus.extension.name>
  <quarkus.extension.labels>resteasy-jackson,jaxrs-json,resteasy-json,resteasy,jaxrs,json,jackson</quarkus.extension.labels>
 </properties>
...
</project>
```

independent-projects/bootstrap/maven-plugin/src/main/java/io/quarkus/maven/ExtensionDescriptorMojo.java
```diff
@@ -78,5 +95,37 @@ public void execute() throws MojoExecutionException {
         } catch(IOException e) {
             throw new MojoExecutionException("Failed to persist extension descriptor " + output.resolve(BootstrapConstants.DESCRIPTOR_FILE_NAME), e);
         }
+
```
FroMage: @aloubyansky I don't understand this code: it appears to read the file if it exists, and then writes it back. Surely I'm reading this wrong?

aloubyansky: Not entirely wrong :) At this point, the file that's actually on the disk is missing groupId and artifactId, which makes it more like a template that needs to be complete by the plugin. However, you're right, it could originally be complete, in which case persisting it is useless.

FroMage: But you're reading it from the output folder. Shouldn't it be read from the src folder?

aloubyansky: I am reading it from the output folder to support filtering of resources, just in case.

FroMage: OK, if it's done on purpose then fine :)

ia3andy:
> Please, don't merge it yet. I'm looking into a build issue in Eclipse.

We should introduce the WIP bot in here?

maxandersen:
> We should introduce the WIP bot in here?

isn't that superseeded by github now directly supporting drafts pull-requests ?

PR #4426, opened 2019-10-08
Add OIDC Cors test and update the docs

### LINKED ISSUE #3441: OIDC adapter should reuse Quarkus CORS filter
**Description**
Right now there are 2 CORS filters that can be used, if KC is active - the KC one has to be enabled, if not - the Quarkus one. Ideally the users should be able to use the Quarkus one with KC as well
**Implementation ideas**
Feedback from the KC experts, @pedroigor , @sebastienblanc will be welcome :-). Perhaps the KC extension can detect the Quarkus CORS capabiity and depend on it...

sberyozkin:
Fixes #3441

### DIFF
--- a/docs/src/main/asciidoc/http-reference.adoc
+++ b/docs/src/main/asciidoc/http-reference.adoc
@@ -153,12 +153,6 @@ quarkus.http.cors.exposed-headers=Content-Disposition
 quarkus.http.cors.access-control-max-age=24H
 ----
 
-[WARNING]
-====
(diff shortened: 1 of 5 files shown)


sberyozkin:
@stuartwdouglas, @gsmet 
The elytron-security-jdbc test failures are definitely not related to this PR, it only updates the oidc test and the docs, so I think it is good to go as it fixes the docs, but I've updated the docs a bit again and rebased, may be it will help.

PR #4427, opened 2019-10-08
Use the plugins DSL to both resolve and apply plugins

segalaj:
Small doc improvement.
This is now the recommended way to use plugins: https://docs.gradle.org/current/userguide/plugins.html#sec:using_plugins

Don't know who to add as reviewer :smile:

### DIFF
--- a/docs/src/main/asciidoc/gradle-tooling.adoc
+++ b/docs/src/main/asciidoc/gradle-tooling.adoc
@@ -89,8 +89,8 @@ Here's the same build script, using the Gradle Kotlin DSL:
 ----
 plugins {
     java
+    id("io.quarkus") version "{quarkus-version}"
 }
-apply(plugin = "io.quarkus")
 
 repositories {
     mavenCentral()


aloubyansky:
Thanks for opening a PR for this. We actually had a reason for applying it that way. It should be good now though. I am going to review the docs today to see whether there are more updates for Gradle docs.

docs/src/main/asciidoc/gradle-tooling.adoc
```diff
@@ -89,8 +89,8 @@ Here's the same build script, using the Gradle Kotlin DSL:
 ----
 plugins {
     java
+    id("io.quarkus") version "{quarkus-version}"
```
Sanne: could you use single quotes ''' rather than double quotes here '"' please ?

Sanne:
The conflict confuses me. Was this example entirely removed, or just moves to somewhere else?

gsmet:
@geoand @gastaldi could you check if we still need a change here? I thought this was addressed shortly after we published the plugin?

geoand:
I haven't followed the lates Gradle stuff unfortunately so I would welcome comments from @aloubyansky or @gastaldi

gastaldi:
AFAIR The `apply:` was necessary before the plugin was published in the Gradle portal. Now it can be used like this: https://github.com/gastaldi/gradle-quarkus-project/blob/master/build.gradle#L3

gsmet:
AFAICS, this part has been removed from the guide since we published the plugin so let's close this one.

Issue #4371, opened 2019-10-04
Make native image test use the test profile or not


emmanuelbernard:
I know we had some conversations recently on the subject but I don't recall the conclusion.
It was related to the shift of hibernate's import file feature having different defaults in prod vs dev.

So let's capture what we really want here and see if we need code change and / or documentation changes before 1.0

Today @cescoffier and @agoncal got hit by IT run in prod and thus confusion.

FroMage:
My suggestion was that if you want to test a native-image you need to build it with the test profile.

gsmet:
I agree on that. We need to find a workflow to do that that is Maven compatible and not too cumbersome.

Typically, with a standard Maven behavior, you would have:
```
$ mvn clean install -Dnative
> ...
> build test native image
> test native image
> build prod native image
```

```
$ mvn clean install -DskipITs -Dnative
> ...
> build prod native image
```

That works.

Now, what I'm wondering is how we can only build the test image and run the tests (and not build the prod one) while still being in a somewhat natural Maven workflow. Is it something we want to expose to the users (we will need it for our CI but we can have an internal thing)? If yes, how?

rsvoboda:
> ... build test native image
> ... test native image
> ... build prod native image

How is ensured that test profile and prod profile are functionally the same without any chance for regression? With my super-skeptical hat on: Different bits => no guarantees.

gsmet:
Well, you won't have more guarantees than when you use H2 to test whereas you use a PostgreSQL instance in production. There is a tradeoff for that. But the current situation makes it's very cumbersome to properly test a Java + native image application as both are using different profiles during the test phase.

cescoffier:
In #4430 I've allowed the user to specify the profile he wants.

stuartwdouglas:
I think this can be closed, there are no changes around this planned, I don't think we want to build separate native image test executables.

PR #4430, opened 2019-10-08
Allows configuring the profile used during the native tests

cescoffier:
Related to https://github.com/quarkusio/quarkus/issues/4371

### DIFF
--- a/docs/src/main/asciidoc/building-native-image-guide.adoc
+++ b/docs/src/main/asciidoc/building-native-image-guide.adoc
@@ -209,6 +209,13 @@ duration can be changed using the `quarkus.test.native-image-wait-time` system p
 to 300 seconds, use: `./mvnw verify -Pnative -Dquarkus.test.native-image-wait-time=300`.
 ====
 
+By default, native tests runs using the `prod` profile.
+This can be overridden using the `quarkus.test.native-image-profile` property.
+For example, in your `application.properties` file, add: `quarkus.test.native-image-profile=test`.
+Alternatively, you can run your tests with: `./mvnw verify -Pnative -Dquarkus.test.native-image-profile=test`.
+However, don't forget that when the native executable is built the `prod` profile is enabled.
+So, the profile you enable this way must be compatible with the produced executable.
+
 === Excluding tests when running as a native executable
 
 When running tests this way, the only things that actually run natively are you application endpoints, which
--- a/test-framework/common/src/main/java/io/quarkus/test/common/NativeImageLauncher.java
+++ b/test-framework/common/src/main/java/io/quarkus/test/common/NativeImageLauncher.java
@@ -25,6 +25,7 @@ public class NativeImageLauncher implements Closeable {
     private static final long DEFAULT_IMAGE_WAIT_TIME = 60;
 
     private final Class<?> testClass;
+    private final String profile;
     private Process quarkusProcess;
     private final int port;
     private final long imageWaitTime;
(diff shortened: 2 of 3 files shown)


docs/src/main/asciidoc/building-native-image-guide.adoc
```diff
@@ -209,6 +209,13 @@ duration can be changed using the `quarkus.test.native-image-wait-time` system p
 to 300 seconds, use: `./mvnw verify -Pnative -Dquarkus.test.native-image-wait-time=300`.
 ====
 
+By default, native tests runs using the `prod` profile.
```
aloubyansky: IMO, that's confusing. Especially taking into account the way we are writing the tests for the quickstarts where the native test simply extends the JVM one. So if you have something in the test profile, the JVM one will pass and the native one will fail (unless you explicitly enable the profile for it, of course).

test-framework/common/src/main/java/io/quarkus/test/common/NativeImageLauncher.java
```diff
@@ -35,10 +36,12 @@ public NativeImageLauncher(Class<?> testClass) {
         this(testClass,
                 ConfigProvider.getConfig().getOptionalValue("quarkus.http.test-port", Integer.class).orElse(DEFAULT_PORT),
                 ConfigProvider.getConfig().getOptionalValue("quarkus.test.native-image-wait-time", Long.class)
-                        .orElse(DEFAULT_IMAGE_WAIT_TIME));
+                        .orElse(DEFAULT_IMAGE_WAIT_TIME),
+                ConfigProvider.getConfig().getOptionalValue("quarkus.test.native-image-profile", String.class)
```
gsmet: Please add it here: https://github.com/quarkusio/quarkus/blob/master/core/deployment/src/main/java/io/quarkus/deployment/configuration/ConfigDefinition.java#L74 .

test-framework/common/src/main/java/io/quarkus/test/common/NativeImageLauncher.java
```diff
@@ -60,6 +64,9 @@ public void start() throws IOException {
         args.add("-Dquarkus.http.port=" + port);
         args.add("-Dtest.url=" + TestHTTPResourceManager.getUri());
         args.add("-Dquarkus.log.file.path=" + PropertyTestUtil.getLogFileLocation());
+        if (profile != null) {
+            args.add("-Dquarkus.profile=" + profile);
```
gsmet: `prod` will be the default value here? I'm not totally sure.

gsmet (changes requested):

test-framework/common/src/main/java/io/quarkus/test/common/NativeImageLauncher.java
cescoffier: Yes, but there is an ongoing discussion about what should be the default.
(https://github.com/quarkusio/quarkus/issues/4371)

So, I've decided to not override the current default or next-to-be default. It's something we could revisit. 

So, this PR just offers the user the possibility to customize it.

docs/src/main/asciidoc/building-native-image-guide.adoc
cescoffier: @aloubyansky I learned that the hard way (dropping my prod database, oupsy...)
More discussion is happening here: https://github.com/quarkusio/quarkus/issues/4371

test-framework/common/src/main/java/io/quarkus/test/common/NativeImageLauncher.java
cescoffier: Fixed. I didn't know about this actually.

aloubyansky (approved):

gsmet (approved):

docs/src/main/asciidoc/building-native-image-guide.adoc
```diff
@@ -209,6 +209,13 @@ duration can be changed using the `quarkus.test.native-image-wait-time` system p
...
+This can be overridden using the `quarkus.test.native-image-profile` property.
+For example, in your `application.properties` file, add: `quarkus.test.native-image-profile=test`.
+Alternatively, you can run your tests with: `./mvnw verify -Pnative -Dquarkus.test.native-image-profile=test`.
+However, don't forget that when the native executable is built the `prod` profile is enabled.
```
emmanuelbernard: I find this sentence confusing. Are you saying that your tests will run with the test profile but pointing to the image that was built as prod.
Or that your image will be build with test and that should be compatible with the "prod" environment you have in mind.

My fear is that this is really inside baseball here and that's typically this is written for the implementer to feel happy that he has covered all cases than in the spirit of helping the user. If we see different use cases, we should enunciate them better so people could connect to the problem + solution.

This is about improvement so if we cannot make it I won't die for it.

PR #4434, opened 2019-10-08
[WIP] Start the TransactionStatusManager at runtime

gsmet:
It uses to be started at runtime as TxControl starts it in its static
initialization.

Draft PR for now, I want to see what CI has to say.

### DIFF
--- a/extensions/narayana-jta/deployment/src/main/java/io/quarkus/narayana/jta/deployment/NarayanaJtaProcessor.java
+++ b/extensions/narayana-jta/deployment/src/main/java/io/quarkus/narayana/jta/deployment/NarayanaJtaProcessor.java
@@ -51,9 +51,16 @@ class NarayanaJtaProcessor {
     @Record(RUNTIME_INIT)
     public void build(NarayanaJtaRecorder recorder, BuildProducer<FeatureBuildItem> feature) {
         feature.produce(new FeatureBuildItem(FeatureBuildItem.NARAYANA_JTA));
+
         additionalBeans.produce(new AdditionalBeanBuildItem(NarayanaJtaProducers.class));
+
         runtimeInit.produce(new RuntimeInitializedClassBuildItem(
                 "com.arjuna.ats.internal.jta.resources.arjunacore.CommitMarkableResourceRecord"));
+        runtimeInit.produce(new RuntimeInitializedClassBuildItem("com.arjuna.ats.arjuna.coordinator.TxControl"));
+        runtimeInit.produce(new RuntimeInitializedClassBuildItem("com.arjuna.ats.arjuna.recovery.ActionStatusService"));
+        runtimeInit.produce(new RuntimeInitializedClassBuildItem(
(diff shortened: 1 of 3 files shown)


Sanne:
@gsmet I suspect we'll need something around these lines. Will you have time to finish, need any help?

gsmet:
So this PR works. The only thing about it is that it hardcodes a few things that are supposed to be configurable. That's why it was a draft open for discussion.

extensions/narayana-jta/deployment/src/main/java/io/quarkus/narayana/jta/deployment/NarayanaJtaProcessor.java
```diff
@@ -51,9 +51,16 @@
     @Record(RUNTIME_INIT)
     public void build(NarayanaJtaRecorder recorder, BuildProducer<FeatureBuildItem> feature) {
         feature.produce(new FeatureBuildItem(FeatureBuildItem.NARAYANA_JTA));
+
         additionalBeans.produce(new AdditionalBeanBuildItem(NarayanaJtaProducers.class));
+
         runtimeInit.produce(new RuntimeInitializedClassBuildItem(
                 "com.arjuna.ats.internal.jta.resources.arjunacore.CommitMarkableResourceRecord"));
+        runtimeInit.produce(new RuntimeInitializedClassBuildItem("com.arjuna.ats.arjuna.coordinator.TxControl"));
+        runtimeInit.produce(new RuntimeInitializedClassBuildItem("com.arjuna.ats.arjuna.recovery.ActionStatusService"));
```
Sanne: the `com.arjuna.ats.arjuna.recovery.ActionStatusService` doesn't seem to have static initializers. Did you mean to force re-initialization to null of the one field instead? I don't think that's accomplished with this specific build item.

gsmet: I added it at some point because otherwise native-image complained for some obscure reasons. I made several iterations though so maybe it's not needed in the end.

gsmet: What happens if you comment it? Does it still work?

Sanne: I'll try.

Sanne: I got no complaints.. I guess it's no longer necessary.  But I'll try to see if that static field needs resetting.

How confident are you that you tracked down all references to `Uid` s? Do we have a way to ensure there's none in the captured heap? I kinda remember doing such checks but don't remember if we had a way to do it in an integration test.

gsmet:
This is not useful anymore considering https://github.com/quarkusio/quarkus/pull/4799 .

Closing.

PR #4436, opened 2019-10-08
Class loader fixes

stuartwdouglas:
Also contains a horrible hack to work around an OpenAPI problem that should be reverted.


aloubyansky:
There is a CCE failure.

geoand:
I am going to close this as it's been superseded by #4438

Issue #3284, opened 2019-07-19
Hibernate validator and annotated static method on enum


pehunka:
**Describe the bug**
Application fails to start when an enum type has annotated static method with bean validation constraint.

**Expected behavior**
I will expect that quarkus app will compile and ignore the validation constraint as not supported by the bean validation specification.

**Actual behavior**
Application crashes with following stack trace

```
[io.qua.dev.DevModeMain] Failed to start quarkus: java.lang.ExceptionInInitializerError ..
Caused by: java.lang.ArrayIndexOutOfBoundsException: 1
at io.quarkus.hibernate.validator.runtime.HibernateValidatorRecorder
```

**To Reproduce**
```
mvn io.quarkus:quarkus-maven-plugin:0.19.1:create \
    -DprojectGroupId=org.acme \
    -DprojectArtifactId=getting-started \
    -DclassName="org.acme.quickstart.GreetingResource" \
    -Dpath="/hello"
```

Add enum with static factory method annotated with a bean validation constraint:

```java
public enum Gender {
    MALE,
    FEMALE;

    public static Gender fromCode(@NotNull String code) {
        return Gender.valueOf(code);
    }
}
```

Add dependency on hibernate validator to your pom:

```xml
 <dependency>
     <groupId>io.quarkus</groupId>
     <artifactId>quarkus-hibernate-validator</artifactId>
 </dependency>
```

```
mvn compile quarkus:dev
```

**Environment (please complete the following information):**
```terminal
 Java version: 1.8.0_212, vendor: Amazon.com Inc., runtime: /Library/Java/JavaVirtualMachines/amazon-corretto-8.jdk/Contents/Home/jre
Default locale: en_CZ, platform encoding: UTF-8
OS name: "mac os x", version: "10.14.5", arch: "x86_64", family: "mac"
```

**Additional context**
I've tried POJO instead of Enum and it works.

gsmet:
Usually these issues are related to synthetic parameters being added in enum methods.

I'll have a look.

gsmet:
There is a tentative PR here: https://github.com/hibernate/hibernate-validator/pull/1049 .

While I suppose the PR fixes the issue mentioned here, it looks like there is something weird going on so we need to fully understand the issue before merging.

yrodiere:
[...]

Issue #2556, opened 2019-05-22
Adding javax.validation @Valid in JAX-RS resource method which override an interface break HibernateValidatorTemplate


masini:
**Describe the bug**

I have an interface with a single method:

```java
public interface GreetingResourceInterface<T> {
    String hello(T pojo);
}
```
and a JAX-RS resource which implements it:

```java
@Path("/hello")
public class GreetingResource implements GreetingResourceInterface<MyPojo> {

    @Override
    @POST
    @Produces(MediaType.TEXT_PLAIN)
    @Consumes(MediaType.APPLICATION_JSON)
    public String hello(@Valid MyPojo pojo) {
        return "hello";
    }
}
```

**Expected behavior**

The parameter is "validated" with javax.validation.

**Actual behavior**

At startup with:

    mvn clean compile quarkus:dev
    
I got this Exception:

```java
14:02:18,137 ERROR [io.qua.dev.DevModeMain] Failed to start quarkus: java.lang.ExceptionInInitializerError
        at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
        at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
        at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
        at java.lang.reflect.Constructor.newInstance(Constructor.java:423)
        at java.lang.Class.newInstance(Class.java:442)
        at io.quarkus.runner.RuntimeRunner.run(RuntimeRunner.java:121)
        at io.quarkus.dev.DevModeMain.doStart(DevModeMain.java:159)
        at io.quarkus.dev.DevModeMain.main(DevModeMain.java:93)
Caused by: java.lang.RuntimeException: Failed to start quarkus
        at io.quarkus.runner.ApplicationImpl1.<clinit>(Unknown Source)
        ... 8 more
Caused by: java.lang.ArrayIndexOutOfBoundsException: 1
        at org.hibernate.validator.internal.properties.javabean.JavaBeanExecutable.getParameters(JavaBeanExecutable.java:213)
        at org.hibernate.validator.internal.properties.javabean.JavaBeanExecutable.<init>(JavaBeanExecutable.java:45)
        at org.hibernate.validator.internal.properties.javabean.JavaBeanConstructor.<init>(JavaBeanConstructor.java:20)
[...]

PR #4439, opened 2019-10-08
Test case for issues #3284 and #2556

yrodiere:
This tests:

* #3284 enums annotated with Hibernate Validator annotations
* #2556 implementations of generic interfaces with parameters annotated with Hibernate Validator annotations.

This fails with the current version of Hibernate Validator. We'll need a Hibernate Validator upgrade on top of this to fix the problem (HV-1730).


gsmet:
I just rebased it after the HV upgrade.

gsmet (approved):

PR #4441, opened 2019-10-08
Force the version of the Maven Compiler Plugin in the templates and the -parameters option

gsmet:
Right now, with our default project template, the version used is 3.1 (latest is 3.8.1). This is an issue.

Also using 3.6.2+ is needed to be able to use `maven.compiler.parameters`, which is better than overriding the compiler arguments.

Let's merge this at the beginning of the next development cycle.

### DIFF
--- a/build-parent/pom.xml
+++ b/build-parent/pom.xml
@@ -25,6 +25,7 @@
         <compiler-plugin.version>3.8.1</compiler-plugin.version>
         <kotlin.version>1.3.21</kotlin.version>
         <scala.version>2.12.8</scala.version>
+        <scala-plugin.version>4.1.1</scala-plugin.version>
 
         <!-- These properties are needed in order for them to be resolvable by the documentation -->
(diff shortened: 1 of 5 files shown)


geoand:
Should be easy to do in gradle as well: https://stackoverflow.com/a/37482148/2504224

cescoffier (approved):

yrodiere (approved):

aureamunoz:
gradle stuff done [here](https://github.com/quarkusio/quarkus/pull/4444)

geoand:
Very nice addition, merging!

geoand:
@gsmet I merged by accident without reading the comment. Should we revert?

gsmet:
@geoand well, you now have to verify it thoroughly :).

I did test with a basic template, Scala and Kotlin project but let's be on the safe side.

geoand:
@gsmet OK, I'll do some tests and let you know.

PR #4443, opened 2019-10-08
Periodically validate JDBC connections

### LINKED ISSUE #2336: Broken pipe (Write failed) using Hibernate ORM and MariaDB
Using ```quarkus-hibernate-orm``` along with ```quarkus-jdbc-mariadb``` with the following application.properties doesn't always work after the application been idle for a few hours. After the application been idling during the night the first few transaction fails (see stacktrace). Could it be that stale connections doesn't get evicted from the connection pool properly? 

(Quarkus version 0.14.0, MariaDB 10.2.15, MacOS Mojave and OpenJDK Runtime Environment 18.9 build 11+28)

```
quarkus.datasource.url=jdbc:mariadb://some-url...
quarkus.datasource.driver=org.mariadb.jdbc.Driver
quarkus.hibernate-orm.dialect=org.hibernate.dialect.MariaDB102Dialect
quarkus.datasource.background-validation-interval=PT1M
quarkus.datasource.idle-removal-interval=PT1M
quarkus.datasource.username=someUserName
quarkus.datasource.password=somePwd
```

stacktrace

yrodiere:
Fixes #2336 

Agroal uses `ConnectionValidator.emptyValidator()` by default,
so even if we set the default validation interval to 2 minutes by
default, it will just do nothing every 2 minutes.

Tested manually by putting a thread-suspending breakpoint on the line `storeTestPersons(entityManagerFactory);` in `JPAFunctionalityTestEndpoint`.
Just run `io.quarkus.it.jpa.mariadb.JPAFunctionalityTest`, and when the breakpoint is hit, restart the database, wait 2 minutes, and resume execution. Before this patch, the call to `storeTestPersons(entityManagerFactory);` would fail because a stale connection was in the ppol; after this patch, it will succeed because the connection was removed and renewed.

### DIFF
--- a/extensions/agroal/runtime/src/main/java/io/quarkus/agroal/runtime/AbstractDataSourceProducer.java
+++ b/extensions/agroal/runtime/src/main/java/io/quarkus/agroal/runtime/AbstractDataSourceProducer.java
@@ -16,6 +16,7 @@ import javax.transaction.TransactionSynchronizationRegistry;
 import org.jboss.logging.Logger;
 
 import io.agroal.api.AgroalDataSource;
+import io.agroal.api.configuration.AgroalConnectionPoolConfiguration.ConnectionValidator;
 import io.agroal.api.configuration.supplier.AgroalConnectionFactoryConfigurationSupplier;
 import io.agroal.api.configuration.supplier.AgroalConnectionPoolConfigurationSupplier;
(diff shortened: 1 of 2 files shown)


gsmet (approved):

Sanne:
thanks @yrodiere !

gsmet:
Windows failure is unrelated, I'll merge this.

Thanks @yrodiere !

Issue #3992, opened 2019-09-12
Support not supplying a value for name in Spring Web @PathVariable


geoand:
In the current Quarkus integration when using Spring Web's `@PathVariable` one needs to supply a value for `name` (or `value` which is an alias) like so:

```
    public Greeting greet(@PathVariable(name = "message") String message) {
...
    }
```

In Spring Web this is not absolutely necessary because Spring can use the method parameter name (if the `-parameter` flag has been supplied during source code compilation).
We should try and do the same.

See: https://github.com/quarkusio/quarkusio.github.io/issues/289

jclingan:
BTW, I can confirm that @RequestParam requires a name attribute.

aureamunoz:
fixed in part on RESTEasy [PR 2177](https://github.com/resteasy/Resteasy/pull/2177)

geoand:
This has already been implemented

PR #4444, opened 2019-10-08
feat: configure java compiler plugin in gradle to add parameters arg

aureamunoz:
Related to issue #3992
In order to allow quarkus integration when using Spring Web's @PathVariable works without a value for `name` we need to enable the java compilation of parameters to the bytecode.
The maven part seems to be done in this [PR](https://github.com/quarkusio/quarkus/pull/4441), this is the gradle part.
I let you take a look and give some feed backs @gastaldi @gsmet @geoand

### DIFF
--- a/devtools/common/src/main/java/io/quarkus/cli/commands/file/GradleBuildFile.java
+++ b/devtools/common/src/main/java/io/quarkus/cli/commands/file/GradleBuildFile.java
@@ -42,7 +42,8 @@ public class GradleBuildFile extends BuildFile {
     }
 
     @Override
-    public void completeFile(String groupId, String artifactId, String version) throws IOException {
+    public void completeFile(String groupId, String artifactId, String version)
+            throws IOException {
         init();
         completeSettingsContent(artifactId);
         completeBuildContent(groupId, version);
--- a/devtools/common/src/main/java/io/quarkus/cli/commands/file/BuildFile.java
+++ b/devtools/common/src/main/java/io/quarkus/cli/commands/file/BuildFile.java
@@ -116,7 +116,8 @@ public abstract class BuildFile implements Closeable {
 
     protected abstract List<Dependency> getManagedDependencies() throws IOException;
 
-    public abstract void completeFile(String groupId, String artifactId, String version) throws IOException;
+    public abstract void completeFile(String groupId, String artifactId, String version)
+            throws IOException;
 
     public BuildTool getBuildTool() {
         return buildTool;
--- a/devtools/common/src/main/java/io/quarkus/cli/commands/file/MavenBuildFile.java
+++ b/devtools/common/src/main/java/io/quarkus/cli/commands/file/MavenBuildFile.java
@@ -86,7 +86,8 @@ public class MavenBuildFile extends BuildFile {
     }
 
     @Override
-    public void completeFile(String groupId, String artifactId, String version) throws IOException {
+    public void completeFile(String groupId, String artifactId, String version)
+            throws IOException {
         addVersionProperty();
         addBom();
         addMainPluginConfig();
--- a/devtools/common/src/main/resources/templates/basic-rest/java/build.gradle-template.ftl
+++ b/devtools/common/src/main/resources/templates/basic-rest/java/build.gradle-template.ftl
@@ -30,3 +30,10 @@ dependencies {
 
 group '${project_groupId}'
 version '${project_version}'
+
+compileJava {
+    options.compilerArgs << '-parameters'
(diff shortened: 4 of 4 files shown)


geoand:
I think that based on what is being done for #4441 we want this to be enabled always, not just for Spring?

gsmet:
Yes, I confirm we always want that to be enabled. It causes issues in Hibernate Validator too.

geoand:
@aureamunoz can you update the PR please to always include `-parameters`?

aureamunoz:
Yes, ok, I'll do it

aureamunoz:
PR updated :-)

devtools/common/src/main/java/io/quarkus/cli/commands/file/GradleBuildFile.java
```diff
@@ -81,6 +82,14 @@ private void completeBuildContent(String groupId, String version) {
         if (!buildContent.contains("apply plugin: 'io.quarkus'") && !buildContent.contains("id 'io.quarkus'")) {
             res.append(System.lineSeparator()).append("apply plugin: 'io.quarkus'").append(System.lineSeparator());
         }
+        if (!buildContent.contains("apply plugin: 'java'")) {
+            res.append(System.lineSeparator());
+            res.append("apply plugin: 'java'").append(System.lineSeparator());
```
geoand: Are you certain we need `apply plugin: 'java'`? Does it work without it, I mean by just adding the `compileJava` Gradle task with the configuration you have supplied?

geoand: Actually I am almost certain we shouldn't do `apply plugin: 'java'` since the `java` plugin has is being applied using:

```
plugins {
    id 'java'
}
```

aureamunoz: yes, I verified also. I'm going to do the change on the code

geoand: cool

geoand:
@aureamunoz Thanks for updating. I'll test it either today or tomorrow since I don't think we plan on including it in the `0.24` release

aureamunoz:
Ok, thank you @geoand

devtools/common/src/main/java/io/quarkus/cli/commands/file/GradleBuildFile.java
```diff
@@ -81,6 +82,13 @@ private void completeBuildContent(String groupId, String version) {
         if (!buildContent.contains("apply plugin: 'io.quarkus'") && !buildContent.contains("id 'io.quarkus'")) {
             res.append(System.lineSeparator()).append("apply plugin: 'io.quarkus'").append(System.lineSeparator());
         }
+        if (!buildContent.contains("apply plugin: 'java'")) {
```
geoand: This check is not correct, since `apply plugin: 'java'` will never be there. What I think we need to have is a check for `compileJava`

geoand (changes requested):
@aureamunoz Looks good apart from the comment I made

devtools/common/src/main/java/io/quarkus/cli/commands/file/GradleBuildFile.java
aureamunoz: OK, yes, makes sense. It is done in the last commit :-)

devtools/common/src/main/java/io/quarkus/cli/commands/file/GradleBuildFile.java
```diff
@@ -81,6 +82,13 @@ private void completeBuildContent(String groupId, String version) {
         if (!buildContent.contains("apply plugin: 'io.quarkus'") && !buildContent.contains("id 'io.quarkus'")) {
             res.append(System.lineSeparator()).append("apply plugin: 'io.quarkus'").append(System.lineSeparator());
         }
+        if (!buildContent.contains("compileJava {")) {
+            res.append(System.lineSeparator());
+            res.append("compileJava {").append(System.lineSeparator());
+            res.append("    options.compilerArgs << '-parameters'").append(System.lineSeparator());
+            res.append(" }").append(System.lineSeparator());
```
geoand: No need for a space here.

geoand:
@aureamunoz I added a couple super minor comments. Once you address them we can get this in

geoand (approved):
Thanks for this @aureamunoz!

PR #4445, opened 2019-10-08
Log non-authentications related errors when using security

geoand:


extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/security/HttpSecurityRecorder.java
```diff
@@ -38,6 +42,7 @@ public void run() {
                                     }
                                 });
                             } else {
+                                log.error(throwable);
```
stuartwdouglas: This should be done a different way, I will open a PR

stuartwdouglas (changes requested):

stuartwdouglas:
superseded by https://github.com/quarkusio/quarkus/pull/4454 which just delegates to the vert.x failure handler

PR #4450, opened 2019-10-08
Removing redundant code, adding jwks_path property

sberyozkin:
I've removed some code which was marked for the deletion, and also remove the doc references to the secret properties since it is not tested right now, so I don't know if we can doc it right now. I propose to doc only those properties which are being tested. 

Thanks

### DIFF
--- a/extensions/oidc/runtime/src/main/java/io/quarkus/oidc/VertxKeycloakRecorder.java
+++ b/extensions/oidc/runtime/src/main/java/io/quarkus/oidc/VertxKeycloakRecorder.java
@@ -1,7 +1,5 @@
 package io.quarkus.oidc;
 
-import java.security.SecureRandom;
-import java.util.Base64;
 import java.util.concurrent.CompletableFuture;
 
 import io.quarkus.arc.runtime.BeanContainer;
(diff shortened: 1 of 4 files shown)


sberyozkin:
Note credentials.secret property is still obviously supported but until we have a test confirming it is supported at the Vertx level with a test, it is probably better not to type it everywhere :-) Pedro may help with a test so we'll make it visible asap

extensions/oidc/runtime/src/main/java/io/quarkus/oidc/VertxKeycloakRecorder.java
```diff
@@ -25,7 +22,14 @@ public void setup(OidcConfig config, RuntimeValue<Vertx> vertx, BeanContainer be
...
         options.setSite(config.authServerUrl);
         // RFC7662 introspection service address
-        options.setIntrospectionPath(config.introspectionPath);
+        if (config.introspectionPath.isPresent()) {
+            options.setIntrospectionPath(config.introspectionPath.get());
+        }
+
+        // RFC7662 JWKS service address
+        if (config.jwksPath.isPresent()) {
+            options.setIntrospectionPath(config.jwksPath.get());
```
stuartwdouglas: This looks like a copy/paste error?

sberyozkin: It does, thanks for catching it

sberyozkin: Funny, I was staring at the code yesterday evening for a long time :-)

sberyozkin:
@stuartwdouglas, @gsmet Not sure if it is already too late for 0.24.0, that temp code was really my target, I'll add a 0.24.0 milestone, Guillaume, please move it to 0.25.0 if it is too tight

gsmet (approved):
Looks OK to me now, let's merge.

PR #4457, opened 2019-10-09
Do better cleanup of tests if the JVM is killed

stuartwdouglas:
We write out all the created files to a file, and if this
file is still present we know that the JVM was killed and
we clean up all listed entries. Note that this can still
give problems for classes that use QuarkusTestExtension,
as JUnit 5 does not let us hook in before classloading.


stuartwdouglas:
@FroMage I think you have been complaining at various times about test cleanup in Eclipse? I think this might help.

FroMage:
OK lemme try

gsmet:
@FroMage can you check if it's OK for you? If so let's merge it and be done with it.

FroMage:
Finally ready to test.

FroMage:
I confirm it fixes my issue, thanks a lot!

FroMage (approved):

PR #4463, opened 2019-10-09
Some new Agroal datasource improvements

Sanne:
- Move validations we can do at built time, to actually run at build time
- Change the "XA / Not XA" boolean configuration property into a three-state enum:  Transactions (ENABLED, DISABLED, XA)
- Expose the new feature of Agroal to skip tracking of JDBC Statements

### DIFF
--- a/extensions/agroal/deployment/src/main/java/io/quarkus/agroal/deployment/AgroalProcessor.java
+++ b/extensions/agroal/deployment/src/main/java/io/quarkus/agroal/deployment/AgroalProcessor.java
@@ -2,6 +2,7 @@ package io.quarkus.agroal.deployment;
 
 import static io.quarkus.deployment.annotations.ExecutionTime.STATIC_INIT;
 
+import java.sql.Driver;
 import java.util.Arrays;
 import java.util.HashSet;
 import java.util.Map.Entry;
@@ -11,6 +12,8 @@ import java.util.Set;
 import javax.enterprise.context.ApplicationScoped;
 import javax.enterprise.inject.Default;
 import javax.enterprise.inject.Produces;
+import javax.enterprise.inject.spi.DeploymentException;
+import javax.sql.XADataSource;
 
 import org.jboss.jandex.AnnotationInstance;
 import org.jboss.jandex.AnnotationValue;
@@ -20,6 +23,7 @@ import org.jboss.logging.Logger;
 
 import io.agroal.api.AgroalDataSource;
 import io.quarkus.agroal.DataSource;
+import io.quarkus.agroal.TransactionIntegration;
 import io.quarkus.agroal.runtime.AbstractDataSourceProducer;
 import io.quarkus.agroal.runtime.AgroalBuildTimeConfig;
 import io.quarkus.agroal.runtime.AgroalRecorder;
@@ -96,6 +100,9 @@ class AgroalProcessor {
                 java.sql.ResultSet.class.getName(),
                 java.sql.ResultSet[].class.getName()));
 
+        validateBuildTimeConfig(null, agroalBuildTimeConfig.defaultDataSource);
+        agroalBuildTimeConfig.namedDataSources.forEach(AgroalProcessor::validateBuildTimeConfig);
+
         // Add reflection for the drivers
         if (agroalBuildTimeConfig.defaultDataSource.driver.isPresent()) {
             reflectiveClass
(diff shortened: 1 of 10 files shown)


gsmet:
This is too late for 0.24.0, targeting 0.25.0.

extensions/agroal/runtime/src/main/java/io/quarkus/agroal/runtime/DataSourceBuildTimeConfig.java
```diff
@@ -15,11 +16,11 @@
...
     /**
-     * Whether we want to use XA.
+     * Whether we want to use regular JDBC transactions, XA, or disable all transactional capabilities.
      * <p>
-     * If used, the driver has to support it.
+     * When enabling XA you will need a driver implementing {@link javax.sql.XADataSource}.
      */
-    @ConfigItem(defaultValue = "false")
-    public boolean xa;
+    @ConfigItem(defaultValue = "ENABLED")
```
gsmet: This is very minor but for the best doc generation experience, it should be `enabled` lowercase (the accepted values are listed as lowercase with hyphens in the doc).

extensions/agroal/runtime/src/main/java/io/quarkus/agroal/runtime/DataSourceBuildTimeConfig.java
```diff
@@ -15,11 +16,11 @@
...
+    public Optional<TransactionIntegration> transactions;
```
gsmet: It shouldn't be an `Optional` if it has a default value as it will never be empty.

extensions/agroal/deployment/src/main/java/io/quarkus/agroal/deployment/AgroalProcessor.java
```diff
@@ -127,6 +134,52 @@ BeanContainerListenerBuildItem build(
                 sslNativeConfig.isExplicitlyDisabled()));
     }
 
+    private static void validateBuildTimeConfig(final String datasourceName, final DataSourceBuildTimeConfig ds) {
+        //When the driver is not defined on the default datasource, we need to be more lenient as the datasource component might not be enabled at all:
+        if (!ds.driver.isPresent() && datasourceName == null) {
+            return;
+        }
+
+        String driverName = ds.driver.get();
```
gsmet: What if the driver for a **named** datasource is not set?

gsmet (changes requested):
I added a couple of comments. Most are very minor but one not so much.

extensions/agroal/deployment/src/main/java/io/quarkus/agroal/deployment/AgroalProcessor.java
Sanne: if a named datasource is defined, it implies that there are some configuration lines for it.

Having a definition for a named datasource w/o the driver will throw an error; this is to be expected as it's an incomplete configuration.

We only allow for the default datasource to not have a driver, as there might be no intention to define a datasource (we decided early on that an extension being present but unconfigured is to be treated as "dormant").

extensions/agroal/runtime/src/main/java/io/quarkus/agroal/runtime/DataSourceBuildTimeConfig.java
Sanne: thanks, didn't know that case didn't matter - expected it had to match the Enum name. weird.

Sanne:
thanks @gsmet , pushed a revised version addressing the ConfigItem changes.

extensions/agroal/deployment/src/main/java/io/quarkus/agroal/deployment/AgroalProcessor.java
gsmet: Well, an error right but not an `Optional` error.

If we think this case shouldn't happen, we should throw a proper error as the other you created.

Sanne: ?
As I said, it's not an error to not have the driver and not have any other property: that means there is no unnamed datasource defined.

Sanne: Also remember this is _build time_ . Some additional checks need to be deferred to runtime.

gsmet: The issue is that your code would throw an exception when calling `String driverName = ds.driver.get();` in the case of a named datasource.

I don't think this is acceptable. I pushed an additional commit to fix it.

extensions/agroal/deployment/src/test/java/io/quarkus/agroal/test/XaRequiresXaDatasourceTest.java
```diff
@@ -0,0 +1,31 @@
...
+            .setArchiveProducer(() -> ShrinkWrap.create(JavaArchive.class)
+                    .addAsResource("application-wrongdriverkind-datasource.properties", "application.properties"))
+            .assertException(t -> {
+                assertEquals(DeploymentException.class, t.getClass());
+            });
+
+    @Test
+    public void xaRequiresJta() {
+        //Should not be reached: verify
+        Assert.assertTrue(false);
```
yrodiere: This uses `org.wildfly.common.Assert`, probably not what you're looking for. Maybe rather `org.junit.jupiter.api.Assertions#fail(java.lang.String)`?

yrodiere (approved):
Looks good to me. I included a few comments, but as far as I can tell nothing prevents this PR from being merged.

extensions/agroal/deployment/src/main/java/io/quarkus/agroal/deployment/AgroalProcessor.java
```diff
@@ -135,8 +135,12 @@ BeanContainerListenerBuildItem build(
     }
 
     private static void validateBuildTimeConfig(final String datasourceName, final DataSourceBuildTimeConfig ds) {
-        //When the driver is not defined on the default datasource, we need to be more lenient as the datasource component might not be enabled at all:
-        if (!ds.driver.isPresent() && datasourceName == null) {
+        if (!ds.driver.isPresent()) {
+            // When the driver is not defined on the default datasource, we need to be more lenient as the datasource
+            // component might not be enabled at all so we only throw an exception for named datasources
+            if (datasourceName != null) {
+                throw new DeploymentException("Named datasource '" + datasourceName + "' doesn't have a driver defined.");
```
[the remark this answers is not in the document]
gsmet: Yeah, I used this wording as this check is specific to named datasources. Not sure if it makes sense?

extensions/agroal/deployment/src/test/java/io/quarkus/agroal/test/DisabledTransactionDataSourceConfigTest.java
```diff
@@ -0,0 +1,44 @@
...
+    static final QuarkusUnitTest config = new QuarkusUnitTest().setArchiveProducer(
+            () -> ShrinkWrap.create(JavaArchive.class)
+                    .addAsResource("application-disabledjta-datasource.properties",
+                            "application.properties"));
+
+    @Test
+    public void testNonTransactionalDataSourceInjection() throws SQLException {
+        AgroalConnectionPoolConfiguration configuration = defaultDataSource.getConfiguration().connectionPoolConfiguration();
+
+        assertTrue(!(configuration.transactionIntegration() instanceof NarayanaTransactionIntegration));
```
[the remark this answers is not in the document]
gsmet: Yeah, changed it.

extensions/agroal/deployment/src/main/java/io/quarkus/agroal/deployment/AgroalProcessor.java
Sanne: ah, thanks!  I finally understand what you mean.

extensions/agroal/deployment/src/test/java/io/quarkus/agroal/test/XaRequiresXaDatasourceTest.java
Sanne: Interesting. Got that as I copied another test as starting point.. a quick grep reveals that the WildFly assert is being used a lot all over :/

gsmet: OK, let's fix that in another PR, I need this one to go through CI or I won't be able to backport it.

Sanne: Sure, that's unrelated. I was just wondering how this could happen :)

gsmet (approved):

gsmet:
Merged. I backported it. CI is running on the 0.24 branch right now.

PR #4465, opened 2019-10-09
Fix issue with RESTEasy GET requests when using standalone vert.x

stuartwdouglas:


geoand:
Closing in favor of #4466 which has the same fix commit but also adds a test

PR #4466, opened 2019-10-09
Fix issue with RESTEasy GET requests when using standalone vert.x 

geoand:
Supersedes: #4465


gsmet (approved):
Apart that the test doesn't trigger the issue so we will need to tweak it. Let's do it later so that we can merge this.

geoand:
The new test that was added now does fail without the fix in the first commit so it should able of catching similar regressions in the future

gsmet:
Perfect, let's wait for CI.

PR #4468, opened 2019-10-09
Arc - AroundInvoke interceptors should now use different implementati…

### LINKED ISSUE #4467: Change how Arc handles around invoke interception
Original problem description is here, text below is a shameless copy and paste - https://groups.google.com/forum/#!topic/quarkus-dev/POdMKKe24ts

tl;dr:
MP FT `@Asynchronous` relies on unspecified behavior of
javax.interceptor.InvocationContext (Interceptors spec). Quarkus impl of
InvocationContext does not allow an interception to continue on a
different thread after the intercepted method returns.

Q: What's the problem?
A: An InvocationContext impl must be able to recover when an exception
is thrown [1]. And the same IC instance must be passed to each
interceptor in a chain [2]. To implement [1] Quarkus impl has some state
that is modified whenever a interceptor invocation completes [3]. This
means that when an intercepted method returns before an MP FT
`@Asynchronous` executes the interception chain is restarted. And may even
result in stack overflow ([4] not the website ;-).

Q: Why does it work on Thorntail/Wdilfly?
A: Weld has a special implementation of InvocationContext for
`@AroundInvoke` that allows for similar use cases. However, the impl
violates the interceptors spec [2] because each interceptor receives a
new InvocationContext instance that is effectively immutable. On the
other hand, the TCK does not test it ;-). A disadvantage is performance
overhead - each invocation results in few more object allocations.

Q: Could we implement the same for Quarkus?
A: Yes, if we're ok with spec violation and increased perf overhead.

[1]
"Interceptor methods are allowed to catch and suppress exceptions and to
recover by calling the InvocationContext.proceed method."

[2]
"The same InvocationContext instance will be passed to each interceptor
method for a given target class method or lifecycle event interception."

[3]
https://github.com/quarkusio/quarkus/blob/master/independent-projects/arc/runtime/src/main/java/io/quarkus/arc/InvocationContextImpl.java#L130-L137

[4]
https://en.wikipedia.org/wiki/Stack_overflow

manovotn:
…on allowing for async interception execution.

Fixes #4467

### DIFF
--- a/extensions/arc/runtime/src/main/java/io/quarkus/arc/runtime/InterceptorBindings.java
+++ b/extensions/arc/runtime/src/main/java/io/quarkus/arc/runtime/InterceptorBindings.java
@@ -5,6 +5,7 @@ import java.util.Set;
 
 import javax.interceptor.InvocationContext;
 
+import io.quarkus.arc.AroundInvokeInvocationContext;
 import io.quarkus.arc.InvocationContextImpl;
 
 public class InterceptorBindings {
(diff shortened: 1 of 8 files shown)


kenfinnigan (approved):

kenfinnigan:
@gsmet so we shouldn't be merging this to master yet? Is that right?

gsmet:
I think I will branch 0.24 now as I don't know when I'll be able to push it.

gsmet:
@kenfinnigan feel free to merge from now on.

Issue #2702, opened 2019-06-04
Running Quarkus JVM app on OpenShift requires root access (SCC)


jfwork:
Running Quarkus app in `jvm` mode on OpenShift requires `root` acces (configuring SCC).
App uses Narayana JTS.

```
2019-06-04 09:04:44,385 [-] WARN  [com.arj.ats.common] (main) ARJUNA048006: cannot create new instance of com.arjuna.ats.internal.arjuna.objectstore.ShadowNoFileLockStore: java.lang.reflect.InvocationTargetException
	at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
	at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.lang.reflect.Constructor.newInstance(Constructor.java:423)
	at com.arjuna.common.internal.util.ClassloadingUtility.loadAndInstantiateClass(ClassloadingUtility.java:129)
	at com.arjuna.ats.arjuna.objectstore.StoreManager.initStore(StoreManager.java:152)
	at com.arjuna.ats.arjuna.objectstore.StoreManager.getActionStore(StoreManager.java:111)
	at com.arjuna.ats.arjuna.objectstore.StoreManager.getRecoveryStore(StoreManager.java:68)
	at com.arjuna.ats.arjuna.recovery.ActionStatusService.<init>(ActionStatusService.java:65)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
	at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.lang.reflect.Constructor.newInstance(Constructor.java:423)
	at java.lang.Class.newInstance(Class.java:442)
	at com.arjuna.common.internal.util.ClassloadingUtility.loadAndInstantiateClass(ClassloadingUtility.java:135)
	at com.arjuna.ats.arjuna.recovery.TransactionStatusManager.start(TransactionStatusManager.java:125)
	at com.arjuna.ats.arjuna.recovery.TransactionStatusManager.<init>(TransactionStatusManager.java:58)
[...]

PR #4470, opened 2019-10-09
run with quarkus user (read-write access)

haraldatbmw:
In OpenShift it is not allowed to run the image with the root user. Therefor OpenShift creates a read-only user. The JTA implementation used by Quarkus needs write access (see #2702). So an exception is thrown if running without a custom user.

To overcome this problem I created a quarkus user with read and write access to the /deployments directory.


geoand:
Thanks for this @haraldatbmw! 

It looks good to me, @cescoffier WDYT?

cescoffier:
2 questions:

* what about using 1001 as user?
* should we do the same for the native docker file?

aloubyansky (approved):

aloubyansky:
@cescoffier do we want to clarify whether the native image one needs the same before merging this or check later?

Issue #4406, opened 2019-10-07
Dynamodb: Add support for Jdk buitin HttpClient - UrlConnectionHttpClient


Ravisankar-Challa:
Hi Team,

Please also add support for UrlConnectionHttpClient as http client for Dynamo DB as an alternative to ApacheHttpClient.

Thanks,
Ravi

gsmet:
Do you have any good reason of not using the Apache HTTP client?

(Real question, I'm not familiar with Amazon DynamoDB)

/cc @marcinczeczko

Ravisankar-Challa:
Not suited for AWS LAMDA

gsmet:
@marcinczeczko I let you reply on this one and see if we can do something about it?

marcinczeczko:
Well, honesty there were no a good reason for that, except the counter-example for async client that was my primary goal :) It's fairly easy to add such support as AWS SDK supports it.

marcinczeczko:
Please see #4471

Ravisankar-Challa:
@marcinczeczko Tested it with 999-SNAPSHOT in aws lamda with **quarkus.dynamodb.sync-client.type=URL**. Workd fine. How to get this property in vscode auto complete feature. Do I have to raise a separate feature request for this?

gsmet:
@marcinczeczko 's work has been merged.

PR #4471, opened 2019-10-09
Dynamodb - URL connection http client support

marcinczeczko:
Implementation for #4406


extensions/amazon-dynamodb/runtime/src/main/java/io/quarkus/dynamodb/runtime/SyncHttpClientConfig.java
```diff
@@ -0,0 +1,211 @@
...
+import java.util.Optional;
+
+import io.quarkus.runtime.annotations.ConfigGroup;
+import io.quarkus.runtime.annotations.ConfigItem;
+
+@ConfigGroup
+public class SyncHttpClientConfig {
+
+    /**
+     * Type of the sync HTTP client implementation
```
cescoffier: You should mention the accepted values here (APACHE, URL)

extensions/amazon-dynamodb/runtime/src/main/java/io/quarkus/dynamodb/runtime/SyncHttpClientConfig.java
```diff
@@ -0,0 +1,211 @@
...
+    public Duration socketTimeout;
+
+    /**
+     * Apache HTTP client additional configuration
+     */
+    @ConfigItem(name = ConfigItem.PARENT)
+    public ApacheHttpClientConfig apacheHttp;
+
+    @ConfigGroup
+    public static class ApacheHttpClientConfig {
```
cescoffier: If you don't use the APACHE client, how do you configure all these details? Proxy, TLS?

cescoffier (dismissed):

extensions/amazon-dynamodb/runtime/src/main/java/io/quarkus/dynamodb/runtime/SyncHttpClientConfig.java
marcinczeczko: There are no configuration options for URL connection other than socket timeout and connection timeout. But it's fine, as the lightweight url conn as sync client is fine for lambdas. For other app models like container talking to the dynamodb the apache sync client can be used that enables configurations for more connectivity aspects

extensions/amazon-dynamodb/runtime/src/main/java/io/quarkus/dynamodb/runtime/SyncHttpClientConfig.java
marcinczeczko: Added

gsmet:
Just a heads up that I plan to review this one.

extensions/amazon-dynamodb/runtime/src/main/java/io/quarkus/dynamodb/runtime/SyncHttpClientConfig.java
```diff
@@ -0,0 +1,214 @@
...
+import io.quarkus.runtime.annotations.ConfigItem;
+
+@ConfigGroup
+public class SyncHttpClientConfig {
+
+    /**
+     * Type of the sync HTTP client implementation
+     *
+     * <p>
+     * Accepted values are `URL` or `APACHE`
```
gsmet: In fact, you shouldn't :). They will be automatically added to the documentation as it's an enum. So you can get rid of it.

extensions/amazon-dynamodb/runtime/src/main/java/io/quarkus/dynamodb/runtime/SyncHttpClientConfig.java
```diff
@@ -0,0 +1,214 @@
...
+     */
+    @ConfigItem(defaultValue = "url")
+    public SyncClientType type;
+
+    /**
+     * The maximum amount of time to establish a connection before timing out.
+     *
+     * <p>
+     * Used by `URL` and `APACHE` clients
```
gsmet: I think I would drop all these (the last line of documentation).

extensions/amazon-dynamodb/runtime/src/main/java/io/quarkus/dynamodb/runtime/SyncHttpClientConfig.java
```diff
@@ -0,0 +1,214 @@
...
+     */
+    @ConfigItem(defaultValue = "30S")
+    public Duration socketTimeout;
+
+    /**
+     * Apache HTTP client additional configuration
+     */
+    @ConfigItem(name = ConfigItem.PARENT)
+    public ApacheHttpClientConfig apacheHttp;
```
gsmet: Wondering if it shouldn't be in an `apache` group instead of added to the parent? This way, it's clear it's Apache specific and you don't need any additional documentation.

extensions/amazon-dynamodb/runtime/src/main/java/io/quarkus/dynamodb/runtime/SyncHttpClientConfig.java
```diff
@@ -0,0 +1,214 @@
...
+
+    @ConfigGroup
+    public static class ApacheHttpClientConfig {
+
+        /**
+         * The amount of time to wait when acquiring a connection from the pool before giving up and timing out.
+         * 
+         * <p>
+         * Used by `APACHE` client only
```
gsmet: Then you could drop this. I will tweak the documentation with some sections and it will be pretty much OK I think.

extensions/amazon-dynamodb/deployment/src/test/resources/sync-apache-broken-config.properties
```diff
@@ -1,3 +1,4 @@
+quarkus.dynamodb.sync-client.type = APACHE
```
gsmet: So for enums, the recommended way is to use lower case, dashed versions in the `application.properties`. They will be automatically converted. Typically the documentation will contain `apache` and `url` so let's be consistent.

bom/runtime/pom.xml
```diff
@@ -2177,6 +2177,11 @@
                 <artifactId>apache-client</artifactId>
                 <version>${awssdk.version}</version>
             </dependency>
+            <dependency>
+                <groupId>software.amazon.awssdk</groupId>
+                <artifactId>url-connection-client</artifactId>
+                <version>${awssdk.version}</version>
+            </dependency>
```
gsmet: What annoys me a bit is that you get all the dependencies with the extension. I wonder if it's a wise choice and we shouldn't be more careful about that especially in the lambda use case.

That being said, I'm not totally sure we can find a nice way to do that properly but IMHO it's worth a try.

If we can get to the point of the dependencies being optional and we can throw a warning or an error if something is configured for a "backend" and the correct dependencies are not in the classpath, that would be OK IMHO.

I don't think it's blocking for this PR tbh so I think we could merge this and see how we can improve things later.

gsmet (changes requested):
I suggested a few things. Open to discussion so don't hesitate to push back :).

extensions/amazon-dynamodb/runtime/src/main/java/io/quarkus/dynamodb/runtime/SyncHttpClientConfig.java
marcinczeczko: Thanks, I forgot about that :) Will fix.

extensions/amazon-dynamodb/runtime/src/main/java/io/quarkus/dynamodb/runtime/SyncHttpClientConfig.java
marcinczeczko: Will drop

extensions/amazon-dynamodb/runtime/src/main/java/io/quarkus/dynamodb/runtime/SyncHttpClientConfig.java
marcinczeczko: will drop it

extensions/amazon-dynamodb/runtime/src/main/java/io/quarkus/dynamodb/runtime/SyncHttpClientConfig.java
marcinczeczko: I think it make sense. However, since I wanted to avoid duplications, the connectionTimeout & socketTimeout that are on the parent and apply to URL connection are also applies to Apache. Wondering if it worth to reflect that on the documentation somehow..

extensions/amazon-dynamodb/deployment/src/test/resources/sync-apache-broken-config.properties
marcinczeczko: will clean it up

bom/runtime/pom.xml
marcinczeczko: I like the idea and in fact tried it, check the latest commits. In general:
1. I made client deps as optional
2. the deployment processor checks if required service exists on the classpath, if not throws deploymentException
3. the application should explicitly add required dependency to the classpath

Do you think it make sense ? Once you verify the idea, I'll update docs & quickstart

marcinczeczko (commented):
Please find the latest changes

gsmet:
@marcinczeczko looks good to me. Can you rebase, squash everything and work on the doc/quickstart update?

I'll move that to 0.27.0 for now as I'm freezing the 0.26.0 code very soon.

marcinczeczko:
> @marcinczeczko looks good to me. Can you rebase, squash everything and work on the doc/quickstart update?
> 
> I'll move that to 0.27.0 for now as I'm freezing the 0.26.0 code very soon.

Thanks. I updated guide & quickstart (https://github.com/quarkusio/quarkus-quickstarts/pull/328)

gsmet:
I force-pushed a small change to fix a doc issue and rebased, I'll merge this one as soon as CI is green.

gsmet:
CI failures is unrelated.

gsmet (approved):

marcinczeczko:
@gsmet - thanks !. By the way, I was working on another iteration of quarkified AWS SDK that spans across more AWS clients not only DynamoDB- #4968. It's WIP at the moment, but I'd like to validate if this is something you might be interested so I can continue on this. Thanks

PR #4476, opened 2019-10-09
Arc - use different package for client proxy generation if the bean comes from default package

### LINKED ISSUE #4469: CNFE on scope defining annotations in the default package
**Describe the bug**
Quarkus fails ungracefully with scope-defining annotations on beans in the default package.
The error: https://gist.github.com/michalszynkiewicz/c489af90ed958b4ff43c7236f4a1b09a

**Expected behavior**
Either just work or produce a user-friendly message.

**To Reproduce**
1. clone https://github.com/michalszynkiewicz/quarkus-krakow/tree/failing-scope-on-default/ads
2. `mvn clean compile quarkus:dev` or `mvn clean package && java -jar target/*-runner.jar`

CC @manovotn

manovotn:
…omes from default package.

Fixes #4469 

Tried with Michal's reproducer, otherwise not sure how to write an automated test without bloating the test structure with default package classes.

### DIFF
--- a/independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/AbstractGenerator.java
+++ b/independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/AbstractGenerator.java
@@ -81,8 +81,8 @@ abstract class AbstractGenerator {
             }
         }
         String packageName = DotNames.packageName(providerTypeName);
-        if (packageName.startsWith("java.")) {
-            // It is not possible to place a class in a JDK package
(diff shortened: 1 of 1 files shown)


manovotn:
Theoretically we could also log an `INFO` about defualt package being a poor practice but depending on what scope user has, that logging might get intrusive. Or we would need to do that during discovery.

kenfinnigan (approved):

mkouba:
For the record - package-private injection points and interception of package-private methods would not work because we only use reflection fallback for private members.

PR #4478, opened 2019-10-09
Upgrade Hibernate Search to 6.0.0.Beta1

yrodiere:
@gsmet 

I also added a small change to how the predicate is built, following our discussion the other day. Feel free to remove that change if you think it's too complex.

Also, as discussed over Zulip, I had to add a commit to workaround https://hibernate.atlassian.net/browse/HSEARCH-3734 . Without this commit, native mode just won't work.


docs/src/main/asciidoc/hibernate-search-guide.adoc
```diff
@@ -474,17 +475,20 @@ In our existing `LibraryResource`, we just need to inject the following methods
...
-                .predicate(f ->
-                    pattern == null || pattern.trim().isEmpty() ?
-                            f.matchAll() : // <6>
-                            f.simpleQueryString()
-                                .onFields("firstName", "lastName", "books.title").matching(pattern) // <7>
-                )
-                .sort(f -> f.byField("lastName_sort").then().byField("firstName_sort")) // <8>
-                .fetchHits(); // <9>
+                .predicate(f -> f.bool(b -> {
+                    b.must(f.matchAll()); // <6>
```
gsmet: Wouldn't it be easier to not have a `bool` but only add one or the other `must` depending on the condition?

gsmet (changes requested):
Looks good apart from one question about one of the examples.

docs/src/main/asciidoc/hibernate-search-guide.adoc
yrodiere: You mean the previous code? I think I'm missing something here.

The trick here is that the `matchAll` will only be taken into account if there is no other `must`. So if you need to add *other* predicates, you just need to append other `if` blocks after the current one. If we used `if ( ... )  { matchAll } else { match().field( ... )... }` and wanted to add another predicate, the `if/else` nesting would get more complex.

I.e. with the current code, adding other predicates would lead to this:

```
                .predicate(f -> f.bool(b -> {
                    b.matchAllIfEmpty(); // <6>
                    if (pattern != null && !pattern.trim().isEmpty()) {
                        b.must(f.simpleQueryString()
                                .fields("firstName", "lastName", "books.title")
                                .matching(pattern)); // <7>
                    }
                    if (somethingElse) {
                        b.must(f.somethingElse());
                    }
                    if (yetSomethingElse) {
                        b.must(f.yetSomethingElse());
                    }
                })) 
```

Maybe I should add `matchAllIfEmpty`/`matchNoneIfEmpty` methods to the boolean predicate builder? The "trick" would then be more obvious.
I'm also considering whether we should add an overload to `predicate()` that takes a bi-function, to make the syntax easier on the eye. In the end we'd get something like that (in Beta2, that is):

```
                .predicate((f, b) -> { // maybe rather "predicates"?
                    b.matchAllIfEmpty(); // <6>
                    if (pattern != null && !pattern.trim().isEmpty()) {
                        b.must(f.simpleQueryString()
                                .fields("firstName", "lastName", "books.title")
                                .matching(pattern)); // <7>
                    }
                })
```

But currently, the original code and the one I replaced it with are our only solutions, I think.

yrodiere: As discussed, I removed this change.

gsmet (approved):

gsmet:
Merged, thanks. I'm not entirely convinced by the config changes but you're the man.

PR #4485, opened 2019-10-09
Remove redundant patterns compilations

gwenneg:
While looking for the root cause of a bug involving regex, I noticed a few redundant patterns compilations during `String` splits. All modified splits in this PR are called from an enclosing loop.

### DIFF
--- a/extensions/mongodb-client/runtime/src/main/java/io/quarkus/mongodb/runtime/MongoClientRecorder.java
+++ b/extensions/mongodb-client/runtime/src/main/java/io/quarkus/mongodb/runtime/MongoClientRecorder.java
@@ -11,6 +11,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.TimeUnit;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
 import org.bson.codecs.configuration.CodecProvider;
@@ -43,6 +44,7 @@ import io.quarkus.runtime.annotations.Recorder;
 public class MongoClientRecorder {
 
     private static final Logger LOGGER = Logger.getLogger(MongoClientRecorder.class);
+    private static final Pattern COLON_PATTERN = Pattern.compile(":");
 
     private static volatile MongoClient client;
     private static volatile ReactiveMongoClient reactiveMongoClient;
@@ -251,7 +253,7 @@ public class MongoClientRecorder {
         return addresses.stream()
                 .map(String::trim)
                 .map(address -> {
-                    String[] segments = address.split(":");
+                    String[] segments = COLON_PATTERN.split(address);
                     if (segments.length == 1) {
                         // Host only, default port
                         return new ServerAddress(address);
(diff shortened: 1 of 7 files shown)


gwenneg:
@cescoffier I also noticed there's a lot of duplicated code between `VertxCoreRecorder` and `VertxOptionsBuilder`, and I can't find any call to `VertxOptionsBuilder.buildVertxOptions()` in Quarkus. I was about to change that and then I saw `VertxOptionsBuilder` was added recently. Is there something in progress there or can I clean the duplicated code?

extensions/mongodb-client/runtime/src/main/java/io/quarkus/mongodb/runtime/MongoClientRecorder.java
```diff
@@ -43,6 +44,7 @@
 public class MongoClientRecorder {
 
     private static final Logger LOGGER = Logger.getLogger(MongoClientRecorder.class);
+    private static Pattern COLON_PATTERN = Pattern.compile(":");
```
gsmet: `final`?

extensions/resteasy/runtime/src/main/java/io/quarkus/resteasy/runtime/standalone/VertxUtil.java
```diff
@@ -23,6 +24,9 @@
  * @version $Revision: 1 $
  */
 public class VertxUtil {
+
+    private static Pattern COMMA_PATTERN = Pattern.compile(",");
```
gsmet: `final`?

extensions/vertx-core/runtime/src/main/java/io/quarkus/vertx/core/runtime/VertxCoreRecorder.java
```diff
@@ -39,6 +39,8 @@
 @Recorder
 public class VertxCoreRecorder {
 
+    private static Pattern COMMA_PATTERN = Pattern.compile(",");
```
gsmet: `final`?

extensions/vertx-core/runtime/src/main/java/io/quarkus/vertx/core/runtime/VertxOptionsBuilder.java
```diff
@@ -22,6 +22,8 @@
 
 public class VertxOptionsBuilder {
 
+    private static Pattern COMMA_PATTERN = Pattern.compile(",");
```
gsmet: `final`?

gsmet (changes requested):
Looks good thanks, except some ``final``s are missing. Could you take care of it?

gwenneg:
Oops, that's fixed! Thanks @gsmet.

gsmet (approved):

gsmet:
I let @cescoffier give you an answer to your question. I think it's probably something we missed when we reorganized the Vert.x extensions.

cescoffier:
@gwenneg yes we need to do some cleanup. The split has introduced some duplications.

PR #4487, opened 2019-10-10
Initial HTTP permission layer implementation

stuartwdouglas:
This allows for HTTP based permissions to be defined
in application.properties, and for pluggable permission
checkers to be defined.


extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/security/HttpPermissionChecker.java
```diff
@@ -0,0 +1,43 @@
...
+
+public interface HttpPermissionChecker {
+
+    /**
+     * The priority of the permission checker. Permission checkers are evaluated from highest to lowest
+     * priority.
+     *
+     * @return The priority
+     */
+    int getPriority();
```
sberyozkin: @stuartwdouglas I'm not sure this is required, all the registered checkers have to give their approval, the order in which they do it is likely to be insignificant

stuartwdouglas: With the current implementation they don't. They can return PERMIT which grants access, or IGNORE, which will grant access if subsequent checkers don't deny (and default-deny is false)

sberyozkin: May be worth having a default implementation of this method ? I'm not sure one can reliably predict the ordering by returning `100` or something like that :-)

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/PermissionSetConfig.java
```diff
@@ -0,0 +1,64 @@
+package io.quarkus.vertx.http.runtime;
+
+import java.util.List;
+
+import io.quarkus.runtime.annotations.ConfigGroup;
+import io.quarkus.runtime.annotations.ConfigItem;
+
+@ConfigGroup
+public class PermissionSetConfig {
```
sberyozkin: Can that be renamed to `RolePermissionSetConfig` or `DefaultPermissonSetConfig`. If say an OIDC adapter would deny an access because the client ip is blacklisted then it won't about the roles ?

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/AuthConfig.java
```diff
@@ -1,14 +1,16 @@
...
-import io.quarkus.runtime.annotations.ConfigRoot;
 
 /**
  * Authentication mechanism and SecurityRealm name information used for configuring HTTP auth
  * instance for the deployment.
  */
-@ConfigRoot
-public class HttpAuthConfig {
+@ConfigGroup
+public class AuthConfig {
```
sberyozkin: Should it be named `DefaultAuthConfig`

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/security/DefaultHttpPermissionChecker.java
```diff
@@ -0,0 +1,114 @@
...
+import io.quarkus.security.identity.SecurityIdentity;
+import io.quarkus.vertx.http.runtime.HttpBuildTimeConfig;
+import io.quarkus.vertx.http.runtime.PermissionSetConfig;
+import io.vertx.core.http.HttpServerRequest;
+
+/**
+ * permission checker that handles permissions defined in application.properties
+ */
+@ApplicationScoped
+public class DefaultHttpPermissionChecker implements HttpPermissionChecker {
```
sberyozkin: Would it make sense to have a property, for example in the `(Default)AuthConfig` to have this checker disabled when some more specific checker is only required, say some claim based OIDC checker ?

sberyozkin:
Stuart, thinking more about this priorities...
I wonder would it really make sense to change `HttpPermissionChecker` a bit and have text-based checker groupings, by categories, `HTTP` or `Kafka` may be, etc.
So we'd have:
```
public interface PermissionChecker {

    CompletionStage<CheckResult> checkPermission(SecurityIdentity identity);

}
```
and then, for HTTP:
```
@RequestScoped
@AppliesTo(Permissions.HTTP)
public class DefaultHttpPermissionChecker implements PermissionChecker {
  @Inject
  HttpServerRequest httpRequest;

  CompletionStage<CheckResult> checkPermission(SecurityIdentity identity) {
  }
  
}
```

I'm not entirely sure it is a sound idea but wanted to check your opinion

pedroigor:
As a HTTP-specific implementation, it looks good for me.

Would it be possible to extend somehow the permission config so that we can parametrize custom permission checkers? I'm thinking about the policy enforcer functionality that we have in Keycloak which provides an additional configuration.

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/security/HttpPermissionChecker.java
stuartwdouglas: No, in fact I should throw an error if we have multiple checkers with the same priority. The main thing that we need is deterministic ordering, if multiple checkers have the same priority then it could result in non-deterministic behaviour.

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/PermissionSetConfig.java
stuartwdouglas: This implementation is all about roles. The OIDC adaptor will likely not use these classes.

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/security/DefaultHttpPermissionChecker.java
stuartwdouglas: This checker is enabled if role based permission checks have been added. If another checker is added as well then both will be used.

stuartwdouglas:
This is HTTP specific and I don't really want to generalize it. 

Something to bear in mind is that we also have SecurityIdentity.checkPermission that can also be used for some permission checks, however it does not really work in this case.

stuartwdouglas:
@pedroigor at the moment our config does not really allow for easy parameterisation, other that something like a generic Map<String, String> type setup. 

Maybe we could add something like 'permission-ref' as a third option (as well as permit, deny, roles-allowed). This would be a string that referenced a custom permission check, that could be defined in the OIDC config?

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/security/HttpPermissionChecker.java
sberyozkin: Hi Stuart, it feels things are being over complicated. How can multiple modules can make it deterministic ? Each module has its own expectations what it needs to check. For example, with the Config converters it is a different story, but here, 

If I have to ship my custom checker I need to make sure it is invoked, so I need to be aware what are the actual numbers the other modules are bringing in, otherwise if I just put random 100 then there is no guarantee my checker will even be invoked.

IMHO all the checkers have to be invoked as opposed to building some relationship between them, it is either PERMIT-ALL or 403.
Can you think of the practical case where a PERMIT-ALL mode can be too lax or too restrictive, etc ?
Thanks

sberyozkin:
@stuartwdouglas OK, I'm happy enough to drop the generalization idea :-)
I don't want to block this PR because it is critical, IMHO the priorities idea does not work well, I'll let to decide if you want to keep it or not. Thanks

sberyozkin (approved):

pedroigor:
> @pedroigor at the moment our config does not really allow for easy parameterisation, other that something like a generic Map<String, String> type setup.
> 
> Maybe we could add something like 'permission-ref' as a fouth option (as well as permit, deny, roles-allowed). This would be a string that referenced a custom permission check, that could be defined in the OIDC config?

The `permission-ref` sounds good. But I guess you did not mean tying it to a reference to OIDC config, right?

Basically, I need to be able to obtain this config from a custom `HttpPermissionChecker`:

```json
# Global Config
?.policy-enforcer.enable=true
?.policy-enforcer.enforcement-mode=PERMISSIVE
?.policy-enforcer.lazy-load-paths=true
?.policy-enforcer.claim-information-point.claims.global-claim=global-claim

# Path-Based Config
?.policy-enforcer.paths.1.claim-information-point.claims.claim-a=claim-a
?.policy-enforcer.paths.2.claim-information-point.http.claims.user-name=/userName
?.policy-enforcer.paths.2.claim-information-point.http.url=http://localhost:8081/api/users/me
?.policy-enforcer.paths.2.claim-information-point.http.method=GET
?.policy-enforcer.paths.2.claim-information-point.http.headers.Content-Type=application/x-www-form-urlencoded
?.policy-enforcer.paths.2.claim-information-point.http.headers.Authorization=Bearer {keycloak.access_token}
```

Regarding `Path-Based Config` maybe we could also be able to reference a path from the `http` config (policy-enforcer.paths.1) ? So that I can offer additional capabilities for a given path.

dmlloyd:
This is just a thought, but I was thinking there might be a point where it's better for the user to specify things using a builder of some sort (like the router API) rather than having super complex configurations.

sberyozkin:
One possible issue here is that by improving the option for this KC specific configuration we can make it more difficult to eventually migrate users to a generic adapter, especially of they will have something in their code :-). JSON format or a custom builder would be nice, but in this case we are dealing with the users of the original Keycloak adapter who can't just move to the one yet and already have the configuration prepared. 
For UMA we can do something really nice with what David and Pedro suggests, perhaps we can prioritize on UMA soon enough for the new users be starting with it. Especially given that `quarkus-oidc` is expected to work not only with Keycloak

sberyozkin:
@pedroigor Hi Pedro, by the way, FYI, I've approved this PR but haven't merged yet as I was hoping to try to convince Stuart to remove the priorities, but we can merge immediately to avoid blocking you, just let us know

sberyozkin:
@pedroigor You can probably have something like this, have `.runtime.PolicyEnforcer` bean and get it initialized in a new extemsion's deployment processor, and pass it to the runtime recoder where you'd set  this `PolicyEnforcer` instance on the CDI producer which produces a custom `HttpPermissionChecker'

sberyozkin:
Going to merge, Pedro, if I can help somehow ping me please, lets chat more at #4495

pedroigor:
Sure. I'm going to consider your proposal too.

PR #4492, opened 2019-10-10
Properly fix resource path in the OIDC guide

gsmet:
@geoand thanks for catching the issue!


geoand (approved):
I should have caught it the first time around :)

PR #4496, opened 2019-10-10
Remove High contention on TransactionManager class

johnaohara:
… causes high contention on TransactionManager class.

The current Narayana integration frequently calls `TransactionalInterceptorBase.invokeInOurTx()`, which looks up the current configured TX timeout via 
`((TransactionManagerImple) com.arjuna.ats.jta.TransactionManager.transactionManager()).getTimeout()`

This is a very expensive call, as `com.arjuna.ats.jta.TransactionManager.transactionManager()` is synchronized on the `com.arjuna.ats.jta.TransactionManager` class.

We *could* use the @ApplicationScoped TransationManager injected into io.quarkus.narayana.jta.runtime.interceptor.TransactionalInterceptorBase if the API defined `getTimeout()`, but it does not.  At present `@inject TransationManager transactionManager` can not be cast to `TransactionManagerImple` as it is a proxy.

I am not 100% happy with this PR, as it changes the injected bean to an impl bean instead of an api bean, but it does remove the monitor contention under heavy load.

@stuartwdouglas @mkouba can you think of any other way of resolving this issue?

### DIFF
--- a/extensions/narayana-jta/runtime/src/main/java/io/quarkus/narayana/jta/runtime/NarayanaJtaProducers.java
+++ b/extensions/narayana-jta/runtime/src/main/java/io/quarkus/narayana/jta/runtime/NarayanaJtaProducers.java
@@ -3,6 +3,7 @@ package io.quarkus.narayana.jta.runtime;
 import javax.enterprise.context.ApplicationScoped;
 import javax.enterprise.context.Dependent;
 import javax.enterprise.inject.Produces;
+import javax.inject.Singleton;
 import javax.transaction.TransactionSynchronizationRegistry;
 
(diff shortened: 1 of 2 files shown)


johnaohara:
Have changed producer from `@ApplicationScoped' to '@Singleton' so proxy is not generated by arc, and we cast TransactionManager to TransactionManagerImple successfully.

gsmet:
Removing the backport label as 0.24.0 has been tagged this morning.

stuartwdouglas (approved):

stuartwdouglas:
@mmusgrov do you happen to know why this method is synchronized? Looking at the code it does not look like it needs to be.

PR #4498, opened 2019-10-10
fix-3592: remove file schema from uri before loading it

dufoli:
tested on my side and fix the issue for me.

### DIFF
--- a/core/devmode/src/main/java/io/quarkus/dev/ClassLoaderCompiler.java
+++ b/core/devmode/src/main/java/io/quarkus/dev/ClassLoaderCompiler.java
@@ -104,6 +104,10 @@ public class ClassLoaderCompiler {
                                 for (String i : classPath.toString().split(" ")) {
                                     File f;
                                     try {
+                                        if (i.startsWith("file:")) {
+                                            i = i.substring(5, i.lastIndexOf('!'));
+                                        }
+
                                         f = Paths.get(new URI("file", null, "/", null).resolve(new URI(i))).toFile();
                                     } catch (URISyntaxException e) {
                                         f = new File(file.getParentFile(), i);


gsmet:
@dmlloyd does it make sense to you?

gsmet:
@dufoli what value do you have exactly?

I'm also concerned that we might not have a `!` every time and the code is not entirely safe.

dmlloyd:
Hmm, something is fishy.  Dropping `file:` makes the URI relative (even if it starts with `/`), which forces it to be resolved against... `file:` on the next line, which should result in the same URI being given back again!

I'd like to see a test, or example, or some evidence of what this change actually does.

core/devmode/src/main/java/io/quarkus/dev/ClassLoaderCompiler.java
```diff
@@ -104,6 +104,10 @@ public ClassLoaderCompiler(ClassLoader classLoader,
                                 for (String i : classPath.toString().split(" ")) {
                                     File f;
                                     try {
+                                        if (i.startsWith("file:")) {
+                                            i = i.substring(5, i.lastIndexOf('!'));
```
jaikiran: Hello @dufoli , won't this cause a problem is the `!` character isn't part of the `String` `i`, in which case `lastIndexOf` will return -1?

dufoli:
meaculpa, it do not work at all.. I was failing on another error and because build is not stable, I was thinking this issue is fixed but no. Fix produce other error.
Can someonce close the PR? because I do not find how to on my side?

What I have was a stacktrace with ":" not allowed after file. So I was guessing that path must not been uri but regular local path. So I try to remove the file part.
I try few thing to remove file but never succeed to get absolute path working.
file:\C:\.... do not worked.
The odd thing to me is that we have file:\C:\... and not file:\\C:\... 

Anyway, I try with file:\\C:\, \C:\..., C:\,... it never worked.
I have an even before assert because the jar file is not found in temp forlder...

jaikiran:
I have been able to narrow this down. A lengthy explanation will follow in the original issue. I think we'll need @dmlloyd's help to fix this properly.

jaikiran:
https://github.com/quarkusio/quarkus/pull/4527 which has been merged, should take care of this.

gsmet:
Yup, thanks for noticing, I will close this one.

PR #4502, opened 2019-10-10
Externalize hotrod-client.properties with infinispan-client

mswiderski:
@wburns could you please review?

I went ahead with config properties to avoid loading file on startup. Focused on properties from hotrod.properties that relates to authentication. We can extend it more with any other properties we think make sense to be overridden at runtime.

### DIFF
--- a/extensions/infinispan-client/runtime/src/main/java/io/quarkus/infinispan/client/runtime/InfinispanClientProducer.java
+++ b/extensions/infinispan-client/runtime/src/main/java/io/quarkus/infinispan/client/runtime/InfinispanClientProducer.java
@@ -4,7 +4,6 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.lang.annotation.Annotation;
 import java.util.Map;
-import java.util.Optional;
 import java.util.Properties;
 import java.util.Scanner;
 import java.util.Set;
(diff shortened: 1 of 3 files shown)


mswiderski:
@wburns would appreciate any feedback on this

extensions/infinispan-client/runtime/src/main/java/io/quarkus/infinispan/client/runtime/InfinispanClientRuntimeConfig.java
```diff
@@ -18,6 +18,12 @@
     @ConfigItem
     public Optional<String> serverList;
 
+    /**
+     * Configures hotrod protocol settings
+     */
+    @ConfigItem
+    public InfinispanHotRodClientRuntimeConfig hotrod;
```
wburns: I would just say to add it as properties in this class directly, no reason to make another sub property for the runtime config

wburns (changes requested):
We also need to document these settings in the user guide. https://github.com/quarkusio/quarkus/blob/master/docs/src/main/asciidoc/infinispan-client-guide.adoc

There is a table talking about the various configuration options and what they are used for and if runtime override-able, which all of these are.

extensions/infinispan-client/runtime/src/main/java/io/quarkus/infinispan/client/runtime/InfinispanClientRuntimeConfig.java
mswiderski: The reason why it was moved to separate class was to have the prefix ‘hotrod’ in the config. Just for clarity in case there will be other approach to configure communication and to separate it from general Infinispan config params 

But if you prefer to have it all in single class I can refactor it

wburns: I would say to keep it in the same class, all of these properties are the same level in the hotrod-client.properties file as well.

mswiderski:
@wburns changes applied and docs updated

extensions/infinispan-client/runtime/src/main/java/io/quarkus/infinispan/client/runtime/InfinispanClientProducer.java
```diff
@@ -167,12 +167,34 @@ private ConfigurationBuilder builderFromProperties(Properties properties) {
             builder.marshaller((Marshaller) marshallerInstance);
         }
 
-        // Override serverList property value at runtime if such configuration exists
         if (infinispanClientRuntimeConfig != null) {
+            // Override serverList property value at runtime if such configuration exists
             Optional<String> runtimeServerList = infinispanClientRuntimeConfig.serverList;
```
wburns: Sorry to ask this, but can we change this one to be the same format (ie. using ifPresent lambda) so the code is all consistent.

wburns (approved):
Changes look good to me. Don't know if my approval matters though :D

extensions/infinispan-client/runtime/src/main/java/io/quarkus/infinispan/client/runtime/InfinispanClientProducer.java
mswiderski: updated

mswiderski:
@gsmet would you mind to review as well? Would be great to have this included in 0.25.0 if possible

extensions/infinispan-client/runtime/src/main/java/io/quarkus/infinispan/client/runtime/InfinispanClientProducer.java
```diff
@@ -167,12 +166,31 @@ private ConfigurationBuilder builderFromProperties(Properties properties) {
             builder.marshaller((Marshaller) marshallerInstance);
         }
 
-        // Override serverList property value at runtime if such configuration exists
         if (infinispanClientRuntimeConfig != null) {
-            Optional<String> runtimeServerList = infinispanClientRuntimeConfig.serverList;
-            if (runtimeServerList.isPresent()) {
-                properties.put(ConfigurationProperties.SERVER_LIST, runtimeServerList.get());
-            }
```
gsmet: I have an issue with this change. It doesn't look related and could potentially break things. Is it safe? Is the server list still overridable at runtime?

gsmet (changes requested):

extensions/infinispan-client/runtime/src/main/java/io/quarkus/infinispan/client/runtime/InfinispanClientProducer.java
mswiderski: yes, it actually the most frequently overridden property as the Infinispan server will differ between different environments

mswiderski: see here https://github.com/quarkusio/quarkus/pull/4502#discussion_r335095046

gsmet (approved):

extensions/infinispan-client/runtime/src/main/java/io/quarkus/infinispan/client/runtime/InfinispanClientProducer.java
gsmet: Ah I totally missed it was still there.

PR #4504, opened 2019-10-10
Have the Agroal datasources initialized eagerly

Sanne:
This also removes the proxy layer around it, so as a nice side-effect it should be more efficient; certainly less code.

Main goal though is to stop inizializing it "on demand", as it doesn't give a good experience by allowing a successfull boot with misconfigured datasource.


gsmet:
Hmmm. I might misremember but I think I made them singleton at first and then something didn't work very well in native.

But if it's working this way, it's certainly better.

Sanne:
Opening an issue to track this as I need to move on other topics #4507

PR #4506, opened 2019-10-10
Fix Gradle tests on release

### LINKED ISSUE #4500: Gradle tests not passing when releasing
If you get this tag: https://github.com/quarkusio/quarkus/commits/0.24.0 and removes the commit skipping the Gradle tests (the one before the last one), you will have the Gradle tests failing.

We need to fix that as it's blocking the release process.

Note that I haven't pushed this commit to master so if we don't find a solution by next release, I will have to do it.

gastaldi:
The quarkus version specified in the quarkus.properties file can be fixed, since we're testing the dependency resolution

Fixes #4500

### DIFF
--- a/devtools/gradle/src/test/resources/gradle-project/gradle.properties
+++ b/devtools/gradle/src/test/resources/gradle-project/gradle.properties
@@ -1,1 +1,1 @@
-quarkusVersion = 999-SNAPSHOT
+quarkusVersion = 0.23.2
--- a/devtools/gradle/src/test/java/io/quarkus/gradle/GradleBuildFileTest.java
+++ b/devtools/gradle/src/test/java/io/quarkus/gradle/GradleBuildFileTest.java
@@ -47,14 +47,14 @@ class GradleBuildFileTest {
             }
             depsString.add(managementKey);
         }
-        assertThat(depsString).contains("io.quarkus:quarkus-jsonp:jar:999-SNAPSHOT",
-                "io.quarkus:quarkus-jsonb:jar:999-SNAPSHOT", "io.quarkus:quarkus-resteasy:jar:999-SNAPSHOT");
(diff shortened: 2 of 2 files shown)


devtools/gradle/src/test/resources/gradle-project/gradle.properties
```diff
@@ -1 +1 @@
-quarkusVersion = 999-SNAPSHOT
\ No newline at end of file
+quarkusVersion = 0.23.2
```
aloubyansky: This will make it actually resolve 0.23.2 deps, right? We should avoid that. What if we keep it equal to the current Quarkus version and also use the same value in the test? The test could read this property file directly, right?

gastaldi: @aloubyansky I am not sure I agree with that. Using the same quarkus version as the current release would require mvn installing the dependent artifacts first, which may not be an ideal approach. Also the dependencies between them may change between releases, which would impact this test.

aloubyansky: > @aloubyansky I am not sure I agree with that. Using the same quarkus version as the current release would require mvn installing the dependent artifacts first, which may not be an ideal approach.
Actually why not? The dependencies are currently installed before this test runs anyway.

> Also the dependencies between them may change between releases, which would impact this test.
It can easily be adjusted in the test though. OTOH, the advantage of using the current version will be to save build/test time by downloading less artifacts.

What we could do in perspective is create a fake platform for this kind of tests.

gastaldi: Perhaps we could merge this solution in order to re-enable the disabled tests and then revisit when this fake platform exists?

aloubyansky (approved):

PR #4508, opened 2019-10-10
Support running tests from third party test-jars transformed using

ppalaga:
rpkgtests-maven-plugin

cc @aloubyansky


aloubyansky:
I'll try one alternative approach tonight.

aloubyansky:
> I'll try one alternative approach tonight.

I'm getting some classpath issues but it looks like it might work, so I'll continue looking into it tomorrow.

gsmet:
@aloubyansky Max asked me to review this one. Not sure what I can do, the patch looks safe enough and not risky at all so if it helps working around some issue, I would say let's get it in?

aloubyansky:
Ok @gsmet

gsmet (approved):
Let's get this in for now, we can still remove it later if this approach doesn't end up being viable.

Issue #4324, opened 2019-10-02
Add a way to disable http and have only https available


mazegreg:
**Description**
Add the possibility to disable HTTP, and to have only HTTPSremaining.

**Implementation ideas**
Support a parameter that could be called "quarkus.http.port.state" (and value = "enabled" by default) that could be set to "disabled" by the application.properties.
If set so, the application would not open the HTTP port (not sure where this should be implemented!).

As @stuartwdouglas says in this zulip topic (https://quarkusio.zulipchat.com/#narrow/stream/187030-users/topic/Disable.20http/near/177142685), if HTTP is disabled, there should be a mechanism that automatically redirects to HTTPS (maybe with usual 443 port?).

CSTDev:
I've attempted to resolve this, what I've done works (as far as I've tested) but I'm having issues getting the unit test to work, if anyone could take a look and offer suggestions that'd be great, thanks!

emmanuelbernard:
Fixed by #7588 I think

PR #4509, opened 2019-10-10
Add option to disable the http port when https is enabled

CSTDev:
Looks to address #4324  

Redirects any http calls to the https endpoint when the property is set, but defaults to true:
`quarkus.http.port-enabled=false`

Sorted the unit test to test it sets the Location header and returns a 301 response.

### DIFF
--- a/extensions/vertx-http/deployment/src/test/java/io/quarkus/vertx/http/DisableHttpPortTest.java
+++ b/extensions/vertx-http/deployment/src/test/java/io/quarkus/vertx/http/DisableHttpPortTest.java
@@ -1,10 +1,11 @@
 package io.quarkus.vertx.http;
 
+import static io.restassured.RestAssured.given;
+import static io.restassured.config.RedirectConfig.redirectConfig;
+import static io.restassured.config.RestAssuredConfig.newConfig;
 import static org.hamcrest.core.Is.is;
 
 import java.io.File;
-import java.net.MalformedURLException;
-import java.net.URL;
 
 import javax.enterprise.context.ApplicationScoped;
 import javax.enterprise.event.Observes;
@@ -14,13 +15,11 @@ import org.jboss.shrinkwrap.api.ShrinkWrap;
 import org.jboss.shrinkwrap.api.spec.JavaArchive;
 import org.junit.jupiter.api.AfterAll;
 import org.junit.jupiter.api.BeforeAll;
-import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.RegisterExtension;
 
 import io.quarkus.test.QuarkusUnitTest;
 import io.restassured.RestAssured;
-import io.restassured.response.ValidatableResponse;
 import io.vertx.ext.web.Router;
 
 public class DisableHttpPortTest {
@@ -30,7 +29,8 @@ public class DisableHttpPortTest {
             .setArchiveProducer(() -> ShrinkWrap.create(JavaArchive.class)
                     .addClasses(MyBean.class)
                     .addAsResource(new File("src/test/resources/conf/disable-http.conf"), "application.properties")
-                    .addAsResource(new File("src/test/resources/conf/server-keystore.jks"), "server-keystore.jks"));
+                    .addAsResource(new File("src/test/resources/conf/server-key.pem"), "server-key.pem")
+                    .addAsResource(new File("src/test/resources/conf/server-cert.pem"), "server-cert.pem"));
 
     @BeforeAll
     public static void setupRestAssured() {
@@ -42,13 +42,11 @@ public class DisableHttpPortTest {
         RestAssured.reset();
     }
 
-    @Disabled
     @Test
-    public void testDisabledHttpPortForwardsRequest() throws MalformedURLException {
-        URL url = new URL("http://localhost:8081/ssl");
-        RestAssured.config().getRedirectConfig().followRedirects(false);
-        ValidatableResponse response = RestAssured.get(url).then().statusCode(301).body(is("ssl"));
-        response.header("Location", "https://localhost:8443/ssl");
+    public void testDisabledHttpPortForwardsRequest() {
+        given().config(newConfig().redirect(redirectConfig().followRedirects(false)))
+                .get("/ssl").then().statusCode(301)
+                .header("Location", is("https://localhost:8444/ssl"));
     }
 
     @ApplicationScoped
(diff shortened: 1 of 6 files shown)


extensions/vertx-http/deployment/src/test/java/io/quarkus/vertx/http/DisableHttpPortTest.java
```diff
@@ -0,0 +1,67 @@
...
+    @AfterAll
+    public static void restoreRestAssured() {
+        RestAssured.reset();
+    }
+
+    @Disabled
+    @Test
+    public void testDisabledHttpPortForwardsRequest() throws MalformedURLException {
+        URL url = new URL("http://localhost:8081/ssl");
+        RestAssured.config().getRedirectConfig().followRedirects(false);
```
stuartwdouglas: I think this is global config that will affect all subsequent tests, so it needs to be reverted in a finally block.

extensions/vertx-http/deployment/src/test/java/io/quarkus/vertx/http/DisableHttpPortTest.java
```diff
@@ -0,0 +1,67 @@
...
+        ValidatableResponse response = RestAssured.get(url).then().statusCode(301).body(is("ssl"));
```
stuartwdouglas: I think you need .relaxedHTTPSValidation() here to allow for self signed certificates.

stuartwdouglas (approved):

stuartwdouglas (changes requested):
oops, wrong button

extensions/vertx-http/deployment/src/test/java/io/quarkus/vertx/http/DisableHttpPortTest.java
CSTDev: This was it, it wasn't actually being set for the request. I've moved it to be request specific so it won't affect subsequent tests and it actually does it for this one. Thanks!

extensions/vertx-http/deployment/src/test/java/io/quarkus/vertx/http/DisableHttpPortTest.java
CSTDev: It sets it in the BeforeAll method which works ok.

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/HttpConfiguration.java
```diff
@@ -47,6 +47,13 @@
     @ConfigItem(defaultValue = "8444")
     public int testSslPort;
 
+    /**
+     * Whether the HTTP port should be available for use
+     * if not any requests to the HTTP port will be redirected to the HTTPS port
+     */
+    @ConfigItem(defaultValue = "true")
+    public boolean portEnabled;
```
stuartwdouglas: I don't think this is a very descriptive config property, both http and https are configured here so it is not clear what this does.. Maybe 'requireHttps' would be a better name.

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/VertxHttpRecorder.java
```diff
@@ -385,12 +390,16 @@ private static void createPemKeyCertOptions(Path certFile, Path keyFile,
...
             LaunchMode launchMode) {
-        // TODO other config properties
-        HttpServerOptions options = new HttpServerOptions();
-        options.setHost(httpConfiguration.host);
-        options.setPort(httpConfiguration.determinePort(launchMode));
-        options.setMaxHeaderSize(httpConfiguration.limits.maxHeaderSize.asBigInteger().intValueExact());
-        return options;
+        if (httpConfiguration.portEnabled) {
+            // TODO other config properties
+            HttpServerOptions options = new HttpServerOptions();
```
stuartwdouglas: You still need to create these options if you are going to start the server and send a redirect, otherwise the user can't control the port. The header limits are also important.

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/VertxHttpRecorder.java
```diff
@@ -438,9 +447,25 @@ public WebDeploymentVerticle(int port, int httpsPort, String host, HttpServerOpt
...
+            if (httpOptions != null) {
+                httpServer = vertx.createHttpServer(httpOptions);
+                httpServer.requestHandler(ACTUAL_ROOT);
+            } else {
+                httpServer = vertx.createHttpServer(new HttpServerOptions().setPort(port).setHost(host))
+                        .requestHandler(req -> {
+                            req.response()
+                                    .setStatusCode(301)
+                                    .putHeader("Location",
+                                            req.absoluteURI().replace("http", "https").replace(Integer.toString(port),
```
stuartwdouglas: This is a more general issue that we need to address but this will fail if you are behind a reverse proxy. This needs to be addressed in a more general way however.

stuartwdouglas (changes requested):

mazegreg:
@stuartwdouglas , any idea how to fix this one? I think this is an important point that comes again and again as requests.

stuartwdouglas:
I have fixed the conflicts and made some other minor changes and force pushed a new version.

stuartwdouglas:
@machi1990 The most recent changes are mine, so someone else should review them.

liquidblock:
Are there any reasons to not allow for both configuration options `quarkus.http.redirect-insecure-requests` and `quarkus.http.port-enabled`? As of commit 190b7e8e5db8648ec4afcb026a38e3126cc53c82 only redirections are possible.

When using the smallrye-jwt quarkus extension, the JSON Web Tokens are passed to the application via the Authorization-Header. In that case a HTTP redirection could reveal the token since it could be sent in the header of the unencrypted HTTP connection and it can be therefore be prefered to disable the HTTP port. This will also reduce unneccessary allocated resources (e.g. a system TCP port).
Redirection is, of course, useful in other scenarios.

stuartwdouglas:
We can do both, but it needs a better name than port-enable. Something like https-only.

stuartwdouglas:
actually I think it should be: insecure-requests=allow|redirect|disabled

gsmet:
@CSTDev any chance you could revisit your PR with Stuart's proposal?

stuartwdouglas:
Superseded by https://github.com/quarkusio/quarkus/pull/7588

I took this work and changed it to allow all three options, and resolved the conflicts

PR #4515, opened 2019-10-11
add custom configuration property validation

### LINKED ISSUE #4431: Bean Validation and Eclipse MicroProfile Spec
**Description**

Hi today I was thinking that it would be interesting to have bean validation integrated with Config API (maybe not per spec but as a feature we are offering in Quarkus) so for example, you can do:

```
@Min(1)
@Max(3)
@ConfigProperty(name = "greeting.repeat", defaultValue = "1")
 int numberOfRepetitions;
```

And if a criteria is not met at runtime, then we can abort the boot process and show the message.

**Implementation ideas**

https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/quarkus-dev/y95HxiHX1iI/NqYLqU5-BwAJ

machi1990:
Fixes #4431 

Opening as a draft to welcome some early feedbacks. 

/cc @geoand @gsmet  thoughts?

### DIFF
--- a/extensions/hibernate-validator/deployment/src/test/java/io/quarkus/hibernate/validator/test/configuration/InvalidConfigurationPropertyTest.java
+++ b/extensions/hibernate-validator/deployment/src/test/java/io/quarkus/hibernate/validator/test/configuration/InvalidConfigurationPropertyTest.java
@@ -1,0 +1,41 @@
+package io.quarkus.hibernate.validator.test.configuration;
+
+import static org.assertj.core.api.Assertions.fail;
+
+import javax.enterprise.context.Dependent;
+import javax.inject.Inject;
+import javax.validation.ConstraintViolationException;
+import javax.validation.constraints.Max;
+
+import org.eclipse.microprofile.config.inject.ConfigProperty;
+import org.jboss.shrinkwrap.api.ShrinkWrap;
+import org.jboss.shrinkwrap.api.asset.StringAsset;
+import org.jboss.shrinkwrap.api.spec.JavaArchive;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.RegisterExtension;
+
+import io.quarkus.test.QuarkusUnitTest;
+
+public class InvalidConfigurationPropertyTest {
+    @Inject
+    InvalidNumberConfigService numberConfigService;
+
+    @RegisterExtension
+    static QuarkusUnitTest runner = new QuarkusUnitTest()
+            .setExpectedException(ConstraintViolationException.class)
+            .setArchiveProducer(() -> ShrinkWrap.create(JavaArchive.class)
+                    .addClasses(InvalidNumberConfigService.class)
+                    .addAsResource(new StringAsset("invalid-number=10"), "application.properties"));
+
+    @Test
+    public void shouldFail() {
+        fail("should not reach here because configuration value is not valid");
+    }
+
+    @Dependent
+    public static class InvalidNumberConfigService {
+        @Max(4)
+        @ConfigProperty(name = "invalid-number")
+        int invalidNumber;
+    }
+}
--- a/extensions/hibernate-validator/runtime/src/main/java/io/quarkus/hibernate/validator/runtime/HibernateValidatorRecorder.java
+++ b/extensions/hibernate-validator/runtime/src/main/java/io/quarkus/hibernate/validator/runtime/HibernateValidatorRecorder.java
@@ -2,7 +2,9 @@ package io.quarkus.hibernate.validator.runtime;
 
 import java.lang.reflect.Method;
 import java.util.Collections;
+import java.util.List;
 import java.util.Locale;
+import java.util.Map;
 import java.util.Set;
 
 import javax.validation.ClockProvider;
@@ -24,6 +26,7 @@ import io.quarkus.arc.Arc;
 import io.quarkus.arc.InstanceHandle;
 import io.quarkus.arc.runtime.BeanContainer;
 import io.quarkus.arc.runtime.BeanContainerListener;
+import io.quarkus.hibernate.validator.runtime.configuration_validation.ClassWithConfigProperties;
 import io.quarkus.runtime.ShutdownContext;
 import io.quarkus.runtime.annotations.Recorder;
 
@@ -31,7 +34,7 @@ import io.quarkus.runtime.annotations.Recorder;
 public class HibernateValidatorRecorder {
 
     public BeanContainerListener initializeValidatorFactory(Set<Class<?>> classesToBeValidated,
-            ShutdownContext shutdownContext) {
+            ShutdownContext shutdownContext, Map<Class<?>, List<String>> configClassWithProperties) {
         BeanContainerListener beanContainerListener = new BeanContainerListener() {
 
             @Override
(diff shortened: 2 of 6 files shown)


geoand:
Thanks for this @machi1990! I will take a look soon and let you know what I think

machi1990:
> Thanks for this @machi1990! I will take a look soon and let you know what I think

Thanks @geoand, looking forward to your feedbacks.

extensions/hibernate-validator/runtime/src/main/java/io/quarkus/hibernate/validator/runtime/HibernateValidatorRecorder.java
```diff
@@ -1,9 +1,7 @@
 package io.quarkus.hibernate.validator.runtime;
 
 import java.lang.reflect.Method;
-import java.util.Collections;
-import java.util.Locale;
-import java.util.Set;
+import java.util.*;
```
geoand: This should be reverted

extensions/hibernate-validator/runtime/src/main/java/io/quarkus/hibernate/validator/runtime/configuration_validation/ClassWithConfigProperties.java
```diff
@@ -0,0 +1,69 @@
...
+import javax.validation.Validator;
+
+/**
+ * Holds the class and the configuration properties i.e class fields annotated with
+ * {@link org.eclipse.microprofile.config.inject.ConfigProperty}
+ */
+public class ClassWithConfigProperties {
+    private final Class<?> aClass;
+    private final List<String> configurationProperties;
+    static List<ClassWithConfigProperties> classWithConfigurationProperties;
```
geoand: I personally find the name of the field confusing. I would have named it `instances` or something

extensions/hibernate-validator/runtime/src/main/java/io/quarkus/hibernate/validator/runtime/configuration_validation/ClassWithConfigProperties.java
```diff
@@ -0,0 +1,69 @@
...
+        this.aClass = aClass;
+        this.configurationProperties = configurationProperties;
+    }
+
+    /**
+     * Throw an {@link ConstraintViolationException} when a configuration property is not valid
+     */
+    void validate(Validator validator) {
+        for (Object obj : CDI.current().select(aClass)) {
+            for (String property : configurationProperties) {
```
geoand: Is there a reason to validate each property on it's own and not just validate the whole class?

extensions/hibernate-validator/runtime/src/main/java/io/quarkus/hibernate/validator/runtime/configuration_validation/ClassWithConfigProperties.java
```diff
@@ -0,0 +1,69 @@
...
+
+    @Override
+    public String toString() {
+        return "ClassWithConfigProperties{" +
+                "aClass=" + aClass +
+                ", configurationProperties=" + configurationProperties +
+                '}';
+    }
+
+    public static void initialize(Map<Class<?>, List<String>> configClassWithProperties) {
```
geoand: I personally don't like this idiom of using static methods to initialize things, but 1) I know I am a minority since the alternative of generating bytecode is harder, 2) it's already being done in various parts of the code.

extensions/hibernate-validator/deployment/src/test/java/io/quarkus/hibernate/validator/test/configuration/InvalidConfigurationPropertyTest.java
```diff
@@ -0,0 +1,41 @@
...
+import org.eclipse.microprofile.config.inject.ConfigProperty;
+import org.jboss.shrinkwrap.api.ShrinkWrap;
+import org.jboss.shrinkwrap.api.asset.StringAsset;
+import org.jboss.shrinkwrap.api.spec.JavaArchive;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.RegisterExtension;
+
+import io.quarkus.test.QuarkusUnitTest;
+
+public class InvalidConfigurationPropertyTest {
```
geoand: :+1:

extensions/hibernate-validator/runtime/src/main/java/io/quarkus/hibernate/validator/runtime/configuration_validation/ClassWithConfigProperties.java
machi1990: To be honest, I also wanted to try the class generation kung-fu, only done a couple of time so I dod not have the "black" belt. I leaned toward this option as it maintainable but If we both think that  generating bytecode is the way to go, I'll be happy to  give it a try.  

A few options I was planning to explore:
-  For each `ClassWithConfigProperties` construct a "bridge" class with a `@ConfigProperties` and add all proper annotations for each field so that they will be validated automatically since we have bean validation already in place for `ConfigProperties` class. 
- Find a way to hook in `ClassConfigPropertiesUtil` and add validations specific code for found classes (ignore fields that are not config properties).

extensions/hibernate-validator/runtime/src/main/java/io/quarkus/hibernate/validator/runtime/configuration_validation/ClassWithConfigProperties.java
machi1990: Oh yes, we would not want to throw an exception for the wrong reasons i.e class might contain other  fields (not config property) that needs to be validated. 

That said, I was undecided between validating at once and construct the message only considering constraints violation for config properties, or this current behaviour. I kept he behaviour for clarity also, i.e `only validate a config properties`

extensions/hibernate-validator/runtime/src/main/java/io/quarkus/hibernate/validator/runtime/configuration_validation/ClassWithConfigProperties.java
machi1990: Thanks, I like the proposed naming.

extensions/hibernate-validator/runtime/src/main/java/io/quarkus/hibernate/validator/runtime/HibernateValidatorRecorder.java
machi1990: Ooops, should be okay now.

extensions/hibernate-validator/runtime/src/main/java/io/quarkus/hibernate/validator/runtime/configuration_validation/ClassWithConfigProperties.java
geoand: @gsmet is the expert here, so let's see what he says :)

extensions/hibernate-validator/runtime/src/main/java/io/quarkus/hibernate/validator/runtime/configuration_validation/ClassWithConfigProperties.java
geoand: I wouldn't leverage `@ConfigProperties`  for this since it doesn't seem natural to generate such an ex class just to leverage validation.
If it were me implementing this, I would just generate some classes that integrate with validation in a similar way that `@ConfigProperties` does.
But I would say don't do it yet. Let's see what @gsmet thinks.

machi1990:
Let's keep this open, we still need some feedbacks from @gsmet @geoand on how we take this from here.

geoand:
I like this feature and I think we should get it in sometime soon

machi1990:
@gsmet  ping..I'll like your opinion on this PR.

machi1990:
Closing as it will have to be solved upstream. See for more info
https://github.com/quarkusio/quarkus/issues/4431#issuecomment-607317781

PR #4231, opened 2019-09-27
Bump jackson to 2.10.0

gastaldi:
Release notes: https://github.com/FasterXML/jackson/wiki/Jackson-Release-2.10

PR #4524, opened 2019-10-11
Update Jackson to 2.10.0

wburns:
Trivial change to just update to latest version of Jackson.

### DIFF
--- a/bom/runtime/pom.xml
+++ b/bom/runtime/pom.xml
@@ -68,7 +68,7 @@
            is not available in Maven fast enough: -->
         <graal-sdk.version>19.2.0.1</graal-sdk.version>
         <gizmo.version>1.0.0.Alpha8</gizmo.version>
-        <jackson.version>2.9.10</jackson.version>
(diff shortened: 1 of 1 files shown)


geoand:
@cescoffier I believe you wanted to hold off from this upgrade because vert.x wasn't using it?

cescoffier:
Exactly. @wburns, which API / bug has been fixed in this version?

wburns:
Infinispan updated our dependency version to 2.10.0 so this causes a version mismatch; The two versions are not binary compatible since Jackson added some new interfaces. Fortunately we can work around it for now though as this only manifests itself in our upcoming server outside of Quarkus repo.

machi1990:
We have a similar change here https://github.com/quarkusio/quarkus/pull/4231, maybe we can close either of the two and continue the discussion on a single PR? WDYT? @gsmet

wburns:
Ah, yes, sorry for the duplicate. I will just close this one.

PR #4525, opened 2019-10-11
Introduce YAML configuration extension

### LINKED ISSUE #3904: Add support for YAML configuration files
Some people prefer yaml for the hierarchical configuration structure we have.
It would be useful I think to support it as an extra `configsource` https://github.com/eclipse/microprofile-config/blob/master/spec/src/main/asciidoc/configsources.asciidoc

`application.yaml`

and make sure we enlist that file for hot reload etc (everywhere `application.properties` is used really).

As for priority, we could decide it's the `.properties` priority + 1 (no strong preference).

Anyone against this?
CC @stuartwdouglas @kenfinnigan @dmlloyd 

I don't think it needs to be documented beyond the configuration guide.

I was wondering if it could/should be an optional extension. Not sure how tied to core it is.

gwenneg:
Fixes #3904

### DIFF
--- a/extensions/config-yaml/runtime/src/main/java/io/quarkus/yaml/configuration/runtime/YamlConfigConstants.java
+++ b/extensions/config-yaml/runtime/src/main/java/io/quarkus/yaml/configuration/runtime/YamlConfigConstants.java
@@ -1,0 +1,12 @@
+package io.quarkus.yaml.configuration.runtime;
+
+public class YamlConfigConstants {
+
+    public static final String APPLICATION_YML_FILE = "application.yaml";
+    public static final String MICROPROFILE_CONFIG_YML_FILE = "META-INF/microprofile-config.yaml";
+
+    public static final int MICROPROFILE_CONFIG_YML_JAR_PRIORITY = 251;
+    public static final int APPLICATION_YML_JAR_PRIORITY = 252;
+    public static final int MICROPROFILE_CONFIG_YML_FILESYSTEM_PRIORITY = 261;
+    public static final int APPLICATION_YML_FILESYSTEM_PRIORITY = 262;
+}
--- a/extensions/config-yaml/runtime/src/main/java/io/quarkus/yaml/configuration/runtime/YamlConfigSource.java
+++ b/extensions/config-yaml/runtime/src/main/java/io/quarkus/yaml/configuration/runtime/YamlConfigSource.java
@@ -1,0 +1,110 @@
+package io.quarkus.yaml.configuration.runtime;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.Serializable;
+import java.io.UncheckedIOException;
+import java.net.URL;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.Map.Entry;
+import java.util.StringJoiner;
+import java.util.TreeMap;
+
+import org.eclipse.microprofile.config.spi.ConfigSource;
+import org.yaml.snakeyaml.Yaml;
+
+public class YamlConfigSource implements ConfigSource, Serializable {
+
+    private static final long serialVersionUID = -875865410658150582L;
+
+    private static final String CONFIG_ORDINAL_KEY = "config_ordinal";
+    private static final String CONFIG_ORDINAL_DEFAULT_VALUE = "100";
+    private static final String PROPERTY_KEY_SEPARATOR = ".";
+    private static final String COMMA = ",";
+    private static final String ESCAPED_COMMA = "\\,";
+
+    private final URL source;
+    private final int ordinal;
+    private final Map<String, String> properties;
+
+    public YamlConfigSource(URL source, int ordinal) {
+        this.source = source;
+        this.properties = buildConfigPropertiesFromYaml();
+        if (properties.containsKey(CONFIG_ORDINAL_KEY)) {
+            this.ordinal = Integer.parseInt(properties.getOrDefault(CONFIG_ORDINAL_KEY, CONFIG_ORDINAL_DEFAULT_VALUE));
+        } else {
+            this.ordinal = ordinal;
+        }
+    }
+
+    private Map<String, String> buildConfigPropertiesFromYaml() {
+        if (source == null) {
+            return Collections.emptyMap();
+        }
+        try (InputStream is = source.openStream()) {
+            @SuppressWarnings("unchecked")
+            Map<String, Object> yamlContent = new Yaml().loadAs(is, TreeMap.class);
+            Map<String, String> configProperties = new TreeMap<>();
+            convertYamlMapToProperties(configProperties, "", yamlContent); // The root `baseKey` is empty.
+            return configProperties;
+        } catch (IOException e) {
+            throw new UncheckedIOException(e);
+        }
+    }
+
+    private void convertYamlMapToProperties(Map<String, String> properties, String baseKey, Object object) {
+        if (object instanceof Map) {
+            for (Entry<?, ?> entry : ((Map<?, ?>) object).entrySet()) {
+                String entryKey = baseKey + (baseKey.isEmpty() ? "" : PROPERTY_KEY_SEPARATOR) + entry.getKey();
+                convertYamlMapToProperties(properties, entryKey, entry.getValue());
+            }
+        } else {
+            String value;
+            if (object instanceof List) {
+                value = stringifyList(object);
+            } else {
+                value = String.valueOf(object);
+            }
+            properties.put(baseKey, value);
+        }
+    }
+
+    private String stringifyList(Object list) {
+        StringJoiner stringifiedElements = new StringJoiner(COMMA);
+        for (Object elem : (List<?>) list) {
+            String stringifiedElem = String.valueOf(elem);
+            if (stringifiedElem.contains(COMMA)) {
+                stringifiedElem = stringifiedElem.replace(COMMA, ESCAPED_COMMA);
+            }
+            stringifiedElements.add(stringifiedElem);
+        }
+        return stringifiedElements.toString();
+    }
+
+    @Override
+    public Map<String, String> getProperties() {
+        return Collections.unmodifiableMap(properties);
+    }
+
+    @Override
+    public String getValue(String key) {
+        return properties.get(key);
+    }
+
+    @Override
+    public int getOrdinal() {
+        return ordinal;
+    }
+
+    @Override
+    public String getName() {
+        return "YamlConfigSource[source=" + source + "]";
+    }
+
+    @Override
+    public String toString() {
+        return getName();
+    }
+}
--- a/extensions/config-yaml/runtime/src/main/java/io/quarkus/yaml/configuration/runtime/YamlConfigSourceProvider.java
+++ b/extensions/config-yaml/runtime/src/main/java/io/quarkus/yaml/configuration/runtime/YamlConfigSourceProvider.java
@@ -1,0 +1,61 @@
+package io.quarkus.yaml.configuration.runtime;
+
+import static io.quarkus.yaml.configuration.runtime.YamlConfigConstants.APPLICATION_YML_FILE;
+import static io.quarkus.yaml.configuration.runtime.YamlConfigConstants.APPLICATION_YML_FILESYSTEM_PRIORITY;
+import static io.quarkus.yaml.configuration.runtime.YamlConfigConstants.APPLICATION_YML_JAR_PRIORITY;
+import static io.quarkus.yaml.configuration.runtime.YamlConfigConstants.MICROPROFILE_CONFIG_YML_FILE;
+import static io.quarkus.yaml.configuration.runtime.YamlConfigConstants.MICROPROFILE_CONFIG_YML_FILESYSTEM_PRIORITY;
+import static io.quarkus.yaml.configuration.runtime.YamlConfigConstants.MICROPROFILE_CONFIG_YML_JAR_PRIORITY;
+
+import java.io.UncheckedIOException;
+import java.net.MalformedURLException;
+import java.net.URL;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.ArrayList;
+import java.util.List;
+
+import org.eclipse.microprofile.config.spi.ConfigSource;
+import org.eclipse.microprofile.config.spi.ConfigSourceProvider;
+
+public class YamlConfigSourceProvider implements ConfigSourceProvider {
+
+    @Override
+    public Iterable<ConfigSource> getConfigSources(ClassLoader classLoader) {
+        List<ConfigSource> configSources = new ArrayList<>();
+
+        URL jarMicroprofileConfigYml = classLoader.getResource(MICROPROFILE_CONFIG_YML_FILE);
+        if (jarMicroprofileConfigYml != null) {
+            configSources.add(new YamlConfigSource(jarMicroprofileConfigYml, MICROPROFILE_CONFIG_YML_JAR_PRIORITY));
+        }
+        URL jarApplicationYml = classLoader.getResource(APPLICATION_YML_FILE);
+        if (jarApplicationYml != null) {
+            configSources.add(new YamlConfigSource(jarApplicationYml, APPLICATION_YML_JAR_PRIORITY));
+        }
+
+        URL fsMicroprofileConfigYml = getFileSystemURL(MICROPROFILE_CONFIG_YML_FILE);
+        if (fsMicroprofileConfigYml != null) {
+            configSources.add(new YamlConfigSource(fsMicroprofileConfigYml, MICROPROFILE_CONFIG_YML_FILESYSTEM_PRIORITY));
+        }
+        URL fsApplicationYml = getFileSystemURL(APPLICATION_YML_FILE);
+        if (fsApplicationYml != null) {
+            configSources.add(new YamlConfigSource(fsApplicationYml, APPLICATION_YML_FILESYSTEM_PRIORITY));
+        }
+
+        return configSources;
+    }
+
+    private URL getFileSystemURL(String source) {
+        Path path = Paths.get("config", source);
+        if (Files.exists(path)) {
+            try {
+                return path.toUri().toURL();
+            } catch (MalformedURLException e) {
+                throw new UncheckedIOException(e);
+            }
+        } else {
+            return null;
+        }
+    }
+}
--- a/extensions/config-yaml/runtime/src/main/resources/META-INF/quarkus-extension.yaml
+++ b/extensions/config-yaml/runtime/src/main/resources/META-INF/quarkus-extension.yaml
@@ -1,0 +1,10 @@
+---
+name: "YAML Configuration"
+metadata:
+  keywords:
+  - "yaml"
+  - "config"
+  - "configuration"
+  categories:
+  - "config"
+  status: "stable"
--- a/bom/deployment/pom.xml
+++ b/bom/deployment/pom.xml
@@ -488,6 +488,11 @@
                 <artifactId>quarkus-scala-deployment</artifactId>
                 <version>${project.version}</version>
             </dependency>
+            <dependency>
+                <groupId>io.quarkus</groupId>
+                <artifactId>quarkus-config-yaml-deployment</artifactId>
+                <version>${project.version}</version>
+            </dependency>
 
             <!-- Quarkus test dependencies -->
 
--- a/bom/runtime/pom.xml
+++ b/bom/runtime/pom.xml
@@ -172,6 +172,7 @@
         <antlr.version>4.7.2</antlr.version>
         <quarkus-security.version>1.0.0.Beta1</quarkus-security.version>
         <keycloak.version>7.0.1</keycloak.version>
+        <snakeyaml.version>1.25</snakeyaml.version>
     </properties>
 
     <dependencyManagement>
@@ -696,6 +697,11 @@
                 <artifactId>quarkus-vault-spi</artifactId>
                 <version>${project.version}</version>
             </dependency>
+            <dependency>
+                <groupId>io.quarkus</groupId>
+                <artifactId>quarkus-config-yaml</artifactId>
+                <version>${project.version}</version>
+            </dependency>
 
             <!-- Quarkus test dependencies -->
             <dependency>
@@ -1165,6 +1171,11 @@
                     </exclusion>
                 </exclusions>
             </dependency>
+	        <dependency>
+	            <groupId>org.yaml</groupId>
+	            <artifactId>snakeyaml</artifactId>
+	            <version>${snakeyaml.version}</version>
+	        </dependency>
             <dependency>
                 <groupId>io.sundr</groupId>
                 <artifactId>builder-annotations</artifactId>
--- a/ci-templates/stages.yml
+++ b/ci-templates/stages.yml
@@ -333,6 +333,7 @@ stages:
             - tika
             - hibernate-validator
             - test-extension
+            - config-yaml
           name: misc_2
 
       - template: native-build-steps.yaml
--- a/core/deployment/src/main/java/io/quarkus/deployment/builditem/FeatureBuildItem.java
+++ b/core/deployment/src/main/java/io/quarkus/deployment/builditem/FeatureBuildItem.java
@@ -14,6 +14,7 @@ public final class FeatureBuildItem extends MultiBuildItem {
     public static final String ARTEMIS_CORE = "artemis-core";
     public static final String ARTEMIS_JMS = "artemis-jms";
     public static final String CDI = "cdi";
+    public static final String CONFIG_YAML = "config-yaml";
     public static final String DYNAMODB = "dynamodb";
     public static final String ELASTICSEARCH_REST_CLIENT = "elasticsearch-rest-client";
     public static final String FLYWAY = "flyway";
--- a/docs/src/main/asciidoc/config.adoc
+++ b/docs/src/main/asciidoc/config.adoc
@@ -444,6 +444,58 @@ For example:
 ----
 <1> The command will run with the `prod-aws` profile. This can be overridden using the `quarkus.profile` system property.
 
+== Using a YAML configuration file
+
+You can also configure your Quarkus application using a hierarchical YAML file thanks to the `quarkus-config-yaml` extension.
+
+First, you need to add the extension to your project:
+
+[source,shell]
+--
+./mvnw quarkus:add-extension -Dextensions="config-yaml"
+--
+
+Then, create the `src/main/resources/application.yaml` file with a content corresponding to your desired configuration.
+
+[WARNING]
+====
+The configuration values from `application.yaml` will take precedence over the ones from `application.properties` if both files are present in your application.
+But if a configuration key is present in `application.properties` and not in `application.yaml`, the value from `application.properties` will be used.
+====
+
+Here's an example of a Quarkus `application.properties` configuration file, followed by its `application.yaml` equivalent:
+
+.application.properties
+[source,properties]
+--
+quarkus.http.port=9090
+quarkus.http.test-port=9091
+quarkus.http.ssl.protocols=TLSv1.3,TLSv1.2
+quarkus.log.category."foo.bar".level=DEBUG
+%dev.quarkus.http.port=8080
+--
+
+.application.yaml
+[source,yaml]
+--
+quarkus:
+  http:
+    port: 9090
+    test-port: 9091
+    ssl:
+      protocols:
+        - TLSv1.3
+        - TLSv1.2
+  log:
+    category:
+      "\"foo.bar\"":
+        level: DEBUG
+"%dev":
+  quarkus:
+    http:
+      port: 8080
+--
+
 == Custom Configuration
 
 === Custom configuration sources
--- a/extensions/config-yaml/deployment/pom.xml
+++ b/extensions/config-yaml/deployment/pom.xml
@@ -1,0 +1,61 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <artifactId>quarkus-config-yaml-parent</artifactId>
+        <groupId>io.quarkus</groupId>
+        <version>999-SNAPSHOT</version>
+        <relativePath>../</relativePath>
+    </parent>
+
+    <artifactId>quarkus-config-yaml-deployment</artifactId>
+    <name>Quarkus - YAML Configuration - Deployment</name>
+
+    <dependencies>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-core-deployment</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-config-yaml</artifactId>
+        </dependency>
+
+        <!-- Test dependencies -->
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-junit5-internal</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-resteasy-deployment</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>io.rest-assured</groupId>
+            <artifactId>rest-assured</artifactId>
+            <scope>test</scope>
+        </dependency>
+    </dependencies>
+
+    <build>
+        <plugins>
+            <plugin>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <configuration>
+                    <annotationProcessorPaths>
+                        <path>
+                            <groupId>io.quarkus</groupId>
+                            <artifactId>quarkus-extension-processor</artifactId>
+                            <version>${project.version}</version>
+                        </path>
+                    </annotationProcessorPaths>
+                </configuration>
+            </plugin>
+        </plugins>
+    </build>
+</project>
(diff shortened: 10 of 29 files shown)


gwenneg:
I'm not done yet with this PR, but since I'd like it to be included in `0.25.0`, it's already opened for comments.

extensions/yaml-configuration/runtime/src/main/java/io/quarkus/yaml/configuration/runtime/JarApplicationYamlConfigSource.java
```diff
@@ -0,0 +1,30 @@
+package io.quarkus.yaml.configuration.runtime;
+
+import java.io.InputStream;
+
+public class JarApplicationYamlConfigSource extends YamlConfigSource {
+
+    private static final long serialVersionUID = 2819560924855482475L;
+
+    private static final String YAML_FILE = "application.yml";
+    private static final int PRIORITY = 270;
```
gwenneg: The properties-based version of this config source priority is `250`. See `ApplicationPropertiesConfigSource$InJar`.

extensions/yaml-configuration/runtime/src/main/java/io/quarkus/yaml/configuration/runtime/FileSystemApplicationYamlConfigSource.java
```diff
@@ -0,0 +1,38 @@
...
+import java.nio.file.NoSuchFileException;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+
+public class FileSystemApplicationYamlConfigSource extends YamlConfigSource {
+
+    private static final long serialVersionUID = -3881751839186883952L;
+
+    private static final String YAML_FILE = "application.yml";
+    private static final int PRIORITY = 280;
```
gwenneg: The properties-based version of this config source priority is `260`. See `ApplicationPropertiesConfigSource$InFileSystem`.

extensions/yaml-configuration/runtime/src/main/java/io/quarkus/yaml/configuration/runtime/YamlConfigSource.java
```diff
@@ -0,0 +1,125 @@
...
+            return Collections.singletonList(new YamlConfigProperty(baseKey, value));
+        }
+    }
+
+    private String stringifyList(Object list) {
+        List<String> stringifiedElements = new ArrayList<>();
+        for (Object elem : (List<?>) list) {
+            String stringifiedElem = String.valueOf(elem);
+            if (stringifiedElem.contains(COMMA)) {
+                stringifiedElem.replace(COMMA, ESCAPED_COMMA);
```
gwenneg: `Pattern` compilation could improve the performances here with JDK 8, but the implementation of `String.replace()` was changed in both JDK 9 and 13 and made the `Pattern` compilation useless. I chose to favor the most recent JDKs.

kenfinnigan:
Can't recall if it was discussed previously, but I know there will be a Yaml ConfigSource in https://github.com/smallrye/smallrye-config soon, as it's being donated from https://github.com/microprofile-extensions/config-ext/tree/master/configsource-yaml

gwenneg:
Yes,  @gastaldi and @emmanuelbernard mentioned that in #3904 and this is where I started from.  It shouldn't be hard to move from this PR to the smallrye-config `YamlConfigSource` implementation when it's available.

extensions/yaml-configuration/runtime/src/main/java/io/quarkus/yaml/configuration/runtime/YamlConfigSource.java
gastaldi: Except that Strings are immutable and you forgot to assign the return value to `stringifiedElem` :)

extensions/yaml-configuration/runtime/src/main/java/io/quarkus/yaml/configuration/runtime/YamlConfigSource.java
```diff
@@ -0,0 +1,125 @@
...
+
+    private final Map<String, String> properties;
+    private final String source;
+    private final int ordinal;
+
+    protected YamlConfigSource(String source, int ordinal) {
+        this.properties = buildConfigPropertiesFromYaml();
+        this.source = source;
+        if (properties.containsKey(CONFIG_ORDINAL_KEY)) {
+            this.ordinal = Integer.valueOf(properties.getOrDefault(CONFIG_ORDINAL_KEY, CONFIG_ORDINAL_DEFAULT_VALUE));
```
gastaldi: Use Integer.parseInt here, no need for unboxing

extensions/yaml-configuration/runtime/src/main/java/io/quarkus/yaml/configuration/runtime/YamlConfigSource.java
```diff
@@ -0,0 +1,125 @@
...
+    private static final String ESCAPED_COMMA = "\\\\,";
+
+    private final Map<String, String> properties;
+    private final String source;
+    private final int ordinal;
+
+    protected YamlConfigSource(String source, int ordinal) {
+        this.properties = buildConfigPropertiesFromYaml();
+        this.source = source;
+        if (properties.containsKey(CONFIG_ORDINAL_KEY)) {
```
gastaldi: It seems to be used in the `toString` only? Maybe  pass the URL or a Path and then you don't need the `openIYamlFilestream` method

gastaldi (changes requested):
Added some comments

extensions/yaml-configuration/runtime/src/main/java/io/quarkus/yaml/configuration/runtime/YamlConfigSource.java
gastaldi: Also you won't need two implementation classes with a URL or a Path in the constructor :)

extensions/yaml-configuration/runtime/src/main/java/io/quarkus/yaml/configuration/runtime/YamlConfigSource.java
gwenneg: Thanks for pointing this out! The return value assignment was there at first, I must have missed a copy/paste at some point :) I really need to add that integration test.

gwenneg:
Thanks for your review @gastaldi! I'll fix everything and add the missing content in the PR by the end of the day.

gwenneg:
I just noticed a configuration case that works with properties but does not seem possible using YAML: the Quarkus CORS filter configuration. I couldn't find a valid YAML version of the following properties (but maybe I missed something) :
```
quarkus.http.cors=true
quarkus.http.cors.origins=http://foo.com,http://www.bar.io
quarkus.http.cors.methods=GET,PUT,POST
quarkus.http.cors.headers=X-Custom
quarkus.http.cors.exposed-headers=Content-Disposition
```

To convert this to YAML, we'd need the `quarkus.http.cors` key to be both the key of a key/value pair and the key of a dictionary.

This problem could be easily fixed by using `quarkus.http.cors.enabled=true` instead of `quarkus.http.cors=true` to enable the filter though.

gwenneg:
New version pushed and it is still a draft. Here's a non-exhaustive list of the improvements I'll add tomorrow:
- Support of `microprofile-config.yaml`.
- `YamlConfigSource` improvements suggested by @gastaldi.
- Documentation update.
- Better integration tests (the current version is just there to see if CI is happy with the native tests).

extensions/yaml-configuration/runtime/src/main/java/io/quarkus/yaml/configuration/runtime/YamlConfigSource.java
gwenneg: Fixed, thanks!

extensions/yaml-configuration/runtime/src/main/java/io/quarkus/yaml/configuration/runtime/YamlConfigSource.java
gwenneg: Fixed, thanks!

extensions/yaml-configuration/runtime/pom.xml
```diff
@@ -0,0 +1,48 @@
...
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <artifactId>quarkus-yaml-configuration-parent</artifactId>
+        <groupId>io.quarkus</groupId>
+        <version>999-SNAPSHOT</version>
+        <relativePath>../</relativePath>
+    </parent>
+
+    <artifactId>quarkus-yaml-configuration</artifactId>
```
gastaldi: `quarkus-config-yaml` would be an easier name (since it depends on the *Config* API)

geoand:
The test that is failing seems related

extensions/yaml-configuration/runtime/pom.xml
gwenneg: You're right, `quarkus-config-yaml` sounds much better than `quarkus-yaml-configuration`. I renamed everything.

gwenneg:
`microprofile-config.yml` support was just added.

Here's the list of config sources priorities (a higher one takes precedence over a lower one):
- `application.properties` loaded from JAR : 250
- `application.properties` loaded from file system : 260
- `microprofile-config.yml` loaded from JAR : 270
- `application.yml` loaded from JAR : 280
- `microprofile-config.yml` loaded from file system : 290
- `application.yml` loaded from file system : 300

This list might need to be reordered.

geoand:
Having application.yaml have higher priority than application.properties from the file system is wrong IMHO.
I would expect all the YAML sources to have the same (or off be one) priority as properties sources.

gwenneg:
@geoand The YAML priority over properties was discussed in #3904 but I'm fine with any order.

geoand:
@gwenneg it seems to me that whay was agreed was what I propose as well (properties priorities +1, not +10)

gwenneg:
Yes, but there was already a +10 difference between the two existing sources, so I kept this value and posted the list above to start a discussion about this subject :)

geoand:
Understood 👌.

I propose to use 251 for `application.yaml` in the jar, 261 for the YAML from the file system.

gwenneg:
This would mean that `application.yml` from a JAR has a higher priority than `application.properties` from a JAR, but a lower priority than `application.properties` from the file system. Would it be OK? What about `microprofile-config.yml`, should it be above or below `application.yml` in terms of priority?

geoand:
Yes, my idea is to file system config files to always overrides jar ones. Otherwise what's the point of having them?

I didn't add all the YAML sources, but my idea is pretty simple: use whatever the corresponding `. properties` source uses and just add 1 😂.

I think that anything else will create a lot of confusion.

gwenneg:
Makes sense. I pushed new priorities values and updated the list [above](https://github.com/quarkusio/quarkus/pull/4525#issuecomment-541423835). Thanks @geoand!

geoand:
You are most welcome!

gwenneg:
@gastaldi `IOError` and `RuntimeException` replaced with `UncheckedIOException`, thanks for pointing this out.

extensions/config-yaml/runtime/src/main/java/io/quarkus/yaml/configuration/runtime/FileSystemYamlConfigSource.java
```diff
@@ -0,0 +1,35 @@
...
+
+    private static final long serialVersionUID = -3881751839186883952L;
+
+    protected FileSystemYamlConfigSource(String source, int priority) {
+        super(source, priority);
+    }
+
+    @Override
+    protected InputStream openYamlFileStream() {
+        Path path = Paths.get("config", source);
```
gastaldi: So this always assumes that a `config/<file>` exists in the project?

extensions/config-yaml/runtime/src/main/resources/META-INF/services/org.eclipse.microprofile.config.spi.ConfigSource
```diff
@@ -0,0 +1,4 @@
+io.quarkus.yaml.configuration.runtime.MicroprofileConfigJarYamlConfigSource
+io.quarkus.yaml.configuration.runtime.ApplicationJarYamlConfigSource
+io.quarkus.yaml.configuration.runtime.MicroprofileConfigFileSystemYamlConfigSource
+io.quarkus.yaml.configuration.runtime.ApplicationFileSystemYamlConfigSource
```
gastaldi: Not sure I like these many ConfigSources for just reading YAML files, but maybe I am missing how the Config API works

gastaldi: Ah you could implement a ConfigSourceProvider instead: https://github.com/eclipse/microprofile-config/blob/master/api/src/main/java/org/eclipse/microprofile/config/spi/ConfigSourceProvider.java#L51

extensions/config-yaml/runtime/src/main/java/io/quarkus/yaml/configuration/runtime/YamlConfigSource.java
```diff
@@ -0,0 +1,127 @@
...
+            if (object instanceof List) {
+                value = stringifyList(object);
+            } else {
+                value = String.valueOf(object);
+            }
+            return Collections.singletonList(new YamlConfigProperty(baseKey, value));
+        }
+    }
+
+    private String stringifyList(Object list) {
```
machi1990: @gwenneg just a minor comment - could we build the string inside the loop and avoid `line 89` (looping twice)?

extensions/config-yaml/runtime/src/main/java/io/quarkus/yaml/configuration/runtime/FileSystemYamlConfigSource.java
gwenneg: No, building the `Path` doesn't require the file to exist.

extensions/config-yaml/runtime/src/main/java/io/quarkus/yaml/configuration/runtime/YamlConfigSource.java
gwenneg: Sure, I've replaced the `List` with a `StringJoiner`, thanks!

extensions/config-yaml/runtime/src/main/resources/META-INF/services/org.eclipse.microprofile.config.spi.ConfigSource
gwenneg: I added a `ConfigSourceProvider` implementation but it doesn't really change the fact that we need 4 config sources to cover all YAML cases.

gastaldi: You don't really need 4 classes if you pass an URL and the ordinal in the YamlConfigSource class constructor, just saying :)

gwenneg: Yeah, I still need to think about that, I didn't forget your previous comment ;)

extensions/config-yaml/runtime/src/main/java/io/quarkus/yaml/configuration/runtime/YamlConfigSource.java
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    private static class YamlConfigProperty {
+
+        private YamlConfigProperty(String key, String value) {
+            this.key = key;
+            this.value = value;
+        }
+
+        private String key;
```
gastaldi: Add `final` in these attributes

gwenneg: Fixed.

gwenneg:
@gastaldi Is my last commit close to what you were thinking about regarding the use of URL in `YamlConfigSource`?

extensions/config-yaml/runtime/src/main/java/io/quarkus/yaml/configuration/runtime/YamlConfigSourceProvider.java
```diff
@@ -0,0 +1,57 @@
...
+import java.util.Arrays;
+
+import org.eclipse.microprofile.config.spi.ConfigSource;
+import org.eclipse.microprofile.config.spi.ConfigSourceProvider;
+
+public class YamlConfigSourceProvider implements ConfigSourceProvider {
+
+    @Override
+    public Iterable<ConfigSource> getConfigSources(ClassLoader forClassLoader) {
+        ClassLoader cl = getClassLoader();
```
gastaldi: What's wrong with the classloader in the parameter?

extensions/config-yaml/runtime/src/main/java/io/quarkus/yaml/configuration/runtime/YamlConfigSourceProvider.java
```diff
@@ -0,0 +1,57 @@
...
+        return Arrays.asList(
+                new YamlConfigSource(cl.getResource(MICROPROFILE_CONFIG_YML_FILE), MICROPROFILE_CONFIG_YML_JAR_PRIORITY),
+                new YamlConfigSource(cl.getResource(APPLICATION_YML_FILE), APPLICATION_YML_JAR_PRIORITY),
+                new YamlConfigSource(getFileSystemURL(MICROPROFILE_CONFIG_YML_FILE),
```
gastaldi: Perhaps check if the URL is not null before adding to the result list?

gastaldi:
@gwenneg yes! Excellent! Added a few comments

extensions/config-yaml/runtime/src/main/java/io/quarkus/yaml/configuration/runtime/YamlConfigSourceProvider.java
gwenneg: Done.

extensions/config-yaml/runtime/src/main/java/io/quarkus/yaml/configuration/runtime/YamlConfigSourceProvider.java
gwenneg: Nothing, I was focused on the classloader logic coming from the initial version of the code and forgot it was already available as a parameter :)

gwenneg:
> @gwenneg yes! Excellent! Added a few comments

Thanks for your guidance! This PR looks much better now.

extensions/config-yaml/runtime/src/main/java/io/quarkus/yaml/configuration/runtime/YamlConfigSourceProvider.java
```diff
@@ -0,0 +1,61 @@
...
+        URL fsApplicationYml = getFileSystemURL(APPLICATION_YML_FILE);
+        if (fsApplicationYml != null) {
+            configSources.add(new YamlConfigSource(fsApplicationYml, APPLICATION_YML_FILESYSTEM_PRIORITY));
+        }
+
+        return configSources;
+    }
+
+    private URL getFileSystemURL(String source) {
+        Path path = Paths.get("config", source);
```
gastaldi: Not a big deal, but maybe we shouldn't assume that a `config` directory will exist in the same directory as the application

gastaldi (approved):
LGTM, added one final observation :wink:

extensions/config-yaml/runtime/src/main/java/io/quarkus/yaml/configuration/runtime/YamlConfigSourceProvider.java
geoand: Ideally yes, but I think that the `.properties` equivallent does assume this. @dmlloyd had some work in progress to make this configurable on the command line IIRC.

geoand:
I think that can be taken out of draft state once there is documentation in place :)

extensions/config-yaml/runtime/src/main/java/io/quarkus/yaml/configuration/runtime/YamlConfigSourceProvider.java
gwenneg: Yes, this is also true for the properties-based configuration and documented as well in the [configuration guide](https://github.com/quarkusio/quarkus/blob/master/docs/src/main/asciidoc/application-configuration-guide.adoc):

>using a configuration file placed in $PWD/config/application.properties
>
>By placing an application.properties file inside a directory named config which resides in the directory where the application runs, any runtime properties defined in that file will override the default configuration. Furthermore any runtime properties added to this file that were not part of the original application.properties file will also be taken into account.
>
> This works in the same way for runner jar and the native executable

devtools/common/src/main/filtered/extensions.json
```diff
@@ -743,5 +743,15 @@
...
+  },
+  {
+    "name": "YAML Configuration",
+    "labels": [
+      "yaml",
+      "config",
+      "configuration"
+    ],
+    "groupId": "io.quarkus",
+    "artifactId": "config-yaml"
```
gastaldi:
```suggestion
    "artifactId": "quarkus-config-yaml"
```

extensions/config-yaml/runtime/src/main/java/io/quarkus/yaml/configuration/runtime/YamlConfigConstants.java
```diff
@@ -0,0 +1,12 @@
+package io.quarkus.yaml.configuration.runtime;
+
+public class YamlConfigConstants {
+
+    public static final String APPLICATION_YML_FILE = "application.yml";
+    public static final String MICROPROFILE_CONFIG_YML_FILE = "microprofile-config.yml";
```
gastaldi: This file should be inside `META-INF`, to be close to the spec: https://github.com/eclipse/microprofile-config/#design

extensions/config-yaml/runtime/src/main/java/io/quarkus/yaml/configuration/runtime/YamlConfigSourceProvider.java
```diff
@@ -0,0 +1,61 @@
...
+        URL jarMicroprofileConfigYml = classLoader.getResource(MICROPROFILE_CONFIG_YML_FILE);
+        if (jarMicroprofileConfigYml != null) {
+            configSources.add(new YamlConfigSource(jarMicroprofileConfigYml, MICROPROFILE_CONFIG_YML_JAR_PRIORITY));
+        }
+        URL jarApplicationYml = classLoader.getResource(APPLICATION_YML_FILE);
+        if (jarApplicationYml != null) {
+            configSources.add(new YamlConfigSource(jarApplicationYml, APPLICATION_YML_JAR_PRIORITY));
+        }
+
+        URL fsMicroprofileConfigYml = getFileSystemURL(MICROPROFILE_CONFIG_YML_FILE);
```
gastaldi: In practice, I don't think people will use the `microprofile-config.yml` file externally, but that's me

gastaldi:
Also, should `.yaml` files also be supported or simply `.yml` is enough?

extensions/config-yaml/runtime/src/main/java/io/quarkus/yaml/configuration/runtime/YamlConfigSourceProvider.java
```diff
@@ -0,0 +1,61 @@
...
+import org.eclipse.microprofile.config.spi.ConfigSource;
+import org.eclipse.microprofile.config.spi.ConfigSourceProvider;
+
+public class YamlConfigSourceProvider implements ConfigSourceProvider {
+
+    @Override
+    public Iterable<ConfigSource> getConfigSources(ClassLoader classLoader) {
+        List<ConfigSource> configSources = new ArrayList<>();
+
+        URL jarMicroprofileConfigYml = classLoader.getResource(MICROPROFILE_CONFIG_YML_FILE);
```
gastaldi: Perhaps it would be nice to use `classLoader.getResources` to get the `application.yml` in other JARs as well?

extensions/config-yaml/runtime/src/main/java/io/quarkus/yaml/configuration/runtime/YamlConfigSourceProvider.java
gwenneg: It could happen for a project relying on a MicroProfile Config implementation and being migrated to Quarkus. And then, having `microprofile-config.yml` in the META-INF directory would definitely help, so I'm moving it as you suggested in your previous comment.

extensions/config-yaml/runtime/src/main/java/io/quarkus/yaml/configuration/runtime/YamlConfigSourceProvider.java
gastaldi: Right, unless the `config` directory is added to the classpath, then the `ClassLoader.getResource` above would grab it

kenfinnigan:
I'm going to repeat myself here, but I'm concerned that this is being developed separately from SmallRye Config.

I say that because once SmallRye Config has a Yaml ConfigSource, Quarkus is going to want to use it, but what kind of breaking changes, or user application changes, will that require because of decisions that have been taken in this PR?

gwenneg:
> Also, should `.yaml` files also be supported or simply `.yml` is enough?

Considering this extension will be documented in the configuration guide, I don't think we need to support `.yaml` (which also seems less common than `.yml`).

gastaldi:
> I'm going to repeat myself here, but I'm concerned that this is being developed separately from SmallRye Config.

Correct me if I am wrong but I believe the `YamlConfigSource` and `YamlConfigSourceProvider` will be extracted to a separate module  inside smallrye-config (https://github.com/smallrye/smallrye-config/tree/master/sources) and this extension will depend on that new source?

kenfinnigan:
Do you mean the classes from this PR? If so, then the answer is no.

The Yaml config source for SmallRye Config is coming from the MP Extensions project that is in the process of being donated

gastaldi:
I think it is based on MP Extensions, but without the mutability (which was seen as a bad idea by @dmlloyd in https://github.com/smallrye/smallrye-config/pull/165)

kenfinnigan:
From a quick look at the sources, I'm not sure they're anything alike

gwenneg:
The sources are indeed different since I tried to make the `YAML map to properties conversion` simpler, but would it be difficult to move from this PR to the MP Extensions implementation? Is there a specific point in this PR that looks like it would create a breaking change?

kenfinnigan:
I don't know, but there are a lot of decisions being made about file name, file type, file location, in this PR that I don't know if it's consistent with the MP Extension implementation.

If it isn't, then we end up with a nasty migration problem

extensions/config-yaml/runtime/src/main/java/io/quarkus/yaml/configuration/runtime/YamlConfigSource.java
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    private Map<String, String> buildConfigPropertiesFromYaml() {
+        if (source == null) {
+            return Collections.emptyMap();
+        }
+        try (InputStream is = source.openStream()) {
+            @SuppressWarnings("unchecked")
+            Map<String, Object> yamlContent = new Yaml().loadAs(is, HashMap.class);
+            Map<String, String> configProperties = new HashMap<>();
```
gastaldi:
```suggestion
            Map<String, String> configProperties = new TreeMap<>();
```

gastaldi:
If you look at the [MP Ext implementation](https://github.com/microprofile-extensions/config-ext/blob/master/configsource-yaml/src/main/java/org/microprofileext/config/source/yaml/YamlConfigSource.java) - which is only this class, there aren't any decisions regarding the file used. That's similar to [@gwenneg 's implementation](https://github.com/gwenneg/quarkus/blob/issue-3904-yaml-configuration/extensions/config-yaml/runtime/src/main/java/io/quarkus/yaml/configuration/runtime/YamlConfigSource.java) which also does not care about the file location

gwenneg:
> > Also, should `.yaml` files also be supported or simply `.yml` is enough?
> 
> Considering this extension will be documented in the configuration guide, I don't think we need to support `.yaml` (which also seems less common than `.yml`).

I retract this comment since the `.yaml` extension seems to be the one being used in the MP Extensions implementation. It this PR ends up being merged, it should use the same extension.

extensions/config-yaml/runtime/src/main/java/io/quarkus/yaml/configuration/runtime/YamlConfigSource.java
```diff
@@ -0,0 +1,127 @@
...
+
+    private Map<String, String> buildConfigPropertiesFromYaml() {
+        if (source == null) {
+            return Collections.emptyMap();
+        }
+        try (InputStream is = source.openStream()) {
+            @SuppressWarnings("unchecked")
+            Map<String, Object> yamlContent = new Yaml().loadAs(is, HashMap.class);
+            Map<String, String> configProperties = new TreeMap<>();
+            for (YamlConfigProperty prop : convertYamlMapToProperties("", yamlContent)) { // The root `baseKey` is empty.
```
gastaldi: I believe If you pass the `Map<String,String>` as a parameter, you can remove the `YamlConfigProperty` class

extensions/config-yaml/runtime/src/main/java/io/quarkus/yaml/configuration/runtime/YamlConfigSource.java
```diff
@@ -0,0 +1,127 @@
...
+        }
+    }
+
+    private Map<String, String> buildConfigPropertiesFromYaml() {
+        if (source == null) {
+            return Collections.emptyMap();
+        }
+        try (InputStream is = source.openStream()) {
+            @SuppressWarnings("unchecked")
+            Map<String, Object> yamlContent = new Yaml().loadAs(is, HashMap.class);
```
gastaldi:
```suggestion
            Map<String, Object> yamlContent = new Yaml().loadAs(is, TreeMap.class);
```

Minor lookup optimization

extensions/config-yaml/runtime/src/main/java/io/quarkus/yaml/configuration/runtime/YamlConfigSource.java
gwenneg: That's true, but I usually try to avoid using a method parameter to store/pass an output value, so I did this on purpose.

If you think passing the map would be better, I'll change this.

gwenneg:
Documentation changes pushed. I'm not sure whether I should mention the config sources priorities (list them all) in the `Custom Configuration` or not.

I'm removing the `coding-in-progress` label label since it's no longer true (still thinking about the classloader comments though) but I'm letting the PR in a draft status until its fate is decided. I obviously hope this will be merged but if it doesn't, I still had fun working on this! :)

extensions/config-yaml/runtime/src/main/java/io/quarkus/yaml/configuration/runtime/YamlConfigSource.java
gastaldi: No preference, but it seems that it would create less objects than necessary

gastaldi (approved):
Good job, this PR can now be promoted to a real PR instead of a Draft :smile:

gwenneg:
@kenfinnigan I took a deep look at [microprofile-extensions's YamlConfigSource](https://github.com/microprofile-extensions/config-ext/blob/master/configsource-yaml/src/main/java/org/microprofileext/config/source/yaml/YamlConfigSource.java) and its parent classes/interfaces and I don't see how the implementation from this PR would create any migration issue. You were right to raise a warning about the file extension though, that was an issue which is now fixed.

The current minimal Quarkus documentation only says two things:
- we're expecting a file located at `src\main\resources\application.yaml` which I believe would still be true after the migration
- `application.yaml` has a higher ordinal/priority than `application.properties` but this decision has nothing to do with the YAML config implementation

I also think it's good to remain silent in the Quarkus doc about the config sources and the `microprofile-config.yaml` support for now.

gsmet:
@kenfinnigan what is the ETA for having yaml in SmallRye Config? I think it's an important element for the decision making process.

kenfinnigan:
@gsmet it's dependent on when @phillip-kruger donates the code to SmallRye Config. He's going through it all piece by piece.

Realistically, probably a couple of weeks

phillip-kruger:
I can make sure thus is the next PR as soon as the currect one is accepted and merged

emmanuelbernard:
We can live without it in 1.0 IMO

extensions/config-yaml/runtime/src/main/resources/META-INF/quarkus-extension.yaml
```diff
@@ -0,0 +1,10 @@
+---
+name: "YAML Configuration"
+metadata:
+  keywords:
+  - "yaml"
+  - "config"
+  - "configuration"
+  categories:
+  - "config"
+  status: "stable"
```
gwenneg: This PR was just rebased onto master to solve conflicts. There's a good chance lines 8 to 10 are not correct.

djajcevic:
> We can live without it in 1.0 IMO

Depends when 1.0 will be released... :smile: 

Would be nice to use the existing config server (eq. spring config server) as config source (yaml or properties). I've played with it for some time and got it to work but with build time limitations (could not configure REST clients without local properties file).

geoand:
> Would be nice to use the existing config server (eq. spring config server) as config source (yaml or properties). 

This is something I've heard from others lately as well and I think it's something we should do soon.

djajcevic:
SCC server has the ability to provide aggregated properties/yaml files for profiles `http://spring-cloud.company.com/configuration-{profile}.yaml` and when that URL is provided to MP https://github.com/microprofile-extensions/config-ext/tree/master/configsource-yaml it works. Therefore we already have it - just not with an extension. But I don't know if waiting for SmallRye is an option if there's no clear date for its release. Having this just in documentation/guides can bring a lot of developers in.

For the end-users there should be no changes in terms of usage, so be it yaml or yml file extension is less relevant. Even though Quarkus is opinionated no one should argue about it, just support most variations.

dmlloyd:
This brings up three ideas:

* The idea of supporting YAML as a config format (obviously)
* Supporting URL locators for configuration sources (on the command line maybe?)
* Incorporating the profile name into the configuration file (something we discussed inconclusively elsewhere)

geoand:
Will this be able to be taken of ice once https://github.com/quarkusio/quarkus/pull/5387 is in?

gwenneg:
@geoand I don't think so, the `YamlConfigSource` donation PR hasn't been merged into `smallrye-config` yet: https://github.com/smallrye/smallrye-config/pull/174

geoand:
Thanks @gwenneg

PR #4527, opened 2019-10-12
Work around an issue in JDK which causes hot deployment compilation to fail

jaikiran:
The commit in this PR introduces a workaround to get past the issue noted in https://github.com/quarkusio/quarkus/issues/3592. The complete details of what's causing this issue is explained starting this comment here https://github.com/quarkusio/quarkus/issues/3592#issuecomment-541078904

The real fix should come in from https://bugs.openjdk.java.net/browse/JDK-8232170 but until then this should get us past it.

I've manually tested this issue on a Windows setup, both with JDK 8 and 11 and it used to fail (on JDK 11) without this fix. It now passes with this fix.

I don't see a way to add an automated test for this given the way we launch the tests and the different code path(s) that are involved in our testing infrastructure.

### DIFF
--- a/core/devmode/src/main/java/io/quarkus/dev/ClassLoaderCompiler.java
+++ b/core/devmode/src/main/java/io/quarkus/dev/ClassLoaderCompiler.java
@@ -83,6 +83,9 @@ public class ClassLoaderCompiler {
                 classPathElements.add(new File(i.getClassesPath()));
             }
         }
+        final String devModeRunnerJarAbsolutePath = context.getDevModeRunnerJarFile() == null
+                ? null
+                : context.getDevModeRunnerJarFile().getAbsolutePath();
         while (!toParse.isEmpty()) {
             String s = toParse.poll();
             if (!parsedFiles.contains(s)) {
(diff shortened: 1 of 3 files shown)


machi1990:
> I've manually tested this issue on a Windows setup, both with JDK 8 and 11 and it used to fail (on JDK 11) without this fix. It now passes with this fix.
> 
> I don't see a way to add an automated test for this given the way we launch the tests and the different code path(s) that are involved in our testing infrastructure.

We have tests for different JDK versions for Linux, maybe we should have a somehow similar strategy for windows? WDYT? @stuartwdouglas @geoand @gsmet

jaikiran:
> We have tests for different JDK versions for Linux, maybe we should have a somehow similar strategy for windows? WDYT? @stuartwdouglas @geoand @gsmet

Hello @machi1990, the problem isn't really about where the tests are run but more about how the `QuarkusDevModeTest` extension builds and runs the application under test. As far as I see, it doesn't (re)use the code in the `DevMojo` of Maven (which is where this whole classpath manifest gets setup) and instead uses its own code to come up with a way to trigger the `DevModeMain`. So even if I write a `QuarkusDevModeTest` and do all the hot deployment tests in it, it's not going to test the actual issue that this PR is for.

geoand:
Given the fact that the issue only occured on Windows and only on JDK 11+, I think it makes sense to have that setup as well

machi1990:
> > We have tests for different JDK versions for Linux, maybe we should have a somehow similar strategy for windows? WDYT? @stuartwdouglas @geoand @gsmet
> 
> Hello @machi1990, the problem isn't really about where the tests are run but more about how the `QuarkusDevModeTest` extension builds and runs the application under test. As far as I see, it doesn't (re)use the code in the `DevMojo` of Maven (which is where this whole classpath manifest gets setup) and instead uses its own code to come up with a way to trigger the `DevModeMain`. So even if I write a `QuarkusDevModeTest` and do all the hot deployment tests in it, it's not going to test the actual issue that this PR is for.

Hi @jaikiran my comment was more general and not to about writing a test for this PR. What I was thinking is maybe we can have a CI setup where we could also test different JDK versions for Window platform - at least the classic dev workflow [maven tests](https://github.com/quarkusio/quarkus/tree/master/integration-tests/maven).

gsmet:
@jaikiran could you rebase please?

I will refrain from merging this so late in the release cycle. I will merge it at the beginning of the next one so probably tomorrow or on Thursday.

stuartwdouglas (approved):

gsmet:
I force pushed a rebase, let's see what CI has to say.

Considering we had this issue reported so many times, Clément and I decided to include it in tomorrow's release.

jaikiran:
@gsmet, I have been away so didn't notice these messages and am just syncing up. I see that you have rebased and force pushed (thank you). Let me know if you need anything more from me.

gsmet:
@jaikiran nope, everything looks fine, thanks!

I would have waited but we really want that one in 0.25.0 so I pushed things a bit.

Thanks for taking the time to figure out this one, really helpful.

PR #4548, opened 2019-10-14
jackson-databind library dependency added

jithset:


cescoffier:
I may miss the context. Can you remind me of the reason for this addition?

jitihn:
adding jackson-databind  library is not mentioned in the guide.

It will be helpful for users to follow the guide step by step.

geoand:
Do the tests of the fail without this dependency?

gsmet (changes requested):
This PR requires a few wording changes and it looks like `nimbus-jose-jwt` is not used anymore and I would prefer getting rid of Jackson Databind.

See https://github.com/quarkusio/quarkus-quickstarts/issues/325

sberyozkin:
Yes, looks like it is not even used now...

gsmet:
This has been fixed another way. Thanks for the report.

PR #4549, opened 2019-10-14
Call the health check method on the worker pool

### LINKED ISSUE #4531: Health check called on the event loop
**Describe the bug**

With the following class:

```
@Readiness
@ApplicationScoped
public class DatabaseConnectionHealthCheck implements HealthCheck {

    @Inject
    FightService fightService;

    @Override
    public HealthCheckResponse call() {
        HealthCheckResponseBuilder responseBuilder = HealthCheckResponse.named("Database connection health check");

        System.out.println("Thread: " + Thread.currentThread().getName());
        try {
            List<Fight> fights = fightService.findAllFights();
            responseBuilder.withData("Number of rows in the database", fights.size()).up();
        } catch (IllegalStateException e) {
            e.printStackTrace();
            responseBuilder.down();
        }

        return responseBuilder.build();
    }
}
```

The call fails because it tries to access the database from an IO Thread.

**Expected behavior**
It should use a worker thread.

**Actual behavior**
It uses the event loop thread

cescoffier:
Fix #4531

### DIFF
--- a/extensions/smallrye-health/runtime/src/main/java/io/quarkus/smallrye/health/runtime/SmallRyeHealthHandler.java
+++ b/extensions/smallrye-health/runtime/src/main/java/io/quarkus/smallrye/health/runtime/SmallRyeHealthHandler.java
@@ -4,6 +4,7 @@ import java.io.ByteArrayOutputStream;
 
 import javax.enterprise.inject.spi.CDI;
 
+import io.quarkus.arc.Arc;
 import io.smallrye.health.SmallRyeHealth;
 import io.smallrye.health.SmallRyeHealthReporter;
 import io.vertx.core.Handler;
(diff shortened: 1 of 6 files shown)


kenfinnigan (dismissed):

extensions/smallrye-health/runtime/src/main/java/io/quarkus/smallrye/health/runtime/SmallRyeHealthHandler.java
```diff
@@ -17,7 +20,10 @@
 
     @Override
     public void handle(RoutingContext event) {
-
+        InjectableContext context = Arc.container().getActiveContext(RequestScoped.class);
```
kenfinnigan: @mkouba is this ok?

stuartwdouglas: Nope, it needs to be cleaned up afterwards. Also at the moment there is no way it can be active when this method is called.

gsmet (dismissed):
Just blocking this thing so that nobody merges it by mistake as it has been approved.

Please dismiss once everything is ready.

stuartwdouglas (approved):

extensions/smallrye-health/runtime/src/main/java/io/quarkus/smallrye/health/runtime/SmallRyeHealthHandler.java
mkouba: Also it's easier to use `io.quarkus.arc.ArcContainer.requestContext()` instead...

mkouba: Oh, and the method returns null if there is no request context active..

cescoffier: Fixed.

cescoffier:
@gsmet changes have been made.

Issue #4541, opened 2019-10-14
Error during WebSocket handshake: net::ERR_INVALID_HTTP_RESPONSE


cescoffier:
**Describe the bug**

WebSocket does not work anymore in 0.24.0

The following code:
```
 @OnOpen
    public void onOpen(Session session) throws IOException {
        session.getBasicRemote().sendText("23.0");
        session.getBasicRemote().sendText("1.0");
        session.getBasicRemote().sendText("99.0");
        session.getBasicRemote().sendText("10.0");
    }
```

triggers: ERR_INVALID_HTTP_RESPONSE

cescoffier:
Fun fact: it only happens from the browser.

PR #4550, opened 2019-10-14
Fix for #4541

stuartwdouglas:


gsmet:
@stuartwdouglas it has been fixed somehow and you're adding a test or we still need to fix it?

cescoffier:
@gsmet It's related to the issue I've opened this morning (#4541)

We found out that the issue only happen in browsers...

stuartwdouglas:
Turns out it is racey, I wanted to see if it failed in CI. Now that I can reproduce I will have a fix shortly

integration-tests/main/src/main/java/io/quarkus/it/websocket/WebSocketOpenEndpoint.java
```diff
@@ -0,0 +1,22 @@
...
+
+@ServerEndpoint("/wsopen")
+@ApplicationScoped
+public class WebSocketOpenEndpoint {
+
+    public static String[] messages = { "23.0", "1.0", "99.0", "10.0", "10.0", "10.0", "10.0", "10.0" };
+
+    @OnOpen
+    public void onOpen(Session session) throws IOException {
+        for (String i : messages) {
```
cescoffier: Because the registration of the handler happens after the opening, we may miss messages. That's expected.
We could modify the test to receive a first ping before sending the messages.

cescoffier (changes requested):

cescoffier (approved):

PR #4554, opened 2019-10-14
Only start Quartz when there are some @Scheduled annotations

### LINKED ISSUE #4243: Quartz dependency shouldn't start threads when it's not being used
**Describe the bug**
For some reason an application I was profiling had the scheduler extension listed among its dependencies.

This was likely a mistake as there is nothing being scheduled.

**Expected behavior**
The extension should be "harmless"

**Actual behavior**
A couple of scheduling related threads are being initialized, and are running in background: stealing CPU time and memory from the main app.

**Additional context**
This clearly isn't critical, but considering we do several optimisations based on the "closed world assumption" I believe Quarkus should be able to enable the scheduler only if it's actually being used.

SlyngDK:
…en there are some Schedules.

Fixes: #4243

### DIFF
--- a/extensions/scheduler/runtime/src/main/java/io/quarkus/scheduler/runtime/QuartzScheduler.java
+++ b/extensions/scheduler/runtime/src/main/java/io/quarkus/scheduler/runtime/QuartzScheduler.java
@@ -110,6 +110,10 @@ public class QuartzScheduler implements Scheduler {
     }
 
     void start(@Observes StartupEvent startupEvent) {
+        if (schedulerConfig.getSchedules().isEmpty()) {
+            LOGGER.warn("No @Scheduled methods found, scheduler will not be started.");
+            return;
+        }
         if (running.compareAndSet(false, true)) {
 
             try {


mkouba (approved):
Looks good.

extensions/scheduler/runtime/src/main/java/io/quarkus/scheduler/runtime/QuartzScheduler.java
```diff
@@ -110,6 +110,10 @@ public void startTimer(long delay, Runnable action) {
     }
 
     void start(@Observes StartupEvent startupEvent) {
+        if (schedulerConfig.getSchedules().isEmpty()) {
+            LOGGER.warn("No Schedules found, Scheduler will not be started.");
```
gsmet:
```suggestion
            LOGGER.warn("No @Scheduled annotations found, Scheduler will not be started.");
```

@mkouba that being said, I wonder if it should really be a warn?

gsmet (changes requested):

extensions/scheduler/runtime/src/main/java/io/quarkus/scheduler/runtime/QuartzScheduler.java
mkouba: I believe so. The extension should be removed if there is no scheduled method...

mkouba: Unless we provide a way to schedule a job programatically...

gsmet: @SlyngDK could you apply my wording suggestion and squash everything into a single commit? Thanks!

mkouba: Actually, it should probably read `No @Scheduled methods found`...

gsmet: Whatever makes you happy :)

gsmet:
I rebased and force-pushed an adjusted message to try to get that one into 0.25.0.

gsmet (approved):

jaikiran:
> Unless we provide a way to schedule a job programatically...
@mkouba, Looking at the code, it looks like the `io.quarkus.scheduler.Scheduler` can be injected and can be used programatically to schedule jobs https://github.com/quarkusio/quarkus/blob/master/extensions/scheduler/runtime/src/main/java/io/quarkus/scheduler/Scheduler.java.

The implementation of these APIs, require the scheduler to be started before calling something like `startTimer`. So if we don't start scheduler, in the absence of a `@Schedule`, I think that would mean the scheduler can no longer be used programatically.

gsmet:
CI failures are unrelated, merging.

gsmet:
Thanks @SlyngDK !

mkouba:
@jaikiran I've completely forgotten about this one! In theory, we could detect `Scheduler` injection points too. In which case the only problematic scenario would be "programmatic lookup of a `Scheduler` + programmatic scheduling" combo.

jaikiran:
> In theory, we could detect Scheduler injection points too.
Yes, that's a good point.
> In which case the only problematic scenario would be "programmatic lookup of a Scheduler + programmatic scheduling" combo.
So I had a look at what prompted this change and it appears to be this enhancement suggestion (a valid one IMO) - https://github.com/quarkusio/quarkus/issues/4243. I think if that has to be solved and yet be able to start the scheduler eagerly, perhaps a better way would be to implement a (lazy?) `org.quartz.spi.ThreadPool` which doesn't start those threads eagerly.

mkouba:
@jaikiran Hm, maybe we should reuse the internal thread pool. Pls file a new tracking issue.

jaikiran:
@mkouba Done - https://github.com/quarkusio/quarkus/issues/4606

Issue #4345, opened 2019-10-03
Optimise RESTEasy for closed world


Sanne:
Profling a Quarkus application it's immediatley visible that the current RESTEasy extension isn't taking advantage of the closed world assumption:

Annotation scanning and type introspection:

```
Stack Trace	TLABs	Total TLAB Size(bytes)	Pressure(%)
java.lang.reflect.Method.getParameterTypes()	28	137,118,288	6.174
   sun.reflect.annotation.AnnotationInvocationHandler.invoke(Object, Method, Object[])	28	137,118,288	6.174
      com.sun.proxy.$Proxy17.annotationType()	16	77,359,528	3.483
         org.jboss.resteasy.spi.util.FindAnnotation.findAnnotation(Annotation[], Class)	16	77,359,528	3.483
            org.jboss.resteasy.plugins.providers.sse.SseEventSinkInterceptor.filter(ContainerRequestContext)	16	77,359,528	3.483
```

And more..
```
Stack Trace	TLABs	Total TLAB Size(bytes)	Pressure(%)
java.lang.Class.getInterfaces()	425	2,083,949,208	93.826
   org.jboss.resteasy.core.providerfactory.Utils.createHeaderDelegateFromInterfaces(Map, Class[])	219	1,071,276,576	48.233
   org.jboss.resteasy.core.providerfactory.Utils.createHeaderDelegate(Map, Class)	206	1,012,672,632	45.594
```

I suppose this implies we can improve it a lot still?

The above methods have been identified when looking for strong allocators; this implies memory consumption could be cut down by dodging these operations.

Sanne:
@patriot1burke could you take this or assign someone?

emmanuelbernard:
I read it as RESTEasy reads the annotation on the method being call for every request without even caching anything or preparing its metadata? Is that correct?

Sanne:
> I read it as RESTEasy reads the annotation on the method being call for every request without even caching anything or preparing its metadata? Is that correct?

Correct, these are profiling snapshots taken at runtime: under load, after the application booted and after it completed warmup.

Sanne:
@geoand assiging this to you :)

@asoldano could you help / mentor ? We might need some fixes from resteasy too

geoand:
For the JSON part of things, I'll try and revive #3553 which has fallen far behind the latest RESTEasy + JSON-B developments.

asoldano:
[...]

PR #4563, opened 2019-10-15
Introduce caching of ResteasyUriInfo.InitData

geoand:
This is done because the computation of the data is expensive

Relates to: #4345
~~Depends on: https://github.com/resteasy/Resteasy/pull/2184~~
~~Depends on: RESTEasy `4.4.0.Final` (still unreleased)~~

### DIFF
--- a/extensions/resteasy/runtime/src/main/java/io/quarkus/resteasy/runtime/standalone/VertxRequestHandler.java
+++ b/extensions/resteasy/runtime/src/main/java/io/quarkus/resteasy/runtime/standalone/VertxRequestHandler.java
@@ -2,6 +2,8 @@ package io.quarkus.resteasy.runtime.standalone;
 
 import java.io.IOException;
 import java.io.InputStream;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 
 import javax.enterprise.inject.Instance;
 import javax.enterprise.inject.spi.CDI;
@@ -35,20 +37,25 @@ public class VertxRequestHandler implements Handler<RoutingContext> {
     protected final Vertx vertx;
     protected final RequestDispatcher dispatcher;
     protected final String rootPath;
+    protected final int resteasyUriInfoCacheMaxSize;
     protected final BufferAllocator allocator;
     protected final BeanContainer beanContainer;
     protected final CurrentIdentityAssociation association;
+
+    private final Map<String, ResteasyUriInfo.InitData> resteasyUriInfoInitDataMap = new ConcurrentHashMap<>();
 
     public VertxRequestHandler(Vertx vertx,
             BeanContainer beanContainer,
             ResteasyDeployment deployment,
             String rootPath,
+            int resteasyUriInfoCacheMaxSize,
             BufferAllocator allocator) {
         this.vertx = vertx;
         this.beanContainer = beanContainer;
         this.dispatcher = new RequestDispatcher((SynchronousDispatcher) deployment.getDispatcher(),
                 deployment.getProviderFactory(), null);
         this.rootPath = rootPath;
+        this.resteasyUriInfoCacheMaxSize = resteasyUriInfoCacheMaxSize;
         this.allocator = allocator;
         Instance<CurrentIdentityAssociation> association = CDI.current().select(CurrentIdentityAssociation.class);
         this.association = association.isResolvable() ? association.get() : null;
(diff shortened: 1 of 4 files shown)


geoand:
@asoldano, @Sanne this is what I had in mind based on our discussion in #4345. 

WDYT?

extensions/resteasy/runtime/src/main/java/io/quarkus/resteasy/runtime/standalone/VertxRequestHandler.java
```diff
@@ -114,6 +136,19 @@ private void dispatch(RoutingContext routingContext, InputStream is, VertxOutput
             if (!vertxRequest.getAsyncContext().isSuspended()) {
                 try {
                     vertxResponse.finish();
+                    // the reason we only cache successful responses is to ensure that a torrent of erroneous URLs
+                    // doesn't fill up the cache and cause a DoS
```
Sanne: I think you should cache unsuccesfull responses as well, especially if we go for having a proper cache.

DDoS are caused by the system being overwhelmed; hitting the "bad response cache" would be the best way to avoid it as the cache would help fighting load.

extensions/resteasy/runtime/src/main/java/io/quarkus/resteasy/runtime/standalone/VertxRequestHandler.java
```diff
@@ -114,6 +136,19 @@ private void dispatch(RoutingContext routingContext, InputStream is, VertxOutput
...
+                    if (setInitDataUponSuccess && vertxResponse.getStatus() >= 200 && vertxResponse.getStatus() <= 300) {
+                        // we don't want the cache to grow unbounded since values path params could potentially be infinite
+                        // and if left unchecked would take up all memory if the application is up for long enough
+                        if (resteasyUriInfoInitDataMap.size() > resteasyUriInfoCacheMaxSize) {
+                            resteasyUriInfoInitDataMap.clear(); // this is super lame and should probably be revisited
```
Sanne: yes it's dangerous. See my other comments on the design discussion.

extensions/resteasy/runtime/src/main/java/io/quarkus/resteasy/runtime/standalone/VertxRequestHandler.java
geoand: I agree 💯%. If we have a proper cache than all this nastiness can go away.

extensions/resteasy/runtime/src/main/java/io/quarkus/resteasy/runtime/standalone/VertxRequestHandler.java
geoand: If we have the RESTEasy part of the PR in which puts in the necessary hooks, we can absolutely improve this based on what you mention about the existence or not of Caffeine.
Note that the way I have done the RESTEasy PR (and as of course you have seen from looking at this PR), the caching itself is handled exclusively in Quarkus so we can probably get as creative as we like.

extensions/resteasy/runtime/src/main/java/io/quarkus/resteasy/runtime/standalone/VertxRequestHandler.java
```diff
@@ -35,20 +37,25 @@
     protected final Vertx vertx;
     protected final RequestDispatcher dispatcher;
     protected final String servletMappingPrefix;
+    protected final int resteasyUriInfoCacheMaxSize;
     protected final BufferAllocator allocator;
     protected final BeanContainer beanContainer;
     protected final CurrentIdentityAssociation association;
 
+    private final Map<String, ResteasyUriInfo.InitData> resteasyUriInfoInitDataMap = new ConcurrentHashMap<>();;
```
gsmet: Minor nitpick: you have two semi-colon.

geoand: LOL, I most certainly do, good find! I'll fix it soon

geoand: Fixed

Sanne:
@geoand what's the status of this? I guess it might be too late but I'd still love to get some improvements in this area.

geoand:
@Sanne I can rebase onto `master` ready for merging if we feel it should be added.

geoand:
I didn't do it so far because this wasn't critical, but it won't be a problem to update it (it's a temporary solution anyway until we have a better "cache")

Sanne:
@geoand yes please do :) And ping me for review, I can quickly run some perf tests.

geoand:
@Sanne can you give it a shot and see if it gives you any significant improvement?

gsmet:
I'm cancelling the build for this one.

geoand:
Sure, no problem

Sanne:
@geoand my results are being puzzling so far. I'll need to repeat the runs and gather some more data.

geoand:
OK, thanks for keeping me up to date

geoand:
@Sanne were you able to see if this change actually had much of an effect?

Sanne:
It actually made performance a little worse :-)  But I haven't been able to check why though... I will try diagnosing this today, as the patch looks sensible but maybe we're missing something.

geoand:
Interesting! If you find anything let me know so we can see how we can improve the situation.

Thanks!

Sanne:
Sure, I will. We need some luck for me to be able to focus on this long enough though, I'm using it as a "fun task" to do between priority tasks  ;)

geoand:
I see :)

gastaldi:
@geoand Is this PR still being worked on or can it be safely closed?

geoand:
Let's close it

PR #4567, opened 2019-10-15
Make InterceptorBindingRegistrarBuildItem final

### LINKED ISSUE #4565: BuildStep utilizing InterceptorBindingRegistrarBuildItem results in IllegalArgumentException
**Describe the bug**

I followed the example from the CDI guide for implementing [additional interceptor bindings](https://quarkus.io/guides/cdi-reference#additional-interceptor-bindings). It fails with:

```
java.lang.IllegalArgumentException: Build item class must be leaf (final) types: class io.quarkus.arc.deployment.InterceptorBindingRegistrarBuildItem
	at io.quarkus.builder.item.BuildItem.<init>(BuildItem.java:22)
	at io.quarkus.builder.item.MultiBuildItem.<init>(MultiBuildItem.java:8)
	at io.quarkus.arc.deployment.InterceptorBindingRegistrarBuildItem.<init>(InterceptorBindingRegistrarBuildItem.java:9)
```

Presumably `InterceptorBindingRegistrarBuildItem` should be declared as `final` like the other `BuildItem` implementations?

**Expected behavior**

`InterceptorBindingRegistrarBuildItem` successfully registers the interceptor binding.

gsmet:
A build item has to be final.

Fixes #4565

### DIFF
--- a/extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/InterceptorBindingRegistrarBuildItem.java
+++ b/extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/InterceptorBindingRegistrarBuildItem.java
@@ -3,7 +3,7 @@ package io.quarkus.arc.deployment;
 import io.quarkus.arc.processor.InterceptorBindingRegistrar;
 import io.quarkus.builder.item.MultiBuildItem;
 
-public class InterceptorBindingRegistrarBuildItem extends MultiBuildItem {
+public final class InterceptorBindingRegistrarBuildItem extends MultiBuildItem {
     private final InterceptorBindingRegistrar registrar;
 
     public InterceptorBindingRegistrarBuildItem(InterceptorBindingRegistrar registrar) {


gsmet:
@manovotn @mkouba we should probably have a test for that? I'm creating a PR with just the fix to get that one into 0.25.0 as a user reported it. Could you please work on adding a test?

mkouba:
@dmlloyd I thought that we validate the build item classes in the annotation processor, or?

dmlloyd:
There's validation in the processor and also when loading.  But I think it would be reasonable to add this validation to the annotation processor.  I don't think it would be able to exhaustively analyze every build item, but it would definitely be able to check every build item that is *used* by a build step.

gsmet:
Could we get this in anyway? It would be nice to unblock the user right away.

mkouba (approved):
I think we can merge this PR now and add a test later. I'll create a separate issue.

mkouba:
Although, without the test we can't be sure it really works ;-).

gsmet:
Let's merge then :).

PR #4577, opened 2019-10-15
Cleanup and optimize InvocationContext implementation

### LINKED ISSUE #4552: Cleanup and optimize InvocationContext implementation
- allocate less `InvocationContext` instances
- cache interceptor bindings
- refactor the code

mkouba:
- resolves #4552

A simple (and naive) [microbenchmark](https://github.com/mkouba/arc-benchmarks) shows that we can get almost 5x bigger throughput compared to 0.24.0. Also the number of allocations should be significantly reduced for each intercepted method invocation.


independent-projects/arc/runtime/src/main/java/io/quarkus/arc/LifecycleCallbackInvocationContext.java
```diff
@@ -0,0 +1,64 @@
...
+import java.lang.reflect.InvocationTargetException;
+import java.util.List;
+import java.util.Set;
+
+/**
+ * A simple InvocationContext implementation used for PostConstruct and PreDestroy callbacks.
+ */
+class LifecycleCallbackInvocationContext extends AbstractInvocationContext {
+
+    private volatile int position;
```
Sanne: why the need for `volatile` here?

I'd assume this class is not meant to be threadsafe; if it's meant to, perhaps it should use AtomicInteger instead?

independent-projects/arc/runtime/src/main/java/io/quarkus/arc/LifecycleCallbackInvocationContext.java
```diff
@@ -0,0 +1,64 @@
...
+    protected void interceptorChainCompleted() throws Exception {
+        // No-op
+    }
+
+    protected Object invokeNext() throws Exception {
+        int oldPosition = position;
+        try {
+            return chain.get(position++).invoke(this);
+        } finally {
+            position = oldPosition;
```
Sanne: Do you expect oldPosition to be different than position-- in some cases?

independent-projects/arc/runtime/src/main/java/io/quarkus/arc/AbstractInvocationContext.java
```diff
@@ -52,32 +46,36 @@ protected AbstractInvocationContext(Object target, Method method, Function<Invoc
...
     @Override
-    public void setParameters(Object[] params) throws IllegalStateException, IllegalArgumentException {
-        if (parameters == null) {
-            throw new IllegalStateException();
+    public void setParameters(Object[] params) {
+        int newParametersCount = Objects.requireNonNull(params).length;
+        Class<?>[] parameterTypes = method.getParameterTypes();
+        if (parameterTypes.length != newParametersCount) {
+            throw new IllegalArgumentException(
+                    "Wrong number of parameters - method has " + Arrays.toString(parameterTypes) + ", attempting to set "
```
Sanne: I think it might help to have some quotes around the parameter types (both of them) ?

independent-projects/arc/runtime/src/main/java/io/quarkus/arc/AbstractInvocationContext.java
```diff
@@ -3,41 +3,35 @@
...
+import java.util.Objects;
 import java.util.Set;
-import java.util.function.Function;
 import javax.interceptor.InvocationContext;
 
-public abstract class AbstractInvocationContext implements InvocationContext {
+abstract class AbstractInvocationContext implements InvocationContext {
 
-    protected Map<String, Object> contextData;
+    protected volatile Map<String, Object> contextData;
```
Sanne: Could you add some comments on the need for `volatile` ? 

maybe it's just me being unfamiliar with CDI, but it would highly help me understand the possibilities here with some more code comments :)

Sanne:
I've added some comments hoping they could be useful, but this is not my area of expertise so I'll leave it for others to do a real review. Feel free to ignore me, I'm mostly trying to understand this :)

independent-projects/arc/runtime/src/main/java/io/quarkus/arc/AbstractInvocationContext.java
mkouba: Well, the `contextData` is initialized lazily, when needed (as it's usually not needed), see the `newContextData()` method, and since it could be accessed from a different thread (see my email about MP FT requirements/problems) we need to make sure it's safely propagated. I'll try to update the javadoc to make it more clear.

independent-projects/arc/runtime/src/main/java/io/quarkus/arc/AbstractInvocationContext.java
mkouba: `Arrays.toString()` is using `[]` so I don't think it's necessary.

independent-projects/arc/runtime/src/main/java/io/quarkus/arc/LifecycleCallbackInvocationContext.java
mkouba: That's a good question. In theory, `position--` should work.

independent-projects/arc/runtime/src/main/java/io/quarkus/arc/LifecycleCallbackInvocationContext.java
mkouba: So no impl of IC in ArC is meant to be threadsafe but it should be possible to continue interception on a different thread, ie. we should safely propagate the values. 

WRT `AtomicInteger` - I'm personally a fan of atomic constructs but there are other team members who say that the fewer allocations the better ;-).

mkouba:
Thanks @Sanne!

independent-projects/arc/runtime/src/main/java/io/quarkus/arc/AbstractInvocationContext.java
manovotn: > we need to make sure it's safely propagated.

The current implementation makes the around invoke (the only candidate to be passed around multiple threads) create a new instance in each step, hence the different thread will get a new instance of IC, which just gets the map as a parameter. I am probably missing some obvious bit as for why we need volatile?

independent-projects/arc/runtime/src/main/java/io/quarkus/arc/InvocationContexts.java
```diff
@@ -0,0 +1,72 @@
...
+import java.lang.reflect.Method;
+import java.util.List;
+import java.util.Set;
+import java.util.function.Function;
+import java.util.function.Supplier;
+import javax.interceptor.InvocationContext;
+
+public final class InvocationContexts {
+
+    public static final String KEY_INTERCEPTOR_BINDINGS = "io.quarkus.arc.interceptorBindings";
```
manovotn: Shouldn't we place this on some more "API/SPI looking" class? The methods on this class are otherwise purely internal whereas this key can be used by integrators (the same way that they can use the util class).
How about we put that inside `extensions/arc/runtime/src/main/java/io/quarkus/arc/runtime/InterceptorBindings.java`?

independent-projects/arc/runtime/src/main/java/io/quarkus/arc/LifecycleCallbackInvocationContext.java
manovotn: > it should be possible to continue interception on a different thread

I thought we wanted that only for around invoke so as to minimize new object creation during interceptions?

manovotn (commented):
I really like the optimizations, thanks for finding the time to do that! :)

independent-projects/arc/runtime/src/main/java/io/quarkus/arc/LifecycleCallbackInvocationContext.java
manovotn: Why do you need to set this back to old position? Once we invoke the next lifecycle interceptor, then the position information in here is no longer important. What am I missing?

independent-projects/arc/runtime/src/main/java/io/quarkus/arc/InvocationContexts.java
mkouba: So the problem with the `io.quarkus.arc.runtime` package is that it should only contain runtime internals. The API stuff should be located in `io.quarkus.arc` but we already use this package in the ArC project itself. And that brings me to the fact that this key is not specific for Quarkus. In theory, we can introduce a new interface similar to `WeldInvocationContext`, i.e. `ArcInvocationContext` and put the key there.

independent-projects/arc/runtime/src/main/java/io/quarkus/arc/LifecycleCallbackInvocationContext.java
mkouba: It is important. Note that when an exception is thrown an interceptor should be able to recover by calling the `InvocationContext.proceed()` method. In other words, we have to execute the specific part of the chain again.

independent-projects/arc/runtime/src/main/java/io/quarkus/arc/AbstractInvocationContext.java
mkouba: You're probably right. I was thinking about accessing the context data from the "previous" IC instance which woudln't work anyway (neither works in Weld imho):
```
--> INT1 - proceed() 
  --> INT2 - proceed() and set some context data
    --> invoke target method
  --> INT2 return from the `@AroundInvoke` method
--> INT1 - access context data after proceed()
```

independent-projects/arc/runtime/src/main/java/io/quarkus/arc/LifecycleCallbackInvocationContext.java
mkouba: That's right. For now we only need this for AroundInvoke interceptors. I'll remove the `volatile` modifier and make it clear in javadoc that all lifecycle callback interceptors must be invoked on the same thread.

independent-projects/arc/runtime/src/main/java/io/quarkus/arc/AbstractInvocationContext.java
manovotn: Hmm, I'd guess that `INT1` will only see whatever context data was added before it was invoked or what `INT1` itself added, right?
Then again, looking at the code, we just pass around the same map, so _theoretically_ you could see that? Might be worth trying out...

independent-projects/arc/runtime/src/main/java/io/quarkus/arc/InvocationContexts.java
manovotn: It's probably not worth it for a single key but I get your point.

independent-projects/arc/runtime/src/main/java/io/quarkus/arc/LifecycleCallbackInvocationContext.java
manovotn: Ah true, thanks for the reminder.

independent-projects/arc/runtime/src/main/java/io/quarkus/arc/AbstractInvocationContext.java
dmlloyd: You don't need volatile to continue execution in a different thread.  In order for an object X to be "moved" from thread 1 to thread 2, the object X must be safely published.  Publishing X implicitly publishes everything that was written before doing so, including indexes etc.

You'd only need `volatile` or atomic constructs if thread 1 continued to access object X *after* it was published to thread 2.  A clean handoff means you don't need any other publication or synchronization primitives.

independent-projects/arc/runtime/src/main/java/io/quarkus/arc/LifecycleCallbackInvocationContext.java
Sanne: > I'm personally a fan of atomic constructs but there are other team members who say that the fewer allocations the better ;-).

The fewer allocations the better! But the increment in this case is not atomic, so if it's meant to work across threads the current pattern isn't safe enough.

dmlloyd: As I stated elsewhere, you can definitely safely continue invocation in another thread without `volatile`.

mkouba: > But the increment in this case is not atomic, so if it's meant to work across threads the current pattern isn't safe enough.

We don't need atomicity. The IC must not be used concurrently.

independent-projects/arc/runtime/src/main/java/io/quarkus/arc/AbstractInvocationContext.java
mkouba: > Hmm, I'd guess that INT1 will only see whatever context data was added before it was invoked or what INT1 itself added, right?

@manovotn Yes. And I think we should fix it.

> Then again, looking at the code, we just pass around the same map, so theoretically you could see that?

I don't think so. `IC1.contextData` is null.

> You'd only need volatile or atomic constructs if thread 1 continued to access object X after it was published to thread 2...

@dmlloyd thanks for explanation David. I think that our use case is a bit more complicated, ie. I thought we will need to access object X after it was published to thread 2. But I need to think about it.

manovotn (approved):

mkouba:
The failing test is very likely unrelated: https://github.com/quarkusio/quarkus/issues/4626

PR #4580, opened 2019-10-15
Fix #4562

### LINKED ISSUE #4562: SwaggerUI not displaying new resources
**Describe the bug**

* Create a project and run it in dev mode
* Check the swagger ui (http://localhost:8080/swagger-ui/#/default)
* Add a new resource with at least one endpoint
* Refresh the swagger-ui page
* The new endpoint is not displayed

Even new methods added to existing resources are not displayed.

cescoffier:
It's an ugly hack but after discussion it's a quick fix that will be revisited later.
Explanation are in the code.

Also add hot reload tests


stuartwdouglas (approved):

gsmet:
Just force pushed to fix the imports order.

PR #4583, opened 2019-10-15
Update REST Assured version in build-parent pom.xml

gsmet:
It was updated in the BOM but not there.


gsmet:
Let's push this in 0.26.0 to be on the safe side.

geoand (approved):

PR #4424, opened 2019-10-07
Initial copy of Derby extension

aguibert:
Working directly with @Sanne on this, but anyone is welcome to commend and discuss!

Fixes #4442

PR #4585, opened 2019-10-15
Add doc for using Derby

aguibert:
Forgot to add doc under PR #4424 so I'm adding it in a follow-on PR here.


gsmet (approved):
Perfect, thanks!

PR #4592, opened 2019-10-15
Always specify the project version when building the Gradle plugin

gastaldi:
This avoids errors during the release. 
@geoand @gsmet please review

### DIFF
--- a/devtools/gradle/gradle.properties
+++ b/devtools/gradle/gradle.properties
@@ -1,1 +1,0 @@
-version = 999-SNAPSHOT
--- a/devtools/gradle/pom.xml
+++ b/devtools/gradle/pom.xml
@@ -120,6 +120,7 @@
                                 <argument>${gradle.task}</argument>
                                 <argument>-Pdescription=${project.description}</argument>
                                 <argument>-S</argument>
+                                <argument>-Pversion=${project.version}</argument>
                             </arguments>
                             <environmentVariables>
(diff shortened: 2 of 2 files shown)


geoand:
I'll test it in the morning, but it looks very reasonable.

gsmet:
We have to check with @aloubyansky as he introduced this file.

geoand:
I tried it and it works. So it's a +1 from me but let's let @aloubyansky have the final say.

devtools/gradle/gradle.properties
```diff
@@ -1 +0,0 @@
-version = 999-SNAPSHOT
```
aloubyansky: Have you tried building the plugin the gradle way?

aloubyansky: Also if you import it into the IDE as a Gradle project, won't it complain that the version is missing?

gastaldi: Yes, you're right, it won't work when building using Gradle. Since the release uses Maven, I guess the `-Pversion` should override this value, right? In this case I'll add the properties file back but keep the parameter in the pom.xml, WDYT?

aloubyansky: I'll let you test and confirm that. BTW, I am not sure what the actual issue you fixing here.

gastaldi:
Closing as this didn't impact the release and Maven workarounds will probably be gone when migrating the plugin to a Gradle-only repository

PR #4598, opened 2019-10-16
Disable a racey test

cescoffier:
Disable a test that is racey because the message handler may be registered after the messages are sent.


gsmet (approved):

gsmet:
Fast tracking it as we want to release 0.25.0 now.

PR #4600, opened 2019-10-16
Fix gradle - junit5 integration templates and instructions

geoand:


devtools/common/src/main/resources/templates/basic-rest/kotlin/build.gradle-template.ftl
```diff
@@ -32,6 +32,11 @@ dependencies {
 group '${project_groupId}'
 version '${project_version}'
 
+test {
+    useJUnitPlatform()
+    exclude '**/Native*'
```
gsmet: Don't we need a way to run the native tests somehow?

geoand: Yes we do (I'm currently looking into that). However if one doesn't exclude the native tests the test job fails completely.

I'll update the PR when I have a way to run the native tests as well.

geoand: @gsmet when you say different directories you mean still under `src/test` or having a completely new source set like `src/it` (which seems to be the gradle way of doign things)?

gsmet: I remember seeing someone doing exactly that. Let me try to find it.

geoand: I'm working on it. I have it half working, but let's see how the other half goes :)

gsmet: I can't find it. But the idea IMHO is too be in line with the Gradle way. We need to be sure we can extend one test from an IT though, not sure how the test classpaths are built.

geoand: It looks like there is a general issue with the native tests and gradle... Haven't pinpointed it yet, I'll update when I do.

geoand: The classpath thing I think I have figured out. The actual test execution is causing a lot problems

geoand: I think I've come up with a proper solution, I'll update shortly

geoand: I added a new commit that adds a source set. I tested on both Java and Kotlin projects and the tests were ran correctly.

geoand: Actually it's not ready... Seems like something is missing...

geoand:
@gsmet This should be ready now.
@gastaldi Mind giving it a spin as well?

geoand:
The regular tests can be run using `./gradlew test` while the native tests run using `./gradlew check` (or `./gradle nativeTest`)

gastaldi:
It would be nice if the CreateProject also generated a README.md file with instructions on how to build the project.

gastaldi:
There you go: https://github.com/quarkusio/quarkus/issues/4608

geoand:
Nice idea!

gastaldi:
@geoand it looks like the native image is always built when you `gradle build` the project, is that intended?

gastaldi:
Not opposed to any of those options btw

geoand:
@Sanne WDYT? Is the more Gradle friendly option to hide tasks behind flags or to leave it up to the users to execute `-x ....` for whatever they don't want?

gastaldi:
btw, `nativeTest` should be renamed to `testNative` to match `buildNative` :)

devtools/common/src/main/resources/templates/basic-rest/java/build.gradle-template.ftl
```diff
@@ -35,5 +52,22 @@ compileJava {
...
+    useJUnitPlatform()
+    description = 'Runs native image tests'
+    group = 'verification'
+
+    testClassesDirs = sourceSets.nativeTest.output.classesDirs
+    classpath = sourceSets.nativeTest.runtimeClasspath
+    shouldRunAfter test
+}
+
+testNative.dependsOn buildNative
```
Sanne: I suspect ideally such "dependsOn" links could be made implicit?

geoand: What do you have in mind? I am sure you have some cool Gradle example I am missing 😉

geoand: Actually it might make sense to have this task provided by the plugin itself. WDYT?

Sanne: Yes that's what I mean. Sorry actually not having an example at hand, I'm a bit rusty on plugin development (and Gradle 5 updated many APIs, haven't used it yet).

But conceptually, such things should be setup by the plugin. People don't need to know, so they won't screw it up ;)

geoand: Makes sense! I'll check it out 👍

geoand: Everything has now been moved into the plugin.

geoand:
@Sanne @gastaldi @gsmet The PR has now been updated to generate a clean `build.gradle` file.
The configuration logic that was needed to set things up has now been moved into the Gradle plugin itself.

devtools/common/src/main/resources/templates/basic-rest/java/build.gradle-template.ftl
```diff
@@ -26,6 +26,9 @@ dependencies {
 
     testImplementation 'io.quarkus:quarkus-junit5'
     testImplementation 'io.rest-assured:rest-assured'
+
+    nativeTestImplementation 'io.quarkus:quarkus-junit5'
```
gastaldi: Can we rename to `testNativeImplementation`? Also rename other occurrences to `testNative` instead of `nativeTest`?

geoand: Yeah, that makes sense.

devtools/common/src/main/java/io/quarkus/generators/rest/BasicRestProjectGenerator.java
```diff
@@ -76,13 +77,19 @@ private boolean initProject() throws IOException {
 
             srcMainPath = writer.mkdirs(type.getSrcDir());
             testMainPath = writer.mkdirs(type.getTestSrcDir());
+            // for gradle we want to place the native tests in under 'src/nativeTest/java'
```
gsmet: camel case for paths is not very appealing, I think I would prefer `native-test`.

geoand: Sure, let's do that

devtools/common/src/main/resources/templates/basic-rest/java/build.gradle-template.ftl
geoand: Fixed

devtools/common/src/main/java/io/quarkus/generators/rest/BasicRestProjectGenerator.java
geoand: Fixed

geoand:
Arg.... Gradle has more surprises in store... Something is not working after the rename :man_facepalming:

geoand:
@gastaldi actually using `testNative...` for the source set is now working. It's probably some obscure gradle issue. I propose we don't spend any more time on it and get this in as is since currently generated gradle projects can't run tests.

test-framework/common/src/main/java/io/quarkus/test/common/PathTestHelper.java
```diff
@@ -21,12 +21,21 @@
                 "bin" + File.separator + "test",
                 "bin" + File.separator + "main");
         // gradle
+        TEST_TO_MAIN_DIR_FRAGMENTS.put(
+                "classes" + File.separator + "java" + File.separator + "nativeTest",
```
gastaldi: This should be `native-test` no?

geoand: Yup. I'm about to push changes

geoand: Done

geoand:
OK, the final version has been pushed

docs/src/main/asciidoc/gradle-tooling.adoc
```diff
@@ -126,23 +126,9 @@ dependencies {
...
-
-[source,kotlin,subs=attributes+]
-----
-tasks.test {
-    useJUnitPlatform()
-}
-----
+When the Quarkus Gradle plugin is applied to a build, the `test` task is configured to run regular tests while the `testNative` task
+is meant to run the native image tests.
+The source code for the native image tests should reside in `src/nativeTest` instead of `src/test` where the source code for regular tests resides.
```
gsmet:
```suggestion
The source code for the native image tests should reside in `src/native-test` instead of `src/test` where the source code for regular tests resides.
```

gsmet (changes requested):
Can you take care of my comment?

docs/src/main/asciidoc/gradle-tooling.adoc
geoand: Ah darn, I missed it, sorry!

geoand: Fixed, thanks

gsmet (approved):

PR #4603, opened 2019-10-16
opentracing: Update to the instructions to reflect that agent (via UDP) can now be used

objectiser:
…P) can now be used

Signed-off-by: Gary Brown <gary@brownuk.com>


gsmet (approved):
Looks good to me thanks. Can you update the quickstart accordingly if it's needed? https://github.com/quarkusio/quarkus-quickstarts/tree/development/using-opentracing

objectiser:
@gsmet I've tested the instructions in the quickstart - but one question I was going to ask - the quickstart has been updated to include a chain request (aswell as the hello request) - which results in a call out back to the hello endpoint. So it provides a slightly more complex trace.

I think the guide is fine just showing how to create a simpler example - but was wondering whether we should refer to the existing example as including the additional endpoint, and how to run it?

PR #4381, opened 2019-10-04
Implement a searchable and expandable configuration property table

FroMage:
This builds on https://github.com/quarkusio/quarkus/pull/4050 to make the config key table collapsible using JS. I think it should work out on PDF too since most of the behaviour is controlled by JS/CSS.

Please @gsmet @emmanuelbernard @machi1990 try it on for size.

Collapsed:

Screenshot from 2019-10-04 18-20-07

Expanded:

Screenshot from 2019-10-04 18-20-16

I'm not entirely sure of the behaviour of expanding the cell to the whole table, but it felt better to read the expanded docs. It has the drawback of hiding the default value and type, but I guess I could dynamically add it when expanding if you think it's important.

I have added collapse/expand all links on each collapsible row, rather than at the top of the table because I'm thinking of the "all config" table and assume that link will be out of reach. I could add it _as well_ though if you think it's important.

I think we can remove the unstructured view with this, because it's just a lot more readable (don't look at the font size, I wish I knew why it became so small all of a sudden, I intend to fix it).

For the "but I can't search collapsed text" I hope users will click on "expand all". Or I can add a search box on every table which will expand all rows that match the searched text. I believe that would be very nice.

Anyway, the pictures don't do it justice to feel the UX: please check it out, compile and open `docs/target/generated-docs/hibernate-orm-guide.html` once it's all compiled/rendered.

PR #4605, opened 2019-10-16
docs: add a tip about overriding the generated default value

machi1990:
The option was added in https://github.com/quarkusio/quarkus/pull/4381

/cc @FroMage @gsmet

### DIFF
--- a/docs/src/main/asciidoc/extension-authors-guide.adoc
+++ b/docs/src/main/asciidoc/extension-authors-guide.adoc
@@ -2189,6 +2189,24 @@ public class SomeConfig {
 1. This will generate a configuration map key named `quarkus.some."cache-name"` instead of `quarkus.some."namespace"`.
 ====
 
+[TIP]
+====
+By default, the doc generator will display the default value supplied via the `defaultValue` attribute of `@ConfigItem`.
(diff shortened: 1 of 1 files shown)


docs/src/main/asciidoc/extension-authors-guide.adoc
```diff
@@ -2189,6 +2189,24 @@ public class SomeConfig {
...
+If you wish to display the default value without hardcoding it in the config item e.g when there are different default values depending on the `quarkus.profile`,
+you can use the `defaultValueDocumentation` attribute as shown below.
+[source,java]
+----
+@ConfigRoot
+public class SomeConfig {
+    /**
+     * Namespace configuration.
+     */
+    @ConfigItem(defaultValueDocumentation = "*value-dev (DEV,TEST)\n*value-prod(DEV)")
```
gsmet: So I haven't followed that but I don't thin we should clutter the `@ConfigItem` annotation with documentation things. I would have preferred it to be separated.

machi1990: What you are saying is we should have an annotation like `ConfigDocDefaultValue`?

machi1990:
Closing this for now. Will revisit later.

PR #4607, opened 2019-10-16
Add the Infinispan Embedded extension to extensions.json

gsmet:


Sanne:
I thought we agreed this extension would stay "hidden" ? There's no much use for it, other than the Infinispan team ramping up their new server efforts.

gsmet:
Ah, OK. Missed that, I stumbled upon that while updating code.quarkus.io. Let's close then.

PR #4614, opened 2019-10-16
Fix Gradle add extension and add tests

### LINKED ISSUE #4601: Adding extension using gradle plugin erase gradle config files content
**Describe the bug**
When adding an extension using gradle plugin:
```
$ ./gradlew addExtension --extensions="io.quarkus:quarkus-resteasy-jackson"
Starting a Gradle Daemon (subsequent builds will be faster)

> Task :addExtension
✅ Adding dependency io.quarkus:quarkus-resteasy-jackson:jar

BUILD SUCCESSFUL in 49s
1 actionable task: 1 executed
```

It erases gradle config files content:
- build.gradle
- gradle.properties
- settings.gradle

**Expected behavior**
It should update the files with the new extension but not erase it.

**To Reproduce**
1. Create a gradle project 
2. Add an extension using `./gradlew addExtension --extensions="..."`
3. Check the files

 - Quarkus version or git rev: 0.24.0

ia3andy:
Closes #4601


ia3andy:
I abstracted MavenExtensionsTest so we have the same coverage for Gradle.

Note that one test is not passing with Gradle and I think it should come in another issue/PR:
https://github.com/quarkusio/quarkus/pull/4614/files#diff-cec51f6e2476bda2790e7fe4074caabdR19

devtools/common/src/main/java/io/quarkus/cli/commands/file/GradleBuildFile.java
```diff
@@ -185,25 +170,25 @@ public void accept(String currentLine) {
...
         return getDependencies().stream()
                 .anyMatch(d -> extension.getGroupId().equals(d.getGroupId())
                         && extension.getArtifactId().equals(d.getArtifactId()));
     }
 
     @Override
-    protected boolean containsBOM() {
-        return buildContent.contains("enforcedPlatform(\"io.quarkus:quarkus-bom:");
+    protected boolean containsBOM() throws IOException {
+        return getModel().getBuildContent().contains("enforcedPlatform(\"io.quarkus:quarkus-bom:");
```
aguibert: although `G:A:V` syntax is more common, people can also declare gradle dependencies with expanded syntax like this:
```groovy
group: 'io.quarkus', name: 'quarkus-bom', version: '0.25.0'
```
with any amount of spacing or newlines in between

devtools/common/src/main/java/io/quarkus/cli/commands/file/GradleBuildFile.java
```diff
@@ -130,20 +115,20 @@ private void completeSettingsContent(String artifactId) {
...
         }
     }
 
     @Override
-    protected void addDependencyInBuildFile(Dependency dependency) {
+    protected void addDependencyInBuildFile(Dependency dependency) throws IOException {
         StringBuilder newBuildContent = new StringBuilder();
-        readLineByLine(buildContent, new AppendDependency(newBuildContent, dependency));
-        buildContent = newBuildContent.toString();
+        readLineByLine(getModel().getBuildContent(), new AppendDependency(newBuildContent, dependency));
```
aguibert: while it's not an issue with this PR specifically, I was looking into the code for AppendDependency and I noticed 2 corner cases that don't currently work:
1. It looks for `line.startsWith("dependencies {"` which doesn't account for additional spacing between `dependencies` and `{` and also doesn't account for leading whitespace (which is normal if this is being applied in a `subprojects{}` block for example).
1. If we have the quarkus plugin installed already, but no `dependencies{}` at all, nothing gets added. For this case I think we should probably add the quarkus-bom dependency with a version corresponding to the quarkus-gradle-plugin, followed by the rest of the requested extension dependencies as normal

ia3andy: TBH I didn't really look at the business or the code itself, and I don't really want to.. I just fixed the bug using the same system used in MavenBuildFile and added the tests that should have been there in the first place..

ia3andy: At some point we are going to rewrite most of the codegen part with a real template engine for Quarkus that is beeing developed https://github.com/quarkusio/qute.

But for now, any corner case you see, please feel free to provide a PR with some tests :)

geoand:
@ia3andy seems like the JVM tests are failing

devtools/common/src/main/java/io/quarkus/cli/commands/file/GradleBuildFile.java
geoand: Yeah this part of the code is messy currently... So trying to improve little things probably doesn't make a ton sense, tests on the other hand definitely do :).
@ia3andy did you sign up for rewriting the part this `qute`? :P

ia3andy: @geoand why not, that might be fun, though for now I have plenty on my plate 😅

ia3andy:
@gsmet we need that merged asap too

devtools/common/src/main/java/io/quarkus/cli/commands/file/GradleBuildFile.java
aloubyansky: Right, not mentioning the code formatting could vary. This functionality needs to be re-written properly.

geoand: `qute` to the rescue :)

aloubyansky (approved):

PR #4622, opened 2019-10-16
chore: move the "elytron-security-jdbc" alongside security related modules.

machi1990:
It was previously added to "kogito" related section

A very minor cleanup PR, which I saw when I was working on adding another extension. 

Also removes trailing "keycloak" label in FeatureBuildItem class

/cc @geoand @gsmet


gsmet (approved):
Looks good, thanks!

PR #4623, opened 2019-10-16
avoid generating config documentation for the test-extension

machi1990:
/cc @stuartwdouglas


stuartwdouglas (approved):
This feels a bit hacky, but I don't really have any better ideas.

PR #4633, opened 2019-10-17
Fix resources update on multi module projects

nelsongraca:
I found this issue while integrating Quarkus on a project I'm developing, I tried to write tests for it but does not seem very trivial and goes beyond my knowledge. I will be happy to provide additional details if needed.

Fix an issue where in multi-module projects the files of a first module would be deleted when the second one was checked for file changes.
The correspondingResources Set became a Map to store the resources on a per-module basis instead of globally.


gsmet (approved):
Makes sense. I agree adding a test for that will be hard.

gsmet:
Merged, thanks!

PR #4642, opened 2019-10-17
Use canonical path for checking -dev.jar location

jaikiran:
The final piece to get https://github.com/quarkusio/quarkus/issues/3592 working. The user has tested this change and verified that this works https://github.com/quarkusio/quarkus/issues/3592#issuecomment-543167894.
I should have anyway started off with canonical path in my original PR, for a check like this.


gsmet (approved):

gsmet:
Merged, thanks for fighting this fight!

PR #4657, opened 2019-10-18
Getting the OIDC test running first

sberyozkin:


sberyozkin:
So I see I only fixed the label, but the OIDC test is still running last so another try is needed

PR #4658, opened 2019-10-18
Ensure QuarkusJsonbComponentInstanceCreator can be used in unit tests

### LINKED ISSUE #4090: Why Quarkus-JsonB is using ArC?
As I was writing a unit test (not `@QuarkusTest`) for a class that uses `Jsonb.fromJson(String,Class)`, I got a NullPointerException on an adapted type field:

```
    @NotNull
    @JsonbProperty("instant")
    @JsonbTypeAdapter(EpochSecondsAdapter.class)
    private Instant instant;  //UTC epoch second Long
```

I'm not sure that the adapted field is actually what causing the problem but It failed on that particular field so I'm writing it down. The way I solved it is by using `Arc.initialize()` as a `BeforeAll` step in the test and `Arc.shutdown()` as a `AfterAll` step.

Why is JsonB using ArC, and is initializing ArC a good practice or I can do something else to make the test pass?

geoand:
Fixes: #4090

### DIFF
--- a/extensions/jsonb/runtime/src/main/java/io/quarkus/jsonb/QuarkusJsonbComponentInstanceCreator.java
+++ b/extensions/jsonb/runtime/src/main/java/io/quarkus/jsonb/QuarkusJsonbComponentInstanceCreator.java
@@ -9,6 +9,7 @@ import java.util.concurrent.ConcurrentHashMap;
 import org.eclipse.yasson.spi.JsonbComponentInstanceCreator;
 
 import io.quarkus.arc.Arc;
+import io.quarkus.arc.ArcContainer;
 import io.quarkus.arc.InstanceHandle;
 
 public class QuarkusJsonbComponentInstanceCreator implements JsonbComponentInstanceCreator {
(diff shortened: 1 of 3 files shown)


extensions/jsonb/runtime/src/main/java/io/quarkus/jsonb/QuarkusJsonbComponentInstanceCreator.java
```diff
@@ -34,17 +35,26 @@ public void close() throws IOException {
...
-            }
+            return fallbackCreate(componentClass);
         });
     }
 
+    private <T> Object fallbackCreate(Class<T> componentClass) {
+        try {
+            return componentClass.newInstance();
+        } catch (InstantiationException | IllegalAccessException e) {
+            throw new IllegalStateException("Cannot instantiate Jsonb component: " + componentClass, e);
```
gsmet:
```suggestion
            throw new IllegalStateException("Cannot instantiate JSON-B component: " + componentClass, e);
```

gsmet (changes requested):
Looks good to me except for a very minor suggestion.

extensions/jsonb/runtime/src/main/java/io/quarkus/jsonb/QuarkusJsonbComponentInstanceCreator.java
geoand: Fixed :)

gsmet (approved):

mkouba:
I know it's too late but it looks good ;-).

PR #4668, opened 2019-10-18
Do not crash when encountering a RESTEasy param annotation on field/property

FroMage:
This has a test, but it breaks openapi due to https://github.com/smallrye/smallrye-open-api/issues/197

I discovered this while editing the quickstarts, so it will be tested there.

### DIFF
--- a/extensions/resteasy-server-common/deployment/src/main/java/io/quarkus/resteasy/server/common/deployment/ResteasyServerCommonProcessor.java
+++ b/extensions/resteasy-server-common/deployment/src/main/java/io/quarkus/resteasy/server/common/deployment/ResteasyServerCommonProcessor.java
@@ -585,6 +585,10 @@ public class ResteasyServerCommonProcessor {
         OUTER: for (DotName annotationType : methodParameterAnnotations) {
             Collection<AnnotationInstance> instances = index.getAnnotations(annotationType);
             for (AnnotationInstance instance : instances) {
+                // we only care about method parameters, because properties or fields always work
+                if (instance.target().kind() != Kind.METHOD_PARAMETER) {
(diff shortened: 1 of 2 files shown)


geoand (approved):

extensions/resteasy-server-common/deployment/src/main/java/io/quarkus/resteasy/server/common/deployment/ResteasyServerCommonProcessor.java
```diff
@@ -585,6 +585,9 @@ private static void checkParameterNames(IndexView index,
         OUTER: for (DotName annotationType : methodParameterAnnotations) {
             Collection<AnnotationInstance> instances = index.getAnnotations(annotationType);
             for (AnnotationInstance instance : instances) {
+                // we only care about method parameters, because properties or fields always work
+                if (instance.target().kind() != Kind.METHOD_PARAMETER)
+                    continue;
```
gsmet: We try to have curly braces even for one line.

geoand: Indeed, that one slipped they eye. Actually we should probably have a formatting rule for that

gsmet: I force pushed a change for that, and a rebase.

gsmet (approved):

FroMage:
OK thanks. CI failure doesn't seem related, so merging.

PR #4676, opened 2019-10-18
ArC - minor cleanup and replace secure random call with id generator

### LINKED ISSUE #4490: ArcContainerImpl could skip SecureRandom initialization
**Description**
The 'ArcContainerImpl' is eagerly initializing an `uuid` field, which requires a Secure random call; according to flamegraphs including native time, this is taking 30% of the time it takes to boot Arc so it would be nice to skip it.

**Implementation ideas**
The `uuid` field seems only necessary for `toString()`. I think it could be lazily initialized, I'm not sure what expectation there is in terms of having the same uuid (or not) of the same app when it's build as native.

I suppose it should mimick the JVM semantics, which implies lazy initialization would actually be better.

mkouba:
- resolves #4490


independent-projects/arc/runtime/src/main/java/io/quarkus/arc/ArcContainerImpl.java
```diff
@@ -69,7 +70,7 @@
     private final List<ResourceReferenceProvider> resourceProviders;
 
     public ArcContainerImpl() {
-        id = UUID.randomUUID().toString();
+        id = "" + ID_GENERATOR.incrementAndGet();
```
Sanne: nice!

manovotn (approved):
Just one question, is `InstanceHandlesContextState` something that's supposed to be used by other (custom) contexts? If so, then it would perhaps deserve a bit of javadoc?

mkouba:
@manovotn  It is not ATM - that's why it's package-private. We could change it later.

PR #4677, opened 2019-10-18
Fix ConfigRecorder missing classloader assignment

gwenneg:
I was taking a look at the `ConfigRecorder` code for another reason when I saw this.


mkouba (approved):
Great catch! The TCCL should be probably always set. Anyway, thanks a lot.

Issue #3064, opened 2019-07-02
OIDC adapter should use smallrye-jwt extension


sberyozkin:
**Description**
smallrye-jwt should be able to let the Keycloak adapter validate the tokens as well take care of all the subtleties associated with the OIDC flows and 'extend' it by letting the users access the tokens via MP JWT Api

**Implementation ideas**
Bill: "smalrye-jwt could just look for a marker BuildItem and not wire in the defaut implementation if that exists.  Keycloak would just publish that marker BuildItem"
plus: possibly provide a handler for converting the token into a `Principal`; KC adapter would check for it before providing its own `Principal`

sberyozkin:
Checking one of the KC @BuildStep capabilities should be enough

sberyozkin:
@stuartwdouglas FYI

sberyozkin:
Hi @cescoffier, @stuartwdouglas 
Here is one possible way which it might work. 
1. Introduce `PrincipalProvider` spi:
```
interface PrincipalProvider {
    Principal toPrincipal(Credential cred);
}
```
2. `quarkus-smallrye-jwt` ships `JsonWebTokenProvider`:

```
class JsonWebTokenProvider implements PrincipalProvider {
    public Principal toPrincipal(Credential cred) {
          if (cred instanceof TokenCredential) {
              return convertToJsonWebToken(cred);
          } else {
              return null;
          }
    }
}
```
3. Have the 'master' token validators, example, a keycloak adapter, checking for `PrincipalProvider` before using its own Principal implementation

It looks like it can work...thanks

stuartwdouglas:
You can already do this via io.quarkus.security.identity.SecurityIdentityAugmentor

sberyozkin:
@stuartwdouglas this is very nice :-) I guess `smallrye-jwt` should first be updated to work with the vertx-web, but I can definitely experiment with this enhancement independently

sberyozkin:
Hi Stuart, @stuartwdouglas, 
[...]

PR #4681, opened 2019-10-19
Add a jsonp dependency to quarkus-oidc

### LINKED ISSUE #4672: OIDC Extension throws ClassNotFoundException when using quarkus-resteasy-jackson
**Describe the bug**
I am migration a project with the Keycloak extension (Quarkus 0.22.0) to the new OIDC extension (Quarkus 0.25.0). I am using the `quarkus-resteasy-jackson` extension instead of jsonb. Unfortunately, these extensions are no longer interchangeable: the access token cannot be parsed, because org.eclipse.microprofile.jwt.Claims specifies JsonObject types, e.g. for the address.
See [stacktrace_JsonObject_notFound.txt](https://github.com/quarkusio/quarkus/files/3744710/stacktrace_JsonObject_notFound.txt)

**Expected behavior**
The app does not crash. Required dependencies are provided.

**Actual behavior**
ClassNotFoundException

**To Reproduce**
1. In the `using-openid-connect` starter pom, change the dependency `quarkus-resteasy-jsonb` to `quarkus-resteasy-jackson` 
2. access UserResource as described in the [guide](https://quarkus.io/guides/oidc-guide.html).

sberyozkin:
Fixes #4672

### DIFF
--- a/extensions/oidc/runtime/pom.xml
+++ b/extensions/oidc/runtime/pom.xml
@@ -31,6 +31,10 @@
             <artifactId>quarkus-security</artifactId>
         </dependency>
         <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-jsonp</artifactId>
+        </dependency>
+        <dependency>
             <groupId>io.smallrye</groupId>
             <artifactId>smallrye-jwt</artifactId>
         </dependency>
--- a/extensions/oidc/deployment/pom.xml
+++ b/extensions/oidc/deployment/pom.xml
@@ -30,6 +30,10 @@
             <groupId>io.quarkus</groupId>
             <artifactId>quarkus-security-deployment</artifactId>
         </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-jsonp-deployment</artifactId>
+        </dependency>
         <!-- Test dependencies -->
         <dependency>
             <groupId>io.quarkus</groupId>
--- a/integration-tests/oidc/pom.xml
+++ b/integration-tests/oidc/pom.xml
@@ -25,7 +25,7 @@
         </dependency>
         <dependency>
             <groupId>io.quarkus</groupId>
-            <artifactId>quarkus-resteasy-jsonb</artifactId>
+            <artifactId>quarkus-resteasy-jackson</artifactId>
         </dependency>
         <dependency>
             <groupId>io.quarkus.keycloak</groupId>
--- a/integration-tests/oidc/src/main/java/io/quarkus/it/keycloak/PermissionResource.java
+++ b/integration-tests/oidc/src/main/java/io/quarkus/it/keycloak/PermissionResource.java
@@ -5,6 +5,7 @@ import java.util.Map;
 
 import javax.annotation.security.RolesAllowed;
 import javax.inject.Inject;
+import javax.json.JsonString;
 import javax.ws.rs.GET;
 import javax.ws.rs.Path;
 import javax.ws.rs.Produces;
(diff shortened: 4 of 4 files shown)


sberyozkin:
The #4672 issue is that `quarkus-oidc` does not include `quarkus-jsonp` as it currently depends directly on `smallrye-jwt` as opposed to `quarkus-smallrye-jwt` which does include `quarkus-jsonp`.

geoand:
@sberzyokin can you elaborate on this a little?

sberyozkin:
@geoand Just typed it :-)

gsmet:
So it's not supposed to depend on `quarkus-smallrye-jwt`?

gsmet:
Looks like our JSON-P extension is missing something:
```
Caused by: java.lang.InstantiationException: Type `org.glassfish.json.JsonProviderImpl` can not be instantiated reflectively as it does not have a no-parameter constructor or the no-parameter constructor has not been added explicitly to the native image.
	at java.lang.Class.newInstance(DynamicHub.java:770)
	at javax.json.spi.JsonProvider.provider(JsonProvider.java:97)
	... 71 more
```

extensions/oidc/runtime/pom.xml
```diff
@@ -30,6 +30,10 @@
             <groupId>io.quarkus</groupId>
             <artifactId>quarkus-security</artifactId>
         </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-jsonp</artifactId>
+        </dependency>
```
gsmet: I don't know if the right fix is to add this one or to depend on `quarkus-smallrye-jwt`. I let you judge of that, not knowing exactly what the SmallRye JWT extension is doing (but my fear is that we might miss some important native image-related stuff somehow).

In any case, you also need to add the `-deployment` dependency to the deployment module. That's why the tests are failing right now.

gsmet (changes requested):

extensions/oidc/runtime/pom.xml
sberyozkin: @gsmet sorry, I was working in replacing the test code with the Jackson provider and it worked, but not sure why I forgot to recompile all of the `oidc` extension...Let me answer about `quarkus-smallrye-jwt` in a few mins

sberyozkin:
@geoand, @gsmet Hi, sorry for a delay... So there is #3064. Stuart did the initial integration directly with `smallrye-jwt` to support the `OIDC token to be accessed via MP-JWT API` case as at the moment we don't have a mechanism for `quarkus-smallrye-jwt` to complement the `primary` authentication mechanism offered in this case by `quarkus-oidc` while disabling its own authentication mechanism and due to this issue it is not obvious yet if this dependency has to be introduced after all because it may be simpler for `quarkus-oidc` to work directly with `smallrye-jwt`, not sure yet :-)

geoand:
Thanks for the explanation! Although I have to admit that I haven't followed the oidc stuff so I can't really provide any meaningful feedback here...
If you are confident that this is the proper fix and if @gsmet doesn't have any objections, I propose we merge it.

sberyozkin:
Hi Georgios, thanks, yes, at this moment of time this is the right fix

sberyozkin:
Well, a bit more work to be done, but it is purely a test issue now, sorry

geoand:
Cool, np!

sberyozkin:
@geoand I'm a bit all over the place here at home tight now :-), I've done a quick native test fix, the real point of the PR is to prove that Jackson is not interfering into the OIDC adapter work, and as far as a complete JSONP to String/List/Map is concerned - we can deal with it once the test starts growing to check more complex claims...

sberyozkin:
By the way, not sure why Jackson only started complaining about failing to convert a JSONString only in the native mode, I forgot earlier to add `-Dnative` and it all worked OK...

geoand:
I'm cool with it 😎. Let's see if @gsmet agrees.

gsmet (approved):

gsmet:
Merged, thanks.

PR #4682, opened 2019-10-19
Check ConfigProperties missing config values

### LINKED ISSUE #4604: DeploymentException not thrown when using @ConfigProperties
`@ConfigProperties` does not behave the same as `@ConfigProperty`

When using `@ConfigProperty` to inject configuration, if the property is missing, a DeploymentException is thrown at start-up.

For example, adding the following to a Controller:
```
@ConfigProperty(name = "cassandra.connection.node") public String node;
@ConfigProperty(name = "cassandra.connection.port") public int port;
```
and the following to application.properties (notice additional r on node) :
```
cassandra.connection.noder=0.0.0.0
cassandra.connection.port=9042
```
results in an exception at start-up:
```
javax.enterprise.inject.spi.DeploymentException: No config value of type [java.lang.String] exists for: cassandra.connection.node
```

However, when using `@ConfigProperties` as follows:
```
@ConfigProperties(prefix = "cassandra")
public class CassandraConfiguration
{
  public ConnectionData connection;

  public static class ConnectionData
  {
    public String node;
    public int port;
  }
}
```
Given the same application.properties file as above, an exception is not thrown at start-up. Instead, the application starts, but throws an UnhandledException, which wraps a NoSuchElementException, when the property is accessed:
```
UnhandledException: java.lang.RuntimeException: Error injecting com.***.CassandraConfiguration com.***.cassandraConfiguration
....
Caused by: java.util.NoSuchElementException: Property cassandra.connection.node not found
```

`@ConfigProperties` works as expected when the application.properties are correct:
```
cassandra.connection.node=0.0.0.0
cassandra.connection.port=9042
```

gwenneg:
Fixes #4604

### DIFF
--- a/extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/configproperties/ClassConfigPropertiesUtil.java
+++ b/extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/configproperties/ClassConfigPropertiesUtil.java
@@ -1,15 +1,14 @@
 package io.quarkus.arc.deployment.configproperties;
 
-import static io.quarkus.arc.deployment.configproperties.ConfigPropertiesUtil.ReadOptionalResponse;
 import static io.quarkus.arc.deployment.configproperties.ConfigPropertiesUtil.createReadMandatoryValueAndConvertIfNeeded;
 import static io.quarkus.arc.deployment.configproperties.ConfigPropertiesUtil.createReadOptionalValueAndConvertIfNeeded;
 import static io.quarkus.arc.deployment.configproperties.ConfigPropertiesUtil.determineSingleGenericType;
 
 import java.lang.reflect.Modifier;
+import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.NoSuchElementException;
 import java.util.Optional;
 import java.util.Set;
 
@@ -26,6 +25,9 @@ import org.jboss.jandex.IndexView;
 import org.jboss.jandex.MethodInfo;
 import org.jboss.jandex.Type;
 
+import io.quarkus.arc.deployment.ConfigPropertyBuildItem;
+import io.quarkus.arc.deployment.configproperties.ConfigPropertiesUtil.ReadOptionalResponse;
+import io.quarkus.deployment.annotations.BuildProducer;
 import io.quarkus.deployment.bean.JavaBeanUtil;
 import io.quarkus.deployment.util.HashUtil;
 import io.quarkus.gizmo.BranchResult;
@@ -112,8 +114,9 @@ final class ClassConfigPropertiesUtil {
     /**
      * @return true if the configuration class needs validation
      */
-    static boolean addProducerMethodForClassConfigProperties(ClassInfo configPropertiesClassInfo,
-            ClassCreator producerClassCreator, String prefixStr, IndexView applicationIndex) {
+    static boolean addProducerMethodForClassConfigProperties(ClassLoader classLoader, ClassInfo configPropertiesClassInfo,
+            ClassCreator producerClassCreator, String prefixStr, IndexView applicationIndex,
+            BuildProducer<ConfigPropertyBuildItem> configProperties) {
 
         if (!DotNames.OBJECT.equals(configPropertiesClassInfo.superName())) {
             throw new IllegalArgumentException(
@@ -164,9 +167,8 @@ final class ClassConfigPropertiesUtil {
                 configObjectClassStr, produceMethodParameterTypes)) {
             methodCreator.addAnnotation(Produces.class);
 
-            ResultHandle configObject = populateConfigObject(configPropertiesClassInfo,
-                    prefixStr,
-                    methodCreator, applicationIndex);
+            ResultHandle configObject = populateConfigObject(classLoader, configPropertiesClassInfo, prefixStr, methodCreator,
+                    applicationIndex, configProperties);
 
             if (needsValidation) {
                 createValidationCodePath(methodCreator, configObject, prefixStr);
@@ -195,11 +197,13 @@ final class ClassConfigPropertiesUtil {
         }
     }
 
-    private static ResultHandle populateConfigObject(ClassInfo configClassInfo,
-            String prefixStr,
-            MethodCreator methodCreator, IndexView applicationIndex) {
+    private static ResultHandle populateConfigObject(ClassLoader classLoader, ClassInfo configClassInfo, String prefixStr,
+            MethodCreator methodCreator, IndexView applicationIndex, BuildProducer<ConfigPropertyBuildItem> configProperties) {
         String configObjectClassStr = configClassInfo.name().toString();
         ResultHandle configObject = methodCreator.newInstance(MethodDescriptor.ofConstructor(configObjectClassStr));
+
+        // Fields with a default value will be removed from this list at the end of the method.
+        List<ConfigPropertyBuildItemCandidate> configPropertyBuildItemCandidates = new ArrayList<>();
 
         // For each field of the class try to pull it out of MP Config and call the corresponding setter
         List<FieldInfo> fields = configClassInfo.fields();
@@ -240,9 +244,8 @@ final class ClassConfigPropertiesUtil {
                             "Nested configuration class " + fieldTypeClassInfo + " must be public ");
                 }
 
-                ResultHandle nestedConfigObject = populateConfigObject(fieldTypeClassInfo,
-                        prefixStr + "." + field.name(),
-                        methodCreator, applicationIndex);
+                ResultHandle nestedConfigObject = populateConfigObject(classLoader, fieldTypeClassInfo,
+                        prefixStr + "." + field.name(), methodCreator, applicationIndex, configProperties);
                 createWriteValue(methodCreator, configObject, field, setter, useFieldAccess, nestedConfigObject);
 
             } else {
@@ -286,16 +289,6 @@ final class ClassConfigPropertiesUtil {
                                     MethodDescriptor.ofMethod(configObjectClassStr, getterName, fieldTypeStr),
                                     configObject);
                         }
-
-                        /*
-                         * If the getter value is null (meaning that no default was set)
-                         * throw the exception that MP Config would.
-                         * If the value is not null, then we don't need to do anything since the class will already
-                         * contain the default value the user intended to use
-                         */
-                        isPresentFalse.ifNull(defaultValue).trueBranch()
-                                .throwException(NoSuchElementException.class,
-                                        "Property " + fullConfigName + " not found");
                     } else {
                         /*
                          * In this case we want a missing property to cause an exception that we don't handle
@@ -307,9 +300,19 @@ final class ClassConfigPropertiesUtil {
                         createWriteValue(methodCreator, configObject, field, setter, useFieldAccess, setterValue);
 
                     }
+                    configPropertyBuildItemCandidates
+                            .add(new ConfigPropertyBuildItemCandidate(field.name(), fullConfigName, fieldTypeStr));
                 }
             }
         }
+
+        ConfigPropertyBuildItemCandidateUtil.removePropertiesWithDefaultValue(classLoader, configObjectClassStr,
+                configPropertyBuildItemCandidates);
+        for (ConfigPropertyBuildItemCandidate candidate : configPropertyBuildItemCandidates) {
+            configProperties
+                    .produce(new ConfigPropertyBuildItem(candidate.getConfigPropertyName(), candidate.getConfigPropertyType()));
+        }
+
         return configObject;
     }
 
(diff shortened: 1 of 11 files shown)


gwenneg:
The `InterfaceConfigProperties` part is working great and I wish I could say the same for the `ClassConfigProperties` part, but I'm still looking for a way to detect a class field default value at build time.

geoand:
@gwenneg Thanks fot this PR!

Let me know when you want me to review it

gwenneg:
I'm not sure yet of the best way to determine at build time whether or not a config class field as a user-chosen default value (not the default Java one). I'm pretty sure my sentence is not clear so I'll give an example.

I'm currently unable to tell the difference between...

```
@ConfigProperties
public static class DummyProperties {
    private boolean boolWithDefault = false;
}
```

... and...

```
@ConfigProperties
public static class DummyProperties {
    private boolean boolWithDefault;
}
```

... using Reflection at build time.

I'd like to exclude the first case from the validation performed in `ConfigBuildStep.validateConfigProperties()` but not the second one. Currently, both are excluded from validation in my local code (I haven't pushed this part in the PR yet).

But maybe Reflection is not the way to go to validate config classes (I'm not using it for interfaces which is a much simpler case already included in this PR).

geoand:
That's a very good point!

We definitely dont want to throw a validation error when there is a default value set (and perhaps also when the type of the field is a primitive).
I think the proper way to implement this would be to have ASM visitor that visits the class and checks if the constructor contains a corresponding with you set field instruction.
Not trivial, but it shouldn't be terribly difficult to implement either.

WDYT?

gwenneg:
@geoand I never had the occasion to play with ASM, so I started looking at it tonight. It looks like the good way to solve my issue indeed. I'm now working on an ASM-based PR :) Thanks for the suggestion!

geoand:
@gwenneg your welcome!

I am thinking that ASM will only be needed for the part described above. I don't think that ASM will be needed for anything else but determining which fields have been initialized.

gwenneg:
Yes, I've only replaced the Reflection part with ASM, I'm not using ASM for the config properties declared from interfaces.

geoand:
Cool, just wanted to make sure you don't do any unncessary work since after all ASM isn't fun to work with :)

geoand:
@gwenneg nice progress!

The `main` integration test is failing however with something like:

```
[ERROR] Failed to execute goal io.quarkus:quarkus-maven-plugin:999-SNAPSHOT:build (default) on project quarkus-integration-test-main: Failed to build a runnable JAR: Failed to augment application classes: Build failure: Build failed due to errors
[ERROR] 	[error]: Build step io.quarkus.arc.deployment.configproperties.ConfigPropertiesBuildStep#setup threw an exception: java.io.UncheckedIOException: io.quarkus.it.config.ConfigPropertiesResource$GreetingConfiguration class reading failed
[ERROR] 	at io.quarkus.arc.deployment.configproperties.ConfigPropertyBuildItemCandidateUtil.removePropertiesWithDefaultValue(ConfigPropertyBuildItemCandidateUtil.java:26)
[ERROR] 	at io.quarkus.arc.deployment.configproperties.ClassConfigPropertiesUtil.populateConfigObject(ClassConfigPropertiesUtil.java:312)
[ERROR] 	at io.quarkus.arc.deployment.configproperties.ClassConfigPropertiesUtil.addProducerMethodForClassConfigProperties(ClassConfigPropertiesUtil.java:170)
[ERROR] 	at io.quarkus.arc.deployment.configproperties.ConfigPropertiesBuildStep.setup(ConfigPropertiesBuildStep.java:95)
[ERROR] 	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
[ERROR] 	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
[ERROR] 	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
[ERROR] 	at java.lang.reflect.Method.invoke(Method.java:498)
[ERROR] 	at io.quarkus.deployment.ExtensionLoader$1.execute(ExtensionLoader.java:940)
[ERROR] 	at io.quarkus.builder.BuildContext.run(BuildContext.java:415)
[ERROR] 	at org.jboss.threads.ContextClassLoaderSavingRunnable.run(ContextClassLoaderSavingRunnable.java:35)
[ERROR] 	at org.jboss.threads.EnhancedQueueExecutor.safeRun(EnhancedQueueExecutor.java:2011)
[ERROR] 	at org.jboss.threads.EnhancedQueueExecutor$ThreadBody.doRunTask(EnhancedQueueExecutor.java:1535)
[ERROR] 	at org.jboss.threads.EnhancedQueueExecutor$ThreadBody.run(EnhancedQueueExecutor.java:1426)
[ERROR] 	at java.lang.Thread.run(Thread.java:748)
[ERROR] 	at org.jboss.threads.JBossThread.run(JBossThread.java:479)
[ERROR] Caused by: java.io.IOException: Class not found
[ERROR] 	at org.objectweb.asm.ClassReader.readStream(ClassReader.java:310)
[ERROR] 	at org.objectweb.asm.ClassReader.<init>(ClassReader.java:295)
[ERROR] 	at io.quarkus.arc.deployment.configproperties.ConfigPropertyBuildItemCandidateUtil.removePropertiesWithDefaultValue(ConfigPropertyBuildItemCandidateUtil.java:23)
[ERROR] 	... 15 more
[ERROR] -> [Help 1]
```
Also, when everything is ready to go it would be great to have the documentation updated to include this as well.

gwenneg:
Oops sorry, I missed that failure locally.

geoand:
No problem! Happens to everyone 😉

gwenneg:
CI should be happier with this new version. I'll take care of the documentation soon.

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/configproperties/ConfigPropertyBuildItemCandidateUtil.java
```diff
@@ -0,0 +1,71 @@
...
+public class ConfigPropertyBuildItemCandidateUtil {
+
+    /**
+     * This method inspects the {@code configClass} bytecode to identify all fields that have a default value set in the class
+     * constructor. These fields are removed from the {@link ConfigPropertyBuildItemCandidate} list because we don't want to
+     * throw an exception if no config property value was provided for them. There is no bytecode modification performed during
+     * this process.
+     */
+    public static void removePropertiesWithDefaultValue(String configClass, List<ConfigPropertyBuildItemCandidate> candidates) {
+        ClassLoader cl = Thread.currentThread().getContextClassLoader();
```
geoand: We now have a better option that the TCCL (which has always been a little fuzzy): Check out `DeploymentClassLoaderBuildItem` and specifically how it's being used - I think that the way it's being usef for `smallrye-reactive-messaging` should apply here

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/configproperties/ClassConfigPropertiesUtil.java
```diff
@@ -286,16 +292,6 @@ private static ResultHandle populateConfigObject(ClassInfo configClassInfo,
...
                                     configObject);
                         }
-
-                        /*
-                         * If the getter value is null (meaning that no default was set)
-                         * throw the exception that MP Config would.
-                         * If the value is not null, then we don't need to do anything since the class will already
-                         * contain the default value the user intended to use
-                         */
-                        isPresentFalse.ifNull(defaultValue).trueBranch()
```
geoand: Can you please explain why this needs to be removed?

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/configproperties/ClassConfigPropertiesUtil.java
```diff
@@ -307,9 +303,19 @@ private static ResultHandle populateConfigObject(ClassInfo configClassInfo,
                         createWriteValue(methodCreator, configObject, field, setter, useFieldAccess, setterValue);
 
                     }
+                    configPropertyBuildItemCandidates
+                            .add(new ConfigPropertyBuildItemCandidate(field.name(), fullConfigName, fieldTypeStr));
                 }
             }
         }
+
+        ConfigPropertyBuildItemCandidateUtil.removePropertiesWithDefaultValue(configObjectClassStr,
```
geoand: :+1:

geoand:
Excellent work! I think that when the points I mentioned above are covered, we should be good to go.

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/configproperties/ClassConfigPropertiesUtil.java
gwenneg: Since missing properties are detected at build time with this PR, can this runtime exception still be thrown? It looks like the answer is no to me, but maybe I'm missing something.

geoand: Hold on, the `build time` part is scaring me :)

You mean application startup time, right?

gwenneg: Yes sorry, I meant startup time :)

geoand: OK, I thought that I had massively missunderstood this PR :)

gwenneg: My fingers answered while my brain was focused on another matter, my bad :D

geoand: NP! Just wanted to get things straight.

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/configproperties/ConfigPropertyBuildItemCandidateUtil.java
gwenneg: Classloader replaced. I didn't know about `DeploymentClassLoaderBuildItem`, thanks for this very useful comment!

gwenneg:
I'm not sure about what exactly needs to be added to the doc.

There's already the `If the validation fails with the given configuration, the application will fail to start and indicate the corresponding validation errors in the log.` sentence in the `Using @ConfigProperties` section.

I could make this sentence more detailed (giving an example of the exception thrown like the one in `Injecting configuration value`) and/or add an extra attribute without a provided config value and its legend to indicate that the app will start to fail.

WDYT?

geoand:
You can just add a comment that it start-up fails for missing properties.

gwenneg:
I didn't realize the sentence I quoted was about bean validation and not about missing properties, sorry. I just added something in the doc, I hope it's OK.

geoand:
> I didn't realize the sentence I quoted was about bean validation and not about missing properties, sorry. I just added something in the doc, I hope it's OK.

The fact that you didn't realize means that the sentence (I most likely wrote) wasn't good enough :).
I'll check your latest changes in the morning.

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/configproperties/ConfigPropertyBuildItemCandidateUtil.java
geoand: Yeah, it's pretty new and definitely very useful!

gwenneg:
It is actually pretty clear, I was just too focused on the missing properties values and not paying attention the bean validation part of the doc ;)

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/configproperties/ConfigPropertyBuildItemCandidate.java
```diff
@@ -0,0 +1,40 @@
+package io.quarkus.arc.deployment.configproperties;
+
+public class ConfigPropertyBuildItemCandidate {
```
geoand: Minor nitpick:

I assume that the fields of this class can be made final and we can do away with the setters, correct?

gwenneg: You are absolutely right! Fixed.

geoand: Cool, thanks

geoand (approved):
Thanks a lot of this @gwenneg!

gwenneg:
Thanks for your help reviewing this and especially for the ASM part, it was an interesting discovery.

geoand:
You are most welcome!

Issue #4684, opened 2019-10-20
Add variable expansion in MANIFEST.MF file for CI/CD builds


rmcguinness:
**Description**
When building a quarkus runner project, the MANIFEST file may contain existing META-INF/MANIFEST.MF entries, but those entries MUST be static. CI/CD systems often add attributes to the environment during execution of pipelines that are then used by downstream system for monitoring, routing and running the produced JAR files.

**Implementation ideas**
Allow the META-INF/MANIFEST.MF file to contain place holders wrapped in ${} notation and the System.getenv(<key_name>) to match a system property if the value is annotated, and the system property is found. Such that:

```
MY_PROPERTY_VALUE=${property.key}
```
Could resolve the '-Dproperty.key="My Value"' and the resulting MANIFEST entry would be:

```
MY_PROPERTY_VALUE=My Value
```

The other way to fix this would be honoring options in the plugin similar to how the JAR plugin works with Maven.

geoand:
This has now been implemented. See https://github.com/quarkusio/quarkus/blob/b8dfcd92bef5f7d750a3f02a958a78598f3dc9d6/core/deployment/src/main/java/io/quarkus/deployment/pkg/ManifestConfig.java#L9

PR #4698, opened 2019-10-21
Manifest Variable Expansion #4684

rmcguinness:
The following code will handle variable expansion if a static MANIFEST.MF is used for the runner jar.

I broke the current manifest method into two smaller parts, these could become static for easier unit testing.

- Note: Unable to add a clean test for this, guidance welcome here as there were no additional tests for supporting the static META-INF/MANIFEST.MF file.

### DIFF
--- a/core/creator/src/main/java/io/quarkus/creator/phase/runnerjar/RunnerJarPhase.java
+++ b/core/creator/src/main/java/io/quarkus/creator/phase/runnerjar/RunnerJarPhase.java
@@ -16,7 +16,18 @@ import java.nio.file.Paths;
 import java.nio.file.SimpleFileVisitor;
 import java.nio.file.StandardCopyOption;
 import java.nio.file.attribute.BasicFileAttributes;
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.EnumSet;
+import java.util.Enumeration;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Set;
 import java.util.function.Consumer;
 import java.util.jar.Attributes;
 import java.util.jar.Manifest;
@@ -444,8 +455,6 @@ public class RunnerJarPhase implements AppCreationPhase<RunnerJarPhase>, RunnerJ
         attributes.put(Attributes.Name.MAIN_CLASS, mainClass);
     }
 
-
-
     /**
      * Resolve any attribute name that is a place holder between ${...} notation.
      * This is used by build systems that inject metadata to the manifest at build time,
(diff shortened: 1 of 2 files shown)


gsmet:
@stuartwdouglas I let you decide if it makes sense to support this feature?

core/creator/src/main/java/io/quarkus/creator/phase/runnerjar/RunnerJarPhase.java
```diff
@@ -426,6 +431,55 @@ private void filterZipFile(Path resolvedDep, Path targetPath, Set<String> transf
...
+
+    /**
+     * Resolve any attribute name that is a place holder between ${...} notation.
+     * This is used by build systems that inject metadata to the manifest at build time,
+     * and generally created by CI/CD systems.
+     *
+     * @param attributes - Manifest Attributes to be resolved
+     */
+    private void handleResolvableAttributes(final Attributes attributes) {
+        attributes.keySet().forEach(k -> {
```
stuartwdouglas: Please just use loops for simple stuff like this. Lambda based iteration is way slower and uses a lot of metaspace that hangs around for the life of the program. This is not that much of an issue for deployment time code but the lambda does not even make the code less verbose in this case.

core/creator/src/main/java/io/quarkus/creator/phase/runnerjar/RunnerJarPhase.java
```diff
@@ -426,6 +431,55 @@ private void filterZipFile(Path resolvedDep, Path targetPath, Set<String> transf
...
+            String value = (String) attributes.get(k);
+            if (Objects.nonNull(value)) {
+                String trimmedValue = value.trim();
+                if (!trimmedValue.isEmpty() && trimmedValue.startsWith(PLACEHOLDER_PREFIX) &&
```
stuartwdouglas: If we are going to do this it should be based on our existing expansion code, and take the value from MP config rather than a system property.

I think something like `org.wildfly.common.expression.Expression.compile(value, Expression.Flag.LENIENT_SYNTAX, Expression.Flag.NO_TRIM).evaluate(ConfigExpander.INSTANCE)`

core/creator/src/test/java/io/quarkus/creator/phase/runnerjar/test/VariableSubstitutionTest.java
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * This tests the core logic of the variable substitution;
+ */
+public class VariableSubstitutionTest {
+    private static final String PLACEHOLDER_PREFIX = "${";
+    private static final String PLACEHOLDER_SUFFIX = "}";
+
+    @Test
+    public void testVariableSubstitution() {
```
stuartwdouglas: This is not really a test, but we don't really have any test infrastructure at the moment for running the generated jar files (they get tested indirectly via the native image tests). You could probably write a real test in integration/main  using @NativeImageTest

stuartwdouglas (changes requested):

kunal-kushwaha:
@rmcguinness : Is this PR still a WIP?

rmcguinness:
Still a WIP. There’s another developer on the team working on it now. Thanks for asking!

Sincerely,
Ryan

“Stay away from negative people, they have a problem for every solution” - Albert Einstein

geoand:
Superceded by #8199

PR #4702, opened 2019-10-21
Ensure Spring RestControllers paths show up in dev-mode 404

### LINKED ISSUE #4637: Spring Controller's don't show up in quarkus:dev 404 page
The live reload mode (quarkus:dev) provides a great error page that displays all the endpoints.
However it currently only displays JAX-RS endpoints and should be extended to support Spring Controllers as well

geoand:
Fixes #4637

### DIFF
--- a/extensions/resteasy/runtime/src/main/java/io/quarkus/resteasy/runtime/ExceptionMapperRecorder.java
+++ b/extensions/resteasy/runtime/src/main/java/io/quarkus/resteasy/runtime/ExceptionMapperRecorder.java
@@ -17,4 +17,15 @@ public class ExceptionMapperRecorder {
         NotFoundExceptionMapper.servlets(servletToMapping);
     }
 
+    /**
+     * Uses to register the paths of classes that are not annotated with JAX-RS annotations (like Spring Controllers for
+     * example)
+     * 
+     * @param nonJaxRsClassNameToMethodPaths A map that contains the class name as a key and a map that
+     *        contains the method name to path as a value
+     */
+    public void nonJaxRsClassNameToMethodPaths(Map<String, NonJaxRsClassMappings> nonJaxRsClassNameToMethodPaths) {
+        NotFoundExceptionMapper.nonJaxRsClassNameToMethodPaths(nonJaxRsClassNameToMethodPaths);
+    }
+
 }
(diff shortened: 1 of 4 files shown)


geoand:
@gsmet am I too late for `0.26` :P?

extensions/resteasy/runtime/src/main/java/io/quarkus/resteasy/runtime/NotFoundExceptionMapper.java
```diff
@@ -45,6 +45,7 @@
 
     private volatile static List<String> servletMappings = Collections.EMPTY_LIST;
     private volatile static List<String> staticResources = Collections.EMPTY_LIST;
+    private volatile static Map<String, NonJaxRsClassMappings> nonJaxRsPaths = Collections.EMPTY_MAP;
```
FroMage: I would rather have a name that indicates that we're mapping from class names, which appears to be the case.

FroMage (changes requested):
LGTM except for one ambiguous variable name,

extensions/resteasy/runtime/src/main/java/io/quarkus/resteasy/runtime/NotFoundExceptionMapper.java
geoand: Makes sense, will fix.

extensions/resteasy/runtime/src/main/java/io/quarkus/resteasy/runtime/ExceptionMapperRecorder.java
```diff
@@ -17,4 +17,15 @@ public void setServlets(Map<String, List<String>> servletToMapping) {
...
     }
 
+    /**
+     * Uses to register the paths of classes that are not annotated with JAX-RS annotations (like Spring Controllers for
+     * example)
+     * 
+     * @param nonJaxRsPaths A that contains the class name as a key and class that contains the base path and a map that
+     *        contains the method name to path as a value
+     */
+    public void setNonJaxRsPaths(Map<String, NonJaxRsClassMappings> nonJaxRsPaths) {
```
FroMage: Actually, same here: I'd use a variable name that makes the javadoc non-required.

geoand: :+1:

extensions/resteasy/runtime/src/main/java/io/quarkus/resteasy/runtime/NotFoundExceptionMapper.java
geoand: Better now?

extensions/resteasy/runtime/src/main/java/io/quarkus/resteasy/runtime/ExceptionMapperRecorder.java
geoand: Any better?

FroMage (approved):
Thanks.

PR #4703, opened 2019-10-21
Add note about prefixing properties with "quarkus."

geoand:
This PR was opened because we have users using properties `quarkus.message` to configure their own application.

See [this](https://quarkusio.zulipchat.com/#narrow/stream/187030-users/topic/Config.20properties.20and.20nativ.20images) discussion.


geoand:
@gsmet If you agree, I think this is safe to be part of `0.26`.

gsmet (approved):

PR #4715, opened 2019-10-21
fix(docs): propose a mapKey for generated config docs of a Map config item only for config group

### LINKED ISSUE #4700: Doc generation - issue with passthrough maps
The MongoDB client guide has a passthrough maps and it ends up being weird in the documentation.

In this particular case of a `Map<String, NOT CONFIG OBJECT>`, the type should be `Map<String, NOT CONFIG OBJECT>` and we shouldn't have the key.

issue-map

machi1990:
Fixes #4700

/cc @gsmet can you try this fix please? thanks.

### DIFF
--- a/core/processor/src/main/java/io/quarkus/annotation/processor/generate_doc/ConfigDoItemFinder.java
+++ b/core/processor/src/main/java/io/quarkus/annotation/processor/generate_doc/ConfigDoItemFinder.java
@@ -176,18 +176,18 @@ class ConfigDoItemFinder {
                     if (!typeArguments.isEmpty()) {
                         // FIXME: this is super dodgy: we should check the type!!
                         if (typeArguments.size() == 2) {
-                            final String mapKey = String.format(NAMED_MAP_CONFIG_ITEM_FORMAT, configDocMapKey);
                             type = typeArguments.get(1).toString();
                             configGroup = configGroups.get(type);
-                            name += mapKey;
 
                             if (configGroup != null) {
                                 name += String.format(NAMED_MAP_CONFIG_ITEM_FORMAT, configDocMapKey);
                                 List<ConfigDocItem> groupConfigItems = recordConfigItemsFromConfigGroup(configPhase, name,
-                                        configGroup, configSection, true, sectionLevel);
+                                        configGroup,
+                                        configSection, true, sectionLevel);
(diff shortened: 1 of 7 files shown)


gsmet:
@machi1990 so I finally got to this one: I rebased and generated the doc but it doesn't work.

There are several elements with the same anchor in the KeycloakPolicyEnforcerConfig. I wonder if it's not the two maps at the bottom causing issues.

Could you have a look?

machi1990:
> @machi1990 so I finally got to this one: I rebased and generated the doc but it doesn't work.
> 

> There are several elements with the same anchor in the KeycloakPolicyEnforcerConfig. I wonder if it's not the two maps at the bottom causing issues.

Hi Guillaume, I think we the `@ConfigItem(name = ConfigItem.PARENT)` parent name, the two configs will end up with parent name/key i.e `quarkus.keycloak.policy-enforcer."paths".claim-information-point`.  

What we can try to detect this configuration and fallback to not treating the keys as passthrough - we'll have something like this `quarkus.keycloak.policy-enforcer."paths".claim-information-point."complex-config"` and `quarkus.keycloak.policy-enforcer."paths".claim-information-point."simple-config"` WDYT? @gsmet 

Also, the description on the two config items could use some love. If we display the two keys just like suggested above with a proper description - what is the complex config or a simple config and when to use one or the other, an example usage will be useful. 

In the meantime we have empty java docs 
```
        @ConfigGroup
        public static class ClaimInformationPointConfig {

            /**
             *
             */
            Map<String, Map<String, Map<String, String>>> complexConfig;

            /**
             *
             */
            Map<String, Map<String, String>> simpleConfig;
        }
```

WDYT ? @pedroigor @sberyozkin

gsmet:
I hadn't followed this but you're right it looks very fishy.

gsmet:
So... the config property is OK but... the type is not. In the MongoDB case I put in the original issue, you need the type to be `Map<String, String>` not `String` as it is now.

machi1990:
@gsmet I just rebased with master. This one is ready for review now.

gsmet (approved):
OK, it fixes some very important issues with doc generation so let's merge it and backport it.

gsmet:
Thanks @machi1990 !

machi1990:
> Thanks @machi1990 !

My pleasure :-)

PR #4722, opened 2019-10-21
Rename Substrate* build items to NativeImage*

### LINKED ISSUE #4704: Rename Substrate build items
The `Substrate*` build items should be renamed to be `NativeImage*` instead.  The Graal people have repeatedly tried to pull back from the `Substrate` name, so it's probably best to avoid using product names period and instead name these items by their function.

Not really sure if this is a "bug" or an "enhancement".  Maybe we need another category. :)

gwenneg:
Fixes #4704


gsmet:
Do we really want to potentially break all the external extensions out there? It seems to me we should deprecate things properly and add new classes but better discuss it.

@dmlloyd ?

dmlloyd:
Better to do it now than after 1.0 IMO, but deprecating things properly seems like a reasonable idea.  I'm just not quite sure of the best way to go about it....

gwenneg:
It seems all files that need some renaming are identified in the PR (I also took the liberty to replace some `Substrate` references with `native mode` in various comments).

I will now try to keep and deprecate the `Substrate*` classes to avoid breaking existing extensions.

lburgazzoli:
From the Camel Quarkus POV, no problem to switch to the new naming now as it is not gogin to be a user facing issue

gwenneg:
This PR is currently mergeable if forward compatibility is not a requirement.

Otherwise, I will continue to work on the non-breaking version tonight.

Sanne:
I'm not aware of many "work in progress" extensions out there; there's a couple but they don't even need Substrate* items so far.

IMO it should be safe to merge this now at minimal disruption.

gwenneg:
Then this PR is ready for a review, but I can still work on the non-breaking version if it's really needed.

gsmet:
@Sanne please have a look at the PR: it's moving `ReflectiveClassBuildItem` from one package to another and it's certainly the most used build item.

gwenneg:
Yes, there were several classes without `Substrate` in their name but still located in a `substrate` package that were moved to a `nativeimage` package.

Sanne:
my comment was related to the question before vs after 1.0. I'd do before 1.0.

Not meaning to interfere with the review you started @gsmet : your call if you want deprecations - I don't know how much work that is and what the tradeoffs are.

gsmet:
I think it would be a bad idea to break all the existing extensions out there now. But I also think we should do it the change before 1.0.

So my vote would be:
- let's do the change
- for the user facing ones (I don't care about the ones that are internal), let's keep the old ones and deprecate them for removal in future versions. Hopefully, there are not too many user facing ones.

gwenneg:
@gsmet I'll need some help to identify the classes that require deprecation.

gwenneg:
I'll be back on this PR tonight and I still need some help to determine exactly which classes need deprecation :)

gwenneg:
I started deprecating things but it's not finished yet (the PR should be ready for review tomorrow). You can still have a look at the code to check if the changes I introduced are OK (I haven't review it myself, I still have to double-check everything).

I duplicated some methods (one version using deprecated classes, the other one using renamed classes) so that we'll be able to easily delete the deprecated stuff with minimal changes to the rest of the code.

I'm not stashing commits for now, I want to keep one commit for the renaming and another one for deprecation while I'm working on the PR.

gsmet:
@gwenneg unfortunately most of them AFAICS:
```
ReflectiveClassBuildItem
ReflectiveFieldBuildItem
ReflectiveHierarchyBuildItem
ReflectiveMethodBuildItem
RuntimeInitializedClassBuildItem
RuntimeReinitializedClassBuildItem
NativeImageProxyDefinitionBuildItem
NativeImageResourceBuildItem
NativeImageResourceBundleBuildItem
NativeImageSystemPropertyBuildItem
ServiceProviderBuildItem
```

What I'm wondering is if we could have a processor with a `@BuildStep` per class, consuming the old classes and producing the new classes. If it works, that would be the best way of doing that IMHO.

gwenneg:
@gsmet Thanks for the list. I really like the processor idea, it would be much simpler to add, review, maintain and remove. I'll git it a try!

gwenneg:
Here's the implementation of @gsmet's suggestion about a processor converting deprecated classes to new classes.

Classes inside the `io.quarkus.deployment.builditem.nativeimage` are the same at the old ones, there were no code changes (apart from the renaming).

The deprecated stuff removal will be super simple: just delete the entire `io.quarkus.deployment.builditem.substrate` package and that's it.

I'm not sure of how this should be tested.

gsmet:
@gwenneg have you tested it works properly with a sample extension producing an old build item? I think it would be OK for me for now, until someone complains and we decide we need more tests.

gsmet (approved):

gsmet:
@gwenneg I let you confirm me you made a quick test with an old build item to check the processor machinery works ok.

gsmet:
Hmmm, too bad, we have a lot of conflicting files now. Don't know which PR have changed all this.

gwenneg:
It happened a few times since this PR was opened, I'll fix that easily. I'm running a test right now, I'll let you know when it's finished.

gwenneg:
Conflicts solved and commits squashed, but **DO NOT MERGE**, test is still running.

gsmet:
@lburgazzoli tested with the Camel extensions and it worked OK. So I think it's good enough.

gsmet:
Thanks @gwenneg and @lburgazzoli !

gwenneg:
@gsmet It feels like deprecation is worth being mentioned in the release note. Feel free to remove the label if you think otherwise :)

Issue #2241, opened 2019-04-26
PrintAnalysisCallTree cause a lot of heap allocation during native image build


johnaohara:
By default we enabled  -H:+PrintAnalysisCallTree for native binary builds.  For a simple rest-crud application (https://github.com/johnaohara/rest-http-crud) during a native image build;

1) With -H:+PrintAnalysisCallTree; 47.87GB of objects are allocated to the heap 
2) Without ; 36.02GB of objects are allocated on the heap

Disabling this option by default would save us 11.85GB of heap allocations (25.2%) during a native image build for the rest-crud application.

@stuartwdouglas I know that we have historically kept this option enabled by default, but given that it causes so much allocation on the heap, is it worth disabling by default for users building native images?

johnaohara:
Native image build time (for rest-crud application) also reduced from 103771 ms to 84908 ms (~20% reduction)

zeeZ:
We shaved 12 minutes off a 15 minute build by disabling this option.

```
2019-10-18T15:34:43.858328327Z [xxx-runner:24]    classlist:  11,035.77 ms
2019-10-18T15:34:45.343044601Z [xxx-runner:24]        (cap):     975.09 ms
2019-10-18T15:34:46.262863629Z [xxx-runner:24]        setup:   2,397.18 ms
2019-10-18T15:38:19.837367346Z [xxx-runner:24]   (typeflow): 155,318.64 ms
2019-10-18T15:38:19.837497343Z [xxx-runner:24]    (objects):  49,119.68 ms
2019-10-18T15:38:19.837650594Z [xxx-runner:24]   (features):   1,929.46 ms
2019-10-18T15:38:19.837863327Z [xxx-runner:24]     analysis: 213,322.67 ms
2019-10-18T15:47:50.863414370Z Printing call tree to /project/reports/...
2019-10-18T15:50:01.645147897Z Printing list of used classes to /project/reports/...
2019-10-18T15:50:01.877119695Z Printing list of used packages to /project/reports/...
2019-10-18T15:50:10.573867399Z [xxx-runner:24]     (clinit):   1,651.35 ms
2019-10-18T15:50:10.609210111Z [xxx-runner:24]     universe:   8,547.64 ms
2019-10-18T15:50:21.322212545Z [xxx-runner:24]      (parse):  10,540.83 ms
2019-10-18T15:50:32.650969959Z [xxx-runner:24]     (inline):  10,720.39 ms
[...]

PR #4734, opened 2019-10-22
Disable native image PrintAnalysisCallTree by default, change semantics to 'enabledReports=true'

johnaohara:
…cs to 'enabledReports=true'

Reduces heap allocation and increase build speed of a default native image build

### DIFF
--- a/devtools/maven/src/main/java/io/quarkus/maven/NativeImageMojo.java
+++ b/devtools/maven/src/main/java/io/quarkus/maven/NativeImageMojo.java
@@ -130,8 +130,12 @@ public class NativeImageMojo extends AbstractMojo {
     @Parameter(defaultValue = "true")
     private Boolean fullStackTraces;
 
+    @Deprecated
     @Parameter(defaultValue = "${native-image.disable-reports}")
     private Boolean disableReports;
+
+    @Parameter(defaultValue = "${native-image.enable-reports}")
+    private Boolean enableReports;
 
     @Parameter
     private List<String> additionalBuildArgs;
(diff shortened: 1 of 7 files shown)


johnaohara:
This PR resolves #2241 but also retains current behaviour, i.e. if users currently define `<disableReports>true</disableReports>` report generation will be disabled

devtools/maven/src/main/java/io/quarkus/maven/NativeImageMojo.java
```diff
@@ -333,7 +337,10 @@ public void accept(ConfigBuilder configBuilder) {
                     configs.add("quarkus.native.debug-symbols", debugSymbols.toString());
                 }
                 if (disableReports != null) {
-                    configs.add("quarkus.native.disable-reports", disableReports.toString());
+                    configs.add("quarkus.native.enable-reports", new Boolean(!disableReports).toString());
```
gsmet: Is it safe? Could `disableReports` be null?

gsmet (changes requested):
Added a small question.

devtools/maven/src/main/java/io/quarkus/maven/NativeImageMojo.java
johnaohara: `disableReports` null on line 340? there is a null check on line 339

gsmet: Oops sorry :).

gsmet (approved):

Sanne:
could we avoid the double negation? I almost misunderstood the option :)

I'd have `<enableReports>` an have it default to `false`.

johnaohara:
@Sanne That is what I have changed it to. Previously the configuration was defined as

gsmet:
Thanks @johnaohara !

PR #4736, opened 2019-10-22
Properly handle JSON-B serializer / deserializer in Kafka Client

geoand:
This uses the same idea that is used in Mongo Panache


geoand:
This doesn't make sense since the serializers / deserializers are Kafka ones, not JSON-B ones

PR #4739, opened 2019-10-22
Tighten integration between Kafka client and jsonb - Add Jackson support

geoand:
* Use Jsonb bean from Arc if possible
* Only register serializer / deserializer for reflection if JSON-B is present
* Add Jackson module that does the same things as the JSON-B module
* Update Kafka test to ensure that everything works when one of these json dependencies is missing

### DIFF
--- a/core/deployment/src/main/java/io/quarkus/deployment/Capabilities.java
+++ b/core/deployment/src/main/java/io/quarkus/deployment/Capabilities.java
@@ -12,6 +12,7 @@ public final class Capabilities extends SimpleBuildItem {
     public static final String CDI_ARC = "io.quarkus.cdi";
     public static final String SERVLET = "io.quarkus.servlet";
     public static final String TRANSACTIONS = "io.quarkus.transactions";
+    public static final String JSONB = "io.quarkus.jsonb";
     public static final String RESTEASY_JSON_EXTENSION = "io.quarkus.resteasy-json";
     public static final String SECURITY = "io.quarkus.security";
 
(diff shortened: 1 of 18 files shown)


gsmet:
A bit unrelated to that one but should we support Jackson too? I think this has probably been done at a time when we didn't really support it.

geoand:
Yeah, that makes sense. I'll add another commit to the PR.

extensions/kafka-client/runtime/src/main/java/io/quarkus/kafka/client/serialization/JsonProducer.java
```diff
@@ -0,0 +1,36 @@
...
+
+    // same comment as above but for Jackson's ObjectMapper
+    static ObjectMapper getObjectMapper() {
+        ObjectMapper objectMapper = null;
+        ArcContainer container = Arc.container();
+        if (container != null) {
+            objectMapper = container.instance(ObjectMapper.class).get();
+        }
+        return objectMapper != null ? objectMapper : new ObjectMapper();
+    }
```
gsmet: Won't you have an error if you only have one of the dependencies at runtime. I mean either `Jsonb` or `ObjectMapper` won't exist in the classpath?

geoand: I'm waiting for CI to tell me that :)

geoand: CI passed. I will however later on do some more manual tests and see how if there is anything that we might have missed.

geoand: Thanks for catching this!

The PR has been updated to make sure this is not a problem.

gsmet (approved):
Looks good now, thanks!

PR #4748, opened 2019-10-22
Pass on the relevant config to control whether the Narayana TSM starts (even for native mode)

tomjenkinson:
…s (even for native mode)


jaikiran:
Just a FYI @tomjenkinson - I manually applied this change locally and built the native image. With this change, I no longer see the previous port binding message, so this looks good.

gsmet (approved):

gsmet:
@jaikiran thanks for checking!

Sanne:
awesome, thanks all!

PR #4750, opened 2019-10-22
Ignore META-INF/quarkus-extension.json when building uberjars

gsmet:


aloubyansky (approved):

maxandersen:
wait - educate me here - where is uber jars made ? would loose metadata access in that case...might not be an issue but just wondering where this is used ?

gsmet:
It's used when the users want to build an uberJar for their application. So runner.jar + all the dependencies.

Obviously, you can't get several of the same files in the jar thus the issue.

If these information are not necessary at runtime then it shouldn't be an issue. But for sure, if you need them at runtime, it's gonna be a problem.

maxandersen:
thanks for the info - looking at all the other files that gets ignored I'm sure we'll be able to survive or at least be in as good a shape as anyone wanting to look up module info or other fun metadata ;)

Issue #4503, opened 2019-10-10
Agroal: failure to connect on boot is swallowed


Sanne:
**Describe the bug**
The datasource has an "housekeeping thread" which is being used to open the initial connections; when these are failed to be opened, exceptions are being swallowed as it's happening in this alterenative thread.

It's not a big deal as exceptions are shown properly when a connection is actually requested by application code, but it's a bit confusing as e.g. there's no visible exception on boot even though the DB is unreachable.

PR #4754, opened 2019-10-22
Add logging to the connection pool

barreiro:
Fix for #4503


Sanne (approved):

Sanne:
this seems to do the trick. thanks @barreiro !

I still don't get on-boot feedback, but that's not an Agroal problem: the datasource is started lazily because of how the beans are setup in Arc.

PR #4756, opened 2019-10-22
Arc - log a warning when we detect wildcard as a bean type.

### LINKED ISSUE #4292: Arc - parameterized type that contains a wildcard type parameter is not a legal bean type
Issue is based on a discussion (and code) in this PR - https://github.com/quarkusio/quarkus/pull/4279

> A parameterized type that contains a wildcard type parameter is not a legal bean type.

According to CDI specification, namely paragraph about [Legal Bean Types](https://docs.jboss.org/cdi/spec/2.0/cdi-spec.html#legal_bean_types), a param. type with wildcard is an illegal bean type.
Later on the spec says that:
> The resulting set of bean types for a managed bean consists only of legal bean types, all other types are removed from the set of bean types.

Hence this isn't really an error blowing the application up, it would just be silently ignored.

We should add some detection in Arc and log a warning if we encounter such bean type.

manovotn:
Fixes #4292 

This will log an `INFO` warning that looks like this:
`INFO: Producer method io.quarkus.arc.test.producer.generic.WildcardTest$Bar<? extends io.quarkus.arc.test.producer.generic.WildcardTest$Foo> create() contains a parameterized typed with a wildcard. This type is not a legal bean type according to CDI specification and will be ignored during bean resolution.`

### DIFF
--- a/independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/Types.java
+++ b/independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/Types.java
@@ -19,6 +19,7 @@ import java.util.Map;
 import java.util.Set;
 import java.util.function.BiConsumer;
 import org.jboss.jandex.AnnotationInstance;
+import org.jboss.jandex.AnnotationTarget;
 import org.jboss.jandex.AnnotationValue;
 import org.jboss.jandex.ClassInfo;
 import org.jboss.jandex.DotName;
@@ -31,12 +32,15 @@ import org.jboss.jandex.Type;
 import org.jboss.jandex.Type.Kind;
 import org.jboss.jandex.TypeVariable;
 import org.jboss.jandex.WildcardType;
+import org.jboss.logging.Logger;
 
 /**
  *
  * @author Martin Kouba
  */
 final class Types {
+
+    static final Logger LOGGER = Logger.getLogger(Types.class);
 
     private static final Type OBJECT_TYPE = Type.create(DotNames.OBJECT, Kind.CLASS);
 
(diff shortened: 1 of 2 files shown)


independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/Types.java
```diff
@@ -199,7 +202,29 @@ static Type getProviderType(ClassInfo classInfo) {
...
+    }
+
+    static Set<Type> getTypeClosure(ClassInfo classInfo, MethodInfo producerMethod,
+            Map<TypeVariable, Type> resolvedTypeParameters,
+            BeanDeployment beanDeployment, BiConsumer<ClassInfo, Map<TypeVariable, Type>> resolvedTypeVariablesConsumer) {
+        return getTypeClosure(classInfo, producerMethod, null, resolvedTypeParameters, beanDeployment,
+                resolvedTypeVariablesConsumer);
+    }
+
+    static Set<Type> getTypeClosure(ClassInfo classInfo, MethodInfo producerMethod, FieldInfo producerField,
```
mkouba: I would replace the `MethodInfo/FieldInfo` with `AnnotationTarget` so that we wouldn't have to introduce two more getTypeClose() signatures...

independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/Types.java
```diff
@@ -212,6 +237,14 @@ static Type getProviderType(ClassInfo classInfo) {
             Type[] typeParams = new Type[typeParameters.size()];
             for (int i = 0; i < typeParameters.size(); i++) {
                 typeParams[i] = resolvedTypeParameters.get(typeParameters.get(i));
+                // this should only be the case for producers; wildcard is not a legal bean type
+                // see https://docs.jboss.org/cdi/spec/2.0/cdi-spec.html#legal_bean_types
+                if (typeParams[i].kind().equals(Kind.WILDCARD_TYPE) && (producerField != null || producerMethod != null)) {
+                    LOGGER.info("Producer " + (producerMethod == null ? "field " : "method ") +
```
mkouba: I wonder if we should skip the type as well, ie. do not add it to the set of types. This is what we do in Weld, right?

mkouba (changes requested):

mkouba (approved):
Looks good.

PR #4757, opened 2019-10-22
Make Jackson + JSON-B @JsonSerializer/Deserializer for fields work in native

geoand:


extensions/jackson/deployment/src/main/java/io/quarkus/jackson/deployment/JacksonProcessor.java
```diff
@@ -73,21 +78,40 @@ void register() {
...
+                AnnotationValue usingValue = deserializeInstance.value("using");
+                if (usingValue != null) {
+                    // the Deserializers are constructed internally by Jackson using a no-args constructor
+                    reflectiveClass.produce(new ReflectiveClassBuildItem(false, false, usingValue.asClass().toString()));
+                }
+            }
+        }
+
+        // handle the various @JsonSerialize cases
+        for (AnnotationInstance serializeInstance : index.getAnnotations(JSON_SERIALIZE)) {
```
gsmet: Why do we only add the class hierarchy for reflection in the deserialize case?

integration-tests/pom.xml
```diff
@@ -65,6 +65,7 @@
         <module>neo4j</module>
         <module>mongodb-client</module>
         <module>jackson</module>
+        <module>json</module>
```
gsmet: Please call it jsonb. And it needs to be added to the native builds pipeline, otherwise it won't be executed on CI.

gsmet (changes requested):
Looks mostly good. I added a couple of comments.

integration-tests/pom.xml
geoand: I am surprised I didn't do this the first time around :)

extensions/jackson/deployment/src/main/java/io/quarkus/jackson/deployment/JacksonProcessor.java
geoand: We register the hierarchy when using the annotation on a field or method in either case.
The hierarchy comes into to play for deserialization when using `@JsonDeserialize` on a class which is a different case.
For the deserialization case we register the hierarchy when the annotation is applied to a Class.

Does that make sense?

integration-tests/pom.xml
geoand: Fixed

gsmet (approved):

extensions/jackson/deployment/src/main/java/io/quarkus/jackson/deployment/JacksonProcessor.java
gsmet: I suppose so :)

geoand: I cleaned up what I wrote a little more for posterity :)

gsmet:
CI failures are unrelated (Azure issues).

PR #4760, opened 2019-10-22
Security guides tweaks

FroMage:
@stuartwdouglas let's do the discussion about security here then.

First, here are some doc tweaks.

Now, questions:

- Why is the security-jdbc extension not listed in the main security guide along with the other extensions: https://quarkus.io/guides/security-guide#security-implementations ?
- On the main security guide: https://quarkus.io/guides/security-guide#securing-web-endpoints presents a way to secure by path, but doesn't say how it compares to and combines with the annotation-based way?
- "Policies are pluggable so it is possible for other Quarkus extensions to provide additional policies" on the main security guide. OK, but how? Link to the extension author guide's relevant section? Or example?
- I've added examples to the rules for matching paths and methods, but while the first rule says that longer paths match first, other rules then go on to say that more specific methods match first, and I'm not sure how the two rules combine.
- Some guides use `SecurityContext` to get the identity (main guide, OAuth, JWT), OIDC uses `SecurityIdentity` and JWT uses `JsonWebToken`. I guess I understand why JWT is special, but why does OIDC have its own class and does not mention `SecurityContext`?

### DIFF
--- a/docs/src/main/asciidoc/security-guide.adoc
+++ b/docs/src/main/asciidoc/security-guide.adoc
@@ -260,9 +260,8 @@ for JAX-RS or `JsonWebToken` for JWT.
 
 There are two configuration settings that alter the RBAC behavior:
 
-- `quarkus.jaxrs.deny-uncovered=true|false` - if set to true, the access will be denied for all JAX-RS endpoints will be
-denied by default. That is if the security annotations do not define the access control.
-Defaults to `false`
+- `quarkus.jaxrs.deny-uncovered=true|false` - if set to true, the access will be denied for all JAX-RS endpoints by default.
+That is if the security annotations do not define the access control. Defaults to `false`
 - `quarkus.security.deny-unannotated=true|false` - if set to true, the access will be denied to all CDI methods
 and JAX-RS endpoints that do not have security annotations but are defined in classes that contain methods with
 security annotations. Defaults to `false`.
--- a/docs/src/main/asciidoc/security-guide.adoc
+++ b/docs/src/main/asciidoc/security-guide.adoc
@@ -260,9 +260,9 @@ for JAX-RS or `JsonWebToken` for JWT.
 
 There are two configuration settings that alter the RBAC behavior:
 
-- `quarkus.jaxrs.deny-uncovered=true|false` - if set to true, the access will be denied for all JAX-RS endpoints by default.
+- `quarkus.security.jaxrs.deny-unannotated-endpoints=true|false` - if set to true, the access will be denied for all JAX-RS endpoints by default.
 That is if the security annotations do not define the access control. Defaults to `false`
-- `quarkus.security.deny-unannotated=true|false` - if set to true, the access will be denied to all CDI methods
+- `quarkus.security.deny-unannotated-members=true|false` - if set to true, the access will be denied to all CDI methods
 and JAX-RS endpoints that do not have security annotations but are defined in classes that contain methods with
 security annotations. Defaults to `false`.
 
--- a/docs/src/main/asciidoc/security-guide.adoc
+++ b/docs/src/main/asciidoc/security-guide.adoc
@@ -133,6 +133,9 @@ public class SubjectExposingResource {
 Quarkus has an integrated plugable web security layer. If security is enabled all HTTP requests will have a permission
 check performed to make sure they are permitted to continue.
 
+NOTE: Configuration authorization checks are executed before any annotation-based authorization check is done, so both
+checks have to pass for a request to be allowed.
+
 The default implementation allows you to define permissions using config in `application.properties`. An example
 config is shown below:
 
--- a/docs/src/main/asciidoc/security-guide.adoc
+++ b/docs/src/main/asciidoc/security-guide.adoc
@@ -163,8 +163,6 @@ which permit all, deny all and only allow authenticated users respectively.
 It is also possible to define role based policies, as shown in the example. These policies will only allow users with the
 specified roles to access the resources.
 
-Policies are pluggable so it is possible for other Quarkus extensions to provide additional policies.
-
 ### Matching on paths, methods
 
 Permission sets can also specify paths and methods as a comma separated list. If a path ends with '*' then it is considered
--- a/docs/src/main/asciidoc/security-guide.adoc
+++ b/docs/src/main/asciidoc/security-guide.adoc
@@ -7,7 +7,9 @@ https://github.com/quarkusio/quarkus/tree/master/docs/src/main/asciidoc
 
 include::./attributes.adoc[]
 
-Quarkus comes with build in security to allow for RBAC (role based access control)
+## Securing REST endpoints
+
+Quarkus comes with build in security to allow for Role-Based Access Control (link:https://en.wikipedia.org/wiki/Role-based_access_control[RBAC])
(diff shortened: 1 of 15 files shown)


sberyozkin:
@FroMage I've created #4792 to make sure those users who prefer to work with `SecurityContext` should be able to continue doing so when `SecurityContext` API is sufficient for them

stuartwdouglas:
LGTM

stuartwdouglas:
1) JDBC not being linked is a bug
2) Paths are matched first, then annotations. Both need to allow a request for it to be permitted
3) There are no examples of this at the moment. Maybe just remove it till we have a concrete example
4) Longest path matches, but if you have two of the same patch and one of them has a method that matches and the other one does not only the method matching one is taken into account.
5) SecurityIdentity is our internal representation, the other ones mentioned are defined by various specs. SecurityIdentity is usable anywhere, but the spec specific ones may have limited scope (e.g. JAX-RS SecurityContext is only usable in JAX-RS requests).

FroMage:
> Longest path matches, but if you have two of the same patch and one of them has a method that matches and the other one does not only the method matching one is taken into account.

This runs contrary to what I documented then, no?

### Matching multiple paths: longest wins

Matching is always done on a longest path basis, less specific permission sets are not considered if a more specific one
has been matched:

```
quarkus.http.auth.permission.permit1.paths=/public/*
quarkus.http.auth.permission.permit1.policy=permit
quarkus.http.auth.permission.permit1.methods=GET,HEAD

quarkus.http.auth.permission.deny1.paths=/public/forbidden-folder/*
quarkus.http.auth.permission.deny1.policy=deny
```

TIP: Given the above permission set, `GET /public/forbidden-folder/foo` would match both permission sets' paths, but because it matches the `deny1` permission set's path on a longer match, `deny1` will be chosen and the request will be rejected.

FroMage:
Push my latest update of the guide.

docs/src/main/asciidoc/security-guide.adoc
```diff
@@ -134,29 +135,109 @@ which permit all, deny all and only allow authenticated users respectively.
...
+quarkus.http.auth.permission.roles2.paths=/api/*,/admin/*
+quarkus.http.auth.permission.roles2.policy=admin-policy1
+--
+
+TIP: Given the above permission set, `GET /api/foo` would match both permission sets' paths,
+so would require both the `user` and `admin` roles.
 
+## Authenticated representation
+
+For every authenticated resource, you can inject an `SecurityIdentity` instance to get the authenticated identity information.
```
gwenneg:
```suggestion
For every authenticated resource, you can inject a `SecurityIdentity` instance to get the authenticated identity information.
```

docs/src/main/asciidoc/security-guide.adoc
```diff
@@ -134,29 +135,109 @@ which permit all, deny all and only allow authenticated users respectively.
...
+
+In some other contexts you may have other parallel representations of the same information (or parts of it) such as `SecurityContext`
+for JAX-RS or `JsonWebToken` for JWT.
+
+## Configuration
+
+There are two configuration settings that alter the RBAC behavior:
+
+- `quarkus.jaxrs.deny-uncovered=true|false` - if set to true, the access will be denied for all JAX-RS endpoints will be
+denied by default. That is if the security annotations security annotations do not define the access control.
```
gwenneg:
```suggestion
denied by default. That is if the security annotations do not define the access control.
```

The previous sentence also needs to be corrected, there's too much denial in it :)

gwenneg (changes requested):
I added a few suggestions.

stuartwdouglas:
> > Longest path matches, but if you have two of the same patch and one of them has a method that matches and the other one does not only the method matching one is taken into account.
> 
> This runs contrary to what I documented then, no?
> 
> ### Matching multiple paths: longest wins
> Matching is always done on a longest path basis, less specific permission sets are not considered if a more specific one
> has been matched:
> 
> ```
> quarkus.http.auth.permission.permit1.paths=/public/*
> quarkus.http.auth.permission.permit1.policy=permit
> quarkus.http.auth.permission.permit1.methods=GET,HEAD
> 
> quarkus.http.auth.permission.deny1.paths=/public/forbidden-folder/*
> quarkus.http.auth.permission.deny1.policy=deny
> ```
> 
> TIP: Given the above permission set, `GET /public/forbidden-folder/foo` would match both permission sets' paths, but because it matches the `deny1` permission set's path on a longer match, `deny1` will be chosen and the request will be rejected.

This is correct. Longest path always wins, so once it has matched the longest path none of the other rules are considered (but there could be multiple rules with the same longest path).

If the method matches at least one of the rules then only the rules with the method match are taken into account. All of them must permit the request for it to be allowed through.

If there is some rules that matched based on path that do not have a method match, and the request did not match any methods then these rules will take effect.

If every rule had method matching criteria, and the request did not match any of them then it is implicitly denied. e.g. if you have a rules that matches /foo with GET and PUT permissions, but no other rule for /foo, then all other method types are implicitly denied.

stuartwdouglas:
> @stuartwdouglas I don't understand why `quarkus.jaxrs.deny-uncovered` is not called `quarkus.jaxrs.deny-unannotated` like `quarkus.security.deny-unannotated`.
> 
> It appears to require annotations, and not config-paths:
> 
> > if set to true, the access will be denied for all JAX-RS endpoints will be
> > denied by default. That is if the security annotations security annotations do not define the access control

This is what it is called in the Servlet spec so I just suggested that because it makes sense to me, but I agree that deny-unannotated makes more sense. Can you change it?

FroMage:
> This is correct

OK cool, then I didn't add wrong examples.

> This is what it is called in the Servlet spec so I just suggested that because it makes sense to me, but I agree that deny-unannotated makes more sense. Can you change it?

OK, will do.

emmanuelbernard:
Not a massively strong opinion but yes quarkus.security.jax-rs (is it jaxrs or jax-rs?) feels better.
how about `quarkus.security.jaxrs.deny-unannotated-endpoints` (see endpoints at the end to be clearer)

FroMage:
Here's what I ended up with:

- `quarkus.security.jaxrs.deny-unannotated-endpoints`: for jax-rs only, default to deny
- `quarkus.security.deny-unannotated-members`: for jax-rs and cdi, default to deny if other members have security annotations

That second one is impossible to describe properly. At least its documentation is clear, so hopefully that will help.

sberyozkin:
@FroMage Hi Stephane, since you are updating these guides, can you please check if #4796 can be fixed along the way ? My understanding TOC has to be added.

sberyozkin:
@FroMage I can try to do a quick #4796 right now if you are busy with something else...

docs/src/main/asciidoc/security-guide.adoc
FroMage: I've fixed it.

FroMage:
Just rebased on master.

@gwenneg can you change your review please? I've fixed your last issue.

stuartwdouglas (approved):

gwenneg:
> 
> 
> Just rebased on master.
> 
> @gwenneg can you change your review please? I've fixed your last issue.

Thanks for taking my suggestions into account! I'd like to approve the changes but it no longer seems possible nor useful since the PR was merged :)

PR #4770, opened 2019-10-22
Add descriptions to the extensions

### LINKED ISSUE #4724: Fix extension descriptions in the generated JSON
The maven plugin generating `quarkus-extension.json` file is taking description from the pom.xml file. Most of the extensions don't actually include the description. In that case the description gets inherited from the closest parent that has it. Currently what ends up in the JSON file looks completely wrong.

gastaldi:
Fixes #4724


aloubyansky (approved):
Great, thanks a lot @gastaldi !

PR #4775, opened 2019-10-22
Change Kafka Stream config topics type from String to List<String>

### LINKED ISSUE #4706: Kafka Streams - Config - Topics should be a list
Right now, it's a string but it should rather be a proper list in config:

https://github.com/quarkusio/quarkus/blob/master/extensions/kafka-streams/runtime/src/main/java/io/quarkus/kafka/streams/runtime/KafkaStreamsRuntimeConfig.java#L37

@gunnarmorling can you have a look? Or maybe you had a particular reason to do it this way?

lhauspie:
fixes #4706

### DIFF
--- a/extensions/kafka-streams/runtime/src/main/java/io/quarkus/kafka/streams/runtime/KafkaStreamsRuntimeConfig.java
+++ b/extensions/kafka-streams/runtime/src/main/java/io/quarkus/kafka/streams/runtime/KafkaStreamsRuntimeConfig.java
@@ -3,6 +3,7 @@ package io.quarkus.kafka.streams.runtime;
 import java.net.InetSocketAddress;
 import java.util.List;
 import java.util.Optional;
+import java.util.stream.Collectors;
 
 import io.quarkus.runtime.annotations.ConfigItem;
 import io.quarkus.runtime.annotations.ConfigPhase;
@@ -30,15 +31,19 @@ public class KafkaStreamsRuntimeConfig {
     public Optional<String> applicationServer;
 
     /**
-     * A comma-separated list of topic names processed by this stream processing application.
+     * A comma-separated list of topic names.
      * The pipeline will only be started once all these topics are present in the Kafka cluster.
      */
     @ConfigItem
-    public Optional<String> topics;
+    public List<String> topics;
 
     @Override
     public String toString() {
         return "KafkaStreamsRuntimeConfig [applicationId=" + applicationId + ", bootstrapServers=" + bootstrapServers
                 + ", applicationServer=" + applicationServer + ", topics=" + topics + "]";
+    }
+
+    public List<String> getTrimmedTopics() {
+        return topics.stream().map(String::trim).collect(Collectors.toList());
     }
 }
--- a/extensions/kafka-streams/runtime/src/main/java/io/quarkus/kafka/streams/runtime/KafkaStreamsTopologyManager.java
+++ b/extensions/kafka-streams/runtime/src/main/java/io/quarkus/kafka/streams/runtime/KafkaStreamsTopologyManager.java
@@ -1,8 +1,7 @@
 package io.quarkus.kafka.streams.runtime;
 
 import java.net.InetSocketAddress;
-import java.util.Arrays;
-import java.util.Collections;
+import java.util.Collection;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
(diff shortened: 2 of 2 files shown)


extensions/kafka-streams/runtime/src/main/java/io/quarkus/kafka/streams/runtime/KafkaStreamsRuntimeConfig.java
```diff
@@ -34,11 +34,11 @@
...
      */
     @ConfigItem
-    public Optional<String> topics;
+    public List<String> topics;
 
     @Override
     public String toString() {
         return "KafkaStreamsRuntimeConfig [applicationId=" + applicationId + ", bootstrapServers=" + bootstrapServers
-                + ", applicationServer=" + applicationServer + ", topics=" + topics + "]";
+                + ", applicationServer=" + applicationServer + ", topics=[" + String.join(",", topics) + "]]";
```
gsmet: You could have just used `topics`. The default `toString()` will do what you did automatically.

gsmet (changes requested):
Added one comment and could you squash the 2 commits into one?

Thanks!

gunnarmorling (commented):
LGTM. One question perhaps: would this trim any whitespace? Not that the previous implementation was doing this, but this would be nice to have, as it could be a common pit-fall.

gsmet:
I don't think MP Config is trimming the whitespace for you from what I saw in the code.

lhauspie:
> I don't think MP Config is trimming the whitespace for you from what I saw in the code.

You're right, I tested it this morning and they're not.

Let me add the triming ;)

extensions/kafka-streams/runtime/src/main/java/io/quarkus/kafka/streams/runtime/KafkaStreamsRuntimeConfig.java
lhauspie: You're right. I will revert this.
Thanks for th review ;)

lhauspie:
@gsmet I would like to add triming just after the "topics" config is set.
Is there any annotation to call a method after the config bean initialization ?

gsmet:
I don’t think so. You will have to do it when you consume them.

lhauspie:
Maybe I can add this line at the very beginning of the `KafkaStreamsTopologyManager#onStart` method ?
```Java
runtimeConfig.topics=runtimeConfig.topics.stream().map(String::trim).collect(Collectors.toList());
```

So no need to always remind to trim the `topics` if we need to use them again.

extensions/kafka-streams/runtime/src/main/java/io/quarkus/kafka/streams/runtime/KafkaStreamsTopologyManager.java
```diff
@@ -103,22 +100,17 @@ void onStart(@Observes StartupEvent ev) {
             return;
         }
 
+        runtimeConfig.topics = runtimeConfig.topics.stream().map(String::trim).collect(Collectors.toList());
```
gsmet: I'm not excited about this: I don't think we should change the config value.

Can you use a local variable instead?

Thanks!

gsmet (changes requested):

extensions/kafka-streams/runtime/src/main/java/io/quarkus/kafka/streams/runtime/KafkaStreamsTopologyManager.java
lhauspie: I refactored this piece of code.

I added the `getTrimedTopics()` method in the `KafkaStreamsRuntimeConfig` class to avoid futur code duplication about triming on `topics`.

extensions/kafka-streams/runtime/src/main/java/io/quarkus/kafka/streams/runtime/KafkaStreamsRuntimeConfig.java
```diff
@@ -30,15 +31,19 @@
...
-    public Optional<String> topics;
+    public List<String> topics;
 
     @Override
     public String toString() {
         return "KafkaStreamsRuntimeConfig [applicationId=" + applicationId + ", bootstrapServers=" + bootstrapServers
                 + ", applicationServer=" + applicationServer + ", topics=" + topics + "]";
     }
+
+    public List<String> getTrimedTopics() {
```
gsmet: Sorry to be picky but I think it should be `getTrimmedTopics()` with 2 m.

Could you fix that and amend your commit? Then I think we are good to go!

lhauspie: Damn, you're right.

It's fixed now.
Thanks for review.

gsmet (approved):

gsmet:
Merged, thanks!

PR #4776, opened 2019-10-22
[fixes #4495] - Quarkus Keycloak Authorization Extension

### LINKED ISSUE #4495: OIDC adapter has to provide HttpPermissionChecker
**Description**
Example of the original `quarkus-keycloak` configuration:
```
quarkus.keycloak.policy-enforcer.claim-information-point.claims.request-uri={request.relativePath}
quarkus.keycloak.policy-enforcer.claim-information-point.claims.request-method={request.method}
quarkus.keycloak.policy-enforcer.lazy-load-paths=true

quarkus.keycloak.policy-enforcer.paths.1.path=/asset-service/resources/tenants
quarkus.keycloak.policy-enforcer.paths.1.methods.1.method=POST
quarkus.keycloak.policy-enforcer.paths.1.methods.1.scopes=tenants:create

quarkus.keycloak.policy-enforcer.paths.10.path=/asset-service/resources/tenants/{tenantId}
quarkus.keycloak.policy-enforcer.paths.10.methods.1.method=PUT
quarkus.keycloak.policy-enforcer.paths.10.methods.1.scopes=tenant:edit
quarkus.keycloak.policy-enforcer.paths.10.methods.2.method=DELETE
quarkus.keycloak.policy-enforcer.paths.10.methods.2.scopes=tenant:delete
quarkus.keycloak.policy-enforcer.paths.11.path=/asset-service/resources/tenants/{tenantId}/disable
quarkus.keycloak.policy-enforcer.paths.11.methods.1.method=PUT
quarkus.keycloak.policy-enforcer.paths.11.methods.1.scopes=tenant:disable
quarkus.keycloak.policy-enforcer.paths.12.path=/asset-service/resources/tenants/{tenantId}/enable
quarkus.keycloak.policy-enforcer.paths.12.methods.1.method=PUT
quarkus.keycloak.policy-enforcer.paths.12.methods.1.scopes=tenant:enable
```
@dfranssen:

E.g. Keycloak javascript permission using the request-uri. Our path is like /asset-service/resources/tenants/abc where abc is the tenant name (uriPart[4] in the code below) for which we check the actual authenticated user is part of. In Keycloak we have a group per tenant (e.g. /abc/users) where the user is put into.
```
var attributes = context.getAttributes();
var httpUri = attributes.getValue('request-uri');
function isInTenantGroup(tenantId) {
    var result = false;
    var wanted = "/" + tenantId + "/users";
    var identity = context.getIdentity();
    var userGroups = identity.getAttributes().getValue('user-groups');
    if (userGroups !== null) {
        for(var i=0; i < userGroups.size(); i++) {
            if(wanted === userGroups.asString(i)) {
                result = true;
                break;
            }
        }
    }
    return result;
}

if (httpUri) {
    var uriParts = httpUri.asString(0).split('/');
    if (isInTenantGroup(uriParts[4])) {
        $evaluation.grant();
    }
}
```

**Implementation ideas**
The adapter has to introspect a `scope` claim (or Keycloak specific claim - in this case we should have a claim name configured to avoid tying the adapter to KC -  though we can do it later). PolicyEnforcer configuration group can be introduced if needed

pedroigor:
Quickstart: https://github.com/quarkusio/quarkus-quickstarts/pull/331


extensions/keycloak-authorization/runtime/src/main/java/io/quarkus/keycloak/pep/KeycloakPolicyEnforcerAuthorizer.java
```diff
@@ -0,0 +1,191 @@
...
+import io.quarkus.oidc.OidcConfig;
+import io.quarkus.security.identity.SecurityIdentity;
+import io.quarkus.security.runtime.QuarkusSecurityIdentity;
+import io.quarkus.vertx.http.runtime.security.HttpAuthorizer;
+import io.quarkus.vertx.http.runtime.security.QuarkusHttpUser;
+import io.vertx.ext.web.RoutingContext;
+
+@Singleton
+@AlternativePriority(1)
+public class KeycloakPolicyEnforcerAuthorizer extends HttpAuthorizer {
```
stuartwdouglas: So this is not really integrating with our HTTP permission layer, but just completely replacing it?

We need to discuss this, as it means that all other security rules will be ignored.

extensions/keycloak-authorization/runtime/src/main/java/io/quarkus/keycloak/pep/KeycloakPolicyEnforcerAuthorizer.java
```diff
@@ -0,0 +1,191 @@
...
+
+    private KeycloakAdapterPolicyEnforcer delegate;
+
+    @Override
+    public CompletionStage<SecurityIdentity> checkPermission(RoutingContext routingContext) {
+        VertxHttpFacade httpFacade = new VertxHttpFacade(routingContext);
+        AuthorizationContext result = delegate.authorize(httpFacade);
```
stuartwdouglas: This is a blocking operation, we need to make sure we are not on an IO thread. This API will need to be changed to add an easy way to run blocking operations.

extensions/keycloak-authorization/runtime/src/main/java/io/quarkus/keycloak/pep/KeycloakPolicyEnforcerConfig.java
```diff
@@ -0,0 +1,192 @@
...
+         * Specifies how policies are enforced.
+         */
+        @ConfigItem(defaultValue = "ENFORCING")
+        String enforcementMode;
+
+        /**
+         * Specifies the paths to protect.
+         */
+        @ConfigItem
+        Map<String, PathConfig> paths;
```
stuartwdouglas: Is there any reason that this can't use the existing path stuff from our current authorization layer? The original idea was that the policies would be pluggable, and there would be one place to map paths to policies.

stuartwdouglas (changes requested):

extensions/keycloak-authorization/runtime/src/main/java/io/quarkus/keycloak/pep/KeycloakPolicyEnforcerAuthorizer.java
stuartwdouglas: Actually I think this needs two layers, the existing policy based path stuff, and the ability to add additional checks that do their own path matching (which servlet web.xml integration will need).

sberyozkin:
@pedroigor thanks, I'll be commenting as well during the next few days. One thing is that we indeed should try to reuse as much as possible from the lower levels like the core security and oidc, and if it is not possible, push down some changes. (re the comment from @stuartwdouglas about the paths for ex)

docs/src/main/asciidoc/keycloak-authorization-guide.adoc
```diff
@@ -0,0 +1,269 @@
...
+The OpenID Connect extension allows you to define the adapter configuration using the `application.properties` file which should be located at the `src/main/resources` directory.
+
+=== Configuring using the application.properties file
+
+[source,properties]
+----
+# OIDC Configuration
+quarkus.oidc.auth-server-url=http://localhost:8180/auth/realms/quarkus
+quarkus.oidc.client-id=backend-service
+quarkus.oidc.credentials.secret=secret
```
sberyozkin: Is this `secret` actually used by the enforcer ? Just checking, it was not used in the `integration-tests/oidc`, but may be the enforcer does need it

extensions/keycloak-authorization/deployment/pom.xml
```diff
@@ -0,0 +1,63 @@
...
+    <parent>
+        <artifactId>quarkus-keycloak-authorization-parent</artifactId>
+        <groupId>io.quarkus</groupId>
+        <version>999-SNAPSHOT</version>
+        <relativePath>../</relativePath>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>quarkus-keycloak-authorization-deployment</artifactId>
+    <name>Quarkus - Keycloak Policy Enforcer - Deployment</name>
```
sberyozkin: This should probably be `Quarkus - Keycloak Authorization - Deployment` (same for the /runtime) but the description can indeed be more specific

extensions/keycloak-authorization/runtime/src/main/java/io/quarkus/keycloak/pep/KeycloakPolicyEnforcerAuthorizer.java
```diff
@@ -0,0 +1,191 @@
...
+
+        if (result.isGranted()) {
```
sberyozkin: There probably should be some abstract code based on what Stuart @stuartwdouglas did, so that the default authorizer and this one do not duplicate. 
What was the problem with only plugging in a custom `HttpSecurityPolicy` ?

extensions/keycloak-authorization/runtime/src/main/resources/META-INF/quarkus-extension.json
```diff
@@ -0,0 +1,14 @@
...
+        "oauth2",
+        "openid-connect",
+        "keycloak",
+        "authorization-services",
+        "policy-enforcer",
+        "fine-grained-permission",
+        "resource-based-authorization"
+    ],
+    "guide": "https://quarkus.io/guides/keycloak-authorization-guide"
+}
```
gsmet: Please also add it in the legacy json for now: ./devtools/platform-descriptor-legacy/src/main/filtered/extensions.json

gsmet (changes requested):
I reviewed the doc and added some suggestions.

Question is: should this extension go in if it doesn't support native? Until now, it was a strict condition.

extensions/keycloak-authorization/runtime/src/main/java/io/quarkus/keycloak/pep/KeycloakPolicyEnforcerConfig.java
```diff
@@ -0,0 +1,192 @@
...
+         */
+        @ConfigItem(defaultValue = "true")
+        boolean lazyLoadPaths;
+
+        /**
+         * Defines a URL where a client request is redirected when an "access denied" message is obtained from the server.
+         * By default, the adapter responds with a 403 HTTP status code
+         */
+        @ConfigItem
+        Optional<String> onDenyRedirectTo;
```
sberyozkin: I believe this has to be managed by the `quarkus-oidc` - this is exactly what it will already do as part of your authorization code flow :-) or the existing bearer-only check. I mean it will do 401 if the token is invalid (signature for ex), but if the token has not enough permissions, it is equally invalid. And we let the user be redirected to something local then we'll lose the flow control, what will happen after the user lands on some local page, as opposed to be redirected to IDP ?

extensions/keycloak-authorization/runtime/src/main/java/io/quarkus/keycloak/pep/KeycloakPolicyEnforcerConfig.java
```diff
@@ -0,0 +1,192 @@
...
+
+        /**
+         * Specifies that the adapter uses the UMA protocol.
+         */
+        @ConfigItem
+        boolean userManagedAccess;
```
sberyozkin: IMHO everything not related to the Keycloak specific policy-enforcer configuration should not be here at all; we've promised `quarkus-oidc` will work with not only KC :-), and UMA is supported not only by Keycloak

extensions/keycloak-authorization/runtime/src/main/java/io/quarkus/keycloak/pep/KeycloakPolicyEnforcerConfig.java
```diff
@@ -0,0 +1,192 @@
...
+            Map<String, Map<String, Map<String, String>>> complexConfig;
+
+            /**
+             *
+             */
+            @ConfigItem(name = ConfigItem.PARENT)
+            Map<String, Map<String, String>> simpleConfig;
+        }
+    }
+}
```
sberyozkin: I'm not sure how many of those properties are related to the KC policy enforcer configuration the user was trying to migrate. Please review and leave only those which are needed to support that policy-enforcer configuration; everything else can go later on to `quarkus-oidc` or `quarkus-oidc-uma` for example

integration-tests/keycloak-authorization/src/main/resources/keycloak.json
```diff
@@ -0,0 +1,26 @@
+{
```
sberyozkin: where is this `keycloak.json` resource used ?

extensions/keycloak-authorization/runtime/src/main/resources/META-INF/quarkus-extension.json
pedroigor: It is there already.

pedroigor:
> I reviewed the doc and added some suggestions.
> 
> Question is: should this extension go in if it doesn't support native? Until now, it was a strict condition.

I can work on native and it should not take too much time. We had native work for the old Keycloak extension before so is just a matter of doing the same changes.

@sberyozkin suggested leaving native for now though. But I can deliver it too.

extensions/keycloak-authorization/runtime/src/main/java/io/quarkus/keycloak/pep/KeycloakPolicyEnforcerAuthorizer.java
pedroigor: @stuartwdouglas this extension is kind of a replacement of Quarkus permissions. Differently than what we have in Quarkus, the policies are externalized, completely decoupled from the application, and centralized managed through Keycloak.

The functionality provided by our policy enforcer has more requirements in regards to path mapping and matching. We also have some requirements in regards to not authorized or forbidden responses (e.g.: when doing UMA).

This extension is all about bringing a Keycloak specific feature into Quarkus. We have users familiar with the configuration I'm proposing here, making easier to migrate their applications.

I'm not sure if it would make sense to have part of your resources protected by Quarkus policies and other parts protected by Keycloak. However, we could review the changes I'm proposing and delegate to the Quarkus permission layer when the extension denies access to a resource.

extensions/keycloak-authorization/runtime/src/main/java/io/quarkus/keycloak/pep/KeycloakPolicyEnforcerConfig.java
pedroigor: The key point here is that we are not just plugging in a new policy, but a completely different solution to authorization.

I tried to do it using what Quarkus provides, but that would mean bringing too many things that are really specific to this feature in Keycloak.

extensions/keycloak-authorization/runtime/src/main/java/io/quarkus/keycloak/pep/KeycloakPolicyEnforcerAuthorizer.java
pedroigor: I don't think there is a duplication. The authorization process is different if you compare both solutions.

This extension just brings an existing and specific feature from Keycloak into Quarkus. It is related to centralized authorization and how to externalize authorization from applications based on a Resource-Based Access Control.

extensions/keycloak-authorization/runtime/src/main/java/io/quarkus/keycloak/pep/KeycloakPolicyEnforcerConfig.java
pedroigor: This is not really related to code flow. This option, in particular, is just providing an option to customize requests denied by the server. The user is already authenticated and if not, this page is not meant to start any flow.

For instance, instead of returning a 403 response the application may want to show a more fancy page.

extensions/keycloak-authorization/runtime/src/main/java/io/quarkus/keycloak/pep/KeycloakPolicyEnforcerConfig.java
pedroigor: I see. So we would be providing a separated extension for UMA support if someone demands it ?

Sounds good for me. I can remove UMA support from this extension.

extensions/keycloak-authorization/runtime/src/main/java/io/quarkus/keycloak/pep/KeycloakPolicyEnforcerConfig.java
pedroigor: Except for the `userManagedAccess` all other properties are specific to KC policy enforcer.

pedroigor:
@gsmet Native support added.

extensions/keycloak-authorization/runtime/src/main/java/io/quarkus/keycloak/pep/KeycloakPolicyEnforcerAuthorizer.java
pedroigor: Can we follow the same approach as you did to the `IdentityProviderManager`? So that we pass an object to authorizers similar to `AuthenticationRequestContext` (maybe rename it to become more generic) in order to run blocking tasks?

extensions/keycloak-authorization/runtime/src/main/java/io/quarkus/keycloak/pep/KeycloakPolicyEnforcerConfig.java
sberyozkin: @pedroigor hey Pedro, exactly, lets have it done either directly as part of `quarkus-oidc` (it is nor really that fat so may be it should be there) or a new extension like `quarkus-oidc-uma` - so that we can say Quarkus can be UMA protected with any IDP which does it, which will be massive. Let me open an issue to track it, it would probably be a copy and paste for you :-), may be the only minor effort would be required to replace the KC adapter client code with some basic HTTP client....

extensions/keycloak-authorization/runtime/src/main/java/io/quarkus/keycloak/pep/KeycloakPolicyEnforcerConfig.java
sberyozkin: @pedroigor It looks like the error can only occur in the process of the user authenticating (for the bearer only flow it is only 403 anyway). So if it is SPA then it is only really an adapter level issue what happens to a user view. And if it is a server web app code flow then the user has signed in at the IDP site, is being redirected now to the adapter, the token fails a check, and with custom page the user lands there seeing that the access is denied, But what is next ? The user stays authenticated ? Does not seem like the right idea.
The other thing, I doubt that in production people will want to display the details outside of its log :-)

Perhaps we should have an option for OIDC adapter itself return a page like "Your access to this resource has been denied, please contact the administrator, you will be redirected to IDO in 5 sec..." (and this page can be customized). 

But whatever we do here, it feels like it is not a KC specific extension concern since 403 will be possible at the `quarkus-oidc` level as well :-)

extensions/keycloak-authorization/runtime/src/main/java/io/quarkus/keycloak/pep/KeycloakPolicyEnforcerAuthorizer.java
sberyozkin: @pedroigor hi, so given we've agreed `UMA` will be managed differently, how will that be done, will it have its own authorization mechanism as well ? Should we give it a try and have the core authorization mechanism be pluggable/customizabl as needed to meet KV ploicy-enforcer requirements ?
I have to admit I don't quite understand, if you could explain what is too rigid/etc in the core one ? Sorry if I'm missing something...

sberyozkin:
@pedroigor @gsmet yes, since Pedro was having some issues I suggested leaving it for now in case if would block the PR, the user who can't migrate may not be using the native mode...Nice to see the native is supported now after all :-)

pedroigor:
@sberyozkin Yeah. I was missing something when facing that error. Sorry for the false alarm :)

extensions/keycloak-authorization/runtime/src/main/java/io/quarkus/keycloak/pep/KeycloakPolicyEnforcerConfig.java
pedroigor: Client-side, UMA is quite straight forward to implement. I agree that a separated extension is the best way.

extensions/keycloak-authorization/runtime/src/main/java/io/quarkus/keycloak/pep/KeycloakPolicyEnforcerConfig.java
pedroigor: That is the idea. This requirement came from the community a long time ago.

The user would just see this custom page with whatever the application wants to show. Either a simple and fancy "Access Denied" or something else.

If we want to have that on `quarkus-oidc`, that is fine and I can remove this.

extensions/keycloak-authorization/runtime/src/main/java/io/quarkus/keycloak/pep/KeycloakPolicyEnforcerAuthorizer.java
pedroigor: The core one is not rigid, on the contrary, it is quite flexible. I just couldn't find a nice way to plug the policy enforcer capabilities. 

However, that is not Quarkus fault. But the nature of this extension that aims a different authorization model.

If you look at the guide, for instance, you'll see that users can enable the policy enforcer without even specify a single path. Where the mapping between paths and policies are done by the extension itself, through some communication with the Keycloak server.

I see this extension as similar to any other that brings functionality from a 3rd party framework/library. We are also constantly involving the capabilities, adding new ones, etc. We may even end up in a situation where a new capability would require changes in Quarkus.

So, what Stuart did is flexible enough from both a user perspective in regards to declarative authorization (support for different/custom policies, reuse policies across different permissions, etc) as well as from a developer perspective, like in this case, where we are enabling centralized authorization for Quarkus apps.

extensions/keycloak-authorization/runtime/src/main/java/io/quarkus/keycloak/pep/KeycloakPolicyEnforcerConfig.java
stuartwdouglas: I won't hold up this PR for this, but we should create an issue to track it.

stuartwdouglas:
I have pushed an extra commit onto this that will run the native tests in CI. If this passes then I think we are good to go.

stuartwdouglas (approved):

extensions/keycloak-authorization/runtime/src/main/java/io/quarkus/keycloak/pep/KeycloakPolicyEnforcerConfig.java
sberyozkin: @pedroigor yes please lets remove if for the moment and then figure out where to have this property - the same concept should work across the OIDC bits in Quarkus

extensions/keycloak-authorization/runtime/src/main/java/io/quarkus/keycloak/pep/KeycloakPolicyEnforcerConfig.java
sberyozkin: @stuartwdouglas we get it in too fast and the users will see the extra properties in the wrong namespace and a new migration headache is on the way

sberyozkin:
@stuartwdouglas we don't have to rush. It is nearly ready but UMa does not belong at the KC level, similarly to the redirect page property - these are the extra bits which are not required for the KC policy-enforcer migration support. IMHO the consolidation around the quarkus-oidc is the key

sberyozkin (changes requested):
Hi Pedro, Stuart, @pedroigor @stuartwdouglas I'm only asking here not for some extra improvements, but only to limit what is currently exposed at the quarkus-keycloak-authorization configuration level, specifically `UMA-related and redirect related` configs.
My point really is we've gone for a major upgrade of the OIDC code, dropped the original keycloak extension, hence IMHO we should be trying to keep all what can be generic at the oidc level to avoid people migrating again when we will support oidc-uma, etc. If we get these properties going into 0.27.0 then it is a new concern. Hope you are all OK with it.

pedroigor:
@sberyozkin Sure. Both options are not available anymore.

sberyozkin:
@pedroigor thanks :-) and we'll get them reborn pretty soon for sure :-)

integration-tests/keycloak-authorization/src/main/java/io/quarkus/it/keycloak/ProtectedResource.java
```diff
@@ -0,0 +1,45 @@
...
+@Path("/api/permission")
+public class ProtectedResource {
+
+    @Inject
+    SecurityIdentity identity;
+
+    @GET
+    @Produces(MediaType.APPLICATION_JSON)
+    public List<Permission> permissions() {
+        if (identity.checkPermissionBlocking(new AuthPermission("Permission Resource"))) {
```
sberyozkin: @pedroigor I'll add a follow up issue for us to come up with Quarkus permission class so that the users can write a portable code with or without KC policy enforcer, not a prob for this PR 
CC @stuartwdouglas

sberyozkin (approved):

PR #4777, opened 2019-10-22
Native support for Amazon Lambda

patriot1burke:
* Still have to generate yaml files from OutputTargetBuildItem
* Needs some cleanup
* Needs doc update too


extensions/amazon-lambda/deployment/src/main/java/io/quarkus/amazon/lambda/deployment/LambdaConfig.java
```diff
@@ -1,12 +1,12 @@
...
+package io.quarkus.amazon.lambda.deployment;
 
 import java.util.Optional;
 
 import io.quarkus.runtime.annotations.ConfigItem;
 import io.quarkus.runtime.annotations.ConfigPhase;
 import io.quarkus.runtime.annotations.ConfigRoot;
 
-@ConfigRoot(phase = ConfigPhase.RUN_TIME)
+@ConfigRoot(phase = ConfigPhase.BUILD_TIME)
```
stuartwdouglas: This should really be runtime. The idea with the runtime config of the handler name was that you could create a single native image with multiple handlers, and use environment variables to select the handler to use at deployment time.

If this is build time and you have a deployment that has 10 different functions you would need 10 different native image builds.

stuartwdouglas (changes requested):

extensions/amazon-lambda/deployment/src/main/java/io/quarkus/amazon/lambda/deployment/LambdaConfig.java
patriot1burke: You shouldn't have a deployment with 10 different functions.

patriot1burke: Moving this back to runtime would also require moving error condition checks from build time to runtime which can be very problematic.  You don't want to have to debug build time issues at runtime.

docs/src/main/asciidoc/amazon-lambda-http-guide.adoc
```diff
@@ -3,36 +3,230 @@ This guide is maintained in the main Quarkus repository
...
+for installing AWS SAM CLI.
+
+== Creating the Maven Deployment Project
+
+Create the Quarkus AWS Lambda maven project using our Maven Archetype.
+
+
+[source, subs=attributes+]
+----
+mvn archetype:generate \
```
stuartwdouglas: This is fine for now as we don't have anything better, but eventually I hope we can have better create-project tooling that can be extended to handle cases like this.

extensions/amazon-lambda/runtime/src/main/java/io/quarkus/amazon/lambda/runtime/LambdaConfig.java
```diff
@@ -19,6 +19,14 @@
...
-    Optional<String> handler;
+    public Optional<String> handler;
+
+    /**
+     * If true, this will enable the aws event poll loop within a Quarkus test run. This loop normally only runs in native
+     * image. This option is strictly for testing purposes.
+     *
+     */
+    @ConfigItem
+    public boolean enablePollingJvmMode;
```
stuartwdouglas: Do we need this to be config, can't we handle this transparently? e.g. have the io.quarkus.amazon.lambda.test.LambdaResourceManager set a system property, and just look at the system property to decide if we are in a test environment.

stuartwdouglas: Either way it is not worth holding up the PR over it.

stuartwdouglas (approved):

extensions/amazon-lambda/runtime/src/main/java/io/quarkus/amazon/lambda/runtime/LambdaConfig.java
patriot1burke: I'll try that out in another PR.

PR #4784, opened 2019-10-23
Remove not used extension registry plugins

maxandersen:
Why:

 * want to improve Extension metadata and old registry plugins
   makes that harder.

 * remove the extension plugin.
 * next step after is to remove the old registry platform but that is
   currently being used.


maxandersen:
checked i can build (without tests) so lets see what the PR completes if we want to be sure it doesn't break any other IT tests.

maxandersen:
or @aloubyansky if you can verify you agree i'm fine applying this immediately ;)

aloubyansky (approved):

geoand:
CI isn't liking this :(

maxandersen:
> CI isn't liking this :(

my bad - removed a module but didn't remove the previous built one in .m2 cache thus my local build was running fine. pushed update that should hopefully get us further!

PR #4790, opened 2019-10-23
Upgrade to jackson-databind 2.9.10.1 to fix CVE

loicmathieu:
This upgrade will fix these new CVE-2019-16942, CVE-2019-16943, CVE-2019-17531

Unfortunatly, this version in not in the Jackson BOM 2.9.10 and there is no patch version of BOM so we have to add an explicit dependency management to it

### DIFF
--- a/bom/runtime/pom.xml
+++ b/bom/runtime/pom.xml
@@ -68,7 +68,7 @@
            is not available in Maven fast enough: -->
         <graal-sdk.version>19.2.0.1</graal-sdk.version>
         <gizmo.version>1.0.0.Alpha8</gizmo.version>
-        <jackson.version>2.9.10</jackson.version>
+        <jackson.version>2.9.10.20191020</jackson.version>
         <commons-logging-jboss-logging.version>1.0.0.Final</commons-logging-jboss-logging.version>
         <commons-lang3.version>3.8.1</commons-lang3.version>
         <commons-codec.version>1.11</commons-codec.version>


Sanne (approved):

gsmet:
We need @cescoffier 's blessing for it as IIRC we were waiting for some Vert.x feedback.

loicmathieu:
> We need @cescoffier 's blessing for it as IIRC we were waiting for some Vert.x feedback.

I think it's for 2.10 that we need Vert.x feedback. 2.9.10.1 is just a security update ...

bom/runtime/pom.xml
```diff
@@ -69,6 +69,8 @@
         <graal-sdk.version>19.2.0.1</graal-sdk.version>
         <gizmo.version>1.0.0.Alpha8</gizmo.version>
         <jackson.version>2.9.10</jackson.version>
+        <!-- Fix a databing version higher that the one from the BOM for CVE vulnerabilities: CVE-2019-16942, CVE-2019-16943, CVE-2019-17531 -->
+        <jackson-databind.version>2.9.10.1</jackson-databind.version>
```
jorsol: Adding another property and dependency is not need, the BOM is used in Quarkus, so the BOM  should be updated.

The Jackson BOM version corresponding to this fix is `2.9.10.20191020` so is enough to change
```xml
<jackson.version>2.9.10.20191020</jackson.version>
```

jorsol (changes requested):

jorsol (approved):

gsmet:
> I think it's for 2.10 that we need Vert.x feedback. 2.9.10.1 is just a security update ...

Ah yes, sorry I missed that it was a 2.9. I will merge it right away. Thanks!

PR #4798, opened 2019-10-23
Remove reference to deleted artifactId quarkus-undertow-common-substitutions

Sanne:
…tutions

This fixes the build, currently broken.

### DIFF
--- a/bom/runtime/pom.xml
+++ b/bom/runtime/pom.xml
@@ -532,11 +532,6 @@
             </dependency>
             <dependency>
                 <groupId>io.quarkus</groupId>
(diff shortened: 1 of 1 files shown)


FroMage (approved):

gsmet:
Where did it fail? Should we make some code changes to be more lenient about this sort of issues?

Sanne:
@gsmet the `Quarkus - Documentation` module would fail as it tries to extract metadata from each extension listed in the bom; it was a bit puzzling to relate it to this entry..

I wouldn't make it more lenient; the Documentation building is not production code, and it's kinda welcome that it brings such issues to our attention. A warning could have been better, but it's not an easy change.

PR #4799, opened 2019-10-23
Hard disable the Narayana TSM

tomjenkinson:


Sanne (approved):
Great, thanks.  Initially I thought this would also disable it for STM but that's not the case. Is @mmusgrov working on that or should I do it?

Issue #4782, opened 2019-10-23
Quarkus produces JAR that fails CRC32 checks


saturnism:
**Describe the bug**
```
# Create a new project
$ mvn io.quarkus:quarkus-maven-plugin:0.26.0:create \
    -DprojectGroupId=org.acme \
    -DprojectArtifactId=getting-started \
    -DclassName="org.acme.quickstart.GreetingResource" \
    -Dpath="/hello"
$ cd getting-started/

# Produce the JAR
$ ./mvnw package

# App runs fine
$ java -jar target/getting-started-1.0-SNAPSHOT-runner.jar

# There are CRC check errors in the produced JAR
$ unzip -tqql target/getting-started-1.0-SNAPSHOT-runner.jar                                                       ↵ META-INF/services/io.quarkus.arc.ComponentsProvider  bad CRC 55fb23e4  (should be 95e5dbd1)
META-INF/MANIFEST.MF    bad CRC 66911b11  (should be 3569519a)

# Alternative checks:
$ zip -T target/getting-started-1.0-SNAPSHOT-runner.jar
META-INF/services/io.quarkus.arc.ComponentsProvider  bad CRC 55fb23e4  (should be 95e5dbd1)
META-INF/MANIFEST.MF    bad CRC 66911b11  (should be 4f04ef90)
test of target/getting-started-1.0-SNAPSHOT-runner.jar FAILED

$ cd target; jar -xf getting-started-1.0-SNAPSHOT-runner.jar
java.io.IOException: Push back buffer is full
	at java.base/java.io.PushbackInputStream.unread(PushbackInputStream.java:229)
	at java.base/java.util.zip.ZipInputStream.readEnd(ZipInputStream.java:359)
	at java.base/java.util.zip.ZipInputStream.read(ZipInputStream.java:197)
	at java.base/java.util.zip.ZipInputStream.closeEntry(ZipInputStream.java:141)
	at jdk.jartool/sun.tools.jar.Main.extractFile(Main.java:1457)
	at jdk.jartool/sun.tools.jar.Main.extract(Main.java:1364)
	at jdk.jartool/sun.tools.jar.Main.run(Main.java:409)
	at jdk.jartool/sun.tools.jar.Main.main(Main.java:1681)
```

**Expected behavior**
No CRC check errors

**Actual behavior**
CRC check errors for unknown reason.

**To Reproduce**
Steps to reproduce the behavior: See above.

**Configuration**
None

**Screenshots**
N/A

[...]

PR #4801, opened 2019-10-23
Work around JDK bug to produce JAR files with proper CRC32 entries

jaikiran:
The commit here introduces a workaround to fix the issue noted in https://github.com/quarkusio/quarkus/issues/4782.

The root cause of the issue resides in recent versions of Java itself and needs to come in as a fix for https://bugs.openjdk.java.net/browse/JDK-8232879. Until that happens and the fix is available in (all) prominent recent versions of Java, the workaround in this commit should allow users to user Quarkus with recent Java versions on platforms like google cloud (which is where the user seems to have run into issues)

### DIFF
--- a/core/deployment/src/main/java/io/quarkus/deployment/pkg/steps/JarResultBuildStep.java
+++ b/core/deployment/src/main/java/io/quarkus/deployment/pkg/steps/JarResultBuildStep.java
@@ -1,5 +1,7 @@
 package io.quarkus.deployment.pkg.steps;
 
+import static io.quarkus.bootstrap.util.ZipUtils.wrapForJDK8232879;
+
 import java.io.ByteArrayOutputStream;
 import java.io.FileOutputStream;
 import java.io.IOException;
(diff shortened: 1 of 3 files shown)


gsmet:
@jaikiran can you confirm me that you tested to generate a runner jar with this patch and that running it was OK? I'm not sure we have tests really executing the runner jar.

jaikiran:
Hello @gsmet, Apart from that automated test I added, I manually ran various different versions of the tests. I generated the runner jar against different java versions with and without this fix and ran `jar -xf` against it to reproduce and verify the fix.
Also ran `mvn quarkus:dev` which booted and served fine.
Finally, ran:
```
java -jar target/code-with-quarkus-1.0.0-SNAPSHOT-runner.jar 
2019-10-23 20:27:07,609 INFO  [io.quarkus] (main) code-with-quarkus 1.0.0-SNAPSHOT (running on Quarkus 999-SNAPSHOT) started in 5.720s. Listening on: http://0.0.0.0:8080
2019-10-23 20:27:07,633 INFO  [io.quarkus] (main) Profile prod activated. 
2019-10-23 20:27:07,634 INFO  [io.quarkus] (main) Installed features: [cdi, narayana-jta, resteasy]
```
and then issued a `curl` against a rest resource which too went fine. All this against ```999-SNAPSHOT``` which includes this fix.

jaikiran:
> I'm not sure we have tests really executing the runner jar.

There's a `JarRunnerIT` which actually tests the runner jar, so as long as that passes (against different JDKs), I think we can be confident of this change.

gsmet (approved):

gsmet:
Yes, perfect, just wanted to be sure we didn't mess with the jars.

Thanks for the fix, let's merge it now that CI is green.

PR #4806, opened 2019-10-23
[#4765] fix node name to be harmonized with xa node name

### LINKED ISSUE #4765: narayana-jta extension: harmonise the node name and the xa node name to a single value
**Description**
The Narayana does not recommend to diverge the node identifier and the xa node names. This setup should be used only in rare specific cases and it should not be explicitly available as a Microprofile config property
https://github.com/quarkusio/quarkus/blob/0.25.0/extensions/narayana-jta/runtime/src/main/java/io/quarkus/narayana/jta/runtime/TransactionManagerConfiguration.java#L18

In short: the `nodeName` and the `xaNodeName` should not diverge and only `nodeName` should be configurable.

**The Narayana procesing background**

The 'nodeName' is used by transaction manager to bake the identifier into the Xid when transaction is in-doubt in memory. The transaction object can be then persisted to the log store during prepare phase. The 'xaNodeName' is a list which is used by recovery manager. It defines a set of identifiers which are considered as owned by the recovery manager. When the orphan detection runs then this set determines if the recovery manager is permitted to rollback the transaction.

**Implementation ideas**

* remove possibility to configure the `xaNodeName` and set the `xaNodeName` to the same value as the `nodeName` is set
* document what is the `nodeName` good for and emphasize that it's important to be unique amognst all the quarkus deployments

ochaloup:
fixes #4765

Narayana node name needs to be correctly set and it should be documented what it's for in the Quarkus documentation.

### DIFF
--- a/extensions/narayana-jta/runtime/src/main/java/io/quarkus/narayana/jta/runtime/NarayanaJtaRecorder.java
+++ b/extensions/narayana-jta/runtime/src/main/java/io/quarkus/narayana/jta/runtime/NarayanaJtaRecorder.java
@@ -1,6 +1,7 @@
 package io.quarkus.narayana.jta.runtime;
 
 import java.lang.reflect.Field;
+import java.util.Collections;
 import java.util.Properties;
 
 import org.jboss.logging.Logger;
@@ -8,6 +9,7 @@ import org.jboss.logging.Logger;
 import com.arjuna.ats.arjuna.common.CoreEnvironmentBeanException;
 import com.arjuna.ats.arjuna.common.arjPropertyManager;
 import com.arjuna.ats.arjuna.coordinator.TxControl;
+import com.arjuna.ats.jta.common.jtaPropertyManager;
 import com.arjuna.common.util.propertyservice.PropertiesFactory;
 
 import io.quarkus.runtime.annotations.Recorder;
@@ -23,7 +25,8 @@ public class NarayanaJtaRecorder {
 
         try {
             arjPropertyManager.getCoreEnvironmentBean().setNodeIdentifier(transactions.nodeName);
-            TxControl.setXANodeName(transactions.xaNodeName.orElse(transactions.nodeName));
+            jtaPropertyManager.getJTAEnvironmentBean().setXaRecoveryNodes(Collections.singletonList(transactions.nodeName));
+            TxControl.setXANodeName(transactions.nodeName);
         } catch (CoreEnvironmentBeanException e) {
             e.printStackTrace();
         }
@@ -46,6 +49,7 @@ public class NarayanaJtaRecorder {
 
     public void setDefaultTimeout(TransactionManagerConfiguration transactions) {
         transactions.defaultTransactionTimeout.ifPresent(defaultTimeout -> {
+            arjPropertyManager.getCoordinatorEnvironmentBean().setDefaultTimeout((int) defaultTimeout.getSeconds());
             TxControl.setDefaultTimeout((int) defaultTimeout.getSeconds());
         });
     }
--- a/docs/src/main/asciidoc/transaction.adoc
+++ b/docs/src/main/asciidoc/transaction.adoc
@@ -186,6 +186,22 @@ include::duration-format-note.adoc[]
 
 The default value is 60 seconds.
 
+== Configuring transaction node name identifier
+
+Narayana, as the underlaying transaction manager, has an concept of unique node identifier.
+This is important if you consider to use XA transactions with involve multiple resources.
+
+The node name identifier plays a crucial part in the identification of a transaction.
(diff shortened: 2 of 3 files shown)


ochaloup:
@tomjenkinson would you be so kind and check this PR?

extensions/narayana-jta/runtime/src/main/java/io/quarkus/narayana/jta/runtime/NarayanaJtaRecorder.java
```diff
@@ -23,7 +25,8 @@ public void setNodeName(final TransactionManagerConfiguration transactions) {
 
         try {
             arjPropertyManager.getCoreEnvironmentBean().setNodeIdentifier(transactions.nodeName);
-            TxControl.setXANodeName(transactions.xaNodeName.orElse(transactions.nodeName));
+            jtaPropertyManager.getJTAEnvironmentBean().setXaRecoveryNodes(Collections.singletonList(transactions.nodeName));
+            TxControl.setXANodeName(transactions.nodeName);
```
tomjenkinson: Do we genuinely need the TxControl setting here? I would imagine it can be controlled through the arjPropertyManager directly?

extensions/narayana-jta/runtime/src/main/java/io/quarkus/narayana/jta/runtime/NarayanaJtaRecorder.java
```diff
@@ -46,6 +49,7 @@ public void setDefaultProperties(Properties properties) {
 
     public void setDefaultTimeout(TransactionManagerConfiguration transactions) {
         transactions.defaultTransactionTimeout.ifPresent(defaultTimeout -> {
+            arjPropertyManager.getCoordinatorEnvironmentBean().setDefaultTimeout((int) defaultTimeout.getSeconds());
```
tomjenkinson: +1 - could the TxControl statement immediately following be removed too do you think?

extensions/narayana-jta/runtime/src/main/java/io/quarkus/narayana/jta/runtime/NarayanaJtaRecorder.java
ochaloup: this is the problematic issue of the whole narayana configuration. Once the `TxControl` class is first  initialized (and that's decision of the JVM, not possible to be influence. normally the `TxControl` is initialized as class when accessed first time) then any change done in the `arjPropertyManager` has no effect for the further use.
As we can't be sure what setter is started first or where the `TxControl` data is needed for the first time during Quarkus startup I consider this as the safe way (or maybe even necessary).

A bit related point is that I created a jira about rethinking the configuration of the whole Narayana. That came from discussion with @mmusgrov, here https://issues.jboss.org/browse/JBTM-3215

tomjenkinson: What we do have to be certain of is there are no transactions that could be created before this is called and indeed the recovery manager has not started running before it. Are you sure that is the case?

tomjenkinson: I assume that this code must run before the transaction service is created? In which case I would expect no transactions could be created before this runs? Or is that not the case in Quarkus? IOW could there be a transaction created with the default (non-unique) node ID and that could therefore lead to data integrity issue?

/cc @mmusgrov

ochaloup: @tomjenkinson the need of use `TxControl.setXANodeName` is not caused by time when the transaction service is created. The need is caused by the fact that we can't be sure that nobody else calls the `TxControl` `setter`. When it happens then all data are initialized. When this method is called after the `setter` was used before then changes in the `bean` have no effect.

In summary, by me the `TxControl` call is needed here.

tomjenkinson: Thanks for confirming. I suppose you are less concerned about the setter being called (which as you say it is [public](https://github.com/jbosstm/narayana/blob/master/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/TxControl.java#L168)) but rather when the class is loaded it will read the value as set in the arjPropertyManager: https://github.com/jbosstm/narayana/blob/master/ArjunaCore/arjuna/classes/com/arjuna/ats/arjuna/coordinator/TxControl.java#L238

In that case, I understand the point - thanks!

tomjenkinson (approved):

Sanne (approved):
Great, thanks!

PR #4807, opened 2019-10-23
Fix broken gradle buildNative

geoand:
The task needs to be executed after `assemble`.

When `build` was being used that was the case which is why that one was working.

### DIFF
--- a/devtools/gradle/src/main/java/io/quarkus/gradle/QuarkusPlugin.java
+++ b/devtools/gradle/src/main/java/io/quarkus/gradle/QuarkusPlugin.java
@@ -60,11 +60,13 @@ public class QuarkusPlugin implements Plugin<Project> {
                     quarkusBuild.dependsOn(classesTask);
                 });
 
-        Task buildNative = tasks.create("buildNative", QuarkusNative.class).dependsOn(quarkusBuild);
+        Task buildNative = tasks.create("buildNative", QuarkusNative.class);
 
         // set up the source set for the testNative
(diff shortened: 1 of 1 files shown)


gastaldi (approved):
LGTM

ia3andy (approved):
LGTM

geoand:
Need to look at the CI failure

devtools/gradle/src/main/java/io/quarkus/gradle/QuarkusPlugin.java
```diff
@@ -60,11 +60,13 @@ private void registerTasks(Project project) {
                     quarkusBuild.dependsOn(classesTask);
                 });
 
-        Task buildNative = tasks.create("buildNative", QuarkusNative.class).dependsOn(quarkusBuild);
+        Task buildNative = tasks.create("buildNative", QuarkusNative.class);
```
gsmet: Stupid question but why don't you create the task in the if below?

geoand: I think one of the Gradle tests would fail. Not 100% certain since I don't have the code in front of me, but I think that was the reason.

gsmet: OK, let's not waste more time on this then :).

geoand: Yeah the existing Gradle tests are pretty useless I think. Once @gastaldi finishes with the proper integration tests, we should maybe remove them.

stuartwdouglas (approved):

PR #4811, opened 2019-10-23
Assorted OIDC changes

gsmet:
Let's see what CI has to say and I added a few comments here and there to be sure I don't do anything wrong.

@pedroigor @sberyozkin @michalszynkiewicz can you check my comments inline and answer my questions? Thanks!

### DIFF
--- a/docs/src/main/asciidoc/oidc-guide.adoc
+++ b/docs/src/main/asciidoc/oidc-guide.adoc
@@ -219,14 +219,15 @@ The application is using bearer token authorization and the first
 thing to do is obtain an access token from the Keycloak Server in
 order to access the application resources:
 
-```bash
+[source,bash]
+----
 export access_token=$(\
     curl -X POST http://localhost:8180/auth/realms/quarkus/protocol/openid-connect/token \
     --user backend-service:secret \
     -H 'content-type: application/x-www-form-urlencoded' \
     -d 'username=alice&password=alice&grant_type=password' | jq --raw-output '.access_token' \
  )
-```
+----
 
 The example above obtains an access token for user `alice`.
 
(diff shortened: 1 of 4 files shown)


extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/OidcConfig.java
```diff
@@ -7,7 +7,7 @@
 import io.quarkus.runtime.annotations.ConfigPhase;
 import io.quarkus.runtime.annotations.ConfigRoot;
 
-@ConfigRoot(phase = ConfigPhase.BUILD_AND_RUN_TIME_FIXED)
+@ConfigRoot(phase = ConfigPhase.RUN_TIME)
```
gsmet: Why is this configuration build time? It's only used at the runtime init phase AFAICS and it's very useful to be able to override it at runtime.

extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/VertxOAuth2AuthenticationMechanism.java
```diff
@@ -24,16 +24,6 @@
...
-    private volatile OidcConfig config;
-
-    public OidcConfig getConfig() {
-        return config;
-    }
-
-    public VertxOAuth2AuthenticationMechanism setConfig(OidcConfig config) {
-        this.config = config;
-        return this;
-    }
```
gsmet: I don't think we should expose this and AFAICS it's not used. @michalszynkiewicz it's you who introduced that, can you confirm it's not used?

extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/VertxOAuth2AuthenticationMechanism.java
```diff
@@ -1,4 +1,4 @@
-package io.quarkus.oidc;
+package io.quarkus.oidc.runtime;
```
gsmet: Is this class supposed to be exposed to the users? It looks like an internal one but I would like to check.

extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/VertxOAuth2IdentityProvider.java
```diff
@@ -8,6 +8,7 @@
 import org.jose4j.jwt.JwtClaims;
 import org.jose4j.jwt.consumer.InvalidJwtException;
 
+import io.quarkus.oidc.VertxJwtCallerPrincipal;
```
gsmet: I'm not convinced by the name of this class. It's exposed to the user AFAICS, could we find a better name? I don't see why we have `Vertx` in it typically. It looks historical to me. `OidcJwtCallerPrincipal` maybe?

I will do the changes. Just asking for advice.

extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/VertxOAuth2AuthenticationMechanism.java
michalszynkiewicz: Neither my memory nor git blame thinks I introduced that but I don't see `getConfig` used anywhere and the field is not used either.

michalszynkiewicz:
I'm definitely no expert at this area but the changes look good to me

extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/VertxOAuth2AuthenticationMechanism.java
gsmet: Oh sorry, my Eclipse history made me think you added this when working on the security interceptors :).

Yeah, I think it's safe to remove it. Will wait for @stuartwdouglas .

extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/OidcConfig.java
gsmet: BTW, I looked at this code because I had a user asking how to override the realm at runtime so it's not a rhetorical question.

sberyozkin:
@gsmet Hi Guillaume, it is all looks fine. 
FYI, Pedro has started renaming things at https://github.com/pedroigor/quarkus/tree/issue-4480/extensions/oidc. Might have problems resolving the conflicts but not sure :-). But that PR unlikely to make it to 0.27.0 I guess, I've also branched from  Pedro's branch to have a few minor fixes applied :-).
If we want to have the files renamed before 0.27.0 then lets do it, otherwise may be we can wait for a week for Pedro's PR to go in ?

sberyozkin:
Though restricting the visibility by moving into `oidc.runtime` may indeed be better done earlier...

gsmet:
OK, looks like I wasted my time, apart for the doc fix and the move to runtime config.

@pedroigor any chance you could submit a PR with your changes. They are 12 days old AFAICS and they would be better in the Quarkus tree than in your own repo :).

gsmet:
@sberyozkin I removed the renaming part as @pedroigor 's work should rather be included. Can you check with him if he can rebase once this is in and submit a PR?

Thanks!

sberyozkin:
Hi @gsmet. We've talked with @pedroigor a bit earlier, the PR needs a bit more work (as the new flow proposed has to be distinguished from the default one where the user is not redirected and few other minor tweaks - I've actually started working on it on one of branches). We'll try to complete with Pedro next once the policy-enforcer PR is closed

sberyozkin (approved):

PR #4814, opened 2019-10-23
RESTEasy standalone - handle @ApplicationPath value correctly

### LINKED ISSUE #4787: RESTEasy standalone - @ApplicationPath not honored if it does not start with a slash
Moreover, if it starts with a slash the resource is served from both the root and the app path. 

For example:
```java
@ApplicationPath("/foo")
public class MyApp extends Application {
}
@Path("/greeting")
    @GET
    public String hello() { return "hello"; }
}
```
HTTP requests to `/greeting` and `/foo/greeting` are both handled by `GreetingResource`.

mkouba:
- application path value does not have to start with a slash
- resource should not be served from both the root and the app path
- resolves #4787

IMPL NOTE: We register the delegating handler in the default route (matches `HttpBuildTimeConfig.path`) for a JAX-RS root resource path of a default value `/`. However, for any other value we do register the handler in a special route that matches the root resource path. In fact, we register two routes that share the same handler. For example, for the JAX-RS root resource path `/foo` we register two routes, first matches `/foo` (exact match) and the second one matches `/foo/*`. And the reason is that otherwise the resources would be served from both the root and the app path - see the issue description for an example.

### DIFF
--- a/extensions/resteasy/deployment/src/main/java/io/quarkus/resteasy/deployment/ResteasyStandaloneBuildStep.java
+++ b/extensions/resteasy/deployment/src/main/java/io/quarkus/resteasy/deployment/ResteasyStandaloneBuildStep.java
@@ -31,14 +31,28 @@ import io.quarkus.resteasy.server.common.deployment.ResteasyDeploymentBuildItem;
 import io.quarkus.vertx.core.deployment.InternalWebVertxBuildItem;
 import io.quarkus.vertx.http.deployment.DefaultRouteBuildItem;
 import io.quarkus.vertx.http.deployment.RequireVirtualHttpBuildItem;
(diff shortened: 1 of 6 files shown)


patriot1burke (dismissed):
Should write a regression test for this.  I don't see one.   I imagine we might want to rework route processing int he future and there's no test for this case.

stuartwdouglas:
LGTM except the lack of tests

mkouba:
Good point with the tests. I'll add some.

mkouba:
Tests added...

mkouba:
@patriot1burke Pls could you re-review this PR again?

stuartwdouglas (approved):

gsmet (approved):
We have a proper test now, let's get it in the tree early.

PR #4816, opened 2019-10-23
Rename Postgres to PostgreSQL

rsvoboda:
Rename Postgres to PostgreSQL

https://vertx.io/docs/vertx-pg-client/js/ is referring to Reactive PostgreSQL Client, not Reactive Postgres Client.


gsmet:
Ah ah, thanks for that. I wanted to do that for a while but thought people might discover my OCD :).

gsmet (approved):

PR #4818, opened 2019-10-23
Fix elytron properties help

gsmet:
No rocket science but let's get these properties right.


gwenneg:
I know it's not part of your changes, but do we really need all those getters and setters? :) If yes, shouldn't the `plainText` field also have its getter and setter (and be part of the `toString` method)?

geoand (approved):

gsmet:
> I know it's not part of your changes, but do we really need all those getters and setters? :) If yes, shouldn't the plainText field also have its getter and setter (and be part of the toString method)?

Yeah, let's try to clean this up.

gsmet:
@gwenneg done.

Issue #3937, opened 2019-09-09
Support the configuration property indirection with custom ConfigSource


sberyozkin:
**Description**
It is possible to use the property indirection when all the properties are set in `application.properties`:
```
quarkus.datasource.password=${password}
password=bar
```
but not when a resolution of the `password` property is supported by a custom `ConfigSource`, so the following does not work:

```
quarkus.datasource.password=${password}
```
Having it supported would benefit Quarkus immensely, especially from the security point of view: for example, all the extensions requiring secret or indeed even public key materials, could all use some kind of Vault such as the HashiCorp one to fetch the secret values.   

**Implementation ideas**
Try to update the way the property indirection resolution works - for example, in the above case, if no `password` property is available within `application.properties` then check the custom config source

kenfinnigan:
This would likely require updates to https://github.com/smallrye/smallrye-config

sberyozkin:
Hi @kenfinnigan, thanks, will have a look

dmlloyd:
This happens because of how we wrap config sources.  But once property expansion hits the MP-config spec (and subsequently SmallRye Config), our custom property expansion will be removed and it should all begin to work consistently.

vsevel:
@dmlloyd excellent - it opens up numerous opportunities

PR #4821, opened 2019-10-24
Add form based authentication

### LINKED ISSUE #4348: Implement Vert.x based FORM auth in the new security layer
This allows for standard form based auth with IdentityManager implementation that support username/password auth.

State should be stored in a JWT token, however time constraints may mean that we may have to look into a simpler approach (DB store, HTTP session). 

We should also provide an 'automatic' login page for a quick getting started experience.

stuartwdouglas:
Fixes #4348

### DIFF
--- a/docs/src/main/asciidoc/security-guide.adoc
+++ b/docs/src/main/asciidoc/security-guide.adoc
@@ -99,6 +99,34 @@ very much a work in progress, so this list will be expanded over the coming week
 
 Please see the linked documents above for details on how to setup the various extensions.
 
+### Authenticating Via HTTP
+
+Quarkus has two built in authentication mechanisms for HTTP based FORM and BASIC auth. This mechanism is pluggable
+however so extensions can add additional mechanisms (most notably OpenID Connect for Keycloak based auth).
+
+#### Basic Authentication
+
+To enable basic authentication set `quarkus.http.auth.basic=true`. You must also have at least one extension installed
+that provides a username/password based `IdentityProvider`, such as link:security-jdbc-guide.html[Elytron JDBC].
+
+#### Form Based Authentication
+
+Quarkus provides form based authentication that works in a similar manner to traditional Servlet form based auth. Unlike
+traditional form authentication the authenticated user is not stored in a HTTP session, as Quarkus does not provide
+clustered HTTP session support. Instead the authentication information is stored in an encrypted cookie, which can
+be read by all members of the cluster (provided they all share the same encryption key).
+
+The encryption key can be set using the `quarkus.http.auth.session.encryption-key` property, and it must be at least 16 characters
+long. This key is hashed using SHA-256 and the resulting digest is used as a key for AES-256 encryption of the cookie
+value. This cookie contains a expiry time as part of the encrypted value, so all nodes in the cluster must have their
+clocks synchronised. At one minute intervals a new cookie will be generated with an updated expiry time if the session
+is in use.
+
+The following properties can be used to configure form based auth:
+
+include::{generated-dir}/config/quarkus-http-auth-form.adoc[opts=optional, leveloffset=+1]
+
+
 ### Securing Web Endpoints
 
 Quarkus has an integrated plugable web security layer. If security is enabled all HTTP requests will have a permission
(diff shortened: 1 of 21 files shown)


docs/src/main/asciidoc/security-guide.adoc
```diff
@@ -99,6 +99,34 @@ very much a work in progress, so this list will be expanded over the coming week
...
+that provides a username/password based `IdentityProvider`, such as link:security-jdbc-guide.html[Elytron JDBC].
+
+#### Form Based Authentication
+
+Quarkus provides form based authentication that works in a similar manner to traditional Servlet form based auth. Unlike
+traditional form authentication the authenticated user is not stored in a HTTP session, as Quarkus does not provide
+clustered HTTP session support. Instead the authentication information is stored in an encrypted cookie, which can
+be read by all members of the cluster (provided they all share the same encryption key).
+
+The encryption key can be set using the `auth.session.encryption-key` property, and it must be at least 16 characters
```
gsmet: Very minor but should it be:

```suggestion
The encryption key can be set using the `quarkus.http.auth.session.encryption-key` property, and it must be at least 16 characters
```
?

gsmet (changes requested):
I added one minor comment.

And CI failed with:
```
[ERROR]   NoConfiguredRealmsTestCase.testSecureAccessFailure:32 1 expectation failed.
Expected status code <403> but was <401>.
```
which looks related to your changes.

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/security/PersistentLoginManager.java
```diff
@@ -0,0 +1,118 @@
...
+        if (restoreResult != null) {
+            if (!restoreResult.newCookieNeeded) {
+                return;
+            }
+        }
+        try {
+            Cipher cipher = Cipher.getInstance("AES/ECB/PKCS5Padding");
+
+            StringBuilder contents = new StringBuilder();
+            //TODO: do we need random padding?
```
pedroigor: Can you take a look at this change set https://github.com/pedroigor/quarkus/commit/fb0b13d36d1adae9da4398c4fda8c69a5baed20c ?

It is about using AES GCM so that we have another random value to encrypt blocks. I have also changed to 128 which I think is a enough and help to balance with performance. We can make this configurable too.

I'm not using association data as that would require us to keep some state. But would be nice in order to make sure the user presenting the cookie is in fact the one to which the cookie was set. But it does not compromise the security.

pedroigor (approved):

stuartwdouglas:
I have addressed all the comments

pedroigor (approved):

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/HttpConfiguration.java
```diff
@@ -85,6 +85,16 @@
...
 
+    /**
+     * The encryption key that is used to store persistent logins (e.g. for form auth). Logins are stored in a persistent
+     * cookie that is encrypted with AES-256 using a key derived from a SHA-256 hash of the key that is provided here.
+     *
+     * If no key is provided then an in-memory one will be generated, this will change on every restart though so it
+     * is not suitable for production environments. This must be more than 16 characters long for security reasons
+     */
+    @ConfigItem(name = "auth.session.encryption-key")
+    public String encryptionKey;
```
sberyozkin: @stuartwdouglas should an `auth.session.encryption-key-location` be added and recommended by default ? `encryptionKey` is a good start but I'm not sure it is good for the production either, unless we have #3937, which would be good, as the users would just type 
```
auth.session.encryption-key=${my.encryption.key.from.vault}
```

stuartwdouglas: You can just set the QUARKUS_HTTP_AUTH_SESSION_ENCRYPTION_KEY env var and it will override.

stuartwdouglas: Either way this can be done as a future enhancement if there is demand for it. FORM is supposed to be mostly for learning/small apps, for big apps keycloak is our recommended solution.

sberyozkin: @stuartwdouglas That is fine. I'll create an issue to follow up. Was not really my intention to slow down the PR - but if we let the users to for ex provide the user passwords with the `elytron-security-jdbc` then there should be a way for them not put the secret key for protecting the cookie data in the clear form. But as I said, with Vault and the config indirection it will already be possible; adding a simple check to load it from a (hopefully secure) file system will just complement it.

Issue #4619, opened 2019-10-16
Quarkus OIDC shouldn't fail, when Auth Server is not up 


UnvirtualHH:
**Describe the bug**
When I start a quarkus application with the oidc extension used, the startup will fail, when the auth server is not reachable. 

**Expected behavior**
Application should start normal, maybe with a warning and an automatic retry

**Actual behavior**
Application stops with following exception:
`java.util.concurrent.CompletionException: io.netty.channel.AbstractChannel$AnnotatedConnectException: Connection refused: localhost/127.0.0.1:8180
	at java.base/java.util.concurrent.CompletableFuture.reportJoin(CompletableFuture.java:412)
	at java.base/java.util.concurrent.CompletableFuture.join(CompletableFuture.java:2044)
	at io.quarkus.oidc.VertxKeycloakRecorder.setup(VertxKeycloakRecorder.java:59)
	at io.quarkus.deployment.steps.VertxKeycloakBuildStep$setup28.deploy_0(VertxKeycloakBuildStep$setup28.zig:92)
	at io.quarkus.deployment.steps.VertxKeycloakBuildStep$setup28.deploy(VertxKeycloakBuildStep$setup28.zig:36)
	at io.quarkus.runner.ApplicationImpl1.doStart(ApplicationImpl1.zig:137)
	at io.quarkus.runtime.Application.start(Application.java:94)
	at io.quarkus.runner.RuntimeRunner.run(RuntimeRunner.java:135)
	at io.quarkus.dev.DevModeMain.doStart(DevModeMain.java:180)
	at io.quarkus.dev.DevModeMain.start(DevModeMain.java:94)
	at io.quarkus.dev.DevModeMain.main(DevModeMain.java:66)
Caused by: io.netty.channel.AbstractChannel$AnnotatedConnectException: Connection refused: localhost/127.0.0.1:8180
Caused by: java.net.ConnectException: Connection refused
	at java.base/sun.nio.ch.SocketChannelImpl.checkConnect(Native Method)
	at java.base/sun.nio.ch.SocketChannelImpl.finishConnect(SocketChannelImpl.java:779)
	at io.netty.channel.socket.nio.NioSocketChannel.doFinishConnect(NioSocketChannel.java:327)
	at io.netty.channel.nio.AbstractNioChannel$AbstractNioUnsafe.finishConnect(AbstractNioChannel.java:336)
[...]

PR #4828, opened 2019-10-24
All to disable the oidc extension

loicmathieu:
This is done for #4619 but not sure it fixes it.
Anyway, it needs to be done because we need to be able to switch authentication implementation for dev/prod as all other authentication mechanism permits.

I also add a featurebuilditem in order the extension to shows up on the log :)


gsmet (approved):
I think it makes sense. Thanks for that.

Let's wait for @sberyozkin as it might collide with other work.

sberyozkin:
@loicmathieu @gsmet Thanks, it looks good to go in :-)

gsmet:
Merged, thanks!

Issue #4836, opened 2019-10-24
vault-app tests fail to build native image


rsvoboda:
integration-tests/vault-app fails to build native image

```
uname -a
Darwin RS 18.7.0 Darwin Kernel Version 18.7.0: Tue Aug 20 16:57:14 PDT 2019; root:xnu-4903.271.2~2/RELEASE_X86_64 x86_64

mvn -version
Apache Maven 3.6.2 (40f52333136460af0dc0d7232c0dc0bcf0d9e117; 2019-08-27T17:06:16+02:00)
Maven home: /Users/rsvoboda/.sdkman/candidates/maven/current
Java version: 1.8.0_222, vendor: Oracle Corporation, runtime: /Users/rsvoboda/.sdkman/candidates/java/19.2.0.1-grl/jre
Default locale: en_CZ, platform encoding: UTF-8
OS name: "mac os x", version: "10.14.6", arch: "x86_64", family: "mac"
```
```
vault-app$ mvn clean package -Dnative -DskipTests -DskipITs
...
...
[INFO] --- quarkus-maven-plugin:0.23.0:native-image (native-image) @ quarkus-integration-test-vault-app ---
[INFO] [io.quarkus.deployment.QuarkusAugmentor] Beginning quarkus augmentation
[WARNING] [io.quarkus.config] Unrecognized configuration key "quarkus.build.skip" provided
[INFO] [org.jboss.threads] JBoss Threads version 3.0.0.Final
[INFO] [org.hibernate.jpa.boot.internal.PersistenceXmlParser] HHH000318: Could not find any META-INF/persistence.xml file in the classpath
[INFO] [org.hibernate.Version] HHH000412: Hibernate Core {5.4.7.Final}
[INFO] [io.quarkus.resteasy] Resteasy running without servlet container.
[INFO] [io.quarkus.resteasy] - Add quarkus-undertow to run Resteasy within a servlet container
[INFO] [io.quarkus.deployment.QuarkusAugmentor] Quarkus augmentation completed in 1650ms
[INFO] [io.quarkus.creator.phase.runnerjar.RunnerJarPhase] Building jar: /Users/rsvoboda/git/quarkus/integration-tests/vault-app/target/quarkus-integration-test-vault-app-999-SNAPSHOT-runner.jar
[INFO] [io.quarkus.creator.phase.nativeimage.NativeImagePhase] Running Quarkus native-image plugin on OpenJDK 64-Bit GraalVM CE 19.2.0.1
[...]

PR #4843, opened 2019-10-24
integration-tests - Use the right version of quarkus-maven-plugin in the native profile

jaikiran:
The commit here should fix the issue reported in https://github.com/quarkusio/quarkus/issues/4836. 

Ideally, I think, the parent pom.xml of the integration tests should probably dictate this in `pluginManagement` for the `native` profile too. But I couldn't get that working. Nor do I see any conclusive documentation that the `pluginManagement` for profiles is supported from the parent pom.xml. So at least for now, this goes directly in the `vault-app`'s `native` profile in pom.xml

### DIFF
--- a/integration-tests/amazon-lambda/pom.xml
+++ b/integration-tests/amazon-lambda/pom.xml
@@ -86,6 +86,7 @@
                     <plugin>
                         <groupId>io.quarkus</groupId>
                         <artifactId>quarkus-maven-plugin</artifactId>
+                        <version>${project.version}</version>
                         <executions>
                             <execution>
                                 <id>native-image</id>
--- a/integration-tests/artemis-core/pom.xml
+++ b/integration-tests/artemis-core/pom.xml
@@ -110,6 +110,7 @@
                     <plugin>
                         <groupId>io.quarkus</groupId>
                         <artifactId>quarkus-maven-plugin</artifactId>
+                        <version>${project.version}</version>
                         <executions>
                             <execution>
                                 <id>native-image</id>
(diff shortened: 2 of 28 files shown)


gsmet:
Could you do a quick sanity check that it’s not an issue with some other ITs? And that our Pom templates generating the projet poms are Ok? They are ftl files.

vsevel:
> Could you do a quick sanity check that it’s not an issue with some other ITs?

hello @gsmet I found quite a few. see https://github.com/quarkusio/quarkus/issues/4836#issuecomment-545975964
can't say if sometimes they are legitimate though

jaikiran:
> Could you do a quick sanity check that it’s not an issue with some other ITs?

I did some manual checks and found a bunch of them in the integration-tests that were missing the version. I've updated this PR to include them.

I was really hoping we could set this `native` profile in the parent pom.xml and then add `pluginManagement` there, just like with the default build/profile, but that isn't working for me. Maybe @aloubyansky would know if that's supported?

As for the ftl files which we use to generate some of these pom.xml files, I had a look at them in the `devtools/templates` and those rightly use the `${quarkus.version}` for the `quarkus-maven-plugin` in them. So this all looks like hand generated pom.xml errors.

geoand:
There will be a parent pom eventually AFAIK, but not soon.

jaikiran:
> There will be a parent pom eventually 

Hello @geoand, there in fact is a parent pom.xml in the `integration-tests` directory. It's just that the plugins within a specific profile of child projects don't seem to be managable through top level `pluginManagement` in the parent pom.xml nor through adding a specific `native` profile in the parent pom.xml

geoand:
Ah sorry, I thought you were referring to a general parent pom.

gsmet (approved):
It looks like a Maven issue to me but if we have at least one Maven version having this problem, we need a workaround anyway.

Thanks for working on this!

PR #4845, opened 2019-10-24
prepend the slash automatically if not present

### LINKED ISSUE #4835: Prepend a slash to a io.quarkus.vertx.web.Route.path if not present
A `Route` path must start with a `/`. So if a user declares `@Route(path = "hello")` Quarkus fails during RUNTIME_INIT. I think we should prepend the slash automatically if not present. For the record - JAX-RS `@Path` can have both values, ie. `@Path("hello")` and `@Path("/hello")`.

qrman:
resolve #4835 issue


stuartwdouglas (approved):

stuartwdouglas:
Merged, thanks

PR #4850, opened 2019-10-24
Must be install dir, not GRAALVM_HOME

ebullient:


docs/src/main/asciidoc/building-native-image-guide.adoc
```diff
@@ -110,7 +110,7 @@ Use the following command to recursively delete the `com.apple.quarantine` exten
 
 [source,shell]
 -----
-xattr -r -d com.apple.quarantine ${GRAALVM_HOME}
+xattr -r -d com.apple.quarantine $HOME/Development/graalvm/
```
stuartwdouglas: What is this directory? Is this basically just your workspace dir?

gsmet: Maybe let's make it `${GRAALVM_INSTALL_DIRECTORY}`

ebullient: It is the directory of the install, which is referenced earlier in the document.

> Configure the runtime environment. Set GRAALVM_HOME environment variable to the GraalVM installation directory, for example:
> export GRAALVM_HOME=$HOME/Development/graalvm/
> On macOS, point the variable to the Home sub-directory:
> export GRAALVM_HOME=$HOME/Development/graalvm/Contents/Home/

If we want to add an explicit set of GRAALVM_INSTALL_DIR at this point, we could do that (and then reference it in this later section). But I reused the string in common with the install instructions.

gsmet (approved):

gsmet:
OK, let's merge, it's probably clear enough.

gsmet:
Thanks!

PR #4853, opened 2019-10-25
Template deprecation warnings suppression

rsvoboda:
Template deprecation warnings suppression.

Template class is deprecated, there is some logic to do the fallback to Recorder behavior.

### DIFF
--- a/core/deployment/src/main/java/io/quarkus/deployment/ExtensionLoader.java
+++ b/core/deployment/src/main/java/io/quarkus/deployment/ExtensionLoader.java
@@ -111,6 +111,7 @@ public final class ExtensionLoader {
 
     private static final String CONFIG_ROOTS_LIST = "META-INF/quarkus-config-roots.list";
 
+    @SuppressWarnings("deprecation")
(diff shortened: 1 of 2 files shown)


core/deployment/src/main/java/io/quarkus/deployment/ExtensionLoader.java
```diff
@@ -530,7 +531,7 @@ private static boolean isRecorder(AnnotatedElement element) {
             if (isRecorder) {
                 boolean recorderFound = false;
                 for (Class<?> p : method.getParameterTypes()) {
-                    if (p.isAnnotationPresent(Recorder.class) || p.isAnnotationPresent(Template.class)) {
```
kenfinnigan: How is this change related to suppressing deprecation warnings?

gsmet: It's related because it uses the `isRecorder` method above which has the suppress warning.

gsmet (approved):

gsmet:
Thanks!

Issue #3436, opened 2019-08-06
Investigate Spring API-only deps


n1hility:
We should probably think about whether its worth creating a spring API only dependency, vs the current real artifacts, that way we don't end up with issues caused by spring implementation classes conflicting with our code or inadvertently getting picked up by the native-image generation process. It also would help establish what APIs we do and don't support.

geoand:
@n1hility I was thinking the same exact thing yesterday

geoand:
@gytis will try and take a look at this while I am on PTO

gytis:
@geoand we could use Maven Shade Plugin to pick the interfaces that are needed for Quarkus and release them under our own GAV name.

geoand:
@gytis Interesting! I think Spring Framework uses gradle however - although yeah the same idea would apply I assume.
Would you like to start experimenting with this? Having this absolutely miminum dependencies would be a great benefit to reduce the footprint of Quarkus applications using the spring modules.

gytis:
I tried it briefly a couple of weeks ago and it works fine. The main issue is to cherry-pick all the required classes and interfaces and list them in the plugin configuration. I'll try it properly on Monday.

geoand:
Awesome, thanks!

gytis:
@geoand take a look at the possible option for spring-di. Here's what would be changed in Quarkus: https://github.com/gytis/quarkus/commit/0c5b2b1460be52bf504fada42413d8aaea6825ec. And here's the dependency pom: https://github.com/gytis/quarkus-spring-context-api/blob/master/pom.xml.

Of course the naming, repository location and ethical aspect (we're taking parts for Spring Framework and renaming it) should be discussed. Once we agree, I could take a stab at spring-web dependencies

geoand:
@gytis I like it!

@n1hility WDYT about the approach? How should we proceed?

geoand:
@gytis we will also need something similar for #4038. 

@n1hility any thoughts on how to proceed with this?

gytis:
@geoand yep, also for Spring Web. It's quite tedious and time consuming to find all the necessary interfaces and classes. That's why I only did Spring DI to see if this approach would work for everybody.

geoand:
Makes sense @gytis. Let's see what @n1hility things before proceeding further.

Also we will need spring data jpa as well

cmoulliard:
> APIs we do and don't support.

[...]

PR #4855, opened 2019-10-25
Spring update to 5.2.0, Spring Data update to 2.2.0

rsvoboda:
Update of Spring Framework bits

 - Spring update to 5.2.0
 - Spring Data update to 2.2.0


geoand:
Actually we probably don't want to do this until #3436 is sorted out

gsmet:
Let's close it then. It's not worth keeping it in the PR list. Spring will have new updates by then.

PR #4856, opened 2019-10-25
Platform descriptor should only be downloaded when targeting a different version

### LINKED ISSUE #4822: Using tooling, platform descriptor should only be downloaded when targeting a different version
**Describe the bug**
When using maven plugin to create a project it downloads the descriptor even if it's the current used version in the plugin:
```
[INFO] --- quarkus-maven-plugin:0.26.1:create (default-cli) @ standalone-pom ---
Downloading from central: https://repo.maven.apache.org/maven2/io/quarkus/quarkus-bom-descriptor-json/0.26.1/quarkus-bom-descriptor-json-0.26.1.json
```

**Expected behavior**
Download only if the version is different and maybe have a cache?

**Actual behavior**
Download every time

aloubyansky:
Fixes #4822 
Removed the legacy extensions.json and enahanced the platform json descriptor loader to load the default version from the classpath


maxandersen:
just noticed the title "Platform descriptor should only be downloaded when targeting a different version" ...what does that mean ?

aloubyansky:
Meaning it should perform resolution only when the version of the plugin is different than the Quarkus version.

maxandersen:
resolution of what exactly ?

maxandersen:
i.e. if I have two platforms with same quarkus version I would still expect a different result...

aloubyansky:
Sure. Non quarkus-bom platforms will be resolved from Maven.

maxandersen (approved):
Looks good now that the build sequence are ensured to have IT run after the bom is built! LGTM!

maxandersen:
had to revert this as it caused master to fail  in quarkus-test-extension-deployment :/

PR #4837, opened 2019-10-24
Updated info about JDK version, highlighted Maven, GraalVM and JDK version

rsvoboda:
Updated info about JDK version to be in sync with https://quarkus.io/get-started/ (1,2,3,4 header)

Highlighted Maven, GraalVM and JDK version so it's more visible on https://quarkus.io/guides/building-native-image-guide page

PR #4859, opened 2019-10-25
Updated info about JDK version

rsvoboda:
Updated info about JDK version to be in sync with https://quarkus.io/get-started/ (1,2,3,4 header)

Resubmit of https://github.com/quarkusio/quarkus/pull/4837 (@gsmet fyi)

### DIFF
--- a/docs/src/main/asciidoc/building-native-image-guide.adoc
+++ b/docs/src/main/asciidoc/building-native-image-guide.adoc
@@ -20,9 +20,9 @@ To complete this guide, you need:
 
 * less than 15 minutes
 * an IDE
-* JDK 1.8+ installed with `JAVA_HOME` configured appropriately
+* JDK 8 installed with `JAVA_HOME` configured appropriately
 * A xref:configuring-c-development[working C development environment]
-* GraalVM version {graalvm-version} installed and  xref:configuring-graalvm[configured appropriately]
+* GraalVM version {graalvm-version} installed and xref:configuring-graalvm[configured appropriately]
 * A working container runtime (Docker, podman)
 * The code of the application developed in the link:getting-started-guide.html[Getting Started Guide].
 
--- a/docs/src/main/asciidoc/getting-started-guide.adoc
+++ b/docs/src/main/asciidoc/getting-started-guide.adoc
@@ -33,7 +33,7 @@ To complete this guide, you need:
 
 * less than 15 minutes
 * an IDE
-* JDK 1.8+ installed with `JAVA_HOME` configured appropriately
+* JDK 8 or 11+ installed with `JAVA_HOME` configured appropriately
 * Apache Maven 3.5.3+
 
 [TIP]


docs/src/main/asciidoc/building-native-image-guide.adoc
```diff
@@ -20,9 +20,9 @@ To complete this guide, you need:
 
 * less than 15 minutes
 * an IDE
-* JDK 1.8+ installed with `JAVA_HOME` configured appropriately
+* `JDK 8 or 11+` installed with `JAVA_HOME` configured appropriately
```
kenfinnigan: I don't think this applies as there is no native image generation for JDK 11 yet

gsmet: Yes, agreed, this is a mistake, it should be JDK 8 for now.

gsmet (changes requested):

docs/src/main/asciidoc/building-native-image-guide.adoc
ebullient: Does it need to have backticks?

rsvoboda: Users can have JAVA_HOME pointing to JDK 11 ++ GRAALVM_HOME pointing to 19.2.0.1.

I understand this is native guide, but limiting to JDK 8 only can be confusing

kenfinnigan: Isn't it more confusing to think Graal is working with JDK 11 when it isn't?

rsvoboda: Another way from this is to mention just GraalVM here and assume line 27 also covers prerequisites from getting-started-guide

kenfinnigan: With you're getting started guide change you still end up in the same boat, that a developer could be using JDK 11 features.

I think it's safer to explicitly say that native image only applies to JDK 8

docs/src/main/asciidoc/building-native-image-guide.adoc
```diff
@@ -20,9 +20,9 @@ To complete this guide, you need:
...
 * A xref:configuring-c-development[working C development environment]
-* GraalVM version {graalvm-version} installed and  xref:configuring-graalvm[configured appropriately]
+* `GraalVM` version `{graalvm-version}` installed and xref:configuring-graalvm[configured appropriately]
```
ebullient: Why backticks around GraalVM? It's a name/noun, not a code snippet?

rsvoboda: Mentioned in PR description

> Highlighted GraalVM and JDK version so it's more visible on https://quarkus.io/guides/building-native-image-guide page

ebullient: Version, yes! Name, why? I know, retentive.. but if you want emphasis, making the name *bold* is more appropriate than backticks. :-P

rsvoboda: Ok, will use bold. I originally picked backticks because`<code>...</code>` in the final html stands out a bit more.

rsvoboda:
PR updated based on the comments, in the end skipping version highlight

cescoffier (approved):

gsmet (approved):

gsmet:
Merged, thanks!

Issue #4320, opened 2019-10-02
Generation of extension.json should allow for overrides


maxandersen:
When a platform generates json it should be possible to override/add metadata:

- Categories
- Excluding extensions
- redefine its labels
- Technical vs exposed extensions
 
current idea:
have `<extension-name>.properties` and any property defined in here overrides what comes from the extensions own `quarkus-extension.properties`

maxandersen:
Just realized this notion of overrides has a challenge - might not be a problem; but something to be aware of. The manifest metada in quarkus-extension.json I assume will be or already is (@aloubyansky ?) available at build and runtime as metadata.

These would always read the data from the manifest and not from the platform - thus the overrides in the platform will be purely cosmetic for when using tools.

aloubyansky:
It will be available, yes. It also depends what kind of overrides we will be doing. And what it would be consumed for.

ia3andy:
We discussed it with @maxandersen and thought that using K8S style labels would be nice:
```
Name: RESTEasy
Description: REST framework implementing JAX-RS and more
Labels:
  ShortName: jax-rs
  Keywords: resteasy, jaxrs, web, rest
  Category: web
  Guide: https://quarkus.io/guides/rest-json-guide
  Status: stable
```

Here are the different labels would be supported (other would be ignored by tooling or could be specific to a sort of tooling):
- "ShortName" is an abbreviation of the name use for search purpose
- "Keywords" is what we previously referred as "labels" used for search purpose
- "Category" or "Categories" can be used to define the cat ids
- "Hidden" is default to false and can be used to hide an extension from tooling
- "Doc", "Documentation" or "Guide" is a link to the documentation previously referred as "guide"
- "Status" is the status of the extension "Draft", "Experimental", "Beta", "Stable" (default to Draft which is not visible), is always overridden by the platform layer.

emmanuelbernard:
I can live with that as long as some labels will be constrained and others are more free.

I'd likely go for lower case though for keys
Remember that one extension could have several guides
Why is status always overridden by the platform layer?

ia3andy:
> I can live with that as long as some labels will be constrained and others are more free.
Sure

> I'd likely go for lower case though for keys
I don't mind either. I just find it nice K8S way, but that must be my frontend side talking

> Why is status always overridden by the platform layer?
Because we want full control over that no? if not then nevermind

emmanuelbernard:
[...]

PR #4861, opened 2019-10-25
Adds the possibility to override values while generating the extension.json as described by issue #4320

quintesse:
Issue #4320

### DIFF
--- a/devtools/maven/src/main/java/io/quarkus/maven/GenerateExtensionsJsonMojo.java
+++ b/devtools/maven/src/main/java/io/quarkus/maven/GenerateExtensionsJsonMojo.java
@@ -1,6 +1,7 @@
 package io.quarkus.maven;
 
 import java.io.File;
+import java.io.FileInputStream;
 import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
@@ -8,13 +9,17 @@ import java.nio.file.FileSystem;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 
 import javax.json.Json;
+import javax.json.JsonArray;
 import javax.json.JsonArrayBuilder;
 import javax.json.JsonObject;
 import javax.json.JsonObjectBuilder;
 import javax.json.JsonReader;
+import javax.json.JsonValue;
 import javax.json.JsonWriter;
 import javax.json.JsonWriterFactory;
 import javax.json.stream.JsonGenerator;
@@ -59,6 +64,9 @@ public class GenerateExtensionsJsonMojo extends AbstractMojo {
     @Parameter(property = "bomVersion", defaultValue = "${project.version}")
     private String bomVersion;
 
+    @Parameter(property = "overridesFile", defaultValue = "${project.basedir}/src/main/resources/extensions-overrides.json")
+    private File overridesFile;
+
     @Parameter(property = "outputFile", defaultValue = "${project.build.directory}/extensions.json")
     private File outputFile;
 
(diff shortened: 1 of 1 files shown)


devtools/maven/src/main/java/io/quarkus/maven/GenerateExtensionsJsonMojo.java
```diff
@@ -59,6 +64,9 @@
     @Parameter(property = "bomVersion", defaultValue = "${project.version}")
     private String bomVersion;
 
+    @Parameter(property = "overridesFile", defaultValue = "${project.basedir}/src/main/resources/extensions-overrides.json")
```
maxandersen: this seems fishy  - users can ovveride the name of resources folder - isn't there a more maven agnostic way to look these up ?

aloubyansky: `${project.build.outputDirectory}` the resources will end up being there, also possibly filtered.

devtools/maven/src/main/java/io/quarkus/maven/GenerateExtensionsJsonMojo.java
```diff
@@ -143,56 +186,90 @@ public void execute() throws MojoExecutionException, MojoFailureException {
...
-                extJsonBuilder.add(object);
+                return object;
             }
         } catch (IOException e) {
             throw new IOException("Failed to parse " + descriptor, e);
         }
     }
 
+    private String extensionId(JsonObject extObject) {
+        return extObject.getString("groupId", "") + "." + extObject.getString("artifactId");
```
maxandersen: this gives a NPE if you forget to put groupid/artifactid in extension entry. need to handle that better.

aloubyansky: Why not keep the traditional `:` as a separator? It'll work either way but it'l be more intuitive and easier to read in case we need to display those if it is `:`.

devtools/maven/src/main/java/io/quarkus/maven/GenerateExtensionsJsonMojo.java
quintesse: Using `${project.build.outputDirectory}` doesn't seem to work. There's no output directory basically (there's just `target`, no `target/classes`)

quintesse: Before I was using `${project.resources[0].resource}` but that doesn't work. The expression is probably too complex.

aloubyansky: Ah, that's right.

quintesse:
> One issue is though that using this does not result in categories being in the final document:

@maxandersen where is that defined? I don't see anything about a categories section at the toplevel of that document being discussed in the issue?

devtools/maven/src/main/java/io/quarkus/maven/GenerateExtensionsJsonMojo.java
quintesse: Done.

quintesse: Done.

devtools/maven/src/main/java/io/quarkus/maven/GenerateExtensionsJsonMojo.java
quintesse: Also, why is this fishy? All paths are basically overridable, right? I mean is there some way this could be abused?

maxandersen: ah this the defaultvalue...oh well then i don't care at this stage - maybe post 1.0 if ever ;)

maxandersen:
> @maxandersen where is that defined? I don't see anything about a categories section at the toplevel of that document being discussed in the issue?

because there aren't any (yet). in this case they are just added during this patching.

maxandersen:
I expected that just as I added a new "status" property inside extension I could add new "categoreies" property at the root.

quintesse:
@maxandersen well that's _very_ easy to do if you want that :-)

quintesse:
@maxandersen well actually it's not _that_ trivial, now that I think about it. So are you sure you want this? It's not the work (probably 15mins) but it's more about adding features that might need to be maintained etc

maxandersen:
We need *someway* to add categories. If not using this patch approach then we can do a separate file and bulk add it in.

maxandersen (approved):
lgtm

maxandersen:
rebased against master and now merged.

PR #4864, opened 2019-10-25
Added pointer to guide for reactive clients

FroMage:
@maxandersen this is still the right file to update, right?


maxandersen (approved):
Yes, this is the right file. 

If/when this changes i'll PR it in one go ;)

FroMage:
OK thanks.

PR #4875, opened 2019-10-25
CDI guide: add method removal documentation

loicmathieu:
Add to the CDI documentation the fact tha Arc also remove method that produces beans, not just beans, and that @Unremovable can be used on methods.

Also adds a line explaining how to debug removal issues

### DIFF
--- a/docs/src/main/asciidoc/cdi-reference.adoc
+++ b/docs/src/main/asciidoc/cdi-reference.adoc
@@ -470,12 +470,22 @@ An unused bean:
 * does not declare any producer which is eligible for injection to any injection point,
 * is not directly eligible for injection into any `javax.enterprise.inject.Instance` or `javax.inject.Provider` injection point
 
+This optimization applies to all forms of bean declarations: bean class, producer method, producer field.
+
 Users can instruct the container to not remove any of their specific beans (even if they satisfy all the rules specified above) by annotating them with `io.quarkus.arc.Unremovable`.
+This annotation can be placed on the types, producer methods, and producer fields.
  
 Furthermore, extensions can eliminate possible false positives by producing `UnremovableBeanBuildItem`.
 
 Finally, Quarkus provides a middle ground for the bean removal optimization where application beans are never removed whether or not they are unused,
 while the optimization proceeds normally for non application classes. To use this mode, set `quarkus.arc.remove-unused-beans` to `fwk` or `framework`.
+
+In order to see more information about which beans are being removed,
+you can enable additional logging via the following line in your `application.properties`.
+
+----
+quarkus.log.category."io.quarkus.arc.producer".level=DEBUG
+----
 
 [[default_beans]]
 == Default Beans
--- a/independent-projects/arc/runtime/src/main/java/io/quarkus/arc/Unremovable.java
+++ b/independent-projects/arc/runtime/src/main/java/io/quarkus/arc/Unremovable.java
@@ -7,7 +7,19 @@ import java.lang.annotation.Target;
 
 /**
  * Indicates that the bean marked with this annotation should never be removed by Arc
- * even if it has no known injection points
+ * even if it's considered unused.
+ *
+ * An unused bean:
+ * <ul>
+ * <li>is not a built-in bean or an interceptor</li>
+ * <li>is not eligible for injection to any injection point,</li>
+ * <li>is not excluded by any extension,</li>
+ * <li>does not have a name,</li>
+ * <li>does not declare an observer,</li>
+ * <li>does not declare any producer which is eligible for injection to any injection point,</li>
+ * <li>is not directly eligible for injection into any `javax.enterprise.inject.Instance` or `javax.inject.Provider` injection
+ * point</li>
+ * </ul>
  */
 @Retention(RetentionPolicy.RUNTIME)
 @Target({ ElementType.TYPE, ElementType.FIELD, ElementType.METHOD })


geoand:
Maybe it's just me but I feel that this shouldn't be needed because a bean producer is a bean itself so what the documentation mentions still applies.
On the other hand, I understand that it can be useful to make things crystal clear.

loicmathieu:
And this is needed, we had the case when working on the multiple mongoclient implementation with a Producer bean that is unremovable but not it's methods, and the methods was removed. We spent multiple hours searching what's happens then I discover by enabling the logs that the bean is not removed but it's methods are.

That's why I propose to add a sentence in the guide for this tricky case :)

geoand:
Fair enough :)

manovotn:
> a bean producer is a bean itself

That's right, producer method is equal declaration to a bean class.
And that's not something we came with either, that's from CDI spec.

Although we can of course make it clearer in our docs.

docs/src/main/asciidoc/cdi-reference.adoc
```diff
@@ -474,9 +474,18 @@ Users can instruct the container to not remove any of their specific beans (even
...
 Furthermore, extensions can eliminate possible false positives by producing `UnremovableBeanBuildItem`.
 
+The container will also remove bean producing methods (methods annotated with `@Produces`) from producer beans if the method produces an unused bean.
+You can also annotate bean producing methods with `io.quarkus.arc.Unremovable`.
+
 Finally, Quarkus provides a middle ground for the bean removal optimization where application beans are never removed whether or not they are unused,
 while the optimization proceeds normally for non application classes. To use this mode, set `quarkus.arc.remove-unused-beans` to `fwk` or `framework`.
 
+If you feel that one of your beans, or one of your methods are removed,
+you can check it by enabling Arc producer beans logging via the following line in your application.properties.
```
manovotn: Let's reword this to:
`In order to see more information about which beans are being removed, you can enable additional logging via the following line in your application.properties.`

docs/src/main/asciidoc/cdi-reference.adoc
```diff
@@ -474,9 +474,18 @@ Users can instruct the container to not remove any of their specific beans (even
  
 Furthermore, extensions can eliminate possible false positives by producing `UnremovableBeanBuildItem`.
 
+The container will also remove bean producing methods (methods annotated with `@Produces`) from producer beans if the method produces an unused bean.
+You can also annotate bean producing methods with `io.quarkus.arc.Unremovable`.
```
manovotn: I'd append this to the sentence above starting `Users can instruct...` in a following way:

```
Users can instruct the container to not remove any of their specific beans (even if they satisfy all the rules specified above) by annotating them with `io.quarkus.arc.Unremovable`.
This applies to all forms of bean declarations - bean class, producer method, producer field.
```

Would that be clear enough?

manovotn (commented):
Added some suggestions.

docs/src/main/asciidoc/cdi-reference.adoc
loicmathieu: done

docs/src/main/asciidoc/cdi-reference.adoc
loicmathieu: I rephrase it a little to be crystal clear by splitting it to two sentences:

- One explaining that the optimization (removing beans) apply to all sort of bean definition
- One explaining that the annotation can be put on all sort of bean definition

I think that it worth being explicit in case the developer don't know the spec very well (like me more used to Spring DI than CDI)

mkouba:
We should probably also clarify the `Unremovable`'s javadoc: "Indicates that the bean marked with this annotation should never be removed by Arc even if **it has no known injection points**". The bold part should be replaced with something like "it's considered unused" + add the description what "unused" means.

loicmathieu:
@mkouba instead of copy/pasting the long description of what an unused bean is I propose to link to the corresponding paragraph in the guide:

_Indicates that the bean marked with this annotation should never be removed by Arc
 even if it's considered unused, you can find the description of an unused bean in the CDI reference guide: https://quarkus.io/guides/cdi-reference#remove_unused_beans_

docs/src/main/asciidoc/cdi-reference.adoc
```diff
@@ -470,13 +470,23 @@ An unused bean:
 * does not declare any producer which is eligible for injection to any injection point,
 * is not directly eligible for injection into any `javax.enterprise.inject.Instance` or `javax.inject.Provider` injection point
 
+This optimization applies to all forms of bean declarations: bean class, producer method, producer field.
+
 Users can instruct the container to not remove any of their specific beans (even if they satisfy all the rules specified above) by annotating them with `io.quarkus.arc.Unremovable`.
+This annotation can be placed on the type declaration, on a producer method or on a producer field.
```
manovotn: The repetition of "on" in this sentence rings false in my ears. I am not a native speaker though; I could be wrong :)
How about we just say the following?
`This annotation can be placed on types, methods and fields.`

manovotn:
> @mkouba instead of copy/pasting the long description of what an unused bean is I propose to link to the corresponding paragraph in the guide:

Link is fine by me, although having it all in javadoc is more comfortable as you don't need to leave the IDE just to read the javadoc elsewhere.

docs/src/main/asciidoc/cdi-reference.adoc
loicmathieu: Yes but with the producer qualifier
```
This annotation can be placed on types, producer methods, and producer fields.
```

manovotn: Fair enough

loicmathieu: done

loicmathieu:
> Link is fine by me, although having it all in javadoc is more comfortable as you don't need to leave the IDE just to read the javadoc elsewhere.

Re-think about it and I must agree :) 
I copy/paste the description from the guide in the JavaDoc

manovotn (approved):

PR #4876, opened 2019-10-25
Kotlin docs - minor fix

oztimpower:
Rest-assured version is specified in the bom, thus remove.  Fix location of group and version for Gradle Groovy DSL.


geoand (approved):
Looks good, thanks a lot!

PR #4877, opened 2019-10-25
Docs fixes

geoand:


gsmet:
I don’t think that’s a good idea. Maybe a specific task to only generate the docs would be better?

geoand:
@gsmet ideally yes. But at the moment we don't know enough about Azure to know how we can detect doc only changes and then launch a specific job.

This has already bitten us so this is a fix that will at least ensure master doesn't break

docs/src/main/asciidoc/extension-authors-guide.adoc
```diff
@@ -1357,7 +1357,6 @@ The Jackson extension will then use the produced build item to register a module
 If you need more customization capabilities than registering a module,
 you can produce a CDI bean that implements `io.quarkus.jackson.ObjectMapperCustomizer` via an `AdditionalBeanBuildItem`.
 More info about customizing Jackson can be found on the JSON guide link:rest-json-guide.html#configuring-json-support[Configuring JSON support]
-----
```
machi1990: Ah you beat me to it. I was preparing a PR :-)

gsmet:
Why not just add a job that builds the doc only. 

It would be easy and shouldn’t have a big cost on standard runs.

geoand:
I don't see how that would work in Azure Pipelines. If you have an example, then that would be great.

gsmet:
Ah but the excludes are global

Frankly it’s the first time it happens and we will probably have a lot of doc fixes, so I would just live with it for now to avoid overloading CI in these critical times.

geoand:
I'm fine with whatever, but let's have a decision in chat please soon. Master is currently broken

geoand:
I removed the second commit. Please review and merge. We can decide on whether or not we want CI to run for the docs later

gsmet (approved):

gsmet:
Merged, thanks.

geoand:
Thanks for the review!

PR #4879, opened 2019-10-25
Add RestClientResponseExceptionMapperTest test

gastaldi:


extensions/rest-client/deployment/src/test/java/io/quarkus/restclient/RestClientResponseExceptionMapperTest.java
```diff
@@ -0,0 +1,62 @@
...
+    @Inject
+    @RestClient
+    Client client;
+
+    @Test
+    public void testExceptionMapperIsCalled() {
+        assertThatExceptionOfType(IllegalArgumentException.class).isThrownBy(client::test);
+    }
+
+    @RegisterRestClient(baseUri = "http://localhost:8081")
```
geoand: Is there perhaps a chance to get the port from the environment? 
I vaguely recall that being done somewhere...

kenfinnigan:
What's the purpose of the test?

It's not related to a fix.

gastaldi:
It's not a fix, It's more about showing that the ResponseExceptionMapper is called correctly, but I don't mind closing it, I just created the test because I wanted to check whether ResponseExceptionMappers were registered correctly

extensions/rest-client/deployment/src/test/java/io/quarkus/restclient/RestClientResponseExceptionMapperTest.java
gastaldi: I couldn't find another way to do that.

kenfinnigan:
Is the situation covered by the TCK: https://github.com/eclipse/microprofile-rest-client/blob/master/tck/src/main/java/org/eclipse/microprofile/rest/client/tck/ExceptionMapperTest.java

If it isn't, then I'd recommend a TCK test instead

gastaldi:
I think the only difference is that it uses an injected @RestClient. I agree it should be part of the TCK. I'll close it for now

Issue #4830, opened 2019-10-24
infinispan-embedded - fail to build native image


rsvoboda:
integration-tests/infinispan-embedded fails to build native image

```
uname -a
Darwin RS 18.7.0 Darwin Kernel Version 18.7.0: Tue Aug 20 16:57:14 PDT 2019; root:xnu-4903.271.2~2/RELEASE_X86_64 x86_64

mvn -version
Apache Maven 3.6.2 (40f52333136460af0dc0d7232c0dc0bcf0d9e117; 2019-08-27T17:06:16+02:00)
Maven home: /Users/rsvoboda/.sdkman/candidates/maven/current
Java version: 1.8.0_222, vendor: Oracle Corporation, runtime: /Users/rsvoboda/.sdkman/candidates/java/19.2.0.1-grl/jre
Default locale: en_CZ, platform encoding: UTF-8
OS name: "mac os x", version: "10.14.6", arch: "x86_64", family: "mac"
```

```
infinispan-embedded$ mvn clean package -Dnative -DskipTests -DskipITs
...
...
[INFO] --- quarkus-maven-plugin:999-SNAPSHOT:native-image (native-image) @ quarkus-integration-test-infinispan-embedded ---
[INFO] [io.quarkus.creator.curator.Curator] provideOutcome depsOrigin=application, versionUpdate=none, versionUpdateNumber=micro
[INFO] [io.quarkus.deployment.QuarkusAugmentor] Beginning quarkus augmentation
[INFO] [org.jboss.threads] JBoss Threads version 3.0.0.Final
[INFO] [io.quarkus.resteasy] Resteasy running without servlet container.
[INFO] [io.quarkus.resteasy] - Add quarkus-undertow to run Resteasy within a servlet container
[INFO] [io.quarkus.deployment.pkg.steps.JarResultBuildStep] Building native image source jar: /Users/rsvoboda/git/quarkus/integration-tests/infinispan-embedded/target/quarkus-integration-test-infinispan-embedded-999-SNAPSHOT-native-image-source-jar/quarkus-integration-test-infinispan-embedded-999-SNAPSHOT-runner.jar
[...]

PR #4887, opened 2019-10-26
Run infinispan-embedded in native pipelines

rsvoboda:
Run infinispan-embedded in native pipelines

To prevent issues like https://github.com/quarkusio/quarkus/issues/4830

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -264,6 +264,7 @@ stages:
           modules:
             - infinispan-cache-jpa
             - infinispan-client
+            - infinispan-embedded
             - mongodb-client
             - mongodb-panache
             - neo4j


jaikiran:
This is a genuine issue in the `infinispan-embedded` integration-tests. It has this:

```
<configuration>
      ...
    <!-- We have some example stack files we want to use -->
    <additionalBuildArgs>
        <additionalBuildArg>-H:ResourceConfigurationFiles=${project.basedir}/src/main/resources/resources-config.json</additionalBuildArg>
    </additionalBuildArgs>
```
Given that the CI build runs `native-image` within `docker`, this path (preferably `${project.build.outputDirectory}` first needs to be exposed via `-v` or `--volume` through `containerRuntimeOptions`, before it can be used as `-H:ResourceConfigurationFiles`

rsvoboda:
`infinispan-embedded` is failing on macOS, don't have linux machine around to play with it :(

rsvoboda:
Looking at infinispan-embedded on Linux machine, transparent linking of resources-config.json in maven is problematic. Shouldn't be Quarkus helping with this stuff?

rsvoboda:
Trying `-H:IncludeResources` approach but resources are not included...

rsvoboda:
Closing, don't see elegant solution for this atm.

PR #4889, opened 2019-10-26
Decouple Spring Web from quarkus-undertow

geoand:
Spring Web can now use standalone RESTEasy when no Servlet capabilities are needed

### DIFF
--- a/integration-tests/spring-web/pom.xml
+++ b/integration-tests/spring-web/pom.xml
@@ -21,6 +21,10 @@
         </dependency>
         <dependency>
             <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-undertow</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
             <artifactId>quarkus-spring-di</artifactId>
         </dependency>
         <dependency>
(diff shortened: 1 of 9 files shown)


geoand:
This improves startup a little and of course is in line with what Quarkus already does.

integration-tests/spring-web/pom.xml
```diff
@@ -19,6 +19,10 @@
             <groupId>io.quarkus</groupId>
             <artifactId>quarkus-spring-web</artifactId>
         </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-undertow</artifactId>
+        </dependency>
```
gsmet: So my only comment here is that we have no idea if things work without undertow being there. Or do we?

geoand: We do because of 2 things:. 

1) I added one if the internal tests
2) I also tested the Spring Web quickstart

gsmet: Ah yes, the added test is in the deployment module. I missed that, sorry.

Looks good to me then.

gsmet (approved):

geoand:
Thanks for checking 😊

PR #4895, opened 2019-10-26
open config doc's type javadocs in new tab

machi1990:
A minor one to open the JavaDoc link of config Type in a new tab instead of the current tab.

/cc @gsmet @FroMage

### DIFF
--- a/core/processor/src/main/java/io/quarkus/annotation/processor/generate_doc/SummaryTableDocFormatter.java
+++ b/core/processor/src/main/java/io/quarkus/annotation/processor/generate_doc/SummaryTableDocFormatter.java
@@ -45,7 +45,7 @@ final class SummaryTableDocFormatter implements DocFormatter {
             typeContent = configDocKey.computeTypeSimpleName();
             final String javaDocLink = configDocKey.getJavaDocSiteLink();
             if (!javaDocLink.isEmpty()) {
-                typeContent = String.format("link:%s[%s]\n", javaDocLink, typeContent);
+                typeContent = String.format("link:%s[%s,window=\"_blank\"]\n", javaDocLink, typeContent);
             }
         }
         if (configDocKey.isList()) {


FroMage:
So, ten years ago I had an argument with @hilton where he convinced me that it was the user's responsability to decide if he wanted to open in a new window or not, and that it was bad UX to decide for him, because there's no UI to let you know this in advance (will it open normally, or in a new window), and if it opens in a new window, there's no UI in browsers to force it to open in the current window (whereas there is UI to force open in new windows if it's not meant ti).

Now, ten years on, and there are a lot of very famous websites which default to "open in new window" without any UI to tell you about it (GMail and many others), so perhaps it's become acceptable.

FroMage:
So, I'm not really sure. I could go either way, but I'd prefer if "new window" links would have at least some UI icon to set them apart.

FroMage:
I haven't looked at the resulting links, though, do they have special UI?

gsmet:
Yeah, I tend to lean towards letting the user decide. But I don't have any strong opinion either.

EXCEPT that I want to be sure that in modern browsers, it opens in a new tab, not in a new window. Because that is a bad behavior IMHO.

machi1990:
> EXCEPT that I want to be sure that in modern browsers, it opens in a new tab, not in a new window. Because that is a bad behavior IMHO.

Tested this in Chrome and Safari and the link is opened in new tab.

hilton:
> there are a lot of very famous websites which default to "open in new window" without any UI to tell you about it (GMail and many others), so perhaps it's become acceptable

Or perhaps dark UX (a.k.a. asshole design) is more common now.

FroMage:
> Or perhaps dark UX (a.k.a. asshole design) is more common now.

Yeah, it's very russian-rouletty

machi1990:
Closing. We can revisit when it is actually needed.

PR #4896, opened 2019-10-27
Restart app on pom.xml change

### LINKED ISSUE #4871: Detect pom changes when in quarkus:dev mode
Today, you need to restart quarkus:dev to take pom changes into action.
We have long discussed fixing that but no issue was capturing this.

CC @stuartwdouglas

stuartwdouglas:
Unlike other hot replacement this simply watches the pom.xml files,
and reloads the complete application on change. This means there is
a short period where the app in unavailible.

Fixes #4871

### DIFF
--- a/devtools/maven/src/main/java/io/quarkus/maven/DevMojo.java
+++ b/devtools/maven/src/main/java/io/quarkus/maven/DevMojo.java
@@ -17,6 +17,8 @@ import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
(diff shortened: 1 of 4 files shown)


devtools/maven/src/main/java/io/quarkus/maven/DevMojo.java
```diff
@@ -297,6 +299,147 @@ public void execute() throws MojoFailureException, MojoExecutionException {
...
 
+            DevModeRunner runner = new DevModeRunner(args);
+
+            runner.prepare();
+            runner.run();
+            long nextCheck = System.currentTimeMillis() + 100;
+            Map<Path, Long> pomFiles = readPomFileTimestamps(runner);
+            for (;;) {
+                //we never suspend after the first run
+                suspend = "n";
```
jaikiran: Hello @stuartwdouglas, is this meant to be passed along to the new `DevModeRunner`, so that it doesn't (start and) suspend in debug mode? If so, I don't see changing this value playing a role in it, since the `args` being passed to the new `DevModeRunner`  already have a bunch of JVM `-Xdebug` options setup in it.

stuartwdouglas: Good point, I will just move all this logic into prepare()

emmanuelbernard:
@stuartwdouglas I think it works well I tested it in a few combinations and never had unexpected behaviors. Except in one case:

- create a new RESTEasy project (from generator)
- add mvn quarkus:add-extensions -Dextensions="jdbc-postgresql,hibernate-orm-panache"
- add Todo class with @Entity on it
- refresh see error
- remove jdbc-postgresql,hibernate-orm-panache fromt he pom
- refresh and don't see error even though the Todo app has missing packages and annotation references that would not compile

We can survive without but that's a bit of an unexpected bahavior.


Also needs testing on Windows I think @maxandersen

emmanuelbernard (approved):
Stuart I approved it with two pending conditions:

- review my "non compilation" error comment and see if that's an easy fix
- I asked @maxandersen to test this PR on Windows, it's probably smart to make sure it works on W before pushing into the release. He will od it today asap after one platform issue is behind us

emmanuelbernard:
@stuartwdouglas ^

stuartwdouglas:
There is nothing I can do about that short of issuing a 'clean compile' invocation after pom changes. Even then I am not really sure how I would go about reporting the error.

I don't think this is a big deal as mostly we care about adding extensions rather than removing them.

maxandersen:
first thing - if you `clean compile` don't you risk loosing any post-compile generated artifacts in the target dir ?  or are they assumed not relevant for devmode ?

stuartwdouglas:
dev mode running happens straight after compile, so if there are any they will be out of date anyway. I don't think it is really a workable solution, and I don't think it is a major problem.

emmanuelbernard:
Yes I'm cool with that little quirk and see how it develops in practical usages

maxandersen:
for the record i tested the basic flow and worked on windows.

PR #4907, opened 2019-10-27
Upgrade to Hibernate Validator 6.1.0.Final, RESTEasy 4.4.0.CR1 and move to Jakarta artifacts

gsmet:
I moved most of the artifacts to Jakarta but not all as RESTEasy optional dependencies are heavily relying on two JBoss API spec artifacts for JAX-RS and JAXB (but not entirely consistently, I will report that). These 2 artifacts were updated to Jakarta.

Note that I added a lot of exclusions to be sure we don't have any duplicates: we had a few duplications present in the current code base.

I have another branch with a complete move to Jakarta but:
- it adds a lot more exclusions + explicit dependencies as JAX-RS and JAXB are injected in a lot of places
- every time you add an optional RESTEasy dependencies (typically the one for RX-Java2), you end up with duplicate artifacts (or requires exclusions + explicit dependencies)
So I thought this was the best compromise for now.

Draft status for now as we need to discuss this.

@FroMage could you have a look at https://github.com/quarkusio/quarkus/pull/4907/commits/e9af116176410384c00b13114e78d5529dbaa44e#diff-806baf513aab8570dfd629e9eaff1dadR136 , I'm not entirely sure where I should get the remote address from. Usually the header don't provide the really remote one so I was wondering exactly what you had in mind? Note that I might have missed something in Vert.x.

### DIFF
--- a/bom/runtime/pom.xml
+++ b/bom/runtime/pom.xml
@@ -52,15 +52,15 @@
         <jakarta.json.version>1.1.6</jakarta.json.version>
         <jakarta.json.bind-api.version>1.0.2</jakarta.json.bind-api.version>
         <jakarta.persistence-api.version>2.2.3</jakarta.persistence-api.version>
+        <jakarta.security.jacc-api.version>1.6.1</jakarta.security.jacc-api.version>
+        <jakarta.security.jaspi-api.version>1.1.3</jakarta.security.jaspi-api.version>
         <jakarta.servlet-api.version>4.0.3</jakarta.servlet-api.version>
         <jakarta.transaction-api.version>1.3.3</jakarta.transaction-api.version>
         <jakarta.validation-api.version>2.0.2</jakarta.validation-api.version>
-        <jakarta.xml.bind-api.version>2.3.2</jakarta.xml.bind-api.version>
+        <jakarta.websocket-api.version>1.1.2</jakarta.websocket-api.version>
         <jaxb-runtime.version>2.3.3-b01</jaxb-runtime.version>
-        <javax.security.jacc-api.version>1.0.2.Final</javax.security.jacc-api.version>
-        <javax.security.jaspi-api.version>1.0.2.Final</javax.security.jaspi-api.version>
-        <jboss-jaxrs-api_2.1_spec.version>1.0.2.Final</jboss-jaxrs-api_2.1_spec.version>
-        <javax.ws.rs-api.version>2.1.1</javax.ws.rs-api.version>
+        <jboss-jaxrs-api_2.1_spec.version>2.0.1.Final</jboss-jaxrs-api_2.1_spec.version>
+        <jboss-jaxb-api_2.3_spec.version>2.0.0.Final</jboss-jaxb-api_2.3_spec.version>
         <asm.version>7.1</asm.version>
         <commons-io.version>2.6</commons-io.version>
         <jboss-metadata-web.version>11.0.0.Final</jboss-metadata-web.version>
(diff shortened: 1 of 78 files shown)


FroMage:
@gsmet `request.remoteAddress().host()` is how I've implemented it in `resteasy-vertx` as well, so that's fine.

Now, if you're talking about https://tools.ietf.org/html/rfc7239#section-4 and the previous headers for this, we should indeed have an answer. But the same question is true of the host/proto parts as well, because they should be used by `UriInfo`.

My guess is that this should be done at a lower-level, but perhaps I'm wrong.

FroMage:
I could not find anything related in netty. @vietj says vertx-core doesn't handle it either. I couldn't find anything in quarkus either. Perhaps we're missing handling for these headers entirely?

Do you know @stuartwdouglas ?

stuartwdouglas:
Undertow can handle it (io.undertow.server.handlers.ForwardedHandler), but I don't think vert.x really has any low level way of handling it. There is also X-Forwarded-For which is more common than the RFC one.

FroMage:
Yeah, I'm talking about the older headers as well as the new one. So it appears we just don't handle it at all in Quarkus, which is frankly surprising to me because I've always used forward proxies to deploy my apps. This means that we can't trust `UriInfo` to generate absolute links (port/proto/host name will be wrong) and we can't trust the remote IP either (I've had plenty of deployments that logged the remote IP on invalid logins so `fail2ban` would ban the IP after too many retries).

I will open an issue but I'm very surprised nobody noticed this before.

FroMage:
Added https://github.com/quarkusio/quarkus/issues/4922

kenfinnigan:
What is the likelihood of this being merged today?

I ask because I have a PR for MP REST Client TCKs that needs the new RESTEasy version

gsmet:
From my point of view: it's mergeable as soon as CI is green.

I just added some nice warnings for the users so there shouldn't be any migration issue.

But we need someone to review it and approve it. And people to generally agree with the approach.

kenfinnigan:
@gsmet FYI this PR now has a conflict and needs rebasing

gsmet:
This is going to be a lot of fun if it does not get merged quickly :).

gsmet:
@kenfinnigan rebased!

bom/runtime/pom.xml
```diff
@@ -72,17 +80,15 @@
         <commons-logging-jboss-logging.version>1.0.0.Final</commons-logging-jboss-logging.version>
         <commons-lang3.version>3.9</commons-lang3.version>
         <commons-codec.version>1.13</commons-codec.version>
-        <validation-api.version>2.0.1.Final</validation-api.version>
         <classmate.version>1.3.4</classmate.version>
-        <javax.el-impl.version>3.0.1-b11</javax.el-impl.version>
-        <hibernate-validator.version>6.1.0.Alpha6</hibernate-validator.version>
+        <jakarta.el-impl.version>3.0.3</jakarta.el-impl.version>
```
kenfinnigan: Should this be moved to where all the other `jakarta.*` properties are?

It can be done in a later PR if it's easier

kenfinnigan (approved):
It would be good if @stuartwdouglas can take a look as well.

Also, should the commits be squashed?

bom/runtime/pom.xml
gsmet: Yeah, it used to be there because it was added with HV. But I can see how it makes sense to move them all at the same place now.

gsmet:
About the commits being squashed: I think we shouldn't squash them so that we can easily revert a part of the approach if it's deemed unpractical. There's a good chance we would keep some of them anyway.

kenfinnigan:
no worries, just asking

geoand:
Looks good to me as well, although my perspective is pretty limited

bom/runtime/pom.xml
gsmet: I reordered things.

gsmet:
I tested the quickstarts and it was pretty much OK, except for the Kafka one which displayed a warning but this was an error in the quickstart which I fixed here: https://github.com/quarkusio/quarkus-quickstarts/pull/340/files .

stuartwdouglas:
This is a big change to make so close to 1.0. I am mostly ok with the implementation, although the banned dependency mechanism will need some work in the future (to make it pluggable, and also to allow banned dependencies to be automatically changed to good ones).

@emmanuelbernard @n1hility are you guys ok with a change this big so close to release?

gsmet:
Yeah, I agree the banned dependencies is very naive for now but it does the work for what we need currently.

I think with the dependencies banned by the enforcer and the warning mechanism, we are pretty safe. We will be able to tune things further if needed depending on the user reports.

emmanuelbernard:
Go.

Sanne:
awesome

PR #4912, opened 2019-10-27
Remove war-launcher

stuartwdouglas:
This was added in the proof of concept phase to provide that this approach was compatible with traditional war archives, however after that it has not been looked at since. It has no tests and is almost certainly 100% broken.


gastaldi (approved):

cescoffier (approved):

cescoffier:
Marking it as noteworthy just in case someone was relying on this. Hopefully... no one was.

PR #4913, opened 2019-10-27
Add a config property to disable Hibernate second level cache

### LINKED ISSUE #4489: Expose Hibernate ORM option to disable 2LC
**Description**
Currently the 2LC is enabled globally, so to opt-in on specific entities.

During development, it's handy to be able to turn it off quickly to diagnose issues. This is currently possible, but only when configuring ORM via the persistence.xml - which makes it quite inconvenient.

gwenneg:
Fixes #4489

For now, I only exposed the `hibernate.cache.use_second_level_cache` Hibernate setting through the `quarkus.hibernate-orm.use-second-level-cache` Quarkus config property (which defaults to `true`), but I'm wondering if some or all of [the other settings we're enabling by default](https://github.com/quarkusio/quarkus/blob/master/extensions/hibernate-orm/runtime/src/main/java/io/quarkus/hibernate/orm/runtime/boot/FastBootMetadataBuilder.java#L333-L339) should be exposed too:

- hibernate.cache.use_reference_entries=true
- hibernate.cache.use_query_cache=true
- javax.persistence.sharedCache.mode=ENABLE_SELECTIVE

### DIFF
--- a/docs/src/main/asciidoc/hibernate-orm.adoc
+++ b/docs/src/main/asciidoc/hibernate-orm.adoc
@@ -450,7 +450,11 @@ On top of immutable data, in certain contexts it might be acceptable to enable c
 Rather than enabling caching on mutable data, ideally a better solution would be to use a clustered cache; however at this time Quarkus doesn't provide any such implementation: feel free to get in touch and let this need known so that the team can take this into account.
 --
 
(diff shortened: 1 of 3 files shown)


Sanne:
Thanks for starting this!

I'll need to test this, as I suspect it's not enough to change these settings - we hardcoded some assumptions.

Sanne:
sorry it took so long to come back to this. Sent a different PR, we can close this now. Thanks!

Issue #4815, opened 2019-10-23
No hot reload on bad configuration wich block the quarkus startup and when http.port is overridden 


Mefl:
**Describe the bug**
If you configure a bad value that block the first application start, there is no way to have hot reload.

If you start on a good value, but after plug a bbad config, the hot reload is ok. A configuration change will trigger the hot reload as wanted. 

**Expected behavior**
The hot reload is trigger if you change the configuration file, or a code change, even it never could be started correctly.

**Actual behavior**
The hot reload is broken if the app never could be started a first time.

**To Reproduce**
1. configure a bad jdbc url on hibernate (or flyway)
2. start the app 
3. change to good value (it won't reload)

**Environment (please complete the following information):**
 - Output of `uname -a` or `ver`: Darwin 18.5.0 Darwin Kernel Version 18.5.0
 - Output of `java -version`: openjdk version "13" 2019-09-17
 - GraalVM version (if different from Java): 
 - Quarkus version or git rev: 0.25.0

**Additional context**
(Add any other context about the problem here.)
@cescoffier the issue that we discuss after the quarkus workshop

gsmet:
Hum, this used to work IIRC. Adding the `regression` label.

jaikiran:
I gave this a try with both 0.25.0 and latest upstream master. It works fine on both these versions.

What I did was included an incorrect JDBC URL in application.properties:
```
quarkus.datasource.url=jdbc2:postgresql://localhost/hibernate_orm_test
quarkus.datasource.driver=org.postgresql.Driver

quarkus.hibernate-orm.log.sql=false
quarkus.hibernate-orm.database.generation=drop-and-create
```
(notice the datasource.url has an incorrect value `jdbc2:...`)
and in my REST endpoint code, just created a connection:

```
@GET
    @Produces(MediaType.TEXT_PLAIN)
    public String hello() {
    	try (final Connection conn = ds.getConnection()) {
    		System.out.println("got connection " + conn);
    	} catch (Exception e) {
    		e.printStackTrace();
    	}
        return "hello2";
    }
```

The first time I accessed this endpoint, in dev mode:
```
curl localhost:8080/hello
```
I did see the exception in the logs (as expected):

```
java.sql.SQLException: Driver does not support the provided URL: jdbc2:postgresql://localhost/hibernate_orm_test
	at io.agroal.pool.ConnectionFactory.connectionSetup(ConnectionFactory.java:215)
[...]

PR #4917, opened 2019-10-28
If start fails make sure to shutdown server

stuartwdouglas:
It's possible start can fail after the server
has started in dev mode. We need to make sure it
is shut down to recover properly.

Possible fix for #4815


cescoffier (approved):

cescoffier:
Is it possible to write a test? 
Wondering if the HotReload test support would work in this case.

stuartwdouglas:
I think you would need to write an extension that was designed to fail at this particular place, otherwise it would be racy.

I don't really think it is worth it.

PR #4923, opened 2019-10-28
Improves Agroal messages

### LINKED ISSUE #4808: Agroal to not log warnings for default configuration
**Describe the bug**
The default in Quarkus for initial-size is 0; when setting any min-size, this will produce a scary warning about 

    "Initial size not whithin min / max bounds"

(warning is new since #4754 )

We should either:

- automatically set the initial-size to match the min size (when it's omitted from configuration)
- de-emphasize the warning to info level
- ?

Sanne:
Fixes #4808

### DIFF
--- a/extensions/agroal/runtime/src/main/java/io/quarkus/agroal/runtime/AgroalEventLoggingListener.java
+++ b/extensions/agroal/runtime/src/main/java/io/quarkus/agroal/runtime/AgroalEventLoggingListener.java
@@ -62,6 +62,11 @@ final class AgroalEventLoggingListener implements AgroalDataSourceListener {
     }
 
     @Override
+    public void onInfo(String message) {
+        log.infov("{0}: {1}", datasourceName, message);
+    }
+
+    @Override
     public void onWarning(Throwable throwable) {
         log.warnv("{0}: {1}", datasourceName, throwable.getMessage());
         log.debug("Cause: ", throwable);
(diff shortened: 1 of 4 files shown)


extensions/agroal/runtime/src/main/java/io/quarkus/agroal/runtime/AgroalEventLoggingListener.java
```diff
@@ -61,6 +61,11 @@ public void onWarning(String warning) {
         log.warnv("{0}: {1}", datasourceName, warning);
     }
 
+    @Override
+    public void onInfo(String message) {
+        log.infov("{0}: {1}", datasourceName, message);
```
gsmet: Quick question that might be stupid: so now, we will log the message about the initial size at info level? Does it really solve the problem? I mean most users will run their application with the default log level.

Sanne: > Quick question that might be stupid: so now, we will log the message about the initial size at info level?

Yes.

> Does it really solve the problem?

What problem exactly? See also discussion on #4808 

It will read like this now, wording suggested by Loic; too late to change the wording as that requires an Agroal release.

` INFO  [io.agr.pool] (main) Datasource '<default>': Initial size smaller than min. Connections will be created when necessary
`

We could make it entirely silent by setting the default min-size to zero: it's 5 now and I'm not entirely sure why.

Sanne: ah that's actually not the wording Loic suggested. Looks like it was too late for the Agroal train.

Sanne:
@gsmet updated, as discussed in chat.

gsmet (approved):

gsmet:
Merged, thanks!

PR #4924, opened 2019-10-28
Ensure that vert.x is closed after Arc

### LINKED ISSUE #4834: Tests are hanging when using a Kafka consumer and the mailer.
**Describe the bug**
Tests are hanging when using a Kafka consumer and the mailer.
The simple POC provide a simple kafka consumer and just a dependency to `quarkus-mailer` to describe this bug.

**Expected behavior**
At the end of the test we should see this message:

`2019-10-24 13:15:54,718 INFO  [io.sma.rea.mes.ext.MediatorManager] (main) Cancel subscriptions`

Next junit should stop and display tests statuses.

**Actual behavior**

The subscription is not canceled. Tests are hanging forever.

**To Reproduce**
1. git clone https://github.com/dcdh/poc-mailer-kafka
2. start containers by running `run_containers.sh` script (warning: it will kill and remove all running containers)
3. execute test by running `mvn clean test`

**Configuration**
application.properties is provided in the POC.

**Screenshots**

 - Output of `uname -a` or `ver`: Linux arch-anywhere 5.3.1-arch1-1-ARCH #1 SMP PREEMPT Sat Sep 21 11:33:49 UTC 2019 x86_64 GNU/Linux
 - Output of `java -version`: 
openjdk version "1.8.0_202"
OpenJDK Runtime Environment (build 1.8.0_202-20190206132807.buildslave.jdk8u-src-tar--b08)
OpenJDK GraalVM CE 1.0.0-rc16 (build 25.202-b08-jvmci-0.59, mixed mode)
 - Quarkus version or git rev: 0.26.1

**Additional context**
To make it working as expected you'll have to remove `quarkus-mailer` dependency however it is not possible in a real application using both kafka and mailer.

My objective is to be able to consume events from kafka to send mails from them.

geoand:
Fixes: #4834

### DIFF
--- a/core/deployment/src/main/java/io/quarkus/deployment/builditem/ShutdownContextBuildItem.java
+++ b/core/deployment/src/main/java/io/quarkus/deployment/builditem/ShutdownContextBuildItem.java
@@ -23,4 +23,9 @@ public final class ShutdownContextBuildItem extends SimpleBuildItem
     public void addShutdownTask(Runnable runnable) {
         throw new IllegalStateException();
     }
+
+    @Override
+    public void addLastShutdownTask(Runnable runnable) {
+        throw new IllegalStateException();
+    }
 }
--- a/core/runtime/src/main/java/io/quarkus/runtime/ShutdownContext.java
+++ b/core/runtime/src/main/java/io/quarkus/runtime/ShutdownContext.java
@@ -8,4 +8,7 @@ package io.quarkus.runtime;
 public interface ShutdownContext {
 
     void addShutdownTask(Runnable runnable);
+
+    // these are executed after all the ones add via addShutdownTask in the reverse order from which they were added
+    void addLastShutdownTask(Runnable runnable);
 }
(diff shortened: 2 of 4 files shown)


geoand:
I am not sure this is proper solution, but from debugging #4834 it seems that we really need to close Vert.x after Arc, otherwise some shutdown tasks of that depend on vert.x to be executed (like in the case of Kafka) can get deadlocked.

cescoffier (approved):

cescoffier:
LGTM, waiting for Stuart opinion.

stuartwdouglas:
I don't really like this, but it will do for now.

At some point I think we are going to have to revisit how we handle ordering of both startup and shutdown tasks.

stuartwdouglas (approved):

geoand:
> I don't really like this, but it will do for now.
> 
> At some point I think we are going to have to revisit how we handle ordering of both startup and shutdown tasks.

Makes sense. The ordering that exists now is totally implicit and susceptible to these kinds of issues.

PR #4927, opened 2019-10-28
Disable Hibernate ORM's inline dirty checking

### LINKED ISSUE #3261: Hibernate cannot track changes of primitive boolean value
**Describe the bug**
Given an entity with a single boolean attribute and a key (an EmbeddedId in my test-case), the changes are tracked only if you change from false to true, instead changing from true to false is not persisted.

As a workaround I shifted to Boolean and it works.

I investigated and the generated method is done this way:
```java
    public void $$_hibernate_write_abilitata(boolean var1) {
        if (var1 - this.abilitata != 0) {
            this.$$_hibernate_trackChange("abilitata");
        }

        if (this.$$_hibernate_getInterceptor() != null) {
            this.abilitata = this.$$_hibernate_getInterceptor().writeBoolean(this, "abilitata", this.abilitata, (boolean)var1);
        } else {
            this.abilitata = (boolean)var1;
        }
    }

```
and I don't know how the JVM works with the bytecode that is decompiled that way (boolean cannot be subtracted), with the values in the stack and the OpCode.ISUB. 

**Expected behavior**
The JPA Provider should update in both case

**Actual behavior**
Only changes to true are persisted

**To Reproduce**
1. mvn install the attached test-case
2. see the two test, one with false, one with true
3. the one that starts with true fails

**Configuration**
```properties
# H2 Database in memory
quarkus.datasource.url: jdbc:h2:mem:testdb
quarkus.datasource.driver: org.h2.Driver
quarkus.datasource.username: sa
quarkus.datasource.password:

quarkus.hibernate-orm.database.generation=drop-and-create

quarkus.hibernate-orm.log.sql: true
```

 - Darwin WMICTLM1P.local 18.6.0 Darwin Kernel Version 18.6.0: Thu Apr 25 23:16:27 PDT 2019; root:xnu-4903.261.4~2/RELEASE_X86_64 x86_64
 - Output of `java -version`: 
openjdk version "1.8.0_212"
OpenJDK Runtime Environment (build 1.8.0_212-20190523183630.graal2.jdk8u-src-tar-gz-b03)
OpenJDK 64-Bit GraalVM CE 19.0.2 (build 25.212-b03-jvmci-19-b04, mixed mode)
 - Quarkus version or git rev: 0.19.1

[test-case.zip](https://github.com/quarkusio/quarkus/files/3402721/test-case.zip)

### LINKED ISSUE #4377: Hibernate - @ManyToOne strange behavior
**Describe the bug**
Imagine the relationship described below:
"one" User -> "many" Message
In Message class, there is @ManyToOne(fetch = FetchType.LAZY, optional = false) User user;  annotation.
When a new Message instance are created, we need to persist the object on database with related User "id".
I find the related User using EntityManager.find(clazz, id)

**Expected behavior**
Insert the new Message on database, do not make any operation with User

**Actual behavior**
After insert Message, Hibernate tries to update the User instance without any field.

**To Reproduce**
1. Create a User instance on database and persist.
2. In other transaction, create a Message and set the existent User
3. Persist the Message

**Configuration**
```properties
# We are using Postgres Database

```
Can reproduce issue only in quarkus 0.23x, on 0.22 works fine with same config.

dreab8:
the PR fixes https://github.com/quarkusio/quarkus/issues/3261

Also fixes https://github.com/quarkusio/quarkus/issues/4377

### DIFF
--- a/extensions/hibernate-orm/deployment/pom.xml
+++ b/extensions/hibernate-orm/deployment/pom.xml
@@ -76,6 +76,11 @@
             <artifactId>jaxb-runtime</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-resteasy-jackson</artifactId>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 
     <build>
--- a/extensions/hibernate-orm/deployment/src/main/java/io/quarkus/hibernate/orm/deployment/HibernateEntityEnhancer.java
+++ b/extensions/hibernate-orm/deployment/src/main/java/io/quarkus/hibernate/orm/deployment/HibernateEntityEnhancer.java
@@ -4,6 +4,7 @@ import java.util.function.BiFunction;
 
 import org.hibernate.bytecode.enhance.spi.DefaultEnhancementContext;
 import org.hibernate.bytecode.enhance.spi.Enhancer;
+import org.hibernate.bytecode.enhance.spi.UnloadedClass;
 import org.hibernate.bytecode.spi.BytecodeProvider;
 import org.objectweb.asm.ClassReader;
 import org.objectweb.asm.ClassVisitor;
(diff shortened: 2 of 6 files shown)


geoand:
This needs the imports to be sorted (`mvn process-resources`)

Sanne:
There's a Panache test failing; this will need to wait for @FroMage 's opinion

_io.quarkus.it.panache.TestEndpoint#testModel1_

Sanne:
Closing this, we won't do this - at least not now. Thanks!

Issue #4919, opened 2019-10-28
Offer a way to blacklist GraalVM metadata from JARs


emmanuelbernard:
Some dependencies define GraalVM metadata in `MANIFEST`. But some goes against possible optimization of Quarkus. We need a mechanism to blacklist GraalVM metadata for specific dependencies.

geoand:
@galderz IIRC there is now some way to do this?

Sanne:
> @galderz IIRC there is now some way to do this?

was about to ask the same :) We'll need it for #19055

galderz:
Hmmmm, I've added https://github.com/oracle/graal/pull/3179 (part of 21.1 onwards), which allows graalvm to exclude the configuration from certain resources within jars. However, I'm not sure if this is exactly what @emmanuelbernard was asking when talking about `MANIFEST`.

Sanne:
Related: #21069 

And since that one is more specific and @galderz already solved oracle/graal#3179 I think we're good - closing this one.

PR #4930, opened 2019-10-28
Update Netty to 4.1.42.Final

cescoffier:


Sanne:
This looks sane, but I'm wondering: during last meeting @emmanuelbernard mentioned you badly needed to disable some native-image directives being picked up from the Manifest in the Netty jar; is that no longer the case?

asking so to affect priority for #4919

gsmet:
Does it bring something to the plate? If we don't strictly need it, I would just postpone that for now.

stuartwdouglas:
@gsmet I assume that comment about postponing was in relation to Sanne, not to this PR in general?

stuartwdouglas (approved):

emmanuelbernard:
The new release does bring CVE fixes.

stuartwdouglas:
Then let's update

stuartwdouglas (approved):

PR #4932, opened 2019-10-28
Support outputDirectory in CreateProjectMojo

gastaldi:
This will allow the quarkus:create to be used on any directory and also useful in pom.xml when wanting to generate a project during the build


geoand:
So I assume this is also useful in integration tests?

gastaldi:
Yes, specially for integration tests :smiley:

geoand (approved):

gsmet:
If we are playing with that, could we make the command entirely consistent and always create the project in a new directory?

The current behavior is highly disturbing, especially when you have hidden files in the directory.

gsmet:
Note that it doesn't have to be done in this PR. I don't want to block it.

gastaldi:
So the default value for `outputDirectory` should be `${basedir}/${projectArtifactId}` ?

PR #4933, opened 2019-10-28
Only execute Vault ITs if -Dtest-vault is specified

### LINKED ISSUE #4925: quarkus-integration-test-vault-app is requiring Docker to build
**Describe the bug**
The module `quarkus-integration-test-vault-app` is expecting the docker daemon to be running; not having it has the whole Quarkus build fail.

Other modules have a conditional dependency controlled by a flag to enable any integration tests requiring docker.

This is particularly annoying for people building Quarkus without docker (possibly because working on an unrelated set of modules), or using `podman` instead of docker.

gsmet:
Fixes #4925

### DIFF
--- a/integration-tests/vault-app/pom.xml
+++ b/integration-tests/vault-app/pom.xml
@@ -75,7 +75,6 @@
             <plugin>
                 <groupId>io.quarkus</groupId>
                 <artifactId>quarkus-maven-plugin</artifactId>
-                <version>${project.version}</version>
                 <executions>
                     <execution>
                         <goals>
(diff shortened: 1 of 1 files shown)


gsmet:
@Sanne this should fix the problem you reported.

integration-tests/vault-app/pom.xml
```diff
@@ -75,7 +75,6 @@
             <plugin>
                 <groupId>io.quarkus</groupId>
                 <artifactId>quarkus-maven-plugin</artifactId>
-                <version>${project.version}</version>
```
Sanne: isn't this reverting what you just blessed in https://github.com/quarkusio/quarkus/pull/4843 ?

Sanne:
If we ignore the maven plugin version I commented about, the other changes to indeed fix the build. Thanks!

integration-tests/vault-app/pom.xml
gsmet: The issue is only if the plugin is in a profile. If it is not in a profile, it's OK. That's why I haven't remove the version in the profile below.

Sanne (approved):

PR #4937, opened 2019-10-28
Fix broken quarkusBuild task

geoand:


gastaldi (approved):
I can confirm that this PR fixes the issue

devtools/gradle/src/main/java/io/quarkus/gradle/tasks/QuarkusBuild.java
```diff
@@ -132,6 +133,12 @@ public void buildQuarkus() {
 
         final AppArtifact appArtifact = extension().getAppArtifact();
         final AppModelResolver modelResolver = extension().resolveAppModel();
+        try {
+            // this needs to be done otherwise the app artifact doesn't get a proper path
+            modelResolver.resolveModel(appArtifact);
```
aloubyansky: Ah, ok, then modelResolver.resolve(appArtifact) should be enough. Resolving the model resolves all the deps + all the deployment deps, which is a waste in this case.

geoand: Let me give it a try.

geoand: I tried `modelResolver.resolve(appArtifact)` but the problem reappeared.

aloubyansky: yes, it's been awhile, the gradle's resolve doesn't support the resolve method. you can merge this pr. It'll cache the model and re-use it during the augmentation. So, not a waste.

aloubyansky (approved):

PR #4939, opened 2019-10-28
Remove the warning about not having undertow

gsmet:
That's definitely not something we want to keep for the next releases.

_Not sure it's something we want to merge right away for this release but I want it for the next one so I prefer creating it now to not forget about it._


gsmet:
@geoand if you have the time to check this out, it would be nice to get this approved. This way, I will merge it as soon as 0.27 is branched and we will be done with it.

geoand (approved):
Makes sense to me

PR #4950, opened 2019-10-28
After failed start make sure to read the HTTP config with the correct TCCL

### LINKED ISSUE #4815: No hot reload on bad configuration wich block the quarkus startup and when http.port is overridden

stuartwdouglas:
… TCCL

Fixes #4815

### DIFF
--- a/core/devmode/src/main/java/io/quarkus/dev/DevModeMain.java
+++ b/core/devmode/src/main/java/io/quarkus/dev/DevModeMain.java
@@ -96,10 +96,6 @@ public class DevModeMain implements Closeable {
             if (context.isAbortOnFailedStart()) {
                 throw new RuntimeException(deploymentProblem == null ? compileProblem : deploymentProblem);
             }
-            log.error("Failed to start Quarkus, attempting to start hot replacement endpoint to recover");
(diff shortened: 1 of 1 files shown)


geoand (approved):
Tested with the scenario I mentioned in #4934 was failing and with this PR it works

PR #4952, opened 2019-10-28
Reduce CI load on failing PR

stuartwdouglas:
This PR changes the staging of CI jobs to not run native tests
if the JVM tests fail. It also removes the dedicated 'build for
native' job, and instead uses the results of the JDK8 build
to run the native build, further reducing the load.


gsmet (approved):
Makes sense.

PR #4956, opened 2019-10-29
HTTP Permissions Refactor

stuartwdouglas:


stuartwdouglas:
It may take a while for quarkus-http to hit central

geoand:
There is also a conflict now

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/security/HttpSecurityPolicy.java
```diff
@@ -1,30 +1,74 @@
...
+        }
+    }
+
+    /**
+     * A context object that can be used to run blocking tasks
+     * <p>
+     * Blocking identity providers should used this context object to run blocking tasks, to prevent excessive and
+     * unnecessary delegation to thread pools
+     */
+    interface AuthorizationRequestContext {
```
sberyozkin: @stuartwdouglas Can `AuthorizationRequestContext` be optionally injected when needed into `HttpSecurityPolicy` implementations ? May be `HttpAuthorizer`'s `CONTEXT` can be CDI-produced ? So that the non-blocking implementations won't see it in the signature.
Or may be simpler, just renamed to something like `BlockingAuthorizationRequestContext` ?

stuartwdouglas: It may end up eventually containing more contextual information

sberyozkin: OK

pedroigor (approved):

gsmet:
quarkus-http Beta3 has hit Central.

I rebased this one to be sure we don't have the doc issue, let's see what CI has to say.

gsmet:
Closing. Will be part of a larger PR.

Issue #4455, opened 2019-10-09
Read timeout for Servlet


stuartwdouglas:
Undertow should have a read timeout for blocking IO. This should be inherited from the vert.x config.

rsvoboda:
@michalszynkiewicz what's the status and plan here ?

michalszynkiewicz:
So far i exposed the idle timeout (that's the thing vert.x gives us for free), it made the issue less urgent.
I have it on my todo list but haven't come back to it yet. 
If it's urgent, I can pick it up in parallel with my current assignment. Do you find it high prio?

rsvoboda:
I don't want to dictate priority, let's see what @gsmet thinks

PR #4963, opened 2019-10-29
Idle timeout for blocking http

michalszynkiewicz:
refs #4455

### DIFF
--- a/extensions/security/deployment/src/main/java/io/quarkus/security/deployment/SecurityAnnotationsRegistrar.java
+++ b/extensions/security/deployment/src/main/java/io/quarkus/security/deployment/SecurityAnnotationsRegistrar.java
@@ -22,7 +22,7 @@ public class SecurityAnnotationsRegistrar implements InterceptorBindingRegistrar
     public static final Map<DotName, Set<String>> SECURITY_BINDINGS = new HashMap<>();
 
     static {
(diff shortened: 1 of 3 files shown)


gsmet:
This uses a SNAPSHOT dependency, it shouldn't be merged.

michalszynkiewicz:
requires https://github.com/quarkusio/quarkus-http/pull/10

gsmet:
I force pushed the Beta3 version change as it has been pushed to Central.

let's see what CI has to say.

michalszynkiewicz:
@gsmet  the released version does not contain https://github.com/quarkusio/quarkus-http/pull/10 so the tests will fail

test-framework/common/src/main/java/io/quarkus/test/common/RestAssuredURLManager.java
```diff
@@ -92,6 +91,15 @@ public void setURL() {
         }
     }
 
+    public static String getHost() {
```
stuartwdouglas: You should not need these changes. Just inject the URL using @TestHTTPResource and then get the host and port from the URL.

stuartwdouglas (changes requested):
Can we get a version of this with just the vert.x idle timeout changes for the 28.0 release?

michalszynkiewicz:
@stuartwdouglas uploaded setting the timeout only

stuartwdouglas (approved):

PR #4970, opened 2019-10-29
Add jackson jdk8 modules

### LINKED ISSUE #4860: Provides Jackson Java 8 modules with quarkus-jackson
**Description**
Quarkus provides Jackson 2 capabilities via the `quarkus-jackson` extension.
Jackson 2 is Java 7 compatible and specific Java 8 data types needs to be registred via the following [Java 8 modules](https://github.com/FasterXML/jackson-modules-java8): 

- com.fasterxml.jackson.datatype.jsr310.JavaTimeModule: for Java DatTime API
- com.fasterxml.jackson.datatype.jdk8.Jdk8Module: for Optional
- com.fasterxml.jackson.module.paramnames.ParameterNamesModule: to avoid `@JsonProperty` on constructor and creator methods (if compiled with `-parameters` javac option)

Using the two first modules is very frequent, so it is a pitty that each developers wanting to use Jackson with Quarkus has to add them manually. It can also leads to issues when migrating from JsonB to Jackson as JsonB correctly handle these types by defaults.

**Implementation ideas**
Add **JavaTimeModule** and **Jdk8Module** and document them, they are small and when we will migrate to Jackson 3 they will be included by default.

Don't inlcude the ParameterNamesModule by default as it needs a compiler option so it's usage is less frequent.

I can provide a PR for this :)

loicmathieu:
Add by default the 3 Jackson JDK8 modues as Quarkus is JDK 8 minimum.
This adds around 100k of libraries ...

Fixes #4860

I add a LocalDate to the resteasy-jackson integration test to check that the module is enabled.

### DIFF
--- a/extensions/jackson/runtime/pom.xml
+++ b/extensions/jackson/runtime/pom.xml
@@ -19,6 +19,18 @@
             <artifactId>jackson-databind</artifactId>
         </dependency>
         <dependency>
+            <groupId>com.fasterxml.jackson.datatype</groupId>
+            <artifactId>jackson-datatype-jsr310</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.datatype</groupId>
+            <artifactId>jackson-datatype-jdk8</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.module</groupId>
+            <artifactId>jackson-module-parameter-names</artifactId>
+        </dependency>
+        <dependency>
             <groupId>io.quarkus</groupId>
             <artifactId>quarkus-arc</artifactId>
         </dependency>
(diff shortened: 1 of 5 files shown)


Sanne (approved):

gsmet:
Thanks @loicmathieu , looks like a detail but a really nice contribution to get things easier for the users.

loicmathieu:
@gsmet I usually always configure date serialized as ISO Date but I didn't turn it on by default as no serialization properties are specified yet by default for Jackson. Date serialization is always hard!

And JSON-B and Jackson has different default today, this can be disturing for someone that want to switch from one to the other.

Jackson is not well documented (it is only part of the writing rest json service guide: https://quarkus.io/guides/rest-json-guide#jackson) maybe we can add to the jackson customizer example a serialization property to be more explicit.

loicmathieu:
@gsmet test failed with a connection refused error. Something seems to be wrong with the CI.
Is there a way to relaunch the CI buid ?

gsmet:
Failures are unrelated. Let's merge.

gsmet:
Thanks @loicmathieu

Issue #4967, opened 2019-10-29
need pinned defined for categories


maxandersen:
code.quarkus.io need pinned for categories.

https://github.com/quarkusio/quarkus/blob/master/devtools/bom-descriptor-json/src/main/resources/extensions-overrides.json needs updating with  a "pinned" attribute like:

```
"categories": [
    {
      "name": "Core",
      "id": "core"
       "metadata": [
           "pinned": [
                "io.quarkus:core-log",
                ...
]
],
 },
...
```

The pins are each groupId:artifactId for what should be put at top and sorted as listed in this array.

maxandersen:
this is done.

PR #4976, opened 2019-10-29
4967 Add pinned info and description for categories

emmanuelbernard:
We don't need to use description in the UI just yet but it's there now.
Also added a reactive driver into the reactive section


ia3andy (approved):
Cool !

emmanuelbernard:
Fix for https://github.com/quarkusio/quarkus/issues/4967

maxandersen:
Thanks - pushed after removing the superfluous period at the end of each descripition since other descriptions we have don't have it.

emmanuelbernard:
Ah, well I was following proper English since some had multiple sencences. But better use consistency.

PR #4978, opened 2019-10-29
Adapt the Maven project creation logic to always create a sub-directory

cescoffier:


cescoffier:
Once merged, the guides must be updated to explicitly indicate to "cd" in the created directory.

maxandersen:
> Once merged, the guides must be updated to explicitly indicate to "cd" in the created directory.

If we don't merge, the guides must updated to explicitly indicate to "mkdir" + "cd" into the directory before running  the commands ;)

j/k - change looks good - makes the behavior consistent for everyone.

maxandersen (approved):
LGTM - now we can update the guides to follow one deterministic approach.

gsmet (approved):

gsmet:
OK, it mostly passed the first build except for the documentation which is unrelated. I don't see why it would fail elsewhere. Let's merge and be done with it.

PR #4979, opened 2019-10-29
HTTP Authorization Refactor

stuartwdouglas:
This is a subset of https://github.com/quarkusio/quarkus/pull/4956 that does not need the new quarkus-http release. If the other artifact is not published in time then ideally this one will go in as it fixes some problems (such as blocking ops on the IO thread)

### DIFF
--- a/extensions/keycloak-authorization/runtime/src/main/java/io/quarkus/keycloak/pep/KeycloakPolicyEnforcerAuthorizer.java
+++ b/extensions/keycloak-authorization/runtime/src/main/java/io/quarkus/keycloak/pep/KeycloakPolicyEnforcerAuthorizer.java
@@ -6,6 +6,7 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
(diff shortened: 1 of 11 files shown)


sberyozkin:
@stuartwdouglas It all looks good, I'd only consider making it obvious that context is meant for the blocking calls (or even move it out of the signature if it is easy enough to do), as per [this comment](https://github.com/quarkusio/quarkus/pull/4956#discussion_r340012037), unless you plan to add more operations to `AuthorizationRequestContext`, for the users be able to do more than just the blocking checks on this context instance.
What do you think ?

stuartwdouglas:
Eventually it may contain more context info, so I would rather leave it to future proof.

sberyozkin:
@stuartwdouglas OK, that makes sense. So may be it is worth pushing the doc about the blocking calls from the class itself to the method which does the blocking call ?

sberyozkin (approved):
LGTM

pedroigor (approved):

gsmet:
Quarkus HTTP Beta3 has hit Central so I'm closing this one.

PR #4980, opened 2019-10-29
issue 4977 - Move short-name under metadata

emmanuelbernard:
Run tests locally


emmanuelbernard:
Taking a risk and pushed without tests on CI.

PR #4954, opened 2019-10-29
Fixes #4520

kenfinnigan:

PR #4984, opened 2019-10-29
Make rest-client aware of deployment profiles

### LINKED ISSUE #4983: rest-client doesn't take deployment profile when reading config
_From the mailing list:_

I have the following configuration (Quarkus 0.26.1) in application.properties:

```
%dev.de.wwu.ulb.registry.RegistrationClient/mp-rest/url=http://wwurdm.uni-muenster.de/registration/api/v1
%dev.de.wwu.ulb.registry.RegistrationClient/mp-rest/scope=javax.inject.Singleton
%prod.de.wwu.ulb.registry.RegistrationClient/mp-rest/url=http://registration-service:8080/registration/api/v1
%prod.de.wwu.ulb.registry.RegistrationClient/mp-rest/scope=javax.inject.Singleton
```

to distinguish dev and prod profile. However, the rest client won't get injected when starting in prod profile while

```
de.wwu.ulb.registry.RegistrationClient/mp-rest/url=http://registration-service:8080/registration/api/v1

de.wwu.ulb.registry.RegistrationClient/mp-rest/scope=javax.inject.Singleton
```

works. Shouldn't the first version work as well?

geoand:
Fixes: #4983

### DIFF
--- a/extensions/rest-client/runtime/src/main/java/io/quarkus/restclient/runtime/RestClientBase.java
+++ b/extensions/rest-client/runtime/src/main/java/io/quarkus/restclient/runtime/RestClientBase.java
@@ -5,7 +5,7 @@ import java.net.URL;
 import java.util.Optional;
 
 import org.eclipse.microprofile.config.Config;
-import org.eclipse.microprofile.config.ConfigProvider;
+import org.eclipse.microprofile.config.spi.ConfigProviderResolver;
 import org.eclipse.microprofile.rest.client.RestClientBuilder;
 
 public class RestClientBase {
@@ -24,7 +24,8 @@ public class RestClientBase {
         this.proxyType = proxyType;
         this.baseUriFromAnnotation = baseUriFromAnnotation;
         this.propertyPrefixFromAnnotation = propertyPrefixFromAnnotation;
-        this.config = ConfigProvider.getConfig();
+        // this is the proper way to obtain the configuration that takes the deployment profiles into account
+        this.config = ConfigProviderResolver.instance().getConfig();
     }
 
     public Object create() {


geoand:
@cescoffier @kenfinnigan given that the only proper way to test this is to add a test that includes `%test....` and `%prod....` and the fact that those tests are slow hard to maintain, would you like me to add it or not?

gsmet:
Stupid question: we don't have other areas doing the same mistake? It looks like an honest one.

geoand:
That's a great question, let me check

geoand:
We have a lot of them unfortunately... I am not sure what we should do so close to a release...

kenfinnigan:
Instead of tests, is there a maven plugin we can use that looks for the bad way to retrieve config and fails the build if it's present?

kenfinnigan:
I can fix this one as part of my REST TCK work, as it needs to be rebased anyway

geoand:
@kenfinnigan not sure, but that would fail master unfortunately because there are places where the other way of doing things is valid and is being used and some where I'm not sure. This close to release I am very hesitant to make such across the board changes

kenfinnigan:
Wasn't suggesting now, just trying to think of a way to prevent them from being added in the future

geoand:
@kenfinnigan yes, we will need to address that for sure

geoand:
CI failure seems to be due to Azure infra. I'll restart the failing tests once the others finish

kenfinnigan:
I've included this fix in the REST TCK fixes, so we could close

geoand:
OK, closing as #4954 includes the fix this PR proposed

PR #4985, opened 2019-10-29
Make const public in descriptor and add new consts

ia3andy:
- added in Extension: `MD_STATUS`
- added in Category: `MD_PINNED`


ia3andy:
@maxandersen it shouldn't affect anything, this is for me to use from Code Quarkus ;-)

maxandersen (approved):
LGTM.

Bonus point next time to add docs to new values ;)

gsmet:
OK, doc is failing due to an unrelated issue. Otherwise first Java build is OK and I can live with that.

PR #4987, opened 2019-10-29
Remove metrics dependency from fault-tolerance extension

### LINKED ISSUE #4857: quarkus-smallrye-fault-tolerance extension should not have a hard dependency to quarkus-smallrye-metrics
**Describe the bug**
When including the `quarkus-smallrye-fault-tolerance` extension, we automatically have the metrics extension as it have a hard dependency to it.

**Expected behavior**
Metrics should be included only if the `quarkus-smallrye-metrics` extension is added in the pom.xml

 - Quarkus version or git rev: 0.26 or master

**Additional context**
The Jaeger extension only includes `smallrye-metrics` maybe the same approach can be use to avoid having the metrics extension directly as a dependency. 
But this does not fully solves the issue of having an un-necessary dependency (smallrye-metrics).

loicmathieu:
Remove the dependency between `quarkus-smallrye-fault-tolerance` and `quarkus-smallrye-metics`.

With this PR, the metrics extension will no longuer be included when the fault-tolerance extension is used.

The fault-tolerance extension now have a dependency to microprofile-fault-tolerance instead and provides a NOOP MetricsRegistry when the metrics extension is not present.

This fixes #4857


loicmathieu:
@Ladicek @jmartisk can you review the implementation ?
I'm not proud of the NoopMetricsRegistry but I don't think there is a better way.

Also the way I switch off metrics for smallrye-fault-tolerance is via a startup event listening methods, I don't know if there is a better way to do this.

@gsmet I added the noteworthy label as with this PR quarkus-smallrye-metrics will no longuer be integrated when we add quarkus-smallrye-fault-tolerance. If some people rely of it they could be surprised.

gsmet:
I'm cancelling CI for now, we need CI resources for other PRs. Sorry about that.

Please do not force push until tomorrow afternoon to avoid CI load.

Thanks!

bom/runtime/pom.xml
```diff
@@ -28,6 +28,7 @@
         <jboss-servlet-api_4.0_spec.version>1.0.0.Final</jboss-servlet-api_4.0_spec.version>
         <microprofile-config-api.version>1.3</microprofile-config-api.version>
         <microprofile-context-propagation.version>1.0.1</microprofile-context-propagation.version>
+        <microprofile-metrics.version>2.1.0</microprofile-metrics.version>
```
Ladicek: I think we probably want 2.2.0?

extensions/smallrye-fault-tolerance/deployment/src/main/java/io/quarkus/smallrye/faulttolerance/deployment/NoopMetricRegistry.java
```diff
@@ -0,0 +1,231 @@
...
+import org.eclipse.microprofile.metrics.*;
+
+import io.quarkus.runtime.StartupEvent;
+
+@ApplicationScoped
+public class NoopMetricRegistry extends MetricRegistry {
+
+    void onStart(@Observes StartupEvent ev) {
+        //the following sys prop will de-activate fault tolerance metrics collector
+        System.setProperty("MP_Fault_Tolerance_Metrics_Enabled", "false");
```
Ladicek: I don't know anything about Quarkus, but I'd expect this configuration change to be done much earlier (during build).

loicmathieu: Build is done on a separate JVM (except on dev mode), so it's not possible. But maybe this can be checked by someone with good knowledge on Quarkus configuration because I don't like neither to use a StartupEvent to set a sys prop but I didn't find an easier/more elegant way to do this.

bom/runtime/pom.xml
loicmathieu: Just saw 2.2 PR was open after this one :)
Will move to 2.2

jmartisk: I find it quite disturbing that we need to declare the same version in two different `pom.xml` files, we might probably want to introduce a property to avoid the danger of having two different versions. But that's not the point of this PR.

extensions/smallrye-fault-tolerance/deployment/src/main/java/io/quarkus/smallrye/faulttolerance/deployment/NoopMetricRegistry.java
jmartisk: Can't we use a `io.quarkus.deployment.builditem.SystemPropertyBuildItem` to do this?

loicmathieu: @jmartisk of course yes :)
I search for such build item yesterday and didn't find any ... I missed it I don't know how ...

bom/runtime/pom.xml
loicmathieu: @jmartisk now that we switch to the latests smallrye-fault-tolerance and smallrye-metrics I can remove this as both use the 2.2 version

extensions/smallrye-fault-tolerance/deployment/src/main/java/io/quarkus/smallrye/faulttolerance/deployment/NoopMetricRegistry.java
jmartisk: So glad I have IntelliJ IDEA with its super sophisticated class searching mechanism where I could write "propertybuild" and received the answer :)

bom/runtime/pom.xml
loicmathieu: done

extensions/smallrye-fault-tolerance/deployment/src/main/java/io/quarkus/smallrye/faulttolerance/deployment/NoopMetricRegistry.java
loicmathieu: > So glad I have IntelliJ IDEA with its super sophisticated class searching mechanism where I could write "propertybuild" and received the answer :)

I too use IDEA ... maybe this is not my IDE fault but ... mine ;)

gsmet:
Let's not push this in 0.28.0. Please wait.

jmartisk (approved):

extensions/smallrye-metrics/deployment/src/main/java/io/quarkus/smallrye/metrics/deployment/SmallRyeMetricsProcessor.java
```diff
@@ -202,15 +203,16 @@ AutoInjectAnnotationBuildItem autoInjectMetric() {
         return new AutoInjectAnnotationBuildItem(SmallRyeMetricsDotNames.METRIC);
     }
 
+    @BuildStep(providesCapabilities = Capabilities.METRICS)
```
jmartisk: This is now deprecated through https://github.com/quarkusio/quarkus/pull/5195/files#diff-909be471039482910981e691b9083e56R77

jmartisk (changes requested):

extensions/smallrye-metrics/deployment/src/main/java/io/quarkus/smallrye/metrics/deployment/SmallRyeMetricsProcessor.java
loicmathieu: done

loicmathieu:
@jmartisk raises concern that this implementation should fails on https://github.com/smallrye/smallrye-fault-tolerance/blob/2.1.2/implementation/src/main/java/io/smallrye/faulttolerance/metrics/BulkheadWaitRecorder.java#L55

As there is not a lot of tests inside Quarkus for fault-tolerence, I will add new ones so we will be able to decide if the implementation is OK or not.

Ladicek:
I don't think it should fail, because that code can only be reached when metrics are enabled. See https://github.com/smallrye/smallrye-fault-tolerance/blob/2.1.2/implementation/src/main/java/io/smallrye/faulttolerance/HystrixCommandInterceptor.java#L202-L209

jmartisk:
Hmm, perhaps this case will work, but we need to really check that SmallRye FT doesn't do anything that will break if we have the no-op registry trick in place, and fix any cases where it could break.. This trick is not a general purpose solution :(

Ladicek:
I don't think it's a trick. I certainly don't want to introduce metrics abstraction in SmallRye FT just to make Metrics optional. MP Metrics is that abstraction. That said, it's probably true that the SmallRye FT code was written with the assumption that an implementation of MP Metrics is always present (and we need to do better with the reimplementation). I think I checked the SmallRye FT codebase when this issue was raised and I believe this approach should work well. I'll also gladly accept any PR in SmallRye FT that would help with this.

loicmathieu:
@jmartisk @Ladicek I add tests for all the annotations of MP fault-tolerance inside quarkus-smallrye-fault-tolerance and all tests works.
As the extension tests uses the no-op impl, it means it works :)

So, is it OK for you guys to merge this ?

Meanwhile I saw that the fault tolerance guide was missing documentation on `@Bulkhead` and `@Asynchronous`. Is it worth opening an issue to covers them in the guide ?

jmartisk:
I just gave this a round of manual experimentation and it worked well for me. So +1

loicmathieu:
@jmartisk if all is OK for you can you approve it ?
@Ladicek is it OK for you also ot is there still some concern about it ?

I'll rebase the PR as there is some conflict

extensions/smallrye-fault-tolerance/deployment/src/test/java/io/quarkus/smallrye/faulttolerance/test/FaultToleranceTest.java
```diff
@@ -0,0 +1,87 @@
...
+    AsynchronousBean asynchronous;
+
+    @Test
+    public void testFallback() {
+        assertEquals(FallbackBean.RecoverFallback.class.getName(), fallback.ping());
+    }
+
+    @Test
+    public void testBulkhead() {
+        ExecutorService executorService = Executors.newFixedThreadPool(10);
```
Ladicek: Perhaps `shutdown()` the executor service at the end of the test?

Ladicek: Actually the asserts here happen on different threads, so they don't really count (unless JUnit 5 is significantly different from JUnit 4 in this regard, which I don't know). Probably easiest would be something like this (writing from memory, some tweaks might be necessary, but the general idea should be clear):

```java
AtomicInteger success = new AtomicInteger();
ExecutorService executorService = Executors.newFixedThreadPool(10);
for (int i = 0; i < 20; i++) {
    executorService.submit(() -> {
        int bulk = bulkhead.bulkhead();
        if (bulk <= 5) {
            success.incrementAndGet();
        }
    });
}
executorService.shutdown();
executorService.awaitTermination(5, TimeUnit.SECOND);
assertEquals(20, success.get());
```

extensions/smallrye-fault-tolerance/deployment/src/test/java/io/quarkus/smallrye/faulttolerance/test/circuitbreaker/CircuitBreakerBean.java
```diff
@@ -0,0 +1,20 @@
...
+@ApplicationScoped
+public class CircuitBreakerBean {
+    AtomicBoolean breakTheCircuit = new AtomicBoolean();
+
+    @CircuitBreaker(requestVolumeThreshold = 5)
+    public void breakCircuit() {
+        if (!breakTheCircuit.getAndSet(true)) {
+            return;
+        }
+        throw new RuntimeException("let's beark it !");
```
Ladicek: typo :-)

Ladicek:
Two little nitpicks in the tests, otherwise LGTM.

extensions/smallrye-fault-tolerance/deployment/src/test/java/io/quarkus/smallrye/faulttolerance/test/FaultToleranceTest.java
loicmathieu: Assertions are done on the same thread (I check the code) so my impl is correct. I assert 20 times that the bulk is less or equals to 5 and it's pretty the same as you code but with an AtomicInteger to count. I prefere mine as it carry the intent better.

I added the closing of the executorService, thanks for pointing this to me :)

extensions/smallrye-fault-tolerance/deployment/src/test/java/io/quarkus/smallrye/faulttolerance/test/circuitbreaker/CircuitBreakerBean.java
loicmathieu: fixed :)

extensions/smallrye-fault-tolerance/deployment/src/test/java/io/quarkus/smallrye/faulttolerance/test/FaultToleranceTest.java
Ladicek: Well, the `assertTrue` call is in the lambda that you pass to `executorService.submit` -- unless I'm seriously mistaken, that means the assert will happen on the thread from the executor thread pool, not on the main thread on which the test runs.

jmartisk (approved):

extensions/smallrye-fault-tolerance/deployment/src/test/java/io/quarkus/smallrye/faulttolerance/test/FaultToleranceTest.java
loicmathieu: @Ladicek you're right, the assert exception is thrown inside the task and ignored ... 

But Bulkhead also throws an exception when too many parallel queries run so the test should assert that an exception was thrown or that bulkhead was never more than five.

loicmathieu: Is the new test implementation OK ?

extensions/smallrye-fault-tolerance/deployment/src/test/java/io/quarkus/smallrye/faulttolerance/test/FaultToleranceTest.java
```diff
@@ -0,0 +1,103 @@
...
+                } catch (BulkheadException be) {
+                    bulkheadFailures.set(true);
+                }
+            });
+            Thread.sleep(1);//give some chance to the bulkhead to happens
+        }
+        executorService.shutdown();
+        executorService.awaitTermination(5, TimeUnit.SECONDS);
+        assertTrue(success.get());
+        assertTrue(bulkheadFailures.get());
```
Ladicek: Should this be `assertFalse`?

Otherwise it looks good.

loicmathieu: No, this proove that Bulkhead works by checking that a `BulkheadException` has been thrown at least once.
I hope that this test will always succeed but as it's a concurency test it can be possible to have some times false negative, maybe I should add a comment on this ?

Ladicek: Ah ah, true, true, thanks!

Ladicek (approved):

loicmathieu:
@gsmet ready to be merged and needed #5272, can we merge it ?

extensions/smallrye-fault-tolerance/deployment/src/main/java/io/quarkus/smallrye/faulttolerance/deployment/SmallRyeFaultToleranceProcessor.java
```diff
@@ -143,6 +147,12 @@ public void transform(TransformationContext context) {
                 DefaultCommandListenersProvider.class,
                 MetricsCollectorFactory.class);
         additionalBean.produce(builder.build());
+
+        if (!capabilities.isCapabilityPresent(Capabilities.METRICS)) {
+            //disable fault tolerance metrics with the MP sys props and provides a No-op metric registry.
+            additionalBean.produce(new AdditionalBeanBuildItem(NoopMetricRegistry.class));
+            systemProperty.produce(new SystemPropertyBuildItem("MP_Fault_Tolerance_Metrics_Enabled", "false"));
```
geoand: Is this property necessary? I don't see it being used anywhere

Ladicek: This is a configuration option specified by MicroProfile Fault Tolerance to switch off metrics collection. SmallRye Fault Tolerance reads it.

geoand: Got it, thanks!

geoand (approved):
Looks good to me and more importantly it's been verified by the experts

PR #5005, opened 2019-10-29
Fix broken native build of generated projects

geoand:
I honestly don't know why stacktraces need to be enabled for the generated native image to work properly.

Without this fix, the native image build of the generated (via the maven tooling) projects would fail:

1) First with a NPE
2) But even going past the NPE, the generated native image when run gave output like this:

```
Exception in thread "Thread-1" java.lang.Error: java.lang.NullPointerException
java.lang.NoClassDefFoundError: Could not initialize class java.util.concurrent.ThreadLocalRandom
java.lang.NoClassDefFoundError: Could not initialize class java.util.concurrent.ThreadLocalRandom
2019-10-29 20:17:38,761 ERROR [io.ver.cor.imp.ContextImpl] (vert.x-eventloop-thread-0) Unhandled exception: java.lang.NullPointerException: channel

Caused by: java.lang.NoClassDefFoundError: Could not initialize class java.util.concurrent.ThreadLocalRandom
```

### DIFF
--- a/core/deployment/src/main/java/io/quarkus/deployment/pkg/NativeConfig.java
+++ b/core/deployment/src/main/java/io/quarkus/deployment/pkg/NativeConfig.java
@@ -165,7 +165,7 @@ public class NativeConfig {
     /**
      * If full stack traces are enabled in the resulting image
      */
-    @ConfigItem(defaultValue = "false")
+    @ConfigItem(defaultValue = "true")
     public boolean fullStackTraces;
 
     /**
--- a/devtools/maven/src/main/java/io/quarkus/maven/BuildMojo.java
+++ b/devtools/maven/src/main/java/io/quarkus/maven/BuildMojo.java
@@ -187,11 +187,13 @@ public class BuildMojo extends AbstractMojo {
                             .setConfigDir(outputDirectory.toPath())
                             .setBuildSystemProperties(realProperties).build());
             Artifact original = project.getArtifact();
-            if (result.getJar().isUberJar() && result.getJar().getOriginalArtifact() != null) {
-                original.setFile(result.getJar().getOriginalArtifact().toFile());
-            }
-            if (result.getJar().isUberJar()) {
-                projectHelper.attachArtifact(project, result.getJar().getPath().toFile(), "runner");
+            if (result.getJar() != null) {
+                if (result.getJar().isUberJar() && result.getJar().getOriginalArtifact() != null) {
+                    original.setFile(result.getJar().getOriginalArtifact().toFile());
+                }
+                if (result.getJar().isUberJar()) {
+                    projectHelper.attachArtifact(project, result.getJar().getPath().toFile(), "runner");
+                }
             }
 
         } catch (AppCreatorException e) {


geoand:
@gsmet can you try it out? Asking because I know you could reproduce the issue

gsmet:
I sure will.

gsmet:
One thing we need is to move one of the current IT to the new way of doing things. It’s not acceptable to discover this by pure luck.

dmlloyd:
My hypothesis is that this fails because the class is reinitialized, but the static init does some things that are disallowed during native image run time.

I think this class cannot be reinitialized at run time.  Instead I think we'll have to reinitialize specific fields manually.

geoand:
> One thing we need is to move one of the current IT to the new way of doing things. It’s not acceptable to discover this by pure luck.

I can do that in a follow up

core/deployment/src/main/java/io/quarkus/deployment/pkg/NativeConfig.java
```diff
@@ -165,7 +165,7 @@
     /**
      * If full stack traces are enabled in the resulting image
      */
-    @ConfigItem(defaultValue = "false")
+    @ConfigItem(defaultValue = "true")
```
gsmet: Yes, this is definitely dark magic.

devtools/maven/src/main/java/io/quarkus/maven/BuildMojo.java
```diff
@@ -187,11 +187,13 @@ public void execute() throws MojoExecutionException {
                             .setConfigDir(outputDirectory.toPath())
                             .setBuildSystemProperties(realProperties).build());
             Artifact original = project.getArtifact();
-            if (result.getJar().isUberJar() && result.getJar().getOriginalArtifact() != null) {
-                original.setFile(result.getJar().getOriginalArtifact().toFile());
-            }
-            if (result.getJar().isUberJar()) {
-                projectHelper.attachArtifact(project, result.getJar().getPath().toFile(), "runner");
+            if (result.getJar() != null) {
```
gsmet: I'm not sure if the issue is the null test or the fact that `result.getJar()` is empty. When using XML config, it's not as we don't have the NPE so I think we need Stuart to say if it's normal or not.

core/deployment/src/main/java/io/quarkus/deployment/pkg/NativeConfig.java
```diff
@@ -165,7 +165,7 @@
...
     public boolean fullStackTraces;
```
gsmet: What bugs me a bit too is that now, the output when creating the native image is very scarce. I don't know if it's normal or not but we used to have a couple of Quarkus lines and now we have:
```
21:02:57,864 INFO  [org.jbo.threads] JBoss Threads version 3.0.0.Final
```
and that's it.

gsmet (commented):
So I confirm it fixes the issue for me too and with a newly created simple project, I'm now able to create a native image.

I think we need Stuart to confirm if that's the right fix though.

For Stuart:
- create a project with `mvn io.quarkus:quarkus-maven-plugin:999-SNAPSHOT:create     -DprojectGroupId=org.acme     -DprojectArtifactId=getting-started     -DclassName="org.acme.quickstart.GreetingResource"     -Dpath="/hello" -Dextensions="openapi"`
- `cd getting-started`
- `mvn clean install -Dnative` will fail with a NPE in BuildMojo at line 190
- with the `if` Georgios added, we pass this step but then we have another NPE that the change of default value for the full stacktrace somehow fixes.

core/deployment/src/main/java/io/quarkus/deployment/pkg/NativeConfig.java
stuartwdouglas: I will look into this

devtools/maven/src/main/java/io/quarkus/maven/BuildMojo.java
stuartwdouglas: This is ok, the NativeImageMojo does not attempt to look at the jar result.

stuartwdouglas (approved):

core/deployment/src/main/java/io/quarkus/deployment/pkg/NativeConfig.java
stuartwdouglas: This change was unintentional. I must have made a mistake when creating this, or maybe it got missed in one of the many rebases.

PR #5012, opened 2019-10-30
Changed Gradle project templates to use the platform BOM properties, …

aloubyansky:
…made Gradle list and add extensions commands to use the correct platform descriptor, enhanced JSON descriptor resolver to determine the JSON artifact corresponding to the specified BOM coords, enhanced Gradle AppModelResolver impl, fixes related to resolution of the latest platform version

@maxandersen @gastaldi I think this PR is good enough for Gradle support. I tested it with the custom built universe which includes Camel too. Specifically, it's about list and add extensions.

QuarkusJsonPlatformDescriptorResolver class is probably going to be used in a similar manner in the Maven case. Except that I'll still need to fallback to the classpath discovery of resources in case we are offline.

### DIFF
--- a/devtools/gradle/src/main/java/io/quarkus/gradle/AppModelGradleResolver.java
+++ b/devtools/gradle/src/main/java/io/quarkus/gradle/AppModelGradleResolver.java
@@ -13,6 +13,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Properties;
+import java.util.Set;
 
 import org.gradle.api.GradleException;
 import org.gradle.api.Project;
@@ -49,6 +50,11 @@ public class AppModelGradleResolver implements AppModelResolver {
     }
 
     @Override
+    public String getLatestVersionFromRange(AppArtifact appArtifact, String range) throws AppModelResolverException {
+        throw new UnsupportedOperationException();
+    }
+
+    @Override
     public String getNextVersion(AppArtifact arg0, String fromVersion, boolean fromVersionIncluded, String arg1, boolean arg2)
             throws AppModelResolverException {
         throw new UnsupportedOperationException();
@@ -67,7 +73,31 @@ public class AppModelGradleResolver implements AppModelResolver {
     @Override
     public Path resolve(AppArtifact appArtifact) throws AppModelResolverException {
         if (!appArtifact.isResolved()) {
-            throw new AppModelResolverException("Artifact has not been resolved: " + appArtifact);
+
+            final GradleDependencyArtifact dep = new GradleDependencyArtifact();
+            dep.setExtension(appArtifact.getType());
+            dep.setType(appArtifact.getType());
+            dep.setName(appArtifact.getArtifactId());
+
+            final QuarkusExtDependency gradleDep = new QuarkusExtDependency(appArtifact.getGroupId(),
+                    appArtifact.getArtifactId(), appArtifact.getVersion(), null);
+            gradleDep.addArtifact(dep);
+
+            final Configuration detachedConfig = project.getConfigurations().detachedConfiguration(gradleDep);
+
+            final ResolvedConfiguration rc = detachedConfig.getResolvedConfiguration();
+            Set<ResolvedArtifact> resolvedArtifacts = rc.getResolvedArtifacts();
+            for (ResolvedArtifact a : resolvedArtifacts) {
+                if (appArtifact.getArtifactId().equals(a.getName())
+                        && appArtifact.getType().equals(a.getType())
+                        && appArtifact.getGroupId().equals(a.getModuleVersion().getId().getGroup())) {
+                    appArtifact.setPath(a.getFile().toPath());
+                }
+            }
+
+            if (!appArtifact.isResolved()) {
+                throw new AppModelResolverException("Failed to resolve " + appArtifact);
+            }
         }
         return appArtifact.getPath();
     }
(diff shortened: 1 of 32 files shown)


aloubyansky:
While the failed tests are not currently relevant and could be ignored, I'm planning to review and fix them.

aloubyansky:
One of the maven projection generating tests will fail because I thought removing managed dependencies from the descriptor was a good idea. I'll fix it a bit later.

devtools/gradle/src/main/java/io/quarkus/gradle/AppModelGradleResolver.java
```diff
@@ -166,6 +185,10 @@ private ModuleIdentifier getModuleId(ResolvedArtifact a) {
 
     private AppDependency toAppDependency(ResolvedArtifact a) {
         final String[] split = a.getModuleVersion().toString().split(":");
+        System.out.println(a.getModuleVersion());
```
maxandersen: no debug logger in gradle ? :)

devtools/gradle/src/main/java/io/quarkus/gradle/GradleDependencyArtifact.java
```diff
@@ -0,0 +1,117 @@
+package io.quarkus.gradle;
+
+import org.gradle.api.artifacts.DependencyArtifact;
+
+public class GradleDependencyArtifact implements DependencyArtifact {
```
maxandersen: why the need for a different one ?

devtools/gradle/src/main/resources/quarkus-platform.properties
```diff
@@ -0,0 +1,3 @@
+quarkus.platform.json.groupId=io.quarkus
```
maxandersen: what is this file representing ?

note - we use kebab-case (i.e. group-id) every where else in properties, json and yaml. Suggest we do the same here.

devtools/gradle/src/main/java/io/quarkus/gradle/AppModelGradleResolver.java
aloubyansky: Removed, thanks.

devtools/gradle/src/main/java/io/quarkus/gradle/GradleDependencyArtifact.java
aloubyansky: What's the alternative?

devtools/gradle/src/main/resources/quarkus-platform.properties
aloubyansky: It's actually not used anymore. I was thinking about externalizing the default platform coordinates for the tools this way instead of hardcoding values in Java classes.
groupId and artifactId is what's common in the Maven world. I don't think you'll find group-id and artifact-id in maven-related configs (be it XML or something else).

devtools/gradle/src/main/java/io/quarkus/gradle/GradleDependencyArtifact.java
maxandersen: ah, i read this as extending Dependency not implementing.

devtools/gradle/src/main/resources/quarkus-platform.properties
maxandersen: not correct, thats what we have in our and even spring similar properties/yaml/json.

aloubyansky:
ConnectionToReplicaSetTest>MongoWithReplicasTestBase.startMongoDatabase:56->MongoWithReplicasTestBase.initializeReplicaSet:90 » MongoTimeout

Isn't related.

devtools/gradle/src/main/java/io/quarkus/gradle/tasks/QuarkusPlatformTask.java
```diff
@@ -0,0 +1,60 @@
...
+            try (InputStream is = Files.newInputStream(gradlePropsPath)) {
+                props.load(is);
+            } catch (IOException e) {
+                throw new IllegalStateException("Failed to load " + gradlePropsPath, e);
+            }
+
+            final QuarkusPlatformDescriptor platform = QuarkusJsonPlatformDescriptorResolver.newInstance()
+                    .setArtifactResolver(extension().resolveAppModel())
+                    .setBomVersion(
+                            getRequiredProperty(props, "quarkusPlatformBomGroupId"),
```
maxandersen: these shuld be something like `quarkus.platform.bom.group-id` to match naming style use elsewhere.

i.e. quarkus.version as used in mvn.

maybe just quarkus.platform.group-id to make it shorter; the bom is implied anyways.

maxandersen (approved):
overall looks good; i think we should cleanup the property names/references but we can do that in a follow up PR as it is following yet another format in the mvn generated projects.

devtools/gradle/src/main/java/io/quarkus/gradle/tasks/QuarkusPlatformTask.java
aloubyansky: No objections changing the style. We've been using quarkusVersion in gradle.properties, so I simply followed that.
As to `quarkus.platform.group-id` w/o the `bom` in it, possibly ok, however we might want to support `quarkus.platform.json.group-id` at some point. QuarkusJsonPlatformDescriptorResolver for example already includes such a property.

aloubyansky: I'm gonna merge it in this state though. Adjusting the style could be done as a separate PR.

gsmet: I think camel case is the usual Gradle style. At least I saw that used a lot in Gradle files.

PR #5034, opened 2019-10-30
Fixes #5018 and enables last MP REST TCK test

### LINKED ISSUE #5018: ScopeInfo should have a public constructor
Otherwise it can't be used for custom scopes in a `BeanConfigurator`. We should also add some javadoc to instruct the users to use `io.quarkus.arc.processor.BuiltinScope` for built-in scopes.

kenfinnigan:


extensions/rest-client/deployment/src/main/java/io/quarkus/restclient/deployment/RestClientProcessor.java
```diff
@@ -258,33 +261,46 @@ private String computeConfigPrefix(String interfaceName, ClassInfo classInfo) {
...
+
         if (scopeConfig.isPresent()) {
             final DotName scope = DotName.createSimple(scopeConfig.get());
             final BuiltinScope builtinScope = BuiltinScope.from(scope);
             if (builtinScope != null) { // override default @Dependent scope with user defined one.
-                scopeInfo = builtinScope.getInfo();
-            } else {
+                scopeToUse = builtinScope.getInfo();
+            } else if (capabilities.isCapabilityPresent(Capabilities.SERVLET)) {
+                if (scope.toString().equals("javax.enterprise.context.SessionScoped")) {
```
mkouba: You should probably use some `DotName` constant or `SessionScoped.class.getName()`?

mkouba (changes requested):
Otherwise looks good...

extensions/rest-client/deployment/src/main/java/io/quarkus/restclient/deployment/RestClientProcessor.java
kenfinnigan: Ok, wasn't sure about that

Will fix

mkouba:
One more thing - could we add a constructor to the `ScopeInfo` with a `DotName` param instead of `Class`? ;-)

kenfinnigan:
And then use that for SessionScoped creation?

kenfinnigan:
@mkouba in the constructor, how do we know if it's @Inherited from the DotName?

We need to retrieve the class from somewhere?

kenfinnigan:
How's that?

mkouba (approved):
Looks good.

PR #5042, opened 2019-10-30
Updating to quarkus-vertx-core and other dep optimizations

### LINKED ISSUE #5040: quarkus-oidc can depend on vertx-core and vertx-http as opposed to the full vertx stack
**Description**
See the subject, as recommended by Clement @cescoffier  

**Implementation ideas**
Have the adapter depending on `quarkus-vertx-core`

sberyozkin:
Fixes #5040

### DIFF
--- a/extensions/oidc/deployment/src/main/java/io/quarkus/oidc/deployment/OidcBuildStep.java
+++ b/extensions/oidc/deployment/src/main/java/io/quarkus/oidc/deployment/OidcBuildStep.java
@@ -16,11 +16,12 @@ import io.quarkus.oidc.runtime.OidcConfig;
 import io.quarkus.oidc.runtime.OidcIdentityProvider;
 import io.quarkus.oidc.runtime.OidcJsonWebTokenProducer;
 import io.quarkus.oidc.runtime.OidcRecorder;
-import io.quarkus.vertx.deployment.VertxBuildItem;
+import io.quarkus.vertx.core.deployment.InternalWebVertxBuildItem;
 import io.smallrye.jwt.auth.cdi.CommonJwtProducer;
 import io.smallrye.jwt.auth.cdi.JsonValueProducer;
 import io.smallrye.jwt.auth.cdi.RawClaimTypeProducer;
 
+@SuppressWarnings("deprecation")
 public class OidcBuildStep {
 
     @BuildStep
@@ -65,7 +66,7 @@ public class OidcBuildStep {
 
     @Record(ExecutionTime.RUNTIME_INIT)
     @BuildStep
-    public void setup(OidcConfig config, OidcRecorder recorder, VertxBuildItem vertxBuildItem,
+    public void setup(OidcConfig config, OidcRecorder recorder, InternalWebVertxBuildItem vertxBuildItem,
             BeanContainerBuildItem bc) {
         if (config.enabled) {
             recorder.setup(config, vertxBuildItem.getVertx(), bc.getValue());
--- a/bom/deployment/pom.xml
+++ b/bom/deployment/pom.xml
@@ -383,6 +383,11 @@
             </dependency>
             <dependency>
                 <groupId>io.quarkus</groupId>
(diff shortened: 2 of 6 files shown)


stuartwdouglas (approved):

extensions/oidc/deployment/src/main/java/io/quarkus/oidc/deployment/OidcBuildStep.java
```diff
@@ -46,7 +46,7 @@ EnableAllSecurityServicesBuildItem security() {
 
     @Record(ExecutionTime.RUNTIME_INIT)
     @BuildStep
-    public void setup(OidcConfig config, OidcRecorder recorder, VertxBuildItem vertxBuildItem,
+    public void setup(OidcConfig config, OidcRecorder recorder, CoreVertxBuildItem vertxBuildItem,
```
stuartwdouglas: Actually this should be using io.quarkus.vertx.core.deployment.InternalWebVertxBuildItem for now. We want OIDC to be on the same vert.x instance as HTTP.

stuartwdouglas (changes requested):

extensions/oidc/deployment/src/main/java/io/quarkus/oidc/deployment/OidcBuildStep.java
sberyozkin: @stuartwdouglas It is deprecated so I used a qualified parameter name to avoid suppressing the deprecation warnings everywhere :-)

stuartwdouglas: It's deprecated because the intention was to remove it for 1.0, but it did not make it in. This is the correct one to use here though. We want to get rid of it, but we just did not have time.

sberyozkin:
@stuartwdouglas @cescoffier I'm assuming this PR should remains in the Draft form for another few days; let me know please if you'd like it to go in earlier (thinking about Stuart's comment about the same Vertx instance here)

extensions/oidc/deployment/src/main/java/io/quarkus/oidc/deployment/OidcBuildStep.java
sberyozkin: OK

extensions/oidc/deployment/src/main/java/io/quarkus/oidc/deployment/OidcBuildStep.java
```diff
@@ -44,9 +43,11 @@ EnableAllSecurityServicesBuildItem security() {
         return new EnableAllSecurityServicesBuildItem();
     }
 
+    @SuppressWarnings("deprecation")
     @Record(ExecutionTime.RUNTIME_INIT)
     @BuildStep
-    public void setup(OidcConfig config, OidcRecorder recorder, VertxBuildItem vertxBuildItem,
+    public void setup(OidcConfig config, OidcRecorder recorder,
+            io.quarkus.vertx.core.deployment.InternalWebVertxBuildItem vertxBuildItem,
```
gsmet: It's minor but could we avoid using the FQCN here and just adding an import?

gsmet (changes requested):
@sberyozkin I added a small comment, could you take care of it and promote your PR? I think we can merge it now.

extensions/oidc/deployment/src/main/java/io/quarkus/oidc/deployment/OidcBuildStep.java
sberyozkin: @gsmet sure, I did it earlier to avoid multiple deprecation warnings

sberyozkin:
@gsmet @stuartwdouglas I had to add a `quarkus-jackson` test dependency to `quarkus-oidc/runtime`, I'm pretty sure it was building OK for me locally when I was only opening a PR, but I guess I did not rebase my branch at a time. All looks good now

stuartwdouglas (approved):

PR #5044, opened 2019-10-30
Generate documentation file for config group

### LINKED ISSUE #5097: generated docs are sometimes overridden when running full build
**Describe the bug**
When running a full build, we sometimes override a previous part of a generated doc

**Expected behavior**
We should merge extensions doc and not overriding a partial or fully generated one

**Actual behavior**
Docs are overridden - e.g when running a full build, the `quarkus-vertx-http.adoc` generated doc does not contain all the information about http configuration. 

But when build the `vertx-http` extension alone, the file in question contains all the information.

 - Quarkus version or git rev: problem observed in latest master 9ed64388726143c99e61db3f966373ae6dcfab88

**Additional context**
I'll come up with a fix, still investigating on ways to work around this.

machi1990:
**1st commit:**
Make it possible to generate a file par config group. 

**2nd commit:** 
use the generated file for `FormAuthConfig` to avoid the usage of an inexistant file which  was generating the following warning
```
asciidoctor: INFO: security-guide.adoc: line 69: optional include dropped because include file not found: target/asciidoc/generated/config/quarkus-http-auth-form.adoc 
```

**3rd commit:**
Resolves https://github.com/quarkusio/quarkus/issues/5097

### DIFF
--- a/core/processor/src/main/java/io/quarkus/annotation/processor/Constants.java
+++ b/core/processor/src/main/java/io/quarkus/annotation/processor/Constants.java
@@ -54,8 +54,18 @@ final public class Constants {
             .get(System.getProperties().getProperty("maven.multiModuleProjectDirectory")
                     + Constants.DOCS_SRC_MAIN_ASCIIDOC_GENERATED);
     public static final File GENERATED_DOCS_DIR = GENERATED_DOCS_PATH.toFile();
+
+    /**
+     * Holds the list of configuration items / configuration sections of each configuration roots.
+     */
     public static final File ALL_CR_GENERATED_DOC = GENERATED_DOCS_PATH
             .resolve("all-configuration-roots-generated-doc.properties").toFile();
+
+    /**
+     * Holds the list of computed file names and the list of configuration roots of this extension
+     */
+    public static final File EXTENSION_CONFIGURATION_ROOT_LIST = GENERATED_DOCS_PATH
+            .resolve("extensions-configuration-roots-list.properties").toFile();
 
     public static final String DURATION_NOTE_ANCHOR = "duration-note-anchor";
     public static final String MEMORY_SIZE_NOTE_ANCHOR = "memory-size-note-anchor";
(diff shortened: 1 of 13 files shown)


docs/src/main/asciidoc/security-guide.adoc
```diff
@@ -64,10 +64,6 @@ value. This cookie contains a expiry time as part of the encrypted value, so all
 clocks synchronised. At one minute intervals a new cookie will be generated with an updated expiry time if the session
 is in use.
 
-The following properties can be used to configure form based auth:
```
machi1990: Should we add a link to the form-auth section of all-doc-page?

gsmet: These properties are not generated in a smaller file?

machi1990: IIRC, they are not generated in a smaller file.

gsmet: Hmm, I don't understand. I remember having all of them in small files. Is it gone?

gsmet: Would it be possible to generate a file per config group too? Looks like we have quite a lot of docs where it makes sense.

machi1990: +1, I was leaning toward that path. I'll look at it.

FroMage: Why is it inexistant? I reviewed that guide last week and didn't see any inexistant file?

machi1990: The `FormAuthConfig` configuration is a config group, and we do not generate specific doc files for config groups. If it was previously there, maybe the file being referenced here was for a config root that is no longer there.

FroMage: Ah, that's possible.

machi1990:
@gsmet I have added a commit which enable the generation of config group as discussed above.

gsmet (approved):
I checked the output and everything looks in order. We can tune it later if needed.

PR #5045, opened 2019-10-30
Fix Security extension metadata

emmanuelbernard:


gsmet:
Ken is right, he had to fix the Jaeger one as it used to be the security one.

I marked this issue with 0.28.0.

gsmet (approved):

Issue #4516, opened 2019-10-11
MP Config TCK - Fix excluded tests


kenfinnigan:
Currently, the below tests are excluded from execution when running the MP Config TCK. We need to either fix Quarkus so the test passes or challenge and fix the TCK if it's a problem there.

- [x] [ArrayConverterTest.*](https://github.com/quarkusio/quarkus/blob/master/tcks/microprofile-config/tck-suite.xml#L17)
- [ ] [ConfigProviderTest.testInjectedConfigSerializable](https://github.com/quarkusio/quarkus/blob/master/tcks/microprofile-config/tck-suite.xml#L26)
- [x] [ConfigProviderTest.testPropertyConfigSource](https://github.com/quarkusio/quarkus/blob/master/tcks/microprofile-config/tck-suite.xml#L28)
- [x] [ConfigProviderTest.testEnvironmentConfigSource](https://github.com/quarkusio/quarkus/blob/master/tcks/microprofile-config/tck-suite.xml#L36)

kenfinnigan:
@dmlloyd @jmesnil I was taking a look at `ConfigProviderTest.testPropertyConfigSource` this afternoon and I'm not sure if the behavior is a problem with us or the TCK.

Basically the test iterates over system properties and verifies that the values from `System.getProperties()` are the same as `config.getValue()` retrieved from the system property config source.

What's weird is that the value for `env.MAVEN_CMD_LINE_ARGS` contains a leading space in the value when coming from the ConfigSource but doesn't from `System.getProperties()` in the test class, even though the ConfigSource uses `System.getProperty()` inside.

Any thoughts?

dmlloyd:
There's an actual property called `env.MAVEN_CMD_LINE_ARGS`?  If so, we're not setting that AFAIK; Maven must be.  And perhaps they're inserting a leading space then?

kenfinnigan:
I believe it's set by Maven.

I'm just confused how accessing it directly through system properties, and through system properties in the config source, lead to one having a leading space and one not

dmlloyd:
Is there a log output I could look at without having to build everything?

kenfinnigan:
Will run the test and capture it for you

kenfinnigan:
@dmlloyd here's the output from running the single test method: https://gist.github.com/kenfinnigan/ede49a21fd7f144f7d58c65e29b295cc

dmlloyd:
[...]

PR #5050, opened 2019-10-30
Fix MP Config TCK tests as part of #4516

kenfinnigan:


dmlloyd (approved):

core/runtime/src/main/java/io/quarkus/runtime/configuration/ExpandingConfigSource.java
```diff
@@ -72,7 +72,7 @@ public static String expandValue(String value, Cache cache) {
         if (value == null)
             return null;
         final Expression compiled = cache.exprCache.computeIfAbsent(value,
-                str -> Expression.compile(str, Expression.Flag.LENIENT_SYNTAX));
+                str -> Expression.compile(str, Expression.Flag.LENIENT_SYNTAX, Expression.Flag.NO_TRIM));
```
gsmet: Hum. Can't this potentially break a lot of applications? E.g. any application having a space at the end of a config line?

I'm not sure I would do it that last minute just to have some more TCK tests passing.

gsmet (dismissed):
I added what seems to me an important question. Let's discuss it before merging please.

core/runtime/src/main/java/io/quarkus/runtime/configuration/ExpandingConfigSource.java
kenfinnigan: Possibly, but if we want the change isn't it better to have it for 1.0 rather than after? As then it's even more of a breaking change

The question is whether we expect users to want leading/trailing spaces in configuration values. If they don't, or we don't think it makes sense, then we should include this change.

If we'd rather allow spaces and trim them, we can look to modify the TCK test to ignore leading/trailing spaces.

gsmet: I'm a bit torn on this. There might be a few specific config property where you would like spaces to be kept (a separator for instance) but for most of them, you don't really care and it might be a real pain.

I think it's worth a discussion on the list rather than sneaking it in like that.

manovotn (approved):

core/runtime/src/main/java/io/quarkus/runtime/configuration/ExpandingConfigSource.java
dmlloyd: No this won't affect most cases.  If anything it removes a behavioral breakage.  And even if it did affect someone somehow, the error message would be very obvious: unlike the opposite case you'd get when you expect to preserve whitespace and it doesn't happen, which will be subtle and hard to detect.

jaikiran:
Looks like we might have our first issue reported related to this https://github.com/quarkusio/quarkus/issues/5305

PR #5053, opened 2019-10-30
Make @Claim injection work when using OIDC

stuartwdouglas:


extensions/oidc/deployment/src/main/java/io/quarkus/oidc/deployment/OidcBuildStep.java
```diff
@@ -22,6 +28,19 @@ FeatureBuildItem featureBuildItem() {
         return new FeatureBuildItem(FeatureBuildItem.OIDC);
     }
 
+    @BuildStep
```
kenfinnigan: Is using capabilities to do this preferable to a common extension that they both depend on?

stuartwdouglas: Adding a new module for 3 lines of code seems like overkill.

kenfinnigan: It's mostly about whether it's intuitive for those coming along later that if there are additional beans that need to be added for JWT in the SmallRye JWT extension, that it also needs to be added to this code block

manovotn (approved):

extensions/oidc/deployment/src/main/java/io/quarkus/oidc/deployment/OidcBuildStep.java
sberyozkin: @kenfinnigan Hi Ken it is a good point. We have an issue to link the OIDC with the `quarkus-smallrye-jwt` (in which case this block would be redundant), but I'm not sure yet if a link between these 2 modules is needed or not, and may be just having `smalrye-jwt` being the shared library is the right way to go with some small duplication as in this build step.

sberyozkin (approved):
@stuartwdouglas FYI, there is CDI code directly in `quarkus-smallrye-jwt` required to have some more MP JWT injection working - but I'm going to work on abstracting as much as possible from the code to `smallrye-jwt`

PR #5054, opened 2019-10-30
Fixes related to resolving Quarkus platform descriptor versions

aloubyansky:
… for a project, for a BOM, for the JSON coordinates, including the version ranges


maxandersen:
on first look this  looks good but i'm a bit hazy wether this completely unifies lookup from both mvn and gradle ?

i'll try run some tests tomorrow morning - too late here now ;)

aloubyansky:
> on first look this looks good but i'm a bit hazy wether this completely unifies lookup from both mvn and gradle ?

It kind of does.

maxandersen (approved):
let get this in. the sooner the better. if issues found around resolving we can still fall back on bundled.

PR #5056, opened 2019-10-30
CORS headers may not be sent on 403

stuartwdouglas:
The authorization and CORS handlers had the same priority,
which means sometimes CORS headers would not be sent on a
403 response.


gastaldi (approved):

gsmet (approved):

extensions/vertx-http/deployment/src/main/java/io/quarkus/vertx/http/deployment/FilterBuildItem.java
```diff
@@ -11,6 +11,11 @@
  */
 public final class FilterBuildItem extends MultiBuildItem {
 
+    //predefined system priorities
+    public static final int AUTHENTICATION = 200;
+    public static final int CORS = 150;
```
gastaldi: CORS should take precedence over AUTHENTICATION and AUTHORIZATION since the credentials may be wrong

Issue #5058, opened 2019-10-31
Include Throwable in startup error log message


brunoabdon:
**Describe the bug**
The error log message "_Failed to start Quarkus, attempting to start hot replacement endpoint to recover_" should include the Throwable that caused it.

This is at this line: https://github.com/quarkusio/quarkus/blob/6ac42657d5c3cc95cf8a09feb8a1f52ebc80b4f7/core/devmode/src/main/java/io/quarkus/dev/DevModeMain.java#L187

**Expected behavior**
Show the Throwable info on error log.

**Actual behavior**
Throwable instance is swallowed.

**Suggestion**
Just pass `t`as second parameter:

`log.error("Failed to start Quarkus, attempting to start hot replacement endpoint to recover",t);`

jaikiran:
https://github.com/quarkusio/quarkus/pull/5061

Sanne:
I think we can consider this resolved by #5061 : closing

PR #5061, opened 2019-10-31
Include the original exception that caused Quarkus startup to fail

jaikiran:
The commit here splits a log message into 2 separate messages, one to include (at `ERROR` level) the original exception that caused Quarkus startup to fail and another at `INFO` level to mention that hot replacement endpoint will be setup to handle recovery.

Should help with https://github.com/quarkusio/quarkus/issues/5058


gsmet (approved):
Nice, thanks!

geoand:
Just now saw this one, :+1: , much better developer experience

PR #5074, opened 2019-10-31
CDI security determined at build time

### LINKED ISSUE #4992: CDI security: Unannotated inherited methods may be restricted when they should not
**Describe the bug**
For the case in the $subject, `determineSecurityAnnotationFromBindings#determineSecurityAnnotationFromBindings` returns annotations from the subclass even if the method is not overriden

**Expected behavior**
If we want to be consistent with JSR 250 section 2.1 point 4, constraint from the subclass should not be applied here.

**To Reproduce**
 Run `CDIAccessDefaultTest.shouldAccessInheritedAllowedMethod` separately

michalszynkiewicz:
Fixes #4992

### DIFF
--- a/build-parent/pom.xml
+++ b/build-parent/pom.xml
@@ -27,6 +27,8 @@
         <scala.version>2.12.8</scala.version>
         <scala-plugin.version>4.1.1</scala-plugin.version>
 
+        <version.surefire.plugin>2.22.1</version.surefire.plugin>
+
         <!-- These properties are needed in order for them to be resolvable by the documentation -->
         <!-- The Graal version we suggest using in documentation - as that's
            what we work with by self downloading it: -->
--- a/extensions/security/deployment/src/main/java/io/quarkus/security/deployment/SecurityProcessor.java
+++ b/extensions/security/deployment/src/main/java/io/quarkus/security/deployment/SecurityProcessor.java
@@ -32,8 +32,9 @@ import io.quarkus.arc.processor.BuiltinScope;
 import io.quarkus.deployment.Capabilities;
 import io.quarkus.deployment.annotations.BuildProducer;
 import io.quarkus.deployment.annotations.BuildStep;
-import io.quarkus.deployment.builditem.CapabilityBuildItem;
+import io.quarkus.deployment.builditem.ApplicationClassPredicateBuildItem;
 import io.quarkus.deployment.builditem.ApplicationIndexBuildItem;
+import io.quarkus.deployment.builditem.CapabilityBuildItem;
 import io.quarkus.deployment.builditem.FeatureBuildItem;
 import io.quarkus.deployment.builditem.nativeimage.ReflectiveClassBuildItem;
 import io.quarkus.gizmo.MethodCreator;
@@ -110,7 +111,9 @@ public class SecurityProcessor {
 
     @BuildStep
     void gatherSecurityChecks(BuildProducer<BeanRegistrarBuildItem> beanRegistrars,
-            ApplicationIndexBuildItem indexBuildItem) {
+            ApplicationIndexBuildItem indexBuildItem,
+            BuildProducer<ApplicationClassPredicateBuildItem> classPredicate) {
+        classPredicate.produce(new ApplicationClassPredicateBuildItem(new SecurityCheckStorage.AppPredicate()));
 
         beanRegistrars.produce(new BeanRegistrarBuildItem(new BeanRegistrar() {
 
--- a/extensions/security/deployment/src/main/java/io/quarkus/security/deployment/SecurityProcessor.java
+++ b/extensions/security/deployment/src/main/java/io/quarkus/security/deployment/SecurityProcessor.java
@@ -1,23 +1,44 @@
 package io.quarkus.security.deployment;
 
+import java.lang.reflect.Method;
 import java.security.Provider;
 import java.security.Security;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 
+import org.jboss.jandex.AnnotationInstance;
+import org.jboss.jandex.AnnotationValue;
+import org.jboss.jandex.ClassInfo;
+import org.jboss.jandex.DotName;
+import org.jboss.jandex.MethodInfo;
+import org.jboss.jandex.Type;
 import org.jboss.logging.Logger;
 
 import io.quarkus.arc.deployment.AdditionalBeanBuildItem;
 import io.quarkus.arc.deployment.AnnotationsTransformerBuildItem;
+import io.quarkus.arc.deployment.BeanRegistrarBuildItem;
 import io.quarkus.arc.deployment.InterceptorBindingRegistrarBuildItem;
+import io.quarkus.arc.processor.AnnotationStore;
+import io.quarkus.arc.processor.BeanConfigurator;
+import io.quarkus.arc.processor.BeanRegistrar;
+import io.quarkus.arc.processor.BuildExtension;
+import io.quarkus.arc.processor.BuiltinScope;
 import io.quarkus.deployment.Capabilities;
 import io.quarkus.deployment.annotations.BuildProducer;
 import io.quarkus.deployment.annotations.BuildStep;
 import io.quarkus.deployment.builditem.CapabilityBuildItem;
+import io.quarkus.deployment.builditem.ApplicationIndexBuildItem;
 import io.quarkus.deployment.builditem.FeatureBuildItem;
 import io.quarkus.deployment.builditem.nativeimage.ReflectiveClassBuildItem;
+import io.quarkus.gizmo.MethodCreator;
+import io.quarkus.gizmo.MethodDescriptor;
+import io.quarkus.gizmo.ResultHandle;
 import io.quarkus.security.runtime.IdentityProviderManagerCreator;
 import io.quarkus.security.runtime.SecurityBuildTimeConfig;
 import io.quarkus.security.runtime.SecurityIdentityAssociation;
@@ -26,6 +47,8 @@ import io.quarkus.security.runtime.interceptor.AuthenticatedInterceptor;
 import io.quarkus.security.runtime.interceptor.DenyAllInterceptor;
 import io.quarkus.security.runtime.interceptor.PermitAllInterceptor;
 import io.quarkus.security.runtime.interceptor.RolesAllowedInterceptor;
+import io.quarkus.security.runtime.interceptor.SecurityCheckStorage;
+import io.quarkus.security.runtime.interceptor.SecurityCheckStorageBuilder;
 import io.quarkus.security.runtime.interceptor.SecurityConstrainer;
 import io.quarkus.security.runtime.interceptor.SecurityHandler;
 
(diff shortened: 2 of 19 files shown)


michalszynkiewicz:
@stuartwdouglas do we want it now?

gsmet:
Marking it for 0.28.0 so that we don't forget to decide what to do with this PR but I will let @stuartwdouglas  decide if it should go in or not.

gsmet:
Rebased and force-pushed to get the higher timeout for TCK tests I pushed to master.

michalszynkiewicz:
thanks @gsmet !

michalszynkiewicz:
There is a test failure in microprofile-jwt tck. The failure seems to be due to the fact that role mappings are not applied.
What's strange is that the same test is in the smallrye-jwt/deployment module, it's called `RolesAllowedUnitTest` and it passes.
The most outstanding difference is arquillian vs `QuarkusUnitTest`.

michalszynkiewicz:
For the successful test (in smallrye-jwt), the mapped "role" is provided by the security identity which makes it hard to believe this PR is the cause.
I tried a couple other revisions of Quarkus that didn't contain my changes and they failed for me too but maybe I have some leftovers in `.m2/repository`

independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/BeanGenerator.java
```diff
@@ -130,6 +131,9 @@ public BeanGenerator(AnnotationLiteralProcessor annotationLiterals, Predicate<Do
         }
         baseNameBuilder.append("_");
         baseNameBuilder.append(bean.getIdentifier());
+        if (Boolean.TRUE.toString().equals(System.getProperty("quarkus.random-synthetic-bean-names"))) {
```
michalszynkiewicz: @mkouba @manovotn  wdyt?

michalszynkiewicz: (this is to allow running tests that need to generate different synthetic beans in one JVM)

mkouba: I think it's a bad idea. We should try to find a different way to fix this.

manovotn: How come you need the property when I see you added `<reuseForks>false</reuseForks>`?
But to be fair, I don't really like this either. +1 for different way but I have to admit I don't have one at hand...

michalszynkiewicz: the property is only used in MP JWT TCK where `reuseForks = false` doesn't work out of the box because the tests are using testng xml. 
I tried working this around by removing the testng xml and listing the tests in the surefire plugin configuration but then one of the tests starts failing

michalszynkiewicz: I don't treat this as the ultimate solution but a workaround until we handle reloading synthetic beans properly. But if this fix can wait until that is fixed, I'm okay with waiting.
BTW, does the fact that synthetic beans are not in the application classloader also impact live reload?

build-parent/pom.xml
```diff
@@ -155,6 +157,7 @@
             <plugins>
                 <plugin>
                     <artifactId>maven-surefire-plugin</artifactId>
+                    <version>${surefire-plugin.version}</version>
```
stuartwdouglas: This should already be defined by the jboss-parent pom. overriding the version.surefire.plugin should be all that is needed to get the newer version.

independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/BeanGenerator.java
stuartwdouglas: You can fix it like this for now: https://github.com/quarkusio/quarkus/commit/d85e2d58af1594892bacd84dc3fbd7c5520468c0

stuartwdouglas (changes requested):

independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/BeanGenerator.java
michalszynkiewicz: works like a charm, thank you @stuartwdouglas !
I'll update the PR soon.

build-parent/pom.xml
michalszynkiewicz: done

manovotn (approved):
Looks good now!
I didn't even know we have `ApplicationClassPredicateBuildItem`, that's really neat way to circumvent the problem.

manovotn:
The CI failure doesn't look related. It is a `NoSuchElementException` coming from Kafka integration tests (`KafkaConsumerTest`).

michalszynkiewicz:
yay! all green!

extensions/security/deployment/src/main/java/io/quarkus/security/deployment/SecurityProcessor.java
```diff
@@ -85,6 +109,147 @@ void registerSecurityInterceptors(BuildProducer<InterceptorBindingRegistrarBuild
...
+        return result;
+    }
+
+    private Map<MethodInfo, AnnotationInstance> gatherSecurityAnnotations(ApplicationIndexBuildItem indexBuildItem,
+            BeanRegistrar.RegistrationContext registrationContext) {
+        Set<DotName> securityAnnotations = SecurityAnnotationsRegistrar.SECURITY_BINDINGS.keySet();
+        AnnotationStore annotationStore = registrationContext.get(BuildExtension.Key.ANNOTATION_STORE);
+        Set<ClassInfo> classesWithSecurity = new HashSet<>();
+
+        Collection<ClassInfo> classes = indexBuildItem.getIndex().getKnownClasses();
```
stuartwdouglas: Does this really need to iterate over all the classes? Can't you just look for security annotations in the index?

stuartwdouglas (changes requested):
Other than iterating over all the classes I think it looks good. If this could be changed to just look up the relevant annotations from the index then it should be good to go.

extensions/security/deployment/src/main/java/io/quarkus/security/deployment/SecurityProcessor.java
michalszynkiewicz: The problem with using the index is with the synthetic annotations (like the `@DenyAll` when the specific switches are enabled or the `@Secured` transalted to `@RolesAllowed` that @aureamunoz is working on).
I was trying to work around it by emitting build items that would contain such information. But it led to build step cycle with resteasy. I think it was because I use ResteasyDeployment for denying JAX-RS by default.
I'm not 100% sure if it cannot be worked around but thought going through the application index only shouldn't be that costly.

michalszynkiewicz: Plus annotationStore is populated lazily. That's why I could not get a list of classes with synthetic annotations from it.

geoand: I haven't followed this PR lately, but from the start I thought that the BuildItem approach would be the cleanest way to go. 
I'll look into the PR more tomorrow and see if / how it can be done

extensions/security/runtime/src/main/java/io/quarkus/security/runtime/interceptor/SecurityConstrainer.java
```diff
@@ -1,188 +1,35 @@
...
  * @author Michal Szynkiewicz, michal.l.szynkiewicz@gmail.com
  */
 @Singleton
 public class SecurityConstrainer {
 
-    private static final List<Class<? extends Annotation>> SECURITY_ANNOTATIONS = asList(Authenticated.class, DenyAll.class,
-            PermitAll.class, RolesAllowed.class);
-
-    private final Map<Method, Optional<Check>> checkForMethod = new ConcurrentHashMap<>();
+    private final Map<Method, Optional<SecurityCheck>> checkForMethod = new ConcurrentHashMap<>();
```
geoand: I think this is no longer needed, correct?

extensions/security/deployment/src/main/java/io/quarkus/security/deployment/SecurityProcessor.java
```diff
@@ -85,6 +109,147 @@ void registerSecurityInterceptors(BuildProducer<InterceptorBindingRegistrarBuild
...
+        }
+        return methodAnnotations;
+    }
+
+    private AnnotationInstance getSingle(Collection<AnnotationInstance> classAnnotations, Set<DotName> securityAnnotations) {
+        AnnotationInstance result = null;
+        for (AnnotationInstance annotation : classAnnotations) {
+            if (securityAnnotations.contains(annotation.name())) {
+                if (result != null) {
+                    throw new IllegalStateException("Duplicate security annotations on class " + annotation.target());
```
geoand: I think the error message can be improved. `Duplicate` isn't the correct word here IMHO because a developer could do:

```
    @RolesAllowed("admin")
    @DenyAll
    public void doSomething() {
    }
```

and get this error

extensions/security/deployment/src/main/java/io/quarkus/security/deployment/SecurityProcessor.java
```diff
@@ -85,6 +109,147 @@ void registerSecurityInterceptors(BuildProducer<InterceptorBindingRegistrarBuild
...
+    }
+
+    private ResultHandle annotationValues(MethodCreator methodCreator, AnnotationInstance instance) {
+        AnnotationValue value = instance.value();
+        if (value != null && value.asStringArray() != null) {
+            String[] values = value.asStringArray();
+            ResultHandle result = methodCreator.newArray(String.class, methodCreator.load(values.length));
+            int i = 0;
+            for (String val : values) {
+                methodCreator.writeArrayValue(result, i, methodCreator.load(val));
```
geoand: `i` needs to be incremented

geoand:
I am pretty close - I still need to fix a few failing tests but I'll do that later on.
@michalszynkiewicz would you like me to open a PR against your PR when done or would you like me to check push to just to your branch?

michalszynkiewicz:
PR to my branch would be optimal for me

geoand:
Cool. I'll open a PR today or tomorrow if all goes well

geoand:
Actually it seems that I spoke too soon. The `config.denyUnannotated` is creating problems when trying to remove the dependency on the annotation store and the loop over all classes in the index.
I'll keep digging

geoand:
I was able to get around the problems I had and the updated I have passes the security tests. The change seems however to be failing in random HTTP tests which is something I'll need to check out.

extensions/security/deployment/src/main/java/io/quarkus/security/deployment/SecurityProcessor.java
michalszynkiewicz: yeah, and I thought duplicate covers that. Would be grateful for some tips on new name

geoand: I would just say that multiple security annotations are present

extensions/security/deployment/src/main/java/io/quarkus/security/deployment/SecurityProcessor.java
michalszynkiewicz: done

extensions/security/deployment/src/main/java/io/quarkus/security/deployment/SecurityProcessor.java
michalszynkiewicz: done

extensions/security/deployment/src/main/java/io/quarkus/security/deployment/SecurityProcessor.java
michalszynkiewicz: @geoand suggested looping through the index only if there are synthetically added annotations (only if the config options that result in adding `@DenyAll`'s are present).
The newest patchset does it. 
WDYT @stuartwdouglas ?

michalszynkiewicz:
Thank you @geoand !

geoand:
Looks good!

extensions/security/deployment/src/main/java/io/quarkus/security/deployment/SecurityProcessor.java
geoand: The reason we need to loop through the index when denyAll is enabled is because like @michalszynkiewicz mentione above, the annotation store doesn't provide a way to give all the instances of an annotation in the way the Jandex index does.
And the implementation of denyAll is based on the annotation store, since it's the easient way to do it and be sure that `@DenyAll` is only added to CDI beans an not everything under the sun.

geoand:
@michalszynkiewicz can you please rebase your PR onto the latest master?

michalszynkiewicz:
@geoand done, let's see

geoand:
@michalszynkiewicz can you do another rebase please? It looks like CI is holding on to an old version of the spring artifacts

geoand:
Another Spring related PR was merged into master and leaked into the native artifacts. Another rebase is needed...

michalszynkiewicz:
rebased once again...

gsmet:
@stuartwdouglas could you have another look at that one?

TBH, I'm a bit torn about this as it's a big patch but it seems like a fix we need. I marked it as `backport?` so that we try to take care of it before Final but we need to be sure it won't break anything.

geoand:
Once this is in a I have a change ready that will target `1.1` and will pave the way for easier custom security integration (it will change the way https://github.com/quarkusio/quarkus/pull/5225 is implemented and make it possible to implement things like `@PreAuthorize`)

extensions/security/deployment/src/main/java/io/quarkus/security/deployment/SecurityProcessor.java
```diff
@@ -85,6 +114,168 @@ void registerSecurityInterceptors(BuildProducer<InterceptorBindingRegistrarBuild
...
+
+    private Map<MethodInfo, AnnotationInstance> gatherSecurityAnnotationsByLooping(ApplicationIndexBuildItem indexBuildItem,
+            BeanRegistrar.RegistrationContext registrationContext,
+            boolean loopThroughIndex) {
+        Set<DotName> securityAnnotations = SecurityAnnotationsRegistrar.SECURITY_BINDINGS.keySet();
+        AnnotationStore annotationStore = registrationContext.get(BuildExtension.Key.ANNOTATION_STORE);
+        Set<ClassInfo> classesWithSecurity = new HashSet<>();
+
+        Index index = indexBuildItem.getIndex();
+        if (loopThroughIndex) { // only needed when synthetic security annotations are added
```
stuartwdouglas: I don't really like this approach with two different code paths. IMHO a better approach would be to DenyAllBuildItem  be AdditionalSecuredClassesBuildIem, and have it just direct provide the class names of the JAX-RS endpoints, so no iteration is required and there is only one code path.

stuartwdouglas (changes requested):
LGTM except for that one issue I mentioned

extensions/security/deployment/src/main/java/io/quarkus/security/deployment/SecurityProcessor.java
geoand: How would you handle the other denyAll case though, I mean the one that applies to all CDI beans? There isn't a way to get all beans as far as I know.

stuartwdouglas: Look up @DenyAll from the annotation index

stuartwdouglas: Basically for the other case where we want to deny it only applies if there is another security annotation on the CdI bean. This means the existing logic in the other branch should discover them.

geoand: That won't work for the `quarkus.securitu.deny-unannotated` option.
It won't work because that option means that we need to implicitly add DenyAll to all beans that don't have any security annotation, therefore there is nothing to look up in the index.

geoand: Now that raises the question of whether this config option makes sense at all. I personally don't understand it, but I am assuming it's been discussed before and there was good reason to add it.

michalszynkiewicz: for CDI beans, the synthetic deny all only applies to beans that have some security annotations, so that shouldn't be a problem.
Testing once again. I had problems with build steps cycle but that clearly shouldn't be a problem since `DenyAllBuildItem` is passed from step consuming a RestEasy deployment...

stuartwdouglas: It adds DenyAll to methods of a bean that have at least one security annotation:

```
class Foo {
@RolesAllowed("foo")
 void doStuff1() {}

void doOtherStuff() {}

}
```
In this case it will mark doOtherStuff as DenyAll. Basically it is a failsafe against forgetting to add security metadata to a secured bean, you need to explicitly use @PermitAll to state your intentions.

geoand: @michalszynkiewicz if you can get rid of the loop for CDI beans, then the JAX-RS case could be easily handled like @stuartwdouglas proposes.
I am however sceptical that the CDI case can be handled without the loop, but I'd like it very much if it can

geoand: Aha! Know I see what I misunderstood... I thought that DenyAll applies to ALL beans, ever without any security annotation.
So yeah I agree this should be easy and there is no need for the loop

extensions/security/deployment/src/main/java/io/quarkus/security/deployment/SecurityProcessor.java
```diff
@@ -85,6 +112,158 @@ void registerSecurityInterceptors(BuildProducer<InterceptorBindingRegistrarBuild
...
 
+    @BuildStep
+    void gatherSecurityChecks(BuildProducer<BeanRegistrarBuildItem> beanRegistrars,
+            ApplicationIndexBuildItem indexBuildItem,
+            BuildProducer<ApplicationClassPredicateBuildItem> classPredicate,
+            List<AdditionalSecuredClassesBuildIem> additionalSecuredClasses) {
+        classPredicate.produce(new ApplicationClassPredicateBuildItem(new SecurityCheckStorage.AppPredicate()));
+
+        Set<ClassInfo> additionalSecured = new HashSet<>();
+        if (additionalSecuredClasses != null) {
```
stuartwdouglas: This will never be null

stuartwdouglas (approved):
Lgtm, except for one minor comment that is not worth holding the pr over

extensions/security/deployment/src/main/java/io/quarkus/security/deployment/SecurityProcessor.java
geoand: I'll be changing this in a subsequent PR anyway

gsmet:
This one was a pain to backport but it's done. I will run the tests a bit later.

geoand:
Thanks so much @gsmet! A true Quarkus hero :)

michalszynkiewicz:
@gsmet sorry to hear that it was difficult and thank you for doing it :)

PR #5076, opened 2019-10-31
Fix Category.getDescription() and adjust setDescription

rsvoboda:
Fix Category.getDescription() and adjust setDescription
Code located in independent-projects/tools/platform-descriptor-api


gsmet (approved):

gsmet:
Looks good to me but I'll let @aloubyansky or @maxandersen decide as there might be hidden reasons for it to be broken :).

maxandersen:
Merged this while on phone - thanks for catching it. The code was for now just there to allow Emmanuel to add descriptions to categories without breaking jackson parsing. Thus noone should have cross-wired name with description just yet ;)

PR #5077, opened 2019-10-31
ComponentsProviderGenerator - break getComponents into multiple methods

### LINKED ISSUE #5063: ComponentsProviderGenerator - break the generated getComponents() method for large deployments into multiple methods
Otherwise we end up with `org.objectweb.asm.MethodTooLargeException`. `BytecodeRecorderImpl` already has something similar implemented.

mkouba:
- resolves #5063


stuartwdouglas (approved):

manovotn (commented):
I know I am coming late but I did review it nonetheless.
It's a really good solution and as a side effect, it makes the generated class more readable IMO.
Let's see if someone can breach the method limit with that ;-)

PR #5078, opened 2019-10-31
Fix of hasNonArquillianDataProvider which resolves always to true

rsvoboda:
Fix of hasNonArquillianDataProvider which resolves always to `true`


mkouba (approved):

gsmet:
The TCK tests have timed out. As they use Arquillian, I rebased and force pushed (I set a higher timeout in master).

PR #5084, opened 2019-10-31
Use platform for project create

maxandersen:
Why:

 * instead of referring to platform as a bom go back to use platform
   as in future it is not all enforced by a bom.

 * rollback bomXXX change
 * change use of Throwable to Exception as if Throwable happens
   we have worse problems.

### DIFF
--- a/independent-projects/tools/platform-descriptor-resolver-json/src/main/java/io/quarkus/platform/descriptor/resolver/json/QuarkusJsonPlatformDescriptorResolver.java
+++ b/independent-projects/tools/platform-descriptor-resolver-json/src/main/java/io/quarkus/platform/descriptor/resolver/json/QuarkusJsonPlatformDescriptorResolver.java
@@ -278,14 +278,15 @@ public class QuarkusJsonPlatformDescriptorResolver {
         try {
             log.debug("Attempting to resolve Quarkus JSON platform descriptor as %s", jsonArtifact);
             return loadFromFile(artifactResolver.resolve(jsonArtifact));
-        } catch (Throwable e) {
+        } catch (Exception e) {
+            log.debug("Failed to resolve %s due to %s", jsonArtifact, e);
             // it didn't work, now we are trying artifactId-descriptor-json
             final AppArtifact fallbackArtifact = new AppArtifact(jsonArtifact.getGroupId(), jsonArtifact.getArtifactId() + "-descriptor-json", null, "json", jsonArtifact.getVersion());
             log.debug("Attempting to resolve Quarkus JSON platform descriptor as %s", fallbackArtifact);
             try {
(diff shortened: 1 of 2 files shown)


gastaldi:
When you mean "in the future" you mean a `quarkus-platform-bom` and not `quarkus-universe-bom`? Because if we're keeping `quarkus-universe-bom` I don't think it makes sense to call it `platformXXX`

independent-projects/tools/platform-descriptor-resolver-json/src/main/java/io/quarkus/platform/descriptor/resolver/json/QuarkusJsonPlatformDescriptorResolver.java
```diff
@@ -278,14 +278,15 @@ private QuarkusPlatformDescriptor loadFromJsonArtifact(AppModelResolver artifact
         try {
             log.debug("Attempting to resolve Quarkus JSON platform descriptor as %s", jsonArtifact);
             return loadFromFile(artifactResolver.resolve(jsonArtifact));
-        } catch (Throwable e) {
+        } catch (Exception e) {
```
aloubyansky: > change use of Throwable to Exception as if Throwable happens we have worse problems.

For the current state of the code, changing it to Exception is going to break it. IllegalStateException is currently thrown from every method. So until proper checked exceptions are in place, don't change this.

aloubyansky (changes requested):
Please, don't merge changes until you've verified it all works. Or wait for me to test.

independent-projects/tools/platform-descriptor-resolver-json/src/main/java/io/quarkus/platform/descriptor/resolver/json/QuarkusJsonPlatformDescriptorResolver.java
maxandersen: IllegalStateException is an Exception.

The only things that can throw descendents of Throwables that are not Exceptions are things like OutOfMemory or worse.  If that happens you want it to surface not swallow it.

maxandersen:
> When you mean "in the future" you mean a `quarkus-platform-bom` and not `quarkus-universe-bom`? Because if we're keeping `quarkus-universe-bom` I don't think it makes sense to call it `platformXXX`

platform is the generic name for something that is representable as a bom + metadata. At the moment we have universal and core. In future we expect to have more.

maxandersen:
> Please, don't merge changes until you've verified it all works. Or wait for me to test.

what parts is it that shouldn't work ? this is purely changing the name we expose to users for what is used to look up platform (Which result in identifying a bom + json).

The PR just rename back the properties so really interested in knowing what issues it could have ?

aloubyansky (approved):

PR #5087, opened 2019-10-31
Fixes #5080

### LINKED ISSUE #5080: MP Fault Tolerance TCK - Fix excluded tests
Currently, the below tests are excluded from execution when running the MP FT TCK. We need to either fix Quarkus so the test passes or challenge and fix the TCK if it's a problem there.

- [ ] [testBulkheadClassSemaphore3](https://github.com/quarkusio/quarkus/blob/master/tcks/microprofile-fault-tolerance/tck-suite.xml#L28)
- [ ] [testCircuitBreakerMetric](https://github.com/quarkusio/quarkus/blob/master/tcks/microprofile-fault-tolerance/tck-suite.xml#L35)

kenfinnigan:
- Also updates to latest SmallRye Fault Tolerance to align with MP Metrics and SmallRye Metrics versions already present in Quarkus


tcks/microprofile-fault-tolerance/src/test/java/io/quarkus/tck/faulttolerance/TimeoutUninterruptableTest.java
```diff
@@ -0,0 +1,267 @@
...
+import org.jboss.shrinkwrap.api.spec.WebArchive;
+import org.testng.annotations.AfterMethod;
+import org.testng.annotations.Test;
+
+/**
+ * Customized test from MP FT TCK because `testTimeout` fails because Hystrix is unable to interrupt the calling thread, as it
+ * didn't create it.
+ * https://github.com/eclipse/microprofile-fault-tolerance/issues/408
+ */
+public class TimeoutUninterruptableTest extends Arquillian {
```
manovotn: Why don't we instead keep the exclude list with the precise method?
Because what if there is a change/addition on MP side? Then this class will be outdated and pretty much worthless....

kenfinnigan:
It’s because using the xml suite file isn’t compatible with reuseForks=false

manovotn:
Is that a known bug or what's the cause?

With no other way around, couldn't we instead have a test class that extends `TimeoutUninterruptableTest` and then override the one test method and put `@Disabled` on it (since `@Test` is inherited)? That would keep us up to date with what the TCKs do in that class while only excluding the one test.

kenfinnigan:
It's a known bug that they don't operate together, but I don't know the cause.

I will try the approach you suggested

gsmet:
@kenfinnigan can you rebase to latest master? I pushed a higher timeout for the TCK tests.

kenfinnigan:
I'm actually in the process of making the change for @manovotn and I've rebased on master for that, so should be good

kenfinnigan:
Updates pushed, should be good now

manovotn (approved):

kenfinnigan:
@gsmet could you retrigger CI for this? I don't have access.

@manovotn both tested locally and the one TCK test failure we can't reproduce, so wondering if it's just a spurious failure

kenfinnigan:
Nevermind figured out I can rerun just the failed check.

Will see what happens

manovotn:
Added a commit that should *hopefully* fix the CI-exlusive failure.
The cause was random ordering of 2 `org.jboss.arquillian.test.spi.event.suite.Before` observers - one of which was activating request context, the other relying on it being already active.

manovotn:
@kenfinnigan hmm, formatting problem now, could you please fix that? I am no longer near my work station. Shame on me though, it's in my code, sorry...

kenfinnigan:
it's already fixed and new builds running

gsmet (approved):
Looks OK to me.

PR #5089, opened 2019-10-31
Ensure that Security annotation are always placed on non-final methods

### LINKED ISSUE #5051: Quarkus 0.27 - JWT role validation does not work
Hi,

I have just migrated my small project from Quarkus 0.25 to 0.27 and I ran into an issue. 

Let's image we have the [SmallRye JWT extension](https://quarkus.io/guides/jwt-guide) configured and the following class:

```
@Path("/hello")
@RequestScoped
open class GreetingResource {

    @GET
    @Path("/protected")
    @RolesAllowed(value = ["Echoer", "Subscriber"])
    fun helloProtected(@Context ctx: SecurityContext): String {
        val callerName = ctx.userPrincipal?.name ?: "unknown"
        return "hello $callerName"
    }
}
```

Corresponding test:

```
    @Test
    fun testHelloEndpoint_protectedAsUnknown() {
        given()
                .`when`().get("/hello/protected")
                .then()
                .statusCode(401)
    }
```

Everything works fine with 0.25. Unfortunately on Quarkus 0.27 my test fails because a request ends with HTTP 200 (```helloProtected``` method gets called).

I started looking into the problem and I found that in Quarkus 0.27:
- ```RolesFilterRegistrar#configure``` is not called and does not register ```RolesAllowedFilter```
- ```ResteasyDeploymentImpl``` contains a list (```providerClasses```) that in previous version contains ```io.quarkus.resteasy.runtime.RolesFilterRegistrar``` but in 0.27 this class is not there

 in Quarkus 0.25 ```RolesFilterRegistrar``` and ```RolesAllowedFilter``` are in ```quarkus-resteasy``` module, but 0.27 does not contain them. 

It looks that due to the fact that  ```io.quarkus.resteasy.runtime.RolesFilterRegistrar``` is not registered (present on the ```providerClasses``` list) JWT RBAC does not work. If there is a new way how ```RolesAllowedFilter``` is configured then maybe it is not enabled by default. ```RolesAllowedFilter``` class is also present in ```smallrye-jwt-2.0.9.jar```).

geoand:
We need security annotations to result in the creation of an interceptor
which is not possible when a method is final.
The default behavior of Arc to simply warn about the final method is not
good enough for security annotations because it could cause methods
that should have been secure to not be by a simple oversight.

Fixes: #5051

A note on the implementation: I went for a simple approach here instead of introducing some more build items and tying deeper into Arc since IMHO that would require us to handle other Arc warnings in a uniform manner instead of having a specific implementation for interceptors

### DIFF
--- a/extensions/security/deployment/src/main/java/io/quarkus/security/deployment/SecurityProcessor.java
+++ b/extensions/security/deployment/src/main/java/io/quarkus/security/deployment/SecurityProcessor.java
@@ -1,5 +1,6 @@
 package io.quarkus.security.deployment;
 
+import java.lang.reflect.Modifier;
 import java.security.Provider;
 import java.security.Security;
 import java.util.ArrayList;
@@ -7,11 +8,22 @@ import java.util.Arrays;
 import java.util.List;
 import java.util.Set;
 
+import javax.annotation.security.DenyAll;
+import javax.annotation.security.PermitAll;
+import javax.annotation.security.RolesAllowed;
+import javax.enterprise.inject.spi.DeploymentException;
+
+import org.jboss.jandex.DotName;
+import org.jboss.jandex.MethodInfo;
 import org.jboss.logging.Logger;
 
 import io.quarkus.arc.deployment.AdditionalBeanBuildItem;
 import io.quarkus.arc.deployment.AnnotationsTransformerBuildItem;
 import io.quarkus.arc.deployment.InterceptorBindingRegistrarBuildItem;
+import io.quarkus.arc.deployment.ValidationPhaseBuildItem;
+import io.quarkus.arc.processor.AnnotationStore;
+import io.quarkus.arc.processor.BeanInfo;
+import io.quarkus.arc.processor.BuildExtension;
 import io.quarkus.deployment.Capabilities;
 import io.quarkus.deployment.annotations.BuildProducer;
 import io.quarkus.deployment.annotations.BuildStep;
@@ -33,6 +45,11 @@ public class SecurityProcessor {
     private static final Logger log = Logger.getLogger(SecurityProcessor.class);
 
     SecurityConfig security;
+
+    private static final DotName ROLES_ALLOWED = DotName.createSimple(RolesAllowed.class.getName());
+    private static final DotName DENY_ALL = DotName.createSimple(DenyAll.class.getName());
+    private static final DotName PERMIT_ALL = DotName.createSimple(PermitAll.class.getName());
+    private static final List<DotName> STANDARD_SECURITY_ANNOTATIONS = Arrays.asList(ROLES_ALLOWED, DENY_ALL, PERMIT_ALL);
 
     /**
      * Register the Elytron-provided password factory SPI implementation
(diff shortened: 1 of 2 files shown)


gsmet (approved):
It looks good to me but I'll let Stuart have the final say on this matter.

extensions/security/deployment/src/main/java/io/quarkus/security/deployment/SecurityProcessor.java
```diff
@@ -119,4 +136,26 @@ void registerAdditionalBeans(BuildProducer<AdditionalBeanBuildItem> beans) {
...
+    @BuildStep
+    void validateAnnotatedMethods(ValidationPhaseBuildItem validationPhase,
+            BuildProducer<ValidationPhaseBuildItem.ValidationErrorBuildItem> errors) {
+
+        AnnotationStore annotationStore = validationPhase.getContext().get(BuildExtension.Key.ANNOTATION_STORE);
+        for (BeanInfo bean : validationPhase.getContext().get(BuildExtension.Key.BEANS)) {
+            if (bean.isClassBean()) {
+                // Ensure that all security annotations are placed on non final methods
+                // this is done because Arc only warns about final methods not being interceptable
+                for (MethodInfo method : bean.getTarget().get().asClass().methods()) {
```
mkouba: Inherited methods are not processed. Also iterating over all methods of all beans could be quite time-consuming (for large deployments). I'd rather move the validation to `io.quarkus.arc.processor.BeanInfo.initInterceptedMethods()` or somewhere "nearby".

mkouba (changes requested):

extensions/security/deployment/src/main/java/io/quarkus/security/deployment/SecurityProcessor.java
geoand: I pondered that option as well but went with the current solution because it seemed cleaner since it wouldn't involve baking specific security annotations inside the general Arc code.
I can certainly move things inside `initInterceptedMethods` but I assume that the security annotations will have to be hard coded - or at best have some kind of build item that is dedicated to this case and will propagate the information down the Arc layers.
What would be your preference?

gsmet: It might be a stupid question but why don't we throw an error for any interceptor binding annotation on final methods? I mean it won't work and could have very bad consequences. Security is one of the occurrences of this problem but there might be other.

geoand: That's fine with me TBH. I just didn't do it because it was mentioned in the issue that we might not want that as a default.

mkouba: I think that we should modify [BeanDeploymentValidator#skipValidation()](https://github.com/quarkusio/quarkus/blob/master/independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/BeanDeploymentValidator.java#L31) to handle similar use cases...

geoand: @mkouba the problem with that approach is that at the point those are executed, we have long lost the interceptor information.

mkouba: > It might be a stupid question ...

We could and we should (according to the spec) but we had some user requests to ignore those methods and in CDI 2.0 we introduced things like `ProcessBeanAttributes#ignoreFinalMethods()` and `InterceptionFactory#ignoreFinalMethods()`, and in Weld we have a [special config property](https://docs.jboss.org/weld/reference/latest/en-US/html/configure.html#_proxying_classes_with_final_methods) to ignore final methods on some classes.

> but I assume that the security annotations will have to be hard coded...

I'd like to propose some API change to the `BeanDeploymentValidator` API. Hard-coded values are definitely a "no go" ;-).

geoand: OK cool. Waiting for your ping then :)

stuartwdouglas: I think the default behaviour should be to fail on final methods. If users want to turn off that behaviour that is fine, but any mistakes are then on them.

geoand: I tend to agree although I don't have the experience you folks have on this.
 
From my perspective it does make sense to fail when an interceptor can't be created since that creates the least surprise and the reason for the failure it absolutely clear. Heck if we wanted to we could even do some extra work to tell if the class is a Kotlin class and let the user know that then need to `open` the method.

So it probably makes sense to have a configuration option that default to failing instead of coming up with a more elaborate Arc API for validating.

But like I said before, I'm cool with whatever you folks think is best. Just let me know and I'll update the PR :)

stuartwdouglas: Really I think that automatically removing the final modifier is probably the most user friendly approach, especially for Kotlin applications.

geoand: @stuartwdouglas are you thinking for this interceptor use case or for all types of issues we might encounter with Kotlin?

stuartwdouglas: I was just thinking for the interceptor use case.

geoand: OK, I'll do it and we can see what @mkouba thinks :)

geoand: Here you go: https://github.com/quarkusio/quarkus/pull/5104

geoand:
I am going to close this one as #5104 is the more accepted solution

PR #5094, opened 2019-10-31
Add config for incoming Kafka topic

thescouser89:


gsmet:
@cescoffier and if correct, it will probably require a quickstart update too?

thescouser89:
I've been playing with the guide and I now understand why the incoming topic is optional. From what I see, if the incoming "channel" has no topic defined, the channel name is assumed to be the topic.

In this case, since the channel name is 'prices', and the Kafka topic name is 'prices', that config is optional.

Feel free to close this ticket if you think this PR is not appropriate :)

geoand:
Thanks for the PR @thescouser89! 

It seems like it needs a rebase however. 
Of course @cescoffier is the SME here so I'll let him comment

docs/src/main/asciidoc/kafka-guide.adoc
```diff
@@ -238,6 +238,7 @@ mp.messaging.outgoing.generated-price.value.serializer=org.apache.kafka.common.s
 
 # Configure the Kafka source (we read from it)
 mp.messaging.incoming.prices.connector=smallrye-kafka
+mp.messaging.incoming.prices.topic=prices
```
cescoffier: not required. By default, the channel name is used as topic name

PR #5098, opened 2019-10-31
camel: override extensions description

lburgazzoli:


maxandersen:
thanks @lburgazzoli  - this overrides file here applies to core platform; at the moment we need to manually keep it in sync with platform at https://github.com/quarkusio/quarkus-platform/blob/master/bom/runtime/src/main/resources/extensions-overrides.json

I've done that based on this PR.

Issue #5023, opened 2019-10-30
Add additional metadata to MANIFEST.MF


gsmet:
Following this SO question - https://stackoverflow.com/questions/58623758/quarkus-maven-plugin-does-not-add-implementation-entries-to-manifest - stating that we don't push some useful information to the manifest, I think we could push at least a couple more, especially the suggested ones.

This should be done in `JarResultBuildStep`, we have a method to generate the manifest. The information are available in `CurateOutcomeBuildItem#getEffectiveModel()` which exposes at least the artifact name and the version, which could be a good first step.

When we generate the manifest, we read if there's an existing one, we should probably not override the values if they have been defined by the user.

It should probably be some sort of opt-in option as we could see that as an information leak if users don't want to push these info to the published jar.

The maven-jar-plugin option is called `manifest > addDefaultImplementationEntries` so let's mimic it.

gsmet:
@anamarija this one might be a good issue to start with and understand the concept of build step and build items that is core to Quarkus.

geoand:
And of course feel free to reach out for any help you might need

anamarija:
Hi,
Could you please provide some additional info what you mean with the opt-in option. Should it be a custom gradle/maven configuration param which when set should define a build param which in the JarResultBuildStep enables implementationEntries in the manifest. As I understood from the SO question only these two are needed currently

Implementation-Title:  ${appArtifact.artifactId}
Implementation-Version: ${appArtifact.version}

or did you have something else in mind :)

anamarija:
[#5100](https://github.com/quarkusio/quarkus/pull/5100)

PR #5100, opened 2019-10-31
#5023 Add additional metadata to MANIFEST.MF

anamarija:
Added two additional information to MANIFEST.MF

```
Implementation-Title:  ${appArtifact.artifactId}
Implementation-Version: ${appArtifact.version}
```

The fields are only added to the manifest if the addDefaultImplementationEntries is set to true

```xml
     <plugin>
        <groupId>io.quarkus</groupId>
        <configuration>
            <addDefaultImplementationEntries>true</addDefaultImplementationEntries>
        </configuration>
        <executions>
          <execution>
            <goals>
              <goal>build</goal>
            </goals>
          </execution>
        </executions>
      </plugin>
```

fix for [#5023](https://github.com/quarkusio/quarkus/issues/5023)

### DIFF
--- a/core/deployment/src/main/java/io/quarkus/deployment/pkg/ManifestConfig.java
+++ b/core/deployment/src/main/java/io/quarkus/deployment/pkg/ManifestConfig.java
@@ -1,0 +1,15 @@
+package io.quarkus.deployment.pkg;
+
+import io.quarkus.runtime.annotations.ConfigGroup;
+import io.quarkus.runtime.annotations.ConfigItem;
+
+@ConfigGroup
+public class ManifestConfig {
+
+    /**
+     * If the Implementation information should be included in the runner jar's MANIFEST.MF.
+     */
+    @ConfigItem(defaultValue = "true")
+    public boolean addImplementationEntries;
+
+}
--- a/core/deployment/src/main/java/io/quarkus/deployment/pkg/PackageConfig.java
+++ b/core/deployment/src/main/java/io/quarkus/deployment/pkg/PackageConfig.java
@@ -26,6 +26,12 @@ public class PackageConfig {
     public boolean uberJar;
 
     /**
+     * Manifest configuration of the runner jar.
+     */
+    @ConfigItem
+    public ManifestConfig manifest;
+
+    /**
      * The entry point of the application. In most cases this should not be modified.
      */
     @ConfigItem(defaultValue = "io.quarkus.runner.GeneratedMain")
--- a/core/deployment/src/main/java/io/quarkus/deployment/pkg/steps/JarResultBuildStep.java
+++ b/core/deployment/src/main/java/io/quarkus/deployment/pkg/steps/JarResultBuildStep.java
@@ -50,6 +50,7 @@ import io.quarkus.bootstrap.util.IoUtils;
 import io.quarkus.bootstrap.util.ZipUtils;
 import io.quarkus.deployment.annotations.BuildStep;
 import io.quarkus.deployment.builditem.ApplicationArchivesBuildItem;
+import io.quarkus.deployment.builditem.ApplicationInfoBuildItem;
 import io.quarkus.deployment.builditem.GeneratedClassBuildItem;
 import io.quarkus.deployment.builditem.GeneratedFileSystemResourceBuildItem;
 import io.quarkus.deployment.builditem.GeneratedNativeImageClassBuildItem;
@@ -126,6 +127,7 @@ public class JarResultBuildStep {
             OutputTargetBuildItem outputTargetBuildItem,
             TransformedClassesBuildItem transformedClasses,
             ApplicationArchivesBuildItem applicationArchivesBuildItem,
+            ApplicationInfoBuildItem applicationInfo,
             PackageConfig packageConfig,
             List<GeneratedClassBuildItem> generatedClasses,
             List<GeneratedResourceBuildItem> generatedResources,
@@ -133,10 +135,10 @@ public class JarResultBuildStep {
             List<GeneratedFileSystemResourceBuildItem> generatedFileSystemResources) throws Exception {
         if (!uberJarRequired.isEmpty() || packageConfig.uberJar) {
             return buildUberJar(curateOutcomeBuildItem, outputTargetBuildItem, transformedClasses, applicationArchivesBuildItem,
-                    packageConfig, generatedClasses, generatedResources, generatedFileSystemResources);
+                    packageConfig, applicationInfo, generatedClasses, generatedResources, generatedFileSystemResources);
         } else {
             return buildThinJar(curateOutcomeBuildItem, outputTargetBuildItem, transformedClasses, applicationArchivesBuildItem,
-                    packageConfig, generatedClasses, generatedResources, generatedFileSystemResources);
+                    packageConfig, applicationInfo, generatedClasses, generatedResources, generatedFileSystemResources);
         }
     }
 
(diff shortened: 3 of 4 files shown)


devtools/maven/src/main/java/io/quarkus/maven/BuildMojo.java
```diff
@@ -128,6 +128,9 @@
     @Parameter(defaultValue = "false")
     private boolean skip = false;
 
+    @Parameter(defaultValue = "false")
+    private boolean addDefaultImplementationEntries = false;
```
stuartwdouglas: This is not necessary, we are moving away from configuring the output in the mojo, and instead are moving to configuring everything in application.xml.

We are keeping the existing config for backwards compatibility but new config should not be added to the mojo.

stuartwdouglas (dismissed):

devtools/maven/src/main/java/io/quarkus/maven/BuildMojo.java
anamarija: I removed the maven configuration option.

core/deployment/src/main/java/io/quarkus/deployment/pkg/PackageConfig.java
```diff
@@ -25,6 +25,12 @@
     @ConfigItem(defaultValue = "false")
     public boolean uberJar;
 
+    /**
+     * If the java runner jar should contain implementation details information in MANIFEST
```
gsmet:
```suggestion
     * If the java runner jar should include implementation details information in the MANIFEST.MF.
```

core/deployment/src/main/java/io/quarkus/deployment/pkg/PackageConfig.java
```diff
@@ -25,6 +25,12 @@
...
+     */
+    @ConfigItem(defaultValue = "false")
+    public boolean implementationEntries;
```
gsmet: I wonder if we will end up having more options for the manifest so I think I would make `manifest` a proper config group and then have a `addImplementationEntries` boolean in this config group.

core/deployment/src/main/java/io/quarkus/deployment/pkg/steps/JarResultBuildStep.java
```diff
@@ -260,7 +260,8 @@ public FileVisitResult visitFile(Path file, BasicFileAttributes attrs)
 
             copyCommonContent(runnerZipFs, services, applicationArchivesBuildItem, transformedClasses, generatedClasses,
                     generatedResources, seen);
-            generateManifest(runnerZipFs, classPath.toString(), packageConfig);
+            AppArtifact appArtifact = curateOutcomeBuildItem.getEffectiveModel().getAppArtifact();
+            generateManifest(runnerZipFs, classPath.toString(), packageConfig, appArtifact);
```
gsmet: Thinking more about it, I think we should also take into account `ApplicationInfoBuildItem` if we can inject it there (I put this comment here but it's true for all the other occurrences).

Basically, we would take into account `ApplicationInfoBuildItem` information if it is set and if they are `UNSET_VALUE` (it should be tested separately for `name` and `version`), we would default to the artifactId/version from the pom.

gsmet (changes requested):
Looks good, thanks! I added a few suggestions.

core/deployment/src/main/java/io/quarkus/deployment/pkg/PackageConfig.java
gsmet: And after giving it some more thoughts (and changing my mind), I would say let's make it true by default. People who don't want to expose it will be able to disable it. That's OK IMHO.

core/deployment/src/main/java/io/quarkus/deployment/pkg/steps/JarResultBuildStep.java
anamarija: :) Made the changes.

core/deployment/src/main/java/io/quarkus/deployment/pkg/PackageConfig.java
anamarija: No problem.

core/deployment/src/main/java/io/quarkus/deployment/pkg/steps/JarResultBuildStep.java
```diff
@@ -106,6 +107,8 @@
             "META-INF/jandex.idx",
             "LICENSE")));
 
+    private static final String UNSET_VALUE = "<<unset>>";
```
gsmet: Let's make the constant public in `ApplicationInfoBuildItem` instead?

I think we already needed that somewhere else, it's definitely something that should be public if it needs to be consumed from somewhere else.

core/deployment/src/main/java/io/quarkus/deployment/pkg/ManifestConfig.java
```diff
@@ -0,0 +1,20 @@
...
+import io.quarkus.runtime.annotations.ConfigItem;
+
+@ConfigGroup
+public class ManifestConfig {
+
+    /**
+     * Manifest specification version.
+     */
+    @ConfigItem(defaultValue = "1.0")
+    public String version;
```
gsmet: Is it something that people change? I must admit that I'm not very accustomed to MANIFEST.MF fiddling so it's an honest question :).

core/deployment/src/main/java/io/quarkus/deployment/pkg/ManifestConfig.java
```diff
@@ -0,0 +1,20 @@
...
+    /**
+     * If the Implementation information should be included in the runner jar's MANIFEST.MF.
+     */
+    @ConfigItem(defaultValue = "true")
+    public boolean implementationEntries;
```
gsmet: Sorry to be picky but I think `addImplementationEntries` would be more explicit.

core/deployment/src/main/java/io/quarkus/deployment/pkg/steps/JarResultBuildStep.java
```diff
@@ -530,6 +544,16 @@ private void generateManifest(FileSystem runnerZipFs, final String classPath, Pa
             }
         }
         attributes.put(Attributes.Name.MAIN_CLASS, config.mainClass);
+        if (config.manifest.implementationEntries && !attributes.containsKey(Attributes.Name.IMPLEMENTATION_TITLE)) {
+            String name = applicationInfo.getName().equals(UNSET_VALUE) ? appArtifact.getArtifactId()
```
gsmet: It doesn't really matter in this case but it's good practice to do it the other way around:
```
UNSET_VALUE.equals(applicationInfo.getName())
```
It automatically protects you from null values.

gsmet (changes requested):
I added some more comments. We are nearly good to go IMHO.

Once you have adjusted everything, could you squash everything into one commit?

core/deployment/src/main/java/io/quarkus/deployment/pkg/PackageConfig.java
```diff
@@ -25,6 +25,11 @@
     @ConfigItem(defaultValue = "false")
     public boolean uberJar;
 
+    /**
+     * Manifest configuration of the runner jar.
+     */
+    public ManifestConfig manifest;
```
gsmet: Ah sorry I missed that one but we need a `@ConfigItem` annotation even for a config group. It works OK for the config but we need that for the doc generation.

gsmet (changes requested):
Sorry, I added one more comment that I missed earlier. Once this is fixed (and please squash everything), it's good to go.

I will merge it on Monday afternoon, though.

gsmet (approved):

gsmet:
Looks good now, thanks! I will merge it on Monday afternoon.

gsmet:
I just force pushed an update to your branch to update the commit message as it was off with the implementation we now have.

PR #5104, opened 2019-10-31
Ensure that final methods don't prevent CDI interceptors from being applied

### LINKED ISSUE #5051: Quarkus 0.27 - JWT role validation does not work

geoand:
We need security annotations to result in the creation of an interceptor
which is not possible when a method is final.
The solution we follow is to remove the final modifier from methods
that need intercepting.
This is now configurable with the default value true, meaning that
Arc will remove the final flag. If the value is set to false
Arc throws an exception at build time.

Fixes: #5051

Provides an alternative implementation to #5089 (which has now been closed) based on [this](https://github.com/quarkusio/quarkus/pull/5089#discussion_r341386797) comment

### DIFF
--- a/independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/Methods.java
+++ b/independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/Methods.java
@@ -2,13 +2,17 @@ package io.quarkus.arc.processor;
 
 import static io.quarkus.arc.processor.IndexClassLookupUtils.getClassByName;
 
+import io.quarkus.gizmo.DescriptorUtils;
 import java.lang.reflect.Modifier;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
+import java.util.function.BiFunction;
+import java.util.function.Consumer;
 import java.util.stream.Collectors;
 import org.jboss.jandex.AnnotationInstance;
 import org.jboss.jandex.ClassInfo;
@@ -18,6 +22,9 @@ import org.jboss.jandex.MethodInfo;
 import org.jboss.jandex.Type;
 import org.jboss.jandex.TypeVariable;
 import org.jboss.logging.Logger;
+import org.objectweb.asm.ClassVisitor;
+import org.objectweb.asm.MethodVisitor;
+import org.objectweb.asm.Opcodes;
 
 /**
  * 
@@ -115,8 +122,11 @@ final class Methods {
 
     static Set<MethodInfo> addInterceptedMethodCandidates(BeanDeployment beanDeployment, ClassInfo classInfo,
             Map<MethodKey, Set<AnnotationInstance>> candidates,
-            List<AnnotationInstance> classLevelBindings) {
-        Set<MethodInfo> finalMethods = new HashSet<>();
+            List<AnnotationInstance> classLevelBindings, Consumer<BytecodeTransformer> bytecodeTransformerConsumer,
+            boolean removeFinalForProxyableMethods) {
+
+        Set<NameAndDescriptor> methodsFromWhichToRemoveFinal = new HashSet<>();
+        Set<MethodInfo> finalMethodsFoundAndNotChanged = new HashSet<>();
         for (MethodInfo method : classInfo.methods()) {
             if (skipForSubclass(method)) {
                 continue;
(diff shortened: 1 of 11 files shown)


independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/Methods.java
```diff
@@ -115,7 +120,9 @@ static boolean isObjectToString(MethodInfo method) {
 
     static void addInterceptedMethodCandidates(BeanDeployment beanDeployment, ClassInfo classInfo,
             Map<MethodKey, Set<AnnotationInstance>> candidates,
-            List<AnnotationInstance> classLevelBindings) {
+            List<AnnotationInstance> classLevelBindings, Consumer<BytecodeTransformer> bytecodeTransformerConsumer) {
+
+        Set<String> methodsFromWhichToRemoveFinal = new HashSet<>();
```
geoand: This could potentially remove `final` from more methods than intended, but that shouldn't be a problem, right?

independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/Methods.java
```diff
@@ -136,20 +143,35 @@ static void addInterceptedMethodCandidates(BeanDeployment beanDeployment, ClassI
...
+        if (!methodsFromWhichToRemoveFinal.isEmpty()) {
+            bytecodeTransformerConsumer.accept(
+                    new BytecodeTransformer(classInfo.name().toString(), new BiFunction<String, ClassVisitor, ClassVisitor>() {
+                        @Override
+                        public ClassVisitor apply(String s, ClassVisitor classVisitor) {
+                            return new ClassVisitor(Opcodes.ASM7, classVisitor) {
+                                @Override
+                                public MethodVisitor visitMethod(int access, String name, String descriptor, String signature,
+                                        String[] exceptions) {
+                                    if (methodsFromWhichToRemoveFinal.contains(name)) {
```
stuartwdouglas: Technically this should match the signature as well as there could be multiple methods with the same name. In practice I don't think it will make any real difference though, removing final is not a big deal for what we are doing.

geoand: I made the same comment above only with a few seconds difference :P

geoand: This should be fixed now that I lifted some of the utils from gizmo

geoand:
I need to rebase this on top of #5108

geoand:
This is now rebased on top of #5108.

geoand:
I cleared the milestone since it remains to be decided if it should go in for `0.28`

gsmet:
Decision on this one is to wait a bit and see the feedback we have on Martin's changes before deciding if we should take the risk to include it pre 0.29.0.

geoand:
Note to self: rebase onto master and get this one in for 1.1

geoand:
@mkouba @gsmet should we get this one in for `1.1`?

manovotn (approved):
Looks good to me and +1 for tests for both cases right away!
I am not familiar with `ClassVisitor` pattern, so take my review there with a pinch of salt, but it does seem to do what I'd expect it to do ;-)

geoand:
Thanks for the review @manovotn! 

I'll wait for another review then :)

geoand:
@mkouba / @stuartwdouglas are you good with this?

stuartwdouglas (approved):

stuartwdouglas:
LGTM but I will let @mkouba have the final say.

mkouba (changes requested):
Looks really good! The only thing I'm missing is a log message that a final modifier was removed. I know that users usually don't care about these messages but it's a good habit that helps with debugging ;-).

stuartwdouglas:
It should be debug level though

geoand:
Added a debug level logging message

mkouba (approved):

mkouba:
> [ERROR] Failed to execute goal net.revelc.code.formatter:formatter-maven-plugin:2.8.1:validate (default) on project arc-processor: File '/home/vsts/work/1/s/independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/Methods.java' has not been previously formatted.  Please format file and commit before running validation! -> [Help 1]
[ERROR] 

@geoand ;-)

geoand:
> > [ERROR] Failed to execute goal net.revelc.code.formatter:formatter-maven-plugin:2.8.1:validate (default) on project arc-processor: File '/home/vsts/work/1/s/independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/Methods.java' has not been previously formatted.  Please format file and commit before running validation! -> [Help 1]
> > [ERROR]
> 
> @geoand ;-)

LOL, I'll fix it :)

PR #5114, opened 2019-11-01
Avoid null in list-extensions output

maxandersen:


gsmet (approved):
I think I would have done the cleanup directly but it's minor.

PR #5121, opened 2019-11-01
Fix MP Config TCKs to only exclude two failing tests

kenfinnigan:
- Override test class to exclude the two methods, enabling the other methods on the class to execute


gsmet (approved):
Looks good. Let's see what CI thinks.

PR #5125, opened 2019-11-02
Add readiness health check for Kafka Streams

### LINKED ISSUE #4793: Support Health Check for Kafka Streams extension
**Description**
As @gunnarmorling said in the #4675 issue, kafka streams doesn't expose any health checks or metrics. It could be interesting to add the health check capability (at least readiness) as the #4675 seems to be only about metrics.

**Implementation ideas**
Just create the `@Readiness` that will check if the `quarkus.kafka-streams.topics` are reachable or not. Maybe by reusing a part of the source code of `KafkaStreamsTopologyManager#waitForTopicsToBeCreated`:
```java
        try (AdminClient adminClient = AdminClient.create(config)) {
            while (true) {
                try {
                    ListTopicsResult topics = adminClient.listTopics();
                    Set<String> topicNames = topics.names().get(10, TimeUnit.SECONDS);

                    if (topicNames.containsAll(topicsToAwait)) {
                        LOGGER.debug("All expected topics created");
                        return;
                    } else {
                        HashSet<String> missing = new HashSet<>(topicsToAwait);
                        missing.removeAll(topicNames);
                        LOGGER.debug("Waiting for topic(s) to be created: " + missing);
                    }

                    Thread.sleep(1_000);
                } catch (ExecutionException | TimeoutException e) {
                    LOGGER.error("Failed to get topic names from broker", e);
                }
            }
        }
```
Then add the `@BuildStep` in deployment module that will `addHealthCheck` by creating the `HealthBuildItem` refering to the previous `@Readiness` class.

lhauspie:
fixes #4793

### DIFF
--- a/docs/src/main/asciidoc/kafka-streams.adoc
+++ b/docs/src/main/asciidoc/kafka-streams.adoc
@@ -1112,6 +1112,107 @@ CMD ["./application", "-Dquarkus.http.host=0.0.0.0", "-Xmx32m"]
 Now start Docker Compose as described above
 (don't forget to rebuild the container images).
 
+== Kafka Streams Health Checks
+
+If you are using the `quarkus-smallrye-health` extension, `quarkus-kafka-streams` will automatically add:
+
+* a readiness health check to validate that all topics declared in the `quarkus.kafka-streams.topics` property are created,
+* a liveness health check based on the Kafka Streams state.
+
+So when you access the `/health` endpoint of your application you will have information about the state of the Kafka Streams and the available and/or missing topics.
+
+This is an example of when the status is `DOWN`:
+[source, subs=attributes+]
+----
+curl -i http://aggregator:8080/health
+
+HTTP/1.1 503 Service Unavailable
+content-type: application/json; charset=UTF-8
+content-length: 454
+
+{
+    "status": "DOWN",
+    "checks": [
+        {
+            "name": "Kafka Streams state health check",  <1>
+            "status": "DOWN",
+            "data": {
+                "state": "CREATED"
+            }
+        },
+        {
+            "name": "Kafka Streams topics health check",  <2>
+            "status": "DOWN",
+            "data": {
+                "available_topics": "weather-stations,temperature-values",
+                "missing_topics": "hygrometry-values"
+            }
+        }
+    ]
+}
+----
+<1> Liveness health check. Also available at `/health/live` endpoint.
+<2> Readiness health check. Also available at `/health/ready` endpoint.
+
+So as you can see, the status is `DOWN` as soon as one of the `quarkus.kafka-streams.topics` is missing or the Kafka Streams `state` is not `RUNNING`.
+
+If no topics are available, the `available_topics` key will not be present in the `data` field of the `Kafka Streams topics health check`.
+As well as if no topics are missing, the `missing_topics` key will not be present in the `data` field of the `Kafka Streams topics health check`.
+
+You can of course disable the health check of the `quarkus-kafka-streams` extension by setting the `quarkus.kafka-streams.health.enabled` property to `false` in your `application.properties`.
+
+Obviously you can create your liveness and readiness probes based on the respective endpoints `/health/live` and `/health/ready`.
+
+=== Liveness health check
+
+Here is an example of the liveness check:
+```
+curl -i http://aggregator:8080/health/live
+
+HTTP/1.1 503 Service Unavailable
+content-type: application/json; charset=UTF-8
+content-length: 225
+
+{
+    "status": "DOWN",
+    "checks": [
+        {
+            "name": "Kafka Streams state health check",
+            "status": "DOWN",
+            "data": {
+                "state": "CREATED"
+            }
+        }
+    ]
+}
+```
+The `state` is coming from the `KafkaStreams.State` enum.
+
+=== Readiness health check
+
+Here is an example of the readiness check:
+```
+curl -i http://aggregator:8080/health/ready
+
+HTTP/1.1 503 Service Unavailable
+content-type: application/json; charset=UTF-8
+content-length: 265
+
+{
+    "status": "DOWN",
+    "checks": [
+        {
+            "name": "Kafka Streams topics health check",
+            "status": "DOWN",
+            "data": {
+                "missing_topics": "weather-stations,temperature-values"
+            }
+        }
+    ]
+}
+
+```
+
 == Going Further
 
 This guide has shown how you can build stream processing applications using Quarkus and the Kafka Streams APIs,
--- a/extensions/kafka-streams/deployment/src/main/java/io/quarkus/kafka/streams/deployment/KafkaStreamsBuildTimeConfig.java
+++ b/extensions/kafka-streams/deployment/src/main/java/io/quarkus/kafka/streams/deployment/KafkaStreamsBuildTimeConfig.java
@@ -1,0 +1,15 @@
+package io.quarkus.kafka.streams.deployment;
+
+import io.quarkus.runtime.annotations.ConfigItem;
+import io.quarkus.runtime.annotations.ConfigPhase;
+import io.quarkus.runtime.annotations.ConfigRoot;
+
+@ConfigRoot(name = "kafka-streams", phase = ConfigPhase.BUILD_TIME)
+public class KafkaStreamsBuildTimeConfig {
+
+    /**
+     * Whether or not a health check is published in case the smallrye-health extension is present (defaults to true).
+     */
+    @ConfigItem(name = "health.enabled", defaultValue = "true")
+    public boolean healthEnabled;
+}
--- a/extensions/kafka-streams/deployment/src/main/java/io/quarkus/kafka/streams/deployment/KafkaStreamsProcessor.java
+++ b/extensions/kafka-streams/deployment/src/main/java/io/quarkus/kafka/streams/deployment/KafkaStreamsProcessor.java
@@ -33,6 +33,7 @@ import io.quarkus.kafka.streams.runtime.KafkaStreamsRecorder;
 import io.quarkus.kafka.streams.runtime.KafkaStreamsRuntimeConfig;
 import io.quarkus.kafka.streams.runtime.KafkaStreamsTopologyManager;
 import io.quarkus.runtime.LaunchMode;
+import io.quarkus.smallrye.health.deployment.spi.HealthBuildItem;
 
 class KafkaStreamsProcessor {
 
@@ -195,5 +196,17 @@ class KafkaStreamsProcessor {
     @BuildStep
     AdditionalBeanBuildItem registerBean() {
         return AdditionalBeanBuildItem.unremovableOf(KafkaStreamsTopologyManager.class);
+    }
+
+    @BuildStep
+    void addHealthChecks(KafkaStreamsBuildTimeConfig buildTimeConfig, BuildProducer<HealthBuildItem> healthChecks) {
+        healthChecks.produce(
+                new HealthBuildItem(
+                        "io.quarkus.kafka.streams.runtime.health.KafkaStreamsTopicsHealthCheck",
+                        buildTimeConfig.healthEnabled, "kafka-streams"));
+        healthChecks.produce(
+                new HealthBuildItem(
+                        "io.quarkus.kafka.streams.runtime.health.KafkaStreamsStateHealthCheck",
+                        buildTimeConfig.healthEnabled, "kafka-streams"));
     }
 }
--- a/extensions/kafka-streams/runtime/src/main/java/io/quarkus/kafka/streams/runtime/KafkaStreamsTopologyManager.java
+++ b/extensions/kafka-streams/runtime/src/main/java/io/quarkus/kafka/streams/runtime/KafkaStreamsTopologyManager.java
@@ -2,6 +2,7 @@ package io.quarkus.kafka.streams.runtime;
 
 import java.net.InetSocketAddress;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
@@ -50,6 +51,7 @@ public class KafkaStreamsTopologyManager {
     private KafkaStreamsRuntimeConfig runtimeConfig;
     private Instance<Topology> topology;
     private Properties properties;
+    private Map<String, Object> adminClientConfig;
 
     KafkaStreamsTopologyManager() {
         executor = null;
@@ -105,10 +107,11 @@ public class KafkaStreamsTopologyManager {
         Properties streamsProperties = getStreamsProperties(properties, bootstrapServersConfig, runtimeConfig);
 
         streams = new KafkaStreams(topology.get(), streamsProperties);
+        adminClientConfig = getAdminClientConfig(bootstrapServersConfig);
 
         executor.execute(() -> {
             try {
-                waitForTopicsToBeCreated(runtimeConfig.getTrimmedTopics(), bootstrapServersConfig);
+                waitForTopicsToBeCreated(runtimeConfig.getTrimmedTopics());
             } catch (InterruptedException e) {
                 Thread.currentThread().interrupt();
                 return;
@@ -131,21 +134,9 @@ public class KafkaStreamsTopologyManager {
         return streams;
     }
 
-    private void waitForTopicsToBeCreated(Collection<String> topicsToAwait, String bootstrapServersConfig)
+    private void waitForTopicsToBeCreated(Collection<String> topicsToAwait)
             throws InterruptedException {
-        final Map<String, Object> config = new HashMap<>();
-        config.put(AdminClientConfig.BOOTSTRAP_SERVERS_CONFIG, bootstrapServersConfig);
-        // include other AdminClientConfig(s) that have been configured
-        for (final String knownAdminClientConfig : AdminClientConfig.configNames()) {
-            // give preference to admin.<propname> first
-            if (properties.containsKey(StreamsConfig.ADMIN_CLIENT_PREFIX + knownAdminClientConfig)) {
-                config.put(knownAdminClientConfig, properties.get(StreamsConfig.ADMIN_CLIENT_PREFIX + knownAdminClientConfig));
-            } else if (properties.containsKey(knownAdminClientConfig)) {
-                config.put(knownAdminClientConfig, properties.get(knownAdminClientConfig));
-            }
-        }
-
-        try (AdminClient adminClient = AdminClient.create(config)) {
+        try (AdminClient adminClient = AdminClient.create(adminClientConfig)) {
             while (true) {
                 try {
                     ListTopicsResult topics = adminClient.listTopics();
@@ -155,7 +146,7 @@ public class KafkaStreamsTopologyManager {
                         LOGGER.debug("All expected topics created");
                         return;
                     } else {
-                        HashSet<String> missing = new HashSet<>(topicsToAwait);
+                        Set<String> missing = new HashSet<>(topicsToAwait);
                         missing.removeAll(topicNames);
                         LOGGER.debug("Waiting for topic(s) to be created: " + missing);
                     }
@@ -166,6 +157,42 @@ public class KafkaStreamsTopologyManager {
                 }
             }
         }
+    }
+
+    public Set<String> getMissingTopics(Collection<String> topicsToCheck)
+            throws InterruptedException {
+        HashSet<String> missing = new HashSet<>(topicsToCheck);
+
+        try (AdminClient adminClient = AdminClient.create(adminClientConfig)) {
+            ListTopicsResult topics = adminClient.listTopics();
+            Set<String> topicNames = topics.names().get(10, TimeUnit.SECONDS);
+
+            if (topicNames.containsAll(topicsToCheck)) {
+                return Collections.EMPTY_SET;
+            } else {
+                missing.removeAll(topicNames);
+            }
+        } catch (ExecutionException | TimeoutException e) {
+            LOGGER.error("Failed to get topic names from broker", e);
+        }
+
+        return missing;
+    }
+
+    private Map<String, Object> getAdminClientConfig(String bootstrapServersConfig) {
+        Map<String, Object> adminClientConfig = new HashMap<>();
+        adminClientConfig.put(AdminClientConfig.BOOTSTRAP_SERVERS_CONFIG, bootstrapServersConfig);
+        // include other AdminClientConfig(s) that have been configured
+        for (final String knownAdminClientConfig : AdminClientConfig.configNames()) {
+            // give preference to admin.<propname> first
+            if (properties.containsKey(StreamsConfig.ADMIN_CLIENT_PREFIX + knownAdminClientConfig)) {
+                adminClientConfig.put(knownAdminClientConfig,
+                        properties.get(StreamsConfig.ADMIN_CLIENT_PREFIX + knownAdminClientConfig));
+            } else if (properties.containsKey(knownAdminClientConfig)) {
+                adminClientConfig.put(knownAdminClientConfig, properties.get(knownAdminClientConfig));
+            }
+        }
+        return adminClientConfig;
     }
 
     public void setRuntimeConfig(KafkaStreamsRuntimeConfig runtimeConfig) {
--- a/extensions/kafka-streams/runtime/src/main/java/io/quarkus/kafka/streams/runtime/health/KafkaStreamsStateHealthCheck.java
+++ b/extensions/kafka-streams/runtime/src/main/java/io/quarkus/kafka/streams/runtime/health/KafkaStreamsStateHealthCheck.java
@@ -1,0 +1,33 @@
+package io.quarkus.kafka.streams.runtime.health;
+
+import javax.enterprise.context.ApplicationScoped;
+import javax.inject.Inject;
+
+import org.apache.kafka.streams.KafkaStreams;
+import org.eclipse.microprofile.health.HealthCheck;
+import org.eclipse.microprofile.health.HealthCheckResponse;
+import org.eclipse.microprofile.health.HealthCheckResponseBuilder;
+import org.eclipse.microprofile.health.Liveness;
+
+import io.quarkus.kafka.streams.runtime.KafkaStreamsTopologyManager;
+
+@Liveness
+@ApplicationScoped
+public class KafkaStreamsStateHealthCheck implements HealthCheck {
+
+    @Inject
+    private KafkaStreamsTopologyManager manager;
+
+    @Override
+    public HealthCheckResponse call() {
+        HealthCheckResponseBuilder responseBuilder = HealthCheckResponse.named("Kafka Streams state health check");
+        try {
+            KafkaStreams.State state = manager.getStreams().state();
+            responseBuilder.state(state.isRunning())
+                    .withData("state", state.name());
+        } catch (Exception e) {
+            responseBuilder.down().withData("technical_error", e.getMessage());
+        }
+        return responseBuilder.build();
+    }
+}
--- a/extensions/kafka-streams/runtime/src/main/java/io/quarkus/kafka/streams/runtime/health/KafkaStreamsTopicsHealthCheck.java
+++ b/extensions/kafka-streams/runtime/src/main/java/io/quarkus/kafka/streams/runtime/health/KafkaStreamsTopicsHealthCheck.java
@@ -1,0 +1,67 @@
+package io.quarkus.kafka.streams.runtime.health;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+import javax.annotation.PostConstruct;
+import javax.enterprise.context.ApplicationScoped;
+import javax.inject.Inject;
+
+import org.eclipse.microprofile.config.inject.ConfigProperty;
+import org.eclipse.microprofile.health.HealthCheck;
+import org.eclipse.microprofile.health.HealthCheckResponse;
+import org.eclipse.microprofile.health.HealthCheckResponseBuilder;
+import org.eclipse.microprofile.health.Readiness;
+import org.jboss.logging.Logger;
+
+import io.quarkus.kafka.streams.runtime.KafkaStreamsTopologyManager;
+
+@Readiness
+@ApplicationScoped
+public class KafkaStreamsTopicsHealthCheck implements HealthCheck {
+
+    private static final Logger LOGGER = Logger.getLogger(KafkaStreamsTopicsHealthCheck.class.getName());
+
+    @ConfigProperty(name = "quarkus.kafka-streams.topics")
+    public List<String> topics;
+
+    //    @ConfigProperty(name = "quarkus.kafka-streams.bootstrap-servers")
+    //    public List<String> bootstrapServers;
+
+    @Inject
+    private KafkaStreamsTopologyManager manager;
+
+    //    private String commaSeparatedBootstrapServersConfig;
+    private List<String> trimmedTopics;
+
+    @PostConstruct
+    public void init() {
+        //        commaSeparatedBootstrapServersConfig = String.join(",", bootstrapServers);
+        trimmedTopics = topics.stream().map(String::trim).collect(Collectors.toList());
+    }
+
+    @Override
+    public HealthCheckResponse call() {
+        HealthCheckResponseBuilder builder = HealthCheckResponse.named("Kafka Streams topics health check").up();
+
+        try {
+            Set<String> missingTopics = manager.getMissingTopics(trimmedTopics);
+            List<String> availableTopics = new ArrayList<>(trimmedTopics);
+            availableTopics.removeAll(missingTopics);
+
+            if (!availableTopics.isEmpty()) {
+                builder.withData("available_topics", String.join(",", availableTopics));
+            }
+            if (!missingTopics.isEmpty()) {
+                builder.down().withData("missing_topics", String.join(",", missingTopics));
+            }
+        } catch (InterruptedException e) {
+            LOGGER.error("error when retrieving missing topics", e);
+            builder.down().withData("technical_error", e.getMessage());
+        }
+
+        return builder.build();
+    }
+}
--- a/integration-tests/kafka/src/test/java/io/quarkus/it/kafka/streams/KafkaStreamsTest.java
+++ b/integration-tests/kafka/src/test/java/io/quarkus/it/kafka/streams/KafkaStreamsTest.java
@@ -8,6 +8,7 @@ import java.util.Properties;
 
 import javax.inject.Inject;
 
+import org.apache.http.HttpStatus;
 import org.apache.kafka.clients.consumer.Consumer;
 import org.apache.kafka.clients.consumer.ConsumerConfig;
 import org.apache.kafka.clients.consumer.ConsumerRecord;
@@ -20,6 +21,7 @@ import org.apache.kafka.clients.producer.ProducerRecord;
 import org.apache.kafka.common.serialization.IntegerDeserializer;
 import org.apache.kafka.common.serialization.IntegerSerializer;
 import org.eclipse.microprofile.config.Config;
+import org.hamcrest.CoreMatchers;
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Test;
 
@@ -37,14 +39,14 @@ public class KafkaStreamsTest {
     @Inject
     public Config config;
 
-    private static Producer<Integer, Customer> createProducer() {
+    private static Producer<Integer, Customer> createCustomerProducer() {
         Properties props = new Properties();
         props.put(ProducerConfig.BOOTSTRAP_SERVERS_CONFIG, "localhost:19092");
         props.put(ProducerConfig.CLIENT_ID_CONFIG, "streams-test-producer");
         props.put(ProducerConfig.KEY_SERIALIZER_CLASS_CONFIG, IntegerSerializer.class.getName());
         props.put(ProducerConfig.VALUE_SERIALIZER_CLASS_CONFIG, ObjectMapperSerializer.class.getName());
 
-        return new KafkaProducer<Integer, Customer>(props);
+        return new KafkaProducer<>(props);
     }
 
     private static Producer<Integer, Category> createCategoryProducer() {
@@ -73,6 +75,8 @@ public class KafkaStreamsTest {
 
     @Test
     public void testKafkaStreams() throws Exception {
+        testKafkaStreamsNotAliveAndNotReady();
+
         produceCustomers();
 
         Consumer<Integer, EnrichedCustomer> consumer = createConsumer();
@@ -114,13 +118,49 @@ public class KafkaStreamsTest {
         assertCategoryCount(1, 3);
         assertCategoryCount(2, 1);
 
+        testKafkaStreamsAliveAndReady();
+
         // explicitly stopping the pipeline *before* the broker is shut down, as it
         // otherwise will time out
         RestAssured.post("/kafkastreams/stop");
     }
 
+    public void testKafkaStreamsNotAliveAndNotReady() throws Exception {
+        RestAssured.get("/health/ready").then()
+                .statusCode(HttpStatus.SC_SERVICE_UNAVAILABLE)
+                .body("checks[0].name", CoreMatchers.is("Kafka Streams topics health check"))
+                .body("checks[0].status", CoreMatchers.is("DOWN"))
+                .body("checks[0].data.missing_topics", CoreMatchers.is("streams-test-customers,streams-test-categories"));
+
+        RestAssured.when().get("/health/live").then()
+                .statusCode(HttpStatus.SC_SERVICE_UNAVAILABLE)
+                .body("checks[0].name", CoreMatchers.is("Kafka Streams state health check"))
+                .body("checks[0].status", CoreMatchers.is("DOWN"))
+                .body("checks[0].data.state", CoreMatchers.is("CREATED"));
+
+        RestAssured.when().get("/health").then()
+                .statusCode(HttpStatus.SC_SERVICE_UNAVAILABLE);
+    }
+
+    public void testKafkaStreamsAliveAndReady() throws Exception {
+        RestAssured.get("/health/ready").then()
+                .statusCode(HttpStatus.SC_OK)
+                .body("checks[0].name", CoreMatchers.is("Kafka Streams topics health check"))
+                .body("checks[0].status", CoreMatchers.is("UP"))
+                .body("checks[0].data.available_topics", CoreMatchers.is("streams-test-categories,streams-test-customers"));
+
+        RestAssured.when().get("/health/live").then()
+                .statusCode(HttpStatus.SC_OK)
+                .body("checks[0].name", CoreMatchers.is("Kafka Streams state health check"))
+                .body("checks[0].status", CoreMatchers.is("UP"))
+                .body("checks[0].data.state", CoreMatchers.is("RUNNING"));
+
+        RestAssured.when().get("/health").then()
+                .statusCode(HttpStatus.SC_OK);
+    }
+
     private void produceCustomers() {
-        Producer<Integer, Customer> producer = createProducer();
+        Producer<Integer, Customer> producer = createCustomerProducer();
 
         Producer<Integer, Category> categoryProducer = createCategoryProducer();
 
--- a/extensions/kafka-streams/deployment/pom.xml
+++ b/extensions/kafka-streams/deployment/pom.xml
@@ -30,6 +30,10 @@
             <groupId>io.quarkus</groupId>
             <artifactId>quarkus-kafka-streams</artifactId>
         </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-smallrye-health-spi</artifactId>
+        </dependency>
 
         <!-- test dependencies -->
         <dependency>
--- a/extensions/kafka-streams/runtime/pom.xml
+++ b/extensions/kafka-streams/runtime/pom.xml
@@ -34,6 +34,11 @@
             <groupId>org.graalvm.nativeimage</groupId>
             <artifactId>svm</artifactId>
         </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-smallrye-health</artifactId>
+            <optional>true</optional>
+        </dependency>
     </dependencies>
 
     <build>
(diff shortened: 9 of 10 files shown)


lhauspie:
hmm... I don't really understand why the configuration cannot be injected in the `KafkaStreamsHelthCheck` class:
```Java
@Readiness
@ApplicationScoped
public class KafkaStreamsHealthCheck implements HealthCheck {
    @Inject
    private KafkaStreamsRuntimeConfig config;
...
}
```

Instead of that I'm obliged to duplicate the `@ConfigProperty` to inject the `application.properties` in the `HealthCheck` because of this error;
```console
Caused by: javax.enterprise.inject.UnsatisfiedResolutionException: Unsatisfied dependency for type io.quarkus.kafka.streams.runtime.KafkaStreamsRuntimeConfig and qualifiers [@Default]
        - java member: io.quarkus.kafka.streams.runtime.health.KafkaStreamsHealthCheck#config
        - declared on CLASS bean [types=[io.quarkus.kafka.streams.runtime.health.KafkaStreamsHealthCheck, org.eclipse.microprofile.health.HealthCheck, java.lang.Object], qualifiers=[@Readiness, @Any], target=io.quarkus.kafka.streams.runtime.health.KafkaStreamsHealthCheck]
        at io.quarkus.arc.processor.Beans.resolveInjectionPoint(Beans.java:472)
        at io.quarkus.arc.processor.BeanInfo.init(BeanInfo.java:404)
        at io.quarkus.arc.processor.BeanDeployment.init(BeanDeployment.java:212)
        ... 14 more
```

It's not even possible to set the `runtimeConfig` bean in the `KafkaStreamsHelthCheck` class in the recorder:
```Java
@Recorder
public class KafkaStreamsRecorder {
...
    public void configureRuntimeProperties(KafkaStreamsRuntimeConfig runtimeConfig) {
        Arc.container().instance(KafkaStreamsTopologyManager.class).get().setRuntimeConfig(runtimeConfig);
        Arc.container().instance(KafkaStreamsHealthCheck.class).get().setRuntimeConfig(runtimeConfig);
    }
...
}
```

Because no `instance` is found in Arc. And, unfortunately,  same problem when I tried `.getBean(KafkaStreamsHelthCheck.class)` instead of `.instance(KafkaStreamsHelthCheck.class)`.

Could you explain me why ?

Many thanks in advance.

cc @gunnarmorling @loicmathieu @jmartisk

loicmathieu:
KafkaStreamsHelthCheck is a regular CDI bean so the configuration property should be injected via @ConfigProperty item by item like in application code.

KafkaStreamsRuntimeConfig is not a CDI bean so it cannot be injected. You can use it inside your deployment module on a build step that is runtime.

lhauspie:
Oki, thanks, I understand now.

So if the `KafkaStreamHealthCheck` is a regular CDI bean... why can't I retrieve it from Arc containers ?

Is it not yet instantiated when I do this ?
```java
Arc.container().instance(KafkaStreamsHealthCheck.class).get().setRuntimeConfig(runtimeConfig);
```

I'm a bit confused.

machi1990:
> So if the `KafkaStreamHealthCheck` is a regular CDI bean... why can't I retrieve it from Arc containers ?
> 
> Is it not yet instantiated when I do this ?
> 
> ```java
> Arc.container().instance(KafkaStreamsHealthCheck.class).get().setRuntimeConfig(runtimeConfig);
> ```

I think the bean has been removed since Arc detect that it is unused see https://quarkus.io/guides/cdi-reference#remove_unused_beans. 
You can either annotate it with `@Unremovable` or produce a `AdditionalBeanBuildItem` build item 
as shown here https://github.com/quarkusio/quarkus/blob/master/extensions/flyway/deployment/src/main/java/io/quarkus/flyway/FlywayProcessor.java#L66

HTH

lhauspie:
Oki, thanks @machi1990.

I'll check that tonight.

lhauspie:
I added this line in the `application.properties` of the `kafka-streams-quickstart` project but no additional log about removed beans.
```
quarkus.log.category."io.quarkus.arc.producer".level=DEBUG
```

I tried to add the `@Unremovable` annotation to the `KafkaStreamsHealthCheck` class but it still unretrievable from `Arc.container().instance(KafkaStreamsHealthCheck.class)`.

I also tried adding this register method in `KafkaStreamsProcessor` but it still not retrievable.
```java
    @BuildStep
    AdditionalBeanBuildItem registerHealthCheckBean() {
        return AdditionalBeanBuildItem.unremovableOf(KafkaStreamsHealthCheck.class);
    }
```
I finaly tried to register the HealthCheck class by following your example but still not retrievable.
```
        AdditionalBeanBuildItem unremovableProducer = AdditionalBeanBuildItem.unremovableOf(KafkaStreamsHealthCheck.class);
        additionalBeanProducer.produce(unremovableProducer);
```

Maybe I'm doing something wrong.

loicmathieu:
@lhauspie there is an error on the CDI guide

> I added this line in the application.properties of the kafka-streams-quickstart project but no additional log about removed beans.

This should be `quarkus.log.category."io.quarkus.arc.processor".level=DEBUG` and it only works on dev mode.

extensions/kafka-streams/runtime/src/main/java/io/quarkus/kafka/streams/runtime/KafkaStreamsBuildTimeConfig.java
```diff
@@ -0,0 +1,15 @@
...
+
+import io.quarkus.runtime.annotations.ConfigItem;
+import io.quarkus.runtime.annotations.ConfigPhase;
+import io.quarkus.runtime.annotations.ConfigRoot;
+
+@ConfigRoot(name = "kafka-streams", phase = ConfigPhase.BUILD_TIME)
+public class KafkaStreamsBuildTimeConfig {
+
+    /**
+     * Whether or not an healtcheck is published in case the smallrye-health extension is present (default to true).
```
loicmathieu: Minor nit: please add a space between health and check.
Another PR will add it on the existing Javadoc better have all those aligned

extensions/kafka-streams/runtime/src/main/java/io/quarkus/kafka/streams/runtime/KafkaStreamsTopologyManager.java
```diff
@@ -168,6 +159,42 @@ private void waitForTopicsToBeCreated(Collection<String> topicsToAwait, String b
         }
     }
 
+    public Set<String> getMissingTopics(Collection<String> topicsToCheck, String bootstrapServersConfig)
+            throws InterruptedException {
+        final Map<String, Object> config = getAdminClientConfig(bootstrapServersConfig);
+        HashSet<String> missing = new HashSet<>(topicsToCheck);
```
loicmathieu: `HashSet<String> missing => Set<String> missing`

extensions/kafka-streams/runtime/src/main/java/io/quarkus/kafka/streams/runtime/health/KafkaStreamsHealthCheck.java
```diff
@@ -0,0 +1,58 @@
...
+
+    @Inject
+    private KafkaStreamsTopologyManager manager;
+
+    private String commaSeparatedBootstrapServersConfig;
+    private List<String> trimmedTopics;
+
+    @PostConstruct
+    public void init() {
+        commaSeparatedBootstrapServersConfig = String.join(",", (bootstrapServers));
```
loicmathieu: Can you remove the extra parenthesis around bootstrapServers ?

extensions/kafka-streams/runtime/src/main/java/io/quarkus/kafka/streams/runtime/health/KafkaStreamsHealthCheck.java
```diff
@@ -0,0 +1,58 @@
...
+        trimmedTopics = topics.stream().map(String::trim).collect(Collectors.toList());
+    }
+
+    @Override
+    public HealthCheckResponse call() {
+        HealthCheckResponseBuilder builder = HealthCheckResponse.named("Kafka topic(s) health check").up();
```
loicmathieu: Isn't it Kafka Stream health check ? Better use this name

extensions/kafka-streams/runtime/src/main/java/io/quarkus/kafka/streams/runtime/health/KafkaStreamsHealthCheck.java
```diff
@@ -0,0 +1,58 @@
...
+
+        try {
+            trimmedTopics.forEach(topic -> builder.withData(topic, true));
```
loicmathieu: I would add a single data with the list of available topic instead of one line with 'true' for each topic

extensions/kafka-streams/runtime/src/main/java/io/quarkus/kafka/streams/runtime/health/KafkaStreamsHealthCheck.java
```diff
@@ -0,0 +1,58 @@
...
+            manager.getMissingTopics(trimmedTopics, commaSeparatedBootstrapServersConfig).forEach(
+                    missingTopic -> builder.down().withData(missingTopic, false));
```
loicmathieu: And one line with all the missing topics ...

integration-tests/kafka/src/main/resources/application.properties
```diff
@@ -5,6 +5,7 @@ quarkus.log.category.\"org.apache.zookeeper\".level=WARN
 quarkus.kafka-streams.bootstrap-servers=localhost:19092
 quarkus.kafka-streams.application-id=streams-test-pipeline
 quarkus.kafka-streams.topics=streams-test-categories,streams-test-customers
+quarkus.kafka-streams.health.enabled=true
```
loicmathieu: This is the default so it should not be specified

loicmathieu (changes requested):
Overall looks good.
Some minor change requested. 

I didn't check the kafka-stream part in details, someone else needs to have a look at it.

extensions/kafka-streams/runtime/src/main/java/io/quarkus/kafka/streams/runtime/health/KafkaStreamsHealthCheck.java
lhauspie: I agree with you that one line per topic can be too much.

But IMHO, I think it would be better if we respond with 2 lists:
```
{
    "Available topic(s)" : "topic_1,topic_2,...",
    "Missing topic(s)": "topic_a,topic_b,topic_c,...",
}
```

Because this following example seems wird to me and not very relevant:
```
{
    "Available topic(s)": "topic_1,topic_2,...",
    "topic_a": false,
    "topic_b": false,
    "topic_c": false,
    ...
}
```

integration-tests/kafka/src/main/resources/application.properties
lhauspie: You're right

docs/src/main/asciidoc/kafka-streams.adoc
```diff
@@ -1113,6 +1113,28 @@ CMD ["./application", "-Dquarkus.http.host=0.0.0.0", "-Xmx32m"]
...
+    "name": "Kafka Streams health check",
+    "status": "DOWN",
+    "data": {
+        "Available topic(s)": "weather-stations,temperature-values",
+        "Missing topic(s)": "hygrometry-values"
+    }
+}
+```
+If no topics are available, the `Available topic(s)` key will not be present in `data`.
+As well as if no topics are missing, the `Missing topic(s)` key will not be present in `data` and the `status` will be `UP`.
```
loicmathieu: > If no topics ..
I think this could be removed as it didn't add any usefull information.

Maybe this can be rephrased with something like:
> The check will be DOWN as soon as one topic is missing

loicmathieu (approved):
Overall looks goods to me.
Again, a second review with someone more knowledgeable around Kafka Streams needs to be done :)

docs/src/main/asciidoc/kafka-streams.adoc
lhauspie: The meaning of these 2 sentences is : if the list of topics is empty, then the corresponding key is not in the body.

I precise this kind of information to ops guys that would like to set a kind of monitoring that parse the body of readiness.

Maybe the "and the `status` will be `UP`" part can be replace by you proposal before these two lines ?

integration-tests/kafka/src/test/java/io/quarkus/it/kafka/streams/KafkaStreamsTest.java
```diff
@@ -8,18 +8,16 @@
 
 import javax.inject.Inject;
 
-import org.apache.kafka.clients.consumer.Consumer;
-import org.apache.kafka.clients.consumer.ConsumerConfig;
-import org.apache.kafka.clients.consumer.ConsumerRecord;
-import org.apache.kafka.clients.consumer.ConsumerRecords;
-import org.apache.kafka.clients.consumer.KafkaConsumer;
+import org.apache.http.HttpStatus;
+import org.apache.kafka.clients.consumer.*;
```
machi1990: can we avoid the star imports here?

lhauspie: Of course, I used to check there is no star in imports, but I missed this one.

Don't hesitate if you see others.

docs/src/main/asciidoc/kafka-streams.adoc
```diff
@@ -1113,6 +1113,28 @@ CMD ["./application", "-Dquarkus.http.host=0.0.0.0", "-Xmx32m"]
 Now start Docker Compose as described above
 (don't forget to rebuild the container images).
 
+== Kafka Streams Health Check
+
+If you are using the `quarkus-smallrye-health` extension, `quarkus-kafka-streams` will automatically add a readiness health check to validate that all topics declared in the `quarkus.kafka-streams.topics` property are created.
+
+So when you access the `/health/ready` endpoint of your application you will have information about the available and missing topics.
+
+This is an example of `DOWN` status:
```
gunnarmorling: Very cool idea to add the absence/presence of the input topics to the health check 👍 !

In [my post](https://quarkus.io/blog/kafka-streams-applications-with-quarkus-and-microprofile/#healthchecks) from some time ago I did something slightly different: querying the stream app state:

```
public HealthCheckResponse call() {
        return HealthCheckResponse.named("My Pipeline")
                .state(streams.state().isRunning())
                .build();
    }
```

Should we have both?

extensions/kafka-streams/runtime/src/main/java/io/quarkus/kafka/streams/runtime/KafkaStreamsBuildTimeConfig.java
```diff
@@ -0,0 +1,15 @@
+package io.quarkus.kafka.streams.runtime;
+
+import io.quarkus.runtime.annotations.ConfigItem;
+import io.quarkus.runtime.annotations.ConfigPhase;
+import io.quarkus.runtime.annotations.ConfigRoot;
+
+@ConfigRoot(name = "kafka-streams", phase = ConfigPhase.BUILD_TIME)
```
gunnarmorling: Interesting that it is set to `BUILD_TIME`. Do other extensions handle it like that, too?

gunnarmorling (commented):
Very nice, great work! A few comments inline.

docs/src/main/asciidoc/kafka-streams.adoc
lhauspie: I don't realy know if we should or not, the decision is yours. But we technically could.
Would you like I try to add the liveness of your blog post to get a view of the result ?

extensions/kafka-streams/runtime/src/main/java/io/quarkus/kafka/streams/runtime/KafkaStreamsBuildTimeConfig.java
lhauspie: Only two extensions are handling it like that

extensions/undertow/deployment/.../ServletConfig.java:
```
@ConfigRoot(phase = ConfigPhase.BUILD_TIME)
```

extensions/security/deployment/.../SecurityConfig.java:
```
@ConfigRoot(phase = ConfigPhase.BUILD_TIME)
```

So as far as I understand, if the config is used at `BUILD_TIME` it has to be in the `deployment` part of the extension, and not in the `runtime` one.

Let me refactor this.

gunnarmorling:
> Only two extensions are handling it like that

Yes, so eventually the question is: should health check being enabled or not a build or runtime decision? Personally I feel it should be the latter, but if other extensions fix it at build time, we should align. // CC @gsmet

extensions/kafka-streams/runtime/src/main/java/io/quarkus/kafka/streams/runtime/KafkaStreamsBuildTimeConfig.java
gunnarmorling: > Only two extensions are handling it like that

Yes, so eventually the question is: should health check being enabled or not a build or runtime decision? Personally I feel it should be the latter, but if other extensions fix it at build time, we should align. // CC @gsmet

docs/src/main/asciidoc/kafka-streams.adoc
gunnarmorling: I'd like to see the original check being added. It will tell you that the app is not healthy e.g. when receiving a poisonous message.

extensions/kafka-streams/runtime/src/main/java/io/quarkus/kafka/streams/runtime/KafkaStreamsBuildTimeConfig.java
lhauspie: Hmmm. My bad. I thought your question was about "only BUILD_TIME". I misunderstood.

Other extensions with health checks manage this at both with BUILD_AND_RUN_TIME.

Let me fix that. You're right.

lhauspie:
As we can have several `@Liveness` and `@Readiness` classes, I decided to rename the `KafkaStreamsHealthCheck` to `KafkaStreamsTopicsCheck` because its purpose is to check the availability of topics.

So the Liveness that has to check the state of KafkaStreams will be named `KafkaStreamsStateCheck`.

extensions/kafka-streams/runtime/src/main/java/io/quarkus/kafka/streams/runtime/KafkaStreamsBuildTimeConfig.java
loicmathieu: Adding health check is a build decision: why do you ever wanted to disabled an health heck in one env and not in the other, Just don't use it !

Adding it at build time allow to totaly remove all health check related stuff from runtime, this is the decision taken during the implementation of the infrastructure of the extension health check (and it is done like this in Agroal).

If you have a `MyExtensionBuildTimeConfig` object that provides only build time config, you can put it inside your deployement module so it will not end up being packaged inside your jar if you are using JVM mode. Few extensions does this, not all ...

gunnarmorling:
So I'm getting second thoughts on the readiness probe based on topic existence. Is this actually what one would want? IIUC, this has at least two implications:

* No HTTP requests will be routed to the application (affecting all the nodes as they'll all wait for the input topics); this might not be desirable when having web parts and KStreams in one app
* It may prevent rolling upgrades as no node is ready to serve requests

I like exposing this information, but I feel it's actually not part of the applications readiness state. After all, it *is* functional, there's just no input data for it to process.

loicmathieu:
> So the Liveness that has to check the state of KafkaStreams will be named `KafkaStreamsStateCheck`.

Please, don't provides Liveness for this kind of stuff but only readiness check, I can explain you why in details but basically liveness is to define if your application is up so it should not check any of it's dependencies (like Kafak related stuff).

Regarding namming, please keep the `HealthCheck` suffix: `KafkaStreamsTopicsCheck` => `KafkaStreamsTopicsHealthCheck`.

extensions/kafka-streams/runtime/src/main/java/io/quarkus/kafka/streams/runtime/KafkaStreamsBuildTimeConfig.java
gunnarmorling: Ok, cool, let's do it as other extensions then and fix it at build time.

gunnarmorling:
> it should not check any of it's dependencies (like Kafak related stuff)

The liveness probe we're discussing here is about the Kafka Streams app's own state, *not* a dependency (like the broker not running).

loicmathieu:
> I like exposing this information, but I feel it's actually not part of the applications readiness state. After all, it _is_ functional, there's just no input data for it to process.

I always feels that UP/DOWN is too limited for the state of an application, sometimes it is  UP with warnings and these warnings should be addressed by the monitoring/observability tools and not something that will restart the apps automatically.

Health Check is not specifically designed to have automatic restart/failover (like in Kubernetes) but can be use for this. And it's more complex when we mix functionalities in the same service ...

We discussed this with Logan, and we ends up saying that this health check is usefull for kafka-stream application (so not when you mix kafka-stream and REST endpoint in the same apps), other kind of application should disable it.

loicmathieu:
> The liveness probe we're discussing here is about the Kafka Streams app's own state, _not_ a dependency (like the broker not running).

The question is: "is restarting the process will make any difference ?"

gunnarmorling:
> Health Check is not specifically designed to have automatic restart/failover (like in Kubernetes) but can be use for this. 

Sure; let's just hope sole presence of such health check doesn't trick people into using it as a readiness probe then. As said, this doesn't really represent the application's readiness.

> We discussed this with Logan, and we ends up saying that this health check is usefull for kafka-stream application (so not when you mix kafka-stream and REST endpoint in the same apps),

This combination is very common though, when taking interactive queries into account, where a KStreams pipeline is used to run some streaming query logic and the current state (of aggregations etc.) is exposed via Kafka Streams Interactive Queries and REST.

> The question is: "is restarting the process will make any difference ?"

It depends on the specific issue. Sometimes yes, sometimes no. In any case the restart loop would tell an operator that some action is needed. In case of a bad message it could e.g. mean to manually overwrite the app's offset state so to skip this one, or perhaps send another message for the same key to a topic sourcing a `KTable` etc.

docs/src/main/asciidoc/kafka-streams.adoc
lhauspie: Ok, I added your Liveness in addition to my Readiness. And I think the result is pretty good and meaningfull. Indeed if we only have the Liveness we don't really know why the streams is in `CREATED` state.

I also added the `streams.state()` in `data` of `HealthCheckResponse` to be more precise about why the streams is not running.

> Should we have both?

So my opinion is we really should have both ;)

http://localhost:9081/health
```
{
    "status": "DOWN",
    "checks": [
        {
            "name": "Kafka Streams state check",
            "status": "DOWN",
            "data": {
                "state": "CREATED"
            }
        },
        {
            "name": "Kafka Streams topic(s) check",
            "status": "DOWN",
            "data": {
                "Available topic(s)": "weather-stations,temperature-values",
                "Missing topic(s)": "missing-topic"
            }
        }
    ]
}
```

http://localhost:9081/health/ready
```
{
    "status": "DOWN",
    "checks": [
        {
            "name": "Kafka Streams topic(s) check",
            "status": "DOWN",
            "data": {
                "Available topic(s)": "weather-stations,temperature-values",
                "Missing topic(s)": "missing-topic"
            }
        }
    ]
}
```

http://localhost:9081/health/live
```
{
    "status": "DOWN",
    "checks": [
        {
            "name": "Kafka Streams state check",
            "status": "DOWN",
            "data": {
                "state": "CREATED"
            }
        }
    ]
}
```

lhauspie:
I reviewed a last time my PR that sounds OK.

Could you review ?

docs/src/main/asciidoc/kafka-streams.adoc
```diff
@@ -1113,6 +1113,55 @@ CMD ["./application", "-Dquarkus.http.host=0.0.0.0", "-Xmx32m"]
...
+
+* a readiness health check to validate that all topics declared in the `quarkus.kafka-streams.topics` property are created,
+* a liveness health check based on the Kafka Streams state.
+
+So when you access the `/health` endpoint of your application you will have information about the state of the KStreams and the available and/or missing topics.
+
+This is an example of `DOWN` status:
+[source, subs=attributes+]
+----
+http http://aggregator:8080/health
```
loicmathieu: Please use  ̀curl` example as it's more common than `http`

extensions/kafka-streams/deployment/src/main/java/io/quarkus/kafka/streams/deployment/KafkaStreamsBuildTimeConfig.java
```diff
@@ -0,0 +1,15 @@
...
+import io.quarkus.runtime.annotations.ConfigPhase;
+import io.quarkus.runtime.annotations.ConfigRoot;
+
+@ConfigRoot(name = "kafka-streams", phase = ConfigPhase.BUILD_TIME)
+public class KafkaStreamsBuildTimeConfig {
+
+    /**
+     * Whether or not an health check is published in case the smallrye-health extension is present (default to true).
+     */
+    @ConfigItem(name = "health.enabled", defaultValue = "true")
```
loicmathieu: I think this should be a String allowing the following values:

- true
- false
- the name of the check to enable as there is two different check

=> If you don't want to implent it now, please open a followup issue for this

extensions/kafka-streams/deployment/src/main/java/io/quarkus/kafka/streams/deployment/KafkaStreamsProcessor.java
```diff
@@ -195,6 +196,18 @@ AdditionalBeanBuildItem registerBean() {
...
+        return new HealthBuildItem("io.quarkus.kafka.streams.runtime.health.KafkaStreamsTopicsHealthCheck",
+                buildTimeConfig.healthEnabled, "kafka-streams");
+    }
+
+    @BuildStep
+    HealthBuildItem addKafkaStreamsStateHealthCheck(KafkaStreamsBuildTimeConfig buildTimeConfig) {
+        return new HealthBuildItem("io.quarkus.kafka.streams.runtime.health.KafkaStreamsStateHealthCheck",
+                buildTimeConfig.healthEnabled, "kafka-streams");
+    }
+
```
loicmathieu: You can produce those two build items in the same method using a `BuildItemProducer` (ping me if you need some example). Two methods seems overkill for me

extensions/kafka-streams/runtime/src/main/java/io/quarkus/kafka/streams/runtime/health/KafkaStreamsStateHealthCheck.java
```diff
@@ -0,0 +1,26 @@
...
+    @Inject
+    private KafkaStreamsTopologyManager manager;
+
+    @Override
+    public HealthCheckResponse call() {
+        return HealthCheckResponse.named("Kafka Streams state health check")
+                .state(manager.getStreams().state().isRunning())
+                .withData("state", manager.getStreams().state().name())
+                .build();
+    }
```
loicmathieu: You should add this to a try/catch block and generates a DOWN liveness in case of any exception

integration-tests/kafka/src/test/java/io/quarkus/it/kafka/streams/KafkaStreamsTest.java
```diff
@@ -114,13 +118,50 @@ public void testKafkaStreams() throws Exception {
...
+
         // explicitly stopping the pipeline *before* the broker is shut down, as it
         // otherwise will time out
         RestAssured.post("/kafkastreams/stop");
     }
 
+    public void testKafkaStreamsNotAliveAndNotReady() throws Exception {
+        RestAssured.get("/health/ready").then()
+                .statusCode(HttpStatus.SC_SERVICE_UNAVAILABLE)
+                .body(CoreMatchers.containsString("\"name\": \"Kafka Streams topic(s) health check\""))
```
loicmathieu: Please use jsonpath instead of string assertion.

Something like
`.body("checks.name",  CoreMatchers.is("Kafka Streams topic(s) health check"))`

loicmathieu (changes requested):

docs/src/main/asciidoc/kafka-streams.adoc
```diff
@@ -1113,6 +1113,55 @@ CMD ["./application", "-Dquarkus.http.host=0.0.0.0", "-Xmx32m"]
...
+
+HTTP/1.1 503 Service Unavailable
+content-length: 490
+content-type: application/json; charset=UTF-8
+
+{
+    "status": "DOWN",
```
gunnarmorling: So I'm still a bit skeptical about mixing these two thins in one health check. I will effectively make it impossible to set up a K8s liveness probe for restarting the service after a stream processing error, as also non-existing topics may trigger the restart. Would it be possible to set up two distinct health checks perhaps?

lhauspie: hmmm... I'm affraid I don't understand what you mean.
If you base your liveness probe on `/health/live` then K8S will only restart the pod if streams state is down (a stream processing error for instance) not when topics are non-existing.

machi1990 (commented):
I believe all my comments have been addressed, so LGTM. I am in no means an expert on Kafka, so let's wait @gunnarmorling @loicmathieu final call on this PR. 

I just noticed a conflict, I'll let @lhauspie handle that. 
 
Thanks @lhauspie

extensions/kafka-streams/deployment/src/main/java/io/quarkus/kafka/streams/deployment/KafkaStreamsBuildTimeConfig.java
lhauspie: issue #5572 created

docs/src/main/asciidoc/kafka-streams.adoc
gunnarmorling: Ah, ok. Sorry, that wasn't apparent to me, if the two checks can be queried independently, then all is good.

lhauspie: Maybe the adoc is not clear enough ?

extensions/kafka-streams/deployment/src/main/java/io/quarkus/kafka/streams/deployment/KafkaStreamsProcessor.java
lhauspie: For my personal knowledge, why is it better to use the `BuildProducer` rather than 2 `BuildStep` ?

lhauspie:
> I just noticed a conflict, I'll let @lhauspie handle that.
:+1: 

The conflict is merged.

extensions/kafka-streams/deployment/src/main/java/io/quarkus/kafka/streams/deployment/KafkaStreamsProcessor.java
loicmathieu: It's eaisier to read, you will have a single method with the responsibility to produce all build steps related to health.
But as always, it's questionnable ... so do as you want :)

docs/src/main/asciidoc/kafka-streams.adoc
gunnarmorling: Yes, it might be a good idea to show how the two can be queried independently.

extensions/kafka-streams/deployment/src/main/java/io/quarkus/kafka/streams/deployment/KafkaStreamsProcessor.java
```diff
@@ -196,4 +202,34 @@ void configureAndLoadRocksDb(KafkaStreamsRecorder recorder, KafkaStreamsRuntimeC
...
+                new HealthBuildItem(
+                        "io.quarkus.kafka.streams.runtime.health.KafkaStreamsTopicsHealthCheck",
+                        buildTimeConfig.healthEnabled, "kafka-streams"));
+        healthChecks.produce(
+                new HealthBuildItem(
+                        "io.quarkus.kafka.streams.runtime.health.KafkaStreamsStateHealthCheck",
+                        buildTimeConfig.healthEnabled, "kafka-streams"));
+    }
+
+    private boolean isContainerBuild() {
```
gunnarmorling: Where is this used?

gunnarmorling:
One last question above, otherwise it LGTM. Great work, @lhauspie! A very useful addition to the Kafka Streams extension, looking forward to seeing it released.

extensions/kafka-streams/deployment/src/main/java/io/quarkus/kafka/streams/deployment/KafkaStreamsProcessor.java
```diff
@@ -196,4 +202,34 @@ void configureAndLoadRocksDb(KafkaStreamsRecorder recorder, KafkaStreamsRuntimeC
     AdditionalBeanBuildItem registerBean() {
         return AdditionalBeanBuildItem.unremovableOf(KafkaStreamsTopologyManager.class);
     }
+
+    @BuildStep
+    void addHealthChecks(KafkaStreamsBuildTimeConfig buildTimeConfig) {
```
loicmathieu: You can inject the BuillProducer inside your method, this is how it is usually done

```
 void addHealthChecks(KafkaStreamsBuildTimeConfig buildTimeConfig, BuildProducer<HealthBuildItem> healthChecks)
```

loicmathieu (approved):

extensions/kafka-streams/deployment/src/main/java/io/quarkus/kafka/streams/deployment/KafkaStreamsProcessor.java
lhauspie: I don't know.
And I don't remamber to write this peace of code.
Maybe a bad previous confilct resolving.

I removed this method then run again integration tests and everything is working fine. So I think it's useless.

You have the :eye: of :tiger: :boxing_glove:!!

gunnarmorling: Yes, I can confirm it's a left-over. Got obsoleted by 9a099abadd903219b2baa98612f1c740b7edfab8.

gunnarmorling (approved):
LGTM. I'll leave it to you @lhauspie, whether you'd reflect in the docs how to query liveness and readiness checks independently. E.g. just by adding the curl invocations.

docs/src/main/asciidoc/kafka-streams.adoc
```diff
@@ -1112,6 +1112,107 @@ CMD ["./application", "-Dquarkus.http.host=0.0.0.0", "-Xmx32m"]
 Now start Docker Compose as described above
 (don't forget to rebuild the container images).
 
+== Kafka Streams Health Check
```
gsmet:
```suggestion
== Kafka Streams Health Checks
```

docs/src/main/asciidoc/kafka-streams.adoc
```diff
@@ -1112,6 +1112,107 @@ CMD ["./application", "-Dquarkus.http.host=0.0.0.0", "-Xmx32m"]
...
+
+If you are using the `quarkus-smallrye-health` extension, `quarkus-kafka-streams` will automatically add:
+
+* a readiness health check to validate that all topics declared in the `quarkus.kafka-streams.topics` property are created,
+* a liveness health check based on the Kafka Streams state.
+
+So when you access the `/health` endpoint of your application you will have information about the state of the KStreams and the available and/or missing topics.
```
gsmet:
```suggestion
So when you access the `/health` endpoint of your application you will have information about the state of the Kafka Streams and the available and/or missing topics.
```

docs/src/main/asciidoc/kafka-streams.adoc
```diff
@@ -1112,6 +1112,107 @@ CMD ["./application", "-Dquarkus.http.host=0.0.0.0", "-Xmx32m"]
...
+    "checks": [
+        {
+            "name": "Kafka Streams state health check",  <1>
+            "status": "DOWN",
+            "data": {
+                "state": "CREATED"
+            }
+        },
+        {
+            "name": "Kafka Streams topic(s) health check",  <2>
```
gsmet:
```suggestion
            "name": "Kafka Streams topics health check",  <2>
```
I will get rid of the already existing ones with `(s)`. It's less readable and doesn't bring much.

docs/src/main/asciidoc/kafka-streams.adoc
```diff
@@ -1112,6 +1112,107 @@ CMD ["./application", "-Dquarkus.http.host=0.0.0.0", "-Xmx32m"]
...
+                "missing_topics": "hygrometry-values"
+            }
+        }
+    ]
+}
+----
+<1> Liveness health check. Also available at `/health/live` endpoint.
+<2> Readiness health check. Also available at `/health/ready` endpoint.
+
+So as you can see, the status is `DOWN` as soon as one of `quarkus.kafka-streams.topics` is missing or the Kafka Streams `state` is not `RUNNING`.
```
gsmet:
```suggestion
So as you can see, the status is `DOWN` as soon as one of the `quarkus.kafka-streams.topics` is missing or the Kafka Streams `state` is not `RUNNING`.
```

docs/src/main/asciidoc/kafka-streams.adoc
```diff
@@ -1112,6 +1112,107 @@ CMD ["./application", "-Dquarkus.http.host=0.0.0.0", "-Xmx32m"]
...
+
+If no topics are available, the `available_topics` key will not be present in `data` of `Kafka Streams topic(s) health check`.
```
gsmet:
```suggestion
If no topics are available, the `available_topics` key will not be present in the `data` field of the `Kafka Streams topics health check`.
```

docs/src/main/asciidoc/kafka-streams.adoc
```diff
@@ -1112,6 +1112,107 @@ CMD ["./application", "-Dquarkus.http.host=0.0.0.0", "-Xmx32m"]
...
+As well as if no topics are missing, the `missing_topics` key will not be present in `data` of `Kafka Streams topic(s) health check`.
```
gsmet:
```suggestion
As well as if no topics are missing, the `missing_topics` key will not be present in the `data` field of the `Kafka Streams topics health check`.
```

docs/src/main/asciidoc/kafka-streams.adoc
```diff
@@ -1112,6 +1112,107 @@ CMD ["./application", "-Dquarkus.http.host=0.0.0.0", "-Xmx32m"]
...
+
+You can of course disable the health check of `quarkus-kafka-streams` extension by setting the `quarkus.kafka-streams.health.enabled` property to `false`.
```
gsmet:
```suggestion
You can of course disable the health check of the `quarkus-kafka-streams` extension by setting the `quarkus.kafka-streams.health.enabled` property to `false` in your `application.properties`.
```

docs/src/main/asciidoc/kafka-streams.adoc
```diff
@@ -1112,6 +1112,107 @@ CMD ["./application", "-Dquarkus.http.host=0.0.0.0", "-Xmx32m"]
...
+
+Obviously you can create your liveness and readyness probes based on the respective endpoints `/health/live` and `/health/ready`.
```
gsmet:
```suggestion
Obviously you can create your liveness and readiness probes based on the respective endpoints `/health/live` and `/health/ready`.
```

docs/src/main/asciidoc/kafka-streams.adoc
```diff
@@ -1112,6 +1112,107 @@ CMD ["./application", "-Dquarkus.http.host=0.0.0.0", "-Xmx32m"]
...
+
+=== Liveness health check
+
+Here is an example of liveness:
```
gsmet:
```suggestion
Here is an example of the liveness check:
```

docs/src/main/asciidoc/kafka-streams.adoc
```diff
@@ -1112,6 +1112,107 @@ CMD ["./application", "-Dquarkus.http.host=0.0.0.0", "-Xmx32m"]
...
+== Kafka Streams Health Check
+
+If you are using the `quarkus-smallrye-health` extension, `quarkus-kafka-streams` will automatically add:
+
+* a readiness health check to validate that all topics declared in the `quarkus.kafka-streams.topics` property are created,
+* a liveness health check based on the Kafka Streams state.
+
+So when you access the `/health` endpoint of your application you will have information about the state of the KStreams and the available and/or missing topics.
+
+This is an example of `DOWN` status:
```
gsmet:
```suggestion
This is an example of when the status is `DOWN`:
```

docs/src/main/asciidoc/kafka-streams.adoc
```diff
@@ -1112,6 +1112,107 @@ CMD ["./application", "-Dquarkus.http.host=0.0.0.0", "-Xmx32m"]
...
+            "name": "Kafka Streams state health check",
+            "status": "DOWN",
+            "data": {
+                "state": "CREATED"
+            }
+        }
+    ]
+}
+```
+The `state` is coming from `KafkaStreams.State` enum.
```
gsmet:
```suggestion
The `state` is coming from the `KafkaStreams.State` enum.
```

docs/src/main/asciidoc/kafka-streams.adoc
```diff
@@ -1112,6 +1112,107 @@ CMD ["./application", "-Dquarkus.http.host=0.0.0.0", "-Xmx32m"]
...
+
+=== Readyness health check
```
gsmet:
```suggestion
=== Readiness health check
```

docs/src/main/asciidoc/kafka-streams.adoc
```diff
@@ -1112,6 +1112,107 @@ CMD ["./application", "-Dquarkus.http.host=0.0.0.0", "-Xmx32m"]
...
+
+Here is an example of readyness:
```
gsmet:
```suggestion
Here is an example of the readiness check:
```

docs/src/main/asciidoc/kafka-streams.adoc
```diff
@@ -1112,6 +1112,107 @@ CMD ["./application", "-Dquarkus.http.host=0.0.0.0", "-Xmx32m"]
...
+
+HTTP/1.1 503 Service Unavailable
+content-type: application/json; charset=UTF-8
+content-length: 265
+
+{
+    "status": "DOWN",
+    "checks": [
+        {
+            "name": "Kafka Streams topic(s) health check",
```
gsmet:
```suggestion
            "name": "Kafka Streams topics health check",
```

extensions/kafka-streams/deployment/src/main/java/io/quarkus/kafka/streams/deployment/KafkaStreamsBuildTimeConfig.java
```diff
@@ -0,0 +1,15 @@
...
+
+import io.quarkus.runtime.annotations.ConfigItem;
+import io.quarkus.runtime.annotations.ConfigPhase;
+import io.quarkus.runtime.annotations.ConfigRoot;
+
+@ConfigRoot(name = "kafka-streams", phase = ConfigPhase.BUILD_TIME)
+public class KafkaStreamsBuildTimeConfig {
+
+    /**
+     * Whether or not an health check is published in case the smallrye-health extension is present (default to true).
```
gsmet:
```suggestion
     * Whether or not an health check is published in case the smallrye-health extension is present (defaults to true).
```

extensions/kafka-streams/runtime/src/main/java/io/quarkus/kafka/streams/runtime/KafkaStreamsTopologyManager.java
```diff
@@ -168,6 +159,42 @@ private void waitForTopicsToBeCreated(Collection<String> topicsToAwait, String b
         }
     }
 
+    public Set<String> getMissingTopics(Collection<String> topicsToCheck, String bootstrapServersConfig)
+            throws InterruptedException {
+        final Map<String, Object> config = getAdminClientConfig(bootstrapServersConfig);
```
gsmet: Shouldn't this config be created only once?

extensions/kafka-streams/runtime/src/main/java/io/quarkus/kafka/streams/runtime/health/KafkaStreamsTopicsHealthCheck.java
```diff
@@ -0,0 +1,67 @@
...
+
+    @PostConstruct
+    public void init() {
+        commaSeparatedBootstrapServersConfig = String.join(",", bootstrapServers);
+        trimmedTopics = topics.stream().map(String::trim).collect(Collectors.toList());
+    }
+
+    @Override
+    public HealthCheckResponse call() {
+        HealthCheckResponseBuilder builder = HealthCheckResponse.named("Kafka Streams topic(s) health check").up();
```
gsmet:
```suggestion
        HealthCheckResponseBuilder builder = HealthCheckResponse.named("Kafka Streams topics health check").up();
```

gsmet (changes requested):
Sorry for the late review.

I added a bunch of comments in the doc, a few suggestions and questions.

Could you have a look?

I think we are really not far from being in a mergeable state.

Thanks!

gsmet:
@lhauspie any chance you could have a look at my review?

lhauspie-adeo:
> @lhauspie any chance you could have a look at my review?

Yes of course, sorry for the delay, I was busy at Disneyland this weekend ;)

Let me have a look.

lhauspie:
> @lhauspie any chance you could have a look at my review?

Yes of course, sorry for the delay, I was busy at Disneyland this weekend ;)

Let me have a look.

extensions/kafka-streams/runtime/src/main/java/io/quarkus/kafka/streams/runtime/KafkaStreamsTopologyManager.java
lhauspie: You're right. It's even simpler.

gsmet (approved):

gsmet:
I just force-pushed a rebase. Let's wait for CI to go green.

gsmet:
Merged, thanks!

Issue #5101, opened 2019-10-31
 Better Gradle support out on our quest out of beta


maxandersen:
**Description**

There are some corner cases while using Gradle quarkus tooling or the gradle file is not behaving well.

This issue is to collect the known "bigger" issues for info.

related: 
https://github.com/quarkusio/quarkus/pull/4905#issuecomment-546851800 
update documentation to have both maven and gradle variation

Definition of move Gradle out of beta (2020-05-20) 
Good parts:
  - we have tests
  - classloader isolated
Needed:
  - be a good gradle citizen (in/output)
  - devmode on run from ide
  - deployment/runtime mismatch (affects gradle more than maven)

## Epic items
(this section is auto-generated - manual edits will get lost)
- [x] [#3552](https://github.com/quarkusio/quarkus/issues/3552) [gradle] gradle test still not work
- [x] [#4960](https://github.com/quarkusio/quarkus/issues/4960) Building with Gradle gives a warning about CDI dependencies
- [x] [#4982](https://github.com/quarkusio/quarkus/issues/4982) Do not require Gradle to be installed to generate the Gradle wrapper
- [x] [#5204](https://github.com/quarkusio/quarkus/issues/5204) Gradle plugin does not work with plugins DSL
- [x] [#5488](https://github.com/quarkusio/quarkus/issues/5488) EntityManager in not injected when using devmode with gradle
- [x] [#5722](https://github.com/quarkusio/quarkus/issues/5722) Gradle plugin doesn't support multi module project well.
- [x] [#6035](https://github.com/quarkusio/quarkus/issues/6035) Pass Kotlin compiler plugin options to dev mode in Gradle
- [x] [#9046](https://github.com/quarkusio/quarkus/issues/9046) Improve gradle and maven test harness to give us more release confidence
- [x] [#9281](https://github.com/quarkusio/quarkus/issues/9281) Gradle build cache is not being leveraged

gastaldi:
Here is a Gradle project using the minimum configuration (which should be generated by our tools): https://github.com/gastaldi/gradle-quarkus-project

4ntoine:
It would be great to have Maven/Gradle switch for configuration too

maxandersen:
@4ntoine can you elaborate a bit on what you are looking for ?

4ntoine:
@maxandersen Sure, i'd like to have a switcher maven/gradle(groove/kotlin) in build configuration files, like follows (see my link):
```
<dependencyManagement>
[...]

PR #5130, opened 2019-11-02
Fix native image doc inconsistency

### LINKED ISSUE #5111: Inconsistent guides for native image
I see some inconsistencies in guides for native image

https://quarkus.io/guides/building-native-image-guide mentions new way using `<quarkus.package.type>native</quarkus.package.type>` property. 

https://quarkus.io/guides/maven-tooling mentions only maven plugin definition way.

https://quarkus.io/guides/writing-native-applications-tips mentions only maven plugin definition way.

https://quarkus.io/guides/maven-tooling#build-tool-maven says: If you have not used project scaffolding, add the following elements in your pom.xml
But project scaffolding now uses `<quarkus.package.type>native</quarkus.package.type>` property way but the guide uses old way.

I think there should be more details about two ways users can take, what are their advantages and how they should select their way. Maybe few words for guys migrating from previous versions.

I played a bit with some text changes but the result wasn't very good so I scratched it and creating issue instead ;)

gsmet:
Fixes #5111

@geoand @gastaldi  I'm not totally sure of what should be adjusted for Gradle but I would appreciate if one of you could have a look at what I did here and see if we need something similar for the Gradle parts of the documentation (we probably do).

Note: the change in the config file is just about reordering the options so that the most useful ones are at the top of the documentation.

### DIFF
--- a/docs/src/main/asciidoc/maven-tooling.adoc
+++ b/docs/src/main/asciidoc/maven-tooling.adoc
@@ -3,7 +3,7 @@ This guide is maintained in the main Quarkus repository
 and pull requests should be submitted there:
 https://github.com/quarkusio/quarkus/tree/master/docs/src/main/asciidoc
 ////
-= Building Quarkus apps with Maven
+= Quarkus - Building applications with Maven
 
 include::./attributes.adoc[]
 
@@ -396,24 +396,12 @@ If you have not used <<project-creation,project scaffolding>>, add the following
 <profiles>
     <profile> <3>
         <id>native</id>
+        <properties> <4>
+            <quarkus.package.type>native</quarkus.package.type>
+        </properties>
         <build>
             <plugins>
-                <plugin> <3>
-                    <groupId>io.quarkus</groupId>
-                    <artifactId>quarkus-maven-plugin</artifactId>
-                    <version>${quarkus.version}</version>
-                    <executions>
-                        <execution>
-                            <goals>
-                                <goal>native-image</goal>
-                            </goals>
-                            <configuration>
-                                <enableHttpUrlHandler>true</enableHttpUrlHandler>
-                            </configuration>
-                        </execution>
-                    </executions>
-                </plugin>
-                <plugin> <4>
+                <plugin> <5>
                     <groupId>org.apache.maven.plugins</groupId>
                     <artifactId>maven-failsafe-plugin</artifactId>
                     <version>${surefire-plugin.version}</version>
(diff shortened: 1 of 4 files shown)


geoand:
@gsmet sure thing! I'll take a look tomorrow since I don't think I'll have time today.

gsmet:
@gsmet OK cool thanks! If you could review this one at the same time you have a look at the Gradle part, that would be nice!

docs/src/main/asciidoc/maven-tooling.adoc
```diff
@@ -456,6 +445,16 @@ Uber-Jar creation by default excludes link:https://docs.oracle.com/javase/tutori
...
 
+[[configuration-reference]]
+== Configuring the Project Output
+
+There are a several configuration options that will define what the output of your project build will be.
+These are provided in `application.properties` the same as any other config property.
+
+The properties are shown below:
+
+include::{generated-dir}/config/quarkus-core-package-config.adoc[opts=optional]
```
machi1990: 👍

docs/src/main/asciidoc/writing-native-applications-tips.adoc
```diff
@@ -53,26 +53,24 @@ Now, let's tell the `native-image` executable about this configuration file by t
...
+            <quarkus.package.type>native</quarkus.package.type>
+            <quarkus.native.additional-build-args>-H:ResourceConfigurationFiles=${project.basedir}/resources-config.json</quarkus.native.additional-build-args>
+        </properties>
+    </profile>
+</profiles>
 ----
 
+[TIP]
+====
+Multiple options should be separated by a comma.
```
machi1990:
```suggestion
Multiple options can be separated by a comma.
```

gastaldi: IMHO "should" is correct here, since you must use commas in this case

geoand: I agree with @gastaldi

geoand:
Just taking a super quick look at this, I am pretty sure we'll need to corresponding Gradle changes. @gastaldi my availability will be a little flacky this week, do you think you can this one?

docs/src/main/asciidoc/writing-native-applications-tips.adoc
machi1990: Yes. When I made the suggestion I found `should` a bit high toned for a tip hence the suggestion.  But thinking about it, the original sentence was right as @gastaldi said. Let's keep it that way.

gastaldi:
@geoand sure, I'll have a look at that Monday morning

docs/src/main/asciidoc/writing-native-applications-tips.adoc
gastaldi: Actually `must` is the right term here, for the same reason

gsmet:
I merged this. @gastaldi I let you open a new one with the Gradle adjustments next week?

gastaldi:
Adding to #5101

PR #5141, opened 2019-11-03
Ensure kubernetes.group and kubernetes.registry are used

### LINKED ISSUE #5135: Quarkus Kubernetes Dekorate kubernetes.group partially applied
Create a new project
Add quarkus-kubernetes as dependency

Add the following to the `application.properties` file

```
kubernetes.group=example
quarkus.application.name=fruits-app
```

In `target/kubernetes/kubernetes.yml`, you will see

- spec.template.spec.containers.image : ${username}/fruits-app:1.0-SNAPSHOT
- metadata.labels.group: ${username}

Where ${username} is your environment username and where it should be `example`.

@iocanel @geoand 

To decide if that's fixable by 0.28

geoand:
This is just a quick fix around some Dekorate issues. Once Dekorate itself
properly handles these cases, the workaround (but not the test) should
be removed

Fixes: #5135

### DIFF
--- a/extensions/kubernetes/deployment/src/main/java/io/quarkus/kubernetes/deployment/KubernetesProcessor.java
+++ b/extensions/kubernetes/deployment/src/main/java/io/quarkus/kubernetes/deployment/KubernetesProcessor.java
@@ -51,6 +51,8 @@ class KubernetesProcessor {
 
     private static final String DEPLOYMENT_TARGET = "kubernetes.deployment.target";
     private static final String KUBERNETES = "kubernetes";
+    private static final String DOCKER_REGISTRY_PROPERTY = PROPERTY_PREFIX + "docker.registry";
+    private static final String APP_GROUP_PROPERTY = "app.group";
 
     @Inject
     BuildProducer<GeneratedFileSystemResourceBuildItem> generatedResourceProducer;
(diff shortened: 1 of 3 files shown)


extensions/kubernetes/deployment/src/main/java/io/quarkus/kubernetes/deployment/KubernetesProcessor.java
```diff
@@ -92,20 +94,39 @@ public void build(ApplicationInfoBuildItem applicationInfo,
...
 
         // write the generated resources to the filesystem
         final Map<String, String> generatedResourcesMap = session.close();
+
+        // clear the hacky properties if set
+        if (kubernetesGroup.isPresent()) {
+            System.clearProperty(APP_GROUP_PROPERTY);
+        }
+        if (kubernetesRegistry.isPresent()) {
+            System.clearProperty(DOCKER_REGISTRY_PROPERTY);
```
gsmet: This probably needs to be done in a `try / finally` as you always want them to be cleared.

gsmet (changes requested):
I added a small comment.

Looks a bit like dark magic but that's what tests are for :).

extensions/kubernetes/deployment/src/main/java/io/quarkus/kubernetes/deployment/KubernetesProcessor.java
geoand: Not I'm following. What would the `try / finally` enclose?

geoand:
Yeah, it is a bit magical... I wish it didn't have to be so (not least because it took plenty of digging to come up with these fixes), but we need to work around some subtle Dekorate issues here.

extensions/kubernetes/deployment/src/main/java/io/quarkus/kubernetes/deployment/KubernetesProcessor.java
emmanuelbernard: I guess because of 

```
Throws:
SecurityException - if a security manager exists and its checkPropertyAccess method doesn't allow access to the specified system property.
NullPointerException - if key is null.
IllegalArgumentException - if key is empty.
```

geoand: I pushed a change which I assume is what you wanted.

gsmet (approved):
I'd say it's good enough for something we will drop soon.

geoand:
@iocanel when you have some time, please take a look at the issues this PR addresses temporarily and see if they can be properly addressed in Dekorate.

Thanks!

emmanuelbernard (approved):

PR #5142, opened 2019-11-03
Add resteasy capability

lburgazzoli:
In camel-quarkus we do enable some features or behaviors according to
the presence of resteasy and as today the conditions is the presence
of the quarkus-resteasy extension (checked by lookgin at what features
are on the classpath using FeatureBuildItems). 

This PR introduces a new capability "resteasy" that can be used by our
code instead of basing our conditional on extension names.

### DIFF
--- a/extensions/resteasy/deployment/src/main/java/io/quarkus/resteasy/deployment/ResteasyStandaloneBuildStep.java
+++ b/extensions/resteasy/deployment/src/main/java/io/quarkus/resteasy/deployment/ResteasyStandaloneBuildStep.java
@@ -56,7 +56,7 @@ public class ResteasyStandaloneBuildStep {
 
     }
 
-    @BuildStep()
+    @BuildStep(providesCapabilities = Capabilities.RESTEASY)
     @Record(STATIC_INIT)
     public void staticInit(ResteasyStandaloneRecorder recorder,
             Capabilities capabilities,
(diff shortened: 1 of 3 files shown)


lburgazzoli:
I'm not 100% sure about this implementation as RESTEasy itself relies on capabilities to enable/disable some behaviors.

extensions/resteasy/deployment/src/main/java/io/quarkus/resteasy/deployment/ResteasyStandaloneBuildStep.java
```diff
@@ -56,7 +56,7 @@ public ResteasyStandaloneBuildItem(String deploymentRootPath) {
 
     }
 
-    @BuildStep()
+    @BuildStep(providesCapabilities = Capabilities.RESTEASY)
```
geoand: Is there any specific reason that both this one and the `providesCapabilities` on `build` are needed?

lburgazzoli: In fact no, I've annotated both as both the build step are supposed to provide a resteasy's `FeatureBuildItem` so I've replicated the behavior to the related `BuildStep` annotation.

My first tough was to add a `BuildProducer<CapabilityBuildItem>` to both the methods and conditionally add the capability but then I noticed that one of the two builder step are also leveraging capabilities so adding a build producer could trigger a cycle.

An option would be to add an additional BuildItem to signal that resteasy is in use and add the capability.

stuartwdouglas: We probably should get rid of 'providesCapabilities' and just use a normal build item for it, the reason why it is treated specially is largely for historical reasons.

stuartwdouglas:
Superseded by https://github.com/quarkusio/quarkus/pull/5195

PR #5146, opened 2019-11-03
Remove Arc dead code

gwenneg:


independent-projects/arc/runtime/src/main/java/io/quarkus/arc/impl/ArcContainerImpl.java
```diff
@@ -228,7 +228,7 @@ public InjectableContext getActiveContext(Class<? extends Annotation> scopeType)
     public <T> InstanceHandle<T> instance(InjectableBean<T> bean) {
         Objects.requireNonNull(bean);
         requireRunning();
-        return bean != null ? (InstanceHandle<T>) beanInstanceHandle(bean, null) : InstanceHandleImpl.unavailable();
+        return (InstanceHandle<T>) beanInstanceHandle(bean, null);
```
gsmet: Not exactly dead code but yes looks like the underlying method does the work for us.

gsmet (approved):
Looks good to me but I'll let @mkouba decide.

independent-projects/arc/runtime/src/main/java/io/quarkus/arc/impl/ArcContainerImpl.java
gwenneg: The removed condition always evaluated to `true` because of `Objects.requireNonNull(bean);` at the beginning of the method.

gsmet: Ah yes, sure, missed it. In any case `beanInstanceHandle` also deals with the `null` value so we are extra safe.

mkouba (approved):

PR #5148, opened 2019-11-03
Use Files.walk() within a try-with-resources statement

gwenneg:
As stated in the [Files.walk()](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/nio/file/Files.html#walk(java.nio.file.Path,java.nio.file.FileVisitOption...)) Javadoc:

> API Note:
This method must be used within a try-with-resources statement or similar control structure to ensure that the stream's open directories are closed promptly after the stream's operations have completed.


gsmet (approved):
Let's see what CI has to say.

PR #5151, opened 2019-11-03
Don't log IOException at ERROR

stuartwdouglas:
It is very normal for connections to be termianted by the remote
client (e.g. hitting the cancel button in firefox). There is nothing
we can do about these errors so they should be logged at debug.

Logging at a higher level opens up the possibility of a DOS attack,
as any client can trigger an excessive amount of logging by sending
a request then immediatly closing the connection.


gsmet (approved):
Looks good.

gwenneg:
It looks like there's a formatting issue.

PR #5155, opened 2019-11-04
Make LaunchMode injectable

### LINKED ISSUE #589: The infamous DEV mode
We're still missing a `Mode` enum that has three values: `PROD`, `DEV` and `TEST`.

That mode should be obtainable (injectable) by extensions as well as user code. It can be set by configuration files. It should default to `DEV`, or possibly `TEST` when run from Shamrock test runners.

Certain behaviour defaults change depending on that mode:

- Default landing page in `DEV` lists dev info, such as what you can do, and the list of registered URLs, possibly more deployment info. In prod it would probably be a 404.
- Default exception page in `DEV` mode shows stack trace and perhaps even code snippets. In `PROD` it should just inform the user it had an error and give (and log) that error ID (generated unique ID that users can then pass on to support, who can grep for it in the logs which have the stack trace).

So far I've only used `TEST` mode to load test data fixtures specific to tests, to avoid botching my `DEV` DB, but perhaps there are other use-cases for it.

stuartwdouglas:
Fixes #589

### DIFF
--- a/docs/src/main/asciidoc/lifecycle.adoc
+++ b/docs/src/main/asciidoc/lifecycle.adoc
@@ -110,3 +110,11 @@ When the application is stopped, the second log message is printed.
 
 As usual, the application can be packaged using `./mvnw clean package` and executed using the `-runner.jar` file.
 You can also generate the native executable using `./mvnw clean package -Pnative`.
+
+== Launch Modes
+
+Quarkus has 3 different launch modes, `NORMAL` (i.e. production), `DEVELOPMENT` and `TEST`. If you are running `quarkus:dev`
+then the mode will be `DEVELOPMENT`, if you are running a JUnit test it will be `TEST`, otherwise it will be `NORMAL`.
+
+Your application can get the launch mode by injecting the `io.quarkus.runtime.LaunchMode` enum into a CDI bean,
+or by invoking the static method `io.quarkus.runtime.LaunchMode.current()`.
--- a/core/runtime/src/main/java/io/quarkus/runtime/LaunchMode.java
+++ b/core/runtime/src/main/java/io/quarkus/runtime/LaunchMode.java
@@ -1,5 +1,7 @@
 package io.quarkus.runtime;
 
+import io.quarkus.runtime.configuration.ProfileManager;
+
 public enum LaunchMode {
 
     /**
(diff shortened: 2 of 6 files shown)


docs/src/main/asciidoc/lifecycle.adoc
```diff
@@ -110,3 +110,11 @@ When the application is stopped, the second log message is printed.
 
 As usual, the application can be packaged using `./mvnw clean package` and executed using the `-runner.jar` file.
 You can also generate the native executable using `./mvnw clean package -Pnative`.
+
+== Luanch Modes
```
gwenneg:
```suggestion
== Launch Modes
```

docs/src/main/asciidoc/lifecycle.adoc
```diff
@@ -110,3 +110,11 @@ When the application is stopped, the second log message is printed.
...
+
+Quarkus has 3 different launch modes, `NORMAL` (i.e. production), `DEVELOPMENT` and `TEST`. If you are running `quarkus:dev`
+then the mode will be `DEVELOPMENT`, if you are running a JUnit test it will be `TEST`, otherwise it will be normal.
```
gwenneg:
```suggestion
then the mode will be `DEVELOPMENT`, if you are running a JUnit test it will be `TEST`, otherwise it will be `NORMAL`.
```

gwenneg:
I added a few suggestions about the doc.

gwenneg (approved):
LGTM, thanks!

docs/src/main/asciidoc/lifecycle.adoc
```diff
@@ -110,3 +110,11 @@ When the application is stopped, the second log message is printed.
 
 As usual, the application can be packaged using `./mvnw clean package` and executed using the `-runner.jar` file.
 You can also generate the native executable using `./mvnw clean package -Pnative`.
+
+== Launch Modes
+
+Quarkus has 3 different launch modes, `NORMAL` (i.e. production), `DEVELOPMENT` and `TEST`. If you are running `quarkus:dev`
+then the mode will be `DEVELOPMENT`, if you are running a JUnit test it will be `TEST`, otherwise it will be `NORMAL`.
```
jmartisk: Our Arquillian adapter supports TestNG and there the launch mode is `TEST`  too, I'm not sure how much we want to discourage people from using Arquillian, but it is possible, so perhaps we shouldn't narrow this description to just JUnit?

stuartwdouglas: At the moment we only support Arquillian for the MP TCK, this is not something we document for end users.

PR #5162, opened 2019-11-04
Also configure the Surefire plugin in the Maven IT

gsmet:
Otherwise, we can't build the release as the tests do not point to the locally configured repository and the platform cannot resolve the bom.

### DIFF
--- a/integration-tests/maven/pom.xml
+++ b/integration-tests/maven/pom.xml
@@ -111,6 +111,19 @@
 
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-surefire-plugin</artifactId>
+                <configuration>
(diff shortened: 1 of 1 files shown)


maxandersen (approved):
+1 on this - but we'll need fixing why this is happening.

gsmet:
I don't understand your comment. It's happening for a very good reason: we are building in a totally isolated Maven repo so we need the whole Maven build to point to it.

Right now, we build 0.28.0 in this isolated repository and then the tests are run against the default repo where the current released version is not available.

maxandersen:
my comment was that I assumed we were doing this by using a ~/.m2/settings.xml file which everything in the chain should listen to but from our conversation on zulip sounds like its done via property/env variables instead so when tests are forking things doesn't work as expected.

PR #5171, opened 2019-11-04
Make override listen to original extension data

### LINKED ISSUE #5170: override extension no longer seem to merge extenions but instead override
current platform overrides for camel are completely ignoring metadata from camel extensions.

maxandersen:
trying to fix https://github.com/quarkusio/quarkus/issues/5170

not tested yet thus don't merge


devtools/platform-descriptor-json-plugin/src/main/java/io/quarkus/maven/GenerateExtensionsJsonMojo.java
```diff
@@ -335,6 +335,12 @@ private String extensionId(JsonObject extObject) {
 
     private JsonObject mergeObject(JsonObject extObject, JsonObject extOverride) {
         final JsonObjectBuilder mergedObject = Json.createObjectBuilder();
```
quintesse: What I think is that something went missing in this line. I think this should be `Json.createObjectBuilder(extObject);` or something like that (didn't test yet)

maxandersen: that could probably work too!

maxandersen:
replaced by https://github.com/quarkusio/quarkus/pull/5173/files

PR #5173, opened 2019-11-04
Fixes overriding of extensions

### LINKED ISSUE #5170: override extension no longer seem to merge extenions but instead override

quintesse:
Fixes #5170


gsmet (approved):
Max says it's right so let's merge it!

PR #5184, opened 2019-11-04
Use ThreadLocal to store the global value only for gradle add and list

aloubyansky:
… extensions tasks, otherwise the default global config should use a static field

Here is the story. Originally, I'd use a static field to share the default config (it's ugly and will be re-worked after 1.0) but then I realized it doesn't work for Gradle. Specifically, add and list extensions (only these two commands use the platform). Apparently, the tasks are initialized once and re-used accross projects. So, I switched to the ThreadLocal as a holder (and didn't properly clean it up). Which in its turn created a problem for code.quarkus.io with a thread pool in the background.

This change introduces an option how to share the default config: thread local or static field. Thread local and proper cleanup is used for Gradle add and list extensions. Maven Mojos are using the static field. code.quarkus.io will use the static field.

### DIFF
--- a/devtools/gradle/src/main/java/io/quarkus/gradle/tasks/QuarkusAddExtension.java
+++ b/devtools/gradle/src/main/java/io/quarkus/gradle/tasks/QuarkusAddExtension.java
@@ -39,8 +39,11 @@ public class QuarkusAddExtension extends QuarkusPlatformTask {
 
     @TaskAction
     public void addExtension() {
-        setupPlatformDescriptor();
+        execute();
+    }
 
+    @Override
+    protected void doExecute() {
         Set<String> extensionsSet = getExtensionsToAdd()
                 .stream()
                 .flatMap(ext -> stream(ext.split(",")))
--- a/devtools/gradle/src/main/java/io/quarkus/gradle/tasks/QuarkusListExtensions.java
+++ b/devtools/gradle/src/main/java/io/quarkus/gradle/tasks/QuarkusListExtensions.java
@@ -62,9 +62,11 @@ public class QuarkusListExtensions extends QuarkusPlatformTask {
 
     @TaskAction
     public void listExtensions() {
+        execute();
+    }
 
-        setupPlatformDescriptor();
-
+    @Override
+    protected void doExecute() {
         try {
             new ListExtensions(new GradleBuildFileFromConnector(new FileProjectWriter(getProject().getProjectDir())))
                     .listExtensions(
--- a/devtools/gradle/src/main/java/io/quarkus/gradle/tasks/QuarkusPlatformTask.java
+++ b/devtools/gradle/src/main/java/io/quarkus/gradle/tasks/QuarkusPlatformTask.java
@@ -10,16 +10,30 @@ import io.quarkus.platform.descriptor.QuarkusPlatformDescriptor;
 import io.quarkus.platform.descriptor.resolver.json.QuarkusJsonPlatformDescriptorResolver;
 import io.quarkus.platform.tools.config.QuarkusPlatformConfig;
 
-public class QuarkusPlatformTask extends QuarkusTask {
+public abstract class QuarkusPlatformTask extends QuarkusTask {
 
     QuarkusPlatformTask(String description) {
         super(description);
     }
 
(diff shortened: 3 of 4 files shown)


gastaldi (approved):
LGTM

gsmet (changes requested):
Please don't merge yet, apparently it has some issues in native mode.

gsmet:
@aloubyansky @ia3andy would it be useful to add a native IT that does what code.quarkus.io does? Definitely not a good idea to find the issues when upgrading code.quarkus.io.

aloubyansky:
@gsmet I've figured it out, I did miss one line in this PR. It can be merged now. Tested with the current code.quarkus.io master and with the following changes

```
diff --git a/pom.xml b/pom.xml
index 414a026..66b4494 100644
--- a/pom.xml
+++ b/pom.xml
@@ -19,10 +19,10 @@
     <version.commons-compress>1.19</version.commons-compress>
 
     <!-- Quarkus version is used for bundling Code Quarkus -->
-    <version.quarkus>0.28.0</version.quarkus>
+    <version.quarkus>999-SNAPSHOT</version.quarkus>
 
     <!-- Quarkus Platform version must be compatible with Quarkus version -->
-    <version.quarkus-platform>0.0.6</version.quarkus-platform>
+    <version.quarkus-platform>999-SNAPSHOT</version.quarkus-platform>
   </properties>
   <dependencyManagement>
     <dependencies>
diff --git a/src/main/kotlin/io/quarkus/code/services/QuarkusExtensionCatalog.kt b/src/main/kotlin/io/quarkus/code/services/QuarkusExtensionCatalog.kt
index 36d99af..024183c 100644
--- a/src/main/kotlin/io/quarkus/code/services/QuarkusExtensionCatalog.kt
+++ b/src/main/kotlin/io/quarkus/code/services/QuarkusExtensionCatalog.kt
@@ -3,6 +3,7 @@ package io.quarkus.code.services
 import com.google.common.base.Preconditions.checkState
 import io.quarkus.code.model.CodeQuarkusExtension
 import io.quarkus.platform.descriptor.resolver.json.QuarkusJsonPlatformDescriptorResolver
+import io.quarkus.platform.tools.config.QuarkusPlatformConfig
 import org.eclipse.microprofile.config.spi.ConfigProviderResolver
 import javax.inject.Singleton
 
@@ -21,6 +22,7 @@ open class QuarkusExtensionCatalog {
 
         init {
             checkState(descriptor.quarkusVersion == bundledQuarkusVersion, "The platform version (%s) must be compatible with the bundled Quarkus version (%s != %s)", descriptor.bomVersion, descriptor.quarkusVersion, bundledQuarkusVersion)
+            QuarkusPlatformConfig.defaultConfigBuilder().setPlatformDescriptor(descriptor).build()
         }
     }
 
diff --git a/src/main/kotlin/io/quarkus/code/services/QuarkusProjectCreator.kt b/src/main/kotlin/io/quarkus/code/services/QuarkusProjectCreator.kt
index ebdaf0f..abe555e 100644
--- a/src/main/kotlin/io/quarkus/code/services/QuarkusProjectCreator.kt
+++ b/src/main/kotlin/io/quarkus/code/services/QuarkusProjectCreator.kt
@@ -31,9 +31,6 @@ open class QuarkusProjectCreator {
     }
 
     open fun create(project: QuarkusProject): ByteArray {
-        if (!QuarkusPlatformConfig.hasGlobalDefault()) {
-            QuarkusPlatformConfig.defaultConfigBuilder().setPlatformDescriptor(QuarkusExtensionCatalog.descriptor).build()
-        }
         val baos = ByteArrayOutputStream()
         baos.use {
             val zipWriter = CommonsZipProjectWriter.createWriter(baos, project.artifactId)

```

@ia3andy

gwenneg:
@aloubyansky There's a conflicting file, could you please rebase this PR?

aloubyansky:
Could somebody merge this please?

maxandersen (approved):
+1 on the change to use of threadlocal (which I hope we can get rid of in 1.1+ :)

If verified to work with code.quarkus we shuold be good here too.

gsmet (approved):

PR #5185, opened 2019-11-04
Introduce the ability to configure Agroal validation query

### LINKED ISSUE #4082: Add datasource configurations for initSQL and validationQuery
**Description**
Just tried to set up a connection pool using Quarkus. The first limitation is that there are no configuration keys for initSQL and validationQuery which can be configured using the application.properties file.

anamarija:
Fixes [4082](https://github.com/quarkusio/quarkus/issues/4082)
added configuration 
quarkus.datasource.validation-query-sql

### DIFF
--- a/extensions/agroal/runtime/src/main/java/io/quarkus/agroal/runtime/AbstractDataSourceProducer.java
+++ b/extensions/agroal/runtime/src/main/java/io/quarkus/agroal/runtime/AbstractDataSourceProducer.java
@@ -1,6 +1,8 @@
 package io.quarkus.agroal.runtime;
 
+import java.sql.Connection;
 import java.sql.Driver;
+import java.sql.Statement;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Optional;
@@ -166,6 +168,22 @@ public abstract class AbstractDataSourceProducer {
         }
         if (dataSourceRuntimeConfig.backgroundValidationInterval.isPresent()) {
             poolConfiguration.validationTimeout(dataSourceRuntimeConfig.backgroundValidationInterval.get());
+        }
+        if (dataSourceRuntimeConfig.validationQuerySql.isPresent()) {
+            String validationQuery = dataSourceRuntimeConfig.validationQuerySql.get();
+            poolConfiguration.connectionValidator(new ConnectionValidator() {
+
+                @Override
+                public boolean isValid(Connection connection) {
+                    try (Statement stmt = connection.createStatement()) {
+                        stmt.execute(validationQuery);
+                        return true;
+                    } catch (Exception e) {
+                        log.warn("Connection validation failed", e);
+                    }
+                    return false;
+                }
+            });
         }
         if (dataSourceRuntimeConfig.idleRemovalInterval.isPresent()) {
             poolConfiguration.reapTimeout(dataSourceRuntimeConfig.idleRemovalInterval.get());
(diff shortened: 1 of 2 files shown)


gsmet:
Don't worry about the CI failure, it's unrelated to your changes.

Could you rebase on top of master so that we have only one clean commit? Thanks!

extensions/agroal/runtime/src/main/java/io/quarkus/agroal/runtime/AbstractDataSourceProducer.java
```diff
@@ -167,6 +169,22 @@ public AgroalDataSource createDataSource(String dataSourceName,
...
+            String validationQuery = dataSourceRuntimeConfig.validationQuerySql.get();
+            poolConfiguration.connectionValidator(new ConnectionValidator() {
+
+                @Override
+                public boolean isValid(Connection connection) {
+                    try (Statement stmt = connection.createStatement()) {
+                        stmt.execute(validationQuery);
+                        return true;
+                    } catch (Exception e) {
+                        log.warn("Connection validation failed", e);
```
NetoDevel: Wouldn't an `ERROR `log be better?

anamarija: if a validation check of a connection fails, it indicates that the connection needs to be dropped and a new one needs to be provided. As such IMHO this does not necessary indicate an error event, but more of a warning information that this happened. 
But please correct me if I am wrong.

gsmet (commented):
Looks good to me, I just have an hesitation on the name of the config. Let's see what @Sanne thinks.

Sanne:
The name of the property is ok for me.

Higher level question: is such a configuration property still actually useful?

I have the impression such things were "fashionable" in the past as the JDBC standard didn't have an `isValid()` method on Connection in older times.
It's been a while and now all drivers I know of, implement it reasonably well - so why override the strategy?

My personal impression is that people think this configuration aspect is still necessary, while actually it's just "we've always done it that way": cargo-cult.

@barreiro and @jesperpedersen  you have a suggestion?

jesperpedersen:
@Sanne Yes, this is from before `isValid()` was added, so it should be a fall-back if `SQLFeatureNotSupportedException` is thrown.

I think that `validationQuerySql` is ok, as it is clear that the user has to write the SQL used against the database, like `SELECT 1`.

gsmet:
A user asked for it and, be sure that we have a lot of weird/malfunctioning/old JDBC drivers out there.

So I think it's worth it.

The name seems OK to everyone so I'll merge it.

gsmet (approved):
Thanks @anamarija !

extensions/agroal/runtime/src/main/java/io/quarkus/agroal/runtime/AbstractDataSourceProducer.java
anamarija: No problem :)

Sanne:
ok to merge and many thanks @anamarija, @NetoDevel & @gsmet !

Still, since we support a closed set of JDBC drivers and at a well defined version, would be nice to check which ones ever need it. Also maybe we should move the `ConnectionValidator` implementation into Agroal's sources, and improve it to act only as a fallback as suggested by Jesper.

PR #5193, opened 2019-11-05
Fix potential NPEs and add missing hashCode() methods

gwenneg:


devtools/platform-descriptor-json-plugin/src/main/java/io/quarkus/maven/GenerateExtensionsJsonMojo.java
```diff
@@ -275,6 +275,7 @@ private JsonNode processMetaInfDir(Artifact artifact, Path metaInfDir)
             mapper = getMapper(true);
             jsonOrYaml = yaml;
         } else {
+            mapper = getMapper(false);
```
gwenneg: A NullPointerException will be thrown if line 283 (`return mapper.createObjectNode()...`) is executed without this change. That's probably not so important considering the conditions to get there, but that's still a potential NPE :)

gsmet (approved):

gsmet:
I had to cancel CI due to more pressing things, will relaunch ASAP.

PR #5195, opened 2019-11-05
Deprecate providesCapabilities and applicationArchiveMarkers

### LINKED ISSUE #5194: Deprecated @BuildStep#providesCapabilities
This should really just use CapabilityBuildItem directly, there is nothing special about capabilities where they should required their own value on the annotation.

stuartwdouglas:
These are just build items so there is really no need for them
to be given special treatment.

Also adds RESTEasy capability

Fixes #5194

### DIFF
--- a/extensions/panache/mongodb-panache/deployment/src/main/java/io/quarkus/mongodb/panache/deployment/PanacheResourceProcessor.java
+++ b/extensions/panache/mongodb-panache/deployment/src/main/java/io/quarkus/mongodb/panache/deployment/PanacheResourceProcessor.java
@@ -10,10 +10,12 @@ import org.jboss.jandex.DotName;
 import org.jboss.jandex.Indexer;
 import org.jboss.jandex.Type;
 
+import io.quarkus.deployment.Capabilities;
 import io.quarkus.deployment.annotations.BuildProducer;
 import io.quarkus.deployment.annotations.BuildStep;
 import io.quarkus.deployment.builditem.ApplicationIndexBuildItem;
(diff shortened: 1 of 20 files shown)


extensions/panache/mongodb-panache/deployment/src/main/java/io/quarkus/mongodb/panache/deployment/PanacheResourceProcessor.java
```diff
@@ -35,7 +37,12 @@
 
     private static final DotName DOTNAME_OBJECT_ID = DotName.createSimple(ObjectId.class.getName());
 
-    @BuildStep(providesCapabilities = "io.quarkus.mongodb.panache")
+    @BuildStep
+    CapabilityBuildItem capability() {
+        return new CapabilityBuildItem(Capabilities.MONGO_PANACHE);
```
gsmet: It's very minor but could we call it `MONGODB_PANACHE` as the feature? See a few lines below.

gsmet (changes requested):
Looks good to be except for a very minor gripe on a constant name.

extensions/panache/mongodb-panache/deployment/src/main/java/io/quarkus/mongodb/panache/deployment/PanacheResourceProcessor.java
gsmet: I had some spare cycles waiting for CI so I fixed it and fixed another typo on the OAUTH2 constant.

I just force pushed.

gsmet (approved):

dmlloyd (approved):

PR #5202, opened 2019-11-05
Documentation cleanup for the Hibernate Search extension

yrodiere:

PR #5203, opened 2019-11-05
Configuration structure improvements in the Hibernate Search extension

yrodiere:
~Based on #5202 , which should be merged first.~ => Done

1. Remove the "elasticsearch" prefix for the "query-loading" and "automatic-indexing" properties, which are not specific to Elasticsearch or to a backend in particular. This is consistent with what we already do for "default-backend" and "background-failure-handler".
2. Clearly separate the configuration of the default backend in the documentation by moving it to a separate section.
3. Move the "default-backend" property to the "additional backends" section of the documentation, since it's only useful in that case.

Please have a look at the documentation before/after to make your mind about this changeset.
Run this:

```
./mvnw clean install -DskipTests -pl docs,extensions/hibernate-search-elasticsearch/runtime,extensions/hibernate-search-elasticsearch/deployment -am && xdg-open docs/target/generated-docs/hibernate-search-elasticsearch.html
```

And compare the structure of the documentation with what we already have [here](https://quarkus.io/guides/hibernate-search-elasticsearch#configuration-reference).

### DIFF
--- a/docs/src/main/asciidoc/hibernate-search-elasticsearch.adoc
+++ b/docs/src/main/asciidoc/hibernate-search-elasticsearch.adoc
@@ -531,9 +531,9 @@ quarkus.hibernate-orm.sql-load-script=import.sql <4>
 
 quarkus.hibernate-search.elasticsearch.version=7 <5>
 quarkus.hibernate-search.elasticsearch.analysis.configurer=org.acme.hibernate.search.elasticsearch.config.AnalysisConfigurer <6>
-quarkus.hibernate-search.elasticsearch.automatic-indexing.synchronization.strategy=searchable <7>
-quarkus.hibernate-search.elasticsearch.index-defaults.lifecycle.strategy=drop-and-create <8>
-quarkus.hibernate-search.elasticsearch.index-defaults.lifecycle.required-status=yellow <9>
+quarkus.hibernate-search.elasticsearch.index-defaults.lifecycle.strategy=drop-and-create <7>
+quarkus.hibernate-search.elasticsearch.index-defaults.lifecycle.required-status=yellow <8>
+quarkus.hibernate-search.automatic-indexing.synchronization.strategy=searchable <9>
 ----
 <1> We won't use SSL so we disable it to have a more compact native executable.
 <2> Let's create a PostgreSQL datasource.
@@ -543,12 +543,12 @@ quarkus.hibernate-search.elasticsearch.index-defaults.lifecycle.required-status=
 It is important because there are significant differences between Elasticsearch mapping syntax depending on the version.
 Since the mapping is created at build time to reduce startup time, Hibernate Search cannot connect to the cluster to automatically detect the version.
 <6> We point to the custom `AnalysisConfigurer` which defines the configuration of our analyzers and normalizers.
-<7> This means that we wait for the entities to be searchable before considering a write complete.
-While, on a production setup, the `committed` default might be a suitable value, using `searchable` is especially important when testing as you need the entities to be searchable immediately.
-<8> Obviously, this is not for production: we drop and recreate the index every time we start the application.
-<9> We consider the `yellow` status is sufficient to connect to the Elasticsearch cluster.
+<7> Obviously, this is not for production: we drop and recreate the index every time we start the application.
+<8> We consider the `yellow` status is sufficient to connect to the Elasticsearch cluster.
 This is for testing purposes with the Elasticsearch Docker container.
 It should not be used in production.
+<9> This means that we wait for the entities to be searchable before considering a write complete.
+While, on a production setup, the `committed` default might be a suitable value, using `searchable` is especially important when testing as you need the entities to be searchable immediately.
 
 [TIP]
 For more information about the Hibernate Search extension configuration please refer to the <<configuration-reference, Configuration Reference>>.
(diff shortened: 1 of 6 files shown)


yrodiere:
Rebased on master just now.

Sanne (approved):
Great, thanks! Merging..

gsmet:
Please let me review the Search PRs before merging them.

docs/src/main/asciidoc/hibernate-search-elasticsearch.adoc
```diff
@@ -531,9 +531,9 @@ quarkus.hibernate-orm.sql-load-script=import.sql <4>
 
 quarkus.hibernate-search.elasticsearch.version=7 <5>
 quarkus.hibernate-search.elasticsearch.analysis.configurer=org.acme.hibernate.search.elasticsearch.config.AnalysisConfigurer <6>
-quarkus.hibernate-search.elasticsearch.automatic-indexing.synchronization.strategy=searchable <7>
-quarkus.hibernate-search.elasticsearch.index-defaults.lifecycle.strategy=drop-and-create <8>
-quarkus.hibernate-search.elasticsearch.index-defaults.lifecycle.required-status=yellow <9>
+quarkus.hibernate-search.elasticsearch.index-defaults.lifecycle.strategy=drop-and-create <7>
+quarkus.hibernate-search.elasticsearch.index-defaults.lifecycle.required-status=yellow <8>
+quarkus.hibernate-search.automatic-indexing.synchronization.strategy=searchable <9>
```
gsmet: TBH, I'm not totally sold on this change. It makes the configuration all weird.

I'm also surprised it's not something you can configure per backend in Search.

gsmet:
This is a breaking change, we should be careful about that, I removed the backport tag, I don't want to do that just after a CR.

docs/src/main/asciidoc/hibernate-search-elasticsearch.adoc
yrodiere: Synchronization is about:

1. Whether to force a refresh on the backend side before considering indexing "done" (`committed` vs. `searchable`).
2. Whether we wait for indexing to be "done" after a transaction, or we just proceed without one works are queued (`queued` vs. `committed`/`searchable`).

Number 1 could theoretically be configured per backend, or even per index, though I don't really see the point.
Number 2 definitely cannot.

Maybe at some point we could split that into two separate configuration options, number 1 being configurable per backend while number 2 would not. However, that would introduce nonsensical combinations (do not wait for indexing to be "done", but force a refresh before considering it's "done"...). I also think it would make it even harder to explain the whole concept, but maybe that's just me.

Sanne:
@gsmet the general attitude so far has been that breaking changes _have to_ be backported before the final - not the opposite.

> Please let me review the Search PRs before merging them.

hu? now that's confusing

gsmet:
You don't backport breaking changes between a CR and a final, or you do another CR. We don't plan to do another CR so... At least, that's how I learned to do things :).

> hu? now that's confusing

why is it? This one has breaking changes that needs to be carefully thought out and I'm the one who worked on it initially.

Having it merged silently without adding `noteworthy-feature` was an issue as it could break users application/testing. This configuration knob is important.

docs/src/main/asciidoc/hibernate-search-elasticsearch.adoc
gsmet: What bugs me a bit tbh is having this property outside of the `elasticsearch` context.

The other ones we had there were for advanced configurations. Here the default configuration looks a bit asymmetrical.

Sanne:
> why is it? This one has breaking changes that needs to be carefully thought out and I'm the one who worked on it initially.

because the PR is sent to the team, and we're one team - everyone trying to help.

Mind you, I'm happy to follow your suggestion but I suspect tomorrow someone else will merge this, trespassing your territory again.

If you'd like to change the structure, I would support you. I already hinted at this: there's a lot of noise trying to follow ALL changes for everyone, I don't see that scaling.

But also I regularly get my PRs reviewed (and even merged) by people out of the same expert zone; one needs to learn accepting that a PR is not a draft, it's a bitter-sweet situation as I'm also not happy that lots of stuff I hoped would be done differently got merged before I could even have look.

It's a consequence of the current process; if you want to "own" all changes on an extension, you'll have to make this clear upfront as I can't see this stopping. Or ask Yoann to work with you on your own fork.

gsmet:
When there is a breaking change, the maintainer of the extension should at least have a look at the PR.

Another PR was merged on the HV extension while I was away without giving me a chance to review it while I explicitly asked to (I assigned it to me and posted a comment about it). The PR broke CI and then people are asking me to fix that.

I'm not talking about simple/obvious bugfixes or documentation fixes. This PR breaks the compatibility for a configuration property that is probably used by all the persons using this extension (as it's strictly needed to have tests working reliably).

And there was a child PR on the quickstarts that you didn't merge (https://github.com/quarkusio/quarkus-quickstarts/pull/359) so it broke the quickstarts for several days and I was the one people pinged about it.

I'm not asking for hard rules, just some common courtesy so that a person familiar with an extension has a look before merging something structuring. And this change is.

Typically, I wouldn't merge anything not obvious on the Spring-compatibility extensions without asking Georgios to have a look first.

Sanne:
> Another PR was merged on the HV extension while I was away without giving me a chance to review it

Yes that's what I'm referring to. I'm concerned as well, and think we need to change the process.

Comments on github won't do the trick, as you've seen on the other PRs.

PR #5205, opened 2019-11-05
Upgrade Gradle wrapper version to 5.6.4

gsmet:
@gastaldi could you test this when you get a chance?

### DIFF
--- a/build-parent/pom.xml
+++ b/build-parent/pom.xml
@@ -53,7 +53,7 @@
         <!-- These 2 properties are used by CreateProjectMojo to add the Maven Wrapper -->
         <proposed-maven-version>3.6.2</proposed-maven-version>
         <maven-wrapper.version>0.7.6</maven-wrapper.version>
-        <gradle-wrapper.version>5.5.1</gradle-wrapper.version>
+        <gradle-wrapper.version>5.6.4</gradle-wrapper.version>
 
         <!-- MicroProfile TCK versions -->
         <microprofile-health-api.version>2.1</microprofile-health-api.version>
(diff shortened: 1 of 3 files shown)


gastaldi (commented):
This doesn't change the Gradle wrapper used in new projects. You need to update the `<gradle-wrapper.version>` in `build-parent/pom.xml` too

gsmet:
Oh it's not using the tools to generate the new projects?

gsmet:
why do we have this variable in the tools, then?

gastaldi:
@aloubyansky may confirm but I think that's still a work in progress

gsmet:
@gastaldi thanks for noticing, I force-pushed an update.

gastaldi (approved):

gsmet:
@gastaldi can you test that creating a new Gradle project with it works and the project works too?

gastaldi:
Yes, I tested that before approving this PR :smile:  I can confirm it works and the project works too

gsmet:
Let's say you are suspiciously fast ;). Thanks then!

gwenneg:
Shouldn't [this file](https://github.com/quarkusio/quarkus/blob/master/devtools/gradle/gradle/wrapper/gradle-wrapper.properties) be updated too?

gsmet:
Yeah, it's the version we need to build. Initially I just wanted to update the version used in the created projects.

But it's nice to have a reference for when to upgrade Gradle. So I included everything in this PR.

Thanks everyone!

gwenneg (approved):

PR #5216, opened 2019-11-05
Replace gradle-tooling-api with gradleApi()

### LINKED ISSUE #5204: Gradle plugin does not work with plugins DSL
**Describe the bug**
As described in https://plugins.gradle.org/plugin/io.quarkus the preferred way to include the Quarkus Gradle plugin should be
```
plugins {
    id "io.quarkus" version "0.28.0"
}
```

**Expected behavior**
The Gradle plugins API should work for Quarkus.

**Actual behavior**
```
FAILURE: Build failed with an exception.

* What went wrong:
A problem occurred configuring root project 'extension-registry'.
> Could not resolve all artifacts for configuration ':classpath'.
   > Could not find org.gradle:gradle-tooling-api:5.5.1.
     Searched in the following locations:
       - https://plugins.gradle.org/m2/org/gradle/gradle-tooling-api/5.5.1/gradle-tooling-api-5.5.1.pom
     Required by:
         project : > io.quarkus:io.quarkus.gradle.plugin:0.28.0 > gradle.plugin.io.quarkus:quarkus-gradle-plugin:0.28.0
```

It works only when using the [legacy plugin syntax](https://plugins.gradle.org/plugin/io.quarkus).

**To Reproduce**
1. Create a Gradle project.
2. Add the `"io.quarkus"` plugin as described above.
3. `./gradlew build`

**Environment**
 - Output of `uname -a`: 
```
Linux ws-b83fff23-645e-4022-baec-7cf6c437f8cc 4.14.145+ #1 SMP Tue Oct 8 03:03:11 PDT 2019 x86_64 x86_64 x86_64 GNU/Linux
```
 - Output of `java -version`: 
```
openjdk version "11.0.2" 2019-01-15 LTS
OpenJDK Runtime Environment Zulu11.29+3-CA (build 11.0.2+7-LTS)
OpenJDK 64-Bit Server VM Zulu11.29+3-CA (build 11.0.2+7-LTS, mixed mode)
```
 - Quarkus version: `0.28.0`
 - Gradle version: `5.6.3`

gastaldi:
As reported in #5204

### DIFF
--- a/devtools/gradle/build.gradle
+++ b/devtools/gradle/build.gradle
@@ -27,7 +27,7 @@ dependencies {
     compile "io.quarkus:quarkus-platform-descriptor-resolver-json:${version}"
     compile "io.quarkus:quarkus-development-mode:${version}"
     compile "io.quarkus:quarkus-creator:${version}"
-    compile "org.gradle:gradle-tooling-api:5.5.1"
+    compile gradleApi()
 
     testImplementation 'org.assertj:assertj-core:3.13.2'
     testImplementation 'org.junit.jupiter:junit-jupiter-api:5.5.2'
--- a/devtools/gradle/gradle/wrapper/gradle-wrapper.properties
+++ b/devtools/gradle/gradle/wrapper/gradle-wrapper.properties
@@ -1,5 +1,6 @@
+#Tue Nov 05 11:34:08 BRST 2019
+distributionUrl=https\://services.gradle.org/distributions/gradle-5.6.4-all.zip
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-5.6.4-bin.zip
-zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
+zipStoreBase=GRADLE_USER_HOME


geoand (commented):
I won't have time to test this, but it does look reasonable. I'll let someone that can actually test it approve

gsmet:
@maxandersen or @aloubyansky could you have a look?

gsmet:
@gastaldi why is the Arquillian bump in this PR?

gastaldi:
@gsmet my fault, messed with my Git branches locally. It should be fixed now

geoand:
Are you sure this fixes #5204 ?

I tried built the PR and when I changed 

```
plugins {
    id 'java'
}

apply plugin: 'io.quarkus'
```

to 

```
plugins {
    id 'java'
    id 'io.quarkus' version "${quarkusVersion}"
}
```

then `./gradlew quarkusBuild` fails with:

```
A problem occurred evaluating root project 'gradle-test'.
> Could not find method nativeTestImplementation() for arguments [io.quarkus:quarkus-junit5] on object of type org.gradle.api.internal.artifacts.dsl.dependencies.DefaultDependencyHandler.
```

gastaldi:
@geoand I see the same error, in my case I removed the `nativeTestImplementation` calls and it passed. 

Perhaps we should also remove that from the generated project template?

geoand:
@gastaldi If we do that then the native tests probably won't work because their dependencies probably won't be setup properly.

I'm really curious as to what causes this strange behavior...

gastaldi:
The native tests still ran and working after I removed that, it seems to use what's defined in `testImplementation` too.

geoand:
So in any case the PR isn't going to break things more, but as it is it's not going to #5204.
Your suggestion does make sense, but I'm a little concerned about what could happen down the road if we remove it

maxandersen:
yes, I would say lets not touch gradle more now before post v1 and do #5101 properly - with actual real tests and not near a release frenzy :)

gastaldi:
Right, so this PR only fixes the dependency resolution issue for `org.gradle:gradle-tooling-api:5.5.1` (which doesn't seem the correct way to depend on the Gradle API inside a Gradle build file).

I created this project with the minimal Gradle configuration to make it work: https://github.com/gastaldi/gradle-quarkus-project/

geoand:
I agree with the PR, just not about it saying that it fixes the issue.
It's IMHO definitely a fix we want now

gastaldi:
Cool, I changed the commit message to not automatically close the issue

geoand (approved):
Cool, thanks!

PR #5217, opened 2019-11-05
fix: Dev mode - 404 Page - Live reloaded list of static files

### LINKED ISSUE #5201: Dev mode - 404 - List of static files not reloaded
Launch dev mode, go to an unknown page, you get the 404 with a list of static html files.

Create a new html file, reload the 404, the file is not there.

machi1990:
/cc @gsmet 

Fixes #5201

### DIFF
--- a/core/devmode/src/main/java/io/quarkus/dev/RuntimeUpdatesProcessor.java
+++ b/core/devmode/src/main/java/io/quarkus/dev/RuntimeUpdatesProcessor.java
@@ -342,7 +342,6 @@ public class RuntimeUpdatesProcessor implements HotReplacementContext {
                 } catch (IOException e) {
                     log.error("Failed to copy resources", e);
                 }
-
             }
 
             for (String path : watchedFilePaths.keySet()) {
@@ -354,7 +353,7 @@ public class RuntimeUpdatesProcessor implements HotReplacementContext {
                         if (value > existing) {
                             ret.add(path);
                             log.infof("File change detected: %s", file);
-                            if (doCopy) {
+                            if (doCopy && !Files.isDirectory(file)) {
                                 Path target = classesDir.resolve(path);
                                 byte[] data = Files.readAllBytes(file);
                                 try (FileOutputStream out = new FileOutputStream(target.toFile())) {
--- a/core/devmode/src/main/java/io/quarkus/dev/RuntimeUpdatesProcessor.java
+++ b/core/devmode/src/main/java/io/quarkus/dev/RuntimeUpdatesProcessor.java
@@ -6,9 +6,11 @@ import static java.util.stream.Collectors.toList;
 import java.io.File;
 import java.io.FileOutputStream;
 import java.io.IOException;
+import java.io.UncheckedIOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.nio.file.attribute.FileTime;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashSet;
@@ -249,7 +251,7 @@ public class RuntimeUpdatesProcessor implements HotReplacementContext {
                 }
             }
         } catch (IOException e) {
-            throw new RuntimeException(e);
+            throw new UncheckedIOException(e);
         }
 
         return hasChanges;
@@ -363,7 +365,7 @@ public class RuntimeUpdatesProcessor implements HotReplacementContext {
                             watchedFileTimestamps.put(file, value);
                         }
                     } catch (IOException e) {
-                        throw new RuntimeException(e);
+                        throw new UncheckedIOException(e);
                     }
                 } else {
                     watchedFileTimestamps.put(file, 0L);
@@ -371,7 +373,7 @@ public class RuntimeUpdatesProcessor implements HotReplacementContext {
                     try {
                         Files.deleteIfExists(target);
                     } catch (IOException e) {
-                        throw new RuntimeException(e);
+                        throw new UncheckedIOException(e);
                     }
                 }
             }
@@ -405,7 +407,7 @@ public class RuntimeUpdatesProcessor implements HotReplacementContext {
 
             return false;
         } catch (IOException e) {
-            throw new RuntimeException(e);
+            throw new UncheckedIOException(e);
         }
     }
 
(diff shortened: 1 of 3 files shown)


extensions/resteasy/deployment/src/test/java/io/quarkus/resteasy/test/NotFoundExceptionMapperTestCase.java
```diff
@@ -45,4 +53,46 @@ public void testJsonResourceNotFound() {
                 .contentType(ContentType.JSON);
     }
 
+    @Test
+    public void shouldDisplayNewAddedFileIn404ErrorPage() {
+        String CONTENT = "html content";
+        test.addResourceFile(META_INF_RESOURCES + "index2.html", CONTENT);
+
+        RestAssured.get("/index2.html")
```
aguibert: Is any sort of wait needed here between adding the resource file and invoking the `GET /index2.html` to ensure that dev mode has processed the new file?

machi1990: Hi @aguibert. No, it is not needed. Changes will be picked detected and picked automatically as indicated in the line 63 and 64.

aguibert (approved):
LGTM

machi1990:
@gwenneg would also gave some time to review this? Related to improving the not found page in dev mode as #5235 thanks.

gwenneg:
Sure, I'll take a look at it tonight.

core/devmode/src/main/java/io/quarkus/dev/RuntimeUpdatesProcessor.java
```diff
@@ -354,7 +353,7 @@ private String getFileExtension(File file) {
                         if (value > existing) {
                             ret.add(path);
                             log.infof("File change detected: %s", file);
-                            if (doCopy) {
+                            if (doCopy && !Files.isDirectory(file)) {
```
gwenneg: It makes sense to add this condition for files, but it also makes me wondering if directories would need to be created. WDYT?

core/devmode/src/main/java/io/quarkus/dev/RuntimeUpdatesProcessor.java
```diff
@@ -354,7 +353,7 @@ private String getFileExtension(File file) {
...
                                 Path target = classesDir.resolve(path);
                                 byte[] data = Files.readAllBytes(file);
```
gwenneg: That's not part of your PR and I'm not even sure that's a problem since static files are usually small, but this won't work for large files. Is that an acceptable risk? There are two occurences of this method call in this class.

extensions/resteasy/deployment/src/test/java/io/quarkus/resteasy/test/NotFoundExceptionMapperTestCase.java
```diff
@@ -45,4 +53,46 @@ public void testJsonResourceNotFound() {
...
+    @Test
+    public void shouldNotDisplayDeletedFileIn404ErrorPage() {
+        String TEST_CONTENT = "test html content";
+        test.addResourceFile(META_INF_RESOURCES + "test.html", TEST_CONTENT);
+
+        RestAssured
+                .get("/test.html")
+                .then()
+                .statusCode(200)
+                .body(containsString(TEST_CONTENT)); // check that another.html is reloaded
```
gwenneg:
```suggestion
                .body(containsString(TEST_CONTENT)); // check that test.html is live reloaded
```

gwenneg (commented):
That's unrelated with your changes, but I've seen a few `IOException` being caught and rethrown encapsulated into a `RuntimeException` in `RuntimeUpdatesProcessor`. It would be nice to throw an `UncheckedIOException` instead while we're at it.

extensions/resteasy/deployment/src/test/java/io/quarkus/resteasy/test/NotFoundExceptionMapperTestCase.java
machi1990: Good catch.

core/devmode/src/main/java/io/quarkus/dev/RuntimeUpdatesProcessor.java
machi1990: A directory is only watched for changes. 

It will be created when coping its files resource files here - https://github.com/quarkusio/quarkus/blob/27381651e25fd30febf3017e04c9a73c96273aaf/core/devmode/src/main/java/io/quarkus/dev/RuntimeUpdatesProcessor.java#L323

core/devmode/src/main/java/io/quarkus/dev/RuntimeUpdatesProcessor.java
machi1990: Yes, the static file will have to be bigger than 2GB. I bet it is safe to keep it this way for now.

machi1990:
> That's unrelated with your changes, but I've seen a few `IOException` being caught and rethrown encapsulated into a `RuntimeException` in `RuntimeUpdatesProcessor`. It would be nice to throw an `UncheckedIOException` instead while we're at it.

Yes, let me do it in a separate refactoring commit.

extensions/resteasy/deployment/src/test/java/io/quarkus/resteasy/test/NotFoundExceptionMapperTestCase.java
machi1990: done

machi1990:
@gwenneg thanks for the review. I pushed a second refactoring commit - WDYT?

core/devmode/src/main/java/io/quarkus/dev/RuntimeUpdatesProcessor.java
gwenneg: If I'm not mistaken, the exact value depends on the heap size. But we're talking about dev mode and a very unlikely static file size, I agree to leave it that way.

gwenneg (approved):
LGTM, thanks!

gsmet:
In the end I decided to not backport it. It's not critical enough to justify touching the hot reload logic now.

machi1990:
Thanks @gsmet for the update. It will give me time (during the weekend) to look on why the test case added was failing in JDK 8, as pointed out by @dmlloyd here https://github.com/quarkusio/quarkus/issues/5424.

PR #5220, opened 2019-11-05
RESTEasy http performance

Sanne:
A couple minor optimisations in VertxHttpRequest based on profiling data.

Also, removed some unused fields.

### DIFF
--- a/extensions/resteasy/runtime/src/main/java/io/quarkus/resteasy/runtime/standalone/VertxRequestHandler.java
+++ b/extensions/resteasy/runtime/src/main/java/io/quarkus/resteasy/runtime/standalone/VertxRequestHandler.java
@@ -2,6 +2,7 @@ package io.quarkus.resteasy.runtime.standalone;
 
 import java.io.IOException;
 import java.io.InputStream;
+import java.util.function.Supplier;
 
 import javax.enterprise.inject.Instance;
 import javax.enterprise.inject.spi.CDI;
@@ -96,12 +97,15 @@ public class VertxRequestHandler implements Handler<RoutingContext> {
             HttpServerResponse response = request.response();
             VertxHttpResponse vertxResponse = new VertxHttpResponse(request, dispatcher.getProviderFactory(),
                     request.method(), allocator, output);
-            // client address may not be available with VirtualHttp
-            SocketAddress socketAddress = request.remoteAddress();
-            String host = socketAddress != null ? socketAddress.host() : null;
+            // client address may not be available with VirtualHttp;
+            // using a supplier to make the remote Address resolution lazy: often it's not needed.
+            Supplier<String> hostNameProvider = () -> {
+                SocketAddress socketAddress = request.remoteAddress();
+                String host = socketAddress != null ? socketAddress.host() : null;
+                return host;
+            };
 
-            VertxHttpRequest vertxRequest = new VertxHttpRequest(ctx, headers, uriInfo, request.rawMethod(),
-                    host, dispatcher.getDispatcher(), vertxResponse, false);
+            VertxHttpRequest vertxRequest = new VertxHttpRequest(ctx, headers, uriInfo, request.rawMethod(), hostNameProvider, dispatcher.getDispatcher(), vertxResponse, false);
             vertxRequest.setInputStream(is);
             try {
                 ResteasyContext.pushContext(SecurityContext.class, new QuarkusResteasySecurityContext(request));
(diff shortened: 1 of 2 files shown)


gsmet (approved):
Looks good to me but I would like @stuartwdouglas to check the `isFlushed()` method shouldn't have been used somehow.

gsmet:
Test failures doesn't look related, I asked for a recheck.

Sanne:
Is the need for @stuartwdouglas 's opinion blocking?  I'd suggest it shouldn't, as while it indeed looks fishy, that was definitely dead code.  I can open a different issue to have him have a look as independent task.

Or if you'd rather wait for his feedback first I could split this PR so to get the performance fix already :)

Sanne:
For more context: making that address field lazy makes a significant difference ;)

patriot1burke:
LGTM

geoand:
Is there a reason why I can't rerun the failed tests? I used to be able to :)

gsmet:
It comes and go. Sometimes I don't have the buttons either.

geoand:
Gotcha, thanks!

extensions/resteasy/runtime/src/main/java/io/quarkus/resteasy/runtime/standalone/VertxRequestHandler.java
```diff
@@ -96,12 +97,16 @@ private void dispatch(RoutingContext routingContext, InputStream is, VertxOutput
             HttpServerResponse response = request.response();
             VertxHttpResponse vertxResponse = new VertxHttpResponse(request, dispatcher.getProviderFactory(),
                     request.method(), allocator, output);
-            // client address may not be available with VirtualHttp
-            SocketAddress socketAddress = request.remoteAddress();
-            String host = socketAddress != null ? socketAddress.host() : null;
+            // client address may not be available with VirtualHttp;
+            // using a supplier to make the remote Address resolution lazy: often it's not needed.
+            Supplier<String> hostNameProvider = () -> {
```
stuartwdouglas: We try and avoid lambdas in the runtime part of the code as they are much more costly in terms of metaspsace than an anonymous class. 

If this is expensive shouldn't the result of this be cached either in the Supplier or in VertxHttpRequest in case getRemoteAddress is called multiple times?

stuartwdouglas (changes requested):

extensions/resteasy/runtime/src/main/java/io/quarkus/resteasy/runtime/standalone/VertxRequestHandler.java
Sanne: ok I can change the lambda to an anonymous class. I know the metaspace cost, TBH I gave up on avoiding them in runtime code as we have many of them anyway, I had the impression nobody except me was still following that recommendation :)

It's not clear to me if it's worth caching the result. I can do that "just in case", but FYI the getter is not invoked during hot-path code in the benchmarks I've run: I verified this by initially running a modified version which was throing an exception.
I had to remove the exception-throwing code as some tests would fail, but it seems seldomly used - I don't know for sure though I guess I could have seen an happy path in my runs.

Sanne: P.S. the reason I ask if it's worth caching, is because `null` is a valid value - so to cache it I will need to add multiple additional fields.

stuartwdouglas: Maybe don't bother then, it is probably not a big deal.

gsmet: @Sanne Just for the record, you're not the only one avoiding lambdas in runtime code. We have some in build time but we really try to be careful for the runtime part.

But certainly some have slipped along the way.

geoand: I can confirm what @gsmet mentions from my part as well.
We have instructed contributors multiple times to avoid them.
Of course some can sneak in 😁

Sanne: I mean there's plenty in the libraries we use - even if we avoid them in `io.quarkus.*` code ...

But I agree with the principle of not opting for a lambda without a strong reason, I'll be more careful. Thanks!

Issue #3343, opened 2019-07-26
Quarkus MongoDB Client: Connect to multiple MongoDB clusters from the same app


csotiriou:
**Description**
Reading the guide about mongo here: https://quarkus.io/guides/mongo-guide there doesn't seem to be a straightfoward way of connecting to two clusters simultaneously in the same app. 

Reading the code, I couldn't find this feature anywhere - unless I am missing something.

I would expect a configuration like the Quarkus Rest Client (https://quarkus.io/guides/mongo-guide), something maybe like:

```
quarkus.mongodb.cluster1.connection-string = mongodb://mongo1:27017,mongo2:27017
quarkus.mongodb.cluster2.connection-string = mongodb://mongo1:27017,mongo2:27017
```

And then use annotations above the instances of MongoClient to indicate which instance we are going to use (example imaginary `@Connection` annotation used):

```
@Connection("cluster1")
@Inject
ReactiveMongoClient mongoClientCluster1
```

gsmet:
Yes, that would be doable. A good example of that is the `agroal` extension where we support multiple datasources.

I think it's a good issue for a beginner as the `agroal` extension should . Would you be interested in taking a look at it?

MMarus:
@gsmet @csotiriou Hello guys is someone working on this? Could I take a stab at this one? Thank you for the answer :).

csotiriou:
I will not be able to touch this for at least a month, due to heavy workload, so at least from my part I am not working on it.

geoand:
@MMarus I don't see anyone assigned to it or anyone mention they are working on it, so if you would like to take it up, feel free :)

MMarus:
@geoand Great, thank you very much :).

dufoli:
hello, I have start to work on this one. It was pretty complicated because I do not understand all the concept in quarkus right now. But I have start to work on it. How Can I tag my branch to this bug and get some help ?

dufoli:
here is the branch https://github.com/dufoli/quarkus/tree/mongo-cluster

geoand:
Hi, you can create a draft Pull Request where you mention that it fixes this issue.
In that PR you describe the details of what you need help with, what might not be working currently, what you think the limitations of your approach are etc.

dufoli:
I have commit and send the draft PR with some fixes.
I just need help to fix one remaining test because error is really strange to me.
CollectionManagementTest

dufoli:
[...]

PR #5222, opened 2019-11-05
Add health checks for Mongodb, Kafka and Neo4j

### LINKED ISSUE #5073: Provides extension HealthCheck
**Description**
Provides health check for some of the existing extensions.
It will not be exhaustive, please advise if some other extensions deserve an health check in the same batch of health check.

**Implementation ideas**

### MongoDB Health Check
Provides a **readiness** health check for MongoDB.
Proposed implementation: use the `MongoClient` to list the databases.
**Enabled** by default.

### Neo4J Health Check
Provides a **readiness** health check for Neo4J.
Proposed implementation: use the Bolt `Driver` to make a query to Neo4J.
**Enabled** by default.

### Kafka Health Check
Provides a **readiness** health check for Kafka.
Proposed implementation: use the Kafka `AdminClient` to connect to the cluster and check it's state.
**Disabled** by default.

_This is my opinion that as Kafka has built in resiliency for both producer and consumer, the health check should not be enabled by default. Some must have good arguments to enabled it by default and I'm open to discussion_

loicmathieu:
This fixes #5073

This is a draft PR as the Kafka health check is subject to discussion: 

- it instanciate a new AdminClient as the check is on `quarkus-kafka-client` that didn't provides any Kafka client inside CDI.
- it uses the (standard) property `kafka.bootstrap.servers` as a temporary solution : a better solution should be to scan all properties and find the one that defines a Kafka cluster (from mp-messaging or kafka-stream config) and creates an AdminClient for each.
- it is disabled by default because I think it's the best things to do (see the rational on the issue description).

When #3343 lands in, we must impact the MongoDB health check to check all the mongo clients.


gastaldi:
This looks great. 

Here is an idea: would it be possible to enable them automatically if the `quarkus-smallrye-health` extension is present? I think this would save the developer the trouble of having to define the `health=enabled` flag for these services.

gastaldi:
Ah ignore my idea, I think you already do that :) Awesome!

extensions/neo4j/deployment/src/main/java/io/quarkus/neo4j/deployment/Neo4jBuildTimeConfig.java
```diff
@@ -0,0 +1,14 @@
+package io.quarkus.neo4j.deployment;
+
+import io.quarkus.runtime.annotations.ConfigItem;
+import io.quarkus.runtime.annotations.ConfigPhase;
+import io.quarkus.runtime.annotations.ConfigRoot;
+
+@ConfigRoot(name = "neo4j", phase = ConfigPhase.BUILD_TIME)
+public class Neo4jBuildTimeConfig {
+    /**
+     * Whether or not an healt check is published in case the smallrye-health extension is present (default to true).
```
machi1990: Maybe we can remove the `(default to true)`  from the description. It will be automatically picked from ConfigItem during doc generation.

loicmathieu: Yes but the JavaDoc is also usefull for Quarkus developpers :)
Maybe I could but best is to make it after to keep all health check enabled comments the same (agroal, these 3 new ones, kafka-streams)

loicmathieu:
@gunnarmorling can you review the Kafka health check implementation and some question I raised inside the PR description ?

extensions/neo4j/runtime/src/main/java/io/quarkus/neo4j/runtime/heath/Neo4jHealthCheck.java
```diff
@@ -0,0 +1,28 @@
...
+    Driver driver;
+
+    @Override
+    public HealthCheckResponse call() {
+        HealthCheckResponseBuilder builder = HealthCheckResponse.named("Neo4j connection health check").up();
+        try {
+            driver.verifyConnectivity();
+            return builder.build();
+        } catch (Exception e) {
+            return builder.down().withData("reason", e.getMessage()).build();
```
jmartisk: Not that it matters much, so just nitpicking that it is a little weird if you first call `up()` and then change it to `down()`, you could call `up()` only after `verifyConnectivity()` passes. Otherwise a big +1 from me

loicmathieu:
Can someone review the Kafka part and gives feedback of the fact that I choose to disabled the health check by default (rational is in the issue description)? Maybe @gunnarmorling or @cescoffier ?

docs/src/main/asciidoc/kafka.adoc
```diff
@@ -308,6 +308,15 @@ You can build the native executable with:
 ./mvnw package -Pnative
 ----
 
+== Kafka Health Check
+
+If you are using the `quarkus-smallrye-health` extension, `quarkus-kafka` could automatically add a readiness health check
+to validate the connection to the broker. This is disabled by default.
```
gunnarmorling: Why is it disabled by default, wasn't the default to enable it?

extensions/kafka-client/deployment/src/main/java/io/quarkus/kafka/client/deployment/KafkaBuildTimeConfig.java
```diff
@@ -0,0 +1,15 @@
...
+
+import io.quarkus.runtime.annotations.ConfigItem;
+import io.quarkus.runtime.annotations.ConfigPhase;
+import io.quarkus.runtime.annotations.ConfigRoot;
+
+@ConfigRoot(name = "kafka", phase = ConfigPhase.BUILD_TIME)
+public class KafkaBuildTimeConfig {
+    /**
+     * Whether or not an health check is published in case the smallrye-health extension is present (default to false).
+     * If you enable the heath check, you must specify the `kafka.bootstrap.servers` property.
```
gunnarmorling:
```suggestion
     * If you enable the health check, you must specify the `kafka.bootstrap.servers` property.
```

extensions/kafka-client/deployment/src/main/java/io/quarkus/kafka/client/deployment/KafkaProcessor.java
```diff
@@ -145,4 +146,10 @@ public void write(final String name, final byte[] data) {
         }
         return new GeneratedClassBuildItem(false, TARGET_JAVA_9_CHECKSUM_FACTORY, holder[0]);
     }
+
+    @BuildStep
+    HealthBuildItem addHealthCheck(KafkaBuildTimeConfig buildTimeConfig) {
+        return new HealthBuildItem("io.quarkus.kafka.client.health.KafkaHealthCheck",
+                buildTimeConfig.healthEnabled, "datasource");
```
gunnarmorling: datasource?

extensions/neo4j/deployment/src/main/java/io/quarkus/neo4j/deployment/Neo4jBuildTimeConfig.java
gunnarmorling:
```suggestion
     * Whether or not an health check is published in case the smallrye-health extension is present (default to true).
```

extensions/kafka-client/runtime/src/main/java/io/quarkus/kafka/client/health/KafkaHealthCheck.java
```diff
@@ -0,0 +1,56 @@
...
+    private String bootstrapServer;
+
+    private AdminClient client;
+
+    @PostConstruct
+    void init() {
+        Map<String, Object> conf = new HashMap<>();
+        conf.put(AdminClientConfig.BOOTSTRAP_SERVERS_CONFIG, bootstrapServer);
+        conf.put(AdminClientConfig.REQUEST_TIMEOUT_MS_CONFIG, "5000");
+        client = AdminClient.create(conf);
```
gunnarmorling: How about creating the client (and closing it) only in `call()`? That way we don't hold this client for the entire lifetime of the application, also if it's actually not needed any longer once the container has received traffic and the readiness probe thus shouldn't matter any longer.

gunnarmorling (commented):
Nice work, @loicmathieu; a few comments on the Kafka part.

docs/src/main/asciidoc/kafka.adoc
loicmathieu: This is debatable :)

This is a quote from the issue description:
> This is my opinion that as Kafka has built in resiliency for both producer and consumer, the health check should not be enabled by default. Some must have good arguments to enabled it by default and I'm open to discussion

loicmathieu: I always disable Kafka Health Check from _other frameworks_ that why I propose to disabled it by default in Quarkus ... but I'm OK to enabled it by default if it is not a popular decision :)

extensions/kafka-client/runtime/src/main/java/io/quarkus/kafka/client/health/KafkaHealthCheck.java
loicmathieu: Typically, a readiness probe will be called every n seconds (5s for example) so I'm concern of the cost of recreating a client each time.

docs/src/main/asciidoc/kafka.adoc
loicmathieu: I always disable Kafka Health Check from other frameworks that why I propose to disabled it by default in Quarkus ... but I'm OK to enabled it by default if it is not a popular decision :)

loicmathieu: I always disable Kafka Health Check from other frameworks that why I propose to disabled it by default in Quarkus ... but I'm OK to enabled it by default if it is not a popular decision :)

gunnarmorling: I brought it up because it's different than what's suggested for MongoDB, so this caught my attention. I'm not 100% sure either. So what happens if a request comes in which triggers the sending of a Kafka message while the broker isn't accessible. Does this block the incoming request? I.e. does it make sense to have requests routed to that pod or not?

loicmathieu: It vastely depends of the way you use the `KafkaProducer` see the introduction JavaDoc: https://kafka.apache.org/20/javadoc/org/apache/kafka/clients/producer/KafkaProducer.html

The send method is asynchronous, so if you didn't wait for it inside your HTTP request, it will be proceed.

Depending on the configuration of acknowledge mode, retry and buffer size your message can be batched (so not send immediatly), retried or not. All this can lead to a not direct failure in case of sending a message.

So I would say that in the more common cases, a broker failure will lead to messages being kept inside the buffer of the `KafkaProducer` untill the broker is back and no failures of the incomming HTTP request that uses Kafak, nor any blocking.

As it can varies due to the way the user configure Kafka, it should be the user choice to enable the health check or not.

If the user uses quarkus-smallrye-reactive-messaging-kafka, by default the KafkaSink (https://github.com/smallrye/smallrye-reactive-messaging/blob/master/smallrye-reactive-messaging-kafka/src/main/java/io/smallrye/reactive/messaging/kafka/KafkaSink.java) didn't wait for the message to be produces (waitForWriteCompletion is true by default and we didn't override it inside Quarkus) so it should be OK.

extensions/kafka-client/runtime/src/main/java/io/quarkus/kafka/client/health/KafkaHealthCheck.java
loicmathieu: I looked into the Vert.x Kafka code and they hold the KafkaClient (inside the Sender) for the entire life of the application so I don't think something is wrong here. But I'm OK to reniew it each time if you prefer. Just tell me :)

extensions/neo4j/runtime/src/main/java/io/quarkus/neo4j/runtime/heath/Neo4jHealthCheck.java
loicmathieu: This constuct is inspired by the Agroal health check because of multiple datasource support. When we will support multiple MongoDB connection/ Kafka connection /  ... this construct will be needed (because you will fails the health check if one of the multiple connection fails).
So better keep it this way.

extensions/neo4j/deployment/src/main/java/io/quarkus/neo4j/deployment/Neo4jBuildTimeConfig.java
loicmathieu: done

extensions/kafka-client/deployment/src/main/java/io/quarkus/kafka/client/deployment/KafkaProcessor.java
loicmathieu: Should have been ` kafka` :) Fixed!

extensions/kafka-client/deployment/src/main/java/io/quarkus/kafka/client/deployment/KafkaBuildTimeConfig.java
loicmathieu: done

gsmet:
Just a heads up that I started a review of this one.

I'm working on the Qute PR today but will get back to this one soon.

loicmathieu:
@gunnarmorling WDYT regarding my latests comments on this PR?

extensions/kafka-client/runtime/src/main/java/io/quarkus/kafka/client/health/KafkaHealthCheck.java
gunnarmorling: Yeah, seems fine as is then. Thanks for checking with Vert.x.

docs/src/main/asciidoc/kafka.adoc
gunnarmorling: > didn't wait for the message to be produces (waitForWriteCompletion is true by default and we didn't override it inside Quarkus)

Isn't this self-contradictory?

gunnarmorling: > This is my opinion that as Kafka has built in resiliency for both producer and consumer, the health check should not be enabled by default.

Yes, so based on that I think disabling by default is reasonable.

extensions/kafka-client/deployment/src/main/java/io/quarkus/kafka/client/deployment/KafkaBuildTimeConfig.java
```diff
@@ -0,0 +1,15 @@
...
+
+import io.quarkus.runtime.annotations.ConfigItem;
+import io.quarkus.runtime.annotations.ConfigPhase;
+import io.quarkus.runtime.annotations.ConfigRoot;
+
+@ConfigRoot(name = "kafka", phase = ConfigPhase.BUILD_TIME)
+public class KafkaBuildTimeConfig {
+    /**
+     * Whether or not an health check is published in case the smallrye-health extension is present (default to false).
+     * If you enable the heatlh check, you must specify the `kafka.bootstrap.servers` property.
```
gunnarmorling:
```suggestion
     * If you enable the healtlh check, you must specify the `kafka.bootstrap.servers` property.
```

loicmathieu: fixed

loicmathieu:
I rebase on master to fix a conflict.

@gsmet did you have time to review it as you intended ?

gsmet:
I adjusted a few things.

gsmet (approved):

gsmet:
Merged, thanks.

loicmathieu:
@gsmet you move the bootstrap server property from the standard one to a quarkus specific one.
According to me it's an issue as the standard one is already used when using smallrye-reactive-messaging Kafka (as it uses standard producer and consumer properties).
So if a user uses the quarkus-smallrye-reactive-messaging-kafka extension it will needs to provides the boostrap-server in two properties: a quarkus one and a standard one.

zeljkot:
Having two separate properties for reactive streams and health is very confusing. The need for additional property in "Guides" can be easily overlooked. Please emphasize that in the documentation, or add all messaging bootstrap servers automatically. IMHO health check should not be general Kafka health check, but rather reactive messaging specific one; otherwise, whoever writes a Kafka client should somehow register it with health check functionality.

PR #5225, opened 2019-11-05
Add support for Spring Security's @Secured

aureamunoz:
This PR contains a new extension `spring-security` implementing `@Secured` annotation. 
It's based on https://github.com/quarkusio/quarkus/pull/5074 and is a draft.
I let you, @geoand and @michalszynkiewicz , take a look and forward to others if needed.
Thank you


gsmet:
@aureamunoz quick question, I haven't looked at the implementation: does it support the EL Spring Security has (things such as `hasRole()` and so on)? I suppose not. I wonder what would be necessary for us to support that because it's very useful in practice.

gsmet:
BTW, I should have started by that: it's an exciting new addition.

geoand:
Nice work!

To answer your question @gsmet, the stuff you describe I am pretty sure is covered by Spring's `@PreAuthorize` which I'll probably be picking up soon

NetoDevel:
@aureamunoz,

**Suggestion:** 

I don't know if it's possible, but it would be nice to have security configuration support.
Something like:

```
http
.authorizeRequests()
.antMatchers("/hello/**").allowAll()
```

geoand:
@NetoDevel we aren't planning to support that style of configuration at this time. That doesn't mean that we won't in the future, but just that it's not in the immediate plans.

NetoDevel:
@geoand, 
Ok, I'll follow this issue to understand and try to contribute in the future.
Thanks for answering me.

geoand:
You are welcome!

geoand:
@michalszynkiewicz @aureamunoz what is the status of #5074?

Is there anything I can do to help?

michalszynkiewicz:
@geoand imo done, Stuart asked if I can use index instead of looping throug all classes but AFAIK I can't

geoand:
Which part of your PR is the code in question @michalszynkiewicz ?

I can take a look on Monday if you like

michalszynkiewicz:
https://github.com/quarkusio/quarkus/pull/5074/files#diff-36e991a830b10f10b19a4360212eebcaR194

geoand:
I'll be looking at #5074 today

geoand:
@michalszynkiewicz Nice work!

I was thinking of implementing this in a slightly different manner however. The basic premise that I have in mind is that methods need to be "mapped" to a `SecurityCheck`.
With that in mind things like Spring's `@Secured` would not need to use the annotation transformer, the extension could just map methods to the appropriate `RolesAllowed` check.
There are probably a few mechanics involved to make this work properly but I will start from your PR and build on it to incorporate the solution I have in mind which I think should turn out to be simpler and avoid the need to go over all classes in the index.

geoand:
I'm going to try to come up with a way to do this that will also allow for the types of SecurityChecks that will be needed for Spring's `@PreAuthorize`

geoand:
I just updated the PR to align with the changes made in the core Quarkus security

geoand:
One thing that is missing is an integration test that will ensure this works in native mode

michalszynkiewicz:
@geoand maybe add it to the `main` module of the integration tests?  It already has security enabled

geoand:
@michalszynkiewicz we don't want to add Spring Security to the main module because that one should not depend on any Spring things. 
We could probably just add it to the Spring Web integration test in order to avoid having a completely new integration test that will take up a lot of time.
The reason I want an integration test for Spring security is that there could perhaps be problems with the Spring Security artifact that will only show up in native mode  - and that might require maven exclusions or GraalVM substitutions.

michalszynkiewicz:
@geoand if you ask me, we should have native tests for all the stuff. OTOH, they native test modules are rellay costly.
+1 from me for adding it to spring web integration tests if we have a module for them.

geoand:
We do have a spring Web integration test module so I propose @aureamunoz add them there

aureamunoz:
> We do have a Spring Web integration test module so I propose @aureamunoz add them there

Yes, of course. It's what I'm doing, sorry for not saying nothing until now :smirk:

geoand:
@aureamunoz can you please squash the second commit?
I don't think that in this case it makes sense to have a second one.
If you add a guide however, then that would make an excellent second commit

extensions/spring-security/deployment/src/main/java/io/quarkus/spring/security/deployment/SpringSecurityTransformerUtils.java
```diff
@@ -0,0 +1,46 @@
...
+
+import javax.annotation.security.RolesAllowed;
+
+import org.jboss.jandex.AnnotationInstance;
+import org.jboss.jandex.ClassInfo;
+import org.jboss.jandex.DotName;
+import org.jboss.jandex.MethodInfo;
+import org.springframework.security.access.annotation.Secured;
+
+public class SpringSecurityTransformerUtils {
```
geoand: Is this class needed? I don't think it's used, correct?

extensions/spring-security/deployment/src/test/java/io/quarkus/spring/security/deployment/SpringSecuredAnnotatedTest.java
```diff
@@ -0,0 +1,85 @@
...
+import io.quarkus.security.UnauthorizedException;
+import io.quarkus.security.test.cdi.SecurityTestUtils;
+import io.quarkus.security.test.utils.AuthData;
+import io.quarkus.security.test.utils.IdentityMock;
+import io.quarkus.spring.security.deployment.app.BeanWithSpringSecurityAnnotations;
+import io.quarkus.spring.security.deployment.app.BeanWithSpringSecurityMethodAnnotations;
+import io.quarkus.spring.security.deployment.app.SpringSecuredSubClass;
+import io.quarkus.test.QuarkusUnitTest;
+
+public class SpringSecuredAnnotatedTest {
```
geoand: +1

aureamunoz: What do you mean? :-)

geoand: That I like this update :)

geoand:
@aureamunoz will provide a follow up PR for docs

geoand (approved):
:+1: :+1: :+1:

bom/runtime/pom.xml
```diff
@@ -2434,6 +2440,33 @@
...
+                        <artifactId>spring-expression</artifactId>
+                    </exclusion>
+                    <exclusion>
+                        <groupId>org.springframework</groupId>
+                        <artifactId>spring-aop</artifactId>
+                    </exclusion>
+                    <exclusion>
+                        <groupId>org.springframework</groupId>
+                        <artifactId>spring-context</artifactId>
+                    </exclusion>
```
gsmet: Maybe we should just exclude `*:*` ? I suppose we just want the annotations and we don't need the runtime to work?

extensions/spring-security/deployment/pom.xml
```diff
@@ -0,0 +1,71 @@
...
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>io.quarkus</groupId>
+        <artifactId>quarkus-spring-security-parent</artifactId>
+        <version>999-SNAPSHOT</version>
+        <relativePath>../</relativePath>
+    </parent>
+
+    <artifactId>quarkus-spring-security-deployment</artifactId>
+    <name>Spring Security - Deployment</name>
```
gsmet:
```suggestion
    <name>Quarkus - Spring Security - Deployment</name>
```

extensions/spring-security/deployment/pom.xml
```diff
@@ -0,0 +1,71 @@
...
+            <exclusions>
+                <exclusion>
+                    <groupId>javax.enterprise</groupId>
+                    <artifactId>cdi-api</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-security-test-utils</artifactId>
```
gsmet: Shouldn't this be a `test` dependency.

extensions/spring-security/deployment/src/main/java/io/quarkus/spring/security/deployment/SpringSecurityProcessor.java
```diff
@@ -0,0 +1,74 @@
...
+import io.quarkus.deployment.annotations.BuildStep;
+import io.quarkus.deployment.builditem.ApplicationIndexBuildItem;
+import io.quarkus.deployment.builditem.FeatureBuildItem;
+import io.quarkus.security.deployment.AdditionalSecurityCheckBuildItem;
+import io.quarkus.security.deployment.SecurityCheckInstantiationUtil;
+import io.quarkus.spring.security.runtime.interceptor.SpringSecuredInterceptor;
+
+class SpringSecurityProcessor {
+
+    private static final String FEATURE = "spring-security";
```
gsmet: This should be in `FeatureBuildItem`. Please respect the alphabetical order.

extensions/spring-security/pom.xml
```diff
@@ -0,0 +1,21 @@
...
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <artifactId>quarkus-build-parent</artifactId>
+        <groupId>io.quarkus</groupId>
+        <version>999-SNAPSHOT</version>
+        <relativePath>../../build-parent/pom.xml</relativePath>
+    </parent>
+
+    <artifactId>quarkus-spring-security-parent</artifactId>
+    <name>Spring Security - Parent</name>
```
gsmet:
```suggestion
    <name>Quarkus - Spring Security - Parent</name>
```

extensions/spring-security/runtime/pom.xml
```diff
@@ -0,0 +1,35 @@
...
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>io.quarkus</groupId>
+        <artifactId>quarkus-spring-security-parent</artifactId>
+        <version>999-SNAPSHOT</version>
+        <relativePath>../</relativePath>
+    </parent>
+
+    <artifactId>quarkus-spring-security</artifactId>
+    <name>Spring Security - Runtime</name>
```
gsmet:
```suggestion
    <name>Quarkus - Spring Security - Runtime</name>
```

gsmet (dismissed):
I added a few comments. Could you have a look?

Thanks.

extensions/spring-security/deployment/src/main/java/io/quarkus/spring/security/deployment/SpringSecurityProcessor.java
```diff
@@ -0,0 +1,74 @@
...
+        registrars.produce(new InterceptorBindingRegistrarBuildItem(new SpringSecurityAnnotationsRegistrar()));
+        beans.produce(new AdditionalBeanBuildItem(SpringSecuredInterceptor.class));
+    }
+
+    @BuildStep
+    void addSpringSecuredSecurityCheck(ApplicationIndexBuildItem index,
+            BuildProducer<AdditionalSecurityCheckBuildItem> additionalSecurityCheckBuildItems) {
+
+        Set<MethodInfo> methodsWithSecurityAnnotation = new HashSet<>();
+        for (AnnotationInstance instance : index.getIndex().getAnnotations(DotNames.SPRING_SECURED)) {
```
michalszynkiewicz: it looks like we're assuming that noone mixes `@Secured` with the spec annotations. 
I think it would be good to validate if it's true before making this assumption.
Or not add the class-level annotation to methods that have these annotations.

geoand: I would say that at this point we should probably not complicate things with mix and match. We can maybe just log a warning if a spec annotation is used?

bom/runtime/pom.xml
gsmet: Didn't work, let's not do it.

extensions/spring-security/deployment/src/main/java/io/quarkus/spring/security/deployment/SpringSecurityProcessor.java
gsmet: +1 for the warning.

michalszynkiewicz: not sure if a warning is enough...
The current logic would add a duplicate security annotation to a method in such case, wouldn't it?

geoand: Well if people go an mix everything they can of course end up with weird situations. I think that a clear warning that Spring and spec annotations should not be mixed (on classes or methods) should be enough.
I don't think we want to start handling cases where Spring annotations are mixed with other annotations, unless Spring itself supports this. It could unecessarily complicate things I think.

michalszynkiewicz: agree on not supporting a mix.
Still not sure if warning is enough.
Do we even know what happens if a user has `@Secured("user") on the class and `@DenyAll` on its method?
I think it will fail in the runtime having two checks for a single method, but I'm not sure (this is how it worked before latest changes).
IMO a compile time error would be much better in such case.

geoand: A build time error sounds good indeed. Let's do that, OK @aureamunoz ?

aureamunoz: Ok for the build time error, let's do it.

aureamunoz:
@geoand you can take a look now :-)

geoand:
I force pushes to the PR, so my approval shouldn't count :).

@michalszynkiewicz can you please take a look?

extensions/spring-security/deployment/src/main/java/io/quarkus/spring/security/deployment/SpringSecurityProcessor.java
geoand: This was fixed

michalszynkiewicz:
It wouldn't hurt to have a test for the newly added check but that's not a big issue.
LGTM

michalszynkiewicz (approved):

geoand:
> It wouldn't hurt to have a test for the newly added check but that's not a big issue.
> LGTM

There is one test, indeed there could have been more :)

geoand:
Thanks for all the work @aureamunoz!

michalszynkiewicz:
> Thanks for all the work @aureamunoz !
:+1: good job!

> > It wouldn't hurt to have a test for the newly added check but that's not a big issue.
> > LGTM
> 
> There is one test, indeed there could have been more :)
sorry, I missed that.

geoand:
> sorry, I missed that.

No worries! 
Thanks again for your help!

PR #5237, opened 2019-11-06
Complete the Hibernate ORM guide, clarify the limitations

Sanne:
long overdue, finally clarifying the limitations we still have.

### DIFF
--- a/docs/src/main/asciidoc/hibernate-orm.adoc
+++ b/docs/src/main/asciidoc/hibernate-orm.adoc
@@ -11,22 +11,34 @@ include::./attributes.adoc[]
 Hibernate ORM is the de facto JPA implementation and offers you the full breath of an Object Relational Mapper.
 It works beautifully in Quarkus.
 
-== Setting up and configuring Hibernate ORM without `persistence.xml` (recommended)
+== Setting up and configuring Hibernate ORM
 
-More often than not, you need one _persistence unit_ with few configuration options.
+When using Hibernate ORM in Quarkus, you don't need to have a `persistence.xml` resource to configure it.
+
+Using such a classic configuration file is an option, but unnecessary unless you have specific advanced needs;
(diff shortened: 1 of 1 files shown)


docs/src/main/asciidoc/hibernate-orm.adoc
```diff
@@ -385,6 +431,85 @@ Finally, the second-level cache can be disabled globally by setting `hibernate.c
...
+
+Transaction Manager integration::
+You don't need to set this up, Quarkus automatically injects the reference to the Narayana Transaction Manager.
+The dependency is included automatically as a transitive dependency of the Hibernate ORM extension.
+All configuration is optional; for more details see link:transaction.html[Using Transactions in Quarkus].
+
+Connection pool::
+Don't need to choose one either. Quarkus automatically includes the Agroal connection pool;
+just configure your datasource as in the above examples and it will setup Hibernate ORM to use Agroal.
+More details about this connection pool can be found in link:datasource-guide.html[Quarkus - Datasources].
```
gsmet: This URL is outdated, it should point to: https://quarkus.io/guides/datasource

There is a redirect so it's no biggie but better get this straight.

gsmet (changes requested):
Very nice work, thanks a lot for that.

There's just one URL to adapt and it's good to go.

Thanks!

docs/src/main/asciidoc/hibernate-orm.adoc
Sanne: ah thanks!  well, without the `.html` too :)

gsmet: Yeah, I will make a pass on all the guides to remove the `.html`. That's why I didn't mention it.

Sanne:
@gsmet applied rebased, push forced.

I stripped all .html also from a couple other cross-guide links this had.

gsmet (approved):

PR #5238, opened 2019-11-06
Fix incorrect GraalVM option and use correct build path for Gradle resource config file

aguibert:


aguibert:
@geoand can you please review?

geoand (approved):

PR #5242, opened 2019-11-06
Upgrade to Neo4j driver 4.0.0-beta03.

michael-simons:
This version of our driver now includes all substitutions and neccessary instructions to make our driver work natively, so those substitutions are no longer necessary to have in quarkus. The commit removes all Netty substitutions and instructions for reflection access.

I'd like to take the opportunity to say thank you for the great discussions. In the end, our driver is better than before and every project that wants to compile it natively can benefit from something we started here.


gsmet (approved):

gsmet:
Thanks!

Issue #5215, opened 2019-11-05
Kafka native ITs are unstable


gsmet:
We have a lot of failures in the native Kafka tests lately.

I see this error:

```
2019-11-05 12:03:40,520 ERROR [io.qua.ver.htt.run.QuarkusErrorHandler] (vert.x-worker-thread-0) HTTP Request to /kafka failed, error id: 64c77992-c376-4f49-9187-9b6cd08432d5-1: org.jboss.resteasy.spi.UnhandledException: java.util.NoSuchElementException
	at org.jboss.resteasy.core.ExceptionHandler.handleApplicationException(ExceptionHandler.java:106)
	at org.jboss.resteasy.core.ExceptionHandler.handleException(ExceptionHandler.java:372)
	at org.jboss.resteasy.core.SynchronousDispatcher.writeException(SynchronousDispatcher.java:209)
	at org.jboss.resteasy.core.SynchronousDispatcher.invoke(SynchronousDispatcher.java:496)
	at org.jboss.resteasy.core.SynchronousDispatcher.lambda$invoke$4(SynchronousDispatcher.java:252)
	at org.jboss.resteasy.core.SynchronousDispatcher.lambda$preprocess$0(SynchronousDispatcher.java:153)
	at org.jboss.resteasy.core.interception.jaxrs.PreMatchContainerRequestContext.filter(PreMatchContainerRequestContext.java:363)
	at org.jboss.resteasy.core.SynchronousDispatcher.preprocess(SynchronousDispatcher.java:156)
	at org.jboss.resteasy.core.SynchronousDispatcher.invoke(SynchronousDispatcher.java:238)
	at io.quarkus.resteasy.runtime.standalone.RequestDispatcher.service(RequestDispatcher.java:73)
	at io.quarkus.resteasy.runtime.standalone.VertxRequestHandler.dispatch(VertxRequestHandler.java:109)
	at io.quarkus.resteasy.runtime.standalone.VertxRequestHandler.dispatchRequestContext(VertxRequestHandler.java:84)
	at io.quarkus.resteasy.runtime.standalone.VertxRequestHandler.lambda$handle$0(VertxRequestHandler.java:71)
	at io.vertx.core.impl.ContextImpl.lambda$executeBlocking$2(ContextImpl.java:316)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
[...]

PR #5245, opened 2019-11-06
Fix Kafka test to handle empty consumer records on poll

jaikiran:
This fixes the `java.util.NoSuchElementException` exceptions that are reported in the Kafka native tests, as noted here https://github.com/quarkusio/quarkus/issues/5215.
Of course, this doesn't mean the test will still pass (since that exception was a sign that there were no messages that were consumed by the consumer). So I bumped the _maximum_ poll time to `60000` milli seconds from the current `10000` milli seconds.


gsmet (approved):
Looks good, thanks @jaikiran . Let's see what CI has to say.

jaikiran:
@gsmet, I just added another commit to this PR. Since it was a very trivial change, I decided to use this PR itself instead of creating another one. Plus, this second commit too is related to #5215 issue.

gsmet:
OK, thanks for pointing this out. Looks ok to me, let's see what CI says.

gsmet:
Merged, thanks!

Issue #5247, opened 2019-11-06
0.28.1 -> amazon-lamda-http not working


Ravisankar-Challa:
I am getting below error when a basic hellowrold lambda deploy to AWS and invoked through API gateway. you can see the stacktrace in the attached image

![image](https://user-images.githubusercontent.com/11788842/68294071-45e5f380-00b5-11ea-87ba-a5c73add22d6.png)

jaikiran:
This looks like an issue in vertx itself where it can't handle unresolved `InetSocketAddress`. The only way for now to get past this is probably an ugly hack in Quarkus, until the fix is available in vertx. 
I'll propose a PR here and open an issue in vertx and see what gets decided.

jaikiran:
I've opened a PR which proposes a workaround https://github.com/quarkusio/quarkus/pull/5250

Ravisankar-Challa:
@patriot1burke

gsmet:
@Ravisankar-Challa can you try @jaikiran 's PR mentioned above?

Get the code from the PR, build with `mvn clean install -DskipTests -DskipITs` then use `999-SNAPSHOT` as a version.

Thanks

Ravisankar-Challa:
Tested with 999-SNAPSHOT now I am getting a proper response. Thanks for fixing.

machi1990:
Re-opening as the PR #5250 is not yet merged.

jaikiran:
Closing now since the PR got merged.

PR #5250, opened 2019-11-06
Work around an issue in vertx where getting remote/client SocketAddress of the request leads to a NPE

jaikiran:
The commit in this PR works around the vertx issue https://github.com/eclipse-vertx/vert.x/issues/3181.

The existing code in this class already seems capable of handling `null` value for a remote address, since it already had:
```
String host = socketAddress != null ? socketAddress.host() : null;
```
So I think it should be fine to consider `host` as `null` when we run into an exception and ignore it.

### DIFF
--- a/extensions/resteasy/runtime/src/main/java/io/quarkus/resteasy/runtime/standalone/LazyHostSupplier.java
+++ b/extensions/resteasy/runtime/src/main/java/io/quarkus/resteasy/runtime/standalone/LazyHostSupplier.java
@@ -2,10 +2,13 @@ package io.quarkus.resteasy.runtime.standalone;
 
 import java.util.Objects;
 
+import org.jboss.logging.Logger;
+
 import io.vertx.core.http.HttpServerRequest;
 import io.vertx.core.net.SocketAddress;
 
 final class LazyHostSupplier {
+    private static final Logger logger = Logger.getLogger(LazyHostSupplier.class.getName());
 
     private final HttpServerRequest request;
     private String remoteHost = null;
@@ -20,7 +23,17 @@ final class LazyHostSupplier {
         if (initialized) {
             return this.remoteHost;
         } else {
-            SocketAddress socketAddress = request.remoteAddress();
+            SocketAddress socketAddress = null;
+            try {
+                socketAddress = request.remoteAddress();
+            } catch (NullPointerException npe) {
+                // ideally there shouldn't be an exception for this call.
+                // This is just to workaround a current issue in vertx
+                // https://github.com/quarkusio/quarkus/issues/5247
+                // https://github.com/eclipse-vertx/vert.x/issues/3181
+                // TODO: remove this workaround once vertx issue is resolved
+                logger.debug("Ignoring exception that occurred when obtaining remote address of request " + request, npe);
+            }
             // client address may not be available with VirtualHttp;
             this.remoteHost = socketAddress != null ? socketAddress.host() : null;
             initialized = true;


patriot1burke (approved):
LGTM

extensions/resteasy/runtime/src/main/java/io/quarkus/resteasy/runtime/standalone/VertxRequestHandler.java
```diff
@@ -97,7 +97,16 @@ private void dispatch(RoutingContext routingContext, InputStream is, VertxOutput
             VertxHttpResponse vertxResponse = new VertxHttpResponse(request, dispatcher.getProviderFactory(),
                     request.method(), allocator, output);
             // client address may not be available with VirtualHttp
-            SocketAddress socketAddress = request.remoteAddress();
+            SocketAddress socketAddress = null;
+            try {
+                socketAddress = request.remoteAddress();
+            } catch (Exception e) {
```
evanchooly: catching `Exception` is probably overly broad.  If we know that vertx is throwing an NPE why don't we catch that here?

jaikiran: @evanchooly, that sounds fine to me. I've now updated this PR to just catch `NullPointerException`.

machi1990:
Fixes https://github.com/quarkusio/quarkus/issues/5247

extensions/resteasy/runtime/src/main/java/io/quarkus/resteasy/runtime/standalone/VertxRequestHandler.java
```diff
@@ -97,7 +97,16 @@ private void dispatch(RoutingContext routingContext, InputStream is, VertxOutput
...
                     request.method(), allocator, output);
             // client address may not be available with VirtualHttp
-            SocketAddress socketAddress = request.remoteAddress();
+            SocketAddress socketAddress = null;
+            try {
+                socketAddress = request.remoteAddress();
+            } catch (NullPointerException npe) {
+                // ideally there shouldn't be an exception for this call.
+                // This is just to workaround a current issue in vertx
+                // https://github.com/quarkusio/quarkus/issues/5247
```
machi1990: can we also add the link to the vertx issue eclipse-vertx/vert.x#3181?

jaikiran: @machi1990, Done

jaikiran:
The failures look unrelated:

```
[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 21.692 s - in io.quarkus.vault.VaultAppRoleITCase
2019-11-07 03:55:51,689 INFO  [io.qua.vau.tes.VaultTestExtension] (main) stop containers
[INFO] 
[INFO] Results:
[INFO] 
[ERROR] Failures: 
[ERROR]   VaultITCase Docker environment should have more than 2GB free disk space
[INFO] 
[ERROR] Tests run: 2, Failures: 1, Errors: 0, Skipped: 0
```

machi1990 (approved):

extensions/resteasy/runtime/src/main/java/io/quarkus/resteasy/runtime/standalone/LazyHostSupplier.java
```diff
@@ -20,7 +23,17 @@ public String getRemoteHost() {
...
+            SocketAddress socketAddress = null;
+            try {
+                socketAddress = request.remoteAddress();
+            } catch (NullPointerException npe) {
+                // ideally there shouldn't be an exception for this call.
+                // This is just to workaround a current issue in vertx
+                // https://github.com/quarkusio/quarkus/issues/5247
+                // https://github.com/eclipse-vertx/vert.x/issues/3181
+                // TODO: remove this workaround once vertx issue is resolved
+                logger.debug("Ignoring exception that occurred when obtaining remote address of request " + request, npe);
```
gwenneg: Logging this at debug level seems a bit low to me, is it really ok considering the remote host will be cached after the NPE occurs?

jaikiran: Hello @gwenneg, I decided to use `DEBUG` since the inability to get the client IP shouldn't usually be a concern to the users. In fact the code here (and the previous version of the code) is designed to handle cases where client address isn't available. That's also the reason why caching this null value is fine.

gwenneg: Ok, that seems reasonnable. Thanks for the explanation.

gwenneg (approved):

gsmet:
I backported it (the code had moved a bit though since then so I did it manually).

extensions/resteasy/runtime/src/main/java/io/quarkus/resteasy/runtime/standalone/LazyHostSupplier.java
cescoffier: This has been fixed in Vert.x 3.8.5. So the workaround it not required anymore.

PR #5268, opened 2019-11-06
Make the tooling resolve the latest version of the universe

aloubyansky:
...by default and fall back to the bundled quarkus-bom in offline mode

### DIFF
--- a/devtools/maven/src/main/java/io/quarkus/maven/CreateProjectMojo.java
+++ b/devtools/maven/src/main/java/io/quarkus/maven/CreateProjectMojo.java
@@ -44,7 +44,6 @@ import org.eclipse.aether.repository.RemoteRepository;
 import org.fusesource.jansi.Ansi;
 
 import io.quarkus.bootstrap.resolver.AppModelResolverException;
-import io.quarkus.bootstrap.resolver.BootstrapAppModelResolver;
 import io.quarkus.bootstrap.resolver.maven.MavenArtifactResolver;
 import io.quarkus.cli.commands.AddExtensionResult;
 import io.quarkus.cli.commands.AddExtensions;
@@ -56,10 +55,6 @@ import io.quarkus.generators.SourceType;
 import io.quarkus.maven.components.MavenVersionEnforcer;
 import io.quarkus.maven.components.Prompter;
 import io.quarkus.maven.utilities.MojoUtils;
-import io.quarkus.platform.descriptor.QuarkusPlatformDescriptor;
-import io.quarkus.platform.descriptor.resolver.json.QuarkusJsonPlatformDescriptorResolver;
-import io.quarkus.platform.tools.config.QuarkusPlatformConfig;
-import io.quarkus.platform.tools.maven.MojoMessageWriter;
 
 /**
  * This goal helps in setting up Quarkus Maven project with quarkus-maven-plugin, with sensible defaults
@@ -83,12 +78,21 @@ public class CreateProjectMojo extends AbstractMojo {
     @Parameter(property = "projectVersion")
     private String projectVersion;
 
-    @Parameter(property = "platformGroupId", defaultValue = CreateUtils.DEFAULT_PLATFORM_BOM_GROUP_ID)
+    /**
+     * Group ID of the target platform BOM
+     */
+    @Parameter(property = "platformGroupId", required = false)
     private String bomGroupId;
 
-    @Parameter(property = "platformArtifactId", defaultValue = CreateUtils.DEFAULT_PLATFORM_BOM_ARTIFACT_ID)
+    /**
+     * Artifact ID of the target platform BOM
+     */
+    @Parameter(property = "platformArtifactId", required = false)
     private String bomArtifactId;
 
+    /**
+     * Version of the target platform BOM
+     */
     @Parameter(property = "platformVersion", required = false)
     private String bomVersion;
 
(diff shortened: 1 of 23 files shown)


integration-tests/maven/src/it/setup-on-existing-pom-it/invoker.properties
```diff
@@ -1 +1 @@
-invoker.goals=io.quarkus:quarkus-maven-plugin:${project.version}:create
+invoker.goals=io.quarkus:quarkus-maven-plugin:${project.version}:create -DplatformArtifactId=quarkus-bom -DplatformVersion=${project.version}
```
maxandersen: these explicit set of platform properties is to just make current tests pass using the "built-in" platform catalog, right ? 

could we add some tests (possibly just unit tests) for the various combinations of the resolution logic since there seem to be quite a lot of different execution paths in the logic ?

stuartwdouglas:
The failure looks related

integration-tests/maven/src/it/setup-on-existing-pom-it/invoker.properties
aloubyansky: > these explicit set of platform properties is to just make current tests pass using the "built-in" platform catalog, right ?

Right.

> could we add some tests (possibly just unit tests) for the various combinations of the resolution logic since there seem to be quite a lot of different execution paths in the logic ?

I added tests for the resolver in the tools resolver.

maxandersen:
trying to run this now (waiting on build ;)

Am I right in seeing that `-DplatformVersion=` actually supports taking a maven version expression ? which then translates into actual `platformVersion` actually being resolved ?

devtools/maven/src/main/java/io/quarkus/maven/CreateProjectMojo.java
```diff
@@ -83,12 +78,21 @@
...
-    @Parameter(property = "platformArtifactId", defaultValue = CreateUtils.DEFAULT_PLATFORM_BOM_ARTIFACT_ID)
+    /**
+     * Artifact ID of the target platform BOM
+     */
+    @Parameter(property = "platformArtifactId", required = false)
     private String bomArtifactId;
 
+    /**
+     * Version of the target platform BOM
+     */
```
maxandersen: This is treated as a "Version or version range of the target platform" which is then used to resolve bom and/or json, right?

maxandersen: hmm - where are the limits for the version range search actually set ?

maxandersen:
so the feature seem to work but I didn't find a way to specify it shouldn't try resolve anything nor control the version range. Is that not supposed to be available ?

devtools/maven/src/main/java/io/quarkus/maven/CreateProjectMojo.java
aloubyansky: `bomVersion` is the actual version. If it's not set then the latest one will be resolved.
The range is set here 
https://github.com/quarkusio/quarkus/pull/5268/files#diff-20e78dc83510b6617d2789e2e291bb1aR235
and here
https://github.com/quarkusio/quarkus/pull/5268/files#diff-20e78dc83510b6617d2789e2e291bb1aR429
The range can be customized using this property https://github.com/quarkusio/quarkus/blob/master/independent-projects/tools/platform-descriptor-resolver-json/src/main/java/io/quarkus/platform/descriptor/resolver/json/QuarkusJsonPlatformDescriptorResolver.java#L51
This can definitely be improved.

maxandersen: yeah so this feels a tad confusing. this why I changed the names of what the user requests at create time to be about what platform he is looking for not wether it’s a bom or json. That it maps directly to a bom is a implementation detail imo ;) 

is there a reason to not make the `platformVersion` parameter version range and be applicable no matter if artifact/groupid . are specified ? (since exact version is possible there too) ?

aloubyansky: Yes, it could be convenient to make it a range as well.
About requiring the artifactId if the version is provided, I didn't like making assumptions about what the default artifactId is in the mojo. But that's arguable and I won't mind not requiring it.

aloubyansky: Ok, it's done. You can provide a range through `platformVersion` (quoted to get it accepted as a cmd arg) and `platformArtifactId` isn't required.

maxandersen:
This looks good!  Still testing out a bit for edge cases - get that reported later today.

Would love to see others use this though just to increase our confidence in it - @gastaldi  and @geoand  could you try it out ?

geoand:
I think the PR needs a rebase onto the latest master to pick up some CI fixes

devtools/maven/src/main/java/io/quarkus/maven/CreateUtils.java
```diff
@@ -28,6 +35,57 @@ private CreateUtils() {
...
+            return true;
+        }
+        c = versionStr.charAt(versionStr.length() - 1);
+        if (c == ']' || c == ')') {
+            return true;
+        }
+        return versionStr.indexOf(',') >= 0;
+    }
+
+    private static String nonEmptyOrNull(String value) {
```
gastaldi: Reuse `org.apache.commons.lang3.StringUtils` instead

aloubyansky: Is there any other benefit? I kind of don't like to add a new dependency on the whole library just for one trivial method.

gastaldi: You'll not add a new dependency, since this class already uses `StringUtils`  (see the imports)

aloubyansky: Ah, you're right, sorry, will change then.

aloubyansky: Done.

gastaldi: The suggestion was to remove the method completely :) See https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/StringUtils.html#isNotEmpty-java.lang.CharSequence-

gastaldi: Actually the method in this case is: https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/StringUtils.html#defaultIfBlank-T-T-

aloubyansky: Done. Thanks.

cescoffier:
Are we good to go on this one?

gastaldi:
CI didn't triggered the tests yet

maxandersen:
force pushed after rebase with master to trigger CI

maxandersen (approved):
This looks good to me and @geoand verified it work for him too. waiting just for CI to finish.

gsmet:
I unbackported it as it broke my build.

PR #5270, opened 2019-11-07
Remove augmentation messages from test and dev mode

stuartwdouglas:


machi1990 (approved):

gwenneg (approved):

loicmathieu:
Just saw this change, and I think it's a pity to remove the augmentation time log on dev mode.
As I understand it (but maybe I'm wrong), on dev mode the augmentation is made at runtime/hot reload and on production mode during the maven/gradle build.

So the augmentation time reported on dev mode is more or less the time saved on bootstraping thanks to the Quarkus deployement process relying eavily on build time optimization.

When I do demos of Quarkus, I explain that Quarkus is "slow" to start in dev mode (4s for a typicall cloud native stack with REST/JPA/Health/... quick for some but slow for Quarkus) but that the time reported for the augmentation will not exist when running on production mode ...

PR #5272, opened 2019-11-07
Unified way for extensions to register metrics

jmartisk:
\+ metrics for Agroal

Work in progress, any feedback welcome, especially from extension maintainers who plan to introduce metrics...
@lhauspie

### DIFF
--- a/docs/src/main/asciidoc/datasource.adoc
+++ b/docs/src/main/asciidoc/datasource.adoc
@@ -265,6 +265,24 @@ If you have multiple datasources, all datasources will be checked and the status
 
 This behavior can be disabled via the property `quarkus.datasource.health.enabled`.
 
+== Datasource Metrics
+
+If you are using the `quarkus-smallrye-metrics` extension, `quarkus-agroal` can expose some data source metrics on the
+`/metrics` endpoint. This can be turned on by setting the property `quarkus.datasource.metrics.enabled` to true.
+
+For the exposed metrics to contain any actual values, it is necessary that metric collection is enabled internally
+by Agroal mechanisms. By default, this metric collection mechanism gets turned on for all data sources if the `quarkus-smallrye-metrics`
+is present and metrics for the Agroal extension are enabled. If you want to disable metrics for a particular data source,
+this can be done by setting `quarkus.datasource.enable-metrics` to `false` (or `quarkus.datasource.<datasource name>.enable-metrics` for a named datasource). This disables
+collecting the metrics as well as exposing them in the `/metrics` endpoint, because it does not make sense to
+expose metrics if the mechanism to collect them is disabled.
+
+Conversely, setting `quarkus.datasource.enable-metrics` to `true` (or `quarkus.datasource.<datasource name>.enable-metrics` for a named datasource) explicitly can be used to enable collection of metrics even if
+the `quarkus-smallrye-metrics` extension is not in use. This can be useful if you need to access the collected metrics programmatically.
+They are available after calling `dataSource.getMetrics()` on an injected `AgroalDataSource` instance. If collection of metrics is disabled
+for this data source, all values will be zero.
+
+
 == Narayana Transaction Manager integration
 
 If the Narayana JTA extension is also available, integration is automatic.
@@ -319,4 +337,3 @@ quarkus.datasource.driver=org.h2.Driver
 == Agroal Configuration Reference
 
 include::{generated-dir}/config/quarkus-agroal.adoc[opts=optional, leveloffset=+1]
-
--- a/docs/src/main/asciidoc/writing-extensions.adoc
+++ b/docs/src/main/asciidoc/writing-extensions.adoc
@@ -1290,6 +1290,143 @@ HealthBuildItem addHealthCheck(AgroalBuildTimeConfig agroalBuildTimeConfig) {
 }
 ----
 
+=== Extension Metrics
+
+An extension can decide to provide metrics through the `quarkus-smallrye-metrics` extension.
+A typical use case for this would be that an extension scans the application code for relevant components (like entities, messaging endpoints, etc.)
+and creates a set of metrics for each of these components. A unified mechanism for metric registration is provided via the `MetricBuildItem` class
+provided by the `quarkus-smallrye-metrics-spi` module.
+
+There are several distinct situations that can occur and each requires slightly different handling:
+
+1. The underlying library used by your extension is using the MicroProfile Metrics API directly.
+2. The underlying library uses its own way for collecting metrics and makes them available at runtime using its own API.
+3. The underlying library does not provide metrics (or there is no library at all) and you need to insert some code in the extension's codebase that will collect the metrics.
+
+What is common for all cases is that the extension should have a build-time config property that enables metrics exposure for the extension,
+it should be named `quarkus.<extension>.metrics.enabled` and be `false` by default.
+
+==== Case 1: The library uses MP Metrics
+If the library exposes metrics by itself, we don't have to do much. However, there are a few points to consider:
+
+- It should be possible to disable all metrics for an extension, which can be a bit problematic if the library registers them directly
+rather than through the unified registration mechanism in the `quarkus-smallrye-metrics-spi` module. Therefore the library should contain
+a way to turn all metrics off, for example using a library-specific system property that can be set during build time (by emitting a `SystemPropertyBuildItem`),
+and will disable metrics if at least one of the properties `quarkus.<extensionName>.metrics.enabled` and `quarkus.smallrye-metrics.extensions.enabled` is `false`.
+- It is desirable to be able to omit the MP Metrics dependency at runtime, so if possible, the library should be written in a way that
+it will still work when the MP Metrics dependencies (or at least the implementation, `io.smallrye:smallrye-metrics`) are unavailable. This can be achieved by
+wrapping all code that does something metric-related into an `if` condition that checks whether metrics integration is enabled. If the library performs injections
+of the `MetricRegistry`, which is not avoidable by introducing an `if` condition, it will unfortunately not be possible to remove the `microprofile-metrics` dependency,
+but it is still possible to avoid the need for `smallrye-metrics` (which contains the implementation class of the metric registry) by introducing a custom "no-op"
+implementation of the `MetricRegistry` which can, for example, return `null` from all its methods. This no-op implementation should be added to the application's classes
+so that it can be injected instead of the regular `io.smallrye.metrics.MetricRegistryImpl`. An example of a no-op implementation can be found at
+https://github.com/quarkusio/quarkus/blob/master/extensions/smallrye-fault-tolerance/runtime/src/main/java/io/quarkus/smallrye/faulttolerance/runtime/NoopMetricRegistry.java[NoopMetricRegistry]
+
+==== Case 2: The library provides its own metric API
+In this case, the extension can make use of the `quarkus-smallrye-metrics-spi` module to expose the metrics from the library using MP Metrics.
+
+You should do the following:
+
+- Import the `quarkus-smallrye-metrics-spi` library in your deployment module.
+- Import the `quarkus-smallrye-metrics` extension as an **optional** dependency in your runtime module so it will not impact the size of the application if
+metrics are not included.
+- In your processor, produce any number of `MetricBuildItem` items as shown in the example below.
+The metric name of all metrics should start with a short name of the extension that owns them.
+- Pass the value of `quarkus.<extension>.metrics.enabled` to the constructor of `MetricBuildItem` objects that the extension produces and if it is false, the metric will be ignored.
+
+An example build step for build-time registration of metrics:
+
+[source%nowrap.java]
+----
+@BuildStep
+void registerMetrics(BuildProducer<MetricBuildItem> metrics) {
+    Metadata activeCountMetadata = Metadata.builder()
+            .withName("agroal.active.count")
+            .withDescription("Count of active connections")
+            .withType(MetricType.GAUGE)
+            .build();
+    List<DataSource> dataSources = getAllDataSources(); // your relevant components specific to the extension
+    for(DataSource dataSource : dataSources) {
+        metrics.produce(new MetricBuildItem(activeCountMetadata,
+                new ActiveCountGauge(dataSource.getName()),
+                agroalBuildTimeConfig.metricsEnabled,
+                configRootName, // name of the config root pertaining to this extension
+                new Tag("datasource", dataSource.getName())));
+    }
+}
+----
+
+In this example, there is a gauge per each data source that shows the current number of active connections.
+In this example, we provide an object that implements the logic of the metric, in the case of a gauge,
+it needs to implement `org.eclipse.microprofile.metrics.Gauge`. Example:
+
+[source%nowrap.java]
+----
+public class ActiveCountGauge implements Gauge<Long> {
+
+    private String dataSourceName;
+
+    public ActiveCountGauge() {
+
+    }
+
+    public ActiveCountGauge(String dataSourceName) {
+        this.dataSourceName = dataSourceName;
+    }
+
+    @Override
+    public Long getValue() {
+        DataSource ds = obtainDataSource(dataSourceName); // some logic to get hold of the data source
+        return ds.getMetrics().getActiveCount();
+    }
+}
+----
+
+When registering gauges, it is required to provide a custom object that implements the correct metric type.
+For metric types other than gauges, this is optional - you can choose whether to provide an implementation,
+or whether the default implementation class from `SmallRye Metrics` should be instantiated for the metric and registered
+in the metric registry. To update the metric value, it will be necessary to look them up in the metric registry
+and call the methods specific to each metric type.
+
+Look into the `AgroalProcessor#registerMetrics` method for an example how this was done for the Agroal extension.
+
+==== Case 3: It is necessary to collect metrics within the extension code
+
+In this case, the dependency setup and build-time registration is the same as in case 2. The difference will be that instead of introducing
+custom metric objects that bridge between library-specific metrics and MP Metrics, the extension's code will contain metric collection code like this:
+
+[source%nowrap.java]
+----
+public void methodThatShouldBeCounted() {
+    if(metricsEnabled) {
+        Counter counter = MetricRegistries.get(MetricRegistry.Type.VENDOR).counter("name");
+        counter.inc();
+    }
+    // proceed with the invocation normally
+}
+----
+
+In this case, be aware that the MP Metrics API and implementation might not be available on the runtime classpath!
+This is why there is a `if(metricsEnabled)` check that should make sure that if metrics are not enabled, we don't touch MP Metrics classes,
+because that would fail. The extension itself will need to provide such check, and it must return true only if both of these conditions hold:
+
+- The Metrics capability is present and therefore the SmallRye Metrics library is available
+- Metrics are not disabled for this extension
+
+The recommended way to implement this check would be to produce a specific `SystemPropertyBuildItem` during build, which will make it easy
+to evaluate these two conditions, and then check back to this system property at runtime. Example:
+
+----
+@BuildStep
+SystemPropertyBuildItem produceMetricsEnabledProperty(AgroalBuildTimeConfig extensionSpecificConfig,
+        Capabilities capabilities) {
+    return new SystemPropertyBuildItem("metrics.enabled",
+            String.valueOf(capabilities.isCapabilityPresent(Capabilities.METRICS) &&
+                            extensionSpecificConfig.metricsEnabled));
+}
+----
+
+
 === Customizing JSON handling from an extension
 Extensions often need to register serializers and/or deserializers for types the extension provides.
 
--- a/extensions/agroal/deployment/src/main/java/io/quarkus/agroal/deployment/AgroalProcessor.java
+++ b/extensions/agroal/deployment/src/main/java/io/quarkus/agroal/deployment/AgroalProcessor.java
@@ -1,9 +1,11 @@
 package io.quarkus.agroal.deployment;
 
+import static io.quarkus.deployment.annotations.ExecutionTime.RUNTIME_INIT;
 import static io.quarkus.deployment.annotations.ExecutionTime.STATIC_INIT;
 
 import java.sql.Driver;
 import java.util.Arrays;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map.Entry;
 import java.util.Optional;
@@ -15,6 +17,10 @@ import javax.enterprise.inject.Produces;
 import javax.enterprise.inject.spi.DeploymentException;
 import javax.sql.XADataSource;
 
+import org.eclipse.microprofile.metrics.Metadata;
+import org.eclipse.microprofile.metrics.MetricType;
+import org.eclipse.microprofile.metrics.MetricUnits;
+import org.eclipse.microprofile.metrics.Tag;
 import org.jboss.jandex.AnnotationInstance;
 import org.jboss.jandex.AnnotationValue;
 import org.jboss.jandex.DotName;
@@ -23,6 +29,8 @@ import org.jboss.logging.Logger;
 
 import io.agroal.api.AgroalDataSource;
 import io.quarkus.agroal.DataSource;
+import io.quarkus.agroal.metrics.AgroalCounter;
+import io.quarkus.agroal.metrics.AgroalGauge;
 import io.quarkus.agroal.runtime.AbstractDataSourceProducer;
 import io.quarkus.agroal.runtime.AgroalBuildTimeConfig;
 import io.quarkus.agroal.runtime.AgroalRecorder;
@@ -34,6 +42,7 @@ import io.quarkus.arc.deployment.GeneratedBeanBuildItem;
 import io.quarkus.arc.deployment.GeneratedBeanGizmoAdaptor;
 import io.quarkus.arc.deployment.UnremovableBeanBuildItem;
 import io.quarkus.arc.processor.DotNames;
+import io.quarkus.deployment.Capabilities;
 import io.quarkus.deployment.annotations.BuildProducer;
 import io.quarkus.deployment.annotations.BuildStep;
 import io.quarkus.deployment.annotations.ExecutionTime;
@@ -51,6 +60,7 @@ import io.quarkus.gizmo.MethodCreator;
 import io.quarkus.gizmo.MethodDescriptor;
 import io.quarkus.gizmo.ResultHandle;
 import io.quarkus.smallrye.health.deployment.spi.HealthBuildItem;
+import io.quarkus.smallrye.metrics.deployment.spi.MetricBuildItem;
 
 class AgroalProcessor {
 
@@ -76,7 +86,8 @@ class AgroalProcessor {
             BuildProducer<NativeImageResourceBuildItem> resource,
             BuildProducer<DataSourceDriverBuildItem> dataSourceDriver,
             SslNativeConfigBuildItem sslNativeConfig, BuildProducer<ExtensionSslNativeSupportBuildItem> sslNativeSupport,
-            BuildProducer<GeneratedBeanBuildItem> generatedBean) throws Exception {
+            BuildProducer<GeneratedBeanBuildItem> generatedBean,
+            Capabilities capabilities) throws Exception {
 
         feature.produce(new FeatureBuildItem(FeatureBuildItem.AGROAL));
 
@@ -127,7 +138,8 @@ class AgroalProcessor {
         String dataSourceProducerClassName = AbstractDataSourceProducer.class.getPackage().getName() + "."
                 + "DataSourceProducer";
 
-        createDataSourceProducerBean(generatedBean, dataSourceProducerClassName);
+        createDataSourceProducerBean(generatedBean, dataSourceProducerClassName,
+                capabilities.isCapabilityPresent(Capabilities.METRICS));
 
         return new BeanContainerListenerBuildItem(recorder.addDataSource(
                 (Class<? extends AbstractDataSourceProducer>) recorderContext.classProxy(dataSourceProducerClassName),
@@ -228,7 +240,7 @@ class AgroalProcessor {
      * Build time and runtime configuration are both injected into this bean.
      */
     private void createDataSourceProducerBean(BuildProducer<GeneratedBeanBuildItem> generatedBean,
-            String dataSourceProducerClassName) {
+            String dataSourceProducerClassName, boolean metricsCapabilityPresent) {
         ClassOutput classOutput = new GeneratedBeanGizmoAdaptor(generatedBean);
 
         ClassCreator classCreator = ClassCreator.builder().classOutput(classOutput)
@@ -252,15 +264,16 @@ class AgroalProcessor {
             ResultHandle dataSourceRuntimeConfig = defaultDataSourceMethodCreator.invokeVirtualMethod(
                     MethodDescriptor.ofMethod(AbstractDataSourceProducer.class, "getDefaultRuntimeConfig", Optional.class),
                     defaultDataSourceMethodCreator.getThis());
+            ResultHandle mpMetricsEnabled = defaultDataSourceMethodCreator.load(metricsCapabilityPresent);
 
             defaultDataSourceMethodCreator.returnValue(
                     defaultDataSourceMethodCreator.invokeVirtualMethod(
                             MethodDescriptor.ofMethod(AbstractDataSourceProducer.class, "createDataSource",
                                     AgroalDataSource.class, String.class,
-                                    DataSourceBuildTimeConfig.class, Optional.class),
+                                    DataSourceBuildTimeConfig.class, Optional.class, boolean.class),
                             defaultDataSourceMethodCreator.getThis(),
                             dataSourceName,
-                            dataSourceBuildTimeConfig, dataSourceRuntimeConfig));
+                            dataSourceBuildTimeConfig, dataSourceRuntimeConfig, mpMetricsEnabled));
         }
 
         for (Entry<String, DataSourceBuildTimeConfig> namedDataSourceEntry : agroalBuildTimeConfig.namedDataSources
@@ -292,15 +305,16 @@ class AgroalProcessor {
                     MethodDescriptor.ofMethod(AbstractDataSourceProducer.class, "getRuntimeConfig", Optional.class,
                             String.class),
                     namedDataSourceMethodCreator.getThis(), namedDataSourceNameRH);
+            ResultHandle mpMetricsEnabled = namedDataSourceMethodCreator.load(metricsCapabilityPresent);
 
             namedDataSourceMethodCreator.returnValue(
                     namedDataSourceMethodCreator.invokeVirtualMethod(
                             MethodDescriptor.ofMethod(AbstractDataSourceProducer.class, "createDataSource",
                                     AgroalDataSource.class, String.class,
-                                    DataSourceBuildTimeConfig.class, Optional.class),
+                                    DataSourceBuildTimeConfig.class, Optional.class, boolean.class),
                             namedDataSourceMethodCreator.getThis(),
                             namedDataSourceNameRH,
-                            namedDataSourceBuildTimeConfig, namedDataSourceRuntimeConfig));
+                            namedDataSourceBuildTimeConfig, namedDataSourceRuntimeConfig, mpMetricsEnabled));
         }
 
         classCreator.close();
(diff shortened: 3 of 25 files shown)


loicmathieu (commented):
I have several questions about this:

- Metrics consumes resources on each call, so should we enable it by default? Enabling health by default is not an issue because it's only used when calls ...
- For your datasource implementation, I saw that  you develop one specific Metric class per metric. I have concern about this as it will not scale well when you want to expose 50 metrics for an extension.
- You didn't implements metrics by annotations like in my prototype, is it something that you plan to integrate next ?

lhauspie:
I'll try to use that on kafka-streams extension and give you feedback.
But I still have some work to add readiness check due to very interesting discussions on my #4793

docs/src/main/asciidoc/datasource.adoc
```diff
@@ -219,18 +219,21 @@ quarkus.datasource.url=jdbc:h2:tcp://localhost/mem:default
 quarkus.datasource.username=username-default
 quarkus.datasource.min-size=3
 quarkus.datasource.max-size=13
+quarkus.datasource.enable-metrics=true
```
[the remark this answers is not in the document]
lhauspie: I think this property should be `quarkus.datasource.metrics.enabled=true` to be consistent with health check e.g. `quarkus.kafka-streams.health.enabled=false`.

lhauspie (changes requested):

docs/src/main/asciidoc/datasource.adoc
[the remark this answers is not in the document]
jmartisk: This is the `enable-metrics` property that enables metric collection, not exposition, and it is per data source. It is already in Quarkus, I'm only adding it to the documentation example, because you need to enable metrics collection if you also enable metrics exposition and don't want to have all values equal to zero.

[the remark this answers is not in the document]
jmartisk: This is the `enable-metrics` property that enables metric collection, not exposition, and it is per data source. It is already in Quarkus, I'm only adding it to the documentation example, because you need to enable metrics collection if you also enable metrics exposition and don't want to have all values equal to zero.

extensions/agroal/deployment/src/main/java/io/quarkus/agroal/deployment/AgroalProcessor.java
```diff
@@ -307,4 +315,64 @@ HealthBuildItem addHealthCheck(AgroalBuildTimeConfig agroalBuildTimeConfig) {
...
+        Metadata acquireCountMetadata = Metadata.builder()
+                .withName("agroal.acquire.count")
+                .withDescription("Count of acquired connections")
+                .withType(MetricType.COUNTER)
+                .build();
+        if (agroalRuntimeConfig.defaultDataSource != null) {
+            /*
+             * TODO: what is the correct thing to do when metrics for the extension are enabled,
+             * but a data source has its metrics collection disabled? Should we log a warning that the values
+             * will be zero, or should we not publish metrics for this data source?
```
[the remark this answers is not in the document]
jmartisk: But on which data sources, if there are many? On all of them? And even if the user explicitly set `enable-metrics=false` on some data source? It would probably be fine to automatically enable it if the user didn't specify `enable-metrics` at all, but overriding the user's decision doesn't seem right.
 And I'm not sure if we have the facility to find out if the user specified `false` or didn't specify it at all (in which case it defaults to `false` too)

extensions/agroal/deployment/src/main/java/io/quarkus/agroal/deployment/AgroalProcessor.java
```diff
@@ -307,4 +315,64 @@ HealthBuildItem addHealthCheck(AgroalBuildTimeConfig agroalBuildTimeConfig) {
         return new HealthBuildItem("io.quarkus.agroal.runtime.health.DataSourceHealthCheck",
                 agroalBuildTimeConfig.healthEnabled, "datasource");
     }
+
+    // TODO: introduce more metrics
+    @BuildStep
+    void registerMetrics(AgroalRuntimeConfig agroalRuntimeConfig,
```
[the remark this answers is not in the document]
jmartisk: I need to inject the runtime config, because the build config doesn't contain the `metrics-enabled` property which enables/disables metric collection, and we need to work with its value here.

docs/src/main/asciidoc/writing-extensions.adoc
```diff
@@ -1238,6 +1238,123 @@ HealthBuildItem addHealthCheck(AgroalBuildTimeConfig agroalBuildTimeConfig) {
...
+and creates a set of metrics for each of these components. There is a unified mechanism for metric registration - it uses the `MetricBuildItem` class
+provided by the `quarkus-smallrye-metrics-spi` module.
+
+There are several distinct situations that can happen and each requires a bit different handling:
+
+1. The underlying library used by your extension is using MP Metrics API directly.
+2. The underlying library uses its own way for collecting metrics and makes them available at runtime using its own API.
+3. The underlying library does not provide metrics (or there is no library at all) and you need to insert some code in the extension's codebase that will collect the metrics.
+
+==== Case 1: The library uses MP Metrics
```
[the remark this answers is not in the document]
jmartisk: Hmmmmm, I'm thinking again whether that pattern is correct at all - will it really work?
See for example https://github.com/smallrye/smallrye-fault-tolerance/blob/2.1.2/implementation/src/main/java/io/smallrye/faulttolerance/metrics/BulkheadWaitRecorder.java#L55
I think a no-op registry returning null on this line will cause a NPE if we don't prevent this method from executing. It's generally quite a fragile pattern that can break due to a lot of things

jmartisk:
>     * Metrics consumes resources on each call, so should we enable it by default? Enabling health by default is not an issue because it's only used when calls ...

I'm not sure what exactly you mean by enabling it by default here. The default is that an application's pom.xml doesn't have a dependency on `quarkus-smallrye-metrics` so it will not get metrics. And when the dependency is present, I would assume that the user indeed wants metrics.

> 
>     * For your datasource implementation, I saw that  you develop one specific Metric class per metric. I have concern about this as it will not scale well when you want to expose 50 metrics for an extension.
That's a good point, we could probably move all the metric retrieval logic into one bigger application-scoped bean and then the `Callable`-s registered in the metric registry will be just anonymous instances that obtain a reference to that bean and call one of its methods (and pass the return value on). Would that sound better?

> 
>     * You didn't implements metrics by annotations like in my prototype, is it something that you plan to integrate next ?
We could do that but we first need to find out how to get some annotated metrics into the `vendor` metric registry without making the whole API a big mess. It will most likely need some changes in SmallRye Metrics side. So I thought we would start with the programmatic way.

extensions/agroal/deployment/src/main/java/io/quarkus/agroal/deployment/AgroalProcessor.java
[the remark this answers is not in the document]
loicmathieu: So we need to switch to a String boolean `enable-metrics` with a default of auto :

- true: always enabled
- false: never enabled
- auto: enabled if smallrye-metrics is in the classpath ..

Or we need to switch to an Optional config item, this should also do the trick

docs/src/main/asciidoc/writing-extensions.adoc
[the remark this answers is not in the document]
loicmathieu: @jmartisk so you think that the implementation of #5388 will not works ? If I understand your concern the Bulkhead functionality should generates a NPE.

Only fallback (inside the extension) and retry (inside the main integration test) are tested for fault-tolerance inside Quarkus. Maybe adding more test will give us confidence about it.

extensions/agroal/deployment/src/main/java/io/quarkus/agroal/deployment/AgroalProcessor.java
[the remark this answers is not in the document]
jmartisk: With the optional config item, you mean that if it's undefined, we will treat is as "auto"? I think that would make sense. Sounds better to me than turning it into a string. 

Btw this could get affected by the incoming config changes https://github.com/quarkusio/quarkus/pull/5387, I'm particularly worried about the "Tighten rules about accessing run time configuration values from build steps.", but I'm not familiar with what exactly the changes will be

docs/src/main/asciidoc/writing-extensions.adoc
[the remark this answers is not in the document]
jmartisk: Ok; after the discussion on that PR I added a few sentences about this to the end of "Case 1" section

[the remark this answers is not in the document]
loicmathieu: Maybe adding a reference to `smallrye-fault-tolerane` extension will be good, as it's an extension author guide adding guidance as example implementation is a good thing.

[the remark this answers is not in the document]
jmartisk: Ok, but I will add that after your PR is merged :)

jmartisk:
I just addressed @loicmathieu 's concern about the fact that each metric would need its own class - now it's just one class per metric type. See `AgroalGauge` and `AgroalCounter` classes. The processor passes the name of the metric that should be retrieved as a `String`. A lambda or something would be nicer, but we can't pass lambdas from the deployment code as they would not be available at runtime.

Also introduced the option to provide a custom metric object for all metric types (this is required for gauges and optional for other metric types). It's no longer a `Callable`, I realized we can use the MP Metrics interfaces directly, so the build item now expects a serializable implementation of `org.eclipse.microprofile.metrics.Gauge` etc.

Cleaned up the logic in `AgroalProcessor` to avoid too much code duplication.

Didn't yet address the mismatch between `metrics.enabled` globally and `enable-metrics` set per data source. I need to find out if we will indeed be able to read and especially change the runtime configuration after the incoming config rework

extensions/agroal/runtime/src/main/java/io/quarkus/agroal/metrics/AgroalCounter.java
```diff
@@ -0,0 +1,81 @@
...
+    public void setDataSourceName(String dataSourceName) {
+        this.dataSourceName = dataSourceName;
+    }
+
+    private AgroalDataSource getDataSource() {
+        if (dataSource == null) {
+            if (dataSourceName == null) {
+                dataSource = CDI.current().select(AgroalDataSource.class).get();
+            } else {
+                dataSource = CDI.current().select(AgroalDataSource.class, new DataSourceLiteral(dataSourceName)).get();
```
loicmathieu: Inside extensions will usually direclty use `Arc` instead of `CDI`.
Not sure it have a performance impact.

extensions/smallrye-metrics/spi/src/main/java/io/quarkus/smallrye/metrics/deployment/spi/MetricBuildItem.java
```diff
@@ -0,0 +1,75 @@
...
+    /**
+     * Create a metric build item from the specified metadata and tags.
+     * Such metric will be picked up by the Metrics extension and registered in the VENDOR registry.
+     * This constructor is applicable to all metric types except gauges.
+     *
+     * @param metadata The metadata that should be applied to the registered metric
+     * @param enabled Whether this metric is enabled
+     * @param tags The tags that will be applied to this metric
+     */
+    public MetricBuildItem(Metadata metadata, boolean enabled, Tag... tags) {
```
loicmathieu: You should also add the `configRoot` field to your MetricBuildItem as it is done to Health.
The configRoot is not the same as the extension name (for agroal, the config root is datasource).

This is for a future usage to implements more complex enabling/disabling of extension metrics like 
```
quarkus.metrics.extensions=transaction,kafka
# enabled metrics for the transaction and kafka extension only
```

loicmathieu (changes requested):

loicmathieu:
> A lambda or something would be nicer, but we can't pass lambdas from the deployment code as they would not be available at runtime.

Are you sure? I already seen some inside deployment code that pass callbacks at runtime like `io.quarkus.kafka.streams.deployment.KafkaStreamsHotReplacementSetup.OnMessage` but maybe it's not the same construct.

jmartisk:
> Are you sure? I already seen some inside deployment code that pass callbacks at runtime like `io.quarkus.kafka.streams.deployment.KafkaStreamsHotReplacementSetup.OnMessage` but maybe it's not the same construct.

I don't understand these technical details too much, but I'd say with `KafkaStreamsHotReplacementSetup` it works because that's used by devmode and the devmode-related code can use deployment classes. In my case I tried passing a `Function<String, Long>` implemented as a lambda, instead of just `String`, and I got a `NoClassDefFoundError` at runtime.

extensions/agroal/runtime/src/main/java/io/quarkus/agroal/metrics/AgroalCounter.java
jmartisk: Ok, changed that to Arc

extensions/smallrye-metrics/spi/src/main/java/io/quarkus/smallrye/metrics/deployment/spi/MetricBuildItem.java
jmartisk: Done

loicmathieu:
> I don't understand these technical details too much,  I'd say with `KafkaStreamsHotReplacementSetup` it works because that's used by devmode and the devmode-related code can use deployment classes.

You're right, I think you understand these technical details better than me :)

extensions/agroal/deployment/src/main/java/io/quarkus/agroal/deployment/AgroalProcessor.java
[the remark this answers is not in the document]
loicmathieu: Yes, you're right, datasource `enable-metrics` is RUNTIME so cannot be used inside a build step.
Your warning must be emitter inside your Metric classes or inside a recorder that is used inside a RUN_TIME build step.

loicmathieu (changes requested):
As said, the ` enable-metrics` datasource config item is runtime so cannot be read inside a build step.

docs/src/main/asciidoc/writing-extensions.adoc
```diff
@@ -1238,6 +1238,129 @@ HealthBuildItem addHealthCheck(AgroalBuildTimeConfig agroalBuildTimeConfig) {
...
+            .withName("agroal.active.count")
+            .withDescription("Count of active connections")
+            .withType(MetricType.GAUGE)
+            .build();
+    List<DataSource> dataSources = getAllDataSources(); // your relevant components specific to the extension
+    for(DataSource dataSource : dataSources) {
+        metrics.produce(new MetricBuildItem(activeCountMetadata,
+                new ActiveCountGauge(dataSource.getName()),
+                agroalBuildTimeConfig.metricsEnabled,
+                new Tag("datasource", dataSource.getName())));
```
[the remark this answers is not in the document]
jmartisk: Good catch, was planning to add this but somehow forgot

emmanuelbernard:
> As said, the ` enable-metrics` datasource config item is runtime so cannot be read inside a build step.

Hum `quarkus.datasource.health.enabled` is build time, should we do the same?

jmartisk:
> > As said, the ` enable-metrics` datasource config item is runtime so cannot be read inside a build step.
> 
> Hum `quarkus.datasource.health.enabled` is build time, should we do the same?

Do you mean turn `quarkus.datasource.DS_NAME.enable-metrics` into a build time property? (It's the one that enables metric collection per each data source)
Are you sure you're not confusing it with `quarkus.datasource.metrics.enabled`, which *is* a build time property and enables metric exposition in MP Metrics for the whole extension?
Damn, these two properties are really similar, a lot of people will be confused. We need to make sure we have very clear docs.
I would rather name the `metrics.enabled` property `metrics.exposed`, but then it would not be consistent with what other extensions have (unless the other extensions do it too).

extensions/agroal/deployment/src/main/java/io/quarkus/agroal/deployment/AgroalProcessor.java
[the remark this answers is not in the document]
jmartisk: Hm, I just changed that to inject the runtime config into the processor and only pass it to a recorder, but it seems https://github.com/quarkusio/quarkus/issues/5514 will forbid this usage. So I will change it again.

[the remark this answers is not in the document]
dmlloyd: Injecting to the build step and passing it to the recorder is OK for now.  We don't yet support directly injecting the config into the recorder so that has to be implemented before we can take any other approach.

emmanuelbernard:
> > > As said, the ` enable-metrics` datasource config item is runtime so cannot be read inside a build step.
> > 
> > 
> > Hum `quarkus.datasource.health.enabled` is build time, should we do the same?
> 
> Do you mean turn `quarkus.datasource.DS_NAME.enable-metrics` into a build time property? (It's the one that enables metric collection per each data source)
> Are you sure you're not confusing it with `quarkus.datasource.metrics.enabled`, which _is_ a build time property and enables metric exposition in MP Metrics for the whole extension?
> Damn, these two properties are really similar, a lot of people will be confused. We need to make sure we have very clear docs.
> I would rather name the `metrics.enabled` property `metrics.exposed`, but then it would not be consistent with what other extensions have (unless the other extensions do it too).
> Or we could rename `enable-metrics` to `collect-metrics` or something, but that would be a breaking change and I'm not sure if it would help much

All of the properties with `health` in them according to https://quarkus.io/guides/all-config are build time.
So I'm wondering why enable-metrics should be a runtime property?

Note that we can't rename properties now, we are in the 1.0+ phase.

jmartisk:
Perhaps it was made a runtime property because Agroal supports enabling/disabling metrics dynamically: https://github.com/agroal/agroal/blob/1.7/agroal-api/src/main/java/io/agroal/api/configuration/AgroalDataSourceConfiguration.java#L33

IIUC Quarkus is not making use of this at the moment, it treats the data source configuration as immutable once the data source is built. But we could detect a change in the config value and dynamically update the data source's configuration without having to create a new instance.

emmanuelbernard:
> Perhaps it was made a runtime property because Agroal supports enabling/disabling metrics dynamically: https://github.com/agroal/agroal/blob/1.7/agroal-api/src/main/java/io/agroal/api/configuration/AgroalDataSourceConfiguration.java#L33
> 
> IIUC Quarkus is not making use of this at the moment, it treats the data source configuration as immutable once the data source is built. But we could detect a change in the config value and dynamically update the data source's configuration without having to create a new instance.

I don't think we need such a feature. It was most likely an oversight.

jmartisk:
@emmanuelbernard  Yep, I agree. I'll change it to a build time property and attach it to this PR

docs/src/main/asciidoc/writing-extensions.adoc
```diff
@@ -1238,6 +1238,130 @@ HealthBuildItem addHealthCheck(AgroalBuildTimeConfig agroalBuildTimeConfig) {
...
+wrapping all code that does something metric-related into an `if` condition that checks whether metrics integration is enabled. If the library performs injections
+of the `MetricRegistry`, which is not avoidable by introducing an `if` condition, it will unfortunately not be possible to remove the `microprofile-metrics` dependency,
+but it is still possible to avoid the need for `smallrye-metrics` (which contains the implementation class of the metric registry) by introducing a custom "no-op"
+implementation of the `MetricRegistry` which can, for example, return `null` from all its methods. This no-op implementation should be added to the application's classes
+so that it can be injected instead of the regular `io.smallrye.metrics.MetricRegistryImpl`.
+
+==== Case 2: The library provides its own metric API
+In this case, the extension can make use of the `quarkus-smallrye-metrics-spi` module to expose the metrics from the library using MP Metrics.
+
+The extension author should to the following:
```
gunnarmorling:
```suggestion
The extension author should do the following:
```

docs/src/main/asciidoc/writing-extensions.adoc
```diff
@@ -1238,6 +1238,130 @@ HealthBuildItem addHealthCheck(AgroalBuildTimeConfig agroalBuildTimeConfig) {
...
+- In your processor, produce any number of `MetricBuildItem` items as shown in the example below.
+The metric name of all metrics should start with a short name of the extension that owns them.
+- Add a way to disable the extension's metrics via a config item `quarkus.<extension>.metrics.enabled` that should be enabled by default. The value
+of this property will be passed to all `MetricBuildItem` objects that the extension produces and if it is false, the metric will be ignored.
+
+An example build step for build-time registration of metrics:
+
+[source%nowrap.java]
+----
+@BuildStep
```
gunnarmorling: @lhauspie, what's your take on this? IIUC, it'd require the KafkaStreams extension to list all the metrics explicitly, instead of the generic implementation we currently have (in the blog post). It seems that'd be quite a bit more work. WDYT?

docs/src/main/asciidoc/writing-extensions.adoc
```diff
@@ -1238,6 +1238,130 @@ HealthBuildItem addHealthCheck(AgroalBuildTimeConfig agroalBuildTimeConfig) {
...
+
+In this case, be aware that the MP Metrics API and implementation might not be available in the runtime classpath!
+This is why there is a `if(metricsEnabled)` check that should make sure that if metrics are not enabled, we don't touch MP Metrics classes,
+because that would fail. The extension itself will need to provide such check, and it must return true only if both of these conditions hold:
+
+- The Metrics capability is present and therefore the SmallRye Metrics library is available
+- Metrics are not disabled for this extension
+
+A recommended way to implement this check would be to produce a specific `SystemPropertyBuildItem` during build, where it is easy
+to evaluate these two conditions, and then check back to this system property at runtime.
```
gunnarmorling: Perhaps show an implementation of `metricsEnabled`?

extensions/agroal/runtime/src/main/java/io/quarkus/agroal/metrics/AgroalCounter.java
```diff
@@ -0,0 +1,95 @@
...
+    public String getDataSourceName() {
+        return dataSourceName;
+    }
+
+    public void setDataSourceName(String dataSourceName) {
+        this.dataSourceName = dataSourceName;
+    }
+
+    private AgroalDataSource getDataSource() {
+        if (dataSource == null) {
```
gunnarmorling: This double-check locking idiom makes me think `dataSource` should be volatile and also be read into a local variable.

gunnarmorling (commented):
Some comments inline.

extensions/agroal/runtime/src/main/java/io/quarkus/agroal/metrics/AgroalCounter.java
loicmathieu: +1 `dataSource` must be volatile here. 
But I will avoid using synchronization here and go for getting the datasource from an `@PostConstruct` block

jmartisk: I guess you're right. I'm not sure if there's any better way to do this, I would need a callback that gets invoked after the Arc container is started, but before any metric requests arrive. I don't want to turn this into a CDI bean just for the purpose of being able to do `@PostConstruct`, due to the performance penalty, and also potentially breaking user applications by unexpectedly introducing some CDI beans that implement `Gauge` or `Counter`. I will fix the synchronization for now, perhaps we will find a better solution where this is not needed.

jmartisk:
Summary of the recent changes that I just did:
- `metrics-enabled` (enables collection of metrics internally in agroal) is now a build time property and it is now an `Optional`. If unspecified, it will be treated as `true` only if MP Metrics are enabled (which are enabled by default if the `smallrye-metrics` extension is present). So I also removed the WARN log that was in place for the cases where MP Metrics are enabled, but Agroal metrics are disabled for some data sources.
- Fixed synchronization in Agroal metric classes per @gunnarmorling 's comment
- Updated the docs with an example how to produce a system property to be used by frameworks to detect whether they should collect metrics. This references `Capabilities.METRICS` capability so this PR is now basically dependent on https://github.com/quarkusio/quarkus/pull/4987

docs/src/main/asciidoc/writing-extensions.adoc
jmartisk: fixed in my commit

docs/src/main/asciidoc/writing-extensions.adoc
jmartisk: I added an example how to produce the system property

extensions/agroal/runtime/src/main/java/io/quarkus/agroal/metrics/AgroalCounter.java
```diff
@@ -0,0 +1,98 @@
...
+    public String getDataSourceName() {
+        return dataSourceName;
+    }
+
+    public void setDataSourceName(String dataSourceName) {
+        this.dataSourceName = dataSourceName;
+    }
+
+    private AgroalDataSource getDataSource() {
+        AgroalDataSource dsLocal = dataSource;
```
loicmathieu: It's ot mandatory to store the attribute inside a local variable. Volatile is enought to make it correct.
See the last example on https://www.cs.umd.edu/~pugh/java/memoryModel/DoubleCheckedLocking.html

loicmathieu:
@jmartisk moving the property from runtime to build time is a breaking change so I added the noteworthy label. Anyway adding new metrics worth a note :)

The only concern that I still have is the fact that the metrics are enabled by default. Computing metrics have a cost that is way more than health check. So even if the user enables metrics I don't think each extension should enable it's metrics by default.
I can totaly think of enabling metrics inside my application but didn't care of metrics from datasource or Kafka.

jmartisk:
@loicmathieu  So your concern is that if a user adds the extension, he might not want to have metrics for all extensions, but just some selected ones for which they will enable them? I don't know, that could be a valid point.
What do others think about this?

There's currently an issue that the agroal internal metrics get enabled automatically even if `smallrye-metrics` extension is not there, that's why the CI failed. I am planning to fix that. The problem is that to determine whether they should be enabled, it's no longer enough to just know the value of `quarkus.datasource.DS_NAME.enable-metrics` property, you also need to know whether the metrics capability is present... But if I add a check for this, this PR will become fully dependent on https://github.com/quarkusio/quarkus/pull/4987  so I would like if we could speed up merging that one :)

extensions/agroal/runtime/src/main/java/io/quarkus/agroal/metrics/AgroalCounter.java
jmartisk: I got this from the interview with Joshua Bloch https://www.oracle.com/technical-resources/articles/javase/bloch-effective-08-qa.html, also the book Effective Java says that thanks to the local variable you get about a performance gain, because you only need to access the volatile field once if it's already non-null.  From wikipedia:
> Note the local variable "localRef", which seems unnecessary. The effect of this is that in cases where helper is already initialized (i.e., most of the time), the volatile field is only accessed once (due to "return localRef;" instead of "return helper;"), which can improve the method's overall performance by as much as 25 percent.

loicmathieu: I never thought about these 2 volatile reads, I really need to read again Effective Java ;)
Thanks for the pointers !

jmartisk:
- Fixed the issue with metrics collection getting enabled even if `quarkus-smallrye-metrics` is not present. This however means this PR will not compile until merging https://github.com/quarkusio/quarkus/pull/4987 and rebasing
- Added some text about metrics into Agroal docs

jmartisk:
Applied a change so that metrics from extensions should be disabled by default.
Rebased on top of the PR on which this depends so it will compile now.
Squashed commits into some reasonable chunks.
I think this should be ready to evolve from draft

loicmathieu (approved):
As metrics are disabled by default for extension, we need a way to tell the user that it exist.

I'm thinking of two things that we can do:

- Explain on the metric page that some extension may provides metrics and that this needs to be enabled (like we did for health but saying it's disabled by default: https://quarkus.io/guides/microprofile-health#extension-health-checks)
- Add a log in INFO in an extension providing metrics when smallrye-metrics is included but the extension metrics are disabled.

Anyway, it's all good to me, good job ;)

emmanuelbernard:
I think the metrics documentation should be in each extension providing these metrics, that's where you want to see them I suppose.

jmartisk:
Yeah I think each extension that supports it can have a short paragraph about it in its documentation. I included exactly that in the agroal documentation.

loicmathieu:
@emmanuelbernard @jmartisk keep in minds that developers often don't read documentation untill the end ... I don't :) Usually I read the first section then start coding and go back to documentation if I cannot make something works by myself. So some (a lot) may miss the section on the extension documentation.

It's important to also have a section, early in the metrics guide (maybe before the example application), explaining that a lot of Quarkus extension provides metrics that are disabled by default and the before implementing your own metrics you should check the extension documentation ...

emmanuelbernard:
Sure can double reference.

jmartisk:
Ok, added a short paragraph to the beginning of the general metrics guide

jmartisk:
Hmm, CI started failing due to dependency divergence. I'll release SmallRye Metrics 2.3.2 to align the MP Metrics API dependency correctly. Should be ready within a few hours.

jmartisk:
After upgrading to SR-M 2.3.2, I faced another divergence problem, this time for Fault Tolerance extension depending on MP Metrics 2.2 (while the rest is at 2.2.1). I employed a not very nice solution to this by overriding the dependency in the Fault Tolerance extension to keep the enforcer plugin happy. I also raised a thread in SmallRye group (https://groups.google.com/forum/#!topic/smallrye/TitSifqEXvU) to find a better long term strategy to avoid having to do these things.

jmartisk:
I was hoping that after my paternity leave, this would be merged :) anyone have any more comments on it?

loicmathieu:
@jmartisk I already approve the PR so all is OK for me

jmartisk:
I just removed the workaround mentioned in my comment  because we now have updated SR Fault Tolerance that aligns correctly.
Would be nice to have this acked+merged soon because the work to get metrics working in other extensions depends on this, I'd like to start working on them. @emmanuelbernard does this have your ack?

emmanuelbernard:
Sorry guys, I lost track a bit so don't wait for me. I'll remove me from the equation by acking the review

emmanuelbernard:
@jmartisk but... are you talking about the fault tolerance that no longer use Hystrix that @michalszynkiewicz mentioned? Because that one will  land in 1.2. so that PR will do too.

jmartisk:
@emmanuelbernard No, in this case this was just a micro update to SR-FaultTolerance 2.1.3, this still depends on Hystrix

gunnarmorling:
Hey there @loicmathieu, @lhauspie, has either of you already looked into (or is planning to) using this for Kafka Streams?

jmartisk:
I already have metrics for Hibernate ORM almost ready, they depend on this mechanism too, so I would really appreciate if we could merge this soon.
CI failures are unrelated

jmartisk:
Rebased and fixed a conflict. @lhauspie have you had a chance to look please? I'm growing a bit impatient here :)

gsmet:
@jmartisk I'll have a look today.

docs/src/main/asciidoc/datasource.adoc
```diff
@@ -265,6 +265,24 @@ If you have multiple datasources, all datasources will be checked and the status
...
+
+For the exposed metrics to contain any actual values, it is necessary that metric collection is enabled internally
+by Agroal mechanisms. By default, this metric collection mechanism gets turned on for all data sources if the `quarkus-smallrye-metrics`
+is present and metrics for the Agroal extension are enabled. If you want to disable metrics for a particular data source,
+this can be done by setting `quarkus.datasource.DATASOURCE_NAME.enable-metrics` to `false`. This disables
+collecting the metrics as well as exposing them in the `/metrics` endpoint, because it does not make sense to
+expose metrics if the mechanism to collect them is disabled.
+
+Conversely, setting `quarkus.datasource.DATASOURCE_NAME.enable-metrics` to `true` explicitly can be used to enable collecting of metrics even if
+`quarkus-smallrye-metrics` extension is not in use. This can be useful if you need to access the collected metrics programmatically.
```
gsmet:
```suggestion
the `quarkus-smallrye-metrics` extension is not in use. This can be useful if you need to access the collected metrics programmatically.
```

docs/src/main/asciidoc/datasource.adoc
```diff
@@ -265,6 +265,24 @@ If you have multiple datasources, all datasources will be checked and the status
...
 
+== Datasource Metrics
+
+If you are using the `quarkus-smallrye-metrics` extension, `quarkus-agroal` can expose some data source metrics on the
+`/metrics` endpoint. This can be turned on by setting the property `quarkus.datasource.metrics.enabled` to true.
+
+For the exposed metrics to contain any actual values, it is necessary that metric collection is enabled internally
+by Agroal mechanisms. By default, this metric collection mechanism gets turned on for all data sources if the `quarkus-smallrye-metrics`
+is present and metrics for the Agroal extension are enabled. If you want to disable metrics for a particular data source,
+this can be done by setting `quarkus.datasource.DATASOURCE_NAME.enable-metrics` to `false`. This disables
```
gsmet:
```suggestion
this can be done by setting `quarkus.datasource.enable-metrics` to `false` (or `quarkus.datasource.<datasource name>.enable-metrics` for a named datasource). This disables
```

docs/src/main/asciidoc/datasource.adoc
```diff
@@ -265,6 +265,24 @@ If you have multiple datasources, all datasources will be checked and the status
...
+collecting the metrics as well as exposing them in the `/metrics` endpoint, because it does not make sense to
+expose metrics if the mechanism to collect them is disabled.
+
+Conversely, setting `quarkus.datasource.DATASOURCE_NAME.enable-metrics` to `true` explicitly can be used to enable collecting of metrics even if
```
gsmet:
```suggestion
Conversely, setting `quarkus.datasource.enable-metrics` to `true` (or `quarkus.datasource.<datasource name>.enable-metrics` for a named datasource) explicitly can be used to enable collecting of metrics even if
```

docs/src/main/asciidoc/writing-extensions.adoc
```diff
@@ -1290,6 +1290,143 @@ HealthBuildItem addHealthCheck(AgroalBuildTimeConfig agroalBuildTimeConfig) {
...
+=== Extension Metrics
+
+An extension can decide to provide metrics through the `quarkus-smallrye-metrics` extension.
+A typical use case for this would be that an extension scans the application code for relevant components (like entities, messaging endpoints, etc.)
+and creates a set of metrics for each of these components. There is a unified mechanism for metric registration - it uses the `MetricBuildItem` class
+provided by the `quarkus-smallrye-metrics-spi` module.
+
+There are several distinct situations that can happen and each requires a bit different handling:
+
+1. The underlying library used by your extension is using MP Metrics API directly.
```
gsmet:
```suggestion
1. The underlying library used by your extension is using the MicroProfile Metrics API directly.
```
(for the first occurrence, this might not be obvious to the reader)

docs/src/main/asciidoc/writing-extensions.adoc
```diff
@@ -1290,6 +1290,143 @@ HealthBuildItem addHealthCheck(AgroalBuildTimeConfig agroalBuildTimeConfig) {
...
+it will still work when the MP Metrics dependencies (or at least the implementation, `io.smallrye:smallrye-metrics`) are unavailable. This can be achieved by
+wrapping all code that does something metric-related into an `if` condition that checks whether metrics integration is enabled. If the library performs injections
+of the `MetricRegistry`, which is not avoidable by introducing an `if` condition, it will unfortunately not be possible to remove the `microprofile-metrics` dependency,
+but it is still possible to avoid the need for `smallrye-metrics` (which contains the implementation class of the metric registry) by introducing a custom "no-op"
+implementation of the `MetricRegistry` which can, for example, return `null` from all its methods. This no-op implementation should be added to the application's classes
+so that it can be injected instead of the regular `io.smallrye.metrics.MetricRegistryImpl`. An example how this was implemented can be found
+by looking at the SmallRye Fault Tolerance library and `smallrye-fault-tolerance` Quarkus extension.
```
gsmet: Maybe point to the appropriate class in the GitHub repo?

I know it can be broken in the future but it seems a bit too hard to find it so I would do it anyway.

docs/src/main/asciidoc/writing-extensions.adoc
```diff
@@ -1290,6 +1290,143 @@ HealthBuildItem addHealthCheck(AgroalBuildTimeConfig agroalBuildTimeConfig) {
...
+
+==== Case 2: The library provides its own metric API
+In this case, the extension can make use of the `quarkus-smallrye-metrics-spi` module to expose the metrics from the library using MP Metrics.
+
+The extension author should do the following:
```
gsmet:
```suggestion
You should do the following:
```
(you're using `your` everywhere after that). Another option would be to use `Extension authors` and move to `they/their/them`. Not sure how it is in the rest of the document.

extensions/agroal/deployment/src/main/java/io/quarkus/agroal/deployment/AgroalProcessor.java
```diff
@@ -311,4 +325,200 @@ HealthBuildItem addHealthCheck(AgroalBuildTimeConfig agroalBuildTimeConfig) {
         return new HealthBuildItem("io.quarkus.agroal.runtime.health.DataSourceHealthCheck",
                 agroalBuildTimeConfig.healthEnabled, "datasource");
     }
+
+    @BuildStep
+    void registerMetrics(AgroalBuildTimeConfig agroalBuildTimeConfig,
+            BuildProducer<MetricBuildItem> metrics) {
+        Metadata activeCountMetadata = Metadata.builder()
+                .withName("agroal.active.count")
+                .withDescription("Number active of connections. This connections are in use and not available to be acquired.")
```
gsmet:
```suggestion
                .withDescription("Number of active connections. These connections are in use and not available to be acquired.")
```

extensions/agroal/runtime/src/main/java/io/quarkus/agroal/metrics/AgroalCounter.java
```diff
@@ -0,0 +1,98 @@
...
+            case "destroyCount":
+                return metrics.destroyCount();
+            case "flushCount":
+                return metrics.flushCount();
+            case "invalidCount":
+                return metrics.invalidCount();
+            case "reapCount":
+                return metrics.reapCount();
+            default:
+                throw new IllegalArgumentException("Unknown data source metric");
```
gsmet:
```suggestion
                throw new IllegalArgumentException("Unknown datasource metric");
```
(I know we're not very consistent with this especially with the class names being `DataSource` (this comes from javax.) but let's stick to `datasource` in text)

extensions/agroal/runtime/src/main/java/io/quarkus/agroal/runtime/DataSourceBuildTimeConfig.java
```diff
@@ -22,4 +22,11 @@
     @ConfigItem(defaultValue = "enabled")
     public TransactionIntegration transactions;
 
+    /**
+     * Enable datasource metrics collection. If unspecified, collecting metrics will be enabled by default if the
+     * smallrye-metrics extension is active.
+     */
+    @ConfigItem
+    public Optional<Boolean> enableMetrics;
```
gsmet: So we don't have a way to disable metrics at runtime?

It seems like a very common request.

gsmet (changes requested):
I added a couple of suggestions and (I think) an important question.

extensions/agroal/runtime/src/main/java/io/quarkus/agroal/runtime/DataSourceBuildTimeConfig.java
jmartisk: Agroal seems to support disabling metrics dynamically, but I made this a build time property after discussing with Emmanuel who thinks we don't need that: https://github.com/quarkusio/quarkus/pull/5272#issuecomment-555106563

extensions/agroal/deployment/src/main/java/io/quarkus/agroal/deployment/AgroalProcessor.java
jmartisk: Good catch, I admit I was mostly just copying javadoc from Hibernate codebase without thinking too hard about it. There are some typos apparently.

docs/src/main/asciidoc/writing-extensions.adoc
jmartisk: Ok, I added a link to the `NoopMetricRegistry` class

jmartisk:
Hey @gsmet I've addressed your comments, does it look better to you now, please?

geoand:
@jmartisk just to be on the safe side, can you please rebase onto the latest master?

docs/src/main/asciidoc/writing-extensions.adoc
```diff
@@ -1290,6 +1290,144 @@ HealthBuildItem addHealthCheck(AgroalBuildTimeConfig agroalBuildTimeConfig) {
...
+public void methodThatShouldBeCounted() {
+    if(metricsEnabled) {
+        Counter counter = MetricRegistries.get(MetricRegistry.Type.VENDOR).counter("name");
+        counter.inc();
+    }
+    // proceed with the invocation normally
+}
+----
+
+In this case, be aware that the MP Metrics API and implementation might not be available in the runtime classpath!
```
geoand:
```suggestion
In this case, be aware that the MP Metrics API and implementation might not be available on the runtime classpath!
```

docs/src/main/asciidoc/writing-extensions.adoc
```diff
@@ -1290,6 +1290,144 @@ HealthBuildItem addHealthCheck(AgroalBuildTimeConfig agroalBuildTimeConfig) {
...
+This is why there is a `if(metricsEnabled)` check that should make sure that if metrics are not enabled, we don't touch MP Metrics classes,
+because that would fail. The extension itself will need to provide such check, and it must return true only if both of these conditions hold:
+
+- The Metrics capability is present and therefore the SmallRye Metrics library is available
+- Metrics are not disabled for this extension
+
+A recommended way to implement this check would be to produce a specific `SystemPropertyBuildItem` during build, which will make it is easy
```
geoand:
```suggestion
The recommended way to implement this check would be to produce a specific `SystemPropertyBuildItem` during build, which will make it easy
```

docs/src/main/asciidoc/datasource.adoc
```diff
@@ -265,6 +265,24 @@ If you have multiple datasources, all datasources will be checked and the status
...
+`/metrics` endpoint. This can be turned on by setting the property `quarkus.datasource.metrics.enabled` to true.
+
+For the exposed metrics to contain any actual values, it is necessary that metric collection is enabled internally
+by Agroal mechanisms. By default, this metric collection mechanism gets turned on for all data sources if the `quarkus-smallrye-metrics`
+is present and metrics for the Agroal extension are enabled. If you want to disable metrics for a particular data source,
+this can be done by setting `quarkus.datasource.enable-metrics` to `false` (or `quarkus.datasource.<datasource name>.enable-metrics` for a named datasource). This disables
+collecting the metrics as well as exposing them in the `/metrics` endpoint, because it does not make sense to
+expose metrics if the mechanism to collect them is disabled.
+
+Conversely, setting `quarkus.datasource.enable-metrics` to `true` (or `quarkus.datasource.<datasource name>.enable-metrics` for a named datasource) explicitly can be used to enable collecting of metrics even if
```
geoand:
```suggestion
Conversely, setting `quarkus.datasource.enable-metrics` to `true` (or `quarkus.datasource.<datasource name>.enable-metrics` for a named datasource) explicitly can be used to enable collection of metrics even if
```

geoand (commented):
I added some suggestions on the docs side of things. I'll also check out the code but I am pretty sure that after all the discussions things will be in place :)

geoand:
I also added the `breaking-change` label since the metrics property for the datasource config has moved to build time

extensions/agroal/runtime/src/main/java/io/quarkus/agroal/runtime/DataSourceBuildTimeConfig.java
geoand: I don't know if we should actually be able to turn it off at runtime given everything the user has to do to actually turn it on :). I mean it seems like the user really wants it explicitly. 

Now I am sure there scenaria were someone will want to turn it off, but perhaps we can keep it as is for now and reexamine if we have such a request in the future?

jmartisk: Yeah I guess, I don't see much usefulness in dynamically turning off something that you had to explicitly turn on in the first place. 
And especially since turning off metric collection internally will mean that the already collected exposed metrics will suddenly "freeze" at their current values, which can become a pitfall, unless we add a hook that will also de-register the externally exposed Metrics as soon as this gets turned off.

geoand (approved):
I'm good with the state of this, thanks @jmartisk.

We probably want to squash the commits however

jmartisk:
Just working on squashing and rebasing, give me 3 minutes :)

jmartisk:
Done, I addressed @geoand 's comments, squashed it back into 3 commits and rebased against current master

geoand:
Excellent, thanks!

extensions/agroal/runtime/src/main/java/io/quarkus/agroal/runtime/DataSourceBuildTimeConfig.java
gsmet: So FTR, my use case was: metrics are not free, if you're in a constrained environment, you might want to disable them most of the time and enable them in case something is going wrong.

But yes, it can probably wait.

gsmet (approved):

gsmet:
@geoand I approved it too. I let you merge it?

geoand:
Sure thing @gsmet!

PR #5278, opened 2019-11-07
Remove the lambda use in RESTEasy's VertxRequestHandler

Sanne:
PR cleaned up as follow up to comments #5220 (which was merged already, so new pr..)


gwenneg (approved):
LGTM

Sanne:
thanks @gwenneg !

Issue #5115, opened 2019-11-01
quarkus-amazon-dynamodb and quarkus-smallrye-fault-tolerance combination causes CNFE


rsvoboda:
quarkus-amazon-dynamodb and quarkus-smallrye-fault-tolerance combination causes CNFE
java.lang.ClassNotFoundException: software.amazon.awssdk.http.urlconnection.UrlConnectionHttpClient$Builder

url-connection-client is marked as optional - https://github.com/quarkusio/quarkus/blob/master/extensions/amazon-dynamodb/runtime/pom.xml#L45

Steps to reproduce:
use microprofile-fault-tolerance-quickstart https://github.com/quarkusio/quarkus-quickstarts/tree/development/microprofile-fault-tolerance-quickstart
```
cd microprofile-fault-tolerance-quickstart
mvn quarkus:add-extension -Dextensions="quarkus-amazon-dynamodb"
mvn clean verify
```

```
[INFO] Running org.acme.faulttolerance.CoffeeResourceTest
2019-11-01 13:40:34,689 INFO  [io.qua.dep.QuarkusAugmentor] (main) Beginning quarkus augmentation
2019-11-01 13:40:35,969 INFO  [io.qua.resteasy] (build-22) Resteasy running without servlet container.
2019-11-01 13:40:35,969 INFO  [io.qua.resteasy] (build-22) - Add quarkus-undertow to run Resteasy within a servlet container
2019-11-01 13:40:36,007 INFO  [io.qua.dep.QuarkusAugmentor] (main) Quarkus augmentation completed in 1318ms
java.lang.NoClassDefFoundError: software/amazon/awssdk/http/urlconnection/UrlConnectionHttpClient$Builder
	at java.base/java.lang.Class.getDeclaredMethods0(Native Method)
	at java.base/java.lang.Class.privateGetDeclaredMethods(Class.java:3167)
	at java.base/java.lang.Class.getDeclaredMethods(Class.java:2310)
	at io.quarkus.smallrye.faulttolerance.runtime.SmallryeFaultToleranceRecorder.getMethodsForValidation(SmallryeFaultToleranceRecorder.java:93)
	at io.quarkus.smallrye.faulttolerance.runtime.SmallryeFaultToleranceRecorder.validate(SmallryeFaultToleranceRecorder.java:52)
[...]

PR #5280, opened 2019-11-07
Fix for #5115 - Do not fire DynamoDB recorder if no DynamoDB clients used

marcinczeczko:
Do not fire DynamoDB recorder if no DynamoDB clients used


gsmet (approved):

gsmet:
Looks good, thanks @marcinczeczko .

Issue #5282, opened 2019-11-07
narayana-jta extension: Improve the way narayana properties are handled


mmusgrov:
The extension has some issues around how the Narayana transaction manger properties are initialised:

1. It records the default properties at build time - Narayana normally sets config properties using reflective access to setters on various beans. Version 5.10.0.Final of Narayana adds a method for directly passing pre-configured beans (and this will avoid the need for reflective access code).
2. It sets the Narayana property factory by reflectively changing the Java access mode of a private field that holds the factory.
3. The extension hard codes the properties in the Narayana ConfigurationInfo bean. The previous version of Narayana set these properties by looking inside the Narayana Java archive which is not efficient in Quarkus. The latest version of Narayana sets the properties directly in the code as constants. Therefore the extension no longer needs to hard code the properties in the ConfigurationInfoSubstitution extension class

gastaldi:
Closing for now, as it had too much pushback on https://github.com/quarkusio/quarkus/pull/5283, please reopen if it's still valid

PR #5283, opened 2019-11-07
Improve the way that the extension handles Narayana configuration properties

mmusgrov:
https://github.com/quarkusio/quarkus/issues/5282

This PR makes use of new features added to version of narayana (5.10.0.Final) which has already been merged. The changes means that the extension:

- Does not need to record the default properties at build time - Narayana normally sets config
  properties using reflective access to setters on various beans. But this latest Narayana version
  adds a method for directly passing pre-configured beans (thus avoiding the relection code).
- Does not need to set its own property factory. The previous vesion of the extension did this
  by reflectively changing the Java access mode of a private field that holds the factory.
- removes the hardcoding of the properties in the Narayana ConfigurationInfo bean
  (earlier versons of Narayana set these proeprties by looking inside the narayana Java archive
  but this version sets directly in the code as constants).

Note that the PR makes the default Narayana properties configurable via the MP config mechanism (see the class NarayanaQuarkusConfiguration in this PR).


Sanne:
hi @mmusgrov , would you mind rebasing this and resolve the conflict?

And please check the native tests, this seems to have broken something:
 - https://dev.azure.com/quarkus-ci/quarkus/_build/results?buildId=11767

Sanne:
Since the upgrade was actually done in the other PR which was already merged, the commit log of this one is now a bit misleading?

cescoffier:
Are we good on this one?

Doesn't it break the API?

extensions/narayana-jta/runtime/src/main/java/io/quarkus/narayana/jta/runtime/NarayanaQuarkusConfiguration.java
```diff
@@ -0,0 +1,147 @@
...
+     */
+    @ConfigItem(name = "CoordinatorEnvironmentBean.asyncAfterSynchronization", defaultValue = "false")
+    public boolean asyncAfterSynchronization;
+    /**
+     * config
+     */
+    @ConfigItem(name = "CoordinatorEnvironmentBean.dynamic1PC", defaultValue = "true")
+    public boolean dynamic1PC;
+
+    //    @ConfigItem(name = "CoordinatorEnvironmentBean.transactionStatusManagerEnable", defaultValue = "false")
```
ochaloup: I assume there should not be a commented out text in the code. If it's not necessary it should be just removed.

extensions/narayana-jta/runtime/src/main/java/io/quarkus/narayana/jta/runtime/NarayanaJtaRecorder.java
```diff
@@ -32,34 +47,53 @@ public void setNodeName(final TransactionManagerConfiguration transactions) {
...
-    public static Properties getDefaultProperties() {
-        return defaultProperties;
-    }
-
     public void disableTransactionStatusManager() {
         arjPropertyManager.getCoordinatorEnvironmentBean()
                 .setTransactionStatusManagerEnable(false);
     }
+
+    public void setConfig(TransactionManagerConfiguration transactions, NarayanaQuarkusConfiguration config) {
```
ochaloup: would not be better to unite the `TransactionManagerConfiguration` and `NarayanaQuarkusConfiguration` under one class. It would be easier to get where the configuration is defined

extensions/narayana-jta/runtime/src/main/java/io/quarkus/narayana/jta/runtime/NarayanaQuarkusConfiguration.java
```diff
@@ -0,0 +1,147 @@
...
+     */
+    @ConfigItem(name = "ObjectStoreEnvironmentBean.objectStoreDir", defaultValue = "PutObjectStoreDirHere")
+    public String objectStoreDir;
+    /**
+     * config
+     */
+    @ConfigItem(name = "ObjectStoreEnvironmentBean.transactionSync", defaultValue = "true")
+    public boolean transactionSync;
+
+    //    @ConfigItem(name = "CoreEnvironmentBean.nodeIdentifier", defaultValue = "1")
```
ochaloup: I assume this commented text should be removed from the PR.

extensions/narayana-jta/runtime/src/main/java/io/quarkus/narayana/jta/runtime/NarayanaQuarkusConfiguration.java
```diff
@@ -0,0 +1,147 @@
...
+    //    public String nodeIdentifier;
+    /**
+     * config
+     */
+    @ConfigItem(name = "JTAEnvironmentBean.xaRecoveryNodes", defaultValue = "1")
```
ochaloup: I think there should not be configurable to change the `xaRecoveryNodes`. The xa recovery nodes should be changed to what is set in node name. Having the node name and recovery nodes to be the same value is safer as it diminishes possibility of misconfiguration.
See `NarayanaJtaRecorder.setNodeName` method.

It's a way how WFLY integration works.

extensions/narayana-jta/runtime/src/main/java/io/quarkus/narayana/jta/runtime/NarayanaQuarkusConfiguration.java
```diff
@@ -0,0 +1,147 @@
...
+
+import io.quarkus.runtime.annotations.ConfigItem;
+import io.quarkus.runtime.annotations.ConfigPhase;
+import io.quarkus.runtime.annotations.ConfigRoot;
+
+@ConfigRoot(phase = ConfigPhase.RUN_TIME)
+public class NarayanaQuarkusConfiguration {
+    // non final properties
+    /**
+     * config
```
ochaloup: I assume comment /** config **/ may be omitted.

extensions/narayana-jta/runtime/src/main/java/io/quarkus/narayana/jta/runtime/NarayanaQuarkusConfiguration.java
```diff
@@ -0,0 +1,147 @@
...
+     */
+    @ConfigItem(name = "JTAEnvironmentBean.xaRecoveryNodes", defaultValue = "1")
+    public List<String> xaRecoveryNodes;
+    /**
+     * config
+     */
+    @ConfigItem(name = "JTAEnvironmentBean.xaResourceOrphanFilterClassNames", defaultValue = "com.arjuna.ats.internal.jta.recovery.arjunacore.JTATransactionLogXAResourceOrphanFilter,"
+            + "com.arjuna.ats.internal.jta.recovery.arjunacore.JTANodeNameXAResourceOrphanFilter,"
+            + "com.arjuna.ats.internal.jta.recovery.arjunacore.SubordinationManagerXAResourceOrphanFilter")
```
ochaloup: adding `SubordinateJTAXAResourceOrphanFilter` should be considered

extensions/narayana-jta/runtime/src/main/java/io/quarkus/narayana/jta/runtime/NarayanaQuarkusConfiguration.java
```diff
@@ -0,0 +1,147 @@
...
+    public List<String> xaResourceOrphanFilterClassNames;
+    /**
+     * config
+     */
+    @ConfigItem(name = "CoreEnvironmentBean.socketProcessIdPort", defaultValue = "0")
+    public int socketProcessIdPort;
+    /**
+     * config
+     */
+    @ConfigItem(name = "RecoveryEnvironmentBean.recoveryModuleClassNames", defaultValue = "com.arjuna.ats.internal.arjuna.recovery.AtomicActionRecoveryModule,"
```
ochaloup: should not be added the `CommitMarkableResourceRecordRecoveryModule` as well?

extensions/narayana-jta/runtime/src/main/java/io/quarkus/narayana/jta/runtime/NarayanaQuarkusConfiguration.java
```diff
@@ -0,0 +1,147 @@
...
+    public String recoveryAddress;
+    /**
+     * config
+     */
+    @ConfigItem(name = "RecoveryEnvironmentBean.transactionStatusManagerPort", defaultValue = "0")
+    public int transactionStatusManagerPort;
+    /**
+     * config
+     */
+    @ConfigItem(name = "RecoveryEnvironmentBean.transactionStatusManagerAddress")
```
ochaloup: this should not be configurable as the status manager seems to be disabled all the time, see `NarayanaJtaProcessor.java.disableTransactionStatusManager`

extensions/narayana-jta/runtime/src/main/java/io/quarkus/narayana/jta/runtime/NarayanaJtaRecorder.java
```diff
@@ -32,34 +47,53 @@ public void setNodeName(final TransactionManagerConfiguration transactions) {
...
-            log.error("Could not override transaction properties factory", e);
-        }
-
-        defaultProperties = properties;
-    }
-
     public void setDefaultTimeout(TransactionManagerConfiguration transactions) {
         transactions.defaultTransactionTimeout.ifPresent(defaultTimeout -> {
             arjPropertyManager.getCoordinatorEnvironmentBean().setDefaultTimeout((int) defaultTimeout.getSeconds());
             TxControl.setDefaultTimeout((int) defaultTimeout.getSeconds());
```
ochaloup: with this change the `TxControl.setter` can't be called here. When it's called then `TxControl` stays frozen at current state. No changes into beans properties are not taken into consideration.

extensions/narayana-jta/runtime/src/main/java/io/quarkus/narayana/jta/runtime/NarayanaQuarkusConfiguration.java
```diff
@@ -0,0 +1,147 @@
...
+    public int recoveryPort;
+    /**
+     * config
+     */
+    @ConfigItem(name = "RecoveryEnvironmentBean.recoveryAddress")
+    public String recoveryAddress;
+    /**
+     * config
+     */
+    @ConfigItem(name = "RecoveryEnvironmentBean.transactionStatusManagerPort", defaultValue = "0")
```
ochaloup: it should not be configurable as `NarayanaJtaProcessor.java.disableTransactionStatusManager` defines it to be disabled anyway

extensions/narayana-jta/runtime/src/main/java/io/quarkus/narayana/jta/runtime/NarayanaQuarkusConfiguration.java
```diff
@@ -0,0 +1,147 @@
...
+    public boolean readonlyOptimisation;
+    /**
+     * config
+     */
+    @ConfigItem(name = "CoordinatorEnvironmentBean.startDisabled", defaultValue = "false")
+    public boolean startDisabled;
+    /**
+     * config
+     */
+    @ConfigItem(name = "CoordinatorEnvironmentBean.defaultTimeout", defaultValue = "60")
```
ochaloup: is it expected value when we are used to with 5 minutes in Narayana/WFLY?

ochaloup (changes requested):
I added several comments. Some of them I consider sever (e.g. call of the `TxControl` before the beans are set up).

extensions/narayana-jta/runtime/src/main/java/io/quarkus/narayana/jta/runtime/NarayanaJtaRecorder.java
```diff
@@ -1,28 +1,43 @@
...
 
 @Recorder
 public class NarayanaJtaRecorder {
 
-    private static Properties defaultProperties;
+    private static final CoordinatorEnvironmentBean coordinatorEnvironmentBean = new CoordinatorEnvironmentBean();
+    private static final RecoveryEnvironmentBean recoveryEnvironmentBean = new RecoveryEnvironmentBean();
+    private static final CoreEnvironmentBean coreEnvironmentBean = new CoreEnvironmentBean();
+    private static final JTAEnvironmentBean jtaEnvironmentBean = new JTAEnvironmentBean();
+    private static final ObjectStoreEnvironmentBean objectStoreEnvironmentBean = new ObjectStoreEnvironmentBean();
```
gsmet: What's the value of making them static final here instead of initializing them below? It would be nice if could initialize them only if needed btw.

extensions/narayana-jta/runtime/src/main/java/io/quarkus/narayana/jta/runtime/NarayanaQuarkusConfiguration.java
```diff
@@ -0,0 +1,147 @@
...
+@ConfigRoot(phase = ConfigPhase.RUN_TIME)
+public class NarayanaQuarkusConfiguration {
+    // non final properties
+    /**
+     * config
+     */
+    @ConfigItem(name = "CoordinatorEnvironmentBean.commitOnePhase", defaultValue = "true")
+    public boolean commitOnePhase;
+    /**
+     * config
```
gsmet: We need proper Javadoc please.

extensions/narayana-jta/runtime/src/main/java/io/quarkus/narayana/jta/runtime/NarayanaQuarkusConfiguration.java
```diff
@@ -0,0 +1,147 @@
...
+     */
+    @ConfigItem(name = "ObjectStoreEnvironmentBean.objectStoreDir", defaultValue = "PutObjectStoreDirHere")
```
gsmet: The default value shouldn't be a random thing.

extensions/narayana-jta/runtime/src/main/java/io/quarkus/narayana/jta/runtime/NarayanaQuarkusConfiguration.java
```diff
@@ -0,0 +1,147 @@
...
+import io.quarkus.runtime.annotations.ConfigPhase;
+import io.quarkus.runtime.annotations.ConfigRoot;
+
+@ConfigRoot(phase = ConfigPhase.RUN_TIME)
+public class NarayanaQuarkusConfiguration {
+    // non final properties
+    /**
+     * config
+     */
+    @ConfigItem(name = "CoordinatorEnvironmentBean.commitOnePhase", defaultValue = "true")
```
gsmet: The keys do not follow our configuration guidelines at all. It's definitely not something we want.

gsmet (changes requested):
This patch is definitely not ready:
- we need proper javadoc for the config
- the config item names are not following our conventions
- there is at least one random default value,  I haven't checked them all, that's definitely not something we want.

extensions/narayana-jta/runtime/src/main/java/io/quarkus/narayana/jta/runtime/NarayanaQuarkusConfiguration.java
mmusgrov: Okay. The initial PR duplicates the default properties bundled in with the Narayana jar.
Should we still set the property but not allow the user to reconfigure it.

extensions/narayana-jta/runtime/src/main/java/io/quarkus/narayana/jta/runtime/NarayanaQuarkusConfiguration.java
mmusgrov: So do you think I should be taking most of the defaults from what we set in the wildfly transactions subsystem?

extensions/narayana-jta/runtime/src/main/java/io/quarkus/narayana/jta/runtime/NarayanaQuarkusConfiguration.java
mmusgrov: +1

extensions/narayana-jta/runtime/src/main/java/io/quarkus/narayana/jta/runtime/NarayanaJtaRecorder.java
mmusgrov: This is in the recorder phase during build time so it gets baked into the executable.

extensions/narayana-jta/runtime/src/main/java/io/quarkus/narayana/jta/runtime/NarayanaQuarkusConfiguration.java
mmusgrov: +1

extensions/narayana-jta/runtime/src/main/java/io/quarkus/narayana/jta/runtime/NarayanaQuarkusConfiguration.java
mmusgrov: Same comment as before: do you think I should be taking most of the defaults from what we set in the wildfly transactions subsystem?

extensions/narayana-jta/runtime/src/main/java/io/quarkus/narayana/jta/runtime/NarayanaJtaRecorder.java
mmusgrov: I agree it is not ready. I wanted to get something ready for discussion.

extensions/narayana-jta/runtime/src/main/java/io/quarkus/narayana/jta/runtime/NarayanaQuarkusConfiguration.java
mmusgrov: +1 Once we have agreed on which properties should be configurable then I will add it.
For properties that are not configurable then they need to be set a build time.

extensions/narayana-jta/runtime/src/main/java/io/quarkus/narayana/jta/runtime/NarayanaQuarkusConfiguration.java
mmusgrov: Fair enough. We use that so that users are aware that they need to configure the location.

extensions/narayana-jta/runtime/src/main/java/io/quarkus/narayana/jta/runtime/NarayanaQuarkusConfiguration.java
mmusgrov: Okay, in this initial PR I wanted to keep everything as close as possible to what narayana does. Once we have agreed on what we want to be configurable I will add sensible defaults. I asked above if I should be taking the initial set of defaults from what we do on wildfly and then modify based on the specifics of quarkus.

extensions/narayana-jta/runtime/src/main/java/io/quarkus/narayana/jta/runtime/NarayanaQuarkusConfiguration.java
mmusgrov: Again, I just duplicated our own defaults. If this property makes sense for quarkus then +1;
Is this property part of your analysis of the orphan filters that is being discussed on our forum?

extensions/narayana-jta/runtime/src/main/java/io/quarkus/narayana/jta/runtime/NarayanaQuarkusConfiguration.java
mmusgrov: Yes I was waiting for the review to happen before deciding on which properties we want to be configurable and which ones we want setting at build time. The initial PR just takes all of the existing defaults and makes them configurable. It is a basis for us to decide which should be set at build time and which ones we want to be configurable.

mmusgrov:
@gsmet @ochaloup This one is ready for another review.
The transaction manager has many config properties and we do not want to expose all of them so I have kept the config minimal. But I have allowed any Narayana TM config option to be configurable via a java system property (like we do in Narayana and WildFly). 

As we get user feedback we should continue adding quarkus config properties but let's keep it minimal for now. For example my next PR will add full config for the transaction log stores (ie the jdbc and artemis journal stores).

extensions/narayana-jta/runtime/src/main/java/io/quarkus/narayana/jta/runtime/NarayanaJtaRecorder.java
mmusgrov: +1

extensions/narayana-jta/runtime/src/main/java/io/quarkus/narayana/jta/runtime/NarayanaQuarkusConfiguration.java
mmusgrov: +1

extensions/narayana-jta/runtime/src/main/java/io/quarkus/narayana/jta/runtime/NarayanaQuarkusConfiguration.java
mmusgrov: +1

extensions/narayana-jta/runtime/src/main/java/io/quarkus/narayana/jta/runtime/NarayanaQuarkusConfiguration.java
mmusgrov: +1

extensions/narayana-jta/runtime/src/main/java/io/quarkus/narayana/jta/runtime/NarayanaJtaRecorder.java
mmusgrov: -1

extensions/narayana-jta/runtime/src/main/java/io/quarkus/narayana/jta/runtime/NarayanaQuarkusConfiguration.java
mmusgrov: +1

extensions/narayana-jta/runtime/src/main/java/io/quarkus/narayana/jta/runtime/NarayanaJtaRecorder.java
```diff
@@ -32,34 +31,115 @@ public void setNodeName(final TransactionManagerConfiguration transactions) {
...
+        recoveryEnvironmentBean.setRecoveryPort(4712);
+    }
+
+    private void setRecoveryListener(RecoveryEnvironmentBean recoveryEnvironmentBean) {
+        recoveryEnvironmentBean.setRecoveryListener(true);
+    }
+
+    public void setConfig(TransactionManagerConfiguration transactions) {
+        Properties systemProperties;
+        final SecurityManager sm = System.getSecurityManager();
```
ochaloup: by interest, is the `SecurityManger` considered to need to be handled in Quarkus code when it's in general one app per one JVM system?

extensions/narayana-jta/runtime/src/main/java/io/quarkus/narayana/jta/runtime/NarayanaJtaRecorder.java
```diff
@@ -32,34 +31,115 @@ public void setNodeName(final TransactionManagerConfiguration transactions) {
...
-    }
-
     public void disableTransactionStatusManager() {
         arjPropertyManager.getCoordinatorEnvironmentBean()
                 .setTransactionStatusManagerEnable(false);
     }
+
+    private void setXaResourceOrphanFilterClassNames(JTAEnvironmentBean jtaEnvironmentBean) {
+        jtaEnvironmentBean.setXaResourceOrphanFilterClassNames(Arrays.asList(
+                "com.arjuna.ats.internal.jta.recovery.arjunacore.JTATransactionLogXAResourceOrphanFilter",
```
ochaloup: would not be better to use the `...class.getName()` instead of the `String` representation of the class name? I just think there could be hidden typo here.

extensions/narayana-jta/runtime/src/main/java/io/quarkus/narayana/jta/runtime/NarayanaJtaRecorder.java
```diff
@@ -32,34 +31,115 @@ public void setNodeName(final TransactionManagerConfiguration transactions) {
...
+                "com.arjuna.ats.internal.jta.recovery.arjunacore.SubordinationManagerXAResourceOrphanFilter"));
+    }
+
+    private void setRecoveryModuleClassNames(RecoveryEnvironmentBean recoveryEnvironmentBean) {
+        recoveryEnvironmentBean.setRecoveryModuleClassNames(Arrays.asList(
+                "com.arjuna.ats.internal.arjuna.recovery.AtomicActionRecoveryModule",
+                "com.arjuna.ats.internal.txoj.recovery.TORecoveryModule",
+                "com.arjuna.ats.internal.jta.recovery.arjunacore.SubordinateAtomicActionRecoveryModule",
+                "com.arjuna.ats.internal.jta.recovery.arjunacore.XARecoveryModule",
+                "com.arjuna.ats.internal.jta.recovery.arjunacore.CommitMarkableResourceRecordRecoveryModule"));
```
ochaloup: this has to be first, https://github.com/wildfly/wildfly/blob/master/transactions/src/main/java/org/jboss/as/txn/service/ArjunaRecoveryManagerService.java#L107

extensions/narayana-jta/runtime/src/main/java/io/quarkus/narayana/jta/runtime/NarayanaJtaRecorder.java
```diff
@@ -32,34 +31,115 @@ public void setNodeName(final TransactionManagerConfiguration transactions) {
...
+        recoveryEnvironmentBean.setExpiryScannerClassNames(Collections.singletonList(
+                "com.arjuna.ats.internal.arjuna.recovery.ExpiredTransactionStatusManagerScanner"));
+    }
+
+    private void setRecoveryPort(RecoveryEnvironmentBean recoveryEnvironmentBean) {
+        recoveryEnvironmentBean.setRecoveryPort(4712);
+    }
+
+    private void setRecoveryListener(RecoveryEnvironmentBean recoveryEnvironmentBean) {
+        recoveryEnvironmentBean.setRecoveryListener(true);
```
ochaloup: is it good idea to enable the recovery listener by default? That way it opens one more port anytime the JTA is used, or am I wrong?

extensions/narayana-jta/runtime/src/main/java/io/quarkus/narayana/jta/runtime/NarayanaJtaRecorder.java
ochaloup: @mmusgrov is that for sure? Have you considered that the `@BuildStep` you refer to is annotated with `@Record(RUNTIME_INIT)`?

extensions/narayana-jta/runtime/src/main/java/io/quarkus/narayana/jta/runtime/NarayanaJtaRecorder.java
```diff
@@ -32,34 +31,115 @@ public void setNodeName(final TransactionManagerConfiguration transactions) {
...
+                "com.arjuna.ats.internal.jta.recovery.arjunacore.CommitMarkableResourceRecordRecoveryModule"));
+    }
+
+    private void setExpiryScannerClassNames(RecoveryEnvironmentBean recoveryEnvironmentBean) {
+        recoveryEnvironmentBean.setExpiryScannerClassNames(Collections.singletonList(
+                "com.arjuna.ats.internal.arjuna.recovery.ExpiredTransactionStatusManagerScanner"));
+    }
+
+    private void setRecoveryPort(RecoveryEnvironmentBean recoveryEnvironmentBean) {
+        recoveryEnvironmentBean.setRecoveryPort(4712);
```
ochaloup: would not be good to make this configurable?

extensions/narayana-jta/runtime/src/main/java/io/quarkus/narayana/jta/runtime/NarayanaJtaRecorder.java
```diff
@@ -32,34 +31,115 @@ public void setNodeName(final TransactionManagerConfiguration transactions) {
...
+
+            recoveryEnvironmentBean = BeanPopulator.getDefaultInstance(RecoveryEnvironmentBean.class, systemProperties);
+            jtaEnvironmentBean = BeanPopulator.getDefaultInstance(JTAEnvironmentBean.class, systemProperties);
+
+            // set defaults if not set via system properties
+            if (!systemProperties.containsKey("JTAEnvironmentBean.xaResourceOrphanFilterClassNames")) {
+                setXaResourceOrphanFilterClassNames(jtaEnvironmentBean);
+            }
+
+            if (!systemProperties.containsKey("RecoveryEnvironmentBean.recoveryModuleClassNames")) {
```
ochaloup: in Narayana property model it's possible to define the system property in this shortened way of `RecoveryEnvironmentBean.recoveryModuleClassNames` and the other valid property naming is with class name `com.arjuna.ats.arjuna.common.RecoveryEnvironmentBean.recoveryModuleClassNames`
(plus for some of the properties there are defined aliases as e.g. for this one I assume it's valid `com.arjuna.ats.arjuna.recovery.recoveryExtension`)

extensions/narayana-jta/runtime/src/main/java/io/quarkus/narayana/jta/runtime/TransactionManagerConfiguration.java
```diff
@@ -19,8 +16,92 @@
...
+     * {@link TransactionManagerConfiguration#nodeName}
+     * </p>
+     * <p>
+     * <em></em>CoordinatorEnvironmentBean.transactionStatusManagerEnable</em> which is always initialised to the value false
+     * </p>
+     * <p>
+     * Note that quarkus config properties will override Narayana system properties.
+     * </p>
+     */
+    @ConfigItem(name = "checkForNarayanaSystemProperties", defaultValue = "false")
```
ochaloup: isn't this approach a bit in contrast (or in parallel) to how the Quarkus works with properties normally?

ochaloup (changes requested):
I have few comments in my review.

extensions/narayana-jta/runtime/src/main/java/io/quarkus/narayana/jta/runtime/NarayanaJtaRecorder.java
mmusgrov: Okay I removed the security check.

extensions/narayana-jta/runtime/src/main/java/io/quarkus/narayana/jta/runtime/NarayanaJtaRecorder.java
mmusgrov: okay

extensions/narayana-jta/runtime/src/main/java/io/quarkus/narayana/jta/runtime/NarayanaJtaRecorder.java
mmusgrov: okay

extensions/narayana-jta/runtime/src/main/java/io/quarkus/narayana/jta/runtime/NarayanaJtaRecorder.java
mmusgrov: Okay (I set it to true because that was the existing behaviour but using false makes sense).

extensions/narayana-jta/runtime/src/main/java/io/quarkus/narayana/jta/runtime/NarayanaJtaRecorder.java
mmusgrov: It is configurable via a system property.

extensions/narayana-jta/runtime/src/main/java/io/quarkus/narayana/jta/runtime/NarayanaJtaRecorder.java
mmusgrov: `NameEnvironmentBean.propertyName` is the preferred format. The other format is legacy.

The goal of allowing system properties is to provide a mechanism to configure an option that isn't exposed in the quarkus config model. It is not the intention to make everything configurable via the quarkus model (it is similar to WildFly where we still allow a back door to enable access to certain Narayana functionality).

extensions/narayana-jta/runtime/src/main/java/io/quarkus/narayana/jta/runtime/TransactionManagerConfiguration.java
mmusgrov: The goal of allowing system properties is to provide a mechanism to configure an option that isn't exposed in the quarkus config model. It is not the intention to make everything configurable via the quarkus model.

The intention is to gradually expose certain behaviours as we get requests for them. For example, in a subsequent issue I will be adding quarkus properties to configure the journal and JDBC stores. But providing access to the Narayana property mechanism means that these and other features can be enabled now if desired.

extensions/narayana-jta/runtime/src/main/java/io/quarkus/narayana/jta/runtime/TransactionManagerConfiguration.java
```diff
@@ -19,8 +16,92 @@
     public String nodeName;
 
     /**
-     * The default transaction timeout
+     * The default transaction timeout in seconds
```
dmlloyd: The type is `Duration`, so the unit is understood already to be seconds unless it is overridden.  To that end, since you're giving a default of 300 seconds, it's probably more user-friendly to use the string `5m` instead.

extensions/narayana-jta/runtime/src/main/java/io/quarkus/narayana/jta/runtime/TransactionManagerConfiguration.java
dmlloyd: You don't need the `name` or `defaultValue` on this one (the property name would then be `check-for-narayana-system-properties` though, in line with our normal naming convention).

dmlloyd: To address @ochaloup's concern: if we do allow this, we'd want to make a note that this option is specifically for advanced users only.  The problem is that users are going to rely on this capability if we expose it.  Our established practice in situations such as this one has been to let users complain/request a feature which we then explicitly expose via configuration rather than support any sort of pass-through.

extensions/narayana-jta/runtime/src/main/java/io/quarkus/narayana/jta/runtime/TransactionManagerConfiguration.java
```diff
@@ -19,8 +16,92 @@
...
-    @ConfigItem(defaultValue = "60")
+    @ConfigItem(defaultValue = "300")
     public Optional<Duration> defaultTransactionTimeout;
+
+    /**
+     * The directory name of location of the transaction logs.
+     * If the value is not absolute then the directory is relative
+     * to the <em>user.dir</em> system property.
+     */
+    @ConfigItem(name = "objectStoreDir", defaultValue = "ObjectStore")
```
dmlloyd: The `name` option should not be given here; the property would be called `object-store-dir`.

extensions/narayana-jta/runtime/src/main/java/io/quarkus/narayana/jta/runtime/NarayanaJtaRecorder.java
mmusgrov: I am not clear on your issue with this code: when the app starts up the default timeout value is read from the quarkus confg and that value is then set on the bean and on the TxControl. The bytecode for doing this is run from the main method and therefore the default timeout will have been set when the application starts running transactions.

The timeout value in TxControl is not marked as `final`. When TxControl.setDefaultTimeout() is called the class will be initialized and it is initialized with the correct value for default timeout as read from the quarkus config.

extensions/narayana-jta/runtime/src/main/java/io/quarkus/narayana/jta/runtime/TransactionManagerConfiguration.java
mmusgrov: It was my intention to call the property `object-store-dir` since that is the quarkus naming convention. Please can you elaborate.

dmlloyd: Giving the `name` property sets the literal property name to that string, so in configuration, this property would be called `objectStoreDir` (which is the name given in the example configuration in a different part of the PR).

The default value for the `name` property of `ConfigItem` is `ConfigItem#HYPHENATED_ELEMENT_NAME`, which tells the config parser system to de-camel-case and hyphenate the field name and use that as the property name.  By giving an explicit `name` property on this annotation, you've defeated that mechanism and overriden the name which would have been `object-store-dir`, causing it to be `objectStoreDir` instead, which does not follow the naming convention.

extensions/narayana-jta/runtime/src/main/java/io/quarkus/narayana/jta/runtime/TransactionManagerConfiguration.java
mmusgrov: Do you mean 

```java
    @ConfigItem(defaultValue = "5m")
    public Optional<Duration> defaultTransactionTimeout;
```

If so, then that would be liable to be misinterpreted. How would he define seconds or hour units for example.

It is less ambiguous if we pick a single unit (namely seconds) and say that any value the user specifies will always be interpreted as seconds.

extensions/narayana-jta/runtime/src/main/java/io/quarkus/narayana/jta/runtime/TransactionManagerConfiguration.java
mmusgrov: Ah I think I understand. So if no name is given then quarkus will automatically select the name `object-store-dir`. That makes sense.

mmusgrov: and your comment would also apply to the property `checkForNarayanaSystemProperties`. Thanks.

extensions/narayana-jta/runtime/src/main/java/io/quarkus/narayana/jta/runtime/TransactionManagerConfiguration.java
dmlloyd: That's not how we configure time durations in Quarkus.

Seconds are the default unit for all duration properties.  Users can use an abbreviated version of ISO-8601 duration strings to specify other units.  For example, `10` or `10S` or `PT10S` all indicate 10 seconds; `10M` is 10 minutes; `5H10M2S` is 5 hours, 10 minutes, and 2 seconds; `P2DT4H` is 2 days, 4 hours, etc.

It's the same format used for all duration-typed properties in all of Quarkus, so I would not be terribly worried about users getting confused about the format.

extensions/narayana-jta/runtime/src/main/java/io/quarkus/narayana/jta/runtime/TransactionManagerConfiguration.java
ochaloup: Yes, @dmlloyd expressed exactly what I was in doubt about.

Plus, if the goal would be to gradually expose certain behaviours then the properties handling needs to be much more complicated than it's in this PR. What's the part of the PR ((https://github.com/quarkusio/quarkus/blob/95ef7f637da601458273eb8625f67d6baaff7256/extensions/narayana-jta/runtime/src/main/java/io/quarkus/narayana/jta/runtime/NarayanaJtaRecorder.java#L87) you may use either the "quarkus property system" or "narayana system property system". For the gradual evolution, I assume, there should be the rule that a defined quarkus property overrides any used system property.

extensions/narayana-jta/runtime/src/main/java/io/quarkus/narayana/jta/runtime/NarayanaJtaRecorder.java
ochaloup: But only if you decide to use the system properties "configuration model", right?

extensions/narayana-jta/runtime/src/main/java/io/quarkus/narayana/jta/runtime/NarayanaJtaRecorder.java
ochaloup: @mmusgrov ok, I see. Have we stated the preferred format somewhere?

extensions/narayana-jta/runtime/src/main/java/io/quarkus/narayana/jta/runtime/TransactionManagerConfiguration.java
mmusgrov: The PR already states the convention. It is present in the javadoc that I created, ie:

```java
     * Note that quarkus config properties will override Narayana system properties.
```

But I am still of the opinion that we should evolve which set of properties to expose via the quarkus config system. Without this mechanism users will have to wait for us to add the feature.

extensions/narayana-jta/runtime/src/main/java/io/quarkus/narayana/jta/runtime/NarayanaJtaRecorder.java
mmusgrov: It is in our docs

extensions/narayana-jta/runtime/src/main/java/io/quarkus/narayana/jta/runtime/NarayanaJtaRecorder.java
mmusgrov: Exactly. It is a conscious decision to be made by the app developer.

extensions/narayana-jta/runtime/src/main/java/io/quarkus/narayana/jta/runtime/TransactionManagerConfiguration.java
mmusgrov: ie the javadoc for the system property override states how the user can enable Narayana features. But note that the user can already configure Narayana using XML property files (but this PR removes that capability) so not having the new option that this PR introduces is removing functionality.

dmlloyd: > But I am still of the opinion that we should evolve which set of properties to expose via the quarkus config system. Without this mechanism users will have to wait for us to add the feature.

That is exactly the intent.  We've done all other extensions this way as well.

> But [...] not having the new option that this PR introduces is removing functionality.

Right, we are deliberately avoiding exposing every bit of functionality with extensions (not just this one but all of them); rather we put a heavy preference towards only exposing the most commonly used configuration items, at least to start with.  A user should have to configure few or no properties in order to get a working, useful system, and having too many knobs can be an attractive nuisance.  As users need additional control, we can easily add configuration properties to expose what is needed (which have a low implementation complexity, generally speaking).

mmusgrov: > > But [...] not having the new option that this PR introduces is removing functionality.
> al control, we can easily add configuration properties to expose what is needed (which have a low implementation complexity, generally speaking).

My point was that with the existing extension it is possible to provide a jbossts-properties.xml config file. Do you still want me to remove that capability because if I remove that capability and don't allow the use of system properties then I will be introducing a regression?

extensions/narayana-jta/runtime/src/main/java/io/quarkus/narayana/jta/runtime/NarayanaJtaRecorder.java
mmusgrov: @ochaloup Yes we have (both in the javadoc for this PR and in the narayana documentation).

extensions/narayana-jta/runtime/src/main/java/io/quarkus/narayana/jta/runtime/TransactionManagerConfiguration.java
dmlloyd: We can't remove functionality which existed in 1.0.  I would just avoid the system property layer and leave the XML in.  But that's a recommendation, not a hard requirement.

mmusgrov: The main goal of the PR is to remove the need to parse XML properties files. Leaving the XML parsing in the code would imply that this PR is not needed,

dmlloyd: That goal is still valid, in the greater time frame of the project.  But replacing XML with system properties is a purely lateral move, so if the change were just that, it would not be worthwhile.

If support for XML configuration were considered unimportant enough that we would discontinue it without deprecation (this would be a question for the project team, I'm not sure exactly what the criteria are), then I would just drop it with no replacement and let the squeaky wheels squeak.  TBH this is in many ways the best option.

If, however, I am right in my guess that we can't just drop support for it due to compatibility requirements, then chronologically, the plan would necessarily be something like this:

 * Deprecate XML configuration (with a warning)
 * Introduce any missing, desired configuration options as requests are raised
 * Drop the XML configuration when all of the following occur:
    1. The volume of requests has dropped off sufficiently to allay worries of lost functionality
    2. We reach a sufficient compatibility milestone (2.0 perhaps, depends on the outcome of the ongoing deprecation process discussion)

But which course should be taken is not something I can answer directly.

mmusgrov: Okay

extensions/narayana-jta/runtime/src/main/java/io/quarkus/narayana/jta/runtime/NarayanaJtaRecorder.java
mmusgrov: @ochaloup yes we have stated it. It is in the javadoc for this PR and it is in the Narayana documentation.

extensions/narayana-jta/runtime/src/main/java/io/quarkus/narayana/jta/runtime/NarayanaJtaRecorder.java
mmusgrov: @ochaloup ^^ this comment still needs your input.

ochaloup: ah, right. I missed your response, sorry. 

You are right. My first response was on PR as it was in November and my second one was a reaction to statement that all is baked into executable.

As the code is now you are right the the `TxControl.setDefaultTimeout()` is called as the last call in the quarkus `setConfig` method. The only trouble I can see is if the methods are reordered in manner that the `disableTransactionStatusManager()` is called after `setDefaultTimeout(transactions)` in the quarkus `setConfig` method.

From what I can see my comment regarding the `TxControl` is not relevant in the current state.

ochaloup (approved):
On the review now I can't see any particular problem to talk about. The discussion about system property setup is about a point of view.

mmusgrov:
There is too much push back on this PR so I am about to withdraw it.

PR #5285, opened 2019-11-07
Updated Kogito extension description and labels

mswiderski:


mswiderski:
@emmanuelbernard here is the PR we talked about ... though I cannot set any labels myself

extensions/kogito/runtime/src/main/resources/META-INF/quarkus-extension.yaml
```diff
@@ -1,10 +1,14 @@
 ---
 name: "Kogito"
+description: "Add business automation capabilities - processes and rules with Kogito (a toolkit that originates from projects Drools and jBPM)."
```
gastaldi: Usually we put the description in the pom.xml. Can you update https://github.com/quarkusio/quarkus/blob/master/extensions/kogito/runtime/pom.xml#L14 instead?

The keywords are fine

gastaldi (changes requested):

extensions/kogito/runtime/src/main/resources/META-INF/quarkus-extension.yaml
gastaldi: Sorry to not have it documented somewhere, but we're working on this :wink:

emmanuelbernard:
You can define labels as

Keywords:
- foo
- bar

Check out
https://github.com/quarkusio/quarkus/blob/master/extensions/hibernate-orm/runtime/src/main/resources/META-INF/quarkus-extension.yaml#L5

extensions/kogito/runtime/src/main/resources/META-INF/quarkus-extension.yaml
mswiderski: description moved to pom.xml

mswiderski:
@emmanuelbernard that's exactly what I did, added new labels to provide more options to find it.

gastaldi (approved):
Excellent! Thank you!

gwenneg:
The test failure is unrelated, I'm relaunching the build.

PR #5287, opened 2019-11-07
Add docs for fixing read only file system deployments

johnoliver:


machi1990 (approved):
Thanks @johnoliver for the contribution. 
LGTM.

Sanne:
Interesting, thanks for the suggestion! I have a follow-up question: any reason to not do this by default?
 @cescoffier ?

johnoliver:
@Sanne Just noticed there is a slightly more in depth answer in vertxs docs at: https://vertx.io/docs/vertx-core/java/#_configuring_vert_x_cache

May answer your question to some extent

gwenneg:
Thanks for your contribution @johnoliver!

> Interesting, thanks for the suggestion! I have a follow-up question: any reason to not do this by default?
> @cescoffier ?

@cescoffier If it's not done by default, should we create a Quarkus config property in `VertxConfiguration` for this property?

Sanne:
thanks for the pointer @johnoliver . Now I'm wondering if we shouldn't set `vertx.disableFileCPResolving=true` by default, as there shouldn't be any blcking resource loading in Quarkus, assuming the extensions are taking care of that... but I guess this would need some more work first.

cescoffier (approved):

cescoffier:
Hi, @johnoliver nice to see you here!

I'm wondering why this code does not do it automatically:
https://github.com/quarkusio/quarkus/blob/master/extensions/vertx-core/runtime/src/main/java/io/quarkus/vertx/core/runtime/VertxCoreRecorder.java#L141-L147

cescoffier:
Wondering if it's because `java.io.tmpdir` may not be defined.

gsmet:
@cescoffier shouldn't we have a proper Quarkus config property for that instead of a magic system property?

Issue #5288, opened 2019-11-07
Access Token is not available to the OIDC 'web-app' applications


sberyozkin:
**Description**
They only see `IdToken`, but if they need to do the outbound calls with the access tokens then no `AccessTokenCredential` will be available, and attempting to inject it as `JsonWebToken` will throw an exception

**Implementation ideas**
Make the access token returned as part of the code grant exchange available as `AccessTokenCredential` and also attach to it a refresh token if it is also available

Issue #5001, opened 2019-10-29
Support for the pre-emptive AccessToken refreshment


sberyozkin:
**Description**
Both access and id tokens can now be supported for the "web-app" applications as part of the code flow. Code for the refresh tokens has also been prototyped. The next step is to use the refresh tokens to pre-emptively refresh ATs when a given access token is close to being expired to avoid the web-app getting 401 when it makes an outbound call with this AT  

**Implementation ideas**
Add a time related property for the adapter to decide when it needs to refresh the token; 
@pedroigor IMHO we'll need to drop the code supporting the injection of RT into the user code, and probably drop `RefreshToken`, because RT can be totally opaque (and as such not map to `JsonWebToken`) but also because it is really an internal adapter task (if we will let the the user code to start refreshing then we'll need to explain how to do it, where to get the IDP address, etc)

IMHO it is not an urgent task

sberyozkin:
@pedroigor OK, but we don't want the users themselves dealing with it, correct ?

pedroigor:
If we provide the necessary capability to automatically refresh tokens, probably yes. But having access to the RT may be useful too.

sberyozkin:
@pedroigor RT is not a token credential really :-). It only enables the refreshment of the AT. So if we do let users see it then we should explain what and how they can do it :-). But that would move us away from the 'make it simple for the users' OIDC strategy. IMHO it will give us more flexibility if we won't start with exposing it, we will always be able to do if there will be a convincing story for it :-)
Thanks

sberyozkin:
@pedroigor Hey Pedro, I thought a bit more about, you are right,  it is not always possible to reliably detect the AT has to be refreshed, there will always be a possibility for the `web-app` doing an outbound call and the AT expiring somewhere in the transit :-).
However I'm still not sure we would ask the users to deal with refreshing the access tokens as there is a bit of complexity involved in the both the request and response validation (esp if OIDC server chooses to return ID token as well).

IMHO, what we should do here is to ship a JAX-RS client filter which will intercept 401/403, check if the AT has expired, and if yes then refresh it and retry the call. This may be tricky to set up (at the level managing the retry HTTP calls with the user code being not aware of it). Or may be we can have an injectable `OidcClient` (which knows how to connect to OIDC, etc) and the user code can then use it to refresh the AT token when needed. Something like that, but I'd consider it be a follow-up issue rather than part of this issue.
[...]

Issue #5002, opened 2019-10-29
Incorrect extension metadata for Jaeger


kenfinnigan:
https://github.com/quarkusio/quarkus/blob/master/extensions/jaeger/runtime/src/main/resources/META-INF/quarkus-extension.yaml identifies the extension as being for "security", which is incorrect

maxandersen:
https://github.com/quarkusio/quarkus/issues/5002

PR #5290, opened 2019-11-07
Have RefreshToken attached to AccessTokenCredential only

sberyozkin:
This minor clean up is primarily related to #5001 (as agreed, refresh token, if available, should only be attached to the current access token), and also prepares a bit #5288.

It is the right time to do this clean up because neither ATs nor RTs are checked as part of the code flow, with RTs not available any way to the 'service' applications


sberyozkin:
@pedroigor So I was under impression after the discussion at #5001 `RefreshToken` was not injected but it is actually being injected, by a `JsonWebToken` producer, though it is not a credential really. And it is already in the quickstarts...Well, do we want to keep it as it is now ? I can close this PR and do a minor update to AccessToken anyway

sberyozkin:
I'm having some thought conflicts about this `RefreshToken` thing. Like right now we support the injection of it without any support for what they can do, and it is not document yet, so we have a chance to tweak something. Specifically, as discussed at #5002, it is not a standalone credential, and it is confusing to me to see it being injected as such.
What I'd like to do is to update the test to get RT as String from AccessTokenCredential.

By the way, `RefreshToken` and `AccessTokenCredential` are in the different packages which is also confusing. I'll move `AccessTokenCredential` (as well as `IdTokenCredential` out of `runtime` one segment up - should be OK since neither of them is doc-ed or used in the quickstarts)

sberyozkin:
Sorry all, I've messed up with the deleted branches so will close again and reopen later

Issue #4791, opened 2019-10-23
The Software Transactional Memory extension should not start the TransactionStatusManager during native build


Sanne:
(repurposing this bug as I originally misunderstood the problem)

**Describe the bug**
The JTA Narayana extension is setting a JVM flag to disable the TSM; something similar should be done for the Software Transactional Memory extension, as it's also starting the TSM during native builds.

See also #4748

Sanne:
Fixed by #5292

PR #5292, opened 2019-11-07
issue 4791 Do not open ports at build time (disable TSM)

mmusgrov:
https://github.com/quarkusio/quarkus/issues/4791

The fix disables the Narayana TransactionStatusManager during native image build and at runtime since it is not needed (and opens ports resulting in unexpected log messages during native image build).

### DIFF
--- a/extensions/narayana-stm/deployment/src/main/java/io/quarkus/narayana/stm/deployment/NarayanaSTMProcessor.java
+++ b/extensions/narayana-stm/deployment/src/main/java/io/quarkus/narayana/stm/deployment/NarayanaSTMProcessor.java
@@ -1,5 +1,7 @@
 package io.quarkus.narayana.stm.deployment;
 
+import static io.quarkus.deployment.annotations.ExecutionTime.RUNTIME_INIT;
+
 import java.util.ArrayList;
 import java.util.Collection;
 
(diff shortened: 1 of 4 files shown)


extensions/narayana-stm/deployment/src/main/java/io/quarkus/narayana/stm/deployment/NarayanaSTMProcessor.java
```diff
@@ -39,14 +45,30 @@
...
+    public NativeImageSystemPropertyBuildItem substrateSystemPropertyBuildItem() {
+        return new NativeImageSystemPropertyBuildItem("CoordinatorEnvironmentBean.transactionStatusManagerEnable", "false");
+    }
+
+    // the software transactional memory implementation does not require a TSM
+    // so disable it at runtime
+    @BuildStep()
+    @Record(RUNTIME_INIT)
+    public void configureRuntimeProperties(NarayanaSTMRecorder recorder) {
+        recorder.disableTransactionStatusManager();
```
machi1990: Just for my own curiosity: Why do we have this? Isn't 
```
return new NativeImageSystemPropertyBuildItem("CoordinatorEnvironmentBean.transactionStatusManagerEnable", "false");
``` 
done above enough?

Sanne (approved):
Perfect, thanks @mmusgrov !

extensions/narayana-stm/deployment/src/main/java/io/quarkus/narayana/stm/deployment/NarayanaSTMProcessor.java
mmusgrov: @machi1990 I will retest the extension without it ...

mmusgrov: It is required since when we run the non native image (standard java jar) we do not want the TSM to be started.

machi1990: Ah sure. Thanks  for the explanation.

PR #5295, opened 2019-11-07
Update Infinispan to 10.0.0.Final

wburns:


integration-tests/infinispan-client/pom.xml
```diff
@@ -76,6 +80,24 @@
             <type>test-jar</type>
             <scope>test</scope>
         </dependency>
+
+        <!-- We override these to satisfy Infinispan server -->
```
gastaldi: Does this mean that it won't work with 2.9.10.20191020? IMHO the BOM should be updated to Jackson 2.10.0 first

wburns: The problem is that Infinispan Server uses a new version of Jackson that isn't compatible with the version Quarkus is currently using. The Server is only started to test the client, so this doesn't affect any Quarkus code.

gastaldi: But doesn't that means that in these tests the client will use Jackson 2.10.0 too?

Sanne (dismissed):

integration-tests/infinispan-client/pom.xml
wburns: The tests don't use anything that requires Jackson to be loaded afaik.

wburns: At least it doesn't fail 🤷‍♂

Sanne: good question.

Sanne: Actually, is the Infinispan client using Jackson at all? I guess if it's only used in the server, then it's ok.

wburns: The Infinispan Client does not use Jackson at all, it is all hotrod. I am only overriding the Jackson version for the Infinispan Server (which does use Jackson for REST), so that the client can actually talk to something.

I was referring to whether or not the Quarkus web layer used Jackson, maybe it only uses it for non String types? All my return types are String in the test.

Sanne (changes requested):
This failure seems legit, and will need to be fixed:
 - https://dev.azure.com/quarkus-ci/quarkus/_build/results?buildId=11781&view=logs&jobId=f3eb6b4d-852c-547d-9cce-3f04f51ea340&taskId=1dd5db7a-ab23-5086-58ca-148e537fe622&lineStart=17044&lineEnd=17057&colStart=1&colEnd=80


```
[WARNING] 
Dependency convergence error for org.eclipse.microprofile.metrics:microprofile-metrics-api:2.0.2 paths to dependency are:
+-io.quarkus:quarkus-infinispan-embedded:999-SNAPSHOT
  +-org.infinispan:infinispan-core:10.0.0.Final
    +-org.eclipse.microprofile.metrics:microprofile-metrics-api:2.0.2
and
+-io.quarkus:quarkus-infinispan-embedded:999-SNAPSHOT
  +-org.infinispan:infinispan-core:10.0.0.Final
    +-io.smallrye:smallrye-metrics:2.3.0
      +-org.eclipse.microprofile.metrics:microprofile-metrics-api:2.2

[WARNING] Rule 0: org.apache.maven.plugins.enforcer.DependencyConvergence failed with message:
Failed while enforcing releasability. See above detailed error message.
```

wburns:
Updated, should be fixed now :)

Sanne (approved):
Great! Tested it with native images locally as well.  I'll wait for CI, then merge.

wburns:
@Sanne so to backport do I just cherry pick into the 1.0 branch? I haven't been keeping up with how you guys have been doing release dev.

Sanne:
@wburns we target the next release (1.1) and add a label "backport" to have it considered for the 1.0.0.Final gatekeepers. It's just a hint for them though, risk assesment will decide..

wburns:
Okay, so nothing more on my side once this is integrated, thanks.

wburns:
I will say I would love for this to be backported to 1.0 though 😀

gastaldi (approved):

extensions/infinispan-embedded/runtime/pom.xml
```diff
@@ -11,7 +11,7 @@
 
     <artifactId>quarkus-infinispan-embedded</artifactId>
     <name>Quarkus - Infinispan - Embedded - Runtime</name>
-    <description>Run an embedded Infinispan data grid server for distributed caching</description>
```
gsmet: I'm a bit late to the party but isn't this going to be a problem with code.quarkus.io once we make this extension public? @maxandersen am I right?

Sanne: There's no plan to make this public. It's more likely that we'll move it to a separate repository.

maxandersen: whats the problem you for see @gsmet  ? I feel like i'm missing some context ;)

wburns: To be honest I am not sure why this line was removed. I am more than fine adding it back in if it fixes something :)

maxandersen: ooh - now i get it. yes you'll want to make sure you have a description either here or in quarkus-extension.yaml as otherwise it will show up blank on code.quarkus

PR #5302, opened 2019-11-07
Link RT with AT, moving ID and AT token credentials out of 'runtime' subpackage

### LINKED ISSUE #5301: JsonWebToken raw_token claim is not set in the OIDC flows
**Describe the bug**
MP JWT `JsonWebToken` has a `getRawToken()` method which should return an original token; could be useful in scope of OIDC `web-app` applications which need to use AT to do the outbound calls (though both `IdTokenCredential` and `AccessTokenCredential` are also of help)

**Expected behavior**
`JsonWebToken.getRawToken()` should return an original id or access token

sberyozkin:
Fixes #5301.
Related to #5002. 

This PR:
* Removes two RefreshToken classes, and makes RT only reachable via AccessTokenCredential. IMHO this better represents the purpose of RT, and lets us delay providing something directly for the user code to use to refresh the access tokens which should ideally be done only by the adapter (#5002), or via an injected `OidcClient` (yet to be considered)
* Adds the raw token to `JsonWebToken`
* Moves `IdTokenCredential` and `AccessTokenCredential` out of `runtime` sub-package
* Updates the test

Quickstart will be updated next. Only the way Refresh tokens are accessed will be updated which will be fine as there is no any support yet for those RTs anyway. RTs are not referenced in the guide yet, so updating the way they are accessed should be just fine and in time


pedroigor (approved):

sberyozkin:
@pedroigor thanks; I've been thinking of the alternative, keep one `RefreshToken` but rename it to `RefreshTokenCredential` to have the naming consistent, and still avoid injecting it directly but only make visible through `AccessTokenCredential` but I do not feel like we'll make much progress as far as the simplification is concerned :-), so just making it available as String should be enough, probably it would still be of interest to the adapter only for the preemptive refreshment...
Thanks

stuartwdouglas (approved):

gsmet:
I won't backport this one as it is a breaking change.

gsmet:
Apparently, people want it anyway.

@sberyozkin if users are required to change their applications, I need a blurb explaining how.

Issue #5219, opened 2019-11-05
Allow setting a custom Hibernate ConnectionProvider


bvarner:
**Description**
Allow setting `hibernate.connection.provider-class` through quarkus app application properties.

**Implementation ideas**
Implement adding the ability to specify a custom connection provider with a hibernate.orm setting such as: `quarkus.hibernate-orm.connection.provider-class` that would set `hibernate.connection.provider-class` within a jpa persistence context when not using a persistence.xml.

Sanne:
Hi,
sorry I don't think that's a good idea. Historically Hibernate allowed a lot of flexibility in this area, as there were many buggy connection pools so it was a hard choice.. today it's more like the opposite, we can actually better prevent issues by doing better integration - while the connection provider patterns are well understood.

Why would you want a different provider?

bvarner:
I have worked on _many_ hibernate projects over the years where the ability to customize the SQL connection context and state before it was handed off to the ORM provider was necessary.

Specifically, (in my experience) I've used it (and seen it used) to enable data-base level auditing that maps back to the Principal in use at the application level, or to enhance the instrumentation connection usage. I've also seen environments where it was essential to execute stored procedures from draconian DBAs before you'd have rights to execute anything else.

Normally the use-cases I see (specifically with ConnectionProvider -- I'm going to ignore the multitennancy provider options for the sake of this request) never intend to _replace_ the connection provider, but would be better implemented with  with interceptor type semantics with aroundInvoke on the ConnectionProvider. But...

* ConnectionProvider isn't a CDI bean.
* The ConnectionProviderInitiator service upstream from hibernate isn't really conducive to this.
* Applications normally know exactly what kind of ConnectionProvider they need, certainly within the scope of Quarkus this is true (as you point out).

My immediate use-case is for database-trigger level auditing, which depends upon certain statements / calls to be run on the Connection after being retrieved from the pool, or before it's returned to the pool. I did not see any 'easy' or obvious hook or interceptor option for doing this on the datasource itself -- and in the applications I've run into over the years, this was one way they handled it.

Sanne:
That's awesome feedback, thanks!

Agreed then let's make sure all use cases can be covered - and yes let's skip multi-tenancy at this point as that's a complex one which will need a series of dedicated issues.

Agroal supports listeners; I'm not sure if they are flexible enough for all cases you describe, and we're not exposing any way of registering them, but that's certainly a reasonable request.

[...]

PR #5304, opened 2019-11-07
Expose hibernate ConnectionProvider config option. 

bvarner:
#5219 Allows setting a custom ConnectionProvider for Quarkus apps that may want to:
* use a different connection provider
* customize their connection provider to setup / manage SQL connection conversation state within hibernate contexts.


gastaldi:
@bvarner any chance to add some tests? Also could you please rebase your PR to a single commit?

Sanne:
sorry @bvarner , I understand were you're coming from but as I mentioned on #5219 I don't think it's wise to offer this configuration option, not yet at least.

There's more changes coming in how ORM is being booted, and the datasource integration has a key role in that (BTW did you try if this works even today? I guess this could be enough but I'd be surprised, keep in mind that the service initializations are heavily customized, not all ORM configuration properties will be read by the SessionFactory bootstrap code).

Either way, going this way would prevent us to make several more strong optimisations so let's not rush it in please. Let's clarify the exact needs on the issue, I can help making the listeners happen in Agroal.

PR #5313, opened 2019-11-08
Provide a way to set the TCCL in @BuildStep

stuartwdouglas:
This allows for the TCCL to automatically set the TCCL
in a build step. We can't do it for everyone as the metadata
needed to build this CL is created by @BuildStep methods.

Any build step that needs to load application classes should
use this flag to get a transformation safe ClassLoader.

### DIFF
--- a/core/deployment/src/main/java/io/quarkus/deployment/ExtensionLoader.java
+++ b/core/deployment/src/main/java/io/quarkus/deployment/ExtensionLoader.java
@@ -65,6 +65,7 @@ import io.quarkus.deployment.builditem.AdditionalApplicationArchiveMarkerBuildIt
 import io.quarkus.deployment.builditem.BuildTimeConfigurationBuildItem;
 import io.quarkus.deployment.builditem.BuildTimeRunTimeFixedConfigurationBuildItem;
 import io.quarkus.deployment.builditem.CapabilityBuildItem;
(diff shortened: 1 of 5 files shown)


gsmet:
There are test failures that look related.

stuartwdouglas:
The other way to do this would be via an 'opt out' based approach, so all the pre-classloading build steps would not have their TCCL set.

jaikiran:
Hello @stuartwdouglas 

> Any build step that needs to load application classes

Have things changed when it comes to build steps (as part of build processors) loading application classes? I'm still not fully caught up with Quarkus augmentation and other details, but the documentation that I had read previously (and even the live one) states (https://quarkus.io/guides/writing-extensions#bootstrap-three-phases):

> Augmentation
>     This is the first phase, and is done by the Build Step Processors. 
> **These processors** have access to Jandex annotation information and can parse any descriptors 
> and read annotations, but **should not attempt to load any application classes**.

stuartwdouglas:
That has changed, unfortunately it turns out that there are cases where do no need to load application classes. This approach allows us to load them in a temp class loader, so they won't affect transformation, in much the same way that the JPA spec does.

jaikiran:
Understood. Thank you.

dmlloyd (approved):

PR #5315, opened 2019-11-08
Fix OIDC flow issue

stuartwdouglas:
This would attempt to authenticate for unprotected resources

### DIFF
--- a/extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/CodeAuthenticationMechanism.java
+++ b/extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/CodeAuthenticationMechanism.java
@@ -96,10 +96,14 @@ public class CodeAuthenticationMechanism extends AbstractOidcAuthenticationMecha
 
     private CompletionStage<SecurityIdentity> performCodeFlow(IdentityProviderManager identityProviderManager,
             RoutingContext context) {
-        CompletableFuture<SecurityIdentity> cf = new CompletableFuture<>();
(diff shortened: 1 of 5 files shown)


pedroigor (approved):

extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/CodeAuthenticationMechanism.java
```diff
@@ -96,10 +96,14 @@ private static QuarkusSecurityIdentity augmentIdentity(SecurityIdentity security
 
     private CompletionStage<SecurityIdentity> performCodeFlow(IdentityProviderManager identityProviderManager,
             RoutingContext context) {
-        CompletableFuture<SecurityIdentity> cf = new CompletableFuture<>();
         JsonObject params = new JsonObject();
 
-        params.put("code", context.request().getParam("code"));
+        String code = context.request().getParam("code");
+        if (code == null) {
```
sberyozkin: @stuartwdouglas Might make sense to log at this point that the IDP has not completed the user authentication

sberyozkin (approved):
Please add that comment if you agree

extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/CodeAuthenticationMechanism.java
stuartwdouglas: you could add some debug logging I guess, but it should not be logged at any level higher.

PR #5319, opened 2019-11-08
Fixes #5286 - DynamoDB: split config into build & runtime configs.

### LINKED ISSUE #5286: DynamoDB processor accesses run time config
**Describe the bug**
The `io.quarkus.dynamodb.deployment.DynamodbProcessor` class accesses `io.quarkus.dynamodb.runtime.DynamodbConfig` from non-recorder steps.  The run time configuration is not normally available during build and this functionality will fail in upcoming releases.  Run time configuration must be checked at run time only.  Build time configuration must be defined as a build time configuration object.

**Fix options**
1. Change `io.quarkus.dynamodb.runtime.DynamodbConfig` to have a phase of `ConfigPhase#BUILD_AND_RUN_TIME_FIXED`.
2. Skip the check in `io.quarkus.dynamodb.deployment.DynamodbProcessor#setup` and possibly do the check at run time instead.
3. After the config update is merged, go back and use the various built-in validation constructs (optional and required properties and config groups) to validate instead of using an explicit check method

/cc @marcinczeczko

marcinczeczko:
Fix for #5286

- Extracted two properties into the build level (Interceptors & sync client type)
- Refactored validation on the build time for build time props only
- Moved runtime config validations on the processor level.

/cc @gsmet @dmlloyd

### DIFF
--- a/extensions/amazon-dynamodb/deployment/src/main/java/io/quarkus/dynamodb/deployment/DynamodbProcessor.java
+++ b/extensions/amazon-dynamodb/deployment/src/main/java/io/quarkus/dynamodb/deployment/DynamodbProcessor.java
@@ -1,6 +1,5 @@
 package io.quarkus.dynamodb.deployment;
 
-import java.net.URI;
 import java.util.Arrays;
 import java.util.List;
 import java.util.stream.Collectors;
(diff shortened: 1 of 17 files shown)


dmlloyd:
Useful tip: if you use the exact syntax "Fixes #5286" in your PR or commit message, the corresponding issue will automatically close when your PR is merged.

marcinczeczko:
@dmlloyd - thanks. Actually I knew that, but for some reason I typed something else :)

dmlloyd (approved):

Issue #4088, opened 2019-09-18
CodeSource#location always null inside when running in dev mode


jordanglassman:
**Describe the bug**

When running a Flyway java-based migration programatically, it failed with NPE when running:

```
            ProtectionDomain protectionDomain = aClass.getProtectionDomain();
            CodeSource codeSource = protectionDomain.getCodeSource();
            String url = codeSource.getLocation().getPath();
```

because all fields in CodeSource are null:

![image](https://user-images.githubusercontent.com/1238718/65154997-776a0900-d9fa-11e9-8e45-92552c93aee2.png)

I discovered that this is the case everywhere in the application.  Not sure if this is expected or not, but it is preventing me from using Flyway programmatically.  

Unknown if the problem recurs when using Flyway declaratively per [the guide](https://quarkus.io/guides/flyway-guide).

**Expected behavior**
CodeSource#location to be populated with the URL to the source file.

**Actual behavior**
CodeSource#location is null.

**To Reproduce**
1.  Add a breakpoint anywhere inside a running quarkus app.
2. Run `this.getClass().getProtectionDomain().getCodeSource().getLocation()`

**Configuration**
N/A

**Environment (please complete the following information):**
 - Output of `uname -a` or `ver`: `Darwin ip-192-168-1-137.ec2.internal 18.2.0 Darwin Kernel Version 18.2.0: Mon Nov 12 20:24:46 PST 2018; root:xnu-4903.231.4~2/RELEASE_X86_64 x86_64`
 - Output of `java -version`: 
```
java version "1.8.0_221"
Java(TM) SE Runtime Environment (build 1.8.0_221-b11)
Java HotSpot(TM) 64-Bit GraalVM EE 19.1.1 (build 25.221-b11-jvmci-19.1-b01, mixed mode)
```
 - Quarkus version or git rev: 0.22.0

jordanglassman:
Update:  verified same error recurs with programmatic Flyway configuration:

```
2019-09-18 10:23:23,992 ERROR [io.qua.dev.DevModeMain] (main) Failed to start quarkus: java.lang.RuntimeException: Failed to start quarkus
	at io.quarkus.runner.ApplicationImpl1.doStart(ApplicationImpl1.zig:142)
	at io.quarkus.runtime.Application.start(Application.java:91)
	at io.quarkus.runner.RuntimeRunner.run(RuntimeRunner.java:127)
[...]

PR #5325, opened 2019-11-08
4088- CodeSource#location always null inside when running in dev mode

anamarija:
Hi this is a fix for the [4088](https://github.com/quarkusio/quarkus/issues/4088) issue regarding  Flyway java-based migration in quarkus:dev mode.

The issue was when Flyway tries to retrieve class Location from the codeSource of the class it gets a NullPointerException. Here is the code that triggers the NPE
```java
public static String getLocationOnDisk(Class<?> aClass) {
        try {
            ProtectionDomain protectionDomain = aClass.getProtectionDomain();
            if (protectionDomain == null) {
                //Android
                return null;
            }
            CodeSource codeSource = protectionDomain.getCodeSource();
            if (codeSource == null) {
                //Custom classloader with for example classes defined using URLClassLoader#defineClass(String name, byte[] b, int off, int len)
                return null;
            }
            String url = codeSource.getLocation().getPath();
            return URLDecoder.decode(url, "UTF-8");
        } catch (UnsupportedEncodingException e) {
            //Can never happen.
            return null;
        }
```
In the dev mode application classes are loaded as byte arrays with the defaultProtectionDomain object which is defined like 

```java
private final ProtectionDomain defaultDomain =
        new ProtectionDomain(new CodeSource(null, (Certificate[]) null),
                             null, this, null);
```

So the source of the error is that FlyWay class utils null checks codeSource but not the codeSource.location.
I tried to compare dumped class values from the RuntimeClassLoader(dev mode) and URLClassLoader( regular mode-running a jar).
The dumps look the same to me. I don’t see any information missing which could trigger such a behaviour.

So my proposed solution is creation of a protectionDomain which has codeSource.location which is appliation package location or codeSource = null

### DIFF
--- a/core/deployment/src/main/java/io/quarkus/runner/RuntimeClassLoader.java
+++ b/core/deployment/src/main/java/io/quarkus/runner/RuntimeClassLoader.java
@@ -9,12 +9,17 @@ import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.net.MalformedURLException;
+import java.net.URI;
+import java.net.URISyntaxException;
 import java.net.URL;
 import java.net.URLConnection;
 import java.net.URLStreamHandler;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.security.CodeSource;
 import java.security.MessageDigest;
+import java.security.ProtectionDomain;
+import java.security.cert.Certificate;
 import java.util.ArrayList;
 import java.util.Base64;
 import java.util.Collections;
@@ -54,6 +59,8 @@ public class RuntimeClassLoader extends ClassLoader implements ClassOutput, Tran
 
     private final Map<String, Path> applicationClasses;
 
+    private final ProtectionDomain defaultProtectionDomain;
+
     private final Path frameworkClassesPath;
     private final Path transformerCache;
 
@@ -85,6 +92,7 @@ public class RuntimeClassLoader extends ClassLoader implements ClassOutput, Tran
                 }
             }
 
+            this.defaultProtectionDomain = createDefaultProtectionDomain(applicationClassesDirectories.get(0));
             this.applicationClasses = applicationClasses;
 
         } catch (IOException e) {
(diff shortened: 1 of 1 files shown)


core/deployment/src/main/java/io/quarkus/runner/RuntimeClassLoader.java
```diff
@@ -486,4 +492,17 @@ public InputStream getInputStream() throws IOException {
         }
         return null;
     }
+
+    private ProtectionDomain createProtectionDomainWithClasspath(String name) {
+        URL url = null;
+        try {
+            url = new URL("file:///" + getPackageNameFromClassName(name).replace('.', '/'));
```
stuartwdouglas: Will is cause problems to have these non-existant url's as the location? I would have thought using the first location from applicationClassDirectories would make more sense.

dmlloyd: Frameworks sometimes (right or wrong) use the code source to introspect the class path, so it should definitely be a real path whenever possible.

There should only be one protection domain (and thus one code source, and thus one URL) instance for each artifact.  If all of the artifacts are squashed into one place, then there should just be one protection domain instance for the whole class loader.

Also I would recommend using `java.net.URI` with a component constructor, calling `toURL()` on the result, to avoid encoding issues and that sort of thing.

stuartwdouglas (approved):

core/deployment/src/main/java/io/quarkus/runner/RuntimeClassLoader.java
```diff
@@ -486,4 +494,19 @@ public InputStream getInputStream() throws IOException {
         }
         return null;
     }
+
+    private ProtectionDomain createDefaultProtectionDomain(Path applicationClasspath) {
+        URL url = null;
+        if (applicationClasspath != null) {
+            try {
+                URI uri = new URI("file://" + applicationClasspath);
```
dmlloyd: Because of possible encoding issues, the `URI(scheme,host,path,fragment)` constructor should be used instead.  Just give `null` for `host` and `fragment`.

dmlloyd (changes requested):
See comments for change.

dmlloyd (approved):

dmlloyd (approved):

stuartwdouglas (approved):

PR #5328, opened 2019-11-08
Properly match paths ending with /*

### LINKED ISSUE #5318: Unexpected behavior of path matching for configuration-based security policies
If I have a permission for a path like
`quarkus.http.auth.permission.mypermission.paths=/a/*`
Then requests to `/a/anything` should be matched to it, but they are not.

If I change it to 
`quarkus.http.auth.permission.mypermission.paths=/a*`
Only then requests to `/a/anything` will start getting authorization checks.

This is contradictory to what is described in the documentation https://quarkus.io/guides/security#matching-on-paths-methods

jmartisk:
Fixes #5318

### DIFF
--- a/extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/security/PathMatchingHttpSecurityPolicy.java
+++ b/extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/security/PathMatchingHttpSecurityPolicy.java
@@ -93,7 +93,10 @@ public class PathMatchingHttpSecurityPolicy implements HttpSecurityPolicy {
                     List<HttpMatcher> perms = new ArrayList<>();
                     tempMap.put(path, perms);
                     perms.add(m);
-                    if (path.endsWith("*")) {
+                    if (path.endsWith("/*")) {
+                        String stripped = path.substring(0, path.length() - 2);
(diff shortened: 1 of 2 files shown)


extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/security/PathMatchingHttpSecurityPolicy.java
```diff
@@ -93,7 +93,10 @@ void init(HttpBuildTimeConfig config, Map<String, Supplier<HttpSecurityPolicy>>
                     List<HttpMatcher> perms = new ArrayList<>();
                     tempMap.put(path, perms);
                     perms.add(m);
-                    if (path.endsWith("*")) {
+                    if (path.endsWith("/*")) {
```
sberyozkin: The original check should be retained as well IMHO, so that the last URL segment paths can be wildcard matched, those not ending with "/"

jmartisk: It is retained, it was just moved in to the `else if` branch (line 99)

sberyozkin (approved):

geoand:
Is this supposed to be backported into `1.0.x`?

sberyozkin:
@geoand Yes, let me merge it, thanks

geoand:
Excellent, thanks!

PR #5331, opened 2019-11-08
Restore one of RT classes, produce all token credentials

### LINKED ISSUE #5303: Support the injection of AccessTokenCredential and IdTokenCredential
**Description**
Let users do
```
@Inject
AccessTokenCredential at;
``` 
instead of
```
@Inject
SecurityIdentity identity;
...
AccessTokenCredential at = identity.getTokenCredential(AccessTokenCredential.class);
``` 

and the same for `IdTokenCredential`.

A bit simpler when nothing else is required from `SecurityIdentity`.
(Based on Pedro's idea to inject refresh tokens as token creds)

**Implementation ideas**
Add a simple producer

sberyozkin:
Also
Fixes #5303

### DIFF
--- a/extensions/oidc/runtime/src/main/java/io/quarkus/oidc/RefreshToken.java
+++ b/extensions/oidc/runtime/src/main/java/io/quarkus/oidc/RefreshToken.java
@@ -1,0 +1,14 @@
+package io.quarkus.oidc;
+
+import io.quarkus.security.credential.TokenCredential;
+
+public class RefreshToken extends TokenCredential {
+
+    public RefreshToken() {
+        this(null);
+    }
+
+    public RefreshToken(String refreshToken) {
+        super(refreshToken, "refresh_token");
+    }
+}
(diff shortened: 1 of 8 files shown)


pedroigor (approved):

sberyozkin:
Please get #5315 in first, I'll rebase

extensions/oidc/runtime/src/main/java/io/quarkus/oidc/RefreshToken.java
```diff
@@ -0,0 +1,14 @@
+package io.quarkus.oidc;
```
rsvoboda: Previously the package was `io.quarkus.oidc.runtime` 

google groups thread mentions change in package - *though in a new package*
What's the reason for package change ?

sberyozkin: @rsvoboda sure, it is mean tot be part of a public API, so it needs to be located outside of the `runtime`. 
As I mentioned in the google thread, in fact `RefreshToken` was duplicated both in `io.quarkus.oidc` and `io.quarkus.runtime` before I removed both of them with an earlier PR, so this PR actually restores the one in `io.quarkus.oidc` to be in the same package as `AccessTokenCredential`/etc.
Does it answer your question ? I know, it looks not good when we see a package name change. But it is the right fix and is just in time.

rsvoboda: thanks for details

sberyozkin:
Hi All, going ahead with the merge as it blocks the quickstarts, given Pedro's approval

gastaldi (approved):

PR #5352, opened 2019-11-09
remove unused portProducer parameter and use runtime httpConfiguration in runtime recording

machi1990:
/cc @geoand minor cleanups in VertxHttpProcessor

### DIFF
--- a/extensions/vertx-http/deployment/src/main/java/io/quarkus/vertx/http/deployment/VertxHttpProcessor.java
+++ b/extensions/vertx-http/deployment/src/main/java/io/quarkus/vertx/http/deployment/VertxHttpProcessor.java
@@ -38,8 +38,6 @@ import io.vertx.ext.web.Router;
 
 class VertxHttpProcessor {
 
-    HttpConfiguration httpConfiguration;
-
     @BuildStep
     HttpRootPathBuildItem httpRoot(HttpBuildTimeConfig config) {
         return new HttpRootPathBuildItem(config.rootPath);
@@ -71,8 +69,7 @@ class VertxHttpProcessor {
 
     @BuildStep(onlyIf = IsNormal.class)
     @Record(value = ExecutionTime.RUNTIME_INIT, optional = true)
-    public KubernetesPortBuildItem kubernetes(HttpConfiguration config, BuildProducer<KubernetesPortBuildItem> portProducer,
-            VertxHttpRecorder recorder) {
+    public KubernetesPortBuildItem kubernetes(HttpConfiguration config, VertxHttpRecorder recorder) {
         int port = ConfigProvider.getConfig().getOptionalValue("quarkus.http.port", Integer.class).orElse(8080);
         recorder.warnIfPortChanged(config, port);
         return new KubernetesPortBuildItem(config.port, "http");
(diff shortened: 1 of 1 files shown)


extensions/vertx-http/deployment/src/main/java/io/quarkus/vertx/http/deployment/VertxHttpProcessor.java
```diff
@@ -126,6 +124,10 @@ ServiceStartBuildItem finalizeRouter(
                 listOfFilters, vertx.getVertx(), router.getRouter(), httpBuildTimeConfig.rootPath, launchMode.getLaunchMode());
 
         boolean startVirtual = requireVirtual.isPresent() || httpConfiguration.virtual;
+        if (requireVirtual.isPresent() || httpConfiguration.virtual) {
```
geoand: Can you revert this change and the relevant removal please?

machi1990: The `HttpConfiguration` class is a runtime configuration, we are trying to access a value from a normal build step (build time), that's what motivated to move that part here (inside a RUNTIME recorder).  Is this okay?

geoand: The fact that the build step is also a recorder doesn't affect the access of build / runtime time configuration.
However your latest addition does make things a little more clear so let's go with it. I'll just add one comment about the name.

extensions/vertx-http/deployment/src/main/java/io/quarkus/vertx/http/deployment/VertxHttpProcessor.java
```diff
@@ -99,10 +96,11 @@ VertxWebRouterBuildItem initializeRouter(VertxHttpRecorder recorder,
...
             Optional<RequireVirtualHttpBuildItem> requireVirtual,
-            InternalWebVertxBuildItem vertx,
-            LaunchModeBuildItem launchMode, ShutdownContextBuildItem shutdown, List<DefaultRouteBuildItem> defaultRoutes,
-            List<FilterBuildItem> filters, VertxWebRouterBuildItem router, EventLoopCountBuildItem eventLoopCountBuildItem,
-            HttpBuildTimeConfig httpBuildTimeConfig,
+            InternalWebVertxBuildItem vertx, LaunchModeBuildItem launchMode, ShutdownContextBuildItem shutdown,
+            List<DefaultRouteBuildItem> defaultRoutes, List<FilterBuildItem> filters,
+            VertxWebRouterBuildItem router, EventLoopCountBuildItem eventLoopCountBuildItem,
+            HttpBuildTimeConfig httpBuildTimeConfig, HttpConfiguration httpConfiguration,
+            BuildProducer<ReflectiveClassBuildItem> reflectiveClassBuildItemBuildProducer,
```
geoand: Let's be consistent with the names of the parameters. None of the other parameters contains `BuildItem` in them, nor is `Producer` used for producers in other methods of the class.

Sorry for these nitpicks, but I strongly think we should try and be as consistent as possible, especially when it doesn't come at the cost of anything else.

geoand (approved):
Thanks!

machi1990:
Thanks @geoand

geoand:
Thank you @machi1990

PR #5365, opened 2019-11-11
Generated Gradle code should use the build/ directory

### LINKED ISSUE #5363: Cannot do docker build with gradle based on native image
Hi folks.

I would like to know how we can do a docker build with gradle based on native build. 

I read this doc : https://quarkus.io/guides/gradle-tooling

I build a native image with this command : `./gradlew buildNative --docker-build=true`, it works.

But when i see the DockerFile.native, i would like to build my docker image with this command : 
`docker build -f src/main/docker/Dockerfile.native -t quarkus/openapi-swaggerui-quickstart .`

But i have an error into the DockerFile, at this line : `COPY target/*-runner /work/application`, with this error : 

> Step 3/6 : COPY target/*-runner /work/application
> COPY failed: no source files were specified

The cause is when i do this with maven package it works because a target folder is created but in gradle it's not. I tried to switch the **target** folder to **build** folder (where the runner is present) in the DockerFile, but i have the same error :

> Step 3/6 : COPY build/*-runner /work/application
> COPY failed: no source files were specified

Someone can help me about this problem ?

Ps : My project is based on this tutorial : https://quarkus.io/guides/openapi-swaggerui

Thanks a lot for your help, feel free to ask.

gastaldi:
Fixes #5363

### DIFF
--- a/devtools/platform-descriptor-json/src/main/resources/templates/dockerfile-jvm.ftl
+++ b/devtools/platform-descriptor-json/src/main/resources/templates/dockerfile-jvm.ftl
@@ -17,8 +17,8 @@
 FROM fabric8/java-alpine-openjdk8-jre
 ENV JAVA_OPTIONS="-Dquarkus.http.host=0.0.0.0 -Djava.util.logging.manager=org.jboss.logmanager.LogManager"
 ENV AB_ENABLED=jmx_exporter
-COPY target/lib/* /deployments/lib/
-COPY target/*-runner.jar /deployments/app.jar
+COPY ${build_dir}/lib/* /deployments/lib/
+COPY ${build_dir}/*-runner.jar /deployments/app.jar
 EXPOSE 8080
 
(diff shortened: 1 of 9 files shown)


geoand (approved):
Excellent!

aloubyansky (approved):

oztimpower:
And if I use Kotlin Grade DSL?  build.gradle.kts

gastaldi:
@oztimpower you mean the build directory is different in this case?

oztimpower:
Build directory is the same, I just noticed a hard coded reference to the gradle build files.
independent-projects/tools/common/src/main/java/io/quarkus/generators/BuildTool.java

In addition, what if I use both maven and gradle? Preference is gradle, yet I find myself having to validate what I'm doing with the maven plugin also, so I typically hard code the docker ignore to allow both, given the lack of gradle support.

gastaldi:
The maven plugin only allows you to select one single build tool, so I guess that's fine when creating new projects.

PR #5367, opened 2019-11-11
Create and use a view role binding when the kubernetes-client extension is used.

iocanel:
To use the `kubernetes-client` from within kubernetes, some RBAC configuration is required.
At minimum the `view` role needs to be granted to the pod.

This pull request introduces a new build item for extensions to be able to express their RBAC requirements. KubernetesProcessor has been updated to read those items and generate the manifests accordingly.


geoand:
Excellent, really good usability improvement!

Can you also have a mention of this in the kubernetes-client guide?

geoand (approved):

PR #5372, opened 2019-11-11
Remove unused gradle build parameters

gastaldi:
These parameters are never used in the task (probably were copied from somewhere else)


devtools/gradle/src/main/java/io/quarkus/gradle/tasks/QuarkusBuild.java
```diff
@@ -24,64 +23,16 @@
  */
 public class QuarkusBuild extends QuarkusTask {
 
-    private String transformedClassesDirectory;
-
-    private String wiringClassesDirectory;
-
-    private String libDir;
```
geoand: Are we sure this one isn't used?

The rest do seem redundant

gastaldi: Yes, apparently they were copied from the `QuarkusDev` task and are never used in the `buildQuarkus()` method

gastaldi: Ah ignore that comment, they are leftovers from the original code

gastaldi: They were used here https://github.com/quarkusio/quarkus/commit/2fa569b35d248a11e27fea892a46d76e6d8fe2b7#diff-cfb0ca87755414fb0943c90f63f1419eR144-R156 but not anymore since @aloubyansky 's refactoring

geoand: OK, I'm good with this :+1:

aloubyansky:
Looks good. What about the corresponding methods in QuarkusPluginExtension? They seem to be useless too. Could you please check @gastaldi ?

aloubyansky (approved):

PR #5385, opened 2019-11-11
Use Class.forName to prevent IllegalAccessError

### LINKED ISSUE #5326: IllegalAccessError trying to run UberJar when public class C extends public class B extends package class A
My class inheritance looks like this:

class Apackage.A
public class Apackage.B extends A
public class Bpackage.C extends B

This has been working up to now with Wildfly/Thorntail, etc.
When trying to run an UberJar quarkus built, I'm getting:

java.lang.IllegalAccessError: tried to access class A from class C_Bean

I couldn't find any good documentation on transitive class access though, so maybe this is not a bug as such. Also have no idea how C_Bean comes into life (presumably already including all of B's funtionality, therefore stripping access to B and directly accessing A?)

I cannot easily change Apackage.A to have public access as that comes from a library.

stuartwdouglas:
Fixes #5326 

Thanks @jaikiran for your excellent analysis and test case

### DIFF
--- a/independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/Types.java
+++ b/independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/Types.java
@@ -1,6 +1,7 @@
 package io.quarkus.arc.processor;
 
 import static io.quarkus.arc.processor.IndexClassLookupUtils.getClassByName;
+import static io.quarkus.gizmo.MethodDescriptor.ofMethod;
 
 import io.quarkus.arc.impl.GenericArrayTypeImpl;
 import io.quarkus.arc.impl.ParameterizedTypeImpl;
@@ -49,7 +50,8 @@ final class Types {
 
     static ResultHandle getTypeHandle(BytecodeCreator creator, Type type) {
         if (Kind.CLASS.equals(type.kind())) {
-            return creator.loadClass(type.asClassType().name().toString());
+            String className = type.asClassType().name().toString();
+            return doLoadClass(creator, className);
         } else if (Kind.TYPE_VARIABLE.equals(type.kind())) {
             // E.g. T -> new TypeVariableImpl("T")
             TypeVariable typeVariable = type.asTypeVariable();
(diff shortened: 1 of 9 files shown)


independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/Types.java
```diff
@@ -49,7 +50,8 @@ private Types() {
 
     static ResultHandle getTypeHandle(BytecodeCreator creator, Type type) {
         if (Kind.CLASS.equals(type.kind())) {
-            return creator.loadClass(type.asClassType().name().toString());
+            String className = type.asClassType().name().toString();
+            return doLoadClass(creator, className);
```
mkouba: Hm, we use `BytecodeCreator.loadClass()` in many places. Why do we actually get an IAE when trying to use `Hello.class` in the `MorningGreeting` bean? I mean it works with javac so why is it a problem for gizmo?

stuartwdouglas: The interface is not actually accessible from the bean, so LDC does not work. It's a bit of a corner case but it's just how the JVM works

mkouba: Ok, I see. I've verified this again and it's indeed not possible ;-). So the only thing I'm not sure about is that we do add the `TCCL` and `Class.forName()` for every bean type. Which means that the constructor will grow for a bean with large number of bean types.

Also I wonder if should verify all those places where we use `BytecodeCreator.loadClass()`...

stuartwdouglas: The recorders already use the TCCL because this problem popped up a while ago. In general though this code should not be slower (even thought it is a few more bytecodes), and the number of types should never be large enough that the method size limit becomes a problem.

mkouba: > should never be large enough that the method size limit becomes a problem.

We do a lot more stuff in the constructor but you're right that we should not hit the limit for common use cases.

I'm going to approve this PR but create a follow-up issue to verify all other occurences of `BytecodeCreator.loadClass()` in arc.

mkouba (approved):

PR #5396, opened 2019-11-12
Interceptor resolution - fix handling of default values of bindings

### LINKED ISSUE #5373: @InterceptorBinding no longer recognized/fired when they have additional fields
This interceptor binding is working fine, if there is an interceptor (having @Priority annotation or not, @AroundInvoke gets fired wherever @StorageThreadSession is annotated):

```
@Target({TYPE, METHOD})
@Retention(RUNTIME)
@InterceptorBinding
public @interface StorageThreadSession {
}
```

adding a field to the binding definition, @AroundInvoke stops working, it seems the interceptor binding is not recognized at all in this case:

```
@Target({TYPE, METHOD})
@Retention(RUNTIME)
@InterceptorBinding
public @interface StorageThreadSession {
	
	Class qualifier() default Object.class;
}
```

I can split up my annotation but I don't think that this is the desired behaviour.

mkouba:
- resolves #5373


manovotn (approved):

gwenneg (approved):
LGTM, the Windows build error seems unrelated.

mkouba:
It seems the CI is down :-(.

Issue #5381, opened 2019-11-11
Allow addition of AgroalDataSourceListener implementations during run-time production of DataSources.


bvarner:
**Description**
Allow CDI @ApplicationScoped AgroalDataSourceListeners to be registered, prioritized, and applied to appropriate data sources during production of an AgroalDataSource.

**Implementation ideas**
Per discussion on #5219, [a CDI implementation](https://github.com/quarkusio/quarkus/issues/5219#issuecomment-551938067) may be an elegant way to handle this.

Sanne:
hi @barreiro could you have a look here please from an Agroal perspective? Also see the linked previous discussion on #5219 to get some context

@mkouba and we'll need you for the CDI aspects :)

bvarner:
Sorry for the confusing stuff on here. I've squashed my commits and rebased with a new PR.

Sanne:
Related PR #9009 was merged

PR #5405, opened 2019-11-12
Feature/cdi agroal datasource listeners

bvarner:
To help foster discussion on #5381 here's a quick first-pass at implementing the CDI run-time bits.

There's some things here I'm not proud of or thrilled with, like cribbing the Interceptor.Priority constants.

From an Agroal perspective, I also haven't checked the Agroal listener invocation order on things like `onConnectionAcquire()` and `onConnectionReturn()`.

In my (private closed source) use-case, this is working fantastically.

If you could all steer me in the right directions here with regards to documentation and testing, I'll be happy to contribute additional changes.


bvarner:
Gah. This is what I get for trying to squash commits after opening the PR. :-(

PR #5408, opened 2019-11-12
CDI detection of AgroalDataSourceListeners

bvarner:
First-pass Implementation of https://github.com/quarkusio/quarkus/issues/5381


extensions/agroal/runtime/src/main/java/io/quarkus/agroal/runtime/AbstractDataSourceProducer.java
```diff
@@ -246,4 +267,40 @@ public void stop() {
...
+        }
+
+        PrioritizedDataSourceListenerWrapper(final Bean<AgroalDataSourceListener> listener) {
+            this.listener = (AgroalDataSourceListener) beanManager.getReference(listener, listener.getBeanClass(),
+                    beanManager.createCreationalContext(listener));
+
+            Priority p = listener.getBeanClass().getAnnotation(Priority.class);
+            this.priority = p != null ? p.value() : Interceptor.Priority.APPLICATION;
+
+            io.quarkus.agroal.DataSource ds = listener.getBeanClass().getAnnotation(io.quarkus.agroal.DataSource.class);
```
mkouba: `io.quarkus.agroal.DataSource` is a qualifier so there's no need to use reflection to obtain the name. `Bean.getQualifiers()` could be used to obtain the annotation instance.

extensions/agroal/runtime/src/main/java/io/quarkus/agroal/runtime/AbstractDataSourceProducer.java
```diff
@@ -209,9 +218,21 @@ public boolean isValid(Connection connection) {
             }
         }
 
+        // Identify AgroalDataSourceListener beans and filter appropriately to this dataSourceName.
+        List<PrioritizedDataSourceListenerWrapper> listeners = beanManager.getBeans(AgroalDataSourceListener.class).stream()
```
mkouba: The problem with this approach is that beans implementing `AgroalDataSourceListener` would be normally removed since they're considered "unused" (see https://quarkus.io/guides/cdi-reference#remove_unused_beans). When using `javax.enterprise.inject.Instance`, on the other hand, the bean class is not available (DataSource name is - see my comment bellow).

Maybe we should introduce a quarkus-specific interface that would extend `AgroalDataSourceListener` and `javax.enterprise.inject.spi.Prioritized`. That would make it possible to simply `@Inject Instance<PrioritizedDataSourceListener> instance` then `select(DataSourceLiteral.of("dataSourceName"))` (DataSourceLiteral would be something similar to `javax.enterprise.inject.literal.NamedLiteral`) and then simply `instance.stream().sorted()...`.

mkouba (changes requested):
I like the idea but I'd try to avoid using `BeanManager` and `Bean` APIs. 

Maybe we should also try to move the logic to the build phase instead of runtime?

And of course, the PR is missing tests ;-).

extensions/agroal/runtime/src/main/java/io/quarkus/agroal/runtime/AbstractDataSourceProducer.java
bvarner: When I tried that, all I was getting was DEFAULT and Any

extensions/agroal/runtime/src/main/java/io/quarkus/agroal/runtime/AbstractDataSourceProducer.java
bvarner: In my project using this, I have an intermediate library that has the listener, so I guess that's why it's not getting removed. Good to know.

I like the suggestion of the custom interface. I was trying to make this as small a change as possible, but I'd rather make it as correct a change as possible. :-)

I think the combined quarkus-specific thing is a good idea - it also provides a place to define some priority constants specific to this use.

Regarding testing, I didn't see any tests in the quarkus-agroal projects to base testing off of.
If you have a good crash-course on how to test quarkus stuff -- I'd love to see it. This has been nagging at me.

extensions/agroal/runtime/src/main/java/io/quarkus/agroal/runtime/AbstractDataSourceProducer.java
bvarner: > I like the idea but I'd try to avoid using `BeanManager` and `Bean` APIs.

Yeah. I wasn't sure how to go about that without introducing a bunch of a new stuff into Quarkus, and I didn't want to go that route without having some discussion with the folks in charge. :-)

mkouba: You will get `Default` and `Any` for the default DS. For a DS defined like [this](https://github.com/quarkusio/quarkus/blob/master/extensions/agroal/deployment/src/test/resources/application-multiple-datasources.properties#L7-L11), you'll get `Any`, `DataSource("users")` and `Named("users")` (not sure why, maybe @Sanne knows?).

bvarner: Except this isn't getting the DataSource -- this is getting the @DataSource qualifier off the @ApplicationScoped listener bean.

Example of a listener that gets picked up by this: (wow this really was a bad PR, huh?)
```
@DataSource("someDataSourceName")
@ApplicationScoped
@Priority(10)
public class AuditingDataSourceListener implements AgroalDataSourceListener {
...
}
```

In this case, the qualifier is used to determine what datasource to apply the listener to (rather than all of them).

extensions/agroal/runtime/src/main/java/io/quarkus/agroal/runtime/AbstractDataSourceProducer.java
mkouba: So we should have a test in the deployment module: https://github.com/quarkusio/quarkus/tree/master/extensions/agroal/deployment/src/test and ideally also a test for native image, somewhere here: https://github.com/quarkusio/quarkus/tree/master/integration-tests.

extensions/agroal/runtime/src/main/java/io/quarkus/agroal/runtime/AbstractDataSourceProducer.java
```diff
@@ -246,4 +267,40 @@ public void stop() {
...
+            this.dataSourceName = ds != null ? ds.value() : null;
+        }
+
+        boolean appliesTo(final String dataSourceNamed) {
+            return dataSourceName == null || dataSourceName.equals(dataSourceNamed);
+        }
+
+        @Override
+        public int compareTo(final PrioritizedDataSourceListenerWrapper o) {
+            return Integer.compare(this.priority, o.priority);
```
aguibert: we should also add a secondary comparison to ensure consistent ordering for the cases where two listeners are applied with the same priority. The secondary comparison could be considered an implementation detail and therefore doesn't need to be documented. Perhaps something like comparing on `listener.getClass().getCanonicalName()`?

extensions/agroal/runtime/src/main/java/io/quarkus/agroal/runtime/AbstractDataSourceProducer.java
bvarner: @mkouba - To implement the existing AgroalEventLoggingListener there'll need to be a build-time injection point transformer, correct?

bvarner: Also, will ARC automagically sort based on priority if things extend Prioritized?

Sanne:
hi @barreiro , could you have a look? I believe this stalled in waiting for you.

we'll need this feature I believe as it's essential to set e.g. roles in a secured environment.

extensions/agroal/runtime/src/main/java/io/quarkus/agroal/runtime/AbstractDataSourceProducer.java
gsmet: @mkouba could you answer the question above ^.

mkouba: Sorry, I lost the context and I will probably need a simplified summary...

In any case, ArC does not sort the beans. However, it's possible to inject `javax.enterprise.inject.Instance` and then use `instance.stream().sorted()`.

barreiro:
@Sanne I have some concerns with this approach. 

- The AgroalDataSourceListener interface was not designed with the Interceptor concept in mind. Agroal always invoke the listeners in the given order, so if there are 2 listeners with common concerns,only one of them will be effective. 
- The Listener gives access to the raw java.sql.Connection and therefore any error may cause chaos on the pool (leave an invalid connection on the pool, etc.) 
- Some methods on the listener are invoked by Agroal's own thread and therefore should not execute any long running task (that is not a problem in the context of this issue, but may be if we expose this interface as API)
- When running Agroal with a TransactionManager, the life-cycle of the connections can be slightly different as this is associated with the transaction. This does not map well on the current listener interface.

All in all, I think having a new interface in Agroal that is a callback just for acquire / return, is consistent with the current transaction and can handle errors would be a far superior approach. I'll try to code something around these lines.

Sanne:
good points @barreiro . But yes we'll need an alternative API then, this seems a rather important use case that needs to be covered.

I guess the good news is that I don't expect significant changes to be needed in this PR; eager to see your API proposal.

gsmet:
@barreiro ping?

nderwin:
Any updates on this?  I'd really like to upgrade to the latest Quarkus version for some projects I'm working on, but can't due to this issue.

barreiro:
@nderwin the target is to have this feature in Quarkus 1.4, as it requires support in Agroal. There have been some discussions around what goes where, and there is an elegant design that will be implemented in time for the release.

gsmet:
@barreiro is the 1.8 upgrade related to this issue? Could we make progress on this?

Sanne:
@gsmet yes it's related. We needed the changes from Agroal 1.8 to be able to now finally progress with this one. AFAIK @barreiro is looking at adapting this to the new API proposals now.

gsmet:
I'm closing this one given https://github.com/quarkusio/quarkus/pull/9009 has been merged.

PR #5410, opened 2019-11-12
Fix failing tests in master

kenfinnigan:
I know the fix isn't ideal, as it adds sleep, but any alternative suggestions are appreciated

### DIFF
--- a/extensions/resteasy/deployment/src/test/java/io/quarkus/resteasy/test/NotFoundExceptionMapperTestCase.java
+++ b/extensions/resteasy/deployment/src/test/java/io/quarkus/resteasy/test/NotFoundExceptionMapperTestCase.java
@@ -54,14 +54,18 @@ public class NotFoundExceptionMapperTestCase {
     }
 
     @Test
-    public void shouldDisplayNewAddedFileIn404ErrorPage() {
+    public void shouldDisplayNewAddedFileIn404ErrorPage() throws InterruptedException {
         String CONTENT = "html content";
         test.addResourceFile(META_INF_RESOURCES + "index2.html", CONTENT);
+
+        Thread.sleep(700);
 
         RestAssured.get("/index2.html")
                 .then()
                 .statusCode(200)
                 .body(containsString(CONTENT)); // check that index2.html is live reloaded
+
+        Thread.sleep(500);
 
         RestAssured.given()
                 .accept(ContentType.HTML)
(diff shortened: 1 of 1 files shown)


geoand (approved):
Let's get this in sooner rather than later

extensions/resteasy/deployment/src/test/java/io/quarkus/resteasy/test/NotFoundExceptionMapperTestCase.java
```diff
@@ -54,15 +54,19 @@ public void testJsonResourceNotFound() {
     }
 
     @Test
-    public void shouldDisplayNewAddedFileIn404ErrorPage() {
+    public void shouldDisplayNewAddedFileIn404ErrorPage() throws InterruptedException {
         String CONTENT = "html content";
         test.addResourceFile(META_INF_RESOURCES + "index2.html", CONTENT);
 
+        Thread.sleep(700);
```
cescoffier: wouldn't it be better to use Awaitibility and poll until it passes?

kenfinnigan: Await on the `RestAssured` call passing?

And then have a max time of 2 seconds?

Is that preferred? I can adjust

kenfinnigan: We currently don't have Awaitility as a dependency, is it ok to add?

cescoffier: I have extensions using it I believe.

kenfinnigan: ugh, I was looking at the wrong BOM!

kenfinnigan: Tried switching to awaitility, and it has a failure with one, not entirely sure why

kenfinnigan: Awaitility seems to lead to just as much raciness unfortunately

PR #5411, opened 2019-11-12
configure/enable stale bot

maxandersen:
enables stale bot.

marks any issue or pr that has no activity for 4 weeks with label 'stale' and put a comment on it.
then after 7 days close it if no activity since then and if you want to keep an issue "forever" mark it as 'pinned'.

We could also make it so any issues with a  milestone or project link or specific labels (i.e. epics) will stay around - but haven't done that yet; we can adjust as we go.

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,57 @@
+# Configuration for probot-stale - https://github.com/probot/stale
+
+# Number of days of inactivity before an Issue or Pull Request becomes stale
+daysUntilStale: 28
+
+# Number of days of inactivity before an Issue or Pull Request with the stale label is closed.
+# Set to false to disable. If disabled, issues still need to be closed manually, but will remain marked as stale.
+daysUntilClose: 7
+
+# Only issues or pull requests with all of these labels are check if stale. Defaults to `[]` (disabled)
+onlyLabels: []
+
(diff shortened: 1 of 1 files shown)


.github/stale.yml
```diff
@@ -0,0 +1,57 @@
...
+# Number of days of inactivity before an Issue or Pull Request with the stale label is closed.
+# Set to false to disable. If disabled, issues still need to be closed manually, but will remain marked as stale.
+daysUntilClose: 7
+
+# Only issues or pull requests with all of these labels are check if stale. Defaults to `[]` (disabled)
+onlyLabels: []
+
+# Issues or Pull Requests with these labels will never be considered stale. Set to `[]` to disable
+exemptLabels:
+  - pinned
```
cescoffier: do we need to create the label first?

maxandersen: already done ;)

maxandersen:
also, not merging this before we send a mail to the mailing list.
once that is done the stalebot should activate within ~24 hours and do about 30 issues updates per hour which will mean we'll get ~400 issues updated the next 10-12 hours after first run.

maxandersen:
mail sent to mailing list.

cescoffier (approved):

PR #5414, opened 2019-11-12
Fix the application-type property in the oidc web authentication guide

sberyozkin:


docs/src/main/asciidoc/security-openid-connect-web-authentication.adoc
```diff
@@ -65,11 +65,11 @@ The OpenID Connect extension allows you to define the configuration using the `a
 ----
 quarkus.oidc.auth-server-url=http://localhost:8180/auth/realms/quarkus
 quarkus.oidc.client-id=frontend
-quarkus.oidc.client-type=web-app
+quarkus.oidc.application-type=web-app
```
cescoffier: Was it a typo in the doc or is it related to some recent code changes?

cescoffier:
I've added a comment, just to be sure it's not a breaking change that needs to be tracked.

docs/src/main/asciidoc/security-openid-connect-web-authentication.adoc
sberyozkin: It was something we discussed with Pedro, `client-type` was in the PR only but `application-type` was what made it to the master

sberyozkin:
@cescoffier sure, it probably made it to the docs while we were still discussing PR, so it is OK

pedroigor (approved):

rsvoboda (approved):

geoand:
So are we good to merge this?

PR #5415, opened 2019-11-12
Take Spring Data Pageable's Sort into account

### LINKED ISSUE #5406: Spring data Jpa extension : sort with Pageable does not work
When I use the extension of spring data Jpa, the sort with Pageable does not work.

This is my exp:  

```
@GET
    @Produces("application/json")
    public List<Fruit> findAll() {
        return fruitRepository.findAll(PageRequest.of(0,3, Sort.by(Sort.Direction.DESC, "id"))).getContent();
    }
```

the method sort does not work as if it does not exist

geoand:
Fixes #5406


cescoffier:
LGTM but would prefer having the opinion of @FroMage

FroMage (approved):

PR #5421, opened 2019-11-12
Persistent attributes of a Panache entity should not be generated when using Jackson

the-bug:
When using JSONB the the method *isPersistent()* in PanacheEntityBase is ignored from JSONB and does not produce the *persistant*-attribute. This was fixed in https://github.com/FroMage/quarkus/commit/86edaa10eba0da2c1052566d752a257068da2b0f
But if Jackson is used, the *persistant*-attribute is still generated. 

This PR tackles this issue.

This is my first PR in an open source project and I do not know if i mess things up....

### DIFF
--- a/extensions/panache/hibernate-orm-panache/runtime/pom.xml
+++ b/extensions/panache/hibernate-orm-panache/runtime/pom.xml
@@ -34,6 +34,11 @@
             <artifactId>jakarta.json.bind-api</artifactId>
             <optional>true</optional>
         </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-jackson</artifactId>
+            <optional>true</optional>
+        </dependency>
     </dependencies>
 
     <build>
--- a/extensions/panache/hibernate-orm-panache/runtime/src/main/java/io/quarkus/hibernate/orm/panache/PanacheEntityBase.java
+++ b/extensions/panache/hibernate-orm-panache/runtime/src/main/java/io/quarkus/hibernate/orm/panache/PanacheEntityBase.java
@@ -7,6 +7,8 @@ import java.util.stream.Stream;
 import javax.json.bind.annotation.JsonbTransient;
 import javax.persistence.Transient;
 
+import com.fasterxml.jackson.annotation.JsonIgnore;
+
 import io.quarkus.hibernate.orm.panache.runtime.JpaOperations;
 import io.quarkus.panache.common.Parameters;
 import io.quarkus.panache.common.Sort;
@@ -74,6 +76,8 @@ public abstract class PanacheEntityBase {
      * @return true if this entity is persistent in the database.
      */
     @JsonbTransient
+    // @JsonIgnore is here to avoid serialization of this property with jackson
+    @JsonIgnore
     public boolean isPersistent() {
         return JpaOperations.isPersistent(this);
     }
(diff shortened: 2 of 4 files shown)


cescoffier:
@FroMage can you have a look?

integration-tests/hibernate-orm-panache/jackson-integration/src/test/java/io/quarkus/it/panache/TestEndpointTest.java
```diff
@@ -0,0 +1,26 @@
...
+public class TestEndpointTest {
+
+    @Test
+    public void testHelloEndpoint_noPersistantAttributeIsThere() {
+        given()
+                .when().get("/test") //
+                .then() //
+                .statusCode(200) //
+                // here is checked that the property 'persistent' is missing
+                .body(is("{\"id\":null,\"name\":\"max\"}"));
```
FroMage: This doesn't seem right: the body should be `{"name": "max"}`.

FroMage (changes requested):
Hi. Thanks for the PR. The fix looks good, but I don't think we should add a new integration test just for this. Can't we add a unit test to the existing integration test using Jackson explicitely instead?

integration-tests/hibernate-orm-panache/jackson-integration/src/test/java/io/quarkus/it/panache/TestEndpointTest.java
the-bug: > Hi. Thanks for the PR. The fix looks good, but I don't think we should add a new integration test just for this. Can't we add a unit test to the existing integration test using Jackson explicitely instead?

Hi,
thanks for the review!
I have build a new integration test module for this because I do not know how to configure Quarkus to use JSONB and Jackson together... I was following this guide: https://quarkus.io/guides/rest-json 
How can I tell Quarkus to use JSONB or Jackson for a resource or method explicitly?

> This doesn't seem right: the body should be `{"name": "max"}`.

Yes, this is not right.
The behaviour with jsonb is different. There it is `{"name": "max"}` with no `"id":null`.
In PanacheEntity is mentioned that the accessors for id are generated but I do not know how...
I do not know to fix this. Do you have a hint?

FroMage: Don't use REST, just write a unit test that serialises using the Jackson API?

Apologies, my bad, I misread and I think it's right that if the `id` is null it can be serialised as such. I initially thought it was `persistent` instead of `id`. Just ignore this :)

the-bug: > Don't use REST, just write a unit test that serialises using the Jackson API?

Thank you for that hint. Did it.

FroMage (changes requested):
OK, thanks, this is much better. Can you please remove the formatting changes of the `pom.xml` files? This  doesn't let us view the changes, and I don't think they're necessary.

the-bug:
Sorry, I did not see I messed with the formatting...

FroMage (approved):
Perfect, thanks!

FroMage:
One CI test failed, looks unrelated, rerunning it.

the-bug:
Can the PR be merged?

FroMage:
Absolutely, thanks!

PR #5426, opened 2019-11-12
Add caffeine integration tests

lburgazzoli:


stuartwdouglas (changes requested):
If you want this to run in native mode on CI it needs to be added to ci-templates/stages.yaml

cescoffier:
As mentioned by @stuartwdouglas you need to add it to the stage list and increase the timeout accordingly. I let you guess the additional time required.

lburgazzoli:
which category fits better ? I would say `infinispan` as `caffeine` is also used by those extensions

stuartwdouglas:
sounds good to me

lburgazzoli:
done, I haven't increased the time-out as now it fails pretty fast :)

lburgazzoli:
I think this can safely be closed as superseded by the cache extension tests

PR #5430, opened 2019-11-13
Provide integration between RestAssured and Jackson

geoand:
When a Quarkus application used Jackson (and therefore has configured
an ObjectMapper CDI bean), the ObjectMapper is configured to work with
RestAssured.
This makes it easy for users to reuse their Jackson configuration in tests
leaving their tests very clean and avoiding duplicate configuration.

Originally mentioned here: https://stackoverflow.com/q/58794871/2504224

### DIFF
--- a/test-framework/common/src/main/java/io/quarkus/test/restassured/QuarkusJackson2ObjectMapperFactory.java
+++ b/test-framework/common/src/main/java/io/quarkus/test/restassured/QuarkusJackson2ObjectMapperFactory.java
@@ -1,0 +1,28 @@
+package io.quarkus.test.restassured;
+
+import java.lang.reflect.Type;
+
+import javax.enterprise.inject.spi.CDI;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+import io.restassured.path.json.mapper.factory.DefaultJackson2ObjectMapperFactory;
+
+/**
+ * A RestAssured Jackson2ObjectMapperFactory that pulls the ObjectMapper from CDI
+ * and delegates to the parent if no CDI bean is found
+ *
+ * It is important that this class not be referenced directly anywhere in the code so the dependencies
+ * on RestAssured and Jackson can remain optional
+ */
+public class QuarkusJackson2ObjectMapperFactory extends DefaultJackson2ObjectMapperFactory {
+
+    @Override
+    public ObjectMapper create(Type type, String s) {
+        ObjectMapper result = CDI.current().select(ObjectMapper.class).get();
+        if (result == null) {
+            super.create(type, s);
+        }
+        return result;
+    }
+}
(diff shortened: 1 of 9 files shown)


geoand:
Oops, forgot to add the proper test - incoming

geoand:
Test added

test-framework/common/src/main/java/io/quarkus/test/restassured/QuarkusJackson2ObjectMapperFactory.java
```diff
@@ -0,0 +1,28 @@
...
+ * and delegates to the parent if no CDI bean is found
+ *
+ * It is important that this class not be referenced directly anywhere in the code so the dependencies
+ * on RestAssured and Jackson can remain optional
+ */
+public class QuarkusJackson2ObjectMapperFactory extends DefaultJackson2ObjectMapperFactory {
+
+    @Override
+    public ObjectMapper create(Type type, String s) {
+        ObjectMapper result = CDI.current().select(ObjectMapper.class).get();
```
machi1990: Nice feature. Just curios if this will also work in native IT? https://quarkus.io/guides/getting-started-testing#native-executable-testing

geoand: What makes you think it won't?

geoand:
There is a weird test failure, I'm looking into it

test-framework/common/src/main/java/io/quarkus/test/restassured/QuarkusJackson2ObjectMapperFactory.java
machi1990: I am trying to fully understand the limit of CDI in NativeImageTest and specifically this note `...This supports all the features mentioned in this guide except injecting into tests` from the link above about NativeImageTest.

geoand: Very good catch. I didn't even think of that TBH

cescoffier: Yes, you cannot reuse the same ObjectMapper in native, you will need your own (with the same set of modules).

geoand:
I'm going to close this since it's not going to work in native tests... It was a neat idea I think, but oh well :)

test-framework/common/src/main/java/io/quarkus/test/restassured/QuarkusJackson2ObjectMapperFactory.java
geoand: Indeed it won't work. Thanks for pointing it out!

geoand: Technically we probably could start Arc inside the native test as well and use it to pull out beans, but it's probably more of a hassle than it's worth...

PR #5308, opened 2019-11-07
hibernate validator: Remember annotated interface methods

ebullient:
resolves #1888

This includes #1913

PR #5444, opened 2019-11-13
Revert "hibernate validator: Remember annotated interface methods"

cescoffier:
Reverts quarkusio/quarkus#5308 because it breaks the build on various PR.

@gsmet is going to have a deeper look.


geoand:
I don't think this is needed, my reasoning is described here: https://quarkusio.zulipchat.com/#narrow/stream/187038-dev/topic/Hibernate.20validator.20IT.20issue.3F/near/180647719

PR #5445, opened 2019-11-13
Fix a race condition on Kafka Streams hot reload

loicmathieu:
This PR fixes a race condition on the kafka-stream hot replacement setup.

Some kind of double-locked idiom was used without a volatile field.
After discussion with @cescoffier about this, I moe the implementation to use an `AtomicLong` instead.

Tested OK manually using the integration test (no test added).

I take the liberty to also add a small fix to the mongodb-panache guide to avoid creating a separate PR for this. I can remove this comit and create a separate PR if needed.

### DIFF
--- a/extensions/kafka-streams/deployment/src/main/java/io/quarkus/kafka/streams/deployment/KafkaStreamsHotReplacementSetup.java
+++ b/extensions/kafka-streams/deployment/src/main/java/io/quarkus/kafka/streams/deployment/KafkaStreamsHotReplacementSetup.java
@@ -12,7 +12,7 @@ public class KafkaStreamsHotReplacementSetup implements HotReplacementSetup {
     private static final long TWO_SECONDS = 2000;
 
     private HotReplacementContext context;
-    private long nextUpdate;
+    private volatile long nextUpdate;
     private final Executor executor = Executors.newSingleThreadExecutor();
 
     @Override
--- a/docs/src/main/asciidoc/mongodb-panache.adoc
+++ b/docs/src/main/asciidoc/mongodb-panache.adoc
@@ -448,7 +448,7 @@ In MongoDB with Panache the ID are defined by a field named `id` of the `org.bso
 but if you want ot customize them, once again we have you covered.
 
 You can specify your own ID strategy by extending `PanacheMongoEntityBase` instead of `PanacheMongoEntity`. Then
-you just declare whatever ID you want as a public field by annotating it by @BsonId:
+you just declare whatever ID you want as a public field by annotating it by `@BsonId`:
 
 [source,java]
 ----


gsmet:
@gunnarmorling can you have a look?

extensions/kafka-streams/deployment/src/main/java/io/quarkus/kafka/streams/deployment/KafkaStreamsHotReplacementSetup.java
```diff
@@ -25,22 +26,19 @@ public void setupHotDeployment(HotReplacementContext context) {
...
-                            } catch (RuntimeException e) {
-                                throw e;
-                            } catch (Exception e) {
-                                throw new RuntimeException(e);
-                            }
-                        });
-                        // we update at most once every 2s
-                        nextUpdate = System.currentTimeMillis() + TWO_SECONDS;
+            final long now = System.currentTimeMillis();
+            // The getAndUpdate we will update the nextUpdate field if it's value is more that now then will return the current value.
```
gunnarmorling: it's=its. The entire sentence sounds a bit odd, something seems wrong with wording/grammar.

extensions/kafka-streams/deployment/src/main/java/io/quarkus/kafka/streams/deployment/KafkaStreamsHotReplacementSetup.java
```diff
@@ -12,7 +13,7 @@
     private static final long TWO_SECONDS = 2000;
 
     private HotReplacementContext context;
-    private long nextUpdate;
+    private AtomicLong nextUpdate = new AtomicLong(System.currentTimeMillis() + TWO_SECONDS);
```
cescoffier: cosmetic - make is final.

extensions/kafka-streams/deployment/src/main/java/io/quarkus/kafka/streams/deployment/KafkaStreamsHotReplacementSetup.java
cescoffier: I believe it should be something like:

```
Check if we need to trigger the scan. The scan is triggered if the nextUpdate is greater than the current time (now).
nextUpdate is also updated if required to now + 2 seconds
```

Or something like this.

cescoffier (dismissed):
I've made a proposal for the comment.

extensions/kafka-streams/deployment/src/main/java/io/quarkus/kafka/streams/deployment/KafkaStreamsHotReplacementSetup.java
loicmathieu: This was just a comment targeting the next developer that will look at this code and wondering what it does :)

Maybe I need more to explain why it is like that then what it does, the trick here is that the same condition will enter the loop and update the AtomicLong atomically while protecting multiple thread doing it concurrently ...

So maybe the best comment will look like this
```
Check if we need to trigger the scan. 
The AtomicLong will protect the check from threads doing it concurrently.
The scan is triggered if the nextUpdate is greater than the current time (now). 
As the same condition is used inside the LongUnaryOperator that update the nexUpdate, when the condition is true, it will also be updated to now + 2 seconds.
```

gunnarmorling (commented):
Btw. the entire code was copied from the Undertow live reload. So it might deserve fixing over there, too.

extensions/kafka-streams/deployment/src/main/java/io/quarkus/kafka/streams/deployment/KafkaStreamsHotReplacementSetup.java
```diff
@@ -25,22 +26,19 @@ public void setupHotDeployment(HotReplacementContext context) {
...
-                            } catch (Exception e) {
-                                throw new RuntimeException(e);
-                            }
-                        });
-                        // we update at most once every 2s
-                        nextUpdate = System.currentTimeMillis() + TWO_SECONDS;
+            final long now = System.currentTimeMillis();
+            // The getAndUpdate we will update the nextUpdate field if it's value is more that now then will return the current value.
+            // If the value is more than now, a new task will be send to the executor.
+            if (nextUpdate.getAndUpdate((value) -> value < now ? value : now + TWO_SECONDS) > now) {
```
gunnarmorling: Probably just me, but I find this entire line hard to grasp. Not sure how to improve it, though.

gsmet: It's not just you :).

loicmathieu: I propose a new version yesterday https://github.com/quarkusio/quarkus/pull/5445#discussion_r346210543

loicmathieu:
@gunnarmorling @cescoffier @gsmet can we decide on the comment change and proceed to this PR ?

- My ne proposal: https://github.com/quarkusio/quarkus/pull/5445#discussion_r346210543
- Clement's proposal: https://github.com/quarkusio/quarkus/pull/5445#discussion_r346151033
- Removing the comment

I don't care of the comment as I know what I did but it's for a future reader of this complex piece of code :)

gsmet:
OK, so, executive decision on this one: please let's get back to the existing code and just add a `volatile`.

Thanks!

loicmathieu:
@gsmet OK, done.

gsmet (approved):

PR #5450, opened 2019-11-13
Use same ResumingRouter in CDI as in Observes events.

famarting:
In vertx-http the router passed in `@Observes Router` events is a ResumingRouter, this is because it's needed to resume the http request after the root handler first pauses it... However the router we were using in CDI was the directly vertx router so user code defining routes in a router got from injection would get missbehaviours receiving the request body...
As per the docs we don't show the usage of the router getting it from injection but because we use it this way in some of our tests I think we should fix that and pass the ResumingRouter instance to CDI.

### DIFF
--- a/extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/VertxHttpRecorder.java
+++ b/extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/VertxHttpRecorder.java
@@ -181,8 +181,9 @@ public class VertxHttpRecorder {
 
         filterList.addAll(filters.getFilters());
 
-        // Then, fire the router
-        event.select(Router.class).fire(new ResumingRouter(router));
+        // Then, fire the resuming router
+        ResumingRouter resumingRouter = new ResumingRouter(router);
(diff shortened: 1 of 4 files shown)


extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/VertxHttpRecorder.java
```diff
@@ -181,8 +181,9 @@ public void finalizeRouter(BeanContainer container, Consumer<Route> defaultRoute
 
         filterList.addAll(filters.getFilters());
 
-        // Then, fire the router
-        event.select(Router.class).fire(new ResumingRouter(router));
+        // Then, fire the routernew ResumingRouter(router)
```
machi1990:
```suggestion
        // Then, fire the resuming router
```

machi1990 (commented):
@famartinrh thanks for the PR. 

I guess there is a typo in the comment. Can you check?

stuartwdouglas (approved):

stuartwdouglas:
@famartinrh can you squash those two commits?

cescoffier (approved):

PR #5452, opened 2019-11-13
WIP: Reenable stalebot

maxandersen:
Do NOT merge until discussion on mailing list have concluded.

This reverts commit d54198ddde329ac62e6f0e63d03291e928a4e1a0.

see . https://groups.google.com/forum/#!topic/quarkus-dev/LTSI2JoP4WE for discussion.

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,69 @@
+# Configuration for probot-stale - https://github.com/probot/stale
+
+# Number of days of inactivity before an Issue or Pull Request becomes stale
+daysUntilStale: 60
+
+# Number of days of inactivity before an Issue or Pull Request with the stale label is closed.
+# Set to false to disable. If disabled, issues still need to be closed manually, but will remain marked as stale.
+daysUntilClose: 7
+
+# Only issues or pull requests with all of these labels are check if stale. Defaults to `[]` (disabled)
+onlyLabels: []
+
+# Issues or Pull Requests with these labels will never be considered stale. Set to `[]` to disable
+exemptLabels:
+  - pinned
+
+# Set to true to ignore issues in a project (defaults to false)
+exemptProjects: true
+
+# Set to true to ignore issues in a milestone (defaults to false)
+exemptMilestones: true
+
+# Set to true to ignore issues with an assignee (defaults to false)
+exemptAssignees: true
+
+# Label to use when marking as stale
+staleLabel: stale
+
+# Comment to post when marking as stale. Set to `false` to disable
+markComment: >
+  This issue has been automatically marked as stale because it has not had
+  recent activity. It will be closed if no further activity occurs. Thank you!
+
+  We are doing this automatically to ensure out-of-date issues does not stay around
+  indefinitely. 
+
+  If you believe this issue is still relevant please put a comment on it on why
+  and if it truly needs to stay request or add 'pinned' label.
+
+
+
+# Comment to post when removing the stale label.
+# unmarkComment: >
+#   Your comment here.
+
+# Comment to post when closing a stale Issue or Pull Request.
+# closeComment: >
+
+# Limit the number of actions per hour, from 1-30. Default is 30
+limitPerRun: 30
+
+# Limit to only `issues` or `pulls`
+# only: issues
+
+# Optionally, specify configuration settings that are specific to just 'issues' or 'pulls':
+pulls:
+   daysUntilStale: 30
+   markComment: >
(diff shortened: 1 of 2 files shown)


gastaldi:
You may also want to add this file to the list of ignored files in azure-pipelines.yaml

maxandersen:
> You may also want to add this file to the list of ignored files in azure-pipelines.yaml

done.

.github/stale.yml
```diff
@@ -0,0 +1,69 @@
...
+
+# Issues or Pull Requests with these labels will never be considered stale. Set to `[]` to disable
+exemptLabels:
+  - pinned
+
+# Set to true to ignore issues in a project (defaults to false)
+exemptProjects: false
+
+# Set to true to ignore issues in a milestone (defaults to false)
+exemptMilestones: false
```
gastaldi: I think this should be `true`: Issues and PRs already set to a milestone should not never be marked as stale

maxandersen: fixed.

.github/stale.yml
```diff
@@ -0,0 +1,69 @@
...
+  - pinned
+
+# Set to true to ignore issues in a project (defaults to false)
+exemptProjects: true
+
+# Set to true to ignore issues in a milestone (defaults to false)
+exemptMilestones: true
+
+# Set to true to ignore issues with an assignee (defaults to false)
+exemptAssignees: false
```
dmlloyd: This should be `true` too IMO.

.github/stale.yml
```diff
@@ -0,0 +1,69 @@
...
+
+# Label to use when marking as stale
+staleLabel: stale
+
+# Comment to post when marking as stale. Set to `false` to disable
+markComment: >
+  This issue has been automatically marked as stale because it has not had
+  recent activity. It will be closed if no further activity occurs. Thank you!
```
dmlloyd: IMO we still need to update the bot to define "activity" to include references from PRs and other issues.  That will probably require a code change.

.github/stale.yml
```diff
@@ -0,0 +1,69 @@
+# Configuration for probot-stale - https://github.com/probot/stale
+
+# Number of days of inactivity before an Issue or Pull Request becomes stale
+daysUntilStale: 60
+
+# Number of days of inactivity before an Issue or Pull Request with the stale label is closed.
+# Set to false to disable. If disabled, issues still need to be closed manually, but will remain marked as stale.
+daysUntilClose: 7
```
dmlloyd: This would be very harsh if someone is in the middle of things but goes on PTO for example.

dmlloyd (commented):
Still not feeling great here.  Some comments added.

.github/stale.yml
maxandersen: What do you suggest as alternative ? (7 days is the one I've seen used the most and the default)

.github/stale.yml
maxandersen: I agree it would be nice to have; but can you think of when an issue that is being worked on from a PR is not supposed to be assigned a milestone or be in a project? would it be too much to ask those triaging issues to set that info ? then this issue aren't going to happen and if it does you comment and get 60 more days or mark it as pinned and you get infinity ;)

.github/stale.yml
maxandersen: ok, toggling.

.github/stale.yml
dmlloyd: 14 maybe?

.github/stale.yml
dmlloyd: I try to always put my issues into projects, but projects are not universally used and most people don't have rights to create them (of course).  And I don't think milestone is really right unless someone really wants to go and constantly update those (I know I'm not going to do it).  Most people also don't have rights to assign an issue or to add labels or milestones.

This seems to leave frequently pinging the issue as the only alternative unless a maintainer decides to pin the issue for the contributor.  That's not really ideal.

This is really looking like an XY problem to me.  We have this solution (stalebot) but I don't think there's really a broad agreement as to what the problem actually is, and who we're really trying to help.  XY problems can result in two possible outcomes: infinite discussion, or an inadequate/flawed solution.

Sanne:
Could you add these labels to be considered "pinned" as well:

- "good first issue"
- "help wanted" 

Thanks!

gastaldi:
`extension-proposal` also would be a label issue to be pinned afaik

maxandersen:
closing as stalebot was deeemed to noisy/intrusive in its current form.

PR #5387, opened 2019-11-11
Introduce new configuration framework and update SmallRye Config version

dmlloyd:
- Fixes a variety of configuration related issues.
- Changes the strategy for configuration reading so that all groups are read eagerly always (in the case of config roots) or as soon as a property within the group is detected (in the case of map values or leaves).  This allows environment variables to override most configuration properties without requiring the property to be present in `application.properties`.
- Use the SmallRye Config converter infrastructure to dramatically reduce the number of special cases related to configuration property analysis.  Only nested maps generally need significantly specialized behavior now.
- Explicitly defines more behaviors regarding empty value handling, default values for primitives, optional versus required, arrays, and collections.
- Tighten rules about accessing run time configuration values from build steps.
- Explicitly make run time configuration available to dev mode on deploy/start failure.
- Make doc generation support arrays.
- Use new config factory SPI to avoid issues around multiple `ConfigProviderResolver` instances.
- Support optional `@ConfigGroup`s.

Before graduating from draft:
- [x] SmallRye Config ~~1.4.0~~ 1.4.1 release (@kenfinnigan) and PR update (@dmlloyd)
- [x] Tests for `Map<String, Map<String, Xxx>>` where `Xxx` is a leaf or config group (@machi1990)
- [x] Tests to verify environment var overlay works as advertised (@machi1990)
- [x] Verify that config documentation website generator works (@machi1990)
- [x] Documentation: describe default values for primitives (@dmlloyd)
- [x] Documentation: update extension author's guide describe eager config loading strategy and optional config groups (@dmlloyd)
- [x] Documentation: describe (for users) empty value handling, how to explicitly clear a property (@dmlloyd)
- [ ] Documentation: list (for users) all the default configuration sources and their priority (@dmlloyd)
- [x] Documentation: update extension author's guide to clarify rules around run time configuration object handling (@dmlloyd)
- [x] Documentation: update extension author's guide to describe empty value handling in extensions (@dmlloyd)
- [x] Verify Quickstarts (@gastaldi)
- [x] Verify Docker build (@dmlloyd)
- [x] Verify native builds (@dmlloyd)

Please submit supplementary PRs to https://github.com/dmlloyd/quarkus/tree/config-updates-final.

PR #5455, opened 2019-11-13
Support * Wildcard for quarkus.http.cors.origins

### LINKED ISSUE #5422: quarkus.http.cors.origins=* should be supported
**Describe the bug**
When I read the documentation about configuring cors for the first time, I tought that leaving "quarkus.http.cors.origins" as empty would be the same as "Access-Control-Allow-Origin: *".
The * wildcard allows any origin to access ressources.

Furthermore, a value of "*" is currently not supported. See CORSFilter.java#L93.
`boolean allowsOrigin = corsConfig.origins.isEmpty() || corsConfig.origins.contains(origin);`

We should try to improve the documentation to reflect what is actually going on in the filter.

**Expected behavior**

> For requests without credentials, the literal value "*" can be specified, as a wildcard; the value tells browsers to allow requesting code from any origin to access the resource.

If I use "*" as wildcard, I expect quarkus to allow any origin to access my resources.
If I remove "quarkus.http.cors.origins" in my configuration, I expect quarkus to use "*" as default.
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin

**Actual behavior**
Requests get rejected, altough quarkus.http.cors.origins is set to "*".

**Configuration**
```properties
uarkus.http.cors=true
quarkus.http.cors.origins=*

```

SYS_NT-10.0 DESKTOP-0TKUM9I 2.11.2(0.329/5/3) 2018-11-10 14:38 x86_64 Msys
 - Output of `java -version`:
java version "1.8.0_211"
Java(TM) SE Runtime Environment (build 1.8.0_211-b12)
Java HotSpot(TM) 64-Bit Server VM (build 25.211-b12, mixed mode)
 - Quarkus version or git rev: 1.0.0.CR1

Postremus:
The star wildcard for the origins means, that any origin can access the service.
Previosly, this behavior could only be achieved by leaving the config property out completly.
Now you can either not set the origins config property, or use "*".

Adjust the documentation for * origins.
Fix naming of one of the tests.

fixes #5422

### DIFF
--- a/docs/src/main/asciidoc/http-reference.adoc
+++ b/docs/src/main/asciidoc/http-reference.adoc
@@ -127,13 +127,13 @@ following properties will be applied before passing the request on to its actual
 [cols="<m,<m,<2",options="header"]
 |===
 |Property Name|Default|Description
-|quarkus.http.cors.origins||The comma-separated list of origins allowed for CORS. The filter allows any origin if this is not
-set.
-|quarkus.http.cors.methods||The comma-separated list of HTTP methods allowed for CORS. The filter allows any method if this is
-not set.
-|quarkus.http.cors.headers||The comma-separated list of HTTP headers allowed for CORS. The filter allows any header if this is
-not set.
-|quarkus.http.cors.exposed-headers||The comma-separated list of HTTP headers exposed in CORS.
+|quarkus.http.cors.origins|*|The comma-separated list of origins allowed for CORS. The filter allows any origin if this is not set or set to '*'.
+|quarkus.http.cors.methods|*|The comma-separated list of HTTP methods allowed for CORS. The filter allows any method if this is
+not set or set to '*'.
+|quarkus.http.cors.headers|*|The comma-separated list of HTTP headers allowed for CORS. The filter allows any header if this is
+not set or set to '*'.
+|quarkus.http.cors.exposed-headers|*|The comma-separated list of HTTP headers exposed in CORS. The filter allows any headers to be exposed if this is
+not set or set to '*'.
 |quarkus.http.cors.access-control-max-age||The duration (see note below) indicating how long the results of a pre-flight request can be cached.
 This value will be returned in a `Access-Control-Max-Age` response header.
 |===
--- a/extensions/vertx-http/deployment/src/test/java/io/quarkus/vertx/http/cors/CORSSecurityTestCase.java
+++ b/extensions/vertx-http/deployment/src/test/java/io/quarkus/vertx/http/cors/CORSSecurityTestCase.java
@@ -18,7 +18,7 @@ import io.quarkus.vertx.http.security.PathHandler;
 import io.quarkus.vertx.http.security.TestIdentityController;
 import io.quarkus.vertx.http.security.TestIdentityProvider;
 
-public class CoresSecurityTestCase {
+public class CORSSecurityTestCase {
 
     private static final String APP_PROPS = "" +
             "quarkus.http.cors=true\n" +
--- a/extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/cors/CORSFilter.java
+++ b/extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/cors/CORSFilter.java
@@ -27,7 +27,7 @@ public class CORSFilter implements Handler<RoutingContext> {
     }
 
     private void processRequestedHeaders(HttpServerResponse response, String allowHeadersValue) {
-        if (!corsConfig.headers.isPresent()) {
+        if (!corsConfig.headers.isPresent() || containsOnlyWildcard(corsConfig.headers)) {
             response.headers().set(HttpHeaders.ACCESS_CONTROL_ALLOW_HEADERS, allowHeadersValue);
         } else {
             List<String> requestedHeaders = new ArrayList<>();
@@ -49,7 +49,7 @@ public class CORSFilter implements Handler<RoutingContext> {
     }
 
     private void processMethods(HttpServerResponse response, String allowMethodsValue) {
-        if (!corsConfig.methods.isPresent()) {
+        if (!corsConfig.methods.isPresent() || containsOnlyWildcard(corsConfig.methods)) {
             response.headers().set(HttpHeaders.ACCESS_CONTROL_ALLOW_METHODS, allowMethodsValue);
         } else {
             List<String> requestedMethods = new ArrayList<>();
@@ -58,9 +58,10 @@ public class CORSFilter implements Handler<RoutingContext> {
             }
 
             List<String> validRequestedMethods = new ArrayList<>();
-            for (HttpMethod configMethod : corsConfig.methods.get()) {
-                if (requestedMethods.contains(configMethod.name().toLowerCase())) {
-                    validRequestedMethods.add(configMethod.name());
+            for (String configMethod : corsConfig.methods.get()) {
+                String methodName = configMethod.trim();
+                if (requestedMethods.contains(methodName.toLowerCase())) {
+                    validRequestedMethods.add(methodName);
                 }
             }
 
@@ -91,7 +92,8 @@ public class CORSFilter implements Handler<RoutingContext> {
                 processRequestedHeaders(response, requestedHeaders);
             }
 
-            boolean allowsOrigin = !corsConfig.origins.isPresent() || corsConfig.origins.get().contains(origin);
+            boolean allowsOrigin = containsOnlyWildcard(corsConfig.origins)
+                    || corsConfig.origins.orElse(Collections.emptyList()).contains(origin);
 
             if (allowsOrigin) {
                 response.headers().set(HttpHeaders.ACCESS_CONTROL_ALLOW_ORIGIN, origin);
@@ -99,22 +101,33 @@ public class CORSFilter implements Handler<RoutingContext> {
 
             response.headers().set(HttpHeaders.ACCESS_CONTROL_ALLOW_CREDENTIALS, "true");
 
-            final Optional<List<String>> exposedHeaders = corsConfig.exposedHeaders;
+            if (corsConfig.exposedHeaders.isPresent()) {
+                List<String> exposedHeaders = corsConfig.exposedHeaders.orElse(Collections.emptyList());
+                if (exposedHeaders.contains("*") && exposedHeaders.size() > 1) {
+                    exposedHeaders.remove("*");
+                }
 
-            if (exposedHeaders.isPresent()) {
-                response.headers().set(HttpHeaders.ACCESS_CONTROL_EXPOSE_HEADERS,
-                        String.join(",", exposedHeaders.orElse(Collections.emptyList())));
+                response.headers().set(HttpHeaders.ACCESS_CONTROL_EXPOSE_HEADERS, String.join(",", exposedHeaders));
             }
 
             if (request.method().equals(HttpMethod.OPTIONS)) {
                 if ((requestedHeaders != null || requestedMethods != null) && corsConfig.accessControlMaxAge.isPresent()) {
-                    response.putHeader(HttpHeaders.ACCESS_CONTROL_MAX_AGE,
-                            String.valueOf(corsConfig.accessControlMaxAge.get().getSeconds()));
+                    response.putHeader(HttpHeaders.ACCESS_CONTROL_MAX_AGE, String.valueOf(corsConfig.accessControlMaxAge.get()
+                            .getSeconds()));
                 }
                 response.end();
             } else {
                 event.next();
             }
         }
+    }
+
+    private boolean containsOnlyWildcard(Optional<List<String>> optionalList) {
+        if (!optionalList.isPresent()) {
+            return true;
+        }
+
+        List<String> list = optionalList.get();
+        return list.isEmpty() || list.size() == 1 && "*".equals(list.get(0));
     }
 }
--- a/extensions/vertx-http/deployment/src/test/java/io/quarkus/vertx/http/cors/CORSWildcardSecurityTestCase.java
+++ b/extensions/vertx-http/deployment/src/test/java/io/quarkus/vertx/http/cors/CORSWildcardSecurityTestCase.java
@@ -1,0 +1,149 @@
+package io.quarkus.vertx.http.cors;
+
+import static io.restassured.RestAssured.given;
+
+import java.util.function.Supplier;
+
+import org.hamcrest.Matchers;
+import org.jboss.shrinkwrap.api.ShrinkWrap;
+import org.jboss.shrinkwrap.api.asset.StringAsset;
+import org.jboss.shrinkwrap.api.spec.JavaArchive;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.RegisterExtension;
+
+import io.quarkus.test.QuarkusUnitTest;
+import io.quarkus.vertx.http.security.PathHandler;
+import io.quarkus.vertx.http.security.TestIdentityController;
+import io.quarkus.vertx.http.security.TestIdentityProvider;
+
+public class CORSWildcardSecurityTestCase {
+
+    private static final String APP_PROPS = "" +
+            "quarkus.http.cors=true\n" +
+            "quarkus.http.auth.basic=true\n" +
+            "quarkus.http.auth.policy.r1.roles-allowed=test\n" +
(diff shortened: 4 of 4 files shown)


extensions/vertx-http/deployment/src/test/java/io/quarkus/vertx/http/cors/CORSSecurityTestCase.java
```diff
@@ -18,7 +18,7 @@
 import io.quarkus.vertx.http.security.TestIdentityController;
 import io.quarkus.vertx.http.security.TestIdentityProvider;
 
-public class CoresSecurityTestCase {
+public class CORSSecurityTestCase {
```
machi1990: Good catch.

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/cors/CORSFilter.java
```diff
@@ -90,7 +90,7 @@ public void handle(RoutingContext event) {
                 processRequestedHeaders(response, requestedHeaders);
             }
 
-            boolean allowsOrigin = corsConfig.origins.isEmpty() || corsConfig.origins.contains(origin);
+            boolean allowsOrigin = corsConfig.origins.contains("*") || corsConfig.origins.contains(origin);
```
sberyozkin: @Postremus Can you please  check it is a single value list too and use `get(0).equals` ? We probably should not allow some random lists which might have non wildcard and wildcard values :-)
There also should be a unit test, please copy and paste one of the tests there to test a `*`

extensions/vertx-http/deployment/src/test/java/io/quarkus/vertx/http/cors/CORSSecurityTestCase.java
sberyozkin: @Postremus thanks for spotting the typo, it is a minor issue but lets keep the original naming pattern, `CorsSecurityTestCase`

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/cors/CORSConfig.java
```diff
@@ -19,7 +19,7 @@
      *
      * default: returns any requested origin as valid
      */
-    @ConfigItem
+    @ConfigItem(defaultValue = "*")
```
sberyozkin: I'm not sure it should be a default value - that would mean all other properties should also be defaulted to `*`

extensions/vertx-http/deployment/src/test/java/io/quarkus/vertx/http/cors/CORSSecurityTestCase.java
Postremus: @sberyozkin The other test classes in this package are also named CORSXXXXTestCase, which is why I went for this name.

docs/src/main/asciidoc/http-reference.adoc
```diff
@@ -127,7 +127,7 @@ following properties will be applied before passing the request on to its actual
 [cols="<m,<m,<2",options="header"]
 |===
 |Property Name|Default|Description
-|quarkus.http.cors.origins||The comma-separated list of origins allowed for CORS. The filter allows any origin if this is not
+|quarkus.http.cors.origins|*|The comma-separated list of origins allowed for CORS. The filter allows any origin if this is not
```
sberyozkin: See my note about the default value. Perhaps it should read: `The filter allows any origin if this is not set or set to '*'' ?

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/cors/CORSConfig.java
machi1990: We should probably revert this and allowsAllOrigin if the list contains only `*` or `isEmpty()`

sberyozkin: @machi1990 yes, I agree

Postremus: We may need to allow `*` for more of the cors properties anyway in the future.
`Access-Control-Expose-Headers`, `Access-Control-Allow-Methods` and `Access-Control-Allow-Headers` allow * as wildcard value.

sberyozkin: @Postremus sure, it makes sense for the consistency purposes. But lets keep the current pattern, it is either absent or set to `*`

sberyozkin:
@Postremus Hi, can you please do `mvn process-sources` in `quarkus-vertx-http` ?

As mentioned above, for the sake of consistency, all other CORS properties which were allowing for a wildcard when omitted should also be updated.

I have to admit I'm a bit not sure now what is the best approach, the one your originally started with (property default to `*`) or the one I proposed yesterday (keep the original pattern). @machi1990, should we ask @Postremus to continue with his original idea ? It seemed to me late yesterday that it was changing things, but now, I'm thinking, if a given property is omitted => it is an indirect `*` or it can be defaulted to `*` from the start, in which a given property is never omitted OOB, yet it does not change anything, it is still `*` by default :-).
CC @gsmet

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/cors/CORSFilter.java
```diff
@@ -90,7 +90,8 @@ public void handle(RoutingContext event) {
                 processRequestedHeaders(response, requestedHeaders);
             }
 
-            boolean allowsOrigin = corsConfig.origins.isEmpty() || corsConfig.origins.contains(origin);
+            boolean isWildcardOrigin = corsConfig.origins.size() == 1 && "*".equals(corsConfig.origins.get(0));
```
gsmet: Hmmm, isn't it a breaking change? If it's empty, it used to pass, now it doesn't. Is it really what we want? Or did I miss something.

Blocking merge until I got an answer to that.

gsmet (changes requested):

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/cors/CORSFilter.java
sberyozkin: Good catch, `corsConfig.origins.isEmpty()` should still be there.

sberyozkin:
@Postremus 
I'm sorry for my rather confusing review, I did it in a great rush. So, here is a proposal: please revert to your original idea by defaulting to `*` as it makes things simpler without breaking the compatibility, only enforce that a list containing a `*` is a single element list. And make sure that all other similar properties are supported consistently. Give it a go please when you get a chance

gsmet:
I would have liked to have this one in 1.0.0.Final. Can someone finish it before Friday end of day?

Postremus:
@sberyozkin sorry for not responding earlier.
@gsmet I will give it a go.

machi1990:
> @sberyozkin sorry for not responding earlier.
> @gsmet I will give it a go.

Thanks @Postremus

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/cors/CORSConfig.java
```diff
@@ -6,42 +6,37 @@
...
      *
-     * Comma separated list of valid methods. ex: GET,PUT,POST
-     * The filter allows any method if this is not set.
+     * Comma separated list of valid methods. ex: GET,PUT,POST The filter allows any method if this is not set.
      *
      * default: returns any requested method as valid
      */
-    @ConfigItem
-    public List<HttpMethod> methods;
+    public List<String> methods;
```
gsmet: Why this change?

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/cors/CORSFilter.java
```diff
@@ -48,7 +47,7 @@ private void processRequestedHeaders(HttpServerResponse response, String allowHe
     }
 
     private void processMethods(HttpServerResponse response, String allowMethodsValue) {
-        if (corsConfig.methods.isEmpty()) {
+        if (containsOnlyWildcard(corsConfig.headers)) {
```
gsmet: I don't understand this change?

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/cors/CORSFilter.java
```diff
@@ -90,29 +89,36 @@ public void handle(RoutingContext event) {
...
                 }
                 response.end();
             } else {
                 event.next();
             }
         }
     }
+
+    private boolean containsOnlyWildcard(List<String> list) {
+        return list.size() == 1 && "*".equals(list.get(0));
```
gsmet: I asked for it to support the empty list and I don't think you have addressed it.

gsmet (changes requested):
I added a few comments/questions, I don't think we will converge on that one today.

Let's target the next release.

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/cors/CORSConfig.java
Postremus: We can't use an Enum here, when we want to support * wildcard for methods as well.

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/cors/CORSFilter.java
Postremus: This change is done so that we also support * for methods as well. See new diff.

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/cors/CORSFilter.java
Postremus: done

gsmet:
Oh and I just stumbled upon the fact that this will conflict with https://github.com/quarkusio/quarkus/pull/5387/commits . We want to get this one in the tree ASAP as it's a very large one and it's a pain to rebase.

We will have to revisit that one after the merge.

machi1990:
@Postremus I noticed a formatting issue in the CI, could you run `mvn formatter:format` in the `vertx-http` module? and force push your changes. Thanks.

Postremus:
@machi1990 done.

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/cors/CORSConfig.java
```diff
@@ -6,69 +6,58 @@
...
-     * Comma separated list of valid headers. ex: X-Custom,Content-Disposition
-     *
-     * default: empty
+     * HTTP headers exposed in CORS.
+     * <p>
+     * Comma separated list of exposed headers e.g. X-Custom,Content-Disposition.
      */
-    @ConfigItem
+    @ConfigItem(defaultValue = "*")
     public Optional<List<String>> exposedHeaders;
```
gsmet: @Postremus sorry I'm getting back to this one a bit late. I rebased and added a commit to improve the documentation formatting.

And while doing so, I saw this change. It looks like a significant change in behavior or am I wrong? Did it really expose all headers by default before?

gsmet (changes requested):
I asked an additional question.

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/cors/CORSConfig.java
gsmet: @Postremus ping?

Postremus: @gsmet pong :)
You are right. That is quite the change from previous behavior. I removed this change.

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/cors/CORSFilter.java
```diff
@@ -27,7 +27,7 @@ public CORSFilter(CORSConfig corsConfig) {
     }
 
     private void processRequestedHeaders(HttpServerResponse response, String allowHeadersValue) {
-        if (!corsConfig.headers.isPresent()) {
+        if (!corsConfig.headers.isPresent() || containsOnlyWildcard(corsConfig.headers)) {
```
gsmet: You could remove the first part of the `if`.

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/cors/CORSFilter.java
```diff
@@ -49,7 +49,7 @@ private void processRequestedHeaders(HttpServerResponse response, String allowHe
     }
 
     private void processMethods(HttpServerResponse response, String allowMethodsValue) {
-        if (!corsConfig.methods.isPresent()) {
+        if (!corsConfig.methods.isPresent() || containsOnlyWildcard(corsConfig.methods)) {
```
gsmet: You could remove the first part of the `if`.

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/cors/CORSFilter.java
```diff
@@ -58,9 +58,10 @@ private void processMethods(HttpServerResponse response, String allowMethodsValu
             }
 
             List<String> validRequestedMethods = new ArrayList<>();
-            for (HttpMethod configMethod : corsConfig.methods.get()) {
-                if (requestedMethods.contains(configMethod.name().toLowerCase())) {
-                    validRequestedMethods.add(configMethod.name());
+            for (String configMethod : corsConfig.methods.get()) {
```
gsmet: I think you mistakenly reverted the config changes necessary for this to work?

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/cors/CORSFilter.java
```diff
@@ -91,30 +92,42 @@ public void handle(RoutingContext event) {
...
                 response.headers().set(HttpHeaders.ACCESS_CONTROL_ALLOW_ORIGIN, origin);
             }
 
             response.headers().set(HttpHeaders.ACCESS_CONTROL_ALLOW_CREDENTIALS, "true");
 
-            final Optional<List<String>> exposedHeaders = corsConfig.exposedHeaders;
+            if (corsConfig.exposedHeaders.isPresent()) {
+                List<String> exposedHeaders = corsConfig.exposedHeaders.orElse(Collections.emptyList());
+                if (exposedHeaders.contains("*") && exposedHeaders.size() > 1) {
+                    exposedHeaders.remove("*");
```
gsmet: I don't understand why this is necessary?

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/cors/CORSFilter.java
```diff
@@ -91,30 +92,42 @@ public void handle(RoutingContext event) {
...
 
             if (allowsOrigin) {
                 response.headers().set(HttpHeaders.ACCESS_CONTROL_ALLOW_ORIGIN, origin);
             }
 
             response.headers().set(HttpHeaders.ACCESS_CONTROL_ALLOW_CREDENTIALS, "true");
 
-            final Optional<List<String>> exposedHeaders = corsConfig.exposedHeaders;
+            if (corsConfig.exposedHeaders.isPresent()) {
+                List<String> exposedHeaders = corsConfig.exposedHeaders.orElse(Collections.emptyList());
```
gsmet: You already checked the value is present. No need for a `orElse` (I know it was preexisting but let's get rid of it).

gsmet (changes requested):
I added some comments inline.

I think you reverted some config changes that are needed. Unless I miss something.

sberyozkin:
Closing this one after #12142 has been merged, @Postremus thanks for starting this effort earlier

PR #5459, opened 2019-11-13
Change the container build detection for Kafka Streams

stuartwdouglas:


cescoffier (approved):

cescoffier:
BTW, this pattern should be documented in the extension guide.

PR #5464, opened 2019-11-14
Maven plugin's devmode doesn't take into account versions overridden through system properties or profiles

### LINKED ISSUE #4289: Maven plugin's devmode doesn't take into account versions overridden through system properties or profiles
If I run devmode with an application with a pom like this:
```
(...)
<properties>
  <quarkus.version>0.21.0</quarkus.version>
</properties>

    <dependencyManagement>
        <dependencies>
            <dependency>
                <groupId>io.quarkus</groupId>
                <artifactId>quarkus-bom</artifactId>
                <type>pom</type>
                <scope>import</scope>
            </dependency>
        </dependencies>
    </dependencyManagement>

(... and then whatever dependencies I need)
```

but I override the `quarkus.version` from the command line (as an argument to the `mvn` command), it will still spin up Quarkus 0.21.0 instead of the version that I requested (it will execute the Maven plugin of the correct version, but the Maven plugin will run the wrong version of Quarkus). The same thing happens if I add and activate a specific Maven profile which overrides the version.

aloubyansky:
For some reason the pom provided by the LocalWorkspace instance is not interpolated by Maven.

It is also important to use MojoExecutor API to invoke `compile` on the project instead of using the Maven Invoker because MojoExecutor will share the reactor with the dev mojo that can later be re-used as the workspace reader.

Fixes #4289 

Also enables a workaround for #3539 by adding `maven.home` system property set by Maven launcher to propagate the Maven home dir to Quarkus bootstrap used in the tests. E.g.
````
      <plugin>
        <version>${surefire-plugin.version}</version>
        <configuration>
          <systemProperties>
            <java.util.logging.manager>org.jboss.logmanager.LogManager</java.util.logging.manager>
            <maven.home>${maven.home}</maven.home>
          </systemProperties>
        </configuration>
      </plugin>
````

### DIFF
--- a/devtools/maven/pom.xml
+++ b/devtools/maven/pom.xml
@@ -21,12 +21,6 @@
         <dependency>
             <groupId>io.quarkus</groupId>
             <artifactId>quarkus-bootstrap-core</artifactId>
-            <exclusions>
-                <exclusion>
-                    <groupId>org.apache.maven.wagon</groupId>
-                    <artifactId>wagon-provider-api</artifactId>
-                </exclusion>
-            </exclusions>
         </dependency>
         <dependency>
             <groupId>io.quarkus</groupId>
@@ -59,10 +53,6 @@
             <artifactId>jakarta.enterprise.cdi-api</artifactId>
         </dependency>
         <dependency>
-            <groupId>org.apache.maven.shared</groupId>
-            <artifactId>maven-invoker</artifactId>
-        </dependency>
-        <dependency>
             <groupId>org.apache.maven</groupId>
             <artifactId>maven-core</artifactId>
             <exclusions>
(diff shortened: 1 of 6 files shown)


gsmet:
I think I prefer we wait for 1.1.0 for this one.

gsmet:
@aloubyansky can you rebase?

aloubyansky:
Apparently, using Maven reactor as a workspace reader breaks POM updates in dev mode.

aloubyansky:
@gsmet do you think we can get this in 1.1.0.Final?

gsmet:
Well, the issue is that it's very last minute and we have absolutely 0 feedback on it.

aloubyansky:
Ok, I won't mind postponing it. Thanks.

gsmet (approved):

gsmet:
Let's merge it to master though.

PR #5466, opened 2019-11-14
ArC - fix resolution if the required type is wildcard

### LINKED ISSUE #3311: ArC removes indirect referenced beans
Quarkus in version 19.1 removes indirect referenced beans.

Lets have following scenario. You define an interface for your strategies and an services that depending on any argument, returns the appropriate strategy. These strategies are defined as beans. As these beans are not referenced directly with injections, but only through the interface, the are removed by ArC. The only way to fix it, is by setting `quarkus.arc.remove-unused-beans=false`

```java
public interface Exporter<T extends Exportable> {
    boolean accept(Class<T> clazz);
    String export(T object);
}

@ApplicationScoped
public class ApplicationExporter implements Exporter<Application> {
   boolean accept(Class<Application> clazz) { return clazz instanceof Application; }
   String export(Application application) {
     ...
   }
}

@ApplicationScoped
public class ExporterService {
  @Inject
  Instance<Exporter<? extends Exportable>> exporters; 
  public <T>Exporter<T> getExporter(Class<T> exporterClazz) {
    // BUG: exporters are empty!
    return StreamSupport.stream(exporters.spliterator(), false)
            .filter(exporter -> exporter.accept(exporterClazz))
            .findFirst().orElseThrow(...);
  }
}
```

If there is an `@Inject javax.enterprise.inject.Instance<T>` Quarkus (ArC) should keep all classes implementing `T`.

mkouba:
- resolves #3311

### DIFF
--- a/independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/BeanResolver.java
+++ b/independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/BeanResolver.java
@@ -125,6 +125,8 @@ class BeanResolver {
                 }
                 return true;
             }
+        } else if (WILDCARD_TYPE.equals(requiredType.kind())) {
+            return parametersMatch(requiredType, beanType);
         }
         return false;
     }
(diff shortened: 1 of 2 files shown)


mkouba:
@dkellenb this PR should fix the bug I've mentioned in https://github.com/quarkusio/quarkus/issues/3311. It would be great if you could verify whether it fixes your problem too.

independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/BeanResolver.java
```diff
@@ -125,6 +125,8 @@ boolean matchesNoBoxing(Type requiredType, Type beanType) {
                 }
                 return true;
             }
+        } else if (WILDCARD_TYPE.equals(requiredType.kind()) || TYPE_VARIABLE.equals(requiredType.kind())) {
```
manovotn: The tests don't cover the type variable case, or am I missing something?

manovotn (approved):

PR #5470, opened 2019-11-14
Add quarkus-jackson to quarkus-oidc

### LINKED ISSUE #5342: quarkus-oidc does not work in the native mode without quarkus-jackson
**Description**
I had to add `quatkus-jackson` in a few cases as a test dependency (VertX JSONObject depends on Jackson I believe). in `extension/oidc/runtime` for example. 

**Implementation ideas**
Add it as a strong dependency to avoid asking the users to add it; it should not interfere with JSONB JAX_RS providers for example

sberyozkin:
Fixes #5342.


sberyozkin:
Hi @gsmet @cescoffier I'm afraid of breaking things and not adding a Backport label, I'd rather say to 1.0.0 users they need to add a quarkus-jackson dependency manually if they need to run the oidc adapter in the native mode

gsmet:
Failure is unrelated, merging.

gsmet (changes requested):
@sberyozkin what's the relation between adding `quarkus-jackson` and removing the Keycloak test resource?

Blocking the merge until I've got an answer for that.

sberyozkin:
@gsmet not related, but I was cleaning up along the way, while checking the tests, loos like there were needed at some point of time to ensure `keycloak.url` was available to the tests (cc @pedroigor ) but I've checked if I can remove them and it worked :-)

gsmet:
And it works locally too without a Keycloak server started? Because I seem to recall we have one running on CI.

pedroigor:
> @gsmet not related, but I was cleaning up along the way, while checking the tests, looks like they were needed at some point of time to ensure `keycloak.url` was available to the tests (cc @pedroigor ) but I've checked if I can remove them and it worked :-)

I've added that due to https://github.com/quarkusio/quarkus/blob/c497d79e48c334955bc2acfeb03d4e249b382a92/integration-tests/oidc/src/test/java/io/quarkus/it/keycloak/KeycloakTestResource.java#L13. If it works now without it, +1 to remove it.

sberyozkin:
@pedroigor @gsmet do you mean if I've run the tests locally ? Yes I did.

sberyozkin:
@pedroigor I was not changing the URLs wit `-Dkeycloak.utl=someother url`- I just run the tests with `-Pdocker`. I can just restore those `KeycloakResource` files if that is still needed in some situations

sberyozkin:
@gsmet by the way I'm not sure if the OIDC native test passed in CI, looks like the job was terminated due to to some download issue ?

sberyozkin:
@gsmet IMHO this is good to go, as I said I've confirmed the native tests work in the master, both `oidc` and `oidc-code-flow` against a locally launched KC image. But the actual fix is not for 1.0.0.Final so the merge can wait till you may want to check it locally as well... Thanks

gsmet (approved):

gsmet:
Thanks!

PR #5475, opened 2019-11-14
Make sure system properties dominate the properties defined in pom.xml

### LINKED ISSUE #3702: QuarkusTest does not take into consideration dependencies defined in the maven profile
**Describe the bug**
When running maven build with tests (`@QuarkusTest`) that requires dependencies defined in the profile that fails with Arc dependency resolution exceptions due to missing classes.

**Expected behavior**
`@QuarkusTest` should see dependencies define in the profile when the profile is active 

**Actual behavior**
Dependencies defined in the profile are not seen by `@QuarkusTest`

**To Reproduce**
1. Create a test case that expects to have bean injected from a dependency that is defined in maven profile
2. run a test with activated profile

**Configuration**
no config needed

**Screenshots**

 - Output of `uname -a` or `ver`: Darwin mac 18.7.0 Darwin Kernel Version 18.7.0: Thu Jun 20 18:42:21 PDT 2019; root:xnu-4903.270.47~4/RELEASE_X86_64 x86_64
 - Output of `java -version`: java version "1.8.0_171"
 - GraalVM version (if different from Java): 19.2.0
 - Quarkus version or git rev: master

aloubyansky:
… when bootstrap LocalWorkspace is used as a WorkspaceReader

Fixes #3702

### DIFF
--- a/independent-projects/bootstrap/core/src/test/java/io/quarkus/bootstrap/resolver/ResolverSetupCleanup.java
+++ b/independent-projects/bootstrap/core/src/test/java/io/quarkus/bootstrap/resolver/ResolverSetupCleanup.java
@@ -5,6 +5,7 @@ import java.nio.file.Path;
 import java.util.UUID;
 
 import io.quarkus.bootstrap.resolver.maven.MavenArtifactResolver;
+import io.quarkus.bootstrap.resolver.maven.workspace.LocalWorkspace;
 import io.quarkus.bootstrap.util.IoUtils;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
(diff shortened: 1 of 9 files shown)


gsmet:
It's too late to backport this one IMHO. Let's be safe.

independent-projects/bootstrap/core/src/test/java/io/quarkus/bootstrap/resolver/ResolverSetupCleanup.java
```diff
@@ -24,7 +25,7 @@
     public void setup() throws Exception {
         workDir = initWorkDir();
         repoHome = IoUtils.mkdirs(workDir.resolve("repo"));
-        resolver = initResolver();
+        resolver = initResolver(null);
```
gsmet: We really want to pass null here?

gsmet (changes requested):
@aloubyansky I added one small question. Once it's cleared, I think we can merge it.

Better do it early in the cycle.

independent-projects/bootstrap/core/src/test/java/io/quarkus/bootstrap/resolver/ResolverSetupCleanup.java
aloubyansky: Yes, it means we are not setting up a workspace reader.

gsmet (approved):

gsmet:
Merged, thanks!

PR #5476, opened 2019-11-14
Remove duplicate dependency from integration test module

Sanne:
This dependency is listed twice in the same module. Maven seems to not complain during a regular build, but this fails hard when trying to build subsets of the Quarkus project using Maven's `-amd` to build all dependant modules of a list of projects.


aloubyansky (approved):

machi1990 (approved):
Good catch.

gsmet:
Adding the `backport?` as the issue is also present in the `1.0` branch.

PR #5477, opened 2019-11-14
Fix #2299: mention service files in extension guide

### LINKED ISSUE #2299: Extension author guide: explain how to deal with `META-INF/services` files
https://quarkus.io/guides/extension-authors-guide#register-resources-for-use-in-native-image led me to believe I only needed to register the file, but actually I need to call `ServiceUtil` to register the contents of the file for reflection, at the very least.

FroMage:


FroMage:
@stuartwdouglas you wanna review that?

docs/src/main/asciidoc/writing-extensions.adoc
```diff
@@ -2025,16 +2025,103 @@ public final class MyExtProcessor {
         resource.produce(new NativeImageResourceBuildItem("/security/runtime.keys")); //<1>
 
         resource.produce(new NativeImageResourceBuildItem(
-                "META-INF/services/" + io.quarkus.SomeService.class.getName())); //<2>
+                "META-INF/my-descriptor.xml")); //<2>
```
Sanne: why is that having an `xml` postfix?

docs/src/main/asciidoc/writing-extensions.adoc
```diff
@@ -2025,16 +2025,103 @@ public final class MyExtProcessor {
...
+        Set<String> implementations = 
+            ServiceUtil.classNamesNamedIn(Thread.currentThread().getContextClassLoader(),
+                                          service);
+        reflectionClasses.produce(
+            new ReflectiveClassBuildItem(true, true, implementations.toArray(new String[0])));
+    }
+}
+----
+
+This is the easiest way to get your services running natively, but it's less efficient than scanning the implementation
```
Sanne: You do hint that this is not the preferred solution, but could we lead with that? It's really not what we want

I'd have something like this at the very beginning:

```
Services normally are loaded by using a combination of a resource file and reflection on the classname; this implies that registering a service can be done with a quick hack by registering the resource and the reflection as in the following example, but this is not the best solution:
```

docs/src/main/asciidoc/writing-extensions.adoc
```diff
@@ -2025,16 +2025,103 @@ public final class MyExtProcessor {
...
+        Collection<Class<? extends io.quarkus.SomeService>> implementationClasses = new LinkedHashSet<>();
+        Set<String> implementations = ServiceUtil.classNamesNamedIn(Thread.currentThread().getContextClassLoader(),
+                                                                    service);
+        for(String implementation : implementations) {
+            implementationClasses.add((Class<? extends io.quarkus.SomeService>) 
+                recorderContext.classProxy(implementation));
+        }
+        
+        // produce a static-initializer with those classes
+        recorder.configure(implementationClasses);
```
Sanne: this is "second tier" preference.. better to store the instances than the class?

docs/src/main/asciidoc/writing-extensions.adoc
FroMage: I didn't know what to put there, and didn't want to mention service files because that's never enough. Better suggestion?

docs/src/main/asciidoc/writing-extensions.adoc
FroMage: I can, but in general I prefer having simpler examples first, because they help understanding the bigger one later. If you feel we must reverse them, I can. An alternative is to intro like you said with a strong statement/warning, but keep the "hack/wrong/simple" solution first.

docs/src/main/asciidoc/writing-extensions.adoc
FroMage: So, in my MP-CP extension I did that, except I just checked and I'm the only extension to do that. All others (including ORM) pass classes to the recorder instead of instances. I don't understand why, but assumed I must be doing it wrong, so didn't document my way :(

docs/src/main/asciidoc/writing-extensions.adoc
Sanne: nevermind, I got confused from the previous example - I thought it was hinting that we have a way to load services from XML files :D

docs/src/main/asciidoc/writing-extensions.adoc
Sanne: ok fair enough

docs/src/main/asciidoc/writing-extensions.adoc
Sanne: no ORM actually does something entirely different for most services ..  you probably saw how integrators are registered, that's the only case following such a pattern.

but ok I wouldn't want to have to explain that, it's a mess and definitely not how we should suggest others doing it. But the ORM approach results in far better optimisations, I guess we'll have to explore how to generalize the solution.

Sanne: (that's just FYI, the patch is fine)

docs/src/main/asciidoc/writing-extensions.adoc
Sanne: I guess if this is about general purpose resources I'd not have it prefixed with META-INF, but I'm happy to merge it as is.

Sanne (approved):

docs/src/main/asciidoc/writing-extensions.adoc
```diff
@@ -2025,16 +2025,103 @@ public final class MyExtProcessor {
...
+        // register the service file so it is visible in native-image
+        resource.produce(new NativeImageResourceBuildItem(service));
+        
+        // register every listed implementation class so they can be inspected/instantiated 
+        // in native-image at run-time
+        Set<String> implementations = 
+            ServiceUtil.classNamesNamedIn(Thread.currentThread().getContextClassLoader(),
+                                          service);
+        reflectionClasses.produce(
+            new ReflectiveClassBuildItem(true, true, implementations.toArray(new String[0])));
```
gsmet: What about `ServiceProviderBuildItem` ? I think this would simplify most of what you did there.

docs/src/main/asciidoc/writing-extensions.adoc
```diff
@@ -2025,16 +2025,103 @@ public final class MyExtProcessor {
...
+            implementationClasses.add((Class<? extends io.quarkus.SomeService>) 
+                recorderContext.classProxy(implementation));
+        }
+        
+        // produce a static-initializer with those classes
+        recorder.configure(implementationClasses);
+    }
+}
+
+@Recorder
```
gsmet: I think I would separate the two classes with a sentence making it clear that they are in 2 different contexts (deployment vs runtime artifact)

gsmet:
I won't backport it for now as there are things that should be improved.

Will backport it though if I have a follow up PR.

docs/src/main/asciidoc/writing-extensions.adoc
FroMage: HAH. Well, it's brand new (three weeks) and undocumented, but actually worse: there's a competing implementation of the same name in the extension author guide, which doesn't do the same at all: it aggregates all services of the same name into a single service file.

This is VERY confusing. How can we resolve this?

docs/src/main/asciidoc/writing-extensions.adoc
FroMage: OK, then I should make another PR, right?

docs/src/main/asciidoc/writing-extensions.adoc
stuartwdouglas: The newness is only because of the package rename, the two that are there are exactly the same.

gsmet: The one that should be used is the one in the `.nativeimage.` package. Otherwise, it has been there since before the shamrock renaming in the `substrate` package.

The `substrate` one is deprecated and deprecation is handled via producing the `.nativeimage.` ones.

docs/src/main/asciidoc/writing-extensions.adoc
gsmet: Yes, please. With the `backport?` tag too.

docs/src/main/asciidoc/writing-extensions.adoc
FroMage: OK, so it was just not documented in the guide and we still have a class with the same name in the guide. I'll rename the guide class then.

FroMage: OK wait. I don't understand this build item. It appears to do two things:

- register the service file as a resource
- register the classes for reflection

But it's invoked mostly while specifying specific implementation classes, and only once by reading the actual service file to extract the implementation lists. It does not appear to write the service file with the list of resources it's given (unlike the version with the same name in the guide). So how does it actually work? If all the callers specify implementation classes that do not actually end up in the service files, how can the service loader see them?

Or are they _also_ manually put in the service file and the list of parameters happens to match what's in the existing file?

That's a pretty important thing to clear up @stuartwdouglas please?

FroMage: @stuartwdouglas I can document the fact that this build item depends on the list of implementations passed as params matching existing service files, but I need your confirmation that it does this to unstuck my PR.

stuartwdouglas: It does. Really we should just have a simple way of enabling all services by reading the service files from the deployment, but I guess it is not a major issue as nobody has really complained yet.

FroMage: OK thanks, I'll document this then.

FroMage:
OK @gsmet follow up at https://github.com/quarkusio/quarkus/pull/5582

Issue #5474, opened 2019-11-14
Provide a build item that contains the module root path.


iocanel:
**Description**
Every now then when working with extensions, I need to use the path to the module root directory. This information is currently not available and one can only use workarounds.

It would be great if there was a build item provided that contained this information.

**Implementation ideas**
The maven BuildMojo does have access to this information so does the gradle BuildTask.
From there we could pass the information forward to the QuarkusAugmentor that could produce the desired build item.

stuartwdouglas:
The issue with this is that it assumes that augmentation always happens from within the context of a Java project. One of our initial goals with Quarkus was the ability too easily do upgrades in a cloud based environment without needing to rebuild your project. This use case would involve augmenting a pre-built jar that does not necessarily still have the Java project present.

In addition to this if we provide it to the build then extensions are going to make assumptions about the structure of the project that is not always correct.

I think we need to look at the use cases here and see if there are other ways to solve this problem.

cc @emmanuelbernard @n1hility

iocanel:
So, it seems that we won't proceed with this change.
I am closing the issue for now.

PR #5478, opened 2019-11-14
Provide a ModuleBuildItem that points to the module root.

iocanel:
This is meant to fulfil #5474

### DIFF
--- a/core/creator/src/main/java/io/quarkus/creator/CuratedApplicationCreator.java
+++ b/core/creator/src/main/java/io/quarkus/creator/CuratedApplicationCreator.java
@@ -28,6 +28,7 @@ public class CuratedApplicationCreator implements AutoCloseable {
 
     private final AppModelResolver artifactResolver;
     private final AppArtifact appArtifact;
+    private final Path moduleDir;
     private final Path workDir;
     private boolean deleteTmpDir = true;
 
@@ -44,6 +45,7 @@ public class CuratedApplicationCreator implements AutoCloseable {
         this.update = builder.update;
         this.updateNumber = builder.updateNumber;
         this.localRepo = builder.localRepo;
+        this.moduleDir = builder.moduleDir;
         boolean del;
         if (builder.workDir != null) {
             del = false;
@@ -67,6 +69,15 @@ public class CuratedApplicationCreator implements AutoCloseable {
             }
         }
         this.baseName = finalName;
+    }
+
+    /**
+     * Module module directory.
+     *
+     * @return module dir
+     */
+    public Path getModuleDir() {
+        return moduleDir;
     }
 
     /**
@@ -186,6 +197,7 @@ public class CuratedApplicationCreator implements AutoCloseable {
 
         public String baseName;
         private AppArtifact appArtifact;
+        private Path moduleDir;
         private Path workDir;
         private AppModelResolver modelResolver;
 
@@ -199,6 +211,17 @@ public class CuratedApplicationCreator implements AutoCloseable {
 
         public Builder setBaseName(String baseName) {
             this.baseName = baseName;
+            return this;
+        }
+
+        /**
+         * Base directory ofr the current module.
+         * 
+         * @param dir base directory
+         * @return this AppCreator instance
+         */
+        public Builder setModuleDir(Path dir) {
+            this.moduleDir = dir;
             return this;
         }
 
--- a/core/deployment/src/main/java/io/quarkus/deployment/pkg/builditem/ModuleDirBuildItem.java
+++ b/core/deployment/src/main/java/io/quarkus/deployment/pkg/builditem/ModuleDirBuildItem.java
@@ -1,0 +1,19 @@
+package io.quarkus.deployment.pkg.builditem;
+
+import java.nio.file.Path;
+
+import io.quarkus.builder.item.SimpleBuildItem;
+
+public final class ModuleDirBuildItem extends SimpleBuildItem {
+
+    private final Path path;
+
+    public ModuleDirBuildItem(Path path) {
+        this.path = path;
+    }
+
+    public Path getPath() {
+        return this.path;
+    }
+
+}
--- a/core/creator/src/main/java/io/quarkus/creator/phase/augment/AugmentTask.java
+++ b/core/creator/src/main/java/io/quarkus/creator/phase/augment/AugmentTask.java
@@ -168,6 +168,7 @@ public class AugmentTask implements CuratedTask<AugmentOutcome> {
                 QuarkusAugmentor.Builder builder = QuarkusAugmentor.builder();
                 builder.setRoot(appClassesDir);
                 builder.setBaseName(ctx.getBaseName());
+                builder.setModuleDir(ctx.getModuleDir());
                 builder.setTargetDir(outputDir);
                 builder.setResolver(appState.getArtifactResolver());
                 builder.setEffectiveModel(appState.getEffectiveModel());
--- a/core/deployment/src/main/java/io/quarkus/deployment/QuarkusAugmentor.java
+++ b/core/deployment/src/main/java/io/quarkus/deployment/QuarkusAugmentor.java
@@ -32,6 +32,7 @@ import io.quarkus.deployment.builditem.LaunchModeBuildItem;
 import io.quarkus.deployment.builditem.LiveReloadBuildItem;
 import io.quarkus.deployment.builditem.ShutdownContextBuildItem;
 import io.quarkus.deployment.pkg.builditem.CurateOutcomeBuildItem;
+import io.quarkus.deployment.pkg.builditem.ModuleDirBuildItem;
 import io.quarkus.deployment.pkg.builditem.OutputTargetBuildItem;
 import io.quarkus.runtime.LaunchMode;
 
@@ -48,6 +49,7 @@ public class QuarkusAugmentor {
     private final Collection<Path> excludedFromIndexing;
     private final LiveReloadBuildItem liveReloadBuildItem;
     private final Properties buildSystemProperties;
+    private final Path moduleDir;
     private final Path targetDir;
     private final AppModel effectiveModel;
     private final AppModelResolver resolver;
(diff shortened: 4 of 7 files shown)


geoand:
The test failure seems related:

```
[INFO] [INFO] ------------------------------------------------------------------------
[INFO] [INFO] BUILD FAILURE
[INFO] [INFO] ------------------------------------------------------------------------
[INFO] [INFO] Total time:  4.912 s
[INFO] [INFO] Finished at: 2019-11-14T16:16:24Z
[INFO] [INFO] ------------------------------------------------------------------------
[INFO] [ERROR] Failed to execute goal io.quarkus:quarkus-maven-plugin:999-SNAPSHOT:build (default) on project kubernetes-with-application-properties: Failed to build a runnable JAR: Failed to augment application classes: Build failure: Build failed due to errors
[INFO] [ERROR] 	[error]: Build step io.quarkus.kubernetes.deployment.KubernetesProcessor#build threw an exception: java.lang.NullPointerException
[INFO] [ERROR] 	at io.quarkus.kubernetes.deployment.KubernetesProcessor.createProject(KubernetesProcessor.java:222)
[INFO] [ERROR] 	at io.quarkus.kubernetes.deployment.KubernetesProcessor.build(KubernetesProcessor.java:118)
[INFO] [ERROR] 	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
[INFO] [ERROR] 	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
[INFO] [ERROR] 	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
[INFO] [ERROR] 	at java.lang.reflect.Method.invoke(Method.java:498)
[INFO] [ERROR] 	at io.quarkus.deployment.ExtensionLoader$1.execute(ExtensionLoader.java:948)
[INFO] [ERROR] 	at io.quarkus.builder.BuildContext.run(BuildContext.java:415)
[INFO] [ERROR] 	at org.jboss.threads.ContextClassLoaderSavingRunnable.run(ContextClassLoaderSavingRunnable.java:35)
[INFO] [ERROR] 	at org.jboss.threads.EnhancedQueueExecutor.safeRun(EnhancedQueueExecutor.java:2011)
[INFO] [ERROR] 	at org.jboss.threads.EnhancedQueueExecutor$ThreadBody.doRunTask(EnhancedQueueExecutor.java:1535)
[INFO] [ERROR] 	at org.jboss.threads.EnhancedQueueExecutor$ThreadBody.run(EnhancedQueueExecutor.java:1426)
[INFO] [ERROR] 	at java.lang.Thread.run(Thread.java:748)
[INFO] [ERROR] 	at org.jboss.threads.JBossThread.run(JBossThread.java:479)
[INFO] [ERROR] -> [Help 1]
[INFO] [ERROR] 
[INFO] [ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[INFO] [ERROR] Re-run Maven using the -X switch to enable full debug logging.
[INFO] [ERROR] 
[INFO] [ERROR] For more information about the errors and possible solutions, please read the following articles:
[INFO] [ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/MojoExecutionException
[INFO] run post-build script verify.groovy
[INFO] Running post-build script: /home/vsts/work/1/s/integration-tests/kubernetes/target/it/kubernetes-with-application-properties/verify.groovy
[INFO]           kubernetes-with-application-properties/pom.xml ... FAILED (6.5 s)
[INFO]   The post-build script did not succeed. assert kubernetesYml.exists()
```

dmlloyd:
This change doesn't seem very clear to me designwise.  Are the existing archive build items somehow insufficient?  Was there an email thread about this?

iocanel:
> This change doesn't seem very clear to me designwise. Are the existing archive build items somehow insufficient? Was there an email thread about this?

I have asked about it via mailing list a while back, with no satisfying answer. If there is an alternative to get the module root, I'd gladly use it and save myself the trouble if fixing the tests :-D

dmlloyd:
I don't really understand what the "module root" is.  Do you mean the base project that is being built?  Sometimes what the user considers to be the "base project" could come from a dependency so really the main build module has to be considered to be an equal peer with all of the application archives in most respects.

iocanel:
> I don't really understand what the "module root" is. Do you mean the base project that is being built? 

By module root, I mean the module directory (the dir that contains the pom.xml for that particular module). If this is still not clear I mean the maven `project.baseDir`.

dmlloyd:
I see.  I have one more question then: what kinds of tasks are you performing on the build root which is not possible to perform generically across all the input archives?

iocanel:
I am not sure I understand the question, but I can try.
There have been two cases for me so far.

1. I needed to provide to dekorate the module directory.
2. I wanted to trigger a docker build while also needed to happen from within the module directory.

Does this answer the question?

dmlloyd:
Ah I see.  This is not what I expected, I was thinking it was something classloader related or something.

Might it not be better to make it configuration based?  This seems similar to `quarkus.application.name` in that the build system would have to propagate it and that its value would depend on the build system.  It would be nice to consolidate these things into one concept somehow.

iocanel:
Well, the module root is something fixed. It's not something that can change. And for the use cases mentioned above, it wouldn't make sense for the user to override it in some way.

It's not that we are talking for something like the output directory, which a user could possibly need to change.

dmlloyd:
That is very true.

dmlloyd:
My only concern with this patch would be misuse.  I can easily imagine an extension author using this path by mistake, and causing user builds to break because the user decided to move their packaging to a different submodule which includes the original one.

iocanel:
It shouldn't really matter. If the user moves the packaging to another submodule, then the suggested build item would point to that submodule and everyone would be happy. In fact that is one of the reasons I created the build item and didn't go with something like `new File(".")`.

dmlloyd:
Ah, sure, that makes sense.

dmlloyd (approved):

geoand:
Now you can fix the tests @iocanel 😎

core/deployment/src/main/java/io/quarkus/deployment/pkg/builditem/ModuleDirBuildItem.java
```diff
@@ -0,0 +1,19 @@
+package io.quarkus.deployment.pkg.builditem;
+
+import java.nio.file.Path;
+
+import io.quarkus.builder.item.SimpleBuildItem;
+
+public final class ModuleDirBuildItem extends SimpleBuildItem {
```
ppalaga: Please add a class level JavaDoc, something like `A BuildItem holding the base directory of the current module (<code>${basedir}</code> for Maven projects or <code>${projectDir}</code> for Gradle projects).`

core/deployment/src/main/java/io/quarkus/deployment/pkg/builditem/ModuleDirBuildItem.java
```diff
@@ -0,0 +1,19 @@
...
+
+    private final Path path;
+
+    public ModuleDirBuildItem(Path path) {
+        this.path = path;
+    }
+
+    public Path getPath() {
```
ppalaga: JavaDoc: `@returns the base directory of the current module (<code>${basedir}</code> for Maven projects or <code>${projectDir}</code> for Gradle projects).`

core/creator/src/main/java/io/quarkus/creator/CuratedApplicationCreator.java
```diff
@@ -202,6 +214,17 @@ public Builder setBaseName(String baseName) {
             return this;
         }
 
+        /**
+         * Base directory ofr the current module.
+         * 
+         * @param dir base directory
+         * @return this AppCreator instance
```
ppalaga: This method does not return an AppCreator instance

core/creator/src/main/java/io/quarkus/creator/CuratedApplicationCreator.java
```diff
@@ -202,6 +214,17 @@ public Builder setBaseName(String baseName) {
             return this;
         }
 
+        /**
+         * Base directory ofr the current module.
```
ppalaga:
```suggestion
         * Base directory of the current module (<code>${basedir}</code> for Maven projects or <code>${projectDir}</code> for Gradle projects).
```

ppalaga (changes requested):
A couple of missing JavaDoc comments suggested inline.

iocanel:
@ppalaga: These commits are also included in #5496. In that pull request there have been several discussions with @stuartwdouglas on wether we should add this change or not. At the momemnt I intend to explore alternative routes and close this pull request. 

Before I do close this, you might want to take some time and disucss your use cases, to see if they could also be worked around or if we `have to have` this change.

Thanks

stuartwdouglas:
The reason why I have concerns about this is that one of our use cases is to be able augment existing jars (e.g. to provide security updates by building an app against an updated version of the Quarkus platform). This build item would break that use case if build steps start relying on it.

ppalaga:
> These commits are also included in #5496

Thanks for the info, looking there.

> you might want to take some time and disucss your use cases

Yes, in Camel Quarkus we try to support as much as possible from Camel. In this particular case, plain Camel allows to declare XSL template files as either absolute paths or paths relative to the `baseDir`. We are compiling the templates to translets at build time so we have to resolve those relative paths there too. 

The XSL templates can also be declared as classpath resources (and we support that already) so if we are forced to ditch the support for relative files, the users could actually use classpath instead. 

@stuartwdouglas thanks for mentioning the patching use case. This probably means that we should also stop supporting `http://` and other external XSL template URIs. WDYT, @lburgazzoli ?

lburgazzoli:
> 
> @stuartwdouglas thanks for mentioning the patching use case. This probably means that we should also stop supporting `http://` and other external XSL template URIs. WDYT, @lburgazzoli ?

I've already said that we should  probably support only templates from the classpath but looks like you ignored my comment

ppalaga:
> I've already said that we should probably support only templates from the classpath but looks like you ignored my comment

Sorry, I was missing a good reason to do that. Now there is one.

So, let's conclude that we (Camel Quarkus) are fine without having a baseDir BuildItem for now. Thanks for the discussion!

iocanel:
So, it's seems that we can close this pull request then.

Issue #5370, opened 2019-11-11
Support spring-web wildcard syntax in @RequestMapping


johnaohara:
**Describe the bug**
Spring-web RequestMapping allows for wildcard definitions within paths that have a different syntax from valid jax-rs paths. 

In order to facilitate support of spring-web applications without developers having to re-define path mappings, quarkus should fully support the spring-web syntax in @RequestMapping definitions.

**Expected behavior**
e.g `@GetMapping("/wildcard/*/{name}")` should map `/wildcard/1/Quarkus`, currently returns HTTP 404, whereas `/wildcard/*/Quarkus` returns HTTP 200

`@RequestMapping("/car?/s?o?/info")` should also map urls  "/cars/shop/info" and "/cart/show/info"

**To Reproduce**
1. `git clone git@github.com:johnaohara/quarkusSpringWebWildcards.git `
2. `cd quarkusSpringWebWildcards`
3. `mvn clean test`

geoand:
@asoldano @patriot1burke @FroMage does JAX-RS not support expressions like `@PATH("/wildcard/*/{name}")`?

asoldano:
@geoand see https://docs.jboss.org/resteasy/docs/4.4.0.Final/userguide/html/Using_Path.html#_Path_and_regular_expression_mappings

geoand:
Excellent, thanks @asoldano!

geoand:
RESTEasy PR opened to address these cases. Once it's been merged and released I can update Quarkus to add some tests cases for this

geoand:
I forgot to add tests, but actually they shouldn't be needed since all relevant the tests were added in RESTEasy

PR #5482, opened 2019-11-14
Upgrade to RESTEasy 4.4.1.Final

### LINKED ISSUE #5374: ClassCastException in Spring Web when both a controller method and an advice method return ResponseEntity
When a Spring controller is something like:

```
 @GetMapping(value = "/{id}")
    public ResponseEntity<Post> getPost(@PathVariable("id") String id) {

        Post post = this.postRepository.findById(id).orElseThrow(
                () -> new PostNotFoundException(id)
        );

        return ok(post);
    }
```

and the controller advice looks like:

```
@ExceptionHandler(PostNotFoundException.class)
    public ResponseEntity notFound(PostNotFoundException ex) {
        Map<String, String> errors = new HashMap<>();
        errors.put("entity", "POST");
        errors.put("id", "" + ex.getSlug());
        errors.put("code", "not_found");
        errors.put("message", ex.getMessage());

        return status(HttpStatus.NOT_FOUND).body(errors);
    }
```

then a ClassCastException is thrown when the controller method is used with an "illegal" ID.
Something like:

```
Caused by: java.lang.ClassCastException: class java.util.HashMap cannot be cast to class org.springframework.http.ResponseEntity
```

Originally reported [here](https://github.com/quarkusio/quarkus/issues/4042#issuecomment-552443998).

asoldano:
Fixes: #5374 #5370

### DIFF
--- a/bom/runtime/pom.xml
+++ b/bom/runtime/pom.xml
@@ -16,7 +16,7 @@
     <properties>
         <aesh.version>1.11</aesh.version>
         <jandex.version>2.1.1.Final</jandex.version>
-        <resteasy.version>4.4.0.Final</resteasy.version>
+        <resteasy.version>4.4.1.Final</resteasy.version>
         <opentracing.version>0.31.0</opentracing.version>
         <opentracing-jaxrs.version>0.4.1</opentracing-jaxrs.version>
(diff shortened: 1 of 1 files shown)


gsmet (approved):

machi1990:
CI is green, let's merge this one.

PR #5483, opened 2019-11-14
Make build registration of security interceptors more flexible

geoand:
This is done by allowing the registration of arbitrary
SecurityCheck implementations.
Furthermore the use of the AnnotationStore is hidden so as
to not give the impression that it's a feature that magically
makes security annotations work

This change paves the way for #5225 and more related improvements
in the future

### DIFF
--- a/extensions/security/deployment/src/main/java/io/quarkus/security/deployment/SecurityProcessor.java
+++ b/extensions/security/deployment/src/main/java/io/quarkus/security/deployment/SecurityProcessor.java
@@ -1,23 +1,26 @@
 package io.quarkus.security.deployment;
 
-import java.lang.reflect.Method;
+import static io.quarkus.security.deployment.SecurityCheckInstantiationUtil.authenticatedSecurityCheck;
+import static io.quarkus.security.deployment.SecurityCheckInstantiationUtil.denyAllSecurityCheck;
+import static io.quarkus.security.deployment.SecurityCheckInstantiationUtil.permitAllSecurityCheck;
+import static io.quarkus.security.deployment.SecurityCheckInstantiationUtil.rolesAllowedSecurityCheck;
+
+import java.lang.reflect.Modifier;
 import java.security.Provider;
 import java.security.Security;
 import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.Collection;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.function.Function;
 
 import org.jboss.jandex.AnnotationInstance;
 import org.jboss.jandex.AnnotationTarget;
-import org.jboss.jandex.AnnotationValue;
 import org.jboss.jandex.ClassInfo;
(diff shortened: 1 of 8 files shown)


extensions/security/deployment/src/main/java/io/quarkus/security/deployment/SecurityProcessor.java
```diff
@@ -203,61 +203,106 @@ private ResultHandle paramTypes(MethodCreator ctor, List<Type> parameters) {
...
+    private Map<MethodInfo, Function<BytecodeCreator, ResultHandle>> gatherSecurityAnnotations(
+            ApplicationIndexBuildItem indexBuildItem, DotName dotName,
+            Map<MethodInfo, AnnotationInstance> alreadyCheckedMethods,
+            Function<AnnotationInstance, Function<BytecodeCreator, ResultHandle>> securityCheckInstanceCreator) {
+
+        Map<MethodInfo, Function<BytecodeCreator, ResultHandle>> result = new HashMap<>();
+
+        List<AnnotationInstance> instances = indexBuildItem.getIndex().getAnnotations(dotName);
+        // make sure we process annotations on methods first
+        for (AnnotationInstance instance : instances) {
```
michalszynkiewicz: It would be nice to have some more straightforward logic, without the need of `alreadyCheckedMethods` (you need them to, e.g., not override method level `@RolesAllowed` by class level `@PermitAll`, right?)

Maybe I'm missing something but I don't see the need to modify what was here before.

michalszynkiewicz: oh, the cause for the change is probably that what I had before relied on scanning and you started coding it in a different way :)

geoand: Basically I needed a way to both collect the annotations and at the same time determine the security check of the method.
So the logic needed to change from what you had

geoand:
@michalszynkiewicz Any more comments on this PR?

michalszynkiewicz (approved):

michalszynkiewicz:
I think it all could be simplified a bit (it being also the stuff that I coded previously). 
I'll create an issue for it because I have other stuff to work on right now but maybe I'll have a chance to come back to it later. And also we will have a chance to discuss it there :)
I see nothing wrong with this change, just the logic got a tiny bit more difficult to follow. Let's merge it before we have any conflicts.

geoand:
Thanks for checking @michalszynkiewicz!

It's probably a bit more complicated because I have a use case in mind that I would like to get the fundamentals in for now (basically an extension needs to be able to specify a custom security check per method and not have it determined by the security extension itself).
Of course that isn't being used in the current code, but it will come.

michalszynkiewicz:
@geoand I think https://github.com/quarkusio/quarkus/issues/5550 should cover your case pretty well :)

geoand:
The list is gret! Thanks for opening the issue!

I'll comment on it when I have something concrete

PR #5496, opened 2019-11-14
Add support for building and deploying container to Kubernetes

### LINKED ISSUE #5493: Allow triggering container builds and deployments during the build.
**Description**
Currently, every quarkus project comes with a set of Dockerfiles for jvm and native images. Also, its trivial to have kubernetes/openshift manifests generated using the `kubernetes` extension.

To further bump the developer joy, it would be great if we could trigger a docker build and deployment from the command line, with something like:

    mvn clean package -Dquarkus.container.deploy=true

This could hands down provide the faster way to get a java application running on Kubernetes / Openshift.

iocanel:
Fixes #5493

This proof of concept introduces two new features:

- Perform docker builds as part of the build.
- Perfrom deployment to kubernetes.

# Docker builds
To enable the docker build:

    mvn clean package -Dquarkus.container.build=true

This command will trigger a docker build, right after the augmentation process is done.

It should work for both jvm and native modes.

This action requires the presence of `docker` binary in the path & also working docker environment.

# Kubernetes deployment

To trigger deployment to kubernetes, add the kubernetes extension and then:

    mvn clean package -Dquarkus.container.build=true -Dquarkus.container.deploy=true

This command will trigger a deployment right after the container build is done.

This actuion requires the presence of `oc` binary in the path & also ~/.kube/config for a working cluster.

## Integration test

Also added an integration test under `kubernetes` called `kubernetes-build-and-deploy`.

Note: The existing implementation does not cover pushing to registires, so if you try it make sure you kubernetes and docker point to the same docker daemon.


maxandersen:
just one question before i have looked in details - how do we get s2i builds to fit into this ? If it doesn't that is fine and we can work from there; i'm just wondering if there is a way I just haven't spotted ;)

iocanel:
We already generate build config and image streams. So it's just a matter of imvoking
the actual build. This should be as simple as doing an exec or just calling build from the client. Will implement it ASAP..

geoand:
@iocanel is there any chance that the new build items proposed in #5499 are useful for thjis?

core/deployment/src/main/java/io/quarkus/deployment/util/ExecUtil.java
```diff
@@ -0,0 +1,62 @@
...
+
+    /**
+     * Execute the specified command from within the specified directory.
+     * 
+     * @param directory The directory
+     * @param command The command
+     * @param args The command arguments
+     * @return true if commands where executed successfully
+     */
+    public static boolean exec(File directory, String command, String... args) {
```
geoand: I wonder if it would be better to have the command executed on another thread, like is done at :https://github.com/quarkusio/quarkus/blob/master/core/deployment/src/main/java/io/quarkus/deployment/pkg/steps/NativeImageBuildStep.java#L278

iocanel:
> @iocanel is there any chance that the new build items proposed in #5499 are useful for this?

Not something I need, but rather something I would have to use once it gets merged.

core/deployment/src/main/java/io/quarkus/deployment/util/ExecUtil.java
iocanel: I am not sure why they are pumping the output stream in a different thread.
I wouldn't mind do that same here, but I'd like to know what's the benefit.

extensions/kubernetes/deployment/src/main/java/io/quarkus/kubernetes/deployment/KubernetesConfig.java
```diff
@@ -0,0 +1,24 @@
...
+@ConfigRoot
+public class KubernetesConfig {
+
+    private static final String KUBERNETES = "kubernetes";
+    private static final String OPENSHIFT = "openshift";
+    private static final String KNATIVE = "knative";
+
+    /**
+     * The target deployment platform.
+     * Defaults to kubernetes. Can be kubernetes, openshift, knative or any combination of the above as coma separated list.
```
machi1990:
```suggestion
     * Defaults to kubernetes. Can be kubernetes, openshift, knative or any combination of the above as comma separated list.
```

extensions/kubernetes/deployment/src/main/java/io/quarkus/kubernetes/deployment/KubernetesConfig.java
```diff
@@ -0,0 +1,24 @@
...
+
+import io.quarkus.runtime.annotations.ConfigItem;
+import io.quarkus.runtime.annotations.ConfigRoot;
+
+@ConfigRoot
+public class KubernetesConfig {
+
+    private static final String KUBERNETES = "kubernetes";
+    private static final String OPENSHIFT = "openshift";
+    private static final String KNATIVE = "knative";
```
machi1990: Can we have an enum `DeploymentTarget` for these values? 

This way, accepted values will be probably displayed in the generated docs.

extensions/kubernetes/deployment/src/main/java/io/quarkus/kubernetes/deployment/KubernetesConfig.java
```diff
@@ -0,0 +1,24 @@
...
+
+    /**
+     * The target deployment platform.
+     * Defaults to kubernetes. Can be kubernetes, openshift, knative or any combination of the above as coma separated list.
+     */
+    @ConfigItem(defaultValue = KUBERNETES)
+    String deploymentTarget = KUBERNETES;
```
machi1990: Leading from the JavaDoc above, I think we can we make this a list with a default value of `KUBERNETES`

```
@ConfigItem(defaultValue = KUBERNETES)
List<DeploymentTarge> deploymentTarget;
```

core/deployment/src/main/java/io/quarkus/deployment/pkg/builditem/DeploymentResultBuildItem.java
```diff
@@ -0,0 +1,26 @@
...
+package io.quarkus.deployment.pkg.builditem;
+
+import java.util.Map;
+
+import io.quarkus.builder.item.SimpleBuildItem;
+
+public final class DeploymentResultBuildItem extends SimpleBuildItem {
+
+    private final String name;
+    private final Map<String, String> labels;
```
machi1990: Here https://github.com/quarkusio/quarkus/pull/5496/files#diff-5335230ed301829601e651decd6c5577R25 we pass `null` and `null`, so I am wondering if we need the two fields?

core/deployment/src/main/java/io/quarkus/deployment/pkg/builditem/DockerImageResultBuildItem.java
```diff
@@ -0,0 +1,30 @@
+
+package io.quarkus.deployment.pkg.builditem;
+
+import io.quarkus.builder.item.SimpleBuildItem;
+
+public final class DockerImageResultBuildItem extends SimpleBuildItem {
+
+    private final String imageId;
+    private final String repository;
+    private final String tag;
```
machi1990: Do we need these fields? They seem to be initialised to `null` all the time 
- https://github.com/quarkusio/quarkus/pull/5496/files#diff-807046a1a852a9c1d26116f35cf9ef16R35
- https://github.com/quarkusio/quarkus/pull/5496/files#diff-807046a1a852a9c1d26116f35cf9ef16R51

core/deployment/src/main/java/io/quarkus/deployment/pkg/builditem/DeploymentResultBuildItem.java
iocanel: The intention is to collect the name of the deployment and its labels and expose them as a build item, for anyone interested. This is not implemented yet.

core/deployment/src/main/java/io/quarkus/deployment/pkg/builditem/DockerImageResultBuildItem.java
iocanel: Same as above.

core/deployment/src/main/java/io/quarkus/deployment/pkg/builditem/DeploymentResultBuildItem.java
machi1990: Thanks, for the explanation. I suppose I reviewed a little too early.

extensions/kubernetes/deployment/src/main/java/io/quarkus/kubernetes/deployment/KubernetesConfig.java
gsmet: +1 on that one.

geoand: Good to know about the docs

core/creator/src/main/java/io/quarkus/creator/phase/augment/AugmentOutcome.java
```diff
@@ -16,12 +17,14 @@
     private final List<ArtifactResultBuildItem> packageOutput;
     private final JarBuildItem jar;
     private final NativeImageBuildItem nativeImage;
+    private final DockerImageResultBuildItem dockerImage;
```
stuartwdouglas: I think this should be in an extension (i.e. there should be no docker stuff in core). Native will likely be moved out to it own extension eventually as well. 

In terms of this particular class the image can be represented by a ArtifactResultBuildItem with type container-image.

core/creator/src/main/java/io/quarkus/creator/phase/augment/AugmentOutcome.java
```diff
@@ -38,4 +41,9 @@ public JarBuildItem getJar() {
     public NativeImageBuildItem getNativeImage() {
         return nativeImage;
     }
+
+    public DockerImageResultBuildItem getDockerimage() {
```
stuartwdouglas: I don't think we should be referring to docker explicitly. It should be a generic container image extension.

core/creator/src/main/java/io/quarkus/creator/phase/augment/AugmentTask.java
```diff
@@ -177,14 +180,17 @@ public AugmentOutcome run(CurateOutcome appState, CuratedApplicationCreator ctx)
                 builder.addFinal(BytecodeTransformerBuildItem.class)
                         .addFinal(ApplicationArchivesBuildItem.class)
                         .addFinal(MainClassBuildItem.class)
-                        .addFinal(ArtifactResultBuildItem.class);
+                        .addFinal(ArtifactResultBuildItem.class)
+                        .addFinal(DockerImageResultBuildItem.class)
+                        .addFinal(DeploymentResultBuildItem.class); //This should probably go to a different class. io.quarkus.deployer.DeployerTask, maybe?
```
stuartwdouglas: We need to have a bit more of a think about how to handle deployment, we certainly don't want to deploy after every build.

core/creator/src/main/java/io/quarkus/creator/phase/augment/AugmentTask.java
```diff
@@ -177,14 +180,17 @@ public AugmentOutcome run(CurateOutcome appState, CuratedApplicationCreator ctx)
                 builder.addFinal(BytecodeTransformerBuildItem.class)
                         .addFinal(ApplicationArchivesBuildItem.class)
                         .addFinal(MainClassBuildItem.class)
-                        .addFinal(ArtifactResultBuildItem.class);
+                        .addFinal(ArtifactResultBuildItem.class)
+                        .addFinal(DockerImageResultBuildItem.class)
```
stuartwdouglas: This should not be needed as ArtifactResultBuildItem is intended to be the generic output type when extensions add additional outputs.

core/deployment/src/main/java/io/quarkus/deployment/pkg/builditem/DockerBuild.java
```diff
@@ -0,0 +1,21 @@
...
+
+    private final ContainerConfig containerConfig;
+
+    DockerBuild(ContainerConfig containerConfig) {
+        this.containerConfig = containerConfig;
+    }
+
+    @Override
+    public boolean getAsBoolean() {
+        return containerConfig.build && ExecUtil.exec("docker", "version");
```
stuartwdouglas: This should maybe re-use io.quarkus.deployment.pkg.PackageConfig#type, and add docker-jvm and docker-native as possible output targets. 

This was the original intent anyway (and is why the package type is not an enum).

core/deployment/src/main/java/io/quarkus/deployment/pkg/steps/DockerBuildStep.java
```diff
@@ -0,0 +1,53 @@
...
+
+    private static final Logger log = Logger.getLogger(DockerBuildStep.class);
+
+    @BuildStep(onlyIf = DockerBuild.class, onlyIfNot = NativeBuild.class)
+    public DockerImageResultBuildItem dockerBuildFromJar(ApplicationInfoBuildItem app, ModuleDirBuildItem moduledir,
+            Optional<DockerImageBuildItem> dockerImage,
+            JarBuildItem artifact) {
+        log.info("Building docker image for jar.");
+        ExecUtil.exec(moduledir.getPath().toFile(),
+                "docker", "build",
```
stuartwdouglas: We need to make this configurable, so we are not tied to docker. We also should provide the option of using some default docker files if there is none in the project.

core/deployment/src/main/java/io/quarkus/deployment/util/ExecUtil.java
stuartwdouglas: It is done that way because we were filtering the error stream and enhancing error messages. If you are dealing with two streams you need multiple threads.

core/deployment/src/main/java/io/quarkus/deployment/util/ExecUtil.java
```diff
@@ -0,0 +1,62 @@
...
+            try (InputStreamReader isr = new InputStreamReader(process.getInputStream());
+                    BufferedReader reader = new BufferedReader(isr)) {
+
+                for (String line = reader.readLine(); line != null; line = reader.readLine()) {
+                    System.out.println(line);
+                }
+                process.waitFor();
+            }
+        } catch (IOException e) {
+            return false;
```
stuartwdouglas: These should be wrapped in a runtime exception with some more details and re-thrown

core/deployment/src/main/java/io/quarkus/deployment/QuarkusAugmentor.java
```diff
@@ -64,6 +66,7 @@
         this.excludedFromIndexing = builder.excludedFromIndexing;
         this.liveReloadBuildItem = builder.liveReloadState;
         this.buildSystemProperties = builder.buildSystemProperties;
+        this.moduleDir = builder.moduleDir;
```
stuartwdouglas: We really don't want this, as it makes assumptions that the augment process only happens from within the scope of a project, which may not always be the case in the future.

core/deployment/src/main/java/io/quarkus/deployment/pkg/steps/DockerBuildStep.java
```diff
@@ -0,0 +1,53 @@
...
+                "-f",
+                moduledir.getPath().resolve("src").resolve("main").resolve("docker").resolve("Dockerfile.jvm").toAbsolutePath()
```
stuartwdouglas: The dockerfile should be a config option.

core/deployment/src/main/java/io/quarkus/deployment/pkg/steps/DockerBuildStep.java
```diff
@@ -0,0 +1,53 @@
...
+                        .toString(),
+                "-t", dockerImage.map(d -> d.getImage()).orElse(app.getName() + ":" + app.getVersion()),
```
stuartwdouglas: This should all be configurable.

stuartwdouglas (changes requested):

core/creator/src/main/java/io/quarkus/creator/phase/augment/AugmentTask.java
iocanel: Of course, in the current POC deployments are only triggered when using `-Dquarkus.container.deployment=true`.

core/deployment/src/main/java/io/quarkus/deployment/QuarkusAugmentor.java
stuartwdouglas: Don't you just need the target directory rather than the whole project?

If you are doing this to get the dockerfile then I think it would actually make sense to move the dockerfile to src/main/resources so it ends up in the final archive. This means that it would still be possible to re-run augmentation without the project being present.

core/deployment/src/main/java/io/quarkus/deployment/util/ExecUtil.java
geoand: Thanks for the explanation

core/deployment/src/main/java/io/quarkus/deployment/QuarkusAugmentor.java
iocanel: No, target directory won't do.
Also, having the path to dockerfiles won't do either (although it would be nice to have).

The quarkus dockerfiles assume that the build is executed from within the module dir. Why?

The following piece of code is taken from `Dockerfile.jvm`:

    COPY target/lib/* /deployments/lib/
    COPY target/*-runner.jar /deployments/app.jar

The instructions above only make sense from within the module dir. If the build is executed with any other path, it will just fail.

So, if we want to be able to trigger a docker build, we do need to know this directory.

Now, a potential alternative would be to modify the dockerfiles on the fly, to remove any reference to `target/` and then just use the `targetDir` for the build, which IMHO is more hacky than just adding the `moduleDir` to the Augmentor.

stuartwdouglas: The problem is that there might not be a module dir, we want to be able to augment standalone jar files as well. We are likely going to need to modify our Dockerfile for this use case.

iocanel: Modify the Dockerfiles provided by quarkus will result in ugly Dockerfiles. I will try and see if I can modify them on demand, right before the build. I'll update the pull request shortly

stuartwdouglas: I mean't that we would probably need to look at producing different docker files in our quickstarts. Possibly we could even include the generic dockerfile in the extension and use it so the user does not have to do anything at all, and give the user the ability to customise it. 

This probably needs to be discussed with a wider audience.

iocanel: I'll start working towards this direction.

core/creator/src/main/java/io/quarkus/creator/phase/augment/AugmentTask.java
iocanel: It doesn't feel that ArtfiactResultBuildItem is suitable for representing a docker image.
It's something more tailored around physical files.

core/deployment/src/main/java/io/quarkus/deployment/pkg/builditem/DockerBuild.java
iocanel: In terms of code it doesn't really make any real difference (doesn't make code cleaner, smarter etc). So, the only difference that it makes is the actual way to configure it.

So, with your suggestion, the user would have to:

    mvn clean package -Dquarkus.package.type=docker-jvm

with the existing code, the user woruld have to use:

    mvn clean package -Dquarkus.container.build=true


Of course, this is something subjective, however the existing approach rhymes better with:

     mvn clean package -Dquarkus.container.deploy=true

So, I'd like to suggest reconsidering this point.

extensions/kubernetes/deployment/src/main/java/io/quarkus/kubernetes/deployment/KubernetesConfig.java
iocanel: Done

iocanel:
@stuartwdouglas: I think that most of your concerns have been addressed. Is there anything else you would like me to check?

core/deployment/src/main/java/io/quarkus/deployment/util/ExecUtil.java
```diff
@@ -0,0 +1,120 @@
...
+import java.io.File;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.util.function.Function;
+
+public class ExecUtil {
+
+    private static final Function<InputStream, Runnable> PRINT_OUTPUT = i -> new PrintOutput(i);
+    private static final Function<InputStream, Runnable> SILENT = i -> new Runnable() {
```
stuartwdouglas: This will fail if the process produces a lot of output. You need to consume the output otherwise the process will hang once the output buffer fill up (output buffer size is system dependent).

extensions/docker/deployment/src/main/java/io/quarkus/docker/deployment/DockerBuildStep.java
```diff
@@ -0,0 +1,123 @@
...
+                    try (InputStreamReader isr = new InputStreamReader(t);
+                            BufferedReader reader = new BufferedReader(isr)) {
+
+                        for (String line = reader.readLine(); line != null; line = reader.readLine()) {
+                            if (line.startsWith("Succesfully built")) {
+                                String[] parts = line.split(" ");
+                                if (parts.length == 3)
+                                    id.set(parts[2]);
+                            }
+                            System.out.println(line);
```
stuartwdouglas: This should be removed, or changed to use a logger.

integration-tests/kubernetes/src/it/kubernetes-build-and-deploy/src/main/docker/Dockerfile.jvm
```diff
@@ -0,0 +1,22 @@
+####
+# This Dockerfile is used in order to build a container that runs the Quarkus application in JVM mode
```
stuartwdouglas: Why are these Dockerfiles in the test? I thought it was extracting the Dockferfile from the extension?

extensions/docker/deployment/src/main/java/io/quarkus/docker/deployment/DockerBuildStep.java
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    @BuildStep(onlyIf = { DockerBuild.class, NativeBuild.class })
+    public ContainerImageResultBuildItem dockerBuildFromNativeImage(ApplicationInfoBuildItem app,
+            OutputTargetBuildItem out,
+            Optional<ContainerImageBuildItem> dockerImage,
+            NativeImageBuildItem nativeImage) {
+        log.info("Building docker image for native image.");
+        String image = dockerImage.map(d -> d.getImage()).orElse(app.getName() + ":" + app.getVersion() + "-native");
+        Path dockerFile = extractDockerfile(DOCKERFILE_NATIVE);
```
stuartwdouglas: Not a priority, but it would be nice to allow the user to package a dockerfile in their app and use that to provide a way to customise it.

integration-tests/kubernetes/src/it/kubernetes-build-and-deploy/src/main/resources/META-INF/resources/index.html
```diff
@@ -0,0 +1,152 @@
+<!DOCTYPE html>
+<html lang="en">
```
stuartwdouglas: Do we really need the full index.html in the tests?

stuartwdouglas (dismissed):
I commented on a few minor things, but other than that I am pretty happy with the implementation.

The main thing that is missing is some documentation, and there are probably also a few questions around deployment. At the moment the build process is just a build, but once we start making the build also do deployment it is a bit of a conceptual change that we might want to talk about a bit more. 

In particular the concept of a deployment action will likely be applicable in other places as well (e.g. AWS Lambda) so we might want to have some kinda of a more general deployment concept. I am going to start a conversation about deployment so will hopefully have an answer soon (If I don't have an answer soon we might just merge anyway but document that it might change).

I am also not sure if we want to call the extension 'docker', as we will probably want to support other tools such as Buildah as well. Maybe 'container-image' but that does not seem great either.

cc @emmanuelbernard @n1hility

emmanuelbernard:
yes, container is probably better than docker  in the extension name

integration-tests/kubernetes/src/it/kubernetes-build-and-deploy/src/main/resources/META-INF/resources/index.html
iocanel: No we don't I'll remove them.

extensions/docker/deployment/src/main/java/io/quarkus/docker/deployment/DockerBuildStep.java
iocanel: Agree, but this needs discussion...

My initial intention was to use the provided Dockerfiles. The user would be able to customize them.

However, in order to use them I had to introduce the `ModuleRootBuildItem` so the container extension can read the provided Dockerfiles. In earlier comments it was suggested that we want to avoid using something like the `ModuleRootBuildItem`.

integration-tests/kubernetes/src/it/kubernetes-build-and-deploy/src/main/docker/Dockerfile.jvm
iocanel: They are not needed. I'll ditch them.

core/deployment/src/main/java/io/quarkus/deployment/util/ExecUtil.java
iocanel: Thanks for the advice. Will apply that soon.

extensions/docker/deployment/src/main/java/io/quarkus/docker/deployment/DockerBuildStep.java
iocanel: Changed to logger.

iocanel:
@stuartwdouglas:

Applied feedback:
- renamed `docker` extension to `container`.
- ditched unneeded resources from tests.
- fixed ExecUtil (not always consuming the output).
- added docs.

The only thing that's left at this point is allowing the user to `bring your own Dockerfiles` for the actual build. If we have a suggestion on how to pickup user Dockerfiles, I'll add that two.

extensions/docker/deployment/src/main/java/io/quarkus/docker/deployment/DockerBuildStep.java
stuartwdouglas: One option here would be to put the dockerfiles in src/main/resources, so they can be read out of the class path.

Serkan80:
I would recommend a simpler command line for making deployments:

Build & run as a docker/container image:
`mvn quarkus:container-deploy -Dmode=jvm|native`  

Redeploy (stops & kills previous one):
mvn quarkus:container-redeploy 

Optional params:
mvn quarkus:container-redeploy  -Dkill=false

and for K8s:
`mvn quarkus:container-deploy-k8s -Dmode=jvm|native` 

optional params:
`mvn quarkus:container-deploy-k8s -Dreplicas=2 -Dnamespace=dev` 

This one also makes sure that the docker image is built (from the current project) when it's missing (reading it from the k8s-deployment.yml file).

maxandersen:
wanted to try this one out today but it does not build for me nor in azure/ci.

```
[ERROR] [ERROR] Some problems were encountered while processing the POMs:
[ERROR] Child module /Users/max/code/work/quarkus/quarkus/extensions/container of /Users/max/code/work/quarkus/quarkus/extensions/pom.xml does not exist @
 @
[ERROR] The build could not read 1 project -> [Help 1]
[ERROR]
[ERROR]   The project io.quarkus:quarkus-extensions-parent:999-SNAPSHOT (/Users/max/code/work/quarkus/quarkus/extensions/pom.xml) has 1 error
[ERROR]     Child module /Users/max/code/work/quarkus/quarkus/extensions/container of /Users/max/code/work/quarkus/quarkus/extensions/pom.xml does not exist
```

@iocanel got a fix ? :)

maxandersen:
yaay - its building now .... let the fun begin ;)

iocanel:
There are still some references to the old `docker` artifacts and thus CI fails. 
Most proabably its going to fail for you too @maxandersen.

Let me wipe my local repo, so that I can reproduce and fix.

maxandersen:
yes, it just failed..i'll hold up ;)

iocanel:
> yes, it just failed..i'll hold up ;)

Made 2 quickfixes. Wiped repo! Rebuilt! It should be ok now @maxandersen

maxandersen:
it built! 3rd time was the charm...

cmoulliard:
> This actuion requires the presence of `oc` binary

Why `oc` and `not oc or kubectl` ? @iocanel

cmoulliard:
> a deployment

Do you deploy as resources a `Deployment`, `Service` and `Route/Ingress` ? @iocanel

iocanel:
> > This actuion requires the presence of `oc` binary
> 
> Why `oc` and `not oc or kubectl` ? @iocanel

To also support deploying to openshift (not sure if kubectl supports openshift resources, never tried tbh).
 
It doesn't really matter, as this will soon be replaced by a direct call to the kubernetes client.

iocanel:
> > a deployment
> 
> Do you deploy as resources a `Deployment`, `Service` and `Route/Ingress` ? @iocanel

Yeah, we the deployment target is kubernetes we use these Deployment, Service, Ingress.
When, the target is openshift we use DeploymentConfig, Service and Route.
And when the target is knative we just use Service  (from the serving api).

cmoulliard:
> To also support deploying to openshift

To support kubernetes or Openshift deployment, then creates a `Deployment` resource and not `ImageStream/DeploymentConfig` and uses JIB to build the container image to avoid to rely on s2i and openshift buildconfig ;-)

iocanel:
> > To also support deploying to openshift
> 
> To support kubernetes or Openshift deployment, then creates a `Deployment` resource and not `ImageStream/DeploymentConfig` and uses JIB to build the container image to avoid to rely on s2i and openshift buildconfig ;-)

One can still select kubernetes as a deployment target and deploy to Openshift. This will result in creating Deployment/Service/Ingrees on openshift.

geoand:
It seems like the PR needs to be rebased onto master and have some conflicts resolved

geoand:
For the record, @iocanel and I are trying to see how we can consolidate this and the Jib [PR](https://github.com/quarkusio/quarkus/pull/6980) to provide a unified and friction-less experience

geoand:
@iocanel seems like there is a conflict. Do you want to resolve it in your next push?

rsvoboda:
@geoand  / @gsmet is this something for 1.3.0 milestone ?

geoand:
@rsvoboda not really, we don't absolutely need it in a milestone, but we definitely want it in for the final 1.3. @iocanel is putting the finishing touches on this (after him and I have been going back on forth on it the last week) so it should be ready real soon, but I don't think it needs to be rushed for `Alpha2`

maxandersen:
its part of the required ones for product so should be in for the final one yes for sure.

@iocanel @geoand ping me when you think it’s a good time to try it out and give feedback and i'll get to it asap!

geoand:
I'll let Ioannis comment since he has been working on it the past couple of days

iocanel:
@maxandersen: Just a heads up: I just pushed s2i binary build support for this pull request. I want to make sure that ci passes and that documentation is in place and then I'll ping you so that you can have a look.

To summarize what went in this pull request the last couple of weeks:

1. jib support added.
2. s2i binary build support added (for non-native builds).
3. eliminated the need for `docker`, `oc` and `kubectl` binaries.

Things that need to be done:
1. documentation.
2. review of the tirgger properiy names.

maxandersen:
excellent - definitely look forward to try it out ;)

iocanel:
FWIW, the only changes to the core are:

1. Created `DeploymentResultBuildItem`.
2. `DeploymentResultBuildItem` added as a final step in `QuarkusAugmentor`.

iocanel:
@maxandersen: I think that you can have a look now. There are still some CI issues, though they seem unrelated to the pull request. Ping me if you need anything.

maxandersen:
one feedback; when running `❯ mvn clean package -Dquarkus.container.build=true -Dquarkus.kubernetes.deploy=true -DskipTests` it does a lot of stuff (at least it prints out its building and deploying but no kube/openshift context info is printed out leaving me in a funky state of ...where did it just deploy this to. 

can we print out context name and optimally something about url to cluster / the service(if any) urls deployed (not sure if we can get that info easily but would help users)

maxandersen:
okey - good news: it *seem* to be doing the right thing ....bad news: it always ends up in my openshift cluster with errimagepull ....even with s2i.enabled=true ... pretty sure i'm missing some step so lets try and catch up tomorrow on this.

maxandersen:
a few other comments (which might just be doc tasks):

* how does one do the opposite of deploy ...how to clean up ?
* routes/ingress does not seem to be setup by default - should it ? if not - could we print out the command to enable/find it
* whats the minimal steps to use this with s2i on openshift with proper productized image ?

maxandersen:
new issue this morning:

```
ERROR] Failed to execute goal io.quarkus:quarkus-maven-plugin:999-SNAPSHOT:build (default) on project kubernetes-quickstart: Failed to build quarkus application: io.quarkus.builder.BuildException: Build failure: Build failed due to errors
[ERROR] 	[error]: Build step io.quarkus.kubernetes.deployment.KubernetesDeployer#deploy threw an exception: io.dekorate.deps.kubernetes.client.KubernetesClientException: Operation: [get]  for kind: [ServiceAccount]  with name: [manderse-kubernetes-quickstart]  in namespace: [default]  failed.
[ERROR] 	at io.dekorate.deps.kubernetes.client.KubernetesClientException.launderThrowable(KubernetesClientException.java:64)
[ERROR] 	at io.dekorate.deps.kubernetes.client.KubernetesClientException.launderThrowable(KubernetesClientException.java:72)
[ERROR] 	at io.dekorate.deps.kubernetes.client.dsl.base.BaseOperation.getMandatory(BaseOperation.java:225)
[ERROR] 	at io.dekorate.deps.kubernetes.client.dsl.base.BaseOperation.get(BaseOperation.java:168)
[ERROR] 	at io.dekorate.deps.kubernetes.client.handlers.ServiceAccountHandler.reload(ServiceAccountHandler.java:63)
[ERROR] 	at io.dekorate.deps.kubernetes.client.handlers.ServiceAccountHandler.reload(ServiceAccountHandler.java:39)
```

this is on a brand new cluster; but I think the issue here is that I'm now using:
`export KUBECONFIG=~/.kube/config:~/.kube/kubeconfig` i.e. kubeconfig with multiple elements - does the client support that ?

iocanel:
>  it always ends up in my openshift cluster with errimagepull ....even with s2i.enabled=true ... pretty sure i'm missing some step so lets try and catch up tomorrow on this.

I think that you need to make sure that `quarkus-container-image-s2i` is added to the dependencies.

iocanel:
> I think the issue here is that I'm now using:
> `export KUBECONFIG=~/.kube/config:~/.kube/kubeconfig` i.e. kubeconfig with multiple elements - does the client support that ?
 
Yeah, this is supported. 
Can you please provide the full stack trace?

maxandersen:
> > I think the issue here is that I'm now using:
> > `export KUBECONFIG=~/.kube/config:~/.kube/kubeconfig` i.e. kubeconfig with multiple elements - does the client support that ?
> 
> Yeah, this is supported.
> Can you please provide the full stack trace

turned out it was because the combined config's had overlapping names/credentials - when cleaned it up those errors disappeared.

I did note though that the errors one get when not logged in aren't actually helping a noob user to know its a bad credential issue.

maxandersen:
> I think that you need to make sure that `quarkus-container-image-s2i` is added to the dependencies.

when I add that I get this:

```
[ERROR] Failed to execute goal io.quarkus:quarkus-maven-plugin:999-SNAPSHOT:build (default) on project kubernetes-quickstart: Failed to build quarkus application: io.quarkus.builder.BuildException: Build failure: Build failed due to errors
[ERROR] 	[error]: Build step io.quarkus.container.image.s2i.deployment.S2iProcessor#s2iBuildFromJar threw an exception: java.lang.IllegalStateException: Could not find kubernetes/openshift.yml
[ERROR] 	at io.quarkus.container.image.s2i.deployment.S2iProcessor.lambda$s2iBuildFromJar$2(S2iProcessor.java:112)
[ERROR] 	at java.util.Optional.orElseThrow(Optional.java:290)
[ERROR] 	at io.quarkus.container.image.s2i.deployment.S2iProcessor.s2iBuildFromJar(S2iProcessor.java:112)
[ERROR] 	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
[ERROR] 	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
[ERROR] 	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
[ERROR] 	at java.lang.reflect.Method.invoke(Method.java:498)
[ERROR] 	at io.quarkus.deployment.ExtensionLoader$2.execute(ExtensionLoader.java:929)
[ERROR] 	at io.quarkus.builder.BuildContext.run(BuildContext.java:279)
[ERROR] 	at org.jboss.threads.ContextClassLoaderSavingRunnable.run(ContextClassLoaderSavingRunnable.java:35)
[ERROR] 	at org.jboss.threads.EnhancedQueueExecutor.safeRun(EnhancedQueueExecutor.java:2027)
[ERROR] 	at org.jboss.threads.EnhancedQueueExecutor$ThreadBody.doRunTask(EnhancedQueueExecutor.java:1551)
[ERROR] 	at org.jboss.threads.EnhancedQueueExecutor$ThreadBody.run(EnhancedQueueExecutor.java:1442)
[ERROR] 	at java.lang.Thread.run(Thread.java:748)
[ERROR] 	at org.jboss.threads.JBossThread.run(JBossThread.java:479)
[ERROR] -> [Help 1]
```

so it makes a difference just not a working one ;)

docs/src/main/asciidoc/container-image.adoc
```diff
@@ -0,0 +1,136 @@
...
+* docker
+* s2i
+
+== Building
+
+To build a container image for your project, you just need to specify `quarkus.container-image.execution=build` either to the `application.properties` or as a system property.
+
+[source, subs=attributes+]
+----
+mvn clean package -Dquarkus.container-image.exection=build
```
maxandersen: before this command will do anything I would need to at least add one of the extensions found below, right? maybe swap the sections or at least point out you shouldn;t run that command without first adding the deeondency.

maxandersen: also should be `.execution=build`

maxandersen (changes requested):

docs/src/main/asciidoc/kubernetes.adoc
```diff
@@ -560,3 +561,28 @@ The generated service can be customized using the following properties:
...
+
+[source, subs=attributes+]
+----
+mvn clean package -Dquarkus.container.build=true -Dquarkus.kubernetes.deploy=true
+----
+
+=== Building
+
+At the moment this is only possible, using `docker`.
+The dockerfiles used are internal (Dockerfiles found under `src/main/docker` are NOT used).
```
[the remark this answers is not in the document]
geoand: Actually this documentation is old. When using `quarkus-container-image-docker`, the Dockerfile that is used is by default `src/main/docker/Dockerfile.jvm`. See `src/main/docker/Dockerfile.jvm`

docs/src/main/asciidoc/kubernetes.adoc
```diff
@@ -560,3 +561,28 @@ The generated service can be customized using the following properties:
...
 |====
+
+== Deployment
+
+To trigger building and deploying a container image you need to enable `quarkus.container.build` and `quarkus.container.deploy` flags respectively.
+This can be easily done from the command line:
+
+[source, subs=attributes+]
+----
+mvn clean package -Dquarkus.container.build=true -Dquarkus.kubernetes.deploy=true
```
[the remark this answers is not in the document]
geoand: It's kind of tricky to debug when doing both a package and running the tests. So I usually do one or the other to figure out problems easier.

`mvnDebug package -DskipTests` will debug the creation of the actual prod jar.

`mvn test -Dmaven.surefire.debug` will debug the test launching process, which also includes a quarkus build (just not a jar build)

core/deployment/src/main/java/io/quarkus/deployment/util/ExecUtil.java
iocanel: This has been fixed.

extensions/kubernetes/deployment/src/main/java/io/quarkus/kubernetes/deployment/KubernetesConfig.java
iocanel: Suggestion has been adopted.

core/creator/src/main/java/io/quarkus/creator/phase/augment/AugmentOutcome.java
iocanel: code has been cleaned from Docker / Container specific stuff.

core/creator/src/main/java/io/quarkus/creator/phase/augment/AugmentTask.java
iocanel: DockerImageResultItem has been removed.

core/deployment/src/main/java/io/quarkus/deployment/QuarkusAugmentor.java
iocanel: This was completely re-written, we are now using the Dockerfiles under `src/main/docker` or from any user provided location.

Detection of `src/main/docker`is done by recursively navigating upwards from the output folder, till `src/main/docker` exists.

docs/src/main/asciidoc/container-image.adoc
```diff
@@ -0,0 +1,136 @@
...
+    <dependency>
+      <groupId>io.quarkus</groupId>
+      <artifactId>quarkus-container-image-docker</artifactId>
+    </dependency>
+----
+
+==== S2i
+
+The extension `quarkus-container-image-s2i` is using s2i binary builds in order to perform container builds inside the Openshift cluster.
+The idea behind the binary build is that you just upload the the artifact and its dependencies to the cluster and during the build they will be merged to a builder image (defaults to `fabric8/s2i-java`).
```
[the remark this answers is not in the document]
iocanel: I am not  aware of a better alternative.

docs/src/main/asciidoc/kubernetes.adoc
```diff
@@ -560,3 +561,28 @@ The generated service can be customized using the following properties:
...
+----
+
+=== Building
+
+At the moment this is only possible, using `docker`.
```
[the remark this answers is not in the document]
iocanel: Hmm this is outdated.

docs/src/main/asciidoc/container-image.adoc
iocanel: Swapped sections.

docs/src/main/asciidoc/kubernetes.adoc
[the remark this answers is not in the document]
iocanel: Removed outdated docs.

maxandersen (approved):
Base idea is good; works where tested. lets move forward and fix whatever we find incrementally.

PR #5499, opened 2019-11-14
Add the ability to control the target dir via config

stuartwdouglas:


core/deployment/src/main/java/io/quarkus/deployment/pkg/PackageConfig.java
```diff
@@ -48,4 +49,17 @@
...
+     * systems target directory.
+     */
+    @ConfigItem
+    public Optional<String> outputDirectory;
+
+    /**
+     * The name of the final artifact
+     */
+    @ConfigItem
+    public Optional<String> outputName;
```
gsmet: I'm a bit concerned about the consequences of this on the ``Dockerfile``s we provide.

I'm more and more thinking that having some templates for these files is probably not a good idea and they shoud be generated by the build.

Do my interrogations make sense?

gsmet (changes requested):
Blocking until we figure out if my interrogations are dumb or not.

core/deployment/src/main/java/io/quarkus/deployment/pkg/PackageConfig.java
stuartwdouglas: I like the idea of generating them in the build, I had already been thinking about it in the context of https://github.com/quarkusio/quarkus/pull/5496

stuartwdouglas: I don't think this will actually break the dockerfile, as they seem to be based on suffix, this only changes the prefix in the same way as build.finalName in maven.

gsmet (approved):

gsmet:
I rebased and force pushed. Let's wait for CI.

Issue #4218, opened 2019-09-26
Upgrade to GraalVM 19.3.0 / Java 8 edition


gsmet:
I created this issue to gather notes about the next GraalVM upgrade.

Per Codrut Stancu from the GraalVM team on Zulip:

> ~~FYI this changeset https://github.com/oracle/graal/commit/fa58e36e0b8337b5e9d9dcc136d60095da3d90a8 introduces a substitution for Field.getAnnotatedType() which conflicts with the Quarkus substitution in https://github.com/quarkusio/quarkus/blob/master/core/runtime/src/main/java/io/quarkus/runtime/graal/FieldReplacement.java. I think the Quarkus substitution for Field.getAnnotatedType() can be removed with the next GraalVM release. In fact, I think that the entire FieldReplacement.java can be removed.~~ DONE!

Per Codrut again

> FYI additional changes that Quarkus will need for GraalVM 19.3:
> - Update Netty version to include https://github.com/netty/netty/pull/9631
> - Additional command-line options --initialize-at-run->time=org.wildfly.common.net.CidrAddress,org.wildfly.common.net.Inet. This is needed due to a new restriction that prohibits the presence of java.net.Inet4Address in the image heap.

Note that this means an upgrade of Netty to `4.1.43.Final`. Or another option is to port this patch https://github.com/netty/netty/pull/9631/files to Quarkus via usage of our reflection registration API but I'm not sure if we have `allowUnsafeAccess` support right now.

loicmathieu:
This release should also add support for Java 11 on GraalVM. 
So this means we also need to add test for this combination in the CI :  Native on 11.

gsmet:
Yeah or drop Java 8 and only test Java 11. /cc @dmlloyd

loicmathieu:
> Yeah or drop Java 8 and only test Java 11

dmlloyd:
Hopefully people aren't using Java 8 for new projects :)

loicmathieu:
> Hopefully people aren't using Java 8 for new projects :)

You will be very surprised ... or disapointed :)

gsmet:
Reopening, we haven't upgraded yet.

cstancu:
Additional patch required for GraalVM 19.3.0:
```
[...]

PR #5500, opened 2019-11-14
Introduce UnsafeAccessedFieldBuildItem

gwenneg:
This PR introduces the GraalVM `allowUnsafeAccess` support for fields through the Quarkus reflection registration API.

It was created to fix one the GraalVM `19.3.0` requirements mentioned in #4218. It makes the Netty upgrade to `4.1.43.Final` optional.

cc @gsmet, @cescoffier

### DIFF
--- a/core/deployment/src/main/java/io/quarkus/deployment/steps/NativeImageAutoFeatureStep.java
+++ b/core/deployment/src/main/java/io/quarkus/deployment/steps/NativeImageAutoFeatureStep.java
@@ -32,6 +32,7 @@ import io.quarkus.deployment.builditem.nativeimage.ReflectiveMethodBuildItem;
 import io.quarkus.deployment.builditem.nativeimage.RuntimeInitializedClassBuildItem;
 import io.quarkus.deployment.builditem.nativeimage.RuntimeReinitializedClassBuildItem;
 import io.quarkus.deployment.builditem.nativeimage.ServiceProviderBuildItem;
+import io.quarkus.deployment.builditem.nativeimage.UnsafeAccessedFieldBuildItem;
 import io.quarkus.gizmo.CatchBlockCreator;
 import io.quarkus.gizmo.ClassCreator;
 import io.quarkus.gizmo.ClassOutput;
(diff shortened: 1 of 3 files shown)


core/deployment/src/main/java/io/quarkus/deployment/steps/NativeImageAutoFeatureStep.java
```diff
@@ -80,6 +82,22 @@ public void write(String s, byte[] bytes) {
...
+                    ofMethod(Class.class, "forName", Class.class, String.class),
+                    tc.load(unsafeAccessedField.getDeclaringClass()));
+            ResultHandle fieldHandle = tc.invokeVirtualMethod(
+                    ofMethod(Class.class, "getDeclaredField", Field.class, String.class), declaringClassHandle,
+                    tc.load(unsafeAccessedField.getFieldName()));
+            tc.invokeInterfaceMethod(
+                    ofMethod(Feature.BeforeAnalysisAccess.class, "registerAsUnsafeAccessed", void.class, Field.class),
+                    beforeAnalysisParam, fieldHandle);
+            CatchBlockCreator cc = tc.addCatch(Throwable.class);
+            cc.invokeVirtualMethod(ofMethod(Throwable.class, "printStackTrace", void.class), cc.getCaughtException());
```
gsmet: Hmmm, we don't use a proper logger here?

gsmet (changes requested):
I added a small question.

Can this be safely merged now (once the question is resolved) or do we need to wait for GraalVM 19.3?

core/deployment/src/main/java/io/quarkus/deployment/steps/NativeImageAutoFeatureStep.java
gwenneg: I don't know why, but this is how it's done in the entire class.

gwenneg:
I think it can be safely merged, but I'd rather have another (more experienced) opinion about it :)

gsmet (approved):

gsmet:
That was also my understanding. Let's see what CI has to say and decide then :).

geoand:
> I think it can be safely merged, but I'd rather have another (more experienced) opinion about it :)

Who better than @dmlloyd :)

dmlloyd (approved):

gwenneg:
Thanks for the review!

Sanne:
Is there an issue already to track the future TODO mentioned in this patch?

gwenneg:
I just created #5686, thanks for pointing this out!

PR #5502, opened 2019-11-15
Platform resolver to fallback to the bundled platform if the artifact metadata is there but not the artifacts themselves

### LINKED ISSUE #5501: Platform resolver fails in offline mode when the local repo contains metadata but not the artifacts
In that case, the resolver is able to determine the latest version of the platform but fails to actually resolve it.
Unless the user has requested explicitly the desired platform coordinates, the agreed upon behavior in this case is to log a warn and fallback to the platform bundled with the tools.

aloubyansky:
If the local Maven repo includes the metadata for the default platform artifacts but the artifacts themselves are missing, the platform resolver running in offline mode should still fallback to the plaform descriptor bundled with the tools unless the user has explicitly requested the desired platform coordinates.

Fixes #5501


maxandersen (approved):
+1 for this. Verified it works and the logic of throwing explicit exception to catch is sound.

Issue #5324, opened 2019-11-08
quarkus.native.enable-code-size-reporting not work in GraalVM 19.2.1


jorsol:
**Describe the bug**
Set property: `quarkus.native.enable-code-size-reporting` activates `-H:+PrintCodeSizeReport` in GraalVM native image parameters.

**Expected behavior**
The option/property should be probably removed since is not supported in GraalVM.

**Actual behavior**
```
Error: Could not find option 'PrintCodeSizeReport'. Use -H:PrintFlags= to list all available options.
Error: Use -H:+ReportExceptionStackTraces to print stacktrace of underlying exception
Error: Image build request failed with exit status 1
```

**Environment (please complete the following information):**
 - Output of `uname -a` or `ver`: Linux 5.0.0-32-generic
 - Output of `java -version`: OpenJDK 64-Bit GraalVM CE 19.2.1
 - Quarkus version or git rev: 1.0.0.CR1

machi1990:
I'll have a look.

PR #5505, opened 2019-11-15
Opentracing without Servlet

stuartwdouglas:


geoand:
The test failure seems to be related to the changes

extensions/smallrye-opentracing/runtime/src/main/java/io/quarkus/smallrye/opentracing/runtime/QuarkusSmallRyeTracingStandaloneVertxDynamicFeature.java
```diff
@@ -0,0 +1,79 @@
...
+
+@Provider
+public class QuarkusSmallRyeTracingStandaloneVertxDynamicFeature implements DynamicFeature {
+
+    @Override
+    public void configure(ResourceInfo resourceInfo, FeatureContext context) {
+        context.register(StandaloneFilter.class);
+    }
+
+    public static class StandaloneFilter implements ContainerRequestFilter {
```
sberyozkin: This request filter won't catch the issues to do with matching for example. Adding `@PreMatching` would fix that, in which the dynamic feature is not needed. But `ContainerResponseFilter` (without DynamicFeature) should be a safer bet in case some earlier priority pre-matching request filters throw an exception

sberyozkin: @stuartwdouglas looks like your solution is the best one at the moment, registering a listener from `ContainerRequestFilter` given the feedback from @pavolloffay and @objectiser at #5324 

Only please try to make it a Prematching one and it should do

sberyozkin: @stuartwdouglas Hi Stuart, as I mentioned, if JAX-RS layer throws 404 or something else before the filters on the matched method will run (such as those registered in your PR), then these exceptions won't be traced. Unless Resteasy implements it somehow differently

stuartwdouglas: When I changed it to @PreMatch everything failed, I think because the existing tracing filter is not pre match.

sberyozkin: @stuartwdouglas Hmm... All the filter in your PR does it registers a Vertx request completion handler, but this filter will only be invoked if RestEasy has found a match and if it has not then it won't be registered. So I'm not sure why would moving this handler registration to the earlier/pre-match chain would cause the failure - please note a pre-matching filter should be registered without a dynamic feature which assumes one attaches the filters to some specific classes - may be it failed because of that.
It is not a big deal, we can follow up with a minor issue to tweak things to make sure JAX-RS exceptions thrown from the pre-matching filters are also traced. If there is some time, then please try to register a pre-match filter without a feature, otherwise please merge :-)

sberyozkin (approved):

extensions/smallrye-opentracing/runtime/pom.xml
```diff
@@ -46,7 +46,7 @@
         </dependency>
         <dependency>
             <groupId>io.quarkus</groupId>
-            <artifactId>quarkus-undertow</artifactId>
+            <artifactId>quarkus-resteasy</artifactId>
```
gsmet: I think at some point we should really avoid a strong dependency on RESTEasy. Using OpenTracing should be doable even if you don't have REST services.

But I suppose it can wait.

gsmet (approved):

extensions/smallrye-opentracing/runtime/pom.xml
geoand: That makes sense, but is probably an edge case that will need more thought and

sberyozkin: @gsmet, indeed, tracing non-HTTP requests would be very useful too

PR #5509, opened 2019-11-15
Ensure resolver tests runs unaffected by local ~/.m2 repo

maxandersen:
When resolver tests was running they got affected by users local repo causing test assumptions to fail.

this fix explcilty set the local repo home during tests rather than listen to global default that makes behvaior different between machines that runs them.

### DIFF
--- a/independent-projects/bootstrap/core/src/main/java/io/quarkus/bootstrap/resolver/maven/MavenArtifactResolver.java
+++ b/independent-projects/bootstrap/core/src/main/java/io/quarkus/bootstrap/resolver/maven/MavenArtifactResolver.java
@@ -58,6 +58,7 @@ public class MavenArtifactResolver {
     public static class Builder {
 
         private Path repoHome;
+        private boolean reTryFailedResolutionsAgainstDefaultLocalRepo;
         private RepositorySystem repoSystem;
         private RepositorySystemSession repoSession;
         private List<RemoteRepository> remoteRepos = null;
@@ -65,6 +66,23 @@ public class MavenArtifactResolver {
         private LocalWorkspace workspace;
 
         private Builder() {
+        }
+
+        /**
+         * In case custom local repository location is configured using {@link #setRepoHome(Path)},
+         * this method can be used to enable artifact resolutions that failed for the configured
+         * custom local repository to be re-tried against the default user local repository before
+         * failing.
+         * <p>NOTE: the default behavior is <b>not</b> to use the default user local repository as the fallback one.
+         *
+         * @param value  true if the failed resolution requests should be re-tried against the default
+         * user local repo before failing
+         *
+         * @return  this builder instance
+         */
+        public Builder setReTryFailedResolutionsAgainstDefaultLocalRepo(boolean value) {
+            this.reTryFailedResolutionsAgainstDefaultLocalRepo = value;
+            return this;
         }
 
         public Builder setRepoHome(Path home) {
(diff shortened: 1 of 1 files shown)


independent-projects/bootstrap/core/src/main/java/io/quarkus/bootstrap/resolver/maven/MavenArtifactResolver.java
```diff
@@ -127,7 +133,7 @@ private MavenArtifactResolver(Builder builder) throws AppModelResolverException
         if(builder.repoHome != null) {
             final MavenLocalRepositoryManager appCreatorLocalRepoManager = new MavenLocalRepositoryManager(
                     repoSystem.newLocalRepositoryManager(newSession, new LocalRepository(builder.repoHome.toString())),
-                    Paths.get(MavenRepoInitializer.getLocalRepo(MavenRepoInitializer.getSettings())));
+                    builder.localRepoHome==null ? Paths.get(MavenRepoInitializer.getLocalRepo(MavenRepoInitializer.getSettings())) : builder.localRepoHome);
```
aloubyansky: I think it should just be `builder.repoHome`.

aloubyansky (dismissed):
Let's use `repoHome` everywhere.

independent-projects/bootstrap/core/src/main/java/io/quarkus/bootstrap/resolver/maven/MavenArtifactResolver.java
maxandersen: but then you are using the same dir as the remote repository and the local cache. Maven treats that differently internally.

if you as a user want it to be the same you should just do that when calling the builder.

aloubyansky: There is no remote repo here. This is a local repo config. Now that I'm looking into how it's implemented, the benefit of having a delegate was that installed artifacts would end up in `repoHome` instead of the user local one. That was useful for me at the time.
We could keep this as a feature but it shouldn't be the default behavior.

I propose to remove `localRepoHome` and introduce a boolean e.g. `fallbackToUserRepo` which if set to true (the default should be false) would trigger the current behavior, otherwise `repoHome` should be used as the only local repo.

aloubyansky: i.e. this should be the default local repo manager if `repoHome` isn't null: `repoSystem.newLocalRepositoryManager(newSession, new LocalRepository(builder.repoHome.toString()))`

aloubyansky: if the user wants delegation then its the current `new MavenLocalRepositoryManager(
                    repoSystem.newLocalRepositoryManager(newSession, new LocalRepository(builder.repoHome.toString())),
                    Paths.get(MavenRepoInitializer.getLocalRepo(MavenRepoInitializer.getSettings())));`

maxandersen: as long as you are sure we are not by accident changing maven resolutoin logic for everything else but this one test then i'm fine - if unsure i suggest we keep the change for 1.0 absolute minimal and then in 1+ look at change the logic.

aloubyansky: I'm sure.

aloubyansky:
@gsmet i updated this one

gsmet (approved):

PR #5510, opened 2019-11-15
Show more robust gradle graal configuration in docs

geoand:
@aguibert could you also take a look?

### DIFF
--- a/docs/src/main/asciidoc/writing-native-applications-tips.adoc
+++ b/docs/src/main/asciidoc/writing-native-applications-tips.adoc
@@ -77,7 +77,7 @@ When using Gradle, we need to configure the additional build arguments in the pl
 ----
 buildNative {
     additionalBuildArgs = [
-            '-H:ResourceConfigurationFiles=../resources/main/resources-config.json'
+            "-H:ResourceConfigurationFiles=${buildDir}/resources/main/resources-config.json".toString()
     ]
 }
 ----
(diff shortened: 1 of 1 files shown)


docs/src/main/asciidoc/writing-native-applications-tips.adoc
```diff
@@ -77,7 +77,7 @@ When using Gradle, we need to configure the additional build arguments in the pl
 ----
 buildNative {
     additionalBuildArgs = [
-            '-H:ResourceConfigurationFiles=../resources/main/resources-config.json'
+            "-H:ResourceConfigurationFiles=${buildDir}/resources/main/resources-config.json".toString()
```
gsmet: I don't know Gradle very well but is the `toString()` needed? It looks weird.

(Just asking, I suppose you added it because it was required)

aguibert: yea the `toString()` is needed otherwise Groovy interprets it as a `groovy.lang.GString` and not a `java.lang.String` and then we get a classcastexception. An alternative would be to do:
```
'-H:ResourceConfigurationFiles=' + buildDir + '/resources/main/resources-config.json'
```

aguibert (approved):

docs/src/main/asciidoc/writing-native-applications-tips.adoc
geoand: Exactly, groovy has its own GString which needs to be converted to regular String

gastaldi (approved):
LGTM

PR #5517, opened 2019-11-15
Allow falling back to the bundled platform in case quarkus-platform-descriptor-json:jar couldn't be resolved

aloubyansky:
If quarkus-platform-descriptor-json:jar couldn't be resolved and the user didn't provide any specific platform coordinates then fallback to the platform bundled with the tools


aloubyansky:
@gsmet fyi, this should fix your issue

maxandersen (commented):
Looks good to me but I can't verify it fixes as I can't reproduce the error @gsmet sees - I'll let him verify!

gsmet (approved):
It fixed my issue.

PR #5519, opened 2019-11-15
refactor: remove trailing usage of @SubstrateTest

machi1990:
follows up https://github.com/quarkusio/quarkus/pull/5007

/cc @gwenneg

### DIFF
--- a/independent-projects/tools/common/src/test/resources/templates/basic-rest/kotlin/native-test-resource-template.ftl
+++ b/independent-projects/tools/common/src/test/resources/templates/basic-rest/kotlin/native-test-resource-template.ftl
@@ -1,6 +1,6 @@
 package ${package_name}
 
-import io.quarkus.test.junit.SubstrateTest
+import io.quarkus.test.junit.NativeImageTest
 
-@SubstrateTest
+@NativeImageTest
 open class Native${class_name}IT : ${class_name}Test()
--- a/integration-tests/main/src/test/java/io/quarkus/it/main/HealthCheckTestCase.java
+++ b/integration-tests/main/src/test/java/io/quarkus/it/main/HealthCheckTestCase.java
@@ -4,26 +4,24 @@ import javax.inject.Inject;
 
 import org.eclipse.microprofile.health.HealthCheckResponse;
 import org.eclipse.microprofile.health.Liveness;
-import org.junit.jupiter.api.Assumptions;
 import org.junit.jupiter.api.Test;
 import org.wildfly.common.Assert;
 
 import io.quarkus.it.health.SimpleHealthCheck;
-import io.quarkus.test.junit.NativeImageTest;
+import io.quarkus.test.junit.DisabledOnNativeImage;
 import io.quarkus.test.junit.QuarkusTest;
-import io.quarkus.test.junit.SubstrateTest;
 
 @QuarkusTest
+@DisabledOnNativeImage("This test is not meant to be ran in native mode as Quarkus does not yet support injection " +
+        "in native " + "tests - see https://quarkus.io/guides/getting-started-testing#native-executable-testing")
 public class HealthCheckTestCase {
 
     @Inject
     @Liveness
-    SimpleHealthCheck checkks;
+    SimpleHealthCheck simpleHealthCheck;
 
     @Test
     public void testInjection() {
-        Assumptions.assumeFalse(getClass().isAnnotationPresent(SubstrateTest.class));
-        Assumptions.assumeFalse(getClass().isAnnotationPresent(NativeImageTest.class));
-        Assert.assertTrue(checkks.call().getState() == HealthCheckResponse.State.UP);
+        Assert.assertTrue(simpleHealthCheck.call().getState() == HealthCheckResponse.State.UP);
     }
 }
--- a/integration-tests/test-extension/src/test/java/io/quarkus/it/extension/FinalFieldReflectionInGraalITCase.java
+++ b/integration-tests/test-extension/src/test/java/io/quarkus/it/extension/FinalFieldReflectionInGraalITCase.java
@@ -7,11 +7,13 @@ import org.junit.jupiter.api.Test;
 import io.quarkus.test.junit.SubstrateTest;
 import io.restassured.RestAssured;
 
+// TODO replace with @NativeImageTest once @SubstrateTest is removed.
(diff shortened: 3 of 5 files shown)


independent-projects/tools/common/src/test/resources/templates/basic-rest/kotlin/native-test-resource-template.ftl
```diff
@@ -1,6 +1,6 @@
 package ${package_name}
 
-import io.quarkus.test.junit.SubstrateTest
+import io.quarkus.test.junit.NativeImageTestTest
```
gsmet: This and a few others below looks wrong. `Test` is repeated.

gsmet (changes requested):

independent-projects/tools/common/src/test/resources/templates/basic-rest/kotlin/native-test-resource-template.ftl
machi1990: Good catch.

machi1990: done

gsmet (approved):

gsmet:
Looks better, I'll let @gwenneg review.

integration-tests/test-extension/src/test/java/io/quarkus/it/extension/FinalFieldReflectionInGraalITCase.java
```diff
@@ -4,10 +4,10 @@
 
 import org.junit.jupiter.api.Test;
 
-import io.quarkus.test.junit.SubstrateTest;
+import io.quarkus.test.junit.NativeImageTest;
 import io.restassured.RestAssured;
 
-@SubstrateTest
+@NativeImageTest
 public class FinalFieldReflectionInGraalITCase {
```
gwenneg: This test is still annotated with `@SubstrateTest` to make sure we don't break the deprecated classes until their removal. We should probably keep it that way for now and add a comment to explain why it is not annotated with `@NativeImageTest`.

integration-tests/main/src/test/java/io/quarkus/it/main/HealthCheckTestCase.java
```diff
@@ -22,7 +21,6 @@
 
     @Test
     public void testInjection() {
-        Assumptions.assumeFalse(getClass().isAnnotationPresent(SubstrateTest.class));
```
gwenneg: I don't know the reason, but this test was designed to prevent it from being launch in native mode. That's why the presence of both `@SubstrateTest` and `@NativeImageTest` is checked here. It should probably stay that way until `@SubstrateTest` is removed.

gwenneg:
The build failure seems unrelated.

integration-tests/main/src/test/java/io/quarkus/it/main/HealthCheckTestCase.java
machi1990: The reason is because injection is not supported in native mode. I think we can safely remove Substrate here, if it is launched as native mode, the test will fail.

integration-tests/test-extension/src/test/java/io/quarkus/it/extension/FinalFieldReflectionInGraalITCase.java
machi1990: Good to know this. 
```
// Use @SubstrateTest to ensure backward compatibility. 
// TODO replace with @NativeImageTest once SubstrateTest is removed. 
```

WDYT?

gwenneg: That sounds good!

integration-tests/main/src/test/java/io/quarkus/it/main/HealthCheckTestCase.java
gwenneg: That's good to know, but this reason might not be obvious to everyone. I'd rather keep these two checks and add an explicit message in case of failure. And also a comment explaining why we're testing `@SubstrateTest` :) That way, we'd be sure nobody will ever try to run this test in native mode even if the injection is failing.

gwenneg: Actually, we should probably use `@DisabledOnNativeImage` instead.

machi1990: It is minor but I still think we can remove the `@SubstrateTest` check here: I do not think we'll try to run this test in native mode with `@SubstrateTest` - since it is deprecated. 

- I can however add a comment - 
```
The Injection test is not meant to be ran in native mode as it is not supported - https://quarkus.io/guides/getting-started-testing#native-executable-testing
```

gwenneg (approved):
LGTM, thanks!

gwenneg:
The build failure was caused by a `FinalFieldReflectionInGraalITCase` imports sort issue. Could you fix that please @machi1990?

independent-projects/tools/common/src/test/resources/templates/basic-rest/kotlin/native-test-resource-template.ftl
Sanne: wasn't fixed yet? :)

Sanne (changes requested):
Wrong name `io.quarkus.test.junit.NativeImageTestTest` in the templates

machi1990:
> Wrong name `io.quarkus.test.junit.NativeImageTestTest` in the templates

Good catch. These ones should have been fixed earlier though ;-)

Thanks @Sanne

independent-projects/tools/common/src/test/resources/templates/basic-rest/kotlin/native-test-resource-template.ftl
machi1990: I might have walked on my feet myself after changing machines :-(

Thanks for catching this.

Sanne (approved):

Sanne:
NP, thanks! Let's wait for CI one more time.

machi1990:
CI failure tracked by #5536 and is unrelated to the change.

Merging. Thanks @gwenneg @gsmet @Sanne

PR #5532, opened 2019-11-16
refactor(LogConfig): Remove optional @DefaultConverter

aoudiamoncef:
Hi @dmlloyd, 
According to [writing extensions guide](https://quarkus.io/guides/writing-extensions) we should use `@DefaultConverter` if we have one, i thought that this is the case of `Level.class` && `InetSocketAddress`. 

Thanks

### DIFF
--- a/core/runtime/src/main/java/io/quarkus/runtime/ExecutorRecorder.java
+++ b/core/runtime/src/main/java/io/quarkus/runtime/ExecutorRecorder.java
@@ -98,10 +98,10 @@ public class ExecutorRecorder {
                                 final List<Runnable> runnables = executor.shutdownNow();
                                 if (!runnables.isEmpty()) {
                                     log.warnf("Thread pool shutdown failed: discarding %d tasks, %d threads still running",
-                                            Integer.valueOf(runnables.size()), Integer.valueOf(executor.getActiveCount()));
+                                            runnables.size(), executor.getActiveCount());
                                 } else {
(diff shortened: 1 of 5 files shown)


dmlloyd:
No, that annotation is only necessary when we have a non-default converter for a given type, which we do not in these cases.  Right now the default converter is used in most cases already, so this change is not needed.

aoudiamoncef:
According to the current code in `FileConfig` && `ConsoleConfig`: 
```java
    /**
     * The level of logs to be written into the file.
     */
    @DefaultConverter
    @ConfigItem(defaultValue = "ALL")
    Level level;
```
So it could be removed ?

dmlloyd:
Correct, it should be removed.

aoudiamoncef:
@dmlloyd, 
Please, could you tell me why did this code is using unboxing ?

```java
log.warnf("Thread pool shutdown failed: %d threads still running",
                                            Integer.valueOf(executor.getActiveCount()));
```

dmlloyd:
I'm sure it was a mistake.

aoudiamoncef:
I removed it in my last commit.

aoudiamoncef:
Hi @dmlloyd, 
I think that i finished all the  modifications on this PR.

dmlloyd (approved):

dmlloyd:
Just have to wait for CI testing to finish.

machi1990:
Merged, thanks @aoudiamoncef

PR #5533, opened 2019-11-16
fix: remove build warning by making injected spring data jpa fragments protected in a generated repository

machi1990:
This was causing the following warning:
```
[INFO] [io.quarkus.arc.processor.BeanProcessor] Found unrecommended usage of private members (use package-private instead) in application beans:
- @Inject field io.quarkus.it.spring.data.jpa.PersonRepository_bcb609283abd593bd6e8ce51955e38a20ba9d528Impl#customImplClass1,
- @Inject field io.quarkus.it.spring.data.jpa.PersonRepository_bcb609283abd593bd6e8ce51955e38a20ba9d528Impl#customImplClass2
```

### DIFF
--- a/extensions/spring-data-jpa/deployment/src/main/java/io/quarkus/spring/data/deployment/generate/SpringDataRepositoryCreator.java
+++ b/extensions/spring-data-jpa/deployment/src/main/java/io/quarkus/spring/data/deployment/generate/SpringDataRepositoryCreator.java
@@ -163,7 +163,7 @@ public class SpringDataRepositoryCreator {
         for (String customImplClassName : customImplClassNames) {
             FieldCreator customClassField = repositoryImpl
                     .getFieldCreator("customImplClass" + (i + 1), customImplClassName)
-                    .setModifiers(Modifier.PRIVATE);
+                    .setModifiers(Modifier.PROTECTED); // done to prevent warning during the build
             customClassField.addAnnotation(Inject.class);
 
             customImplNameToFieldDescriptor.put(customImplClassName,


extensions/spring-data-jpa/deployment/src/main/java/io/quarkus/spring/data/deployment/generate/SpringDataRepositoryCreator.java
```diff
@@ -163,7 +163,7 @@ private void createCustomImplFields(ClassCreator repositoryImpl, List<DotName> c
         for (String customImplClassName : customImplClassNames) {
             FieldCreator customClassField = repositoryImpl
                     .getFieldCreator("customImplClass" + (i + 1), customImplClassName)
-                    .setModifiers(Modifier.PRIVATE);
+                    .setModifiers(Modifier.PUBLIC); // done to prevent warning during the build
```
geoand: Nice catch!

I wonder, if you remove the modifier completely, does it make the field package private and still remove the warning?

machi1990: > Nice catch!

Thanks.
 
> I wonder, if you remove the modifier completely, does it make the field package private and still remove the warning?

By default the modifier is `private`, so I think only removing the modifier not make the field package private.

geoand: I don't really like the public field, but of course if there aren't any other alternatives, it's much better that having a build warning

machi1990: Likewise here. let's use `protected` :-)

geoand (approved):
Thanks!

PR #5534, opened 2019-11-16
fix: NotFoundExceptionMapperTestCase fails on some environments

### LINKED ISSUE #5424: NotFoundExceptionMapperTestCase fails on some environments
**Describe the bug**
The two tests in `io.quarkus.resteasy.test.NotFoundExceptionMapperTestCase` added by @machi1990 which replace resources and expect to see the changed resources fail in some environments and not in others.

@kenfinnigan added some sleeps to these tests which seemed to fix the problems sometimes, yet the problem is still not consistently fixed.

The problem may be due to inadequate resolution of file and/or directory time stamps.  This was a known problem in Java 8 (where the test fails for me) that was fixed in Java 11 (where the test does not fail for me).

The commit which introduced the problematic test was d81e508b6fb543724bd526c01304c28ebedc3d8f.  The commit which added the sleeps was 24876d0b7a9be4445f4ee4188f97aba4dc9cd1cb.

**To Reproduce**
1. Run this test class on Java 8, hopefully the problem reappears.  Removing the sleeps might be necessary.

machi1990:
Fixes #5424 

/cc @dmlloyd can you have a look? Thanks.

### DIFF
--- a/core/devmode/src/main/java/io/quarkus/dev/RuntimeUpdatesProcessor.java
+++ b/core/devmode/src/main/java/io/quarkus/dev/RuntimeUpdatesProcessor.java
@@ -286,7 +286,7 @@ public class RuntimeUpdatesProcessor implements HotReplacementContext {
     }
 
     Set<String> checkForFileChange() {
-        Set<String> ret = new HashSet<>();
+        Set<String> changedFiles = new HashSet<>();
         for (DevModeContext.ModuleInfo module : context.getModules()) {
             final Set<Path> moduleResources = correspondingResources.computeIfAbsent(module.getName(),
                     m -> Collections.newSetFromMap(new ConcurrentHashMap<>()));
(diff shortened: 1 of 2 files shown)


dmlloyd (commented):
I'm not really sure whether or not this fixes the problem.  I don't see anything that looks like a mistake per se, but someone who is more familiar with this area should review the change.

machi1990:
> I'm not really sure whether or not this fixes the problem. I don't see anything that looks like a mistake per se, but someone who is more familiar with this area should review the change.

Thanks David for the feedback. I should probably ping @stuartwdouglas for a review. Were you able to quick launch the scripts that ran the failing tests in question and see if the issue was fixed? Thanks

dmlloyd:
I'm in the middle of a big ugly checkout/rebase right now so I can't verify this right away.

gsmet:
@jaikiran any chance you could have a look at this one before the end of the week? It would be nice to have it in 1.1.0.

If not, just drop me a note and I'll see if Stuart or I can have a look.

Thanks!

jaikiran:
Hello @gsmet, this turned out to be a test framework issue. I've opened https://github.com/quarkusio/quarkus/pull/5913 to fix this and have added @dmlloyd and @machi1990 as reviewers. @machi1990, can you please take a look at the other PR and see if it's good enough?

machi1990:
Hey @jaikiran thanks for taking care of this one and for proposing a reliable fix. 
Let me close this one in favor of your PR.

PR #5542, opened 2019-11-17
fix CNFE when running an app packaged with custom main class

machi1990:
When a user defined the main class via `quarkus.package.main-class=foo.bar.MainClass` property the
following error `Error: Could not find or load main class foo.bar.MainClass. Caused by:
java.lang.ClassNotFoundException: foo.bar.MainClass` occurred when launching the application and app could not be started

/cc @dmlloyd @stuartwdouglas


integration-tests/maven/src/test/java/io/quarkus/maven/it/JarRunnerIT.java
```diff
@@ -28,7 +28,9 @@ public void testThatJarRunnerConsoleOutputWorksCorrectly() throws MavenInvocatio
         File testDir = initProject("projects/classic", "projects/project-classic-console-output");
         RunningInvoker running = new RunningInvoker(testDir, false);
 
-        MavenProcessInvocationResult result = running.execute(Arrays.asList("package", "-DskipTests"), Collections.emptyMap());
+        MavenProcessInvocationResult result = running.execute(
+                Arrays.asList("package", "-DskipTests", "-Dquarkus.package.main-class=foo.bar.MainClass"),
+                Collections.emptyMap());
```
gsmet: Please add another test when you do that.

core/runtime/src/main/java/io/quarkus/runtime/Quarkus.java
```diff
@@ -1,20 +0,0 @@
...
-
-/**
- * The main entry point class, calling main allows you to bootstrap quarkus
- * <p>
- * Note that at native image generation time this is replaced by {@link io.quarkus.runtime.graal.QuarkusReplacement}
- * which will avoid the need for reflection.
- * <p>
- * TODO: how do we deal with static init
- */
-public class Quarkus {
```
gsmet: These classes are unused?

gsmet (changes requested):
Added a couple of comments/questions.

core/runtime/src/main/java/io/quarkus/runtime/Quarkus.java
machi1990: Yes, looks so to me. I'd like another opinion though @stuartwdouglas @dmlloyd

integration-tests/maven/src/test/java/io/quarkus/maven/it/JarRunnerIT.java
machi1990: I think the test in here is enough as without this patch if you add the mainClass property (via build tool e.g maven or in application.properties )you could not launch the application - https://github.com/quarkusio/quarkus/blob/95590681ea0a9c94be8f81a16114e0eca0495ab5/integration-tests/maven/src/test/java/io/quarkus/maven/it/JarRunnerIT.java#L44-L47.

core/runtime/src/main/java/io/quarkus/runtime/Quarkus.java
dmlloyd: I believe that the intent of this class was to provide a class that can be run from an IDE.

machi1990: Oh good to know, thanks David. So we should keep it. Same question for `QuarkusReplacement.java` contains a substitution of the generated main class? Can it be deleted?

dmlloyd: I'm not sure what the current status of that is TBH.

machi1990: I’ll try removing it and keep the Substitutee and see if it the substitutions are still needed.

machi1990: I kept the `Quarkus` for IDE runs and removed the `QuarkusReplacement`.

stuartwdouglas: These are leftovers from the early days, they can both go.

machi1990: So the two classes are gone. This is ready for another round of review.

integration-tests/maven/src/test/java/io/quarkus/maven/it/JarRunnerIT.java
stuartwdouglas: I think this should really be a different test to avoid confusion. I also don't understand how this test passes? Where does foo.bar.MainClass come from?

machi1990: Yes, we can have another test. I wanted to quickly add the mainClass property when packaging in this existing JarRunner test to avoid straining the CI with yet another similar test. 

foo.bar.MainClass is just an arbitrary class name supplied as configuration property to override the default ‘io.quarkus.runner.GeneratedMain’

stuartwdouglas: But the class doesn't exist so why does the test pass?

machi1990: It passes because of my second commit https://github.com/quarkusio/quarkus/pull/5542/commits/a069ecba35fcd5d9cf0f977b3ec9d091b2233562

machi1990: Is this how the mainClass property should be used? Or it is a user defined one?

stuartwdouglas:
Can we hold this one for now, as this likely ties into what we are going to do with command mode and main methods (which is an ongoing discussion on the list). Hopefully we will have an answer for that soon and we can define the semantics exactly what the main-class attribute is, as at the moment it is not particularly well defined.

machi1990:
> Can we hold this one for now, as this likely ties into what we are going to do with command mode and main methods (which is an ongoing discussion on the list). Hopefully we will have an answer for that soon and we can define the semantics exactly what the main-class attribute is, as at the moment it is not particularly well defined.

Thanks for the feedback @stuartwdouglas. I noticed the issue when I was playing with jib.  

Sure, let's hold this. Can we have a separate PR with only the first commit removing the `Quarkus` and `QuarkusReplacement` classes ?

stuartwdouglas:
Sure, sounds good.

machi1990:
> Sure, sounds good.

Let me prepare it.

gsmet:
I think I will close this one for now. All this will probably be changed soon with the command mode.

PR #5547, opened 2019-11-17
Specify keycloak version in tests

stuartwdouglas:


integration-tests/oidc-code-flow/pom.xml
```diff
@@ -218,7 +218,7 @@
                         <configuration>
                             <images>
                                 <image>
-                                    <name>quay.io/keycloak/keycloak</name>
+                                    <name>quay.io/keycloak/keycloak::7.0.1</name>
```
gsmet: Double colons here.

gsmet (changes requested):

integration-tests/keycloak-authorization/pom.xml
```diff
@@ -193,7 +193,7 @@
                         <configuration>
                             <images>
                                 <image>
-                                    <name>quay.io/keycloak/keycloak</name>
+                                    <name>quay.io/keycloak/keycloak:7.0.1</name>
```
gsmet: Couldn't we use a centralized property for that? It would have to be in the `build-parent` pom.xml.

stuartwdouglas:
It looks like https://github.com/quarkusio/quarkus/commit/60c4d44982b361907571f7b5b77dd6e47d6537c1 removed the fixed version number for keycloak

PR #5560, opened 2019-11-18
Add some standards to the README

### LINKED ISSUE #2404: We call our "standards compliant" but don't actually mention any standards
In quarkus repo the README states:

"Standards-based: Based on the standards and frameworks you love and use (RESTEasy, Hibernate, Netty, Eclipse Vert.x, Apache Camel...)"

Everything listed in parentheses is a project and whilst some standards are implied (e.g., JAX-RS or JPA) they're not called out explicitly. If this is an oversight then let's add a few standards (I'd also suggest MicroProfile). If it's deliberate then let's discuss.

gsmet:
Fix #2404

This is very minor but that's one more issue closed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ Embraces [12 factor architecture](https://12factor.net) in environments like Kub
 * **Unify imperative and reactive**:
 Brings under one programming model non blocking and imperative styles of development.
 * **Standards-based**:
-Based on the standards and frameworks you love and use (RESTEasy, Hibernate, Netty, Eclipse Vert.x, Apache Camel...)
+Based on the standards and frameworks you love and use (RESTEasy and JAX-RS, Hibernate ORM and JPA, Netty, Eclipse Vert.x, Eclipse MicroProfile, Apache Camel...)
 * **Microservice First**:
 Brings lightning fast startup time and code turn around to Java apps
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -16,7 +16,7 @@ Embraces [12 factor architecture](https://12factor.net) in environments like Kub
 * **Unify imperative and reactive**:
 Brings under one programming model non blocking and imperative styles of development.
 * **Standards-based**:
-Based on the standards and frameworks you love and use (RESTEasy, Hibernate, Netty, Eclipse Vert.x, Apache Camel...)
+Based on the standards and frameworks you love and use (RESTEasy and JAX-RS, Hibernate ORM and JPA, Netty, Eclipse Vert.x, Eclipse MicroProfile, Apache Camel...)
```
machi1990: We have some Spring API supports, maybe mention that too?

gsmet: This is supposed to be an easy and non polemic one :).

Communication around our Spring compatibility layer is a bit more complicated than that so let's keep it simple for now.

stuartwdouglas (approved):

README.md
machi1990: Good to know, Thanks :-)

PR #5481, opened 2019-11-14
Consolidate scheduler extension - part 1

mkouba:
- `quarkus-scheduler` contains the API (`io.quarkus.scheduler.Scheduled` and friends), shared logic to collect/validate scheduled methods and a simple scheduler impl (we use `com.cronutils:cron-utils` to parse/validate/migrate CRON expressions, a `ScheduledExecutorService` with one "scheduler" thread that delegates the execution of scheduled methods to the `ExecutorService` obtained from `ExecutorBuildItem`; if `quarkus-scheduler` is present this impl is disabled)
- `quarkus-quartz` depends on `quarkus-scheduler` and reuses most of the build steps and API; however, it's using Quartz as the underlying scheduler impl and provides some quartz-specific configuration
- CRON syntax is configurable through the `quarkus.scheduler.cron-type` property; the default value is `quartz` to retain backward compatibility; other options are `unix` and `cron4j`
- the `Scheduler#startTimer()` method was removed - it was a mistake
- there is a new intergration test because previously scheduler was included in `quarkus-integration-test-main` but there were no assertions

TODO (next PRs):
- update the docs, although almost everything should work as before; mention the `quarkus.scheduler.cron-syntax` property - PR  #5481 opened
- consider adding more quartz-specific configuration (@machi1990 clustered job store etc.) PR opened for clustered jobs https://github.com/quarkusio/quarkus/pull/5529
- maybe add some more tests ;-)

PR #5561, opened 2019-11-18
docs: mention the quarkus.scheduler.cron-type property

machi1990:
Follows up https://github.com/quarkusio/quarkus/pull/5481

### DIFF
--- a/docs/src/main/asciidoc/scheduler.adoc
+++ b/docs/src/main/asciidoc/scheduler.adoc
@@ -109,7 +109,7 @@ public class CounterBean {
 (Quarkus is using 10 worker threads for the scheduler). If it is not available the method invocation should be re-scheduled by default i.e
 it should be invoked as soon as possible. The invocation of the scheduled method does not depend on the status or result of the previous invocation.
 4. The code is pretty straightforward. Every 10 seconds, the counter is incremented.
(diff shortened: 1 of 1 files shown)


docs/src/main/asciidoc/scheduler.adoc
```diff
@@ -109,16 +109,18 @@ public class CounterBean {
...
-5. Define a job with a cron-like expression. The syntax is currently based on http://www.quartz-scheduler.org/documentation/quartz-2.3.0/tutorials/crontrigger.html[Quartz Cron Trigger]. The annotated method is executed at 10:15am every day.
+5. Define a job with a cron-like expression. The annotated method is executed at 10:15am every day.
 6. Define a job with a cron-like expression `cron.expr` which is configurable in `application.properties`.
 
 == Updating the application configuration file
 
 Edit the `application.properties` file and add the `cron.expr` configuration:
 [source,shell]
 ----
+quarkus.scheduler.cron-type=quartz <1>
```
mkouba: Hm, `quartz` is the default value. So I wonder if we should comment out this line and add one more line above with _"Define the cron syntax to use..."_?

machi1990: You are suggesting that we remove this config line?

mkouba: Not removing, but something like this instead:
```properties
# By default, the syntax used for cron expressions is based on Quartz
# You can change the syntax using the following property:
# quarkus.scheduler.cron-type=unix

cron.expr=*/5 * * * * ?
```

WDYT?

mkouba (approved):

Issue #1349, opened 2019-03-08
Extension to view the values configured by MicroProfile config in development mode


hpehl:
Integrate the [config extension](https://github.com/hpehl/quarkus-config-extension) into Quarkus. Can come handy in both dev and production mode.

emmanuelbernard:
I was thinking of dev mode only actually to avoid some security disaster.

hpehl:
Yes, makes sense to me. We could make it part of the configuration: Default would be enabled in devmode, disabled in production mode.

emmanuelbernard:
We hate configuration toggles. They complicate the user experience. You really need an important reason to bring one in. What usage flow would not work if we  were not allowing this to be seen in PROD mode?

hpehl:
Agreed, no toggles. I've updated the config extension to only work in development mode:
https://github.com/hpehl/quarkus-config-extension/commit/3530c88c908288c530f82047b8150741f2c9e152

gsmet:
@hpehl do you plan to do it or should I close it?

hpehl:
I'll update the extension and open a PR.

stuartwdouglas:
We can now do this in the config editor in Dev UI, so IMHO this can be closed.

hpehl:
No objections from my side.

PR #5562, opened 2019-11-18
Extension to view MicroProfile config values in dev mode

hpehl:
PR for #1349. 

Adds a servlet in dev mode which lists the configured values for MicroProfile Configuration. The servlet is registered using the path `/config` by default and lists all properties of all config sources. The config sources are sorted descending by ordinal, the properties by name. If no config is defined an empty JSON object is returned.

There's no integration test in this PR. Since the servlet is only added in dev mode, `@QuarkusTest` does not work as expected. 

If necessary I can provide documentation.


extensions/config/deployment/pom.xml
```diff
@@ -0,0 +1,49 @@
...
+
+    <artifactId>quarkus-config-deployment</artifactId>
+    <name>Quarkus - Show Configuration - Deployment</name>
+
+    <dependencies>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-core-deployment</artifactId>
+        </dependency>
+        <dependency>
```
kenfinnigan: I would be inclined to go lower and just have it depend on vertx instead of resteasy

kenfinnigan (changes requested):
I have a problem with the names being "config" for the config and "quarkus-config" for the extension.

To me that implies these bring with them the ability to configure an application when that functionality is already present.

Maybe a name like "config-viewer", not sure tbh

extensions/config/deployment/pom.xml
hpehl: Good point. I'll adjust the dependencies.

hpehl:
I agree about the naming issue. Since it just dumps the current configuration, maybe `quarkus-config-dump` is a better choice.

kenfinnigan:
Maybe `quarkus-config-viewer`?

hpehl:
Sorry my last commit introduced too many changes. I'll close this PR and re-submit into a new PR.

hpehl:
Reopened as #5692

PR #5563, opened 2019-11-18
Add @Cache and @NoCache tests

### LINKED ISSUE #5551: Resteasy @NoCache Annotation doesn't work
Hey,
we were using the `@NoCache` Annotation of RESTEasy to disable the caching for some rest interfaces. It is not working with Quarkus 1.0.0.CR1

It is ignored and no cache related headers are set.

gastaldi:
This PR provides tests for #5551 that would ensure that `@NoCache` and `@Cache` work as advertised.

gsmet: But apparently, the OP still has the issue.

### DIFF
--- a/extensions/resteasy/deployment/src/test/java/io/quarkus/resteasy/test/CacheControlFeatureTest.java
+++ b/extensions/resteasy/deployment/src/test/java/io/quarkus/resteasy/test/CacheControlFeatureTest.java
@@ -1,0 +1,30 @@
+package io.quarkus.resteasy.test;
+
+import static io.restassured.RestAssured.when;
+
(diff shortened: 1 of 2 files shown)


gsmet (changes requested):
Let's wait for the OP's feedback before merging this PR and (thus) closing the issue.

geoand:
Seems like a resonable change, but @gsmet's suggestion also makes perfect sense

geoand:
I see from #5551 that this doesn't solve your issue @rboei ?

gsmet (approved):
Doesn't hurt to have a test. Let's merge this even if it doesn't seem to solve the OP's original issue.

PR #5564, opened 2019-11-18
Replacing quarkus.properties with quarkus.platform.version

### LINKED ISSUE #4744: Generated maven project should have a property for quarkus platform and version
**Describe the bug**
when mvn project generated and quarkus version != platform version we fall back to hardcoded values.
Should use property replacements instead., i.e. ${quarkus.platform.version} and ${quarkus.platform.id} 

**To Reproduce**
use latest master/0.26 with universe platform with a version number updated to be 998-SNAPSHOT in all its pom.xml files and run:

`mvn io.quarkus:quarkus-maven-plugin:999-SNAPSHOT:create -DprojectGroupId=org.acme -DprojectArtifactId=application-configuratxio557 -DclassName="org.acme.config.GreetingResource" -Dpath="/greeting" -DplatformArtifactId=quarkus-universe-bom -DplatformVersion=998-SNAPSHOT`

aloubyansky:
...and quarkus-plugin.version in Maven templates and quarkusVersion with quarkusPluginVersion in the Gradle ones, also adding quarkus.platform.group-is and quarkus.platform.artifact-id to pom.xml templates and sorting of properties before persisting pom.xml.

Fixes #4744

### DIFF
--- a/core/deployment/src/main/java/io/quarkus/deployment/configuration/ConfigDefinition.java
+++ b/core/deployment/src/main/java/io/quarkus/deployment/configuration/ConfigDefinition.java
@@ -71,6 +71,9 @@ public class ConfigDefinition extends CompoundConfigType {
             .asList(QUARKUS_NAMESPACE + ".live-reload.password", QUARKUS_NAMESPACE + ".live-reload.url",
                     QUARKUS_NAMESPACE + ".debug.generated-classes-dir", QUARKUS_NAMESPACE + ".debug.reflection",
                     QUARKUS_NAMESPACE + ".build.skip",
+                    QUARKUS_NAMESPACE + ".platform.group-id",
+                    QUARKUS_NAMESPACE + ".platform.artifact-id",
+                    QUARKUS_NAMESPACE + ".platform.version",
                     QUARKUS_NAMESPACE + ".version", QUARKUS_NAMESPACE + ".profile", QUARKUS_NAMESPACE + ".test.profile",
                     QUARKUS_NAMESPACE + ".test.native-image-wait-time",
                     QUARKUS_NAMESPACE + ".test.native-image-profile");
--- a/independent-projects/tools/common/src/test/java/io/quarkus/cli/commands/ListExtensionsTest.java
+++ b/independent-projects/tools/common/src/test/java/io/quarkus/cli/commands/ListExtensionsTest.java
@@ -36,7 +36,7 @@ public class ListExtensionsTest {
 
         FileProjectWriter writer = new FileProjectWriter(pom.getParentFile());
         new CreateProject(writer)
-                .groupId(getPluginGroupId())
+                .groupId("org.acme")
                 .artifactId("add-extension-test")
                 .version("0.0.1-SNAPSHOT")
                 .doCreateProject(context);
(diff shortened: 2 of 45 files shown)


independent-projects/tools/common/src/test/java/io/quarkus/cli/commands/ListExtensionsTest.java
```diff
@@ -36,7 +36,7 @@ public void listWithBom() throws IOException {
 
         FileProjectWriter writer = new FileProjectWriter(pom.getParentFile());
         new CreateProject(writer)
-                .groupId(getPluginGroupId())
+                .groupId("org.acme")
```
maxandersen: why are all these "org.acme" reference replaced? shouldn't they stay the same for this test ?

core/deployment/src/main/java/io/quarkus/deployment/configuration/ConfigDefinition.java
```diff
@@ -71,6 +71,10 @@
             .asList(QUARKUS_NAMESPACE + ".live-reload.password", QUARKUS_NAMESPACE + ".live-reload.url",
                     QUARKUS_NAMESPACE + ".debug.generated-classes-dir", QUARKUS_NAMESPACE + ".debug.reflection",
                     QUARKUS_NAMESPACE + ".build.skip",
+                    QUARKUS_NAMESPACE + ".platform.group-id",
+                    QUARKUS_NAMESPACE + ".platform.artifact-id",
+                    QUARKUS_NAMESPACE + ".platform.version",
+                    QUARKUS_NAMESPACE + "-plugin.version",
```
maxandersen: hmm - when seeing it here and in the lower pom.xml's then the `quarkus-plugin` does feel out of place as one would expect `quarkus.plugin`. 

should we make it `quarkus.plugin.version` just to not trigger other peoples OCD ? :)

maxandersen: okey - just did a quick search on github and "-plugin.version" seems waay more prominent (millions vs thousands) than ".plugin.version" so lets just keep it as is.

maxandersen (changes requested):
i tested it with and without `-o` and for mvn/gradle and it works.

Just want to know if those org.acme changes in the test was really intended.

core/deployment/src/main/java/io/quarkus/deployment/configuration/ConfigDefinition.java
dmlloyd: I'm pretty sure that this entry isn't needed.  The config subsystem should only warn on properties whose name starts with `quarkus.`

independent-projects/tools/common/src/test/java/io/quarkus/cli/commands/ListExtensionsTest.java
aloubyansky: Because the groupId of quarkus-maven-plugin is not related to the groupId of a random test project. Or am I missing something?

core/deployment/src/main/java/io/quarkus/deployment/configuration/ConfigDefinition.java
aloubyansky: Ok, removing it from here. Thanks.

independent-projects/tools/common/src/test/java/io/quarkus/cli/commands/ListExtensionsTest.java
maxandersen: ooooh - I misread the diff. you are right. your fix is correct.

maxandersen (approved):

PR #5565, opened 2019-11-18
Check if format is null, if so default to \'name\'.

stalep:


independent-projects/tools/common/src/main/java/io/quarkus/cli/commands/ListExtensions.java
```diff
@@ -33,6 +33,8 @@ public ListExtensions(final BuildFile buildFile) throws IOException {
 
     public void listExtensions(boolean all, String format, String search) throws IOException {
         final Map<String, Dependency> installed = findInstalled();
+        if(format == null)
+            format = "name";
```
gsmet: Could you add curly braces, please?

gsmet:
@aloubyansky I let you review that one.

@stalep when is this one an issue?

stalep:
I had some time to work on the CLI and when I played with the commands I just noticed this. It didn't used to require that format was set. I can fix it in the command itself, but IMO it's good to check.

independent-projects/tools/common/src/main/java/io/quarkus/cli/commands/ListExtensions.java
aloubyansky: Shouldn't it be "concise", though? Just looking at the switch/default.

stalep:
I do need to add another method to this class as it's limiting to not have a method that returns the list of extensions as well. ATM there is only this method that prints out the extensions to std out. In the CLI you would often want a different formatting.

stalep:
Closing this after talking with @aloubyansky. I'll base the latest CLI work on his incoming changes.

Issue #2746, opened 2019-06-06
Consider adding src/main/resources to native image


geoand:
**Description**
It seems to come up somewhat often that application developers want to include the contents of `src/main/resources` into their native image. 
Currently we don't provide an easy way of doing that, other making the user add `additionalBuildArgs`.
This has been discussed before in #388, but I believe that this need is not limited to extension authors only (see for example [this](https://quarkusio.zulipchat.com/#narrow/stream/187030-users/topic/filenotfoundexception) discussion, or envision a scenario were a use developing a Kafka application wants to read from Twitter and includes `twitter4j.properties`).

**Implementation ideas**
My idea would be to follow the everything configured via `application.properties` paradigm. We could provide some properties that the user could set which would then be easily processed by by a processor to generate the necessary `SubstrateResourceBuildItem`.

ChristianHuff-DEV:
@geoand Could you give an example on how to add a resource using `additionalBuildArgs`. Based on #388 and the [substratevm documentation](https://github.com/oracle/graal/blob/master/substratevm/RESOURCES.md), I haven't been able to figure it out.

I'm using Quarkus *0.18.0* with GraalVM *ce-1.0.0-rc16*.

I added a file called *version.txt* to the `src/main/resources` folder (same folder as *application.properties*). In my code I access this file using `File file = new File(getClass().getClassLoader().getResource("version.txt").getFile());`. This works as expected. But as soon as I build a native image a `NullPointerException` is thrown.

My build plugins configuration looks like this:
```
          <plugin>
            <groupId>io.quarkus</groupId>
            <artifactId>quarkus-maven-plugin</artifactId>
            <version>${quarkus.version}</version>
            <executions>
              <execution>
                <goals>
                  <goal>native-image</goal>
                </goals>
                <configuration>
                  <enableHttpUrlHandler>true</enableHttpUrlHandler>
                  <additionalBuildArgs>-H:Log=registerResource:</additionalBuildArgs>
                  <additionalBuildArgs>-H:IncludeResources=.*/version.*txt$</additionalBuildArgs>
                </configuration>
              </execution>
            </executions>
          </plugin>
```

(I also tried `.*/version.txt$` for the regex.)

Can you spot what I am doing wrong? Do you have a complete example I can have a look at?

geoand:
Hello @ChristianHuff-DEV,

I would advise you to take a look at my comment [here](https://github.com/quarkusio/quarkus/pull/2910#issuecomment-504671096).

[...]

PR #5566, opened 2019-11-18
Place json configuration files next to thin jar when building

### LINKED ISSUE #5390: Adding a reflection configuration file fails when building native using Docker
We are currently trying to update an old 0.21.2 Quarkus Project to 0.28.1. In this project we used a reflect-config.json to keep classes from being deleted. With 0.21.2 we could successfully build a native image and a docker container, but now we can't.

We already changed to quarkus.native.additional-build-args in the application.properties file with these parameters: 
```
quarkus.native.additional-build-args=\
-H:ResourceConfigurationFiles=resources/main/resource-config.json,\
-H:ReflectionConfigurationFiles=resources/main/reflect-config.json

quarkus.native.enable-jni=true
quarkus.native.enable-http-url-handler=true
quarkus.native.enable-https-url-handler=true
quarkus.native.enable-all-security-services=true
```

Trying to execute the build ends in an error that he can't find project/resources/main/reflect-config.json.

What Quarkus tries to execute is this: 
`docker run -v /home/bfr/git/minical-multi/minical/build/minical-0.2.0-native-image-source-jar:/project:z --user 1000:1001 --rm quay.io/quarkus/ubi-quarkus-native-image:19.2.1 -J-Dsun.nio.ch.maxUpdateArraySize=100 -J-Djava.util.logging.manager=org.jboss.logmanager.LogManager -J-Dvertx.logger-delegate-factory-class-name=io.quarkus.vertx.core.runtime.VertxLogDelegateFactory -J-Dvertx.disableDnsResolver=true -J-Dio.netty.leakDetection.level=DISABLED -J-Dio.netty.allocator.maxOrder=1 -H:ResourceConfigurationFiles=resources/main/resource-config.json -H:ReflectionConfigurationFiles=resources/main/reflect-config.json --initialize-at-build-time= -H:InitialCollectionPolicy=com.oracle.svm.core.genscavenge.CollectionPolicy$BySpaceAndTime -jar minical-0.2.0-runner.jar -J-Djava.util.concurrent.ForkJoinPool.common.parallelism=1 -H:FallbackThreshold=0 -H:+ReportExceptionStackTraces -H:-AddAllCharsets -H:EnableURLProtocols=http,https --enable-all-security-services -H:-SpawnIsolates -H:+JNI --no-server -H:-UseServiceLoaderFeature -H:+StackTrace minical-0.2.0-runner`

In the earlier versions of quarkus just the build directory was bound to docker as a volume and all files were found correctly, but now it is only the native-image-source-jar and we can't figure out a way to get the reflect-config.json into it, because always on build the directory is new generated and deletes the already copied file. 

The reflect-config.json itself is in the .jar file and can't also be accessed. Is there a way to do it, like change the volume path for the `docker run` or any other way we oversaw?

We are using gradle to build our project.

geoand:
This makes it very easy and predicable to configure
add extra json files to configure GraalVM natives.
Furthermore, this makes these files available to
docker when a container build of the native image
is being used - something which fails without this change.

Fixes: #5390

### DIFF
--- a/core/deployment/src/main/java/io/quarkus/deployment/pkg/steps/JarResultBuildStep.java
+++ b/core/deployment/src/main/java/io/quarkus/deployment/pkg/steps/JarResultBuildStep.java
@@ -10,6 +10,7 @@ import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
+import java.io.UncheckedIOException;
 import java.nio.file.FileAlreadyExistsException;
 import java.nio.file.FileSystem;
 import java.nio.file.FileVisitOption;
@@ -30,6 +31,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.function.BiPredicate;
 import java.util.function.BooleanSupplier;
 import java.util.function.Consumer;
 import java.util.jar.Attributes;
@@ -337,6 +339,7 @@ public class JarResultBuildStep {
                 .resolve(outputTargetBuildItem.getBaseName() + "-native-image-source-jar");
         IoUtils.recursiveDelete(thinJarDirectory);
         Files.createDirectories(thinJarDirectory);
+        copyJsonConfigFiles(applicationArchivesBuildItem, thinJarDirectory);
 
         Path runnerJar = thinJarDirectory
                 .resolve(outputTargetBuildItem.getBaseName() + packageConfig.runnerSuffix + ".jar");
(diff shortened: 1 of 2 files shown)


core/deployment/src/main/java/io/quarkus/deployment/pkg/steps/JarResultBuildStep.java
```diff
@@ -358,6 +360,34 @@ public NativeImageSourceJarBuildItem buildNativeImageJar(CurateOutcomeBuildItem
...
+            public boolean test(Path path, BasicFileAttributes basicFileAttributes) {
+                return basicFileAttributes.isRegularFile() && path.toString().endsWith(".json");
+            }
+        }).forEach(new Consumer<Path>() {
+            @Override
+            public void accept(Path jsonPath) {
+                try {
+                    Files.copy(jsonPath, thinJarDirectory.resolve(jsonPath.getFileName()));
+                } catch (IOException e) {
+                    throw new RuntimeException("Unable to copy json config file from " + jsonPath + " to " + thinJarDirectory,
```
gastaldi: `throw new UncheckedIOException(..., e )`

geoand:
I could add tests for these if you feel it's warranted, but they would of course slow down the build a lot.
I tested by creating a maven and gradle project then adding a json config file for GraalVM and finally building using both regular and docker builds.

docs/src/main/asciidoc/writing-native-applications-tips.adoc
```diff
@@ -58,37 +86,23 @@ Now, let's tell the `native-image` executable about this configuration file by t
...
-
-When using Gradle, we need to configure the additional build arguments in the plugin configuration:
+When using Gradle, configuring the `buildNative` task like so suffices:
 
 [source,groovy]
 ----
 buildNative {
     additionalBuildArgs = [
-            "-H:ResourceConfigurationFiles=${buildDir}/resources/main/resources-config.json".toString()
+            'resources-config.json'
```
gastaldi: Don't you need to still specify the `-H:ResourceConfigurationFiles` here?

gastaldi (commented):
Added some comments

docs/src/main/asciidoc/writing-native-applications-tips.adoc
geoand: Oops!

core/deployment/src/main/java/io/quarkus/deployment/pkg/steps/JarResultBuildStep.java
geoand: Fixed, thanks!

docs/src/main/asciidoc/writing-native-applications-tips.adoc
geoand: Fixed

gastaldi (approved):
LGTM

geoand:
@gsmet are you ok with this one?

stuartwdouglas:
I don't really like this solution. Even though it improves the status quo I still don't think it provides a great user experience. 

Maybe the approach should be: 
- Everything in the application (src/main/resources) is included by default (otherwise why would it be in the application if it was not needed?)
- Additional resources can be included through either an annotation that specifies the paths to include, or a config file in a well known location (quarkus-resources.properties?) that we parse at build time and then automatically handle. 
- Maybe the ability to include resources via application.properties, but I don't know if this is really config

I think we should discuss this, as the user should not have to mess around with JSON files or the substrate command line to include some resources in their image.

geoand:
@stuartwdouglas you have my vote when it comes to adding everything in `src/main/resources`, I had even opened an issue about it: https://github.com/quarkusio/quarkus/issues/2746.

This PR didn't really attempt to solve that problem generally. It was motivated primarily by the fact that currently one can't due a native build using docker and specify the json config files

geoand:
Even if we do add some dedicated configuration options for resource and reflection handling (plus whatever other json file could be added), we'll still need the copy mechanics of this PR.

I do like the idea of having dedicated `quarkus.native.reflection-config-files` or something like that to users can just specify what they want to include.

PR #5577, opened 2019-11-19
Updates of rest-assured, assertj

rsvoboda:
Updates of rest-assured, assertj and jetty


stuartwdouglas:
Why do we even have Jetty? I think it may be a leftover from when Camel was part of the main code base.

rsvoboda:
I think I saw that in TCK modules, let me check

rsvoboda:
got confused, tck module is not using it, it is excluding it
https://github.com/quarkusio/quarkus/blob/master/tcks/microprofile-rest-client/pom.xml#L98-L110

I will update the pr to skip jetty for now, will prepare dedicated pr for jetty removal

stuartwdouglas:
Thanks

stuartwdouglas (approved):

PR #5580, opened 2019-11-19
Docs Reactive Routes - minor updates/fixes

mkouba:
- use RoutingExchange
- @Route is repeatable


mkouba:
While reading the guide I also wonder if it would make sense to introduce `@RouteFilter` to simplify the `Filters` definition. It's probably a very low priority though.

cescoffier (approved):

cescoffier:
LGTM, about `@RouteFilter`, why not, we could add it.

PR #5581, opened 2019-11-19
Document replacement for HttpServletRequest

### LINKED ISSUE #4125: Provide a replacement for injecting HttpServletRequest
When using RESTEasy without Undertow, we can't inject `HttpServletRequest` so we need a replacement for it (and all the other servlet related injections).

A good example of that is the REST JSON quickstart: https://github.com/quarkusio/quarkus-quickstarts/blob/master/rest-json/src/main/java/org/acme/rest/json/LoggingFilter.java .

I think we need something for 0.24.0 as adding Undertow in the quickstart won't really help the case of RESTEasy without Undertow as it's our basic example of a REST service.

/cc @patriot1burke @stuartwdouglas

FroMage:
Fixes #4125


gsmet (approved):

gsmet:
Merged, thanks!

PR #5583, opened 2019-11-19
Provides lock support for Hibernate with Panache

### LINKED ISSUE #2744: LockModeType support with Panache
**Description**
Panache in both mode (Entity and Repository) don't have primary support for JPA LockModeType. So if we want to implements a lock mechanism we must inject the EntityManager and use the lock method on the entity.
This is not very convenient and moreover there is no way to make a lock without first retrieving the entity so hitting the database 2 times.

**Implementation ideas**
I have a prototype in progress to add the following methods ;
- PanageEntityBase::lock(LockModeType)
- PanageEntityBase::findById(Id, LockModeType)
- PanageRepositoryBase::lock(Entity, LockModeType)
- PanageRepositoryBase::findById(Id, LockModeType)

I can share my work in progress in a PR, but I still have some issue with the findById methods, they generate a core dump at runtime (still need to understand the code generation code based on ASM).

loicmathieu:
Fixes #2744 

cc @emmanuelbernard and @FroMage

### DIFF
--- a/docs/src/main/asciidoc/hibernate-orm-panache.adoc
+++ b/docs/src/main/asciidoc/hibernate-orm-panache.adoc
@@ -412,55 +412,45 @@ public void create(Parameter parameter){
 
 == Lock management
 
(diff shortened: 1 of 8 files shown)


FroMage (approved):
LGTM

PR #5587, opened 2019-11-19
chore: Option -DskipTests also covers maven invoker tests.

iocanel:
Currently when using `-DskipTests` all tests are excluded.
All?
Not all, there are still two modules that are using `maven-invoker-plugin`. These integration tests do run even when `-DskipTests` is used.

This pull request changes the `maven-invoker-plugin` coniguration so that invoker tests are skiped when using `-DskipTests`.


geoand (approved):
I tried it and it worked great!

gsmet (approved):
Woot! I wanted that for a while!

PR #5598, opened 2019-11-19
Fix PanacheQuery.firstResult()

### LINKED ISSUE #5589: PanacheQueryImpl.firstResult() retrieves more than one item from the database
**Describe the bug**
Reading the code of `PanacheQueryImpl.firstResult() ` it seems to retrieves more than one item from the database.

```
@Override
    public <T extends Entity> T firstResult() {
        List<T> list = list();
        jpaQuery.setMaxResults(1);
        return list.isEmpty() ? null : list.get(0);
    }
```

As the `setMaxQueryResult` is called after listing the results, it is useless and the query will be sent to the database with the previously set max result (Integer.MAX_VALUE by default)

**Expected behavior**
`firstResult()` retrieve one or zero item from the database.

 - Quarkus version or git rev: 1.0.0.CR1 and current master

**Additional context**
If the bug is aknowledge, I can propose a PR to fix it.

loicmathieu:
Fixes #5589

PanacheQuery.firstResult() calls `setMaxResults` too late so it is not taken into account


FroMage (approved):
Ooops.

PR #5605, opened 2019-11-19
Set the lower boundary of the default range for Quarkus platforms to 1.0.0.CR2

aloubyansky:
Due to recently introduced changes in the project templates used for project generation and updates, it'll be safe to move the lower boundary of the default range of Quarkus platforms discovered by the tooling to the upcoming version 1.0.0.CR2.
Note that 1.0.0.CR2 has not yet been released, so the SNAPSHOT version of the tooling will now be falling back to the bundled version of the platform (which is based on quarkus-bom, SNAPSHOT) until 1.0.0.CR2 has been released and available in the remote repos.


gsmet (approved):

gsmet:
Merging per our discussion.

Issue #5588, opened 2019-11-19
Native image gets generated even when all classes are unchanged / up to date


rsvoboda:
Native image gets generated even when all classes unchanged / are up to date
maven-compiler-plugin detects that case, quarkus-maven-plugin always runs the build

Discussed in https://quarkusio.zulipchat.com/#narrow/stream/187038-dev/topic/skip.20native.20%28re%29build.2C.20run.20just.20test/near/181113795

```
mvn io.quarkus:quarkus-maven-plugin:999-SNAPSHOT:create \
    -DprojectGroupId=org.acme \
    -DprojectArtifactId=getting-started \
    -DclassName="org.acme.quickstart.GreetingResource" \
    -Dpath="/hello"
```
```
mvn verify -Dnative  # runs the native compilation
mvn verify -Dnative  # second run should not run the native compilation but it runs it
```

mkouba:
For the record - the fact that no class has changed does not imply that the native image generation should be skipped - for example you can add extesions, change the configuration, resource files, etc.

It's possible to skip the native image manually. Which is not very practical but sometimes it could be useful ;-)

- configure the `maven-failsafe-plugin` outside the "native" profile and skip the "native" profile activation for the second run:
   ```
   mvn verify -Dnative
   # This one will reuse the existing native image
   mvn verify
   ```
* configure the `maven-failsafe-plugin` and do not use the "native" profile at all:
   ```
   mvn verify -Dquarkus.package.type=native
   mvn verify
   ```

rsvoboda:
> you can add extesions, change the configuration, resource files

dev mode can detect these things, standard quarkus build could leverage these capabilities too

> workaround with failsafe

I'd rather have `mvn verify -Dquarkus.native.skip-build=true` for better out-of-the-box experience

gsmet:
I don’t see how that could become a priority anytime soon.

rsvoboda:
I think https://github.com/quarkusio/quarkus/pull/5609 addresses this quite well

Native image gets built only once when following commands are used
```
mvn verify -Dnative
mvn failsafe:integration-test
```

rsvoboda:
@mkouba / @gsmet I think this can be closed, 5609 is good enough for me.

PR #5609, opened 2019-11-19
Allow testing of an existing executable

stuartwdouglas:


test-framework/common/src/main/java/io/quarkus/test/common/NativeImageLauncher.java
```diff
@@ -109,11 +109,24 @@ private static String guessPath(Class<?> testClass) {
                             return file.getAbsolutePath();
                         }
                     }
+                } else if (url.getProtocol().equals("file") && url.getPath().contains("/target/surefire/")) {
```
geoand: Will these paths work on Windows?

stuartwdouglas: I am not sure, it is a URL path so I think so, either way windows support for native is pretty limited at the moment AFAIK, and the runner would probably have a .exe extension so that would not work either.

geoand: Makes sense

geoand (approved):

gsmet (dismissed):
Please let @rsvoboda test this one before merging because he was the one asking.

@rsvoboda can you check that it solves your initial complaint?

Thanks!

rsvoboda:
Sure, but need to tackle wf/eap stuff first. Will look into it today for sure

rsvoboda (commented):
I did `mvn verify -Dnative && mvn failsafe:integration-test` and native image got compiled only once !

Small disadvantage is that 3.0.0-M4 gets used - `[INFO] --- maven-failsafe-plugin:3.0.0-M4:integration-test (default-cli) @ getting-started ---`, pom.xml uses 2.22.0. 

`mvn org.apache.maven.plugins:maven-failsafe-plugin:2.22.0:integration-test` is workaround for this.

Also tried to modify one test, changes were reflected.

One thing which I didn't like are these super long `=` lines:
```
[INFO] Running org.acme.quickstart.NativeGreetingResourceIT
=================================================================================================================================================================
=native.image.path was not set, making a guess that  /Users/rsvoboda/tmp/getting-started/target/getting-started-1.0-SNAPSHOT-runner is the correct native image=
=================================================================================================================================================================
```
But it's not related to this change

rsvoboda (approved):

mkouba:
> Small disadvantage is that 3.0.0-M4 gets used - [INFO] --- maven-failsafe-plugin:3.0.0-M4:integration-test (default-cli) @ getting-started ---, pom.xml uses 2.22.0.

Ok, that's because your profile defines a different version of failsafe, right? In that case, I'd prefer to define the failsafe version outside the native profile. However, in that case you wouldn't need to run `mvn verify -Dnative && mvn failsafe:integration-test` because `mvn verify -Dnative && mvn verify` would work as well (https://github.com/quarkusio/quarkus/issues/5588#issuecomment-555558300). Or am I missing something obvious?

Issue #5601, opened 2019-11-19
Expose mini extension APIs for applications


paulrobinson:
### Description
Allow applications write build-time operations or resource enlistment for GraalVM without having to use the nasty command line arguments or write a full extension.

### Analysis
_Links to analysis docs (E.g Design, Requirements, etc)_

### Tasks
- [ ] Support for reflective hierarchies
- [ ] Support for Runtime Initialised classes
- [ ] Support for Runtime Reinitialised classes
- [ ] Support for Service providers
- [ ] Support for defining JDK proxy classes
- [ ] Support for including resource bundles
- [ ] Better application resource handling

n1hility:
retiring issue

Issue #2277, opened 2019-04-29
Improve extension author guide


emmanuelbernard:
- [ ] Describe where to put APIs (runtime, deployment etc?)
- [ ] Describe where Substitutions and templates should live
- [ ] Dealing with META-INF/services files https://github.com/quarkusio/quarkus/issues/2299
- [ ] Offer a getting started section to ease the entry into Extension writing https://github.com/quarkusio/quarkus/pull/2406

emmanuelbernard:
See also https://github.com/quarkusio/quarkus/pull/2406 for an easier guide start

cmoulliard:
Here is a new list of things to be improved after spending 2 days to try to create a simple Servlet Extension - https://github.com/cmoulliard/quarkus-my-ext

To Do

- [ ] Update the maven archetype to create an extension as it is outdated
- [ ] Review Guide in order to avoid to have to create the extension under the quarkus/extensions project
- [ ] Add Junit test to show how to test deployment/runtime code
- [ ] Explain which BOM o be used and dependencies to include to register servlet, ... on undertow
- [ ] Explain when we must use @Recorder, @record
- [ ] Better introduce the role of the `deployment` and runtime` module, how the code is build and next used by a Quarkus application
- [ ] Do we need the runtime module for all the extensions ? If yes -> Why and when we should use it

PR #5610, opened 2019-11-20
Initial work on native user API

stuartwdouglas:
First part of https://github.com/quarkusio/quarkus/issues/5601

### DIFF
--- a/core/runtime/src/main/java/io/quarkus/runtime/annotations/RuntimeInitialized.java
+++ b/core/runtime/src/main/java/io/quarkus/runtime/annotations/RuntimeInitialized.java
@@ -1,0 +1,25 @@
+package io.quarkus.runtime.annotations;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Annotation that can be used to force a class to be initialized at runtime
+ * when using native image mode. This means that any static init code will be run
+ * at startup, rather than when the image is built.
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.TYPE)
+public @interface RuntimeInitialized {
+
+    /**
+     * Alternative classes that should actually be initialized at runtime instead of the current class.
+     *
+     * This allows for classes in 3rd party libraries to be initialized at runtime without modification
+     * or writing an extension. If this is set then the class it is placed on is not initialized at runtime,
+     * so this should generally just be placed on an empty class that is not otherwise used.
+     */
+    Class<?>[] targets() default {};
+}
--- a/docs/src/main/asciidoc/writing-native-applications-tips.adoc
+++ b/docs/src/main/asciidoc/writing-native-applications-tips.adoc
@@ -20,6 +20,14 @@ Please refer to the appropriate section depending on your context.
 
 GraalVM imposes a number of constraints and making your application a native executable might require a few tweaks.
 
+=== Quarkus Annotations
+
+This guide covers some of the annotations that Quarkus supports to help generate native images. In some cases these
+annotations are applied directly to a relevant class, but in others the class that they are placed on does not actually
+matter (e.g. if you are referencing a class in a 3rd party jar). In this case we recommend creating an empty `NativeConfig`
+class somewhere in your application and placing all annotation that control native image generation on it, so they are all
+in the same place.
+
 === Including resources
 
 By default, when building a native executable, GraalVM will not include any of the resources that are on the classpath into the native executable it creates.
@@ -279,7 +287,7 @@ buildNative {
 }
 ----
 
-=== Delaying class initialization
+=== Delaying and re-running class initialization
 
 By default, Quarkus initializes all classes at build time.
 
@@ -294,9 +302,14 @@ method com.amazonaws.services.s3.model.CryptoConfiguration.<init>(CryptoMode)
 Call path from entry point to com.amazonaws.services.s3.model.CryptoConfiguration.<init>(CryptoMode):
 ----
 
-If you need to delay the initialization of a class, you can use the `--initialize-at-run-time=<package or class>` configuration knob.
+If you need to delay the initialization of a class, you can use the `@io.quarkus.runtime.annotations.RuntimeInitialized` annotation
+to indicate that a class should be initialized at runtime. If the class cannot be modified (e.g. it is part of a 3rd party jar)
+then you can place this annotation on a different class in your application, and use the `targets` member to indicate the
+actual classes that should be runtime initialized.
 
-It should be added to the `native-image` configuration using an `<additionalBuildArg>` as shown in the examples above.
+Native mode also supports the notion of 'runtime reinitialization' where static init is run at image generation time, and
+then re-run at startup. This can be done via the `@io.quarkus.runtime.annotations.RuntimeReinitialized` annotation in a
+similar manner to above.
 
 [NOTE]
 ====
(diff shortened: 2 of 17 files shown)


mkouba:
Do you plan on adding a test for `@RegisterResourceBundle`? Otherwise looks very good!

core/runtime/src/main/java/io/quarkus/runtime/annotations/RuntimeInitialized.java
```diff
@@ -0,0 +1,25 @@
...
+import java.lang.annotation.Target;
+
+/**
+ * Annotation that can be used to force a class to be initialized at runtime
+ * when using native image mode. This means that any static init code will be run
+ * at startup, rather than when the image is built.
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.TYPE)
+public @interface RuntimeInitialized {
```
gsmet: It might be a silly question and I'm not totally convinced myself but should we stick to the `Register` prefix?

docs/src/main/asciidoc/writing-native-applications-tips.adoc
```diff
@@ -20,6 +20,14 @@ Please refer to the appropriate section depending on your context.
 
 GraalVM imposes a number of constraints and making your application a native executable might require a few tweaks.
 
+=== Quarkus Annotations
+
+This guide covers some of the annotations that Quarkus supports to help generate native images. In some cases these
+annotations are applied directly to a relevant class, but in others the class that they are placed on does not actually
+matter (e.g. if you are referencing a class in a 3rd party jar). In this case we recommend creating an empty `NativeConfig`
+class somewhere in your application and placing all annotation that control native image generation on it, so they are all
```
gsmet:
```suggestion
class somewhere in your application and placing all annotations that control native image generation on it, so they are all
```

docs/src/main/asciidoc/writing-native-applications-tips.adoc
```diff
@@ -20,6 +20,14 @@ Please refer to the appropriate section depending on your context.
...
+in the same place.
```
gsmet: I wonder if we should have a table here summarizing the annotations so that the user has a global overview of what's possible.

gsmet (changes requested):
Definitely a very nice user experience improvement.

I added a few comments. Only the typo is a must fix, the rest is really to be discussed.

docs/src/main/asciidoc/writing-native-applications-tips.adoc
```diff
@@ -183,9 +191,25 @@ public class MyClass {
...
+that you actually want to register in the `targets` member of the annotation:
+
+[source,java]
+----
+@RegisterForReflection(targets={ThirdPartyClass.class, OtherClass.class})
+public class EmptyClass {
+}
+----
+
+In this example only the `ThirdPartyClass` and `ThirdPartyClass` other class are registered, `EmptyClass` is ignored.
```
emmanuelbernard: I find that a bit counter intuitive. To be fair the whole annotation approach feels counter intuitive. I wonder if we could allow the class enlistment via an API and annotated toe block of code to be called.

```
@QuarkusBuild
public void buildOperations(QuarkusBuildOperations ops) {
    ops.enlistClassForReflecton(ThirdPartyClass.clas);
    ops.otherOperations(...);
}
```

docs/src/main/asciidoc/writing-native-applications-tips.adoc
```diff
@@ -20,6 +20,14 @@ Please refer to the appropriate section depending on your context.
 
 GraalVM imposes a number of constraints and making your application a native executable might require a few tweaks.
 
+=== Quarkus Annotations
```
emmanuelbernard: Let's mark new features as `Preview` in both docs and APIs/Annotations until we are happy we will support them for the while 1.x series.

docs/src/main/asciidoc/writing-native-applications-tips.adoc
```diff
@@ -278,9 +302,14 @@ method com.amazonaws.services.s3.model.CryptoConfiguration.<init>(CryptoMode)
 Call path from entry point to com.amazonaws.services.s3.model.CryptoConfiguration.<init>(CryptoMode):
 ----
 
-If you need to delay the initialization of a class, you can use the `--initialize-at-run-time=<package or class>` configuration knob.
+If you need to delay the initialization of a class, you can use the `@io.quarkus.runtime.annotations.RuntimeInitialized` annotation
```
emmanuelbernard: Same here for API vs annotation.
Also we should explain why a user would want to do that.

docs/src/main/asciidoc/writing-native-applications-tips.adoc
```diff
@@ -278,9 +302,14 @@ method com.amazonaws.services.s3.model.CryptoConfiguration.<init>(CryptoMode)
...
+to indicate that a class should be initialized at runtime. If the class cannot be modified (e.g. it is part of a 3rd party jar)
+then you can place this annotation on a different class in your application, and use the `targets` member to indicate the
+actual classes that should be runtime initialized.
 
-It should be added to the `native-image` configuration using an `<additionalBuildArg>` as shown in the examples above.
+Native mode also supports the notion of 'runtime reinitialization' where static init is run at image generation time, and
+then re-run at startup. This can be done via the `@io.quarkus.runtime.annotations.RuntimeReinitialized` annotation in a
```
emmanuelbernard: Explain why you would want to do that. It's dry otherwise.

emmanuelbernard (changes requested):

docs/src/main/asciidoc/writing-native-applications-tips.adoc
```diff
@@ -183,9 +191,25 @@ public class MyClass {
...
 ----
 
-==== Using a configuration file
+If you cannot modify the class (e.g. it is in a 3rd party jar) then you can instead specify the classes
+that you actually want to register in the `targets` member of the annotation:
+
+[source,java]
+----
+@RegisterForReflection(targets={ThirdPartyClass.class, OtherClass.class})
+public class EmptyClass {
```
FroMage: Shouldn't it be `NativeConfig` then?

docs/src/main/asciidoc/writing-native-applications-tips.adoc
FroMage: Grouping those operations into a single interface looks like a nice idea.

core/runtime/src/main/java/io/quarkus/runtime/annotations/RuntimeInitialized.java
stuartwdouglas: I don't know, I did think about it but I think it reads better this way. 'RegisterAsRuntimeInitialized' could be ok as well.

docs/src/main/asciidoc/writing-native-applications-tips.adoc
stuartwdouglas: Once we have decided on the API I don't think there is any issue supporting these, they are just a new facade on functionality we already have.

core/runtime/src/main/java/io/quarkus/runtime/annotations/RuntimeInitialized.java
stuartwdouglas: Actually I wonder if all of these should go in a 'native' package, although it would cause problems with the existing RegisterForReflection.

docs/src/main/asciidoc/writing-native-applications-tips.adoc
stuartwdouglas: This is actually pretty common in Jakarta EE (annotations that are not directly applicable to the class they are placed on). Some examples:

CDI response producer fields:

````
@Produces
@EntityManager
EntityManager entityManger;
````

This enables EntityManager injection in you application, but it can be literally anywhere and the class it is on makes no difference. Same with things like @DataSourceDefinition.

I do agree that it is a bit weird but your solution does not solve the problem of code affecting the whole app that could be anywhere.

stuartwdouglas: Also we would need both, as for the cases where the class is in your application then an annotation is the easiest to understand.

emmanuelbernard:
When I looked at the list of subjects, I think while there is value in this extension. It might actually be moot or less important if we really manage to improve https://github.com/quarkusio/quarkus/issues/2277

stuartwdouglas:
@emmanuelbernard do we still want this? If not we should close the PR.

emmanuelbernard:
We want it in some form at some point. Whether I would spend our energy on in in the next 2/3 releases, I don't think so.

PR #5611, opened 2019-11-20
Add more tests for Spring Data JPA

geoand:
These tests where inspired from
https://dzone.com/articles/spring2quarkus-spring-boot-to-quarkus-migration
which mentions some issues (that are no longer present)


gsmet (approved):
More tests passing is always good, let's merge that one.

PR #5618, opened 2019-11-20
Reactive routes - declarative way of defining Filter

### LINKED ISSUE #5591: Reactive routes - declarative way of defining Filter
Currently, it's possible to declare an observer method and register an instance of `Filter` - see https://quarkus.io/guides/reactive-routes#intercepting-http-requests. It could be useful to declare a filter method like this:
```java
public class MyFilters {

    @RouteFilter(100)
    void filter(RoutingContext rc) {
        rc.response().putHeader("X-Header", "intercepting the request");
        rc.next();
    }
}
```

mkouba:
- resolves #5591

### DIFF
--- a/extensions/vertx-web/runtime/src/main/java/io/quarkus/vertx/web/RouteFilter.java
+++ b/extensions/vertx-web/runtime/src/main/java/io/quarkus/vertx/web/RouteFilter.java
@@ -1,0 +1,28 @@
+package io.quarkus.vertx.web;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import io.vertx.ext.web.RoutingContext;
+
+/**
+ * This annotation is used to define a "filter", i.e. a route called on every HTTP request.
+ * <p>
+ * The target business method must return {@code void} and accept exactly one argument of type {@link RoutingContext}.
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.METHOD)
+public @interface RouteFilter {
+
+    static final int DEFAULT_PRIORITY = 10;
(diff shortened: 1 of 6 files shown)


extensions/vertx-web/runtime/src/main/java/io/quarkus/vertx/web/RouteFilter.java
```diff
@@ -0,0 +1,28 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import io.vertx.ext.web.RoutingContext;
+
+/**
+ * This annotation is used to define a "filter", i.e. a route called on every HTTP request.
+ * <p>
+ * The target business method must return {@code void} and accept exactly one argument of type {@link RoutingContext}.
```
cescoffier: Might be useful to write a few words on the default priority. WDYT?

cescoffier (changes requested):
I've just added a comment about the default priority.

extensions/vertx-web/runtime/src/main/java/io/quarkus/vertx/web/RouteFilter.java
mkouba: Hm, I don't know what to say. `FilterBuildItem` has some pre-defined system priorities but I can't reference them in the runtime module. I chose `10` so that it's called after these system filters...

cescoffier: Maybe just mentioned that the default priority is 10.

mkouba: Ok, will do.

cescoffier (approved):

PR #5620, opened 2019-11-20
Avoid creating redundant instances of SecurityCheck implementations

geoand:
The reason we want to reuse `RolesAllowedCheck` is that it is very common
to have a lot of methods using the same roles in the security check.
In such cases there is no need to have multiple instances of the class hanging around
for the entire lifecycle of the application.

### DIFF
--- a/extensions/security/runtime/src/main/java/io/quarkus/security/runtime/interceptor/check/RolesAllowedCheck.java
+++ b/extensions/security/runtime/src/main/java/io/quarkus/security/runtime/interceptor/check/RolesAllowedCheck.java
@@ -1,16 +1,39 @@
 package io.quarkus.security.runtime.interceptor.check;
 
+import java.util.Arrays;
+import java.util.List;
+import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.Function;
 
 import io.quarkus.security.ForbiddenException;
 import io.quarkus.security.UnauthorizedException;
 import io.quarkus.security.identity.SecurityIdentity;
 
 public class RolesAllowedCheck implements SecurityCheck {
+
+    /*
+     * The reason we want to cache RolesAllowedCheck is that it is very common
+     * to have a lot of methods using the same roles in the security check
+     * In such cases there is no need to have multiple instances of the class hanging around
+     * for the entire lifecycle of the application
+     */
+    private static final Map<List<String>, RolesAllowedCheck> CACHE = new ConcurrentHashMap<>();
+
     private final String[] allowedRoles;
 
-    public RolesAllowedCheck(String[] allowedRoles) {
+    private RolesAllowedCheck(String[] allowedRoles) {
         this.allowedRoles = allowedRoles;
+    }
+
+    public static RolesAllowedCheck of(String[] allowedRoles) {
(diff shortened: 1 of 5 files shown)


geoand:
Mongo error is unrelated - it seemed like MongoDB didn't come up in time. I'll rerun the failing test once all others complete

gsmet (approved):

extensions/security/runtime/src/main/java/io/quarkus/security/runtime/interceptor/check/RolesAllowedCheck.java
```diff
@@ -1,18 +1,41 @@
...
+
     private final String[] allowedRoles;
 
-    public RolesAllowedCheck(String[] allowedRoles) {
+    private RolesAllowedCheck(String[] allowedRoles) {
         this.allowedRoles = allowedRoles;
     }
 
+    public static RolesAllowedCheck of(String[] allowedRoles) {
+        return CACHE.computeIfAbsent(Arrays.asList(allowedRoles), new Function<List<String>, RolesAllowedCheck>() {
```
gsmet: Just so you know, if you expect a lot of hits, it's significantly faster to check the key first with a simple `get()`.

I don't think it's worth the hassle here so let's go with this.

gsmet (approved):

extensions/security/runtime/src/main/java/io/quarkus/security/runtime/interceptor/check/RolesAllowedCheck.java
geoand: Ah yes! I remember you mentioning this in another context! I'll keep it in mind, thanks!

michalszynkiewicz: @geoand do you sort the roles anywhere? Otherwise `@RolesAllowed("a", "b")` would be treated as different than `@RolesAllowed("b", "a")` and we can do better ;)

gsmet: Ah ah I thought about it too but decided to not bother you with this but if we are two, let's be picky :). Yeah, it would sure be a nice improvement. Maybe, use a `Set` instead of a `List`. I suppose a `HashSet` would be good enough.

geoand: Good idea, I didn't think about that one. Would you like to add new PR @michalszynkiewicz?

gsmet: And if we want to go into the micro-optimization world: there's a good chance that for a lot of cases, you will have only one role so you could use a singleton collection in this case.

michalszynkiewicz: I dont have time for that now, unfortunately

geoand: Here is it: https://github.com/quarkusio/quarkus/pull/5643

PR #5628, opened 2019-11-20
RegisterRestclient configkey: improve error message and add documentation

### LINKED ISSUE #5619: Unable to use @RegisterRestClient(configKey = "some.config.key")
**UPDATE: this is not a bug, only the error message is misleading.**

Unable to use @RegisterRestClient(configKey = "some.config.key") ot have a more friendly configuration key for the rest client URL.

**Describe the bug***

When defining a rest client with `@RegisterRestClient(configKey = "some.config.key")`
```
@Path("/v1/accounts")
@RegisterRestClient(configKey = "account-api.base-uri")
@Produces("application/json")
@Consumes("application/json")
public interface AccountRestClient {
    [...]
}
```

And having the following line in my `application.properties`
```
account-api.base-uri=http://localhost:8080
```

It fails with the following error:
```
[...]
Caused by: java.lang.IllegalArgumentException: Unable to determine the proper baseUrl/baseUri. Consider registering using @RegisterRestClient(baseUri="someuri", configKey="orkey"), or by adding 'account-api.base-uri' to your Quarkus configuration
	at io.quarkus.restclient.runtime.RestClientBase.configureBaseUrl(RestClientBase.java:191)
	at io.quarkus.restclient.runtime.RestClientBase.create(RestClientBase.java:52)
```

**Expected behavior**
The MP Rest Client spec seems to allow using a different property but this seems not to works on Quarkus.

```
public @interface RegisterRestClient {
    [...]

    /**
     * Associates the annotated rest client interface with this configuration key. By specifying a non-empty value,
     * this interface can be configured more simply using the configuration key rather than the fully-qualified class
     * name of the interface.
     *
     * @return the configuration key in use by this client interface. An empty value means that this interface is not
     *  associated with a configuration key.
     * @since 1.3
     */
        String configKey() default "";
}
```

**To Reproduce**
1. Create a rest client
2. Use `RegisterRestClient(configKey="some.config")`
3. Add the following line to your application.properties `some.config=http://myhost:myport`
4. Call the rest client

 - Quarkus version or git rev: 1.0.0.CR1 or current master

**Additional context**
`io.quarkus.restclient.runtime.RestClientBase.configureBaseUrl(RestClientBuilder)` doesn't take into account the confgKey (in fact, the configKey is used ... but only to be used inside the error message).

I will provides a PR for this. 

I will also add an entry inside the documentation as it is not documented but really a good thing to use as the default config key is very long and not very friendly.

loicmathieu:
Fixes #5619

I provides a better error message (the default one makes me think there is a bug:) ).

I add some documentation because this is really a usefull feature.

I didn't document the `baseUrl` annotation property because I don't think it something a user should use ...

### DIFF
--- a/docs/src/main/asciidoc/rest-client.adoc
+++ b/docs/src/main/asciidoc/rest-client.adoc
@@ -141,6 +141,24 @@ The default scope can also be defined on the interface.
 
 Note that `org.acme.restclient.CountriesService` _must_ match the fully qualified name of the `CountriesService` interface we created in the previous section.
 
+To facilitate the configuration, you can use the `@RegisterRestClient` `configKey` property that allows to use another configuration root than the fully qualified name of your interface.
+
+[source, java]
+----
+
+@RegisterRestClient(configKey="country-api")
+public interface CountriesService {
+    [...]
+}
+----
+
+[source,properties]
+----
+# Your configuration properties
+country-api/mp-rest/url=https://restcountries.eu/rest #
+country-api/mp-rest/scope=javax.inject.Singleton # /
+----
+
 == Update the JAX-RS resource
 
 Open the `src/main/java/org/acme/restclient/CountriesResource.java` file and update it with the following content:
--- a/extensions/rest-client/runtime/src/main/java/io/quarkus/restclient/runtime/RestClientBase.java
+++ b/extensions/rest-client/runtime/src/main/java/io/quarkus/restclient/runtime/RestClientBase.java
@@ -189,8 +189,11 @@ public class RestClientBase {
                 && !propertyOptional.isPresent()) {
             throw new IllegalArgumentException(
                     String.format(
-                            "Unable to determine the proper baseUrl/baseUri. Consider registering using @RegisterRestClient(baseUri=\"someuri\", configKey=\"orkey\"), or by adding '%s' to your Quarkus configuration",
-                            propertyPrefix));
+                            "Unable to determine the proper baseUrl/baseUri. " +
+                                    "Consider registering using @RegisterRestClient(baseUri=\"someuri\"), @RegisterRestClient(configKey=\"orkey\"), "
+                                    +
+                                    "or by adding '%s' or '%s' to your Quarkus configuration",
+                            String.format(REST_URL_FORMAT, propertyPrefix), String.format(REST_URI_FORMAT, propertyPrefix)));
         }
         String baseUrl = propertyOptional.orElse(baseUriFromAnnotation);
 


extensions/rest-client/runtime/src/main/java/io/quarkus/restclient/runtime/RestClientBase.java
```diff
@@ -189,8 +189,11 @@ private void configureBaseUrl(RestClientBuilder builder) {
                 && !propertyOptional.isPresent()) {
             throw new IllegalArgumentException(
                     String.format(
-                            "Unable to determine the proper baseUrl/baseUri. Consider registering using @RegisterRestClient(baseUri=\"someuri\", configKey=\"orkey\"), or by adding '%s' to your Quarkus configuration",
-                            propertyPrefix));
+                            "Unable to determine the proper baseUrl/baseUri. " +
+                                    "Consider registering using @RegisterRestClient(baseUri=\"someuri\"), @RegisterRestClient(configKey=\"orkey\"), "
+                                    +
+                                    "or by adding '%s' or '%s'to your Quarkus configuration",
```
gsmet:
```suggestion
                                    "or by adding '%s' or '%s' to your Quarkus configuration",
```

gsmet (dismissed):
I don't have an opinion about that one, just spotted a typo.

/cc @kenfinnigan

extensions/rest-client/runtime/src/main/java/io/quarkus/restclient/runtime/RestClientBase.java
gastaldi:
```suggestion
                                    "or by adding '%s' or '%s' to your Quarkus configuration",
```

docs/src/main/asciidoc/rest-client.adoc
```diff
@@ -141,6 +141,24 @@ The default scope can also be defined on the interface.
...
+[source, java]
+----
+
+@RegisterRestClient(configKey="country-api")
+public interface CountriesService {
+    [...]
+}
+----
+
+[source,shell]
```
gastaldi:
```suggestion
[source,properties]
```

loicmathieu:
@gsmet @gastaldi I udapted with your changes

gastaldi (approved):
LGTM

docs/src/main/asciidoc/rest-client.adoc
```diff
@@ -141,6 +141,24 @@ The default scope can also be defined on the interface.
 
 Note that `org.acme.restclient.CountriesService` _must_ match the fully qualified name of the `CountriesService` interface we created in the previous section.
 
+To facilitate the configuration, you can use the `@RegisterRestClient` `configKey` property that allows to use another configuration root than the name of your interface.
```
kenfinnigan: This should be more descriptive than just saying "name", as it's actually replacing the full package and class name

kenfinnigan (changes requested):

docs/src/main/asciidoc/rest-client.adoc
loicmathieu: Is `the full qualified name of your interface` better ?
The example is self explainatory :)

kenfinnigan: That works

loicmathieu: done

kenfinnigan (approved):

gsmet:
So, I suppose we should merge it?

PR #5632, opened 2019-11-20
Replace TCCL with CL from DeploymentClassLoaderBuildItem in deployment modules

geoand:
This doesn't fix a bug or anything, it's just the "proper" way of obtaining a ClassLoader that can load classes in a build step

### DIFF
--- a/extensions/agroal/deployment/src/main/java/io/quarkus/agroal/deployment/AgroalProcessor.java
+++ b/extensions/agroal/deployment/src/main/java/io/quarkus/agroal/deployment/AgroalProcessor.java
@@ -66,7 +66,7 @@ class AgroalProcessor {
 
     @SuppressWarnings("unchecked")
     @Record(STATIC_INIT)
-    @BuildStep
+    @BuildStep(loadsApplicationClasses = true)
     BeanContainerListenerBuildItem build(
             RecorderContext recorderContext,
             AgroalRecorder recorder,
--- a/extensions/infinispan-client/deployment/src/main/java/io/quarkus/infinispan/client/deployment/InfinispanClientProcessor.java
+++ b/extensions/infinispan-client/deployment/src/main/java/io/quarkus/infinispan/client/deployment/InfinispanClientProcessor.java
@@ -47,7 +47,6 @@ import io.quarkus.deployment.annotations.ExecutionTime;
 import io.quarkus.deployment.annotations.Record;
 import io.quarkus.deployment.builditem.ApplicationArchivesBuildItem;
 import io.quarkus.deployment.builditem.ApplicationIndexBuildItem;
-import io.quarkus.deployment.builditem.DeploymentClassLoaderBuildItem;
 import io.quarkus.deployment.builditem.ExtensionSslNativeSupportBuildItem;
 import io.quarkus.deployment.builditem.FeatureBuildItem;
 import io.quarkus.deployment.builditem.HotDeploymentWatchedFileBuildItem;
@@ -71,7 +70,7 @@ class InfinispanClientProcessor {
      */
     InfinispanClientBuildTimeConfig infinispanClient;
 
-    @BuildStep
+    @BuildStep(loadsApplicationClasses = true)
     InfinispanPropertiesBuildItem setup(ApplicationArchivesBuildItem applicationArchivesBuildItem,
             BuildProducer<ReflectiveClassBuildItem> reflectiveClass,
             BuildProducer<HotDeploymentWatchedFileBuildItem> hotDeployment,
@@ -80,7 +79,6 @@ class InfinispanClientProcessor {
             BuildProducer<AdditionalBeanBuildItem> additionalBeans,
             BuildProducer<ExtensionSslNativeSupportBuildItem> sslNativeSupport,
             BuildProducer<NativeImageConfigBuildItem> nativeImageConfig,
-            DeploymentClassLoaderBuildItem classLoaderBuildItem,
             ApplicationIndexBuildItem applicationIndexBuildItem) throws ClassNotFoundException, IOException {
 
         feature.produce(new FeatureBuildItem(FeatureBuildItem.INFINISPAN_CLIENT));
@@ -91,8 +89,7 @@ class InfinispanClientProcessor {
         // Enable SSL support by default
         sslNativeSupport.produce(new ExtensionSslNativeSupportBuildItem(FeatureBuildItem.INFINISPAN_CLIENT));
 
-        ClassLoader cl = Thread.currentThread().getContextClassLoader();
-        InputStream stream = cl.getResourceAsStream(HOTROD_CLIENT_PROPERTIES);
+        InputStream stream = Thread.currentThread().getContextClassLoader().getResourceAsStream(HOTROD_CLIENT_PROPERTIES);
         Properties properties;
         if (stream == null) {
             properties = new Properties();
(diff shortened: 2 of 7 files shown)


mkouba:
Hm, we should probably check all occurences of `Thread.currentThread().getContextClassLoader()` in the "deployment" code. AFAIR there was no better way before the `DeploymentClassLoaderBuildItem` existed so I'd expect more places to fix.

geoand:
Indeed, however I was familiar with this piece of the code :). I don't want to go messing with everyone's build steps :P

gsmet:
Better do it globally and once and for all IMHO. Otherwise people will continue copying bad patterns.

geoand:
OK, I'll check it out

geoand:
I'm running all the tests locally now to see if what I have passes

geoand:
I updated the deployment modules of extensions that were using TCCL

extensions/agroal/deployment/src/main/java/io/quarkus/agroal/deployment/AgroalProcessor.java
```diff
@@ -100,8 +103,12 @@ BeanContainerListenerBuildItem build(
                 java.sql.ResultSet.class.getName(),
                 java.sql.ResultSet[].class.getName()));
 
-        validateBuildTimeConfig(null, agroalBuildTimeConfig.defaultDataSource);
-        agroalBuildTimeConfig.namedDataSources.forEach(AgroalProcessor::validateBuildTimeConfig);
+        validateBuildTimeConfig(null, agroalBuildTimeConfig.defaultDataSource, deploymentClassLoader.getClassLoader());
+        Map<String, DataSourceBuildTimeConfig> namedDataSources = agroalBuildTimeConfig.namedDataSources;
```
gsmet: It's done, it's done but lambdas are usually accepted at build time (unless it doesn't make the code clearer) but in this case, it was a bit more readable before.

Anyway, it's no biggie, let's keep it the way you did it.

gsmet (approved):

extensions/agroal/deployment/src/main/java/io/quarkus/agroal/deployment/AgroalProcessor.java
geoand: I have developed lambda aversion 😂

extensions/agroal/deployment/src/main/java/io/quarkus/agroal/deployment/AgroalProcessor.java
```diff
@@ -70,6 +72,7 @@
     BeanContainerListenerBuildItem build(
             RecorderContext recorderContext,
             AgroalRecorder recorder,
+            DeploymentClassLoaderBuildItem deploymentClassLoader,
```
stuartwdouglas: You can just set io.quarkus.deployment.annotations.BuildStep#loadsApplicationClasses to true and the TCCL will be set to this CL.

extensions/hibernate-orm/deployment/src/main/java/io/quarkus/hibernate/orm/deployment/HibernateEntityEnhancer.java
```diff
@@ -26,11 +26,16 @@
  */
 public final class HibernateEntityEnhancer implements BiFunction<String, ClassVisitor, ClassVisitor> {
 
+    private final ClassLoader classLoader;
     private final BytecodeProvider provider = new org.hibernate.bytecode.internal.bytebuddy.BytecodeProviderImpl();
 
+    public HibernateEntityEnhancer(ClassLoader classLoader) {
```
stuartwdouglas: Transformations are already run with the deployment ClassLoader as the TCCL

extensions/smallrye-openapi/deployment/src/main/java/io/quarkus/smallrye/openapi/deployment/SmallRyeOpenApiProcessor.java
```diff
@@ -234,6 +234,7 @@ private void registerReflectionForApiResponseSchemaSerialization(BuildProducer<R
 
     @BuildStep(loadsApplicationClasses = true)
```
stuartwdouglas: This is already using the correct TCCL due to loadsApplicationClasses= true

stuartwdouglas (changes requested):

extensions/smallrye-openapi/deployment/src/main/java/io/quarkus/smallrye/openapi/deployment/SmallRyeOpenApiProcessor.java
geoand: I wasn't aware of that flag. So which of the two solutions is preferred?
Asking because people are definitely going to be copying this kind of thing in other extensions

stuartwdouglas: loadsApplicationClasses is should be preferred, as it automatically sets the correct TCCL. If you are loading classes from the DeploymentClassLoader but have not set it to the TCCL there is the possibility for subtle bugs if the classes being loaded use the TCCL.

I did actually think about defaulting to true and requiring build steps that feed into generating the DeploymentClassLoader to set it to false but I am not sure how invasive that change would be.

geoand: Gotcha, thanks. I'll update the PR accordingly

geoand:
@stuartwdouglas PR updated

extensions/infinispan-client/deployment/src/main/java/io/quarkus/infinispan/client/deployment/InfinispanClientProcessor.java
```diff
@@ -91,7 +91,7 @@ InfinispanPropertiesBuildItem setup(ApplicationArchivesBuildItem applicationArch
         // Enable SSL support by default
         sslNativeSupport.produce(new ExtensionSslNativeSupportBuildItem(FeatureBuildItem.INFINISPAN_CLIENT));
 
-        ClassLoader cl = Thread.currentThread().getContextClassLoader();
```
stuartwdouglas: This should also use loadsApplicationClasses and just use the TCCL (not the fault of this PR, but while we are fixing these issues might as well get this one too).

I know it seems like a bit of a corner case (the wrong TCCL being set during class loading), but this did actually cause real issues in the early days of WildFly so it can be a problem.

stuartwdouglas (changes requested):

extensions/infinispan-client/deployment/src/main/java/io/quarkus/infinispan/client/deployment/InfinispanClientProcessor.java
geoand: That makes sense. PR updated

stuartwdouglas (approved):

mkouba:
I think we should also document `BuildStep.loadsApplicationClasses()`; there is no mention in the guide and even no javadoc.

geoand:
@mkouba are you signing up for that? :P

mkouba:
@geoand oook, I'll try to prepare something in a different PR. Note to myself: next time pls keep silent ;-)

geoand:
> @geoand oook, I'll try to prepare something in a different PR. Note to myself: next time pls keep silent ;-)

Hahaha

PR #5646, opened 2019-11-21
Use GizmoAdaptor where possible

stuartwdouglas:


mkouba (approved):

geoand (approved):
Wasn't aware of this one, nice!

There are definitely more occurences of this in the code that I guess we'll track down little by little

stuartwdouglas:
I don't think so, I just looked at all implementers in my IDE and removed the duplicated classes.

geoand:
One occurence is this: https://github.com/quarkusio/quarkus/blob/master/extensions/spring-web/deployment/src/main/java/io/quarkus/spring/web/deployment/SpringWebProcessor.java#L340

I guess all the other occurences use `GeneratedBeanBuildItem` like this: https://github.com/quarkusio/quarkus/blob/master/extensions/spring-data-jpa/deployment/src/main/java/io/quarkus/spring/data/deployment/SpringDataJPAProcessor.java#L101

stuartwdouglas:
For some reason I don't understand Intellij has been randomly marking maven modules as ignored for me, which is why I did not pick this one up. 

We should also create a version of this for GeneratedBeanBuildItem.

geoand:
> For some reason I don't understand Intellij has been randomly marking maven modules as ignored for me, which is why I did not pick this one up.
> 

I have seen the same exact problem every now and then. 

> We should also create a version of this for GeneratedBeanBuildItem.

Makes sense. I'll do that later if no one beats me to it.

PR #5652, opened 2019-11-21
Take into account repeated constraint annotations

### LINKED ISSUE #5549: Repeated constraints are not supported
Right now, I don't think we properly support repeated constraints in the Hibernate Validator extension. I was recently made aware that Jandex does not return the repeated annotation occurrences when querying the index and that we should query the index with the container annotation too.

That means that for all the repeatable constraints (`@Pattern` is a good example of that and could be used for testing), we will also need to look for the associated containers.

So we should add the containers to the considered annotations here: https://github.com/quarkusio/quarkus/blob/master/extensions/hibernate-validator/deployment/src/main/java/io/quarkus/hibernate/validator/deployment/HibernateValidatorProcessor.java#L136-L141 .

The containers should probably be extracted from the constraint itself by querying the index and look for the container defined in the `@Repeatable` annotation.

As we don't really analyze the content of the constraints but are just looking for the occurrences, we shouldn't have to interpret the container as it is done here: https://github.com/quarkusio/quarkus/blob/master/extensions/scheduler/deployment/src/main/java/io/quarkus/scheduler/deployment/SchedulerProcessor.java#L148 . Just mentioning it to be thorough.

A test should be added here (and we should start with that): https://github.com/quarkusio/quarkus/tree/master/extensions/hibernate-validator/deployment/src/test/java/io/quarkus/hibernate/validator/test .

irenakezic:
Take into account repeated constraint annotations

Fixes  [#5549](https://github.com/quarkusio/quarkus/issues/5549)

### DIFF
--- a/extensions/hibernate-validator/deployment/src/test/java/io/quarkus/hibernate/validator/test/RepeatedConstraintsTest.java
+++ b/extensions/hibernate-validator/deployment/src/test/java/io/quarkus/hibernate/validator/test/RepeatedConstraintsTest.java
@@ -1,0 +1,109 @@
+package io.quarkus.hibernate.validator.test;
+
+import static java.lang.annotation.ElementType.ANNOTATION_TYPE;
+import static java.lang.annotation.ElementType.CONSTRUCTOR;
+import static java.lang.annotation.ElementType.FIELD;
+import static java.lang.annotation.ElementType.METHOD;
+import static java.lang.annotation.ElementType.PARAMETER;
+import static java.lang.annotation.ElementType.TYPE_USE;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.Repeatable;
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+import javax.enterprise.context.ApplicationScoped;
+import javax.inject.Inject;
+import javax.validation.Constraint;
+import javax.validation.ConstraintValidator;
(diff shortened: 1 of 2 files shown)


extensions/hibernate-validator/deployment/src/test/java/io/quarkus/hibernate/validator/test/RepeatedConstraintsTest.java
```diff
@@ -0,0 +1,98 @@
...
+            this.testValue2 = testValue2;
+        }
+
+        @Pattern(regexp = ".*[0-9]+.*", message = "Error message")
+        @Pattern(regexp = ".*[a-z]+.*", message = "Error message")
+        public String testValue1;
+
+        @MyConstraint("value1")
+        @MyConstraint("value2")
+        public String testValue2;
```
gsmet: Can you separate the tests in 2 bean classes?

Because, if at least one works, then all the constraints will be detected so we really need to test the built-in constraints and the custom constraints in 2 separate beans.

gsmet (changes requested):
I added one small comment about the test.

Otherwise, it looks very good, thanks!

extensions/hibernate-validator/deployment/src/test/java/io/quarkus/hibernate/validator/test/RepeatedConstraintsTest.java
irenakezic: ok sure

gsmet (approved):

gsmet:
Looks good, thanks!

Congrats for your first contribution!

Issue #4347, opened 2019-10-03
Add a JPA Identity Extension


stuartwdouglas:
This extension would allow for an easy annotation driven approach to creating JPA entities that represent a user. An entity can be annotated with @Username and @Password (and maybe @Salt) to inform Quarkus that this entity represents a user. Quarkus will then automatically wire up authentication based on this entity.

FroMage:
Looking at it for prototyping.

FroMage:
Well, I stopped looking at it, but I do have a branch with the start of this. I hit the issue that it appears that the credentials are being looked up outside of a scope in which ArC can deliver an `EntityManager` (so probably outside the request scope), which I'm not sure how to fix. I can probably get an `EntityManagerFactory` and create one, but it's probably not right performance-wise?

stuartwdouglas:
We need to create the request scope earlier then I guess. I was trying to avoid this as there is some complexity in dealing with this if the thread is propagated between different threads. 

Creating an EntityManager is also a solution (and should not really be a perf issue, they are supposed to be fairly cheap), but it means that if we are going to stick the user entity into the SecurityIdentity then it will be a detached entity, although this might be ok.

FroMage:
It's almost done here: https://github.com/quarkusio/quarkus/pull/5660

Issue #5113, opened 2019-11-01
Provide API to create bcrypt hash and salt


gnunn1:
**Description**
Quarkus provides a bcrypt password mapper for the elyton JDBC security provider but doesn't appear to offer a way to create a bcrypt salt and hash thus limiting it's usefulness. It is possible to do this using JARs from Wildfly however they will not compile natively. 

Example using code from Wildfly, pom.xml:

```
    <dependency>
      <groupId>org.wildfly.security</groupId>
      <artifactId>wildfly-elytron-credential</artifactId>
      <version>1.10.3.Final</version>
    </dependency>
```
Code:

```
    public void setPasswordHash(String password) {
        byte[] salt = generateRandomSalt(BCRYPT_SALT_SIZE);
        int iterationCount = 10;

        try {
            PasswordFactory factory = PasswordFactory.getInstance(ALGORITHM_BCRYPT);
            BCryptPassword bCryptPassword = (BCryptPassword) factory.generatePassword(
                    new EncryptablePasswordSpec(password.toCharArray(), new IteratedSaltedPasswordAlgorithmSpec(iterationCount, salt))
            );
            this.passwordHash = Base64.getEncoder().encodeToString(bCryptPassword.getHash());
            this.salt = Base64.getEncoder().encodeToString(bCryptPassword.getSalt());
            this.iterations = bCryptPassword.getIterationCount();
            log.info("hash: " + this.passwordHash + ", salt:" + this.salt + ", iterations:" + this.iterations );
        } catch (InvalidKeySpecException e) {
            throw new RuntimeException("Password encryption failed, invalid key spec", e);
        } catch (NoSuchAlgorithmException e) {
            throw new RuntimeException("Password encryption failed, no such algorithm", e);
        }
    }

    private static byte[] generateRandomSalt(int saltSize) {
        byte[] randomSalt = new byte[saltSize];
        ThreadLocalRandom.current().nextBytes(randomSalt);
        return randomSalt;
    }
```

sberyozkin:
@danielpetisme Hi Daniel, if you can find some time to check this one then it would be nice

dmlloyd:
> It is possible to do this using JARs from Wildfly however they will not compile natively.

What do you mean by "they will not compile natively"?  Is native image generation failing with these JARs, or...?

gnunn1:
Yes, native image generation is failing, below is full output:

```
mvn package -Pnative
[INFO] Scanning for projects...
[INFO] 
[...]

PR #5660, opened 2019-11-21
Security JPA implementation

FroMage:
This is for #4347, and it's Almost Done™ (Famous Last Words©).

You can define your security user entity like this:

```java
@UserDefinition
@Entity
public class PanacheUserEntity extends PanacheEntity {
    @Username
    public String name;

    @Password
    public String pass;

    @Roles
    public String roles;
}
```

Or with multiple tables for roles:

```java
@UserDefinition
@Entity
public class PanacheUserEntity extends PanacheEntity {
    @Username
    public String name;

    @Password
    public String pass;

    @ManyToMany
    @Roles
    public List<PanacheRoleEntity> roles = new ArrayList<>();
}

@Entity
public class PanacheRoleEntity extends PanacheEntity {

    @ManyToMany(mappedBy = "roles")
    public List<PanacheUserEntity> users;

    @RolesValue
    public String role;

}
```

- Supported role types of types `String`, `Collection/List/Set<String>`, `Collection/List/Set<role-entity>`, which can all contain comma-separated strings of roles.
- Supports plain Hibernate ORM or ORM with Panache
- No config required
- Supports `@NaturalId` on the same field as `@Username`
- Works _before_ the request, using `JPAConfig`'s support to get `EntityManagerFactory`
- Supports clear and MCF passwords

Under the hood, it generates this sort of `IdentityProvider` implementation:

```java
@Singleton
public class UserEntityIdentityProvider implements JpaIdentityProvider {

    @Override
    public SecurityIdentity authenticate(EntityManager em,
            UsernamePasswordAuthenticationRequest request) {

        // With Natural Id
        Session session = em.unwrap(Session.class);
        SimpleNaturalIdLoadAccess<PlainUserEntity> naturalIdLoadAccess = session.bySimpleNaturalId(PlainUserEntity.class);
        PlainUserEntity user = naturalIdLoadAccess.load(request.getUsername());
        // With normal query
        //        Query query = em.createQuery("FROM PlainUserEntity WHERE name = :name");
        //        query.setParameter("name", request.getUsername());
        //        PlainUserEntity user = getSingleUser(query);
        if (user == null)
            return null;

        // for MCF:
        //               Password storedPassword = getMcfPasword(user.pass);
        // for clear:
        Password storedPassword = getClearPassword(user.pass);

        QuarkusSecurityIdentity.Builder builder = checkPassword(storedPassword, request);

        addRoles(builder, user.role);
        return builder.build();
    }

}
```

So overall it's pretty complete, but I have a few questions before I'm ready to merge (besides the lack of docs):

- I default to MCF (#5667) passwords in single column (which supports bcrypt and many others), with clear text password being supported. I think this is fine for the start, right?
- I wanted to only require a single `@UserDefinition` annotation and default to looking up fields if they're named "user", "username", "pass", "password", "passwd", etc… but then I wondered if that would not be a security risk if a user adds a `@UserDefinition` annotation to his class and we pick up `user` for the username, but it turns out that in his model, this is not a primary key column and we should actually use the `email` column, so this could be a dangerous risk so I didn't do it. It feels lame to have to use 4 annotations, though, so WDYT? I'd love to use convention over configuration here.

### DIFF
--- a/bom/deployment/pom.xml
+++ b/bom/deployment/pom.xml
@@ -614,6 +614,11 @@
             </dependency>
             <dependency>
                 <groupId>io.quarkus</groupId>
+                <artifactId>quarkus-security-jpa-deployment</artifactId>
+                <version>${project.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>io.quarkus</groupId>
                 <artifactId>quarkus-security-test-utils</artifactId>
                 <version>${project.version}</version>
                 <scope>test</scope>
--- a/bom/runtime/pom.xml
+++ b/bom/runtime/pom.xml
@@ -742,6 +742,11 @@
             </dependency>
             <dependency>
                 <groupId>io.quarkus</groupId>
+                <artifactId>quarkus-security-jpa</artifactId>
+                <version>${project.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>io.quarkus</groupId>
                 <artifactId>quarkus-vault</artifactId>
                 <version>${project.version}</version>
             </dependency>
--- a/core/deployment/src/main/java/io/quarkus/deployment/Capabilities.java
+++ b/core/deployment/src/main/java/io/quarkus/deployment/Capabilities.java
@@ -25,6 +25,7 @@ public final class Capabilities extends SimpleBuildItem {
     public static final String SECURITY_ELYTRON_OAUTH2 = "io.quarkus.elytron.security.oauth2";
     public static final String SECURITY_ELYTRON_JDBC = "io.quarkus.elytron.security.jdbc";
     public static final String SECURITY_ELYTRON_LDAP = "io.quarkus.elytron.security.ldap";
+    public static final String SECURITY_JPA = "io.quarkus.security.jpa";
     public static final String QUARTZ = "io.quarkus.quartz";
     public static final String METRICS = "io.quarkus.metrics";
 
--- a/docs/src/main/asciidoc/security-jpa.adoc
+++ b/docs/src/main/asciidoc/security-jpa.adoc
@@ -1,0 +1,360 @@
+////
+This guide is maintained in the main Quarkus repository
+and pull requests should be submitted there:
+https://github.com/quarkusio/quarkus/tree/master/docs/src/main/asciidoc
+////
+= Quarkus - Using Security from JPA
+
+include::./attributes.adoc[]
+
+This guide demonstrates how your Quarkus application can use a database to store your user identities with 
+link:hibernate-orm[Hibernate ORM] or link:hibernate-orm-panache[Hibernate ORM with Panache].
+
+
+== Prerequisites
+
+To complete this guide, you need:
+
+* less than 15 minutes
+* an IDE
+* JDK 1.8+ installed with `JAVA_HOME` configured appropriately
+* Apache Maven 3.5.3+
+
+== Architecture
+
+In this example, we build a very simple microservice which offers three endpoints:
+
+* `/api/public`
+* `/api/users/me`
+* `/api/admin`
+
+The `/api/public` endpoint can be accessed anonymously.
+The `/api/admin` endpoint is protected with RBAC (Role-Based Access Control) where only users granted with the `admin` role can access. At this endpoint, we use the `@RolesAllowed` annotation to declaratively enforce the access constraint.
+The `/api/users/me` endpoint is also protected with RBAC (Role-Based Access Control) where only users granted with the `user` role can access. As a response, it returns a JSON document with details about the user.
+
+== Solution
+
+We recommend that you follow the instructions in the next sections and create the application step by step.
+However, you can go right to the completed example.
+
+Clone the Git repository: `git clone {quickstarts-clone-url}`, or download an {quickstarts-archive-url}[archive].
+
+The solution is located in the `security-jpa-quickstart` {quickstarts-tree-url}/security-jpa-quickstart[directory].
+
+== Creating the Maven Project
+
+First, we need a new project. Create a new project with the following command:
+
+[source, subs=attributes+]
+----
+mvn io.quarkus:quarkus-maven-plugin:{quarkus-version}:create \
+    -DprojectGroupId=org.acme \
+    -DprojectArtifactId=security-jpa-quickstart \
+    -Dextensions="security-jpa, jdbc-postgresql, resteasy, hibernate-orm-panache"
+cd security-jpa-quickstart
+----
+
+[NOTE]
+====
+Don't forget to add the database connector library of choice. Here we are using PostgreSQL as identity store.
+====
+
+This command generates a Maven project, importing the `security-jpa` extension
+which allows you to map your security source to JPA entities.
+
+== Writing the application
+
+Let's start by implementing the `/api/public` endpoint. As you can see from the source code below, it is just a regular JAX-RS resource:
+
+[source,java]
+----
+package org.acme.security.jpa;
+
+import javax.annotation.security.PermitAll;
+import javax.ws.rs.GET;
+import javax.ws.rs.Path;
+import javax.ws.rs.Produces;
+import javax.ws.rs.core.MediaType;
+
+@Path("/api/public")
+public class PublicResource {
+
+    @GET
+    @PermitAll
+    @Produces(MediaType.TEXT_PLAIN)
+    public String publicResource() {
+        return "public";
+   }
+}
+----
+
+The source code for the `/api/admin` endpoint is also very simple. The main difference here is that we are using a `@RolesAllowed` annotation to make sure that only users granted with the `admin` role can access the endpoint:
+
+
+[source,java]
+----
+package org.acme.security.jpa;
+
+import javax.annotation.security.RolesAllowed;
+import javax.ws.rs.GET;
+import javax.ws.rs.Path;
+import javax.ws.rs.Produces;
+import javax.ws.rs.core.MediaType;
+
+@Path("/api/admin")
+public class AdminResource {
+
+    @GET
+    @RolesAllowed("admin")
+    @Produces(MediaType.TEXT_PLAIN)
+    public String adminResource() {
+         return "admin";
+    }
+}
+----
+
+Finally, let's consider the `/api/users/me` endpoint. As you can see from the source code below, we are trusting only users with the `user` role.
+We are using `SecurityContext` to get access to the current authenticated Principal and we return the user's name. This information is loaded from the database.
+
+[source,java]
+----
+package org.acme.security.jpa;
+
+import javax.annotation.security.RolesAllowed;
+import javax.inject.Inject;
+import javax.ws.rs.GET;
+import javax.ws.rs.Path;
+import javax.ws.rs.Produces;
+import javax.ws.rs.core.Context;
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.SecurityContext;
+
+@Path("/api/users")
+public class UserResource {
+
+    @GET
+    @RolesAllowed("user")
+    @Path("/me")
+    @Produces(MediaType.APPLICATION_JSON)
+    public String me(@Context SecurityContext securityContext) {
+        return securityContext.getUserPrincipal().getName();
+    }
+}
+----
+
+=== Defining our user entity
+
+We can now describe how our security information is stored in our model by adding a few annotations to our `User` entity:
+
+[source,java]
+----
+package org.acme.security.jpa;
+
+import javax.persistence.Entity;
+import javax.persistence.Table;
+
+import io.quarkus.hibernate.orm.panache.PanacheEntity;
+import io.quarkus.security.common.BcryptUtil;
+import io.quarkus.security.jpa.Password;
+import io.quarkus.security.jpa.Roles;
+import io.quarkus.security.jpa.UserDefinition;
+import io.quarkus.security.jpa.Username;
+
+@Entity
+@Table(name = "test_user")
+@UserDefinition <1>
+public class User extends PanacheEntity {
+    @Username <2>
+    public String username;
+    @Password <3>
+    public String password;
+    @Roles <4>
+    public String role;
+    
+    /**
+     * Adds a new user in the database
+     * @param username the user name
+     * @param password the unencrypted password (it will be encrypted with bcrypt)
+     * @param role the comma-separated roles
+     */
+    public static void add(String username, String password, String role) { <5>
+        User user = new User();
+        user.username = username;
+        user.password = BcryptUtil.bcryptHash(password);
+        user.role = role;
+        user.persist();
+    }
+}
+
+----
+
+The `security-jpa` extension is only initialized if there is a single entity annotated with `@UserDefinition`.
+
+<1> This annotation must be present on a single entity. It can be a regular Hibernate ORM entity or a Hibernate ORM with Panache entity as in this example.
+<2> This indicates the field used for the user name.
+<3> This indicates the field used for the password. This defaults to using bcrypt hashed passwords, but you can also configure it for clear text passwords.
+<4> This indicates the comma-separated list of roles added to the target Principal representation attributes.
+<5> This method allows us to add users while hashing the password with the proper bcrypt hash.
+
+=== Configuring the Application
+
+The `security-jpa` extension requires at least one datasource to access to your database.
+
+[source,properties]
+----
+quarkus.datasource.url=jdbc:postgresql:security_jpa
+quarkus.datasource.driver=org.postgresql.Driver
+quarkus.datasource.username=quarkus
+quarkus.datasource.password=quarkus
+
+quarkus.hibernate-orm.database.generation=drop-and-create
+----
+
+In our context, we are using PostgreSQL as identity store. The database schema is created by Hibernate ORM automatically
+on startup (change this in production) and we initialise the database with users and roles in the `Startup` class:
+
+[source,java]
+----
+package org.acme.security.jpa;
+
+import javax.enterprise.event.Observes;
+import javax.inject.Singleton;
+import javax.transaction.Transactional;
+
+import io.quarkus.runtime.StartupEvent;
+
+
+@Singleton
+public class Startup {
+    @Transactional
+    public void loadUsers(@Observes StartupEvent evt) {
+        // reset and load all test users
+        User.deleteAll();
+        User.add("admin", "admin", "admin");
+        User.add("user", "user", "user");
+    }
+}
+----
+
+[NOTE]
+====
+It is probably useless but we kindly remind you that you must not store clear-text passwords in production environments ;-).
+As a result, the `security-jpa` defaults to using bcrypt-hashed passwords.
+====
+
+== Testing the Application
+
+The application is now protected and the identities are provided by our database.
+The very first thing to check is to ensure the anonymous access works.
+
+[source,shell]
+----
+$ curl -i -X GET http://localhost:8080/api/public
+HTTP/1.1 200 OK
+Content-Length: 6
+Content-Type: text/plain;charset=UTF-8
+
+public%
+----
+
+Now, let's try a to hit a protected resource anonymously.
+
+[source,shell]
+----
+$ curl -i -X GET http://localhost:8080/api/admin
+HTTP/1.1 401 Unauthorized
+Content-Length: 14
+Content-Type: text/html;charset=UTF-8
+
+Not authorized%
+----
+
+So far so good, now let's try with an allowed user.
+
+[source,shell]
+----
+$ curl -i -X GET -u admin:admin http://localhost:8080/api/admin
+HTTP/1.1 200 OK
+Content-Length: 5
+Content-Type: text/plain;charset=UTF-8
+
+admin%
+----
+By providing the `admin:admin` credentials, the extension authenticated the user and loaded their roles.
+The `admin` user is authorized to access to the protected resources.
+
+The user `admin` should be forbidden to access a resource protected with `@RolesAllowed("user")` because it doesn't have this role.
+[source,shell]
+----
+$ curl -i -X GET -u admin:admin http://localhost:8080/api/users/me
+HTTP/1.1 403 Forbidden
+Content-Length: 34
+Content-Type: text/html;charset=UTF-8
+
+Forbidden%
+----
+
+Finally, using the user `user` works and the security context contains the principal details (username for instance).
+[source,shell]
+----
+curl -i -X GET -u user:user http://localhost:8080/api/users/me
+HTTP/1.1 200 OK
+Content-Length: 4
+Content-Type: text/plain;charset=UTF-8
+
+user%
+----
+
+== Supported model types
+
+- The `@UserDefinition` class must be a JPA entity (with Panache or not).
+- The `@Username` and `@Password` field types must be of type `String`.
+- The `@Roles` field must either be of type `String` or `Collection<String>` or alternately a `Collection<X>` where `X` is an entity class with one `String` field annotated with the `@RolesValue` annotation.
+- Each `String` role element type will be parsed as a comma-separated list of roles.
+
+== Storing roles in another entity
+
+You can also store roles in another entity:
+
+[source,java]
+----
+@UserDefinition
+@Table(name = "test_user")
+@Entity
+public class User extends PanacheEntity {
+    @Username
+    public String name;
+
+    @Password
+    public String pass;
+
+    @ManyToMany
+    @Roles
+    public List<Role> roles = new ArrayList<>();
+}
+
+@Entity
+public class Role extends PanacheEntity {
+
+    @ManyToMany(mappedBy = "roles")
+    public List<ExternalRolesUserEntity> users;
+
+    @RolesValue
+    public String role;
+}
+----
+
+== Password storage and hashing
+
+By default, we consider passwords to be stored hashed with https://en.wikipedia.org/wiki/Bcrypt[bcrypt] under the 
+https://en.wikipedia.org/wiki/Crypt_(C)[Modular Crypt Format] (MCF).
+
+When you need to create such a hashed password we provide the convenient `String BcryptUtil.bcryptHash(String password)`
+function, which defaults to creating a random salt and hashing in 10 iterations (though you can specify the iterations and salt
+too).
+
+NOTE: with MCF you don't need dedicated columns to store the hashing algo, the iterations count or the salt because
+they're all stored in the hashed value.
+
+WARN: you can also store passwords in clear text with `@Password(PasswordType.CLEAR)` but we strongly recommend against
+it in production.
--- a/docs/src/main/asciidoc/security.adoc
+++ b/docs/src/main/asciidoc/security.adoc
@@ -22,6 +22,9 @@ in order for Quarkus to know how to find the authentication information to check
 |link:security-properties[quarkus-elytron-security-properties-file]
 |Provides support for simple properties files that can be used for testing security. This supports both embedding user info in `application.properties` and standalone properties files.
 
+|link:security-jpa[quarkus-security-jpa]
+|Provides support for authenticating via JPA.
+
 |link:security-jdbc[quarkus-elytron-security-jdbc]
 |Provides support for authenticating via JDBC.
 
--- a/extensions/elytron-security-common/deployment/pom.xml
+++ b/extensions/elytron-security-common/deployment/pom.xml
@@ -18,6 +18,10 @@
             <groupId>io.quarkus</groupId>
             <artifactId>quarkus-core-deployment</artifactId>
         </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-elytron-security-common</artifactId>
+        </dependency>
     </dependencies>
 
     <build>
--- a/extensions/panache/hibernate-orm-panache/deployment/src/main/java/io/quarkus/hibernate/orm/panache/deployment/PanacheResourceProcessor.java
+++ b/extensions/panache/hibernate-orm-panache/deployment/src/main/java/io/quarkus/hibernate/orm/panache/deployment/PanacheResourceProcessor.java
@@ -26,6 +26,7 @@ import io.quarkus.hibernate.orm.panache.PanacheRepositoryBase;
 import io.quarkus.panache.common.deployment.EntityField;
 import io.quarkus.panache.common.deployment.EntityModel;
 import io.quarkus.panache.common.deployment.MetamodelInfo;
+import io.quarkus.panache.common.deployment.PanacheEntityClassesBuildItem;
 import io.quarkus.panache.common.deployment.PanacheFieldAccessEnhancer;
 import io.quarkus.panache.common.deployment.PanacheRepositoryEnhancer;
 
@@ -60,7 +61,8 @@ public final class PanacheResourceProcessor {
     void build(CombinedIndexBuildItem index,
             ApplicationIndexBuildItem applicationIndex,
             BuildProducer<BytecodeTransformerBuildItem> transformers,
-            HibernateEnhancersRegisteredBuildItem hibernateMarker) throws Exception {
+            HibernateEnhancersRegisteredBuildItem hibernateMarker,
+            BuildProducer<PanacheEntityClassesBuildItem> entityClasses) throws Exception {
 
         PanacheJpaRepositoryEnhancer daoEnhancer = new PanacheJpaRepositoryEnhancer(index.getIndex());
         Set<String> daoClasses = new HashSet<>();
@@ -99,6 +101,9 @@ public final class PanacheResourceProcessor {
         }
         for (String modelClass : modelClasses) {
             transformers.produce(new BytecodeTransformerBuildItem(modelClass, modelEnhancer));
+        }
+        if (!modelClasses.isEmpty()) {
+            entityClasses.produce(new PanacheEntityClassesBuildItem(modelClasses));
         }
 
         MetamodelInfo<EntityModel<EntityField>> modelInfo = modelEnhancer.getModelInfo();
--- a/extensions/panache/mongodb-panache/deployment/src/main/java/io/quarkus/mongodb/panache/deployment/PanacheResourceProcessor.java
+++ b/extensions/panache/mongodb-panache/deployment/src/main/java/io/quarkus/mongodb/panache/deployment/PanacheResourceProcessor.java
@@ -45,6 +45,7 @@ import io.quarkus.mongodb.panache.PanacheMongoRecorder;
 import io.quarkus.mongodb.panache.PanacheMongoRepository;
 import io.quarkus.mongodb.panache.PanacheMongoRepositoryBase;
 import io.quarkus.mongodb.panache.ProjectionFor;
+import io.quarkus.panache.common.deployment.PanacheEntityClassesBuildItem;
 import io.quarkus.panache.common.deployment.PanacheFieldAccessEnhancer;
 import io.quarkus.panache.common.deployment.PanacheRepositoryEnhancer;
 
@@ -118,7 +119,8 @@ public class PanacheResourceProcessor {
             ApplicationIndexBuildItem applicationIndex,
             BuildProducer<BytecodeTransformerBuildItem> transformers,
             BuildProducer<ReflectiveClassBuildItem> reflectiveClass,
-            BuildProducer<PropertyMappingClassBuildStep> propertyMappingClass) throws Exception {
+            BuildProducer<PropertyMappingClassBuildStep> propertyMappingClass,
+            BuildProducer<PanacheEntityClassesBuildItem> entityClasses) throws Exception {
 
         PanacheMongoRepositoryEnhancer daoEnhancer = new PanacheMongoRepositoryEnhancer(index.getIndex());
         Set<String> daoClasses = new HashSet<>();
@@ -176,6 +178,9 @@ public class PanacheResourceProcessor {
 
             // Register for building the property mapping cache
             propertyMappingClass.produce(new PropertyMappingClassBuildStep(modelClass));
+        }
+        if (!modelClasses.isEmpty()) {
+            entityClasses.produce(new PanacheEntityClassesBuildItem(modelClasses));
         }
 
         if (!modelEnhancer.entities.isEmpty()) {
--- a/extensions/pom.xml
+++ b/extensions/pom.xml
@@ -97,6 +97,7 @@
 
         <!-- Security -->
         <module>security</module>
+        <module>security-jpa</module>
         <module>elytron-security-common</module>
         <module>elytron-security</module>
         <module>elytron-security-jdbc</module>
--- a/extensions/security-jpa/deployment/pom.xml
+++ b/extensions/security-jpa/deployment/pom.xml
@@ -1,0 +1,114 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>io.quarkus</groupId>
+        <artifactId>quarkus-security-jpa-parent</artifactId>
+        <version>999-SNAPSHOT</version>
+        <relativePath>../pom.xml</relativePath>
+    </parent>
+
+    <artifactId>quarkus-security-jpa-deployment</artifactId>
+    <name>Quarkus - Security JPA - Deployment</name>
+
+    <dependencies>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-security-jpa</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-core-deployment</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-arc-deployment</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-security-deployment</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-elytron-security-common-deployment</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-agroal-deployment</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-panache-common-deployment</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-undertow</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-undertow-deployment</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-resteasy-deployment</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-hibernate-orm-panache-deployment</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-junit5-internal</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>io.rest-assured</groupId>
+            <artifactId>rest-assured</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-test-h2</artifactId>
+            <scope>test</scope>
+        </dependency>
+
+        <!-- Hibernate ORM specific dependencies -->
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-hibernate-orm-deployment</artifactId>
+            <scope>test</scope>
+        </dependency>
+
+        <!-- JDBC driver dependencies -->
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-jdbc-h2-deployment</artifactId>
+            <scope>test</scope>
+        </dependency>
+    </dependencies>
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <configuration>
+                    <annotationProcessorPaths>
+                        <path>
+                            <groupId>io.quarkus</groupId>
+                            <artifactId>quarkus-extension-processor</artifactId>
+                            <version>${project.version}</version>
+                        </path>
+                    </annotationProcessorPaths>
+                </configuration>
+            </plugin>
+        </plugins>
+    </build>
+
+</project>
--- a/extensions/security-jpa/deployment/src/main/java/io/quarkus/security/jpa/deployment/JpaSecurityDefinition.java
+++ b/extensions/security-jpa/deployment/src/main/java/io/quarkus/security/jpa/deployment/JpaSecurityDefinition.java
@@ -1,0 +1,132 @@
+package io.quarkus.security.jpa.deployment;
+
+import java.lang.reflect.Modifier;
+
+import org.jboss.jandex.AnnotationInstance;
+import org.jboss.jandex.AnnotationTarget;
+import org.jboss.jandex.ClassInfo;
+import org.jboss.jandex.DotName;
+import org.jboss.jandex.FieldInfo;
+import org.jboss.jandex.Index;
+import org.jboss.jandex.MethodInfo;
+import org.jboss.jandex.Type;
+
+import io.quarkus.deployment.bean.JavaBeanUtil;
+import io.quarkus.gizmo.AssignableResultHandle;
+import io.quarkus.gizmo.BytecodeCreator;
+import io.quarkus.gizmo.FieldDescriptor;
+import io.quarkus.gizmo.MethodDescriptor;
+import io.quarkus.gizmo.ResultHandle;
+
+public class JpaSecurityDefinition {
+    public static class FieldOrMethod {
+        public final FieldInfo field;
+        public final MethodInfo getter;
+
+        public FieldOrMethod(FieldInfo field, MethodInfo getter) {
+            this.field = field;
+            this.getter = getter;
+        }
+
+        public AnnotationInstance annotation(DotName annotationName) {
+            if (field != null) {
+                return field.annotation(annotationName);
+            }
+            return getter.annotation(annotationName);
+        }
+
+        public String name() {
+            if (field != null) {
+                return field.name();
+            }
+            return JavaBeanUtil.getPropertyNameFromGetter(getter.name());
+        }
+
+        public ResultHandle readValue(BytecodeCreator methodCreator, AssignableResultHandle userVar) {
+            // favour the getter
+            if (getter != null) {
+                return methodCreator.invokeVirtualMethod(MethodDescriptor.of(getter), userVar);
+            }
+            return methodCreator.readInstanceField(FieldDescriptor.of(field), userVar);
+        }
+
+        public Type type() {
+            // FIXME: this doesn't work with generics-inherited types
+            if (field != null) {
+                return field.type();
+            }
+            return getter.returnType();
+        }
+    }
+
+    public final FieldOrMethod username;
+    public final FieldOrMethod password;
+    public final FieldOrMethod roles;
+    public final ClassInfo annotatedClass;
+
+    public JpaSecurityDefinition(Index index,
+            ClassInfo annotatedClass,
+            boolean isPanache,
+            AnnotationTarget usernameFieldOrMethod,
+            AnnotationTarget passwordFieldOrMethod,
+            AnnotationTarget rolesFieldOrMethod) {
+        this.annotatedClass = annotatedClass;
+        this.username = getFieldOrMethod(index, annotatedClass, usernameFieldOrMethod, isPanache);
+        this.password = getFieldOrMethod(index, annotatedClass, passwordFieldOrMethod, isPanache);
+        this.roles = getFieldOrMethod(index, annotatedClass, rolesFieldOrMethod, isPanache);
+    }
+
+    public static FieldOrMethod getFieldOrMethod(Index index, ClassInfo annotatedClass,
+            AnnotationTarget annotatedFieldOrMethod, boolean isPanache) {
+        switch (annotatedFieldOrMethod.kind()) {
+            case FIELD:
+                // try to find a getter for this field
+                FieldInfo field = annotatedFieldOrMethod.asField();
+                return new FieldOrMethod(field,
+                        findGetter(index, annotatedClass, field, Modifier.isPublic(field.flags()) && isPanache));
+            case METHOD:
+                // skip the field entirely
+                return new FieldOrMethod(null, annotatedFieldOrMethod.asMethod());
+            default:
+                throw new IllegalArgumentException(
+                        "annotatedFieldOrMethod must be a field or method: " + annotatedFieldOrMethod);
+        }
+    }
+
+    // FIXME: in order to check for the getter type we need to apply type parameters, that's too complex so assume it matches
+    private static MethodInfo findGetter(Index index, ClassInfo annotatedClass, FieldInfo field, boolean isPanache) {
+        // if it's a panache field, we won't see the getter but it will be there
+        String methodName = "get" + JavaBeanUtil.capitalize(field.name());
+        if (isPanache) {
+            return MethodInfo.create(field.declaringClass(), methodName, new Type[0], field.type(), (short) Modifier.PUBLIC);
+        }
+        return findGetter(index, annotatedClass, methodName);
+    }
+
+    private static MethodInfo findGetter(Index index, ClassInfo annotatedClass, String methodName) {
+        MethodInfo method = annotatedClass.method(methodName);
+        if (method != null) {
+            return method;
+        }
+        DotName superName = annotatedClass.superName();
+        if (superName != null && !superName.equals(QuarkusSecurityJpaProcessor.DOTNAME_OBJECT)) {
+            ClassInfo superClass = index.getClassByName(superName);
+            if (superClass != null) {
+                method = findGetter(index, superClass, methodName);
+                if (method != null) {
+                    return method;
+                }
+            }
+        }
+        for (DotName interfaceName : annotatedClass.interfaceNames()) {
+            ClassInfo interf = index.getClassByName(interfaceName);
+            if (interf != null) {
+                method = findGetter(index, interf, methodName);
+                if (method != null) {
+                    return method;
+                }
+            }
+        }
+        return null;
+    }
+}
--- a/extensions/security-jpa/deployment/src/main/java/io/quarkus/security/jpa/deployment/QuarkusSecurityJpaProcessor.java
+++ b/extensions/security-jpa/deployment/src/main/java/io/quarkus/security/jpa/deployment/QuarkusSecurityJpaProcessor.java
@@ -1,0 +1,345 @@
+package io.quarkus.security.jpa.deployment;
+
+import java.lang.reflect.Modifier;
+import java.util.Collection;
+import java.util.HashSet;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Set;
+import java.util.function.BiConsumer;
+
+import javax.inject.Singleton;
+import javax.persistence.EntityManager;
+import javax.persistence.Query;
+
+import org.hibernate.Session;
+import org.hibernate.SimpleNaturalIdLoadAccess;
+import org.hibernate.annotations.NaturalId;
+import org.jboss.jandex.AnnotationInstance;
+import org.jboss.jandex.AnnotationTarget;
+import org.jboss.jandex.AnnotationValue;
+import org.jboss.jandex.ClassInfo;
+import org.jboss.jandex.DotName;
+import org.jboss.jandex.Index;
+import org.jboss.jandex.Type;
+
+import io.quarkus.arc.deployment.GeneratedBeanBuildItem;
+import io.quarkus.arc.deployment.GeneratedBeanGizmoAdaptor;
+import io.quarkus.arc.deployment.UnremovableBeanBuildItem;
+import io.quarkus.deployment.Capabilities;
+import io.quarkus.deployment.annotations.BuildProducer;
+import io.quarkus.deployment.annotations.BuildStep;
+import io.quarkus.deployment.builditem.ApplicationIndexBuildItem;
+import io.quarkus.deployment.builditem.CapabilityBuildItem;
+import io.quarkus.deployment.builditem.FeatureBuildItem;
+import io.quarkus.gizmo.AssignableResultHandle;
+import io.quarkus.gizmo.BranchResult;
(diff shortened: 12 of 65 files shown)


Sanne:
Some questions / ideas

- do you enforce/expect/require uniqueness of userids in any way?

- Should you load credentials by natural id? The great benefit is you'd not have to generate an HQL query - and parse them each time it's used. I think we should encourage people to use naturalids, or "process" their entities to enable the feature.

- I think I'd allow people to use a named query rather than hard code the query strategy. I guess this could be a future enhancement, but you might want to need take it into account so to allow evolution of the configuration-less use case?

- if you want this enabled and be configuration-less: having just the model won't be enough, you will need at least the datasource configuration.

- Do you have a plan for how this will look like when multiple persistence units / multiple datasources will be allowed?

- I'm confused that you load the credentials model just by userid and then seem to delegate checking the password to other business logic; I remember most other systems regularly require for both the userid and the password to match what's in the store before even allowing to load this. Essentially that's why most systems can't tell if the user exists vs the user exists but you don't have the right password - I don't know why that's the standard but I wonder if that's an important guideline?

- what's the expected semantics of `io.quarkus.elytron.security.jpa.UserEntityCredentialObtainer#obtain` ? It's returning managed entities, it would be far more efficient to use a stateless session. Clearly, not necessary for the implementation in integration testing but it would be good to apply this when generating an implementation. Speaking of, why have all that complexity expressed as gizmo code, couldn't you have a simple abstract class to extend via code generation? Might be easier to make such improvements in the future.

 - should "clear text" really be an option?

FroMage:
Thanks for the great feedback!

> do you enforce/expect/require uniqueness of userids in any way?

I don't, but I can't imagine a user lookup string being anything else than unique. I would rather avoid inspecting JPA annotations as much as possible because I'm sure I'll get them wrong (field/getter/setter/inheritance). What I can do is when generating the query, request at most 2 results and throw if `> 1`?  WDYT? Or do you advise I check for a unicity constraint? How?

> Should you load credentials by natural id? The great benefit is you'd not have to generate an HQL query - and parse them each time it's used. I think we should encourage people to use naturalids, or "process" their entities to enable the feature.

I can totally do that, good idea. Given an annotated field or getter, can I rely on the presence of `@NaturalId` on it? Do I need to check if that annotation is present on any other field to avoid composite natural ids?

> I think I'd allow people to use a named query rather than hard code the query strategy. I guess this could be a future enhancement, but you might want to need take it into account so to allow evolution of the configuration-less use case?

Named query for the entity loading, I guess, right? Not to return all the separate fields (we have the security-jdbc extension for something like that). I guess I can add a `@UserDefinition(namedQuery = "foo")` parameter which would make `@Username` optional. I assume that named query has a single username parameter, right?

> if you want this enabled and be configuration-less: having just the model won't be enough, you will need at least the datasource configuration.

Yes, but it was more a question of wether we need `security.jpa.enabled=true` or not. That's the only security config required ATM, and I think it should not be required.

> Do you have a plan for how this will look like when multiple persistence units / multiple datasources will be allowed?

Not more than for Hibernate ORM with Panache: I expect that alternate PU/DS will be specifiable as a package or entity annotation and we'll be able to use that.

> I'm confused that you load the credentials model just by userid and then seem to delegate checking the password to other business logic; I remember most other systems regularly require for both the userid and the password to match what's in the store before even allowing to load this. Essentially that's why most systems can't tell if the user exists vs the user exists but you don't have the right password - I don't know why that's the standard but I wonder if that's an important guideline?

So, I did wonder about that too, because I always looked users up by both user/pass, but Elytron does not. This is how the JDBC implementation works as well and the underlying API seems to force this. I just followed.

what's the expected semantics of io.quarkus.elytron.security.jpa.UserEntityCredentialObtainer#obtain ? It's returning managed entities, it would be far more efficient to use a stateless session. Clearly, not necessary for the implementation in integration testing but it would be good to apply this when generating an implementation. Speaking of, why have all that complexity expressed as gizmo code, couldn't you have a simple abstract class to extend via code generation? Might be easier to make such improvements in the future.

> No, this method fills a POJO with the results of the JPA session/query. It could use a stateless session if I knew how to use it. Would it be more efficient than this (this is what we generate):

```java
    @Override
    public CredentialHolder obtain(EntityManager entityManager, String username) {
        Query query = entityManager.createQuery("FROM PlainUserEntity WHERE name = :name");
        query.setParameter("name", username);
        List<PlainUserEntity> results = (List<PlainUserEntity>) query.getResultList();
        if (results.isEmpty())
            return null;
        // FIXME: more than one result?
        PlainUserEntity user = results.get(0);

        CredentialHolder ret = new CredentialHolder();

        ret.username = username;

        ret.setClearPassword(user.pass);

        ret.addRoles(user.role);
        return ret;
    }
```

> should "clear text" really be an option?

Yes, for demos. The other security impls support it too.

> setClearPassword .. is this implying it's a requirement to use symmetric cryptography?

Well, it's a clear text password, that's all you can infer. But clear text is indeed symmetric.

stuartwdouglas:
I have a few comments around this:

- What does Elytron add here? The existing Elytron extensions are here as they allowed us to re-use the existing Elytron realms while still integrating with our common security architecture. If you are providing a new Elytron realm then IMHO this should just be a new IdentityProvider rather than an Elytron realm.

- How does a user add an entity here in terms of setting the password? One approach would be to automagically transform the entity so that if you set the password to a plaintext string then we automatically hash it before flushing, but I think that is probably a bit too magic. A simpler approach would be to provide a PasswordService bean that allows the user to turn a plaintext password into a DB password. 

- I don't really like CredentialObtainer, IMHO it is to close to IdentityProvider. e.g. the example above is broadly similar to the following code:

```java
public class TestIdentityProvider implements IdentityProvider<UsernamePasswordAuthenticationRequest> {

    @Inject
    EntityManagerFactory emf;
    
    @Inject
    PasswordService passwordService;

    @Override
    public Class<UsernamePasswordAuthenticationRequest> getRequestType() {
        return UsernamePasswordAuthenticationRequest.class;
    }

    @Override
    public CompletionStage<SecurityIdentity> authenticate(UsernamePasswordAuthenticationRequest request,
            AuthenticationRequestContext context) {
        return context.runBlocking(new Supplier<SecurityIdentity>() {
            @Override
            public SecurityIdentity get() {
                EntityManager entityManager = emf.createEntityManager();
                try {
                    Query query = entityManager.createQuery("FROM PlainUserEntity WHERE name = :name");
                    query.setParameter("name", request.getUsername());
                    List<PlainUserEntity> results = (List<PlainUserEntity>) query.getResultList();
                    if (results.isEmpty())
                        return null;
                    // FIXME: more than one result?
                    PlainUserEntity user = results.get(0);
                    passwordService.verify(user.pass, request.getPassword());
                    return QuarkusSecurityIdentity.builder()
                            .setPrincipal(new QuarkusPrincipal(request.getUsername()))
                            .addRole(user.role)
                            .addCredential(request.getPassword())
                            .build();
                } finally {
                    entityManager.close();
                }

            }
        });
    }
}
```

I think this mostly just adds extra abstraction around our core API's.

Sanne:
> Thanks for the great feedback!

thanks :)

> > do you enforce/expect/require uniqueness of userids in any way?
> 
> I don't, but I can't imagine a user lookup string being anything else than unique. I would rather avoid inspecting JPA annotations as much as possible because I'm sure I'll get them wrong (field/getter/setter/inheritance). What I can do is when generating the query, request at most 2 results and throw if `> 1`? WDYT? Or do you advise I check for a unicity constraint? How?

If you use the natural id it enforces uniqueness. You'd still need to enforce the user to actually make use of the `@NaturalId` annotation but I think that would come ... natural. You wouldn't have to inspect them yourself, other than possibly check there's one to provide a great error message, but even if you don't, if there isn't one it would fail when attempting to load the entity by natural id.

> > Should you load credentials by natural id? The great benefit is you'd not have to generate an HQL query - and parse them each time it's used. I think we should encourage people to use naturalids, or "process" their entities to enable the feature.
> 
> I can totally do that, good idea. Given an annotated field or getter, can I rely on the presence of `@NaturalId` on it? Do I need to check if that annotation is present on any other field to avoid composite natural ids?

Hadn't thought about composite natural ids. Isn't it a given that you need the "unique id" to be the username?
I'd say that if users need to map a complex existing schema doing fancy stuffl like encoding the username into multiple columns, enforce them to still map a single `@NaturalId` property, they can always declare it as a composite embedded object and workaround such limitations.

That way your API also needs to only deal with this single object as id, rather than have alternative code to deal with tuplets.

> 
> > I think I'd allow people to use a named query rather than hard code the query strategy. I guess this could be a future enhancement, but you might want to need take it into account so to allow evolution of the configuration-less use case?
> 
> Named query for the entity loading, I guess, right? Not to return all the separate fields (we have the security-jdbc extension for something like that). I guess I can add a `@UserDefinition(namedQuery = "foo")` parameter which would make `@Username` optional. I assume that named query has a single username parameter, right?

A NamedQuery is very flexible, it can take as many parameters as the user wants.

But I just mentioned them as alternative to natural-ids. I'd say choose one approach, no need to go overboard with flexibility.

> 
> > if you want this enabled and be configuration-less: having just the model won't be enough, you will need at least the datasource configuration.
> 
> Yes, but it was more a question of wether we need `security.jpa.enabled=true` or not. That's the only security config required ATM, and I think it should not be required.

I don't think it can be the default. For example, even if there was a default datasource to bind to, it might not be mean for you to create additional tables in there without explicit opt-in, especially as they might end up storing sensitive information.

> > Do you have a plan for how this will look like when multiple persistence units / multiple datasources will be allowed?
> 
> Not more than for Hibernate ORM with Panache: I expect that alternate PU/DS will be specifiable as a package or entity annotation and we'll be able to use that.

It won't be an entity annotation, that would not be user friendly at all. I hope to write down a proposal soon and discuss about it.. just please keep in mind it will need to play nicely with it.

> 
> > I'm confused that you load the credentials model just by userid and then seem to delegate checking the password to other business logic; I remember most other systems regularly require for both the userid and the password to match what's in the store before even allowing to load this. Essentially that's why most systems can't tell if the user exists vs the user exists but you don't have the right password - I don't know why that's the standard but I wonder if that's an important guideline?
> 
> So, I did wonder about that too, because I always looked users up by both user/pass, but Elytron does not. This is how the JDBC implementation works as well and the underlying API seems to force this. I just followed.

Ok, was just wondering but it's not my field ./

> 
> what's the expected semantics of io.quarkus.elytron.security.jpa.UserEntityCredentialObtainer#obtain ? It's returning managed entities, it would be far more efficient to use a stateless session. Clearly, not necessary for the implementation in integration testing but it would be good to apply this when generating an implementation.
> 
> No, this method fills a POJO with the results of the JPA session/query. It could use a stateless session if I knew how to use it. Would it be more efficient than this (this is what we generate):

Yes it would. A lot.

> 
> ```java
>     @Override
>     public CredentialHolder obtain(EntityManager entityManager, String username) {
>         Query query = entityManager.createQuery("FROM PlainUserEntity WHERE name = :name");
>         query.setParameter("name", username);
>         List<PlainUserEntity> results = (List<PlainUserEntity>) query.getResultList();
>         if (results.isEmpty())
>             return null;
>         // FIXME: more than one result?
>         PlainUserEntity user = results.get(0);
> 
>         CredentialHolder ret = new CredentialHolder();
> 
>         ret.username = username;
> 
>         ret.setClearPassword(user.pass);
> 
>         ret.addRoles(user.role);
>         return ret;
>     }
> ```
> 
> > Speaking of, why have all that complexity expressed as gizmo code, couldn't you have a simple abstract class to extend via code generation? Might be easier to make such improvements in the future.
> 
> I do, this code above is what I generate. It mostly deals with reading fields/getters into the POJO. I could abstract the query part, but it seemed so trival that I didn't think it worth abstracting more.
> 
> > should "clear text" really be an option?
> 
> Yes, for demos. The other security impls support it too.

Ok
 
> > setClearPassword .. is this implying it's a requirement to use symmetric cryptography?
> 
> Well, it's a clear text password, that's all you can infer. But clear text is indeed symmetric.

ok so this method is specific to the "clear text" strategy? Just wondering if it wouldn't get in the way in case of non-symmetric strategies potentially coming later.

Sanne:
P.S. I agree with Stuart that `CredentialObtainer` seems weird looking :)

FroMage:
> What does Elytron add here? The existing Elytron extensions are here as they allowed us to re-use the existing Elytron realms while still integrating with our common security architecture. If you are providing a new Elytron realm then IMHO this should just be a new IdentityProvider rather than an Elytron realm.

I started as a copy of elytron-jpa, then trimmed from there. It still depends on elytron for the specifying the credentials. Can I really do an implementation that doesn't depend on it? Note that I'm not familiar at all with `IdentityProvider`. Where is it documented?

> How does a user add an entity here in terms of setting the password? One approach would be to automagically transform the entity so that if you set the password to a plaintext string then we automatically hash it before flushing, but I think that is probably a bit too magic. A simpler approach would be to provide a PasswordService bean that allows the user to turn a plaintext password into a DB password.

That's not our business. We're in the business of reading entities, not writing them. The user is responsible for storing the entities, and they can use the elytron API to create the right hashes. That's pretty trivial, and I really don't think we should meddle with entity storage logic.

FroMage:
> I don't really like CredentialObtainer, IMHO it is to close to IdentityProvider. e.g. the example above is broadly similar to the following code:

```java
public class TestIdentityProvider implements IdentityProvider<UsernamePasswordAuthenticationRequest> {

    @Inject
    EntityManagerFactory emf;
    
    @Inject
    PasswordService passwordService;

    @Override
    public Class<UsernamePasswordAuthenticationRequest> getRequestType() {
        return UsernamePasswordAuthenticationRequest.class;
    }

    @Override
    public CompletionStage<SecurityIdentity> authenticate(UsernamePasswordAuthenticationRequest request,
            AuthenticationRequestContext context) {
        return context.runBlocking(new Supplier<SecurityIdentity>() {
            @Override
            public SecurityIdentity get() {
                EntityManager entityManager = emf.createEntityManager();
                try {
                    Query query = entityManager.createQuery("FROM PlainUserEntity WHERE name = :name");
                    query.setParameter("name", request.getUsername());
                    List<PlainUserEntity> results = (List<PlainUserEntity>) query.getResultList();
                    if (results.isEmpty())
                        return null;
                    // FIXME: more than one result?
                    PlainUserEntity user = results.get(0);
                    passwordService.verify(user.pass, request.getPassword());
                    return QuarkusSecurityIdentity.builder()
                            .setPrincipal(new QuarkusPrincipal(request.getUsername()))
                            .addRole(user.role)
                            .addCredential(request.getPassword())
                            .build();
                } finally {
                    entityManager.close();
                }

            }
        });
    }
}
```

OK, I'll admit I'd never heard of this API to define a pluggable security provider. 
- Is it documented somewhere? 
- Is is auto-discovered (one of the nice points of `CredentialsObtainer`)?
- What's `PasswordService`?
- How do we specify anything else than plan-text passwords?

Otherwise, I agree it's the same idea as `CredentialsObtainer`, except way more verbose, but I can get around that by providing a supertype to override. The one thing I don't understand is whether it does the password checking or not. `CredentialsObtainer` does not: it's only responsible for data extraction.

FroMage:
@Sanne 

> Hadn't thought about composite natural ids. Isn't it a given that you need the "unique id" to be the username?
> I'd say that if users need to map a complex existing schema doing fancy stuffl like encoding the username into multiple columns, enforce them to still map a single @NaturalId property, they can always declare it as a composite embedded object and workaround such limitations.

> That way your API also needs to only deal with this single object as id, rather than have alternative code to deal with tuplets.

We really look entities up by username, which can only be a `String`, so I don't know how composite embedded objects can work.

> A NamedQuery is very flexible, it can take as many parameters as the user wants.

OK, but the only dynamic param we can pass is the username. Unless we have a source of other params, but I wonder which.

> But I just mentioned them as alternative to natural-ids. I'd say choose one approach, no need to go overboard with flexibility.

OK let's skip it for now then ;)

> I don't think it can be the default. For example, even if there was a default datasource to bind to, it might not be meant for you to create additional tables in there without explicit opt-in, especially as they might end up storing sensitive information.

We're not creating any tables in the security extension. We're just reading how to map the user model to our security source.

>> No, this method fills a POJO with the results of the JPA session/query. It could use a stateless session if I knew how to use it. Would it be more efficient than this (this is what we generate):

> Yes it would. A lot.

Please show me equivalent code with stateless session then :)

>>> setClearPassword .. is this implying it's a requirement to use symmetric cryptography?

>> Well, it's a clear text password, that's all you can infer. But clear text is indeed symmetric.

> ok so this method is specific to the "clear text" strategy? Just wondering if it wouldn't get in the way in case of non-symmetric strategies potentially coming later.

The hashed password code we generate is:

```java
ret.setMdfPassword(user.pass);
```

The generated code depends on how the `@Password` annotation is declared.

Sanne:
ok so, summarizing as I feel you only have 2 open points still:

embeeded natural ids:
yes I also was thinking only for String, but you brought the composite-key possibility up. Ignore it then and assume there's a unique string?
That was my suggestion, as the user can bypass the limitation by using a more complex mapping I think - but also it feels very far fetched.

Just use NaturalId and expect a String for username.

> Please show me equivalent code with stateless session then :)

    try (StatelessSession statelessSession = sessionFactory.openStatelessSession()) {
        Person p = statelessSession.get( Person.class, id );
        return p;
    }

Unfortunatley I just noticed the stateless session API doesn't do natural id loading. Seems like an oversight, but no big deal just forget about my suggestion to use it and go for named queries instead?

Would be cool to add some helpers in Panache?

FroMage:
> Unfortunatley I just noticed the stateless session API doesn't do natural id loading. Seems like an oversight, but no big deal just forget about my suggestion to use it and go for named queries instead?

- Do stateless sessions allow me to have the roles in a relation?
- I can revert to `EntityManager` if using a `@NaturalId`
- I'd rather also keep the current option to just annotate a field with `@Username` for lookup: I find named queries are harder to write than just annotating the password field. I'd rather add the named query support later.

> Would be cool to add some helpers in Panache?

I think there's an open issue for this.

stuartwdouglas:
IdentityProvider is the core of our security API. It was documented in the security design doc from Paris (you were there when we designed it). 

Originally the plan was not to use Elytron, but due to time constraints the existing Elytron stuff was modified to just use our new core API's. This is an extensible API, the user can simply include it in their application and make it a CDI bean and it will be picked up.

This has no support for passwords though, so we should use the Elytron stuff for that, but ideally behind some kind of easy to use API that the users can just use directly. Elytron is modular, so you can include the password support without also including the realm support.

FroMage:
> IdentityProvider is the core of our security API. It was documented in the security design doc from Paris (you were there when we designed it).

I didn't mean to imply it was not justified. It's just that I forgot about this particular interface, and didn't read anything about it when I reviewed the security guides a few weeks ago. We should document it in our guides.

OK lemme rewrite that in terms of `IdentityProvider` then.

Sanne:
> > Unfortunatley I just noticed the stateless session API doesn't do natural id loading. Seems like an oversight, but no big deal just forget about my suggestion to use it and go for named queries instead?
> 
> * Do stateless sessions allow me to have the roles in a relation?
> * I can revert to `EntityManager` if using a `@NaturalId`
> * I'd rather also keep the current option to just annotate a field with `@Username` for lookup: I find named queries are harder to write than just annotating the password field. I'd rather add the named query support later.
> 
> > Would be cool to add some helpers in Panache?
> 
> I think there's an open issue for this.

Ok to go with NaturalId then. But no you won't be ableto use `EntityManager` as it doesn't have the method
https://docs.jboss.org/hibernate/orm/5.4/userguide/html_single/Hibernate_User_Guide.html#naturalid-api

You can get most of the performance benefits of a StatelessSession by setting flush mode to manual, and set the session in read-only mode.
I'd recommend doing it also for safety (not just performance) so to make sure you don't unintentionally modify the database.

FroMage:
OK I'll try.

nimo23:
Why not using `Java EE 8 Security API` behind the scenes? It already has Built-in Identity Stores, e.g. `@DatabaseIdentityStoreDefinition`. Why not providing the `@Username`, `@Password`, `@Roles` an an extension of `Java EE 8 Security API`?

FroMage:
@Sanne I can't actually use a stateless session if my roles are in another table:

```
2019-12-04 17:19:51,217 ERROR [io.qua.ver.htt.run.QuarkusErrorHandler] (executor-thread-1) HTTP Request to /jaxrs-secured/parameterized-paths/my/banking/view failed, error id: 9bb728e4-477a-4961-9074-5ed2ff311af1-24: org.hibernate.LazyInitializationException: Unable to perform requested lazy initialization [io.quarkus.security.jpa.PanacheUserEntity.roles] - no session and settings disallow loading outside the Session
	at org.hibernate.bytecode.enhance.spi.interceptor.EnhancementHelper.throwLazyInitializationException(EnhancementHelper.java:196)
	at org.hibernate.bytecode.enhance.spi.interceptor.EnhancementHelper.performWork(EnhancementHelper.java:86)
	at org.hibernate.bytecode.enhance.spi.interceptor.LazyAttributeLoadingInterceptor.loadAttribute(LazyAttributeLoadingInterceptor.java:76)
	at org.hibernate.bytecode.enhance.spi.interceptor.LazyAttributeLoadingInterceptor.fetchAttribute(LazyAttributeLoadingInterceptor.java:72)
	at org.hibernate.bytecode.enhance.spi.interceptor.LazyAttributeLoadingInterceptor.handleRead(LazyAttributeLoadingInterceptor.java:53)
	at org.hibernate.bytecode.enhance.spi.interceptor.AbstractInterceptor.readObject(AbstractInterceptor.java:153)
	at io.quarkus.security.jpa.PanacheUserEntity.$$_hibernate_read_roles(PanacheUserEntity.java)
	at io.quarkus.security.jpa.PanacheUserEntity.getRoles(PanacheUserEntity.java)
	at io.quarkus.security.jpa.PanacheUserEntity__JpaIdentityProviderImpl.authenticate(PanacheUserEntity__JpaIdentityProviderImpl.zig:76)
```

Sanne:
@FroMage sure, see my last comment  - I thought we had agreed to not use stateless sessions but rather prefer using natural Ids ?

The need to handle relations is one of the reasons.

FroMage:
I guess I'm trying things out in comment order :(

Sanne:
very efficient :dancer: have fun :)

sorry about that, I hadn't initially thought about the roles being mapped there as well.

FroMage:
OK, I've added support for `@NaturalId` @Sanne (if it's not there it will revert to a normal query), and turned the session read-only with no flush. We'll deal with named queries if/when someone requests it.

@stuartwdouglas I've rewritten in terms of `IdentityProvider`.

As a result, I've stopped depending on `quarkus-security-elytron-deployment` (but still on the runtime because it contains SVM substs I need), and also there's no config anymore: if there's an annotated entity, it generates an `IdentityProvider`, if not: it doesn't do anything.

If you guys think it's fine, I'll write the docs.

FroMage:
@nimo23 
> It already has Built-in Identity Stores, e.g. @DatabaseIdentityStoreDefinition

Given this example:

```java
@DatabaseIdentityStoreDefinition(
    callerQuery = "#{'select password from caller where name = ?'}",
    groupsQuery = "select group_name from caller_groups where caller_name = ?",
    hashAlgorithm = Pbkdf2PasswordHash.class,
    priorityExpression = "#{100}",
    hashAlgorithmParameters = {
        "Pbkdf2PasswordHash.Iterations=3072",
        "${applicationConfig.dyna}"
    }
)
```

It looks like we already support this in our `quarkus-security-elytron-jdbc` extension. I don't think an annotation is the right place for this sort of config.

We could probably support our annotated entities on top of that, but I'd much rather plug into the existing Quarkus security APIs than EE security which comes with a lot of baggage.

FroMage:
It would be nice to get this in 1.1 @gsmet if I can get it reviewed and finished in time.

FroMage:
Just updated the PR description with latest features.

FroMage:
Rebased on master, let's see if it helps wrt CI. Without feedback from @Sanne and @stuartwdouglas I don't think it can make 1.1.

extensions/security-jpa/deployment/src/main/java/io/quarkus/security/jpa/deployment/QuarkusSecurityJpaProcessor.java
```diff
@@ -0,0 +1,370 @@
...
+            BuildProducer<GeneratedBeanBuildItem> beanProducer, Set<String> panacheClasses) {
+        ClassOutput classOutput = new ClassOutput() {
+            @Override
+            public void write(String name, byte[] data) {
+                beanProducer.produce(new GeneratedBeanBuildItem(name, data));
+            }
+
+            @Override
+            public Writer getSourceWriter(String className) {
+                return new OutputStreamWriter(System.err) {
```
stuartwdouglas: Is this supposed to be here or was this a debugging thing? If it is not supposed to be there you can use GeneratedBeanGizmoAdaptor

extensions/security-jpa/runtime/src/main/java/io/quarkus/security/jpa/JpaRecorder.java
```diff
@@ -0,0 +1,16 @@
...
+
+import org.wildfly.security.password.WildFlyElytronPasswordProvider;
+
+import io.quarkus.runtime.annotations.Recorder;
+
+@Recorder
+public class JpaRecorder {
+
+    public void registerSecurityProvider() {
+        Security.addProvider(new WildFlyElytronPasswordProvider());
```
stuartwdouglas: This is also registered by io.quarkus.elytron.security.runtime.ElytronRecorder#registerPasswordProvider, I don't think we want it twice (not sure if it matters)? If registering it twice causes problems we probably need some build items to manage the provider registration.

extensions/security-jpa/runtime/src/main/java/io/quarkus/security/jpa/JpaSecurityConfig.java
```diff
@@ -0,0 +1,17 @@
+package io.quarkus.security.jpa;
+
+import io.quarkus.runtime.annotations.ConfigPhase;
+import io.quarkus.runtime.annotations.ConfigRoot;
+
+/**
+ * A configuration object for JPA-based security.
+ */
+@ConfigRoot(name = "security.jpa", phase = ConfigPhase.BUILD_AND_RUN_TIME_FIXED)
+public class JpaSecurityConfig {
```
stuartwdouglas: What is the point of the empty config? I think this should at least have the unit name.

stuartwdouglas (changes requested):
I think it looks good, the only think I think it is missing (other than docs) is an easy way to generate the passwords. If I am implementing the 'new user' screen in my app there should be a really simple way to set the password on the entity, including generating the salt etc, I should not have to know anything about MCF passwords.

extensions/security-jpa/deployment/src/main/java/io/quarkus/security/jpa/deployment/QuarkusSecurityJpaProcessor.java
FroMage: I will remove this.

extensions/security-jpa/runtime/src/main/java/io/quarkus/security/jpa/JpaRecorder.java
FroMage: IIRC it is conditionally added only if you have a realm, which is not the case here.

extensions/security-jpa/runtime/src/main/java/io/quarkus/security/jpa/JpaSecurityConfig.java
FroMage: Yeah, I will remove the empty config. I wasn't sure if I was going to need config or not, but I can't think of anything.

FroMage:
Finally got the stack trace of the failing test. This actually doesn't seem to be related to native, but the test passes in the IDE, but not on the Maven CLI.

```
sun.misc.Unsafe.park(Native Method)
java.util.concurrent.locks.LockSupport.park(LockSupport.java:175)
java.util.concurrent.CompletableFuture$Signaller.block(CompletableFuture.java:1693)
java.util.concurrent.ForkJoinPool.managedBlock(ForkJoinPool.java:3323)
java.util.concurrent.CompletableFuture.waitingGet(CompletableFuture.java:1729)
java.util.concurrent.CompletableFuture.get(CompletableFuture.java:1895)
io.quarkus.runner.RuntimeClassLoader.findClass(RuntimeClassLoader.java:214)
io.quarkus.runner.RuntimeClassLoader.loadClass(RuntimeClassLoader.java:179)
java.lang.ClassLoader.loadClass(ClassLoader.java:357)
java.lang.Class.forName0(Native Method)
java.lang.Class.forName(Class.java:348)
org.objectweb.asm.ClassWriter.getCommonSuperClass(ClassWriter.java:935)
org.objectweb.asm.SymbolTable.addMergedType(SymbolTable.java:1200)
org.objectweb.asm.Frame.merge(Frame.java:1299)
org.objectweb.asm.Frame.merge(Frame.java:1197)
org.objectweb.asm.MethodWriter.computeAllFrames(MethodWriter.java:1607)
org.objectweb.asm.MethodWriter.visitMaxs(MethodWriter.java:1543)
org.objectweb.asm.ClassReader.readCode(ClassReader.java:2431)
org.objectweb.asm.ClassReader.readMethod(ClassReader.java:1283)
org.objectweb.asm.ClassReader.accept(ClassReader.java:688)
org.objectweb.asm.ClassReader.accept(ClassReader.java:400)
io.quarkus.hibernate.orm.deployment.HibernateEntityEnhancer$HibernateEnhancingClassVisitor.visitEnd(HibernateEntityEnhancer.java:67)
org.objectweb.asm.ClassReader.accept(ClassReader.java:692)
org.objectweb.asm.ClassReader.accept(ClassReader.java:400)
io.quarkus.runner.RuntimeClassLoader.handleTransform(RuntimeClassLoader.java:430)
io.quarkus.runner.RuntimeClassLoader.findClass(RuntimeClassLoader.java:225)
io.quarkus.runner.RuntimeClassLoader.loadClass(RuntimeClassLoader.java:179)
java.lang.ClassLoader.loadClass(ClassLoader.java:357)
java.lang.Class.forName0(Native Method)
java.lang.Class.forName(Class.java:348)
io.quarkus.hibernate.orm.runtime.service.FlatClassLoaderService.classForName(FlatClassLoaderService.java:36)
org.hibernate.boot.model.source.internal.annotations.AnnotationMetadataSourceProcessorImpl.<init>(AnnotationMetadataSourceProcessorImpl.java:109)
org.hibernate.boot.model.process.spi.MetadataBuildingProcess$1.<init>(MetadataBuildingProcess.java:155)
org.hibernate.boot.model.process.spi.MetadataBuildingProcess.complete(MetadataBuildingProcess.java:149)
io.quarkus.hibernate.orm.runtime.boot.FastBootMetadataBuilder.build(FastBootMetadataBuilder.java:342)
io.quarkus.hibernate.orm.runtime.PersistenceUnitsHolder.createMetadata(PersistenceUnitsHolder.java:111)
io.quarkus.hibernate.orm.runtime.PersistenceUnitsHolder.constructMetadataAdvance(PersistenceUnitsHolder.java:84)
io.quarkus.hibernate.orm.runtime.PersistenceUnitsHolder.initializeJpa(PersistenceUnitsHolder.java:47)
io.quarkus.hibernate.orm.runtime.HibernateOrmRecorder$4.created(HibernateOrmRecorder.java:75)
io.quarkus.arc.runtime.ArcRecorder.initBeanContainer(ArcRecorder.java:101)
io.quarkus.deployment.steps.ArcProcessor$generateResources24.deploy_0(ArcProcessor$generateResources24.zig:442)
io.quarkus.deployment.steps.ArcProcessor$generateResources24.deploy(ArcProcessor$generateResources24.zig:36)
io.quarkus.runner.ApplicationImpl.<clinit>(ApplicationImpl.zig:378)
java.lang.Class.forName0(Native Method)
```

Apparently it's a deadlock in the class loading, in an area (find common supertype) we recently touched, right @stuartwdouglas ?

FroMage:
Rebased on master which has a fix for that.

extensions/security-jpa/runtime/src/main/java/io/quarkus/security/jpa/JpaRecorder.java
FroMage: So it's not conditional anymore, but I'd have to make my extension depend on `quarkus-security-elytron` to get it. Do you think that's fine?

FroMage: I removed my recorder and depend on that extension. I guess it should be fine.

extensions/security-jpa/deployment/src/main/java/io/quarkus/security/jpa/deployment/QuarkusSecurityJpaProcessor.java
FroMage: Removed

extensions/security-jpa/runtime/src/main/java/io/quarkus/security/jpa/JpaSecurityConfig.java
FroMage: Removed

FroMage:
>  think it is missing (other than docs) is an easy way to generate the passwords. If I am implementing the 'new user' screen in my app there should be a really simple way to set the password on the entity, including generating the salt etc, I should not have to know anything about MCF passwords.

Done:

```java
        String testPassword = "fubar";

// this is the hash to store in DB        
        String testPasswordHash = JpaSecurityUtil.bcryptHash(testPassword);
        
        PasswordGuessEvidence correctPasswordEvidence = new PasswordGuessEvidence(testPassword.toCharArray());
        PasswordGuessEvidence incorrectPasswordEvidence = new PasswordGuessEvidence("stef".toCharArray());
        PasswordCredential producedPasswordCredential = new PasswordCredential(ModularCrypt.decode(testPasswordHash));
        Assertions.assertTrue(producedPasswordCredential.verify(correctPasswordEvidence));
        Assertions.assertFalse(producedPasswordCredential.verify(incorrectPasswordEvidence));

```

FroMage:
Tomorrow I'll write the docs and let's get rid of this PR.

FroMage:
I guess the `JpaSecurityUtil` file could do with a review by @dmlloyd :)

extensions/security-jpa/runtime/src/main/java/io/quarkus/security/jpa/JpaRecorder.java
dmlloyd: We might consider switching this to be ServiceLoader-based.  All of the Elytron providers are available via `ServiceLoader.load(Provider.class)` normally; we could have a startup step where we register all providers on the class path at boot.

The JDK has already starting moving to this behavior in fact, and moving away from requiring global provider registration in some cases [1].  So really, we probably should look at enhancing the Elytron APIs to use this approach as well.

FroMage:
OK it's ready: it has javadocs, the util class, a guide, a quickstart, tests, the works.

sberyozkin:
@FroMage It looks good. What is the migration path for `quarkus-elytron-security-jdbc` users with regard to the password protection algorithms ? I.e, should the two extensions, "-jdbc" and this new one, support a common way to protect the passwords or it is not possible (referring here to the MCF passwords) ? thanks

FroMage:
@sberyozkin I don't know if we should have a migration strategy. I do think "clear text" should not have been the default and that bcrypt in MCF is the only default that has made sense for the last few years. I was very frustrated that the JDBC security seemed to force me to split my hashed password in many columns, but perhaps I was missing an option.

I'm not too tempted to add support for split passwords in this new extension because I think it's a bad design in many cases and wanted to wait for users to request it (or not). Perhaps we'l have no choice, but I'd rather not start with it to not encourage people to add complexity to their tables.

sberyozkin:
@FroMage Sure that sounds reasonable. Might make sense to mention in the docs why the new protection mechanism offered by this new extension should be preferred. thanks

FroMage:
@stuartwdouglas I think this is waiting for your review :)

extensions/security-jpa/runtime/src/main/java/io/quarkus/security/jpa/runtime/JpaSecurityUtil.java
```diff
@@ -0,0 +1,90 @@
...
+import org.wildfly.security.password.WildFlyElytronPasswordProvider;
+import org.wildfly.security.password.interfaces.BCryptPassword;
+import org.wildfly.security.password.spec.EncryptablePasswordSpec;
+import org.wildfly.security.password.spec.IteratedSaltedPasswordAlgorithmSpec;
+import org.wildfly.security.password.util.ModularCrypt;
+
+/**
+ * Utility class used to produce bcrypt hashes using the Modular Crypt Format.
+ */
+public class JpaSecurityUtil {
```
danielpetisme: Based on #5113 and #5667 would it make sense to move an rename the class to something les "JPA related"? Like `io.quarkus.security.CryptoUtil` or similar?

stuartwdouglas (approved):
Other than my minor comment I think this is good to go.

core/deployment/src/main/java/io/quarkus/deployment/Capabilities.java
```diff
@@ -24,6 +24,7 @@
     public static final String SECURITY = "io.quarkus.security";
     public static final String SECURITY_ELYTRON_OAUTH2 = "io.quarkus.elytron.security.oauth2";
     public static final String SECURITY_ELYTRON_JDBC = "io.quarkus.elytron.security.jdbc";
+    public static final String SECURITY_ELYTRON_JPA = "io.quarkus.elytron.security.jpa";
```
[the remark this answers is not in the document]
FroMage: Ah, right.

extensions/security-jpa/runtime/src/main/java/io/quarkus/security/jpa/runtime/JpaSecurityUtil.java
FroMage: That's a good point, I'll try to move it.

FroMage: Given that it depends on Elytron and `quarkus-security` does not depend on it, I can move it to `quarkus-elytron-security-common` which ATM only holds the GraalVM substs that I need anyway.

I don't think it's worth adding it to ``quarkus-elytron-security` itself since it comes with baggage.

WDYT @stuartwdouglas that place would work for you?

stuartwdouglas: Works for me.

sberyozkin:
@FroMage Hi Stephane, can you resolver the conflicts please ? Stuart is good with this PR so we can just merge it

core/deployment/src/main/java/io/quarkus/deployment/Capabilities.java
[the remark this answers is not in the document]
FroMage: Done.

extensions/security-jpa/runtime/src/main/java/io/quarkus/security/jpa/runtime/JpaSecurityUtil.java
FroMage: Done. Had to move some things that init elytron/password and register them as service at runtime there, but I think it's reasonable: https://github.com/quarkusio/quarkus/pull/5660/commits/d80322935c72c81a61eb305d0de2013f3894501d

FroMage:
Done. Perhaps needs a final review of https://github.com/quarkusio/quarkus/pull/5660/commits/d80322935c72c81a61eb305d0de2013f3894501d by whoever is responsible for `quarkus-elytron-security` and `quarkus-elytron-security-common`?

sberyozkin:
@FroMage It looks OK. Can you please squash the commits as well

stuartwdouglas:
There is a failure that may be related. I have seen is before related to registering security providers at static init time rather than runtime, but I am not sure why these changes would have caused it.

FroMage:
It must be related because infinispan is the only consumer of `quarkus-elytron-security-common` which I touched. But that NPE sure is not helping find the cause.

stuartwdouglas:
To debug stuff like this load up the graal sources, and run the native build with -Dquarkus.native.debug-build-process=true then attach to the native image compilation with a breakpoint at the problem line. I am pretty sure it is because something is registering a security provider at static init, but not sure why these changes would cause it.

FroMage:
It works here after I rebased on master and used the latest graalvm 19.3.1, so repushed and let's see what CI says.

gsmet (changes requested):
This needs a squash before merging.

extensions/security-jpa/runtime/src/main/resources/META-INF/quarkus-extension.json
```diff
@@ -0,0 +1,9 @@
+
+{
+    "name": "Quarkus Security JPA Support",
+    "labels": [
+        "security",
+        "jpa"
+    ],
+    "guide": "https://quarkus.io/guides/security-jpa"
```
gsmet: This should be a YAML file now.

gsmet (changes requested):

extensions/security-jpa/runtime/src/main/resources/META-INF/quarkus-extension.json
FroMage: OK thanks, fixed. Damn this was a long-standing PR :(

FroMage:
Actually, does it make sense to make the extension status `stable` @gsmet ? Should it be preview or something else?

FroMage:
OK switched extension status to preview. We should be able to merge this without re-running CI, no?

stuartwdouglas:
This needs a rebase

FroMage:
OK, here's another rebase.

FroMage:
This never ends. Now the tests fail with:

```
2020-01-23 09:31:50,082 ERROR [io.qua.ver.htt.run.QuarkusErrorHandler] (executor-thread-1) HTTP Request to /jaxrs-secured/roles-class failed, error id: 4afd1ea0-b20b-40f2-9278-ad328ca2aae0-8: java.lang.ClassCastException: org.wildfly.security.password.impl.PasswordFactorySpiImpl cannot be cast to org.wildfly.security.password.PasswordFactorySpi
```

@stuartwdouglas could it be related to the CP changes?

stuartwdouglas:
Probably, if I don't get to it tonight I will have a look tomorrow.

FroMage:
Thanks.

stuartwdouglas:
It's something to do with the password providers being loaded with the wrong class loader. It may be that we need to make sure Elytron is always loaded in the system class loader, or there may be a simpler fix, I will look tomorrow.

dmlloyd:
I think it's likely the opposite problem: there is more than one copy of the API JAR on the class path.  If we're supporting distinct class loaders, it's critically important that we do not allow classes to be duplicated between them.

In JBoss Modules we solved this by unconditionally using child-first delegation for every class loader, but also deleting any package path from the child that is also present in any parent, so it behaves like parent-first but without the drawbacks inherent in that delegation model.

stuartwdouglas:
Classes have to be duplicated, it is the old way we can support transformers, and make sure that our class path is 'clean' and not linked against old versions of API's that just happen to be in the parent class loader (e.g. gradle, although this one can likely be worked around in other ways, so is not a hard requirement).  

Basically the main requirements that lead to duplicated classes are:
- JUnit loads classes eagerly, so classes present in the system class loader cannot be reliably transformed, this needs them to be duplicated into a new class loader
- Also any dependency may also need to be transformed. These suffer from the same eager loading issues, but also need to be loaded into a 'flatish' class path, where they have two way visibility with other dependencies. 
- Isolation between deployment and runtime to allow for deployment to have additional dependencies. This was Sanne's use case, but this also means that the test environment is much more similar to the production environment, as static state left over from deployment is not visible at runtime.

Some things obviously can't be duplicated though and need to be in the system class loader (e.g. JBoss Logging), so there is a facility to allow that. I think it will likely turn out that the solution here is to force security artifacts to be parent first.

stuartwdouglas:
@FroMage I have pushed a commit that should help

FroMage:
Well, now it's the Windows build that fails:

```
[ERROR] testThatTheApplicationIsReloadedOnKotlinChange  Time elapsed: 121.786 s  <<< ERROR!
```

That can't be related to this PR, can it?

FroMage:
YAR (Yet Another Rebase).

FroMage:
YAR.

bom/deployment/pom.xml
```diff
@@ -612,6 +612,11 @@
                 <artifactId>quarkus-elytron-security-jdbc-deployment</artifactId>
                 <version>${project.version}</version>
             </dependency>
+            <dependency>
+                <groupId>io.quarkus</groupId>
+                <artifactId>quarkus-security-jpa-deployment</artifactId>
```
gsmet: I think the name here is incorrect. At least it's inconsistent with the other `elytron-security` extensions (they have elytron in the name).

bom/runtime/pom.xml
```diff
@@ -733,6 +733,11 @@
                 <artifactId>quarkus-elytron-security-jdbc</artifactId>
                 <version>${project.version}</version>
             </dependency>
+            <dependency>
+                <groupId>io.quarkus</groupId>
+                <artifactId>quarkus-security-jpa</artifactId>
```
gsmet: Same here.

docs/src/main/asciidoc/security-jpa.adoc
```diff
@@ -0,0 +1,360 @@
...
+== Creating the Maven Project
+
+First, we need a new project. Create a new project with the following command:
+
+[source, subs=attributes+]
+----
+mvn io.quarkus:quarkus-maven-plugin:{quarkus-version}:create \
+    -DprojectGroupId=org.acme \
+    -DprojectArtifactId=security-jpa-quickstart \
+    -Dextensions="elytron-security-jpa, jdbc-postgresql, resteasy, hibernate-orm-panache"
```
gsmet: And see here you have the right name.

docs/src/main/asciidoc/security-jpa.adoc
```diff
@@ -0,0 +1,360 @@
...
+
+import javax.persistence.Entity;
+import javax.persistence.Table;
+
+import io.quarkus.hibernate.orm.panache.PanacheEntity;
+import io.quarkus.elytron.security.common.runtime.BcryptUtil;
+import io.quarkus.security.jpa.runtime.Password;
+import io.quarkus.security.jpa.runtime.Roles;
+import io.quarkus.security.jpa.runtime.UserDefinition;
+import io.quarkus.security.jpa.runtime.Username;
```
gsmet: Everything that is user API should be at the root, not in the `runtime` package. `runtime` is an internal package.

The package should also be `elytron.security.jpa` to be consistent with the other Elytron extensions.

docs/src/main/asciidoc/security-jpa.adoc
```diff
@@ -0,0 +1,360 @@
...
+    }
+}
+
+----
+
+The `elytron-security-jpa` extension is only initialized if there is a single entity annotated with `@UserDefinition`.
+
+<1> This annotation must be present on a single entity. It can be a regular Hibernate ORM entity or a Hibernate ORM with Panache entity as in this example.
+<2> This indicates the field used for the user name.
+<3> This indicates the field used for the password. This defaults to using bcrypt hashed passwords, but you can add also configure it for clear text passwords.
```
gsmet:
```suggestion
<3> This indicates the field used for the password. This defaults to using bcrypt hashed passwords, but you can also configure it for clear text passwords.
```

docs/src/main/asciidoc/security-jpa.adoc
```diff
@@ -0,0 +1,360 @@
...
+----
+quarkus.datasource.url=jdbc:postgresql:elytron_security_jpa
+quarkus.datasource.driver=org.postgresql.Driver
+quarkus.datasource.username=quarkus
+quarkus.datasource.password=quarkus
+
+quarkus.hibernate-orm.database.generation=drop-and-create
+----
+
+In our context, we are using PostgreSQL as identity store. The database schema is created by hibernate automatically
```
gsmet:
```suggestion
In our context, we are using PostgreSQL as identity store. The database schema is created by Hibernate ORM automatically
```

docs/src/main/asciidoc/security-jpa.adoc
```diff
@@ -0,0 +1,360 @@
...
+        User.deleteAll();
+        User.add("admin", "admin", "admin");
+        User.add("user", "user", "user");
+    }
+}
+----
+
+[NOTE]
+====
+It is probably useless but we kindly remind you that you must not store clear-text passwords in production environment ;-).
```
gsmet:
```suggestion
It is probably useless but we kindly remind you that you must not store clear-text passwords in production environments ;-).
```

docs/src/main/asciidoc/security-jpa.adoc
```diff
@@ -0,0 +1,360 @@
...
+    public String role;
+}
+----
+
+== Password storage and hashing
+
+By default, we consider passwords to be stored hashed with https://en.wikipedia.org/wiki/Bcrypt[bcrypt] under the 
+https://en.wikipedia.org/wiki/Crypt_(C)[Modular Crypt Format] (MCF).
+
+When you need to create such a hashed password we provide the convenient `String BcryptUtil.bcryptHash(String password)`
```
gsmet: If we expect the user to use it, we will need that one to be API too so outside of the runtime package.

docs/src/main/asciidoc/security.adoc
```diff
@@ -22,6 +22,9 @@ in order for Quarkus to know how to find the authentication information to check
 |link:security-properties[quarkus-elytron-security-properties-file]
 |Provides support for simple properties files that can be used for testing security. This supports both embedding user info in `application.properties` and standalone properties files.
 
+|link:security-jpa[quarkus-security-jpa]
```
gsmet:
```suggestion
|link:security-jpa[quarkus-elytron-security-jpa]
```

extensions/elytron-security-common/deployment/pom.xml
```diff
@@ -18,6 +18,10 @@
             <groupId>io.quarkus</groupId>
             <artifactId>quarkus-core-deployment</artifactId>
         </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-elytron-security-common</artifactId>
+        </dependency>
```
gsmet: Good catch.

extensions/elytron-security-common/runtime/src/main/java/io/quarkus/elytron/security/common/runtime/BcryptUtil.java
```diff
@@ -0,0 +1,90 @@
...
+        if (iterationCount <= 0)
+            throw new IllegalArgumentException("Iteration count must be greater than zero");
+        Objects.requireNonNull(password, "password is required");
+        Objects.requireNonNull(salt, "salt is required");
+        if (salt.length != BCryptPassword.BCRYPT_SALT_SIZE)
+            throw new IllegalArgumentException("Salt lenght must be exactly " + BCryptPassword.BCRYPT_SALT_SIZE + " bytes");
+
+        PasswordFactory passwordFactory;
+        try {
+            passwordFactory = PasswordFactory.getInstance(BCryptPassword.ALGORITHM_BCRYPT, provider);
```
gsmet: Can't that be static and called once and for all? It might or might not be slow.

extensions/elytron-security-common/runtime/src/main/java/io/quarkus/elytron/security/common/runtime/BcryptUtil.java
```diff
@@ -0,0 +1,90 @@
...
+     * @throws IllegalArgumentException if the iterationCount parameter is negative or zero, or if the salt length is not equal
+     *         to 16
+     */
+    public static String bcryptHash(String password, int iterationCount, byte[] salt) {
+        if (iterationCount <= 0)
+            throw new IllegalArgumentException("Iteration count must be greater than zero");
+        Objects.requireNonNull(password, "password is required");
+        Objects.requireNonNull(salt, "salt is required");
+        if (salt.length != BCryptPassword.BCRYPT_SALT_SIZE)
+            throw new IllegalArgumentException("Salt lenght must be exactly " + BCryptPassword.BCRYPT_SALT_SIZE + " bytes");
```
gsmet: Here and also above, please use curly braces.

gsmet:
```suggestion
            throw new IllegalArgumentException("Salt length must be exactly " + BCryptPassword.BCRYPT_SALT_SIZE + " bytes");
```

extensions/panache/hibernate-orm-panache/deployment/src/main/java/io/quarkus/hibernate/orm/panache/deployment/PanacheResourceProcessor.java
```diff
@@ -115,6 +117,8 @@ void build(CombinedIndexBuildItem index,
         for (String modelClass : modelClasses) {
             transformers.produce(new BytecodeTransformerBuildItem(modelClass, modelEnhancer));
         }
+        if (!modelClasses.isEmpty())
+            entityClasses.produce(new PanacheEntityClassesBuildItem(modelClasses));
```
gsmet: Same here about curly braces. Maybe configure your Quarkus workspace for it?

extensions/panache/mongodb-panache/deployment/src/main/java/io/quarkus/mongodb/panache/deployment/PanacheResourceProcessor.java
```diff
@@ -163,6 +165,8 @@ void build(CombinedIndexBuildItem index,
             // Register for building the property mapping cache
             propertyMappingClass.produce(new PropertyMappingClassBuildStep(modelClass));
         }
+        if (!modelClasses.isEmpty())
+            entityClasses.produce(new PanacheEntityClassesBuildItem(modelClasses));
```
gsmet: And here again.

extensions/pom.xml
```diff
@@ -97,6 +97,7 @@
 
         <!-- Security -->
         <module>security</module>
+        <module>security-jpa</module>
```
gsmet:
```suggestion
        <module>elytron-security-jpa</module>
```

extensions/security-jpa/deployment/pom.xml
```diff
@@ -0,0 +1,114 @@
...
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>io.quarkus</groupId>
+        <artifactId>quarkus-security-jpa-parent</artifactId>
+        <version>999-SNAPSHOT</version>
+        <relativePath>../pom.xml</relativePath>
+    </parent>
+
+    <artifactId>quarkus-security-jpa-deployment</artifactId>
+    <name>Quarkus - Security JPA - Deployment</name>
```
gsmet:
```suggestion
    <name>Quarkus - Elytron Security JPA - Deployment</name>
```
(btw, I agree it's ugly and nobody cares about Elytron but that's how all the others are)

extensions/security-jpa/deployment/src/main/java/io/quarkus/security/jpa/deployment/JpaSecurityDefinition.java
```diff
@@ -0,0 +1,125 @@
...
+                        "annotatedFieldOrMethod must be a field or method: " + annotatedFieldOrMethod);
+        }
+    }
+
+    // FIXME: in order to check for the getter type we need to apply type parameters, that's too complex so assume it matches
+    private static MethodInfo findGetter(Index index, ClassInfo annotatedClass, FieldInfo field, boolean isPanache) {
+        // if it's a panache field, we won't see the getter but it will be there
+        String methodName = "get" + JavaBeanUtil.capitalize(field.name());
+        if (isPanache)
+            return MethodInfo.create(field.declaringClass(), methodName, new Type[0], field.type(), (short) Modifier.PUBLIC);
```
gsmet: Curly braces.

extensions/security-jpa/deployment/src/main/java/io/quarkus/security/jpa/deployment/JpaSecurityDefinition.java
```diff
@@ -0,0 +1,125 @@
...
+            }
+        }
+        for (DotName interfaceName : annotatedClass.interfaceNames()) {
+            ClassInfo interf = index.getClassByName(interfaceName);
+            if (interf != null) {
+                method = findGetter(index, interf, methodName);
+                if (method != null)
+                    return method;
+            }
+        }
```
gsmet: Curly braces in all this method.

extensions/security-jpa/deployment/src/main/java/io/quarkus/security/jpa/deployment/QuarkusSecurityJpaProcessor.java
```diff
@@ -0,0 +1,339 @@
...
+        return new CapabilityBuildItem(Capabilities.SECURITY_JPA);
+    }
+
+    @BuildStep
+    void configureJpaAuthConfig(ApplicationIndexBuildItem index,
+            BuildProducer<UnremovableBeanBuildItem> unremovable,
+            BuildProducer<GeneratedBeanBuildItem> beanProducer,
+            List<PanacheEntityClassesBuildItem> panacheEntityClasses) throws Exception {
+
+        // Generate an IdentityProvider if we have a @UserDefinition
```
gsmet: Same in this class about the curly braces. Please configure your IDE once and for all as it's a recurring issue in your PRs.

extensions/security-jpa/runtime/pom.xml
```diff
@@ -0,0 +1,68 @@
...
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>io.quarkus</groupId>
+        <artifactId>quarkus-security-jpa-parent</artifactId>
+        <version>999-SNAPSHOT</version>
+        <relativePath>../pom.xml</relativePath>
+    </parent>
+
+    <artifactId>quarkus-security-jpa</artifactId>
+    <name>Quarkus - Security JPA - Runtime</name>
```
gsmet:
```suggestion
    <name>Quarkus - Elytron Security JPA - Runtime</name>
```

extensions/security-jpa/runtime/src/main/java/io/quarkus/security/jpa/runtime/Password.java
```diff
@@ -0,0 +1,26 @@
...
+import java.lang.annotation.Target;
+
+/**
+ * <p>
+ * Indicates that this field or property should be used as a source of password for security. Only
+ * supports the {@link String} type.
+ * </p>
+ * <p>
+ * Defaults to considering the password as hashed with bcrypt in the Modular Crypt Format.
+ * </p>
```
gsmet: For the record, Javadoc complains about the closing `</p>`.

extensions/security-jpa/runtime/src/main/resources/META-INF/quarkus-extension.yaml
```diff
@@ -0,0 +1,12 @@
+---
+name: "Quarkus Security JPA Support"
```
gsmet:
```suggestion
name: "Elytron Security JPA"
```

gsmet (dismissed):
Some comments are purely cosmetic but some are really important and needs fixing.

bom/deployment/pom.xml
stuartwdouglas: So this is a bit of a weird one, this is not really an Elytron extension even though it does use some of the Elytron password functionality, it does not use the Elytron realm mechanism.

FroMage: It's only using Elytron under the hood for bCrypt, which is an implementation detail.

docs/src/main/asciidoc/security-jpa.adoc
FroMage: Agree they should be in sync, this one's a mistake.

docs/src/main/asciidoc/security-jpa.adoc
FroMage: Oh, I didn't know that. Will move.

bom/deployment/pom.xml
gsmet: @stuartwdouglas apparently you were the one asking for a rename. I'm just a bit concerned that all the other security modules have Elytron in the name and I'm not really sure they depend more on Elytron than this one but maybe I'm mistaken? Not being consistent is an issue IMHO. I wasn't really fond of having Elytron in the name to begin with but I'm not sure it's a good idea to not be consistent.

docs/src/main/asciidoc/security-jpa.adoc
FroMage: OK

docs/src/main/asciidoc/security-jpa.adoc
gsmet: You have another occurrence in the doc then IIRC

extensions/elytron-security-common/runtime/src/main/java/io/quarkus/elytron/security/common/runtime/BcryptUtil.java
FroMage: @dmlloyd I'm not sure. Can it? It probably will depend on the security provider being registered in the static runtime init before this one, but I'm not sure we can easily guarantee order, right?

extensions/elytron-security-common/runtime/src/main/java/io/quarkus/elytron/security/common/runtime/BcryptUtil.java
FroMage: If this is required, why doesn't the formatter do that, then?

extensions/security-jpa/deployment/src/main/java/io/quarkus/security/jpa/deployment/QuarkusSecurityJpaProcessor.java
FroMage: If it's a requirement the formatter should do it.

extensions/security-jpa/runtime/src/main/java/io/quarkus/security/jpa/runtime/Password.java
FroMage: I can't imagine why.

dmlloyd: I guess because the end tag is optional.

FroMage: What nonsense.

dmlloyd: It doesn't help matters that IntelliJ aggressively creates end tags...

FroMage:
I've put changes for every comment @gsmet made except the elytron/not debate. I'll let @gsmet and @stuartwdouglas decide here and I'll made any eventual change required.

geoand:
At last CI is not throwing random errors at us!

FroMage:
FINALLY :) Now we're "just" waiting for @gsmet and @stuartwdouglas to agree on an extension name.

bom/deployment/pom.xml
stuartwdouglas: OIDC does not have Elytron in the name, as it does not use Elytron. Elytron is not the core of our security infrastructure, we have some modules that use it because we did not have the time or resources to implement some of that stuff ourselves, but Elytron is inherently blocking.

FroMage:
@gsmet and @stuartwdouglas any conclusion?

stuartwdouglas:
I don't think it should have Elytron in the name, as it does not provide an Elytron realm.

FroMage:
In that case we're all good?

extensions/security-jpa/runtime/src/main/resources/META-INF/quarkus-extension.yaml
```diff
@@ -0,0 +1,12 @@
+---
+name: "Quarkus Security JPA"
```
gsmet:
```suggestion
name: "Security JPA"
```
Given what I see on `code.quarkus.io`, it shouldn't have Quarkus in the name.

gsmet (changes requested):
If you can just adjust the extension name and squash everything, let's merge it.

FroMage:
Rebased and squashed.

gsmet (approved):

gsmet:
Merged, thanks for your patience!

FroMage:
Great, thanks!

PR #5670, opened 2019-11-21
Add more GizmoAdaptor stuff

geoand:
Follow up of: #5646


gsmet:
BTW, it might be a stupid question but can we use Gizmo when transforming an existing class?

For an HV POC, I ended up using ASM directly as that's what others did but it would be nice to be able to use Gizmo for that too.

geoand:
From what I have seen from Gizmo, that isn't supported.
@stuartwdouglas would of course be the authority

gsmet (approved):

gsmet:
You asked for @mkouba 's review so I let him have the final say.

geoand:
Only because @mkouba had reviewed the previous PR, no other special reason

mkouba (approved):

PR #5692, opened 2019-11-22
Extension to view MicroProfile config values in dev mode

hpehl:
PR for #1349.

Adds a servlet in dev mode which lists the configured values for MicroProfile Configuration. The servlet is registered using the path /config by default and lists all properties of all config sources. The config sources are sorted descending by ordinal, the properties by name. If no config is defined an empty JSON object is returned.

There's no integration test in this PR. Since the servlet is only added in dev mode, @QuarkusTest does not work as expected.

If necessary I can provide documentation.


extensions/pom.xml
```diff
@@ -87,7 +88,6 @@
         <module>spring-di</module>
         <module>spring-web</module>
         <module>spring-data-jpa</module>
-        <module>spring-security</module>
```
kenfinnigan: are these modules meant to be removed?

extensions/config-viewer/runtime/src/main/java/io/quarkus/config/ConfigViewerServlet.java
```diff
@@ -0,0 +1,94 @@
...
+ *       "properties": {
+ *         "greeting.message": "hello",
+ *         "greeting.name": "quarkus"
+ *       }
+ *     }
+ *   ]
+ * }
+ * </pre>
+ */
+@WebServlet
```
kenfinnigan: Ideally I think this should be a pure Vert.x Handler like Health and Metrics: https://github.com/quarkusio/quarkus/blob/master/extensions/smallrye-health/runtime/src/main/java/io/quarkus/smallrye/health/runtime/SmallRyeHealthHandler.java

And thus change the dependency from undertow to vertx. That's going to be the lowest common denominator

kenfinnigan (changes requested):
I'm not sure if we need some extension metadata to indicate that this is dev only, if that's possible?

extensions/config-viewer/runtime/src/main/java/io/quarkus/config/ConfigViewerServlet.java
hpehl: I'm using a pure Vert.x handler now.

extensions/pom.xml
hpehl: Should be fixed now.

extensions/config-viewer/runtime/src/main/java/io/quarkus/config/ConfigHolder.java
```diff
@@ -0,0 +1,11 @@
+package io.quarkus.config;
+
+import javax.inject.Inject;
+
+import org.eclipse.microprofile.config.Config;
+
+public class ConfigHolder {
```
geoand: This shouldn't be needed IIUC, you could just pull `Config` out of CDI.current() instead of wrapping in this class

geoand:
Looks nice!

I am wondering if this could be part of the standard dev mode experience without the user having to add another extension to have access to the view.

extensions/config-viewer/runtime/src/main/java/io/quarkus/config/ConfigHolder.java
hpehl: Ok, I'll try to simplify that.

hpehl: Removed `ConfigHolder`.

geoand:
@hpehl did you perhaps look into having this available in dev mode by default without requiring the user to add an extra dependency?

hpehl:
Sounds reasonable. 
How would that work? Still as an extension or do I need to integrate it in a different way?

geoand:
I was think that if the code could live in the `io.quarkus:quarkus-development-mode` module that would be great since it wouldn't require users to add anything new

hpehl:
Ok. I'm going to refactor and update the PR.

geoand:
@hpehl thanks!

Best have the refactoring in a separate commit in case someone else comes up with a good reason why this change shouldn't be in the core devmode code.

Thanks

hpehl:
@geoand Adding the code directly in `quarkus-development-mode` would add dependencies to `quarkus-vertx-http` and `quarkus-jsonp`. Not sure if that's what we want?

DevModeMain.java:167
```java 
String configPath = buildSystemProperties.getProperty("quarkus.config.path", "/config");
builder.addChainCustomizer(new Consumer<BuildChainBuilder>() {
    @Override
    public void accept(BuildChainBuilder buildChainBuilder) {
        buildChainBuilder.addBuildStep(new BuildStep() {
            @Override
            public void execute(BuildContext context) {
                // this would add dependencies to quarkus-vertx-http and quarkus-jsonp
                context.produce(new RouteBuildItem(configPath, new ConfigViewerHandler(), HandlerType.BLOCKING));
            }
        });
    }
});
```

Are there other ways to get this into `quarkus-development-mode`?

gsmet:
@geoand should we try to get that one into 1.3? It looks useful.

geoand:
Yeah it would definitely be nice to have.

vertx-http (or perhaps it's internal version) should already be a dependency of the dev-mode, so I don't think that should be a problem.
Adding jsonp I would like to avoid if at all possible...

hpehl:
Thanks for the feedback @gsmet and @geoand. I'll try to update the PR to the latest versions.

geoand:
So should this be closed now?

hpehl:
Superseded by https://github.com/quarkusio/quarkus/pull/7430

Issue #3166, opened 2019-07-10
Support and/or document pushing the logs to EFK (elastic fluentd kibana) and ELK (elastic logstash kibana)


gsmet:
Apparently, we don't have anything built in for that in JBoss Logging but we have a generic JSON handler that could maybe be used.

Maybe it's just documentation needed but we had a question about that on Zulip and we need a proper answer.

/cc @emmanuelbernard to prioritize

emmanuelbernard:
Looks like a good contribution opportunity

loicmathieu:
Hello,

There is two sides on this issue:
1. Formatting the log to EFK or ELK: usually, it's advised to just generate standard logs and making in EFK/ELK the logs mapping to the internal representation. So I don't think there is something specific to be done in Quarkus. Of course it exists some solution for log4j to generate log events in logstrash native format but I don't think it's a good idea.
2. Sending the log to EFK/ELK: again, usually it's advised to just log to a file (or the console in containerized application) and let the log shipper do the job (tail the file or read the console with the docker log driver). Again in log4j it exists a socket appender that can send the log to Logstash directly but it's not a good idea as it will couple together the apps and the log solution and in case the event cannot be send to the socket it's lost

That's why I don't think something specific needs to be done at Quarkus level. Of course JBoss loggin can propose formatter/appender for logstash or fluentd but the current avised way (I discuss this with devops experts) is to log locally (file or console) in a parseable pattern and use the tool to parse and ship.

For me, centralized logging is something, like monitoring, that is not the concern of the application.

gsmet:
Even if we don't end up doing anything (it appears it's not needed), I think it would be interesting to have a blog post explaining how to properly map the fields and so on.

This way, we could also check that we have all the necessary information to do it.

dmlloyd:
Related #453, #2348

bamling:
Any idea when a solution for this will be introduced? We're currently experimenting with Quarkus and would like to end up using it in production, but we require (if possible, easily configured) JSON logging for our ELK stack.

gsmet:
We have preliminary work here: https://github.com/quarkusio/quarkus/pull/2348 .

But it needs to be moved to a separate extension to avoid having the JSON dependency in our core artifacts.

It's typically something a new contributor should be able to do. Would you be interested? I can provide some guidance along the way.

gsmet:
[...]

PR #5694, opened 2019-11-22
New extension: GELF logging support

### LINKED ISSUE #4908: Graylog log management
**Describe the extension**
With distributed systems, centralized logs are required to unify everything. Otherwise, it is impossible to manage the scale.

Graylog is a great open source log management. 

Java client lib is: https://github.com/mp911de/logstash-gelf

Interested in this extension, please +1 via the emoji/reaction feature of GitHub (top right).

**Configuration suggestion**
```properties
level=(INFO, DEBUG...)
protocol=(UDP,TCP)
host=localhost
port=12201
gelfVersion=1.1
extractStackTrace=true
filterStackTrace=true
timestampPattern=yyyy-MM-dd HH:mm:ss
maximumMessageSize=8192(UDP datagrams are usually limited to a size of 8192 bytes)
additionalFields=fieldName1=fieldValue1,fieldName2=fieldValue2
additionalFieldTypes=fieldName1=String,fieldName2=Double,fieldName3=Long


```

**Additional context**
(Add any other context about the proposal here.)

loicmathieu:
Provides support for GELF Log Handler.
Fixes #4908 and #3166 (maybe not entirely as I don't know if it is usable for Fluentd)

This allows to send logs directly to Graylog, Logstash or Fluentd via TCP or UDP socket in the GELF Format.

~~This is still a WIP:~~

- [x] Native is not working (but don't generate an error)
- [x] Logstash needs to be tested but should works
- [x] Needs to be evaluated if it can be used with Fluentd
- [x] Guide needs to be done

The following issues needs to be opened after this PR will be merged to improve it:

- Resoves the issue with the NPE on `java.net.InetAddress.getHostFromNameService()` on native and remove the current workaround
- Fix the issue of the first log event missing in native mode (can be linked to the previous issue)


loicmathieu:
@viniciusfcf this PR allows to send logs to Graylog via TCP or UDP.

viniciusfcf:
Great @loicmathieu, I will test it as soon as possible

loicmathieu:
Tested with ELK, logs arrived correctly inside Logstash.

loicmathieu:
Tested with Graylog, ELK and EFK.
Created a guide with the all three solution + SYSLOG with ELK.
Still needs some native support ...

loicmathieu:
This is ready to review.

@gsmet, I add a new guide that I called "Central log management" (didn't find a better name) and document how to send logs to Graylog, ELK, EKF with this library and also using SYSLOG to send logs to Fluentd. I know you will find a lots of wording issue just feel free to send a PR with them I will not discuss :)

@dmlloyd this is using the new logging integration facility that you create to provides a new handler to send logs to a remote host in GELF format. Can you check the native part ?

gsmet:
Please squash all the commits into one.

loicmathieu:
@gsmet I was planning to squash everything after the review but OK, done :)

extensions/logging-gelf/runtime/src/main/java/io/quarkus/logging/gelf/GelfConfig.java
```diff
@@ -0,0 +1,52 @@
...
+import io.quarkus.runtime.annotations.ConfigPhase;
+import io.quarkus.runtime.annotations.ConfigRoot;
+
+@ConfigRoot(phase = ConfigPhase.RUN_TIME, name = "log.handler.gelf")
+public class GelfConfig {
+    /**
+     * Determine whether to enable the GELF logging handler
+     */
+    @ConfigItem
+    boolean enabled;
```
dmlloyd: This will be `false` by default; was that intentional?

extensions/logging-gelf/runtime/src/main/java/io/quarkus/logging/gelf/GelfConfig.java
```diff
@@ -0,0 +1,52 @@
...
+     */
+    @ConfigItem(defaultValue = "1.1")
+    public String version;
+
+    /**
+     * Whether to post Stack-Trace to StackTrace field (true/false/throwable reference [0 = throwable, 1 =
+     * throwable.cause, -1 = root cause]).
+     */
+    @ConfigItem(defaultValue = "true")
+    public String extractStackTrace;
```
dmlloyd: So... this can be `true` or `false` or one of three integer values?  At that point maybe an `enum` is better.

extensions/logging-gelf/runtime/src/main/java/io/quarkus/logging/gelf/GelfConfig.java
```diff
@@ -0,0 +1,52 @@
...
+
+    /**
+     * Whether to perform Stack-Trace filtering
+     */
+    @ConfigItem(defaultValue = "false")
+    public boolean filterStackTrace;
```
dmlloyd: The default is already `false` and doesn't need to be given.

extensions/logging-gelf/runtime/src/main/java/io/quarkus/logging/gelf/graal/JulLogEventSubstitution.java
```diff
@@ -0,0 +1,39 @@
...
+@TargetClass(className = "biz.paluch.logging.gelf.jul.JulLogEvent")
+public final class JulLogEventSubstitution {
+    @Alias
+    private static Map<String, String> threadNameCache;
+
+    @Substitute
+    private String getThreadName(LogRecord record) {
+        // the default implementation uses the ThreadMXBean to get the name of the thread corresponding to this thead ID
+        // SVM didn't support JMX so we use the Thread.getAllStackTraces().keySet(); instead
+        // an issue is openned upstream: https://github.com/mp911de/logstash-gelf/issues/213
```
dmlloyd: Have you tried this?

```java
if (record instanceof ExtLogRecord) { // from jboss-logmanager
   return ((ExtLogRecord)record).getThreadName();
}
```
I would expect `record` will *always* be an instance of `ExtLogRecord`; you should be able to assume that is always true.

dmlloyd (changes requested):

extensions/logging-gelf/runtime/src/main/java/io/quarkus/logging/gelf/GelfConfig.java
loicmathieu: Yes, I try to use the same construct as in https://github.com/quarkusio/quarkus/blob/master/extensions/logging-json/runtime/src/main/java/io/quarkus/logging/json/runtime/JsonConfig.java#L20 to be able to automatically switch to `true` when a user configure the host but didn't succeed.

Typical use case would be to not use a central log management tool in local/dev and only in production.

extensions/logging-gelf/runtime/src/main/java/io/quarkus/logging/gelf/graal/JulLogEventSubstitution.java
loicmathieu: The substitution's code would certainly be accepted upstream, in this case I would remove it.
If we keep our substitution, I will try what you propose

dmlloyd: It would be nice to use the better code in JVM mode too, if possible, but having it just in the substitution should be OK for now I would think.

extensions/logging-gelf/runtime/src/main/java/io/quarkus/logging/gelf/GelfConfig.java
loicmathieu: In fact it's more complicated than this, I'll update the JavaDoc you can pass any integer and it walk throught the cause tree ...

extensions/logging-gelf/runtime/src/main/java/io/quarkus/logging/gelf/graal/JulLogEventSubstitution.java
loicmathieu: I add a fast path in case the thread that log the message is the same that will handle it. Testing it in native this fast path is used.
I prefere to keep the substitution the same as the implementation that will be done upstream. Is it OK for you ?

extensions/logging-gelf/runtime/src/main/java/io/quarkus/logging/gelf/GelfConfig.java
loicmathieu: done

extensions/logging-gelf/runtime/src/main/java/io/quarkus/logging/gelf/GelfConfig.java
loicmathieu: JavaDoc updated to explain the use of throwable reference

dmlloyd: You've got two values (`true` and `0`) that give the same behavior, and two fundamentally different *kinds* of values.

You might instead do something like this:

```java
    @ConfigItem(defaultValue = "true")
    public boolean extractStackTrace;

    @ConfigItem(defaultValue = "0")
    public int stackTraceSearchIndex;
```

The names are just examples, but this way each property expresses one concept, and the search index can be documented to be only used when stack trace extraction is enabled.

In general we want one property to have one type.

loicmathieu: I agree that it is more elegant but we will no longuer have a one-to-one correspondance between Quarkus and the library.
This can be problematic:
- when the library evolves
- for people already knowing the library
- for people looking at the documentation of the library and trying to do the same with Quarkus inetgration of it

extensions/logging-gelf/runtime/src/main/java/io/quarkus/logging/gelf/graal/JulLogEventSubstitution.java
loicmathieu: @dmlloyd I update the subsitution to the exact changes that will be done upstream. As I said there is a fast path to the current thread and it should be used in most cases.

extensions/logging-gelf/runtime/src/main/java/io/quarkus/logging/gelf/GelfConfig.java
loicmathieu: I implements a config change as you requested in a separate commit to be able to revert it easily. If you prefere it I can squash the commit.

loicmathieu:
@dmlloyd can you review the changes I made ?

There is still an open point that the extension is diabled by default as I would expect users will not use it in dev but only on prod.

I would like to have a way to automatically enabled it if a configuration item exist for it (as it is done here https://github.com/quarkusio/quarkus/blob/master/extensions/logging-json/runtime/src/main/java/io/quarkus/logging/json/runtime/JsonConfig.java#L20) but I test it and it didn't works :( (maybe I misunderstood something).

extensions/logging-gelf/runtime/src/main/java/io/quarkus/logging/gelf/graal/JulLogEventSubstitution.java
dmlloyd: Looking at the GELF code, I think that none of the classes are really appropriate for Quarkus.  It seems like there should be a custom handler that includes proper support for our extended fields.

I see that `JBoss7JulLogEvent` exists but that's pretty old and doesn't account for the extra fields.  I think we want to just use a custom handler with a custom log event type instead of relying on substitutions for this.

loicmathieu: As I said, the substitution is temporary as it is already merged upstream.

Maybe the best is to use the [JBoss7GelfLogHandler](https://github.com/mp911de/logstash-gelf/blob/master/src/main/java/biz/paluch/logging/gelf/jboss7/JBoss7GelfLogHandler.java) that is designed to be used with jboss-logmanager and send a PR to the `logstash-gelf` library if we want to provides some features/fixes like gettting the thread name from the log event.

Regarding extended field, do you mean MDC and additionale filelds? It is supported inside logstahs-gelf but not yet inside this extension.

dmlloyd: No, I mean the thread ID and name.  It's all in the log record already; the rest of this code should not be needed in Quarkus (or in WildFly for that matter).

loicmathieu: So the best is to use the JBoss7GelfLogHandler and send a PR to `logstash-gelf` to override the `getThreadName()` method right ?

So we don't need to maintain our own GelfLogHandler implementation and others that uses `logstash-gelf` will also take advantage of this optimization!

dmlloyd: Yeah something like that.  They also have a WildFly one.  I think JBoss 7 used a version of LogManager from before when we were recording the thread name & ID though, so the WildFly one might be better to extend.  There could be a WildFly record which extends the JBoss7 one with the new stuff.

loicmathieu: The WildFly and the JBoss7 ones seems to be very identical, they both use the ExtLogRegord so if one didn't have access to the thread name the other will not have neither.

Maybe we need to upstream an upgrade of the `jboss-logmanager` library.

I will have a look later, native is not working again I need to understand why :(

dmlloyd: > The WildFly and the JBoss7 ones seems to be very identical, they both use the ExtLogRegord so if one didn't have access to the thread name the other will not have neither.

`ExtLogRecord` got the thread name field in version 1.1.0.  But looking back, the oldest log manager version ever used by JBoss AS 7 was 1.2.0, so I think we don't need to worry about it.  So I will pivot and suggest simply that `JBoss7JulLogEvent` be updated upstream to override `getThreadName` as I suggested earlier.

docs/src/main/asciidoc/central-logging.adoc
```diff
@@ -0,0 +1,377 @@
...
+```
+
+- Configure the gelf log handler to send logs to an external UDP endpoint on the port 12201
+```
+quarkus.log.handler.gelf.enabled=true
+quarkus.log.handler.gelf.host=localhost
+quarkus.log.handler.gelf.port=12201
+```
+
+WARNING: UDP logging is currently not supported on native mode, use TCP if you plan to deploy your application as a native image.
```
dmlloyd: This might not be true anymore now that we're on GraalVM 19.3.0.

docs/src/main/asciidoc/central-logging.adoc
```diff
@@ -0,0 +1,377 @@
...
+    -DclassName="org.acme.quickstart.GelfLoggingResource" \
+    -Dpath="/gelf-logging" \
+    -Dextensions="logging-gelf"
+```
+
+- Add a log inside your rest endpoint
+```
+@Path("/gelf-logging")
+@ApplicationScoped
+public class GelfLoggingResource {
```
dmlloyd: The purpose of this endpoint is just to have something that triggers logging, right?  If so, it should probably be made clear that you don't need this endpoint in order to make this work.

docs/src/main/asciidoc/central-logging.adoc
```diff
@@ -0,0 +1,377 @@
...
+
+```
+curl -H "Content-Type: application/json" -H "Authorization: Basic YWRtaW46YWRtaW4=" -H "X-Requested-By: curl" -X POST -v -d \
+'{"title":"udp input","configuration":{"recv_buffer_size":262144,"bind_address":"0.0.0.0","port":12201,"decompress_size_limit":8388608},"type":"org.graylog2.inputs.gelf.udp.GELFUDPInput","global":true}' \
+http://localhost:9000/api/system/inputs
+```
+
+Launch your application, you should see your logs arriving inside Graylog.
+
+NOTE: due to the initialization time of the GELF log handler, it is possible that not all startup logs are sent to Graylog.
```
dmlloyd: Because of the way we set up our log handlers, the few "missed" log messages should be automatically replayed.  If this isn't working, let's have a chat to figure out the cause of the problem.

docs/src/main/asciidoc/central-logging.adoc
```diff
@@ -0,0 +1,377 @@
...
+
+networks:
+  elk:
+    driver: bridge
+
+```
+
+Launch your application, you should see your logs arriving inside the Elastic Stack; you can use Kibana to access them.
+
+NOTE: due to the initialization time of the GELF log handler, it is possible that not all startup logs are sent to the Elastic Stack.
```
dmlloyd: Here too.

dmlloyd:
The code looks OK at least from my read-through; if the `enabled` switch is `true` then it should just work.  What are the symptoms?  The handler is not being installed?

docs/src/main/asciidoc/central-logging.adoc
loicmathieu: Yes, it's the case. I wasn't sure so I write this note. Will delete it.

docs/src/main/asciidoc/central-logging.adoc
loicmathieu: On GraalVM 19.3 it didn't works anymore!

```
	at io.quarkus.runner.ApplicationImpl.doStart(ApplicationImpl.zig:191)
	at io.quarkus.runtime.Application.start(Application.java:87)
	at io.quarkus.runtime.Application.run(Application.java:210)
	at java.net.InetAddress.getHostFromNameService(InetAddress.java:615)
	at java.net.InetAddress.getCanonicalHostName(InetAddress.java:589)
	at biz.paluch.logging.RuntimeContainer.isQualified(RuntimeContainer.java:148)
	at biz.paluch.logging.RuntimeContainer.getInetAddressWithHostname(RuntimeContainer.java:137)
	at biz.paluch.logging.RuntimeContainer.lookupHostname(RuntimeContainer.java:90)
	at biz.paluch.logging.RuntimeContainer.initialize(RuntimeContainer.java:67)
	at biz.paluch.logging.gelf.jul.GelfLogHandler.<init>(GelfLogHandler.java:54)
	at io.quarkus.logging.gelf.GelfLogHandlerRecorder.initializeHandler(GelfLogHandlerRecorder.java:17)
	at io.quarkus.deployment.steps.GelfLogHandlerProcessor$build28.deploy_0(GelfLogHandlerProcessor$build28.zig:76)
	at io.quarkus.deployment.steps.GelfLogHandlerProcessor$build28.deploy(GelfLogHandlerProcessor$build28.zig:36)
	at io.quarkus.runner.ApplicationImpl.doStart(ApplicationImpl.zig:68)
	... 3 more
```

extensions/logging-gelf/runtime/src/main/java/io/quarkus/logging/gelf/graal/JulLogEventSubstitution.java
```diff
@@ -0,0 +1,47 @@
...
+public final class JulLogEventSubstitution {
+    @Alias
+    private static Map<String, String> threadNameCache;
+
+    @Substitute
+    private String getThreadName(LogRecord record) {
+        // the default implementation uses the ThreadMXBean to get the name of the thread corresponding to this thead ID
+        // SVM didn't support JMX so we use Thread.getAllStackTraces().keySet() instead
+        // The following PR fixes the issue: https://github.com/mp911de/logstash-gelf/issues/213
+        // This substitution implements the same while waiting for a new release of logstash-gelf library
```
dmlloyd: Again, this whole thing should be replaced with the cast to `ExtLogRecord`.  Even if upstream isn't there yet.  It will be far lest costly in this case.

docs/src/main/asciidoc/central-logging.adoc
dmlloyd: Interesting; any idea where the null reference is coming from?

extensions/logging-gelf/runtime/src/main/java/io/quarkus/logging/gelf/graal/JulLogEventSubstitution.java
loicmathieu: done

docs/src/main/asciidoc/central-logging.adoc
loicmathieu: I use a workaround to configure the hostname resolver to localhost before network and it fixes the issue. 
I propose to open a followup issue to fix this properly after merging this, I suspect and issue on GraalVM side as the NPE is from code inside the JDK : `java.net.InetAddress.getHostFromNameService()`

docs/src/main/asciidoc/central-logging.adoc
loicmathieu: I add a comment

docs/src/main/asciidoc/central-logging.adoc
loicmathieu: deleted

extensions/logging-gelf/runtime/src/main/java/io/quarkus/logging/gelf/graal/JulLogEventSubstitution.java
loicmathieu: done

loicmathieu:
@dmlloyd I success in making this works with GraalVM 19.3. I use a workaround (there is a FIXME in the code for this, I will open a followup issue after this PR will be merged).

Native works also in UDP now :tada: 

The first log event is missing in native, maybe it's related to the first issue (host name resolver). Anyway I propose to open two followup issues after merging this one (I list the issue on the description of the PR).

I create a PR upstream to have the fastpath inside logging-gelf: https://github.com/mp911de/logstash-gelf/pull/217

extensions/logging-gelf/deployment/src/main/java/io/quarkus/logging/gelf/deployment/GelfLogHandlerProcessor.java
```diff
@@ -0,0 +1,58 @@
...
+         * at biz.paluch.logging.gelf.jboss7.JBoss7GelfLogHandler.<init>(JBoss7GelfLogHandler.java:75)
+         * at io.quarkus.logging.gelf.GelfLogHandlerRecorder.initializeHandler(GelfLogHandlerRecorder.java:17)
+         * at io.quarkus.deployment.steps.GelfLogHandlerProcessor$build27.deploy_0(GelfLogHandlerProcessor$build27.zig:76)
+         * at io.quarkus.deployment.steps.GelfLogHandlerProcessor$build27.deploy(GelfLogHandlerProcessor$build27.zig:36)
+         * at io.quarkus.runner.ApplicationImpl.doStart(ApplicationImpl.zig:68)
+         * at io.quarkus.runtime.Application.start(Application.java:87)
+         * at io.quarkus.runtime.Application.run(Application.java:210)
+         * at io.quarkus.runner.GeneratedMain.main(GeneratedMain.zig:41)
+         */
+        return new SystemPropertyBuildItem("logstash-gelf.resolutionOrder", "localhost,network");
```
dmlloyd: This is a comment to look into for later, but: we already include the source host name in `ExtLogRecord`, so maybe there's some way to utilize that information and bypass this whole thing.

dmlloyd (approved):

docs/src/main/asciidoc/central-logging.adoc
```diff
@@ -0,0 +1,369 @@
+////
+This guide is maintained in the main Quarkus repository
+and pull requests should be submitted there:
+https://github.com/quarkusio/quarkus/tree/master/docs/src/main/asciidoc
+////
+= Quarkus - Central log management (Graylog, Logstash, Fluentd)
+
+include::./attributes.adoc[]
+
+This guide explains how you can send you log to a central logging system like Graylog, Logstash (inside the Elastic Stack or ELK - Elasticsearch, Logstash, Kibana) or
```
gsmet:
```suggestion
This guide explains how you can send your logs to a central logging system like Graylog, Logstash (inside the Elastic Stack or ELK - Elasticsearch, Logstash, Kibana) or
```

docs/src/main/asciidoc/central-logging.adoc
```diff
@@ -0,0 +1,369 @@
...
+Fluentd (inside EFK - Elasticsearch, Fluentd, Kibana).
+
+There is a lot of different way to centralize your logs (if you're using Kubernetes, the simplest way is to log to the console and ask you cluster administrator to integrate a central log manager inside your cluster).
```
gsmet:
```suggestion
There are a lot of different ways to centralize your logs (if you are using Kubernetes, the simplest way is to log to the console and ask you cluster administrator to integrate a central log manager inside your cluster).
```

docs/src/main/asciidoc/central-logging.adoc
```diff
@@ -0,0 +1,369 @@
...
+In this guide, we will expose how to send them to an external tool using the `quarkus-logging-gelf` extension that can use TCP or UDP to send logs in the Graylog Extented Log Format (GELF).
```
gsmet:
```suggestion
In this guide, we will expose how to send them to an external tool using the `quarkus-logging-gelf` extension that can use TCP or UDP to send logs in the Graylog Extended Log Format (GELF).
```

docs/src/main/asciidoc/central-logging.adoc
```diff
@@ -0,0 +1,369 @@
...
+
+The `quarkus-logging-gelf` extension will add a GELF log handler to the underlying logging backend that Quarkus uses (jboss-logmanager).
+By default, it is disabled, if you enable it but still use another handler (by default the console handler is enabled), your logs will be sent to both handlers.
+
+== Example application
+
+The following examples will all be based on the same example application that you can create based on the following steps:
```
gsmet:
```suggestion
The following examples will all be based on the same example application that you can create with the following steps:
```

docs/src/main/asciidoc/central-logging.adoc
```diff
@@ -0,0 +1,369 @@
...
+
+- Create an application with the `quarkus-logging-gelf` extension. You can use the following Maven command to create it:
+```
```
gsmet: Please use proper Asciidoc source for all the source extracts.

docs/src/main/asciidoc/central-logging.adoc
```diff
@@ -0,0 +1,369 @@
...
+mvn io.quarkus:quarkus-maven-plugin:1.0.0.Final:create \
+    -DprojectGroupId=org.acme \
+    -DprojectArtifactId=gelf-logging \
+    -DclassName="org.acme.quickstart.GelfLoggingResource" \
+    -Dpath="/gelf-logging" \
+    -Dextensions="logging-gelf"
+```
+
+- For demonstration purpose, we create an endpoint that do nothing but log a sentence. You don't need to do this inside your application.
+```
```
gsmet: Same here about using source + you need to say it's Java.

docs/src/main/asciidoc/central-logging.adoc
```diff
@@ -0,0 +1,369 @@
...
+mvn io.quarkus:quarkus-maven-plugin:1.0.0.Final:create \
+    -DprojectGroupId=org.acme \
+    -DprojectArtifactId=gelf-logging \
+    -DclassName="org.acme.quickstart.GelfLoggingResource" \
+    -Dpath="/gelf-logging" \
+    -Dextensions="logging-gelf"
+```
+
+- For demonstration purpose, we create an endpoint that do nothing but log a sentence. You don't need to do this inside your application.
```
gsmet:
```suggestion
- For demonstration purpose, we create an endpoint that does nothing but log a sentence. You don't need to do this inside your application.
```

docs/src/main/asciidoc/central-logging.adoc
```diff
@@ -0,0 +1,369 @@
...
+```
+@Path("/gelf-logging")
```
gsmet: Please include the imports. in the source file. Otherwise people have a hard time reproducing the steps.

docs/src/main/asciidoc/central-logging.adoc
```diff
@@ -0,0 +1,369 @@
...
+
+    @GET
+    public void log() {
+        LOG.info("Some useful log message");
+    }
+
+}
+```
+
+- Configure the gelf log handler to send logs to an external UDP endpoint on the port 12201
```
gsmet:
```suggestion
- Configure the GELF log handler to send logs to an external UDP endpoint on the port 12201:
```

docs/src/main/asciidoc/central-logging.adoc
```diff
@@ -0,0 +1,369 @@
...
+- Graylog
+
+You can do this via the following docker-compose file that you can launch via `docker-compose run -d`:
+
+```
+version: '3.2'
+
+services:
+  elasticsearch:
+    image: docker.elastic.co/elasticsearch/elasticsearch-oss:6.8.2
```
gsmet: We can't use a more recent version?

docs/src/main/asciidoc/central-logging.adoc
```diff
@@ -0,0 +1,369 @@
...
+curl -H "Content-Type: application/json" -H "Authorization: Basic YWRtaW46YWRtaW4=" -H "X-Requested-By: curl" -X POST -v -d \
+'{"title":"udp input","configuration":{"recv_buffer_size":262144,"bind_address":"0.0.0.0","port":12201,"decompress_size_limit":8388608},"type":"org.graylog2.inputs.gelf.udp.GELFUDPInput","global":true}' \
+http://localhost:9000/api/system/inputs
+```
+
+Launch your application, you should see your logs arriving inside Graylog.
+
+== Send logs to Logstash / the Elastic Stack (ELK)
+
+Logstash comes by default with an Input plugin that can understand the GELF format, we will first create a pipeline that enable this plugin.
```
gsmet:
```suggestion
Logstash comes by default with an Input plugin that can understand the GELF format, we will first create a pipeline that enables this plugin.
```

docs/src/main/asciidoc/central-logging.adoc
```diff
@@ -0,0 +1,369 @@
...
+
+You can do this via the following docker-compose file that you can launch via `docker-compose run -d`:
+
+```
+# Launch Elasticsearch
+version: '3.2'
+
+services:
+  elasticsearch:
+    image: docker.elastic.co/elasticsearch/elasticsearch-oss:6.8.2
```
gsmet: Same here about the Elastic stack version.

docs/src/main/asciidoc/central-logging.adoc
```diff
@@ -0,0 +1,369 @@
...
+    depends_on:
+      - elasticsearch
+
+networks:
+  elk:
+    driver: bridge
+
+```
+
+Launch your application, you should see your logs arriving inside the Elastic Stack; you can use Kibana to access them.
```
gsmet: Maybe add a link to the Kibana UI?

docs/src/main/asciidoc/central-logging.adoc
```diff
@@ -0,0 +1,369 @@
...
+- Fluentd
+- Kibana
+
+You can do this via the following docker-compose file that you can launch via `docker-compose run -d`:
+```
+version: '3.2'
+
+services:
+  elasticsearch:
+    image: docker.elastic.co/elasticsearch/elasticsearch-oss:6.8.2
```
gsmet: Same here about Elastic version.

docs/src/main/asciidoc/central-logging.adoc
```diff
@@ -0,0 +1,369 @@
...
+      - efk
+    depends_on:
+      - elasticsearch
+
+networks:
+  efk:
+    driver: bridge
+```
+
+Launch your application, you should see your logs arriving inside EFK: you can use Kibana to access them.
```
gsmet: Same here about the Kibana URI.

docs/src/main/asciidoc/central-logging.adoc
```diff
@@ -0,0 +1,369 @@
...
+
+== Fluentd alternative: use SYSLOG
+
+You can also send your logs to Fluentd using a SYSLOG input.
```
gsmet: can you do `s/SYSLOG/Syslog/g`?

docs/src/main/asciidoc/central-logging.adoc
```diff
@@ -0,0 +1,369 @@
...
+As opposed to GELF input, SYSLOG input will not render multiline logs in one event, that's why we advise to use the GELF input that we implement in Quarkus.
```
gsmet:
```suggestion
As opposed to the GELF input, the SYSLOG input will not render multiline logs in one event, that's why we advise to use the GELF input that we implement in Quarkus.
```

docs/src/main/asciidoc/central-logging.adoc
```diff
@@ -0,0 +1,369 @@
...
+
+First, you need to create a Fluentd image with the needed plugins: elasticsearch and input-gelf.
```
gsmet: You don't have `input-gelf` in the code below and it doesn't seem needed.

docs/src/main/asciidoc/central-logging.adoc
```diff
@@ -0,0 +1,369 @@
...
+- Fluentd
+- Kibana
+
+You can do this via the following docker-compose file that you can launch via `docker-compose run -d`:
+```
+version: '3.2'
+
+services:
+  elasticsearch:
+    image: docker.elastic.co/elasticsearch/elasticsearch-oss:6.8.2
```
gsmet: Same here about Elastic version.

Maybe it's worth making it a variable of the Asciidoc file?

docs/src/main/asciidoc/central-logging.adoc
```diff
@@ -0,0 +1,369 @@
...
+Finally, configure your application to send logs to EFK using SYSLOG:
+```
+quarkus.log.syslog.enable=true
+quarkus.log.syslog.endpoint=localhost:5140
+quarkus.log.syslog.protocol=udp
+quarkus.log.syslog.app-name=quarkus
+quarkus.log.syslog.hostname=quarkus-test
+```
+
+Launch your application, you should see your logs arriving inside EFK: you can use Kibana to access them.
```
gsmet: Same here about the link to the Kibana UI.

docs/src/main/asciidoc/central-logging.adoc
```diff
@@ -0,0 +1,369 @@
...
+
+
+[[configuration-reference]]
+== Configuration Reference
+
+Configuration is done through the normal `application.properties` file.
```
gsmet:
```suggestion
Configuration is done through the usual `application.properties` file.
```

docs/src/main/asciidoc/central-logging.adoc
```diff
@@ -0,0 +1,369 @@
...
+
+include::{generated-dir}/config/quarkus-logging-gelf.adoc[opts=optional, leveloffset=+1]
+
+This extension uses the `logstash-gelf` library that allow more configuration options via system properties,
+you can access it's documentation here: https://logging.paluch.biz/
```
gsmet:
```suggestion
you can access its documentation here: https://logging.paluch.biz/
```

gsmet (changes requested):
I haven't reviewed the code but the documentation requires some work.

docs/src/main/asciidoc/central-logging.adoc
loicmathieu: This is the version used inside the Graylog documentation: http://docs.graylog.org/en/3.1/pages/installation/docker.html#here
A more recent version should works but I will need to re-run all the test ...

docs/src/main/asciidoc/central-logging.adoc
loicmathieu: done

docs/src/main/asciidoc/central-logging.adoc
loicmathieu: Done, sorry for this, markdown habits :)

docs/src/main/asciidoc/central-logging.adoc
loicmathieu: done

docs/src/main/asciidoc/central-logging.adoc
loicmathieu: done

docs/src/main/asciidoc/central-logging.adoc
loicmathieu: done

docs/src/main/asciidoc/central-logging.adoc
loicmathieu: done

docs/src/main/asciidoc/central-logging.adoc
loicmathieu: done

docs/src/main/asciidoc/central-logging.adoc
loicmathieu: done

extensions/logging-gelf/deployment/src/main/java/io/quarkus/logging/gelf/deployment/GelfLogHandlerProcessor.java
loicmathieu: I check the code and it's not possible as it is resolved once at startup and in the part of the code that creates the GELF message totaly de-couple from the log handler part.
I will try to make a small reproducer to see if it's not a bug in GraalVM

docs/src/main/asciidoc/central-logging.adoc
```diff
@@ -97,18 +101,19 @@ services:
...
-```
+----
 
 Then, you need to create a UDP input in Graylog.
 You can do it from the Graylog web console (System -> Input -> Select GELF UDP) or via the API.
 
 This curl example will create a new Input of type GELF UDP, it uses the default login from Graylog (admin/admin).
 
-```
+[source,shell script]
```
gsmet: I think it should be just `shell`.

loicmathieu: This is what IDEA proposes.
I try to generate the doc with both and this does not seems to be any difference so I replace by script.

gsmet:
@loicmathieu can you ping me when you're done? I'll do a doc generation and a final review of the doc

docs/src/main/asciidoc/central-logging.adoc
loicmathieu: In fact Graylog is only compatible with Elasticsearch 5.6 and 6.8: https://docs.graylog.org/en/3.1/pages/configuration/elasticsearch.html

The other examples should work on more recent version of Elasticsearch but to keep it simple I prefere to use the same for all

loicmathieu:
@gsmet I think I incorporate all your concern about the documentation. I generate it and it seems OK. You can do your general review

dmlloyd (approved):

gsmet (approved):

PR #5710, opened 2019-11-22
Refactoring of project-related commands

### LINKED ISSUE #5695: Tools: common Quarkus project command API
Currently project-related commands such as create, add and list extensions are getting hold of the platform descriptor the project is based on using a global instance cached in a static variable, which was a temporary workaround to avoid major refactoring of commands before 1.0 release.

The idea behind this issue is to introduce a basic command API and a way to provide arguments and other contextual data using an invocation object.

aloubyansky:
Currently project-related commands such as create, add and list extensions are getting hold of the platform descriptor the project is based on using a global instance cached in a static variable, which was a temporary workaround to avoid major refactoring of commands before 1.0 release.

The idea behind this change is to introduce a basic command API and a way to provide arguments and other contextual data using an invocation object.

Fixes #5695


stalep (approved):
I think the code changes looks good. A unified way of handling all the commands is very welcome and should make it a lot easier to keep the functionality of the different tools in sync.
I've tested the changes locally using the gradle plugin to build (jar/native), devmode, list/addExtensions it all works (a simple rest example generated from code.quarkus.io).

PR #5713, opened 2019-11-22
Kogito quickstart Person example fails with list of objects

spolti:
…jects

Signed-off-by: Filippe Spolti <fspolti@redhat.com>


spolti:
@mswiderski f.y.i.

mswiderski (approved):

mswiderski:
@spolti thanks for the fix!

Just a comment, this change itself does not need new kogito version as it fixes the immediate bug in handling list of simple types

mswiderski:
What I meant is that it can be safely merged and no need to wait for 0.6.0 of kogito

gsmet (approved):

gsmet:
@mswiderski I'm merging it then.

PR #5721, opened 2019-11-24
Disable the unnecessary JVM debug launch of integration tests

jaikiran:
`quarkus:dev` by default launches the JVM in debug mode. As a result, our integration tests which launch this mojo end up launching the JVM unnecessarily in debug mode with port opening on `5005`. This can be avoided (especially for CI runs).

The commit here disables this by default `debug` launch of these tests, with a change in the `test-framework`. Unless the test intentionally passes a debug attribute (either the `-Ddebug` or `-Dsuspend`), the test framework will disable the JVM debug launch. This should avoid the unnecessary opening and closing of the debug port as well as, in theory, probably speed it up a bit (although I don't know how noticeable that's going to be).


geoand:
Nice catch!

geoand (approved):

machi1990 (approved):
Good catch.

gsmet:
Let's merge it, we can adjust later if needed.

Issue #5683, opened 2019-11-22
quarkus-maven-plugin libDir configuration entry is being ignored


Arcnor:
**Describe the bug**
When configuring the `quarkus-maven-plugin` with the `libDir` setting, nothing really happens. Taking a quick look at the code, the variable is being set but nothing seems to be happening with it at all.

gsmet:
It was removed.

Arcnor:
As you can imagine, I was using this functionality, so can you tell me what's the alternative to what the `libDir` configuration parameter was doing?

gsmet:
Hmmm, but it wasn't wired at all? How did you make use of it?

And what were you trying to accomplish?

Arcnor:
This was working on a previous version. But I just checked the linked issue and it looks like the `lib` directory will go wherever the final build directory is, which I guess it's wherever the `finalName` config setting points to?

That will work for my usecase.

Arcnor:
Unfortunately it doesn't seem to work as I expected, `finalName` will move the generated `jar` to wherever I point, but the `lib` directory stays at `target`.

Is there anything I can do to move the `lib` directory to the right place, besides doing it myself with maven trickery?

gsmet:
IMHO, it makes sense to have the `lib` directory together with the `-runner.jar`. @Arcnor when you define a `finalName`, the `-runner.jar` is generated in the appropriate directory?

Arcnor:
Yes, the `-runner` is in the `finalName` directory, but `lib` doesn't move with it.

gsmet:
@jaikiran could you see if it's easily doable and if it makes sense?

What is a bit annoying is that our generated Dockerfiles won't be accurate for this kind of setup.

But I'm more and more thinking that we should generate them dynamically during the build.

jaikiran:
I'll take a more detailed look.

diogocarleto:
@Arcnor you can use some like that as a workaround:

> <plugin>
        <artifactId>maven-resources-plugin</artifactId>
        <version>3.1.0</version>
        <executions>
          <execution>
            <id>copy-resources</id>
            <phase>verify</phase>
            <goals>
              <goal>copy-resources</goal>
            </goals>
            <configuration>
              <outputDirectory>${project.build.directory}/appengine-staging/lib</outputDirectory>
              <resources>
                <resource>
                  <directory>${project.build.directory}/lib</directory>
                  <filtering>true</filtering>
                </resource>
              </resources>
[...]

PR #5723, opened 2019-11-24
Remove unused "libDir" param

jaikiran:
Relates to https://github.com/quarkusio/quarkus/issues/5683

The `BuildMojo` no longer uses the `libDir` param. The Mojos now all seem to use the (configured) build directory as the root for any sub-directories it wants to resolve, like `classes`, `lib` etc...

The commit here removes this unused param. 

Note that although this was unusued internally, it was still advertized as a valid parameter configurable for the `quarkus-maven-plugin`. So I'm marking this as a backward incompatible change.

Also adding a backport tag to see if we want this included in `1.0.0` to reduce the chances of this param being used by users.


gsmet (approved):

gsmet:
Merged, thanks!

PR #5726, opened 2019-11-24
Document property indirection for passwords coming from Vault - Fixes #3279

### LINKED ISSUE #3279: Use encrypted password for quarkus.mailer.password
**Description**
As of now, the password value for quarkus.mailer.password has to be written in the application.properties file as plain text. 
It would be interesting and more secure to be able to have the value encrypted in the properties file.

vsevel:
This provides documentation (and a test) on the ability to store a password in vault, and refer to it anywhere in the configuration using property indirection.

This was made possible thanks to https://github.com/quarkusio/quarkus/issues/3937.

This solves for instance https://github.com/quarkusio/quarkus/issues/3279, assuming Vault is an option in the user's environment.

cc @sberyozkin

### DIFF
--- a/docs/src/main/asciidoc/mailer.adoc
+++ b/docs/src/main/asciidoc/mailer.adoc
@@ -62,6 +62,24 @@ quarkus.mailer.username=....
 quarkus.mailer.password=....
 ----
 
+[NOTE]
+====
+It is recommended to encrypt any sensitive data, such as the `quarkus.mailer.password`.
+One approach is to save the value into a secure store like HashiCorp Vault, and refer to it from the configuration.
+Assuming for instance that Vault contains key `mail-password` at path `myapps/myapp/myconfig`, then the mailer
+extension can be simply configured as:
+```
+...
+# path within the kv secret engine where is located the application sensitive configuration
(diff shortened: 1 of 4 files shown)


sberyozkin (approved):
@vsevel IMHO it is top class, the extensions can continue using their own properties to work with Vault :-)

vsevel:
@sberyozkin thanks for merging; you were too fast for me ;) I just finished running a full example on mail+vault quickstart using the property indirection.

here is the config I used:
```
quarkus.vault.secret-config-kv-path=myapps/vault-quickstart/config
        
quarkus.mailer.from=test@quarkus.io
quarkus.mailer.host=localhost
quarkus.mailer.port=1025
quarkus.mailer.ssl=false
quarkus.mailer.username=foo
quarkus.mailer.password=${a-private-key}
```

and for the smtp server:
```
docker run -p 1080:1080 -p 1025:1025 djfarrelly/maildev --incoming-user foo --incoming-pass 123456 -v
````

> the extensions can continue using their own properties to work with Vault :-)

indeed. it is pretty nice.
(except for the lack of property value refresh)

Issue #5728, opened 2019-11-24
Proposal: vertx.http.security FORM auth: Make redirect to fixed landing page after login optional


Karm:
I am beginning to learn [Vue.js](https://vuejs.org/) web app programming with [AXIOS](https://vuejs.org/v2/cookbook/using-axios-to-consume-apis.html) library to talk to my backend.

I want to programmatically do a form auth POST request (j_security_check, ``` `j_username=`+user+`&j_password=`+password``` etc.) from my web app running in user's browser to my Quarkus backend, receive a ```quarkus-credential``` cookie and go on with calling REST endpoints with that cookie.

it works fine, but since there is no page to go back to - because I directly call ```/j_security_check``` to do the login, Quarkus redirects me to ```quarkus.http.auth.form.landing-page``` just after login.  

It creates a problem to me: If the Quarkus was the actual server serving the Vue web app, the redirect after successful login would be merely superfluous, but when I run Vue app in live dev mode via [Express / Node](https://expressjs.com/) e.g. on http://localhost:8080 and Quarkus backend also in live :dev mode e.g. on http://localhost:8098, the forced redirect to e.g. http://localhost:8098 or http://localhost:8098/whateverlanding manifests as a routing error I would have to somehow workaround in Axios. Note that cors is setup for me just fine in the Express dev mode.

Being able to edit both the frontend Vue web app code and the backend Quarkus code live without any interruptions is so powerful experience I definitely don't want to give it up.

### Code
e.g. backed-api.js:
```js
import axios from 'axios'
const AXIOS = axios.create({
    baseURL: `/`,
});
export default {
    getSecured(user, password) {
        console.log("Entering get secured...");
        return AXIOS.post(`/api/j_security_check`, `j_username=`+user+`&j_password=`+password,
            withCredentials: true,
            redirect: "manual",   // does not help in browser...
            maxRedirects: 0
       });
    }
}
```

vue.config.js:

```js
module.exports = {
  devServer: {
    proxy: {
      '^/api': {
        target: 'http://localhost:8098',
        ws: true,
        changeOrigin: true
      }
    }
  },
  outputDir: 'target/dist',
  assetsDir: 'static'
};
```
### Browser debug console on login
What is happening by default with  ```quarkus.http.auth.form.landing-page=/```:
![bad](https://user-images.githubusercontent.com/691097/69494645-6dfb9200-0ebe-11ea-9618-8a7ee8167480.png)
 
# Proposal
[...]

PR #5729, opened 2019-11-24
Implements proposal 5728, adds quarkus.http.auth.form.redirect-after-login

Karm:
If there is no page to redirect back to, Quarkus redirects to ```quarkus.http.auth.form.landing-page```
after a successful login. This PR makes this behaviour optional, i.e. with ```quarkus.http.auth.form.redirect-after-login=false```
no such redirect takes place if there is no stored location to go back to, i.e.  ```quarkus.http.auth.form.landing-page``` remains unused.
See https://github.com/quarkusio/quarkus/issues/5728 for details on the proposal.

### DIFF
--- a/extensions/vertx-http/deployment/src/test/java/io/quarkus/vertx/http/security/FormAuthNoRedirectTestCase.java
+++ b/extensions/vertx-http/deployment/src/test/java/io/quarkus/vertx/http/security/FormAuthNoRedirectTestCase.java
@@ -1,0 +1,139 @@
+package io.quarkus.vertx.http.security;
+
+import static org.hamcrest.Matchers.*;
+
+import java.util.function.Supplier;
+
+import org.jboss.shrinkwrap.api.ShrinkWrap;
+import org.jboss.shrinkwrap.api.asset.StringAsset;
+import org.jboss.shrinkwrap.api.spec.JavaArchive;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.RegisterExtension;
+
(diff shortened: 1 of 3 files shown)


sberyozkin:
Hi @Karm It looks good; what do you think about this alternative: do not redirect if neither saved page nor the landing page (resource) exists ? Ideally the landing page property were optional but not sure we can change it now.

stuartwdouglas (approved):

Karm:
Thank you guys for the review @sberyozkin @stuartwdouglas 

 > Hi @Karm It looks good; what do you think about this alternative: do not redirect if neither
 > saved page nor the landing page (resource) exists ? Ideally the landing page property
 > were optional but not sure we can change it now.

I figured I could either 
1) make ```quarkus.http.auth.form.landing-page``` optional (which it is not now) and base the do-not-redirect-to-landing-page logic on that or
2) add another property that controls it, e.g. ```quarkus.http.auth.form.redirect-after-login```

I went with 2), because I thought you wouldn't approve of changing the current default behavior of an already established property. In fact, I would much rather do 1), because it does not introduce yet another property and it is IMHO safe to make a currently mandatory property optional. The other way around would be problematic, but this is not the case.

People who use FORM auth have to have the property set now so if we make it optional, nothing breaks for them...

sberyozkin:
@Karm Thanks for the clarification, yes, I was thinking that may be you could avoid introducing a new property and do not redirect to the landing page if a class loader resource check reveals that landing resource does not exist... Sorry for a delay in responding.
I don't want to delay the PR, it does look good and a new property is a way for the users to tell to the runtime what to do as opposed to it guessing :-)
Thanks

sberyozkin (approved):

PR #5730, opened 2019-11-24
Flyway healthcheck

viniciusfcf:
Validate pending migrations, available locally, but not yet applied to the DB.

### DIFF
--- a/docs/src/main/asciidoc/flyway.adoc
+++ b/docs/src/main/asciidoc/flyway.adoc
@@ -194,3 +194,12 @@ public class MigrationService {
 <1> Inject the Flyway object if you want to use it directly
 <2> Use the Flyway instance directly
 
+
+== Flyway database migrations Health Check
+
+If you are using the `quarkus-smallrye-health` extension, `quarkus-flyway` will automatically add a readiness health check
+to validate the datasource.
+
+So when you access the `/health/ready` endpoint of your application you will have information about the Flyway migration validation status.
+
+This behavior can be disabled by setting `quarkus.flyway.health.enabled`.
--- a/extensions/flyway/runtime/src/main/java/io/quarkus/flyway/runtime/FlywayBuildConfig.java
+++ b/extensions/flyway/runtime/src/main/java/io/quarkus/flyway/runtime/FlywayBuildConfig.java
@@ -18,4 +18,10 @@ public final class FlywayBuildConfig {
      */
     @ConfigItem
     public Optional<List<String>> locations;
+
+    /**
+     * Whether or not a health check is published in case the smallrye-health extension is present.
+     */
+    @ConfigItem(name = "health.enabled", defaultValue = "true")
+    public boolean healthEnabled;
 }
--- a/extensions/flyway/runtime/src/main/java/io/quarkus/flyway/runtime/health/DatabaseMigrationHealthCheck.java
+++ b/extensions/flyway/runtime/src/main/java/io/quarkus/flyway/runtime/health/DatabaseMigrationHealthCheck.java
@@ -1,0 +1,36 @@
+package io.quarkus.flyway.runtime.health;
+
+import javax.annotation.PostConstruct;
+import javax.enterprise.context.ApplicationScoped;
+
+import org.eclipse.microprofile.health.HealthCheck;
+import org.eclipse.microprofile.health.HealthCheckResponse;
+import org.eclipse.microprofile.health.HealthCheckResponseBuilder;
+import org.eclipse.microprofile.health.Readiness;
+import org.flywaydb.core.Flyway;
+
+import io.quarkus.arc.Arc;
+
+@Readiness
+@ApplicationScoped
+public class DatabaseMigrationHealthCheck implements HealthCheck {
+
+    private Flyway flyway;
+
+    @PostConstruct
+    protected void init() {
+        flyway = Arc.container().instance(Flyway.class).get();
+    }
+
+    @Override
+    public HealthCheckResponse call() {
+        HealthCheckResponseBuilder builder = HealthCheckResponse.named("Flyway database migrations health check").up();
+        int pendingLength = flyway.info().pending().length;
+        if (pendingLength != 0) {
+            String data = "Validation check failed with " + pendingLength + " pending migrations";
+            String dsName = "quarkus-flyway";
+            builder.down().withData(dsName, data);
+        }
+        return builder.build();
+    }
+}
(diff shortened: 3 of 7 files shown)


docs/src/main/asciidoc/flyway.adoc
```diff
@@ -194,3 +194,12 @@ public class MigrationService {
 <1> Inject the Flyway object if you want to use it directly
 <2> Use the Flyway instance directly
 
+
+== Flyway migration Health Check
+
+If you are using the `quarkus-smallrye-health` extension, `quarkus-flyway` will automatically add a readiness health check
+to validate the datasource.
+
+So when you access the `/health/ready` endpoint of your application you will have information about the flyway migration validation status.
```
gsmet:
```suggestion
So when you access the `/health/ready` endpoint of your application you will have information about the Flyway migration validation status.
```

docs/src/main/asciidoc/flyway.adoc
```diff
@@ -194,3 +194,12 @@ public class MigrationService {
...
+
+This behavior can be disabled via the property `quarkus.flyway.health.enabled`.
```
gsmet:
```suggestion
This behavior can be disabled by setting `quarkus.flyway.health.enabled` to `false`.
```

extensions/flyway/runtime/src/main/java/io/quarkus/flyway/runtime/FlywayBuildConfig.java
```diff
@@ -18,4 +18,10 @@
      */
     @ConfigItem
     public Optional<List<String>> locations;
+
+    /**
+     * Whether or not an healtcheck is published in case the smallrye-health extension is present (default to true).
```
gsmet:
```suggestion
     * Whether or not a health check is published in case the smallrye-health extension is present.
```
The default will be automatically included in the generated documentation.

extensions/flyway/runtime/src/main/java/io/quarkus/flyway/runtime/health/DatabaseMigrationHealthCheck.java
```diff
@@ -0,0 +1,36 @@
...
+    private Flyway flyway;
+
+    @PostConstruct
+    protected void init() {
+        flyway = Arc.container().instance(Flyway.class).get();
+    }
+
+    @Override
+    public HealthCheckResponse call() {
+        HealthCheckResponseBuilder builder = HealthCheckResponse.named("Database migration(s) health check").up();
```
gsmet:
```suggestion
        HealthCheckResponseBuilder builder = HealthCheckResponse.named("Flyway database migrations health check").up();
```

extensions/flyway/runtime/src/main/java/io/quarkus/flyway/runtime/health/DatabaseMigrationHealthCheck.java
```diff
@@ -0,0 +1,36 @@
...
+        int pendingLength = flyway.info().pending().length;
+        if (pendingLength != 0) {
+            String data = "validation check failed with " + pendingLength + " pending migrations";
```
gsmet:
```suggestion
            String data = "Validation check failed with " + pendingLength + " pending migrations";
```

integration-tests/main/src/test/java/io/quarkus/it/main/HealthTestCase.java
```diff
@@ -27,8 +27,9 @@ public void testHealthCheck() {
                     .contentType(ContentType.JSON)
                     .header("Content-Type", containsString("charset=UTF-8"))
                     .body("status", is("UP"),
-                            "checks.status", containsInAnyOrder("UP"),
-                            "checks.name", containsInAnyOrder("Database connection(s) health check"));
+                            "checks.status", containsInAnyOrder("UP", "UP"),
+                            "checks.name",
+                            containsInAnyOrder("Database connection(s) health check", "Database migration(s) health check"));
```
gsmet:
```suggestion
                            containsInAnyOrder("Database connection(s) health check", "Flyway database migrations health check"));
```

docs/src/main/asciidoc/flyway.adoc
```diff
@@ -194,3 +194,12 @@ public class MigrationService {
 <1> Inject the Flyway object if you want to use it directly
 <2> Use the Flyway instance directly
 
+
+== Flyway migration Health Check
```
gsmet:
```suggestion
== Flyway database migrations Health Check
```

gsmet (changes requested):
Thanks for your contribution! I added a few (minor) suggestions.

Could you have a look?

viniciusfcf:
All done, and rebased.

loicmathieu:
I don't this healthcheck is needed, and I will definilty not enabled it by default.

If you use `migrate-at-start=true` it's not possible for this health check to be DOWN as if Flyway didn't succeed in migrating your database your application will not start. And using this property is the right way to do Flyway migrations.

If you migrate manually, you need to do it very early to avoid potential application issues ... so normally this will be done before your application is stated so it's really not a health check that needs to be done (when the application is started, the migration should have been done).

I don't see any possibility where the applicaiton starts with migration OK and suddently the migrations  are not OK ...

machi1990:
> I don't this healthcheck is needed, and I will definilty not enabled it by default.
> 
> If you use `migrate-at-start=true` it's not possible for this health check to be DOWN as if Flyway didn't succeed in migrating your database your application will not start. And using this property is the right way to do Flyway migrations.
> 
> If you migrate manually, you need to do it very early to avoid potential application issues ... so normally this will be done before your application is stated so it's really not a health check that needs to be done (when the application is started, the migration should have been done).
> 
> I don't see any possibility where the applicaiton starts with migration OK and suddently the migrations are not OK ...

+1 @loicmathieu

gsmet:
Maybe, let's ask what @viniciusfcf had in mind and what are the cases he envisioned?

viniciusfcf:
@loicmathieu is right, the default  value "true" does not make sense if the right way is migration-at-start. However still is possible manual migration.

I was doing local tests using manual migration to simulate a drop-and-create using clean and migrate commands

gsmet:
My understanding is that it can be useful with certain configuration and doesn't do any harm in others. If so, I would tend to think it can be useful.

Or did I miss something?

loicmathieu:
If it has little usage (I expect 80% of user using migration-at-start so not needing it) as it implies a cost (memory, CPU, storage) I would say not include it!

At least: disabled it by default as it will pollute the heath response with meaningless information if a user use migration-at-start.

gsmet:
Yeah, the thing is that I'm a bit torn on consistency here. If they are enabled for all the extensions but one, that will be a bit confusing. And the cost if really minimal.

I haven't had the time to check the other health checks yet. Are there others that are disabled?

loicmathieu:
@gsmet all health checks will be enabled by default, at least the 5 ones that are on developement and the one that is already merged.

viniciusfcf:
Do you prefer close this PR or change to false the default value?

Issue #5687, opened 2019-11-22
document how to use 999-SNAPSHOT of quarkus with a universe-platform based project


maxandersen:
see https://quarkusio.zulipchat.com/#narrow/stream/187038-dev/topic/create.20cmd.20issue.20with.20new.20extension/near/181634045

geoand:
cc @aureamunoz

gsmet:
@aureamunoz so should I close this one? I see the PR has been merged.

aureamunoz:
Yes! @gsmet, you can close it. I has been already merged. 
Thank you!!

gsmet:
Thanks for the feedback.

PR #5736, opened 2019-11-25
doc: document how to use 999-SNAPSHOT of quarkus

aureamunoz:
Related to #5687

### DIFF
--- a/docs/src/main/asciidoc/maven-tooling.adoc
+++ b/docs/src/main/asciidoc/maven-tooling.adoc
@@ -44,6 +44,18 @@ The following table lists the attributes you can pass to the `create` command:
 | `1.0-SNAPSHOT`
 | The version of the created project
 
+| `platformGroupId`
+| `io.quarkus`
+| The group id of the target platform. Given that all the existing platforms are coming from `io.quarkus` this one won't practically be used explicitly. But it's still an option.
+
+| `platformArtifactId`
+| `quarkus-universe-bom`
+| The artifact id of the target platform BOM. It should be `quarkus-bom` in order to use the locally built Quarkus.
+
+| `platformVersion`
+| If it's not specified, the latest one will be resolved.
+| The version of the platform you want the project to use. It can also accept a version range, in which case the latest from the specified range will be used.
+
 | `className`
 | _Not created if omitted_
 | The fully qualified name of the generated resource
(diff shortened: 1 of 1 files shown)


aureamunoz:
Hi, here the PR to doc how to be able to generate a project that uses locally built Quarkus. WDYT @geoand @maxandersen @aloubyansky ?

docs/src/main/asciidoc/maven-tooling.adoc
```diff
@@ -40,6 +40,10 @@ The following table lists the attributes you can pass to the `create` command:
 | _mandatory_
 | The artifact id of the created project. Not passing it triggers the interactive mode.
 
+| `platformArtifactId`
+| `quarkus-universe-bom`
+| The artifact id of the of the target platform BOM. It should be `quarkus-bom` in order to use the locally built Quarkus.
```
aloubyansky: Please remove one `of the` in `The artifact id of the of the...`.

aloubyansky (changes requested):
Thanks @aureamunoz There is a small misprint. Could you please fix it?
There actually are other relevant arguments.
`-DplatformGroupId` is the groupId of the target platform, the default value is `io.quarkus`, given that all the existing `platforms` are coming from `io.quarkus` this one won't practically be used explicitly. But it's still an option.
`-DplatformVersion` is the version of the platform you want the project to use. If it's not specified, the latest one will be resolved. It can also accept a version range, in which case the latest from the specified range will be used.
By default, the command will target the latest version of `quarkus-universe-bom` (unless specific coordinates have been specified). If you run offline however, it will look for the latest locally available. If `quarkus-universe-bom` (satisfying the default version range which is currently up to 2.0), it will fallback to the bundled platform based on `quarkus-bom` (the version will match the version of the plugin).
If you execute the command with

aureamunoz:
@aloubyansky do you want also add this paragraphe to the doc? or it's just for clarification?
_By default, the command will target the latest version of quarkus-universe-bom (unless specific coordinates have been specified). If you run offline however, it will look for the latest locally available. If quarkus-universe-bom (satisfying the default version range which is currently up to 2.0), it will fallback to the bundled platform based on quarkus-bom (the version will match the version of the plugin).
If you execute the command with_

aloubyansky:
I think it'd be good to mention it. Please correct this phrase
````
...available. If quarkus-universe-bom (satisfying the default version range which is currently up to 2.0),
````
to 
````
...available and if quarkus-universe-bom (satisfying the default version range which is currently up to 2.0) is not available locally,
````
Thank you!

aloubyansky (approved):

PR #5738, opened 2019-11-25
fix: add spring-security extension metadata

aureamunoz:
I think I missed that when adding the spring-security extension.
WDYT @geoand @michalszynkiewicz @gsmet ?
Related to #5225

### DIFF
--- a/extensions/spring-security/runtime/src/main/resources/META-INF/quarkus-extension.yaml
+++ b/extensions/spring-security/runtime/src/main/resources/META-INF/quarkus-extension.yaml
@@ -1,0 +1,11 @@
+---
+name: "Quarkus Extension for Spring Security API"
+metadata:
+  keywords:
+  - "spring-security"
+  - "spring"
+  - "security"
+  guide: "https://quarkus.io/guides/spring-security"
+  categories:
+  - "compatibility"
+  status: "preview"


extensions/spring-security/runtime/src/main/resources/META-INF/quarkus-extension.yaml
```diff
@@ -0,0 +1,10 @@
+---
+name: "Quarkus Extension for Spring Security API"
+metadata:
+  keywords:
+  - "spring-security"
+  - "spring"
```
gsmet: Maybe add `security` as a keyword too.

gsmet (changes requested):
Looks good, thanks for taking care of that, I added a minor comment.

gsmet:
@aloubyansky @maxandersen any chance we could fail if one extension is missing the metadata?

extensions/spring-security/runtime/src/main/resources/META-INF/quarkus-extension.yaml
aureamunoz: yep, good idea.

aloubyansky:
We surely can fail if no descriptor is provided. OTOH, I like the idea of not requiring a metadata and generating the basic one as the default. It'd be one of the requirements however to have a detailed enough metadata to be accepted in the platform.

gsmet:
@aloubyansky any chance it could be optional and we could enforce that in the Quarkus and Platform builds?

aloubyansky:
I think so. However it actually has to be reviewed by somebody accepting
the extension, it's not something that can be relied upon blindly.

gsmet:
@aloubyansky I think if the file is there, people will review it properly. But I'm not sure they will think of it if it's missing (we have a good example here).

extensions/spring-security/runtime/src/main/resources/META-INF/quarkus-extension.yaml
gsmet: I just force-pushed an amended commit with the `security` keyword added. Let's wait for CI.

gsmet (approved):

maxandersen:
we definitely could do a "platform-lint" that could warn/errror on issues; especially something we could enforce in platform and at least in quarkus hosted extensions.

gsmet:
Merged, thanks!

PR #5761, opened 2019-11-26
Improvements to Gradle plugin build and enable gradle testing again

gastaldi:
This PR refactors the Gradle plugin: 

- Fixes the plugin build warnings 
- Support build using the Gradle build system (`mvn install` still supported for testing locally)
- Bump Gradle to 6.0.1
- Support running Gradle integration and functional tests
- Migrates the generated Gradle project to use the plugin DSL
- Removes the `@author` tags


stalep:
Give me a ping when it's ready George and I'll try to have a look :)

devtools/gradle/src/test/resources/gradle-project/build.gradle
```diff
@@ -5,9 +5,7 @@ plugins {
 }
 
 repositories {
-    maven {
-        url uri(System.getenv('MAVEN_LOCAL_REPO'))
-    }
+    mavenLocal()
```
gsmet: Are we sure this won't be an issue when I'm building things with a specific Maven repo (i.e. when I release)?

devtools/gradle/pom.xml
```diff
@@ -137,83 +77,6 @@
...
-                                    <file>target/${project.artifactId}-${project.version}-sources.jar</file>
-                                    <type>jar</type>
-                                    <classifier>sources</classifier>
-                                </artifact>
-                            </artifacts>
-                            <skipAttach>${skip.gradle.build}</skipAttach>
-                        </configuration>
-                    </execution>
-                </executions>
-            </plugin>
```
gsmet: I'm a bit concerned by all the code you removed. How does it work now?

gsmet (dismissed):
I added 2 questions inline.

And I have another one: shouldn't we also upgrade the generated projects to Gradle 6?

devtools/gradle/src/test/resources/gradle-project/build.gradle
gastaldi: Yes, because `mavenLocal()` reads the `maven.repo.local` that is set in https://github.com/quarkusio/quarkus/pull/5761/files#diff-32bff5c9d7d29fc887c0eb01a4f45724R67

devtools/gradle/pom.xml
gastaldi: This plugin is not supposed to be deployed in Sonatype. These removed snippets were necessary to do that, which are no longer required once we started publishing to the plugin portal

gastaldi:
> shouldn't we also upgrade the generated projects to Gradle 6?

Definitely.

gastaldi:
Marking as WIP until I can verify that the release process is still working

gastaldi:
The `version` in `gradle.properties` is updated as part of the release process, so that's fine

stalep (approved):
Tested locally using Gradle 6.0.1 on a project generated from code.quarkus.io and just changing the pluginVersion and platformVersion to 999-SNAPSHOT (also had to copy the gradle plugin to my .m2/repository since it doesn't do that automatically anymore).
Building the jar, tests and plugins worked as expected. 
The native-image builder did not work, it fails with: "No value has been specified for property 'nativeImageXmx'". I do not believe that is caused by these changes, but rather something in master.

gastaldi:
>also had to copy the gradle plugin to my .m2/repository since it doesn't do that automatically anymore

@stalep yes, the plugin should not be published in Maven anymore, that impacts the generated `settings.gradle` and `build.gradle` (which I am still working on it)

gastaldi:
>The native-image builder did not work, it fails with: "No value has been specified for property 'nativeImageXmx'"

Good catch, I removed the `@Optional` there mistakenly :) I'll fix it asap

gastaldi:
There is only one caveat though: since the plugin is no longer installed in Maven, Gradle projects using `99-SNAPSHOT` will not work. The solution I found so far is to write a Functional test, like in https://github.com/quarkusio/quarkus/pull/5761/files#diff-a43f7db32d7e0ee45ea954a2db621dc5

gastaldi:
Maybe I'll put the code that installs in the local Maven repository back and just skip the deploy plugin, that could work for testing stuff locally if I don't find an elegant solution

geoand:
I'll check it out over the weekend or on Monday cause I'm to sick now to be able to deal with Gradle :P

devtools/platform-descriptor-json/src/main/resources/templates/basic-rest/java/build.gradle-template.ftl
```diff
@@ -26,9 +14,6 @@ dependencies {
 
     testImplementation 'io.quarkus:quarkus-junit5'
     testImplementation 'io.rest-assured:rest-assured'
-
-    nativeTestImplementation 'io.quarkus:quarkus-junit5'
-    nativeTestImplementation 'io.rest-assured:rest-assured'
```
geoand: Were these removed because the dependencies are "inherrited" from the test dependencies?

geoand:
Looks like a very nice improvement!

devtools/platform-descriptor-json/src/main/resources/templates/basic-rest/java/build.gradle-template.ftl
gastaldi: Yes

gastaldi:
Finally the CI passed

devtools/gradle/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
-#Tue Nov 05 11:34:08 BRST 2019
-distributionUrl=https\://services.gradle.org/distributions/gradle-5.6.4-all.zip
+#Tue Nov 26 00:14:48 BRST 2019
+distributionUrl=https\://services.gradle.org/distributions/gradle-6.0.1-all.zip
```
geoand: I wonder if we should just be using the `-bin` variant instead of `-all`. Not that it matters much :)

devtools/gradle/pom.xml
```diff
@@ -21,30 +21,7 @@
         <skip.gradle.build>false</skip.gradle.build>
```
geoand: When do we envision that this will be set? Is it even needed anymore?

devtools/gradle/settings.gradle
```diff
@@ -1 +1,15 @@
+plugins {
+    id "com.gradle.enterprise" version "3.1"
+}
+
+gradleEnterprise {
```
geoand: Are there any implication for using this?

devtools/gradle/src/functionalTest/java/io/quarkus/gradle/QuarkusPluginFunctionalTest.java
```diff
@@ -0,0 +1,71 @@
...
+import io.quarkus.generators.BuildTool;
+import org.gradle.testkit.runner.BuildResult;
+import org.gradle.testkit.runner.GradleRunner;
+import org.gradle.testkit.runner.TaskOutcome;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class QuarkusPluginFunctionalTest {
```
geoand: :+1: :+1: :+1:

devtools/gradle/src/main/java/io/quarkus/gradle/QuarkusPlugin.java
```diff
@@ -57,51 +71,49 @@ private void registerTasks(Project project) {
...
-            t.dependsOn(quarkusTestConfig);
-            // also make each task use the JUnit platform since it's the only supported test environment
-            t.useJUnitPlatform();
-        });
     }
 
     private void verifyGradleVersion() {
-        if (GradleVersion.current().compareTo(GradleVersion.version("5.0")) < 0) {
-            throw new GradleException("Quarkus plugin requires Gradle 5.0 or later. Current version is: " +
+        if (GradleVersion.current().compareTo(GradleVersion.version("6.0")) < 0) {
```
geoand: So we will require Gradle 6.x from now on?

This is something we definitely need to note. 
Also, have you tried generating a gradle project with `1.0.x.y` and tested if just bumping the gradle version will work with this PR?

devtools/gradle/src/test/java/io/quarkus/gradle/QuarkusPluginTest.java
```diff
@@ -33,7 +34,7 @@ public void shouldMakeAssembleDependOnQuarkusBuild() {
 
         TaskContainer tasks = project.getTasks();
 
-        assertTrue(tasks.getByName("assemble").getDependsOn().contains(tasks.getByName("quarkusBuild")));
+        assertThat(tasks.getByName("assemble").getDependsOn()).contains(tasks.getByName("quarkusBuild"));
```
geoand: +100 for AssertJ!

devtools/gradle/src/main/java/io/quarkus/gradle/QuarkusPlugin.java
maxandersen: if it’s the case, would be good to know why we require gradle 6 ...as gradle users aren't agressive at updating afaik :)

devtools/gradle/gradle/wrapper/gradle-wrapper.properties
gastaldi: Intellij changed it so that Gradle sources are available

devtools/gradle/settings.gradle
gastaldi: No, that's the recommended approach for publishing scan builds in scans.gradle.com. see https://docs.gradle.com/enterprise/gradle-plugin/#gradle_6_x_and_later

devtools/gradle/pom.xml
gastaldi: This is useful for speeding up the build when you don't need to test the Gradle plugin. I use it sometimes but I am not opposed to removing it if necessary

devtools/gradle/src/main/java/io/quarkus/gradle/QuarkusPlugin.java
gastaldi: That's why I added the `noteworthy-feature` label :) 

Since I had to change the artifactId, 1.0.x projects will no longer work (the `useModule("quakus-gradle-plugin")` would fail), but I am planning on putting these migration steps in the migration guide: https://github.com/quarkusio/quarkus/wiki/Migration-Guide

devtools/gradle/gradle/wrapper/gradle-wrapper.properties
geoand: But do we need them? I guess they help when developing the plugin...

devtools/gradle/src/main/java/io/quarkus/gradle/QuarkusPlugin.java
gastaldi: > if it’s the case, would be good to know why we require gradle 6 ...as gradle users aren't agressive at updating afaik :)

TBH we are not using any specific Gradle 6.x feature, I'll have a.look

devtools/gradle/pom.xml
geoand: Just asking to understand if it still makes sense, I'm not saying we should remove it

devtools/gradle/gradle/wrapper/gradle-wrapper.properties
gastaldi: Yes it helped me a lot to debug and understand how some Gradle features work.

devtools/gradle/src/main/java/io/quarkus/gradle/QuarkusPlugin.java
maxandersen: wait - why did you *have* to change the artifactid ?  

on gradle 5 vs 6 then I think better to stay conservative here and support 5+ (actually think we know we need higher than just 5.0, to make the enforced platform feature work)

devtools/gradle/pom.xml
```diff
@@ -10,7 +10,7 @@
     </parent>
     <modelVersion>4.0.0</modelVersion>
 
-    <artifactId>quarkus-gradle-plugin</artifactId>
+    <artifactId>io.quarkus.gradle.plugin</artifactId>
```
maxandersen: do we absolutely need to change this ? 

if we do - doesn't gradle have some capability of handling renames somewhat transparently ?

devtools/gradle/src/main/java/io/quarkus/gradle/QuarkusPlugin.java
gastaldi: @maxandersen  as I explained above, changing the artifactId allows us to remove this snippet from the settings.gradle: 

```groovy
    resolutionStrategy {
        eachPlugin {
            if (requested.id.id == 'io.quarkus') {
                useModule("io.quarkus:quarkus-gradle-plugin:1.0.1.Final")
            }
        }
    }
```

I just tested, the plugin still works in Gradle 5, it's just that generated projects should target 6.x from now on. If you still need this plugin running on Gradle 5, some (minor) adjustments to the build.gradle and settings.gradle will be necessary (hint: it simplifies it a lot)

I'll make sure to put the necessary migration steps in the Migration guide for Gradle 5.x users.

devtools/gradle/pom.xml
gastaldi: See https://github.com/quarkusio/quarkus/pull/5761#discussion_r354286243. This was corrected to the artifactId that is understood by Gradle, allowing us to remove the `resolutionStrategy` snippet in settings.gradle

devtools/gradle/src/main/java/io/quarkus/gradle/QuarkusPlugin.java
gastaldi: Also, since the plugin is no longer deployed to Maven central, `useModule` will not work

maxandersen: I'm all for simplification :) 

Can we update the generated code part in this PR too ? just so that does not get forgotten...

I for one would like to see the actual differences incl. how gradle 5 and 6 usages will differ since then generation *wont* simple ;)

gastaldi: I have updated `build.gradle` and `settings.gradle` in this PR already, feel free to compare its contents with the one produced by 1.0.1.Final. I'm sure you'll like it :)

gastaldi:
Okay, I think this PR is ready to be merged.

geoand:
> Okay, I think this PR is ready to be merged.

I haven't really tested it but the code looks OK, so I think @aloubyansky should have the final say

maxandersen:
Here is my only concern with this PR - it will break users existing gradle apps; which in itself is kinda ok since we already with #5101 and gradle preview status indicated things would have to break/change.

How confident are we that once we start fixing core issues of the gradle plugin (especially the classloader issues) that we don't have other user breaking things happening in 1.2 ?

I *think* we should be fine as with the changes there are very little touchpoints left in the build.gradle/settings files which is a *good* thing.  Curious to hear what other think of this and if there are more we can do or this is as good as it gets for now ;)

geoand:
I must admit that I haven't been able to follow this one as much as I would like. Hopefully I can dedicate some time over the weekend to do so.

aloubyansky:
> Here is my only concern with this PR - it will break users existing gradle apps;

Let's clarify this. What is going to break?

aloubyansky:
I generated a project with 1.0.1.Final, changed 1.0.1.Final to 999-SNAPSHOT in the properties (using @gastaldi's branch for 999-SNAPSHOT) and re-built the project successfully. Then generated a project with 999-SNAPSHOT, changed 999-SNAPSHOT to 1.0.1.Final in the properties and re-build went w/o an issue.
I tried both Gradle 6.0.1 and 5.6.2.

maxandersen:
> Let's clarify this. What is going to break?

users that created gradle projects with 1.0 will have to update build.gradle/settings as currently documented at https://github.com/quarkusio/quarkus/wiki/Migration-Guide#gradle-plugin

and at 1.0 we changed the setup to; so i'm just trying to reduce the problem of having to tell gradle users to change their setup at once again at 1.1 and then again at 1.2 ( or whenever we get the classloader issues fixed :)

aloubyansky:
Still, what is going to break?
I generated a project with 1.0.0.Final using Gradle 5.6.2., changed the properties to 999-SNAPSHOT, added a new extension using `gradle addExtension...` and re-built it successfully.

aloubyansky:
BTW, classloader fixes are going to fix Gradle builds, I don't think they will introduce any incompatibility.

maxandersen:
> Still, what is going to break?
> I generated a project with 1.0.0.Final using Gradle 5.6.2., changed the properties to 999-SNAPSHOT, added a new extension using `gradle addExtension...` and re-built it successfully.

I assume you were then actually using older version of plugin as in this change the artifact id change so generated projects wont be using the new plugin.

and when you update the version it shuold fail as the plugin would not be avaialble when we release 1.1

gastaldi:
> BTW, classloader fixes are going to fix Gradle builds, I don't think they will introduce any incompatibility.

Yes, just like this PR, there shouldn't be any API changes. What this PR proposes is what a regular Gradle user would expect to have in their build files, I just removed the hacked part :smile:

gsmet:
> I assume you were then actually using older version of plugin as in this change the artifact id change so generated projects wont be using the new plugin.

Can you answer that one? That would be a big issue IMHO.

BTW, even if it doesn't break anything, I really don't like the artifactId change, is it really necessary? It's really not consistent with our other artifactIds.

gastaldi:
> Still, what is going to break?
> I generated a project with 1.0.0.Final using Gradle 5.6.2., changed the properties to 999-SNAPSHOT, added a new extension using `gradle addExtension...` and re-built it successfully.

Try removing your `~/.m2/repository/io/quarkus/quarkus-gradle-plugin`. You should see the error Max is talking about

gastaldi:
> > I assume you were then actually using older version of plugin as in this change the artifact id change so generated projects wont be using the new plugin.
> 
> Can you answer that one? That would be a big issue IMHO.
> 
> BTW, even if it doesn't break anything, I really don't like the artifactId change, is it really necessary? It's really not consistent with our other artifactIds.

Again, this new artifactId is what Gradle expects when looking for a plugin in the local Maven repository without using `useModule()` in settings.gradle. Since this plugin will never be published in Maven central, that won't be a problem

geoand:
I think following Gradle conventions is a good thing to do and the sooner we do it, the better.

However I am still not clear on what will happen for users with generated `1.0.x` projects when they try to move to `1.1.0.x`. 
What will they have to change?

gastaldi:
It's in the migration guide: https://github.com/quarkusio/quarkus/wiki/Migration-Guide#gradle-plugin

geoand:
Thanks @gastaldi!  

I think it's reasonable and hopefully it will be the last change that users will need to make

gastaldi:
Yes, I am a bit surprised of why it wasn't like this PR proposes since day one :)

geoand:
I'm +1 for this as long as it's properly communicated

gastaldi:
That depends on what goes in the release notes, a link to the migration guide should be enough I think ☺️

geoand:
Well, a big fat warning in the release notes would be nice as well :)

jaikiran:
Just a FYI - A user has raised a valid point here about the artifact id change https://github.com/quarkusio/quarkus/issues/3592#issuecomment-562631400. We would need to change/update code.quarkus.io to use this new artifact id, once we release 1.1.0 of Quarkus, right?

maxandersen:
@jaikiran yes, but since code.quarkus.io uses quarkus to do the generation that happens autoatmically when code.quarkus.io gets updated :)

PR #5768, opened 2019-11-26
Fix HTTPS issue on native for OAuth2

### LINKED ISSUE #5731: How to set correctly ssl certificate oauth2
Hi folks,

I'm trying to use oauth2 following this guide : https://quarkus.io/guides/security-oauth2 and also https://quarkus.io/guides/native-and-ssl

In application.properties, i use for the `quarkus.oauth2.introspection-url` an **https** adress ([auth0](https://auth0.com/)).

When i execute these commands separatly : 

```
./mvnw compile quarkus:dev
./mvnw clean install -Pnative 
```

I getting this error : 
> Caused by: java.lang.IllegalArgumentException: Parameter 'sslContext' may not be null

The complete stacktrace : [error-log.txt](https://github.com/quarkusio/quarkus/files/3884061/error-log.txt)

I use graalVM, the `GRAALVM_HOME` is set and the certificate of auth0 is added to the truststore of graalVM, follwing these commands :

```
echo | openssl s_client -servername NAME -connect HOST:PORT |\
  sed -ne '/-BEGIN CERTIFICATE-/,/-END CERTIFICATE-/p' > certificate.crt

/home/xerxes/graalvm-ce-19.2.1/bin/keytool -import -alias auth0 -storepass 'changeit' -keystore /home/xerxes/graalvm-ce-19.2.1/jre/lib/security/cacerts -file certificate.crt
```
I check with this command `keytool -list -v -keystore /home/xerxes/graalvm-ce-19.2.1/jre/lib/security/cacerts`, i have my certificate in there.

Could you explain me what i missed in my steps ?

I saw this option `quarkus.oauth2.ca-cert-file` but i don't know very well how to use it.

Thank you so much guys for your help !

loicmathieu:
Fixes #5731

~~I don't like the way I fix it but I cannot find a better way.~~

The issue was that the Elytron `OAuth2IntrospectValidator` needs a SSLContext inside it's constructor in case of HTTPS introspect URL.

As all security related build steps are STATIC_INIT, the constructor is called at static initilization time during the native process, and GraalVM refuses to have an SSLContext in it's image heap so it crash.

~~I try to switch the security related build items to RUNTIME_INIT but cannot success, maybe it's possible but there is a lot of different build items procudes and consumes by some other extensions so it's risky to change that.~~

~~I try to provide a substitution but as the `OAuth2IntrospectValidator` use a private builder with private fields I cannot substitute the constructor (maybe I will also need to substitute the builder ...).~~

~~So the best implementation I can find is to copy/paste the `OAuth2IntrospectValidator` and make the SSLContext related stuff lazy.~~

**UPDATE:** All security related build steps has been moved to RUNTIME_INIT so an easy fix can be done.

Tested OK in both JVM and native builds.

### DIFF
--- a/extensions/elytron-security-oauth2/runtime/src/main/java/io/quarkus/elytron/security/oauth2/runtime/OAuth2Recorder.java
+++ b/extensions/elytron-security-oauth2/runtime/src/main/java/io/quarkus/elytron/security/oauth2/runtime/OAuth2Recorder.java
@@ -39,6 +39,8 @@ public class OAuth2Recorder {
 
         if (config.caCertFile.isPresent()) {
             validatorBuilder.useSslContext(createSSLContext(config));
+        } else {
+            validatorBuilder.useSslContext(SSLContext.getDefault());
         }
 
         OAuth2IntrospectValidator validator = validatorBuilder.build();


extensions/elytron-security-oauth2/runtime/src/main/java/io/quarkus/elytron/security/oauth2/runtime/auth/QuarkusOAuth2IntrospectValidator.java
```diff
@@ -0,0 +1,291 @@
...
+    private final String clientSecret;
+    private final SSLContext sslContext;
+    private final HostnameVerifier hostnameVerifier;
+
+    QuarkusOAuth2IntrospectValidator(QuarkusOAuth2IntrospectValidator.Builder configuration) {
+        this.tokenIntrospectionUrl = Assert.checkNotNullParam("tokenIntrospectionUrl", configuration.tokenIntrospectionUrl);
+        this.clientId = Assert.checkNotNullParam("clientId", configuration.clientId);
+        this.clientSecret = Assert.checkNotNullParam("clientSecret", configuration.clientSecret);
+
+        // SSLContext is lazily provided so we remove the check
```
sberyozkin: Please remove this commented block

extensions/elytron-security-oauth2/runtime/src/main/java/io/quarkus/elytron/security/oauth2/runtime/auth/QuarkusOAuth2IntrospectValidator.java
```diff
@@ -0,0 +1,291 @@
...
+     * @param clientId the identifier of a client within the OAUth2 Authorization Server
+     * @param clientSecret the secret of the client
+     * @param token the access token to introspect
+     * @param sslContext the ssl context
+     * @param hostnameVerifier the hostname verifier
+     * @return a @{JsonObject} representing the response from the introspection endpoint or null if
+     */
+    private JsonObject introspectAccessToken(URL tokenIntrospectionUrl, String clientId, String clientSecret, String token,
+            SSLContext sslContext, HostnameVerifier hostnameVerifier) throws RealmUnavailableException {
+        Assert.checkNotNullParam("clientId", clientId);
```
sberyozkin: Please remove the 1st 2 checks since they are already done in the constructor, and I'd also drop this `sslContext` parameter and set up the sslContext in this method as there is also a same named property.

extensions/elytron-security-oauth2/runtime/pom.xml
```diff
@@ -22,6 +22,10 @@
             <groupId>org.wildfly.security</groupId>
             <artifactId>wildfly-elytron-realm-token</artifactId>
         </dependency>
+        <dependency>
```
sberyozkin: Should it be `quarkus-jsonb` ?

sberyozkin:
@loicmathieu IMHO copying it is not a problem, ideally we'd have `quarkus-oidc` dealing with the opaque tokens too (as OIDC may have binary tokens) or as suggested by Stuart this extension wiill be rewritten to become reactive, so this is why I believe copying to get past this issue is fine :-)

loicmathieu:
@sberyozkin the changes your requested are all coming from the fact I copy the class and wants to keep it as close as possible than the existing one in case we upstream the issue on elytron side.

Let's wait from what others think of it, and if we don't plan to upstream the issue I will make the change to make it pass Quarkus standard :)

sberyozkin:
@loicmathieu Sure, you are right, while we have this extension it does make sense to try to fix it in the wildfly code as well.

gsmet:
I would prefer us to fix the issue at the source rather to have a copy of it.

Security things are tricky and I would hate having a security issue because we copied a class in Quarkus.

loicmathieu:
@gsmet the intent is to provide a temporary fix (this PR) while proposing a solution on the wildfly side. I'm not sure my fix will be accepted at wildfly as it's really the way Quarkus uses GraalVM that causes the isssue (initializaing classes at static init).

An other solution should be to move to RUNTIME_INIT all the security steps but I didn't succeed (maybe someone else can try).

If you prefere to wait for an answer at Elytron side I will open an issue and we will see ...

loicmathieu:
Bug open at Elytron side: https://issues.jboss.org/browse/ELY-1911

gsmet:
@sberyozkin do you know who is responsible for Elytron development?

@loicmathieu I agree that having a band aid for now is mostly OK but I want to have a clear plan on the Elytron side to be sure the band aid doesn't become permanent. I don't mean waiting for a release but at least having a plan.

sberyozkin:
@darranl Hi, can you please recommend who can follow up with a minor fix on the Elytron side for this issue ?

loicmathieu:
On ice because of #5937

loicmathieu:
@sberyozkin @gsmet now that #5937 I can provides an easy fix for this :)
Can you review it? 
By providing a default SSLContext, https can be used whitout a custom certificat file.

Fix on Elytron side is not needed anymore, I will close the bug report (nobody took action on it anyway).

I will try to use a certificat file in native mode to see if it's possible now that we move everything ot runtime. If yes I will update the documentation.

gsmet (approved):

loicmathieu:
CI failure on Quarks integration test, I rebase to master to trigger a new build as it's unrelated to the change.

gsmet:
Merged, thanks.

PR #5777, opened 2019-11-26
Fix AnalysisConfigurer example in Hibernate Search documentation

viniciusfcf:
Fix configure parameter Classs. Changing ElasticsearchAnalysisDefinitionContainerContext to ElasticsearchAnalysisConfigurationContext

### DIFF
--- a/docs/src/main/asciidoc/hibernate-search-elasticsearch.adoc
+++ b/docs/src/main/asciidoc/hibernate-search-elasticsearch.adoc
@@ -431,13 +431,13 @@ To fulfill our requirements, let's create the following implementation:
 ----
 package org.acme.hibernate.search.elasticsearch.config;
 
+import org.hibernate.search.backend.elasticsearch.analysis.ElasticsearchAnalysisConfigurationContext;
 import org.hibernate.search.backend.elasticsearch.analysis.ElasticsearchAnalysisConfigurer;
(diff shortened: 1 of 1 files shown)


docs/src/main/asciidoc/hibernate-search-elasticsearch.adoc
```diff
@@ -476,8 +476,8 @@ In our existing `LibraryResource`, we just need to inject the following methods
     @GET
     @Path("author/search") // <3>
     @Transactional
-    public List<Author> searchAuthors(@QueryParam String pattern, // <4>
-            @QueryParam Optional<Integer> size) {
+    public List<Author> searchAuthors(@QueryParam("pattern") String pattern, // <4>
+            @QueryParam("size") Optional<Integer> size) {
```
gsmet: This change shouldn't be needed, see the note: you should use `org.jboss.resteasy.annotations.jaxrs.QueryParam` for the import.

gsmet (changes requested):
The first commit is legit, thanks for the fix. But the second one is not. Could you remove the second commit?

Thanks!

docs/src/main/asciidoc/hibernate-search-elasticsearch.adoc
viniciusfcf: Ok, sorry.

gsmet (approved):

gsmet:
Cool! Thanks for taking the time to create a PR, very much appreciated.

Issue #5722, opened 2019-11-24
Gradle plugin doesn't support multi module project well.


SetoKaiba:
**Describe the bug**
I want to use quarkus project as sub module project. But the plugin is not working well.

**Expected behavior**
It should work with multi module project.

**Actual behavior**
https://github.com/SetoKaiba/blueeyes
Here's the test multi module project.
1. It does not recognize the gradle.properties file in the root project. It will work with placing it in the sub module. I use the subprojects DSL to simplify the configuration. So I have to duplicate the properties in parent and sub moudules.
```
#Gradle properties
#Sat Nov 23 16:34:25 UTC 2019
quarkusPluginVersion=1.0.0.CR2
quarkusPlatformArtifactId=quarkus-universe-bom
quarkusPlatformVersion=1.0.0.CR2
quarkusPlatformGroupId=io.quarkus
```

```
subprojects{
    // this block is necessary to make enforcedPlatform work for Quarkus plugin available
    // only locally (snapshot) that is also importing the Quarkus BOM
    buildscript {
        repositories {
            mavenLocal()
            maven {
                url "https://repository.jboss.org/nexus/content/groups/public-jboss"
            }
            jcenter()
        }
        dependencies {
            classpath "io.quarkus:quarkus-gradle-plugin:${quarkusPluginVersion}"
        }
    }
}
```

2. gradlew :blueeyes-web:addExtension --extensions="io.quarkus:quarkus-hibernate-orm"
It will generate settings.gradle in sub module. It will cause problems for sub module.
The problem is below after settings.gradle is generated. I have to delete it manually to avoid the problem.

```
14:37:22: Executing task 'listExtensions'...


FAILURE: Build failed with an exception.

* Where:
Build file 'D:\Quarkus\blueeyes\blueeyes-web\build.gradle' line: 1

* What went wrong:
A problem occurred evaluating root project 'blueeyes-web'.
> Plugin with id 'io.quarkus' not found.

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Get more help at https://help.gradle.org

BUILD FAILED in 0s
14:37:23: Task execution finished 'listExtensions'.
```

3. Argument in IDEA is not working. It seems a bug of IntelliJ IDEA. 
https://youtrack.jetbrains.com/issue/IDEA-227496
https://youtrack.jetbrains.com/issue/IDEA-196952

**To Reproduce**
1. 
2. 
[...]

PR #5780, opened 2019-11-26
Better multi-module support for gradle when adding extension

SetoKaiba:
I made a PR. But I didn't test it. Because I get a lot of error when open the maven projects when solving dependencies in IDEA. #5722

### DIFF
--- a/independent-projects/tools/common/src/main/java/io/quarkus/cli/commands/file/GradleBuildFile.java
+++ b/independent-projects/tools/common/src/main/java/io/quarkus/cli/commands/file/GradleBuildFile.java
@@ -23,19 +23,43 @@ public class GradleBuildFile extends BuildFile {
     private static final String SETTINGS_GRADLE_PATH = "settings.gradle";
     private static final String GRADLE_PROPERTIES_PATH = "gradle.properties";
 
+    private final ProjectWriter rootWriter;
+
     private Model model;
 
     public GradleBuildFile(ProjectWriter writer) {
         super(writer, BuildTool.GRADLE);
+        rootWriter = writer;
+    }
+
+    public GradleBuildFile(ProjectWriter writer, ProjectWriter rootWriter) {
+        super(writer, BuildTool.GRADLE);
+        this.rootWriter = rootWriter;
+    }
+
+    protected void rootWrite(String fileName, String content) throws IOException {
+        rootWriter.write(fileName, content);
+    }
+
+    protected ProjectWriter getRootWriter() {
+        return rootWriter;
     }
 
     @Override
     public void close() throws IOException {
-        write(SETTINGS_GRADLE_PATH, getModel().getSettingsContent());
+        if (getWriter() == getRootWriter()) {
+            write(SETTINGS_GRADLE_PATH, getModel().getSettingsContent());
+        } else {
+            rootWrite(SETTINGS_GRADLE_PATH, getModel().getSettingsContent());
+        }
         write(BUILD_GRADLE_PATH, getModel().getBuildContent());
         try (ByteArrayOutputStream out = new ByteArrayOutputStream()) {
             getModel().getPropertiesContent().store(out, "Gradle properties");
-            write(GRADLE_PROPERTIES_PATH, out.toString(StandardCharsets.UTF_8.toString()));
+            if (getModel().isRootProperties()) {
+                rootWrite(GRADLE_PROPERTIES_PATH, out.toString(StandardCharsets.UTF_8.toString()));
+            } else {
+                write(GRADLE_PROPERTIES_PATH, out.toString(StandardCharsets.UTF_8.toString()));
+            }
         }
     }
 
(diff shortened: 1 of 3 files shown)


SetoKaiba:
I finally sucessfully tested working on my side. Force push with the sucessfully tested commit.

geoand:
Thanks for the PR @SetoKaiba!

@maxandersen how would you like to handle PRs like this?

aloubyansky:
Thanks for your contribution @SetoKaiba A couple of questions though. Looks like this change is adding support for multimodule projects of two levels, i.e. a parent and children but not more, correct? It looks like you could move all the changes from `BuildFile` to `GradleBuildFile`? Maven won't really benefit from this change, right?
We'll be re-working this API and the impl. However if this change helps you in your development I'll be happy to merge it.

SetoKaiba:
>  Looks like this change is adding support for multimodule projects of two levels, i.e. a parent and children but not more, correct?

@aloubyansky Not exactly. If it's with a parent, it's a sub module. The parent is just to determine whether it's a sub module. It does support multi module project with 2 levels and more.

- For settings.gradle, always use the root project's. With it exists, the task of sub module won't run.
- For gradle.properties, if there's one in sub module. Apply changes to it. If not, apply to the root project's. Because the sub module gradle properties will override the root's.

> It looks like you could move all the changes from BuildFile to GradleBuildFile? Maven won't really benefit from this change, right?

Yes. I can move them all to the BuildFile. I just add to the BuildFile in case that the maven will need a root project writer in the future.

> We'll be re-working this API and the impl. However if this change helps you in your development I'll be happy to merge it.

Yes. It would help. Because I may add or remove extentions. Deleting settings.gradle every time is frustrate. And also, I'd like to use gradle.properties in the root project to manage the properties universally,

aloubyansky:
Alright, thanks @SetoKaiba Could you please move the changes from BuildFile to GradleBuildFile then for now?

SetoKaiba:
>Alright, thanks @SetoKaiba Could you please move the changes from BuildFile to GradleBuildFile then for now?
OK. Force push one.

But for now the gradle properties file is still not perfect. Sub module one will overide the root one. But only the shared properties. But for now, when there's sub module I'll write it to sub module. If not, it will write to the root one. But someone may want to manage the versions in the root properties while there're some sub module specific properties. Any better idea? Or any improvement on the re-working one?

aloubyansky:
I suppose, theoretically, if both the submodule and its parent include the file, you could check which properties the submodule overrides and based on that figure out which property belongs where?

SetoKaiba:
@aloubyansky Thank you. I made it with a better check for the gradle properties file now. It should be ok.

aloubyansky (approved):
Given the current state of things, imo, it's good enough. Anybody else has any other remarks?
Thanks @SetoKaiba !

geoand:
If it's good for @aloubyansky it's good for me :)

maxandersen:
updated description to match what it does - please object/change if it applies to more than addextension for gradle multi model projects.

could we add a Integration test for this usecase ? just to be sure we actually have an example of what kind of multi module project we support.

gsmet:
@maxandersen my understanding is that we don't have any Gradle IT at the moment due to some infrastructure issues.

maxandersen:
I though the lack of running gradle tests are being fixed in #5761 ? in any case - at leats having a base example added which we then can hook up when #5761 is ready to go would be great.

gastaldi:
BTW #5761 is ready to be reviewed. That would allow adding some tests in this PR

aloubyansky:
Let's merge this one. WDYT @gastaldi?

gastaldi:
I think it's better to merge #5761 first and then add some tests to this PR, but it's your call :)

aloubyansky:
@SetoKaiba would you like to push this a bit further by adding a test?

gsmet:
@gastaldi could you point @SetoKaiba to the existing tests so that he can mimic what's existing?

Thanks.

gastaldi:
Sure, the functional tests for the Gradle plugin are in https://github.com/quarkusio/quarkus/blob/master/devtools/gradle/src/functionalTest/java/io/quarkus/gradle/QuarkusPluginFunctionalTest.java. Feel free to add new tests in there as needed

SetoKaiba:
Hi. I don't know how to create a multi-module project with the tests. I modify the created project with to support multi module myself.

independent-projects/tools/common/src/main/java/io/quarkus/cli/commands/file/GradleBuildFile.java
```diff
@@ -23,19 +23,43 @@
...
+    private final ProjectWriter rootWriter;
+
     private Model model;
 
     public GradleBuildFile(ProjectWriter writer) {
         super(writer, BuildTool.GRADLE);
+        rootWriter = writer;
+    }
+
+    public GradleBuildFile(ProjectWriter writer, ProjectWriter rootWriter) {
```
gastaldi: It seems that you don't need two ProjectWriters here, but two `GradleBuildFile` instances in QuarkusAddExtension? I'll review more carefully next week.

SetoKaiba: I don't think so. Or you have to make the AddExtensions more complex. Because you're reading the different folder based on whether it's multi-module to init the model and writing the different folder when it's closed.

gastaldi: I may be missing something but I'd expect each `GradleBuildFile` to be responsible for maintaining a single `build.gradle`. Perhaps `GradleBuildFile` should have a `GradleBuildFile getParent()` method instead. 

Not saying your solution is wrong, just that it's confusing to have a single `GradleBuildFile` changing two files internally

SetoKaiba: - Single Module
project/build.gradle
project/gradle.properties
project/settings.gradle

- Multi Module (required properties from parent or no required properties in both gradle.properties)
sub/build.gradle
parent/gradle.properties
parent/settings.gradle

- Multi Module (required properties from sub)
sub/build.gradle
sub/gradle.properties
parent/settings.gradle

I think three files combined is a GradleBuildFile.

SetoKaiba: There's only one file for one type. It does not dealing with both.

SetoKaiba:
When will it be merged?
I tried with my custom build plugin 999-SNAPSHOT. But I encounter this bug.
https://github.com/quarkusio/quarkus-quickstarts/issues/401

gsmet:
@gastaldi what's missing to drive this puppy home?

andreas-eberle:
What's missing for this? The lack of gradle multi module support is a real pain.

SetoKaiba:
Yes. I'd like to know as well. This is solving the add extension and list extensions.

aloubyansky:
@SetoKaiba I've rebased your commit and added a test in #7853. I am closing this one. Thanks a lot for you contribution!

PR #5782, opened 2019-11-26
Support for the fixed redirect URI

### LINKED ISSUE #5733: OIDC Redirect URL should be static absolute URL
**Describe the bug**
Currently Quarkus is including the full [request path](https://github.com/quarkusio/quarkus/blob/21cd452f7efe9936cd0cbc104a8644d872d13815/extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/CodeAuthenticationMechanism.java#L162) in the redirect URL. According to the OAuth 2.0 specification that OIDC is based on the redirect URL should be [absolute](https://tools.ietf.org/html/rfc6749#section-3.1.2).

Without a constant redirect path it is not possible to configure a single whitelisted login redirect URI in the OIDC IDP. 

For example if http://localhost:5000/ is added to the IDP allowed redirect URLs if an unauthenticated user access http://localhost:5000/profile the redirect URL is sent as http://localhost:5000/profile which is not allowed in the IDP and the IDP displays an error.

**Expected behavior**
The redirect URL should have a static path like / or /oidc. The initially requested path should be stored in the web session or relayState and after the OAuth code authentication is performed the response should be redirected back to the initially requested path. 

**Actual behavior**
Whatever unauthenticated request path that is submitted to Quarkus is set as the redirect URL. There are an infinite possible request path variations and it is not possible to include them all or configure wildcards in the IDP which would be a violation of the OAuth standard.

**To Reproduce**
1. Enable the OIDC extension
2. In the IDP add the application's base URL as the redirect URL, i.e. http://localhost:5000
3. Access the local application, OIDC authentication works as expected.
4. Open an anonymous browser session, access http://localhost:5000/test
5. http://localhost:5000/test is sent as the redirect URL, this value is not in the applications approved redirect URL list, the IDP rejects the login request.

**Configuration**
```
quarkus.oidc.auth-server-url=https://xxxxxxx.okta.com
quarkus.oidc.client-id=XXXXXXXXXXXXXXX
quarkus.oidc.credentials.secret=XXXXXXXXXXXXXXXXX
quarkus.oidc.authentication.scopes=openid,profile,groups


quarkus.http.auth.permission.authenticated1.paths=/*
quarkus.http.auth.permission.authenticated1.policy=authenticated

quarkus.http.auth.permission.permit1.paths=/health/*
quarkus.http.auth.permission.permit1.policy=permit
quarkus.http.auth.permission.permit1.methods=GET

```

sberyozkin:
Fixes #5733 . 
This PR introduces a (relative) `redirect-path` property, which, if set, will be used to calculate a static `redirect_uri` parameter. The remaining (dynamic part) of the current request URI will be stored in a `q_auth` state cookie (queries parameters can also be saved but it can be done in a follow up PR).

Right now the problem is that a `RoutingContext.reroute` method does not seem to work.
Note a 'state' parameter (as well as a state cookie) are not added unless a dynamic path is available because there is nothing else at the moment that is being saved.

### DIFF
--- a/extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/CodeAuthenticationMechanism.java
+++ b/extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/CodeAuthenticationMechanism.java
@@ -11,6 +11,8 @@ import java.util.function.Function;
 
 import javax.enterprise.context.ApplicationScoped;
 
+import org.jboss.logging.Logger;
+
 import io.netty.handler.codec.http.HttpResponseStatus;
 import io.quarkus.oidc.AccessTokenCredential;
 import io.quarkus.oidc.IdTokenCredential;
@@ -19,6 +21,7 @@ import io.quarkus.security.AuthenticationFailedException;
 import io.quarkus.security.identity.IdentityProviderManager;
 import io.quarkus.security.identity.SecurityIdentity;
 import io.quarkus.security.runtime.QuarkusSecurityIdentity;
+import io.quarkus.vertx.http.runtime.security.AuthenticationRedirectException;
 import io.quarkus.vertx.http.runtime.security.ChallengeData;
 import io.vertx.core.http.Cookie;
 import io.vertx.core.http.HttpHeaders;
@@ -31,9 +34,11 @@ import io.vertx.ext.web.impl.CookieImpl;
 @ApplicationScoped
 public class CodeAuthenticationMechanism extends AbstractOidcAuthenticationMechanism {
 
+    private static final Logger LOG = Logger.getLogger(CodeAuthenticationMechanism.class);
+
     private static final String STATE_COOKIE_NAME = "q_auth";
     private static final String SESSION_COOKIE_NAME = "q_session";
-    private static final String SESSION_COOKIE_DELIM = "___";
+    private static final String COOKIE_DELIM = "___";
 
     private static QuarkusSecurityIdentity augmentIdentity(SecurityIdentity securityIdentity,
             String accessToken,
@@ -62,7 +67,7 @@ public class CodeAuthenticationMechanism extends AbstractOidcAuthenticationMecha
 
         // if session already established, try to re-authenticate
         if (sessionCookie != null) {
-            String[] tokens = sessionCookie.getValue().split(SESSION_COOKIE_DELIM);
+            String[] tokens = sessionCookie.getValue().split(COOKIE_DELIM);
             return authenticate(identityProviderManager, new IdTokenCredential(tokens[0]))
                     .thenCompose(new Function<SecurityIdentity, CompletionStage<SecurityIdentity>>() {
                         @Override
@@ -78,9 +83,8 @@ public class CodeAuthenticationMechanism extends AbstractOidcAuthenticationMecha
 
     @Override
     public CompletionStage<ChallengeData> getChallenge(RoutingContext context) {
-        removeSessionCookie(context);
+        removeCookie(context, SESSION_COOKIE_NAME);
         ChallengeData challenge;
-
         JsonObject params = new JsonObject();
 
         List<Object> scopes = new ArrayList<>();
@@ -89,8 +93,12 @@ public class CodeAuthenticationMechanism extends AbstractOidcAuthenticationMecha
         config.authentication.scopes.ifPresent(scopes::addAll);
 
         params.put("scopes", new JsonArray(scopes));
-        params.put("redirect_uri", buildRedirectUri(context));
-        params.put("state", generateState(context));
+
+        URI absoluteUri = URI.create(context.request().absoluteURI());
+        String dynamicPath = getDynamicPath(context, absoluteUri);
+        params.put("redirect_uri", buildCodeRedirectUri(context, absoluteUri, dynamicPath));
+
+        params.put("state", generateState(context, dynamicPath));
 
         challenge = new ChallengeData(HttpResponseStatus.FOUND.code(), HttpHeaders.LOCATION, auth.authorizeURL(params));
 
@@ -105,9 +113,53 @@ public class CodeAuthenticationMechanism extends AbstractOidcAuthenticationMecha
         if (code == null) {
             return CompletableFuture.completedFuture(null);
         }
+
         CompletableFuture<SecurityIdentity> cf = new CompletableFuture<>();
+
+        URI absoluteUri = URI.create(context.request().absoluteURI());
+
+        Cookie stateCookie = context.getCookie(STATE_COOKIE_NAME);
+        if (stateCookie != null) {
+            List<String> values = context.queryParam("state");
+            // IDP must return a 'state' query parameter and the value of the state cookie must start with this parameter's value
+            if (values.size() != 1 || !stateCookie.getValue().startsWith(values.get(0))) {
+                cf.completeExceptionally(new AuthenticationFailedException());
+                return cf;
+            } else if (context.queryParam("pathChecked").isEmpty()) {
+                // This is an original redirect from IDP, check if the request path needs to be updated
+                String[] pair = stateCookie.getValue().split(COOKIE_DELIM);
+                if (pair.length == 2) {
+                    // The extra path that needs to be added to the current request path
+                    String extraPath = pair[1];
+                    // Adding a query marker that the state cookie has already been used to restore the path
+                    // as deleting it now would increase the risk of CSRF
+                    String extraQuery = "?pathChecked=true";
+
+                    // The query parameters returned from IDP need to be included
+                    if (absoluteUri.getRawQuery() != null) {
+                        extraQuery += ("&" + absoluteUri.getRawQuery());
+                    }
+
+                    String localRedirectUri = buildLocalRedirectUri(context, absoluteUri, extraPath + extraQuery);
+                    LOG.debug("Local redirectUri: " + localRedirectUri);
+
+                    cf.completeExceptionally(new AuthenticationRedirectException(localRedirectUri));
+                    return cf;
+                }
+                // The redirect path matches the original request path, the state cookie is no longer needed
+                removeCookie(context, STATE_COOKIE_NAME);
+            } else {
+                // Local redirect restoring the original request path, the state cookie is no longer needed
+                removeCookie(context, STATE_COOKIE_NAME);
+            }
+        } else {
+            // State cookie must be available to minimize the risk of CSRF
+            cf.completeExceptionally(new AuthenticationFailedException());
+            return cf;
+        }
+
         params.put("code", code);
-        params.put("redirect_uri", buildRedirectUri(context));
+        params.put("redirect_uri", buildCodeRedirectUri(context, absoluteUri, getDynamicPath(context, absoluteUri)));
 
         auth.authenticate(params, userAsyncResult -> {
             if (userAsyncResult.failed()) {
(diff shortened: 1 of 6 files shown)


sberyozkin:
CC @stianst @pedroigor @stuartwdouglas

sberyozkin:
note I've updated the test `application.properties` with `quarkus.oidc.authentication.redirect-path=/web-app/refresh` - this does not affect any of the tests and tests that unless this value is a subset of the actual request raw path then nothing is affected.
If I set this value to `quarkus.oidc.authentication.redirect-path=/web-app` then the newly added test fails because the `reroute` code from `web-app` to `web-app/refresh` does not seem to work

extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/CodeAuthenticationMechanism.java
```diff
@@ -105,9 +110,27 @@ private static QuarkusSecurityIdentity augmentIdentity(SecurityIdentity security
...
+        Cookie stateCookie = removeCookie(context, STATE_COOKIE_NAME);
+        if (stateCookie != null) {
+            List<String> values = context.queryParam("state");
+            if (values == null || values.size() != 1 || !stateCookie.getValue().startsWith(values.get(0))) {
+                cf.completeExceptionally(new AuthenticationFailedException());
+                return cf;
+            } else {
+                String extraPath = stateCookie.getValue().split(COOKIE_DELIM)[1];
+                // This code does not work
+                context.reroute(absoluteUri.getRawPath() + extraPath);
```
pedroigor: Isn't enough to send a redirect ?

extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/CodeAuthenticationMechanism.java
```diff
@@ -89,8 +88,14 @@ private static QuarkusSecurityIdentity augmentIdentity(SecurityIdentity security
...
         params.put("scopes", new JsonArray(scopes));
-        params.put("redirect_uri", buildRedirectUri(context));
-        params.put("state", generateState(context));
+
+        URI absoluteUri = URI.create(context.request().absoluteURI());
+        String dynamicPath = getDynamicPath(context, absoluteUri);
+        params.put("redirect_uri", buildRedirectUri(context, absoluteUri, dynamicPath));
+
+        if (dynamicPath != null) {
+            params.put("state", generateState(context, dynamicPath));
```
pedroigor: I think we should always send the state parameter. It provides a barrier to CSRF.

extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/CodeAuthenticationMechanism.java
```diff
@@ -129,46 +152,63 @@ private static QuarkusSecurityIdentity augmentIdentity(SecurityIdentity security
...
+            if (requestPath.startsWith(redirectPath) && requestPath.length() > redirectPath.length()) {
+                return requestPath.substring(redirectPath.length());
+            }
+        }
+        return null;
+    }
+
+    private String generateState(RoutingContext context, String dynamicPath) {
+        String uuid = UUID.randomUUID().toString();
+        String cookieValue = uuid + COOKIE_DELIM + dynamicPath;
```
pedroigor: You could probably reuse the state parameter to bind the callback uri.

extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/CodeAuthenticationMechanism.java
```diff
@@ -129,46 +152,63 @@ private static QuarkusSecurityIdentity augmentIdentity(SecurityIdentity security
...
+
+        CookieImpl cookie = new CookieImpl(STATE_COOKIE_NAME, cookieValue);
 
         cookie.setHttpOnly(true);
         cookie.setSecure(context.request().isSSL());
-        cookie.setMaxAge(-1);
+        cookie.setMaxAge(60 * 5);
```
pedroigor: Is this really necessary given that we always reset the cookie when doing challenges? Assuming that the user may take more time than usual to log in at the AS, the cookie will be automatically removed and the redirect will fail.

extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/CodeAuthenticationMechanism.java
pedroigor: In fact, a future improvement here would be to support PKCE. Just a side comment :)

gastaldi:
@sberyozkin are you sure that this PR fixes #5773 ? The changes seem unrelated

extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/CodeAuthenticationMechanism.java
sberyozkin: @pedroigor yeah, that should be enough actually, will try :-)

extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/CodeAuthenticationMechanism.java
sberyozkin: @pedroigor You are right, I'll just keep a check that compares it with the the cookie on the way back

extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/CodeAuthenticationMechanism.java
sberyozkin: @pedroigor yeah, I thought about it but if it goes in the URL it be encoded so it may end up being double encoded and as such we can lose the original percent encoding if any in the original path, hence I added it to the cookie. Going forward I guess we can expect the cookie become a (secured) container of various properties...

extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/CodeAuthenticationMechanism.java
sberyozkin: @pedroigor I updated it because when a `security-oidc-web-app-authentication-quickstart` is built a lot of warnings is displayed about an invalid `max-age=-1` cookie parameter :-). I set to to 5 mins, but we can set it to 30 mins, that should be enough, what do you think ?

sberyozkin:
@gastaldi sorry, fixed it

extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/CodeAuthenticationMechanism.java
stianst: PKCE is aimed at public clients. Confidential clients should use state and nonce though.

stianst:
This should be the default behaviour, and not something that requires redirect-path option to enable.

sberyozkin:
@stianst I tried to approach it this way originally, I did not even had this property, and wrote the code which would settle on the 1st path segment and save the remaining part in the state cookie as per my comment in the issue, but it proved impossible to fix generically, as there could be the static resources sitting on the URI paths which are not overlapping with the JAX-RS paths.

For example, a request may start with an access to a protected resource `http://localhost:8081/pages/index.html` (I've found it while trying to come up with a test) and the JAX-RS endpoint paths are starting from `/web-app`. My code would set a `redirect_uri` to `http://localhost:8081/pages` but then on the way back a 404 is reported before the OIDC adapter gets a chance to restore a complete path.
It did work for the existing tests starting from a `http://localhost:8081/index.html` request but only because on the way back Quarkus would resolve `http://localhost:8081/` to the root index.html but it is a special case.

extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/CodeAuthenticationMechanism.java
pedroigor: It should be fine, maybe we can also add an additional step after successfully obtaining the ID Token and remove the cookie ?

extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/CodeAuthenticationMechanism.java
pedroigor: My point is you can still have `web-app` type applications as public (although much more secure would be if people just configure their clients as confidential).

extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/CodeAuthenticationMechanism.java
sberyozkin: @pedroigor yes, it is removed in this PR, but before the code step is started but I'll tweak it

extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/CodeAuthenticationMechanism.java
sberyozkin: done

extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/CodeAuthenticationMechanism.java
sberyozkin: State cookie is always used now

extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/CodeAuthenticationMechanism.java
sberyozkin: It is now removed only after the local redirect

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/security/HttpSecurityRecorder.java
```diff
@@ -47,6 +48,11 @@ public void run() {
                                         event.response().end();
                                     }
                                 });
+                            } else if (throwable instanceof AuthenticationRedirectException) {
```
stuartwdouglas: I don't think this is correct. As I understand it this is basically a challenge, so it should be send as part of sendChallenge()? 

Otherwise you loose the ability to have unprotected resources, as this will attempt to log in on every request even if the page is supposed to be public.

stuartwdouglas (changes requested):

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/security/HttpSecurityRecorder.java
sberyozkin: @stuartwdouglas This exception may only be thrown on the return leg of the OIDC code flow, not when the user is initially trying to access a resource. (Ex, user accesses a protected resource, user is redirected, with some of the original path saved in the cookie, user is redirected back - and now if the original path needs to be restored the user will be redirect to it)
FYI, `CodeFlowTest` test accessing an `UnprotectedResource` which you added passes with this PR.

stuartwdouglas: ah, ok. In that case I think it is ok.

sberyozkin:
@pedroigor, @stianst Hi, I'm hoping to merge it soon for it to make it to 1.1.0 now that Stuart @stuartwdouglas 's concern about the unprotected resources has been resolved :-), if you have some other concerns then let me know today

stuartwdouglas (approved):

Issue #4037, opened 2019-09-16
Introduce support for Spring Security


geoand:
### Description
Decide on what features from Spring security should be supported then implement the important ones for this Epic.

### Analysis
(links to analysis docs containing architecture design work, requirements gathering, etc)

### Tasks
- [x] Support Spring Security's `@Secured` (#5225)
- [x] Support Spring Security's `@PreAuthorize` (open PR: #5787)

hantsy:
Currently there are some issues I found when copying a very simple Spring sample to Quarkus. I think port Spring in Quarkus is a lot of work. 
 
Instead of introducing Spring Security I hope there is a more generic programmatic approach to configure Security rules by matching HTTP url/method etc, like the configuration in Spring Security.

geoand:
Quarkus is going to introduce a native security solution soon. 
Any work we do towards this ticket will be based on that.

ebullient:
Spring 5 Security is pretty tidy now.. we really should figure out how to pull this in, as Spring applications will tend to use it.

ppalaga:
Two people recently asked me about Spring Security on Quarkus on a Java developer conference. (And I got no questions about other missing Spring components).

geoand:
@ppalaga thanks for the info!

geoand:
Has been implemented

PR #5787, opened 2019-11-26
Introduce support for the most common features of Spring Security's @PreAuthorize

geoand:
Docs will follow and be based on what @aureamunoz is already preparing

Relates to: #4037

### DIFF
--- a/extensions/spring-di/deployment/src/main/java/io/quarkus/spring/di/deployment/SpringBeanNameToDotNameBuildItem.java
+++ b/extensions/spring-di/deployment/src/main/java/io/quarkus/spring/di/deployment/SpringBeanNameToDotNameBuildItem.java
@@ -1,0 +1,24 @@
+package io.quarkus.spring.di.deployment;
+
+import java.util.Map;
+
+import org.jboss.jandex.DotName;
+
+import io.quarkus.builder.item.SimpleBuildItem;
+
+/**
+ * The purpose of this bean is to map the names of the Spring Beans to their associated DotName
+ * This info is needed when trying to convert SpEL expressions that reference beans by name, to bytecode
+ */
+public final class SpringBeanNameToDotNameBuildItem extends SimpleBuildItem {
+
+    private final Map<String, DotName> map;
+
+    public SpringBeanNameToDotNameBuildItem(Map<String, DotName> map) {
+        this.map = map;
+    }
+
+    public Map<String, DotName> getMap() {
+        return map;
+    }
+}
--- a/extensions/spring-security/deployment/src/main/java/io/quarkus/spring/security/deployment/SpringSecurityProcessor.java
+++ b/extensions/spring-security/deployment/src/main/java/io/quarkus/spring/security/deployment/SpringSecurityProcessor.java
@@ -4,27 +4,64 @@ import static io.quarkus.security.deployment.SecurityTransformerUtils.findFirstS
 import static io.quarkus.security.deployment.SecurityTransformerUtils.hasStandardSecurityAnnotation;
 
 import java.lang.reflect.Modifier;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
 import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
+import java.util.function.Function;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import org.jboss.jandex.AnnotationInstance;
 import org.jboss.jandex.AnnotationTarget;
+import org.jboss.jandex.AnnotationValue;
 import org.jboss.jandex.ClassInfo;
+import org.jboss.jandex.DotName;
+import org.jboss.jandex.FieldInfo;
 import org.jboss.jandex.MethodInfo;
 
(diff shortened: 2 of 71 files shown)


geoand:
@michalszynkiewicz can you please review the CDI related logic?

@gsmet would you like to go over the basics to see if I have missed anything obvious? I don't expect you to go over the actual code since it's quite involved :)

extensions/spring-di/deployment/src/main/java/io/quarkus/spring/di/deployment/SpringBeanNameToDotNameBuildItem.java
```diff
@@ -0,0 +1,24 @@
...
+
+import java.util.Map;
+
+import org.jboss.jandex.DotName;
+
+import io.quarkus.builder.item.SimpleBuildItem;
+
+/**
+ * The purpose of this bean is to map the names of the Spring Beans to their associated DotName
+ * This info is needed when trying to some SpEL expressions to bytecode
```
gsmet: There is a word missing.

extensions/spring-security/deployment/src/main/java/io/quarkus/spring/security/deployment/SpringSecurityProcessor.java
```diff
@@ -4,28 +4,64 @@
...
+import io.quarkus.spring.di.deployment.SpringBeanNameToDotNameBuildItem;
+import io.quarkus.spring.security.deployment.roles.HasRoleValueProducer;
+import io.quarkus.spring.security.runtime.interceptor.SpringPreauthorizeInterceptor;
 import io.quarkus.spring.security.runtime.interceptor.SpringSecuredInterceptor;
 
 class SpringSecurityProcessor {
 
+    private static final String PARAMETER_EQ_PRINCIPAL_USERNAME_REGEX = "#(\\w+)(\\.(\\w+))?\\s+==\\s+(authentication.)?principal.username";
+    private static final Pattern PARAMETER_EQ_PRINCIPAL_USERNAME_PATTERN = Pattern
+            .compile(PARAMETER_EQ_PRINCIPAL_USERNAME_REGEX);
```
gsmet: You don't support `!=`?

gsmet (changes requested):
I added a few minor comment/questions.

Tests look good so it should be pretty safe.

gsmet:
By the way, additional question: do you support precedence via parenthesis, IIRC, that's something we used in our conditions.

extensions/spring-security/deployment/src/main/java/io/quarkus/spring/security/deployment/SpringSecurityProcessor.java
geoand: I didn't think of that use case at all. Is that something you've seen in the wild?

geoand:
In this iteration there is precedence support I am afraid. I really wanted to get the basics out there and see what people think and then try to work out what people would like and use the most.
That said, if you have a good and easy way to deal with parentheses and precedence that doesn't require me to learn ANTLR, I'm all ears 😁

extensions/spring-security/deployment/src/main/java/io/quarkus/spring/security/deployment/SpringSecurityProcessor.java
gsmet: Yes, I'm pretty sure I already used it. If it's not too complicated, I would say let's do it. But it can come in a follow-up if you prefer.

gsmet:
Couldn't we use an EL implementation somehow instead of doing dark magic? That would be taken care of I suppose.

That being said, I'm for merging this. But you can be sure you will need proper precedence support as that's very important for conditions.

I'm totally positive we used that in our applications at $previousJob.

geoand:
I really don't want to pull in SpEL because who knows what it will take to get it working on native and how many classes it depends on.
I agree that we will definitely need proper precedence and similar features.
I'll ask @mkouba how he implemented those in Qute 😉

extensions/spring-di/deployment/src/main/java/io/quarkus/spring/di/deployment/SpringBeanNameToDotNameBuildItem.java
geoand: Fixed

extensions/spring-security/deployment/src/main/java/io/quarkus/spring/security/deployment/SpringSecurityProcessor.java
geoand: Nah, I'll do it tomorrow most likely and get it out of the way :)

geoand: Done

gsmet (approved):

gsmet:
@geoand I merged this one, thanks. I also merged @aureamunoz 's work on the Spring Security guide so you should be able to work on the documentation.

gsmet:
BTW, will your parser fail if there is some precedence rules defined in an expression? I would rather have it fail than having it do incorrect things.

geoand:
@gsmet Thanks!

Yes it will currently fail instead of providing incorrect results.

I will work on the documentation next week

PR #5788, opened 2019-11-26
Remove the nimbus test dependency

sberyozkin:
This PR removes a Nimbus test dependency. Our own JWT generation API is in the works, but for now this utility code needs an update


gastaldi:
I don't see the dependency being removed in the pom.xml? Is that right?

gastaldi (approved):

sberyozkin:
@gastaldi Thanks, you are super fast :-) my fault, I was 'forced' to close my laptop immediately after I opened this PR yesterday so missed your comment yesterday.

PR #5801, opened 2019-11-27
Launching command line should be copy/paste friendly

gastaldi:
It also performs a minor refactor on how ProcessBuilder is used

### DIFF
--- a/devtools/maven/src/main/java/io/quarkus/maven/DevMojo.java
+++ b/devtools/maven/src/main/java/io/quarkus/maven/DevMojo.java
@@ -1,5 +1,7 @@
 package io.quarkus.maven;
 
+import static java.util.stream.Collectors.joining;
+
 import java.io.ByteArrayOutputStream;
 import java.io.DataOutputStream;
 import java.io.File;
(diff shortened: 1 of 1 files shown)


devtools/maven/src/main/java/io/quarkus/maven/DevMojo.java
```diff
@@ -617,14 +619,11 @@ void prepare() throws Exception {
...
             // Display the launch command line in dev mode
-            getLog().info("Launching JVM with command line: " + args.toString());
-            ProcessBuilder pb = new ProcessBuilder(args.toArray(new String[0]));
-            pb.redirectError(ProcessBuilder.Redirect.INHERIT);
-            pb.redirectOutput(ProcessBuilder.Redirect.INHERIT);
-            pb.redirectInput(ProcessBuilder.Redirect.INHERIT);
-            pb.directory(workingDir);
-            process = pb.start();
-
+            getLog().info("Launching JVM with command line: " + args.stream().collect(joining(" ")));
```
gastaldi: Actually, I think this should be using the debug level instead. It doesn't seem to be a really useful information to be displayed everytime

jaikiran: Hello @gastaldi, this was intentionally changed to `INFO` as part of https://github.com/quarkusio/quarkus/pull/5366 to help quickly notice any obvious issues for cases noted in that PR.

gastaldi: Okay, I reverted my commit that undo'ed that. Thanks for the pointer :)

stuartwdouglas (approved):

PR #5803, opened 2019-11-27
Fix various usability issues in Spring @RestControllerAdvice handling

### LINKED ISSUE #5796: @RestControllerAdvice that returns an object but handles errors from a void controller method does not return json
A simple way to reproduce this error is to change the existing spring-web integration to use : 

```
    @GetMapping("/responseEntity")
    public Greeting handledByResponseEntity() {
        throw new IllegalStateException("bad state");
    }
```

to:

```
    @GetMapping("/responseEntity")
    public void handledByResponseEntity() {
        throw new IllegalStateException("bad state");
    }
``` 

in the `ExceptionThrowingController` class

geoand:
Fixes: #5796

### DIFF
--- a/extensions/spring-web/deployment/src/main/java/io/quarkus/spring/web/deployment/ControllerAdviceAbstractExceptionMapperGenerator.java
+++ b/extensions/spring-web/deployment/src/main/java/io/quarkus/spring/web/deployment/ControllerAdviceAbstractExceptionMapperGenerator.java
@@ -9,6 +9,7 @@ import java.util.Map;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 import javax.ws.rs.core.Context;
+import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
 
 import org.jboss.jandex.AnnotationInstance;
@@ -31,6 +32,7 @@ import io.quarkus.spring.web.runtime.ResponseEntityConverter;
 class ControllerAdviceAbstractExceptionMapperGenerator extends AbstractExceptionMapperGenerator {
 
     private static final DotName RESPONSE_ENTITY = DotName.createSimple("org.springframework.http.ResponseEntity");
+    private static final DotName STRING = DotName.createSimple(String.class.getName());
 
     private final MethodInfo controllerAdviceMethod;
     private final TypesUtil typesUtil;
@@ -135,17 +137,39 @@ class ControllerAdviceAbstractExceptionMapperGenerator extends AbstractException
 
             ResultHandle response;
             if (RESPONSE_ENTITY.equals(returnType.name())) {
+                /*
+                 * By default we will send JSON back unless the ResponseEntity has a String body
+                 */
+                boolean addDefaultJsonContentType = true;
+                if (returnType.kind() == Type.Kind.PARAMETERIZED_TYPE) {
+                    if (returnType.asParameterizedType().arguments().size() == 1) {
+                        Type responseEntityParameterType = returnType.asParameterizedType().arguments().get(0);
+                        if (STRING.equals(responseEntityParameterType.name())) {
+                            addDefaultJsonContentType = false;
+                        }
+                    }
+                }
+
                 // convert Spring's ResponseEntity to JAX-RS Response
                 response = toResponse.invokeStaticMethod(
                         MethodDescriptor.ofMethod(ResponseEntityConverter.class.getName(), "toResponse",
-                                Response.class.getName(), RESPONSE_ENTITY.toString()),
-                        exceptionHandlerMethodResponse);
+                                Response.class.getName(), RESPONSE_ENTITY.toString(), boolean.class.getName()),
+                        exceptionHandlerMethodResponse, toResponse.load(addDefaultJsonContentType));
             } else {
                 ResultHandle status = toResponse.load(getStatus(controllerAdviceMethod.annotation(RESPONSE_STATUS)));
 
                 ResultHandle responseBuilder = toResponse.invokeStaticMethod(
                         MethodDescriptor.ofMethod(Response.class, "status", Response.ResponseBuilder.class, int.class),
                         status);
+                /*
+                 * By default we will send JSON back unless the ResponseEntity has a String body
+                 */
+                if (!STRING.equals(returnType.name())) {
+                    toResponse.invokeVirtualMethod(
+                            MethodDescriptor.ofMethod(Response.ResponseBuilder.class, "type", Response.ResponseBuilder.class,
+                                    String.class),
+                            responseBuilder, toResponse.load(MediaType.APPLICATION_JSON));
+                }
 
                 toResponse.invokeVirtualMethod(
                         MethodDescriptor.ofMethod(Response.ResponseBuilder.class, "entity", Response.ResponseBuilder.class,
--- a/extensions/spring-web/deployment/src/main/java/io/quarkus/spring/web/deployment/ControllerAdviceAbstractExceptionMapperGenerator.java
+++ b/extensions/spring-web/deployment/src/main/java/io/quarkus/spring/web/deployment/ControllerAdviceAbstractExceptionMapperGenerator.java
@@ -214,9 +214,7 @@ class ControllerAdviceAbstractExceptionMapperGenerator extends AbstractException
     }
 
     private ResultHandle controllerAdviceInstance(MethodCreator toResponse) {
-        ResultHandle controllerAdviceClass = toResponse.invokeStaticMethod(
-                MethodDescriptor.ofMethod(Class.class, "forName", Class.class, String.class),
-                toResponse.load(declaringClassName));
+        ResultHandle controllerAdviceClass = toResponse.loadClass(declaringClassName);
 
         ResultHandle container = toResponse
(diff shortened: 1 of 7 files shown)


geoand:
@gsmet this should be easy to review commit by commit

extensions/spring-web/runtime/src/main/java/io/quarkus/spring/web/runtime/ResponseEntityConverter.java
```diff
@@ -22,7 +22,7 @@ public static Response toResponse(ResponseEntity responseEntity) {
     private static Headers<Object> toHeaders(HttpHeaders springHeaders) {
         Headers<Object> jaxRsHeaders = new Headers<>();
         for (Map.Entry<String, List<String>> entry : springHeaders.entrySet()) {
-            jaxRsHeaders.addAll(entry.getKey(), entry.getValue());
+            jaxRsHeaders.addAll(entry.getKey(), entry.getValue().toArray(new Object[0]));
```
gsmet: Maybe let's properly initialize the size of the array to avoid creating a new one?

extensions/spring-web/runtime/src/main/java/io/quarkus/spring/web/runtime/ResponseEntityConverter.java
```diff
@@ -4,26 +4,38 @@
...
+    private static Headers<Object> toHeaders(HttpHeaders springHeaders, boolean addJsonContentTypeIfNotSet) {
         Headers<Object> jaxRsHeaders = new Headers<>();
         for (Map.Entry<String, List<String>> entry : springHeaders.entrySet()) {
             jaxRsHeaders.addAll(entry.getKey(), entry.getValue().toArray(new Object[0]));
         }
+        /*
+         * We add the default application/json content type if no content type is specified
+         * since this is the default value when returning an object from a Spring RestController
+         */
+        if (!jaxRsHeaders.containsKey(javax.ws.rs.core.HttpHeaders.CONTENT_TYPE) && addJsonContentTypeIfNotSet) {
```
gsmet: It's very minor as it's probably the default case but I would revert the condition to avoid the map lookup when we can.

extensions/spring-web/deployment/src/main/java/io/quarkus/spring/web/deployment/ControllerAdviceAbstractExceptionMapperGenerator.java
```diff
@@ -135,17 +137,39 @@ void generateMethodBody(MethodCreator toResponse) {
...
             if (RESPONSE_ENTITY.equals(returnType.name())) {
+                /*
+                 * By default we will send JSON back unless the ResponseEntity has a String body
+                 */
+                boolean addDefaultJsonContentType = true;
+                if (returnType.kind() == Type.Kind.PARAMETERIZED_TYPE) {
+                    if (returnType.asParameterizedType().arguments().size() == 1) {
+                        Type responseEntityParameterType = returnType.asParameterizedType().arguments().get(0);
+                        if (STRING.equals(responseEntityParameterType.name())) {
+                            addDefaultJsonContentType = false;
```
gsmet: There's no way in Spring to explicitly define the response media type?

extensions/spring-web/deployment/src/main/java/io/quarkus/spring/web/deployment/SpringWebProcessor.java
```diff
@@ -338,7 +338,7 @@ public void generateExceptionMapperProviders(BeanArchiveIndexBuildItem beanArchi
         // Look for all exception classes that are annotated with @ResponseStatus
 
         IndexView index = beanArchiveIndexBuildItem.getIndex();
-        ClassOutput classOutput = new GeneratedClassGizmoAdaptor(generatedExceptionMappers, false);
+        ClassOutput classOutput = new GeneratedClassGizmoAdaptor(generatedExceptionMappers, true);
```
gsmet: Oops :).

gsmet (changes requested):
I added a few very minor comments and a question.

Feel free to ignore some of them if you don't feel like it but I'd like an answer to my question :).

extensions/spring-web/runtime/src/main/java/io/quarkus/spring/web/runtime/ResponseEntityConverter.java
geoand: Makes sense

extensions/spring-web/deployment/src/main/java/io/quarkus/spring/web/deployment/SpringWebProcessor.java
geoand: Yeah, that cost me a good hour or so to figure out... Paying for my own mistakes :P

extensions/spring-web/deployment/src/main/java/io/quarkus/spring/web/deployment/ControllerAdviceAbstractExceptionMapperGenerator.java
geoand: For the regular controller method sure there is and we handle it. But for error handler you either return an object and rely on the default or return a `ResponseEntity` where you can set whatever you like.

extensions/spring-web/runtime/src/main/java/io/quarkus/spring/web/runtime/ResponseEntityConverter.java
geoand: Not sure I follow. We just need to convert a list to an array and according to http://shipilev.net/blog/2016/arrays-wisdom-ancients/#_conclusion this is the proper way to do it.
Am I missing something?

extensions/spring-web/runtime/src/main/java/io/quarkus/spring/web/runtime/ResponseEntityConverter.java
geoand: Fixed

extensions/spring-web/deployment/src/main/java/io/quarkus/spring/web/deployment/ControllerAdviceAbstractExceptionMapperGenerator.java
gsmet: Are you sure about that? Wouldn't it use the content type defined in the `@RequestMapping` annotation of the original method?

And just to be clear, we don't support overriding the content-type in the `ResponseEntity` response of the exception mapper?

extensions/spring-web/runtime/src/main/java/io/quarkus/spring/web/runtime/ResponseEntityConverter.java
gsmet: I don't know. The List javadoc clearly pushes towards using the size but I suppose it's not critical enough here to have us waste our time discussing it :).

extensions/spring-web/deployment/src/main/java/io/quarkus/spring/web/deployment/ControllerAdviceAbstractExceptionMapperGenerator.java
geoand: > Are you sure about that? Wouldn't it use the content type defined in the `@RequestMapping` annotation of the original method?
> 

Yes it would, and that is what causes the problem, since the controller method uses a type that doesn't match what the exception handler returns.

> And just to be clear, we don't support overriding the content-type in the `ResponseEntity` response of the exception mapper?

That's part of the responsibility of the application code. If the application sets it, we don't change it: https://github.com/quarkusio/quarkus/blob/63673a7189713c822366b518550248778ab4a66d/extensions/spring-web/runtime/src/main/java/io/quarkus/spring/web/runtime/ResponseEntityConverter.java#L36

extensions/spring-web/runtime/src/main/java/io/quarkus/spring/web/runtime/ResponseEntityConverter.java
geoand: Yeah, this isn't on a critical path anyway so it doesn't really matter.

extensions/spring-web/deployment/src/main/java/io/quarkus/spring/web/deployment/ControllerAdviceAbstractExceptionMapperGenerator.java
gsmet: We discussed this on Zulip with @geoand. Our current support is a bit limited as we will always return JSON (except in some edge cases), even if the original method was pushing some XML.

That's not very user friendly and we will probably need to improve that. But that can probably wait so we decided to merge that one first.

gsmet (approved):

extensions/spring-web/deployment/src/main/java/io/quarkus/spring/web/deployment/ControllerAdviceAbstractExceptionMapperGenerator.java
geoand: As discussed I will research this a bit more and open an issue for a follow-up implementation

geoand:
@gsmet all commits from the PR can be back-ported except: 324a213be6f34651182f9fd69f92070a3a0ce835

PR #5805, opened 2019-11-27
Fixes the extension author guide WRT the new quarkus-extension.yaml

loicmathieu:


docs/src/main/asciidoc/writing-extensions.adoc
```diff
@@ -378,7 +378,7 @@ The above sequence of commands does the following:
 A Maven build performed immediately after generating the modules should fail due to a `fail()` assertion in one of the test classes.
 
 There is one step (specific to the Quarkus source tree) that you should do manually when creating a new extension:
-Add the extension coordinates to `devtools/common/src/main/filtered/extensions.json`.
+Create a `quarkus-extension.yaml` file that describe your extension inside the runtime module `src/main/resources/META-INF` folder.
```
gsmet: So the error was here before but `Create` should be `create` after a colon. Can you fix that while you're at it?

I also think an example of an existing extension that has all the right information would be helpful.

gsmet (changes requested):
Thanks for catching that!

I added a comment inline.

docs/src/main/asciidoc/writing-extensions.adoc
loicmathieu: done, I add an example from Agroal

gsmet (approved):

gsmet:
Thanks!

PR #5834, opened 2019-11-28
Set current/explicit Keycloak version in docs examples

rsvoboda:
Set current/explicit Keycloak version in docs examples

Similar fix happened in QS - https://github.com/quarkusio/quarkus-quickstarts/pull/387
I was playing with Keycloak 8 and hit issues in native mode.

### DIFF
--- a/docs/pom.xml
+++ b/docs/pom.xml
@@ -97,6 +97,7 @@
                         <surefire-version>${version.surefire.plugin}</surefire-version>
                         <graalvm-version>${graal-sdk.version-for-documentation}</graalvm-version>
                         <restassured-version>${rest-assured.version}</restassured-version>
+                        <keycloak-docker-image>${keycloak.docker.image}</keycloak-docker-image>
                         <!-- Project website home page -->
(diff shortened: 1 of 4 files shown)


sberyozkin (approved):

gsmet (changes requested):
There is a build property in the parent pom, we should inject it in the doc and use it.

See https://github.com/quarkusio/quarkus/blob/master/build-parent/pom.xml#L82 .

rsvoboda:
@gsmet can you point me to an example ? Something like `${keycloak.docker.image}` ?

rsvoboda:
Looking at asciidoctor-maven-plugin config and following `graalvm-version` way

gsmet:
Yeah sorry. We already do that for a couple of attributes, you just need to do exactly the same!

You might have to use the following if the variable is used in a source block:
```
[source,subs=attributes+]
```
This enables variable substitution in a source block.

Ping me on Zulip if you have any further issue!

rsvoboda:
@gsmet PR updated, already figured out my way

gsmet (approved):

Issue #5821, opened 2019-11-28
RuntimeClassLoader.findApplicationResource() does not handle the path name correctly 


mkouba:
This method is used in `getResource()` and `getResourceAsStream()` but it does not convert the path name in any way and so if you use the `/` separator in the path (as defined in those methods) it fails on windows machines because it simply walks the app directories and resolves the name against them (but system-dependent separator is used).

mkouba:
@stuartwdouglas am I right?

stuartwdouglas:
yes, but all this is going away soon, so I don't know if it is worth spending time on it at this stage.

mkouba:
We can keep this issue open and verify/close it once the new CL model is in place...

mkouba:
Hm, on the other hand, it blocks my PR: https://github.com/quarkusio/quarkus/pull/5793. So if there is a quick fix I'd vote to fix it now.

jaikiran:
@mkouba I created https://github.com/quarkusio/quarkus/pull/5835. Would you be able to cherry-pick that one into your Qute extension PR and see if it helps?

mkouba:
@jaikiran Of course! Let's see how it goes and if the CI is green we can close https://github.com/quarkusio/quarkus/pull/5835 and merge it as part of the https://github.com/quarkusio/quarkus/pull/5793. Thanks!

jaikiran:
This got fixed as part of the recently merged https://github.com/quarkusio/quarkus/pull/5793. Closing.

PR #5835, opened 2019-11-28
Use filesystem specific name separator while trying to resolve a resource in RuntimeClassLoader 

jaikiran:
Potential fix for https://github.com/quarkusio/quarkus/issues/5821

I haven't done any actual testing of Windows and even on *nix, I've run minimal tests (`core/deployment` passed fine). Given that @mkouba had already narrowed it down to this method, I decided to just try out this change and see how CI responds.

### DIFF
--- a/core/deployment/src/main/java/io/quarkus/runner/RuntimeClassLoader.java
+++ b/core/deployment/src/main/java/io/quarkus/runner/RuntimeClassLoader.java
@@ -462,7 +462,16 @@ public class RuntimeClassLoader extends ClassLoader implements ClassOutput, Tran
         Path resourcePath = null;
 
         for (Path i : applicationClassDirectories) {
-            resourcePath = i.resolve(name);
+            // Resource names are always separated by the "/" character.
+            // Here we are trying to resolve those resources using a filesystem
+            // Path, so we replace the "/" character with the filesystem
+            // specific separator before resolving
+            String path = name;
+            final String pathSeparator = i.getFileSystem().getSeparator();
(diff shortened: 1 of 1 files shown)


stuartwdouglas:
@mkouba can you verify that this fixes your issue?

core/deployment/src/main/java/io/quarkus/runner/RuntimeClassLoader.java
```diff
@@ -462,7 +462,16 @@ private URL findApplicationResource(String name) {
         Path resourcePath = null;
 
         for (Path i : applicationClassDirectories) {
-            resourcePath = i.resolve(name);
+            // Resource names are always separated by the "/" character.
+            // Here we are trying to resolve those resources using a filesystem
+            // Path, so we replace the "/" character with the filesystem
+            // specific separator before resolving
+            String path = name;
+            final String pathSeparator = i.getFileSystem().getSeparator();
```
dmlloyd: This is a small nit but if you do this instead:

```java
if (File.separatorChar != '/') {
    path = name.replace('/', File.separatorChar);
}
```
then the code is eliminated on *NIX-like platforms.

jaikiran: Hello David, I'll actually close this PR. I had opened this to help with https://github.com/quarkusio/quarkus/pull/5793. @mkouba has already pulled this commit into that PR and even incorporated your review comment into that change.

PR #5847, opened 2019-11-28
Feature/add behaviours description

uaihebert:
add description for CORS filters and Spring PostConstruct.

### DIFF
--- a/docs/src/main/asciidoc/http-reference.adoc
+++ b/docs/src/main/asciidoc/http-reference.adoc
@@ -153,6 +153,15 @@ quarkus.http.cors.exposed-headers=Content-Disposition
 quarkus.http.cors.access-control-max-age=24H
 ----
 
+[TIP]
+====
+When CORS functionality is activated your filters - e.g. ContainerRequestFilter or ContainerResponseFilter - will not be invoked.
+
+By default, a browser will make first the `preflight` request and after that the real request to your server. For the preflight requests, your filters will not be invoked only Quarkus internal filters.
+
+That is why you will not see any custom header that is defined in your project, only those defined in quarkus.http.xxx configurations.
+====
+
 == HTTP Limits Configuration
 
 The following properties are supported.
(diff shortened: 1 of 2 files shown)


docs/src/main/asciidoc/http-reference.adoc
```diff
@@ -153,6 +153,15 @@ quarkus.http.cors.exposed-headers=Content-Disposition
 quarkus.http.cors.access-control-max-age=24H
 ----
 
+[TIP]
+====
+When CORS functionality is activated your filters - e.g. ContainerRequestFilter or ContainerResponseFilter - will not be invoked.
```
gsmet: I'm not sure it's precise enough.

I suppose they are not enabled for the CORS requests but they are still enabled for standard requests? If not, I don't think this is expected.

gsmet (changes requested):
Asked a question we need to sort out before thinking about merging.

I also think it should be rewritten to be:
- more precise for the concerned requests
- more general for the consequences: a filter can do a lot more things than adding headers.

gsmet:
BTW, sorry, I'm in serial review mode but thanks for taking the time to contribute :).

docs/src/main/asciidoc/http-reference.adoc
uaihebert: What I experienced with quarkus is. 
When there are no CORS your filters will be invoked for every request.

When you got CORS your filters are not used for CORS request. That is what I am trying to explain here.

uaihebert:
no problem at all. I just want to make it clear some behaviors that were not described.

uaihebert:
> Asked a question we need to sort out before thinking about merging.
> 
> I also think it should be rewritten to be:
> 
> * more precise for the concerned requests
> * more general for the consequences: a filter can do a lot more things than adding headers.

Indeed can do more than add headers, I just added the headers as an example. A lot of companies requires that some header should always be present.

docs/src/main/asciidoc/http-reference.adoc
uaihebert: ps. I am just trying first to detail what I am trying to cover to hear from you. I will improve the text later! :D

geoand:
@gsmet when your concerns have been addressed please let me know so I can give this a go as well.
Thanks

gsmet:
Closing for now. Please open another PR taking the comments into account if you wish to.

Thanks.

PR #5850, opened 2019-11-28
Fix compilation error

gastaldi:
This fixes the error introduced by https://github.com/quarkusio/quarkus/commit/8f439467bb504112a0b34f3f377d690d9ecb872c

### DIFF
--- a/independent-projects/tools/common/src/main/java/io/quarkus/generators/rest/BasicRestProjectGenerator.java
+++ b/independent-projects/tools/common/src/main/java/io/quarkus/generators/rest/BasicRestProjectGenerator.java
@@ -171,10 +171,10 @@ public class BasicRestProjectGenerator implements ProjectGenerator {
             BuildTool buildTool = getBuildTool();
             switch (buildTool) {
                 case MAVEN:
-                    generate("templates/README.maven.ftl", context, readme, "read me");
+                    generate("templates/README.maven.ftl", invocation, readme,"read me");
                     break;
                 case GRADLE:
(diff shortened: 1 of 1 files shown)


geoand (approved):
Works for me, thanks!

independent-projects/tools/common/src/main/java/io/quarkus/generators/rest/BasicRestProjectGenerator.java
```diff
@@ -171,10 +171,10 @@ private void createReadme() throws IOException {
             BuildTool buildTool = getBuildTool();
             switch (buildTool) {
                 case MAVEN:
-                    generate("templates/README.maven.ftl", context, readme, "read me");
```
stuartwdouglas: Will this whitespace change break the formatter?

gastaldi: I ran the full build before submitting, so I guess it won't

machi1990 (approved):

independent-projects/tools/common/src/main/java/io/quarkus/generators/rest/BasicRestProjectGenerator.java
stuartwdouglas: Apparently not because it is in independent-projects

geoand: I executed a sample `mvn clean install` and the source wasn't changed so I guess the formatter didn't have any issues

PR #5853, opened 2019-11-28
Check if the user ran compile and otherwise run it for them

### LINKED ISSUE #5832: changed application.properties not taken into account starting devmode
`mvn quarkus:devmode` picks up .java changes but does not pick up changes to `application.properties` for that I need to explicitly do `mvn compile quarkus:devmode`.

I would expect it to be picked up if .java changes are OR that all "compilable" source changes will be ignored by `quarkus:devmode`

stuartwdouglas:
Fixes #5832

### DIFF
--- a/devtools/maven/src/main/java/io/quarkus/maven/DevMojo.java
+++ b/devtools/maven/src/main/java/io/quarkus/maven/DevMojo.java
@@ -71,6 +71,30 @@ import io.quarkus.utilities.JavaBinFinder;
  */
 @Mojo(name = "dev", defaultPhase = LifecyclePhase.PREPARE_PACKAGE, requiresDependencyResolution = ResolutionScope.COMPILE_PLUS_RUNTIME)
 public class DevMojo extends AbstractMojo {
+
+    /**
+     * running any one of these phases means the compile phase will have been run, if these have
+     * not been run we manually run compile
+     */
+    private static final Set<String> POST_COMPILE_PHASES = new HashSet<>(Arrays.asList(
+            "compile",
+            "process-classes",
+            "generate-test-sources",
+            "process-test-sources",
+            "generate-test-resources",
+            "process-test-resources",
+            "test-compile",
+            "process-test-classes",
+            "test",
+            "prepare-package",
+            "package",
+            "pre-integration-test",
+            "integration-test",
+            "post-integration-test",
+            "verify",
+            "install",
+            "deploy"));
+
     /**
      * The directory for compiled classes.
      */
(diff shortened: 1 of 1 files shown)


gastaldi (approved):

stuartwdouglas:
Maybe we should also change the docs to leave off compile?

gastaldi:
Yeah, that would be a good thing. Although I haven't been able to reproduce the error Max is seeing even without this PR

stuartwdouglas:
I think the steps are:

- new project
- add application.properties, make change
- mvn quarkus:dev, compile is run because first build, changes are picked up
- stop
- make changes to application.properties
- mvn quarkus:dev, change is not picked up, as compile has not copied the resource

devtools/maven/src/main/java/io/quarkus/maven/DevMojo.java
```diff
@@ -225,8 +249,20 @@ public void execute() throws MojoFailureException, MojoExecutionException {
         if (!sourceDir.isDirectory()) {
             getLog().warn("The project's sources directory does not exist " + sourceDir);
         }
+        //we check to see if there was a compile (or later) goal before this plugin
```
gastaldi: I think you only need to check if `compile` was executed? Not sure if you can `install` or `deploy` without the `compile` phase is invoked

gastaldi:
> I think the steps are:
> 
> * new project
> * add application.properties, make change
> * mvn quarkus:dev, compile is run because first build, changes are picked up
> * stop
> * make changes to application.properties
> * mvn quarkus:dev, change is not picked up, as compile has not copied the resource

I followed your exact steps and still works for me.

gastaldi:
PS: I am using VI, so it's unlikely that an IDE copied the file to the classes folder

devtools/maven/src/main/java/io/quarkus/maven/DevMojo.java
gastaldi: Just realized these are the goals set from the `mvn` command. Looks ok

machi1990:
> I think the steps are:
> 
> * new project
> * add application.properties, make change
> * mvn quarkus:dev, compile is run because first build, changes are picked up
> * stop
> * make changes to application.properties
> * mvn quarkus:dev, change is not picked up, as compile has not copied the resource

I guess this will work either way since we have:

https://github.com/quarkusio/quarkus/blob/d9bc53056c3647e627a2f21ea74a5355e3b1ebd1/devtools/maven/src/main/java/io/quarkus/maven/DevMojo.java#L231-L235 

@evanchooly added the auto-compile IIRC.  

Maybe we just need to change the docs and drop `compile` ?

stuartwdouglas:
The current code only auto compiles if target/classes does not exist. If they exist but are out of date  there will not be a compile.

gsmet (approved):
This one looks good to go IMHO.

I rebased it on top of master as it was a bit old. Let's wait for CI.

PR #5358, opened 2019-11-10
Upgrade GraalVM to 19.3.0

gwenneg:
This PR is related to #4218

It is incomplete and not meant to be merged for now. The build will most probably fail until the GraalVM 19.3.0 dependency is available and integrated into Quarkus.

PR #5854, opened 2019-11-29
Remove GraalVM internal class usage for JDK11 compatibility

### LINKED ISSUE #5682: Support GraalVM / native image builds using the GraalVM Java 11 build
**Description**
A follow up to #4218 which did the upgrade to GraalVM version 19.3 but focused only on the "JDK8 flavoured edition"

gwenneg:
Fixes #5682

### DIFF
--- a/core/deployment/src/main/java/io/quarkus/deployment/steps/NativeImageAutoFeatureStep.java
+++ b/core/deployment/src/main/java/io/quarkus/deployment/steps/NativeImageAutoFeatureStep.java
@@ -22,12 +22,11 @@ import java.util.Map;
 import java.util.Set;
 import java.util.jar.JarEntry;
 import java.util.jar.JarFile;
-import java.util.stream.Collectors;
 
 import org.graalvm.nativeimage.ImageSingletons;
 import org.graalvm.nativeimage.hosted.Feature;
+import org.graalvm.nativeimage.hosted.RuntimeClassInitialization;
 import org.graalvm.nativeimage.hosted.RuntimeReflection;
-import org.graalvm.nativeimage.impl.RuntimeClassInitializationSupport;
 
 import io.quarkus.deployment.annotations.BuildProducer;
 import io.quarkus.deployment.annotations.BuildStep;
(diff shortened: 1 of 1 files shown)


gwenneg:
This PR is a first step towards our GraalVM compatibility with JDK 11. I only had time to check it doesn't break the Quarkus build with JDK 8 last night, I will run the test with JDK 11 tonight.

It fixes the runtime initialization for sure, but the runtime **re**initialization will probably fail during the native image generation since the `RuntimeClassInitializationSupport` class is still used in our generated bytecode (but it was removed from the Quarkus sources).

One thing I've started investigating is the possibility to stop reinitializing classes at runtime since it was a hack (according to a comment in our sources) from the beginning and GraalVM deprecated the API option: https://github.com/oracle/graal/blob/e0cb8e79782f4d863180c9ce84519c634922cbfe/substratevm/src/com.oracle.svm.hosted/src/com/oracle/svm/hosted/classinitialization/ClassInitializationFeature.java#L124-L126. We only have a few cases where this "hack" is used in Quarkus.

gwenneg:
The current code of this PR seems to be enough to fix the GraalVM internal class usage in Quarkus, which is a requirement for the Quarkus sources compatible with GraalVM `19.3.0`, JDK 11 edition. It was successfully tested with several native builds including `quarkus-integration-test-kafka` which is one of the few builds reinitializing classes at runtime. The tests were run using both the `quay.io/quarkus/ubi-quarkus-native-image:19.3.0-java11` image and a local GraalVM `19.3.0 CE` installation.

Since it is not breaking anything in the `master` (JDK 8) CI, I think it is safe to be merged now even if the GraalVM `19.3.0` support is not ready yet (see #5358).

If you want to test Quarkus with GraalVM `19.3.0` and JDK 11 before this PR is merged, you will need to checkout #5358 and then cherry-pick the commit from this PR. I'll rebase #5358 on `master` as soon as this one is merged.

I didn't spend much time investigating the possibility to stop reinitializing classes at runtime, but that's still on my TODO list and will be treated in another PR if there's something to merge eventually.

cc @bobmcwhirter, @dmlloyd, @Sanne

dmlloyd (approved):

gwenneg:
Thanks for reviewing this @dmlloyd.

PR #5856, opened 2019-11-29
Fix exception mapper when root path is set

stuartwdouglas:


extensions/resteasy/runtime/src/main/java/io/quarkus/resteasy/runtime/NotFoundExceptionMapper.java
```diff
@@ -255,6 +260,17 @@ private Response respond(List<ResourceDescription> descriptions) {
...
 
+    private String adjustRoot(String basePath) {
+        if (httpRoot.equals("/")) {
+            //leave it alone
+            return basePath;
+        }
+        if (basePath.startsWith("/")) {
+            return httpRoot + basePath;
+        }
+        return httpRoot.substring(1) + "/" + basePath;
```
gsmet: I might be missing something obvious but why the substring in this very particular case?

stuartwdouglas: It's not obvious, for this case the template already has a '/' as it is expecting a non-slash prefixed path. As the httpRoot starts with '/' we need to remove it.

gsmet:
@machi1990 could you check that one with a few examples and report back?

Thanks!

machi1990:
> @machi1990 could you check that one with a few examples and report back?
> 
> Thanks!

Hey @gsmet 

I tried this one and  did not see any major issue other than when the root path ends with a slash, we end up with two consecutive slashes after concatenating the root path and base path e.g `/root-path/` and `base-path/` reads to `/root-path//base-path`. Likewise for the generated swagger.

machi1990 (approved):
LGTM.

PR #5858, opened 2019-11-29
Vert.x is resumed when using non-default paths

stuartwdouglas:
If the servlet context path is set then Undertow would be
registered using a RouteBuildItem instead of a DefaultRouteBuildItem.

This behaves differently in terms of resuming in the request,
so if both the context path was set and security was in use
the request body could disappear sometimes.

### DIFF
--- a/integration-tests/elytron-undertow/src/test/java/io/quarkus/it/undertow/elytron/BaseAuthTest.java
+++ b/integration-tests/elytron-undertow/src/test/java/io/quarkus/it/undertow/elytron/BaseAuthTest.java
@@ -3,6 +3,7 @@ package io.quarkus.it.undertow.elytron;
 import static io.restassured.RestAssured.given;
 import static org.hamcrest.Matchers.is;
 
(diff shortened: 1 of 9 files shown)


integration-tests/elytron-undertow/src/test/java/io/quarkus/it/undertow/elytron/BaseAuthTest.java
```diff
@@ -13,26 +13,30 @@
...
-                .contentType(ContentType.TEXT)
-                .when()
-                .post("/")
-                .then()
-                .statusCode(200)
-                .body(is("hello Bill"));
+        StringBuilder sb = new StringBuilder();
+
+        for (int i = 0; i < 100; ++i) {
+            // This is a regression test in that we had a problem where the Vert.x request was not paused
```
cescoffier: Should we use `@RepeatedTest(100)` instead?

cescoffier (approved):

PR #5861, opened 2019-11-29
Add possibility to override the logging configuration of each handler on category level

debae:
… on category level.

### DIFF
--- a/core/runtime/src/main/java/io/quarkus/runtime/logging/CategoryConfig.java
+++ b/core/runtime/src/main/java/io/quarkus/runtime/logging/CategoryConfig.java
@@ -1,5 +1,8 @@
 package io.quarkus.runtime.logging;
 
+import java.util.List;
+import java.util.Optional;
+
 import io.quarkus.runtime.annotations.ConfigGroup;
 import io.quarkus.runtime.annotations.ConfigItem;
 
@@ -11,4 +14,17 @@ public class CategoryConfig {
      */
     @ConfigItem(defaultValue = "inherit")
     String level;
+
+    /**
+     * The names of the handlers to link to this category.
+     */
+    @ConfigItem
+    Optional<List<String>> handlers;
+
+    /**
+     * Specify whether or not this logger should send its output to its parent Logger
+     */
+    @ConfigItem(defaultValue = "true")
+    boolean useParentHandlers;
+
 }
--- a/core/runtime/src/main/java/io/quarkus/runtime/logging/LoggingSetupRecorder.java
+++ b/core/runtime/src/main/java/io/quarkus/runtime/logging/LoggingSetupRecorder.java
@@ -6,6 +6,7 @@ import static org.wildfly.common.os.Process.getProcessName;
 import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
@@ -73,21 +74,15 @@ public class LoggingSetupRecorder {
         final Logger rootLogger = logContext.getLogger("");
 
         rootLogger.setLevel(config.level.orElse(Level.INFO));
-        for (Map.Entry<String, CategoryConfig> entry : categories.entrySet()) {
-            final String name = entry.getKey();
-            final Logger logger = logContext.getLogger(name);
-            final CategoryConfig categoryConfig = entry.getValue();
-            if (!"inherit".equals(categoryConfig.level)) {
-                logger.setLevelName(categoryConfig.level);
-            }
-        }
+
+        ErrorManager errorManager = new OnlyOnceErrorManager();
         final Map<String, CleanupFilterConfig> filters = config.filters;
         List<LogCleanupFilterElement> filterElements = new ArrayList<>(filters.size());
         for (Entry<String, CleanupFilterConfig> entry : filters.entrySet()) {
             filterElements.add(new LogCleanupFilterElement(entry.getKey(), entry.getValue().ifStartsWith));
         }
+
         final ArrayList<Handler> handlers = new ArrayList<>(3 + additionalHandlers.size());
-        ErrorManager errorManager = new OnlyOnceErrorManager();
 
         if (config.console.enable) {
             final Handler consoleHandler = configureConsoleHandler(config.console, errorManager, filterElements,
@@ -107,6 +102,21 @@ public class LoggingSetupRecorder {
             }
         }
 
+        Map<String, Handler> namedHandlers = createNamedHandlers(config, possibleFormatters, errorManager, filterElements);
+
+        for (Map.Entry<String, CategoryConfig> entry : categories.entrySet()) {
+            final String name = entry.getKey();
+            final Logger categoryLogger = logContext.getLogger(name);
+            final CategoryConfig categoryConfig = entry.getValue();
+            if (!"inherit".equals(categoryConfig.level)) {
+                categoryLogger.setLevelName(categoryConfig.level);
+            }
+            categoryLogger.setUseParentHandlers(categoryConfig.useParentHandlers);
+            if (categoryConfig.handlers.isPresent()) {
+                addNamedHandlersToCategory(categoryConfig, namedHandlers, categoryLogger, errorManager);
+            }
+        }
+
         for (RuntimeValue<Optional<Handler>> additionalHandler : additionalHandlers) {
             final Optional<Handler> optional = additionalHandler.getValue();
             if (optional.isPresent()) {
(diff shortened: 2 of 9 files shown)


core/runtime/src/main/java/io/quarkus/runtime/logging/CategoryConfig.java
```diff
@@ -1,20 +1,32 @@
 package io.quarkus.runtime.logging;
 
-import io.quarkus.runtime.annotations.ConfigGroup;
-import io.quarkus.runtime.annotations.ConfigItem;
+import io.quarkus.runtime.annotations.*;
```
geoand: Can you please convert this to non-start imports?

geoand:
It seems like there are formatting errors. Please run `./mvnw process-resources` and include the changes it makes into this PR

core/runtime/src/main/java/io/quarkus/runtime/logging/CategoryConfig.java
debae: Done

core/runtime/src/main/java/io/quarkus/runtime/logging/LoggingSetupRecorder.java
```diff
@@ -120,6 +111,62 @@ public void initializeLogging(LogConfig config, final List<RuntimeValue<Optional
...
+            List<LogCleanupFilterElement> filterElements, ArrayList<Handler> handlers, ErrorManager errorManager) {
+        final Handler syslogHandler = configureSyslogHandler(config.syslog, errorManager, filterElements);
+        if (syslogHandler != null) {
+            handlers.add(syslogHandler);
+        }
+        for (Entry<String, CategoryConfig> entry : categories.entrySet()) {
+            final CategoryConfig categoryConfig = entry.getValue();
+            final String name = entry.getKey();
+            final Logger categoryLogger = logContext.getLogger(name);
+            categoryLogger.setUseParentHandlers(false);
```
dmlloyd: A concern I have with this approach is that it will squelch all other handlers, because they do not have a hook into the category configuration, and adding such a hook will be complex.  This change completely takes over the log category so it becomes impossible to mix handlers which delegate to parents, and ones which don't.  The decision of whether to delegate to a parent handler for a given logger should lie with the user.

I think this problem might be better solved at the formatter level, by adding a mechanism for the formatter to choose a format based on factors such as category.

core/runtime/src/main/java/io/quarkus/runtime/logging/LoggingSetupRecorder.java
```diff
@@ -5,12 +5,8 @@
 
 import java.io.FileNotFoundException;
 import java.io.IOException;
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Locale;
-import java.util.Map;
+import java.util.*;
```
dmlloyd: `*` imports are not allowed.

core/runtime/src/main/java/io/quarkus/runtime/logging/CategoryConfig.java
```diff
@@ -1,20 +1,34 @@
...
 
 @ConfigGroup
 public class CategoryConfig {
 
     /**
-     * The minimum level that this category can be set to
+     * The log level level for this category
      */
     @ConfigItem(defaultValue = "inherit")
-    String minLevel;
```
dmlloyd: This should be removed in a separate PR.

dmlloyd (changes requested):

core/runtime/src/main/java/io/quarkus/runtime/logging/LoggingSetupRecorder.java
debae: I thought the setUseParentHandler() was used wether or not to take the handlers from the "root" logger. I'll give it another look, but indeed it was not my intention to break the hierarchy.

However, the suggestion you make is to add a format property in CategoryConfig instead of giving the ability to override the complete ( console/file/syslog ) handler ?
In our application for example we could write everything to file, and certain appenders only to console. If we only allow to override the format this is still an "issue" and we pollute the file as well.

dmlloyd: Each handler has its own formatter, so you don't have to worry about cross-pollution between them at that level.

Maybe we do need to work up a more generic configuration setup for log handlers.  We could introduce a `useParentHandler` property on the category configuration, and otherwise introduce per-handler configuration in the manner that you have done.

The problem to solve would be figuring out how to deal with pluggable formatters, since they could potentially apply to any handler on any category, and the current infrastructure doesn't account for that.

Finding a way to do this compatibly is an added challenge.

core/runtime/src/main/java/io/quarkus/runtime/logging/CategoryConfig.java
debae: Done, pull request https://github.com/quarkusio/quarkus/pull/5894 is made for this change.

dmlloyd (changes requested):
This is good so far, but a couple changes are needed.

core/runtime/src/main/java/io/quarkus/runtime/logging/LoggingSetupRecorder.java
```diff
@@ -69,18 +70,12 @@ public void initializeLogging(LogConfig config, final List<RuntimeValue<Optional
             final List<RuntimeValue<Optional<Formatter>>> possibleFormatters) {
 
         final Map<String, CategoryConfig> categories = config.categories;
+        final Map<String, NamedLogHandlerConfig> namedLogHandlers = config.handlers;
```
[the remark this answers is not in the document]
debae: So just to make sure, the following config :

`/**
     * Console handlers.
     * <p>
     * The named console handlers configured here can be linked on one or more categories.
     */
    @ConfigItem(name = "handler.console")
    @ConfigDocSection
    public Map<String, ConsoleConfig> consoleHandlers;

    /**
     * File handlers.
     * <p>
     * The named file handlers configured here can be linked on one or more categories.
     */
    @ConfigItem(name = "handler.file")
    @ConfigDocSection
    public Map<String, FileConfig> fileHandlers;

    /**
     * Syslog handlers.
     * <p>
     * The named syslog handlers configured here can be linked on one or more categories.
     */
    @ConfigItem(name = "handler.syslog")
    @ConfigDocSection
    public Map<String, SyslogConfig> syslogHandlers;
`

Should NOT be included in the **LogConfig** object ?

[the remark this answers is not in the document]
dmlloyd: These config maps can be in `LogConfig`.  But they should not be *processed* by the main step; rather, a new step should be introduced to map these to the new build item.

core/runtime/src/main/java/io/quarkus/runtime/logging/CategoryConfig.java
```diff
@@ -11,4 +14,11 @@
      */
     @ConfigItem(defaultValue = "inherit")
     String level;
+
+    /**
+     * The names of the handlers to link to this category.
+     */
+    @ConfigItem
+    Optional<List<String>> handlers;
+
```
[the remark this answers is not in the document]
debae: Added and put default value to true, because that is default behaviour.

core/deployment/src/main/java/io/quarkus/deployment/logging/LoggingResourceProcessor.java
```diff
@@ -79,11 +82,25 @@ void miscSetup(
...
     void setupLoggingRuntimeInit(LoggingSetupRecorder recorder, LogConfig log, List<LogHandlerBuildItem> handlers,
-            List<LogConsoleFormatBuildItem> consoleFormatItems) {
-        final List<RuntimeValue<Optional<Handler>>> list = handlers.stream().map(LogHandlerBuildItem::getHandlerValue)
+            List<LogConsoleFormatBuildItem> consoleFormatItems, List<NamedLogHandlersBuildItem> namedLogHandlersBuildItems) {
+        final List<RuntimeValue<Optional<Handler>>> additionalHandlers = handlers.stream()
+                .map(LogHandlerBuildItem::getHandlerValue)
+                .collect(Collectors.toList());
+        final List<RuntimeValue<Optional<Formatter>>> possibleFormatters = consoleFormatItems.stream()
+                .map(LogConsoleFormatBuildItem::getFormatterValue).collect(Collectors.toList());
+        final List<RuntimeValue<Optional<Map<String, List<Handler>>>>> namedHandlers = namedLogHandlersBuildItems.stream()
```
dmlloyd: This is still wrong, it has to be a `Map<String, List<RuntimeValue<Optional<Handler>>>>` I think.  I'll review more after I get back from running an errand.

dmlloyd: Rather it has to be a `Map<String, RuntimeValue<Optional<Handler>>>`.  If there is more than one handler with a given name, we need to fail the build.  Then the map is passed to the recorder, so the `RuntimeValue` is not dereferenced until the proper time.

gsmet (changes requested):
Just a note that this needs to be rebased and squashed before merging.

gsmet:
@debae, any progress on that one? We are not far from a conclusion from what I can see.

While at it, could you rebase and squash everything into a single commit?

Thanks!

debae:
> @debae, any progress on that one? We are not far from a conclusion from what I can see.
> 
> While at it, could you rebase and squash everything into a single commit?
> 
> Thanks!

@gsmet , almost got it working. The key to get it working at build time was changing the annotation on LogConfig and use phase BUILD_AND_RUN_TIME_FIXED. This however seems to have some side affects which I need to tackle... Will request some help if I cannot get it fixed, because the requested changes from @dmlloyd are ready.

dmlloyd:
That'll be an incompatible change though, so you unfortunately cannot use that solution.

hartimcwildfly:
@debae Thank you for your contribution! Could you add the possibility to provide additional named handlers?
(add something like List<RuntimeValue<Entry<String, Handler>>> additionalNamedHandlers to method LoggingSetupRecorder::initializeLogging(...)) So we can link a categoryLogger with a custom named handler. Would be great to see this feature in v1.2.0.

hartimcwildfly:
@debae [This commit](https://github.com/hartimcwildfly/quarkus/commit/e486b26b775a672653eea85058e9894301b49070)
could be enough.

core/runtime/src/main/java/io/quarkus/runtime/logging/LoggingSetupRecorder.java
```diff
@@ -73,21 +74,30 @@ public void initializeLogging(LogConfig config, final List<RuntimeValue<Optional
...
         rootLogger.setLevel(config.level.orElse(Level.INFO));
+
+        ErrorManager errorManager = new OnlyOnceErrorManager();
+        final Map<String, CleanupFilterConfig> filters = config.filters;
+        List<LogCleanupFilterElement> filterElements = new ArrayList<>(filters.size());
+        for (Entry<String, CleanupFilterConfig> entry : filters.entrySet()) {
+            filterElements.add(new LogCleanupFilterElement(entry.getKey(), entry.getValue().ifStartsWith));
+        }
+
+        Map<String, Handler> namedHandlers = createNamedHandlers(config, possibleFormatters, errorManager, filterElements);
```
dmlloyd: The console handler has to be created before the named handlers, otherwise the error manager is not correctly set up in time.  AFAICT this didn't need to be factored out to a separate method.

dmlloyd (changes requested):

core/runtime/src/main/java/io/quarkus/runtime/logging/LoggingSetupRecorder.java
debae: Moved the creation of the named handlers after creation of the handlers at root level.

dmlloyd (approved):

dmlloyd:
Everything OK with you @gsmet?

integration-tests/maven/src/test/java/io/quarkus/maven/it/GenerateConfigIT.java
```diff
@@ -36,16 +36,16 @@ void testAddExtensionWithASingleExtension() throws MavenInvocationException, IOE
...
         Assertions.assertTrue(file.contains("#quarkus.thread-pool.growth-resistance=0"));
         Assertions.assertTrue(file.contains("The executor growth resistance"));
 
         generateConfig("application.properties");
         //the existing file should not add properties that already exist
         file = loadFile("application.properties");
-        Assertions.assertTrue(file.contains("quarkus.log.file.enable=false"));
-        Assertions.assertFalse(file.contains("If file logging should be enabled"));
+        Assertions.assertTrue(file.contains("quarkus.log.level=INFO"));
+        Assertions.assertFalse(file.contains("The default log level"));
```
gsmet: I'm just wondering why this test was changed? I don't understand why it's needed so an explanation would be welcome.

gsmet: Per Baerten on Zulip: "Oh I changed them , because the structure now contains twice the value ""If file logging should be enabled"" , once for the root level, and once for configured categories. So this field was no longer a good candidate to assert on. That's why I changed it to level."

gsmet (approved):

debae:
Below the summary of the changes in this PR: 

Right now Quarkus supports logging to console / file / syslog. At this moment it only support configuring a format on that level.
When you application makes use of structured logging or has some specific class where you want a different format to be used when logging this is not possible.

The changes made allow you to configure named handlers for the 3 types of handlers ( console / file / syslog ) that are provided out of the box by Quarkus. These named handlers can then be linked to categories, by doing so you can deviate the log format for those specific categories.

Issue #5860, opened 2019-11-29
Cannot deserialize LocalDate in @QueryParam due to missing ParamConverterProvider


cg-d:
**Describe the bug**
I'm using Jackson (via quarkus-resteasy-jackson) as `ObjectMapper` with the `KotlinModule` and the `JavaTimeModule` registered using an `ObjectMapperCustomizer`. When I have an endpoint with a `@QueryParam` of type `LocalDate` quarkus does not start.

**Expected behavior**
I expect resteasy to use the provided `ObjectMapper` to deserialize a String formated "YYYY-MM-DD" into an instance of `LocalDate`.

**Actual behavior**
Omitting the name and method signature for brevity I get

```
RESTEASY003875: Unable to find a constructor that takes a String param or a valueOf() or fromString() method for javax.ws.rs.QueryParam("adate") for basetype: java.time.LocalDate
```

**To Reproduce**
1. Use 
```xml
        <dependency>
            <groupId>io.quarkus</groupId>
            <artifactId>quarkus-resteasy</artifactId>
        </dependency>
        <dependency>
            <groupId>io.quarkus</groupId>
            <artifactId>quarkus-resteasy-jackson</artifactId>
        </dependency>
        <dependency>
            <groupId>com.fasterxml.jackson.module</groupId>
            <artifactId>jackson-module-kotlin</artifactId>
        </dependency>
        <dependency>
            <groupId>com.fasterxml.jackson.datatype</groupId>
            <artifactId>jackson-datatype-jsr310</artifactId>
        </dependency>
```
2. Register a resource like
```kotlin
@GET
@Path("path")
fun broken( @QueryParam("adate") aDate: LocalDate) {}
```
3. Run a `@QuarkusTest` to try a start up the application.

**Environment (please complete the following information):**
 - Output of `uname -a` or `ver`: 
Linux ninja-cgd 5.3.0-23-generic #25-Ubuntu SMP Tue Nov 12 09:22:33 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux
 - Output of `java -version`: 
openjdk version "11.0.5-ea" 2019-10-15
OpenJDK Runtime Environment (build 11.0.5-ea+10-post-Ubuntu-0ubuntu1)
OpenJDK 64-Bit Server VM (build 11.0.5-ea+10-post-Ubuntu-0ubuntu1, mixed mode, sharing)
 - GraalVM version (if different from Java): 
not used
 - Quarkus version or git rev: 
[...]

PR #5886, opened 2019-11-29
Generalize registration of jackson modules

geoand:
This change makes use of the generated `ObjectMapperCustomizer` in order to allow auto-registration of discovered Jackson modules (thus removing the need of hard-coding the modules in the bean producer).

I initially got the idea for this by looking at #5860 but it's more general.

### DIFF
--- a/docs/src/main/asciidoc/kotlin.adoc
+++ b/docs/src/main/asciidoc/kotlin.adoc
@@ -389,4 +389,9 @@ One thing to note is that the live reload feature is not available when making c
 
 == Packaging the application
 
-As usual, the application can be packaged using `./mvnw clean package` and executed using the `-runner.jar` file. You can also build the native executable using `./mvnw package -Pnative`, or `./gradlew buildNative`.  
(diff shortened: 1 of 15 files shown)


loicmathieu:
This is a bit complicated... maybe we can just enable the auto discovery of modules inside the ObjectMapper ?

geoand:
Well it fits best with the what we already do in Quarkus and with the everything at build time logic.

It's not that complicated actually since everything we need is already in the code base

aguibert (approved):
LGTM

loicmathieu (commented):
@geoand now that I read again all the code I get it :)

Maybe it's worth adding the Scala module inside the Scala extension as you did for Kotlin ?

Maybe also add a note on the Kotlin guide about the fact that we auto register the Kotlin module to avoid some user trying to do it by themself :)

geoand:
@loicmathieu those are both excellent ideas! I didn't even know there was a Scala Jackson modul TBH :)

geoand:
@loicmathieu I incorporated your suggestions, thanks

loicmathieu (approved):
LGTM

gsmet (approved):

Issue #5890, opened 2019-12-01
add SSLContext customization for MongoDB extension


maxday:
**Description**

When creating a MongoClient, we may want to specify a custom SSLContext.

**Implementation ideas**

Since the SSLContext is not a primitive type, it's not possible to specify it via configuration.
Maybe a annotation is the solution :
 
```
@Inject
@SSLContext(context = MyCustomSSLContext.class)
MongoClient mongoClient;
```

MyCustomSSLContext could be a class implementing a SSLContextConfig interface with only one method returning a javax.net.ssl.SSLContext.

Then in MongoClientRecorder.java we can inject this SSLContext via [applyToSslSettings](https://mongodb.github.io/mongo-java-driver/3.9/javadoc/com/mongodb/connection/SslSettings.Builder.html#context-javax.net.ssl.SSLContext-)

What to you think ? I would love to create a PR for this.

danilopiazza:
I believe this issue to be superseded (and closed) by #46293, which added [TLS registry support for MongoDB Client](https://quarkus.io/blog/quarkus-3-21-0-released/#tls-registry-support-for-mongodb-client) in 3.21.

Issue #2809, opened 2019-06-12
ClassNotFoundException when using quarkus:dev


sarxos:
Hello @quarkus Team :)

I have a ~~multimodule~~ [project](https://github.com/sarxos/abberwoult) with ~~one of modules being~~ an example subproject which exposes very simple REST API. Everything works fine when I start this example with a runner:

```plain
java -jar target/abberwoult-example-0.0.1-SNAPSHOT-runner.jar -cp target/lib/*
```

The resource is started and I can access REST API. This is ok but takes some time, so I wanted to use `mvn quarkus:dev`. ~~I see that support for `quarkus:dev` in multimodule projects was introduced by https://github.com/quarkusio/quarkus/pull/1973 and thus should be working fine,~~ But when I try to run my example with a dev mojo, that is:

```plain
$ cd abberwoult-example
$ mvn quarkus:dev
```

I'm getting the following error:

```plain
18:12:22,936 ERROR [io.qua.dev.DevModeMain] Failed to start quarkus: java.lang.ExceptionInInitializerError
	at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
	at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.lang.reflect.Constructor.newInstance(Constructor.java:423)
	at java.lang.Class.newInstance(Class.java:442)
	at io.quarkus.runner.RuntimeRunner.run(RuntimeRunner.java:129)
	at io.quarkus.dev.DevModeMain.doStart(DevModeMain.java:173)
	at io.quarkus.dev.DevModeMain.main(DevModeMain.java:100)
Caused by: java.lang.RuntimeException: Failed to start quarkus
	at io.quarkus.runner.ApplicationImpl1.<clinit>(Unknown Source)
	... 8 more
Caused by: java.lang.IllegalStateException: Cannot load com.example.FruitRepositoryActor
	at com.github.sarxos.abberwoult.deployment.ActorInterceptorRegistryTemplate.register(ActorInterceptorRegistryTemplate.java:25)
	at io.quarkus.deployment.steps.AbberwoultProcessor$doRegisterActors9.deploy_0(Unknown Source)
	at io.quarkus.deployment.steps.AbberwoultProcessor$doRegisterActors9.deploy(Unknown Source)
[...]

PR #5893, opened 2019-12-01
Issue5890/add ssl context customization to mongo db extension

maxday:
Hello,

This is a PR to fix this issue : https://github.com/quarkusio/quarkus/issues/5890
Sample usage : 

1 - Create a Class which implements SSLContextConfig, here is a sample 
2 - Reference this class in the new property 

`quarkus.mongodb.ssl-context-class=io.maxday.MyCustomSSLContext`

```
    @RegisterForReflection
    public class MyCustomSSLContext implements SSLContextConfig {

        private final String KEYSTORE_PATH = "./ca-serti.jks";
        private final String KEYSTORE_PASSWORD = "changeit";

        @Override
        public SSLContext getSSLContext() {
            try {
                KeyStore keystore = KeyStore.getInstance(KeyStore.getDefaultType());
                try (InputStream in = new FileInputStream(KEYSTORE_PATH)) {
                    keystore.load(in, KEYSTORE_PASSWORD.toCharArray());
                }
                KeyManagerFactory keyManagerFactory =
                    KeyManagerFactory.getInstance(KeyManagerFactory.getDefaultAlgorithm());
                keyManagerFactory.init(keystore, KEYSTORE_PASSWORD.toCharArray());

                TrustManagerFactory trustManagerFactory =
                    TrustManagerFactory.getInstance(TrustManagerFactory.getDefaultAlgorithm());
                trustManagerFactory.init(keystore);

                SSLContext sslContext = SSLContext.getInstance("TLS");
                sslContext.init(
                        keyManagerFactory.getKeyManagers(),
                        trustManagerFactory.getTrustManagers(),
                        new SecureRandom());
                return sslContext;
            } catch(Exception e) {
                return null;
            }
        }
    }
```

If everything looks good for you, I'll add tests of course

(Please note that you won't be able to use quarkus:dev to test it due to https://github.com/quarkusio/quarkus/issues/2809)

Thanks


cescoffier (changes requested):
Is there any way we could cover this with a test?

gsmet:
Thanks for your contribution. Could you squash the commits into one? We try to avoid broken commits in the tree.

Then, I think we probably need to take a step back and see how we could fix that one globally. Because there's a chance we will need to fix this issue for other extensions.

maxday:
Sure, I'll add a test + squash into one commit today

loicmathieu:
@maxday did you try it in native mode?

You cannot have an SSLContext inside the image snapshoted heap when building a native image so I'm not sure you can create an SSLContext inside a recorder ...

When creating the OAuth2 extension, I favor offering a way to sepcify a certificate file instead of a whole SSLContext, this is the more common use case of creating our own SSLContext. Maybe we need to provides this (and maybe at Quarkus core level to avoid doing it for all extensions).
This is the OAuth2 extension implementation (not working in native mode): https://github.com/quarkusio/quarkus/blob/master/extensions/elytron-security-oauth2/runtime/src/main/java/io/quarkus/elytron/security/oauth2/runtime/OAuth2Recorder.java#L67

maxday:
@loicmathieu yes it works in native mode :) maybe I can push the tests and we'll discuss if we want to restrict and allow specifying just a certificate ?

I wrote "deployments" tests but I also want to push **an integration test with real certificates**. 
Is that OK if a create a new module under the integration-tests (mongo-client-ssl) ? @cescoffier 
I need to change the content of application.properties in the existing mongo-client but adding ssl=true breaks existing tests and I could not find an easy way to "swap" properties file in integration tests (see : https://groups.google.com/forum/#!searchin/quarkus-dev/multiple$20application$20properties%7Csort:date/quarkus-dev/Rsy30WZKv80/0py7_GHjDAAJ)

maxday:
Hi @gsmet @cescoffier 
I’ve added tests (unit + integration)
Let me know if I need to make additional changes :)

extensions/mongodb-client/runtime/src/main/java/io/quarkus/mongodb/runtime/MongoClientRecorder.java
```diff
@@ -266,4 +272,22 @@ private MongoCredential createMongoCredential(MongoClientConfig config) {
                     }
                 }).collect(Collectors.toList());
     }
+
+    private SSLContext getSSLContext(String className) {
+        try {
+            Class clazz = Class.forName(className);
```
cescoffier: the Class.forName might be a problem. @stuartwdouglas WDYT? Would that work with your ongoing changes?

extensions/mongodb-client/runtime/src/main/java/io/quarkus/mongodb/runtime/MongoClientRecorder.java
```diff
@@ -266,4 +272,22 @@ private MongoCredential createMongoCredential(MongoClientConfig config) {
...
+            if (SSLContextConfig.class.isAssignableFrom(clazz)) {
+                SSLContextConfig SSLContextConfig = (SSLContextConfig) clazz.getDeclaredConstructor().newInstance();
```
cescoffier: This won't work in native if the class is not registered for reflection.

integration-tests/mongodb-client-ssl/src/test/java/io/quarkus/it/mongodb/SSLMongoTest.java
```diff
@@ -0,0 +1,64 @@
...
+import de.flapdoodle.embed.mongo.MongodStarter;
+import de.flapdoodle.embed.mongo.config.IMongodConfig;
+import de.flapdoodle.embed.mongo.config.MongodConfigBuilder;
+import de.flapdoodle.embed.mongo.config.Net;
+import de.flapdoodle.embed.mongo.distribution.Version;
+import de.flapdoodle.embed.process.runtime.Network;
+import io.quarkus.test.junit.QuarkusTest;
+import io.restassured.RestAssured;
+
+@DisabledOnOs(OS.LINUX)
```
cescoffier: It would be great to get that tested on our CI (using linux).
@evanchooly any idea about this?

cescoffier (changes requested):

gsmet:
This one is outdated. Please open a new PR if you want to revisit this.

PR #5900, opened 2019-12-01
Feature/create artemis test module

### LINKED ISSUE #3833: ArtemisTestResource available
**Description**
Artemis integration test has an `ArtemisTestResource` https://github.com/quarkusio/quarkus/pull/2930/files#diff-76fa3fe6358cacc251f9cb4d29ae82ac to manage the lifecycle of an embedded Artemis instance for testing purposes.

The problem is that this class is not available for all users, I think that this kind of utilities should be available for any user because the way we test these integrations might be used by any dev.

So I think that this class should be moved to either an `artemis-test` artifact or in a place that can be used by anyone.

Generally speaking, I think that any util used in our integration tests should not be implemented in the integration tests itself but in a place that could be used by anyone, because at the end our necessities for testing probably are going to be the same for anyone writing test for Quarkus.

uaihebert:
Fixes #3833
The idea of this pr is:
1) create the test module with ArtemisResourceTest
1) expose it to be used by the developers that use Quarkus
1) replace integration tests that used the test class

### DIFF
--- a/bom/runtime/pom.xml
+++ b/bom/runtime/pom.xml
@@ -302,6 +302,11 @@
             </dependency>
             <dependency>
                 <groupId>io.quarkus</groupId>
+                <artifactId>quarkus-test-artemis</artifactId>
+                <version>${project.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>io.quarkus</groupId>
                 <artifactId>quarkus-artemis-jms</artifactId>
                 <version>${project.version}</version>
             </dependency>
--- a/integration-tests/artemis-core/pom.xml
+++ b/integration-tests/artemis-core/pom.xml
@@ -29,6 +29,10 @@
             <groupId>io.quarkus</groupId>
             <artifactId>quarkus-artemis-core</artifactId>
         </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-test-artemis</artifactId>
+        </dependency>
 
         <!-- Health check -->
         <dependency>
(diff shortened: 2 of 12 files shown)


extensions/artemis-test/pom.xml
```diff
@@ -0,0 +1,65 @@
...
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <artifactId>quarkus-build-parent</artifactId>
+        <groupId>io.quarkus</groupId>
+        <version>999-SNAPSHOT</version>
+        <relativePath>../../build-parent/pom.xml</relativePath>
+    </parent>
+
+    <modelVersion>4.0.0</modelVersion>
+    <artifactId>quarkus-artemis-test</artifactId>
```
aguibert: we should put this test module under `quarkus/test-framework/` with the other test modules, and set the pom parent to `quarkus-test-framework`

aguibert (commented):
Overall looks good, my only comment is on placement of the new module

gastaldi (changes requested):
I agree with @aguibert's comment in https://github.com/quarkusio/quarkus/pull/5900/files#r352410389. The module should be moved under `quarkus-test-framework`

gsmet:
Agreed and you will need to name it `quarkus-test-artemis` to be consistent with the others.

uaihebert:
Thank you for the feedback @aguibert @gastaldi @gsmet 

I updated it as requested.

gastaldi:
That looks good. Can you squash all your commits in a single one? Thank you!

uaihebert:
@gastaldi sure, no problem.

gastaldi (approved):

uaihebert:
hello @gastaldi the build failed on the windows test.

any tip for what should I look in the log? the logs are huge and I could not find the error in there.

gastaldi:
@uaihebert it doesn't seem related to your changes, let me see if I can trigger the build again

uaihebert:
hello @gastaldi indeed it was not related.

let me know if you need anything else.

gastaldi:
@uaihebert Merged. Thank you!

uaihebert:
thank you too.

Issue #5824, opened 2019-11-28
Use of quarkus-agroal with GCP CloudSql


jtama-op:
**Description**
As of now using quarkus-agroal to connect with GCP is possible but nor documented or trivial.

**Implementation ideas**
First option is to add com.google.cloud.sql:postgres-socket-factory or com.google.cloud.sql:mysql-socket-factory to the application dependencies and configuring datasource as follow : 
```properties
jdbc:postgresql:///<DB_NAME>?socketFactory=com.google.cloud.sql.postgres.SocketFactory&cloudSqlInstance=<CLOUD_SQL_INSTANCE>
```

Second option would be to add a dedicated options such as use-cloud-sql and cloud-sql-instance to  delegate the url forging to the `io.quarkus.agroal.runtime.AbstractDataSourceProducer` class. 
I don't know if we could arrange some kind of optional maven dependency to the google socket factory jar, to prevent loading unneeded jar.

I tried using both options. You can check the code for agroal extension here (https://github.com/jtama-op/quarkus/commit/eb27113c050b2c98a0ef046a84d366484adf6e21) (though it won't package due to `Dependency convergence error for com.google.http-client:google-http-client-jackson2:1.30.1`)

Anyway documenting this would be great as quarkus is great tool for deploying apps on Cloud and GCP beeing one of the top 3 provider.

gsmet:
@jtama-op would you be up for it?

It would be nice to have a guide similar to https://quarkus.io/guides/deploying-to-azure-cloud with all the possible GCP-related issues.

The sources are there: https://github.com/quarkusio/quarkus/tree/master/docs/src/main/asciidoc .

It doesn't need to be a big one, we can start to collect the solutions for the issues you solved.

jtama-op:
Ok. Should I create a dedicated branch With documentation only ? Or do I add it to my master where there is also my code to modify the extension?

gsmet:
Maybe create a PR with all you have and we can discuss from there what is better?

jtama-op:
Please check https://github.com/quarkusio/quarkus/pull/5901 as a base for discussion

PR #5901, opened 2019-12-02
Use of quarkus-agroal with GCP CloudSQL

jtama:
This pull request contains code to ease the use of GCP CloudSQL as well as documentation to use it without any code amendments.


extensions/agroal/runtime/src/main/java/io/quarkus/agroal/runtime/DataSourceRuntimeConfig.java
```diff
@@ -130,4 +130,16 @@
...
+     * When enabled jdbc url will be forged to enable google socket factory
+     */
+    @ConfigItem(defaultValue = "false")
+    public boolean useGcp;
+
+    /**
+     * Query executed to validate a connection.
+     */
+    @ConfigItem
+    public Optional<String> cloudSqlInstance;
```
Sanne: isn't this replicating the `validationQuerySql` property ?

extensions/agroal/runtime/src/main/java/io/quarkus/agroal/runtime/DataSourceRuntimeConfig.java
```diff
@@ -130,4 +130,16 @@
      */
     @ConfigItem
     public Optional<String> validationQuerySql;
+
+    /**
+     * When enabled jdbc url will be forged to enable google socket factory
+     */
+    @ConfigItem(defaultValue = "false")
+    public boolean useGcp;
```
Sanne: Just wondering, would it be possible to automate this in any way?

Would be great if we could detect it's being run on GCP and not need such a configuration flag

Sanne (changes requested):

Sanne:
hey, many thanks for proposing this. It's exciting to see more support for CGP!

I hope you could help me figure out if we can make it really easy and slick to run on GCP, so without needing any additional configuration flags? Please see my review comments for details.

It would also be great as to not have configuration properties which only apply to a single cloud.

docs/src/main/asciidoc/deploying-to-google-cloud.adoc
```diff
@@ -0,0 +1,109 @@
...
+This guide covers:
+
+* Connecting to a CloudSQL postgresql instance with quarkus-agroal and service account using CloudRun
+
+== Prerequisites
+
+For this guide you need:
+
+* roughly 1 hour
+* having access to an Google cloud plateform project with owner rights.
```
Sanne: a typo : s/plateform/platform/

extensions/agroal/runtime/src/main/java/io/quarkus/agroal/runtime/DataSourceRuntimeConfig.java
jtama-op: we can detect we are running on GCP by connecting to the `metadata.google.internal` as followed : 
```java 
try {
        Socket socket = new Socket("meta.google.internal", 80);
        System.out.println("Running in GCP!!");
    } catch (IOException e) {
        e.printStackTrace();
    }
```

But as stated in the [Deploying to google cloud documentation](docs/src/main/asciidoc/deploying-to-google-cloud.adoc), connecting with service account is just an option. I don't think Quarkus should force it. We could combine this detection and the presence of the cloudSqlInstance property to automate this.

Again as stated in the https://github.com/quarkusio/quarkus/issues/5824, for it to work as is, we would need to add a dependency to `com.google.cloud.sql:postgres-socket-factory:1.0.15` and/or `com.google.cloud.sql:mysql-socket-factory-connector-j-8:1.0.15` plus dependencies to the corresponding drivers.

As of now it is not possible as maven-enforcer-plugin will fail the build.

Alternatively, we could delegate the responsibility to the applications using the quarkus-agroal extension.

jtama-op:
Links to #5824

docs/src/main/asciidoc/deploying-to-google-cloud.adoc
jtama-op: Done

extensions/agroal/runtime/src/main/java/io/quarkus/agroal/runtime/DataSourceRuntimeConfig.java
jtama-op: Nope, this property value will end up in the jdbc url and will be used by the `SocketFactory` to provide access to the CloudSQL instance.

extensions/agroal/runtime/src/main/java/io/quarkus/agroal/runtime/DataSourceRuntimeConfig.java
jtama-op: I have removed postgres and postgres-socket-factory so that the build doesn't fail.

Again we could force the use of service account, but it would be making decisions instead of letting developper choose their connection mode

jtama-op:
Hi,
@Sanne I am not sure about the merge request process, do I need to do something more ?

Mihai-B:
After this is merged, in what version will it be available ?

extensions/agroal/runtime/src/main/java/io/quarkus/agroal/runtime/AbstractDataSourceProducer.java
```diff
@@ -195,6 +196,21 @@ public boolean isValid(Connection connection) {
...
+        // CloudSQL config
+        if (dataSourceRuntimeConfig.useGcp) {
+            disableSslSupport();
+            String cloudSqlInstance = dataSourceRuntimeConfig.cloudSqlInstance.orElseThrow(
+                    () -> new RuntimeException("Cloud sql instance property is mandatory to use Gcp Cloudsql"));
+            if (cloudSqlInstance.split(":").length != 3) {
+                throw new RuntimeException("Cloud sql instance should match the pattern project-id:zone:cloudSqlInstance");
+            }
+            agroalConnectionFactoryConfigurationSupplier.jdbcUrl(
+                    MessageFormat.format("{0}?socketFactory=com.google.cloud.sql.postgres.SocketFactory&cloudSqlInstance={1}"
```
Mihai-B: The socketFactory is bound to postgres driver and if you would like to use a different driver is not possible with current implementation. Is it possible to make this configurable from a property as well ?

jtama-op: Currently there are only two supported SGBD postgres and mysql, but yes I should make this configurable.

Mihai-B: But I think @Sanne is correct, this implementation is very GCP integrated. maybe we should switch a little bit the way we make the implementation for this.

I think the easiest way is to use the `agroalConnectionFactoryConfigurationSupplier.jdbcProperty(<property_name>, <value>)` method

extensions/agroal/runtime/src/main/java/io/quarkus/agroal/runtime/DataSourceRuntimeConfig.java
```diff
@@ -130,4 +130,16 @@
...
+
+    /**
+     * Query executed to validate a connection.
```
Mihai-B: I think the message here is wrong

extensions/agroal/runtime/src/main/java/io/quarkus/agroal/runtime/AbstractDataSourceProducer.java
```diff
@@ -195,6 +196,20 @@ public boolean isValid(Connection connection) {
...
+        // CloudSQL config
+        if (dataSourceRuntimeConfig.useGcp) {
+            disableSslSupport();
+            String cloudSqlInstance = dataSourceRuntimeConfig.cloudSqlInstance.orElseThrow(
+                    () -> new RuntimeException("Cloud sql instance property is mandatory to use Gcp Cloudsql"));
+            if (cloudSqlInstance.split(":").length != 3) {
+                throw new RuntimeException("Cloud sql instance should match the pattern project-id:zone:cloudSqlInstance");
+            }
+            agroalConnectionFactoryConfigurationSupplier.jdbcUrl(
+                    MessageFormat.format("{0}?socketFactory=com.google.cloud.sql.postgres.SocketFactory&cloudSqlInstance={1}",
```
Mihai-B: Maybe we should use the `agroalConnectionFactoryConfigurationSupplier.jdbcProperty(<property_name>, <value>) ` method here ?

Mihai-B (commented):
just added some ideas

extensions/agroal/runtime/src/main/java/io/quarkus/agroal/runtime/AbstractDataSourceProducer.java
jtama-op: I tried this but it didn't worked. I will re-take a loot at it. 

In fact I think I should create a separate extension for gcp sql datasources.

extensions/agroal/runtime/src/main/java/io/quarkus/agroal/runtime/DataSourceRuntimeConfig.java
jtama-op: You are right I will change this

extensions/agroal/runtime/src/main/java/io/quarkus/agroal/runtime/AbstractDataSourceProducer.java
jtama-op: I don't think it works

extensions/agroal/runtime/src/main/java/io/quarkus/agroal/runtime/AbstractDataSourceProducer.java
Mihai-B: Can you check: https://github.com/quarkusio/quarkus/pull/6634 ?

jtama-op: I think it won't work (i tried this already)

gsmet:
This PR is invalid and contains a lot of unrelated commits.

Closing. Please open a new PR to pursue this.

PR #5912, opened 2019-12-02
Vertx Graphql test enhancement for websocket based connections

cristianonicolai:


gsmet:
Could you explain what problem you're trying to solve?

It's not exactly obvious from the diff :).

cristianonicolai:
@gsmet the initial test just verifies that a connection can be established, with this change I'm actually sending a graphql query and getting a response back... so a small enhancement to make sure there is nothing else missing in handling graphql protocol over websocket.

gsmet (approved):

gsmet:
OK, thanks for the explanation, let's merge it!

PR #5914, opened 2019-12-03
Do not use latest in the Dockerfile.jvm

cescoffier:
Avoid issues such as https://github.com/quarkusio/quarkus/issues/5848

### DIFF
--- a/devtools/platform-descriptor-json/src/main/resources/templates/dockerfile-jvm.ftl
+++ b/devtools/platform-descriptor-json/src/main/resources/templates/dockerfile-jvm.ftl
@@ -14,7 +14,7 @@
 # docker run -i --rm -p 8080:8080 quarkus/${project_artifactId}-jvm
 #
 ###
-FROM fabric8/java-alpine-openjdk8-jre
+FROM fabric8/java-alpine-openjdk8-jre:1.6.5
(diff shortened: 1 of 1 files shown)


geoand (approved):
Of course yes for anything that doesn't use `latest`!

gsmet:
How will we deal with security updates to the image?

geoand:
We'll probably have to check it periodically

cescoffier:
Yes, we will have to check periodically.

With `latest` security updates are not covered either as you get the `latest` from the initial pull, and then you need to pull again explicitly.

jaikiran:
I was under the impression that `latest` would mean that docker, during the build process, would fetch any changes as compared to any previously locally downloaded image. Apparently not. Looks like `--pull` needs to be explicitly passed to `docker build` for that to work https://github.com/moby/moby/pull/9281, which would mean updating our docs to show the usage of this option. But then that's not enforcable in any way. So I think sticking to a specific version and then regularly updating that version in our `Dockerfile` is the better way to go about it.

geoand:
Yes, `latest` causes way to many problems and in my experience should avoided like the plague! 
The even worse thing is that you usually have no idea what went wrong since PRs or examples failing for no apparent reason.

gastaldi (approved):

PR #5916, opened 2019-12-03
Format profile - activeByDefault removal, keeping just no-format property

rsvoboda:
Format profile - `activeByDefault` removal, keeping just `no-format` property.

`activeByDefault` behavior already achieved by default via `!no-format` activation property.

### DIFF
--- a/build-parent/pom.xml
+++ b/build-parent/pom.xml
@@ -369,7 +369,6 @@
         <profile>
             <id>format</id>
             <activation>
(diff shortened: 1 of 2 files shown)


gsmet:
While I agree it's redundant, it doesn't hurt to be over clear IMHO so I would prefer keeping it that way.

Thanks anyway.

rsvoboda:
ok, np at all :)

PR #5930, opened 2019-12-04
First version of Qute Reference Guide

mkouba:
- also add TemplateDataTest and fix some bugs in ValueResolverGenerator

### DIFF
--- a/docs/src/main/asciidoc/qute-reference.adoc
+++ b/docs/src/main/asciidoc/qute-reference.adoc
@@ -1,0 +1,699 @@
+////
+This guide is maintained in the main Quarkus repository
+and pull requests should be submitted there:
+https://github.com/quarkusio/quarkus/tree/master/docs/src/main/asciidoc
+////
+= Quarkus - Qute Reference Guide
+
+include::./attributes.adoc[]
+
+:numbered:
+:sectnums:
+:sectnumlevels: 4
+:toc:
+
+== Hello World Example
+
+In this example, we'd like to demonstrate the basic workflow when working with Qute templates.
+Let's start with a simple hello world example.
+We will always need some *template contents*:
+
+.hello.html
+[source,html]
+----
+<html>
+ <p>Hello {name}! <1>
+</html>
+----
+<1> `{name}` is a value expression that is evaluated when the template is rendered.
+
+Then, we will need to parse the contents into a *template definition* Java object.
+A template definition is an instance of `io.quarkus.qute.Template`.
+
+If using Qute "standalone" you'll need to create an instance of `io.quarkus.qute.Engine` first.
+The `Engine` represents a central point for template management with dedicated configuration. 
+Let's use the convenient builder:
+
+[source,java]
+----
+Engine engine = Engine.builder().addDefaults().build();
+----
+
+TIP: In Quarkus, there is a preconfigured `Engine` available for injection - see <<quarkus_integration>>.
+
+If we have an `Engine` instance we could parse the template contents:
+
+[source,java]
+----
+Template helloTemplate = engine.parse(helloHtmlContent);
+----
+
+TIP: In Quarkus, you can simply inject the template definition. The template is automatically parsed and cached - see <<quarkus_integration>>.
+
+Finally, we will create a *template instance*, set the data and render the output:
+
+[source,java]
+----
+// Renders <html><p>Hello Jim!</p></html>
+helloTemplate.data("name", "Jim").render(); <1> 
+----
+<1> `Template.data(String, Object)` is a convenient method that creates a template instance and sets the data in one step.
+
+So the workflow is simple: 
+
+1. Create template contents (`hello.html`),
+2. Parse template definition (`io.quarkus.qute.Template`),
+3. Create template instance (`io.quarkus.qute.TemplateInstance`),
+4. Render output.
+
+The `Engine` is able to cache the definitions so that it's not necessary to parse the contents again and again.
+
+NOTE: In Quarkus, the caching is done automatically.
+
+== Core Features
+
+=== Syntax and Building Blocks
+
+The dynamic parts of a template include:
+
+* *comment* 
+** `{! This is a comment !}`,
+** could be multi-line,
+** may contain expressions and sections: `{! {#if true} !}`.
+* *expression*
+** outputs the evaluated value,
+** simple properties: `{foo}`, `{item.name}`,
+** virtual methods: `{item.get(name)}`, `{name ?: 'John'}`,
+** with namespace: `{inject:colors}`.
+* *section tag*
+** may contain expressions and sections: `{#if foo}{foo.name}{/if}`,
+** the name in the closing tag is optional: `{#if active}ACTIVE!{/}`,
+** can be empty: `{#myTag image=true /}`,
+** may declare nested section blocks: `{#if item.valid} Valid. {#else} Invalid. {/if}` and decide which block to render.
+
+==== Expressions
+
+An expression consists of:
+
+* an optional namespace followed by a colon `:`, 
+* parts separated by dot `.`.
+
+The first part of the expression is always resolved against the <<current_context_object, current context object>>.
+If no result is found for the first part it's resolved against the parent context object (if available).
+For an expression that starts with a namespace the current context object is found using the available ``NamespaceResolver``s.
+For an expression that does not start with a namespace the current context object is *derived from the position* of the tag.
+All other parts are resolved using `ValueResolver` s against the result of the previous resolution.
+
+For example, expression `{name}` has no namespace and single part - "name".
+The "name" will be resolved using all available value resolvers against the current context object.
+However, the expression `{global:colors}` has the namespace "global" and single part - "colors".
+First, all available `NamespaceResolver` s will be used to find the current context object.
+And afterwards value resolvers will be used to resolve "colors" against the context object found. 
+
+[source]
+----
+{name} <1>
+{item.name} <2>
+{global:colors} <3> 
+----
+<1> no namespace, one part -`name`
+<2> no namespace, two parts - `item`, `name`
+<3> namespace `global`, one part - `colors`
+
+An expression part could be a "virtual method" in which case the name can be followed by parameters in parentheses. 
+
+[source]
+----
+{item.getLabels(1)} <1>
+{name or 'John'} <2>
+----
+<1> no namespace, two parts - `item`, `getLabels(1)`, the second part is a virtual method with name `getLabels` and params `1`
+<2> infix notation, translated to `name.or('John')`; no namespace, two parts - `name`, `or('John')`
+
+[[current_context_object]]
+===== Current Context
+
+If an expression does not specify a namespace the current context object is derived from the position of the tag.
+By default, the current context object represents the data passed to the template instance.
+However, sections may change the current context object.
+A typical example is the for/each loop - during iteration the content of the section is rendered with each element as the current context object:
+
+[source]
+----
+{#each items}
+ {count}. {it.name} <1>
+{/each}
+
+{! Another form of iteration... !}
+{#for item in items}
+  {count}. {item.name} <2>
+{/for}
+----
+<1> `it` is an implicit alias. `name` is resolved against the current iteration element.
+<2> Loop with an explicit alias `item`.
+
+[TIP]
+====
+Data passed to the template instance are always accessible using the `data` namespace. 
+This could be useful to access data for which the key is overriden:  
+
+[source,html]
+----
+<html>
+{item.name} <1>
+<ul>
+{#for item in item.getDerivedItems()} <2>
+  <li>
+  {item.name} <3> 
+  is derived from
+  {data:item.name} <4>
+  </li>
+{/for}
+</ul>
+</html>
+----
+<1> `item` is passed to the template instance as a data object.
+<2> Iterate over the list of derived items.
+<3> `item` is an alias for the iterated element.
+<4> Use the `data` namespace to access the `item` data object.
+
+====
+
+==== Sections
+
+A section:
+
+* has a start tag
+** starts with `#`, followed by the name of the section such as `{#if}` and `{#each}`,
+* may be empty 
+** tag ends with `/`, ie. `{#emptySection /}`
+* may contain other expression, sections, etc.
+** the end tag starts with `/` and contains the name of the section (optional): `{#if foo}Foo!{/if}` or `{#if foo}Foo!{/}`,
+
+The start tag can also define parameters.
+The parameters have optional names.
+A section may contain several content *blocks*. 
+The "main" block is always present.
+Additional/nested blocks also start with `#` and can have parameters too - `{#else if item.isActive}`.
+A section helper that defines the logic of a section can "execute" any of the blocks and evaluate the parameters.
+
+[source]
+----
+{#if item.name is 'sword'}
+  It's a sword!
+{#else if item.name is 'shield'}
+  It's a shield!
+{#else}
+  Item is neither a sword nor a shield.
+{/if}
+----
+
+===== Loop Section
+
+The loop section makes it possible to iterate over an instance of `Iterable`, `Map` 's entry set and `Stream`.
+It has two flavors.
+The first one is using the `each` name alias.
+
+[source]
+----
+{#each items}
+ {it.name} <1>
+{/each}
+----
+<1> `it` is an implicit alias. `name` is resolved against the current iteration element.
+
+The other form is using the `for` name alias and can specify the alias used to reference the iteration element:
+
+[source]
+----
+{#for item in items}
+  {item.name}
+{/for}
+----
+
+It's also possible to access the iteration metadata inside the loop:
+
+[source]
+----
+{#each items}
+ {count}. {it.name} <1>
+{/each}
+----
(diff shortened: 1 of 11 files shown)


extensions/qute/qute-reference-guide.adoc
```diff
@@ -0,0 +1,673 @@
+:toc:
+:toc-placement!:
```
gsmet: Why is it in the extensions directory? It won't be visible on the website, whereas it looks very valuable.

I think it should be placed alongside the cdi guide and be called `qute-reference` to be consistent with https://quarkus.io/guides/cdi-reference .

extensions/qute/qute-reference-guide.adoc
```diff
@@ -0,0 +1,673 @@
...
+
+= Qute Reference Guide
+
+toc::[]
```
gsmet: There's a good chance this will need to be removed.

extensions/qute/qute-reference-guide.adoc
```diff
@@ -0,0 +1,673 @@
...
+The dynamic parts of a template include:
+
+* *comment* 
+** `{! This is a comment !}`,
+** could be multi-line,
+** may contain expressions and sections: `{! {#if true} !}`.
+* *expression*
+** outputs the evaluated value,
+** simple properties: `{foo}`, `{item.name}`,
+** virtual methods: `{item.get(name)}`, `{name ?: 'John'}`,
```
gsmet: Do we support, `{name ? 'something' : 'else'}` ? I don't see it in any example?

Also do we have a `concat` operator?

extensions/qute/qute-reference-guide.adoc
```diff
@@ -0,0 +1,673 @@
...
+==== Expressions
+
+An expression consists of:
+
+* an optional namespace followed by a colon `:`, 
+* parts separated by dot `.`.
+
+The first part of the expression is always resolved against the <<current_context_object, current context object>>.
+If no result is found for the first part it's resolved against the parent context object (if available).
+For an expression that starts with a namespace the current context object is found using the available `NamespaceResolver` s.
```
gsmet: You can use double backticks when you're in this case. Then you don't need a space after.

extensions/qute/qute-reference-guide.adoc
```diff
@@ -0,0 +1,673 @@
...
+For an expression that does not start with a namespace the current context object is *derived from the position* of the tag.
+All other parts are resolved using `ValueResolver` s against the result of the previous resolution.
+
+For example, expression `{name}` has no namespace and single part - "name".
+The "name" will be resolved using all available value resolvers s against the current context object.
```
gsmet:
```suggestion
The "name" will be resolved using all available value resolvers against the current context object.
```

extensions/qute/qute-reference-guide.adoc
```diff
@@ -0,0 +1,673 @@
...
+<1> no namespace, two parts - `item`, `getLabels(1)`, the second part is a virtual method with name `getLabels` and params `1`
+<2> infix notation, translated to `name.or('John')`; no namespace, two parts - `name`, `or('John')`
+
+[[current_context_object]]
+===== Current Context
+
+If an expression does not specify a namespace the current context object is derived from the position of the tag.
+By default, the current context object represents the data passed to the template instance.
+However, sections may change the current context object.
+A typical example is for/each loop - during iteration the content of the section is rendered with each element as the current context object:
```
gsmet:
```suggestion
A typical example is the for/each loop - during iteration the content of the section is rendered with each element as the current context object:
```
?

extensions/qute/qute-reference-guide.adoc
```diff
@@ -0,0 +1,673 @@
...
+  {count}. {item.name} <2>
+{/for}
+----
+<1> `it` is an implicit alias. `name` is resolved against the current iteration element.
+<2> Loop with an explicit alias `item`.
+
+[TIP]
+====
+Data passed to the template instance are always accessible using the `data` namespace. 
+This could be useful to access data for which the key is overrided:
```
gsmet:
```suggestion
This could be useful to access data for which the key is overridden:  
```

extensions/qute/qute-reference-guide.adoc
```diff
@@ -0,0 +1,673 @@
...
+Additional/nested blocks also start with `#` and can have parameters too - `{#else if item.isActive}`.
+A section helper that defines the logic of a section can "execute" any of the blocks and evaluate the parameters.
+
+----
+{#if item.name is 'sword'}
+  It's a sword!
+{#else if item.name is 'shield'}
+  It's a shield!
+{#else}
+  Item is nor a sword nor a shield.
```
gsmet:
```suggestion
  Item is neither a sword nor a shield.
```

extensions/qute/qute-reference-guide.adoc
```diff
@@ -0,0 +1,673 @@
...
+** the end tag starts with `/` and contains the name of the section (optional): `{#if foo}Foo!{/if}` or `{#if foo}Foo!{/}`,
+
+The start tag can also define parameters.
+The parameters have optional names.
+A section may contain several content *blocks*. 
+The "main" block is always present.
+Additional/nested blocks also start with `#` and can have parameters too - `{#else if item.isActive}`.
+A section helper that defines the logic of a section can "execute" any of the blocks and evaluate the parameters.
+
+----
```
gsmet: FTR, Asciidoc behaves a bit differently when you define block like that and proper `[source]`. I recommend to always add `[source]`.

extensions/qute/qute-reference-guide.adoc
```diff
@@ -0,0 +1,673 @@
...
+{#else}
+  Item is nor a sword nor a shield.
+{/if}
+----
+
+===== Loop Section
+
+The loop section makes it possible to iterate over an instance of `Iterable`, `Map` 's entry set and `Stream`.
+It has two flavors.
+The first one is using `each` name alias.
```
gsmet:
```suggestion
The first one is using the `each` name alias.
```
?

extensions/qute/qute-reference-guide.adoc
```diff
@@ -0,0 +1,673 @@
...
+
+----
+{#each items}
+ {it.name} <1>
+{/each}
+----
+<1> `it` is an implicit alias. `name` is resolved against the current iteration element.
+
+The other form is using `for` name alias and can specify the alias used to reference the iteration element:
```
gsmet:
```suggestion
The other form is using the `for` name alias and can specify the alias used to reference the iteration element:
```
?

extensions/qute/qute-reference-guide.adoc
```diff
@@ -0,0 +1,673 @@
...
+{#each items}
+ {count}. {it.name} <1>
+{/each}
+----
+<1> `count` represents one-based index. Metadata also include zero-based `index`, `hasNext`, `odd`, `even`.
+
+===== If Section
+
+A basic control flow section.
+The simplest possible version accepts a single parameter and renders the contet if it's evaluated to `true` (or `Boolean.TRUE`).
```
gsmet:
```suggestion
The simplest possible version accepts a single parameter and renders the content if it's evaluated to `true` (or `Boolean.TRUE`).
```

extensions/qute/qute-reference-guide.adoc
```diff
@@ -0,0 +1,673 @@
...
+|equals 
+|`eq`, `==`, `is` 
+
+|not equals 
+|`ne`, `!=`
+
+|greater than 
+|`gt`, `>`
+
+|greater equals
```
gsmet:
```suggestion
|greater than or equal to
```

extensions/qute/qute-reference-guide.adoc
```diff
@@ -0,0 +1,673 @@
...
+|`ge`, `>=`
+
+|less than 
+|`lt`, `<`
+
+|less equals
```
gsmet:
```suggestion
|less than or equal to
```

extensions/qute/qute-reference-guide.adoc
```diff
@@ -0,0 +1,673 @@
...
+
+|===
+
+----
+{#if item.age > 10}
+ This item is very old.
+{/if}
+----
+
+NOTE: Multiple conditions are not supported.
```
gsmet: Mmmh, this will probably be an annoying limitation. @geoand has the same issue with his `@PreAuthorize` conditions.

extensions/qute/qute-reference-guide.adoc
```diff
@@ -0,0 +1,673 @@
...
+
+.Template "detail"
+[source,html]
+----
+{#include base} <1>
+    {#title}My Title{/title} <2>
+    {#body}
+    <div>
+        My body.
+    </div>
```
gsmet: Is it normal you don't end the body block?

It probably is considering the note below but it feels a bit weird.

extensions/qute/qute-reference-guide.adoc
```diff
@@ -0,0 +1,673 @@
...
+
+[source,xml]
+----
+<dependency>
+   <groupId>io.quarkus</groupId>
+   <artifactId>quarkus-qute</artifactId>
+</dependency>
+----
+
+Quarkus extension provides a preconfigured engine instance and registers a bean with scope `@Singleton`, bean type `io.quarkus.qute.Engine` and qualifier `@Default`.
```
gsmet:
```suggestion
The Quarkus extension provides a preconfigured engine instance and registers a bean with scope `@Singleton`, bean type `io.quarkus.qute.Engine` and qualifier `@Default`.
```

extensions/qute/qute-reference-guide.adoc
```diff
@@ -0,0 +1,673 @@
...
+<title>Qute Hello</title>
+</head>
+<body>
+    <h1>{title}</h1> <2>
+    Hello {foo.message}! <3>
+</body>
+</html>
+----
+<1> Parameter declaration - maps `foo` to `org.acme.Foo`.
+<2> Not validated - no matching a param declaration.
```
gsmet:
```suggestion
<2> Not validated - not matching a param declaration.
```

extensions/qute/qute-reference-guide.adoc
```diff
@@ -0,0 +1,673 @@
...
+----
+{#each items} <1>
+   {it.price.setScale(2, rounding)} <1>
+{/each}
+----
+<1> The generated value resolver knows how to invoke the `BigDecimal.setScale()` method.
+
+=== RESTEasy Integration
+
+If you want to use Qute in your JAX-RS application, you'll need to add the `quarkus-qute-resteasy` extension first.
```
gsmet:
```suggestion
If you want to use Qute in your JAX-RS application, you'll need to add the `quarkus-resteasy-qute` extension first.
```

extensions/qute/qute-reference-guide.adoc
```diff
@@ -0,0 +1,673 @@
...
+  @GET
+  @Produces({ MediaType.TEXT_HTML, MediaType.TEXT_PLAIN })
+  public Rendering item() {
+     return item.data(new Item("Alpha", 1000)); <2>
+  }
+}
+----
+<1> Inject a variant template with base path derived from the injected field - `src/main/resources/templates/item`.
+<2> The resulting output depends on the `Accept` header received from the client. For `text/plain` the `src/main/resources/templates/item.txt` template is used. For `text/html` the `META-INF/resources/templates/item.html` template is used.
+
```
gsmet: Don't you also support `Accept-Language`? I have a vague recollection of seeing something about it.

gsmet (changes requested):
I reviewed the documentation.

extensions/qute/qute-reference-guide.adoc
mkouba: Well, I wasn't sure that the docs could contain a "real" reference guide. The CDI ref guide is kind of special - it's not a real reference guide but back then we did not find a better name.

extensions/qute/qute-reference-guide.adoc
mkouba: > Do we support, {name ? 'something' : 'else'}

No, we don't.

> Also do we have a concat operator?

Not out of the box. What's the use case?

extensions/qute/qute-reference-guide.adoc
mkouba: Yes, end tag is optional for section blocks. The start tag of an additional section block is usually more like a separator:

```
{#if foo}
...
{#else} -> the same situation
...
{/if} 
```

extensions/qute/qute-reference-guide.adoc
mkouba: Not yet.

extensions/qute/qute-reference-guide.adoc
gsmet: I remember seeing a lot of templates which uses concat and also some sort of pipe operator to deal with transformations.

I think the latter is probably covered with the methods you can add? This was especially used to format dates. It would be nice to have something documented for that.

As for the former, it could typically be used when using a ternary operator if we end up having one :).

Thinking out loud, one other thing that I saw in other templating languages is some constructs to easily deal with plurals. But I suppose the elvis operator for simple cases and a ternary operator for more complex cases could do?

extensions/qute/qute-reference-guide.adoc
mkouba: Yes, I know. I just didn't have time to implement something like that ;-).

docs/src/main/asciidoc/qute-reference.adoc
```diff
@@ -0,0 +1,698 @@
...
+https://github.com/quarkusio/quarkus/tree/master/docs/src/main/asciidoc
+////
+= Quarkus - Qute Reference Guide
+
+include::./attributes.adoc[]
+
+:numbered:
+:sectnums:
+:sectnumlevels: 4
+:toc:
```
gsmet: I think we don't have that one in the other guides. The TOC is generated by the Jekyll plugin IIRC.

mkouba: Yep, I use this in the CDI ref guide to generate a numbered TOC. It seems to work fine.

docs/src/main/asciidoc/qute-reference.adoc
```diff
@@ -0,0 +1,698 @@
...
+|`gt`, `>`
+
+|greater than or equal to
+|`ge`, `>=`
+
+|less than 
+|`lt`, `<`
+
+|less than or equal to
+|`le`, `<=`
```
gsmet: I just checked the generated output and this gets replaced by an arrow by Asciidoctor. It needs some escaping.

Using `\<=` does the trick.

docs/src/main/asciidoc/qute-reference.adoc
```diff
@@ -0,0 +1,698 @@
...
+{/if}
+----
+<1> `showImage` is a named parameter.
+<2> `it` is a special key that is replaced with the first unnamed param of the tag.
+
+Now if we register this template under the name `item` and if we add a `UserTagSectionHelper` to the engine:
+
+[source,java]
+----
+Engine engine = Engine.builder().addSectionHelper(new UserTagSectionHelper.Factory("item"))
```
gsmet: Putting `addSectionHelper` on its own line as `build()` will help.

docs/src/main/asciidoc/qute-reference.adoc
```diff
@@ -0,0 +1,698 @@
...
+        this.price = price;
+    }
+}
+
+class MyExtensions {
+
+  @TemplateExtension
+  static BigDecimal discountedPrice(Item item) { <1>
+     return item.getPrice().multiply(new BigDecimal("0.9"));
+  }
```
gsmet: This is minor but indentation is not consistent here. You have four spaces then 2.

docs/src/main/asciidoc/qute-reference.adoc
```diff
@@ -0,0 +1,698 @@
...
+<2> The resulting output depends on the `Accept` header received from the client. For `text/plain` the `src/main/resources/templates/item.txt` template is used. For `text/html` the `META-INF/resources/templates/item.html` template is used.
+
+
+=== Development Mode
+
+In the development mode, all files located in `src/main/resources/templates` are watched for changes and modifications are immediately visible.
+
+=== Configuration
+
+* `quarkus.qute.suffixes` - a set of suffixes used when attempting to locate a template; by default, `engine.getTemplate(foo)` would result in several lookups: `src/main/resources/templates/foo`, `src/main/resources/templates/foo.html` and `src/main/resources/templates/foo.txt`
```
gsmet: I would use the configuration reference include and add more information to the javadoc if you need more.

docs/src/main/asciidoc/qute-reference.adoc
mkouba: Oh, I didn't know about this one. Thanks!

extensions/qute/qute-reference-guide.adoc
mkouba: For the record I added a value resolver to support `{condition ? 'something' : 'else'}`. We need to decide whether to test `true` or `truthy` (from JS). Currently, the condition must be true.

gsmet (approved):
Thanks for this work. Let's wait for CI to go green to be sure.

PR #5932, opened 2019-12-04
MainClassBuildStep - log error when app fails to start

mkouba:
- remove Throwable.printStackTrace()

### DIFF
--- a/core/deployment/src/main/java/io/quarkus/deployment/steps/MainClassBuildStep.java
+++ b/core/deployment/src/main/java/io/quarkus/deployment/steps/MainClassBuildStep.java
@@ -12,6 +12,7 @@ import java.util.Optional;
 import java.util.stream.Collectors;
 
 import org.graalvm.nativeimage.ImageInfo;
+import org.jboss.logging.Logger;
 
 import io.quarkus.builder.Version;
 import io.quarkus.deployment.GeneratedClassGizmoAdaptor;
@@ -58,6 +59,7 @@ class MainClassBuildStep {
     private static final String APP_CLASS = "io.quarkus.runner.ApplicationImpl";
     private static final String MAIN_CLASS = "io.quarkus.runner.GeneratedMain";
     private static final String STARTUP_CONTEXT = "STARTUP_CONTEXT";
+    private static final String LOG = "LOG";
     private static final String JAVA_LIBRARY_PATH = "java.library.path";
     private static final String JAVAX_NET_SSL_TRUST_STORE = "javax.net.ssl.trustStore";
 
@@ -101,6 +103,9 @@ class MainClassBuildStep {
 
         // Application class: static init
 
+        // LOG static field
+        FieldCreator logField = file.getFieldCreator(LOG, Logger.class).setModifiers(Modifier.STATIC);
+
         FieldCreator scField = file.getFieldCreator(STARTUP_CONTEXT, StartupContext.class);
         scField.setModifiers(Modifier.STATIC);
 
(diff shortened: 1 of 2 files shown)


stuartwdouglas (approved):

core/deployment/src/main/java/io/quarkus/deployment/steps/MainClassBuildStep.java
```diff
@@ -235,9 +244,11 @@ MainClassBuildItem build(List<StaticBytecodeRecorderBuildItem> staticInitTasks,
                 tryBlock.load(LaunchMode.DEVELOPMENT.equals(launchMode.getLaunchMode())));
 
         cb = tryBlock.addCatch(Throwable.class);
-        cb.invokeVirtualMethod(ofMethod(Throwable.class, "printStackTrace", void.class), cb.getCaughtException());
+        cb.invokeVirtualMethod(ofMethod(Logger.class, "error", void.class, Object.class, Throwable.class),
+                cb.readStaticField(logField.getFieldDescriptor()), cb.load("Failed to start application"),
+                cb.getCaughtException());
         cb.invokeVirtualMethod(ofMethod(StartupContext.class, "close", void.class), startupContext);
-        cb.throwException(RuntimeException.class, "Failed to start quarkus", cb.getCaughtException());
+        cb.throwException(RuntimeException.class, "Failed to start quarkus");
```
gsmet: There are test failures related to this change.

From what I saw in the log, the exception wasn't logged twice so I would keep the cause here.

gsmet (changes requested):
Thanks @mkouba . It wasn't an Arc issue after all then. Could you have a look at my comment?

core/deployment/src/main/java/io/quarkus/deployment/steps/MainClassBuildStep.java
mkouba: Hm, I tested my PR with quickstart and it was logged twice when I tried to run the app using the `-runner.jar`. I can try the dev mode and native image too..

mkouba: Actually some ArC tests failed because of this ;-).

mkouba: So I verified that it's logged twice but I agree that we should keep the cause here... I will update the PR shortly.

mkouba:
The windows job failed with:
```
Caused by: org.quartz.JobPersistenceException: Couldn't check for existence of job: Table "QRTZ_JOB_DETAILS" not found; SQL statement:
SELECT JOB_NAME FROM QRTZ_JOB_DETAILS WHERE SCHED_NAME = 'QuarkusQuartzScheduler' AND JOB_NAME = ? AND JOB_GROUP = ? [42102-197] [See nested exception: org.h2.jdbc.JdbcSQLException: Table "QRTZ_JOB_DETAILS" not found; SQL statement:
SELECT JOB_NAME FROM QRTZ_JOB_DETAILS WHERE SCHED_NAME = 'QuarkusQuartzScheduler' AND JOB_NAME = ? AND JOB_GROUP = ? [42102-197]
```

gsmet:
@mkouba so I might be saying something stupid but if it's logged twice, should we simply get rid of it?

mkouba:
@gsmet the problem is that afterwards we will not be able to assert the build exception in tests and possibly in the dev mode?

gsmet (approved):

Issue #5800, opened 2019-11-27
Cannot override OAuth URL at run time


berkyvz:
**Describe the bug**
We decide to use Quarkus in cloud native application as a company but it seems there is a bug. I can not change the properties of `quarkus.oauth2`
**Expected behavior**
The oauth properties should be change in runtime without using profile.
**Actual behavior**
Github Link : https://github.com/berkyvz/quarkus-issues/tree/master/quarkus-oauth2
I have a project with quarkus 1.0.0.Final with gradle. There is .properties file that keeping OAuth properties. The properties shown below;
```properties
quarkus.oauth2.client-id=SampleClientId
quarkus.oauth2.client-secret=secret
quarkus.oauth2.introspection-url= http://localhost:8081/auth/oauth/introspect
```
I am trying to override these properties with using `-D` flags at run time. Actually, It seems like it is changing the properties. However, the truth is it can't change the `quarkus.oauth2.introspection-url` property. I am building project with command `./gradlew clean buildNative --docker-build=true` and run with 
```
./build/quarkus-oauth2-0.0.1-SNAPSHOT-runner -Dquarkus.oauth2.client-id=SampleClientId-native -Dquarkus.oauth2.client-secret=secret-native  -Dquarkus.oauth2.introspection-url=http://abc:8081/auth/oauth/introspect-native -Dquarkus.log.level=DEBUG
```
When I call the controller that simply returns the configuration properties as `HashMap<String,String>`, the reponse is shown below as `JSON`;
```JSON
{
    "quarkus.oauth2.introspection-url": "http://abc:8081/auth/oauth/introspect-native",
    "quarkus.oauth2.client-secret": "secret-native",
    "quarkus.oauth2.client-id": "SampleClientId-native"
}
```
The controller;
__hello/props__
```Java
@ConfigProperty(name = "quarkus.oauth2.client-id")
private String clientId;
	
@ConfigProperty(name = "quarkus.oauth2.client-secret")
private String clientSecret;
	
@ConfigProperty(name = "quarkus.oauth2.introspection-url")
private String introspectionURL;
	
	
@GET
@Path("/props")
@Produces(MediaType.APPLICATION_JSON)
public HashMap<String, String> getProps(){
	HashMap<String, String> props = new HashMap<String, String>();
	props.put("quarkus.oauth2.client-id", clientId);
	props.put("quarkus.oauth2.client-secret", clientSecret);
	props.put("quarkus.oauth2.introspection-url", introspectionURL);
[...]

PR #5937, opened 2019-12-04
Move to RUNTIME_INIT all security related build items

loicmathieu:
Security configuration should be overritable at runtime as it vastly depends on the environement you run inside.

This PR will move the configuration of Oauth2, JDBC and properties file from STATIC to RUNTIME so that native images can override them at runtime like JVM mode does.

### DIFF
--- a/extensions/elytron-security-jdbc/deployment/src/main/java/io/quarkus/elytron/security/jdbc/deployment/ElytronSecurityJdbcProcessor.java
+++ b/extensions/elytron-security-jdbc/deployment/src/main/java/io/quarkus/elytron/security/jdbc/deployment/ElytronSecurityJdbcProcessor.java
@@ -46,7 +46,7 @@ class ElytronSecurityJdbcProcessor {
      * @throws Exception - on any failure
      */
     @BuildStep
-    @Record(ExecutionTime.STATIC_INIT)
+    @Record(ExecutionTime.RUNTIME_INIT)
     void configureJdbcRealmAuthConfig(JdbcRecorder recorder,
             BuildProducer<SecurityRealmBuildItem> securityRealm,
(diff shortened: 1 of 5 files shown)


loicmathieu:
@sberyozkin @stuartwdouglas this PR moves at RUNTIME_INIT elytron-security related steps to be able to configure them at runtime.

I'm not sure it is correct for elytron-security-properties-file, maybe I need to keep it's step at STATIC_INIT (or at least move in a STATIC_INIT step the recording of the properties file) because the properties file needs to be included inside the native-image.

This will fixes #5800 and allow to provides a clean way to implement what needs to be done for #5731

sberyozkin:
@loicmathieu thanks, yes, the properties file should be in the image so it should stay at STATIC_INIT. It probably should be deprecated, the properties approach, as it does not fit into the general Quarkus approach, Stuart @stuartwdouglas, what do you think ?

loicmathieu:
@sberyozkin moved to build time

sberyozkin (approved):
@loicmathieu thanks, please wait for Stuart to comment too in case I've missed something, cheers

stuartwdouglas:
Looks like there is a CI failure?

loicmathieu:
This seems unrelated to the changes. I rebase on master to trigger a new CI, let's wait for it.

sberyozkin:
@loicmathieu Hi Loic, can you please resolve the conflict ?

loicmathieu:
@sberyozkin conflict resolved

loicmathieu:
@stuartwdouglas this has been merged but you didn't aprove it (at least, you didn't make your review as aproved in Github:) ). Did you had time to review it prior to it's merged ?

stuartwdouglas:
yea, I thought it looked fine.

PR #5938, opened 2019-12-04
Allow RuntimeBeanBuildItem to pick static or runtime init

stuartwdouglas:


extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/RuntimeBeanProcessor.java
```diff
@@ -82,6 +85,30 @@ public boolean test(BeanInfo bean) {
...
+    void initStatic(BeanInfoBuildItem infoBuildItem,
+            ArcRecorder recorder) {
+        recorder.initSupplierBeans(infoBuildItem.staticBeans);
+    }
+
+    @BuildStep
+    @Record(ExecutionTime.RUNTIME_INIT)
+    void initRuntime(BeanInfoBuildItem infoBuildItem,
+            ArcRecorder recorder) {
+        recorder.initSupplierBeans(infoBuildItem.runtimeBeans);
```
mkouba: This will overwrite the map for `STATIC_INIT` runtime beans, or?

stuartwdouglas: oops

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/RuntimeBeanProcessor.java
```diff
@@ -82,6 +85,30 @@ public boolean test(BeanInfo bean) {
...
+    }
+
+    @BuildStep
+    @Record(ExecutionTime.RUNTIME_INIT)
+    void initRuntime(BeanInfoBuildItem infoBuildItem,
+            ArcRecorder recorder) {
+        recorder.initRuntimeSupplierBeans(infoBuildItem.runtimeBeans);
+    }
+
+    static final class BeanInfoBuildItem extends SimpleBuildItem {
```
mkouba: I'd like to rename this build item to something like `RuntimeBeanSuppliersBuildItem`. `BeanInfo` is confusing because of `io.quarkus.arc.processor.BeanInfo`.

mkouba (changes requested):
Ideally, we should have a test for this. But I understand that `QuarkusUnitTest` is not quite ready for this kind of stuff.

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/RuntimeBeanProcessor.java
```diff
@@ -82,6 +85,30 @@ public boolean test(BeanInfo bean) {
...
+        final Map<String, Supplier<Object>> staticBeans;
```
gsmet: It's minor but let's call it `staticInitBeans`. Othewise we could have the impression that these beans are static.

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/RuntimeBeanProcessor.java
gsmet: I think it would be better but I would drop the `Runtime` as we also have the static init ones and it could be confusing (I understand you don't mean runtime in this sense but that's a bit confusing IMHO).

In any case, it's local to this class so that's not a big issue.

gsmet (changes requested):
Added some micro nitpicking. Feel free to ignore.

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/RuntimeBeanProcessor.java
mkouba: This build item is package-private so it's probably not a big problem. The `RuntimeBeanBuildItem` is de facto part of the API and the name is confusing. However, 1) we would break backward compatibility and 2) I'm not quite sure what the correct name should be...

mkouba:
Hm, when I think about it there is one last issue with this PR that should be at least documented. If a `RuntimeBean` with `staticInit=false` is used during `STATIC_INIT` the client either receives `null` for `@Dependent` or `IllegalProductException` for any normal scope (because it's implemented as a producer method).

mkouba (approved):

stuartwdouglas:
@gsmet this should be good to merge now.

PR #5944, opened 2019-12-04
Initial token validation support for the issuer and audience claims

### LINKED ISSUE #5829: Quarkus OIDC does not work with Auth0
**Describe the bug**
Given the configuration :
```
quarkus.oidc.auth-server-url=https://myauthtenant.eu.auth0.com/
```
On start-up Quarkus tries to access URL : `https://myauthtenant.eu.auth0.com//.well-known/openid-configuration` (see double //) which causes an exception
```
OIDC server is not available at the 'quarkus.oidc.auth-server-url' URL (...)
(...)
Cannot GET //.well-known/openid-configuration
```
Alternatively, when removing trailing slash:
```
quarkus.oidc.auth-server-url=https://myauthtenant.eu.auth0.com
```
an error is thrown
```
issuer validation failed: received [https://myauthtenant.eu.auth0.com/]
```
as the issuer field in the OIDC JSON has the trailing slash
```
"issuer": "https://myauthtenant.eu.auth0.com/",
```

I've tried adding

```
quarkus.oidc.introspection-path=.well-known/openid-configuration
quarkus.oidc.jwks-path=.well-known/jwks.json
```
with no effect

**Expected behavior**
Quarkus could be smart enough not to add another "/" before `.well-known` when the auth-server-url ends with trailing `/`

**Actual behavior**
Quarkus start up error

**To Reproduce**
1. Create a free Auth0 tenant
2. Configure Quarkus as in https://quarkus.io/guides/security-openid-connect-web-authentication
3. Configure properties to point to Auth0
4. Start app

**Configuration**
```properties
quarkus.oidc.auth-server-url=https://myauthtenant.eu.auth0.com
quarkus.oidc.introspection-path=.well-known/openid-configuration
quarkus.oidc.jwks-path=.well-known/jwks.json
quarkus.oidc.client-id=XXX
quarkus.oidc.credentials.secret=XXX
quarkus.oidc.authentication.scopes=email,profile,roles
```

**Screenshots**

 - Output of `java -version`: 
 - Quarkus version or git rev: 

**Additional context**

sberyozkin:
Fixes #5829

### DIFF
--- a/extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/OidcConfig.java
+++ b/extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/OidcConfig.java
@@ -1,6 +1,7 @@
 package io.quarkus.oidc.runtime;
 
 import java.time.Duration;
+import java.util.Arrays;
 import java.util.List;
 import java.util.Optional;
 
@@ -57,6 +58,12 @@ public class OidcConfig {
      */
     @ConfigItem
     Roles roles;
+
+    /**
+     * Configuration how to validate the token claims.
+     */
+    @ConfigItem
+    Token token;
 
     /**
      * Credentials which the OIDC adapter will use to authenticate to the OIDC server.
@@ -159,5 +166,35 @@ public class OidcConfig {
          */
         @ConfigItem
         public Optional<List<String>> scopes;
+    }
+
+    @ConfigGroup
+    public static class Token {
+
+        /**
+         * Expected issuer 'iss' claim value
+         */
+        @ConfigItem
+        public Optional<String> issuer;
+
+        /**
+         * Expected audience `aud` claim value which may be a string or an array of strings
+         */
+        @ConfigItem
+        public Optional<List<String>> audience;
+
+        public static Token fromIssuer(String issuer) {
+            Token tokenClaims = new Token();
+            tokenClaims.issuer = Optional.of(issuer);
+            tokenClaims.audience = Optional.ofNullable(null);
+            return tokenClaims;
+        }
+
+        public static Token fromAudience(String... audience) {
+            Token tokenClaims = new Token();
+            tokenClaims.issuer = Optional.ofNullable(null);
+            tokenClaims.audience = Optional.of(Arrays.asList(audience));
+            return tokenClaims;
+        }
     }
 }
--- a/extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/OidcIdentityProvider.java
+++ b/extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/OidcIdentityProvider.java
@@ -9,6 +9,7 @@ import org.eclipse.microprofile.jwt.JsonWebToken;
 import org.jose4j.jwt.JwtClaims;
 import org.jose4j.jwt.consumer.InvalidJwtException;
 
+import io.quarkus.oidc.OIDCException;
 import io.quarkus.security.AuthenticationFailedException;
 import io.quarkus.security.ForbiddenException;
 import io.quarkus.security.identity.AuthenticationRequestContext;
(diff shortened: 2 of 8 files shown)


sberyozkin:
@pedroigor, @stuartwdouglas please note that setting an `issuer` property disables the default Vertx validation as per comment at #5829, as the Vertx validation is only centered around the site URL match which can not be sufficient and indeed, going forward it would be good if the users were setting the expected claim values (issuer, audience, etc) themselves.
Ideally we'd get this in for 1.1.0, as it is a basic and isolated new code

extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/OidcConfig.java
```diff
@@ -58,6 +59,12 @@
     @ConfigItem
     Roles roles;
 
+    /**
+     * Configuration how to validate the token claims.
+     */
+    @ConfigItem
+    Claims claims;
```
pedroigor: I'm not sure about `quarkus.oidc.claims` config property. It may clash with the `claims` request parameter as defined by https://openid.net/specs/openid-connect-core-1_0.html#ClaimsParameter.

Maybe, we could use `quarkus.oidc.token.issuer|audience` ?

pedroigor (changes requested):

extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/OidcConfig.java
sberyozkin: @pedroigor Good point, please believe me a `token` qualifier was something I thought for a sec or two about as well :-), because we may want to actually have an option to have both `id token` and `access token` validated separately, right now this validation will affect either the `id token` or `access token` depending on the application type. But in time we may qualify further and introduce an `idtoken` group if needed.
Recall though we already have an `Authentication` group which has `scopes` but will also host more properties related to the redirect, but in any case, it is good to avoid the confusion, so I'll rename shortly
Thanks

sberyozkin: @pedroigor done

sberyozkin:
@gsmet Just downloaded the logs, KC tests are good, I can't even see what is failing :-)

sberyozkin:
@stuartwdouglas @pedroigor please check it again

extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/OidcIdentityProvider.java
```diff
@@ -59,6 +60,13 @@ public void handle(AsyncResult<AccessToken> event) {
                     return;
                 }
                 AccessToken token = event.result();
+                try {
+                    OidcUtils.validateClaims(config.token, token.accessToken());
+                } catch (OIDCException e) {
+                    result.completeExceptionally(new AuthenticationFailedException());
```
stuartwdouglas: This should log at debug level or include the cause.

extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/OidcUtils.java
```diff
@@ -16,14 +18,36 @@ private OidcUtils() {
...
+            String issuer = json.getString(Claims.iss.name());
+            if (!tokenConfig.issuer.get().equals(issuer)) {
+                throw new OIDCException("Invalid issuer");
+            }
+        }
+        if (tokenConfig.audience.isPresent()) {
+            Object claimValue = json.getValue(Claims.aud.name());
+            List<String> audience = Collections.emptyList();
+            if (claimValue instanceof JsonArray) {
+                audience = ((JsonArray) claimValue).stream().map(v -> v.toString()).collect(Collectors.toList());
```
stuartwdouglas: We don't allow lambda in runtime code, they use up a significant amount of additional memory.

stuartwdouglas (dismissed):

extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/OidcIdentityProvider.java
sberyozkin: @stuartwdouglas sure, I forgot to update it to include the cause

sberyozkin: done

extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/OidcUtils.java
sberyozkin: done

gsmet (approved):

PR #5972, opened 2019-12-05
Improve GraalVM detection by asking the version to native-image

gsmet:
Until now we were using the VM name, which might just be a standard JVM
if you're using a standard JVM and just have $GRAALVM_HOME set up.

By calling native-image --version, we really check the proper version.

@cescoffier @stuartwdouglas @gwenneg could you have a look to that one?

### DIFF
--- a/core/deployment/src/main/java/io/quarkus/deployment/pkg/steps/NativeImageBuildStep.java
+++ b/core/deployment/src/main/java/io/quarkus/deployment/pkg/steps/NativeImageBuildStep.java
@@ -73,8 +73,6 @@ public class NativeImageBuildStep {
 
         final String runnerJarName = runnerJar.getFileName().toString();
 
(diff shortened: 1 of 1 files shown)


cescoffier (approved):

gwenneg:
Will this be replaced by the fixed GraalVM Version API when 19.3.1 is released? (assuming it works at that particular place in the Quarkus code which I'm not certain of)

See https://github.com/oracle/graal/issues/1889 for one confirmed issue. There might be another one that I need to confirm.

gsmet:
Well, not really. I, for one, uses a standard JDK to compile and GraalVM only for `native-image` compilation.

I'm pretty sure I'm not the only one. At least Clément's CI setup was similar.

stuartwdouglas:
Lot's of people only use the docker build which does not require any local Graal.

PR #5973, opened 2019-12-05
Introduce YAML configuration extension

### LINKED ISSUE #3904: Add support for YAML configuration files

### LINKED ISSUE #4568: CORSConfig is not representable as YAML
As seen during #4525 development, `CORSConfig` cannot be represented as YAML because we'd need the `quarkus.http.cors` key to be both the key of a key/value pair and the key of a dictionary, which is not possible.

A properties-based CORS filter configuration currently looks like this:
```
quarkus.http.cors=true
quarkus.http.cors.origins=http://foo.com,http://www.bar.io
quarkus.http.cors.methods=GET,PUT,POST
quarkus.http.cors.headers=X-Custom
quarkus.http.cors.exposed-headers=Content-Disposition
```

We only need to change the `quarkus.http.cors=true` property to `quarkus.http.cors.enabled=true` to make it convertible to YAML.

dmlloyd:
Fixes #3904, fixes #4568, fixes #4525.


gastaldi:
Will this also close https://github.com/quarkusio/quarkus/pull/4525 ?

gastaldi (approved):
LGTM

dmlloyd:
Yes, it will.

dmlloyd:
It looks like there's still some relocation or dependency related issue.  Part of the build is grabbing an old version of `smallrye-config` and dying as a result.  Looking into it.

dmlloyd:
TBH it almost looks like it's using a stale BOM.

geoand:
@gsmet any idea what could be going on with the dependencies?

gsmet:
How did you solve the issue with `quarkus.http.cors` in the end?

gsmet:
I have a fix for the build issue. It is a funny one. I will push additional commits to David's branch as soon as the tests are OK.

gsmet:
@dmlloyd so... Kogito, Scala and Kotlin test projects had issues so I fixed them all.

I also reordered the commits (even if GitHub doesn't say so) so that my fixes comes before your commits.

Let's see what CI has to say.

dmlloyd:
Thanks @gsmet!  I couldn't even reproduce the problem locally for some reason.  It might have had something to do with the reactor order or something?

dmlloyd:
The way the CORS style problem was solved is by using a null key when you want an entry for the parent name, so given properties `foo.bar` and `foo.bar.baz`, you do:

```yaml
foo:
    bar:
        ~: fooBarValue
        baz: fooBarBazValue
```

gsmet:
Mkay. Cannot say I'm excited about this :).

This should probably be documented somewhere.

gsmet:
@dmlloyd we will need some documentation for this one before Final so work can be done next week.

PR #5977, opened 2019-12-05
update directory for kotlin guide maven archetype

sherl0cks:
Updating the directory to make the one generated by the archetype.

Also worth noting that I could not get the CDI inject example working, even when adding `<option>all-open:annotation=javax.enterprise.context.ApplicationScoped</option>`. I was getting errors about `GreetingService` being marked as final, which I thought that maven switch would fix. I chose to give up on it.


geoand (approved):
Thanks for spotting and fixing this!

PR #5978, opened 2019-12-06
Use Vert.x BodyHandler

### LINKED ISSUE #5959: Keycloak Claim Information Point - NPE when trying to read body
When you try to read body through Claim Information Point the result is Null Pointer Exception.

When you add the following to your application.properties file:

`quarkus.keycloak.policy-enforcer.claim-information-point.claims.claim-from-body={request.body}`

And execute the post request against the endpoint /api/auth-entry

The result is:
```
ERROR: HTTP Request to /api/auth-entry failed, error id: d73bb86a-aa3e-40bc-9497-7683a11b1a92-1
java.lang.NullPointerException
	at io.quarkus.keycloak.pep.VertxHttpFacade$1.getInputStream(VertxHttpFacade.java:124)
	at org.keycloak.adapters.authorization.util.RequestPlaceHolderResolver.resolve(RequestPlaceHolderResolver.java:107)
	at org.keycloak.adapters.authorization.util.PlaceHolders.parsePlaceHolders(PlaceHolders.java:93)
	at org.keycloak.adapters.authorization.util.PlaceHolders.resolve(PlaceHolders.java:45)

```

Here is the [Sample Project](https://github.com/tpenakov/org-otaibe-enforcer-claim-can-not-read-body)

You can create the quarkus realm from src/test/resources/keycloack/quarkus-realm.json file.

Then you can retrieve the token for user (alice/alice)  as described here: https://quarkus.io/guides/security-keycloak-authorization

Then you have to put the token in org.otaibe.enforcer.claim.can.not.read.body.web.controller.RestControllerTest#TOKEN

If you execute the test org.otaibe.enforcer.claim.can.not.read.body.web.controller.RestControllerTest#testPostEndpoint

Then you will receive the NPE

stuartwdouglas:
Fixes #5959

This change allows the body handler to be used for Undertow
and RESTEasy standalone. When the request is fully buffered
it can be consumed multiple times, which allows keycloak
to also process it.

### DIFF
--- a/extensions/keycloak-authorization/deployment/src/main/java/io/quarkus/keycloak/pep/deployment/KeycloakPolicyEnforcerBuildStep.java
+++ b/extensions/keycloak-authorization/deployment/src/main/java/io/quarkus/keycloak/pep/deployment/KeycloakPolicyEnforcerBuildStep.java
@@ -1,5 +1,7 @@
 package io.quarkus.keycloak.pep.deployment;
 
+import java.util.Map;
+
 import io.quarkus.arc.deployment.AdditionalBeanBuildItem;
 import io.quarkus.arc.deployment.BeanContainerBuildItem;
 import io.quarkus.deployment.annotations.BuildStep;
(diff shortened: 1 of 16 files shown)


extensions/keycloak-authorization/deployment/src/main/java/io/quarkus/keycloak/pep/deployment/KeycloakPolicyEnforcerBuildStep.java
```diff
@@ -21,6 +22,12 @@ FeatureBuildItem featureBuildItem() {
         return new FeatureBuildItem(FeatureBuildItem.KEYCLOAK_AUTHORIZATION);
     }
 
+    @BuildStep
+    RequireBodyHandlerBuildItem requireBody() {
+        //TODO: this should only be produced if required, ask Pedro about config for this
```
stuartwdouglas: @pedroigor how can we tell if the request needs to be fully buffered or not?

extensions/keycloak-authorization/runtime/src/main/java/io/quarkus/keycloak/pep/runtime/VertxHttpFacade.java
```diff
@@ -107,7 +108,15 @@ public Cookie getCookie(String cookieName) {
...
-                return request.getHeader(name);
+                String value = request.getHeader(name);
+
+                if (name.equalsIgnoreCase(HttpHeaders.CONTENT_TYPE.toString())) {
+                    if (value.indexOf(';') != -1) {
+                        return value.substring(0, value.indexOf(';'));
+                    }
+                }
+
+                return value;
```
gsmet: What's the reason of this change? It looks a bit unrelated and unexpected.

stuartwdouglas: That was part of @pedroigor original approach, I don't think it is needed so I might modify the PR to just keep the tests and not the rest of the changes.

pedroigor (approved):

extensions/keycloak-authorization/runtime/src/main/java/io/quarkus/keycloak/pep/runtime/VertxHttpFacade.java
pedroigor: Yeah, the change is not related with this one but to #5959

gsmet: OK, then this needs proper documentation with a TODO pointing to a GH issue so that we remove it in the future.

pedroigor (changes requested):
@stuartwdouglas you need to remove any change related to tests because the previous change I did to content type (VertxHttpFacade) is not here anymore (but will be sent separately on top of your work). Otherwise, the test will fail. The files that should be reverted are: `ProtectedResource`, `application.properties` and `PolicyEnforcerTest`.

gsmet:
Superseded by https://github.com/quarkusio/quarkus/pull/5981 which includes this work.

PR #5987, opened 2019-12-06
BOM - exclude jakarta.ejb-api in the jakarta.interceptor-api dependency

mkouba:
- there is a bug in v1.2.5, should be fixed in the next version


gsmet (approved):

gsmet:
CI error is unrelated.

Issue #5980, opened 2019-12-06
Quarkus shows warning messages for properties that are actually used but not part of config classes


geoand:
**Describe the bug**
It seems like the big config PR removed he list of false positives keys that wouldn't trigger a warning when set.

**Expected behavior**
Setting `-Dquarkus.profile` (as is mentioned for example [here](https://quarkus.io/guides/config#configuration-profiles)) should not trigger the logging of any warning.

**Actual behavior**
Setting `-Dquarkus.profile` yields:

```
2019-12-06 11:48:32,749 WARN  [io.qua.config] (main) Unrecognized configuration key "quarkus.profile" was provided; it will be ignored
```

**To Reproduce**
1. Use any of the Quarkus quickstarts (from the `development` branch).
2. Build
3. Run with `-Dquarkus.profile` set

geoand:
cc @dmlloyd

gastaldi:
Duplicate of https://github.com/quarkusio/quarkus/issues/5855?

geoand:
Yes it is. Although I'm pretty certain that it's not a `Good first issue` :), that is unless I missed something

gastaldi:
According to @stuartwdouglas, the fix is to create a Config object for the missing properties, that's why I marked as a good first issue :)

geoand:
That works for sure. But there are plenty of properties that need to be ignored: https://github.com/quarkusio/quarkus/blob/1.0.1.Final/core/deployment/src/main/java/io/quarkus/deployment/configuration/ConfigDefinition.java#L70

so I'm not sure if adding all those Config objects is the right way to go.

gsmet:
I think it's OK to create a proper config for these. @gastaldi did you plan to work on it? I think we need it for CR so basically today.

If nobody's working on it, I'll do it.

gastaldi:
> did you plan to work on it?

Nope, I'll leave the fun for you this time ;)

geoand:
Having Config objects still won't solve the problem for `quarkus.profile` and `quarkus.version`

gsmet:
@geoand why?

geoand:
What Config object would you use to capture `quarkus.profile`? I guess there is some fundamental piece I am missing :)

gsmet:
Hum. That's a good question.

Maybe a config root called `profile` and then a config item with a `ConfigItem#PARENT` name would work.

[...]

PR #5989, opened 2019-12-06
Allow anonymous config roots

dmlloyd:
Enables #5980

### DIFF
--- a/core/deployment/src/main/java/io/quarkus/deployment/configuration/BuildTimeConfigurationReader.java
+++ b/core/deployment/src/main/java/io/quarkus/deployment/configuration/BuildTimeConfigurationReader.java
@@ -267,7 +267,9 @@ public final class BuildTimeConfigurationReader {
                 }
                 objectsByRootClass.put(clazz, instance);
                 String rootName = root.getRootName();
-                nameBuilder.append('.').append(rootName);
+                if (!rootName.isEmpty()) {
+                    nameBuilder.append('.').append(rootName);
+                }
                 readConfigGroup(root, instance, nameBuilder);
                 nameBuilder.setLength(len);
             }
--- a/core/deployment/src/main/java/io/quarkus/deployment/configuration/matching/PatternMapBuilder.java
+++ b/core/deployment/src/main/java/io/quarkus/deployment/configuration/matching/PatternMapBuilder.java
@@ -19,18 +19,19 @@ public final class PatternMapBuilder {
         for (RootDefinition rootDefinition : rootDefinitions) {
             final String rootName = rootDefinition.getRootName();
             ConfigPatternMap<Container> addTo = patternMap, child;
-            assert !rootName.isEmpty();
-            NameIterator ni = new NameIterator(rootName);
-            assert ni.hasNext();
-            do {
-                final String seg = ni.getNextSegment();
-                child = addTo.getChild(seg);
-                ni.next();
-                if (child == null) {
-                    addTo.addChild(seg, child = new ConfigPatternMap<>());
-                }
-                addTo = child;
-            } while (ni.hasNext());
+            if (!rootName.isEmpty()) {
+                NameIterator ni = new NameIterator(rootName);
+                assert ni.hasNext();
+                do {
+                    final String seg = ni.getNextSegment();
+                    child = addTo.getChild(seg);
+                    ni.next();
+                    if (child == null) {
+                        addTo.addChild(seg, child = new ConfigPatternMap<>());
+                    }
+                    addTo = child;
+                } while (ni.hasNext());
+            }
             addGroup(addTo, rootDefinition, null);
         }
         return patternMap;
--- a/core/deployment/src/main/java/io/quarkus/deployment/configuration/RunTimeConfigurationGenerator.java
+++ b/core/deployment/src/main/java/io/quarkus/deployment/configuration/RunTimeConfigurationGenerator.java
@@ -287,7 +287,7 @@ public final class RunTimeConfigurationGenerator {
             clinit.setModifiers(Opcodes.ACC_STATIC);
             clinit.invokeStaticMethod(PM_SET_RUNTIME_DEFAULT_PROFILE, clinit.load(ProfileManager.getActiveProfile()));
             clinitNameBuilder = clinit.newInstance(SB_NEW);
-            clinit.invokeVirtualMethod(SB_APPEND_STRING, clinitNameBuilder, clinit.load("quarkus."));
+            clinit.invokeVirtualMethod(SB_APPEND_STRING, clinitNameBuilder, clinit.load("quarkus"));
 
             // create the map for build time config source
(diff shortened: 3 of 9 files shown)


geoand (approved):

gsmet:
Thanks @dmlloyd !

core/deployment/src/main/java/io/quarkus/deployment/configuration/matching/PatternMapBuilder.java
```diff
@@ -19,18 +19,19 @@ private PatternMapBuilder() {
...
-                child = addTo.getChild(seg);
-                ni.next();
-                if (child == null) {
-                    addTo.addChild(seg, child = new ConfigPatternMap<>());
-                }
-                addTo = child;
-            } while (ni.hasNext());
+            if (!rootName.isEmpty()) {
+                NameIterator ni = new NameIterator(rootName);
+                assert ni.hasNext();
```
gastaldi: The `assert` keyword is only triggered when `-ea` is specified in the `java` cmd line. Not sure it is a good idea to use it

gastaldi: Perhaps changing the `do...while` loop with a `while` should be enough?

core/deployment/src/main/java/io/quarkus/deployment/configuration/BuildTimeConfigurationReader.java
```diff
@@ -267,7 +267,9 @@ ReadResult run() {
                 }
                 objectsByRootClass.put(clazz, instance);
                 String rootName = root.getRootName();
-                nameBuilder.append('.').append(rootName);
+                if (!rootName.isEmpty()) {
+                    nameBuilder.append('.').append(rootName);
```
geoand: Actually I think that `.append(.)` shouldn't be there since `"quarkus."` is what is added by default. Am I right or am I missing something @dmlloyd ?

geoand: Actually scratch that, I am making looking at wrong piece of code.

However I do see generate code containing:

```
        int var8 = var1.length();
        var1.append('.');
        var1.append("version");
        String var3 = var1.toString();
        Converter var4 = conv$0;
```

inside `initGroup$io$quarkus$runtime$TopLevelRootConfig` when I create a `TopLevelRootConfig`containing `version`

geoand: I think https://github.com/quarkusio/quarkus/blob/bb3a6f47c56fc52598fdbc84fb18ee6ab2f96a00/core/deployment/src/main/java/io/quarkus/deployment/configuration/RunTimeConfigurationGenerator.java#L660 needs to be updated

geoand: I have a quick and dirty fix for this.

core/deployment/src/main/java/io/quarkus/deployment/configuration/matching/PatternMapBuilder.java
dmlloyd: I use assertions as a code comment to show conditions that are always true, and can be relied on without explicit testing.  This assertion has no side-effects.

core/deployment/src/main/java/io/quarkus/deployment/configuration/BuildTimeConfigurationReader.java
dmlloyd: No, I don't think so, unless you can make a test that shows the problem?

geoand: Sure yeah, I'll come up with one once I get back home.

geoand: So here it is: https://github.com/geoand/quarkus/tree/pull/5989

Basically that branch is this PR plus one commit to add a `TopLevelRootConfig` (in https://github.com/geoand/quarkus/commit/4217332e6bad2e2a5156aa45ca6c02b6cbe44305).

If you use that branch and then just run the `main` integration test, everything fails.
When I debugged I found that the generated code was trying to load property `quarkus..profile` instead of `quarkus.profile` which of course doesn't exist.

Does that help?

dmlloyd: Yes! Thanks.

dmlloyd: Fix is pushed.

geoand: Excellent, thanks!

geoand:
Darn, some kind of doc issue with the class that was added:

```
ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.8.0-jboss-2:compile (default-compile) on project quarkus-core: Fatal error compiling: java.nio.file.InvalidPathException: Illegal char <<> at index 8: quarkus-<<parent>>-top-level-root-config.adoc -> [Help 1]
```

Any ideas @gsmet @machi1990 ?

gsmet:
well, it says it all, the doc generation needs to be adjusted.

geoand:
Yeah I figured, the problem is that I've looked at that code before so I'm trying to figure it out in the dark :P

geoand:
I think I have it, I'll give it a try

geoand:
I guess I don't have it after all. I can't even the build to fail locally....

geoand:
I think I have something... @dmlloyd Do you mind if I push to your PR?

geoand:
Right.... no wonder I can't reproduce... the failure is only on Windows :man_facepalming: . Of course the docs generation still needs to be fixed

gsmet:
If nobody beats me to it, I’ll have a look tomorrow. 

@geoand could you push the warning thing somewhere? I’ll create an aggregated PR.

geoand:
Well l have a fix but I don't have anyway of testing it for one and it could very well be the wrong fix since I've never seen how the documentation is produced.

I'll push the config warning fixes to another branch and let you know @gsmet

gsmet:
Well, as long as you have dropped the <<parent>> part, it should work ok.

geoand:
> Well, as long as you have dropped the <<parent>> part, it should work ok.

That is the only thing I did :P

geoand:
@gsmet  here is the branch were I pushed some fixes: https://github.com/geoand/quarkus/tree/pull/5989%2B5980

gsmet:
Superseded by https://github.com/quarkusio/quarkus/pull/6005 .

PR #5990, opened 2019-12-06
Log a warning if Arc finds observer for @Initialized(ApplicationScope…

### LINKED ISSUE #5946: Log a big warning if an application bean observes @Initialized(ApplicationScoped.class)
There is a big difference in behavior when working with native images. More details are explained in the guide: https://quarkus.io/guides/lifecycle#what-is-the-difference-from-initializedapplicationscoped-class-and-destroyedapplicationscoped-class.

manovotn:
Fixes #5946 

This will give you following warning in the console on build:
`WARN: The method io.my.app.ContextObserver#observeContextInit is an observer for @Initialized(ApplicationScoped.class). Observer notification for this event may vary between JVM and native modes! We strongly recommend to observe StartupEvent instead as that one is consistently delivered in both modes once the container is running.`

@mkouba I am open to better WARN message suggestions; just couldn't come up with anything short yet explanatory enough. We could also just link to the docs we have, but I kinda don't like doing links from code to website...

### DIFF
--- a/extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/ObserverValidationProcessor.java
+++ b/extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/ObserverValidationProcessor.java
@@ -1,0 +1,59 @@
+package io.quarkus.arc.deployment;
+
+import java.util.Collection;
+
+import org.jboss.jandex.AnnotationInstance;
+import org.jboss.jandex.DotName;
+import org.jboss.jandex.IndexView;
+import org.jboss.logging.Logger;
+
+import io.quarkus.arc.processor.Annotations;
+import io.quarkus.arc.processor.BeanDeploymentValidator;
+import io.quarkus.arc.processor.BuiltinScope;
+import io.quarkus.arc.processor.DotNames;
+import io.quarkus.arc.processor.ObserverInfo;
+import io.quarkus.deployment.annotations.BuildProducer;
+import io.quarkus.deployment.annotations.BuildStep;
+import io.quarkus.deployment.builditem.ApplicationArchivesBuildItem;
+
+/**
+ * Validates observer methods from application classes.
+ * If an observer listening for {@code @Initialized(ApplicationScoped.class)} is found, it logs a warning.
+ */
+public class ObserverValidationProcessor {
+
+    private static final Logger LOGGER = Logger.getLogger(ObserverValidationProcessor.class.getName());
+
+    @BuildStep
+    public void validateApplicationObserver(ApplicationArchivesBuildItem applicationArchivesBuildItem,
+            BuildProducer<BeanDeploymentValidatorBuildItem> validators) {
+        // an index of all root archive classes (usually src/main/classes)
+        IndexView applicationClassesIndex = applicationArchivesBuildItem.getRootArchive().getIndex();
+
+        validators.produce(new BeanDeploymentValidatorBuildItem(new BeanDeploymentValidator() {
+
+            @Override
+            public void validate(ValidationContext context) {
+                Collection<ObserverInfo> allObservers = context.get(Key.OBSERVERS);
+                // do the validation for each observer that can be found within application classes
+                for (ObserverInfo observer : allObservers) {
+                    DotName declaringBeanDotName = observer.getDeclaringBean().getBeanClass();
+                    AnnotationInstance instance = Annotations.getParameterAnnotation(observer.getObserverMethod(),
+                            DotNames.INITIALIZED);
+                    if (applicationClassesIndex.getClassByName(declaringBeanDotName) != null && instance != null &&
+                            instance.value().asClass().name().equals(BuiltinScope.APPLICATION.getName())) {
+                        // found an observer for @Initialized(ApplicationScoped.class)
+                        // log a warning and recommend to use StartupEvent instead
+                        final String observerWarning = "The method %s#%s is an observer for " +
+                                "@Initialized(ApplicationScoped.class). Observer notification for this event may " +
+                                "vary between JVM and native modes! We strongly recommend to observe StartupEvent " +
+                                "instead as that one is consistently delivered in both modes once the container is " +
+                                "running.";
+                        LOGGER.warnf(observerWarning, observer.getDeclaringBean().getImplClazz(),
+                                observer.getObserverMethod().name());
+                    }
+                }
+            }
+        }));
+    }
+}
--- a/independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/DotNames.java
+++ b/independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/DotNames.java
@@ -7,6 +7,7 @@ import java.util.Optional;
 import javax.annotation.PostConstruct;
 import javax.annotation.PreDestroy;
 import javax.annotation.Priority;
+import javax.enterprise.context.Initialized;
 import javax.enterprise.context.control.ActivateRequestContext;
 import javax.enterprise.event.Event;
 import javax.enterprise.event.Observes;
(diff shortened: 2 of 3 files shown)


independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/ObserverInfo.java
```diff
@@ -143,6 +145,17 @@ MethodParameterInfo initEventParam(MethodInfo observerMethod) {
                     LOGGER.warnf(transactionalObserverWarning, clazz, observerMethod.name(), during.asEnum());
                 }
             }
+            if (annotation.name().equals(DotNames.create(Initialized.class))
```
mkouba: I think that we should only log this message for app classes, ie. classes that are under the control of the user. We do this for private members as well.

Also I think that we should handle this in a build step in the `ArcProcessor`.

manovotn:
Alright, second attempt.
I've scratched the previous one and turned it into a separate validation processor based on `BeanDeploymentValidatorBuildItem`. It should also check if the given declaring observer class is within index that contains app classes and only then perform validation.
The log message stays the same for now.

Is this any better?

manovotn:
Also, no test as this is purely about logging a warning, though locally I've adapted the `AsyncObserverTest` to check if the warning is correctly emitted.

independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/DotNames.java
```diff
@@ -81,6 +83,8 @@
     public static final DotName NAMED = create(Named.class);
     public static final DotName ACTIVATE_REQUEST_CONTEXT = create(ActivateRequestContext.class);
     public static final DotName TRANSACTION_PHASE = create(TransactionPhase.class);
+    public static final DotName INITIALIZED = create(Initialized.class);
+    public static final DotName APPLICATION_SCOPED = create(ApplicationScoped.class);
```
mkouba: You can use `BuiltinScope.APPLICATION.getName()` instead...

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/ObserverValidationProcessor.java
```diff
@@ -0,0 +1,63 @@
...
+
+        validators.produce(new BeanDeploymentValidatorBuildItem(new BeanDeploymentValidator() {
+
+            @Override
+            public void validate(ValidationContext context) {
+                Collection<ObserverInfo> allObservers = context.get(Key.OBSERVERS);
+                // do the validation for each observer that can be found within application classes
+                for (ObserverInfo observer : allObservers) {
+                    DotName declaringBeanDotName = observer.getDeclaringBean().getBeanClass();
+                    AnnotationInstance annotationInstance = observer.getObserverMethod().annotation(DotNames.INITIALIZED);
```
mkouba: Just a little detail - `observer.getObserverMethod().annotation(DotNames.INITIALIZED)` will return a non-null value even if the annotation is accidentally declared on the method...

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/ObserverValidationProcessor.java
```diff
@@ -0,0 +1,63 @@
...
+/**
+ * Validates observer methods from application classes.
+ * If an observer listening for {@code @Initialized(ApplicationScoped.class)} is found, it logs a warning.
+ */
+public class ObserverValidationProcessor {
+
+    private static final Logger LOGGER = Logger.getLogger(ObserverValidationProcessor.class.getName());
+
+    @Inject
+    BuildProducer<BeanDeploymentValidatorBuildItem> validators;
```
mkouba: There is no need to use an injection field for a producer. I'd just add the `BuildProducer<BeanDeploymentValidatorBuildItem> validators` next to the `applicationArchivesBuildItem` parameter.

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/ObserverValidationProcessor.java
```diff
@@ -0,0 +1,63 @@
...
+            @Override
+            public void validate(ValidationContext context) {
+                Collection<ObserverInfo> allObservers = context.get(Key.OBSERVERS);
+                // do the validation for each observer that can be found within application classes
+                for (ObserverInfo observer : allObservers) {
+                    DotName declaringBeanDotName = observer.getDeclaringBean().getBeanClass();
+                    AnnotationInstance annotationInstance = observer.getObserverMethod().annotation(DotNames.INITIALIZED);
+                    if (applicationClassesIndex.getClassByName(declaringBeanDotName) != null &&
+                            annotationInstance != null &&
+                            annotationInstance.value().kind().equals(AnnotationValue.Kind.CLASS) &&
```
mkouba: This check is probably superfluous. If it's not `Kind.CLASS` then it's an error.

mkouba (changes requested):

independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/DotNames.java
manovotn: Good point, I forgot it's there.

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/ObserverValidationProcessor.java
```diff
@@ -0,0 +1,63 @@
...
+
+    private static final Logger LOGGER = Logger.getLogger(ObserverValidationProcessor.class.getName());
+
+    @Inject
+    BuildProducer<BeanDeploymentValidatorBuildItem> validators;
+
+    @BuildStep
+    public void validateApplicationObserver(ApplicationArchivesBuildItem applicationArchivesBuildItem) {
+        // an index of all application classes
+        IndexView applicationClassesIndex = applicationArchivesBuildItem.getRootArchive().getIndex();
```
mkouba: Actually, this index does not contain all the application classes but the classes from the root archive (usually `src/main/classes`). It does not contain classes from other modules in a multi-module project. But I think that we have the same issue with private injection points. So let's keep it this way for now...

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/ObserverValidationProcessor.java
manovotn: Yeah I know, will do that if it'd preferable.

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/ObserverValidationProcessor.java
manovotn: Hmm, I don't think there is a way to query for just method param annotations.
Therefore, to avoid situations where you accidentally declare it on method and on method param, we should probably get all annotations and iterate over them?

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/ObserverValidationProcessor.java
manovotn: > probably

That was my thinking exactly, I was just paranoid ;-)

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/ObserverValidationProcessor.java
manovotn: Hmm, didn't know that. If you know a better way, I am all ears.
For now, I'll at least adjust the javadoc.

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/ObserverValidationProcessor.java
mkouba: Yes, there is no such method in the API. And yes, you'll need to filter the annotation instances by target. Something like https://github.com/quarkusio/quarkus/blob/master/independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/Annotations.java#L54-L63.

mkouba: BTW feel free to add a utility method to the `Annotations` class.

manovotn: Hm, pushed changes before I read these comments, I'll add the util method too.

mkouba (approved):

PR #5999, opened 2019-12-07
Remove Override annotation from generated bytecode

### LINKED ISSUE #5682: Support GraalVM / native image builds using the GraalVM Java 11 build

gwenneg:
Fixes one of the issues mentioned in #5682.

### DIFF
--- a/core/deployment/src/main/java/io/quarkus/deployment/configuration/RunTimeConfigurationGenerator.java
+++ b/core/deployment/src/main/java/io/quarkus/deployment/configuration/RunTimeConfigurationGenerator.java
@@ -578,10 +578,9 @@ public final class RunTimeConfigurationGenerator {
             // generate run time default values config source class
             try (ClassCreator dvcc = ClassCreator.builder().classOutput(classOutput).className(RTDVCS_CLASS_NAME)
                     .superClass(AbstractRawDefaultConfigSource.class).setFinal(true).build()) {
+                // implements abstract method AbstractRawDefaultConfigSource#getValue(NameIterator)
                 try (MethodCreator mc = dvcc.getMethodCreator("getValue", String.class, NameIterator.class)) {
                     final ResultHandle keyIter = mc.getMethodParam(0);
-                    // implements abstract method AbstractRawDefaultConfigSource#getValue(NameIterator)
-                    mc.addAnnotation(Override.class);
                     final MethodDescriptor md = generateDefaultValueParse(dvcc, runTimePatternMap,
                             new StringBuilder("getDefaultFor"));
                     if (md != null) {
@@ -606,10 +605,9 @@ public final class RunTimeConfigurationGenerator {
             // generate build time run time visible default values config source class
             try (ClassCreator dvcc = ClassCreator.builder().classOutput(classOutput).className(BTRTDVCS_CLASS_NAME)
                     .superClass(AbstractRawDefaultConfigSource.class).setFinal(true).build()) {
+                // implements abstract method AbstractRawDefaultConfigSource#getValue(NameIterator)
                 try (MethodCreator mc = dvcc.getMethodCreator("getValue", String.class, NameIterator.class)) {
                     final ResultHandle keyIter = mc.getMethodParam(0);
-                    // implements abstract method AbstractRawDefaultConfigSource#getValue(NameIterator)
-                    mc.addAnnotation(Override.class);
                     final MethodDescriptor md = generateDefaultValueParse(dvcc, buildTimeRunTimePatternMap,
                             new StringBuilder("getDefaultFor"));
                     if (md != null) {


gwenneg:
For now, I'm submitting this PR to see if CI is happy with it. There's no reason it wouldn't be but I want to make sure of it. In the meantine, I need to test this fix with all the Quarkus native integration. I'll explain why this fix is required once I'm sure it's the good one.

gwenneg:
The CI failure is unrelated.

jaikiran:
Hello @gwenneg, I'm guessing you did this change to get past https://github.com/oracle/graal/issues/1941? If so, I think this is the correct fix and not a workaround :) The `Override` is a `SOURCE` retention type annotation and shouldn't end up in the generated bytecode.

Of course, I think this should be handled by Gizmo. @stuartwdouglas, it appears that the `addAnnotation` APIs exposed by Gizmo assume/expect that the annotation type is a "visible" at runtime and thus should be recorded in the bytecode? If that's the case then maybe wherever possible, like here, where we have access to the annotation type, Gizmo should probably throw an error if the retention type is `SOURCE`? Of course, it won't cover cases where the `addAnnotation` is called using a `String` value of the annotation type, but in such cases we perhaps make it an expectation (through javadoc maybe) that the caller is responsible for doing the necessary validation?

gwenneg:
You're guessing right @jaikiran and thanks for the Gizmo part analysis! I agree it would be great to improve Gizmo and avoid such situation whenever possible.

I had to debug the GraalVM native image generation and analyze a lot of call stacks including tons of `Proxy.newProxyInstance` before I was able to trace the issue back to Quarkus. It was a long journey for a 2 lines PR :)

My tests were delayed by the broken CI issue, I just started the full Quarkus native build locally. I also rebased this PR on master so that it is built using the Quarkus JDK 8 CI. We might be ready for GraalVM with JDK 11 after this one. I'll update the PR to simply remove the two `mc.addAnnotation` lines and promote it when my local tests are done.

cc @dmlloyd since the changes affect the new configuration code.

gsmet:
I think I would like to get this one into CR1 so if you can get it done today or tomorrow, that would be nice.

Thanks!

gwenneg:
@gsmet: How would you like to communicate in the doc about the GraalVM + JDK 11 support in Quarkus? Since CI doesn't run JDK 11 native tests for now, I suppose we can't just write that we provide a production-grade support. Should it be qualified as `preview` in the doc instead?

gwenneg:
I rephrased the commit and PR title as this was not caused by a GraalVM bug even if I believe GraalVM could do a better job at detecting this (it will be discussed in https://github.com/oracle/graal/issues/1941).

I also promoted this PR since I confirmed locally that the fix works with several Quarkus native tests modules. I feel like it can be merged once CI is happy with this latest version.

My last point is about the full Quarkus native build: I'm experiencing difficulties running all the native tests because the native image generation takes much more time with JDK 11 than with JDK 8. This is not a problem with "small" Quarkus integration-test modules which build fine with JDK 11, but it is a problem with larger modules because the native image isn't started fast enough. The computer I'm using is certainly contributing to this global slowness as it is not very recent.

dmlloyd (approved):

gsmet (approved):

gwenneg:
The `quarkus-integration-test-main` native tests just failed locally with an unrelated error:

```
Error: No instances of java.io.FilePermission are allowed in the image heap as this class should be initialized at image runtime. To see how this object got instantiated use -H:+TraceClassInitialization.
Detailed message:
Trace: 	object java.util.concurrent.ConcurrentHashMap$Node
	object java.util.concurrent.ConcurrentHashMap$Node[]
	object java.util.concurrent.ConcurrentHashMap
	object java.io.FilePermissionCollection
	object java.util.concurrent.ConcurrentHashMap$Node
	object java.util.concurrent.ConcurrentHashMap$Node[]
	object java.util.concurrent.ConcurrentHashMap
	object java.security.Permissions
	object java.security.ProtectionDomain
	object java.security.ProtectionDomain[]
	object java.security.AccessControlContext
	object java.util.concurrent.Executors$PrivilegedThreadFactory
	object io.smallrye.faulttolerance.DefaultHystrixConcurrencyStrategy
	object io.quarkus.arc.impl.InstanceHandleImpl
	object java.lang.Object[]
	object java.util.ArrayList
	object java.util.Collections$SynchronizedRandomAccessList
	object io.quarkus.arc.impl.CreationalContextImpl
	object io.quarkus.arc.impl.InstanceImpl
	object io.smallrye.faulttolerance.HystrixInitializer
	object io.quarkus.arc.impl.ContextInstanceHandleImpl
	object io.quarkus.arc.impl.LazyValue
	object java.util.concurrent.ConcurrentHashMap$Node
	object java.util.concurrent.ConcurrentHashMap$Node[]
	object java.util.concurrent.ConcurrentHashMap
	object io.quarkus.arc.impl.ComputingCache
	object io.quarkus.arc.impl.ApplicationContext
	object io.quarkus.arc.impl.ArcContainerImpl
	object java.util.concurrent.atomic.AtomicReference
	method io.quarkus.arc.Arc.shutdown()
Call path from entry point to io.quarkus.arc.Arc.shutdown(): 
	at io.quarkus.arc.Arc.shutdown(Arc.java:45)
	at io.quarkus.arc.runtime.ArcRecorder$1.run(ArcRecorder.java:38)
	at com.oracle.svm.core.jdk.RuntimeSupport.executeHooks(RuntimeSupport.java:144)
	at com.oracle.svm.core.jdk.RuntimeSupport.executeStartupHooks(RuntimeSupport.java:89)
	at com.oracle.svm.core.JavaMainWrapper.runCore(JavaMainWrapper.java:143)
	at com.oracle.svm.core.JavaMainWrapper.run(JavaMainWrapper.java:186)
	at com.oracle.svm.core.code.IsolateEnterStub.JavaMainWrapper_run_5087f5482cc9a6abc971913ece43acb471d2631b(generated:0)
```
I'll investigate this and submit another PR as soon as I have a fix for it (before the end of the day hopefully).

gsmet:
Merged, nice detective work!

stuartwdouglas:
I have updated Gizmo: https://github.com/quarkusio/gizmo/commit/478fdb2ea6961325cf9dde719ff3f5d449ace431

PR #6030, opened 2019-12-09
Properly pass Kotlin compiler plugin options to dev-mode

### LINKED ISSUE #2811: Recompilation kotlin sources in dev mode should use compiler plugins
**Description**
Using quarkus 0.16.0 in a kotlin project, when hot reloading is triggered it seems like the kotlin files are recompiled without any kotlin compiler plugins. Consider a resource like the following

    @Path("/greeting")
    @ApplicationScoped
    class GreetingResource @Inject constructor(val greetingService: GreetingService) { ... }

In order for this class to work correctly, it needs to not be final and to have a default constructor. Both can be achieved by configuring the maven kotlin-maven-plugin

    <configuration>
        <compilerPlugins>
            <plugin>all-open</plugin>
            <plugin>no-arg</plugin>
        </compilerPlugins>
        <pluginOptions>
            <option>all-open:annotation=javax.enterprise.context.ApplicationScoped</option>
            <option>no-arg:annotation=javax.enterprise.context.ApplicationScoped</option>
        </pluginOptions>
    </configuration>

And indeed the application is working when initially starting with

    mvn compile quarkus:dev

But as soon as the source is modified the reload fails with

    Caused by: javax.enterprise.inject.spi.DeploymentException: Found 2 deployment problems: 
    [1] Normal scoped bean must not be final: CLASS bean [types=[GreetingResource, java.lang.Object], qualifiers=[@Default, @Any], target=GreetingResource]
    [2] Normal scoped beans must declare a non-private constructor with no parameters: CLASS bean [types=[GreetingResource, java.lang.Object], qualifiers=[@Default, @Any], target=GreetingResource]

**Implementation ideas**
Ideally the kotlin extension would reuse the configuration of the kotlin-maven-plugin from pom.xml. Alternatively there would need to be a separate configuration option for plugins of the quarkus-kotlin extension.

Issue #1157 is related, but this issue is actually more generic as there might be more classes or annotations that  need special treatment by the kotlin compiler, which would then break after hot reloading.

geoand:
Partially fixes: #2811 (Gradle support will be part of #6035)

### DIFF
--- a/bom/runtime/pom.xml
+++ b/bom/runtime/pom.xml
@@ -2263,6 +2263,11 @@
             </dependency>
             <dependency>
                 <groupId>org.jetbrains.kotlin</groupId>
+                <artifactId>kotlin-compiler</artifactId>
+                <version>${kotlin.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>org.jetbrains.kotlin</groupId>
                 <artifactId>kotlin-stdlib-jdk8</artifactId>
                 <version>${kotlin.version}</version>
             </dependency>
--- a/bom/runtime/pom.xml
+++ b/bom/runtime/pom.xml
@@ -2258,11 +2258,6 @@
 
             <dependency>
                 <groupId>org.jetbrains.kotlin</groupId>
-                <artifactId>kotlin-compiler-embeddable</artifactId>
-                <version>${kotlin.version}</version>
-            </dependency>
-            <dependency>
-                <groupId>org.jetbrains.kotlin</groupId>
                 <artifactId>kotlin-compiler</artifactId>
                 <version>${kotlin.version}</version>
             </dependency>
--- a/core/devmode/src/main/java/io/quarkus/dev/DevModeContext.java
+++ b/core/devmode/src/main/java/io/quarkus/dev/DevModeContext.java
@@ -36,6 +36,9 @@ public class DevModeContext implements Serializable {
     private String sourceJavaVersion;
     private String targetJvmVersion;
 
+    private List<String> compilerPluginArtifacts;
+    private List<String> compilerPluginsOptions;
+
     public List<URL> getClassPath() {
         return classPath;
     }
@@ -120,6 +123,22 @@ public class DevModeContext implements Serializable {
         this.targetJvmVersion = targetJvmVersion;
     }
 
+    public List<String> getCompilerPluginArtifacts() {
+        return compilerPluginArtifacts;
+    }
+
+    public void setCompilerPluginArtifacts(List<String> compilerPluginArtifacts) {
+        this.compilerPluginArtifacts = compilerPluginArtifacts;
+    }
+
+    public List<String> getCompilerPluginsOptions() {
+        return compilerPluginsOptions;
+    }
+
+    public void setCompilerPluginsOptions(List<String> compilerPluginsOptions) {
+        this.compilerPluginsOptions = compilerPluginsOptions;
+    }
+
     public File getDevModeRunnerJarFile() {
         return devModeRunnerJarFile;
     }
@@ -173,5 +192,4 @@ public class DevModeContext implements Serializable {
             return resourcePath;
         }
     }
-
 }
--- a/devtools/maven/src/main/java/io/quarkus/maven/DevMojo.java
+++ b/devtools/maven/src/main/java/io/quarkus/maven/DevMojo.java
@@ -31,7 +31,11 @@ import java.util.stream.Collectors;
 import java.util.zip.ZipEntry;
 import java.util.zip.ZipOutputStream;
 
+import org.apache.maven.artifact.Artifact;
+import org.apache.maven.artifact.resolver.ArtifactResolutionRequest;
+import org.apache.maven.artifact.resolver.ArtifactResolutionResult;
 import org.apache.maven.execution.MavenSession;
+import org.apache.maven.model.Dependency;
 import org.apache.maven.model.Plugin;
 import org.apache.maven.plugin.AbstractMojo;
 import org.apache.maven.plugin.MojoExecutionException;
@@ -179,6 +183,9 @@ public class DevMojo extends AbstractMojo {
 
     @Component
     private RepositorySystem repoSystem;
+
+    @Component
+    protected org.apache.maven.repository.RepositorySystem system;
 
     @Component
     private Invoker invoker;
@@ -429,6 +436,7 @@ public class DevMojo extends AbstractMojo {
 
     class DevModeRunner {
 
+        private static final String KOTLIN_MAVEN_PLUGIN_GA = "org.jetbrains.kotlin:kotlin-maven-plugin";
         private final List<String> args;
         private Process process;
         private Set<Path> pomFiles = new HashSet<>();
@@ -511,6 +519,8 @@ public class DevMojo extends AbstractMojo {
                     break;
                 }
             }
+
+            setKotlinSpecificFlags(devModeContext);
 
             final AppModel appModel;
             try {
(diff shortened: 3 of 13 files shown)


geoand:
I tested this and it worked great on a generated project. I also updated the Kotlin integration test to reflect the test case it's supposed to solve.

Feedback more than welcome :)

geoand:
@ia3andy this doesn't fix your Java 11 issue with Kotlin, but it should make it a lot easier for you or anyone else to implement it. Did you ever open an issue for that?

core/devmode/src/main/java/io/quarkus/dev/DevModeContext.java
```diff
@@ -120,6 +123,14 @@ public void setTargetJvmVersion(String targetJvmVersion) {
         this.targetJvmVersion = targetJvmVersion;
     }
 
+    public List<String> getCompilerPluginArtifacts() {
+        return compilerPluginArtifacts;
+    }
+
+    public List<String> getCompilerPluginsOptions() {
+        return compilerPluginsOptions;
+    }
```
gsmet: I would prefer having setters for those two. Similar to `setCompilerOptions`.

extensions/kotlin/deployment/pom.xml
```diff
@@ -31,7 +31,7 @@
 
         <dependency>
             <groupId>org.jetbrains.kotlin</groupId>
-            <artifactId>kotlin-compiler-embeddable</artifactId>
+            <artifactId>kotlin-compiler</artifactId>
```
gsmet: Why that change?

devtools/maven/src/main/java/io/quarkus/maven/DevMojo.java
```diff
@@ -638,6 +648,49 @@ void prepare() throws Exception {
...
+            }
+
+            getLog().debug("Kotlin Maven plugin detected");
+
+            // get the absolute path of the all-open plugin
+            String fullPath = null;
+            List<Dependency> dependencies = kotlinMavenPlugin.getDependencies();
+            for (Dependency dependency : dependencies) {
+                // TODO do we actually need this to resolve the artifact or can we use the repoSystem?
+                Artifact artifact = system.createDependencyArtifact(dependency);
```
geoand: I would like if I didn't have to add `system` to `DevMojo` and just use `repoSystem`. Is that possible @aloubyansky ?

What I added here I pretty much copied from [here](https://github.com/JetBrains/kotlin/blob/v1.3.41/libraries/tools/kotlin-maven-plugin/src/main/java/org/jetbrains/kotlin/maven/KotlinCompileMojoBase.java#L294)

gsmet (commented):
I added a suggestion and a question.

extensions/kotlin/deployment/pom.xml
geoand: This is needed otherwise the Kotlin compiler plugins can't integrate with the compiler (you get some kind of incompatibility exception).
FWIW, we probably should have had this dependency from the start.

core/devmode/src/main/java/io/quarkus/dev/DevModeContext.java
geoand: OK, I'll update

geoand: Fixed

geoand:
@gsmet I updated per your suggestion and added an answer to your question

devtools/maven/src/main/java/io/quarkus/maven/DevMojo.java
aloubyansky: Yes, sure, here is how https://github.com/quarkusio/quarkus/blob/master/independent-projects/bootstrap/core/src/main/java/io/quarkus/bootstrap/resolver/maven/MavenArtifactResolver.java#L197

geoand: Thanks! I'll update the PR to use this

bom/runtime/pom.xml
```diff
@@ -2261,6 +2261,11 @@
                 <artifactId>kotlin-compiler-embeddable</artifactId>
                 <version>${kotlin.version}</version>
             </dependency>
+            <dependency>
+                <groupId>org.jetbrains.kotlin</groupId>
+                <artifactId>kotlin-compiler</artifactId>
+                <version>${kotlin.version}</version>
+            </dependency>
```
aloubyansky: Do we still need `kotlin-compiler-embeddable`? AFAICS, it's used only in `kotlin/deployment/pom.xml`. Which also raises a question why it is in the runtime BOM instead of the deployment one.

geoand: Nice catches, thanks! 
I think we can remove `embeddable`. I'll also check and see if moving it to the deployment BOM works.

geoand: Actually, are you sure it should in the deployment BOM? I don't see any other non-quarkus artifacts thee

geoand: I removed the embeddable dependency

devtools/maven/src/main/java/io/quarkus/maven/DevMojo.java
geoand: Fixed

bom/runtime/pom.xml
aloubyansky: > Actually, are you sure it should in the deployment BOM? I don't see any other non-quarkus artifacts there

The runtime BOM is for runtime dependencies. It is also imported directly by user applications to align application runtime dependencies with the Quarkus platform's runtime dependencies.
Theoretically, if the artifact is only a deployment dependency then it belongs to the deployment BOM. One caveat in the current implementation (and afaict in the to be implemented classloading model) if there are difference versions of the artifact in the runtime CP and the deployment CP then the runtime one will end up being used during the augmentation.
In this given case, however, I'd probably put it in the deployment BOM. If you are not comfortable doing that then don't :) Could be a separate issue.
The BOMs should probably go through a thorough review.

aloubyansky (approved):

bom/runtime/pom.xml
geoand: I did a test that moved the dependency to the deployment BOM and as you expected it worked. The only caveat however is that we need to redeclare the kotlin version and keep it in sync with the runtime BOM.
Should we go for that or keep it as it is?

aloubyansky: Ah, the properties. I think the properties should move to the root pom.xml so that they can be shared across all the subprojects including the tests. Some tests define/duplicate those versions, which is not a good thing.

geoand: @aloubyansky makes sense!

@gsmet I think that what @aloubyansky proposes should be handled in a dedicated issue, WDYT?

gsmet: I fully agree about having the properties in their own pom. I suggested that a while ago.

Not entirely sure it will work to have them in the root pom, depends if all the pom requiring them has it as parent. Otherwise, it can be a specific pom.

I agree it should be done in a separate issue.

gsmet (approved):

bom/runtime/pom.xml
geoand: Thanks both of you for your input!

gastaldi:
Looks good. Haven't fully tested yet, but does it work in Gradle?

geoand:
> Looks good. Haven't fully tested yet, but does it work in Gradle?

Nope, it will require a lot of work to get it working in Gradle as well. That is because in Gradle we can't just read the configuration, we'll need to read the Kotlin Gradle plugin data. It should definitely be done when all the other Gradle improvements things are in place.

geoand:
@gastaldi would you be interested in doing a follow up for Gradle as some point?

bom/runtime/pom.xml
aloubyansky: > Not entirely sure it will work to have them in the root pom, depends if all the pom requiring them has it as parent. Otherwise, it can be a specific pom.

Either way that properties pom will have to be a parent pom for every subproject. Which is what the root pom is supposed to be, otherwise the hierarchy for the project is going to break which effectively means independent projects.

gastaldi:
I created #6035 And referenced it in #5101

gastaldi (approved):

geoand:
> I created #6035 And referenced it in #5101

Excellent!

PR #6031, opened 2019-12-09
Read from the process before waiting for exit

stuartwdouglas:
If you do not read from the Process InputStream is it
possible that the process will block if the process buffer
is too small to contain the full output.

There is no need to call waitFor as reading the InputStream
will have the same effect.

### DIFF
--- a/core/deployment/src/main/java/io/quarkus/deployment/pkg/steps/NativeImageBuildStep.java
+++ b/core/deployment/src/main/java/io/quarkus/deployment/pkg/steps/NativeImageBuildStep.java
@@ -134,15 +134,15 @@ public class NativeImageBuildStep {
             nativeImage = Collections.singletonList(getNativeImageExecutable(graal, java, env).getAbsolutePath());
         }
 
-        Optional<String> graalVMVersion = Optional.empty();
+        final Optional<String> graalVMVersion;
 
         try {
             List<String> versionCommand = new ArrayList<>(nativeImage);
             versionCommand.add("--version");
 
             Process versionProcess = new ProcessBuilder(versionCommand.toArray(new String[0]))
+                    .redirectErrorStream(true)
                     .start();
-            versionProcess.waitFor();
             try (BufferedReader reader = new BufferedReader(new InputStreamReader(versionProcess.getInputStream()))) {
                 graalVMVersion = reader.lines().filter((l) -> l.startsWith("GraalVM Version")).findFirst();
             }


stuartwdouglas:
Windows failure appears to be due to https://github.com/quarkusio/quarkus/pull/6032

gsmet (approved):

core/deployment/src/main/java/io/quarkus/deployment/pkg/steps/NativeImageBuildStep.java
```diff
@@ -134,15 +134,15 @@ public NativeImageBuildItem build(NativeConfig nativeConfig, NativeImageSourceJa
             nativeImage = Collections.singletonList(getNativeImageExecutable(graal, java, env).getAbsolutePath());
         }
 
-        Optional<String> graalVMVersion = Optional.empty();
+        final Optional<String> graalVMVersion;
```
gsmet: Ah yes, thanks, this was coming from the first version of the patch.

core/deployment/src/main/java/io/quarkus/deployment/pkg/steps/NativeImageBuildStep.java
```diff
@@ -134,15 +134,15 @@ public NativeImageBuildItem build(NativeConfig nativeConfig, NativeImageSourceJa
...
 
         try {
             List<String> versionCommand = new ArrayList<>(nativeImage);
             versionCommand.add("--version");
 
             Process versionProcess = new ProcessBuilder(versionCommand.toArray(new String[0]))
+                    .redirectErrorStream(true)
                     .start();
-            versionProcess.waitFor();
```
gsmet: Ah, nice, @gwenneg had issues with this when running inside a Docker container on Windows.

dzwicker: For me it is not clear why you have removed waiting to finish? I want to provide the container name flag using `containerRuntimeOptions` which fails because of the container already exist in the docker daemon.
The daemon is just not cleanup fast enough!

stuartwdouglas: Does adding waitFor() fix this issue? The stream should not return -1 until the process has exited, so they should have the same effect.

dzwicker: I'll try

dzwicker: waitFor does the trick. But I'm not sure if it is just the additional time spent to call the method?

stuartwdouglas: Maybe the stream gets closed as part of the process exit, but before the exit is actually complete, so it is still necessary to call waitFor(). Probably safest just to add it back, can you open a PR?

dzwicker: I'm ready to create the PR. I have tried to find a unit/integration test to adjust (or copy) to reproduce the possible race condition. But I'm unsure which to use. Any help on choosing one?

stuartwdouglas: I don't think you will be able to test for it, as it is a race you can't actually prove that it is fixed with a test , and the test will be slow and require a lot of code too not really prove anything, I think it just needs to be a one liner.

dzwicker: Created the PR https://github.com/quarkusio/quarkus/pull/10562

PR #6036, opened 2019-12-09
Close connection in flyway tests

rsvoboda:
Close connection in flyway tests

### DIFF
--- a/extensions/flyway/deployment/src/test/java/io/quarkus/flyway/test/FlywayExtensionCleanAtStartTest.java
+++ b/extensions/flyway/deployment/src/test/java/io/quarkus/flyway/test/FlywayExtensionCleanAtStartTest.java
@@ -40,9 +40,7 @@ public class FlywayExtensionCleanAtStartTest {
     @DisplayName("Clean at start correctly")
     public void testFlywayConfigInjection() throws SQLException {
 
-        Connection connection = defaultDataSource.getConnection();
-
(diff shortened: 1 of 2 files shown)


extensions/flyway/deployment/src/test/java/io/quarkus/flyway/test/FlywayExtensionCleanAtStartTest.java
```diff
@@ -48,6 +48,8 @@ public void testFlywayConfigInjection() throws SQLException {
             } catch (JdbcSQLException e) {
                 // expected fake_existing_tbl does not exist
             }
+        } finally {
+            connection.close();
```
gsmet: Maybe we should use a try with resources for this, too?

gsmet (changes requested):

extensions/flyway/deployment/src/test/java/io/quarkus/flyway/test/FlywayExtensionCleanAtStartTest.java
rsvoboda: ok, I will change it to this approach: `try (Connection connection = defaultDataSource.getConnection(); Statement stat = connection.createStatement()) {`

gsmet (approved):

gsmet:
Good catch, thanks.

gastaldi (approved):

PR #6053, opened 2019-12-09
Update KC authorization test to return the data only if 'from-body' key exists

### LINKED ISSUE #6037: keycloak-authorization - ProtectedResource - if-check in bodyClaim
https://github.com/quarkusio/quarkus/blob/master/integration-tests/keycloak-authorization/src/main/java/io/quarkus/it/keycloak/ProtectedResource.java#L62

current `if (body == null && !body.containsKey("from-body")) {` is weird

shouldn't it be
 * `if (body != null && !body.containsKey("from-body")) {` 
or 
 * `if (body == null || !body.containsKey("from-body")) { `
?

sberyozkin:
Fixes #6037


pedroigor (changes requested):
Sorry for this one @sberyozkin. It should be a `||` to also check if the body is null so that we make sure that even if the body is parsed before reaching that method it will be available to the application.

sberyozkin:
@pedroigor np :-), updated...

pedroigor (approved):

sberyozkin:
@gsmet Hi Guillaume can you approve it please for it to be merged ? Pedro has already approved

gsmet (approved):

PR #6054, opened 2019-12-09
Update Kafka to 2.3.1 and Debezium to 0.10.0

gsmet:


gastaldi (approved):

gwenneg:
The failure is related. There's a new argument in Kafka 2.3.x for one of the method we're substituting: https://github.com/apache/kafka/blob/2.3/clients/src/main/java/org/apache/kafka/common/utils/AppInfoParser.java#L60. We only need to add the `long nowMs` argument to the current Quarkus substitution and it should work just fine.

gsmet:
Yes, I have to work on it. Will probably do so today.

gwenneg (approved):
LGTM

PR #6055, opened 2019-12-09
Add banner extension module

### LINKED ISSUE #5267: Offer a Quarkus asciiart banner
It should be displayed once per Quarkus dev run. And probably not in tests. 
And be overridable for the usâmes like Keycloak with a small powered by Quarkus subline. 

CC @cescoffier

RotBolt:
this is work in progress PR to fix #5267 

Work Done
- Add Banner extension module (I have followed the work from quarkus-superheroes-workshop)

Problems I am facing
- Where to put the `asciiart.txt`? 

In superheroes-workshop there was `rest-fight` module in which put the `banner.txt` in resources folder and made config in application.properties

How to do that here ?


extensions/banner/pom.xml
```diff
@@ -0,0 +1,50 @@
...
+
+<!--    <properties>-->
+<!--        <quarkus.version>1.0.0.CR1</quarkus.version>-->
+<!--        <surefire-plugin.version>2.22.0</surefire-plugin.version>-->
+<!--        <maven.compiler.source>1.8</maven.compiler.source>-->
+<!--        <maven.compiler.target>1.8</maven.compiler.target>-->
+<!--        <maven.build.timestamp.format>yyyy-MM-dd</maven.build.timestamp.format>-->
+<!--        <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>-->
+<!--        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>-->
+<!--    </properties>-->
```
gastaldi: Delete these lines

KaenDagger: Deleted

extensions/banner/deployment/pom.xml
```diff
@@ -0,0 +1,65 @@
...
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>quarkus-banner-deployment</artifactId>
+    <name>Quarkus - Banner - Deployment</name>
+
+    <dependencies>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-core-deployment</artifactId>
```
geoand: This dependency isn't really needed since you are using `quarkus-arc-deployment` which already includes this one.

extensions/banner/deployment/pom.xml
```diff
@@ -0,0 +1,65 @@
...
+            <version>999-SNAPSHOT</version>
+            <scope>compile</scope>
+        </dependency>
+    </dependencies>
+
+    <build>
+        <plugins>
+            <plugin>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <version>3.8.1</version>
```
geoand: I am pretty sure you don't need to specify the plugin version (and it's actually desirable not do so)

extensions/banner/deployment/pom.xml
```diff
@@ -0,0 +1,65 @@
...
+                    <annotationProcessorPaths>
+                        <path>
+                            <groupId>io.quarkus</groupId>
+                            <artifactId>quarkus-extension-processor</artifactId>
+                            <version>${project.version}</version>
+                        </path>
+                    </annotationProcessorPaths>
+                </configuration>
+            </plugin>
+            <plugin>
```
geoand: I think you can replace this whole plugin section with:

```
            <plugin>
            </plugin>
```

extensions/banner/deployment/src/main/java/io/quarkus/banner/deployment/BannerProcessor.java
```diff
@@ -0,0 +1,49 @@
...
+import io.quarkus.deployment.annotations.BuildStep;
+import io.quarkus.deployment.annotations.ExecutionTime;
+import io.quarkus.deployment.annotations.Record;
+import io.quarkus.deployment.builditem.HotDeploymentWatchedFileBuildItem;
+import io.quarkus.deployment.util.FileUtil;
+
+public class BannerProcessor {
+
+    @BuildStep
+
```
geoand: This should probably be removed :)

gastaldi:
> * Where to put the `asciiart.txt`?

In `src/main/resources` would be more logical I think

extensions/banner/runtime/src/main/java/io/quarkus/banner/runtime/BannerRecorder.java
```diff
@@ -0,0 +1,11 @@
+package io.quarkus.banner.runtime;
+
+import io.quarkus.runtime.annotations.Recorder;
+
+@Recorder
+public class BannerRecorder {
+
+    public void print(String banner) {
+        System.err.println(banner);
```
gastaldi: Why `System.err`? Better use a logger here, no?

extensions/banner/deployment/src/main/java/io/quarkus/banner/deployment/BannerProcessor.java
KaenDagger: Isn't the `@BuildStep`annoation necessary? I thought this needed as this would run at build time

extensions/banner/runtime/src/main/java/io/quarkus/banner/runtime/BannerRecorder.java
KaenDagger: Ok sure

KaenDagger:
> In `src/main/resources` would be more logical I think

Of banner/runtime module ? I was following the work of superheroes workshop. In that `asciiart.txt` was put under the resources of particular module `fight`. So the asciiart worked for that module only
[banner in rest-fight](https://github.com/quarkusio/quarkus-workshops/blob/master/quarkus-workshop-super-heroes/super-heroes/rest-fight/src/main/resources/banner.txt) only ?

extensions/banner/deployment/src/main/java/io/quarkus/banner/deployment/BannerProcessor.java
geoand: I meant you can remove the extra line 😄

KaenDagger: oh sure

extensions/banner/deployment/pom.xml
KaenDagger: You mean remove all plugins and use only sure-fire plugin

geoand: I mean that you can drop all the configuration of the plugin (including its version) since it will be automatically configured.

KaenDagger:
Please review the proposed [asciiArt](https://github.com/KaenDagger/quarkus/blob/AddBanner/extensions/banner/runtime/src/main/resources/banner.txt)

jaikiran:
> Please review the proposed asciiArt

Hello @KaenDagger, do you have a version of that which has a space between each letter? I find this current form a bit hard to read, so maybe the space version would be easier to read?

gastaldi:
> Of banner/runtime module ? 

No, in the ' src/main/resources` of the project that uses this extension. I believe you want people to provide their own `ascii-art.txt` to be displayed in the log? 

Bonus point: It would be nice if Qute could be used to render that file, so the ascii-art could have more info, like the application name taken from application.properties

geoand:
You'll also need to properly format the changed code. You can do that very easily using `mvn process-resources` and commiting (and most probably amending)  the changes that are made automatically

KaenDagger:
> > Of banner/runtime module ?
> 
> No, in the `src/main/resources` of the project that uses this extension. I believe you want people to provide their own `banner.txt` to be displayed in the log?

Actually I thought to give the default one. 

> 
> Bonus point: It would be nice if Qute could be used to render that file, so the ascii-art could have more info, like the application name taken from application.properties

I have to check Qute for that. IG It would require me some more time then

KaenDagger:
> > Please review the proposed asciiArt
> 
> Hello @KaenDagger, do you have a version of that which has a space between each letter? I find this current form a bit hard to read, so maybe the space version would be easier to read?

Ok lemme try

extensions/banner/runtime/src/main/java/io/quarkus/banner/runtime/BannerRecorder.java
```diff
@@ -0,0 +1,13 @@
...
+
+import io.quarkus.runtime.annotations.Recorder;
+import org.jboss.logging.Logger;
+
+
+@Recorder
+public class BannerRecorder {
+    public void print(String banner) {
+        Logger LOGGER = Logger.getLogger(BannerRecorder.class.getPackage().getName());
+        LOGGER.debug(banner);
```
gastaldi: `info` level is more adequate here, also make the `LOGGER` variable all lowercase, to conform to the Java coding standard

KaenDagger: I checked Loggers in other files those were capital. Sure I will change it

gastaldi: Probably because they were declared as static final attributes, and even in this case they should be lowercased: https://stackoverflow.com/questions/1417190/should-a-static-final-logger-be-declared-in-upper-case

gastaldi:
This is a nice website to generate ASCII Art: http://patorjk.com/software/taag/#p=display&h=1&f=Big&t=Quarkus

gastaldi:
This looks cool BTW :) : 

```

 ██████╗ ██╗   ██╗ █████╗ ██████╗ ██╗  ██╗██╗   ██╗███████╗
██╔═══██╗██║   ██║██╔══██╗██╔══██╗██║ ██╔╝██║   ██║██╔════╝
██║   ██║██║   ██║███████║██████╔╝█████╔╝ ██║   ██║███████╗
██║▄▄ ██║██║   ██║██╔══██║██╔══██╗██╔═██╗ ██║   ██║╚════██║
╚██████╔╝╚██████╔╝██║  ██║██║  ██║██║  ██╗╚██████╔╝███████║
 ╚══▀▀═╝  ╚═════╝ ╚═╝  ╚═╝╚═╝  ╚═╝╚═╝  ╚═╝ ╚═════╝ ╚══════╝
                                                           
```

geoand:
Asciiart definitely calls for some serious bikesheding :P

KaenDagger:
After updating pom.xml I am having this fail test 
[io.quarkus.it.main.ConfigPropertiesTestCase.txt](https://github.com/quarkusio/quarkus/files/3963921/io.quarkus.it.main.ConfigPropertiesTestCase.txt)

extensions/banner/deployment/src/main/java/io/quarkus/banner/deployment/BannerProcessor.java
```diff
@@ -0,0 +1,53 @@
...
+    @BuildStep
+    @Record(ExecutionTime.RUNTIME_INIT)
+    public void recordBanner(BannerRecorder recorder, BannerConfig config) {
+        String content = readBannerFile(config.path);
+        recorder.print(content);
+    }
+
+    @BuildStep
+    List<HotDeploymentWatchedFileBuildItem> watchBannerChanges(BannerConfig config) {
```
machi1990: Since we are returning a list containing only one item the piece of code could be refactored to 

```
@BuildStep
HotDeploymentWatchedFileBuildItem watchBannerChanges(BannerConfig config) {
        return new HotDeploymentWatchedFileBuildItem(config.path);
    }
```

extensions/banner/deployment/src/main/java/io/quarkus/banner/deployment/BannerProcessor.java
```diff
@@ -0,0 +1,53 @@
...
+import io.quarkus.banner.runtime.BannerRecorder;
+import io.quarkus.deployment.annotations.BuildStep;
+import io.quarkus.deployment.annotations.ExecutionTime;
+import io.quarkus.deployment.annotations.Record;
+import io.quarkus.deployment.builditem.HotDeploymentWatchedFileBuildItem;
+import io.quarkus.deployment.util.FileUtil;
+
+public class BannerProcessor {
+
+    private static final String DEFAULT_BANNER_PATH = "src/main/resources/banner.txt";
```
machi1990: Probably make this the `defaultValue` for `BannerConfig #path` field so that it is properly  and automatically documented with our config doc generation infrastructure.

KaenDagger: Sure ! That makes more sense

extensions/banner/deployment/pom.xml
```diff
@@ -0,0 +1,51 @@
...
+        </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-junit5-internal</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-banner-runtime</artifactId>
+            <version>999-SNAPSHOT</version>
```
gastaldi: Don't specify the version here, `quarkus-banner` should be declared in the `bom/runtime/pom.xml`

KaenDagger:
both options working default one 

quarkus-def

user provided one in application.properties
quarkus-user

extensions/banner/deployment/pom.xml
KaenDagger: done

extensions/banner/deployment/pom.xml
```diff
@@ -0,0 +1,50 @@
...
+        </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-junit5-internal</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-banner</artifactId>
+            <scope>compile</scope>
```
geoand: We don't need the `scope` here do we?

extensions/banner/deployment/src/main/java/io/quarkus/banner/deployment/BannerConfig.java
```diff
@@ -0,0 +1,22 @@
...
+import io.quarkus.runtime.annotations.ConfigPhase;
+import io.quarkus.runtime.annotations.ConfigRoot;
+
+@ConfigRoot(name = "banner", phase = ConfigPhase.BUILD_TIME)
+public class BannerConfig {
+
+    /**
+     * The path of the default quarkus banner
+     */
+    public static final String DEFAULT_BANNER_PATH = "default_banner.txt";
```
geoand: Can we utilize the `defaultValue` from `@ConfigItem` for this?

extensions/banner/runtime/pom.xml
```diff
@@ -0,0 +1,48 @@
...
+        <relativePath>../</relativePath>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>quarkus-banner</artifactId>
+    <name>Quarkus - Banner - Runtime</name>
+
+    <dependencies>
+        <dependency>
+            <groupId>io.quarkus</groupId>
```
geoand: This dependency shouldn't be needed since you are including Arc

geoand:
I added some minor comments

extensions/banner/deployment/src/main/java/io/quarkus/banner/deployment/BannerConfig.java
KaenDagger: No. Otherwise it will ask for default_path parameter from user. Besides Its default for Quarkus, so IG it makes more sense to be static final

geoand: I still don't understand why it needs to be a separate field instead of just using: https://github.com/quarkusio/quarkus/blob/master/core/runtime/src/main/java/io/quarkus/runtime/annotations/ConfigItem.java#L55.
Something like

```
@ConfigItem(defaultValue = "default_banner.txt")
public String path;
```

Why would that not work?

extensions/banner/deployment/pom.xml
KaenDagger: No we don't. I checked it

geoand: :+1:

extensions/banner/deployment/src/main/java/io/quarkus/banner/deployment/BannerConfig.java
KaenDagger: Default value is present in Quarkus-banner resources for that only classLoader works
but for the user provided banner. It exists in their project resources for that, we need `contexrClassLoader`. If I give default value here, there won't be no way to differentiate between the two

geoand: OK, thanks

extensions/banner/runtime/src/main/java/io/quarkus/banner/runtime/BannerRecorder.java
```diff
@@ -0,0 +1,13 @@
+package io.quarkus.banner.runtime;
+
+import org.jboss.logging.Logger;
+
+import io.quarkus.runtime.annotations.Recorder;
+
+@Recorder
+public class BannerRecorder {
+    public void print(String banner) {
+        Logger LOGGER = Logger.getLogger(BannerRecorder.class.getPackage().getName());
```
gastaldi:
```suggestion
        Logger LOGGER = Logger.getLogger("io.quarkus");
```

extensions/banner/deployment/src/main/java/io/quarkus/banner/deployment/BannerConfig.java
KaenDagger: Sorry my bad. Your suggestion is working. I used it default value directly as string and didn't use annotation's default value. 
Thanks for telling me. improving the code :sweat_smile:

geoand: Great to hear :wink:

extensions/banner/deployment/src/main/java/io/quarkus/banner/deployment/BannerConfig.java
```diff
@@ -0,0 +1,14 @@
+package io.quarkus.banner.deployment;
+
+import io.quarkus.runtime.annotations.ConfigItem;
+import io.quarkus.runtime.annotations.ConfigPhase;
+import io.quarkus.runtime.annotations.ConfigRoot;
+
+@ConfigRoot(name = "banner", phase = ConfigPhase.BUILD_TIME)
+public class BannerConfig {
+    /**
+     * The path of the banner which could be provided by user
```
geoand: I think it would be nice to have an example here showing how a user can put a custom banner (and where they need to add this file).

Other than this, I'm good with the PR

geoand: I am thinking of telling the user that adding `my_banner.txt` in `src/main/resources` then just using `quarkus.banner.path=my_banner.txt`

KaenDagger: Wouldn't that be good for documentation purpose ? A proper document would be more clear IMO

geoand: The good thing is that these config properties generate documentation automatically (see other config objects for example).
So it would be a nice first step to have a nice javadoc for the property since Quarkus users will know where to see the documentation.

KaenDagger: how to do `javaDoc` then ?

gastaldi:
LGTM but it would be really nice if you had a guide ( in the `docs` module) explaining how to add a custom banner, like @geoand mentioned above

extensions/banner/deployment/src/main/java/io/quarkus/banner/deployment/BannerConfig.java
geoand: See for `MongoClientConfig` for example.

geoand:
I think that in the beginning we can just go with javadoc on the config item and look at the creating a guide if needed later.

KaenDagger:
In docs module those are `.adoc` files. I could add one .adoc file in that

extensions/banner/deployment/src/main/java/io/quarkus/banner/deployment/BannerConfig.java
KaenDagger: Could you please tell in which module ? :sweat_smile:

geoand:
The problem is which adoc file to add the stuff to. Perhaps the `getting-started` document?

extensions/banner/deployment/src/main/java/io/quarkus/banner/deployment/BannerConfig.java
geoand: `io.quarkus.mongodb.runtime.MongoClientConfig` in `extensions/mongodb-client/runtime`. Just look at `connectionString` string where a detailed javadoc has been added

gastaldi:
Why not a `banner.adoc` guide?

Em Ter, 17 de dez de 2019 07:52, Georgios Andrianakis <
notifications@github.com> escreveu:

geoand:
Sounds too much to have a whole guide for such a feature, no? 
If you both think we should have a separate guide for it, then I'm OK with it, I just think it's a little over the top :)

KaenDagger:
I know its small feature. Isn't there any Collective extensions .adoc. We can add banner to that

gastaldi:
Yeah, you're right. I was also thinking that maybe this feature could be
part of the generated main Application class instead of being a separate
extension since it's trivial

Em Ter, 17 de dez de 2019 08:16, Georgios Andrianakis <
notifications@github.com> escreveu:

geoand:
> Yeah, you're right. I was also thinking that maybe this feature could be part of the generated main Application class instead of being a separate extension since it's trivial

I was thinking the same thing to be honest

geoand:
> I know its small feature. Isn't there any Collective extensions .adoc ? We can add banner to that

We don't AFAIK. You could just add it to the getting-started guide I guess

geoand:
How about we start it out as an extension and move it into the core later on if necessary?

extensions/banner/runtime/src/main/java/io/quarkus/banner/runtime/BannerRecorder.java
```diff
@@ -0,0 +1,13 @@
...
+
+import org.jboss.logging.Logger;
+
+import io.quarkus.runtime.annotations.Recorder;
+
+@Recorder
+public class BannerRecorder {
+    public void print(String banner) {
+        Logger logger = Logger.getLogger("io.quarkus");
+        logger.info(banner);
```
dmlloyd: The banner shouldn't be logged, it should be [implemented as a header](https://docs.oracle.com/en/java/javase/11/docs/api/java.logging/java/util/logging/Formatter.html#getHead(java.util.logging.Handler)) within [the console setup code](https://github.com/quarkusio/quarkus/blob/master/core/runtime/src/main/java/io/quarkus/runtime/logging/LoggingSetupRecorder.java#L152-L186).  This will likely require an additional build item (e.g. a `ConsoleFormatterBannerBuildItem` which contains a `RuntimeValue<Optional<Supplier<String>>>` or something of that sort) and a bit of core integration to wrap the default formatter with a delegating `ExtFormatter` which supplies the header and delegates the other methods to the default formatter.

The reason for this is:

* The banner should only appear on the console.  We don't want it in a file, published to syslog, ending up in fluentd, etc.
* The formatting of multi-line log messages is a bit awkward by default, with the first line indentation not matching up with the rest.
* If something like JSON formatting is enabled for the console, we don't want a banner because it won't end up making sense to whatever system is parsing it.

extensions/banner/deployment/src/main/java/io/quarkus/banner/deployment/BannerConfig.java
```diff
@@ -0,0 +1,14 @@
...
+     */
+    @ConfigItem(defaultValue = "default_banner.txt")
+    public String path;
```
dmlloyd: I think there should be different paths depending on the console color support (no color, 8 color, 16 color, 256 color, truecolor) that are searched in order from most to least color support (since color modes are backwards compatible).

One thing I had looked into which might be interesting to explore further was rendering PNG images to full color text banners.  I have a working(ish) prototype around somewhere for 256 and 16M colors.  But that's a separate discussion I guess!

extensions/banner/deployment/src/main/java/io/quarkus/banner/deployment/BannerProcessor.java
```diff
@@ -0,0 +1,43 @@
...
+    @BuildStep
+    @Record(ExecutionTime.RUNTIME_INIT)
+    public void recordBanner(BannerRecorder recorder, BannerConfig config) {
+        String content = readBannerFile(config);
+        recorder.print(content);
+    }
+
+    @BuildStep
+    HotDeploymentWatchedFileBuildItem watchBannerChanges(BannerConfig config) {
+        return new HotDeploymentWatchedFileBuildItem(config.path);
```
dmlloyd: :grin:

dmlloyd (dismissed):
This is a great start but I think there needs to be a couple of changes to this code before we bring this in to the main code base.

KaenDagger:
So Do I have to change according to requested here ?

geoand:
@dmlloyd is the logging (among many other things) expert here, so best follow his guidelines :)

extensions/banner/deployment/src/main/java/io/quarkus/banner/deployment/BannerConfig.java
KaenDagger: That would be completely new. IG. Could you give me little leads for this ? 😅

extensions/banner/runtime/src/main/java/io/quarkus/banner/runtime/BannerRecorder.java
KaenDagger: So should I remove this recorder step and use only BannerProcessor and BannerConfig only ?

extensions/banner/deployment/src/main/java/io/quarkus/banner/deployment/BannerConfig.java
dmlloyd: For the color stuff... it looks like we don't (yet) expose that information, so it might be too hard to utilize for now.  But, we can make a note for later. :slightly_smiling_face:

extensions/banner/runtime/src/main/java/io/quarkus/banner/runtime/BannerRecorder.java
dmlloyd: I think the recorder should be modified to return the string supplier runtime value which in turn gets the banner as a string.  The processor step would pass that in to the logger setup using a build item, and the core logging setup processor would have to be modified to merge that in to the formatter configuration.

core/runtime/src/main/java/io/quarkus/runtime/logging/LoggingSetupRecorder.java
```diff
@@ -151,7 +153,8 @@ private static boolean hasColorSupport() {
 
     private static Handler configureConsoleHandler(final ConsoleConfig config, final ErrorManager defaultErrorManager,
             final List<LogCleanupFilterElement> filterElements,
-            final List<RuntimeValue<Optional<Formatter>>> possibleFormatters) {
+            final List<RuntimeValue<Optional<Formatter>>> possibleFormatters,
+            final Supplier<String> bannerSupplier) {
```
KaenDagger: I have added Banner Supplier which itself contains Banner as String. As per @dmlloyd  I should integrate banner here. How to integrate the supplier string to formatter here ?

KaenDagger: Do I have to create a custom formatter that returns the banner when `formatter.getHead(Handler)` is called ? or make changes to PatternFormatter?

KaenDagger:
As per @dmlloyd  
I have added  
- `BannerBuildItem`
- CustomBannerFormatter extending ColorPatternFormatter and PatterFormatter to implement Banner as Header (formatter.getHead)

Results
Default 

quarkus-def

User provided

quarkus-user

geoand:
It seems like CI is failing

KaenDagger:
Probably because BannerBuildItem was not optional. Because of that Basic hello-quarkus application was not working without banner extension.
 I have made it optional now so now it works without that

jaikiran:
@KaenDagger I haven't been following this PR much, but just saw that you updated the (default) banner to include a space between each letter. This looks a lot a better IMO. Thank you for doing that.

core/deployment/src/main/java/io/quarkus/deployment/logging/LoggingResourceProcessor.java
```diff
@@ -12,11 +13,7 @@
 import io.quarkus.deployment.annotations.BuildStep;
 import io.quarkus.deployment.annotations.ExecutionTime;
 import io.quarkus.deployment.annotations.Record;
-import io.quarkus.deployment.builditem.LogCategoryBuildItem;
-import io.quarkus.deployment.builditem.LogConsoleFormatBuildItem;
-import io.quarkus.deployment.builditem.LogHandlerBuildItem;
-import io.quarkus.deployment.builditem.RunTimeConfigurationDefaultBuildItem;
-import io.quarkus.deployment.builditem.SystemPropertyBuildItem;
+import io.quarkus.deployment.builditem.*;
```
gastaldi: We don't use star imports. Make sure to configure your IDE to use the https://github.com/quarkusio/quarkus/blob/master/ide-config/eclipse-format.xml format file

KaenDagger: Done

geoand:
@dmlloyd have your concerns been addressed by the latest changes?

KaenDagger:
If conditions are fulfilled. Could this be merged :sweat_smile:

core/runtime/src/main/java/io/quarkus/runtime/logging/PatternFormatterWithBanner.java
```diff
@@ -0,0 +1,30 @@
+package io.quarkus.runtime.logging;
+
+import java.util.logging.Handler;
+
+import org.jboss.logmanager.formatters.PatternFormatter;
+
+public class PatternFormatterWithBanner extends PatternFormatter {
+    private String banner;
```
dmlloyd: This should be a `private final Supplier<String> bannerSupplier`, probably.

core/runtime/src/main/java/io/quarkus/runtime/logging/ColorPatternFormatterWithBanner.java
```diff
@@ -0,0 +1,34 @@
+package io.quarkus.runtime.logging;
+
+import java.util.logging.Handler;
+
+import org.jboss.logmanager.formatters.ColorPatternFormatter;
+
+public class ColorPatternFormatterWithBanner extends ColorPatternFormatter {
```
dmlloyd: Rather than having two formatter subclasses, what about having a single formatter called `BannerFormatter` which delegates to another `ExtFormatter`?

core/runtime/src/main/java/io/quarkus/runtime/logging/ColorPatternFormatterWithBanner.java
```diff
@@ -0,0 +1,34 @@
...
+    private String banner;
```
dmlloyd: See my other comment; the supplier should be used here, and it should be final.

extensions/banner/runtime/src/main/java/io/quarkus/banner/runtime/BannerRecorder.java
```diff
@@ -0,0 +1,36 @@
...
+
+import io.quarkus.deployment.util.FileUtil;
+import io.quarkus.runtime.RuntimeValue;
+import io.quarkus.runtime.annotations.Recorder;
+
+@Recorder
+public class BannerRecorder {
+    public RuntimeValue<Optional<Supplier<String>>> provideBannerSupplier(BannerConfig config) {
+        String banner = readBannerFile(config);
+        Supplier<String> supplierValue = () -> banner;
```
dmlloyd: This supplier should do the actual banner file reading, so that the banner is only held in memory when it is actually being displayed.

extensions/banner/runtime/src/main/resources/default_banner.txt
```diff
@@ -0,0 +1,9 @@
+
```
dmlloyd: I do wish we had a better default banner.

dmlloyd (dismissed):

extensions/banner/runtime/src/main/java/io/quarkus/banner/runtime/BannerRecorder.java
KaenDagger: So I would pass the path of banner.txt to supplier. And supplier will read it from the path when it provides string. ?

extensions/banner/runtime/src/main/resources/default_banner.txt
KaenDagger: I will try to add different one.

extensions/banner/runtime/src/main/java/io/quarkus/banner/runtime/BannerRecorder.java
KaenDagger: Done

core/runtime/src/main/java/io/quarkus/runtime/logging/ColorPatternFormatterWithBanner.java
KaenDagger: I have extended ColorPatternFormatter, sinxe ConsolehandlerTest asserts for PatternFormatter instance only. In the single BannerFormatter , I have overriden  methods which were different in Pattern and ColorPattern and delegated  them to corresponding formatter instances

core/runtime/src/main/java/io/quarkus/runtime/logging/PatternFormatterWithBanner.java
KaenDagger: done

extensions/banner/runtime/src/main/resources/default_banner.txt
KaenDagger: I have added this one as suggested by @gastaldi with slight modifications

core/runtime/src/main/java/io/quarkus/runtime/logging/BannerFormatter.java
```diff
@@ -0,0 +1,62 @@
...
+        if (isColorPattern) {
+            return ((ColorPatternFormatter) formatter).format(extLogRecord);
+        } else {
+            return ((PatternFormatter) formatter).format(extLogRecord);
+        }
+    }
+
+    @Override
+    public String getHead(Handler h) {
+        return bannerSupplier.get();
```
gastaldi: Better check if bannerSupplier is null and call `super.getHead(h)` if it does

KaenDagger: But creating the instance of bannerFormatter supplier is necessary. So it wouldn't be null in case of bannerFormatter

gastaldi: In this case, better make sure bannerSupplier is never null in the constructor

KaenDagger: I checked the quarkus application without banner extension. It was throwing NPE. It was careless  of me to miss that thing. Thanks for pointing that out. In case banner-extension is not used, application will use normal formatters now

KaenDagger:
@dmlloyd could you please review again ?

geoand:
@dmlloyd I see you dismissed your review, does that mean you are OK with the changes?

dmlloyd:
It just means I took myself out of the voting.

geoand:
OK, thanks!

geoand:
@KaenDagger I see that the banner is read from the classpath, however I didn't see the resource being registered for the native image generation. 

Have you tried this PR with an application that is compiled to a native image?

extensions/banner/runtime/src/main/java/io/quarkus/banner/runtime/BannerRecorder.java
```diff
@@ -0,0 +1,38 @@
...
+            try (InputStream is = resource.openStream()) {
+                byte[] content = FileUtil.readFileContents(is);
+                String bannerTitle = new String(content, StandardCharsets.UTF_8);
+                return bannerTitle +
+                        ":: Quarkus :: v" + Version.getVersion() + '\n';
+            } catch (IOException e) {
+                throw new UncheckedIOException(e);
+            }
+        } else {
+            throw new IllegalArgumentException("Cannot find the banner file");
```
geoand: Do we really want to be this strict?

KaenDagger:
No I didn't , I will try it

geoand:
Thank you

geoand:
@KaenDagger were you able to test the extension in native?

KaenDagger:
Yeah ! And there is WRITE_FAILURE for banner

extensions/banner/runtime/src/main/java/io/quarkus/banner/runtime/BannerRecorder.java
KaenDagger: I will change it to warning

extensions/banner/runtime/pom.xml
machi1990: +1

extensions/banner/runtime/pom.xml
```diff
@@ -0,0 +1,48 @@
...
+            <artifactId>quarkus-arc</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-core-deployment</artifactId>
```
machi1990: runtime module should not depend on deployment module.

extensions/banner/runtime/src/main/java/io/quarkus/banner/runtime/BannerRecorder.java
```diff
@@ -0,0 +1,43 @@
...
+@Recorder
+public class BannerRecorder {
+    private Logger logger = Logger.getLogger(BannerRecorder.class);
+
+    public RuntimeValue<Optional<Supplier<String>>> provideBannerSupplier(BannerConfig config) {
+        Supplier<String> supplierValue = () -> readBannerFile(config);
+        return new RuntimeValue<>(Optional.of(supplierValue));
+    }
+
+    private String readBannerFile(BannerConfig config) {
```
machi1990: Can this be done during build time? And pass the banner text in the recorder? 
i.e modify the signature of `provideBannerSupplier` to  
```
 public RuntimeValue<Optional<Supplier<String>>> provideBannerSupplier(String bannerText) {
        Supplier<String> supplierValue = () -> bannerText;
        return new RuntimeValue<>(Optional.of(supplierValue));
    }
```

extensions/banner/runtime/src/main/java/io/quarkus/banner/runtime/BannerConfig.java
```diff
@@ -0,0 +1,14 @@
+package io.quarkus.banner.runtime;
+
+import io.quarkus.runtime.annotations.ConfigItem;
+import io.quarkus.runtime.annotations.ConfigPhase;
+import io.quarkus.runtime.annotations.ConfigRoot;
+
+@ConfigRoot(name = "banner", phase = ConfigPhase.BUILD_TIME)
+public class BannerConfig {
```
machi1990: This can be moved to the deployment module

extensions/banner/runtime/src/main/java/io/quarkus/banner/runtime/BannerConfig.java
```diff
@@ -0,0 +1,14 @@
+package io.quarkus.banner.runtime;
+
+import io.quarkus.runtime.annotations.ConfigItem;
+import io.quarkus.runtime.annotations.ConfigPhase;
+import io.quarkus.runtime.annotations.ConfigRoot;
+
+@ConfigRoot(name = "banner", phase = ConfigPhase.BUILD_TIME)
```
machi1990:
```suggestion
@ConfigRoot(name = "banner")
```

nitpick: the default phase is `BUILD_TIME`

extensions/banner/deployment/pom.xml
```diff
@@ -0,0 +1,49 @@
...
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>quarkus-banner-deployment</artifactId>
+    <name>Quarkus - Banner - Deployment</name>
+
+    <dependencies>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-arc-deployment</artifactId>
+        </dependency>
```
machi1990: we should have a dependency on `quarkus-core-deployment` artifact and not on `arc`.

extensions/banner/deployment/pom.xml
```diff
@@ -0,0 +1,49 @@
...
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-junit5-internal</artifactId>
+            <scope>test</scope>
```
machi1990: I am seeing that we have this dependency but not associated tests, it would be good to have some if possible or remove this dependency.

extensions/banner/deployment/src/main/java/io/quarkus/banner/deployment/BannerProcessor.java
```diff
@@ -0,0 +1,29 @@
...
+    }
+
+    @BuildStep
+    NativeImageResourceBuildItem nativeImageResourceBuildItem(BannerConfig config) {
+        return new NativeImageResourceBuildItem(config.path);
+    }
```
machi1990: if we move the code that read the banner file to build time, we shouldn't need this build step.

machi1990 (changes requested):
I have added a couple of suggestions. Let me know what do you think.

extensions/banner/runtime/pom.xml
geoand: Very very good catch

extensions/banner/runtime/src/main/java/io/quarkus/banner/runtime/BannerConfig.java
geoand: I personally like having the phase explicitly set, but yeah it's just a matter of preference :)

machi1990: Yes, it is a minor one. We can leave it the way it is :-)

extensions/banner/runtime/pom.xml
KaenDagger: Why so ? 😅

geoand: The runtime module should never depend on anything from the deployment module. The deployment module is meant for build time only code and dependencies

KaenDagger:
Alot of requested changes are from the time. When I didn't know about deployment and runtime. I will take care of these things. Thank you for catching these things

geoand:
Unfortunately things like this are easy to miss in a review, which is why we are extremely grateful to have plenty of eyes on PRs :)

extensions/banner/runtime/src/main/java/io/quarkus/banner/runtime/BannerRecorder.java
KaenDagger: Actually david said this 
```
This supplier should do the actual banner file reading, so that the banner is only held in memory when it is actually being displayed.
```

extensions/banner/deployment/src/main/java/io/quarkus/banner/deployment/BannerProcessor.java
geoand: Very good point

extensions/banner/runtime/src/main/java/io/quarkus/banner/runtime/BannerRecorder.java
KaenDagger: But for native image part, It makes sense, we do not have to use nativeImageBuildItem then

KaenDagger: umm So how to proceed now ?

geoand: I think we can probably just leave it as it is for now. This isn't terribly important IMHO

KaenDagger: Lemme make the final changes as for now. Also new suggested changes makes it easier to read the code more. :)

extensions/banner/deployment/pom.xml
KaenDagger: removed

extensions/banner/runtime/src/main/java/io/quarkus/banner/runtime/BannerConfig.java
KaenDagger: moved

extensions/banner/runtime/src/main/java/io/quarkus/banner/runtime/BannerRecorder.java
machi1990: Oh yes, let's keep it this way for now.

machi1990 (approved):

geoand:
@KaenDagger what was the reason you closed this one? To lift the curse 😂?

KaenDagger:
I will make new PR , 😶😶. Probably won't fail with that.

KaenDagger:
PR shifted to #7093

PR #6058, opened 2019-12-09
StandardCharsets.UTF_8 usage

rsvoboda:
StandardCharsets.UTF_8 usage to avoid platform defaults


gastaldi (approved):

gsmet:
The re-run button is gone for whatever reason, I force-pushed an update.

PR #6064, opened 2019-12-09
Drop unused build item classes

dmlloyd:
These classes are not exposed to build step APIs and should be dropped


gastaldi (approved):

gastaldi:
I think this is OK to merge, WDYT @gsmet ?

dmlloyd:
It should be perfectly safe. These types and methods have never been used.

PR #6071, opened 2019-12-10
Add Qute experimental status and fix metadata

ia3andy:


gsmet (approved):

extensions/resteasy-qute/runtime/src/main/resources/META-INF/quarkus-extension.yaml
```diff
@@ -0,0 +1,11 @@
+---
+name: "RESTEasy Qute"
+metadata:
+  keywords:
+  - "templating"
+  - "templates"
+  - "resteasy"
+  guide: "https://quarkus.io/guides/qute"
```
dufoli: url do not worked. Is it expected?

gsmet: Yes. The guide will be pushed when we release 1.1.0.Final.

PR #6072, opened 2019-12-10
Remove Kotlin from dependabot config

gastaldi:
Closes #6051

### DIFF
--- a/.dependabot/config.yml
+++ b/.dependabot/config.yml
@@ -54,15 +54,6 @@ update_configs:
       # Debezium
       - match:
           dependency_name: "io.debezium:debezium-core"
-      # Kotlin
-      - match:
-          dependency_name: "org.jetbrains.kotlin:kotlin-compiler"
-      - match:
-          dependency_name: "org.jetbrains.kotlin:kotlin-stdlib-jdk8"
-      - match:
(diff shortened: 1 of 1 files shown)


geoand (approved):

gsmet:
I'm not entirely sure I agree tbh. It's a nice reminder and we could have changed the other version manually.

Not sure it's better to ship with an outdated Kotlin, given how Kotlin people are bleeding edge.

gastaldi:
In this case, we should dedicate some time to see if we can reduce everything to just use one version (as suggested by @geoand). I can provide a PR reverting this commit if you like

geoand:
Well I think it's better to ship an outdated but consistent experience.
Of course ideally we would have the Kotlin version in a single place and then be able to use the bot. I think we should look into doing this.

gsmet:
Well, it would be consistent if we add a commit when the bot creates a PR. I'm not saying it shouldn't be.

I can see how people might be tempted to just merge things when CI is green though.

And yes, I agree we should try to fix this issue somehow.

geoand:
Exactly, it would be too easy to just merge when everything is green and forgetting about the other properties.
I can absolutely see that happening to me 😁.

I'll try and take a look over the weekend and see if we can unify the Kotlin version properties in an easy way.

PR #6077, opened 2019-12-10
Fix an artemis-commons dependency convergence issue

gsmet:
We don't have it in Quarkus but we have it in Camel Quarkus. Artemis
dependencies are not consistent.


gastaldi:
Better add in https://github.com/quarkusio/quarkus/blob/master/.dependabot/config.yml#L10-L15 too, just in case

gsmet:
@gastaldi good catch, will do. Thanks.

gsmet:
@gastaldi done!

gastaldi (approved):

PR #6082, opened 2019-12-10
revert change to run()

evanchooly:
Simply changing the `start()` call to `run()` causes the system to block on AWS and it doesn't serve any requests.  This change reverts back to calling `start()` and adds a shutdown hook to call `stop()` so that any external resources are properly cleaned up.


dmlloyd:
I'm not really sure how this is supposed to work to start with.  The change seems reasonable enough on the surface, though once again it seems we're diverging to N different entry/exit paths rather than having one that works in all cases.

In conclusion, +0

evanchooly:
@dmlloyd ok.  i added you because i figured you'd have a general opinion at the quarkus/jvm level.  no worries, though.  thanks.

dmlloyd:
I'm not against it per se, I just don't know enough about this launch environment to have a useful opinion.

stuartwdouglas (approved):

Issue #3449, opened 2019-08-07
Flyway should support named datasources


JohT:
**Description**
"application.properties" support defining multiple named datasources:
`
#default datasource:
quarkus.datasource.url 
#datasource for database1:
quarkus.datasource.database1.url
#datasource for database2:
quarkus.datasource.database2.url 
`

Currently, flyway only uses the default datasource.
It would be nice, if it can be configured for every named datasource in a similar manner:
`
#flyway for default datasource:
quarkus.flyway.migrate-at-start
#flyway for database1 datasource:
quarkus.flyway.database1.migrate-at-start
#flyway for database2 datasource:
quarkus.flyway.database2.migrate-at-start
` 

**Implementation ideas**
Discussion see: https://groups.google.com/forum/#!topic/quarkus-dev/NrGrBJBdiG4

I'll provide an pull/merge request for this.

JohT:
Example (format corrected):
`
#default datasource: 
quarkus.datasource.url 
#datasource for database1: 
quarkus.datasource.database1.url 
#datasource for database2: 
quarkus.datasource.database2.url

#flyway for default datasource: 
quarkus.flyway.migrate-at-start 
#flyway for database1 datasource: 
quarkus.flyway.database1.migrate-at-start 
#flyway for database2 datasource: 
quarkus.flyway.database2.migrate-at-start
`

JohT:
Since this is solved, i'll close this issue.

PR #6086, opened 2019-12-10
Documentation for Flyway support for multiple datasources #3449

JohT:
Pull Request for the documentation on how to use Flyway with multiple datasources as described in
#3449.

### DIFF
--- a/docs/src/main/asciidoc/flyway.adoc
+++ b/docs/src/main/asciidoc/flyway.adoc
@@ -43,8 +43,10 @@ In your `pom.xml`, add the following dependencies:
 </dependencies>
 --
 
-Flyway support relies on the Quarkus default datasource config, you must add the default datasource properties
-to the `{config-file}` file in order to allow Flyway to manage the schema.
+Flyway support relies on the Quarkus datasource config. 
+It can be customized for the default datasource as well as for every <<multiple-datasources,named datasource>>.
+First, you need to add the datasource config to the `{config-file}` file 
+in order to allow Flyway to manage the schema.
 Also, you can customize the Flyway behaviour by using the following properties:
 
 `quarkus.flyway.migrate-at-start`::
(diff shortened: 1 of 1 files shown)


docs/src/main/asciidoc/flyway.adoc
```diff
@@ -170,6 +172,52 @@ public class MigrationService {
 
 <1> Inject the Flyway object if you want to use it directly
 
+== Multiple datasources
+
+Flyway can be configured for multiple datasources. 
+It works exactly the same way as for named datasource, for example:
```
gsmet:
```suggestion
It works exactly the same way as for the default datasource, for example:
```
?

docs/src/main/asciidoc/flyway.adoc
```diff
@@ -170,6 +172,52 @@ public class MigrationService {
...
+# Flyway configuration for the "inventory" datasource
+quarkus.flyway.inventory.schemas=INVENTORY_TEST_SCHEMA
+quarkus.flyway.inventory.locations=db/inventory/location1,db/inventory/location2
+quarkus.flyway.inventory.migrate-at-start=true
+--
+
+Notice there's an extra bit in the key.
+The syntax is as follows: `quarkus.flyway.[optional name.][datasource property]`.
+
+NOTE: Without configuration, Flyway is setup for every datasource using the default settings.
```
gsmet:
```suggestion
NOTE: Without configuration, Flyway is set up for every datasource using the default settings.
```
So... I was wondering if it was a wise choice or if we should enable it only if a location is set? I mean there's very little chance that you will want all your datasources to share the same location? Or am I wrong?

docs/src/main/asciidoc/flyway.adoc
```diff
@@ -185,9 +233,17 @@ public class MigrationService {
     @Inject
     Flyway flyway; <1>
 
+    @Inject
+    @Named("flyway_users") <2>
+    Flyway flywayForUsers;
+
+    @Inject
+    @FlywayDataSource("inventory") <3>
```
gsmet: Let's make it the first proposal as I would prefer to recommend this usage.

docs/src/main/asciidoc/flyway.adoc
```diff
@@ -196,5 +252,7 @@ public class MigrationService {
 --
 
 <1> Inject the Flyway object if you want to use it directly
-<2> Use the Flyway instance directly
+<2> Inject Flyway for named datasources
+<3> Inject Flyway for named datasources using the quarkus qualifier
```
gsmet:
```suggestion
<3> Inject Flyway for named datasources using the Quarkus `FlywayDataSource` qualifier
```

gsmet (changes requested):
@JohT thanks!

I added a few suggestions and (I think) one important question that we should figure out before Final IMHO.

docs/src/main/asciidoc/flyway.adoc
JohT: I hear you. I think, it makes sense as it is. 

If there is only one datasource (named or default, doesn't matter) it is perfectly fine, that the default location is used. The most important point IMO is, that "quarkus.flyway.migrate-at-start" is false by default. Nothing "bad" happens, if there are multiple datasources with default Flyway configuration. 

Everything else is in the responsibility of the developer. Even if Flyway would only be activated with a configured location, it would still be possible to copy&paste the same location. It might not be the standard use case, but it could also be used to have two databases with the same scheme (e.g. to synchronize a backup system).

Let me know, if i should start a discussion on google.groups to get more opinions.

gsmet: Let's try it your way and we'll see what feedback we have.

I let you fix the other things? Thanks!

docs/src/main/asciidoc/flyway.adoc
JohT: ok

docs/src/main/asciidoc/flyway.adoc
JohT: ok

docs/src/main/asciidoc/flyway.adoc
JohT: i reworded the sentence.

docs/src/main/asciidoc/flyway.adoc
JohT: i did and squashed it into a single commit.

gsmet (approved):

gsmet:
Thanks @JohT !

PR #6093, opened 2019-12-11
The quarkus.index-dependency.<name>.classifier property should be optional

dwalluck:
…ional


gsmet (approved):
Good catch.

gsmet:
Merged, thanks. I will backport it to the Final.

PR #6099, opened 2019-12-11
Controller advice exception content type

### LINKED ISSUE #5816: Take controller method into account for setting content-type in @RestControllerAdvice
Currently the generated `javax.ws.rs.ext.ExceptionMapper` set `application/json` as the content type of the response when the handler method of the `@RestControllerAdvice` returns an object.

This is fine for most cases, but there can be a case where the controller method needs to return say XML (which is specified in the `@RequestMapping`) then the generated `ExceptionMapper` should not force JSON.

The proper way to handle this is to include 

`@Context ResourceInfo resourceInfo` 

in generated the `ExceptionMapper` and use that to if there are any annotations on the original controller method (that threw the exception) that set the content type. 

This is a follow up from [this](https://github.com/quarkusio/quarkus/pull/5803#discussion_r351340721) discussion

gytis:
Fixes https://github.com/quarkusio/quarkus/issues/5816

cc @geoand

### DIFF
--- a/extensions/spring-web/deployment/src/main/java/io/quarkus/spring/web/deployment/ControllerAdviceAbstractExceptionMapperGenerator.java
+++ b/extensions/spring-web/deployment/src/main/java/io/quarkus/spring/web/deployment/ControllerAdviceAbstractExceptionMapperGenerator.java
@@ -2,6 +2,7 @@ package io.quarkus.spring.web.deployment;
 
 import java.lang.annotation.Annotation;
 import java.lang.reflect.Modifier;
+import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -9,6 +10,7 @@ import java.util.Map;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 import javax.ws.rs.core.Context;
+import javax.ws.rs.core.HttpHeaders;
 import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
 
@@ -27,12 +29,19 @@ import io.quarkus.gizmo.FieldDescriptor;
 import io.quarkus.gizmo.MethodCreator;
 import io.quarkus.gizmo.MethodDescriptor;
 import io.quarkus.gizmo.ResultHandle;
+import io.quarkus.spring.web.runtime.ResponseContentTypeResolver;
 import io.quarkus.spring.web.runtime.ResponseEntityConverter;
 
 class ControllerAdviceAbstractExceptionMapperGenerator extends AbstractExceptionMapperGenerator {
 
     private static final DotName RESPONSE_ENTITY = DotName.createSimple("org.springframework.http.ResponseEntity");
-    private static final DotName STRING = DotName.createSimple(String.class.getName());
+
+    // Preferred content types order for String or primitive type responses
+    private static final List<String> TEXT_MEDIA_TYPES = Arrays.asList(
+            MediaType.TEXT_PLAIN, MediaType.APPLICATION_JSON, MediaType.APPLICATION_XML, MediaType.TEXT_XML);
+    // Preferred content types order for object type responses
+    private static final List<String> OBJECT_MEDIA_TYPES = Arrays.asList(
+            MediaType.APPLICATION_JSON, MediaType.APPLICATION_XML, MediaType.TEXT_XML, MediaType.TEXT_PLAIN);
 
     private final MethodInfo controllerAdviceMethod;
     private final TypesUtil typesUtil;
@@ -41,6 +50,8 @@ class ControllerAdviceAbstractExceptionMapperGenerator extends AbstractException
     private final String declaringClassName;
 
     private final Map<Type, FieldDescriptor> parameterTypeToField = new HashMap<>();
+
+    private FieldDescriptor httpHeadersField;
 
     ControllerAdviceAbstractExceptionMapperGenerator(MethodInfo controllerAdviceMethod, DotName exceptionDotName,
             ClassOutput classOutput, TypesUtil typesUtil) {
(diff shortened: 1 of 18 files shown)


geoand:
Thanks Gytis!

I'll take a look soon. 

Can you please squash the commits though?

gytis:
I was planning to squash them after the review, thought it make it easier for you to review the changes. But if you prefer, I could squash them now.

geoand:
You can go ahead and squash, reviewing shouldn't be a problem.

geoand:
I took a quick look at it and it looks good! But of course I'll need to spend more time with it.

extensions/spring-web/deployment/src/main/java/io/quarkus/spring/web/deployment/ControllerAdviceAbstractExceptionMapperGenerator.java
```diff
@@ -27,12 +29,17 @@
...
 import io.quarkus.spring.web.runtime.ResponseEntityConverter;
 
 class ControllerAdviceAbstractExceptionMapperGenerator extends AbstractExceptionMapperGenerator {
 
     private static final DotName RESPONSE_ENTITY = DotName.createSimple("org.springframework.http.ResponseEntity");
-    private static final DotName STRING = DotName.createSimple(String.class.getName());
+
+    private static final List<String> TEXT_MEDIA_TYPES = Arrays.asList(
+            MediaType.TEXT_PLAIN, MediaType.APPLICATION_JSON, MediaType.APPLICATION_XML, MediaType.TEXT_XML);
+    private static final List<String> OBJECT_MEDIA_TYPES = Arrays.asList(
```
geoand: I am not sure I understand the distinction betweem `TEXT_MEDIA_TYPES` and `OBJECT_MEDIA_TYPES`. They seem to contains the same values?

extensions/spring-web/deployment/src/main/java/io/quarkus/spring/web/deployment/ResponseBuilder.java
```diff
@@ -0,0 +1,46 @@
...
+
+import static io.quarkus.gizmo.MethodDescriptor.ofMethod;
+
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response;
+
+import io.quarkus.gizmo.MethodCreator;
+import io.quarkus.gizmo.ResultHandle;
+
+final class ResponseBuilder {
```
geoand: +1

extensions/spring-web/runtime/src/main/java/io/quarkus/spring/web/runtime/ResponseContentTypeResolver.java
```diff
@@ -0,0 +1,56 @@
...
+        ServerDrivenNegotiation negotiation = new ServerDrivenNegotiation();
+        negotiation.setAcceptHeaders(acceptHeaders);
+
+        return negotiation.getBestMatch(variants);
+    }
+
+    private static List<Variant> getMediaTypeVariants(String... mediaTypes) {
+        Variant.VariantListBuilder variantListBuilder = Variant.VariantListBuilder.newInstance();
+
+        Arrays.stream(mediaTypes)
```
geoand: Generally we try to stay away from lambdas in code that exists in runtime. The reason is that they introduce a small memory overhead.
Now of course this code is not on the critical, but I think we should try and get rid of the lambdas nonetheless in this case, since it will very use a for loop that will be just as readable

integration-tests/spring-web/src/main/java/io/quarkus/it/spring/web/ExceptionThrowingController.java
```diff
@@ -10,49 +11,63 @@
...
-        throw new IllegalArgumentException("hello from error");
+    @GetMapping("/re/void/xml")
+    public void voidWithResponseEntityExceptionAsXml() {
+        throw new HandledResponseEntityException("bad state", MediaType.APPLICATION_XML);
     }
 
-    @GetMapping("/re")
-    public ResponseEntity<Greeting> responseEntityWithIllegalArgumentException() {
-        throw new IllegalStateException("hello from error");
+    @GetMapping(value = "/pojo/re")
```
geoand: Perhaps drop `value`?

integration-tests/spring-web/src/main/java/io/quarkus/it/spring/web/ExceptionThrowingController.java
```diff
@@ -10,49 +11,63 @@
...
+    public ResponseEntity responseEntityWithPojoException() {
+        throw new HandledPojoException("bad state");
     }
 
+    @GetMapping(value = "/pojo/pojo")
```
geoand: Same here

integration-tests/spring-web/src/main/java/io/quarkus/it/spring/web/ExceptionThrowingController.java
```diff
@@ -10,49 +11,63 @@
...
+    public Greeting pojoWithPojoException() {
+        throw new HandledPojoException("bad state");
+    }
+
+    @GetMapping("/pojo/void")
+    public void voidWithPojoException() {
+        throw new HandledPojoException("bad state");
+    }
+
+    @GetMapping(value = "/string")
```
geoand: And here

geoand:
Seems like the imports are not sorted:

```
[ERROR] Failed to execute goal net.revelc.code:impsort-maven-plugin:1.2.0:check (check-imports) on project quarkus-spring-web: Imports are not sorted in /home/vsts/work/1/s/extensions/spring-web/runtime/src/main/java/io/quarkus/spring/web/runtime/ResponseContentTypeResolver.java -> [Help 1]
```

So what you need to do is basically do an `mvn process-resources` and the just commit the changes

extensions/spring-web/deployment/src/main/java/io/quarkus/spring/web/deployment/ControllerAdviceAbstractExceptionMapperGenerator.java
geoand: Aha, gotcha! Perhaps a comment in the code is in order? Because now it looks like a prime target for refactoring 😂

geoand:
Thanks for the updates @gytis!

I'll try and a proper review soon

gsmet:
@geoand I'll let you tell me if you want that backported.

geoand:
@gsmet I hope to be able to review tomorrow and hopefully we can get it in for backporting

geoand:
@gytis this part of the docs should also be updated: https://github.com/quarkusio/quarkus/blob/master/docs/src/main/asciidoc/spring-web.adoc#L247

extensions/spring-web/deployment/src/main/java/io/quarkus/spring/web/deployment/ControllerAdviceAbstractExceptionMapperGenerator.java
```diff
@@ -104,89 +113,87 @@ protected void preGenerateMethodBody(ClassCreator cc) {
                     "Parameter type " + parameterTypes.get(notAllowedParameterIndex).name() + " is not supported for method"
                             + controllerAdviceMethod.name() + " of class" + controllerAdviceMethod.declaringClass().name());
         }
+
+        injectHttpHeadersField(cc);
+    }
+
+    private void injectHttpHeadersField(ClassCreator classCreator) {
```
geoand: Minor complaint: I would call this `createHttpHeadersField`. The reason is that this part of the is actually creating a field in bytecode, it's not actually doing any injection which you obvisouly know happens at runtime.

geoand:
Excellent work if I may say so @gytis! Once you take care of the minor comments I have we can get this in.

gytis:
> @gytis this part of the docs should also be updated: https://github.com/quarkusio/quarkus/blob/master/docs/src/main/asciidoc/spring-web.adoc#L247

Hmm, it's a bit tricky, because we've only changed @ExceptionHandler content type for controller advice.

geoand:
> > @gytis this part of the docs should also be updated: https://github.com/quarkusio/quarkus/blob/master/docs/src/main/asciidoc/spring-web.adoc#L247
> 
> Hmm, it's a bit tricky, because we've only changed @ExceptionHandler content type for controller advice.

OK let's leave it as is for now

geoand (approved):
Thanks a lot @gytis!!!

PR #6117, opened 2019-12-11
Enroll Neo4j in native image testing

gsmet:
This will fail because we have a regression but let's get CI straight.

Let's see if CI is OK with that (at least to have the tests working then we will need to fix the regression somehow).


rsvoboda:
Neo4j would be running just in native pipeline ?
And not in JVM one - https://github.com/quarkusio/quarkus/blob/master/ci-templates/jvm-build-steps.yaml#L32

gsmet:
Superseded by what @machi1990 did here: https://github.com/quarkusio/quarkus/pull/6364 .

PR #6120, opened 2019-12-11
Introduce JSch extension

### LINKED ISSUE #6110: Move jsch bits out of quarkus-jgit
**Description**
in camel-quarkus we do use jsch as part of the camel-quarkus-ftp extensions and the same library is also used by the quarkus-jgit extension which provides some dedicated build items and substitutions so it could make sense to create a quarkus-jsch extension that can be re-used by camel-quarkus-ftp  and quarkus-jgit

gastaldi:
Fixes #6110

### DIFF
--- a/bom/runtime/pom.xml
+++ b/bom/runtime/pom.xml
@@ -173,6 +173,8 @@
         <quarkus-security.version>1.0.1.Final</quarkus-security.version>
         <keycloak.version>8.0.1</keycloak.version>
         <logstash-gelf.version>1.13.0</logstash-gelf.version>
+        <jsch.version>0.1.55</jsch.version>
+        <jzlib.version>1.1.1</jzlib.version>
     </properties>
 
     <dependencyManagement>
@@ -695,6 +697,11 @@
             <dependency>
                 <groupId>io.quarkus</groupId>
                 <artifactId>quarkus-jgit</artifactId>
+                <version>${project.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>io.quarkus</groupId>
+                <artifactId>quarkus-jsch</artifactId>
                 <version>${project.version}</version>
             </dependency>
             <dependency>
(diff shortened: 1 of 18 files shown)


lburgazzoli (commented):
LGTM

gastaldi:
I'm adding some integration tests, will let you know when it's ready

core/deployment/src/main/java/io/quarkus/deployment/builditem/FeatureBuildItem.java
```diff
@@ -31,6 +31,7 @@
     public static final String JDBC_MSSQL = "jdbc-mssql";
     public static final String JDBC_MYSQL = "jdbc-mysql";
     public static final String JGIT = "jgit";
+    public static final String JSCH = "jsch";
```
gsmet: Open question: do we really want to expose it or should it stay internal for now? My inclination would be to make it internal until someone asked for it.

bom/runtime/pom.xml
```diff
@@ -173,6 +173,8 @@
         <quarkus-security.version>1.0.1.Final</quarkus-security.version>
         <keycloak.version>8.0.1</keycloak.version>
         <logstash-gelf.version>1.13.0</logstash-gelf.version>
+        <jsch.version>0.1.55</jsch.version>
+        <jzlib.version>1.1.1</jzlib.version>
```
gsmet: Maybe we can add those 2 to dependabot? (It can be done separately)

gsmet (dismissed):
I added one suggestion and one question.

Let's discuss the question before merging.

bom/runtime/pom.xml
gastaldi: Maybe, but I rather reuse the same version used by JGit

core/deployment/src/main/java/io/quarkus/deployment/builditem/FeatureBuildItem.java
gastaldi: I'm not opposed to making it internal. So far jGit and camel-ftp were the only extensions using it. How do I make it internal BTW? Just removing the FeatureBuildItem?

gsmet: That and https://github.com/quarkusio/quarkus/blob/master/extensions/caffeine/runtime/src/main/resources/META-INF/quarkus-extension.yaml#L9 .

gastaldi: Should it be internal ? @lburgazzoli any suggestion?

gsmet: The question is basically:
- do we want users to be able to use it directly?
- do we want to expose it in the extension list at startup?

I think, for now, the answers to these questions are both no.

geoand: I am with @gsmet on this, it doesn't look like this is something that would be used directly by users.

gastaldi:
Ok, the extension should be internal now

gsmet (approved):

Issue #6115, opened 2019-12-11
Neo4j doesn't work in native mode anymore


gsmet:
Apparently, CI didn't run the Neo4j tests and we missed that the Neo4j extension doesn't work in native mode anymore:

```
2019-12-10T17:22:12.8145240Z 2019-12-10 17:22:12,810 INFO  [io.quarkus] (main) Installed features: [cdi, neo4j, resteasy, resteasy-jsonb, smallrye-health]
2019-12-10T17:22:13.9081262Z 2019-12-10 17:22:13,905 WARN  [ChannelErrorHandler] (Neo4jDriverIO-3-2) [0xf593c5b0][localhost:60513][] Fatal error occurred in the pipeline: org.neo4j.driver.internal.shaded.io.netty.handler.codec.DecoderException: Failed to read inbound message:
2019-12-10T17:22:13.9088669Z 70a2867365727665728b4e656f346a2f332e352e368d636f6e6e656374696f6e5f696486626f6c742d31
2019-12-10T17:22:13.9089433Z 
2019-12-10T17:22:13.9103746Z 	at org.neo4j.driver.internal.async.inbound.InboundMessageHandler.channelRead0(InboundMessageHandler.java:87)
2019-12-10T17:22:13.9104472Z 	at org.neo4j.driver.internal.async.inbound.InboundMessageHandler.channelRead0(InboundMessageHandler.java:35)
2019-12-10T17:22:13.9111392Z 	at org.neo4j.driver.internal.shaded.io.netty.channel.SimpleChannelInboundHandler.channelRead(SimpleChannelInboundHandler.java:105)
2019-12-10T17:22:13.9111960Z 	at org.neo4j.driver.internal.shaded.io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:374)
2019-12-10T17:22:13.9119813Z 	at org.neo4j.driver.internal.shaded.io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:360)
2019-12-10T17:22:13.9120438Z 	at org.neo4j.driver.internal.shaded.io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:352)
[...]

PR #6124, opened 2019-12-11
Update to Neo4j 4.0.0 driver, add more information to health check.

michael-simons:
This is the first final release of our 4.0 driver. The API is now published and stable.

This also brings the same features to the health check as we provide for Spring Boot.

In addition I changed the database image being used during test to our public 4.0.0.RC1 milestone, so that we can test reactive database access as well.

I would like to propose this upgrade regardless the current problems with native compilation. As of now, native compilation works both with GraalVM 19.2 and 19.3, but fails during runtimes with GraalVM 19.3, regardless of beta03 or final driver (For a discussion please have a look here https://github.com/quarkusio/quarkus/issues/6115)

We might have to update the driver in the near future, but the user facing API won't change anymore, so that Quarkus users now have a stable surface.

Thanks for considering this.

### DIFF
--- a/docs/src/main/asciidoc/neo4j.adoc
+++ b/docs/src/main/asciidoc/neo4j.adoc
@@ -4,7 +4,7 @@ and pull requests should be submitted there:
 https://github.com/quarkusio/quarkus/tree/master/docs/src/main/asciidoc
 ////
 = Quarkus - Neo4j
-:neo4j_version: 3.5.6
+:neo4j_version: 3.5.13
 
 include::./attributes.adoc[]
 :extension-status: preview
@@ -429,7 +429,7 @@ Please add the following dependency management and dependency to your `pom.xml`
       <dependency>
           <groupId>io.projectreactor</groupId>
           <artifactId>reactor-bom</artifactId>
-          <version>Californium-SR4</version>
+          <version>Dysprosium-RELEASE</version>
           <type>pom</type>
           <scope>import</scope>
       </dependency>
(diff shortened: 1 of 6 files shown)


docs/src/main/asciidoc/neo4j.adoc
```diff
@@ -429,7 +429,7 @@ Please add the following dependency management and dependency to your `pom.xml`
       <dependency>
           <groupId>io.projectreactor</groupId>
           <artifactId>reactor-bom</artifactId>
-          <version>Californium-SR4</version>
+          <version>Californium-RELEASE</version>
```
gsmet: I'm uncomfortable not using the same versions in documentation and ITs.

Considering the extension is experimental, we could point to the new Neo4j version in the doc too?

gsmet (changes requested):
Added a question.

docs/src/main/asciidoc/neo4j.adoc
michael-simons: And rightfully uncomfortable. That was an editing mistake on my side. I fixed this.

docs/src/main/asciidoc/neo4j.adoc
```diff
@@ -4,7 +4,7 @@ and pull requests should be submitted there:
 https://github.com/quarkusio/quarkus/tree/master/docs/src/main/asciidoc
 ////
 = Quarkus - Neo4j
-:neo4j_version: 3.5.6
+:neo4j_version: 3.5.13
```
gsmet: Hmmm, well, I wasn't entirely clear but my main concern was this one :).

gsmet:
I squashed the 2 commits and rebased, let's wait for CI.

gsmet (approved):

docs/src/main/asciidoc/neo4j.adoc
michael-simons: Oh sorry, I didn't get that :/

PR #6140, opened 2019-12-12
only base64 encode binary types

### LINKED ISSUE #5196: aws-lambda-http how to turn off response body base64 encoding
Hi Team,

I am using aws-lambda-http body is always base64 encoded (see sample response below) in response. Is there an option to not to do base 64 encoding?

{
	"statusCode": 200,
	"multiValueHeaders": {
		"Content-Length": ["23"],
		"Content-Type": ["application/json"]
	},
	"body": "eyJoZWxsbyI6ImhlbGxvIGpheHJzIn0=",
	"isBase64Encoded": true
}

Thanks,
Ravi

evanchooly:
This restricts the base64 encoding only to known binary types.

The one question i have is the encoding of the byte[] for non-binary types.  There was a cleaner way rather than hardcoding UTF-8 but I can't recall what that was offhand.

fixes  #5196

### DIFF
--- a/bom/runtime/pom.xml
+++ b/bom/runtime/pom.xml
@@ -2318,6 +2318,20 @@
                 <groupId>com.amazonaws.serverless</groupId>
                 <artifactId>aws-serverless-java-container-core</artifactId>
                 <version>${aws-lambda-serverless-java-container.version}</version>
+                <exclusions>
+                    <exclusion>
+                        <groupId>commons-logging</groupId>
+                        <artifactId>commons-logging</artifactId>
+                    </exclusion>
+                    <exclusion>
+                        <groupId>javax.servlet</groupId>
+                        <artifactId>javax.servlet-api</artifactId>
+                    </exclusion>
+                    <exclusion>
+                        <groupId>javax.ws.rs</groupId>
+                        <artifactId>javax.ws.rs-api</artifactId>
+                    </exclusion>
+                </exclusions>
             </dependency>
 
             <dependency>
--- a/extensions/amazon-lambda-http/runtime/src/main/java/io/quarkus/amazon/lambda/http/LambdaHttpHandler.java
+++ b/extensions/amazon-lambda-http/runtime/src/main/java/io/quarkus/amazon/lambda/http/LambdaHttpHandler.java
@@ -9,6 +9,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.concurrent.TimeUnit;
 
+import com.amazonaws.serverless.proxy.internal.LambdaContainerHandler;
 import com.amazonaws.services.lambda.runtime.Context;
 import com.amazonaws.services.lambda.runtime.RequestHandler;
 
@@ -144,8 +145,12 @@ public class LambdaHttpHandler implements RequestHandler<AwsProxyRequest, AwsPro
                 }
                 if (msg instanceof LastHttpContent) {
                     if (baos != null) {
-                        responseBuilder.setBase64Encoded(true);
-                        responseBuilder.setBody(Base64.getMimeEncoder().encodeToString(baos.toByteArray()));
+                        if (isBinary(responseBuilder.getMultiValueHeaders().getFirst("Content-Type"))) {
+                            responseBuilder.setBase64Encoded(true);
+                            responseBuilder.setBody(Base64.getMimeEncoder().encodeToString(baos.toByteArray()));
+                        } else {
+                            responseBuilder.setBody(new String(baos.toByteArray(), "UTF-8"));
+                        }
                     }
                     return responseBuilder;
                 }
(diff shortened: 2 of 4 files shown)


bom/runtime/pom.xml
```diff
@@ -2292,6 +2292,20 @@
                 <groupId>com.amazonaws.serverless</groupId>
                 <artifactId>aws-serverless-java-container-core</artifactId>
                 <version>${aws-lambda-serverless-java-container.version}</version>
+                <exclusions>
+                    <exclusion>
+                        <groupId>commons-logging</groupId>
```
dmlloyd: Isn't this artifact banned?  We'd want (iirc) `org.jboss.logging:commons-logging-jboss-logging` (or maybe `-logmanager`, I forget which one we're using`.

evanchooly: without the exclusion I get enforcer errors.

dmlloyd: Oh! My bad, I'm reading this on the go and didn't notice that it's an exclusion, not a dependency.  Carry on!

bom/runtime/pom.xml
```diff
@@ -2292,6 +2292,20 @@
...
+                        <artifactId>commons-logging</artifactId>
+                    </exclusion>
+                    <exclusion>
+                        <groupId>javax.servlet</groupId>
+                        <artifactId>javax.servlet-api</artifactId>
+                    </exclusion>
+                    <exclusion>
+                        <groupId>javax.ws.rs</groupId>
+                        <artifactId>javax.ws.rs-api</artifactId>
+                    </exclusion>
```
gsmet: You don't need to add the replacement dependencies when you use this artifact?

gsmet:
@evanchooly I think it's mostly OK and should probably be backported. Should someone else review it?

BTW, could you squash everything into one commit?

bom/runtime/pom.xml
baztian: Why do you need to exclude these two dependencies?

extensions/amazon-lambda-http/runtime/src/main/java/io/quarkus/amazon/lambda/http/LambdaHttpHandler.java
```diff
@@ -9,6 +9,7 @@
 import java.util.Map;
 import java.util.concurrent.TimeUnit;
 
+import com.amazonaws.serverless.proxy.internal.LambdaContainerHandler;
```
baztian: I'm not an expert on this one but `internal` sounds to be something that might get changed without any further deprecation phase.
Not completely sure but looking at the aws code it is actually nothing more than checking if the content type is either application/octet-stream, image/jpeg, image/png or image/gif

bom/runtime/pom.xml
gsmet: We are using Jakarta dependencies so we don't want the javax ones.

extensions/amazon-lambda-http/runtime/src/main/java/io/quarkus/amazon/lambda/http/LambdaHttpHandler.java
evanchooly: That check comes from the java serverless project on amazon's github account.  It's probably safe enough.  I used it to mirror the same checks they do for their own integrations.

bom/runtime/pom.xml
evanchooly: nope!

evanchooly:
@gsmet squashed

baztian:
I think the documentation should also reflect the changes.

gsmet:
@evanchooly could you add something to the documentation? I think @baztian is right and a note in the doc would be better to indicate when we encode and when not.

I plan to wrap up 1.1 sometimes today.

Thanks.

evanchooly:
@gsmet doc'd and squashed.

gsmet (approved):

gsmet:
Thanks!

baztian:
Regarding the sam.yaml example in the docs and the archetype. Is it still correct/necessary to have the BinaryMediaTypes setting?

Ravisankar-Challa:
How to deal with other binary content types (eg: application/pdf)

Ravisankar-Challa:
> 
> 
> Regarding the sam.yaml example in the docs and the archetype. Is it still correct/necessary to have the BinaryMediaTypes setting?

Yes it is required. I have tested this setting in AWS API Gateway. BinaryMediaTypes  */* setting and the field 'isBase64Encoded' in json (LambdaHttpHandler.java -> vert.x response) tells api gateway whether to decode the response body or not.

extensions/amazon-lambda-http/runtime/src/main/java/io/quarkus/amazon/lambda/http/LambdaHttpHandler.java
Ravisankar-Challa: In future versions please consider providing a new quarkus application property for binary content types apart from the default ones (application/octet-stream, image/jpeg, image/png or image/gif). Generating pdf is a real world use case (application/pdf )

evanchooly:
i have a subsequent patch i just ran out of time to submit that broadens that check.  I'll get it in probably on monday.

gsmet:
@evanchooly it would be nice if we could include it in 1.1 so having it today would be cool.

patriot1burke:
Don't forget that any string media type that is not UTF-8 most be encoded.  For example, the built in text/plain provider for resteasy assumes a specific ISO format, NOT UTF-8.  JSON needs to be encoded as well.  In fact, JSON tests should be added to make sure you don't break anything.

evanchooly:
@patriot1burke noted, thanks.  I'll look in to that today.

PR #6167, opened 2019-12-13
Let the delegate handlers decide to autoflush

dmlloyd:
Fixes the problem where Vert.x threads can block even if async logging is enabled.


gsmet:
Looks like something we would want to include in 1.1. Let me know how it goes.

dmlloyd:
Looks like it does solve the issue.

gsmet (approved):

PR #6174, opened 2019-12-14
Add Support to Multiple Vault KV Paths

### LINKED ISSUE #5638: [Vault]: Add Support to Multiple Vault KV Paths
It would be nice if the `quarkus.vault.secret-config-kv-path` property supported multiple vault kv-path entries.

vsevel:
Fixes #5638.

This provides the ability to express 
 - multiple paths for property `quarkus.vault.secret-config-kv-path` and
 - provide an optional property prefix

Consider for instance those vault paths containing secrets:
```
multi/default1:
  color=blue
  size=XL

multi/default2:
  color=red
  weight=3

multi/singer1:
  firstname=paul
  lastname=shaffer

multi/singer2:
  color=green
  lastname=simon
  age=78
```

And the following `application.properties`:
```
quarkus.vault.secret-config-kv-path=multi/default1,multi/default2
quarkus.vault.secret-config-kv-path.singer=multi/singer1,multi/singer2
```

Then the quarkus application would see the following properties:
```
color=red
size=XL
weight=3

singer.color=green
singer.firstname=paul
singer.lastname=simon
singer.age=78
```

If multiple paths contain the same key for a particular prefix, the last path will win (going from the most general, to the most specific). This rule applies also on the _no prefix_ property, but not in between different prefix zones.

In the example above the following overrides apply:
 - `color=red` from `multi/default2` overrides `color=blue` from `multi/default1`
 - `singer.lastname=simon` from `multi/singer2` overrides `singer.lastname=shaffer` from `multi/singer1`
 - `singer.color=green` from `multi/singer2` does **not** override `color=red` from `multi/default2`, because they are in different prefix zones

cc @sberyozkin @valones @gsmet

### DIFF
--- a/extensions/vault/runtime/src/main/java/io/quarkus/vault/runtime/config/VaultConfigSource.java
+++ b/extensions/vault/runtime/src/main/java/io/quarkus/vault/runtime/config/VaultConfigSource.java
@@ -14,14 +14,18 @@ import static io.quarkus.vault.runtime.config.VaultRuntimeConfig.DEFAULT_TLS_USE
 import static io.quarkus.vault.runtime.config.VaultRuntimeConfig.KV_SECRET_ENGINE_VERSION_V1;
 import static java.lang.Boolean.parseBoolean;
 import static java.lang.Integer.parseInt;
+import static java.util.Collections.emptyMap;
+import static java.util.Optional.empty;
+import static java.util.stream.Collectors.toList;
 import static java.util.stream.Collectors.toMap;
 
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.time.Duration;
 import java.util.AbstractMap.SimpleEntry;
-import java.util.Collections;
+import java.util.Arrays;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
@@ -48,6 +52,8 @@ public class VaultConfigSource implements ConfigSource {
 
     private static final String PROPERTY_PREFIX = "quarkus.vault.";
     public static final Pattern CREDENTIAL_PATTERN = Pattern.compile("^quarkus\\.vault\\.credentials-provider\\.([^.]+)\\.");
+    public static final Pattern SECRET_CONFIG_KV_PATH_PATTERN = Pattern
+            .compile("^quarkus\\.vault\\.secret-config-kv-path\\.([^.]+)$");
 
     private AtomicReference<VaultCacheEntry<Map<String, String>>> cache = new AtomicReference<>(null);
     private AtomicReference<VaultRuntimeConfig> serverConfig = new AtomicReference<>(null);
@@ -76,7 +82,7 @@ public class VaultConfigSource implements ConfigSource {
      */
     @Override
     public Map<String, String> getProperties() {
-        return Collections.emptyMap();
+        return emptyMap();
     }
 
     @Override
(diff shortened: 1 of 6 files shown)


sberyozkin:
@vsevel it looks good, what about [VaultKVSecretEngine](https://github.com/quarkusio/quarkus/blob/master/extensions/vault/runtime/src/main/java/io/quarkus/vault/VaultKVSecretEngine.java) ?

vsevel:
> @vsevel it looks good, what about [VaultKVSecretEngine](https://github.com/quarkusio/quarkus/blob/master/extensions/vault/runtime/src/main/java/io/quarkus/vault/VaultKVSecretEngine.java) ?

hello @sberyozkin thanks for your feedback. what do you think is missing in VaultKVSecretEngine? something like this?
```
Map<String, String> readSecret(String... paths);
```

sberyozkin:
@vsevel Hi, I'm not sure what would go into a single Map if multiple paths are available ?

May be there should be a new method 
```
Map<String, Map<String, String>> readSecret(Set<String> paths)
```
?

vsevel:
> @vsevel Hi, I'm not sure what would go into a single Map if multiple paths are available ?

@sberyozkin why not, if we think it is missing. but I see that as independent from the present PR, which was targeting specifically the vault config source.

so we are talking about this, right?
```
    @Override
    public Map<String, Map<String, String>> readSecrets(Set<String> paths) {
        return paths.stream().collect(toMap(identity(), this::readSecret));
    }
```
Note: it should be plural, because each path is a secret.

I think the user could as easily do it on his own. but if we think that this use case will come very often, then may be it is worth make it easier for users. do you think this is a common use case?

sberyozkin:
@vsevel my understanding that these engine interfaces are for the users who want to deal with Vault directly (as opposed to relying on the injection of the config properties) so I thought it would make sense to get the multi-paths supported at that level too for us to assert the multi-paths are fully supported ? If you agree then indeed a plural form is good :-). 
thanks

sberyozkin:
@vsevel Lets decide on whether `VaultKVSecretEngine` should be updated or not and then I can merge the PR. I have a minor preference to have it supported at the interface level too but since adding it alter won't break any user code, we can just indeed do it later, so it is your choice :-)

vsevel:
> @vsevel Lets decide on whether `VaultKVSecretEngine` should be updated or not and then I can merge the PR. I have a minor preference to have it supported at the interface level too but since adding it later won't break any user code, we can just indeed do it later, so it is your choice :-)

hello @sberyozkin 
indeed I would leave it aside from this PR. I do not see a common "multiple kv paths" theme across the vault config source and the kv secret engine. I think both serve different use cases. and I am not convinced a `Map<String, Map<String, String>> readSecrets(Set<String>)` method is adding a lot of value for the user: the code can be easily written, and we are not providing any magic beside standard java stream stuff. I could have been tempted by a `Map<String, String> readSecrets(Set<String>)` that provides overriding, hence some level of usefulness. but even there I do not see it as adding a lot of value, not mentioning being trivial to implement by the user. for these different reasons, I would rather resist adding new methods to the api, until we encounter real life use cases we can discuss with users.

sberyozkin (approved):

sberyozkin:
@vsevel I'm not sure if the tests have actually run, because there was a "WIP" in the title. Can you please rebase and push again just to get the tests running ? CC Guillaume @gsmet

vsevel:
@sberyozkin I have renamed and squashed the commits. waiting for the CI to complete.
thank for you continuous support :)

sberyozkin:
@vsevel np, Vault is good for Quarkus and thanks for continuing with the quality Vault contributions :-). Looks like the CI managed to finish well before the current CI issues have surfaced. So I'll just go ahead with the merge, @gsmet hope you good with it :-)

PR #6177, opened 2019-12-14
[WIP] Use optional config groups in Hibernate ORM configuration

### LINKED ISSUE #5608: Hibernate ORM should use new optional config groups
**Description**
After #5387 is merged, we gain support for optional config groups.  Optional config groups, as explained in the docs, allow a config group _as a whole_ to be considered optional, which also implies that the required properties in the group are _only_ required when _any other_ property within the group or its nested children are specified.

This should fit the Hibernate ORM configuration well.  Presently the Hibernate ORM configuration uses many individually-Optional properties with a query method typically named `isAnyPropertySet`, coupled with manual validation of combinations of properties.  This extra validation step is ripe to be omitted, with a resultant savings in code complexity.  Instead, it can simply query for the presence of the various configuration groups to determine if any of the corresponding properties have been set.

/cc @Sanne

gwenneg:
Fixes #5608

### DIFF
--- a/extensions/hibernate-orm/deployment/src/main/java/io/quarkus/hibernate/orm/deployment/HibernateOrmConfig.java
+++ b/extensions/hibernate-orm/deployment/src/main/java/io/quarkus/hibernate/orm/deployment/HibernateOrmConfig.java
@@ -3,6 +3,7 @@ package io.quarkus.hibernate.orm.deployment;
 import java.time.Duration;
 import java.util.Map;
 import java.util.Optional;
+import java.util.OptionalInt;
 import java.util.OptionalLong;
 
 import io.quarkus.runtime.annotations.ConfigDocSection;
@@ -77,13 +78,9 @@ public class HibernateOrmConfig {
 
     /**
      * The size of the batches used when loading entities and collections.
-     *
-     * `-1` means batch loading is disabled. This is the default.
-     *
-     * @asciidoclet
      */
-    @ConfigItem(defaultValue = "-1")
-    public int batchFetchSize;
+    @ConfigItem
+    public OptionalInt batchFetchSize;
 
     /**
      * Pluggable strategy contract for applying physical naming rules for database object names.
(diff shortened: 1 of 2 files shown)


gwenneg:
This was on a branch for a while, I'll finish it during the week-end.

I tried to get rid of `HibernateOrmConfig.isAnyPropertySet()` a few weeks ago but it didn't end well. I'll try again and then promote the PR for a review.

extensions/hibernate-orm/deployment/src/main/java/io/quarkus/hibernate/orm/deployment/HibernateOrmConfig.java
```diff
@@ -78,40 +78,40 @@
     /**
      * The size of the batches used when loading entities and collections.
      *
-     * `-1` means batch loading is disabled. This is the default.
+     * `0` means batch loading is disabled. This is the default.
      *
      * @asciidoclet
      */
-    @ConfigItem(defaultValue = "-1")
```
gwenneg: This needs extra attention. I don't see any reason to use `-1` here but maybe I missed something.

gwenneg:
I'll fix the build failure (which is certainly related) tonight.

extensions/hibernate-orm/deployment/src/main/java/io/quarkus/hibernate/orm/deployment/HibernateOrmConfig.java
dmlloyd: Maybe it should be an `OptionalInt`?

extensions/hibernate-orm/deployment/src/main/java/io/quarkus/hibernate/orm/deployment/HibernateOrmConfig.java
```diff
@@ -131,10 +131,10 @@ public boolean isAnyPropertySet() {
...
                 batchFetchSize > 0 ||
                 statistics ||
-                query.isAnyPropertySet() ||
-                database.isAnyPropertySet() ||
-                jdbc.isAnyPropertySet() ||
-                log.isAnyPropertySet() ||
+                query.isPresent() ||
+                database.isPresent() ||
+                jdbc.isPresent() ||
+                log.isPresent() ||
```
dmlloyd: I suppose the next step would be to see if we can make some of the properties in the optional groups be individually non-optional.

gwenneg:
The failures are caused by the optional config groups not being populated while it seems they should be. I'll dig this further tomorrow.

gsmet:
This is out of date. Please open a new PR if you want to pursue this.

PR #6191, opened 2019-12-16
Fix #5885: generics break panache repo enhancer

### LINKED ISSUE #5885: Abstraction layer over PanacheRepositoryBase
**Describe the bug**
When implementing an abstraction over the PanacheRepositoryBase, I get the error when calling findById method:

org.hibernate.UnknownEntityTypeException: Unable to locate persister: org.acme.base.AbstractBaseLongEntity

**Expected behavior**
findById() should work when using an Abstraction layer, as findAll() just works.

**Actual behavior**

**To Reproduce**
1. clone this repo: https://github.com/lssoares/panache-abstraction-problem
2. run ./mvn clean package && mvn quarkus:dev 
3. Use swagger to execute the following webservice: http://0.0.0.0:8080/example/1

**Additional context**
1.Calling the endpoint that uses findAll() works fine: http://0.0.0.0:8080/example    
2. If you edit ExampleRepository.java and swap the lines indicated there, both endpoints will work.

FroMage:
Do not enhance generic/abstract repos
Use proper generics resolution when looking up the repository type
arguments

Fixes: #5885

### DIFF
--- a/extensions/panache/hibernate-orm-panache/deployment/src/main/java/io/quarkus/hibernate/orm/panache/deployment/PanacheResourceProcessor.java
+++ b/extensions/panache/hibernate-orm-panache/deployment/src/main/java/io/quarkus/hibernate/orm/panache/deployment/PanacheResourceProcessor.java
@@ -1,6 +1,5 @@
 package io.quarkus.hibernate.orm.panache.deployment;
 
-import java.lang.reflect.Modifier;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
@@ -31,6 +30,7 @@ import io.quarkus.panache.common.deployment.EntityField;
 import io.quarkus.panache.common.deployment.EntityModel;
 import io.quarkus.panache.common.deployment.MetamodelInfo;
 import io.quarkus.panache.common.deployment.PanacheFieldAccessEnhancer;
+import io.quarkus.panache.common.deployment.PanacheRepositoryEnhancer;
 
 public final class PanacheResourceProcessor {
 
@@ -83,12 +83,12 @@ public final class PanacheResourceProcessor {
             // Skip PanacheRepository
             if (classInfo.name().equals(DOTNAME_PANACHE_REPOSITORY))
                 continue;
-            if (skipRepository(classInfo))
+            if (PanacheRepositoryEnhancer.skipRepository(classInfo))
                 continue;
             daoClasses.add(classInfo.name().toString());
         }
         for (ClassInfo classInfo : index.getIndex().getAllKnownImplementors(DOTNAME_PANACHE_REPOSITORY)) {
-            if (skipRepository(classInfo))
+            if (PanacheRepositoryEnhancer.skipRepository(classInfo))
                 continue;
             daoClasses.add(classInfo.name().toString());
         }
@@ -127,12 +127,4 @@ public final class PanacheResourceProcessor {
             }
         }
     }
-
-    private boolean skipRepository(ClassInfo classInfo) {
-        // we don't want to add methods to abstract/generic entities/repositories: they get added to bottom types
-        // which can't be either
-        return Modifier.isAbstract(classInfo.flags())
-                || !classInfo.typeParameters().isEmpty();
-    }
-
 }
--- a/extensions/panache/hibernate-orm-panache/deployment/src/main/java/io/quarkus/hibernate/orm/panache/deployment/PanacheResourceProcessor.java
+++ b/extensions/panache/hibernate-orm-panache/deployment/src/main/java/io/quarkus/hibernate/orm/panache/deployment/PanacheResourceProcessor.java
@@ -1,5 +1,6 @@
 package io.quarkus.hibernate.orm.panache.deployment;
 
+import java.lang.reflect.Modifier;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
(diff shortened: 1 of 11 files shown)


FroMage:
@loicmathieu it's very possible you have the same bug in mongo-panache, and will need the same fix (do not enhance abstract/generic repos).

It's also possible that you have the same issue in your reactive-mongo-panache repo IF your entity base type is generic.

geoand (approved):
:+1:

extensions/panache/hibernate-orm-panache/deployment/src/main/java/io/quarkus/hibernate/orm/panache/deployment/PanacheResourceProcessor.java
```diff
@@ -122,4 +128,11 @@ void build(CombinedIndexBuildItem index,
         }
     }
 
+    private boolean skipRepository(ClassInfo classInfo) {
```
loicmathieu: Maybe rename to `isNotLeafType` and put it in the common processor ?

extensions/panache/panache-common/deployment/src/main/java/io/quarkus/panache/common/deployment/PanacheRepositoryEnhancer.java
```diff
@@ -90,15 +90,13 @@ private String recursivelyFindEntityTypeFromClass(DotName clazz, DotName reposit
...
-            final ClassInfo classByName = indexView.getClassByName(clazz);
-            for (org.jboss.jandex.Type type : classByName.interfaceTypes()) {
-                if (type.name().equals(repositoryDotName)) {
-                    org.jboss.jandex.Type entityType = type.asParameterizedType().arguments().get(0);
-                    return entityType.name().toString().replace('.', '/');
-                }
-            }
-
-            return recursivelyFindEntityTypeFromClass(classByName.superName(), repositoryDotName);
+            List<org.jboss.jandex.Type> typeParameters = io.quarkus.deployment.util.JandexUtil
```
loicmathieu: `io.quarkus.deployment.util.JandexUtil` maybe use import statement ?

loicmathieu (changes requested):
`mongodb-panache` and the future `mongodb-panache-reactive` suffers from the same bug.
If you don't have time to fix it in `mongodb-panache`inside the same PR, can you open an issue for it ?

extensions/panache/panache-common/deployment/src/main/java/io/quarkus/panache/common/deployment/PanacheRepositoryEnhancer.java
FroMage: Well there are more than one…

extensions/panache/hibernate-orm-panache/deployment/src/main/java/io/quarkus/hibernate/orm/panache/deployment/PanacheResourceProcessor.java
FroMage: I can pull it up, but I wanted the name to be clear that this is only for repositories.

FroMage:
> If you don't have time to fix it in `mongodb-panache` inside the same PR, can you open an issue for it ?

I can do it in this PR, should be simple.

extensions/panache/panache-common/deployment/src/main/java/io/quarkus/panache/common/deployment/PanacheRepositoryEnhancer.java
loicmathieu: There are more than one `JandexUtil` used inside the enhancer ? Stange ...

extensions/panache/hibernate-orm-panache/deployment/src/main/java/io/quarkus/hibernate/orm/panache/deployment/PanacheResourceProcessor.java
FroMage: moved up.

FroMage:
Added test/fix for mongo too.

loicmathieu (approved):
LGTM

FroMage:
Huh, so native test fails, due to who knows what, but I am already rebased on master.

FroMage:
It passed after restart

MartinX3:
@FroMage Here we have the problem, that a Repository listAll() call gives us every DataBase entry in every table.
As example we made a trainer and the call for courses gives us all trainer, too.

`abstract class BaseRepository<T>(@Inject val entityManager: EntityManager) : PanacheRepositoryBase<T, UUID> {`

`@Dependent
class TrainerRepository(entityManager: EntityManager) : BaseRepository<Trainer>(entityManager) {`

`@Dependent
class CourseRepository(entityManager: EntityManager) : BaseRepository<Course>(entityManager) {`

FroMage:
@MartinX3 is this with the current master which has the fix? I think it's the same bug and it should be fixed.

MartinX3:
@FroMage Thank you for your fast response.
We use the maven 1.1.0.final release in our company.
Nice to hear that it will get fixed :)
Will it be released as 1.1.1 or do we need to wait for 1.2.0 (which has no ETA?)

Thank you all for the fix. :)

FroMage:
That's a question for @gsmet but IIRC 1.2 is planned for January.

PR #6194, opened 2019-12-16
Qute - escape expressions in HTML by default

### LINKED ISSUE #6155: Handle escaping in Qute
Right now, Qute doesn't handle escaping at all and doesn't provide any facility for escaping. This is an issue as that means that users will easily have XSS issues.

My inclination would be to enable escaping by default.

The escaping should be different for each content-type.

We need to check how we will properly handle tag content escaping and attribute escaping in HTML. We need to be sure our escaping works for both and a value escaped in an attribute still has the same value as previously when submitted through a form (think of `input value=""`: if we inject a value here, we want the submitted value to be exactly the same as the injected one).

mkouba:
- resolves #6155
- API changes:
 - introduce TemplateLocator
 - add Template#getVariant()

What this PR does?

1. For HTML templates `'`,`"`,`<`, `>` and `&` are escaped by default
2. It's possilble to use `raw` and `safe` extension methods for String to "unescape" the value; ie. `{foo.myString.raw}`
3. `RawString` is never escaped
4. `Template` now has an optional `Variant`

### DIFF
--- a/extensions/qute/runtime/src/main/java/io/quarkus/qute/runtime/EngineProducer.java
+++ b/extensions/qute/runtime/src/main/java/io/quarkus/qute/runtime/EngineProducer.java
@@ -1,8 +1,9 @@
 package io.quarkus.qute.runtime;
 
-import java.io.InputStream;
+import java.io.IOException;
 import java.io.InputStreamReader;
 import java.io.Reader;
+import java.net.URL;
 import java.nio.charset.Charset;
 import java.util.List;
 import java.util.Optional;
@@ -19,12 +20,19 @@ import io.quarkus.arc.Arc;
 import io.quarkus.arc.InstanceHandle;
 import io.quarkus.qute.Engine;
 import io.quarkus.qute.EngineBuilder;
+import io.quarkus.qute.Escaper;
+import io.quarkus.qute.Expression;
 import io.quarkus.qute.NamespaceResolver;
+import io.quarkus.qute.RawString;
 import io.quarkus.qute.ReflectionValueResolver;
+import io.quarkus.qute.ResultMapper;
 import io.quarkus.qute.Results.Result;
+import io.quarkus.qute.TemplateLocator.TemplateLocation;
+import io.quarkus.qute.TemplateNode.Origin;
 import io.quarkus.qute.UserTagSectionHelper;
 import io.quarkus.qute.ValueResolver;
 import io.quarkus.qute.ValueResolvers;
+import io.quarkus.qute.Variant;
 
 @Singleton
 public class EngineProducer {
@@ -59,8 +67,31 @@ public class EngineProducer {
 
         // We don't register the map resolver because of param declaration validation
         // See DefaultTemplateExtensions
-        builder.addValueResolvers(ValueResolvers.thisResolver(), ValueResolvers.orResolver(), ValueResolvers.trueResolver(),
-                ValueResolvers.collectionResolver(), ValueResolvers.mapperResolver(), ValueResolvers.mapEntryResolver());
+        builder.addValueResolver(ValueResolvers.thisResolver());
+        builder.addValueResolver(ValueResolvers.orResolver());
+        builder.addValueResolver(ValueResolvers.trueResolver());
+        builder.addValueResolver(ValueResolvers.collectionResolver());
+        builder.addValueResolver(ValueResolvers.mapperResolver());
+        builder.addValueResolver(ValueResolvers.mapEntryResolver());
+        // foo.string.raw returns a RawString which is never escaped
+        builder.addValueResolver(ValueResolvers.rawResolver());
+
+        // Escape some characters for HTML templates
+        Escaper htmlEscaper = Escaper.builder().add('"', "&quot;").add('\'', "&#39;")
+                .add('&', "&amp;").add('<', "&lt;").add('>', "&gt;").build();
+        builder.addResultMapper(new ResultMapper() {
+
+            @Override
+            public boolean appliesTo(Origin origin, Object result) {
+                return !(result instanceof RawString)
+                        && origin.getVariant().filter(EngineProducer::requiresDefaultEscaping).isPresent();
+            }
+
+            @Override
+            public String map(Object result, Expression expression) {
+                return htmlEscaper.escape(result.toString());
+            }
+        });
 
         // Fallback reflection resolver
         builder.addValueResolver(new ReflectionValueResolver());
@@ -132,37 +163,79 @@ public class EngineProducer {
      * @param path
      * @return the optional reader
      */
-    private Optional<Reader> locate(String path) {
-        InputStream in = null;
+    private Optional<TemplateLocation> locate(String path) {
+        URL resource = null;
         // First try to locate a tag template
         if (tags.stream().anyMatch(tag -> tag.startsWith(path))) {
             LOGGER.debugf("Locate tag for %s", path);
-            in = locatePath(tagPath + path);
+            resource = locatePath(tagPath + path);
             // Try path with suffixes
             for (String suffix : suffixes) {
-                in = locatePath(tagPath + path + "." + suffix);
-                if (in != null) {
+                resource = locatePath(tagPath + path + "." + suffix);
+                if (resource != null) {
                     break;
                 }
             }
         }
-        if (in == null) {
+        if (resource == null) {
             String templatePath = basePath + path;
             LOGGER.debugf("Locate template for %s", templatePath);
-            in = locatePath(templatePath);
+            resource = locatePath(templatePath);
         }
-        if (in != null) {
-            return Optional.of(new InputStreamReader(in, Charset.forName("utf-8")));
+        if (resource != null) {
+            return Optional.of(new ResourceTemplateLocation(resource, guessVariant(path)));
         }
         return Optional.empty();
     }
 
-    private InputStream locatePath(String path) {
+    private URL locatePath(String path) {
         ClassLoader cl = Thread.currentThread().getContextClassLoader();
         if (cl == null) {
             cl = EngineProducer.class.getClassLoader();
         }
-        return cl.getResourceAsStream(path);
+        return cl.getResource(path);
+    }
+
+    static Variant guessVariant(String path) {
+        // TODO we need a proper way to detect the variant
+        int suffixIdx = path.lastIndexOf('.');
+        if (suffixIdx != -1) {
+            String suffix = path.substring(suffixIdx);
+            return new Variant(null, VariantTemplateProducer.parseMediaType(suffix), null);
+        }
+        return null;
+    }
+
+    static boolean requiresDefaultEscaping(Variant variant) {
+        return variant.mediaType != null
+                ? (Variant.TEXT_HTML.equals(variant.mediaType) || Variant.TEXT_XML.equals(variant.mediaType))
+                : false;
+    }
+
+    static class ResourceTemplateLocation implements TemplateLocation {
+
+        private final URL resource;
+        private final Optional<Variant> variant;
+
+        public ResourceTemplateLocation(URL resource, Variant variant) {
+            this.resource = resource;
+            this.variant = Optional.ofNullable(variant);
+        }
+
+        @Override
+        public Reader read() {
+            try {
+                return new InputStreamReader(resource.openStream(), Charset.forName("utf-8"));
+            } catch (IOException e) {
+                return null;
+            }
+        }
+
+        @Override
+        public Optional<Variant> getVariant() {
+            return variant;
+        }
+
     }
 
 }
--- a/extensions/qute/runtime/src/main/java/io/quarkus/qute/runtime/VariantTemplateProducer.java
+++ b/extensions/qute/runtime/src/main/java/io/quarkus/qute/runtime/VariantTemplateProducer.java
@@ -7,6 +7,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.CompletionStage;
 import java.util.function.Consumer;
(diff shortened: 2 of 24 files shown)


gsmet:
I would lean towards including that one into 1.1. I'm not very happy to release a templating engine without default escaping. WDYT?

Any chance you could add some documentation about this behavior?

mkouba:
> I would lean towards including that one into 1.1

I thought that 1.1 is released already ;-). If it's easy to backport (and I believe it should be because we did not change in Qute since 1.1 code freeze) then +1.

> Any chance you could add some documentation about this behavior?

+1, I'll mention this in the docs.

extensions/qute/runtime/src/main/java/io/quarkus/qute/runtime/VariantTemplateProducer.java
```diff
@@ -166,8 +172,8 @@ public TemplateVariants(Map<Variant, String> variants, String defaultTemplate) {
     }
 
     static String parseMediaType(String suffix) {
-        // TODO support more media types...
-        if (suffix.equalsIgnoreCase(".html")) {
+        // TODO we need a proper way to parse the media type
+        if (suffix.equalsIgnoreCase(".html") || suffix.equalsIgnoreCase(".htm")) {
             return "text/html";
         } else if (suffix.equalsIgnoreCase(".xml")) {
             return "text/xml";
```
gsmet: Shouldn't we apply that one for XML too?

mkouba: Yes, we could.

independent-projects/qute/core/src/main/java/io/quarkus/qute/Escaper.java
```diff
@@ -0,0 +1,85 @@
...
+     * @param value
+     * @return an escaped value
+     */
+    public String escape(CharSequence value) {
+        Objects.requireNonNull(value);
+        if (value.length() == 0) {
+            return value.toString();
+        }
+        for (int i = 0; i < value.length(); i++) {
+            String replacement = replacements.get(value.charAt(i));
```
gastaldi: The auto-boxing here looks like an expensive operation to be done on every character.

mkouba: Well, `Map#get()` is probably even more expensive. Anyway, it's an implementation detail that can be improved in the future... feel free to improve the algorithm ;-).

gastaldi: Yeah, I was looking at https://www.eclipse.org/collections/javadoc/10.0.0/org/eclipse/collections/impl/map/mutable/primitive/CharObjectHashMap.html. I'll wait until your PR is merged to perform some performance tests using it ;)

mkouba: There is also `io.netty.util.collection.CharObjectHashMap<V>`.

gastaldi: Yeah, I've been looking at that too initially. Maybe it's simple to just copy it over to Qute than to depend on another library. OSS FTW :)

mkouba:
Hm, I've been thinking about the implementation and it's very likely not correct. The `ResultMapper` that performs escaping is only applied to expression values that resolve to a `String`. However, the escaping should be performed _after_ the value is converted to `String`. 

I'll need to change the default `ResultMapper` to convert everything to String first (if needed) and then escape the value. Custom `ResultMapper` can be still implemented if it's using higher `getPriority()`.

independent-projects/qute/core/src/main/java/io/quarkus/qute/Escaper.java
```diff
@@ -0,0 +1,85 @@
...
+                return doEscape(value, i, new StringBuilder(value.subSequence(0, i)).append(replacement));
+            }
+        }
+        return value.toString();
+    }
+
+    private String doEscape(CharSequence value, int index, StringBuilder builder) {
+        int length = value.length();
+        while (++index < length) {
+            char c = value.charAt(index);
```
FroMage: Watch out that this is not codepoint-compliant. Not sure if 16-bit components of 32-bit codepoints can be in the ascii plane, but we should check.

mkouba: I have no idea what you're talking about ;-). Chars in Java are unicode, or? I would be more than grateful if you could check whatever needs to be checked.

FroMage: chars in java are "half" unicode. they're 16 bits, so for any "unicode char" (which is called a "codepoint") that requires more than 16 bits they're encoded on two consecutive 16-bit chars using UTF-16 (standard encoding of all Java Strings at runtime).

FroMage: Thus iterating chars for comparison is rarely right: you must use the codepoint static methods on `Character` to ask if a codepoint is on 1 or 2 chars. Now, it's possible that the 2-char codepoints have both chars greater than the ascii range, and you're only escaping (5) ascii values so your code may work, but we should check.

FroMage: Both surrogate pairs have their sign bits at 1 so they can't be mistaken for the ascii we're replacing, so we're good.

extensions/qute/runtime/src/main/java/io/quarkus/qute/runtime/EngineProducer.java
```diff
@@ -59,8 +67,31 @@ void init(QuteConfig config, List<String> resolverClasses, List<String> template
...
+        builder.addValueResolver(ValueResolvers.orResolver());
+        builder.addValueResolver(ValueResolvers.trueResolver());
+        builder.addValueResolver(ValueResolvers.collectionResolver());
+        builder.addValueResolver(ValueResolvers.mapperResolver());
+        builder.addValueResolver(ValueResolvers.mapEntryResolver());
+        // foo.string.raw returns a RawString which is never escaped
+        builder.addValueResolver(ValueResolvers.rawResolver());
+
+        // Escape some characters for HTML templates
+        Escaper htmlEscaper = Escaper.builder().add('"', "&quot;").add('\'', "&#39;")
```
FroMage: https://en.wikipedia.org/wiki/List_of_XML_and_HTML_character_entity_references#Predefined_entities_in_XML says `&apos;` is available for both xml and html.

extensions/qute/runtime/src/main/java/io/quarkus/qute/runtime/EngineProducer.java
```diff
@@ -59,8 +67,31 @@ void init(QuteConfig config, List<String> resolverClasses, List<String> template
...
+                .add('&', "&amp;").add('<', "&lt;").add('>', "&gt;").build();
+        builder.addResultMapper(new ResultMapper() {
+
+            @Override
+            public boolean appliesTo(Origin origin, Object result) {
+                return !(result instanceof RawString)
+                        && origin.getVariant().filter(EngineProducer::requiresDefaultEscaping).isPresent();
```
FroMage: I'm guessing in the future we'll need to turn this into an SPI for custom escaping of other variant content types. But it can wait.

extensions/qute/runtime/src/main/java/io/quarkus/qute/runtime/EngineProducer.java
```diff
@@ -59,8 +67,31 @@ void init(QuteConfig config, List<String> resolverClasses, List<String> template
...
+        builder.addValueResolver(ValueResolvers.collectionResolver());
+        builder.addValueResolver(ValueResolvers.mapperResolver());
+        builder.addValueResolver(ValueResolvers.mapEntryResolver());
+        // foo.string.raw returns a RawString which is never escaped
+        builder.addValueResolver(ValueResolvers.rawResolver());
+
+        // Escape some characters for HTML templates
+        Escaper htmlEscaper = Escaper.builder().add('"', "&quot;").add('\'', "&#39;")
+                .add('&', "&amp;").add('<', "&lt;").add('>', "&gt;").build();
+        builder.addResultMapper(new ResultMapper() {
```
FroMage: Are we sure it's the last mapper? I don't think users should be able to register mappers _after_ this one, no?

extensions/qute/runtime/src/main/java/io/quarkus/qute/runtime/VariantTemplateProducer.java
```diff
@@ -166,15 +172,15 @@ public TemplateVariants(Map<Variant, String> variants, String defaultTemplate) {
...
+            return Variant.TEXT_HTML;
         } else if (suffix.equalsIgnoreCase(".xml")) {
-            return "text/xml";
+            return Variant.TEXT_XML;
         } else if (suffix.equalsIgnoreCase(".txt")) {
-            return "text/plain";
+            return Variant.TEXT_PLAIN;
         } else if (suffix.equalsIgnoreCase(".json")) {
-            return "application/json";
+            return Variant.APPLICATION_JSON;
```
FroMage: We support JSON but I didn't see any escaper for JSON.

independent-projects/qute/core/src/main/java/io/quarkus/qute/Variant.java
```diff
@@ -1,12 +1,17 @@
...
 
 /**
  * Media type, locale and encoding.
  */
-public class Variant {
+public final class Variant {
+
+    public final static String TEXT_HTML = "text/html";
+    public final static String TEXT_PLAIN = "text/plain";
+    public final static String TEXT_XML = "text/xml";
```
FroMage: Note that XML can be `application/xml` in many cases, but I don't think it matters here. https://www.ietf.org/rfc/rfc2376.txt

FroMage (approved):
LGTM

extensions/qute/runtime/src/main/java/io/quarkus/qute/runtime/EngineProducer.java
mkouba: Mappers with higher priority are tried first. So if a user wants to map some object then the priority should be > `io.quarkus.qute.WithPriority.DEFAULT_PRIORITY`.

extensions/qute/runtime/src/main/java/io/quarkus/qute/runtime/VariantTemplateProducer.java
mkouba: Well, actually we don't support JSON ATM ;-). TBD

independent-projects/qute/core/src/main/java/io/quarkus/qute/Variant.java
mkouba: Yes, I know. But in our case we guess the content type from a file suffix and set always `text/xml` for `*.xml` templates.

extensions/qute/runtime/src/main/java/io/quarkus/qute/runtime/EngineProducer.java
mkouba: +1 for proper SPI... in the furure ;-).

independent-projects/qute/core/src/test/java/io/quarkus/qute/EscaperTest.java
```diff
@@ -0,0 +1,48 @@
...
+        assertEquals("b", escaper.escape("b"));
+
+        Escaper html = Escaper.builder().add('"', "&quot;").add('\'', "&#39;")
+                .add('&', "&amp;").add('<', "&lt;").add('>', "&gt;").build();
+        assertEquals("&lt;strong&gt;Čolek&lt;/strong&gt;", html.escape("<strong>Čolek</strong>"));
+        assertEquals("&lt;a&gt;&amp;link&quot;&#39;&lt;/a&gt;", html.escape("<a>&link\"'</a>"));
+    }
+
+    @Test
+    public void testRawStringRevolver() {
```
gastaldi:
```suggestion
    public void testRawStringResolver() {
```

independent-projects/qute/core/src/main/java/io/quarkus/qute/TemplateLocator.java
```diff
@@ -0,0 +1,37 @@
...
+         * 
+         * @return
+         */
+        Reader read();
+
+        /**
+         * 
+         * @return the template variant
+         */
+        Optional<Variant> getVariant();
```
gastaldi: What about multiple localized templates for a given template? Should this return a List?

gsmet (approved):

gsmet:
Let's get moving, we can adjust later but we need at least some escaping in 1.1.

independent-projects/qute/core/src/main/java/io/quarkus/qute/TemplateLocator.java
mkouba: I don't think so. This represents a single template file.

independent-projects/qute/core/src/test/java/io/quarkus/qute/EscaperTest.java
mkouba: I may be blind but what's the suggested change?

gastaldi: Revolver != resolver :smile:

mkouba: Ok, it's confirmed. I'm blind...

FroMage: Most beautiful typo I've seen so far :)

mkouba: I'm very good at typos!

PR #6201, opened 2019-12-16
Use the bean index to scan subresources, check parameters and look for @Context

gsmet:
In the case of generated resource beans such as in Kogito, this was
causing failures.

It should fix the quickstart issue we have right now.

/cc @mswiderski


mswiderski (approved):

geoand (approved):
LGTM

PR #6204, opened 2019-12-16
Upgrade to Hibernate Search 6.0.0 beta3

yrodiere:
https://in.relation.to/2019/12/16/hibernate-search-6-0-0-Beta3/

@gsmet I will need your advice, especially on the reflection stuff. I'm not sure I fully grasped what was going on there, so I may have removed important code.

To summarize, starting with Beta3 all Hibernate Search mapping annotations are meta-annotated, so we do not need to list them all. Custom annotations must also be meta-annotated with the same meta-annotation, so they are no longer a specific use case.

However, it seems we were behaving differently for built-in annotations and custom bridge annotations (the code wasn't exactly the same). I did as well as I could, but that may not be enough...

### DIFF
--- a/extensions/hibernate-search-elasticsearch/runtime/src/main/java/io/quarkus/hibernate/search/elasticsearch/runtime/HibernateSearchElasticsearchRecorder.java
+++ b/extensions/hibernate-search-elasticsearch/runtime/src/main/java/io/quarkus/hibernate/search/elasticsearch/runtime/HibernateSearchElasticsearchRecorder.java
@@ -124,6 +124,8 @@ public class HibernateSearchElasticsearchRecorder {
                 ElasticsearchBackendRuntimeConfig elasticsearchBackendConfig) {
             addBackendConfig(propertyCollector, backendName, ElasticsearchBackendSettings.HOSTS,
                     elasticsearchBackendConfig.hosts);
+            addBackendConfig(propertyCollector, backendName, ElasticsearchBackendSettings.PROTOCOL,
+                    elasticsearchBackendConfig.protocol.getHibernateSearchString());
             addBackendConfig(propertyCollector, backendName, ElasticsearchBackendSettings.USERNAME,
                     elasticsearchBackendConfig.username);
             addBackendConfig(propertyCollector, backendName, ElasticsearchBackendSettings.PASSWORD,
@@ -140,8 +142,6 @@ public class HibernateSearchElasticsearchRecorder {
             if (elasticsearchBackendConfig.discovery.enabled) {
                 addBackendConfig(propertyCollector, backendName, ElasticsearchBackendSettings.DISCOVERY_REFRESH_INTERVAL,
                         elasticsearchBackendConfig.discovery.refreshInterval.getSeconds());
-                addBackendConfig(propertyCollector, backendName, ElasticsearchBackendSettings.DISCOVERY_SCHEME,
-                        elasticsearchBackendConfig.discovery.defaultScheme);
             }
 
             addBackendDefaultIndexConfig(propertyCollector, backendName, ElasticsearchIndexSettings.LIFECYCLE_STRATEGY,
--- a/bom/runtime/pom.xml
+++ b/bom/runtime/pom.xml
@@ -83,12 +83,12 @@
         <classmate.version>1.3.4</classmate.version>
         <hibernate-validator.version>6.1.0.Final</hibernate-validator.version>
         <hibernate-orm.version>5.4.10.Final</hibernate-orm.version>
-        <hibernate-search.version>6.0.0.Beta2</hibernate-search.version>
+        <hibernate-search.version>6.0.0.Beta3</hibernate-search.version>
         <narayana.version>5.10.0.Final</narayana.version>
         <jboss-transaction-api_1.2_spec.version>1.1.1.Final</jboss-transaction-api_1.2_spec.version>
         <agroal.version>1.7</agroal.version>
         <jboss-transaction-spi.version>7.6.0.Final</jboss-transaction-spi.version>
-        <elasticsearch-rest-client.version>7.4.0</elasticsearch-rest-client.version>
+        <elasticsearch-rest-client.version>7.5.0</elasticsearch-rest-client.version>
         <rxjava1.version>1.3.8</rxjava1.version>
         <rxjava.version>2.2.15</rxjava.version>
         <wildfly.openssl.version>1.0.6.Final</wildfly.openssl.version>
--- a/build-parent/pom.xml
+++ b/build-parent/pom.xml
@@ -16,8 +16,8 @@
 
     <properties>
         <!-- Maven plugin versions -->
-        <elasticsearch-maven-plugin.version>6.14</elasticsearch-maven-plugin.version>
-        <elasticsearch-server.version>7.4.0</elasticsearch-server.version>
+        <elasticsearch-maven-plugin.version>6.15</elasticsearch-maven-plugin.version>
+        <elasticsearch-server.version>7.5.0</elasticsearch-server.version>
         <scala-maven-plugin.version>4.1.1</scala-maven-plugin.version>
 
         <!-- These properties are needed in order for them to be resolvable by the generated projects -->
--- a/docs/src/main/asciidoc/hibernate-search-elasticsearch.adoc
+++ b/docs/src/main/asciidoc/hibernate-search-elasticsearch.adoc
@@ -594,7 +594,7 @@ Let's use Docker to start one of each:
 
 [source, shell]
 ----
-docker run -it --rm=true --name elasticsearch_quarkus_test -p 9200:9200 -p 9300:9300 -e "discovery.type=single-node" docker.elastic.co/elasticsearch/elasticsearch:7.4.0
+docker run -it --rm=true --name elasticsearch_quarkus_test -p 9200:9200 -p 9300:9300 -e "discovery.type=single-node" docker.elastic.co/elasticsearch/elasticsearch:7.5.0
 ----
 
 [source, shell]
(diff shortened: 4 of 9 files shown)


gsmet:
CI is unhappy: the native build is failing with some weird logging issue:
```
	at io.quarkus.runner.ApplicationImpl.doStart(ApplicationImpl.zig:231)
	at io.quarkus.runtime.Application.start(Application.java:87)
	at io.quarkus.runtime.Application.run(Application.java:210)
Caused by: javax.persistence.PersistenceException: [PersistenceUnit: default] Unable to build Hibernate SessionFactory
	at io.quarkus.hibernate.orm.runtime.boot.FastBootEntityManagerFactoryBuilder.persistenceException(FastBootEntityManagerFactoryBuilder.java:113)
	at io.quarkus.hibernate.orm.runtime.boot.FastBootEntityManagerFactoryBuilder.build(FastBootEntityManagerFactoryBuilder.java:67)
	at io.quarkus.hibernate.orm.runtime.FastBootHibernatePersistenceProvider.createEntityManagerFactory(FastBootHibernatePersistenceProvider.java:54)
	at javax.persistence.Persistence.createEntityManagerFactory(Persistence.java:80)
	at javax.persistence.Persistence.createEntityManagerFactory(Persistence.java:55)
	at io.quarkus.hibernate.orm.runtime.JPAConfig$LazyPersistenceUnit.get(JPAConfig.java:109)
	at io.quarkus.hibernate.orm.runtime.JPAConfig.startAll(JPAConfig.java:57)
	at io.quarkus.hibernate.orm.runtime.HibernateOrmRecorder.startAllPersistenceUnits(HibernateOrmRecorder.java:82)
	at io.quarkus.deployment.steps.HibernateOrmProcessor$startPersistenceUnits59.deploy_0(HibernateOrmProcessor$startPersistenceUnits59.zig:51)
	at io.quarkus.deployment.steps.HibernateOrmProcessor$startPersistenceUnits59.deploy(HibernateOrmProcessor$startPersistenceUnits59.zig:70)
	at io.quarkus.runner.ApplicationImpl.doStart(ApplicationImpl.zig:161)
	... 3 more
Caused by: java.lang.IllegalArgumentException: Invalid logger interface org.hibernate.search.engine.logging.impl.Log (implementation not found in java.lang.ClassLoader@3477f70)
	at org.jboss.logging.Logger$1.run(Logger.java:2556)
	at java.security.AccessController.doPrivileged(AccessController.java:80)
	at org.jboss.logging.Logger.getMessageLogger(Logger.java:2529)
	at org.jboss.logging.Logger.getMessageLogger(Logger.java:2516)
	at org.hibernate.search.util.common.logging.impl.LoggerFactory.make(LoggerFactory.java:21)
	at org.hibernate.search.engine.Version.logVersion(Version.java:32)
	at org.hibernate.search.mapper.orm.bootstrap.impl.HibernateSearchIntegrator.integrate(HibernateSearchIntegrator.java:65)
	at org.hibernate.internal.SessionFactoryImpl.<init>(SessionFactoryImpl.java:286)
	at org.hibernate.boot.internal.SessionFactoryBuilderImpl.build(SessionFactoryBuilderImpl.java:462)
	at io.quarkus.hibernate.orm.runtime.boot.FastBootEntityManagerFactoryBuilder.build(FastBootEntityManagerFactoryBuilder.java:65)
	... 12 more
```
Did you change something regarding logging?

yrodiere:
@gsmet This part of the stack is new in Beta3:

```
	at org.hibernate.search.engine.Version.logVersion(Version.java:32)
```

However, I do not understand why it's failing. I would have expected reflection to be enabled for all JBoss loggers as part of the jboss-logger Quarkus extension.

Do I have to make sure all loggers are instantiated during deployment? If so, we may have a problem, because most loggers are instantiated when a class is first loaded, and I'm pretty sure we have several other classes using a logger that are not used until runtime (`SimpleIndexingProgressMonitor` being the most obvious example I have in mind).

gsmet:
@dmlloyd I think we need your advice here.

dmlloyd:
We do not enable reflection in any logging classes that I am aware of.  This kind of problem happens when loggers are not created in static initializers, which should always be the case.

We did a fairly significant cleanup effort around this a while ago for performance reasons.  I think if Hibernate Search has started using non-constant loggers again, I would consider it a regression in that project for that reason alone.  Also: does Hibernate Search *really* need its own logging abstraction?

That said, we do need a good substitution strategy for loggers which support i18n, because I believe right now it will pin the locale of the building process rather than the run time environment.

yrodiere:
> We did a fairly significant cleanup effort around this a while ago for performance reasons. I think if Hibernate Search has started using non-constant loggers again, I would consider it a regression in that project for that reason alone.

Looking closer, there is indeed a behavior change: the logger used to be created (then dropped, to be garbage collected) in a static intiailizer, and is now initialized at runtime (then dropped).

Not sure it's a performance regression, though, since the logger is only instantiated once (just as before).

> Also: does Hibernate Search really need its own logging abstraction?

As a rule, no. `Version.logVersion` is an exception. I guess I can get rid of it.

yrodiere:
Created https://hibernate.atlassian.net/browse/HSEARCH-3795 to address this. I'll see what I can do in this PR to make it work with `6.0.0.Beta3`.

Thanks for the insight, @dmlloyd .

extensions/hibernate-search-elasticsearch/deployment/src/main/java/io/quarkus/hibernate/search/elasticsearch/HibernateSearchLogFilterBuildStep.java
```diff
@@ -0,0 +1,16 @@
...
+import io.quarkus.deployment.logging.LogCleanupFilterBuildItem;
+
+/**
+ * Processor that sets up log filters for Hibernate Search
+ */
+public final class HibernateSearchLogFilterBuildStep {
+
+    @BuildStep
+    void setupLogFilters(BuildProducer<LogCleanupFilterBuildItem> filters) {
+        filters.produce(new LogCleanupFilterBuildItem("org.hibernate.search.engine.Version", "HSEARCH000034"));
```
gsmet: I wouldn't make it a separate class. We tend to centralize these things in the processor.

extensions/hibernate-search-elasticsearch/runtime/src/main/java/io/quarkus/hibernate/search/elasticsearch/runtime/HibernateSearchElasticsearchRecorder.java
```diff
@@ -124,6 +124,8 @@ private void contributeBackendRuntimeProperties(BiConsumer<String, Object> prope
                 ElasticsearchBackendRuntimeConfig elasticsearchBackendConfig) {
             addBackendConfig(propertyCollector, backendName, ElasticsearchBackendSettings.HOSTS,
                     elasticsearchBackendConfig.hosts);
+            addBackendConfig(propertyCollector, backendName, ElasticsearchBackendSettings.PROTOCOL,
+                    elasticsearchBackendConfig.protocol.getHibernateSearchString());
```
gsmet: Can you add a note here: https://github.com/quarkusio/quarkus/wiki/Migration-Guide ?

gsmet (changes requested):
Added two minor comments.

The reflection changes looks good. I was maybe a bit too conservative before but with meta annotations, you don't have a choice.

extensions/hibernate-search-elasticsearch/runtime/src/main/java/io/quarkus/hibernate/search/elasticsearch/runtime/HibernateSearchElasticsearchRecorder.java
yrodiere: As a matter of fact, I cannot. I don't have the sufficient rights.

Here's something, if you want to add it yourself:

````
### Hibernate Search + Elasticsearch (Preview)

The protocol used to contact Elasticsearch is now specified in a separate configuration property, `quarkus.hibernate-search.elasticsearch.protocol`. This property defaults to `http`. 

As a result, the property `quarkus.hibernate-search.elasticsearch.discovery.default_scheme` was removed.

So if your configuration looked like this:

```
quarkus.hibernate-search.elasticsearch.hosts = http://es1.mycompany.com
```

You should now have this:

```
quarkus.hibernate-search.elasticsearch.hosts = es1.mycompany.com
```

And if it looked like this:

```
quarkus.hibernate-search.elasticsearch.hosts = https://es1.mycompany.com
quarkus.hibernate-search.elasticsearch.discovery.enabled = true
quarkus.hibernate-search.elasticsearch.discovery.default_scheme = https
```

You should now have this:

```
quarkus.hibernate-search.elasticsearch.hosts = es1.mycompany.com
quarkus.hibernate-search.elasticsearch.protocol = https
quarkus.hibernate-search.elasticsearch.discovery.enabled = true
```
````

extensions/hibernate-search-elasticsearch/deployment/src/main/java/io/quarkus/hibernate/search/elasticsearch/HibernateSearchLogFilterBuildStep.java
yrodiere: Ok, I just copy-pasted from someplace else. Will do.

yrodiere:
@gsmet I addressed your comments.
I also added some assertions in the mass indexing test following our discussion (where I was worried that maybe, mass indexing wasn't effective -- spoiler alert: it works).

extensions/hibernate-search-elasticsearch/runtime/src/main/java/io/quarkus/hibernate/search/elasticsearch/runtime/HibernateSearchElasticsearchRecorder.java
gsmet: Added, thanks.

gsmet (approved):

gsmet:
Merged, thanks!

PR #6206, opened 2019-12-16
updated binary type check

evanchooly:
removed dependence on amazon serverless libraries internals
added tests

### DIFF
--- a/bom/runtime/pom.xml
+++ b/bom/runtime/pom.xml
@@ -135,7 +135,6 @@
         <tika.version>1.22</tika.version>
         <aws-lambda-java.version>1.1.0</aws-lambda-java.version>
         <aws-lambda-java-events.version>2.2.5</aws-lambda-java-events.version>
-        <aws-lambda-serverless-java-container.version>1.3.1</aws-lambda-serverless-java-container.version>
         <awssdk.version>2.10.3</awssdk.version>
         <azure-functions-java-library.version>1.3.0</azure-functions-java-library.version>
         <kotlin.version>1.3.41</kotlin.version>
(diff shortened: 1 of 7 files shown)


bom/runtime/pom.xml
```diff
@@ -2315,25 +2315,6 @@
                 <artifactId>aws-lambda-java-events</artifactId>
                 <version>${aws-lambda-java-events.version}</version>
             </dependency>
-            <dependency>
-                <groupId>com.amazonaws.serverless</groupId>
-                <artifactId>aws-serverless-java-container-core</artifactId>
-                <version>${aws-lambda-serverless-java-container.version}</version>
```
gastaldi: Is the `aws-lambda-serverless-java-container.version` property still used?

extensions/amazon-lambda-http/runtime/src/main/java/io/quarkus/amazon/lambda/http/LambdaHttpHandler.java
```diff
@@ -165,12 +164,15 @@ private boolean isBinary(String contentType) {
...
             } else {
-                return LambdaContainerHandler.getContainerConfig().isBinaryContentType(contentType);
+                return isBinaryContentType(contentType);
             }
         }
         return false;
     }
 
+    private boolean isBinaryContentType(String contentType) {
+        return contentType != null && !contentType.startsWith("text/");
```
gastaldi: Should that test against `application/` types too?

evanchooly:
Bill said json should get base64'd and that lumped gets lumped in since
it's application/json.  Im not 100% on this check but its currently the
simplest thing that could possibly work.  I won't be surprised or offended
if it needs refinement.

bom/runtime/pom.xml
evanchooly: no, it isn't.  I'll remove it.

extensions/amazon-lambda-http/runtime/src/main/java/io/quarkus/amazon/lambda/http/LambdaHttpHandler.java
Ravisankar-Challa: Please provide an option to read Content-Types from quarkus application properties so that application developers can decide.

gsmet:
This needs a rebase but I'm not sure what we want to do here.

@oztimpower I think your expertise could be valuable here in defining precisely what we should do and how things should be configured.

oztimpower:
Ok, thanks for looping me in.  I will validate inside Amazon.

evanchooly:
old and outdated

PR #6207, opened 2019-12-16
Transfer JAX-B annotations from fields to accessor

### LINKED ISSUE #6021: JAXB annotations on a Panache entity causes REST call to fail
I want to include JAXB annotations on my Panache entity so that I can return the entity in XML format in a REST response.
Adding `@XmlAttribute` on the public property of the entity causes the REST call to fail with an exception.
It is possible to change the access modifier of the property to private to prevent the exception however this approach then loses one of the benefits of Panache.

```
2019-12-08 14:33:28,640 ERROR [org.jbo.res.res.i18n] (vert.x-worker-thread-1) RESTEASY002005: Failed executing GET /person: org.jboss.resteasy.plugins.providers.jaxb.JAXBMarshalException: com.sun.xml.bind.v2.runtime.IllegalAnnotationsException: 1 counts of IllegalAnnotationExceptions
org.acme.quarkus.sample.Person#name has mutually exclusive annotations @javax.xml.bind.annotation.XmlTransient and @javax.xml.bind.annotation.XmlAttribute
	this problem is related to the following location:
		at @javax.xml.bind.annotation.XmlTransient()
		at org.acme.quarkus.sample.Person
	this problem is related to the following location:
		at @javax.xml.bind.annotation.XmlAttribute(name="Name", namespace="##default", required=false)
		at org.acme.quarkus.sample.Person

	at org.jboss.resteasy.plugins.providers.jaxb.AbstractJAXBProvider.getMarshaller(AbstractJAXBProvider.java:187)
	at org.jboss.resteasy.plugins.providers.jaxb.AbstractJAXBProvider.writeTo(AbstractJAXBProvider.java:149)
	at org.jboss.resteasy.core.interception.jaxrs.AbstractWriterInterceptorContext.writeTo(AbstractWriterInterceptorContext.java:193)
	at org.jboss.resteasy.core.interception.jaxrs.ServerWriterInterceptorContext.writeTo(ServerWriterInterceptorContext.java:64)
	at org.jboss.resteasy.core.interception.jaxrs.AbstractWriterInterceptorContext.proceed(AbstractWriterInterceptorContext.java:155)
	at org.jboss.resteasy.core.ServerResponseWriter.lambda$writeNomapResponse$2(ServerResponseWriter.java:156)
	at org.jboss.resteasy.core.interception.jaxrs.ContainerResponseContextImpl.filter(ContainerResponseContextImpl.java:404)
	at org.jboss.resteasy.core.ServerResponseWriter.executeFilters(ServerResponseWriter.java:232)
	at org.jboss.resteasy.core.ServerResponseWriter.writeNomapResponse(ServerResponseWriter.java:97)
	at org.jboss.resteasy.core.ServerResponseWriter.writeNomapResponse(ServerResponseWriter.java:70)
	at org.jboss.resteasy.core.SynchronousDispatcher.writeResponse(SynchronousDispatcher.java:578)
	at org.jboss.resteasy.core.SynchronousDispatcher.invoke(SynchronousDispatcher.java:508)
	at org.jboss.resteasy.core.SynchronousDispatcher.lambda$invoke$4(SynchronousDispatcher.java:252)
	at org.jboss.resteasy.core.SynchronousDispatcher.lambda$preprocess$0(SynchronousDispatcher.java:153)
	at org.jboss.resteasy.core.interception.jaxrs.PreMatchContainerRequestContext.filter(PreMatchContainerRequestContext.java:363)
	at org.jboss.resteasy.core.SynchronousDispatcher.preprocess(SynchronousDispatcher.java:156)
	at org.jboss.resteasy.core.SynchronousDispatcher.invoke(SynchronousDispatcher.java:238)
	at io.quarkus.resteasy.runtime.standalone.RequestDispatcher.service(RequestDispatcher.java:73)
	at io.quarkus.resteasy.runtime.standalone.VertxRequestHandler.dispatch(VertxRequestHandler.java:118)
	at io.quarkus.resteasy.runtime.standalone.VertxRequestHandler.lambda$handle$0(VertxRequestHandler.java:74)
	at io.vertx.core.impl.ContextImpl.lambda$executeBlocking$2(ContextImpl.java:316)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30)
	at java.base/java.lang.Thread.run(Thread.java:834)
Caused by: com.sun.xml.bind.v2.runtime.IllegalAnnotationsException: 1 counts of IllegalAnnotationExceptions
org.acme.quarkus.sample.Person#name has mutually exclusive annotations @javax.xml.bind.annotation.XmlTransient and @javax.xml.bind.annotation.XmlAttribute
	this problem is related to the following location:
		at @javax.xml.bind.annotation.XmlTransient()
		at org.acme.quarkus.sample.Person
	this problem is related to the following location:
		at @javax.xml.bind.annotation.XmlAttribute(name="Name", namespace="##default", required=false)
		at org.acme.quarkus.sample.Person

	at com.sun.xml.bind.v2.runtime.IllegalAnnotationsException$Builder.check(IllegalAnnotationsException.java:76)
	at com.sun.xml.bind.v2.runtime.JAXBContextImpl.getTypeInfoSet(JAXBContextImpl.java:441)
	at com.sun.xml.bind.v2.runtime.JAXBContextImpl.<init>(JAXBContextImpl.java:273)
	at com.sun.xml.bind.v2.runtime.JAXBContextImpl.<init>(JAXBContextImpl.java:109)
	at com.sun.xml.bind.v2.runtime.JAXBContextImpl$JAXBContextBuilder.build(JAXBContextImpl.java:1126)
	at com.sun.xml.bind.v2.ContextFactory.createContext(ContextFactory.java:135)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at javax.xml.bind.ContextFinder.newInstance(ContextFinder.java:275)
	at javax.xml.bind.ContextFinder.newInstance(ContextFinder.java:264)
	at javax.xml.bind.ContextFinder.find(ContextFinder.java:403)
	at javax.xml.bind.JAXBContext.newInstance(JAXBContext.java:691)
	at org.jboss.resteasy.plugins.providers.jaxb.JAXBContextWrapper.<init>(JAXBContextWrapper.java:112)
	at org.jboss.resteasy.plugins.providers.jaxb.JAXBContextWrapper.<init>(JAXBContextWrapper.java:175)
	at org.jboss.resteasy.plugins.providers.jaxb.XmlJAXBContextFinder.createContextObject(XmlJAXBContextFinder.java:52)
	at org.jboss.resteasy.plugins.providers.jaxb.AbstractJAXBContextFinder.createContext(AbstractJAXBContextFinder.java:173)
	at org.jboss.resteasy.plugins.providers.jaxb.XmlJAXBContextFinder.findCachedContext(XmlJAXBContextFinder.java:41)
	at org.jboss.resteasy.plugins.providers.jaxb.AbstractJAXBProvider.findJAXBContext(AbstractJAXBProvider.java:83)
	at org.jboss.resteasy.plugins.providers.jaxb.AbstractJAXBProvider.getMarshaller(AbstractJAXBProvider.java:173)
	... 24 more
```

**To Reproduce**
1. Create starter project.
2. Add extensions: hibernate-orm-panache, agroal, resteasy-jaxb, and a JDBC driver extension
3. Add datasource configuration to appliction.properties
4. Create a Panache entity that includes a property with a JAXB annotation:

```
@Entity
@XmlRootElement(name = "Person")
@XmlAccessorType(XmlAccessType.NONE)
public class Person extends PanacheEntity {

	@XmlAttribute(name = "Name")
	public String name;
}
```

5. Create a REST resource to return the entity in XML:

```
@Path("/person")
@Produces(MediaType.APPLICATION_XML)
public class PersonResource {

	@GET
	public Person find() {
		return Person.findAll().firstResult();
	}	
}
```

6. Ensure that the database will be populated with a Person record (e.g. use import-dev.sql)
7. Start the application and go to http://localhost:8080/person

**Environment:**
- Quarkus 1.0.1.Final
- openjdk 11
- maven 3.6.2

aguibert:
For Panache entity enhancement we add @XmlTransient to the original fields
to force Jackson to use the generated accessor methods. However, this can
cause conflicts if the field was originally annotated with JAX-B annotations
that are mutually exclusive with @XmlTransient.

Signed-off-by: Andrew Guibert <andy.guibert@gmail.com>

Fixes https://github.com/quarkusio/quarkus/issues/6021

### DIFF
--- a/extensions/panache/panache-common/deployment/src/main/java/io/quarkus/panache/common/deployment/PanacheEntityEnhancer.java
+++ b/extensions/panache/panache-common/deployment/src/main/java/io/quarkus/panache/common/deployment/PanacheEntityEnhancer.java
@@ -84,17 +84,12 @@ public abstract class PanacheEntityEnhancer<MetamodelType extends MetamodelInfo<
                     descriptors.add(descriptor);
                     if (!descriptor.startsWith(JAXB_ANNOTATION_PREFIX)) {
                         return super.visitAnnotation(descriptor, visible);
+                    } else {
+                        // Save off JAX-B annotations on the field so they can be applied to the generated getter later
+                        EntityFieldAnnotation efAnno = new EntityFieldAnnotation(descriptor);
+                        ef.annotations.add(efAnno);
+                        return new PanacheMovingAnnotationVisitor(efAnno);
                     }
-                    // Save off JAX-B annotations on the field so they can be applied to the generated getter later
-                    EntityFieldAnnotation efAnno = new EntityFieldAnnotation(descriptor);
-                    ef.annotations.add(efAnno);
-                    return new AnnotationVisitor(Opcodes.ASM7) {
-                        @Override
-                        public void visit(String name, Object value) {
-                            efAnno.name = name;
-                            efAnno.value = value;
-                        }
-                    };
                 }
 
                 @Override
@@ -200,10 +195,7 @@ public abstract class PanacheEntityEnhancer<MetamodelType extends MetamodelInfo<
                     mv.visitMaxs(0, 0);
                     // Apply JAX-B annotations that are being transferred from the field
                     for (EntityFieldAnnotation anno : field.annotations) {
-                        AnnotationVisitor av = mv.visitAnnotation(anno.descriptor, true);
-                        if (anno.name != null)
-                            av.visit(anno.name, anno.value);
-                        av.visitEnd();
+                        anno.writeToVisitor(mv);
                     }
                     mv.visitEnd();
                 }
--- a/extensions/panache/panache-common/deployment/src/main/java/io/quarkus/panache/common/deployment/PanacheEntityEnhancer.java
+++ b/extensions/panache/panache-common/deployment/src/main/java/io/quarkus/panache/common/deployment/PanacheEntityEnhancer.java
@@ -21,6 +21,7 @@ import org.objectweb.asm.Type;
 
 import io.quarkus.panache.common.Parameters;
 import io.quarkus.panache.common.Sort;
+import io.quarkus.panache.common.deployment.EntityField.EntityFieldAnnotation;
 
 public abstract class PanacheEntityEnhancer<MetamodelType extends MetamodelInfo<?>>
         implements BiFunction<String, ClassVisitor, ClassVisitor> {
@@ -33,6 +34,7 @@ public abstract class PanacheEntityEnhancer<MetamodelType extends MetamodelInfo<
     public final static String PARAMETERS_BINARY_NAME = PARAMETERS_NAME.replace('.', '/');
     public final static String PARAMETERS_SIGNATURE = "L" + PARAMETERS_BINARY_NAME + ";";
 
+    private static final String JAXB_ANNOTATION_PREFIX = "Ljavax/xml/bind/annotation/";
     private static final String JAXB_TRANSIENT_BINARY_NAME = "javax/xml/bind/annotation/XmlTransient";
     private static final String JAXB_TRANSIENT_SIGNATURE = "L" + JAXB_TRANSIENT_BINARY_NAME + ";";
 
@@ -80,15 +82,26 @@ public abstract class PanacheEntityEnhancer<MetamodelType extends MetamodelInfo<
                 @Override
                 public AnnotationVisitor visitAnnotation(String descriptor, boolean visible) {
                     descriptors.add(descriptor);
-                    return super.visitAnnotation(descriptor, visible);
+                    if (!descriptor.startsWith(JAXB_ANNOTATION_PREFIX)) {
+                        return super.visitAnnotation(descriptor, visible);
+                    }
+                    // Save off JAX-B annotations on the field so they can be applied to the generated getter later
+                    EntityFieldAnnotation efAnno = new EntityFieldAnnotation(descriptor);
+                    ef.annotations.add(efAnno);
+                    return new AnnotationVisitor(Opcodes.ASM7) {
+                        @Override
+                        public void visit(String name, Object value) {
+                            efAnno.name = name;
+                            efAnno.value = value;
+                        }
+                    };
                 }
 
                 @Override
                 public void visitEnd() {
                     // add the @JaxbTransient property to the field so that Jackson prefers the generated getter
                     // jsonb will already use the getter so we're good
-                    if (!descriptors.contains(JAXB_TRANSIENT_SIGNATURE))
-                        super.visitAnnotation(JAXB_TRANSIENT_SIGNATURE, true);
+                    super.visitAnnotation(JAXB_TRANSIENT_SIGNATURE, true);
                     super.visitEnd();
                 }
             };
@@ -185,6 +198,13 @@ public abstract class PanacheEntityEnhancer<MetamodelType extends MetamodelInfo<
                     int returnCode = JandexUtil.getReturnInstruction(field.descriptor);
                     mv.visitInsn(returnCode);
                     mv.visitMaxs(0, 0);
+                    // Apply JAX-B annotations that are being transferred from the field
+                    for (EntityFieldAnnotation anno : field.annotations) {
+                        AnnotationVisitor av = mv.visitAnnotation(anno.descriptor, true);
+                        if (anno.name != null)
+                            av.visit(anno.name, anno.value);
+                        av.visitEnd();
+                    }
                     mv.visitEnd();
                 }
 
--- a/extensions/panache/panache-common/deployment/src/main/java/io/quarkus/panache/common/deployment/EntityField.java
+++ b/extensions/panache/panache-common/deployment/src/main/java/io/quarkus/panache/common/deployment/EntityField.java
@@ -1,5 +1,8 @@
 package io.quarkus.panache.common.deployment;
 
+import java.util.HashSet;
+import java.util.Set;
+
 import io.quarkus.deployment.bean.JavaBeanUtil;
 
 public class EntityField {
@@ -7,6 +10,7 @@ public class EntityField {
     public final String name;
     public final String descriptor;
     public String signature;
+    public final Set<EntityFieldAnnotation> annotations = new HashSet<>(2);
 
     public EntityField(String name, String descriptor) {
         this.name = name;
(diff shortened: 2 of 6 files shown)


aguibert:
@FroMage can you please review?

extensions/panache/panache-common/deployment/src/main/java/io/quarkus/panache/common/deployment/PanacheEntityEnhancer.java
```diff
@@ -80,15 +82,26 @@ public FieldVisitor visitField(int access, String name, String descriptor, Strin
...
                 public AnnotationVisitor visitAnnotation(String descriptor, boolean visible) {
                     descriptors.add(descriptor);
-                    return super.visitAnnotation(descriptor, visible);
+                    if (!descriptor.startsWith(JAXB_ANNOTATION_PREFIX)) {
+                        return super.visitAnnotation(descriptor, visible);
+                    }
+                    // Save off JAX-B annotations on the field so they can be applied to the generated getter later
+                    EntityFieldAnnotation efAnno = new EntityFieldAnnotation(descriptor);
+                    ef.annotations.add(efAnno);
+                    return new AnnotationVisitor(Opcodes.ASM7) {
```
FroMage: I forgot what the ASM API does with annotation visitors wrt nested annotations. Does the jaxb module have nested annotations? Are we sure that this won't be called for nested annotation attributes, or fail to properly forward those calls?

aguibert: you're correct, there are separate visit method to override for:
- nested annotation --> visitAnnotation
- enums --> visitEnum
- arrays --> visitArray (javadoc indicates it calls `visit()` for each array element)

I had a look through  the annotations in the `javax.xml.bind.annotation` package, and I see there are no annotations with enum or nested annotation attributes, but I do see there is a `XmlElements` and `XmlElementRefs` container annotations which take arrays of annotations. Looking back at the code I don't think these cases will be handled properly, so I'll add a test for it and ensure they work properly.

aguibert: @FroMage I pushed another commit which properly handles the `@XmlElements` and `@XmlElementRefs` scenario. Ready for re-review

FroMage: If you've checked the JAXB API and found only a single annotation whose value is an array of nested annotations, I'm torn between saying this is Good Enough™ or saying this has to be more general because if they ever add a new attribute with an enum value or a non-"value" non-array nested annotation, this code will suddenly break.

On the other hand, "collecting all annotations that match a given predicate and moving them to another element" sounds like a sufficiently reusable functionality that it might be worth doing this comprehensively into a quarkus-core utility class that others can later reuse.

WDYT @stuartwdouglas @dmlloyd?

aguibert: I think JAXB is rather unlikely to add new annotations in the future, what I'd be more concerned about is the impending `javax.*` --> `jakarta.*` package rename and handling that case properly. However, I've left that out because I assume some sort of general solution will be used for Quarkus and it may cover this scenario too (but I doubt it)

FroMage: I do wonder if there isn't already something in ASM that lets us "capture" a visitor's "visits" and apply it somewhere else. It feels like a problem that must happen pretty often.

The other thing when it comes to Good Enough™ is that you officially becomes the maintainer of that piece of code and by tradition you have to add a `// FIXME: this only works for XXX case but I'll fix it if it ever breaks` ;)

aguibert: Yea, recording the annotation processing flow and redirecting it to a different element was my preference as well, but I could not see a way to do this because in order to redirect the annotation processing we would need the Visitor instance from the generated accessor method so we could delegate to it. However, the visitor for the generated getter doesn't get created until later on in the code flow after we've processed all of the fields. So at this point it's a dead-end chicken-and-the-egg issue because we cannot delegate to the getter method visitor until after we've scanned the field and determined we need to generate a getter in the first place.

I'm pretty confident in the code and the test coverage (as long as the scope remains moving JAX-B annotations and not other annotations), so I could certainly add a `FIXME` to put myself on the hook for maintaining it in the future.

FroMage: OK, let's just leave a day for @stuartwdouglas or @dmlloyd to confirm they don't have or know of an API that already does this annotation visitor transfer, which would render the discussion moot, and then I'll approve this PR.

Let's also ask @geoand who also has some m4d 4SM ski11z

geoand: I certainly don't, but let me read trough this and see if I have anything to suggest :)

geoand: I don't know of any ASM API off hand for what you are looking for here.  The general solution @FroMage talks about would be great, but I would assume reading through the PR that this is just fine for the time being and perhaps we can reexamine in the future if need be (which I assume we won't have to :) )?

FroMage (approved):

extensions/panache/panache-common/deployment/src/main/java/io/quarkus/panache/common/deployment/PanacheEntityEnhancer.java
FroMage: OK, let's not wait for further input, this is fine, and we can revisit later if need be.

FroMage:
Thanks a lot @aguibert

PR #6224, opened 2019-12-17
Fix a small documentation error

loicmathieu:


docs/src/main/asciidoc/mongodb-panache.adoc
```diff
@@ -312,7 +312,7 @@ MongoDB with Panache will then map it to a MongoDB native query.
 
 If your query does not start with `{`, we will consider it a PanacheQL query:
 
-- `<singleColumnName>` (and single parameter) which will expand to `{'singleColumnName': '?'}`
+- `<singlePropertyName>` (and single parameter) which will expand to `{'singleColumnName': '?'}`
```
gsmet: Shouldn't you change the other occurrence too (the one at the end of the sentence)?

gsmet (changes requested):

docs/src/main/asciidoc/mongodb-panache.adoc
loicmathieu: No, we have propety to column mapping based on `@BsonProperty`.
But maybe this needs to be more clear on the documentatin :)

This PR only fix an error inside the documentation, maybe this all mapping stuff needs to be more clearly specified ...

gsmet (approved):

PR #6225, opened 2019-12-17
ArC - activate request context for any observer notification

### LINKED ISSUE #6221: ArC - activate request context for any observer notification (if not already active)
Right now, we do activate the request context for async observers notification. It could be useful to activate the request context for any observer notification (if not already active).

mkouba:
- resolves #6221

@manovotn I've modified the `BuiltInBeansAreResolvableTest` because the observer was notified for `@Initialized(RequestScoped.class)`

FYI @FroMage @stuartwdouglas @jaikiran...

### DIFF
--- a/docs/src/main/asciidoc/cdi-reference.adoc
+++ b/docs/src/main/asciidoc/cdi-reference.adoc
@@ -251,6 +251,18 @@ class MyBeanStarter {
 
 NOTE: Quarkus users are encouraged to always prefer the `@Observes StartupEvent` to `@Initialized(ApplicationScoped.class)` as explained in the link:lifecycle[Application Initialization and Termination] guide. 
 
+=== Request Context Lifecycle 
+
+The request context is also active:
+
+* during notification of a synchronous observer method.
+
+The request context is destroyed:
+
+* after the observer notification completes for an event, if it was not already active when the notification started.
+
+NOTE: An event with qualifier `@Initialized(RequestScoped.class)` is fired when the request context is initialized for an observer notification. Moreover, the events with qualifiers `@BeforeDestroyed(RequestScoped.class)` and `@Destroyed(RequestScoped.class)` are fired when the request context is destroyed. 
+
 === Qualified Injected Fields
 
 In CDI, if you declare a field injection point you need to use `@Inject` and optionally a set of qualifiers:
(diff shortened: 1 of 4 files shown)


docs/src/main/asciidoc/cdi-reference.adoc
```diff
@@ -251,6 +251,16 @@ class MyBeanStarter {
 
 NOTE: Quarkus users are encouraged to always prefer the `@Observes StartupEvent` to `@Initialized(ApplicationScoped.class)` as explained in the link:lifecycle[Application Initialization and Termination] guide. 
 
+=== Request Context Activation
```
manovotn: `Activation` -> `Automatic Activation`? (under current title I imagine a paragraph on how I manually activate it on demand)
It might also be worth noting that this will trigger all respective req. context lifecycle event such as firing `@Initilized`.

manovotn (approved):
Looks good, added just one tiny comment.

docs/src/main/asciidoc/cdi-reference.adoc
mkouba: `Automatic Activation` sounds odd to me. Maybe we could just copy the wording from the spec: `Request Context Lifecycle`?

> It might also be worth noting that this will trigger...

Yes, we could add a note.

Sanne (commented):
I'll leave it to the CDI experts to decide (I'm not) but I feel like we shouldn't overload the meaning of a Request scope - as there's no request in this context. See my email on the mailing list as well.

manovotn:
@Sanne the spec already defines that req. context is active during some action that don't really have one such as - async observer notification or post contruct of any bean.
See this chapter - https://docs.jboss.org/cdi/spec/2.0/cdi-spec.html#request_context

mkouba:
It's more like a "virtual" request...

Sanne:
so it's like an "whatever" scope ? Funny but ok go ahead :)

Sanne:
And thanks for explaining that to me BTW, I appreciate it.

manovotn:
> so it's like an "whatever" scope ? Funny but ok go ahead :)

Nah, it's  a swiss army knife scope - it can do everything and you use it for everything :-D

mkouba:
> so it's like an "whatever" scope ? Funny but ok go ahead :)

Yes, I'd say that it's a short-lived scope that is not shared accross multiple threads and has clear boundaries.

FroMage:
> Nah, it's a swiss army knife scope - it can do everything and you use it for everything :-D

Can I also use it for conversations, then?

(just kidding).

mkouba:
> Can I also use it for conversations, then?

In fact, you could use it for _transient_ conversations as defined by the spec ;-)

PR #6235, opened 2019-12-17
Fix file system path for overriding application.properties when using dev-mode with Gradle

### LINKED ISSUE #6227: Overriding application.properties via file in build/config folder doesn't work with Gradle
**Describe the bug**
According to the [documentation](https://quarkus.io/guides/config#overriding-properties-at-runtime) it should be possible to override the default _application.properties_ file by placing a file with the same name in the _build/config_ folder (when using Gradle). This doesn't work currently.

**Expected behavior**
It should be possible to override the properties by using the mentioned approach.

**Actual behavior**
The override approach currently works only when using Maven, but not when using Gradle.

**To Reproduce**
1. Use the [code generator](https://code.quarkus.io/) to generate the default project. The build tool needs to be set to Gradle, everything else can be left unchanged.
2. Extract the package and add the following to _src/main/resources_:
`quarkus.http.port = 41234`
3. Run the project with `./gradlew quarkusDev`. Note the correct port in the console log:
`Listening on: http://0.0.0.0:41234`
4. Create the file _build/config/application.properties_ and set a custom port:
`quarkus.http.port = 44444`
5. After restarting via `./gradlew quarkusDev` the new port is not used:
`Listening on: http://0.0.0.0:41234`

**Environment:**
 - Operating system: Windows 10 Enterprise 64 bit
 - Java version: AdoptOpenJDK version "1.8.0_202"
 - Quarkus version: 1.0.1.Final

paschi:
* resolves #6227 (documentation issue only)

### DIFF
--- a/docs/src/main/asciidoc/config.adoc
+++ b/docs/src/main/asciidoc/config.adoc
@@ -396,7 +396,7 @@ _will also_ be taken into account.
 
 NOTE: Environment variables names are following the conversion rules of link:https://github.com/eclipse/microprofile-config/blob/master/spec/src/main/asciidoc/configsources.asciidoc#default-configsources[Eclipse MicroProfile]
 
-NOTE: The `config/application.properties` features is available in development mode as well. To make use of it, `config/application.properties` needs to be placed inside the build tool's output directory (`target` for Maven and `build` for Gradle).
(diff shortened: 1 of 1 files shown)


geoand (approved):
Thanks!

PR #3137, opened 2019-07-08
Build system properties support

dmlloyd:
Fixes #3012

PR #6263, opened 2019-12-19
Make application version and name available as (fixed) runtime properties

### LINKED ISSUE #6255: quarkus.application.version config property is unavailable
**Describe the bug**
The following config property injection doesn't work: 
```java
@Path("/hello")

    @ConfigProperty(name = "quarkus.application.version", defaultValue = "not-set")
    String version;

    @GET
    public String hello() {
        return "hello " + version;
    }
}
```
**Expected behavior**

When running `http localhost:8080/hello`, to return `hello 1.0.0-SNAPSHOT`

**Actual behavior**
When running `http localhost:8080/hello`, it returns `hello not-set`

**To Reproduce**
1. Create a sample project and add the endpoint above

**Configuration**
```properties

```

geoand:
Fixes: #6255

Relates to: #6267

### DIFF
--- a/core/runtime/src/main/java/io/quarkus/runtime/ApplicationConfig.java
+++ b/core/runtime/src/main/java/io/quarkus/runtime/ApplicationConfig.java
@@ -1,4 +1,4 @@
-package io.quarkus.deployment;
+package io.quarkus.runtime;
 
 import java.util.Optional;
 
@@ -6,19 +6,19 @@ import io.quarkus.runtime.annotations.ConfigItem;
 import io.quarkus.runtime.annotations.ConfigPhase;
 import io.quarkus.runtime.annotations.ConfigRoot;
 
-@ConfigRoot(phase = ConfigPhase.BUILD_TIME)
+@ConfigRoot(phase = ConfigPhase.BUILD_AND_RUN_TIME_FIXED)
 public class ApplicationConfig {
 
     /**
      * The name of the application.
-     * If not set, defaults to the name of the project.
+     * If not set, defaults to the name of the project (except for tests where it is not set at all).
      */
     @ConfigItem
     public Optional<String> name;
 
     /**
      * The version of the application.
-     * If not set, defaults to the version of the project
+     * If not set, defaults to the version of the project (except for tests where it is not set at all).
      */
     @ConfigItem
     public Optional<String> version;
--- a/devtools/maven/src/main/java/io/quarkus/maven/DevMojo.java
+++ b/devtools/maven/src/main/java/io/quarkus/maven/DevMojo.java
@@ -490,7 +490,13 @@ public class DevMojo extends AbstractMojo {
             for (Map.Entry<Object, Object> e : System.getProperties().entrySet()) {
                 devModeContext.getSystemProperties().put(e.getKey().toString(), (String) e.getValue());
             }
+
             devModeContext.getBuildSystemProperties().putAll((Map) project.getProperties());
+
+            //  this is a minor hack to allow ApplicationConfig to be populated with defaults
+            devModeContext.getBuildSystemProperties().putIfAbsent("quarkus.application.name", project.getArtifactId());
+            devModeContext.getBuildSystemProperties().putIfAbsent("quarkus.application.version", project.getVersion());
+
             devModeContext.setSourceEncoding(getSourceEncoding());
             devModeContext.setSourceJavaVersion(source);
             devModeContext.setTargetJvmVersion(target);
--- a/integration-tests/maven/src/test/java/io/quarkus/maven/it/NativeImageIT.java
+++ b/integration-tests/maven/src/test/java/io/quarkus/maven/it/NativeImageIT.java
@@ -20,9 +20,7 @@ import org.junit.jupiter.api.Test;
 import io.quarkus.maven.it.verifier.MavenProcessInvocationResult;
 import io.quarkus.maven.it.verifier.RunningInvoker;
 
-/**
- *
- */
+@EnableForNative
 public class NativeImageIT extends MojoTestBase {
 
     /**
--- a/ci-templates/stages.yml
+++ b/ci-templates/stages.yml
@@ -324,8 +324,9 @@ stages:
         parameters:
           poolSettings: ${{parameters.poolSettings}}
           expectUseVMs: ${{parameters.expectUseVMs}}
-          timeoutInMinutes: 30
+          timeoutInMinutes: 40
           modules:
+            - maven
             - jackson
             - jsonb
             - jsch
--- a/core/deployment/src/main/java/io/quarkus/deployment/steps/ApplicationInfoBuildStep.java
+++ b/core/deployment/src/main/java/io/quarkus/deployment/steps/ApplicationInfoBuildStep.java
@@ -1,8 +1,8 @@
 package io.quarkus.deployment.steps;
 
-import io.quarkus.deployment.ApplicationConfig;
 import io.quarkus.deployment.annotations.BuildStep;
 import io.quarkus.deployment.builditem.ApplicationInfoBuildItem;
+import io.quarkus.runtime.ApplicationConfig;
 
 public class ApplicationInfoBuildStep {
 
--- a/core/deployment/src/main/java/io/quarkus/runner/RuntimeRunner.java
+++ b/core/deployment/src/main/java/io/quarkus/runner/RuntimeRunner.java
@@ -97,6 +97,7 @@ public class RuntimeRunner implements Runnable, Closeable {
             builder.setRoot(target);
             builder.setClassLoader(loader);
             builder.setLaunchMode(launchMode);
+            builder.setBuildSystemProperties(buildSystemProperties);
             if (liveReloadState != null) {
                 builder.setLiveReloadState(liveReloadState);
             }
--- a/integration-tests/main/src/main/java/io/quarkus/it/config/ApplicationInfoResource.java
+++ b/integration-tests/main/src/main/java/io/quarkus/it/config/ApplicationInfoResource.java
@@ -1,0 +1,24 @@
+package io.quarkus.it.config;
+
+import javax.ws.rs.GET;
+import javax.ws.rs.Path;
+import javax.ws.rs.Produces;
+import javax.ws.rs.core.MediaType;
+
+import org.eclipse.microprofile.config.inject.ConfigProperty;
+
+@Path("/application-info")
+public class ApplicationInfoResource {
+
+    @ConfigProperty(name = "quarkus.application.version")
+    String applicationVersion;
+
+    @ConfigProperty(name = "quarkus.application.name")
+    String applicationName;
+
+    @GET
+    @Produces(MediaType.TEXT_PLAIN)
+    public String hello() {
+        return applicationName + "/" + applicationVersion;
+    }
+}
--- a/integration-tests/main/src/main/resources/application.properties
+++ b/integration-tests/main/src/main/resources/application.properties
@@ -13,6 +13,11 @@ org.eclipse.microprofile.rest.client.propagateHeaders=header-name
 # we probably won't be able to test SSL proper so it seems reasonable to set it globally
 quarkus.ssl.native=false
 
+# in tests 'quarkus.application.name' and `quarkus.application.version` do not get set from the build tool
+# here we are testing that override from properties works
+quarkus.application.name=main-integration-test
+quarkus.application.version=1.0
+
 quarkus.datasource.url: ${datasource.url}
 quarkus.datasource.driver: ${datasource.driver}
 quarkus.datasource.username: ${datasource.username}
--- a/integration-tests/main/src/test/java/io/quarkus/it/main/ApplicationInfoTestCase.java
+++ b/integration-tests/main/src/test/java/io/quarkus/it/main/ApplicationInfoTestCase.java
@@ -1,0 +1,19 @@
+package io.quarkus.it.main;
+
+import static org.hamcrest.Matchers.is;
+
+import org.junit.jupiter.api.Test;
+
+import io.quarkus.test.junit.QuarkusTest;
+import io.restassured.RestAssured;
+
+@QuarkusTest
+public class ApplicationInfoTestCase {
+
+    @Test
+    public void testConfigPropertiesProperlyInjected() {
+        RestAssured
+                .when().get("/application-info")
+                .then().body(is("main-integration-test/1.0"));
+    }
+}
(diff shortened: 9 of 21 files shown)


geoand:
I am making this a draft because it only solves part of the problem. With this change `quarkus.application.name` and `quarkus.application.version` are available when running the prod mode of the application. It doesn't work for dev-mode or tests.
@dmlloyd I can dig further, but maybe you have some insight why this is?

gsmet:
We need to add a test for that one.

geoand:
Yes for sure. However at the moment it only works for prod mode so the only test we can have is one of the heavyweight tests that invoke Maven.

geoand:
Never mind, I think I have a fix for dev mode as well.

geoand:
Actually this needs more thought since with my fix you can't override the value in dev-mode.

geoand:
@gsmet had the idea to use the build system properties introduced in https://github.com/quarkusio/quarkus/pull/3137.

This is well worth exploring so I'll do that in a few hours and see what happens

core/runtime/src/main/java/io/quarkus/runtime/ApplicationConfig.java
```diff
@@ -0,0 +1,31 @@
...
+
+    /**
+     * The version of the application.
+     * If not set, defaults to the version of the project
+     */
+    @ConfigItem(defaultValue = "${project.version:unset}")
+    public String version;
+
+    public static boolean isNameUnset(String name) {
+        return name.contains("unset");
```
gsmet: shouldn't it be `equals()` instead of contains? What if you have `unset` in your project name or even if the artifact is called `unset` (which might be a nice project name)?

I would make it `__UNSET__` or something similar.

gsmet (changes requested):
I added a small suggestion. It's nice to have tests for that one.

I think what we need now is a Gradle equivalent.

core/runtime/src/main/java/io/quarkus/runtime/ApplicationConfig.java
geoand: Yes you are right. This got left over from previous stuff I was checking. Let me fix it

geoand: Done

geoand:
prod mode of Gradle will work, so if we get this in I am hoping @gastaldi will take care of the necessary changes for dev mode :)

geoand:
I'm taking this out of draft since I also fixed some CI issues the PR had

core/runtime/src/main/java/io/quarkus/runtime/ApplicationConfig.java
```diff
@@ -0,0 +1,31 @@
...
+
+@ConfigRoot(phase = ConfigPhase.BUILD_AND_RUN_TIME_FIXED)
+public class ApplicationConfig {
+
+    /**
+     * The name of the application.
+     * If not set, defaults to the name of the project.
+     */
+    @ConfigItem(defaultValue = "${project.name:__UNSET__}")
+    public String name;
```
dmlloyd: If we allow it to be unset, it should just be optional.

geoand: Can a property have a default value and also be `Optional`?

geoand: The reason I went this was is because of https://github.com/quarkusio/quarkus/blob/master/core/deployment/src/main/java/io/quarkus/deployment/configuration/RunTimeConfigurationGenerator.java#L738. 
With the current code of `ApplicationConfig` in master the line above is what (I believe) causes the value to not be used at runtime.

dmlloyd: > Can a property have a default value and also be `Optional`?

Yes, it can.  But the default value should be meaningful.  It looks like you might be trying to hack around something the wrong way here.

> The reason I went with this is, is because of https://github.com/quarkusio/quarkus/blob/master/core/deployment/src/main/java/io/quarkus/deployment/configuration/RunTimeConfigurationGenerator.java#L738.
> With the current code of `ApplicationConfig` in master the line above is what (I believe) causes the value to not be used at runtime.

This logic essentially means "if this property has no default value, and no other sub-trees of this value have default values, don't generate pointless default value methods which would always return null for every name".  This logic only controls what values end up as defaults at run time based on what's in the parse tree.  If a property has no default value, then no default value needs to be returned for it.

This logic does *not* generate run-time-readable config values for values which were specified in build time config, which I suspect is what you're trying to do here.  Trying to use this logic for that purpose will yield incorrect results.

geoand: Well I wasn't trying to use something in any particular way. I was lead to that piece of code by debugging and comparing what made this type of configuration different from others that did behave the way we expected. 
But let me go back and take another look.

core/runtime/src/main/java/io/quarkus/runtime/ApplicationConfig.java
```diff
@@ -0,0 +1,31 @@
...
+
+    /**
+     * The version of the application.
+     * If not set, defaults to the version of the project
+     */
+    @ConfigItem(defaultValue = "${project.version:__UNSET__}")
+    public String version;
+
+    public static boolean isNameUnset(String name) {
+        return name.equals("__UNSET__");
```
gastaldi:
```suggestion
        return (name == null) || ("__UNSET__".equals(name));
```

gastaldi: For the cases where the Config object is uninitialized properly

geoand: I'm going to change this stuff based on @dmlloyd's comments, but thanks for the suggestion

core/runtime/src/main/java/io/quarkus/runtime/ApplicationConfig.java
geoand: > This logic does _not_ generate run-time-readable config values for values which were specified in build time config, which I suspect is what you're trying to do here. 

And yes, that is exactly the end result I am looking for :)

geoand: Thanks for the invaluable input! I think this iteration is better. WDYT?

geoand:
@dmlloyd what do you think about the latest iteration?

dmlloyd (approved):
Seems like a legitimate incremental improvement.

geoand:
Thanks!

geoand:
@gsmet OK to merge?

devtools/maven/src/main/java/io/quarkus/maven/DevMojo.java
```diff
@@ -490,7 +490,13 @@ void prepare() throws Exception {
             for (Map.Entry<Object, Object> e : System.getProperties().entrySet()) {
                 devModeContext.getSystemProperties().put(e.getKey().toString(), (String) e.getValue());
             }
+
             devModeContext.getBuildSystemProperties().putAll((Map) project.getProperties());
+
+            // this is a minor hack to allows up to populate ApplicationConfig with default
```
gsmet:
```suggestion
            // this is a minor hack to allow ApplicationConfig to be populated with defaults
```

integration-tests/maven/src/test/java/io/quarkus/maven/it/NativeImageIT.java
```diff
@@ -16,13 +16,13 @@
...
+import org.junit.jupiter.api.condition.EnabledIfEnvironmentVariable;
 
 import io.quarkus.maven.it.verifier.MavenProcessInvocationResult;
 import io.quarkus.maven.it.verifier.RunningInvoker;
 
-/**
- *
- */
+// ensure that this test only runs when GraalVM is configured
+@EnabledIfEnvironmentVariable(named = "GRAALVM_HOME", matches = ".*")
```
gsmet: So we don't run it on CI with Docker? I don't think GRAALVM_HOME is set in this case?

gsmet (changes requested):
I added a suggestion and a question.

integration-tests/maven/src/test/java/io/quarkus/maven/it/NativeImageIT.java
geoand: We do, I checked the docker image that we use and it does set it

gsmet:
BTW, you got back to a previous version? I don't see the thing we talked about.

integration-tests/maven/src/test/java/io/quarkus/maven/it/NativeImageIT.java
gsmet: The Docker image sets it. But is it set when the test is executed?

geoand: It's part of the Docker image, so it's present when the container is run

geoand:
> BTW, you got back to a previous version? I don't see the thing we talked about.

I changed it based on the feedback from @dmlloyd. It essentially does the use the build parameters, but for setting the application name and version directly.

devtools/maven/src/main/java/io/quarkus/maven/DevMojo.java
geoand: Fixed, thanks

integration-tests/maven/src/test/java/io/quarkus/maven/it/NativeImageIT.java
gsmet: I might be missing something but my understanding is the test is not executed inside the Docker image and thus the `GRAALVM_HOME` is not set. Am I missing something?

geoand: Aha, I see what you are saying! Indeed you are absolutely right! I'll try and think of some other way to do this.
However I don't knoq how this test ever passed in the first place... The reason I disabled it is because the PR was failing on it, because of course native-image doesn't exist

geoand: I pushed a change. WDYT?

integration-tests/maven/src/test/java/io/quarkus/maven/it/NativeImageIT.java
```diff
@@ -37,8 +34,13 @@ public void testJavaLibraryPathAtRuntime() throws Exception {
         final RunningInvoker running = new RunningInvoker(testDir, false);
 
         // trigger mvn package -Pnative -Dquarkus.ssl.native=true
-        final String[] mvnArgs = new String[] { "package", "-DskipTests", "-Pnative", "-Dquarkus.ssl.native=true" };
-        final MavenProcessInvocationResult result = running.execute(Arrays.asList(mvnArgs), Collections.emptyMap());
+        List<String> mavenArgs = Arrays.asList("package", "-DskipTests", "-Pnative", "-Dquarkus.ssl.native=true");
+        if (System.getenv("GRAALVM_HOME") == null || System.getenv("GRAALVM_HOME").isEmpty()) {
+            // If the GRAALVM_HOME env var isn't set, try to use docker to build the image.
+            // This is needed because CI doesn't have a GraalVM installation
+            mavenArgs.add("-Dquarkus.native.container-runtime=docker");
```
gsmet: Can't say I'm excited about this because that means we will require Docker to be installed for a standard JDK run? Or am I mistaken?

geoand: Well unless I have completely missread this test, it can't succeed without GraalVM being installed, right?
So the change just allows us to fallback to docker if it's not.

geoand: So essentially now it's safer than it was before, right?

geoand:
It seems like there is an issue with the Windows test.... Seems like Docker on Windows isn't playing nicely in CI...
Should we perhaps ignore this test for Windows?

geoand:
> Should we perhaps ignore this test for Windows?

This is what I ended up doing

devtools/maven/src/main/java/io/quarkus/maven/DevMojo.java
```diff
@@ -490,7 +490,13 @@ void prepare() throws Exception {
             for (Map.Entry<Object, Object> e : System.getProperties().entrySet()) {
                 devModeContext.getSystemProperties().put(e.getKey().toString(), (String) e.getValue());
             }
+
             devModeContext.getBuildSystemProperties().putAll((Map) project.getProperties());
+
+            //  this is a minor hack to allow ApplicationConfig to be populated with defaults
+            devModeContext.getBuildSystemProperties().putIfAbsent("quarkus.application.name", project.getArtifactId());
```
gastaldi: Should it resolve to `project.name` first and fallback to `artifactId`?

geoand: Does the former exist for Gradle as well?

Also note that we already do the same thing for the regular build, it was just missing from dev mode

gastaldi: > Does the former exist for Gradle as well?

I have no idea, but now that you mention the `QuarkusDev` task in the Gradle plugin needs to be updated too, no?

> Also note that we already do the same thing for the regular build, it was just missing from dev mode

:+1:

geoand: It does yes, I was hoping you would take that up once this is merged 😉

geoand:
@gsmet unless you have any suggestions for getting around the NativeImageIT problem some other way, should we merge this?

geoand:
PR updated with a suggestion from @gsmet regarding better handling of the problematic `NativeImageIT`

integration-tests/maven/src/test/java/io/quarkus/maven/it/NativeImageIT.java
geoand: Fixed a different way

integration-tests/maven/src/test/java/io/quarkus/maven/it/NativeImageIT.java
geoand: Fixed a different way

gsmet (approved):
As mentioned on Zulip, I'm not entirely convinced by the annotation approach (especially the fact that it was needed) but since it's there, let's make progress on this one.

PR #6265, opened 2019-12-19
Make Neo4j transport encryption configurable

michael-simons:
This makes the encryption-option of the driver configurable (at least when not in native image mode). I chose the name `Config` for the class to align the naming with the integrations with other platforms. In case you can tell me how to have another class name map to `config` in the properties, than I would rather name `DriverSettings` (analogue to https://github.com/neo4j/neo4j-java-driver-spring-boot-starter/blob/master/neo4j-java-driver-spring-boot-autoconfigure/src/main/java/org/neo4j/driver/springframework/boot/autoconfigure/Neo4jDriverProperties.java#L287).

### DIFF
--- a/extensions/neo4j/runtime/src/main/java/io/quarkus/neo4j/runtime/Neo4jConfiguration.java
+++ b/extensions/neo4j/runtime/src/main/java/io/quarkus/neo4j/runtime/Neo4jConfiguration.java
@@ -1,6 +1,11 @@
 package io.quarkus.neo4j.runtime;
 
+import java.io.File;
+import java.nio.file.Path;
 import java.time.Duration;
+import java.util.Optional;
+
+import org.neo4j.driver.Config;
 
 import io.quarkus.runtime.annotations.ConfigDocSection;
 import io.quarkus.runtime.annotations.ConfigGroup;
@@ -29,6 +34,19 @@ public class Neo4jConfiguration {
     public Authentication authentication;
 
     /**
+     * If the driver should use encrypted traffic.
+     */
+    @ConfigItem
+    public boolean encrypted;
+
+    /**
+     * Configure trust settings for encrypted traffic.
+     */
+    @ConfigItem
+    @ConfigDocSection
+    public TrustSettings trustSettings;
+
+    /**
      * Connection pool.
      */
     @ConfigItem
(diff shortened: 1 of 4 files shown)


extensions/neo4j/runtime/src/main/java/io/quarkus/neo4j/runtime/Neo4jConfiguration.java
```diff
@@ -28,6 +28,13 @@
     @ConfigDocSection
     public Authentication authentication;
 
+    /**
+     * General driver configuration.
+     */
+    @ConfigItem
+    @ConfigDocSection
+    public Config config;
```
gsmet: You can name it `DriverSettingsConfig` and name the property `driverSettings`. Or maybe just `driver` (we are in the config anyway so you clearly expect settings).

It's the name of the property that is taken as the name of the config key. So you would have `quarkus.neo4j.driver-settings.encrypted` for instance.

gsmet (changes requested):

extensions/neo4j/runtime/src/main/java/io/quarkus/neo4j/runtime/Neo4jConfiguration.java
michael-simons: I think I than would rather prefer pulling that setting up into top level. I don't expect an object mapper for Neo4j in Quarkus anytime soon. Wdyt?

gsmet: That was my first thought when I saw the patch so I think it will be good enough.

extensions/neo4j/runtime/src/main/java/io/quarkus/neo4j/runtime/Neo4jConfiguration.java
```diff
@@ -57,6 +64,15 @@
         public boolean disabled = false;
     }
 
+    @ConfigGroup
+    static class Config {
+
+        /**
+         * Flag, if the driver should use encrypted traffic.
```
gsmet:
```suggestion
         * If the driver should use encrypted traffic.
```

michael-simons:
I have implemented the suggestion and added further settings, which I think will be necessary. Thanks for looking into this.

michael-simons:
Ping.

extensions/neo4j/runtime/src/main/java/io/quarkus/neo4j/runtime/Neo4jConfiguration.java
```diff
@@ -28,6 +33,19 @@
     @ConfigDocSection
     public Authentication authentication;
 
+    /**
+     * If the driver should use encrypted traffic.
+     */
+    @ConfigItem(defaultValue = "false")
+    public boolean encrypted = false;
```
machi1990:
```suggestion
    public boolean encrypted;
```

extensions/neo4j/runtime/src/main/java/io/quarkus/neo4j/runtime/Neo4jConfiguration.java
```diff
@@ -57,6 +75,65 @@
...
+            TRUST_CUSTOM_CA_SIGNED_CERTIFICATES,
+
+            TRUST_SYSTEM_CA_SIGNED_CERTIFICATES
+        }
+
+        /**
+         * Configures which trust strategy to apply when using encrypted traffic.
+         */
+        @ConfigItem(defaultValue = "TRUST_SYSTEM_CA_SIGNED_CERTIFICATES")
+        public Strategy strategy = Strategy.TRUST_SYSTEM_CA_SIGNED_CERTIFICATES;
```
machi1990:
```suggestion
        public Strategy strategy;
```

extensions/neo4j/runtime/src/main/java/io/quarkus/neo4j/runtime/Neo4jConfiguration.java
```diff
@@ -57,6 +75,65 @@
...
+         * The file of the certificate to use.
+         */
+        @ConfigItem
+        public Optional<Path> certFile = Optional.empty();
+
+        /**
+         * If hostname verification is used.
+         */
+        @ConfigItem(defaultValue = "false")
+        public boolean hostnameVerificationEnabled = false;
```
machi1990:
```suggestion
        public boolean hostnameVerificationEnabled;
```

extensions/neo4j/runtime/src/main/java/io/quarkus/neo4j/runtime/Neo4jDriverRecorder.java
```diff
@@ -1,6 +1,6 @@
 package io.quarkus.neo4j.runtime;
 
-import static java.util.concurrent.TimeUnit.MILLISECONDS;
+import static java.util.concurrent.TimeUnit.*;
```
machi1990: Can we have explicit imports instead of wildcard one here?

extensions/neo4j/runtime/src/test/java/io/quarkus/neo4j/runtime/Neo4jConfigurationTest.java
```diff
@@ -0,0 +1,87 @@
+package io.quarkus.neo4j.runtime;
+
+import static org.junit.jupiter.api.Assertions.*;
```
machi1990: Let's go with explicit imports over wildcard one here.

extensions/neo4j/runtime/src/main/java/io/quarkus/neo4j/runtime/Neo4jConfiguration.java
machi1990: Oh I just saw that there is test `defaultsShouldWork()` so maybe let's keep it the way it was.

michael-simons:
Thanks @machi1990 for your kind review. I adapted to not using default values in the class itself (I'm used to the way of Boot, need to switch here when working on Quarkus) and also rebased onto the current master.

machi1990:
> Thanks @machi1990 for your kind review. I adapted to not using default values in the class itself (I'm used to the way of Boot, need to switch here when working on Quarkus) and also rebased onto the current master.

Thanks for the contribution.  

I do not know much about neo4j, @michael-simons you are the expert here but LGTM otherwise. @gsmet WDYT?

michael-simons:
The change will allow connecting to Neo4j server using encrypted connection without manually providing the driver bean. 

I plan to do a Quarkus example connecting to our cloud offering with it, so yes, I think it fulfills this goal.

michael-simons:
Ping?

michael-simons:
Ping?

gsmet:
@michael-simons just a heads up that I'm back to addressing the huge backlog of PRs so you will have some news soon.

Sorry about the delay. Things have been a bit crazy until now!

gastaldi:
@michael-simons mind rebasing this PR and squashing the commits? Looks good to me

michael-simons:
I’d love to. Will do next Monday (away from keyboard and things)

michael-simons:
Rebased, squashed and pushed…. Thanks for reviving this. 

I noticed that the configuration has been polished. While I of course accept Quarkus' way of doing it, I do think it's easier to read the explicitly stated defaults, even for scalar values (related: https://twitter.com/rotnroll666/status/1288009552777097216) 

Thanks again!

gsmet (approved):

gsmet:
Sorry it took me so long to review it and merge it! Thanks!

PR #6267, opened 2019-12-19
Make all build time properties visible at run time

### LINKED ISSUE #6255: quarkus.application.version config property is unavailable

dmlloyd:
Fixes #6255


geoand:
I just tried this and it doesn't seem to fix the original issue

dmlloyd:
This makes sense in retrospect.  But this stil might be a legitimate issue.  I'll come back to it later.

gsmet:
@dmlloyd is this PR still a thing or should we close it?

dmlloyd:
It's probably still legitimate.  I'll update it and see what happens.

dmlloyd:
Never mind, looks like it can't be merged anymore.

PR #6269, opened 2019-12-19
Spring Boot properties extension

gytis:
cc @geoand


geoand:
I hope to be able to review this tomorrow but I can't promise :)

devtools/platform-descriptor-json/src/test/resources/fakeextensions.json
```diff
@@ -1000,6 +1000,19 @@
...
+                "keywords": [
+                    "spring-boot",
+                    "properties"
+                ]
+            },
+            "group-id": "io.quarkus",
+            "artifact-id": "quarkus-spring-boot-properties",
+            "version": "999-SNAPSHOT",
+            "description": "Use Spring Boot properties annotations to configure our application"
+        },
```
gsmet: I don't think you need that.

gsmet (commented):
Added one very minor comment.

devtools/platform-descriptor-json/src/test/resources/fakeextensions.json
gytis: OK I'll remove it. I saw other spring modules mentioned in this file so I followed the trend.

geoand:
@gytis did you check why CI is failing?

gytis:
> @gytis did you check why CI is failing?

I think it failed because of `fakeextensions.json`. I've removed it so let's see if it will pass now.

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/configproperties/ConfigPropertiesMetadataBuildItem.java
```diff
@@ -0,0 +1,68 @@
...
+
+import org.jboss.jandex.AnnotationInstance;
+import org.jboss.jandex.AnnotationValue;
+import org.jboss.jandex.ClassInfo;
+import org.jboss.jandex.DotName;
+
+import io.quarkus.arc.config.ConfigProperties;
+import io.quarkus.builder.item.MultiBuildItem;
+
+public final class ConfigPropertiesMetadataBuildItem extends MultiBuildItem {
```
geoand: I really like this, :+1:

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/configproperties/ConfigPropertiesBuildStep.java
```diff
@@ -35,17 +26,26 @@
 
 public class ConfigPropertiesBuildStep {
 
+    @BuildStep
+    void produceConfigPropertiesMetadata(CombinedIndexBuildItem combinedIndex,
```
geoand: This is a really minor complaint, but the rest of the class doesn't use lambda's so can we keep it consistent please?

gytis: OK will do. But I still think that streams are more readable :)

extensions/spring-boot-properties/deployment/src/main/java/io/quarkus/spring/boot/properties/deployment/ConfigurationPropertiesProcessor.java
```diff
@@ -0,0 +1,68 @@
...
+    private ConfigPropertiesMetadataBuildItem createConfigPropertiesMetadataFromClass(AnnotationInstance annotation) {
+        return new ConfigPropertiesMetadataBuildItem(annotation.target().asClass(), getPrefix(annotation));
+    }
+
+    private ConfigPropertiesMetadataBuildItem createConfigPropertiesMetadataFromMethod(AnnotationInstance annotation,
+            IndexView index) {
+        return new ConfigPropertiesMetadataBuildItem(getReturnClassInfo(index, annotation), getPrefix(annotation));
+    }
+
+    private ClassInfo getReturnClassInfo(IndexView index, AnnotationInstance annotation) {
```
geoand: I would just inline this method since it's only called at one spot and more importantly it can only work when the target is a method

devtools/platform-descriptor-json/src/test/resources/fakeextensions.json
geoand: I used to think the same :). Let's keep the streams in the other uses since those are for new classes

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/configproperties/ConfigPropertiesBuildStep.java
geoand: I used to think the same :). Let's keep the streams in the other uses since those are for new classes

integration-tests/spring-boot-properties/src/main/java/io/quarkus/it/spring/boot/InterfacePropertiesResource.java
```diff
@@ -0,0 +1,18 @@
...
+
+import javax.inject.Inject;
+import javax.ws.rs.GET;
+import javax.ws.rs.Path;
+
+@Path("/interface")
+public class InterfacePropertiesResource {
+
+    @Inject
+    private InterfaceProperties properties;
```
geoand: Can we drop `private` from here (and the other instances in the integration test application)? The reason is that quarkus logs a warning for injections into `private` fields and we could use the cleanest test output we can get.

gytis: Didn't know that. Why is package-private preferred to private?

geoand: Because Quarkus then doesn't need to use reflection to set the field. You can read more details here: https://quarkus.io/guides/cdi-reference#native-executables-and-private-members

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/configproperties/ConfigPropertiesBuildStep.java
gytis: Used to? What changed?

geoand: Disillusionment :). I liked them because they were cool, then once I stepped back from them for a while I started disliking them and find that I need to think more about what they are doing

gytis: I sort of agree with simple cases, like the one in this method. Both stream and loop are easy to read in this case. But if it would be a method with a bunch of it/else and loops, I don't think it would more readable than stream with lambdas.

geoand: I agree with you there

extensions/spring-boot-properties/deployment/pom.xml
```diff
@@ -0,0 +1,46 @@
...
+    <name>Quarkus - Spring Boot - Properties - Deployment</name>
+
+    <dependencies>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-spring-boot-properties</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-core-deployment</artifactId>
```
geoand: This isn't necessary because `quarkus-arc-deployment` brings it in

integration-tests/spring-boot-properties/README.md
```diff
@@ -0,0 +1,12 @@
+### Current differences with Spring Boot properties.
+
+1. Default values assigned with `=` operator are not supported.
```
geoand: Are you sure this is true?

integration-tests/spring-boot-properties/README.md
```diff
@@ -0,0 +1,12 @@
...
+2. Getters and setters are needed for all fields, including nested classes.
```
geoand: Are you sure about this?

integration-tests/spring-boot-properties/README.md
```diff
@@ -0,0 +1,12 @@
...
+3. `@Bean` needs to be removed from a method which is annotated with `@ConfigurationProperties`.
```
geoand: It shouldn't be hard to make this work - basically by just ignoring `@Bean`. We should follow up this PR with that

integration-tests/spring-boot-properties/README.md
```diff
@@ -0,0 +1,12 @@
...
+Otherwise two beans will be created and injection will fail.
+4. `@NestedConfigurationProperty` can be used but is not required.
+Nested property classes work out of the box with inner and outer classes.
+5. `@ConstructorBinding` is not supported.
```
geoand: This would be a nice follow up that would also work for Quarkus's `@ConfigProperties`.

integration-tests/spring-boot-properties/README.md
```diff
@@ -0,0 +1,12 @@
...
+6. Spring Boot properties conversion is not supported.
+7. Random values are not supported
```
geoand: What are random values?

integration-tests/spring-boot-properties/README.md
gytis: My bad, this readme wasn't supposed to be in this PR and is not fully correct. I'll remove it. `=` operator works for Spring Boot properties

integration-tests/spring-boot-properties/README.md
gytis: Yes. I tried the Spring Boot way where a config field of a pojo type has only a getter and I ended up getting Quarkus extension that the field doesn't have a setter

integration-tests/spring-boot-properties/README.md
gytis: Spring Boot properties support various expressions in `application.properties`. One of them are random values where you can set something like this: `my.number=${random.int}`. Of course, if it's needed it should be a feature of Quarkus properties rather than Spring Boot properties extension

geoand: I see, thanks!

geoand (approved):
LGTM, thanks!

@gsmet is there anything that you can spot that doesn't add up?

geoand (changes requested):
Actually, the CI config needs to be updated as well: https://github.com/quarkusio/quarkus/blob/master/ci-templates/stages.yml#L362

gytis:
> Actually, the CI config needs to be updated as well: https://github.com/quarkusio/quarkus/blob/master/ci-templates/stages.yml#L362

Should I only include extension name?

geoand:
It's just the (integration test) directory name that needs to be added

geoand (approved):

maxandersen:
does this thing do what I think it does ? adding a quarkus-springboot-properties extension that will pickup springboot keys in application.properties or ?

geoand:
@maxandersen no.

This allows users to use Spring Boot's `@ConfigurationProperties` annotation.

gsmet:
I have one question about that one: we created an extension that specifically introduces support for Spring Boot properties. What if we want to introduce more features? We will create one extension per feature? That doesn't seem ideal.

geoand:
Any ideas on what to call it?

gsmet:
Well, I don't even know if you plan to add more features? But if so, it would probably be "spring-boot" instead of having 5 different extensions with different features.

But I'm not totally sure what our target is here, just raising the issue.

geoand:
I see your point and I do have the same concerns. However, naming the extension spring-boot would probably lead to massive confusion and a ton of issues being opened asking why this or that spring Boot feature doesn't work.
Another option would be to roll this into Spring DI.
The only problem is that the jar size would increase

gytis:
Maybe we could merge the extensions in the future if the number becomes too big? At the end of the day all the Spring and Spring Boot features are marked as preview.

However, merging this extension with Spring DI would not be an ideal scenario IMO. The properties API that is used in this extension comes from Spring Boot project directly. Spring DI on the other had is a separate project and has a much smaller scope.

geoand:
> Maybe we could merge the extensions in the future if the number becomes too big? At the end of the day all the Spring and Spring Boot features are marked as preview.
> 
+1

geoand:
I propose that for the time being we leave the name as is and leave it in preview state, that way we can change it later on if needed

gsmet (approved):

gsmet:
OK, I suppose I can live with that :).

PR #6272, opened 2019-12-19
Reactive routes - detect unordered conflicting routes

### LINKED ISSUE #6165: Reactive routes - log a warning if multiple routes match the same path and no order is declared
The routes are tested/handlers executed in a random order.

See https://groups.google.com/forum/#!topic/quarkus-dev/pJDC32SBLWA.

mkouba:
- resolves #6165

The warning looks like:
```
[WARNING] Route org.acme.quarkus.sample.MyDeclarativeRoutes#hello() can match the same request and has the same order [0] as:
	- org.acme.quarkus.sample.MyDeclarativeRoutes#greetings()
[WARNING] You can use @Route#order() to ensure the routes are not executed in random order
```

### DIFF
--- a/extensions/vertx-web/deployment/src/main/java/io/quarkus/vertx/web/deployment/VertxWebProcessor.java
+++ b/extensions/vertx-web/deployment/src/main/java/io/quarkus/vertx/web/deployment/VertxWebProcessor.java
@@ -1,9 +1,18 @@
 package io.quarkus.vertx.web.deployment;
 
 import java.io.IOException;
+import java.util.Arrays;
 import java.util.Collections;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Iterator;
+import java.util.LinkedHashSet;
 import java.util.LinkedList;
 import java.util.List;
+import java.util.Map;
+import java.util.Map.Entry;
+import java.util.Objects;
+import java.util.Set;
 import java.util.function.Function;
 
 import javax.inject.Singleton;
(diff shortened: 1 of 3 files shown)


jaikiran:
Hello @mkouba, I don't have much knowledge in this area, so probably a basic question - is there ever a case where it is OK to have conflicting routes with the same order? In order words, should this change introduce an exception/error instead of just a WARN?

mkouba:
@jaikiran well, I think that it is legal to register multiple routes with the same order in Vertx. @cescoffier am I right? So that's the reason why I chose warning.

cescoffier:
Yes, it's legal to have conflicting routes with the same order. It's used when building filters for example.

cescoffier:
I checked the changes and I'm ok with them. If someone once to have more control over the routes, there is always the possibility to inject the `Router` directly and register the routes manually.

extensions/vertx-web/deployment/src/main/java/io/quarkus/vertx/web/deployment/VertxWebProcessor.java
```diff
@@ -397,4 +428,84 @@ private static String dashify(String value) {
...
+                            .append(method.name()).append("()");
+                }
+                LOGGER.warnf(
+                        "Route %s#%s() can match the same request and has the same order [%s] as:%s",
+                        firstMethod.declaringClass().name(),
+                        firstMethod.name(), firstMatcher.getOrder(), conflictingRoutes);
+            }
+        }
+        if (conflictExists) {
+            LOGGER.warn("You can use @Route#order() to ensure the routes are not executed in random order");
```
cescoffier: it's actually not random, it's the insertion order.

cescoffier (approved):

extensions/vertx-web/deployment/src/main/java/io/quarkus/vertx/web/deployment/VertxWebProcessor.java
mkouba: Yes, but it's not guaranteed that we register the methods in the same order.

PR #6274, opened 2019-12-19
Put trace context into MDC

pavolloffay:
Resolves https://groups.google.com/forum/#!msg/quarkus-dev/dej_Ref-YAQ/yH_hZysTBQAJ

* This change adds trace context (traceId, spanId, sampled) into MDC. 
* The feature is automatically enabled and can be configured via `quarkus.jaeger.log-trace-context=true`

Follow up feature can be to add baggage/distributed context into MDC. As we don't want to log all baggage content we might provide a white list configuration to identify which keys should be logged.

Example log output with `%X` pattern
```
11:33:14 INFO  {sampled=true, spanId=6f81a2edfb7c7699, traceId=6f81a2edfb7c7699} [or.ac.qu.GreetingResource] (executor-thread-63) hello
```

Signed-off-by: Pavol Loffay <ploffay@redhat.com>

### DIFF
--- a/extensions/jaeger/runtime/src/main/java/io/quarkus/jaeger/runtime/JaegerConfig.java
+++ b/extensions/jaeger/runtime/src/main/java/io/quarkus/jaeger/runtime/JaegerConfig.java
@@ -111,4 +111,10 @@ public class JaegerConfig {
     @ConfigItem
     public Optional<String> senderFactory;
 
+    /**
+     * Whether the trace context should be logged.
+     */
+    @ConfigItem(defaultValue = "true")
+    public Boolean logTraceContext;
+
 }
--- a/extensions/jaeger/runtime/src/main/java/io/quarkus/jaeger/runtime/JaegerDeploymentRecorder.java
+++ b/extensions/jaeger/runtime/src/main/java/io/quarkus/jaeger/runtime/JaegerDeploymentRecorder.java
@@ -60,6 +60,8 @@ public class JaegerDeploymentRecorder {
         initTracerProperty("JAEGER_TAGS", jaeger.tags, tags -> tags.toString());
         initTracerProperty("JAEGER_PROPAGATION", jaeger.propagation, format -> format.toString());
         initTracerProperty("JAEGER_SENDER_FACTORY", jaeger.senderFactory, sender -> sender);
+        initTracerProperty(QuarkusJaegerTracer.LOG_TRACE_CONTEXT, Optional.of(jaeger.logTraceContext),
+                logTraceContext -> logTraceContext.toString());
     }
 
     private <T> void initTracerProperty(String property, Optional<T> value, Function<T, String> accessor) {
--- a/extensions/jaeger/runtime/src/main/java/io/quarkus/jaeger/runtime/QuarkusJaegerTracer.java
+++ b/extensions/jaeger/runtime/src/main/java/io/quarkus/jaeger/runtime/QuarkusJaegerTracer.java
@@ -6,9 +6,11 @@ import io.opentracing.Span;
 import io.opentracing.SpanContext;
 import io.opentracing.Tracer;
 import io.opentracing.propagation.Format;
+import io.opentracing.util.ThreadLocalScopeManager;
 
 public class QuarkusJaegerTracer implements Tracer {
 
+    static final String LOG_TRACE_CONTEXT = "JAEGER_LOG_TRACE_CONTEXT";
     private static volatile Tracer tracer;
 
     @Override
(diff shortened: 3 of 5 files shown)


extensions/jaeger/runtime/src/main/java/io/quarkus/jaeger/runtime/QuarkusJaegerTracer.java
```diff
@@ -6,9 +6,11 @@
 import io.opentracing.SpanContext;
 import io.opentracing.Tracer;
 import io.opentracing.propagation.Format;
+import io.opentracing.util.ThreadLocalScopeManager;
 
 public class QuarkusJaegerTracer implements Tracer {
 
+    static final String LOG_TRACE_CONTEXT = "JAEGER_LOG_TRACE_CONTEXT";
     private static volatile Tracer tracer;
```
pavolloffay: I have noticed that changes to Jaeger configuration are not applied in dev mode (`maven quarkus:dev`). This is because tracer is a singleton. This could be fixed in a separate PR.

extensions/jaeger/runtime/src/main/java/io/quarkus/jaeger/runtime/JaegerConfig.java
```diff
@@ -111,4 +111,10 @@
     @ConfigItem
     public Optional<String> senderFactory;
 
+    /**
+     * Whether the trace context should be logged.
+     */
+    @ConfigItem(defaultValue = "true")
+    public Optional<Boolean> logTraceContext;
```
gastaldi: You don't need to use `Optional` here if you have a defaultValue set in the @ConfigItem

extensions/jaeger/runtime/src/main/java/io/quarkus/jaeger/runtime/JaegerDeploymentRecorder.java
```diff
@@ -60,6 +60,8 @@ private void initTracerConfig(JaegerConfig jaeger) {
         initTracerProperty("JAEGER_TAGS", jaeger.tags, tags -> tags.toString());
         initTracerProperty("JAEGER_PROPAGATION", jaeger.propagation, format -> format.toString());
         initTracerProperty("JAEGER_SENDER_FACTORY", jaeger.senderFactory, sender -> sender);
+        initTracerProperty(QuarkusJaegerTracer.LOG_TRACE_CONTEXT, jaeger.logTraceContext,
```
gastaldi: Why is this property the only one defined  in a constant? I think it should be an all-or-nothing here :)

gsmet: I think it's a constant because it's used twice, once to be set and another time to be read. It makes mostly sense to me in this situation (and I'm the consistency maniac ;)).

I added a small comment about this below, though.

extensions/jaeger/runtime/src/main/java/io/quarkus/jaeger/runtime/QuarkusJaegerTracer.java
```diff
@@ -21,13 +23,25 @@ private static Tracer tracer() {
...
+                            .build();
                 }
             }
         }
         return tracer;
     }
 
+    private static ScopeManager getScopeManager() {
+        ScopeManager scopeManager = new ThreadLocalScopeManager();
+        String logTraceContext = System.getProperty(LOG_TRACE_CONTEXT);
```
gsmet: I'm a bit torn on this one. Not sure using system properties is the best way to pass this value.

It's a bit different from the others because we set the others to push values to Jaeger.

gsmet (changes requested):
Added a few things.

extensions/jaeger/runtime/src/main/java/io/quarkus/jaeger/runtime/QuarkusJaegerTracer.java
pavolloffay: The problem is that quarkus tracer is a singleton and `tracer()` method which resolves tracer (if null) is `static` - and it's called on every tracer usage.

I would rather go with this - as it is the simplest change and conceptually follows the same approach. I can open refactoring PR which will fix https://github.com/quarkusio/quarkus/pull/6274#discussion_r360306758 and also remove passing configuration via a system property.

extensions/jaeger/runtime/src/main/java/io/quarkus/jaeger/runtime/JaegerDeploymentRecorder.java
pavolloffay: The constant is used because it is used two times.

pavolloffay:
@gsmet could you please have a look again?

gsmet (approved):

gsmet:
Merged, thanks!

PR #6277, opened 2019-12-19
Mention possibility to disable JSON for non prod

mejlholm:
* Mention possibility to disable JSON for non prod

Remove <build> surrounding extension - it's not valid. 

Add small example of how to use human readable logging for dev/test and structured JSON for production.

### DIFF
--- a/docs/src/main/asciidoc/logging.adoc
+++ b/docs/src/main/asciidoc/logging.adoc
@@ -91,7 +91,6 @@ application POM as the following snippet illustrates.
 .Modifications to POM file to add the JSON logging extension
 [source,xml]
 ----
-<build>
   <dependencies>
     <!-- ... your other dependencies are here ... -->
     <dependency>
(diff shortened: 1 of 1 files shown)


docs/src/main/asciidoc/logging.adoc
```diff
@@ -91,21 +91,28 @@ application POM as the following snippet illustrates.
...
     </dependency>
   </dependencies>
-</build>
 ----
 
 The presence of this extension will, by default, replace the output format configuration from the console configuration.
 This means that the format string and the color settings (if any) will be ignored.  The other console configuration items
 (including those controlling asynchronous logging and the log level) will continue to be applied.
 
+For some, it will make sense to use logging that is humanly readable (unstructured) in dev mode and JSON logging (structured) in production mode. This can be achieved using the different profiles, as shown in the following configuration.
```
gsmet:
```suggestion
For some, it will make sense to use logging that is humanly readable (unstructured) in dev mode and JSON logging (structured) in production mode. This can be achieved using different profiles, as shown in the following configuration. 
```
would be better I think.

gsmet (changes requested):
Nice catch and nice addition. I added a suggestion and the commit message looks weird, could you fix it?

mejlholm:
Fixed now, but resulted in 4 commits. Can you squash when merging or should i try to cleanup it up first?

oscerd (approved):

gsmet (approved):

Issue #5791, opened 2019-11-27
NPE is thrown with a Spring Web Controller contains a @PostConstruct annotation


geoand:

geoand:
This RESTEasy [PR](https://github.com/resteasy/Resteasy/pull/2233) should fix the problem.

rsvoboda:
PR is in https://github.com/resteasy/Resteasy/commits/4.4.2.Final
Master now uses 4.4.2.Final

Can be this issue closed ?

geoand:
Has been fixed in RESTEasy

Issue #4404, opened 2019-10-07
Header Propagation does not work when using Asynchronous Rest Clients


csotiriou:
**Describe the bug**
Rest Client Header Propagation does not work when using Asynchronous Rest Clients. I have created a small sample which reproduces the problem with Quarkus 0.23.2, with both `CompletionStage` and `Single`

[Archive.zip](https://github.com/quarkusio/quarkus/files/3696314/Archive.zip)

**Expected behavior**
When using `Single` or `CompletionStage` in the REST clients' responses, it seems that the header propagation does not work.

**Actual behavior**
when `HeaderPropagator` is called, the container request headers are empty.

**To Reproduce**
1. Open the attached project and run it. Put a breakpoint inside the `HeaderPropagator`
2. GET `http://0.0.0.0:8080/hello`. The breakpoint will get hit, the `containerRequestHeaders` will be full of your browser's headers.
3. GET `http://0.0.0.0:8080/hello/async` or `http://0.0.0.0:8080/hello/single`. The same array will be empty.

I have also included all necessary libraries as instructred here https://quarkus.io/guides/context-propagation-guide and had the same results. Header propagation only works in synchronous mode.

kenfinnigan:
The issue is that https://github.com/quarkusio/quarkus/blob/master/extensions/rest-client/runtime/src/main/java/io/quarkus/restclient/runtime/IncomingHeadersProvider.java#L28 returns `null` because the context data is held in a `ThreadLocal`.

With a sync call, we're on the same thread, a worker thread, but when using `CompletionStage` we're on a separate thread which is why it's empty.

@FroMage any thoughts on where/how to propagate the `ThreadLocal` for RESTEasy context data?

FroMage:
The headers of what? I thought client calls had no context _at all_. The spec says:

> Except for Configuration and Providers, which are injectable in both client and server-side providers, all the other types are server-side only.

So `HttpRequest` is server-side and should not be visible in clients and if they are visible it's by accident and they point to the server's request (when the client is running in a jax-rs server endpoint).

kenfinnigan:
When it's used on its own, then yes.

When REST Client is used inside a JAX-RS endpoint, it's possible for the headers to be propagated from that incoming request to the REST Client call

FroMage:
[...]

PR #6283, opened 2019-12-20
Bump RESTEasy to 4.4.2.Final

### LINKED ISSUE #5752: Microprofile rest client - Can't inject config/beans to providers
**Describe the bug**
Any provider added to rest client via `@Provider` annotation is unable to inject config bean or any other application scoped bean.
```java
@RegisterRestClient
@RegisterProvider(RequestFilter.class)
public interface EchoService {....}
```
And example the provider impl:
```java
public class RequestFilter implements ClientRequestFilter {
  @Inject
  AppConfig config;  //is always null

  @Inject
  SomeBean bean; //is always null

  @Override
  public void filter(ClientRequestContext requestContext) throws IOException {...}
```

**Expected behavior**
Config or application scope bean injected and not null

**Actual behavior**
Any bean injected via `@Inject` is alway null (no errors in logs during argumentation nor during runtime.

**To Reproduce**
1. Use reproducer: https://github.com/marcinczeczko/quarkus-rest-providers-reproducer
2. `mvn compile quarkus:dev`
3. `curl http://localhost:8080/echo`
4. Check the logs - the injected config & app bean shows `null` reference

**Configuration**
- See reproducer

**Screenshots**
N/A

 - Output of `uname -a` or `ver`: `18.7.0 Darwin Kernel Version 18.7.0: Sat Oct 12 00:02:19 PDT 2019; root:xnu-4903.278.12~1/RELEASE_X86_64 x86_64`
 - Output of `java -version`: 
```
java version "1.8.0_192"
Java(TM) SE Runtime Environment (build 1.8.0_192-b12)
Java HotSpot(TM) 64-Bit Server VM (build 25.192-b12, mixed mode)
```
 - GraalVM version (if different from Java): Didn't check on graalvm
 - Quarkus version or git rev: 1.0.0.Final

**Additional context**

I found out there is a closed issue for similar problem, but looks like the fix doesn't work, see: https://github.com/quarkusio/quarkus/issues/2773

geoand:
Fixes #5752, #5791


geoand:
Seems like `org.jboss.resteasy.microprofile.client.header.IncomingHeadersProvider` is gone in `4.4.2.Final`. I am not sure how this should be handled @asoldano

gsmet:
Well, I don't think we should remove a class in a micro release. It seems wrong.

geoand:
I agree, let's wait and see what @asoldano says

asoldano:
This went unnoticed because we were previously planning for a minor and then decided to do a micro as there were basically bug fixes only.
I'm on mobile and can't check for a couple of hours, but is this actually a problem here? I think I reme2mber the change came from a big fix from Stephane. If you prefer not upgrading to this version I can probably retag as 4.5.0 in the next few days. Generally speaking strict backward compatibility is granted on the spi modules only.

geoand:
It's a problem for Quarkus because we have a class (`io.quarkus.restclient.runtime.IncomingHeadersProvider`) that extends it.

@FroMage do we absolutely need that class or can we get rid of it in Quarkus as well?

FroMage:
This release includes a fix for https://github.com/quarkusio/quarkus/issues/4404 and that provider class was no longer useful, so we're good to upgrade.

FroMage:
We can remove our subclass.

geoand:
Thanks for the input @FroMage!

PR updated

asoldano (approved):
Thanks

geoand:
There seems to be some issue with Spring Web that I'll need to sort out

geoand:
@gsmet do you remember why we are on Jackson 2.9.10 instead of `2.10`? RESTEasy uses the latter (in both `4.4.1.Final` and `4.4.2.Final` so I am not sure why we are only encountering the problem in this PR and not `master`).

geoand:
OK, so it fails in the PR because now RESTEasy uses this type (for example `Jackson2JsonpInterceptor` references it indirectly in `clinit` - which wasn't the case in `4.4.1.Final`).
So we'll need to update to Jackson 2.10.0 if we want to pick this up.

gsmet:
We are waiting for the Vert.x team to give us the green light.

@cescoffier @vietj is there any plan to move to Jackson 2.10 soon?

asoldano:
You should definitely move to jackson 2.10 series for security reasons...

gsmet:
I force pushed a rebase of this one on top of the Jackson 2.10 upgrade.

gsmet (approved):

machi1990 (approved):

machi1990:
Merged thanks @asoldano

PR #6299, opened 2019-12-20
Do not assume that option values are always String[]

gastaldi:
From @michalszynkiewicz feedback in https://quarkusio.zulipchat.com/#narrow/stream/187038-dev/topic/ClassCastException.20on.20master

### DIFF
--- a/independent-projects/bootstrap/core/src/main/java/io/quarkus/bootstrap/resolver/maven/options/BootstrapMavenOptions.java
+++ b/independent-projects/bootstrap/core/src/main/java/io/quarkus/bootstrap/resolver/maven/options/BootstrapMavenOptions.java
@@ -90,7 +90,11 @@ public class BootstrapMavenOptions {
 
     public String[] getOptionValues(String name) {
         final Object o = options.get(name);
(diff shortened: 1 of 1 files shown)


aloubyansky:
Actually, it should be fixed here https://github.com/quarkusio/quarkus/blob/96067aeeffe3abdf4fc9984036d354c15c31e699/independent-projects/bootstrap/core/src/main/java/io/quarkus/bootstrap/resolver/maven/options/BootstrapMavenOptionsParser.java#L23

Instead of calling getOptionValue(..) for the active profiles, it should be calling getOptionValues(..).

gastaldi:
Anyway putting everything in a Map with Object as value seems to be a bit creepy. Maybe change the generic parameter to the map value to `String[]` would be better

gastaldi:
Or a Multimap data structure

aloubyansky:
I hope that class is gonna go away with the refactoring I'm working on right now. So don't go too creative there.

gastaldi:
Awesome, so I'll close it and wait for your changes then ;)

PR #6302, opened 2019-12-21
sentry.io is not opensource

maxandersen:


oscerd (approved):

gastaldi (approved):

gsmet:
This is redundant with the note added here: https://github.com/quarkusio/quarkus/pull/6284 .

Not saying one is better than the other but I think we should only have one not two.

gsmet:
I created https://github.com/quarkusio/quarkus/pull/6305 .

PR #6306, opened 2019-12-21
[WIP] Deprecate enableJni option and always consider it true

### LINKED ISSUE #6101: Deprecate enableJni option as GraalVM 19.3 requires JNI support
With GraalVM 19.3.0, JNI always needs to be enabled.

We should:
- deprecate enableJni (and any other option allowing to set it)
- always consider it true
- warn if it has been set to false

gwenneg:
Fixes #6101 

This PR is ready for a review, but it shouldn't be merged into `master` before we're sure that the following Quarkus release depends on GraalVM `19.3.0` or greater.

### DIFF
--- a/core/deployment/src/main/java/io/quarkus/deployment/JniProcessor.java
+++ b/core/deployment/src/main/java/io/quarkus/deployment/JniProcessor.java
@@ -1,8 +1,9 @@
 package io.quarkus.deployment;
 
-import java.util.Collections;
 import java.util.List;
 import java.util.Optional;
+
+import org.jboss.logging.Logger;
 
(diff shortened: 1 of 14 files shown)


gwenneg:
I'm putting a WIP label to make sure it isn't merged too soon.

gsmet (approved):
Looks good to me but I agree we need to merge it very last minute.

I'm still unsure we will be able to have the RSS regression fixed in 19.3.1.

gwenneg:
Thanks for the review!

gsmet:
If we get to support both versions, we will have to tune that one a bit.

gwenneg:
Superseded by #6792.

PR #6307, opened 2019-12-22
feat(flyway): integrate flyway's validate functionality into quarkus

### LINKED ISSUE #6240: Push Flyway integration to the next level
**Description**

Since Quarkus has already support for Flyway, it would be great to support some other "operations" (or commands) to make our life easier while dealing with it:

- `quarkus.flyway.validate-on-migrate` — Option to automatically call `validate` when performing the migration
- `quarkus.flyway.out-of-order` — Allow migrations to be run out of order or not
- `quarkus.flyway.clean-disabled` — Whether to disable cleaning of the database; extremely useful during development and when tests are about (or prepared to) be run, so we don't have to do this manually all the time

x80486:
Closes #6240

### DIFF
--- a/extensions/flyway/runtime/src/main/java/io/quarkus/flyway/runtime/FlywayRecorder.java
+++ b/extensions/flyway/runtime/src/main/java/io/quarkus/flyway/runtime/FlywayRecorder.java
@@ -45,12 +45,10 @@ public class FlywayRecorder {
     }
 
     private void clean(BeanContainer container, AnnotationLiteral<? extends Annotation> qualifier) {
-        Flyway flyway = container.instance(Flyway.class, qualifier);
-        flyway.clean();
+        container.instance(Flyway.class, qualifier).clean();
     }
 
     private void migrate(BeanContainer container, AnnotationLiteral<? extends Annotation> qualifier) {
-        Flyway flyway = container.instance(Flyway.class, qualifier);
-        flyway.migrate();
+        container.instance(Flyway.class, qualifier).migrate();
     }
 }
--- a/docs/src/main/asciidoc/flyway.adoc
+++ b/docs/src/main/asciidoc/flyway.adoc
@@ -43,15 +43,19 @@ In your `pom.xml`, add the following dependencies:
 </dependencies>
 --
 
-Flyway support relies on the Quarkus datasource config. 
+Flyway support relies on the Quarkus datasource config.
 It can be customized for the default datasource as well as for every <<multiple-datasources,named datasource>>.
-First, you need to add the datasource config to the `{config-file}` file 
+First, you need to add the datasource config to the `{config-file}` file
 in order to allow Flyway to manage the schema.
 Also, you can customize the Flyway behaviour by using the following properties:
 
 `quarkus.flyway.migrate-at-start`::
 **true** to execute Flyway automatically when the application starts, **false** otherwise. +
 **default:** false
+
+`quarkus.flyway.validate-on-migrate`::
+**true** to validate the applied migrations against the available ones, **false** otherwise. +
+**default:** true
 
 `quarkus.flyway.clean-at-start`::
 **true** to execute Flyway clean command automatically when the application starts, **false** otherwise. +
@@ -174,7 +178,7 @@ public class MigrationService {
 
 == Multiple datasources
 
-Flyway can be configured for multiple datasources. 
+Flyway can be configured for multiple datasources.
 The Flyway properties are prefixed exactly the same way as the named datasources, for example:
 
 [source,properties]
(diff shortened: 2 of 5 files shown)


extensions/flyway/deployment/src/test/java/io/quarkus/flyway/test/FlywayExtensionValidateOnMigrateFailureTest.java
```diff
@@ -0,0 +1,31 @@
...
+    @RegisterExtension
+    final QuarkusUnitTest config = new QuarkusUnitTest().setArchiveProducer(() -> ShrinkWrap.create(JavaArchive.class)
+            .addAsResource("db/migration/V1.0.0__Quarkus.sql")
+            .addAsResource("validate-on-migrate-failure-config.properties", "application.properties"))
+            .setExpectedException(FlywayException.class);
+
+    @Test
+    @DisplayName("validate on migrate fails if migrations applied do not match the one(s) available locally")
+    public void testFlywayConfigInjection() {
+        // ...
```
gastaldi: Add a fail() if this test is supposed to never be called

extensions/flyway/deployment/src/test/java/io/quarkus/flyway/test/FlywayExtensionValidateOnMigrateTest.java
```diff
@@ -0,0 +1,44 @@
...
+import io.agroal.api.AgroalDataSource;
+import io.quarkus.test.QuarkusUnitTest;
+
+/**
+ * Test suite for Flyway's validate.
+ * <p>
+ * Assures Flyway's validate runs successfully after SQL migrations are applied, by checking that
+ * they do match the one(s) available locally.
+ */
+public class FlywayExtensionValidateOnMigrateTest {
```
gastaldi: There is no need to test Flyway (they have their own tests) :)

extensions/flyway/runtime/src/main/java/io/quarkus/flyway/runtime/FlywayRecorder.java
```diff
@@ -34,23 +34,31 @@ public void doStartActions(FlywayRuntimeConfig config, BeanContainer container)
...
+        }
         for (Entry<String, FlywayDataSourceRuntimeConfig> configPerDataSource : config.namedDataSources.entrySet()) {
             if (configPerDataSource.getValue().cleanAtStart) {
                 clean(container, FlywayDataSource.FlywayDataSourceLiteral.of(configPerDataSource.getKey()));
             }
             if (configPerDataSource.getValue().migrateAtStart) {
                 migrate(container, FlywayDataSource.FlywayDataSourceLiteral.of(configPerDataSource.getKey()));
             }
+            if (configPerDataSource.getValue().validateOnMigrate) {
+                validate(container, FlywayDataSource.FlywayDataSourceLiteral.of(configPerDataSource.getKey()));
```
gastaldi: There is no need to call validate again since you are already setting` configure.validateOnMigrate` in `FlywayCreator`

extensions/flyway/deployment/src/test/java/io/quarkus/flyway/test/FlywayExtensionValidateOnMigrateFailureTest.java
x80486: This one is tricky. The `setExpectedException` call will take of the test, failing when the application boostraps (the correct behavior), but the method is only there because otherwise the class won't execute any test(s).

If you have a better way to test this I'm all ears.

extensions/flyway/deployment/src/test/java/io/quarkus/flyway/test/FlywayExtensionValidateOnMigrateTest.java
x80486: This is actually not testing `Flyway` (in my opinion) but the integration with Quarkus: `quarkus.flyway.validate-on-migrate`, which I defaulted to run all the time (read: `true`). The call to `flyway.validate()` in `testFlywayConfigInjection` can be actually deleted since everything happens when the application starts.

extensions/flyway/runtime/src/main/java/io/quarkus/flyway/runtime/FlywayRecorder.java
x80486: OK, I can't probably relate that other than this setting might be a global one rather than per datasource, because I see the other calls being made for `cleat at start` and `migrate`...but I will leave this up to you since you know that better than myself.

In the end, you are stating that: this `if` statement (inside the `for` loop) is unnecessary and can be deleted, right?

gastaldi: I believe so, yes. Looking at the Flyway sources it seems thaf validate() is called internally when the `validateOnMigrate` is true

gsmet (approved):

gsmet:
Merged, thanks!

PR #6310, opened 2019-12-22
Ensure that the Kotlin compiler adds method param names to bytecode

### LINKED ISSUE #6041: Kotlin data class not work without annotation
I try https://quarkus.io/guides/rest-json  on kotlin. I want to use simple data class, without any annotation. 
```
data class Fruit(
        val name : String,
        val description : String)
```

if I use quarkus-resteasy-jackson
**(no Creators, like default construct, exist): cannot deserialize from Object value (no delegate- or property-based Creator)** 
If I use quarkus-resteasy-jsonb
I get   **No default constructor found.**
This is bad practice to use default constructor in this case.  
related  [topic](https://github.com/quarkusio/quarkus/issues/3954)

<quarkus.platform.version>1.0.1.Final</quarkus.platform.version>

geoand:
We already do this for Java, so let's make it consistent for Kotlin too

Fixes: #6041


gastaldi (approved):
LGTM

PR #6313, opened 2019-12-22
Adopt new label nomenclature in issue templates

gastaldi:


maxandersen (approved):
I missed those. Thanks. Looks good.

Issue #4766, opened 2019-10-22
swagger-ui is not mapping oauth2-redirect.html


btsibr:
**Describe the bug**
In Quarkus, swagger-UI sets the redirect_uri for OAUTH2 requests to /oauth2-redirect.html.  If this file doesn't exist, the Authorize dialog (after asking for scopes) returns to a 404 page (that includes a suggestion for /webjars/swagger-ui/3.20.9/oauth2-redirect.html).

Quarkus 0.25.0 w/ swagger-ui and smallrye-jwt extensions.

**Expected behavior**
The authorize dialog successfully returns to swagger-ui with a token.

This can currently be validated by extracting oauth2-redirect.html from the swagger-ui webjar and placing it in src/main/resources/META-INF.

Expectation is that one of the following happens:
* swagger-ui extension correctly configures swagger's redirect_uri to be /swagger-ui/oauth2-redirect.html (respecting parameters that may have changed /swagger-ui)
* OR Quarkus is configured to respond to /oauth2-redirect.html

**Actual behavior**
The Authorize dialog (after asking for scopes) returns to a 404 page (that includes a suggestion for /webjars/swagger-ui/3.20.9/oauth2-redirect.html).

**To Reproduce**
1. Follow the JWT quickstart steps from https://quarkus.io/guides/jwt-guide
2. Add the swagger-ui extension
3. Attempt to access the secured endpoint using the swagger-ui Authorize button

**Environment (please complete the following information):**
 - CentOS7
 - 1.8.0_222
 - quarkus:dev
 - 0.25.0

btsibr:
I believe that it just needs to set the oauth2RedirectUrl when it modifies the swagger-ui index.html

https://swagger.io/docs/open-source-tools/swagger-ui/usage/configuration/.

As an aside, it'd be nice to be able to set the validatorUrl.

gsmet:
@irenakezic this might be a good issue for you. From what I can see, we need to expose `oauth2-redirect.html` when doing the Swagger UI magic.

Better check beforehand if it's still inaccessible as it's a pretty old issue.

irenakezic:
@gsmet ok thx! I will give it a try but probably not before the weekend.

irenakezic:
@btsibr I tried reproducing this but the Authorize dialog seems to work fine for me. 
Could you share a exact project you use to reproduce this issue?
Did you add any open api annotations to your code or do you have an openapi.yaml to share?
Thanks!

[...]

Issue #5639, opened 2019-11-20
[openapi] Add Ability to Configure Swagger UI


valones:
I can configure Swagger UI via URL query params, but would be nice if I could configure some params in `application.properties` file, like the `docExpansion` param for example.

swagger oficial doc: https://swagger.io/docs/open-source-tools/swagger-ui/usage/configuration/

gsmet:
If you're familiar with Swagger UI, that could be a nice first issue to start contributing!

valones:
To be honest, my first contact with swagger was thanks to quarkus!  :)
I just investigated how I could load the interface with the elements collapsed:
`http://localhost:8080/swagger-ui/?docExpansion=none`

aoudiamoncef:
@gsmet, what approach would you like to implement for configuring Swagger UI ? 

Swagger UI accepts configuration parameters in four locations.
From lowest to highest precedence:

1. The swagger-config.yaml in the project root directory, if it exists, is baked into the application
2. configuration object passed as an argument to Swagger UI (SwaggerUI({ ... }))
3. configuration document fetched from a specified configUrl
4. configuration items passed as key/value pairs in the URL query string

Thanks

aoudiamoncef:
Hi @dmlloyd,
I studied the swagger-ui extension in order  add this feature.
We could add all available [customisations](https://swagger.io/docs/open-source-tools/swagger-ui/usage/configuration/) customisations as query parameters(All tagged as Docker variables in the doc). 
Could i use an existing URL builder from Apache common for example or, i should build own utility class to avoid adding external dependencies ?

Thanks

dmlloyd:
Fewer dependencies is better, but other than that, I don't have an opinion.  The most important thing is that it's correct (meaning, all URL components are correctly encoded).

gsmet:
Wouldn't `The swagger-config.yaml in the project root directory, if it exists, is baked into the application` be the more practical for us? We could generate it based on the configuration while building the application?

gsmet:
Oh wait, wouldn't it work to just have the user add a `swagger-config.yaml` to the `src/main/resources/` directory?

If so, I would say let's go for it and document it properly. I don't think it's a big problem if the file is only used in dev mode but still present in the prod application.

aoudiamoncef:
@gsmet,
[...]

PR #6314, opened 2019-12-22
feat(SwaggerUI): Add Swagger UI configuration parameters

aoudiamoncef:
Hi @dmlloyd ,
This PR is related to #5639. I'm waiting for feedbacks
Moncef


extensions/swagger-ui/deployment/src/main/java/io/quarkus/swaggerui/deployment/SwaggerUiConfig.java
```diff
@@ -0,0 +1,115 @@
...
+    /**
+     * URL to fetch external configuration document from.
+     */
+    @ConfigItem
+    Optional<String> configUrl;
+
+    /**
+     * Enables deep linking for tags and operations.
+     * 
+     * @see <a href="https://swagger.io/docs/open-source-tools/swagger-ui/usage/deep-linking</a>
```
gsmet:
```suggestion
     * @see <a href="https://swagger.io/docs/open-source-tools/swagger-ui/usage/deep-linking">Swagger UI documentation</a>
```

extensions/swagger-ui/deployment/src/main/java/io/quarkus/swaggerui/deployment/SwaggerUiConfig.java
```diff
@@ -0,0 +1,115 @@
...
+     * The default expansion depth for models (set to -1 completely hide the models).
+     */
+    @ConfigItem
+    OptionalInt defaultModelsExpandDepth;
+
+    /**
+     * The default expansion depth for the model on the model-example section.
+     */
+    @ConfigItem
+    OptionalInt defaultModelExpandDepth;
```
gsmet: I find it a bit confusing to have an option with an s and one without. It's like that in Swagger UI?

extensions/swagger-ui/deployment/src/main/java/io/quarkus/swaggerui/deployment/SwaggerUiConfig.java
```diff
@@ -0,0 +1,115 @@
...
+    Optional<Boolean> deepLinking;
+
+    /**
+     * The default expansion depth for models (set to -1 completely hide the models).
+     */
+    @ConfigItem
+    OptionalInt defaultModelsExpandDepth;
+
+    /**
+     * The default expansion depth for the model on the model-example section.
```
gsmet:
```suggestion
     * The default expansion depth for the model in the model-example section.
```

extensions/swagger-ui/deployment/src/main/java/io/quarkus/swaggerui/deployment/SwaggerUiConfig.java
```diff
@@ -0,0 +1,115 @@
...
+    Optional<Boolean> showExtensions;
+
+    /**
+     * Controls the display of extensions
+     */
+    @ConfigItem
+    Optional<Boolean> showCommonExtensions;
+
+    /**
+     * The url pointing to API definition (normally swagger.json/swagger.yaml/openapi.json/openapi.yaml).
```
gsmet: I'm not entirely sure we want to expose that. Or at least it should be clear that the default points to the URL generated by the OpenAPI extension.

extensions/swagger-ui/runtime/src/main/java/io/quarkus/swaggerui/runtime/SwaggerUiRecorder.java
```diff
@@ -26,9 +27,8 @@
             @Override
             public void handle(RoutingContext event) {
                 if (event.normalisedPath().length() == swaggerUiPath.length()) {
-
                     event.response().setStatusCode(302);
-                    event.response().headers().set(HttpHeaders.LOCATION, swaggerUiPath + "/");
+                    event.response().headers().set(HttpHeaders.LOCATION, swaggerUiPath + "/" + swaggerUiParams);
```
gsmet: If I understand correctly what you did, you only redirect with the parameters if someone went to `/swagger-ui` and then it's redirected to `/swagger-ui/?...options...`. What if the user goes to `/swagger-ui/` directly? Does it work?

gsmet (changes requested):
Thanks. I added a couple of suggestions and questions.

extensions/swagger-ui/deployment/src/main/java/io/quarkus/swaggerui/deployment/SwaggerUiConfig.java
aoudiamoncef: Me too, i just copied the original configuration properties names.

extensions/swagger-ui/deployment/src/main/java/io/quarkus/swaggerui/deployment/SwaggerUiConfig.java
aoudiamoncef: We could improve the documentation.

extensions/swagger-ui/runtime/src/main/java/io/quarkus/swaggerui/runtime/SwaggerUiRecorder.java
aoudiamoncef: The response is: i tried to find a way to handle ```/swagger-ui/?``` but i failed. I need a help to find a way to fix this.

machi1990: Maybe the re-direction rule can be based on the presence or absence of configuration query params? If not, redirect using to the same request route with configuration query params. 

PS: This is a wild guess, I am not near my PC to check whether this could actually work.

aoudiamoncef: Hi, we should take in consideration the fact that the user could override this parameters at runtime. This is just a start-up configuration. 

```java
            Handler<RoutingContext> handler = recorder
                    .handler(SWAGGER_UI_FINAL_DESTINATION, httpRootPathBuildItem.adjustPath(swaggerUiConfig.path),
                            getSwaggerUiParams(swaggerUiConfig));
            routes.produce(new RouteBuildItem(swaggerUiConfig.path, handler));
            routes.produce(new RouteBuildItem(swaggerUiConfig.path + "/*", handler));
```
When i try to redirect ```swgger-ui/``` like this for example: 
```java
                } else if (event.normalisedPath().length() == swaggerUiPath.length() + 1) {
                    if (event.normalisedPath().endsWith("/")) {
                        event.response().setStatusCode(302);
                        event.response().headers().set(HttpHeaders.LOCATION, swaggerUiPath + swaggerUiParams);
                        event.response().end();
                        return;
                    }
                    event.reroute(swaggerUiPath + "/index.html");
                    return;
                }
```
Nothing happens. 

Notice that i haven't worked with Vert.x before.

gsmet:
My opinion on this one is that adding things to the URL won't work very well.

I would lean towards trying to find another solution.

Postremus:
@aoudiamoncef Do you need any help on this pr? I kinda need this feature.

aoudiamoncef:
I'm agree with @gsmet, we should be able to inject swagger ui configuration at build time.
We are using the pre-packaged webjar of swagger, we couldn't inject a swagger.yml configuration file which could be the best solution.

Postremus:
I think we are still able to do this, or at least something similar.
The unpackaged swagger-ui is inside "META-INF/swagger-ui-files". Would it be enough to put a generated swagger.yml in this directory?

Alternativly, we could also "just" generate a custom index.html, and configure the swagger-ui this way.

aoudiamoncef:
@Postremus, we should look at how they inject the [config file](https://github.com/swagger-api/swagger-ui/blob/master/swagger-config.yaml).

Postremus:
Ah, I undertstand. The swagger.yml config is staticly included into "swagger-ui.js" during the compilation of the ui.

Another option:
We could generate an index.html, where we configure swagger-ui to load the configuration from a custom configUrl.
We could then just generate a swagger.yml config file at build time, and serve it with quarkus.
The configUrl would point to this static config file.

WDYT?

aoudiamoncef:
I think that keeping the config properties and inject them inside the html file could be the easiest way to do it.

gsmet:
@aoudiamoncef any chance you could revisit this by pushing the configuration directly by adjusting the `index.html`?

It would be similar to what we do here: https://github.com/quarkusio/quarkus/blob/master/extensions/swagger-ui/deployment/src/main/java/io/quarkus/swaggerui/deployment/SwaggerUiProcessor.java#L189 . Except we would push all the configuration properties there.

IMHO, this would be far more future proof.

gsmet:
Just a note that you will need to be a bit careful about escaping :).

aoudiamoncef:
I'm working on pushing configuration to `index.html`, my idea is to write a Java RegEx to get: 
```javascript
 const ui = SwaggerUIBundle({ // start with this line
        url: "https://petstore.swagger.io/v2/swagger.json",
        dom_id: '#swagger-ui',
        deepLinking: true,
        presets: [
          SwaggerUIBundle.presets.apis,
          SwaggerUIStandalonePreset
        ],
        plugins: [
          SwaggerUIBundle.plugins.DownloadUrl
        ],
        layout: "StandaloneLayout"
      }) // and ends with this one
```
Then i'll write a kind of string template and inject the configuration Map. Anyone have a better approach ?

gsmet:
Yeah, I don't think it would be too hard. Just catch `SwaggerUIBundle\\(({[^}]+)}` with a multiline regexp and it would be safe enough.

With this approach, we could also fix https://github.com/quarkusio/quarkus/issues/4766 (see my last comment).

gsmet:
Additional though: `jackson-databind` is a dependency of `smallrye-openapi`. Considering there's a good chance Swagger UI will always be used with openapi, I think you could add the `quarkus-jackson` extension to handle the JSON generation in the Swagger UI extension.

aoudiamoncef:
I'm using this RegEx `(SwaggerUIBundle\()(.*)(\))` with Dotall and i got:
image
The idea is to be able to generate a Javascript object using `jackson`: 
```java
ObjectMapper mapper = new ObjectMapper()
        .disable(JsonGenerator.Feature.QUOTE_FIELD_NAMES)
        .enable(SerializationFeature.INDENT_OUTPUT);
```
We have to replace the group(2) .

gsmet:
@aoudiamoncef are you making progress? Do you need any help?

aoudiamoncef:
@gsmet

I'm still working on it but with another approach. It's not a big deal to create a string and replace the matched group. 

I'm trying to validate the configurations to fail at build time if the configurations are wrongs.

Handling Authentication need more work and we should tests to avoid regressions in the future.

aoudiamoncef:
Ho @gsmet,
Updates:
Inject available Swagger UI settings: done
Inject Oauth2 settings: done 
Integration test: done (it works as expected on openapi swagger quickstart)
Add Swagger group handling: work in progress(took me more time than i expected).

I'll push my work ASAP

gsmet:
Cool! Thanks for the heads up.

aoudiamoncef:
Hi @gsmet,
I'm waiting for feedbacks to improve the code(it works but...)  and try to pretty print all the generated configuration.

extensions/swagger-ui/deployment/pom.xml
```diff
@@ -30,6 +30,10 @@
             <groupId>io.quarkus</groupId>
             <artifactId>quarkus-smallrye-openapi-common-deployment</artifactId>
         </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-jackson</artifactId>
```
gsmet, line 35: You need to include the `-deployment` dependency here and the runtime one in the runtime pom.

extensions/swagger-ui/deployment/src/main/java/io/quarkus/swaggerui/deployment/SwaggerUiProcessor.java
```diff
@@ -186,47 +204,70 @@ private void extractSwaggerUi(ResolvedArtifact artifact, Path resourceDir) throw
...
+    }
+
+    private void initObjectMapper() {
+        objectMapper = JsonMapper.builder()
+                .serializationInclusion(JsonInclude.Include.NON_NULL)
+                .serializationInclusion(JsonInclude.Include.NON_ABSENT)
+                .disable(JsonWriteFeature.QUOTE_FIELD_NAMES)
+                .enable(SerializationFeature.INDENT_OUTPUT)
+                .addModule(new Jdk8Module())
+                .build();
```
gsmet, line 270: Could we initialize this one statically?

extensions/swagger-ui/deployment/src/main/java/io/quarkus/swaggerui/deployment/SwaggerUiProcessor.java
```diff
@@ -186,47 +204,70 @@ private void extractSwaggerUi(ResolvedArtifact artifact, Path resourceDir) throw
...
-         * If Swagger UI should be enabled. By default, Swagger UI is enabled.
-         */
-        @ConfigItem(defaultValue = "true")
-        boolean enable;
+    private void updateApiUrl(String openApiPath) {
+        if (!StringUtils.isEmpty(openApiPath)) {
+            final Optional<URI> openApiUri = Optional.of(URI.create(openApiPath));
+            if (!swaggerUiConfig.url.isPresent()) {
+                swaggerUiConfig.url = openApiUri;
+            }
```
gsmet, line 240: This is not a good pattern. You shouldn't update the config. Better get the method to return you whatever you want and then use it.

extensions/swagger-ui/deployment/src/main/resources/application-custom-config.properties
```diff
@@ -0,0 +1,33 @@
+quarkus.swagger-ui=true
```
gsmet, line 1: shouldn't it be `.enable`?

extensions/swagger-ui/deployment/src/main/resources/application-custom-config.properties
```diff
@@ -0,0 +1,33 @@
...
+quarkus.swagger-ui.path=/custom
+quarkus.swagger-ui.show-common-extensions=true
+quarkus.swagger-ui.show-extensions=true
+quarkus.swagger-ui.supported-submit-methods=OPTIONS, GET, HEAD, POST, PUT, DELETE, TRACE, CONNECT, PATCH, OTHER
+quarkus.swagger-ui.tags-sorter=ALPHA
+quarkus.swagger-ui.urls-primary-name=/openapi?group=test
+quarkus.swagger-ui.urls=/openapi?group=test, /openapi?group=lahzouz
+quarkus.swagger-ui.validator-url=https://validator.swagger.io/validator
+
+quarkus.swagger-ui.oauth=true
```
gsmet, line 25: Same here?

extensions/swagger-ui/runtime/src/main/java/io/quarkus/swaggerui/runtime/SwaggerUiRecorder.java
```diff
@@ -13,31 +11,21 @@
...
+            if (event.normalisedPath().length() == swaggerUiPath.length()) {
+                event.response().setStatusCode(302);
+                event.response().headers().set(HttpHeaders.LOCATION, swaggerUiPath + "/");
+                event.response().end();
+                return;
+            } else if (event.normalisedPath().length() == swaggerUiPath.length() + 1) {
+                event.reroute(swaggerUiPath + "/index.html");
+                return;
             }
+            handler.handle(event);
```
gsmet, line 28: Please don't rewrite code like that as I have absolutely no idea of what you changed.

There is a good reason we don't use lambdas in runtime code: they have a memory cost that we want to avoid.

extensions/swagger-ui/deployment/src/main/java/io/quarkus/swaggerui/deployment/SwaggerUiConfig.java
```diff
@@ -0,0 +1,409 @@
...
+    /**
+     * Set a different validator URL, for example for locally deployed validators.
+     */
+    @ConfigItem
+    Optional<URI> validatorUrl;
+
+    /**
+     * OAuth redirect URL.
+     */
+    @ConfigItem(defaultValue = "/swagger-ui/oauth2-redirect.html")
```
gsmet, line 182: I think you cannot define the defaultValue here. You need to make it optional and create it from the path if not defined as the path is customizable.

Also I would make it a String not a URI.

gsmet (changes requested):
Overall, very nice and thorough work, thanks a lot for that!

Could you have a look at my comments?

(sorry for the late review)

extensions/swagger-ui/runtime/src/main/java/io/quarkus/swaggerui/runtime/SwaggerUiRecorder.java
aoudiamoncef: I got it :)

extensions/swagger-ui/deployment/src/main/java/io/quarkus/swaggerui/deployment/SwaggerUiConfig.java
aoudiamoncef: @gsmet , would ou transfom all URI to String ?

aoudiamoncef:
I'll fix this very quickly

aoudiamoncef:
@gsmet, i did the requested changes. I'm waiting for your feedbacks.

gsmet:
Hi Moncef,

Sorry it took me so long to get back to you.

The PR seems broken, I had to change this in the test:
```
public class CustomConfigTest {

    @RegisterExtension
    static final QuarkusUnitTest config = new QuarkusUnitTest()
            .withConfigurationResource("application-custom-config.properties");
```
but then the tests are failing anyway.

I think it won't fly to generate the URIs in the config as you somehow need to use `httpRootPathBuildItem.adjustPath()` on the paths as we do it in the processor. I would probably move all the JSON object building logic to a proper method in the processor.

Then there are other minor things that could be improved but I think we need that to be addressed first.

aoudiamoncef:
Hi @gsmet,
I'll update the PR ASAP. 
I remember @phillip-kruger proposal to fuse SwaggeUI extension with OpenApi one. It could make sense with GROUPS, OpenApi extension have to know about this, in order to create supplied GROUPS.

Thanks

phillip-kruger:
When I suggested that we merge the two, there was concerns that maybe we want to add support for other UIs. I still think we can merge and add support for other UIs via config, as there will not be many (if ever more than one or two). This will make things easier to maintain I think.

wtrocki:
Maybe other editors could be a separate issue once this one is merged?
Things like Apicurio/Apicurito will be an amazing addition as they can support other schemas.

yntelectual:
Hi guys, what is the state of this MR? We would love to get support for swagger-ui showing multiple openapi files (dropdown in header) for which we need to be able to specify `urls` attribute of the Swagger ui config. 
```
{url: "<url1>", name: "<name1>"},{url: "<url2>", name: "<name2>"}
```

I understand that this MR would allow that among other things.
thanks[[m56

phillip-kruger:
Hi all.
I am looking to get this into smallrye. So far the plan is to allow creating a index html using a basic template and some options. w.d.y.t ?

yntelectual:
any solution that would allow customization of the swagger config options is fine. Right now there is no way how to customize it(I tried adding a webfilter that would modify the html response, but the swagger-ui is server via vertx...). Not sure how long the integration w smallrye would take. We will probably create a custom extension (based on this MR)in the meantime as we need it rather urgently.  Thanks for the effort, will keep an eye on this.

aoudiamoncef:
Hi all,
I'll try to finish this PR ASAP.
 I'm sorry but with my current daily work, I haven't a lot of free time to work on open source projects.

phillip-kruger:
@aoudiamoncef i have added this to SmallRye and will pull that into Quarkus soon.

phillip-kruger:
#12782 will supersede and close this.

PR #6315, opened 2019-12-23
Exclude integration-tests module if skipITs is set

gastaldi:
This excludes the `integration-tests` module if `-DskipITs` is specified during the build, making the build faster

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -63,9 +63,6 @@
         <!-- relying on the fact that the extension descriptor, generated as part of an extension build, is available -->
         <module>devtools</module>
 
-        <!-- Integration Tests -->
-        <module>integration-tests</module>
-
         <!-- Misc. -->
         <module>docs</module>
     </modules>
@@ -289,6 +286,18 @@
                     </plugin>
                 </plugins>
             </build>
+        </profile>
+        <profile>
+            <id>integration-tests</id>
+            <activation>
+                <property>
+                    <name>!skipITs</name>
+                </property>
+            </activation>
+            <modules>
+                <!-- Integration Tests -->
+                <module>integration-tests</module>
+            </modules>
         </profile>
         <profile>
             <id>tcks</id>


oscerd (approved):

gsmet (changes requested):
I'm not convinced by this one: even if you don't run the ITs, it's good to know everything compiles.

I'm pretty sure it happened to me a few times that the ITs didn't compile anymore and I had to fix them.

Let's wait for some more opinions. @geoand WDYT?

geoand:
I am very much on the fence about this one, but I'm probably 60 - 40 in favor of what @gsmet says. I would personally leave it as it is

gastaldi:
>even if you don't run the ITs, it's good to know everything compiles

In this case you don't specify `-DskipITs`. If you still want to compile the ITs but don't want to run them you just specify `-DskipTests`

gsmet:
AFAIK, `-DskipTests` still runs the ITs. At least that was the case at the beginning of the project.

To skip running the ITs, you needed `-DskipITs`.

In any case, it's very different not running something and potentially have compilation errors.

I'm really not about having a common setup allowing for compilation errors. That will require unnecessary CI runs and we certainly don't want that. I think you will waste less time just running the compilation of everything.

geoand:
One more thing to consider perhaps: Do IDEs properly recognize the `!....` syntax, or would this make it harder to get good IDE warnings for the integration tests?
I am pretty sure IntelliJ works as expected, but I have no idea about the rest

gsmet:
I decided to not merge that one.

If we had unlimited CI resources, we could consider it but resources are scarce right now. We can reconsider if one day we get sufficient CI resources so don't drop the branch and keep it around :).

gastaldi:
>If we had unlimited CI resources, we could consider it but resources are
scarce right now

Not sure how that relates to this PR but that's fine ;)

gsmet:
I explained it here:

> I'm really not about having a common setup allowing for compilation errors. That will require unnecessary CI runs and we certainly don't want that. I think you will waste less time just running the compilation of everything.

Basically, if you have a compilation issue in the IT, you will have a CI run for ~40 mn until it gets to the compilation error.

I think it's better to waste ~1mn on your box in the current situation.

PR #6325, opened 2019-12-23
Ensure that Kotlin Data classes with default values work in native with JAX-RS

### LINKED ISSUE #3954: Using Jackson support for Kotlin fails in native mode.
**Describe the bug**
When using the Jackson Kotlin module with a Kotlin data class, it fails in native mode with the following:

```
com.fasterxml.jackson.databind.exc.InvalidDefinitionException: Cannot construct instance of `com.capitalone.polyglot.quarkus.model.Address` (no Creators, like default construct, exist): cannot deserialize from Object value (no delegate- or property-based Creator)
```

The example code is pretty straight forward:

```
@ApplicationScoped
class CustomObjectMapperConfig {
    @Singleton
    @Produces
    fun objectMapper(): ObjectMapper {
        return ObjectMapper().registerModule(KotlinModule())
    }
}

data class Address(val addressLine1 : String, val addressLine2 : String, val addressLine3 : String, val addressLine4 : String, val city : String, val state : String, val country : String)

@Path("/deposits/tooling/validate-address")
@Consumes(MediaType.APPLICATION_JSON)
class AddressValidationResource {

    @POST
    fun validateAddress(address : Address) = address.copy(
                addressLine1 = "${address.addressLine1} ${address.addressLine2} ${address.addressLine3} ${address.addressLine4}",
                addressLine2 = "",
                addressLine3 = "",
                addressLine4 = "")
}
```

**Expected behavior**
(Describe the expected behavior clearly and concisely.)
If I run this in JVM mode, aka `mvn compile quarkus:dev`, or as a runnable jar it works just fine.

**Actual behavior**

**To Reproduce**
1. 
2. 
3. 

**Configuration**
```properties

```

**Screenshots**

 - Output of `java -version`: 
```
openjdk version "1.8.0_212"
OpenJDK Runtime Environment (build 1.8.0_212-20190523183630.graal2.jdk8u-src-tar-gz-b03)
OpenJDK 64-Bit GraalVM CE 19.1.0 (build 25.212-b03-jvmci-20-b04, mixed mode)
```
 - Quarkus version or git rev: 
`<quarkus.version>0.21.2</quarkus.version>`

**Additional context**
Related #3652

geoand:
…th JAX-RS

Fixes: #3954

### DIFF
--- a/core/deployment/src/main/java/io/quarkus/deployment/steps/ReflectiveHierarchyStep.java
+++ b/core/deployment/src/main/java/io/quarkus/deployment/steps/ReflectiveHierarchyStep.java
@@ -28,6 +28,7 @@ import io.quarkus.deployment.annotations.BuildProducer;
 import io.quarkus.deployment.annotations.BuildStep;
 import io.quarkus.deployment.builditem.CombinedIndexBuildItem;
 import io.quarkus.deployment.builditem.nativeimage.ReflectiveClassBuildItem;
+import io.quarkus.deployment.builditem.nativeimage.ReflectiveClassFinalFieldsWritablePredicateBuildItem;
 import io.quarkus.deployment.builditem.nativeimage.ReflectiveHierarchyBuildItem;
 import io.quarkus.deployment.builditem.nativeimage.ReflectiveHierarchyIgnoreWarningBuildItem;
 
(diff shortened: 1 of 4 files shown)


geoand:
I can of course add an integration test if we think it's valuable

core/deployment/src/main/java/io/quarkus/deployment/steps/ReflectiveHierarchyStep.java
```diff
@@ -144,4 +157,21 @@ private void addClassTypeHierarchy(ReflectiveHierarchyBuildItem reflectiveHierar
...
+        if (finalFieldsWritablePredicates.isEmpty()) {
+            return false;
+        }
+        if (finalFieldsWritablePredicate == null) {
+            // create a predicate that returns true if any of the predicates says that final fields need to be writable
+            finalFieldsWritablePredicate = finalFieldsWritablePredicates
+                    .stream()
+                    .map(ReflectiveClassFinalFieldsWritablePredicateBuildItem::getPredicate)
+                    .reduce(c -> false, Predicate::or);
+        }
```
gsmet: I don't think I would have done it this way. Can't we build the predicate once where we need it and pass it as a parameter?

gsmet (changes requested):
I added a suggestion.

core/deployment/src/main/java/io/quarkus/deployment/steps/ReflectiveHierarchyStep.java
geoand: You mean instead of stuffing the result in the `private static` field?

gsmet: Yes.

geoand: Good idea, fixed

geoand:
@gsmet let me know if you think that an integration test should be added.

Thanks

gsmet (approved):

gsmet:
I marked it for backport. I'm not totally sure I will but let's consider it when we prepare 1.1.1.

geoand:
Makes sense, thanks!

PR #6326, opened 2019-12-23
Enlist repository parameterized types for reflection

### LINKED ISSUE #6324: MongoDB with Panache didn't enlist for reflection Repository parameterized type
**Describe the bug**

When using the repository aproach and resteasy with only `Response` return type, native mode miss some reflection configuration.

Imagine the following classes

```
public class NeedReflection {
    public String comment;
}

@ApplicationScoped
public class NeedReflectionRepository implements PanacheMongoRepository<NeedReflection> {
}

@Path("/bugs")
@Consumes(MediaType.TEXT_PLAIN)
public class BugResource {

    @Inject
    NeedReflectionRepository needReflectionRepository;

    @GET
    @Path("reflect")
    public Response testNeedReflection() {
        return Response.ok(needReflectionRepository.listAll()).build();
    }
}
```

Calling /bugs/reflect will leads to the exception `org.bson.codecs.configuration.CodecConfigurationException: Can't find a codec for class io.quarkus.it.mongodb.panache.bugs.NeedReflection` that is fixed via the usage of `@RegisterForReflection`.

The entity aproach didn't have the same issue.

**Expected behavior**
We shouldn't need to add `@RegisterForReflection` to a repository entity class, the extension should enlist for reflection all repository parameterized type classes.

 - Output of `java -version`: Java 11
 - GraalVM version (if different from Java): 19.3.0
 - Quarkus version or git rev: 1.1.0 or current master

loicmathieu:
Fixes #6324 

Enlist for relfection types parameters from interfaces and super classes.
This fixes the issue with the repository implementation where the repository type parameter (it should be the MongoDB persisted class) is not registered for reflection so it can fails if not direct usage is made that GraalVM can be aware of.

I didn't climb up the hierarchy of interfaces and super classes so it's still possible to have relfection issue for complex type hierarchies. I don't now if it worth it to implement it.

### DIFF
--- a/extensions/panache/mongodb-panache/deployment/src/main/java/io/quarkus/mongodb/panache/deployment/PanacheResourceProcessor.java
+++ b/extensions/panache/mongodb-panache/deployment/src/main/java/io/quarkus/mongodb/panache/deployment/PanacheResourceProcessor.java
@@ -1,12 +1,15 @@
 package io.quarkus.mongodb.panache.deployment;
 
+import java.util.ArrayList;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Set;
 
 import org.bson.types.ObjectId;
 import org.jboss.jandex.ClassInfo;
 import org.jboss.jandex.CompositeIndex;
 import org.jboss.jandex.DotName;
+import org.jboss.jandex.IndexView;
 import org.jboss.jandex.Indexer;
 import org.jboss.jandex.Type;
 
@@ -18,6 +21,7 @@ import io.quarkus.deployment.builditem.BytecodeTransformerBuildItem;
 import io.quarkus.deployment.builditem.CapabilityBuildItem;
 import io.quarkus.deployment.builditem.CombinedIndexBuildItem;
 import io.quarkus.deployment.builditem.FeatureBuildItem;
+import io.quarkus.deployment.builditem.nativeimage.ReflectiveClassBuildItem;
 import io.quarkus.deployment.builditem.nativeimage.ReflectiveHierarchyBuildItem;
 import io.quarkus.deployment.index.IndexingUtil;
 import io.quarkus.jackson.spi.JacksonModuleBuildItem;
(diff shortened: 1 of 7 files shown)


gsmet (dismissed):
Can you fix the commit message? `relfection`. Thanks.

loicmathieu:
@gsmet commit message updated

gsmet (dismissed):
We need to take into account hierarchy of superclasses. It happens and dealing with the superclass only is clearly not sufficient.

Or did I miss something?

machi1990:
> We need to take into account hierarchy of superclasses. It happens and dealing with the superclass only is clearly not sufficient.
> 
> Or did I miss something?

+1  with @gsmet here

loicmathieu:
@gsmet hierarchy of super classes where the direct super classes is not the parameterized one seems to be not very common but I can handle it.
But this will need to wait for a few days ...

loicmathieu:
@gsmet @machi1990 I climb up the hierarchy of super types in the PR.

gsmet:
I will let @FroMage check that one and decide of its fate :).

FroMage:
I don't understand this PR. The test you added is for when an entity type is never registered via reflection. This can only happen for Mongo, right? And only because you don't have an equivalent to `@Entity` that ORM uses to register those types. So for Mongo they become "unknown" unless they extend `MongoPanacheEntity`.

So, perhaps you should have a `@MongoEntity` annotation?

Alternately, what you want to do is use `JandexUtil.resolveTypeParameters(DotName.createSimple(userRepoSubclassName), DotName.createSimple(MongoPanacheRepositoryBase.class.getName()), index)` and get the right type parameter that points to the entity.

And I think you want to do that only for mongo, so not in panache-common.

loicmathieu:
@FroMage yes, when you define "mongodb entities" it's not mandatory to add the `@MongoEntity` annotation that only exist to define the database name (if not comming from application.properties) and the collection name (by default the name of the class).

We can ask to always add the annotation, but as it's not enforced inside the framework, situations will always exist where a user will not add it ...

I will refactory with `JandexUtil.resolveTypeParameters()`.

FroMage:
> We can ask to always add the annotation, but as it's not enforced inside the framework, situations will always exist where a user will not add it ...

Well, you can enforce it ;)

It would be more consistent with ORM, but on the other hand this appears to be the first use-case for requiring it, and we can solve it by scanning type params, so probably we can keep that strategy for now and keep an eye out for new use-cases that would require that annotation.

loicmathieu:
@FroMage updated with the usage of `JandexUtil.resolveTypeParameters()`

FroMage (approved):
Look @geoand we're all using this method now :)

geoand:
Excellent, you all know who to direct your rage if things don't work :P

loicmathieu:
@gsmet ready to be merged :)

loicmathieu:
There is an SO question about it: https://stackoverflow.com/questions/59722717/problem-with-native-image-in-quarkus-and-mongopanache

geoand:
Let's merge

PR #6332, opened 2019-12-23
Fix YAML profile example

lordofthejars:


lordofthejars:
I close the PR as the fix is not complete

PR #6301, opened 2019-12-21
Introduce a deterministic ordering, between extensions, for database schema updates

jaikiran:
Fixes https://github.com/quarkusio/quarkus/issues/5942

Sorry this took a bit longer than expected since I had to spend some time to make sure I could absolutely reproduce this issue in an automated testcase before verifying the fix. I'm now satisfied that I can verify this fix with some minor changes, that are included in this PR, to the existing `FlywayFunctionalityTest` test.

PR #6334, opened 2019-12-23
Liquibase extension

andrejpetras:
Liquibase extension

- Add the liquibase-core and snakeyml (it is used for liquibase json, yaml format) dependencies.
- Because Liquibase object requires a JDBC connection I created a QLiquibase wrapper with datasource instance and common liquibase methods to open and close connection after each liquibase method.
- Multiple datasources (Datasoruce annotation) are supported (like flyway)
- Integration tests are for xml, json, yaml and sql format
- Liquibase has own implementation of the class-path scanner to load the Services (Interface implementation). For this I generated txt files (list of implementation classes) at build time which are used in the native run time.

I am really not sure if the QLiquibase (wrapper class) is the right name.

Update:
I renamed the QLiquibase class in Liquibase
Add the org.osgi dependency


extensions/liquibase/deployment/src/main/java/io/quarkus/liquibase/LiquibaseProcessor.java
```diff
@@ -0,0 +1,319 @@
...
+                new NativeImageResourceBuildItem(getChangeLogs(dataSourceInitializedBuildItem).toArray(new String[0])));
+
+        // liquibase XSD
+        resourceProducer.produce(new NativeImageResourceBuildItem(
+                "www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.5.xsd",
+                "www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.6.xsd",
+                "www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.7.xsd",
+                "www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.8.xsd",
+                "www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd",
+                "www.liquibase.org/xml/ns/pro/liquibase-pro-3.8.xsd",
```
gsmet: Can you check if `NativeImageResourceBuildItem` accepts regexps? I would be more permissive here to avoid having to update that in the future.

extensions/liquibase/deployment/src/main/java/io/quarkus/liquibase/LiquibaseProcessor.java
```diff
@@ -0,0 +1,319 @@
...
+    @BuildStep
+    ServiceStartBuildItem configureRuntimeProperties(LiquibaseRecorder recorder,
+            LiquibaseRuntimeConfig liquibaseRuntimeConfig,
+            BeanContainerBuildItem beanContainer) {
+        try {
+            recorder.configureLiquibaseProperties(liquibaseRuntimeConfig, beanContainer.getValue());
+            recorder.doStartActions(liquibaseRuntimeConfig, beanContainer.getValue());
+            return new ServiceStartBuildItem("liquibase");
+        } catch (Throwable t) {
+            t.printStackTrace();
```
gsmet: Hmmm? We definitely don't want to use that.

extensions/liquibase/deployment/src/main/java/io/quarkus/liquibase/LiquibaseProcessor.java
```diff
@@ -0,0 +1,319 @@
...
+            throw t;
+        }
+    }
+
+    /**
+     * Adds the liquibase services
```
gsmet: Same here, useless javadoc.

extensions/liquibase/deployment/src/main/java/io/quarkus/liquibase/LiquibaseProcessor.java
```diff
@@ -0,0 +1,319 @@
...
+     * @param liquibaseRuntimeConfig The Liquibase configuration
+     */
+    @Record(ExecutionTime.RUNTIME_INIT)
+    @BuildStep
+    ServiceStartBuildItem configureRuntimeProperties(LiquibaseRecorder recorder,
+            LiquibaseRuntimeConfig liquibaseRuntimeConfig,
+            BeanContainerBuildItem beanContainer) {
+        try {
+            recorder.configureLiquibaseProperties(liquibaseRuntimeConfig, beanContainer.getValue());
+            recorder.doStartActions(liquibaseRuntimeConfig, beanContainer.getValue());
```
gsmet: You will need something similar to https://github.com/quarkusio/quarkus/pull/6301/files#diff-1fdde7b94eca712fd8aa08fe51e38c39R111 . It was just merged into the Flyway extension, that's normal you missed it.

extensions/liquibase/runtime/pom.xml
```diff
@@ -0,0 +1,69 @@
...
+    <parent>
+        <artifactId>quarkus-liquibase-parent</artifactId>
+        <groupId>io.quarkus</groupId>
+        <version>999-SNAPSHOT</version>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>quarkus-liquibase</artifactId>
+    <name>Quarkus - Liquibase - Runtime</name>
+    <description>Handle your database schema migrations</description>
```
gsmet: Maybe add `with Liquibase` at the end?

extensions/liquibase/runtime/src/main/java/io/quarkus/liquibase/QLiquibase.java
```diff
@@ -0,0 +1,932 @@
...
+import liquibase.diff.compare.CompareControl;
+import liquibase.exception.DatabaseException;
+import liquibase.exception.LiquibaseException;
+import liquibase.lockservice.DatabaseChangeLogLock;
+import liquibase.resource.ResourceAccessor;
+
+/**
+ * The quarkus liquibase bean
+ */
+public class QLiquibase {
```
gsmet: Why `Q`? It doesn't sound very natural to me. Especially for an API.

Why do we need this wrapper? In the Flyway extension, we inject the `Flyway` instance which seems more future proof.

extensions/liquibase/runtime/src/main/java/io/quarkus/liquibase/runtime/LiquibaseConfig.java
```diff
@@ -0,0 +1,80 @@
...
+import static io.quarkus.liquibase.runtime.LiquibaseDataSourceRuntimeConfig.DEFAULT_LOCK_TABLE;
+import static io.quarkus.liquibase.runtime.LiquibaseDataSourceRuntimeConfig.DEFAULT_LOG_TABLE;
+
+import java.util.List;
+import java.util.Optional;
+
+/**
+ * The liquibase configuration
+ */
+public class LiquibaseConfig {
```
gsmet: If this is a Quarkus config class, the annotations are missing.

extensions/liquibase/runtime/src/main/java/io/quarkus/liquibase/runtime/graal/LiquibaseServiceLoader.java
```diff
@@ -0,0 +1,66 @@
...
+import java.util.List;
+import java.util.Objects;
+import java.util.stream.Collectors;
+
+import org.jboss.logging.Logger;
+
+/**
+ * The liquibase service class loader.
+ */
+public class LiquibaseServiceLoader {
```
gsmet: Why do we need that one?

extensions/liquibase/runtime/src/main/java/io/quarkus/liquibase/runtime/graal/SubstituteKp.java
```diff
@@ -0,0 +1,16 @@
+package io.quarkus.liquibase.runtime.graal;
+
+import com.oracle.svm.core.annotate.Substitute;
+import com.oracle.svm.core.annotate.TargetClass;
+
+/**
+ * The liquibase license validation substitute
```
gsmet: So this comment is useless. Explain why you have a substitution instead.

extensions/liquibase/runtime/src/main/java/io/quarkus/liquibase/runtime/graal/SubstituteServiceLocator.java
```diff
@@ -0,0 +1,19 @@
+package io.quarkus.liquibase.runtime.graal;
+
+import java.util.List;
+
+import com.oracle.svm.core.annotate.Substitute;
+import com.oracle.svm.core.annotate.TargetClass;
+
+/**
+ * The liquibase service locator substitute
```
gsmet: Same here.

integration-tests/pom.xml
```diff
@@ -53,6 +53,7 @@
         <module>elytron-resteasy</module>
         <module>elytron-undertow</module>
         <module>flyway</module>
+        <module>liquibase</module>
         <module>oidc</module>
         <module>oidc-code-flow</module>
         <module>keycloak-authorization</module>
```
gsmet: Note: we also need to add the native tests to the pipeline here: https://github.com/quarkusio/quarkus/blob/master/ci-templates/stages.yml#L187 .

gsmet (changes requested):
Woot! Thanks for this work, very cool.

I added some suggestions and questions.

extensions/liquibase/deployment/src/main/java/io/quarkus/liquibase/LiquibaseProcessor.java
andrejpetras: I tried but it doesn't work for me.
```
        resourceProducer.produce(new NativeImageResourceBuildItem(
                ".*properties$",
                "www.liquibase.org/xml/ns/.*$"));
```
The native image of the liquibase has this configuration
https://github.com/1000kit/liquibase/blob/9ebb46978344aab2f4a531e61ec6c7d20a656fbc/src/main/resources/META-INF/native-image/org.tkit.liquibase/liquibase/resource-config.json#L9

extensions/liquibase/runtime/src/main/java/io/quarkus/liquibase/QLiquibase.java
andrejpetras: I am really not sure if the QLiquibase (wrapper class) is the right name. 

I started like the Flyway:
- The Liquibase is expecting open JDBC connection in the constructor (it reads the info about the database from the connection). In the case of Flyway we have Datasource as input parameter
- The Liquibase does not open or close the connection
- I tried to extend the Liquibase object and overwrite the method to open and close the connection but there are final methods like dropAll()

They have CDI integration: https://github.com/liquibase/liquibase/blob/628a95e7344ad2bc108746fd3491e3957879668e/liquibase-cdi/src/main/java/liquibase/integration/cdi/CDILiquibase.java#L130

If we use the Liquibase object in the producer we can open connection and the developer needs to close the connection.
If we use the extended Liquibase implementation in the producer we need to mark the final methods which do not close the connection. 
If we use the wrapper (with correct name) we can open and close the connection for the developer for each method but we have much more code.

The QLiquibase also contains the ```LiquibaseConfig``` object which is merge of the run-time configuration and build configuration.

I am open to any solution

extensions/liquibase/runtime/src/main/java/io/quarkus/liquibase/runtime/LiquibaseConfig.java
andrejpetras: The QLiquibase contains the LiquibaseConfig object which is merge of the run-time configuration and build configuration. It could be replaced by the run-time and build configuration.

extensions/liquibase/runtime/src/main/java/io/quarkus/liquibase/runtime/graal/SubstituteKp.java
andrejpetras: Javadoc update
The liquibase license service for the commercial liquibase license throw UnsupportedFeatureException: Non-reducible loop for the liquibase.pro.packaged.aO.parseNumberText2(boolean)

extensions/liquibase/deployment/pom.xml
```diff
@@ -0,0 +1,71 @@
...
+        </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-test-h2</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.jboss.spec.javax.xml.bind</groupId>
+            <artifactId>jboss-jaxb-api_2.3_spec</artifactId>
+            <scope>test</scope>
```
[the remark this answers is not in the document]
andrejpetras: Done

extensions/liquibase/deployment/src/main/java/io/quarkus/liquibase/LiquibaseDatasourceBeanGenerator.java
```diff
@@ -0,0 +1,131 @@
...
+
+import org.jboss.jandex.AnnotationInstance;
+import org.jboss.jandex.AnnotationValue;
+import org.jboss.jandex.DotName;
+
+import io.quarkus.arc.deployment.GeneratedBeanBuildItem;
+import io.quarkus.arc.processor.DotNames;
+import io.quarkus.deployment.annotations.BuildProducer;
+import io.quarkus.deployment.util.HashUtil;
+import io.quarkus.gizmo.*;
```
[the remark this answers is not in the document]
andrejpetras: Done

extensions/liquibase/deployment/src/main/java/io/quarkus/liquibase/LiquibaseDatasourceBeanGenerator.java
```diff
@@ -0,0 +1,131 @@
...
+import io.quarkus.liquibase.runtime.LiquibaseProducer;
+
+/**
+ * Generates the CDI producer bean for {@link QLiquibase} at build time.<br>
+ * Supports multiple {@link Named} {@link DataSource}s.
```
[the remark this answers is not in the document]
andrejpetras: Done

extensions/liquibase/deployment/src/main/java/io/quarkus/liquibase/LiquibaseProcessor.java
```diff
@@ -0,0 +1,319 @@
...
+import liquibase.resource.ClassLoaderResourceAccessor;
+import liquibase.servicelocator.ServiceLocator;
+
+/**
+ * The liquibase processor
+ */
+class LiquibaseProcessor {
+
+    /**
+     * The logger for this class
```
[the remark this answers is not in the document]
andrejpetras: Done

extensions/liquibase/deployment/src/main/java/io/quarkus/liquibase/LiquibaseProcessor.java
```diff
@@ -0,0 +1,319 @@
...
+import liquibase.changelog.ChangeSet;
+import liquibase.changelog.DatabaseChangeLog;
+import liquibase.exception.LiquibaseException;
+import liquibase.parser.ChangeLogParser;
+import liquibase.parser.ChangeLogParserFactory;
+import liquibase.resource.ClassLoaderResourceAccessor;
+import liquibase.servicelocator.ServiceLocator;
+
+/**
+ * The liquibase processor
```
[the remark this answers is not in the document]
andrejpetras: Done

extensions/liquibase/runtime/src/main/java/io/quarkus/liquibase/runtime/graal/LiquibaseServiceLoader.java
andrejpetras: JavaDoc updated

extensions/liquibase/runtime/src/main/java/io/quarkus/liquibase/runtime/graal/SubstituteServiceLocator.java
andrejpetras: JavaDoc updated

integration-tests/pom.xml
andrejpetras: Done

gastaldi:
@andrejpetras looks like you need to rebase this branch, as it has some unrelated commits. Here are the steps in case you need some help:

1) `git checkout master`
2) `git pull`
3) `git checkout 5957-liquibase-extension`
4) `git rebase master`
5) `git push -f`

andrejpetras:
> @andrejpetras looks like you need to rebase this branch, as it has some unrelated commits. Here are the steps in case you need some help:
> 
> 1. `git checkout master`
> 2. `git pull`
> 3. `git checkout 5957-liquibase-extension`
> 4. `git rebase master`
> 5. `git push -f`

@gastaldi thanks for the hint. I will do that.

extensions/liquibase/runtime/src/main/java/io/quarkus/liquibase/QLiquibase.java
andrejpetras: I renamed the QLiquibase class in Liquibase

extensions/liquibase/deployment/src/main/java/io/quarkus/liquibase/LiquibaseProcessor.java
```diff
@@ -0,0 +1,319 @@
...
+     */
+    private static final ChangeLogParserFactory CHANGE_LOG_PARSER_FACTORY = ChangeLogParserFactory.getInstance();
+
+    /**
+     * Liquibase build config
+     */
+    LiquibaseBuildTimeConfig liquibaseBuildConfig;
+
+    /**
+     * Gets the capability for the liquibase extension
```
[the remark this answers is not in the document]
andrejpetras: Done

extensions/liquibase/deployment/src/main/java/io/quarkus/liquibase/LiquibaseProcessor.java
```diff
@@ -0,0 +1,319 @@
...
+    CapabilityBuildItem capability() {
+        return new CapabilityBuildItem(Capabilities.LIQUIBASE);
+    }
+
+    @BuildStep
+    void reflectionMethods(BuildProducer<ReflectiveMethodBuildItem> reflective) throws Exception {
+        Method isSplitStatements = liquibase.change.AbstractSQLChange.class.getMethod("isSplitStatements");
+        reflective.produce(new ReflectiveMethodBuildItem(isSplitStatements));
+        Method isStripComments = liquibase.change.AbstractSQLChange.class.getMethod("isStripComments");
+        reflective.produce(new ReflectiveMethodBuildItem(isStripComments));
```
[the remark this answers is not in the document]
andrejpetras: Done

extensions/liquibase/deployment/src/main/java/io/quarkus/liquibase/LiquibaseProcessor.java
```diff
@@ -0,0 +1,319 @@
...
+    void reflection(BuildProducer<ReflectiveClassBuildItem> reflective,
+            BuildProducer<NativeImageResourceBuildItem> resourceProducer,
+            BuildProducer<GeneratedResourceBuildItem> generatedResourceProducer,
+            BuildProducer<RuntimeInitializedClassBuildItem> runtimeInitialized) {
+
+        Stream.of("liquibase.util.StringUtils",
+                "liquibase.servicelocator.ServiceLocator",
+                "liquibase.diff.compare.CompareControl")
+                .map(RuntimeInitializedClassBuildItem::new)
+                .forEach(runtimeInitialized::produce);
```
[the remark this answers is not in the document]
andrejpetras: Done

extensions/liquibase/deployment/src/main/java/io/quarkus/liquibase/LiquibaseProcessor.java
```diff
@@ -0,0 +1,319 @@
...
+                liquibase.changelog.ChangeLogHistoryService.class,
+                liquibase.datatype.LiquibaseDataType.class,
+                liquibase.executor.Executor.class,
+                liquibase.lockservice.LockService.class,
+                liquibase.sqlgenerator.SqlGenerator.class)
+                .forEach(t -> addService(reflective, generatedResourceProducer, resourceProducer, t));
+    }
+
+    /**
+     * Build the static init phase of the liquibase extension
```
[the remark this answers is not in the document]
andrejpetras: Done

extensions/liquibase/deployment/src/main/java/io/quarkus/liquibase/LiquibaseProcessor.java
andrejpetras: Done

extensions/liquibase/deployment/src/main/java/io/quarkus/liquibase/LiquibaseProcessor.java
andrejpetras: Done

extensions/liquibase/deployment/src/main/java/io/quarkus/liquibase/LiquibaseProcessor.java
andrejpetras: Done

extensions/liquibase/runtime/pom.xml
andrejpetras: Done

gastaldi:
Looks like it needs to be rebased again :)

extensions/liquibase/deployment/src/main/java/io/quarkus/liquibase/LiquibaseProcessor.java
```diff
@@ -0,0 +1,298 @@
...
+
+        runtimeInitialized.produce(new RuntimeInitializedClassBuildItem("liquibase.util.StringUtils"));
+        runtimeInitialized.produce(new RuntimeInitializedClassBuildItem("liquibase.servicelocator.ServiceLocator"));
+        runtimeInitialized.produce(new RuntimeInitializedClassBuildItem("liquibase.diff.compare.CompareControl"));
+
+        //Classes that use reflection
+        reflective.produce(new ReflectiveClassBuildItem(true, false, false,
+                "com.sun.org.apache.xerces.internal.impl.dv.xs.SchemaDVFactoryImpl",
+                "com.sun.org.apache.xerces.internal.jaxp.SAXParserFactoryImpl",
+                "com.sun.org.apache.xerces.internal.jaxp.DocumentBuilderFactoryImpl"));
```
gastaldi: It seems to me that this should belong to a jaxp extension

andrejpetras: @gastaldi should I put jaxb-runtime extension as dependency?
Currently we have only this for the liquibase-runtime
```
        <dependency>
            <groupId>org.jboss.spec.javax.xml.bind</groupId>
            <artifactId>jboss-jaxb-api_2.3_spec</artifactId>
        </dependency>
```

gastaldi: Not sure if that belongs to jaxb, perhaps a separate (unlisted) jaxp extension containing this code and the jaxp deps but I'm unsure if it's worth the effort

gsmet: If you need the JAXB API, then it probably makes sense to add the JAXB extension.

bom/runtime/pom.xml
```diff
@@ -2446,7 +2454,38 @@
...
+                <version>${snakeyaml.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>org.osgi</groupId>
+                <artifactId>org.osgi.core</artifactId>
+                <version>${osgi.version}</version>
+                <exclusions>
+                    <exclusion>
+                        <groupId>*</groupId>
+                        <artifactId>*</artifactId>
```
gsmet: Does it really work? We had some bad surprises with star excludes.

extensions/liquibase/deployment/pom.xml
```diff
@@ -0,0 +1,66 @@
...
+        </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-test-h2</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-jdbc-postgresql</artifactId>
```
gsmet: Do we really need to test things with PostgreSQL? Testing with H2 is not enough?

extensions/liquibase/deployment/src/main/java/io/quarkus/liquibase/LiquibaseProcessor.java
gastaldi: These classes don't seem to belong to JAXB, but yes, if the JAXB API is used the extension should be added as @gsmet mentioned

extensions/liquibase/runtime/src/main/java/io/quarkus/liquibase/Liquibase.java
```diff
@@ -0,0 +1,932 @@
...
+     * @throws DatabaseException if the method fails
+     */
+    private liquibase.Liquibase createLiquibase(Connection con) throws DatabaseException {
+        Database database = DatabaseFactory.getInstance().findCorrectDatabaseImplementation(new JdbcConnection(con));
+        database.setDatabaseChangeLogLockTableName(config.databaseChangeLogLockTableName);
+        database.setDatabaseChangeLogTableName(config.databaseChangeLogTableName);
+        config.currentDateTimeFunction.ifPresent(database::setCurrentDateTimeFunction);
+        config.liquibaseCatalogName.ifPresent(database::setLiquibaseCatalogName);
+        config.liquibaseSchemaName.ifPresent(database::setLiquibaseSchemaName);
+        config.liquibaseTablespaceName.ifPresent(database::setLiquibaseTablespaceName);
```
gsmet: Is it a good idea to create a Liquibase instance for each operation?

I'm not familiar with Liquibase but it doesn't seem right to me. I'm not sure I see another solution though as I suppose we need to close it somehow but I thought I should raise the issue.

extensions/liquibase/runtime/src/main/java/io/quarkus/liquibase/Liquibase.java
```diff
@@ -0,0 +1,932 @@
...
+
+        config.defaultCatalogName.ifPresent(d -> {
```
gsmet: Generally, we try to avoid lambdas in runtime code as it has a non negligible memory cost.

Let's get to the bottom of this Liquibase instantiation thing before changing that though.

extensions/liquibase/runtime/src/main/java/io/quarkus/liquibase/runtime/graal/SubstituteKp.java
```diff
@@ -0,0 +1,19 @@
+package io.quarkus.liquibase.runtime.graal;
+
+import com.oracle.svm.core.annotate.Substitute;
+import com.oracle.svm.core.annotate.TargetClass;
+
+/**
+ * The liquibase license service for the commercial liquibase license
+ * throw UnsupportedFeatureException: Non-reducible loop for the liquibase.pro.packaged.aO.parseNumberText2(boolean):
```
gsmet:
```suggestion
 * throws UnsupportedFeatureException: Non-reducible loop for the liquibase.pro.packaged.aO.parseNumberText2(boolean):
```

extensions/liquibase/runtime/src/main/java/io/quarkus/liquibase/runtime/graal/SubstituteServiceLocator.java
```diff
@@ -0,0 +1,21 @@
+package io.quarkus.liquibase.runtime.graal;
+
+import java.util.List;
+
+import com.oracle.svm.core.annotate.Substitute;
+import com.oracle.svm.core.annotate.TargetClass;
+
+/**
+ * The liquibase service locator substitute replace liquibase classpath scanner method
```
gsmet:
```suggestion
 * The liquibase service locator substitute replaces liquibase classpath scanner method
```

extensions/liquibase/runtime/src/main/java/io/quarkus/liquibase/runtime/graal/SubstituteServiceLocator.java
```diff
@@ -0,0 +1,21 @@
...
+ * {@link liquibase.servicelocator.ServiceLocator#findClasses(Class)} with a custom implementation
+ * {@link LiquibaseServiceLoader#findClassesImpl(Class)} which used the prebuild txt file.
```
gsmet:
```suggestion
 * {@link LiquibaseServiceLoader#findClassesImpl(Class)} which used the prebuilt txt file.
```

gsmet (changes requested):
I added a couple of small questions and remarks.

I'm a bit worried about the instantiation of a `Liquibase` instance for each operation. I think that's the one remaining issue but I'm not sure what we can do about it.

extensions/liquibase/deployment/pom.xml
andrejpetras: We don't need a PostgreSQL dependency there. I will remove it.

bom/runtime/pom.xml
andrejpetras: I would like to avoid the osgi dependency in general. The osgi dependency is used only because of this static code which is not used in the quarkus: https://github.com/liquibase/liquibase/blob/9908038ad4fd6341470feb4d1aa1c5d084d0eaa9/liquibase-core/src/main/java/liquibase/servicelocator/ServiceLocator.java#L38. Is there any way to avoid the dependency?

I will remove the excludes.


[INFO] +- io.quarkus:quarkus-liquibase:jar:999-SNAPSHOT:compile
[INFO] |  +- org.liquibase:liquibase-core:jar:3.8.4:compile
[INFO] |  +- org.yaml:snakeyaml:jar:1.25:compile
[INFO] |  +- io.quarkus:quarkus-jaxb:jar:999-SNAPSHOT:compile
[INFO] |  |  +- org.glassfish.jaxb:jaxb-runtime:jar:2.3.3-b01:compile
[INFO] |  |  |  +- org.glassfish.jaxb:txw2:jar:2.3.3-b01:compile
[INFO] |  |  |  +- com.sun.istack:istack-commons-runtime:jar:3.0.10:compile
[INFO] |  |  |  \- com.sun.activation:jakarta.activation:jar:1.2.1:runtime
[INFO] |  |  \- org.jboss.spec.javax.xml.bind:jboss-jaxb-api_2.3_spec:jar:2.0.0.Final:compile
[INFO] |  +- io.quarkus:quarkus-narayana-jta:jar:999-SNAPSHOT:compile
[INFO] |  |  +- io.smallrye.reactive:smallrye-reactive-streams-operators:jar:1.0.10:compile
[INFO] |  |  |  +- org.eclipse.microprofile.reactive-streams-operators:microprofile-reactive-streams-operators-api:jar:1.0:compile
[INFO] |  |  |  |  \- org.reactivestreams:reactive-streams:jar:1.0.3:compile
[INFO] |  |  |  +- org.eclipse.microprofile.reactive-streams-operators:microprofile-reactive-streams-operators-core:jar:1.0:compile
[INFO] |  |  |  \- io.reactivex.rxjava2:rxjava:jar:2.2.15:compile
[INFO] |  |  +- io.smallrye:smallrye-context-propagation-jta:jar:1.0.11:compile
[INFO] |  |  \- io.smallrye.reactive:smallrye-reactive-converter-api:jar:1.0.10:compile
[INFO] |  \- org.osgi:org.osgi.core:jar:6.0.0:compile

extensions/liquibase/runtime/src/main/java/io/quarkus/liquibase/Liquibase.java
andrejpetras: @gsmet I am not sure but I found another option which could be the one

I am using the data-source driver to create a liquibase database and liquibase instance is created only once. 

Then we have in each Liquibase bean method this logic:
Open JDBC connection
Execute liquibase.Liquibase method
Close JDBC connection

https://github.com/andrejpetras/quarkus/blob/de19e6e3de27ce9ede78069addce066dfb8c483e/extensions/liquibase/runtime/src/main/java/io/quarkus/liquibase/Liquibase.java#L65

Like hibernate-orm or quartz extension I have guess method for the liquibase database:
https://github.com/andrejpetras/quarkus/blob/de19e6e3de27ce9ede78069addce066dfb8c483e/extensions/liquibase/runtime/src/main/java/io/quarkus/liquibase/Liquibase.java#L105

I am not sure if it is possible to move this part in the LiquibaseProcessor but I found this:
https://github.com/quarkusio/quarkus/blob/f8ccf737a65461b4c123e961986a3d4c35adefd6/extensions/agroal/deployment/src/main/java/io/quarkus/agroal/deployment/AgroalProcessor.java#L113

It looks like DataSourceDriverBuildItem currently supports only a default datasource

Commit: 
https://github.com/andrejpetras/quarkus/commit/de19e6e3de27ce9ede78069addce066dfb8c483e

extensions/liquibase/deployment/src/main/java/io/quarkus/liquibase/LiquibaseProcessor.java
andrejpetras: Ok, I add to the liquibase-runtime the jaxb extension

andrejpetras:
@gsmet @gastaldi 

How can I fix the problem with an Azure job?
Error: The job running on agent Azure Pipelines 12 ran longer than the maximum time of 30 minutes.

gastaldi:
I usually.hit the "Re-run failed checks" button in the Checks tab. You probably need special permissions if that is not available for you. Anyway I triggered that for you in this PR

andrejpetras:
> I usually.hit the "Re-run failed checks" button in the Checks tab. You probably need special permissions if that is not available for you. Anyway I triggered that for you in this PR

Oh, thank you.

extensions/liquibase/runtime/src/main/java/io/quarkus/liquibase/Liquibase.java
andrejpetras: @gsmet could you please check the solution in this commit? https://github.com/andrejpetras/quarkus/commit/de19e6e3de27ce9ede78069addce066dfb8c483e
Looks like this is the last open point in this PR

andrejpetras:
@gastaldi @gsmet Can I also create the documentation page?

gastaldi:
@andrejpetras yes please, that would be great

andrejpetras:
@gastaldi should I create separate PR for it?

gastaldi:
No, I'd prefer if you added to this PR

extensions/liquibase/runtime/src/main/java/io/quarkus/liquibase/Liquibase.java
andrejpetras: Only the JDBC connection is created for each method execution

extensions/liquibase/runtime/src/main/java/io/quarkus/liquibase/Liquibase.java
andrejpetras: Done

andrejpetras:
@gsmet @gastaldi
I added the documentation page.
The Liquibase object is now created only once, and only the JDBC connection is created for each Liquibase method.

Could please review this PR again.

docs/src/main/asciidoc/liquibase.adoc
```diff
@@ -0,0 +1,270 @@
...
+The name of the catalog with the liquibase tables +
+**default:** <none>
+
+`quarkus.liquibase.liquibase-schema-name`::
+The name of the schema with the liquibase tables. +
+**default:** <none>
+
+`quarkus.liquibase.liquibase-tablespace-name`::
+The name of the tablespace where the -LOG and -LOCK tables will be created (if they do not exist yet) +
+**default:** <none>
```
gastaldi: I think this section can be generated automatically if you add 
```adoc
include::{generated-dir}/config/quarkus-liquibase.adoc[opts=optional, leveloffset=+2]
```

andrejpetras: @gastaldi thanks for the hint. Done

andrejpetras:
@gsmet @gastaldi do we need to fix any  other task or issue in this PR?

docs/src/main/asciidoc/liquibase.adoc
```diff
@@ -0,0 +1,219 @@
...
+The syntax is as follows: `quarkus.liquibase.[optional name.][datasource property]`.
+
+NOTE: Without configuration, Liquibase is set up for every datasource using the default settings.
+
+== Using the Liquibase object
+
+In case you are interested in using the `Liquibase` object directly, you can inject it as follows:
+
+NOTE: The Liquibase object is the instance of the org.quarkus.liquibase.Liquibase class. This kind of wrapper object to handle the
+JDBC connection for the liquibase.Liquibase object. The instance open and close the database connection for each method separately.
```
gastaldi: This wording is a bit confusing.

extensions/liquibase/deployment/src/test/java/io/quarkus/liquibase/test/LiquibaseExtensionConfigFixture.java
```diff
@@ -0,0 +1,156 @@
...
+import io.quarkus.liquibase.runtime.LiquibaseConfig;
+import io.quarkus.liquibase.runtime.LiquibaseDataSourceBuildTimeConfig;
+import liquibase.configuration.GlobalConfiguration;
+import liquibase.configuration.LiquibaseConfiguration;
+
+/**
+ * This fixture provides access to read the expected and the actual configuration of liquibase.
+ * It also provides a method combining all assertions to be reused for multiple tests.
+ */
+@Disabled
```
gastaldi: Why this class needs a `Disabled` annotation?

docs/src/main/asciidoc/liquibase.adoc
andrejpetras: What would you suggest?
It is deep technical point and for guides is not so relevant, I can remove this note if you are ok with that.

extensions/liquibase/deployment/src/test/java/io/quarkus/liquibase/test/LiquibaseExtensionConfigFixture.java
andrejpetras: It's copy & paste from the Flyway extension. My understanding is that we do not use this class as a test class, but only as a test help class.

andrejpetras: @gastaldi should I remove it?

extensions/liquibase/runtime/src/main/java/io/quarkus/liquibase/Liquibase.java
gsmet: Let's not mark conversation as resolved if we are discussing it :). Sorry, I was away for a bit, will have a look at your commit.

docs/src/main/asciidoc/liquibase.adoc
```diff
@@ -0,0 +1,219 @@
...
+        liquibase.dropAll(); <4>
+        liquibase.validate();
+        liquibase.update();
+        // Get the list of liquibase change set statuses
+        List<ChangeSetStatus> status = liquibase.getChangeSetStatuses(); <5>
+    }
+}
+--
+
+<1> Inject the Liquibase object if you want to use it directly
```
gastaldi: Better move the deep technical explanation here

docs/src/main/asciidoc/liquibase.adoc
gastaldi: Better move this to the bullet point description (<1>), maybe something along the lines of:

```
NOTE: As a technical detail, the injected `Liquibase` object is an `org.quarkus.liquibase.Liquibase` instance, which acts as a wrapper for the underlying JDBC connection and is used to open and close the database connection for each method separately

File: extensions/liquibase/deployment/src/test/java/io/quarkus/liquibase/test/LiquibaseExtensionConfigFixture.java (hunk 31 above)
Comment (gastaldi) [reply extensions/liquibase/deployment/src/test/java/io/quarkus/liquibase/test/LiquibaseExtensionConfigFixture.java]: If it is not necessary I believe it's better to remove it :wink:

File: extensions/liquibase/runtime/src/main/java/io/quarkus/liquibase/Liquibase.java (hunk 24 above)
Comment (gsmet) [reply extensions/liquibase/runtime/src/main/java/io/quarkus/liquibase/Liquibase.java]: I'm not convinced it's a good idea to hardcode the databases here. We want to get rid of it elsewhere so I don't think generalizing it is a good idea.

What I was wondering is if we should have something like:

```java
@Inject
LiquibaseHandler liquibaseHandler;

public void myMethod() {
    try (Liquibase liquibase = liquibaseHandler.open()) {
         // do your Liquibase operations here
    }
}
```

The `LiquibaseHandler` name might not be ideal but it's the pattern I'm interested in. We would use a proper try with resource to handle the close (providing `Liquibase` is closeable, I'm not sure?).

That way the connection would be short lived but we could still reuse the same connection for several operations and we wouldn't have to wrap the `Liquibase` class.

WDYT? /cc @gastaldi

gsmet: BTW, it's just an idea, I haven't checked it could work so maybe check if it could be doable but let's discuss it with you and @gastaldi before doing too much work.

gastaldi: AFAIR there isn't a concept of "closing" a Liquibase object in the Liquibase API, but I may be wrong (it's been a while since I used it). I think the connection is kept open until it `DatabaseConnection.close` is explicitly closed

andrejpetras: @gsmet sorry for the "resolved conversation".

Please take a look at the current implementation in PR
extensions/liquibase/runtime/src/main/java/io/quarkus/liquibase/Liquibase.java

gastaldi: But yeah, I can see why a try-with-resources block may be useful here

andrejpetras: > AFAIR there isn't a concept of "closing" a Liquibase object in the Liquibase API, but I may be wrong (it's been a while since I used it). I think the connection is kept open until it `DatabaseConnection.close` is explicitly closed

Yes, this: ```Liquibase.getDatabase().close();```

andrejpetras: >The LiquibaseHandler name might not be ideal but it's the pattern I'm interested in. We would use a proper try with resource to handle the close (providing Liquibase is closeable, I'm not sure?).

The Liquibase object doesn't implement Closeable.

Also some of the methods like dropAll are final
```
public final void dropAll() throws DatabaseException {
        this.dropAll(new CatalogAndSchema(this.getDatabase().getDefaultCatalogName(), this.getDatabase().getDefaultSchemaName()));
    }
```

docs/src/main/asciidoc/liquibase.adoc
andrejpetras: Like this?
image

gastaldi: yeah, that looks cool

extensions/liquibase/runtime/src/main/java/io/quarkus/liquibase/Liquibase.java
andrejpetras: @gsmet the ```LiquibaseHandler``` makes sense to me.

This should be possible:
```
package org.quarkus.liquibase;
public class Liquibase implements Closeable {
}
```

```
@Inject
LiquibaseHandler liquibaseHandler;

public void myMethod() {
    try (org.quarkus.liquibase.Liquibase liquibase = liquibaseHandler.open()) {
         // do your Liquibase operations here
    }
}
```

gastaldi: Another possibility is to leave the closing operations away from the user code. Say you have something like:

```java
public interface LiquibaseConsumer {
    void consume(Liquibase liquibase);
}
```

And in the extension you could iterate through all the LiquibaseConsumer implementations and close the connection afterwards

gastaldi: Although I have seen in practice that users don't really use the Liquibase object, they just put the changelog.xml files in a specific folder and Liquibase automatically uses them to perform the necessary DB operations. (Through a [LiquibaseServletListener](https://www.liquibase.org/javadoc/liquibase/integration/servlet/LiquibaseServletListener.html))

andrejpetras: mmh, what exactly would be the point where we wanted to close the connection?

andrejpetras: Would it be possible to remove the "final" flag from the Liquibase method by modifying the byte code?

andrejpetras: I can try this:
```
public class LiquibaseImpl extends liquibase.Liquibase implements Closeable { }
```

```
    try (liquibase.Liquibase liquibase = liquibaseHandler.open()) {
         // do your Liquibase operations here
    }
```

gastaldi: > mmh, what exactly would be the point where we wanted to close the connection?

What would be the use case where you'll want to close the connection before all changes are performed?

andrejpetras: For example, the user can start the migration and then read the Liquibase information about the changes.

```
liquibase.update();
List<ChangeSetStatus> status = liquibase.getChangeSetStatuses();
```

andrejpetras: @gsmet @gastaldi 

Please check this:
https://github.com/andrejpetras/quarkus/commit/e061b821b4fe60084f1bccbae0acf60c75177738

```java
    @Inject
    LiquibaseFactory liquibaseFactory;

    try (LiquibaseContext liquibase = liquibaseFactory.createContext()) {
         liquibase.update();
    }
```

- no hardcode databases
- the names LiquibaseFactory and LiquibaseContext might not be ideal.

gastaldi: That looks cool

andrejpetras: Should I merge it into PR?

andrejpetras: @gsmet please have a look at the solution.

andrejpetras: PR update for liquibase 3.8.7 version
```   @Inject
    LiquibaseFactory liquibaseFactory;

    try (Liquibase liquibase = liquibaseFactory.createLiquibase()) {
         liquibase.update(liquibaseFactory.createContexts(),liquibaseFactory.createLabels());
    }```

Comment (andrejpetras) [reply extensions/liquibase/runtime/src/main/java/io/quarkus/liquibase/Liquibase.java]: The liquibase version 3.8.7 will be release in the next few weeks
Liquibase PR: https://github.com/liquibase/liquibase/pull/987

File: extensions/liquibase/deployment/src/main/java/io/quarkus/liquibase/LiquibaseProcessor.java (hunk 1 above)
Comment (andrejpetras) [reply extensions/liquibase/deployment/src/main/java/io/quarkus/liquibase/LiquibaseProcessor.java]: @gsme please check the regex for the native resources build item:
https://github.com/andrejpetras/quarkus/commit/0c5eee023ec085ff77470d485f42a4cfe90b289c

Comment (andrejpetras) [reply extensions/liquibase/deployment/src/main/java/io/quarkus/liquibase/LiquibaseProcessor.java]: https://github.com/quarkusio/quarkus/issues/7033

File: bom/runtime/pom.xml [hunk 33]
```diff
@@ -2685,6 +2693,31 @@
...
+            </dependency>
+            <dependency>
+                <groupId>org.yaml</groupId>
+                <artifactId>snakeyaml</artifactId>
+                <version>${snakeyaml.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>org.osgi</groupId>
+                <artifactId>org.osgi.core</artifactId>
+                <version>${osgi.version}</version>
```
gsmet: Is this strictly needed or can we use an exclusion instead?

gsmet (changes requested):
@andrejpetras I pushed a rebase with an upgrade to 3.8.7.

I added a small question about the osgi dependency. Could you have a look?

bom/runtime/pom.xml
andrejpetras: @gsmet cool, thanks

I believe that I add this dependency for native build because of this class:
https://github.com/liquibase/liquibase/blob/9626494719cf3f6e698bcdd76ee97b9eee1ff717/liquibase-core/src/main/java/liquibase/servicelocator/ServiceLocator.java#L7

The OSGI is used in the static block of the class
```java
    static {
        try {
            Class<?> scanner = Class.forName("Liquibase.ServiceLocator.ClrServiceLocator, Liquibase");
            instance = (ServiceLocator) scanner.newInstance();
        } catch (Exception e) {
            try {
                if (OSGiUtil.isLiquibaseLoadedAsOSGiBundle()) {
                    Bundle liquibaseBundle = FrameworkUtil.getBundle(ServiceLocator.class);
                    instance = new ServiceLocator(new OSGiPackageScanClassResolver(liquibaseBundle), 
                            new OSGiResourceAccessor(liquibaseBundle));
                } else {
                    instance = new ServiceLocator();
                }
            } catch (Throwable e1) {
                LogService.getLog(ServiceLocator.class).severe(LogType.LOG, "Cannot build ServiceLocator", e1);
            }
        }
    }
```
We are not using this in the quarkus at all. 
I have no idea how it is possible to substitute the static block without replacing the whole class for the native build.

gsmet: @andrejpetras we would need to move that code to a static method and substitute that method. Maybe you can prepare another PR for Liquibase?

gsmet: Oh wait silly me, we would still need the dependency anyway so let's leave it at that.

docs/src/main/asciidoc/liquibase.adoc
andrejpetras: Done

extensions/liquibase/deployment/src/test/java/io/quarkus/liquibase/test/LiquibaseExtensionConfigFixture.java
andrejpetras: Done

bom/runtime/pom.xml
andrejpetras: @gsmet I check the 4.0.x branch of the liquibase and the class does not have osgi packages anymore. We can check the liquibase extension again after liquibase release 4.0.0 version. (This will happen in few months or more ...) But this change should not affect the liquibase extension API for developer. We can change it later in background and clean up the dependencies and also fix the native resources with a regex. We can have new task for it and I can take it to synchronize the liquibase changes in the extension.

gsmet:
Yes, the idea is to merge it today but it's in conflict with my big datasource config PR so I would like to get this one in and then update the Liquibase PR.

gsmet:
I just force-pushed an update with a migration to the new datasource config stuff and some fixes.

Let's see how it goes with CI.

gsmet (approved):

gsmet:
@andrejpetras I'm going to merge that one and include it in CR2. The code is totally isolated so it's not a problem.

Thanks a lot for your work and your tenacity.

PR #6343, opened 2019-12-26
spring-di: Handle javax.inject.Named on implict singleton bean

### LINKED ISSUE #6342: spring-di: @Named should be a bean defining annotation
**Describe the bug**

While migrating a spring war application based on JSR-303 to Quarkus, beans annotated with `@javax.inject.Named` and without an explicit scope are not seen by Arc.

Indeed CDI, `@javax.inject.Named` is not a bean defining annotation by itself.
But in Spring DI it is an equivalent to `@org.springframework.stereotype.Component` (with the difference that it is not a composable annotation) - cf https://github.com/spring-projects/spring-framework/blob/76645644b1cb99cbee155dfea2852ee13914b3e3/src/docs/asciidoc/core/core-beans.adoc#named-and-managedbean-standard-equivalents-to-the-component-annotation

Here is a working spring test:

```java
@ExtendWith(SpringExtension.class)
@ContextConfiguration(classes = { SpringNamedTest.MyBean.class })
public class SpringNamedTest {

  @Inject
  MyBean myBean;

  @Test
  public void testInjection() {
    assertNotNull(myBean);
  }

  @Named
  public static class MyBean {
  }
}
```

**Expected behavior**

Dependency injection of MyBean works.

**Actual behavior**

No qualifying bean found by Arc.

 - Output of `java -version`: 
java version "1.8.0_201"
Java(TM) SE Runtime Environment (build 1.8.0_201-b09)
Java HotSpot(TM) 64-Bit Server VM (build 25.201-b09, mixed mode)
 - Quarkus version or git rev: 1.1.0.Final

**Additional context**
I'll open a PR with proposed reproducer on integration-tests/spring-di and a fix.

rquinio:
Fixes #6342

### DIFF
--- a/extensions/spring-di/deployment/src/main/java/io/quarkus/spring/di/deployment/SpringDIProcessor.java
+++ b/extensions/spring-di/deployment/src/main/java/io/quarkus/spring/di/deployment/SpringDIProcessor.java
@@ -313,7 +313,11 @@ public class SpringDIProcessor {
                     }
                 }
             }
-            final DotName declaredScope = getScope(classInfo);
+            DotName declaredScope = getScope(classInfo);
+            // @Named is a bean-defining annotation in Spring, but not in Arc.
+            if (declaredScope == null && classInfo.classAnnotation(CDI_NAMED_ANNOTATION) != null) {
+                declaredScope = CDI_SINGLETON_ANNOTATION; // implicit default scope in spring
+            }
             final boolean isAnnotation = isAnnotation(classInfo.flags());
             if (declaredScope != null) {
                 annotationsToAdd.add(create(
--- a/integration-tests/spring-di/src/main/java/io/quarkus/it/spring/AppConfiguration.java
+++ b/integration-tests/spring-di/src/main/java/io/quarkus/it/spring/AppConfiguration.java
@@ -1,5 +1,6 @@
 package io.quarkus.it.spring;
 
+import javax.inject.Named;
 import javax.inject.Singleton;
 
 import org.springframework.beans.factory.annotation.Qualifier;
@@ -48,6 +49,11 @@ public class AppConfiguration {
     }
 
     public static class CustomPrototypeBean {
+
+    }
+
+    @Named
+    public static class NamedBean {
 
     }
 }
--- a/integration-tests/spring-di/src/main/java/io/quarkus/it/spring/InjectedSpringBeansResource.java
+++ b/integration-tests/spring-di/src/main/java/io/quarkus/it/spring/InjectedSpringBeansResource.java
@@ -1,7 +1,5 @@
 package io.quarkus.it.spring;
 
-import static io.quarkus.it.spring.AppConfiguration.CustomPrototypeBean;
-
 import javax.inject.Inject;
 import javax.servlet.http.HttpServletRequest;
 import javax.ws.rs.GET;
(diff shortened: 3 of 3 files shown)


gsmet:
There is a formatting issue. Can you run `mvn process-sources` and amend your commit?

Thanks!

extensions/spring-di/deployment/src/main/java/io/quarkus/spring/di/deployment/SpringDIProcessor.java
```diff
@@ -88,6 +89,14 @@ FeatureBuildItem registerFeature() {
         return new FeatureBuildItem(FeatureBuildItem.SPRING_DI);
     }
 
+    /*
+     * In spring-di @Named is a bean defining annotation, while it's not in CDI.
+     */
+    @BuildStep
+    BeanDefiningAnnotationBuildItem registerNamedAnnotation() {
+        return new BeanDefiningAnnotationBuildItem(CDI_NAMED_ANNOTATION);
```
gsmet: What's the default scope of these `@Named` beans in Spring?

geoand: I assume Singleton? But I am not sure

gsmet: Yes, my question basically was: we need to check it and be consistent with what Spring does :).

geoand: Yup 👍. I am 90% certain that's the case. Hopefully @rquinio can provide some info since I don't really have much computer access for a few days

rquinio: I confirm it defaults to singleton.
I'll put that explicitly and also allow those beans to be removable if never injected anywhere as it's not the default: 
`BeanDefiningAnnotationBuildItem(CDI_NAMED_ANNOTATION, CDI_SINGLETON_ANNOTATION, true)`

gsmet: Does Spring have this notion of removable beans? If not, I would keep them unremovable.

rquinio: Not that I know, I'll do that.

rquinio:
> There is a formatting issue. Can you run `mvn process-sources` and amend your commit?
> 
> Thanks!

Actually I'm fighting a bit with the formatter due to line endings on Windows :)

I need to use -Dlineendings=LF  (https://code.revelc.net/formatter-maven-plugin/format-mojo.html#lineEnding) on every Maven command otherwise it reformats everything since the default is AUTO, not great.
So I guess I should be using git core.autocrlf = true to checkout with line endings that match the formatter defaults ? Should I do anther PR to add these info to CONTRIBUTING.md ?

geoand:
Thanks for the contribution!

However there is an issue... With the current implementation, when the spring-di extension is added, `@Named` becomes a bean defining annotation for everything, which we don't really want.
A more correct implementation would be to change SpringDIProcessor to properly handle the annotation (it's already handled, to get the name of the bean, so it shouldn't be difficult)

rquinio:
@geoand Indeed ! I've pushing something more subtle :)
Had to do it at a different place than when reading the name though.
In fact the buggy case is only `@Named` beans without an explicit scope. We need to add the `@Singleton` explicitly, so it gets picked by Arc, like for the case of explicit scopes.

geoand:
@rquinio thanks! I don't have much access to a computer so it will be a few days before I can review more thoroughly

geoand (approved):
Looks good, thanks for this!

PR #6347, opened 2019-12-27
Upgrade mongodb clients to the latest versions

### LINKED ISSUE #6418: Upgrade MongoDB dependencies to 3.12.0
**Description**

It'd be useful if the Quarkus MongoDB dependencies could be bumped to the latest version(s) - currently 3.12.0, as it'd help the Camel Mongo extension to remain compatible.

loicmathieu:
Upgrades mongo-java-driver to 3.12
Upgrades mongodb-driver-reactivestreams to 1.13
Add the new mongodb-crypt library for field encryption inside the client

Fixes #6418

### DIFF
--- a/docs/src/main/asciidoc/mongodb.adoc
+++ b/docs/src/main/asciidoc/mongodb.adoc
@@ -515,6 +515,8 @@ You can then point your browser to `http://localhost:8080/fruits.html` and use y
 [WARNING]
 ====
 Currently, Quarkus doesn't support the `mongodb+srv` protocol in native mode.
+
+link:https://docs.mongodb.com/manual/core/security-client-side-encryption/[Client-Side Field Level Encryption] is also not supported in native mode.
 ====
 
 == Conclusion
--- a/bom/runtime/pom.xml
+++ b/bom/runtime/pom.xml
@@ -150,8 +150,9 @@
         <flyway.version>6.2.1</flyway.version>
         <yasson.version>1.0.6</yasson.version>
         <neo4j-java-driver.version>4.0.0</neo4j-java-driver.version>
-        <mongo-client.version>3.10.2</mongo-client.version>
-        <mongo-reactivestreams-client.version>1.11.0</mongo-reactivestreams-client.version>
+        <mongo-client.version>3.12.0</mongo-client.version>
+        <mongo-reactivestreams-client.version>1.13.0</mongo-reactivestreams-client.version>
+        <mongo-crypt.version>1.0.0</mongo-crypt.version>
         <artemis.version>2.11.0</artemis.version>
         <okhttp.version>3.14.6</okhttp.version>
         <sentry.version>1.7.28</sentry.version>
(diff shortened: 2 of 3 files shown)


loicmathieu:
I added the new mongodb-crypt (1.4 MB) to avoid issues on native image generation.
This library should normaly be included only if client-side field encryption is used.
It contains native libraries (libmongocrypt.so), I have no idea if native libraries can works in native applications.

An other solution could have been to add substitution to `MongClientImpl` to remove the dependency to MongoCrypt but this has the disadvantage of disabling the usage of field client-side field encryption and to add maintenance cost of the susbtitution on each driver update.

gsmet (changes requested):
> It contains native libraries (libmongocrypt.so), I have no idea if native libraries can works in native applications.

Can we check? Because it makes no sense including it if it's not supported at all in native image. There's a good chance some JNI magic will be needed to include the .so.

In any case, one option could be to have a build time config to disable it if we think the cost is too high. Or we could have substitutions if the dependency is not around, which would probably have my preference.

At least if this feature is not widely used.

loicmathieu:
@gsmet reading the JNI section of graalvm it seems that it needs some specific configuration. 
So to make it run we will need to do some reverse engineering on the code first ....
This seems a little too much so I propose to add a substitution for the moment and a note on the documentation. 

The substitution should be done on a constructor and the faulty class is in the constructor signature. Not sure how I should do it ? Substitute the class and the constructor that uses it ?

loicmathieu:
@gsmet I tried to provides substitutions but apparently SubstratVM try to load the original methods/classes so if I remove the library from the pom the substitution didn't works

```
Caused by: java.lang.NoClassDefFoundError: com/mongodb/crypt/capi/MongoCryptException
	at java.lang.Class.getDeclaredMethods0(Native Method)
	at java.lang.Class.privateGetDeclaredMethods(Class.java:2701)
	at java.lang.Class.getDeclaredMethod(Class.java:2128)
	at com.oracle.svm.hosted.substitute.AnnotationSubstitutionProcessor.findOriginalMethod(AnnotationSubstitutionProcessor.java:620)
```

If I need the mongo-crypt library to makes the native build works I didn't see any needs to provides the substitutions.

Or maybe someone that knows more about graalvm substitutions should have a look ...

gastaldi:
Fixes #6418

loicmathieu:
@gsmet I think the best for now is to keep the added `mongodb-crypt` librairy because I didn't succeed, despite a lot of attemps, to provides substitutions for it.

It it's OK for you, I'll open a followup issue to investiguate if we can provides substitutions (someone more knowledgeable on GraalVM can try) or ask upstream to refactor some part of the driver code to make a substitution more easily doable.

docs/src/main/asciidoc/mongodb.adoc
```diff
@@ -515,6 +515,8 @@ You can then point your browser to `http://localhost:8080/fruits.html` and use y
 [WARNING]
 ====
 Currently, Quarkus doesn't support the `mongodb+srv` protocol in native mode.
+
+link:https://docs.mongodb.com/manual/core/security-client-side-encryption/[Client-Side Field Level Encryption] is also not supported on native mode.
```
gsmet:
```suggestion
link:https://docs.mongodb.com/manual/core/security-client-side-encryption/[Client-Side Field Level Encryption] is also not supported in native mode.
```

gsmet (changes requested):
I asked for a small doc change, otherwise let's merge it ASAP.

loicmathieu:
@gsmet done

gsmet (approved):

PR #6348, opened 2019-12-27
[fixes #4448] - OIDC Multi-tenancy Support

### LINKED ISSUE #4448: quarkus-oidc adapter has to support KeycloakConfigResolver alternative available in VertxOAuth2
@pedroigor FYI

pedroigor:


pedroigor:
@sberyozkin docs on the way.

sberyozkin:
@pedroigor Hey Pedro, thanks for starting on it, sorry for a delay, will comment shortly

extensions/keycloak-authorization/runtime/src/main/java/io/quarkus/keycloak/pep/runtime/KeycloakPolicyEnforcerAuthorizer.java
```diff
@@ -29,6 +32,9 @@
 public class KeycloakPolicyEnforcerAuthorizer
         implements HttpSecurityPolicy, BiFunction<RoutingContext, SecurityIdentity, HttpSecurityPolicy.CheckResult> {
 
+    @Inject
```
sberyozkin: This injection is probably not needed yet as `quarkus-keycloak-authorization` will get the multi-tenancy support in a follow up PR ?

extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/OidcRuntimeConfig.java
```diff
@@ -0,0 +1,212 @@
...
+    /**
+     * The client-id of the application. Each application has a client-id that is used to identify the application
+     */
+    @ConfigItem
+    Optional<String> clientId;
+    /**
+     * Configuration to find and parse a custom claim containing the roles information.
+     */
+    @ConfigItem
+    OidcRuntimeConfig.Roles roles;
```
sberyozkin: Can `OidRuntimeConfig.` qualifier be dropped for this and few follow up properties ?

extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/TenantAuthenticationMechanism.java
```diff
@@ -0,0 +1,26 @@
...
+import javax.enterprise.context.ApplicationScoped;
+
+import io.quarkus.security.identity.IdentityProviderManager;
+import io.quarkus.security.identity.SecurityIdentity;
+import io.quarkus.vertx.http.runtime.security.ChallengeData;
+import io.quarkus.vertx.http.runtime.security.HttpAuthenticationMechanism;
+import io.vertx.ext.web.RoutingContext;
+
+@ApplicationScoped
+public class TenantAuthenticationMechanism implements HttpAuthenticationMechanism {
```
sberyozkin: What is the idea behind this class ?

sberyozkin:
@pedroigor It looks brilliant to me, it is quite an involved PR but it is already very precise, nice work :-)  I've left few cosmetic comments, and I I guess `TokenAuthenticationMechanism` is not  needed.

One thing I'm not sure about is how can we tie it to this [MultiTenantHandler](https://github.com/vert-x3/vertx-web/pull/1398/files#diff-7c105f6d214bca719528dad0d93b8e07R46). I think it can fit well with your own idea of supporting an injectable bean to support a more dynamic configuration resolution, what do you think ?  

Indeed, as you suggest, lets do the `quarkus-keycloak-authorization` multitetnancy update in a follow up PR, lets have this one affecting `quarkus-oidc` settling in first.

pedroigor:
> One thing I'm not sure about is how can we tie it to this [MultiTenantHandler](https://github.com/vert-x3/vertx-web/pull/1398/files#diff-7c105f6d214bca719528dad0d93b8e07R46). I think it can fit well with your own idea of supporting an injectable bean to support a more dynamic configuration resolution, what do you think ?

I've looked at the `MultitenantHandler` but I did not figure out how we could leverage it. That is one of the reasons for bringing most of the tenant resolution logic to the extension. One of the issues I found was how to register the handler given that most of the handler registration is done upstream by the security extension. I think the way it is implemented is quite simple though and flexible enough to support different forms of tenant resolution.

extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/OidcRuntimeConfig.java
pedroigor: Not sure what you mean ... Could you elaborate more please ?

sberyozkin:
@pedroigor Yes, it looks like we can get more control/flexibility with managing it directly at the extension level. I suppose we can wire in that handler as well in due time

extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/OidcRuntimeConfig.java
sberyozkin: @pedroigor I mean, can it just be typed as 
`Roles roles;` ?

pedroigor: Ahh, right. I'll change, sorry ...

extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/OidcRuntimeConfig.java
```diff
@@ -0,0 +1,212 @@
...
+import java.time.Duration;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Optional;
+
+import io.quarkus.runtime.annotations.ConfigGroup;
+import io.quarkus.runtime.annotations.ConfigItem;
+
+@ConfigGroup
+public class OidcRuntimeConfig {
```
sberyozkin: @pedroigor I've just noticed that we also have `OidcConfig` in the runtime package. I'd say we should have this class renamed to `OidcTenantConfig`, while `TenantConfig` to something else, say `CurrentTenantConfig` or `TenantConfigContext`, something like that, what do you think ?

sberyozkin:
@pedroigor Hi Pedro, I'm also looking at `io.quarkus.oidc.TenantResolver` which resolves a tenant identifier. And then there is an internal `io.quarkus.oidc.runtime.TenantConfigResolver` which uses `TenantResolver` to get the identifier and fetch the correct configuration which was set in the `application.properties`. This is all works very good, but I'm keeping thinking now of how we'd wire in the custom resolver as you also suggested earlier and it feels like some renaming/refactoring may be needed.

It feels like we may need to have `io.quarkus.oidc.TenantConfigResolver` interface, which, given `RoutingContext`, would return what is currently named  `TenantConfig` (possibly renamed as per the earlier comment), and the current `io.quarkus.oidc.runtiume.TenantConfigResolver` then becomes `io.quarkus.oidc.runtiume.TenantConfigResolverImpl` which would check the custom `TenantConfigResolver`s and if they return null, then it would return the config as per the current code.

How does it sound ?

extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/OidcRuntimeConfig.java
pedroigor: Done

pedroigor:
I'm not sure @sberyozkin. I did the renaming as you suggested though.

Users should just provide a bean implementing the `TenantResolver` interface as demonstrated by the integration tests.

The `TenantConfigResolver` is really an internal component/intermediary (between app and extension) that we can do whatever we need to actually map the tenant identifier to the proper configuration.

sberyozkin:
@pedroigor great, thanks for renaming the config classes, IMHO it is a bit clearer this way.

Yeah, I see how the current mechanism works which is neat, but all this mechanism is about currently is about loading a configuration fragment from `application.properties`. If, as per the earlier discussion, we want to let the users to load the configuration objects (similarly to the old `KeycloakConfigResolver`) then we'd need to introduce some interface, see what I mean ?

sberyozkin:
@pedroigor I'd just like us to have an idea now how the configuration as an object can be provided, I'd be fine with it going in the follow up PR. For example, `TenantResolver` is good in what it does but if there will be another interface where the user beans will return a config object then `TenantResolver` might become redundant because the would be future custom beans will just return the config, either from the `application.properties` or created in the code. May be  `TenantResolver` would stay either way, but lets analyze it a bit more, thanks

pedroigor:
@sberyozkin I see now, it is indeed missing.

In fact, if you have multiple tenants forcing users to have then configured in the properties file is not good.

I think one of the reasons for not doing it was that we would need to build the `OAuth2Auth` whenever a tenant config is resolved (if not previously resolved). I was unsure about the implications in terms of performance and footprint given that when creating `OAuth2Auth` instances there is a discovery call and its relationship with a single client (client_id).

sberyozkin:
@pedroigor Yeah, that is a reasonable concern. (though `client_id` should be `OidcTenantConfig` specific now). [This Vertx example](https://github.com/vert-x3/vertx-web/pull/1398/files#diff-146a764389fbe05673312b2c7cfa59c1R1476) though looks quite promising. 
I mean we do not have to use that `MultiTenancyHandler`, but the custom resolver beans (if we were to support them) would just create `OAuth2Auth` objects similarly to that example at the bean post-construct initialization time.
I don't know, may be we should detect these custom beans and internally register them using the static `MultiTenancyHandler` helper methods.

pedroigor:
@sberyozkin I did some changes in the second commit in order to support resolving tenant config dynamically.

extensions/oidc/runtime/src/main/java/io/quarkus/oidc/TenantConfigResolver.java
```diff
@@ -0,0 +1,24 @@
...
+/**
+ * <p>
+ * A tenant resolver is responsible for resolving the {@link OidcTenantConfig} for tenants, dynamically.
+ * 
+ * <p>
+ * Instead of implementing a {@link TenantResolver} that maps the tenant configuration based on an identifier and its
+ * corresponding entry in the application configuration file, beans implementing this interface can dynamically construct the
+ * tenant configuration without having to define each tenant in the application configuration file.
+ */
+public interface TenantConfigResolver {
```
sberyozkin: Hi Pedro @pedroigor, this is a step in the right direction IMHO :-). I've been thinking today how to pre-calculate the connections, similarly to the way it is demoed with the Vertx multi-tenancy handler.
So here is a proposal for this interface:
```
public interface TenantConfgResolver extends TenantResolver {
    // Map key is the tenant id 
    Map<String, OidcTenantConfig> getTenantConfigurations();
}
```
Given this interface, the default resolver will precalculate all OAuth2Auth objects and cache them in the recorder, and at runtime, once the tenant id is resolved, it will return TenantConfigContext. Users won't have to implement,  only if they want to.

How does it sound ? Otherwise, in order to avoid recalculating `OAuth2Auth`, we need to have hashcode equals/impementations on the OidcTenantConfig, etc...

sberyozkin:
@pedroigor thanks, it is looking good, it is very close IMHO, somehow it would be good for the perf reasons to precalcuate the connections, so I've typed one proposal, may be there could be better options as well.
Thanks

sberyozkin:
@pedroigor by the way, can you please check it will work if no default configuration is available, and only tenant-specific one is available

extensions/oidc/runtime/src/main/java/io/quarkus/oidc/TenantConfigResolver.java
pedroigor: I'm not sure ... Maybe if we get more feedback and any requirement that justify doing this?

I think the current strategy of caching the tenant context is enough ... However, something is missing here. We need to add some concurrency logic when creating the tenant context for the first time for a given tenant.

pedroigor:
@sberyozkin you spotted an issue. I fixed and now we can run without a default config.

Shall I create a new module to test this scenario (ideally we could just have a single module, but I think we discussed this before) ?

stuartwdouglas (approved):
I think this looks ok, but it is not really my area.

extensions/oidc/runtime/src/main/java/io/quarkus/oidc/TenantConfigResolver.java
sberyozkin: @pedroigor Hi Pedro, just got back... The proposed modification does not introduce anything new but the option to precalculate the connections. 
If the user sets all the multi-tenant options in `application.properties` then `OidcRecorder` does it and at runtime the connections are already available. 
If the user goes with `TenantConfgResolver` then the connection is established at runtime, it is cached though.
If we can have a case where `TenantConfgResolver` can provided an open set of `OidcTenantConfig` then my proposal won't make sense.
If `TenantConfigResolver` implementation is provided solely because putting the tenant config into `application.properties` is not possible, then there is a performance loss compared to a case where `application.properties` are used.
So I'm not exactly sure what can be improved but it feels like something can be improved :-)

OK, here it is: can we rename `TenantConfgResolver` in your current version to `RuntimeTenantConfgResolver` ? If we do it then we can quickly add later on  `StaticTenantConfgResolver` typed as per my proposal above to address the case where a set of tenant configurations is not open ended (which would be typical enough I guess), we'd just have `OidcRecorder` check an `StaticTenantConfgResolver` bean and deal with it the same way it does with the tenant config in `application.properties`. 

So I'd be happy with just `TenantConfgResolver` to `RuntimeTenantConfgResolver` name change. I can add `StaticTenantConfgResolver` later, or if you don't mind, add it now too. I think we'd cover almost 100% of all the cases here.

sberyozkin:
@stuartwdouglas thanks, yes, it looks good. I've added [one more comment](https://github.com/quarkusio/quarkus/pull/6348#discussion_r365844179) for Pedro @pedroigor (and yourself) to check (just a rename of one interface would do for now or may be a bit more :-) ), we can merge it shortly.

sberyozkin:
@pedroigor Keeping things as they already are would be fine too, I can come up with some name similar to `StaticTenantConfigResolver` :-) such as `TenantConfigMapSource`, etc.
`RuntimeTenantConfigResolver` still seems not bad to me but it is not a big deal.

Lets make a final decision and I can merge immediately afterwards, thanks

sberyozkin:
@pedroigor I've realized that the static resolver that I've proposed is most likely redundant at this stage and would duplicate the combination of `TenantResolver` and per-tenant config in the application.properties

sberyozkin (approved):

DorianMaliszewski:
Hi,

Did you finished this feature to implement multi tenancy ? I just show all of your code, commits, ... and I don't see what can I implement to make it possible.

On the official documentation I saw quarkus.oidc."tenant" configuration properties, in my case I try quarkus.oidc.hospitalink and try passing the tenant name in Headers:
- tenant : hospitalink
- X-Tenant : hospitalink
- realm : hospitalink

In cookies : 
- X-Tenant : hospitalink

It always tries to get the default oidc configuration. Do you need some help on that feature ?

FYI, I didn't try to implement some classes, I just tried with the configuration properties file. Moreover, quarkus completely ignore my KeycloakConfigResolver ApplicationScoped Bean :sweat_smile: 

Thank you in advance, :smile:

PR #6351, opened 2019-12-28
fix typo in quarkus-amazon-lambda extension's maven archetype resources

kdnakt:
Method name is `process`, not `proces`.
 
https://github.com/quarkusio/quarkus/blob/master/extensions/amazon-lambda/maven-archetype/src/main/resources/archetype-resources/src/main/java/ProcessingService.java#L10


machi1990 (approved):

machi1990:
Good catch, thanks @kdnakt

PR #6357, opened 2019-12-30
Provides additional configuration options for logging-gelf

### LINKED ISSUE #6336: Add more config options for quarkus-logging-gelf
Currently, logging-gelf extension is not allowing to set facility and other options to log to graylog.
Please, consider add all possible settings to make it fully configurable

loicmathieu:
Fixes #6336

### DIFF
--- a/extensions/logging-gelf/runtime/src/main/java/io/quarkus/logging/gelf/AdditionalFieldConfig.java
+++ b/extensions/logging-gelf/runtime/src/main/java/io/quarkus/logging/gelf/AdditionalFieldConfig.java
@@ -1,0 +1,24 @@
+package io.quarkus.logging.gelf;
+
+import io.quarkus.runtime.annotations.ConfigGroup;
+import io.quarkus.runtime.annotations.ConfigItem;
+
+/**
+ * Post additional fields. E.g. `fieldName1=value1,fieldName2=value2`.
+ */
+@ConfigGroup
+public class AdditionalFieldConfig {
+    /**
+     * Additional field value.
+     */
+    @ConfigItem
+    public String value;
+
+    /**
+     * Additional field type specification.
+     * Supported types: String, long, Long, double, Double and discover.
+     * Discover is the default if not specified, it discovers field type based on parseability.
+     */
+    @ConfigItem(defaultValue = "discover")
+    public String type;
+}
--- a/extensions/logging-gelf/runtime/src/main/java/io/quarkus/logging/gelf/GelfConfig.java
+++ b/extensions/logging-gelf/runtime/src/main/java/io/quarkus/logging/gelf/GelfConfig.java
@@ -1,5 +1,8 @@
 package io.quarkus.logging.gelf;
 
+import java.util.Optional;
+import java.util.logging.Level;
+
 import io.quarkus.runtime.annotations.ConfigItem;
 import io.quarkus.runtime.annotations.ConfigPhase;
 import io.quarkus.runtime.annotations.ConfigRoot;
@@ -61,4 +64,32 @@ public class GelfConfig {
      */
     @ConfigItem(defaultValue = "yyyy-MM-dd HH:mm:ss,SSS")
     public String timestampPattern;
+
+    /**
+     * The logging-gelf log level.
+     */
+    @ConfigItem(defaultValue = "ALL")
+    Level level;
+
+    /**
+     * Name of the Facility.
+     */
+    @ConfigItem(defaultValue = "jboss-logmanager")
+    String facility;
+
+    /**
+     * Post additional fields. E.g. `fieldName1=value1,fieldName2=value2`.
+     */
+    @ConfigItem
+    Optional<String> additionalFields;
+
+    /**
+     * Type specification for additional and MDC fields.
+     * Supported types: String, long, Long, double, Double and discover (default if not specified, discover field type on
+     * parseability).
+     * E.g. `field=String,field2=double`.
+     */
+    @ConfigItem
+    Optional<String> additionalFieldsTypes;
+
 }
--- a/extensions/logging-gelf/runtime/src/main/java/io/quarkus/logging/gelf/GelfConfig.java
+++ b/extensions/logging-gelf/runtime/src/main/java/io/quarkus/logging/gelf/GelfConfig.java
@@ -1,8 +1,10 @@
 package io.quarkus.logging.gelf;
 
-import java.util.Optional;
+import java.util.Map;
 import java.util.logging.Level;
 
+import io.quarkus.runtime.annotations.ConfigDocMapKey;
+import io.quarkus.runtime.annotations.ConfigDocSection;
 import io.quarkus.runtime.annotations.ConfigItem;
 import io.quarkus.runtime.annotations.ConfigPhase;
 import io.quarkus.runtime.annotations.ConfigRoot;
@@ -13,7 +15,7 @@ public class GelfConfig {
      * Determine whether to enable the GELF logging handler
      */
     @ConfigItem
-    boolean enabled;
+    public boolean enabled;
 
     /**
      * Hostname/IP-Address of the Logstash/Graylog Host
@@ -69,27 +71,26 @@ public class GelfConfig {
      * The logging-gelf log level.
      */
     @ConfigItem(defaultValue = "ALL")
-    Level level;
+    public Level level;
 
     /**
-     * Name of the Facility.
+     * Name of the facility.
      */
     @ConfigItem(defaultValue = "jboss-logmanager")
-    String facility;
+    public String facility;
 
     /**
-     * Post additional fields. E.g. `fieldName1=value1,fieldName2=value2`.
+     * Post additional fields.
+     * You can add static fields to each log event in the following form:
+     * 
+     * <pre>
+     * quarkus.log.handler.gelf.additional-field.field1.value=value1
+     * quarkus.log.handler.gelf.additional-field.field1.type=String
+     * </pre>
      */
     @ConfigItem
-    Optional<String> additionalFields;
-
-    /**
-     * Type specification for additional and MDC fields.
-     * Supported types: String, long, Long, double, Double and discover (default if not specified, discover field type on
-     * parseability).
-     * E.g. `field=String,field2=double`.
-     */
-    @ConfigItem
-    Optional<String> additionalFieldsTypes;
+    @ConfigDocMapKey("field-name")
+    @ConfigDocSection
+    public Map<String, AdditionalFieldConfig> additionalField;
 
 }
--- a/extensions/logging-gelf/runtime/src/main/java/io/quarkus/logging/gelf/GelfLogHandlerRecorder.java
+++ b/extensions/logging-gelf/runtime/src/main/java/io/quarkus/logging/gelf/GelfLogHandlerRecorder.java
@@ -1,5 +1,6 @@
 package io.quarkus.logging.gelf;
 
+import java.util.Map;
 import java.util.Optional;
 import java.util.logging.Handler;
 
@@ -27,12 +28,27 @@ public class GelfLogHandlerRecorder {
         handler.setHost(config.host);
         handler.setPort(config.port);
         handler.setLevel(config.level);
-        if (config.additionalFields.isPresent()) {
-            handler.setAdditionalFields(config.additionalFields.get());
+
+        // handle additional fields
+        if (!config.additionalField.isEmpty()) {
+            StringBuilder additionalFieldsValue = new StringBuilder();
+            StringBuilder additionalFieldsType = new StringBuilder();
+            for (Map.Entry<String, AdditionalFieldConfig> additionalField : config.additionalField.entrySet()) {
+                if (additionalFieldsValue.length() > 0) {
+                    additionalFieldsValue.append(',');
+                }
+                additionalFieldsValue.append(additionalField.getKey()).append('=').append(additionalField.getValue().value);
+
+                if (additionalFieldsType.length() > 0) {
+                    additionalFieldsType.append(',');
+                }
+                additionalFieldsType.append(additionalField.getKey()).append('=').append(additionalField.getValue().type);
+            }
+
+            handler.setAdditionalFields(additionalFieldsValue.toString());
+            handler.setAdditionalFieldTypes(additionalFieldsType.toString());
         }
-        if (config.additionalFieldsTypes.isPresent()) {
-            handler.setAdditionalFieldTypes(config.additionalFieldsTypes.get());
-        }
+
         return new RuntimeValue<>(Optional.of(handler));
     }
 }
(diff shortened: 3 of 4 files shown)


extensions/logging-gelf/runtime/src/main/java/io/quarkus/logging/gelf/GelfConfig.java
```diff
@@ -61,4 +64,32 @@
...
+
+    /**
+     * Type specification for additional and MDC fields.
+     * Supported types: String, long, Long, double, Double and discover (default if not specified, discover field type on
+     * parseability).
+     * Eg. `field=String,field2=double`.
+     */
+    @ConfigItem
+    Optional<String> additionalFieldsTypes;
+
```
machi1990: Maybe this should be a map of config group?
 
Something like in pseudo code

```
@ConfigItem
Map<String, AdditionalField> additionalField;
```

And the `AdditionalField` class being 

```
@ConfigGroup
AdditionalField {
 @ConfigItem
 String value;

 @ConfigItem
 String type;
}

WDYT? @loicmathieu @gsmet

File: extensions/logging-gelf/runtime/src/main/java/io/quarkus/logging/gelf/GelfConfig.java [hunk 2]
```diff
@@ -61,4 +64,32 @@
...
+    /**
+     * The logging-gelf log level.
+     */
+    @ConfigItem(defaultValue = "ALL")
+    Level level;
+
+    /**
+     * Name of the Facility.
+     */
+    @ConfigItem(defaultValue = "jboss-logmanager")
```
machi1990: Is it possible to have an enum here?

machi1990:
Thank you @loicmathieu for this PR I have added two comments which are open for discussion.

extensions/logging-gelf/runtime/src/main/java/io/quarkus/logging/gelf/GelfConfig.java
loicmathieu: I don't think an enum will be OK here.
The facility shoud be the name of the logging system, that's why the default value is `jboss-logmanager` but user can override it to whatever they want (typical use case: the name of the application)

extensions/logging-gelf/runtime/src/main/java/io/quarkus/logging/gelf/GelfConfig.java
loicmathieu: I see your point, but I don't know if it will be easier to the user to configure them like this.

Current situation:
```
quarkus.log.handler.gelf.additional-fields=field1=value,field2=1
quarkus.log.handler.gelf.additional-fields-types=field1=String,field2=long
```

Your proposal:
```
quarkus.log.handler.gelf.additional-fields.field1.value=value
quarkus.log.handler.gelf.additional-fields.field1.type=String
quarkus.log.handler.gelf.additional-fields.field2.value=1
quarkus.log.handler.gelf.additional-fields.field2.type=long
```

machi1990: Thanks for the reply. 

With my proposal, the property will be in the generated doc. I think it also shows that the value and type go hand in hand (maybe that is pretty clear to some user but surely not all users). 

What happens if we provide fields value without the type or vice-versa with the current proposal?

extensions/logging-gelf/runtime/src/main/java/io/quarkus/logging/gelf/GelfConfig.java
machi1990: Oh okay, thanks for the explanation.

extensions/logging-gelf/runtime/src/main/java/io/quarkus/logging/gelf/GelfConfig.java
loicmathieu: - field value without type: default type will be used (discover: auto mode based on parseability of the field)
- field type without value: no idea

machi1990: Great, thanks. 


> * field value without type: default type will be used (discover: auto mode based on parseability of the field)

Great thanks, just saw it on the Javadoc documentation. 

> * field type without value: no idea

Okay, can we have a test? If it is something not desirable, I think we should throw a `ConfigurationException`.  

Let us wait and see what others think of the two proposals or if we can find another option to make these configuration knobs more appealing to the user.

extensions/logging-gelf/runtime/src/main/java/io/quarkus/logging/gelf/GelfConfig.java
```diff
@@ -61,4 +64,32 @@
...
+    String facility;
+
+    /**
+     * Post additional fields. Eg. `fieldName=Value,field2=value2`.
```
gsmet:
```suggestion
     * Post additional fields. E.g. `fieldName1=value1,fieldName2=value2`.
```

extensions/logging-gelf/runtime/src/main/java/io/quarkus/logging/gelf/GelfConfig.java
```diff
@@ -61,4 +64,32 @@
...
+     */
+    @ConfigItem
+    Optional<String> additionalFields;
+
+    /**
+     * Type specification for additional and MDC fields.
+     * Supported types: String, long, Long, double, Double and discover (default if not specified, discover field type on
+     * parseability).
+     * Eg. `field=String,field2=double`.
```
gsmet:
```suggestion
     * E.g. `field=String,field2=double`.
```

gsmet (changes requested):
I agree we should experiment with Manyanda's proposal and see how it goes. It looks cleaner and easier to understand.

Also, if the type is optional, it reads very well.

loicmathieu:
@gsmet  @machi1990 I review the way additional fields are configured.
If it's OK for you, then I'll squashed the commits

extensions/logging-gelf/runtime/src/main/java/io/quarkus/logging/gelf/GelfConfig.java
```diff
@@ -61,4 +64,28 @@
...
+    /**
+     * Name of the Facility.
+     */
+    @ConfigItem(defaultValue = "jboss-logmanager")
+    public String facility;
+
+    /**
+     * Post additional fields.
+     * You can add static fields to each log event in the following form:
+     * `̀`
```
machi1990:
```suggestion
     * ```
```

extensions/logging-gelf/runtime/src/main/java/io/quarkus/logging/gelf/GelfLogHandlerRecorder.java
```diff
@@ -26,6 +27,31 @@
...
+            StringBuilder additionalFieldsValue = new StringBuilder();
+            StringBuilder additionalFieldsType = new StringBuilder();
+            for (Map.Entry<String, AdditionalFieldConfig> additionalField : config.additionalField.entrySet()) {
+                if (additionalFieldsValue.length() > 0) {
+                    additionalFieldsValue.append(',');
+                }
+                additionalFieldsValue.append(additionalField.getKey()).append('=').append(additionalField.getValue().value);
+
+                additionalField.getValue().type.ifPresent(
+                        type -> {
```
machi1990: We try to avoid lambdas in runtime module

machi1990:
> @gsmet @machi1990 I review the way additional fields are configured.
> If it's OK for you, then I'll squashed the commits

I have added two comments, LGTM once they are addressed.

extensions/logging-gelf/runtime/src/main/java/io/quarkus/logging/gelf/GelfConfig.java
loicmathieu: done

extensions/logging-gelf/runtime/src/main/java/io/quarkus/logging/gelf/GelfLogHandlerRecorder.java
loicmathieu: done

extensions/logging-gelf/runtime/src/main/java/io/quarkus/logging/gelf/AdditionalFieldConfig.java
```diff
@@ -0,0 +1,26 @@
...
+@ConfigGroup
+public class AdditionalFieldConfig {
+    /**
+     * Additional field value.
+     */
+    @ConfigItem
+    public String value;
+
+    /**
+     * Additional field type specification .
```
gsmet:
```suggestion
     * Additional field type specification.
```

extensions/logging-gelf/runtime/src/main/java/io/quarkus/logging/gelf/AdditionalFieldConfig.java
```diff
@@ -0,0 +1,26 @@
...
+     * Supported types: String, long, Long, double, Double and discover.
+     * Discover is the default if not specified, it discovers field type based on parseability.
+     */
+    @ConfigItem
+    public Optional<String> type;
```
gsmet: Maybe make it a `String` and make `discover` the `defaultValue`? This way, it's all auto-documented?

I was wondering if we could use an enum for that but I'm pretty sure the magic case handling we added will conflict with having `Long` and `long`.

extensions/logging-gelf/runtime/src/main/java/io/quarkus/logging/gelf/GelfConfig.java
```diff
@@ -61,4 +64,28 @@
...
     public String timestampPattern;
+
+    /**
+     * The logging-gelf log level.
+     */
+    @ConfigItem(defaultValue = "ALL")
+    public Level level;
+
+    /**
+     * Name of the Facility.
```
gsmet:
```suggestion
     * Name of the facility.
```

extensions/logging-gelf/runtime/src/main/java/io/quarkus/logging/gelf/GelfConfig.java
```diff
@@ -61,4 +64,28 @@
...
+    @ConfigItem(defaultValue = "jboss-logmanager")
+    public String facility;
+
+    /**
+     * Post additional fields.
+     * You can add static fields to each log event in the following form:
+     * ```
+     * quarkus.log.handler.gelf.additional-field.field1.value=value1
+     * quarkus.log.handler.gelf.additional-field.field1.type=String
+     * ```
```
gsmet: Markdown is not supported in Javadoc. Either use a `<pre>` or make it Asciidoc and use `@asciidoclet`.

extensions/logging-gelf/runtime/src/main/java/io/quarkus/logging/gelf/GelfConfig.java
```diff
@@ -61,4 +64,28 @@
...
+     */
+    @ConfigItem
+    public Map<String, AdditionalFieldConfig> additionalField;
```
gsmet:
```suggestion
    public Map<String, AdditionalFieldConfig> additionalFields;
```
?

extensions/logging-gelf/runtime/src/main/java/io/quarkus/logging/gelf/GelfConfig.java
```diff
@@ -61,4 +64,28 @@
...
+
+    /**
+     * Post additional fields.
+     * You can add static fields to each log event in the following form:
+     * ```
+     * quarkus.log.handler.gelf.additional-field.field1.value=value1
+     * quarkus.log.handler.gelf.additional-field.field1.type=String
+     * ```
+     */
+    @ConfigItem
```
gsmet: Please add a `@ConfigDocMapKey("field-name")` annotation.

extensions/logging-gelf/runtime/src/main/java/io/quarkus/logging/gelf/GelfLogHandlerRecorder.java
```diff
@@ -26,6 +28,33 @@
...
+
+        // handle additional fields
+        if (!config.additionalField.isEmpty()) {
+            StringBuilder additionalFieldsValue = new StringBuilder();
+            StringBuilder additionalFieldsType = new StringBuilder();
+            for (Map.Entry<String, AdditionalFieldConfig> additionalField : config.additionalField.entrySet()) {
+                if (additionalFieldsValue.length() > 0) {
+                    additionalFieldsValue.append(',');
+                }
+                additionalFieldsValue.append(additionalField.getKey()).append('=').append(additionalField.getValue().value);
```
gsmet: There's no escaping needed?

gsmet (changes requested):
I added a couple of suggestions.

extensions/logging-gelf/runtime/src/main/java/io/quarkus/logging/gelf/AdditionalFieldConfig.java
loicmathieu: I'm not comfortable with enum because of `Long` and `long`, or we can choose to only support one of them as both will ends up the same I think ...

If I have no value, I didn't emit any type configuration at all. If I add a default value the default type configuration will be created for each additional type. I have no string opinion on this ... as you prefere ...

extensions/logging-gelf/runtime/src/main/java/io/quarkus/logging/gelf/GelfLogHandlerRecorder.java
loicmathieu: In all examples that I saw, there was no escaping.
I don't know what will happens if a user put a special character in an additional field.
As it comes from a configuration file, there is no security risk so I propose to let it this way

extensions/logging-gelf/runtime/src/main/java/io/quarkus/logging/gelf/AdditionalFieldConfig.java
gsmet: I think I prefer having a `defaultValue`.

extensions/logging-gelf/runtime/src/main/java/io/quarkus/logging/gelf/GelfConfig.java
loicmathieu: Updated to use pre

extensions/logging-gelf/runtime/src/main/java/io/quarkus/logging/gelf/GelfConfig.java
loicmathieu: I was not aware of this annotation, updated to use it.

extensions/logging-gelf/runtime/src/main/java/io/quarkus/logging/gelf/AdditionalFieldConfig.java
loicmathieu: Done

extensions/logging-gelf/runtime/src/main/java/io/quarkus/logging/gelf/GelfConfig.java
loicmathieu: I also add a `@ConfigDocSection` it adds clarity to the configuration documentation.

loicmathieu:
@gsmet I think I apply all your suggestions now.

gsmet (approved):

gsmet:
Cool, let's wait for CI and merge.

PR #6358, opened 2019-12-30
Fix elytron properties perf issues

loicmathieu:
During some performance testing, I discover, thanks to profiler CPU flamegraph, that the Elytron Properties implementation uses a lot of CPU cycles to create the `WilldFlyElytronProvider`.

It apears that the security provider supplier is called on each request calls, and triggers the creation of a new `WilldFlyElytronProvider`.

Creating a unique `WilldFlyElytronProvider` and returing it from the supplier instead of creating it each time fix the issue.

### DIFF
--- a/extensions/elytron-security-properties-file/runtime/src/main/java/io/quarkus/elytron/security/runtime/ElytronPropertiesFileRecorder.java
+++ b/extensions/elytron-security-properties-file/runtime/src/main/java/io/quarkus/elytron/security/runtime/ElytronPropertiesFileRecorder.java
@@ -44,6 +44,8 @@ import io.quarkus.runtime.annotations.Recorder;
 public class ElytronPropertiesFileRecorder {
     static final Logger log = Logger.getLogger(ElytronPropertiesFileRecorder.class);
 
+    private static final Provider[] PROVIDERS = new Provider[] { new WildFlyElytronProvider() };
+
     /**
      * Load the user.properties and roles.properties files into the {@linkplain SecurityRealm}
      *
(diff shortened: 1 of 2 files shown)


extensions/elytron-security-properties-file/runtime/src/main/java/io/quarkus/elytron/security/runtime/ElytronPropertiesFileRecorder.java
```diff
@@ -170,9 +170,11 @@ public void run() {
         SecurityRealm realm = LegacyPropertiesSecurityRealm.builder()
                 .setDefaultRealm(config.realmName)
                 .setProviders(new Supplier<Provider[]>() {
+                    Provider[] providers = new Provider[] { new WildFlyElytronProvider() };
+
                     @Override
                     public Provider[] get() {
-                        return new Provider[] { new WildFlyElytronProvider() };
+                        return providers;
```
gsmet: Maybe you can make the whole supplier static and reuse it at both locations?

gsmet (changes requested):
Good catch. I added a comment to further improve the situation.

gsmet:
Also, can you check in the other security implementation that we don't have this pattern?

extensions/elytron-security-properties-file/runtime/src/main/java/io/quarkus/elytron/security/runtime/ElytronPropertiesFileRecorder.java
loicmathieu: done

loicmathieu:
@gsmet I add the same performance optimization to the JDBC implementation.

OAuth2 and OIDC didn't suffers from the same issue.

loicmathieu:
@gsmet I also review the LDAP and JPA security extensions currently under review, LDAP has the same issue (I add a review comment), JPA doesn't seems to have it (not 100% sure as it's a complex extension and I just have a quick look).

stuartwdouglas (approved):

gsmet (approved):

gsmet:
Merged, thanks!

PR #6361, opened 2019-12-30
Improve Hibernate Validator localization support with RESTEasy

### LINKED ISSUE #5908: Improve Hibernate Validator localization support with RESTEasy
We have a very low level of integration between RESTEasy and HV on the localization side.

HV supports localization, RESTEasy has support for dealing with the `Accept-Language` header (not sure how much support there is, we'll need to figure that out).

But... they are not coupled at all. Meaning you will always have the messages in the default locale.

The idea would be to take this into account in the `JaxrsEndPointValidationInterceptor` and build a specific `Validator` there, passing it a specific `MessageInterpolator` which would have the locale determined from:
- some config: for now, let's create a `HibernateValidatorBuildTimeConfig` with a `locales` property. We might generalize that but that will be good enough. This set of locales will need to be passed to the ValidatorFactory here: https://github.com/quarkusio/quarkus/blob/master/extensions/hibernate-validator/runtime/src/main/java/io/quarkus/hibernate/validator/runtime/HibernateValidatorRecorder.java#L43 . Right now, we only pass the default one so if you try to use another locale, you will get an error as reported here https://github.com/quarkusio/quarkus/issues/5882.
- the headers of the request. I think we can get them from `ResteasyContext`. If you call `ResteasyContext.getContextDataMap()`, there should be some headers there, we need to check they are properly populated for the given method that is currently called. Playing with the `integration-tests/hibernate-validator/` should help.
The `MessageInterpolator` would be a simple wrapper around the existing `MessageInterpolator` and would just resolve the locale then pass it to the existing interpolator.

We usually have a way to define a MessageInterpolator specific to a given context via:
```
validatorFactory.usingContext()
    .messageInterpolator(specificMessageInterpolator)
    .getValidator();
```
The issue is that... I was a bit too conservative in HV and made it forbidden in `PredefinedScopeValidatorContextImpl`. I think we could allow it and it was just me being too conservative but we will need to check if it works correctly (that means that you will need to get HV master, adjust it and recompile).

I don't know if there is a proper `Accept-Language` resolution algorithm in RESTEasy, easily accessible for us. It would really nice better if we didn't have to implement our one. @asoldano any pointers?

I think this should be done in several small steps:
- support several locales in HV via configuration
- see if we can safely override the MessageInterpolator in the PredefinedScopeVF. Let's not prepare the PR for HV right away, we can do that at the end.
- see if we can resolve the locale from the RESTEasy headers and use that locale for HV.

In the end, we might make a RESTEasy-aware message interpolator the default one if we can get to a working solution. If we do that, all the HV calls would be localized by default if they are done in the context of a RESTEasy call. But that's really something for the end.

irenakezic:
Improve Hibernate Validator localization support with RESTEasy
Fixes [#5908](https://github.com/quarkusio/quarkus/issues/5908)


machi1990:
@irenakezic thanks for the PR. Looks like there is a compilation failure

```
 /home/vsts/work/1/s/extensions/hibernate-validator/runtime/src/main/java/io/quarkus/hibernate/validator/runtime/ResteasyContextLocaleResolver.java:[19,55] cannot find symbol
[ERROR]   symbol: class LocaleResolver
[ERROR] /home/vsts/work/1/s/extensions/hibernate-validator/runtime/src/main/java/io/quarkus/hibernate/validator/runtime/ResteasyContextLocaleResolver.java:[22,27] cannot find symbol
[ERROR]   symbol:   class LocaleResolverContext
[ERROR]   location: class io.quarkus.hibernate.validator.runtime.ResteasyContextLocaleResolver
[ERROR] /home/vsts/work/1/s/extensions/hibernate-validator/runtime/src/main/java/io/quarkus/hibernate/validator/runtime/HibernateValidatorRecorder.java:[20,56] package org.hibernate.validator.spi.messageinterpolation does not exist
```

Can you check? Thanks :-)

irenakezic:
@machi1990  This is expected as there is a dependency on new HV functionalities which will be added in the next Hibernate Validator release (6.1.1). I created a PR anyway as instructed by @gsmet in [#5908](https://github.com/quarkusio/quarkus/issues/5908) to enable us to collaborate further on this.

machi1990:
> @machi1990 This is expected as there is a dependency on new HV functionalities which will be added in the next Hibernate Validator release (6.1.1). I created a PR anyway as instructed by @gsmet in [#5908](https://github.com/quarkusio/quarkus/issues/5908) to enable us to collaborate further on this.

Oh sure, it is a draft PR.  I just realised that after reading your discussion in the issue. Please,  ignore my comment.

gsmet:
@irenakezic sorry it took so long but I released HV 6.1.1.Final yesterday.

I will wait for CI and merge.

gsmet (approved):

PR #6365, opened 2019-12-31
WIP - Support transaction inside MongoDB with Panache

loicmathieu:
Basic transaction support inside MongoDB with Panache.

I implemented a basic TransactionManager (not using Narayana JTA), maybe it's a too naive implementation but it works ;)

It needs the latests version of the MongoDB driver, so it needs the PR #6347 to be merged first.

This is a draft PR to gather feedback on the API design and the implementation choices, please don't go into deep review yet.


FroMage:
So, why not use Narayana? It supports async transactions, which you do not. You should be able to integrate with it too.

https://github.com/FroMage/quarkus/commit/663df8f9692f37e445c67b9f7f75ecd9c28a851d#diff-99d18a6d69c7c51a2e7ed99ed888615fR45 has all the bits:

```java
    @Inject
    TransactionSynchronizationRegistry tsr;

    @Inject
    TransactionManager transactionManager;

MongoTransaction getMongoTransaction() {
        if (isInTransaction()) {
            // already have a transaction
            MongoTransaction tr = (MongoTransaction) tsr.getResource(transactionKey);
            if (tr != null) {
                return tr;
            }
            tr = new MongoTransaction();// whatever
                tsr.putResource(transactionKey, tr);
                tsr.registerInterposedSynchronization(new Synchronization() {
                    @Override
                    public void beforeCompletion() {
                    }

                    @Override
                    public void afterCompletion(int i) {
                        try {
                            // FIXME: async?
                            // FIXME: statuses?
                            if(transactionManager.getStatus() == Status.STATUS_ROLLEDBACK)
                                tr.rollback();
                            else
                                tr.commit();
                        } catch (SystemException e) {
                            // TODO Auto-generated catch block
                            e.printStackTrace();
                        }
                        tr.close();
                    }
                });
                return tr;
        } else {
          // no transaction case
          return null;
        }
    }

    private boolean isInTransaction() {
        try {
            switch (transactionManager.getStatus()) {
                case Status.STATUS_ACTIVE:
                case Status.STATUS_COMMITTING:
                case Status.STATUS_MARKED_ROLLBACK:
                case Status.STATUS_PREPARED:
                case Status.STATUS_PREPARING:
                    return true;
                default:
                    return false;
            }
        } catch (Exception e) {
        }
    }
```

loicmathieu:
> So, why not use Narayana?

It will make the mongodb extension dependents on narayana and transaction-api. I found it a little too much but I can give it a try.

What will happens if a user defines a transaction with a mix of JPA and MongoDB, will it works OOTB ?

FroMage:
> What will happens if a user defines a transaction with a mix of JPA and MongoDB, will it works OOTB ?

That is the idea.

FroMage:
Hold on: I don't see any commit/rollback in your PR. Don't those `ClientSession` objects support rollbacks?

loicmathieu:
> Hold on: I don't see any commit/rollback

`clientSession.withTransaction(txnBody, txnOptions.build());` will take care of committing the transaction if everything is successfull and rolling back otherwise.

We can deal with commit and rollback by ourselves if we want but I use the _transaction body_ implementation that is the most easy to use.

loicmathieu:
Superseded by #7222

PR #6370, opened 2020-01-02
docs: fix project artifact id for jackson

kdnakt:
According to the following command `cd rest-json-quickstart`,  project artifact id should be `rest-json-quickstart` rather than `rest-json`.


geoand (approved):
Good catch, thanks!

PR #6373, opened 2020-01-02
Ensure that MySQL integration test works with Docker Maven plugin

### LINKED ISSUE #6372: integration-tests/jpa-mysql test fails against docker instance
integration-tests/jpa-mysql test fails against docker instance - `mvn clean install -Dtest-mysql -Ddocker -pl integration-tests/jpa-mysql`

From the stacktrace the most important message seems `Caused by: java.io.EOFException: Can not read response from server. Expected to read 4 bytes, read 0 bytes before connection was unexpectedly lost.`

Tests are passing when running against already running database - `mvn clean install -Dtest-mysql -pl integration-tests/jpa-mysql`

Seems that `-Ddocker` way doesn't wait long enough or there is problem with exposed ports.
Running on macOS 10.15.2, OpenJDK 64-Bit GraalVM CE 19.3.0.

geoand:
Fixes: #6372

### DIFF
--- a/integration-tests/jpa-mysql/pom.xml
+++ b/integration-tests/jpa-mysql/pom.xml
@@ -195,14 +195,12 @@
                                         <!-- Speed things up a bit by not actually flushing writes to disk -->
                                         <tmpfs>/var/lib/mysql</tmpfs>
                                         <wait>
-                                            <!-- good docs found at: http://dmp.fabric8.io/#build-healthcheck -->
-                                            <tcp>
-                                                <mode>mapped</mode>
-                                                <ports>
-                                                    <port>3306</port>
-                                                </ports>
-                                            </tcp>
-                                            <time>40000</time>
+                                            <!-- good docs found at: http://dmp.fabric8.io/#start-wait -->
+                                            <time>20000</time>
+                                            <!-- wait until MySQL is actually up by checking if mysqladmin can ping the server with specified username/password -->
+                                            <exec>
+                                                <postStart>mysqladmin ping -h localhost -u hibernate_orm_test -phibernate_orm_test</postStart>
+                                            </exec>
                                         </wait>
                                     </run>
                                 </image>


integration-tests/jpa-mysql/pom.xml
```diff
@@ -196,13 +196,11 @@
                                         <tmpfs>/var/lib/mysql</tmpfs>
                                         <wait>
                                             <!-- good docs found at: http://dmp.fabric8.io/#build-healthcheck -->
```
rsvoboda: This line seems no longer relevant,  `<postStart>mysqladmin ...` is not mentioned there

rsvoboda (commented):
Test is now running with `-Ddocker`.

The only downside is quite long wait for mysqladmin based check - 30 seconds in this case (see lines for time 10:37:48.870 and 10:38:18.417):
```
10:37:38.432 MySQL:Initializing database
....
10:37:48.870 MySQL:2020-01-02T09:37:48.869831Z 0 [System] [MY-011323] [Server] X Plugin ready for connections. Socket: '/var/run/mysqld/mysqlx.sock' bind-address: '::' port: 33060
10:38:18.417 MySQL:mysqladmin: [Warning] Using a password on the command line interface can be insecure.
10:38:18.425 MySQL:mysqld is alive
```

it's connected with `<time>40000</time>` settings, in my case MySQL was initiated within 10s so I could safely reduce the time to 20000.

integration-tests/jpa-mysql/pom.xml
geoand: Good catch!, I'll update it

geoand:
@rsvoboda yes indeed the wait time is long... We can of course reduce it (and I initially did have it at 20 seconds and then opted to switch back just in case some machine is slow to start the docker container).

But I guess I can move it back to 20 seconds and if we have problems we can always increase it.

gsmet:
> But I guess I can move it back to 20 seconds and if we have problems we can always increase it.

Yes let's try that.

geoand:
@rsvoboda PR updated

rsvoboda (approved):

PR #6375, opened 2020-01-02
Use mongodb-driver-sync instead of mongo-java-driver

loicmathieu:
Today Quarkus uses two MongoDB libraries:

- mongo-java-driver: this is an uberjar of mongodb-driver-sync, mongodb-driver-async, mongodb-driver-core and bson
- mongodb-driver-reactivestreams: it have dependencies to mongodb-driver-async, mongodb-driver-core and bson

So a lot of mongodb classes are duplicated, we also include some legacy classes that are replaced and should not be used anymore.

This PR replace the usage of mongo-java-driver by  mongodb-driver-sync.

It breaks some compatibilities because `com.mongodb.MongoClient` is no more accessible as it is replaced by `com.mongodb.client.MongoClient`. As the Quarkus code only provides bean instance of `com.mongodb.client.MongoClient` it should be safe to do it.

Only some code related to codec didn't works anymore as the default codec registry should be obtain from `MongoClientSettings` inside of from the old `MongoClient`.

A dependency management has been added to `mongodb-client-legacy` that includes the classes from mongo-java-driver that are not present in other libraries. User can add this dependency to have access to the old classes.

We should have a note about this small breaking change inside the release note (or we can add some WARNING in the guide) explaining the changes needed to be done on Codec code and the possibility to add the legacy driver if needed (not recommended).

### DIFF
--- a/docs/src/main/asciidoc/mongodb.adoc
+++ b/docs/src/main/asciidoc/mongodb.adoc
@@ -331,7 +331,7 @@ More information in the codec documentation: https://mongodb.github.io/mongo-jav
 ----
 package org.acme.rest.json.codec;
 
-import com.mongodb.MongoClient;
+import com.mongodb.MongoClientSettings;
 import org.acme.rest.json.Fruit;
 import org.bson.*;
 import org.bson.codecs.Codec;
@@ -346,7 +346,7 @@ public class FruitCodec implements CollectibleCodec<Fruit> {
     private final Codec<Document> documentCodec;
 
     public FruitCodec() {
-        this.documentCodec = MongoClient.getDefaultCodecRegistry().get(Document.class);
+        this.documentCodec = MongoClientSettings.getDefaultCodecRegistry().get(Document.class);
     }
 
     @Override
@@ -478,6 +478,21 @@ to validate the connection to the cluster.
 So when you access the `/health/ready` endpoint of your application you will have information about the connection validation status.
 
 This behavior can be disabled by setting the `quarkus.mongodb.health.enabled` property to `false` in your `application.properties`.
+
+== The legacy client
+
+We don't include the legacy MongoDB client by default. It contains the now retired MongoDB Java API (DB, DBCollection,... )
+and the `com.mongodb.MongoClient` that is now superseded by `com.mongodb.client.MongoClient`.
+
+If you want to use the legacy API, you need to add the following dependency to your `pom.xml`:
+
+[source,xml]
+----
+<dependency>
+    <groupId>org.mongodb</groupId>
+    <artifactId>mongodb-driver-legacy</artifactId>
+</dependency>
+----
 
 == Building a native executable
 
--- a/extensions/mongodb-client/runtime/pom.xml
+++ b/extensions/mongodb-client/runtime/pom.xml
@@ -13,6 +13,7 @@
     <artifactId>quarkus-mongodb-client</artifactId>
     <name>Quarkus - MongoDB Client - Runtime</name>
     <description>An imperative and reactive client for MongoDB</description>
+
     <dependencies>
 
         <dependency>
@@ -29,7 +30,7 @@
         </dependency>
         <dependency>
             <groupId>org.mongodb</groupId>
-            <artifactId>mongo-java-driver</artifactId>
+            <artifactId>mongodb-driver-sync</artifactId>
         </dependency>
         <dependency>
             <groupId>org.mongodb</groupId>
--- a/integration-tests/mongodb-client/src/main/java/io/quarkus/it/mongodb/BookCodec.java
+++ b/integration-tests/mongodb-client/src/main/java/io/quarkus/it/mongodb/BookCodec.java
@@ -9,14 +9,14 @@ import org.bson.codecs.CollectibleCodec;
 import org.bson.codecs.DecoderContext;
 import org.bson.codecs.EncoderContext;
 
-import com.mongodb.MongoClient;
+import com.mongodb.MongoClientSettings;
 
 public class BookCodec implements CollectibleCodec<Book> {
 
     private final Codec<Document> documentCodec;
 
     public BookCodec() {
-        this.documentCodec = MongoClient.getDefaultCodecRegistry().get(Document.class);
+        this.documentCodec = MongoClientSettings.getDefaultCodecRegistry().get(Document.class);
     }
 
     @Override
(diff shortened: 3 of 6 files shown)


loicmathieu:
Note: PR #6347 will need to be updated after this one merged

extensions/mongodb-client/runtime/pom.xml
```diff
@@ -13,6 +13,17 @@
     <artifactId>quarkus-mongodb-client</artifactId>
     <name>Quarkus - MongoDB Client - Runtime</name>
     <description>An imperative and reactive client for MongoDB</description>
+
+    <dependencyManagement>
+        <dependencies>
+            <dependency>
+                <groupId>org.mongodb</groupId>
+                <artifactId>mongodb-driver-async</artifactId>
+                <version>3.10.2</version>
```
cescoffier: It should be declared in the bom.

integration-tests/mongodb-client/src/main/java/io/quarkus/it/mongodb/BookCodec.java
```diff
@@ -9,14 +9,14 @@
...
-import com.mongodb.MongoClient;
+import com.mongodb.MongoClientSettings;
 
 public class BookCodec implements CollectibleCodec<Book> {
 
     private final Codec<Document> documentCodec;
 
     public BookCodec() {
-        this.documentCodec = MongoClient.getDefaultCodecRegistry().get(Document.class);
+        this.documentCodec = MongoClientSettings.getDefaultCodecRegistry().get(Document.class);
```
cescoffier: This might be considered as a breaking change as the application code need to be updated.

cescoffier (changes requested):

extensions/mongodb-client/runtime/pom.xml
loicmathieu: It was a leftover: removed

integration-tests/mongodb-client/src/main/java/io/quarkus/it/mongodb/BookCodec.java
loicmathieu: Yes, it is described in the PR and I added the breaking change label.
The extension is still in preview so I think we can have this kind of small breaking change but if youre against it I'm OK.
In fact, it's not really a breaking change as the user can include the legacy driver that still contains this class.

extensions/mongodb-client/runtime/pom.xml
```diff
@@ -29,11 +30,25 @@
...
+            <groupId>org.mongodb</groupId>
+            <artifactId>mongodb-driver-async</artifactId>
         </dependency>
         <dependency>
             <groupId>org.mongodb</groupId>
             <artifactId>mongodb-driver-reactivestreams</artifactId>
+            <exclusions>
+                <!--
+                    Exclude mongodb-driver-async as it shares dependencies with mongodb-driver-sync so we need
+                    to have both versions aligned which is not the case if it is a transitive dependency of mongodb-driver-reactivestreams.
```
gastaldi: If you declare the `mongo-driver-async` in the BOM I think you can remove this exclusion

loicmathieu: Yes, this was again a leftover of my attemp to fix the dependencies issue

loicmathieu:
@cescoffier I can add a warning inside the mongodb user guide that we don't include the legacy API of MongoDB (com.mongodb.MongoClient and all the DB* classes) and that a user can still add the dependency org.mongodb:mongodb-driver-legacy to have access to it.

The fact that the exemples previously get the codec registry from `MongoClient` and it now needs to retrieve it from `MongoClientSettings` should be described inside the release note as it is a dropin replacement I don't think it is a big issue.

cescoffier:
@loicmathieu +1, thanks!

loicmathieu:
@cescoffier I added a note on the mongodb guide about the legacy client.

gastaldi:
This PR looks good to be in 1.2.0, any considerations @cescoffier ?

cescoffier (approved):

docs/src/main/asciidoc/mongodb.adoc
```diff
@@ -479,6 +479,21 @@ So when you access the `/health/ready` endpoint of your application you will hav
 
 This behavior can be disabled by setting the `quarkus.mongodb.health.enabled` property to `false` in your `application.properties`.
 
+== The legacy client
+
+We didn't include the legacy MongoDB client per default. It contains the old MongoDB Java API (DB, DBCollection,... )
```
gsmet:
```suggestion
We don't include the legacy MongoDB client by default. It contains the now retired MongoDB Java API (DB, DBCollection,... )
```

docs/src/main/asciidoc/mongodb.adoc
```diff
@@ -479,6 +479,21 @@ So when you access the `/health/ready` endpoint of your application you will hav
...
+and the `com.mongodb.MongoClient` that is now superseded by `com.mongodb.client.MongoClient`.
+
+If you want to use the legacy API, you need to include the following dependency inside your `pom.xml`.
```
gsmet:
```suggestion
If you want to use the legacy API, you need to add the following dependency to your `pom.xml`:
```

gsmet (changes requested):
Added a few comments on the doc addition.

docs/src/main/asciidoc/mongodb.adoc
```diff
@@ -346,7 +346,7 @@ public class FruitCodec implements CollectibleCodec<Fruit> {
     private final Codec<Document> documentCodec;
 
     public FruitCodec() {
-        this.documentCodec = MongoClient.getDefaultCodecRegistry().get(Document.class);
+        this.documentCodec = MongoClientSettings.getDefaultCodecRegistry().get(Document.class);
```
gsmet: Do we need a PR for the quickstart too (if there is one)?

loicmathieu: Yes, I'll send it today or tomorrow.

As we add some stuff to Hibernate with Panache and MongoDB with Panache lately I will check these two quickstart also to align them with the guides.

loicmathieu:
@gsmet Quickstart PR is created https://github.com/quarkusio/quarkus-quickstarts/pull/415

I think we're good to go :)

gsmet (approved):

gsmet:
We also probably need to prepare something for the migration guide. I'll try to work on that.

PR #6378, opened 2020-01-02
Add a quick workaround to get javadoc generation happy

gsmet:
This is ugly but, to properly build a Javadoc jar, we need at least one public class.

And we need a javadoc jar to be compliant with Central publishing rules.


gsmet:
This one is already in the 1.1 branch. I had to push it urgently when I released 1.1.0.

geoand (approved):

PR #6380, opened 2020-01-02
Upgrade Jackson to 2.10.1

gsmet:
This is a draft PR as we need to discuss it (and I want CI to check it).

The current situation is uncomfortable:
- Vert.x is still based on 2.9 (but we don't really know it 2.10 would break it);
- RESTEasy 4.4.2 uses 2.10-specific Jackson stuff and we need to upgrade to it to fix a few issues.

Thus, I'm wondering if we should take the risk to upgrade to Jackson 2.10.

/cc @geoand @cescoffier

### DIFF
--- a/bom/runtime/pom.xml
+++ b/bom/runtime/pom.xml
@@ -76,7 +76,7 @@
         <!-- What we actually depend on for the annotations, as latest Graal is not available in Maven fast enough: -->
         <graal-sdk.version>19.3.0</graal-sdk.version>
         <gizmo.version>1.0.0.Final</gizmo.version>
-        <jackson.version>2.9.10.20191020</jackson.version>
+        <jackson.version>2.10.1</jackson.version>
         <commons-logging-jboss-logging.version>1.0.0.Final</commons-logging-jboss-logging.version>
         <commons-lang3.version>3.9</commons-lang3.version>
         <commons-codec.version>1.13</commons-codec.version>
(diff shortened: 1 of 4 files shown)


geoand:
Yeah we need to get on this version as soon as possible...

geoand:
I am seeing: 

```
[WARNING] Rule 2: org.apache.maven.plugins.enforcer.BannedDependencies failed with message:
Found Banned Dependency: javax.xml.bind:jaxb-api:jar:2.3.1
Found Banned Dependency: javax.activation:javax.activation-api:jar:1.2.0
Use 'mvn dependency:tree' to locate the source of the banned dependencies
```

gsmet:
Hum. I tested with JDK 8. Looks like with JDK 11, additional artifacts are coming. I'll retest with JDK 11.

geoand:
Yeah CI is also passing with Java 8

gastaldi:
And here comes another micro patch: https://twitter.com/fasterxml/status/1212930048618950656?s=19

gsmet:
@geoand it should be fixed now.

geoand:
Excellent, thanks!

geoand:
I personally think we need this sooner rather than later however I have no idea what impact it has on Vert.x

gsmet:
@cescoffier WDYT? I don't know if there's a way to quickly check that Vert.x doesn't have an issue with Jackson 2.10. CI is green but not sure our test coverage is sufficient.

cescoffier:
Should be fine. Go ahead.

gsmet:
I promoted the PR. Now I need a review :).

geoand (approved):
LGTM!

Thanks for taking care of it!

PR #6384, opened 2020-01-02
Properly initialize Maven profile options during bootstrap

### LINKED ISSUE #6383: Using mvn package -Pnative instead of -Dnative, leads to CCE
**Describe the bug**
When executing `./mvnw package -Pnative` against a newly generated project, instead of a getting a native image built, we get the following error:

```
org.junit.jupiter.api.extension.TestInstantiationException: TestInstanceFactory [io.quarkus.test.junit.QuarkusTestExtension] failed to instantiate test class [org.acme.GreetingResourceTest]: java.lang.String cannot be cast to [Ljava.lang.String;
Caused by: java.lang.ClassCastException: java.lang.String cannot be cast to [Ljava.lang.String;
```

This only happens with `master`, Quarkus `1.1.0.Final` works fine.

The issue was originally found by @aureamunoz when she was working on some Quarkus material.

There is a workaround which is to use `-Dnative`, but we should fix this nonetheless since using profiles is currently broken.

geoand:
Fixes: #6383

### DIFF
--- a/independent-projects/bootstrap/core/src/main/java/io/quarkus/bootstrap/resolver/maven/options/BootstrapMavenOptionsParser.java
+++ b/independent-projects/bootstrap/core/src/main/java/io/quarkus/bootstrap/resolver/maven/options/BootstrapMavenOptionsParser.java
@@ -20,7 +20,7 @@ public class BootstrapMavenOptionsParser {
         final Map<String, Object> map = new HashMap<>();
         put(cmdLine, map, CLIManager.ALTERNATE_USER_SETTINGS);
         put(cmdLine, map, CLIManager.ALTERNATE_GLOBAL_SETTINGS);
-        put(cmdLine, map, CLIManager.ACTIVATE_PROFILES);
+        put(map, String.valueOf(CLIManager.ACTIVATE_PROFILES), cmdLine.getOptionValues(CLIManager.ACTIVATE_PROFILES));
 
         putBoolean(cmdLine, map, CLIManager.SUPRESS_SNAPSHOT_UPDATES);
         putBoolean(cmdLine, map, CLIManager.UPDATE_SNAPSHOTS);


independent-projects/bootstrap/core/src/main/java/io/quarkus/bootstrap/resolver/maven/options/BootstrapMavenOptions.java
```diff
@@ -90,7 +90,14 @@ public String getOptionValue(String name) {
 
     public String[] getOptionValues(String name) {
         final Object o = options.get(name);
-        return o == null ? null : (String[]) o;
+        if (o == null) {
+            return null;
+        } else if (o instanceof String[]) {
+            return (String[]) o;
+        } else if (o instanceof String) {
+            return new String[]{(String)o};
```
geoand: I didn't perform a `((String) o).split(",");` here because the code that uses the returned array does that already, but I am wondering if it should be done here in addition...

aloubyansky:
Would you mind changing it a bit? See my comment here https://github.com/quarkusio/quarkus/pull/6299#issuecomment-568099166

geoand:
Sure thing, I'll fix it ASAP

geoand:
@aloubyansky actually it is `getOptionValues` that is being called, however the call site is not the one you indicated, but this one: https://github.com/quarkusio/quarkus/blob/master/independent-projects/bootstrap/core/src/main/java/io/quarkus/bootstrap/resolver/maven/MavenRepoInitializer.java#L270

So I assume the fix is valid?

aloubyansky:
Yes, but it's not properly initialized. It's initialized with `cmdLine.getOptionValue(name)` instead of `getOptionValues`.

geoand:
Gotcha, I pushed the change

aloubyansky:
That's actually not what I meant. I meant this only for the active profiles. On which option did it fail in your case?

geoand:
It failed for me when using `-Pnative`.

With the initial fix the problem went away

aloubyansky:
Right, so that's profiles. They should always be initialized as String[]. Instead of creating an array on every get.

geoand:
Aha, I see what you mean

aloubyansky:
map.put(String.valueOf(CLIManager.ACTIVE_PROFILES), cmdLine.getOptionValues(CLIManager.ACTIVE_PROFILES))

geoand:
:+1:

geoand:
@aloubyansky PR updated.

Thanks for your patience :)

aloubyansky (approved):
Thanks a lot @geoand!

geoand:
You are welcome!

Issue #6385, opened 2020-01-02
ConfigProperty quarkus.profile always set to "prod"


zapho:
**Describe the bug**
When using `@ConfigProperty(name = "quarkus.profile") String profile`, `profile` is always set to "prod" regardless of the execution mode (dev, test, prod).

**Expected behavior**
When running quarkus:dev, `profile` should be "dev".
When running @QuarkusTest, `profile` should be "test".

**Actual behavior**
When running quarkus:dev, `profile` is "prod".
When running @QuarkusTest, `profile` is "prod".

**To Reproduce**
1. Write a class making of `@ConfigProperty(name = "quarkus.profile") String profile` and log the value of `profile`, e.g. 
```java @ApplicationScoped
public class MyBean {
    @ConfigProperty(name = "quarkus.profile")
    Optional<String> quarkusProfile;

   public void anyMethod() {
     LOG.debug("Quarkus profile is {}", quarkusProfile.orElse(null));
   }
}
```

2. Run `quarkus:dev` and use the application to trigger logging
3. Write a @QuarkusTest, run it

**Configuration**
n/a

**Screenshots**
n/a

**Environment (please complete the following information):**
 - Output of `uname -a` or `ver`: CYGWIN_NT-10.0-WOW MORLN256 2.6.1(0.305/5/3) 2016-12-16 11:50 i686 Cygwin
 - Output of `java -version`: java version "1.8.0_221"
Java(TM) SE Runtime Environment (build 1.8.0_221-b11)
Java HotSpot(TM) 64-Bit Server VM (build 25.221-b11, mixed mode)
 - GraalVM version (if different from Java): n/a JVM mode only
 - Quarkus version or git rev: 1.1.0.Final

**Additional context**
n/a

geoand:
You can retrieve profile information is via  `ProfileManager#getActiveProfile`. However I'll take a look at your issue and see if we can do something about it

geoand:
@gsmet @dmlloyd should we perhaps document the proper way to retrieve the profile? 
This is not the first time the problem has shown up...

zapho:
@geoand thanks for the tip, it works fine.

`ProfileManager#getActiveProfile` vs `@ConfigProperty(name = "quarkus.profile")`, it is indeed confusing from an external point of view...

geoand:
@zapho Yes, that is absolutely understandable and we should make the situation more clear

geoand:
Documentation has now been added in the configuration guide.

[...]

PR #6393, opened 2020-01-03
Document how to retrieve the active profile

geoand:
Has come up in chat and in issues (like #6385)


geoand:
The one reason I am not certain about this one is that the `ProfileManager` class is in the `runtime` package meaning it's not exactly guaranteed to be stable :)

gsmet:
Well, the API/internal separation in the core artifacts is blurry at best, unfortunately.

gsmet (approved):

geoand:
OK, thanks

PR #6397, opened 2020-01-03
Hibernate Validator dev mode tests added

irenakezic:
Hibernate Validator dev mode tests added:
- adding a constraint on a property of an existing bean
- adding a class constraint on an existing bean
- adding a method constraint on a CDI bean
- adding an entirely new constrained bean

### DIFF
--- a/extensions/hibernate-validator/deployment/pom.xml
+++ b/extensions/hibernate-validator/deployment/pom.xml
@@ -41,6 +41,16 @@
             <artifactId>assertj-core</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-resteasy-jsonb-deployment</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>io.rest-assured</groupId>
+            <artifactId>rest-assured</artifactId>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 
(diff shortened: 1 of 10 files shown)


irenakezic:
@gsmet please review

gsmet:
@irenakezic looks good, thanks.

I wonder if we should add a test adding a brand new constraint annotation and using it at the same time. That would check that the new contraints are correctly taken into account. WDYT?

irenakezic:
Agree, will add

irenakezic:
@gsmet I added one more test, please review

gsmet (approved):
I moved the tests to the `test` subpackage. Otherwise, it looks very good.

Let's wait for CI and merge. Thanks!

gsmet:
Merged, thanks!

extensions/hibernate-validator/deployment/pom.xml
```diff
@@ -41,6 +41,16 @@
             <artifactId>assertj-core</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-resteasy-jsonb-deployment</artifactId>
```
gsmet: @stuartwdouglas I reckon you will have to rebase and change that one to runtime in your ClassLoader PR?

stuartwdouglas: thanks for the heads up.

PR #6405, opened 2020-01-05
[WIP] Quarkus All Open compiler plugin

### LINKED ISSUE #1157: Create Kotlin compiler plugin for open classes
We should create a Kotlin compiler plugin that will stand in place of the all-open plugin and will contain pre-configured all the annotations that result in classes that must not be final (much the same way that the Kotlin spring does).
~~This will also solve [this](https://github.com/jbossas/protean-shamrock/pull/1155#discussion_r261866936) issue.~~

RotBolt:
resolves #1157


devtools/allopen/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -0,0 +1,5 @@
+distributionBase=GRADLE_USER_HOME
+distributionPath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-5.2.1-bin.zip
```
gsmet: Shouldn't we use Gradle 6 here?

devtools/allopen/build.gradle
```diff
@@ -0,0 +1,48 @@
+plugins {
+    id 'org.jetbrains.kotlin.jvm' version '1.3.61'
+    id "org.jetbrains.kotlin.kapt" version "1.3.61"
+    id 'com.gradle.plugin-publish' version "0.10.1"
+    id 'java-gradle-plugin'
+}
+
+group 'io.quarkus'
+version '999-SNAPSHOT'
```
gsmet: I don't know exactly where the name of the plugin is defined but we should be careful about it.

I think having `kotlin-` somewhere would be necessary. Same for the Maven artifact id and the directory of the module.

gsmet (changes requested):
Thanks for your contribution!

I added a couple of very general remarks but I haven't reviewed the code. I'll let that to @gastaldi and @geoand .

Also, the commit message needs to be far more explicit than `Initial commit`. Something like "Add an allopen Gradle plugin for Kotlin support", maybe?

geoand:
Thanks for this!

I would like to know what the high level plan is here. Do you plan to add a Maven plugin as well?
If so, how much code gets shared?

I am also wondering if this sort of thing should be a separate repo and not tied to the Quarkus release cycle

KaenDagger:
> Also, the commit message needs to be far more explicit than Initial commit

@gsmet  this was WIP , so thought when I would rebase , I would change the commit message

> I am also wondering if this sort of thing should be a separate repo and not tied to the Quarkus release cycle

@geoand I could make separate repo on my account. My initial plan was to make that only.

geoand:
I'm not sure about the release cycle, just wondering.

@gastaldi @gsmet WDYT?

gastaldi:
How different is this plugin from
https://plugins.gradle.org/plugin/org.jetbrains.kotlin.plugin.allopen?

IMHO I'd rather prefer to use the plugin provided by the Kotlin team than
to maintain our own :)

>

geoand:
@gastaldi we would like out own for the reasons outlined here: https://github.com/quarkusio/quarkus/issues/1157#issuecomment-570486347

gastaldi:
Ah thanks for the pointer, I see now the discussion in the original issue

Em Dom, 5 de jan de 2020 13:00, Georgios Andrianakis <
notifications@github.com> escreveu:

gastaldi:
I wonder if the quarkus gradle plugin could detect this and apply these
settings automatically without having to depend on another plugin

Em Dom, 5 de jan de 2020 13:02, George Gastaldi <gegastaldi@gmail.com>
escreveu:

geoand:
One problem with that approach is dev mode.
In dev mode you need to re-apply everything and its much easier to just apply a Kotlin compiler plugin that takes care of everything on it's own

geoand:
@KaenDagger please disregard my comment about the Maven plugin

KaenDagger:
For the maven I have seen in Kotlin all open , there is only one extension written for it. Most of the code is written for gradle only. 
As @geoand said, I also prefer completely different plugin (not integrated with this repo)

devtools/allopen/pom.xml
```diff
@@ -0,0 +1,21 @@
...
+
+    <artifactId>quarkus-allopen</artifactId>
+    <name>Quarkus - AllOpen Plugin</name>
+
+    <properties>
+        <gradle.executable>./gradlew</gradle.executable>
+        <gradle.task>build</gradle.task>
+        <skip.gradle.build>false</skip.gradle.build>
+    </properties>
+</project>
```
gastaldi: I think your pom.xml is incomplete? These properties are not used anywhere

KaenDagger:
Actually I haven't written tasks yet. 😅
These are for plugins tasks

devtools/allopen/pom.xml
KaenDagger: Actually I haven't written tasks yet. 😅
These are for plugins tasks

KaenDagger:
this is separate repository for this plugin which I am testing. 

https://github.com/KaenDagger/quarkus-allopen

It will be integrated when its in working condition

gastaldi:
Nice! Mind closing this PR for now and opening a new one when it's ready? That will help us in our daily tasks for reviewing and triaging PRs :)

KaenDagger:
sure !

gastaldi:
Thank you!

PR #6409, opened 2020-01-05
Improve the Neo4j guide

michael-simons:
The Neo4j driver has been updated to 4.0.0 final and Neo4j 4.0 can be
used from Docker, the preview flag can now be removed.

The final driver revealed some errors in the existing guide:

The reading of the transactional result of a transactional function
needs to happen inside the same transaction.
Therefor the guide needs a fix so that the composition of the future
extracting the node happens in the transaction, not outside.

A corresponding PR to the Guides itself will be created in a couple of seconds.

Thanks @gunnarmorling for bringing this to my attention, much appreciated.

### DIFF
--- a/docs/src/main/asciidoc/neo4j.adoc
+++ b/docs/src/main/asciidoc/neo4j.adoc
@@ -4,7 +4,7 @@ and pull requests should be submitted there:
 https://github.com/quarkusio/quarkus/tree/master/docs/src/main/asciidoc
 ////
 = Quarkus - Neo4j
-:neo4j_version: 3.5.13
+:neo4j_version: 4.0.0
 
 include::./attributes.adoc[]
 :extension-status: preview
@@ -35,12 +35,6 @@ while Neo4j itself is available in a GPL3-licensed open-source "community editio
 with online backup and high availability extensions licensed under a closed-source commercial license.
 
 include::./status-include.adoc[]
-
-[WARNING]
-====
-It is based on an alpha version of the Neo4j driver.
-Hence the reason for it to be _preview_
-====
 
 == Programming model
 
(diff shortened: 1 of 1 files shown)


docs/src/main/asciidoc/neo4j.adoc
```diff
@@ -4,10 +4,9 @@ and pull requests should be submitted there:
 https://github.com/quarkusio/quarkus/tree/master/docs/src/main/asciidoc
 ////
 = Quarkus - Neo4j
-:neo4j_version: 3.5.13
+:neo4j_version: 4.0.0
 
 include::./attributes.adoc[]
-:extension-status: preview
```
gunnarmorling: There's still a note above the warning you removed which mentions that it is preview.

docs/src/main/asciidoc/neo4j.adoc
```diff
@@ -321,10 +314,10 @@ We also add some exception handling, in case the resource is called with an inva
 public CompletionStage<Response> getSingle(@PathParam("id") Long id) {
     AsyncSession session = driver.asyncSession();
     return session
-        .readTransactionAsync(tx ->
```
gunnarmorling: Not related to the change itself, but rather to the entire method: this is 22 lines for a single get-by-id call :(

I understand it's all async and non-blocking and so on, but e.g. in comparison to Panache where this would be one line essentially, the signal-to-noise ratio seems not great. Is there any way for simplifying this? Perhaps first show imperative, too, before going to async/reactive?

gunnarmorling (commented):
Thanks a lot, @michael-simons! Two comments inline, one of them just loosely related.

docs/src/main/asciidoc/neo4j.adoc
michael-simons: Comparing Panache with the raw Driver is like comparing access via Panache to plain JDBC driver. I don’t see the point here.

gunnarmorling: Good point on the comparison, it's not 100% fair. Still the amount of code for this piece of functionality is overwhelming. Anyways, seems my comment isn't really actionable right now, so feel free to ignore it.

docs/src/main/asciidoc/neo4j.adoc
gsmet: I would keep the extension as `preview` for now until we get sufficient feedback.

Agreed to remove the warning below.

gsmet (changes requested):
Added a small comment about the status.

docs/src/main/asciidoc/neo4j.adoc
michael-simons: We have detailed docs on this: https://neo4j.com/docs/driver-manual/4.0-preview/session-api/ This is about choice. 
Actually, most Neo4j users that use Neo4j-OGM (our object mapper) keep on asking how to work around that object mapping. So we think it's in the interest of our (potential) users to make access to the driver as easy as possible and than let the users chose: Either just using imperative (I think that will be good enough in the vast majority of use cases), using an asynchronous connection (available on all Neo4j versions supported by the driver) or going reactive (available on the Neo4j 4.0 and the best choice in a reactive world based on Quarkus).

docs/src/main/asciidoc/neo4j.adoc
michael-simons: Agreed and added back in.

gsmet (approved):

Issue #5354, opened 2019-11-10
java.lang.LinkageError from quarkus:dev but not from java -jar 


brunoabdon:
**Describe the bug**
Quarkus startup fails with `java.lang.LinkageError` from a specific seemingly unharmful combination of facts:

* The app has a JAX `@Path` resource that extends a super from "_dependency-b_"
* This JAX resource overrides a method from its super that returns a `ReturnType` from "_dependency-a_"
* "_dependency-a_" has a class annotated with [`@MappedSuperClass`](https://javaee.github.io/javaee-spec/javadocs/javax/persistence/MappedSuperclass.html) (that is never used).

**Expected behavior**
Startup with `mvn quarkus:dev` should not fail with `java.lang.LinkageError`

**Actual behavior**
Quarkust fails to start, complaining about `java.lang.LinkageError`.

**To Reproduce**
1. Clone this showcase
`git clone git@github.com:brunoabdon/quarkus-issue.git`
2. Build dependency-a
`cd quarkus-issue/dependency-a/`
`mvn package install`
3. Build dependency-b
`cd ../dependency-b/`
`mvn package install`
4. Build and run the quarkus-app 
`cd ../quarkus-app/`
`mvn quarkus:dev`

**Configuration**
It's already at `git@github.com:brunoabdon/quarkus-issue.git`

**Screenshots**
<pre>
22:10:21,454 ERROR [io.qua.dev.DevModeMain] Failed to start Quarkus: java.lang.ExceptionInInitializerError
	at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
	at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.lang.reflect.Constructor.newInstance(Constructor.java:423)
	at java.lang.Class.newInstance(Class.java:442)
	at io.quarkus.runner.RuntimeRunner.run(RuntimeRunner.java:142)
	at io.quarkus.dev.DevModeMain.doStart(DevModeMain.java:177)
	at io.quarkus.dev.DevModeMain.start(DevModeMain.java:95)
	at io.quarkus.dev.DevModeMain.main(DevModeMain.java:66)
Caused by: java.lang.RuntimeException: Failed to start quarkus
[...]

Issue #3917, opened 2019-09-07
Failed to build a runner jar: java.lang.NoSuchFieldError: AROUND_CONSTRUCT


4ntoine:
**Describe the bug**
Can't build jar written in Kotlin and build with Gradle.

**Expected behavior**
Jar file is compiled

**Actual behavior**
Getting build error:
```
* Exception is:
org.gradle.api.tasks.TaskExecutionException: Execution failed for task ':app-quarkus:quarkusBuild'.
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter$3.accept(ExecuteActionsTaskExecuter.java:151)
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter$3.accept(ExecuteActionsTaskExecuter.java:148)
        at org.gradle.internal.Try$Failure.ifSuccessfulOrElse(Try.java:191)
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.execute(ExecuteActionsTaskExecuter.java:141)
        at org.gradle.api.internal.tasks.execution.ResolveBeforeExecutionStateTaskExecuter.execute(ResolveBeforeExecutionStateTaskExecuter.java:75)
        at org.gradle.api.internal.tasks.execution.ValidatingTaskExecuter.execute(ValidatingTaskExecuter.java:62)
        at org.gradle.api.internal.tasks.execution.SkipEmptySourceFilesTaskExecuter.execute(SkipEmptySourceFilesTaskExecuter.java:108)
        at org.gradle.api.internal.tasks.execution.ResolveBeforeExecutionOutputsTaskExecuter.execute(ResolveBeforeExecutionOutputsTaskExecuter.java:67)
        at org.gradle.api.internal.tasks.execution.ResolveAfterPreviousExecutionStateTaskExecuter.execute(ResolveAfterPreviousExecutionStateTaskExecuter.java:46)
        at org.gradle.api.internal.tasks.execution.CleanupStaleOutputsExecuter.execute(CleanupStaleOutputsExecuter.java:94)
        at org.gradle.api.internal.tasks.execution.FinalizePropertiesTaskExecuter.execute(FinalizePropertiesTaskExecuter.java:46)
        at org.gradle.api.internal.tasks.execution.ResolveTaskExecutionModeExecuter.execute(ResolveTaskExecutionModeExecuter.java:95)
[...]

Issue #770, opened 2019-02-06
ClassLoader and augmentation issues


FroMage:
For example, Hibernate and Panache enhance entity `Foo`, and Panache then generates a `Foo$generated` class, which uses reflection on `Foo` to get some info that's hard to get using jandex. It doesn't care if it sees a non-enhanced version of the class, using its class' class loader. But if it uses theTCCL (`RuntimeClassLoader`) that will trigger a load of the entity class before the `transformers` have been set and all hell breaks loose later on with really weird failures.

And if I don't use the TCCL `shamrock:dev` does not work, for some _other_ reason.

We're going ot have to clarify what reflection is allowed during bytecode generation (Hibernate uses it, ASM uses it), and how to avoid these issues of difference between `shamrock:dev` and the normal runner.

rsvoboda:
@FroMage is this fixed by the big CL change commit ?

FroMage:
Yeah, at the very least it clarified how things work.

PR #6411, opened 2020-01-06
The big ClassLoader change

### LINKED ISSUE #5597: Classloader Isolation Model
### Description
Due to the Gradle specificities and the bug around Panache entities + typed queries, we need this clarified CL model.

### Analysis
Design work has started on the public list https://groups.google.com/forum/#!topic/quarkus-dev/3u24xR6H7zs and https://docs.google.com/document/d/1XOD4_OWi93gD0mAT39fyZvDCVSwPbH3yVpkBbM1yPVg/edit?usp=sharing once everyone is happy I will start to implement it.

### Related Issues
- [ ] https://github.com/quarkusio/quarkus/issues/6190
- [ ] https://github.com/quarkusio/quarkus/issues/1724
- [ ] https://github.com/quarkusio/quarkus/issues/5354
- [ ] https://github.com/quarkusio/quarkus/issues/770
- [ ] https://github.com/quarkusio/quarkus/issues/4960
- [ ] https://github.com/quarkusio/quarkus/issues/3917

### Tasks
- [ ] _Checklist of sub tasks goes here_

stuartwdouglas:
Fixes #5597 #1724 #5354 #770 #3917 #6652

This changes the way Quarkus ClassLoading works,
to allow for isolated class loaders.

It also unifies how Quarkus is launched, so every
different mode we support uses the same mechanism
for both curation and launch.

Tests are now run in an isolated ClassLoader, which
means that a proxy is created that runs the tests
from within the isolated ClassLoader. This currently
has a quirk where @BeforeAll methods are run twice,
which will be fixed in the next JUnit release. This
can be worked around by using @QuarkusBeforeAll.

Apparently JUnit 5.6 which will fix this should be released on the 20th according to https://github.com/junit-team/junit5/milestone/46, which will allow this restriction to be removed.

This still needs performance testing, but otherwise is very close to being complete.

All of these should be addressed in the new few days.


core/deployment/src/main/java/io/quarkus/runner/bootstrap/GenerateConfigTask.java
```diff
@@ -0,0 +1,161 @@
...
+                    }
+                }
+            });
+
+            List<ConfigDescriptionBuildItem> descriptions = buildResult.consumeMulti(ConfigDescriptionBuildItem.class);
+            Collections.sort(descriptions);
+
+            String existing = "";
+            if (Files.exists(configFile)) {
+                try (InputStream in = new FileInputStream(configFile.toFile())) {
```
aloubyansky: Just curious, any reason not to do it the nio way with Files.newInputStream(configFile)?

core/deployment/src/main/java/io/quarkus/runner/bootstrap/GenerateConfigTask.java
```diff
@@ -0,0 +1,161 @@
...
+                }
+
+                sb.append("\n#\n");
+                sb.append(formatDocs(i.getDocs()));
+                sb.append("\n#\n#");
+                sb.append(i.getPropertyName() + "=" + i.getDefaultValue());
+                sb.append("\n");
+            }
+
+            try (FileOutputStream out = new FileOutputStream(configFile.toFile(), true)) {
```
aloubyansky: Same here. Or it looks like alternatively you could do `Files.newBufferedWriter` and write directly there instead of `sb` using `newLine()`.

aloubyansky (commented):
I haven't yet reviewed it completely. I'll probably do it in a few steps as it's quite large and not trivial.

test-framework/junit5/src/main/java/io/quarkus/test/junit/QuarkusTestExtension.java
```diff
@@ -3,220 +3,90 @@
...
-        final LinkedBlockingDeque<Runnable> shutdownTasks = new LinkedBlockingDeque<>();
+        try {
+            final LinkedBlockingDeque<Runnable> shutdownTasks = new LinkedBlockingDeque<>();
 
-        Path appClassLocation = getAppClassLocation(context.getRequiredTestClass());
+            Path appClassLocation = getAppClassLocation(context.getRequiredTestClass());
 
-        appCl = createQuarkusBuildClassLoader(appClassLocation);
-        originalCl = setCCL(appCl);
+            QuarkusBootstrap.builder(appClassLocation);
```
geoand: Not that it should be an issue, but I assume this is not needed since `QuarkusBootstrap.builder(appClassLocation).setMode(QuarkusBootstrap.Mode.TEST);` is being used later

docs/src/main/asciidoc/class-loading-reference.adoc
```diff
@@ -0,0 +1,23 @@
+////
+This guide is maintained in the main Quarkus repository
+and pull requests should be submitted there:
+https://github.com/quarkusio/quarkus/tree/master/docs/src/main/asciidoc
+////
+= Quarkus - Class Loading Reference
```
geoand: +100 for this doc!

It would be wonderful to have this document completed so people that haven't followed the ClassLoader system design discussions can have a good reference point and perhaps even be able to contribute somewhat to the review process

stuartwdouglas:
This commit includes a class loading guide that gives a bit of an explanation as to how it all works, when reviewing this is probably a good place to start.

docs/src/main/asciidoc/class-loading-reference.adoc
```diff
@@ -0,0 +1,163 @@
...
+Base ClassLoader:: 
+
+This is usually the normal JVM System ClassLoader. In some environments such as Maven it may be different. This ClassLoader
+is used to load the bootstrap classes, and other ClassLoader instances will delegate the loading of JDK classes to it.
+
+Augment ClassLoader::
+
+This loads all the `-deployment` artifacts and their dependencies, as well as other user dependencies. It does not load the
+application root or any hot deployed code. This ClassLoader is persistent, even if the application restarts it will remain
+(which is why it cannot load application classes that may be hot deployed). It's parent is the base ClassLoader, and it is
```
gsmet:
```suggestion
(which is why it cannot load application classes that may be hot deployed). Its parent is the base ClassLoader, and it is
```

docs/src/main/asciidoc/class-loading-reference.adoc
```diff
@@ -0,0 +1,163 @@
...
+transformer safe.
```
gsmet: I think we should have an explanation of what that means.

docs/src/main/asciidoc/class-loading-reference.adoc
```diff
@@ -0,0 +1,163 @@
...
+
+At present this can be configured to delegate to the Base ClassLoader, however the plan is for this option to go away and
+always have this as an isolated ClassLoader. Making this an isolated ClassLoader is complicated as it means that all
+the builder classes are isolated, which means that use cases that want to customise the build chains are slightly more complex.
+
+Deployment ClassLoader::
+
+This can load all application classes, it's parent is the Augment ClassLoader so it can also load all deployment classes.
```
gsmet:
```suggestion
This can load all application classes, its parent is the Augment ClassLoader so it can also load all deployment classes.
```

docs/src/main/asciidoc/class-loading-reference.adoc
```diff
@@ -0,0 +1,163 @@
...
+
+This ClassLoader is non-persistent, it will be re-created when the application is started, and is isolated. This ClassLoader
+is the context ClassLoader that is used when running the build steps. It is transformer safe so loading application classes
+will not prevent them from being transformed.
```
gsmet: Oh, you explain it there.

docs/src/main/asciidoc/class-loading-reference.adoc
```diff
@@ -0,0 +1,163 @@
...
+
+Base Runtime ClassLoader::
+
+This loads all the runtime extension dependencies, as well as other user dependencies. It does not load the
+application root or any hot deployed code. This ClassLoader is persistent, even if the application restarts it will remain
+(which is why it cannot load application classes that may be hot deployed). It's parent is the base ClassLoader.
```
gsmet:
```suggestion
(which is why it cannot load application classes that may be hot deployed). Its parent is the base ClassLoader.
```

docs/src/main/asciidoc/class-loading-reference.adoc
```diff
@@ -0,0 +1,163 @@
...
+
+This loads code that is not hot-reloadable, but it does support transformation (although once the class is loaded this
+transformation is no longer possible). This means that only transformers registered in the first application start
+will take effect, however as these transformers are expected to be idempotent this should not cause problems.
+
+This ClassLoader is isolated from the Augment and Deployment ClassLoaders. This means that it is not possible to set
+values in a static field in the deployment side, and expect to read it at runtime. This allows dev and test applications
+to behave more like a production application (production applications are isolated in that they run in a whole new JVM).
+
+This also means that the runtime version can be linked against a different set of dependencies, e.g. the hibernate
```
gsmet:
```suggestion
This also means that the runtime version can be linked against a different set of dependencies, e.g. the Hibernate ORM
```

docs/src/main/asciidoc/class-loading-reference.adoc
```diff
@@ -0,0 +1,163 @@
...
+version used at deployment time might want to include ByteBuddy, while the version used at runtime does not.
+
+Runtime Class Loader::
+
+This ClassLoader is used to load the application classes and other hot deployable resources. It's parent it the base runtime
```
gsmet:
```suggestion
This ClassLoader is used to load the application classes and other hot deployable resources. Its parent is the base runtime
```

docs/src/main/asciidoc/class-loading-reference.adoc
```diff
@@ -0,0 +1,163 @@
...
+ClassLoader, and it is recreated when the application is restarted.
+
+
+== Isolated ClassLoaders
+
+The runtime ClassLoader is always isolated. This means that it will have it's own copies of almost every class from the
```
gsmet:
```suggestion
The runtime ClassLoader is always isolated. This means that it will have its own copies of almost every class from the
```

docs/src/main/asciidoc/class-loading-reference.adoc
```diff
@@ -0,0 +1,163 @@
...
+
+There are some classes that should not be loaded in an isolated manner, but that should always be loaded by the system
+ClassLoader (or whatever ClassLoader is responsible for bootstrapping Quarkus). Most extensions do not need to worry about
+this, however there are a few cases where this is necessary:
+
+- Some logging related classes, as logging must be loaded by the system ClassLoader
+- Quarkus bootstrap itself
+
+If this is required it can be configured in the `quarkus-bootstrap-maven-plugin`. Note that if you
+mark a dependency as parent first then all of it's dependencies must also be parent first,
```
gsmet:
```suggestion
mark a dependency as parent first then all of its dependencies must also be parent first,
```

docs/src/main/asciidoc/class-loading-reference.adoc
```diff
@@ -0,0 +1,163 @@
...
+    </configuration>
+</plugin>
+----
+
+=== Banned Dependencies
+
+There are some dependencies that we can be sure we do not want. This generally happens when a dependency has had a name
+change (e.g. smallrye-config changing groups from `org.smallrye` to `org.smallrye.config`, the `javax` -> `jakarta` rename).
+This can cause problems, as if these artifacts end up in the dependency tree out of date classes can be loaded that are
+not compatible with Quarkus. To deal with this this extensions can specify artifacts that should never be loaded. This is
```
gsmet:
```suggestion
not compatible with Quarkus. To deal with this, extensions can specify artifacts that should never be loaded. This is
```

docs/src/main/asciidoc/class-loading-reference.adoc
```diff
@@ -0,0 +1,163 @@
...
+    <configuration>
+        <excludedArtifacts>
+            <excludedArtifact>io.smallrye:smallrye-config</excludedArtifact>
+            <excludedArtifact>javax.enterprise:cdi-api</excludedArtifact>
+        </excludedArtifacts>
+    </configuration>
+</plugin>
+----
+
+This should only be done if the extension depends on a never version of these artifacts. If the extension does not bring
```
gsmet:
```suggestion
This should only be done if the extension depends on a newer version of these artifacts. If the extension does not bring
```

gsmet (dismissed):
I will dig more into this but here is a small review of the documentation part.

docs/src/main/asciidoc/class-loading-reference.adoc
```diff
@@ -0,0 +1,163 @@
...
+
+This can load all application classes, it's parent is the Augment ClassLoader so it can also load all deployment classes.
+
+This ClassLoader is non-persistent, it will be re-created when the application is started, and is isolated. This ClassLoader
+is the context ClassLoader that is used when running the build steps. It is transformer safe so loading application classes
+will not prevent them from being transformed.
+
+Base Runtime ClassLoader::
+
+This loads all the runtime extension dependencies, as well as other user dependencies. It does not load the
```
geoand: `as well as other user dependencies` is mentioned here for `Base Runtime Classloader`, however `Augment ClassLoader` also mentions `as well as other user dependencies`.

This seems kind of confusing, but I assume that the former is done at build time and the latter at runtime? Maybe a note should be made about this?

docs/src/main/asciidoc/class-loading-reference.adoc
geoand: The explanation exists, but it's further down, so I agree it should be moved up here

docs/src/main/asciidoc/class-loading-reference.adoc
```diff
@@ -0,0 +1,163 @@
...
+application root or any hot deployed code. This ClassLoader is persistent, even if the application restarts it will remain
+(which is why it cannot load application classes that may be hot deployed). It's parent is the base ClassLoader.
+
+This loads code that is not hot-reloadable, but it does support transformation (although once the class is loaded this
+transformation is no longer possible). This means that only transformers registered in the first application start
+will take effect, however as these transformers are expected to be idempotent this should not cause problems.
```
geoand: Could we have an example of what sort of transformations are meant here?

docs/src/main/asciidoc/class-loading-reference.adoc
geoand: I think this needs to have more attention drawn to it. Extension authors would probably need to know more about this

docs/src/main/asciidoc/class-loading-reference.adoc
geoand: Can you also please add an explanation why this classloader needs to be transformer safe? Is this need driven by the fact that we need to transform dependencies in some cases?

geoand:
I took at this a little and I really love the way things get a lot cleaner

integration-tests/hibernate-orm-panache/src/test/java/io/quarkus/it/panache/PanacheFunctionalityTest.java
```diff
@@ -45,6 +55,7 @@ public void testPanacheSerialisation() {
 
     @DisabledOnNativeImage
     @Test
+    @ActivateRequestContext
```
FroMage: Is the request context no longer active by default for tests, then?

stuartwdouglas: ah, good catch. I need to fix this, the test scope setup is being run from outside the scope of the deployment so it is not actually activating the scope.

stuartwdouglas:
The windows tests are failing because something somewhere is not closing a file handle for the libs/$app}.jar file in the gradle build. The gradle tests can't delete their temp folders and fail as a result.

core/deployment/src/main/java/io/quarkus/runner/bootstrap/AugmentActionImpl.java
```diff
@@ -0,0 +1,214 @@
...
+
+    @Override
+    public AugmentResult createProductionApplication() {
+        if (launchMode != LaunchMode.NORMAL) {
+            throw new IllegalStateException("Can only create a production application when using NORMAL launch mode");
+        }
+        BuildResult result = runAugment(true, Collections.emptySet(), ArtifactResultBuildItem.class);
+        JarBuildItem jarBuildItem = result.consumeOptional(JarBuildItem.class);
+        NativeImageBuildItem nativeImageBuildItem = result.consumeOptional(NativeImageBuildItem.class);
+        curatedApplication.close();
```
aloubyansky: I guess it'd be safer to close it in `finally` in case `runAugment` throws a runtime exception. In general though, I don't see `CuratedApplication` being closed anywhere else which looks like a major issue. Did I overlook something?
I get that depending on the mode it may have different lifecycles. But it'd be great to improve the design around it to make it more clear.
We do need to test prove it on Windows which is more sensitive to closing resources.

aloubyansky: Ah, sorry, I see RunningQuarkusApplicationImpl.

core/deployment/src/main/java/io/quarkus/runner/bootstrap/StartupActionImpl.java
```diff
@@ -0,0 +1,172 @@
...
+
+    static final String DEBUG_CLASSES_DIR = System.getProperty("quarkus.debug.generated-classes-dir");
+
+    private final CuratedApplication curatedApplication;
+    private final AugmentAction augmentAction;
+    private final BuildResult buildResult;
+
+    public StartupActionImpl(CuratedApplication curatedApplication, AugmentAction augmentAction, BuildResult buildResult) {
+        this.curatedApplication = curatedApplication;
+        this.augmentAction = augmentAction;
```
aloubyansky: augmentAction looks like a leftover. Can be removed?

independent-projects/bootstrap/core/src/main/java/io/quarkus/bootstrap/app/CuratedApplication.java
```diff
@@ -0,0 +1,279 @@
...
+            //                    builder.addBannedElement(element);
+            //                }
+            //            }
+            augmentClassLoader = builder.build();
+
+        }
+        return augmentClassLoader;
+    }
+
+    private AppArtifactKey getKey(AppDependency i) {
```
aloubyansky: This method looks equivalent to `i.getArtifact().getKey()`.

devtools/gradle/src/main/java/io/quarkus/gradle/AppModelGradleResolver.java
```diff
@@ -228,4 +267,31 @@ private Properties resolveDescriptor(final Path path) {
         }
         return rtProps;
     }
+
+    static final class AppKey {
```
aloubyansky: Why not use AppArtifactKey instead?

devtools/gradle/src/main/java/io/quarkus/gradle/tasks/QuarkusTestConfig.java
```diff
@@ -34,7 +33,7 @@ public void setupTest() {
 
             for (Test test : getProject().getTasks().withType(Test.class)) {
                 final Map<String, Object> props = test.getSystemProperties();
-                props.put(BootstrapClassLoaderFactory.PROP_DEPLOYMENT_CP, deploymentCp);
+                //props.put(BootstrapClassLoaderFactory.PROP_DEPLOYMENT_CP, deploymentCp);
```
aloubyansky: That's an interesting one. If it's not needed any more, this whole task is questionable. We shouldn't be resolving the app model here, at a minimum.

devtools/maven/src/main/java/io/quarkus/maven/DevMojo.java
```diff
@@ -528,49 +528,45 @@ void prepare() throws Exception {
...
+                repoSystem = MavenRepoInitializer.getRepositorySystem(repoSession.isOffline(), localProject.getWorkspace());
             }
-            for (AppDependency appDep : appModel.getAllDependencies()) {
-                addToClassPaths(classPathManifest, devModeContext, appDep.getArtifact().getPath().toFile());
+            DefaultArtifact bootstrap = new DefaultArtifact("io.quarkus", "quarkus-development-mode", "jar",
+                    pluginDef.getVersion());
+            MavenArtifactResolver mavenArtifactResolver = MavenArtifactResolver.builder()
+                    .setRepositorySystem(repoSystem)
+                    .setRepositorySystemSession(repoSession)
+                    .setRemoteRepositories(repos)
```
aloubyansky: I think `.setWorkspace(localProject.getWorkspace())` should still be necessary. Otherwise, it won't be able to resolve artifacts that are supposed to be produced by the workspace but haven't actually been installed.

devtools/maven/src/main/java/io/quarkus/maven/DevMojo.java
```diff
@@ -528,49 +528,45 @@ void prepare() throws Exception {
...
+                    .build();
+            DependencyResult cpRes = mavenArtifactResolver.resolveManagedDependencies(
```
aloubyansky: I'd simply use `mavenArtifactResolver.resolveDependencies(` instead. It does delegate to this method after all, it's purely to use softer words :) and possibly create less confusion for those reading the code and not to learn what these method names actually mean.

core/devmode/src/main/java/io/quarkus/dev/DevModeMain.java
```diff
@@ -43,24 +29,16 @@
...
 
     public DevModeMain(DevModeContext context) {
         this.context = context;
     }
 
     public static void main(String... args) throws Exception {
-        Timing.staticInitStarted();
 
         try (InputStream devModeCp = DevModeMain.class.getClassLoader().getResourceAsStream(DEV_MODE_CONTEXT)) {
             DevModeContext context = (DevModeContext) new ObjectInputStream(new DataInputStream(devModeCp)).readObject();
```
aloubyansky: `context` doesn't seem to be closed, btw.

devtools/maven/src/main/java/io/quarkus/maven/DevMojo.java
aloubyansky: Actually, you don't need to set the workspace here since you are not resolving any app deps with this resolver. Which means you can also simply use `DevMojo.this.repoSystem` and remove `repoSystem` local variable.

devtools/maven/src/main/java/io/quarkus/maven/GenerateConfigMojo.java
```diff
@@ -95,48 +91,28 @@ public void execute() throws MojoExecutionException {
...
-                            .build());
-            appModel = modelResolver.resolveModel(appArtifact);
-        } catch (Exception e) {
-            throw new MojoExecutionException("Failed to resolve application model " + appArtifact + " dependencies", e);
-        }
         if (project.getResources().isEmpty()) {
             throw new MojoExecutionException("No resources directory, cannot create application.properties");
         }
+        try {
+            CuratedApplication curatedApplication = QuarkusBootstrap.builder(Paths.get(project.getBuild().getOutputDirectory()))
```
aloubyansky: Looks like `curatedApplication` isn't closed. I would also strongly recommend creating maven artifact resolvers in mojos and passing them to the bootstrap.

devtools/maven/src/main/java/io/quarkus/maven/NativeImageMojo.java
```diff
@@ -186,278 +175,157 @@ public NativeImageMojo() {
...
+            realProperties.putIfAbsent("quarkus.application.version", project.getVersion());
+
+            Map<String, String> config = createCustomConfig();
+            Map<String, String> old = new HashMap<>();
+            for (Map.Entry<String, String> e : config.entrySet()) {
+                old.put(e.getKey(), System.getProperty(e.getKey()));
+                System.setProperty(e.getKey(), e.getValue());
+            }
+            try {
+                CuratedApplication curatedApplication = QuarkusBootstrap.builder(outputDirectory.toPath())
```
aloubyansky: Let's create the maven artifact resolver and pass it to the bootstrap here.

devtools/gradle/src/main/java/io/quarkus/gradle/tasks/QuarkusTestConfig.java
aloubyansky: IIUIC, here is a potential issue with this change. Gradle and Maven perform version convergence differently. The way it currently is in master we are using the same Gradle resolver for tests and actual builds.
With this change it looks like we are going to be using the Maven resolver for the tests and the Gradle one for the actual builds. And the dev mode will always be using the Maven resolver in `DevModeMain`.

core/deployment/src/main/java/io/quarkus/runner/bootstrap/AugmentActionImpl.java
stuartwdouglas: There were some gradle failures on windows that were due to this (the app not being closed). The gradle tests create a temp dir and the test fails if the dir cannot be released.

devtools/gradle/src/main/java/io/quarkus/gradle/tasks/QuarkusTestConfig.java
stuartwdouglas: This is something I am not sure about, and wanted some feedback on.

If this is removed we fall back to class path based discovery in the maven resolver (unless there happens to be a pom present as well). The main advantage of this is that it should allow gradle tests to also run in the IDE, which does not work at the moment. 

If that is not possible then we need to somehow encode the app model into this system prop and restore the old behaviour.

core/devmode/src/main/java/io/quarkus/dev/DevModeMain.java
stuartwdouglas: context is not Closeable

devtools/maven/src/main/java/io/quarkus/maven/NativeImageMojo.java
stuartwdouglas: Ok. All the resolution stuff is what I am least sure about, I think it is likely that some of my changes are not 100% correct, or at the very least not as efficient as they could be.

aloubyansky: First, the repo system, session, repos, etc here are properly initialized in the mojo, let's call them the original ones, because in tests and now in bootstrap we are re-creating them (currently, most probably using different versions of the api and impl and there have been lots of issues related to this initialization). Second, avoiding them being re-initialized is going to be a lot more efficient, given that Maven has already cached some artifact and resolution request-related data.

stuartwdouglas:
CI is now passing

core/devmode/src/main/java/io/quarkus/dev/DevModeMain.java
aloubyansky: Sorry, I meant the instance of DevModeMain.

aloubyansky: instantiated below at line 45

devtools/gradle/src/main/java/io/quarkus/gradle/tasks/QuarkusTestConfig.java
aloubyansky: While it does not look elegant in this case, we are ending up testing against a different set of deps than the ones used for the build. They should be aligned. Either we use those resolved by Gradle all the way or we use those that we resolve in QuarkusBootstrap.

aloubyansky: By not looking elegant I refer to the classpath shared through the system property.

devtools/maven/src/main/java/io/quarkus/maven/NativeImageMojo.java
```diff
@@ -186,278 +176,164 @@ public NativeImageMojo() {
     @Override
     public void execute() throws MojoExecutionException, MojoFailureException {
 
-        if (project.getPackaging().equals("pom") && appArtifact == null) {
```
aloubyansky: Was there a reason to remove `appArtifact` check here?
`appArtifact` is used in the platform integration tests. The reason I allowed `pom` project and `appArtifact` combination was that the project was simply pulling all the code from the plugin config and dependencies and didn't produce any JAR.

aloubyansky: It looks like somebody has removed pom packaging from the platform tests, although the rest of the config remains the same. It could still be just a POM though.

independent-projects/bootstrap/core/src/main/java/io/quarkus/bootstrap/BootstrapAppModelFactory.java
```diff
@@ -0,0 +1,512 @@
...
+                        urls.add(new File(URLDecoder.decode(path, StandardCharsets.UTF_8.name())).toURI().toURL());
+                    }
+                }
+            }
+            List<AppDependency> artifacts = new ArrayList<>();
+            for (URL jarUrl : urls) {
+                try (JarInputStream file = new JarInputStream(jarUrl.openConnection().getInputStream())) {
+                    JarEntry entry = file.getNextJarEntry();
+                    while (entry != null) {
+                        if (entry.getName().endsWith("/pom.properties") && entry.getName().startsWith("META-INF/maven")) {
```
aloubyansky: The issue is that this is not very reliable. Neither `pom.properties` nor `pom.xml` are always present. I don't think JARs built with Gradle contain those. You can try a simple test: scan JARs in your local repo and list those that don't include `META-INF/maven`.
So this approach can be used only as "the best effort" in cases when nothing else is available.

independent-projects/bootstrap/core/src/main/java/io/quarkus/bootstrap/BootstrapAppModelFactory.java
```diff
@@ -0,0 +1,512 @@
...
+                    loadedFromState, appArtifact,
+                    stateArtifact);
+        }
+    }
+
+    private static Path resolveCachedCpPath(LocalProject project) {
+        return project.getOutputDir().resolve(QUARKUS).resolve(BOOTSTRAP).resolve(DEPLOYMENT_CP);
+    }
+
+    private static org.apache.maven.model.RepositoryPolicy toMavenRepoPolicy(RepositoryPolicy policy) {
```
aloubyansky: Looks like this one isn't used anywhere.

independent-projects/bootstrap/core/src/main/java/io/quarkus/bootstrap/classloading/QuarkusClassLoader.java
```diff
@@ -0,0 +1,504 @@
...
+        //    }
+        //}
+        for (ClassPathElement i : elements) {
+            ClassPathResource res = i.getResource(nm);
+            if (res != null) {
+                resources.add(res.getUrl());
+            }
+        }
+        if (!banned) {
+            if(resources.isEmpty() || aggregateParentResources) {
```
aloubyansky: This one looks somewhat confusing. If `aggregateParentResources` is false then it will still be looking for the resources in the parent but only if nothing was found locally. Is there a specific use-case where this is needed? I'm just thinking wouldn't simply consulting the parent only if `aggregateParentResources` was true work?

independent-projects/bootstrap/core/src/main/java/io/quarkus/bootstrap/classloading/QuarkusClassLoader.java
```diff
@@ -0,0 +1,504 @@
...
+        //            return providers[0].getResource(nm).getUrl();
+        //        }
+        //TODO: because of dev mode we can't use the fast path her, we need to iterate
+        for (ClassPathElement i : elements) {
+            ClassPathResource res = i.getResource(name);
+            if (res != null) {
+                return res.getUrl();
+            }
+        }
+        return parent.getResource(nm);
```
aloubyansky: How about taking the value of `aggregateParentResources` into account here and in `getResourceAsStream` below? Wondering whether it's this way on purpose. Otherwise, it'd be probably a good idea to have a consistent resource access switch.

independent-projects/bootstrap/core/src/main/java/io/quarkus/bootstrap/resolver/BootstrapAppModelResolver.java
```diff
@@ -130,24 +143,42 @@ public AppModel resolveModel(AppArtifact appArtifact, List<AppDependency> direct
...
         final Set<AppArtifactKey> appDeps = new HashSet<>();
         final List<AppDependency> userDeps = new ArrayList<>();
+        DependencyNode resolvedDeps;
+        if (appArtifact != null) {
+            resolvedDeps = mvn.resolveManagedDependencies(toAetherArtifact(appArtifact),
+                    directMvnDeps, managedDeps, managedRepos, excludedScopes.toArray(new String[0])).getRoot();
+        } else {
+            //if there is no main artifact we assume we already have all the deps we need
+            //we just turn them into a DependencyNode
+            resolvedDeps = mvn.toDependencyTree(directMvnDeps, managedRepos).getRoot();
```
aloubyansky: `directMvnDeps` are supposed to be only the direct deps, their transitive deps aren't in there.

independent-projects/bootstrap/core/src/main/java/io/quarkus/bootstrap/resolver/BootstrapAppModelResolver.java
```diff
@@ -157,28 +188,38 @@ public boolean visitEnter(DependencyNode node) {
...
         } catch (BootstrapDependencyProcessingException e) {
-            throw new AppModelResolverException("Failed to inject extension deployment dependencies for " + resolvedDeps.getArtifact(), e.getCause());
+            throw new AppModelResolverException(
+                    "Failed to inject extension deployment dependencies for " + resolvedDeps.getArtifact(), e.getCause());
         }
 
         List<AppDependency> deploymentDeps = Collections.emptyList();
-        if(deploymentInjector.isInjectedDeps()) {
-            final DependencyGraphTransformationContext context = new SimpleDependencyGraphTransformationContext(mvn.getSession());
+        List<Dependency> deploymentDependencies = new ArrayList<>();
```
aloubyansky: Looks like `deploymentDependencies` is not used for anything useful.

independent-projects/bootstrap/core/src/main/java/io/quarkus/bootstrap/resolver/BootstrapAppModelResolver.java
```diff
@@ -336,4 +418,31 @@ private static AppArtifact toAppArtifact(Artifact artifact) {
         }
         return directMvnDeps;
     }
+
+    static final class AppKey {
```
aloubyansky: Let's keep using `AppArtifactKey` instead.

independent-projects/bootstrap/core/src/main/java/io/quarkus/bootstrap/resolver/BootstrapAppModelResolver.java
```diff
@@ -297,37 +352,64 @@ private VersionRangeResult resolveVersionRangeResult(AppArtifact appArtifact, St
...
+                artifact.getClassifier(), artifact.getExtension(), artifact.getVersion());
         final File file = artifact.getFile();
-        if(file != null) {
+        if (file != null) {
             appArtifact.setPath(file.toPath());
         }
         return appArtifact;
     }
 
+    private static AppArtifact toAppArtifact(Artifact artifact, Map<AppKey, AppArtifact> versions) {
```
aloubyansky: Ah, this method is actually not used. Is it? So AppKey can simply be removed with this method.

independent-projects/bootstrap/core/src/main/java/io/quarkus/bootstrap/resolver/maven/DeploymentInjectingDependencyVisitor.java
```diff
@@ -111,19 +115,23 @@ private void processPlatformArtifact(DependencyNode node, Path descriptor) throw
...
-            managedDeps.add(new Dependency(node.getArtifact(), JavaScopes.COMPILE));
-            managedDeps.add(new Dependency(deploymentArtifact, JavaScopes.COMPILE));
+        Artifact deploymentArtifact = toArtifact(value);
+        if(deploymentArtifact.getVersion() == null || deploymentArtifact.getVersion().isEmpty()) {
+            deploymentArtifact = deploymentArtifact.setVersion(node.getArtifact().getVersion());
         }
+        node.setData(QUARKUS_DEPLOYMENT_ARTIFACT, deploymentArtifact);
+        runtimeNodes.add(node);
+        managedDeps.add(new Dependency(node.getArtifact(), JavaScopes.COMPILE));
+        runtimeExtensionDeps.add(new Dependency(node.getArtifact(), JavaScopes.COMPILE));
```
aloubyansky: I guess it's safe to re-use `new Dependency(node.getArtifact(), JavaScopes.COMPILE)` created above.

independent-projects/bootstrap/core/src/main/java/io/quarkus/bootstrap/resolver/maven/MavenArtifactResolver.java
```diff
@@ -312,6 +317,46 @@ public DependencyResult resolveManagedDependencies(Artifact artifact, List<Depen
...
+                node.setVersionConstraint(vs.parseVersionConstraint(i.getArtifact().getVersion()));
+                node.setVersion(vs.parseVersion(i.getArtifact().getVersion()));
+            } catch (InvalidVersionSpecificationException e) {
+                throw new RuntimeException(e);
+            }
+            root.getChildren().add(node);
+        }
+        return result;
+    }
+    public DependencyResult resolveManagedDependencies(List<Dependency> deps, List<Dependency> managedDeps, List<RemoteRepository> mainRepos) throws AppModelResolverException {
```
aloubyansky: This method doesn't seem to be used anywhere. Let's remove it for now.

independent-projects/bootstrap/core/src/main/java/io/quarkus/bootstrap/resolver/maven/MavenArtifactResolver.java
```diff
@@ -328,7 +373,7 @@ private CollectRequest newCollectManagedRequest(Artifact artifact, List<Dependen
         final ArtifactDescriptorResult descr = resolveDescriptor(artifact);
         Collection<String> excluded;
         if(excludedScopes.length == 0) {
-            excluded = Arrays.asList(new String[] {"test", "provided"});
+            excluded = Collections.emptyList();
```
aloubyansky: I agree this was an ugly choice to design it this way. But by default "test" and "provided" should be excluded. That is, if the caller doesn't provide anything.

independent-projects/bootstrap/core/src/main/java/io/quarkus/bootstrap/resolver/maven/options/BootstrapMavenOptions.java
```diff
@@ -90,7 +93,7 @@ public String getOptionValue(String name) {
 
     public String[] getOptionValues(String name) {
         final Object o = options.get(name);
-        return o == null ? null : (String[]) o;
+        return o == null ? null : o instanceof String ? new String[] { o.toString() } : (String[]) o;
```
aloubyansky: Just curious, was there a case when it actually was a `String`? Which option was that?

independent-projects/bootstrap/core/src/main/java/io/quarkus/bootstrap/util/IoUtils.java
```diff
@@ -1,6 +1,7 @@
 package io.quarkus.bootstrap.util;
 
 import java.io.BufferedReader;
+import java.io.ByteArrayOutputStream;
```
aloubyansky: This one is redundant.

independent-projects/bootstrap/core/src/test/java/io/quarkus/bootstrap/resolver/CollectDependenciesBase.java
```diff
@@ -43,8 +44,8 @@ public void testCollectedDependencies() throws Exception {
             expected.addAll(expectedResult);
             expected.addAll(deploymentDeps);
         }
-        final List<AppDependency> resolvedDeps = getTestResolver().resolveModel(root.toAppArtifact()).getAllDependencies();
-        assertEquals(expected, resolvedDeps);
+        final List<AppDependency> resolvedDeps = getTestResolver().resolveModel(root.toAppArtifact()).getFullDeploymentDeps();
+        assertEquals(new HashSet<>(expected), new HashSet<>(resolvedDeps));
```
aloubyansky: Why? I think it's important to keep the list.

independent-projects/bootstrap/core/pom.xml
```diff
@@ -14,6 +14,10 @@
     <name>Quarkus - Bootstrap - Core</name>
 
     <dependencies>
+        <dependency>
+            <groupId>org.ow2.asm</groupId>
+            <artifactId>asm</artifactId>
+        </dependency>
```
aloubyansky: Where is this used in core?

integration-tests/amazon-lambda/src/test/java/io/quarkus/it/amazon/lambda/AmazonLambdaSimpleTestCase.java
```diff
@@ -29,9 +28,9 @@ public void testSimpleLambdaFailure() throws Exception {
             OutputObject out = LambdaClient.invoke(OutputObject.class, in);
             out.getResult();
             Assertions.fail();
-        } catch (LambdaException e) {
+        } catch (Exception e) {
             Assertions.assertEquals(ProcessingService.CAN_ONLY_GREET_NICKNAMES, e.getMessage());
-            Assertions.assertEquals(IllegalArgumentException.class.getName(), e.getType());
+            //Assertions.assertEquals(IllegalArgumentException.class.getName(), e.getType());
```
aloubyansky: TBH, I'm not familiar with this test but if for LambdaException it's still relevant to make sure the type is IAE, shouldn't this turn into multiple catches?

test-framework/common/src/main/java/io/quarkus/test/common/TestResourceManager.java
```diff
@@ -84,10 +84,10 @@ public void stop() {
                 throw new RuntimeException("Unable to stop Quarkus test resource " + testResource, e);
             }
         }
-        ConfigProviderResolver cpr = ConfigProviderResolver.instance();
         try {
+            ConfigProviderResolver cpr = ConfigProviderResolver.instance();
             cpr.releaseConfig(cpr.getConfig());
-        } catch (IllegalStateException ignored) {
+        } catch (Throwable ignored) {
```
aloubyansky: `Exception` won't be generic enough here?

test-framework/junit5/src/main/java/io/quarkus/test/junit/QuarkusTestExtension.java
```diff
@@ -3,259 +3,123 @@
...
+        try {
+            final LinkedBlockingDeque<Runnable> shutdownTasks = new LinkedBlockingDeque<>();
 
-        Path appClassLocation = getAppClassLocation(context.getRequiredTestClass());
+            Path appClassLocation = getAppClassLocation(context.getRequiredTestClass());
 
-        appCl = createQuarkusBuildClassLoader(appClassLocation);
-        originalCl = setCCL(appCl);
+            System.err
+                    .println("INIT TIME " + (System.currentTimeMillis() - ManagementFactory.getRuntimeMXBean().getStartTime()));
```
aloubyansky: Is this a leftover?

test-framework/junit5/src/main/java/io/quarkus/test/junit/QuarkusTestExtension.java
```diff
@@ -3,259 +3,123 @@
...
-                        public void writeClass(boolean applicationClass, String className, byte[] data) throws IOException {
-                            Path location = testWiringClassesDir.resolve(className.replace('.', '/') + ".class");
-                            Files.createDirectories(location.getParent());
-                            Files.write(location, data);
-                            handleCreatedFile(location, created, testWiringClassesDir, shutdownTasks);
-                        }
+            CuratedApplication curatedApplication = runnerBuilder.setTest(true).build().bootstrap();
+            Timing.staticInitStarted(curatedApplication.getBaseRuntimeClassLoader());
+            System.err.println(
+                    "CURATE TIME " + (System.currentTimeMillis() - ManagementFactory.getRuntimeMXBean().getStartTime()));
```
aloubyansky: This one too?

test-framework/junit5/src/main/java/io/quarkus/test/junit/QuarkusTestExtension.java
```diff
@@ -3,259 +3,123 @@
...
-                            handleCreatedFile(location, created, testWiringClassesDir, shutdownTasks);
-                        }
-                    })
-                    .setTransformerTarget(new TransformerTarget() {
-                        @Override
-                        public void setTransformers(
-                                Map<String, List<BiFunction<String, ClassVisitor, ClassVisitor>>> functions) {
-                            ClassLoader main = Thread.currentThread().getContextClassLoader();
+            System.err.println(
+                    "REAL START TIME " + (System.currentTimeMillis() - ManagementFactory.getRuntimeMXBean().getStartTime()));
```
aloubyansky: And this one?

aloubyansky (commented):
@stuartwdouglas have you tried the platform TS with this PR? https://github.com/quarkusio/quarkus-platform

independent-projects/bootstrap/core/src/main/java/io/quarkus/bootstrap/BootstrapAppModelFactory.java
stuartwdouglas: Thinking about it we don't actually need the artifact names here, the path is what is really important. We just need some way to make sure we don't get jdk jars.

devtools/maven/src/main/java/io/quarkus/maven/GenerateConfigMojo.java
stuartwdouglas: how do you do that? This is a seperate JVM. 

One option could be to resolve the app model in the mojo and serialize the resolved app.

aloubyansky: Isn't curated application going to be usually created in the Mojo when running from maven?

stuartwdouglas: Except for dev mode, and tests (which have always done their own resolution). I can make dev mode work by serialising the result into the dev mode runner jar, but there is not much I can do with tests.

core/deployment/src/main/java/io/quarkus/deployment/QuarkusAugmentor.java
```diff
@@ -141,6 +145,9 @@ public BuildResult run() throws Exception {
                 } catch (Exception e) {
                 }
             }
+            if (deploymentClassLoader instanceof Closeable) {
+                ((Closeable) deploymentClassLoader).close();
+            }
```
aloubyansky: Just wondering whether `classLoader` which is the application's augment classloader should also be closed here.

stuartwdouglas: No, they have different lifecycles. This one is recreated on each hot reload, the other lasts for the life of dev mode apps.

core/deployment/src/main/java/io/quarkus/runner/bootstrap/GenerateConfigTask.java
```diff
@@ -0,0 +1,157 @@
...
+                @Override
+                public void accept(BuildChainBuilder chainBuilder) {
+                    chainBuilder.addFinal(ConfigDescriptionBuildItem.class);
+                    chainBuilder.addInitial(ArchiveRootBuildItem.class);
+                }
+            }, new Consumer<BuildExecutionBuilder>() {
+                @Override
+                public void accept(BuildExecutionBuilder buildExecutionBuilder) {
+                    try {
+                        buildExecutionBuilder.produce(new ArchiveRootBuildItem(Files.createTempDirectory("empty")));
```
aloubyansky: Shouldn't we `deleteOnExit` this dir? Or do a more controlled cleanup? Perhaps we should consider introducing some kind of Quarkus-post-build-clean-up tasks in perspective.

independent-projects/bootstrap/core/src/main/java/io/quarkus/bootstrap/resolver/maven/options/BootstrapMavenOptions.java
stuartwdouglas: I think it was these ones: https://github.com/quarkusio/quarkus/blob/8f0d1587609fd1d055caaa92bce99f70998b20b3/independent-projects/bootstrap/core/src/main/java/io/quarkus/bootstrap/resolver/maven/options/BootstrapMavenOptionsParser.java#L25

independent-projects/bootstrap/core/src/main/java/io/quarkus/bootstrap/classloading/QuarkusClassLoader.java
stuartwdouglas: this is used for the deployment and runtime class loaders, so that getResources gets everything that would be on the actual application class path, without anything that may have leaked in on the system class path.

I am not sure about just never consulting the parent, as it would effectively hide JDK resources. Arc needs to be able to index JDK classes at times, so if we did not delegate at all this would cause problems.

independent-projects/bootstrap/core/src/main/java/io/quarkus/bootstrap/resolver/maven/options/BootstrapMavenOptions.java
aloubyansky: If you recall how to reproduce that, please let me know. If not we'll merge it anyway. But I'd like to see the stacktrace.

independent-projects/bootstrap/core/src/main/java/io/quarkus/bootstrap/classloading/QuarkusClassLoader.java
stuartwdouglas: So we can't completely hide the JDK resources or ArC can't index them, but we want to hide duplicate/unwanted copies of META-INF/services files. I was not 100% sure about this but I can't really think of a better approach.

core/deployment/src/main/java/io/quarkus/runner/bootstrap/GenerateConfigTask.java
stuartwdouglas: good catch, but in this case it can just be deleted immediately after.

independent-projects/bootstrap/core/src/main/java/io/quarkus/bootstrap/classloading/QuarkusClassLoader.java
aloubyansky: Ok

test-framework/common/src/main/java/io/quarkus/test/common/TestResourceManager.java
stuartwdouglas: This may not be nessesary any more, there were some smallrye class loading problems that resulted in Error's being thrown. I think they are fixed now as the CL always returns null from getParent, but I need to re-check.

independent-projects/bootstrap/core/pom.xml
stuartwdouglas: The ClassLoader implementation allows for transformers, and the transformer API is based on ASM.

independent-projects/bootstrap/core/src/main/java/io/quarkus/bootstrap/resolver/maven/options/BootstrapMavenOptions.java
stuartwdouglas: I can't actually remember, and it might not even be a problem any more as this PR has gone through so many changes.

independent-projects/bootstrap/core/src/main/java/io/quarkus/bootstrap/resolver/maven/MavenArtifactResolver.java
stuartwdouglas: This is why there is now explicit devmode and test flags, as if you are running the dev mode tests you need both.

stuartwdouglas: That is why I changed this, as there are cases where there are no excluded scopes, so now it is up to the caller.

independent-projects/bootstrap/core/src/main/java/io/quarkus/bootstrap/BootstrapAppModelFactory.java
```diff
@@ -0,0 +1,522 @@
...
+
+            Set<URL> urls = new HashSet<>();
+            //this is pretty yuck, but under JDK11 the URLClassLoader trick does not work
+            Enumeration<URL> manifests = Thread.currentThread().getContextClassLoader().getResources("META-INF/MANIFEST.MF");
+            while (manifests.hasMoreElements()) {
+                URL url = manifests.nextElement();
+                if (url.getProtocol().equals("jar")) {
+                    String path = url.getPath();
+                    if (path.startsWith("file:")) {
+                        path = path.substring(5, path.lastIndexOf('!'));
```
aloubyansky: The code here is assuming it's a JAR. It could be a directory containing classes, couldn't it?

independent-projects/bootstrap/core/src/main/java/io/quarkus/bootstrap/BootstrapAppModelFactory.java
aloubyansky: Right, unless we wanted to apply a filter.

independent-projects/bootstrap/core/src/main/java/io/quarkus/bootstrap/BootstrapAppModelFactory.java
```diff
@@ -0,0 +1,522 @@
...
+                            artifact.setPath(Paths.get(jarUrl.toURI()));
+                            artifacts.add(
+                                    new AppDependency(artifact, "compile"));
+                        }
+                        entry = file.getNextJarEntry();
+                    }
+                }
+            }
+
+            //we now have our deployment time artifacts, lets resolve all their deps
```
aloubyansky: Aren't these the runtime artifacts?

independent-projects/bootstrap/core/src/main/java/io/quarkus/bootstrap/BootstrapAppModelFactory.java
```diff
@@ -0,0 +1,522 @@
...
+            AppModel model = resolver.resolveManagedModel(appArtifact, artifacts, managingProject);
```
aloubyansky: For this invocation we'd need the actual GAVs.

independent-projects/bootstrap/core/src/main/java/io/quarkus/bootstrap/BootstrapAppModelFactory.java
```diff
@@ -0,0 +1,522 @@
...
+            return new CurationResult(model, new FixedSuppler<>(resolver));
+        } catch (Exception e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    private CurationResult createAppModelForJar(Path appArtifactPath) {
+        log.debug("provideOutcome depsOrigin=" + dependenciesOrigin + ", versionUpdate=" + versionUpdate
+                + ", versionUpdateNumber="
+                + versionUpdateNumber);
```
aloubyansky: Could you please change it to `.log.debugf(...)` instead?

independent-projects/bootstrap/core/src/main/java/io/quarkus/bootstrap/resolver/maven/MavenArtifactResolver.java
aloubyansky: This is a different default behavior though. But let's leave it for now. I didn't do it well in the first place, so I'll probably review it later.

independent-projects/bootstrap/core/src/main/java/io/quarkus/bootstrap/BootstrapAppModelFactory.java
stuartwdouglas: no, see the if statement above.

aloubyansky: Ok, my point was though about directories containing classes that are on classpath. If you are launching from an IDE that is what should appear on the classpath, right? So why is this method adding only the JARs?

independent-projects/bootstrap/core/src/main/java/io/quarkus/bootstrap/BootstrapAppModelFactory.java
```diff
@@ -212,6 +213,18 @@ public AppModelResolver getAppModelResolver() {
     }
 
     public CurationResult resolveAppModel() throws BootstrapException {
+        if (test) {
+            //gradle tests encode the result on the class path
+
+            try (InputStream existing = getClass().getResourceAsStream(BootstrapConstants.SERIALIZED_APP_MODEL)){
```
aloubyansky: Wouldn't it be safer and perhaps more efficient to look for that file in `appClasses`? Now that the file is created in the classes dir, it will end up in the jar, right? Just wondering where there could be multiple files on the classpath.

independent-projects/bootstrap/core/src/main/java/io/quarkus/bootstrap/classloading/QuarkusClassLoader.java
```diff
@@ -274,8 +274,17 @@ public InputStream getResourceAsStream(String nm) {
...
-                            protectionDomains.computeIfAbsent(classPathElement, (ce) -> ce.getProtectionDomain(this)));
+                    try {
+                        return defineClass(name, data, 0, data.length,
+                                protectionDomains.computeIfAbsent(classPathElement, (ce) -> ce.getProtectionDomain(this)));
+                    } catch (LinkageError e) {
+                        //TODO: is this a good idea?
+                        //if we are already the initiating loader we might want to load this from the parent
+                        if (!parentFirst) {
+                            return parent.loadClass(name);
+                        }
```
aloubyansky: Is there a test or specific use-case this is fixing in our build or TS?

stuartwdouglas: This fixes some issues in the camel test suite that I am still investigating.

independent-projects/bootstrap/core/src/main/java/io/quarkus/bootstrap/BootstrapAppModelFactory.java
stuartwdouglas: It is created in the test class directory, so it should not end up in the jar. It does stay on the file system though so if you attempt to run a test from the IDE after a build it should still be picked up.

bom/runtime/pom.xml
```diff
@@ -114,7 +114,7 @@
         <derby-jdbc.version>10.14.2.0</derby-jdbc.version>
         <shrinkwrap.version>1.2.6</shrinkwrap.version>
         <rest-assured.version>4.1.2</rest-assured.version>
-        <junit.jupiter.version>5.5.2</junit.jupiter.version>
+        <junit.jupiter.version>5.6.0-RC1</junit.jupiter.version>
```
gastaldi: https://github.com/quarkusio/quarkus/pull/6663 updates to 5.6.0. It would be nice to rebase once that PR is merged

independent-projects/bootstrap/core/src/main/java/io/quarkus/bootstrap/classloading/QuarkusClassLoader.java
stuartwdouglas: I have removed this. The issue in Camel is that they are packaging parts of XML parsers, e.g. a jar contains a copy of QName but not the parser that is using it. This causes problems as the parser uses the JDK version of QName but the app sees the copy in the isolated ClassLoader. 

Once this is in I am planning on working with the camel team to resolve it.

stuartwdouglas:
@aloubyansky I believe I have fixed all the issues you brought up. 

@gsmet I have fixed the docs issues. I have tested against the platform test suite, and there are some issues that will likely need to be fixed on the camel side (e.g. the XML parsing issue I mention above). Once this is in I will work with them to get this resolved.

core/deployment/src/main/java/io/quarkus/runner/bootstrap/GenerateConfigTask.java
```diff
@@ -0,0 +1,154 @@
...
+                sb.append(formatDocs(i.getDocs()));
+                sb.append("\n#\n#");
+                sb.append(i.getPropertyName() + "=" + i.getDefaultValue());
+                sb.append("\n");
+            }
+
+            Files.createDirectories(configFile.getParent());
+            Files.write(configFile, sb.toString().getBytes(StandardCharsets.UTF_8),
+                    Files.exists(configFile) ? new OpenOption[] { StandardOpenOption.APPEND } : new OpenOption[] {});
+            Files.deleteIfExists(temp);
```
aloubyansky: Shouldn't it be deleted in the `finally`?

core/deployment/src/main/java/io/quarkus/runner/bootstrap/StartupActionImpl.java
```diff
@@ -0,0 +1,170 @@
...
+                        if (!debugPath.exists()) {
+                            debugPath.mkdir();
+                        }
+                        File classFile = new File(debugPath, i.getName() + ".class");
+                        classFile.getParentFile().mkdirs();
+                        try (FileOutputStream classWriter = new FileOutputStream(classFile)) {
+                            classWriter.write(i.getClassData());
+                        }
+                        log.infof("Wrote %s", classFile.getAbsolutePath());
+                    } catch (Throwable t) {
```
aloubyansky: Does it really have to be a Throwable? Swallowing throwables in a loop may go pretty bad.

core/devmode/src/main/java/io/quarkus/dev/DevModeMain.java
aloubyansky: So what about closing the instance of DevModeMain?

independent-projects/bootstrap/core/src/main/java/io/quarkus/bootstrap/resolver/BootstrapAppModelResolver.java
aloubyansky: Please remove `deploymentDependencies`.

independent-projects/bootstrap/core/src/main/java/io/quarkus/bootstrap/resolver/maven/MavenArtifactResolver.java
aloubyansky: Please remove this method.

core/devmode/src/main/java/io/quarkus/dev/DevModeMain.java
stuartwdouglas: I have changed it to be closed after the park, but in practice it would have been cleaned up by the normal shutdown hook anyway.

independent-projects/bootstrap/core/src/main/java/io/quarkus/bootstrap/BootstrapAppModelFactory.java
stuartwdouglas: I need to think about this one a bit more, but it is really a last resort case when we have no info. I could just delete it but I think it is worth keeping and trying to solve any reported problems as they arise.

aloubyansky: I think you can keep it for now. AFAIU, this method may get called when (re)building an app from a JAR in the first place. Probably a pretty good chance that everything comes in a JAR at that point.

aloubyansky (approved):
Let's get this in, ASAP!

geoand:
Damn CI is acting up again...

stuartwdouglas:
I made another modification today as I found a class loader leak, if this passed it should be good to merge

stuartwdouglas:
CI seems to be working again, @gsmet if it passes can you hit the button?

gsmet:
Sure.

gsmet:
I rebased to trigger a new CI run... This CI issue is a nightmare :/

PR #6413, opened 2020-01-06
Fix Kubernetes manifest port handling

### LINKED ISSUE #6408: Container Port is 0 in Manifests generated by Kubernetes Extension
When generating Kubernetes Manifest with `./mvnw package` the resulting containerPort in the Pod Spec is always 0. As a side effect the Service is generated with port and targetPort set to 0 too.

**Expected behavior**

The Pod containerPort should be the same as `quarkus.http.port`, and consequently port and targetPort in service Spec.

**Actual behavior**

The Pod Template Spec generated is:
```yaml
      spec:
        containers:
        - env:
          - name: "KUBERNETES_NAMESPACE"
            valueFrom:
              fieldRef:
                fieldPath: "metadata.namespace"
          image: "cedric.lamalle/code-with-quarkus:1.0.0-SNAPSHOT"
          imagePullPolicy: "IfNotPresent"
          name: "code-with-quarkus"
          ports:
          - containerPort: 0
            name: "http"
            protocol: "TCP"
```

And the Service Spec is:

```yaml
  spec:
    ports:
    - name: "http"
      port: 0
      targetPort: 0
```

**To Reproduce**
1. Go to https://code.quarkus.io/
2. Add Kubernetes Extension
3. Click Generate Application
4. Unzip the Application and go to the code-with-quarkus folder
5. Run `./mvnw package`
6. Open target/kubernetes/kubernetes.yml to see the generated manifests

**Configuration**
```properties
# Configuration file
# key = value
```

 - Output of `uname -a` or `ver`: ``Darwin cedric-lamalle.local 18.7.0 Darwin Kernel Version 18.7.0: Tue Aug 20 16:57:14 PDT 2019; root:xnu-4903.271.2~2/RELEASE_X86_64 x86_64``
 - Output of `java -version`: ``java version "1.8.0_171"
Java(TM) SE Runtime Environment (build 1.8.0_171-b11)
Java HotSpot(TM) 64-Bit Server VM (build 25.171-b11, mixed mode)``
 - GraalVM version (if different from Java):  -
 - Quarkus version or git rev: ``1.1.0.Final``

**Additional context**

Editing `pom.xml` and setting the following properties: `quarkus-plugin.version` and `quarkus.platform.version` to `1.0.0.Final` and running `./mvnw clean package` generates the expected Manifests.

I have noticed that if I had the following in `application.properties`:

```properties
kubernetes.ports[0].name=test
kubernetes.ports[0].containerPort=8081
```

The new Port is generated correctly:

Pod Template Spec:

```yaml
      spec:
        containers:
        - env:
          - name: "KUBERNETES_NAMESPACE"
            valueFrom:
              fieldRef:
                fieldPath: "metadata.namespace"
          image: "cedric.lamalle/code-with-quarkus:1.0.0-SNAPSHOT"
          imagePullPolicy: "IfNotPresent"
          name: "code-with-quarkus"
          ports:
          - containerPort: 0
            name: "http"
            protocol: "TCP"
          - containerPort: 8081
            name: "test"
            protocol: "TCP"
        serviceAccount: "code-with-quarkus"
```

Service Spec:

```
  spec:
    ports:
    - name: "http"
      port: 0
      targetPort: 0
    - name: "test"
      port: 8081
      targetPort: 8081
```

If I put the following in `application.properties` (kebab case):

```properties
kubernetes.ports[0].name=test
kubernetes.ports[0].container-port=8081
```

The new Port is generated with containerPort equal to 0. Note that the kebab case form is the one documented in:

- https://quarkus.io/guides/kubernetes#configuration-options Table 6
- https://github.com/dekorateio/dekorate/blob/master/assets/config.md#port

geoand:
This is needed because of the config changes that went in 1.1.
One thing that needed to be removed because of those changes was
the warning we used to produce about potential mismatch between
the port in the Kubernetes manifests and the port set at runtime

Fixes: #6408

### DIFF
--- a/extensions/vertx-http/deployment/src/main/java/io/quarkus/vertx/http/deployment/VertxHttpProcessor.java
+++ b/extensions/vertx-http/deployment/src/main/java/io/quarkus/vertx/http/deployment/VertxHttpProcessor.java
@@ -4,7 +4,6 @@ import java.io.IOException;
 import java.util.Collections;
 import java.util.List;
 import java.util.Optional;
-import java.util.OptionalInt;
 import java.util.stream.Collectors;
 
 import org.eclipse.microprofile.config.ConfigProvider;
@@ -76,11 +75,9 @@ class VertxHttpProcessor {
     }
 
     @BuildStep(onlyIf = IsNormal.class)
-    @Record(value = ExecutionTime.RUNTIME_INIT, optional = true)
-    public KubernetesPortBuildItem kubernetes(HttpConfiguration config, VertxHttpRecorder recorder) {
-        int port = ConfigProvider.getConfig().getValue("quarkus.http.port", OptionalInt.class).orElse(8080);
-        recorder.warnIfPortChanged(config, port);
-        return new KubernetesPortBuildItem(config.port, "http");
+    public KubernetesPortBuildItem kubernetes() {
+        int port = ConfigProvider.getConfig().getOptionalValue("quarkus.http.port", Integer.class).orElse(8080);
+        return new KubernetesPortBuildItem(port, "http");
     }
 
     @BuildStep
--- a/integration-tests/kubernetes/src/it/kubernetes-with-application-properties/src/main/resources/application.properties
+++ b/integration-tests/kubernetes/src/it/kubernetes-with-application-properties/src/main/resources/application.properties
@@ -1,4 +1,5 @@
 # Configuration file
+quarkus.http.port=9090
 kubernetes.name=test-it
 kubernetes.labels[0].key=foo
 kubernetes.labels[0].value=bar
--- a/integration-tests/kubernetes/src/it/kubernetes-with-application-properties/verify.groovy
+++ b/integration-tests/kubernetes/src/it/kubernetes-with-application-properties/verify.groovy
@@ -29,3 +29,9 @@ assert env.value == "SOMEVALUE"
 
 //Check the image
 assert deployment.spec.template.spec.containers[0].image == "quay.io/grp/kubernetes-with-application-properties:0.1-SNAPSHOT"
+assert deployment.spec.template.spec.containers[0].ports[0].containerPort == 9090
+
+//Check the Service
+Service service = list.items.find{r -> r.kind == "Service"}
+assert service != null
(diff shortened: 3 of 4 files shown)


gsmet (approved):

geoand:
I don't really like how these tests are executed, I'll try and look into making them simpler based on something like `QuarkusUnitTest` when the new Classloader stuff (with the unified way of creating/launching is in)

extensions/vertx-http/deployment/src/main/java/io/quarkus/vertx/http/deployment/VertxHttpProcessor.java
```diff
@@ -76,11 +75,9 @@ void filterMultipleVertxInstancesWarning(LaunchModeBuildItem launchModeBuildItem
     }
 
     @BuildStep(onlyIf = IsNormal.class)
-    @Record(value = ExecutionTime.RUNTIME_INIT, optional = true)
-    public KubernetesPortBuildItem kubernetes(HttpConfiguration config, VertxHttpRecorder recorder) {
-        int port = ConfigProvider.getConfig().getValue("quarkus.http.port", OptionalInt.class).orElse(8080);
-        recorder.warnIfPortChanged(config, port);
-        return new KubernetesPortBuildItem(config.port, "http");
+    public KubernetesPortBuildItem kubernetes() {
+        int port = ConfigProvider.getConfig().getOptionalValue("quarkus.http.port", Integer.class).orElse(8080);
```
dmlloyd: This is really wrong; it's a run time configuration property.  I think we need to revisit this.

geoand: I'll think of a better solution for 1.2. Unfortunately without this change, Kubernetes manifest generation for 1.1 was sort of broken

gsmet: @dmlloyd yeah, we agreed with @geoand that there was something fishy. This is just a quick fix for 1.1.1. We need to discuss how we can solve this properly.

Issue #5751, opened 2019-11-25
Allow quarkus-oidc to customize response body of 403


looorent:
**Description**
When using quarkus-oidc, providing a wrong token (expired, wrong signature, realm, ...) returns a forbidden status, as expected. However, the response body is always empty.

In order to migrate an existing API (which already defines its own response bodies), we would require from quarkus-oidc the possibility to customize and give more information in the response body.

For example, an expired token could produce this response body:
```
{
   "reason": "token_expired",
   "message": "Your token is expired."
}
```

sberyozkin:
This is a problem that we have to tackle somehow, the first problem is that we need to be able to launch an OIDC authentication from a pre-match JAX-RS chain so as to get the exception mappers utilized. The smaller problem would be to make `OIDCException` informative enough...
@stuartwdouglas Hi Stuart, what do you think ? The same issue was reported at the pure smallrye-jwt path, the users would like to customize the exceptions. 
Perhaps we can introduce quarkus specific exception mappers which would work across the security layer ?

loicmathieu:
@sberyozkin the issue has already been discuss but not yet implemented :)

When we implements JSON representation for resteasy 404 (inside `io.quarkus.resteasy.runtime.NotFoundExceptionMapper`) and for undertow/Vet.x 500, I raise the fact that there is also exception mappers for authentification issues, but we didn't implement JSON representation for them. This was a leftover.

So we need to provide a default JSON reprensentation (close to the one we did for 404 and 500) inside `io.quarkus.resteasy.runtime.ForbiddenExceptionMapper` and `io.quarkus.resteasy.runtime.UnauthorizedExceptionMapper`.

A user can still override these ones by providing it's own Exception mapper as their priority is lower than the default one.

sberyozkin:
@loicmathieu Hi Loic, not sure how much of that we can apply in this case, when the token has been validated and the roles extracted then yes, if 403 is reported is enforced in the JAX-RS chain we can map catch it, but in this case a user would like to catch the exceptions at the OIDC token verification time which is happening before the JAX-RS chain has started...

loicmathieu:
@sberyozkin maybe we need more that what I propose, but what I propose must be done :)

It's hard to provide exception mappers for all possible exception generated by Quarkus. 

[...]

PR #6419, opened 2020-01-06
Allow vertx-http to customize body of security errors

looorent:
As explained in #5751 , `io.quarkus.vertx.http.runtime.security.HttpSecurityRecorder` ends the response as soon as a authentication is detected. This prevents us to customize the response's body elsewhere.

The idea of this pull request is to manage this body with a dedicated bean.

### DIFF
--- a/extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/security/AuthenticationErrorResponseHandler.java
+++ b/extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/security/AuthenticationErrorResponseHandler.java
@@ -1,0 +1,10 @@
+package io.quarkus.vertx.http.runtime.security;
+
+import io.quarkus.security.AuthenticationFailedException;
+import io.vertx.core.buffer.Buffer;
+
+public interface AuthenticationErrorResponseHandler {
+    Buffer body(AuthenticationFailedException exception);
+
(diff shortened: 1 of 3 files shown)


gsmet:
There is a formatting issue. Could you run `mvn process-sources` and amend your commit?

looorent:
I have defined a `DefaultBean` to declare a default implementation (that could be overriden by any project). However, I've never worked on a Quarkus extension so I have no idea whether it is the best way to do it or not. Any clue on this?

```
@Dependent
class AuthenticationErrorResponseConfiguration {

    @Produces
    @DefaultBean
    public AuthenticationErrorResponseHandler createHandler() {
        return new DefaultHandler();
    }
}
```

gsmet:
I rebased this one.

It looks interesting but I think we need proper tests, otherwise we don't know if it works at all.

@stuartwdouglas could you have a look at the general idea?

stuartwdouglas (changes requested):
I don't think this is sufficient. It does not address the io.quarkus.vertx.http.runtime.security.HttpAuthenticator#getChallenge use case.

I think that this body data needs to be added to ChallengeData, and then AuthenticationErrorResponseHandler be modified to take the existing challenge data and return a new modified one. I am not sure if this needs the actual exception, as in some cases there isn't one (e.g. io.quarkus.vertx.http.runtime.security.HttpAuthorizer#doDeny).

PR #6432, opened 2020-01-06
ArC - optimize CreationalContext

### LINKED ISSUE #5281: ArC - optimize CreationalContext
Do not store dependent instances that have no `@PreDestroy`/disposal method and no transitive dependency with `@PreDestroy`/disposal method.

See https://github.com/weld/core/blob/master/impl/src/main/java/org/jboss/weld/contexts/unbound/DependentContextImpl.java#L80-L109 and https://issues.jboss.org/browse/WELD-1076.

mkouba:
- do not always store all dependent instances in CC
- also introduce io.quarkus.arc.InjectableInstance
- resolves #5281

### DIFF
--- a/independent-projects/arc/runtime/src/main/java/io/quarkus/arc/impl/InstanceHandleImpl.java
+++ b/independent-projects/arc/runtime/src/main/java/io/quarkus/arc/impl/InstanceHandleImpl.java
@@ -4,6 +4,7 @@ import io.quarkus.arc.Arc;
 import io.quarkus.arc.InjectableBean;
 import io.quarkus.arc.InstanceHandle;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.function.Consumer;
 import javax.enterprise.context.Dependent;
 import javax.enterprise.context.spi.CreationalContext;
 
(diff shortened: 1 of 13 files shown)


independent-projects/arc/runtime/src/main/java/io/quarkus/arc/impl/InstanceHandleImpl.java
```diff
@@ -61,10 +60,14 @@ public T get() {
...
-            if (bean.getScope().equals(Dependent.class)) {
-                destroyInternal();
+            if (destroyLogic != null) {
+                destroyLogic.accept(instance);
             } else {
-                Arc.container().getActiveContext(bean.getScope()).destroy(bean);
+                if (bean.getScope().equals(Dependent.class)) {
+                    destroyInternal();
+                } else {
+                    Arc.container().getActiveContext(bean.getScope()).destroy(bean);
```
manovotn: `getActiveContext` can return `null` if no context is active, we should probably check that.

independent-projects/arc/runtime/src/main/java/io/quarkus/arc/impl/InstanceImpl.java
```diff
@@ -120,7 +113,7 @@ public boolean isAmbiguous() {
     }
 
     @Override
-    public void destroy(T instance) {
+    public void destroy(Object instance) {
```
manovotn: What `Object`, the instance should still be of type `T`, right?

independent-projects/arc/runtime/src/main/java/io/quarkus/arc/impl/InstanceHandleImpl.java
mkouba: Good point. I did not check the existing code...

independent-projects/arc/runtime/src/main/java/io/quarkus/arc/impl/InstanceImpl.java
mkouba: Because of the generics hell... yes, it should be `T` but it should be safe use `Object`.

manovotn: Yea, `Object` works too. It's just that I saw `T` everywhere else, so this struck me as weird :)

manovotn (approved):

manovotn:
@mkouba unsorted imports ;-)

machi1990:
Merged, thanks!

PR #6443, opened 2020-01-07
Extending smoke tests for SR context propagation integration.

manovotn:
This PR revisits/extends smoke tests that SR context propagation currently has.

@FroMage please sanity check this and if you have ideas for any other part of the spec that should be smoke tested within Quarkus, let me know and I can add it.


FroMage (approved):
LGTM, can't think of any other sort of test.

PR #6444, opened 2020-01-07
Qute - support @TemplateExtension declared on a class

### LINKED ISSUE #6438: Qute - support @TemplateExtension declared on a class
A `ValueResolver` will be generated for every public, protected and package-private static method declared on the class. Methods that return `void` or have no arguments are ignored.

mkouba:
- also update docs and add some tests
- resolves #6438

### DIFF
--- a/docs/src/main/asciidoc/qute-reference.adoc
+++ b/docs/src/main/asciidoc/qute-reference.adoc
@@ -538,10 +538,25 @@ NOTE: A value resolver is also generated for all types used in parameter declara
 [[template_extension_methods]]
 === Template Extension Methods
 
-A value resolver is automatically generated for a template extension method annotated with `@TemplateExtension`.
-The method must be static, must not return `void` and must accept at least one parameter. 
-The class of the first parameter is used to match the base object and the method name is used to match the property name.
+Extension methods can be used to extend the data classes with new functionality.
+For example, it is possible to add "computed properties" and "virtual methods".
+A value resolver is automatically generated for a method annotated with `@TemplateExtension`.
+If declared on a class a value resolver is generated for every non-private method declared on the class.
+Methods that do not meet the following requirements are ignored.
 
+A template extension method:
+
+* must be static, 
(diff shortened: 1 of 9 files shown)


docs/src/main/asciidoc/qute-reference.adoc
```diff
@@ -538,10 +538,25 @@ NOTE: A value resolver is also generated for all types used in parameter declara
...
+
+* must be static, 
+* must not return `void`,
+* must accept at least one parameter.
+ 
+The class of the first parameter is always used to match the base object.
+The method name is used to match the property name by default. 
+However, it is possible to specify the matching name with `TemplateExtension#matchName()`.
+
+NOTE: A special constant - `ANY` - may be used to specify that the extension method matches any name. In that case, the method must declare at least two parameters and the second parameter must be a string.
```
FroMage: So the method can be named `Integer ANY(Integer i, String methodName)`? I guess you should add an example for this, no?

FroMage (approved):
LGTM :)

docs/src/main/asciidoc/qute-reference.adoc
mkouba: Not really - the constant must be used in the `TemplateExtension#matchName()`. There is an example in the test.

FroMage: Well, it should probably be supported, because otherwise we end up with this sort of thing:

```java
@RegisterExtension
public class MyExtensions {

 public static String foo(String f){}
 public static String bar(String f){}

 @RegisterExtension(matchName = "ANY")
 public static Integer wtf(Integer f, String method){}
}
```

And I suspect this will break because we will have the same method registered twice, no?

mkouba: > And I suspect this will break because we will have the same method registered twice, no?

Nope. The annotation declared on the method takes precedence. Note that so far there is only `matchName` but there could be others in the future...

PR #6446, opened 2020-01-07
docs: fix link to native-and-ssl in mailer.adoc

kdnakt:
https://quarkus.io/guides/mailer

Currently it looks like this:

スクリーンショット 2020-01-08 0 30 52


machi1990 (approved):
Good catch, thanks @kdnakt

PR #6449, opened 2020-01-07
Remove unused Vert.x GraalVM substitution

gwenneg:
This substitution is no longer required no matter which GraalVM version is supported by the next Quarkus release.


gsmet (approved):
Nice, thanks!

PR #6451, opened 2020-01-07
Support configuring PhysicalNamingStrategy in Hibernate ORM

### LINKED ISSUE #6346: Add support for Hibernate ORM naming strategy
**Description**
Using Java naming convention, all the Entity names and all the attributes of an entity must be written in camelCase.
But, in some cases, on the database side it would be more convenient to use "all lower case" or "snake case" names, expecially when you are using an already existent database. #5935 

[Database, Table and Column Naming Conventions?](https://stackoverflow.com/questions/7662/database-table-and-column-naming-conventions)

[How I Write SQL, Part 1: Naming Conventions](https://launchbylunch.com/posts/2014/Feb/16/sql-naming-conventions/)

Now, to support this naming conventions, the only two option are the following:

1) breaking the Java entity/field naming convention
2) annotate all entities and attributes using the `@Entity` and `@Column` annotation

Using 2 is time expensive, prone to error, and sometimes breaks (partially) the readability of the entity.

Panache uses Hibernate 5, which provides two different naming strategies for use with Hibernate entities: an Implicit Naming Strategy (from DB to attribute) and a Physical Naming Strategy (from attribute to DB).

[This tutorial](https://www.baeldung.com/hibernate-naming-strategy) shows an example of implementation.

It requires just a line of configuration in the `application.properties`

`hibernate.physical_naming_strategy=<naming strategy class>`

**Implementation ideas**
Allowing the use of custom class could bring many problems, to keep thing simple and save time maybe it would be enough to support just the 3 most common naming strategy:

- camelCase (already supported)
- all lowercase
- snake_case

gastaldi:
Fixes #6346

### DIFF
--- a/extensions/hibernate-orm/deployment/src/main/java/io/quarkus/hibernate/orm/deployment/HibernateOrmConfig.java
+++ b/extensions/hibernate-orm/deployment/src/main/java/io/quarkus/hibernate/orm/deployment/HibernateOrmConfig.java
@@ -89,8 +89,6 @@ public class HibernateOrmConfig {
      * Pluggable strategy contract for applying physical naming rules for database object names.
      *
      * Class name of the Hibernate PhysicalNamingStrategy implementation
-     *
-     * @asciidoclet
      */
     @ConfigItem
     Optional<String> physicalNamingStrategy;
--- a/extensions/hibernate-orm/deployment/src/main/java/io/quarkus/hibernate/orm/deployment/HibernateOrmConfig.java
+++ b/extensions/hibernate-orm/deployment/src/main/java/io/quarkus/hibernate/orm/deployment/HibernateOrmConfig.java
@@ -86,6 +86,16 @@ public class HibernateOrmConfig {
     public int batchFetchSize;
 
     /**
+     * Pluggable strategy contract for applying physical naming rules for database object names.
+     *
+     * Class name of the Hibernate PhysicalNamingStrategy implementation
+     *
+     * @asciidoclet
+     */
+    @ConfigItem
+    Optional<String> physicalNamingStrategy;
+
+    /**
      * Query related configuration.
      */
     @ConfigItem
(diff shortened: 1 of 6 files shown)


extensions/hibernate-orm/runtime/src/main/java/io/quarkus/hibernate/orm/runtime/recording/RecordableBootstrap.java
```diff
@@ -72,7 +72,8 @@ public RecordableBootstrap(BootstrapServiceRegistry bootstrapServiceRegistry) {
     }
 
     public RecordableBootstrap(BootstrapServiceRegistry bootstrapServiceRegistry, LoadedConfig loadedConfigBaseline) {
-        this.settings = QuarkusEnvironment.getInitialProperties();
+        this.settings = new HashMap();
+        this.settings.putAll(QuarkusEnvironment.getInitialProperties());
```
Sanne: :1st_place_medal:

extensions/hibernate-orm/deployment/src/main/java/io/quarkus/hibernate/orm/deployment/HibernateOrmConfig.java
```diff
@@ -85,6 +85,16 @@
     @ConfigItem(defaultValue = "-1")
     public int batchFetchSize;
 
+    /**
+     * Pluggable strategy contract for applying physical naming rules for database object names.
+     *
+     * Class name of the Hibernate PhysicalNamingStrategy implementation
+     *
+     * @asciidoclet
```
machi1990: This tag can be removed.

Sanne (approved):
great, thanks!

extensions/hibernate-orm/deployment/src/main/java/io/quarkus/hibernate/orm/deployment/HibernateOrmConfig.java
gastaldi: For the other properties as well?

gsmet (approved):

PR #6455, opened 2020-01-08
Fix for #6442 quarkusDev not working in multi module projects

### LINKED ISSUE #6442: quarkusDev doesn't work in multi module gradle project
**Describe the bug**
If using a multi module gradle project quarkusDev will fail complaining that beans are defined multiple times. This is due to both common/buld/classes and common/buld/lib/common.jar being on the classpath inside of dev-mode-context.dat

Project Structure
```
project-root
... common
... app
```
App depends on common via 
```
compile project(":commons")
```

The error that happens is

```        Suppressed: javax.enterprise.inject.AmbiguousResolutionException: Ambiguous dependencies for type commons.CommonsConfig and qualifiers [@Default]
        - java member: commons.cache.impl.CacheServiceImpl#config
        - declared on CLASS bean [types=[commons.cache.CacheService, commons.cache.impl.CacheServiceImpl, java.lang.Object], qualifiers=[@Default, @Any], target=commons.cache.impl.CacheServiceImpl]
        - available beans:
                - CLASS bean [types=[commons.CommonsConfig, java.lang.Object], qualifiers=[@Default, @Any], target=commons.CommonsConfig]
                - CLASS bean [types=[commons.CommonsConfig, java.lang.Object], qualifiers=[@Default, @Any], target=commons.CommonsConfig]
```

Inside META-INF/dev-mode-context.dat
```
commons/build/classes/java/main
commons/build/libs/commons-v383-1-gf5b4d8a.dirty.jar
```

**Expected behavior**
QuarkusDev should start without issues

**Actual behavior**
QuarkusDev fails due to AmbiguousResolutionException

 - Output of `uname -a` or `ver`: Darwin Kernel Version 19.0.0
 - Output of `java -version`:  OpenJDK 64-Bit GraalVM CE 19.2.1 (build 25.232-b07-jvmci-19.2-b03, mixed mode)
 - Quarkus version or git rev: 1.1.0.FINAL

tobad357:
Fixes #6442 
The fix is to register if a dependency is a Project Component and ignore it if it is as it will be added to the classpath later on

### DIFF
--- a/devtools/gradle/src/main/java/io/quarkus/gradle/tasks/QuarkusDev.java
+++ b/devtools/gradle/src/main/java/io/quarkus/gradle/tasks/QuarkusDev.java
@@ -52,6 +52,7 @@ import org.gradle.api.tasks.compile.JavaCompile;
 import org.gradle.api.tasks.options.Option;
 
 import io.quarkus.bootstrap.model.AppArtifact;
+import io.quarkus.bootstrap.model.AppArtifactKey;
 import io.quarkus.bootstrap.model.AppDependency;
 import io.quarkus.bootstrap.model.AppModel;
 import io.quarkus.bootstrap.resolver.AppModelResolver;
@@ -241,22 +242,13 @@ public class QuarkusDev extends QuarkusTask {
                 throw new GradleException("Failed to resolve application model " + extension.getAppArtifact() + " dependencies",
                         e);
             }
-            for (AppDependency appDep : appModel.getAllDependencies()) {
-                addToClassPaths(classPathManifest, context, appDep.getArtifact().getPath().toFile());
-            }
 
             args.add("-Djava.util.logging.manager=org.jboss.logmanager.LogManager");
             File wiringClassesDirectory = new File(getBuildDir(), "wiring-classes");
             wiringClassesDirectory.mkdirs();
             addToClassPaths(classPathManifest, context, wiringClassesDirectory);
 
-            //we also want to add the maven plugin jar to the class path
-            //this allows us to just directly use classes, without messing around copying them
-            //to the runner jar
-            addGradlePluginDeps(classPathManifest, context);
-
             //now we need to build a temporary jar to actually run
-
             File tempFile = new File(getBuildDir(), extension.finalName() + "-dev.jar");
             tempFile.delete();
             tempFile.deleteOnExit();
@@ -271,6 +263,7 @@ public class QuarkusDev extends QuarkusTask {
                 res = file.getAbsolutePath();
             }
 
+            final Set<AppArtifactKey> projectDependencies = new HashSet<>();
             final Configuration compileCp = project.getConfigurations()
                     .getByName(JavaPlugin.COMPILE_CLASSPATH_CONFIGURATION_NAME);
             final DependencySet compileCpDependencies = compileCp.getAllDependencies();
@@ -279,6 +272,11 @@ public class QuarkusDev extends QuarkusTask {
                 if (!(dependency instanceof ProjectDependency)) {
                     continue;
                 }
+
+                // Create the key via AppArtifact to make sure we use same defaults for type and classifier
+                AppArtifactKey key = new AppArtifact(dependency.getGroup(), dependency.getName(), dependency.getVersion())
+                        .getKey();
+                projectDependencies.add(key);
 
                 Project dependencyProject = ((ProjectDependency) dependency).getDependencyProject();
                 Convention convention = dependencyProject.getConvention();
@@ -307,6 +305,17 @@ public class QuarkusDev extends QuarkusTask {
 
                 context.getModules().add(wsModuleInfo);
             }
+
+            for (AppDependency appDependency : appModel.getAllDependencies()) {
+                if (!projectDependencies.contains(appDependency.getArtifact().getKey())) {
+                    addToClassPaths(classPathManifest, context, appDependency.getArtifact().getPath().toFile());
+                }
+            }
+
+            //we also want to add the maven plugin jar to the class path
+            //this allows us to just directly use classes, without messing around copying them
+            //to the runner jar
+            addGradlePluginDeps(classPathManifest, context);
 
             DevModeContext.ModuleInfo moduleInfo = new DevModeContext.ModuleInfo(
                     project.getName(),


devtools/gradle/src/main/java/io/quarkus/gradle/tasks/QuarkusDev.java
```diff
@@ -242,7 +242,11 @@ public void startDev() {
                         e);
             }
             for (AppDependency appDep : appModel.getAllDependencies()) {
-                addToClassPaths(classPathManifest, context, appDep.getArtifact().getPath().toFile());
+                if (!appDep.getArtifact().isSubModule()) {
+                    addToClassPaths(classPathManifest, context, appDep.getArtifact().getPath().toFile());
+                } else {
+                    getLogger().info("Ignoring {} as it is a project dependency", appDep);
```
gsmet: Should this one be debug instead of info?

tobad357: It could be.. but the logging of the added dependencies in addToClasspaths is info so I thought it made sense to be on same level

gastaldi: +1 to be debug

gastaldi:
This looks good, but I'll let @aloubyansky chime in this since there are changes to AppArtifact (which should probably also reflect in the Maven plugin)

gastaldi:
Can you squash these commits into a single one? Thanks!

devtools/gradle/src/main/java/io/quarkus/gradle/AppModelGradleResolver.java
```diff
@@ -192,6 +194,10 @@ private ModuleIdentifier getModuleId(ResolvedArtifact a) {
     private AppDependency toAppDependency(ResolvedArtifact a) {
         final String[] split = a.getModuleVersion().toString().split(":");
         final AppArtifact appArtifact = new AppArtifact(split[0], split[1], split[2]);
+        final ComponentIdentifier id = a.getId().getComponentIdentifier();
+        if (id instanceof ProjectComponentIdentifier) {
+            appArtifact.setSubModule(true);
+        }
```
gastaldi: You can also replace this snippet with `appArtifact.setSubModule(id instanceof ProjectComponentIdentifier)` :wink:

gastaldi: Or simply `appArtifact.setSubModule(a.getId().getComponentIdentifier() instanceof ProjectComponentIdentifier);`

devtools/gradle/src/main/java/io/quarkus/gradle/AppModelGradleResolver.java
```diff
@@ -23,6 +23,8 @@
 import org.gradle.api.artifacts.ModuleVersionIdentifier;
 import org.gradle.api.artifacts.ResolvedArtifact;
 import org.gradle.api.artifacts.ResolvedConfiguration;
+import org.gradle.api.artifacts.component.ComponentIdentifier;
```
gastaldi: I think this import is unused now?

tobad357: Sorted

aloubyansky:
> This looks good, but I'll let @aloubyansky chime in this since there are changes to AppArtifact (which should probably also reflect in the Maven plugin)

Right, I'm not sure about that. We are already iterating though the classpath in QuarkusDev (line 278), could we do the filtering in that loop instead? Thanks.

tobad357:
> > This looks good, but I'll let @aloubyansky chime in this since there are changes to AppArtifact (which should probably also reflect in the Maven plugin)
> 
> Right, I'm not sure about that. We are already iterating though the classpath in QuarkusDev (line 278), could we do the filtering in that loop instead? Thanks.

Don't we want to avoid adding the Jar artifact to the classpath as I would guess it would prevent reloading of the classes in the submodule.

aloubyansky:
Sure, that's what the fix is about, right? I'm not questioning that. My comment was about isSubModule() in AppArtifact. I don't see how we can easily implement that in Maven case and whether we need it in that case. So, I was wondering whether we could do filter deps that should not be added to the classpath in the for loop I mentioned above.

tobad357:
> Sure, that's what the fix is about, right? I'm not questioning that. My comment was about isSubModule() in AppArtifact. I don't see how we can easily implement that in Maven case and whether we need it in that case. So, I was wondering whether we could do filter deps that should not be added to the classpath in the for loop I mentioned above.

The problem is the Jar artifact is being added at line 244-250 and there is currently no way to check if a AppDependency is a project dependency.
My initial plan was to get the Jar artifact name and path from the Configuration and filter as you mentioned but couldn't find a way (this may be due to me not knowing all the Gradle APIs)

aloubyansky:
I think instead of adding the deps to the classpath here https://github.com/quarkusio/quarkus/blob/master/devtools/gradle/src/main/java/io/quarkus/gradle/tasks/QuarkusDev.java#L245 You could create a map of the deps with AppArtifactKey as keys of all the deps.

Then here https://github.com/quarkusio/quarkus/blob/master/devtools/gradle/src/main/java/io/quarkus/gradle/tasks/QuarkusDev.java#L282 you can remove module deps from that map.
And then add to the classpath whatever remained in that map.
WDYT?

tobad357:
@aloubyansky yes that makes sense. I did roughly what you said but instead added all project deps to a Set and then filter while adding to classpath. I think that makes the code a bit clearer vs manipulating a Map of deps in several places.

If you are happy with the change I'll squash the commits

independent-projects/bootstrap/core/src/main/java/io/quarkus/bootstrap/model/AppArtifact.java
```diff
@@ -11,6 +11,8 @@
 
     protected Path path;
 
+    protected boolean subModule;
```
aloubyansky: Please remove this leftover.

aloubyansky (changes requested):
Yes, looks good, thanks a lot! Please squash the commits.

tobad357:
@aloubyansky thanks for your time, I've removed the property and squashed

aloubyansky (approved):
You're welcome @tobad357! Great work :)

PR #6461, opened 2020-01-08
Native executable must be made executable in Dockerfile

haraldatbmw:
The native executable must be changed to be executable **/work/application** explicit. 
Otherwise, if built with docker (mvn package -Pnative -Dquarkus.native.container-build=true) from a **windows system** (windows cmd) the application can not start inside docker/openshift.

```
container_linux.go:235: starting container process caused "exec: \"./application\": permission denied"
```

### DIFF
--- a/devtools/platform-descriptor-json/src/main/resources/templates/dockerfile-native.ftl
+++ b/devtools/platform-descriptor-json/src/main/resources/templates/dockerfile-native.ftl
@@ -17,6 +17,6 @@
 FROM registry.access.redhat.com/ubi8/ubi-minimal
 WORKDIR /work/
 COPY ${build_dir}/*-runner /work/application
-RUN chmod 775 /work
+RUN chmod 775 /work /work/application
 EXPOSE 8080
 CMD ["./application", "-Dquarkus.http.host=0.0.0.0"]


devtools/platform-descriptor-json/src/main/resources/templates/dockerfile-native.ftl
```diff
@@ -17,6 +17,6 @@
 FROM registry.access.redhat.com/ubi8/ubi-minimal
 WORKDIR /work/
 COPY ${build_dir}/*-runner /work/application
-RUN chmod 775 /work
```
gsmet: My guess is that you should keep this line too and have both.

gsmet (changes requested):

devtools/platform-descriptor-json/src/main/resources/templates/dockerfile-native.ftl
haraldatbmw: @gsmet What about chmod -R 775 /work for recursiv setting?

haraldatbmw: But would be to follow your advice.

gsmet: Yeah, I'm not a big fan of making things executable recursively. Users could add files there by mistake.

gsmet:
Looks OK to me now. Could you squash everything into one commit?

haraldatbmw:
@gsmet Not sure if I did the squash of the commits correct. Please check.

gsmet:
@haraldatbmw I fixed the history that got a bit messed up.

Can you check the commit reflects your original intent? I got weird merge conflicts just squashing so I prefer asking.

Thanks!

haraldatbmw:
@gsmet The commit looks good for me!

gsmet (approved):

gsmet:
Merged, thanks!

PR #6364, opened 2019-12-31
Update GraalVM to 19.3.0.2

machi1990:
- Fixes _DynamicProxySupport is not thread safe_ and remove related work around
- Fixes https://github.com/quarkusio/quarkus/issues/6115

PR #6468, opened 2020-01-08
Fix GraalVM SDK version in documentation

gwenneg:
This should be the same version as [the one from the BOM](https://github.com/quarkusio/quarkus/blob/master/bom/runtime/pom.xml#L77) since we [removed](https://github.com/quarkusio/quarkus/pull/6364/commits/2ab449625a62034f0b1c86203d124d8480ff7715) a `19.3.0` substitution that was incompatible with `19.3.0.2`.


gsmet (approved):
Good catch.

I think it's OK on the website though as the attributes are coming from a different place.

PR #6472, opened 2020-01-09
Add support for kebab case in @ConfigProperties

### LINKED ISSUE #6457: Provide option on the @ConfigProperties annotation to control the kebab-case naming of properties
As discusses https://groups.google.com/forum/#!topic/quarkus-dev/NgbXygc1rps

Looking at https://quarkus.io/guides/all-config all config properties seem to use a kebab-case naming strategy.

When creating a custom config properties class (or interface in this case) like:

```
@ConfigProperties(prefix = "graphql")
public interface GraphQLConfiguration {

    String getSchemaLocation();

    String getSchemaFileExtension();

    Integer getMaxQueryComplexity();

    Integer getMaxQueryDepth();

    Boolean getTracing();
}
```

I would expect that these would also follow the kebab-case, but this isn't the case :-)

It is expecting the Java bean convention name, which is not consistent with out-of-the-box Quarkus configuration properties.

It can be workaround adding a `@ConfigProperty(name = "schema-location", defaultValue = "/graphqls")` annotation explicitly; but this will lead to a lot of boilerplate.

The idea mentioned on the forum is to provide an option on the `@ConfigProperties` annotation to control the kebab-case naming.

geoand:
Fixes: #6457


geoand:
@gytis Do you want to take a look as well?

gsmet:
I have a question on this one: considering our conventions, shouldn't kebab case be the one and only thing we support?

geoand:
That would indeed make sense. The reason I didn't do it is because of backward compatibility

gytis:
@geoand, assign this to me. I'll take care of it once I'm done with Spring Data REST extension.

geoand:
@gytis should just be a matter of a small review

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/configproperties/ClassConfigPropertiesUtil.java
```diff
@@ -167,7 +170,8 @@ static boolean addProducerMethodForClassConfigProperties(ClassLoader classLoader
                 configObjectClassStr, produceMethodParameterTypes)) {
             methodCreator.addAnnotation(Produces.class);
 
-            ResultHandle configObject = populateConfigObject(classLoader, configPropertiesClassInfo, prefixStr, methodCreator,
+            ResultHandle configObject = populateConfigObject(classLoader, configPropertiesClassInfo, prefixStr, namingStrategy,
+                    methodCreator,
```
gytis: Small thing, but maybe move this inline with `applicationIndex, configProperties);`

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/configproperties/InterfaceConfigPropertiesUtil.java
```diff
@@ -196,26 +199,33 @@ private static boolean isDefault(short flags) {
...
+    private static String getPropertyName(MethodInfo method, ConfigProperties.NamingStrategy namingStrategy) {
+        if (namingStrategy == ConfigProperties.NamingStrategy.DEFAULT) {
+            try {
+                return JavaBeanUtil.getPropertyNameFromGetter(method.name());
+            } catch (IllegalArgumentException e) {
+                throw new IllegalArgumentException("Method " + method.name() + " of interface " + method.declaringClass()
+                        + " is not a getter method. Either rename the method to follow getter name conventions or annotate the method with @ConfigProperty");
+            }
+        } else if (namingStrategy == ConfigProperties.NamingStrategy.KEBAB_CASE) {
+            return StringUtil.hyphenate(method.name());
```
gytis: How about checking for a getter in here the same as in the default case? Looking at the original issue (https://github.com/quarkusio/quarkus/issues/6457), this seems to be the expected scenario i.e. `schema-location` > `getSchemaLocation`

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/configproperties/ConfigPropertiesBuildStep.java
```diff
@@ -74,6 +74,7 @@ void setup(CombinedIndexBuildItem combinedIndex,
 
                 String generatedClassName = InterfaceConfigPropertiesUtil.generateImplementationForInterfaceConfigProperties(
                         classInfo, nonBeansClassOutput, combinedIndex.getIndex(), configPropertiesMetadata.getPrefix(),
+                        configPropertiesMetadata.getNamingStrategy(),
```
gytis: Move inline with `defaultConfigValues, configProperties);`?

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/configproperties/InterfaceConfigPropertiesUtil.java
geoand: Good idea

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/configproperties/ConfigPropertiesBuildStep.java
geoand: Done

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/configproperties/InterfaceConfigPropertiesUtil.java
geoand: Done

gsmet:
@geoand I think maybe we should make it the default and allow to go back to previous behavior for compatibility.

We can document it in the migration guide.

Maybe send an email on the list with which use case it will break and how people will be able to fix it with the annotation parameter and see if anyone complains about that.

And I think we need @mkouba 's blessing too as it's part of ArC.

geoand:
@gsmet good idea!

I'll do that tonight or tomorrow

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/configproperties/ClassConfigPropertiesUtil.java
```diff
@@ -316,6 +322,15 @@ private static ResultHandle populateConfigObject(ClassLoader classLoader, ClassI
         return configObject;
     }
 
+    private static String getPropertyName(FieldInfo field, ConfigProperties.NamingStrategy namingStrategy) {
+        if (namingStrategy == ConfigProperties.NamingStrategy.DEFAULT) {
+            return field.name();
+        } else if (namingStrategy == ConfigProperties.NamingStrategy.KEBAB_CASE) {
+            return StringUtil.hyphenate(field.name());
+        }
```
gastaldi: Maybe move this behavior to a method in the NamingStrategy enum? eg. `String getPropertyName(String name)`

geoand: That makes sense, I'll update it soon

geoand: That makes sense, I'll update it soon

geoand: Done

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/configproperties/ClassConfigPropertiesUtil.java
```diff
@@ -316,6 +321,10 @@ private static ResultHandle populateConfigObject(ClassLoader classLoader, ClassI
         return configObject;
     }
 
+    private static String getPropertyName(FieldInfo field, ConfigProperties.NamingStrategy namingStrategy) {
```
gastaldi: Is this method still needed? I think you can call `namingStrategy.getName(field.name())` directly, no?

geoand: Yeah, you are right. should have got rid of it, but I got mixed up because of the similar (but not exactly the same) case for interfaces

geoand: So now it should be properly done :)

extensions/arc/runtime/src/main/java/io/quarkus/arc/config/ConfigProperties.java
```diff
@@ -19,4 +21,47 @@
...
+            }
+        },
+        KEBAB_CASE {
+            @Override
+            public String getName(String name) {
+                return StringUtil.hyphenate(name);
+            }
+        };
+
+        public String getName(String name) {
```
gastaldi: This method can be declared as abstract

geoand: Damn why didn't I think of that...

geoand: I'll push a fix in a few hours

geoand: Thanks for catching this, it should be fixed now

geoand:
PR has been updated to change the default behavior and to also add a config property determining the default behavior

gsmet (approved):

gsmet:
Merged thanks!

Could you add a note here: https://github.com/quarkusio/quarkus/wiki/Migration-Guide-1.2 ?

geoand:
Yes of course! Will do over the weekend, or if I'm not too tired, later on tonight

geoand:
An entry was added to the migration guide

PR #6474, opened 2020-01-09
Document git config for contributing to Quarkus from Windows

rquinio:
I mentioned this during https://github.com/quarkusio/quarkus/pull/6343#issuecomment-569077715

As Quarkus build re-formats all files, the configuration of the formatter-maven-plugin for line endings (https://code.revelc.net/formatter-maven-plugin/format-mojo.html#lineEnding - by default AUTO) needs to be aligned with git configuration for line endoings (core.autocrlf), otherwise it generates noise on all files after running a mvn command.

An alternative is to specify -Dlineendings=LF on every mvn command, but it's not very practical.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -50,6 +50,8 @@ Be sure to test your pull request in:
 If you have not done so on this machine, you need to:
  
 * Install Git and configure your GitHub access
+    * Windows
+        * Use `git config core.autocrlf true`, as default system line endings are expected by formatter-maven-plugin
 * Install Java SDK (OpenJDK recommended)
 * Install [GraalVM](https://quarkus.io/guides/building-native-image)
 * Install platform C developer tools:


gsmet:
I'm not entirely sure it's the right fix. I wonder if we should get the formatter to put Unix line endings everywhere instead.

@stuartwdouglas WDYT?

stuartwdouglas:
Does https://github.com/stuartwdouglas/quarkus/pull/new/line-endings fix it?

gastaldi:
@stuartwdouglas even with that parameter set I get lots of warnings in `git status`: 
```
warning: CRLF will be replaced by LF in independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/Annotations.java.
```

maxandersen:
that’s because @stuartwdouglas  fix only fixes main quarkus, not independent projects.

+1 on getting the forced LF setting in as I hit this on windows too.

gsmet:
Superseded by https://github.com/quarkusio/quarkus/pull/8748 .

PR #6478, opened 2020-01-09
Feat/update mongo client substitution

loicmathieu:
Same as PR #6418 but with providing substitution for mongo-crypt instead of the library


loicmathieu:
I'm not able to make a working susbtitution set for mongodb-crypt, closing it for now.

PR #6480, opened 2020-01-09
Generate PDFs for the documentation and build an assemble zip

gsmet:
The idea is to have a link on the website allowing to download all the
PDF documents at once.

First step for fixing
https://github.com/quarkusio/quarkusio.github.io/issues/419 .

To test it, in the `docs` folder, you shoud run `mvn clean install -Ddocumentation-pdf`.

Note: output was kept simple for printing (so no logo). Also the configuration table layout is far from being perfect, I think some of the issues are due to bugs in asciidoctor-pdf (especially the random space after the icon when the property name is on two lines) but it's a nice first step and we can improve later.

The theme was adapted from the HIbernate one.


cescoffier:
Very nice!

cescoffier (approved):

PR #6484, opened 2020-01-09
Fix spelling in extension doc

mattyb678:
I think that parley is not the word needed here; it should be parlance, which means: 
> a particular way of speaking or using words, especially a way common to those with a particular job or interest.


gastaldi (approved):

gastaldi:
Nice catch!

PR #6500, opened 2020-01-10
add auto labeling of pull-requests

maxandersen:
This is a basic first attempt on autolabeler for pullrequests.

I just matched labels against folder names to get an initial setup. Not all matched; but wee can tweak it as we go.

note: I generated this in late december so if new extensions or integraitons tests folders was added then they aren't listed in here.

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -5,7 +5,7 @@
 ### committed updates would be done manually.
 
 ### you can test this script by running the following in quarkus folder:
-### jbang bang https://git.io/JvJR5 .github/labeler.yml <optional path>
+### jbang https://git.io/JvJR5 .github/labeler.yml <optional path>
 
 ### Non-covered dirs....
 ### - extensions
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -4,6 +4,8 @@
 ### should be verified/tuned by changing the script and once
 ### committed updates would be done manually.
 
+### you can test this script by running the following in quarkus folder:
+### jbang bang https://git.io/JvJR5 .github/labeler.yml <optional path>
 
 ### Non-covered dirs....
 ### - extensions
@@ -49,7 +51,6 @@
 ### area/core
 ### area/dependencies
 ### area/dev-mode
-### area/gradle
 ### area/housekeeping
 ### area/infra
 ### area/new-http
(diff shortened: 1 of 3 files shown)


.github/labeler.yml
```diff
@@ -0,0 +1,273 @@
...
+
+area/config:
+ - extensions/config-yaml/**/*
+
+area/documentation:
+ - docs/**/*
+
+area/hibernate-orm:
+ - extensions/hibernate-orm/**/*
+ - integration-tests/hibernate-orm-panache/**/*
```
gsmet: You also can add all the `jpa-*` ITs

.github/labeler.yml
```diff
@@ -0,0 +1,273 @@
...
+ - integration-tests/hibernate-search-elasticsearch/**/*
+
+area/hibernate-validator:
+ - extensions/hibernate-validator/**/*
+ - integration-tests/hibernate-validator/**/*
+
+area/infinispan:
+ - extensions/infinispan-client/**/*
+ - extensions/infinispan-embedded/**/*
+ - integration-tests/infinispan-cache-jpa/**/*
```
gsmet: You can add this one to Hibernate too. It needs both labels IMHO.

.github/labeler.yml
```diff
@@ -0,0 +1,273 @@
...
+ - integration-tests/kubernetes-client/**/*
+
+area/logging:
+ - extensions/logging-gelf/**/*
+ - extensions/logging-json/**/*
+ - extensions/logging-sentry/**/*
+ - integration-tests/logging-gelf/**/*
+
+area/maven:
+ - integration-tests/maven/**/*
```
gsmet: Shouldn't we add the Maven plugin directory too?

.github/labeler.yml
```diff
@@ -0,0 +1,273 @@
...
+
+area/mongodb:
+ - extensions/mongodb-client/**/*
+ - integration-tests/mongodb-client/**/*
+ - integration-tests/mongodb-panache/**/*
```
gsmet: The MongoDB Panache extension should be added here too.

.github/labeler.yml
```diff
@@ -0,0 +1,273 @@
...
+area/persistence:
+ - integration-tests/jpa/**/*
+ - integration-tests/jpa-derby/**/*
+ - integration-tests/jpa-h2/**/*
+ - integration-tests/jpa-mariadb/**/*
+ - integration-tests/jpa-mssql/**/*
+ - integration-tests/jpa-mysql/**/*
+ - integration-tests/jpa-postgresql/**/*
+ - integration-tests/jpa-without-entity/**/*
+ - integration-tests/common-jpa-entities/**/*
```
gsmet: As mentioned above, I think all of these should have the Hibernate ORM label.

.github/labeler.yml
```diff
@@ -0,0 +1,273 @@
...
+
+area/qute:
+ - extensions/qute/**/*
```
gsmet: shouldn't you add the independent-project too?

.github/labeler.yml
```diff
@@ -0,0 +1,273 @@
...
+area/reactive:
+ - extensions/reactive-mysql-client/**/*
+ - extensions/reactive-pg-client/**/*
+ - extensions/reactive-streams-operators/**/*
+ - integration-tests/reactive-mysql-client/**/*
+ - integration-tests/reactive-pg-client/**/*
+
+area/rest-client:
+ - extensions/rest-client/**/*
+ - extensions/elasticsearch-rest-client/**/*
```
gsmet: This one shouldn't be here but with the hibernate-search label. At least for now. It's not exposed, just an internal library of the Search extension.

.github/labeler.yml
```diff
@@ -0,0 +1,273 @@
...
+
+area/qute:
+ - extensions/qute/**/*
+
+area/reactive:
+ - extensions/reactive-mysql-client/**/*
+ - extensions/reactive-pg-client/**/*
+ - extensions/reactive-streams-operators/**/*
+ - integration-tests/reactive-mysql-client/**/*
+ - integration-tests/reactive-pg-client/**/*
```
gsmet: Shouldn't also we have there all the `smallrye-reactive-messaging` ones?

.github/labeler.yml
```diff
@@ -0,0 +1,273 @@
...
+ - integration-tests/vault/**/*
+ - integration-tests/vault-agroal/**/*
+ - integration-tests/vault-app/**/*
+
+area/vertx:
+ - extensions/vertx/**/*
+ - extensions/vertx-core/**/*
+ - extensions/vertx-graphql/**/*
+ - extensions/vertx-http/**/*
+ - extensions/vertx-keycloak/**/*
```
gsmet: This one should probably have the `security` label too.

.github/labeler.yml
```diff
@@ -0,0 +1,273 @@
...
+ - extensions/vertx-web/**/*
+ - integration-tests/vertx/**/*
+ - integration-tests/vertx-graphql/**/*
+ - integration-tests/vertx-http/**/*
```
gsmet: Wondering if all these should have the `reactive` one?

.github/labeler.yml
```diff
@@ -0,0 +1,273 @@
...
+### area/dependencies
+### area/dev-mode
+### area/gradle
+### area/housekeeping
+### area/infra
+### area/new-http
+### area/oidc-interoperability
+### area/openapi
+### area/platform
+### area/reactive-messaging
```
gsmet: This one should be used for the `smallrye-reactive-messaging*` extensions.

.github/labeler.yml
```diff
@@ -0,0 +1,273 @@
...
+### area/security-consolidation
+### area/testing
```
gsmet: Maybe we can put this one on all the `test-framework` directory?

gsmet (changes requested):
That's amazing, Max, thanks a lot for that one.

I added some comments inline.

.github/labeler.yml
maxandersen: i'll probably write a new script to generate this stuff automatically ;) 
so can we say any folder with `jpa` in it should get the hibernate label ?

.github/labeler.yml
maxandersen: yeah - after submitting it I realized the whole `devtools` system was missing which is actually my personal interest to get auto-labelled :) I'll get that added.

.github/labeler.yml
maxandersen: where is that actually ? only spotting mongodb-client ....

.github/labeler.yml
maxandersen: yup - I'll regen where "jpa" => hibernate

.github/labeler.yml
```diff
@@ -0,0 +1,273 @@
...
+ - extensions/arc/**/*
+
+area/artemis:
+ - extensions/artemis-core/**/*
+ - extensions/artemis-jms/**/*
+ - integration-tests/artemis-core/**/*
+ - integration-tests/artemis-jms/**/*
+
+area/config:
+ - extensions/config-yaml/**/*
```
dmlloyd: I'd add `core/deployment/src/main/java/io/quarkus/deployment/configuration/**/*` and `core/runtime/src/main/java/io/quarkus/runtime/configuration/**/*` here as well.

maxandersen:
actually - this is easier to update by hand.. update incoming soon(tm) :)

.github/labeler.yml
maxandersen: done.

.github/labeler.yml
maxandersen: done

.github/labeler.yml
maxandersen: done

.github/labeler.yml
maxandersen: ah - had to search a level deeper - extensions/panache/mongodb-*

.github/labeler.yml
maxandersen: in addition to area/reactive-messaging ?

maxandersen:
I've pushed an update and I wrote  a [jbang script](https://gist.github.com/maxandersen/590b8a0e824faeb3ee7ddfad741ce842) to verify this file. 

do the following in root:

```
$ sdk install jbang 
$ jbang https://git.io/JvJR5 .github/labeler.yml <optional path>
```

it has a few options you can see by running: `jbang https://git.io/JvJR5`

gastaldi:
Make sure to add these files to be ignored by the azure pipeline

maxandersen:
> Make sure to add these files to be ignored by the azure pipeline

done.

.github/labeler.yml
```diff
@@ -0,0 +1,273 @@
...
+ - extensions/elytron-security-common/**/*
+ - extensions/elytron-security-jdbc/**/*
+ - extensions/elytron-security-oauth2/**/*
+ - extensions/elytron-security-properties-file/**/*
+ - integration-tests/elytron-security/**/*
+ - integration-tests/elytron-security-jdbc/**/*
+ - integration-tests/elytron-security-oauth2/**/*
+
+area/smallrye:
+ - extensions/smallrye-context-propagation/**/*
```
michalszynkiewicz: To me the specs are different worlds. I think JWT fits better in security, openapi could live together with resteasy, reactive-messaging-* in reactive

gastaldi:
> > Make sure to add these files to be ignored by the azure pipeline
> 
> done.

@maxandersen I think you need to add your new files to https://github.com/quarkusio/quarkus/blob/master/azure-pipelines.yml#L13 to ignore it?

maxandersen:
> @maxandersen I think you need to add your new files to https://github.com/quarkusio/quarkus/blob/master/azure-pipelines.yml#L13 to ignore it?

sorry was missing a push. done now.

.github/labeler.yml
maxandersen: I'm not fully following - can you make a propsed edit here? otherwise we can update it after I push this initial setup.

geoand:
Can you please rebase the PR onto master to hopefully make CI happy?

Thanks

PR #6530, opened 2020-01-13
Optimize @BsonProperty replacement

loicmathieu:
Cache the replacement map to avoid reflection lookup each time

### DIFF
--- a/extensions/panache/mongodb-panache/deployment/src/main/java/io/quarkus/mongodb/panache/deployment/PanacheResourceProcessor.java
+++ b/extensions/panache/mongodb-panache/deployment/src/main/java/io/quarkus/mongodb/panache/deployment/PanacheResourceProcessor.java
@@ -1,9 +1,12 @@
 package io.quarkus.mongodb.panache.deployment;
 
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
 
 import org.bson.codecs.pojo.annotations.BsonProperty;
 import org.bson.types.ObjectId;
@@ -19,6 +22,9 @@ import org.jboss.jandex.Type;
 import io.quarkus.deployment.Capabilities;
 import io.quarkus.deployment.annotations.BuildProducer;
 import io.quarkus.deployment.annotations.BuildStep;
+import io.quarkus.deployment.annotations.ExecutionTime;
+import io.quarkus.deployment.annotations.Record;
+import io.quarkus.deployment.bean.JavaBeanUtil;
 import io.quarkus.deployment.builditem.ApplicationIndexBuildItem;
 import io.quarkus.deployment.builditem.BytecodeTransformerBuildItem;
 import io.quarkus.deployment.builditem.CapabilityBuildItem;
@@ -33,6 +39,7 @@ import io.quarkus.jsonb.spi.JsonbDeserializerBuildItem;
 import io.quarkus.jsonb.spi.JsonbSerializerBuildItem;
 import io.quarkus.mongodb.panache.PanacheMongoEntity;
 import io.quarkus.mongodb.panache.PanacheMongoEntityBase;
+import io.quarkus.mongodb.panache.PanacheMongoRecorder;
 import io.quarkus.mongodb.panache.PanacheMongoRepository;
 import io.quarkus.mongodb.panache.PanacheMongoRepositoryBase;
 import io.quarkus.mongodb.panache.ProjectionFor;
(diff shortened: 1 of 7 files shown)


loicmathieu:
@FroMage this small enhancement avoids reflection lookup for both projection and PanacheQL queries by caching the relfection map.

FroMage:
Can't you compute this at build-time and generate a class returning the cached data instead?

loicmathieu:
@FroMage of course but I will need to change more than 9 lines for this ...

Can you show me an example ? How this class instance can be calculated at build time and accessible at runtime ?

geoand:
@loicmathieu I assume the idea would be to use information from Jandex to build the Map you need.
You would then pass it to a recorder which would stuff it into a `static volatile` field thus making it available

FroMage:
Yeah, something like that indeed.

BTW, why does it need to be `volatile`? I never understood that.

geoand:
The best explanation I have seen is: https://www.ibm.com/developerworks/library/j-jtp06197/index.html (`Pattern #2`)

FroMage:
Well, I have a decent idea what `volatile` does, but why do we use it on `static` fields in Quarkus?

geoand:
I don't know if the semantics change for a `static` field. I think we just use it for the same reason, 
"synchronizing" the value between a piece of code that writes it and a piece that reads it later on

loicmathieu:
@geoand I will try what you describe.

Regarding volatile, I didn't understand neither why it is needed as it will be written once by the build step and read later on so the volatile "lightweigth" synchronization should not be needed ...
Is the augmentation phase multi-threaded ?
As te augmentation will occurs before the first request volatile should not be needed ...

geoand:
I agree, it shouldn't be since the steps are not multithreaded. Perhaps it's a forward looking thing

loicmathieu:
@geoand @FroMage should I keep a failsafe implementation via relfection (the same as today) in case the augmentation phase didn't create the replacement map for this class ?

At least it can be handy for unit testing ...

geoand:
If I were doing it I would probably drop it, but up to you :)

loicmathieu:
OK, if there is no failsafe we would easily spot bugs ;)

loicmathieu:
Hum, doing this will madate to use the `@ProjectionFor` annotation which is not mandatory today (if you don't use `@BsonProperty` inside a projection class you don't need `@ProjectionFor`).

I prefere to keep a failsafe implementation that will log at INFO level that a replacement map didn't exist for this class and to make sure it is inside the Jandex index and that, if using projection, the class is annotated with `@ProjectionFor`.

Or we can make the `@ProjectionFor` annotation mandatory but this will imply checking it at runtime ... via reflection ?

geoand:
OK, so I don't know the details of that stuff so I'll let @FroMage take over the discussion for here :)

loicmathieu:
@FroMage @geoand I pre-compute the replacement map for `@BsonProperty` at build time using Jandex.

I also register for reflection all entity classes as it was not the case before.

I keep a reflection safety net so that if a class has no replacement map we recompute it (can be the case fir projection class that didn't use the `@ProjectionFor` annotation).

Now, reflection is only used to access fields/getters of the projection class at runtime. I don't think we can do it another way ...

FroMage:
@loicmathieu I'm not sure I understand.

`@ProjectionFor` is not mandatory today, unless you have `@BsonProperty` in the entity, right? You said "projection class" instead of "entity".

I'm not sure what your runtime fallback does? It scans a projection class that was not annotated with `@ProjectionFor` in case they have `@BsonProperty`?

loicmathieu:
@FroMage "projection class" means the class that we project to (so the one that will replace the entity).

`@ProjectionFor` is not mandatory because we didn't check it's presence inside the projection implemetantation. If some `@BsonProperty` is used inside the entity class, without the `@ProjectionFor` annotation they will not by "copied" to the projection class.

The fallback use reflection if the information is not inside the "replacement" cache. So if we fail at build time to index the entity class or the projection class.

FroMage:
> @ProjectionFor is not mandatory because we didn't check it's presence inside the projection implemetantation. If some @BsonProperty is used inside the entity class, without the @ProjectionFor annotation they will not by "copied" to the projection class.

Well, you can always store the `@BsonProperty` mappings at build time for the entity classes, because those are always scanned, right?

Then at run-time if you have a projection class that has no projection mapping (it had no `@ProjectionFor`) you can use the entity mappings without having to re-scan them, no?

loicmathieu:
> Well, you can always store the @BsonProperty mappings at build time for the entity classes, because those are always scanned, right?

It is already done

> Then at run-time if you have a projection class that has no projection mapping (it had no @ProjectionFor) you can use the entity mappings without having to re-scan them, no?

If a projection class as no `@BsonProperty` mapping and has no `@ProjectionFor` it will not be enhanced (we add the `@BsonProperty annotation from the entity to the projection class via bytecode enhancement at build time) so it will not works using the MongoDB automatic pojo codec. 

The "replacement map" that I optimized in this PR is to be able to use the entity property instead of the database fields in query. And we decided that a projection class must defines the same fields that the entity class. So both functionality uses the same code for different reasons.

loicmathieu:
> Well, you can always store the @BsonProperty mappings at build time for the entity classes, because those are always scanned, right?

In case of a repository implementation, if the repository type parameter (the entity class) is not in the Jandex index this could works so the entity class can be, in some rare cases, not in the index (typical use case: business object in a separate jar) ... That's why I added a fallback mechanism via relfection with a big WARNING in the log file (index your separate jar or add the @ProjectionFor annotation).

FroMage (approved):

FroMage:
OK sorry, I get it now.

gsmet:
@FroMage should this one be merged?

FroMage:
It can, yes.

PR #6533, opened 2020-01-13
update and align RX Java 2 version to 2.2.17

cescoffier:


gsmet:
Does it make sense to add it to Dependabot and have it handled for us? I mean we are not forced to accept all the changes but it will update everything in a consistent way.

gsmet (approved):

cescoffier:
Agreed, I was surprised actually.

geoand:
+1 for the dependabot change

cescoffier:
Just opened: https://github.com/quarkusio/quarkus/pull/6542/files

PR #6547, opened 2020-01-14
Support configuring ImplicitNamingStrategy in Hibernate ORM

### LINKED ISSUE #6546: Hibernate ORM Implicit Naming Strategy Config
Initial discussion - https://groups.google.com/forum/#!topic/quarkus-dev/F2Xu_1dKick

> I'm trying to apply a custom implicit naming strategy using hibernate.implicit_naming_strategy configuration option. 
> 
> There is no possibility to put it to application.property file yet. So I have to put it to persistence.xml. But in my case some of the hibernate options must be in application.properties as I want them to be configured via the environment variables, for example, quarkus.datasource.url.
> 
> It isn't possible to use application.properties and persistence.xml together at the same time to configure hibernate.
> 
> After some experiments, I've found that I can use hibernate.properties instead of persistence.xml to configure naming strategy. 
> 
> Running app in dev mode works as expected. I see in the log:
> 
> 020-01-14 10:39:18,119 INFO [org.hib.cfg.Environment] (build-34) HHH000205: Loaded properties from resource hibernate.properties: {hibernate.bytecode.use_reflection_optimizer=false, hibernate.implicit_naming_strategy=org.example.CustomNamingStrategy}
> 
> and further logs shows that the naming strategy is applied and everything works fine.
> 
> But after deploying to Kubernetes in prod mode it works differently. There is the same log message that configuration is read, but custom naming strategy isn't applied.
> 
> As well test-mode works same as prod - naming strategy isn't applied.

We need to support implicit naming strategy in application.properties

gastaldi:
Fixes #6546

### DIFF
--- a/extensions/hibernate-orm/deployment/src/main/java/io/quarkus/hibernate/orm/deployment/HibernateOrmConfig.java
+++ b/extensions/hibernate-orm/deployment/src/main/java/io/quarkus/hibernate/orm/deployment/HibernateOrmConfig.java
@@ -94,6 +94,14 @@ public class HibernateOrmConfig {
     Optional<String> physicalNamingStrategy;
 
     /**
+     * Pluggable strategy for applying implicit naming rules when an explicit name is not given.
+     *
+     * Class name of the Hibernate ImplicitNamingStrategy implementation
+     */
+    @ConfigItem
+    Optional<String> implicitNamingStrategy;
+
+    /**
      * Query related configuration.
      */
     @ConfigItem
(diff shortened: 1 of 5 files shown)


gsmet:
Let's not go too fast on that one, we need to discuss it with Sanne. And, at least, we need to properly understand the use case.

Sanne:
yes let's make sure there is a strong reason to have this before merging please.

gastaldi:
@Sanne the reason why this is needed is explained by @dima-fedotovs in https://github.com/quarkusio/quarkus/issues/6546#issuecomment-574137538

Sanne (approved):

PR #6548, opened 2020-01-14
Improving the mailer adoc by adding quarkus.mailer.mock=false in the initial set of properties.

gochev:
The sendgrid example in the **current state** will not work for anyone who tried, because the quarkus.mailer.mock=false is not present. If you just try to send the email it will fail and you will have to read the whole testing section to find out that the quarkus.mailer.mock is true by default.

I find this misleading I personally have lost around 40 mins finding and debugging why the email is not send and finding that I am in a mock mode.

I greatly believe that this will help the others using mailer to send email with or without sendgrid.

Maybe it is wise to even add a comment somewhere below the properties similar on how the password vault is explained.

### DIFF
--- a/docs/src/main/asciidoc/mailer.adoc
+++ b/docs/src/main/asciidoc/mailer.adoc
@@ -60,6 +60,7 @@ quarkus.mailer.port=465
 quarkus.mailer.ssl=true
 quarkus.mailer.username=....
 quarkus.mailer.password=....
+quarkus.mailer.mock=false
 ----
 
 [NOTE]


cescoffier (approved):

machi1990:
> The sendgrid example in the **current state** will not work for anyone who tried, because the quarkus.mailer.mock=false is not present. If you just try to send the email it will fail and you will have to read the whole testing section to find out that the quarkus.mailer.mock is true by default.
> 
> I find this misleading I personally have lost around 40 mins finding and debugging why the email is not send and finding that I am in a mock mode.

@gochev Was this in dev mode? IIRC, the mock mode is enabled by default in dev and test modes. 

@cescoffier maybe we should enable mocking only in test mode?

cescoffier:
Mock can be used outside of test mode, like in dev mode you don't necessarily want to send emails.
However, there are case where you want to do so.

gochev:
yes it was in dev mode since I was developing and spend a lot of time finding  out well why this doesn't work, is it going to the smtp server at all and so on. On top when using things like sendgrid, google smtp it is even harder to find out whats going on. This flags when set to TRUE is a bit strange even for a dev mode, my assumption would be if it is missing then it is false 🙄 in any case I believe a note is nice to be added since people run quarkus in dev mode mostly.

Maybe it is wise to be put as:
quarkus.mailer.mock=false  # enabled by default on quarkus:dev, read more about it in the test section

or something similar

PR #6557, opened 2020-01-14
Support Hibernate ORM's map entity mode

gsmet:
@gunnarmorling @Naros is it sufficient for you?


Naros:
Hi @gsmet that's exactly what we need.  We actually worked around this today by simply specifying those two classes for reflective access in our extension processor.  I'll add this PR in my follow-up task as to remove those from our processor when this gets merged & released.

gastaldi (approved):

PR #6560, opened 2020-01-15
Some tiny English fixes from reading guides

hutchig:
Just a few small spelling/English corrections for the website/guides.
Signed-off-by: Gordon Hutchison <Gordon.Hutchison@gmail.com>

### DIFF
--- a/docs/src/main/asciidoc/deploying-to-kubernetes.adoc
+++ b/docs/src/main/asciidoc/deploying-to-kubernetes.adoc
@@ -93,5 +93,5 @@ Your application is accessible at the printed URL.
 
 This guide covered the deployment of a Quarkus application on Kubernetes and OpenShift.
 However, there is much more, and the integration with these environments has been tailored to make Quarkus applications execution very smooth.
-For instance, the health extension can be used for health check; the configuration support allows mounting the application configuration using config map, the metric extension produces data _scrappable_ by Prometheus and so on.
+For instance, the health extension can be used for health check; the configuration support allows mounting the application configuration using config map, the metric extension produces data _scrapable_ by Prometheus and so on.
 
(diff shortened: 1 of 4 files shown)


docs/src/main/asciidoc/deploying-to-kubernetes.adoc
```diff
@@ -93,5 +93,5 @@ Your application is accessible at the printed URL.
 
 This guide covered the deployment of a Quarkus application on Kubernetes and OpenShift.
 However, there is much more, and the integration with these environments has been tailored to make Quarkus applications execution very smooth.
-For instance, the health extension can be used for health check; the configuration support allows mounting the application configuration using config map, the metric extension produces data _scrappable_ by Prometheus and so on.
+For instance, the health extension can be used for health check; the configuration support allows mounting the application configuration using config map, the metric extension produces data _scrapable_ by Prometheus and so on.
```
geoand: Are you sure about this one? When I search google I see that both are valid, or am I mistaken?

geoand (commented):
Thanks for this!

I added a comment

docs/src/main/asciidoc/deploying-to-kubernetes.adoc
hutchig: Yes :-) 
https://en.wiktionary.org/wiki/scrappable
https://en.wiktionary.org/wiki/scrapable

https://www.thefreedictionary.com/scrapable  (and find "scrapable adj" )

I can also derive it..
There are two distinct meanings - for scrapage and scrappage - one from 'scrape' (as in scraping ice off my car window) and one from 'scrap' (as in scrapping my old rusty car)  It is the long 'a' sound in scrape made with the 'split digraph' that is carried into scrapable and this only works with the single consonant 'p'. Indeed this is sometimes written with a 'magic e' as 'scrapeable' but scrapable is an acceptable alternative.
https://www.yourdictionary.com/scrapeable

geoand: Got it, thanks!

geoand (approved):
Thanks a lot!

Issue #6392, opened 2020-01-03
Infinispan Embedded: config file not included in native image


gsmet:
This one was reported by a user so we apparently have people using this extension.

Currently the configuration file is configured at runtime (see https://github.com/quarkusio/quarkus/blob/master/extensions/infinispan-embedded/runtime/src/main/java/io/quarkus/infinispan/embedded/runtime/InfinispanEmbeddedRuntimeConfig.java#L19) and not included in the native image which makes the native image fail.

If we keep it configurable at runtime, there's no way to include it automatically in the native image so we need to choose one evil.

/cc @wburns @Sanne

Sanne:
I believe the configuration aspect is very much work in progress: we shouldn't parse the whole thing at runtime of course :)

But I'll leave it to @wburns to decide if we should implement a stop-gap solution by including the resource, or if he'd rather directly go for the better long term solution.

thanks @gsmet.

wburns:
Sorry, I didn't respond earlier.

The goal originally was to have this be a runtime config option for now. An example is we will want to generate a native image that runs the Infinispan server allowing a user to provide an arbitrary Infinispan config file without rebuilding.

Maybe we will need to deviate from this for the embedded extension though.

wburns:
@jmartisk Can you describe what you are trying to do with the Infinispan embedded extension here? That way we can discuss it regarding runtime or build time config.

gsmet:
@wburns how do you expect it to work if the configuration file is not included in the native image?

Because that's the issue we have here. You can override it but as the file is not there in the native image, it doesn't work.

wburns:
The file would be on disk. But the more I think about it I am thinking this functionality should probably be moved up higher to our server extension (that doesn't live in Quarkus).

I personally am open to however people may use the embedded extension at this point. It is really barebones and ready to be molded as we need. I just don't know if we should ignore the xml and only go the programmatic configuration way or not.

gsmet:
Yeah, so the problem is that we have users expecting to be able to use files in the classpath. We had one user complaining about it.

geoand:
Is this still an issue?

geoand:
Closing for lack of feedback

PR #6564, opened 2020-01-15
Watch infinispan-embedded.xml-config file for changes in devmode

### LINKED ISSUE #6563: Dev mode: reload on infinispan-embedded.xml-config file changes
In dev mode, we should trigger a reload when there is a change in the `quarkus.infinispan-embedded.xml-config` file

jmartisk:
Fixes https://github.com/quarkusio/quarkus/issues/6563

### DIFF
--- a/extensions/infinispan-embedded/runtime/src/main/java/io/quarkus/infinispan/embedded/runtime/InfinispanEmbeddedRuntimeConfig.java
+++ b/extensions/infinispan-embedded/runtime/src/main/java/io/quarkus/infinispan/embedded/runtime/InfinispanEmbeddedRuntimeConfig.java
@@ -9,7 +9,7 @@ import io.quarkus.runtime.annotations.ConfigRoot;
 /**
  * @author wburns
  */
-@ConfigRoot(name = "infinispan-embedded", phase = ConfigPhase.RUN_TIME)
+@ConfigRoot(name = "infinispan-embedded", phase = ConfigPhase.BUILD_AND_RUN_TIME_FIXED)
 public class InfinispanEmbeddedRuntimeConfig {
 
     /**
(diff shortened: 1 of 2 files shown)


extensions/infinispan-embedded/runtime/src/main/java/io/quarkus/infinispan/embedded/runtime/InfinispanEmbeddedRuntimeConfig.java
```diff
@@ -9,7 +9,7 @@
 /**
  * @author wburns
  */
-@ConfigRoot(name = "infinispan-embedded", phase = ConfigPhase.RUN_TIME)
+@ConfigRoot(name = "infinispan-embedded", phase = ConfigPhase.BUILD_AND_RUN_TIME_FIXED)
```
machi1990: Do we really want this change? As configurations knobs won’t be overridable at runtime.

jmartisk: There's just this one property in that config and IMO it doesn't make sense to override that one at runtime. So logically I think this should be a build-time-only config, but it had to be made available at run-time because of how Infinispan works - it is being read by Infinispan at runtime.

geoand: At the very least we would need input from @wburns before making this change

gsmet:
I opened an issue about the existing behavior being incorrect here: https://github.com/quarkusio/quarkus/issues/6392 .

I had no feedback from @wburns though.

gsmet:
OK, well, it's so broken now that we should try to fix it. Could you also add the file as a native resource? Otherwise it's not included in the native image.

Thanks!

extensions/infinispan-embedded/runtime/src/main/java/io/quarkus/infinispan/embedded/runtime/InfinispanEmbeddedRuntimeConfig.java
wburns: 👍 for separating this concern to the other issue. I have responded to why it is RUN_TIME for now at https://github.com/quarkusio/quarkus/issues/6392#issuecomment-575227225

jmartisk:
Sorry now I'm a bit confused what is expected we will be doing with it. So I will add the file as a native resource, but should we split out the config phase change into #6392 or a different issue? (that is a prerequisite for getting the hot reload working, so it would have to be done first)

gsmet:
I'm closing that one. AFAIK the Infinispan Embedded extension is not really supported so let's not waste too much time on it if we don't have user feedback.

PR #6569, opened 2020-01-15
Disable automatic association management in Hibernate ORM entities

Sanne:
During the Hibernate team meeting we reviewed the usage of bytecode enhancement in Quarkus and decided that the "automatic assosciation management" is not quite a good idea to have enabled here.

I suspect nobody actually noticed that it was on, but it would be a good idea to mention this in release notes so for people to be aware of this when upgrading.

### DIFF
--- a/extensions/hibernate-orm/deployment/src/main/java/io/quarkus/hibernate/orm/deployment/HibernateEntityEnhancer.java
+++ b/extensions/hibernate-orm/deployment/src/main/java/io/quarkus/hibernate/orm/deployment/HibernateEntityEnhancer.java
@@ -4,6 +4,7 @@ import java.util.function.BiFunction;
 
 import org.hibernate.bytecode.enhance.spi.DefaultEnhancementContext;
 import org.hibernate.bytecode.enhance.spi.Enhancer;
+import org.hibernate.bytecode.enhance.spi.UnloadedField;
 import org.hibernate.bytecode.spi.BytecodeProvider;
 import org.objectweb.asm.ClassReader;
 import org.objectweb.asm.ClassVisitor;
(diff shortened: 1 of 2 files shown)


geoand:
The Hibernate ORM with Panache integration test seems to be failing in CI

Sanne:
thanks, pushed a fix for that.

Sanne:
@FroMage FYI - especially 17e75d2 . Sorry, that code looked better without my commit :)

geoand:
Cool, thanks

gsmet:
So why is it not a good idea? The description should at least explain that for posterity.

And tbh, that's one of the nice thing of using the enhancement as it's very painful to do it manually.

(and, yes, I missed the discussion but I asked to be there IIRC and nobody pinged me for this breakout so I'm asking now :))

Sanne:
I have a comment reminding why it's not a good idea in the code, but right let me expand a bit on it.

The main problem is that this automatic handling might be unexpected, and apply unintended changes to the data.

For example if you conbine this with collections which allow repeated element, ordered or not (such as bags), you might expect to need to handle "the other side" explicitly by adding an entry to the collection, but since it's automatically added as well you end up with duplicates.

Removal can get even worse; it's less likely but if you intended to have duplicates, you might want to be picky about which element is being removed - while the automatic association management doesn't allow you to make an explicit choice, it will just remove one: this choice is unspecified, and perhaps not even expected.

I think it's clear that this feature should not be on by default, mainly because users will likely not expect these things to happen.

If in the future we can figure out how to implement this exclusively for safe use cases, such as on collections on which the operation would be idempotent, we could reconsider enabling it (as it wouldn't be a compatibility concern either).
We'd still have the problem of such magic applying only on some collections and not others, so the consistency should also be considered.

gsmet (approved):

FroMage:
> @FroMage FYI - especially 17e75d2 . Sorry, that code looked better without my commit :)

ACK

PR #6570, opened 2020-01-15
Qute "if" section - support multiple conditions

mkouba:
This PR makes it possible to use more complex expressions in `{#if}` sections:
`{#if !item.active || (item.sold && item.price == 100)}`
`{#if item.age > 10 && item.price > 500}`
`{#if (item.age > 10 || item.price > 500) && user.loggedIn}`

We can switch to a common reusable EL (once we have it ;-) but this should be powerful enough in the meantime.

TODO: improve validation of parameters.


gsmet:
I won't have the time to review that one before 1.2. @gastaldi could you give it a shot?

gastaldi:
Sure, I'll give it a try tomorrow morning first thing

gastaldi:
@mkouba doesn't seem to work with me. I have a template like: 
```
{#if (name.equals("George")) || name.startsWith("C")}
Hello {name}!
{#else}
OI {name}!
{/if}
```
Having a GreetingResource like: 

```java
@Path("/hello")

    @Inject
    Template hello; 

    @GET
    public TemplateInstance hello(@QueryParam("name") String name) {
        return hello.data("name",name);
    }
}
```
And I did a http://localhost:8080/hello?name=George but I get the `OI George!`  output

gastaldi:
Ah my fault for not RTFM:
```
{#if name is "George" || name startsWith "C"}
Hello {name}!
{#else}
OI {name}!
{/if}
```

It seems that startsWith is not supported yet

mkouba:
@gastaldi `equals` and `startsWith` methods are not automatically available. By default, only public properties work out of the box (public getter or field). You can use `@TemplateData(target=String.class)` or `@TemplateExtension`.

gastaldi (approved):
Works like a charm. Good job!

PR #6579, opened 2020-01-16
Add temporary workaround for OpenTracing issue with https in Maven Central

geoand:
…ntral

Maven Central no longer support http:
https://support.sonatype.com/hc/en-us/articles/360041287334
And the OpenTracing TCK test resolves via Central using http so we need to
workaround that for now

### DIFF
--- a/tcks/microprofile-opentracing/base/pom.xml
+++ b/tcks/microprofile-opentracing/base/pom.xml
@@ -34,6 +34,15 @@
                     </dependenciesToScan>
                 </configuration>
             </plugin>
+
+            <!-- Temporarily disable enforcer because of comment about test dependencies (see later) -->
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-enforcer-plugin</artifactId>
+                <configuration>
+                    <skip>true</skip>
+                </configuration>
+            </plugin>
         </plugins>
     </build>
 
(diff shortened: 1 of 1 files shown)


gsmet:
Shouldn't we try to use a more recent Shrinkwrap version instead? It has been fixed in 2017: https://github.com/shrinkwrap/resolver/commit/c529fa2e99e70556171e9911f33c60a7d800c369#diff-728eaf5d6e16290570bc8ac51c1b5377 .

Not saying we shouldn't merge the quick workaround but if forcing a more recent Shrinkwrap version fixes the issue, that would have my preference.

There's a good chance the API has changed though so we would need someone from MP OpenTracing to have a look at that.

@kenfinnigan can you coordinate that?

geoand:
@gsmet Yes for sure, I wanted to get this in quick now because our CI is pilling up PRs that can't be merged and I don't have much experience with Shrinkwrap or the TCKs to dive into any issues that might occur.

So I propose that we merge this PR now (if CI passes of course) and then @kenfinnigan and @pavolloffay can provide a proper fix when they are ready

gsmet:
@geoand I wonder if calling:
```
Maven.configureResolver().withMavenCentralRepo(false)
                .withRemoteRepo("Maven Central", "https://repo.maven.apache.org/maven2/", "default");
```
would be a better way to fix it?

That would make sure it uses HTTPS for everything and would avoid future breakage. WDYT?

gsmet (approved):

geoand:
That would be a good way to fix it yes, however the test is not inside our repo

geoand:
I'm gonna close this since the approach won't work after all

PR #6582, opened 2020-01-16
Kubernetes Extension: Include openshift, knative and more in the list of supported prefixes

iocanel:
Currently we filter out all propertiest that don't have the kubernetes prefix.
This means that we ignore everything but kubernetes related properties.

This pull request addresses that and also adds:

- openshift
- knative 
- docker 
- s2i 

to the list of supported prefixes.


extensions/kubernetes/deployment/src/main/java/io/quarkus/kubernetes/deployment/KubernetesProcessor.java
```diff
@@ -51,7 +51,8 @@
 class KubernetesProcessor {
 
     private static final String PROPERTY_PREFIX = "dekorate.";
-    private static final String ALLOWED_GENERATOR = "kubernetes";
+    private static final List<String> ALLOWED_GENERATORS = Arrays.asList("kubernetes", "openshift", "knative", "docker", "s2i");
```
geoand: How about making this a `Set` so `contains` will be O(1)?

iocanel: Done

extensions/kubernetes/deployment/src/main/java/io/quarkus/kubernetes/deployment/KubernetesProcessor.java
```diff
@@ -51,7 +51,22 @@
 class KubernetesProcessor {
 
     private static final String PROPERTY_PREFIX = "dekorate.";
-    private static final String ALLOWED_GENERATOR = "kubernetes";
+    private static final Set<String> ALLOWED_GENERATORS = new HashSet<String>() {
```
geoand: Just nitpicking here, but the codebase generally shies away from this syntax and uses the more verbose `new HashSet(Arrays.asList())`

extensions/kubernetes/deployment/src/main/java/io/quarkus/kubernetes/deployment/KubernetesProcessor.java
```diff
@@ -97,18 +99,25 @@ public void build(ApplicationInfoBuildItem applicationInfo,
                 .collect(Collectors.toList());
 
         Map<String, Object> configAsMap = StreamSupport.stream(config.getPropertyNames().spliterator(), false)
-                .filter(k -> ALLOWED_GENERATOR.equals(generatorName(k)))
+                .filter(k -> ALLOWED_GENERATORS.contains(generatorName(k)))
                 .collect(Collectors.toMap(k -> PROPERTY_PREFIX + k, k -> config.getValue(k, String.class)));
+
         // this is a hack to get kubernetes.registry working because currently it's not supported as is in Dekorate
```
geoand: Does this comment still apply?

extensions/kubernetes/deployment/src/main/java/io/quarkus/kubernetes/deployment/KubernetesProcessor.java
```diff
@@ -97,18 +99,25 @@ public void build(ApplicationInfoBuildItem applicationInfo,
...
-        }
+        Optional<String> dockerRegistry = IMAGE_GENERATORS.stream()
+                .map(g -> config.getOptionalValue(g + ".registry", String.class))
+                .filter(Optional::isPresent)
+                .map(Optional::get)
+                .findFirst();
+
+        dockerRegistry.ifPresent(v -> System.setProperty(DOCKER_REGISTRY_PROPERTY, v));
+
         // this is a hack to work around Dekorate using the default group for some of the properties
```
geoand: Same for this

gsmet (dismissed):
I'm a bit torn on that one because that means that all those prefixes are now taken by this extension.

I wonder if they shouldn't all have the `kubernetes.` prefix so that it's clear they all belong to this extension.

We might want to use these prefixes for other extensions at some point.

@geoand WDYT?

extensions/kubernetes/deployment/src/main/java/io/quarkus/kubernetes/deployment/KubernetesProcessor.java
iocanel: I don't remember why it was initially added.

iocanel:
@gsmet: As this extension is used for populating manifests for `openshift` and `knative` I don't see an issue with that. Same goes for `s2i` that goes hand in hand with `openshift`.

I only see that being a potential issue with `docker` for obvious reasons. Even if at some point we decide to have a docker extension, why would that be a problem to share properties like `docker-file` or `registry`?

extensions/kubernetes/deployment/src/main/java/io/quarkus/kubernetes/deployment/KubernetesProcessor.java
geoand: I had added it because setting `kubernetes.registry` was not having an effect. I assume this has been fixed now? If so, the comment is no longer needed

geoand:
I am with @iocanel on this one, I think it's fine to have the extension handle these prefixes

gsmet:
OK, you clear that it means we won't be able to use these prefixes for any other extensions?

Just checking.

geoand:
Yes exactly. I could be wrong, but I don't see other extensions needing to use these prefixes. 
If that turns out not to be the case, you can always say "I told you so" :)

gsmet:
@geoand I dismissed my review.

geoand:
@gsmet OK, thank you

extensions/kubernetes/deployment/src/main/java/io/quarkus/kubernetes/deployment/KubernetesProcessor.java
geoand: OK, let's leave it for now

geoand (approved):

PR #6595, opened 2020-01-16
RESTeasy standalone does not respect max request size

### LINKED ISSUE #6590: Resteasy Vertx. does not respect max body size
**Describe the bug**
Configuration setting "quarkus.http.limits.max-body-size" gets ignored when using resteasy vertx.
This setting should limit the max size of a request body.
However, the request does not get denied.

Using undertow, this works as expected.

**Expected behavior**
"413 Request Entity To Large" as response status code.

**Actual behavior**
"200 OK" as response. This means, that the entity was accepted, and my application is working with the to big entity.

**To Reproduce**
See reproducer project [code-with-quarkus.zip](https://github.com/quarkusio/quarkus/files/4072372/code-with-quarkus.zip)

1. Run reproducer project in dev mode or as runner jar - does not matter
2. The project as a max body size of 10 byte set
3. Execute following curl:

> curl -i -X POST -H "Content-Type: application/json" -d ", says the brown fox" http://localhost:8080/hello

4. This returns a 200 status code, although the body is bigger than 10 byte.

5. Change the project to also use undertow
6. Run the same curl as before
7. -> Now a 413 is returned.

**Configuration**
```properties 
quarkus.http.limits.max-body-size=10

```

MSYS_NT-10.0 SOMEHOSTNAME 2.10.0(0.325/5/3) 2018-06-13 23:34 x86_64 Msys
 - Output of `java -version`:  
java version "1.8.0_241"
Java(TM) SE Runtime Environment (build 1.8.0_241-b07)
Java HotSpot(TM) 64-Bit Server VM (build 25.241-b07, mixed mode)
 - Quarkus version or git rev: 1.1.1.Final

stuartwdouglas:
Fixes #6590

### DIFF
--- a/extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/VertxHttpRecorder.java
+++ b/extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/VertxHttpRecorder.java
@@ -24,6 +24,8 @@ import org.wildfly.common.cpu.ProcessorInfo;
 
 import io.netty.bootstrap.ServerBootstrap;
 import io.netty.channel.ChannelInitializer;
+import io.netty.handler.codec.http.HttpHeaderNames;
+import io.netty.handler.codec.http.HttpResponseStatus;
 import io.quarkus.arc.Arc;
 import io.quarkus.arc.runtime.BeanContainer;
 import io.quarkus.netty.runtime.virtual.VirtualAddress;
@@ -66,6 +68,8 @@ import io.vertx.ext.web.handler.BodyHandler;
 
 @Recorder
 public class VertxHttpRecorder {
+
+    public static final String MAX_REQUEST_SIZE_KEY = "io.quarkus.max-request-size";
 
     private static final Logger LOGGER = Logger.getLogger(VertxHttpRecorder.class.getName());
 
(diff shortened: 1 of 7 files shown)


extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/VertxHttpRecorder.java
```diff
@@ -215,6 +219,36 @@ public void handle(RoutingContext routingContext) {
             });
         }
 
+        if (httpConfiguration.limits.maxBodySize.isPresent()) {
+            long limit = httpConfiguration.limits.maxBodySize.get().asLongValue();
+            Long limitObj = limit;
+            router.route().order(-2).handler(new Handler<RoutingContext>() {
+                @Override
+                public void handle(RoutingContext event) {
```
cescoffier: Any reason to not use https://vertx.io/docs/apidocs/io/vertx/ext/web/handler/BodyHandler.html#setBodyLimit-long-

cescoffier (approved):
LGTM, just wondering if we could use the Vert.x support for this. It may change a bit the protocol that has been defined.

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/VertxHttpRecorder.java
stuartwdouglas: We do, but not everything uses the body handler, as it requires the full request to be buffered.

cescoffier: oh I see! Thanks!

PR #6599, opened 2020-01-17
Make use of QuteConfig.suffixes when validating template injections

mkouba:
- also change the default value to support VS Code extension


maxandersen (approved):
looks good. `.qute.*` for the win :)

PR #6611, opened 2020-01-17
QuarkusUnitTest - detect usage of non-static fields and fail correctly

mkouba:
- if a non-static field is used the test app is not started
- also update docs

### DIFF
--- a/test-framework/junit5-internal/src/main/java/io/quarkus/test/QuarkusUnitTest.java
+++ b/test-framework/junit5-internal/src/main/java/io/quarkus/test/QuarkusUnitTest.java
@@ -8,18 +8,22 @@ import java.io.ByteArrayOutputStream;
 import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
+import java.lang.reflect.Field;
 import java.lang.reflect.InvocationHandler;
 import java.lang.reflect.Method;
+import java.lang.reflect.Modifier;
 import java.nio.file.FileVisitResult;
 import java.nio.file.FileVisitor;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.attribute.BasicFileAttributes;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Comparator;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Optional;
 import java.util.Properties;
 import java.util.Timer;
 import java.util.TimerTask;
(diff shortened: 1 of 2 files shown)


manovotn (approved):

test-framework/junit5-internal/src/main/java/io/quarkus/test/QuarkusUnitTest.java
```diff
@@ -381,6 +386,20 @@ public void afterEach(ExtensionContext context) throws Exception {
 
     @Override
     public void beforeEach(ExtensionContext context) throws Exception {
```
geoand: Does this need to be done for each test method? Wouldn't it make sense to do it in `beforeAll`?

mkouba: `beforeAll` is not executed at all when a non-static field is used... that's the whole point of the issue ;-)

geoand: OK, thanks

geoand: Although I would have hoped there would be a more useful lifecycle method. Perhaps where the test class is constructed?

mkouba: `TestInstanceFactory.createTestInstance()` is not executed either.

geoand: I rest my case then :)

manovotn: I like how @geoand asks the exact same questions I did xD
BTW this bit documents what is/isn't invoked if the registration is non-static https://junit.org/junit5/docs/current/user-guide/#extensions-registration-programmatic-instance-fields

geoand: Great reference, thanks @manovotn !

geoand:
It seems like some tests in `Arc Deployment` are failing

mkouba:
All the tests that expect a deployment failure seem to fail. Which is weird because for such tests I would expect `beforeEach()` not to be called (the test methods are not invoked). @manovotn any idea?

manovotn:
FTR after a short Zulip exchange we found it was a missing check for tests that assert on deployment failure.

PR #6613, opened 2020-01-17
MongoDB with Panache : avoid to project on the class field

loicmathieu:
Class is always added to the projection fields as all Java classes have a getClass method.
We should avoid this, it is not realy a bug as MongoDB will ignore unknow fields but it's useless.


machi1990 (approved):
Makes sense!

machi1990 (changes requested):
Can you fix the formatting issue?

loicmathieu:
@machi1990 sorry, fixed.

machi1990 (approved):

machi1990:
Thanks! Merged!

PR #6617, opened 2020-01-18
Disable native tests execution for elytron-security-ldap

### LINKED ISSUE #6606: quarkus-elytron-security-ldap-integration-test fails in the native mode
**Describe the bug**
Georgios @geoand has spotted it is not part of [the native CI build](https://github.com/quarkusio/quarkus/blob/master/ci-templates/stages.yml#L262) and fails locally when run in the native mode with:
```
[quarkus-elytron-security-ldap-integration-test-999-SNAPSHOT-runner:26719]     analysis:  27,176.27 ms
Error: Unsupported features in 2 methods
Detailed message:
Error: Class initialization of org.wildfly.security.auth.client.ElytronMessages failed. Use the option --initialize-at-run-time=org.wildfly.security.auth.client.ElytronMessages to explicitly request delayed initialization of this class.
Original exception that caused the problem: java.lang.NoClassDefFoundError: org/wildfly/client/config/ConfigXMLParseException
        at java.lang.Class.forName0(Native Method)
```

gastaldi:
This PR enables native compilation but fails native execution failing with:

`Fatal error: Failed to leave the current IsolateThread context and to detach the current thread. (code 12)` 

Fixes #6606

### DIFF
--- a/ci-templates/stages.yml
+++ b/ci-templates/stages.yml
@@ -259,6 +259,7 @@ stages:
             - elytron-security-oauth2
             - elytron-security
             - elytron-security-jdbc
+            - elytron-security-ldap
             - elytron-undertow
           name: security_1
           keycloak: true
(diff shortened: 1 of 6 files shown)


geoand:
This effort is great, but given the problems that still exist, I don't really know what we should do with this module.


@sberyozkin I think it should be up to you

gastaldi:
Yeah, definitely it doesn't seem to work in native. The application is killed with that error (may be a bug in GraalVM?)

sberyozkin:
Hi George, Georgios; George, thanks for looking into it, I'm sorry, the code of the extension looked sound to me, so from a non-LDAP expert's point of view it was good to go and it works in the JVM mode. IMHO we should ask the author of the PR to spend some time on this issue too, I'll CC in the issue. Perhaps we should mark this extension as experimental ?

sberyozkin:
FYI `quarkus-smallrye-jwt` did not work for a while in the native mode too :-)

sberyozkin (approved):

PR #6629, opened 2020-01-20
doc: fix compilation error

kdnakt:
When I tried the guide `USING JWT RBAC` (https://quarkus.io/guides/security-jwt), I found a compilation error.

I fixed it according to the quickstart project.
https://github.com/quarkusio/quarkus-quickstarts/blob/master/security-jwt-quickstart/src/main/java/org/acme/jwt/TokenSecuredResource.java#L46


geoand (approved):
Thanks for spotting this!

PR #6637, opened 2020-01-20
Fix labeler

geoand:


maxandersen:
weird - didn't need that added when I tried it initially....well lets see ;)

PR #6643, opened 2020-01-20
Remove synthetic flag for created acessor methods in Panache

### LINKED ISSUE #6641: Panache entites are not marked as dirty after deserialization with Jackson
**Describe the bug**
Using Jackson for JSON-Deserializiation Panache entities (w/out explicit setters) is problematic because generated getters & setters are marked "synthetic" by Quarkus and Jackson skips synthetic methods, i.e. it uses just the public fields. But that means that Hibernate's dirty tracking is bypassed and the (changed) entities are not persisted.

**Expected behavior**
Panache entity is persistable (aka. dirty) after deserialization with Jackson.

**Actual behavior**
Panache entity is not marked as dirty after deserialization with Jackson.

**To Reproduce**
1. Use Jackson for JSON handling
2. Deserialize a Panache entity w/out explicit getters & setters
3. Try to persist this entity

**Solution**
As discussed with @FroMage on the chat, the synthetic flag should be removed to make this work as it is not relevant for any (known) use case right now (see getter & setter generation in PanacheEntityEnhancer).

gastaldi:
Fixes #6641

### DIFF
--- a/extensions/panache/panache-common/deployment/src/main/java/io/quarkus/panache/common/deployment/PanacheEntityEnhancer.java
+++ b/extensions/panache/panache-common/deployment/src/main/java/io/quarkus/panache/common/deployment/PanacheEntityEnhancer.java
@@ -185,7 +185,7 @@ public abstract class PanacheEntityEnhancer<MetamodelType extends MetamodelInfo<
                 String getterName = field.getGetterName();
                 String getterDescriptor = "()" + field.descriptor;
                 if (!methods.contains(getterName + "/" + getterDescriptor)) {
-                    MethodVisitor mv = super.visitMethod(Opcodes.ACC_PUBLIC | Opcodes.ACC_SYNTHETIC,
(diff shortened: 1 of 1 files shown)


geoand:
I think an integration test failing without this PR would also be nice :)

FroMage (changes requested):
I think only accessors should have the synthetic bit removed, not the other methods.

FroMage:
Well, I think you deleted the change for mongodb-panache accessors too, and even though they don't have a hibernate bug to workaround, so their accessors are not as significant, I support they should drop the synthetic bit as well just to be consistent. WDYT @loicmathieu ?

loicmathieu:
@FroMage yes, better keep consistent both Panache implementations

gastaldi:
Ah crap, I reverted too much. Let me add it back

gastaldi:
Hm, actually I reverted the RepositoryEnhancer changes, I don't think we need those, right?

FroMage:
Oh, my bad, this is used by both variants.

FroMage (approved):

PR #6644, opened 2020-01-20
Config for boring cyborg auto labeling

maxandersen:
Why:

 * action/labeler does not work for pull-request when the PR is on a fork
 * attempted using paulfnatom/periodic-labler but it fails with other
   permission issues https://github.com/paulfantom/periodic-labeler/issues/6

 * enable use of https://github.com/kaxil/boring-cyborg which supports same label
   syntax. It also has more features but just using the labeling feature for now.


azure-pipelines.yml
```diff
@@ -20,6 +20,7 @@ pr:
       - dco.txt
       - .github/ISSUE_TEMPLATE/*.md
       - .github/labeler.yml
```
gastaldi: Is this path still necessary to be listed here? Same with `github/workflows/*`?

maxandersen: they don't hurt and once github actions Prs comes in they wont use precious CI :)

PR #6655, opened 2020-01-20
Add various test to the SmallRye Reactive Messaging extensions

cescoffier:
* Verifies all valid signatures
* Added hot reload tests for config and code, as well as recovery after an error
* Verifies that profiles are supported


cescoffier:
@paulrobinson FYI.

mkouba (approved):
Looks good.

PR #6630, opened 2020-01-20
Don't fire the startup event until the vert.x producer has been initialized

stuartwdouglas:
Fixes #6612

PR #6657, opened 2020-01-20
Don't fire the startup event until the vert.x producer has been initialized

cescoffier:
Add tests to the fix proposed in https://github.com/quarkusio/quarkus/pull/6630.


gsmet:
I just force-pushed an import order fix.

gastaldi (approved):

Issue #6669, opened 2020-01-21
Add support for Vertx' preferNativeTransport


pcasaes:
**Description**
We can get significant performance boosts by enabling native transport on Vert.x
https://vertx.io/docs/vertx-core/java/#_native_transports

**Implementation ideas**
We could add a new field to `prefer-native-transport` to `VertxConfiguration`.

On `HttpConfiguration` we can add:
* so-reuseport
* tcp-quickack
* tcp-cork
* tcp-fastopen

On `HttpConfiguration` we can also configure *unix domain sockets*
* host-enabled (default true)
* domain-socket
* domain-socket-enabled (default false)

PR #6670, opened 2020-01-21
Add configuration for vert.x with native transport #6669

pcasaes:
Description in issue #6669

### DIFF
--- a/docs/src/main/asciidoc/vertx.adoc
+++ b/docs/src/main/asciidoc/vertx.adoc
@@ -49,6 +49,74 @@ Otherwise, you can manually add this to the dependencies section of your `pom.xm
 </dependency>
 ----
 
+== Native Transport
+
+Vert.x is capable of using https://netty.io/wiki/native-transports.html[Netty's native transports] which offers
+performance improvements on certain platforms. To enable them you must include the appropriate dependency for your
+platform. It's usually a good idea to include both to keep your application platform agnostic. Netty is smart enough
+to use the correct one, that includes none at all on unsupported platforms:
+
+[source,xml]
+----
+<dependency>
+  <groupId>io.netty</groupId>
+  <artifactId>netty-transport-native-epoll</artifactId>
+  <classifier>linux-x86_64</classifier>
+</dependency>
+
+<dependency>
+  <groupId>io.netty</groupId>
+  <artifactId>netty-transport-native-kqueue</artifactId>
+  <classifier>osx-x86_64</classifier>
+</dependency>
+----
+
+You will also have to explicitly configure Vert.x to use the native transport. In `application.properties` add:
+
+[source]
+----
+quarkus.vertx.prefer-native-transport=true
+----
+
+Or in `application.yml`:
+
+[source,yml]
+----
+quarkus:
+  vertx:
+    prefer-native-transport: true
+----
+
+=== Native Linux Transport
+
+On Linux you can enable the following socket options:
+
+* SO_REUSEPORT
+----
+quarkus.http.so-reuse-port=true
+----
+* TCP_QUICKACK
+----
+quarkus.http.tcp-quick-ack=true
+----
+* TCP_CORK
+----
+quarkus.http.tcp-cork=true
+----
+* TCP_FASTOPEN
+----
+quarkus.http.tcp-fast-open=true
+----
+
+=== Native MacOS Transport
+
+On MacOS Sierra and above you can enable the following socket options:
+
+* SO_REUSEPORT
+----
+quarkus.http.so-reuse-port=true
+----
+
 == Accessing Vert.x
 
 Once the extension has been added, you can access the _managed_ Vert.x instance using `@Inject`:
@@ -583,6 +651,32 @@ public void init(@Observes StartupEvent e, Vertx vertx, Instance<AbstractVerticl
     }
 }
 ====
+
+== Listening to a Unix Domain Socket
+
+Listening on a unix domain socket allows us to dispense with the overhead of TCP
+if the connection to the server is established from the same host. This can happen
+if access to the service goes through a proxy which is often the case
+if you're setting up a service mesh with a proxy like Envoy.
+
+IMPORTANT: This will only work on platforms that support <<native-transport>>.
+
+To setup please enable the appropriate <<native-transport>> and set the following
+environment property:
+
+----
+quarkus.http.domain-socket=/var/run/quarkus-http.socket
+----
+
+By itself this will not disable the tcp socket which by default will open on
+`0.0.0.0:8080`. You can disable it by leaving the following environment property empty:
+
+----
+quarkus.http.host=
+----
+
+These properties can be set through Java's `-D` command line parameter or
+on `application.properties`.
 
 == Read only deployment environments
 
--- a/docs/src/main/asciidoc/vertx.adoc
+++ b/docs/src/main/asciidoc/vertx.adoc
@@ -87,6 +87,12 @@ quarkus:
     prefer-native-transport: true
 ----
 
+If all is well quarkus will log:
+
+----
+[io.qua.ver.cor.run.VertxCoreRecorder] (main) Vertx has Native Transport Enabled: true
+----
+
 === Native Linux Transport
 
 On Linux you can enable the following socket options:
@@ -655,7 +661,7 @@ public void init(@Observes StartupEvent e, Vertx vertx, Instance<AbstractVerticl
 == Listening to a Unix Domain Socket
 
 Listening on a unix domain socket allows us to dispense with the overhead of TCP
-if the connection to the server is established from the same host. This can happen
+if the connection to the quarkus service is established from the same host. This can happen
 if access to the service goes through a proxy which is often the case
 if you're setting up a service mesh with a proxy like Envoy.
 
@@ -665,14 +671,15 @@ To setup please enable the appropriate <<native-transport>> and set the followin
 environment property:
 
 ----
-quarkus.http.domain-socket=/var/run/quarkus-http.socket
+quarkus.http.domain-socket=/var/run/io.quarkus.app.socket
+quarkus.http.domain-socket-enabled=true
 ----
 
 By itself this will not disable the tcp socket which by default will open on
-`0.0.0.0:8080`. You can disable it by leaving the following environment property empty:
+`0.0.0.0:8080`. It can be explicitly disabled:
 
 ----
-quarkus.http.host=
+quarkus.http.host-enabled=false
 ----
 
 These properties can be set through Java's `-D` command line parameter or
--- a/extensions/netty/runtime/pom.xml
+++ b/extensions/netty/runtime/pom.xml
@@ -39,6 +39,19 @@
             <groupId>org.graalvm.nativeimage</groupId>
             <artifactId>svm</artifactId>
         </dependency>
+
+        <dependency>
+            <groupId>io.netty</groupId>
+            <artifactId>netty-transport-native-epoll</artifactId>
+            <classifier>linux-x86_64</classifier>
+            <optional>true</optional>
+        </dependency>
+        <dependency>
+            <groupId>io.netty</groupId>
+            <artifactId>netty-transport-native-kqueue</artifactId>
+            <classifier>osx-x86_64</classifier>
+            <optional>true</optional>
+        </dependency>
     </dependencies>
 
     <build>
--- a/extensions/vertx-core/runtime/src/test/java/io/quarkus/vertx/core/runtime/VertxCoreProducerTest.java
+++ b/extensions/vertx-core/runtime/src/test/java/io/quarkus/vertx/core/runtime/VertxCoreProducerTest.java
@@ -100,6 +100,7 @@ public class VertxCoreProducerTest {
         vc.maxWorkerExecuteTime = Optional.of(Duration.ofSeconds(1));
         vc.internalBlockingPoolSize = 20;
         vc.useAsyncDNS = false;
+        vc.preferNativeTransport = false;
         vc.eventbus = new EventBusConfiguration();
         vc.eventbus.keyCertificatePem = new PemKeyCertConfiguration();
         vc.eventbus.keyCertificatePem.keys = Optional.empty();
--- a/extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/HttpConfiguration.java
+++ b/extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/HttpConfiguration.java
@@ -113,25 +113,25 @@ public class HttpConfiguration {
     /**
      * Enable socket reuse port (linux/macOs native transport only)
      */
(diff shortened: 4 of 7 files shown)


cescoffier:
LGTM, but would like the POV from @vietj.

cescoffier:
Wondering about native actually. Did you test in native mode?

pcasaes:
I did a negative test in native (it worked). I didn't actually try to use the _native transport_ in native though. I don't know if that is actually possible.

pcasaes:
To be more specific _native transport_ was enabled in native, but I didn't check to see if it was a NOOP.

gastaldi:
Is this setting ignored in Windows 10?

extensions/netty/runtime/pom.xml
```diff
@@ -39,6 +39,17 @@
...
+        <dependency>
+            <groupId>io.netty</groupId>
+            <artifactId>netty-transport-native-epoll</artifactId>
+            <classifier>linux-x86_64</classifier>
+        </dependency>
+        <dependency>
+            <groupId>io.netty</groupId>
+            <artifactId>netty-transport-native-kqueue</artifactId>
+            <classifier>osx-x86_64</classifier>
+        </dependency>
```
gastaldi: Maybe leave these dependencies to be manually added in the project if the property is set? Not sure it makes sense to include the OSX and Linux dependencies in `quarkus-netty`

pcasaes: It's perfectly safe to include these dependencies and they are ignored in Windows. Both Spring Boot and Wildfly have them baked in (the later in netty-all). We could do as yu suggest but then we would have to add some documentation.

extensions/netty/runtime/pom.xml
```diff
@@ -39,6 +39,19 @@
...
+            <artifactId>netty-transport-native-epoll</artifactId>
+            <classifier>linux-x86_64</classifier>
+            <optional>true</optional>
+        </dependency>
+        <dependency>
+            <groupId>io.netty</groupId>
+            <artifactId>netty-transport-native-kqueue</artifactId>
+            <classifier>osx-x86_64</classifier>
+            <optional>true</optional>
+        </dependency>
```
pcasaes: The native transport dependencies are optional. You'll have to add them in manually to your project but the version will be taken care for you.

docs/src/main/asciidoc/vertx.adoc
```diff
@@ -47,6 +47,44 @@ Otherwise, you can manually add this to the dependencies section of your `pom.xm
 </dependency>
 ----
 
+== Native Transport
+
+Vert.x is capable of using https://netty.io/wiki/native-transports.html[Netty's native transports] which offers
+performance improvements on certain platforms. To enable them you must include one or both of the following dependencies:
```
gastaldi: How to know when to include one or both? Aren't these dependencies OS specific?

pcasaes: The referenced documentation https://netty.io/wiki/native-transports.html is where you get that information. But it won't break if you include both of them. Netty is smart enough to fallback to a non native transport. Usually projects include both so they'll be platform agnostic.

pcasaes: To be clear. If you include both netty is smart enough to know which one it should use (or non tall) depending on the platform.

gastaldi: Gotcha. Thanks!

pcasaes: I improved the documentation to make it clearer.

extensions/vertx-core/runtime/src/test/java/io/quarkus/vertx/core/runtime/VertxCoreProducerTest.java
```diff
@@ -61,6 +61,7 @@ public void shouldNotFailWithDefaultConfig() {
         configuration.workerPoolSize = 10;
         configuration.warningExceptionTime = Duration.ofSeconds(1);
         configuration.internalBlockingPoolSize = 5;
+        configuration.preferNativeTransport = false;
```
gastaldi: No need to initialize as the default value is false anyway

pcasaes: You're right.

extensions/vertx-core/runtime/src/test/java/io/quarkus/vertx/core/runtime/VertxCoreProducerTest.java
```diff
@@ -100,6 +100,7 @@ private VertxConfiguration createDefaultConfiguration() {
         vc.maxWorkerExecuteTime = Optional.of(Duration.ofSeconds(1));
         vc.internalBlockingPoolSize = 20;
         vc.useAsyncDNS = false;
+        vc.preferNativeTransport = false;
```
gastaldi: No need to initialize here too, I suppose? :)

vietj (commented):
A few comments / questions :

- would it be good to log if using native actually worked using `vertx.isNativeTransportEnabled()` ?
- would be interesting to allow binding of the HTTP server on a domain socket path via configuration ?
- it would be interesting to document that Vert.x Web Client can then access domain socket bound servers or give a way in Quarkus to benefit from this

pcasaes:
> would it be good to log if using native actually worked using vertx.isNativeTransportEnabled() ?

It certainly would!

> would be interesting to allow binding of the HTTP server on a domain socket path via configuration ?
it would be interesting to document that Vert.x Web Client can then access domain socket bound servers or give a way in Quarkus to benefit from this

It would be, but maybe on a separate ticket/pr?

pcasaes:
> would be interesting to allow binding of the HTTP server on a domain socket path via configuration ?

Are you thinking either/or or both domain sockets and tcp?

extensions/vertx-core/runtime/src/test/java/io/quarkus/vertx/core/runtime/VertxCoreProducerTest.java
pcasaes: This method creates a default configuration and sets all values. I figured best to keep it consistent.

pcasaes:
@vietj I added in the ability to configure a domain socket with documentation on how and why to do this.

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/VertxHttpRecorder.java
```diff
@@ -541,39 +592,85 @@ public void addRoute(RuntimeValue<Router> router, Function<Router, Route> route,
 
     private static class WebDeploymentVerticle extends AbstractVerticle {
 
-        private final int port;
-        private final int httpsPort;
-        private final String host;
```
pcasaes: This variables are redundant since their values are in `httpOptions` and `httpsOptions`.

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/HttpConfiguration.java
```diff
@@ -96,6 +96,36 @@
     @ConfigItem(name = "auth.session.encryption-key")
     public Optional<String> encryptionKey;
 
+    /**
+     * Enable socket reuse port
+     */
+    @ConfigItem(defaultValue = "false", name = "so-reuse-port")
+    public boolean soReusePort;
```
stuartwdouglas: Why have you changed the default to false? I thought the usual default value here would be true?

pcasaes: It's a new configuration. In this case one that enables [SO_REUSEPORT](https://lwn.net/Articles/542629/). At a first glance I figured it safer to leave this new configurations turned off. 

All the new configurations here only take effect if native transport is enabled. I'll add that to the documentation.

stuartwdouglas: Sorry, I should have read it more closely, I was thinking it was SO_REUSEADDR

pcasaes: No worries. I added this to the docs.

vietj (approved):
This is looking good and helpful to me.

pcasaes:
@cescoffier veitj gave his thumbs up. Can we get this merged?

gastaldi:
Can you squash your commits into a single one? Thanks!

pcasaes:
Sure

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/HttpConfiguration.java
```diff
@@ -35,7 +35,7 @@
      * The HTTP host
      */
     @ConfigItem(defaultValue = "0.0.0.0")
-    public String host;
+    public Optional<String> host;
```
cescoffier: Why switching this property as optional? It has a default value, so the optional will never be empty.

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/HttpConfiguration.java
```diff
@@ -96,6 +96,36 @@
     @ConfigItem(name = "auth.session.encryption-key")
     public Optional<String> encryptionKey;
 
+    /**
+     * Enable socket reuse port (linux/macOs native transport only)
+     */
+    @ConfigItem(defaultValue = "false", name = "so-reuse-port")
+    public boolean soReusePort;
```
cescoffier: Any reason to set the name? The name you use are the default ones.

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/VertxHttpRecorder.java
```diff
@@ -420,9 +447,13 @@ private static HttpServerOptions createSslOptions(HttpConfiguration httpConfigur
             }
         }
         serverOptions.setSsl(true);
-        serverOptions.setHost(httpConfiguration.host);
+        serverOptions.setHost(httpConfiguration.host.get());
```
cescoffier: no need for `get` if you remove the optional.

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/VertxHttpRecorder.java
```diff
@@ -345,6 +368,10 @@ public void handle(AsyncResult<Void> event) {
      */
     private static HttpServerOptions createSslOptions(HttpConfiguration httpConfiguration, LaunchMode launchMode)
             throws IOException {
+        if (!httpConfiguration.host.isPresent()) {
```
cescoffier: cannot be false, it has a default value.

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/VertxHttpRecorder.java
```diff
@@ -497,13 +528,33 @@ private static String findKeystoreFileType(Path storePath) {
 
     private static HttpServerOptions createHttpServerOptions(HttpConfiguration httpConfiguration,
             LaunchMode launchMode, String websocketSubProtocols) {
+        if (!httpConfiguration.host.isPresent()) {
```
cescoffier: cannot be `false`. It has a default value.
In addition, the caller need to be aware of the possibility to get `null`.

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/VertxHttpRecorder.java
```diff
@@ -497,13 +528,33 @@ private static String findKeystoreFileType(Path storePath) {
...
+            return null;
+        }
         // TODO other config properties
         HttpServerOptions options = new HttpServerOptions();
-        options.setHost(httpConfiguration.host);
+        options.setHost(httpConfiguration.host.get());
```
cescoffier: no need for `get` if you remove the optional.

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/VertxHttpRecorder.java
cescoffier: It has a `null`-check, so it's ok, but it can still never be `false`.

cescoffier (changes requested):

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/HttpConfiguration.java
pcasaes: So if you only want to use unix domain sockets you can set
`
quarkus.http.host=
`

And it will disable listening on `host:port`. Do you think it would be clearer to use a different property to disable this?

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/HttpConfiguration.java
pcasaes: You're right. I'll fix this.

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/HttpConfiguration.java
pcasaes: maybe have remove the `Optional` and add this properties with these default?

```
quarkus.http.host.enabled=true
quarkus.http.host.domain-socket.enabled=false
```

pcasaes: `quarkus.http.host.domain-socket` won't have a default value so it must be optional.

So then we'd only need
```
quarkus.http.host.enabled=true
```

But having `domain-socket` and `http` enabled differently is a bit strange.

pcasaes: Another option is to give `domain-sockets` a default value `/var/run/io.quarkus.app.socket`. That would keep it consistent.

pcasaes: @cescoffier  I went ahead and made these changes.

gastaldi:
@cescoffier can you review the PR again when possible? Thanks

gsmet:
@cescoffier should we make progress on that one?

cescoffier:
Everything is fine from my point of view.

cescoffier (approved):

cescoffier:
Everything is fine from my point of view.

PR #6672, opened 2020-01-21
Add support for .yml in addition to .yaml

### LINKED ISSUE #6668: quarkus-config-yaml does not recognize "application.yml"
**Describe the bug**
While the plugin recognizes `application.yaml`, it does not recognize `application.yml`, and thus does not get the configuration as normally expected. (`.yml` is accepted by Spring, for instance)

**Expected behavior**
Quarkus would pick up the configuration placed in `application.yml`

**Actual behavior**
Quarkus does not pick up the configuration placed in `application.yml`

geoand:
Fixes: #6668

### DIFF
--- a/docs/src/main/asciidoc/config.adoc
+++ b/docs/src/main/asciidoc/config.adoc
@@ -625,6 +625,11 @@ quarkus:
     password: quarkus
 ----
 
+[NOTE]
+====
+Quarkus also supports using `application.yml` as the name of the YAML file. The same rules apply for this file as for `application.yaml`.
+====
+
 === Profile dependent configuration
 
 Providing profile dependent configuration with YAML is done like with properties.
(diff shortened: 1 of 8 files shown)


geoand:
@dmlloyd was there any specific reason that `.yml` wasn't added initially?

docs/src/main/asciidoc/config.adoc
```diff
@@ -625,6 +625,11 @@ quarkus:
     password: quarkus
 ----
 
+[NOTE]
+====
+Quarkus also supports using `application.yml` as the name of YAML file. The same rules apply for this file as for `application.yaml`.
```
gsmet:
```suggestion
Quarkus also supports using `application.yml` as the name of the YAML file. The same rules apply for this file as for `application.yaml`.
```

gsmet (changes requested):
Looks OK to me except for a minor gripe on the doc.

docs/src/main/asciidoc/config.adoc
geoand: Good one! A classic mistake I always make :). You can probably spot those with closed eyes now :P

gsmet (approved):

dmlloyd:
Just avoiding confusion I guess.  What about microprofile-config.yml?

geoand:
Should we add `microprofile-config.yml` or wait until someone complains?

gsmet:
I would say let's leave it at that.

geoand:
> I would say let's leave it at that.

@dmlloyd OK with this approach?

dmlloyd (approved):

PR #6673, opened 2020-01-21
add arc, devtools and infra labeling

maxandersen:
added area/devtools label for this to work.

### DIFF
--- a/.github/boring-cyborg.yml
+++ b/.github/boring-cyborg.yml
@@ -70,6 +70,7 @@ labelPRBasedOnFilePath:
         - integration-tests/amazon-lambda-http-resteasy/**/*
     area/arc:
         - extensions/arc/**/*
+        - independent-projects/arc/**/*
     area/artemis:
         - extensions/artemis-core/**/*
         - extensions/artemis-jms/**/*
@@ -85,6 +86,10 @@ labelPRBasedOnFilePath:
         - .dependabot/config.yml
         - bom/**/*
         - build-parent/**/*
+    area/devtools:
+        - devtools/**/*
+        - independent-projects/bootstrap/**/*
+        - independent-projects/tools/**/*
     area/documentation:
         - docs/**/*
     area/gradle:
--- a/.github/boring-cyborg.yml
+++ b/.github/boring-cyborg.yml
@@ -124,6 +124,11 @@ labelPRBasedOnFilePath:
         - integration-tests/infinispan-cache-jpa/**/*
         - integration-tests/infinispan-client/**/*
         - integration-tests/infinispan-embedded/**/*
+     area/infra-automation:
+        - .github/**/*
+        - azure-mvn-settings.xml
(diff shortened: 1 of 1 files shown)


geoand (approved):

maxandersen:
you are too fast @geoand  ;) just added area/infra for .github / azure changes ...

geoand:
Then you should probably change the title as well :P

.github/boring-cyborg.yml
```diff
@@ -119,6 +124,10 @@ labelPRBasedOnFilePath:
         - integration-tests/infinispan-cache-jpa/**/*
         - integration-tests/infinispan-client/**/*
         - integration-tests/infinispan-embedded/**/*
+     area/infra:
+        - .github/**/*
+        - azure-mvn-settings.xml
+        - azure-pipelines.yml
```
gsmet: Woot! I was going to add that one :).

gsmet: You should add the `ci-templates` directory too.

gsmet: I withdraw this one. If we add a something new in the native pipeline, we don't want to have `area/infra` otherwise it won't be properly detected as an orphan.

Is there a way to add labels only if we haven't added any others? I'm thinking about this for infra and dependencies. They should be added if the PR doesn't touch anything else. In the case of dependencies, it's just unnecessary noise, in the case of infra, it's more annoying as issues with this label are ignored in the orphan detection.

maxandersen: at the moment we can't tweak the label adds.

For dependencies - why wouldn't you want that added whenever the bom.xml is modified ? 

What is it you are concretely looking for in your orphan deteection ? Maybe we want to add a separate label for that and leave area/infra as what is used to be able to find pullrequests that relates to a specific area ? (That’s what I'm after - being able to setup a query  that lists the areas i'm interested in being aware of - area/infra and dependencies being one no matter if its happening together with 0 or many other changes)

gastaldi (approved):
Looks good, nice one!

.github/boring-cyborg.yml
gsmet: `area/infra` has been specifically added for the orphan detection. Everything having this label is ignored as it's about the infra and not about Quarkus itself.

maxandersen: okey, so what name do you want to use for "this pull-request relates to area/infra but that doesn't mean it should be included in orphan detection" ... how about "area/build-automation" ?

if that works i'll update the PR and use that instead and you can continue to use "area/infra" as a special marker ?

maxandersen: @gsmet I've added `ci-templates` and made it use area/infra-automation rather than `area/infra` which you prefer to have as a magic marker for non-orphans.

okey to push this one now as this one would actually label a good bunch of the pullrequests not being labeled today.

gsmet (approved):

PR #6674, opened 2020-01-21
Thread safety issues

### LINKED ISSUE #6608: OIDC test checking the custom principal name is unstable

stuartwdouglas:
This commit fixes a lot of potential thread safety issues, by making sure that fields on application scoped beans that are initialised via a recorder are thread safe.

Fixes #6608

### DIFF
--- a/extensions/hibernate-orm/runtime/src/main/java/io/quarkus/hibernate/orm/runtime/DefaultEntityManagerFactoryProducer.java
+++ b/extensions/hibernate-orm/runtime/src/main/java/io/quarkus/hibernate/orm/runtime/DefaultEntityManagerFactoryProducer.java
@@ -5,11 +5,11 @@ import javax.inject.Singleton;
 import javax.persistence.EntityManagerFactory;
 import javax.persistence.PersistenceUnit;
 
-@Singleton
 public class DefaultEntityManagerFactoryProducer {
 
     @Produces
     @PersistenceUnit
+    @Singleton
     EntityManagerFactory entityManagerFactory;
 
 }
--- a/extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/DefaultTenantConfigResolver.java
+++ b/extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/DefaultTenantConfigResolver.java
@@ -20,9 +20,9 @@ public class DefaultTenantConfigResolver {
     @Inject
     Instance<TenantConfigResolver> tenantConfigResolver;
 
-    private Map<String, TenantConfigContext> tenantsConfig;
-    private TenantConfigContext defaultTenant;
-    private Function<OidcTenantConfig, TenantConfigContext> tenantConfigContextFactory;
+    private volatile Map<String, TenantConfigContext> tenantsConfig;
+    private volatile TenantConfigContext defaultTenant;
+    private volatile Function<OidcTenantConfig, TenantConfigContext> tenantConfigContextFactory;
 
     TenantConfigContext resolve(RoutingContext context) {
         if (tenantConfigResolver.isAmbiguous()) {
--- a/extensions/agroal/runtime/src/main/java/io/quarkus/agroal/runtime/health/DataSourceHealthCheck.java
+++ b/extensions/agroal/runtime/src/main/java/io/quarkus/agroal/runtime/health/DataSourceHealthCheck.java
@@ -22,7 +22,7 @@ import io.quarkus.arc.Arc;
 @ApplicationScoped
 public class DataSourceHealthCheck implements HealthCheck {
     private static final String DEFAULT_DS = "__default__";
-    private Map<String, DataSource> dataSources = new HashMap<>();
+    private final Map<String, DataSource> dataSources = new HashMap<>();
 
     @PostConstruct
     protected void init() {
--- a/extensions/elytron-security-oauth2/runtime/src/main/java/io/quarkus/elytron/security/oauth2/runtime/auth/OAuth2AuthMechanism.java
+++ b/extensions/elytron-security-oauth2/runtime/src/main/java/io/quarkus/elytron/security/oauth2/runtime/auth/OAuth2AuthMechanism.java
@@ -5,8 +5,6 @@ import java.util.concurrent.CompletionStage;
 
 import javax.enterprise.context.ApplicationScoped;
 
-import org.jboss.logging.Logger;
-
 import io.netty.handler.codec.http.HttpHeaderNames;
 import io.netty.handler.codec.http.HttpResponseStatus;
 import io.quarkus.security.credential.TokenCredential;
@@ -22,8 +20,6 @@ import io.vertx.ext.web.RoutingContext;
  */
 @ApplicationScoped
 public class OAuth2AuthMechanism implements HttpAuthenticationMechanism {
-
-    private static final Logger log = Logger.getLogger(OAuth2AuthMechanism.class);
 
     /**
      * Extract the Authorization header and validate the bearer token if it exists. If it does, and is validated, this
--- a/extensions/elytron-security/runtime/src/main/java/io/quarkus/elytron/security/runtime/ElytronSecurityDomainManager.java
+++ b/extensions/elytron-security/runtime/src/main/java/io/quarkus/elytron/security/runtime/ElytronSecurityDomainManager.java
@@ -3,13 +3,10 @@ package io.quarkus.elytron.security.runtime;
 import javax.enterprise.context.ApplicationScoped;
 import javax.enterprise.inject.Produces;
 
-import org.jboss.logging.Logger;
 import org.wildfly.security.auth.server.SecurityDomain;
 
 @ApplicationScoped
 public class ElytronSecurityDomainManager {
-
-    private static Logger log = Logger.getLogger(ElytronSecurityDomainManager.class);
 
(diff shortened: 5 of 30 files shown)


extensions/hibernate-orm/runtime/src/main/java/io/quarkus/hibernate/orm/runtime/DefaultEntityManagerFactoryProducer.java
```diff
@@ -1,11 +1,9 @@
 package io.quarkus.hibernate.orm.runtime;
 
 import javax.enterprise.inject.Produces;
-import javax.inject.Singleton;
 import javax.persistence.EntityManagerFactory;
 import javax.persistence.PersistenceUnit;
 
-@Singleton
```
mkouba: Why do you remove the scope here? Since the produced `EntityManagerFactory` is dependent a new instance of this producer will be created for each `EntityManagerFactory`  injection point...

stuartwdouglas: I could have sworn the producer also had a scope on it, oops.

extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/DefaultTenantConfigResolver.java
```diff
@@ -20,9 +20,9 @@
     @Inject
     Instance<TenantConfigResolver> tenantConfigResolver;
 
-    private Map<String, TenantConfigContext> tenantsConfig;
-    private TenantConfigContext defaultTenant;
-    private Function<OidcTenantConfig, TenantConfigContext> tenantConfigContextFactory;
+    private volatile Map<String, TenantConfigContext> tenantsConfig;
+    private volatile TenantConfigContext defaultTenant;
+    private volatile Function<OidcTenantConfig, TenantConfigContext> tenantConfigContextFactory;
```
geoand: So I assume this is the part that actually fixes #6608?

stuartwdouglas: I hope so. Without this the HTTP thread could read incorrect config, so was likely missing the principalClaim. This is the only explanation I can see for the type of failure we are seeing.

geoand: I see, thanks for the explanation. Let's see what CI thinks :)

extensions/hibernate-orm/runtime/src/main/java/io/quarkus/hibernate/orm/runtime/DefaultEntityManagerFactoryProducer.java
geoand: @mkouba are you good with the move of the annotation to the producer method? I would like to get this approved (I think all the changes make sense) and merged if CI passes.

geoand:
@sberyozkin mind taking a look at the security stuff as well?

extensions/hibernate-orm/runtime/src/main/java/io/quarkus/hibernate/orm/runtime/DefaultEntityManagerFactoryProducer.java
mkouba: Looks good so +1 if CI passes ;-).

sberyozkin:
@geoand Sure, looks good to me. 
@stuartwdouglas I believe Pedro @pedroigor  initialized various OIDC config fields to make it simpler for the custom tenant resolvers to write the code like `config.getToken.setIssuer(...)`, but I know Pedro had a nice idea of creating a builder API, so we can do it with him later

sberyozkin (approved):

geoand (approved):

geoand:
The main IT is failing in native:

```
[ERROR] Errors: 
[ERROR]   RestClientITCase>RestClientTestCase.testMicroprofileClientComplex:70 » JsonPath
[ERROR]   RestClientITCase>RestClientTestCase.testMicroprofileClientComplexCdi:79 » JsonPath
[ERROR]   RestClientITCase>RestClientTestCase.testMicroprofileClientData:48 » JsonPath F...
[ERROR]   RestClientITCase>RestClientTestCase.testMicroprofileClientDataCdi:55 » JsonPath
[INFO] 
[ERROR] Tests run: 119, Failures: 37, Errors: 4, Skipped: 2
[INFO] 
[INFO] 
[INFO] --- maven-failsafe-plugin:2.22.1:verify (default) @ quarkus-integration-test-main ---
[INFO] BUILD FAILURE
[INFO] Total time:  09:44 min
[INFO] Finished at: 2020-01-21T10:39:37Z

```

sberyozkin:
Seems unrelated as this PR updates the fields visibility

stuartwdouglas:
The failure appears to be  https://github.com/oracle/graal/issues/1927 which is a race condition. It would be pretty ironic if fixing our own race conditions changed the timing and made the Graal race always appear. This is fixed in 19.3.1, but not in 19.2 which is what CI is using.

mkouba:
`io.quarkus.it.keycloak.BearerTokenAuthorizationTest` also fails with:
```
2020-01-21 10:45:26,375 ERROR [io.qua.ver.htt.run.QuarkusErrorHandler] (executor-thread-1) HTTP Request to /tenant/tenant-d/api/user failed, error id: 1cadd722-25c5-46e8-b881-996f7b81c8ae-1: java.lang.NullPointerException
	at io.quarkus.oidc.runtime.OidcRecorder.createTenantContext(OidcRecorder.java:57)
```

geoand:
> The failure appears to be [oracle/graal#1927](https://github.com/oracle/graal/issues/1927) which is a race condition. It would be pretty ironic if fixing our own race conditions changed the timing and made the Graal race always appear. This is fixed in 19.3.1, but not in 19.2 which is what CI is using.

Very ironic indeed.

We are going to drop 19.2.1 soon (but I don't see that being possible for 1.2 since 1.2.CR1 has already been released), so maybe this PR needs to wait for our move to 19.3.1 only?

stuartwdouglas:
looks like the config is being directly instantiated so it was not safe to drop all the fields being initalized. I don't really like it but I will add it back.

sberyozkin:
Hi @geoand I guess the backport can be avoided if there is some concern that 19.2.1 users might get affected though the PR really makes things safer, so not 100% sure if the backport would be problematic or not :-)

geoand:
Well if CI passes, then we should be just fine with backporting (since CI used 19.2.1). However if we run into oracle/graal#1927 like @stuartwdouglas mentioned, we unfortunately have to wait since that fix is only part of 19.3.1

dmlloyd:
Many of these properties seem to be read-only in practice... maybe it would be nicer (for users especially) if we instead ensure that publication happens after `initialize` returns (like how constructors work) from the framework?

sberyozkin:
Hi All, unless there are some objections, I'm going to go ahead with the merge.
Hi David @dmlloyd, are you referring to the existing pattern of some producers setting the now `volatile` fields from `initialize` ? I can only assume that changing the pattern will require some extra updates :-)

dmlloyd:
> Hi All, unless there are some objections, I'm going to go ahead with the merge.
> Hi David @dmlloyd, are you referring to the existing pattern of some producers setting the now `volatile` fields from `initialize` ? I can only assume that changing the pattern will require some extra updates :-)

Right... I'm just thinking that if we're adding `volatile` all over the place then we probably have a design problem going on.

sberyozkin:
@dmlloyd thanks, not sure I can propose anything meaningful but if there will be some issue open then I might try :-)
OK. going ahead with the merge, please don't kill me if some next PR will start failing with the OIDC `principalClaim` test failure :-)

stuartwdouglas:
@dmlloyd yea, this is not great, as there is still a window where in theory a request could come in before the bean is initialised. I can think of two options:

- Add direct support in ArC to allow some way for the required info to be injected into the constructor
- Add a 3rd phase after RUNTIME_INIT (SOCKET_OPEN ?) and make sure that all actions that open a socket or otherwise begin request processing happen in this phase, while everything else happens in RUNTIME_INIT. This should establish a happens before relationship that means this should always be visible.

mkouba:
> Add direct support in ArC to allow some way for the required info to be injected into the constructor

So I think the correct approach would be to register a `RuntimeBeanBuildItem` for the "initialization context" (contains all the recorder init params) so that the a singleton/application bean can inject this context in the constructor. The problem is that `RuntimeBeanBuildItem` can currently introduce build chain cycles easily (`RuntimeBeanProcessor` produces `GeneratedBeanBuildItem` which is consumed by `BeanArchiveProcessor`). I think that we should try to reimplement `RuntimeBeanBuildItem` with synthetic beans and `BeanRegistrar`. I'll look into this for 1.3...

mkouba:
FYI: https://github.com/quarkusio/quarkus/pull/6721

If we merge this PR I can prepare a POC to replace the need for volatile fields in app scoped beans.

mkouba:
This is my proposal to replace the need for volatile fields: https://github.com/quarkusio/quarkus/pull/6862

It's basically a combination of a synthetic bean whose instance is supplied by a recorder and constructor injection.

PR #6676, opened 2020-01-21
Use network bridge for docker to avoid hostname can't be null fail

rsvoboda:
Use network bridge for docker to avoid `hostname can't be null` fail

checked via `./mvnw -fn clean verify -Ddocker -Dtest-mariadb -Dtest-mssql -Dtest-mysql -Dtest-postgresql -Dtest-gelf -Dtest-neo4j -Dtest-keycloak -pl integration-tests/jpa-mariadb,integration-tests/jpa-mssql,integration-tests/reactive-mysql-client` command


gastaldi (approved):
LGTM

geoand (approved):

PR #6681, opened 2020-01-21
Ensure that a missing user.country system property doesn't break the application

geoand:


gsmet (approved):
Looks good, thanks!

geoand:
You are welcome!

geoand:
Marking as WIP because we need to take care of `user.language` missing as well.

geoand:
@gsmet I added another commit that makes `user.language` optional as well

core/runtime/src/main/java/io/quarkus/runtime/configuration/LocaleConverter.java
```diff
@@ -29,6 +32,11 @@ public Locale convert(final String value) {
             return null;
         }
 
+        if (localeValue.equals("-")) {
```
geoand: The idea here is to return `null` only when no locale value is set explicitly, but instead the defaults are both empty. This allows the user explicitly set locale strings like `en-` or `en-US` even when there are no default values.
If the default values are empty and the user doesn't specify any locales, then a higher up layer will end up throwing an exception like:

```
java.lang.RuntimeException: java.util.NoSuchElementException: Property quarkus.locales not found
```

geoand: In this case the warning message will at least give the user the opportunity to know to set `user.language` instead of having to set `quarkus.locales` and `quarkus.default-locale`

geoand: After talking we @gsmet we decided to use `en` as the last line of defence, so this stuff is no longer needed

gsmet (approved):

geoand:
I forced pushed because CI failed to check out the PR...

test-framework/junit5-internal/src/main/java/io/quarkus/test/QuarkusUnitTest.java
```diff
@@ -133,6 +135,18 @@ public QuarkusUnitTest setLogFileName(String logFileName) {
         return this;
     }
 
+    // set a Runnable that will run before ANYTHING else is done
+    public QuarkusUnitTest setBeforeAllCustomizer(Runnable beforeAllCustomizer) {
+        this.beforeAllCustomizer = beforeAllCustomizer;
```
gastaldi: Why not use the `@BeforeAll` annotation in a static method?

extensions/hibernate-validator/deployment/src/test/java/io/quarkus/hibernate/validator/test/UserCountryNotSetValidatorLocaleTest.java
```diff
@@ -0,0 +1,56 @@
...
+                public void run() {
+                    userCountry = System.clearProperty("user.country");
+                    userLanguage = System.clearProperty("user.language");
+                }
+            }).setAfterAllCustomizer(new Runnable() {
+                @Override
+                public void run() {
+                    System.setProperty("user.country", userCountry);
+                    System.setProperty("user.language", userLanguage);
+                }
```
gastaldi: Using @BeforeAll and @AfterAll static methods should be the standard way of invoking code before and after tests. Not sure if it's worth adding another strategy to do the same thing, unless I am missing something, of course

gastaldi (changes requested):
I would like to know the reason behind the QuarkusUnitTest changes before merging this PR

test-framework/junit5-internal/src/main/java/io/quarkus/test/QuarkusUnitTest.java
geoand: That's the first thing I tried :wink: . However that is executed _after_ the Quarkus extension and I needed it to execute before the build.

extensions/hibernate-validator/deployment/src/test/java/io/quarkus/hibernate/validator/test/UserCountryNotSetValidatorLocaleTest.java
geoand: See my comment above

gastaldi (approved):
Thanks @geoand, yeah, you're right. 

Probably there is another elegant way of enabling this in the JUnit 5 API but this already looks good enough :smiley:

geoand:
> Thanks @geoand, yeah, you're right.
> 
> Probably there is another elegant way of enabling this in the JUnit 5 API but this already looks good enough

Ping me if you create a PR that uses it :)

gsmet:
Rebase to trigger CI again. There's something weird with the native tests failing with a git error. Hopefully, it was only transient but we had it for several PRs in a row.

geoand:
Thanks for the heads-up.

PR rebased

gsmet:
Looks like CI will never pass for this one. Let's merge it, the JDK 8 build went OK, that's all we need.

PR #6686, opened 2020-01-21
Vert.x Verticles support

cescoffier:
* Verticle classes should be registered for reflection (when `deployVerticle` is used with class name)
* Router restarted before Vert.x which led to NPE
* Add IT and live reload tests


gsmet:
Rebased to trigger CI.

gsmet (approved):
The changes make sense to me and the rest is tests so let's see what CI has to say.

PR #6691, opened 2020-01-21
spring-data-jpa: support custom return type for @Query annotation

### LINKED ISSUE #6205: Quarkus spring data jpa does not support interface/class mapping for custom JPQL @Query in CrudRepository
JPQL Query does not return list of mapped interface objects as in standard Spring data but returns only list of object arrays.

ebullient:
resolves #6205


geoand:
@gytis would you like to also review this one?

geoand:
Thanks a lot for the contribution @ebullient!

I will hopefully have a look tomorrow

gytis:
> @gytis would you like to also review this one?

I'll go through it first thing in the morning.

geoand:
Darn CI is being flaky lately... I would really like a full CI run before reviewing. @ebullient could you rebase onto master and see if CI unblocks?

extensions/spring-data-jpa/deployment/src/main/java/io/quarkus/spring/data/deployment/SpringDataJPAProcessor.java
```diff
@@ -58,7 +61,8 @@ void build(CombinedIndexBuildItem index,
                 indexIndex);
 
         removeNoRepositoryBeanClasses(interfacesExtendingCrudRepository);
-        implementCrudRepositories(generatedBeans, additionalBeans, interfacesExtendingCrudRepository, indexIndex);
+        implementCrudRepositories(generatedBeans, additionalBeans, generatedClasses,
```
geoand: How about we move `generatedClasses` right after `generatedBeans` instead of next to `additionalBeans`. It will make things a tad more consistent I believe.

extensions/spring-data-jpa/deployment/src/main/java/io/quarkus/spring/data/deployment/generate/GenerationUtil.java
```diff
@@ -2,21 +2,28 @@
...
 import org.jboss.jandex.Type;
 
+import io.quarkus.deployment.util.HashUtil;
 import io.quarkus.gizmo.MethodDescriptor;
 import io.quarkus.spring.data.deployment.DotNames;
 
 public final class GenerationUtil {
 
+    private static final Pattern SELECT_CLAUSE = Pattern.compile("select\\s+(.+)\\s+from", Pattern.CASE_INSENSITIVE);
+    private static final Pattern FIELD_ALIAS = Pattern.compile("\\s+[as|AS]+\\s+(([\\w\\.]+))");
```
geoand: The IDE is telling warning me about `Redundant group nesting` for `(([\w\.]+))`

extensions/spring-data-jpa/deployment/src/main/java/io/quarkus/spring/data/deployment/generate/GenerationUtil.java
```diff
@@ -53,4 +60,51 @@ static MethodDescriptor toMethodDescriptor(String generatedClassName, MethodInfo
         return MethodDescriptor.ofMethod(generatedClassName, methodInfo.name(), methodInfo.returnType().name().toString(),
                 parameterTypesStr.toArray(new String[0]));
     }
+
+    static String getSimpleName(DotName ifaceName) {
```
geoand: I think this method is unused

extensions/spring-data-jpa/deployment/src/main/java/io/quarkus/spring/data/deployment/generate/GenerationUtil.java
```diff
@@ -53,4 +60,51 @@ static MethodDescriptor toMethodDescriptor(String generatedClassName, MethodInfo
...
+        return (ifaceName.isInner() ? ifaceName.local() : ifaceName.withoutPackagePrefix());
+    }
+
+    static DotName createImplDotName(DotName ifaceName) {
```
geoand: Perhaps this and the other newly added methods should be moved to a different utility class that is more aptly named?

extensions/spring-data-jpa/deployment/src/main/java/io/quarkus/spring/data/deployment/generate/CustomQueryMethodsAdder.java
```diff
@@ -218,6 +229,32 @@ public void add(ClassCreator classCreator, FieldDescriptor entityClassFieldDescr
...
+                        if (fieldNames.isEmpty()) {
+                            throw new IllegalArgumentException(
+                                    "Custom Query annotations must specify field aliases when used with a custom type. "
+                                            + "Offending query string is \"" + queryString + "\" on method " + methodName
+                                            + " of Repository " + repositoryName);
+                        }
+
+                        // If the custom type is an interface, we may need to generate the implementation
+                        ClassInfo resultClassInfo = index.getClassByName(customResultTypeName);
+                        if (Modifier.isInterface(resultClassInfo.flags())) {
```
geoand: What happens if it's not an interface?

In other places we do have code that can deal with this situation (by instantiating a JavaBean for example), but does Spring Data support that?

extensions/spring-data-jpa/deployment/src/main/java/io/quarkus/spring/data/deployment/generate/CustomQueryMethodsAdder.java
```diff
@@ -31,13 +36,18 @@
 
 public class CustomQueryMethodsAdder extends AbstractMethodsAdder {
 
+    private static final Logger log = Logger.getLogger(CustomQueryMethodsAdder.class.getName());
```
geoand: It seems like this is unused

extensions/spring-data-jpa/deployment/src/main/java/io/quarkus/spring/data/deployment/generate/CustomQueryMethodsAdder.java
```diff
@@ -298,4 +336,119 @@ private ResultHandle generateSort(Integer sortParameterIndex, MethodCreator meth
...
+        if (customImplName == null) {
+            customImplName = GenerationUtil.createImplDotName(interfaceName);
+            try (ClassCreator implClassCreator = ClassCreator.builder().classOutput(nonBeansClassOutput)
+                    .interfaces(interfaceName.toString()).className(customImplName.toString())
+                    .build()) {
+
+                List<MethodInfo> methods = interfaceInfo.methods();
+
+                // Work with methods in declared order
+                for (int i = 0; i < methods.size(); i++) {
```
geoand: for each loop perhaps would be nicer?

extensions/spring-data-jpa/deployment/src/main/java/io/quarkus/spring/data/deployment/generate/CustomQueryMethodsAdder.java
```diff
@@ -298,4 +336,119 @@ private ResultHandle generateSort(Integer sortParameterIndex, MethodCreator meth
...
+                    if (!method.parameters().isEmpty()) {
+                        throw new IllegalArgumentException("Method " + method.name() + " of interface " + interfaceName
+                                + " is not a getter method since it defined parameters");
+                    }
+                    Type returnType = method.returnType();
+                    if (returnType.kind() == Type.Kind.VOID) {
+                        throw new IllegalArgumentException("Method " + method.name() + " of interface " + interfaceName
+                                + " is not a getter method since it returns void");
+                    }
+
```
geoand: I think we also need to be checking for default methods and skip over those

extensions/spring-data-jpa/deployment/src/main/java/io/quarkus/spring/data/deployment/generate/CustomQueryMethodsAdder.java
```diff
@@ -298,4 +336,119 @@ private ResultHandle generateSort(Integer sortParameterIndex, MethodCreator meth
...
+                    // create getter (based on the interface)
+                    try (MethodCreator getter = implClassCreator.getMethodCreator(getterName, returnType.toString())) {
+                        getter.setModifiers(Modifier.PUBLIC);
+                        getter.returnValue(getter.readInstanceField(field, getter.getThis()));
+                    }
+                }
+
+                // Generate a static 'convert' method, to convert Object[] to field values
+                try (MethodCreator convert = implClassCreator.getMethodCreator("convert",
```
geoand: Although this does do what it's intended to do, I suspect however that it will be problematic when the same Class is returned from multiple repository methods that don't use the same ordering for fields.

In you test you have:

```
@Query(value = "SELECT publicationYear as publicationYear, COUNT(*) as count FROM Book GROUP BY publicationYear")
    List<BookCountByYear> findAllByPublicationYear();
```

What happens if I do this:

```
@Query(value = "SELECT COUNT(*) as count, publicationYear as publicationYear FROM Book GROUP BY publicationYear")
    List<BookCountByYear> findAllByPublicationYear2();
```

I actually did try this and it failed :(

geoand (changes requested):
Thanks a lot of this work! It's a great step!
I did add a  bunch of comments most of which are small but there are a couple serious ones in there.

extensions/spring-data-jpa/deployment/src/main/java/io/quarkus/spring/data/deployment/generate/CustomQueryMethodsAdder.java
geoand: What I think you'll need to do is take into account the `fieldNames` in the `convert` method by adding it as a parameter and then doing some magic to work out the proper order

extensions/spring-data-jpa/deployment/src/main/java/io/quarkus/spring/data/deployment/generate/CustomQueryMethodsAdder.java
```diff
@@ -298,4 +336,119 @@ private ResultHandle generateSort(Integer sortParameterIndex, MethodCreator meth
...
+        return t;
+    }
+
+    // Verify the implementation class exists (generate it if necessary)
+    private DotName verifyGeneratedImplementationClass(ClassInfo interfaceInfo, List<String> fieldNames) {
+        DotName interfaceName = interfaceInfo.name();
+        DotName customImplName = customReturnTypes.get(interfaceName);
+
+        Map<String, FieldDescriptor> fields = new HashMap<>(fieldNames.size());
+        if (customImplName == null) {
```
gytis: Content of this block could be a private method e.g. `generateInterfaceImplementation(...)`

gytis: Same for further blocks: `generateMethod`, `generateField`, `generateConverter` etc.. I think it might improve readability.

extensions/spring-data-jpa/deployment/src/main/java/io/quarkus/spring/data/deployment/generate/GenerationUtil.java
gytis: This only seems to be used in one class, so could also stay as a private method there.

integration-tests/spring-data-jpa/src/main/java/io/quarkus/it/spring/data/jpa/BookRepository.java
```diff
@@ -22,4 +23,14 @@
     boolean existsBookByPublicationYearBetween(Integer start, Integer end);
 
     Optional<Book> findByPublicationYear(Integer year);
+
+    // issue 6205
+    @Query(value = "SELECT publicationYear as publicationYear, COUNT(*) as count FROM Book GROUP BY publicationYear")
```
gytis: I don't know how it's done in Spring Data, but why can't field name without alias be treated as is? Now `publicationYear as publicationYear` seems redundant.

extensions/spring-data-jpa/deployment/src/main/java/io/quarkus/spring/data/deployment/generate/CustomQueryMethodsAdder.java
ebullient: Woops! Yes. I had plans to use it..

extensions/spring-data-jpa/deployment/src/main/java/io/quarkus/spring/data/deployment/generate/CustomQueryMethodsAdder.java
ebullient: I can go to a foreach loop now, yes. Remember that thing where I was dependent on order? That's from then. ;) I'll fix

extensions/spring-data-jpa/deployment/src/main/java/io/quarkus/spring/data/deployment/generate/CustomQueryMethodsAdder.java
ebullient: I don't know that Spring supports it, I can test. I'll also see if the newInstance works, though I imagine we'll also miss the field ordering.

extensions/spring-data-jpa/deployment/src/main/java/io/quarkus/spring/data/deployment/generate/CustomQueryMethodsAdder.java
ebullient: I think I'd have to mark the query uniquely, and include that in a separate convert method. The order of the columns is significant, so while those have the same custom return type, it is probably safer to treat as two separate implementations. .. maybe. I will have to see if I still have the field order available when calling convert later.

integration-tests/spring-data-jpa/src/main/java/io/quarkus/it/spring/data/jpa/BookRepository.java
ebullient: I can try using the original name if it is otherwise missing. Will see where that gets me

geoand: Yeah I agree with @gytis . From an implementation perspective it shouldn't be any harder (except probably for a more complex regex)

extensions/spring-data-jpa/deployment/src/main/java/io/quarkus/spring/data/deployment/generate/GenerationUtil.java
ebullient: Moved to private methods re: other feedback from @gytis

integration-tests/spring-data-jpa/src/main/java/io/quarkus/it/spring/data/jpa/BookRepository.java
ebullient: agree

integration-tests/spring-data-jpa/src/main/resources/application.properties
```diff
@@ -8,3 +8,5 @@ quarkus.hibernate-orm.database.generation=drop-and-create
 #quarkus.hibernate-orm.log.sql=true
 
 %prod.quarkus.hibernate-orm.sql-load-script=import.sql
+
+quarkus.log.category.\"io.quarkus.spring.data\".level=TRACE
```
geoand: I doubt we wan't this :)

extensions/spring-data-jpa/deployment/src/main/java/io/quarkus/spring/data/deployment/generate/CustomQueryMethodsAdder.java
ebullient: I've changed it a bit to only look for getters .. Given the impl has to be generated, it can't do/won't build anything else. I will follow-up and ask what is allowed, now that I know what question to ask. ;)

integration-tests/spring-data-jpa/src/main/resources/application.properties
ebullient: woops!

extensions/spring-data-jpa/deployment/src/main/java/io/quarkus/spring/data/deployment/generate/CustomQueryMethodsAdder.java
geoand: Sounds good

extensions/spring-data-jpa/deployment/src/main/java/io/quarkus/spring/data/deployment/generate/CustomQueryMethodsAdder.java
ebullient: I throw an exception here. I could check to ensure the fields exist, but the code that attempts to convert from one type to the other depends on a method in the generated impl.

geoand: Looks good

integration-tests/spring-data-jpa/src/main/resources/application.properties
geoand: It seems to still be there, can you remove it please?

integration-tests/spring-data-jpa/src/main/java/io/quarkus/it/spring/data/jpa/BookResource.java
```diff
@@ -58,4 +59,33 @@ public Response findByPublicationYear(@PathParam("year") Integer year) {
         Optional<Book> book = bookRepository.findByPublicationYear(year);
         return book.map(b -> Response.ok(book).build()).orElse(Response.noContent().build());
     }
+
+    @GET
+    @Path("/count/year")
+    @Produces("application/json")
+    public List<BookRepository.BookCountByYear> countAllByPublicationYear() {
```
geoand: I am also thinking that the tests should include another custom type as well, just to be sure nothing weird gets tripped

integration-tests/spring-data-jpa/src/main/resources/application.properties
ebullient: Crap. I thought I had. I bet it got lost when I got frustrated with messy attempts and just reset things back to what I'd pushed already in a fit of pique.

integration-tests/spring-data-jpa/src/main/java/io/quarkus/it/spring/data/jpa/BookResource.java
ebullient: Will add another. Give me a few. Will use at least one different return type. I'm afraid my brain fails at anything useful for a Join query.

geoand: :+1:

geoand (approved):
Looks good! Thanks a lot!

geoand:
@ebullient can you please squash the last commit with the first one?

Thanks

geoand:
I force pushed a commit that should make CI happy - basically the `QuarkusUnitTest` became stricter while the PR was open

PR #6699, opened 2020-01-21
Enable JNI for logging-gelf

gsmet:


cescoffier (approved):

geoand:
Oh boy... I hope we don't end up with a mess because it wasn't in CI before...

geoand (approved):

gsmet:
@geoand I tested it before including it. I think @loicmathieu probably tested it when master was 19.3 with JNI enabled all the time.

geoand:
Yeah, I figured you tested it and that was the reason why you created the PR. 
But still we have seen strange things happen :)

PR #6700, opened 2020-01-21
Set a higher timeout for Vert.x tests

gsmet:
Clément can't help adding more and more.


cescoffier:
@gsmet not my fault, @paulrobinson asked :-)... But yes, I should have increased the time out. Another round of test will come next week.

cescoffier (approved):

gsmet:
CI failed for a totally unrelated reason. I think it's safe enough to be merged, let's avoid overloading CI with this.

PR #6703, opened 2020-01-21
Hibernate ORM should not resolve placeholders in configuration properties

Sanne:
…ties

This feature is mimicking the configuration parsing of upstream Hibernate ORM, but I feel it's redundant in Quarkus as we already have a layer for such replacements (right?).

So let's shave a little more from bootime :)

Also, reasons:

- I prefer to remove such things now than later (easier to add back if we change our minds)
- Consistency with the Quarkus configuration


geoand:
@Sanne can you show an example of what that feature did? 

I am pretty sure I understand (and if I do then yes we have that in Quarkus already), but I just want to be on the safe side

Sanne:
@geoand it allowed to refer to _environment variables_ within the value of configuration properties.

e.g.

    hibernate.config.x= somethin_${VARX}

then run the JVM with `-DVARX=whatever`

It's not something we should support as Quarkus already has similar capabilities.

geoand:
@Sanne thanks!

Yeah quarkus config already has all we need here.

geoand:
I would approve but I don't know much of anything about the Hibernate extension so better to leave it to someone else :)

gsmet:
Rebased to trigger CI.

gsmet (approved):

Issue #6678, opened 2020-01-21
Optimize Vert.x startup by avoiding DNS resolver initialization


gsmet:
If we do something about this, it should be done in Vert.x but I'd like the discussion to start here to be sure it's really something that should be fixed in Vert.x or if it's something we should live with.

When starting Vert.x in `io/vertx/core/http/impl/HttpServerImpl.listen`, we resolve the address in `io/vertx/core/net/impl/AsyncResolveConnectHelper.doBind` via `io/vertx/core/impl/VertxImpl.resolveAddress`.

Most of the time (if not all), we will pass an IPv4/IPv6 address and I'm not sure we really need to resolve the name? Especially when using the default `0.0.0.0`.

From what I can see, with the default `0.0.0.0`, we set up the whole Vert.x DNS infrastructure at that point.

I have a few questions:
- do we need to do that for an IP address? It seems useless?
- if we don't, will the DNS infrastructure be set up later? What are the operations requiring address resolution? Will a RESTEasy request require DNS resolution?

Basically, is the effort to drop this worth it?

Note: I did some bootstrap profiling as my demo started to lag a lot in dev mode.

/cc @vietj @cescoffier @stuartwdouglas

vietj:
the code will use a quick path in the resolver, 0.0.0.0 will be resolved quickly to an `InetSocketAddress` representing this address and not actual DNS resolution will happen (i.e no network call). Then this address will be used to do a Netty bind on it.

Not sure what is the issue here, unless you want to add a specific fast path that resolves hardcode the resolution of the string `0.0.0.0` to the `InetSocketAddress` and not call Netty.

gsmet:
@vietj so... most of the bootstrap slowness is related to class loading. It's what really slows us down.

In this case, it's mostly related to the initialization of `NioDatagramChannel` and allegates. AFAICS this class is only called when initializing the DnsResolver layer (Eclipse says so but it might not be true, please correct me if incorrect).

If we don't resolve any other address later on, that would be quite beneficial to avoid loading all this machinery.

Now, I'm not sure if we do or don't. You're the expert here :).

vietj:
have you tried in Quarkus native to turn JDK DNS resolver using `vertx.disableDnsResolver` system property, this should use the JDK dns resolver which will bypass this machinery ?

gsmet:
I'm working on optimizing the dev mode experience so in JVM mode. Is your proposal valid for JVM mode, too?

[...]

PR #6706, opened 2020-01-21
Disable the DNS resolver when creating the options

gsmet:
So that it's properly disabled even if only the web Vert.x instance is
created.

Related to https://github.com/quarkusio/quarkus/issues/6678

/cc @cescoffier


cescoffier (approved):

gsmet:
CI failed again with a git error after a fresh rebase. Will open a new PR to see if things are better.

lordvlad:
@gsmet @cescoffier 

Three years later, is it still warranted to disable async dns by default for quarkus?

Just spent a good while to figure out why we were getting loads of `thread blocked` warnings, with a little help from @vietj I found that async dns is disabled for quarkus by default.

cescoffier:
Yes, the async dns client is still causing many issues.

cescoffier:
We have issues when resolving services in some Kubernetes environmens.

I also had issues in docker-compose but it may have been fixed. 

Recently, I have seen an issue in AWS Lambda.

So the safest for no is to not use the async dns resolver, and make sure it works in your environment before enabling it.

PR #6718, opened 2020-01-22
Add first take on native image CI in cron

geoand:


geoand:
@gsmet here is what we talked about

.github/workflows/native-cron-build.yml
```diff
@@ -0,0 +1,64 @@
...
+          echo "The report step got status: ${STATUS}"
+          sudo apt-get update -o Dir::Etc::sourcelist="sources.list" \
+            -o Dir::Etc::sourceparts="-" -o APT::Get::List-Cleanup="0"
+          sudo apt-get install -y gnupg2 gnupg-agent
+          echo Installing SDKMAN
+          curl -s "https://get.sdkman.io" | bash
+          source ~/.sdkman/bin/sdkman-init.sh && \
+          sdk install kotlin 1.3.61 && \
+          sdk install kscript 2.9.0
+          kscript .github/NativeBuildReport.kts ${GITHUB_TOKEN} ${STATUS}
```
gsmet: We would need the Java version in the report, wouldn't we?

gsmet (changes requested):
Added a small question.

.github/NativeBuildReport.kts
```diff
@@ -0,0 +1,104 @@
...
+
+import org.kohsuke.github.*
+import java.io.File
+import java.io.IOException
+import java.util.concurrent.TimeUnit
+
+val token = args[0]; 
+val status = args[1];
+
+val ISSUE_NUMBER=6717
```
gsmet: I would make the issue number a parameter so that we can easily reuse this script for other reports.

.github/workflows/native-cron-build.yml
geoand: Very good point. We actually will probably need two issues. One for Java 8 and one for Java 11 since we need to open/closing of the issues to happen on each specific one.

WDYT?

geoand: The point should be addressed now. It's not the most elegant syntax in the world, but it seems like github actions doesn't give us many options

.github/NativeBuildReport.kts
geoand: Done

gsmet (approved):
I would say let's push it ASAP and see how it goes :).

geoand:
Yeah, that makes perfect IMHO

gsmet:
@geoand I'll push it later tonight once all the other PRs have been run on CI.

geoand:
👌

PR #4529, opened 2019-10-12
3343: add multi mongo db support

dufoli:
fix #3343

GOAL:
```
quarkus.mongodb.cluster1.connection-string = mongodb://mongo1:27017,mongo2:27017
quarkus.mongodb.cluster2.connection-string = mongodb://mongo1:27017,mongo2:27017
```
And then use annotations above the instances of MongoClient to indicate which instance we are going to use (example imaginary @Connection annotation used):
```

@Connection("cluster1")
@Inject
ReactiveMongoClient mongoClientCluster1
```

I reuse(copy) a lot of code from agroal extention to have same structure. when I do not set codecs, it work well but codec test is failing. I need help to understand how to load codec during runtime and not deployment. 

previous code: load mongoClient during runtime build step
new code: create producer for each named connection (like agroal extention) and buildstep during compile time. 

depend of #5387. Have to wait that it land on master

PR #6720, opened 2020-01-22
Simplify Mongo client

geoand:
MongoClientBuildItem is not used anywhere neither
are the RuntimeValue values produced by the recorder


extensions/mongodb-client/deployment/src/main/java/io/quarkus/mongodb/deployment/MongoClientProcessor.java
```diff
@@ -63,11 +59,9 @@ MongoClientBuildItem build(BuildProducer<FeatureBuildItem> feature, MongoClientR
 
         sslNativeSupport.produce(new ExtensionSslNativeSupportBuildItem(FeatureBuildItem.MONGODB_CLIENT));
 
-        RuntimeValue<MongoClient> client = recorder.configureTheClient(config, beanContainer.getValue(),
+        recorder.configureTheClient(config, beanContainer.getValue(),
                 launchMode.getLaunchMode(), shutdown,
                 codecs.getCodecProviderClassNames());
-        RuntimeValue<ReactiveMongoClient> reactiveClient = recorder.configureTheReactiveClient();
-        return new MongoClientBuildItem(client, reactiveClient);
```
cescoffier: That allows another extension to get the client. Why removing this feature? 
Not sure if Panache Mongo requires it But I had something using this (for demo purpose, so no problem to break it).

geoand: Mongo Panache is not using this (at least as far as I could tell and when I ran the tests locally nothing failed).

I removed it because nothing was using it and the implementation threw me off. If you feel it's valid to keep it around for future use, then I have absolutely no problem closing this PR

loicmathieu:
MongoDB with Panache didn't use it but maybe it should to be sure that the client is initialized before it. For the record, HIbernate with Panache uses the `HibernateEnhancersRegisteredBuildItem` as a marker build item.

This is also changed by the PR that provides multiple MongoDB Client support : #4529 ... better not merging this before it as it would need again a rebased ...

geoand:
Yeah, let's see what #4529 does and I'll update this PR accordingly (or close it)

gsmet:
I had a request to expose the `ValidatorFactory` like that for Camel. So I think this is indeed useful and should probably be used by Camel.

geoand:
I'll check with Camel as well. If it's not used, then I don't really see the point of keeping it around

gsmet:
Well, it might be a good idea to check if they would be interested and if we should generalize things like that instead of removing them.

geoand:
That's a good idea. I'll see what's up on that side of things

cescoffier:
This pattern lets us manage the "I'm using Vert.x before you" issue we had when injecting Vert.x instances or the router instances. So, I believe it should be generalized. It does not cost much and allows extensions to be called once their dependencies are ready.

geoand:
Closing this as Camel uses it

PR #6721, opened 2020-01-22
Reimplement RuntimeBeanProcessor to register synthetic beans

mkouba:
- previously for each RuntimeBeanBuildItem a producer method was added
to a generated RuntimeBeanProducers class
- the disadvantage of the previous approach was that
RuntimeBeanBuildItems were processed before the bean archive index was
produced, ie. phase 0 of the bean processing
- also fix an issue in the type-safe resolution - matching qualifier
array values


mkouba:
I think that we should rename the `RuntimeBeanBuildItem` to `SyntheticBeanBuildItem` and adjust the API a little bit. But I'm quite not sure what the current rules for backward compatibility are. For example we could deprecate the `RuntimeBeanBuildItem` add `SyntheticBeanBuildItem` in 1.3, and remove `RuntimeBeanBuildItem` in 1.4.

gsmet:
Force pushed a rebase to trigger CI.

mkouba:
@stuartwdouglas I can alo incorporate your changes from https://github.com/quarkusio/quarkus/pull/5938.

gsmet:
@mkouba there is a formatting issue

stuartwdouglas (approved):

manovotn (approved):

manovotn:
> I think that we should rename the RuntimeBeanBuildItem to SyntheticBeanBuildItem and adjust the API a little bit. 

I agree, `SyntheticBean` is a known concept in CDI whereas `RuntimeBean` sounds a bit confusing. The gradual changewith deprecation is IMO a good way to go.

PR #6727, opened 2020-01-22
Make optional the oauth2 configuration when the extension is disabled

loicmathieu:


extensions/elytron-security-oauth2/runtime/src/main/java/io/quarkus/elytron/security/oauth2/runtime/OAuth2Recorder.java
```diff
@@ -29,13 +30,17 @@
 
 @Recorder
 public class OAuth2Recorder {
+    private static final Logger LOG = Logger.getLogger(OAuth2Recorder.class);
 
     public RuntimeValue<SecurityRealm> createRealm(OAuth2Config config)
             throws IOException, NoSuchAlgorithmException, CertificateException, KeyStoreException, KeyManagementException {
+        if (!config.clientId.isPresent() || !config.clientSecret.isPresent() || !config.introspectionUrl.isPresent()) {
+            LOG.error("client-id, client-secret and introspection-url must be configured when the oauth2 extension is enabled");
```
machi1990: should it be a `ConfigurationException` instead?

loicmathieu: It will be a `NoSuchElementException` with the current impl.
Maybe a `ConfigurationException` is better yes :)

gsmet: Yes we need it to be an exception. +1 for `ConfigurationException`.

loicmathieu: done

gsmet (approved):

gsmet:
I force pushed a rebase to trigger CI.

machi1990 (approved):

PR #6730, opened 2020-01-22
Add value to PathParam

pugmarx:
PathParam needs value to be specified. Added that.

### DIFF
--- a/docs/src/main/asciidoc/getting-started.adoc
+++ b/docs/src/main/asciidoc/getting-started.adoc
@@ -281,7 +281,7 @@ public class GreetingResource {
     @GET
     @Produces(MediaType.TEXT_PLAIN)
     @Path("/greeting/{name}")
-    public String greeting(@PathParam String name) {
+    public String greeting(@PathParam("name") String name) {
         return service.greeting(name);
     }
 


docs/src/main/asciidoc/getting-started.adoc
```diff
@@ -281,7 +281,7 @@ public class GreetingResource {
     @GET
     @Produces(MediaType.TEXT_PLAIN)
     @Path("/greeting/{name}")
-    public String greeting(@PathParam String name) {
+    public String greeting(@PathParam("name") String name) {
```
gsmet: It does not if you use the right import. You should use `org.jboss.resteasy.annotations.jaxrs.PathParam` not the JAX-RS one.

gsmet (changes requested):
I don't think this one is valid. See my comment.

docs/src/main/asciidoc/getting-started.adoc
pugmarx: Ah! Ok. I think I got confused since rest of the annotations come from `javax.ws.rs.*`

gsmet: Yeah, that's why we put the imports in the source excerpts otherwise it's impossible to get it right :).

Issue #5809, opened 2019-11-27
Keycloak with Reactive Routes Error: Request has already been read


tpenakov:
I've tried to run a quarkus with keyckoak policy enforcer and reactive routes (no resteasy at all). It fails with the following error:
`ERROR: HTTP Request to /fhir/test failed, error id: f92ad12f-7a15-4fc4-b1cd-7d99ab09df0d-1
java.lang.IllegalStateException: Request has already been read`
`	at io.vertx.core.http.impl.HttpServerRequestImpl.checkEnded(HttpServerRequestImpl.java:600)`
`	at io.vertx.core.http.impl.HttpServerRequestImpl.handler(HttpServerRequestImpl.java:307)`

This is my config:

`quarkus.keycloak.policy-enforcer.enable=true`
`quarkus.keycloak.policy-enforcer.enforcement-mode=PERMISSIVE`
`quarkus.keycloak.policy-enforcer.path-cache.lifespan=36000000`
`quarkus.keycloak.policy-enforcer.paths.0.path=${quarkus.http.root-path}/test`
`quarkus.keycloak.policy-enforcer.paths.0.enforcement-mode=DISABLED`
`quarkus.keycloak.policy-enforcer.paths.999.path=/*`
`quarkus.keycloak.policy-enforcer.paths.999.enforcement-mode=ENFORCING`

The paths where it fails is the path disabled for policy enforcing. I guess the same error will occur and for secured paths

Sample project : [keycloak-with-reactive-routes-error](https://github.com/tpenakov/keycloak-with-reactive-routes-error)

gsmet:
@tpenakov any chance you could put together a small reproducer?

tpenakov:
No problem - added in bug description.

sberyozkin:
CC @pedroigor @stianst

sberyozkin:
@tpenakov Can you please confirm that disabling the policy enforcer 'fixes' it ?

tpenakov:
Hi @sberyozkin ,
You mean this:
`quarkus.keycloak.policy-enforcer.enable=false`
Right?
Now I've got NPE:
`Nov 27, 2019 6:13:40 PM io.quarkus.vertx.http.runtime.QuarkusErrorHandler handle`
`ERROR: HTTP Request to /fhir/test failed, error id: d7186528-0b82-4a0f-91ae-fe2f731f2edd-1`
`java.lang.NullPointerException`
[...]

Issue #4922, opened 2019-10-28
Handle forwarded headers


FroMage:
When sitting behind a forwarding proxy (that has always been the case for my apps at least) we cannot trust the underlying TCP connection for the remote IP and current port, nor the HTTP `Host` header, so the information we expose to users is wrong. The impact is they cannot use JAX-RS `UriInfo` to build absolute URIs nor `HttpResponse.getRemoteAddress()` to log security violations, nor can they know if the current connection is encrypted.

We should read the https://tools.ietf.org/html/rfc7239#section-4 `Forwarded` header and the older equivalent headers to make sure the information we expose to users is accurate.

I would be tempted to say this is of the highest priority, but the fact that nobody noticed this before probably means deployments these days either don't use forwarding proxies, or mitigate their effects using alternative strategies.

FroMage:
As noticed in https://github.com/quarkusio/quarkus/pull/4907#issuecomment-546870817

stuartwdouglas:
I think we should be able to just stick this in the RoutingContext under some well key, and then RESTEasy and Undertow can be modified to look at this key for the remote address. 

All this code existing in Undertow already, so it should be pretty simple.

lodrantl:
This setting work with vertx http, but as soon as quarkus-undertow is added it breaks. Maybe just the ForwardedHandler should be added to the deployment?

stuartwdouglas:
This is a bug that will be fixed by https://github.com/quarkusio/quarkus-http/pull/20

For now you can use the forwarded handler as a workaround.

lodrantl:
I tried with the master version of quarkus-http and nothing changes. X-Forwarded-Proto and X-Forwarded-Port are used correctly, but X-Forwarded-Host is just disregarded completely. Tracked it down to getHostandPort() method in HttpServerExchange.java.
https://github.com/undertow-io/undertow/blob/master/core/src/main/java/io/undertow/server/HttpServerExchange.java#L646

The ForwardedHandler from Undertow actually overwrites the Host field with the X-Forwarded-Host internally. 
https://github.com/undertow-io/undertow/blob/master/core/src/main/java/io/undertow/server/handlers/ForwardedHandler.java#L55

PR #6733, opened 2020-01-22
Add support for proxy address forwarding

stuartwdouglas:
This also now includes the fix for #5809 as they would conflict.


extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/HttpConfiguration.java
```diff
@@ -49,6 +49,20 @@
     @ConfigItem(defaultValue = "8444")
     public int testSslPort;
 
+    /**
+     * If this is true then the address, scheme etc will be set from headers forwarded by the proxy server, such as
+     * {@code X-Forwarded-For}. This should only be set if you are behind a proxy that sets these headers.
+     */
+    @ConfigItem(defaultValue = "false")
+    public boolean proxyAddressForwarding;
```
FroMage: Why isn't it the default to use those values if present? Security reasons because any user could use them?

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/HttpConfiguration.java
```diff
@@ -49,6 +49,20 @@
...
+
+    /**
+     * If this is true and proxy address forwarding is enabled then the standard {@code Forwarded} header will be used,
+     * rather than the more common but not standard {@code X-Forwarded-For}.
+     */
+    @ConfigItem(defaultValue = "false")
+    public boolean allowForwarded;
```
FroMage: Why not use `Forwarded` if present (since it's the new preferred header) and fall-back to `X-Forwarded` otherwise? Why do we need a config?

FroMage:
Original issue is #4922

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/HttpConfiguration.java
stuartwdouglas: Yea, any client can add them if this is not configured and spoof their address etc.

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/HttpConfiguration.java
stuartwdouglas: Because an old proxy that is using X-Forwarded likely does not understand Forwarded, and will just pass it through to the backend, allowing any remote client to spoof their address. The code I copied this from actually does this, and I think it is wrong: https://github.com/vert-x3/vertx-web/issues/1519

FroMage: Oh damn, that's subtle. OK.

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/HttpConfiguration.java
FroMage: OK thanks.

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/ForwardedServerRequestWrapper.java
```diff
@@ -0,0 +1,314 @@
...
+import javax.security.cert.X509Certificate;
+
+import io.vertx.core.Handler;
+import io.vertx.core.MultiMap;
+import io.vertx.core.buffer.Buffer;
+import io.vertx.core.http.*;
+import io.vertx.core.net.NetSocket;
+import io.vertx.core.net.SocketAddress;
+
+class ForwardedServerRequestWrapper implements HttpServerRequest {
```
sberyozkin: @stuartwdouglas Hi Stuart, can it be of some help if the request info being replaced with this wrapper be contained as `HttpServerRequest` attribute ? So that it can be logged if needed etc ? Not really sure, just checking.
All looks good

stuartwdouglas: I don't really know if it is needed, generally you know what your proxies are. Maybe wait and see if anyone asks for it?

sberyozkin: @stuartwdouglas sorry, I keep missing the comments from Github...yes, makes sense

sberyozkin (approved):

sberyozkin:
Kotlin integration tests are failing, seems unrelated

sberyozkin:
@stuartwdouglas Can you please rebase ?

gsmet:
I think it's good to go. The failure is unrelated.

PR #6735, opened 2020-01-23
Micrometer compatibility mode for Metrics

### LINKED ISSUE #6544: Micrometer compatibility mode for Metrics
Users migrating from Spring Boot and Micrometer might want to expose the same JVM metrics as they used to, so that they don't have to rework their Grafana dashboards. 
The naming and organization of base/vendor metrics exposed by Quarkus is different than what Micrometer does, so it would be useful to have a configuration option that changes them to align with that scheme. In particular, we should mimic the way how `jvm` metrics are handled in https://github.com/micrometer-metrics/micrometer/tree/master/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jvm so we will have to add some new metrics that we don't normally have, and change metadata of some metrics that we do normally have

jmartisk:
Fixes #6544

### DIFF
--- a/extensions/smallrye-metrics/deployment/src/main/java/io/quarkus/smallrye/metrics/deployment/SmallRyeMetricsProcessor.java
+++ b/extensions/smallrye-metrics/deployment/src/main/java/io/quarkus/smallrye/metrics/deployment/SmallRyeMetricsProcessor.java
@@ -103,6 +103,15 @@ public class SmallRyeMetricsProcessor {
         @ConfigItem(name = "extensions.enabled", defaultValue = "true")
         public boolean extensionsEnabled;
 
+        /**
+         * Apply Micrometer compatibility mode, where instead of regular 'base' and 'vendor' metrics,
+         * Quarkus exposes the same 'jvm' metrics that Micrometer does. Application metrics are unaffected by this mode.
+         * The use case is to facilitate migration from Micrometer-based metrics, because original dashboards for JVM metrics
+         * will continue working without having to rewrite them.
+         */
+        @ConfigItem(name = "micrometer.compatibility", defaultValue = "false")
+        public boolean micrometerCompatibility;
+
     }
 
     SmallRyeMetricsConfig metrics;
@@ -249,9 +258,15 @@ public class SmallRyeMetricsProcessor {
 
     @BuildStep
     @Record(RUNTIME_INIT)
-    void registerBaseAndVendorMetrics(SmallRyeMetricsRecorder metrics, ShutdownContextBuildItem shutdown) {
-        metrics.registerBaseMetrics(shutdown);
-        metrics.registerVendorMetrics(shutdown);
+    void registerBaseAndVendorMetrics(SmallRyeMetricsRecorder metrics,
+            ShutdownContextBuildItem shutdown,
+            SmallRyeMetricsConfig config) {
+        if (config.micrometerCompatibility) {
+            metrics.registerMicrometerJvmMetrics(shutdown);
+        } else {
+            metrics.registerBaseMetrics(shutdown);
+            metrics.registerVendorMetrics(shutdown);
+        }
     }
 
     @BuildStep
--- a/extensions/smallrye-metrics/runtime/src/main/java/io/quarkus/smallrye/metrics/runtime/SmallRyeMetricsRecorder.java
+++ b/extensions/smallrye-metrics/runtime/src/main/java/io/quarkus/smallrye/metrics/runtime/SmallRyeMetricsRecorder.java
@@ -1,10 +1,12 @@
 package io.quarkus.smallrye.metrics.runtime;
 
+import java.lang.management.BufferPoolMXBean;
 import java.lang.management.ClassLoadingMXBean;
 import java.lang.management.GarbageCollectorMXBean;
 import java.lang.management.ManagementFactory;
 import java.lang.management.MemoryMXBean;
 import java.lang.management.MemoryPoolMXBean;
+import java.lang.management.MemoryType;
 import java.lang.management.OperatingSystemMXBean;
 import java.lang.management.RuntimeMXBean;
 import java.lang.management.ThreadMXBean;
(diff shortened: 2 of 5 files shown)


geoand:
@ebullient I know you were interested in this :)

ebullient:
I am, indeed. I'll likely bash on the boundaries of this one some more for my DevNexus talk.. :-P Thanks, Jan!

gsmet:
@jmartisk could you ping the original user asking for that so that he gives it a try?

jmartisk:
@gsmet Sure, I asked in the mailing list. Let's see.
For the record I tested it against this dashboard https://grafana.com/grafana/dashboards/3308 (this dashboard uses a bit more metrics than just `jvm` ones, so far this PR contains just `jvm` metrics, we could potentially add a few more if there is enough interest)

wedamartinez:
@jmartisk This is Alejandro, thank you very much for your work, to you and your team!  It looks awesome. I tested and I have almost all that the jvm metrics that we had with micrometer.

wedamartinez:
@jmartisk I have a question if I have this parameter activated, I will be able to add customized "application" metrics? for the moment we don't have any Quarkus microservice using this feature but for sure we are going to use that soon. I am going to create a POC to test that use case.

ebullient:
@wedamartinez .. I am on that path as well, as I have a talk that uses micrometer's custom metrics, and I would like those to be usable w/o change. So we'll see how that goes. ;)

wedamartinez:
Cool also works! thanks!

# HELP application_com_weekendesk_activities_activitytype_http_ActivityTypeResource_performedChecks_total How many primality checks have been performed.
# TYPE application_com_weekendesk_activities_activitytype_http_ActivityTypeResource_performedChecks_total counter
application_com_weekendesk_activities_activitytype_http_ActivityTypeResource_performedChecks_total 5.0

extensions/smallrye-metrics/deployment/src/main/java/io/quarkus/smallrye/metrics/deployment/SmallRyeMetricsProcessor.java
```diff
@@ -103,6 +103,15 @@
         @ConfigItem(name = "extensions.enabled", defaultValue = "true")
         public boolean extensionsEnabled;
 
+        /**
+         * Apply Micrometer compatibility mode, where instead of regular 'base' and 'vendor' metrics,
+         * Quarkus exposes the same 'jvm' metrics that Micrometer does. Application metrics are unaffected by this mode.
```
wedamartinez: It's seems that they are affected

extensions/smallrye-metrics/deployment/src/main/java/io/quarkus/smallrye/metrics/deployment/SmallRyeMetricsProcessor.java
```diff
@@ -103,6 +103,15 @@
...
+         * The use case is to facilitate migration from Micrometer-based metrics, because original dashboards for JVM metrics
+         * will continue working without having to rewrite them.
```
wedamartinez: thanks 👍

extensions/smallrye-metrics/deployment/src/main/java/io/quarkus/smallrye/metrics/deployment/SmallRyeMetricsProcessor.java
jmartisk: I don't know what you mean here. Is it about the `application_` prefix being there? That's the way it is normally (without the compatibility mode)

jmartisk:
> @jmartisk I have a question if I have this property activated, I will be able to add customized "application" metrics? for the moment we don't have any Quarkus microservice using any custom metric but for sure we are going to use that soon. I am going to create a POC to test that use case.

@wedamartinez Well that's currently out of scope, that would be a tough one. Micrometer uses a completely different programming API (annotations etc.) than Quarkus (which is based on MicroProfile Metrics), has different metric types, different strategy for exporting the metrics, etc. So what you're asking about sounds more like we would need to introduce a completely new alternative extension that supports Micrometer metrics. I can't quite imagine supporting the Micrometer API and somehow automatically 'bridging' all of it into MicroProfile Metrics without turning it into a complete mess.
The scope of this enhancement is to at least map the basic (built-in) metrics. Application-specific metrics will have to be rewritten manually to use the MP Metrics API.

wedamartinez:
> > @jmartisk I have a question if I have this property activated, I will be able to add customized "application" metrics? for the moment we don't have any Quarkus microservice using any custom metric but for sure we are going to use that soon. I am going to create a POC to test that use case.
> 
> @wedamartinez Well that's currently out of scope, that would be a tough one. Micrometer uses a completely different programming API (annotations etc.) than Quarkus (which is based on MicroProfile Metrics), has different metric types, different strategy for exporting the metrics, etc. So what you're asking about sounds more like we would need to introduce a completely new alternative extension that supports Micrometer metrics. I can't quite imagine supporting the Micrometer API and somehow automatically 'bridging' all of it into MicroProfile Metrics without turning it into a complete mess.
> The scope of this enhancement is to at least map the basic (built-in) metrics. Application-specific metrics will have to be rewritten manually to use the MP Metrics API.

Thank you very much!, I completely agree.

jmartisk:
To be more clear, my previous comment was to say that it's not supported to use an application that includes application metrics through Micrometer API without rewriting them to MP Metrics. The ability to use MP Metrics API is not affected if this mode is active.

@gsmet @geoand It looks to me this is ready, do you guys have any concerns?

extensions/smallrye-metrics/runtime/src/main/java/io/quarkus/smallrye/metrics/runtime/SmallRyeMetricsRecorder.java
```diff
@@ -583,4 +604,193 @@ private void memoryPoolMetrics(MetricRegistry registry, List<String> names) {
         }
     }
 
+    private void micrometerJvmGcMetrics(MetricRegistry registry, List<String> names, ShutdownContext shutdown) {
+        if (!ImageInfo.inImageCode()) {
```
geoand: I was just about to ask how the code knows to not expose metrics in GraalVM mode and stumbled upon this before I could :).

:+1:

geoand (approved):
I don't know a ton about metrics, but this seems reasonable to me

gastaldi:
Works great! Thanks!

wedamartinez:
Thanks! god job!

PR #6750, opened 2020-01-23
Allows empty Sort.by

### LINKED ISSUE #6664: Support for empty Sort object
**Description**

Support an empty Sort object to denote "no-sorting needed".

Currently it uses `null`, but I use the Sort object to validate the sorting passed is valid (i.e. that the app allows to sort by that column). 
I can of course validate if the sort is null before trying, but would be more conveniente to just pass an "empty sort".

I tried using Sort.by() to create an empty Sort, but this line https://github.com/quarkusio/quarkus/blob/master/extensions/panache/hibernate-orm-panache/runtime/src/main/java/io/quarkus/hibernate/orm/panache/runtime/JpaOperations.java#L209 produces an incomplete "  SORT BY " that crashes the query (`unexpected token: null`)

**Implementation ideas**

I think that the only needed thing to do is validate [here](https://github.com/quarkusio/quarkus/blob/master/extensions/panache/hibernate-orm-panache/runtime/src/main/java/io/quarkus/hibernate/orm/panache/runtime/JpaOperations.java#L209) that sort.getColumns().size() > 0 and should be fine as far I have seen.
If this is possible, i could send the PR.

josejulio:
- When Sort.by without arguments is used, there won't be any
   Sort by clause in the query

Closes #6664

### DIFF
--- a/extensions/panache/hibernate-orm-panache/deployment/src/test/java/io/quarkus/hibernate/orm/panache/test/JpaOperationsSortTest.java
+++ b/extensions/panache/hibernate-orm-panache/deployment/src/test/java/io/quarkus/hibernate/orm/panache/test/JpaOperationsSortTest.java
@@ -1,0 +1,24 @@
+package io.quarkus.hibernate.orm.panache.test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import org.junit.jupiter.api.Test;
+
+import io.quarkus.hibernate.orm.panache.runtime.JpaOperations;
+import io.quarkus.panache.common.Sort;
+
+public class JpaOperationsSortTest {
+
+    @Test
+    public void testSortBy() {
+        Sort sort = Sort.by("foo", "bar");
+        assertEquals(" ORDER BY foo , bar", JpaOperations.toOrderBy(sort));
+    }
+
+    @Test
+    public void testEmptySortByYieldsEmptyString() {
+        Sort emptySort = Sort.by();
+        assertEquals("", JpaOperations.toOrderBy(emptySort));
+    }
+
+}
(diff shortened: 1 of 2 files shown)


josejulio:
I haven't been able to test this code, I'm figuring out how to build and test.

gastaldi:
It would be nice to have a small test for this change

josejulio:
> It would be nice to have a small test for this change

Sure, I see that there are no test for this module (please point me if I'm wrong here). If so, i might try to create some tests that cover the change.

gastaldi:
There are some tests in the `deployment` module and integration tests in https://github.com/quarkusio/quarkus/tree/master/integration-tests/hibernate-orm-panache.

I think in your case you can add a simple unit test in the `deployment` module, next to the existing ones :)

josejulio:
> There are some tests in the `deployment` module and integration tests in https://github.com/quarkusio/quarkus/tree/master/integration-tests/hibernate-orm-panache.
> 
> I think in your case you can add a simple unit test in the `deployment` module, next to the existing ones :)

Great, will do, thanks!

josejulio:
Done, added a couple of tests.

gastaldi (commented):
Nice! Can you squash your commits into a single one? Thanks!

josejulio:
> Nice! Can you squash your commits into a single one? Thanks!

Done

FroMage (approved):
LGTM, thanks!

FroMage:
@loicmathieu do you have a similar bug in mongodb as well?

loicmathieu:
@FroMage in MongoDB with Panache empty sort will lead to empty sort Document, this is useless but it's legit so not bug: https://github.com/quarkusio/quarkus/blob/master/extensions/panache/mongodb-panache/runtime/src/main/java/io/quarkus/mongodb/panache/runtime/MongoOperations.java#L461

FroMage:
OK perfect. Thanks

PR #6752, opened 2020-01-23
Generate Tika XML configuration during build time

### LINKED ISSUE #5700: TikaProcessorTest should use Quarkus test framework
**Describe the bug**
In order to get #5387 to pass, I had to add code to `TikaProcessorTest` to handle setup and tear down of the configuration.  But it would probably be better to let the Quarkus test framework code do that instead.

### LINKED ISSUE #6746: Tika extension doesn't like the new class loading
@machi1990 stumbled upon an issue while working on improving a test: `java.lang.ClassCastException: io.quarkus.tika.runtime.TikaParserParameter cannot be cast to io.quarkus.tika.runtime.TikaParserParameter`.

The gist of it is that there is a class loader issue in `TikaProcessor`.

I think the issue is that ``TikaParserParameter``s are created during the deployment and pushed to `TikaParsersConfigBuildItem`.

```java
        Map<String, List<TikaParserParameter>> parsersConfig = getSupportedParserConfig(config.tikaConfigPath, config.parsers,
                config.parserOptions, config.parser);
        recorder.initTikaParser(beanContainer.getValue(), config, parsersConfig);
        return new TikaParsersConfigBuildItem(parsersConfig);
```

The config should be passed to the recorder and the TikaParserParameters created there, not at build time.

This should be integrated in the other BuildStep and use the config too:
```java
    @BuildStep
    public void registerTikaProviders(BuildProducer<ServiceProviderBuildItem> serviceProvider,
            TikaParsersConfigBuildItem parserConfigItem) throws Exception {
        serviceProvider.produce(
                new ServiceProviderBuildItem(Parser.class.getName(),
                        new ArrayList<>(parserConfigItem.getConfiguration().keySet())));
        serviceProvider.produce(
                new ServiceProviderBuildItem(Detector.class.getName(), getProviderNames(Detector.class.getName())));
        serviceProvider.produce(
                new ServiceProviderBuildItem(EncodingDetector.class.getName(),
                        getProviderNames(EncodingDetector.class.getName())));
    }
```
so that we can get rid of this `BuildItem`.

machi1990:
test: TikaProcessorTest should use Quarkus test framework

Fixes https://github.com/quarkusio/quarkus/issues/6746
Fixes https://github.com/quarkusio/quarkus/issues/5700


extensions/tika/deployment/src/main/java/io/quarkus/tika/deployment/TikaProcessor.java
```diff
@@ -91,42 +79,48 @@ public void registerRuntimeInitializedClasses(BuildProducer<RuntimeInitializedCl
...
-                new ServiceProviderBuildItem(EncodingDetector.class.getName(),
-                        getProviderNames(EncodingDetector.class.getName())));
-    }
-
-    static List<String> getProviderNames(String serviceProviderName) throws Exception {
+    @Record(ExecutionTime.STATIC_INIT)
+    void initializeTikaParser(BeanContainerBuildItem beanContainer, TikaRecorder recorder,
+            BuildProducer<ServiceProviderBuildItem> serviceProvider, TikaConfiguration configuration)
+            throws Exception {
+        Map<String, List<TikaParserParameter>> parsers = getSupportedParserConfig(configuration.tikaConfigPath,
```
sberyozkin: @machi1990 first of all, thanks for this effort :-)
I'm not sure that removing a build item containing this Map. My plan, with the current master, has been to make this Map visible to various steps dealing with the individual parsers. For example, I suggested to @tpenakov in the issue where he works on OOXML support to group all PDF related build steps into a single method like `preparePdfParser` and check this map if PDF parser key is available and only then do all those PDF specific native registrations, the same for OOXML.

So I think that build item still has it be retained. Let me know please if it makes sense. CC @gsmet

machi1990: > @machi1990 first of all, thanks for this effort :-)

Hi @sberyozkin thank you. 

> I'm not sure about removing a build item containing this Map. My plan, with the current master, has been to make this Map visible to various steps dealing with the individual parsers. For example, I suggested to @tpenakov in the issue where he works on OOXML support to group all PDF related build steps into a single method like `preparePdfParser` and check this map if PDF parser key is available and only then do all those PDF specific native registrations, the same for OOXML.

Oh sorry, I was not aware of the future plans around this class. 

If I understood the description above correctly, it seems to be that these various methods (e.g `preparePdfParser`) will be consuming this build item from within the `TikaProcessor` class, correct? 

If so, the same be achieved with an invocation of a `private` method passing the parsers' map as a parameter ?

sberyozkin: @machi1990 sorry, had to sign off for a bit.
Yes, I actually thought about it while being offline :-), but then TikaProcessor would just have a large single method :-). It just looks a bit cleaner to me to have dedicated build steps per specific parsers, or does the new CL model does not allow for it as far as Tika is concerned ?

sberyozkin: Hi @machi1990 @gsmet so what do you think; I've been thinking today, I'd like, I think, to have a build step per each parser which requires some support. The main build step is the one which generates the tika config. may be it can be called as such, `generateTikaConfig` or `createTikaConfig`, it would return a Map with the build item, and then there would PDF build step, OOXML build step, etc as required which will optionally register the extra bits if the map has the parser key. Otherwise it would indeed have to be this master build step keeping adding private calls to `preparePDF`.etc and it would feel a bit less cool to me :-)

machi1990: > @machi1990 sorry, had to sign off for a bit.

It is okay, I missed your ping too. 

> Yes, I actually thought about it while being offline :-), but then TikaProcessor would just have a large single method :-), calling `preparePdfParser`, `prepareOOXmlParser`, etc. It just looks a bit cleaner to me to have dedicated build steps per specific parser, or does the new CL model does not allow for it as far as Tika is concerned ? 

Hi @sberyozkin the change is not CL related but on whether we need the BuildItem to achieve the above. 

> Hi @machi1990 @gsmet so what do you think; I've been thinking today, I'd like, I think, to have a build step per each parser which requires some support. The main build step is the one which generates the tika config. may be it can be called as such, generateTikaConfig or createTikaConfig, it would return a Map with the build item, and then there would be PDF build step, OOXML build step, etc as required which will optionally register the extra bits if the map has the parser key. Otherwise it would indeed have to be this master build step keeping adding private calls to preparePDF.etc and it would feel a bit less cool to me :-)

If you prefer to keep it around, I can bring it back. @gsmet WDYT?

sberyozkin: @machi1990 sorry, missed your comment. Yeah, it would be nice, I reckon it is not really related to the class loading issue, though I can see why Guillaume saw it being redundant, because the optimizations plans I'm referring to above are not yet implemented :-) and that item was only used to facilitate the sub-optimal tika config generation

gsmet: It's not about being redundant. We certainly cannot pass those classes from build time to runtime and expect them to work.

I don't see the need for a build item if everything is done in the same extension.

sberyozkin: @gsmet This PR itself has preserved a reasonable number of build steps, one generates the configuration now, there are few build steps which deal with loading some resources, common to all of the Tika parsers, and some - common to PDF, and there would be OOXML specific build step coming in, we are not going ahead with collapsing them all into a single step in this PR. So it is clean and nice. Now, if this build item is removed, then for me to implement the optimization idea I had in mind I'd have to basically collapse everything (but a single step involving the Tika-common resources) into a single build step (generate the config, optionally load PDF/OOXML specific resources).
I appreciate a build item is great for coordinating between different extensions, but in this case it would help the tika processor keep the parser specific build steps separate, keeping the native optimization in mind. Np if you don't agree, please go ahead with the merge.
@machi1990 thanks for fix, cheers

sberyozkin (approved):
Hi @gsmet @machi1990 Lets go ahead with the merge as this fix may help @tpenakov in [this issue](https://github.com/quarkusio/quarkus/issues/6746).

I can reintroduce the build item in a follow up PR if I feel like it :-)

Thanks

gsmet (approved):
Nice work.

PR #6758, opened 2020-01-24
Add hazelcast-client extension

pivovarit:
Hi!

I prepared the initial version of the `hazelcast-client` extension. 

Design-wise, we need to resolve and parse configuration files at runtime in order to maintain environment independence. We could push most of it to build-time gaining a few milliseconds, but we'd need to rebuild the native image for each environment.

Please review and advise.

----
Related: https://github.com/quarkusio/quarkus/pull/6792

### DIFF
--- a/extensions/hazelcast-client/deployment/pom.xml
+++ b/extensions/hazelcast-client/deployment/pom.xml
@@ -1,0 +1,51 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>io.quarkus</groupId>
+        <artifactId>quarkus-hazelcast-client-parent</artifactId>
+        <version>999-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>quarkus-hazelcast-client-deployment</artifactId>
+    <name>Quarkus - Hazelcast - Client - Deployment</name>
+
+    <dependencies>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-core-deployment</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-arc-deployment</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-jaxb-deployment</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-hazelcast-client</artifactId>
+        </dependency>
+    </dependencies>
+
+    <build>
+        <plugins>
+            <plugin>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <configuration>
+                    <annotationProcessorPaths>
+                        <path>
+                            <groupId>io.quarkus</groupId>
+                            <artifactId>quarkus-extension-processor</artifactId>
+                            <version>${project.version}</version>
+                        </path>
+                    </annotationProcessorPaths>
+                </configuration>
+            </plugin>
+        </plugins>
+    </build>
+    
+</project>
--- a/extensions/hazelcast-client/deployment/src/main/java/io/quarkus/hazelcast/client/deployment/HazelcastClientProcessor.java
+++ b/extensions/hazelcast-client/deployment/src/main/java/io/quarkus/hazelcast/client/deployment/HazelcastClientProcessor.java
@@ -1,0 +1,235 @@
+package io.quarkus.hazelcast.client.deployment;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Set;
+
+import org.jboss.jandex.DotName;
+import org.jboss.jandex.Type;
+
+import com.hazelcast.client.cache.impl.HazelcastClientCachingProvider;
+import com.hazelcast.client.impl.ClientExtension;
+import com.hazelcast.client.impl.spi.ClientProxyFactory;
+import com.hazelcast.cluster.MembershipListener;
+import com.hazelcast.collection.ItemListener;
+import com.hazelcast.com.fasterxml.jackson.core.JsonFactory;
+import com.hazelcast.config.EventJournalConfig;
+import com.hazelcast.config.MerkleTreeConfig;
+import com.hazelcast.config.replacer.EncryptionReplacer;
+import com.hazelcast.config.replacer.PropertyReplacer;
+import com.hazelcast.config.replacer.spi.ConfigReplacer;
+import com.hazelcast.core.EntryListener;
+import com.hazelcast.internal.config.DomConfigHelper;
+import com.hazelcast.internal.util.ICMPHelper;
+import com.hazelcast.map.listener.MapListener;
+import com.hazelcast.nio.SocketInterceptor;
+import com.hazelcast.nio.serialization.DataSerializable;
+import com.hazelcast.nio.serialization.DataSerializableFactory;
+import com.hazelcast.nio.serialization.PortableFactory;
+import com.hazelcast.nio.serialization.Serializer;
+import com.hazelcast.nio.ssl.BasicSSLContextFactory;
+import com.hazelcast.partition.MigrationListener;
+import com.hazelcast.spi.discovery.DiscoveryStrategy;
+import com.hazelcast.spi.discovery.DiscoveryStrategyFactory;
+import com.hazelcast.spi.discovery.NodeFilter;
+import com.hazelcast.spi.discovery.multicast.MulticastDiscoveryStrategy;
+import com.hazelcast.topic.MessageListener;
+
+import io.quarkus.arc.deployment.AdditionalBeanBuildItem;
+import io.quarkus.deployment.annotations.BuildProducer;
+import io.quarkus.deployment.annotations.BuildStep;
+import io.quarkus.deployment.annotations.ExecutionTime;
+import io.quarkus.deployment.annotations.Record;
+import io.quarkus.deployment.builditem.ExtensionSslNativeSupportBuildItem;
+import io.quarkus.deployment.builditem.FeatureBuildItem;
+import io.quarkus.deployment.builditem.HotDeploymentWatchedFileBuildItem;
+import io.quarkus.deployment.builditem.nativeimage.NativeImageResourceBuildItem;
+import io.quarkus.deployment.builditem.nativeimage.ReflectiveClassBuildItem;
+import io.quarkus.deployment.builditem.nativeimage.ReflectiveHierarchyBuildItem;
+import io.quarkus.deployment.builditem.nativeimage.ReflectiveHierarchyIgnoreWarningBuildItem;
+import io.quarkus.deployment.builditem.nativeimage.RuntimeReinitializedClassBuildItem;
+import io.quarkus.deployment.builditem.nativeimage.ServiceProviderBuildItem;
+import io.quarkus.deployment.util.ServiceUtil;
+import io.quarkus.hazelcast.client.runtime.HazelcastClientBytecodeRecorder;
+import io.quarkus.hazelcast.client.runtime.HazelcastClientConfig;
+import io.quarkus.hazelcast.client.runtime.HazelcastClientProducer;
+
+/**
+ * @author Grzegorz Piwowarek
+ */
+class HazelcastClientProcessor {
+
+    @BuildStep
+    FeatureBuildItem feature() {
+        return new FeatureBuildItem(FeatureBuildItem.HAZELCAST_CLIENT);
+    }
+
+    @BuildStep
+    void enableSSL(BuildProducer<ExtensionSslNativeSupportBuildItem> ssl) {
+        ssl.produce(new ExtensionSslNativeSupportBuildItem(FeatureBuildItem.HAZELCAST_CLIENT));
+    }
+
+    @BuildStep
+    void registerServiceProviders(BuildProducer<ServiceProviderBuildItem> services) throws IOException {
+        registerServiceProviders(DiscoveryStrategyFactory.class, services);
+        registerServiceProviders(ClientExtension.class, services);
+        registerServiceProviders(JsonFactory.class, services);
+    }
+
+    @BuildStep
+    void setup(BuildProducer<AdditionalBeanBuildItem> additionalBeans) {
+        additionalBeans.produce(AdditionalBeanBuildItem.unremovableOf(HazelcastClientProducer.class));
+    }
+
+    @BuildStep
+    @Record(ExecutionTime.RUNTIME_INIT)
+    HazelcastClientConfiguredBuildItem resolveClientProperties(HazelcastClientBytecodeRecorder recorder,
+            HazelcastClientConfig config) {
+        recorder.configureRuntimeProperties(config);
+        return new HazelcastClientConfiguredBuildItem();
+    }
+
+    @BuildStep
+    void registerConfigurationFiles(
+            BuildProducer<NativeImageResourceBuildItem> resources,
+            BuildProducer<HotDeploymentWatchedFileBuildItem> watchedFiles) {
+        resources.produce(new NativeImageResourceBuildItem(
+                "hazelcast-client.yml",
+                "hazelcast-client-default.xml",
+                "hazelcast-client.yaml",
+                "hazelcast-client.xml"));
+
+        watchedFiles.produce(new HotDeploymentWatchedFileBuildItem("hazelcast-client.yml"));
+        watchedFiles.produce(new HotDeploymentWatchedFileBuildItem("hazelcast-client.yaml"));
+        watchedFiles.produce(new HotDeploymentWatchedFileBuildItem("hazelcast-client.xml"));
+    }
+
+    @BuildStep
+    void registerReflectivelyCreatedClasses(BuildProducer<ReflectiveClassBuildItem> reflectiveClasses) {
+        reflectiveClasses.produce(new ReflectiveClassBuildItem(false, false,
+                HazelcastClientCachingProvider.class,
+                DomConfigHelper.class,
+                EventJournalConfig.class,
+                MerkleTreeConfig.class));
+    }
+
+    @BuildStep
+    void registerCustomImplementationClasses(BuildProducer<ReflectiveHierarchyBuildItem> reflectiveClassHierarchies,
+            BuildProducer<ReflectiveHierarchyIgnoreWarningBuildItem> ignoreWarnings) {
+
+        registerTypeHierarchy(reflectiveClassHierarchies, ignoreWarnings,
+                SocketInterceptor.class,
+                MembershipListener.class,
+                MigrationListener.class,
+                EntryListener.class,
+                MessageListener.class,
+                ItemListener.class,
+                MapListener.class,
+                com.hazelcast.client.impl.ClientExtension.class,
+                ClientProxyFactory.class);
+    }
+
+    @BuildStep
+    void registerCustomConfigReplacerClasses(BuildProducer<ReflectiveClassBuildItem> reflectiveClasses,
+            BuildProducer<ReflectiveHierarchyBuildItem> reflectiveClassHierarchies,
+            BuildProducer<ReflectiveHierarchyIgnoreWarningBuildItem> ignoreWarnings) {
+
+        registerTypeHierarchy(reflectiveClassHierarchies, ignoreWarnings, ConfigReplacer.class);
+        reflectiveClasses.produce(new ReflectiveClassBuildItem(false, false,
+                EncryptionReplacer.class,
+                PropertyReplacer.class));
+    }
+
+    @BuildStep
+    void registerCustomDiscoveryStrategiesClasses(BuildProducer<ReflectiveClassBuildItem> reflectiveClasses,
+            BuildProducer<ReflectiveHierarchyBuildItem> reflectiveClassHierarchies,
+            BuildProducer<ReflectiveHierarchyIgnoreWarningBuildItem> ignoreWarnings) {
+
+        registerTypeHierarchy(reflectiveClassHierarchies, ignoreWarnings,
+                DiscoveryStrategy.class,
+                NodeFilter.class);
+
+        reflectiveClasses.produce(new ReflectiveClassBuildItem(false, false, MulticastDiscoveryStrategy.class));
+        reflectiveClasses.produce(new ReflectiveClassBuildItem(false, false,
+                "com.hazelcast.aws.AwsDiscoveryStrategy",
+                "com.hazelcast.aws.AwsDiscoveryStrategyFactory",
+                "com.hazelcast.gcp.GcpDiscoveryStrategy",
+                "com.hazelcast.gcp.GcpDiscoveryStrategyFactory"));
+
+        reflectiveClasses.produce(new ReflectiveClassBuildItem(false, false,
+                "com.hazelcast.kubernetes.HazelcastKubernetesDiscoveryStrategyFactory",
+                "com.hazelcast.kubernetes.HazelcastKubernetesDiscoveryStrategy"));
+    }
+
+    void registerServiceProviders(Class<?> klass, BuildProducer<ServiceProviderBuildItem> services) throws IOException {
+        String service = "META-INF/services/" + klass.getName();
+
+        Set<String> implementations = ServiceUtil
+                .classNamesNamedIn(Thread.currentThread().getContextClassLoader(), service);
+
+        services.produce(new ServiceProviderBuildItem(klass.getName(), new ArrayList<>(implementations)));
+    }
+
+    @BuildStep
+    void registerCustomCredentialFactories(BuildProducer<ReflectiveClassBuildItem> reflectiveClasses,
+            BuildProducer<ReflectiveHierarchyBuildItem> reflectiveClassHierarchies,
+            BuildProducer<ReflectiveHierarchyIgnoreWarningBuildItem> ignoreWarnings) {
+        registerTypeHierarchy(
+                reflectiveClassHierarchies, ignoreWarnings,
+                com.hazelcast.security.ICredentialsFactory.class);
+
+        reflectiveClasses.produce(
+                new ReflectiveClassBuildItem(false, false, com.hazelcast.config.security.StaticCredentialsFactory.class));
+    }
+
+    @BuildStep
+    void registerSSLUtilities(BuildProducer<ReflectiveClassBuildItem> reflectiveClasses,
+            BuildProducer<ReflectiveHierarchyBuildItem> reflectiveClassHierarchies,
+            BuildProducer<ReflectiveHierarchyIgnoreWarningBuildItem> ignoreWarnings) {
+
+        registerTypeHierarchy(
+                reflectiveClassHierarchies, ignoreWarnings,
+                com.hazelcast.nio.ssl.SSLContextFactory.class);
+        reflectiveClasses.produce(
+                new ReflectiveClassBuildItem(false, false, BasicSSLContextFactory.class));
+    }
+
+    @BuildStep
+    void registerUserImplementationsOfSerializableUtilities(
+            BuildProducer<ReflectiveHierarchyBuildItem> reflectiveClassHierarchies,
+            BuildProducer<ReflectiveHierarchyIgnoreWarningBuildItem> ignoreWarnings) {
+        registerTypeHierarchy(reflectiveClassHierarchies, ignoreWarnings,
+                DataSerializable.class,
+                DataSerializableFactory.class,
+                PortableFactory.class,
+                Serializer.class);
+    }
+
+    @BuildStep
+    void registerICMPHelper(BuildProducer<NativeImageResourceBuildItem> resources,
+            BuildProducer<RuntimeReinitializedClassBuildItem> reinitializedClasses) {
+        resources.produce(new NativeImageResourceBuildItem(
+                "lib/linux-x86/libicmp_helper.so",
+                "lib/linux-x86_64/libicmp_helper.so"));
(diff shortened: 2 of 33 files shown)


pivovarit:
Should I add the `hazelcast-client` module to one of the pipelines?

geoand:
Thanks a lot for this.

@pivovarit yes please add it. Otherwise the native tests won't be run.

extensions/hazelcast-client/deployment/src/main/java/io/quarkus/hazelcast/client/deployment/HazelcastClientProcessor.java
```diff
@@ -0,0 +1,249 @@
...
+import io.quarkus.deployment.builditem.nativeimage.ReflectiveHierarchyBuildItem;
+import io.quarkus.deployment.builditem.nativeimage.RuntimeReinitializedClassBuildItem;
+import io.quarkus.deployment.builditem.nativeimage.ServiceProviderBuildItem;
+import io.quarkus.deployment.util.ServiceUtil;
+import io.quarkus.hazelcast.client.HazelcastClientBytecodeRecorder;
+import io.quarkus.hazelcast.client.HazelcastClientConfig;
+import io.quarkus.hazelcast.client.HazelcastClientProducer;
+
+/**
+ * @author Grzegorz Piwowarek
```
geoand: Please remove the author tags, we have decided to not add them to new code.

pivovarit:
> Thanks a lot for this.
> 
> @pivovarit yes please add it. Otherwise the native tests won't be run.

You are welcome!

I believe it's already running. If I just `mvn clean install -Dquarkus.native.container-build=true ` everything looks good. 

However, I still haven't deciphered your pipelines so I might be missing a place or two, could you point me to the right place?

geoand:
> 
> However, I still haven't deciphered your pipelines so I might be missing a place or two, could you point me to the right place?

Here is the place where we define the various native tests: https://github.com/quarkusio/quarkus/blob/master/ci-templates/stages.yml#L190

extensions/hazelcast-client/runtime/src/main/resources/META-INF/quarkus-extension.yaml
```diff
@@ -0,0 +1,13 @@
+---
+name: "Hazelcast Client"
+metadata:
+  keywords:
+  - "hazelcast-client"
+  - "hazelcast"
+  - "cache"
+  - "distributed cache"
+  - "caching"
+  guide: "https://quarkus.io/guides/hazelcast-client"
```
gastaldi: You need to add a `hazelcast-client.adoc` in https://github.com/quarkusio/quarkus/tree/master/docs/src/main/asciidoc to avoid the broken link

integration-tests/hazelcast-client/src/main/java/io/quarkus/it/hazelcast/client/CustomNodeFilter.java
```diff
@@ -0,0 +1,12 @@
+package io.quarkus.it.hazelcast.client;
+
+import com.hazelcast.spi.discovery.DiscoveryNode;
+import com.hazelcast.spi.discovery.NodeFilter;
+
+class CustomNodeFilter implements NodeFilter {
+    @Override
+    public boolean test(DiscoveryNode candidate) {
+        System.out.println("using custom node filter");
```
gastaldi: Remove any call to System.out in tests

extensions/hazelcast-client/runtime/pom.xml
```diff
@@ -0,0 +1,57 @@
...
+            <artifactId>quarkus-core</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-arc</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>com.hazelcast</groupId>
+            <artifactId>hazelcast-all</artifactId>
+            <version>3.12.5</version>
```
gastaldi: Add these versions to the bom/runtime/pom.xml BOM

gastaldi: Also out of curiosity, do you need the `-all` dependency or can you use separate artifacts? It looks like it's 12Mb and I am not sure if all these classes are required, as they would impact the size of the resulting application

pivovarit: In a new 4.0 release, we'll be merging some artifacts, but I believe for 3.0 we should be able to strip it down a bit

extensions/hazelcast-client/deployment/src/main/java/io/quarkus/hazelcast/client/deployment/HazelcastClientProcessor.java
```diff
@@ -0,0 +1,246 @@
...
+                "com.sun.org.apache.xml.internal.serializer.utils.SerializerMessages"));
+        bundles.produce(new NativeImageResourceBundleBuildItem(
+                "com.sun.org.apache.xerces.internal.impl.msg.XMLMessages"));
+        bundles.produce(new NativeImageResourceBundleBuildItem(
+                "com.sun.org.apache.xerces.internal.impl.msg.XMLSchemaMessages"));
+        bundles.produce(new NativeImageResourceBundleBuildItem(
+                "com.sun.org.apache.xerces.internal.impl.xpath.regex.message"));
+
+        resources.produce(new NativeImageResourceBuildItem(
+                "com/sun/org/apache/xml/internal/serializer/output_xml.properties"));
```
gastaldi: It looks like they are already handled in the `jaxb` extension? Try depending your extension on it and see if that's sufficient.

extensions/hazelcast-client/deployment/src/main/java/io/quarkus/hazelcast/client/deployment/HazelcastClientProcessor.java
```diff
@@ -0,0 +1,246 @@
...
+import io.quarkus.deployment.builditem.nativeimage.RuntimeReinitializedClassBuildItem;
+import io.quarkus.deployment.builditem.nativeimage.ServiceProviderBuildItem;
+import io.quarkus.deployment.util.ServiceUtil;
+import io.quarkus.hazelcast.client.HazelcastClientBytecodeRecorder;
+import io.quarkus.hazelcast.client.HazelcastClientConfig;
+import io.quarkus.hazelcast.client.HazelcastClientProducer;
+
+class HazelcastClientProcessor {
+
+    private static final String FEATURE = "hazelcast-client";
```
gastaldi: Move this to the `FeatureBuildItem`  class for now

extensions/hazelcast-client/deployment/src/main/java/io/quarkus/hazelcast/client/deployment/HazelcastClientProcessor.java
```diff
@@ -0,0 +1,246 @@
...
+
+    CombinedIndexBuildItem buildIndex;
+
+    BuildProducer<ReflectiveClassBuildItem> reflectiveClasses;
+    BuildProducer<NativeImageResourceBuildItem> resources;
+    BuildProducer<ReflectiveHierarchyBuildItem> reflectiveClassHierarchies;
+    BuildProducer<NativeImageResourceBundleBuildItem> bundles;
+    BuildProducer<RuntimeReinitializedClassBuildItem> reinitializedClasses;
+    BuildProducer<ServiceProviderBuildItem> services;
```
gastaldi: As a good practice, avoid storing the BuildProducers as attributes and pass them as parameters in the methods you need them. This helps extension maintainers to understand what's necessary in each method

extensions/hazelcast-client/deployment/src/main/java/io/quarkus/hazelcast/client/deployment/HazelcastClientProcessor.java
```diff
@@ -0,0 +1,246 @@
...
+
+    @BuildStep
+    void enableSSL(BuildProducer<ExtensionSslNativeSupportBuildItem> ssl) {
+        ssl.produce(new ExtensionSslNativeSupportBuildItem(FEATURE));
+    }
+
+    @BuildStep
+    void enableJNI(BuildProducer<JniBuildItem> jni) {
+        jni.produce(new JniBuildItem());
+    }
```
gastaldi: Not a big deal, but you can just return the instantiated BuildItems in these methods :)

integration-tests/hazelcast-client/src/test/java/io/quarkus/it/hazelcast/client/HazelcastServerTestResource.java
```diff
@@ -0,0 +1,25 @@
...
+
+    @Override
+    public Map<String, String> start() {
+        member1 = Hazelcast.newHazelcastInstance();
+        return Collections.emptyMap();
+    }
+
+    @Override
+    public void stop() {
+        Hazelcast.shutdownAll();
```
gastaldi: Maybe call member1.shutdown() just in case?

geoand:
`master` (hopefully) contains a fix for the CI issue so rebasing onto it is advised

pivovarit:
@geoand sure thing, the work here will slow down a bit since I'm reworking a few things - we made a decision to target the next version of `hazelcast` which requires a couple more native mode fixes

geoand:
Sure thing, thanks for the update!

extensions/hazelcast-client/deployment/src/main/java/io/quarkus/hazelcast/client/deployment/HazelcastClientProcessor.java
pivovarit: Unfortunately, it doesn't really help - I will investigate and ideally fix the issues in `jaxb` so that we can reuse it

extensions/hazelcast-client/README.MD
```diff
@@ -0,0 +1,62 @@
...
+- User code deployment is not supported
+- SSL support requires extra runtime configuration
+
+### SSL
+In order for SSL to work, one needs to properly set `java.library.path` and `javax.net.ssl.trustStore` runtime parameters.
+
+Example Dockerfile:
+
+```
+FROM quay.io/quarkus/ubi-quarkus-native-image:19.2.1 as graalvm
```
gsmet: This is outdated as we now only support 19.3.1 in master and that's my problem with having a README here: it will quickly be outdated. All this should be in the guide we publish to the website instead IMHO. We have attributes there and they are probably updated for each release.

extensions/hazelcast-client/deployment/src/main/java/io/quarkus/hazelcast/client/deployment/HazelcastClientProcessor.java
gsmet: +1

extensions/hazelcast-client/deployment/src/main/java/io/quarkus/hazelcast/client/deployment/HazelcastClientProcessor.java
```diff
@@ -0,0 +1,237 @@
...
+
+    @BuildStep
+    void enableSSL(BuildProducer<ExtensionSslNativeSupportBuildItem> ssl) {
+        ssl.produce(new ExtensionSslNativeSupportBuildItem(FeatureBuildItem.HAZELCAST_CLIENT));
+    }
+
+    @BuildStep
+    void enableJNI(BuildProducer<JniBuildItem> jni) {
+        jni.produce(new JniBuildItem());
```
gsmet: You can drop that one, it's not needed anymore in master as 19.3 has JNI always enabled.

extensions/hazelcast-client/runtime/pom.xml
```diff
@@ -0,0 +1,61 @@
...
+        </dependency>
+        <dependency>
+            <groupId>javax.cache</groupId>
+            <artifactId>cache-api</artifactId>
+            <version>1.1.1</version>
```
gsmet: Same here it should be in the BOM. But it doesn't come with Hazelcast? I'm surprised you even need to add it.

I'm also totally unsure it's useful as we don't have JSR 107 support in Quarkus.

extensions/hazelcast-client/runtime/src/main/java/io/quarkus/hazelcast/client/runtime/HazelcastClientConfig.java
```diff
@@ -0,0 +1,65 @@
...
+import io.quarkus.runtime.annotations.ConfigRoot;
+
+@ConfigRoot(name = "hazelcast-client", phase = ConfigPhase.RUN_TIME)
+public class HazelcastClientConfig {
+
+    /**
+     * Hazelcast Cluster address
+     */
+    @ConfigItem
+    public Optional<String> clusterMembers;
```
gsmet: Can't they be a list of `InetSocketAddress`?

It would be better to not have it optional and have a default value instead (if possible). That way, it's automatically documented.

extensions/hazelcast-client/runtime/src/main/java/io/quarkus/hazelcast/client/runtime/HazelcastClientConfig.java
```diff
@@ -0,0 +1,65 @@
...
+     * Outbound port definition
+     */
+    @ConfigItem
+    public Optional<String> outboundPortDefinitions;
+
+    /**
+     * Connection attempt limit
+     */
+    @ConfigItem
+    public Optional<Integer> connectionAttemptLimit;
```
gsmet: Better use `OptionalInt` for all of these.

extensions/hazelcast-client/runtime/src/main/java/io/quarkus/hazelcast/client/runtime/HazelcastConfigurationParser.java
```diff
@@ -0,0 +1,75 @@
...
+        setConnectionAttemptPeriod(clientConfig, config);
+
+        setExecutorPoolSize(clientConfig, config);
+
+        return clientConfig;
+    }
+
+    private void setClusterAddress(ClientConfig clientConfig, HazelcastClientConfig config) {
+        config.clusterMembers
+                .ifPresent(members -> clientConfig.getNetworkConfig().addAddress(members.split(",")));
```
gsmet: We avoid lambdas in runtime code. They have a cost.

gsmet: Your config should be a List and the split on `,` will be done automatically by the config framework.

extensions/hazelcast-client/runtime/src/main/java/io/quarkus/hazelcast/client/runtime/HazelcastConfigurationParser.java
```diff
@@ -0,0 +1,75 @@
...
+
+    private void setGroupName(ClientConfig clientConfig, HazelcastClientConfig config) {
+        config.groupName
+                .filter(s -> !s.isEmpty())
+                .ifPresent(groupName -> clientConfig.getGroupConfig().setName(groupName));
+    }
+
+    private void setLabels(ClientConfig clientConfig, HazelcastClientConfig config) {
+        config.labels
+                .map(s -> Arrays.stream(s.split(","))).orElseGet(Stream::empty)
```
gsmet: Same here and for all the other elements that are split.

extensions/hazelcast-client/runtime/src/main/resources/META-INF/quarkus-extension.yaml
```diff
@@ -0,0 +1,13 @@
...
+  categories:
+  - "data"
+  status: "stable"
```
gsmet: Let's make it `preview` first.

gsmet (changes requested):
Thanks for this work. I added a couple of comments.

pivovarit:
@gsmet thanks for all the comments, keep in mind it's still work in progress - I will call all of you once all the changes are done - I will wait a week or two for the Hazelcast 4.0.0

extensions/hazelcast-client/README.MD
pivovarit: I agree, I know already that there's a separate place for guides and I'm planning to move it all there!

extensions/hazelcast-client/runtime/pom.xml
pivovarit: Hazelcast 3.0.0 has `cache-api` in as _provided_ dependency - most likely we should be able to get rid of it in the next release where we're reshaping our artifacts

extensions/hazelcast-client/deployment/src/main/java/io/quarkus/hazelcast/client/deployment/HazelcastClientProcessor.java
pivovarit: Fix'd: https://github.com/quarkusio/quarkus/pull/6863/files

pivovarit:
[Hazelcast 4.0 is already in Maven Central](https://search.maven.org/artifact/com.hazelcast/hazelcast/4.0/jar) so I will update the PR and it should be ready today/tomorrow for the ultimate review :)

integration-tests/hazelcast-client/src/test/java/io/quarkus/it/hazelcast/client/HazelcastServerTestResource.java
pivovarit: All the client instances get registered internally so I'd say this is the "just in case" option

pivovarit:
It's almost ready. However, I need to consult one thing.

At the moment, the bean produced by the producer is initialized lazily (I believe it's CDI's default).
This results in our client establishing a connection on the first call touching the client bean.

Not sure if this is desired or not. Should we leave the choice to users and leave it as it or force eager init?

geoand:
> It's almost ready. However, I need to consult a few things.
> 
> At the moment, the bean produced by the producer is initialized lazily (I believe it's CDI's default).
> This results in our client establishing a connection on the first call touching the client bean.
> 
> Not sure if this is desired or not. Should we leave the choice to users and leave it as it or force eager init?
> 

@karesti how is Infinispan handled? I think we should follow the same model here.

> Another thing, there's no separation between `hazelcast` and `hazelcast-client` anymore, it's just `hazelcast` since 4.0. Is it ok if we provide both embedded and client functionality through a single `quarkus-hazelcast` extension?

I am pretty sure that for Infinispan there are separate extensions and I think we should try to be consistent. Again @karesti can provide more input here :)

extensions/hazelcast-client/runtime/pom.xml
pivovarit: That's no longer the case in 4.0, we are on the bare minimum.

extensions/hazelcast-client/deployment/src/main/java/io/quarkus/hazelcast/client/deployment/HazelcastClientProcessor.java
```diff
@@ -0,0 +1,232 @@
...
+        return new HazelcastClientConfiguredBuildItem();
+    }
+
+    @BuildStep
+    void registerConfigurationFiles(BuildProducer<NativeImageResourceBuildItem> resources) {
+        resources.produce(new NativeImageResourceBuildItem(
+                "hazelcast-client.yml",
+                "hazelcast-client-default.xml",
+                "hazelcast-client.yaml",
+                "hazelcast-client.xml"));
```
machi1990: Maybe also watch these files for modification with `HotDeploymentWatchedFileBuildItem`?

pivovarit: Good idea, thanks!

extensions/hazelcast-client/deployment/src/main/java/io/quarkus/hazelcast/client/deployment/HazelcastClientProcessor.java
```diff
@@ -0,0 +1,232 @@
...
+    }
+
+    @BuildStep
+    void reinitializeCommonFJP(BuildProducer<RuntimeReinitializedClassBuildItem> reinitializedClasses) {
+        // TODO remove after migrating to GraalVM 19.3.1 (deferred ForkJoinPool.commonPool())
```
machi1990: Current master is on 19.3.1, I guess we can remove this.

extensions/hazelcast-client/deployment/pom.xml
```diff
@@ -0,0 +1,52 @@
...
+            <artifactId>quarkus-arc-deployment</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-jaxb-deployment</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-hazelcast-client</artifactId>
+            <version>${project.version}</version>
```
gastaldi: You can remove this if you have added `quarkus-hazelcast-client` to the `bom/runtime/pom.xml`

extensions/hazelcast-client/runtime/pom.xml
gastaldi: Is this dependency still needed now that Hazelcast Client 4.0 is used?

gastaldi (changes requested):
Looks good. Added some minor comments that would be nice to be addressed before merging this

extensions/hazelcast-client/runtime/pom.xml
pivovarit: Unfortunately, yes. The dependency is not used for anything other than making GraalVM happy, and manual removal of it from core classes via substitutions would last weeks and be super fragile.

The real hope to get rid of is if this gets implemented: https://github.com/oracle/graal/issues/1664 and then similar feature could be exposed through Quarkus build item

gastaldi (approved):
LGTM

extensions/hazelcast-client/runtime/pom.xml
geoand: Are you sure that stripping out the classes via substitutions would be that difficult? We really try to minimize runtime dependencies as much as possible.

pivovarit: To give you a clue what the volume of changes is: https://github.com/hazelcast/hazelcast/search?q=javax.cache&unscoped_q=javax.cache

We're totally coupled to jcache

geoand: I see, thanks

wburns:
> > It's almost ready. However, I need to consult a few things.
> > At the moment, the bean produced by the producer is initialized lazily (I believe it's CDI's default).
> > This results in our client establishing a connection on the first call touching the client bean.
> > Not sure if this is desired or not. Should we leave the choice to users and leave it as it or force eager init?
> 
> @karesti how is Infinispan handled? I think we should follow the same model here.

This is the same for Infinispan Client. The connection isn't initialized until the bean is referenced.

> 
> > Another thing, there's no separation between `hazelcast` and `hazelcast-client` anymore, it's just `hazelcast` since 4.0. Is it ok if we provide both embedded and client functionality through a single `quarkus-hazelcast` extension?
> 
> I am pretty sure that for Infinispan there are separate extensions and I think we should try to be consistent. Again @karesti can provide more input here :)

I would think you would want to have them separate. In the Infinispan case, the client is a lot leaner than the embedded extension so we don't want them together.

pivovarit:
@wburns thanks, I took the same approach

extensions/hazelcast-client/runtime/pom.xml
pivovarit: What we could do instead is to make the dependency optional and include a note in the README that it's required for native... let me know if this is ok

geoand: I think it's probably better as is. We really try to make the switch from jvm mode to native just a matter of adding `-Dnative`.

integration-tests/hazelcast-client/src/test/java/io/quarkus/it/hazelcast/client/HazelcastServerTestResource.java
```diff
@@ -0,0 +1,25 @@
...
+
+import java.util.Collections;
+import java.util.Map;
+
+import com.hazelcast.core.Hazelcast;
+import com.hazelcast.core.HazelcastInstance;
+
+import io.quarkus.test.common.QuarkusTestResourceLifecycleManager;
+
+public class HazelcastServerTestResource implements QuarkusTestResourceLifecycleManager {
```
geoand: :+1:

geoand: Perhaps this could also be mentioned in the doc? Otherwise I think it will be unlikely most users will discover it.

pivovarit: That's just a part of testing infrastructure under `src/test`, I believe that should not be exposed. If it is somehow, it's a bug, I believe

pivovarit: Used only here: https://github.com/quarkusio/quarkus/blob/6b7fbe650cd3fee005ad6b94485f21491ac1053b/integration-tests/hazelcast-client/src/test/java/io/quarkus/it/hazelcast/client/HazelcastClientFunctionalityTest.java#L12

geoand: Yeah, but aren't users supposed to be able to use this resource? I mean the other resources we have users can use and are encouraged to use actually

pivovarit: I did not really realize it's going to be public, but well, no problem :) I will double-check and add a note in the docs

geoand: See for example: `DerbyDatabaseTestResource` or `H2DatabaseTestResource`. I mean it's up to you if you want to expose it or not.
But wouldn't it be useful for users to have?

pivovarit: Actually, it's quite useful :) yeah, would love to move forward with that

geoand: Excellent!

pivovarit: I exposed the functionality via `quarkus-test-hazelcast` :)

geoand: With accompanying documentation, :+1:

geoand:
@pivovarit Thank you a lot for this work! 

I'm sorry to let you know that because we are in a real sprint to get 1.3 out of the door, I am unlikely to be able to take another good look at this before then. 
I'll circle back as soon as possible.

Thanks for your patience!

pivovarit:
Just a random thought here. After this gets in, I'd love to integrate it with Hibernate 2LC alongside Caffeine and Infinispan

pivovarit:
Can I help you somehow with pushing this forward? :)

btw, I just updated the newly introduced GitHub Actions pipelines.

emmanuelbernard:
Hey @pivovarit 

Thanks for the PR!

tl;dr; We want the extension to be in Quarkus but that does not mean it has to go in the quarkus repo (we call it incorrectly quarkus-core).

Longer version
So, we definitely want the Hazelcast client extension to be present in the ecosystem (code.quarkus.io being the main window of it). We tried the "monorepo" approach but it is straining our systems too much (humans to keep track of everything, machines e.g. CI - we are close to the GH action limits already, etc). So what we are doing now is move new extensions to their own GH repos and integrate them in quarkus-platform (which is the BOM + metadata we use to create a quarkus release) https://github.com/quarkusio/quarkus-platform/
Right now the Debezium outbox pattern and the JMS for QPid extensions are being hosted that way. This is still a bit of work in progress but that's where we are aiming and will help us to scale.

For a user, there is no difference between extensions hosted on quarkusio/quarkus or elsewhere. In particular, the doc and code.quarkus.io presence is the same. For an extension developer, the difference is that
- the release cycle is more in your control
- issue management is in your control too
- it's not part of the initial CI build upon each change but now that we have moved to GH actions, we believe it will be easy to chain GH Action builds and thus trigger a platform extension build after core has changed

I am sorry it's taking so long and for this extra step. If you all are ok with this plan, my hope is for the hazelcast extension to be amongst the early adopters of this model (with QPid and Debezium).

BTW another point, you plan to keep maintaining it, right? If an extension is part of the platform, sometimes we change extension contracts to keep the core evolving and we would need your help in keeping it up to date to ensure the platform does not lose features.

geoand:
@pivovarit Thank you very much for your hard work and understanding on this. I'll also like to express my sympathies for you getting caught up in the middle of our repo reorganization.

Following up on what @emmanuelbernard mentions, there are a few simple concrete steps that you can follow to make your extension part of the Quarkus platform.

* Create your own repo to host the extension
* Open a PR against the Quarkus Platform repo: https://github.com/quarkusio/quarkus-platform

The good thing is that there are already two PRs in flight doing the exact same thing: https://github.com/quarkusio/quarkus-platform/pull/28, https://github.com/quarkusio/quarkus-platform/pull/25 so you should be have some good examples of what needs to be done.
And of course any help you need, don't hesitate to ping us

pivovarit:
Thank you for the update - I feel the monorepo pain and will gladly help you with migrating to the new setup. I will move the content to a separate repository and integrate it with the platform soon.

pivovarit:
Superseded with: https://github.com/hazelcast/quarkus-hazelcast-client/pull/1

pivovarit:
@geoand I have one question. Where should I add the documentation so that it's visible on the Quarkus site?

geoand:
@gsmet would be the most qualified to answer this question :)

gastaldi:
@pivovarit docs go in https://github.com/quarkusio/quarkus/tree/master/docs/src/main/asciidoc for now. Since it's a community extension, It should preferably just contain a reference to the docs maintained in your repository (See https://github.com/quarkusio/quarkus/blob/master/docs/src/main/asciidoc/camel.adoc)

PR #6762, opened 2020-01-24
Upgrade to Hibernate Search 6.0.0.Beta4

yrodiere:
The quickstart does not need to change: it works out of the box after this upgrade.


gsmet (approved):
Woot!

PR #6765, opened 2020-01-24
Remove configuration related code that is no longer used

geoand:
This stuff see to be left over from the previous iteration of the Configuration support and resulted in the generation of a `io.quarkus.runtime.generated.ConfigSourceProviderImpl` class that always returned an empty array


geoand:
Nevermind... It seems like me IDE was ignoring the vault module. 
Sorry for the noise

PR #6771, opened 2020-01-24
Fix FaultTolerance perf issue related to the creation of FaultToleranceOperation

### LINKED ISSUE #6761: QuarkusFaultToleranceOperationProvider should cache the FaultToleranceOperation
**Description**
`QuarkusFaultToleranceOperationProvider` re-creates a new `FaultToleranceOperation` object for each request. 

CPU profiling shows that creating such object can uses a significant amount of CPU, see here for details: https://groups.google.com/d/msg/quarkus-dev/2jFpwMt4QaI/edUfNagDCgAJ

Smallrye-fault-tolerance provides a CDI extension that caches the `FaultToleranceOperation` but Quarkus didn't use it (CDI portable extensions are not supported in Quarkus ) so we need to create something dedicated to Quarkus.

**Implementation ideas**
Fault tolerance extension can be found here: https://github.com/smallrye/smallrye-fault-tolerance/blob/master/implementation/fault-tolerance/src/main/java/io/smallrye/faulttolerance/DefaultFaultToleranceOperationProvider.java#L43-L56

A prototype implementation that needs to be reworked can be found here:  https://github.com/quarkusio/quarkus/compare/master...loicmathieu:fix/ft-perf-issue

loicmathieu:
Fixes #6761

Still a draft PR as I want some feedback on the implementation before polishing it (if needed).

As discuss on the mailing list and on the issue, there is some performance issue as currently the `FaultToleranceOperation` is re-created on each request.

The first commit address this issue by using a cache that is filled at `RUNTIME_INIT` by the recorder. It works except for FT annotations on interfaces (like the one on RestClient interfaces). So there is a fallback that create it on the fly. If we only keep this change we need to store inside the cache the `FaultToleranceOperation` created on the fly to fix the issue correctly.

I made a second change when I discover that the recorder load all bean classes to analyse them for FT annotations. I restrict the list of bean classes to analyse using Jandex indexes and it improve a lot the performance of scanning the classpath for FT annotations. On the main integration test, the time to `validate` the annotations drop from 200ms to 7ms => improvement of 20x at runtime !

If we keep the second commit, we will also create `FaultToleranceOperation` for RestClient interfaces correctly at `RUNTIME_INIT` by the recorder as Jandex indexe interfaces.

@Ladicek @mkouba @michalszynkiewicz WDYT ?

### DIFF
--- a/extensions/smallrye-fault-tolerance/deployment/src/main/java/io/quarkus/smallrye/faulttolerance/deployment/SmallRyeFaultToleranceProcessor.java
+++ b/extensions/smallrye-fault-tolerance/deployment/src/main/java/io/quarkus/smallrye/faulttolerance/deployment/SmallRyeFaultToleranceProcessor.java
@@ -1,14 +1,11 @@
 package io.quarkus.smallrye.faulttolerance.deployment;
 
 import java.time.temporal.ChronoUnit;
-import java.util.ArrayList;
 import java.util.HashSet;
-import java.util.List;
 import java.util.OptionalInt;
 import java.util.Set;
 
 import javax.annotation.Priority;
-import javax.inject.Inject;
 import javax.inject.Singleton;
 
 import org.eclipse.microprofile.config.Config;
@@ -25,15 +22,19 @@ import org.jboss.jandex.AnnotationValue;
 import org.jboss.jandex.ClassInfo;
 import org.jboss.jandex.DotName;
 import org.jboss.jandex.IndexView;
+import org.jboss.jandex.MethodInfo;
 
 import io.quarkus.arc.deployment.AdditionalBeanBuildItem;
 import io.quarkus.arc.deployment.AnnotationsTransformerBuildItem;
 import io.quarkus.arc.deployment.BeanArchiveIndexBuildItem;
 import io.quarkus.arc.deployment.BeanDefiningAnnotationBuildItem;
 import io.quarkus.arc.deployment.ValidationPhaseBuildItem;
+import io.quarkus.arc.processor.AnnotationStore;
 import io.quarkus.arc.processor.AnnotationsTransformer;
 import io.quarkus.arc.processor.BeanInfo;
+import io.quarkus.arc.processor.BuildExtension;
 import io.quarkus.arc.processor.BuiltinScope;
+import io.quarkus.arc.processor.DotNames;
 import io.quarkus.deployment.Capabilities;
 import io.quarkus.deployment.annotations.BuildProducer;
 import io.quarkus.deployment.annotations.BuildStep;
@@ -60,14 +61,15 @@ import io.smallrye.faulttolerance.propagation.ContextPropagationExecutorFactory;
 
 public class SmallRyeFaultToleranceProcessor {
 
-    @Inject
-    BuildProducer<ReflectiveClassBuildItem> reflectiveClass;
-
-    @Inject
-    BuildProducer<NativeImageSystemPropertyBuildItem> nativeImageSystemProperty;
-
-    @Inject
-    CombinedIndexBuildItem combinedIndexBuildItem;
+    private static final Set<DotName> FT_ANNOTATIONS = new HashSet<>();
+    static {
+        FT_ANNOTATIONS.add(DotName.createSimple(Asynchronous.class.getName()));
+        FT_ANNOTATIONS.add(DotName.createSimple(Bulkhead.class.getName()));
+        FT_ANNOTATIONS.add(DotName.createSimple(CircuitBreaker.class.getName()));
+        FT_ANNOTATIONS.add(DotName.createSimple(Fallback.class.getName()));
+        FT_ANNOTATIONS.add(DotName.createSimple(Retry.class.getName()));
+        FT_ANNOTATIONS.add(DotName.createSimple(Timeout.class.getName()));
+    }
 
     @BuildStep
     public void build(BuildProducer<AnnotationsTransformerBuildItem> annotationsTransformer,
@@ -75,20 +77,15 @@ public class SmallRyeFaultToleranceProcessor {
             BuildProducer<ServiceProviderBuildItem> serviceProvider,
             BuildProducer<BeanDefiningAnnotationBuildItem> additionalBda,
             Capabilities capabilities,
-            BuildProducer<SystemPropertyBuildItem> systemProperty) {
+            BuildProducer<SystemPropertyBuildItem> systemProperty,
+            CombinedIndexBuildItem combinedIndexBuildItem,
+            BuildProducer<ReflectiveClassBuildItem> reflectiveClass,
+            BuildProducer<NativeImageSystemPropertyBuildItem> nativeImageSystemProperty) {
 
         feature.produce(new FeatureBuildItem(FeatureBuildItem.SMALLRYE_FAULT_TOLERANCE));
 
         serviceProvider.produce(new ServiceProviderBuildItem(ExecutorFactory.class.getName(),
                 ContextPropagationExecutorFactory.class.getName()));
-
-        Set<DotName> ftAnnotations = new HashSet<>();
-        ftAnnotations.add(DotName.createSimple(Asynchronous.class.getName()));
-        ftAnnotations.add(DotName.createSimple(Bulkhead.class.getName()));
-        ftAnnotations.add(DotName.createSimple(CircuitBreaker.class.getName()));
-        ftAnnotations.add(DotName.createSimple(Fallback.class.getName()));
-        ftAnnotations.add(DotName.createSimple(Retry.class.getName()));
-        ftAnnotations.add(DotName.createSimple(Timeout.class.getName()));
 
         IndexView index = combinedIndexBuildItem.getIndex();
 
@@ -111,7 +108,7 @@ public class SmallRyeFaultToleranceProcessor {
             additionalBean.produce(fallbackHandlersBeans.build());
         }
 
-        for (DotName annotation : ftAnnotations) {
+        for (DotName annotation : FT_ANNOTATIONS) {
             reflectiveClass.produce(new ReflectiveClassBuildItem(true, false, annotation.toString()));
             // also make them bean defining annotations
             additionalBda.produce(new BeanDefiningAnnotationBuildItem(annotation));
@@ -126,7 +123,7 @@ public class SmallRyeFaultToleranceProcessor {
 
             @Override
             public void transform(TransformationContext context) {
-                if (ftAnnotations.contains(context.getTarget().asClass().name())) {
+                if (FT_ANNOTATIONS.contains(context.getTarget().asClass().name())) {
                     context.transform().add(FaultToleranceBinding.class).done();
                 }
             }
@@ -136,7 +133,7 @@ public class SmallRyeFaultToleranceProcessor {
         AdditionalBeanBuildItem.Builder builder = AdditionalBeanBuildItem.builder();
         // Also register MP FT annotations so that they are recognized as interceptor bindings
         // Note that MP FT API jar is nor indexed, nor contains beans.xml so it is not part of the app index
-        for (DotName ftAnnotation : ftAnnotations) {
+        for (DotName ftAnnotation : FT_ANNOTATIONS) {
             builder.addBeanClass(ftAnnotation.toString());
         }
         builder.addBeanClasses(FaultToleranceInterceptor.class,
@@ -187,13 +184,43 @@ public class SmallRyeFaultToleranceProcessor {
     @BuildStep
     // needs to be RUNTIME_INIT because we need to read MP Config
     @Record(ExecutionTime.RUNTIME_INIT)
-    void validateFaultToleranceAnnotations(
-            ValidationPhaseBuildItem validationPhase, SmallryeFaultToleranceRecorder recorder) {
-        List<String> beanNames = new ArrayList<>();
-        for (BeanInfo bean : validationPhase.getContext().beans().classBeans()) {
-            beanNames.add(bean.getBeanClass().toString());
+    void validateFaultToleranceAnnotations(SmallryeFaultToleranceRecorder recorder,
+            ValidationPhaseBuildItem validationPhase,
+            BeanArchiveIndexBuildItem beanArchiveIndexBuildItem) {
+        AnnotationStore annotationStore = validationPhase.getContext().get(BuildExtension.Key.ANNOTATION_STORE);
+        Set<String> beanNames = new HashSet<>();
+        IndexView index = beanArchiveIndexBuildItem.getIndex();
+
+        for (BeanInfo info : validationPhase.getContext().beans()) {
+            if (hasFTAnnotations(index, annotationStore, info.getImplClazz())) {
+                beanNames.add(info.getBeanClass().toString());
+            }
         }
-        recorder.validate(beanNames);
+
+        recorder.createFaultToleranceOperation(beanNames);
+    }
+
+    private boolean hasFTAnnotations(IndexView index, AnnotationStore annotationStore, ClassInfo info) {
+        // first check annotations on type
+        if (annotationStore.hasAnyAnnotation(info, FT_ANNOTATIONS)) {
+            return true;
+        }
+
+        // then check on the methods
+        for (MethodInfo method : info.methods()) {
+            if (annotationStore.hasAnyAnnotation(method, FT_ANNOTATIONS)) {
+                return true;
+            }
+        }
+
+        // then check on the parent
+        DotName parentClassName = info.superName();
+        ClassInfo parentClassInfo = index.getClassByName(parentClassName);
+        if (parentClassName.equals(DotNames.OBJECT) || parentClassInfo == null) {
+            return false; //no more parents
+        }
+
+        return hasFTAnnotations(index, annotationStore, parentClassInfo);
     }
 
     @BuildStep
--- a/extensions/smallrye-fault-tolerance/runtime/src/main/java/io/quarkus/smallrye/faulttolerance/runtime/QuarkusFaultToleranceOperationProvider.java
+++ b/extensions/smallrye-fault-tolerance/runtime/src/main/java/io/quarkus/smallrye/faulttolerance/runtime/QuarkusFaultToleranceOperationProvider.java
@@ -1,22 +1,80 @@
 package io.quarkus.smallrye.faulttolerance.runtime;
 
 import java.lang.reflect.Method;
+import java.util.Map;
+import java.util.Objects;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.Function;
 
 import javax.annotation.Priority;
-import javax.enterprise.context.Dependent;
 import javax.enterprise.inject.Alternative;
+import javax.inject.Singleton;
+
+import org.jboss.logging.Logger;
 
 import io.smallrye.faulttolerance.FaultToleranceOperationProvider;
 import io.smallrye.faulttolerance.config.FaultToleranceOperation;
 
-@Dependent
+@Singleton
 @Alternative
 @Priority(1)
 public class QuarkusFaultToleranceOperationProvider implements FaultToleranceOperationProvider {
+    private static final Logger LOG = Logger.getLogger(QuarkusFaultToleranceOperationProvider.class);
+
+    private final Map<CacheKey, FaultToleranceOperation> operationCache = new ConcurrentHashMap<>();
+    private final Function<CacheKey, FaultToleranceOperation> cacheFunction = new Function<CacheKey, FaultToleranceOperation>() {
+        @Override
+        public FaultToleranceOperation apply(CacheKey key) {
+            return createAtRuntime(key);
+        }
+    };
+
+    /**
+     * Called by SmallryeFaultToleranceRecorder to init the operation cache.
+     */
+    void init(Map<CacheKey, FaultToleranceOperation> operationCache) {
+        this.operationCache.putAll(operationCache);
+    }
 
     @Override
     public FaultToleranceOperation get(Class<?> beanClass, Method method) {
-        return FaultToleranceOperation.of(beanClass, method);
+        CacheKey key = new CacheKey(beanClass, method);
+        FaultToleranceOperation existing = operationCache.get(key);
+        return existing != null ? existing : operationCache.computeIfAbsent(key, cacheFunction);
     }
 
+    private FaultToleranceOperation createAtRuntime(CacheKey key) {
+        LOG.debugf("FaultToleranceOperation not found in the cache for %s creating it at runtime", key);
+        return FaultToleranceOperation.of(key.beanClass, key.method);
+    }
+
+    static class CacheKey {
+        private Class<?> beanClass;
+        private Method method;
+        private int hashCode;
+
+        public CacheKey(Class<?> beanClass, Method method) {
+            this.beanClass = beanClass;
+            this.method = method;
+            this.hashCode = Objects.hash(beanClass, method);
+        }
+
+        @Override
+        public boolean equals(Object o) {
+            if (this == o) {
+                return true;
+            }
+            if (o == null || getClass() != o.getClass()) {
+                return false;
+            }
+            CacheKey cacheKey = (CacheKey) o;
+            return Objects.equals(beanClass, cacheKey.beanClass) &&
+                    Objects.equals(method, cacheKey.method);
+        }
+
+        @Override
+        public int hashCode() {
+            return this.hashCode;
+        }
+    }
 }
--- a/extensions/smallrye-fault-tolerance/runtime/src/main/java/io/quarkus/smallrye/faulttolerance/runtime/SmallryeFaultToleranceRecorder.java
+++ b/extensions/smallrye-fault-tolerance/runtime/src/main/java/io/quarkus/smallrye/faulttolerance/runtime/SmallryeFaultToleranceRecorder.java
@@ -2,22 +2,27 @@ package io.quarkus.smallrye.faulttolerance.runtime;
 
 import java.lang.reflect.Method;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 
 import javax.enterprise.inject.spi.DeploymentException;
 
 import org.eclipse.microprofile.faulttolerance.exceptions.FaultToleranceDefinitionException;
 
+import io.quarkus.arc.Arc;
 import io.quarkus.runtime.annotations.Recorder;
 import io.smallrye.faulttolerance.config.FaultToleranceOperation;
 
 @Recorder
 public class SmallryeFaultToleranceRecorder {
 
-    public void validate(List<String> beanNames) {
+    public void createFaultToleranceOperation(Set<String> beanNames) {
         List<Throwable> allExceptions = new ArrayList<>();
+        Map<QuarkusFaultToleranceOperationProvider.CacheKey, FaultToleranceOperation> operationCache = new HashMap<>(
+                beanNames.size());
         for (String beanName : beanNames) {
             try {
                 ClassLoader classLoader = Thread.currentThread().getContextClassLoader();
@@ -26,11 +31,16 @@ public class SmallryeFaultToleranceRecorder {
                 }
                 Class<?> beanClass = Class.forName(beanName, true, classLoader);
 
-                for (Method method : getMethodsForValidation(beanClass)) {
+                for (Method method : getAllMethods(beanClass)) {
                     FaultToleranceOperation operation = FaultToleranceOperation.of(beanClass, method);
                     if (operation.isLegitimate()) {
                         try {
                             operation.validate();
+
+                            // register the operation at validation time to avoid re-creating it at runtime
+                            QuarkusFaultToleranceOperationProvider.CacheKey cacheKey = new QuarkusFaultToleranceOperationProvider.CacheKey(
+                                    beanClass, method);
+                            operationCache.put(cacheKey, operation);
                         } catch (FaultToleranceDefinitionException e) {
                             allExceptions.add(e);
                         }
@@ -61,16 +71,17 @@ public class SmallryeFaultToleranceRecorder {
                 throw deploymentException;
             }
         }
+        Arc.container().instance(QuarkusFaultToleranceOperationProvider.class).get().init(operationCache);
     }
 
-    private Set<Method> getMethodsForValidation(Class<?> beanClass) {
+    private Set<Method> getAllMethods(Class<?> beanClass) {
         Set<Method> allMethods = new HashSet<>();
         Class<?> currentClass = beanClass;
-        while (!currentClass.equals(Object.class)) {
+        while (currentClass != null && !currentClass.equals(Object.class)) {
             for (Method m : currentClass.getDeclaredMethods()) {
                 allMethods.add(m);
             }
-            currentClass = currentClass.getSuperclass();
+            currentClass = currentClass.getSuperclass(); // this will be null for interfaces
         }
         return allMethods;
     }
--- a/extensions/vertx-web/deployment/src/main/java/io/quarkus/vertx/web/deployment/VertxWebProcessor.java
+++ b/extensions/vertx-web/deployment/src/main/java/io/quarkus/vertx/web/deployment/VertxWebProcessor.java
@@ -119,43 +119,41 @@ class VertxWebProcessor {
         // Collect all business methods annotated with @Route and @RouteFilter
         AnnotationStore annotationStore = validationPhase.getContext().get(BuildExtension.Key.ANNOTATION_STORE);
         for (BeanInfo bean : validationPhase.getContext().beans().classBeans()) {
-            if (bean.isClassBean()) {
-                // NOTE: inherited business methods are not taken into account
-                ClassInfo beanClass = bean.getTarget().get().asClass();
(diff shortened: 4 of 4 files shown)


extensions/smallrye-fault-tolerance/runtime/src/main/java/io/quarkus/smallrye/faulttolerance/runtime/QuarkusFaultToleranceOperationProvider.java
```diff
@@ -1,22 +1,39 @@
...
+
 import io.smallrye.faulttolerance.FaultToleranceOperationProvider;
 import io.smallrye.faulttolerance.config.FaultToleranceOperation;
 
 @Dependent
 @Alternative
 @Priority(1)
 public class QuarkusFaultToleranceOperationProvider implements FaultToleranceOperationProvider {
+    private static final Logger LOG = Logger.getLogger(QuarkusFaultToleranceOperationProvider.class);
+    static final Map<String, FaultToleranceOperation> OPERATIONS = new ConcurrentHashMap<>();
```
mkouba: You can make the `QuarkusFaultToleranceOperationProvider` a `@Singleton` and then there is no need to use static field here. Instead, you can use `Arc.container().instance(QuarkusFaultToleranceOperationProvider.class).init(Map<String,FaultToleranceOperation> ops)` in the recorder.

extensions/smallrye-fault-tolerance/runtime/src/main/java/io/quarkus/smallrye/faulttolerance/runtime/QuarkusFaultToleranceOperationProvider.java
```diff
@@ -1,22 +1,39 @@
...
 
     @Override
     public FaultToleranceOperation get(Class<?> beanClass, Method method) {
+        String key = getCacheKey(beanClass, method);
+        FaultToleranceOperation operation = OPERATIONS.get(key);
+        return operation != null ? operation : createAtRuntime(beanClass, method, key);
```
mkouba: We could use `Map.computeIfAbsent()` if a `FaultToleranceOperation` is not found in the cache...

extensions/smallrye-fault-tolerance/runtime/src/main/java/io/quarkus/smallrye/faulttolerance/runtime/SmallryeFaultToleranceRecorder.java
```diff
@@ -66,11 +70,12 @@ public void validate(List<String> beanNames) {
     private Set<Method> getMethodsForValidation(Class<?> beanClass) {
         Set<Method> allMethods = new HashSet<>();
         Class<?> currentClass = beanClass;
-        while (!currentClass.equals(Object.class)) {
+        System.out.println(currentClass);
```
mkouba: No `System.out.print` please ;-)

mkouba (changes requested):

extensions/smallrye-fault-tolerance/runtime/src/main/java/io/quarkus/smallrye/faulttolerance/runtime/SmallryeFaultToleranceRecorder.java
loicmathieu: Still a draft :).
I will remove it

extensions/smallrye-fault-tolerance/runtime/src/main/java/io/quarkus/smallrye/faulttolerance/runtime/QuarkusFaultToleranceOperationProvider.java
loicmathieu: This was on my first implémentation, but as with the current one the operation should always be in the cache I don't know if it worth it.
It is here to catch potential bugs, and as it will be inside the hot code path I prefere to use `get` instead of `computeIfAbsent`. We could revisit it later if needed.

gsmet: My HV benchmarks taught me the hard way that using `get()` is faster if having the operation in the cache is the most common use case.

mkouba: +1

extensions/smallrye-fault-tolerance/runtime/src/main/java/io/quarkus/smallrye/faulttolerance/runtime/QuarkusFaultToleranceOperationProvider.java
loicmathieu: OK, will do.
And if I don't implements `computeIfAbsent` I should be able to use an HashMap directly instead of a concurrent one as the map will never be modified outside the enhancer.

extensions/smallrye-fault-tolerance/runtime/src/main/java/io/quarkus/smallrye/faulttolerance/runtime/QuarkusFaultToleranceOperationProvider.java
geoand: I remember having similar discussions on the PRs :). I say no `computeIfAbsent` anywhere on the hot path :)

extensions/smallrye-fault-tolerance/runtime/src/main/java/io/quarkus/smallrye/faulttolerance/runtime/QuarkusFaultToleranceOperationProvider.java
mkouba: Yep, an optimized immutable map would be a perfect fit but unfotunately there is no such impl in JDK.

extensions/smallrye-fault-tolerance/runtime/src/main/java/io/quarkus/smallrye/faulttolerance/runtime/QuarkusFaultToleranceOperationProvider.java
mkouba: Well, it depends on how many times it's invoked ;-). From my experience microbenchmarks are often tricky and the bottlenecks identified in the infra code might not be bottlenecks at all in a real app and vice versa.

loicmathieu: Well, with native image, stuff are a little different as there is no optimisation based on branch prediction (unless for paid customers using the new PGO mechanism of GraalVM). So computeIfAbsent could be more slow on native image than on JVM mode ...
Having said that, if I didn't saw any benchmark written with JMH that prove this I will deny it ;)

Anyway, I switch to regular HashMap and uses `get` as the operation should always be present (in fact I can move to a runtime exception if not in the cache if we prefere ... like a NPE :) )

extensions/smallrye-fault-tolerance/runtime/src/main/java/io/quarkus/smallrye/faulttolerance/runtime/SmallryeFaultToleranceRecorder.java
loicmathieu: Removed

gsmet (changes requested):
I added a couple of very minor comments.

mkouba:
It seems that a test from MP FT TCK failed:
```
[ERROR]   FallbackMethodGenericAbstractTest>Arquillian.arquillianBeforeClass:96 
```

loicmathieu:
@gsmet I updated the PR with your proposed changes.

@mkouba I'll check the TCK

loicmathieu:
@mkouba the issue from the TCK seems unrelated to my changes

```
java.lang.AssertionError: Unexpected exceptionjava.util.concurrent.ExecutionException: java.util.ServiceConfigurationError: io.smallrye.config.SmallRyeConfigFactory: io.quarkus.runtime.configuration.QuarkusConfigFactory not a subtype
```

loicmathieu:
@mkouba just testing the TCK from master and it fails with the same error

mkouba:
> @mkouba just testing the TCK from master and it fails with the same error

Hm, interesting. The last job does not seem to fail with this issue. Let me try this locally...

mkouba:
So I've just tried to run the TCKs on master and the result is BUILD SUCCESS.

loicmathieu:
@mkouba I will double-check ...

loicmathieu:
@mkouba master is still failing for me but it only have 4 failures where my branch have 5.

So I chase down the new failing test to `FallbackMethodGenericAbstractTest`, I will investiguate.

mkouba:
> master is still failing for me...

Hm, that's suspicious... 

> So I chase down the new failing test to FallbackMethodGenericAbstractTest

@loicmathieu yep, that's the one that fails on CI too.

loicmathieu:
@mkouba thanks to the TCK I chased down a bug when the annotation is not on the bean class but on one of it's parent; the previous implementation didn't work.

I switch back to using the `validationPhase.getContext().beans()` to scan all beans but filter the list of bean to create FT operations for by using Jandex to check if an FT annotation is present on a class (or one of it's methods) or on one of it's parents (or on one of the parent's method). And now all FT operations for all tests (main integration test and TCK) are correctly found at build time !

As the startup performance issue was due to loading all bean classes via reflection, filtering the bean at the validation stage like this still give me the same performance improvements (on my personal projects 250ms startup improvements over 2.5s, same kind of startup improvements can be found for the main integration test).

@gsmet I check all the other usage of the ValidationBuildItem and none are suffering from the same issue (none load all beans via refleciton). When checking other usages, I spotted a small unecessary if condition on `VertxWebProcessor` that I remove (the same check was done as a stream predicate and in a if statement when iterating the stream), it is on a separate commit.

I will merge the first three commits when we agree on the implementation.

extensions/smallrye-fault-tolerance/deployment/src/main/java/io/quarkus/smallrye/faulttolerance/deployment/SmallRyeFaultToleranceProcessor.java
```diff
@@ -187,13 +192,42 @@ public void transform(TransformationContext ctx) {
...
-        List<String> beanNames = new ArrayList<>();
-        for (BeanInfo bean : validationPhase.getContext().beans().classBeans()) {
-            beanNames.add(bean.getBeanClass().toString());
+    void validateFaultToleranceAnnotations(SmallryeFaultToleranceRecorder recorder,
+            ValidationPhaseBuildItem validationPhase) {
+        AnnotationStore annotationStore = validationPhase.getContext().get(BuildExtension.Key.ANNOTATION_STORE);
+        Set<String> beanNames = new HashSet<>();
+        IndexView index = combinedIndexBuildItem.getIndex();
+
+        for (BeanInfo info : validationPhase.getContext().beans()) {
```
mkouba: We're only interested in class beans which have at least one around invoke interceptor... so something like this could slightly improve the perf: `validationPhase.getContext().beans().stream().filter(BeanInfo::hasAroundInvokeInterceptors).iterator()`

extensions/smallrye-fault-tolerance/deployment/src/main/java/io/quarkus/smallrye/faulttolerance/deployment/SmallRyeFaultToleranceProcessor.java
```diff
@@ -187,13 +192,42 @@ public void transform(TransformationContext ctx) {
...
-    void validateFaultToleranceAnnotations(
-            ValidationPhaseBuildItem validationPhase, SmallryeFaultToleranceRecorder recorder) {
-        List<String> beanNames = new ArrayList<>();
-        for (BeanInfo bean : validationPhase.getContext().beans().classBeans()) {
-            beanNames.add(bean.getBeanClass().toString());
+    void validateFaultToleranceAnnotations(SmallryeFaultToleranceRecorder recorder,
+            ValidationPhaseBuildItem validationPhase) {
+        AnnotationStore annotationStore = validationPhase.getContext().get(BuildExtension.Key.ANNOTATION_STORE);
+        Set<String> beanNames = new HashSet<>();
+        IndexView index = combinedIndexBuildItem.getIndex();
```
mkouba: Unused variable?

extensions/smallrye-fault-tolerance/deployment/src/main/java/io/quarkus/smallrye/faulttolerance/deployment/SmallRyeFaultToleranceProcessor.java
```diff
@@ -187,13 +192,42 @@ public void transform(TransformationContext ctx) {
...
+        // then check on the methods
+        for (MethodInfo method : info.methods()) {
+            if (annotationStore.hasAnyAnnotation(method, FT_ANNOTATIONS)) {
+                return true;
+            }
+        }
+
+        // then check on the parent
+        DotName parentClassName = info.superName();
+        ClassInfo parentClassInfo = combinedIndexBuildItem.getIndex().getClassByName(parentClassName);
```
mkouba: I think that using `BeanArchiveIndexBuildItem` would be more appropriate in this case. Also the injected fields of `SmallRyeFaultToleranceProcessor` should be probably turned into injected params of the `SmallRyeFaultToleranceProcessor.build()` build step (they're not shared so it does not make much sense to use fields). I know that it's not a subject of this PR but it would deserve a small refactoring ;-).

extensions/smallrye-fault-tolerance/deployment/src/main/java/io/quarkus/smallrye/faulttolerance/deployment/SmallRyeFaultToleranceProcessor.java
```diff
@@ -187,13 +192,42 @@ public void transform(TransformationContext ctx) {
...
+        if (parentClassName.equals(DOTNAME_OBJECT) || parentClassInfo == null) {
```
mkouba: You can reuse `io.quarkus.arc.processor.DotNames.OBJECT`.

mkouba (changes requested):
Looks good but with this approach the rest client interfaces will not be processed, or?

loicmathieu:
> Looks good but with this approach the rest client interfaces will not be processed, or?

No, so I must lazily creates it at runtime (using a CHM and a computeIfAbsent as in my first implementation). 
In fact, the previous implementation also didn't validate it at build time (so my current code improve the startup of the application) as it was also based on `validationPhase.context().beans()`.

loicmathieu:
@mkouba I double-checked and in fact using `validationPhase.getContext().beans()` and not `validationPhase.getContext().beans().classBeans()` allow to access interfaces (they are indexed by Jandex). I tested on my project and it works for my rest client.

No need to cache the missing fault tolerance operations at runtime as they should all be in the cache. I will keep the WARNING as I expect users to open issue if they see WARNING in the log so if something is missing we can add it later to the cache.

extensions/smallrye-fault-tolerance/deployment/src/main/java/io/quarkus/smallrye/faulttolerance/deployment/SmallRyeFaultToleranceProcessor.java
loicmathieu: No possible or we will miss the annotations on interfaces (like the one for rest client)

extensions/smallrye-fault-tolerance/deployment/src/main/java/io/quarkus/smallrye/faulttolerance/deployment/SmallRyeFaultToleranceProcessor.java
loicmathieu: Yes, removed, sorry.

extensions/smallrye-fault-tolerance/deployment/src/main/java/io/quarkus/smallrye/faulttolerance/deployment/SmallRyeFaultToleranceProcessor.java
loicmathieu: Switch to `BeanArchiveIndexBuildItem`.
Refactor to inject everything via parameters.

extensions/smallrye-fault-tolerance/deployment/src/main/java/io/quarkus/smallrye/faulttolerance/deployment/SmallRyeFaultToleranceProcessor.java
loicmathieu: Done

mkouba:
Hm, `validationPhase.getContext().beans()` also includes synthetic beans and AFAIK we do register synthetic beans for rest client interfaces iff they are annotated with `@RegisterRestClient`. In other words, yes this should cover most of the rest client interfaces but not those accessed directly through a builder.

gsmet (changes requested):
Please squash these commits into one once you're done (and this review can be dismissed as soon as it's done).

loicmathieu:
@mkouba 

> In other words, yes this should cover most of the rest client interfaces but not those accessed directly through a builder.

Will FT annotations works in this case ? If yes, so I must update the implementation to put in cache the missing FTOperations at runtime ...

mkouba:
> Will FT annotations works in this case 

I think they should. It is not supported by the CDI spec - interceptor bindings on interfaces - but our Rest Client impl (previously SmallRye and now RESTEasy) resolves/invokes the CDI interceptors manually.

loicmathieu:
@mkouba OK, I will switch back to using `computeIfAbsent` to allow runtime creation of FTOperations to be stored inside the cache.

loicmathieu:
@mkouba I update my PR to store FT operations at runtime if not already done inside the build step.
Is it OK for you now ?

mkouba (changes requested):
I added few minor comments. Otherwise looks good.

mkouba:
@loicmathieu And thanks for your patience. Sometimes it takes time to find the correct solution ;-).

extensions/smallrye-fault-tolerance/runtime/src/main/java/io/quarkus/smallrye/faulttolerance/runtime/QuarkusFaultToleranceOperationProvider.java
```diff
@@ -1,22 +1,50 @@
...
 import io.smallrye.faulttolerance.FaultToleranceOperationProvider;
 import io.smallrye.faulttolerance.config.FaultToleranceOperation;
 
-@Dependent
+@Singleton
 @Alternative
 @Priority(1)
 public class QuarkusFaultToleranceOperationProvider implements FaultToleranceOperationProvider {
+    private static final Logger LOG = Logger.getLogger(QuarkusFaultToleranceOperationProvider.class);
+    private Map<String, FaultToleranceOperation> operationCache;
```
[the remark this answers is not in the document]
loicmathieu: Ah sorry, it was in my first implementation, I forgot to switch it.

extensions/smallrye-fault-tolerance/runtime/src/main/java/io/quarkus/smallrye/faulttolerance/runtime/QuarkusFaultToleranceOperationProvider.java
```diff
@@ -1,22 +1,50 @@
...
+        return operationCache.computeIfAbsent(key, new Function<String, FaultToleranceOperation>() {
+            @Override
+            public FaultToleranceOperation apply(String key) {
+                return createAtRuntime(beanClass, method, key);
+            }
+        });
+    }
+
+    private FaultToleranceOperation createAtRuntime(Class<?> beanClass, Method method, String key) {
+        LOG.warnf("FaultToleranceOperation not found in the cache for %s creating it at runtime", key);
```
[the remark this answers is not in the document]
loicmathieu: Yes, now that we define that it can occurs and it's not a bug, INFO seems better.

loicmathieu:
@mkouba it should be OK now, I also squash the commit but I keep separate the small no-op vertx change.

extensions/smallrye-fault-tolerance/runtime/src/main/java/io/quarkus/smallrye/faulttolerance/runtime/QuarkusFaultToleranceOperationProvider.java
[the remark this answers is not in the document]
gsmet: Do we really need an INFO if it's the normal behavior? Or it's not a normal behavior?

If it's normal behavior, it should just be debug.

extensions/smallrye-fault-tolerance/runtime/src/main/java/io/quarkus/smallrye/faulttolerance/runtime/QuarkusFaultToleranceOperationProvider.java
```diff
@@ -1,22 +1,50 @@
...
+     * Called by SmallryeFaultToleranceRecorder to init the operation cache.
+     */
+    void init(Map<String, FaultToleranceOperation> operationCache) {
+        this.operationCache = operationCache;
+    }
 
     @Override
     public FaultToleranceOperation get(Class<?> beanClass, Method method) {
+        String key = getCacheKey(beanClass, method);
+        return operationCache.computeIfAbsent(key, new Function<String, FaultToleranceOperation>() {
```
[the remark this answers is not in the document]
gsmet: I don't see why we don't make the function a constant? Is there something forbidding it?

extensions/smallrye-fault-tolerance/runtime/src/main/java/io/quarkus/smallrye/faulttolerance/runtime/QuarkusFaultToleranceOperationProvider.java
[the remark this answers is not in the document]
loicmathieu: It will be logged only once, and should arrives in very rare conditions (using the builder aproach for RestClient is one that we thin of, this aproach is not documented on the Quarkus site).
But I'm OK with DEBUG ...

extensions/smallrye-fault-tolerance/runtime/src/main/java/io/quarkus/smallrye/faulttolerance/runtime/QuarkusFaultToleranceOperationProvider.java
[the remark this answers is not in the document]
mkouba: Because we use the method arguments inside the function -> `createAtRuntime(beanClass, method, key)`.

mkouba (approved):

extensions/smallrye-fault-tolerance/runtime/src/main/java/io/quarkus/smallrye/faulttolerance/runtime/QuarkusFaultToleranceOperationProvider.java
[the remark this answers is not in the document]
loicmathieu: It's possible if I used a CompositeKey with beanClass and method instead of a String. So everything needed to create the operation will be on the key and the Function could be a constant.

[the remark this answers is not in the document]
gsmet: Yeah, I was thinking about that.

TBH, the best solution here is to do what I suggested first: we do a first `get()` call and only call `computeIfAbsent` if we have nothing. As previously mentioned, I did some extensive benchmarks on HV and it's significantly faster if this is called often (and I suppose it will be called for every call to a function that is marked with FT annotations or am I missing something?). And this will avoid creating the function if this is in the cache.

extensions/smallrye-fault-tolerance/runtime/src/main/java/io/quarkus/smallrye/faulttolerance/runtime/QuarkusFaultToleranceOperationProvider.java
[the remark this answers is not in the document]
gsmet: If it's something that can happen and doesn't need any action by the user then debug is enough IMHO.

extensions/smallrye-fault-tolerance/runtime/src/main/java/io/quarkus/smallrye/faulttolerance/runtime/QuarkusFaultToleranceOperationProvider.java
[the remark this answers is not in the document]
mkouba: That would not help much - you'll have to create a new key object for each `get()` anyway...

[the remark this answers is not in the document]
loicmathieu: Previsouly I created a new String for each object so it should be the same ...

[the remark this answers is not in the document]
loicmathieu: And the key object just hold two references for existing object so should be small regarding a new string ...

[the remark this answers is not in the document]
loicmathieu: @gsmet @mkouba updated with a composite CacheKey to avoid re-creating the Function on each call and implemented @gsmet optimization on the CHM.

If it's OK I'll squash the last commit and we'll be ready to go :)

extensions/smallrye-fault-tolerance/runtime/src/main/java/io/quarkus/smallrye/faulttolerance/runtime/QuarkusFaultToleranceOperationProvider.java
```diff
@@ -19,33 +20,73 @@
...
+    static class CacheKey {
+        private Class<?> beanClass;
+        private Method method;
+
+        public CacheKey(Class<?> beanClass, Method method) {
+            this.beanClass = beanClass;
+            this.method = method;
+        }
+
+        public Class<?> getBeanClass() {
```
mkouba: What are the getters/setters for?

loicmathieu: Useless, removed, sorry.

mkouba:
Looks good. Could you squash the commits? Thanks!

extensions/smallrye-fault-tolerance/runtime/src/main/java/io/quarkus/smallrye/faulttolerance/runtime/QuarkusFaultToleranceOperationProvider.java
```diff
@@ -1,22 +1,76 @@
...
+            if (o == null || getClass() != o.getClass())
+                return false;
+            CacheKey cacheKey = (CacheKey) o;
+            return Objects.equals(beanClass, cacheKey.beanClass) &&
+                    Objects.equals(method, cacheKey.method);
+        }
+
+        @Override
+        public int hashCode() {
+            return Objects.hash(beanClass, method);
```
gsmet: It's minor but I would compute the hashCode in the constructor if it can become a hot path. That's what we do in HV.

That allows to cache the hashCode once and for all for all the items that are in the map.

extensions/smallrye-fault-tolerance/runtime/src/main/java/io/quarkus/smallrye/faulttolerance/runtime/QuarkusFaultToleranceOperationProvider.java
```diff
@@ -1,22 +1,76 @@
...
+            this.beanClass = beanClass;
+            this.method = method;
+        }
+
+        @Override
+        public boolean equals(Object o) {
+            if (this == o)
+                return true;
+            if (o == null || getClass() != o.getClass())
+                return false;
```
gsmet: If you end up taking into account my comment below, please use curly braces for the two statements above.

gsmet:
I added two more small comments. They are minor but since you are working on optimizing things, it might be worth it.

extensions/smallrye-fault-tolerance/runtime/src/main/java/io/quarkus/smallrye/faulttolerance/runtime/QuarkusFaultToleranceOperationProvider.java
mkouba: +1 and also make the fields final please.

extensions/smallrye-fault-tolerance/runtime/src/main/java/io/quarkus/smallrye/faulttolerance/runtime/QuarkusFaultToleranceOperationProvider.java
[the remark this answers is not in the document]
gsmet: @mkouba you won't avoid creating the key but you can avoid creating the function.

extensions/smallrye-fault-tolerance/runtime/src/main/java/io/quarkus/smallrye/faulttolerance/runtime/QuarkusFaultToleranceOperationProvider.java
gsmet: Actually, I'm torn on that, not having the fields final make the JDK do some weird optimizations :). I removed them from the key objects in HV after Sanne gave me the idea to test it and benchmark it.

mkouba: So I'm not following. Was it better or worse ;-). In any case, the key should be immutable and it's not.

extensions/smallrye-fault-tolerance/runtime/src/main/java/io/quarkus/smallrye/faulttolerance/runtime/QuarkusFaultToleranceOperationProvider.java
loicmathieu: Done, I should update my IDEA templates as I prefere having the curly braces ...

extensions/smallrye-fault-tolerance/runtime/src/main/java/io/quarkus/smallrye/faulttolerance/runtime/QuarkusFaultToleranceOperationProvider.java
loicmathieu: @gmset good catch on the cache of the hashCode, I'm not used to do it but totaly worth it :)

Regarding final fields, I add final to all three fields, I can remove it.

@gmset it's very interesting, I was not aware of performance penalty on final field writes inside a constructor so I make some research and yes, there is a small performance penalty because the VM emmit a memory barrier when one of the field written inside a constructor is final. The cost of the barrier is minimal (5ns on ARM, almost nefligeable on x86). As always some good explaination/experiments on this can be found on Alexey Shipilev blog: https://shipilev.net/blog/2014/all-fields-are-final/

I you want, I can remove the three final modifier ...

loicmathieu:
@gsmet @mkouba I take into account all yours request except for the final fields: I keep them. Please tell me if I need to remove them. Else, ready to be merged :)

extensions/smallrye-fault-tolerance/runtime/src/main/java/io/quarkus/smallrye/faulttolerance/runtime/QuarkusFaultToleranceOperationProvider.java
gsmet: In my HV benchmarks (and I have no idea why), removing the final from the fields did help. In HV, the only classes that don't have final fields are the classes used as keys. It's a very hot path though.

loicmathieu: OK, to please @gsmet ;) I remove the final modifier.
Ready to be merged

gsmet (approved):

gsmet:
@loicmathieu it would be interesting to check what your flame graph looks like now. I know I had some bad surprises with the `equals` from the reflection API being slow in the past. But I merged and let's iterate from there!

loicmathieu:
@gsmet on the mailing list I send flamegraph with a fix like this one (with a CacheKey) and it shows that FT was no longer an issue. I'll test again with this fix to be 100% sure (probably next week) and send update to the mailing list.

Issue #6754, opened 2020-01-23
Incorrect resources included in UberJar with multiple dependencies


bcluap:
**Describe the bug**
If you have a pom which pulls in multiple jars (each has some rest services) as well as the Quarkus features, and each individual jar has a src/main/resources/application.properties, then the application.properties from the first dependency is put in the uberjar instead of one on the uberjar project itself

**Expected behavior**
The uberjar projects src/main/resources/application.properties should take precedence and be put in the jar file

**Actual behavior**
The src/main/resources/application.properties from the first dependency is put in the uberjar

**To Reproduce**
git clone https://github.com/bcluap/quarkus.git
cd quarkus/uberjar
mvn install
look in the jar getting-started-uberjar/target/uberjar-runner.jar and see that application.properties is the one from getting-started1 but should be from uberjar

**Configuration**
```properties
maven 3.6.1
Latest Quarkus master

**Environment (please complete the following information):**
 - Output of `uname -a` or `ver`: 
Linux PCBLaptop 4.15.0-70-generic #79-Ubuntu SMP Tue Nov 12 10:36:11 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux
 - Output of `java -version`: 
openjdk version "11.0.5" 2019-10-15
OpenJDK Runtime Environment (build 11.0.5+10-post-Ubuntu-0ubuntu1.118.04)
OpenJDK 64-Bit Server VM (build 11.0.5+10-post-Ubuntu-0ubuntu1.118.04, mixed mode, sharing)
 - GraalVM version (if different from Java): 
 - Quarkus version or git rev: 
master

**Additional context**

damirnovosel:
Hi guys. Will this issue be addressed in foreseeable time? Is there a way to work around it, such as by explicitly defining application.properties file in maven config or something like that?

Thanks,
Damir

geoand:
Closing for lack of feedback

PR #6777, opened 2020-01-24
Fix issue #6754

bcluap:
Local repository resources should be able to overwrite existing files from the project dependencies


gastaldi:
This PR seems invalid and outdated, I'll close it for the time being.

PR #6782, opened 2020-01-24
Allow multiple caching annotations on a single method

### LINKED ISSUE #6682: Allow multiple caching annotations on a single method
The `quarkus-cache` extension (which was just merged into `master`) currently has a limitation of one `@Cache*` annotation per method.

This means none of the following code snippets are allowed for now:

```
@CacheInvalidate(cacheName = "foo")
@CacheResult(cacheName = "foo")
public Object load(String key) {
    return new Object();
}
```

```
@CacheInvalidateAll(cacheName = "foo")
@CacheResult(cacheName = "foo")
public Object load(String key) {
    return new Object();
}
```

```
@CacheInvalidate(cacheName = "foo")
@CacheInvalidate(cacheName = "bar")
public void load(String key) {
    // ...
}
```

I will submit a PR that removes this limitation.

gwenneg:
Fixes #6682

- All `AugmentedCache*` classes have been replaced with a much simpler code using `CacheInterceptorBinding` and `CacheInterceptor`.
- It is now possible to combine multiple caching annotations on a single method. The operations order will always be the same: `CacheInvalidateAll` first, then `CacheInvalidate` and finally `CacheResult`.
- Using a combination of `@CacheInvalidate` and `@CacheResult` on a method is equivalent to the old `@CacheStore` annotation that existed at the beginning of this extension development.
- The only restriction is a one `@CacheResult` annotation limit per method. Everything else is allowed in terms of annotations combination and/or caches involved in a single method.


gwenneg:
cc @emmanuelbernard, @galderz

gwenneg:
@gsmet Is creating a new `area/cache` label enough to enable auto-labelling of PRs that include changes for the `quarkus-cache` extension?

geoand:
@gwenneg you also need to add it to: https://github.com/quarkusio/quarkus/blob/master/.github/boring-cyborg.yml

gwenneg:
Thanks @geoand!

extensions/cache/runtime/src/main/java/io/quarkus/cache/CacheInvalidate.java
```diff
@@ -1,31 +1,43 @@
...
 
     /**
      * The name of the cache.
      */
     @Nonbinding
     String cacheName();
+
+    @Retention(RetentionPolicy.RUNTIME)
+    @Target(ElementType.METHOD)
+    @interface CacheInvalidateContainer {
```
gastaldi: Name it `List` to match the same pattern used in other places (including Bean validation)

extensions/cache/runtime/src/main/java/io/quarkus/cache/CacheInvalidateAll.java
```diff
@@ -1,27 +1,39 @@
...
 
     /**
      * The name of the cache.
      */
     @Nonbinding
     String cacheName();
+
+    @Retention(RetentionPolicy.RUNTIME)
+    @Target(ElementType.METHOD)
+    @interface CacheInvalidateAllContainer {
```
gastaldi: Same here

extensions/cache/runtime/src/main/java/io/quarkus/cache/CacheInvalidate.java
gwenneg: Done, thanks for reviewing this!

extensions/cache/runtime/src/main/java/io/quarkus/cache/runtime/CacheInterceptor.java
```diff
@@ -0,0 +1,88 @@
...
+
+        CacheResult cacheResult = method.getDeclaredAnnotation(CacheResult.class);
+        if (cacheResult != null) {
+            Object key = getCacheKey(context, cacheResult.cacheName());
+            if (LOGGER.isDebugEnabled()) {
+                LOGGER.debugf("Loading entry with key [%s] from cache [%s]", key, cacheResult.cacheName());
+            }
+            CaffeineCache cache = cacheRepository.getCache(cacheResult.cacheName());
+            return cache.get(key, () -> context.proceed(), cacheResult.lockTimeout());
+        }
```
gsmet: I wonder how efficient this code will be considering it will call `declaredAnnotations()` (which is a synchronized method) 3 times.

extensions/cache/runtime/src/main/java/io/quarkus/cache/runtime/CacheInterceptor.java
```diff
@@ -0,0 +1,88 @@
...
+    private Object getCacheKey(InvocationContext context, String cacheName) {
+        // If the method doesn't have any parameter, then a unique default key is generated and used.
+        if (context.getParameters().length == 0) {
+            return CacheKeyBuilder.buildDefault(cacheName);
+        } else {
+            List<Object> cacheKeyElements = new ArrayList<>();
+            // If at least one of the method parameters is annotated with @CacheKey, then the key is composed of all
+            // @CacheKey-annotated parameters.
+            for (int i = 0; i < context.getParameters().length; i++) {
+                if (context.getMethod().getParameters()[i].isAnnotationPresent(CacheKey.class)) {
```
gsmet: Same here, I'm pretty sure it will end up calling `declaredAnnotations()` again.

gsmet (changes requested):
Needs a rebase as I renamed the guide and I added one question.

@mkouba your input might be interesting on how to best design interceptors for performance.

extensions/cache/runtime/src/main/java/io/quarkus/cache/runtime/CacheInterceptor.java
gwenneg: The `synchronized` block should only be called once since the declared annotations will already be parsed when the 2nd and 3rd calls happen, but that's still a very interesting question. Thanks for taking a look at this!

gwenneg:
I just pushed a small improvement regarding how cache keys are built but that's still based on reflection.

extensions/cache/runtime/src/main/java/io/quarkus/cache/runtime/CacheInterceptor.java
gwenneg: FYI, I tried to avoid using reflection and to do things at build time using annotations transformation originally, but there's a limitation in the current Quarkus annotation transformation logic that blocked me (I already had a discussion with @mkouba about it). I'd be happy to remove all reflection calls if that's possible :)

emmanuelbernard:
@karesti can you also review this one as you are slowly erasing Galder from history ;)

docs/src/main/asciidoc/cache.adoc
```diff
@@ -288,7 +288,8 @@ If the method has one or more arguments, the key computation is done from all th
 This annotation can also be used on a method with no arguments, a default key derived from the cache name is generated in that case.
 If the key does not identify any cache entry, nothing will happen.
 
-You can only use one of the cache operations (and this annotation) on a given method: `@CacheResult`, `@CacheInvalidate` or `@CacheInvalidateAll`.
+This annotation can be combined with multiple other caching annotations on a single method.
```
emmanuelbernard: I wonder if you need to mention that, do you think it helps the reader or is it just a remnent of the old implementation that you want to emphasize the change of?

docs/src/main/asciidoc/cache.adoc
```diff
@@ -299,7 +300,8 @@ If the `@CacheResult` or `@CacheInvalidate` annotations are used on a method wit
 
 When a method annotated with `@CacheInvalidateAll` is invoked, Quarkus will remove all entries from the cache.
 
-You can only use one of the cache operations (and this annotation) on a given method: `@CacheResult`, `@CacheInvalidate` or `@CacheInvalidateAll`.
+This annotation can be combined with multiple other caching annotations on a single method.
```
emmanuelbernard: same here

extensions/cache/runtime/src/main/java/io/quarkus/cache/runtime/CacheInterceptor.java
emmanuelbernard: So I'm a bit disappointed in the sense that witht he build time knowledge, we should be able to add an interceptor with the right state. We should know the annotations effectively present and even the cache name state and it should be burnt into the interceptor instance. I don't know how much of this is doable with the CDI model but that would be a nice improvement. Is that the conversation you had with @mkouba @gwenneg ?

emmanuelbernard: But as a temp measure, it looks like the synchronize call will only happen the first time the service class is requested its annotations. We could do this in a static_init block to do it at "startup time" if we feel its a problem to pay the cost lazily.

docs/src/main/asciidoc/cache.adoc
gwenneg: The first sentence is only there to introduce the second one about the execution order. I think the execution order is an important detail as it allows the user to force a cache value refresh for example (by combining `@CacheInvalidate` and `@CacheResult`). But maybe I should go one step further and write down that specific example in the doc.

extensions/cache/runtime/src/main/java/io/quarkus/cache/runtime/CacheInterceptor.java
gwenneg: I talked with @mkouba about the following annotation that I tried to introduce through transformation at build time:
```
@CacheInterceptorBinding(
    cacheInvalidateAll = { @CacheInvalidateAll(cacheName = "foo"), @CacheInvalidateAll(cacheName = "bar") },
    cacheInvalidate = { @CacheInvalidate(cacheName = "foo", @CacheInvalidate(cacheName = "bar") },
    cacheResult = @CacheResult(cacheName = "foo")
)
```
With such an annotation, most of the reflection calls wouldn't be required anymore (there would still be some reflection involved to identify the `@CacheKey`-annotated parameters).

gwenneg: Maybe handling everything in a single interceptor is not a reasonnable idea. Another solution would be to keep one interceptor by caching operation and change the priorities to make sure the execution order is always the same.

gwenneg:
@emmanuelbernard I don't know why but some of my answers to your comments are not showing up in the `Conversation` tab of this PR, but you can read them from the `Files changed` tab.

emmanuelbernard:
> @emmanuelbernard ~I don't know why but some of my answers to your comments are not showing up in the `Conversation` tab of this PR, but you can read them from the `Files changed` tab.~
> 
> Here's what I wrote:
> 
> I talked with @mkouba about the following annotation that I tried to introduce through transformation at build time:
> 
> ```
> @CacheInterceptorBinding(
>     cacheInvalidateAll = { @CacheInvalidateAll(cacheName = "foo"), @CacheInvalidateAll(cacheName = "bar") },
>     cacheInvalidate = { @CacheInvalidate(cacheName = "foo", @CacheInvalidate(cacheName = "bar") },
>     cacheResult = @CacheResult(cacheName = "foo")
> )
> ```
> 
> With such an annotation, most of the reflection calls wouldn't be required anymore (there would still be some reflection involved to identify the `@CacheKey`-annotated parameters).
> 
> Maybe handling everything in a single interceptor is not a reasonnable idea. Another solution would be to keep one interceptor by caching operation and change the priorities to make sure the execution order is always the same. This would be pretty close from what's currently on `master`.

you would still need the `method.getDeclaredAnnotations()` call which in theory could be entirely avoided if interceptors could receive state and be call side specific.

extensions/cache/runtime/src/main/java/io/quarkus/cache/runtime/CacheInterceptor.java
gsmet: > it looks like the synchronize call will only happen the first time the service class is requested its annotations

That's not what I see as AFAICS `declaredAnnotations()` will always be called when calling `getDeclaredAnnotationsByType()`. It will be very fast after the first invocation but you still have the synchronized part. Unless I miss something obvious?

gwenneg:
I don't think `method.getDeclaredAnnotations()` would still be called since `@CacheInterceptorBinding` and everything it contains would be retrieved with `InterceptorBindings.getInterceptorBindings(context)` (but I haven't checked what this method does internally).

gwenneg:
That being said, I completely agree that interceptors should be call side specific. I'll rework this PR tonight and submit a better version.

gsmet:
@gwenneg hmmm, but `intercept()` will be called for each intercepted call, won't it? That's my concern.

mkouba:
So I haven't seen the details of this PR but (1) an interceptor instance is associated with the intercepted bean instance, (2) `@AroundInvoke` is called for each invocation of an intercepted method (unless a another interceptor breaks the chain).

gwenneg:
> 
> 
> @gwenneg hmmm, but `intercept()` will be called for each intercepted call, won't it? That's my concern.

Yes, it will. I thought your first comment was about the `declaredAnnotations()` calls happening within a single interception.

Anyway, reflection is clearly not the way to go performance-wise, I'll move back to build-time computation as much as possible in the next version of this PR.

gwenneg:
I have a working solution locally without a single drop of reflection. The entire code that relied on reflection (including the @CacheKey parameters resolution) is now run at build time. It’ll be ready tomorrow.

gwenneg:
This PR is ready for another review.

- We're back to one interceptor per caching operation, but each interceptor has a different priority to guarantee the operations execution order: `CacheInvalidateAll` first, then `CacheInvalidate` and finally `CacheResult`.
- Reflection was completely removed from the runtime code. It's been replaced with build time code using Jandex in `CacheAnnotationsTransformer`.
- `@CacheKey`-annotated method parameters are identified at build time and their position within the method parameters list is stored in the `cacheKeyParameterPositions` parameter of `CacheInvalidateInterceptorBinding` and `CacheResultInterceptorBinding`. These positions are then used at runtime to build the cache key in `CacheInterceptor`. From my understanding and tests, Jandex always provides the same positions for a method parameters, but it would be great to confirm this very important detail.
- Since `@CacheInvalidate` and `@CacheInvalidateAll` are `@Repeatable`, the corresponding interceptors bindings (`CacheInvalidateInterceptorBinding` and `CacheInvalidateAllInterceptorBinding`) are also `@Repeatable`. @mkouba I'll need your approval on this particular detail. Everything's tested and is working fine but I'd like to be sure it's ok from a CDI spec point of view.
- @emmanuelbernard I didn't change the guide or Javadoc since your last review, I'll need you confirmation if you want to remove the sentences you commented earlier.

The two following points were in my initial comment and are still true:
- Using a combination of `@CacheInvalidate` and `@CacheResult` on a method is equivalent to the old `@CacheStore` annotation that existed at the beginning of this extension development.
- The only restriction is a one `@CacheResult` annotation limit per method. Everything else is allowed in terms of annotations combination and/or caches involved in a single method.

emmanuelbernard:
I find the sentences make the documentation harder to read and grasp as I need to figure our why you wrote these sentences. But if you feel strongly about it, we can keep them.

gwenneg:
@emmanuelbernard I like the idea of keeping the doc easy to read/understand so I removed the sentences there. They are still present in the annotations Javadoc though since I think the caching operations execution order has to be mentioned somewhere.

gsmet (approved):

gsmet:
Merged, thanks!

Issue #6786, opened 2020-01-25
Can't compile project using Java 14


SR-Lut3t1um:
I downloaded a new project using the "start-coding" option on the website. After updating gradle 6 -> 6.1.1 and switching the sdk form 11->14 gradle breaks. After going back to java 13 it work's again.

Exception:
```
java.lang.NoClassDefFoundError: Could not initialize class org.codehaus.groovy.runtime.InvokerHelper
        at org.gradle.internal.extensibility.DefaultExtraPropertiesExtension.<init>(DefaultExtraPropertiesExtension.java:29)
        at org.gradle.internal.extensibility.DefaultConvention.<init>(DefaultConvention.java:49)
        at org.gradle.internal.extensibility.ExtensibleDynamicObject.<init>(ExtensibleDynamicObject.java:60)
        at org.gradle.internal.instantiation.generator.MixInExtensibleDynamicObject.<init>(MixInExtensibleDynamicObject.java:35)
        at org.gradle.initialization.DefaultSettings_Decorated.getAsDynamicObject(Unknown Source)
        at org.gradle.initialization.SettingsFactory.createSettings(SettingsFactory.java:58)
        at org.gradle.initialization.ScriptEvaluatingSettingsProcessor.process(ScriptEvaluatingSettingsProcessor.java:61)
        at org.gradle.initialization.PropertiesLoadingSettingsProcessor.process(PropertiesLoadingSettingsProcessor.java:38)
        at org.gradle.initialization.SettingsEvaluatedCallbackFiringSettingsProcessor.process(SettingsEvaluatedCallbackFiringSettingsProcessor.java:34)
        at org.gradle.initialization.RootBuildCacheControllerSettingsProcessor.process(RootBuildCacheControllerSettingsProcessor.java:36)
        at org.gradle.initialization.BuildOperationSettingsProcessor$2.call(BuildOperationSettingsProcessor.java:50)
        at org.gradle.initialization.BuildOperationSettingsProcessor$2.call(BuildOperationSettingsProcessor.java:47)
        at org.gradle.internal.operations.DefaultBuildOperationExecutor$CallableBuildOperationWorker.execute(DefaultBuildOperationExecutor.java:416)
        at org.gradle.internal.operations.DefaultBuildOperationExecutor$CallableBuildOperationWorker.execute(DefaultBuildOperationExecutor.java:406)
[...]

PR #6788, opened 2020-01-25
Upgrade to ASM 7.3.1

gunnarmorling:
Note this solves the problem I had with JDK 14, not the other one reported on #6786 which seems related to Gradle/Groovy and likely needs another fix.

CC @gsmet

### DIFF
--- a/bom/runtime/pom.xml
+++ b/bom/runtime/pom.xml
@@ -63,7 +63,7 @@
         <jaxb-api.version>2.3.1</jaxb-api.version>
         <jboss-jaxrs-api_2.1_spec.version>2.0.1.Final</jboss-jaxrs-api_2.1_spec.version>
         <jboss-jaxb-api_2.3_spec.version>2.0.0.Final</jboss-jaxb-api_2.3_spec.version>
-        <asm.version>7.1</asm.version>
+        <asm.version>7.3.1</asm.version>
         <commons-io.version>2.6</commons-io.version>
         <jboss-metadata-web.version>11.0.0.Final</jboss-metadata-web.version>
         <maven-core.version>3.5.4</maven-core.version>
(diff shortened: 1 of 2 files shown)


gsmet (changes requested):
Thanks but please let's avoid the emojis in issue title and commit message. I'm also wondering if we should change the versions used when generating the classes (the ASM constants with the version).

We probably should upgrade Gizmo to use that version too.

gunnarmorling:
> let's avoid the emojis in issue title and commit message.

Why?

> change the ASM constants with the version
> We probably should upgrade Gizmo to use that version too.

Do you have any pointers where to look into for these things?

geoand:
> change the ASM constants with the version We probably should upgrade Gizmo to use that version too.
> Do you have any pointers where to look into for these things?

This should be it: https://github.com/quarkusio/gizmo/blob/master/pom.xml#L38

geoand:
> let's avoid the emojis in issue title and commit message.
> Why?
My guess is that it doesn't look in the generated changelog. @gsmet can of course confirm or deny

gastaldi:
How about adding it to the dependabot descriptor?

gastaldi:
I created https://github.com/quarkusio/gizmo/pull/24 bumping ASM and Jandex to their latest versions

gsmet:
> Why?

Because I don't want the changelog to look like a Christmas tree. And if we start accepting them, where do we set the limit? Typically in this case, it's totally useless, it's just visual noise. So let's drop it please.

AFAICS the latest version of ASM is 7.3.1 so we should upgrade to this one instead. I upgraded Gizmo to use it and set up some additional CI.

The version of the API is still 7 so we can just upgrade in place without further changes.

gunnarmorling:
Ok, will send an update for this PR later today, updating to ASM 7.3.1 and
replacing the emoji with a Christm..., eh, removing the emoji ;)

>

gunnarmorling:
Reworked the commit and updated this PR 🤟.

gsmet:
Merged, thanks.

PR #6791, opened 2020-01-25
Flyway - Add an out-of-order configuration knob

### LINKED ISSUE #6790: Expose Configuration for Out-of-Order Flyway Migrations
**Description**
By default, Flyway will run all migrations in version order and throw an error if a new migration is introduced with a number lower than the last applied migration.  There is a boolean configuration option to allow Flyway to run migrations out of order.

**Implementation ideas**
Would it be possible to expose this as a configuration property in `application.properties` similar to how, for example, `quarkus.flyway.migrate-at-start` works.

machi1990:
Fixes https://github.com/quarkusio/quarkus/issues/6790

### DIFF
--- a/extensions/flyway/runtime/src/test/java/io/quarkus/flyway/runtime/FlywayCreatorTest.java
+++ b/extensions/flyway/runtime/src/test/java/io/quarkus/flyway/runtime/FlywayCreatorTest.java
@@ -2,6 +2,8 @@ package io.quarkus.flyway.runtime;
 
 import static java.util.Arrays.asList;
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import java.util.Arrays;
 import java.util.List;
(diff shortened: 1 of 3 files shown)


extensions/flyway/runtime/src/test/java/io/quarkus/flyway/runtime/FlywayCreatorTest.java
```diff
@@ -165,6 +165,18 @@ void testValidateOnMigrate() {
         assertEquals(runtimeConfig.validateOnMigrate, true);
     }
 
+    @Test
+    @DisplayName("outOfOrder is correctly set")
+    void testOutOfOrder() {
+        runtimeConfig.outOfOrder = false;
+        creator = new FlywayCreator(runtimeConfig, buildConfig);
+        assertEquals(false, createdFlywayConfig().isOutOfOrder());
```
gastaldi: Not a big deal but you can use `assertFalse` here

extensions/flyway/runtime/src/test/java/io/quarkus/flyway/runtime/FlywayCreatorTest.java
```diff
@@ -165,6 +165,18 @@ void testValidateOnMigrate() {
...
+
+        runtimeConfig.outOfOrder = true;
+        creator = new FlywayCreator(runtimeConfig, buildConfig);
+        assertEquals(true, createdFlywayConfig().isOutOfOrder());
```
gsmet: And `assertTrue` here.

gsmet (changes requested):
Looks good. We let you fix the assertions?

extensions/flyway/runtime/src/test/java/io/quarkus/flyway/runtime/FlywayCreatorTest.java
machi1990: Yeap, let's do it!

extensions/flyway/runtime/src/test/java/io/quarkus/flyway/runtime/FlywayCreatorTest.java
machi1990: Let's do it!

gsmet (approved):

PR #6792, opened 2020-01-25
Drop GraalVM 19.2.1 support and deprecate enableJni option

### LINKED ISSUE #6709: Drop the GraalVM 19.2.1 compatibility
Quarkus `1.2` supports both GraalVM `19.2.1` and `19.3.1`.

Now that we know https://github.com/oracle/graal/issues/1984 was fixed in GraalVM `19.3.1`, it should be OK to drop the GraalVM `19.2.1` compatibility in Quarkus `1.3`.

I'll submit a PR.

gwenneg:
Fixes #6709 and #6101


core/deployment/src/main/java/io/quarkus/deployment/pkg/steps/NativeImageBuildStep.java
```diff
@@ -176,8 +176,10 @@ public NativeImageBuildItem build(NativeConfig nativeConfig, NativeImageSourceJa
                 //todo: this should be specific build items
                 if (prop.getKey().equals("quarkus.ssl.native") && prop.getValue() != null) {
                     enableSslNative = Boolean.parseBoolean(prop.getValue());
-                } else if (prop.getKey().equals("quarkus.jni.enable") && prop.getValue() != null) {
-                    nativeConfig.enableJni |= Boolean.parseBoolean(prop.getValue());
+                } else if (prop.getKey().equals("quarkus.jni.enable") && prop.getValue().equals("false")) {
+                    log.warn("Your application is setting the deprecated 'quarkus.jni.enable' configuration key to false."
+                            + " Please consider removing this configuration key as it is ignored (JNI is always enabled) and it"
+                            + " will be removed in a future Quarkus version.");
```
gsmet: Just to be sure: have you tested that you don't have several identical messages when setting `quarkus.jni.enable=false`?

gsmet (changes requested):
This looks very good. I added one question inline and I have another one: I wonder if adding the SunEC library is still necessary to get SSL working? But that's material for another PR if it ends up being useless.

(Marking as `Request changes` so that we can discuss things before it being merged)

core/deployment/src/main/java/io/quarkus/deployment/pkg/steps/NativeImageBuildStep.java
gwenneg: I did it with #6306 but I'll restest it again with this PR just be sure.

gwenneg (commented):
Including the SunEC lib should not be required anymore, but I agree this should be treated separately.

I need to test this PR with JDK 11 in native mode (btw, thanks @geoand for your indications about how to do it with GitHub actions!), I'll promote the PR when I'm done with the tests.

geoand:
Looking forward to it! Thanks @gwenneg!

gwenneg:
For the record, the GitHub action got triggered with these two changes in `native-cron-build.yml`:
```
on:
  pull_request:
    branches:
      - master
```
and
```
jobs:
  build:
[...]
    steps:
[...]
      - name: Checkout Quarkus
        uses: actions/checkout@v2
        with:
          repository: gwenneg/quarkus
          ref: issue-6709-drop-graalvm-19.2.1-compatibility
```

gwenneg:
It's really great to have this GH action available, I don't have to run JDK 11 native tests locally for an entire night anymore :satisfied:

gwenneg:
@geoand It seems the JDK 11 native tests got canceled because of their duration. Is there a way to configure the GitHub action and prevent this? I tried something in https://github.com/quarkusio/quarkus/pull/6792/commits/c8d7090296305244b10196528b4668a225fd8e1c but it didn't do anything.

geoand:
I think 6 hours is the max.  Did you try executing "only" the integration tests and nothing else like the current cron job does?

gwenneg:
The native tests got canceled after 3h 34m 19s at step 47/75 so it need to run roughly for another couple hours (JDK 11 tests take so much time compared with the JDK 8 version...). I didn't change anything apart from the branch and job launch condition so it should behave like the current cron job.

gwenneg:
It seems the usual cron job runs these tests in approximately 5 hours so mine shouldn't be interrupted like that. Maybe there's an interruption cause unrelated with the tests duration.

gsmet:
If you have one of the tests of the matrix failing, it auto-cancels the others AFAICS.

gsmet:
We want to set that one to false: https://help.github.com/en/actions/automating-your-workflow-with-github-actions/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast

gwenneg:
Oh then the cancelation could be cause by the fail in the `Report` step of the JDK 8 action! It happened at the same time.

gsmet:
BTW, there's little change the report part can work in a PR as you're not supposed to have the token.

And things shouldn't be reported to the same issues anyway. So I would drop that part.

geoand:
Like @gsmet mentioned, please change the reporting for your testing so we don't get mixed information in the issue :)

gwenneg:
The native integration tests run finally ended successfully for both GH actions. I'll check the absence of duplicate warnings about `quarkus.jni.enable=false` tonight.

geoand:
Excellent news!

gastaldi:
Looks like this Draft PR is ready to be promoted to a real PR? :)

geoand:
As long as the changes to the github actions are reverted :)

gwenneg:
@gastaldi It’s not ready to be promoted yet, I need to check for duplicate warnings locally first. And I will revert the actions changes of course.

core/deployment/src/main/java/io/quarkus/deployment/pkg/steps/NativeImageBuildStep.java
gwenneg: I couldn't trigger this warning using `quarkus.jni.enable=false`, the only one being displayed is the one from `JniProcessor`. So, there's no double warning but I don't know if we need to keep this one.

gwenneg: Unless I missed something, this warning is never displayed with the current Quarkus codebase. I suppose this is an old property that is no longer used by the core extensions, but it can still be used by 3rd parties extensions.

On the other hand, the warning from `JniProcessor` is displayed when `quarkus.jni.enable` is set to `false`.

I don't think we should worry about duplicate warnings here since it's very unlikely to happen.

gwenneg:
The JDK 8 and 11 native build and tests all passed successfully (see details [here](https://github.com/quarkusio/quarkus/actions/runs/31974448)) and I'm done looking for duplicate warnings, so I'm promoting this PR.

core/deployment/src/main/java/io/quarkus/deployment/pkg/NativeConfig.java
```diff
@@ -132,7 +133,7 @@
     /**
      * The docker image to use to do the image build
      */
-    @ConfigItem(defaultValue = "quay.io/quarkus/ubi-quarkus-native-image:19.2.1")
+    @ConfigItem(defaultValue = "quay.io/quarkus/ubi-quarkus-native-image:19.3.1-java8")
```
gastaldi: Should we start using Java 11 here?

geoand: I don't think so since it still experimental isn't it?

gastaldi: That's fine by me, maybe flip the switch when we require JDK 11 as the minimum version in the future

gastaldi (approved):
Looks good to me. I'll let @gsmet review and decide when to land this PR

gsmet (approved):

gsmet:
Merged, thanks.

core/deployment/src/main/java/io/quarkus/deployment/pkg/NativeConfig.java
geoand: Makes sense

PR #6793, opened 2020-01-26
Add area/cache auto-labelling

gwenneg:
I created the `area/cache` label for the `quarkus-cache` extension.

This PR is also needed to enable auto-labelling.


machi1990 (approved):
LGTM

PR #6797, opened 2020-01-26
Efficient caching for multi stage build

crumohr:


docs/src/main/asciidoc/building-native-image.adoc
```diff
@@ -348,12 +348,13 @@ In this guide we will use the first stage to generate the native executable usin
...
 ## Stage 1 : build with maven builder image with native capabilities
 FROM quay.io/quarkus/centos-quarkus-maven:{graalvm-version} AS build
-COPY src /usr/src/app/src
-COPY pom.xml /usr/src/app
+COPY pom.xml /project/usr/src/app/
+RUN mvn -f /project/usr/src/app/pom.xml -B de.qaware.maven:go-offline-maven-plugin:1.2.4:resolve-dependencies
+COPY src /project/usr/src/app/src
 USER root
-RUN chown -R quarkus /usr/src/app
+RUN chown -R quarkus /project/usr/src/app
```
geoand: Why this change? I don't see how it's related to the other part

docs/src/main/asciidoc/building-native-image.adoc
```diff
@@ -348,12 +348,13 @@ In this guide we will use the first stage to generate the native executable usin
...
 USER quarkus
-RUN mvn -f /usr/src/app/pom.xml -Pnative clean package
+RUN mvn -f /project/usr/src/app/pom.xml -Pnative clean package
```
geoand: Same here

geoand (commented):
Thanks for the contribution!

I've added a question

docs/src/main/asciidoc/building-native-image.adoc
crumohr: Sorry for the delay, I was out of office for several weeks. You are correct, probably it's not really related. But without this change I wasn't able to build the final image. Maybe there was a change in the `quay.io/quarkus/centos-quarkus-maven` image.

docs/src/main/asciidoc/building-native-image.adoc
crumohr: Same reason as above.

geoand (changes requested):
I like the idea, however I was able to do it with fewer code changes:

```
FROM quay.io/quarkus/centos-quarkus-maven:19.3.1-java8 AS build
COPY pom.xml /usr/src/app/pom.xml
RUN mvn -f /usr/src/app/pom.xml -B de.qaware.maven:go-offline-maven-plugin:1.2.4:resolve-dependencies
COPY src /usr/src/app/src
USER root
RUN chown -R quarkus /usr/src/app
USER quarkus
RUN mvn -f /usr/src/app/pom.xml -Pnative clean package

## Stage 2 : create the docker final image
FROM registry.access.redhat.com/ubi8/ubi-minimal
WORKDIR /work/
COPY --from=build /usr/src/app/target/*-runner /work/application
RUN chmod 775 /work
EXPOSE 8080
CMD ["./application", "-Dquarkus.http.host=0.0.0.0"]
```

You would mind checking it and adapting the PR?

crumohr:
> I like the idea, however I was able to do it with fewer code changes:

I'm using the `jaxb2-maven-plugin` for some XSD processing. The only way I get the multistage build working is when I add `/project` to all paths. Otherwise the plugin always reports that it's searching under `/project/usr/src/app/src/main/resources/` for some XSD which it cannot find.

I already did a fulltext search on my project for the "project" term - it's nowhere to be found.

Maybe somebody else is using this plugin as well, my changes fix this "bug" and has no other negative side effects.

gsmet:
I think we should keep the project out. And this one needs a rebase.

@geoand can you drive this puppy home? Looks like your version was OK for a general purpose documentation.

geoand:
@gsmet sure. This actually ties in with some stuff that I need to look into that @devclaus brought to our attention.

So I'll try and have a look this week.

geoand (approved):
I like it, thanks for the contribution!

PR #6798, opened 2020-01-27
Fix broken native cron ci

geoand:
Sorry about all the noise with this one... Bash is hard :(


cescoffier (approved):

cescoffier:
Bash is hard, but Bash is life!

geoand:
Absolutely 😊

Issue #4767, opened 2019-10-22
Java 11/Gradle - NullPointerException on build Reactive Messaging


collin-sullivan:
NullPointerException is thrown at build time when using `@Incoming` or `@Outgoing` method annotations from quarkus-smallrye-reactive-messaging on Java 11.  Build completes successfully on java 1.8 or if I remove the  `@Incoming` or `@Outgoing` method annotations when using Java 11.

Last Quarkus Version it worked:
- 0.23.2

Quarkus Versions this is broken:
- 0.24.0
- 0.25.0
- 0.26.0

Sample Code:
```java
    @Incoming("requests")
    @Outgoing("responses")
    @Broadcast
    public String test() {
        return "test";
    }
```

Properties:
```properties
amqp-username=quarkus
amqp-password=quarkus


mp.messaging.incoming.requests.connector=smallrye-amqp
mp.messaging.incoming.requests.address=request
mp.messaging.incoming.requests.durable=true
mp.messaging.incoming.requests.port=5672

mp.messaging.outgoing.responses.connector=smallrye-amqp
mp.messaging.outgoing.responses.address=result
mp.messaging.outgoing.responses.durable=true
mp.messaging.outgoing.responses.broadcast=true
mp.messaging.outgoing.responses.port=5672
```

Stack Trace:
```java
org.gradle.api.tasks.TaskExecutionException: Execution failed for task ':quarkusBuild'.
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter$3.accept(ExecuteActionsTaskExecuter.java:166)
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter$3.accept(ExecuteActionsTaskExecuter.java:163)
        at org.gradle.internal.Try$Failure.ifSuccessfulOrElse(Try.java:191)
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.execute(ExecuteActionsTaskExecuter.java:156)
        at org.gradle.api.internal.tasks.execution.ValidatingTaskExecuter.execute(ValidatingTaskExecuter.java:62)
        at org.gradle.api.internal.tasks.execution.SkipEmptySourceFilesTaskExecuter.execute(SkipEmptySourceFilesTaskExecuter.java:108)
        at org.gradle.api.internal.tasks.execution.ResolveBeforeExecutionOutputsTaskExecuter.execute(ResolveBeforeExecutionOutputsTaskExecuter.java:67)
[...]

PR #6811, opened 2020-01-27
Better support for reactive-messaging builds on Java11

acoburn:
Related to #4767

The `quarkus-bootstrap-core` project depends transitively on an older
version of `commons-lang3`, which doesn't work with Java 11 when using
the `smallrye-reactive-messaging` extension. This change causes the
`quarkus-bootstrap-core` project to use the version of `commons-lang3`
that is already managed by the parent project (i.e. version 3.9)

### DIFF
--- a/independent-projects/bootstrap/core/pom.xml
+++ b/independent-projects/bootstrap/core/pom.xml
@@ -19,6 +19,10 @@
             <artifactId>asm</artifactId>
         </dependency>
         <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-lang3</artifactId>
+        </dependency>
+        <dependency>
             <groupId>org.apache.maven</groupId>
             <artifactId>maven-embedder</artifactId>
         </dependency>


geoand:
Thanks @acoburn!

Seems reasonable to me but let's see what the SME @aloubyansky has to say

cescoffier:
Another solution would be to remove the commons-lang dependency from smallrye reactive messaging. I don't remember why it is pulled (probably type/class methods)

geoand:
Last time I checked it was being used for easily finding out associations between classes

acoburn:
@cescoffier it looks like the commons-lang3 dependency [has already been removed](https://github.com/smallrye/smallrye-reactive-messaging/commit/b85cb0a96a87455fea2fabad846b3e2891b83be5) from the master branch of smallrye-reactive-messaging. So perhaps this is a matter of simply waiting on a new release of smallrye-reactive-messaging.

acoburn:
I spent some more time trying to get to the bottom of this, and it appears that the issue I encountered related to an intersection of two unrelated gradle plugins. Upgrading the other plugin seemed to resolve this issue, so I don't think anything needs to be done on the quarkus side.

PR #6835, opened 2020-01-28
KubernetesProcessor is now using OutputTargetBuildItem to determine jar name

### LINKED ISSUE #6827: Kubernetes: wrong .jar file name in JAVA_APP_JAR env var (for OpenShift)
**Describe the bug**

The `JAVA_APP_JAR` environment variable, which the OpenShift Java images use to know which JAR they should run, is generated wrongly.

Dekorate generates this env var here: https://github.com/dekorateio/dekorate/blob/master/annotations/openshift-annotations/src/main/java/io/dekorate/openshift/handler/OpenshiftHandler.java#L200

From that code, it is apparent that Dekorate tries to use its knowledge about the build to figure out the correct file name.

However, if I define `quarkus.application.name`, this value is used as the file name, which is wrong, as `quarkus.application.name` can be something different than, e.g., Maven's `${project.build.finalName}`.

**Expected behavior**

The actual JAR file name is always used, even if `quarkus.application.name` is defined.

**Actual behavior**

The value of `quarkus.application.name` is used, even if it doesn't match the file name.

**To Reproduce**

```properties
quarkus.application.name=fooooo-baaaaaar
kubernetes.deployment.target=openshift
```

See `target/kubernetes/openshift.yml`:

```yaml
apiVersion: "apps.openshift.io/v1"
kind: "DeploymentConfig"
metadata:
  ...
  name: "fooooo-baaaaaar"
spec:
  ...
  template:
    ...
    spec:
      containers:
      - env:
        ...
        - name: "JAVA_APP_JAR"
          value: "/deployments/fooooo-baaaaaar-1.0.0-SNAPSHOT-runner.jar"
        image: ""
      ...
```

iocanel:
determine jar name.

This pull request is meant to fix #6827. 
The idea is that instead of checking internal properties, we use the OutputTargetBuildItem, in order to determine which is that generated jar and configure the generated manifests accordingly.

The pull request also includes integration-test that verifies the fix.


iocanel:
@Ladicek: I can hardly keep up with your pace, but I am on it :-D

Ladicek:
Thanks @iocanel! :-)

gsmet (approved):

PR #6837, opened 2020-01-28
Use quarkus.http.root-path when creating kubernetes health related build items

### LINKED ISSUE #6833: Kubernetes: liveness and readiness probes ignore the `quarkus.http.root-path` setting
**Describe the bug**

When I configure `quarkus.http.root-path`, the `/health` endpoint is moved under it (as well as `/metrics`, but that is irrelevant here). That seems to be intentional, per https://github.com/quarkusio/quarkus/issues/3745. The generated Kubernetes probes don't reflect that, however.

**Expected behavior**

Autogenerated liveness and readiness probes point to the real health endpoint, even if `quarkus.http.root-path` is defined.

**Actual behavior**

If `quarkus.http.root-path` is defined, the autogenerated liveness and readiness probes point to a non-existing path and the pod never starts successfully.

**To Reproduce**

```properties
quarkus.http.root-path=/api
```

The generated Kubernetes YAML looks like:

```yaml
apiVersion: "apps/v1"
kind: "Deployment"
metadata:
  ...
  name: "http"
spec:
  ...
  template:
    ...
    spec:
      containers:
      - ...
        livenessProbe:
          failureThreshold: 3
          httpGet:
            path: "/health/live"
            port: 8080
            scheme: "HTTP"
          ...
        ...
        readinessProbe:
          failureThreshold: 3
          httpGet:
            path: "/health/ready"
            port: 8080
            scheme: "HTTP"
          ...
```

The OpenShift YAML is identical in this regard.

However, the health endpoint is at `/api/health`.

 - Quarkus version or git rev: `76fcdc4b8e6c113c53bb55f965a116afdee5290a`

iocanel:
This is meant to fix #6833, by populating the correct build items (ones that include quarkus.http.root-path).

In addition it adds an integration test to verify the fix.
Last but not least it adds a missing pom.xml from a relevant integration test.


extensions/smallrye-health/deployment/src/main/java/io/quarkus/smallrye/health/deployment/SmallRyeHealthProcessor.java
```diff
@@ -88,6 +89,8 @@ void healthCheck(BuildProducer<AdditionalBeanBuildItem> buildItemBuildProducer,
     void build(SmallRyeHealthRecorder recorder, RecorderContext recorderContext,
             BuildProducer<FeatureBuildItem> feature,
             BuildProducer<RouteBuildItem> routes,
+
+            HttpBuildTimeConfig httpConfig,
```
Ladicek: This seems unnecessary?

iocanel: Right! Let me get rid of it.

gsmet:
@Ladicek I let you check it fixes your issue? Thanks.

Ladicek (approved):
I've checked Quarkus built from this PR (`09b5cba67933155fa0dbfd8a5682454a83f3615b`) and I confirm that both #6827 (fix already merged in `master` on top of which this PR is based) and #6833 are fixed.

gsmet (approved):
Thanks @Ladicek for checking!

PR #6840, opened 2020-01-28
Make sure TCCL is correct for outer test class

stuartwdouglas:
We still need to create an instance of the original test,
make sure the TCCL is correct to avoid problems


geoand:
cc @radcortez

gsmet (approved):

radcortez:
Thanks!

PR #6856, opened 2020-01-29
Rename application-data-caching.adoc to cache.adoc

gsmet:
I already did that on the website for consistency reasons: we try to use
the extension name as much as possible (if it makes sense).


gastaldi (approved):

gastaldi (changes requested):
You also need to change the link to the guide: https://github.com/quarkusio/quarkus/blob/master/extensions/cache/runtime/src/main/resources/META-INF/quarkus-extension.yaml#L7

gsmet:
Raaah, forgot about that one. It's pointing to an incorrect page right now. I suppose I will add it as a backportable fix if I end up doing a 1.2.1.

gastaldi (approved):

PR #6857, opened 2020-01-29
Do not build with GraalVM 19.3.1 JDK 8 in the periodic build

gsmet:
It's now the default build of the main CI so we can reduce the matrix.


geoand (approved):
Makes sense!

gastaldi:
I think this PR could hold until we move completely to JDK 11 since the default is still set to JDK8?

geoand:
The idea of this PR is that the native tests for 19.3.1-jdk8 are already tested in the normal PR CI runs (which wasn't the case until today) so now we don't need it here as well

gastaldi:
Ok, in this case I think this makes sense

PR #6861, opened 2020-01-29
JaCoCo update to 0.8.5

rsvoboda:
JaCoCo update to 0.8.5


gsmet (approved):

gsmet:
Should we have dependabot update it?

rsvoboda:
I checked jacoco profile if the report works. So I'm like 50/50 about dependabot.

PR #6862, opened 2020-01-29
Introduce QuteContext synthetic bean

mkouba:
- this should replace the need for volatile fields declared on the bean class

TLDR> we use final fields, constructor injection and a synthetic bean whose instance is supplied by a recorder.

~The only downside of this approach I'm aware of is that you need to add `@Observes StartupEvent` to init the bean eagerly. We could probably introduce a new annotation like `@Startup`.~


geoand:
I personally like it a lot! Having a way to do constructor injection for this kind of this is great!

I do however believe that we need to get rid of the ugly 

```
    void onStart(@Observes @Priority(Interceptor.Priority.PLATFORM_BEFORE) StartupEvent event)   {
        LOGGER.debugf("Initializing Qute");
    }
```

and replace it with some sort of annotation :)

mkouba:
Yes, we already have https://github.com/quarkusio/quarkus/issues/6864.

manovotn (approved):

geoand (approved):
:+1:

stuartwdouglas (approved):

PR #6867, opened 2020-01-29
Have the JSchTest open ports on localhost rather than try guessing the address

### LINKED ISSUE #6866: JSch integration tests are failing on a vanilla Fedora installation
**Description**
The JSch integration test is attempting to open a server on the address returned by `InetAddress.getLocalHost().getHostName()`. This by default on Fedora will result in an address which is bound to Ipv6 exclusively, which in turn gets the JSch server fail to start.

**Implementation ideas**
Simply using the "localhost" constant avoids such problems.

Sanne:
…e address

Fixes #6866


gsmet (approved):
Makes sense.

gsmet:
@gastaldi I let you validate this one before I even think of merging it :).

gastaldi (approved):
Good catch!

PR #6872, opened 2020-01-29
N1hility patch 1

n1hility:


gsmet:
@n1hility please squash everything into one commit once you are done, this is definitely not what we can call a clean history :).

Thanks.

PR #6876, opened 2020-01-30
exclude shrinkwrap from openapi and bump vert.x in build-parent

michalszynkiewicz:


cescoffier (approved):
LGTM, need to check why the CI has failed.

bom/runtime/pom.xml
```diff
@@ -517,6 +517,12 @@
                 <groupId>io.quarkus</groupId>
                 <artifactId>quarkus-smallrye-openapi</artifactId>
                 <version>${project.version}</version>
+                <exclusions>
+                    <exclusion>
+                        <groupId>org.jboss.shrinkwrap</groupId>
+                        <artifactId>shrinkwrap-api</artifactId>
+                    </exclusion>
+                </exclusions>
```
gsmet: Wondering if it's an issue in `smallrye-openapi` itself and if the dependency should be test there?

gastaldi: ShrinkWrap is used by this class in SmallRye OpenAPI: https://github.com/smallrye/smallrye-open-api/blob/master/implementation/src/main/java/io/smallrye/openapi/api/util/ArchiveUtil.

gsmet: I wonder if this should be a test class or if it's really supposed to be used in user applications.

@EricWittmann ?

gsmet:
I merged but let's try to get to the bottom of it on the SmallRye OpenAPI side.

gastaldi:
I created https://github.com/smallrye/smallrye-open-api/pull/249 moving that class to the TCK project (which seems to be the only place where it is used)

gsmet:
Thanks! That's exactly what I suspected.

EricWittmann:
Well, we're having this discussion here:  https://github.com/smallrye/smallrye-open-api/issues/246

It turns out that Thorntail uses the `ArchiveUtil` class, so it can't be removed.  The latest proposal is to mark ShrinkWrap as optional...which means consumers of SmallRye OpenAPI will need to BYO-ShrinkWrap. :)

bom/runtime/pom.xml
EricWittmann: See my comment in this PR.

PR #6887, opened 2020-01-30
MP Config support for injection of all primitive types

radcortez:
MP Config 1.4 will support injection for Short, Byte and Char. This is preparation work to support these native types in Quarkus. This also requires 
updates to MP Config API (not released yet) and SmallRye Config, but it shouldn't impact the current version.

### DIFF
--- a/extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/ConfigBuildStep.java
+++ b/extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/ConfigBuildStep.java
@@ -157,19 +157,7 @@ public class ConfigBuildStep {
             return false;
         }
         if (type.kind() == Kind.PRIMITIVE) {
-            switch (type.asPrimitiveType().primitive()) {
-                case BOOLEAN:
-                case DOUBLE:
-                case FLOAT:
-                case LONG:
-                case INT:
-                case SHORT:
-                case BYTE:
-                case CHAR:
-                    return true;
-                default:
-                    return false;
-            }
+            return true;
         }
         return DotNames.STRING.equals(type.name()) ||
                 DotNames.OPTIONAL.equals(type.name()) ||
--- a/extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/ConfigBuildStep.java
+++ b/extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/ConfigBuildStep.java
@@ -163,15 +163,26 @@ public class ConfigBuildStep {
                 case FLOAT:
                 case LONG:
                 case INT:
+                case SHORT:
+                case BYTE:
+                case CHAR:
                     return true;
                 default:
                     return false;
             }
         }
-        return DotNames.STRING.equals(type.name()) || DotNames.OPTIONAL.equals(type.name()) || SET_NAME.equals(type.name())
-                || LIST_NAME.equals(type.name()) || DotNames.LONG.equals(type.name()) || DotNames.FLOAT.equals(type.name())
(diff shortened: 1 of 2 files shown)


mkouba:
There is a small conflict with https://github.com/quarkusio/quarkus/pull/6888. If possible, I'd like to merge the other PR first.

> This also requires updates to MP Config API (not released yet) and SmallRye Config

Does it mean that we have to wait for these API artifacts?

radcortez:
> There is a small conflict with #6888. If possible, I'd like to merge the other PR first.

Sure. Please go ahead. When done, I'll rebase and resolve any conflict.

 > Does it mean that we have to wait for these API artifacts?

No. Is just that we are not able to test this without the API / SmallRye support. I did run this locally with SNAPSHOT versions and seems to work fine. When updating to the new API's, the TCK already has tests and we will get the tests.

mkouba:
> When done, I'll rebase and resolve any conflict.

+1 :-)

> No. Is just that we are not able to test this without the API / SmallRye support.

I see.

radcortez:
Should be good to go now :)

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/ConfigBuildStep.java
```diff
@@ -163,15 +163,26 @@ private boolean isHandledByProducers(Type type) {
                 case FLOAT:
                 case LONG:
                 case INT:
+                case SHORT:
+                case BYTE:
+                case CHAR:
```
dmlloyd: Since all primitives are handled here now, you could just `return type.asPrimitiveType().primitive()` instead couldn't you?  Or even just return `true` perhaps?

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/ConfigBuildStep.java
```diff
@@ -163,15 +163,26 @@ private boolean isHandledByProducers(Type type) {
...
                 default:
                     return false;
             }
         }
-        return DotNames.STRING.equals(type.name()) || DotNames.OPTIONAL.equals(type.name()) || SET_NAME.equals(type.name())
-                || LIST_NAME.equals(type.name()) || DotNames.LONG.equals(type.name()) || DotNames.FLOAT.equals(type.name())
-                || DotNames.INTEGER.equals(type.name()) || DotNames.BOOLEAN.equals(type.name())
-                || DotNames.DOUBLE.equals(type.name());
+        return DotNames.STRING.equals(type.name()) ||
+                DotNames.OPTIONAL.equals(type.name()) ||
```
dmlloyd: I just noticed that `OptionalInt`/`OptionalLong`/`OptionalDouble` are missing from this list.  Might be worth fixing now or later.

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/ConfigBuildStep.java
radcortez: You are right.

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/ConfigBuildStep.java
radcortez: Sure. I'll do it in a separate PR.

gsmet (approved):

gsmet:
@mkouba gave me his blessing to merge that one. I'm going to cherry-pick it for 1.3.0.Alpha1.

radcortez:
Cool. Thanks :)

PR #6888, opened 2020-01-30
ArC - support synthetic observer methods added at build time

### LINKED ISSUE #6879: ArC - support synthetic observer methods added at build time
This is an alternative to `javax.enterprise.inject.spi.AfterBeanDiscovery.addObserverMethod()`. See https://docs.jboss.org/cdi/spec/2.0/cdi-spec.html#after_bean_discovery for more information.

mkouba:
- resolves #6879

### DIFF
--- a/independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/AbstractGenerator.java
+++ b/independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/AbstractGenerator.java
@@ -10,6 +10,8 @@ import org.jboss.jandex.Type.Kind;
 abstract class AbstractGenerator {
 
     static final String DEFAULT_PACKAGE = Arc.class.getPackage().getName() + ".generator";
+    static final String UNDERSCORE = "_";
+    static final String SYNTHETIC_SUFFIX = "Synthetic";
 
     /**
      * Create a generated bean name from a bean package. When bean is located
--- a/independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/BeanRegistrar.java
+++ b/independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/BeanRegistrar.java
@@ -1,10 +1,9 @@
 package io.quarkus.arc.processor;
 
-import io.quarkus.arc.InjectableBean;
 import org.jboss.jandex.DotName;
 
 /**
- * Allows a build-time extension to register synthetic {@link InjectableBean} implementations.
+ * Allows a build-time extension to register synthetic beans and observers.
  *
  * @author Martin Kouba
  */
@@ -19,6 +18,7 @@ public interface BeanRegistrar extends BuildExtension {
     interface RegistrationContext extends BuildContext {
 
         /**
+         * The synthetic bean is not added to the deployment unless the {@link BeanConfigurator#done()} method is called.
          *
          * @param beanClass
          * @return a new synthetic bean configurator
@@ -26,6 +26,7 @@ public interface BeanRegistrar extends BuildExtension {
         <T> BeanConfigurator<T> configure(DotName beanClassName);
 
         /**
+         * The synthetic bean is not added to the deployment unless the {@link BeanConfigurator#done()} method is called.
          * 
          * @param beanClass
          * @return a new synthetic bean configurator
@@ -33,6 +34,14 @@ public interface BeanRegistrar extends BuildExtension {
         default <T> BeanConfigurator<T> configure(Class<?> beanClass) {
             return configure(DotName.createSimple(beanClass.getName()));
         }
+
+        /**
+         * The synthetic observer is not added to the deployment unless the {@link ObserverConfigurator#done()} method is
+         * called.
+         * 
+         * @return a new synthetic observer configurator
+         */
+        ObserverConfigurator configureObserver();
 
         /**
          * The returned stream contains all non-synthetic beans (beans derived from classes) and beans
--- a/independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/BeanRegistrar.java
+++ b/independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/BeanRegistrar.java
@@ -18,7 +18,8 @@ public interface BeanRegistrar extends BuildExtension {
     interface RegistrationContext extends BuildContext {
 
         /**
-         * The synthetic bean is not added to the deployment unless the {@link BeanConfigurator#done()} method is called.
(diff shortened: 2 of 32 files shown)


independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/ObserverGenerator.java
```diff
@@ -72,35 +74,52 @@ public ObserverGenerator(AnnotationLiteralProcessor annotationLiterals, Predicat
...
         }
 
         StringBuilder sigBuilder = new StringBuilder();
-        sigBuilder.append(observer.getObserverMethod().name())
-                .append("_")
-                .append(observer.getObserverMethod().returnType().name().toString());
-        for (org.jboss.jandex.Type paramType : observer.getObserverMethod().parameters()) {
-            sigBuilder.append(paramType.name().toString());
+        if (observer.isSynthetic()) {
+            // This is not unique but best effort
```
manovotn: Hmm if we want to have this unique, we could append object hash of the `ObserverInfo` object? Something like that instead of listing the params. Depends on whether we want to have it human-readable or unique :)

independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/BeanRegistrar.java
```diff
@@ -34,6 +35,14 @@
             return configure(DotName.createSimple(beanClass.getName()));
         }
 
+        /**
+         * The synthetic observer is not added to the deployment unless the {@link ObserverConfigurator#done()} method is
+         * called.
+         * 
+         * @return a new synthetic observer configurator
+         */
+        ObserverConfigurator configureObserver();
```
manovotn: I am thinking this might deserve an `ObserverRegistrar` instead of merging it into `BeanRegistrar`.

independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/ObserverInfo.java
```diff
@@ -75,16 +88,18 @@ static ObserverInfo create(BeanInfo declaringBean, MethodInfo observerMethod, In
 
         if (!transactionPhase.equals(TransactionPhase.IN_PROGRESS) && !jtaCapabilities) {
             final ClassInfo clazz = observerMethod.declaringClass();
-            LOGGER.warnf("The method %s#%s makes use of '%s' transactional observers but no " +
+            LOGGER.warnf("The observer %s#%s makes use of '%s' transactional observers but no " +
```
manovotn: `observer method`? ;-)

independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/QualifierConfigurator.java
```diff
@@ -0,0 +1,90 @@
...
+import java.util.List;
+import java.util.function.Consumer;
+import org.jboss.jandex.AnnotationInstance;
+import org.jboss.jandex.AnnotationValue;
+import org.jboss.jandex.DotName;
+import org.jboss.jandex.Type;
+import org.jboss.jandex.Type.Kind;
+
+/**
+ * This construct is not thread-safe.
```
manovotn: This will be user facing API, can we have some more verbose javadoc please?

manovotn (commented):
Looks good overall, added few comments.

independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/ObserverGenerator.java
mkouba: Ah, sorry. It should be unique and it should not contain any random element so that it only changes if you change the source. I'm not sure it's 100% needed but use a similar strategy for bean classes where it's definitely needed.

manovotn: Oh, I see what you mean.
I recall we had one issue where synthetic bean re-generated in between tests was already loaded in the CL in its own version. So I suppose this is going to be the same, right?
It was a rare case but I don't think we anyhow solved that (e.g. made it unique on a per-build basis).

manovotn: BTW it might be worth adding something along those lines to the javadoc (of both, observer and bean).

independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/BeanRegistrar.java
mkouba: Well, it would be more correct and so but in fact, I'd rather rename the `BeanRegistrar` to `ComponentRegistrar`... but this would require some compatibility tweaks.

independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/ObserverInfo.java
mkouba: Good catch! This could even fail with NPE!

independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/QualifierConfigurator.java
mkouba: I can try to...

manovotn: @mkouba You know the drill - lorem ipsum ... :-D

independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/AbstractGenerator.java
```diff
@@ -10,6 +10,8 @@
 abstract class AbstractGenerator {
 
     static final String DEFAULT_PACKAGE = Arc.class.getPackage().getName() + ".generator";
+    static final String UNDERSCORE = "_";
+    static final String SYNTHETIC_SUFFIX = "Synthetic";
```
gastaldi: Smaller class names mean smaller JARs :wink: Maybe use the abbreviation `Syn`?

mkouba: Well, this is what I call a premature optimization :smile:. No seriously the readability is IMHO much more important in this case.

mkouba:
The windows job fails with unrelated error: `Caused by: java.net.BindException: Address already in use: bind`.

manovotn (approved):

PR #6890, opened 2020-01-30
Add some tests to the reactive streams operators extension

cescoffier:
It covers basic usage and hot reload, just to be sure it works as expected.
We already have ITs covering this extension.


mkouba (approved):

cescoffier:
I've added the infra label just to avoid having this PR in the release notes. It's just a few tests.

PR #6893, opened 2020-01-30
ArC and Qute apply the Maven impsort plugin unconditionally

Sanne:
The core Quarkus modules honour the `format.skip` build property, which is useful when needing to quickly deploy snapshots locally during prototypes. Could also be useful by CI I guess.

This plugin does slow down the build of these modules considerably, so not having this option on the other projects has been nagging me.

### DIFF
--- a/independent-projects/arc/pom.xml
+++ b/independent-projects/arc/pom.xml
@@ -263,6 +263,7 @@
                     <version>1.2.0</version>
                     <configuration>
                         <removeUnused>true</removeUnused>
+                        <skip>${format.skip}</skip>
                     </configuration>
                 </plugin>
             </plugins>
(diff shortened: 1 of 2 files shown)


gsmet:
Looks good to me but I let @mkouba decide :).

mkouba (approved):

mkouba:
@Sanne so for ArC `mvn clean install` takes ~ 24s on my machine and `formatter-maven-plugin` takes ~ 1s if no change is detected. For Qute it's ~ 10s and ~1s respectively. That being said, I agree that it makes sense to be consistent ;-).

@gsmet Should we ignore the windows job for now? It seems to be very unstable lately:
```
java.net.BindException: Address already in use: bind
```

Sanne:
@mkouba I mean with tests disabled :) When I do quick turnaround POCs, my scritps will use:

   mvn clean install -DskipTests -Dno-native -Dformat.skip

Sure a single module doesn't make a big difference, yet overall skipping the formatting on all modules gives me a significant time saving, around 30% total times so I'd like this to be honored consistently.

mkouba:
Yeah, I know. In my case it can save ~ 30s!

PR #6895, opened 2020-01-30
Don't close the EntityManager twice

Sanne:
The EntityManager is being closed twice: both during the `beforeCompletion` and during  `afterCompletion` of the transaction sync.

It's not a problem as we're being leanient about this, but even so closing the EM is not particularly cheap and this should be avoided.

N.B. this would be a problem if we eventually decided to enable Hibernate ORM's strict JPA compliance, as in that case it would throw an exception.


gsmet (approved):

gsmet:
Thanks!

PR #6908, opened 2020-01-31
Enhance command for tests run against existing native executable

rsvoboda:
Enhance command for tests run against existing native executable - adding `test-compile` phase to the command

To me the primary use case is when adding, editing or investigating failed tests


geoand (approved):
Makes sense, thanks!

PR #6912, opened 2020-01-31
Move comment text to @Disabled annotation text

rsvoboda:
Move comment text to @Disabled annotation text


machi1990 (approved):
LGTM.

geoand (approved):

gsmet:
Could we try to aggregate those small housekeeping changes into one PR with several commits and only submit one per day for instance?

Having CI running for 3 hours for a set of small changes like that is really not good considering how constrained we are right now.

We need to find the right balance sure and probably don't aggregate things that could be controversial but you get the idea.

(Too late for that one but let's try to think about it for the next ones)

Thanks!

rsvoboda:
> Could we try to aggregate those small housekeeping changes into one PR with several commits and only submit one per day for instance?

OK, will do

PR #6941, opened 2020-02-02
Adjust the count down latch in the verticle IT

lordofthejars:
Fixes small problem in VerticleDeployer IT by fixing the number of `countDown` and the initialization number of latch.


cescoffier (approved):

cescoffier:
Thanks!

lordofthejars:
Things that one see when updating the cheat sheet 🤣

PR #6946, opened 2020-02-02
Fix Scheduling Quartz Tasks with MSSQL JDBC Store

Vinche59:
The current implementation does not recognize mssql driver and thus does not load the correct delegate class.

### DIFF
--- a/extensions/quartz/deployment/src/main/java/io/quarkus/quartz/deployment/QuartzProcessor.java
+++ b/extensions/quartz/deployment/src/main/java/io/quarkus/quartz/deployment/QuartzProcessor.java
@@ -100,7 +100,7 @@ public class QuartzProcessor {
             return HSQLDBDelegate.class.getName();
         }
 
-        if (resolvedDriver.contains("com.microsoft.sqlserver.jdbc.SQLServerResource")) {
+        if (resolvedDriver.contains("com.microsoft.sqlserver.jdbc.SQLServerDriver")) {
             return MSSQLDelegate.class.getName();
         }
 


extensions/quartz/deployment/src/main/java/io/quarkus/quartz/deployment/QuartzProcessor.java
```diff
@@ -100,7 +100,7 @@ private String guessDriver(Optional<DataSourceDriverBuildItem> dataSourceDriver)
             return HSQLDBDelegate.class.getName();
         }
 
-        if (resolvedDriver.contains("com.microsoft.sqlserver.jdbc.SQLServerResource")) {
+        if (resolvedDriver.contains("com.microsoft.sqlserver.jdbc")) {
```
machi1990: Good catch. Maybe we can make it `com.microsoft.sqlserver.jdbc.SQLServerDriver` instead?

machi1990 (changes requested):

extensions/quartz/deployment/src/main/java/io/quarkus/quartz/deployment/QuartzProcessor.java
Vinche59: Indeed you're right.
Will do

machi1990: Thanks. Can we squash the two commits into one, please! 
Once that is done, this PR should be good to be merged.

machi1990 (approved):

machi1990:
merged, thanks @Vinche59

PR #6949, opened 2020-02-03
Always compile quarkus:dev

stuartwdouglas:


devtools/maven/src/main/java/io/quarkus/maven/DevMojo.java
```diff
@@ -233,6 +233,8 @@
     @Override
     public void execute() throws MojoFailureException, MojoExecutionException {
 
+        //we always want to compile if needed, so if it is run from the parent it will compile dependent projects
+        handleAutoCompile();
```
gsmet: I would have put that one below the line that enforces the Maven version?

geoand:
Closing as the commit was cherry-picked into #6953

devtools/maven/src/main/java/io/quarkus/maven/DevMojo.java
geoand: Addressed in #6953

PR #6952, opened 2020-02-03
Add librocksdbjni-linux64.so to the build when only containerRuntime is set

swastl:
When the native image is build via mvn clean package -Pnative -Dnative-image.container-runtime=docker the librocksdbjni-linux64.so was not added to the build. Therefore I added the check if the containerRuntime is present.


gunnarmorling (approved):
Should do the trick, thanks. Although I'm curious why `containerBuild` isn't true automatically when `containerRuntime` is set?

gsmet (approved):

gsmet:
I agree the config is a bit convoluted here but I don't think writing to it is a good idea as it could have some unexpected consequences. Thus let's go for this fix.

Thanks!

Issue #5686, opened 2019-11-22
Clean NettyProcessor when `netty.version` is 4.1.43.Final or greater


gwenneg:
The following code was introduced in the quarkus-netty extension in #5500 as a workaround for a GraalVM 19.3.0 compatibility [issue](https://github.com/netty/netty/pull/9631):

```
class NettyProcessor {
    // Other methods...

    // TODO: Remove this when netty.version is 4.1.43.Final or greater.
    @BuildStep
    public List<UnsafeAccessedFieldBuildItem> unsafeAccessedFields() {
        return Arrays.asList(
                new UnsafeAccessedFieldBuildItem("sun.nio.ch.SelectorImpl", "selectedKeys"),
                new UnsafeAccessedFieldBuildItem("sun.nio.ch.SelectorImpl", "publicSelectedKeys"));
    }
}
```

This code should be removed as soon as the Netty dependencies are updated to 4.1.43.Final or greater.

mejlholm:
Looks like there is a 4.1.45 version of netty now - and 4.1.42 has reported vulnerabilities:

https://snyk.io/vuln/SNYK-JAVA-IONETTY-543490
https://snyk.io/vuln/SNYK-JAVA-IONETTY-543669

Is there anything holding bumping the netty version back?

gwenneg:
Netty should be upgraded to `4.1.44.Final` on the Quarkus `master` branch before the end of the day: https://github.com/quarkusio/quarkus/pull/6956#issuecomment-581427352

PR #6956, opened 2020-02-03
Upgrade netty to 4.1.44.Final to fix known vulnerabilities

loicmathieu:
Fixes CVE-2019-20444 and CVE-2019-20445

### DIFF
--- a/bom/runtime/pom.xml
+++ b/bom/runtime/pom.xml
@@ -121,7 +121,7 @@
         <infinispan.version>10.0.0.Final</infinispan.version>
         <infinispan.protostream.version>4.3.1.Final</infinispan.protostream.version>
         <caffeine.version>2.8.0</caffeine.version>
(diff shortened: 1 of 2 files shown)


gsmet (changes requested):
There is a typo in the commit message. s/nettyt/netty.

Also, let's wait for @cescoffier 's blessing before even thinking about merging that one.

cescoffier:
The change looks ok. I briefly reviewed the Netty changes and it should be fine. But only tests would give us the "go".

cescoffier (approved):

gwenneg (changes requested):
@loicmathieu Could you please remove the GraalVM workaround described in #5686?

loicmathieu:
@gwenneg done, I tested the main integration test in native with the change and it passes.

gwenneg (approved):

PR #6962, opened 2020-02-03
Exclude osgi.annotation.versioning dependency

RotBolt:
Exclude the org.osgi:osgi.annotation.versioning from dependency graph


KaenDagger:
CC @michalszynkiewicz

gsmet (changes requested):
If you really don't want that one, could you add it here: https://github.com/quarkusio/quarkus/blob/master/build-parent/pom.xml#L211 with a small comment explaning the why?

Thanks!

KaenDagger:
@michalszynkiewicz had discussion to remove this dependency for the productized version. There is no productized version of this available . Also No github repository is there for this dependency. This could excluded from Metrics API

gsmet (approved):

gsmet:
Merged, thanks!

PR #6968, opened 2020-02-03
Fix #6966 Reproducible builds: order native image BuildItems where ne…

### LINKED ISSUE #6966: Reproducible builds: order native image BuildItems where necessary
The order of BuildItems consumed by NativeImageAutoFeatureStep (and possibly other steps) impacts the order of instructions output to the runner jar. To make the build binary reproducible, we need to make the ordering of the BuildItems consumed by NativeImageAutoFeatureStep deterministic. This can be done by making those BuildItems implement Comparable. 

Note that there are other build items causing similar problems. I'll create separate issues for those, to keep the size of the PRs manageable.

ppalaga:
…cessary


ppalaga:
This change may turn out not to be necessary based on the discussion in https://github.com/quarkusio/quarkus/pull/6972#discussion_r374348000 Closing for now

Issue #6028, opened 2019-12-08
tweak special classes with @ApplicationScoped or @Singleton by default


nimo23:
**Description**
There are classes in quarkus which are `@ApplicationScoped` by default, for example, classes annotated with `@Path`. 

Classes annotated with `@javax.persistence.Converter` could also be `@ApplicationScoped` by default. So user does not need to annotate such classes explicitly with `@ApplicationScoped`. Maybe some other types of annotated classes can also be `@ApplicationScoped` by default in quarkus.

**Implementation ideas**
Look for annotated classes which can be annotated with `@ApplicationScoped` by default.

geoand:
Closing as not planned

PR #6969, opened 2020-02-03
Introduce Startup annotation

### LINKED ISSUE #5955: Allow the easy instantiation of CDI beans
**Description**

I was wondering if it would make sense to make eager instantiation super easy for users via either an annotation (for per bean control) and / or a config property that would make bean instantation eager by default.

**Implementation ideas**

We could basically generate the necessary `void startup(@Observes StartupEvent event)` under the hood - or perhaps tie deeper into Arc to make this controllable.

mkouba:
- resolves #5955

I've decided to place the `Startup` annotation in the quarkus core (`io.quarkus.runtime` package), next to `StartupEvent`. I'm not so sure it's the best idea but since it's used to emulate an `StartupEvent` observer it makes probably more sense than to place it in `io.quarkus.arc`. WDYT?

### DIFF
--- a/extensions/arc/deployment/src/test/java/io/quarkus/arc/test/startup/StartupAnnotationTest.java
+++ b/extensions/arc/deployment/src/test/java/io/quarkus/arc/test/startup/StartupAnnotationTest.java
@@ -1,0 +1,174 @@
+package io.quarkus.arc.test.startup;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import java.util.List;
+import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.function.Consumer;
+
+import javax.annotation.PostConstruct;
+import javax.annotation.PreDestroy;
+import javax.enterprise.context.ApplicationScoped;
+import javax.enterprise.context.Dependent;
+import javax.enterprise.inject.Produces;
+import javax.enterprise.inject.spi.ObserverMethod;
+import javax.inject.Singleton;
+
+import org.jboss.jandex.AnnotationTarget;
+import org.jboss.jandex.AnnotationTarget.Kind;
+import org.jboss.shrinkwrap.api.ShrinkWrap;
+import org.jboss.shrinkwrap.api.spec.JavaArchive;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.RegisterExtension;
+
+import io.quarkus.arc.Unremovable;
+import io.quarkus.arc.deployment.AnnotationsTransformerBuildItem;
+import io.quarkus.arc.processor.AnnotationsTransformer;
+import io.quarkus.builder.BuildChainBuilder;
+import io.quarkus.builder.BuildContext;
+import io.quarkus.builder.BuildStep;
+import io.quarkus.runtime.Startup;
+import io.quarkus.test.QuarkusUnitTest;
+
+public class StartupAnnotationTest {
+
+    static final List<String> LOG = new CopyOnWriteArrayList<String>();
+
+    @RegisterExtension
+    static final QuarkusUnitTest config = new QuarkusUnitTest()
+            .setArchiveProducer(() -> ShrinkWrap.create(JavaArchive.class)
+                    .addClasses(StartMe.class, SingletonStartMe.class, DependentStartMe.class, ProducerStartMe.class))
+            .addBuildChainCustomizer(buildCustomizer());
+
+    static Consumer<BuildChainBuilder> buildCustomizer() {
+        return new Consumer<BuildChainBuilder>() {
+
+            @Override
+            public void accept(BuildChainBuilder builder) {
+                builder.addBuildStep(new BuildStep() {
+
+                    @Override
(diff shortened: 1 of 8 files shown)


geoand:
I would place it in `io.quarkus.arc` (same as for `@DefaultBean` for example)

extensions/arc/deployment/src/test/java/io/quarkus/arc/test/startup/StartupAnnotationTest.java
```diff
@@ -0,0 +1,174 @@
...
+
+    static final List<String> LOG = new CopyOnWriteArrayList<String>();
+
+    @RegisterExtension
+    static final QuarkusUnitTest config = new QuarkusUnitTest()
+            .setArchiveProducer(() -> ShrinkWrap.create(JavaArchive.class)
+                    .addClasses(StartMe.class, SingletonStartMe.class, DependentStartMe.class, ProducerStartMe.class))
+            .addBuildChainCustomizer(buildCustomizer());
+
+    static Consumer<BuildChainBuilder> buildCustomizer() {
```
geoand: Nice test, :+1:

mkouba:
Both locations are fine IMO but my point is that it's more a quarkus-specific feature (compared to `@DefaultBean` for example) than an ArC core feature and if we place in `io.quarkus.arc` we'll have to remove the link to `StartupEvent`.

geoand:
Oh right, `StartupEvent` is in `runtime`.... Then I guess having it under runtime makes more sense

manovotn (approved):
Now, this is a noteworthy feature. Looks good and the test is awesome :)

geoand:
@mkouba I think the PR needs a rebase onto master because there was a change that went in in the meantime to some of the testing stuff

manovotn:
I don't mind `@Startup` being in either of those places, so keeping it as-is now works for me.

mkouba:
Rebased...

nimo23:
@mkouba one minor question: using `@Startup` other than `@ApplicationScoped `(or @Singleton?) makes no sense. 

Following the same simplification such as jax-rs components which are `@ApplicationScoped` by default in quarkus or others (https://github.com/quarkusio/quarkus/issues/6028), could you make a `@Startup` annotated bean `@ApplicationScoped` by default. So users does not need to add this redundant information to the bean.


```
@Startup 
// @ApplicationScoped no need to declare this
@Singleton (if user annotates it explicitly as @Singleton than use @Singleton)
public class Startup{

}
 
```

mkouba:
> Following the same simplification such as jax-rs components which are @ApplicationScoped by default 

In fact, they're `@Singleton` by default.

@nimo23 it probably does not make a lot of sense to support e.g. `@Startup @Dependent` but it's still a valid use case:

```java
@Startup
@Dependent
public class Startup {
 
    @PostConstruct
    private void onStart() {
      // Startup instace is created to do some startup logic and destroyed immediately afterwards
    }
}
```
However, I agree that it would be reasonable to use `@ApplicationScoped` if no scope annotation is declared. @geoand @manovotn WDYT?

Would you care to send a PR? It should be straightforward - just register `@Startup` as a bean defining annotation with a default scope...

geoand:
> 
> However, I agree that it would be reasonable to use `@ApplicationScoped` if no scope annotation is declared. @geoand @manovotn WDYT?
> 

I think it's a good idea

manovotn:
> However, I agree that it would be reasonable to use @ApplicationScoped if no scope annotation is declared. @geoand @manovotn WDYT?

Sounds good, +1

mkouba:
Ok, anyone mentioned here pls feel free to create an issue, assign it to yourself and send a PR ;-)

geoand:
I'll take care of it tomorrow if no one else has done it by then :)

nimo23:
I can, but how to add the corresponding type for the annotation:

>just register @Startup as a bean defining annotation with a default scope...

```
@Target({ TYPE, METHOD, FIELD })
@Retention(RUNTIME)
@ApplicationScoped // is this right? dont want to make it as unchangable..
public @interface Startup {

    /**
     * 
     * @return the priority
     * @see javax.annotation.Priority
     */
    int value() default ObserverMethod.DEFAULT_PRIORITY;

}
```

mkouba:
We should just produce a new `io.quarkus.arc.deployment.BeanDefiningAnnotationBuildItem`; the name should be a `DotName` created from `Startup` and the default scope should be `BuiltinScope.APPLICATION.getName()`.

nimo23:
So something like this..

`static final BeanDefiningAnnotationBuildItem STARTUP_BUILDITEM = new BeanDefiningAnnotationBuildItem(DotName.createSimple(Startup.class.getName()), BuiltinScope.APPLICATION.getName())`

However, @geoand please create a PR for this..I am unsure where and how to put this code.. I will learn from it. Thanks!

geoand:
Ok no problem, I'll do it tomorrow and ping you 😎

geoand:
The PR is at #7048. Didn't go as smoothly as I expected :)

PR #6970, opened 2020-02-03
Run blocking when tenant config resolver exists

### LINKED ISSUE #6950: OIDC now always delegates to a thread pool
Since the multi tenacy changes OIDC will now always delegate to a blocking thread pool:

https://github.com/quarkusio/quarkus/blob/3c2cb0d8afd39e60f5dfc1fe601a75f32f2a8b0b/extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/OidcIdentityProvider.java#L41

This has a negative performance impact, this should only delegate if required.

pedroigor:


sberyozkin:
@pedroigor Hi Pedro, thanks. This looks good, it would only be blocking if `TenantConfigResolver` were used so definitely a good progress. But may be I haven't looked carefully at your concern about `AuthenticationRequestContext.runBlocking`, i.e, it still feels we can do something about `TenantConfigResolver`.
Perhaps add `TenantConfigResolver.isBlocking(RoutingContext)` and have  `DefaultTenantConfigResolver` check if `tenantConfigResolver.isResolvable() && tenantConfigResolver.get().isBlocking(context)` ? I suspect most of these providers will just return a preloaded config or create it on the fly once they check `RoutingContext`.
We can even push `AuthenticationContext` to `DefaultTenantConfigResolver.resolve` and have it returning `CompletableFuture<OidcTenantConfig>` but I suspect it won't make it easier as we'd need to keep pushing it to `TenantConfigResolver`.
So perhaps  `TenantConfigResolver.isBlocking(RoutingContext)` ?
Thanks

pedroigor:
@sberyozkin Not sure if it makes sense to add that method because after the config resolution the configuration will be used to build `OAuth2Auth` which in turn does blocking operations (e.g.: discovery and load JWKS).

pedroigor:
@sberyozkin One thing we could do though is run blocking only in case the tenant config does not exist.

stuartwdouglas (approved):

sberyozkin:
@pedroigor Oh yes, got it. The updated `isBlocking` check is better.

sberyozkin:
@pedroigor Hi Pedro, that said, is there really a possibility of the same `RequestContext` going through `OidcIdentityProvider` ? Referring here to storing the resolved tenant config context in `RoutingContext`... I suspect I'm just missing something

pedroigor:
Yes, there is, when first checking blocking and later when actually obtaining the config to authenticate.

pedroigor:
@sberyozkin do you have any clue why this https://github.com/quarkusio/quarkus/blob/f91dd8e998c7cba82eacc7b8b9c30cb490186bdc/integration-tests/oidc-code-flow/src/main/java/io/quarkus/it/keycloak/ProtectedResource.java#L38 check is failing in CI ?

sberyozkin:
@pedroigor yes, see it, nice optimization. I've been wondering for the last 30 mins or so if some other convention can be employed on top of it, as right now `OAuth2Auth` is created per request if `TenantConfigResolver` is used but not really sure :-). We can review later, I'm sure we will optimize, `TenantResolver` is a good way anyway...
thanks

sberyozkin (approved):

sberyozkin:
@pedroigor Ouch :-). That most likely means there is some thread related issue which is either being lurking around and has been exposed by this PR or possibly introduced.
[This](https://github.com/quarkusio/quarkus/blob/master/extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/OidcJsonWebTokenProducer.java#L72) or [this](https://github.com/quarkusio/quarkus/blob/master/extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/OidcIdentityProvider.java#L74) line has not been executed, it appears.

pedroigor:
@sberyozkin I also thought about those two places. But locally, even if repeating the tests 20 times, I just can't reproduce it.

Will try to rebase.

pedroigor:
@sberyozkin got the error after a rebase. At least I can now reproduce it. Will try to check what changed.

sberyozkin:
@pedroigor Thanks, nothing really changed in the OIDC code I think... Right that was the same issue with the PR I was having the other day, it was failing in CI only and Stuart fixed with [this PR](https://github.com/quarkusio/quarkus/pull/6674).

pedroigor:
@sberyozkin This line does not seem right https://github.com/quarkusio/quarkus/blob/master/extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/OidcIdentityProvider.java#L74. It should obtain the raw token from `credential.getToken().getToken()`.

After changing to that, it works fine. Wondering why we get this only now ?

sberyozkin:
@pedroigor Hi Pedro, `credential.getToken()` returns String ?

pedroigor:
@sberyozkin No, and that is the issue. But `TokenCredential`, so the comparison would never match.

sberyozkin:
@pedroigor Hmm, I'm in Eclipse right now :-). `request.getToken()` returns `TokenCedential` but the code was using ` credential`...

sberyozkin:
@pedroigor Oh I see, it is your PR which dropped `ContextAwareTokenCredential credential = (ContextAwareTokenCredential) request.getToken();` and then it replaced `credential.getToken()` with `request.getToken()` so it answers the question `why only now` :-)

sberyozkin:
@pedroigor It was super fast though how you detected the typo, would've taken me 2 days :-)

pedroigor:
> @pedroigor It was super fast though how you detected the typo, would've taken me 2 days :-)

Aha .. so it was me ! Lost of time ...

sberyozkin:
@pedroigor No problems, I guess `TokenAuthenticationRequest` could've had `getTokenCredential()` instead :-)

PR #6971, opened 2020-02-03
Qute: Support iterating by number

### LINKED ISSUE #6963: Qute iterating by number
**Description**

Iterate by a number instead of by a list of elements.

**Implementation ideas**

```
{#repeat var.total}
   ...
{/repeat}
```

I think we could do things more complex like in FreeMarker but at least having this structure would help a lot in some cases.

gastaldi:
I reused the `for` instruction to iterate by number. Support for descending order is not yet implemented. Let me see how that works for you @lordofthejars 

Fixes #6963

### DIFF
--- a/docs/src/main/asciidoc/qute-reference.adoc
+++ b/docs/src/main/asciidoc/qute-reference.adoc
@@ -231,7 +231,7 @@ A section helper that defines the logic of a section can "execute" any of the bl
 
 ===== Loop Section
 
-The loop section makes it possible to iterate over an instance of `Iterable`, `Map` 's entry set and `Stream`.
+The loop section makes it possible to iterate over an instance of `Iterable`, `Map` 's entry set, `Stream` and an Integer.
 It has two flavors.
 The first one is using the `each` name alias.
 
@@ -261,6 +261,22 @@ It's also possible to access the iteration metadata inside the loop:
 {/each}
 ----
 <1> `count` represents one-based index. Metadata also include zero-based `index`, `hasNext`, `odd`, `even`.
+
+The `for` statement also works with integers, starting from 1. In the example below, considering that `total = 3`:
+
+[source]
+----
+{#for i in total}
+  {i}:
+{/for}
+----
+
+The output will be:
+
+[source]
+----
+1:2:3:
+----
 
 ===== If Section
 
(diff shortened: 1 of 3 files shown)


gastaldi:
Also feel free to reject/close this PR if it doesn't make sense @mkouba :)

lordofthejars:
For me, it works really well as the only thing I want to do is print the number of stars depending on a rating number. So if it is 4 I just want to add 4 stars. The other thing as you said is the descending order but for now, if this is accepted I can work around the problem with not much effort.

Thank you very much.

gastaldi:
> The other thing as you said is the descending order but for now, if this is accepted I can work around the problem with not much effort.

Yeah, you can always do a `total - i` instead of `i` if you want a descending order anyway :)

mkouba:
@gastaldi looks good ;-).

> It seems like math expressions are not supported yet

Yep, complex expressions are not supported in loops.

By the way, you can also use the `{#each}` variant here, e.g. something like:
```
{#each movie.numberOfStars}
  ...print star
{/each}
```

gastaldi:
Can we get this PR approved and merged in time for 1.3.0.Alpha1?

mkouba (approved):

gsmet:
Just a small comment: let's not put pressure when none is needed. 1.3.0.Alpha1 is released for people to test the new class loading infra. We don't need to hurry to get this sort of fixes in: we can take our time to get a proper design if needed.

gastaldi:
Hm, while reading the docs, I noticed that you could have also used an `IntStream` instead in an extension method: 

Eg.
```java 
public IntStream getMovies() {
   return IntStream.rangeClosed(1, getMoviesCount()); 
}
```

Anyway, I added some docs adding support for plain Integers, let me know what you guys think :)

gsmet:
@mkouba could you review the doc?

mkouba:
@gastaldi Yes, extension methods would work too, ie. something like:
```java
public static IntStream getMovieStars(Movie movie) {
   return IntStream.rangeClosed(1, movie.getStarsCount()); 
}
```
Anyway, I think it's ok to support this directly.

DaHoC:
Proposing to also support iterating Long numbers: https://github.com/quarkusio/quarkus/pull/41679

PR #6972, opened 2020-02-03
Fix #1020 Stable main and static init method generation

ppalaga:


core/builder/src/main/java/io/quarkus/builder/Execution.java
```diff
@@ -56,12 +68,36 @@
             done = true;
     }
 
+    Map<StepInfo, Integer> orderTopologically(Set<StepInfo> startSteps) {
+        if (!(startSteps instanceof SortedSet)) {
```
dmlloyd: If this value must be a `SortedSet`, then its type should be `SortedSet` rather than having all these dynamic checks around.  All this stuff is package-private I believe.

core/builder/src/main/java/io/quarkus/builder/Execution.java
```diff
@@ -39,8 +45,14 @@
     private volatile Thread runningThread;
     private volatile boolean done;
 
+    /**
+     * A topological ordering of {@link StepInfo}s.
+     */
+    private final Map<StepInfo, Integer> stepOrdinals;
```
dmlloyd: Rather than have maps associating things with `StepInfo`, why not just put fields right on the `StepInfo`?

core/builder/src/main/java/io/quarkus/builder/BuildChainBuilder.java
```diff
@@ -64,6 +66,11 @@
      * @return the builder for the build step
      */
     public BuildStepBuilder addBuildStep(BuildStep buildStep) {
+        final String id = buildStep.toString();
```
dmlloyd: I'm a bit concerned about using these string values for ordering purposes since they're pretty arbitrary.  I wonder if you considered any other approaches to determining an ordinal value for each step?

core/builder/src/main/java/io/quarkus/builder/Execution.java
```diff
@@ -56,12 +68,36 @@
             done = true;
     }
 
+    Map<StepInfo, Integer> orderTopologically(Set<StepInfo> startSteps) {
+        if (!(startSteps instanceof SortedSet)) {
+            throw new IllegalArgumentException("startSteps must be a SortedSet");
+        }
+
+        /* Kahn's algorithm */
+        final Map<StepInfo, AtomicInteger> indegrees = new HashMap<StepInfo, AtomicInteger>();
```
dmlloyd: This is another place where it seems like a field on `StepInfo` would be simpler and faster.

core/builder/src/main/java/io/quarkus/builder/Execution.java
```diff
@@ -56,12 +68,36 @@
...
+        final Deque<StepInfo> queue = new ArrayDeque<>(startSteps);
+        final Map<StepInfo, Integer> result = new LinkedHashMap<StepInfo, Integer>();
```
dmlloyd: One more thing that could be a field on `StepInfo` instead.

core/deployment/src/main/java/io/quarkus/deployment/ExtensionLoader.java
```diff
@@ -870,7 +871,7 @@ public boolean canHandleObject(final Object obj, final boolean staticInit) {
             final Consumer<BuildStepBuilder> finalStepConfig = stepConfig.andThen(methodStepConfig)
                     .andThen(buildStepBuilder -> buildStepBuilder.buildIf(finalAddStep));
             final BiConsumer<BuildContext, Object> finalStepInstanceSetup = stepInstanceSetup;
-            final String name = clazz.getName() + "#" + method.getName();
+            final String name = clazz.getName() + "#" + method.getName() + "_" + HashUtil.methodSignatureSha1(method);
```
dmlloyd: I'm a bit concerned about this because IIRC until now these names were used *only* for diagnostic (human readable) messages.  Why does it need to be a hash?

core/deployment/src/main/java/io/quarkus/deployment/builditem/StaticBytecodeRecorderBuildItem.java
```diff
@@ -3,15 +3,23 @@
...
 
-public final class StaticBytecodeRecorderBuildItem extends MultiBuildItem {
+public final class StaticBytecodeRecorderBuildItem extends MultiBuildItem
+        implements Comparable<StaticBytecodeRecorderBuildItem> {
 
     private final BytecodeRecorderImpl bytecodeRecorder;
+    private final int ordinal;
 
-    public StaticBytecodeRecorderBuildItem(BytecodeRecorderImpl bytecodeRecorder) {
+    public StaticBytecodeRecorderBuildItem(BytecodeRecorderImpl bytecodeRecorder, int ordinal) {
```
dmlloyd: Adding an ordinal here shouldn't be necessary because the steps were already topologically sorted, therefore the build items produced will also be sorted by the same topology.

core/deployment/src/main/java/io/quarkus/deployment/builditem/SystemPropertyBuildItem.java
```diff
@@ -1,12 +1,18 @@
...
+import java.util.Comparator;
+
 import io.quarkus.builder.item.MultiBuildItem;
 
 /**
  * Represents a system property that will be set immediately on application startup.
  *
  */
-public final class SystemPropertyBuildItem extends MultiBuildItem {
+public final class SystemPropertyBuildItem extends MultiBuildItem implements Comparable<SystemPropertyBuildItem> {
```
dmlloyd: These also should not need to be comparable.

core/deployment/src/main/java/io/quarkus/deployment/logging/LogCleanupFilterBuildItem.java
```diff
@@ -11,8 +12,10 @@
  *
  * @author Stéphane Épardaud
  */
-public final class LogCleanupFilterBuildItem extends MultiBuildItem {
+public final class LogCleanupFilterBuildItem extends MultiBuildItem implements Comparable<LogCleanupFilterBuildItem> {
```
dmlloyd: This also should not need to be comparable.

dmlloyd (changes requested):
Overall I think this is a good start.  I have a few questions and suggestions to get started with in this PR and then once those are addressed I want to talk about the ordering of the steps themselves.

core/builder/src/main/java/io/quarkus/builder/Execution.java
ppalaga: Good point, thanks. I originally thought I'd have to break `StepInfo`'s immutability to do that, but now checking again it looks like it is doable.

core/builder/src/main/java/io/quarkus/builder/BuildChainBuilder.java
ppalaga: This worked for me in practice, because most of the build steps happen to have a quite well defined `toString()`. I would not mind modelling the string sort key properly, but you perhaps have a better idea?

core/builder/src/main/java/io/quarkus/builder/Execution.java
ppalaga: +1

core/builder/src/main/java/io/quarkus/builder/Execution.java
ppalaga: +1

core/deployment/src/main/java/io/quarkus/deployment/ExtensionLoader.java
ppalaga: ... because this `name` translates to `buildStep.toString()` and `BuildInfo.id` which I intransparently use for sorting BuildInfos.

core/deployment/src/main/java/io/quarkus/deployment/builditem/StaticBytecodeRecorderBuildItem.java
ppalaga: This is an interesting point. You mean it is possible to impose the canonical ordering of the Buildstep executions? I thought this was not possible due to the concurrent nature of the implementation.

core/builder/src/main/java/io/quarkus/builder/BuildChainBuilder.java
dmlloyd: I began work on this about a year ago but only got a couple ideas in code.  The first was this: https://github.com/dmlloyd/quarkus/commit/build-ordinals which defines a class that represents an overall "path" style sequence.  The second I can't find which sets the ordinal value for a `StepInfo` based on its dependency path.  The dependency with the lowest ordinal is the "parent" of a given step's ordinal, and a step has no parent if it has no dependencies.  The ordinals are assigned sequentially and (I think) deterministically (because the extension load process is sequental, so for a given build, it should run in the same order each time).

core/deployment/src/main/java/io/quarkus/deployment/ExtensionLoader.java
dmlloyd: I understood that, but why hash it, especially given that the value is meant to be human-readable?  The hash doesn't seem to contribute anything in terms of algorithmic behavior.

core/deployment/src/main/java/io/quarkus/deployment/builditem/StaticBytecodeRecorderBuildItem.java
dmlloyd: Well... to refine/correct my thought a bit... step execution has a partial order, with multiple possible correct total orders, hence our original problem, I suppose.  So you are correct that in this case we're trying to pin things down to a single correct total order and to do that requires yet another adjustment.  But sorting build items on an ad-hoc basis isn't a sustainable solution IMO.

We essentially need a way to guarantee that each multi build item type has a total order that reflects the total order of the initial step addition regardless of the order they executed in.  This also has the benefit if not requiring a sort step at the end.  I don't think we need any greater concept of total order beyond this.

I imagine one possible solution would be an array of lists.  The outer array has one "slot" for each step that can contribute the multi type, and the inner lists are the instances which are actually contributed by each step.  Then the array of lists can be flattened into a single "real" list once the producers have completed and the consumers begin.  This is marginally more efficient (in big-O terms) than sorting (`O(n)` vs `O(n log n)`) but also doesn't require items to be comparable.

core/deployment/src/main/java/io/quarkus/deployment/ExtensionLoader.java
ppalaga: Yes, both the hash could be replaced by some string representation of the method signature here. Why I took hash: because I copy-pasted it from another commit of mine where the string is a part of a class name. The hash guarantees a lexically correct and reasonably long class name there (which is indeed not necessary here).

core/builder/src/main/java/io/quarkus/builder/Execution.java
ppalaga: The whole method is gone in aacdcf7

core/builder/src/main/java/io/quarkus/builder/Execution.java
ppalaga: `StepInfo` now has an `ordinal` field in aacdcf7.

core/builder/src/main/java/io/quarkus/builder/BuildChainBuilder.java
ppalaga: `buildStep.toString()` is not used for ordering anymore in aacdcf7

core/builder/src/main/java/io/quarkus/builder/Execution.java
ppalaga: There is `BuildStepBuilder.indegree` for this now in aacdcf7

core/deployment/src/main/java/io/quarkus/deployment/ExtensionLoader.java
ppalaga: No change on this place in aacdcf7

core/deployment/src/main/java/io/quarkus/deployment/builditem/StaticBytecodeRecorderBuildItem.java
ppalaga: > step execution has a partial order, with multiple possible correct total orders, hence our original problem, I suppose.

Yes, exactly.

> the extension load process is sequental, so for a given build, it should run in the same order each time

I was not aware of this, thanks for pointing out. It is really the case and it makes the whole thing much easier.

> one possible solution would be an array of lists

I tried to implement that idea in aacdcf7 - see the new `Multis` class. It works quite well.

core/deployment/src/main/java/io/quarkus/deployment/logging/LogCleanupFilterBuildItem.java
ppalaga: +1

core/deployment/src/main/java/io/quarkus/deployment/builditem/SystemPropertyBuildItem.java
ppalaga: +1

ppalaga:
aacdcf7 adopts @dmlloyd 's proposal to leverage the sequential nature of extensions loading nad store the MultiBuildItems by the given BuildSteps ordinal to guarantee a reproducible ordering.

core/builder/src/main/java/io/quarkus/builder/BuildChainBuilder.java
```diff
@@ -34,9 +36,9 @@
 
     private final BuildStepBuilder finalStep;
     private final List<BuildProvider> providers = new ArrayList<>();
-    private final Map<BuildStepBuilder, StackTraceElement[]> steps = new HashMap<BuildStepBuilder, StackTraceElement[]>();
-    private final Set<ItemId> initialIds = new HashSet<>();
-    private final Set<ItemId> finalIds = new HashSet<>();
+    private final Map<BuildStepBuilder, StackTraceElement[]> steps = new LinkedHashMap<BuildStepBuilder, StackTraceElement[]>();
+    private final Set<ItemId> initialIds = new LinkedHashSet<>();
+    private final Set<ItemId> finalIds = new LinkedHashSet<>();
```
dmlloyd: I'm fairly certain these don't need to be linked in order to produce a reproducible order (since steps are already run concurrently).

core/builder/src/main/java/io/quarkus/builder/BuildChainBuilder.java
```diff
@@ -417,19 +425,46 @@ private StepInfo buildOne(BuildStepBuilder toBuild, Set<BuildStepBuilder> includ
...
+        toBuild.dependencyCount(includedDependencies);
+
+        if (toBuild.getDependencyCount() == 0) {
+            // it's a start step!
+            startSteps.add(toBuild);
+        }
+
+    }
+
+    void setOrdinals(Set<BuildStepBuilder> startSteps) {
```
dmlloyd: Could the ordinals now be set at the time the build steps are each actually added?

core/builder/src/main/java/io/quarkus/builder/BuildContext.java
```diff
@@ -240,17 +224,7 @@ private void doProduce(ItemId id, BuildItem value) {
...
-                if (Comparable.class.isAssignableFrom(id.getType())) {
-                    int pos = Collections.binarySearch((List) list, value);
-                    if (pos < 0)
-                        pos = -(pos + 1);
-                    list.add(pos, value);
-                } else {
-                    list.add(value);
-                }
-            }
+            execution.getMultis().put(stepInfo.getOrdinal(), id, (MultiBuildItem) value);
```
dmlloyd: This entire sorting behavior was probably a mistake and I'd be glad to see it go away TBH.

core/builder/src/main/java/io/quarkus/builder/ItemId.java
```diff
@@ -11,14 +11,22 @@
...
+        this.comparable = Comparable.class.isAssignableFrom(itemType);
     }
 
     boolean isMulti() {
-        return MultiBuildItem.class.isAssignableFrom(itemType);
+        return multi;
+    }
+
+    boolean isComparable() {
+        return comparable;
```
dmlloyd: It doesn't seem worth wasting fields for what can be done in a run time type check.

dmlloyd (changes requested):
Next round, then I want to look into the Multis impl directly next.

core/builder/src/main/java/io/quarkus/builder/BuildChain.java
```diff
@@ -102,7 +102,7 @@ int getInitialMultiCount() {
         return initialMultiCount;
     }
 
-    List<StepInfo> getStartSteps() {
+    Set<StepInfo> getStartSteps() {
```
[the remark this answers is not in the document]
ppalaga: No, it is not required to solve the reproducibility problem, it just simplifies the flow that existed even before this change: the `startSteps` field was initialized from a Set anyway. Now the Set is kept instead of being copied to a new ArrayList. Do you prefer to get back to a List for some reason?

core/builder/src/main/java/io/quarkus/builder/BuildChainBuilder.java
ppalaga: +1 Good catch, thanks!

core/builder/src/main/java/io/quarkus/builder/BuildChainBuilder.java
ppalaga: Yes, they can, thanks for asking :) Calling `buildOne()` for each `included` is actually a depth-first search - a well known topological sort algorithm.

core/builder/src/main/java/io/quarkus/builder/ItemId.java
ppalaga: no problem.

core/builder/src/main/java/io/quarkus/builder/BuildChainBuilder.java
ppalaga: After some testing, `finalIds` turns out to require a LinkedHashSet, while initialIds does not.

ppalaga:
351945e : addressed all comments I was able to.

core/builder/src/main/java/io/quarkus/builder/ItemId.java
ppalaga: addressed in 351945e

core/builder/src/main/java/io/quarkus/builder/BuildChainBuilder.java
ppalaga: Solved in 351945e

core/builder/src/main/java/io/quarkus/builder/BuildChainBuilder.java
ppalaga: `finalIds` still a LinkedHashSet, initialIds back a HashSet in 351945e

core/builder/src/main/java/io/quarkus/builder/BuildContext.java
ppalaga: I found that only `ConfigDescriptionBuildItem` implements `Comparable` in the Quarkus source tree and no Comparable BuildItem in Camel Quarkus. I can have a look if we can eliminate it. WDYT?

dmlloyd: I think that's a good idea.  It was initially just a random idea but it's not useful in practice.

core/builder/src/main/java/io/quarkus/builder/BuildChain.java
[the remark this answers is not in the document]
dmlloyd: No, I suppose it doesn't matter if the steps were already a set.

core/builder/src/main/java/io/quarkus/builder/BuildContext.java
ppalaga: Done in 4810080eb79800718dcd1a42e8c2fba4f5620db1

ppalaga:
Multi sorting behavior removed in 4810080eb79800718dcd1a42e8c2fba4f5620db1

core/builder/src/main/java/io/quarkus/builder/ItemId.java
```diff
@@ -11,14 +11,16 @@
...
+    private final boolean multi;
 
     ItemId(final Class<? extends BuildItem> itemType) {
         Assert.checkNotNullParam("itemType", itemType);
         this.itemType = itemType;
+        this.multi = MultiBuildItem.class.isAssignableFrom(itemType);
     }
 
     boolean isMulti() {
-        return MultiBuildItem.class.isAssignableFrom(itemType);
```
dmlloyd: I don't think this change is needed here, as the check should be fast already.

core/builder/src/main/java/io/quarkus/builder/Multis.java
```diff
@@ -0,0 +1,165 @@
...
+import io.quarkus.builder.item.MultiBuildItem;
+
+/**
+ * A facility for storing and retrieving {@link MultiBuildItem}s by {@link ItemId}.
+ */
+class Multis {
+    /** A special ordinal to use for entries that are available initially before the {@link Execution} has started */
+    private static final int INITIAL_ORDINAL = -1;
+
+    private final ConcurrentHashMap<ItemId, TopoMultiList<MultiBuildItem>> multis;
```
dmlloyd: This seems overly complex.  Could it not be a `List<List<MultiBuildItem>>`?  The outer index would be the position relative to the ordinal of the producing step.  The size of the outer list should be known ahead of time because it is bounded by the number of producing steps (which should be a small-ish number).  You could treat a `null` entry as an empty list to avoid constructing useless objects.

core/builder/src/main/java/io/quarkus/builder/BuildStepBuilder.java
```diff
@@ -185,6 +192,22 @@ public BuildStepBuilder consumes(Class<? extends BuildItem> type, ConsumeFlags f
         return this;
     }
 
+    public BuildStepBuilder ordinal(int ordinal) {
```
dmlloyd: We wouldn't want this to be public.  The ordinal could possibly be automatically set using a global counter on `BuildChainBuilder`.

dmlloyd (changes requested):
Next stage of review.

core/builder/src/main/java/io/quarkus/builder/ItemId.java
ppalaga: Sorry, forgot to revert this.

core/builder/src/main/java/io/quarkus/builder/BuildStepBuilder.java
ppalaga: Good catch. The method and field are actually unused since the last iteration, so they can be removed.

core/builder/src/main/java/io/quarkus/builder/Multis.java
ppalaga: See 26312da

ppalaga:
26312da simplifies the Multis implementation using precomputed ordinal lookup tables.

ppalaga:
Is this good to merge now?

gsmet:
OK, this now needs a rebase. I think it's something we want. @dmlloyd can you re-review once rebased and get this merged?

gsmet:
@ppalaga let's wait for @dmlloyd to confirm it's something we want to pursue before you rebasing again.

dmlloyd:
Well, I think it's worth pursuing generally speaking.  I can't say I'm excited to review it all again though. :)

gastaldi:
@ppalaga This needs a rebase, so @dmlloyd can happily review it again ;)

ppalaga:
Yes, binary reproducibility of Quarkus builds would be awesome. But without a broader commitment towards the goal, it is very hard to make sure that code in the current PR is correct. I mean there are many other sources of irreproducibility in Quarkus, Gizmo and Jandex and without fixing those, even a simple reproducibility test will fail. So I lean towards keeping this on ice until there is a broad consensus to fix all related issues.

gastaldi:
@ppalaga okay. I'll close it in the meantime then. Please reopen when there is a consensus about this

Issue #6726, opened 2020-01-22
Kubernetes: inconsistent handling of kubernetes.name/group/version properties


Ladicek:
**Describe the bug**

I'm using the `kubernetes` extension to generate Kubernetes and OpenShift resources, and find that the `kubernetes.name/group/version` configuration properties are handled wildly inconsistently.

The problem seems to be in Dekorate, but the `KubernetesProcessor` in Quarkus is also involved (it sets the `app.group` system property). I think the root cause is that the Dekorate `Resources` object, which is shared among the entire Dekorate `Session`, is being modified in the middle of Dekorate processing (by the callers of `AbstractKubernetesHandler.setApplicationInfo`).

Here's my `application.properties` file:

```properties
quarkus.application.name=test
kubernetes.name=foo
kubernetes.group=bar
kubernetes.version=my1.0version
kubernetes.expose=true
```

And the generated `kubernetes.yml`; marking the relevant lines by `!!!`:

```yaml
---
apiVersion: "v1"
kind: "ServiceAccount"
metadata:
  labels:
    app: "test" # !!!
    version: "1.0.0-SNAPSHOT" # !!!
    group: "bar" # !!!
  name: "foo" # !!!
---
apiVersion: "v1"
kind: "Service"
metadata:
  labels:
    app: "test" # !!!
    version: "1.0.0-SNAPSHOT" # !!!
    group: "bar" # !!!
  name: "foo" # !!!
spec:
  ports:
  - name: "http"
    port: 8080
    targetPort: 8080
  selector:
    app: "foo" # !!!
    version: "my1.0version" # !!!
    group: "bar" # !!!
  type: "ClusterIP"
---
apiVersion: "apps/v1"
kind: "Deployment"
metadata:
  labels:
    app: "foo" # !!!
    version: "my1.0version" # !!!
    group: "bar" # !!!
  name: "foo" # !!!
spec:
  replicas: 1
  selector:
    matchLabels:
      app: "foo" # !!!
      version: "my1.0version" # !!!
      group: "bar" # !!!
  template:
    metadata:
      labels:
        app: "foo" # !!!
        version: "my1.0version" # !!!
        group: "bar" # !!!
    spec:
      containers:
      - env:
        - name: "KUBERNETES_NAMESPACE"
          valueFrom:
            fieldRef:
              fieldPath: "metadata.namespace"
        image: "bar/test:1.0.0-SNAPSHOT" # !!!
        imagePullPolicy: "IfNotPresent"
[...]

PR #6977, opened 2020-02-03
fix (#6726): Bump dekorate to 0.10.8 and add integration test that

iocanel:
verifies the issue has been fixed.

#6726 has been fixed in the dekorate 0.10.8.
This pull request upgrades to `0.10.8` and also adds an integration test that verifies that the reported issue is fixed.


geoand (approved):
Thanks!

PR #6990, opened 2020-02-04
ArC - improve performance of Reflections.findX() methods

### LINKED ISSUE #6959: Improve io.quarkus.arc.impl.Reflections.findX() performance
As described in https://github.com/quarkusio/quarkus/pull/6922:

At runtime we need to find an appropriate `java.lang.reflect` construct for intercepted beans and beans with an injection point that resolves to a dependent bean that injects `InjectionPoint` metadata. The current implementation demonstrates a poor performance for class hierarchies where we need to iterate over subclasses and `NoSuchFieldException`/`NoSuchMethodException` exceptions are currently thrown.

A typical use case is a `@RequestScoped` resource with `@Transactional` and/or injecting `@ConfigProperty`.

Note that we canot get rid of the reflection because of interceptors API, e.g. `javax.interceptor.InvocationContext.getMethod()`.

We should consider two solutions:
1. Use caching
2. Switch to `Class.getDeclaredFields()` + iteration.

mkouba:
- resolves #6959

### Implementation notes
- we do cache all results in static caches
- we intentionally do not use weak references for keys/values - caches are cleared when the when the container is shut down

### Microbenchmarks

I've added a microbenchmark to compare the performance results: https://github.com/mkouba/arc-benchmarks/blob/master/src/main/java/io/quarkus/arc/benchmarks/ReflectionsBenchmark.java. And the results are:
```
1.2.0.Final
Benchmark                         Mode  Cnt       Score       Error  Units
ReflectionsBenchmark.findField   thrpt   25  468775,169 ± 21317,025  ops/s
ReflectionsBenchmark.findMethod  thrpt   25  320135,824 ±  6466,329  ops/s

999-SNAPSHOT
Benchmark                         Mode  Cnt         Score         Error  Units
ReflectionsBenchmark.findField   thrpt   25  62357079,271 ± 2309336,554  ops/s
ReflectionsBenchmark.findMethod  thrpt   25  54075745,129 ± 1846794,145  ops/s
```

### DIFF
--- a/independent-projects/arc/runtime/src/main/java/io/quarkus/arc/impl/Reflections.java
+++ b/independent-projects/arc/runtime/src/main/java/io/quarkus/arc/impl/Reflections.java
@@ -202,20 +202,23 @@ public final class Reflections {
         final Class<?> clazz;
         final String methodName;
         final Class<?>[] parameterTypes;
+        final int hashCode;
 
         public MethodKey(Class<?> clazz, String methodName, Class<?>[] parameterTypes) {
             this.clazz = clazz;
             this.methodName = methodName;
             this.parameterTypes = parameterTypes;
+            final int prime = 31;
+            int result = 1;
+            result = prime * result + Arrays.hashCode(parameterTypes);
+            result = prime * result + ((clazz == null) ? 0 : clazz.hashCode());
+            result = prime * result + ((methodName == null) ? 0 : methodName.hashCode());
+            this.hashCode = result;
         }
 
         @Override
         public int hashCode() {
-            final int prime = 31;
-            int result = 1;
-            result = prime * result + Arrays.hashCode(parameterTypes);
-            result = prime * result + Objects.hash(clazz, methodName);
-            return result;
+            return hashCode;
         }
 
         @Override
@@ -240,15 +243,21 @@ public final class Reflections {
 
         final Class<?> clazz;
         final String fieldName;
+        final int hashCode;
 
         public FieldKey(Class<?> clazz, String fieldName) {
             this.clazz = clazz;
             this.fieldName = fieldName;
+            final int prime = 31;
+            int result = 1;
+            result = prime * result + ((clazz == null) ? 0 : clazz.hashCode());
+            result = prime * result + ((fieldName == null) ? 0 : fieldName.hashCode());
+            this.hashCode = result;
         }
 
         @Override
         public int hashCode() {
-            return Objects.hash(clazz, fieldName);
+            return hashCode;
         }
 
         @Override
@@ -267,4 +276,5 @@ public final class Reflections {
         }
 
     }
+
 }
--- a/independent-projects/arc/runtime/src/main/java/io/quarkus/arc/impl/Reflections.java
+++ b/independent-projects/arc/runtime/src/main/java/io/quarkus/arc/impl/Reflections.java
@@ -11,29 +11,78 @@ import java.lang.reflect.Type;
 import java.lang.reflect.TypeVariable;
 import java.lang.reflect.WildcardType;
 import java.util.Arrays;
+import java.util.Objects;
+import java.util.function.Function;
 
+/**
+ * Neither the class nor its methods are considered a public API and should only be used internally.
+ */
 public final class Reflections {
+
+    // Note that we intentionally do not use weak references for keys/values
+    // The reason is that:
+    // (1) ArC does not support multiple deployments on the classpath
+    // (2) the caches are cleared when the container is shut down
+    private static final ComputingCache<FieldKey, Field> FIELDS_CACHE = new ComputingCache<>(new Function<FieldKey, Field>() {
+        @Override
+        public Field apply(FieldKey key) {
+            return findFieldInternal(key.clazz, key.fieldName);
+        }
+    });
+    private static final ComputingCache<MethodKey, Method> METHODS_CACHE = new ComputingCache<>(
+            new Function<MethodKey, Method>() {
+                @Override
+                public Method apply(MethodKey key) {
+                    return findMethodInternal(key.clazz, key.methodName, key.parameterTypes);
+                }
+            });
+
+    static void clearCaches() {
+        FIELDS_CACHE.clear();
+        METHODS_CACHE.clear();
+    }
 
     private Reflections() {
     }
 
+    /**
+     * 
+     * @param clazz
+     * @param fieldName
+     * @return the field declared in the class hierarchy
+     */
     public static Field findField(Class<?> clazz, String fieldName) {
+        return FIELDS_CACHE.getValue(new FieldKey(clazz, fieldName));
+    }
+
+    private static Field findFieldInternal(Class<?> clazz, String fieldName) {
         try {
             return clazz.getDeclaredField(fieldName);
         } catch (NoSuchFieldException e) {
             if (clazz.getSuperclass() != null) {
-                return findField(clazz.getSuperclass(), fieldName);
+                return findFieldInternal(clazz.getSuperclass(), fieldName);
             }
             throw new IllegalArgumentException(e);
         }
     }
 
+    /**
+     * 
+     * @param clazz
+     * @param methodName
+     * @param parameterTypes
+     * @return the method declared in the class hierarchy
+     */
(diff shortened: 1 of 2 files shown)


mkouba:
@bcluap It would be great if you could verify/re-run your test with this PR.

mkouba:
@bcluap And by the way thanks a lot for catching this problem! :-)

independent-projects/arc/runtime/src/main/java/io/quarkus/arc/impl/Reflections.java
```diff
@@ -148,4 +197,74 @@ public static Object invokeMethod(Class<?> clazz, String name, Class<?>[] paramT
         }
     }
 
+    static final class MethodKey {
+
+        final Class<?> clazz;
+        final String methodName;
+        final Class<?>[] parameterTypes;
```
gsmet: Could you try with:
- removing the finals here and in the other key class
- caching the hashCode in the constructor
- not using `Objects.hash(clazz, methodName)` as it does one more array allocation

That's what I did in HV and it was beneficial. If it doesn't make any significant difference, we can keep it your way but as you have a benchmark and we are optimizing things, it's worth a quick check IMHO

mkouba: I can give it a try!

gsmet: Btw, the remark on `Objects.hash()` is also valid for the Method one.

gsmet: After giving it some thoughts, let's keep the ``final``s and just experiment with the rest.

It's a bit different than in HV as you don't really publish these maps safely before accessing them in a multi-threaded environments.

mkouba: So I've tried to cache the hashCodes and the results are:
```
Benchmark                         Mode  Cnt         Score         Error  Units
ReflectionsBenchmark.findField   thrpt   25  70640655,243 ± 1919015,675  ops/s
ReflectionsBenchmark.findMethod  thrpt   25  57293358,785 ± 1065670,627  ops/s
```
Now I'm going to replace the `Objects.hash()` ...

mkouba:
```
Benchmark                         Mode  Cnt          Score         Error  Units
ReflectionsBenchmark.findField   thrpt   25  104686242,346 ± 3802225,528  ops/s
ReflectionsBenchmark.findMethod  thrpt   25   81447915,881 ±  661570,727  ops/s
```
Awesome, thanks for your tips ;-)

gsmet: Yeah, we have a lot of pressure on the memory in HV as the path is really very optimized CPU-wise so we ended up with all sort of tricks to optimize the map lookups :).

Glad it could be of use for another project.

independent-projects/arc/runtime/src/main/java/io/quarkus/arc/impl/Reflections.java
```diff
@@ -148,4 +197,84 @@ public static Object invokeMethod(Class<?> clazz, String name, Class<?>[] paramT
...
+        public MethodKey(Class<?> clazz, String methodName, Class<?>[] parameterTypes) {
+            this.clazz = clazz;
+            this.methodName = methodName;
+            this.parameterTypes = parameterTypes;
+            final int prime = 31;
+            int result = 1;
+            result = prime * result + Arrays.hashCode(parameterTypes);
+            result = prime * result + ((clazz == null) ? 0 : clazz.hashCode());
+            result = prime * result + ((methodName == null) ? 0 : methodName.hashCode());
+            this.hashCode = result;
```
gsmet: Maybe move that part to a `buildHashCode` method? :)

mkouba: I don't think it's necessary to create a separate method for this ;-).

gsmet: As you prefer. Looks good to me, then!

gsmet (approved):

gsmet:
Let's wait for @bcluap 's feedback before merging.

bcluap:
Thanks guys. My worst case scenario of 160 methods on the superclass goes from around 1200 tps to 2800 tps. So about 3X better

gsmet:
OK, so I played a bit with asyncprofiler, using the app @bcluap posted here: `git clone git@github.com:bcluap/quarkus.git` then the `reflection` directory.

I _think_ the situation is quite bad as the constructor of `org/acme/quickstart/GreetingResource_Subclass` has a very important cost.

I'm wondering why we need to initialize all this for each instance? Shouldn't it be partially cached somehow? Also there are a lot of `AnnotationLiteral` `hashCode` that has a very high cost. Calling reflection in a `hashCode` is a very bad idea. Either we need to fix that or have a wrapper around it to cache the `hashCode` (fixing it would have my preference as a wrapper has a cost too).

@mkouba you know far more about that than me so I let you have a look at the attached profile and maybe we can discuss it tomorrow?

Also I'm not entirely sure if having a lot of `$$function$$` classes is a good choice. But it's a separate discussion.

[profile.zip](https://github.com/quarkusio/quarkus/files/4155131/profile.zip)

gsmet:
Just a few notes:

As root:
```
echo 1 > /proc/sys/kernel/perf_event_paranoid
```

Then download https://github.com/jvm-profiling-tools/async-profiler .

Then run your Quarkus app with:
```
java "-agentpath:$HOME/Downloads/build/libasyncProfiler.so=start,e=wall,file=profile.svg,interval=1000000" -jar target/reflection-1.0-SNAPSHOT-runner.jar
```

Then in another term:
```
ab -c 10 -n 100000 http://localhost:8080/hellofast
```

Then when finished, you `CTRL+C` the Quarkus process and you get your `profile.svg` ready for analysis. You can zoom on things when clicking but the nav isn't very good so don't expect a back button, you have to click again on the line at the bottom to unzoom.

gsmet:
I'm merging this one but see my comments above for possible follow-ups.

mkouba:
> I think the situation is quite bad as the constructor of org/acme/quickstart/GreetingResource_Subclass has a very important cost.

Yes, this use case was not optimized yet. On the other hand, it's probably not a common one (quarkus has `@Singleton` resources by default, JAX-RS resource with 100 methods?).

> I'm wondering why we need to initialize all this for each instance? Shouldn't it be partially cached somehow?

We could cache and even share some metadata, in this particular case we can share one  `Transactional_Shared_AnnotationLiteral` instance and also avoid creating empty arrays. I'll send a PR later today.

>  Either we need to fix that or have a wrapper around it to cache the hashCode (fixing it would have my preference as a wrapper has a cost too)

I think that we could replace the `AnnotationLiteral` from the spec with our own version in some cases. WRT caching the hashcode - there was a discussion in the CDI EG and the problem is that `AnnotationLiteral` can be mutable [1]. That was the reason why couldn't change it on the spec level. However, it's OK to use our own optimized version to replace the `java.lang.annotation.Annotation`.

[1]
http://cdi-development-mailing-list.1064426.n5.nabble.com/Mutable-annotation-literals-td5706867.html

> Also I'm not entirely sure if having a lot of $$function$$ classes is a good choice. But it's a separate discussion.

Unfortunately, I haven't found a better way. Those function classes represent instances of functional interfaces that are used to call the original method. The other possibility is to use reflection but I don't think this would help in terms of performance.

gsmet:
So I totally agree this is an extreme case but I wouldn't be surprised by services with ~20 endpoints and, considering the cost it has, it would certainly make a difference there too.

I know it's not the default, it's just that you pay a very high price when suddenly you need to make your bean `@RequestScoped` so optimizing this case makes sense too.

And btw, I'm pretty sure you will find it interesting ;). I can rerun the benchmarks if needed, just thought you might be interested in a way to reproduce the issue and check your progress.

mkouba:
> And btw, I'm pretty sure you will find it interesting ;).

+100 :D

PR #6994, opened 2020-02-04
Remove Elasticsearch port bindings in the logging-gelf docker-compose

loicmathieu:
@gsmet I remove the port binding for Elasticsearch for the docker-compose shiped inside the logging-gelf integration tests as they are not needed (it can be handy if you need raw access to elasticsearch). Maybe it will be easier now to integrate the test inside the CI ;)


geoand (approved):
Makes sense

gsmet (approved):
Perfect, thanks.

PR #7002, opened 2020-02-05
Gradle test app model

aloubyansky:
This PR fixes a couple of issues in our Gradle plugin.

1. To pass the resolved application dependencies to Quarkus test extension we serialize the AppModel resolve in QuarkusTestConfig to the test classes directory. The issue is that the file may get removed before the tests are executed as "stale content". This issue is currently fixed by saving the serialized AppModel in a temporary directory outside of the project directory tree.

2. When AppModel is resolved by the Gradle AppModel resolver only the application compile configuration is taken into account. When running tests however, the test dependencies are missing from the model. This is fixed by making the Gradle AppModel resolver aware of the launch mode the AppModel is intended to be used for.

### DIFF
--- a/independent-projects/bootstrap/core/src/main/java/io/quarkus/bootstrap/BootstrapConstants.java
+++ b/independent-projects/bootstrap/core/src/main/java/io/quarkus/bootstrap/BootstrapConstants.java
@@ -6,12 +6,12 @@ package io.quarkus.bootstrap;
  */
 public interface BootstrapConstants {
 
-    String SERIALIZED_APP_MODEL = "serialized-app-model.data";
+    String SERIALIZED_APP_MODEL = "quarkus-internal.serialized-app-model.path";
     String DESCRIPTOR_FILE_NAME = "quarkus-extension.properties";
-    
+
     @Deprecated
     String EXTENSION_PROPS_JSON_FILE_NAME = "quarkus-extension.json";
-    
+
     String QUARKUS_EXTENSION_FILE_NAME = "quarkus-extension.yaml";
 
     String META_INF = "META-INF";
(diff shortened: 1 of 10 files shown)


independent-projects/bootstrap/core/src/main/java/io/quarkus/bootstrap/BootstrapConstants.java
```diff
@@ -6,12 +6,12 @@
  */
 public interface BootstrapConstants {
 
-    String SERIALIZED_APP_MODEL = "serialized-app-model.data";
+    String SERIALIZED_APP_MODEL = "quarkus.serialized-app-model.path";
```
gsmet: I wouldn't make it `quarkus.` prefixed as we might have some warning coming. Better use `quarkus-internal.` that we already use a few times in the code.

gsmet (changes requested):
Makes sense, thanks.

I just added a small comment about the name of the system property.

maxandersen:
> 1. The issue is that the file may get removed before the tests are executed as "stale content". This issue is currently fixed by saving the serialized AppModel in a temporary directory outside of the project directory tree.

this makes me think we aren't doing this right in gradle - we should be able to declare it is not stale unless certain things changes; writing outside in a temporary directory sounds like a workaround that will eventually also fail. good enough for a quick fix but is there a not better approach to take here in the long run ?

aloubyansky:
> > 1. The issue is that the file may get removed before the tests are executed as "stale content". This issue is currently fixed by saving the serialized AppModel in a temporary directory outside of the project directory tree.
> 
> this makes me think we aren't doing this right in gradle - we should be able to declare it is not stale unless certain things changes; writing outside in a temporary directory sounds like a workaround that will eventually also fail. good enough for a quick fix but is there a not better approach to take here in the long run ?

Hopefully, there is/will be. Not sure at this point how to do that otherwise.

gsmet (approved):

PR #7003, opened 2020-02-05
ArC- reduce overhead of metadata initialization in subclass constructors

mkouba:
- use shared literal instances for interceptor bindings
- use shared interceptor chains

### DIFF
--- a/independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/AnnotationLiteralGenerator.java
+++ b/independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/AnnotationLiteralGenerator.java
@@ -30,6 +30,7 @@ import org.jboss.jandex.ArrayType;
 import org.jboss.jandex.ClassInfo;
 import org.jboss.jandex.DotName;
 import org.jboss.jandex.MethodInfo;
+import org.jboss.jandex.PrimitiveType;
 import org.jboss.jandex.Type;
 import org.jboss.logging.Logger;
 
@@ -228,7 +229,20 @@ public class AnnotationLiteralGenerator extends AbstractGenerator {
                     valueMethod.writeArrayValue(retValue, i, valueMethod.load(charArray[i]));
                 }
                 break;
-            // TODO: handle other less common types of array components
+            case FLOAT:
+                float[] floatArray = value.asFloatArray();
+                retValue = valueMethod.newArray(componentType(method), valueMethod.load(floatArray.length));
+                for (int i = 0; i < floatArray.length; i++) {
+                    valueMethod.writeArrayValue(retValue, i, valueMethod.load(floatArray[i]));
+                }
+                break;
+            case DOUBLE:
+                double[] doubleArray = value.asDoubleArray();
+                retValue = valueMethod.newArray(componentType(method), valueMethod.load(doubleArray.length));
+                for (int i = 0; i < doubleArray.length; i++) {
+                    valueMethod.writeArrayValue(retValue, i, valueMethod.load(doubleArray[i]));
+                }
+                break;
             default:
                 // Return empty array for empty arrays and unsupported types
                 // For an empty array the component kind is UNKNOWN
@@ -241,14 +255,30 @@ public class AnnotationLiteralGenerator extends AbstractGenerator {
                                 annotationClass);
                     }
                 }
-                retValue = valueMethod.newArray(componentType(method), valueMethod.load(0));
+                DotName componentName = componentTypeName(method);
+                // Use empty array constants for common component kinds
+                if (DotNames.CLASS.equals(componentName)) {
+                    retValue = valueMethod.readStaticField(FieldDescriptors.ANNOTATION_LITERALS_EMPTY_CLASS_ARRAY);
+                } else if (DotNames.STRING.equals(componentName)) {
+                    retValue = valueMethod.readStaticField(FieldDescriptors.ANNOTATION_LITERALS_EMPTY_STRING_ARRAY);
+                } else if (PrimitiveType.LONG.name().equals(componentName)) {
+                    retValue = valueMethod.readStaticField(FieldDescriptors.ANNOTATION_LITERALS_EMPTY_LONG_ARRAY);
+                } else if (PrimitiveType.INT.name().equals(componentName)) {
+                    retValue = valueMethod.readStaticField(FieldDescriptors.ANNOTATION_LITERALS_EMPTY_INT_ARRAY);
+                } else {
+                    retValue = valueMethod.newArray(componentName.toString(), valueMethod.load(0));
+                }
         }
         return retValue;
     }
 
     static String componentType(MethodInfo method) {
+        return componentTypeName(method).toString();
+    }
+
+    static DotName componentTypeName(MethodInfo method) {
         ArrayType arrayType = method.returnType().asArrayType();
-        return arrayType.component().name().toString();
+        return arrayType.component().name();
     }
 
     static String generatedSharedName(DotName annotationName) {
--- a/independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/FieldDescriptors.java
+++ b/independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/FieldDescriptors.java
@@ -1,5 +1,6 @@
 package io.quarkus.arc.processor;
 
+import io.quarkus.arc.impl.AnnotationLiterals;
 import io.quarkus.arc.impl.Qualifiers;
 import io.quarkus.gizmo.FieldDescriptor;
 import java.util.Set;
@@ -12,6 +13,22 @@ final class FieldDescriptors {
     static final FieldDescriptor QUALIFIERS_IP_QUALIFIERS = FieldDescriptor.of(Qualifiers.class, "IP_DEFAULT_QUALIFIERS",
             Set.class);
 
+    static final FieldDescriptor ANNOTATION_LITERALS_EMPTY_CLASS_ARRAY = FieldDescriptor.of(AnnotationLiterals.class,
+            "EMPTY_CLASS_ARRAY",
+            Class[].class);
+
+    static final FieldDescriptor ANNOTATION_LITERALS_EMPTY_STRING_ARRAY = FieldDescriptor.of(AnnotationLiterals.class,
+            "EMPTY_STRING_ARRAY",
+            String[].class);
+
+    static final FieldDescriptor ANNOTATION_LITERALS_EMPTY_LONG_ARRAY = FieldDescriptor.of(AnnotationLiterals.class,
+            "EMPTY_LONG_ARRAY",
+            long[].class);
+    
+    static final FieldDescriptor ANNOTATION_LITERALS_EMPTY_INT_ARRAY = FieldDescriptor.of(AnnotationLiterals.class,
+            "EMPTY_INT_ARRAY",
+            int[].class);
+    
     private FieldDescriptors() {
     }
 
--- a/independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/FieldDescriptors.java
+++ b/independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/FieldDescriptors.java
@@ -24,11 +24,11 @@ final class FieldDescriptors {
     static final FieldDescriptor ANNOTATION_LITERALS_EMPTY_LONG_ARRAY = FieldDescriptor.of(AnnotationLiterals.class,
             "EMPTY_LONG_ARRAY",
             long[].class);
-    
+
     static final FieldDescriptor ANNOTATION_LITERALS_EMPTY_INT_ARRAY = FieldDescriptor.of(AnnotationLiterals.class,
             "EMPTY_INT_ARRAY",
             int[].class);
-    
+
     private FieldDescriptors() {
     }
 
--- a/independent-projects/arc/runtime/src/main/java/io/quarkus/arc/impl/AbstractInvocationContext.java
+++ b/independent-projects/arc/runtime/src/main/java/io/quarkus/arc/impl/AbstractInvocationContext.java
@@ -5,6 +5,7 @@ import java.lang.annotation.Annotation;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.Method;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -43,7 +44,7 @@ abstract class AbstractInvocationContext implements ArcInvocationContext, Suppli
 
     @Override
     public Set<Annotation> getInterceptorBindings() {
-        return interceptorBindings;
+        return Collections.unmodifiableSet(interceptorBindings);
     }
 
     @Override
--- a/independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/MethodDescriptors.java
+++ b/independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/MethodDescriptors.java
@@ -168,6 +168,12 @@ final class MethodDescriptors {
     static final MethodDescriptor COLLECTIONS_UNMODIFIABLE_SET = MethodDescriptor.ofMethod(Collections.class, "unmodifiableSet",
             Set.class, Set.class);
 
+    static final MethodDescriptor COLLECTIONS_SINGLETON = MethodDescriptor.ofMethod(Collections.class, "singleton",
+            Set.class, Object.class);
+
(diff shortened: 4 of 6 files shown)


mkouba:
This is a follow-up of https://github.com/quarkusio/quarkus/pull/6990.

I've added a simple microbenchmark here:
https://github.com/mkouba/arc-benchmarks/blob/master/src/main/java/io/quarkus/arc/benchmarks/SubclassInstantiationBenchmark.java

The results are:
```
1.2.0:
Benchmark                                                    Mode  Cnt       Score       Error  Units
SubclassInstantiationBenchmark.instantiateDependentSublass  thrpt   25  459530,775 ± 20652,591  ops/s

MASTER:
Benchmark                                                    Mode  Cnt        Score        Error  Units
SubclassInstantiationBenchmark.instantiateDependentSublass  thrpt   25  2516775,307 ± 110357,068  ops/s

This PR:
Benchmark                                                    Mode  Cnt        Score        Error  Units
SubclassInstantiationBenchmark.instantiateDependentSublass  thrpt   25  4117451,397 ± 126190,354  ops/s
```

Use cases with complex class hierarchy and many methods should yield even better results.

mkouba:
CC @bcluap

independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/AnnotationLiteralGenerator.java
```diff
@@ -241,14 +254,30 @@ static ResultHandle arrayValue(AnnotationValue value, BytecodeCreator valueMetho
...
                 }
-                retValue = valueMethod.newArray(componentType(method), valueMethod.load(0));
+                DotName componentName = componentTypeName(method);
+                // Use empty array constants for common component kinds
+                if (DotNames.CLASS.equals(componentName)) {
+                    retValue = valueMethod.readStaticField(FieldDescriptors.ANNOTATION_LITERALS_EMPTY_CLASS_ARRAY);
+                } else if (DotNames.STRING.equals(componentName)) {
+                    retValue = valueMethod.readStaticField(FieldDescriptors.ANNOTATION_LITERALS_EMPTY_STRING_ARRAY);
+                } else if (DotNames.LONG.equals(componentName)) {
+                    retValue = valueMethod.readStaticField(FieldDescriptors.ANNOTATION_LITERALS_EMPTY_STRING_ARRAY);
```
gsmet: Shouldn't you use the LONG version?

independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/AnnotationLiteralGenerator.java
```diff
@@ -241,14 +254,30 @@ static ResultHandle arrayValue(AnnotationValue value, BytecodeCreator valueMetho
...
+                } else if (DotNames.INTEGER.equals(componentName)) {
+                    retValue = valueMethod.readStaticField(FieldDescriptors.ANNOTATION_LITERALS_EMPTY_STRING_ARRAY);
```
gsmet: Shouldn't you use the INT version?

independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/FieldDescriptors.java
```diff
@@ -12,6 +13,22 @@
...
+            "EMPTY_CLASS_ARRAY",
+            Class[].class);
+
+    static final FieldDescriptor ANNOTATION_LITERALS_EMPTY_STRING_ARRAY = FieldDescriptor.of(AnnotationLiterals.class,
+            "EMPTY_STRING_ARRAY",
+            String[].class);
+
+    static final FieldDescriptor ANNOTATION_LITERALS_EMPTY_LONG_ARRAY = FieldDescriptor.of(AnnotationLiterals.class,
+            "EMPTY_LONG_ARRAY",
+            String[].class);
```
gsmet: Is it normal it's String here?

independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/FieldDescriptors.java
```diff
@@ -12,6 +13,22 @@
...
+    
+    static final FieldDescriptor ANNOTATION_LITERALS_EMPTY_INT_ARRAY = FieldDescriptor.of(AnnotationLiterals.class,
+            "EMPTY_INT_ARRAY",
+            String[].class);
```
gsmet: Same here?

independent-projects/arc/runtime/src/main/java/io/quarkus/arc/impl/AbstractInvocationContext.java
```diff
@@ -43,7 +44,7 @@ protected AbstractInvocationContext(Object target, Method method,
 
     @Override
     public Set<Annotation> getInterceptorBindings() {
-        return interceptorBindings;
+        return Collections.unmodifiableSet(interceptorBindings);
```
gsmet: Just to be sure, you call that one often or not? If you do better move that one to the constructor as they look fully immutable.

gsmet (changes requested):
I added a couple of comments noted with a quick scan.

I'll do some further testing later today.

johnaohara:
Taking a look now

independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/AnnotationLiteralGenerator.java
mkouba: Yes, of course.. I'll fix it.

independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/FieldDescriptors.java
mkouba: Eh, of course not. I was just too tired...

independent-projects/arc/runtime/src/main/java/io/quarkus/arc/impl/AbstractInvocationContext.java
mkouba: No. It's a consumer of the `InvocationContext` who can invoke this method inside the interceptor...

mkouba:
BTW there's a formatting error...

gsmet:
Using the original benchmark, I still see a couple of things:

 * a significant amount of time spent in `InterpreterRuntime::ldc(JavaThread*, bool)`, @johnaohara do you know what it means?
 * then we have: some hashmap manipulation, see below
 * and finally: `javax/transaction/Transactional_Shared_AnnotationLiteral.<init>`. Shouldn't it be instantiated only once in our case and not for every instance?

So for the HashMap, if I uncompile things, I get:
```
        HashMap hashMap = new HashMap();
        this.metadata = hashMap;
        List<InterceptorInvocation> list = Collections.singletonList(InterceptorInvocation.aroundInvoke((InjectableInterceptor)injectableInterceptor, (Object)object));
        Class[] arrclass = AnnotationLiterals.EMPTY_CLASS_ARRAY;
        Method method = Reflections.findMethod(GreetingResource.class, (String)"xxy4", (Class[])arrclass);
        Class[] arrclass2 = AnnotationLiterals.EMPTY_CLASS_ARRAY;
        Class[] arrclass3 = AnnotationLiterals.EMPTY_CLASS_ARRAY;
        Transactional.TxType txType = Transactional.TxType.REQUIRED;
        Set<Transactional_Shared_AnnotationLiteral> set = Collections.singleton(new Transactional_Shared_AnnotationLiteral(arrclass2, arrclass3, txType));
        SubclassMethodMetadata subclassMethodMetadata = new SubclassMethodMetadata(list, method, set);
        ((Map)hashMap).put("m1", subclassMethodMetadata);
        Class[] arrclass4 = AnnotationLiterals.EMPTY_CLASS_ARRAY;
        Method method2 = Reflections.findMethod(GreetingResource.class, (String)"ho9", (Class[])arrclass4);
        SubclassMethodMetadata subclassMethodMetadata2 = new SubclassMethodMetadata(list, method2, set);
        ((Map)hashMap).put("m2", subclassMethodMetadata2);
        Class[] arrclass5 = AnnotationLiterals.EMPTY_CLASS_ARRAY;
        Method method3 = Reflections.findMethod(GreetingResource.class, (String)"goy3", (Class[])arrclass5);
        SubclassMethodMetadata subclassMethodMetadata3 = new SubclassMethodMetadata(list, method3, set);
        ((Map)hashMap).put("m3", subclassMethodMetadata3);
        Class[] arrclass6 = AnnotationLiterals.EMPTY_CLASS_ARRAY;
        Method method4 = Reflections.findMethod(GreetingResource.class, (String)"xxhos2", (Class[])arrclass6);
        SubclassMethodMetadata subclassMethodMetadata4 = new SubclassMethodMetadata(list, method4, set);
        ((Map)hashMap).put("m4", subclassMethodMetadata4);
        Class[] arrclass7 = AnnotationLiterals.EMPTY_CLASS_ARRAY;
        Method method5 = Reflections.findMethod(GreetingResource.class, (String)"go6", (Class[])arrclass7);
        SubclassMethodMetadata subclassMethodMetadata5 = new SubclassMethodMetadata(list, method5, set);
        ((Map)hashMap).put("m5", subclassMethodMetadata5);
```
This looks like static metadata to me, at least at the class level? Or am I incorrect?

Also, can we get an approximation of the size of the map when initializing it? I see a significant amount of time spent in resizing the Map. Obviously, the example is a bit extreme but you would probably have one resize you could avoid even with 20 methods.

My 2 cents.

mkouba:
> This looks like static metadata to me, at least at the class level? Or am I incorrect?

So currently the `SubclassMethodMetadata` also holds a reference to the interceptor chains (instances of interceptors) that are bound to a specific subclass instance. We could try to split this class in two and cache some metadata in a static variable. But that would be the next step as it could be a little bit more complicated and I'm also not so sure it's the best way to cache so many things in statics.

>  I see a significant amount of time spent in resizing the Map.

It's definitely possible. I'm not sure if avoiding one resize operation would help in usual use case. But we can definitely give it a try! ;-)

> javax/transaction/Transactional_Shared_AnnotationLiteral.<init>. Shouldn't it be instantiated only once in our case and not for every instance?

It could be cached if we go for ^ static metadata.

> a significant amount of time spent in InterpreterRuntime::ldc(JavaThread*, bool)

I have no idea ;-)

mkouba:
@gsmet And thanks a lot for your input!

Did you observe any improvement?

gsmet:
I'm releasing things so I'm not really able to do a proper benchmark with numbers just to generate some profiles.

If it's easy to properly size the Map, I would say go for it. If not, then I suppose we can live with it. It's not the resizing that is an issue, it's more than you end up rehashing the entire map. Obviously it's more of an issue with this benchmark.

johnaohara:
@gsmet which tool are you using to profile the benchmark?  I have run with async profiler, comments below
 
> * a significant amount of time spent in `InterpreterRuntime::ldc(JavaThread*, bool)`, @johnaohara do you know what it means?

I don't see it in the profiling I have done, please can you share any data/graphs?

> * then we have: some hashmap manipulation, see below
> * and finally: `javax/transaction/Transactional_Shared_AnnotationLiteral.<init>`. Shouldn't it be instantiated only once in our case and not for every instance?
> 
Again, I don't see this so if you have anything you can share that would be great. 


> Also, can we get an approximation of the size of the map when initializing it? I see a significant amount of time spent in resizing the Map. Obviously, the example is a bit extreme but you would probably have one resize you could avoid even with 20 methods.
> 
This is taking 5.6% of cpu time so it if we could determine inital size of map, then benchmark throughput would improve slightly

johnaohara:
@gsmet sry, forgot `InterpreterRuntime::ldc` is the interpreter executing the LDC bytecode instruction, am surprised to see this in profiling, would have thought that would have been compiled by JIT

johnaohara:
@mkouba Apart from the HashMap resize, I don't see any other issues with this PR. I have spotted something else that I think can be improved, but is in addition to these changes. I will open a separate PR for those changes

gsmet:
@johnaohara it's with this benchmark: https://github.com/quarkusio/quarkus/pull/6990#issuecomment-582043343 (see this comment and the one below). This is not very realistic but it allows to see some extreme behaviors.

Here is the profile I got:

[profile.zip](https://github.com/quarkusio/quarkus/files/4160996/profile.zip)

gsmet:
I think it might be due to the 160 functions class generated.

bcluap:
That example was just to show the extent of the issue. I agree it's not
realistic but it does help to understand the impact of any improvements.

Paul Carter-Brown
Director
Jini Guru
m: +27 (0) 83 442 7179 <+27834427179>
a: 1st Floor, Golf House, Design Quarter, Cnr. William Nicol and Leslie
  Johannesburg, South Africa
w: jini.guru  e: paul@jini.guru

Disclaimer: This message and/or attachment(s) may contain
privileged, confidential and/or personal information. If you are not the
intended recipient you may not disclose or distribute any of
the information contained within this message. In such case you must
destroy this message and inform the sender of the error. Jini Guru may not
accept liability for any errors, omissions, information and viruses
contained in the transmission of this message. Any opinions, conclusions
and other information contained within this message not related to Jini
Guru official business is deemed to be that of the individual only and is
not endorsed by Jini Guru.

gsmet:
@bcluap yes, exactly. It helped narrowing what was wrong. It's just that in the case of the interpreted kicking in, it might be due to having the 160 function classes. But still worth some feedback from @johnaohara .

johnaohara:
@gsmet thanks, looking now

mkouba:
@gsmet Setting the initial capacity of the map quite improved the results!

I've added a benchmark for a little more complex subclass:
https://github.com/mkouba/arc-benchmarks/blob/master/src/main/java/io/quarkus/arc/benchmarks/SubclassInstantiationBenchmark.java#L46-L52

The results [1] are ~ 9x better for simple use case and ~ 70x better for complex use case. I think that's good enough for this round. I'd like to merge this PR now (if CI passes) and continue with improvements in the next round. WDYT?

@bcluap Thanks again for spotting this problem! 
@gsmet Thanks again for all your tips.
@johnaohara Hopefully, you'll be able to verify my findings... ;-)

[1]
```
1.2.0:
Benchmark                                        Mode  Cnt       Score       Error  Units
SubclassInstantiationBenchmark.complexSubclass  thrpt   25   14724,932 ±   229,155  ops/s
SubclassInstantiationBenchmark.simpleSubclass   thrpt   25  475942,048 ± 46076,935  ops/s

999-SNAPSHOT:
Benchmark                                        Mode  Cnt        Score        Error  Units
SubclassInstantiationBenchmark.complexSubclass  thrpt   25  1082657,197 ±   8218,298  ops/s
SubclassInstantiationBenchmark.simpleSubclass   thrpt   25  4117640,138 ± 202821,409  ops/s
```

gsmet (approved):
I totally agree with the approach, let's make some iterative improvements.

Results already are impressive!

gsmet:
Here are the results I got from this morning on my laptop with @bcluap 's benchmark:

# 1.2.0.Final

```
Concurrency Level:      10
Time taken for tests:   19.517 seconds
Complete requests:      100000
Failed requests:        0
Total transferred:      8000000 bytes
HTML transferred:       200000 bytes
Requests per second:    5123.81 [#/sec] (mean)
Time per request:       1.952 [ms] (mean)
Time per request:       0.195 [ms] (mean, across all concurrent requests)
Transfer rate:          400.30 [Kbytes/sec] received
```

# current master

```
Concurrency Level:      10
Time taken for tests:   8.745 seconds
Complete requests:      100000
Failed requests:        0
Total transferred:      8000000 bytes
HTML transferred:       200000 bytes
Requests per second:    11435.62 [#/sec] (mean)
Time per request:       0.874 [ms] (mean)
Time per request:       0.087 [ms] (mean, across all concurrent requests)
Transfer rate:          893.41 [Kbytes/sec] received
```

The profile is far better than before. I think the only way to improve it now is to radically change things to initialize things statically but that's for another day.

bcluap:
I think I need a new laptop :-)

Paul Carter-Brown
Director
Jini Guru
m: +27 (0) 83 442 7179 <+27834427179>
a: 1st Floor, Golf House, Design Quarter, Cnr. William Nicol and Leslie
  Johannesburg, South Africa
w: jini.guru  e: paul@jini.guru

Disclaimer: This message and/or attachment(s) may contain
privileged, confidential and/or personal information. If you are not the
intended recipient you may not disclose or distribute any of
the information contained within this message. In such case you must
destroy this message and inform the sender of the error. Jini Guru may not
accept liability for any errors, omissions, information and viruses
contained in the transmission of this message. Any opinions, conclusions
and other information contained within this message not related to Jini
Guru official business is deemed to be that of the individual only and is
not endorsed by Jini Guru.

gsmet:
@bcluap and it's on a nearly 4 years old laptop :).

PR #7004, opened 2020-02-05
Maven create extension test fixes

aloubyansky:
* compare the project files after the resource filtering;
* escape properties that are not supposed to be replaced during resource filtering


gsmet (approved):
Tested locally.

PR #7006, opened 2020-02-05
Fixes callout and property value cache from doc

lordofthejars:


docs/src/main/asciidoc/cache.adoc
```diff
@@ -324,8 +324,8 @@ Here's what your cache configuration could look like:
 [source,properties]
 ----
 quarkus.cache.caffeine."foo".initial-capacity=10 <1>
-quarkus.cache.caffeine."foo".maximum-size=20 <1>
-quarkus.cache.caffeine."foo".expire-after-write <1>
```
gsmet: I think the callouts were on purpose. Do you have a warning/error? /cc @gwenneg

lordofthejars: But then they should be <1>, <2>, ... Because if not it looks really strange when they are rendered.

gwenneg: There's a missing property value indeed, but I don't see any problem with the callouts. The same one is repeated three times on purpose since the same cache is being configured and it doesn't generate any warning/error AFAIK.

gwenneg: image

This looks fine to me (except for the missing value).

lordofthejars: Well now we are entering to the subjective things for me it is weird, but of course I understand that for somebody else this looks good :)

gwenneg: I wouldn't mind removing the 2nd and 3rd callout if it makes the doc clearer :)

gwenneg (approved):

docs/src/main/asciidoc/cache.adoc
lordofthejars: In the end, it is my personal opinion, but of course, I might be wrong.

gwenneg:
Considering your amazing work on the Quarkus Cheat-Sheet, I think your opinion matters more than mine here :)

PR #7009, opened 2020-02-05
Fix the native image path passed to NativeImageBuildItem

geoand:


geoand:
I stumbled upon this when looking into Jib support for the native binary. The code that my PR changes seems wrong, but I may be missing something

aloubyansky (approved):

stuartwdouglas (approved):

PR #7010, opened 2020-02-05
Adjust the artifact name of the PDF documentation to avoid conflicts

gsmet:
The final name is not used when pushing to Central and we need a
classifier to avoid conflicts.
The good thing is that we will be able to point the downloads to Central
instead of serving them ourselves.

I just noticed that when doing the release so better fix it now rather than never.


geoand:
This seems reasonable to me but I assume there is no way to test?

gsmet:
You mean test it yourself? Basically, with the old way of doing things, the zip artifact was overwriting the other artifact because the final name is ignored when pushing the artifact and the classifier was unused. Now the artifact is installed with a classifier. Just do a `mvn -Ddocumentation-pdf clean install` in the `docs/` directory and you should see it.

geoand (approved):
Excellent! Tested and verified :)

PR #7011, opened 2020-02-05
Allow specifying quarkus.native.builder-image via gradle plugin parameter native-builder-image

andreas-repp:
Currently it is not possible to specify this in the gradle file. Especially for multi-module builds, it is nice to be able to set this via gradle in the main module.


gastaldi (approved):
LGTM

PR #7012, opened 2020-02-05
Check to see if injection point is same as current thread local value

johnaohara:
While running a benchmark [1] for changes in https://github.com/quarkusio/quarkus/pull/7003 28.9% of cpu time was spent in `io/quarkus/arc/impl/InjectionPointProvider.set()`.  By checking whether the same injectionPoint passed to InjectionPointProvider.set() was already stored in the ThreadLocal, cpu time for this method call dropped to 13.25%. 

@mkouba How likely is the scenario that the injectionPoint passed to `io/quarkus/arc/impl/InjectionPointProvider.set()` will be the same? Or is it just a consequence of this particular benchmark? I am not certain that in a real-world scanerio whether this change will make as big a difference for this method call

1 - https://github.com/mkouba/arc-benchmarks/blob/master/src/main/java/io/quarkus/arc/benchmarks/SubclassInstantiationBenchmark.java

### DIFF
--- a/independent-projects/arc/runtime/src/main/java/io/quarkus/arc/impl/InjectionPointProvider.java
+++ b/independent-projects/arc/runtime/src/main/java/io/quarkus/arc/impl/InjectionPointProvider.java
@@ -26,8 +26,12 @@ public class InjectionPointProvider implements InjectableReferenceProvider<Injec
     static InjectionPoint set(InjectionPoint injectionPoint) {
         if (injectionPoint != null) {
             InjectionPoint prev = InjectionPointProvider.CURRENT.get();
-            InjectionPointProvider.CURRENT.set(injectionPoint);
-            return prev;
+            if (injectionPoint.equals(prev)) {
+                return injectionPoint;
+            } else {
+                InjectionPointProvider.CURRENT.set(injectionPoint);
+                return prev;
+            }
         } else {
             CURRENT.remove();
             return null;


independent-projects/arc/runtime/src/main/java/io/quarkus/arc/impl/InjectionPointProvider.java
```diff
@@ -25,9 +25,13 @@ public InjectionPoint get(CreationalContext<InjectionPoint> creationalContext) {
      */
     static InjectionPoint set(InjectionPoint injectionPoint) {
         if (injectionPoint != null) {
-            InjectionPoint prev = InjectionPointProvider.CURRENT.get();
-            InjectionPointProvider.CURRENT.set(injectionPoint);
-            return prev;
+            if (injectionPoint.equals(InjectionPointProvider.CURRENT.get())) {
+                return injectionPoint;
+            } else {
+                InjectionPoint prev = InjectionPointProvider.CURRENT.get();
```
gsmet: You could move that one outside of the `else` and call `InjectionPointProvider.CURRENT.get()` only once in all cases. Or do I miss something?

gsmet: You could move that one outside of the else and call InjectionPointProvider.CURRENT.get() only once in all cases. Or do I miss something?

gsmet (changes requested):

independent-projects/arc/runtime/src/main/java/io/quarkus/arc/impl/InjectionPointProvider.java
johnaohara: Good point, I will update the PR

gsmet (approved):

gsmet:
Looks better, thanks! Let's wait for @mkouba 's blessing!

mkouba:
Hm, @johnaohara I've tried to print a log message in the `injectionPoint.equals(prev)` branch and I don't see a single output message when running the `SubclassInstantiationBenchmark` and the arc test suite... I'm not saying it's impossible but I just can't think of such a use case. On the other, it wouldn't hurt to merge this PR so if it helps I'm +1 ;-).

johnaohara:
these are the flamegraphs of before and after this change;

before:
Screenshot from 2020-02-06 08-47-50

after:
Screenshot from 2020-02-06 08-47-51

PR #7015, opened 2020-02-05
make ide-config shareable

maxandersen:
Why:

 * ide-config currently replicated at `ide-config`, `ip/arc/ide-config` and `ip/qute/ide-config`
 * quickstarts could benefit from reusing quarkus formatting

 * move ide-config to independent-projects and make it a real project
 * update refs to `eclipse-format.xml` to use <dependency> and resource lookup
 * update CONTRIBUTING.MD


maxandersen:
this is to enforce formatting done in https://github.com/quarkusio/quarkus-quickstarts/pull/451 stays.

gsmet (approved):
Looks good to me. I fixed a few XML formatting issues and amended your commit.

Now that we have that, I wonder if we could apply it to the other independent projects? We would probably need to sync with Alexey so that it is not in his way.

maxandersen:
> Now that we have that, I wonder if we could apply it to the other independent projects? We would probably need to sync with Alexey so that it is not in his way.

this PR updated independent-projects to use it so already done - since ide-config is changed to independnet project i don't see @aloubyansky or @mkouba having a problem with it (crossing fingers :)

gsmet:
AFAIK it's not used in `tools` and `bootstrap`, thus my request and my concern about it being potentially disruptive for Alexey's work.

aloubyansky:
That's ok, let's do it now.

maxandersen:
> AFAIK it's not used in `tools` and `bootstrap`, thus my request and my concern about it being potentially disruptive for Alexey's work.

yes, didn't change them as they had nothing setup as fear of circular/quarkus dependency. with this change we can ;)

maxandersen:
Pushed update of ide-config that includes enabling it for tools/bootstrap
+ a separate commit where the formatting is applied to tools/bootstrap (that’s why the file count sky rocketed here)

gastaldi:
Looks like this PR needs to be rebased

maxandersen:
rebased - please ack so we can get this merged ;)

gastaldi:
Oops, you'll need to rebase this again @maxandersen, sorry about that ;)

maxandersen:
rebased @gastaldi

PR #7029, opened 2020-02-06
Add back EnvConfigSource that was removed by accident

geoand:
This fixes failing Quickstarts CI


gastaldi (approved):

dmlloyd (approved):

gsmet (approved):
@geoand I think we should migrate all the projects and quickstarts to use the new way of configuring things for native images.

That's definitely not something we want to break again. I know you had something in the works for that so +1 from me on this.

geoand:
@gsmet unfortunately for the integration tests that would not have caught this issue because we use the container build type for the native deployment.

Regardless, we should definitely do what you propose.

gsmet:
Yeah, I'm still torn on whether we shoud move to using a proper GraalVM install or stick with containers when moving to GH actions. In the end we want to test both.

So I suppose the current situation with core on container and quickstarts on install is as good as it can be.

But yeah, we need to migrate massively to the new config.

PR #7032, opened 2020-02-06
Revert "Make the Kafka health check configuration more future proof"

gsmet:
This reverts commit 397424e2f1d1578348aeb08d47fd509946cad2a2.

It was a bad idea of mine. I really don't like us not using Quarkus
config but making things inconsistent and requiring another property is
an even worse idea.

/cc @geoand per our discussion
/cc @loicmathieu and @zeljkot


geoand (approved):
I agree that is the best possible thing we can do in a tough situation

loicmathieu (approved):
LGTM

zeljkot:
I cannot recall all the details (I guess it was about duplicate broker key), but IMHO health check should be the part of microprofile reactive messaging Kafka. An unavailable broker is just one way how a subsystem can get sick. Thrown exception or serde issue can break the reactive part and then no further messages can be processed. That kind of stuff should be also taken into account.

Health OK = I can send and receive messages.
I cannot send and receive messages = health NOK.

loicmathieu:
@zeljkot this health check is on kafka side, not on reactive-messaging-kafka so it cannot do more than check if the kafka cluster is healthy.

aguibert:
just saw this issue. Perhaps we could introduce a way to document "external properties" like `kafka.bootstrap.servers` on the quarkus all-config page? This would be very helpful because it would still collect relevant/required external config options for the user in one place.

zeljkot:
> @zeljkot this health check is on kafka side, not on reactive-messaging-kafka so it cannot do more than check if the kafka cluster is healthy.

I agree, but my point is that check is in the wrong component. Kafka brokers are configured in reactive messaging configuration, so IMHO health checks should also be also part of it. In that case, it can do more than checking is Kafka alive, like checking are streams still processing the messages. I got burnt by the resilience of reactive messaging, so it would be nice if it can at least initiate a restart if something is broken.

loicmathieu:
@zeljkot health check are designed to be multiple, if you feel that the existing ones around Kafka are not enought you can open an issue to create new ones.

Currently we have :

- A check to the cluster to switc the readiness probe off if the cluster is not available (disabled by default)
- A check inside the `kafka-stream` extension to check at the kafka-stream level

zeljkot:
Having separate probes for Kafka cluster and reactive messaging is also OK, if they all use the same configuration parameters for common stuff, like brokers. By "all" I mean Kafka health + Reactive messaging health + Reactive messaging actual messaging.

PR #7037, opened 2020-02-06
Build reproducibility #7025 #7026 #7027

### LINKED ISSUE #7025: Output property files reproducibly
`java.util.Properties.store(...)` has at least three reproducibility issues:

* Ordering of the entries in the file is non-reproducible due to `ConcurrentHashMap` being used internally
* Usage of System.getProperty("line.separator") as a line sepatator
* Always writing a timestamp into the file

This causes that jars produced by our projects change with every build even if there was no change in the source tree. 

There are two property files that need to get produced reproducibly: 

* META-INF/quarkus-extension.properties
* META-INF/quarkus-javadoc.properties

A PR follows.

### LINKED ISSUE #7026: Set project.build.outputTimestamp for reproducible JARs
`project.build.outputTimestamp` property signals to `maven-jar-plugin` 3.2.0+ that it should switch to the reproducible mode. Among other things, this makes the plugin to use the timestamp stored in the named Maven property for the entries in the ZIP file. See https://blogs.apache.org/maven/entry/apache-maven-jar-plugin-version3

A PR follows

### LINKED ISSUE #7027: Output runner JARs reproducibly
We write into several JAR files during our augmentation process. The ZIP entries get timestamps of the instant when they were written. Therefore, when repeating the build from an unchanged source tree, the resulting JAR files are binary different for every build. 

We should take the same strategy as the maven-jar-plugin - i.e. take the stable value stored in the `project.build.outputTimestamp` property in the pom.xml file - see https://github.com/quarkusio/quarkus/issues/7026

A PR follows.

ppalaga:
Fixes #7025
Fixes #7026 
Fixes #7027

### DIFF
--- a/build-parent/pom.xml
+++ b/build-parent/pom.xml
@@ -85,6 +85,13 @@
         <!-- Code Coverage Properties-->
         <jacoco.agent.argLine></jacoco.agent.argLine>
         <jacoco.version>0.8.5</jacoco.version>
+
+        <version.jar.plugin>3.2.0</version.jar.plugin><!-- This can be removed after jboss-parent 37 -->
+        <version.release.plugin>3.0.0-M1</version.release.plugin><!-- This can be removed after jboss-parent 37 -->
+
+        <!-- Reproducible builds see https://maven.apache.org/guides/mini/guide-reproducible-builds.html -->
+        <!-- project.build.outputTimestamp is updated by maven-release-plugin during release -->
+        <project.build.outputTimestamp>2020-01-23T09:06:00Z</project.build.outputTimestamp>
     </properties>
 
     <dependencyManagement>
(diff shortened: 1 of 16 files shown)


ppalaga:
@aloubyansky could you please review?

@gsmet not sure who else should review?

ppalaga:
As agreed on the developer mailing list, Quarkus build reproducibility is not required at this time, see https://groups.google.com/forum/#!topic/quarkus-dev/6geWtb5HOdc Thus closing.

In case this becomes a topic in the future, here is my latest code https://github.com/ppalaga/quarkus/commits/native-image-cache-and-build-reproducibility

gsmet:
I included the fixes in this PR in my current work regarding reproducibility.

Too many things had changed to use this PR but I think I was able to port the changes to the new code, one way or another.

PR #7050, opened 2020-02-06
Allow the standalone use of @Startup

### LINKED ISSUE #7048: Allow the standalone use of @Startup
Currently the new `@Startup` annotation requires a scope annotation to go along with it.

It was [suggested](https://github.com/quarkusio/quarkus/pull/6969#issuecomment-582784981) to make `@Startup` a standalone annotation that is `@ApplicationScoped` by default

geoand:
Fixes: #7048

### DIFF
--- a/extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/StartupBuildSteps.java
+++ b/extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/StartupBuildSteps.java
@@ -46,6 +46,11 @@ public class StartupBuildSteps {
             "destroy", void.class, Object.class, CreationalContext.class);
 
     @BuildStep
+    BeanDefiningAnnotationBuildItem beanDefiningAnnotation() {
+        return new BeanDefiningAnnotationBuildItem(STARTUP_NAME);
+    }
+
+    @BuildStep
     UnremovableBeanBuildItem unremovableBeans() {
         // Make all classes annotated with @Startup unremovable
         return new UnremovableBeanBuildItem(new Predicate<BeanInfo>() {


geoand:
@mkouba I didn't add the `ApplicationScoped` as the default scope in the BeanDefiningBuildItem because when I do so the test fails:

```
Caused by: javax.enterprise.inject.spi.DefinitionException: Producer method for a normal scoped bean must not have a return type that is final: PRODUCER METHOD bean [types=[java.io.Serializable, java.lang.String, java.lang.CharSequence, java.lang.Object, java.lang.Comparable<java.lang.String>], qualifiers=[@Default, @Any], target=java.lang.String produceString(), declaringBean=io.quarkus.arc.test.startup.StartupAnnotationTest$ProducerStartMe]
```

That seems to me to indicate that a bean producer method and the BeanDefiningAnnotation don't play well nicely?
Because it seems to me that what happened is that because there was not scope on `@Startup`, `@ApplicationScoped` was added even to the producer method, which probably shouldn't have been the case. WDYT?

mkouba:
Hm, bean defining annotations should have no effect on producer methods/fields. I'll take a look.

mkouba:
Ok, we have a bug here:
https://github.com/quarkusio/quarkus/blob/master/independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/BeanDeployment.java#L517

A bean defining annotation with the default scope is registered as a stereotype...

mkouba:
Anyway, for `Startup` we could just use an annotation transformer. Something like this: https://github.com/quarkusio/quarkus/blob/master/extensions/scheduler/deployment/src/main/java/io/quarkus/scheduler/deployment/SchedulerProcessor.java#L95-L117.

geoand:
> Anyway, for `Startup` we could just use an annotation transformer. Something like this: https://github.com/quarkusio/quarkus/blob/master/extensions/scheduler/deployment/src/main/java/io/quarkus/scheduler/deployment/SchedulerProcessor.java#L95-L117.

I tried that as well :). And when I did the tests failed because the methods didn't seem to be called...

mkouba:
Ok, I'll try it right now ;-).

geoand:
Thanks :)

mkouba:
Closing, superseded by https://github.com/quarkusio/quarkus/pull/7058.

@geoand the problem was you can't consider `org.jboss.jandex.ClassInfo.annotations()` in the transformer but only those declared on the class...

geoand:
Aha, thanks!

nimo23:
@mkouba > Ok, we have a bug here: A bean defining annotation with the default scope is registered as a stereotype...

In https://docs.jboss.org/weld/reference/latest/en-US/html/stereotypes.html it is said that:

> 
> A stereotype encapsulates any combination of:
> 
>     a default scope, and 

According that it is not said that a default scope encapsulates a stereotype but vice versa. Is it because of that a bug ?

mkouba:
> A bean defining annotation with the default scope is registered as a stereotype...

I'm talking about quarkus-specific stuff, i.e. an implementation detail of `io.quarkus.arc.deployment.BeanDefiningAnnotationBuildItem`. It's not 100% correct but it should work in most cases.

nimo23:
>I'm talking about quarkus-specific stuff..It's not 100% correct but it should work in most cases.

Ok, got it. So it did not worked in this case but it works almost in most cases:) Would be maybe better to resolve this bug..anyway thanks to make it `@Startup` possible!

mkouba:
Yeah, the problem is that we did not expect a bean defining annotation to be used on a producer method... which makes sense, because bean defining annotations must be declared on a class... but it does not work in this particular case ;-)

Issue #7014, opened 2020-02-05
Building native docker image fails due to Failed to augment application classes error


kamilgregorczyk:
**Describe the bug**
I created a simple project via the web wizard with Gradle and creating a native image works by running `./gradlew buildNative` but doing the same thing with building docker by running ` ./gradlew buildNative --docker-build=true ` does not work, it ends with such error:
```
➜  blog ./gradlew buildNative --docker-build=true --stacktrace        

> Task :quarkusBuild
building quarkus runner

> Task :buildNative
building native image

> Task :buildNative FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':buildNative'.
> Failed to generate a native image

* Try:
Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Exception is:
org.gradle.api.tasks.TaskExecutionException: Execution failed for task ':buildNative'.
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.lambda$executeIfValid$1(ExecuteActionsTaskExecuter.java:187)
        at org.gradle.internal.Try$Failure.ifSuccessfulOrElse(Try.java:263)
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.executeIfValid(ExecuteActionsTaskExecuter.java:185)
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.execute(ExecuteActionsTaskExecuter.java:166)
        at org.gradle.api.internal.tasks.execution.CleanupStaleOutputsExecuter.execute(CleanupStaleOutputsExecuter.java:109)
        at org.gradle.api.internal.tasks.execution.FinalizePropertiesTaskExecuter.execute(FinalizePropertiesTaskExecuter.java:46)
        at org.gradle.api.internal.tasks.execution.ResolveTaskExecutionModeExecuter.execute(ResolveTaskExecutionModeExecuter.java:62)
        at org.gradle.api.internal.tasks.execution.SkipTaskWithNoActionsExecuter.execute(SkipTaskWithNoActionsExecuter.java:57)
        at org.gradle.api.internal.tasks.execution.SkipOnlyIfTaskExecuter.execute(SkipOnlyIfTaskExecuter.java:56)
        at org.gradle.api.internal.tasks.execution.CatchExceptionTaskExecuter.execute(CatchExceptionTaskExecuter.java:36)
[...]

PR #7051, opened 2020-02-06
Add tooling guides tip about JDK 11 native image generation from Docker

gwenneg:
Partially addresses #7014.

I'm not sure it's ok to mention the full `ubi-quarkus-native-image` URL like that in the doc, but this is only a temporary tip since the JDK 8 support will be dropped in Quarkus in the near future hopefully.

@gsmet WDYT?

### DIFF
--- a/docs/src/main/asciidoc/gradle-tooling.adoc
+++ b/docs/src/main/asciidoc/gradle-tooling.adoc
@@ -277,9 +277,15 @@ The easiest way to supply custom configuration is via the command line. For exam
 ./gradlew buildNative --docker-build=true
 ----
 
-The produced executable will be a 64 bit Linux executable, so depending on your operating system
-it may no longer be runnable.
+The produced executable will be a 64 bit Linux executable, so depending on your operating system it may no longer be runnable.
 However, it's not an issue as we are going to copy it to a Docker container.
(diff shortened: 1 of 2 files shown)


gwenneg:
FYI, I checked that the `{graalvm-version}` string is correctly replaced by the GraalVM version when the doc is built:

image

I also noticed the Gradle tooling guide contains deprecated build instructions using `QuarkusNative` which should have been deprecated at the same time as `NativeImageMojo`. I posted [something about it on Zulip](https://quarkusio.zulipchat.com/#narrow/stream/187030-users/topic/configure.20builderImage.20via.20gradle/near/187596485), waiting for a confirmation now.

gastaldi (approved):

geoand:
We should certainly update the Gradle tooling docs as well to no longer point to the Gradle tasks

gwenneg:
@geoand I agree, but it looks like there's more to do than simply update the doc. For example, if `QuarkusNative` is officially deprecated, we have to do something about `QuarkusTestNative` which depends on `QuarkusNative`. I'd rather have this subject treated in another PR :)

gwenneg:
I created #7104 for that purpose.

Let's merge this PR if that's ok with everyone.

PR #7056, opened 2020-02-07
enable code formatting and import sorting on tools/bootstrap

maxandersen:


maxandersen:
might as well just do it in #7015

PR #7058, opened 2020-02-07
Turn a class annotated with @Startup but no scope into a bean

### LINKED ISSUE #7048: Allow the standalone use of @Startup

mkouba:
- add @ApplicationScope automatically
- resolves #7048

### DIFF
--- a/extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/StartupBuildSteps.java
+++ b/extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/StartupBuildSteps.java
@@ -2,12 +2,14 @@ package io.quarkus.arc.deployment;
 
 import java.util.function.Predicate;
 
+import javax.enterprise.context.ApplicationScoped;
 import javax.enterprise.context.spi.Contextual;
 import javax.enterprise.context.spi.CreationalContext;
 
 import org.jboss.jandex.AnnotationInstance;
 import org.jboss.jandex.AnnotationValue;
 import org.jboss.jandex.DotName;
+import org.jboss.logging.Logger;
 
 import io.quarkus.arc.Arc;
 import io.quarkus.arc.ArcContainer;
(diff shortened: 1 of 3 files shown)


geoand (approved):
Weird, I tried this and it didn't work for me. I must have made some subtle mistake :)

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/StartupBuildSteps.java
```diff
@@ -45,6 +49,31 @@
...
+        return new AnnotationsTransformerBuildItem(new AnnotationsTransformer() {
+
+            @Override
+            public boolean appliesTo(org.jboss.jandex.AnnotationTarget.Kind kind) {
+                return kind == org.jboss.jandex.AnnotationTarget.Kind.CLASS;
+            }
+
+            @Override
+            public void transform(TransformationContext context) {
+                if (context.isClass() && !BuiltinScope.isDeclaredOn(context.getTarget().asClass())) {
```
manovotn: I know it's somewhat far-fetched but should we care about non-built in scopes? Users can have custom ones, or there is `Transactional` one already present. Both cases will pass this check and try to add yet another scope on the bean.

manovotn (commented):
Ah, I see you beat me to this PR. Was planning on sending it but got distracted by other issues...
I left one comment, otherwise it looks good as there is no black magic in this PR :)

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/StartupBuildSteps.java
mkouba: That's a good point. Although in this case I doubt that a `@Startup` bean would be annotated with `@Transactional` or a custom scope. Let's create a separate issue to handle this corretly. WDYT?

mkouba: Hm, the problem is that annotation transformers are registered before the final set of scopes is known...

mkouba: I've created https://github.com/quarkusio/quarkus/issues/7076

PR #7059, opened 2020-02-07
Update to SmallRye Reactive Messaging 1.1.0

cescoffier:
This version brings a new set of features:

* Metadata propagation
* New Emitter support (part of the spec, the old one is deprecated but still supported)
* `@Stream` has been removed (already deprecated)

### DIFF
--- a/extensions/smallrye-reactive-messaging/deployment/src/main/java/io/quarkus/smallrye/reactivemessaging/deployment/DotNames.java
+++ b/extensions/smallrye-reactive-messaging/deployment/src/main/java/io/quarkus/smallrye/reactivemessaging/deployment/DotNames.java
@@ -17,12 +17,12 @@ public final class DotNames {
     static final DotName INCOMING = DotName.createSimple(Incoming.class.getName());
     static final DotName OUTGOING = DotName.createSimple(Outgoing.class.getName());
     static final DotName CHANNEL = DotName.createSimple(org.eclipse.microprofile.reactive.messaging.Channel.class.getName());
-    static final DotName DEPRECATED_CHANNEL = DotName.createSimple(Channel.class.getName());
+    static final DotName LEGACY_CHANNEL = DotName.createSimple(Channel.class.getName());
     static final DotName EMITTER = DotName.createSimple(org.eclipse.microprofile.reactive.messaging.Emitter.class.getName());
-    static final DotName DEPRECATED_EMITTER = DotName.createSimple(Emitter.class.getName());
+    static final DotName LEGACY_EMITTER = DotName.createSimple(Emitter.class.getName());
     static final DotName ON_OVERFLOW = DotName
             .createSimple(org.eclipse.microprofile.reactive.messaging.OnOverflow.class.getName());
-    static final DotName DEPRECATED_ON_OVERFLOW = DotName.createSimple(OnOverflow.class.getName());
+    static final DotName LEGACY_ON_OVERFLOW = DotName.createSimple(OnOverflow.class.getName());
     static final DotName ACKNOWLEDGMENT = DotName.createSimple(Acknowledgment.class.getName());
     static final DotName MERGE = DotName.createSimple(Merge.class.getName());
     static final DotName BROADCAST = DotName.createSimple(Broadcast.class.getName());
--- a/extensions/smallrye-reactive-messaging/deployment/src/main/java/io/quarkus/smallrye/reactivemessaging/deployment/DotNames.java
+++ b/extensions/smallrye-reactive-messaging/deployment/src/main/java/io/quarkus/smallrye/reactivemessaging/deployment/DotNames.java
@@ -17,12 +17,12 @@ public final class DotNames {
     static final DotName INCOMING = DotName.createSimple(Incoming.class.getName());
     static final DotName OUTGOING = DotName.createSimple(Outgoing.class.getName());
     static final DotName CHANNEL = DotName.createSimple(org.eclipse.microprofile.reactive.messaging.Channel.class.getName());
-    static final DotName OLD_CHANNEL = DotName.createSimple(Channel.class.getName());
+    static final DotName DEPRECATED_CHANNEL = DotName.createSimple(Channel.class.getName());
     static final DotName EMITTER = DotName.createSimple(org.eclipse.microprofile.reactive.messaging.Emitter.class.getName());
-    static final DotName OLD_EMITTER = DotName.createSimple(Emitter.class.getName());
+    static final DotName DEPRECATED_EMITTER = DotName.createSimple(Emitter.class.getName());
     static final DotName ON_OVERFLOW = DotName
             .createSimple(org.eclipse.microprofile.reactive.messaging.OnOverflow.class.getName());
-    static final DotName OLD_ON_OVERFLOW = DotName.createSimple(OnOverflow.class.getName());
+    static final DotName DEPRECATED_ON_OVERFLOW = DotName.createSimple(OnOverflow.class.getName());
     static final DotName ACKNOWLEDGMENT = DotName.createSimple(Acknowledgment.class.getName());
     static final DotName MERGE = DotName.createSimple(Merge.class.getName());
     static final DotName BROADCAST = DotName.createSimple(Broadcast.class.getName());
--- a/extensions/smallrye-reactive-messaging/deployment/src/main/java/io/quarkus/smallrye/reactivemessaging/deployment/DotNames.java
+++ b/extensions/smallrye-reactive-messaging/deployment/src/main/java/io/quarkus/smallrye/reactivemessaging/deployment/DotNames.java
@@ -10,6 +10,7 @@ import io.smallrye.reactive.messaging.annotations.Channel;
 import io.smallrye.reactive.messaging.annotations.Emitter;
 import io.smallrye.reactive.messaging.annotations.Merge;
 import io.smallrye.reactive.messaging.annotations.OnOverflow;
+import io.smallrye.reactive.messaging.metrics.MetricDecorator;
 
 public final class DotNames {
 
@@ -26,6 +27,8 @@ public final class DotNames {
     static final DotName ACKNOWLEDGMENT = DotName.createSimple(Acknowledgment.class.getName());
     static final DotName MERGE = DotName.createSimple(Merge.class.getName());
     static final DotName BROADCAST = DotName.createSimple(Broadcast.class.getName());
+
+    static final DotName METRIC_DECORATOR = DotName.createSimple(MetricDecorator.class.getName());
 
     private DotNames() {
     }
--- a/extensions/smallrye-reactive-messaging/deployment/src/main/java/io/quarkus/smallrye/reactivemessaging/deployment/DotNames.java
+++ b/extensions/smallrye-reactive-messaging/deployment/src/main/java/io/quarkus/smallrye/reactivemessaging/deployment/DotNames.java
@@ -10,17 +10,19 @@ import io.smallrye.reactive.messaging.annotations.Channel;
 import io.smallrye.reactive.messaging.annotations.Emitter;
 import io.smallrye.reactive.messaging.annotations.Merge;
 import io.smallrye.reactive.messaging.annotations.OnOverflow;
-import io.smallrye.reactive.messaging.annotations.Stream;
 
 public final class DotNames {
 
     static final DotName VOID = DotName.createSimple(void.class.getName());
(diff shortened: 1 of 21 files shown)


cescoffier:
@kenfinnigan @geoand @mkouba There is something where I'm not quite sure: the metric support.
SR Reactive Messaging defines a metric (defined in the spec) : https://github.com/smallrye/smallrye-reactive-messaging/blob/master/smallrye-reactive-messaging-provider/src/main/java/io/smallrye/reactive/messaging/metrics/MetricDecorator.java (1 metric per channel).

In the PR, I just add this bean if the metric capability is enabled. Is there anything else required?

geoand:
I don't really know about metrics, so either the aforementioned people can help out or @jmartisk can chime in

extensions/smallrye-reactive-messaging/deployment/src/main/java/io/quarkus/smallrye/reactivemessaging/deployment/DotNames.java
```diff
@@ -10,17 +10,19 @@
...
 
     static final DotName VOID = DotName.createSimple(void.class.getName());
     static final DotName INCOMING = DotName.createSimple(Incoming.class.getName());
     static final DotName OUTGOING = DotName.createSimple(Outgoing.class.getName());
-    static final DotName CHANNEL = DotName.createSimple(Stream.class.getName());
-    static final DotName STREAM = DotName.createSimple(Channel.class.getName());
-    static final DotName EMITTER = DotName.createSimple(Emitter.class.getName());
-    static final DotName ON_OVERFLOW = DotName.createSimple(OnOverflow.class.getName());
+    static final DotName CHANNEL = DotName.createSimple(org.eclipse.microprofile.reactive.messaging.Channel.class.getName());
+    static final DotName OLD_CHANNEL = DotName.createSimple(Channel.class.getName());
```
geoand: Minor complaint: Maybe all this `old` stuff should be renamed to `deprecated`?

geoand:
I like it, :+1:

jmartisk:
IIUC the `MetricDecorator` puts the metrics into the `application` metric registry, I think `vendor` registry would be more correct for this, as these are not declared directly by an application. That can be achieved by injecting it together with `@org.eclipse.microprofile.metrics.annotation.RegistryType(type=VENDOR)`

Also we usually introduce a config property like `quarkus.EXTENSION.metrics.enabled` to enable metrics for a particular extension and the default value is false, see https://github.com/quarkusio/quarkus/blob/master/extensions/agroal/runtime/src/main/java/io/quarkus/agroal/runtime/AgroalBuildTimeConfig.java#L40 so that metrics for each extension have to be enabled separately

jmartisk:
Also if the set of channels (and therefore metrics) is known already at build time, the processor could optionally emit `MetricBuildItem`-s to get the metrics registered eagerly during build

extensions/smallrye-reactive-messaging/deployment/src/main/java/io/quarkus/smallrye/reactivemessaging/deployment/DotNames.java
cescoffier: I was wondering, "legacy" vs. "old" vs. "deprecated". I'm good with all of them. Any opinion?

cescoffier:
@jmartisk Thanks! 

So, unfortunately right now the set of channels is not known at build time. But I really want to go in this direction (it requires a bit more code and configuration analysis, nothing very hard).

I will add the flag to enable and disable the metric, that makes sense! So, we expose them only if 1. the metric capability is enabled 2. the extension metrics are enabled. Am I right?

extensions/smallrye-reactive-messaging/deployment/src/main/java/io/quarkus/smallrye/reactivemessaging/deployment/DotNames.java
geoand: I like `legacy`! I didn't think of that one, excellent!

jmartisk:
> So, we expose them only if 1. the metric capability is enabled 2. the extension metrics are enabled. Am I right?

Yes.

cescoffier:
@jmartisk About _vendor_, do you think the spec should mention it (See https://github.com/eclipse/microprofile-reactive-messaging/blob/22b40135a6dbffca17599ac79b89bd19174f2f98/spec/src/main/asciidoc/architecture.asciidoc#metrics). 

Right now the TCK uses the "default" registry:
https://github.com/eclipse/microprofile-reactive-messaging/blob/22b40135a6dbffca17599ac79b89bd19174f2f98/tck/src/main/java/org/eclipse/microprofile/reactive/messaging/tck/metrics/MetricsTest.java#L69

jmartisk:
Hm, if the TCKs already expect them in default (which is `application`), then I suppose we'll have to leave them there..  This has unfortunately only very recently been clarified in the MP Metrics spec (https://github.com/eclipse/microprofile-metrics/blob/master/spec/src/main/asciidoc/architecture.adoc#required-base-metrics) - this will be in the upcoming 2.3 version:

> The base scope is used for, and only for, any metrics that are defined in MicroProfile specifications. Metrics in the base scope are intended to be portable between different MicroProfile-compatible runtimes.

So I would suggest the next version of RM spec should move them to `base`

cescoffier:
@jmartisk Actually, the spec has not been released with these changes yet... So, there is still time to fix it :-). Let me create an issue in the spec (will ping you from there).

jmartisk:
@cescoffier Oh, that's good news then :) sure.

geoand:
The TCK failures seem related

kenfinnigan:
For the TCKs, I think https://github.com/quarkusio/quarkus/blob/master/tcks/microprofile-reactive-messaging/pom.xml#L36 needs to add the Metrics extension, as by default it's not brought in through a required dependency

cescoffier:
The TCK failure is expected, the failing tests have been removed from the TCK, it was not correct.

cescoffier:
I've applied a workaround for the TCK.

extensions/smallrye-reactive-messaging/deployment/src/main/java/io/quarkus/smallrye/reactivemessaging/deployment/DotNames.java
cescoffier: Changed!

cescoffier:
About the metrics change, let's wait until it's part of the upstream specification. It is going to happen soon I believe.

extensions/smallrye-reactive-messaging/deployment/src/main/java/io/quarkus/smallrye/reactivemessaging/deployment/DotNames.java
geoand: Lovely! Now that the naming has been addressed, the hard part is over :laughing:

geoand (approved):
LGTM but let's get an approval from @kenfinnigan and @jmartisk as well :)

extensions/smallrye-reactive-messaging/deployment/src/main/java/io/quarkus/smallrye/reactivemessaging/deployment/SmallRyeReactiveMessagingProcessor.java
```diff
@@ -143,54 +149,98 @@ void validateBeanDeployment(
...
-                        new BeanClassAnnotationExclusion(io.quarkus.smallrye.reactivemessaging.deployment.DotNames.OUTGOING)));
+                        new BeanClassAnnotationExclusion(
+                                io.quarkus.smallrye.reactivemessaging.deployment.DotNames.OUTGOING)));
+    }
+
+    @BuildStep
+    public void enableMetrics(Capabilities capabilities, ReactiveMessagingConfiguration configuration,
+            BuildProducer<AdditionalBeanBuildItem> beans) {
+        if (capabilities.isCapabilityPresent(Capabilities.METRICS) && configuration.metricsEnabled) {
+            beans.produce(new AdditionalBeanBuildItem(MetricDecorator.class.getName()));
```
jmartisk: I'm trying it and it looks like the `MetricDecorator` is included always (!), even without this line for some reason. And if the Metrics dependency is not present on the application, then it fails to boot due to a `ClassNotFoundException: org.eclipse.microprofile.metrics.MetricRegistry` because the `MetricDecorator` references that class. We need to find out why this bean is being included even when we don't want it.

jmartisk: Hm, I think it's because SR-RM provider jar contains a `beans.xml` so all its beans are automatically picked up: 
https://github.com/smallrye/smallrye-reactive-messaging/blob/master/smallrye-reactive-messaging-provider/src/main/resources/META-INF/beans.xml
Not sure if we can specifically exclude `MetricDecorator` without having to get rid of that `beans.xml`

cescoffier: It contains a bean.xml as it was requested. But yes, I can see the issue... Not sure how I can work around this issue. Dropping the beans.xml in reactive messaging would be seen as a breaking change.

kenfinnigan: What's the purpose of it being there? What's the use case it was needed for?

jmartisk: We could potentially "remove" the `@ApplicationScoped` annotation from that bean by an annotation transformer, but that's ugly. Or add a build-time mechanism for Arc for excluding a particular bean, if there's no such thing yet. Also quite ugly.
I feel we should get rid of the `beans.xml` if possible. If it's needed during testing for example, it can be added using ShrinkWrap. Removing it could be a breaking change from the PoV of the runtime that imports SR-RM, but if implemented properly it shouldn't be of any concern to user applications, I guess

cescoffier: Removing the `beans.xml` is a dead end. It breaks too many things (such as the Java SE support). 

One approach would be to move the metric decorator to its own module - without a beans.xml, and add the bean only in Quarkus. That potentially means more work to pass the TCK (mandating the metric support) and providing an implementation that would only be compliant in Quarkus. So not sure about this approach either.

jmartisk: One more option would be to make the decorator a no-op when the metrics extension is not available. To not trigger a `ClassNotFoundException` we would have to get rid of the `MetricRegistry` injection in it, and instead retrieve the registry by using static methods of the SmallRye-specific `MetricRegistries` class.
Not that it's nice, but it's possible.

cescoffier: That would require a change in SR Reactive Messaging that would use an SR specific API? Unfortunately, it's not possible as this implementation is used in other environments (that do not use SR Metrics)

cescoffier: I discussed with @mkouba and it seems that using a transformer to veto the bean is the best approach.

jmartisk: Hm, yes that would be using SmallRye-specific API.
Vetoing sounds like a good idea!

cescoffier: I've implemented the veto approach.

jmartisk (approved):

kenfinnigan (approved):

PR #6728, opened 2020-01-22
Update wildfly-elytron version to 2.0.0.Alpha6

sberyozkin:
@rsvoboda FYI, thanks for letting me know about this version. Will update accordingly should the next/final version become available in time for 1.3.0

PR #7060, opened 2020-02-07
Move the WildFly Elytron version to 1.11.2.Final

darranl:
I have been cleaning up our master branches in preparation for creating a .Final release of WildFly Elytron for Quarkus, however after reviewing the end results of the branches there are no changes required by Quarkus in the master branch that are not also in our existing 1.x release streams so for now I think it makes more sense to continue to use the releases we are already producing off 1.x.

If at any point changes are required which can only be applied to the master branch of Elytron there is no problem at all creating .Final releases from that branch but that doesn't seem required today.

### DIFF
--- a/bom/runtime/pom.xml
+++ b/bom/runtime/pom.xml
@@ -97,7 +97,7 @@
         <slf4j-jboss-logging.version>1.2.0.Final</slf4j-jboss-logging.version>
         <wildfly-common.version>1.5.3.Final-format-001</wildfly-common.version>
         <wildfly-client-config.version>1.0.1.Final</wildfly-client-config.version>
-        <wildfly-elytron.version>2.0.0.Alpha6</wildfly-elytron.version>
+        <wildfly-elytron.version>1.11.2.Final</wildfly-elytron.version>
         <jboss-modules.version>1.8.7.Final</jboss-modules.version>
         <jboss-threads.version>3.0.1.Final</jboss-threads.version>
         <vertx.version>3.8.5</vertx.version>


sberyozkin:
@darranl Thanks, so both versions offer the equivalent set of features from the Quarkus point of view. @rsvoboda Do you prefer to come down to `1.11.2.Final` from `2.0.0.Alpha6` ? The dependencies pulled by either of them are likely different but whether it might have some side-effects for the existing users I'm not sure.
@darranl By the way, [these are the changes](https://github.com/quarkusio/quarkus/pull/6728/files) @gsmet did to help me update to `2.0.0.Alpha6`. There were not required for `2.0.0.Alpha4`, so I'm not sure how those changes do in the native mode with `1.11.2.Final`

darranl:
I can confirm there are no differences in the dependencies in the master branch, we keep our 1.x branch up to date with the latest versions as they become available and the 1.x branch is routinely merged to the master branch so at this point in time both refer to the exact same set of external dependencies.

darranl:
@sberyozkin regarding the changes you linked for Alpha6 there is no difference between 1.x and master in relation to those password classes.

However if future issues are encountered please do feel free to contact us, we have no issue with adapting the Elytron project where needed and in some cases making changes to Elytron may be better than needing a Quarkus workaround.

sberyozkin:
@darranl OK, thanks for all the information. Lets wait till the native build passes. Cheers

sberyozkin:
@stuartwdouglas @gsmet Hi, I'm going to go ahead with the merge once this PR passes, let me know please if you have some concerns, I don't see a problem for Quarkus with Darran himself having checked it all at the Elytron side.

gsmet:
Can't say I'm overly excited about this. We really need to be very careful and please don't merge it before @stuartwdouglas had the time to check this out.

darranl:
@gsmet which bit is concerning you?

sberyozkin:
@darranl From my point of view the actual version downgrade is unusual (and I'm generally not excited about certain requirements like having `-Final` in the version for example :-)) but if this change will make things simpler for different processes and does not change anything then I guess it is fine.
@gsmet Sure, will wait for Stuart.
Cheers

darranl:
@sberyozkin this sounds more like a psychological issue

gsmet:
@darranl just switching to a different version now. I just want Stuart to have a say and be sure we didn't have any magic hidden in those versions. Maybe @dmlloyd too.

dmlloyd (approved):
+1 from me

darranl:
FYI here is the comparison between the proposed tag and the current state of master: -

https://github.com/wildfly-security/wildfly-elytron/compare/1.11.2.Final...master

Presently the only difference between those branches is the deletion of some deprecated code not in use within Quarkus and the change in version number.

stuartwdouglas (approved):

PR #7064, opened 2020-02-07
Dockerfile.native - run with non-root user 1001

haraldatbmw:
The native docker-image should run with non-root user 1001.


gsmet (approved):
Thanks!

Makes sense to me but I prefer we have @cescoffier 's blessing before merging.

gsmet:
@cescoffier can you check that one?

cescoffier (approved):

cescoffier:
Thanks! That was a bug actually (never run as root)

PR #7069, opened 2020-02-07
Updated MP Config to 1.4 and SmallRye Config to 1.6.1

radcortez:


gastaldi (approved):

gastaldi:
@radcortez minor thing, but can we use `SmallRyeConfigBuilder.withSources(Collection)` in https://github.com/quarkusio/quarkus/blob/master/core/runtime/src/main/java/io/quarkus/runtime/configuration/ConfigUtils.java#L84?

radcortez:
With this new version, yes. I see you added that method there :) It is included.

Do you want me to do the change, or we do it in a separate PR?

gastaldi:
If you could do the change in this PR that would be great (it can be in a separate commit), otherwise I'll send a PR after this one is merged

radcortez:
I'll do it. No worries.

We actually found a regression in 1.6.0, so we may need to hold this a little bit for a new minor version.

gastaldi:
Nice, can you squash the first commit with the second one so you have only two commits in this PR?

radcortez:
Sure. Sec.

radcortez:
Done.

PR #7071, opened 2020-02-07
QuarkusDev should be serializing the AppModel in the same manner as QuarkusTestConfig

aloubyansky:
i.e. in a temp file and passing its path using quarkus-internal.serialized-app-model.path

It's tricky to write a test specifically for this issue since all the required deps are going to be built and added to the local maven repo and eventually QuarkusBootstrap will pick them up.


geoand (approved):
Works great!

PR #7073, opened 2020-02-07
Handle MongoDB BsonDiscriminator

### LINKED ISSUE #6292: Improve the usage of MongoDB @BsonDiscriminator
**Description**
The `@BsonDiscriminator` annotation can be use to persist different entity types inside a collection. It's very handy for persisting a hierarchy of entities inside the same collection when using abstract classes: http://mongodb.github.io/mongo-java-driver/3.12/bson/pojos/#properties-with-abstract-or-interface-types

To use it, all classes that have a `@BsonDiscriminator` annotation should be registered inside the `PojoCodec`. This is done automatically when you write an entity inside the collection, but if you read it first the codec didn't know yet wich class to use so it didn't work.

I found a workaround to force the registration of the discriminator in a `@PostConstruct` method but we should facilitate this by regitering these classes inside the deployement module of the `mongodb-client` extension.

For the record my current workaround is:

```
    @PostConstruct
    void init(){
        mongoCollection().getCodecRegistry().get(OrderTransactionRecord.class);
        mongoCollection().getCodecRegistry().get(SaleTransactionRecord.class);
        mongoCollection().getCodecRegistry().get(TicketTransactionRecord.class);
    }
```

**Implementation ideas**
Register all the classes that have a `@BsonDiscriminator` annotation inside the `mongodb-client` extension.
Document also the fact that we do this.

loicmathieu:
Fixes #6292

This PR :
- Configure the POJO codec used by MongoDB with Panache with all the classes annotated with `@BsonDiscriminator` so that a user don't need to do the hack described in #6292 
- Register for reflection all classes annotated by `@BsonDiscriminator`

I also add a section in the mongodb guide about the POJO codec.

Maybe I need to have a few lines about this inside the MongoDB with Panache guide (on the section where we describe the various annotations we can use with the POJO codec) ?

### DIFF
--- a/docs/src/main/asciidoc/mongodb.adoc
+++ b/docs/src/main/asciidoc/mongodb.adoc
@@ -510,6 +510,16 @@ public class CodecFruitService {
 }
 ----
 
+== The POJO Codec
+
+The link:http://mongodb.github.io/mongo-java-driver/3.12/bson/pojos[POJO Codec] provides a set of annotations that enable the customization of
+the way a POJO is mapped to a MongoDB collection and this codec is initialized automatically by Quarkus
+
+One of these annotations is the `@BsonDiscriminator` annotation that allows to storage multiple Java types in a single MongoDB collection by adding
+a discriminator field inside the document. It can be useful when working with abstract types or interfaces.
+
+Quarkus will automatically register all the classes annotated with `@BsonDiscriminator` with the POJO codec.
+
 == Simplifying MongoDB with Panache
 
 The link:mongodb-panache[MongoDB with Panache] extension facilitates the usage of MongoDB by providing active record style entities (and repositories) like you have in link:hibernate-orm-panache.html[Hibernate ORM with Panache] and focuses on making your entities trivial and fun to write in Quarkus.
--- a/extensions/mongodb-client/runtime/src/main/java/io/quarkus/mongodb/runtime/AbstractMongoClientProducer.java
+++ b/extensions/mongodb-client/runtime/src/main/java/io/quarkus/mongodb/runtime/AbstractMongoClientProducer.java
@@ -22,6 +22,7 @@ import javax.annotation.PreDestroy;
 import org.bson.codecs.configuration.CodecProvider;
 import org.bson.codecs.configuration.CodecRegistries;
 import org.bson.codecs.configuration.CodecRegistry;
+import org.bson.codecs.pojo.ClassModel;
 import org.bson.codecs.pojo.Conventions;
 import org.bson.codecs.pojo.PojoCodecProvider;
 import org.jboss.logging.Logger;
@@ -54,6 +55,7 @@ public abstract class AbstractMongoClientProducer {
     private MongodbConfig mongodbConfig;
     private boolean disableSslSupport = false;
     private List<String> codecProviders;
+    private List<String> bsonDiscriminators;
     private Map<String, MongoClient> mongoclients = new HashMap<>();
     private Map<String, ReactiveMongoClient> reactiveMongoClients = new HashMap<>();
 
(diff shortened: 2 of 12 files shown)


docs/src/main/asciidoc/mongodb.adoc
```diff
@@ -502,6 +502,18 @@ public class CodecFruitService {
 }
 ----
 
+== The POJO Codec
+
+Quarkus initializes the link:http://mongodb.github.io/mongo-java-driver/3.12/bson/pojos[MongoDB POJO Codec], this codec is capable of automatically working with POJOs so in many cases,
```
geoand: I don't think this sentence provides any useful information. I propose we drop it and add the link to the next sentence

docs/src/main/asciidoc/mongodb.adoc
```diff
@@ -502,6 +502,18 @@ public class CodecFruitService {
...
+you will be able to avoid the creation of a specific codec.
+
+The POJO Codec provides a rich set of features to customize the way your POJO is mapped to a collection using a set of annotations.
```
geoand:
```suggestion
The POJO Codec provides a set of annotations that enable the customization of the way a POJO is mapped to a MongoDB collection.
```

docs/src/main/asciidoc/mongodb.adoc
```diff
@@ -502,6 +502,18 @@ public class CodecFruitService {
...
+
+One of these annotations is the `@BsonDiscriminator` annotation that allows to store in a single collection multiple Java types by adding
```
geoand:
```suggestion
One of these annotations is the `@BsonDiscriminator` annotation that allows to storage multiple Java types in a single MongoDB collection by adding
```

docs/src/main/asciidoc/mongodb.adoc
```diff
@@ -502,6 +502,18 @@ public class CodecFruitService {
...
+a discriminator field inside the document. It can be useful when working with abstract types or interfaces.
+
+Quarkus will automatically register all the classes annotated with `@BsonDiscriminator` inside the POJO codec, so you don't have to do it by yourself.
```
geoand:
```suggestion
Quarkus will automatically register all the classes annotated with `@BsonDiscriminator` with the POJO codec.
```

extensions/mongodb-client/runtime/src/main/java/io/quarkus/mongodb/runtime/AbstractMongoClientProducer.java
```diff
@@ -217,11 +219,24 @@ private MongoClientSettings createMongoConfiguration(MongoClientConfig config) {
...
-        CodecProvider pojoCodecProvider = PojoCodecProvider.builder()
+        PojoCodecProvider.Builder pojoCodecProviderBuilder = PojoCodecProvider.builder()
                 .automatic(true)
-                .conventions(Conventions.DEFAULT_CONVENTIONS)
-                .build();
-        providers.add(pojoCodecProvider);
+                .conventions(Conventions.DEFAULT_CONVENTIONS);
+        // register bson discriminators
+        ClassLoader classLoader = Thread.currentThread().getContextClassLoader();
+        if (classLoader == null) {
```
geoand: I don't think this can happen really... And if it does it's doubtful that `AbstractMongoClientProducer.class.getClassLoader()` will save you (certainly not dev mode that is :))

geoand (commented):
Thanks for this!

I added some minor comments

docs/src/main/asciidoc/mongodb.adoc
loicmathieu: I think we need to tell the user that the POJO codec is initialized by Quarkus in some way, and in the previous section we talk about creating custom codec so I wanted to make it clear that this is not mandatory as in most cases the pojo codec will works OOTB.

geoand: Then perhaps the 2 sentences can be combined like so:

```
The POJO Codec provides a set of annotations that enable the customization of the way a POJO is mapped to a MongoDB collection and This codec is initialized automatically by Quarkus
```

extensions/mongodb-client/runtime/src/main/java/io/quarkus/mongodb/runtime/AbstractMongoClientProducer.java
loicmathieu: I copied this pattern from the Smallrye Fault Tolerance extension, and I just check almost all extensions directly used `Thread.currentThread().getContextClassLoader()` so it should be safe to use it.

geoand: Yeah, I can't see a scenario where `Thread.currentThread().getContextClassLoader()` will return null. In any case this is very minor so it's not worth putting too much thought into it

geoand:
@loicmathieu you know the drill of squashing the commits :laughing:

geoand:
Pushed the wrong button and closed by accident :(

geoand (approved):
Nice one, thanks a lot!

PR #7070, opened 2020-02-07
Move generation of GeneratedFileSystemResourceBuildItem out of jar

geoand:
This helps to decouple the build steps producing them from the actual jar
generation thus opening up more possibilities in the future

PR #7075, opened 2020-02-07
Add option to force a build step to run even if it doesn't produce ou…

geoand:
…tput


geoand:
This has been mentioned before, but I had the need for it because of #7070

geoand:
@dmlloyd I tried this and it seems to work but I am wondering if perhaps there could be some unintended side-effect that I haven't foreseen?

dmlloyd:
Technically there are no problems, but logically I think there are.  We never really want to "always" run a step.  If we think we do, we're almost certainly basing that thought on an unwarranted assumption of some sort.

We should be using empty build items as abstract goals, and producing those instead.

geoand:
I remember Bill Burke bringing it up lately when trying to do something (his use-case admittedly could didn' t absolutely require it).
But for example in #7070, I needed to use a dummy build item in order to actually write things as output

dmlloyd:
We should have a build item directly for things that are written as part of the build, which would avoid that trouble.  Failing that, an abstract goal that represents the files written should be used.  If a build requires such files, then the corresponding steps are run.  If not (e.g. what if we're generating docs or just doing an analysis stage of some sort?) then the steps are not run.

geoand:
Understood, I'll go ahead and close this then

PR #7086, opened 2020-02-07
Add a composite OidcAuthenticationMechanism

### LINKED ISSUE #6997: Improve OIDC multi-tenancy and "web-app" configuration support
**Description**
Right now the tenant-specifc `web-app` `application-type` will be ignored unless a default configuration is provided.
For example, configuring only:
```
quarkus.oidc.tenant-1.auth-server-url=${keycloak.url}/realms/quarkus
quarkus.oidc.tenant-1.application-type=web-app
quarkus.oidc.tenant-1.client-id=quarkus-app
quarkus.oidc.tenant-1.credentials.secret=secret
```
will log a warning that `quarkus.oidc.tenant-1.application-type` property is not supported.
The reason for that is that the `application-type` property is a build time property and as such it is not available at the moment the tenant configuration is resolved.  And because a `service` type is a default value, only the Bearer mechanism would be registered at the deployment time.

So for the code-flow to work with the multi-tenancy, one needs to have a default configuration even if it is not used:
```
# Default config just to make the Tenant specific configuration work in the code flow mode
quarkus.oidc.auth-server-url=${keycloak.url}/realms/quarkus

# Tenant 1 configuration
quarkus.oidc.tenant-1.auth-server-url=${keycloak.url}/realms/quarkus

# Tenant 2 configuration
quarkus.oidc.tenant-2.auth-server-url=${keycloak.url}/realms/quarkus

#etc
```

**Implementation ideas**
- Make `applicatiion-type` a runtime property
- Instead of registering either Bearer or CodeFlow authentication mechanism at the deployment time, register a composite `OidcAutenticationMechanism` which will have both of those authentication mechanisms available and delegate to the right one depending on whatever the resolved context's application type is.

This way we will also be able to support an application which can act both as a `service` and a `web-app` application.

CC @pedroigor @stuartwdouglas

### LINKED ISSUE #7013: Optimize the use of TenantConfigResolver by making OidcTenantConfig a good map key
**Description**
Right now using `TenantConfigProvider` (which can create `OidcTenantConfig` on the fly) involves reconnecting to the OIDC server on every request (note a combination of `TenantResolver` + having a tenant specific config in `application.properties` is very effective).

**Implementation ideas**
Implement `OidcTenantConfig` `equals` and `hashCode` and use it as a map key in the concurrent map to retrieve the already calculated `OAuth2Auth` objects. It will also make it possible to avoid blocking for `TenantConfigProvider` for all but the very 1st request

CC @pedroigor

sberyozkin:
Fixes #6997.
Fixes #7013.

### DIFF
--- a/extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/CodeAuthenticationMechanism.java
+++ b/extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/CodeAuthenticationMechanism.java
@@ -10,8 +10,6 @@ import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
 import java.util.function.Function;
 
-import javax.enterprise.context.ApplicationScoped;
-
 import org.jboss.logging.Logger;
 
 import io.netty.handler.codec.http.HttpResponseStatus;
@@ -33,7 +31,6 @@ import io.vertx.ext.auth.oauth2.AccessToken;
 import io.vertx.ext.web.RoutingContext;
 import io.vertx.ext.web.impl.CookieImpl;
 
-@ApplicationScoped
 public class CodeAuthenticationMechanism extends AbstractOidcAuthenticationMechanism {
 
     private static final Logger LOG = Logger.getLogger(CodeAuthenticationMechanism.class);
@@ -63,9 +60,9 @@ public class CodeAuthenticationMechanism extends AbstractOidcAuthenticationMecha
                 .build();
     }
 
-    @Override
     public CompletionStage<SecurityIdentity> authenticate(RoutingContext context,
-            IdentityProviderManager identityProviderManager) {
+            IdentityProviderManager identityProviderManager,
+            DefaultTenantConfigResolver resolver) {
         Cookie sessionCookie = context.request().getCookie(SESSION_COOKIE_NAME);
 
         // if session already established, try to re-authenticate
@@ -82,17 +79,16 @@ public class CodeAuthenticationMechanism extends AbstractOidcAuthenticationMecha
         }
 
         // start a new session by starting the code flow dance
-        return performCodeFlow(identityProviderManager, context);
+        return performCodeFlow(identityProviderManager, context, resolver);
     }
 
-    @Override
-    public CompletionStage<ChallengeData> getChallenge(RoutingContext context) {
+    public CompletionStage<ChallengeData> getChallenge(RoutingContext context, DefaultTenantConfigResolver resolver) {
         removeCookie(context, SESSION_COOKIE_NAME);
+
+        TenantConfigContext configContext = resolver.resolve(context, false);
 
         ChallengeData challenge;
         JsonObject params = new JsonObject();
-
-        TenantConfigContext configContext = tenantConfigResolver.resolve(context);
 
         // scope
         List<Object> scopes = new ArrayList<>();
@@ -102,13 +98,13 @@ public class CodeAuthenticationMechanism extends AbstractOidcAuthenticationMecha
 
         // redirect_uri
         URI absoluteUri = URI.create(context.request().absoluteURI());
-        String redirectPath = getRedirectPath(context, absoluteUri);
+        String redirectPath = getRedirectPath(configContext, absoluteUri);
         String redirectUriParam = buildRedirectUri(context, absoluteUri, redirectPath);
         LOG.debugf("Authentication request redirect_uri parameter: %s", redirectUriParam);
         params.put("redirect_uri", redirectUriParam);
 
         // state
-        params.put("state", generateState(context, absoluteUri, redirectPath));
+        params.put("state", generateState(context, configContext, absoluteUri, redirectPath));
 
         // extra redirect parameters, see https://openid.net/specs/openid-connect-core-1_0.html#AuthRequests
         if (configContext.oidcConfig.authentication.getExtraParams() != null) {
@@ -124,7 +120,9 @@ public class CodeAuthenticationMechanism extends AbstractOidcAuthenticationMecha
     }
 
     private CompletionStage<SecurityIdentity> performCodeFlow(IdentityProviderManager identityProviderManager,
-            RoutingContext context) {
+            RoutingContext context, DefaultTenantConfigResolver resolver) {
+        TenantConfigContext configContext = resolver.resolve(context, true);
+
         JsonObject params = new JsonObject();
 
         String code = context.request().getParam("code");
@@ -177,12 +175,12 @@ public class CodeAuthenticationMechanism extends AbstractOidcAuthenticationMecha
         }
 
         params.put("code", code);
-        String redirectPath = getRedirectPath(context, absoluteUri);
+        String redirectPath = getRedirectPath(configContext, absoluteUri);
         String redirectUriParam = buildRedirectUri(context, absoluteUri, redirectPath);
         LOG.debugf("Token request redirect_uri parameter: %s", redirectUriParam);
         params.put("redirect_uri", redirectUriParam);
 
-        tenantConfigResolver.resolve(context).auth.authenticate(params, userAsyncResult -> {
+        configContext.auth.authenticate(params, userAsyncResult -> {
             if (userAsyncResult.failed()) {
                 cf.completeExceptionally(new AuthenticationFailedException());
             } else {
@@ -221,16 +219,17 @@ public class CodeAuthenticationMechanism extends AbstractOidcAuthenticationMecha
                 result.opaqueRefreshToken(), context));
     }
 
-    private String getRedirectPath(RoutingContext context, URI absoluteUri) {
-        Authentication auth = tenantConfigResolver.resolve(context).oidcConfig.getAuthentication();
+    private String getRedirectPath(TenantConfigContext configContext, URI absoluteUri) {
+        Authentication auth = configContext.oidcConfig.getAuthentication();
         return auth.getRedirectPath().isPresent() ? auth.getRedirectPath().get() : absoluteUri.getRawPath();
     }
 
-    private String generateState(RoutingContext context, URI absoluteUri, String redirectPath) {
+    private String generateState(RoutingContext context, TenantConfigContext configContext, URI absoluteUri,
+            String redirectPath) {
         String uuid = UUID.randomUUID().toString();
         String cookieValue = uuid;
 
-        Authentication auth = tenantConfigResolver.resolve(context).oidcConfig.getAuthentication();
+        Authentication auth = configContext.oidcConfig.getAuthentication();
         if (auth.isRestorePathAfterRedirect() && !redirectPath.equals(absoluteUri.getRawPath())) {
             cookieValue += (COOKIE_DELIM + absoluteUri.getRawPath());
         }
--- a/extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/DefaultTenantConfigResolver.java
+++ b/extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/DefaultTenantConfigResolver.java
@@ -3,6 +3,7 @@ package io.quarkus.oidc.runtime;
 import java.util.Map;
 import java.util.function.Function;
 
+import javax.annotation.PostConstruct;
 import javax.enterprise.context.ApplicationScoped;
 import javax.enterprise.inject.Instance;
 import javax.inject.Inject;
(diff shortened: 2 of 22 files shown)


extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/OidcAuthenticationMechanism.java
```diff
@@ -0,0 +1,43 @@
...
+
+    @Inject
+    DefaultTenantConfigResolver tenantConfigResolver;
+    private BearerAuthenticationMechanism bearerAuth = new BearerAuthenticationMechanism();
+    private CodeAuthenticationMechanism codeAuth = new CodeAuthenticationMechanism();
+
+    @Override
+    public CompletionStage<SecurityIdentity> authenticate(RoutingContext context,
+            IdentityProviderManager identityProviderManager) {
+        TenantConfigContext configContext = tenantConfigResolver.resolve(context);
```
sberyozkin: @pedroigor, @stuartwdouglas  Hi, this is a work in progress, I have one more test in the works which combines the service and web-app tenants. Happy enough with how it is working out, the only problem is resolving the configContext here deadlocks in `oidc-tenancy` tests :-) in the sync lock `DefaultTenantConfigResolver`.
Pretty sure it can be reproduced if for example `BearerAuthenticationMechanism.authenticate` on the master  would try to resolve early in its `authenticate`. Nearly there though...Thanks

sberyozkin: I've been wondering today how I'm going to resolve this deadlock  and realized what is really needed here is a resolution without attempting to create `OAuth2Auth`, at this point only a check what kind of application it is is needed. `DefaultTenantConfigResolver.resolve` needs a `boolean create` parameter, same one which is used by `isBlocking` check internally :-)

stuartwdouglas:
This needs to be formatted.

extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/OidcAuthenticationMechanism.java
```diff
@@ -0,0 +1,39 @@
...
+import io.quarkus.vertx.http.runtime.security.ChallengeData;
+import io.quarkus.vertx.http.runtime.security.HttpAuthenticationMechanism;
+import io.vertx.ext.web.RoutingContext;
+
+@ApplicationScoped
+public class OidcAuthenticationMechanism implements HttpAuthenticationMechanism {
+
+    @Inject
+    DefaultTenantConfigResolver resolver;
+    private BearerAuthenticationMechanism bearerAuth = new BearerAuthenticationMechanism();
```
pedroigor: It would be nice if we could create the mechanisms only when they are necessary. Also, I think most deployments will just rely on a single mechanism as they will not be using multi-tenancy at all.

So, I'm wondering if we could:

* Decide for a specific mechanism at build time depending on the existence of tenants
* Maybe too much, but if in a multi-tenancy deployment, create the mechanism only if necessary depending on the tenant type

sberyozkin: @pedroigor I think it would complicate things unnecessary for no obvious benefits. And with `TenantConfigResolver` it is not possible anyway as it offers the dynamic option. 
This PR offers a very simple fix for #6997 - please have a look at it, right now the multi-tenancy is not properly supported for `web-app` anyway. As a direct consequence of this fix we will offer an option to the sophisticated users go beyond the `frontend-only` or `backend-only` services which is already being thought after in on external project, please see  #6997.

>  I think most deployments will just rely on a single mechanism as they will not be using multi-tenancy at all

Those deployments will only then have a `cost` of an extra single instance (`Bearer` or `CodeAuth` mechanism) - which is really small issue compared to the enabling hybrid web-apps.

extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/OidcAuthenticationMechanism.java
pedroigor: OK. So only tests are missing. Otherwise, LGTM. Thanks.

extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/OidcAuthenticationMechanism.java
pedroigor: OK. So only tests are missing. Otherwise, LGTM. Thanks.

sberyozkin: @pedroigor sorry may be I've just totally missed some of your points, which would not be surprising :-). I mean, we can create the mechanisms dynamically, with the volatile storage, so for a single type applications only one will ever be created, may be few temp instances depending on how heavy the load is...

sberyozkin: @pedroigor By the way, you are right of course that most of the applications will no go the multi-tenancy route - but our OIDC multi-tenancy feature will be a real differentiator (thanks for wiring it so well :-) ) when people will need something more involved and it was a very thought after feature for a number of users (who watched the original issue)

sberyozkin: @pedroigor Also, apologies for being defensive here, I'm not concerned about my PR code of course:-) but I'm very liking the idea of a flexible per-tenant specific OIDC configuration and believe it will be very good for Quarkus in the real world :-). 
> OK. So only tests are missing. Otherwise, LGTM. Thanks.
Thanks, right, working on an extra test now...

sberyozkin:
@pedroigor @stuartwdouglas thanks for the comments so far, Pedro, thanks for helping out with getting the roles into the id token as well :-)
It is is probably ready to go, I haven't done yet
```
volatile BearerAuthenticationMechanism bearerAuth;
volatile CodeAuthenticationMechanism codeAuth;
```
and then create them on request, but I'm not 100% sure it will be cheaper. Probably yes as in this case GC will collect few extra instances of a single type (which might not be even necessary)...I'll do it a little bit later...

sberyozkin:
Added the volatile qualifiers, that should be more optimal for the mainstream single (default) tenant case

sberyozkin:
@pedroigor @stuartwdouglas I'vw improved the test, `application.properties` now have a per-tenant `application-type` property set without relying on a default (which is `service`) and also updated `TenantResource` to check few things around the tokens and claim names to make sure things are not mixed in...

extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/OidcAuthenticationMechanism.java
```diff
@@ -0,0 +1,53 @@
...
+import io.quarkus.vertx.http.runtime.security.ChallengeData;
+import io.quarkus.vertx.http.runtime.security.HttpAuthenticationMechanism;
+import io.vertx.ext.web.RoutingContext;
+
+@ApplicationScoped
+public class OidcAuthenticationMechanism implements HttpAuthenticationMechanism {
+
+    @Inject
+    DefaultTenantConfigResolver resolver;
+    private volatile BearerAuthenticationMechanism bearerAuth;
```
stuartwdouglas: Why are you doing this lazy creation? These objects don't have any state, there is no measurable benefit to creating them lazily, but in return you add a volatile reads to every request which is more expensive.

extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/OidcAuthenticationMechanism.java
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    @Override
+    public CompletionStage<ChallengeData> getChallenge(RoutingContext context) {
+        return isWebApp(context) ? getCodeAuth().getChallenge(context, resolver)
+                : getBearerAuth().getChallenge(context, resolver);
+    }
+
+    private CodeAuthenticationMechanism getCodeAuth() {
+        if (codeAuth == null) {
```
stuartwdouglas: Not that it matters, but this can actually create multiple instances here if multiple threads hit it at the same time.

extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/CodeAuthenticationMechanism.java
```diff
@@ -124,7 +120,9 @@ private static QuarkusSecurityIdentity augmentIdentity(SecurityIdentity security
     }
 
     private CompletionStage<SecurityIdentity> performCodeFlow(IdentityProviderManager identityProviderManager,
-            RoutingContext context) {
+            RoutingContext context, DefaultTenantConfigResolver resolver) {
+        TenantConfigContext configContext = resolver.resolve(context, true);
```
stuartwdouglas: It looks like resolve is now called lots of times per request. I think it might make sense to add some code to DefaultTenantResolver that stores the result in the routing context so subsequent invocations are effectively cached.

extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/DefaultTenantConfigResolver.java
```diff
@@ -26,12 +26,21 @@
...
+     * the active OIDC connection instance which may be null.
+     * 
+     * @param context the current request context
+     * @param create if true then the OIDC connection must be available or established
+     *        for the resolution to be successful
+     * @return
+     */
+    TenantConfigContext resolve(RoutingContext context, boolean create) {
         if (tenantConfigResolver.isAmbiguous()) {
             throw new IllegalStateException("Multiple " + TenantConfigResolver.class + " beans registered");
```
stuartwdouglas: I know this is not part of the PR, but this code should be in a PostCostruct method

integration-tests/oidc-tenancy/pom.xml
```diff
@@ -16,6 +16,7 @@
 
     <properties>
         <keycloak.url>http://localhost:8180/auth</keycloak.url>
+        <htmlunit.version>2.36.0</htmlunit.version>
```
stuartwdouglas: This should be managed in the build parent.

stuartwdouglas (changes requested):

extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/CodeAuthenticationMechanism.java
sberyozkin: It is only a visual impression :-). This PR only introduces one extra resolver call in `OidcAuthenticationMechanism` to figure which way to go. It is called a few times in `CodeAuthenticationMechanism` right on the master requesting a complete resolution. This PR optimizes it a bit, asking for `OidcTenantConfig` only when no `OAuth2Auth` is needed.
And @pedroigor has already updated `DefaultTenantConfigResolver` to support caching `TenantConfigContext` when needed. So it should be fine

extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/OidcAuthenticationMechanism.java
sberyozkin: Yes, this is what I was typing about [above](https://github.com/quarkusio/quarkus/pull/7086#issuecomment-584300029) but I decided to give it a try to make sure for a single `application-type` no code (or bearer) flow is created given Pedro's concern.
I'm going to return to the original code where I was just creating these 2 instances.
With the dynamic `TenantConfigResolver` one can never be 100% sure there will only be one application-type

extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/OidcAuthenticationMechanism.java
sberyozkin: Pedro was concerned about the eager creation in the constructor, so I decided to give a try, though was not entirely sure. I'll just revert. Please also see my response to your next comment.

stuartwdouglas: If you are really concerned about the object creation all that logic could just be moved to static methods as there is no state in the objects, and they no longer have to implement an interface.

sberyozkin:
@stuartwdouglas Thanks for finding the time to review, will work on the comments tomorrow, have to sign off
@pedroigor Hi Pedro, hope you are happy with the actual test :-)

extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/OidcAuthenticationMechanism.java
sberyozkin: @stuartwdouglas thanks, not sure I understand :-) I guess both objects can be just created as before. There is only one extra single object in the mainstream single `application-type` case which as you said has no state and it is not possible to avoid this extra object's class be added to the native image, so it is a very tiny cost indeed.
But can you explain please what you meant about moving the logic to the static methods, what difference would it make compared to the current lazy initialization done in the instance ? `volatile` class fields would still be read...But I guess it is not what you had in mind :-)

pedroigor:
@sberyozkin Tests seems fine.

sberyozkin:
Hi Pedro, @pedroigor, Stuart, @stuartwdouglas, I'm not really happy with the PR, especially after Pedro has proposed how to resolve TenantConfigResolver issue to do with the tenant ids (Thanks Pedro). Stuart, addressed most of you comments too. 
It should be good to go now IMHO; will deal with the comments as well
Cheers

stuartwdouglas (approved):

PR #7093, opened 2020-02-08
Add banner extension module - default banner & option for custom banner for user

### LINKED ISSUE #5267: Offer a Quarkus asciiart banner

RotBolt:
fixes #5267


KaenDagger:
@geoand finally passed :sob:

machi1990 (approved):
LGTM.

geoand:
Let me it give it a last super quick look in a few hours

core/deployment/src/main/java/io/quarkus/deployment/logging/LoggingResourceProcessor.java
```diff
@@ -79,11 +81,22 @@ void miscSetup(
     @BuildStep
```
geoand: This needs to be:

```
@BuildStep(loadsApplicationClasses = true)
```

Otherwise if you try to set a custom banner (that of course needs to be placed in `src/main/resources`), the banner not found and therefore not displayed at runtime.

extensions/banner/deployment/src/main/java/io/quarkus/banner/deployment/BannerConfig.java
```diff
@@ -0,0 +1,13 @@
+package io.quarkus.banner.deployment;
+
+import io.quarkus.runtime.annotations.ConfigItem;
+import io.quarkus.runtime.annotations.ConfigRoot;
+
+@ConfigRoot(name = "banner")
+public class BannerConfig {
+    /**
+     * The path of the banner which could be provided by user
```
geoand: It should be mentioned mentioned that this path is a path relative to the root of the classpath, not a file system path

geoand (changes requested):
Let's get these minor things fixed and then we can merge

core/deployment/src/main/java/io/quarkus/deployment/logging/LoggingResourceProcessor.java
KaenDagger: It was loading the custom banner, placed inside resources ,

extensions/banner/deployment/src/main/java/io/quarkus/banner/deployment/BannerProcessor.java
```diff
@@ -0,0 +1,52 @@
...
+import io.quarkus.deployment.annotations.Record;
+import io.quarkus.deployment.builditem.ConsoleFormatterBannerBuildItem;
+import io.quarkus.deployment.builditem.HotDeploymentWatchedFileBuildItem;
+import io.quarkus.deployment.util.FileUtil;
+
+public class BannerProcessor {
+
+    private static final Logger logger = Logger.getLogger(BannerProcessor.class);
+
+    @BuildStep(loadsApplicationClasses = true)
```
geoand: This is correct, :+1:

core/deployment/src/main/java/io/quarkus/deployment/logging/LoggingResourceProcessor.java
```diff
@@ -76,14 +78,25 @@ void miscSetup(
                 new ServiceProviderBuildItem(EmbeddedConfigurator.class.getName(), InitialConfigurator.class.getName()));
     }
 
-    @BuildStep
+    @BuildStep(loadsApplicationClasses = true)
```
geoand: Did I mention that this needs to change? If I did, sorry, I was wrong. This should not be needed, only the other `loadsApplicationClasses = true` you added

geoand:
I added a comment. I most likely made the mistake previously, sorry

core/deployment/src/main/java/io/quarkus/deployment/logging/LoggingResourceProcessor.java
KaenDagger: But it was not loading resource. When I changed this too. It did

KaenDagger: Oh wait , did you mean , `loadApplicationClassTrue` in banner extension and not on LoggingResourceProcessor

KaenDagger: @geoand  I checked , I have to add `loadApplicationClasses= true` to both places. Otherwise It is not working
Otherwise it get stucks at this line 
```
[INFO] --- quarkus-maven-plugin:999-SNAPSHOT:dev (default-cli) @ banner-test ---
Listening for transport dt_socket at address: 5005
```

No output further

geoand: I'll check again and let you know, but I really don't think it's necessary

geoand: I double-checked a test application and indeed this is not needed. 

The reason why it's needed in the other build step you added is because you are attempting you read a classpath resource inside a build step - a resource that was added by the user. The default Thread Context ClassLoader that is used when build steps are run (which Quarkus calls the `Augmentation ClassLoader`) does not have access to the application classes and resources.
When using `loadApplicationClasses` then the Thread Context ClassLoader that is set when the build step is run (is what Quarkus calls the `Deployment ClassLoader`) and this does have access to application classes and resources.

The logging processor you have changed here has nothing to do with loading the resource.

KaenDagger: I removed it from LoggingResourceProcessor , but then it doesn't work . As I output stuck

geoand: You must be doing something wrong. I tried it multiple times

KaenDagger: You mean on your side banner is working ?

geoand: Yes, when I use `@BuildStep(loadsApplicationClasses = true)` in `BannerProcessor` only and leave this one without it (like it was before the change), then a custom banner file placed inside `src/main/resources` and configured in `application.properties`, works just fine

KaenDagger: Ok its working now.
I probably erased `@BuildStep` annotation itself instead of removing `loadApplicationClasses = true` only from LoggingResourceProcessor. :sweat:

geoand (approved):

core/deployment/src/main/java/io/quarkus/deployment/logging/LoggingResourceProcessor.java
geoand: Great!

KaenDagger:
@geoand , could you merge it. If all the things are satisfied

geoand:
Thanks for this!

machi1990:
Good work @KaenDagger.

Issue #7018, opened 2020-02-06
Recreate all metric registries on startup?


jmartisk:
Currently, the behavior of dev mode is that `application` metric registry will be discarded and re-created on reload. This however does not include the `vendor` and `base` registries, so extensions must take care of removing their metrics properly in case that they have to be re-registered on reloading. 
While this could theoretically allow some specific metrics, in cases where appropriate, to just 'survive' a reload and continue without change, we're not doing this for any metrics at the moment, and there's actually not many cases where this would work. I can only think of some JVM metrics in the `base` scope. Most other metrics can't do this, especially if they need any sort of reference to application code, because application will be loaded by a different class loader.

That said, my opinion is that allowing this optimization might be not worth it, because it forces extensions to include cleanup code (only those extensions that are registering metrics directly rather than using `MetricBuildItem`, which takes care of the cleanup automatically).
If we changed the behavior such that all registries will get re-created and all metrics will have to be re-registered, we can do a code cleanup and remove all code (shutdown tasks or other) that takes care of removing a metric before it can be re-created.

Note: the cleanup actually happens here in SmallRye code https://github.com/smallrye/smallrye-metrics/blob/2.3.3/implementation/src/main/java/io/smallrye/metrics/MetricRegistries.java#L64

jmartisk:
Another reason to do this is that if an extension places metrics into `base` or `vendor` scopes and it does that dynamically at runtime, because the list of metrics is not known during build, it gets quite difficult for the extension to correctly determine when to remove and re-create the metrics.

PR #7099, opened 2020-02-09
Add Mongo Connection Pool Metrics

Postremus:
Add metrics for opened connection pools in the mongo java driver.
Currently, the following metrics are available:
* size of a pool
* amount of currently used connections (checked-out-count)

See also: https://mongodb.github.io/mongo-java-driver/3.12/driver/reference/monitoring/

### DIFF
--- a/docs/src/main/asciidoc/mongodb.adoc
+++ b/docs/src/main/asciidoc/mongodb.adoc
@@ -543,6 +543,14 @@ So when you access the `/health/ready` endpoint of your application you will hav
 
 This behavior can be disabled by setting the `quarkus.mongodb.health.enabled` property to `false` in your `application.properties`.
 
+== Metrics
+
+If you are using the `quarkus-smallrye-metrics` extension, `quarkus-mongodb` can provide metrics about the connection pools.
+This behavior must first be enabled by setting the `quarkus.mongodb.metrics.enabled` property to `true` in your `application.properties`.
+
+So when you access the `/metrics` endpoint of your application you will have information about the connection pool status in the `vendor` scope.
+
+
 == The legacy client
 
 We don't include the legacy MongoDB client by default. It contains the now retired MongoDB Java API (DB, DBCollection,... )
--- a/extensions/mongodb-client/deployment/pom.xml
+++ b/extensions/mongodb-client/deployment/pom.xml
@@ -40,6 +40,11 @@
         </dependency>
         <dependency>
             <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-smallrye-metrics</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
             <artifactId>quarkus-junit5-internal</artifactId>
             <scope>test</scope>
         </dependency>
--- a/extensions/mongodb-client/deployment/src/main/java/io/quarkus/mongodb/deployment/MongoClientBuildTimeConfig.java
+++ b/extensions/mongodb-client/deployment/src/main/java/io/quarkus/mongodb/deployment/MongoClientBuildTimeConfig.java
@@ -11,4 +11,10 @@ public class MongoClientBuildTimeConfig {
      */
     @ConfigItem(name = "health.enabled", defaultValue = "true")
     public boolean healthEnabled;
+
+    /**
+     * Whether or not metrics are published in case the smallrye-metrics extension is present.
+     */
+    @ConfigItem(name = "metrics.enabled")
+    public boolean metricsEnabled;
 }
--- a/extensions/mongodb-client/deployment/src/main/java/io/quarkus/mongodb/deployment/MongoClientProcessor.java
+++ b/extensions/mongodb-client/deployment/src/main/java/io/quarkus/mongodb/deployment/MongoClientProcessor.java
@@ -24,6 +24,7 @@ import org.jboss.jandex.DotName;
 import org.jboss.jandex.IndexView;
 
 import com.mongodb.client.MongoClient;
+import com.mongodb.event.ConnectionPoolListener;
 
 import io.quarkus.arc.Unremovable;
 import io.quarkus.arc.deployment.BeanContainerListenerBuildItem;
@@ -32,6 +33,7 @@ import io.quarkus.arc.deployment.GeneratedBeanBuildItem;
 import io.quarkus.arc.deployment.GeneratedBeanGizmoAdaptor;
 import io.quarkus.arc.deployment.UnremovableBeanBuildItem;
 import io.quarkus.arc.processor.DotNames;
+import io.quarkus.deployment.Capabilities;
 import io.quarkus.deployment.annotations.BuildProducer;
 import io.quarkus.deployment.annotations.BuildStep;
 import io.quarkus.deployment.annotations.Record;
@@ -49,6 +51,7 @@ import io.quarkus.gizmo.MethodCreator;
 import io.quarkus.gizmo.MethodDescriptor;
 import io.quarkus.gizmo.ResultHandle;
 import io.quarkus.mongodb.ReactiveMongoClient;
+import io.quarkus.mongodb.metrics.MongoMetricsConnectionPoolListener;
 import io.quarkus.mongodb.runtime.AbstractMongoClientProducer;
 import io.quarkus.mongodb.runtime.MongoClientConfig;
 import io.quarkus.mongodb.runtime.MongoClientName;
@@ -75,9 +78,16 @@ public class MongoClientProcessor {
     void configureRuntimeProperties(MongoClientRecorder recorder,
             CodecProviderBuildItem codecProvider,
             BsonDiscriminatorBuildItem bsonDiscriminator,
-            MongodbConfig config) {
+            MongodbConfig config,
+            List<MongoConnectionPoolListenerBuildItem> connectionPoolListenerProvider) {
+        List<ConnectionPoolListener> poolListenerList = connectionPoolListenerProvider.stream()
+                .map(MongoConnectionPoolListenerBuildItem::getConnectionPoolListener)
+                .collect(Collectors.toList());
+
         recorder.configureRuntimeProperties(codecProvider.getCodecProviderClassNames(),
-                bsonDiscriminator.getBsonDiscriminatorClassNames(), config);
+                bsonDiscriminator.getBsonDiscriminatorClassNames(),
+                config,
+                poolListenerList);
     }
 
     @BuildStep
@@ -391,5 +401,15 @@ public class MongoClientProcessor {
     HealthBuildItem addHealthCheck(MongoClientBuildTimeConfig buildTimeConfig) {
         return new HealthBuildItem("io.quarkus.mongodb.health.MongoHealthCheck",
                 buildTimeConfig.healthEnabled, "mongodb");
+    }
+
+    @BuildStep
+    void setupMetrics(
+            MongoClientBuildTimeConfig buildTimeConfig, Capabilities capabilities,
+            BuildProducer<MongoConnectionPoolListenerBuildItem> producer) {
+
+        if (buildTimeConfig.metricsEnabled && capabilities.isCapabilityPresent(Capabilities.METRICS)) {
+            producer.produce(new MongoConnectionPoolListenerBuildItem(new MongoMetricsConnectionPoolListener()));
+        }
     }
 }
--- a/extensions/mongodb-client/deployment/src/test/java/io/quarkus/mongodb/MongoMetricsTest.java
+++ b/extensions/mongodb-client/deployment/src/test/java/io/quarkus/mongodb/MongoMetricsTest.java
@@ -1,0 +1,77 @@
+package io.quarkus.mongodb;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNull;
+
+import javax.inject.Inject;
+
+import org.eclipse.microprofile.metrics.Metric;
+import org.eclipse.microprofile.metrics.MetricID;
+import org.eclipse.microprofile.metrics.MetricRegistry;
+import org.eclipse.microprofile.metrics.Tag;
+import org.eclipse.microprofile.metrics.annotation.RegistryType;
+import org.jboss.shrinkwrap.api.ShrinkWrap;
+import org.jboss.shrinkwrap.api.spec.JavaArchive;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.RegisterExtension;
+
+import com.mongodb.client.MongoClient;
+
+import io.quarkus.mongodb.metrics.ConnectionPoolGauge;
+import io.quarkus.test.QuarkusUnitTest;
+
+public class MongoMetricsTest extends MongoTestBase {
+
+    @Inject
+    MongoClient client;
+
+    @Inject
+    @RegistryType(type = MetricRegistry.Type.VENDOR)
+    MetricRegistry registry;
+
+    @RegisterExtension
+    static final QuarkusUnitTest config = new QuarkusUnitTest()
+            .setArchiveProducer(() -> ShrinkWrap.create(JavaArchive.class).addClasses(MongoTestBase.class))
+            .withConfigurationResource("application-metrics-mongo.properties");
+
+    @AfterEach
+    void cleanup() {
+        if (client != null) {
+            client.close();
+        }
+    }
+
+    @Test
+    void testMetricsInitialization() {
+        assertNull(getGaugeValueOrNull("mongodb.connection-pool.size", getTags()));
+        assertNull(getGaugeValueOrNull("mongodb.connection-pool.checked-out-count", getTags()));
+
+        // Just need to execute something so that an connection is opened
+        String name = client.listDatabaseNames().first();
+
+        assertEquals(1L, getGaugeValueOrNull("mongodb.connection-pool.size", getTags()));
+        assertEquals(0L, getGaugeValueOrNull("mongodb.connection-pool.checked-out-count", getTags()));
+
+        client.close();
+        assertEquals(0L, getGaugeValueOrNull("mongodb.connection-pool.size", getTags()));
+        assertEquals(0L, getGaugeValueOrNull("mongodb.connection-pool.checked-out-count", getTags()));
+    }
+
+    private Long getGaugeValueOrNull(String metricName, Tag[] tags) {
+        MetricID metricID = new MetricID(metricName, tags);
+        Metric metric = registry.getMetrics().get(metricID);
+
+        if (metric == null) {
+            return null;
+        }
+        return ((ConnectionPoolGauge) metric).getValue();
+    }
+
+    private Tag[] getTags() {
+        return new Tag[] {
+                new Tag("host", "localhost"),
+                new Tag("port", "27018"),
+        };
+    }
+}
--- a/extensions/mongodb-client/runtime/src/main/java/io/quarkus/mongodb/metrics/MongoMetricsConnectionPoolListener.java
+++ b/extensions/mongodb-client/runtime/src/main/java/io/quarkus/mongodb/metrics/MongoMetricsConnectionPoolListener.java
@@ -1,0 +1,117 @@
+package io.quarkus.mongodb.metrics;
+
+import org.eclipse.microprofile.metrics.Metadata;
+import org.eclipse.microprofile.metrics.Metric;
+import org.eclipse.microprofile.metrics.MetricID;
+import org.eclipse.microprofile.metrics.MetricRegistry;
+import org.eclipse.microprofile.metrics.MetricType;
+import org.eclipse.microprofile.metrics.Tag;
+
+import com.mongodb.connection.ServerId;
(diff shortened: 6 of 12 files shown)


extensions/mongodb-client/deployment/src/main/java/io/quarkus/mongodb/deployment/MongoClientProcessor.java
```diff
@@ -24,20 +24,13 @@
 import com.mongodb.client.MongoClient;
 
 import io.quarkus.arc.Unremovable;
-import io.quarkus.arc.deployment.BeanContainerListenerBuildItem;
-import io.quarkus.arc.deployment.BeanDefiningAnnotationBuildItem;
-import io.quarkus.arc.deployment.GeneratedBeanBuildItem;
-import io.quarkus.arc.deployment.GeneratedBeanGizmoAdaptor;
-import io.quarkus.arc.deployment.UnremovableBeanBuildItem;
+import io.quarkus.arc.deployment.*;
```
machi1990: can we explicit the imports in use and not use start imports here?

extensions/mongodb-client/deployment/src/main/java/io/quarkus/mongodb/deployment/MongoClientProcessor.java
```diff
@@ -24,20 +24,13 @@
...
+import io.quarkus.deployment.Capabilities;
 import io.quarkus.deployment.annotations.BuildProducer;
 import io.quarkus.deployment.annotations.BuildStep;
 import io.quarkus.deployment.annotations.Record;
-import io.quarkus.deployment.builditem.ApplicationArchivesBuildItem;
-import io.quarkus.deployment.builditem.CombinedIndexBuildItem;
-import io.quarkus.deployment.builditem.ExtensionSslNativeSupportBuildItem;
-import io.quarkus.deployment.builditem.FeatureBuildItem;
-import io.quarkus.deployment.builditem.SslNativeConfigBuildItem;
+import io.quarkus.deployment.builditem.*;
```
machi1990: same here about the usage of star imports..

extensions/mongodb-client/runtime/src/main/java/io/quarkus/mongodb/metrics/MongoMetricsConnectionPoolListener.java
```diff
@@ -0,0 +1,83 @@
+package io.quarkus.mongodb.metrics;
+
+import org.eclipse.microprofile.metrics.*;
+
+import com.mongodb.connection.ServerId;
+import com.mongodb.event.*;
+
```
machi1990: let's get rid of wildcard imports..

extensions/mongodb-client/deployment/src/main/java/io/quarkus/mongodb/deployment/MongoClientProcessor.java
Postremus: done

docs/src/main/asciidoc/mongodb.adoc
```diff
@@ -523,6 +523,14 @@ So when you access the `/health/ready` endpoint of your application you will hav
 
 This behavior can be disabled by setting the `quarkus.mongodb.health.enabled` property to `false` in your `application.properties`.
 
+== Metrics
+
+If you are using the `quarkus-smallrye-metrics` extension, `quarkus-mongodb` will automatically add metrics for the connection pools.
```
jmartisk: "Will automatically add metrics" followed by explanation that it must be explicitly enabled is a bit confusing.

extensions/mongodb-client/runtime/src/main/java/io/quarkus/mongodb/metrics/MongoMetricsConnectionPoolListener.java
```diff
@@ -0,0 +1,95 @@
...
+        Counter sizeCount = getMetricRegistry().counter(sizeName, tags);
+        sizeCount.inc();
+    }
+
+    @Override
+    public void connectionRemoved(ConnectionRemovedEvent event) {
+        Tag[] tags = createTags(event.getConnectionId().getServerId());
+
+        Counter sizeCount = getMetricRegistry().counter(sizeName, tags);
+        sizeCount.inc(-1);
```
jmartisk: Counters are only supposed to go up. If you need a metric that goes down as well, it should be a `Gauge`. This would confuse Prometheus, which only expects counters to go up.

extensions/mongodb-client/deployment/src/test/java/io/quarkus/mongodb/MongoMetricsTest.java
```diff
@@ -0,0 +1,65 @@
...
+
+    private MongoClient client;
+    private MetricRegistry metricRegistry;
+
+    @BeforeEach
+    void init() {
+        client = com.mongodb.client.MongoClients
+                .create(MongoClientSettings.builder().applyConnectionString(new ConnectionString(getConnectionString()))
+                        .applyToConnectionPoolSettings(
+                                builder -> builder.addConnectionPoolListener(new MongoMetricsConnectionPoolListener()))
```
jmartisk: IIUC this connection pool listener is needed for updating the metrics, but here you only explictly add it in a test, but how will it be added in production code? I suppose that when `quarkus.mongodb.metrics.enabled` is true, then we need to do this automatically. So something like this needs to be in the place in Quarkus codebase that produces Mongo clients (`AbstractMongoClientProducer` I think?) To be able to do that, `AbstractMongoClientProducer` will need to be able to see the value of `quarkus.mongodb.metrics.enabled` property, so I think it might need to be turned into a runtime property

extensions/mongodb-client/runtime/src/main/java/io/quarkus/mongodb/metrics/MongoMetricsConnectionPoolListener.java
```diff
@@ -0,0 +1,95 @@
...
+    @Override
+    public void connectionPoolOpened(ConnectionPoolOpenedEvent event) {
+        Tag[] tags = createTags(event.getServerId());
+
+        Metadata sizeMetadata = Metadata.builder().withName(sizeName).withType(MetricType.COUNTER)
+                .withDescription("the current size of the pool, including idle and and in-use members").build();
+        Metadata checkedOutCountMetadata = Metadata.builder().withName(checkedOutCountName).withType(MetricType.COUNTER)
+                .withDescription(" the current count of connections that are currently in use").build();
+
+        getMetricRegistry().register(sizeMetadata, new ConnectionPoolCounter(), tags);
```
jmartisk: In development mode, after a reload, the metrics from previous run may still exist in the registry. At the moment we don't automatically deregister vendor metrics during reload (I think we should, I will take care of that soon, but for now we don't). So temporarily this should probably contain a check if the metric already exists, and if yes, it probably means that we just reloaded and the metric should be removed and re-registered (so it will start again from zero).

extensions/mongodb-client/deployment/src/main/java/io/quarkus/mongodb/deployment/MongoClientProcessor.java
```diff
@@ -313,4 +315,13 @@ HealthBuildItem addHealthCheck(MongoClientBuildTimeConfig buildTimeConfig) {
         return new HealthBuildItem("io.quarkus.mongodb.health.MongoHealthCheck",
                 buildTimeConfig.healthEnabled, "mongodb");
     }
+
+    @BuildStep
+    void produceMetricsEnabledProperty(BuildProducer<SystemPropertyBuildItem> producer,
+            MongoClientBuildTimeConfig buildTimeConfig, Capabilities capabilities) {
+
+        if (buildTimeConfig.metricsEnabled && capabilities.isCapabilityPresent(Capabilities.METRICS)) {
+            producer.produce(new SystemPropertyBuildItem("mongodb.metrics.enabled", "true"));
```
jmartisk: What is this for? Is it something that the Mongo client library recognizes? I haven't been able to find what it is used for at runtime.

jmartisk (changes requested):

extensions/mongodb-client/deployment/src/main/java/io/quarkus/mongodb/deployment/MongoClientProcessor.java
```diff
@@ -313,4 +315,13 @@ HealthBuildItem addHealthCheck(MongoClientBuildTimeConfig buildTimeConfig) {
         return new HealthBuildItem("io.quarkus.mongodb.health.MongoHealthCheck",
                 buildTimeConfig.healthEnabled, "mongodb");
     }
+
+    @BuildStep
+    void produceMetricsEnabledProperty(BuildProducer<SystemPropertyBuildItem> producer,
+            MongoClientBuildTimeConfig buildTimeConfig, Capabilities capabilities) {
```
jmartisk: AFAIK the full list of metrics is known at build time? If so, we should register them here (including their tags). See https://github.com/quarkusio/quarkus/blob/master/extensions/agroal/deployment/src/main/java/io/quarkus/agroal/deployment/AgroalProcessor.java#L427 for example

jmartisk (changes requested):

extensions/mongodb-client/deployment/src/main/java/io/quarkus/mongodb/deployment/MongoClientProcessor.java
Postremus: > AFAIK the full list of metrics is known at build time?
We do not necessarily know all tags.
We use the tags host and port here.

In case of the mongo+srv protocoll, the servers are resolved through dns entries.
https://docs.mongodb.com/manual/reference/connection-string/#dns-seedlist-connection-format
We therefore would not know them during build time.

jmartisk: Ok, that's unfortunate. In that case, the metrics won't currently get automatically deregistered during reload in dev mode (until we include a commit for https://github.com/quarkusio/quarkus/issues/7018), therefore `MongoMetricsConnectionPoolListener` will at least temporarily need to take into account the fact that after a reload, the metrics from previous run can still exist, and instead of just registering them, it should delete and re-register them if they exist, so that they can start again from zero

extensions/mongodb-client/deployment/pom.xml
```diff
@@ -58,6 +58,11 @@
             <artifactId>assertj-core</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-smallrye-metrics-deployment</artifactId>
```
loicmathieu: This should be `quarkus-smallrye-metrics-spi`

extensions/mongodb-client/deployment/src/main/java/io/quarkus/mongodb/deployment/MongoClientBuildTimeConfig.java
```diff
@@ -11,4 +11,10 @@
      */
     @ConfigItem(name = "health.enabled", defaultValue = "true")
     public boolean healthEnabled;
+
+    /**
+     * Whether or not metrics are published in case the smallrye-metrics extension is present.
+     */
+    @ConfigItem(name = "metrics.enabled", defaultValue = "false")
```
loicmathieu: `false` is the default value for boolean so it could be omitted

loicmathieu (changes requested):
+1 for registering the listener automatically at build time

extensions/mongodb-client/deployment/pom.xml
Postremus: quarkus-smallrye-metrics-spi does not have the MetricRegistries class.
If I use an `@Inject RegistryType(VENDOR) MetricRegistry` for the Metric Registry, arc can not find an implementation.

How can I just use quarkus-smallrye-metrics-spi in my tests?

jmartisk: `quarkus-smallrye-metrics-spi` basically only contains the `MetricBuildItem` class which is used for registering metrics at build time. If you need your tests to use the metrics extension, importing `quarkus-smallrye-metrics-deployment` is fine IMO. 
Instead of using `MetricRegistries`, which is a class from SmallRye Metrics, I'd suggest staying with just pure MP Metrics API, that is indeed `@Inject RegistryType(VENDOR) MetricRegistry`. If you import `quarkus-smallrye-metrics-deployment` it should work I think

loicmathieu: > If you need your tests to actually use the metrics extension, importing quarkus-smallrye-metrics-deployment is fine IMO.

I'm not agree, if your test need `quarkus-smallrye-metrics-deployment` you should add it with the `test` scope and also import the SPI. With this, someone that don't want metrics can still avoid the inclusion of all the dependencies at runtime.

jmartisk: - it has the `test` scope here
- What's the point in importing the SPI when you're not using anything that it offers?

loicmathieu: @jmartisk sorry, I didn't notice that it has the test scope.

The fact that it didn't use the `MetricBuildItem` is a little annoying as it will be better than all extension register metrics the same way.

jmartisk: I agree it would be convenient, and we try to do that as much as possible, but it's not achievable always. Not always do extensions know the full list of metrics at build time.

Postremus:
> +1 for registering the listener automatically at build time

I have no idea how to do this. The MongoClientSettings class, and the MongoClientSettings.Builder is not serializeable. I therefore can't provide an instance of the settings directly using a build item. Any idea how to do this instead?

For now, I just produce a MongoConnectionPoolListenerBuildItem, containing the listener.

extensions/mongodb-client/deployment/src/test/java/io/quarkus/mongodb/MongoMetricsTest.java
```diff
@@ -0,0 +1,53 @@
...
+        // Metrics get registered once the driver opens the connection pool
+        // The connection pool has a size of 0 by default
+        //assertNull(getCounterValueOrNull("mongodb.connection-pool.size", getTags()));
+        //assertNull(getCounterValueOrNull("mongodb.connection-pool.checked-out-count", getTags()));
+
+        // Just need to execute something so that an connection is opened
+        //String name = client.listDatabaseNames().first();
+
+        //assertEquals(1L, getCounterValueOrNull("mongodb.connection-pool.size", getTags()));
+        //assertEquals(0L, getCounterValueOrNull("mongodb.connection-pool.checked-out-count", getTags()));
```
Postremus: Yeah, I just noticed that I deactivated this test.

jmartisk:
The new solution that you pushed looks nice, that's what I was envisioning! 
I suppose that's what @loicmathieu  meant too, because you can't really register the listener at build time, if the connections are created  at runtime. 
From my PoV, if we just finish `MongoMetricsTest`, then we should be good to go.

extensions/mongodb-client/deployment/src/main/java/io/quarkus/mongodb/deployment/MongoClientProcessor.java
```diff
@@ -75,9 +78,14 @@ UnremovableBeanBuildItem markBeansAsUnremovable() {
     void configureRuntimeProperties(MongoClientRecorder recorder,
             CodecProviderBuildItem codecProvider,
             BsonDiscriminatorBuildItem bsonDiscriminator,
-            MongodbConfig config) {
+            MongodbConfig config, List<MongoConnectionPoolListenerBuildItem> connectionPoolListenerProvider) {
```
loicmathieu: Can you move the connectionPoolListenerProvider on a new line, this will be easier to read

extensions/mongodb-client/deployment/src/main/java/io/quarkus/mongodb/deployment/MongoClientProcessor.java
```diff
@@ -332,4 +340,14 @@ HealthBuildItem addHealthCheck(MongoClientBuildTimeConfig buildTimeConfig) {
         return new HealthBuildItem("io.quarkus.mongodb.health.MongoHealthCheck",
                 buildTimeConfig.healthEnabled, "mongodb");
     }
+
+    @BuildStep
+    void produceMongoClientSettings(
```
loicmathieu: We usually name a build step method by what it does. Here you setup metrics or you add the listener so maybe `setupMetrics` or `addConnectionPoolListener` are better names.

extensions/mongodb-client/runtime/src/main/java/io/quarkus/mongodb/metrics/MongoMetricsConnectionPoolListener.java
```diff
@@ -0,0 +1,116 @@
...
+import com.mongodb.event.ConnectionPoolListener;
+import com.mongodb.event.ConnectionPoolOpenedEvent;
+import com.mongodb.event.ConnectionPoolWaitQueueEnteredEvent;
+import com.mongodb.event.ConnectionPoolWaitQueueExitedEvent;
+import com.mongodb.event.ConnectionRemovedEvent;
+
+import io.smallrye.metrics.MetricRegistries;
+
+public class MongoMetricsConnectionPoolListener implements ConnectionPoolListener {
+    private final static String sizeName = "mongodb.connection-pool.size";
```
loicmathieu: constants are usally defined as `private static final` and are uppercase.
```java
private static final String SIZE_PROPERTY = "mongodb.connection-pool.size";
```

Same for `checkedOutCountName`.

loicmathieu (approved):
I added some small code changes not mandatory so I aprove the PR.

Postremus:
ci failure seems unrelated

jmartisk (approved):

Postremus:
@cescoffier would you please take a look? Would be kind of nice to have this in 1.3.0.

Postremus:
I rebased and resolved the conflicts.

Current Build failure seems unrelated. Resteasy classes are missing.
`
[ERROR] testAggregate  Time elapsed: 0 s  <<< ERROR!
java.lang.NoClassDefFoundError: org/jboss/resteasy/core/ResteasyContext
	at io.smallrye.context.SmallRyeThreadContext.currentContextExecutor(SmallRyeThreadContext.java:208)`

docs/src/main/asciidoc/mongodb.adoc
```diff
@@ -543,6 +543,14 @@ So when you access the `/health/ready` endpoint of your application you will hav
 
 This behavior can be disabled by setting the `quarkus.mongodb.health.enabled` property to `false` in your `application.properties`.
 
+== Metrics
+
+If you are using the `quarkus-smallrye-metrics` extension, `quarkus-mongodb` con provide metrics about the connection pools.
```
geoand: `con`?

extensions/mongodb-client/deployment/src/main/java/io/quarkus/mongodb/deployment/MongoClientProcessor.java
```diff
@@ -114,7 +124,7 @@ BsonDiscriminatorBuildItem collectBsonDiscriminators(CombinedIndexBuildItem inde
      * Create a producer bean managing the lifecycle of the MongoClient.
      * <p>
      * The generated class will look like
-     *
```
geoand: Any reason this change was made or what it simply done by the formatter? Dittor for the similar changes below

extensions/mongodb-client/runtime/src/main/java/io/quarkus/mongodb/runtime/AbstractMongoClientProducer.java
```diff
@@ -295,6 +297,13 @@ private MongoClientSettings createMongoConfiguration(MongoClientConfig config) {
         if (config.readPreference.isPresent()) {
             settings.readPreference(ReadPreference.valueOf(config.readPreference.get()));
         }
+
+        if (!connectionPoolListeners.isEmpty()) {
+            settings.applyToConnectionPoolSettings(builder -> {
```
geoand: Please change this to not use lambdas. We try very hard to not use lambdas in runtime code because a small startup time and have a small memory footprint

extensions/mongodb-client/runtime/src/main/java/io/quarkus/mongodb/runtime/AbstractMongoClientProducer.java
```diff
@@ -295,6 +297,13 @@ private MongoClientSettings createMongoConfiguration(MongoClientConfig config) {
...
+                connectionPoolListeners.forEach(builder::addConnectionPoolListener);
```
geoand: Same for method references :)

geoand (commented):
I added some minor comments. After addressing them please rebase onto master in order to pick the latest CI changes

extensions/mongodb-client/deployment/src/main/java/io/quarkus/mongodb/deployment/MongoClientProcessor.java
Postremus: I think this might just have been to formatter, I will check.

geoand (approved):

machi1990:
Merged, thanks @Postremus

Issue #7097, opened 2020-02-09
Introducing Mutiny


cescoffier:
### Description
The objective of this epic is to gather all the tasks related to the introduction of Mutiny in Quarkus, as well as the deprecation of Axle and RX APIs, documentation and quickstart updates...

### Tasks
- [X] Create the Mutiny extension - https://github.com/quarkusio/quarkus/pull/7098
- [X] Integrate Mutiny in the Vert.x extension, and manage Axle / RX deprecation - https://github.com/quarkusio/quarkus/pull/7098
- [X] Integrate Mutiny with `@ConsumeEvent` - https://github.com/quarkusio/quarkus/pull/7098
- [x] Integrate Mutiny with RestEasy - https://github.com/quarkusio/quarkus/pull/7177
- [x] Integrate Mutiny with Context propagation - https://github.com/quarkusio/quarkus/pull/7303
- [x] Update Vert.x quickstart to use Mutiny instead of Axle - #7388 
- [X] Integrate Mutiny in Qute - https://github.com/quarkusio/quarkus/pull/7101
- [x] Update the mailer extension as well as the guide - https://github.com/quarkusio/quarkus/pull/7211
- [x] Integrate the Mutiny API with the PostGreSQL and MySQL reactive client - this includes updating the guide - https://github.com/quarkusio/quarkus/pull/7216
- [x] Integrate Mutiny API in the reactive Mongo API - https://github.com/quarkusio/quarkus/pull/7263
- [x] Write a Mutiny quickstart
- [x] Integrate Mutiny Verticle - https://github.com/quarkusio/quarkus/pull/7231
- [x] Add Uni and Multi support to Reactive Messaging -    _partially done_
- [x] Switch Mail Qute template support to Mutiny 
- [x] Update Panache Mongo (reactive version) to provide a Mutiny API #7327 
- [x] Use Mutiny API in Quarkus Security

loicmathieu:
@cescoffier it's still time for MongoDB with Panache to replace the Axle base implementation to a Mutiny one, I can do it easily.

Provinding both implementation is more works (the deployment part of MongoDB with Panache is complexe).

I can work on it this week based on your mongodb-client branch, my only question is do we expose Uni/Multi or do we stick with CompletionStage/Publisher ?

cescoffier:
@loicmathieu ideally both. In your Panache code, the Axle model is in the `axle` package. Just create the mutiny one in the `reactive` package.

So:

[...]

PR #7101, opened 2020-02-09
Integration of Qute and Mutiny

cescoffier:
This PR is part of the #7097 epic. It re-implement the reactive support of Qute to use Mutiny instead of RX:

* Implement a new Publisher Factory using Mutiny
* Remove the RX factory
* Switch the extension to use the new reactive support
* Add some tests
* Extend documentation


docs/src/main/asciidoc/qute.adoc
```diff
@@ -263,6 +263,29 @@ public class ReportGenerator {
...
+The returned `Publisher` is an instance of `io.smallrye.mutiny.Multi`.
+
+It is possible to create an instance of `io.smallrye.mutiny.Uni` as follows:
+
+[source, java]
+----
+Uni<String> uni = Uni.createFrom().completionStage(() -> template.data("name", "neo").renderAsync());
+----
+
+In this case, the rendering only starts once the subscriber requests it.
```
gsmet: Looks like it's the same in both cases? Or did I misunderstand something?

gsmet (approved):
Looks good to me but I'll let @mkouba have the final say on this one.

docs/src/main/asciidoc/qute.adoc
cescoffier: If you use a `renderAsync()` the rendering process starts even if you don't have anyone listening / prepared to receive the output. With Uni or Multi, it would only start if someone is there to receive it.

mkouba (approved):

PR #7106, opened 2020-02-09
Reset the TCCL for each test

stuartwdouglas:
This makes sure the correct TCCL is set if there is a mix
of @QuarkusTest and normal tests


aloubyansky:
Are beforeAll/afterAll called more than once in this case?

stuartwdouglas:
The are called before and after every @QuarkusTest test class. The issue is that if we don't reset it in between classes and the test suite contains non-quarkus tests (e.g. tests with mocks) then these tests will run with the wrong class loader. This will break Mockito.

aloubyansky (approved):

sean-scott-lr:
@stuartwdouglas Thank you, this does fix my mockito error.

PR #7107, opened 2020-02-09
Stop dealing with the SunEC library during the native image generation

### LINKED ISSUE #6939: Stop dealing with the SunEC library during the native image generation
GraalVM `19.3.0` removed the need to ship the SunEC library with the native image:

> GraalVM 19.3 switched to using the JDK native code instead of manual substitutions. For GraalVM Native Image this switch to Java Native Interface (JNI) platform enabled the possibility to provide support for JDK 11 and extended support for Windows OS. It has no impact on startup time or memory footprint, and eliminates the need for shipping JDK libraries such as libsunec.so along with native images that use Java crypto services. GraalVM now ships with statically linkable version of the JDK libraries.
Note that the sunec native library of SunEC provider gets statically linked into an image if required (see #951). Thus native images do not depend on the sunec shared library anymore at runtime. However, if SunEC provider gets statically linked into an image on Linux and macOS the image will depend on libstdc++.

We currently support GraalVM `19.3.1` but we're still dealing with this lib in `NativeImageConfigBuildStep` because of some old code that should be removed.

I'll submit a PR.

gwenneg:
Fixes #6939

### DIFF
--- a/core/deployment/src/main/java/io/quarkus/deployment/steps/NativeImageConfigBuildStep.java
+++ b/core/deployment/src/main/java/io/quarkus/deployment/steps/NativeImageConfigBuildStep.java
@@ -1,13 +1,11 @@
 package io.quarkus.deployment.steps;
 
-import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
 
-import org.graalvm.nativeimage.ImageInfo;
 import org.jboss.logging.Logger;
 
 import io.quarkus.deployment.annotations.BuildProducer;
@@ -35,8 +33,6 @@ import io.quarkus.runtime.ssl.SslContextConfigurationRecorder;
 class NativeImageConfigBuildStep {
 
     private static final Logger log = Logger.getLogger(NativeImageConfigBuildStep.class);
-
-    private static final String LIB_SUN_EC = "libsunec.so";
 
     @BuildStep
     @Record(ExecutionTime.STATIC_INIT)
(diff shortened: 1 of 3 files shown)


gwenneg:
@gsmet I am wondering if `JavaLibraryPathAdditionalPathBuildItem` should deprecated (and deleted later) or if it's still useful for another reason than the SunEC lib.

core/deployment/src/main/java/io/quarkus/deployment/steps/NativeImageConfigBuildStep.java
```diff
@@ -91,33 +84,14 @@ void build(SslContextConfigurationRecorder sslContextConfigurationRecorder,
 
             if (graalVmHome != null) {
                 Path graalVmLibDirectory = Paths.get(graalVmHome, "jre", "lib");
-                Path linuxLibDirectory = graalVmLibDirectory.resolve("amd64");
-                Path linuxPath = linuxLibDirectory.resolve(LIB_SUN_EC);
 
                 // We add . as it might be useful in a containerized world
                 javaLibraryPathAdditionalPath.produce(new JavaLibraryPathAdditionalPathBuildItem("."));
```
gwenneg: @gsmet Do we want to keep this? I'd answer no but I'm not sure if it was added for the SunEC lib or some other reason.

core/deployment/src/main/java/io/quarkus/deployment/steps/NativeImageConfigBuildStep.java
```diff
@@ -91,33 +84,14 @@ void build(SslContextConfigurationRecorder sslContextConfigurationRecorder,
...
-                    // On MacOS, the SunEC library is directly in jre/lib/
-                    // This is useful for testing or if you have a similar environment in production
-                    javaLibraryPathAdditionalPath
-                            .produce(new JavaLibraryPathAdditionalPathBuildItem(graalVmLibDirectory.toString()));
-                }
 
                 // This is useful for testing but the user will have to override it.
                 sslTrustStoreSystemProperty.produce(
                         new SslTrustStoreSystemPropertyBuildItem(
                                 graalVmLibDirectory.resolve(Paths.get("security", "cacerts")).toString()));
```
gwenneg: I suppose we want to keep this around. I'll need it to be confirmed.

gwenneg:
I'll update the `native-and-ssl` guide before promoting the PR, but I need answers to the questions above first.

core/deployment/src/main/java/io/quarkus/deployment/steps/NativeImageConfigBuildStep.java
gsmet: Yes.

core/deployment/src/main/java/io/quarkus/deployment/steps/NativeImageConfigBuildStep.java
gsmet: I think we can remove it.

gsmet (changes requested):
Thanks for working on that one. Comments inline.

gwenneg:
Oops, something's wrong with that rebase, let me do it again :)

gwenneg:
@gsmet Thanks for the review! I think we're good to go now.

gsmet (approved):

Issue #7087, opened 2020-02-07
quarkus-arquillian having issue process app using CompletionStage


rsearls:
Deploying the app results is the following error message and none of
the test methods being called. I have traced the error into io.quarkus.arc.processor.BeanArchives
but have been unsuccessful determining root cause.

    [INFO] Running org.jboss.resteasy.test.asynch.AsyncInjectionTest
    18:07:48,889 INFO  [org.jbo.threads] JBoss Threads version 3.0.0.Final
    18:07:49,150 WARN  [io.qua.arc.pro.BeanArchives] Failed to index boolean: Stream closed
    18:07:49,153 WARN  [io.qua.arc.pro.BeanArchives] Failed to index char: Stream closed
    18:07:49,154 WARN  [io.qua.arc.pro.BeanArchives] Failed to index byte: Stream closed
    18:07:49,155 WARN  [io.qua.arc.pro.BeanArchives] Failed to index short: Stream closed
    18:07:49,157 WARN  [io.qua.arc.pro.BeanArchives] Failed to index int: Stream closed
    18:07:49,158 WARN  [io.qua.arc.pro.BeanArchives] Failed to index long: Stream closed
    18:07:49,159 WARN  [io.qua.arc.pro.BeanArchives] Failed to index float: Stream closed
    18:07:49,160 WARN  [io.qua.arc.pro.BeanArchives] Failed to index double: Stream closed
    [INFO] Tests run: 0, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 3.106 s - in org.jboss.resteasy.test.asynch.AsyncInjectionTest

Execution env.
    jdk 11.0.2
    mvn 3.6.0
    quarkus 1.2.0.Final

  This project is provided to demonstrate the issue
https://github.com/rsearls/quarkus-BeanArchives-index.git

  Follow the directions in the README file.

gwenneg:
Hi @rsearls and thanks for reporting this.

You need to fix `-Dfile=FIX_THIS/lib/utils-arquillian-utils-4.5.0-SNAPSHOT.jar` (from your repository README.md) before we can use your reproducer, right?

rsearls:
No.  If you install utils-arquillian-utils as directed there is no issue.  This archive only contains a utility used by the test to generate the test's url. 
Once you install this archive compile the project,
  mvn clean test -DskipTests=true
Then run the test.
mvn test
You will see the issue.

gwenneg:
[...]

PR #7108, opened 2020-02-10
Stop trying to index primitive types annotated with @Context

gwenneg:
This PR suppresses the warn messages mentioned in #7087.

I'm submitting this as a draft for now since I'm not sure we'll want to treat the issue that way.


gwenneg:
I'm promoting this PR as it seems annotating primitive types with `@Context` is supported by RestEasy (see [here](https://github.com/quarkusio/quarkus/issues/7087#issuecomment-584111361)).

The idea here is to skip the `NativeImageProxyDefinitionBuildItem` creation logic because it doesn't make sense to do that for primitive values and it causes warn messages to be displayed in the log from `BeanArchives.index(Indexer, String, ClassLoader)`.

gsmet (approved):

gsmet:
Merged, thanks!

PR #7109, opened 2020-02-10
Add support for graceful shutdown

stuartwdouglas:
This adds the core support for waiting for requests
to finish before shutting down, and a HTTP implementation
for web requests.

### DIFF
--- a/core/runtime/src/main/java/io/quarkus/runtime/shutdown/ShutdownConfig.java
+++ b/core/runtime/src/main/java/io/quarkus/runtime/shutdown/ShutdownConfig.java
@@ -1,0 +1,25 @@
+package io.quarkus.runtime.shutdown;
+
+import java.time.Duration;
+import java.util.Optional;
+
+import io.quarkus.runtime.annotations.ConfigItem;
+import io.quarkus.runtime.annotations.ConfigPhase;
+import io.quarkus.runtime.annotations.ConfigRoot;
+
+@ConfigRoot(phase = ConfigPhase.RUN_TIME)
+public class ShutdownConfig {
+
+    /**
+     * The timeout to wait for running requests to finish. If this is not set then the application will exit immediately.
+     * 
+     * Setting this timeout will incur a small performance penalty, as it requires active requests to be tracked.
+     */
+    @ConfigItem
+    public Optional<Duration> timeout;
+
+    public boolean isShutdownTimeoutSet() {
+        return timeout.isPresent() && timeout.get().toMillis() > 0;
+    }
+
+}
--- a/core/runtime/src/main/java/io/quarkus/runtime/shutdown/ShutdownListener.java
+++ b/core/runtime/src/main/java/io/quarkus/runtime/shutdown/ShutdownListener.java
@@ -1,0 +1,44 @@
+package io.quarkus.runtime.shutdown;
+
+/**
+ * A listener that can be registered to control the shutdown process and implement
+ * graceful shutdown.
+ *
+ * Shutdown happens in two phases. In the pre shutdown phase the application should
+ * function normally, but should notify external systems that it is about to go away.
+ *
+ * In the shutdown phase the app should disallow new external requests, however
+ * allow existing requests to complete normally.
+ */
+public interface ShutdownListener {
+
+    /**
+     * Pre shutdown notification, the listener can use this to notify external
+     * systems this application is about to shut down.
+     * 
+     * @param notification The notification event
+     */
+    default void preShutdown(ShutdownNotification notification) {
+        notification.done();
+    }
+
+    /**
+     * The shutdown notification. The listener should start rejecting requests
+     * and wait for all existing ones to finish.
+     *
+     * @param notification The notification event
+     */
+    default void shutdown(ShutdownNotification notification) {
+        notification.done();
+    }
+
+    interface ShutdownNotification {
+
+        /**
+         * This method must be invoked when the lister has done it's work,
+         * to allow shutdown to proceed to the next stage.
+         */
+        void done();
+    }
+
+}
--- a/extensions/smallrye-health/runtime/src/main/java/io/quarkus/smallrye/health/runtime/ShutdownReadinessCheck.java
+++ b/extensions/smallrye-health/runtime/src/main/java/io/quarkus/smallrye/health/runtime/ShutdownReadinessCheck.java
@@ -1,0 +1,27 @@
+package io.quarkus.smallrye.health.runtime;
+
+import javax.inject.Singleton;
+
+import org.eclipse.microprofile.health.HealthCheck;
+import org.eclipse.microprofile.health.HealthCheckResponse;
+import org.eclipse.microprofile.health.Readiness;
+
+@Singleton
+@Readiness
+public class ShutdownReadinessCheck implements HealthCheck {
+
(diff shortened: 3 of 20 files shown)


extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/filters/GracefulShutdownFilter.java
```diff
@@ -0,0 +1,68 @@
...
+
+    @Override
+    public void handle(HttpServerRequest event) {
+        if (!running) {
+            event.response().setStatusCode(HttpResponseStatus.SERVICE_UNAVAILABLE.code()).end();
+            return;
+        }
+        currentRequestCount.incrementAndGet();
+        //todo: some way to do this without a wrapper solution
+        next.handle(new ShutdownRequestWrapper(event, requestDoneHandler));
```
stuartwdouglas: @vietj is there a better way to handle this?

Basically I need a reliable way to know when a request is done so I know how many requests are running (that can't be overridden in a downstream handler), and wrapping the request and response like this was the only way I could figure out to do it.

gsmet:
I think this one should go into 1.3. @vietj @cescoffier could you have a look?

gsmet:
@vietj @cescoffier ping?

core/runtime/src/main/java/io/quarkus/runtime/shutdown/ShutdownConfig.java
```diff
@@ -0,0 +1,24 @@
+package io.quarkus.runtime.shutdown;
+
+import java.time.Duration;
+import java.util.Optional;
+
+import io.quarkus.runtime.annotations.ConfigItem;
+import io.quarkus.runtime.annotations.ConfigRoot;
+
+@ConfigRoot
```
machi1990: Maybe add a runtime config phase so that it is runtime overridable?

extensions/vertx-http/deployment/src/test/java/io/quarkus/vertx/http/shutdown/ShutdownTest.java
```diff
@@ -0,0 +1,97 @@
...
+import org.junit.jupiter.api.extension.RegisterExtension;
+
+import io.quarkus.test.QuarkusUnitTest;
+import io.quarkus.test.common.http.TestHTTPResource;
+import io.vertx.core.Handler;
+import io.vertx.ext.web.Router;
+import io.vertx.ext.web.RoutingContext;
+
+/**
+ * Tests that shutddown will wait for current requests to finish.
```
machi1990:
```suggestion
 * Tests that shutdown will wait for current requests to finish.
```

extensions/vertx-http/deployment/src/test/java/io/quarkus/vertx/http/shutdown/ShutdownTimeoutTest.java
```diff
@@ -0,0 +1,90 @@
...
+import org.junit.jupiter.api.extension.RegisterExtension;
+
+import io.quarkus.test.QuarkusUnitTest;
+import io.quarkus.test.common.http.TestHTTPResource;
+import io.vertx.core.Handler;
+import io.vertx.ext.web.Router;
+import io.vertx.ext.web.RoutingContext;
+
+/**
+ * Tests that shutddown will wait for current requests to finish, up to the timeout specified.
```
machi1990:
```suggestion
 * Tests that shutdown will wait for current requests to finish, up to the timeout specified.
```

stuartwdouglas:
Rebased, I am guessing this is not going to make 1.3

gsmet:
@vietj @cescoffier ?

core/runtime/src/main/java/io/quarkus/runtime/shutdown/ShutdownListener.java
```diff
@@ -0,0 +1,44 @@
...
+ */
+public interface ShutdownListener {
+
+    /**
+     * Pre shutdown notification, the listener can use this to notify external
+     * systems this application is about to shut down.
+     * 
+     * @param notification The notification event
+     */
+    default void preShutdown(ShutdownNotification notification) {
```
cescoffier: in terms of API, any reason to not return a `CompletionStage`?

cescoffier: Typically how are reported failures?

extensions/smallrye-health/runtime/src/main/java/io/quarkus/smallrye/health/runtime/ShutdownReadinessCheck.java
```diff
@@ -0,0 +1,27 @@
+package io.quarkus.smallrye.health.runtime;
+
+import javax.inject.Singleton;
+
+import org.eclipse.microprofile.health.HealthCheck;
+import org.eclipse.microprofile.health.HealthCheckResponse;
+import org.eclipse.microprofile.health.Readiness;
+
+@Singleton
```
cescoffier: One day I would understand why sometimes we use @ApplicationScoped, and sometimes @Singleton (I know it's not coming from the same spec, but still puzzled by when use which).

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/filters/GracefulShutdownFilter.java
cescoffier: on the HTTP Server Response, you can attach a https://vertx.io/docs/apidocs/io/vertx/core/http/HttpServerResponse.html#endHandler-io.vertx.core.Handler- to be notified when the response has been written.

cescoffier: It would need to be coupled with the closeHandler when the connection is closed (abruptly).

cescoffier (approved):
I've made a few comments, but it can be improved later.

core/runtime/src/main/java/io/quarkus/runtime/shutdown/ShutdownListener.java
stuartwdouglas: Failures don't really matter, we are shutting down anyway, if a stage has failed then it needs to log and notify, so I am not sure if completion stage works here.

extensions/smallrye-health/runtime/src/main/java/io/quarkus/smallrye/health/runtime/ShutdownReadinessCheck.java
stuartwdouglas: ApplicationScoped requires a proxy to comply with the spec, Singleton is usually better, but because it is not actually in the CDI spec a lot of people just default to application scoped.

stuartwdouglas:
The end and close handlers would be overwritten by Undertow, which also needs this info, and could be overwritten by any user handler.

stuartwdouglas:
Ideally there would be an addCloseHandler and addEndHandler API that lets you add them rather than just having a single one.

gsmet:
Merging this first version per discussion with @cescoffier .

PR #7110, opened 2020-02-10
Exclude jackson-module-scala

stuartwdouglas:
This breaks jackson if it is included and scala is not present

### DIFF
--- a/bom/runtime/pom.xml
+++ b/bom/runtime/pom.xml
@@ -2167,6 +2167,10 @@
                         <artifactId>jackson-core</artifactId>
                     </exclusion>
                     <exclusion>
+                        <groupId>com.fasterxml.jackson.module</groupId>
+                        <artifactId>jackson-module-scala_2.12</artifactId>
+                    </exclusion>
+                    <exclusion>
                         <groupId>org.scala-lang</groupId>
                         <artifactId>scala-reflect</artifactId>
                     </exclusion>


geoand (approved):

bom/runtime/pom.xml
```diff
@@ -2166,6 +2166,10 @@
                         <groupId>com.fasterxml.jackson.core</groupId>
                         <artifactId>jackson-core</artifactId>
                     </exclusion>
+                    <exclusion>
+                        <groupId>com.fasterxml.jackson.module</groupId>
+                        <artifactId>jackson-module-scala_2.12</artifactId>
```
aloubyansky: It's unfortunate that the version appears in the artifactId. I'm wondering whether it'll be a better maintenance compromise to exclude everything under `com.fasterxml.jackson.module` in this case. At this point this is the only artifact from this groupId that appears as a dependency of `io.vertx:vertx-kafka-client` through `org.apache.kafka:kafka_2.12`.
Unfortunately, Maven does not support exclusion expressions, it's either `*` or specific artifactId.

aloubyansky:
```
[INFO] +- org.apache.kafka:kafka_2.12:jar:2.3.0:compile
[INFO] |  +- com.fasterxml.jackson.module:jackson-module-scala_2.12:jar:2.9.9:compile
[INFO] |  |  \- com.fasterxml.jackson.module:jackson-module-paranamer:jar:2.9.9:compile
```

stuartwdouglas: I have changed it to a * exclude

aloubyansky:
Hm, `org.slf4j:slf4j-log4j12` is supposed to be excluded from `io.vertx:vertx-kafka-client`.

bom/runtime/pom.xml
stuartwdouglas: Looks like the * approach does not work, I think it must only work to exclude everything.

PR #7113, opened 2020-02-10
Remove Caffeine ForkJoinPool#commonPool fix

pivovarit:
GraalVM 19.3.1 features [a deferring substitute of ForkJoinPool#commonPool](https://github.com/oracle/graal/pull/1542) and this workaround is no longer needed

### DIFF
--- a/extensions/cache/runtime/src/main/java/io/quarkus/cache/runtime/caffeine/CaffeineCache.java
+++ b/extensions/cache/runtime/src/main/java/io/quarkus/cache/runtime/caffeine/CaffeineCache.java
@@ -6,7 +6,6 @@ import static io.quarkus.cache.runtime.NullValueConverter.toCacheValue;
 import java.time.Duration;
 import java.util.concurrent.Callable;
 import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.ForkJoinPool;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 import java.util.function.Supplier;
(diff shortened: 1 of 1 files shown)


extensions/cache/runtime/src/main/java/io/quarkus/cache/runtime/caffeine/CaffeineCache.java
```diff
@@ -31,11 +31,7 @@
     public CaffeineCache(CaffeineCacheInfo cacheInfo) {
         this.name = cacheInfo.name;
         Caffeine<Object, Object> builder = Caffeine.newBuilder();
-        /*
-         * The following line is a workaround for a GraalVM issue: https://github.com/oracle/graal/issues/1524
-         * TODO: Remove it as soon as a Quarkus release depends on GraalVM 19.3.0 or higher.
-         */
-        builder.executor(task -> ForkJoinPool.commonPool().execute(task));
```
gwenneg: Good catch! This should have been removed in #6792 when we dropped the GraalVM 19.2.1 support.

The build failed because you need to remove the `ForkJoinPool` import. Could you fix that please?

gwenneg (changes requested):

extensions/cache/runtime/src/main/java/io/quarkus/cache/runtime/caffeine/CaffeineCache.java
pivovarit: Sorry, fixed

gwenneg: Thanks! Let's see if CI is happy with this in native mode now.

gwenneg (approved):

extensions/cache/runtime/src/main/java/io/quarkus/cache/runtime/caffeine/CaffeineCache.java
pivovarit: Should be, I went through the same issue with http://github.com/quarkusio/quarkus/pull/6758 :)

gwenneg:
Merged, thanks for your help with this @pivovarit!

PR #7114, opened 2020-02-10
Update base image in multistage native doc

g1raffi:
The base image was renamed and does not work as suggested, the postfix -java11 / -java8 has to be added.

### DIFF
--- a/docs/src/main/asciidoc/building-native-image.adoc
+++ b/docs/src/main/asciidoc/building-native-image.adoc
@@ -347,7 +347,7 @@ In this guide we will use the first stage to generate the native executable usin
 [source,dockerfile,subs=attributes+]
 ----
 ## Stage 1 : build with maven builder image with native capabilities
-FROM quay.io/quarkus/centos-quarkus-maven:{graalvm-version} AS build
+FROM quay.io/quarkus/centos-quarkus-maven:{graalvm-version}-java8 AS build
 COPY src /usr/src/app/src
 COPY pom.xml /usr/src/app
 USER root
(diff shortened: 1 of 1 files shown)


loicmathieu:
I'm not sure we want to communicate for java 11 native image for now as it's still experimental on GraalVM side.

@gsmet do we need to default to java 8 or 11 for native image base docker image ?

g1raffi:
@loicmathieu  Understandable. Unfortunately yes. There is no base image without defining the java version at the moment, if you take a look at the Quay registry (https://quay.io/repository/quarkus/centos-quarkus-maven?tab=tags). 
I'm just a fan of quick starter guides which do work without any changes needed. For persons not familiar with docker and their registries the error given might create unnecessary confusion.

loicmathieu:
@g1raffi maybe I was not clear, I agree that it needs to be corrected, but I'm not sure the default image must be the java 11 or the java 8 ones.

docs/src/main/asciidoc/building-native-image.adoc
```diff
@@ -369,6 +369,8 @@ Save this file in `src/main/docker/Dockerfile.multistage` as it is not included
 
 [WARNING]
 ====
+Replace the `-java11` in the base image with `-java8` if you want to use Java 8.
```
gsmet: As mentioned by Loïc, let's do the opposite for now.

We will migrate to Java 11 by default soon but right now we still use java8 as the default for native images.

gsmet (dismissed):
Thanks a lot for spotting that one! Could you have a look at my comment?

g1raffi:
@gsmet @loicmathieu Alright, I switched the default image to java 8 and updated the comment below accordingly.

loicmathieu (approved):
LGTM

loicmathieu:
@g1raffi can you squashed the two commits into one please ?

g1raffi:
@loicmathieu done!

gastaldi (approved):

PR #7115, opened 2020-02-10
Support Hibernate ORM's generation of Enhanced Proxies

### LINKED ISSUE #6063: HibernateProxy at runtime is not allowed when the configured BytecodeProvider
**Describe the bug**
When run the native image, the hibernate can't load the proxy

**Expected behavior**
Load proxy of all collections and lazy relations

**Actual behavior**
Throw exception -> HibernateProxy instances at runtime is not allowed when the configured BytecodeProvider  is 'none'; your model requires a more advanced BytecodeProvider to be enabled.

**To Reproduce**
1. Get the reproducer project https://github.com/rafaelkloss/quarkus-many-to-one-error
2. config the application resources (database postgres)
3. generate native image -> docker build --build-arg version=homolog -t bug-proxy .
4. docker run image -> docker run -it --rm --name bug-proxy -p 8080:8080 bug-proxy
5. make a get to http://localhost:8080/v1/feedback/create-project (create a project with a lot of relations)
6. make a get to http://localhost:8080/v1/feedback/error-proxy to get project above, the error throw

**Additional context**
This is a recurrent issue after quarkus 0.22.0

Sanne:
The high level ideas is very simple: the proxy class definitions need to be generated before we reach "runtime". Instances of such class can be instantiated later.

The details of boot in ORM made this quite complex though; this needed first to revisit how tuplizers are initialized, and make the BytecodeEnhancers a stateful service which could be injected and hold some sessionfactory specific state; the complexity mainly was that Bytecode enhancers were being used as a global static helper across the codebase.

Also, needs to actually make sure we can extrapolate from the metadata all proxy class definitions which will be needed at runtime; there's a fallback implementation in place which should mimick the same fallback patterns as ORM "vanilla": it's not always possible to use the Enhanced Proxy feature, but we can degrade gracefully to have the same semantics as not having this turned on on specific entities.

Fixes #6063 

See also:
 - https://hibernate.atlassian.net/browse/HHH-13804
 - https://hibernate.atlassian.net/browse/HHH-13849
 - https://hibernate.atlassian.net/browse/HHH-13851
 - https://hibernate.atlassian.net/browse/HHH-13850


Sanne:
Once CI build was cancelled; I'll take the opportunity to rebase.

extensions/hibernate-orm/deployment/src/test/java/io/quarkus/hibernate/orm/HibernateMetricsTestCase.java
```diff
@@ -64,7 +60,8 @@ public void testMetrics() {
         assertEquals(0L, getCounterValueOrNull("hibernate-orm.entities.inserted"));
         Arc.container().requestContext().activate();
         try {
-            DummyEntity entity = new DummyEntity(12345L);
+            DummyEntity entity = new DummyEntity();
```
gsmet: Does that one throw the exception? It needs proxying?

gsmet (approved):
It looks good, I just added one question.

extensions/hibernate-orm/deployment/src/test/java/io/quarkus/hibernate/orm/HibernateMetricsTestCase.java
Sanne: That code wasn't throwing an exception, as it was only storing a `DummyEntity`. If the test would have attempted to load it, you'd have had a runtime exception.

That seemed wrong to me (in many ways), so I made sure we throw an helpful (but fatal) error at build time to highlight the need for an empty constructor. Since it fails now, existing tests had to be fixed.

Sanne: (that specific test doesn't need proxying)

Issue #7000, opened 2020-02-05
QuarkusUnitTest: assertException shadows out setExpectedException


ge0ffrey:
In the testcase below, this unit test is red (which is correct):
```
            new QuarkusUnitTest()
            ...
            .setExpectedException(ArithmeticException.class); // RED, which is correct
```
This unit test is green (which is also correct):
```
            new QuarkusUnitTest()
            ...
            .assertException(throwable -> {
                assertTrue(throwable instanceof IllegalArgumentException);
                assertTrue(throwable.getMessage().contains("scanAnnotatedClasses"));
            }); // GREEN, which is correct
```
Obviously, combining both of these (a green and a red test) should give me a red result. But this test is green (which is wrong):
```
            new QuarkusUnitTest()
            ...
            .setExpectedException(ArithmeticException.class);
            .assertException(throwable -> {
                assertTrue(throwable instanceof IllegalArgumentException);
                assertTrue(throwable.getMessage().contains("scanAnnotatedClasses"));
            }); // GREEN, which is wrong! There is no ArithmeticException!
```
Full source code:
```
public class OptaPlannerProcessorIllegalXMLTest {

    @RegisterExtension
    static final QuarkusUnitTest config = new QuarkusUnitTest()
            .setExpectedException(IllegalArgumentException.class)
            .overrideConfigKey("quarkus.optaplanner.solver-config-xml",
                    "io/quarkus/optaplanner/illegalScanAnnotatedSolverConfig.xml")
            .setArchiveProducer(() -> ShrinkWrap.create(JavaArchive.class)
                    .addAsResource("io/quarkus/optaplanner/illegalScanAnnotatedSolverConfig.xml"))
            .setExpectedException(ArithmeticException.class) // Doesn't happen
            .assertException(throwable -> {
                assertTrue(throwable instanceof IllegalArgumentException);
                assertTrue(throwable.getMessage().contains("scanAnnotatedClasses"));
            });

    @Test
    public void scanAnnotatedClasses() {
        // Should not be called, deployment exception should happen first
        Assertions.fail();
    }
}

```

Proposal A) Check both (do not shadow)
Proposal B) Fail fast if both are used together.

geoand:
And now you contribute a fix to this :)

ge0ffrey:
You want proposal A) or proposal B) to be implemented?

geoand:
[...]

PR #7119, opened 2020-02-10
Fix "QuarkusUnitTest: assertException shadows out setExpectedException"

ge0ffrey:
Fix for https://github.com/quarkusio/quarkus/issues/7000

### DIFF
--- a/test-framework/junit5-internal/src/main/java/io/quarkus/test/QuarkusUnitTest.java
+++ b/test-framework/junit5-internal/src/main/java/io/quarkus/test/QuarkusUnitTest.java
@@ -125,6 +125,10 @@ public class QuarkusUnitTest
     }
 
     public QuarkusUnitTest assertException(Consumer<Throwable> assertException) {
+        if (this.assertException != null) {
+            throw new IllegalStateException("Don't set the asserted or excepted exception twice"
+                    + " to avoid shadowing out the first call.");
+        }
         this.assertException = assertException;
         return this;
     }


geoand (commented):
Shouldn't we also the same kind of check in `setExpectedException`? Also, perhaps the error message could be a little more verbose to mention that these two methods can only be called once?

ge0ffrey:
@geoand setExpectedException calls this method, so it fails-fast in all cases (twice setExpectedException, twice assertException and once of each)

As for exception message, what did you have in mind? I brainstormed a bit, and this message is the best I came up with.

ge0ffrey:
I can add "to avoid shadowing out the first call" in the error message. Wdyt?

geoand:
> @geoand setExpectedException calls this method, so it fails-fast in all cases (twice setExpectedException, twice assertException and once of each)
> 

Good point

> I can add "to avoid shadowing out the first call" in the error message. Wdyt?

Sure, sounds good

test-framework/junit5-internal/src/main/java/io/quarkus/test/QuarkusUnitTest.java
```diff
@@ -125,6 +125,9 @@ private QuarkusUnitTest(boolean useSecureConnection) {
     }
 
     public QuarkusUnitTest assertException(Consumer<Throwable> assertException) {
+        if (this.assertException != null) {
+            throw new IllegalStateException("Don't set the asserted or excepted exception twice.");
```
ge0ffrey:
```suggestion
            throw new IllegalStateException("Don't set the asserted or excepted exception twice"
                    + " to avoid shadowing out the first call.");
```

ge0ffrey:
Done through github interface (small change). Wait for CI to confirm it compiles :)

geoand:
@ge0ffrey the problem with the UI is that it created a second commit and we really want to have a single commit for this :)

gsmet:
Please squash both commits into one.

ge0ffrey:
@gsmet Done

geoand (approved):
Thanks!

PR #7128, opened 2020-02-11
Shutdown all resources on close

stuartwdouglas:


stuartwdouglas:
This may fix the windows issues

gastaldi (approved):

PR #7136, opened 2020-02-11
guides: fix native builder when deploying to OCP

someth2say:
Guides: Quarkus - Deploying on OpenShift with S2I
Despite fixed in other places, this guide didn't move to java8/java11 builder images as pushed to quay.io.


gastaldi (approved):

gastaldi:
Good catch!

cescoffier:
Thanks! Great catch! 

Hopefully, at some point in the future, we will be able to go back to a single image.

PR #7138, opened 2020-02-11
Gather and expose information about all known scopes before transforming annotations

### LINKED ISSUE #7076: ArC - introduce a new build item that represents a scope of a custom context registered by an extension
Some extension register an annotation transformer that adds a scope annotation to a class that is annotated with no scope annotation but meets some specific requirements, e.g. a class with a `@Scheduled` business methods is `@Singleton` by default: https://github.com/quarkusio/quarkus/blob/master/extensions/scheduler/deployment/src/main/java/io/quarkus/scheduler/deployment/SchedulerProcessor.java#L106

However, when an annotation transformer is registered the final set of contexts is not known. An extension that registers a `ContextRegistrar` should also produce a build item that holds a name of the scope that will be used by the custom context, so that annotation transformers would be able to skip classes annotated with custom scope annotations.

The name could be `CustomScopeBuildItem`.

manovotn:
…ing annotations.

Fixes #7076 

Doing this I realized that we already expose [`SCOPES`](https://github.com/quarkusio/quarkus/blob/master/independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/BuildExtension.java#L59) within build extensions. However, this is of type `ScopeInfo` (whereas here we want `DotName`) and is created too late for annotation transformers to use that. So, I've created another `Key` that represents scopes as `Collection<DotName>`. That way we don't break backward compatibility with any usage of `SCOPES`.

However, we still break compatibility with any existing custom scope definitions since `ContextRegistrarBuildItem` has an extra method. While I've fixed that for our own scopes (session and transactional), I highly doubt there are any other user defined custom scopes at this point as this feature was sparsely used even in pure CDI. And if they are, the fix is trivial.

Let me know what you think @mkouba and I can adjust the PR accordingly.

### DIFF
--- a/extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/ContextRegistrarBuildItem.java
+++ b/extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/ContextRegistrarBuildItem.java
@@ -1,20 +1,56 @@
 package io.quarkus.arc.deployment;
 
+import java.lang.annotation.Annotation;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+
+import org.jboss.jandex.DotName;
+import org.jboss.logging.Logger;
+
 import io.quarkus.arc.processor.ContextRegistrar;
 import io.quarkus.builder.item.MultiBuildItem;
 
 /**
- * Make it possible to register a custom CDI context.
+ * Make it possible to register one or more custom CDI contexts.
+ * If you are registering a new context, you also pass in the respective annotation value into the constructor either in
+ * {@link DotName} form, or as {@code Class<? extends Annotation>}.
+ *
+ * This information is then leveraged in {@link CustomScopeAnnotationsBuildItem} which allows consumers to browse
+ * all known custom scoped within deployment even early in the build process.
  */
 public final class ContextRegistrarBuildItem extends MultiBuildItem {
 
-    private final ContextRegistrar contextRegistrar;
+    private static final Logger LOGGER = Logger.getLogger(ContextRegistrarBuildItem.class);
 
-    public ContextRegistrarBuildItem(ContextRegistrar contextRegistrar) {
+    private final ContextRegistrar contextRegistrar;
+    private final Collection<DotName> annotationNames;
+
+    public ContextRegistrarBuildItem(ContextRegistrar contextRegistrar, DotName... annotationsNames) {
         this.contextRegistrar = contextRegistrar;
+        if (annotationsNames.length == 0) {
+            // log info level - usually you want to pass in annotation name as well
+            LOGGER.infof("A ContextRegistrarBuildItem was created but no annotation name/class was specified." +
+                    "This information can be later on consumed by other extensions via CustomScopeAnnotationsBuildItem, " +
+                    "please consider adding it.");
+        }
+        Collection<DotName> names = new ArrayList<>(annotationsNames.length);
+        for (DotName name : annotationsNames) {
+            names.add(name);
+        }
+        this.annotationNames = names;
+    }
+
+    public ContextRegistrarBuildItem(ContextRegistrar contextRegistrar, Class<? extends Annotation>... annotationsClasses) {
+        this(contextRegistrar, Arrays.stream(annotationsClasses).map(Class::getName)
+                .map(DotName::createSimple).toArray(DotName[]::new));
     }
 
     public ContextRegistrar getContextRegistrar() {
         return contextRegistrar;
+    }
+
+    public Collection<DotName> getAnnotationNames() {
+        return annotationNames;
     }
 }
--- a/extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/CustomScopeAnnotationsBuildItem.java
+++ b/extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/CustomScopeAnnotationsBuildItem.java
@@ -1,0 +1,61 @@
+package io.quarkus.arc.deployment;
+
+import java.util.Collection;
+
+import org.jboss.jandex.ClassInfo;
+import org.jboss.jandex.DotName;
+
+import io.quarkus.arc.processor.BuiltinScope;
+import io.quarkus.builder.item.SimpleBuildItem;
+
+/**
+ * Holds information about all known custom scopes in the deployment and has utility methods allowing to check
+ * whether given class has some scope annotation.
+ */
+public final class CustomScopeAnnotationsBuildItem extends SimpleBuildItem {
+
+    private Collection<DotName> customScopeNames;
+
+    public CustomScopeAnnotationsBuildItem(Collection<DotName> customScopeNames) {
+        this.customScopeNames = customScopeNames;
+    }
+
+    /**
+     * Returns a collection of all known custom scopes represented as {@link DotName}.
+     *
+     * @return collection of known custom scopes (built-in scopes are not included)
+     */
+    public Collection<DotName> getCustomScopeNames() {
+        return customScopeNames;
+    }
+
+    /**
+     * Returns true if the given class has some of the custom scope annotations, false otherwise.
+     * List of known custom scopes can be seen via {@link CustomScopeAnnotationsBuildItem#getCustomScopeNames()}.
+     * In order to check for presence of any scope annotation (including built-in ones),
+     * see {@link CustomScopeAnnotationsBuildItem#isScopeDeclaredOn(ClassInfo)}.
+     *
+     * @param clazz Class to check for annotations
+     * @return true if the clazz contains some of the custom scope annotations, false otherwise
+     */
+    public boolean isCustomScopeDeclaredOn(ClassInfo clazz) {
+        for (DotName scope : customScopeNames) {
+            if (clazz.classAnnotation(scope) != null) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    /**
+     * Returns true if the given class has some scope annotations, false otherwise.
+     * This method check for all scope annotations, including built-in ones as well as custom scopes.
+     * List of known custom scopes can be seen via {@link CustomScopeAnnotationsBuildItem#getCustomScopeNames()}.
+     *
+     * @param clazz Class to check for annotations
+     * @return true if the clazz contains any scope annotation, false otherwise
+     */
+    public boolean isScopeDeclaredOn(ClassInfo clazz) {
+        return BuiltinScope.isDeclaredOn(clazz) || isCustomScopeDeclaredOn(clazz);
+    }
+}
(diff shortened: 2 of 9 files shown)


extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/ContextRegistrarBuildItem.java
```diff
@@ -9,12 +11,18 @@
 public final class ContextRegistrarBuildItem extends MultiBuildItem {
 
     private final ContextRegistrar contextRegistrar;
+    private final DotName annotationName;
 
-    public ContextRegistrarBuildItem(ContextRegistrar contextRegistrar) {
+    public ContextRegistrarBuildItem(ContextRegistrar contextRegistrar, DotName annotationName) {
```
mkouba: I'm not so sure it was a good idea to add the annotation name to the `ContextRegistrarBuildItem`. In theory, you can use one `ContextRegistrarBuildItem` to register multiple contexts. It might make more sense to create a separate build item or to make it `DotName[]`...

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/StartupBuildSteps.java
```diff
@@ -74,6 +76,15 @@ public void transform(TransformationContext context) {
         });
     }
 
+    public static boolean isDeclaredOn(Collection<DotName> scopes, ClassInfo clazz) {
```
mkouba: We should probably extract this method in a public util class.

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/ContextRegistrarBuildItem.java
mkouba: Also it would make sense to pass `Class<? extends Annotation>` and create a `DotName`  under the hood. Maybe we should create a a builder to avoid "too many constructors hell".

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/StartupBuildSteps.java
```diff
@@ -62,7 +64,7 @@ public boolean appliesTo(org.jboss.jandex.AnnotationTarget.Kind kind) {
 
             @Override
             public void transform(TransformationContext context) {
-                if (context.isClass() && !BuiltinScope.isDeclaredOn(context.getTarget().asClass())) {
+                if (context.isClass() && !isDeclaredOn(context.get(Key.SCOPES_NAMES), context.getTarget().asClass())) {
```
mkouba: I can see a little inconsistency here - build steps can simply inject the list of `ContextRegistrarBuildItem`, ie. `Key.SCOPES_NAMES` is not needed. However, regular ArC extensions have no way to register custom annotations and so `Key.SCOPES_NAMES` is not very useful outside quarkus.

mkouba (changes requested):

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/ContextRegistrarBuildItem.java
manovotn: In the original issue we discussed that it would be nicer to have just one build item. IMO it's better to have one item for one task whenever possible (task being custom context registration). Otherwise you risk having a context registered but not the name of the scope.

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/StartupBuildSteps.java
manovotn: Note that `KEY.SCOPE_NAMES` is all scopes that ArC knows, not just custom one. That allows any Quarkus-ArC extension to check the list of all known scopes at (almost) any point and using the same way that you would use to grab annotation store, injection points, beans.... Without it, you need to work with a combination of built in scopes and context registrar items. What I attempted here just seemed a lot more user friendly to me.

manovotn:
So after some chat with @mkouba on Zulip we decided to:

* Change `ContextRegistrarBuildItem`, introduce two constructors that will accept either `DotName` or `Class<? extends Annotation>` - this will still be a breaking change, much like what I did previously
* Introduce a new single build item (`CustomScopeAnnotationsBuildItem`) that will grant access to util method `isDeclaredOn()` + `isScopeAnnotationDeclaredOn()` (including built in ones)
* Arc processor will create `CustomScopeAnnotationsBuildItem` in a separate build stem from all the `ContextRegistrarBuildItem` in the deployment
* The above implies we can drop `SCOPE_NAMES` altogether

manovotn:
@mkouba I changed the PR accordingly, WDYT about it now?
(the CI is unfinished because Guillaume killed the job, probably needing the machines for something else)

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/ContextRegistrarBuildItem.java
```diff
@@ -1,20 +1,56 @@
...
+        }
+        Collection<DotName> names = new ArrayList<>(annotationsNames.length);
+        for (DotName name : annotationsNames) {
+            names.add(name);
+        }
+        this.annotationNames = names;
+    }
+
+    public ContextRegistrarBuildItem(ContextRegistrar contextRegistrar, Class<? extends Annotation>... annotationsClasses) {
+        this.contextRegistrar = contextRegistrar;
```
mkouba: You could call the `ContextRegistrarBuildItem(ContextRegistrar contextRegistrar, DotName... annotationsNames)` constructor instead... something like:

```java
this(contextRegistrar, annotationsClasses.stream().map(Class::getName).map(DotName::createSimple).toArray(DotName[]::new));
```

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/CustomScopeAnnotationsBuildItem.java
```diff
@@ -0,0 +1,61 @@
...
+    public CustomScopeAnnotationsBuildItem(Collection<DotName> customScopeNames) {
+        this.customScopeNames = customScopeNames;
+    }
+
+    /**
+     * Returns a collection of all known custom scopes represented as {@link DotName}.
+     *
+     * @return collection of known custom scopes (built-in scopes are not included)
+     */
+    public Collection<DotName> getAllCustomScopeNames() {
```
mkouba: I'd probably use the javabean convention here, ie. `getCustomScopeNames()`.

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/CustomScopeAnnotationsBuildItem.java
```diff
@@ -0,0 +1,61 @@
...
+    /**
+     * Returns true if the given class has some scope annotations, false otherwise.
+     * This method check for all scope annotations, including built-in ones as well as custom scopes.
+     * List of known custom scopes can be seen via {@link CustomScopeAnnotationsBuildItem#getAllCustomScopeNames()}.
+     *
+     * @param clazz Class to check for annotations
+     * @return true if the clazz contains any scope annotation, false otherwise
+     */
+    public boolean isScopeAnnotationDeclaredOn(ClassInfo clazz) {
+        return BuiltinScope.isDeclaredOn(clazz) ? true : isDeclaredOn(clazz);
```
mkouba: `return BuiltinScope.isDeclaredOn(clazz) || isDeclaredOn(clazz)`? ;-)

mkouba:
Looks good! We should also find all occurences of `io.quarkus.arc.processor.BuiltinScope.isDeclaredOn(ClassInfo)` and update where needed...

manovotn:
> Looks good! We should also find all occurences of `io.quarkus.arc.processor.BuiltinScope.isDeclaredOn(ClassInfo)` and update where needed...

That's my plan as soon as we settle upon a solution ;-)

manovotn:
@mkouba rebased and updated - replaced usages of `BuiltInScope` with the new build item. WDYT?

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/ContextRegistrarBuildItem.java
```diff
@@ -1,20 +1,50 @@
...
 public final class ContextRegistrarBuildItem extends MultiBuildItem {
 
     private final ContextRegistrar contextRegistrar;
+    private final Collection<DotName> annotationNames;
 
-    public ContextRegistrarBuildItem(ContextRegistrar contextRegistrar) {
+    public ContextRegistrarBuildItem(ContextRegistrar contextRegistrar, DotName... annotationsNames) {
         this.contextRegistrar = contextRegistrar;
+        if (annotationsNames == null) {
+            throw new IllegalArgumentException("Parameter annotationNames cannot be null!");
```
mkouba: Hm, so if I pass an empty array it's OK? Also if you skip the `annotationsNames` completely on the call site the method receives an empty array. It could be compiler-specific though. This is what I get for oracle JDK 8.

Anyway, I'm not quite sure whether throwing an IAE is the best approach. I'd rather just log a warning if a `ContextRegistrarBuildItem` has no annotations. In theory, the `ContextRegistrar` can register a context conditionally... I know that it's ugly :-(.

mkouba (commented):
Looks really good! I've added few comments ;-).

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/ContextRegistrarBuildItem.java
mkouba: https://stackoverflow.com/questions/10905942/which-part-of-the-java-language-specification-describes-the-behaviour-of-omitted

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/ContextRegistrarBuildItem.java
```diff
@@ -1,20 +1,50 @@
...
+
+import org.jboss.jandex.DotName;
+
 import io.quarkus.arc.processor.ContextRegistrar;
 import io.quarkus.builder.item.MultiBuildItem;
 
 /**
- * Make it possible to register a custom CDI context.
+ * Make it possible to register one or more custom CDI contexts.
+ * If you are registering a new context, you should also pass in the annotation value into the constructor either in
```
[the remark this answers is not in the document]
manovotn: Well, yes but we don't know how many contexts you register and how many annotation names you pass it - e.g. you may register 3 contexts and pass in one annotation name.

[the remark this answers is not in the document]
manovotn: But I'll twist the wording a little.

manovotn:
@mkouba updated with all the changes

mkouba (approved):

PR #7141, opened 2020-02-11
exclude spotbugs and findbugs

michalszynkiewicz:


gsmet (changes requested):
Please add them to the banned dependencies in the build-parent pom. Thanks!

michalszynkiewicz:
I added it and it turns out the enforcer fails on test dependencies :(
I will exclude the test ones too

gsmet:
You can specifically allow things for test dependencies, there is an example at the bottom of the banned dependencies IIRC.

michalszynkiewicz:
should be excluded everywhere now :)

michalszynkiewicz:
So it seems that (unlike plain java), in some cases Quarkus requires declared annotations to be available. 
An example is `SpringDIProcessor` that assumes such annotations are in the index.
I will close this PR now. We can revisit later.

PR #7142, opened 2020-02-11
Ensure that user provided ObjectMapperCustomizer objects always have higher priority

### LINKED ISSUE #7116: Custom serializer in ObjectMapperCustomizer is loaded randomly
**Describe the bug**
A custom serializer registered in Jackson's ObjectMapper through ObjectMapperCustomizer is loaded randomly. Sometimes it works, sometimes it doesn't after a restart or during tests.

**Expected behavior**
Data should always be serialized the same way. The ZonedDateTime field in the below example should always look like this : 2020-02-10T10:39:22Z

**Actual behavior**
The ZonedDateTime field is sometimes serialized like this after a restart : 
2020-02-10T11:41:21.854544+01:00

**To Reproduce**
1. I created a sample app which reproduces the issue : https://github.com/jessydeclerck/quarkus-serialization-issue
2. run the app : mvn compile quarkus:dev
3. get request on the following endpoint : http://localhost:8080/hello
4. Restart app and reproduce step 3 until date format has changed 

 - Output of `uname -a` or `ver`: Darwin 19.2.0 Darwin Kernel Version 19.2.0: Sat Nov  9 03:47:04 PST 2019; root:xnu-6153.61.1~20/RELEASE_X86_64 x86_64
 - Output of `java -version`: 
 - Quarkus version or git rev: 1.3.0.Alpha1 (issue is also there in 1.2.0.Final)

**Additional context :**
ObjectMapperCustomizer can be found here : [DateFormatConfig](https://github.com/jessydeclerck/quarkus-serialization-issue/blob/master/src/main/java/org/acme/config/DateFormatConfig.java)

geoand:
Fixes: #7116

### DIFF
--- a/extensions/jackson/runtime/src/main/java/io/quarkus/jackson/ObjectMapperProducer.java
+++ b/extensions/jackson/runtime/src/main/java/io/quarkus/jackson/ObjectMapperProducer.java
@@ -1,5 +1,9 @@
 package io.quarkus.jackson;
 
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
 import javax.enterprise.context.ApplicationScoped;
 import javax.enterprise.inject.Instance;
 import javax.enterprise.inject.Produces;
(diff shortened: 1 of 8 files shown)


gsmet (approved):

extensions/jackson/runtime/src/main/java/io/quarkus/jackson/ObjectMapperProducer.java
```diff
@@ -17,9 +21,20 @@
...
         ObjectMapper objectMapper = new ObjectMapper();
-        for (ObjectMapperCustomizer customizer : customizers) {
+        List<ObjectMapperCustomizer> sortedCustomizers = sortCustomizersInDescendingPriorityOrder(customizers);
+        for (ObjectMapperCustomizer customizer : sortedCustomizers) {
             customizer.customize(objectMapper);
         }
         return objectMapper;
     }
+
+    private List<ObjectMapperCustomizer> sortCustomizersInDescendingPriorityOrder(
```
gastaldi: That looks odd, is inverting the priority also done somewhere else?

geoand: Higher priority means that the module should take precedence in case of the same module has been applied in a lower priority customizer. 
Same as where higher priority in MP Config means that the Config Source take precedence

PR #7143, opened 2020-02-11
Undertow: unused servlet initParams code, adds tests

Karm:
Two occurrences of

```
- Map<String, String> params = new HashMap<>();
- for (Map.Entry<String, String> i : params.entrySet()) {
-     params.put(i.getKey(), i.getValue());
- }
```

are removed from UndertowBuildStep.

New tests covering init params and descriptions for Servlet and Filter were added, both for xml and annotation.

ping @stuartwdouglas

### DIFF
--- a/extensions/undertow/deployment/src/main/java/io/quarkus/undertow/deployment/UndertowBuildStep.java
+++ b/extensions/undertow/deployment/src/main/java/io/quarkus/undertow/deployment/UndertowBuildStep.java
@@ -12,10 +12,8 @@ import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
-import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
-import java.util.Map;
 import java.util.Set;
 import java.util.function.Consumer;
 import java.util.stream.Collectors;
@@ -329,16 +327,11 @@ public class UndertowBuildStep {
         if (webMetaData.getServlets() != null) {
             for (ServletMetaData servlet : webMetaData.getServlets()) {
                 reflectiveClasses.accept(new ReflectiveClassBuildItem(false, false, servlet.getServletClass()));
-                Map<String, String> params = new HashMap<>();
-                for (Map.Entry<String, String> i : params.entrySet()) {
-                    params.put(i.getKey(), i.getValue());
-                }
                 RuntimeValue<ServletInfo> sref = recorder.registerServlet(deployment, servlet.getServletName(),
                         context.classProxy(servlet.getServletClass()),
                         servlet.isAsyncSupported(),
                         servlet.getLoadOnStartupInt(),
                         bc.getValue(),
-                        params,
                         null);
                 if (servlet.getInitParam() != null) {
                     for (ParamValueMetaData init : servlet.getInitParam()) {
@@ -402,16 +395,12 @@ public class UndertowBuildStep {
         if (webMetaData.getFilters() != null) {
             for (FilterMetaData filter : webMetaData.getFilters()) {
                 reflectiveClasses.accept(new ReflectiveClassBuildItem(false, false, filter.getFilterClass()));
-                Map<String, String> params = new HashMap<>();
-                for (Map.Entry<String, String> i : params.entrySet()) {
-                    params.put(i.getKey(), i.getValue());
-                }
                 RuntimeValue<FilterInfo> sref = recorder.registerFilter(deployment,
                         filter.getFilterName(),
                         context.classProxy(filter.getFilterClass()),
                         filter.isAsyncSupported(),
                         bc.getValue(),
-                        params, null);
+                        null);
                 if (filter.getInitParam() != null) {
                     for (ParamValueMetaData init : filter.getInitParam()) {
                         recorder.addFilterInitParam(sref, init.getParamName(), init.getParamValue());
@@ -487,7 +476,7 @@ public class UndertowBuildStep {
                 reflectiveClasses.accept(new ReflectiveClassBuildItem(false, false, servlet.getServletClass()));
             }
             recorder.registerServlet(deployment, servlet.getName(), context.classProxy(servletClass),
-                    servlet.isAsyncSupported(), servlet.getLoadOnStartup(), bc.getValue(), servlet.getInitParams(),
+                    servlet.isAsyncSupported(), servlet.getLoadOnStartup(), bc.getValue(),
                     servlet.getInstanceFactory());
 
             for (String m : servlet.getMappings()) {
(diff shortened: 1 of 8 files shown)


extensions/undertow/deployment/src/main/java/io/quarkus/undertow/deployment/UndertowBuildStep.java
```diff
@@ -329,16 +328,12 @@ public ServletDeploymentManagerBuildItem build(List<ServletBuildItem> servlets,
         if (webMetaData.getServlets() != null) {
             for (ServletMetaData servlet : webMetaData.getServlets()) {
                 reflectiveClasses.accept(new ReflectiveClassBuildItem(false, false, servlet.getServletClass()));
-                Map<String, String> params = new HashMap<>();
-                for (Map.Entry<String, String> i : params.entrySet()) {
```
gastaldi: Actually I think it should be be:

```java
for (ParamValueMetaData pvm : servlet.getInitParam()) {
   params.put(pvm.getParamName(), pvm.getParamValue());
}
```

extensions/undertow/deployment/src/main/java/io/quarkus/undertow/deployment/UndertowBuildStep.java
```diff
@@ -402,16 +397,13 @@ public ServletDeploymentManagerBuildItem build(List<ServletBuildItem> servlets,
         if (webMetaData.getFilters() != null) {
             for (FilterMetaData filter : webMetaData.getFilters()) {
                 reflectiveClasses.accept(new ReflectiveClassBuildItem(false, false, filter.getFilterClass()));
-                Map<String, String> params = new HashMap<>();
```
gastaldi: This should be:

```java
for (ParamValueMetaData pvm : filter.getInitParam()) {
   params.put(pvm.getParamName(), pvm.getParamValue());
}
```

extensions/undertow/deployment/src/main/java/io/quarkus/undertow/deployment/UndertowBuildStep.java
Karm: @gastaldi The thing is I tried exactly that without any impact :)
That is why I went ahead with the test, to reach out to that part of code.
Do you happen to have any hint on the test case?
No point in having here something that actually doesn't get used.

I'll check it again though...

gastaldi: > Do you happen to have any hint on the test case?

I think your test case looks correct. You mean that the init parameters are not set in the `ServletMetaData` object?

stuartwdouglas: This parameter should just be deleted. The init params are added in the next line with a seperate call to the recorder.

Karm:
@gastaldi  @stuartwdouglas  Removed the superfluous code. Tests keep passing for me.

extensions/undertow/runtime/src/main/java/io/quarkus/undertow/runtime/UndertowDeploymentRecorder.java
```diff
@@ -221,16 +220,13 @@ public static SocketAddress getHttpAddress() {
             Class<?> servletClass,
             boolean asyncSupported,
             int loadOnStartup,
-            BeanContainer beanContainer, Map<String, String> initParams,
+            BeanContainer beanContainer,
```
gastaldi: I am not sure it's a good idea to change the API here. Other extensions (external to core - camel?) may be using it

extensions/undertow/deployment/src/main/java/io/quarkus/undertow/deployment/UndertowBuildStep.java
```diff
@@ -499,7 +488,7 @@ public ServletDeploymentManagerBuildItem build(List<ServletBuildItem> servlets,
             String filterClass = filter.getFilterClass();
             reflectiveClasses.accept(new ReflectiveClassBuildItem(false, false, filterClass));
             recorder.registerFilter(deployment, filter.getName(), context.classProxy(filterClass), filter.isAsyncSupported(),
-                    bc.getValue(), filter.getInitParams(), filter.getInstanceFactory());
+                    bc.getValue(), filter.getInstanceFactory());
```
gastaldi: Where are these init params used?

gastaldi (changes requested):
Added some comments

extensions/undertow/deployment/src/main/java/io/quarkus/undertow/deployment/UndertowBuildStep.java
Karm: @gastaldi 
Well, they were redundantly used in in ```io.quarkus.undertow.runtime.UndertowDeploymentRecorder#registerFilter```.
The real place where initParams are set for filters is this: [UndertowBuildStep.java#L406](https://github.com/quarkusio/quarkus/blob/af0251cfff43b386814f5a44285dd515810996a8/extensions/undertow/deployment/src/main/java/io/quarkus/undertow/deployment/UndertowBuildStep.java#L406).

extensions/undertow/runtime/src/main/java/io/quarkus/undertow/runtime/UndertowDeploymentRecorder.java
Karm: @gastaldi Hmm, such external extension would have to be fiddling with how Servlet is built in Quarkus *and* use Undertow extension for that...

If you have any suggestion what area such extension might appear in, I will try to find it and run its tests...

gastaldi: Check if camel-quarkus is using it, otherwise it may be good to go then

Karm: @gastaldi I run quarkus-camel tests, excluding Mongo (does not build with Quarkus master, talked to @ppalaga ) and JMS (local network shenanigans), and it looks unaffected, e.g. org.apache.camel.quarkus.component.servlet.CamelServletTest is :heavy_check_mark:  etc.

+ basic search for usage of said API did not show anything...

IMHO safe to go.

gastaldi (approved):

stuartwdouglas:
Recorders are not part of an extensions API, you should not be using recorder classes outside their extension. Unfortunately we can't really make them package private as the build steps are in a different package.

gastaldi:
Sounds like something to be considered while improving the Extensions API

Em Qua, 12 de fev de 2020 20:10, Stuart Douglas <notifications@github.com>
escreveu:

PR #7147, opened 2020-02-11
MongoDB with Panache multi-update

loicmathieu:
Fixes 5924

@FroMage @viniciusfcf this is the implementation of Panache multi-update for MongoDB.

I took a slightly different aproach as for MongoDB, an udpate is not a single query but two: a filter (the query part) and an update document. So the method signature is not the same as with Hibernate (but parameters and PanacheQL works for both parts).

I just regret that we didn't call these methods `updateAll` in the first time as it seems more logical for me (and reading the JavaDoc ...), but it's too late to change this for Hibernate with Panache so better keep them this way also in MongoDB with Panache.

### DIFF
--- a/docs/src/main/asciidoc/hibernate-orm-panache.adoc
+++ b/docs/src/main/asciidoc/hibernate-orm-panache.adoc
@@ -198,8 +198,8 @@ Person.deleteAll();
 // delete by id
 boolean deleted = Person.deleteById(personId);
 
-// update all living persons
-Person.update("name = 'Moral' where status = ?1", Status.Alive);
+// set the name of all living persons to 'Mortal'
+Person.update("name = 'Mortal' where status = ?1", Status.Alive);
 
 ----
 
@@ -385,8 +385,8 @@ personRepository.deleteAll();
 // delete by id
 boolean deleted = personRepository.deleteById(personId);
 
-// update all living persons
-personRepository.update("name = 'Moral' where status = ?1", Status.Alive);
+// set the name of all living persons to 'Mortal'
+personRepository.update("name = 'Mortal' where status = ?1", Status.Alive);
 
 ----
 
@@ -515,7 +515,7 @@ link:https://docs.jboss.org/hibernate/orm/5.4/userguide/html_single/Hibernate_Us
 [source,java]
 ----
 Order.find("select distinct o from Order o left join fetch o.lineItems");
-Order.update("update from Person set name = 'Moral' where status = ?", Status.Alive);
+Order.update("update from Person set name = 'Mortal' where status = ?", Status.Alive);
 ----
 
 === Query parameters
--- a/docs/src/main/asciidoc/mongodb-panache.adoc
+++ b/docs/src/main/asciidoc/mongodb-panache.adoc
@@ -225,6 +225,9 @@ Person.deleteAll();
 
 // delete by id
 boolean deleted = Person.deleteById(personId);
+
+// set the name of all living persons to 'Mortal'
+long updated = Person.update("name", "Mortal").where("status", Status.Alive);
 ----
 
 All `list` methods have equivalent `stream` versions.
@@ -388,6 +391,9 @@ personRepository.deleteAll();
 
 // delete by id
 boolean deleted = personRepository.deleteById(personId);
+
+// set the name of all living persons to 'Mortal'
+long updated = personRepository.update("name", "Mortal").where("status", Status.Alive);
 ----
 
 All `list` methods have equivalent `stream` versions.
@@ -495,7 +501,7 @@ MongoDB with Panache will then map it to a MongoDB native query.
 
 If your query does not start with `{`, we will consider it a PanacheQL query:
 
-- `<singlePropertyName>` (and single parameter) which will expand to `{'singleColumnName': '?'}`
+- `<singlePropertyName>` (and single parameter) which will expand to `{'singleColumnName': '?1'}`
 - `<query>` will expand to `{<query>}` where we will map the PanacheQL query to MongoDB native query form. We support the following operators that will be mapped to the corresponding MongoDB operators: 'and', 'or' ( mixing 'and' and 'or' is not currently supported), '=', '>', '>=', '<', '<=', '!=', 'is null', 'is not null', and 'like' that is mapped to the MongoDB `$regex` operator.
 
 Here are some query examples:
@@ -510,6 +516,17 @@ link:https://docs.mongodb.com/manual/reference/bson-types/#document-bson-type-da
 The MongoDB POJO codec doesn't support `ZonedDateTime` and `OffsetDateTime` so you should convert them prior usage.
 
 MongoDB with Panache also supports extended MongoDB queries by providing a `Document` query, this is supported by the find/list/stream/count/delete methods.
+
+MongoDB with Panache offers operations to update multiple documents based on an update document and a query :
+`Person.update("foo = ?1, bar = ?2", fooName, barName).where("name = ?1", name)`.
+
+For these operations, you can express the update document the same way you express your queries, here are some examples:
+
+- `<singlePropertyName>` (and single parameter) which will expand to the update document `{'$set' : {'singleColumnName': '?1'}}`
+- `firstname = ?1, status = ?2` will be mapped to the update document `{'$set' : {'firstname': ?1, 'status': ?2}}`
+- `firstname = :firstname, status = :status` will be mapped to the update document `{'$set' : {'firstname': :firstname, 'status': :status}}`
+- `{'firstname' : ?1, 'status' : ?2}` will be mapped to the update document `{'$set' : {'firstname': ?1, 'status': ?2}}`
+- `{'firstname' : firstname, 'status' : :status}` ` will be mapped to the update document `{'$set' : {'firstname': :firstname, 'status': :status}}`
 
 === Query parameters
 
@@ -809,6 +826,9 @@ deleteCount = ReactivePerson.deleteAll();
 
 // delete by id
 Uni<Boolean> deleted = ReactivePerson.deleteById(personId);
+
+// set the name of all living persons to 'Mortal'
+Uni<Long> updated = ReactivePerson.update("name", "Mortal").where("status", Status.Alive);
 ----
 
 TIP: If you use MongoDB with Panache in conjunction with RESTEasy, you can directly return a reactive type inside your JAX-RS resource endpoint as long as you include the `quarkus-resteasy-mutiny` extension.
--- a/extensions/panache/mongodb-panache/runtime/src/main/java/io/quarkus/mongodb/panache/PanacheMongoEntityBase.java
+++ b/extensions/panache/mongodb-panache/runtime/src/main/java/io/quarkus/mongodb/panache/PanacheMongoEntityBase.java
@@ -794,7 +794,7 @@ public abstract class PanacheMongoEntityBase {
     /**
      * Insert all given entities.
      *
-     * @param entities the entities to update
+     * @param entities the entities to insert
      * @see #persist()
      * @see #persist(Stream)
      * @see #persist(Iterable)
@@ -873,6 +873,55 @@ public abstract class PanacheMongoEntityBase {
      */
     public static void persistOrUpdate(Object firstEntity, Object... entities) {
         MongoOperations.persistOrUpdate(firstEntity, entities);
+    }
+
+    /**
+     * Update all entities of this type by the given update document, with optional indexed parameters.
+     * The returned {@link PanacheUpdate} object will allow to restrict on which document the update should be applied.
+     *
+     * @param update the update document, if it didn't contain <code>$set</code> we add it.
+     *        It can also be expressed as a {@link io.quarkus.mongodb.panache query string}.
+     * @param params optional sequence of indexed parameters
+     * @return a new {@link PanacheUpdate} instance for the given update document
+     * @see #update(String, Map)
+     * @see #update(String, Parameters)
+     */
+    @GenerateBridge
+    public static PanacheUpdate update(String update, Object... params) {
+        throw MongoOperations.implementationInjectionMissing();
+    }
+
+    /**
+     * Update all entities of this type by the given update document, with named parameters.
+     * The returned {@link PanacheUpdate} object will allow to restrict on which document the update should be applied.
+     *
+     * @param update the update document, if it didn't contain <code>$set</code> we add it.
+     *        It can also be expressed as a {@link io.quarkus.mongodb.panache query string}.
+     * @param params {@link Map} of named parameters
+     * @return a new {@link PanacheUpdate} instance for the given update document
+     * @see #update(String, Object...)
+     * @see #update(String, Parameters)
+     *
+     */
+    @GenerateBridge
+    public static PanacheUpdate update(String update, Map<String, Object> params) {
+        throw MongoOperations.implementationInjectionMissing();
+    }
+
+    /**
+     * Update all entities of this type by the given update document, with named parameters.
+     * The returned {@link PanacheUpdate} object will allow to restrict on which document the update should be applied.
+     *
+     * @param update the update document, if it didn't contain <code>$set</code> we add it.
+     *        It can also be expressed as a {@link io.quarkus.mongodb.panache query string}.
+     * @param params {@link Parameters} of named parameters
+     * @return a new {@link PanacheUpdate} instance for the given update document
+     * @see #update(String, Object...)
+     * @see #update(String, Map)
+     */
+    @GenerateBridge
+    public static PanacheUpdate update(String update, Parameters params) {
+        throw MongoOperations.implementationInjectionMissing();
     }
 
     /**
--- a/extensions/panache/hibernate-orm-panache/runtime/src/main/java/io/quarkus/hibernate/orm/panache/PanacheEntityBase.java
+++ b/extensions/panache/hibernate-orm-panache/runtime/src/main/java/io/quarkus/hibernate/orm/panache/PanacheEntityBase.java
@@ -731,7 +731,7 @@ public abstract class PanacheEntityBase {
     }
 
     /**
-     * Update all entities of this type matching the given query, with mandatory indexed parameters.
+     * Update all entities of this type matching the given query, with optional indexed parameters.
      *
      * @param query a {@link io.quarkus.hibernate.orm.panache query string}
      * @param params optional sequence of indexed parameters
--- a/extensions/panache/mongodb-panache/runtime/src/main/java/io/quarkus/mongodb/panache/PanacheMongoRepositoryBase.java
+++ b/extensions/panache/mongodb-panache/runtime/src/main/java/io/quarkus/mongodb/panache/PanacheMongoRepositoryBase.java
@@ -881,6 +881,55 @@ public interface PanacheMongoRepositoryBase<Entity, Id> {
     }
 
     /**
+     * Update all entities of this type by the given update document, with optional indexed parameters.
+     * The returned {@link PanacheUpdate} object will allow to restrict on which documents the update should be applied.
+     *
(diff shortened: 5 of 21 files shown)


docs/src/main/asciidoc/mongodb-panache.adoc
```diff
@@ -378,6 +381,9 @@ personRepository.delete("status", Status.Alive);
 
 // delete all persons
 personRepository.deleteAll();
+
+// update all living persons
```
FroMage: This looks like you mixed up the query/update part, no? It's not updating all living persons, but all persons with a `Moral` name (does that even compile?) to make them `Alive`, no?

That kinda proves the API is confusing, no? It's not clear to me if the query or update should be first.

FroMage (commented):
LGTM. Pity that we can't have a single-query version like `Person.update("status = ?2 where name = ?1")` but if we do that we can't write native queries/updates.

If we wanted to support both `Person.update(String updateWithQuery)` and `Person.update(String query, String update)` it would likely conflict with the `Person.update(String updateWithQuery, Object... args)` version.

Perhaps there should be two versions? One for panaqueql and one for native? Like `Person.update(String updateWithQuery)` and `Person.updateNative(String query, String update)`?

Just an idea to see what you think.

docs/src/main/asciidoc/mongodb-panache.adoc
loicmathieu: Query is first, then the update document (query is a filter in MongoDB terms), what is confusing is that you can pass parameters on both side and it takes the parameters as if the query is a unique one ... Using varags parameters it is not very readable.

Maybe we need to introduce an `UpdateQuery` object that will allow to intoduce the two "query part" one by one
```
Person.update("name = ?1", "Mortal").with("status = ?1", Status.Alive);
```

WDYT ?

FroMage: I like that idea. That's more readable. It also supports shortcut syntax: `Person.update("name", "Mortal").with("status", Status.Alive);`

loicmathieu: OK, I'll give it a try

loicmathieu: The new impl is done, can you have a look ?

FroMage: The comment here (and above) are still wrong: this does not update all living persons. It sets all persons named "Mortal" to living.

extensions/panache/mongodb-panache/runtime/src/main/java/io/quarkus/mongodb/panache/PanacheMongoEntityBase.java
```diff
@@ -864,6 +864,49 @@ public static void persistOrUpdate(Object firstEntity, Object... entities) {
         MongoOperations.persistOrUpdate(firstEntity, entities);
     }
 
+    /**
+     * Update all entities of this type matching the given query, with mandatory indexed parameters.
```
FroMage: I'm not sure what the `mandatory` here means. It should be removed.

FroMage: Below (for repository) it's `optional`…

FroMage (changes requested):
LGTM, but you need to test the update with mongo query syntax as well.

docs/src/main/asciidoc/mongodb-panache.adoc
FroMage: @emmanuelbernard can you give this an eye? I think it looks good, but it's different to the ORM version. Given that queries can be native here, I think it's fine.

extensions/panache/mongodb-panache/runtime/src/main/java/io/quarkus/mongodb/panache/PanacheMongoEntityBase.java
loicmathieu: It's copied from the JavaDoc of Hibernate with Panache where `mandatory` is used. I think it's wrong as we should be able to use it without any parameters : `Person.update("set person.status = 'Dead' where person.status='Alive');`

I check the code of the Hibernate with Panache implementation and no parameters are allowed at code level, I'll update the documentation with `optional` here also.

docs/src/main/asciidoc/mongodb-panache.adoc
loicmathieu: > The comment here (and above) are still wrong: this does not update all living persons. It sets all persons named "Mortal" to living.

This was copied from the Hibernate with Panache documentation, I'll change it there also.

FroMage: Hah, I was afraid this would turn out to be the case ;) Thanks!

extensions/panache/mongodb-panache/runtime/src/main/java/io/quarkus/mongodb/panache/PanacheMongoEntityBase.java
FroMage: Yeah, I've no idea how this could have slipped through. Thanks for fixing it there too.

loicmathieu:
> LGTM, but you need to test the update with mongo query syntax as well.

FroMage:
> @FroMage MongoDB with Panache lack some test coverage on a lot of areas, I opened #7698 for it is it OK to wait for it to cover the missing tests ?

Sure, this is mostly tests for already-merged functionality.

loicmathieu:
@FromMage is this still some work to do for the PR ? If not can you approve it ?

FroMage:
I'd like to check with @emmanuelbernard first to make sure he's on board with this new API.

docs/src/main/asciidoc/mongodb-panache.adoc
emmanuelbernard: Are you asking me for my opinion on

`update("name", "Mortal").with("status", Status.Alive);`
being different than
`update("name = 'Mortal' where status = ?1", Status.Alive);`

Would be cool to be unified but I imagine there is a reason behind that difference?

What makes me uneasy is how can I do a update all name to "Mortal" without condition of being alive? And the counter reason is how is it clear that I am not changing all the DB by accident by forgetting the "where clause" - that could not have happened ever to one of us, right so this is purely hypothetical that an ecomerce site did update its catalog with price=0 for all products ;P

loicmathieu: @emmanuelbernard in fact the implementation is
 `update(filter, params).with(document, params)`

So 
`update("name", "Mortal").with("status", Status.Alive);`

Correspond to the following one implemented on Hibernate with Panache (but cannot be done easily with MongoDB)
`update("name = ?1 where status = ?2", "Mortal", Status.Alive);`

To update without filter you could use the empty String or null (worth testing it)
`update("").with("status", Status.Alive);`

We can also add an overload of update with no parameters for this case, good point :)

FroMage: > Would be cool to be unified but I imagine there is a reason behind that difference?

Yeah, and the reason is that both the query and the update need to support native queries so we can't have them as a single `String` param, and having two `String` params for the query and update parts is confusing (their order is).

Now, you're right that we do need another version with no query part. We can't go with `update("").with(...)` because it looks silly. We could do `update().with(...)` but it pains me that we're going through all that trouble for native queries, which in theory is the less common case.

At this point, we could have `update().with(update)` and `update(query).with(update)` OR `update(panacheQlQuery)` and `updateNative(nativeQuery?).with(updateQuery)`.

Not sure yet which I prefer.

loicmathieu: We choose to swith panacheql/nativeql by inspecting the query string for the presence of `{`. 
I would really prefere to keep it this way so I favor `update().with(update, params)`.

The issue is that a MongoDB update is two queries and not one (a filter query and an update document), and we need to provide parameters substitution for both.

As an alternative (that I don't really support), we can add `set` support for the current panacheQL so we can write a MongoDB update query as `update("query with set", params)` method. But I feel this would not be natural for MongoDB users. 

In this case, native query could be supported by providing the , inside the first varags parameters, the update document (a String or a BSON Document) ...

Examples:
```
update("where name=?1 set status=?2", "Mortal", Status.ALIVE);
update("{'name'=?1}", "{'status'=?2"}, "Mortal", Status.ALIVE);
```

For this example to works we need to support `set` inside the PanacheQL parser and the MongoParserVisitor.

As I said, I'm not a big fan of it and prefere to stick with `update(query, param).with(document, param)`.

emmanuelbernard: > @emmanuelbernard in fact the implementation is
> `update(filter, params).with(document, params)`
> 
> So
> `update("name", "Mortal").with("status", Status.Alive);`
> 
> Correspond to the following one implemented on Hibernate with Panache (but cannot be done easily with MongoDB)
> `update("name = ?1 where status = ?2", "Mortal", Status.Alive);`
> 
> To update without filter you could use the empty String or null (worth testing it)
> `update("").with("status", Status.Alive);`
> 
> We can also add an overload of update with no parameters for this case, good point :)

I am a bit confused. What you call filter is not the filter but the transformation clause. And the narrowing (filtering) is in the with operator? The way you describe it seems to be the opposite but your examples seems ot disagree with your text. In my head that is.

emmanuelbernard: So I think it's likely ok to have the double methods currently explored. But you guys still confuse the hell out of me and I still can't figure out where to put my transformations and where to put my narrowing predicate. Which makes me feel the names can be improved. Proposed alternative names

- update(...).where(...) (e.g. update("some field changes").where("some restrictions on fields") with where being optional
- update(...).for(...) with same semantic above
-

loicmathieu: **This is** `update(filter, params).with(newDocument, params)`. The fact that you are confuse show that the examples was not clear and maybe the API is not clear enougth !

Keep in mind that a typical MongoDB multi-update looks like this (from the mongo shell or a tool like Robot3T) :
```
dp.persons.update(
    {"name": "Mortal"}, 
    {"status": "Alive"},
    {"multi": true}
); 
```
This is a JavaScript method call with 3 parameters: filter, update document and options.

We can replace it with `update().filter(filter, params).withDocument(newDocument, params)` so that it is be more clear and the filter could be optional to provides update on all documents.

With this new API, there will be one more method call ...

The other advantage of this new API is that we can also add it to Hibernate with Panache if some people think it will be a good addition to the existing multi-update support :)

emmanuelbernard: So I guess you copied the notion of filter and document from Mongo (where document is roughly equivalent to SQL SET section - again if I understood but I’m still not sure). So the core or my problem is that the Mongo API confuses me. I would prefer the semantic I proposed earlier (update and where methods) which align with SQL and JP-QL and feel more natural to me.

loicmathieu: Yes, your guess is right :)

I'm OK with `update(updateDocument, params).where(filter, params);`, it will be a little less natural for peoples used to MongoDB but is still easy to understand.

For the records Spring Data do it this way:
```
mongoTemplate.updateMulti(
    new Query(where("accounts.accountType").is(Account.Type.SAVINGS)),
    new Update().inc("accounts.$.balance", 50.00), 
   Account.class
);
```

emmanuelbernard: Can you do the same query but with our API, with parameters and all? Looks like it will be more compact and readable.

loicmathieu: The Spring data example is based on their QueryDsl, a type safe query DSL that is very verbose.

I think that our aproach of update().with() or update().where() better match the Panache way (compact, defines query with strings, parameters, ...).

So I can propose the following API:
```
update(filter, params).with(updateDocument, params)// 1
update(updateDocument, params).where(filter, params)// 2
update().filter(filter, params).set(updateDocument, params)// 3
```

We can change the name of the method (I use `filter` because it matches MongoDB terms but you can use `where` instead).
If we plan to integrate the same facility inside Hibernate with Panache in addition to the current multi-update we need to choose terms that are not tied to SQL nor MongoDB ...

emmanuelbernard: 2

loicmathieu:
@FroMage @emmanuelbernard I refactor the code to swict to 
```
update(update, params).where(query, params)
```

I also added a `update(update, params).where()` to all updating all documents. Maybe we should called this method `all()`.

docs/src/main/asciidoc/mongodb-panache.adoc
FroMage: So you don't want a panache-ql version after all?

How about when there's no filter?

One possibility would be:

```java
update("foo = ?1 where foo is null", something) // panache-ql
updateNative("{'foo'=?1}", something).where("{'status'=?1"}, Status.Alive); // native update queries
updateNative("{'foo'=?1}", something).all(); // or some other no-filter term
```

loicmathieu: `update("foo = ?1 where foo is null", something)`
This didn't feels right to me, it's realy not the way to do it in MongoDB. And we cannot have a panache-ql version and the `update(update, params).where(query, params)` that we discuss with  @emmanuelbernard and the last one has my preference.

I would also prefere to keep the same methods for both native and non-native query as we did for all the other methods, at least inside `PanacheMongoEntityBase` and `PanacheMongoRepositoryBase` that already contains a lot of methods.

For the case there is no filter, I add a `update(update, params).where()` I ask in a comment wheter to have an empty `where()` or to have `all()` instead, none have my preference so you can choose ...

loicmathieu:
@FroMage @emmanuelbernard is it OK now with the last change ?
Should I switch the empty update method to `update(update, params).all()` instead of `update(update, params).where()` that is currentl y implemented ?

emmanuelbernard:
@loicmathieu +1 for `update(...).all()`

FroMage:
Yup, +1 too for `all`

loicmathieu:
PR updated.
I also add more tests.

docs/src/main/asciidoc/hibernate-orm-panache.adoc
```diff
@@ -385,8 +385,8 @@ personRepository.deleteAll();
 // delete by id
 boolean deleted = personRepository.deleteById(personId);
 
-// update all living persons
-personRepository.update("name = 'Moral' where status = ?1", Status.Alive);
+// set all persons named "Mortal" to living.
```
FroMage: This is not what it does. It updates all living persons to change their name to `Mortal`. As a result, the mongodb equivalent example should probably be changed to reflect that.

extensions/panache/mongodb-panache/runtime/src/main/java/io/quarkus/mongodb/panache/PanacheMongoEntityBase.java
```diff
@@ -875,6 +875,55 @@ public static void persistOrUpdate(Object firstEntity, Object... entities) {
         MongoOperations.persistOrUpdate(firstEntity, entities);
     }
 
+    /**
+     * Update all entities of this type by the given update document, with optional indexed parameters.
```
FroMage: It's not really a `document` here, is it?

extensions/panache/mongodb-panache/runtime/src/main/java/io/quarkus/mongodb/panache/PanacheMongoEntityBase.java
```diff
@@ -875,6 +875,55 @@ public static void persistOrUpdate(Object firstEntity, Object... entities) {
...
+     * @see #update(String, Map)
+     * @see #update(String, Parameters)
+     */
+    @GenerateBridge
+    public static PanacheUpdate update(String update, Object... params) {
+        throw MongoOperations.implementationInjectionMissing();
+    }
+
+    /**
+     * Update all entities of this type matching the given query, with named parameters.
```
FroMage: This talks about a query.

extensions/panache/mongodb-panache/runtime/src/main/java/io/quarkus/mongodb/panache/PanacheMongoEntityBase.java
FroMage: Also, I read the whole PR and I'm not sure what syntax is allowed in the `update` param besides the mongo native query. It says this is a query string, but it's not, is it?

Can I do `foo = :name, bar = :lastname` or `gee = 2`?

FroMage (changes requested):
Thanks. A few documentation questions remaining.

extensions/panache/mongodb-panache/runtime/src/main/java/io/quarkus/mongodb/panache/PanacheMongoEntityBase.java
loicmathieu: Yes, you can do it.

In reality, everything is a document in MongoDB, so there is no difference between what we called a query (MongoDB called it a filter document) and a stored document.

So you can use PanacheQL to express the update document, all the following are working update document:
```
FooBar.update("foo = :name and bar = :lastname", parameters).all()
FooBar.update("foo = :name, bar = :lastname", parameters).all()
FooBar.update("foo = ?1 and bar = ?2", name, lastname).all()
FooBar.update("foo = ?1, bar = ?2", name, lastname).all()
FooBar.update("{'foo' : :name, 'bar' : :lastname}", parameters).all()
FooBar.update("{'foo' : ?1, 'bar' : ?2}", name, lastname).all()
FooBar.update("foo", name).all()
```

docs/src/main/asciidoc/hibernate-orm-panache.adoc
loicmathieu: I update all comments and switch the MongoDB examples to mimic the one from Hibernate with Panache guide.

extensions/panache/mongodb-panache/runtime/src/main/java/io/quarkus/mongodb/panache/PanacheMongoEntityBase.java
loicmathieu: I added some sentence on the guide to document this inside the 'Simplified query' section.

extensions/panache/mongodb-panache/runtime/src/main/java/io/quarkus/mongodb/panache/PanacheMongoEntityBase.java
loicmathieu: I fixed the javadoc in all places.

docs/src/main/asciidoc/mongodb-panache.adoc
```diff
@@ -511,6 +517,17 @@ The MongoDB POJO codec doesn't support `ZonedDateTime` and `OffsetDateTime` so y
 
 MongoDB with Panache also supports extended MongoDB queries by providing a `Document` query, this is supported by the find/list/stream/count/delete methods.
 
+MongoDB with Panache offers operations to update multiple documents based on an update document and a query :
+`Person.update("foo = ?1 and bar = ?2", fooName, barName).where("name = ?1", name)`.
```
FroMage: If it works with `,` instead of `and` as the separator I'd prefer we document that and leave the `and` syntax undocumented, because the `and` makes it look like a query rather than a set of update commands.

FroMage (changes requested):

docs/src/main/asciidoc/mongodb-panache.adoc
loicmathieu: Yes, it works, I tested it.
I just updated the documentation part.

FroMage (approved):
OK perfect, thanks.

Issue #7152, opened 2020-02-11
Classes with @ConstructorProperties for Reflection


JohT:
**Description**
Classes with constructors annotated with `@java.beans.ConstructorProperties` should be registered for reflection without further configuration. 

Part of its JavaDoc:
> An annotation on a constructor that shows how the parameters of that constructor correspond to the constructed object's getter methods. [[1]](https://docs.oracle.com/javase/7/docs/api/java/beans/ConstructorProperties.html)

This annotation makes it possible to write immutable value objects that can be serialized with JSON-B [[2]](https://github.com/eclipse-ee4j/jsonb-api/issues/115#issue-398276491) and Jackson without any configuration and only depending on Java SE. This is helpful e.g. for api modules that shouldn't depend on any frameworks, but should be easy to integrate.

Since `@ConstructorProperties` is intended to be used with reflection, it makes IMO sense to register those classes. As a result, they could be used directly with `javax.ws.rs.core.Response`, `javax.json.bind.Jsonb`, `javax.ws.rs.sse.SseEventSink`,... without the need for `@RegisterForReflection` (quarkus-specific).

**Implementation ideas**
I'll provide a pull request as a proposal and for further discussion.

[[1] https://docs.oracle.com/javase/7/docs/api/java/beans/ConstructorProperties.html](https://docs.oracle.com/javase/7/docs/api/java/beans/ConstructorProperties.html)
[[2] https://github.com/eclipse-ee4j/jsonb-api/issues/115#issue-398276491](https://github.com/eclipse-ee4j/jsonb-api/issues/115#issue-398276491)

stuartwdouglas:
Looks like this was merged

PR #7153, opened 2020-02-11
Registering classes with @ConstructorProperties for Reflection for #7152

JohT:
First proposal on how classes with `@ConstructorProperties` annotated constructors could be registered for reflection. There fields aren't registered for now, because `@ConstructorProperties` only relates constructor parameters to getter methods.

### DIFF
--- a/core/deployment/src/main/java/io/quarkus/deployment/ConstructorPropertiesProcessor.java
+++ b/core/deployment/src/main/java/io/quarkus/deployment/ConstructorPropertiesProcessor.java
@@ -1,0 +1,42 @@
+package io.quarkus.deployment;
+
+import org.jboss.jandex.AnnotationInstance;
+import org.jboss.jandex.AnnotationTarget;
+import org.jboss.jandex.DotName;
+import org.jboss.jandex.IndexView;
+import org.jboss.jandex.MethodInfo;
+
+import io.quarkus.deployment.annotations.BuildProducer;
+import io.quarkus.deployment.annotations.BuildStep;
+import io.quarkus.deployment.builditem.CombinedIndexBuildItem;
+import io.quarkus.deployment.builditem.nativeimage.ReflectiveClassBuildItem;
+
+/**
+ * Registers all classes for reflection,
+ * that contain a constructor annotated with <code>@java.beans.ConstructorProperties</code>.
+ */
+public class ConstructorPropertiesProcessor {
+
+    private static final DotName CONSTRUCTOR_PROPERTIES = DotName.createSimple("java.beans.ConstructorProperties");
+
+    @BuildStep
+    void build(BuildProducer<ReflectiveClassBuildItem> reflectiveClass, CombinedIndexBuildItem indexBuildItem) {
+        IndexView index = indexBuildItem.getIndex();
+        for (AnnotationInstance annotationInstance : index.getAnnotations(CONSTRUCTOR_PROPERTIES)) {
+            registerInstance(reflectiveClass, annotationInstance);
+        }
+    }
+
+    private void registerInstance(BuildProducer<ReflectiveClassBuildItem> reflectiveClass, AnnotationInstance instance) {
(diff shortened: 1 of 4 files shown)


JohT:
Proposal for #7152

gastaldi:
Awesome! Any chance to have a test case too?

core/deployment/src/main/java/io/quarkus/deployment/ConstructorPropertiesProcessor.java
```diff
@@ -0,0 +1,38 @@
...
+import io.quarkus.deployment.builditem.CombinedIndexBuildItem;
+import io.quarkus.deployment.builditem.nativeimage.ReflectiveClassBuildItem;
+
+/**
+ * Registers all classes for reflection,
+ * that contain a constructor annotated with {@link ConstructorProperties}.
+ */
+public class ConstructorPropertiesProcessor {
+
+    private static final DotName CONSTRUCTOR_PROPERTIES = DotName.createSimple(ConstructorProperties.class.getName());
```
gsmet: Please use a String for the name instead. I prefer we don't have any dependency to `java.beans`.

core/deployment/src/main/java/io/quarkus/deployment/ConstructorPropertiesProcessor.java
```diff
@@ -0,0 +1,38 @@
...
+    @BuildStep
+    void build(BuildProducer<ReflectiveClassBuildItem> reflectiveClass, CombinedIndexBuildItem indexBuildItem) {
+        indexBuildItem.getIndex().getAnnotations(CONSTRUCTOR_PROPERTIES).stream()
+                .map(AnnotationInstance::target)
+                .filter(MethodInfo.class::isInstance)
+                .map(MethodInfo.class::cast)
+                .map(MethodInfo::declaringClass)
+                .map(ClassInfo::toString)
+                .map(this::asReflectiveClassBuildItem)
+                .forEach(reflectiveClass::produce);
```
gsmet: Honestly, this is far less readable than a good old for loop.

gsmet (changes requested):
Thanks. I added two small comments.

core/deployment/src/main/java/io/quarkus/deployment/ConstructorPropertiesProcessor.java
JohT: I'm kinda used to it, but it'll be indeed more readable using common known technique.

core/deployment/src/main/java/io/quarkus/deployment/ConstructorPropertiesProcessor.java
JohT: makes sense.

JohT:
I had some problems locally to get everything to build. Sorry in advance, if CI fails.

core/deployment/src/main/java/io/quarkus/deployment/ConstructorPropertiesProcessor.java
```diff
@@ -0,0 +1,44 @@
...
+import org.jboss.jandex.MethodInfo;
+
+import io.quarkus.deployment.annotations.BuildProducer;
+import io.quarkus.deployment.annotations.BuildStep;
+import io.quarkus.deployment.builditem.CombinedIndexBuildItem;
+import io.quarkus.deployment.builditem.nativeimage.ReflectiveClassBuildItem;
+
+/**
+ * Registers all classes for reflection,
+ * that contain a constructor annotated with {@link ConstructorProperties}.
```
gsmet: Could you remove the link here? I would like to avoid having the import.

Otherwise it looks good to me.

JohT: done, i missed that.

gsmet (approved):

gsmet:
Merged, thanks @JohT !

PR #7157, opened 2020-02-12
Always register the class of Jackson's JsonDeserialize#using

### LINKED ISSUE #7139: Enhance JsonDeserialize reflective class registration
**Description**
The Jackson extension JacksonProcessor [iterates](https://github.com/quarkusio/quarkus/blob/master/extensions/jackson/deployment/src/main/java/io/quarkus/jackson/deployment/JacksonProcessor.java#L94-L116) over anything that is annotated with `JsonDeserialize`. It only creates `ReflectiveClassBuildItem` for the "using" annotation value, if the annotation target kind is a method or field.

It'd be good if it also considered classes. I was working on an extension to integrate the Camel Olingo component into the Camel Quarkus project, and found that Apache Olingo has classes like:

```java
@JsonDeserialize(using = ClientCsdlAction.ActionDeserializer.class)
class ClientCsdlAction extends CsdlAction implements Serializable {
...
} 
```

`ClientCsdlAction.ActionDeserializer` is never registered for reflection by the Jackson extension so I had to add some custom logic register such classes for reflection in my own extension.

geoand:
Fixes: #7139

### DIFF
--- a/extensions/jackson/deployment/src/main/java/io/quarkus/jackson/deployment/JacksonProcessor.java
+++ b/extensions/jackson/deployment/src/main/java/io/quarkus/jackson/deployment/JacksonProcessor.java
@@ -106,12 +106,11 @@ public class JacksonProcessor {
                         addReflectiveHierarchyClass(builderClassName);
                     }
                 }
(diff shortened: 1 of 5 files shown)


gsmet (approved):
Makes perfect sense!

PR #7169, opened 2020-02-12
feature: Spring API-only deps

aureamunoz:
Hi, 
Here a first try of the Spring APIs, I continued the work done by @gytis.
WDYT? @n1hility @geoand 

Related to #3436

### DIFF
--- a/bom/runtime/pom.xml
+++ b/bom/runtime/pom.xml
@@ -170,10 +170,10 @@
         <sundr.version>0.19.1</sundr.version> <!-- this is to avoid annoying pop-up in eclipse about failure to init Velocity logging -->
         <flapdoodle.mongo.version>2.2.0</flapdoodle.mongo.version>
         <wiremock.version>2.24.1</wiremock.version>
-        <spring.version>5.1.8.RELEASE</spring.version>
-        <spring-data.version>2.1.9.RELEASE</spring-data.version>
-        <spring-security.version>5.2.0.RELEASE</spring-security.version>
-        <spring-boot.version>2.1.10.RELEASE</spring-boot.version>
+        <quarkus-spring-api.version>5.2.Alpha3</quarkus-spring-api.version>
+        <quarkus-spring-data-api.version>2.2.Alpha3</quarkus-spring-data-api.version>
+        <quarkus-spring-security-api.version>5.2.Alpha1</quarkus-spring-security-api.version>
+        <quarkus-spring-boot-api.version>2.1.Alpha1</quarkus-spring-boot-api.version>
         <mvel2.version>2.4.4.Final</mvel2.version>
         <htmlunit.version>2.36.0</htmlunit.version>
         <mockito.version>3.3.0</mockito.version>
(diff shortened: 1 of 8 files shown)


bom/runtime/pom.xml
```diff
@@ -169,6 +169,7 @@
         <spring-data.version>2.1.9.RELEASE</spring-data.version>
         <spring-security.version>5.2.0.RELEASE</spring-security.version>
         <spring-boot.version>2.1.10.RELEASE</spring-boot.version>
+        <quarkus-spring-api.version>0.0.1-SNAPSHOT</quarkus-spring-api.version>
```
geoand: I don't really like the `0.0.1` TBH. I would probably go for `0.1` as the first release

gsmet: Why not use the version of Spring they are based on? That would be far better IMHO.

bom/runtime/pom.xml
```diff
@@ -2571,6 +2572,26 @@
             </dependency>
 
             <!-- Spring -->
+            <dependency>
+                <groupId>me.snowdrop</groupId>
```
gsmet: Shouldn't they be under the `io.quarkus` groupId?

gsmet (changes requested):
Cool! I added a few comments inline.

bom/runtime/pom.xml
geoand: That's an option as well. We did discuss doing something like `2.1.3.Alpha1`, `2.1.3.CR1` etc, but we didn't settle on what the proper strategy should be

gsmet: Ah you would base it on Spring Boot. My inclination would have been to really use the Spring version of the related project. That way, you know what you are using.

gsmet: But thinking a bit more about it, it has been so long I used Spring that nowadays people might always speak in terms of Spring Boot versions.

I just think that it could be cool to check the version of Spring Security APIs and be able to easily find the related javadoc.

geoand: I'm not saying we need the Spring Boot version, I was just giving an example. I personally think that using the actual Spring component version instead of the Spring Boot version makes more sense as you say.
But then again there is the argument that for Quarkus, the Spring version doesn't really matter, since we don't implement evething anyway, nor are we affected by bugs in Spring projects

gsmet: I don't think that's a good call. They might add some features in new versions and knowing that we don't support them is good IMHO.

We could maybe track the major.minor version only.

geoand: That's probably a good compromise

bom/runtime/pom.xml
aureamunoz: As finally it was decided to hosted under snowdrop github organization I took the same `groupId` than we have in the snowdrop-bom, it seems confusing to me if we choose `io.quarkus` for something under `snowdrop`, no?

For the version I like using the Spring one, yes. WDYT @geoand ?

bom/runtime/pom.xml
aureamunoz: please, see https://github.com/quarkusio/quarkus/pull/7169#discussion_r378712096

bom/runtime/pom.xml
geoand: For me personally it doesn't make a difference. I can go either way. Just keep in mind that if we end up with `io.quarkus`, @gsmet you'll most likely be the one doing the releases (or grant me access perhaps).
If we end up with `snowdrop` then we'll be doing the releases and probably grant you access as well.

bom/runtime/pom.xml
aureamunoz: Ok, so 
`<quarkus-spring-api.version>5.1-SNAPSHOT</quarkus-spring-api.version>`?
as :
quarkus-spring-*-api --> `Spring Boot 2.1.12` --> `Spring Framework 5.1.13.RELEASE`

geoand: @aureamunoz I'm not sure what you mean `quarkus-spring-*-api --> Spring Boot 2.1.12 --> Spring Framework 5.1.13.RELEASE`.

Let's do `<quarkus-spring-web-api.version>5.1.Alpha1</quarkus-spring-web-api.version>` for Spring Web 5.1.x.RELEASE. Does that make sense?

aureamunoz: Yes, that makes sense for me.

I meant that currently the quarkus spring dependencies (`quarkus-spring-web-api `, `quarkus-spring-di-api` and so on) depend on Spring Boot 2.1.12.RELEASE which depend on Spring 5.1.13.RELEASE.

geoand: Got it, thanks.

I don't think we need to care about the Spring Boot version. The more I think about it the more I agree with @gsmet's proposal of using only the major and minor version of the actual Spring artifact

aureamunoz: Yes, I definitely agree with you. The commit is in progress ;-)

bom/runtime/pom.xml
gytis: If we got with `*.snowdrop` group id, it might make sense to have `dev.snowdrop` instead.

bom/runtime/pom.xml
gytis: We need to consider that we're likely to make multiple releases that will use the same Spring Boot version.
It's because we only picked a subset of classes from Spring that we need at the moment. If we'll add a new feature to an extension, we might need to include new classes to the dependencies jar.

geoand: Yes of course. My proposal is that we just add the proper qualifiers: `Alpha1`, `Alpha2`, `Beta1`, `Beta2`, `Final`, `CR1`, `CR2` etc.

gsmet: I don't think that will work. Let's say you have a Final and you end up needing a new annotation to implement a new feature in Quarkus?

geoand: That's what CR1 is supposed to be for.

aureamunoz:
Hi, just did the last change. It's ok for you?

bom/runtime/pom.xml
gytis: I assume we'll have to do similar dependencies bundle for Spring Boot as well (e.g. for spring boot properties extension). Are we going to use spring boot version for that?

bom/runtime/pom.xml
cescoffier: I think it would be better to have these modules on the quarkus side. The release process should not be a big burden and we can more easily test future releases and so on. We can grant @geoand (or someone else) the permission to push artifacts on `io.quarkus`.

geoand: That's fine with me

gytis: 100% agree

geoand:
So are we good with the proposed names and versions?

gsmet:
Not sure. Looks like we only have a Spring version, which looks weird for Spring Security for instance, which has a completely different lifecycle.

Also can we move everything to the `quarkus` org and have `io.quarkus` and also set up all the permissions for you to be able to release that?

gsmet:
Also, I think I would have added a third digit that is for us. Because we might have to fix an issue post Final and we will need an extra digit then.

geoand:
> Also can we move everything to the `quarkus` org and have `io.quarkus` and also set up all the permissions for you to be able to release that?

Yes, that works for me

geoand:
> Also, I think I would have added a third digit that is for us. Because we might have to fix an issue post Final and we will need an extra digit then.

I would prefer not to add the third digit. If we need a post-Final then that can just be `SR1`,..., `SRX` right?

aureamunoz:
Ok, I will move the repositories to `quarkus.io`.
So, for the versions, do you agree with following propositions: 
- `5.1.Alpha1` for `quarkus-spring-di-api`, `quarkus-spring-data-jpa-api` and `quarkus-spring-web-api` as Quarkus is using the `5.1.8.RELEASE` of Spring.
- `5.2.Alpha1` for `quarkus-spring-security-api` as Quarkus is using the `5.2.0.RELEASE` version of Spring Security.

Regarding the third digit I agree with @geoand because if we are not following the Spring versions it could be confusing.

bom/runtime/pom.xml
```diff
@@ -2724,6 +2725,26 @@
...
+                <version>${quarkus-spring-api.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>io.quarkus</groupId>
+                <artifactId>quarkus-spring-data-jpa-api</artifactId>
+                <version>${quarkus-spring-api.version}</version>
+            </dependency>
             <dependency>
                 <groupId>org.springframework</groupId>
                 <artifactId>spring-context</artifactId>
```
gsmet: Shouldn't the dependency below be removed from the BOM? We don't expect Spring artifacts anymore, do we?

bom/runtime/pom.xml
```diff
@@ -2724,6 +2725,26 @@
...
+            </dependency>
+            <dependency>
+                <groupId>io.quarkus</groupId>
+                <artifactId>quarkus-spring-security-api</artifactId>
+                <version>${quarkus-spring-api.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>io.quarkus</groupId>
+                <artifactId>quarkus-spring-data-jpa-api</artifactId>
+                <version>${quarkus-spring-api.version}</version>
```
gsmet: So do we go with specific versions for each Spring component?

extensions/spring-di/deployment/pom.xml
```diff
@@ -35,9 +35,9 @@
         </dependency>
 
         <dependency>
-            <groupId>org.springframework</groupId>
-            <artifactId>spring-context</artifactId>
-            <scope>test</scope>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-spring-di-api</artifactId>
+<!--            <scope>test</scope>-->
```
gsmet: This is weird.

The dependency should only be in the runtime artifact, shouldn't it?

extensions/spring-security/deployment/pom.xml
```diff
@@ -50,6 +50,10 @@
             <artifactId>quarkus-security-test-utils</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-spring-security-api</artifactId>
+        </dependency>
```
gsmet: This dependency should only be in the runtime artifact, shouldn't it?

extensions/spring-web/deployment/pom.xml
```diff
@@ -41,6 +41,10 @@
             <artifactId>rest-assured</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-spring-web-api</artifactId>
```
gsmet: Same here, I don't expect to see that one in the deployment artifact.

gsmet (changes requested):
I added some additional comments. Please squash everything once you're done.

geoand:
@aureamunoz the Spring Data JPA stuff failed. It seems like there is a class missing from the Jar:

```
2020-03-05T15:39:22.1379473Z Caused by: java.lang.ClassNotFoundException: org.springframework.data.domain.Chunk
```

geoand:
@gsmet tests are passing

gsmet (approved):

PR #7186, opened 2020-02-13
MicroProfile Metrics 2.3

### LINKED ISSUE #6220: REST metrics
Provide metrics related to usage of REST endpoints. If possible, this should be based on the "optional REST metrics" recommendations that are planned to be included in MP Metrics 2.3 specification, and possibly extend beyond these recommendations.

jmartisk:
Fixes https://github.com/quarkusio/quarkus/issues/6220

### DIFF
--- a/extensions/jaeger/runtime/pom.xml
+++ b/extensions/jaeger/runtime/pom.xml
@@ -34,6 +34,10 @@
             <groupId>io.smallrye</groupId>
             <artifactId>smallrye-metrics</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.eclipse.microprofile.metrics</groupId>
+            <artifactId>microprofile-metrics-api</artifactId>
+        </dependency>
     </dependencies>
 
     <build>
--- a/extensions/resteasy-server-common/deployment/src/main/java/io/quarkus/resteasy/server/common/deployment/ResteasyServerCommonProcessor.java
+++ b/extensions/resteasy-server-common/deployment/src/main/java/io/quarkus/resteasy/server/common/deployment/ResteasyServerCommonProcessor.java
@@ -16,6 +16,8 @@ import java.util.Set;
 import java.util.function.BiFunction;
 import java.util.stream.Collectors;
 
+import javax.servlet.DispatcherType;
+
 import org.jboss.jandex.AnnotationInstance;
 import org.jboss.jandex.AnnotationTarget;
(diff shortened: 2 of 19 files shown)


extensions/jaeger/runtime/pom.xml
```diff
@@ -34,6 +34,10 @@
             <groupId>io.smallrye</groupId>
             <artifactId>smallrye-metrics</artifactId>
         </dependency>
+        <dependency>
```
jmartisk: before someone asks, this is needed because `microprofile-metrics-api` is now a `provided` dependency of SR Metrics so it's not brought in transitively

extensions/resteasy-server-common/deployment/src/main/java/io/quarkus/resteasy/server/common/deployment/ResteasyServerCommonProcessor.java
```diff
@@ -412,6 +424,33 @@ void beanDefiningAnnotations(BuildProducer<BeanDefiningAnnotationBuildItem> bean
                         BuiltinScope.SINGLETON.getName()));
     }
 
+    @BuildStep
+    void enableMetrics(ResteasyConfig buildConfig,
```
gastaldi: Shouldn't this code be in the `smallrye-metrics` extension?

jmartisk: All other extensions that enable metrics do it in their own processor. 
This extension has all the needed dependencies for it, Metrics extension doesn't. And it sounds odd, if it is at all possible, that Metrics extension would inject configuration of a different extension.

gastaldi: I thought that those classes would be available since you added `quarkus-resteasy` as an optional dependency ni `extensions/smallrye-metrics/runtime/pom.xml`, but I'll let someone more experienced review that ;)

jmartisk: That's just the runtime artifact and it's because `QuarkusJaxRsMetricsFilter` uses JAX-RS api to compile

kenfinnigan (approved):

PR #7187, opened 2020-02-13
exclude hikari from the quartz extension

michalszynkiewicz:
Since we added exclusion for hikari to quartz, it changed from using java6 to java7 version.

### DIFF
--- a/extensions/quartz/runtime/pom.xml
+++ b/extensions/quartz/runtime/pom.xml
@@ -33,7 +33,11 @@
             <exclusions>
                 <exclusion>
                     <groupId>com.zaxxer</groupId>
-                    <artifactId>HikariCP-java6</artifactId>
+                    <artifactId>HikariCP</artifactId>
+                </exclusion>
+                <exclusion>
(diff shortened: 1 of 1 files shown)


gastaldi (approved):
LGTM

extensions/quartz/runtime/pom.xml
```diff
@@ -33,7 +33,7 @@
             <exclusions>
                 <exclusion>
                     <groupId>com.zaxxer</groupId>
-                    <artifactId>HikariCP-java6</artifactId>
+                    <artifactId>HikariCP-java7</artifactId>
```
gsmet: I think I would keep excluding the old one and also exclude `HikariCP`.

gsmet (changes requested):

extensions/quartz/runtime/pom.xml
michalszynkiewicz: done

gsmet (approved):

PR #7197, opened 2020-02-14
chore: update kotlin version to 1.3.61

machi1990:
/cc @geoand @gastaldi

### DIFF
--- a/bom/runtime/pom.xml
+++ b/bom/runtime/pom.xml
@@ -141,7 +141,7 @@
         <aws-lambda-serverless-java-container.version>1.3.1</aws-lambda-serverless-java-container.version>
         <awssdk.version>2.10.3</awssdk.version>
         <azure-functions-java-library.version>1.3.0</azure-functions-java-library.version>
-        <kotlin.version>1.3.41</kotlin.version>
+        <kotlin.version>1.3.61</kotlin.version>
         <dekorate.version>0.10.9</dekorate.version>
         <maven-artifact-transfer.version>0.10.0</maven-artifact-transfer.version>
         <jline.version>2.14.6</jline.version>
--- a/build-parent/pom.xml
+++ b/build-parent/pom.xml
@@ -23,7 +23,7 @@
         <!-- These properties are needed in order for them to be resolvable by the generated projects -->
         <!-- Quarkus uses jboss-parent and it comes with 3.8.0-jboss-2, we don't want that in the templates -->
         <compiler-plugin.version>3.8.1</compiler-plugin.version>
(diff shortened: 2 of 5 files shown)


.dependabot/config.yml
```diff
@@ -83,6 +83,15 @@ update_configs:
           dependency_name: "org.scala-lang:scala-compiler"
       - match:
           dependency_name: "net.alchim31.maven:scala-maven-plugin"
+      # Kotlin
```
geoand: I don't think we should add thing configuration since there are too many places where the version is being used and I don't think the bot can catch them all

geoand:
Searching for `1.3.41` in the codebase I see more instances than are updated by the PR.

.dependabot/config.yml
machi1990: I have tried to be exhaustive when adding the configuration in the hope of overcoming that situation.

machi1990:
> Searching for `1.3.41` in the codebase I see more instances than are updated by the PR.

Can you point me where ? A quick search from my side seems like we have all the places covered.

.dependabot/config.yml
geoand: Basically there are test projects that also use this version which explains why it's all over the place. If we find a way to consolidate them, then I think the configuration makes sense. Otherwise I would prefer we do manual updates in order to not end up with mixed versions

geoand:
You are right, I had a few false positives (one comments and some stuff in `target`).

.dependabot/config.yml
geoand: The bot won't be able to update all the instances will it? If not, lets get rid of this configuration

machi1990: Oh yeah, just saw that we have a the version in a Kotlin IT template project. Not sure the bot will see and update that. Let's keep this manual for now. Thanks @geoand

geoand (approved):
Thanks!

machi1990:
> Thanks!

Thanks for the quick review.

gwenneg:
This is very very minor, but should the following line be updated too?

https://github.com/quarkusio/quarkus/blob/8f7eeaae32bb7a2fe30c7677174ae5043cc50c3e/extensions/kotlin/deployment/src/main/java/io/quarkus/kotlin/deployment/KotlinCompilationProvider.java#L28

The new version would be:
`// see: https://github.com/JetBrains/kotlin/blob/v1.3.61/libraries/tools/kotlin-maven-plugin/src/main/java/org/jetbrains/kotlin/maven/KotlinCompileMojoBase.java#L181`

gwenneg (approved):

PR #7194, opened 2020-02-13
if a configuration contains test-related entries, DevMojo crashes.  

evanchooly:
filter out such elements before starting dev mode

PR #7202, opened 2020-02-14
Remove test-related entries in maven-compiler-plugin

gastaldi:
Some properties (like `testSource` and `testTarget`) are valid only for the `testCompile` mojo, however they are allowed to be declared as part of the maven-compiler-plugin configuration

Closes #7194


gsmet:
Open question: should we whitelist options that we know should be there instead?

gastaldi:
That is relative to what's configured in the project's pom.xml. IMHO we should just use what's there, in order to avoid inconsistencies with `mvn compile`

aloubyansky:
I'd think there hardly should be any option that we shouldn't allow in the first place *for the goal that we invoke*. That information is actually available in the `plugin.xml`. It's be a matter of fetching that.

aloubyansky (approved):

PR #7209, opened 2020-02-15
JPA - Criteria

italogustavoborba:


gastaldi:
Any chance to add a test case?

extensions/panache/hibernate-orm-panache/runtime/src/main/java/io/quarkus/hibernate/orm/panache/runtime/JpaOperations.java
```diff
@@ -267,6 +268,13 @@ public static Object findById(Class<?> entityClass, Object id, LockModeType lock
         bindParameters(jpaQuery, params);
         return new PanacheQueryImpl(em, jpaQuery, findQuery, params);
     }
+    
+    @SuppressWarnings("rawtypes")
+    public static PanacheQuery<?> findAll(Class<?> entityClass, CriteriaQuery<?> criteriaQuery) {
```
gastaldi: I'm not sure if JPA objects should be part of the API, do you have an use case?

italogustavoborba: JPA was not used

gastaldi: `CriteriaQuery` is part of the JPA API no?

italogustavoborba: `javax.persistence.criteria. CriteriaQuery`

italogustavoborba: This project already uses: `javax.persistence.*`

extensions/panache/hibernate-orm-panache/runtime/src/main/java/io/quarkus/hibernate/orm/panache/runtime/JpaOperations.java
```diff
@@ -268,6 +269,13 @@ public static Object findById(Class<?> entityClass, Object id, LockModeType lock
         return new PanacheQueryImpl(em, jpaQuery, findQuery, params);
     }
 
+    @SuppressWarnings("rawtypes")
+    public static PanacheQuery<?> findAll(Class<?> entityClass, CriteriaQuery<?> criteriaQuery) {
+        String query = "FROM " + getEntityName(entityClass);
+        EntityManager em = getEntityManager();
+        return new PanacheQueryImpl(em, em.createQuery(criteriaQuery), query, null);
```
FroMage: This looks just wrong. I don't think `PanacheQuery.count()` will work without using the correct count query. A criteria query can filter and so it is not an "all" query.

extensions/panache/hibernate-orm-panache/runtime/src/main/java/io/quarkus/hibernate/orm/panache/PanacheEntityBase.java
```diff
@@ -257,6 +258,18 @@ public void flush() {
...
 
+    /**
+     * Find all entities of this type, in the given criteria.
+     *
+     * @param criteria to use
+     * @return a new {@link PanacheQuery} instance to find all entities of this type.
+     * @see #findAll()
+     */
+    @GenerateBridge
+    public static <T extends PanacheEntityBase> PanacheQuery<T> findAll(CriteriaQuery<T> criteria) {
```
FroMage: This should better be in `find` because `CriteriaQuery` can have `WHERE` clauses and so it's not necessarily "all results". Furthermore, it would likely need the equivalent `list` and `stream` shortcuts.

FroMage (changes requested):
Thanks for the PR. As it is, you need to tweak a few things to make it work.

The biggest question is for @Sanne and @emmanuelbernard as to whether it's worth adding support for `CriteriaQuery` in the API.

extensions/panache/hibernate-orm-panache/runtime/src/main/java/io/quarkus/hibernate/orm/panache/runtime/JpaOperations.java
italogustavoborba: Count ex:
`CriteriaBuilder builder = em.getCriteriaBuilder();
CriteriaQuery<Long> countCriteria = builder.createQuery(Long.class);
Root<Entity> entityRoot = countCriteria.from(criteriaQuery.getResultType());
countCriteria.select(builder.count(entityRoot));
countCriteria.where(criteriaQuery.getRestriction());
return em.createQuery(countCriteria).getSingleResult();`

emmanuelbernard:
So my opinion is that the JPA Criteria API is not worth its weight in complexity. Couldn't you just use non Panache APIs for these usages? What does Panache brings to the table here?

Also @italogustavoborba , I'm really interested in why you feel you want the Criteria API? I'll leave it at that not to influence you further in your response.

italogustavoborba:
> So my opinion is that the JPA Criteria API is not worth its weight in complexity. Couldn't you just use non Panache APIs for these usages? What does Panache brings to the table here?
> 
> Also @italogustavoborba , I'm really interested in why you feel you want the Criteria API? I'll leave it at that not to influence you further in your response.

Complex queries migrated from other systems that were made with criteria (my case) prevent its use. And would follow the pattern of objects for mounting queries

emmanuelbernard:
@italogustavoborba so you can still define your entities using Panache. Just that the queries would use the old way of writing it.

```
@Entity
public Foo extends PanacheEntity {
    public String title;

    // modern query
    public List<Foo> getByTitle(String title) { return find("title", title); }

    // Criteria API
    public List<Foo> getAlsoByTitle(String title) {
        return getEntityManager().em.createQuery(criteriaQuery).setParameter(...).getResultList();    
    }
}
```

@FroMage do we have a getEntityManager() method or equivalent for this soft fallback?

FroMage:
We have `JpaOperations.getEntityManager()` but we should totally add it as an instance method too.

emmanuelbernard:
There you go @FroMage https://github.com/quarkusio/quarkus/issues/7233
@italogustavoborba wanna do a PR for this one? https://github.com/quarkusio/quarkus/issues/7233

italogustavoborba:
> @italogustavoborba so you can still define your entities using Panache. Just that the queries would use the old way of writing it.

Cool, I just wanted to contribute in some way, I already do it using this way.

italogustavoborba:
> wanna do a PR for this one

Sure.

gastaldi:
> We have `JpaOperations.getEntityManager()` but we should totally add it as an instance method too.

Probably should be a protected method, to be accessed only inside the entity?

loicmathieu:
Support for JPA Query has already been discussed.

MongoDB with Panache allows to make queries with a `Document` to allow more advanced use case, so I'm in favor of providing the same capacity for Hibernate with Panache.

But if we agree on this, I think it will be better to accept a parameter type of `javax.persistence.Query` to cover more use cases.

gastaldi:
Superseded by https://github.com/quarkusio/quarkus/pull/10541. Closing this

PR #7211, opened 2020-02-15
Implement Mutiny support for the mailer service

cescoffier:
* Implement the Mutiny API
* Deprecated the previous API
* Update tests and documentation

These changes are backward compatible. 

I've also updated:
* the mock mailer to print `<empty>` instead of `null` 
* fix a few logged messages

The template support still returns a `CompletionStage`. It will be covered once #7101 is merged. This would be a breaking change.

### DIFF
--- a/extensions/mailer/runtime/src/main/java/io/quarkus/mailer/reactive/ReactiveMailer.java
+++ b/extensions/mailer/runtime/src/main/java/io/quarkus/mailer/reactive/ReactiveMailer.java
@@ -1,0 +1,19 @@
+package io.quarkus.mailer.reactive;
+
+import io.quarkus.mailer.Mail;
+import io.smallrye.mutiny.Uni;
+
+/**
+ * A mailer to send email asynchronously.
+ */
+public interface ReactiveMailer {
+
+    /**
+     * Sends the given emails.
+     *
+     * @param mails the emails to send, must not be {@code null}, must not contain {@code null}
+     * @return a {@link Uni} indicating when the mails have been sent. The {@link Uni} may fire a failure if the
+     *         emails cannot be sent.
+     */
+    Uni<Void> send(Mail... mails);
+}
--- a/docs/src/main/asciidoc/mailer.adoc
+++ b/docs/src/main/asciidoc/mailer.adoc
@@ -100,9 +100,15 @@ ReactiveMailer reactiveMailer;
 There are 2 APIs:
 
 * `io.quarkus.mailer.Mailer` provides the imperative (blocking and synchronous) API;
-* `io.quarkus.mailer.ReactiveMailer` provides the reactive (non-blocking and asynchronous) API
+* `io.quarkus.mailer.reactive.ReactiveMailer` provides the reactive (non-blocking and asynchronous) API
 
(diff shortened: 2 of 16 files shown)


extensions/mailer/runtime/src/main/java/io/quarkus/mailer/reactive/ReactiveMailer.java
```diff
@@ -0,0 +1,19 @@
...
+public interface ReactiveMailer {
+
+    /**
+     * Sends the given emails.
+     *
+     * @param mails the emails to send, must not be {@code null}, must not contain {@code null}
+     * @return an {@link Uni} indicating when the mails have been sent. The {@link Uni} may fire a failure if the
+     *         emails cannot be sent.
+     */
+    Uni<Void> send(Mail... mails);
```
gastaldi: Perhaps the returned object may give more info instead of `Void``? `MailResult` perhaps?

cescoffier: And what would it contains? Basically it just tells you success or failure which is conveyed using the Uni events.

cescoffier:
Force pushed to fix the import ordering issue.

extensions/mailer/runtime/src/main/java/io/quarkus/mailer/reactive/ReactiveMailer.java
gastaldi: I was thinking about the message itself but maybe `Uni` already handles that

extensions/mailer/runtime/src/main/java/io/quarkus/mailer/reactive/ReactiveMailer.java
```diff
@@ -0,0 +1,19 @@
...
+/**
+ * A mailer to send email asynchronously.
+ */
+public interface ReactiveMailer {
+
+    /**
+     * Sends the given emails.
+     *
+     * @param mails the emails to send, must not be {@code null}, must not contain {@code null}
+     * @return an {@link Uni} indicating when the mails have been sent. The {@link Uni} may fire a failure if the
```
gsmet:
```suggestion
     * @return a {@link Uni} indicating when the mails have been sent. The {@link Uni} may fire a failure if the
```

gsmet (approved):
I just had one small Javadoc comment. Otherwise looks good.

Feel free to fix and merge or ignore and merge :).

extensions/mailer/runtime/src/main/java/io/quarkus/mailer/reactive/ReactiveMailer.java
gsmet: Fixed.

gsmet:
I fixed the small typo and rebased. Let's wait for CI.

Akejelu-Nelu:
I have difficulty in getting the response code for each email i sent with reactive mailer.

 return reactiveMailer.send(mail)
                .subscribeAsCompletionStage()
                .thenApply(response -> Response.accepted().build())
                .exceptionally(throwable -> Response.serverError().build());

this is what my return statement looks like after building the mail. The mail object also has a list of recipients. I'm not sure how i can get the response code / status of that email if it was sent or not. Any suggestion is appreciated.

PR #7216, opened 2020-02-16
Update the reactive database client extensions to use Mutiny

cescoffier:
* Add Mutiny support to the reactive MySQL extension
* Add Mutiny support to the reactive PostgreSQL extension
* Update documentation
* Update ITs

CC @tsegismont

### DIFF
--- a/docs/src/main/asciidoc/reactive-sql-clients.adoc
+++ b/docs/src/main/asciidoc/reactive-sql-clients.adoc
@@ -88,7 +88,9 @@ Otherwise, you can manually add this to the dependencies section of your `pom.xm
 ----
 
 NOTE: In this guide, we will use the Axle API of the Reactive PostgreSQL Client.
-Read the link:vertx[Using Eclipse Vert.x] guide to understand the differences between the callback, RxJava and Axle based APIs.
+Read the link:vertx[Using Eclipse Vert.x] guide to understand the differences between the callback, Mutiny, RxJava and Axle based APIs.
+The RxJava and Axle API are deprecated, and are planned for removal.
+It is recommended to switch to Mutiny.
 
 === JSON Binding
 
@@ -127,7 +129,7 @@ quarkus.datasource.password=quarkus_test
 IMPORTANT: The eagle eyes among you might have noticed the `url` prefix is different from usual JDBC urls.
 It is required to use the `vertx-reactive:` prefix otherwise the Reactive PostgreSQL Client extension will not use the `quarkus.datasource.url` config property.
 
-With that you may create your `FruitResource` skeleton and `@Inject` a `io.vertx.axle.pgclient.PgPool` instance:
+With that you may create your `FruitResource` skeleton and `@Inject` a `io.vertx.mutiny.pgclient.PgPool` instance:
 
 [source,java]
 .src/main/java/org/acme/vertx/FruitResource.java
@@ -138,7 +140,7 @@ With that you may create your `FruitResource` skeleton and `@Inject` a `io.vertx
 public class FruitResource {
 
     @Inject
-    io.vertx.axle.pgclient.PgPool client;
+    io.vertx.mutiny.pgclient.PgPool client;
 }
 ----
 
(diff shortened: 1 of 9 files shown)


tsegismont:
I'll talk a look tomorrow

extensions/reactive-pg-client/runtime/src/main/java/io/quarkus/reactive/pg/client/runtime/PgPoolProducer.java
```diff
@@ -4,36 +4,87 @@
...
 import javax.inject.Singleton;
 
+import org.jboss.logging.Logger;
+
 import io.vertx.pgclient.PgPool;
 
 @ApplicationScoped
 public class PgPoolProducer {
 
-    private volatile PgPool pgPool;
```
tsegismont: I believe the `volatile` construct was preferred here because the producer can be invoked concurrently at runtime (e.g. in request scoped beans). Using synchronization instead would serialize these calls. Or am I missing something?

docs/src/main/asciidoc/reactive-sql-clients.adoc
```diff
@@ -174,20 +176,19 @@ TIP: You may override the default value of the `myapp.schema.create` property in
...
 ----
-        client.query("DROP TABLE IF EXISTS fruits")
-                .thenCompose(r -> client.query("CREATE TABLE fruits (id SERIAL PRIMARY KEY, name TEXT NOT NULL)"))
-                .thenCompose(r -> client.query("INSERT INTO fruits (name) VALUES ('Orange')"))
-                .thenCompose(r -> client.query("INSERT INTO fruits (name) VALUES ('Pear')"))
-                .thenCompose(r -> client.query("INSERT INTO fruits (name) VALUES ('Apple')"))
-                .toCompletableFuture()
-                .join();
+ client.query("DROP TABLE IF EXISTS fruits")
+    .onItem().produceUni(r -> client.query("CREATE TABLE fruits (id SERIAL PRIMARY KEY, name TEXT NOT NULL)"))
```
tsegismont: Shouldn't we use the same API methods in the quickstart and the doc? (`flatMap` in the quickstart)

tsegismont (commented):
LGTM @cescoffier , a couple of concerns

extensions/reactive-pg-client/runtime/src/main/java/io/quarkus/reactive/pg/client/runtime/PgPoolProducer.java
cescoffier: I actually want the serialization here to catch the creation of the deprecated clients. 
I can restore using a volatile for the `PgPool` but that mix 2 synchronization constructs.

docs/src/main/asciidoc/reactive-sql-clients.adoc
cescoffier: The quickstart has not been updated. So where did you find the `flatMap`?

tsegismont: Sorry, actually `flatMap` is used in the tests (`FruitResource`)

cescoffier: oh good catch, yes, it should be consistent (both are equivalent BTW).

extensions/reactive-pg-client/runtime/src/main/java/io/quarkus/reactive/pg/client/runtime/PgPoolProducer.java
cescoffier: Fixed.

docs/src/main/asciidoc/reactive-sql-clients.adoc
cescoffier: Fixed.

docs/src/main/asciidoc/reactive-sql-clients.adoc
```diff
@@ -174,20 +176,19 @@ TIP: You may override the default value of the `myapp.schema.create` property in
...
 ----
-        client.query("DROP TABLE IF EXISTS fruits")
-                .thenCompose(r -> client.query("CREATE TABLE fruits (id SERIAL PRIMARY KEY, name TEXT NOT NULL)"))
-                .thenCompose(r -> client.query("INSERT INTO fruits (name) VALUES ('Orange')"))
-                .thenCompose(r -> client.query("INSERT INTO fruits (name) VALUES ('Pear')"))
-                .thenCompose(r -> client.query("INSERT INTO fruits (name) VALUES ('Apple')"))
-                .toCompletableFuture()
-                .join();
+ client.query("DROP TABLE IF EXISTS fruits")
+    .flatMap(r -> client.query("CREATE TABLE fruits (id SERIAL PRIMARY KEY, name TEXT NOT NULL)"))
```
tsegismont: `flatMap` is fine with me, so is the Mutiny flavor with `onItem`. Your call.

My only concern is that the test and the doc are consistent, so all good. :+1:

extensions/reactive-mysql-client/runtime/src/main/java/io/quarkus/reactive/mysql/client/runtime/MySQLPoolProducer.java
```diff
@@ -4,36 +4,90 @@
...
 
 @ApplicationScoped
 public class MySQLPoolProducer {
 
+    private static final Logger LOGGER = Logger.getLogger(MySQLPoolProducer.class);
+
     private volatile MySQLPool mysqlPool;
-    private volatile io.vertx.axle.mysqlclient.MySQLPool axleMySQLPool;
-    private volatile io.vertx.reactivex.mysqlclient.MySQLPool rxMySQLPool;
+    private io.vertx.mutiny.mysqlclient.MySQLPool mutinyMySQLPool;
```
tsegismont: How about using `AtomicReference` to remove synchronization without losing the checks in the producer methods?

tsegismont (commented):
Did another pass @cescoffier

extensions/reactive-mysql-client/runtime/src/main/java/io/quarkus/reactive/mysql/client/runtime/MySQLPoolProducer.java
cescoffier: `AtomicReference.compareAndSet(null, instance)` requires the instance to be already created - so it's not lazy. Using `updateAndGet` could work but is rather convoluted (you would need to move the `if` check in the binary operator).

docs/src/main/asciidoc/reactive-sql-clients.adoc
cescoffier: I updated the commit message to explain why I'm ok with `flatMap` and `map` here. Reactive programmers are used to them.

extensions/reactive-mysql-client/runtime/src/main/java/io/quarkus/reactive/mysql/client/runtime/MySQLPoolProducer.java
tsegismont: Ok, let's revisit later if that proves to be a problem

gsmet (approved):

PR #7218, opened 2020-02-16
Provide jpa metamodel for PanacheEntity class

Postremus:
The created meta model for an entity extends PanacheEntity_.
PanacheEntity_ is not created during the quarkus build, and therefore would have to be created by the user, which is tedius.

We now just run the jpamodelgen in the hiberate-orm-panache modul, creating PanacheEntity_ automaticly during the build.

### DIFF
--- a/extensions/panache/hibernate-orm-panache/runtime/pom.xml
+++ b/extensions/panache/hibernate-orm-panache/runtime/pom.xml
@@ -39,6 +39,36 @@
             <artifactId>quarkus-jackson</artifactId>
             <optional>true</optional>
         </dependency>
+
+        <dependency>
+            <groupId>org.hibernate</groupId>
+            <artifactId>hibernate-jpamodelgen</artifactId>
+            <scope>provided</scope>
+            <!-- Exclude these dependencies, since they are banned -->
+            <exclusions>
+                <exclusion>
+                    <groupId>javax.xml.bind</groupId>
+                    <artifactId>jaxb-api</artifactId>
+                </exclusion>
+                <exclusion>
+                    <groupId>javax.activation</groupId>
+                    <artifactId>javax.activation-api</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+
+        <!-- Explicitly provide banned dependencies, in the correct version -->
+        <dependency>
+            <groupId>org.jboss.spec.javax.xml.bind</groupId>
+            <artifactId>jboss-jaxb-api_2.3_spec</artifactId>
+            <scope>provided</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>com.sun.activation</groupId>
+            <artifactId>jakarta.activation</artifactId>
+            <scope>provided</scope>
+        </dependency>
     </dependencies>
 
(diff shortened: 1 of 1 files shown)


extensions/panache/hibernate-orm-panache/runtime/pom.xml
```diff
@@ -39,6 +39,36 @@
             <artifactId>quarkus-jackson</artifactId>
             <optional>true</optional>
         </dependency>
+
+        <dependency>
+            <groupId>org.hibernate</groupId>
+            <artifactId>hibernate-jpamodelgen</artifactId>
+            <scope>provided</scope>
+            <!-- Exclude these dependencies, since the are banned -->
```
gsmet:
```suggestion
            <!-- Exclude these dependencies, since they are banned -->
```

gsmet (changes requested):
Thanks.

I'll let @FroMage decide but I think it makes sense to do it. Could you apply my suggestion and squash everything?

extensions/panache/hibernate-orm-panache/runtime/pom.xml
Postremus: sure, I will do this change this evening.

FroMage:
I've no idea what this will do. What's the JPA meta-model, and why do we need to generate it?

Postremus:
@FroMage See Example 1 and Example 2 from https://docs.jboss.org/hibernate/orm/5.0/topical/html/metamodelgen/MetamodelGenerator.html.

The metamodel allows you to simplify your criteriaquery.

For example, I can write:
`where.add(builder.not(historyEntry.get(HistoryEntry_.eventType).in(filter.getExlusiveEventTypes())));`

Instead of:
`where.add(builder.not(historyEntry.get("eventType").in(filter.getExlusiveEventTypes())));`.

This saves me from referencing "entityType" as String everywhere.
It also allows type safety.

FroMage:
OK, thanks. I guess it can't hurt, but I'm not familiar with this at all, so asking @Sanne and @emmanuelbernard for advice.

gsmet:
Yeah, it's useful if you use JPA criterias. If you have nothing against it, I will merge it once the typo is fixed.

Sanne:
Nice!

Postremus:
@gsmet done

gsmet (approved):

gsmet:
Merged, thanks!

geoand:
I think it would be nice to have a section in the Panache guide showing how this could be used? I personally don't know much about the meta-model and would like to know how to use it :)

emmanuelbernard:
Nan, maybe in the normal Hibernate doc. This tool should be in the
Hibernate ORM extension anyways, not in the Panache one as it's generally
useful.

gsmet:
The idea here is to provide the meta model for `PanacheEntity`.

geoand:
> The idea here is to provide the meta model for `PanacheEntity`.

Exactly. So perhaps some quick info on how to use it with Panache and then perhaps a link to the extensive Hibernate documentation?

PR #7222, opened 2020-02-17
Support transaction inside MongoDB with Panache via Narayana JTA

loicmathieu:
@FroMage I have a working-ish version with Narayana JTA but I'm not sure it's correctly implemented. I'm not an expert on transaction management.

Is it better than #6365


loicmathieu:
@FroMage when you have some time can you look at this implementation proposal of MongoDB transaction with Narayana JTA and if it's better than #6365 ?

loicmathieu:
ping @FroMage can you have a look at this one please ?

extensions/panache/mongodb-panache/runtime/src/main/java/io/quarkus/mongodb/panache/transaction/interceptor/ClienSessionHolder.java
```diff
@@ -0,0 +1,24 @@
+package io.quarkus.mongodb.panache.transaction.interceptor;
+
+import java.util.Optional;
+
+import com.mongodb.client.ClientSession;
+
+public final class ClienSessionHolder {
```
FroMage, line 7: I don't think you need this class. And its name has a missing `t` at `ClientT`.

extensions/panache/mongodb-panache/runtime/src/main/java/io/quarkus/mongodb/panache/transaction/interceptor/MongoTransaction.java
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    @Override
+    public void rollback() throws IllegalStateException, SystemException {
+        clientSession.abortTransaction();
+        this.status = Status.STATUS_ROLLEDBACK;
+    }
+
+    @Override
+    public void setRollbackOnly() throws IllegalStateException, SystemException {
```
FroMage, line 60: You need to support this, it's pretty useful.

FroMage: And easy to implement.

extensions/panache/mongodb-panache/runtime/src/main/java/io/quarkus/mongodb/panache/transaction/interceptor/MongoTransactionalInterceptor.java
```diff
@@ -0,0 +1,136 @@
...
+        if (!transactional.readPreference().equals("<default>")) {
+            ReadPreference readPreference = ReadPreference.valueOf(transactional.readPreference());
+            txnOptions.readPreference(readPreference);
+        }
+
+        try (ClientSession clientSession = client.startSession()) {
+            MongoTransaction transaction = getMongoTransaction(clientSession);
+            // TODO put the transaction somewhere ...
+            clientSession.startTransaction(txnOptions.build());
+            ClienSessionHolder.setClientSession(clientSession);
```
FroMage, line 72: You don't need that: you can obtain it from the `TransactionManager`

extensions/panache/mongodb-panache/runtime/src/main/java/io/quarkus/mongodb/panache/transaction/interceptor/MongoTransactionalInterceptor.java
```diff
@@ -0,0 +1,136 @@
...
+    MongoClient client;
+
+    @Inject
+    TransactionSynchronizationRegistry tsr;
+
+    @Inject
+    TransactionManager transactionManager;
+
+    @AroundInvoke
+    public Object intercept(InvocationContext ic) throws Exception {
```
FroMage, line 44: At the very least you need to support the reactive return types with the same handling as in `TransactionalInterceptorBase.invokeInOurTx`, especially because you have a reactive client for mongo.

integration-tests/mongodb-panache/src/main/java/io/quarkus/it/mongodb/panache/person/PersonEntityResource.java
```diff
@@ -40,6 +41,15 @@ public void addPersons(List<PersonEntity> persons) {
         PersonEntity.persist(persons);
     }
 
+    // This endpoint is not called inside the test because transaction needs a replica set
```
FroMage, line 44: Really? That makes it pretty hard to test transactions. I don't see why either…

integration-tests/mongodb-panache/src/main/resources/application.properties
```diff
@@ -1,4 +1,4 @@
-quarkus.mongodb.connection-string=mongodb://localhost:27018
+quarkus.mongodb.connection-string=mongodb+srv://sarah:connor@cluster0-j57zm.gcp.mongodb.net/test?retryWrites=true&w=majority
```
FroMage, line 1: You sure about this?

FroMage (changes requested):
Really sorry about the late review. This looks very good and on the right path.

extensions/panache/mongodb-panache/runtime/src/main/java/io/quarkus/mongodb/panache/transaction/interceptor/MongoTransactionalInterceptor.java
loicmathieu: Reactive MongoDB with Panache didn' exist at the time I wrote this PR.
I'll take a look

integration-tests/mongodb-panache/src/main/java/io/quarkus/it/mongodb/panache/person/PersonEntityResource.java
loicmathieu: Transactions only works on a replicat set, MongoDB only provides distributed treansaction, that's why.
I'll investiguate a way to setup a cluster via Quakus TestResource support.

integration-tests/mongodb-panache/src/main/resources/application.properties
loicmathieu: I delete the cluster ...

extensions/panache/mongodb-panache/runtime/src/main/java/io/quarkus/mongodb/panache/transaction/interceptor/MongoTransaction.java
loicmathieu: Yes, done ;)

extensions/panache/mongodb-panache/runtime/src/main/java/io/quarkus/mongodb/panache/transaction/interceptor/MongoTransactionalInterceptor.java
loicmathieu: I try to obtain the transaction from the transaction manager and it shows that there is an issue somewhere: `java.lang.IllegalStateException: No transaction is running`.
I need to have a closer look at this ...

extensions/panache/mongodb-panache/runtime/src/main/java/io/quarkus/mongodb/panache/transaction/interceptor/ClienSessionHolder.java
loicmathieu: Added the missing `t` will remove it if I achieve to retrieve the session from the transaction manager.

extensions/panache/mongodb-panache/runtime/src/main/java/io/quarkus/mongodb/panache/transaction/interceptor/MongoTransactionalInterceptor.java
loicmathieu: I now get the transaction from the `TransactionSynchrinizationRegistry`.

loicmathieu:
@FroMage before I implement reactive support and add a dedicated automated test, can you check again my interceptor implementation ?
This is now working as expected I think, I remove the use of  `tsr.registerInterposedSynchronization();` as with it the transaction never ends so I call commit() and rollback() directly on the transaction manager.

extensions/panache/mongodb-panache/runtime/src/main/java/io/quarkus/mongodb/panache/transaction/interceptor/MongoTransaction.java
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    public ClientSession getClientSession() {
+        return clientSession;
+    }
+
+    @Override
+    public void commit() throws RollbackException, HeuristicMixedException, HeuristicRollbackException, SecurityException,
+            IllegalStateException, SystemException {
+        this.status = Status.STATUS_COMMITTING;
```
FroMage, line 29: Is it not up to this class to check the status and only commit if not already done, or not marked as rollback-only? Or is it the caller doing that?

FroMage (changes requested):
Looks better indeed, but I do wonder why you don't need that synchro point :-/

Also note you're still missing the reactive use-case.

extensions/panache/mongodb-panache/runtime/src/main/java/io/quarkus/mongodb/panache/transaction/interceptor/MongoTransaction.java
loicmathieu: Hard to say, the Narayana code is not easy to follow as there is multiple layer of manager and tx classes.
I made several attemps before finding the way to do it, I will check again without this but as far as I remember without this the status didn't change.
Maybe someone that made the Narayana integration inside Quarkus can have a look ?

loicmathieu: @FroMage in fact there is no `setStatus` method inside the transaction object so there is no way for an external system to update the status of the transaction. This needs to be done from the inside of the transaction object.

loicmathieu: I checked on the Narayana and the Infinispan transaction object, and both do the same so it's fine.

loicmathieu:
@FroMage I check on the existing transaction implementation and the `Synchronization` object needs to be called explicitly from the transaction object at commit()/rollback() calls. I can re-introduce it if you want.

loicmathieu:
@FroMage I add async/reactive support.

What still needs to be done are:

- Documentation
- Integration test with a MongoDB cluster
- Refactoring to avoid so much duplication with the interceptor for JDBC transaction

extensions/panache/mongodb-panache/runtime/src/main/java/io/quarkus/mongodb/panache/transaction/interceptor/MongoTransaction.java
FroMage: OK.

extensions/panache/mongodb-panache/runtime/src/main/java/io/quarkus/mongodb/panache/transaction/interceptor/MongoTransaction.java
```diff
@@ -0,0 +1,93 @@
...
+    public void registerSynchronization(Synchronization sync) throws RollbackException, IllegalStateException, SystemException {
+    }
+
+    @Override
+    public void rollback() throws IllegalStateException, SystemException {
+        if (isDone()) {
+            throw new IllegalStateException("Transaction is done. Cannot rollback transaction.");
+        }
+
+        System.out.println("MongoTransaction.rollback()");
```
FroMage, line 64: Traces left.

extensions/panache/mongodb-panache/runtime/src/main/java/io/quarkus/mongodb/panache/transaction/interceptor/MongoTransaction.java
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    @Override
+    public void commit() throws RollbackException, HeuristicMixedException, HeuristicRollbackException, SecurityException,
+            IllegalStateException, SystemException {
+        if (isDone()) {
+            throw new IllegalStateException("Transaction is done. Cannot commit transaction.");
+        }
+
+        System.out.println("MongoTransaction.commit()");
```
FroMage, line 33: Traces left.

FroMage (changes requested):
Looks good. It would be good indeed to share the interceptor common code, not sure how easy that would be, though.

integration-tests/mongodb-panache/src/main/java/io/quarkus/it/mongodb/panache/person/PersonEntityResource.java
loicmathieu: I setup a replica set for the integration test. I copied the existing TestResource from the `mongob-client` extension.
Maybe I need to create an `quarkus-test-mongodb` module inside the `test-framework` directory and put the MongoDB related test resource in it to avoid duplication ?

extensions/panache/mongodb-panache/runtime/src/main/java/io/quarkus/mongodb/panache/transaction/interceptor/MongoTransactionalInterceptor.java
loicmathieu: Done

extensions/panache/mongodb-panache/runtime/src/main/java/io/quarkus/mongodb/panache/transaction/interceptor/MongoTransaction.java
loicmathieu: removed

extensions/panache/mongodb-panache/runtime/src/main/java/io/quarkus/mongodb/panache/transaction/interceptor/MongoTransaction.java
loicmathieu: removed

loicmathieu:
@FroMage I review the implementation and add a test. It is now perfectly working ;)

The reactive story is not done, you can use reactive types with the imperative MongoDB driver, but not with the reactive one. The session class is different between the two so I fear I will maybe need a different interceptor bindings because there is no way to know inside the interceptor wich type of client we use ...

Now that we use Narayana, there is two `@Transactional` annotation in the classpath. I think it will be good to rename the one for MongoDB to `@MongoTransactional` to avoid issues.

loicmathieu:
I added the reactive implementation.
I choose to provides two annotations and two interceptor as there is no way to know if we are using the imperative or the reactive client.

Still some work to do: test is not working for reactive and a refactoring is needed to avoid too much code duplication.

emmanuelbernard:
hey @loicmathieu , nice work as usual to push the boundaries.
It's funny you added the `@Mongo*Transaction` annotations. I have been exploring a similar subject for Hibernate Reactive but wanted to reuse the existing `@Transactional` annotation.

Do you mind starting a thread on quarkus-dev to discuss why you went for several annotations (and an interceptor for the reactive one). I have some ideas that I'll be able to pile on. And quarkus-dev because it will need input for quite a few folks from core, narayana, hibernate reactive, vert.x etc.

emmanuelbernard:
Name the thread `Quarkus and Transactions in reactive and for non RDBMS`

loicmathieu:
@emmanuelbernard I'll try to make a summary of what I did and why, I also think we need a look from folks from narayana before merging this one anyway.

extensions/panache/mongodb-panache/runtime/src/main/java/io/quarkus/mongodb/panache/transaction/interceptor/ReactiveMongoTransaction.java
```diff
@@ -0,0 +1,98 @@
...
+    @Override
+    public void commit() throws RollbackException, HeuristicMixedException, HeuristicRollbackException, SecurityException,
+            IllegalStateException, SystemException {
+        if (isDone()) {
+            throw new IllegalStateException("Transaction is done. Cannot commit transaction.");
+        }
+
+        this.status = Status.STATUS_COMMITTING;
+        try {
+            this.clientSession.commitTransaction();
```
FroMage, line 35: Note that this returns a `Publisher<Void>` (https://mongodb.github.io/mongo-java-driver/4.1/apidocs/mongodb-driver-reactivestreams/com/mongodb/reactivestreams/client/ClientSession.html#commitTransaction()) and so it turns out that we can't implement this with the JTA transaction API.

I didn't realise this until today, so let's continue the discussion on quarkus-dev as soon as you start the thread. It's likely we'll get rid of JTA for reactive transactions and implement it differently, but introducing an abstraction for both Hibernate Reactive and Mongodb Reactive (if we can).

loicmathieu: Thanks, you just point me to why my entities are not persisted when using transaction, it's because the `commitTransaction` is async !

FroMage: :)

loicmathieu: I wait on the `commitTransacton` and `rollbackTransaction` for now on and this work fine.

There is now three places where I block, I add a comment `//FIXME we block here` as I think it's wrong to block on an IO inside the interceptor for reactive MongoDB.
I'll write someting to the ML tomorrow to discuss this.

extensions/panache/mongodb-panache/runtime/src/main/java/io/quarkus/mongodb/panache/transaction/MongoTransactional.java
```diff
@@ -0,0 +1,42 @@
...
+@Inherited
+@InterceptorBinding
+@Target({ ElementType.TYPE, ElementType.METHOD })
+@Retention(RetentionPolicy.RUNTIME)
+public @interface MongoTransactional {
+
+    /**
+     * Read preference for this transaction defined as a String.
+     * ReadPreference#valueOf(String) will be used to retrieve the write concern.
+     * If not specified, the default from the underlying MongoClient will be used.
```
emmanuelbernard, line 25: This JavaDoc would be more useful if you were listing the various read preference options and what they mean. Same for the other ones below

loicmathieu:
Superseded by #14002

PR #7223, opened 2020-02-17
Add release-related metadata to ide-config pom.xml

gsmet:


gastaldi (approved):

gsmet:
FWIW, I had to push it this morning to push the release so it has already been tested in the 1.3 branch (that I will drop for CR1). But we need to push this in master too.

gastaldi:
There is no need to stress CI for that

PR #7225, opened 2020-02-17
Artifacts may not contain the version

### LINKED ISSUE #7217: Quarkus gradle build fails with flatDir
**Describe the bug**
I'm using gradle to build quarkus project.
I've got jar dependencies that are in my projects 'lib' folder that I need to include as dependencies.
When trying to build the project it fails

**Expected behavior**
It should build and start

**Actual behavior**
Build fails without any clear reason.

**To Reproduce**
1.  Create a sample quarkus project with gradle as a build tool
2.  Create a lib folder in your project, add some jar
(I have downloaded the appache commons lang3)

3.  Add a flat dir repository to your build.gradle file, like:
flatDir {
        name = 'Local lib'
        dir file('lib')
    }
4.  Add a dependency in your project like:
implementation ':commons-lang3-3.9'

**Note**
I ran the quarkusBuild target with **--stacktrace** and it gives this error:

`Caused by: java.lang.ArrayIndexOutOfBoundsException: 2
	at io.quarkus.gradle.AppModelGradleResolver.toAppDependency(AppModelGradleResolver.java:194)
	at io.quarkus.gradle.AppModelGradleResolver.resolveModel(AppModelGradleResolver.java:134)`

So I think it must be an issue with the quarkus gradle plugin.

gastaldi:
Fixes #7217

### DIFF
--- a/devtools/gradle/build.gradle
+++ b/devtools/gradle/build.gradle
@@ -26,6 +26,7 @@ dependencies {
     implementation "io.quarkus:quarkus-platform-descriptor-resolver-json:${version}"
     implementation "io.quarkus:quarkus-development-mode:${version}"
 
+    testImplementation 'org.mockito:mockito-core:3.2.4'
     testImplementation 'org.assertj:assertj-core:3.14.0'
     testImplementation 'org.junit.jupiter:junit-jupiter-api:5.5.2'
     testImplementation 'org.junit.jupiter:junit-jupiter-params:5.5.2'
--- a/devtools/gradle/src/main/java/io/quarkus/gradle/AppModelGradleResolver.java
+++ b/devtools/gradle/src/main/java/io/quarkus/gradle/AppModelGradleResolver.java
@@ -140,7 +140,7 @@ public class AppModelGradleResolver implements AppModelResolver {
         final List<AppDependency> fullDeploymentDeps = new ArrayList<>();
         if (!extensionDeps.isEmpty()) {
             final Configuration deploymentConfig = project.getConfigurations()
-                    .detachedConfiguration(extensionDeps.toArray(new Dependency[extensionDeps.size()]));
+                    .detachedConfiguration(extensionDeps.toArray(new Dependency[0]));
             final ResolvedConfiguration rc = deploymentConfig.getResolvedConfiguration();
             for (ResolvedArtifact a : rc.getResolvedArtifacts()) {
                 final ModuleVersionIdentifier userVersion = userModules.get(getModuleId(a));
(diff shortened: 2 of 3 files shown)


devtools/gradle/build.gradle
```diff
@@ -26,6 +26,7 @@ dependencies {
     implementation "io.quarkus:quarkus-platform-descriptor-resolver-json:${version}"
     implementation "io.quarkus:quarkus-development-mode:${version}"
 
+    testImplementation 'org.mockito:mockito-core:3.2.4'
```
geoand: Is this related to the purpose of the PR?

gastaldi: This is for the test. There are too many nested objects in Gradle :)

geoand:
Seems reasonable to me, but I'll leave the final OK to @aloubyansky

devtools/gradle/src/main/java/io/quarkus/gradle/AppModelGradleResolver.java
```diff
@@ -140,7 +140,7 @@ public AppModel resolveModel(AppArtifact appArtifact) throws AppModelResolverExc
         final List<AppDependency> fullDeploymentDeps = new ArrayList<>();
         if (!extensionDeps.isEmpty()) {
             final Configuration deploymentConfig = project.getConfigurations()
-                    .detachedConfiguration(extensionDeps.toArray(new Dependency[extensionDeps.size()]));
```
aloubyansky: Just curious why this line was replaced?

geoand: I assume because of: https://shipilev.net/blog/2016/arrays-wisdom-ancients/#_conclusion

devtools/gradle/build.gradle
aloubyansky: Perhaps not as part of this PR, I think we should extract specific versions to the properties file (`gradle.properties`) that would also make it easier to verify/identify differences between the maven and gradle projects. We could also generate the properties file from a maven plugin, as an option.

devtools/gradle/src/main/java/io/quarkus/gradle/AppModelGradleResolver.java
gastaldi: IntellIJ was complaining about it. Apparently this is not an issue anymore in 2020:


>Call to 'toArray()' with pre-sized array argument 'new Dependency[extensionDeps.size()]' 
> Inspection info: There are two styles to convert a collection to an array: either using a pre-sized array (like c.toArray(new String[c.size()])) or using an empty array (like c.toArray(new String[0]).
>In older Java versions using pre-sized array was recommended, as the reflection call which is necessary to create an array of proper size was quite slow. However since late updates of OpenJDK 6 this call was intrinsified, making the performance of the empty array version the same and sometimes even better, compared to the pre-sized version. Also passing pre-sized array is dangerous for a concurrent or synchronized collection as a data race is possible between the size and toArray call which may result in extra nulls at the end of the array, if the collection was concurrently shrunk during the operation.
This inspection allows to follow the uniform style: either using an empty array (which is recommended in modern Java) or using a pre-sized array (which might be faster in older Java versions or non-HotSpot based JVMs).

aloubyansky (approved):

gastaldi:
CI errors are not related. Merging

Issue #7233, opened 2020-02-17
Add access to the entityManager from a JPA PanacheEntity / Repository


emmanuelbernard:
In a `PanacheEntity` entity or in `PanacheRepository`, I should be able to access the entitymanager (available via `JpaOperations.getEntityManager()` internally).

probably

```
public EntityManager entityManager() {
    return JpaOperations.getEntityManager();
}
```

nimo23:
Related to https://github.com/quarkusio/quarkus/issues/6080.

loicmathieu:
For the record, the entityManager should be accessible from the base classes: `PanacheEntityBase` and `PanacheRepositoryBase`.

PR #7236, opened 2020-02-17
Add access to the entityManager from a JPA PanacheEntity / Repository #7233

italogustavoborba:
#7233

### DIFF
--- a/extensions/panache/hibernate-orm-panache/runtime/src/main/java/io/quarkus/hibernate/orm/panache/PanacheEntity.java
+++ b/extensions/panache/hibernate-orm-panache/runtime/src/main/java/io/quarkus/hibernate/orm/panache/PanacheEntity.java
@@ -1,9 +1,12 @@
 package io.quarkus.hibernate.orm.panache;
 
+import javax.persistence.EntityManager;
 import javax.persistence.GeneratedValue;
 import javax.persistence.Id;
 import javax.persistence.MappedSuperclass;
 import javax.persistence.Transient;
+
+import io.quarkus.hibernate.orm.panache.runtime.JpaOperations;
 
 /**
  * <p>
@@ -38,6 +41,10 @@ public abstract class PanacheEntity extends PanacheEntityBase {
     public String toString() {
         return this.getClass().getSimpleName() + "<" + id + ">";
     }
+    
+    public EntityManager entityManager() {
+	    return JpaOperations.getEntityManager();
+	}
 
     // FIXME: VERSION?
 }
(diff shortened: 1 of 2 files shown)


FroMage:
LGTM, could you just add javadocs to the methods please?

emmanuelbernard:
should we have a line or two in the Panache Guide?

FroMage:
Not sure it's worth it, I think most people will find it with completion.

nimo23:
Why not putting 

```
EntityManager entityManager() {
    return JpaOperations.getEntityManager();
}
```

within `PanacheEntityBase` (and `PanacheRepositoryBase`) instead of its child classes? With the actual version the `entityManager()` is not available when extending those classes directly because of, for example:

> If you want a custom ID type or strategy, you can directly extend {@link PanacheEntityBase}
> instead, and write your own ID field.

extensions/panache/hibernate-orm-panache/runtime/src/main/java/io/quarkus/hibernate/orm/panache/PanacheEntity.java
```diff
@@ -38,6 +41,10 @@
     public String toString() {
         return this.getClass().getSimpleName() + "<" + id + ">";
     }
+    
+    public EntityManager entityManager() {
```
nimo23: Why not put this method within PanacheEntityBase instead?

extensions/panache/hibernate-orm-panache/runtime/src/main/java/io/quarkus/hibernate/orm/panache/PanacheRepository.java
```diff
@@ -13,4 +17,7 @@
  */
 public interface PanacheRepository<Entity> extends PanacheRepositoryBase<Entity, Long> {
 
+	default EntityManager entityManager() {
```
nimo23: Why not put this method within PanacheRepositoryBase instead?

emmanuelbernard:
Ah yes @nimo23 is right, the entityManager method should in on `PanacheEntityBase` and `PanacheRepositoryBase` instead. Can you do that change @italogustavoborba ?

emmanuelbernard:
A small test would be nice too BTW :)

loicmathieu:
@italogustavoborba can you provides the methods inside the base class so we can merge this PR ?

gastaldi:
What's the status of this PR? @italogustavoborba can you perform the requested changes above?

gastaldi:
Superseded by https://github.com/quarkusio/quarkus/pull/10541, closing now

PR #7245, opened 2020-02-18
Recreate all metric registries during reload

### LINKED ISSUE #7018: Recreate all metric registries on startup?

jmartisk:
Fixes https://github.com/quarkusio/quarkus/issues/7018

### DIFF
--- a/extensions/jaeger/runtime/src/main/java/io/quarkus/jaeger/runtime/QuarkusJaegerMetricsFactory.java
+++ b/extensions/jaeger/runtime/src/main/java/io/quarkus/jaeger/runtime/QuarkusJaegerMetricsFactory.java
@@ -21,9 +21,6 @@ public class QuarkusJaegerMetricsFactory implements MetricsFactory {
 
     @Override
     public Counter createCounter(final String name, final Map<String, String> tags) {
-        // On restart (e.g. in dev mode) Jaeger registers
-        // the same metrics which would result in exception
-        registry.remove(name);
         org.eclipse.microprofile.metrics.Counter counter = registry.counter(meta(name, MetricType.COUNTER), toTagArray(tags));
 
         return new Counter() {
@@ -36,7 +33,6 @@ public class QuarkusJaegerMetricsFactory implements MetricsFactory {
 
     @Override
     public Timer createTimer(final String name, final Map<String, String> tags) {
-        registry.remove(name);
         org.eclipse.microprofile.metrics.Timer timer = registry.timer(meta(name, MetricType.TIMER), toTagArray(tags));
 
         return new Timer() {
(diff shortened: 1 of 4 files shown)


geoand:
I like it @jmartisk!
Any chance we can get a dev mode test for this? I am not really concerned about it working - I'm sure it does, I just want to make sure some other change irrelevant doesn't brake it in the future

jmartisk:
Hey @geoand,  `DevModeMetricsTest` basically already verifies this for application metrics, this PR just extends the behavior to other metric scopes.
Given that this is meant for `base` and `vendor` metrics, which are managed by Quarkus extensions, it would make sense to have a test where an extension registers a metric, but I'm not sure how to do that because `QuarkusDevModeTest` does not seem to support adding build chain customizers (unlike `QuarkusUnitTest`), so I think we can't right now create a dev mode test that adds a dynamically added "extension" :(
So the test would have to register a `base`/`vendor` metric within application code, which is not really correct usage, so adding such test sounds awkward.

geoand:
@jmartisk in that case you would probably have to add one of ugly test like this: https://github.com/quarkusio/quarkus/blob/master/integration-tests/maven/src/test/java/io/quarkus/maven/it/DevMojoIT.java

jmartisk:
That looks quite awful for something that would only verify that `base` and `vendor` scopes get cleaned up too :)

geoand:
I'll leave it up to you to decide, but I personally for having something that we'll keep potentialy erroneous future refactorings in check :)

jmartisk:
I added a tiny test that creates base/vendor metrics in application code, it's not how it's supposed to be done in reality, but the test will still verify what it needs to verify, so I suppose it's a valid test

geoand:
Cool, thanks

geoand (approved):

jmartisk:
There were some unrelated CI failures, I rebased it and let's try again

PR #7253, opened 2020-02-18
Use Maven 3.6.3 in new projects

gastaldi:


geoand:
I think there are 3 instances of `3.6.2` in the code base and all should probably be updates

aloubyansky (approved):
We should also upgrade https://github.com/quarkusio/quarkus/blob/master/bom/runtime/pom.xml#L70 to 3.6.3 or something. We don't really have to be using the minimal supported version ourselves.

gastaldi:
@aloubyansky I agree. I'll try to provide another PR bumping that

@geoand I believe I replaced all the cases, the other instances seems to be generated from these properties I think

geoand:
Indeed the 3rd one does seem to be in a generated file

geoand (approved):

gsmet (dismissed):
Question: is the Maven wrapper used by Quarkus itself 3.6.3? I think we should be consistent as we run the tests with it.

So I would also upgrade the wrapper if not already 3.6.3. Feel free to dismiss this review if it already is.

gastaldi:
@gsmet good catch. I updated the wrapper to comply with Maven 3.6.3 too.

gsmet:
Ah ah, so I killed two birds with one stone because I was thinking of the wrapper we have at the root of the Quarkus project :).

gsmet:
@gastaldi if you upgrade the maven-core properties, please also upgrade the wrapper used for Quarkus itself here: https://github.com/quarkusio/quarkus/blob/master/.mvn/wrapper/maven-wrapper.properties .

PR #7264, opened 2020-02-18
Remove redundant synchronized block

pivovarit:
Looks like an old attempt at double-checked locking got synchronized at method level anyway, or am I missing something?

Removing synchronizedception


stuartwdouglas (approved):
Looks like when I was going through fixing some thread safety issues I did not notices that it already had method level sync.

PR #7282, opened 2020-02-19
Introduced quarkus-internal.maven-cmd-line-args system property

aloubyansky:
to be able to pass Maven CLI args as a system property to the MavenRepoInitializer utility


gastaldi (approved):

aloubyansky:
@gsmet what should we do with this one?

PR #7285, opened 2020-02-19
Add oidc-code-flow dev mode test

### LINKED ISSUE #6965: Test that quarkus-oidc works in the dev mode
**Description**
Verify that changing some of the OIDC properties is effective in the dev mode.

sberyozkin:
Fixes #6965.

This test is like the tests in the `oidc-code-flow`. The difference is that it starts with the invalid `client_id` (`client-dev` as opposed to `client-dev-mode` created in Keycloak). As such the exception is expected at the beginning of the test. Next the `client_id` is updated and the test then passes as expected.

### DIFF
--- a/extensions/oidc/deployment/pom.xml
+++ b/extensions/oidc/deployment/pom.xml
@@ -13,6 +13,10 @@
     <artifactId>quarkus-oidc-deployment</artifactId>
     <name>Quarkus - OpenID Connect Adapter - Deployment</name>
 
+    <properties>
+        <keycloak.url>http://localhost:8180/auth</keycloak.url>
+    </properties>
+
     <dependencies>
         <dependency>
             <groupId>io.quarkus</groupId>
@@ -53,9 +57,35 @@
             <artifactId>rest-assured</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.keycloak</groupId>
+            <artifactId>keycloak-adapter-core</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.keycloak</groupId>
+            <artifactId>keycloak-core</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>net.sourceforge.htmlunit</groupId>
+            <artifactId>htmlunit</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-resteasy-jackson</artifactId>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 
     <build>
+        <testResources>
+            <testResource>
+                <directory>src/test/resources</directory>
+                <filtering>true</filtering>
+            </testResource>
+        </testResources>
         <plugins>
             <plugin>
                 <artifactId>maven-compiler-plugin</artifactId>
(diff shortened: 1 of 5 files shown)


sberyozkin:
It does not run in a native mode which is expected for the devmode tests

gastaldi (approved):
LGTM, but I'd rather wait for @pedroigor to approve before merging ;)

sberyozkin:
@gastaldi Hi, thanks; sure, lets wait for Pedro and/or Stuart to check

stuartwdouglas (changes requested):
This should really be in oidc/deployment, dev mode tests should not get their own 'integration-tests' module (we already have hundreds of modules, so we need to resist adding any that are not needed).

sberyozkin:
Hi @stuartwdouglas 
The reason I've added it there is that it requires a docker starting KC. I've tried initially allocate it in the `integration/oidc-code-flow` but it was interfering with the main `CodeFlowTest` (KC was failing to create the realms).
If I move it to `extensions/oidc/deployment`, will it be actually tested in CI (it depends on `docker-keycloak` and `test-keycloak` profiles).
CC @gsmet

stuartwdouglas:
Yes, it will still be tested, keycloak is still started for the JVM mode tests.

sberyozkin:
Hi @stuartwdouglas, how does it look now ?

sberyozkin:
`http://localhost:8180` is not available in CI. How to identify the problem ? Looks lie Keycloak has not completed its own initialization yet by the time the test starts contacting Keycloak.

sberyozkin:
@stuartwdouglas The same test works in `integration-tests/`, so I wonder if there is some timing issue.

stuartwdouglas:
It looks like you have fixed the timing issue?

extensions/oidc/deployment/pom.xml
```diff
@@ -69,7 +99,130 @@
...
                 </configuration>
             </plugin>
+            <plugin>
+                <artifactId>maven-surefire-plugin</artifactId>
+                <configuration>
+                    <skip>true</skip>
+                </configuration>
+            </plugin>
+            <plugin>
+                <artifactId>maven-failsafe-plugin</artifactId>
```
stuartwdouglas: Failsafe is not used here, as there are no native image tests

extensions/oidc/deployment/pom.xml
```diff
@@ -69,7 +99,130 @@
...
+                        <artifactId>maven-surefire-plugin</artifactId>
+                        <configuration>
+                            <skip>false</skip>
+                            <systemPropertyVariables>
+                                <keycloak.url>${keycloak.url}</keycloak.url>
+                            </systemPropertyVariables>
+                        </configuration>
+                    </plugin>
+                    <plugin>
+                        <artifactId>maven-failsafe-plugin</artifactId>
```
stuartwdouglas: same here

extensions/oidc/deployment/pom.xml
```diff
@@ -69,7 +99,130 @@
...
+                        <configuration>
+                            <skip>false</skip>
+                            <systemPropertyVariables>
+                                <keycloak.url>${keycloak.url}</keycloak.url>
+                            </systemPropertyVariables>
+                        </configuration>
+                    </plugin>
+                    <plugin>
+                        <groupId>io.quarkus</groupId>
```
stuartwdouglas: This should not be here.

stuartwdouglas (changes requested):

sberyozkin:
Hi @stuartwdouglas I've only added a `keycloak.url` maven property and the resource filtering as used in the integration tests, but the main 'fix' was about rebasing and force-pushing repeatedly :-).
Let me take care of the new comments and see what happens :-)

stuartwdouglas:
Does that mean it it likely intermittently failing, and just happened to pass this time?

sberyozkin:
Windows build failure is unrelated: ` Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:2.22.1:test (default-test) on project quarkus-resteasy-deployment: There are test failures.`

sberyozkin:
@stuartwdouglas It has passed again, so please merge if all looks good to you

sberyozkin:
@stuartwdouglas It has passed again, so please merge if all looks good to you

stuartwdouglas (approved):

PR #7288, opened 2020-02-19
Introduce built-in number operations in Qute extension

### LINKED ISSUE #7283: qute: modulo operation in iterations
**Description**
`
{#for printable in printables}
<div>{printable.name}</div>   
{#if count % 5 == 0 }
           <div class="clearfix"></div>
   {/if}
{/for}
`

gastaldi:
Fixes #7283

### DIFF
--- a/extensions/qute/runtime/src/main/java/io/quarkus/qute/runtime/extensions/CollectionTemplateExtensions.java
+++ b/extensions/qute/runtime/src/main/java/io/quarkus/qute/runtime/extensions/CollectionTemplateExtensions.java
@@ -1,0 +1,13 @@
+package io.quarkus.qute.runtime.extensions;
+
+import java.util.List;
+
+import io.quarkus.qute.TemplateExtension;
+
+@TemplateExtension
+public class CollectionTemplateExtensions {
+
+    static Object get(List<?> list, int index) {
+        return list.get(index);
+    }
+}
--- a/docs/src/main/asciidoc/qute-reference.adoc
+++ b/docs/src/main/asciidoc/qute-reference.adoc
@@ -694,6 +694,37 @@ class MyExtensions {
 ----
 <1> `item.discountedPrice` is resolved to an instance of `BigDecimal`.
 
+==== Built-in Template Extension methods
+
+Quarkus provides a set of built-in extension methods.
+
+===== Map extension methods:
+
+* `keys` or `keySet`: Returns a Set view of the keys contained in a map
+** `{#for key in map.keySet}`
+
+* `values`: Returns a Collection view of the values contained in a map
+** `{#for value in map.values}`
+
+* `size`: Returns the number of key-value mappings in a map
+** `{map.size}`
+
+* `isEmpty`: Returns true if a map contains no key-value mappings
+** `{#if map.isEmpty}`
+
+* `get(key)`: Returns the value to which the specified key is mapped
+** `{map.get('foo')}` or `{map['foo']}`
+
+===== Collection extension methods:
+
+* `get(index)`: Returns the element at the specified position in a list
+** `{list.get(0)}` or `{list[0]}`
+
+===== Number extension methods:
+
+* `mod`: Modulo operation
+** `{#if counter.mod(5) == 0}`
+
 === @TemplateData
 
 A value resolver is automatically generated for a type annotated with `@TemplateData`.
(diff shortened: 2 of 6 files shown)


gsmet:
I let Martin validate the approach but if we go with this, we need a documentation update.

gastaldi:
I updated the documentation to reflect the changes

gastaldi:
Seems that CI is not being triggered. Let me close and reopen the PR just in case

gsmet:
@mkouba can you check that one, see if that's what we want to do?

mkouba:
@gsmet I'm pretty sure I added some comments yesterday but they're gone :-(. I'm going to add them again.

extensions/qute/runtime/src/main/java/io/quarkus/qute/runtime/extensions/CollectionTemplateExtensions.java
```diff
@@ -9,7 +9,7 @@
 import io.quarkus.qute.TemplateExtension;
 
 @TemplateExtension
-public class BuiltinTemplateExtensions {
+public class CollectionTemplateExtensions {
```
mkouba: `Map` is not a collection so we should probably extract the relevant methods in `MapTemplateExtensions`.

extensions/qute/runtime/src/main/java/io/quarkus/qute/runtime/extensions/IntegerTemplateExtensions.java
```diff
@@ -0,0 +1,12 @@
+package io.quarkus.qute.runtime.extensions;
+
+import io.quarkus.qute.TemplateExtension;
+
+@TemplateExtension
+public class IntegerTemplateExtensions {
```
mkouba: Maybe we should use a more generic name, ie. `NumberTemplateExtensions`?

mkouba (changes requested):
The approach is probably OK. At least for now. We should re-evaluate the builtins again once the API is stabilized enough.

docs/src/main/asciidoc/qute.adoc
```diff
@@ -69,13 +71,16 @@ public class HelloResource {
     }
 }
 ----
-<1> If there is no `@ResourcePath` qualifier provided, the field name is used to locate the template. In this particular case, we're injecting a template with path `templates/hello.txt`.
-<2> `Template.data()` returns a new template instance that can be customized before the actual rendering is triggered. In this case, we put the name value under the key `name`. The data map is accessible during rendering. 
+<1> If there is no `@ResourcePath` qualifier provided, the field name is used to locate the template.
+In this particular case, we're injecting a template with path `templates/hello.txt`.
```
mkouba: Is it legal to use line breaks in source code callouts? I thought that you always need to use `+` here...

gastaldi: Hm, not sure why this part was changed, maybe IntelliJ did it when I saved the doc

docs/src/main/asciidoc/qute.adoc
```diff
@@ -171,6 +171,37 @@ public class ItemResource {
...
 <3> A static template extension method can be used to add "computed properties" to a data class. The class of the first parameter is used to match the base object and the method name is used to match the property name.
 
+== Built-in Template Extension methods
+
+Quarkus provides a set of built-in extension methods.
+
+=== Map extension methods:
+
+* `keys()` or `keySet()`: Returns a Set view of the keys contained in a map
+** `{#for key in map.keySet()}`
```
mkouba: The brackets are optional and I'd say that `map.keySet` is more readable. The same applies to all snippets.

docs/src/main/asciidoc/qute.adoc
```diff
@@ -171,6 +171,37 @@ public class ItemResource {
 <2> Make the `Item` object accessible in the template.
 <3> A static template extension method can be used to add "computed properties" to a data class. The class of the first parameter is used to match the base object and the method name is used to match the property name.
 
+== Built-in Template Extension methods
```
mkouba: I think that we should move this section to the reference guide instead: https://github.com/quarkusio/quarkus/blob/master/docs/src/main/asciidoc/qute-reference.adoc

mkouba (changes requested):

docs/src/main/asciidoc/qute.adoc
gastaldi: The Built in template methods work only if you use the `qute` extension in Quarkus, they are not available if you use Qute standalone, that's why I kept it in `qute.adoc`, but I am fine to move it if you think it's better

mkouba: `qute.adoc` is a simple quickstart guide, `qute-reference.adoc` is a reference guide that contains both the core/standalone and quarkus-specific features.

gastaldi:
@mkouba Changes pushed

gsmet:
@mkouba should we merge that one?

mkouba (approved):

PR #7294, opened 2020-02-19
Add Vault write and delete secret Fixes #7155

### LINKED ISSUE #7155: Vault add writeSecret in VaultKVSecretEngine
**Description**
Currently `VaultKVSecretEngine` allow only to read a secret present in Vault.

An application may need to write secret.

**Implementation ideas**

In the interface `VaultKVSecretEngine` define a method to `writeSecret` like this one:

```java
writeSecret(String path, Map<String, String> secret);
```

In `VaultKvManager` the implementation could be:

```java
    @Override
    public void writeSecret(String path, Map<String, String> secret) {

        String clientToken = vaultAuthManager.getClientToken();
        String mount = serverConfig.kvSecretEngineMountPath;

        if (serverConfig.kvSecretEngineVersion == 1) {
            vaultClient.writeSecretV1(clientToken, mount, path, secret);
        } else {
            vaultClient.writeSecretV2(clientToken, mount, path, secret);
        }
    }
```

In the interface `VaultClient` we could add theses two methods to handle writing secret depending on secret version:

```java
void writeSecretV1(String token,
                              String secretEnginePath,
                              String path,
                              Map<String, String> secret);

void writeSecretV2(String token,
                              String secretEnginePath,
                              String path,
                              Map<String, String> secret);
```

In `OkHttpVaultClient` the implementation could be:

```java
@Override
public void writeSecretV1(String token,
                              String secretEnginePath,
                              String path,
                              Map<String, String> secret) {
        Request request = builder(secretEnginePath + "/" + path, token).post(requestBody(secret)).build();
        try (Response response = client.newCall(request).execute()) {
            if (response.code() != 204) {
                throwVaultException(response);
            }
        }
}

@Override
public void writeSecretV2(String token,
                              String secretEnginePath,
                              String path,
                              Map<String, String> secret) {
        Request request = builder(secretEnginePath + "/data/" + path, token).post(requestBody(secret)).build();
        try (Response response = client.newCall(request).execute()) {
            if (response.code() != 204) {
                throwVaultException(response);
            }
        }
}
```
**Remark**: We cannot use `exec` method as the response from Vault is an empty 204 and so the response cannot be deserialized using jackson to an Object.

vsevel:
Fixes https://github.com/quarkusio/quarkus/issues/7155 plus provides deleteSecret.
readSecret has been changed also to return null if the path does not exist (rather than throwing a 404)

### DIFF
--- a/extensions/vault/runtime/src/main/java/io/quarkus/vault/runtime/client/OkHttpVaultClient.java
+++ b/extensions/vault/runtime/src/main/java/io/quarkus/vault/runtime/client/OkHttpVaultClient.java
@@ -6,6 +6,7 @@ import static io.quarkus.vault.runtime.client.OkHttpClientFactory.createHttpClie
 import java.io.IOException;
 import java.net.MalformedURLException;
 import java.net.URL;
+import java.util.Map;
 
 import org.jboss.logging.Logger;
 
@@ -25,6 +26,8 @@ import io.quarkus.vault.runtime.client.dto.auth.VaultUserPassAuthBody;
 import io.quarkus.vault.runtime.client.dto.database.VaultDatabaseCredentials;
 import io.quarkus.vault.runtime.client.dto.kv.VaultKvSecretV1;
 import io.quarkus.vault.runtime.client.dto.kv.VaultKvSecretV2;
+import io.quarkus.vault.runtime.client.dto.kv.VaultKvSecretV2Write;
+import io.quarkus.vault.runtime.client.dto.kv.VaultKvSecretV2WriteBody;
 import io.quarkus.vault.runtime.client.dto.sys.VaultLeasesBody;
 import io.quarkus.vault.runtime.client.dto.sys.VaultLeasesLookup;
 import io.quarkus.vault.runtime.client.dto.sys.VaultRenewLease;
@@ -89,6 +92,26 @@ public class OkHttpVaultClient implements VaultClient {
     }
 
     @Override
+    public void writeSecretV1(String token, String secretEnginePath, String path, Map<String, String> secret) {
+        post(secretEnginePath + "/" + path, token, secret, null, 204);
+    }
+
+    @Override
+    public void writeSecretV2(String token, String secretEnginePath, String path, VaultKvSecretV2WriteBody body) {
+        post(secretEnginePath + "/data/" + path, token, body, VaultKvSecretV2Write.class);
+    }
+
+    @Override
+    public void deleteSecretV1(String token, String secretEnginePath, String path) {
+        delete(secretEnginePath + "/" + path, token, null, null, 204);
+    }
+
+    @Override
+    public void deleteSecretV2(String token, String secretEnginePath, String path) {
+        delete(secretEnginePath + "/data/" + path, token, null, null, 204);
+    }
+
+    @Override
     public VaultRenewSelf renewSelf(String token, String increment) {
         VaultRenewSelfBody body = new VaultRenewSelfBody(increment);
         return post("auth/token/renew-self", token, body, VaultRenewSelf.class);
@@ -145,6 +168,11 @@ public class OkHttpVaultClient implements VaultClient {
     }
 
     // ---
+
+    protected <T> T delete(String path, String token, Object body, Class<T> resultClass, int expectedCode) {
+        Request request = builder(path, token).delete(requestBody(body)).build();
+        return exec(request, resultClass, expectedCode);
+    }
 
     protected <T> T post(String path, String token, Object body, Class<T> resultClass, int expectedCode) {
         Request request = builder(path, token).post(requestBody(body)).build();
--- a/extensions/vault/runtime/src/main/java/io/quarkus/vault/VaultKVSecretEngine.java
+++ b/extensions/vault/runtime/src/main/java/io/quarkus/vault/VaultKVSecretEngine.java
@@ -20,4 +20,21 @@ public interface VaultKVSecretEngine {
      */
     Map<String, String> readSecret(String path);
 
+    /**
+     * Writes the secret at the given path. If the path does not exist, the secret will
+     * be created. If not the new secret will be merged with the existing one.
+     * 
+     * @param path in Vault, without the kv engine mount path
+     * @param secret to write at path
+     */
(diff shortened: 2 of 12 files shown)


sberyozkin:
Hey Vincent @vsevel, welcome back :-) Will deal with the review tomorrow, have to sign off now, cheers

extensions/vault/runtime/src/main/java/io/quarkus/vault/runtime/client/OkHttpVaultClient.java
```diff
@@ -146,33 +169,45 @@ public VaultTransitEncrypt rewrap(String token, String keyName, VaultTransitRewr
...
-        return exec(request, resultClass, 200);
+    private <T> T exec(Request request, Class<T> resultClass, boolean throwOn404) {
+        return exec(request, resultClass, 200, throwOn404);
     }
 
-    private <T> T exec(Request request, Class<T> resultClass, int expectedCode) {
+    private <T> T exec(Request request, Class<T> resultClass, int expectedCode, boolean throwOn404) {
         try (Response response = client.newCall(request).execute()) {
             if (response.code() != expectedCode) {
+                if (response.code() == 404 && !throwOn404) {
```
sberyozkin: @vsevel Why have you differentiated here ? Can we already have the users who have the code expecting the exceptions if nothing is available on a given path ?

vsevel: @sberyozkin historically, any code other than the expected code would yield an exception. that was the default behavior. as I was writing the tests for the crud use case, it felt that getting a 404 when getting a secret on a non existent path should return a null instead of an exception. so I kept the old behavior for all calls, except for readSecret(), where a null seemed more natural.
how would you do it? what is your suggestion?

gsmet (dismissed):
I will let @sberyozkin review that one but can you squash please?

extensions/vault/runtime/src/main/java/io/quarkus/vault/runtime/client/OkHttpVaultClient.java
sberyozkin: Hi @vsevel if `404` is returned when no path actually exists in the secret engine then it is an exception IMHO (yes, it would also change the expectations for the existing users without a strong reaspn); if it returns 404 when nothing is available at that existing path then `null` is fine; I propose to keep the old behavior for the moment (and perhaps revisit with a dedicated issue ?)

Thanks

vsevel: hello @sberyozkin not sure I completely followed you.
when you read secrets from vault, you get a 404 if the path does not exist, and the secret (which is a Map<String,String>) if it does. if not keys are added to an existing path, I did not test but I think it returns an empty map.

so today the behavior in quarkus is:
 - path does not exist => exception
 - path does exist => map (which might be empty)

in the PR, for the readSecret operation, I changed the behavior to be:
 - path does not exist => null
 - path does exist => map (which might be empty)

do you want me to revert to the old behavior?

sberyozkin: @vsevel Thanks; yes the current behavior looks correct to me, as with this PR's change the difference between the non-existent path and no secrets being available at the existing path becomes `null` vs `empty Map`. I feel an exception is warranted if the users has used the path which has not been configured. Please revert this one, As I said, we can have a comprehensive discussion about the way the Vault responses are treated (for all the supported engines) in a dedicated issue and change as needed :-)

vsevel: ok. I will revert to the old behavior and squash all commits. then you can approve if no other comments.
I will do this tonight.

>  I feel an exception is warranted if a user has used the path which has not been correctly configured

minor correction: we are not talking about using a path that has not been configured correctly, but rather getting a non existent path that is passed as an argument.

if vault was a java object, it would look like this:
```
Map<String, Properties> vault = new HashMap<>()

vault.put("path1", new Properties());

Properties props2 = new Properties();
props2.put("mykey", "myvalue");
vault.put("path2", props2);

vault.get("path1") => empty properties object
vault.get("path2") => properties object with mykey=myvalue
vault.get("foobar") => null
```

so getting a null looks right from a java api perspective.

sberyozkin: Hi Vincent @vsevel that definitely looks good when we have a local map :-). But not IMHO when the `vault` secret engine interface represent a remote Vault server. That `foobar` in your example, it can actually be set via a custom Quarkus property, but the user forgot that the path was in fact `barfoo` :-).
So yeah, please revert it and I'll be happy to merge, thanks

sberyozkin:
Hi Guillaume thanks, so then later on I'll lift your change request as Vincent is going to squash.

extensions/vault/runtime/src/main/java/io/quarkus/vault/runtime/client/OkHttpVaultClient.java
sberyozkin: Hi Vincent @vsevel I was think while I was offline if I was unreasonable in asking you to revert this change, but eventually came to the conclusion the way it currently works is good. 
Returning to your java code example, I should say that the API designers often use the empty List or Map for a pure convenience purposes, to let the users avoid checking for `null` :-), as opposed to overloading the response value to mark an exceptional situation, such that the users can write (they might already be writing it this way now):
```
try {
    return vault.get("path1").get("secret1Key");
} catch (VaultException ex) {}
```
The new code would leak NPE and result in 500.
So IMHO we will be fine with keeping your original code idea in place :-) thanks

vsevel: 404 changes have been reverted. I have squashed the commits. waiting for CI to go through. cheers.

sberyozkin (approved):
@vsevel Looks simple and good to me :-) Thanks a million (please consider a small doc PR as a follow up). Cheers

PR #7303, opened 2020-02-20
Implement Context Propagation support for Mutiny

cescoffier:
This PR enables the support for Context Propagation when using Mutiny.
It includes:

* a bit of functional code (reactive converters, a small change in the transactional filter...)
* a few tests
* ITs
* documentation update

### DIFF
--- a/extensions/narayana-jta/runtime/src/main/java/io/quarkus/narayana/jta/runtime/interceptor/TransactionalInterceptorBase.java
+++ b/extensions/narayana-jta/runtime/src/main/java/io/quarkus/narayana/jta/runtime/interceptor/TransactionalInterceptorBase.java
@@ -125,7 +125,7 @@ public abstract class TransactionalInterceptorBase implements Serializable {
             if (!throwing && ret != null) {
                 ReactiveTypeConverter<Object> converter = null;
                 if (ret instanceof CompletionStage == false
-                        && ret instanceof Publisher == false) {
+                        && (ret instanceof Publisher == false || ic.getMethod().getReturnType() != Publisher.class)) {
                     @SuppressWarnings({ "rawtypes", "unchecked" })
                     Optional<ReactiveTypeConverter<Object>> lookup = Registry.lookup((Class) ret.getClass());
                     if (lookup.isPresent()) {
--- a/bom/runtime/pom.xml
+++ b/bom/runtime/pom.xml
@@ -126,7 +126,7 @@
         <reactive-streams.version>1.0.3</reactive-streams.version>
         <test-containers.version>1.12.4</test-containers.version>
         <jboss-logging.version>3.3.2.Final</jboss-logging.version>
-        <mutiny.version>0.3.1</mutiny.version>
+        <mutiny.version>0.4.0</mutiny.version>
         <axle-client.version>0.0.12</axle-client.version>
         <mutiny-client.version>0.0.12</mutiny-client.version>
         <kafka2.version>2.4.0</kafka2.version>
--- a/bom/runtime/pom.xml
+++ b/bom/runtime/pom.xml
@@ -2506,6 +2506,11 @@
                 <artifactId>smallrye-reactive-converter-api</artifactId>
                 <version>${smallrye-converter-api.version}</version>
             </dependency>
+            <dependency>
+                <groupId>io.smallrye.reactive</groupId>
+                <artifactId>smallrye-reactive-converter-rxjava2</artifactId>
+                <version>${smallrye-converter-api.version}</version>
+            </dependency>
 
             <dependency>
                 <groupId>org.jetbrains.kotlin</groupId>
--- a/bom/runtime/pom.xml
+++ b/bom/runtime/pom.xml
@@ -996,6 +996,11 @@
                 <version>${mutiny.version}</version>
             </dependency>
             <dependency>
+                <groupId>io.smallrye.reactive</groupId>
+                <artifactId>mutiny-context-propagation</artifactId>
+                <version>${mutiny.version}</version>
+            </dependency>
+            <dependency>
                 <groupId>io.quarkus</groupId>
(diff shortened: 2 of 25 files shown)


extensions/narayana-jta/runtime/src/main/java/io/quarkus/narayana/jta/runtime/interceptor/TransactionalInterceptorBase.java
```diff
@@ -124,8 +124,7 @@ protected Object invokeInOurTx(InvocationContext ic, TransactionManager tm, Runn
             // handle asynchronously if not throwing
             if (!throwing && ret != null) {
                 ReactiveTypeConverter<Object> converter = null;
-                if (ret instanceof CompletionStage == false
-                        && ret instanceof Publisher == false) {
+                if (ret instanceof CompletionStage == false) {
```
FroMage: Why did you drop this clause? We don't have to convert it if it's a `Publisher`.

FroMage (changes requested):
LGTM except a weird removed line.

gsmet:
There is a test failure with:
```
Caused by: java.lang.NoClassDefFoundError: org/jboss/resteasy/core/ResteasyContext
	at io.smallrye.context.SmallRyeThreadContext.currentContextExecutor(SmallRyeThreadContext.java:208)
	at io.smallrye.mutiny.context.ContextPropagationUniInterceptor.onUniCreation(ContextPropagationUniInterceptor.java:47)
	at io.smallrye.mutiny.infrastructure.Infrastructure.onUniCreation(Infrastructure.java:70)
	at io.smallrye.mutiny.groups.UniOnItem.apply(UniOnItem.java:50)
	at io.quarkus.vertx.EventBusCodecTest.testWithUserCodecNonLocal(EventBusCodecTest.java:59)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at io.quarkus.test.QuarkusUnitTest.runExtensionMethod(QuarkusUnitTest.java:276)
```

extensions/narayana-jta/runtime/src/main/java/io/quarkus/narayana/jta/runtime/interceptor/TransactionalInterceptorBase.java
cescoffier: We actually do.
If we don't, then the following block does not work as expected (line 145):
```
ret = handleAsync(tm, tx, ic, ret, afterEndTransaction);
// convert back
if (converter != null)
       ret = converter.fromPublisher((Publisher<?>) ret);
```

Imagine we have a `Multi`. As `Multi` is an instance of `Publisher`, you don't look up for the converter (in the current code, before this PR). However, this Publisher gets _decorated_ in `handleAsync` which returns a `Publisher`. As we don't have a converter, it is not converted back to a `Multi`... boom `ClassCastException`.  It works with RX Java because the implementation of Reactive Streams Operators (used in `handleAsync`) returned a Flowable, but if you use a Flux or a Multi it does not work.

FroMage: Ah, damn. Well, OK. Perhaps we could check the intercepted method's return type and allow reusing the `Publisher` if its return type is `Publisher` (as opposed to `Multi`)?

cescoffier: That would remove the need to call the converter in this case. Something like:
```
if (ret instanceof CompletionStage == false && ! ret.getClass().equals(Publisher.class)) {
```

cescoffier: Fixed.

extensions/narayana-jta/runtime/src/main/java/io/quarkus/narayana/jta/runtime/interceptor/TransactionalInterceptorBase.java
```diff
@@ -124,8 +124,7 @@ protected Object invokeInOurTx(InvocationContext ic, TransactionManager tm, Runn
             // handle asynchronously if not throwing
             if (!throwing && ret != null) {
                 ReactiveTypeConverter<Object> converter = null;
-                if (ret instanceof CompletionStage == false
-                        && ret instanceof Publisher == false) {
+                if (ret instanceof CompletionStage == false && !ret.getClass().equals(Publisher.class)) {
```
FroMage: Shouldn't it be:

```java
if(ret instanceof CompletionStage == false
 && (ret instanceof Publisher == false || !ic.getMethod().getReturnType() == Publisher.class))
```

You need to check the method's declared return type, not the instance exact type, which can never be `Publisher` since it's an interface.
Note that I wonder if we don't have the same issue with `CompletionStage` and `CompletableFuture`. Could you add a test to see what happens if an intercepted method is declared to return a `CompletableFuture`?

cescoffier: Adding the test right now.

cescoffier: I fixed the condition.

FroMage (approved):
:shipit:

Issue #6286, opened 2019-12-20
integration-tests/logging-gelf are not executed in CI runs, risk for regression


rsvoboda:
Running integration-tests/logging-gelf tests involves manual steps
https://github.com/quarkusio/quarkus/tree/master/integration-tests/logging-gelf#running-the-tests

This means that integration-tests/logging-gelf are not executed in Quarkus CI runs
In the end there is increased risk for regression for this extension

PR #7307, opened 2020-02-20
Enable missing build profiles for native integration tests

gwenneg:
Some native integration tests are currently skipped in the JDK 11 GH action because the build profiles are not enabled. This fixes that issue.

I also found out that the `logging-gelf` integration tests are never executed even in JVM mode.

cc @gsmet @geoand


gwenneg:
There's actually an open issue about `logging-gelf` (#6286) and it seems it will require more than just enabling the build profile to fix it. I'll remove it from this PR.

geoand (approved):
Thanks!

gwenneg:
Thanks for the review!

PR #7312, opened 2020-02-20
Various GitHub actions adjustments

gsmet:
Let's see what CI has to say.

### DIFF
--- a/.github/workflows/ci-actions.yml
+++ b/.github/workflows/ci-actions.yml
@@ -94,9 +94,6 @@ jobs:
         env:
           ACCEPT_EULA: Y
           SA_PASSWORD: ActuallyRequired11Complexity
-          MYSQL_USER: hibernate_orm_test
-          MYSQL_PASSWORD: hibernate_orm_test
-          MYSQL_DATABASE: hibernate_orm_test
         ports:
           - 1433:1433
       neo4j:
--- a/.github/workflows/ci-actions.yml
+++ b/.github/workflows/ci-actions.yml
@@ -384,6 +384,8 @@ jobs:
               jsonb
               jsch
               jgit
+              quartz
+              qute
           - category: Misc2
             timeout: 35
             test-modules: >
@@ -392,14 +394,13 @@ jobs:
               test-extension
               logging-gelf
               bootstrap-config
+              optaplanner-jackson
+          # kubernetes-client alone takes 16mn+
           - category: Misc3
             timeout: 35
             test-modules: >
               kogito
               kubernetes-client
-              optaplanner-jackson
-              quartz
-              qute
           - category: Spring
             timeout: 30
             test-modules: >
--- a/.github/workflows/ci-actions.yml
+++ b/.github/workflows/ci-actions.yml
@@ -58,6 +58,7 @@ jobs:
     needs: build-jdk8
     timeout-minutes: 120
     strategy:
+      fail-fast: ${{ github.repository == 'quarkusio/quarkus' }}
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-actions.yml
```diff
@@ -58,8 +86,9 @@ jobs:
     needs: build-jdk8
     timeout-minutes: 120
     strategy:
+      fail-fast: false
```
n1hility: -1 on changing tihs. It will hurt our CI bandwidth and doesn't give you any new information

.github/workflows/ci-actions.yml
```diff
@@ -76,15 +76,15 @@ jobs:
           KEYCLOAK_PASSWORD: admin
           JAVA_OPTS: "-server -Xms64m -Xmx512m -XX:MetaspaceSize=96M -XX:MaxMetaspaceSize=256m -Djava.net.preferIPv4Stack=true -Djava.awt.headless=true -Dkeycloak.profile.feature.upload_scripts=enabled"
         ports:
-          - 8180:8080
+          - 127.0.0.1:8180:8080
```
n1hility: Out of curiosity what's the story behind this one?

n1hility (changes requested):

.github/workflows/ci-actions.yml
gsmet: That's useful information. Because knowing it only fails on a particular JDK is useful.

Right now, the bandwidth is more hurt because we run all the native tests even if the JDK 8 fails.

.github/workflows/ci-actions.yml
gsmet: We had some weird failures with SQL Server not being accessible while it was properly started, I'd like to see if it improves things or not.

.github/workflows/ci-actions.yml
gsmet: Or we should go back to our previous set up being we build on JDK 8 first and then we build the rest.

But not having a clue if it's an issue specific to one JDK will be a problem.

.github/workflows/ci-actions.yml
geoand: It's worth a try I guess, but I'm curious to why we thing that this will solve the issue :)

gsmet: I know Sanne tried that at some point so that's why I thought of it.

.github/workflows/ci-actions.yml
```diff
@@ -60,7 +60,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        java-version: [8, 11, 12]
+        java-version: [8, 11, 13]
```
gsmet: Hum. Apparently, we get a 404 for that one whereas it's supposed to be supported.

.github/workflows/ci-actions.yml
geoand: OK Cool. Lets hope it works

.github/workflows/ci-actions.yml
gsmet: @n1hility so I only disabled fail fast for forked repo. That way, we fail fast in the main repo and in their own repo, people have the full feedback.

.github/workflows/ci-actions.yml
gsmet: Got back to 12 as I got a 404 with 13.

n1hility (approved):

n1hility:
perfect thanks!

PR #7315, opened 2020-02-20
Update Keycloak docker image to 9.0.0

### LINKED ISSUE #7314: Update Keycloak docker image to 9.0.0
**Description**
Keycloak 9.0.0 has been released so we need to test against it

sberyozkin:
Fixes #7314

### DIFF
--- a/.github/workflows/ci-actions.yml
+++ b/.github/workflows/ci-actions.yml
@@ -65,7 +65,7 @@ jobs:
 
     services:
       keycloak:
-        image: quay.io/keycloak/keycloak:8.0.1
+        image: quay.io/keycloak/keycloak:9.0.0
         env:
           KEYCLOAK_USER: admin
           KEYCLOAK_PASSWORD: admin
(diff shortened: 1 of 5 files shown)


machi1990 (dismissed):
I think we should update this line https://github.com/quarkusio/quarkus/blob/master/ci-templates/jvm-build-steps.yaml#L33 too.

sberyozkin:
@machi1990 Thanks for spotting it :-), also updated the comment in the `build-parent/pom.xml`

machi1990:
> @machi1990 Thanks for spotting it :-), also updated the comment in the `build-parent/pom.xml`

Thanks @sberyozkin I also wonder if we should updates the keycloak instances in  https://github.com/quarkusio/quarkus/blob/master/.github/workflows/ci-actions.yml ?

@gsmet @sberyozkin WDYT?

gsmet:
Yes, it's what we use now for CI so we need to.

gsmet (changes requested):
Shouldn't `keycloak.version` be updated in the bom?

sberyozkin:
Hi @machi1990 @gsmet Oh, another resource is there :-), let me update it too.

@gsmet So far the update was about testing against Keycloak, so haven't thought about the bom update, makes sense though to sync, will do as well

gsmet:
I rebased to get the latest CI fixes. Please pull before pushing other changes.

sberyozkin:
@gsmet Thanks, does it look OK now ?

machi1990 (approved):
LGTM.

gsmet (approved):

PR #7316, opened 2020-02-20
Resteasy asyncio

FroMage:
Not ready to be merged because we need to wait for RESTEasy to be released, but can already be reviewed.

Some of those changes are not related to async-io and may already be present on a branch of @patriot1burke so it's likely that I'll rebase this branch on his when I know where it is.


FroMage:
@geoand can you review the test version part?

I'm a bit bothered by adding `version.txt` in the main artifact (not the test scope), but I don't think it would end up in the user's classpath if I placed it in the test resources.

FroMage:
@cescoffier can you review the reactive converter commit https://github.com/quarkusio/quarkus/pull/7316/commits/066966ec939c8b2504106a4463393ac84af84034 ?

geoand:
> @geoand can you review the test version part? [d32be2a](https://github.com/quarkusio/quarkus/commit/d32be2ad4ae60a982fd8c58b995148c7d64289cc)
> 
> I'm a bit bothered by adding `version.txt` in the main artifact (not the test scope), but I don't think it would end up in the user's classpath if I placed it in the test resources.

It's just part of junit5-internal anyway, so I don't think we are at much risk of it becoming part of the user's classpath. 
If you want to be on the same side, perhaps we can rename the file to something less generic- like `quarkus-version.txt`? That would practically eliminate the chances of a user reading it by accident

extensions/resteasy/runtime/src/main/java/io/quarkus/resteasy/runtime/standalone/VertxBlockingOutput.java
```diff
@@ -89,6 +92,44 @@ public void write(ByteBuf data, boolean last) throws IOException {
...
+            request.response().end(buffer, handler(ret));
+        } else {
+            request.response().write(buffer, handler(ret));
+        }
+        return ret.handle((v, t) -> {
+            if (t != null) {
+                if (data != null && data.refCnt() > 0) {
+                    data.release();
+                }
+                rethrow(new IOException("Failed to write", t));
```
gastaldi: This `rethrow` method seems useless if you use `UncheckedIOException` instead of the checked version

test-framework/junit5-internal/src/main/java/io/quarkus/test/TestUtil.java
```diff
@@ -0,0 +1,30 @@
...
+
+    private static String version;
+
+    private TestUtil() {
+    }
+
+    /**
+     * Gets the project version from a version.txt (that has been properly setup in maven to contain the version)
+     * This is needed in order to avoid hard-coding Quarkus dependency versions that could break the release process
+     */
```
gastaldi: You can also make this class a filtered resource and use `${project.version}` instead

gsmet:
Note: uses a snapshot so won't build on CI.

patriot1burke:
@FroMage I'll get my branch up to date once resteasy 4.5 is released and push a PR.

extensions/resteasy/runtime/src/main/java/io/quarkus/resteasy/runtime/standalone/VertxBlockingOutput.java
FroMage: Well, only if the caller knows to unwrap them. I prefer to not rely on that.

test-framework/junit5-internal/src/main/java/io/quarkus/test/TestUtil.java
FroMage: I've changed it to use the existing `Version.getVersion()`.

extensions/vertx/runtime/pom.xml
```diff
@@ -18,6 +18,14 @@
             <groupId>io.quarkus</groupId>
             <artifactId>quarkus-core</artifactId>
         </dependency>
+        <dependency>
```
stuartwdouglas: This feels wrong to me. Given that RESTEasy now depends on vert.x maybe it would make more sense to move the MessageBodyWriter implementations into the RESTEasy module?

test-framework/junit5-internal/src/main/java/io/quarkus/test/QuarkusUnitTest.java
```diff
@@ -165,6 +170,15 @@ public QuarkusUnitTest setAfterAllCustomizer(Runnable afterAllCustomizer) {
         return this;
     }
 
+    /**
+     * Provides a convenient way to either add additional dependencies to the application (if it doesn't already contain a
+     * dependency), or override a version (if the dependency already exists)
+     */
+    public QuarkusUnitTest setForcedDependencies(List<AppArtifact> forcedDependencies) {
```
stuartwdouglas: Very nice, I think this will be really useful for testing different dependency combinations.

stuartwdouglas (changes requested):

test-framework/junit5-internal/src/main/java/io/quarkus/test/QuarkusUnitTest.java
geoand: :+1

patriot1burke:
@FroMage @stuartwdouglas @gsmet This PR needs to go in before any work I've done.  My work is just minor boot performance improvements.

extensions/resteasy/runtime/src/main/java/io/quarkus/resteasy/runtime/standalone/VertxOutputStream.java
```diff
@@ -125,4 +128,72 @@ public void close() throws IOException {
...
+            int toWrite = Math.min(rem, buffer.writableBytes());
+            buffer.writeBytes(b, idx, toWrite);
+            rem -= toWrite;
+            idx += toWrite;
+            if (!buffer.isWritable()) {
+                ByteBuf tmpBuf = buffer;
+                this.pooledBuffer = buffer = allocator.allocateBuffer();
+                ret = ret.thenCompose(v -> response.writeNonBlocking(tmpBuf, false)
+                        .whenComplete((v2, t) -> {
+                            if (t != null)
```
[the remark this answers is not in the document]
stuartwdouglas: Actually that first comment is wrong, the buffers will be released by the write, so no change is required here.

stuartwdouglas:
I have rebased, updated to the released RESTEasy, and moved the JSON writers to the RESTEasy module.

stuartwdouglas:
The 4.5 release appears to be corrupted in a weird manner. The constructor for ResteasyJackson2Provider throws a java.lang.Error complaining about a compilation problem (this error is hard coded into the bytecode).

I have never seen this before, and have no idea what could have caused it.

gsmet:
@stuartwdouglas I let you decide if this should go in as a single commit or not. I'm especially wondering if some of the commits are partial and breaks the build but maybe they are meaningful so should stay that way.

stuartwdouglas:
We need to get it passing first, but my inclination would be to squash them, but I am not really sure.

emmanuelbernard:
So for some reason the build passes on my fork via GitHub actions https://github.com/emmanuelbernard/quarkus/actions/runs/46180919

stuartwdouglas:
It is racy, so the failures are intermittent.

stuartwdouglas:
Even though the CI passed it failed on my fork, there is still a (different) intermittent issue.

gsmet (changes requested):
Just my usual annoying note: we will need some squashing at the end. Mostly in case we have to revert this urgently.

asoldano (commented):
As this is effectively also serving as a component upgrade PR for RESTEasy, shouldn't you also bump microprofile rest client api dependency to 1.4.0 (which is what is implemented in RESTEasy currently) ?

asoldano:
Or I can create a new PR for the microprofile rest client upgrade just after this is merged ;-)

stuartwdouglas (approved):

stuartwdouglas:
I have opened https://github.com/quarkusio/quarkus/pull/7586 which is the same PR but with the commits squashed and MP REST client 1.4.

gastaldi:
Closing this PR on behalf of https://github.com/quarkusio/quarkus/pull/7586

PR #7322, opened 2020-02-20
Filter w + ip and Drafts

n1hility:
Don't merge this just yet.

### DIFF
--- a/.github/workflows/ci-actions.yml
+++ b/.github/workflows/ci-actions.yml
@@ -17,6 +17,7 @@ on:
       - '.github/NativeBuildReport.kts'
       - '.dependabot/**'
   pull_request:
+    types: [assigned, opened, synchronize, reopened, ready_for_review, edited]
     paths-ignore:
       - 'docs/src/main/asciidoc/**'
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-actions.yml
```diff
@@ -42,6 +43,13 @@ jobs:
   build-jdk8:
     name: "JDK 8 Build"
     runs-on: ubuntu-latest
+    # Skip draft PRs and those with WIP in the subject, rerun as soon as its removed
+    if: "github.event_name != 'pull_request' || ( \
+           github.event.pull_request.draft == false && \
+           github.event.pull_request.state != 'closed' && \
+           contains(github.event.pull_request.title, 'wip ') == false && \
+           (github.event.action != 'edited' || contains(github.event.changes.title.from, 'wip ')) \
```
gsmet: We have also people using `[WIP]` so we should probably take it into account.

Also is `contains` case insensitive?

n1hility: good point. i can add that as well

Yes its insensitive

PR #7324, opened 2020-02-21
Limit native runs to 8 at a time

n1hility:


geoand:
Weird that the PR fails with:

```
The job running on runner GitHub Actions 8 has exceeded the maximum execution time of 30. 
```

:)

gsmet:
Did someone cancel the build? Because some of the native builds were cancelled whereas they didn't exceed the timeout.

geoand:
> Did someone cancel the build? Because some of the native builds were cancelled whereas they didn't exceed the timeout.

Not me :)

PR #7327, opened 2020-02-21
Mongodb panache/mutiny

loicmathieu:
@cescoffier I wanted to play with Mutiny so I integrate it in MongoDB with Panache instead of Axle.

It works !

I have one big question, for the moment it uses Mutiny on the inside but it still exposes CompletionStage/Publisher.

Is the plan to expose Mutiny at the API level or only used it on the inside and expose standard reactive types at the API level ?


cescoffier:
Damned, didn't see that, I've opened another PR.

cescoffier:
Ok, we would need to merge both, because you did something I didn't - implement the previous API with the new one.

loicmathieu:
@cescoffier I didn't implement the previous API with the new one, I replace the previous API with Mutiny.
As I said in the Mutiny epic, the Axle implementation has been done recently for 1.3 so not already used. That's why we can replace it. No need to deprecate it.

The main difference between your PR and mine is that you return Mutiny types where I stick to CompletionStage and Publisher. This was a question I ask on the description of my PR and now it is answered ;)

So two solution here : update my PR to Mutiny return types and merge it or update you PR by emoving the Axle implementation and merge it. WDYT ?

loicmathieu:
@cescoffier updated to return the Mutiny types at the API level as we discussed.

loicmathieu:
@FroMage I updated the reactive MongoDB with Panache implementation to use Mutiny. 
If you want to have a look before it's merged I think that @cescoffier would want it merged quickly to meet the 1.3 target ...

loicmathieu:
I had to add the `quarkus-smallrye-context-propagation` inside the integration test to make it works. 

I don't really understand why and I think that it should be integrated by resteasy or panache because a user cannot know that it must include it ...

cescoffier:
The context-propagation issue is due to a weird chain of dependency. I will investigate that next.

docs/src/main/asciidoc/mongodb-panache.adoc
```diff
@@ -623,43 +623,43 @@ person.birth = LocalDate.of(1910, Month.FEBRUARY, 1);
...
+Uni<Long>  deleteCount = ReactivePerson.delete("status", Status.Alive);
 
 // delete all persons
 deleteCount = ReactivePerson.deleteAll();
 ----
 
-TIP: If you use MongoDB with Panache in conjunction with RESTEasy, you can directly return a reactive type inside your JAX-RS resource endpoint as RESTEasy will handle it correctly.
+TIP: If you use MongoDB with Panache in conjunction with RESTEasy, you can directly return a reactive type inside your JAX-RS resource endpoint as long as you include the `quarkus-resteasy-mutiny` extension.
 
 The same query facility exists for the reactive types, but the `stream()` methods act differently: they return a reactive stream `Publisher` instead of a `Stream`.
```
cescoffier: They should return a `Multi` (which implement `Publisher`)

cescoffier (changes requested):
a very small comment in the doc where it should mention `Multi` instead of `Publisher`.

docs/src/main/asciidoc/mongodb-panache.adoc
loicmathieu: updated

cescoffier (approved):

cescoffier:
Thanks @loicmathieu - Waiting for CI!

machi1990:
Merged thanks @loicmathieu @cescoffier

PR #7328, opened 2020-02-21
bump netty to 4.1.45.Final

michalszynkiewicz:


gsmet:
I don't have an opinion on that one. I will let @cescoffier decide!

cescoffier:
It should be ok. Let's merge and see :-)

cescoffier (approved):

Issue #7066, opened 2020-02-07
Native build of Kafka Streams fails at runtime with class RocksDBException not found


andreas-eberle:
**Describe the bug**
The native build of kafka streams aggregator fails at runtime with the exception 

```
2020-02-07 10:42:15,466 INFO  [org.apa.kaf.str.KafkaStreams] (pool-1-thread-1) stream-client [temperature-aggregator-9a921186-0418-4988-bc9e-e962abcaaeeb] State transition from CREATED to REBALANCING
RocksDBExceptionJni::ThrowNew/class - Error: unexpected exception!
Exception in thread "temperature-aggregator-9a921186-0418-4988-bc9e-e962abcaaeeb-GlobalStreamThread" java.lang.NoClassDefFoundError: Lorg/rocksdb/RocksDBException;
        at com.oracle.svm.jni.functions.JNIFunctions.FindClass(JNIFunctions.java:328)
        at org.rocksdb.RocksDB.open(RocksDB.java)
        at org.rocksdb.RocksDB.open(RocksDB.java:286)
        at org.apache.kafka.streams.state.internals.RocksDBTimestampedStore.openRocksDB(RocksDBTimestampedStore.java:72)
        at org.apache.kafka.streams.state.internals.RocksDBStore.openDB(RocksDBStore.java:180)
        at org.apache.kafka.streams.state.internals.RocksDBStore.init(RocksDBStore.java:202)
        at org.apache.kafka.streams.state.internals.WrappedStateStore.init(WrappedStateStore.java:48)
        at org.apache.kafka.streams.state.internals.CachingKeyValueStore.init(CachingKeyValueStore.java:58)
        at org.apache.kafka.streams.state.internals.WrappedStateStore.init(WrappedStateStore.java:48)
        at org.apache.kafka.streams.state.internals.MeteredKeyValueStore.lambda$init$0(MeteredKeyValueStore.java:107)
        at org.apache.kafka.streams.state.internals.MeteredKeyValueStore.measureLatency(MeteredKeyValueStore.java:260)
        at org.apache.kafka.streams.state.internals.MeteredKeyValueStore.init(MeteredKeyValueStore.java:105)
        at org.apache.kafka.streams.processor.internals.GlobalStateManagerImpl.initialize(GlobalStateManagerImpl.java:136)
        at org.apache.kafka.streams.processor.internals.GlobalStateUpdateTask.initialize(GlobalStateUpdateTask.java:61)
[...]

PR #7329, opened 2020-02-21
Add support for registering classes for JNI runtime access

gunnarmorling:
Hey @gsmet, here's what I'd hope is needed for addressing #7066.

### DIFF
--- a/core/deployment/src/main/java/io/quarkus/deployment/steps/NativeImageAutoFeatureStep.java
+++ b/core/deployment/src/main/java/io/quarkus/deployment/steps/NativeImageAutoFeatureStep.java
@@ -32,6 +32,7 @@ import io.quarkus.deployment.annotations.BuildProducer;
 import io.quarkus.deployment.annotations.BuildStep;
 import io.quarkus.deployment.builditem.DeploymentClassLoaderBuildItem;
 import io.quarkus.deployment.builditem.GeneratedNativeImageClassBuildItem;
+import io.quarkus.deployment.builditem.nativeimage.JniRuntimeAccessBuildItem;
 import io.quarkus.deployment.builditem.nativeimage.NativeImageProxyDefinitionBuildItem;
 import io.quarkus.deployment.builditem.nativeimage.NativeImageResourceBuildItem;
 import io.quarkus.deployment.builditem.nativeimage.NativeImageResourceBundleBuildItem;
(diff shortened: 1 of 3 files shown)


core/deployment/pom.xml
```diff
@@ -73,6 +73,10 @@
             <groupId>org.graalvm.sdk</groupId>
             <artifactId>graal-sdk</artifactId>
         </dependency>
+        <dependency>
```
gunnarmorling: It's needed in order to reference the `JNIRuntimeAccess` class. Can we add this here?

core/deployment/src/main/java/io/quarkus/deployment/steps/NativeImageAutoFeatureStep.java
```diff
@@ -63,6 +66,7 @@
             "org.graalvm.nativeimage.impl.RuntimeClassInitializationSupport",
             "rerunInitialization", void.class, Class.class, String.class);
     static final String RUNTIME_REFLECTION = RuntimeReflection.class.getName();
+    static final String JNI_RUNTIME_ACCESS = JNIRuntimeAccess.class.getName();
```
gunnarmorling: Maybe I should just give the class name as a string, avoiding the dependency? I'll try that.

gunnarmorling: So that's what I ended up doing. Makes the Kafka Streams quickstart pass for me.

gunnarmorling:
What's not quite clear to me (and a tad concerning): why didn't this issue not show up in the integration tests? There's [KafkaStreamsITCase](https://github.com/quarkusio/quarkus/blob/master/integration-tests/kafka-streams/src/test/java/io/quarkus/it/kafka/streams/KafkaStreamsITCase.java), shouldn't this have revealed the problem? Or is this test not executed for some reason?

gsmet:
I don't know about why the tests didn't fail. `kafka-streams` is in the pipeline so they should be run.

Can you maybe have a look at the run for your PR, check that the native tests have been properly executed? They are in the `Messaging` category.

gunnarmorling:
I took a quick look, and saw the test *class* itself, but not the *method*, in the logs. Will try and take a deeper look. Any other issues with the proposed fix itself?

gsmet (approved):

gsmet:
Merged, thanks!

PR #7336, opened 2020-02-21
Avoid annoying Warning logs when Sentry is disabled

ia3andy:
Those warning logs are not supposed to be visible when Sentry is disabled:
```
2020-02-21 17:13:08,062 WARN  [io.sen.dsn.Dsn] (executor-thread-1) *** Couldn't find a suitable DSN, Sentry operations will do nothing! See documentation: https://docs.sentry.io/clients/java/ ***
2020-02-21 17:13:08,072 WARN  [io.sen.DefaultSentryClientFactory] (executor-thread-1) No 'stacktrace.app.packages' was configured, this option is highly recommended as it affects stacktrace grouping and display on Sentry. See documentation: https://docs.sentry.io/cl
```

### DIFF
--- a/extensions/logging-sentry/runtime/src/main/java/io/quarkus/logging/sentry/SentryHandlerValueFactory.java
+++ b/extensions/logging-sentry/runtime/src/main/java/io/quarkus/logging/sentry/SentryHandlerValueFactory.java
@@ -11,6 +11,7 @@ import io.quarkus.runtime.configuration.ConfigurationException;
 import io.sentry.Sentry;
 import io.sentry.SentryOptions;
 import io.sentry.config.Lookup;
+import io.sentry.dsn.Dsn;
 import io.sentry.jul.SentryHandler;
 
 @Recorder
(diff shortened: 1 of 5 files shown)


extensions/logging-sentry/runtime/src/main/java/io/quarkus/logging/sentry/SentryHandlerValueFactory.java
```diff
@@ -11,14 +11,20 @@
...
 public class SentryHandlerValueFactory {
     private static final Logger LOG = Logger.getLogger(SentryConfigProvider.class);
 
     public RuntimeValue<Optional<Handler>> create(final SentryConfig config) {
+        final SentryConfigProvider provider = new SentryConfigProvider(config);
+        final Lookup lookup = new Lookup(provider, provider);
+
         if (!config.enable) {
+            // Disable Sentry
+            Sentry.init(SentryOptions.from(lookup, Dsn.DEFAULT_DSN));
```
gsmet: Looks a bit like dark magic to me :).

ia3andy: It's a bit magic, but Sentry API is not really made for this kind of use case.

 I Already had to implement the `Lookup` in order to use our own config.

Just to make you understand this code, when there is no DSN, Sentry show a warning and set it to `Dsn.DEFAULT_DSN`, Using this just prevent from showing the warning..

ia3andy:
@gsmet I don't think it fail because of the PR right?

extensions/logging-sentry/runtime/src/main/java/io/quarkus/logging/sentry/SentryHandlerValueFactory.java
ia3andy: This is what I am referring to above: https://github.com/getsentry/sentry-java/blob/master/sentry/src/main/java/io/sentry/dsn/Dsn.java#L101

gsmet (approved):

PR #7356, opened 2020-02-22
Create openshift wrapper extension.

iocanel:
I want to make `Quarkus on Openshift` to be as easy as possible.

Right now the user needs:

1. Add kubernetes extension
2. Add container-image-s2i extension
3. Set `kubernetes.deploymmet.target=openshift`

This pull request suggests an alternative aproach:

1. Add openshift extension

In other words, it provides a wrapper extension the makes it possible to deploy to openshift without additional configuration, while at the same time maintaining backwards compatibility.


iocanel:
Just to clarify this is just a wrapper extension, all the functionality is still left to the existing kubernetes extension.

extensions/kubernetes/openshift/deployment/src/main/resources/config-wrapper.vm
```diff
@@ -0,0 +1,15 @@
+package io.quarkus.kubernetes.deployment;
```
geoand: What is this file needed for?

geoand (commented):
I added a comment about the code gen that I don't really understand

extensions/kubernetes/openshift/deployment/src/main/resources/config-wrapper.vm
iocanel: They are not needed and have been removed.

geoand:
I personally like the idea a lot

maxandersen:
I like this idea of 'starter-extensions' too; wondering if the approach has issues if we start have more and more of these - what happens if you add a openshift and a knative starter - which one wins ?

/cc @emmanuelbernard @tqvarnst for awareness as we discussed these things in past of various degrees.

extensions/kubernetes/vanilla/deployment/src/main/java/io/quarkus/kubernetes/deployment/KubernetesProcessor.java
```diff
@@ -74,6 +75,33 @@
 
     private static final String OUTPUT_ARTIFACT_FORMAT = "%s%s.jar";
 
+    @BuildStep
```
maxandersen: how does these three methods interact....feels like it will be nondeterministic on who "wins" ?

bom/runtime/pom.xml
```diff
@@ -733,6 +733,11 @@
                 <artifactId>quarkus-kubernetes</artifactId>
                 <version>${project.version}</version>
             </dependency>
+            <dependency>
+                <groupId>io.quarkus</groupId>
+                <artifactId>quarkus-openshift</artifactId>
```
maxandersen: would this be better named as quarkus-openshift-s2i as we wouldn't have this if s2i wasn't in the picture and we could imagine more and more users using openshift without needing s2i at all....so maybe down the line there is something a bit more apropriate to be named quarkus-openshift ?

emmanuelbernard:
I think it makes a lot of sense.  I knew at some point some umbrella extension would just profile / tune the usage of other extensions. So +1 to me. A few general comments though (I did not look at the implementation):

- we must be able to use the underlying extensions to do the same
    - my main usage is for someone to be able to swap between Kube and OpenShift deployments
- ideally an aggregator extension would just push "proposed" properties that are read by the config if it's not explicitly set by the user config. (but we don't have such framework today)
- another way is via build items that should have less priority compared to an explicit property

So +1 to the idea.

bom/runtime/pom.xml
iocanel: I am not a big fan of `quarkus-openshift-s2i`. Mostly, because it's not just about s2i but also to provide default properties for openshift e.g.`quarkus.kubernetes.deploymenet-target` etc. In the future we could move all the openshfit  related code, in this extension too (if there wasn't time constraints I would have done it already).

iocanel:
> * another way is via build items that should have less priority compared to an explicit property

That's exactly the approach we've taken here.

bom/runtime/pom.xml
maxandersen: ack -makes sense. I guess question is if s2i.enbled=true  or s2i.enabled=false is the default then ?

extensions/kubernetes/vanilla/deployment/src/main/java/io/quarkus/kubernetes/deployment/KubernetesProcessor.java
iocanel: There is  not a race here.
We just produce build items,  based on the configuration properties that are found.

iocanel:
> I like this idea of 'starter-extensions' too; wondering if the approach has issues if we start have more and more of these - what happens if you add a openshift and a knative starter - which one wins ?

Thats' s an excellent question!

Here are some thoughts: In order to be  able to perform a deployment for any platform, we have the following requirements:

1. Certain APIS to be available on the target API server.
2. A compatible container image to be available (e.g. s2i won't work with vanilla nor knative).

Now, if both of the conditions above are meet for more than one `deployment target` we should pickup the first one, unless the user has explicitly selected something else.

integration-tests/infinispan-client/.gitignore
```diff
@@ -0,0 +1 @@
+/.apt_generated_tests/
```
geoand: I assume this was added by accident?

extensions/kubernetes/vanilla/deployment/src/main/java/io/quarkus/kubernetes/deployment/KubernetesProcessor.java
```diff
@@ -374,7 +399,7 @@ private void applyBuildItems(Session session,
         }
 
         //Hanlde custom s2i builder images
```
geoand:
```suggestion
        //Handle custom s2i builder images
```

extensions/kubernetes/vanilla/deployment/src/main/java/io/quarkus/kubernetes/deployment/KubernetesConfigUtil.java
```diff
@@ -24,13 +25,17 @@
 public class KubernetesConfigUtil {
 
     private static final String DEKORATE_PREFIX = "dekorate.";
+    private static final String QUARKUS_PREFIX = "quarkus.";
```
geoand: This doesn't seem to be used

extensions/kubernetes/openshift/deployment/src/main/java/io/quarkus/openshift/deployment/Constants.java
```diff
@@ -0,0 +1,6 @@
+package io.quarkus.openshift.deployment;
+
+public class Constants {
+    static final String OPENSHIFT = "openshift";
+    static final String DEPLOYMENT_CONFIG = "openshift";
```
geoand: Is this correct?

geoand (commented):
Added a few comments and questions

docs/src/main/asciidoc/openshift.adoc
```diff
@@ -0,0 +1,176 @@
+////
+This guide is maintained in the main Quarkus repository
+and pull requests should be submitted there:
+https://github.com/quarkusio/quarkus/tree/master/docs/src/main/asciidoc
+////
+= Quarkus - Openshift extension
+
+include::./attributes.adoc[]
+
+This guide covers generating and deploying Openshift resources based on sane default and user supplied configuration.
```
geoand:
```suggestion
This guide covers generating and deploying Openshift resources based on sane defaults and user supplied configuration.
```

docs/src/main/asciidoc/openshift.adoc
```diff
@@ -0,0 +1,176 @@
...
+    -Dextensions="openshift"
+
+cd openshift-quickstart
+----
+
+=== Openshift
+
+Quarkus offers the ability to automatically generate Openshift resources based on sane defaults and user supplied configuration.
+The Openshift extension is actually a wrapper extension that brings together the link:kubernetes[kubernetes] and link:container-image-s2i[container-image-s2i]
+extensions with sensible defaults so that its easier for the user to get started with Quarkus on Openshift.
```
geoand:
```suggestion
extensions with sensible defaults so that it's easier for the user to get started with Quarkus on Openshift.
```

docs/src/main/asciidoc/openshift.adoc
```diff
@@ -0,0 +1,176 @@
...
+
+To add an annotation in the generated resources:
+
+[source]
+----
+quarkus.openshift.env-vars.my-env-var.value=foobar
+----
+
+The command above will add `MY_ENV_VAR=foobar` as an environment variable.
+Please, notice that the key `my-env-var` will be converted to uppercase and dashes will be replaced by underscores resulting in `MY_ENV_VAR`.
```
geoand:
```suggestion
Please note that the key `my-env-var` will be converted to uppercase and dashes will be replaced by underscores resulting in `MY_ENV_VAR`.
```

docs/src/main/asciidoc/openshift.adoc
```diff
@@ -0,0 +1,176 @@
...
+
+You may also noticed that in contrast to labels, and annotations for environment variables you don't just use a key=value approach.
```
geoand:
```suggestion
You may also have noticed that in contrast to labels and annotations for environment variables you don't just use a key=value approach.
```

docs/src/main/asciidoc/openshift.adoc
```diff
@@ -0,0 +1,176 @@
...
+That is because for environment variables there are additonal options rather than just value.
```
geoand:
```suggestion
That is because for environment variables there are additional options rather than just defining value.
```

docs/src/main/asciidoc/openshift.adoc
```diff
@@ -0,0 +1,176 @@
...
+
+===== Environment variables from Secret
+
+To add all key value pairs of a `Secret` as envrionment variables:
```
geoand:
```suggestion
To add all key value pairs of a `Secret` as environment variables:
```

docs/src/main/asciidoc/openshift.adoc
```diff
@@ -0,0 +1,176 @@
...
+
+[source]
+----
+quarkus.openshift.env-vars.my-env-var.secret=my-secret
+----
+
+===== Environment variables from ConfigMap
+
+To add all key value pairs of a `ConfigMap` as envrionment variables:
```
geoand:
```suggestion
To add all key value pairs of a `ConfigMap` as environment variables:
```

docs/src/main/asciidoc/openshift.adoc
```diff
@@ -0,0 +1,176 @@
...
+
+[source]
+----
+quarkus.openshift.env-vars.my-env-var.configmap=my-secret
+----
+
+==== Exposing Routes
+
+To expose a route for the Quarkus application:
```
geoand:
```suggestion
To expose a `Route` for the Quarkus application:
```

geoand (commented):
And some comments on the doc :)

integration-tests/infinispan-client/.gitignore
iocanel: Yeap

bom/runtime/pom.xml
iocanel: `s2i.enabled=true` by default.

docs/src/main/asciidoc/openshift.adoc
```diff
@@ -0,0 +1,176 @@
...
+----
+
+==== Exposing Routes
+
+To expose a `Route` for the Quarkus application:
+
+[source]
+----
+quarkus.openshift.expose=true
+----
```
maxandersen: should we add a:

TIP: You do not need to always put these in `application.properties` you can also pass it as a command line argument: `-Dquarkus.openshift.expose=true`

maxandersen:
looks really good - much simpler story in docs too.

One nice addition would be to add labels to the Deployment/Deployment Config so it will get Quarkus icon in dev console. 

needs to have either app.openshift.io/runtime=quarkus or app.kubernetes.io/name=quarkus label.

details at https://github.com/openshift/console/pull/4306

docs/src/main/asciidoc/openshift.adoc
iocanel: Agree! Let me add that.

geoand:
Are we good to go with this?

iocanel:
> Are we good to go with this?

I think that everything is wrapped, so I'd say yeah.

geoand:
I'll let @maxandersen push the magic button then

maxandersen:
Missed this. I'll try rebase it in a few hours when at desktop.

geoand:
I rebased the PR onto master since Github UI was reporting a conflict

extensions/kubernetes/vanilla/runtime/src/main/resources/META-INF/quarkus-extension.yaml
```diff
@@ -0,0 +1,9 @@
+---
+name: "Openshift"
+metadata:
+  keywords:
+  - "openshift"
+  guide: "https://quarkus.io/guides/openshift"
```
maxandersen: this metadata looks wrong - this is for openshift not kubernetes?

geoand: Yup, I'll fix it

maxandersen:
one open question is the status of the extension is marked stable even thuogh its new and depends on container-image that are preview. ill open separate issue for that.

PR #7375, opened 2020-02-24
Make some small adjustments to the container image doc

geoand:


docs/src/main/asciidoc/container-image.adoc
```diff
@@ -19,6 +19,7 @@ Quarkus provides extensions for building (and pushing) container images. Current
 
 The extension `quarkus-container-image-jib` is powered by https://github.com/GoogleContainerTools/jib[Jib] for performing container builds.
 The major benefit of using Jib with quarkus, is that all dependencies (everything found under target/lib) are cached in a different layer than the actual application making rebuilds extra fast and extra small (when it comes to pushing).
+Another important benefit of using this extension is that it can be used to build a container image without having to have any dedicated client side tooling (like Docker).
```
cescoffier: not only tooling, also a running daemon.

docs/src/main/asciidoc/container-image.adoc
```diff
@@ -88,15 +89,16 @@ For customizing the extension the following properties are available:
...
+| quarkus.container-image.tag             | String                          | The tag of the image                                                   | The application version
+| quarkus.container-image.registry        | String                          | The registry to use for pushing                                        |
+| quarkus.container-image.username        | String                          | The registry username                                                  |
+| quarkus.container-image.password        | String                          | The registry password                                                  |
+| quarkus.container-image.insecure        | Boolean                         | Flag to allow insecure registries                                      | false
+| quarkus.container-image.build           | Boolean                         | Flag to enable building of a container image                           | false
+| quarkus.container-image.push            | Boolean                         | Flag to enable pushing of a container image to the configured registry | false
```
cescoffier: Question: which image format does it support?

docs/src/main/asciidoc/container-image.adoc
```diff
@@ -118,7 +120,7 @@ On top the the global container image options, the following jib specific option
 
 On top the the global container image options, the following docker specific options are availble:
 
-.Dokcer Options
+.Docker Options
```
cescoffier: We start the documentation with: no need for docker. And then we have docker options. Should we rename them?

cescoffier (dismissed):

docs/src/main/asciidoc/container-image.adoc
geoand: Docker 2.2. Jib also provides the option to use the OCI format, should I expose this option via a property?

docs/src/main/asciidoc/container-image.adoc
geoand: I went ahead and removed these lines. WDYT?

docs/src/main/asciidoc/container-image.adoc
geoand: very good point, I added it!

docs/src/main/asciidoc/container-image.adoc
```diff
@@ -88,22 +89,22 @@ For customizing the extension the following properties are available:
...
+| quarkus.container-image.username        | String                          | The registry username                                                  |
+| quarkus.container-image.password        | String                          | The registry password                                                  |
+| quarkus.container-image.insecure        | Boolean                         | Flag to allow insecure registries                                      | false
+| quarkus.container-image.build           | Boolean                         | Flag to enable building of a container image                           | false
+| quarkus.container-image.push            | Boolean                         | Flag to enable pushing of a container image to the configured registry | false
 |====
 
 === Jib Options
 
 On top the the global container image options, the following jib specific options are availble:
```
gsmet:
```suggestion
On top the the global container image options, the following jib specific options are available:
```

docs/src/main/asciidoc/container-image.adoc
```diff
@@ -118,7 +119,6 @@ On top the the global container image options, the following jib specific option
 
 On top the the global container image options, the following docker specific options are availble:
```
gsmet:
```suggestion
On top the the global container image options, the following docker specific options are available:
```

docs/src/main/asciidoc/container-image.adoc
```diff
@@ -129,7 +129,6 @@ On top the the global container image options, the following docker specific opt
 
 On top the the global container image options, the following s2i specific options are availble:
```
gsmet:
```suggestion
On top the the global container image options, the following s2i specific options are available:
```

docs/src/main/asciidoc/container-image.adoc
```diff
@@ -19,6 +19,7 @@ Quarkus provides extensions for building (and pushing) container images. Current
 
 The extension `quarkus-container-image-jib` is powered by https://github.com/GoogleContainerTools/jib[Jib] for performing container builds.
 The major benefit of using Jib with quarkus, is that all dependencies (everything found under target/lib) are cached in a different layer than the actual application making rebuilds extra fast and extra small (when it comes to pushing).
```
gsmet:
```suggestion
The major benefit of using Jib with Quarkus, is that all dependencies (everything found under target/lib) are cached in a different layer than the actual application making rebuilds extra fast and extra small (when it comes to pushing).
```

gsmet (changes requested):
Spotted some typos in the original text. I suppose I will have to proofread that one completely once this is in.

geoand:
Fixed. @gsmet I plan to revisit these docs extensively over this course of this week since there are going to be more changes coming in anyway. So I would say that you should probably wait a few days

gsmet:
@geoand OK, please ping me when it's done because there's a good chance I will forget about this :).

geoand:
You bet @gsmet :)

iocanel (approved):
LGTM

geoand:
I will go ahead and merge this one now (if all reviewers agree) and when @iocanel and myself have finished with all the code and docs changes, I'll ping you for a review of everything that has gone into the docs so far.

This will also probably mean that one or more of the existing guides will be made obsolete and might need to be removed

gsmet (approved):

docs/src/main/asciidoc/container-image.adoc
geoand: I actually tried the OCI thing, but at this point it's more trouble than it's worth since I couldn't push such an image to docker or quay

docs/src/main/asciidoc/container-image.adoc
cescoffier: Agreed!

PR #7381, opened 2020-02-24
Reactive Messaging - add @Dependent scope only if not JAX-RS construct

### LINKED ISSUE #7354: CDI lifecycle behavior different when using reactive messaging and @Incoming and @StartupEvent
**Describe the bug**
It appears that although the default lifecycle behavior of JAX-RS resource classes is that of `@Singleton` (as described in the [REST/JSON guide](https://quarkus.io/guides/rest-json#lifecycle-of-resources)), when using reactive messaging and `@Incoming`, multiple instances of the resource appear to be created, so that variables initialized in startup events aren't accessible when JAX-RS endpoints are accessed.

**Expected behavior**
Variables initialized in a `@Observes StartupEvent` method are accessible later on when JAX-RS endpoints are accessed.

**Actual behavior**
Instance variables initialized in a `@Observes StartupEvent` method in a JAX-RS resource class are not set to the right values when the JAX-RS endpoints are accessed. Basic debugging shows multiple instances are created, not "respecting" the default behavior from docs.

**To Reproduce**
1. Generate new basic JAX-RS project
2. Change `GreetingResource` to this:

```java
package org.acme.quickstart;

import java.io.IOException;
import java.util.concurrent.CompletionStage;

import javax.enterprise.event.Observes;
import javax.ws.rs.GET;
import javax.ws.rs.Path;
import javax.ws.rs.Produces;
import javax.ws.rs.core.MediaType;

import org.eclipse.microprofile.reactive.messaging.Incoming;

import io.quarkus.runtime.StartupEvent;
import io.smallrye.reactive.messaging.kafka.KafkaMessage;

@Path("/hello")

    private String foo;

    @GET
    public String getfoo() {
        System.out.println("getfoo() instance: " + this);
            return foo;
    }

    @Incoming("orders")
    public CompletionStage<Void> onMessage(KafkaMessage<String, String> message)
            throws IOException {
        return message.ack();
    }

    public void init(@Observes StartupEvent ev) {
        System.out.println("init() instance: " + this);
        foo = "bar";
    }
}
```
2a. Fire up a temporary kafka server on `localhost:9092` using the `docker-compose.yml` from the [kafka guide](https://quarkus.io/guides/kafka).
3. Compile and run executable jar
4. Access `localhost:8080`.
5. Observe the value returned is *blank*, rather than the expected value of `bar` as set in the `init()` method. Also observe the identifiers generated in the `init()` method are different than those generated in the `getfoo()` method. I would expect these to be identical if indeed the behavior was that of a singleton.
6. Add `@Singleton` to the resource class, rebuild, re-run, and try curl again
7. Observe expected behavior (the value `bar` is output).
8. You can also comment out the `@Incoming` method which makes it work.

**Configuration**
```properties
# no special properties needed
```
`
 - Output of `uname -a` or `ver`:  `Darwin jfalkner-OSX 18.7.0 Darwin Kernel Version 18.7.0: Sun Dec  1 18:59:03 PST 2019; root:xnu-4903.278.19~1/RELEASE_X86_64 x86_64`
 - Output of `java -version`: 

```
java version "1.8.0_201"
Java(TM) SE Runtime Environment (build 1.8.0_201-b09)
Java HotSpot(TM) 64-Bit Server VM (build 25.201-b09, mixed mode)
```

 - GraalVM version (if different from Java): N/A
 - Quarkus version or git rev:  1.2.1.Final

mkouba:
- resolves #7354

### DIFF
--- a/extensions/smallrye-reactive-messaging/deployment/src/main/java/io/quarkus/smallrye/reactivemessaging/deployment/QuarkusMediatorConfigurationUtil.java
+++ b/extensions/smallrye-reactive-messaging/deployment/src/main/java/io/quarkus/smallrye/reactivemessaging/deployment/QuarkusMediatorConfigurationUtil.java
@@ -1,10 +1,10 @@
 package io.quarkus.smallrye.reactivemessaging.deployment;
 
-import static io.quarkus.smallrye.reactivemessaging.deployment.DotNames.ACKNOWLEDGMENT;
-import static io.quarkus.smallrye.reactivemessaging.deployment.DotNames.BROADCAST;
-import static io.quarkus.smallrye.reactivemessaging.deployment.DotNames.INCOMING;
-import static io.quarkus.smallrye.reactivemessaging.deployment.DotNames.MERGE;
-import static io.quarkus.smallrye.reactivemessaging.deployment.DotNames.OUTGOING;
+import static io.quarkus.smallrye.reactivemessaging.deployment.ReactiveMessagingDotNames.ACKNOWLEDGMENT;
+import static io.quarkus.smallrye.reactivemessaging.deployment.ReactiveMessagingDotNames.BROADCAST;
+import static io.quarkus.smallrye.reactivemessaging.deployment.ReactiveMessagingDotNames.INCOMING;
+import static io.quarkus.smallrye.reactivemessaging.deployment.ReactiveMessagingDotNames.MERGE;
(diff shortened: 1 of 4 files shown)


jmartisk (approved):

mkouba:
@cescoffier pls review. I can't force-merge this PR anymore.

cescoffier (approved):

PR #7385, opened 2020-02-24
Fix broken MongoClientBuildItem support

### LINKED ISSUE #7378: MongoClientBuildItem is never produced
**Describe the bug**
`MongoClientProcessor` has a `build` method which is supposed to produce `MongoClientBuildItem`(s) when the relevant Quarkus Mongo configuration is present. I never see the fields in `MongodbConfig` be anything other than `null`, hence the build items never get created.

Having `MongoClientBuildItem` available, is useful for other extensions like the Camel Mongo extension, where it can fetch the client and wire it into the relevant Camel component.

**Expected behavior**
If any of the Mongo configuration options are set, E.g `quarkus.mongodb.hosts` or `quarkus.mongodb.connection-string` then a `MongoClientBuildItem` should be produced.

**Actual behavior**
No `MongoClientBuildItem` ever seems to be produced, regardless of the Mongo configuration used.

**To Reproduce**
1. In your IDE, set a breakpoint in `MongoClientProcessor` at the start of the `build` method
2. Launch test `BookResourceTest` in debug mode
3. Observe conditions for creating `MongoClientBuildItem` are never satisfied

geoand:
Fixes: #7378

### DIFF
--- a/extensions/mongodb-client/deployment/src/main/java/io/quarkus/mongodb/deployment/MongoClientProcessor.java
+++ b/extensions/mongodb-client/deployment/src/main/java/io/quarkus/mongodb/deployment/MongoClientProcessor.java
@@ -7,7 +7,7 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashSet;
 import java.util.List;
-import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
 
@@ -37,8 +37,10 @@ import io.quarkus.deployment.Capabilities;
 import io.quarkus.deployment.annotations.BuildProducer;
 import io.quarkus.deployment.annotations.BuildStep;
 import io.quarkus.deployment.annotations.Record;
+import io.quarkus.deployment.annotations.Weak;
 import io.quarkus.deployment.builditem.ApplicationArchivesBuildItem;
 import io.quarkus.deployment.builditem.CombinedIndexBuildItem;
+import io.quarkus.deployment.builditem.ConfigurationBuildItem;
 import io.quarkus.deployment.builditem.ExtensionSslNativeSupportBuildItem;
 import io.quarkus.deployment.builditem.FeatureBuildItem;
 import io.quarkus.deployment.builditem.SslNativeConfigBuildItem;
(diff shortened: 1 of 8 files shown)


geoand:
There is one thing that needs to change however for `MongoClientBuildItem`.

The `value` of each `RuntimeValue` needs to become an optional instead of an actual client. The reason is that if the client is never injected into application code, the bean is never created.

loicmathieu:
> The reason is that if the client is never injected into application code, the bean is never created.

This is the behaviour I will need to change ...

geoand:
> > The reason is that if the client is never injected into application code, the bean is never created.
> 
> This is the behaviour I will need to change ...

Yes of course, but it's necessary, otherwise we need to make the beans unremovable which goes counter to the point of the optimization :)

loicmathieu:
> we need to make the beans unremovable which goes counter to the point of the optimization :)

I don't really see the point of defining a mongoclient connection and not using it in the code. I don't think that bean removal should cover such case.

geoand:
That's not the issue :wink: 

The issue is that code can use only the blocking client and not the reactive client. In that case, no bean will be created for the reactive client. Ditto if the opposite happens.

We definitely can make the beans unremoveable, but that goes counter to what the optimization actually is meant for

loicmathieu:
> That's not the issue

That's mine, but I will look at it later, let stop the digression here :)

geoand:
I am not hellbent on the optional thing, I'm just saying it's the proper thing to do. If it's too much of a hassle for the consumers of the Build Item to update, then we can make the bean unremovable :)

loicmathieu:
@geoand in fact I just have an _enlightment_ (I don't know if it's the correct word). `@MongoClientName` as a base for removing bean is not suitable for my use case but I have the `@MongoEntity(clientName)` annotation so I should be able to define a `UnremovableBeanBuildItem` to tell Arc to avoid removing the client I need even if the `@MongoClientName` is not used ! This will fix my issue ! :tada:

geoand:
Excellent 👌

extensions/mongodb-client/deployment/src/main/java/io/quarkus/mongodb/deployment/MongoClientProcessor.java
```diff
@@ -365,21 +358,51 @@ BeanContainerListenerBuildItem build(
...
+        recorder.configureRuntimeProperties(codecProvider.getCodecProviderClassNames(),
+                bsonDiscriminator.getBsonDiscriminatorClassNames(), mongodbConfig);
+
+        for (String connectionName : resolveAllConnectionNamesFromConfig(config)) {
+            mongoClients.produce(new MongoClientBuildItem(recorder.getClient(connectionName),
+                    recorder.getReactiveClient(connectionName), connectionName));
+        }
+    }
+
+    private Set<String> resolveAllConnectionNamesFromConfig(ConfigurationBuildItem config) {
```
loicmathieu: Why not using this new methods also in the `createMongoClientProducerBean`methods that do the same but based on Jandex ?

loicmathieu (changes requested):

extensions/mongodb-client/deployment/src/main/java/io/quarkus/mongodb/deployment/MongoClientProcessor.java
geoand: That is an interesting idea! I went the way I did because that solution feels closer to the problem at hand. 
That said, I'll sleep on this idea and once the Camel folks respond with whether or not the whole idea works then I'll have settled :)

geoand: Implemented

loicmathieu (approved):
LGTM but I will need to test it to see if it fixes my issue.
Maybe hold on to merge it untill I have time to test it (end of week)

geoand:
The issue is we really need this for Camel. So I propose merging ASAP and then making any extra charges if necessary when you have the time to test

loicmathieu:
I'm OK with this, I didn't saw prior to my comment that it fixes the issue for Camel.
So let's merge it, I'll give you feedback later on this.

geoand:
Thanks

gsmet (approved):
I don't have the time to carefully review this so I'm gonna trust you on that one, considering it's blocking for Camel and CI is green.

PR #7389, opened 2020-02-24
Prevent NCDFE on thread pool shutdown when CL is closed early

dmlloyd:


dmlloyd:
This can be reverted once upstream JBoss Threads has something equivalent in place.

geoand (approved):

PR #7396, opened 2020-02-24
feat: group name version are inherited from container image.

iocanel:
group, name & version are config properties that are present in:

- container-image
- kubernetes
- opesnhift
- knative 

This pull request changes kubernetes, openshift and knative config objects so that they default to container-image for these 3 properties (group, name & version).

### DIFF
--- a/docs/src/main/asciidoc/kubernetes.adoc
+++ b/docs/src/main/asciidoc/kubernetes.adoc
@@ -210,10 +210,10 @@ The table below describe all the available configuration options.
 .Kubernetes
 |====
 | Property                                           | Type                                      | Description | Default Value
-| quarkus.kubernetes.group                           | String                                    |             |
-| quarkus.kubernetes.name                            | String                                    |             |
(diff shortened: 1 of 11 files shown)


iocanel:
@lordofthejars: We talked about it in twitter and felt that we can do better. So, with this change if the user sets

`quarkus.container-image.group` it will also apply to kubernetes, openshift and knative resource. Still the user is free to override the property, in which case it will only affect labels and resource names (image name is 100% responsibility of container-image).

lordofthejars:
That's awesome thank you very much.

geoand:
CI doesn't seem to like it :)

iocanel:
> CI doesn't seem to like it :)
I know, I am working on it!

geoand:
This has a few issues, I'll take care of them

geoand:
Actually we came to the conclusion that I was wrong, I hadn't properly installed all the dependencies

geoand (approved):

PR #7406, opened 2020-02-25
Remove wrong and unused dependency from the bom

cescoffier:
@michalszynkiewicz found that yesterday. The BOM was defining a dependency on a non-existing artifact. I introduced that bug during the introduction of Mutiny. As the dependency was not used, it never attempted to resolve it (which would have failed).

This PR just removes the dependency.


michalszynkiewicz (approved):

michalszynkiewicz:
thanks!

PR #7416, opened 2020-02-25
Use BeanArchiveIndexBuildItem in hibernate-validator extension to allow working with generated beans

mswiderski:
…ow working with generated beans

@gsmet here comes another one of the same kind :)

I didn't add test for it as I am not really sure what can be done to test this and my test for similar on in security was removed (or replaced but I can't find with what). So if there is a need fr=or a test I'd appreciate some hints on how to approach it.


mswiderski:
failure does not seem to be related to this change

gsmet (approved):

gsmet:
Makes perfect sense. Thanks for the fix!

PR #7417, opened 2020-02-25
Kafka Streams SASL and SSL config

### LINKED ISSUE #4961: [kafka streams] sasl not configurable
**Describe the bug**
The kafka streams extensions only allows for the bootstrap-servers and application-id to be set at runtime. This works in a non authentication environment but when auth is required there are other properties which needs to be configurable at runtime.
Also the confluent schema registry need to be configurable (again with auth properties)

**Expected behavior**
More properties needs to be configurable @ runtime

For me now at least; could be more:
kafka-streams.schema.registry.url
kafka-streams.sasl.jaas.config
kafka-streams.ssl.truststore.location
kafka-streams.ssl.truststore.password
kafka-streams.ssl.keystore.location
kafka-streams.ssl.keystore.password
kafka-streams.ssl.key.password

**Actual behavior**
Properties cannot be overriden @ runtime.

**To Reproduce**

Try to change them @ runtime does not work.

**Solution proposal**
As there could be other properties which might need to be added could there be an option which let the user define which properties he/she whants configurable @ runtime?
Prefix the runtime config with quarkus.kafka-streams and the build time config just with kafka-streams.

alesj:
Fix #4961


extensions/kafka-streams/runtime/src/main/java/io/quarkus/kafka/streams/runtime/SaslConfig.java
```diff
@@ -0,0 +1,90 @@
...
+     * Kerberos kinit command path
+     */
+    @ConfigItem(name = "kerberos.kinit.cmd")
+    public Optional<String> kerberosKinitCmd;
+
+    /**
+     * Login thread will sleep until the specified window factor of time from last refresh
+     */
+    @ConfigItem(name = "kerberos.ticket.renew.window.factor")
+    public Optional<String> kerberosTicketRenewWindowFactor;
```
gunnarmorling: Why is this a string, shouldn't it be rather a double (some for some more props below)?

extensions/kafka-streams/runtime/src/main/java/io/quarkus/kafka/streams/runtime/SaslConfig.java
```diff
@@ -0,0 +1,90 @@
...
+     * The desired minimum time for the login refresh thread to wait before refreshing a credential
+     */
+    @ConfigItem(name = "login.refresh.min.period.seconds")
+    public Optional<String> loginRefreshMinPeriodSeconds;
+
+    /**
+     * The amount of buffer time before credential expiration to maintain when refreshing a credential
+     */
+    @ConfigItem(name = "login.refresh.buffer.seconds")
+    public Optional<String> loginRefreshBufferSeconds;
```
gunnarmorling: Should be an int. But also worth double checking with Quarkus guidelines: should it be rather `Duration`, and be given as such (which would impact the option name, though, which might not be desirable as it's coming from Kafka upstream IIUC).

integration-tests/kafka-streams/src/test/java/io/quarkus/it/kafka/streams/KafkaStreamsPropertiesTest.java
```diff
@@ -0,0 +1,35 @@
...
+
+    @Inject
+    KafkaStreams streams;
+
+    @Test
+    public void testProperties() throws Exception {
+        // reflection hack ... no other way to get raw props ...
+        Field configField = KafkaStreams.class.getDeclaredField("config");
+        configField.setAccessible(true);
+        StreamsConfig config = (StreamsConfig) configField.get(streams);
```
gunnarmorling: How about adding a package visible accessor?

gunnarmorling (commented):
@alesj, a few comments inline. Apart from those: the docs (props table in the guide) should be updated, too. Plus, ideally the existing integration test would at least cover one simple SSL set-up, so to make sure things work in general and future regressions are avoided.

extensions/kafka-streams/runtime/src/main/java/io/quarkus/kafka/streams/runtime/SaslConfig.java
alesj: Yeah, I didn't bother, since KS's config takes care of the conversions anyway.
If you think we should, then this can easily be changed.

integration-tests/kafka-streams/src/test/java/io/quarkus/it/kafka/streams/KafkaStreamsPropertiesTest.java
alesj: What do you mean?

alesj:
> the docs (props table in the guide) should be updated,

I thought that's the beauty of having this config stuff, so it's auto-generated?

alesj:
> ideally the existing integration test would at least cover one simple SSL set-up

Yeah, working on it.

extensions/kafka-streams/runtime/src/main/java/io/quarkus/kafka/streams/runtime/SaslConfig.java
gunnarmorling: Let's ask @gsmet. My understanding is that Quarkus wants to expose properly typed options, so to offer e.g. editor support (validation) down the road.

integration-tests/kafka-streams/src/test/java/io/quarkus/it/kafka/streams/KafkaStreamsPropertiesTest.java
gunnarmorling: Ah, seems you get a field from the upstream `KafkaStreams` object. I thought about having a getter in our own object (topology manager), where we could expose a package-visible getter for testing purposes.

gunnarmorling:
> I thought that's the beauty of having this config stuff, so it's auto-generated?

Ah, I didn't know that :) If so, that's much cooler of course.

alesj:
>> I thought that's the beauty of having this config stuff, so it's auto-generated?

> Ah, I didn't know that :) If so, that's much cooler of course.

Right, @gsmet ?

extensions/kafka-streams/runtime/src/main/java/io/quarkus/kafka/streams/runtime/SaslConfig.java
gsmet: Yes, the Quarkus config should be properly typed.

extensions/kafka-streams/runtime/src/main/java/io/quarkus/kafka/streams/runtime/KafkaStreamsRuntimeConfig.java
```diff
@@ -37,10 +37,42 @@
...
+     * e.g. to diff between different registry impls / instances
+     * as they have this registry url under different property key.
+     */
+    @ConfigItem(name = "schema.registry.key", defaultValue = "schema.registry.url")
+    public String schemaRegistryKey;
+
+    /**
+     * The schema registry url.
+     */
+    @ConfigItem(name = "schema.registry.url")
```
gsmet: In Quarkus config, we use a dot to separate sections and a dash to separate words. It's a general comment about the config key of this PR.

extensions/kafka-streams/runtime/src/main/java/io/quarkus/kafka/streams/runtime/SaslConfig.java
gsmet: Yes, it should be typed.

extensions/kafka-streams/runtime/src/main/java/io/quarkus/kafka/streams/runtime/SaslConfig.java
gsmet: Yeah all these properties should be `Duration` and we should drop the `Seconds` part. They will be documented so it's OK.

gsmet (changes requested):
I added some comments.

As for the documentation, it will be automatically generated so no need to do anything except having proper Javadoc.

alesj:
I just pushed some more work, but it's not done yet ... so no new review yet needed.
Will let you guys know when I think I'm done.

alesj:
@gunnarmorling @gsmet I changed to typed props, used dash, added ssl test ... please re-review

extensions/kafka-streams/runtime/src/main/java/io/quarkus/kafka/streams/runtime/KafkaStreamsRuntimeConfig.java
```diff
@@ -37,10 +37,42 @@
     @ConfigItem
     public List<String> topics;
 
+    /**
+     * The schema registry key.
+     * e.g. to diff between different registry impls / instances
```
gunnarmorling: Perhaps clarify the value to be used for Apicurio?

gunnarmorling (commented):
@alesj, sorry for the delay. LGTM overall. One minor suggestion inline. And one question on the test: wouldn't the test Kafka be needed to configured accordingly, so to accept SSL connections? Maybe I'm missing something?

alesj:
@gunnarmorling > wouldn't the test Kafka be needed to configured accordingly, so to accept SSL connections? Maybe I'm missing something?

So you're saying I should fix the Kafka tests as well, not just Kafka Streams tests?

extensions/kafka-streams/runtime/src/main/java/io/quarkus/kafka/streams/runtime/KafkaStreamsRuntimeConfig.java
alesj: OK, will do.

gunnarmorling:
> So you're saying I should fix the Kafka tests as well, not just Kafka Streams tests?

No, I meant Kafka Streams tests. But they launch Apache Kafka, whose launch configuration I'd expect to be adjusted so to make use of SSL. Maybe I'm missing something?

alesj:
> But they launch Apache Kafka, whose launch configuration I'd expect to be adjusted so to make use of SSL.

This should be done, otherwise it wouldn't work ...

alesj:
@gunnarmorling see changes to KafkaTestResource, its properties

gunnarmorling:
> see KafkaTestResource

Got it, yeah, that's what I was looking for. Not sure how I missed it before :) Thanks!

gunnarmorling (approved):
LGTM; thanks, @alesj!

gunnarmorling:
Hey @gsmet, it looks good to me; perhaps you might double-check on Quarkus stylistics and merge? Thx!

PHameete:
Hi, I've just built a test that is representative for our environment which uses the confluent schema registry and sasl plaintext for authentication.

I was able to configure a working example using the sasl and schema registry runtime properties.

I'm not yet able to check if it also works as a native image, because I have trouble building a native image that works with avro and the confluent schema registry.

gunnarmorling:
Thanks for reporting back; re the issues with native, you wouldn't have to add the registry itself to the native image, this could remain separately. In terms of the actual converters, it'd be interesting to see what problems you encounter. In particular, I'd recommend to check out the serializers from the Apicurio project (which is another schema registry); they are compatible with other registries like Confluent's, too, and I could see us adding support for them by means of flipping the right switches in this extension or maybe the Quarkus Kafka core extension. CC @alesj.

alesj:
Apicurio registry:
* https://github.com/Apicurio/apicurio-registry

PHameete:
For the native image i'll open a different issue at some later time. I don't see us switch away from confluent's SR and (de)serializer but I'm very interested in making it work with native.

I'll let you know what issues I run into and perhaps a minimal example to reproduce at a later time (several weeks from now).

extensions/kafka-streams/runtime/src/main/java/io/quarkus/kafka/streams/runtime/KafkaStreamsTopologyManager.java
```diff
@@ -80,17 +84,90 @@ private static Properties getStreamsProperties(Properties properties, String boo
...
+            setProperty(ssl.cipherSuites, streamsProperties, SslConfigs.SSL_CIPHER_SUITES_CONFIG);
+            setProperty(ssl.enabledProtocols, streamsProperties, SslConfigs.SSL_ENABLED_PROTOCOLS_CONFIG);
+
+            setStoreConfig(ssl.truststore, streamsProperties, "ssl.truststore");
+            setStoreConfig(ssl.keystore, streamsProperties, "ssl.keystore");
+            setStoreConfig(ssl.key, streamsProperties, "ssl.key");
+
+            setProperty(ssl.keymanagerAlgorithm, streamsProperties, SslConfigs.SSL_KEYMANAGER_ALGORITHM_CONFIG);
+            setProperty(ssl.trustmanagerAlgorithm, streamsProperties, SslConfigs.SSL_TRUSTMANAGER_ALGORITHM_CONFIG);
+            setProperty(ssl.endpointIdentificationAlgorithm.map(s -> "\"\"".equals(s) ? "" : s), streamsProperties,
```
gsmet, line 145: I don't understand why we need that here?

gsmet (changes requested):
Hi @alesj ! Sorry for the delay, things have been crazy until now and I'm now looking at the backlog.

I rebased your PR and also added 2 commits.

I added one question inline about some weird filtering you did?

I have one big issue though, I'm not sure we pass the config to Kafka Streams appropriately:
```
2020-04-28 15:38:12,392 WARN  [io.qua.config] (main) Unrecognized configuration key "quarkus.kafka-streams.security.protocol" was provided; it will be ignored
2020-04-28 15:38:13,066 WARN  [org.apa.kaf.cli.adm.AdminClientConfig] (main) The configuration 'sasl.kerberos-ticket-renew-jitter' was supplied but isn't a known config.
2020-04-28 15:38:13,066 WARN  [org.apa.kaf.cli.adm.AdminClientConfig] (main) The configuration 'sasl.login.refresh.buffer.seconds' was supplied but isn't a known config.
2020-04-28 15:38:13,066 WARN  [org.apa.kaf.cli.adm.AdminClientConfig] (main) The configuration 'ssl.endpoint-identification-algorithm' was supplied but isn't a known config.
2020-04-28 15:38:13,066 WARN  [org.apa.kaf.cli.adm.AdminClientConfig] (main) The configuration 'topics' was supplied but isn't a known config.
2020-04-28 15:38:13,067 WARN  [org.apa.kaf.cli.adm.AdminClientConfig] (main) The configuration 'bootstrap-servers' was supplied but isn't a known config.
2020-04-28 15:38:13,067 WARN  [org.apa.kaf.cli.adm.AdminClientConfig] (main) The configuration 'sasl.login-refresh-buffer' was supplied but isn't a known config.
2020-04-28 15:38:13,067 WARN  [org.apa.kaf.cli.adm.AdminClientConfig] (main) The configuration 'schema-registry-key' was supplied but isn't a known config.
2020-04-28 15:38:13,067 WARN  [org.apa.kaf.cli.adm.AdminClientConfig] (main) The configuration 'ssl.truststore.type' was supplied but isn't a known config.
2020-04-28 15:38:13,067 WARN  [org.apa.kaf.cli.adm.AdminClientConfig] (main) The configuration 'sasl.kerberos.ticket.renew.jitter' was supplied but isn't a known config.
2020-04-28 15:38:13,067 WARN  [org.apa.kaf.cli.adm.AdminClientConfig] (main) The configuration 'apicurio.registry.url' was supplied but isn't a known config.
2020-04-28 15:38:13,067 WARN  [org.apa.kaf.cli.adm.AdminClientConfig] (main) The configuration 'ssl.truststore.location' was supplied but isn't a known config.
2020-04-28 15:38:13,067 WARN  [org.apa.kaf.cli.adm.AdminClientConfig] (main) The configuration 'ssl.truststore.password' was supplied but isn't a known config.
2020-04-28 15:38:13,067 WARN  [org.apa.kaf.cli.adm.AdminClientConfig] (main) The configuration 'schema-registry-url' was supplied but isn't a known config.
2020-04-28 15:38:13,067 WARN  [org.apa.kaf.cli.adm.AdminClientConfig] (main) The configuration 'application-id' was supplied but isn't a known config.
2020-04-28 15:38:13,067 WARN  [org.apa.kaf.cli.adm.AdminClientConfig] (main) The configuration 'some-property' was supplied but isn't a known config.
2020-04-28 15:38:13,067 WARN  [org.apa.kaf.cli.adm.AdminClientConfig] (main) The configuration 'ssl.endpoint.identification.algorithm' was supplied but isn't a known config.
```

I thought it might be due to one of my commits but I have the same issue when going back to your commit.

The first Quarkus warning is probably just something that needs to be removed or adjusted in the application.properties but all the Kafka errors look weird to me.

Could you check what's going on? I'll be there to get this thing merged into 1.5.

Thanks!

extensions/kafka-streams/runtime/src/main/java/io/quarkus/kafka/streams/runtime/KafkaStreamsTopologyManager.java
alesj: There is no other (or better / smarter) way to pass-in an empty string.
(that's what I've been told by @dmlloyd )

And in this case you need to have an empty string, to disable this config/setting.

alesj:
> but all the Kafka errors look weird to me

This somehow explains this:
* https://github.com/quarkusio/quarkus/issues/4961#issuecomment-620615544

What you see there is what we also pass into AdminClient creation -- same props from application.properties.
And AdminClient supports only limited number of those props.
To get rid of this, we would need to separate between admin and non-admin props.
Which is imo an overkill, since in most cases you have same bootstrap servers, security, etc

gsmet:
@alesj so in a normal setup, you're saying you won't see all those warnings?

OK, that goes beyond my knowledge. @gunnarmorling @cescoffier WDYT of ^

extensions/kafka-streams/runtime/src/main/java/io/quarkus/kafka/streams/runtime/KafkaStreamsTopologyManager.java
gsmet: OK. Can't say I'm excited about it but can't think of a better solution. Or maybe we could use `-1` to disable? It's a bit of dark magic but could be documented.

Just a proposal, WDYT?

alesj:
Well ... in current/this Quarkus setup you would always see it -- as this add-on/extension uses AdminClient as well.
If a user would hand-pick/custom config/props, it probably wouldn't see this.
But this is harmless warning ... imho

extensions/kafka-streams/runtime/src/main/java/io/quarkus/kafka/streams/runtime/KafkaStreamsTopologyManager.java
alesj: It's a string ... so setting a number ... hmmm, dunno :-)

dmlloyd: > There is no other (or better / smarter) way to pass-in an empty string.
> (that's what I've been told by @dmlloyd )

No it's not; I definitely never told you that.  If you want a string which can be empty, use `Optional<String>` with `orElse("")`.  If you want an integer which can be empty, use `OptionalInt` (`int` has no native empty representation though, so `orElse` won't help in this case).

> And in this case you need to have an empty string, to disable this config/setting.

Then it should be `Optional`.

alesj:
This wont work — not set is not equal to “”.
“” has a special meaning.

dmlloyd:
If an empty string means both an empty string and a special meaning, then the special meaning is wrong.  What is the special meaning that you suggest an empty string corresponds to?

gsmet:
@dmlloyd if you have
```
quarkus.my-config=
```
is the config property value `Optional.empty()` or the empty string?

dmlloyd:
It's `Optional.empty()`.

dmlloyd:
My point is that requiring a user to supply a string of `""` is obviously ridiculous.  If there's some special behavior that has to be specified, then there should be a specific string that corresponds to that special behavior.

gsmet:
Yeah, that's what I was proposing with `-1`. We could potentially use `none` too.

dmlloyd:
An empty `Optional` already represents the idea of "none" or "nothing".  Why not use that?  What meaning were we ascribing to it otherwise?

dmlloyd:
If the default value is `https` then make `https` the default value of the config property, and make it be `Optional`.  A value of `Optional.empty()` means "none", otherwise the value is the value and you don't need any of the `"\"\""` nonsense.

extensions/kafka-streams/runtime/src/main/java/io/quarkus/kafka/streams/runtime/KafkaStreamsTopologyManager.java
dmlloyd, line 145:
```suggestion
            setProperty(ssl.endpointIdentificationAlgorithm.orElse(""), streamsProperties,
```

alesj: How are you gonna get Optional.empty() if default is "https" ... ?
Or it looks like I don't understand how @ConfigItem with defaultValue and Optional work together?

dmlloyd: An empty `Optional` means the value was empty - meaning, the user explicitly gave an empty value.  If the user *does not give* a value, then the default is used, so in this case you'd get an `Optional.of("https")`.

alesj:
Ah, ok ... so this

some.prop=

would translate to Optional.empty()

Yeah, this solves the problem.
Will commit this fix/change.

dmlloyd (dismissed):

dmlloyd:
But the optional part looks good now.

gsmet:
Unfortunately, this won't make it. There's something not working on CI, not sure where it comes from. We have this message going on and on:
```

**Comment by rquinio** [comment]
I've reproduced locally the "SSL handshake failed" issue on graalvm-ce-java11-20.0.0 (but it works fine on openjdk-11.0.5), it seems systematic.

Was the native test passing before the big rebase ?

Btw I think we should put a JUnit timeout on the integration test, since liveness/readiness is not checked externally to shutdown KafkaStreams when something goes wrong, it'll fail faster than waiting for the 65min timeout of the CI.

**Comment by rquinio** [comment]
We'd really need the ability to configure kafka-streams.* properties at runtime in Quarkus 1.6, because some values can vary between platforms and cannot be known at build time.

Unfortunately I don't know how to investigate the "SSL handshake failed" failure that only occurs in native mode... Other extensions like raw Kafka also have SSL native tests which work fine...

An alternative could be to split the PR (ability to configure at runtime / new SSL config & tests).

**Comment by vietk** [comment]
Hello

About the SSL handshake failed, I have some findings :
The Kafka cluster being run by the failsafe plugin it's possible to debug the SSL handshake protocol between the integration test run in native mode, 

with the command : 
```
mvn -Djavax.net.debug=all test-compile failsafe:integration-test
```

here's what I am seeing : 

```
javax.net.ssl|ERROR|4C|data-plane-kafka-network-thread-1-ListenerName(CLIENT)-SSL-1|2020-06-17 01:41:20.146 CEST|TransportContext.java:312|Fatal (INTERNAL_ERROR): problem unwrapping net record (
"throwable" : {
  javax.net.ssl.SSLException: Unrecognized SSL message, plaintext connection?
  	at java.base/sun.security.ssl.SSLEngineInputRecord.bytesInCompletePacket(SSLEngineInputRecord.java:146)
  	at java.base/sun.security.ssl.SSLEngineInputRecord.bytesInCompletePacket(SSLEngineInputRecord.java:64)
  	at java.base/sun.security.ssl.SSLEngineImpl.readRecord(SSLEngineImpl.java:544)
  	at java.base/sun.security.ssl.SSLEngineImpl.unwrap(SSLEngineImpl.java:441)
  	at java.base/sun.security.ssl.SSLEngineImpl.unwrap(SSLEngineImpl.java:420)
  	at java.base/javax.net.ssl.SSLEngine.unwrap(SSLEngine.java:634)
  	at org.apache.kafka.common.network.SslTransportLayer.handshakeUnwrap(SslTransportLayer.java:504)
  	at org.apache.kafka.common.network.SslTransportLayer.doHandshake(SslTransportLayer.java:363)
  	at org.apache.kafka.common.network.SslTransportLayer.handshake(SslTransportLayer.java:286)
  	at org.apache.kafka.common.network.KafkaChannel.prepare(KafkaChannel.java:174)
  	at org.apache.kafka.common.network.Selector.pollSelectionKeys(Selector.java:547)
  	at org.apache.kafka.common.network.Selector.poll(Selector.java:485)
  	at kafka.network.Processor.poll(SocketServer.scala:861)
  	at kafka.network.Processor.run(SocketServer.scala:760)
  	at java.base/java.lang.Thread.run(Thread.java:834)}

)
Jun 17, 2020 1:41:20 AM org.apache.kafka.common.network.Selector pollSelectionKeys
INFO: [SocketServer brokerId=1] Failed authentication with /127.0.0.1 (SSL handshake failed)
```

It seems the SSL layer complains that a non SSL connection is being made upon the the server instance.

Actually looking at the logs of the native compilation we can see the following input :

```
2020-05-20T11:21:10.6006836Z 11:21:10,595 WARN  [io.qua.config] Unrecognized configuration key "quarkus.kafka-streams.security.protocol" was provided; it will be ignored
```

Which make me think that the configuration property from the application.properties has been merely ignored, hence the error at runtime.

I will try to continue investigation when I have time 

Regards
Kevin

vietk:
Hello

Finally I think my feeling was good, there's was an issue with the kafka streams settings "security.protocol". 
This one was not recorded during the native compilation, being left to the default option which is PLAIN_TEXT when run in native mode. 

So the native integration test was failing : the kafka streams application wanted to connect using unsecured channel, hence the SSL handshake failed on the kafka broker.
Also the integration test needed to have enableAllSecurityServices set to true, to be able to run SSL code in the native image. 

I made a commit on my fork on top of @alesj contribution => [here](https://github.com/vietk/quarkus/commits/bugfix/fix-sll-handshake-failed-for-kafka-streams)
The Quarkus-CI also shows that the native test on messaging has been solved => [here](https://github.com/vietk/quarkus/runs/784459441?check_suite_focus=true) 

The remaining CI issue (native Tests - Data 5) is present in both CIs completion, this pull request and on my fork, so I did not bother with it, it's not linked to the content of this PR.

@gsmet : what do you think? can you cherry-pick my commit in order to solve the issue and make this PR move forward? 

Regards

geoand:
In any case,we'll need this PR rebased onto master to solve the merge conflict

vietk:
@alesj can you please rebase your PR please and cherry pick the commit from my fork so that we can move forward ?

@geoand: in case the original contributor is not responding to the PR, can't you force a rebase, as @gsmet did previously in this PR and integrate my commit ?

geoand:
> @geoand: in case the original contributor is not responding to the PR, can't you force a rebase, as @gsmet did previously in this PR and integrate my commit ?

If necessary, sure

alesj:
@vietk @gsmet @geoand I'm fine with whatever you do with my initial PR
Or how you're gonna merge with @vietk 's additional commit, and then rebase.

vietk:
@geoand : if I may, can you force a rebase on the PR branch ?  The merge conflict is easy to solve.

geoand:
PR updated

vietk:
Thanks, @geoand. 
Last request, but not the least : 
I rebased this PR on my fork with [the mandatory commit to fix](https://github.com/vietk/quarkus/commit/51eb83bbc643c529550163851961f944c4eff56c) the native tests, the CI is now green on my fork => https://github.com/vietk/quarkus/actions/runs/143549209
Can you please review and cherry-pick the last commit from my fork and push it to @alesj fork ?

geoand:
@vietk I cherry-picked your commit and pushed it into this PR. 

However I don't have much Kafka experience, so I won't be reviewing the PR. I'll leave that to @gunnarmorling :)

vietk:
@gunnarmorling all checks passed ! Since this PR has already been reviewed and I carried on small changes to make it pass, can you please give a final review ?

gsmet (approved):

gsmet:
Let's get this in and make progress from there.

gsmet:
Thanks everyone and sorry it took so much time to get in.

I think a few things still need to be refined but we can do it iteratively.

PR #7418, opened 2020-02-25
Prevent redirect loops and support a cookie path parameter

### LINKED ISSUE #6802: OIDC goes into a redirect loop if a token request with a code grant fails
**Describe the bug**
The users have reported that `quarkus-oidc` goes into a redirect loop if a token request with a code grant fails.

**Expected behavior**
`401` is returned.

**Actual behavior**
A user is redirected to OIDC server, authenticates, is returned back to with the `code`, Quarkus tries to acquire a token with this code, and if it fails, the user is redirected back again

sberyozkin:
Fixes #6802.

This PR makes the following changes:
- Introduces `AuthenticationCompletionException` to avoid a confusing redirect loop in all the cases when, after the user has already authenticated at the IDP site, the last phase of the authentication flow fails. Two main cases are: missing the state cookie or a code to token exchange failure due to the redirect_uri mismatch. Both of these cases are tested and the log warnings are issued. (I was thinking for a moment to avoid introducing a new exception and instead set a routing context flag and then react to it in the challenge method but decided that having an exception was cleaner)
 
In fact, with the last case (code to token exchange) I've managed, in a weird way, before the fix, to have a test application method successfully invoked despite a code grant failure (multi-tenancy case where after the tenant specific failure a redirect loop has started and a default configuration was picked up). Which is a **major security issue** and has to be fixed before the users will start stressing the multi-tenancy with 1.3.0.

- `cookie-path` property is added and tested to address the case where the `redirect-path` has a root different that of the original request URL which would otherwise cause a redirect loop (before the fix) or `401` with a fix due to a missing state cookie

- Updated `DefaultTenantConfigResolver` to log a debug message when `TenantResolver` returns a non-empty `tenantId` and no matching tenant configuration has been found; I've spent a lot of time figuring out the test issues after I've mistyped `tenant-3` as `tenant3`. I've actually wanted to throw an exception but it might be too forceful for the truly dynamic situations tested in the `oidc-tenancy` (Single ConcurrentHashMap is an alternative but it feels it is not too bad having the dedicated static and dynamic maps)

- Updated `DefaultTenantConfigResolver` to keep `TenantConfigContext`s created with the help from `TenantConfigResolver` in its own map to avoid a possible thread safety issue (one thread gets `TenantConfigResolver` statically created at the startup, another one puts a new `TenantConfigContext`) and updated a lock to minimize a possible contention too.

### DIFF
--- a/extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/CodeAuthenticationMechanism.java
+++ b/extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/CodeAuthenticationMechanism.java
@@ -17,10 +17,10 @@ import io.quarkus.oidc.AccessTokenCredential;
 import io.quarkus.oidc.IdTokenCredential;
 import io.quarkus.oidc.RefreshToken;
 import io.quarkus.oidc.runtime.OidcTenantConfig.Authentication;
-import io.quarkus.security.AuthenticationFailedException;
 import io.quarkus.security.identity.IdentityProviderManager;
 import io.quarkus.security.identity.SecurityIdentity;
 import io.quarkus.security.runtime.QuarkusSecurityIdentity;
+import io.quarkus.vertx.http.runtime.security.AuthenticationCompletionException;
 import io.quarkus.vertx.http.runtime.security.AuthenticationRedirectException;
 import io.quarkus.vertx.http.runtime.security.ChallengeData;
 import io.vertx.core.http.Cookie;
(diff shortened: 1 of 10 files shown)


sberyozkin:
Re the failing [Windows build](https://github.com/quarkusio/quarkus/pull/7418/checks?check_run_id=467549697):
```
[INFO] Quarkus - Integration Tests - OpenID Connect Adapter Code Flow SUCCESS [  2.018 s]
[INFO] Quarkus - Integration Tests - Multi-Tenant OpenID Connect Adapter SUCCESS [  4.788 s]
[INFO] Quarkus - Integration Tests - Keycloak Policy Enforcer SUCCESS [  1.641 s]
```

extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/CodeAuthenticationMechanism.java
```diff
@@ -182,16 +188,19 @@ private static QuarkusSecurityIdentity augmentIdentity(SecurityIdentity security
 
         configContext.auth.authenticate(params, userAsyncResult -> {
             if (userAsyncResult.failed()) {
-                cf.completeExceptionally(new AuthenticationFailedException());
+                if (userAsyncResult.cause() != null) {
+                    LOG.warnf("Exception during the code to token exchange: %s", userAsyncResult.cause().getMessage());
```
stuartwdouglas: This should be DEBUG if it is something a user can trigger with a malformed request. In general you don't want to log anything that a user can trigger higher than DEBUG, as it opens you up to DOS attacks because logging is very expensive.

stuartwdouglas (changes requested):
Just a single minor comment, otherwise looks good

extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/CodeAuthenticationMechanism.java
sberyozkin: @stuartwdouglas done

stuartwdouglas (approved):

sberyozkin:
@stuartwdouglas Thanks Stuart.
@pedroigor Hey Pedro, can you please review too, will address your comments if any. Thanks

pedroigor (approved):

PR #7423, opened 2020-02-25
fix Pass default value to PathParam in Vertx doc

tivrfoa:


gsmet (changes requested):
So we usually use the RESTEasy-specific `PathParam` which doesn't need the name and can extract it from the parameter names.

What's missing here is the imports so that you don't have to do that.

tivrfoa:
> So we usually use the RESTEasy-specific `PathParam` which doesn't need the name and can extract it from the parameter names.
> 
> What's missing here is the imports so that you don't have to do that.

There is no pattern in the guides regarding @PathParam, sometimes from "org.jboss.resteasy.annotations.jaxrs" and sometimes from "javax.ws.rs" like in https://quarkus.io/guides/reactive-messaging
So you want to use from "org.jboss.resteasy.annotations.jaxrs", right?

gsmet:
@tivrfoa sorry, I missed your message. Yes, we favor the RESTEasy ones as they are more flexible.

PR #7320, opened 2020-02-20
[WIP] Support both GraalVM 19.3.1 and 20.0.0 - 19.3.1 SDK & 20.0.0 GraalVM image

gwenneg:
This PR was created to see if the GraalVM `19.3.1` SDK and the GraalVM `20.0.0` image are compatible.

More details in #7269.

PR #7427, opened 2020-02-25
Add Gradle support in multi-stage docker build for native images

### LINKED ISSUE #6513: Provide Gradle GraalVM compatible builder image
**Description**
In [creating-a-container-with-a-multi-stage-docker-build](https://quarkus.io/guides/building-native-image#creating-a-container-with-a-multi-stage-docker-build) guide there is a Maven docker base image with GraalVM capabilities.

It would be nice if there was one for projects using Gradle as well.

fmhwong:
https://github.com/quarkusio/quarkus-images/pull/54 has added Gradle into the yaml files.  This PR is just to update the docs.

Fixes #6513

### DIFF
--- a/docs/src/main/asciidoc/building-native-image.adoc
+++ b/docs/src/main/asciidoc/building-native-image.adoc
@@ -340,10 +340,11 @@ NOTE: Interested by tiny Docker images, check the {quarkus-images-url}/master/di
 
 The previous section showed you how to build a native executable using Maven, but implicitly required that the proper GraalVM version be installed on the building machine (be it your local machine or your CI/CD infrastructure).
 
-In cases where the GraalVM requirement cannot be met, you can use Docker to perform the Maven build by using a multi-stage Docker build. A multi-stage Docker build is like two Dockerfile files combined in one, the first is used to build the artifact used by the second.
+In cases where the GraalVM requirement cannot be met, you can use Docker to perform the Maven or Gradle build by using a multi-stage Docker build. A multi-stage Docker build is like two Dockerfile files combined in one, the first is used to build the artifact used by the second.
 
-In this guide we will use the first stage to generate the native executable using Maven and the second stage to create our runtime image.
+In this guide we will use the first stage to generate the native executable and the second stage to create our runtime image.
 
+Sample Dockerfile for building with Maven:
 [source,dockerfile,subs=attributes+]
 ----
 ## Stage 1 : build with maven builder image with native capabilities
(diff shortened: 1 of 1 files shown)


build-parent/pom.xml
```diff
@@ -32,7 +32,7 @@
         <!-- These properties are needed in order for them to be resolvable by the documentation -->
         <!-- The Graal version we suggest using in documentation - as that's
            what we work with by self downloading it: -->
-        <graal-sdk.version-for-documentation>19.3.1</graal-sdk.version-for-documentation>
+        <graal-sdk.version-for-documentation>20.0.0</graal-sdk.version-for-documentation>
```
gastaldi: I am not sure we upgraded to GraalVM 20 already

fmhwong: This PR https://github.com/quarkusio/quarkus/pull/7427 updated the version to 20.0.0.

gastaldi: Right, the image is there, but we need #7320 and #7321 to be merged first to officially say we support GraalVM 20 :wink:

gsmet: We need to go back to 19.3.1 here.

docs/src/main/asciidoc/building-native-image.adoc
```diff
@@ -364,8 +365,30 @@ EXPOSE 8080
...
+FROM registry.access.redhat.com/ubi8/ubi-minimal
+WORKDIR /work/
+COPY --from=build /usr/src/app/build/*-runner /work/application
+RUN chmod 775 /work
+EXPOSE 8080
+CMD ["./application", "-Dquarkus.http.host=0.0.0.0"]
+----
 
-Save this file in `src/main/docker/Dockerfile.multistage` as it is not included in the getting started quickstart.
+Pick one of the above Dockerfile samples and save it in `src/main/docker/Dockerfile.multistage` as it is not included in the getting started quickstart.
```
gsmet: This needs a bit of rewording IMHO because the getting started project is a Maven one. So it makes no sense putting the Gradle one there.

I would put that sentence just after the Maven example. And then have something like "If you are using Gradle in your project, you can use this sample Dockerfile:"

gsmet (changes requested):

build-parent/pom.xml
fmhwong: Version reverted back to 19.3.1

gsmet (approved):

PR #7429, opened 2020-02-26
Add code example for java 11 native dockerfile to native-and-ssl guide

andreas-repp:
In the Java 11 GraalVM image, the location of cacerts and libsunec.so is different. I added a code sample showing the paths to be used if Java 11 is used.

### DIFF
--- a/docs/src/main/asciidoc/native-and-ssl.adoc
+++ b/docs/src/main/asciidoc/native-and-ssl.adoc
@@ -203,6 +203,7 @@ When working with containers, the idea is to bundle both the SunEC library and t
 
 You can for example modify your `Dockerfile.native` as follows to copy the required files to your final image:
 
+==== Java 8
 [source, subs=attributes+]
 ----
 FROM quay.io/quarkus/ubi-quarkus-native-image:{graalvm-version}-java8 as nativebuilder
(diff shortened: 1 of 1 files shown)


docs/src/main/asciidoc/native-and-ssl.adoc
```diff
@@ -219,6 +220,23 @@ EXPOSE 8080
...
+----
+FROM quay.io/quarkus/ubi-quarkus-native-image:{graalvm-version}-java11 as nativebuilder
+RUN mkdir -p /tmp/ssl-libs/lib \
+  && cp /opt/graalvm/lib/security/cacerts /tmp/ssl-libs \
+  && cp /opt/graalvm/lib/libsunec.so /tmp/ssl-libs/lib/
+
+FROM registry.access.redhat.com/ubi8/ubi-minimal
+WORKDIR /work/
+COPY --from=nativebuilder /tmp/ssl-libs/ /work/
+COPY build/*-runner /work/application
```
geoand: Let's try and keep this the same as the Java 8 example. I assume you were using Gradle when you tried this :)

geoand (changes requested):

docs/src/main/asciidoc/native-and-ssl.adoc
andreas-eberle: Thanks for the hint! I fixed it. 

Btw, is there a "selection" thing for the docs to provide code for different setups and let the user easily switch between them? E.g. Java/Kotlin with Maven/Gradle? I imagine something like you can see on the [Android Developers pages with Java/Kotlin](https://developer.android.com/topic/libraries/architecture/livedata#kotlin). I think this could really help the docs in the future when you make Quarkus fully work with the combinations mentioned above.

geoand: We don't have that yet, but that's something we'll definitely need to address in the future

geoand (approved):
Thanks for this!

PR #7430, opened 2020-02-26
Config viewer as part of vertx-http in dev mode

### LINKED ISSUE #1349: Extension to view the values configured by MicroProfile config in development mode

hpehl:
Fixes: #1349.
Supersedes https://github.com/quarkusio/quarkus/pull/5692.

Adds a route in dev mode which lists the configured values for MicroProfile Configuration. The route is registered using the path "/config" by default and lists all properties of all config sources. The config sources are sorted descending by ordinal, the properties by name. If no config is defined an empty JSON object is returned.

### DIFF
--- a/bom/runtime/pom.xml
+++ b/bom/runtime/pom.xml
@@ -2244,6 +2244,11 @@
             </dependency>
             <dependency>
                 <groupId>io.vertx</groupId>
+                <artifactId>vertx-core</artifactId>
+                <version>${vertx.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>io.vertx</groupId>
                 <artifactId>vertx-web</artifactId>
                 <version>${vertx.version}</version>
             </dependency>
--- a/bom/runtime/pom.xml
+++ b/bom/runtime/pom.xml
@@ -2244,11 +2244,6 @@
             </dependency>
             <dependency>
                 <groupId>io.vertx</groupId>
-                <artifactId>vertx-core</artifactId>
-                <version>${vertx.version}</version>
-            </dependency>
-            <dependency>
-                <groupId>io.vertx</groupId>
                 <artifactId>vertx-web</artifactId>
                 <version>${vertx.version}</version>
             </dependency>
--- a/extensions/vertx-http/runtime/pom.xml
+++ b/extensions/vertx-http/runtime/pom.xml
@@ -50,6 +50,10 @@
                 </exclusion>
             </exclusions>
         </dependency>
+        <dependency>
+            <groupId>io.vertx</groupId>
+            <artifactId>vertx-core</artifactId>
+        </dependency>
     </dependencies>
 
     <build>
--- a/extensions/vertx-http/runtime/pom.xml
+++ b/extensions/vertx-http/runtime/pom.xml
@@ -50,10 +50,6 @@
                 </exclusion>
             </exclusions>
         </dependency>
-        <dependency>
-            <groupId>io.vertx</groupId>
-            <artifactId>vertx-core</artifactId>
-        </dependency>
     </dependencies>
 
     <build>
--- a/extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/devmode/ConfigViewer.java
+++ b/extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/devmode/ConfigViewer.java
@@ -1,0 +1,77 @@
+package io.quarkus.vertx.http.runtime.devmode;
+
+import java.util.Set;
+import java.util.SortedSet;
+import java.util.TreeSet;
+
+import org.eclipse.microprofile.config.Config;
+import org.eclipse.microprofile.config.spi.ConfigSource;
+import org.jboss.logging.Logger;
+
+import io.vertx.core.json.JsonArray;
+import io.vertx.core.json.JsonObject;
+
+/**
+ * Turns a {@link Config} into a JSON object with all config sources and properties as JSON. The config sources are
+ * sorted descending by ordinal, the properties by name. If no config is defined an empty JSON object is returned.
+ *
+ * <p>
+ * A typical output might look like:
+ * </p>
+ *
+ * <pre>
+ * {
+ *   "sources": [
+ *     {
+ *       "source": "source0",
+ *       "ordinal": 200,
+ *       "properties": {
+ *         "key": "value"
+ *       }
+ *     },
+ *     {
+ *       "source": "source1",
+ *       "ordinal": 100,
+ *       "properties": {
+ *         "key": "value"
+ *       }
+ *     }
+ *   ]
+ * }
+ * </pre>
+ */
+class ConfigViewer {
+
+    private static final Logger LOGGER = Logger.getLogger(ConfigViewer.class.getName());
+
+    JsonObject dump(Config config) {
+        JsonObject json = new JsonObject();
+        if (config != null) {
+            if (config.getConfigSources().iterator().hasNext()) {
+                JsonArray jsonSources = new JsonArray();
+                for (ConfigSource source : config.getConfigSources()) {
+                    JsonObject jsonSource = new JsonObject();
+                    jsonSource.put("source", source.getName())
+                            .put("ordinal", source.getOrdinal());
+                    Set<String> propertyNames = source.getPropertyNames();
+                    if (!propertyNames.isEmpty()) {
+                        SortedSet<String> sortedPropertyNames = new TreeSet<>(propertyNames);
+                        JsonObject jsonProperties = new JsonObject();
+                        for (String propertyName : sortedPropertyNames) {
+                            try {
+                                jsonProperties.put(propertyName, source.getValue(propertyName));
+                            } catch (Throwable t) {
+                                LOGGER.errorf("Cannot get configuration value for '%s': %s",
+                                        propertyName, t.getMessage());
+                            }
+                        }
+                        jsonSource.put("properties", jsonProperties);
+                    }
+                    jsonSources.add(jsonSource);
+                }
+                json.put("sources", jsonSources);
+            }
+        }
+        return json;
+    }
+}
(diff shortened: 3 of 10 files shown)


extensions/vertx-http/runtime/pom.xml
```diff
@@ -50,6 +50,10 @@
                 </exclusion>
             </exclusions>
         </dependency>
+        <dependency>
```
geoand: Why is this needed? Isn't the `io.vertx:vertx-web` dependency added above enough?

hpehl: I've tried w/o, but got a `java.lang.ClassNotFoundException: 
```
[INFO] Running io.quarkus.vertx.http.configviewer.ShowConfigTest
2020-02-26 10:57:52,162 INFO  [io.quarkus] (main) Quarkus 999-SNAPSHOT started in 1.513s. Listening on: http://0.0.0.0:8080
2020-02-26 10:57:52,162 INFO  [io.quarkus] (main) Profile dev activated. Live Coding activated.
2020-02-26 10:57:52,162 INFO  [io.quarkus] (main) Installed features: [cdi, security]
2020-02-26 10:57:53,319 ERROR [io.qua.ver.htt.run.QuarkusErrorHandler] (vert.x-eventloop-thread-4) HTTP Request to /config failed, error id: f23efab7-ebff-4242-b10e-c4efa772f718-1: java.lang.NoClassDefFoundError: com/fasterxml/jackson/databind/JsonSerializer
	at io.vertx.core.json.jackson.JacksonCodec.toString(JacksonCodec.java:151)
	at io.vertx.core.json.JsonObject.encode(JsonObject.java:785)
	at io.quarkus.vertx.http.runtime.devmode.ConfigViewerHandler.handle(ConfigViewerHandler.java:19)
	at io.quarkus.vertx.http.runtime.devmode.ConfigViewerHandler.handle(ConfigViewerHandler.java:11)
	at io.vertx.ext.web.impl.BlockingHandlerDecorator.lambda$handle$0(BlockingHandlerDecorator.java:48)
	at io.vertx.core.impl.ContextImpl.lambda$executeBlocking$2(ContextImpl.java:316)
	at io.vertx.core.impl.TaskQueue.run(TaskQueue.java:76)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30)
	at java.lang.Thread.run(Thread.java:748)
Caused by: java.lang.ClassNotFoundException: com.fasterxml.jackson.databind.JsonSerializer
	at java.net.URLClassLoader.findClass(URLClassLoader.java:382)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:349)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
	at io.quarkus.bootstrap.classloading.QuarkusClassLoader.loadClass(QuarkusClassLoader.java:291)
	at io.quarkus.bootstrap.classloading.QuarkusClassLoader.loadClass(QuarkusClassLoader.java:251)
	... 11 more
```

bom/runtime/pom.xml
```diff
@@ -2242,6 +2242,11 @@
                 <artifactId>smallrye-mutiny-vertx-web-client</artifactId>
                 <version>${mutiny-client.version}</version>
             </dependency>
+            <dependency>
+                <groupId>io.vertx</groupId>
+                <artifactId>vertx-core</artifactId>
+                <version>${vertx.version}</version>
+            </dependency>
```
cescoffier: This dependency is already defined in the vert.x bom. Isn't it?

extensions/vertx-http/runtime/pom.xml
geoand: Ah yes, that is because we exclude Jackson databind - and we need to continue doing this, we can't let it sneak in.
Can the JSON part you PR just be based on Jackson core instead of relying on databind?

cescoffier: Yes, using JSON Object "encode" requires Jackson, which we want to avoid in vertx-http. It may require a bit of code, but in your case, Jackson should not be required if you just implement a simple "encode" method doing the ToJson transformation manually. As you know the structure of the result, you can just generate the JSON string.

cescoffier (changes requested):

extensions/vertx-http/runtime/pom.xml
hpehl: Ok, I'm going to remove the dependency to Jackson and do the JSON encoding manually.

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/devmode/ConfigViewer.java
```diff
@@ -0,0 +1,77 @@
...
+                    JsonObject jsonSource = new JsonObject();
+                    jsonSource.put("source", source.getName())
+                            .put("ordinal", source.getOrdinal());
+                    Set<String> propertyNames = source.getPropertyNames();
+                    if (!propertyNames.isEmpty()) {
+                        SortedSet<String> sortedPropertyNames = new TreeSet<>(propertyNames);
+                        JsonObject jsonProperties = new JsonObject();
+                        for (String propertyName : sortedPropertyNames) {
+                            try {
+                                jsonProperties.put(propertyName, source.getValue(propertyName));
```
dmlloyd: It looks like this is going to print every config value from every config source (as opposed to printing only the effective values).  I assume that this was your intent!

One concern I have is that this will expand property expressions.  If you don't want that (and I suspect you don't since you're printing all the raw config keys rather than just the effective configuration contents), then the whole thing should be enclosed like this:

```java
boolean old = ExpandingConfigSource.setExpanding(false);
try {
    // ... do all the work here ...
} finally {
    ExpandingConfigSource.setExpanding(old);
}

```

dmlloyd: Another concern I have is that this might expose secrets/passwords!

hpehl: Thanks @dmlloyd 
Right, I only want to print the effective values. Will wrap the code like you suggested.

I don't have a solution how to hide secrets / passwords.

dmlloyd: Let me post something to the dev list on this. I have an idea... I think.

dmlloyd: Actually I'll post an issue for it...

dmlloyd: I've opened #7442 for this.

dmlloyd (changes requested):
I hate to do this but I think we should :pause_button: this until we solve #7442.  If others vehemently disagree I'll lift this review but I think we should be security-minded here.

hpehl:
In any case I've updated the PR:

1. Remove dependency to Jackson data binding
1. wrap code in `ExpandingConfigSource.setExpanding(false)`

hpehl:
The JSON code is based on https://github.com/stleary/JSON-java. 
Third commit adds the relevant license headers.

gastaldi:
One poor-man solution is to check if the key contains `password` and avoid showing it (or showing anyway, since you're in Dev mode, I don't think this is much of an issue)

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/devmode/JsonObject.java
```diff
@@ -0,0 +1,170 @@
...
+ of this software and associated documentation files (the "Software"), to deal
+ in the Software without restriction, including without limitation the rights
+ to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ copies of the Software, and to permit persons to whom the Software is
+ furnished to do so, subject to the following conditions:
+
+ The above copyright notice and this permission notice shall be included in all
+ copies or substantial portions of the Software.
+
+ The Software shall be used for Good, not Evil.
```
gsmet, line 14: Unfortunately, this makes it non-free software. We can't include it.

gsmet (changes requested):
Added a concern about the license of the added JSON classes.

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/devmode/JsonObject.java
gastaldi: Better use Jackson here

machi1990: > Better use Jackson here

Ah wouldn't that meaning pulling Jackson to the core?

gastaldi: Ah I missed @cescoffier comments in https://github.com/quarkusio/quarkus/pull/7430#discussion_r384390220, Jackson is not an option here

mkouba:
FYI in [this PR](https://github.com/quarkusio/quarkus/pull/9412) there is a minimalistic JSON string generator (adapted from the [Weld Probe](https://github.com/weld/core/blob/master/probe/core/src/main/java/org/jboss/weld/probe/Json.java)) and I believe that it could be used in this PR as well.

hpehl:
Thanks @mkouba. I'll have a look asap.

gastaldi:
Closing as it needs a rebase and the PR needs to reuse the existing JSON libraries instead. Please reopen when done

PR #6923, opened 2020-01-31
doc: active record and repository patterns

loicmathieu:
Fixes #6135
Fixes #6945

The following documentation changes are, for both MongoDB and Hibernate with Panache:

- Rework the documentation to clearly present the two _Panache flavours_ as equals in terms of support/implementation/...
- Clearly name the two pattens : active records and repository
- Add more repository documentation examples
- Restructure the documentation

A PR on the quickstart will follows to duplicate the code to have exampes for both the active record and the repository patterns.

PR #7431, opened 2020-02-26
Provides multi-tenancy capabilities for MongoDB with Panache

### LINKED ISSUE #5183: Multitenancy with mongodb panache extension
**Description**
We're building an app which implement multitenancy by having different mongo databases.

Currently, there seems to be no way to specify which database should be used by PanacheMongoRepository implementations depending on dynamic values like http requests headers or path params.

Currently, since Panache get the database value by calling [ConfigProvider.getConfig()](https://github.com/quarkusio/quarkus/blob/b347c6d9e96b4984013da76e125a7a70a546d35e/extensions/panache/mongodb-panache/runtime/src/main/java/io/quarkus/mongodb/panache/runtime/MongoOperations.java#L272), the only workaround we see is to implement a custom microprofile ConfigSource to provide database value depending on ThreadLocal content set by a request filter earlier. 

Is there any better way to do it or do you plan to implement this feature in future releases ?

**Implementation ideas**
Injecting a custom database provider into PanacheMongoRepository implementations to resolve database to use at runtime may be a way to resolve this issue.

Thanks for your input on the matter.

loicmathieu:
Fixes #5183

~Documentation is missing, waiting for #6923 to be merged before updating the guide~

### DIFF
--- a/docs/src/main/asciidoc/mongodb-panache.adoc
+++ b/docs/src/main/asciidoc/mongodb-panache.adoc
@@ -98,7 +98,7 @@ To get started:
 
 * add your settings in `{config-file}`
 * Make your entities extend `PanacheMongoEntity` (optional if you are using the repository pattern)
-* Use the `@MongoEntity` annotation to specify the name of the database and the name of the collection (it will default to the name of your entity).
+* Optionally, use the `@MongoEntity` annotation to specify the name of the collection, the name of the database or the name of the client.
 
 Then add the relevant configuration properties in `{config-file}`.
 
@@ -110,7 +110,13 @@ quarkus.mongodb.connection-string = mongodb://mongo1:27017,mongo2:27017
 quarkus.mongodb.database = person
 ----
 
-The `quarkus.mongodb.database` property will be used by MongoDB with Panache to determine the name of the database where your entities will be persisted.
+The `quarkus.mongodb.database` property will be used by MongoDB with Panache to determine the name of the database where your entities will be persisted (if not overridden by `@MongoEntity`).
+
+The `@MongoEntity` annotation allows configuring:
+
+* the name of the client for multi-tenant application, see link:mongodb#multiple-mongodb-clients[Multiple MongoDB Clients]. Otherwise, the default client will be used.
+* the name of the database, otherwise, the `quarkus.mongodb.database` property will be used.
+* the name of the collection, otherwise the simple name of the class will be used.
 
 For advanced configuration of the MongoDB client, you can follow the link:mongodb#configuring-the-mongodb-database[Configuring the MongoDB database guide].
 
@@ -119,7 +125,7 @@ For advanced configuration of the MongoDB client, you can follow the link:mongod
 === Defining your entity
 
 To define a Panache entity, simply extend `PanacheMongoEntity` and add your columns as public fields.
-You can add the `@MongoEntity` annotation to your entity if you need to customize the name of the collection and/or the database.
+You can add the `@MongoEntity` annotation to your entity if you need to customize the name of the collection, the database, or the client.
 
 [source,java]
 ----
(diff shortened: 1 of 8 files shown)


loicmathieu:
@geoand with your changes on the MongoDB client build my multi-tenancy PR is now working :tada: 

But I wonder if I can improve it:

- I didn't consume any MongoDB related build items but the list of the needed MongoDB client can be known at build time by scanning the `@MongoEntity` annotations
- I retrieve the `MongoClient` from Arc each time it's needed, I don't know if it is performant or not but maybe I can do this once for all inside a static intializer or inside a `RUNTIME_INIT` build step that store them via a recorder ... WDYT ?

geoand:
> @geoand with your changes on the MongoDB client build my multi-tenancy PR is now working 
>

Glad to hear it!
 
> But I wonder if I can improve it:
> 
> * I didn't consume any MongoDB related build items but the list of the needed MongoDB client can be known at build time by scanning the `@MongoEntity` annotations
> * I retrieve the `MongoClient` from Arc each time it's needed, I don't know if it is performant or not but maybe I can do this once for all inside a static intializer or inside a `RUNTIME_INIT` build step that store them via a recorder ... WDYT ?

Essentially yes, this is why I don't like the `MongoClienBuildItem` and why I don't see the point of it. If you remember I had a PR that removed it at one point :)
The only consumer of `MongoClientBuildItem` seems to be Camel so they would have to vet for its removal.
I am of course +1 for removing

FroMage (changes requested):
The code LGTM, but I'm confused as to the relation between clients and databases.

IMO if you have two databases you get two clients: one for each DB, and they should not be addressed by client name but by DB name. But perhaps this is different in Mongo as it is in ORM?

loicmathieu:
@FroMage In MongoDB you can have multiple databases by MongoDB instance, all would be accessible using the same client (it's closest to the schema of a JDBC database than to a database).
Multiple clients is really multiple database hosts (clusters in fact).

For me, with JDBC, you create a connection by database, not to a database engine, so it's different. You can have multiple database in a single Postgres server but you will need to create separate connections for them ...

That's why we choose the name `client` and not connection to clearly make the distinction between the two.

Maybe the documentation can be improved a bit,

FroMage:
OK, in that case, IF it's true that mongo users are fine understanding that a mongo DB is an SQL schema, and a mongo client is a JDBC connection, then sure, keep that terminology. Not sure it's worth documenting it for SQL/JDBC people like me, if it's such common knowledge for mongo users. After all, this extension assumes you have some knowledge of mongo :)

FroMage (approved):

loicmathieu:
>  Not sure it's worth documenting it for SQL/JDBC people like me, if it's such common knowledge for mongo users

Maybe we can ask @geoand if the current wordings are clear enought, more eyes, better naming ;)

> After all, this extension assumes you have some knowledge of mongo :)

Yes, but as we aim to provides very easy to use API, just a little knowledge should be enought ;)

geoand:
> @FroMage In MongoDB you can have multiple databases by MongoDB instance, all would be accessible using the same client (it's closest to the schema of a JDBC database than to a database).
> Multiple clients is really multiple database hosts (clusters in fact).
> 
> For me, with JDBC, you create a connection by database, not to a database engine, so it's different. You can have multiple database in a single Postgres server but you will need to create separate connections for them ...
> 
> That's why we choose the name `client` and not connection to clearly make the distinction between the two.
> 

I think this is great information to have in the documentation. I don't think this is exactly common knowledge for most people.
Your comparison to JDBC is great is because that is what everyone knows - I would to have that knowledge preserved in the guide.

Perhaps a follow up PR?

loicmathieu:
> Perhaps a follow up PR?

You now it will imply as much work for you to correct it than for me to write it ;)

I'll add it to my TODO list ...

geoand:
> > Perhaps a follow up PR?
> 
> You now it will imply as much work for you to correct it than for me to write it ;)

And then you spotting the missing words in my corrections :laughing:

loicmathieu:
> And then you spotting the missing words in my corrections laughing

Team works !

loicmathieu:
@geoand this PR is validated. Can you merge it ?

geoand:
Sure, I forgot about it!

geoand (approved):

PR #7438, opened 2020-02-26
Avoid having to load ResteasyContextProvider class

cescoffier:
Reasteasy Context Propagation may not be in the classpath, so the previous code led to ClassNotFound error.


cescoffier:
@loicmathieu With this, the Mongo Panache IT would not require the dependency on SmallRye Context Propagation anymore.

loicmathieu:
OK, I'll test it when it'll be merged

gsmet (approved):

PR #7443, opened 2020-02-26
Implements projection inside reactive MongoDB with Panache

loicmathieu:
@cescoffier this is a leftover from the reactive implementation of MongoDB with Panache.
I forget to implements the `project(Class)` methods that was merged recently.

### DIFF
--- a/extensions/panache/mongodb-panache/runtime/src/main/java/io/quarkus/mongodb/panache/runtime/PanacheQueryImpl.java
+++ b/extensions/panache/mongodb-panache/runtime/src/main/java/io/quarkus/mongodb/panache/runtime/PanacheQueryImpl.java
@@ -23,7 +23,6 @@ import io.quarkus.panache.common.exception.PanacheQueryException;
 
 public class PanacheQueryImpl<Entity> implements PanacheQuery<Entity> {
     private MongoCollection collection;
-    private Class<? extends Entity> entityClass;
     private Document mongoQuery;
     private Document sort;
     private Document projections;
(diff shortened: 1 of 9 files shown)


loicmathieu:
@cescoffier can you review it ? 
@gsmet It will be good to have it for 1.3 as it's the only difference between the imperative and the reactive implementation.

loicmathieu:
@cescoffier can you review it ? It's a missing part in the reactive MongoDB with panache implementation.

extensions/panache/mongodb-panache/runtime/src/main/java/io/quarkus/mongodb/panache/runtime/PanacheQueryImpl.java
```diff
@@ -40,7 +39,6 @@
     PanacheQueryImpl(MongoCollection<? extends Entity> collection, Class<? extends Entity> entityClass, Document mongoQuery,
```
FroMage: Might as well remove the parameter if we're not using it, no?

FroMage (changes requested):
Aside from the param to remove, LGTM.

extensions/panache/mongodb-panache/runtime/src/main/java/io/quarkus/mongodb/panache/runtime/PanacheQueryImpl.java
loicmathieu: Yes, this was copied from Hibernate with Panache but not needed here.
I remove it (and also for the reactive implementation), a little cleanup doesn't hurt ;)

FroMage (approved):

PR #7448, opened 2020-02-27
Fix a connection-delay processing

### LINKED ISSUE #7447: OIDC Connection-Delay config parameter doesn't work
**Describe the bug**
If the quarkus.oidc.connection-delay configuration parameter startup fails right away if the OIDC server is not (yet) available instead of retrying for the configured delay period.

**Expected behavior**
If the OIDC identity provider (IDP) is not available, I expect that the quarkus OIDC setup to retry the connect every 2 seconds until the configured delay has expired.  If the IDP is available then quarkus server should start immediately.

**Actual behavior**
Message is logged "Failed to start quarkus" with a stack trace with a root "caused by" java.net.ConnectException: Connection refused: no further information

**To Reproduce**
1. Ensure IDP server is not running
2. Start Quarkus app: I used Eclipse Maven launch:  clean install quarkus:dev 

**Configuration**
```properties
quarkus.http.port=8080


quarkus.arc.remove-unused-beans=framework


# DEBUG console logging
quarkus.log.console.enable=true
quarkus.log.console.format=%d{yyyy-MM-dd HH:mm:ss} %-5p [%c] %s%e%n
quarkus.log.console.level=TRACE
quarkus.log.console.color=false

# TRACE file logging
quarkus.log.file.enable=true
quarkus.log.file.path=/tmp/quarkus.log
quarkus.log.file.level=TRACE
quarkus.log.file.format=%d{yyyy-MM-dd HH:mm:ss} %-5p [%c{2.}]] (%t) %s%e%n

quarkus.log.level=INFO

# OpenID Connect properties -- depends on io.quarkus:quarkus-oidc 
quarkus.oidc.enabled=true
quarkus.oidc.application-type=service
quarkus.oidc.auth-server-url=http://localhost:9090/auth/realms/Quarkus
quarkus.oidc.client-id=quarkus
quarkus.oidc.credentials.secret=592a1849-3bc1-4cdb-92aa-4556f99a3a5a

quarkus.oidc.connection-delay=PT30S

```

 - Output of `uname -a` or `ver`: Microsoft Windows [Version 10.0.18362.657]
 - Output of `java -version`: 
 - GraalVM version (if different from Java): (not run under GraalVM)
 - Quarkus version or git rev: 
		<quarkus.version>1.2.1.Final</quarkus.version>

**Additional context**
Dependency at issue is io.quarkus:quarkus-oidc
OIDC IDP server was keycloak.  

When debugging I observed (what I assume to be) the problem in `OidcRecorder.java` in the section of code between lines 91 and 119 (as of the most recent commit) which is a `for` loop.  In particular the catch clause is incorrect (i believe).  It tries to catch an `OIDCException`, however what is actually being thrown is a `java.util.concurrent.CompletionException`, which wraps the OIDCException.  This exception is thrown from the `cf.join()` at line 105.

Here is the code I am referring to starting at line 91 of `OidcRecorder.java`:

        for (long i = 0; i < connectionRetryCount; i++) {
            try {
                CompletableFuture<OAuth2Auth> cf = new CompletableFuture<>();
                KeycloakAuth.discover(vertx, options, new Handler<AsyncResult<OAuth2Auth>>() {
                    @Override
                    public void handle(AsyncResult<OAuth2Auth> event) {
                        if (event.failed()) {
                            cf.completeExceptionally(toOidcException(event.cause()));
                        } else {
                            cf.complete(event.result());
                        }
                    }
                });

                auth = cf.join();
                break;
            } catch (OIDCException ex) {
                if (i + 1 < connectionRetryCount) {
                    try {
                        Thread.sleep(2000);
                    } catch (InterruptedException iex) {
                        // continue connecting
                    }

                } else {
                    throw ex;
                }
            }
        }

sberyozkin:
Fixes #7447

### DIFF
--- a/extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/OidcRecorder.java
+++ b/extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/OidcRecorder.java
@@ -3,6 +3,7 @@ package io.quarkus.oidc.runtime;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionException;
 import java.util.function.Function;
 
 import org.jboss.logging.Logger;
@@ -119,16 +120,22 @@ public class OidcRecorder {
 
                 auth = cf.join();
                 break;
-            } catch (OIDCException ex) {
-                if (i + 1 < connectionRetryCount) {
-                    try {
-                        Thread.sleep(2000);
-                    } catch (InterruptedException iex) {
-                        // continue connecting
+            } catch (Throwable throwable) {
+                while (throwable instanceof CompletionException && throwable.getCause() != null) {
+                    throwable = throwable.getCause();
+                }
+                if (throwable instanceof OIDCException) {
+                    if (i + 1 < connectionRetryCount) {
+                        try {
+                            Thread.sleep(2000);
+                        } catch (InterruptedException iex) {
+                            // continue connecting
+                        }
+                    } else {
+                        throw (OIDCException) throwable;
                     }
-
                 } else {
-                    throw ex;
+                    throw new OIDCException(throwable);
                 }
             }
         }


extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/OidcRecorder.java
```diff
@@ -119,16 +120,22 @@ public void handle(AsyncResult<OAuth2Auth> event) {
...
-                    } catch (InterruptedException iex) {
-                        // continue connecting
+            } catch (Throwable throwable) {
+                while (throwable instanceof CompletionException && throwable.getCause() != null) {
+                    throwable = throwable.getCause();
+                }
+                if (throwable instanceof OIDCException) {
+                    if (i + 1 < connectionRetryCount) {
+                        try {
+                            Thread.sleep(2000);
```
stuartwdouglas: Can this be called in an IO thread?

sberyozkin: Hi Stuart @stuartwdouglas Good question, I don't know, do you mean avoid using the future at all ? It just joins till the OIDC server has been contacted, may be it is not needed. This is done at the application start up, no client requests are involved.

sberyozkin: @stuartwdouglas `runAsync` I guess, let me try, though I'd not've guessed it implies a more optimal thread consumption in this case :-)

sberyozkin: @stuartwdouglas Should be better now

stuartwdouglas: what I was actually getting at was should we change this to register a timer and callback, rather than blocking a thread for 2s?

extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/OidcRecorder.java
```diff
@@ -119,16 +120,29 @@ public void handle(AsyncResult<OAuth2Auth> event) {
...
-                        Thread.sleep(2000);
-                    } catch (InterruptedException iex) {
-                        // continue connecting
+            } catch (Throwable throwable) {
+                while (throwable instanceof CompletionException && throwable.getCause() != null) {
+                    throwable = throwable.getCause();
+                }
+                if (throwable instanceof OIDCException) {
+                    if (i + 1 < connectionRetryCount) {
+                        CompletableFuture.runAsync(
```
stuartwdouglas: runAsync should never be used, it uses the default fork join pool that we have no control over.

extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/OidcRecorder.java
```diff
@@ -119,16 +120,29 @@ public void handle(AsyncResult<OAuth2Auth> event) {
...
+                                    @Override
+                                    public void run() {
+                                        try {
+                                            Thread.sleep(2000);
+                                        } catch (InterruptedException iex) {
+                                            // continue connecting
+                                        }
+                                    }
+                                })
+                                .join();
```
stuartwdouglas: Also this blocks the thread anyway, so now you are blocking 2 threads instead of just one.

extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/OidcRecorder.java
sberyozkin: @stuartwdouglas Callback to what condition though ? Vertx tells us OIDC is not contactable. It won't let us save 1 sec out of those 2 secs by calling back when OIDC becomes contactable unless this timer you are proposing will be pinging OIDC every 100 milliseconds. Is it what you suggest be done here ? I'm not sure it is really worth it, we are not dealing with the actual client request here. May be just reducing a wait period from 2 secs to 200 milliseconds is a simpler version of the timer/callback approach ?

extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/OidcRecorder.java
sberyozkin: Right, recall you were saying in context of the CL work...

extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/OidcRecorder.java
sberyozkin: Well it can be concluded I did not draw the right conclusion based on the `io thread` proposal :-)

sberyozkin:
@stuartwdouglas Hi Stuart, I've spent some time today exploring a Timer idea, and I have to admit I don't see how it can work better than it was originally.

We can start a timer and make `Consumer<OAuth2Auth>` visible to the timer task, but connecting it with the rest of `OidcRecorder` does not work for me in a better way, as there still has to be a class linking that callback with the thread which sets `TenantConfigContext` in the map, with this thread inevitably being blocked anyway till that callback fires.

My plan is to revert my current `runAsync` blunder and be done with it, but I'd happy to learn something new :-), so may be you can prototype some code here ?

Cheers

extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/OidcRecorder.java
stuartwdouglas: I misread this, I thought it was happening at runtime. It's not ideal that it connects to each tenant sequentially but it is probably fine for now, sorry for the noise.

sberyozkin: @stuartwdouglas Hi Stuart, no problems at all, it has been awhile since I had to think so hard, trying to figure out how to link all those Functions and Consumers :-).

Good point about the runtime, in fact we have 2 cases here, one is when the static TenantContextConfigs are created, those in the application.properties and a new case when the multi-tenancy resolvers may need it, during the client request, and I was thinking may be we should not even support the `connection-delay` property in this latter case, if someone sets to even 10 sec it won't be good. This property has been added to help in the cases when at the initial start up, the OIDC server is a few secs behind, to help with the tests mainly...I'll create a small issue afterwards, so that we can decide if to ignore this property in those cases or if not then at least make the more optimal use of the current io thread as per proposal, etc

sberyozkin:
Hi @stuartwdouglas This PR is now back to the original fix state, where we just correctly get to  `OIDCException` from `CompletionException` :-).
Also opened #7512 to do a possible follow up update to do with this property
Thanks

stuartwdouglas:
This still has the runAsync

sberyozkin:
@stuartwdouglas silly me, forgot to force push yesterday :-)

stuartwdouglas (approved):

PR #7452, opened 2020-02-27
Make updating a built application to k8s a no-brainer

geoand:


iocanel (approved):
LGTM

maxandersen (changes requested):
idea is good as from developer perspective its annoying not getting latest changes on a push.

I'm aok on the changes, but I reckon we should document how to turn this off in a prod profile ?

maxandersen (approved):
lets get this in and update docs as we get them more complete

geoand:
I'll be looking at the docs tomorrow and then next week and plan to add some "how to" style additions to it, one of which would address the above comment.

PR #7470, opened 2020-02-27
feat: don't display registry warning for s2i

iocanel:
Currently we display a warning about docker registry being a requirement for `kubernetes ` deployment. 

This is not the case for s2i and we should not display such warning for s2i builds.


geoand (approved):

geoand:
CI failure is unrelated, merging

Issue #7472, opened 2020-02-27
Qute can not find included template without extension


fishb6nes:
**Describe the bug**
Contrary to the behaviour outlined in https://quarkus.io/guides/qute-reference#include_helper
attempting to include a template without specifying its extension results in an `IllegalStateException: Template not found`

**Expected behavior**
For the following template to correctly include the base template:
```
{#include base} 
  {#title}My Title{/title} 
  {#body}<div>My body.</div>{/body}
{/include}
```

**Actual behavior**
An `IllegalStateException: Template not found` is thrown.

To correctly include the base template its extension needs to be specified:
```
{#include base.html} 
  {#title}My Title{/title} 
  {#body}<div>My body.</div>{/body}
{/include}
```

**Environment (please complete the following information):**
 - Output of `uname -a` or `ver`: Microsoft Windows [Version 10.0.18362.657]
 - Output of `java -version`: 13.0.2
 - GraalVM version (if different from Java): Not installed
 - Quarkus version or git rev: 1.2.1

**Additional context**
If this is a limitation due to the fact that one could also include a plain text template maybe the extension of the template including the base template could be used as a sensible default?

mkouba:
@fishb6nes Hm, I've added a quarkus-specific test here: https://github.com/quarkusio/quarkus/pull/7476 and it seems to work fine.

The truth is that we don't reflect the variant/extension of the template including the base, ie. we just try the suffixes from the `QuteConfig.suffixes` config list.

fishb6nes:
Ah I see what the issue is now. I thought of the included template as a template, not a tag. If the included template is in `templates/tags/` it does indeed work correctly, however if it is a sibiling of the including template in `templates/` it throws the not found exception. I'm assuming it works with the extension as specifying one overrides the default resolution from `templates/tags/` and in stead considers it a relative path from `tempates/`?

If we want to leave it as is for now I've opened https://github.com/quarkusio/quarkus/pull/7487 to clarify. But automatically resolving from a `templates/includes/` folder in stead of `templates/tags/` might better convey the purpose of the templates within both folders.

mkouba:
> If the included template is in templates/tags/ it does indeed work correctly

[...]

PR #7476, opened 2020-02-28
Qute - add test for include section, a small fix for user tags

mkouba:


mkouba:
This PR resolves #7472.

gsmet (approved):

PR #7498, opened 2020-02-29
Add logging of hibernate orm bind params

### LINKED ISSUE #3075: [ORM] Add option for logging bind variables
**Description**
Instead of only seeing the executed DML, I'd also like to see the values for bind parameters (akin to org.hibernate.type=trace in old times).

Postremus:
"quarkus.hibernate-orm.log.sql" now supports an extended option, which enables logging of bind params
closes #3075

### DIFF
--- a/extensions/hibernate-orm/deployment/src/main/java/io/quarkus/hibernate/orm/deployment/HibernateOrmConfig.java
+++ b/extensions/hibernate-orm/deployment/src/main/java/io/quarkus/hibernate/orm/deployment/HibernateOrmConfig.java
@@ -285,13 +285,21 @@ public class HibernateOrmConfig {
         public boolean sql;
 
         /**
+         * Logs SQL bind parameter.
+         * <p>
+         * Setting it to true is obviously not recommended in production.
+         */
+        @ConfigItem(defaultValue = "false")
+        public boolean bindParam;
+
+        /**
          * Whether JDBC warnings should be collected and logged.
          */
         @ConfigItem(defaultValueDocumentation = "depends on dialect")
         public Optional<Boolean> jdbcWarnings;
 
         public boolean isAnyPropertySet() {
-            return sql || jdbcWarnings.isPresent();
+            return sql || bindParam || jdbcWarnings.isPresent();
         }
     }
 
--- a/extensions/hibernate-orm/deployment/src/main/java/io/quarkus/hibernate/orm/deployment/HibernateOrmProcessor.java
+++ b/extensions/hibernate-orm/deployment/src/main/java/io/quarkus/hibernate/orm/deployment/HibernateOrmProcessor.java
@@ -50,6 +50,7 @@ import org.jboss.jandex.CompositeIndex;
 import org.jboss.jandex.DotName;
 import org.jboss.jandex.IndexView;
 import org.jboss.jandex.Indexer;
+import org.jboss.logmanager.Level;
 
 import io.quarkus.agroal.deployment.JdbcDataSourceBuildItem;
 import io.quarkus.agroal.deployment.JdbcDataSourceSchemaReadyBuildItem;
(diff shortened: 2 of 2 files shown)


extensions/hibernate-orm/deployment/src/main/java/io/quarkus/hibernate/orm/deployment/HibernateOrmConfig.java
```diff
@@ -279,10 +279,19 @@ public boolean isAnyPropertySet() {
...
          * Show SQL logs and format them nicely.
          * <p>
+         * true logs sql statements
+         * extended logs sql statements and bind parameter values
+         * <p>
          * Setting it to true is obviously not recommended in production.
          */
-        @ConfigItem(defaultValue = "false")
-        public boolean sql;
+        @ConfigItem(defaultValueDocumentation = "false")
```
machi1990: Shouldn't this be `defaultValue=false`?

Postremus: `defaultValue=false` would also work here, but would require a more in depth check in the `isAnyPropertySet()` method (line 303)

extensions/hibernate-orm/deployment/src/main/java/io/quarkus/hibernate/orm/deployment/HibernateOrmProcessor.java
```diff
@@ -642,9 +644,14 @@ private void handleHibernateORMWithNoPersistenceXml(
                                 fetchSize.toString()));
 
                 // Logging
-                if (hibernateConfig.log.sql) {
-                    desc.getProperties().setProperty(AvailableSettings.SHOW_SQL, "true");
-                    desc.getProperties().setProperty(AvailableSettings.FORMAT_SQL, "true");
+                if (hibernateConfig.log.sql.isPresent()) {
+                    HibernateOrmConfig.HibernateOrmConfigLog.HibernateOrmConfigLogSqlValue value = hibernateConfig.log.sql
+                            .get();
+                    if (value.ordinal() >= HibernateOrmConfig.HibernateOrmConfigLog.HibernateOrmConfigLogSqlValue.TRUE
```
gunnarmorling: Please don't rely on the ordinal, it's prone to errors in case someone re-orders the enum constants down the road. I think simply checking for the two enum constants is good enough; or, if you'd want something more generic, add methods to the enum such as `boolean logsSql()` (returning true for TRUE and EXTENDED) and `boolean logsParameters()`.

extensions/hibernate-orm/deployment/src/main/java/io/quarkus/hibernate/orm/deployment/HibernateOrmConfig.java
```diff
@@ -279,10 +279,19 @@ public boolean isAnyPropertySet() {
...
+        public Optional<HibernateOrmConfigLogSqlValue> sql;
```
gunnarmorling: Why is it optional, if there is a default value?

gunnarmorling (changes requested):
Two comments inline. Thanks!

extensions/hibernate-orm/deployment/src/main/java/io/quarkus/hibernate/orm/deployment/HibernateOrmConfig.java
dmlloyd: I wouldn't do it this way. It would be better to create a separate parameter for this than to have properties which are boolean plus some other thing.

extensions/hibernate-orm/deployment/src/main/java/io/quarkus/hibernate/orm/deployment/HibernateOrmConfig.java
```diff
@@ -279,10 +279,19 @@ public boolean isAnyPropertySet() {
...
+
+        public enum HibernateOrmConfigLogSqlValue {
+            FALSE,
+            TRUE,
+            EXTENDED;
```
dmlloyd: Yeah definitely not.

dmlloyd (changes requested):

extensions/hibernate-orm/deployment/src/main/java/io/quarkus/hibernate/orm/deployment/HibernateOrmConfig.java
Postremus: Done

Postremus:
@dmlloyd @gunnarmorling
I addressed every point of the first round of reviews.
Onto the next round :D

dmlloyd (approved):

Postremus:
@gunnarmorling ping

gunnarmorling (approved):
LGTM, but I'm lacking permission to merge. @gsmet?

Sanne (approved):

Sanne:
thanks all, very useful!

PR #7504, opened 2020-03-02
Fix broken Hibernate Search Quickstart

geoand:
Example of failure can be see by following:
https://github.com/quarkusio/quarkus/issues/6588#issuecomment-593250181

Essentially the problem was that disabling ssl was causing a:
```
Caused by: java.util.NoSuchElementException: No value present
```

### DIFF
--- a/extensions/agroal/runtime/src/main/java/io/quarkus/agroal/runtime/AbstractDataSourceProducer.java
+++ b/extensions/agroal/runtime/src/main/java/io/quarkus/agroal/runtime/AbstractDataSourceProducer.java
@@ -143,8 +143,12 @@ public abstract class AbstractDataSourceProducer {
                 agroalConnectionConfigurerHandle.get().disableSslSupport(dataSourceBuildTimeConfig.dbKind.get(),
                         dataSourceConfiguration);
             } else {
-                log.warnv("Agroal does not support disabling SSL for database kind {0}",
-                        dataSourceBuildTimeConfig.dbKind.get());
+                if (dataSourceBuildTimeConfig.dbKind.isPresent()) {
+                    log.warnv("Agroal does not support disabling SSL for database kind {0}",
+                            dataSourceBuildTimeConfig.dbKind.get());
+                } else {
+                    log.warn("Agroal does not support disabling SSL");
+                }
             }
         }
 


extensions/agroal/runtime/src/main/java/io/quarkus/agroal/runtime/AbstractDataSourceProducer.java
```diff
@@ -143,8 +143,12 @@ public AgroalDataSource createDataSource(String dataSourceName,
                 agroalConnectionConfigurerHandle.get().disableSslSupport(dataSourceBuildTimeConfig.dbKind.get(),
                         dataSourceConfiguration);
             } else {
-                log.warnv("Agroal does not support disabling SSL for database kind {0}",
-                        dataSourceBuildTimeConfig.dbKind.get());
```
gsmet: That's not the right fix. You should just use `resolvedDbKind` here instead of the one from the config.

gsmet (changes requested):

extensions/agroal/runtime/src/main/java/io/quarkus/agroal/runtime/AbstractDataSourceProducer.java
geoand: I wasn't aware of that, thanks

gsmet: We have this layer for compatibility reason as the kind might not come from the config.

extensions/agroal/runtime/src/main/java/io/quarkus/agroal/runtime/AbstractDataSourceProducer.java
```diff
@@ -143,8 +143,12 @@ public AgroalDataSource createDataSource(String dataSourceName,
                 agroalConnectionConfigurerHandle.get().disableSslSupport(dataSourceBuildTimeConfig.dbKind.get(),
```
gsmet: Here we should use `resolvedDbKind` too.

gsmet:
Let's close this one for now. I'll have a look tonight as I want to check and tune a few things.

geoand:
Ok makes sense

PR #7508, opened 2020-03-02
Ensure that Amazon Lambda uses the ObjectMapper bean when available

geoand:
This was the result of me looking into: https://github.com/quarkusio/quarkus/issues/6041#issuecomment-592657931

### DIFF
--- a/extensions/amazon-lambda/runtime/src/main/java/io/quarkus/amazon/lambda/runtime/AmazonLambdaRecorder.java
+++ b/extensions/amazon-lambda/runtime/src/main/java/io/quarkus/amazon/lambda/runtime/AmazonLambdaRecorder.java
@@ -22,6 +22,8 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.ObjectReader;
 import com.fasterxml.jackson.databind.ObjectWriter;
 
+import io.quarkus.arc.Arc;
+import io.quarkus.arc.InstanceHandle;
 import io.quarkus.arc.runtime.BeanContainer;
 import io.quarkus.runtime.Application;
 import io.quarkus.runtime.ShutdownContext;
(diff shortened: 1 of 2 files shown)


extensions/amazon-lambda/runtime/src/main/java/io/quarkus/amazon/lambda/runtime/AmazonLambdaRecorder.java
```diff
@@ -45,14 +47,22 @@
...
                 .configure(MapperFeature.ACCEPT_CASE_INSENSITIVE_PROPERTIES, true);
         Method handlerMethod = discoverHandlerMethod(handlerClass);
         objectReader = objectMapper.readerFor(handlerMethod.getParameterTypes()[0]);
         objectWriter = objectMapper.writerFor(handlerMethod.getReturnType());
     }
 
+    private ObjectMapper getObjectMapper() {
+        InstanceHandle<ObjectMapper> instance = Arc.container().instance(ObjectMapper.class);
+        if (instance.isAvailable()) {
+            return instance.get().copy();
```
gsmet: Why `copy()`?

geoand: Because the recorder then adds its own settings which we don't want to propagate to the configured bean since that would affect all consumers

oztimpower:
Validated as working on my build, post rolling in the changes.  Nice one!

geoand:
Excellent, thanks for checking!

Do you mind adding a comment to the PR as well?

gsmet (changes requested):
I'm a bit torn about that one. I wonder if it's such a good idea to reuse the object mapper if it's one used for a very specific context and that needs a very specific configuration.

Can we discuss it? (Marking it as Request changes so that we have a discussion before merging)

geoand:
Sure, let's talk tomorrow

stuartwdouglas:
Maybe this should have a specific Qualifier so you have a bean that is just the object mapper for Lambda.

geoand:
The idea of using the bean from Arc is that it's been properly configured with various Jackson modules automatically by Quarkus.
So in the case that @oztimpower had, the Kotlin Jackson module is configured by Quarkus making Lambda work OOTB with Kotlin.

@gsmet @stuartwdouglas The alternative would be to create a second ObjectMapper for Lamdba under the covers that applies the same modules but also applies the few extra settings that Lambda needs. Like Stuart said this would have a qualifier and we would use that qualifier when pulling the bean out of Arc in the Lambda Recorder code.
How does that sound?

stuartwdouglas:
Actually I think the current implementation is probably fine. The would be the main use case for using the mapper with AWS lambda anyway, so any customisers the users provide would be expected to be used for this. I don't think we need to complicate it any further.

stuartwdouglas (approved):

geoand:
@gsmet are you OK with this?

gsmet (approved):

gsmet:
Yes, let's move on then.

PR #7510, opened 2020-03-02
feat: Hibernate with Panache ranged query

### LINKED ISSUE #3870: PanacheQuery.range()
**Description**
PanacheQuery support a paging mechanism that allow to fetch the result of a query page by page. This is handy to list all the result of a query but not to give an external application the capability to fetch all the results. 

Moreover, paging is statefull as it implies to store the size of the page and the current page index.

In API developement, range query is often implemented by providing an extra query parameter `?range=0-10`, this is suitable to be used with HTTP range headers defined in the follwing RFC; https://tools.ietf.org/html/rfc7233

Building such range capability where a user can ask for any range within the list of result is not suitable with the current paging implementation.

So I propose to add a PanacheQuery.range() method for this.

**Implementation ideas**
Add a `PanacheQuery.range(startIdx, endIdx)` method that will allow to set `offset` and `limit` on the generated SQL query as the current paging mechanism do.

If it's OK, I can works on a PR for this.

loicmathieu:
This is a ~draft~ implementation of Range queries, that fixes #3870 

~It introduce a `Findable` interface that is returned by the `range()` that only contains query result related methods.~

/cc @emmanuelbernard @FroMage if you validate the design I'll implement it for MongoDB also.

### DIFF
--- a/docs/src/main/asciidoc/hibernate-orm-panache.adoc
+++ b/docs/src/main/asciidoc/hibernate-orm-panache.adoc
@@ -440,6 +440,31 @@ return Person.find("status", Status.Alive)
 
 The `PanacheQuery` type has many other methods to deal with paging and returning streams.
 
+=== Using a range instead of pages
+
+`PanacheQuery` also allows range-based queries.
+
+[source,java]
+----
+// create a query for all living persons
+PanacheQuery<Person> livingPersons = Person.find("status", Status.Alive);
+
+// make it use a range: start at index 0 until index 24 (inclusive).
+livingPersons.range(0, 24);
+
+// get the range
+List<Person> firstRange = livingPersons.list();
+
+// to get the next range, you need to call range again
+List<Person> secondRange = livingPersons.range(25, 49).list();
+----
+
+[WARNING]
+----
+You cannot mix ranges and pages: if you use a range, all methods that depend on having a current page will throw an `UnsupportedOperationException`;
+you can switch back to paging using `page(Page)` or `page(int, int)`.
+----
+
 === Sorting
 
 All methods accepting a query string also accept the following simplified query form:
--- a/docs/src/main/asciidoc/mongodb-panache.adoc
+++ b/docs/src/main/asciidoc/mongodb-panache.adoc
@@ -441,6 +441,31 @@ return Person.find("status", Status.Alive)
 
 The `PanacheQuery` type has many other methods to deal with paging and returning streams.
 
+=== Using a range instead of pages
+
+`PanacheQuery` also allows range-based queries.
+
+[source,java]
+----
+// create a query for all living persons
+PanacheQuery<Person> livingPersons = Person.find("status", Status.Alive);
+
+// make it use a range: start at index 0 until index 24 (inclusive).
+livingPersons.range(0, 24);
+
+// get the range
+List<Person> firstRange = livingPersons.list();
+
+// to get the next range, you need to call range again
+List<Person> secondRange = livingPersons.range(25, 49).list();
+----
+
+[WARNING]
+----
+You cannot mix ranges and pages: if you use a range, all methods that depend on having a current page will throw an `UnsupportedOperationException`;
+you can switch back to paging using `page(Page)` or `page(int, int)`.
+----
+
 === Sorting
 
 All methods accepting a query string also accept an optional `Sort` parameter, which allows you to abstract your sorting:
--- a/extensions/panache/hibernate-orm-panache/runtime/src/main/java/io/quarkus/hibernate/orm/panache/PanacheQuery.java
+++ b/extensions/panache/hibernate-orm-panache/runtime/src/main/java/io/quarkus/hibernate/orm/panache/PanacheQuery.java
@@ -117,6 +117,16 @@ public interface PanacheQuery<Entity> {
     public Page page();
 
     /**
+     * Switch the query to use a fixed range (start index - last index) instead of a page.
+     * As the range is fixed, subsequent pagination of the query is not possible.
+     *
+     * @param startIndex the index of the first element, starting at 0
+     * @param lastIndex the index of the last element
+     * @return this query, modified
+     */
+    public <T extends Entity> PanacheQuery<T> range(int startIndex, int lastIndex);
+
+    /**
      * Define the locking strategy used for this query.
      *
      * @param lockModeType the locking strategy to be used for this query.
--- a/extensions/panache/mongodb-panache/runtime/src/main/java/io/quarkus/mongodb/panache/PanacheQuery.java
+++ b/extensions/panache/mongodb-panache/runtime/src/main/java/io/quarkus/mongodb/panache/PanacheQuery.java
@@ -116,6 +116,16 @@ public interface PanacheQuery<Entity> {
      */
     public Page page();
 
+    /**
+     * Switch the query to use a fixed range (start index - last index) instead of a page.
+     * As the range is fixed, subsequent pagination of the query is not possible.
+     *
+     * @param startIndex the index of the first element, starting at 0
+     * @param lastIndex the index of the last element
+     * @return this query, modified
+     */
+    public <T extends Entity> PanacheQuery<T> range(int startIndex, int lastIndex);
+
     // Results
 
     /**
--- a/extensions/panache/mongodb-panache/runtime/src/main/java/io/quarkus/mongodb/panache/reactive/ReactivePanacheQuery.java
+++ b/extensions/panache/mongodb-panache/runtime/src/main/java/io/quarkus/mongodb/panache/reactive/ReactivePanacheQuery.java
@@ -110,6 +110,16 @@ public interface ReactivePanacheQuery<Entity> {
      */
     public Page page();
 
+    /**
+     * Switch the query to use a fixed range (start index - last index) instead of a page.
+     * As the range is fixed, subsequent pagination of the query is not possible.
+     *
+     * @param startIndex the index of the first element, starting at 0
+     * @param lastIndex the index of the last element
+     * @return this query, modified
+     */
+    public <T extends Entity> ReactivePanacheQuery<T> range(int startIndex, int lastIndex);
+
     // Results
 
     /**
--- a/extensions/panache/panache-common/runtime/src/main/java/io/quarkus/panache/common/Range.java
+++ b/extensions/panache/panache-common/runtime/src/main/java/io/quarkus/panache/common/Range.java
@@ -1,0 +1,36 @@
+package io.quarkus.panache.common;
+
+/**
+ * <p>
+ * Utility class to represent ranging information. Range instances are immutable.
+ * </p>
+ *
+ * <p>
+ * Usage:
+ * </p>
+ *
+ * <code><pre>
+ * Range range = Range.of(0, 5);
+ * </pre></code>
+ */
+public class Range {
+    private final int startIndex;
+    private final int lastIndex;
+
+    public Range(int startIndex, int lastIndex) {
+        this.startIndex = startIndex;
+        this.lastIndex = lastIndex;
+    }
+
+    public static Range of(int startIndex, int lastIndex) {
+        return new Range(startIndex, lastIndex);
+    }
+
+    public int getStartIndex() {
+        return startIndex;
+    }
+
+    public int getLastIndex() {
+        return lastIndex;
+    }
+}
--- a/integration-tests/hibernate-orm-panache/src/main/java/io/quarkus/it/panache/TestEndpoint.java
+++ b/integration-tests/hibernate-orm-panache/src/main/java/io/quarkus/it/panache/TestEndpoint.java
@@ -200,6 +200,10 @@ public class TestEndpoint {
         testPaging(Person.findAll());
         testPaging(Person.find("ORDER BY name"));
 
+        // range
+        testRange(Person.findAll());
+        testRange(Person.find("ORDER BY name"));
+
         try {
             Person.findAll().singleResult();
             Assertions.fail("singleResult should have thrown");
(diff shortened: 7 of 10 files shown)


FroMage (changes requested):
I don't think `Findable` buys us anything here: it only restricts what you can do with the returned value, but the returned value is `this`, so it's still modified.

It prevents you from doing this:

```java
PanacheQuery<String> x = ...;
x.range(1, 2).page(2) // error
```

But it lets you do this:

```java
PanacheQuery<String> x = ...;
x.page(2).range(1, 2) // no error
```

and even worse:

```java
PanacheQuery<String> x = ...;
x.range(1, 2);
x.page(2) // no error
```

So I'd get rid of the `Findable` and make it so:

- You can set a page, or a range, and that overrides whatever previous page or range.
- Setting a range overrides the previous page
- Setting a page overrides the previous range
- Querying anything that requires a range when a page is set results in an exception
- Querying anything that requires a page when a range is set results in an exception (all operations that give you the page count, etc…)

Effectively, your query can then alternate between page mode and range mode.

WDYT?

loicmathieu:
@Fromage I was not sure whether to provides restriction via interface or Exception. Interface restriction add more restrictions but as you said, it's still not coherent.

I will propose a restriction via Exception, it will be a little harder to reason about at the implementation level but maybe better to use it (and the exception should occur at development time so it should be as safe to use as interface based restriction). But as page is statefull and range is stateless I think it should be OK to go from page to range but not the opposite.

FroMage:
> But as page is statefull and range is stateless

They're both stateful: they change the current window. I don't see the difference?

loicmathieu:
> They're both stateful: they change the current window. I don't see the difference?

The issue is that when you iterate over the pages it's statefull but not when you iterate over pages :

```java
PanacheQuery<Person> pagedQuery = Person.findAll().page(5);
pagedQuery.nextPage().list();// use the page state
pagedQuery.nextPage().list();// use the page state

PanacheQuery<Person> rangedQuery = Person.findAll();
rangedQuery.range(0, 5).list();// stateless the range state is on the call site
rangedQuery.range(5, 10).list();// stateless the range state is on the call site
```
According to me, a range after a page should be prohibitted (is the range for the full query or for the page ?).
As range is kind of stateless (but OK, not on the implementation side), calling page() after a range should be allowed, it will reset the range and start using a page.

Maybe we should clear the range after querying to make it really stateless but this can also be surprising:

```
PanacheQuery<Person> rangedQuery = Person.findAll();
rangedQuery.range(0, 5).list();// list of 5 persons
rangedQuery.list();// list of all persons ? or the same 5 persons ?
```

loicmathieu:
@FroMage can you review the new implementation ? When we agree on it I'll update MongoDB accordingly.

FroMage:
> According to me, a range after a page should be prohibitted (is the range for the full query or for the page ?).
> As range is kind of stateless (but OK, not on the implementation side), calling page() after a range should be allowed, it will reset the range and start using a page.

Right, ATM both are stateful.

> Maybe we should clear the range after querying to make it really stateless but this can also be surprising:
>
> ```
> PanacheQuery<Person> rangedQuery = Person.findAll();
> rangedQuery.range(0, 5).list();// list of 5 persons
> rangedQuery.list();// list of all persons ? or the same 5 persons ?
> ```

Nah, I'd leave both `range` and `page` to be stateful and keep whatever "mode" they had last (page or range). It should be fine to switch from one mode to another, but if you call `nextPage` when you're in "range" mode it should throw. I think that's the less confusing/surprising behaviour.

Most people will stick to one of the modes anyway.

integration-tests/hibernate-orm-panache/src/main/java/io/quarkus/it/panache/TestEndpoint.java
```diff
@@ -851,6 +859,47 @@ private void testPaging(PanacheQuery<Person> query) {
...
+        Assertions.assertEquals(1, persons.size());
+        Assertions.assertEquals("stef6", persons.get(0).name);
+
+        persons = query.range(8, 12).list();
+        Assertions.assertEquals(0, persons.size());
+
+        // mix range with page
+        try {
+            // invalid
+            query.range(0, 2).nextPage();
```
FroMage: Let's be thorough and test the rest of the methods that should throw.

FroMage (changes requested):
LGTM, except for a minor test issue.

integration-tests/hibernate-orm-panache/src/main/java/io/quarkus/it/panache/TestEndpoint.java
loicmathieu: I added more test, and discover a missing check so worth it ;)

loicmathieu:
@FroMage I add more test and implement range for MongoDB.

MongoDB lack some tests (no tests on pagination neither), I openned an issue to add them later #7698

docs/src/main/asciidoc/hibernate-orm-panache.adoc
```diff
@@ -440,6 +440,31 @@ return Person.find("status", Status.Alive)
...
+=== Using a range instead of pages
+
+`PanacheQuery` also allow range based query instead of page based.
+
+[source,java]
+----
+// create a query for all living persons
+PanacheQuery<Person> livingPersons = Person.find("status", Status.Alive);
+
+// make it use a range or start index 0 and last index 25.
```
FroMage:
```suggestion
// make it use a range: start at index 0 until index 25 (inclusive).
```

docs/src/main/asciidoc/hibernate-orm-panache.adoc
```diff
@@ -440,6 +440,31 @@ return Person.find("status", Status.Alive)
...
+livingPersons.range(0, 25);
+
+// get the range
+List<Person> firstRange = livingPersons.list();
+
+// to get the next range, you need to call range again
```
FroMage: Oh hold on, I thought the range was inclusive on the last index? If it is, then you get element 25 for the previous range and this one too, no?

docs/src/main/asciidoc/hibernate-orm-panache.adoc
```diff
@@ -440,6 +440,31 @@ return Person.find("status", Status.Alive)
...
+List<Person> secondRange = livingPersons.range(25, 50).list();
+----
+
+[WARNING]
+----
+You cannot mix range and page, if you use a range, all page related methods will throw an `UnsupportedOperationException`;
```
FroMage:
```suggestion
You cannot mix ranges and pages: if you use a range, all methods that depend on having a current page will throw an `UnsupportedOperationException`;
```

docs/src/main/asciidoc/hibernate-orm-panache.adoc
```diff
@@ -440,6 +440,31 @@ return Person.find("status", Status.Alive)
...
+you must use a method that defines a new page before using them : `page(Page)` or `page(int, int)`.
```
FroMage:
```suggestion
you can switch back to paging using `page(Page)` or `page(int, int)`.
```

docs/src/main/asciidoc/mongodb-panache.adoc
```diff
@@ -437,6 +437,31 @@ return Person.find("status", Status.Alive)
 
 The `PanacheQuery` type has many other methods to deal with paging and returning streams.
 
+=== Using a range instead of pages
+
+`PanacheQuery` also allow range based query instead of page based.
```
FroMage:
```suggestion
`PanacheQuery` also allows range-based queries.
```

docs/src/main/asciidoc/mongodb-panache.adoc
```diff
@@ -437,6 +437,31 @@ return Person.find("status", Status.Alive)
...
+
+[source,java]
+----
+// create a query for all living persons
+PanacheQuery<Person> livingPersons = Person.find("status", Status.Alive);
+
+// make it use a range or start index 0 and last index 25.
```
FroMage:
```suggestion
// make it use a range: start at index 0 until index 25 (inclusive).
```

docs/src/main/asciidoc/mongodb-panache.adoc
```diff
@@ -437,6 +437,31 @@ return Person.find("status", Status.Alive)
...
+// get the range
+List<Person> firstRange = livingPersons.list();
+
+// to get the next range, you need to call range again
+List<Person> secondRange = livingPersons.range(25, 50).list();
+----
+
+[WARNING]
+----
+You cannot mix range and page, if you use a range, all page related methods will throw an `UnsupportedOperationException`;
```
FroMage:
```suggestion
You cannot mix ranges and pages: if you use a range, all methods that depend on having a current page will throw an `UnsupportedOperationException`;
```

docs/src/main/asciidoc/mongodb-panache.adoc
```diff
@@ -437,6 +437,31 @@ return Person.find("status", Status.Alive)
...
+you must use a method that defines a new page before using them : `page(Page)` or `page(int, int)`.
```
FroMage:
```suggestion
you can switch back to paging using `page(Page)` or `page(int, int)`.
```

extensions/panache/hibernate-orm-panache/runtime/src/main/java/io/quarkus/hibernate/orm/panache/PanacheQuery.java
```diff
@@ -116,6 +116,16 @@
...
 
+    /**
+     * Switch the query to use a fixed range (start index - last index) instead of a page.
+     * As the range is fixed, subsequent pagination of the query is not possible.
+     *
+     * @param startIdx the index of the first element, starting at 0
+     * @param lastIdx the index of the last element
+     * @return this query, modified
+     */
+    public <T extends Entity> PanacheQuery<T> range(int startIdx, int lastIdx);
```
FroMage:
```suggestion
    public <T extends Entity> PanacheQuery<T> range(int startIndex, int lastIndex);
```

extensions/panache/mongodb-panache/runtime/src/main/java/io/quarkus/mongodb/panache/PanacheQuery.java
```diff
@@ -116,6 +116,16 @@
...
 
+    /**
+     * Switch the query to use a fixed range (start index - last index) instead of a page.
+     * As the range is fixed, subsequent pagination of the query is not possible.
+     *
+     * @param startIdx the index of the first element, starting at 0
+     * @param lastIdx the index of the last element
+     * @return this query, modified
+     */
+    public <T extends Entity> PanacheQuery<T> range(int startIdx, int lastIdx);
```
FroMage:
```suggestion
    public <T extends Entity> PanacheQuery<T> range(int startIndex, int lastIndex);
```

extensions/panache/mongodb-panache/runtime/src/main/java/io/quarkus/mongodb/panache/reactive/ReactivePanacheQuery.java
```diff
@@ -110,6 +110,16 @@
...
 
+    /**
+     * Switch the query to use a fixed range (start index - last index) instead of a page.
+     * As the range is fixed, subsequent pagination of the query is not possible.
+     *
+     * @param startIdx the index of the first element, starting at 0
+     * @param lastIdx the index of the last element
+     * @return this query, modified
+     */
+    public <T extends Entity> ReactivePanacheQuery<T> range(int startIdx, int lastIdx);
```
FroMage:
```suggestion
    public <T extends Entity> ReactivePanacheQuery<T> range(int startIndex, int lastIndex);
```

extensions/panache/panache-common/runtime/src/main/java/io/quarkus/panache/common/Range.java
```diff
@@ -0,0 +1,36 @@
...
+ *
+ * <code><pre>
+ * Range range = Range.of(0, 5);
+ * </pre></code>
+ */
+public class Range {
+    private final int startIdx;
+    private final int lastIdx;
+
+    public Range(int startIdx, int lastIdx) {
```
FroMage: Please `index` everywhere here :)

FroMage (changes requested):
Sorry it seems I overlooked some documentation and parameter names in my last review. Also I want to make sure I understand if the `lastIndex` of a range is inclusive, because your docs example appears to say it's not, or it requests it twice.

docs/src/main/asciidoc/hibernate-orm-panache.adoc
loicmathieu: Yes, the example should be `range(0, 24)` then `range(25, 49)` to have two range of 25 items as range starts at 0.
I'll update the documentation.

We can decide to make it start at 1 if needed, but in IT stuff starts at 0 ;)

loicmathieu:
@FroMage I took into account all your remarks, a CI check is failing in an unrelated part ...

loicmathieu:
@FroMage is it OK for you ?

docs/src/main/asciidoc/hibernate-orm-panache.adoc
FroMage: OK fine.

FroMage (approved):
LGTM.

FroMage:
I'm good, but the CI must pass…

loicmathieu:
@FroMage thanks, I will rebase and force push to run the CI again, it failed on an urelated test ...

loicmathieu:
@FroMage CI passed, can you merge it ?

FroMage:
Thanks!

PR #7514, opened 2020-03-02
Show better error when multiple container image extensions used with Kubernetes

### LINKED ISSUE #7484: Conflicts when multiple container-image modules are present
If we have more than one container-image modules present, for example `container-image-s2i and `container-image-jib` the following error occurs:

```
[ERROR] Failed to execute goal io.quarkus:quarkus-maven-plugin:999-SNAPSHOT:build (default) on project kubernetes-quickstart: 
Failed to build quarkus application: io.quarkus.builder.ChainBuildException: 
Multiple producers of item class io.quarkus.container.spi.ContainerImageResultBuildItem 
(io.quarkus.container.image.s2i.deployment.S2iProcessor#s2iBuildFromJar):
This is the location of the conflicting producer 
(io.quarkus.container.image.jib.deployment.JibProcessor#buildFromJar) -> [Help 1]
```

It would be nice if we could provide a better message.
Also it would be nice if we could enable, disable or explicitly select one of the two using configuration.

geoand:
Fixes: #7484

### DIFF
--- a/extensions/container-image/spi/src/main/java/io/quarkus/container/spi/ContainerImageResultBuildItem.java
+++ b/extensions/container-image/spi/src/main/java/io/quarkus/container/spi/ContainerImageResultBuildItem.java
@@ -1,8 +1,8 @@
 package io.quarkus.container.spi;
 
-import io.quarkus.builder.item.SimpleBuildItem;
+import io.quarkus.builder.item.MultiBuildItem;
 
-public final class ContainerImageResultBuildItem extends SimpleBuildItem {
+public final class ContainerImageResultBuildItem extends MultiBuildItem {
 
     private final String provider;
     private final String imageId;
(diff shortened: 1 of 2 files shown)


gsmet:
@iocanel can you have a look at that one?

geoand:
@iocanel WDYT?

iocanel (commented):
Sorry for the delayed review.

Converting to a multi build item, will fix the error message, but will result in performing a build using multiple services. I feel that this will rarely reflect the actual needs of the user. I can't think of a use case where the user would like to perform a `docker` and a `jib` build, or a `docker` and an `s2i` build.

So, I would expect, something like a selector instead.

geoand:
WDYM by selector?

iocanel:
A mechanism to select which service to use. 
For example:

`quarkus.container-image.preferred-service=jib`

or 

`quarkus.container-image.preferred-service=jib, s2i`

geoand:
That would definitely be the next step. But until we do that we need a better error message, because currently the user has no indication what the problem is

gsmet (approved):

PR #7515, opened 2020-03-02
add short-name to reactive-messaging-amqp extension metadata

gemmellr:
The guide at https://quarkus.io/guides/amqp for the reactive-messaging-amqp extension indicates use of -Dextensions="amqp" when creating application template with quarkus-maven-plugin. When other extensions (e.g. in quarkus platform) also reference AMQP in their details the existing creation command fails as quarkus-maven-plugin detects multiple extensions.

The same scenario of multiple extensions exists for the reactive-messaging-kafka extension and its guide https://quarkus.io/guides/kafka, however in practice the scenario with -Dextensions="kafka" actually does work regardless.

The difference seems to be that the reactive-messaging-kafka extension defines an additional short-name metadata at:
https://github.com/quarkusio/quarkus/blob/1.2.1.Final/extensions/smallrye-reactive-messaging-kafka/runtime/src/main/resources/META-INF/quarkus-extension.yaml#L4

This PR makes similar addition to the reactive-messaging-amqp extension to maintain the historic behaviour.

### DIFF
--- a/extensions/smallrye-reactive-messaging-amqp/runtime/src/main/resources/META-INF/quarkus-extension.yaml
+++ b/extensions/smallrye-reactive-messaging-amqp/runtime/src/main/resources/META-INF/quarkus-extension.yaml
@@ -1,6 +1,7 @@
 ---
 name: "SmallRye Reactive Messaging - AMQP Connector"
 metadata:
+  short-name: "amqp"
   keywords:
   - "amqp"
   - "reactive-amqp"


gemmellr:
CI build run passed on my fork at: https://github.com/gemmellr/quarkus/actions/runs/48190219

gemmellr:
@cescoffier could you give this a look and see what you think?

gemmellr:
The CI issue seems to be from a timeout issue with the KafkaStreamsTest creating a topic, rather than anything from the change itself (which doesn't change that module, and already passed in the fork). I cant restart the test so someone else will need to if required.

gemmellr:
Rebased to provoke a rebuild.

gastaldi (approved):
LGTM

gsmet:
Which extensions are returned when you try to add `amqp` before this patch?

gemmellr:
Historically, using -Dextensions="amqp" just went ahead and used smallrye-reactive-messaging-amqp. Thats what the patch aims to preserve.

I noticed in some testing I was doing that the extension I have a PR open at Quarkus Platform for Qpid's AMQP JMS client (https://github.com/quarkusio/quarkus-platform/pull/28) also then got picked up by just specifying "amqp" as the guide for smallrye-reactive-messaging-amqp suggested. (I had originally thought that -Dextensions="..." needed the artifact id, but evidently its smarter as I came to realise from the guide).

Realising that similar situation arises with the kafka extensions (where -Dextensions="kafka" simply uses smallrye-reactive-messaging-kafka, despite e.g kafka-streams and kafka-client also existing) I came to the conclusion the short-name was what made that work. I tried it out and found it got -Dextensions="amqp" back to simply using smallrye-reactive-messaging-amqp, and so raised this PR to retain the prior behaviour.

gsmet:
Merged, thanks!

PR #7518, opened 2020-03-02
Support OIDC client_secret_post client authentication

### LINKED ISSUE #6747: OIDC confidential clients can not authenticate with form params with Vertx 3.8.5
**Describe the bug**
A quarkus user works with the OIDC server which only accepts the form based authentication info during the token acquisition request. 

**Expected behavior**
Setting `Oauth2ClientOptions.useBasicAuthorizationHeader` to `false` should result in the client id and secret posted as the form parameters,

**Actual behavior**
`Oauth2ClientOptions.isUseBasicAuthorizationHeader` is ignored starting from Vertx `3.8.5`

**Additional context**

https://github.com/vert-x3/vertx-auth/issues/357

sberyozkin:
Fixes #6747.

This PR provides a workaround to support `quarkus-oidc` `web-app` applications working with the IDPs like ForgeRock (or other IDPs) which may require `client_secret_post` (`client_id` and `client_secret` posted as form url encoded parameters).

Along the way a `credentials.type` property is introduced. It will be very straightforward to support `client_secret_jwt`, which I will do a bit later.

### DIFF
--- a/extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/OidcTenantConfig.java
+++ b/extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/OidcTenantConfig.java
@@ -172,10 +172,20 @@ public class OidcTenantConfig {
     public static class Credentials {
 
         /**
-         * The client secret
+         * Client secret which is used for a 'client_secret_basic' authentication method.
+         * Note that a 'client-secret' can be used instead but both properties are mutually exclusive.
          */
         @ConfigItem
         Optional<String> secret = Optional.empty();
+
+        /**
+         * Client secret credentials which can be used for the 'client_secret_basic' (default)
+         * and 'client_secret_post' authentication methods.
+         * Note that a 'secret.value' property can be used instead to support the 'client_secret_basic' method
+         * but both properties are mutually exclusive.
+         */
+        @ConfigItem
+        Secret clientSecret = new Secret();
 
         public Optional<String> getSecret() {
             return secret;
@@ -183,6 +193,64 @@ public class OidcTenantConfig {
 
         public void setSecret(String secret) {
             this.secret = Optional.of(secret);
+        }
+
+        public Secret getClientSecret() {
+            return clientSecret;
+        }
+
+        public void setClientSecret(Secret clientSecret) {
+            this.clientSecret = clientSecret;
+        }
+
+        @ConfigGroup
+        public static class Secret {
+
+            /**
+             * Client secret authentication methods which specify how a client id and client secret
+             * have to be used to authenticate a client.
+             *
+             * @see <a href=
+             *      "https://openid.net/specs/openid-connect-core-1_0.html#ClientAuthentication">https://openid.net/specs/openid-connect-core-1_0.html#ClientAuthentication</a>
+             */
+            public static enum Method {
+                /**
+                 * client_secret_basic (default)
+                 */
+                BASIC,
+                /**
+                 * client_secret_post
+                 */
+                POST
+            }
+
+            /**
+             * The client secret
+             */
+            @ConfigItem
+            Optional<String> value = Optional.empty();
+
+            /**
+             * Authentication method.
+             */
+            @ConfigItem
+            Optional<Method> method = Optional.empty();
+
+            public Optional<String> getValue() {
+                return value;
+            }
+
+            public void setValue(String value) {
+                this.value = Optional.of(value);
+            }
+
+            public Optional<Method> getMethod() {
+                return method;
+            }
+
+            public void setMethod(Method method) {
+                this.method = Optional.of(method);
+            }
         }
     }
 
--- a/extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/CodeAuthenticationMechanism.java
+++ b/extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/CodeAuthenticationMechanism.java
@@ -17,6 +17,8 @@ import io.quarkus.oidc.AccessTokenCredential;
 import io.quarkus.oidc.IdTokenCredential;
 import io.quarkus.oidc.RefreshToken;
 import io.quarkus.oidc.runtime.OidcTenantConfig.Authentication;
+import io.quarkus.oidc.runtime.OidcTenantConfig.Credentials;
+import io.quarkus.oidc.runtime.OidcTenantConfig.Credentials.Secret;
 import io.quarkus.security.identity.IdentityProviderManager;
 import io.quarkus.security.identity.SecurityIdentity;
 import io.quarkus.security.runtime.QuarkusSecurityIdentity;
@@ -180,11 +182,22 @@ public class CodeAuthenticationMechanism extends AbstractOidcAuthenticationMecha
             return cf;
         }
 
+        // Code grant request
+        // 'code': the code grant value returned from IDP
         params.put("code", code);
+
+        // 'redirect_uri': typically it must match the 'redirect_uri' query parameter which was used during the code request.
         String redirectPath = getRedirectPath(configContext, absoluteUri);
         String redirectUriParam = buildRedirectUri(context, absoluteUri, redirectPath);
         LOG.debugf("Token request redirect_uri parameter: %s", redirectUriParam);
         params.put("redirect_uri", redirectUriParam);
+
+        // Client secret has to be posted as a form parameter if OIDC requires the client_secret_post authentication
+        Credentials creds = configContext.oidcConfig.getCredentials();
+        if (creds.clientSecret.value.isPresent() && creds.clientSecret.method.isPresent()
+                && Secret.Method.POST == creds.clientSecret.method.get()) {
+            params.put("client_secret", creds.clientSecret.value.get());
+        }
 
         configContext.auth.authenticate(params, userAsyncResult -> {
             if (userAsyncResult.failed()) {
--- a/extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/OidcRecorder.java
+++ b/extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/OidcRecorder.java
@@ -11,7 +11,10 @@ import org.jboss.logging.Logger;
 
 import io.quarkus.arc.runtime.BeanContainer;
 import io.quarkus.oidc.OIDCException;
+import io.quarkus.oidc.runtime.OidcTenantConfig.Credentials;
+import io.quarkus.oidc.runtime.OidcTenantConfig.Credentials.Secret;
 import io.quarkus.runtime.annotations.Recorder;
+import io.quarkus.runtime.configuration.ConfigurationException;
 import io.vertx.core.AsyncResult;
 import io.vertx.core.Handler;
 import io.vertx.core.Vertx;
@@ -77,9 +80,25 @@ public class OidcRecorder {
             options.setClientID(oidcConfig.getClientId().get());
         }
 
-        if (oidcConfig.getCredentials().secret.isPresent()) {
-            options.setClientSecret(oidcConfig.getCredentials().secret.get());
+        Credentials creds = oidcConfig.getCredentials();
+        if (creds.secret.isPresent() && (creds.clientSecret.value.isPresent() || creds.clientSecret.method.isPresent())) {
+            throw new ConfigurationException(
+                    "'credentials.secret' and 'credentials.client-secret' properties are mutually exclusive");
         }
+        // TODO: The workaround to support client_secret_post is added below and have to be removed once
+        // it is supported again in VertX OAuth2.
+
+        if (creds.secret.isPresent()
+                || creds.clientSecret.value.isPresent()
+                        && creds.clientSecret.method.orElseGet(() -> Secret.Method.BASIC) == Secret.Method.BASIC) {
+            // If it is set for client_secret_post as well then VertX OAuth2 will only use client_secret_basic
+            options.setClientSecret(creds.secret.orElseGet(() -> creds.clientSecret.value.get()));
+        } else {
+            // Avoid the client_secret set in CodeAuthenticationMechanism when client_secret_post is enabled
+            // from being reset to null in VertX OAuth2
+            options.setClientSecretParameterName(null);
+        }
+
         if (oidcConfig.getPublicKey().isPresent()) {
             options.addPubSecKey(new PubSecKeyOptions()
                     .setAlgorithm("RS256")
--- a/integration-tests/oidc-code-flow/src/main/resources/application.properties
+++ b/integration-tests/oidc-code-flow/src/main/resources/application.properties
@@ -3,6 +3,7 @@
 # Default tenant configuration
 quarkus.oidc.auth-server-url=${keycloak.url}/realms/quarkus
 quarkus.oidc.client-id=quarkus-app
+quarkus.oidc.credentials.secret=secret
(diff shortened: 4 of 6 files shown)


extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/OidcTenantConfig.java
```diff
@@ -171,6 +171,20 @@ public void setTenantId(String tenantId) {
     @ConfigGroup
     public static class Credentials {
 
+        public static enum Type {
```
pedroigor: I'm not sure about having this as a credential type. For me, credentials types as jwt, secret, mtls, etc. Which are related to the method that will be used to authenticate the client.

What if you have `quarkus.oidc.credentials.secret.method` instead ? Where `secret` would become a config object ?

pedroigor (approved):

extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/OidcTenantConfig.java
sberyozkin: Hi @pedroigor This looks a bit complex ? In the simple case the user would have to do `quarkus.oidc.credentials.secret.value=secret` ?
I agree `quarkus.oidc.credentials.secret` should've ideally be part of some other object like client authentication, but it is already there now, so I though we could avoid another migration problem and qualify how that secret should be used, with this `Type` covering `client_secret_basic`, `client_secret_post` and easily `client_secret_jwt`.
When we will need to support `private_key_jwt` or mtls, then we'd have some new convention/holder for those and can just doc that `secret` property is ignored unless it is `client_secret_basic`, `client_secret_post` and easily `client_secret_jwt`.

Perhaps I can rename `Type` to `Method` and doc that this `Method` identifies how the credentials like 'secret' (or private_key or cert in the future) are used.

Would it be a reasonable compromise ?

Thanks

sberyozkin: @pedroigor To summarize, adding an extra `credentials.method` property will let us support 3 out of 4 client authentication methods defined in the OIDC core as they are all built around the use of the client secret, all without introducing a migration requirement. 
To support 2 more methods (private key for signing the jwt, mtls) we can either breaking configuration change and have a configuration object per every auth method or 3 at least, or add a Map of Maps to `credentials` and say the key must match the Method enum value, something like that, but we should probably deal with it later :-)

sberyozkin:
@pedroigor Have a look please how does it look with `Method` (just updated the PR). Np if you will still reckon we will need to do it differently from now on, cheers

sberyozkin:
We've had a brief discussion with Pedro, and we've agreed some it is better to put some more thinking into it right now instead of rushing in... I'll think more

sberyozkin:
Hi Pedro, so I've spent the rest of the day looking at the various options. This is what I have right now as one of the options:
```
/**
     * Client secret credentials which the OIDC adapter will use to authenticate to the OIDC server.
     * Note this property can be used to support the client_secret_basic and client_secret_post authentication methods
     * and is mutually exclusive with the 'client-jwt' property.
     */
    @ConfigItem
    SimpleClientSecretMethod credentials = new SimpleClientSecretMethod();
    
    /**
     * Client secret credentials which the OIDC adapter will use to create a signed JWT to authenticate to the OIDC server.
     * Note this property can be used to support the client_secret_jwt authentication methods only
     * and is mutually exclusive with the 'credentials' property.
     */
    @ConfigItem
    ClientSecretJwtMethod clientSecretJwt = new ClientSecretJwtMethod();

    // similarly for the future private_key_jwt and mtls

    private static abstract class AbstractClientSecretMethod {

        /**
         * The client secret
         */
        @ConfigItem
        Optional<String> secret = Optional.empty();

        public Optional<String> getSecret() {
            return secret;
        }

        public void setSecret(String secret) {
            this.secret = Optional.of(secret);
        }
    }
    
    @ConfigGroup
    public static class SimpleClientSecretMethod extends AbstractClientSecretMethod {
    	/**
         * Simple client secret authentication method which control how the client id and a credential
         * such as a client secret must be be used to authenticate the client during the code grant and token introspection requests.
         *
         * @see <a href=
         *      "https://openid.net/specs/openid-connect-core-1_0.html#ClientAuthentication">https://openid.net/specs/openid-connect-core-1_0.html#ClientAuthentication</a>
         */

    	public static enum Method {
            BASIC,
            POST
        }

        @ConfigItem(defaultValue = "BASIC")
        Method method;
    }
    
    @ConfigGroup
    public static class ClientSecretJwtMethod extends AbstractClientSecretMethod {
    }
    
    @ConfigGroup
    public static class ClientPrivateKeyJwtMethod {
    }
    
    @ConfigGroup
    public static class ClientMutualTlsMethod {
    }
```

This looks not too bad but I have this feeling, that it is a bit over the board when it comes to differentiating between `client_secret_basic`, `client_secret_post` and `client_secret_jwt` as the only reason we do a dedicated JWT group is to support either `HS384` or `HS512` and here I'm thinking this is something noone will likely ask us for :-), if people are really concerned about `HS256` then they'd just go for `private_key_jwt` with `RS512` and 5096 bit length key :-).

I've also tried to have something like:
```
// keep the current property to support client_secret_basic only
Credentials credentials;

public static enum ClientAuthenticationMethod {
            SECRET_BASIC,
            SECRET_POST,
            SECRET_JWT,
            PRIVATE_KEY_JWT,
            MTLS
        }
@ConfigItem
Map<ClientAuthenticationMethod, Map<String, String>> clientAuthentication;
```

This works generically but there is no typed support for the specific methods.

I'm still fond enough of what is currently in this PR, I'd only rename `Credentials` to `ClientSecretAuthenticationMethod`, have `Method` with `BASIC`/`POST`/`JWT` (yes, only `HS256` :-)), simple and works, and then in due time go further with the dedicated groups for private_key_jwt and MTLS. If we ever need `HS512` for `client_secret_jwt`, then we may have:
```
// set algorithm=HS512 if Method=JWT
Map<String, String> extraParams;
```

etc...I'll have to start looking at some non OIDC issue tomorrow, let me know what you think please, prototype some other options as well :-)

Thanks

pedroigor:
@sberyozkin what about these changes https://github.com/pedroigor/quarkus/tree/issue-6747 ?

pedroigor:
@sberyozkin I think `credentials.secret.method` won't be read because of the `SecretConverter` but we could just have `secret` for BASIC (as it stands today) and `secret-post` to tell that secret whould be sent using `POST` instead of the default `BASIC`.

pedroigor:
I think there is something we are not considering here. The OpenID Discovery Document gives us the authentication methods supported by the server (https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfigurationResponse). We can infer if we should do `BASIC` or `POST` based on what is advertised by the OP? Would that solve the original issue?

Regarding the credentials config discussion.

I really like the pattern `credentials.<authentication-method>.<options>`. It is clear, objective and easier to read so that we have a config group for each authentication method. It is also easier to extend too so that new methods would have their own config group, independently.

I don't see any issue about having `secret` duplicated in both `client_secret_*` and `client_secret_jwt`. The reason is that when doing `client_secret_jwt` you are actually passing/using a JWT to authenticate and not the secret (ok, the authentication is done checking the signature that is based on the secret, but I think you get my point).

And if we really want to allow users to force `client_secret_post` we should just have a `credentials.secret-post`.

sberyozkin:
@pedroigor 
`credentials.<authentication-method>.<options>` - it is inverted IMHO, there is a method (how to authenticate), the credentials is a secondary thing, they are there to implement the authentication method, this is why I've tried to rename `Credentials`. 
And IMHO  the discovery can be that useful if both methods are supported and either we need to let the user tell how they prefer the clients authenticated, esp if no discovery is available.

OK. Lets try to find a compromise. How about we just leave JWT out of the discussion for now, it is really not related to the original/current issue. We need to support either client basic (default) or 'post'. I prefer no to have 'secret' 'secret_post', 2 mutually exclusive fields, with the converter, which can be easily avoided by a qualifier.

Let me offer another option. We don't touch `Credentials`, because it is not an authentication method. It was fine for an initial quick support to set a secret. But trying to squeeze the authentication methods into it does not seem to work. 

```
Credentials {
    // client basic - this is just a shortcut, we don't really need an extra converter 
    // and only keep it because users already use it widely
    String secret;
}

This gives us the usual:

```
quarkus.credentials.secret=secret
```

Then lets have an independent `AuthenticationMethod` config group, because this is exactly what we trying to do, support different client secret authentication methods.
```
class OidcTenantConfig {

AuthenticationMethod clientAuthentication = new AuthenticationMethod();
class AuthenticationMethod {
    Secret secretMethod = new Secret();
    // more methods like JWT/etc later    
}
// This is close to your git prototype
class Secret {
    String value;
    // default is Basic
    Method method;
    enum Method {
        BASIC, POST
    }
}
}
```
It gives:
```
# same as the credentials.secret
client-authentication.secret-method.secret=mysecret`
#client-authentication.secret-method.method=basic
```

or 

```
# same as the credentials.secret
client-authentication.secret-method.secret=mysecret
client-authentication.secret-method=post
```

I don't. I feel like I'm not getting confused with all these options

**Comment by sberyozkin** [comment]
Hi @pedroigor if you are not happy with my previous proposal too, I thought, what I typed earlier,
```
    we keep Credentials.secret
    we add Credentials.Secret with BASIC/POST/JWT and it becomes Credentials.secret-authentication
    we say Credentials.secret is a shortcut for Credentials.secret-authentication.secret=abc and Credentials.secret-authentication.method=basic, post/jwt with secret is easy too.
    we add your Jwt group renamed to PrivateKeyJwt to Credentials, etc, for MTLS
```

is very close to your point of having a configuration group per authentication method, which becomes even simpler if we put all the JWT things aside for now, no need for converters too. 
Then the only contentious point becomes is how to express client-secret basic vs post :-). As you've said before `client_secret_basic` and `client_secret_post` are special. Can we settle on what I copied above (minus JWT) ? The fact that we will have your `Secret` having a method property is a reasonable compromise IMHO:

```
class Credentials {
    // Nothing changes here, this is a client secret basic shortcut 
    String secret;

    Secret secretMethod;
    class Secret {
      String secret;
      // default is Basic
      Method method;
      enum Method {
          BASIC, POST
      }
  } 

   // later: add ClientSecretJwt, PrivateKeyJwt, Mtls groups
}
```

Thoughts ?

**Comment by pedroigor** [comment]
The `credentials.<authentication-method>.<options>` that I said to like can be changed to `credentials.<type>.<options>`. You have a point when saying when you are actually defining an authentication method, but my point is that it can be inferred from the credential type. I think that is why I think it is simpler and reads better for users.

So, that is my preference.

But your proposal also makes sense for me. So having `quarkus.oidc.authentication.secret.secret` and `quarkus.oidc.authentication.secret.method` looks better (no `client-` prefix and no `-method`suffix).

Do you think we should ask someone else opinion about this?

**Comment by sberyozkin** [comment]
Hi @pedroigor, my latest proposal is is really built around your proposal, whichever way it ends up it will be a joint effort :-). I think we are trying to reach the same goal but only via slightly different paths. It takes us a bit of time to agree :-), but I'd rather us be on the same line.

I've named it `clientAuthentication` because we already have `authentication` for the code flow.
I'm happy with dropping `-method`: `quarkus.oidc.client-authentication.secret.secret=secret`, or may be even `quarkus.oidc.authentication.secret.value=secret` ? 

Re your proposal. So just so that I understand, you propose to have something like per every auth method:
```
Credentials {
    // client-basic 
    Optional<String> secret;
    // client-post
    Optional<String> secretPost;

   // later
   Jwt jwt = new Jwt();
   class Jwt {
       Optional<String> secret;
       Optional<String> algorithm;
   }
   // later
   PrivateJwt jwt = new PrivateJwt();
   class PrivateJwt {
       // various properties on the private key
   }  
}
```

I.e, no converters, no `Secret` group, no converter ? We can go with it too, my only reservation, it is just a client secret, post or basic, it is an extra qualifier, so `secretPost` just does not look too good to me. But not a big deal.

Lets finalize, what is your final preference ?

Thanks

> Do you think we should ask someone else opinion about this?

@stuartwdouglas, Stian, and everyone else are welcome to comment :-)

**Comment by sberyozkin** [comment]
@pedroigor I was trying my best to be totally objective when thinking about it while I was offline :-).
It feels like if we go this route where we keep `Credentials` unchanged, and introduce `AuthenticationMethod` with a Secret group for a start, then it may present an opportunity for us to deprecate `Credentials` over time. Ultimately, `quarkus.oidc.client-authentication.secret.value=secret` for s default basic case is not too bad, a bit more verbose compared to a simpler  `quarkus.oidc.credentials.secret=secret` but we will have a pattern with a method specific config groups, which both of us agree on it for sure :-)

`quarkus.oidc.client-authentication.secret.value=secret` can also become 

`quarkus.oidc.client-method.secret.value=secret`

Anyway, as I said, lets try to finalize as it would be good to have it fixed for 1.3.0, let me know your final preference please, and if @stuartwdouglas or @stianst also have preferences, then it would be good to see them.

**Comment by pedroigor** [comment]
@sberyozkin That is the beauty of our work, discuss ideas :)

I agree about the `secret-post` not looking good. But in a universe of multiple credential types, this is the only exception. I would even agree about a `credentials.secret.value` and `credentials.secret.method`.

**Comment by pedroigor** [comment]
@sberyozkin Btw, there is a saying here: "Every truth has three versions. Mine, yours and the real truth." :)

**Comment by sberyozkin** [comment]
@pedroigor nice, then we can have the both goals achieved, keep the original `secret` and start offering users an option to explicitly configure the methods:
```
// keeping the current shortcut
credentials.secret=mysecret
// new mechanism
credentials.client-secret.value=mysecret
credentials.client-secret.method=post
```
I'll try to prototype something soon enough, thanks

sberyozkin:
@pedroigor How does it look now ? I've added checks preventing the users mixing in `credentials.secret` (we still keep it as planned to support client_secret_basic OOB) and `credentials.client-secret`.
I'll open an issue to support `client_secret_jwt` later (we can then finalize if we'd want a new `Jwt` group for the client secret signing option, or add a new `Method`, either option will be fine for me; IMHO we should have a dedicated group for the `private_key_jwt` as per your prototype)
Would be good to have `Optional<Secret>` but it does not work right now, not a big problem, we can optimize later.
Thanks

sberyozkin:
`in io.quarkus.it.jpa.mssql.JPAFunctionalityTest` has failed

pedroigor (approved):

sberyozkin:
Hi Pedro, @pedroigor, thanks, one more question. Should we consider just have `credentials.secret.value` starting from Quarkus `1.3.0` ? 
I.e `credentials.secret` will be gone, there will be no need to support `credentials.secret` and differentiate with `credentials.client-secret`, we update the migration and tell the users to replace `credentials.secret` with `credentials.secret.value` which will be enough for 100% migration cases, given that `client_secret_post` does not work anyway right now with `1.3.0.CR1`.

I'll be fine if you'd prefer not to do it now. We can then revisit it later on...thanks

pedroigor:
+1000. Having `credentials.secret` as a config group is the middle/long term solution we should aim for. Just easier, simple to read and meaninful.

sberyozkin:
@pedroigor Thanks, I assume your preference is to do it for the post `1.3.0` major release such as `1.4.0` (given your reference to the medium/long term). I can create then an issue for 1.4.0 once this one is resolved.

gsmet (approved):
Let's get this one into CR2 as Pedro approved it.

PR #7520, opened 2020-03-02
Fix some datasource config issues

gsmet:
@geoand this should be the right fix for the issue you reported with the Hibernate Search quickstart.

I added 2 more commits with some minor adjustments.


geoand (commented):
I just have a question about `setUnremovable()`. Why is that needed?

gsmet:
@geoand well, because otherwise they are removed :). They are only consumed dynamically via the Arc API.

geoand:
But how where tests passing without this change?

extensions/agroal/runtime/src/main/java/io/quarkus/agroal/runtime/JdbcDriver.java
```diff
@@ -22,20 +22,20 @@
 @Qualifier
 public @interface JdbcDriver {
 
-    String name();
+    String value();
```
geoand: :+1:

geoand (approved):

gsmet:
Because currently they are only used for an implementation detail (i.e. pass an option to the driver) which doesn't really fail. And it fails gracefully if we don't find one because you can use an external driver.

They should be used more in the future but for now it's not really testable.

The capability not being produced was a merge issue at some point because it was there at the beginning.

gsmet:
I noticed the issue because the exception you had should have been on the previous line, not that one.

PR #7524, opened 2020-03-03
Move Agroal metric classes into private API

jmartisk:
Based on Zulip discussion https://quarkusio.zulipchat.com/#narrow/stream/187038-dev/topic/AgroalGauge.20and.20AgroalCounter


gsmet:
I rebased following another commit in this area. I remove the breaking change tag. I think we can consider it hasn't been used.

gsmet (approved):

PR #7526, opened 2020-03-03
Fix metrics of the default Agroal data source

### LINKED ISSUE #7525: Metrics for default Agroal data source are broken
After https://github.com/quarkusio/quarkus/commit/0d64d8596242d1900b1f62334e2d89c6fbeaa3aa the metrics for Agroal default data source broke, because `AgroalCounter` and `AgroalGauge` expect `null` to be passed as the data source name when referring to the default DS. Now, instead of `null`, we pass them the name `<default>`, so the metric classes are unable to locate the correct data source.

This results in `java.lang.NullPointerException` appearing in the log every time metrics are queried, and metrics for the default data source don't appear in the output.

jmartisk:
Fixes #7525

### DIFF
--- a/extensions/agroal/deployment/pom.xml
+++ b/extensions/agroal/deployment/pom.xml
@@ -73,6 +73,11 @@
             <artifactId>quarkus-jdbc-h2-deployment</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-smallrye-metrics</artifactId>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 
     <build>
--- a/extensions/agroal/deployment/src/test/java/io/quarkus/agroal/test/AgroalMetricsTestCase.java
+++ b/extensions/agroal/deployment/src/test/java/io/quarkus/agroal/test/AgroalMetricsTestCase.java
@@ -1,0 +1,78 @@
+package io.quarkus.agroal.test;
+
+import java.sql.Connection;
+import java.sql.SQLException;
+import java.sql.Statement;
+
+import javax.inject.Inject;
+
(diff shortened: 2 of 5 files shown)


geoand (approved):

extensions/agroal/deployment/src/main/java/io/quarkus/agroal/deployment/AgroalProcessor.java
```diff
@@ -551,89 +551,91 @@ void registerMetrics(
             boolean metricsEnabledForThisDatasource = dataSourcesBuildTimeConfig.metricsEnabled &&
                     aggregatedDataSourceBuildTimeConfig.getJdbcConfig().enableMetrics.orElse(true);
             Tag tag = new Tag("datasource", DataSourceUtil.isDefault(dataSourceName) ? "default" : dataSourceName);
+            // Metric classes expect null as the name for the default DS
+            String dataSourceNameOrNull = DataSourceUtil.isDefault(dataSourceName) ? null : dataSourceName;
```
gsmet: Sorry I missed that but... could we adjust that, pass the name to the gauge and counter and use `DataSourceUtil.isDefault()` there?

That would be more consistent and avoid this sort of issues.

gsmet (changes requested):

extensions/agroal/deployment/src/main/java/io/quarkus/agroal/deployment/AgroalProcessor.java
jmartisk: Sounds like a good idea. Done

gsmet (approved):
Thanks!

gsmet:
Force pushed a rebase.

gsmet:
Thanks, good catch @jmartisk !

PR #7527, opened 2020-03-03
Panache : deleteById()

### LINKED ISSUE #6270: Provide deleteById() for MongoDB and Hibernate with Panache
**Description**
Currently, to delete an entity in both MongoDb and Hibernate by Panache we can only do it from an entity instance.

So in traditional REST API, when we implement a DELETE: /entity/id we usually load the entity from the database then delete it. Alternatively we can use Entity.delete("id", id) but in this case we must check the resulting number of deleted results to check that the entity has been deleted (and generates a 404 if not).

Providing a `Entity.deleteById(id)` method will make it easy to delete from an ID.
It should throw an exception if no delete happens.

loicmathieu:
Add the deleteById() method for both Hibernate and MongoDB with Panache.

Fixes #6270

### DIFF
--- a/docs/src/main/asciidoc/hibernate-orm-panache.adoc
+++ b/docs/src/main/asciidoc/hibernate-orm-panache.adoc
@@ -195,6 +195,9 @@ Person.delete("status", Status.Alive);
 // delete all persons
 Person.deleteAll();
 
+// delete by id
+boolean deleted = Person.deleteById(personId);
+
 // update all living persons
 Person.update("name = 'Moral' where status = ?1", Status.Alive);
 
@@ -378,6 +381,9 @@ personRepository.delete("status", Status.Alive);
 
 // delete all persons
 personRepository.deleteAll();
+
+// delete by id
+boolean deleted = personRepository.deleteById(personId);
 
 // update all living persons
 personRepository.update("name = 'Moral' where status = ?1", Status.Alive);
--- a/docs/src/main/asciidoc/mongodb-panache.adoc
+++ b/docs/src/main/asciidoc/mongodb-panache.adoc
@@ -217,6 +217,9 @@ Person.delete("status", Status.Alive);
 
 // delete all persons
 Person.deleteAll();
+
+// delete by id
+boolean deleted = Person.deleteById(personId);
 ----
 
 All `list` methods have equivalent `stream` versions.
@@ -378,6 +381,9 @@ personRepository.delete("status", Status.Alive);
 
 // delete all persons
 personRepository.deleteAll();
+
+// delete by id
+boolean deleted = personRepository.deleteById(personId);
 ----
 
 All `list` methods have equivalent `stream` versions.
@@ -771,6 +777,9 @@ Uni<Long>  deleteCount = ReactivePerson.delete("status", Status.Alive);
 
 // delete all persons
 deleteCount = ReactivePerson.deleteAll();
+
+// delete by id
+Uni<Boolean> deleted = ReactivePerson.deleteById(personId);
 ----
 
 TIP: If you use MongoDB with Panache in conjunction with RESTEasy, you can directly return a reactive type inside your JAX-RS resource endpoint as long as you include the `quarkus-resteasy-mutiny` extension.
--- a/extensions/panache/hibernate-orm-panache/runtime/src/main/java/io/quarkus/hibernate/orm/panache/PanacheEntityBase.java
+++ b/extensions/panache/hibernate-orm-panache/runtime/src/main/java/io/quarkus/hibernate/orm/panache/PanacheEntityBase.java
@@ -642,7 +642,7 @@ public abstract class PanacheEntityBase {
      * Delete an entity of this type by ID.
      *
      * @param id the ID of the entity to delete.
-     * @return false if the entity is not delete (not found).
+     * @return false if the entity was not deleted (not found).
      */
     @GenerateBridge
     public static boolean deleteById(Object id) {
--- a/extensions/panache/hibernate-orm-panache/runtime/src/main/java/io/quarkus/hibernate/orm/panache/PanacheEntityBase.java
+++ b/extensions/panache/hibernate-orm-panache/runtime/src/main/java/io/quarkus/hibernate/orm/panache/PanacheEntityBase.java
@@ -639,6 +639,17 @@ public abstract class PanacheEntityBase {
     }
 
     /**
+     * Delete an entity of this type by ID.
+     *
+     * @param id the ID of the entity to delete.
+     * @return false if the entity is not delete (not found).
+     */
+    @GenerateBridge
+    public static boolean deleteById(Object id) {
+        throw JpaOperations.implementationInjectionMissing();
+    }
+
+    /**
      * Delete all entities of this type matching the given query, with optional indexed parameters.
      *
      * @param query a {@link io.quarkus.hibernate.orm.panache query string}
--- a/extensions/panache/hibernate-orm-panache/runtime/src/main/java/io/quarkus/hibernate/orm/panache/PanacheRepositoryBase.java
+++ b/extensions/panache/hibernate-orm-panache/runtime/src/main/java/io/quarkus/hibernate/orm/panache/PanacheRepositoryBase.java
@@ -635,6 +635,17 @@ public interface PanacheRepositoryBase<Entity, Id> {
     }
 
     /**
+     * Delete an entity of this type by ID.
+     *
+     * @param id the ID of the entity to delete.
+     * @return false if the entity is not deleted (not found).
+     */
+    @GenerateBridge
+    public default boolean deleteById(Id id) {
+        throw JpaOperations.implementationInjectionMissing();
+    }
+
+    /**
      * Delete all entities of this type matching the given query, with optional indexed parameters.
      * 
      * @param query a {@link io.quarkus.hibernate.orm.panache query string}
--- a/extensions/panache/hibernate-orm-panache/runtime/src/main/java/io/quarkus/hibernate/orm/panache/PanacheRepositoryBase.java
+++ b/extensions/panache/hibernate-orm-panache/runtime/src/main/java/io/quarkus/hibernate/orm/panache/PanacheRepositoryBase.java
@@ -638,7 +638,7 @@ public interface PanacheRepositoryBase<Entity, Id> {
      * Delete an entity of this type by ID.
      *
      * @param id the ID of the entity to delete.
-     * @return false if the entity is not deleted (not found).
+     * @return false if the entity was not deleted (not found).
      */
     @GenerateBridge
     public default boolean deleteById(Id id) {
--- a/extensions/panache/hibernate-orm-panache/runtime/src/main/java/io/quarkus/hibernate/orm/panache/runtime/JpaOperations.java
+++ b/extensions/panache/hibernate-orm-panache/runtime/src/main/java/io/quarkus/hibernate/orm/panache/runtime/JpaOperations.java
@@ -393,6 +393,17 @@ public class JpaOperations {
         return (long) getEntityManager().createQuery("DELETE FROM " + getEntityName(entityClass)).executeUpdate();
     }
 
+    public static boolean deleteById(Class<?> entityClass, Object id) {
+        // Impl note : we load the entity then delete it because it's the only implementation generic enough for any model,
+        // and correct in all cases (composite key, graph of entities, ...). HQL cannot be directly used for these reasons.
+        Object entity = findById(entityClass, id);
+        if (entity == null) {
+            return false;
+        }
+        getEntityManager().remove(entity);
+        return true;
+    }
+
     public static long delete(Class<?> entityClass, String query, Object... params) {
         return bindParameters(getEntityManager().createQuery(createDeleteQuery(entityClass, query, paramCount(params))), params)
                 .executeUpdate();
--- a/extensions/panache/mongodb-panache/runtime/src/main/java/io/quarkus/mongodb/panache/PanacheMongoEntityBase.java
+++ b/extensions/panache/mongodb-panache/runtime/src/main/java/io/quarkus/mongodb/panache/PanacheMongoEntityBase.java
@@ -698,6 +698,17 @@ public abstract class PanacheMongoEntityBase {
     }
 
     /**
+     * Delete an entity of this type by ID.
+     *
+     * @param id the ID of the entity to delete.
+     * @return false if the entity was not deleted (not found).
+     */
+    @GenerateBridge
+    public static boolean deleteById(Object id) {
+        throw MongoOperations.implementationInjectionMissing();
+    }
+
+    /**
      * Delete all entities of this type matching the given query, with optional indexed parameters.
      *
      * @param query a {@link io.quarkus.mongodb.panache query string}
--- a/extensions/panache/mongodb-panache/runtime/src/main/java/io/quarkus/mongodb/panache/PanacheMongoRepositoryBase.java
+++ b/extensions/panache/mongodb-panache/runtime/src/main/java/io/quarkus/mongodb/panache/PanacheMongoRepositoryBase.java
@@ -703,6 +703,17 @@ public interface PanacheMongoRepositoryBase<Entity, Id> {
     }
 
     /**
+     * Delete an entity of this type by ID.
+     *
+     * @param id the ID of the entity to delete.
+     * @return false if the entity was not deleted (not found).
+     */
+    @GenerateBridge
+    public default boolean deleteById(Id id) {
+        throw MongoOperations.implementationInjectionMissing();
+    }
+
+    /**
      * Delete all entities of this type matching the given query, with optional indexed parameters.
      * 
      * @param query a {@link io.quarkus.mongodb.panache query string}
--- a/extensions/panache/mongodb-panache/runtime/src/main/java/io/quarkus/mongodb/panache/reactive/ReactivePanacheMongoEntityBase.java
+++ b/extensions/panache/mongodb-panache/runtime/src/main/java/io/quarkus/mongodb/panache/reactive/ReactivePanacheMongoEntityBase.java
@@ -8,6 +8,7 @@ import java.util.stream.Stream;
 import org.bson.Document;
 
 import io.quarkus.mongodb.panache.reactive.runtime.ReactiveMongoOperations;
+import io.quarkus.mongodb.panache.runtime.MongoOperations;
 import io.quarkus.mongodb.reactive.ReactiveMongoCollection;
 import io.quarkus.mongodb.reactive.ReactiveMongoDatabase;
 import io.quarkus.panache.common.Parameters;
@@ -705,6 +706,17 @@ public abstract class ReactivePanacheMongoEntityBase {
     @GenerateBridge
     public static Uni<Long> deleteAll() {
         throw ReactiveMongoOperations.implementationInjectionMissing();
+    }
+
+    /**
+     * Delete an entity of this type by ID.
+     *
+     * @param id the ID of the entity to delete.
+     * @return false if the entity was not deleted (not found).
+     */
+    @GenerateBridge
+    public static Uni<Boolean> deleteById(Object id) {
+        throw MongoOperations.implementationInjectionMissing();
     }
 
     /**
--- a/extensions/panache/mongodb-panache/runtime/src/main/java/io/quarkus/mongodb/panache/reactive/ReactivePanacheMongoRepositoryBase.java
+++ b/extensions/panache/mongodb-panache/runtime/src/main/java/io/quarkus/mongodb/panache/reactive/ReactivePanacheMongoRepositoryBase.java
@@ -8,6 +8,7 @@ import java.util.stream.Stream;
 import org.bson.Document;
 
 import io.quarkus.mongodb.panache.reactive.runtime.ReactiveMongoOperations;
+import io.quarkus.mongodb.panache.runtime.MongoOperations;
 import io.quarkus.mongodb.reactive.ReactiveMongoCollection;
 import io.quarkus.mongodb.reactive.ReactiveMongoDatabase;
 import io.quarkus.panache.common.Parameters;
@@ -701,6 +702,17 @@ public interface ReactivePanacheMongoRepositoryBase<Entity, Id> {
     @GenerateBridge
     public default Uni<Long> deleteAll() {
         throw ReactiveMongoOperations.implementationInjectionMissing();
+    }
+
+    /**
+     * Delete an entity of this type by ID.
+     *
+     * @param id the ID of the entity to delete.
+     * @return false if the entity was not deleted (not found).
+     */
+    @GenerateBridge
+    public default Uni<Boolean> deleteById(Id id) {
+        throw MongoOperations.implementationInjectionMissing();
     }
 
     /**
--- a/extensions/panache/mongodb-panache/runtime/src/main/java/io/quarkus/mongodb/panache/reactive/runtime/ReactiveMongoOperations.java
+++ b/extensions/panache/mongodb-panache/runtime/src/main/java/io/quarkus/mongodb/panache/reactive/runtime/ReactiveMongoOperations.java
@@ -551,6 +551,12 @@ public class ReactiveMongoOperations {
         return collection.deleteMany(new Document()).map(deleteResult -> deleteResult.getDeletedCount());
     }
 
+    public static Uni<Boolean> deleteById(Class<?> entityClass, Object id) {
+        ReactiveMongoCollection<?> collection = mongoCollection(entityClass);
+        Document query = new Document().append(ID, id);
+        return collection.deleteOne(query).map(results -> results.getDeletedCount() == 1);
+    }
+
     public static Uni<Long> delete(Class<?> entityClass, String query, Object... params) {
         String bindQuery = bindQuery(entityClass, query, params);
         Document docQuery = Document.parse(bindQuery);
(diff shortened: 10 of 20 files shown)


geoand:
@loicmathieu do you mind if I push a commit to your PR to update the Spring Data JPA implementation to use this new method?

loicmathieu:
@geoand of course not ! go ahead !

geoand:
Thanks, done :)

gastaldi (approved):
LGTM

extensions/panache/hibernate-orm-panache/runtime/src/main/java/io/quarkus/hibernate/orm/panache/PanacheEntityBase.java
```diff
@@ -638,6 +638,17 @@ public static long deleteAll() {
         throw JpaOperations.implementationInjectionMissing();
     }
 
+    /**
+     * Delete an entity of this type by ID.
+     *
+     * @param id the ID of the entity to delete.
+     * @return false if the entity is not delete (not found).
```
gastaldi:
```suggestion
     * @return false if the entity was not deleted (not found).
```

extensions/panache/hibernate-orm-panache/runtime/src/main/java/io/quarkus/hibernate/orm/panache/PanacheRepositoryBase.java
```diff
@@ -634,6 +634,17 @@ public default long deleteAll() {
         throw JpaOperations.implementationInjectionMissing();
     }
 
+    /**
+     * Delete an entity of this type by ID.
+     *
+     * @param id the ID of the entity to delete.
+     * @return false if the entity is not deleted (not found).
```
gastaldi:
```suggestion
     * @return false if the entity was not deleted (not found).
```

extensions/panache/mongodb-panache/runtime/src/main/java/io/quarkus/mongodb/panache/PanacheMongoEntityBase.java
```diff
@@ -697,6 +697,17 @@ public static long deleteAll() {
         throw MongoOperations.implementationInjectionMissing();
     }
 
+    /**
+     * Delete an entity of this type by ID.
+     *
+     * @param id the ID of the entity to delete.
+     * @return false if the entity is not delete (not found).
```
gastaldi:
```suggestion
     * @return false if the entity was not deleted (not found).
```

extensions/panache/mongodb-panache/runtime/src/main/java/io/quarkus/mongodb/panache/PanacheMongoRepositoryBase.java
```diff
@@ -702,6 +702,17 @@ public default long deleteAll() {
         throw MongoOperations.implementationInjectionMissing();
     }
 
+    /**
+     * Delete an entity of this type by ID.
+     *
+     * @param id the ID of the entity to delete.
+     * @return false if the entity is not deleted (not found).
```
gastaldi:
```suggestion
     * @return false if the entity was not deleted (not found).
```

extensions/panache/mongodb-panache/runtime/src/main/java/io/quarkus/mongodb/panache/reactive/ReactivePanacheMongoEntityBase.java
```diff
@@ -707,6 +708,17 @@
         throw ReactiveMongoOperations.implementationInjectionMissing();
     }
 
+    /**
+     * Delete an entity of this type by ID.
+     *
+     * @param id the ID of the entity to delete.
+     * @return false if the entity is not delete (not found).
```
gastaldi:
```suggestion
     * @return false if the entity was not deleted (not found).
```

extensions/panache/mongodb-panache/runtime/src/main/java/io/quarkus/mongodb/panache/reactive/ReactivePanacheMongoRepositoryBase.java
```diff
@@ -703,6 +704,17 @@
         throw ReactiveMongoOperations.implementationInjectionMissing();
     }
 
+    /**
+     * Delete an entity of this type by ID.
+     *
+     * @param id the ID of the entity to delete.
+     * @return false if the entity is not deleted (not found).
```
gastaldi:
```suggestion
     * @return false if the entity was not deleted (not found).
```

loicmathieu:
@gastaldi I apply your suggestions then squashed the commits.
Should be OK to merge now :)

geoand:
@FroMage OK with the implementation?

extensions/panache/hibernate-orm-panache/runtime/src/main/java/io/quarkus/hibernate/orm/panache/PanacheRepositoryBase.java
```diff
@@ -634,6 +634,17 @@ public default long deleteAll() {
...
     }
 
+    /**
+     * Delete an entity of this type by ID.
+     *
+     * @param id the ID of the entity to delete.
+     * @return false if the entity was not deleted (not found).
+     */
+    @GenerateBridge
+    public default boolean deleteById(Object id) {
```
FroMage: Shouldn't this be of type `Id` for coherence with `findById(Id)` ?

extensions/panache/hibernate-orm-panache/runtime/src/main/java/io/quarkus/hibernate/orm/panache/runtime/JpaOperations.java
```diff
@@ -393,6 +393,16 @@ public static long deleteAll(Class<?> entityClass) {
         return (long) getEntityManager().createQuery("DELETE FROM " + getEntityName(entityClass)).executeUpdate();
     }
 
+    public static boolean deleteById(Class<?> entityClass, Object id) {
```
FroMage: I think we should start with the improved version ;)

extensions/panache/mongodb-panache/runtime/src/main/java/io/quarkus/mongodb/panache/PanacheMongoRepositoryBase.java
```diff
@@ -702,6 +702,17 @@ public default long deleteAll() {
...
     }
 
+    /**
+     * Delete an entity of this type by ID.
+     *
+     * @param id the ID of the entity to delete.
+     * @return false if the entity was not deleted (not found).
+     */
+    @GenerateBridge
+    public default boolean deleteById(Object id) {
```
FroMage: Same comment for `Id`

extensions/panache/mongodb-panache/runtime/src/main/java/io/quarkus/mongodb/panache/reactive/ReactivePanacheMongoRepositoryBase.java
```diff
@@ -703,6 +704,17 @@
...
     }
 
+    /**
+     * Delete an entity of this type by ID.
+     *
+     * @param id the ID of the entity to delete.
+     * @return false if the entity was not deleted (not found).
+     */
+    @GenerateBridge
+    public default Uni<Boolean> deleteById(Object id) {
```
FroMage: Same comment for the `Id` type.

extensions/panache/mongodb-panache/runtime/src/main/java/io/quarkus/mongodb/panache/reactive/runtime/ReactiveMongoOperations.java
```diff
@@ -551,6 +551,12 @@ private static Document sortToDocument(Sort sort) {
         return collection.deleteMany(new Document()).map(deleteResult -> deleteResult.getDeletedCount());
     }
 
+    public static Uni<Boolean> deleteById(Class<?> entityClass, Object id) {
+        ReactiveMongoCollection<?> collection = mongoCollection(entityClass);
+        Document query = new Document().append(ID, id);
+        return collection.deleteOne(query).map(results -> results.getDeletedCount() == 1);
```
FroMage: See? This one seems optimal, so let's not be unfavourable to ORM ;)

FroMage (changes requested):

extensions/panache/mongodb-panache/runtime/src/main/java/io/quarkus/mongodb/panache/PanacheMongoRepositoryBase.java
loicmathieu: It's not as easy as it looks, for `PanacheEntity` and for the repository pattern we don't know the name of the ID field so we need to retrieve it somehow ...

extensions/panache/hibernate-orm-panache/runtime/src/main/java/io/quarkus/hibernate/orm/panache/runtime/JpaOperations.java
loicmathieu: It's not as easy as it looks, for PanacheEntity and for the repository pattern we don't know the name of the ID field so we need to retrieve it somehow ...

extensions/panache/mongodb-panache/runtime/src/main/java/io/quarkus/mongodb/panache/reactive/runtime/ReactiveMongoOperations.java
loicmathieu: Yes because the ID field name is always `_id` in MongoDB so it's easy to implement ;)

extensions/panache/hibernate-orm-panache/runtime/src/main/java/io/quarkus/hibernate/orm/panache/runtime/JpaOperations.java
loicmathieu: I find a way to retrieve from Hiberate the name of the identifier field:
```
        EntityManager entityManager = getEntityManager();
        Session session = entityManager.unwrap(Session.class);
        SessionFactory sessionFactory = session.getSessionFactory();
        ClassMetadata metadata = sessionFactory.getClassMetadata(entityClass);
        String idField = metadata.getIdentifierPropertyName();
```
WDYT ?

extensions/panache/mongodb-panache/runtime/src/main/java/io/quarkus/mongodb/panache/PanacheMongoRepositoryBase.java
loicmathieu: done

extensions/panache/mongodb-panache/runtime/src/main/java/io/quarkus/mongodb/panache/reactive/ReactivePanacheMongoRepositoryBase.java
loicmathieu: done

extensions/panache/hibernate-orm-panache/runtime/src/main/java/io/quarkus/hibernate/orm/panache/PanacheRepositoryBase.java
loicmathieu: done

extensions/panache/hibernate-orm-panache/runtime/src/main/java/io/quarkus/hibernate/orm/panache/runtime/JpaOperations.java
loicmathieu: @FroMage I sucessfully create a query with the trick I described above. This couple Panache directly with Hibernate and not just JPA but I think it's OK. After all it's Hibernate with Panache not JPA with Panache :)

If the new impl is OK for you I'll squash it and we could merge it

FroMage: This looks good. Can you just check what happens with composite IDs? This should work with `@CompositeId` but not `@IdClass`, I assume.

But this should already be a limitation of `findById()` which only takes a single argument, so perhaps it's acceptable.

FroMage: https://www.baeldung.com/jpa-composite-primary-keys

loicmathieu: @FroMage I added tests with `@IdClass` and `@Embeddable` composite id and it works fine for `@Embeddable` (it generates a delete query with all the fields of the embeddable composite ID) but didn't works for `@IdClass`.

For `@IdClass` Hibernate didn't succeed in generating the query with the following error:
```
Caused by: java.lang.IllegalArgumentException: Could not locate ordinal parameter [1], expecting one of []
	at org.hibernate.query.internal.ParameterMetadataImpl.getOrdinalParameterDescriptor(ParameterMetadataImpl.java:154)
	at org.hibernate.query.internal.ParameterMetadataImpl.getQueryParameter(ParameterMetadataImpl.java:204)
	at org.hibernate.query.internal.QueryParameterBindingsImpl.getBinding(QueryParameterBindingsImpl.java:182)
	at org.hibernate.query.internal.AbstractProducedQuery.setParameter(AbstractProducedQuery.java:507)
	at org.hibernate.query.internal.AbstractProducedQuery.setParameter(AbstractProducedQuery.java:110)
	at io.quarkus.hibernate.orm.panache.runtime.JpaOperations.bindParameters(JpaOperations.java:95)
```

Multiple solutions here:
- It should works and I should log a bug on Hibernate side
- We go back to using `entityManager.find(id).remove()`
- We check if the object have the annotation `@IdClass` by reflection and switch back to `entityManager.find(id).remove()` only for this case (reflection lookup is fast so it should not be an issue)
- We add a varargs to deleteById so when using `@IdClass` based composite id the user can write `entity.deleteById(key.part1, key.part2)` instead of `entity.deleteById(key)` (not tested)

WDYT ?

FroMage: This stack trace is not for `findById` but for the new `deleteById` right?

loicmathieu: Yes, `findById works with composite id for both `@IdClass` and `@Embeddable`. Only `deleteById` with `@IdClass` didn't works

FroMage: I bet this is because we didn't find how to write a query for that.

Perhaps @Sanne knows of a way to do a delete by id in Hibernate that works without loading the entity, when using an `@IdClass`?

Sanne: "id" is a special attribute in HQL, you should be able to refer to it even if the field name is different.

Sanne: But give me some more context please; writing HQL queries for us to parse is typically not the most efficient approach.

FroMage: SERIOUSLY? I never knew that :(

The context is we're trying to implement `deleteById` without doing `findById(id).delete()`.

loicmathieu: @Sanne the purpose of the `deleteById` method is to ease the developement of classical REST endpoint where you want to delete an entity that are not yet loaded inside the persistence context. 

I first implement it via `findById(class, id).delete()` that basically maps to `entityManager.find(id, class); entityManager.remove(entity)`.

As this generates two queries, one SELECT and one DELETE we try to achieve a better implementation with only a DELETE query, something like : `entityManager.createQuery("delete from entity where id = ?1).setParameter(1, key).executeUpdate()`.

Keep in mind that it's inside Panache so the code is not exactly this one ...

It works, except for composite key with `@IdClass` (composite key with `@Embeddable` works fine) where it generates the exception above.

Sanne: you could do:

    Object toBeDeleted = session.byId( entityClass ).getReference( id );
    session.delete( toBeDeleted );

**However** this will not check for the object to actually exist, it will just do the delete statement.

Sanne: to clarify: the first line returns the representation of the object to be deleted in the form of a managed entity.
So if that same object was already loaded in some way, you'll get that one instance returned with the state it currently has. If there was no such object of that type and Id in the current Session, it's NOT loading it from the database but returning you a managed instance which only has the Id initialized.

- 

-

Sanne: thanks @loicmathieu , for some reason I only received your comment after posting my suggestion.
The suggestion doesn't change, just please think if the fact that you might be deleting an entity which actually no longer exists is something that you'd want an error for... I'd say probably not but I'm not 100% sure.

loicmathieu: @Sanne thanks a lot this seems exactly what we need ;)
We return a boolean at true if the entity was removed, I'll test it tomorrow.

Sanne: > We return a boolean at true if the entity was removed, I'll test it tomorrow.

I agree that returning this information to the user is kinda nice, but it might be very complicated to implement: Hibernate ORM won't tell you: if you have a reference to a persistent object then it will existin the DB, since we assume transactionality.

To be clear, it's not as simple as checking for the expected number of rows: it's an object so you're likely deleting what is actually a graph of entries from related tables, and this is most likely done via a combination of cascading via foreign constraints and/or batching of multiple statements into a single roundtrip.
It's very likely that it's actually cheaper to load the object first, if you need to know of its existence.

I would suggest to return `void`; if someone really needs to know, they'll always have the option to load /check explicitly. Sorry, I know it's not ideal...

loicmathieu: @Sanne `getReference()` performs a select in case the entity is not yet inside the persistence context and throws `EntityNotFoundException` if it didn't exist.

So we are back with a select and a delete if we use it (it didn't perform the select if the entity is already loaded inside the persistence context, but in the context of a DELETE Rest endpoint it should not be the case).

For the records, the code I tested is :
```
Object entity = getEntityManager().getReference(entityClass, id);
getEntityManager().remove(entity);
```

loicmathieu: @Sanne @FroMage here is what I propose:

- Implement a direct deleteByIf with HQL as previously if the identifier entity is not annotated with `@IdClass`
- Implements a deleteById via getReference + remove for entity annotated by `@IdClass`
- Document clearly the method: it should be used when the entity is not yet loaded inside the persistence context (like for stateless DELETE rest endoint), if the entity is already loaded calling `delete()` on the entity performs better.

loicmathieu:
@Sanne @FroMage I review the implementation to use an JPQL query where possible and switch back to using find()/remove() in case the entity is annotated by `@IdClass`. We cannot use `getReference()` as it throws an exception if the entity cannot be foud.

I also add some note on the JavaDoc to advise to use remove() if the entity is already loaded inside the persistence contexte, as I understand it this is when remove() is more performant that an JPQL query right ?

If it's OK, I'll squash the commit and made a last review for doc/typos/...

extensions/panache/hibernate-orm-panache/runtime/src/main/java/io/quarkus/hibernate/orm/panache/runtime/JpaOperations.java
Sanne: > @Sanne `getReference()` performs a select in case the entity is not yet inside the persistence context and throws `EntityNotFoundException` if it didn't exist.

Hi @loicmathieu , no that's not accurate. `getReference()` will not hit the database at all.

But like I warned, it will assume that you know for sure that the entity exists: it's not correct to use this method if you're in doubt.
I guess what you observe is that some subsequent operation later on is forcing the reference to be loaded - most likely the delete action itself.

TBH as I suggested previous I don't think you can implement this feature to be all three of:
 - generic enough for any model
 - correct in all such cases
 - skipping the load

It's one of those situations in which you have to pick 2.

loicmathieu: > it will assume that you know for sure that the entity exists: it's not correct to use this method if you're in doubt.

Yes, in fact this is my assumption also, I write my comment too quickly.

OK, so you advise to stick to `find(id) + remove()`, as it's @FroMage that ask for a JPQL implementation I would like to see if he is on the same page (I'm pretty sure yes but there is no hurry)

FroMage: If the Hibernate experts tell us to use one strategy, I won't object. Just make sure to add a comment to the impl explaining why we use what appears to be a supoptimal strategy, so that no-one wants to "improve" it later.

Sanne: yeah. We could explore some optimisations in Hibernate ORM; for example I was chatting with the team about this yesterday as I wondered why we don't do just the update statement at least for the simple cases in which we can proof there is no cascading requirements.

Turns out the team sentiment is that while that would be possible, it's actually highly unlikely for an entity to satisfy such requirements in more realistic models than our simple demos, and  possibly not trivial to do.. so that explains why it doesn't do it but there certainly is the option to eventually improve on that.

Sanne: BTW another angle to this is that even if we *did* (hypotethically speaking) implement such an optimisation in ORM, you'd have no way to return true/false about knowing if the specific entity was actually existing before... so probably really not worth it.

loicmathieu:
@Sanne @FroMage I switch back to find()/remove() for Hibernate.
I add an implementation note to the implementation inside `JpaOperations`.

I also add the documentation part inside the guides.

docs/src/main/asciidoc/hibernate-orm-panache.adoc
```diff
@@ -195,6 +195,9 @@ Person.delete("status", Status.Alive);
 // delete all persons
 Person.deleteAll();
 
+// delete by id
+boolean deleted = Person.deleteByIf(personId);
```
FroMage:
```suggestion
boolean deleted = Person.deleteById(personId);
```

docs/src/main/asciidoc/hibernate-orm-panache.adoc
```diff
@@ -379,6 +382,9 @@ personRepository.delete("status", Status.Alive);
 // delete all persons
 personRepository.deleteAll();
 
+// delete by id
+boolean deleted = personRepository.deleteByIf(personId);
```
FroMage:
```suggestion
boolean deleted = personRepository.deleteById(personId);
```

docs/src/main/asciidoc/mongodb-panache.adoc
```diff
@@ -217,6 +217,9 @@ Person.delete("status", Status.Alive);
 
 // delete all persons
 Person.deleteAll();
+
+// delete by id
+boolean deleted = Person.deleteByIf(personId);
```
FroMage:
```suggestion
boolean deleted = Person.deleteById(personId);
```

docs/src/main/asciidoc/mongodb-panache.adoc
```diff
@@ -378,6 +381,9 @@ personRepository.delete("status", Status.Alive);
 
 // delete all persons
 personRepository.deleteAll();
+
+// delete by id
+boolean deleted = personRepository.deleteByIf(personId);
```
FroMage:
```suggestion
boolean deleted = personRepository.deleteById(personId);
```

docs/src/main/asciidoc/mongodb-panache.adoc
```diff
@@ -771,6 +777,9 @@ Uni<Long>  deleteCount = ReactivePerson.delete("status", Status.Alive);
 
 // delete all persons
 deleteCount = ReactivePerson.deleteAll();
+
+// delete by id
+Uni<Boolean> deleted = ReactivePerson.deleteByIf(personId);
```
FroMage:
```suggestion
Uni<Boolean> deleted = ReactivePerson.deleteById(personId);
```

FroMage (changes requested):
Aside from the typo, LGTM.

loicmathieu:
@FroMage sorry, I double check the documentation but the typo slipped throuhg.

Fixed.

FroMage (approved):
Great!

loicmathieu:
@FroMage @gastaldi can we merge it now ?

PR #7533, opened 2020-03-03
Fix #7532 Introduce NativeConfig.includeAllTimeZones

### LINKED ISSUE #7532: Introduce NativeConfig.includeAllTimeZones
Could we please introduce a config switch for passing `-H:+IncludeAllTimeZones` to the native-image?

The feature would allow users to enable all time zones in the native mode via `application.propreties`:

```
quarkus-core_quarkus.native.include-all-time-zones = true
```

This could work in the very same way as `quarkus-core_quarkus.native.add-all-charsets` https://quarkus.io/guides/all-config#quarkus-core_quarkus.native.add-all-charsets

I can prepare a PR if there is an interest.

ppalaga:


core/deployment/src/main/java/io/quarkus/deployment/pkg/NativeConfig.java
```diff
@@ -48,6 +48,12 @@
     @ConfigItem(defaultValue = "false")
     public boolean addAllCharsets;
 
+    /**
+     * If all time zones should be added to the native image. This increases image size
+     */
+    @ConfigItem(defaultValue = "false")
+    public boolean includeAllTimeZones;
```
gsmet: Could we be consistent for the two config properties for charsets and time zones?

gsmet (changes requested):

core/deployment/src/main/java/io/quarkus/deployment/pkg/NativeConfig.java
ppalaga: Well I see two ways how to be consistent: (a) align with the native-image flag (my current proposal) or (b) align with other available properties.

(a) Might be less confusing for the people knowing the native-image flags.

I have no clear preference and I am ready to go with `all` instead of `include` if you and others prefer that.

gsmet: Yeah no OK, it makes perfect sense.

gsmet (approved):

PR #7538, opened 2020-03-03
Update the JMS guide to use the Qpid JMS extension

gemmellr:
The changes in the PR update the JMS guide as previously discussed, to use the Qpid JMS client extension (related PR for adding it to the platform: https://github.com/quarkusio/quarkus-platform/pull/28), while also detailing the small dependency and config differences for using the Artemis JMS client extension instead.

The project creation step depends on the extension being in the platform. If that would prevent merging for now, I could change that so it doesnt (e.g detail manual addition of the dependency to the pom) for now and then change it back again later.

### DIFF
--- a/docs/src/main/asciidoc/jms.adoc
+++ b/docs/src/main/asciidoc/jms.adoc
@@ -3,11 +3,13 @@ This guide is maintained in the main Quarkus repository
 and pull requests should be submitted there:
 https://github.com/quarkusio/quarkus/tree/master/docs/src/main/asciidoc
 ////
-= Quarkus - Using Artemis JMS extension
+= Quarkus - Using JMS
 include::./attributes.adoc[]
 :extension-status: preview
 
-This guide demonstrates how your Quarkus application can use Artemis JMS messaging.
+
+This guide demonstrates how your Quarkus application can use JMS messaging via the
+Apache Qpid JMS AMQP client, or alternatively the Apache ActiveMQ Artemis JMS client.
 
 include::./status-include.adoc[]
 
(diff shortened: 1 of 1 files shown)


gemmellr:
Rebased to fix conflicts from some of the same changes subsequently being made in #7590

gemmellr:
@emmanuelbernard @gsmet @aloubyansky can you look over these guide changes?

docs/src/main/asciidoc/jms.adoc
```diff
@@ -19,26 +21,40 @@ To complete this guide, you need:
 * an IDE
 * JDK 1.8+ installed with `JAVA_HOME` configured appropriately
 * Apache Maven 3.5.3+
-* A running Artemis server, or Docker Compose to start one
-* GraalVM installed if you want to run in native mode.
+* A running Artemis server, or Docker to start one
+* GraalVM, or Docker, installed if you want to run in native mode.
```
gsmet: The two `or Docker` are weird. This should be reworded.

docs/src/main/asciidoc/jms.adoc
```diff
@@ -19,26 +21,40 @@ To complete this guide, you need:
...
-These prices are written in an JMS queue (`prices`).
-Another component reads from the `prices` queue and stores the last price.
-The data can be fetch from a browser using a fetch button from a JAX-RS resource.
+These prices are written to a queue (`prices`) using a JMS client.
+Another component reads from the `prices` queue and stores the latest price.
+The data can be fetched from a browser using a fetch button from a JAX-RS resource.
+
+
+The guide can be used either via the Apache Qpid JMS AMQP client as detailed immediately below, or
+alternatively with the Apache ActiveMQ Artemis JMS client given some some different configuration
```
gsmet:
```suggestion
alternatively with the Apache ActiveMQ Artemis JMS client given some different configuration
```

docs/src/main/asciidoc/jms.adoc
```diff
@@ -290,6 +319,83 @@ You can build the native executable with:
 ./mvnw package -Pnative
 ----
 
-== Configuration Reference
+Or, if you dont have GraalVM installed, you can instead use Docker to build the native executable using:
```
gsmet:
```suggestion
Or, if you don't have GraalVM installed, you can instead use Docker to build the native executable using:
```

docs/src/main/asciidoc/jms.adoc
```diff
@@ -290,6 +319,83 @@ You can build the native executable with:
...
+and then run with:
+
+[source, shell]
+----
+./target/jms-quickstart-1.0-SNAPSHOT-runner
+----
+
+Open `http://localhost:8080/prices.html` in your browser.
+
+'''
```
gsmet: What is this?

gsmet (changes requested):
Looks mostly good, I added a few suggestions.

docs/src/main/asciidoc/jms.adoc
gemmellr: Any recommendations? I couldn't really think of a better wording as they are really separate things. You dont need Docker for either requirement, but you could use it for one or both of them. Other guides seem to just ignore that you can use docker for native builds and so you dont need GraalVM installed.

docs/src/main/asciidoc/jms.adoc
gemmellr: Oops, will fix.

docs/src/main/asciidoc/jms.adoc
gemmellr: A line separator between the two outer client sections to make it a little more obvious this is the end of the steps.

docs/src/main/asciidoc/jms.adoc
gsmet: Ah no, it's OK, I somehow missed it was a bullet point.

docs/src/main/asciidoc/jms.adoc
gemmellr: Will fix.

gsmet:
OK, I let you fix the two typos and we're done then. Thanks!

gemmellr:
Rebased and pushed with the 2 fixes.

gsmet (approved):

gsmet:
Merged, thanks!

PR #7540, opened 2020-03-03
refactor: do not use deprecated DefaultAnnotationScannerExtension class

machi1990:


EricWittmann:
Seems reasonable to me - what do you think @MikeEdgar ?

gsmet:
Yeah the thing worrying me a bit is the `return null` appearing now.

machi1990:
> Yeah the thing worrying me a bit is the `return null` appearing now.

This seems to be the default value returned by `AnnotationScannerExtension#resolveAsyncType()`.

MikeEdgar:
+1

Returning `null` is valid - it simply means the `Type` could not be resolved. As @machi1990 mentioned, it's the same value being returned by the call of the `super` method today.

gsmet (approved):
Got the approval from the SmallRye OpenAPI team so let's merge!

gsmet:
Thanks for the reviews!

Issue #7269, opened 2020-02-18
Support both GraalVM 19.3.1 and 20.0.0


gwenneg:
As mentioned by @gastaldi on [Zulip](https://quarkusio.zulipchat.com/#narrow/stream/187038-dev/topic/GraalVM.2020), GraalVM `20.0` is now available: https://www.graalvm.org/docs/release-notes/20_0/

@cescoffier We'll need two new `ubi-quarkus-native-image` (JDK 8 and 11) for that version, but I suppose this has a lower priority than the Mutiny integration since `1.3.0.CR1` is getting close.

gwenneg:
I'll start running locally the complete Quarkus build in native mode with GraalVM `20.0` tonight to identify possible issues ASAP (unless the ubi images are ready by then).

gsmet:
FYI https://github.com/quarkusio/quarkus-images/pull/55

iwangxiaodong:
**I use GraalVM 20.0.0 + io.quarkus:quarkus-bom:1.3.0.Alpha2, Output error:**

  Error: com.oracle.graal.pointsto.constraints.UnsupportedFeatureException: type is not available in this platform: io.quarkus.runtime.graal.DiagnosticPrinter$Target_PosixJavaThreads
  Trace:
          at parsing io.quarkus.runtime.graal.DiagnosticPrinter.printDiagnostics(DiagnosticPrinter.java:46)
  Call path from entry point to io.quarkus.runtime.graal.DiagnosticPrinter.printDiagnostics(PrintStream):
          at io.quarkus.runtime.graal.DiagnosticPrinter.printDiagnostics(DiagnosticPrinter.java:29)
          at io.quarkus.runtime.Application$2.handle(Application.java:209)
          at sun.misc.Signal$InternalMiscHandler.handle(Signal.java:198)
          at jdk.internal.misc.Signal$1.run(Signal.java:220)
          at java.lang.Thread.run(Thread.java:834)
          at com.oracle.svm.core.thread.JavaThreads.threadStartRoutine(JavaThreads.java:527)
          at com.oracle.svm.core.windows.WindowsJavaThreads.osThreadStartRoutine(WindowsJavaThreads.java:138)
          at com.oracle.svm.core.code.IsolateEnterStub.WindowsJavaThreads_osThreadStartRoutine_4bc03aa26f8cdfc97ebd54050e8ae4bce1023851(generated:0)
[...]

PR #7546, opened 2020-03-03
Make DiagnosticPrinter compatible with Windows

gwenneg:
This PR fixes the following error which was reported in #7269. It happens while running Quarkus with GraalVM 20.0.0 on Windows.

```
Error: com.oracle.graal.pointsto.constraints.UnsupportedFeatureException: type is not available in this platform: io.quarkus.runtime.graal.DiagnosticPrinter$Target_PosixJavaThreads
Trace:
        at parsing io.quarkus.runtime.graal.DiagnosticPrinter.printDiagnostics(DiagnosticPrinter.java:46)
Call path from entry point to io.quarkus.runtime.graal.DiagnosticPrinter.printDiagnostics(PrintStream):
        at io.quarkus.runtime.graal.DiagnosticPrinter.printDiagnostics(DiagnosticPrinter.java:29)
        at io.quarkus.runtime.Application$2.handle(Application.java:209)
        at sun.misc.Signal$InternalMiscHandler.handle(Signal.java:198)
        at jdk.internal.misc.Signal$1.run(Signal.java:220)
        at java.lang.Thread.run(Thread.java:834)
        at com.oracle.svm.core.thread.JavaThreads.threadStartRoutine(JavaThreads.java:527)
        at com.oracle.svm.core.windows.WindowsJavaThreads.osThreadStartRoutine(WindowsJavaThreads.java:138)
        at com.oracle.svm.core.code.IsolateEnterStub.WindowsJavaThreads_osThreadStartRoutine_4bc03aa26f8cdfc97ebd54050e8ae4bce1023851(generated:0)
```
It is caused by the following Quarkus code:

https://github.com/quarkusio/quarkus/blob/3fbdbc399f4c274054bed23d70061c028e7a59c7/core/runtime/src/main/java/io/quarkus/runtime/graal/DiagnosticPrinter.java#L46-L50

@dmlloyd I'll need your opinion on this fix since that's the first time I'm doing that kind of change.

It was successfully tested locally with GraalVM 20.0.0 (JDK 11 edition) and Windows 10.

### DIFF
--- a/core/runtime/src/main/java/io/quarkus/runtime/graal/DiagnosticPrinter.java
+++ b/core/runtime/src/main/java/io/quarkus/runtime/graal/DiagnosticPrinter.java
@@ -9,6 +9,7 @@ import org.graalvm.nativeimage.Platforms;
 
 import com.oracle.svm.core.annotate.Alias;
 import com.oracle.svm.core.annotate.TargetClass;
+import com.oracle.svm.core.os.IsDefined;
 import com.oracle.svm.core.posix.headers.Pthread;
 
 /**
@@ -43,7 +44,7 @@ public final class DiagnosticPrinter {
             w.print("prio=");
             w.print(thread.getPriority());
             w.print(" tid=");
-            if (Target_PosixJavaThreads.hasThreadIdentifier(thread)) {
+            if ((IsDefined.isLinux() || IsDefined.isDarwin()) && Target_PosixJavaThreads.hasThreadIdentifier(thread)) {
                 final long nativeId = Target_PosixJavaThreads.getPthreadIdentifier(thread).rawValue();
                 w.print("0x");
                 w.println(Long.toHexString(nativeId));


core/runtime/src/main/java/io/quarkus/runtime/graal/WindowsDiagnosticPrinter.java
```diff
@@ -0,0 +1,37 @@
...
+@Platforms(WINDOWS.class)
+public final class WindowsDiagnosticPrinter extends DiagnosticPrinter {
+
+    @Platforms(HOSTED_ONLY.class)
+    WindowsDiagnosticPrinter() {
+    }
+
+    @Override
+    protected void printTid(PrintStream w, Thread thread) {
+        w.println(" tid=(unknown)");
```
gwenneg: I'm not satisfied with this, but I didn't find a way in [WindowsJavaThreads](https://github.com/oracle/graal/blob/master/substratevm/src/com.oracle.svm.core.windows/src/com/oracle/svm/core/windows/WindowsJavaThreads.java) to retrieve the Windows native thread ID. Suggestions are welcome here.

nimo23: Look at this gist from https://gist.github.com/rednaxelafx/843622 to access the Native thread ID: If `sun.jvm.hotspot.tools.Tool` is available in Quarkus you can get it with `ThreadInfo` for any platform:

```
        List<ThreadInfo> infos = ThreadUtils.getAllThreadInfos();
        for (ThreadInfo info : infos) {
            System.out.printf("Thread@%s: tid=%d nid=0x%x\n",
            	info.getAddress(), info.getTid(), info.getNid());
        }
```

You can extend `Tool` within an inner static class in  `WindowsJavaThreads`..

core/runtime/src/main/java/io/quarkus/runtime/graal/DiagnosticPrinter.java
```diff
@@ -70,4 +44,6 @@ public static void printDiagnostics(PrintStream w) {
             }
         }
     }
+
+    protected abstract void printTid(PrintStream w, Thread thread);
```
dmlloyd: This solution is likely far more complex than necessary.  Before we go to this extreme, could you try wrapping the POSIX-specific code in a call to `com.oracle.svm.core.os.IsDefined#isLinux`/`#isDarwin`?  These values are constant folded and thus allow conditional compilation.

dmlloyd (changes requested):

core/runtime/src/main/java/io/quarkus/runtime/graal/DiagnosticPrinter.java
dmlloyd: You might need to move the platform-specific code to methods.  I've had bugs where it didn't work otherwise, though the GraalVM people couldn't explain why.

dmlloyd: You could probably have still used the `OS.getCurrent() != OS.WINDOWS` strategy if you put that conditional into a method with the `@Fold` annotation.  However the problem is that there is no artifact containing this annotation, so there's really no way to add it in real code!

gwenneg: I just tested `IsDefined` and it does solve the issue in a much simpler way. Thanks for the suggestion @dmlloyd!

There's only one drawback to it: we can't use `IsDefined` to include some Windows-specific code since there's no `IsDefined.isWindows()`, unless we can assume that any OS that is not Linux or Darwin is a Windows OS (which does not sound reasonnable to me).

If we want to retrieve and print the Windows native thread ID, then the complex version might be the only way to do it. If not, the `IsDefined` is definitely the best one.

WDYT?

dmlloyd: I think that for the time being it'd be OK to assume that `windows == ! linux && ! macos`.  It can always be revisited later, and/or an issue can be opened upstream to request an `IsWindows` method.

One option could be to introduce a utility method somewhere, marked `@AlwaysInline`, which does this logic so that if/when there's a better way to identify Windows it can be updated in just one place.  But I wouldn't worry too much about it otherwise.

dmlloyd (approved):

gwenneg:
Thanks for the review and suggestions. I just squashed the commits.

Let's keep this PR simple since it's meant to fix a bug and revisit the code later if needed.

machi1990 (approved):

gsmet:
@gwenneg can you open an issue upstream requesting `IsWindows`? Better have it next time we need it.

PR #7552, opened 2020-03-04
Rest Client - support injection for providers used in @RegisterProvider

### LINKED ISSUE #7531: Microprofile rest client - Can't inject config/beans to providers also in 1.2.1.Final
**Describe the bug**
As already described in #5752, it is not possible to inject CDI beans into any kind of providers that are registered by @RegisterProvider or @RegisterClientHeaders. The upgrade of quarkus to 1.2.1.Final and resteasy 4.4.2 seems not to solve this issue.

**To Reproduce**
1. Clone project from marcinczeczko from [here](https://github.com/marcinczeczko/quarkus-rest-providers-reproducer)
2. Upgrade Quarkus versions in pom.xml to 1.2.1.Final (Plugin + universe-bom)
3. Start project and do GET request to echo resource
4. Output on console is null for every injection

mkouba:
- i.e. for providers not annotated with @Provider
- resolves #7531
- follows up on #5752

### DIFF
--- a/extensions/rest-client/deployment/pom.xml
+++ b/extensions/rest-client/deployment/pom.xml
@@ -41,6 +41,11 @@
             <artifactId>quarkus-smallrye-fault-tolerance-deployment</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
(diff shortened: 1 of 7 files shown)


geoand (approved):
Nice one!

gsmet (approved):

PR #7560, opened 2020-03-04
Qute - loop section -support iterators and arrays

mkouba:


gastaldi (approved):
Looks good!

independent-projects/qute/core/src/main/java/io/quarkus/qute/LoopSectionHelper.java
```diff
@@ -40,12 +41,16 @@
...
                 iterator = ((Iterable<?>) it).iterator();
+            } else if (it instanceof Iterator) {
+                iterator = (Iterator<?>) it;
             } else if (it instanceof Map) {
                 iterator = ((Map<?, ?>) it).entrySet().iterator();
             } else if (it instanceof Stream) {
                 iterator = ((Stream<?>) it).sequential().iterator();
             } else if (it instanceof Integer) {
                 iterator = IntStream.rangeClosed(1, (Integer) it).iterator();
+            } else if (it != null && it.getClass().isArray()) {
```
gastaldi: No biggie but maybe the null check could be the first statement to check in this `if-else` chain

mkouba: Well, `instanceof` handles `null` values so I don't think it would help in any way...  also the `IllegalStateException` should be the same for `null` and types that are not considered iterable. WDYT?

mkouba: Hm, I can try to improve the readability a little bit..

PR #7562, opened 2020-03-04
devtools commands to accept platform descriptor as an argument

aloubyansky:
This commit completes the refactoring of project manipulating commands to allow passing QuarkusPlatformDescriptor as an argument instead of relying on a globally shared (using a static or thread-local var) instance of it. It also introduces thread-safe handlers for each command. The legacy commands are now collecting the input into an instance of QuarkusCommandInvocation and simply pass it to the corresponding command handler for execution.


maxandersen (approved):
Looks good! just internal changes to allow for more flexible platform definitions/lookup.

PR #7568, opened 2020-03-04
added possibility to add custom named handlers

hartimcwildfly:
Current master does not build on my local environment. So I have to misuse the CI.

### DIFF
--- a/core/deployment/src/main/java/io/quarkus/deployment/builditem/NamedLogHandlersBuildItem.java
+++ b/core/deployment/src/main/java/io/quarkus/deployment/builditem/NamedLogHandlersBuildItem.java
@@ -1,0 +1,34 @@
+package io.quarkus.deployment.builditem;
+
+import java.util.Map;
+import java.util.logging.Handler;
+
+import org.wildfly.common.Assert;
+
+import io.quarkus.builder.item.MultiBuildItem;
+import io.quarkus.runtime.RuntimeValue;
+
+/**
+ * A build item for adding additional named logging handlers.
+ */
+public final class NamedLogHandlersBuildItem extends MultiBuildItem {
+    private final RuntimeValue<Map<String, Handler>> namedHandlersMap;
+
+    /**
+     * Construct a new instance.
+     *
+     * @param namedHandlersMap the named handlers to add to the run time configuration
+     */
+    public NamedLogHandlersBuildItem(final RuntimeValue<Map<String, Handler>> namedHandlersMap) {
+        this.namedHandlersMap = Assert.checkNotNullParam("namedHandlersMap", namedHandlersMap);
+    }
+
+    /**
+     * Get the named handlers.
+     *
+     * @return the named handlers map
+     */
+    public RuntimeValue<Map<String, Handler>> getNamedHandlersMap() {
+        return namedHandlersMap;
+    }
+}
--- a/core/deployment/src/main/java/io/quarkus/deployment/logging/LoggingResourceProcessor.java
+++ b/core/deployment/src/main/java/io/quarkus/deployment/logging/LoggingResourceProcessor.java
@@ -1,6 +1,7 @@
 package io.quarkus.deployment.logging;
 
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 import java.util.function.Consumer;
 import java.util.function.Supplier;
@@ -17,6 +18,7 @@ import io.quarkus.deployment.builditem.ConsoleFormatterBannerBuildItem;
 import io.quarkus.deployment.builditem.LogCategoryBuildItem;
 import io.quarkus.deployment.builditem.LogConsoleFormatBuildItem;
 import io.quarkus.deployment.builditem.LogHandlerBuildItem;
+import io.quarkus.deployment.builditem.NamedLogHandlersBuildItem;
 import io.quarkus.deployment.builditem.RunTimeConfigurationDefaultBuildItem;
 import io.quarkus.deployment.builditem.SystemPropertyBuildItem;
 import io.quarkus.deployment.builditem.nativeimage.NativeImageSystemPropertyBuildItem;
@@ -80,11 +82,14 @@ public final class LoggingResourceProcessor {
 
     @BuildStep
     @Record(ExecutionTime.RUNTIME_INIT)
-    void setupLoggingRuntimeInit(LoggingSetupRecorder recorder, LogConfig log, List<LogHandlerBuildItem> handlers,
-            List<LogConsoleFormatBuildItem> consoleFormatItems,
+    void setupLoggingRuntimeInit(LoggingSetupRecorder recorder, LogConfig log, List<LogHandlerBuildItem> handlerBuildItems,
+            List<NamedLogHandlersBuildItem> namedHandlerBuildItems, List<LogConsoleFormatBuildItem> consoleFormatItems,
             Optional<ConsoleFormatterBannerBuildItem> possibleBannerBuildItem) {
-        final List<RuntimeValue<Optional<Handler>>> list = handlers.stream().map(LogHandlerBuildItem::getHandlerValue)
+        final List<RuntimeValue<Optional<Handler>>> handlers = handlerBuildItems.stream()
+                .map(LogHandlerBuildItem::getHandlerValue)
                 .collect(Collectors.toList());
+        final List<RuntimeValue<Map<String, Handler>>> namedHandlers = namedHandlerBuildItems.stream()
+                .map(NamedLogHandlersBuildItem::getNamedHandlersMap).collect(Collectors.toList());
 
         ConsoleFormatterBannerBuildItem bannerBuildItem = null;
         RuntimeValue<Optional<Supplier<String>>> possibleSupplier = null;
@@ -94,7 +99,7 @@ public final class LoggingResourceProcessor {
         if (bannerBuildItem != null) {
             possibleSupplier = bannerBuildItem.getBannerSupplier();
         }
-        recorder.initializeLogging(log, list,
+        recorder.initializeLogging(log, handlers, namedHandlers,
                 consoleFormatItems.stream().map(LogConsoleFormatBuildItem::getFormatterValue).collect(Collectors.toList()),
                 possibleSupplier);
     }
--- a/core/runtime/src/main/java/io/quarkus/runtime/logging/LoggingSetupRecorder.java
+++ b/core/runtime/src/main/java/io/quarkus/runtime/logging/LoggingSetupRecorder.java
@@ -18,6 +18,7 @@ import java.util.logging.ErrorManager;
 import java.util.logging.Formatter;
 import java.util.logging.Handler;
 import java.util.logging.Level;
+import java.util.stream.Collectors;
 
 import org.graalvm.nativeimage.ImageInfo;
 import org.jboss.logmanager.EmbeddedConfigurator;
@@ -73,11 +74,12 @@ public class LoggingSetupRecorder {
     public static void handleFailedStart() {
         LogConfig config = new LogConfig();
         ConfigInstantiator.handleObject(config);
-        new LoggingSetupRecorder().initializeLogging(config, Collections.emptyList(),
+        new LoggingSetupRecorder().initializeLogging(config, Collections.emptyList(), Collections.emptyList(),
                 Collections.emptyList(), null);
     }
 
     public void initializeLogging(LogConfig config, final List<RuntimeValue<Optional<Handler>>> additionalHandlers,
+            final List<RuntimeValue<Map<String, Handler>>> additionalNamedHandlers,
             final List<RuntimeValue<Optional<Formatter>>> possibleFormatters,
             final RuntimeValue<Optional<Supplier<String>>> possibleBannerSupplier) {
 
@@ -115,6 +117,16 @@ public class LoggingSetupRecorder {
         }
 
         Map<String, Handler> namedHandlers = createNamedHandlers(config, possibleFormatters, errorManager, filterElements);
+
+        Map<String, Handler> additionalNamedHandlersMap = additionalNamedHandlers.stream().map(RuntimeValue::getValue)
+                .flatMap(map -> map.entrySet().stream())
+                .collect(Collectors.toMap(Entry::getKey, Entry::getValue));
+
+        for (Handler additionalNamedHandler : additionalNamedHandlersMap.values()) {
+            additionalNamedHandler.setErrorManager(errorManager);
+            additionalNamedHandler.setFilter(new LogCleanupFilter(filterElements));
+        }
+        namedHandlers.putAll(additionalNamedHandlersMap);
 
         for (Map.Entry<String, CategoryConfig> entry : categories.entrySet()) {
             final String name = entry.getKey();


geoand:
Please don't misuse CI since we have limited resources.

You can close the pull request and a open a new draft one that will run the tests in the Github Actions of your own fork.
See https://github.com/quarkusio/quarkus/blob/master/CONTRIBUTING.md#continuous-integration

hartimcwildfly:
Sorry. This should be the last time. It should build now.

geoand:
OK :)

core/deployment/src/main/java/io/quarkus/deployment/builditem/NamedLogHandlersBuildItem.java
```diff
@@ -0,0 +1,34 @@
...
+import org.wildfly.common.Assert;
+
+import io.quarkus.builder.item.MultiBuildItem;
+import io.quarkus.runtime.RuntimeValue;
+
+/**
+ * A build item for adding additional named logging handlers.
+ */
+public final class NamedLogHandlersBuildItem extends MultiBuildItem {
+    private final RuntimeValue<Map<String, Handler>> namedHandlersValue;
```
dmlloyd: Why have a multi build item that in turn contains a map of multiple values, instead of having the build item contain only `String name` and `RuntimeValue<Optional<Handler>>` values?  (Apologies if this conversation already happened and I just don't remember).

hartimcwildfly: First I tried using multiple build items with RuntimeValue<Map.Entry<String, Handler>>. But then I have to collect all of the entries to a single map which got me an exception like (I'm not sure but something like) "cannot invoke method getValue() on ProxyObject NamedLogHandlersBuildItem, it is only allowed to pass it to a recorder."
I have one last idea how to do this better. Let me check it out. Then I will comment again.

hartimcwildfly:
So I tried it with RuntimeValue<Map.Entry<String, Handler>> but I could not find another solution.
Problems are:

Doing it like case 1 always results in a NPE when accessing rlogConfig.handlerConfigMap.entrySet()

case 1:

`class RlogProcessor {

    private static final String FEATURE = "rlog";

    @BuildStep
    FeatureBuildItem feature() {
        return new FeatureBuildItem(FEATURE);
    }

    @BuildStep
    @Record(ExecutionTime.RUNTIME_INIT)
    List<NamedLogHandlersBuildItem> addRLogHandlers(final RlogConfig rlogConfig,
            RlogHandlerValueFactory rlogHandlerValueFactory) {
        return rlogConfig.handlerConfigMap.entrySet().stream()
                .map(entry -> rlogHandlerValueFactory.createNamedHandler(entry.getKey(), entry.getValue()))
                .map(NamedLogHandlersBuildItem::new).collect(Collectors.toList());
    }
}

@Recorder
public class RlogHandlerValueFactory {

    private static final Logger LOG = Logger.getLogger(RlogHandlerValueFactory.class);

    private Handler createHandler(HandlerConfig handlerConfig) {
        RlogHandler handler = new RlogHandler();
        handler.setLevel(handlerConfig.level);
        handler.setFilename(handlerConfig.filename);
        handler.setPort(handlerConfig.port);
        handler.setStandalone(handlerConfig.standalone);
        handler.setLocalDir(handlerConfig.localDir);
        handler.setFormatter(new SimpleFormatter());

        return handler;
    }
}`

When returning List<RuntimeValue<Map.Entry<String, Handler>> and then wrap each of it in separate BuildItem like in case2.
I get the exception: "java.lang.RuntimeException: You cannot invoke stream() directly on an object returned from the bytecode recorder, you can only pass it back into the recorder as a parameter"

case 2:

`class RlogProcessor {

    private static final String FEATURE = "rlog";

    @BuildStep
    FeatureBuildItem feature() {
        return new FeatureBuildItem(FEATURE);
    }

    @BuildStep
    @Record(ExecutionTime.RUNTIME_INIT)
    List<NamedLogHandlersBuildItem> addRLogHandlers(final RlogConfig rlogConfig,
            RlogHandlerValueFactory rlogHandlerValueFactory) {
        return rlogHandlerValueFactory.createNamedHandler(rlogConfig).stream().map(NamedLogHandlersBuildItem::new).collect(Collectors.toList());
    }
}

@Recorder
public class RlogHandlerValueFactory {
    private static final Logger LOG = Logger.getLogger(RlogHandlerValueFactory.class);

    public List<RuntimeValue<Map.Entry<String, Handler>>> createNamedHandler(RlogConfig rlogConfig) {
        return rlogConfig.handlerConfigMap.entrySet().stream().collect(Collectors.toMap(Map.Entry::getKey, entry -> createHandler(entry.getValue()))).entrySet().stream()
                .map(RuntimeValue::new).collect(Collectors.toList());
    }

    private Handler createHandler(HandlerConfig handlerConfig) {
        RlogHandler handler = new RlogHandler();
        handler.setLevel(handlerConfig.level);
        handler.setFilename(handlerConfig.filename);
        handler.setPort(handlerConfig.port);
        handler.setStandalone(handlerConfig.standalone);
        handler.setLocalDir(handlerConfig.localDir);
        handler.setFormatter(new SimpleFormatter());

        return handler;
    }
}`

So IMHO the solution with RuntimeValue<Map<String, Handler>> is the only one which works.

core/deployment/src/main/java/io/quarkus/deployment/builditem/NamedLogHandlersBuildItem.java
hartimcwildfly: So I tried it with RuntimeValue<Map.Entry<String, Handler>> but I could not find another solution.
Problems are:

Doing it like case 1 always results in a NPE when accessing rlogConfig.handlerConfigMap.entrySet()

case 1:

`class RlogProcessor {

    private static final String FEATURE = "rlog";

    @BuildStep
    FeatureBuildItem feature() {
        return new FeatureBuildItem(FEATURE);
    }

    @BuildStep
    @Record(ExecutionTime.RUNTIME_INIT)
    List<NamedLogHandlersBuildItem> addRLogHandlers(final RlogConfig rlogConfig,
            RlogHandlerValueFactory rlogHandlerValueFactory) {
        return rlogConfig.handlerConfigMap.entrySet().stream()
                .map(entry -> rlogHandlerValueFactory.createNamedHandler(entry.getKey(), entry.getValue()))
                .map(NamedLogHandlersBuildItem::new).collect(Collectors.toList());
    }
}

@Recorder
public class RlogHandlerValueFactory {

    private static final Logger LOG = Logger.getLogger(RlogHandlerValueFactory.class);

    private Handler createHandler(HandlerConfig handlerConfig) {
        RlogHandler handler = new RlogHandler();
        handler.setLevel(handlerConfig.level);
        handler.setFilename(handlerConfig.filename);
        handler.setPort(handlerConfig.port);
        handler.setStandalone(handlerConfig.standalone);
        handler.setLocalDir(handlerConfig.localDir);
        handler.setFormatter(new SimpleFormatter());

        return handler;
    }
}`

When returning List<RuntimeValue<Map.Entry<String, Handler>> and then wrap each of it in separate BuildItem like in case2.
I get the exception: "java.lang.RuntimeException: You cannot invoke stream() directly on an object returned from the bytecode recorder, you can only pass it back into the recorder as a parameter"

case 2:

`class RlogProcessor {

    private static final String FEATURE = "rlog";

    @BuildStep
    FeatureBuildItem feature() {
        return new FeatureBuildItem(FEATURE);
    }

    @BuildStep
    @Record(ExecutionTime.RUNTIME_INIT)
    List<NamedLogHandlersBuildItem> addRLogHandlers(final RlogConfig rlogConfig,
            RlogHandlerValueFactory rlogHandlerValueFactory) {
        return rlogHandlerValueFactory.createNamedHandler(rlogConfig).stream().map(NamedLogHandlersBuildItem::new).collect(Collectors.toList());
    }
}

@Recorder
public class RlogHandlerValueFactory {
    private static final Logger LOG = Logger.getLogger(RlogHandlerValueFactory.class);

    public List<RuntimeValue<Map.Entry<String, Handler>>> createNamedHandler(RlogConfig rlogConfig) {
        return rlogConfig.handlerConfigMap.entrySet().stream().collect(Collectors.toMap(Map.Entry::getKey, entry -> createHandler(entry.getValue()))).entrySet().stream()
                .map(RuntimeValue::new).collect(Collectors.toList());
    }

    private Handler createHandler(HandlerConfig handlerConfig) {
        RlogHandler handler = new RlogHandler();
        handler.setLevel(handlerConfig.level);
        handler.setFilename(handlerConfig.filename);
        handler.setPort(handlerConfig.port);
        handler.setStandalone(handlerConfig.standalone);
        handler.setLocalDir(handlerConfig.localDir);
        handler.setFormatter(new SimpleFormatter());

        return handler;
    }
}`

So IMHO the solution with RuntimeValue<Map<String, Handler>> is the only one which works.

dmlloyd: Are you trying to set it up so that the list of run time handlers can be completely customized at run time?

hartimcwildfly: Sorry, I don't understand what you mean.
Do you mean adding a new handler while quarkus already runs?

dmlloyd: Normally we set up the set of handlers that are going to be used during build and static initialization.  So for example the build items in this case would consist of a `String` name (which is fixed at build) and a `RuntimeValue<Optional<Handler>>` (which can be set at run time, or it can be left out if the handler is disabled at run time).  This means that you couldn't add a new handler without rebuilding the application, but it also optimizes the code for native images where the unused handler code can be eliminated.

hartimcwildfly: Ahh I see. I did not consider this, but as I'm using RuntimeValue<Map<String, Handler>> the additional named handlers should be fully customizable at runtime. (which is what I want and which IMHO makes sense for additional named handlers). For the sake of completion I will test if my assumption is correct. -> So I will build the quarkus app without application.properties and then start it with a config which consists of some additional named handlers.

hartimcwildfly: Yup works as expected.

hartimcwildfly: What's the status? Can this be merged?

dmlloyd: As I said we normally strive for build time binding to ensure that dead code can be eliminated at run time.  We'd have to verify that this actually is the case.  But my queue is non-empty and I won't have time to look into it until next week.

hartimcwildfly: But I don't think that it makes sense to fix/set the config of the loggers at compile time. So I would leave the PR as is.

dmlloyd: I would not propose fixing the configuration at compile time, but I would consider the possibility of fixing the *possible* set of handlers at compile time, with run time configuration enabling or disabling the handlers as needed, plus allowing some properties to be run time configured.  This is the general Quarkus approach to things like this.

As I said though I'll be looking at this next week, and then I can give more concrete feedback.

hartimcwildfly: Ok. Thank you for your answer. Sorry for bothering.

dmlloyd:
Could you explain why `io.quarkus.deployment.builditem.LogHandlerBuildItem` is not sufficient for your needs?  AFAICT nothing here is actually using the name field for anything; am I missing something?

hartimcwildfly:
> Could you explain why `io.quarkus.deployment.builditem.LogHandlerBuildItem` is not sufficient for your needs? AFAICT nothing here is actually using the name field for anything; am I missing something?

In the diff of the PR you don't see that the keySet of the Map is used. But in class io.quarkus.runtime.logging.LoggingSetupRecorder I add all additional named handlers to the standard named handlers (Line 128 "namedHandlers.putAll(additionalNamedHandlersMap);").
And the namedHandlers map uses the keys indeed.

hartimcwildfly:
Push/Reminder.

core/deployment/src/main/java/io/quarkus/deployment/builditem/NamedLogHandlersBuildItem.java
dmlloyd: Here is my final opinion for this.  Do not use a `Map`.   Create a new static or top-level class called `NamedHandler` with fields `String` and `Handler`.  The `RuntimeValue` should have type `RuntimeValue<Optional<NamedHandler>>`.  Each build item publishes one single named handler.

hartimcwildfly: Ok I will try. Thank you for your answer!

hartimcwildfly: Ok. I think I know why it didn't work. I didn't understand the concept with the BuildItems, BuildSteps, Recorders etc. good enough. I don't understand it completely yet, but things are getting better. So just give me some time.

hartimcwildfly: I spent a few hours. But I couldn't find a solution to generate multiple MultiBuildItems in a BuildStep with a runtime configuration containing a Map. I took a look at other MultiBuildItems. They either generate one MultiBuildItem per BuildStep or they produce multiple ones in a single BuiltStep but with static values not coming from the runtime configuration. The only one which covers the case "one buildstep, runtime config, multiple MultiBuildItems" was the AgroalProcessor. But I couldn't copy the pattern to my Processor because the AgroalProcessor is quite complex and I don't know how to adapt.

I would be very happy if you could provide an example how to generate multiple MultiBuildItems from a runtime config containing some kind of collection.

dmlloyd: In your build step, instead of returning a build item, accept a parameter of type `BuildProducer<YourBuildItem>`.  Then you can call `produce` on it as many times as you need to.

hartimcwildfly: I will provide a minimal example which is compatible to this PR. Maybe you find my mistake.

hartimcwildfly: https://github.com/hartimcwildfly/quarkus-logging-extension If I run a quarkus app with the config in the README file, there should be a console handler with the name "0". But when I debug into io.quarkus.runtime.logging.LoggingSetupRecorder#initializeLogging I can see that no NamedLogHandlerBuildItem was created. When stopping at  io.quarkus.logging.rlog.RlogConfig$$accessor#get_handlerConfigMap I determine that the handler config is present. The getter is called various times. The configMap is only empty at the first call.

dmlloyd: You can't actually construct the handler in the build step.  When we run build steps, the code isn't actually executed the way that you would think; it's actually recorded instead.  So the work of this method should be done in a separate class with the `@Record` annotation.

hartimcwildfly: I feared that it has to to with this construct with the bytecode recording.

hartimcwildfly: On which method of which example are you refering? It's quite confusing. Can you make an example of a general structure?

hartimcwildfly: Ahh I think I understand it now. At runtime the acutal buildstep is not run only the code which was recorded at build time. But in my case the recorder code was not recorded at build time because the map was empty so we never jumped from the build step into the recorder code?

hartimcwildfly: Got it: So I need a recorder which consumes the RlogConfig and another recorder which consumes an Map.Entry<String, HandlerConfig>. And the BuildStep just runs the recorder with the RlogConfig.

dmlloyd: Sorry I realize there's an error in my previous post.

A "recorder" is a plain class annotated with `@Recorder`, which is run at run time.  This is a normal class in every respect; Quarkus creates the instance for you and then calls the methods at the appropriate time.

A "recording step" is a build step that has the `@Record` annotation.  It should do as little as possible other than calling methods on your recorder class.

dmlloyd: I'm told that this is not confusing. :-)

hartimcwildfly: Thank you for your patience. Now I feel so dumb :D I'm going to squash the commits and then you can review it (did you see my last commit?) and hopefully merge.

hartimcwildfly: Hmm I now understand how everythings works with the code recording. But the problem stays the same. The recorded code is called at quarkus startup and the returned value will be wrapped in a wrapped in a BuildItem (whose wrapper was created at build time). This means that we know at build time how many/which builditems exist. This is also the reason why whe use Optional<Handler> in the LogHandlerBuildItem.

tl;dr: Using NamedLogHandlerBuildItem for each named log handler does not work!
We have to pack multiple named handlers into one BuildItem.

The named console, file and syslog handlers are not wrapped in a builditem they are instanciated in the recorder method. This is why the amount of them is not fixed at build time.

dmlloyd: Well we come back to the problem that we need to know what the possible handlers are at build time so we can ensure that they're in the final image.  It'd be OK to have an `enable` switch on them, but they have to be known at build time.

The reason that console/file/syslog get away with it is that they are always included.  But making it dynamic at run time leaves it completely open-ended.

hartimcwildfly: And what's wrong with the fully dynamic approach using NamedLogHandlersBuildItem containing the Map<String, Handler> ? I would prefer the dynamic one, but if you decide for the static one, I will edit the PR and do it static. Then we can finally complete this PR and I don't have to bother you any more :D

dmlloyd: My concern is simply that the native image size might be needlessly inflated.  But, I guess I'm worrying before there is a known problem.  So I guess a map is OK.

hartimcwildfly:
Updated code according to the discussion's result.

gsmet:
@dmlloyd could you check the added comments?

dmlloyd (approved):

PR #7569, opened 2020-03-04
Fix wrong string constant name for `inherit`

dmlloyd:


gastaldi (approved):

ia3andy:
@dmlloyd maybe we could have a test? to avoid changing it again ^^

dmlloyd:
Maybe we could!

PR #7575, opened 2020-03-04
Document the fact that Panache stream methods should be closed

### LINKED ISSUE #7492: JDBC resources leaked when using `PanacheQuery<Entity>.stream()`
### Describe the bug
I have a method like this one:

```java
public Set<MyDomainObject> getThingsCalled(String name)
   return MyEntity.<MyEntity>find("name", name).stream()
      .map(entity -> entity.toDomainObject())
      .collect(Collectors.toSet());
}
```

When executed, the following warning is printed in the logs:

```
2020-02-28 17:14:37,005 WARN  [io.agr.pool] (pool-5-thread-5) Datasource '<default>': Closing open connection prior to commit
2020-02-28 17:14:37,005 WARN  [io.agr.pool] (pool-5-thread-5) Datasource '<default>': JDBC resources leaked: 1 ResultSet(s) and 1 Statement(s)
```

If I change the method code to look like this:

```java
public Set<MyDomainObject> getThingsCalled(String name)
   List<MyEntity> entities = MyEntity.<MyEntity>find("name", name).list();
   return entities.stream() 
      .map(entity -> entity.toDomainObject())
      .collect(Collectors.toSet());
}
```

then there is no warning in the logs.

Judging from the above, **it seams to me that the implementation of `stream()` method of `PanacheQuery<Entity>` does not handle DB resources properly**. According to https://github.com/quarkusio/quarkus/issues/5757#issuecomment-558531664 the issue started to appear after `1.0.0.CR1`

### Environment
 - Java version:  1.8.0_172
 - Quarkus version: 1.2.1.Final
 - Build tool: Maven 3.6.2 

### Additional context
Not sure how relevant it is but the actual code runs on separate thread inside user managed transaction. The way the method is called looks more or less like this:

```java
@Inject protected TransactionManager tm;
...
executor().execute(() -> {
   try {
      tm.begin();
      Set<MyDomainObject> result  = getThingsCalled("something");
      tm.commit();
   } catch (Throwable e) {
      rollback(tm);
   } finally {
      ...
   }
});
```

I wouldn't be surprised if the thread and transaction management contribute to the issue.
That said, I still believe there is a bug since `list()` method works fine (or so it seams) while `stream()` does not in the very same context.

loicmathieu:
Fixes #7492

@Sanne I added a note on the Hibernate with Panache guide about the usage of the `stream()` method (it uses ResultSet.stream()), can you validate that it's correct ? As some people asking questionq about it I think it's a good addition to the existing guide.

### DIFF
--- a/docs/src/main/asciidoc/hibernate-orm-panache.adoc
+++ b/docs/src/main/asciidoc/hibernate-orm-panache.adoc
@@ -204,14 +204,17 @@ All `list` methods have equivalent `stream` versions.
 
 [source,java]
 ----
-Stream<Person> persons = Person.streamAll();
-List<String> namesButEmmanuels = persons
-    .map(p -> p.name.toLowerCase() )
-    .filter( n -> ! "emmanuel".equals(n) )
-    .collect(Collectors.toList());
+try (Stream<Person> persons = Person.streamAll()) {
+    List<String> namesButEmmanuels = persons
+        .map(p -> p.name.toLowerCase() )
+        .filter( n -> ! "emmanuel".equals(n) )
+        .collect(Collectors.toList());
+}
 ----
 
-NOTE: The `stream` methods require a transaction to work.
+NOTE: The `stream` methods require a transaction to work. +
+As they perform I/O operations, they should be closed via the `close()` method or via a try-with-resource to close the underlying `ResultSet`.
+If not, you will see warnings from Agroal that will close the underlying `ResultSet` for you.
 
 === Adding entity methods
 


docs/src/main/asciidoc/hibernate-orm-panache.adoc
```diff
@@ -209,9 +209,12 @@ List<String> namesButEmmanuels = persons
     .map(p -> p.name.toLowerCase() )
     .filter( n -> ! "emmanuel".equals(n) )
     .collect(Collectors.toList());
+persons.close(); // close the stream to release I/O resources
```
gsmet: Better use a try with resources IMHO.

docs/src/main/asciidoc/hibernate-orm-panache.adoc
```diff
@@ -209,9 +209,12 @@ List<String> namesButEmmanuels = persons
...
 ----
 
-NOTE: The `stream` methods require a transaction to work.
+NOTE: The `stream` methods require a transaction to work. +
+As they perform I/O operations, they should be closed via the `close()` method or via a try-with-resource to close the underlying ResultSet.
+If not, you will see warnings from Agroal that will close the underlying ResultSet for you.
```
gsmet: Can you add backticks around the two `ResultSet` occurrences?

gsmet (changes requested):

docs/src/main/asciidoc/hibernate-orm-panache.adoc
loicmathieu: Yes,  I agree,  was too lazy to refactor the code ... will do it tomorrow

loicmathieu:
@gsmet I update my PR with your suggestions, ready to be merged !

gsmet (approved):

PR #7588, opened 2020-03-05
Add option to disable the http port when https is enabled

stuartwdouglas:
Redirects any http calls to the https endpoint


gsmet (approved):
Force pushed an update to use the lowercased version of the enum values in the documentation as that's the ones that will be documented.

Let's wait for CI.

PR #7595, opened 2020-03-05
Ensure that container image tests only run on linux when the docker flag is enabled

### LINKED ISSUE #7585: quarkus master build fails if no docker daemon is running
With the addition of the container image for Jib and friends, the quarkus build no longer run if no docker daemon is present.
These tests should only be run if the docker profile is in use.

geoand:
Fixes: #7585


integration-tests/container-image/pom.xml
```diff
@@ -9,6 +9,9 @@
   </parent>
   <artifactId>quarkus-integration-test-container-image</artifactId>
   <name>Quarkus - Integration Tests - Jib extension</name>
```
gsmet: The name looks a bit off?

gsmet:
@geoand I promoted it to get an earlier CI result as I would like to have that one in CR1.

integration-tests/container-image/pom.xml
geoand: Indeed, thanks for spotting it! It should be fixed now

gsmet (approved):

gsmet:
Restarted CI as we had MongoDB failing to start and all the JVM tests got cancelled :/

geoand:
Damn mongo...

geoand:
We need to get rid of that embedded mongo thing. I'll try and take a look soon

machi1990:
> We need to get rid of that embedded mongo thing. I'll try and take a look soon

+1 We can use a mongo container instead? or testcontainers?

gsmet:
@machi1990 yeah whatever is more stable than what we have. I would go with a Docker container started from the pom as for the other tests if doable.

geoand:
That's my idea yes. The only issue I see now is that currently the tests don't need `docker` to run, meaning that you can run them on any machine (which is desirable for the tests that are part of the extensions themselves)

geoand:
But at this point I also believe that having more stability is more desirable than being able to run all the extension tests without docker

geoand:
Took a quick look. We can get rid of most of the `de.flapdoodle.embed.mongo` tests, but the ones that use a replica set I don't think we can...

PR #7596, opened 2020-03-05
Simplify kubernetes tests

geoand:
Remove the unnecessary version retrieval method that was added
and use the Version of the quarkus-builder module class that
does the same thing


gastaldi (approved):
Excellent!

geoand:
I saw someone use it, so I knew I need to copy it :)

gsmet (approved):

gsmet:
@geoand I promoted it to get an earlier CI result as I would like to have that one in CR1.

geoand:
OK thanks

Issue #7601, opened 2020-03-05
The openshift extension causes performance issues in dev mode


tqvarnst:
**Describe the bug**
When adding the quarkus-openshift extension startup times for quarkus in dev mode go from 0.6 sec to more than 6 sec. 

**Expected behavior**
Approx the same startup time in dev mode as before we added the quarkus-openshift extension

**Actual behavior**

**To Reproduce**
1. Create a Quarkus project without quarkus-openshift extension
2. Start in dev mode, record the startup time, and stop
3. Add the quarkus-openshift extension
4. Start in dev mode, record the startup time, compare

**Configuration**
N/A

**Screenshots**
(If applicable, add screenshots to help explain your problem.)

**Environment (please complete the following information):**
 - Output of `uname -a` or `ver`: 
Darwin ovpn-118-188.ams2.redhat.com 19.3.0 Darwin Kernel Version 19.3.0: Thu Jan  9 20:58:23 PST 2020; root:xnu-6153.81.5~1/RELEASE_X86_64 x86_64
 - Output of `java -version`: 
openjdk version "11.0.5" 2019-10-15
OpenJDK Runtime Environment AdoptOpenJDK (build 11.0.5+10)
OpenJDK 64-Bit Server VM AdoptOpenJDK (build 11.0.5+10, mixed mode)
 - GraalVM version (if different from Java): 
N/A
 - Quarkus version or git rev: 
Quarkus master
 - Build tool (ie. output of `mvnw --version` or `gradlew --version`): 
Apache Maven 3.6.3 (cecedd343002696d0abb50b32b541b8a6ba2883f)
Maven home: /Users/tqvarnst/.sdkman/candidates/maven/current
Java version: 11.0.5, vendor: AdoptOpenJDK, runtime: /Users/tqvarnst/.sdkman/candidates/java/11.0.5.hs-adpt
Default locale: en_SE, platform encoding: UTF-8
OS name: "mac os x", version: "10.15.3", arch: "x86_64", family: "mac"

**Additional context**
N/A

iocanel:
@geoand: you are not looking into it, are you?

geoand:
Nope, I'm not at the moment. Feel free to address it if you are already working in that area

tqvarnst:
Cannot reproduce after PR #7606. Closing this issue

geoand:
Thanks for checking @tqvarnst

PR #7606, opened 2020-03-05
fix: container-image-s2i runtime dependencies

iocanel:
This pull request adds a missing required dependency to `container-image-s2i`.
This fixes tests and run-time failures.

### DIFF
--- a/extensions/container-image/container-image-s2i/deployment/pom.xml
+++ b/extensions/container-image/container-image-s2i/deployment/pom.xml
@@ -20,6 +20,10 @@
         </dependency>
         <dependency>
             <groupId>io.quarkus</groupId>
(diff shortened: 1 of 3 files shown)


iocanel:
Possibly Fixes: #7601 
I am not 100% sure if its this particular commit or one of the previous fixes that merged earlier today. However, I can no longer reproduce #7601 .

geoand:
Can we please have a small comment in `pom.xml` on why the dependency is needed? It's not immediately obvious

gsmet:
Can we add the comment later once this PR is in if the PR looks good? I would like to wrap up CR1 and this one looks important.

geoand:
OK sure

geoand (approved):

PR #7608, opened 2020-03-05
Show hostname to be added to /etc/hosts

gastaldi:
This helps to fix warnings like

```
2020-03-05 09:19:15,953 WARNING [io.ver.cor.imp.BlockedThreadChecker] (vertx-blocked-thread-checker) Thread Thread[vert.x-eventloop-thread-1,5,main]=Thread[vert.x-eventloop-thread-1,5,main] has been blocked for 3866 ms, time limit is 2000 ms: io.vertx.core.VertxException: Thread blocked
```

### DIFF
--- a/extensions/netty/runtime/src/main/java/io/quarkus/netty/runtime/NettyRecorder.java
+++ b/extensions/netty/runtime/src/main/java/io/quarkus/netty/runtime/NettyRecorder.java
@@ -25,7 +25,9 @@ public class NettyRecorder {
                 DefaultChannelId.newInstance();
                 if (System.currentTimeMillis() - start > 1000) {
                     log.warn(
-                            "Localhost lookup took more than one second, you need to add a /etc/hosts entry to improve Quarkus startup time. See https://thoeni.io/post/macos-sierra-java/ for details.");
+                            "Localhost lookup took more than one second, you need to add your hostname ("
+                                    + System.getenv("HOSTNAME")
+                                    + ") to the /etc/hosts entry to improve Quarkus startup time. See https://thoeni.io/post/macos-sierra-java/ for details.");
                 }
             }
         }).start();


extensions/netty/runtime/src/main/java/io/quarkus/netty/runtime/NettyRecorder.java
```diff
@@ -25,7 +25,9 @@ public void run() {
                 DefaultChannelId.newInstance();
                 if (System.currentTimeMillis() - start > 1000) {
```
dmlloyd: You should *always* use `nanoTime` for durations.  While it isn't super likely to cause a problem here, things like NTP might cause false positives.  But anyway it's a good habit to be in!

gastaldi: Thanks for the hint, but I didn't touched that part of the code in this pr :stuck_out_tongue:

extensions/netty/runtime/src/main/java/io/quarkus/netty/runtime/NettyRecorder.java
```diff
@@ -25,7 +25,9 @@ public void run() {
...
                     log.warn(
-                            "Localhost lookup took more than one second, you need to add a /etc/hosts entry to improve Quarkus startup time. See https://thoeni.io/post/macos-sierra-java/ for details.");
+                            "Localhost lookup took more than one second, you need to add your hostname ("
+                                    + System.getenv("HOSTNAME")
```
dmlloyd: The host name might or might not be in the env, and also the host name might not correspond to the actual network name of the network interface.  It might be better to say "You need a hosts entry for `localhost` which maps to `127.0.0.1` and `::1` in order to improve startup time" or "You need a hosts entry corresponding to IP address <whatever>" or something along those lines.

gastaldi: My hosts file already had `localhost`. The error was gone only after I added `ggastald-laptop` in

dmlloyd: Fair 🙂 I didn't "request changes"!

extensions/netty/runtime/src/main/java/io/quarkus/netty/runtime/NettyRecorder.java
dmlloyd: Fair, I didn't "request changes" 🙂

extensions/netty/runtime/src/main/java/io/quarkus/netty/runtime/NettyRecorder.java
gastaldi: > It's the DNS name of the network interface that it used

Right, what I meant is that is what's returned when I type `hostname` in bash :smiley:

dmlloyd: Exactly.  So it's not exactly the right thing, and there might be multiple problematic interfaces besides.  We might want to change this check to run earlier at startup and have it iterate the `NetworkInterface` addresses trivially, with timing info.  Then when it's done we could actually list every IP/host name pair that the user could even cut & paste into the hosts file if they wanted to.

gastaldi:
Closing according @dmlloyd 's comments above

PR #7610, opened 2020-03-05
Adding LANG env to docker build

gastaldi:
This resolves errors while showing the ASCII art banner

### DIFF
--- a/core/deployment/src/main/java/io/quarkus/deployment/pkg/steps/NativeImageBuildStep.java
+++ b/core/deployment/src/main/java/io/quarkus/deployment/pkg/steps/NativeImageBuildStep.java
@@ -88,7 +88,8 @@ public class NativeImageBuildStep {
             if (IS_WINDOWS) {
                 outputPath = FileUtil.translateToVolumePath(outputPath);
             }
-            Collections.addAll(nativeImage, containerRuntime, "run", "-v", outputPath + ":/project:z");
+            Collections.addAll(nativeImage, containerRuntime, "run", "-v", outputPath + ":/project:z",
+                    "--env", "LANG=en_US.UTF-8");
 
             if (IS_LINUX) {
                 if ("docker".equals(containerRuntime)) {


geoand:
Works for me. But I'll let @dmlloyd have the final say

gsmet (changes requested):
I'm -1 for this change, at least until we discuss it first on the mailing list.

maxandersen:
Can't we just pass in the environment charset rather than hardcore it ?

That mean It runs as if you local native build ?

maxandersen:
This doenst not just fix ASCII art banner. It makes charset predictable. 

@gsmet can you state what your concern is ? In what situation is using whatever default charset in our base builders better Than utf-8 or Users default Charset ?

maxandersen:
If we wait we'll not get feedback if it actually breaks and users can still set it where needed.

dmlloyd:
TBH I with @gsmet on this one.  We don't want to force the charset when the problem we're trying to solve is correct banner rendering, which might better be solved at a different layer.

maxandersen:
the way I see it the banner revealed that our call through docker here are using rather custom/random encoding...

geoand:
Given the fact that this only happens when the native image is built through docker, I agree with @maxandersen (although my knowledge of character encodings is very limited)

dmlloyd:
How do we know that UTF-8 is the right charset?  What are the factors that determine the charset in use with Docker and how does that interplay with the JDK or the host OS?  At a minimum that level of analysis should be on the PR and/or commit.

geoand:
That I can completely agree with - essentially if this PR fixes the issue we need to know why

maxandersen:
The image we are talking about here is "quay.io/quarkus/ubi-quarkus-native-image:19.3.1-java8", right?

when I run that guess what LANG is set to ? "en_US.UTF-8" ...

maxandersen:
I really see this as our docker launch here relying on whatever LANG enabled in the graalvm build and thus explicitly setting it should be better.

geoand:
Yeah, it would be great to know how the value ends up being what it is when we don't set anything.

maxandersen:
so, after some trial and error we now know LANG and other env is not set at all by default and the default charset /lang is `en.US-ASCII`which I think we all can agree is *worse* than UTF-8....

geoand:
Superseded by: #7662

PR #7614, opened 2020-03-05
Support for recognizing multiple Quarkus platforms in a project

aloubyansky:
(both maven and gradle supported) and merging their descriptors into a single one (with the platforms imported earlier dominating over those imported later) for the tools to be able to list and add extensions.

@gastaldi if you have time, I'd appreciate if you could review this one. Thanks.

### DIFF
--- a/devtools/maven/src/main/java/io/quarkus/maven/BuildFileMojoBase.java
+++ b/devtools/maven/src/main/java/io/quarkus/maven/BuildFileMojoBase.java
@@ -2,6 +2,7 @@ package io.quarkus.maven;
 
 import java.io.File;
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.List;
 
 import org.apache.maven.model.Dependency;
@@ -22,6 +23,7 @@ import io.quarkus.cli.commands.file.BuildFile;
 import io.quarkus.cli.commands.file.GradleBuildFile;
 import io.quarkus.cli.commands.file.MavenBuildFile;
 import io.quarkus.cli.commands.writer.FileProjectWriter;
+import io.quarkus.platform.descriptor.CombinedQuarkusPlatformDescriptor;
 import io.quarkus.platform.descriptor.QuarkusPlatformDescriptor;
 import io.quarkus.platform.descriptor.resolver.json.QuarkusJsonPlatformDescriptorResolver;
 import io.quarkus.platform.tools.MessageWriter;
@@ -76,7 +78,7 @@ public abstract class BuildFileMojoBase extends AbstractMojo {
                 final MavenBuildFile mvnBuild = new MavenBuildFile(fileProjectWriter);
                 buildFile = mvnBuild;
 
-                Artifact descrArtifact = null;
+                final List<Artifact> descrArtifactList = new ArrayList<>(2);
                 for (Dependency dep : mvnBuild.getManagedDependencies()) {
                     if (!dep.getScope().equals("import") && !dep.getType().equals("pom")) {
                         continue;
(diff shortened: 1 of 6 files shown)


devtools/maven/src/main/java/io/quarkus/maven/BuildFileMojoBase.java
```diff
@@ -137,6 +131,39 @@ public void execute() throws MojoExecutionException {
...
+            if (getLog().isDebugEnabled()) {
+                getLog().debug("Failed to resolve JSON descriptor as " + jsonArtifact);
+            }
+            jsonArtifact = new DefaultArtifact(bomGroupId, bomArtifactId + "-descriptor-json", null, "json",
+                    bomVersion);
+            try {
+                jsonArtifact = mvn.resolve(jsonArtifact).getArtifact();
+            } catch (Exception e1) {
+                if (getLog().isDebugEnabled()) {
+                    getLog().debug("Failed to resolve JSON descriptor as " + jsonArtifact);
```
gastaldi: Add the exception stacktrace to the debug message

devtools/maven/src/main/java/io/quarkus/maven/BuildFileMojoBase.java
```diff
@@ -137,6 +131,39 @@ public void execute() throws MojoExecutionException {
...
+                .resolveFromJson(descrArtifact.getFile().toPath());
+    }
+
+    private Artifact resolveJsonOrNull(MavenArtifactResolver mvn, String bomGroupId, String bomArtifactId, String bomVersion) {
+        Artifact jsonArtifact = new DefaultArtifact(bomGroupId, bomArtifactId, null, "json", bomVersion);
+        try {
+            jsonArtifact = mvn.resolve(jsonArtifact).getArtifact();
+        } catch (Exception e) {
+            if (getLog().isDebugEnabled()) {
+                getLog().debug("Failed to resolve JSON descriptor as " + jsonArtifact);
```
gastaldi: Add the exception stacktrace to the log statement

gastaldi (commented):
I found some minor things so far, will review and give it a try later tonight

maxandersen:
@aloubyansky whats the expected layout in ~/.m2 or similar so I can actually test this "in real life" - got . a real test case or just the artificial/mockup ones created in the test ?

aloubyansky:
You can take quarkus-bom artifacts as a template. I haven't had the time to create a "real life" test. We definitely need more tests than what's included in this PR, involving end-user tools. Creating those is a task in itself though. I preferred to get the core change in first and asap.

devtools/maven/src/main/java/io/quarkus/maven/BuildFileMojoBase.java
aloubyansky: TBH, I'd rather remove this logging altogether. Given that we are trying every imported BOM here, if it happens to be not a platform BOM is pretty much a usual situation. IMO, logging a stacktrace is going to make look scarier than it actually is.
The reason I'd consider removing this log message is that if you enable debug, the resolver itself is going to be pretty verbose about what's been tried and what didn't work.

devtools/maven/src/main/java/io/quarkus/maven/BuildFileMojoBase.java
aloubyansky: Same as above.

gsmet:
What's the status of this? There's a good chance I will wrap a CR2 on Monday evening so it would be a good idea to get this in before that.

Thanks!

devtools/maven/src/main/java/io/quarkus/maven/BuildFileMojoBase.java
```diff
@@ -147,7 +174,9 @@ private String resolveValue(String expr, BuildFile buildFile) throws IOException
         if (expr.startsWith("${") && expr.endsWith("}")) {
             final String v = buildFile.getProperty(expr.substring(2, expr.length() - 1));
             if (v == null) {
-                getLog().debug("Failed to resolve version of " + v);
+                if (getLog().isDebugEnabled()) {
+                    getLog().debug("Failed to resolve version of " + v);
```
gastaldi:
```suggestion
                    getLog().debug("Failed to resolve version of " + expr);
```

Because v == null :wink:

gastaldi (changes requested):

devtools/maven/src/main/java/io/quarkus/maven/BuildFileMojoBase.java
aloubyansky: Fixed, thanks!

maxandersen (approved):
+1 on getting the notion of a combined platform into core in 1.3 so its easier/simpler to evolve this and later expose to users.

One thing I'm concerned about in long run is that all the api is assuming a fully resolved platform (like getCategories, getExtensions, etc.) - for a small platform this isn't an issue; but for multiple layered ones this might become a problem....maybe if the api was more lazy in nature we could encapsulate it better. 

I'm saying +1 on this as platforms will need to be fully resolvable anyways client side for lots of reasons (validation is one  of them) but we'll need some more "queryable" api once we start thinking respository/catalog lookups.

aloubyansky:
@gastaldi anything missing to get this in? Thanks.

gastaldi (approved):
LGTM

Issue #6491, opened 2020-01-10
AWS lambda could not deserialize SNS and SQS events


dkuzmenkov17:
If I subscribe to an AWS lambda for SNS or SQS event, QuarkusStreamHandler could not deserialize the timestamp field from SNS class. It happens because the timestamp is using DateTime from JodaTime library and Jackson could not deserialize it correctly. 

My stack trace 
```
Cannot construct instance of `org.joda.time.DateTime` (although at least one Creator exists): no String-argument constructor/factory method to deserialize from String value ('2019-12-11T13:29:08.731Z')
 at [Source: (lambdainternal.util.NativeMemoryAsInputStream); line: 1, column: 429] (through reference chain: com.amazonaws.services.lambda.runtime.events.SNSEvent["Records"]->java.util.ArrayList[0]->com.amazonaws.services.lambda.runtime.events.SNSEvent$SNSRecord["Sns"]->com.amazonaws.services.lambda.runtime.events.SNSEvent$SNS["Timestamp"]): com.fasterxml.jackson.databind.exc.MismatchedInputException
com.fasterxml.jackson.databind.exc.MismatchedInputException: Cannot construct instance of `org.joda.time.DateTime` (although at least one Creator exists): no String-argument constructor/factory method to deserialize from String value ('2019-12-11T13:29:08.731Z')
 at [Source: (lambdainternal.util.NativeMemoryAsInputStream); line: 1, column: 429] (through reference chain: com.amazonaws.services.lambda.runtime.events.SNSEvent["Records"]->java.util.ArrayList[0]->com.amazonaws.services.lambda.runtime.events.SNSEvent$SNSRecord["Sns"]->com.amazonaws.services.lambda.runtime.events.SNSEvent$SNS["Timestamp"])
	at com.fasterxml.jackson.databind.exc.MismatchedInputException.from(MismatchedInputException.java:63)
	at com.fasterxml.jackson.databind.DeserializationContext.reportInputMismatch(DeserializationContext.java:1342)
	at com.fasterxml.jackson.databind.DeserializationContext.handleMissingInstantiator(DeserializationContext.java:1031)
	at com.fasterxml.jackson.databind.deser.ValueInstantiator._createFromStringFallbacks(ValueInstantiator.java:371)
	at com.fasterxml.jackson.databind.deser.std.StdValueInstantiator.createFromString(StdValueInstantiator.java:323)
[...]

PR #7616, opened 2020-03-05
fix sns event deserialize issue

bnusunny:
SNSEvent, CodeCommitEvent and ScheduledEvent used Joda DateTime. Added JodaModule to Jackson, so that these events can be properly deserialized. 

aws-lambda-java-events uses old Joda-time v2.6, which conflict with JodaModule. I excluded joda-time from aws-lambda-java-events and used the latest version from JodaModule. 

Solves issue #6491

### DIFF
--- a/extensions/amazon-lambda/runtime/pom.xml
+++ b/extensions/amazon-lambda/runtime/pom.xml
@@ -23,6 +23,13 @@
         <dependency>
             <groupId>com.amazonaws</groupId>
             <artifactId>aws-lambda-java-events</artifactId>
+            <version>2.2.7</version>
+            <exclusions>
+                <exclusion>  <!-- declare the exclusion here -->
+                    <groupId>joda-time</groupId>
+                    <artifactId>joda-time</artifactId>
+                </exclusion>
+            </exclusions>
         </dependency>
         <dependency>
             <groupId>io.quarkus</groupId>
@@ -35,6 +42,10 @@
         <dependency>
             <groupId>io.quarkus</groupId>
             <artifactId>quarkus-jackson</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.datatype</groupId>
+            <artifactId>jackson-datatype-joda</artifactId>
         </dependency>
         <dependency>
             <groupId>org.junit.jupiter</groupId>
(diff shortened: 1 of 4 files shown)


machi1990:
@geoand I asked you for a review since it seems jackson related :-)

geoand:
@machi1990 thanks. I'll check in a couple hours

evanchooly (approved):
looks pretty straightforward to me.

extensions/amazon-lambda/runtime/pom.xml
```diff
@@ -23,6 +23,13 @@
         <dependency>
             <groupId>com.amazonaws</groupId>
             <artifactId>aws-lambda-java-events</artifactId>
+            <version>2.2.7</version>
```
geoand: We shouldn't be overriding the version here, but instead change it here: https://github.com/quarkusio/quarkus/blob/master/bom/runtime/pom.xml#L139

extensions/amazon-lambda/runtime/src/main/java/io/quarkus/amazon/lambda/runtime/AmazonLambdaRecorder.java
```diff
@@ -49,7 +50,8 @@ public void setHandlerClass(Class<? extends RequestHandler<?, ?>> handler, BeanC
         beanContainer = container;
         AmazonLambdaRecorder.objectMapper = getObjectMapper()
                 .configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false)
-                .configure(MapperFeature.ACCEPT_CASE_INSENSITIVE_PROPERTIES, true);
+                .configure(MapperFeature.ACCEPT_CASE_INSENSITIVE_PROPERTIES, true)
+                .registerModule(new JodaModule());
```
geoand: I let's go with this for now, but I think we'll want to introduce an `ObjectMapperCustomizer` to the module that takes care of all this

geoand (changes requested):
I added one comment for another time and one request for a change

patriot1burke:
@geoand I'm working on closing out some of the lambda bugs.  I'll add customizer capability to the list.

geoand:
Cool, thanks @patriot1burke

extensions/amazon-lambda/runtime/pom.xml
bnusunny: Sure. Updated.

geoand (approved):

Issue #6701, opened 2020-01-21
Guide for AWS Lambda HTTP yields: io.quarkus.amazon.lambda.runtime.QuarkusStreamHandler ClassNotFoundException in SAM


bluemonk3y:
**Describe the bug**
When following the guide: 
1. https://quarkus.io/guides/amazon-lambda-http I build the GreetingResource etc. 
1. Run > `mvn clean install`
1. Start SAM using: `sam local start-api --template sam.jvm.yaml`
1. Navigating the browsers to: `http://127.0.0.1:3000/hello`
1. Shows class not found on the console:
```
Fetching lambci/lambda:java8 Docker container image......
Mounting /private/var/folders/2x/mp3q5n8d2136yqf5bl2r571m0000gn/T/tmpt84krkw6 as /var/task:ro,delegated inside runtime container
START RequestId: 2d7e86bb-fd7b-1012-35e5-a150c771aa47 Version: $LATEST
java.lang.ClassNotFoundException: io.quarkus.amazon.lambda.runtime.QuarkusStreamHandler
	at java.net.URLClassLoader.findClass(URLClassLoader.java:382)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:424)

```

**Expected behavior**
Should return the result of the get request to the browser (as shown in the guide)

**Actual behavior**
browser returns error message: 
`{"message":"Internal server error"}`

**To Reproduce**
1. Follow the guide
2. run SAM
3. point browser to `http://127.0.0.1:3000/hello`

**Configuration**
```properties
# Add your application.properties here, if applicable.

```

**Screenshots**
(If applicable, add screenshots to help explain your problem.)

**Environment (please complete the following information):**
 - Output of `uname -a` or `ver`: 
 - Output of `java -version`: 
 - GraalVM version (if different from Java): 
 - Quarkus version or git rev: 

**Additional context**
* Running on OSX ElCapitan (10.11.6), 

Java version "1.8.0_231"
Java(TM) SE Runtime Environment (build 1.8.0_231-b11)
Java HotSpot(TM) 64-Bit Server VM (build 25.231-b11, mixed mode)

* Latest Docker-toolbox

geoand:
cc @evanchooly @patriot1burke

logscape:
this looks like an issue with SAM use of DockerMachine permission on sharing/copying files to the container for execution/mounting. 

I reproduced it with the SAM HelloWorld app. 
[...]

PR #7619, opened 2020-03-06
Set a proper temp directory for Amazon Lambda

patriot1burke:


patriot1burke:
@oztimpower Here's the PR too.  Please try it out because I can't reproduce this on my AWS account

oztimpower:
I confirm that #7619, resolves the issue with my Kotlin Amazon Lambda project.  i.e., forcing the tmp dirs in AmazonLambdaProcessor.java works.

oztimpower:
Whilst you are at it, any chance for Amazon Lambda you could force the following property:
`quarkus.package.uber-jar=true
`
Only after some pain and reading the documentation (right down the bottom) does on realise the missing QuarkusStreamHandler is due to a thin jar being built by default.  Indeed #6701 issues likely to be caused by the thin jar and not Docker / Amazon SAM-CLI

patriot1burke:
@oztimpower I'll do that in a separate PR.

extensions/amazon-lambda/deployment/src/main/java/io/quarkus/amazon/lambda/deployment/AmazonLambdaProcessor.java
```diff
@@ -195,6 +195,12 @@ void ipv4Only(BuildProducer<SystemPropertyBuildItem> systemProperty) {
         systemProperty.produce(new SystemPropertyBuildItem("java.net.preferIPv4Stack", "true"));
     }
 
+    @BuildStep
+    void tmpdirs(BuildProducer<SystemPropertyBuildItem> systemProperty) {
+        systemProperty.produce(new SystemPropertyBuildItem("java.io.tmpdir", "/tmp"));
```
gastaldi: I'm not sure this will work in Windows environments

patriot1burke: Lambda doesn't run on Windows.  But, good point, I should not do this if its test mode.

patriot1burke: Ok, I just set the properties if we're not in dev or test mode.

gsmet:
@patriot1burke I'm not convinced building an uberjar by default is such a good idea. I would prefer us to stay consistent with the default Quarkus behavior.

gsmet:
The PR is all borked with other commits. It needs a proper rebase.

gastaldi:
@patriot1burke you need to do a `git rebase master && git push -f` in your `lambda-tmpdir` branch

patriot1burke:
@gsmet Sorry...i forgot to fetch/rebase before squash

patriot1burke:
There's no reason this should be failing where its failing.

patriot1burke:
@gsmet The TCK failure isn't my PR.  Isn't there new rules that this won't block a merge?

gastaldi (approved):
LGTM, although I'd prefer those properties to be configurable somehow

PR #7623, opened 2020-03-06
Don't run postgres tests by default

stuartwdouglas:


geoand:
Shouldn't we do this for the rest of the `reactive-mysql-client` as well?

stuartwdouglas:
yes, that must run after the next failure I hit so I did not notice.

stuartwdouglas:
That can be a seperate PR thought

stuartwdouglas:
Maybe we need a basic mvn install CI test with nothing running to make sure we don't break this again

geoand (approved):

geoand:
I approved the PR, feel free to proceed as you see fit :)

gsmet:
I'm not sure about that one. IIRC, the reactive-pg-client tests are designed to "work" without any PostgreSQL instance.

stuartwdouglas:
well they don't

stuartwdouglas:
They work on CI because we already have PG running, if you run them locally they fail

PR #7625, opened 2020-03-06
Use mongo started from docker instead of flapdoodle

geoand:
This is done because `flapdoodle` has proven unstable in CI. 

Unfortunately this PR doesn't completely remove `flapdoodle`, because the JVM tests that use it, need to have access to a Mongo replica set which is hard to do in docker.

### DIFF
--- a/integration-tests/mongodb-client/pom.xml
+++ b/integration-tests/mongodb-client/pom.xml
@@ -40,8 +40,28 @@
             <scope>test</scope>
         </dependency>
         <dependency>
-            <groupId>de.flapdoodle.embed</groupId>
-            <artifactId>de.flapdoodle.embed.mongo</artifactId>
+            <groupId>org.testcontainers</groupId>
+            <artifactId>testcontainers</artifactId>
+            <scope>test</scope>
+            <exclusions>
+                <exclusion>
+                    <groupId>javax.xml.bind</groupId>
+                    <artifactId>jaxb-api</artifactId>
+                </exclusion>
+                <exclusion>
+                    <groupId>javax.annotation</groupId>
+                    <artifactId>javax.annotation-api</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+        <dependency>
+            <groupId>jakarta.annotation</groupId>
+            <artifactId>jakarta.annotation-api</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.jboss.spec.javax.xml.bind</groupId>
+            <artifactId>jboss-jaxb-api_2.3_spec</artifactId>
             <scope>test</scope>
         </dependency>
     </dependencies>
@@ -54,6 +74,12 @@
             </resource>
         </resources>
         <plugins>
+            <plugin>
+                <artifactId>maven-surefire-plugin</artifactId>
+                <configuration>
+                    <skip>true</skip>
+                </configuration>
+            </plugin>
             <plugin>
                 <artifactId>maven-failsafe-plugin</artifactId>
                 <configuration>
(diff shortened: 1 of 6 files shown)


loicmathieu:
As I remember the issue is that Windows CI previously didn't have a docker engine.

If now it has one, maybe using Testcontainers is easier ?

To run a replica set (it will also be mandatory for mongodb with Panache as transaction support may lands some days) it's not yet possible easily with testcontainers but it can be done via https://github.com/silaev/mongodb-replica-set/

geoand:
That looks interesting indeed!

geoand:
I'll update the PR to use testcontainers since it does indeed seem easier to use

geoand:
@loicmathieu PR updated to use test containers

integration-tests/mongodb-client/pom.xml
```diff
@@ -40,8 +40,27 @@
...
+            <exclusions>
+                <exclusion>
+                    <groupId>javax.xml.bind</groupId>
+                    <artifactId>jaxb-api</artifactId>
+                </exclusion>
+                <exclusion>
+                    <groupId>javax.annotation</groupId>
+                    <artifactId>javax.annotation-api</artifactId>
+                </exclusion>
+            </exclusions>
```
loicmathieu: Missing the test scope ?

integration-tests/mongodb-panache/pom.xml
```diff
@@ -62,6 +57,30 @@
...
+            <exclusions>
+                <exclusion>
+                    <groupId>javax.xml.bind</groupId>
+                    <artifactId>jaxb-api</artifactId>
+                </exclusion>
+                <exclusion>
+                    <groupId>javax.annotation</groupId>
+                    <artifactId>javax.annotation-api</artifactId>
+                </exclusion>
+            </exclusions>
```
loicmathieu: Missing test scope ?

loicmathieu (changes requested):
It appears you start MongoDB but never stop it, I know Ryuk will close it at the end of the test but this can slighly delay the close so maybe you should manually close.

Also, MongoDB client can reconnect, so you can safely use the standard integration way (via annotations https://www.testcontainers.org/test_framework_integration/junit_5/), with it the container will be started after Quarkus but before any test so it will works (I did that in my first implementation of the MongoDB with Panache IT and go back to flapdoodle as Testcontainers didn't works on Quarkus windows CI builds at this time).

integration-tests/mongodb-client/pom.xml
geoand: Fixed, thanks

integration-tests/mongodb-panache/pom.xml
geoand: Fixed

geoand:
> It appears you start MongoDB but never stop it, I know Ryuk will close it at the end of the test but this can slighly delay the close so maybe you should manually close.
> 

Not sure what you mean here because Mongo does get closed: https://github.com/geoand/quarkus/blob/5b490b99bebc0bfcc3cf6589f528e3daa1af07bc/integration-tests/mongodb-panache/src/test/java/io/quarkus/it/mongodb/panache/MongoTestResource.java#L25

> Also, MongoDB client can reconnect, so you can safely use the standard integration way (via annotations https://www.testcontainers.org/test_framework_integration/junit_5/), with it the container will be started after Quarkus but before any test so it will works (I did that in my first implementation of the MongoDB with Panache IT and go back to flapdoodle as Testcontainers didn't works on Quarkus windows CI builds at this time).

Using the Quarkus way of integrating with test containers works under all circumstances, @cescoffier can tell you more since he has had to deal with multiple related issues

geoand:
I played with `silaev/mongodb-replica-set` and although it seems exactly like what we want, I couldn't get it to work reliably

gsmet:
Does it mandate Docker for the default build? Because we don't want that.

geoand:
It does

loicmathieu:
> Not sure what you mean here because Mongo does get closed:

My bad, didn't saw it as I didn't open the full class ...

> Using the Quarkus way of integrating with test containers works under all circumstances, @cescoffier can tell you more since he has had to deal with multiple related issues

It works in both way so it's OK for me

geoand:
@gsmet this is only for the integration tests which require docker anyway

geoand:
I'll close this since we haven't had issues with Mongo in CI lately

PR #7630, opened 2020-03-06
Show useful warning message when @ConfigProperty used in JAX-RS Provider

### LINKED ISSUE #7607: @ConfigProperty not resolved in native mode
Hello, 

a property injected via `org.eclipse.microprofile.config.inject.ConfigProperty` in a filter like the one of the following example is not resolving when the application is running in native mode. Things are working normally if the value is introduced via `org.eclipse.microprofile.config.ConfigProvider.getConfig()` or injected using a Quarkus configuration object annotated with `io.quarkus.arc.config.ConfigProperties`. 

Quarkus version used: **1.2.1.Final**.

```
@Provider
@ApplicationScoped
public class AuthorizationFilter implements ContainerRequestFilter {

    @ConfigProperty(name = "client.key", defaultValue = "undefined")
    String clientKey;

    @Override
    public void filter(ContainerRequestContext context) {
        // clientKey is "undefined" here...
    }
}
```

Thank you.

geoand:
Fixes: #7607

### DIFF
--- a/extensions/resteasy-common/deployment/src/main/java/io/quarkus/resteasy/common/deployment/ResteasyCommonProcessor.java
+++ b/extensions/resteasy-common/deployment/src/main/java/io/quarkus/resteasy/common/deployment/ResteasyCommonProcessor.java
@@ -24,8 +24,11 @@ import org.jboss.jandex.AnnotationTarget;
 import org.jboss.jandex.AnnotationValue;
 import org.jboss.jandex.AnnotationValue.Kind;
 import org.jboss.jandex.DotName;
+import org.jboss.jandex.FieldInfo;
 import org.jboss.jandex.IndexView;
 import org.jboss.jandex.MethodInfo;
+import org.jboss.jandex.ParameterizedType;
+import org.jboss.jandex.Type;
 import org.jboss.logging.Logger;
 import org.jboss.resteasy.core.MediaTypeMap;
 import org.jboss.resteasy.plugins.interceptors.AcceptEncodingGZIPFilter;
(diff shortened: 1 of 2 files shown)


geoand:
An example of the logged message is:

```
[WARNING] [io.quarkus.resteasy.common.deployment.ResteasyCommonProcessor] Directly injecting a @ConfigProperty into a JAX-RS provider may lead to unexpected results. To ensure proper results, please change the type of the field to javax.inject.Provider<java.lang.String>. Offending field is 'message' of class 'org.acme.getting.started.DummyProvider'
```

extensions/resteasy-common/deployment/src/main/java/io/quarkus/resteasy/common/deployment/ResteasyCommonProcessor.java
```diff
@@ -187,6 +191,29 @@ JaxrsProvidersToRegisterBuildItem setupProviders(BuildProducer<ReflectiveClassBu
...
+        if (configPropertyInstances == null) {
+            return;
+        }
+        for (AnnotationInstance configPropertyInstance : configPropertyInstances) {
+            if (configPropertyInstance.target().kind() != AnnotationTarget.Kind.FIELD) {
+                continue;
+            }
+            FieldInfo field = configPropertyInstance.target().asField();
+            Type fieldType = field.type();
+            if (ResteasyDotNames.CDI_PROVIDER.equals(fieldType.name())) {
```
mkouba: `javax.enterprise.inject.Instance` is also legal... and it extends `Provider` ;-).

mkouba (dismissed):

extensions/resteasy-common/deployment/src/main/java/io/quarkus/resteasy/common/deployment/ResteasyCommonProcessor.java
geoand: Actually I think we should just propose `Instance` since the name doesn't class with the JAX-RS Provider. WDYT?

geoand: Fixed

gastaldi (approved):

PR #7648, opened 2020-03-06
Feature BuildItem added for AWS Xray

oztimpower:
Add the feature text "aws-xray" for AWS Xray to display upon startup.  Beneficial to have this info.

`2020-03-06 20:40:29,198 INFO  [io.quarkus] (main) Installed features: [amazon-lambda, aws-xray, cdi, kotlin, resteasy-jackson]
`

### DIFF
--- a/core/deployment/src/main/java/io/quarkus/deployment/builditem/FeatureBuildItem.java
+++ b/core/deployment/src/main/java/io/quarkus/deployment/builditem/FeatureBuildItem.java
@@ -13,6 +13,7 @@ public final class FeatureBuildItem extends MultiBuildItem {
     public static final String AMAZON_LAMBDA = "amazon-lambda";
     public static final String ARTEMIS_CORE = "artemis-core";
     public static final String ARTEMIS_JMS = "artemis-jms";
+    public static final String AWS_XRAY = "aws-xray";
     public static final String CACHE = "cache";
     public static final String CDI = "cdi";
     public static final String CONFIG_YAML = "config-yaml";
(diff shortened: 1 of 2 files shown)


core/deployment/src/main/java/io/quarkus/deployment/builditem/FeatureBuildItem.java
```diff
@@ -13,6 +13,7 @@
     public static final String AMAZON_LAMBDA = "amazon-lambda";
     public static final String ARTEMIS_CORE = "artemis-core";
     public static final String ARTEMIS_JMS = "artemis-jms";
+    public static final String AWS_XRAY = "aws-xray";
```
gsmet: So now I have a bit of a dilemma because we should probably have a consistent prefix between lambda and xray :).

I don't think we will change the extension name, it's a bit late for that probably but maybe we could change the feature names.

AFAIU, the correct naming scheme is AWS Lambda and AWS XRay.

@patriot1burke @evanchooly

gsmet (changes requested):
Marking it as Request change so that we have the discussion before merging.

patriot1burke:
prefix is 'amazon'.  We can change it to aws.

patriot1burke:
@oztimpower Do we really want "aws-xray" listed as a feature?

oztimpower:
I suspected you might say that since you would have created it when you did the build item if you felt it necessary. AWS X-ray is useful for all AWS services, it may not be ideal longer term to subsume under lambda. 

My reasoning for adding a feature for it was that I wanted to be conscious when it was activated since it is heavy, adds a lot of bulk to the deployment, for example a 128MB Quarkus lambda runs out of memory and won’t start. I’m quite a fan of the extensions appearing in the installed features, it’s a real plus of using Aquarius extension. Personally I think every extension should have a feature named. 

However I totally agree if you don’t feel it needs you no loss,  can ignore this PR.

oztimpower:
Closing due to not required.

PR #7654, opened 2020-03-06
updated hibernate

SR-Lut3t1um:


gsmet:
Thanks but we don't want to upgrade to Hibernate 6 Alpha. It's far from being usable for all usages.

SR-Lut3t1um:
Allright I'm gonne rerequest once a more stable comes out.

gsmet:
Keep in mind that some of us are part of the Hibernate team so we will coordinate this update once we think it's doable.

PR #7657, opened 2020-03-06
refactor: remove CompositeMigrationResolverSubstitutions substitution since it is no longer needed

machi1990:
I tried to remove this substitution and seems like [CI in my fork](https://github.com/machi1990/quarkus/runs/490905311?check_suite_focus=true) is happy.

I might be wrong so I am opening the PR to see whether we still need this substitution or not (in which case this PR can be merged).

/cc @gastaldi @Sanne


gastaldi (approved):
That makes sense

PR #7658, opened 2020-03-06
fix: allow Flyway and Liquibase to play together by making JdbcDataSourceSchemaReadyBuildItem a multiple build item

machi1990:
The liquibase extension was merged in #6334 which means that we have multiple possibilities of making a DB schema ready.
This makes the build item that controls whether DB schemas are ready a mutliple build item so that it can be produced by several extensions.

Without this we have the following extensions when we try to cohabite Flyway and Liquibase
```
io.quarkus.builder.ChainBuildException: Multiple producers of item class
io.quarkus.agroal.deployment.JdbcDataSourceSchemaReadyBuildItem(io.quarkus.flyway.FlywayProcessor#configureRuntimeProperties)
```

Also avoid the usage of star imports in liquibase integration tests


gsmet (approved):
I wouldn't expect people to use both but sure better do that.

machi1990:
Thanks @gsmet

Issue #6260, opened 2019-12-19
Improve efficiency of CI pipeline resources by partial rebuild


emmanuelbernard:
**Description**

Currently PR builds run all of our integration tests for every PR, which takes several hours and will probably take longer as time goes on and more tests are added. We could improve CI pipeline build time and resource usage by computing which integration tests are impacted by the files being changed in the PR, and only run the impacted integration tests.

Offer partial build run based on what has changed in a PR, vs the full build run we have today.

See discussion there https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/quarkus-dev/FTLRYTfmELw/wXxWS2mhAgAJ

aguibert:
@emmanuelbernard I don't have authority to edit the description or assign the issue to myself, but can you please assign this one to me and also update the description to be the following?

----

Title: Improve efficiency of CI pipeline resources by only running impacted tests

Description: Currently PR builds run all of our integration tests for every PR, which takes several hours and will probably take longer as time goes on and more tests are added. We could improve CI pipeline build time and resource usage by computing which integration tests are impacted by the files being changed in the PR, and only run the impacted integration tests.

emmanuelbernard:
@aguibert done

emmanuelbernard:
Is it something we will work more on in the next 3-6 months?

aguibert:
Yes, I'm planning to have this done within that timeframe.

However, it depends on Quarkus being a monorepo. If that plan changes please let me know because the strategy I'm using would not work.

emmanuelbernard:
Likely what's inside the repo today will remain (except a few moves as necessary) but newer extensions will not be in the monorepo. That's a likely future anyways.

emmanuelbernard:
hello @aguibert , with the other stuff you have on the fire, when do you think you want and can make this to land ? end of june, end of july, end of... I'm trying to map the task to a Quarkus release on the roadmap with option to move things out / adjust.

aguibert:
hi @emmanuelbernard, sorry, I should have posted an update here sooner. Yes, this is on the back burner since I'm focusing on hibernate reactive and DB2 reactive lately.

For now I'd guess this would be around end of july, if at all.

maxandersen:
related: https://github.com/quarkusio/quarkus/issues/7717

PR #7660, opened 2020-03-06
WIP - Add dependency checker module for reducing amount of native mode tests

aguibert:


.github/workflows/native-cron-build.yml
```diff
@@ -29,8 +32,8 @@ jobs:
       - name: Checkout Quarkus
         uses: actions/checkout@v2
         with:
-          repository: quarkusio/quarkus
-          ref: master
+          repository: aguibert/quarkus
```
machi1990: We should probably revert all changes in this file?

.github/workflows/native-cron-build.yml
```diff
@@ -39,34 +42,21 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-maven-
 
+      - name: Run dependency checker
```
machi1990: this should be in https://github.com/quarkusio/quarkus/blob/master/.github/workflows/ci-actions.yml?

independent-projects/utilities/dependency-checker/pom.xml
```diff
@@ -0,0 +1,86 @@
...
+		<!-- <dependency> -->
+		<!-- <groupId>org.eclipse.jgit</groupId> -->
+		<!-- <artifactId>org.eclipse.jgit</artifactId> -->
+		<!-- <version>5.6.0.201912101111-r</version> -->
+		<!-- </dependency> -->
+		<!-- <dependency> -->
+		<!-- <groupId>org.slf4j</groupId> -->
+		<!-- <artifactId>slf4j-log4j12</artifactId> -->
+		<!-- <version>1.7.26</version> -->
+		<!-- </dependency> -->
```
machi1990: Can be removed?

.github/workflows/native-cron-build.yml
```diff
@@ -39,34 +42,21 @@ jobs:
...
+        run: |
+          echo "@AGG Inside of dependency checker task"
+          echo "GITHUB_SHA=${GITHUB_SHA}"
+          echo "GITHUB_REF=$GITHUB_REF"
+          echo "GITHUB_HEAD_REF=$GITHUB_HEAD_REF"
+          echo "GITHUB_BASE_REF=$GITHUB_BASE_REF"
+          mvn -B -f independent-projects/utilities/dependency-checker/pom.xml package
+          java -jar independent-projects/utilities/dependency-checker/target/quarkus-utilities-dependency-checker-999-SNAPSHOT-jar-with-dependencies.jar ${GITHUB_BASE_REF} ${GITHUB_HEAD_REF}
```
aguibert: This section is just a WIP while I figure out how to properly hook the tool into actions

.github/workflows/native-cron-build.yml
aguibert: hi @machi1990, thanks for having a look at the PR. All the workflow changes right will be rolled back. I've hacked up the workflow files just to make the testing quicker

machi1990: Thanks, realised that I jumped quickly with the review and missed your comment about this being a WIP.

aguibert:
For #6260

gsmet:
@famod isnt what you're working on supposed to replace this?

famod:
@gsmet Since I wasn't really aware of this until now, I was/am not aiming to "replace" this but yeah this is more or less the same approach.
My approach with GIB would leave all those fine grained module-specific jobs untouched, though. You always have to be prepared for a full build and one single job/build might blow up then (timeouts etc.).

gastaldi:
Closing as this needs a rebase and we're aiming to use GIB instead

PR #7664, opened 2020-03-07
changed tarLongFileMode to posix to mitigate problem with long group id

emattheis:
I ran into an issue building Quarkus on my Mac at work which is bound to Active Directory. The assembly plugin failed in the doc project because my group ID was too long ([related post on stack overflow](https://stackoverflow.com/questions/33819438/maven-assembly-plugin-group-id-1377585961-is-too-big-error)).

Switching  `<tarLongFileMode>` from `gnu` to `posix` solves the issue and doesn't break the build. I'm not sure if it has implications on other platforms, though.

### DIFF
--- a/docs/pom.xml
+++ b/docs/pom.xml
@@ -200,7 +200,7 @@
                             <appendAssemblyId>false</appendAssemblyId>
                             <outputDirectory>target/</outputDirectory>
                             <workDirectory>target/assembly/work</workDirectory>
-                            <tarLongFileMode>gnu</tarLongFileMode>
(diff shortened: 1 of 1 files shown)


geoand:
Thanks for your contribution!

I am not qualified to reviews this, @gsmet maybe?

gsmet:
What worries me a bit is that I think we already changed it to the current value.

Let me check.

gsmet:
Apparently, I got things mixed up. It's effectively the value we should use.

gsmet (approved):

gsmet:
Thanks!

PR #7691, opened 2020-03-09
Add some Kubernetes Client build time improvements

geoand:


geoand:
cc @gsmet

integration-tests/kubernetes-client/src/main/java/io/quarkus/it/kubernetes/client/Pods.java
```diff
@@ -28,8 +28,8 @@ public Pods(KubernetesClient kubernetesClient) {
     @GET
     @Produces(MediaType.APPLICATION_JSON)
     @Path("/{namespace}")
-    public List<Pod> pods(@PathParam("namespace") String namespace) {
-        return kubernetesClient.pods().inNamespace(namespace).list().getItems();
+    public Response pods(@PathParam("namespace") String namespace) {
+        return Response.ok(kubernetesClient.pods().inNamespace(namespace).list().getItems()).build();
```
gsmet: I think we should keep it as is as it's probably what users will do. At least, we shouldn't see that as solving the issue.

geoand: I don't agree, at least not in the context of the integration test. The fact that we were returning Pods means that the entire hierarchy was being registered for reflection, which we definitely want to avoid

geoand:
From the email thread discussion (can be found [here](https://groups.google.com/d/msg/quarkus-dev/kZomcIgS6Co/JfOPpVYMAQAJ)), I am adding the performance improvement that this PR gives (numbers from my machine):

master + GraalVM 20 + JDK 8: [total]: 128,918.82 ms,  8.78 GB (I saw max memory of 11.48 GB)
master + GraalVM 20 + JDK 11: [total]: 146,026.04 ms, 14.00 GB (this was the max and final value reported)

PR + GraalVM 20 + JDK 8: [total]: 108,593.92 ms,  7.20 GB (I saw max memory of 8.14 GB)
PR + GraalVM 20 + JDK 11: [total]: 111,277.52 ms, 12.85 GB (this was the max and final value reported)

integration-tests/kubernetes-client/src/main/java/io/quarkus/it/kubernetes/client/Pods.java
geoand: Also if people are looking at integration tests as a model to create their applications, then all the more reason to show them the "better" way of doing this for this use case

geoand:
@gsmet WDYT about this one?

integration-tests/kubernetes-client/src/main/java/io/quarkus/it/kubernetes/client/Pods.java
gsmet: I don't know.

I mean: we need them to be registered for reflection somehow, don't we? Otherwise it won't work well in native image. If our REST-based registration logic doesn't work very well, maybe we need to reconsider it globally?

geoand: The used Kubernetes Model classes are already registered by the Kubernetes extension (weakly) - and that is why the native test works :)

gsmet: Yes I know that. What I'm wondering is: should we generalize that or not? So that the model can be used in the REST API calls without thinking twice about it.

I'm unsure about the consequences thus the reason I ask.

geoand: You can use it now without having to do anything. The use of `Pod` as a result type was really an oversight on my part when I created the integration test, it shouldn't have been done that way because it basically shadows the reflection registration that the Kubernetes extensions does by bringing the RESTEasy reflection generation in the mix

gsmet: Users will do that. You can be sure of that. So should we change the RESTEasy reflection registration to be weak too?

That's what I'm trying to ask since the beginning :).

geoand: We can certainly do that, but I think it should be a following PR. We would need to do introduce a new Build Item that would be consumed by RESTEasy - or maybe it needs to be even more generalized

core/deployment/src/main/java/io/quarkus/deployment/builditem/nativeimage/ReflectiveClassBuildItem.java
```diff
@@ -121,6 +121,7 @@ public static Builder builder(String... className) {
         private boolean methods;
         private boolean fields;
         private boolean finalFieldsWritable;
+        private boolean weak = false;
```
gastaldi:
```suggestion
        private boolean weak;
```

gastaldi (approved):
LGTM. Added a minor optional change

core/deployment/src/main/java/io/quarkus/deployment/builditem/nativeimage/ReflectiveClassBuildItem.java
geoand: Done

integration-tests/kubernetes-client/src/main/java/io/quarkus/it/kubernetes/client/Pods.java
geoand: Furthermore, we haven't had any users complain so far, so I maybe we shouldn't add extra complexity unless it's necessary

geoand:
@gsmet unless you have objections, I will merge this and look at doing the improvement you suggested later

geoand:
This is all the more important with the mode to JDK 11 in CI

PR #7694, opened 2020-03-09
Register methods marked with @JsonCreator for reflection

gsmet:


gsmet:
@geoand does it make sense to you?

extensions/jackson/deployment/src/main/java/io/quarkus/jackson/deployment/JacksonProcessor.java
```diff
@@ -66,6 +68,9 @@
     @Inject
     BuildProducer<ReflectiveHierarchyBuildItem> reflectiveHierarchyClass;
 
+    @Inject
+    BuildProducer<ReflectiveMethodBuildItem> reflectiveMethod;
```
geoand: Man do I dislike this pattern...

But of course I also like consistency, so better be consistently ugly :)

geoand (approved):
Makes perfect sense!

extensions/jackson/deployment/src/main/java/io/quarkus/jackson/deployment/JacksonProcessor.java
gsmet: Yeah, same here. I could change it but we have more pressing things to do, I think :).

geoand: Absolutely!

PR #7701, opened 2020-03-09
Fix #7700: Actually use the JBoss Nexus artifacts in the Azure mvn se…

### LINKED ISSUE #7700: Minor mistake in azure-mvn-settings.xml
**Description**
Looks like we made a mistake in https://github.com/quarkusio/quarkus/commit/3c1c59d875aef0609bf002d34bb439e45fb0e13d , which I've only hit today as I was using the settings file to setup a clean machine.

AFAIK the intention of that patch was to replace the `public-jboss` group with a reference to the repository - so to avoid the "grouping" feature of Nexus which is not very fast.
But what we really did is rather than expose the JBoss repository is to expose its own mirror of Maven Central .. which is pointless as we already have a mirror for central, but is also incorrect as we can't find the dependencies which are exclusively found on the JBoss Nexus.

**Implementation ideas**
Replace the URLs with `https://repository.jboss.org/nexus/content/repositories/public-jboss/` instead;

See for example the content of, and compare:
 - https://repository.jboss.org/nexus/content/repositories/central/org/apache/maven/plugins/maven-compiler-plugin/
 - https://repository.jboss.org/nexus/content/repositories/public-jboss/org/apache/maven/plugins/maven-compiler-plugin/

Sanne:
…ttings

Assigning to @gsmet to make sure we're not going back and forth on this one: we already changed approaches a couple of times and might be missing something.

Fixes #7700

### DIFF
--- a/azure-mvn-settings.xml
+++ b/azure-mvn-settings.xml
@@ -8,10 +8,17 @@
           <name>GCS Maven Central mirror EU</name>
           <url>https://maven-central-eu.storage-download.googleapis.com/repos/central/data/</url>
         </repository>
+        <!-- Add the JBoss Nexus second to mirror Maven Central, as sometimes the Google mirrors aren't fresh enough for our needs -->
         <repository>
           <id>jboss-maven-central-proxy</id>
           <name>JBoss Maven Central proxy</name>
           <url>https://repository.jboss.org/nexus/content/repositories/central/</url>
+        </repository>
+        <!-- Also explicitly list the custom JBoss content for plugins: currently necessary for custom patches to maven-compiler-plugin -->
+        <repository>
+          <id>jboss-public-</id>
+          <name>JBoss Maven Central proxy</name>
+          <url>https://repository.jboss.org/nexus/content/repositories/public-jboss/</url>
         </repository>
         <repository>
             <id>gradle-official-repository</id>
@@ -25,10 +32,17 @@
           <name>GCS Maven Central mirror EU</name>
           <url>https://maven-central-eu.storage-download.googleapis.com/repos/central/data/</url>
         </pluginRepository>
+        <!-- Add the JBoss Nexus second to mirror Maven Central, as sometimes the Google mirrors aren't fresh enough for our needs -->
         <pluginRepository>
           <id>jboss-maven-central-proxy</id>
           <name>JBoss Maven Central proxy</name>
           <url>https://repository.jboss.org/nexus/content/repositories/central/</url>
+        </pluginRepository>
+        <!-- Also explicitly list the custom JBoss content for plugins: currently necessary for custom patches to maven-compiler-plugin -->
+        <pluginRepository>
+          <id>jboss-public-</id>
+          <name>JBoss Maven Central proxy</name>
+          <url>https://repository.jboss.org/nexus/content/repositories/public-jboss/</url>
         </pluginRepository>
       </pluginRepositories>
     </profile>


gsmet:
I have only one small question: how come CI works OK with this? (and it started from scratch quite a few times)

azure-mvn-settings.xml
```diff
@@ -10,8 +10,8 @@
         </repository>
         <repository>
           <id>jboss-maven-central-proxy</id>
```
gsmet: This would need an update.

Sanne:
> I have only one small question: how come CI works OK with this? (and it started from scratch quite a few times)

right I was puzzled by that too. I'd say it's because we use aggressive copies of the maven repo as a cache?

WhatI can see is that the current setup doesn't work for me locally: having a clean repo and no settings.xml

azure-mvn-settings.xml
```diff
@@ -28,7 +28,7 @@
         <pluginRepository>
           <id>jboss-maven-central-proxy</id>
           <name>JBoss Maven Central proxy</name>
```
gsmet: The id and the name should be updated too.

Sanne:
BTW this allows me to build much faster. Apparently the previous setup was getting me throttled somehow; I'm not going to measure accurately but the progress is visible in the terminal.

gsmet (changes requested):
I added a few comments.

You have missed the original intent: the Google mirror is not in our hands and is updated quite infrequently. Thus the second level of proxying to JBoss Nexus which acts more as a proxy so is usually more up to date (but slower).

As for downloading custom JBoss jars, they are usually coming from the JBoss parent and it comes with its own repository thus it should work OK.

CI works OK with that setup and it often starts from a totally empty repository.

If you have issue with that, we need to keep the JBoss Central mirror and add a new one, not replace it.

I know it's cumbersome but as soon as you hit Maven Central, your build is mostly dead so we need to put as many chances as possible on our side to avoid that. At least, that was the case on Azure. Currently unsure what the situation is on GH CI and if they are throttled at the Maven Central level as Azure is.

Sanne:
> You have missed the original intent: the Google mirror is not in our hands and is updated quite infrequently. Thus the second level of proxying to JBoss Nexus which acts more as a proxy so is usually more up to date (but slower).

you're right - sorry I forgot. So we still need the full group then, or both repositories explicitly listed.

> As for downloading custom JBoss jars, they are usually coming from the JBoss parent and it comes with its own repository thus it should work OK.

well that doesn't seem to work.

Sanne:
steps to reproduce:

1. _cd ~/.m2/repository/org/apache/maven/plugins_
2. _rm -Rf maven-compiler-plugin_
3. switch to Quarkus directory
4. _mvn dependency:go-offline -s azure-mvn-settings.xml_

gsmet:
Do you know which artifact it can't find? I wonder if we have a regression here.

Sanne:
> Do you know which artifact it can't find? I wonder if we have a regression here.

sure, it's the usual one, the compiler-plugin:

    org.eclipse.aether.resolution.DependencyResolutionException: Failure to find org.apache.maven.plugins:maven-compiler-plugin:jar:3.8.0-jboss-1

I'll amend this PR to include both repositories.

gsmet:
Hmm, so this one should come from the repositories defined in the JBoss
parent. At least that's how it worked on my box and has worked until now.

I should probably check what's going on.

Sanne:
I would be surprised if relying on repositories defined "somewhere" in your dependency tree would not be affected by various limitations, such as ordering issues or problematic priority semantics.

So I'd recommend to make such a dependency to a different repository explicit; BTW this has always been a policy in jboss projects.

Sanne:
spotted one more thing: if I only list the additional repository as a `pluginRepository` it still fails. 

The `jboss-parent-36.pom` lists it exclusively as a `pluginRepository`, but we have a fully fledged dependency to it via module `IDE/Tools`.  So I guess it's really missing.

Sanne:
fixed, added some comments for our future selves and push forced

gsmet:
What's a bit weird is that I build from a totally empty Maven repo for each release. And it works. That's why I'm wondering what's going on there.

Sanne:
You might have a custom settings.xml ?

I normally have one, only spotted this as I happened to build on a clean
machine.

gastaldi:
This PR needs to be rebased since https://github.com/quarkusio/quarkus/pull/7790 was merged

gastaldi:
Doesn't seem to be needed by CI. Closing for now

PR #7711, opened 2020-03-09
Update NativeImageBuildStep.java

SR-Lut3t1um:
This way also newer versions of GraalVM get's a fitting error.


gsmet:
Sorry but for now we decided to only warn on outdated versions so that we can have 20.0.1 supported for instance. 

We usually release quickly enough to update the message accordingly. 

Don’t hesitate to post on Zulip or on the ML to discuss this sort of changes before doing them.

PR #7714, opened 2020-03-09
Allow @QuarkusTest to use constructor injection

geoand:


stuartwdouglas:
I don't really like this, but it is a probably a good compromise. A better solution would be to use JDK hooks to create an object without calling the constructor, but this is annoying to implement and requires a multi release jar to support both 8 and 11.

stuartwdouglas (approved):

geoand:
@stuartwdouglas but this PR doesn't change anything about how the test is actually created. Or am I missing something?

stuartwdouglas:
@geoand actually it would probably be good to catch NullPointerException (unwrapped) from the constructor.newInstance call and add an informative error message about why the ctor parameter is null. The main problem with this is that if they try to use these injected objects and do more than just assign them to a field they will get a NPE.

stuartwdouglas:
My issue is that some classes that use constructor injection may throw an NPE as they get null injected. Adding an informative error message is probably enough though.

geoand:
Ah yes, that makes sense! I'll update the PR.

geoand:
@stuartwdouglas I pushed a fix. The thing I am not sure about however is logging - I didn't see a way to properly log from an extension, so I just used system error. I assume there is a better way?

test-framework/junit5/src/main/java/io/quarkus/test/junit/QuarkusTestExtension.java
```diff
@@ -231,9 +235,15 @@ public void interceptBeforeAllMethod(Invocation<Void> invocation, ReflectiveInvo
...
         T result;
         ClassLoader old = Thread.currentThread().getContextClassLoader();
+        Class<?> requiredTestClass = extensionContext.getRequiredTestClass();
         try {
-            Thread.currentThread().setContextClassLoader(extensionContext.getRequiredTestClass().getClassLoader());
+            Thread.currentThread().setContextClassLoader(requiredTestClass.getClassLoader());
             result = invocation.proceed();
+        } catch (NullPointerException e) {
+            System.err.println(
+                    "When using constructor injection in a test, the only legal operation is to assign the constructor values to fields. Offending class is "
```
stuartwdouglas: I would wrap the exception with this, not log it, so it shows up in the test failures.

geoand: Thanks for the suggestion, it looks much much better in the logs!

PR #7715, opened 2020-03-09
Document which Gradle version to use

lfmunoz:


docs/src/main/asciidoc/gradle-tooling.adoc
```diff
@@ -31,6 +31,9 @@ mvn io.quarkus:quarkus-maven-plugin:{quarkus-version}:create \
 NOTE: If you just launch `mvn io.quarkus:quarkus-maven-plugin:{quarkus-version}:create` the Maven plugin asks
 for user inputs. You can disable (and use default values) this interactive mode by passing `-B` to the Maven command.
 
+NOTE: It is recommended you match gradle version you use with the version which the Quarkus Gradle Plugin is built with:
```
gsmet: I'm not sure about that. It's recommended to use `./gradlew` so that you don't have to care about anything :).

docs/src/main/asciidoc/gradle-tooling.adoc
```diff
@@ -146,7 +149,7 @@ If no extension is found or if more than one extensions match, you will see a re
 [...]
 ----
 
-You can install all extensions wich match a globbing pattern :
+You can install all extensions which match a globbing pattern :
```
gsmet:
```suggestion
You can install all extensions which match a globbing pattern:
```

gsmet (changes requested):

gsmet:
Thanks! I added a suggestion and a comment.

docs/src/main/asciidoc/gradle-tooling.adoc
lfmunoz: @gsmet that is assuming you use the project generator. In my use case I don't use that because I like using the gradle kotlin dsl which the generator doesn't support. Looking at the documentation I had no idea what gradle version to use.

gsmet: OK. I think we probably need a Gradle version attribute similar to what we have with the Maven version.

I'll try to prepare something along those lines and amend your proposal.

Thanks!

gsmet:
I pushed something along the lines of what I had in mind. Will merge once CI is green.

gsmet (approved):

machi1990 (approved):

gsmet:
@lfmunoz thanks for creating this PR. Keep them coming!

PR #7724, opened 2020-03-10
fix how defaultMode in Kubernetes volumes is parsed

### LINKED ISSUE #7722: defaultMode of a ConfigMap volume is 600 (decimal!) by default
**Describe the bug**

When I specify a ConfigMap volume using the Quarkus Kubernetes extension configuration properties, the volume has by default a `defaultMode` of `600`:

```yaml
      volumes:
      - configMap:
          defaultMode: 600
          name: "app-config"
          optional: false
        name: "app-config"
```

This is a decimal number, because it doesn't begin with `0`.

The Fabric8 Kubernetes client even warns about that: `Invalid value: 600: must be a number between 0 and 0777 (octal), both inclusive`.

The documentation says that default value here is `384` (which is `0600`). The corresponding `*Config` classes have `@ConfigItem(defaultValue = "0600")`, but apparently, when that string is converted to `int`, it's parsed as decimal, not octal.

**Expected behavior**
Default value of `defaultMode` is either `384` or `0600`.

**Actual behavior**
Default value of `defaultMode` is `600`.

**To Reproduce**
1. Add the `quarkus-kubernetes` extension.
2. Specify a ConfigMap in `application.properties`.
3. Run `mvn clean package -DskipTests`.
4. See `target/kubernetes/*.yml`.

**Configuration**
```properties
quarkus.kubernetes.deployment-target=openshift
quarkus.openshift.config-map-volumes.app-config.config-map-name=app-config
quarkus.openshift.mounts.app-config.path=/deployments/config
```

Ladicek:
Previously, the `defaultMode` config property of some Kubernetes
volumes (ConfigMap, secret) was always treated as a decimal number,
even though octal number is a common format (and we actually used
an octal number when specifying the default value!).

This commit changes the type of the config property to `String`,
which is later parsed to an `int` when calling the Dekorate API.
At that time, if the string begins with `0`, it's parsed as octal,
otherwise as decimal.

Fixes #7722.

### DIFF
--- a/docs/src/main/asciidoc/kubernetes.adoc
+++ b/docs/src/main/asciidoc/kubernetes.adoc
@@ -464,7 +464,7 @@ Below you will find tables describing all available types.
 |====
 | Property        | Type    | Description | Default Value
 | config-map-name | String  |             |
-| default-mode    | int     |             | 384
+| default-mode    | int     |             | 0600
 | optional        | boolean |             | false
 |====
 
@@ -472,7 +472,7 @@ Below you will find tables describing all available types.
 |====
 | Property     | Type    | Description | Default Value
 | secret-name  | String  |             |
-| default-mode | int     |             | 384           
+| default-mode | int     |             | 0600
 | optional     | boolean |             | false
 |====
 
--- a/extensions/kubernetes/vanilla/deployment/src/main/java/io/quarkus/kubernetes/deployment/SecretVolumeConfig.java
+++ b/extensions/kubernetes/vanilla/deployment/src/main/java/io/quarkus/kubernetes/deployment/SecretVolumeConfig.java
@@ -1,21 +1,25 @@
 package io.quarkus.kubernetes.deployment;
 
+import io.quarkus.runtime.annotations.ConfigGroup;
 import io.quarkus.runtime.annotations.ConfigItem;
 
+@ConfigGroup
 public class SecretVolumeConfig {
 
     /**
      * The name of the secret to mount.
      */
+    @ConfigItem
     String secretName;
 
     /**
      * Default mode.
+     * When specifying an octal number, leading zero must be present.
      *
      * @return The default mode.
      */
     @ConfigItem(defaultValue = "0600")
-    Integer defaultMode;
+    String defaultMode;
 
     /**
      * Optional
--- a/extensions/kubernetes/vanilla/deployment/src/main/java/io/quarkus/kubernetes/deployment/ConfigMapVolumeConfig.java
+++ b/extensions/kubernetes/vanilla/deployment/src/main/java/io/quarkus/kubernetes/deployment/ConfigMapVolumeConfig.java
@@ -14,11 +14,12 @@ public class ConfigMapVolumeConfig {
 
     /**
      * Default mode.
+     * When specifying an octal number, leading zero must be present.
      *
      * @return The default mode.
      */
     @ConfigItem(defaultValue = "0600")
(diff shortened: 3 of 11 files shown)


extensions/kubernetes/vanilla/deployment/src/main/java/io/quarkus/kubernetes/deployment/SecretVolumeConfig.java
```diff
@@ -1,21 +1,25 @@
 package io.quarkus.kubernetes.deployment;
 
+import io.quarkus.runtime.annotations.ConfigGroup;
 import io.quarkus.runtime.annotations.ConfigItem;
 
+@ConfigGroup
```
Ladicek: I don't know why, but this annotation is present in similar classes (`ConfigMapVolumeConfig`, `PvcVolumeConfig`), but it wasn't here. I added it, but I don't know what I'm doing :-)

extensions/kubernetes/vanilla/deployment/src/main/java/io/quarkus/kubernetes/deployment/SecretVolumeConfig.java
```diff
@@ -1,21 +1,25 @@
...
 public class SecretVolumeConfig {
 
     /**
      * The name of the secret to mount.
      */
+    @ConfigItem
```
Ladicek: Same here.

extensions/kubernetes/vanilla/deployment/src/main/java/io/quarkus/kubernetes/deployment/SecretVolumeConfig.java
geoand: It's correct, nice catch!

geoand (approved):
Excellent, thanks!

docs/src/main/asciidoc/kubernetes.adoc
```diff
@@ -464,15 +464,15 @@ Below you will find tables describing all available types.
...
 |====
 
 .SecretVolume
 |====
 | Property     | Type    | Description | Default Value
 | secret-name  | String  |             |
-| default-mode | int     |             | 384           
+| default-mode | int     |             | 0600
 | optional     | boolean |             | false
 |====
```
gsmet: Could someone really have a look to the generated documentation and see if we can use it?

If we have proper Quarkus config now, it will really be a big win.

geoand: @iocanel didn't you say you were going to look into it?

gsmet: Maybe I will have a look and see how it goes.

gsmet: BTW, one things that bugs me is that we have things like `timeout-seconds` in the doc.

I don't know how they have been translated in Quarkus config but they should be `Duration` and named `timeout` without `-seconds`.

That's really something we want done before Final so either @iocanel can have a look or I will, just tell me.

iocanel: I will

iocanel (approved):
LGTM

docs/src/main/asciidoc/kubernetes.adoc
gsmet: @iocanel feel free to ping me if you need some help or guidance on this one.

PR #7731, opened 2020-03-10
QuarkusPlatformTask - addPlatform to builder

rsvoboda:
Bugfix for QuarkusPlatformTask - invoke addPlatform on builder instead of iterating and modifying on the same list


aloubyansky (approved):
Shame on me. Thanks a lot @rsvoboda!

PR #7733, opened 2020-03-10
Qute enhancements

FroMage:
Finally found the time to follow up on the discussions we had at the last f2f, here's my proposal for improving UX for templating.

ATM, in order to do a non-type-safe template you need this template in `templates/hello.txt`:

```
Salut {name}!
```

And this controller:

```java
@Path("hello")
public class HelloResource {
    @Inject
    Template hello;

    @GET
    public TemplateInstance get(@QueryParam String name) {
        return hello.data("name", name);
    }
}
```

My proposal for non-type-safe templates is to move the template to `templates/HelloResource/hello.txt` and write the following code:

```java
@Path("hello")
public class HelloResource {
    @GET
    public TemplateInstance hello(@QueryParam String name) {
// we can change the class/method name, add a static import, or move it to a superinterface to get it in scope easier
        return ResteasyTemplate.data("name", name);
    }
}
```

This will default to using a template under `{className}/{methodName}`, and will (this is still missing) take an optional template name as parameter if you want to override it.

## Type-safe templates

On the other hand (and orthogonally), ATM if you want type-safe templates you need to write this template:

```
{@java.lang.String name}
Salut {name.toLowerCase()}!
```

While the invocation (in any style: old or new) is not ever type-safe, letting you do this:

```java
@Path("hello")
public class HelloResource {
    @GET
    public TemplateInstance hello(@QueryParam String name) {
        return ResteasyTemplate.data("name", 23); // will fail at run-time
    }
}
```

I propose the following variant, guaranteeing type-safety as discussed at the f2f:

The template doesn't have types in there:

```
Salut {name.toLowerCase()}!
```

And the template declares a constructor for the template instance:

```java
@Path("hello")
public class HelloResource {
    public static class hello extends CheckedTemplateInstance {
        public hello(String name) {
        }
    }

    @GET
    public TemplateInstance hello(@QueryParam String name) {
        return new hello(name); // this is type-safe
    }
}
```

This is implemented in this branch and a little bit tested.

Opinions @mkouba @maxandersen @emmanuelbernard @cescoffier @nicmarti ?

### DIFF
--- a/docs/src/main/asciidoc/qute.adoc
+++ b/docs/src/main/asciidoc/qute.adoc
@@ -81,12 +81,115 @@ $ curl -w "\n" http://localhost:8080/hello?name=Martin
 Hello Martin!
 ----
 
-== Parameter Declarations and Template Extension Methods
+== Type-safe templates
 
-Qute has many useful features.
-In this example, we'll demonstrate two of them.
+There's an alternate way to declare your templates in your Java code, which relies on the following convention:
+
+- Organise your template files in the `/src/main/resources/templates` directory, by grouping them into one directory per resource class. So, if
+  your `ItemResource` class references two templates `hello` and `goodbye`, place them at `/src/main/resources/templates/ItemResource/hello.txt`
+  and `/src/main/resources/templates/ItemResource/goodbye.txt`. Grouping templates per resource class makes it easier to navigate to them.
+- In each of your resource class, declare a `@CheckedTemplate static class Template {}` class within your resource class.
+- Declare one `public static native TemplateInstance method();` per template file for your resource.
+- Use those static methods to build your template instances.
+
+Here's the previous example, rewritten using this style:
+
+We'll start with a very simple template:
+
+.HelloResource/hello.txt
+[source]
+----
+Hello {name}! <1>
+----
+<1> `{name}` is a value expression that is evaluated when the template is rendered.
+
+Now let's declare and use those templates in the resource class.
+
+.HelloResource.java
+[source,java]
+----
+package org.acme.quarkus.sample;
+
+import javax.inject.Inject;
+import javax.ws.rs.GET;
+import javax.ws.rs.Path;
+import javax.ws.rs.QueryParam;
+
+import io.quarkus.qute.TemplateInstance;
+import io.quarkus.qute.api.CheckedTemplate;
+
+@Path("hello")
+public class HelloResource {
+
+    @CheckedTemplate
+    class Templates {
+        public static native TemplateInstance hello(); <1>
+    }
+
+    @GET
+    @Produces(MediaType.TEXT_PLAIN)
+    public TemplateInstance get(@QueryParam("name") String name) {
+        return Templates.hello().data("name", name); <2> <3>
+    }
+}
+----
+<1> This declares a template with path `templates/HelloResource/hello.txt`.
+<2> `Templates.hello()` returns a new template instance that can be customized before the actual rendering is triggered. In this case, we put the name value under the key `name`. The data map is accessible during rendering. 
+<3> Note that we don't trigger the rendering - this is done automatically by a special `ContainerResponseFilter` implementation.
+ 
+NOTE: Once you have declared a `@CheckedTemplate` class, we will check that all its methods point to existing templates, so if you try to use a template
+from your Java code and you forgot to add it, we will let you know at build time :)
+
+Keep in mind this style of declaration allows you to reference templates declared in other resources too:
+
+.HelloResource.java
+[source,java]
+----
+package org.acme.quarkus.sample;
+
+import javax.inject.Inject;
+import javax.ws.rs.GET;
+import javax.ws.rs.Path;
+import javax.ws.rs.QueryParam;
+
+import io.quarkus.qute.TemplateInstance;
+
+@Path("goodbye")
+public class GoodbyeResource {
+
+    @GET
+    @Produces(MediaType.TEXT_PLAIN)
+    public TemplateInstance get(@QueryParam("name") String name) {
+        return HelloResource.Templates.hello().data("name", name);
+    }
+}
+----
+
+=== Toplevel type-safe templates
+
+Naturally, if you want to declare templates at the toplevel, directly in `/src/main/resources/templates/hello.txt`, for example,
+you can declare them in a toplevel (non-nested) `Templates` class:
+
+.HelloResource.java
+[source,java]
+----
+package org.acme.quarkus.sample;
+
+import io.quarkus.qute.TemplateInstance;
+import io.quarkus.qute.Template;
+import io.quarkus.qute.api.CheckedTemplate;
+
+@CheckedTemplate
+public class Templates {
+    public static native TemplateInstance hello(); <1>
+}
+----
+<1> This declares a template with path `templates/hello.txt`.
+
+
+== Template Parameter Declarations 
+
 If you declare a *parameter declaration* in a template then Qute attempts to validate all expressions that reference this parameter and if an incorrect expression is found the build fails.
-*Template extension methods* are used to extend the set of accessible properties of data objects.
 
 Let's suppose we have a simple class like this:
 
@@ -99,9 +202,69 @@ public class Item {
 }
 ----
 
-And we'd like to render a simple HTML page that contains the item name, price and also a discounted price.
-The discounted price is sometimes called a "computed property".
-We will implement a template extension method to render this property easily.
+And we'd like to render a simple HTML page that contains the item name and price.
+
+Let's start again with the template:
+
+.ItemResource/item.html
+[source,html]
+----
+<!DOCTYPE html>
+<html>
+<head>
+<meta charset="UTF-8">
+<title>{item.name}</title> <1>
+</head>
+<body>
+    <h1>{item.name}</h1>
+    <div>Price: {item.price}</div> <2> 
+</body>
+</html>
+----
+<1> This expression is validated. Try to change the expression to `{item.nonSense}` and the build should fail.
+<2> This is also validated.
+
+Finally, let's create a resource class with type-safe templates:
+
+.ItemResource.java
+[source,java]
+----
+package org.acme.quarkus.sample;
+
+import javax.inject.Inject;
+import javax.ws.rs.GET;
+import javax.ws.rs.Path;
+import javax.ws.rs.PathParam;
+import javax.ws.rs.Produces;
+import javax.ws.rs.core.MediaType;
+
+import io.quarkus.qute.TemplateInstance;
+import io.quarkus.qute.Template;
+import io.quarkus.qute.api.CheckedTemplate;
+
+@Path("item")
+public class ItemResource {
+
+    @CheckedTemplate
+    class Templates {
+        public static native TemplateInstance item(Item item); <1>
+    }
+
+    @GET
+    @Path("{id}")
+    @Produces(MediaType.TEXT_HTML)
+    public TemplateInstance get(@PathParam("id") Integer id) {
+        return Templates.item(service.findItem(id)); <2>
+    }
+}
+----
+<1> Declare a method that gives us a `TemplateInstance` for `templates/ItemResource/item.html` and declare its `Item item` parameter so we can validate the template.
+<2> Make the `Item` object accessible in the template.
+
+=== Template parameter declaration inside the template itself
+
+Alternatively, you can declare your template parameters in the template file itself.
+
 Let's start again with the template:
 
 .item.html
@@ -117,16 +280,11 @@ Let's start again with the template:
 <body>
     <h1>{item.name}</h1>
     <div>Price: {item.price}</div> 
-    {#if item.price > 100} <3>
-    <div>Discounted Price: {item.discountedPrice}</div> <4>
-    {/if}
 </body>
 </html>
 ----
 <1> Optional parameter declaration. Qute attempts to validate all expressions that reference the parameter `item`.
 <2> This expression is validated. Try to change the expression to `{item.nonSense}` and the build should fail.
-<3> `if` is a basic control flow section.
-<4> This expression is also validated against the `Item` class and obviously there is no such property declared. However, there is a template extension method declared on the `ItemResource` class - see below.
 
 Finally, let's create a resource class.
 
@@ -160,16 +318,66 @@ public class ItemResource {
     public TemplateInstance get(@PathParam("id") Integer id) {
         return item.data("item", service.findItem(id)); <2>
     }
-
-    @TemplateExtension <3>
-    static BigDecimal discountedPrice(Item item) {
-        return item.price.multiply(new BigDecimal("0.9"));
-    }
 }
 ----
 <1> Inject the template with path `templates/item.html`.
 <2> Make the `Item` object accessible in the template.
-<3> A static template extension method can be used to add "computed properties" to a data class. The class of the first parameter is used to match the base object and the method name is used to match the property name.
+
+== Template Extension Methods
+
+*Template extension methods* are used to extend the set of accessible properties of data objects.
+
+Sometimes, you're not in control of the classes that you want to use in your template, and you cannot add methods
+to them. Template extension methods allows you to declare new method for those classes that will be available
+from your templates just as if they belonged to the target class.
+
+Let's keep extending on our simple HTML page that contains the item name, price and add a discounted price.
+The discounted price is sometimes called a "computed property".
+We will implement a template extension method to render this property easily.
+Let's update our template:
+
+.HelloResource/item.html
+[source,html]
+----
+<!DOCTYPE html>
+<html>
+<head>
+<meta charset="UTF-8">
+<title>{item.name}</title>
+</head>
+<body>
+    <h1>{item.name}</h1>
+    <div>Price: {item.price}</div> 
+    {#if item.price > 100} <1>
+    <div>Discounted Price: {item.discountedPrice}</div> <2>
+    {/if}
+</body>
+</html>
+----
+<1> `if` is a basic control flow section.
+<2> This expression is also validated against the `Item` class and obviously there is no such property declared. However, there is a template extension method declared on the `TemplateExtensions` class - see below.
+
+Finally, let's create a class where we put all our extension methods:
+
+.TemplateExtensions.java
+[source,java]
+----
+package org.acme.quarkus.sample;
+
+import io.quarkus.qute.TemplateExtension;
+
+@TemplateExtension
+public class TemplateExtensions {
+
+    public static BigDecimal discountedPrice(Item item) { <1>
+        return item.price.multiply(new BigDecimal("0.9"));
+    }
+}
+----
+<1> A static template extension method can be used to add "computed properties" to a data class. The class of the first parameter is used to match the base object and the method name is used to match the property name.
+
+NOTE: you can place template extension methods in every class if you annotate them with `@TemplateExtension` but we advise to keep them either
+grouped by target type, or in a single `TemplateExtensions` class by convention.
 
 == Rendering Periodic Reports
 
--- a/extensions/qute/runtime/src/main/java/io/quarkus/qute/api/CheckedTemplate.java
+++ b/extensions/qute/runtime/src/main/java/io/quarkus/qute/api/CheckedTemplate.java
@@ -1,0 +1,59 @@
+package io.quarkus.qute.api;
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import io.quarkus.qute.TemplateInstance;
+
+/**
+ * <p>
+ * If you place this annotation on a class, all its <code>native static</code> methods will be used to declare
+ * templates and the list of parameters they require.
+ * <p>
+ * If this is placed on an inner class of the class <code>X</code>, a <code>native static</code> method of the name
+ * <code>foo</code> will refer to a template at the path <code>X/foo</code> (template file extensions are
+ * not part of the method name) relative to your templates root.
+ * <p>
+ * If this is placed on a toplevel class, a <code>native static</code> method of the name
+ * <code>foo</code> will refer to a template at the path <code>foo</code> (template file extensions are
+ * not part of the method name) at the toplevel of your templates root.
+ * <p>
+ * Each parameter of the <code>native static</code> will be used to validate the template at build time, to
+ * make sure that those parameters are used properly in a type-safe manner. The return type of each
+ * <code>native static</code> method should be {@link TemplateInstance}.
+ * <p>
+ * Example:
+ * <p>
+ * 
+ * <pre>
+ * {
+ *     &#64;code
+ *     &#64;Path("item")
+ *     public class ItemResource {
+ * 
+ *         &#64;CheckedTemplate
+ *         class Templates {
+ *             // defines a template at ItemResource/item, taking an Item parameter named item
+ *             public static native TemplateInstance item(Item item);
+ *         }
+ * 
+ *         &#64;GET
+ *         &#64;Path("{id}")
+ *         &#64;Produces(MediaType.TEXT_HTML)
+ *         public TemplateInstance get(@PathParam("id") Integer id) {
+ *             // instantiate that template and pass it the required template parameter
+ *             return Templates.item(service.findItem(id));
+ *         }
+ *     }
+ * }
+ * </pre>
+ */
+@Documented
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.TYPE)
+public @interface CheckedTemplate {
+
+}
--- a/core/deployment/src/main/java/io/quarkus/deployment/util/AsmUtil.java
+++ b/core/deployment/src/main/java/io/quarkus/deployment/util/AsmUtil.java
@@ -542,4 +542,17 @@ public class AsmUtil {
         return 1;
     }
 
+    /**
+     * Prints the value pushed on the stack (must be an Object) by the given <tt>valuePusher</tt>
+     * to STDERR.
+     * 
+     * @param mv The MethodVisitor to forward printing to.
+     * @param valuePusher The function to invoke to push an Object to print on the stack.
+     */
+    public static void printValueOnStderr(MethodVisitor mv, Runnable valuePusher) {
+        mv.visitFieldInsn(Opcodes.GETSTATIC, "java/lang/System", "err", "Ljava/io/PrintStream;");
+        valuePusher.run();
+        mv.visitMethodInsn(Opcodes.INVOKEVIRTUAL, "java/io/PrintStream", "println",
+                "(Ljava/lang/Object;)V", false);
+    }
 }
--- a/core/deployment/src/main/java/io/quarkus/deployment/util/JandexUtil.java
+++ b/core/deployment/src/main/java/io/quarkus/deployment/util/JandexUtil.java
@@ -318,4 +318,30 @@ public final class JandexUtil {
         }
         return isSubclassOf(index, superClass, parentName);
     }
+
+    @SuppressWarnings("incomplete-switch")
+    public static String getBoxedTypeName(Type type) {
+        switch (type.kind()) {
+            case PRIMITIVE:
+                switch (type.asPrimitiveType().primitive()) {
+                    case BOOLEAN:
+                        return "java.lang.Boolean";
+                    case BYTE:
+                        return "java.lang.Byte";
+                    case CHAR:
+                        return "java.lang.Character";
+                    case DOUBLE:
+                        return "java.lang.Double";
+                    case FLOAT:
+                        return "java.lang.Float";
+                    case INT:
+                        return "java.lang.Integer";
+                    case LONG:
+                        return "java.lang.Long";
+                    case SHORT:
+                        return "java.lang.Short";
+                }
+        }
+        return type.toString();
+    }
 }
--- a/docs/src/main/asciidoc/mailer.adoc
+++ b/docs/src/main/asciidoc/mailer.adoc
@@ -232,6 +232,39 @@ It's also possible to inject a mail template, where the message body is created 
 
 [source, java]
 ----
+@Path("")
+public class MailingResource {
+
+    @CheckedTemplate
+    class Templates {
+        public static native MailTemplateInstance hello(String name); <1>
+    }
+    
+    @GET
+    @Path("/mail")
+    public CompletionStage<Response> send() {
+        // the template looks like: Hello {name}! <2>
+        return Templates.hello("John")
+           .to("to@acme.org") <3>
+           .subject("Hello from Qute template")
+           .send() <4>
+           .subscribeAsCompletionStage()
+           .thenApply(x -> Response.accepted().build());
+    }
+}
+----
+<1> By convention, the enclosing class name and method names are used to locate the template. In this particular case, 
+we will use the `MailingResource/hello.html` and `MailingResource/hello.txt` templates to create the message body.
+<2> Set the data used in the template.
+<3> Create a mail template instance and set the recipient.
+<4> `MailTemplate.send()` triggers the rendering and, once finished, sends the e-mail via a `Mailer` instance. 
+
+TIP: Injected mail templates are validated during build. If there is no matching template in `src/main/resources/templates` the build fails.
+
+You can also do this without type-safe templates:
+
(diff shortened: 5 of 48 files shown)


FroMage:
Forgot @gavinking :)

FroMage:
Now with 100% more correct formatting.

FroMage:
Noone? @mkouba @maxandersen @emmanuelbernard @cescoffier @nicmarti

mkouba:
@FroMage I'll take a look later this week. Anyway, I don't like the fact that the template itself will not declare the type info...

maxandersen:
> @FroMage I'll take a look later this week. Anyway, I don't like the fact that the template itself will not declare the type info...

As I grok this, the declaration of type info in templates are still optional as it is today thus I see this as just a java compatible way of declaring the typeinfo and if you *also* have typeinfo in the template and they are incompatible I would expect a failure.

also this feels like something that would fit nicely with java 14 records. 

@FroMage remind me again why we use "extend CheckedTemplateInstance" and not a annotation ?

FroMage:
> @FroMage remind me again why we use "extend CheckedTemplateInstance" and not a annotation ?

Because we need to be able to return an instance of this type that can later be checked by the renderer filter to see if it can render it. Although I could inject the supertype based on an annotation. 

So yeah, I could do this with an annotation too, I'm just too used to RAILS working with subtyping. Do you think it'd be better?

FroMage:
> @FroMage I'll take a look later this week. Anyway, I don't like the fact that the template itself will not declare the type info...

Oh come on, to be fair, you don't like the type info at all ;)

mkouba:
> Oh come on, to be fair, you don't like the type info at all ;)

I got used to it ;-).

I have few questions...

1. `ResteasyTemplate.data("name", name)` looks good but the path `{className}/{methodName}` would not work once there are multiple resource classes with the same name which is legal although not very common.
2. The `CheckedTemplateInstance` approach looks also good but there are few things I'm not excited about. The template itself does not contain the type info which is imho not a win, esp. for someone working on the template side. Also the fact that one has to create an empty public constructor seems  weird to me. Not to mention that a class name starting with lowercase hurts my eyes!

I'll take a look at the implementation and will try to play with the API. It could be that my concerns are just unfounded...

maxandersen:
> So yeah, I could do this with an annotation too, I'm just too used to RAILS working with subtyping. Do you think it'd be better?

I like ducktyping from rails too :) so at least having annotation as an option would let me put it on something that is not necessarily a CheckedTemplateInstance but simply just any bean with a proper constructor and bean/field accessors.

FroMage:
> ResteasyTemplate.data("name", name) looks good but the path {className}/{methodName} would not work once there are multiple resource classes with the same name which is legal although not very common.

Yeah, but we don't want to encourage that, or support FQDN paths. We should just tell people to use a single `.controller` package by convention and detect clashes at build time.

> The CheckedTemplateInstance approach looks also good but there are few things I'm not excited about. The template itself does not contain the type info which is imho not a win, esp. for someone working on the template side. Also the fact that one has to create an empty public constructor seems weird to me. Not to mention that a class name starting with lowercase hurts my eyes!

I'm not sure about the value of having the type info itself in the template, unless it can be used by the IDE (though I'm sure the IDE can use the type info from the type class too). Plus it's super hard to write since the IDE doesn't help you and you have to type the FQN.

The empty constructor is the only type-safe trick I could think of. If you have better I'll take it.

As for the lowercase class name: 
thuglife

More seriously, we can make them uppercase if you want, but I found lowercase to be more intuitive when it came to making it clear they were method-related.

FroMage:
> having annotation as an option would let me put it on something that is not necessarily a CheckedTemplateInstance but simply just any bean with a proper constructor and bean/field accessors.

Why would you need to do that? Those constructors and accessors would never be used. The class is just a placeholder, the only kind we could think of to declare a list of template parameters in Java.

mkouba:
> We should just tell people to use a single .controller package by convention and detect clashes at build time.

+1

> I'm not sure about the value of having the type info itself in the template...

The value is that once you open a template the presence of a type info is a hint that tells you whether you need to care about about types or not.

> The empty constructor is the only type-safe trick I could think of. If you have better I'll take it.

I have none ;-).

> More seriously, we can make them uppercase if you want...

Well, that's a matter of taste and users can choose their own convention.

BTW I also think we don't need annotation for this.

FroMage:
> Well, that's a matter of taste and users can choose their own convention.

Litteraly the ONLY thing that makes Convention-Over-Configuration work is that users don't choose their own conventions ;)

mkouba:
@FroMage Yeah, sorry "convention" was a poor choice. It should be "style" or something like that... ;-). I know that you'd like to avoid any name transformations so that it's 1:1 but this would only work for simple template names like `hello`. But what about `item-detail.html`, `footer-old-nobanner.html`, etc. For these cases `itemDetail/ItemDetail` -> `item-detail` would be probably more appropriate...

FroMage:
Sure, we can support that. But I'm not sure we need to. I'd rather tell people to name their files `itemDetail.html`.

mkouba:
Fair enough. From what I see the `CheckedTemplateInstance` approach does not work outside a JAX-RS resource? Is it intentional? I mean it could be useful even without JAX-RS. Just declare a top-level class that extends the `CheckedTemplateInstance` etc?

FroMage:
We could make it work outside a jax-rs resource, indeed.

maxandersen:
> itemDetail.html

we use kebab case everywhere else (except java class names); why are we suddenly wanting users to use camel case in file names ?

maxandersen:
> I'm not sure about the value of having the type info itself in the template, unless it can be used by the IDE (though I'm sure the IDE can use the type info from the type class too). Plus it's super hard to write since the IDE doesn't help you and you have to type the FQN.

The IDE can be enhanced to give content assist and verification when its in the template - just saying.

stuartwdouglas:
Just throwing this out there, but another approach you could use is annotated interfaces.

For a templates in `templates/MyTemplates/hello.txt`

```
@TemplateController
class MyTemplates {
  TemplateInstance hello(String name);
}
```

You can then just inject this template controller anywhere you want, and use it in a type safe manner.

FroMage:
> You can then just inject this template controller anywhere you want, and use it in a type safe manner.

This does replace one injection point per method with one per class, but I like the idea, and it's led me to come up with another idea that does away with injection altogether. I'm pretty sure we can use the `native` trick to do this:

```java
@Path("hello")
public class HelloResource {
    private native TemplateInstance renderHello(String name);
    private native TemplateInstance renderGoodbye(String name);

    @GET
    public TemplateInstance hello(@QueryParam String name) {
        return renderHello(name); // this is type-safe
    }

    @Path("bye")
    @GET
    public TemplateInstance goodbye(@QueryParam String name) {
        return renderGoodbye(name); // this is type-safe
    }
}
```

It's shorter than my constructors proposal and doesn't require injection. I think they can even be `static` to allow usage from other classes.

mkouba:
How does this `native` trick work? In any case, I believe we should try to to avoid any tricks if possible...

> This does replace one injection point per method with one per class...

I like Stuart's proposal. The "controller" could be used for all templates in your app. We should at least explore this possibility.

@FroMage BTW the methods in your `HelloResource` should return `CheckedTemplateInstance` because it does not implement `TemplateInstance`.

FroMage:
> How does this native trick work? 

It works by us replacing the bytecode of the method with a normal method. It's pretty much the only way to declare a method without a body unless it's an abstract class or interface. I think @gunnarmorling showed it to me, IIRC.

> In any case, I believe we should try to to avoid any tricks if possible...

Well, the entire Panache APIs are based on using any trick necessary to make UX better, so I disagree :)

> BTW the methods in your HelloResource should return CheckedTemplateInstance because it does not implement TemplateInstance.

Ah yes.

mkouba:
> Well, the entire Panache APIs are based on using any trick necessary to make UX better...

I disagree. I don't think that Panache users are not encouraged to use `native` which is a keyword with a well-defined meaning AFAIK. Less code does not always mean better.

gunnarmorling:
> I think @gunnarmorling showed it to me, IIRC.

I wish I could claim ownership, but it wasn't my idea :) Might be an interesting effort though to make it a small JEP to have an official keyword for body-less methods in non-abstract classes?

emmanuelbernard:
It was @gavinking or I that proposed `native` if I recall (which I clearly don't so this message is just waste of bandwidth).

gavinking:
So this looks very reasonable but let me add some questions/comments:

- I suppose (or at least hope) that it's not necessary to declare it as an inner class.
- Are these things composable? (Even if not now, at least in principle.) I'm thinking about template composition, and being able to compose the template parameter classes. What would that look like?
- Is it really necessary to extend `CheckedTemplateInstance` here? Eyeballing that class, it doesn't really seem to do very much, but I might be missing what restrictions Resteasy imposes on the `@GET` method return type.
- Whatever, I'm not a huge fan of long fairly technical names like `CheckedTemplateInstance` and `ResteasyTemplate` in vanilla user code. Would it be possibly to come up with something a bit snappier here?

WDYT, @FroMage?

gavinking:
> I'm not sure about the value of having the type info itself in the template, unless it can be used by the IDE (though I'm sure the IDE can use the type info from the type class too). Plus it's super hard to write since the IDE doesn't help you and you have to type the FQN.

See what I would have done here is declare the template and the template class in the same location right next to each other. (Yes yes I know this is contrary to the modern convention of keeping different kinds of files under different source roots, but that's a convention I've never particularly loved.)

I sort of agree with @mkouba that if the template class is an inner class of the resource class (which I can see makes lots of sense for other reasons) then it splits the declaration or the template parameters to a totally different location in the source tree, which makes maintenance potentially difficult.

I change my template and then have to go hunting for the template class. Indeed, I might not even know that there *is* a template class. Now, hopefully the build phase type checking will catch any problem there, however with type safety being optional, that's not guaranteed.

gavinking:
Note that from what I see here, and assuming I understand correctly how this works, there's no link to the template itself in the template class, nor any link to the template class in the template, making the typechecker *entirely* dependent upon conventions, if I understand correctly. This worries me a bit from a tooling and maintenance programming perspective.

FroMage:
> I suppose (or at least hope) that it's not necessary to declare it as an inner class.

They could be toplevel too, if we find a way to specify what resource/folder they belong to. In all my experience you mostly need one template per method, and grouping them in a folder which represents the class is a nice way to organise them.

If we put them as toplevel classes we need either an annotation to represent where the template is stored, or another convention. The RoR convention of `templates/ClassName/method.ext` for templates is so good I can't come up with anything better to organise templates and map them to where they're used.

> Are these things composable? (Even if not now, at least in principle.) I'm thinking about template composition, and being able to compose the template parameter classes. What would that look like?

That is a very good question. AFAICT composition works like this:

```
// base.html
<html>
<head>
<meta charset="UTF-8">
<title>{#insert title}Default Title{/}</title> 
</head>
<body>
  {#insert body}No body!{/} 
</body>
</html>

// method.html
{#include base} 
  {#title}My Title{/title} 
  {#body}
    <div>
      My body.
    </div>
  {/body}
{/include}
```

So when I render `method.html` I specify typed params, but inside it, it will invoke the `base.html` template but can only pass `String` parameters, AFAICT. I don't really know if the parameters passed to `method.html` (named and typed) can be used directly from within `base.html`. If we consider `#include` to be a preprocessor macro like in C, I'd say yes. Not sure what the docs say about this. @mkouba ?

> Is it really necessary to extend CheckedTemplateInstance here? Eyeballing that class, it doesn't really seem to do very much, but I might be missing what restrictions Resteasy imposes on the @GET method return type.

Well, that's a convention: this is how you declare typed templates. We could alternately annotate it, but since I need that inner class to have some methods that are generated at build-time and I need to be able to invoke them in the response renderer, this seemed like a fine convention. Otherwise I can make it implement an interface in the processor, but we still need a way to mark this as a template declaration.

> Whatever, I'm not a huge fan of long fairly technical names like CheckedTemplateInstance and ResteasyTemplate in vanilla user code. Would it be possibly to come up with something a bit snappier here?

Bikeshedding is for last ;) I'll propose `BetterTemplate`…

Do note that we're also discussing another convention in the form of native methods. See proposal in https://github.com/quarkusio/quarkus/pull/7733#issuecomment-606498220

FroMage:
> See what I would have done here is declare the template and the template class in the same location right next to each other. (Yes yes I know this is contrary to the modern convention of keeping different kinds of files under different source roots, but that's a convention I've never particularly loved.)

> I sort of agree with @mkouba that if the template class is an inner class of the resource class (which I can see makes lots of sense for other reasons) then it splits the declaration or the template parameters to a totally different location in the source tree, which makes maintenance potentially difficult.

We could add source files in the resource folder, at least my IDE doesn't complain. But it kinda goes against the `native` proposal which has shorter syntax. We could mandate a grouped declaration such as:

```java
public class FolderName {
    public native TemplateInstance renderHello(String name);
    public native TemplateInstance renderGoodbye(String name);
}
```

In each `template/ClassName` folder to contain all templates.

> I change my template and then have to go hunting for the template class. Indeed, I might not even know that there is a template class. Now, hopefully the build phase type checking will catch any problem there, however with type safety being optional, that's not guaranteed.

Well, if you have a type-safe declaration for your template you will get an error if they're out of sync. If you don't, well there's no problem.

FroMage:
> Note that from what I see here, and assuming I understand correctly how this works, there's no link to the template itself in the template class, nor any link to the template class in the template, making the typechecker entirely dependent upon conventions, if I understand correctly. This worries me a bit from a tooling and maintenance programming perspective.

> @FroMage precisely what is the rule that allows the type checker to correlate these two artifacts at build time?

Right now, it's `template/ClassName/methodName.ext` for a template which maps to `pkg.ClassName.methodName` in Java.

gavinking:
> Well, if you have a type-safe declaration for your template you will get an error if they're out of sync. If you don't, well there's no problem.

I guess it's OK as long as you produce an error any time there is a subclass of `CheckedTemplateInstance` that doesn't have a corresponding template.

gavinking:
> If we put them as toplevel classes we need either an annotation to represent where the template is stored, or another convention.

I sorta feel like having this as an option would make me feel a bit safer.

gavinking:
> We could alternately annotate it

An annotation would make it a little easier to compose these things in more complex cases. And since you're planning on bytecode processing these things *anyway*, I'm inclined to say that an annotation is a better choice.

mkouba:
So I'd like to move this PR on but I'm a little bit lost in the comments ;-).

I'm still concerned about using `native` methods as a part of the "API". But if everyone else is OK I can live with this.

I like the idea of a "template controller" interface. But we could explore the idea later.

@FroMage I think it's time to get this PR into the "ready for review" state, i.e. polish the code, tests, docs, etc.

FroMage:
Well I have to implement the `native` option so we can look at it and compare both approaches.

FroMage:
Will try this week.

FroMage:
I pushed a new version with support for `native` declarations.

I tried putting classes next to the templates in `src/main/resources` and it didn't work: my IDE didn't pick them up as classes, they were not visible in the controller class, so I couldn't call them.

So grouping them next to the templates didn't work.

Here's what currently works:

```java
@Path("hello")
public class HelloResource {

    @CheckedTemplate
    public static class Templates {
        public static native TemplateInstance typeError3(String name);
        public static native TemplateInstance typedTemplate(String name, Map<String, Object> other);
    }

    @Path("native/type-error3")
    @GET
    public TemplateInstance nativeTypeError3() {
        return Templates.typeError3("world");
    }

    @Path("native/typed-template")
    @GET
    public TemplateInstance nativeTypedTemplate(@QueryParam String name) {
        if (name == null) {
            name = "world";
        }
        return Templates.typedTemplate(name, null);
    }
}
```

Note that I _had_ to group them into an inner class otherwise their names would conflict with the controller method names that use them.

The previous alternative was:

```java
@Path("hello")
public class HelloResource {

    public static class typeError2 extends CheckedTemplateInstance {
        public typeError2(String name) {
        }
    }

    public static class typedTemplate extends CheckedTemplateInstance {
        public typedTemplate(String name, Map<String, Object> other) {
        }
    }

    @Path("type-error2")
    @GET
    public CheckedTemplateInstance typeError2() {
        return new typeError2("world");
    }

    @Path("typed-template")
    @GET
    public CheckedTemplateInstance typedTemplate(@QueryParam String name) {
        if (name == null) {
            name = "world";
        }
        return new typedTemplate(name, null);
    }

}
```

Or if we go via declaration with an annotation:

```java
@Path("hello")
public class HelloResource {

    @TypedTemplate
    public static class typeError2 {
        public typeError2(String name) {
        }
    }

    @TypedTemplate
    public static class typedTemplate {
        public typedTemplate(String name, Map<String, Object> other) {
        }
    }

    @Path("type-error2")
    @GET
    public typeError2 typeError2() {
        return new typeError2("world");
    }

    @Path("typed-template")
    @GET
    public typedTemplate typedTemplate(@QueryParam String name) {
        if (name == null) {
            name = "world";
        }
        return new typedTemplate(name, null);
    }

}
```

Overall, the `native` option is not much less verbose, because of those annoying `static native` modifiers to repeat, but I think I prefer it.

This would work for emails too (as soon I support return types of `MailTemplateInstance`), where the `Templates` class would be in the email class (I like to group all email sending to a single class):

```java
public class MyMails {
    public static final String FROM = "stef@example.com";

    @CheckedTemplate
    public static class Templates {
        public static native MailTemplateInstance hello(String name);
    }

  public static CompletionStage<Void> hello(String to, String name) {
    return Templates.hello(name).to(to).from(FROM).send();
  }
}
```

This would work for toplevel templates too (not implemented ATM but would be trivial) and would allow us to target `templates/hello.txt` files (not inside a class/subfolder).

Now, I should be able to not require the `@CheckedTemplate` annotation on `Templates` by scanning every JAX-RS resource and attempting to look up an inner `Templates` class by convention.

So, WDYT?

Note that the implementation is not finished, there's still issues to resolve wrt primitives and boxing.

mkouba:
It looks better but I still don't see the point of using `static native` + the need for having the resource name in the template path, e.g. `templates/HelloResource/typeError3.txt`. I'd rather use @stuartwdouglas's proposal with `@TemplateController` interfaces:

```java
@Path("hello")
public class HelloResource {
    
    // Could be placed anywhere in the app, determines the template location, e.g. templates/Templates/typeError3.txt but could be customizable
    @TemplateController
    public interface Templates {
        TemplateInstance typeError3(String name);
        TemplateInstance typedTemplate(String name, Map<String, Object> other);
    }

    // If you really hate injection we could introduce a helper class, i.e. Controller.for(Templates.class).typeError3("world");
    @Inject
    Templates templates;

    @Path("native/type-error3")
    @GET
    public TemplateInstance nativeTypeError3() {
        return templates.typeError3("world");
    }

    @Path("native/typed-template")
    @GET
    public TemplateInstance nativeTypedTemplate(@QueryParam String name) {
        if (name == null) {
            name = "world";
        }
        return templates.typedTemplate(name, null);
    }
}
```

FroMage:
> the need for having the resource name in the template path, e.g. `templates/HelloResource/typeError3.txt`
> templates/Templates/typeError3.txt

Because in template-driven projects we end up with a lot of templates:

- https://github.com/FroMage/RivieraDEV/tree/master/app/views
- https://github.com/eclipse/ceylon-herd/tree/master/app/views

And grouping them by controller is super useful and what makes the most sense, since the controller methods themselves are grouped by class.

As for using `interface` + injection vs `static native` + `Controller.for(Templates.class)`, it's certainly debatable, but I still feel like it flows much better without having to jump through hoops when using templates: so make it slightly longer to declare, and much easier to use.

As for getting rid of the `Template` class holder, we can, but due to methods using the same name as the controller that use them, we would need to prefix every method with `render` (or something else):

```java
@Path("hello")
@CheckedTemplates
public class HelloResource {

    public static native TemplateInstance renderTypeError3(String name);
    public static native TemplateInstance renderTypedTemplate(String name, Map<String, Object> other);

    @Path("native/type-error3")
    @GET
    public TemplateInstance nativeTypeError3() {
        return renterTypeError3("world");
    }

    @Path("native/typed-template")
    @GET
    public TemplateInstance nativeTypedTemplate(@QueryParam String name) {
        if (name == null) {
            name = "world";
        }
        return renderTypedTemplate(name, null);
    }
}
```

Probably this is the less boiler-platy, but I'm not sure it's worth going that far, I kinda like them grouped.

mkouba:
I like them grouped as well and I don't dispute the existence of a "template controller class/interface". On the other hand, I'd prefer using interface to class with native method hacks. TBH I see no advantage in using the native methods. It's not clearer or less verbose and I can imagine users adding static methods without `native` with empty bodies complaining that it does not work.

FroMage:
> I can imagine users adding static methods without native with empty bodies complaining that it does not work.

We could totally support that and make it work, except the body would not be empty because you need to return an instance, so I doubt users will try that. `native` is only used to avoid specifying a body.

mkouba:
> native is only used to avoid specifying a body

I know what it's used for.

FroMage:
Did you close it by accident?

mkouba:
@FroMage yeah, sorry! ;-)

mkouba:
@FroMage I think that need to decide what to do with this PR.

I suggest the following:
1. Pls rebase and get this PR into the "ready for review" state
2. So that we can merge the PR, play with the API and eventually improve things later

FroMage:
I can do that.

stuartwdouglas:
Should we start quarkus-dev discussion around this so it gets more visibility?

FroMage:
I've rebased it.

FroMage:
There are still things to be discussed and settled:

- `native` or constructors? I think `native` scales better.
- If `native`, what annotation name? `@CheckedTemplate`? Convention?

Things still to be done:

- Test with primitives/boxing
- Mail support
- Docs

stuartwdouglas:
I personally prefer the interface approach, as it seems to adhere more to the 'single responsibility' principle. With the native approach your examples have the classes being both REST endpoints and template controllers. It also avoids the native trick. On the flip side I guess the static approach is more similar to what we already do in Panache.

FroMage:
> I personally prefer the interface approach, as it seems to adhere more to the 'single responsibility' principle. With the native approach your examples have the classes being both REST endpoints and template controllers. It also avoids the native trick. On the flip side I guess the static approach is more similar to what we already do in Panache.

In my experience, when building a site with templates, each endpoint method has its own view, so defining it anywhere else than endpoint itself is a waste of time, hence the statics, which still allow for them to be used somewhere else if ever needed for template reuse.

Now, like we do in Data with Panache, we could just support both forms (static and interface) and leave it to users to use the form they prefer. But I'm afraid they'll start putting those interfaces away from the endpoint, which is less good for maintenance: it's already far from the template itself, because we could not find any way to put it closer (technical limitations), so if we put it far from each template's (almost always) single use-site it becomes harder to maintain.

But again, we should probably support both use-cases if people are so set in the interface/field way that this would bother them.

stuartwdouglas:
Maybe just do the native trick first and see if there is any requests for the interface based approach.

FroMage:
New:

- Verify that native methods have at least one template variant present
- Verify that native methods are static and a valid return type
- Support native methods for emails
- Support native method template variants
- Moved Panache's JandexUtil into quarkus-core JandexUtil and (new) AsmUtil (this was required because those methods are useful everywhere, as discussed on quarkus-dev)
- Support boxing/primitives in native method parameters
- Removed the constructor variant
- Support toplevel templates

Now in theory I've only got documentation to write before this is ready.

FroMage:
I've added docs.

I'm not quite sure on which feet to dance about it, because we have conflicting opinions as to what should be the default style, the one we should start with.

ATM, I've updated the docs in this manner:

- how to use templates using `@Inject` (no parameter validation)
- how to declare templates with `@CheckedTemplate` (type-safe declaration but no parameter validation)
- how to declare type-safe parameters with `@CheckedTemplate` (type-safe declaration and type-safe parameters)
- how to declare type-safe parameters using `@Inject` and parameter declarations in the template (type-safe parameters)

As a result, I've moved the template extension docs to their own section, and added some text about the convention of grouping them in a single class, or by target.

I've also updated the mailer docs to mention the type-safe declaration first @cescoffier.

Please @mkouba @stuartwdouglas @emmanuelbernard @cescoffier can you take a look at the docs and let me know if I can improve it: I'm really not sure how to organise it so it's not overly verbose (there's a lot of overlap between the alternative approaches)?

FroMage:
Rebased due to the `JandexUtil` commit having been merged in `master`.

FroMage:
Rebased yet again.

docs/src/main/asciidoc/qute.adoc
```diff
@@ -99,9 +202,69 @@ public class Item {
...
+    @CheckedTemplate
+    class Templates {
+        public static native TemplateInstance item(Item item); <1>
+    }
+
+    @GET
+    @Path("{id}")
+    @Produces(MediaType.TEXT_HTML)
+    public TemplateInstance get(@PathParam("id") Integer id) {
+        return Templates.item("item", service.findItem(id)); <2>
```
mkouba, line 257: This should be probably `return Templates.item(service.findItem(id));`, or?

docs/src/main/asciidoc/qute.adoc
```diff
@@ -99,9 +202,69 @@ public class Item {
...
+    }
+}
+----
+<1> Declare a method that gives us a `TemplateInstance` for `templates/ItemResource/item.html` and declare its `Item item` parameter so we can validate the template.
+<2> Make the `Item` object accessible in the template.
+
+=== Template parameter declaration inside the template itself
+
+Alternately, you can declare your template parameters in the template file itself.
```
mkouba, line 266: `Alternately` -> `Alternatively`

extensions/qute/runtime/src/main/java/io/quarkus/qute/api/CheckedTemplate.java
```diff
@@ -0,0 +1,14 @@
...
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Documented
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.TYPE)
+public @interface CheckedTemplate {
```
mkouba, line 12: We should add some javadoc here...

extensions/resteasy-qute/runtime/src/main/java/io/quarkus/resteasy/qute/runtime/ResteasyTemplate.java
```diff
@@ -0,0 +1,28 @@
...
+import javax.ws.rs.container.ResourceInfo;
+
+import org.jboss.resteasy.core.ResteasyContext;
+
+import io.quarkus.arc.Arc;
+import io.quarkus.qute.Engine;
+import io.quarkus.qute.Template;
+import io.quarkus.qute.TemplateInstance;
+
+public class ResteasyTemplate {
```
mkouba, line 12: I'm not sure about the name... maybe just `RestTemplate`? IDK maybe it's OK. But it should reside in the `io.quarkus.resteasy.qute` package as it's a public API. Plus it wouldn't hurt to make it final with a private constructor.

mkouba (changes requested):
I added few comments. It looks good (although I'm still against using the `native` modifier in the API ;-). I still need to checkout the code and play with the API a little.

docs/src/main/asciidoc/qute.adoc
FroMage: Ouch.

docs/src/main/asciidoc/qute.adoc
FroMage: It does exist, but seems to be an American word. Fixed.

extensions/qute/runtime/src/main/java/io/quarkus/qute/api/CheckedTemplate.java
FroMage: Indeed.

extensions/resteasy-qute/runtime/src/main/java/io/quarkus/resteasy/qute/runtime/ResteasyTemplate.java
FroMage: OK changed.

FroMage:
All updated and rebased. Let's see what CI says.

FroMage:
Windows failure doesn't seem related:

```
2020-06-16T17:13:08.1485725Z 2020-06-16 17:13:08,136 INFO  [io.quarkus] (Quarkus Main Thread) acme 1.0-SNAPSHOT on JVM (powered by Quarkus 999-SNAPSHOT) started in 0.476s. Listening on: http://0.0.0.0:8080
2020-06-16T17:13:08.1486507Z 2020-06-16 17:13:08,137 INFO  [io.quarkus] (Quarkus Main Thread) Profile dev activated. Live Coding activated.
2020-06-16T17:13:08.1487990Z 2020-06-16 17:13:08,139 INFO  [io.quarkus] (Quarkus Main Thread) Installed features: [cdi, resteasy, servlet, undertow-websockets]
2020-06-16T17:13:08.1488358Z 2020-06-16 17:13:08,139 INFO  [io.qua.dep.dev] (vert.x-worker-thread-3) Hot replace total time: 1.059s 
2020-06-16T17:13:11.8111125Z [WARNING] An exceptionCaught() event was fired, and it reached at the tail of the pipeline. It usually means the last handler in the pipeline did not handle the exception.
2020-06-16T17:13:11.8124136Z java.io.IOException: An existing connection was forcibly closed by the remote host
2020-06-16T17:13:11.8130607Z     at sun.nio.ch.SocketDispatcher.read0 (Native Method)
2020-06-16T17:13:11.8157386Z     at sun.nio.ch.SocketDispatcher.read (SocketDispatcher.java:43)
2020-06-16T17:13:11.8180653Z     at sun.nio.ch.IOUtil.readIntoNativeBuffer (IOUtil.java:276)
2020-06-16T17:13:11.8194447Z     at sun.nio.ch.IOUtil.read (IOUtil.java:233)
2020-06-16T17:13:11.8196698Z     at sun.nio.ch.IOUtil.read (IOUtil.java:223)
2020-06-16T17:13:11.8217621Z     at sun.nio.ch.SocketChannelImpl.read (SocketChannelImpl.java:358)
2020-06-16T17:13:11.8224839Z     at io.netty.buffer.PooledByteBuf.setBytes (PooledByteBuf.java:253)
2020-06-16T17:13:11.8227337Z     at io.netty.buffer.AbstractByteBuf.writeBytes (AbstractByteBuf.java:1133)
2020-06-16T17:13:11.8237364Z     at io.netty.channel.socket.nio.NioSocketChannel.doReadBytes (NioSocketChannel.java:350)
2020-06-16T17:13:11.8265315Z     at io.netty.channel.nio.AbstractNioByteChannel$NioByteUnsafe.read (AbstractNioByteChannel.java:148)
2020-06-16T17:13:11.8285478Z     at io.netty.channel.nio.NioEventLoop.processSelectedKey (NioEventLoop.java:714)
2020-06-16T17:13:11.8287157Z     at io.netty.channel.nio.NioEventLoop.processSelectedKeysOptimized (NioEventLoop.java:650)
2020-06-16T17:13:11.8343758Z     at io.netty.channel.nio.NioEventLoop.processSelectedKeys (NioEventLoop.java:576)
2020-06-16T17:13:11.8502849Z     at io.netty.channel.nio.NioEventLoop.run (NioEventLoop.java:493)
2020-06-16T17:13:11.8505994Z     at io.netty.util.concurrent.SingleThreadEventExecutor$4.run (SingleThreadEventExecutor.java:989)
2020-06-16T17:13:11.8677402Z     at io.netty.util.internal.ThreadExecutorMap$2.run (ThreadExecutorMap.java:74)
2020-06-16T17:13:11.8679231Z     at io.netty.util.concurrent.FastThreadLocalRunnable.run (FastThreadLocalRunnable.java:30)
2020-06-16T17:13:11.9055838Z     at java.lang.Thread.run (Thread.java:834)
2020-06-16T17:13:12.5671453Z Connection closed CloseReason[1001]
```
@mkouba can you rereview and we get this PR behind us?

mkouba (approved):

mkouba:
@FroMage could you pls squash the commits? It's not possible to do "squash and merge" in the quarkus repo.

FroMage:
Done.

mkouba:
Awesome, thanks!

gsmet:
@mkouba does this one break something for the users?

mkouba:
@gsmet AFAIK it should not. Also for qute we don't care that much about compatibility because it's still an experimental feature.

PR #7410, opened 2020-02-25
enable support for AMQP in the quarkus-artemis-test module

gemmellr:
As discussed on Zulip, there is a quarkus-test-artemis module at test-framework/artemis/pom.xml for using an embedded Artemis broker in testing, but it only supports the brokers Core protocol. This PR adds the brokers AMQP protocol module also, allowing existing extensions using AMQP to utilise quarkus-test-artemis without having to do this themselves more awkwardly.

PR #7738, opened 2020-03-10
update proton-j to 0.33.3

gemmellr:
When I raised https://github.com/quarkusio/quarkus/pull/7410, I specified the version of proton-j that artemis-amqp-protocol 2.11.0 used, which is 0.33.2.

The proton-j dependency is also used by the smallrye-reactive-messaging-amqp extension (indirectly via vertx-proton), and the Qpid JMS extension. Both of those are using the current proton-j 0.33.3 in their original component releases \[1\] \[2\] so I think it makes more sense to use that here. FWIW, I made the only change contained in 0.33.3, there is little difference.

\[1\] https://github.com/vert-x3/vertx-proton/blob/3.8.5/pom.xml#L41
\[2\] https://github.com/apache/qpid-jms/blob/0.49.0/pom.xml#L42

### DIFF
--- a/bom/runtime/pom.xml
+++ b/bom/runtime/pom.xml
@@ -160,7 +160,7 @@
         <mongo-reactivestreams-client.version>1.13.0</mongo-reactivestreams-client.version>
         <mongo-crypt.version>1.0.0</mongo-crypt.version>
         <artemis.version>2.11.0</artemis.version>
-        <proton-j.version>0.33.2</proton-j.version>
+        <proton-j.version>0.33.3</proton-j.version>
         <okhttp.version>3.14.6</okhttp.version>
         <sentry.version>1.7.30</sentry.version>
         <!-- Used for integration tests, to make sure webjars work-->


gemmellr:
I ran a test backport for 1.3, everything passed:
https://github.com/gemmellr/quarkus/actions/runs/52942426

My forks CI run for master failed the Windows JDK8 JVM Tests, but seemingly due to a timeout in other tests and not due to this change.

gemmellr:
This time \[1\] the Windows JDK11 JVM Tests job passed, but the linux JDK8 JVM Tests job failed (gradle multi module build test). The failing job used precisely the same changes as the passing equivalent on my fork, comparing the checkout logs and branch commit history. Do we need to force another run?

\[1\] https://github.com/quarkusio/quarkus/actions/runs/53022952

gemmellr:
I'll just rebase on current master anyway in case the answer is yes, cant exactly hurt.

michalszynkiewicz:
@gsmet backporting to 1.3 is fine by me

gsmet (approved):

PR #7739, opened 2020-03-10
Ensure that a .env file at the root of the project can be used in dev-mode

### LINKED ISSUE #7519: .env file copy to target in dev mode
**Description**

When you are at dev mode and you have a file at the root of the project a `.env` file, the file is not loaded as the root directory at runtime is at `target/`.

A possible workaround is to use:

```
<build>
<resources>  
   <resource>
       <directory>${project.basedir}</directory><!-- from -->
       <targetPath>${project.build.directory}</targetPath><!-- to -->
       <includes><!-- what -->
          <include>**/.env</include>
       </includes>
    </resource> 
</resources>
</build>
```

But we could integrate this on dev mode so if there is a `.env` the file then we copy directly to `target` (I don't know if this is possible).

I know that the purpose of a `.env` is to reside outside of the project (JAR/native package) but I am thinking in the next use case:

I am demoing in front of a big audience, I have my `.env` file with my credentials to publish a docker image to quai. Then I can just say look I copy this file into root project as it has my credentials and of course, you cannot see them, and then I run mvn .... and the image is published.

I could copy directly to `target/` directory it is true, but if I do a clean then I will miss everything.

So it will be great at the plugin level to do this copy instead of having to configure `pom.xml` with the `resources` and explain to the audience why I am doing this.

geoand:
…-mode

This includes having the capability to update the file

Fixes: #7519

### DIFF
--- a/core/deployment/src/main/java/io/quarkus/deployment/steps/DevModeBuildStep.java
+++ b/core/deployment/src/main/java/io/quarkus/deployment/steps/DevModeBuildStep.java
@@ -1,5 +1,6 @@
 package io.quarkus.deployment.steps;
 
+import java.nio.file.Paths;
 import java.util.Arrays;
 import java.util.List;
 
@@ -11,6 +12,7 @@ class DevModeBuildStep {
     @BuildStep
     List<HotDeploymentWatchedFileBuildItem> config() {
         return Arrays.asList(new HotDeploymentWatchedFileBuildItem("META-INF/microprofile-config.properties"),
-                new HotDeploymentWatchedFileBuildItem("application.properties"));
+                new HotDeploymentWatchedFileBuildItem("application.properties"), new HotDeploymentWatchedFileBuildItem(
+                        Paths.get(".env").toAbsolutePath().toAbsolutePath().toString()));
     }
 }
--- a/core/devmode/src/main/java/io/quarkus/dev/DevModeMain.java
+++ b/core/devmode/src/main/java/io/quarkus/dev/DevModeMain.java
@@ -8,6 +8,7 @@ import java.io.InputStream;
 import java.io.ObjectInputStream;
 import java.net.URI;
 import java.net.URL;
+import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.Collections;
@@ -97,6 +98,9 @@ public class DevModeMain implements Closeable {
 
                 }
             }
+
+            copyDotEnvFile();
+
             Properties buildSystemProperties = new Properties();
             buildSystemProperties.putAll(context.getBuildSystemProperties());
             bootstrapBuilder.setBuildSystemProperties(buildSystemProperties);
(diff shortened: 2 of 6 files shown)


geoand:
This implementation is by no means perfect (for example deleting the root `.env` is not reflected in dev mode), but I think it covers almost all ways users will use the file

geoand:
CI issue fixed

core/devmode/src/main/java/io/quarkus/dev/DevModeMain.java
```diff
@@ -110,6 +114,36 @@ public void start() throws Exception {
         }
     }
 
+    // copies the .env file to the directory where the process is running
+    // this is done because the for the .env file to take effect, it needs to be
```
gsmet:
```suggestion
    // this is done because for the .env file to take effect, it needs to be
```

core/devmode/src/main/java/io/quarkus/dev/DevModeMain.java
```diff
@@ -110,6 +114,36 @@ public void start() throws Exception {
...
+            return;
+        }
+        Path currentDir = Paths.get("").toAbsolutePath().normalize();
+        Path dotEnvPath = projectDir.toPath().resolve(".env");
+        if (Files.exists(dotEnvPath)) {
+            try {
+                Path link = currentDir.resolve(".env");
+                silentDeleteFile(link);
+                // create a symlink to ensure that user updates to the file have the expected effect in dev-mode
+                Files.createSymbolicLink(link, dotEnvPath);
```
gsmet: Does it work on Windows?

gsmet (dismissed):
Added a suggestion and a question?

core/devmode/src/main/java/io/quarkus/dev/DevModeMain.java
gsmet: I suppose so as you have a test.

geoand: Yup, thankfully we have CI running on Windows as well 😎

core/devmode/src/main/java/io/quarkus/dev/DevModeMain.java
geoand: Fixed, thanks

core/deployment/src/main/java/io/quarkus/deployment/steps/DevModeBuildStep.java
```diff
@@ -11,6 +12,7 @@
     @BuildStep
     List<HotDeploymentWatchedFileBuildItem> config() {
         return Arrays.asList(new HotDeploymentWatchedFileBuildItem("META-INF/microprofile-config.properties"),
-                new HotDeploymentWatchedFileBuildItem("application.properties"));
+                new HotDeploymentWatchedFileBuildItem("application.properties"), new HotDeploymentWatchedFileBuildItem(
+                        Paths.get("").toAbsolutePath().normalize().resolve(".env").toAbsolutePath().toString()));
```
dmlloyd: This seems like it might be overkill... would something like this work?

```suggestion
                        Paths.get(".env").toAbsolutePath().toString()));
```

core/devmode/src/main/java/io/quarkus/dev/DevModeMain.java
```diff
@@ -110,6 +114,36 @@ public void start() throws Exception {
...
```
dmlloyd: This seems wrong... unless we know for a fact that the current directory is going to be `target/`?  We don't want to modify anything outside `target` for any reason.  It would be better to modify the config source impl if that's the case.

geoand: The reason I did this is because as a user I expect that if my `${rootDir}/.env` file is going to be picked up when I do `mvn quarkus:dev`, then I would also expect it to be monitored for changes. With this way of doing things, those changes are picked up.

It doesn't really have to do with `./target` at all, unless I am missing something. 

WDYT?

dmlloyd: The question is, where is this symbolic link being created, in practical terms?

geoand: Inside `target`. Which is why I mentioned that updates work, but deleting the original `.env` doesn't (although I wouldn't expect the user to do that)

dmlloyd: OK, that's what I wanted to know.  Thanks!

dmlloyd (approved):

core/deployment/src/main/java/io/quarkus/deployment/steps/DevModeBuildStep.java
geoand: It does yes, thanks

core/devmode/src/main/java/io/quarkus/dev/DevModeMain.java
geoand: :+1:

PR #7742, opened 2020-03-10
Fix links to generated config in container-image.adoc

geoand:


gsmet (approved):
Not ideal but that will do for now until we undestand what's going on.

PR #7748, opened 2020-03-10
feat(docs): do not include deprecated config items in the generated config doc

### LINKED ISSUE #7746: Avoid including the deprecated configuration properties in the generated doc
The datasource config is a good example of that: I have several deprecated properties and they appear in `quarkus-datasource.adoc`.

I would prefer them to not be included.

A good example of that is `quarkus.datasource.url`.

machi1990:
Fixes https://github.com/quarkusio/quarkus/issues/7746

/cc @gsmet

### DIFF
--- a/core/processor/src/main/java/io/quarkus/annotation/processor/generate_doc/ConfigDoItemFinder.java
+++ b/core/processor/src/main/java/io/quarkus/annotation/processor/generate_doc/ConfigDoItemFinder.java
@@ -106,9 +106,14 @@ class ConfigDoItemFinder {
 
             String hyphenatedFieldName = hyphenate(fieldName);
             String configDocMapKey = hyphenatedFieldName;
+            boolean isDeprecated = false;
 
             for (AnnotationMirror annotationMirror : annotationMirrors) {
                 String annotationName = annotationMirror.getAnnotationType().toString();
+                if (annotationName.equals(Deprecated.class.getName())) {
+                    isDeprecated = true;
+                    break;
+                }
                 if (annotationName.equals(Constants.ANNOTATION_CONFIG_ITEM)
                         || annotationName.equals(Constants.ANNOTATION_CONFIG_DOC_MAP_KEY)) {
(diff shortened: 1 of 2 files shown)


core/processor/src/main/java/io/quarkus/annotation/processor/generate_doc/ConfigDoItemFinder.java
```diff
@@ -149,6 +154,10 @@ ScannedConfigDocsItemHolder findInMemoryConfigurationItems() {
                 }
             }
 
+            if (isDeprecated) {
+                continue; // do not include deprecated config items
```
gastaldi: Although it makes sense to not add deprecated config items, I wonder if we shouldn't still display them in the docs but in a different style (as in ~~`do not use`~~)

gastaldi (approved):

core/processor/src/main/java/io/quarkus/annotation/processor/generate_doc/ConfigDoItemFinder.java
gsmet: Yeah, so I asked Manyanda to do that for now.

We have a ton of deprecated stuff in the datasource config and it will really be very confusing if we include it.

So for now, let's go this way, we can refine later if we see a use for it.

gsmet (approved):

gsmet:
I will test that tomorrow with the datasource config and will merge after that if everything is OK.

core/processor/src/main/java/io/quarkus/annotation/processor/generate_doc/ConfigDoItemFinder.java
machi1990: Yes, I'd say let's avoid the clutter for now especially when displaying the global extension config.

machi1990:
Thanks folks @gastaldi @gsmet

gsmet:
Closing as I included your commit in https://github.com/quarkusio/quarkus/pull/7785 to test it!

Issue #7740, opened 2020-03-10
Kubernetes extension generates environment variables twice


Ladicek:
**Describe the bug**

I use the Kubernetes extension configuration properties to add environment variables to the `Deployment` / `DeploymentConfig`, and I'm getting the same piece of YAML generated twice.

This happens at least for generating env var from a ConfigMap and from a secret, both in the "one env var for specific key" and "env vars for all keys" scenarios.

For example, I have this config:

```properties
# one env var for specific key in the secret
quarkus.openshift.env-vars.MY_ENV_VAR.secret=the-secret
quarkus.openshift.env-vars.MY_ENV_VAR.value=key-in-the-secret

# env vars for all keys in the ConfigMap
quarkus.openshift.env-vars.IGNORED.configmap=the-configmap
```

With this, the generated YAML contains:

```yaml
    spec:
      containers:
      - env:
        - name: "MY_ENV_VAR"
          valueFrom:
            secretKeyRef:
              key: "key-in-the-secret"
              name: "the-secret"
              optional: false
        - name: "MY_ENV_VAR"
          valueFrom:
            secretKeyRef:
              key: "key-in-the-secret"
              name: "the-secret"
              optional: false
        envFrom:
        - configMapRef:
            name: "the-configmap"
            optional: false
        - configMapRef:
            name: "the-configmap"
            optional: false
```

In the debugger, I see `AddEnvVarDecorator.andThenVisit` called twice during `session.close()`, but as far as I was able to observe, the `AddEnvVarDecorator` is only added _once_ in the Quarkus extension.

**Expected behavior**
Only one generated piece of YAML for each config key.

**Actual behavior**
Two identical pieces of YAML for each config key.

**To Reproduce**
1. Add the `quarkus-kubernetes` extension.
2. Run `mvn clean package -DskipTests`.
3. See `target/kubernetes/*.yml`.

**Configuration**
See above.

Ladicek:
CC @geoand @iocanel

geoand:
@iocanel a dekorate bug perhaps?

iocanel:
Possibly, I'll have a look.

Ladicek:
This was _partially_ fixed by #7749.

Ladicek:
Clicked the wrong button :-)

PR #7749, opened 2020-03-10
fix(#7712): S2i build no longer renames artifacts

### LINKED ISSUE #7712: presence of `quarkus-container-image-s2i` extension causes wrong `java -jar` argument in pod spec
**Describe the bug**

I use the `quarkus-container-image-s2i` extension to be able to customize the S2I base image for my application. I then `oc apply -f target/kubernetes/openshift.yml` and `oc start-build <app name> --from-dir=target` to deploy my application, which is the usual S2I binary build flow.

However, the mere presence of this extension causes a significant difference in how the Kubernetes `Deployment` (or OpenShift `DeploymentConfig`) is generated. If the S2I extension is present, the pod spec template contains a full `java -jar ... -cp ...` command. This command seems to always read `java -jar /deployments/application-runner.jar ...`, even though the JAR itself is _not_ called `application-runner.jar`. In case of OpenShift, an env var `JAVA_APP_JAR` is also generated, and its value is the correct path to the JAR.

This is probably tailored towards the S2I extension building the image itself, but that is a severely restrictive requirement.

Is generating the `java -jar ... -cp ...` command really necessary? The JDK base images typically know how to use a single JAR in `/deployments` (which already happens if I don't include the `quarkus-container-image-s2i` extension).

**Expected behavior**
I can use the generated `target/kubernetes/openshift.yml` as is, without change.

**Actual behavior**
When the `quarkus-container-image-s2i` extension is present, the generated `target/kubernetes/openshift.yml` uses extraneous configuration that prevents the deployment from succeeding.

**To Reproduce**
1. Add the `quarkus-kubernetes` and `quarkus-container-image-s2i` extensions.
2. Run `mvn clean package -DskipTests`.
3. See `target/kubernetes/*.yml`.

**Configuration**
```properties
quarkus.kubernetes.deployment-target=kubernetes,openshift
quarkus.s2i.base-jvm-image=registry.access.redhat.com/openjdk/openjdk-11-rhel7
```

**Additional context**
Compared to 1.3.0.Alpha2, this is actually a regression.

iocanel:
Fixes: #7712 #7766 #7740

This pull request changes the way that the artifact is handled during the s2i build process.

- **The artifact is no longer renamed**.
- The user is able to specify the target directory and or the target filename (if the s2i builder image changes the name).
- The artifact name that is used in the manifests comes from `JarBuildItem` / `NativeBuildItem` unless the user specifies something else (see above).

### DIFF
--- a/extensions/container-image/container-image-s2i/deployment/src/main/java/io/quarkus/container/image/s2i/deployment/S2iBaseJavaImage.java
+++ b/extensions/container-image/container-image-s2i/deployment/src/main/java/io/quarkus/container/image/s2i/deployment/S2iBaseJavaImage.java
@@ -1,0 +1,71 @@
+
+package io.quarkus.container.image.s2i.deployment;
+
+import java.util.Optional;
+
+import io.quarkus.container.image.deployment.util.ImageUtil;
+
+public enum S2iBaseJavaImage {
+
+    //We only compare `repositories` so registries and tags are stripped
+    FABRIC8("fabric8/s2i-java:latest", "JAVA_MAIN_CLASS", "JAVA_APP_JAR", "JAVA_LIB_DIR", "JAVA_CLASSPATH", "JAVA_OPTIONS"),
+    OPENJDK_8_RHEL7("redhat-openjdk-18/openjdk18-openshift:latest", "JAVA_MAIN_CLASS", "JAVA_APP_JAR", "JAVA_LIB_DIR",
+            "JAVA_CLASSPATH", "JAVA_OPTIONS"),
+    OPENJDK_8_RHEL8("openjdk/openjdk-8-rhel8:latest", "JAVA_MAIN_CLASS", "JAVA_APP_JAR", "JAVA_LIB_DIR", "JAVA_CLASSPATH",
+            "JAVA_OPTIONS"),
+    OPENJDK_11_RHEL7("openjdk/openjdk-11-rhel7:latest", "JAVA_MAIN_CLASS", "JAVA_APP_JAR", "JAVA_LIB_DIR", "JAVA_CLASSPATH",
+            "JAVA_OPTIONS"),
+    OPENJDK_11_RHEL8("openjdk/openjdk-11-rhel8:latest", "JAVA_MAIN_CLASS", "JAVA_APP_JAR", "JAVA_LIB_DIR", "JAVA_CLASSPATH",
+            "JAVA_OPTIONS");
+
+    private final String image;
+    private final String javaMainClassEnvVar;
+    private final String jarEnvVar;
+    private final String jarLibEnvVar;
+    private final String classpathEnvVar;
+    private final String jvmOptionsEnvVar;
+
+    public static Optional<S2iBaseJavaImage> findMatching(String image) {
+        for (S2iBaseJavaImage candidate : S2iBaseJavaImage.values()) {
+            if (ImageUtil.getRepository(candidate.getImage()).equals(ImageUtil.getRepository(image))) {
+                return Optional.of(candidate);
+            }
+        }
+        return Optional.empty();
+    }
+
+    private S2iBaseJavaImage(String image, String javaMainClassEnvVar, String jarEnvVar, String jarLibEnvVar,
+            String classpathEnvVar, String jvmOptionsEnvVar) {
+        this.image = image;
+        this.javaMainClassEnvVar = javaMainClassEnvVar;
+        this.jarEnvVar = jarEnvVar;
+        this.jarLibEnvVar = jarLibEnvVar;
+        this.classpathEnvVar = classpathEnvVar;
+        this.jvmOptionsEnvVar = jvmOptionsEnvVar;
+    }
+
+    public String getImage() {
+        return image;
+    }
+
+    public String getJavaMainClassEnvVar() {
+        return javaMainClassEnvVar;
+    }
+
+    public String getJvmOptionsEnvVar() {
+        return jvmOptionsEnvVar;
+    }
+
+    public String getClasspathEnvVar() {
+        return classpathEnvVar;
+    }
+
+    public String getJarLibEnvVar() {
+        return jarLibEnvVar;
+    }
+
+    public String getJarEnvVar() {
+        return jarEnvVar;
+    }
+
+}
--- a/extensions/container-image/container-image-s2i/deployment/src/main/java/io/quarkus/container/image/s2i/deployment/S2iConfig.java
+++ b/extensions/container-image/container-image-s2i/deployment/src/main/java/io/quarkus/container/image/s2i/deployment/S2iConfig.java
@@ -2,6 +2,7 @@ package io.quarkus.container.image.s2i.deployment;
 
 import java.time.Duration;
 import java.util.List;
+import java.util.Optional;
 
 import io.quarkus.runtime.annotations.ConfigItem;
 import io.quarkus.runtime.annotations.ConfigPhase;
@@ -12,6 +13,7 @@ public class S2iConfig {
 
     public static final String DEFAULT_BASE_JVM_IMAGE = "fabric8/s2i-java:2.3";
     public static final String DEFAULT_BASE_NATIVE_IMAGE = "quay.io/quarkus/ubi-quarkus-native-binary-s2i:19.3.0";
+    public static final String DEFAULT_NATIVE_TARGET_FILENAME = "application";
 
     /**
      * The base image to be used when a container image is being produced for the jar build
@@ -38,18 +40,32 @@ public class S2iConfig {
     public List<String> nativeArguments;
 
     /**
-     * The path to where the jar is added during the assemble phase.
+     * The directory where the jar is added during the assemble phase.
      * This is dependant on the s2i image and should be supplied if a non default image is used.
      */
-    @ConfigItem(defaultValue = "/deployments/application${quarkus.package.runner-suffix}.jar")
-    public String jarPath;
+    @ConfigItem(defaultValue = "/deployments/")
+    public String jarDirectory;
 
     /**
-     * The path to where the native binary is added during the assemble phase.
-     * This is dependant on the s2i image and should be supplied if a non default image is used.
+     * The resulting filename of the jar in the s2i image.
+     * This option may be used if the selected s2i image uses a fixed name for the jar.
      */
-    @ConfigItem(defaultValue = "/home/quarkus/application")
-    public String nativeBinaryPath;
+    @ConfigItem
+    public Optional<String> jarFileName;
+
+    /**
+     * The directory where the native binary is added during the assemble phase.
+     * This is dependant on the s2i image and should be supplied if a non-default image is used.
+     */
+    @ConfigItem(defaultValue = "/home/quarkus/")
+    public String nativeBinaryDirectory;
+
+    /**
+     * The resulting filename of the native binary in the s2i image.
+     * This option may be used if the selected s2i image uses a fixed name for the native binary.
+     */
+    @ConfigItem
+    public Optional<String> nativeBinaryFileName;
 
     /**
      * The build timeout.
--- a/extensions/container-image/container-image-s2i/deployment/src/main/java/io/quarkus/container/image/s2i/deployment/S2iProcessor.java
+++ b/extensions/container-image/container-image-s2i/deployment/src/main/java/io/quarkus/container/image/s2i/deployment/S2iProcessor.java
@@ -12,12 +12,16 @@ import java.nio.file.Paths;
 import java.nio.file.StandardCopyOption;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.Executor;
 import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
+import java.util.function.Predicate;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
@@ -246,14 +250,15 @@ public class S2iProcessor {
         //Let's disable http2 as it causes issues with duplicate build triggers.
         config.setHttp2Disable(true);
         KubernetesClient client = Clients.fromConfig(config);
+        OpenShiftClient openShiftClient = client.adapt(OpenShiftClient.class);
         KubernetesList kubernetesList = Serialization.unmarshalAsList(new ByteArrayInputStream(openshiftManifests.getData()));
 
         List<HasMetadata> buildResources = kubernetesList.getItems().stream()
                 .filter(i -> i instanceof BuildConfig || i instanceof ImageStream || i instanceof Secret)
                 .collect(Collectors.toList());
 
-        applyS2iResources(client, buildResources);
-        s2iBuild(client, buildResources, tar, s2iConfig);
+        applyS2iResources(openShiftClient, buildResources);
+        s2iBuild(openShiftClient, buildResources, tar, s2iConfig);
     }
 
     /**
@@ -262,10 +267,10 @@ public class S2iProcessor {
      * @param client the client instance
      * @param the resources to apply
      */
-    private static void applyS2iResources(KubernetesClient client, List<HasMetadata> buildResources) {
+    private static void applyS2iResources(OpenShiftClient client, List<HasMetadata> buildResources) {
         // Apply build resource requirements
         try {
-            buildResources.forEach(i -> {
+            for (HasMetadata i : distinct(buildResources)) {
                 if (i instanceof BuildConfig) {
                     client.resource(i).cascading(true).delete();
                     try {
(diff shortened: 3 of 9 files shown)


iocanel:
CC @Ladicek @maxandersen

iocanel:
Found some issues with the PR related to native mode ... 
I am on it.

maxandersen:
as this changes the behavior of kubernetes/openshift setup best to get this fixed before final.

extensions/container-image/container-image-s2i/deployment/src/main/java/io/quarkus/container/image/s2i/deployment/S2iProcessor.java
```diff
@@ -102,11 +109,30 @@ public void s2iRequirementsNative(S2iConfig s2iConfig,
...
             BuildProducer<BaseImageInfoBuildItem> builderImageProducer,
             BuildProducer<KubernetesCommandBuildItem> commandProducer) {
 
+        boolean usingDefaultBuilder = ImageUtil.getRepository(S2iConfig.DEFAULT_BASE_NATIVE_IMAGE)
+                .equals(ImageUtil.getRepository(s2iConfig.baseNativeImage));
+        String outputNativeBinaryFileName = nativeImage.getPath().getFileName().toString();
+
+        String nativeBinaryFileName = null;
+
+        //The default s2i builder for native builds, renames the native binary.
```
maxandersen: wait - so our own native s2i builder goes an rename to something even more different ?

geoand:
I would like @Ladicek to test this if possible.

I agree with @maxandersen that we need this in for `Final` from a usability perspective. Moreover it changes the s2i configuration which we really want to do before the `Final` is out to avoid unnecessary breaking changes down the line

extensions/container-image/container-image-s2i/deployment/src/main/java/io/quarkus/container/image/s2i/deployment/S2iProcessor.java
Ladicek: I also think this shouldn't happen. It was probably a shortcut to make the `run` script easier. If anything, it shouldn't be a rename but a symlink.

Ladicek (approved):
This seems to work fine from my perspective. I don't test native binaries in OpenShift yet, but for the JVM mode, this is OK.

(I'm still not sure why we need the full `java ...` cmdline, but it can't hurt either, so I'm fine with that.)

geoand:
Thanks for checking @Ladicek

extensions/container-image/container-image-s2i/deployment/src/main/java/io/quarkus/container/image/s2i/deployment/S2iConfig.java
```diff
@@ -38,18 +40,32 @@
...
      */
-    @ConfigItem(defaultValue = "/deployments/application${quarkus.package.runner-suffix}.jar")
-    public String jarPath;
+    @ConfigItem(defaultValue = "/deployments/")
+    public String jarDirectory;
 
     /**
-     * The path to where the native binary is added during the assemble phase.
+     * The resulting filename of the jar in the s2i image.
+     * This option may be used of the selected s2i image uses a fixed name for the jar.
```
geoand:
```suggestion
     * This option may be used if the selected s2i image uses a fixed name for the jar.
```

extensions/container-image/container-image-s2i/deployment/src/main/java/io/quarkus/container/image/s2i/deployment/S2iConfig.java
```diff
@@ -38,18 +40,32 @@
...
+     */
+    @ConfigItem
+    public Optional<String> jarFileName;
+
+    /**
+     * The directory where the native binary is added during the assemble phase.
      * This is dependant on the s2i image and should be supplied if a non default image is used.
```
geoand:
```suggestion
     * This is dependant on the s2i image and should be supplied if a non-default image is used.
```

extensions/container-image/container-image-s2i/deployment/src/main/java/io/quarkus/container/image/s2i/deployment/S2iConfig.java
```diff
@@ -38,18 +40,32 @@
...
      */
-    @ConfigItem(defaultValue = "/home/quarkus/application")
-    public String nativeBinaryPath;
+    @ConfigItem(defaultValue = "/home/quarkus/")
+    public String nativeBinaryDirectory;
+
+    /**
+     * The resulting filename of the native binary in the s2i image.
+     * This option may be used of the selected s2i image uses a fixed name for the native binary.
```
geoand:
```suggestion
     * This option may be used if the selected s2i image uses a fixed name for the native binary.
```

geoand (commented):
Added some minor comments about the config property javadoc (which should be fixed since they end up in the docs)

iocanel:
> This seems to work fine from my perspective. I don't test native binaries in OpenShift yet, but for the JVM mode, this is OK.

It's the ubi quarkus native binary image that does rename the artifact. It doesn't affect JVM mode.

> (I'm still not sure why we need the full `java ...` cmdline, but it can't hurt either, so I'm fine with that.)

Because there is no guarantee that the s2i builder image the user selects will support `JAVA_APP_JAR` etc.

extensions/container-image/container-image-s2i/deployment/src/main/java/io/quarkus/container/image/s2i/deployment/S2iProcessor.java
iocanel: I logged an issue: https://github.com/quarkusio/quarkus-images/issues/58
Until the issue is address we need this, so I'll mark this discussion as resolved.

geoand:
Please squash the commits when you are done :)

iocanel:
> Please squash the commits when you are done :)

done

geoand (approved):
LGTM

Ladicek:
> > (I'm still not sure why we need the full `java ...` cmdline, but it can't hurt either, so I'm fine with that.)
> 
> Because there is no guarantee that the s2i builder image the user selects will support `JAVA_APP_JAR` etc.

I have actually noticed that it _might_ hurt, or at least be suboptimal. I filed #7766 for that.

gsmet:
@geoand please don't assign PRs to 1.3.0.Final. I do it when I backport things.

As long as they are not backported, they don't end up in 1.3.0.Final as we already have a branch.

Thanks!

geoand:
So the milestone should be `1.4.0.Final`?

iocanel:
Added a couple more commits that fixes: #7766  and #7740. 
Also it changes the default s2i java base image to `openjdk`.

geoand:
Another looks from the s2i master @Ladicek would be great :)

extensions/container-image/container-image-s2i/deployment/src/main/java/io/quarkus/container/image/s2i/deployment/S2iConfig.java
```diff
@@ -11,7 +11,7 @@
 @ConfigRoot(phase = ConfigPhase.BUILD_TIME)
 public class S2iConfig {
 
-    public static final String DEFAULT_BASE_JVM_IMAGE = "fabric8/s2i-java:2.3";
+    public static final String DEFAULT_BASE_JVM_IMAGE = "registry.access.redhat.com/openjdk/openjdk-11-rhel7";
```
Ladicek: I'm not sure if this is a good idea, purely from "legal" point of view. This is RHEL, after all. Too bad UBI Java isn't ready yet :-(

geoand: @maxandersen any idea about this?

extensions/container-image/container-image-s2i/deployment/src/main/java/io/quarkus/container/image/s2i/deployment/S2iBaseJavaImage.java
```diff
@@ -0,0 +1,64 @@
...
+package io.quarkus.container.image.s2i.deployment;
+
+import java.util.Optional;
+
+import io.quarkus.container.image.deployment.util.ImageUtil;
+
+public enum S2iBaseJavaImage {
+
+    FABRIC8("fabric8/s2i-java:2.3", "JAVA_MAIN_CLASS", "JAVA_APP_JAR", "JAVA_LIB_DIR", "JAVA_CLASSPATH", "JAVA_OPTIONS"),
+    OPENJDK("registry.access.redhat.com/openjdk/openjdk-11-rhel7", "JAVA_MAIN_CLASS", "JAVA_APP_JAR", "JAVA_LIB_DIR",
```
Ladicek: There are 4 Red Hat OpenJDK images that I'm aware of (OpenJDK 8 and 11, RHEL 7 and 8), some of them being available from 2 registries (authenticated and unauthenticated).

- OpenJDK 8 (I'm not sure if we still support Java 8, but we probably do. Defaulting to 11 is probably good though.)
    - RHEL 7
        - registry.access.redhat.com/redhat-openjdk-18/openjdk18-openshift
        - registry.redhat.io/redhat-openjdk-18/openjdk18-openshift
    - RHEL 8
        - registry.redhat.io/openjdk/openjdk-8-rhel8 (registry.access.redhat.com variant not available)
- OpenJDK 11
    - RHEL 7
        - registry.access.redhat.com/openjdk/openjdk-11-rhel7
        - registry.redhat.io/openjdk/openjdk-11-rhel7
    - RHEL 8
        - registry.redhat.io/openjdk/openjdk-11-rhel8 (registry.access.redhat.com variant not available)

They all have different tags (as does the Fabric8 image), which should all be recognized. (That is, tag should be ignored when searching.)

Ladicek:
I will take a look in a few hours, I need to sort out some higher-priority items first.

geoand:
@gsmet Just as an FYI, we'll need to get this one in for `Final`. @iocanel will be making some final changes and @Ladicek will hopefully check within the day

extensions/container-image/container-image-s2i/deployment/src/main/java/io/quarkus/container/image/s2i/deployment/S2iBaseJavaImage.java
iocanel: Added more images. Also cleaned them up a little bit, as the registry is stripped when performing the comparison.

gsmet:
Please ping the appropriate people when this is ready.

iocanel:
> Please ping the appropriate people when this is ready.
I don't have anything more to add. So @maxandersen: Ping!

geoand:
I have already approved the PR. I just want @Ladicek to check it since he knows more about s2i and has been the one that finds all these subtle issues

Ladicek:
I've finished the high-priority work now. I'll try to check between the meetings I have now :-)

Ladicek:
I can confirm the duplicate env vars issue (#7740) is partially fixed.

1.

```properties
# one env var for specific key in the secret
quarkus.openshift.env-vars.MY_ENV_VAR.secret=the-secret
quarkus.openshift.env-vars.MY_ENV_VAR.value=key-in-the-secret
```

This case works fine now:

```yaml
        - name: "MY_ENV_VAR"
          valueFrom:
            secretKeyRef:
              key: "key-in-the-secret"
              name: "the-secret"
              optional: false
```

2.

```properties
# env vars for all keys in the ConfigMap
quarkus.openshift.env-vars.IGNORED.configmap=the-configmap
```

This case still causes duplicate entry in the YAML file:

```yaml
        envFrom:
        - configMapRef:
            name: "the-configmap"
            optional: false
        - configMapRef:
            name: "the-configmap"
            optional: false
```

iocanel:
> This case still causes duplicate entry in the YAML file:

As this is a non-blocker issue, I think that the fix for that particular uses case should be deferred post 1.3.0.Final.

geoand:
> > This case still causes duplicate entry in the YAML file:
> 
> As this is a non-blocker issue, I think that the fix for that particular uses case should be deferred post 1.3.0.Final.

I agree

Ladicek:
Yea, there's no way how to indicate priority in GitHub issues (except labels, but I can't assign those). If it were, I'd mark that particular issue as not a blocker -- but since you already tried to fix it, I wanted to report that the fix is incomplete :-)

Ladicek:
Ah, one thing. This doesn't mean I didn't find any other issue. I'm still checking. This one was just too easy to find :-)

geoand:
@Ladicek I propose we merge now and if you find any issues open them along with an indication of how serious they are for `.Final`

Ladicek:
I agree with merging (as I didn't spot any other glaring issue yet), but would like to have these 2 questions answered first:

1. Can we use the `registry.access.redhat.com/openjdk/openjdk-11-rhel7` image by default?
2. Can the fact that "known image" matching algorithm ignores the registry name have security consequences? Or any other consequences? I can only think of the possibility to abuse that by creating identically named image in an internal registry. But I'd like other folks to think about that too, at least a little.

iocanel:
> I agree with merging (as I didn't spot any other glaring issue yet), but would like to have these 2 questions answered first:
> 
> 1. Can we use the `registry.access.redhat.com/openjdk/openjdk-11-rhel7` image by default?

This is already the image we use. If your question is if we should, then maybe @maxandersen  can answer that.

> 2. Can the fact that "known image" matching algorithm ignores the registry name have security consequences? Or any other consequences? I can only think of the possibility to abuse that by creating identically named image in an internal registry. But I'd like other folks to think about that too, at least a little.

Let me start by saying that we can't take the registry into consideration, as this would break all private registries. Now, let's see what the implications might be:

Security wise, nothing changes. We don't grant any special privileges to known images, we just align our generated manifests (something anyone can do by hand anyway).

Other consequences?

Yeah, if someone creates an identically named image that doesn't respect the same environment variable conventions as the original one, publish that to a private, and use that instead, then the resulting container will probably fail to start. But I think we can live with that (at least for now).

Ladicek:
> > I agree with merging (as I didn't spot any other glaring issue yet), but would like to have these 2 questions answered first:
> > 
> > 1. Can we use the `registry.access.redhat.com/openjdk/openjdk-11-rhel7` image by default?
> 
> This is already the image we use.

No we don't. We currently use the Fabric8 image by default. This PR changes that to the Red Hat image. I'm not sure if that's right. Technically, the image is available for everyone. However, I think that "conditions apply". If you don't have a Red Hat subscription (even the $0 developer subscription), you shouldn't use it. Or something like that.

> > 2. Can the fact that "known image" matching algorithm ignores the registry name have security consequences? Or any other consequences? I can only think of the possibility to abuse that by creating identically named image in an internal registry. But I'd like other folks to think about that too, at least a little.
> 
> Let me start by saying that we can't take the registry into consideration, as this would break all private registries. Now, let's see what the implications might be:
> 
> Security wise, nothing changes. We don't grant any special privileges to known images, we just align our generated manifests (something anyone can do by hand anyway).
> 
> Other consequences?
> 
> Yeah, if someone creates an identically named image that doesn't respect the same environment variable conventions as the original one, publish that to a private, and use that instead, then the resulting container will probably fail to start. But I think we can live with that (at least for now).

OK, agree.

iocanel:
> No we don't. We currently use the Fabric8 image by default. This PR changes that to the Red Hat image. 

I stand corrected!

geoand:
> > No we don't. We currently use the Fabric8 image by default. This PR changes that to the Red Hat image.
> 
> I stand corrected!

Which leaves us where with regards to @Ladicek 's comment?

iocanel:
> > > No we don't. We currently use the Fabric8 image by default. This PR changes that to the Red Hat image.
> > 
> > 
> > I stand corrected!
> 
> Which leaves us where with regards to @Ladicek 's comment?

We need to check if we can use `registry.access.redhat.com/openjdk/openjdk-11-rhel7` as a default image. It seems that @maxandersen is swarmed though. @tqvarnst since we discussed about this on Monday, are there any legal obstacles in proceeding with this change (adopting openjdk-11-rhel7 as the default s2i base image)?

Ladicek:
I don't like us using the Fabric8 image, but I don't think there's a better image that we can use by default. When UBI image with Java comes out, we should switch to that for sure, but until then...

gsmet:
For the default Dockerfiles, we use ubi and install Java manually.

geoand:
We can't do that for s2i however

maxandersen:
long thread and not sure what questions are left but just two things to state:

* we are stuck with fabric8 image until ubi-java comes out - for customers they'll need to explicitly set whatever supported image is needed.
* we are not going to use any image that require auth as default.

extensions/container-image/container-image-s2i/deployment/src/main/java/io/quarkus/container/image/s2i/deployment/S2iProcessor.java
```diff
@@ -61,53 +65,100 @@
...
             BuildProducer<BaseImageInfoBuildItem> builderImageProducer,
             BuildProducer<KubernetesCommandBuildItem> commandProducer) {
 
         final List<AppDependency> appDeps = curateOutcomeBuildItem.getEffectiveModel().getUserDependencies();
+        String outputJarFileName = jarBuildItem.getPath().getFileName().toString();
         String classpath = appDeps.stream()
-                .map(d -> Paths.get(s2iConfig.jarPath).getParent().resolve("lib")
-                        .resolve(d.getArtifact().getGroupId() + "." + d.getArtifact().getPath().getFileName()).toAbsolutePath()
+                .map(d -> String.valueOf(d.getArtifact().getGroupId() + "." + d.getArtifact().getPath().getFileName()))
+                .map(s -> Paths.get(s2iConfig.jarDirectory).resolve("lib").resolve(s).toAbsolutePath()
```
Ladicek: I wonder what happens on Windows, would this possibly generate Windows-style paths into the Kubernetes YAML? (Didn't check, just thinking aloud.)

Ladicek (approved):
This works fine for my usecase. There might be follow-up issues, but this is a good base IMHO.

geoand:
Since we are practically out of time, I'm going to merge this to get the needed fixes into `Final`

PR #7756, opened 2020-03-10
Set compiler encoding to UTF-8 in the generated build.gradle

gastaldi:


maxandersen (approved):
this will make gradle projects have same default encoding as the maven ones so +1 for this

devtools/platform-descriptor-json/src/main/resources/templates/basic-rest/scala/build.gradle-template.ftl
```diff
@@ -20,3 +20,6 @@ dependencies {
 group '${project_groupId}'
 version '${project_version}'
 
+compileScala {
+  scalaCompileOptions.encoding = 'UTF-8'
+}
```
gsmet: I see you added something for Scala. We don't need something for Kotlin?

gsmet (changes requested):
Asked a question. Feel free to dismiss if it's a false alarm.

devtools/platform-descriptor-json/src/main/resources/templates/basic-rest/scala/build.gradle-template.ftl
gastaldi: According to https://stackoverflow.com/questions/45564450/how-to-set-kotlin-source-encoding-in-gradle: 

>Kotlin source files are always UTF-8 by design. There is no way to specify any other encoding.

gsmet (approved):

PR #7764, opened 2020-03-11
Add the option to skip the build mojo from the command line

ppalaga:


ppalaga:
This is rather trivial. @aloubyansky @gsmet could you please review?

devtools/maven/src/main/java/io/quarkus/maven/BuildMojo.java
```diff
@@ -116,7 +116,8 @@
     @Parameter(property = "ignoredEntries")
     private String[] ignoredEntries;
 
-    @Parameter(defaultValue = "false")
+    /** Skip the execution of this mojo */
+    @Parameter(defaultValue = "false", property = "quarkus.skip")
```
gastaldi: Perhaps we should name it `quarkus.build.skip`? Other mojos might follow the same convention for the skip flag too :wink:

ppalaga: +1

ppalaga:
0e04db1 now uses `quarkus.build.skip` property name instead of the originally proposed  `quarkus.skip`.

gastaldi (approved):

gastaldi:
@ppalaga thanks! Do you need that for 1.3.0.Final or can it wait for 1.4 ?

ppalaga:
> @ppalaga thanks! Do you need that for 1.3.0.Final

1.3.0.Final would be highly welcome, thanks!

PR #7772, opened 2020-03-11
Remove Kogito extension from Quarkus codebase

mariofusco:


bom/runtime/pom.xml
```diff
@@ -2921,6 +2862,23 @@
...
+            <dependency>
+                <groupId>org.kie.kogito</groupId>
+                <artifactId>kogito-drools</artifactId>
+                <version>${kogito.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>org.kie.kogito</groupId>
+                <artifactId>drools-core-static</artifactId>
+                <version>${kogito.version}</version>
+            </dependency>
```
gastaldi: Why this is still needed if kogito is being removed from the Quarkus codebase?

devtools/bom-descriptor-json/src/main/resources/extensions-overrides.json
```diff
@@ -372,6 +372,12 @@
       "artifact-id": "camel-quarkus-pdf",
       "version": "0.3.1",
       "description": "Camel PDF support"
+    },
+    {
+      "group-id": "org.kie.kogito",
+      "artifact-id": "kogito-quarkus",
+      "version": "0.8.0",
+      "description": "Kogito extension"
```
gastaldi: Why this is still needed if kogito is being removed from the Quarkus codebase?

bom/runtime/pom.xml
mariofusco: @gastaldi That is required by optaplanner extension.

mariofusco:
Hi @galderz as you can see this PR caused the failure of 2 tests in native mode. The failure of the tests is caused by the upgrade of kogito from 0.7.1 to 0.8.0. I'm trying to track down this problem but I must admit that at the moment I have no clue. Do you have any suggestion on what to do to investigate this?

evacchi:
Hello guys, I have built the PR with GraalVM 20 and the error message is a little bit more detailed https://gist.github.com/evacchi/3ed85c9ada1d1382cb500da17e37a6ff  /cc @ge0ffrey

geoand:
I pushed a fix to the PR. But it seems like there are conflicts... I don't know where those are coming from...

mariofusco:
Replaced by https://github.com/quarkusio/quarkus/pull/7815

devtools/bom-descriptor-json/src/main/resources/extensions-overrides.json
maxandersen: it will pickup what is in the extension - which btw. should be more than just "Kogito extension" :)

galderz:
> Hi @galderz as you can see this PR caused the failure of 2 tests in native mode. The failure of the tests is caused by the upgrade of kogito from 0.7.1 to 0.8.0. I'm trying to track down this problem but I must admit that at the moment I have no clue. Do you have any suggestion on what to do to investigate this?

Guess you meant @gastaldi ? ;)

PR #7779, opened 2020-03-11
Support for repeating qualifiers in ArC plus automated tests

### LINKED ISSUE #7732: Support repeatable qualifier annotations
https://docs.jboss.org/cdi/spec/2.0/cdi-spec.html#repeating_qualifiers

manovotn:
Fixes #7732

### DIFF
--- a/independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/BeanDeployment.java
+++ b/independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/BeanDeployment.java
@@ -57,6 +57,8 @@ public class BeanDeployment {
 
     private final Map<DotName, ClassInfo> qualifiers;
 
+    private final Map<DotName, ClassInfo> repeatingQualifierAnnotations;
+
     private final Map<DotName, ClassInfo> interceptorBindings;
 
     private final Map<DotName, Set<String>> nonBindingFields;
@@ -132,6 +134,7 @@ public class BeanDeployment {
         this.customContexts = new ConcurrentHashMap<>();
 
         this.qualifiers = findQualifiers(index);
+        this.repeatingQualifierAnnotations = findContainerAnnotations(qualifiers, index);
         buildContextPut(Key.QUALIFIERS.asString(), Collections.unmodifiableMap(qualifiers));
 
         this.interceptorBindings = findInterceptorBindings(index);
@@ -363,6 +366,30 @@ public class BeanDeployment {
         return qualifiers.get(name);
     }
 
+    /**
+     * Extracts qualifiers from given annotation instance.
+     * This returns a collection because in case of repeating qualifiers there can be multiple.
+     * For most instances this will be a singleton instance (if given annotatation is a qualifier) or an empty list for
+     * cases where the annotation is not a qualifier.
+     *
+     * @param annotation instance to be inspected
+     * @return a collection of qualifiers or an empty collection
+     */
+    Collection<AnnotationInstance> extractQualifiers(AnnotationInstance annotation) {
+        DotName annotationName = annotation.name();
+        if (qualifiers.get(annotationName) != null) {
+            return Collections.singleton(annotation);
+        } else {
+            if (repeatingQualifierAnnotations.get(annotationName) != null) {
+                // container annotation, we need to extract actual qualifiers
+                return new ArrayList<>(Arrays.asList(annotation.value().asNestedArray()));
+            } else {
+                // neither qualifier, nor container annotation, return empty collection
+                return Collections.EMPTY_LIST;
+            }
+        }
+    }
+
     ClassInfo getInterceptorBinding(DotName name) {
         return interceptorBindings.get(name);
     }
@@ -415,6 +442,18 @@ public class BeanDeployment {
             qualifiers.put(qualifier.target().asClass().name(), qualifier.target().asClass());
         }
         return qualifiers;
+    }
+
+    private static Map<DotName, ClassInfo> findContainerAnnotations(Map<DotName, ClassInfo> qualifiers, IndexView index) {
+        Map<DotName, ClassInfo> containerAnnotations = new HashMap<>();
+        for (ClassInfo qualifier : qualifiers.values()) {
+            AnnotationInstance instance = qualifier.classAnnotation(DotNames.REPEATABLE);
+            if (instance != null) {
+                DotName annotationName = instance.value().asClass().name();
+                containerAnnotations.put(annotationName, getClassByName(index, annotationName));
+            }
+        }
+        return containerAnnotations;
     }
 
     private static Map<DotName, ClassInfo> findInterceptorBindings(IndexView index) {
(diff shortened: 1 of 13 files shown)


independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/Beans.java
```diff
@@ -51,11 +51,12 @@ static BeanInfo createClassBean(ClassInfo beanClass, BeanDeployment beanDeployme
...
         for (AnnotationInstance annotation : beanDeployment.getAnnotations(beanClass)) {
-            if (beanDeployment.getQualifier(annotation.name()) != null) {
+            for (AnnotationInstance qualifierAnnotation : beanDeployment.getQualifierExtractor(annotation.name())
+                    .apply(annotation)) {
                 // Qualifiers
-                qualifiers.add(annotation);
-                if (DotNames.NAMED.equals(annotation.name())) {
-                    AnnotationValue nameValue = annotation.value();
+                qualifiers.add(qualifierAnnotation);
+                if (DotNames.NAMED.equals(qualifierAnnotation.name())) {
```
mkouba: I don't think we need to run this for every annotation. I'd place this `if`  block outside the loop.

independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/Beans.java
```diff
@@ -51,11 +51,12 @@ static BeanInfo createClassBean(ClassInfo beanClass, BeanDeployment beanDeployme
         String name = null;
 
         for (AnnotationInstance annotation : beanDeployment.getAnnotations(beanClass)) {
-            if (beanDeployment.getQualifier(annotation.name()) != null) {
+            for (AnnotationInstance qualifierAnnotation : beanDeployment.getQualifierExtractor(annotation.name())
```
mkouba: Hm, maybe we could simplify the code a little bit. `Function<> getQualifierExtractor(annotation.name())` -> `Collection<AnnotationInstance> getQualifiers(annotation)` and the loop could be replaced with `beanDeployment.getQualifiers(annotation).forEach(qualifiers::add);`. WDYT?

mkouba (changes requested):
I think that `Beans` is not the only place where we should handle repeatable qualifiers. Although the spec only mentions the type safe resolution it should work for observers, disposers, injection points, etc. In other words, we should find all occurences of `io.quarkus.arc.processor.BeanDeployment.getQualifier(DotName)` and consider which should be replaced.

Otherwise looks really good!

independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/Beans.java
manovotn: I just realized that my code doesn't call `continue` properly on the outer loop once we know it's qualifier (so we needn't process it any further). And so does yours. We should take that into account, so in fact it should look something like this:

```
            Collection<AnnotationInstance> qualifierCollection = beanDeployment.getQualifierExtractor(annotation.name()).apply(annotation);
            for (AnnotationInstance qualifierAnnotation : qualifierCollection) {
                // Qualifiers
                qualifiers.add(qualifierAnnotation);
            }
            if (!qualifierCollection.isEmpty()) {
                // we needn't process it further, the annotation was a qualifier (or multiple repeating ones)
                continue;
            }
```

independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/BeanDeployment.java
```diff
@@ -363,6 +372,20 @@ ClassInfo getQualifier(DotName name) {
         return qualifiers.get(name);
     }
 
+    Function<AnnotationInstance, Collection<AnnotationInstance>> getQualifierExtractor(DotName name) {
```
mkouba: Looking at the usage I wonder if it makes sense to return a function. `Collection<AnnotationInstance> getQualifiers(AnnotationInstance annotation)` would be sufficient and more convenient in most cases. Alternatively, we could add a convenient method like this:
```java
Collection<AnnotationInstance> getQualifiers(AnnotationInstance annotation) {
    return getQualifierExtractor(annotation.name()).apply(annotation);
}
```

independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/BeanDeployment.java
```diff
@@ -95,6 +97,12 @@
     private final boolean removeFinalForProxyableMethods;
     private final boolean jtaCapabilities;
 
+    private Function<AnnotationInstance, Collection<AnnotationInstance>> singletonFunction = annotationInstance -> Collections
```
mkouba: These could be final... but it's just a minor.

independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/DisposerInfo.java
```diff
@@ -71,9 +72,11 @@ void init(List<Throwable> errors) {
     }
 
     Collection<AnnotationInstance> getDisposedParameterQualifiers() {
-        return Annotations.getParameterAnnotations(declaringBean.getDeployment(), disposerMethod, disposedParameter.position())
-                .stream().filter(a -> declaringBean.getDeployment().getQualifier(a.name()) != null)
-                .collect(Collectors.toList());
+        Set<AnnotationInstance> resultingQualifiers = new HashSet<>();
+        Annotations.getParameterAnnotations(declaringBean.getDeployment(), disposerMethod, disposedParameter.position())
+                .stream().forEach(a -> declaringBean.getDeployment().getQualifierExtractor(a.name()).apply(a)
+                        .forEach(an -> resultingQualifiers.add(an)));
```
mkouba: `resultingQualifiers::add`?

mkouba (changes requested):

independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/BeanDeployment.java
manovotn: Ok, I will change that to the suggested method. You're right that it's easier to read and use.

independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/BeanDeployment.java
manovotn: These will be removed altogether accordingly to your earlier suggestion.

manovotn:
Fixed the things you asked me for - the util method is longer operates on functions.
However, I also found a glitch in test I wrote and fixing it showed me an error. I am now looking at whether that is just me being silly on Mon or whether we have a bug in disposer method resolution.

manovotn:
Turns out it was an actual bug - in typesafe resolution done for disposer methods with qualifiers.
I will fix it as part of this PR.

manovotn:
Updated, I am once again getting green builds locally :)

mkouba (approved):

PR #7786, opened 2020-03-11
Fixes related to Maven projects using the 'revision' property as the project version

### LINKED ISSUE #7747: "Maven CI Friendly Versions" are not (fully) supported, application is incomplete/broken
**Describe the bug**
The `build` goal of the `quarkus-maven-plugin` builds an incomplete application in case ["Maven CI Friendly Versions"](https://maven.apache.org/maven-ci-friendly.html) are used.

E.g.:

Parent `pom.xml`:
```xml
    <groupId>some.project</groupId>
    <artifactId>root</artifactId>
    <version>${revision}</version>
    <packaging>pom</packaging>
    <properties>
        <!-- This is only the default for local development, CI will override it! -->
        <revision>0.1-SNAPSHOT</revision>
    </properties>
```

Submodule `pom.xml` (the one with the `quarkus-maven-plugin` execution):
```xml
  <parent>
    <groupId>some.project</groupId>
    <artifactId>root</artifactId>
    <version>${revision}</version>
  </parent>

  <artifactId>dist</artifactId>
```

As described on  ["Maven CI Friendly Versions"](https://maven.apache.org/maven-ci-friendly.html) (paragraph "Install / Deploy"), `flatten-maven-plugin` is configured properly.

**Expected behavior**
The application is built normally/fully. There should be no difference to e.g. `<version>0.1-SNAPSHOT</version>`.

**Actual behavior**
The runner jar is missing most (all?) of the application classes and the following is logged:
```
[INFO] --- quarkus-maven-plugin:1.3.0.CR2:build (default) @ dist ---
[WARNING] The POM for some.project:dist:jar:${revision} is missing, no dependency information available
[WARNING] The POM for some.project:dist:jar:${revision} is missing, no dependency information available
[INFO] [org.jboss.threads] JBoss Threads version 3.0.1.Final
[INFO] [io.quarkus.deployment.pkg.steps.JarResultBuildStep] Building thin jar: C:\Develop\some-project\dist\target\dist-0.1-SNAPSHOT-runner.jar
[INFO] [io.quarkus.deployment.QuarkusAugmentor] Quarkus augmentation completed in 1106ms
[INFO] BUILD SUCCESS
```

**To Reproduce**
1. set up a small multi module project (parent + one module)
2. introduce `${revision}`
3. introduce `flatten-maven-plugin`
4. try to build the Quarkus application (e.g. `mvn clean install` on root level)

I am pretty swamped at the moment so I cannot provide a sample project right away.
But please do let me know if you need one.

_Update:_ Reproducer created, see https://github.com/quarkusio/quarkus/issues/7747#issuecomment-597336269

**Configuration**
none

**Screenshots**
none

**Environment:**
 - Output of `uname -a` or `ver`: `MINGW64_NT-10.0-18363 W4DEUMSY9003463 3.0.7-338.x86_64 2019-11-21 23:07 UTC x86_64 Msys`
 - Output of `java -version`: `OpenJDK 64-Bit Server VM AdoptOpenJDK (build 11.0.6+10, mixed mode)`
 - GraalVM version (if different from Java): none
 - Quarkus version or git rev: 1.3.0.CR2
 - Build tool (ie. output of `mvnw --version` or `gradlew --version`): `Apache Maven 3.6.3`

**Additional context**

I think I already found the source of the problem:
`io.quarkus.bootstrap.resolver.maven.workspace.LocalProject` is re-reading the `pom.xml` via `ModelUtils` / `MavenXpp3Reader`. This simple XML parsing does _not_ take care of the property interpolation that is required for the resolution of `${revision}`.

_One_ solution might be to prefer a possibly existing `.flattened-pom.xml` to the regular `pom.xml`.
I did something similar here: https://github.com/shrinkwrap/resolver/pull/142
Since the name and location of the flattened file is configurable, Quarkus should probably provide a config option to reflect that.

aloubyansky:
Fixes #7747

### DIFF
--- a/independent-projects/bootstrap/core/src/main/java/io/quarkus/bootstrap/resolver/maven/workspace/LocalProject.java
+++ b/independent-projects/bootstrap/core/src/main/java/io/quarkus/bootstrap/resolver/maven/workspace/LocalProject.java
@@ -10,8 +10,10 @@ import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 import org.apache.maven.model.Dependency;
 import org.apache.maven.model.Model;
@@ -28,17 +30,24 @@ public class LocalProject {
 
     private static final String PROJECT_BASEDIR = "${project.basedir}";
     private static final String POM_XML = "pom.xml";
+    private static final String REVISION = "revision";
+    static final String REVISION_EXPR = "${revision}";
 
     public static LocalProject load(Path path) throws BootstrapException {
         return load(path, true);
     }
 
     public static LocalProject load(Path path, boolean required) throws BootstrapException {
-        Path cpd = locateCurrentProjectDir(path, required);
+        final Path cpd = locateCurrentProjectDir(path, required);
         if (cpd == null) {
             return null;
         }
-        return new LocalProject(readModel(cpd.resolve(POM_XML)), null);
+        try {
+            return new LocalProject(readModel(cpd.resolve(POM_XML)), null);
+        } catch (UnresolvedRevisionException e) {
+            // if the revision couldn't be resolved, we are trying to resolve it from the workspace
+            return loadWorkspace(cpd);
+        }
     }
 
     public static LocalProject loadWorkspace(Path path) throws BootstrapException {
@@ -136,10 +145,37 @@ public class LocalProject {
         this.workspace = workspace;
         this.groupId = ModelUtils.getGroupId(rawModel);
         this.artifactId = rawModel.getArtifactId();
-        this.version = ModelUtils.getVersion(rawModel);
+
+        String version = ModelUtils.getVersion(rawModel);
+        final boolean revisionVersion;
+        if (revisionVersion = REVISION_EXPR.equals(version)) {
+            version = resolveRevision(rawModel);
+        }
+
         if (workspace != null) {
             workspace.addProject(this, rawModel.getPomFile().lastModified());
+            if (revisionVersion) {
+                if (version == null) {
+                    version = workspace.getRevision();
+                    if (version == null) {
+                        throw UnresolvedRevisionException.forGa(groupId, artifactId);
+                    }
+                } else {
+                    workspace.setRevision(version);
+                }
+            }
+        } else if (version == null) {
+            throw UnresolvedRevisionException.forGa(groupId, artifactId);
         }
+
+        this.version = version;
+    }
+
+    private static String resolveRevision(Model rawModel) {
+        final Map<String, String> props = new HashMap<>();
+        rawModel.getProperties().entrySet().forEach(e -> props.put(e.getKey().toString(), e.getValue().toString()));
+        System.getProperties().entrySet().forEach(e -> props.put(e.getKey().toString(), e.getValue().toString()));
+        return props.get(REVISION);
     }
 
     public String getGroupId() {
--- a/devtools/maven/src/main/java/io/quarkus/maven/BuildMojo.java
+++ b/devtools/maven/src/main/java/io/quarkus/maven/BuildMojo.java
@@ -22,6 +22,7 @@ import io.quarkus.bootstrap.app.AugmentAction;
 import io.quarkus.bootstrap.app.AugmentResult;
 import io.quarkus.bootstrap.app.CuratedApplication;
 import io.quarkus.bootstrap.app.QuarkusBootstrap;
+import io.quarkus.bootstrap.model.AppArtifact;
 import io.quarkus.bootstrap.resolver.maven.MavenArtifactResolver;
 
 /**
@@ -158,7 +159,13 @@ public class BuildMojo extends AbstractMojo {
                     .setRemoteRepositories(repos)
                     .build();
 
+            final Artifact projectArtifact = project.getArtifact();
+            final AppArtifact appArtifact = new AppArtifact(projectArtifact.getGroupId(), projectArtifact.getArtifactId(),
+                    projectArtifact.getClassifier(), projectArtifact.getArtifactHandler().getExtension(),
+                    projectArtifact.getVersion());
+
             CuratedApplication curatedApplication = QuarkusBootstrap.builder(outputDirectory.toPath())
+                    .setAppArtifact(appArtifact)
                     .setProjectRoot(project.getBasedir().toPath())
                     .setMavenArtifactResolver(resolver)
                     .setBaseClassLoader(BuildMojo.class.getClassLoader())
--- a/independent-projects/bootstrap/core/src/main/java/io/quarkus/bootstrap/BootstrapAppModelFactory.java
+++ b/independent-projects/bootstrap/core/src/main/java/io/quarkus/bootstrap/BootstrapAppModelFactory.java
@@ -235,30 +235,26 @@ public class BootstrapAppModelFactory {
             return createAppModelForJar(appClasses);
         }
 
-        final LocalProject localProject = isWorkspaceDiscoveryEnabled() || enableClasspathCache
-                ? loadAppClassesWorkspace()
-                : LocalProject.load(appClasses, false);
-        LocalWorkspace workspace = null;
+        LocalProject localProject = null;
         AppArtifact appArtifact = this.appArtifact;
-        if (localProject == null) {
-            log.warn("Unable to locate maven project, falling back to classpath discovery");
-            if (appArtifact == null) {
+        if (appArtifact == null) {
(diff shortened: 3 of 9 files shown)


famod:
Thanks again for the quick reaction. Any chance to get this into 1.3.0.Final?

Btw, is there a SNAPSHOT version available with this fix or do I have to build it myself?

aloubyansky:
Yes, it should end up in 1.3.0.Final. We do upload SNAPSHOT every day. I'm not sure at what time though.

aloubyansky:
It has to get merged first though to end up in a deployed SNAPSHOT.

famod:
> Yes, it should end up in 1.3.0.Final. 

Great!

> It has to get merged first though to end up in a deployed SNAPSHOT.

I see! Then I'll give it a shot once it is merged.

PS: I am getting SNAPSHOTs with different timestamps when using 999-SNAPSHOT and the test in my sample project hangs. I guess this is not normal?

independent-projects/bootstrap/core/src/main/java/io/quarkus/bootstrap/resolver/maven/workspace/LocalProject.java
```diff
@@ -28,17 +28,24 @@
 
     private static final String PROJECT_BASEDIR = "${project.basedir}";
     private static final String POM_XML = "pom.xml";
+    private static final String REVISION = "revision";
```
famod: Just a heads up: https://maven.apache.org/maven-ci-friendly.html does not only support `revision` but also `sha1` and/or `changelist`.

aloubyansky:
> PS: I am getting SNAPSHOTs with different timestamps when using 999-SNAPSHOT and the test in my sample project hangs. I guess this is not normal?

Not sure, tbh. Perhaps it's in the process of being updated?

independent-projects/bootstrap/core/src/main/java/io/quarkus/bootstrap/resolver/maven/workspace/LocalProject.java
aloubyansky: Thanks for pointing this out. I'll probably add support for those in another PR just to get this one in quicker. Unless you tell me you absolutely need them :)

famod: I'm fine with just `revision`!

independent-projects/bootstrap/core/src/main/java/io/quarkus/bootstrap/resolver/maven/workspace/LocalProject.java
```diff
@@ -136,10 +143,30 @@ private LocalProject(Model rawModel, LocalWorkspace workspace) throws BootstrapE
         this.workspace = workspace;
         this.groupId = ModelUtils.getGroupId(rawModel);
         this.artifactId = rawModel.getArtifactId();
-        this.version = ModelUtils.getVersion(rawModel);
+
+        String version = ModelUtils.getVersion(rawModel);
+        final boolean revisionVersion;
+        if (revisionVersion = REVISION_EXPR.equals(version)) {
+            version = rawModel.getProperties().getProperty(REVISION);
```
famod: Does this also cover `-Drevision=<something_different_than_in_pom>`?
This is important, because that is what CI (e.g. Jenkins via `Jenkinsfile`) would be passing to `mvn` to override the default revision.

A variation is of that would be:
- ```xml
  <properties>
      <revision>0.1-${revisionSuffix}</revision>
      <revisionSuffix>SNAPSHOT</revisionSuffix>
  </properties>
  ```
- CI passes `-DrevisionSuffix=build-${BUILD_NUMBER}`

This way local build produces `0.1-SNAPSHOT` and CI produces something like `0.1-build-123`.

geoand:
@aloubyansky is this safe for `1.3.0.Final`?

aloubyansky:
Yes, which is why I requested a backport. However, I'm about to commit one more enhancement in a bit.

independent-projects/bootstrap/core/src/main/java/io/quarkus/bootstrap/resolver/maven/workspace/LocalProject.java
aloubyansky: Actually according to the Maven doc you referenced this is not supported by Maven itself, isn't it?

aloubyansky: I tried adding support for sha1 and changelist but strangely it also didn't work in Maven itself. Unless I somehow did it wrong.

aloubyansky: > Does this also cover `-Drevision=<something_different_than_in_pom>`?

This will work though.

gsmet:
Please ping us when this is ready.

aloubyansky:
It's ready @gsmet I can always follow up with enhancements.

gastaldi (approved):

independent-projects/bootstrap/core/src/main/java/io/quarkus/bootstrap/resolver/maven/workspace/LocalProject.java
famod: > Actually according to the Maven doc you referenced this is not supported by Maven itself, isn't it?

Which part do you mean? `-Drevision=...` is mentioned explicitly multiple times.

The `revisionSuffix` is not mentioned but IMHO this is just basic Maven interpolation logic, no?
You could also have `myFancySomethingProperty` instead of `revisionSuffix`, doesn't matter.

> I tried adding support for sha1 and changelist but strangely it also didn't work in Maven itself.

I swapped `revision` with `sha1` locally in my sample project and it worked (apart from the Quarkus build problem). I even changed the default value from `1.0-SNAPSHOT` to `0.2-SNAPSHOT` and I also tried `-Drevision=1.1-SNAPSHOT`.

> > Does this also cover `-Drevision=<something_different_than_in_pom>`?
> 
> This will work though.

Ok, as long as this works the main problem is solved. The rest could be adressed separately.

aloubyansky: > You could also have `myFancySomethingProperty` instead of `revisionSuffix`, doesn't matter.

Arbitrary property interpolation is not allowed in the version elements. There is an example in that doc.

> I swapped `revision` with `sha1` locally in my sample project and it worked

Ah, I need to try that. I tried combining the three mentioned properties like they had in the doc, which for some reason didn't work.

famod: > > You could also have `myFancySomethingProperty` instead of `revisionSuffix`, doesn't matter.
> 
> Arbitrary property interpolation is not allowed in the version elements. There is an example in that doc.

You are right, I have totally missed that!

> > I swapped `revision` with `sha1` locally in my sample project and it worked
> 
> Ah, I need to try that. I tried combining the three mentioned properties like they had in the doc, which for some reason didn't work.

I just tried the following and everything works as expected (in Maven itself):
```diff
$ git diff
diff --git a/core/pom.xml b/core/pom.xml
index ea45976..b52f070 100644
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -5,7 +5,7 @@
   <parent>
     <groupId>com.github.famod.modmono-quarkus</groupId>
     <artifactId>modmono-quarkus-parent</artifactId>
-    <version>${revision}</version>
+    <version>${revision}${sha1}</version>
   </parent>
   <artifactId>modmono-quarkus-core</artifactId>
   <dependencies>
diff --git a/dist/pom.xml b/dist/pom.xml
index 278e9c5..e0df457 100644
--- a/dist/pom.xml
+++ b/dist/pom.xml
@@ -5,7 +5,7 @@
   <parent>
     <groupId>com.github.famod.modmono-quarkus</groupId>
     <artifactId>modmono-quarkus-parent</artifactId>
-    <version>${revision}</version>
+    <version>${revision}${sha1}</version>
   </parent>
   <artifactId>modmono-quarkus-dist</artifactId>
   <dependencies>
diff --git a/pom.xml b/pom.xml
index b895c66..f72103c 100644
--- a/pom.xml
+++ b/pom.xml
@@ -4,10 +4,11 @@
   <modelVersion>4.0.0</modelVersion>
   <groupId>com.github.famod.modmono-quarkus</groupId>
   <artifactId>modmono-quarkus-parent</artifactId>
-  <version>${revision}</version>
+  <version>${revision}${sha1}</version>
   <packaging>pom</packaging>
   <properties>
-    <revision>1.0-SNAPSHOT</revision>
+    <revision>0.3</revision>
+    <sha1>-SNAPSHOT</sha1>

     <compiler-plugin.version>3.8.1</compiler-plugin.version>
```

PR #7793, opened 2020-03-11
portable banner

maxandersen:
after lots of back and forth I'm saying we go for banner that only uses ascii chars as otherwise non-US windows users using gradle will get blank or weird characters.

included banner here is more portable and smaller than the competition :)


maxandersen:
Screenshots from both osx and windows.

maxandersen:
ignore the WAZZUP in the screenshot :)

geoand:
Looks good to me.

So should we merge and be done with it?

maxandersen:
@geoand sounds good to me, will you +1 review it ? :)

geoand (approved):

geoand:
Done :)

PR #7797, opened 2020-03-11
fix(docs): make sure that mapped properties are sorted properly

### LINKED ISSUE #7781: Generated datasource configuration: incorrect order of properties
When generating the datasource configuration, there is an issue with the additional datasources (the ones defined with a name): the build time properties are not first but last.

See `datasource.html#quarkus-agroal_quarkus.datasource.named-data-sources`.

And the following captures:
capture1

[...]

capture2

I checked the annotations and they look consistent but we are doing a weird thing to have a `jdbc` prefix.

@machi1990 could you have a look?

machi1990:
Use the config phase comparator even for mapped properties to make sure that config items in a Map
are sorted in correct ordering - build time items first and runtime items at the end

Fixes https://github.com/quarkusio/quarkus/issues/7781

### DIFF
--- a/core/processor/src/main/java/io/quarkus/annotation/processor/generate_doc/ConfigDocElement.java
+++ b/core/processor/src/main/java/io/quarkus/annotation/processor/generate_doc/ConfigDocElement.java
@@ -1,5 +1,7 @@
 package io.quarkus.annotation.processor.generate_doc;
 
+import static io.quarkus.annotation.processor.generate_doc.ConfigPhase.COMPARATOR;
+
 import java.io.IOException;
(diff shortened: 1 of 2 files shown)


machi1990:
@gsmet can you try this? I checked the output and it should be okay but would need a second eye.

gsmet (approved):
It did fix my issue.

machi1990:
Thanks @gsmet for taking a look and for the review.

gsmet:
Thanks for the fix. I'll wait for CI to be extra sure and merge.

gastaldi (approved):
LGTM

PR #7808, opened 2020-03-12
Exclude protobuf-java from mysql-connector-jdbc

michalszynkiewicz:
It seems probuf-java is needed only for Protocol X

See https://dev.mysql.com/doc/connector-j/8.0/en/connector-j-installing-maven.html


gsmet (approved):

gsmet:
Let's merge it and see if people complain.

PR #7819, opened 2020-03-12
Fix a few things in the container-image documentation

gsmet:
@geoand could you have a look?

I started this as I though it was the one where I needed to include the generated doc but it turned out I was editing the wrong file :).

Hopefully, I will get the other doc fixed soon.

### DIFF
--- a/docs/src/main/asciidoc/container-image.adoc
+++ b/docs/src/main/asciidoc/container-image.adoc
@@ -19,18 +19,15 @@ Quarkus provides extensions for building (and pushing) container images. Current
 === JIB
 
 The extension `quarkus-container-image-jib` is powered by https://github.com/GoogleContainerTools/jib[Jib] for performing container image builds.
-The major benefit of using Jib with Quarkus is that all dependencies (everything found under target/lib) are cached in a different layer than the actual application making rebuilds really fast and small (when it comes to pushing).
+The major benefit of using Jib with Quarkus is that all the dependencies (everything found under `target/lib`) are cached in a different layer than the actual application making rebuilds really fast and small (when it comes to pushing).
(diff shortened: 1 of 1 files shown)


docs/src/main/asciidoc/container-image.adoc
```diff
@@ -19,18 +19,15 @@ Quarkus provides extensions for building (and pushing) container images. Current
 === JIB
 
 The extension `quarkus-container-image-jib` is powered by https://github.com/GoogleContainerTools/jib[Jib] for performing container image builds.
-The major benefit of using Jib with Quarkus is that all dependencies (everything found under target/lib) are cached in a different layer than the actual application making rebuilds really fast and small (when it comes to pushing).
+The major benefit of using Jib with Quarkus is that all the dependencies (everything found under `target/lib`) are cached in a different layer than the actual application making rebuilds really fast and small (when it comes to pushing).
 Another important benefit of using this extension is that it provides the ability to create a container image without having to have any dedicated client side tooling (like Docker) or running daemon processes (like the Docker daemon)
 when all that is needed is the ability to push to a container image registry.
 
-To use this image extension, you just need to add the following dependency into your project.
+To use this feature, add the following extension to your project:
```
geoand: I have no problem whatsoever doing this, but it seems like a lot of other guides don't

geoand (approved):
Makes sense. I added a comment about consistence with other guides that we can take up later

docs/src/main/asciidoc/container-image.adoc
gsmet: It should be how we do things IMHO. Some of the guides have that command and then an extract from the pom.

geoand: Yeah that's cool with me. We should just make a pass at some point to make all guides you this style.

PR #7822, opened 2020-03-12
Make duration in Kubernetes extension config proper durations

gsmet:
/cc @geoand @iocanel 

I didn't have the time to think about using the generated doc, considering there are a lot of additional info in there. We might need some doc generation tweaks.

We can do that in 1.3.1 but we need to fix the config properties to be consistent with the Quarkus way of doing things.

### DIFF
--- a/docs/src/main/asciidoc/kubernetes.adoc
+++ b/docs/src/main/asciidoc/kubernetes.adoc
@@ -326,8 +326,8 @@ To set the initial delay of the probe to 20 seconds and the period to 45:
 
 [source]
 ----
-quarkus.kubernetes.readiness-probe.initial-delay-seconds=20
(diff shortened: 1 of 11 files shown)


geoand:
This is really good, but let me take a quick look and make sure there aren't any other instances that need to be changed

gsmet:
I tried to be thorough but second eye very welcome.

geoand:
I also need to check and see if the old config properties work

iocanel (approved):
LGTM

geoand:
@iocanel The commit I added is what we discussed

geoand (approved):
Approving (although I pushed a commit as well) so the PR can be merged

iocanel (approved):
The probe fixes also look good.

PR #7830, opened 2020-03-12
Add excluded artifacts for Spring

geoand:


stuartwdouglas (changes requested):
I think these should actually be in the relevant spring extensions. You should only exclude something if you are providing a replacement.

geoand:
PR updated, thanks for the info!

stuartwdouglas (approved):

gsmet (changes requested):
Maybe we should also add them to the banned dependencies of the enforcer?

geoand:
@gsmet done

gsmet (approved):

gsmet:
Perfect, thanks! Let's wait for CI!

geoand:
CI failure is totally unrelated

PR #7837, opened 2020-03-13
ArC - transform unproxyable bean classes by default

mkouba:
- new transformations: remove final flag from a bean class, change
visibility of private no-args constructor
- deprecate quarkus.arc.remove-final-for-proxyable-methods
- introduce quarkus.arc.transfrorm-unproxyable-classes
- this should also help kotlin developers where classes are final by
default


geoand:
I can't really tell if the TCK failure is related or not... The error is rather obscure. Can you give it a check @mkouba ?

mkouba:
That's a bug in the TCK: https://github.com/eclipse/microprofile-config/issues/543

We do transform the [ConfigObserver](https://github.com/eclipse/microprofile-config/blob/master/tck/src/main/java/org/eclipse/microprofile/config/tck/broken/ConfigObserver.java#L40) class but there is obviously some class loading problem, most probably related to our arquillian adapter.

I think I'll try to disable the transformation for MP Config TCK fow now...

geoand:
OK, thanks

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/ArcConfig.java
```diff
@@ -54,13 +52,25 @@
...
      * but have been declared as final. The transformation is simply a matter of removing final.
-     * This ensures that a proxy can be created properly.
+     * This ensures that a proxy/subclass can be created properly.
      * If the value is set to false, then an exception is thrown at build time indicating
      * that a proxy could not be created because a method was final.
+     * 
+     * @deprecated This config property is ignored and will be removed at some point post Quarkus 1.4
+     * @see #transformUnproxyableClasses
      */
+    @Deprecated
```
geoand:
```suggestion
    @Deprecated(forRemoval=true)
```

mkouba: Hm, `forRemoval` is JDK9+. Is it required to build quarkus? Our pom still targets 1.8... and I'm still using JDK8 ;-).

geoand: Ah damn it... You are right I just switched to JDK 11 in my IDE and didn't realize that this was a JDK 8+ thing since I was so used to viewing JDK 8 sources :)

independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/BeanProcessor.java
```diff
@@ -403,8 +413,26 @@ public Builder addRemovalExclusion(Predicate<BeanInfo> exclusion) {
...
+         * @deprecated This method will be removed at some point post Quarkus 1.4
+         */
+        @Deprecated
         public Builder setRemoveFinalFromProxyableMethods(boolean removeFinalForProxyableMethods) {
-            this.removeFinalForProxyableMethods = removeFinalForProxyableMethods;
+            return this;
+        }
+
+        /**
+         * If set to true the container will attempt to remove the final flag from a bean class that must be proxyable during
```
geoand: I think this comment needs to be augmented right? Since the transformation isn't only about removing final anymore

extensions/arc/deployment/src/test/java/io/quarkus/arc/test/unproxyable/PrivateNoArgsConstructorDoNotChangeFlagTest.java
```diff
@@ -0,0 +1,48 @@
...
+import org.jboss.shrinkwrap.api.asset.StringAsset;
+import org.jboss.shrinkwrap.api.spec.JavaArchive;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.RegisterExtension;
+
+import io.quarkus.arc.Unremovable;
+import io.quarkus.test.QuarkusUnitTest;
+
+public class PrivateNoArgsConstructorDoNotChangeFlagTest {
```
geoand: These tests are great! I think that having a extra test where the bean extends some class that doesn't have a no-args constructor would be beneficial - to ensure that a build error is encountered

mkouba: There is/was `FailedToAddMissingNoargsConstructorTest`... is that what you mean?

geoand: Oh yes! That is exactly what I meant. I missed that one

geoand:
I like it a lot, +1. The implementation also looks straight-forward but I'll let @manovotn comment on that since this is his wheelhouse :)

stuartwdouglas (approved):

geoand (approved):

independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/BeanDeploymentValidator.java
```diff
@@ -23,11 +23,14 @@ default void validate(ValidationContext context) {
...
 
     /**
+     * This method is deprecated and is not used internally.
      * 
      * @param target
      * @param rule
      * @return {@code true} if the given validation rule should be skipped for the specified target
+     * @deprecated This method will be removed at some point post Quarkus 1.4
      */
+    @Deprecated
```
manovotn: Do we have a substitute for this now or do we just drop it?

independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/BeanProcessor.java
```diff
@@ -223,16 +232,17 @@ public BeanDeployment getBeanDeployment() {
...
     public BeanDeployment process() throws IOException {
-        registerCustomContexts();
-        registerBeans();
-        registerSyntheticObservers();
-        initialize(new Consumer<BytecodeTransformer>() {
+        Consumer<BytecodeTransformer> unsupportedBytecodeTransformer = new Consumer<BytecodeTransformer>() {
             @Override
             public void accept(BytecodeTransformer transformer) {
-
+                throw new UnsupportedOperationException();
```
manovotn: Why unsupported as opposed to just no-op? Also a comment would be nice to know that this is only for test container and is not going to be actually invoked.

independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/BeanProcessor.java
```diff
@@ -403,8 +413,25 @@ public Builder addRemovalExclusion(Predicate<BeanInfo> exclusion) {
             return this;
         }
 
+        /**
+         * 
+         * @param removeFinalForProxyableMethods
+         * @return self
+         * @deprecated This method will be removed at some point post Quarkus 1.4
+         */
```
manovotn: Add `@see `setRemoveFinalFromProxyableMethods(boolean)`.

manovotn (commented):
Looks really good, this actually fixes quite a few proxyability cases where we wouldn't pass TCK's exception expectations (deployment versus definition) :-)
Added just some minor questions/remarks.

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/ArcConfig.java
```diff
@@ -37,7 +36,6 @@
      * <li>does not declare any producer which is eligible for injection to any injection point,</li>
      * <li>is not directly eligible for injection into any {@link javax.enterprise.inject.Instance} injection point</li>
      * </ul>
-     * </p>
```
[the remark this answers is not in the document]
mkouba: That's ok in javadoc....

independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/BeanDeploymentValidator.java
mkouba: I think that we should just drop it.

independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/BeanProcessor.java
mkouba: "no-op" would just cause a problem later... In theory, it can be used for non-test scenarios as well. From my POV it's just illegal state if `transformUnproxyableClasses=true` and no valid `Consumer<BytecodeTransformer>` is provided.

independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/BeanProcessor.java
mkouba: Did you mean `@see #setTransformUnproxyableClasses()`?

manovotn: Yea, that's what I meant :-D

independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/BeanDeploymentValidator.java
manovotn: Probably yes, I was just wondering since the deprecation doesn't mention it.

independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/BeanProcessor.java
manovotn: >  it's just illegal state if transformUnproxyableClasses=true and no valid Consumer<BytecodeTransformer> is provided.

Fair enough, that makes sense.

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/ArcConfig.java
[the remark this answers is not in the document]
manovotn: I know, just noted it here since I caught a glimpse of it. Doesn't really matter :)

manovotn (approved):

geoand:
Github UI is reporting a conflict

mkouba:
@geoand Yep, just in the middle of rebasing and testing ;-).

geoand:
Given how much this will improve the Kotlin experience (especially on Gradle where we can't yet propagate the correct plugins to dev mode), perhaps we should backport this to 1.3.x?
WDYT @mkouba @gsmet ?

geoand:
I am actually thinking that this type of transformation could be applied in other areas that affect Kotlin as well...
That way we will be able to solve a lot of Kotlin issues in the Quarkus codebase itself and not have to rely on glue code in the build tool plugins which we may or may not be able to do.

manovotn:
> perhaps we should backport this to 1.3.x?

I am not working in Kotlin so I cannot say how much of an issue is this. But if it's a biggie then I see no reason not to backport it.

geoand:
At the moment, Kotlin + Gradle (which is what most Kotlin users use), does not work nice at all in dev-mode.
This PR really improves the situation.

We always knew this, but after actually witnessing this in action I am convinced that this approach is great.

mkouba:
+1 for the approach but I'm not so sure about backporting. (1) it's a breaking change. (2) we modified a lot of files so it might not be that easy to backport this PR.

geoand:
I see, in that case if we are not sure about the feasibility of backporting, then I won't insist :)

stuartwdouglas:
If possible I would like to backport, but I understand that it might not be practical.

PR #5489, opened 2019-11-14
fix Easy Config-related cleanups #5388

Simulant87:
fixing #5388

PR #7849, opened 2020-03-13
Fixes #5388 : Easy Config-related cleanups

### LINKED ISSUE #5388: Easy Config-related cleanups
**Description**
After #5387, there are several configuration constructs that can be cleaned up in core and extensions.

- [x] Any `@ConfigItem` which has a primitive type like `int` or `boolean` that has an explicit default value of `0` or `false` can have the default value removed.
- [x] Any `@ConfigItem` with a primitive wrapper type like `Integer` or `Boolean` with an explicit default value of `0` or `false` should be changed to use the primitive type instead and remove the default value.
- [ ] Any `@ConfigItem` of type `Optional<Integer>` can be changed to `OptionalInt`.
- [ ] Any `@ConfigItem` which is a `List` type that is always converted to an array using `toArray` can be changed to just be an array type instead.

ShubhamRwt:
Rebased PR (#5489)

### DIFF
--- a/core/deployment/src/main/java/io/quarkus/deployment/JniProcessor.java
+++ b/core/deployment/src/main/java/io/quarkus/deployment/JniProcessor.java
@@ -29,9 +29,9 @@ public class JniProcessor {
         /**
          * @deprecated JNI is always enabled starting from GraalVM 19.3.1.
          */
-        @Deprecated
-        @ConfigItem(defaultValue = "true")
-        boolean enable = true;
+
+        @ConfigItem
+        boolean enable;
     }
 
     @BuildStep
--- a/core/deployment/src/main/java/io/quarkus/deployment/pkg/NativeConfig.java
+++ b/core/deployment/src/main/java/io/quarkus/deployment/pkg/NativeConfig.java
@@ -26,26 +26,30 @@ public class NativeConfig {
     /**
      * If the HTTPS url handler should be enabled, allowing you to do URL.openConnection() for HTTPS URLs
      */
-    @ConfigItem(defaultValue = "false")
+    @ConfigItem
     public boolean enableHttpsUrlHandler;
 
     /**
      * If all security services should be added to the native image
      */
-    @ConfigItem(defaultValue = "false")
+    @ConfigItem
     public boolean enableAllSecurityServices;
 
     /**
      * @deprecated JNI is always enabled starting from GraalVM 19.3.1.
      */
+<<<<<<< HEAD
     @Deprecated
     @ConfigItem(defaultValue = "true")
+=======
+    @ConfigItem
+>>>>>>> fix config-related cleanups #5388
     public boolean enableJni;
 
     /**
      * If all character sets should be added to the native image. This increases image size
      */
-    @ConfigItem(defaultValue = "false")
+    @ConfigItem
     public boolean addAllCharsets;
 
     /**
@@ -75,14 +79,14 @@ public class NativeConfig {
     /**
      * If debug symbols should be included
      */
-    @ConfigItem(defaultValue = "false")
+    @ConfigItem
     public boolean debugSymbols;
 
     /**
      * If the native image build should wait for a debugger to be attached before running. This is an advanced option
      * and is generally only intended for those familiar with GraalVM internals
      */
-    @ConfigItem(defaultValue = "false")
+    @ConfigItem
     public boolean debugBuildProcess;
 
     /**
@@ -94,7 +98,7 @@ public class NativeConfig {
     /**
      * If the native image server should be restarted
      */
-    @ConfigItem(defaultValue = "false")
+    @ConfigItem
     public boolean cleanupServer;
 
     /**
@@ -107,33 +111,33 @@ public class NativeConfig {
      * If a JVM based 'fallback image' should be created if native image fails. This is not recommended, as this is
      * functionally the same as just running the application in a JVM
      */
-    @ConfigItem(defaultValue = "false")
+    @ConfigItem
     public boolean enableFallbackImages;
 
     /**
      * If the native image server should be used. This can speed up compilation but can result in changes not always
      * being picked up due to cache invalidation not working 100%
      */
-    @ConfigItem(defaultValue = "false")
+    @ConfigItem
     public boolean enableServer;
 
     /**
      * If all META-INF/services entries should be automatically registered
      */
-    @ConfigItem(defaultValue = "false")
+    @ConfigItem
     public boolean autoServiceLoaderRegistration;
 
     /**
      * If the bytecode of all proxies should be dumped for inspection
      */
-    @ConfigItem(defaultValue = "false")
+    @ConfigItem
     public boolean dumpProxies;
 
     /**
      * If this build should be done using a container runtime. If this is set docker will be used by default,
      * unless container-runtime is also set.
      */
-    @ConfigItem(defaultValue = "false")
+    @ConfigItem
     public boolean containerBuild;
 
     /**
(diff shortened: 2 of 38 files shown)


extensions/flyway/deployment/src/main/java/io/quarkus/flyway/FlywayProcessor.java
```diff
@@ -12,14 +12,7 @@
...
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.Enumeration;
-import java.util.HashMap;
-import java.util.LinkedHashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
+import java.util.*;
 import java.util.stream.Collectors;
```
kunal-kushwaha: Can you please remove wildcard imports?

core/deployment/src/main/java/io/quarkus/deployment/JniProcessor.java
```diff
@@ -29,9 +29,9 @@
         /**
          * @deprecated JNI is always enabled starting from GraalVM 19.3.1.
          */
-        @Deprecated
-        @ConfigItem(defaultValue = "true")
-        boolean enable = true;
+
+        @ConfigItem
+        boolean enable;
```
dmlloyd: This changes the default value from `true` to `false`; we don't want that! I'd suggest to leave this one alone; it was correct already.

core/deployment/src/main/java/io/quarkus/deployment/pkg/NativeConfig.java
```diff
@@ -26,26 +26,26 @@
...
+    @ConfigItem
     public boolean enableAllSecurityServices;
 
     /**
      * @deprecated JNI is always enabled starting from GraalVM 19.3.1.
      */
-    @Deprecated
-    @ConfigItem(defaultValue = "true")
+
+    @ConfigItem
```
dmlloyd: Same here.  Only remove the default values if they're `false` or `0`.

core/test-extension/runtime/src/main/java/io/quarkus/extest/runtime/config/AllValuesConfig.java
```diff
@@ -30,7 +29,7 @@
     public OptionalDouble optDoubleValue;
     /** an optional long value */
     @ConfigItem
-    public Optional<Long> optionalLongValue;
```
dmlloyd: This should not be changed: it is like this deliberately in order to test `Optional`.  If we change it to `OptionalLong` then we won't be testing that anymore.

extensions/flyway/runtime/src/main/java/io/quarkus/flyway/runtime/FlywayDataSourceBuildTimeConfig.java
```diff
@@ -32,5 +31,5 @@ public static final FlywayDataSourceBuildTimeConfig defaultConfig() {
      * scanned recursively down non-hidden directories.
      */
     @ConfigItem(defaultValue = DEFAULT_LOCATION)
-    public List<String> locations;
+    public Optional<String[]> locations;
```
dmlloyd: This one shouldn't change to an array (almost every usage of this property treats it as a list).  And, it wasn't `Optional` before so it should not become `Optional` now.

extensions/hibernate-orm/deployment/src/main/java/io/quarkus/hibernate/orm/deployment/HibernateOrmProcessor.java
```diff
@@ -633,13 +633,15 @@ private void handleHibernateORMWithNoPersistenceXml(
                 hibernateConfig.jdbc.timezone.ifPresent(
                         timezone -> desc.getProperties().setProperty(AvailableSettings.JDBC_TIME_ZONE, timezone));
 
-                hibernateConfig.jdbc.statementFetchSize.ifPresent(
-                        fetchSize -> desc.getProperties().setProperty(AvailableSettings.STATEMENT_FETCH_SIZE,
-                                fetchSize.toString()));
+                if (hibernateConfig.jdbc.statementFetchSize.isPresent()) {
+                    desc.getProperties().setProperty(AvailableSettings.STATEMENT_FETCH_SIZE,
+                            String.valueOf(hibernateConfig.jdbc.statementFetchSize.getAsInt()));
+                }
```
dmlloyd: Maybe instead go with the original but slightly modified:
```suggestion
                hibernateConfig.jdbc.statementFetchSize.ifPresent(
                        fetchSize -> desc.getProperties().setProperty(AvailableSettings.STATEMENT_FETCH_SIZE,
                                Integer.toString(fetchSize));
```
Smaller diffs are usually preferred.

extensions/hibernate-orm/deployment/src/main/java/io/quarkus/hibernate/orm/deployment/HibernateOrmProcessor.java
```diff
@@ -633,13 +633,15 @@ private void handleHibernateORMWithNoPersistenceXml(
...
 
-                hibernateConfig.jdbc.statementBatchSize.ifPresent(
-                        fetchSize -> desc.getProperties().setProperty(AvailableSettings.STATEMENT_BATCH_SIZE,
-                                fetchSize.toString()));
+                if (hibernateConfig.jdbc.statementBatchSize.isPresent()) {
+                    desc.getProperties().setProperty(AvailableSettings.STATEMENT_BATCH_SIZE,
+                            String.valueOf(hibernateConfig.jdbc.statementBatchSize.getAsInt()));
+                }
```
dmlloyd: Same here.

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/ServerSslConfig.java
```diff
@@ -22,7 +22,7 @@
      * The cipher suites to use. If none is given, a reasonable default is selected.
      */
     @ConfigItem
-    public Optional<List<String>> cipherSuites;
+    public Optional<String[]> cipherSuites;
```
dmlloyd: This one doesn't need to change: it's being used as a list, not an array.

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/cors/CORSConfig.java
```diff
@@ -31,7 +31,7 @@
      * default: returns any requested method as valid
      */
     @ConfigItem
-    public Optional<List<HttpMethod>> methods;
+    public Optional<HttpMethod[]> methods;
```
dmlloyd: These ones don't need to change, since they're being used as lists not arrays.

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/cors/CORSConfig.java
```diff
@@ -42,7 +42,7 @@
      * default: returns any requested header as valid
      */
     @ConfigItem
-    public Optional<List<String>> headers;
+    public Optional<String[]> headers;
```
dmlloyd: These ones don't need to change, since they're being used as lists not arrays.

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/cors/CORSConfig.java
```diff
@@ -52,7 +52,7 @@
      * default: empty
      */
     @ConfigItem
-    public Optional<List<String>> exposedHeaders;
+    public Optional<String[]> exposedHeaders;
```
dmlloyd: These ones don't need to change, since they're being used as lists not arrays.

dmlloyd (changes requested):

gsmet:
Closing this one. Too many conflicts. Please open a new PR if you want to pursue.

PR #7851, opened 2020-03-13
Do not print null in generated config

gastaldi:


geoand:
Makes sense to me although I haven't really used the generate config feature

aloubyansky (approved):

PR #7860, opened 2020-03-13
Add healthcheck for reactive datasources

machi1990:
the datasource config had a health check enabled option which was only consumed by Agroal Datasource
and not the reactive ones. Let's use this option for reactive datasource too and send appropriate
check status when database is down.

### DIFF
--- a/extensions/reactive-mysql-client/runtime/src/main/java/io/quarkus/reactive/mysql/client/runtime/health/ReactiveMySQLDataSourceHealthCheck.java
+++ b/extensions/reactive-mysql-client/runtime/src/main/java/io/quarkus/reactive/mysql/client/runtime/health/ReactiveMySQLDataSourceHealthCheck.java
@@ -1,0 +1,48 @@
+package io.quarkus.reactive.mysql.client.runtime.health;
+
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.TimeUnit;
+
+import javax.annotation.PostConstruct;
+import javax.enterprise.context.ApplicationScoped;
+
+import org.eclipse.microprofile.health.HealthCheck;
+import org.eclipse.microprofile.health.HealthCheckResponse;
+import org.eclipse.microprofile.health.HealthCheckResponseBuilder;
+import org.eclipse.microprofile.health.Readiness;
+
+import io.quarkus.arc.Arc;
+import io.vertx.mysqlclient.MySQLPool;
+
+@Readiness
+@ApplicationScoped
+public class ReactiveMySQLDataSourceHealthCheck implements HealthCheck {
+
+    private MySQLPool mySQLPool;
+
+    @PostConstruct
+    protected void init() {
+        mySQLPool = Arc.container().instance(MySQLPool.class).get();
+    }
+
+    @Override
+    public HealthCheckResponse call() {
+        HealthCheckResponseBuilder builder = HealthCheckResponse.named("Reactive MySQL connection health check").up();
+
+        try {
+            CompletableFuture<Void> databaseConnectionAttempt = new CompletableFuture<>();
+            mySQLPool.query("SELECT 1")
+                    .execute(ar -> {
+                        if (ar.failed()) {
+                            builder.down();
+                        }
+                        databaseConnectionAttempt.complete(null);
+                    });
+            databaseConnectionAttempt.get(10, TimeUnit.SECONDS);
+        } catch (Exception exception) {
+            builder.down();
+        }
+
+        return builder.build();
+    }
+}
--- a/extensions/reactive-pg-client/runtime/src/main/java/io/quarkus/reactive/pg/client/runtime/health/ReactivePgDataSourceHealthCheck.java
+++ b/extensions/reactive-pg-client/runtime/src/main/java/io/quarkus/reactive/pg/client/runtime/health/ReactivePgDataSourceHealthCheck.java
@@ -1,0 +1,47 @@
+package io.quarkus.reactive.pg.client.runtime.health;
+
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.TimeUnit;
+
+import javax.annotation.PostConstruct;
+import javax.enterprise.context.ApplicationScoped;
+
+import org.eclipse.microprofile.health.HealthCheck;
+import org.eclipse.microprofile.health.HealthCheckResponse;
+import org.eclipse.microprofile.health.HealthCheckResponseBuilder;
+import org.eclipse.microprofile.health.Readiness;
+
+import io.quarkus.arc.Arc;
+import io.vertx.pgclient.PgPool;
+
+@Readiness
+@ApplicationScoped
+public class ReactivePgDataSourceHealthCheck implements HealthCheck {
+    private PgPool pgPool;
+
+    @PostConstruct
+    protected void init() {
+        pgPool = Arc.container().instance(PgPool.class).get();
+    }
+
+    @Override
+    public HealthCheckResponse call() {
+        HealthCheckResponseBuilder builder = HealthCheckResponse.named("Reactive PostgreSQL connection health check").up();
+
+        try {
+            CompletableFuture<Void> databaseConnectionAttempt = new CompletableFuture<>();
+            pgPool.query("SELECT 1")
+                    .execute(ar -> {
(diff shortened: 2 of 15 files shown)


jmartisk (approved):

extensions/reactive-mysql-client/runtime/src/main/java/io/quarkus/reactive/mysql/client/runtime/health/ReactiveMySQLDataSourceHealthCheck.java
```diff
@@ -0,0 +1,46 @@
...
+
+@Readiness
+@ApplicationScoped
+public class ReactiveMySQLDataSourceHealthCheck implements HealthCheck {
+
+    private MySQLPool mySQLPool;
+
+    @PostConstruct
+    protected void init() {
+        mySQLPool = Arc.container().instance(MySQLPool.class).get();
```
cescoffier: can't we get more than one?

extensions/reactive-mysql-client/runtime/src/main/java/io/quarkus/reactive/mysql/client/runtime/health/ReactiveMySQLDataSourceHealthCheck.java
```diff
@@ -0,0 +1,46 @@
...
+
+    @Override
+    public HealthCheckResponse call() {
+        HealthCheckResponseBuilder builder = HealthCheckResponse.named("Reactive MySQL connection health check").up();
+
+        try {
+            CompletableFuture<Void> databaseConnectionAttempt = new CompletableFuture<>();
+            mySQLPool.query("SELECT 1", ar -> {
+                if (ar.failed()) {
+                    builder.down();
```
cescoffier: You need to check you are not on the IO Thread, because if so you cannot block.

extensions/reactive-pg-client/runtime/src/main/java/io/quarkus/reactive/pg/client/runtime/health/ReactivePgDataSourceHealthCheck.java
```diff
@@ -0,0 +1,45 @@
...
+
+        try {
+            CompletableFuture<Void> databaseConnectionAttempt = new CompletableFuture<>();
+            pgPool.query("SELECT 1", ar -> {
+                if (ar.failed()) {
+                    builder.down();
+                }
+                databaseConnectionAttempt.complete(null);
+            });
+            databaseConnectionAttempt.join();
```
cescoffier: Same comments:

* how to handle when there are several Pools
* you need to be sure to not block the IO Thread.

cescoffier (dismissed):

extensions/reactive-pg-client/runtime/src/main/java/io/quarkus/reactive/pg/client/runtime/health/ReactivePgDataSourceHealthCheck.java
machi1990: @cescoffier thanks for the review.

> Same comments:
> 
> * how to handle when there are several Pools

In the meantime multiple reactive datasources are not handled, once added we will for sure need to handle them the same way we have handled them for the Agroal connection healthcheck. 

> * you need to be sure to not block the IO Thread.

Okay,  are there constructs (a bean that I can inject ) that I can use to check if in IO Thread? 
And if in IO Thread case, what should be done?

jmartisk: I think this will always be in the IO thread
Is this still a problem even if we  mark the Health handler as blocking though? https://github.com/quarkusio/quarkus/blob/1.3.0.Final/extensions/smallrye-health/deployment/src/main/java/io/quarkus/smallrye/health/deployment/SmallRyeHealthProcessor.java#L106

jmartisk: If it's a problem then I suppose we would need a change to run the health checks (probably all, not just this one) in a different thread pool (which one?)

machi1990: > I think this will always be in the IO thread
> Is this still a problem even if we mark the Health handler as blocking though? https://github.com/quarkusio/quarkus/blob/1.3.0.Final/extensions/smallrye-health/deployment/src/main/java/io/quarkus/smallrye/health/deployment/SmallRyeHealthProcessor.java#L106

Thanks @jmartisk for the clarifications and pointers.

 


> If it's a problem then I suppose we would need a change to run the health checks (probably all, not just this one) in a different thread pool (which one?)

Yes, I think so too. @cescoffier WDYT? Preferably we should do it in another followup PR. I can take it if given enough inputs.

jmartisk: +1 for moving this to another PR

machi1990: @cescoffier ping

cescoffier: yes, it's a blocking route. Still, a blocking route should not be blocked indefinitely. 
I would recommend adding a timeout (10 seconds) because we may not answer at all. The client (Kubernetes) would get an HTTP timeout, but we would consume and waste one of the threads of the worker thread pool.

jmartisk: Where should we ideally process it though? Does the health extension need to maintain its own thread pool or is there a common thread pool that would be appropriate to use?

machi1990: @cescoffier I have added a 10s timeout.

+1 with @jmartisk question up there. Should help with the next step.

machi1990: @cescoffier do you have sometime to have another look and see if all comments have been addressed? Thanks.

extensions/reactive-pg-client/runtime/src/main/java/io/quarkus/reactive/pg/client/runtime/health/ReactivePgDataSourceHealthCheck.java
```diff
@@ -0,0 +1,46 @@
...
+        pgPool = Arc.container().instance(PgPool.class).get();
+    }
+
+    @Override
+    public HealthCheckResponse call() {
+        HealthCheckResponseBuilder builder = HealthCheckResponse.named("Reactive PostgreSQL connection health check").up();
+
+        try {
+            CompletableFuture<Void> databaseConnectionAttempt = new CompletableFuture<>();
+            pgPool.query("SELECT 1", ar -> {
```
cescoffier, line 33: Does "SELECT 1" trigger a roundtrip with the database? 
Just want to be sure.

cescoffier (commented):
Looks good to me, I just want one last verification. Does "SELECT 1" trigger a round trip to the database? @vietj do you know by any chance?

extensions/reactive-pg-client/runtime/src/main/java/io/quarkus/reactive/pg/client/runtime/health/ReactivePgDataSourceHealthCheck.java
gsmet: It should except if they do some weird magic in the driver but that would be surprising.

gsmet (approved):

PR #7863, opened 2020-03-14
Use proper configuration for environment variables in JibConfig

geoand:


extensions/container-image/container-image-jib/deployment/src/main/java/io/quarkus/container/image/jib/deployment/JibProcessor.java
```diff
@@ -203,7 +203,7 @@ private JibContainerBuilder createContainerBuilderFromJar(JibConfig jibConfig,
             }
 
             return javaContainerBuilder.toContainerBuilder()
-                    .setEnvironment(jibConfig.environmentVariables.orElse(Collections.emptyMap()))
+                    .setEnvironment(jibConfig.environmentVariables)
```
gastaldi: `jibConfig.environmentVariables` will be null if it is not set.

geoand: It will not 😎. I tried it and it's an empty map. But let me try it again to be 100% sure

geoand: In general, Quarkus config objects never have null values.

Here is the screenshot from the debugger for a test I did:

Screenshot from 2020-03-14 14-28-05

gastaldi: Hm I could swear I saw it being null in https://github.com/quarkusio/quarkus/blob/master/extensions/flyway/runtime/src/main/java/io/quarkus/flyway/runtime/FlywayDataSourceRuntimeConfig.java#L113

But maybe because of this: https://github.com/quarkusio/quarkus/blob/master/extensions/flyway/runtime/src/main/java/io/quarkus/flyway/runtime/FlywayDataSourceRuntimeConfig.java#L20

Thanks for proving me wrong, I was annoyed by that ;)

gastaldi (approved):

extensions/container-image/container-image-jib/deployment/src/main/java/io/quarkus/container/image/jib/deployment/JibProcessor.java
geoand: Yeah, I think you found the cause of the `null` value 👍

Issue #6707, opened 2020-01-21
AmazonLambdaRecorder.chooseHandlerClass needs to support RequestStreamHandler pattern


stdunbar:
I'm unable to get the out of the box AWS Lambda test to work if my Lambda uses the RequestStreamHandler implementation pattern.  My code:

```
@Named("lambdaHandler")
public class LambdaHandler implements RequestStreamHandler {
    public void handleRequest(InputStream inputStream, OutputStream outputStream, Context context) {
        context.getLogger().log( "starting " + this.getClass().getName() );
    }
}
```

And my application.properties has `quarkus.lambda.handler=lambdaHandler`.  At times it is easier to read the InputStream and parse what you need out of it rather than creating a POJO.  As it is I get:

```
2020-01-21 14:30:08,111 ERROR [io.qua.application] (main) Failed to start application: java.lang.RuntimeException: Unable to find handler class with name lambdaHandler make sure there is a handler class in the deployment with the correct @Named annotation
    at io.quarkus.amazon.lambda.runtime.AmazonLambdaRecorder.chooseHandlerClass(AmazonLambdaRecorder.java:98)
    at io.quarkus.deployment.steps.AmazonLambdaProcessor$recordHandlerClass32.deploy_0(AmazonLambdaProcessor$recordHandlerClass32.zig:110)
    at io.quarkus.deployment.steps.AmazonLambdaProcessor$recordHandlerClass32.deploy(AmazonLambdaProcessor$recordHandlerClass32.zig:129)
    at io.quarkus.runner.ApplicationImpl.doStart(ApplicationImpl.zig:172)
    at io.quarkus.runtime.Application.start(Application.java:87)
```

Perhaps the `quarkus.lambda.handler` could work like the AWS Lambda environment and accept a full package name with method name, i.e. `com.example.lambda.handler.LambdaHander::myHandler`

gsmet:
@evanchooly could you have a look at that one?

jmkgreen:
I am seeing exactly the same. If it implements RequestHandler it seems to work.

According to https://quarkus.io/guides/amazon-lambda#choose-your-lambda only `RequestHandler` implementations are scanned for, which may be why.

I am building against `RequestHandler` because I want to use Jackson. The documentation for Quarkus mentioning Jackson does not make clear whether the integration is purely for REST API interfaces or for others such as Amazon Lambda too (which would render my work here redundant).

PhilAndrew:
[...]

PR #7866, opened 2020-03-14
RequestStreamHandler functionality added to Amazon Lambda

oztimpower:
**Summary:**

Amazon Lambda extension currently only supports the interface RequestHandler<?, ?>.  

Enhance the extension, to enable the use of the interface RequestStreamHandler, so that both AWS Lambda interfaces are supported.

#6707 requests this functionality.

The RequestStreamHandler can co-exist in a project with RequestHandler<?, ?>, and the active function toggled via the property quarkus.lambda.handler, as normal.  The requestHandler signature for the former has 3 params, the latter 2.

Validated as working for NativeImage, via SAM and AWS.

**Specifics:**

**Deployment:**

AmazonLambdaBuildItem adds a flag to indicate whether the scanned requestHandler is of type RequestStreamHandler, since it was the only way I could determine this (tried lots of options not to modify this BuildItem class).

`
    public AmazonLambdaBuildItem(String handlerClass, String name, boolean streamHandler) {
`

Various changes to AmazonLambdaProcessesor to recognise and process the RequestStreamHandler::requestHandler implementation.

**Runtime:**

AmazonLambdaRecorder has similar changes to accept the co-existance of the RequestStreamHandler::requestHandler implementation.

Only one lambda handler can be active, per the existing validation in chooseHandlerClass, thus the code is checking which one is present to determine how to branch. 

Added a compatible integration test for the AmazonClient.invoke in the test-framework for Amazon Lambda.  I used this to test both.

**Archetype:**

Added a StreamHandler to the archetype, to demonstrate you can have both in your project, and that you must use the @Named annotation along with the application.properties to select the active lambda function, per existing functionality.

In addition, I added test functionality to include the integration test for Amazon Lambda into the archetype, as that is an easy way for people to test their functions, and wasn't previously in the archetype.  For JVM, there is no need to use SAM when you have the integration test available (see issues with SAM for JVM #6701) which I continue to experience.

**Tested by**:

1. mvn archetype:generate -DarchetypeGroupId=io.quarkus -DarchetypeArtifactId=quarkus-amazon-lambda-archetype -DarchetypeVersion=999-SNAPSHOT
2. mvn package
3. sh manage.sh create
4. sh manage.sh invoke
5. mvn package -Dnative -Dquarkus.native.container-build=true'
6. sam local invoke --template sam.native.yaml --event payload.json
7. sh manage.sh native create invoke
(repeated toggling the application.property to swap handler type quarkus.lambda.handler)

/cc @patriot1burke 
/cc @evanchooly 
/cc @bnusunny

### DIFF
--- a/extensions/amazon-lambda/deployment/src/main/java/io/quarkus/amazon/lambda/deployment/AmazonLambdaProcessor.java
+++ b/extensions/amazon-lambda/deployment/src/main/java/io/quarkus/amazon/lambda/deployment/AmazonLambdaProcessor.java
@@ -19,6 +19,7 @@ import org.jboss.logging.Logger;
 import org.joda.time.DateTime;
 
 import com.amazonaws.services.lambda.runtime.RequestHandler;
+import com.amazonaws.services.lambda.runtime.RequestStreamHandler;
 
 import io.quarkus.amazon.lambda.runtime.AmazonLambdaRecorder;
 import io.quarkus.amazon.lambda.runtime.FunctionError;
@@ -49,6 +50,7 @@ public final class AmazonLambdaProcessor {
     public static final String AWS_LAMBDA_EVENTS_ARCHIVE_MARKERS = "com/amazonaws/services/lambda/runtime/events";
 
     private static final DotName REQUEST_HANDLER = DotName.createSimple(RequestHandler.class.getName());
+    private static final DotName REQUEST_STREAM_HANDLER = DotName.createSimple(RequestStreamHandler.class.getName());
 
     private static final DotName NAMED = DotName.createSimple(Named.class.getName());
     private static final Logger log = Logger.getLogger(AmazonLambdaProcessor.class);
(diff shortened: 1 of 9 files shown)


extensions/amazon-lambda/deployment/src/main/java/io/quarkus/amazon/lambda/deployment/AmazonLambdaProcessor.java
```diff
@@ -90,14 +96,18 @@ AdditionalApplicationArchiveMarkerBuildItem marker() {
             }
 
             final String lambda = name.toString();
-            ret.add(new AmazonLambdaBuildItem(lambda, cdiName));
+            //            ret.add(new AmazonLambdaBuildItem(lambda, cdiName));
```
patriot1burke: Remove commented code

patriot1burke (commented):
Really nice PR!.  Awesome you added the test harness.  The switches you added are useful too.

extensions/amazon-lambda/deployment/src/main/java/io/quarkus/amazon/lambda/deployment/AmazonLambdaProcessor.java
oztimpower: Done

oztimpower:
Good news is that with this new RequestStreamHandler functionality I've been able to get Alexa Lambda functions to work seamlessly with JVM, and also with native (lengthy reflection-config.json however to cater for inner builder classes).

patriot1burke (approved):

bnusunny:
@oztimpower This is great! Top request from lambda users. Thank you Timothy.

PR #7874, opened 2020-03-16
Have the Agroal datasources initialized eagerly

### LINKED ISSUE #4507: Agroal datasource should be initialized eagerly
**Describe the bug**
Currently the datasources are flagged as `@ApplicaitonScoped`, which means they are wrapped by a proxy making them initialized on first use.

This is annoying as
 - we don't respect the configuration of pre-filling the pool on boot
 - we can't validate the connection properties, deferring any failure (however also affected by #4503)
 - it's an overhead

Postremus:
Originally, this was done by @Sanne in https://github.com/quarkusio/quarkus/pull/4504/commits/5209e40b9b40d1bc8a740e9f2d2247c4cc5b7396
But I have no Idea how to cherry pick from a deleted branch..
Created a new commit with the same changes because of this.

Main difference is, that this time around all ITs are passing.

closes #4507

### DIFF
--- a/extensions/agroal/deployment/src/main/java/io/quarkus/agroal/deployment/AgroalProcessor.java
+++ b/extensions/agroal/deployment/src/main/java/io/quarkus/agroal/deployment/AgroalProcessor.java
@@ -11,9 +11,9 @@ import java.util.Map.Entry;
 import java.util.Optional;
 import java.util.Set;
 
-import javax.enterprise.context.ApplicationScoped;
 import javax.enterprise.inject.Default;
 import javax.enterprise.inject.Produces;
+import javax.inject.Singleton;
 import javax.sql.XADataSource;
 
 import org.eclipse.microprofile.metrics.Metadata;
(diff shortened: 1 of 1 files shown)


Sanne (approved):
that's great to know! Thanks @Postremus

dufoli:
@Sanne @Postremus maybe the same patch must be applied to mongo codebase which is close to agroal now.

Sanne:
what do you mean @dufoli ? I've not followed what is happening with the MongoDB extension, how it that related to Agroal?

cc/ @loicmathieu

dufoli:
Multi tenancy support of Mongo client have introduce a mongo processor very close to agroal processor.

Postremus:
@dufoli ah, I see what you mean. I will test if this change is also possible for the mongo processor.

geoand:
I believe this should be marked as a breaking change, WDYT @Sanne @gsmet ?

Sanne:
Technically I see it as a bugfix, and there was a tracking issue open that
people could watch, but +1 to highlight the change somehow

geoand:
I agree it's the right thing to do, but from a user's perpective it is also breaking (some users could possibly be relying on the fact that the application could start even if the connection to the database wasn't available at startup time).

PR #7890, opened 2020-03-16
Changing remote debug connection string for JDK 9+

### LINKED ISSUE #7889: Remote debug into dev mode doesn't work on JDK 9+
**Describe the bug**
If I run `quarkus:dev` remotely (i.e. in a Docker container), I can't remote debug if I'm using JDK 9 or newer (the remote connection string has been changed and requires adapting).

Trying to debug remotely (with IDE / jdb through `localhost:5005`) fails for me with:

```
Error running Debugger: Unable to open debugger port (localhost:5005): 
java.io.IOException "handshake failed - connection prematurally closed"
```

Also see [this StackOverflow answer](https://stackoverflow.com/a/60090750).

 - Output of `java -version`: `openjdk version "13.0.1" 2019-10-15 \n OpenJDK Runtime Environment AdoptOpenJDK (build 13.0.1+9)`
 - Quarkus version or git rev: `1.3.0.Final`

**Additional context**
The cause of this issue is in [DevMojo.java](https://github.com/quarkusio/quarkus/blob/master/devtools/maven/src/main/java/io/quarkus/maven/DevMojo.java#L497) in the debug connection; instead of `transport=dt_socket,address=5005,server=y,suspend=n`, with JDK 9+ you'd need `transport=dt_socket,address=*:5005,server=y,suspend=n` for the same result (allowing connections from every host, not just localhost).

sdaschner:
Fixes: #7889.

Should work for both pre and post JDK 9 (quickly tested with 8 and 13) and has the same effect like before (connections from every host is allowed).

### DIFF
--- a/devtools/gradle/src/main/java/io/quarkus/gradle/tasks/QuarkusDev.java
+++ b/devtools/gradle/src/main/java/io/quarkus/gradle/tasks/QuarkusDev.java
@@ -206,7 +206,7 @@ public class QuarkusDev extends QuarkusTask {
                 args.add("-Xrunjdwp:transport=dt_socket,address=localhost:5005,server=n,suspend=" + debugSuspend);
             } else if (debug.toLowerCase().equals("true") || debug.isEmpty()) {
                 args.add("-Xdebug");
-                args.add("-Xrunjdwp:transport=dt_socket,address=localhost:5005,server=y,suspend=" + debugSuspend);
+                args.add("-Xrunjdwp:transport=dt_socket,address=0.0.0.0:5005,server=y,suspend=" + debugSuspend);
             } else if (!debug.toLowerCase().equals("false")) {
                 try {
                     int port = Integer.parseInt(debug);
--- a/devtools/gradle/src/main/java/io/quarkus/gradle/tasks/QuarkusDev.java
+++ b/devtools/gradle/src/main/java/io/quarkus/gradle/tasks/QuarkusDev.java
@@ -199,7 +199,7 @@ public class QuarkusDev extends QuarkusTask {
                     System.err.println("Port 5005 in use, not starting in debug mode");
                 } catch (IOException e) {
                     args.add("-Xdebug");
-                    args.add("-Xrunjdwp:transport=dt_socket,address=5005,server=y,suspend=" + debugSuspend);
+                    args.add("-Xrunjdwp:transport=dt_socket,address=0.0.0.0:5005,server=y,suspend=" + debugSuspend);
                 }
             } else if (debug.toLowerCase().equals("client")) {
                 args.add("-Xdebug");
@@ -214,7 +214,7 @@ public class QuarkusDev extends QuarkusTask {
                         throw new GradleException("The specified debug port must be greater than 0");
                     }
                     args.add("-Xdebug");
-                    args.add("-Xrunjdwp:transport=dt_socket,address=" + port + ",server=y,suspend=" + debugSuspend);
+                    args.add("-Xrunjdwp:transport=dt_socket,address=0.0.0.0:" + port + ",server=y,suspend=" + debugSuspend);
                 } catch (NumberFormatException e) {
                     throw new GradleException(
                             "Invalid value for debug parameter: " + debug + " must be true|false|client|{port}");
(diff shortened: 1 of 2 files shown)


gastaldi (approved):

geoand:
Thanks for the PR!

Shouldn't we be using `localhost` instead of `0.0.0.0` (like the other instances in the code are using)?

gsmet:
Yeah I think I would prefer `localhost` too for dev mode.

gastaldi:
@geoand good catch, I was going to ask to do the same in the Gradle plugin, then I found this: https://github.com/quarkusio/quarkus/blob/b8e68dd43dbb908fe3dddf8a04eeedc0699d2c27/devtools/gradle/src/main/java/io/quarkus/gradle/tasks/QuarkusDev.java#L195-L222

geoand:
Thinking about it again, `localhost` is good for dev mode, but will cause problems if people try to debug a dev mode container... 
I am not sure that it's such a valid use case though...

gastaldi:
According to https://github.com/quarkusio/quarkus/issues/7889, @sdaschner says that `localhost` fails when running Quarkus inside a Docker container

geoand:
Anyway we choose, we should be consistent across all usages :)

geoand:
> According to #7889, @sdaschner says that `localhost` fails when running Quarkus inside a Docker container

Yeah, I expected that...

So we have the following options:

always use `0.0.0.0` or provide the ability to set change `localhost` to `0.0.0.0`.

sdaschner:
> According to #7889, @sdaschner says that `localhost` fails when running Quarkus inside a Docker container

Yes, exactly. `localhost` is the same as now (port only). But for actual remote access (e.g. container) we need `0.0.0.0:{port}` or `*:{port}`. The latter fails on pre-JDK 9, that's why I tried to use `0.0.0.0`.

gastaldi:
I'm fine with `0.0.0.0` in the Maven and Gradle plugins and a property to change it if necessary (like we do with `port`)

geoand:
I'm fine with `0.0.0.0` as well.

sdaschner:
... is the failing CI build caused by the change? Don't know how to asses that

gastaldi:
@sdaschner can you update the Gradle plugin too?

sdaschner:
Updated that as well. To quote a famous person: 

> Be careful with the code, I have only proved it correct, not tested it :)

devtools/gradle/src/main/java/io/quarkus/gradle/tasks/QuarkusDev.java
```diff
@@ -199,7 +199,7 @@ public void startDev() {
                     System.err.println("Port 5005 in use, not starting in debug mode");
                 } catch (IOException e) {
                     args.add("-Xdebug");
-                    args.add("-Xrunjdwp:transport=dt_socket,address=5005,server=y,suspend=" + debugSuspend);
+                    args.add("-Xrunjdwp:transport=dt_socket,address=0.0.0.0:5005,server=y,suspend=" + debugSuspend);
                 }
             } else if (debug.toLowerCase().equals("client")) {
                 args.add("-Xdebug");
```
gastaldi: The line below uses `localhost`

gastaldi: Here localhost is used too: https://github.com/quarkusio/quarkus/pull/7890/files#diff-72fac779acece3ea68a6666ee0060ad5R209

sdaschner: Yes, the code before had `localhost` there too, therefore I didn't want to change the behavior. But agree, we could think about if it's not more consistent to allow remote connections everywhere

gastaldi: IMHO it should all be `0.0.0.0` unless proven wrong :wink:

sdaschner:
Updated the PR to use `0.0.0.0` for all (except `client` connect). This changes the behavior (from a security perspective) for `-Ddebug=true` but is now consistent with the other server options.

gastaldi (approved):

PR #7894, opened 2020-03-16
Exclude org.checkerframework:checker-qual from runtime dependencies

michalszynkiewicz:


bom/runtime/pom.xml
```diff
@@ -976,6 +976,12 @@
                 <groupId>com.github.ben-manes.caffeine</groupId>
                 <artifactId>caffeine</artifactId>
                 <version>${caffeine.version}</version>
+                <exclusions>
+                    <exclusion>
+                        <groupId>org.checkerframework</groupId>
+                        <artifactId>checker-qual</artifactId>
+                    </exclusion>
+                </exclusions>
```
gsmet: If we know it's an issue, can we add it to the banned dependencies of the enforcer in the build-parent?

gsmet (changes requested):

bom/runtime/pom.xml
michalszynkiewicz: It also goes in through maven-embedder but then it is only used in deployment. 
I'd need to exclude it there too to ban it.
I'll give it a try.

gsmet:
Ok if it’s not obvious, we can live with it. We have better things to do :)

gsmet (approved):
Feel free to merge when you think it’s ready.

michalszynkiewicz:
@gsmet thanks for mentioning banning. It turned out there was a few places I missed when simply looking at the dependency tree.

PR #7896, opened 2020-03-16
Resolve test jars from the local project correctly

### LINKED ISSUE #7887: classloading issue when @QuarkusTest inherits from class from another artifact (Maven module)
**Describe the bug**
I have a database application and I want to support multiple databases. For that, I need to build the application separately for each database. For various reasons, I decided to do this as a multi-module Maven project, structured like this:

- `app`: the application itself, doesn't have `quarkus-maven-plugin`, but has `jandex-maven-plugin` to produce Jandex
- `postgresql`: depends on `app`, adds `quarkus-jdbc-postgresql` and has `quarkus-maven-plugin` to produce the PostgreSQL-specific build
- `mysql`, `mariadb`, `mssql`: same as `postgresql`, but for other databases

With all these different databases, I also want to run integration tests. For that, I created an `abstract` class in the `app` module that has all the tests (`AbstractDatabaseTest`). I also produce a `tests-jar` from the `app` module, so that I can keep `AbstractDatabaseTest` in `src/test`. I use RestAssured for the tests, relying on Quarkus setting up the correct base URL and port in RestAssured for me.

Then, the idea is that in each submodule, I create an empty class that inherits from `AbstractDatabaseTest` and just adds the `@QuarkusTest` annotation.

I also have an empty subclass of `AbstractDatabaseTest` in `app` that runs with H2. That test passes just fine. All the tests in other modules fail with `Connection refused`.

I was able to trace this down to a classloading problem: the test class itself is loaded by `Quarkus Base Runtime ClassLoader`, but its parent class, `AbstractDatabaseTest`, is loaded by the system classloader (some classloader delegation seems to be set up, so this arrangement doesn't immediately crash). The parent class uses RestAssured, so that is loaded by the system classloader as well. And now I have RestAssured loaded by 2 classloaders -- the Quarkus one, which has the correct URL and port set, and the system classloader one, which doesn't have the correct settings, but is the one that is actually used.

**Expected behavior**
Superclasses of test class are loaded by the Quarkus classloader.

**Actual behavior**
Superclasses of test class are loaded by the system classloader.

**To Reproduce**
I have created a minimal reproducer, which doesn't have all the database stuff etc. Just the bare minimum to show what's going on.

1. `git clone https://github.com/Ladicek/quarkus-multimodule-test-inheritance-issue.git`
2. `mvn clean test`

 - Output of `uname -a` or `ver`: `Linux argondie 4.15.0-88-generic #88-Ubuntu SMP Tue Feb 11 20:11:34 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux`
 - Output of `java -version`:
```
openjdk version "1.8.0_242"
OpenJDK Runtime Environment (build 1.8.0_242-8u242-b08-0ubuntu3~18.04-b08)
OpenJDK 64-Bit Server VM (build 25.242-b08, mixed mode)
```
 - GraalVM version (if different from Java): N/A
 - Quarkus version or git rev: `1.3.0.Final`
```
Apache Maven 3.6.3 (cecedd343002696d0abb50b32b541b8a6ba2883f)
Maven home: /home/lthon/software/apache-maven
Java version: 1.8.0_242, vendor: Private Build, runtime: /usr/lib/jvm/java-8-openjdk-amd64/jre
Default locale: cs_CZ, platform encoding: UTF-8
OS name: "linux", version: "4.15.0-88-generic", arch: "amd64", family: "unix"
```

**Additional context**

If you look at the reproducer, you will notice that the `specific` module contains a dummy class (called `Dummy`). That class is completely useless and I really don't want it there. But if I remove it, Maven doesn't create `target/classes` at all (which is correct) and then Quarkus fails with `java.io.FileNotFoundException: /home/lthon/projects/quarkus/reproducers/quarkus-multimodule-test-inheritance-issue/specific/target/classes (No such file or directory)`. I'd consider that a bug too.

stuartwdouglas:
Fixes #7887

### DIFF
--- a/independent-projects/bootstrap/core/src/main/java/io/quarkus/bootstrap/resolver/maven/workspace/LocalProject.java
+++ b/independent-projects/bootstrap/core/src/main/java/io/quarkus/bootstrap/resolver/maven/workspace/LocalProject.java
@@ -202,6 +202,10 @@ public class LocalProject {
         return getOutputDir().resolve("classes");
     }
 
+    public Path getTestClassesDir() {
+        return getOutputDir().resolve("test-classes");
+    }
+
     public Path getSourcesSourcesDir() {
         if (getRawModel().getBuild() != null && getRawModel().getBuild().getSourceDirectory() != null) {
             String originalValue = getRawModel().getBuild().getSourceDirectory();
--- a/independent-projects/bootstrap/core/src/main/java/io/quarkus/bootstrap/resolver/maven/workspace/LocalWorkspace.java
+++ b/independent-projects/bootstrap/core/src/main/java/io/quarkus/bootstrap/resolver/maven/workspace/LocalWorkspace.java
@@ -7,6 +7,7 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import org.apache.maven.model.Model;
 import org.apache.maven.model.resolution.UnresolvableModelException;
 import org.apache.maven.model.resolution.WorkspaceModelResolver;
@@ -87,6 +88,16 @@ public class LocalWorkspace implements WorkspaceModelResolver, WorkspaceReader {
                 || !lp.getVersion().equals(artifact.getVersion())
                         && !(LocalProject.REVISION_EXPR.equals(artifact.getVersion())
                                 && lp.getVersion().equals(revision))) {
+            return null;
+        }
+        if (!Objects.equals(artifact.getClassifier(), lp.getAppArtifact().getClassifier())) {
+            if ("tests".equals(artifact.getClassifier())) {
+                //special classifier used for test jars
+                final File file = lp.getTestClassesDir().toFile();
+                if (file.exists()) {
+                    return file;
+                }
+            }
             return null;
         }
         final String type = artifact.getExtension();


geoand:
@Ladicek can you please test this?

Ladicek:
This works in the simple reproducer, if I use `<type>test-jar</type>`. I'm not quite sure why that's needed, but apparently it's the correct thing to do.

I'll check with my "full" project now.

geoand:
CI failure has nothing to do with the PR (and should be fixed by: https://github.com/quarkusio/quarkus/pull/7901)

stuartwdouglas:
test-jar is not required after this pr

Ladicek:
With this, the tests in my "full", more complex database application also run correctly!

aloubyansky (approved):

aloubyansky:
Looks like it failed the second time on
```
[ERROR] Errors: 
[ERROR]   KubernetesWithQuarkusAppNameTest � Runtime java.lang.RuntimeException: io.quar...
[INFO] 
[ERROR] Tests run: 19, Failures: 0, Errors: 1, Skipped: 0
```
Haven't yet looked into that. Re-running again.

geoand:
@aloubyansky that has been fixed by #7901

PR #7906, opened 2020-03-17
Make it possible to inject Instance<> into observers/disposers

### LINKED ISSUE #7888: ArC - it's not possible to inject Instance<> into an observer method parameter
The code generated by `io.quarkus.arc.processor.BuiltinBean.INSTANCE` only considers injection into a bean.

mkouba:
- also destroy Instance<> correctly when the invocation completes
- resolves #7888


mkouba:
@gsmet this would be probably worth backporting into 1.3.1.

manovotn (approved):
Looks good.
Seeing this I also double checked if injecting Event/BM into those methods works and that seems to be good.

PR #7908, opened 2020-03-17
Add doc about defaults Spring Boot naming strategies

aureamunoz:
…out how to configure the Quarkus properties to use them
Could you, please, take a look @geoand ?

### DIFF
--- a/bom/runtime/pom.xml
+++ b/bom/runtime/pom.xml
@@ -176,7 +176,7 @@
         <quarkus-spring-api.version>5.2.Final</quarkus-spring-api.version>
         <quarkus-spring-data-api.version>2.1.Final</quarkus-spring-data-api.version>
         <quarkus-spring-security-api.version>5.2.Final</quarkus-spring-security-api.version>
-        <quarkus-spring-boot-api.version>2.1.Final</quarkus-spring-boot-api.version>
+        <quarkus-spring-boot-api.version>2.1.SP1</quarkus-spring-boot-api.version>
         <mvel2.version>2.4.4.Final</mvel2.version>
         <htmlunit.version>2.36.0</htmlunit.version>
         <mockito.version>3.3.0</mockito.version>
@@ -2824,6 +2824,11 @@
             </dependency>
             <dependency>
                 <groupId>io.quarkus</groupId>
+                <artifactId>quarkus-spring-boot-orm-api</artifactId>
+                <version>${quarkus-spring-boot-api.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>io.quarkus</groupId>
                 <artifactId>quarkus-spring-data-commons-api</artifactId>
                 <version>${quarkus-spring-data-api.version}</version>
             </dependency>
@@ -2844,7 +2849,7 @@
             </dependency>
             <dependency>
                 <groupId>io.quarkus</groupId>
-                <artifactId>quarkus-spring-boot-api</artifactId>
+                <artifactId>quarkus-spring-boot-properties-api</artifactId>
                 <version>${quarkus-spring-boot-api.version}</version>
             </dependency>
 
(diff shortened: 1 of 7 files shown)


docs/src/main/asciidoc/spring-data-jpa.adoc
```diff
@@ -567,9 +567,18 @@ public interface MovieRepository extends CrudRepository<Movie, Long> {
     Slice<Movie> orderByTitleLength(Pageable pageable);
 }
 ----
-
 All methods that are annotated with `@Modifying` will automatically be annotated with `@Transactional`.
 
+==== Naming Strategies
+
+Hibernate maps field names using a physical strategy and an implicit strategy. If you wish Quarkus to use the defaults Spring Boot Naming Strategies, you need to set up the following properties:
```
geoand:
```suggestion
Hibernate maps field names using a physical strategy and an implicit strategy. If you wish to use Spring Boot's default naming strategies, the following properties need to be se:
```

geoand:
This will require adding the new jar, no? If so, please mark the PR as `WIP`

geoand:
LGTM

Once we have the jar it should be good to go

aureamunoz:
In fact, no, I need to add the new jar dependency to the Spring Data JPA extension and modify the name of another one....Once the build pass locally, I will update this PR

aureamunoz:
I've just pushed the update of quarkus-spring-boot-api new released version, let's see what says the CI

bom/runtime/pom.xml
```diff
@@ -176,7 +176,7 @@
         <quarkus-spring-api.version>5.2.Final</quarkus-spring-api.version>
         <quarkus-spring-data-api.version>2.1.Final</quarkus-spring-data-api.version>
         <quarkus-spring-security-api.version>5.2.Final</quarkus-spring-security-api.version>
-        <quarkus-spring-boot-api.version>2.1.Final</quarkus-spring-boot-api.version>
+        <quarkus-spring-boot-api.version>2.2.Alpha1</quarkus-spring-boot-api.version>
```
geoand: Hm... Did we update the Spring Boot version? If so, why? If not, why did we bump the Spring Boot version here?

aureamunoz: Yes, I updated the spring boot version, I thought it would be ok but as talked before let's keep the 2.1.13

gsmet: We need a Final version here, not an Alpha. This will be backported to 1.3.1.Final.

If 2.2.Final is not an option, use 2.2.SP1 or something like that.

/cc @geoand for verification.

docs/src/main/asciidoc/spring-data-jpa.adoc
```diff
@@ -567,9 +567,18 @@ public interface MovieRepository extends CrudRepository<Movie, Long> {
     Slice<Movie> orderByTitleLength(Pageable pageable);
 }
 ----
-
 All methods that are annotated with `@Modifying` will automatically be annotated with `@Transactional`.
 
+==== Naming Strategies
+
+Hibernate maps field names using a physical strategy and an implicit strategy. If you wish to use Spring Boot's default naming strategies, the following properties need to be set:
```
gsmet:
```suggestion
Hibernate ORM maps property names using a physical naming strategy and an implicit naming strategy. If you wish to use Spring Boot's default naming strategies, the following properties need to be set:
```

gsmet (dismissed):
I added some comments.

Please wait for @geoand's blessing to be sure I'm not asking for dumb changes :).

bom/runtime/pom.xml
geoand: We agreed that the bump to 2.2 was not correct and we need to go back to 2.1.xxx

In any case @aureamunoz do keep in mind that in Quarkus master we can no longer have `Alpha` versions at any time.

aureamunoz: Ok, yes. Thanks. I will do the necessary in the API and open a PR, then update this one

geoand:
Let's update test for the properties please

aureamunoz:
Done, let's see what the CI says...

geoand (approved):

gsmet:
@geoand so should we merge that one?

geoand:
Done!

PR #7920, opened 2020-03-17
Ensure that enums can be used in @ParameterizedTest in @QuarkusTest

### LINKED ISSUE #7916: Error running @ParameterizedTest in Quarkus 1.3.0.Final
**Describe the bug**
Simple method in a `@QuarkusTest` which is also a `@ParameterizedTest` doesn't work.

To reproduce:

```
@QuarkusTest
public class ExampleResourceTest {

    public enum MyEnum {
        VALUE1,
        VALUE2
    }

    @ParameterizedTest(name = "Enum Value for test {0}")
    @EnumSource(value = MyEnum.class)
    public void testHelloEndpoint(MyEnum input) {
    }

}
```

if you try to run it you got:

`java.lang.RuntimeException: Could not find method public void net.lucamasini.testcase.ExampleResourceTest.testHelloEndpoint(net.lucamasini.testcase.ExampleResourceTest$MyEnum) on test class`

one for every enum value.

**Expected behavior**
The runtime container should find the enumerated method

**Actual behavior**
Cannot find the public method

**To Reproduce**
1. Create a project from Quarkus.io
2. Add the test case at the top of this issue
3. Run it

**Configuration**
None

geoand:
Fixes: #7916

### DIFF
--- a/integration-tests/main/src/test/java/io/quarkus/it/main/ParameterizedEnumTestCase.java
+++ b/integration-tests/main/src/test/java/io/quarkus/it/main/ParameterizedEnumTestCase.java
@@ -1,0 +1,21 @@
+package io.quarkus.it.main;
+
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.EnumSource;
+
+import io.quarkus.test.junit.QuarkusTest;
+
+@QuarkusTest
+public class ParameterizedEnumTestCase {
+
+    public enum MyEnum {
+        VALUE1,
+        VALUE2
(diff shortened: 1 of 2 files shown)


geoand:
I really don't like this "solution" - it's more of a hack than anything else...

@stuartwdouglas do you have anything more clever to offer?

stuartwdouglas (approved):

stuartwdouglas:
There is not much else that can be done here, unfortunately the new class loader impl does result in a bit of code like this, but at least it is hidden from the user.

Issue #7910, opened 2020-03-17
Infinispan can't be manually started with Quarkus 1.3.0


burmanm:
**Describe the bug**
When launching the application that uses Infinispan directly (not through infinispan-embedded), the startup fails to loading metrics:

```
2020-03-17 13:35:10,563 ERROR [io.qua.application] (main) Failed to start application: java.lang.IllegalStateException: Display name differs from previous usage
	at io.smallrye.metrics.MetricsRegistryImpl.verifyMetadataEquality(MetricsRegistryImpl.java:203)
	at io.smallrye.metrics.MetricsRegistryImpl.register(MetricsRegistryImpl.java:151)
	at io.quarkus.smallrye.metrics.runtime.SmallRyeMetricsRecorder.garbageCollectionMetrics(SmallRyeMetricsRecorder.java:250)
	at io.quarkus.smallrye.metrics.runtime.SmallRyeMetricsRecorder.registerBaseMetrics(SmallRyeMetricsRecorder.java:110)
	at io.quarkus.deployment.steps.SmallRyeMetricsProcessor$registerBaseAndVendorMetrics18.deploy_0(SmallRyeMetricsProcessor$registerBaseAndVendorMetrics18.zig:65)
	at io.quarkus.deployment.steps.SmallRyeMetricsProcessor$registerBaseAndVendorMetrics18.deploy(SmallRyeMetricsProcessor$registerBaseAndVendorMetrics18.zig:36)
	at io.quarkus.runner.ApplicationImpl.doStart(ApplicationImpl.zig:138)
	at io.quarkus.runtime.Application.start(Application.java:90)
	at io.quarkus.runtime.Application.run(Application.java:228)
	at io.quarkus.runner.GeneratedMain.main(GeneratedMain.zig:41)
```

This is visible here because of #7907 that causes the displayName to differ from the one that Infinispan loaded, since Infinispan's ```new DefaultCacheManager(is)``` loads its own metrics before Quarkus:

```
DefaultMetadata{name='gc.total', type=counter, unit='none', reusable=true, description='Displays the total number of collections that have occurred. This attribute lists -1 if the collection count is undefined for this collector.', displayName='Garbage Collection Count'}
 
register:116, MetricsRegistryImpl (io.smallrye.metrics)
register:74, JmxRegistrar (io.smallrye.metrics.setup)
register:58, JmxRegistrar (io.smallrye.metrics.setup)
[...]

PR #7921, opened 2020-03-17
Update Infinispan to 10.1.5.Final and update test classes

wburns:
https://github.com/quarkusio/quarkus/issues/7910

### DIFF
--- a/extensions/infinispan-embedded/runtime/pom.xml
+++ b/extensions/infinispan-embedded/runtime/pom.xml
@@ -30,6 +30,11 @@
             <groupId>io.smallrye</groupId>
             <artifactId>smallrye-metrics</artifactId>
         </dependency>
+        <!-- Hack to get it to have the API Infinispan uses available -->
+        <dependency>
+            <groupId>org.eclipse.microprofile.metrics</groupId>
+            <artifactId>microprofile-metrics-api</artifactId>
+        </dependency>
         <dependency>
             <groupId>org.infinispan</groupId>
             <artifactId>infinispan-core</artifactId>
--- a/bom/runtime/pom.xml
+++ b/bom/runtime/pom.xml
@@ -119,8 +119,8 @@
         <testng.version>6.14.2</testng.version>
         <assertj.version>3.15.0</assertj.version>
         <json-smart.version>2.3</json-smart.version>
-        <infinispan.version>10.1.2.Final</infinispan.version>
-        <infinispan.protostream.version>4.3.1.Final</infinispan.protostream.version>
+        <infinispan.version>10.1.5.Final</infinispan.version>
+        <infinispan.protostream.version>4.3.2.Final</infinispan.protostream.version>
         <caffeine.version>2.8.0</caffeine.version>
         <netty.version>4.1.45.Final</netty.version>
         <reactive-streams.version>1.0.3</reactive-streams.version>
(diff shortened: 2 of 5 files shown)


gsmet:
@jmartisk can you check it fixes the metrics issue?

wburns:
It is failing native build saying it can't find the metrics classes. I am not quite sure why yet as we have quarkus smallrye metrics as a dependency.

wburns:
So I fixed the xml issues by tweaking it slightly (we added extra sanity checks that the xml was violating).

However I had to add in a hack to have the appropriate eclipse microprofile apis available, @jmartisk maybe you could help a bit? Infinispan is on 2.1.5 of smallrye metrics where as Quarkus is on 2.4.1. To be honest I am not sure why this wasn't complaining before though either.

wburns:
Disregard my previous comment, it should be good now.

gsmet (approved):

jmartisk:
Sorry I was already finished for yesterday
> However I had to add in a hack to have the appropriate eclipse microprofile apis available, @jmartisk maybe you could help a bit? Infinispan is on 2.1.5 of smallrye metrics where as Quarkus is on 2.4.1. To be honest I am not sure why this wasn't complaining before though either.

That sounds like you may want to update soon, we should try to make the versions aligned as much as possible. I see Infinispan master is at 2.3.0, which is a bit better. It shouldn't be a critical problem for now because these versions are backward compatible, but for example in June we will have MicroProfile 4.0 with breaking changes. If you need assistance with upgrading metrics in Infinispan, just drop me a message, but I think it shouldn't be a problem

extensions/infinispan-embedded/runtime/pom.xml
```diff
@@ -30,6 +30,11 @@
             <groupId>io.smallrye</groupId>
             <artifactId>smallrye-metrics</artifactId>
         </dependency>
+        <!-- Hack to get it to have the API Infinispan uses available -->
```
anistor: Could you elaborate why do you call this a hack @wburns ?

jmartisk: I'd rather ask why aren't we importing the `quarkus-smallrye-metrics` module here instead?

wburns: > Could you elaborate why do you call this a hack @wburns ?

It isn't anymore, I forgot to remove the comment is all.

> I'd rather ask why aren't we importing the `quarkus-smallrye-metrics` module here instead?

But I can try this instead and see if it still works. Either way I can make a simple PR here in just a bit.

gsmet: For me, it depends if you need the API or an implementation.

If you just need the API, I wouldn't make the implementation mandatory.

@jmartisk WDYT?

jmartisk: AFAIK the SmallRye Metrics implementation is mandatory for Infinispan to be able to run (ideally it should be able to run without metrics dependency at all if they are disabled, but that's not the case right now)

wburns: I created https://github.com/quarkusio/quarkus/pull/7935 to remove the comment. Let me know if I should change the dependency though.

jmartisk: Other extensions that make use of Metrics declare the `quarkus-smallrye-metrics` dependency in their runtime module, so I'd vote for yes

gsmet: Ah ah, I missed the `smallrye-metrics` dependency just above. So yes, we need the extension to be used.

anistor: > right now

Now (10.1.5.final) it should be able to work without Smallrye metrics being present in classpath at runtime. It's truly optional.

PR #7926, opened 2020-03-18
Have the Mongo Clients initialized eagerly

### LINKED ISSUE #7895: Mongo clients should be initialized eagerly
**Describe the bug**
Currently the mongo clients are flagged as @ApplicaitonScoped, which means they are wrapped by a proxy making them initialized on first use.

This is annoying as

* we don't respect the configuration of pre-filling the pool on boot
* we can't validate the connection properties, deferring any failure
* it's an overhead

relates to #4507

Postremus:
We propably also should add a heads up in the migration guide.

/cc @dufoli 
closes #7895

### DIFF
--- a/extensions/mongodb-client/deployment/src/main/java/io/quarkus/mongodb/deployment/MongoClientProcessor.java
+++ b/extensions/mongodb-client/deployment/src/main/java/io/quarkus/mongodb/deployment/MongoClientProcessor.java
@@ -11,9 +11,9 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
 
-import javax.enterprise.context.ApplicationScoped;
 import javax.enterprise.inject.Default;
 import javax.enterprise.inject.Produces;
+import javax.inject.Singleton;
 
 import org.bson.codecs.configuration.CodecProvider;
 import org.bson.codecs.pojo.annotations.BsonDiscriminator;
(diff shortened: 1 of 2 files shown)


loicmathieu (approved):
LGTM but a review by @cescoffier is needed on this one

loicmathieu (approved):

mkouba (approved):
Looks good but I'm no Mongo expert ;-).

cescoffier:
@mkouba yes, but this change is mostly about ApplicationScoped to Singleton. The mongo part is not impacted.

cescoffier (approved):

cescoffier:
Before merging,  I would prefer having a go from the metrics guy.

mkouba:
> The mongo part is not impacted.

It's ok to change the scope and expect `@Singleton` beans to be initialized eagerly (when injected) but it could impact some other parts of the mongo extension (bootstrap logic, etc.) about which I have no idea...

jmartisk:
IIUC this is semi-eager, because the client will still only initialize once a bean that injects it is being initialized, so it does not necessarily mean initializing right after start, right?

mkouba:
> IIUC this is semi-eager, because the client will still only initialize once a bean that injects it is being initialized, so it does not necessarily mean initializing right after start, right?

Yes, as I said in my previous comment - eagerly, when injected. One could use `@io.quarkus.runtime.Startup` or add `void onStart(@Observes StartupEvent event)` to initialize the bean at startup.

jmartisk (approved):
LGTM

Issue #7771, opened 2020-03-11
Qute 1.3.0.CR2 - property/method [raw] not found on class [java.lang.String]


Maldivia:
**Describe the bug**
Updating from 1.2.1 to 1.3.0.CR2 now gives me the following BuildException, that previously worked in 1.2.1:

```
[1] Incorrect expression: playable.getPlayTitleEmbed().raw
        - property/method [raw] not found on class [java.lang.String] nor handled by an extension method
        - found in template [play.html] on line 6
```

Snippet from play.html contains the following:
```
{@com.foo.model.Playable playable}
{playable.getPlayTitleEmbed().raw}
```

where Playable is an interface:
```
@TemplateData
public interface Playable {
  public String getPlayTitleEmbed();
```

If I change the template code to `{playable.getPlayTitleEmbed}` and change the method to be `public RawString getPlayTitleEmbed();`, I instead get the following warning upon startup:
```
WARN  [io.qua.dep.ste.ReflectiveHierarchyStep] (build-36) Unable to properly register the hierarchy of the following classes for reflection as they are not in the Jandex index:
        - io.quarkus.qute.RawString
Consider adding them to the index either by creating a Jandex index for your dependency via the Maven plugin, an empty META-INF/beans.xml or quarkus.index-dependency properties.");.
```

**Environment (please complete the following information):**
Quarkus 1.3.0.CR2
Java 11.0.6
Maven 3.6.3

mkouba:
@Maldivia good catch! That's a bug. We improved the validation and forgot to add exception for the `raw` value resolver. I'm not so sure about the warning. Is `Playable` also used by other extensions (i.e. jackson, resteasy, etc.)?

Maldivia:
I have two classes that implement the Playable interface, which are deserialized from two different RestClient services, but the template works with both implementations.

mkouba:
> which are deserialized from two different RestClient services

That would be it. I'll add the `RawString` to the index.

Maldivia:
@mkouba 
The implemented solution explicitly checks if the type is String, but raw/safe are extensions of Object, and applies to the toString() value of it.

Given the following usage:
```
return template.data("obj", List.of("<i>Hello</i>", "<b>World</b>"));
```

```
{obj}
{obj.raw}
```
will produce as expected:
```
[&lt;i&gt;Hello&lt;/i&gt;, &lt;b&gt;World&lt;/b&gt;]
[<i>Hello</i>, <b>World</b>]
[...]

PR #7930, opened 2020-03-18
Qute type-safe templates - fix inconsitency in "raw" validation

mkouba:
- "{@java.lang.Object obj}{obj.raw}" should not result in an incorrect expression
- follows up on #7771


gsmet (approved):

gsmet:
Thanks!

PR #7933, opened 2020-03-18
Fix interceptor resolution process for multiple bindings, add automated test

### LINKED ISSUE #7931: Interceptor resolution with multiple bindings is not done accordingly to the specification
Given an interceptor with two bindings and a bean with only one of them, ArC currently triggers the interceptor for methods on that bean. This is not what the specification expects, it should only trigger if bean has both bindings.

manovotn:
…ed test.

Fixes #7931


mkouba (approved):

manovotn:
Side note - this is probably worth backporting.

mkouba:
@manovotn yep, that why I've added the `triage/backport?` label ;-).

PR #7935, opened 2020-03-18
Add SmallRye Metrics extension to Infinispan Embedded

wburns:
Related to https://github.com/quarkusio/quarkus/issues/7910 and https://github.com/quarkusio/quarkus/pull/7921

### DIFF
--- a/extensions/infinispan-embedded/runtime/pom.xml
+++ b/extensions/infinispan-embedded/runtime/pom.xml
@@ -25,10 +25,9 @@
             <groupId>io.quarkus</groupId>
             <artifactId>quarkus-caffeine</artifactId>
         </dependency>
-
         <dependency>
-            <groupId>io.smallrye</groupId>
-            <artifactId>smallrye-metrics</artifactId>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-smallrye-metrics</artifactId>
         </dependency>
         <dependency>
             <groupId>org.eclipse.microprofile.metrics</groupId>
--- a/extensions/infinispan-embedded/runtime/pom.xml
+++ b/extensions/infinispan-embedded/runtime/pom.xml
@@ -30,7 +30,6 @@
             <groupId>io.smallrye</groupId>
             <artifactId>smallrye-metrics</artifactId>
         </dependency>
-        <!-- Hack to get it to have the API Infinispan uses available -->
         <dependency>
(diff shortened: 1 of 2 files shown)


jmartisk (approved):

extensions/infinispan-embedded/runtime/pom.xml
```diff
@@ -30,7 +30,6 @@
             <groupId>io.smallrye</groupId>
             <artifactId>smallrye-metrics</artifactId>
         </dependency>
```
gsmet: You should remove that one and use `quarkus-smallrye-metrics` here and `quarkus-smallrye-metrics-deployment` in the deployment module.

gsmet (changes requested):
/cc @jmartisk

extensions/infinispan-embedded/runtime/pom.xml
wburns: Sounds good. I have no clue about this metrics stuff :)

jmartisk: I think there is no deployment-time dependency on metrics in this case. Metrics are handled by Infinispan directly, not by any Quarkus extension, so it all happens in runtime

wburns: So I must admit I am not sure what you guys want me to do :)

gsmet: @jmartisk we might need some reflection registration at some point and so on.

So if you have the runtime dependency, you need the deployment one.

gsmet: I pushed an additional commit with the change.

@jmartisk can you check it's OK with you?

wburns: @gsmet sorry I could have done that - I was more asking if I needed to actually do that.

gsmet: Ah OK, I misunderstood. Let's have @jmartisk check this. I think it's the right thing to do to be on the safe side and be sure SmallRye Metrics behave correctly in native mode. Even if the deployment extension doesn't have anything special for native atm (I haven't checked), it might not always be the case so better be on the safe side.

jmartisk: LGTM!

gsmet (approved):

PR #7937, opened 2020-03-18
Some proofreading of the recent additions to config.adoc

gsmet:


docs/src/main/asciidoc/config.adoc
```diff
@@ -444,31 +444,31 @@ Using `@ConfigProperty("quarkus.profile")` will *not* work properly.
 
 === Using placeholders
 
-Quarkus supports the use of placeholders in the `application.properties` file on build time.
+Quarkus supports the use of placeholders in the `application.properties` file at build time.
```
gsmet: @dmlloyd could you have a look at this new section?

I would have expected this to be also valid at runtime but maybe I'm too optimistic?

dmlloyd: Expressions work at the time that the configuration property is read.  If it's a build time property, the expression is expanded at build time, but if it's a run time property, the expression is preserved at build time and expanded at run time.

Also I'm not sure we should be using the term "placeholder"; the term we've used in MP Config so far is "expression" or "property expression".

gsmet: Yeah I didn't like the term placeholder either.

OK so we need some further tuning of this section.

dwamara: :-) I used that term translating the one in German that I had in mind when writing it. Will be more careful/inspired next time.

gsmet:
@dmlloyd I gave it another go. Could you have another look?

docs/src/main/asciidoc/config.adoc
```diff
@@ -442,41 +442,47 @@ The proper way to check the active profile programmatically is to use the `getAc
...
 ----
 callable.url=http://${remote.host}/
 ----
 
-will result on the value of the `callable.url` property to be set at build time to
+will result in the value of the `callable.url` property being set to:
 
 [source,properties]
 ----
 callable.url=quarkus.io
```
dmlloyd:
```suggestion
callable.url=http://quarkus.io/
```

dmlloyd (commented):
Only one suggestion (well I think it was an oversight).

docs/src/main/asciidoc/config.adoc
gsmet: Fixed, thanks!

dmlloyd (approved):

Issue #7543, opened 2020-03-03
Quarkus OIDC and HTTPS improvements


sberyozkin:
**Description**
We've agreed with Pedro that it is worth investing more time into testing and documenting that the code grants/introspection requests can go via HTTPS, and also possibly support a custom trust/key stores for the 2-way TLS. This follows a user query how to get Keycloak CA verified.

CC @pedroigor

sberyozkin:
We have also discussed the security of the various client authentication methods with @emmanuelbernard at the forum. The proposal is to warn the users that a client secret may be leaked if `auth-server-url` is insecure HTTP one and either `client_secret_basic` or `client_secret_post` is used.

CC @pedroigor

The only possible side-effect is if OIDC is collocated somewhere in the same protected zone when `HTTP` is enough ?

sberyozkin:
Or we can go even further and introduce the auto method selection, example, if it is `HTTP` and OIDC supports `client_server_basic`, `client_server_post` and `client_server_jwt` then we auto select `client_server_jwt` but I suspect that may have some unexpected side-effects, so we probably should keep to _basic as a default and then let users to choose based on their requirements

sberyozkin:
I'm going to close this one because we have a test confirming the HTTPS connection works now, we can reopen more specific TLS related issues going forward

PR #7940, opened 2020-03-18
[oidc] Add config to trust all certs and skip hostname verification

### LINKED ISSUE #5640: [oidc] Add Ability to Skip TLS Verify
For development purposes it would be interesting to be able to skip the tls verification of the oidc auth server.

fmhwong:
Add 2 new config to trust all server certificates and turn off hostname verification.

Fixes #5640

### DIFF
--- a/extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/OidcTenantConfig.java
+++ b/extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/OidcTenantConfig.java
@@ -91,6 +91,34 @@ public class OidcTenantConfig {
      */
     Authentication authentication = new Authentication();
 
+    /**
+     * TLS configurations
+     */
+    @ConfigItem
+    Tls tls = new Tls();
+
+    @ConfigGroup
+    public static class Tls {
+        public enum Verification {
+            /**
+             * Certificates are validated and hostname verification is enabled. This is the default value.
+             */
+            REQUIRED,
+            /**
+             * All certificated are trusted and hostname verification is disabled.
+             */
+            NONE
+        }
+
+        /**
+         * Certificate validation and hostname verification, which can be one of the following values from enum
+         * {@link Verification}. Default is required.
+         */
+        @ConfigItem(defaultValue = "REQUIRED")
+        Verification verification;
+
+    }
+
     public Optional<Duration> getConnectionDelay() {
         return connectionDelay;
     }
--- a/.github/workflows/ci-actions.yml
+++ b/.github/workflows/ci-actions.yml
@@ -110,6 +110,7 @@ jobs:
           JAVA_OPTS: "-server -Xms64m -Xmx512m -XX:MetaspaceSize=96M -XX:MaxMetaspaceSize=256m -Djava.net.preferIPv4Stack=true -Djava.awt.headless=true -Dkeycloak.profile.feature.upload_scripts=enabled"
         ports:
           - 127.0.0.1:8180:8080
+          - 127.0.0.1:8543:8443
       postgres:
         image: postgres:10.5
         env:
@@ -459,7 +460,7 @@ jobs:
         if: matrix.neo4j
       - name: Keycloak Service
         run: |
-          docker run --rm --publish 8180:8080 --name build-keycloak \
+          docker run --rm --publish 8180:8080 --publish 8543:8443 --name build-keycloak \
             -e KEYCLOAK_USER=admin -e KEYCLOAK_PASSWORD=admin -e JAVA_OPTS=" \
               -server -Xms64m -Xmx512m -XX:MetaspaceSize=96M \
               -XX:MaxMetaspaceSize=256m -Djava.net.preferIPv4Stack=true -Djava.awt.headless=true \
--- a/extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/OidcRecorder.java
+++ b/extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/OidcRecorder.java
@@ -15,6 +15,7 @@ import io.quarkus.oidc.OIDCException;
 import io.quarkus.oidc.runtime.OidcTenantConfig.ApplicationType;
 import io.quarkus.oidc.runtime.OidcTenantConfig.Credentials;
 import io.quarkus.oidc.runtime.OidcTenantConfig.Credentials.Secret;
+import io.quarkus.oidc.runtime.OidcTenantConfig.Tls.Verification;
 import io.quarkus.runtime.annotations.Recorder;
 import io.quarkus.runtime.configuration.ConfigurationException;
 import io.vertx.core.AsyncResult;
@@ -138,6 +139,11 @@ public class OidcRecorder {
         Optional<ProxyOptions> proxyOpt = toProxyOptions(oidcConfig.getProxy());
         if (proxyOpt.isPresent()) {
             options.setProxyOptions(proxyOpt.get());
+        }
+
+        if (oidcConfig.tls.verification == Verification.NONE) {
+            options.setTrustAll(true);
+            options.setVerifyHost(false);
         }
 
         final long connectionDelayInSecs = oidcConfig.getConnectionDelay().isPresent()
--- a/integration-tests/oidc/README.md
+++ b/integration-tests/oidc/README.md
@@ -16,7 +16,7 @@ Additionally, you can generate a native image and run the tests for this native 
 mvn clean install -Dtest-keycloak -Ddocker -Dnative
 ```
 
-If you don't want to run Keycloak Server as a Docker container, you can start your own Keycloak server. It needs to listen on the default port `8180`.
+If you don't want to run Keycloak Server as a Docker container, you can start your own Keycloak server. It needs to listen on the default https port `8543`.
 
 You can then run the tests as follows (either with `-Dnative` or not):
 
@@ -24,4 +24,4 @@ You can then run the tests as follows (either with `-Dnative` or not):
 mvn clean install -Dtest-keycloak
 ```
 
-If you have specific requirements, you can define a specific connection URL with `-Dkeycloak.url=http://keycloak.server.domain:8180/auth`.
+If you have specific requirements, you can define a specific connection URL with `-Dkeycloak.url=https://keycloak.server.domain:8543/auth`.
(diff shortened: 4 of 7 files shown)


sberyozkin:
@fmhwong Thanks for this PR. 

We should probably wrap these 2 properties into a `@ConfigGroup` `Tls` so that we can add more TLS related properties there going forward such as the custom key/truststore locations.

The other thing though, we will support an MTLS authentication method hopefully soon enough as discussed with @pedroigor, so, we should consider now IMHO how these 2 new properties will be used along the MTLS method, @pedroigor, what do you think ?

sberyozkin:
@fmhwong Can you please one of the 4 OIDC related integration test suites, lets choose `integration-tests/oidc` which is simplest, and update it to have Keycloak available at `HTTPS` ? I'm assuming it would require using the new properties and as such will test they are effective :-) Give it a try please.
You can run the test locally in `integration-tests/oidc` like this:
```
mvn clean install -Pdocker-keycloak -Ptest-keycloak
```

fmhwong:
Thanks for the comments! Will add a new `@ConfigGroup` and take a look of the test cases.

pedroigor:
> @fmhwong Thanks for this PR.
> 
> We should probably wrap these 2 properties into a `@ConfigGroup` `Tls` so that we can add more TLS related properties there going forward such as the custom key/truststore locations.

I think it makes sense to have such group so that users can set any global TLS related config.

Btw, what if we have just a single `hostname-verification` option and use a enum to define the behavior ? For instance, `hostname-verification=NONE` would disable it, etc.

> 
> The other thing though, we will support an MTLS authentication method hopefully soon enough as discussed with @pedroigor, so, we should consider now IMHO how these 2 new properties will be used alongside the MTLS method, @pedroigor, what do you think ?

For MTLS, I think this group is going to provide some core properties such as keystore/password/key, so that web type clients can authenticate to the authorization server using certificates. For service type clients, the config group should help to establish the connection to clients so that they can ask their certificates and validate certificate-bound access tokens.

Basically, two main use cases for the different types of applications we support.

sberyozkin:
Hi @pedroigor Thanks, this is helpful...
Sure, making `hostname-verification` an enum can be good, my understanding though it is only either on or off as far as the VertxOAuth2 is concerned. Please comment once @fmhwong introduces a group.
Thanks

fmhwong:
@pedroigor  Just want to clarify that you are suggesting to combine the proposed 2 new config `trustAllCertificates` and `hostnameVerification` into a single config `hostnameVerification` and use `quarkus.oidc.tls.hostname-verification=NONE` to trust all certs and disable hostname verification.

sberyozkin:
@fmhwong Hi, can you please wrap those properties in a group which is what Pedro is also fine with and force-push and then we will all discuss if/how the properties can be optimized, we will grow that `Tls` group anyway going forward. Cheers

fmhwong:
@sberyozkin Just did.  Please take a look. Thanks!

extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/OidcTenantConfig.java
```diff
@@ -86,6 +86,26 @@
      */
     Authentication authentication = new Authentication();
 
+    /**
+     * TLS configurations
+     */
+    @ConfigItem
+    TlsConfig tls;
```
sberyozkin: @fmhwong Please change it to `Tls tls = new Tls()` and rename the group to `Tls` so that we can have `quarkus.oidc.tls.trust-all-certificates`, etc, the `config` is assumed :-)

sberyozkin:
@pedroigor Hey Pedro, OK, the group is here, can you clarify please what you'd like to change ? Cheers

pedroigor:
@sberyozkin @fmhwong I think that we could squash both options in `tls` into a single one. One of the reasons is that they are quite related when you want to disable TLS verifications (hostname and certs). Another reason is that we could eventually support more options on how verification should work. I'm mainly saying that due to the fact that some HTTP client libraries provide different options to set how verification is done (e.g.: apache http client). So I think it may be useful and more flexible.

sberyozkin:
@pedroigor OK, so `Tls` group will have an enum (`Verification` ?), with `REQUIRED` and `NONE` being two initials values, something like this:
```
class Tls {
   enum Verification {
        // Default: the hostname verification is enabled
        REQUIRED,
        // the hostname verification is disabled, all certs are trusted
        NONE
        // other options as needed (trust specific hosts etc) ?       
   }
  @ConfigItem(defaultValue=Verification.REQUIRED) 
  Vericaition verification; 
}
```

something like that ?

fmhwong:
Thanks and I will make the change.

fmhwong:
I noticed the config already has `quarkus.vault.tls.skip-verify`.  Do you want to name this `quarkus.oidc.tls.skip-verify` and set the type to `boolean`?  Default to `false` and set to `true` to trust all certs and skip host verification.

fmhwong:
@sberyozkin @pedroigor What do you think?

sberyozkin:
@fmhwong We don't have it to align it with the `Vault` extension IMHO as it is the only property there and we'd like to keep the options more diverse with the `Tls` group.

pedroigor:
> @sberyozkin @pedroigor What do you think?

@fmhwong Good catch. Align sounds better than introducing a different way of doing it.

sberyozkin:
@pedroigor I'm sorry, I'm not following. That `Vault` property is a pretty basic way to configure it for Vault, why would we want to limit the configuration to a single boolean property now ? Would it also be a step back from what you proposed earlier (we have more flexible TLS support, etc) ?

sberyozkin:
If we were go back from the enum idea then I'd just stick to the current PR:
```
    @ConfigGroup
    public static class TlsConfig {
        /**
         * If all server certificates should be trusted
         */
        boolean trustAllCertificates;
        /**
         * If hostname verification is enabled
         */
        @ConfigItem(defaultValue = "true")
        boolean hostnameVerification;
    }
```
as these are 2 distinctive properties.

pedroigor:
@sberyozkin Yes, it is a step back ... I did not know we already had this in Quarkus.

My proposal is mainly based on what we currently have in Keycloak in regards to this type of verification. There we have an enum. So I thought that future-wise it would be nice to protect us from future changes.

But now that we know we have the same capability in another extension, it makes more sense to me to have the same. I do prefer the enum option, but in favor of a better UX, I'm OK to review my preference.

I also think the Vault approach is better than have two options. As we discussed, most of the times you are going to disable boths, right ?

sberyozkin:
@pedroigor I thought when you suggested to use an enum and mentioning it was more flexible that we'd want to support more than just `true`, `false` with respect to `Tls` verification.
My understanding is that right now, in Quarkus, different extensions where the HTTPS client invocations are involved control things differently. Having a uniform approach across all of the extensions would be a worthy idea to try to realize but it would require a dedicated efforts from many of us :-). In meantime even if we name this property the same as in Vault, users who use both Vault and OIDC would still have to type this property twice :-)

May be we need to back to a drawing board :-).

So this `Tls` group is a start where we will control different things related to `Tls` including `Mtls`.
We have several requirements, possibly somewhat orthogonal but all related to `Tls`:
1. Disable/enable the host/cert verification (this PR)
2. We have #7543 (I suppose we'd need to support setting the custom trustore/keystore ?)
These are to support the discovery and most of the client autentication methods over TLS
3.  and Client authentication MTLS method will need to be added in time
4. Do we want to support some specific hostname verification options ?

So `NONE/REQUIRED` enums lets us do 1. Or we can have instead `quarkus.oidc.tls.skip-verify`.
It also works.

The question is how we can move from there when addressing other requirements above.
Do you think `quarkus.oidc.tls.skip-verify` will let us build upon it ? Or enums are better ? 

`quarkus.oidc.tls.verification=none` or `quarkus.oidc.tls.verification=required` (default) (note Quarkus is fine with converting lower case values like `none` to `NONE`/etc :-) ) reads quite good to me :-)

pedroigor:
@sberyozkin OK, I agree to the `.verification=none|required`.

fmhwong:
Thanks guys! I will make the changes.

sberyozkin:
@pedroigor thanks, I don't really want to push for it, if you also feel OK about it then it is good. I'll CC Guillaume as he knows a lot about these TLS issues, @gsmet, Guillaume (and others), comments are welcome

sberyozkin:
@fmhwong Thanks, can you please update the test as agreed earlier ?

fmhwong:
Yes, will do

sberyozkin:
@fmhwong Hi, thanks for your effort, it looks good :-). But you can simplify. We have enough non-secure `http` already in various integration tests, so if you just update `integration-tests/oidc` then it should be fine IMHO, only update URLs to `https:` and disable the verification in the application properties and it will be enough :-)
Thanks again for spending the time on it

sberyozkin:
@fmhwong Thanks, IMHO it is ready to go as soon as CI passes. Please also squash all your commits.
Thanks

sberyozkin:
@fmhwong Hi, thanks for getting the tests fixed, the native build failed with the unrelated issue. You have definitely helped with starting the TLS related enhancements, it would've taken few cycles to trace that the ci actions have to be fixed, etc, nice work.
Can you please squash all your commits ?

fmhwong:
@sberyozkin I squashed the commits and all checks have passed.  Thanks for your help!

sberyozkin:
@fmhwong Thanks.

Now, before I merge, I'd like to circle a bit. I've been thinking during the last few days if your @fmhwong idea to align with Vault and just have `boolean skip-verify` should be revisited, something @pedroigor also thought was worth an attention. IMHO it is OK to consider it again, I certainly do it all the time with my PRs :-)

I've looked at how Vertx handles these 2 booleans.
`HttpClientOptions` deals with the host verification while `ClientOptionsBase` deals with `trustAll` - if it is false then it checks the trust options. But if we were to add the trust options then it would make `trustAll=false` anyway. I've been thinking that the enums will give us more options based on the original discussion with Pedro, but I'm not really seeing what else we can actually add to these `NONE` and `REQUIRED` now.

So I'd like to hold off the merge a bit. @fmhwong @pedroigor, can you think of at least one more useful enum value ? If yes - then we should keep the enum. If not lets just do `boolean skip-verify` :-)

Thanks all

sberyozkin (approved):
LGTM, but lets do one final discussion around the property names (use enum vs a single `boolean skipVerify`)

sberyozkin:
I'm going to go ahead with the merge later today if there will be no additional comments, the enums are more descriptive so it is still a good approach, I'd just like to give a bit more time to think about it.

PR #7947, opened 2020-03-19
Remove unused ClassOutput

### LINKED ISSUE #2923: Class generation build items are screwy
We have a `GeneratedClassBuildItem` which accepts a class name, byte array, and a flag to indicate what "kind" of class it is.

At the same time, we have a pattern for generating classes, particularly with Gizmo: consume the `ClassOutputBuildItem` and then feed it to Gizmo and have it write the class directly to the `ClassOutput`.

On top of all this, we have two different kinds of `ClassOutput`.

The build item is not ideal because to use it with Gizmo, you have to create your own `ClassOutput` (the Gizmo one) and have it produce the generated classes (i.e. you can't just return a `GeneratedClassBuildItem` unless you stash the object somewhere).

The common Gizmo pattern of consuming the `ClassOutputBuildItem` is not ideal because nothing is produced - so the build step won't run unless you artificially produce something.

We should make the `ClassOutputBuildItem` work nicely with Gizmo and non-Gizmo usages.  Maybe Gizmo should emit classes to an array by default, or maybe `ClassOutputBuildItem` should accept a supplier for class bytes that the Gizmo class factory can produce (if we're worried about heap usage).  Maybe Gizmo's `ClassOutput` API should be replaced by a pull API like `ClassCreator.write(OutputStream)` or `ClassCreator.getBytes()` (or both).

stuartwdouglas:
Fixes #2923


stuartwdouglas:
@dmlloyd I think most of the concerns are addressed with all the recent changes. You still can't easily just return a ClassOutputBuildItem but we have common adaptor types now and I don't really know if it is worth changing, especially as most build steps that can produce classes can produce multiple classes.

gastaldi (approved):

PR #7952, opened 2020-03-19
Document extra log levels

stuartwdouglas:


gastaldi (approved):

core/runtime/src/main/java/io/quarkus/runtime/logging/LogConfig.java
```diff
@@ -16,6 +16,16 @@
...
+     *
+     * In addition to the standard JDK log level JBoss Logging also adds the following:
+     *
+     * {@link org.jboss.logmanager.Level#FATAL}
+     * {@link org.jboss.logmanager.Level#ERROR}
+     * {@link org.jboss.logmanager.Level#WARN}
+     * {@link org.jboss.logmanager.Level#INFO}
+     * {@link org.jboss.logmanager.Level#DEBUG}
+     * {@link org.jboss.logmanager.Level#TRACE}
+     *
```
dmlloyd: If this is meant to be JavaDoc, then this should be formatted (add a `<p>` before `In addition...` and use `<ul>`/`<li>` for the links).  If it's meant to be AsciiDoc (remember that conversation?) then we probably can't use `@link`.

gastaldi: Ah crap, I should have waited for your review before merging. Oh well :stuck_out_tongue_closed_eyes:

PR #7959, opened 2020-03-19
Load wagon providers from the classloader that loaded the bootstrap classes

aloubyansky:
...instead of the current classloader which may already be an application Quarkus classloader


geoand:
@gsmet @ia3andy where you going to test something with regards to this?

gsmet (approved):

PR #7960, opened 2020-03-19
fix NPE in QuarkusTestExtension

### LINKED ISSUE #7958: NPE in QuarkusTestExtension when test method argument is null
**Describe the bug**
If I have a `@QuarkusTest` and some of the test method is passed `null` as an argument, `QuarkusTestExtension` throws a NPE. Getting `null` as an argument can happen for various reasons, for example when I use parameterized tests and want to use `null` as one of the parameters.

**Expected behavior**
`null` is a valid argument that test methods can receive

**Actual behavior**
NPE

**To Reproduce**
1. Create a `@QuarkusTest`.
2. Add a `@ParameterizedTest` with `@NullSource`.
3. Observe NPE.

Ladicek:
Resolves #7958.

### DIFF
--- a/integration-tests/main/src/test/java/io/quarkus/it/main/ParameterizedSimpleTestCase.java
+++ b/integration-tests/main/src/test/java/io/quarkus/it/main/ParameterizedSimpleTestCase.java
@@ -1,0 +1,32 @@
+package io.quarkus.it.main;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNull;
+
+import org.junit.jupiter.params.ParameterizedTest;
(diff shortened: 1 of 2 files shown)


Ladicek:
Created as a draft because I want to clarify one thing: I'm not entirey happy about adding yet another module to the test suite (which makes the build unnecessary longer, even if it's just a few seconds), but I didn't find a suitable place. Any ideas?

geoand:
When I fixed some issue with the `@ParameterizedTest` I just added the new tests to the `main` module.

Ladicek:
I'll do the same, then :-)

Ladicek:
Done.

geoand (approved):
Thanks a lot!

PR #7909, opened 2020-03-17
Quarkus Funqy FaaS Framework - initial implementation

patriot1burke:
* Initial implementaiton of Quarkus Funqy FaaS Framework
* Refactored amazon-lambda extension into some reusable common modules
* Added amazon-lambda stream handler test

PR #7925, opened 2020-03-18
Add quarkus-test-amazon-lambda to BOM

oztimpower:
Add the **quarkus-test-amazon-lambda** test-framework module version to the BOM.  This is useful as the integration test has been added to the Maven archetype for the Amazon Lambda extension.

I can now remove the version from the POM for the Maven archetype, as tested below.

Tested by:

1. mvn archetype:generate -DarchetypeGroupId=io.quarkus -DarchetypeArtifactId=quarkus-amazon-lambda-archetype -DarchetypeVersion=999-SNAPSHOT
2. mvn package

/cc @patriot1burke 
/cc @geoand

PR #7971, opened 2020-03-19
Gradle added to Amazon Lambda archetype

oztimpower:
Added Gradle support to the generic archetype project, to simplify the use of Gradle for new starters.  They have the option now to use Maven or Gradle and use the archetype project as a base for creating AWS Lambda projects.

**code.quarkus.io** only generates a sample HTTP project, it doesn't generate an AWS Lambda sample project, if someone is wanting to use Gradle.

** Tested by: **

mvn archetype:generate -DarchetypeGroupId=io.quarkus -DarchetypeArtifactId=quarkus-amazon-lambda-archetype -DarchetypeVersion=999-SNAPSHOT
mvn package
gradle assemble
(Unit test for Gradle fails per #7909 

/cc @patriot1burke 
/cc @Nxtra

### DIFF
--- a/extensions/amazon-lambda/maven-archetype/src/main/resources/archetype-resources/build.gradle
+++ b/extensions/amazon-lambda/maven-archetype/src/main/resources/archetype-resources/build.gradle
@@ -1,0 +1,32 @@
+plugins {
+    id 'java'
+    id 'io.quarkus'
+}
+
+repositories {
+     mavenLocal()
+     mavenCentral()
+}
+
+dependencies {
+    implementation enforcedPlatform("${quarkusPlatformGroupId}:${quarkusPlatformArtifactId}:${quarkusPlatformVersion}")
+    implementation 'io.quarkus:quarkus-resteasy'
+    implementation 'io.quarkus:quarkus-amazon-lambda'
+
+    testImplementation  "io.quarkus:quarkus-test-amazon-lambda"
+    testImplementation 'io.quarkus:quarkus-junit5'
+    testImplementation 'io.rest-assured:rest-assured'
+}
+
+group '${groupId}'
+version '${version}'
+
+compileJava {
+    options.compilerArgs << '-parameters'
+}
+
+java {
+    sourceCompatibility = JavaVersion.VERSION_1_8
+    targetCompatibility = JavaVersion.VERSION_1_8
+}
+
--- a/extensions/amazon-lambda/maven-archetype/pom.xml
+++ b/extensions/amazon-lambda/maven-archetype/pom.xml
@@ -30,6 +30,13 @@
                     <exclude>archetype-resources/pom.xml</exclude>
                 </excludes>
             </resource>
+            <resource>
+                <directory>src/main/resources</directory>
+                <filtering>true</filtering>
+                <includes>
+                    <include>archetype-resources/gradle.properties</include>
+                </includes>
+            </resource>
         </resources>
         <extensions>
             <extension>
--- a/extensions/amazon-lambda/maven-archetype/src/main/resources/META-INF/maven/archetype-metadata.xml
+++ b/extensions/amazon-lambda/maven-archetype/src/main/resources/META-INF/maven/archetype-metadata.xml
@@ -41,11 +41,19 @@ under the License.
         <include>*</include>
       </includes>
     </fileSet>
+    <fileSet>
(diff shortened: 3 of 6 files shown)


Nxtra:
We also need to update `manage.sh` in the archetype to work with the artifact that Gradle creates in the `/build` directory. 
Then you can update `amazon-lambda.adoc` and add a choice between Gradle or Maven.

Nxtra (changes requested):

extensions/amazon-lambda/maven-archetype/src/main/resources/archetype-resources/build.gradle
```diff
@@ -0,0 +1,32 @@
...
+dependencies {
+    implementation enforcedPlatform("${quarkusPlatformGroupId}:${quarkusPlatformArtifactId}:${quarkusPlatformVersion}")
+    implementation 'io.quarkus:quarkus-resteasy'
+    implementation 'io.quarkus:quarkus-amazon-lambda'
+
+    testImplementation  "io.quarkus:quarkus-test-amazon-lambda"
+    testImplementation 'io.quarkus:quarkus-junit5'
+    testImplementation 'io.rest-assured:rest-assured'
+}
+
```
Nxtra: When shipping the native binary to Lambda we need to create the `function.zip` file with the binary renamed to `bootstrap` insight. Can we use the a combination of the `distribution` plugin, `distZip` or custom `buildZip` task to do that?

Nxtra (changes requested):

extensions/amazon-lambda/maven-archetype/src/main/resources/archetype-resources/build.gradle
patriot1burke: Don't worry too much about that.  I'm currently working on having the extension generate the zip.

extensions/amazon-lambda/maven-archetype/src/main/resources/archetype-resources/build.gradle
```diff
@@ -0,0 +1,32 @@
...
+     mavenLocal()
+     mavenCentral()
+}
+
+dependencies {
+    implementation enforcedPlatform("${quarkusPlatformGroupId}:${quarkusPlatformArtifactId}:${quarkusPlatformVersion}")
+    implementation 'io.quarkus:quarkus-resteasy'
+    implementation 'io.quarkus:quarkus-amazon-lambda'
+
+    testImplementation  "io.quarkus:quarkus-test-amazon-lambda"
```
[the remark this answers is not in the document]
oztimpower: If you're on HEAD (999-SNAPSHOT), it will have this already, per PR #7925

extensions/amazon-lambda/maven-archetype/src/main/resources/archetype-resources/build.gradle
oztimpower: Quarkus will generate this for you, via gradle buildNative.  No need for custom tasks.

patriot1burke (approved):

extensions/amazon-lambda/maven-archetype/src/main/resources/archetype-resources/build.gradle
Nxtra: @patriot1burke okay, so for now I rename the runner to bootstrap and zip it up to `function.zip`?
@oztimpower , quarkus will generate the runner but atm only mvn is creating the `function.zip`. As I understand this will be covered by @patriot1burke 's extension.

oztimpower: Yes apologies, I forgot I had similar workarounds as you on the function.zip with Gradle, I did it via a one liner in manage.sh, and passed a flag if it was gradle or maven (to toggle the target/build).  Maybe I should PR that also.  For me I've always kept a Maven and Gradle build up to date on the projects as Gradle is only Preview and the Maven version put simply "just works"

extensions/amazon-lambda/maven-archetype/src/main/resources/archetype-resources/build.gradle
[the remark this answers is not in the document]
Nxtra: Got it, thanks!

PR #7975, opened 2020-03-19
Fix Maven minimal version

gsmet:


geoand (approved):
Right, I was wondering about that...

gastaldi (approved):

PR #7978, opened 2020-03-19
Update to SmallRye Health 2.2.0

antoinesd:
Add @HealthGroup and @HealthGroups support

### DIFF
--- a/extensions/smallrye-health/deployment/src/main/java/io/quarkus/smallrye/health/deployment/SmallRyeHealthProcessor.java
+++ b/extensions/smallrye-health/deployment/src/main/java/io/quarkus/smallrye/health/deployment/SmallRyeHealthProcessor.java
@@ -1,6 +1,7 @@
 package io.quarkus.smallrye.health.deployment;
 
 import java.io.IOException;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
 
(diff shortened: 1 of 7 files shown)


xstefank (approved):

extensions/smallrye-health/deployment/src/main/java/io/quarkus/smallrye/health/deployment/SmallRyeHealthProcessor.java
```diff
@@ -72,6 +84,12 @@
          */
         @ConfigItem(defaultValue = "/ready")
         String readinessPath;
+
+        /**
+         * The relative path of the health group servlet.
```
gastaldi: We should at some point refer to `endpoints` instead of `servlets` :smiley:

gastaldi (approved):

extensions/smallrye-health/runtime/pom.xml
```diff
@@ -18,12 +18,6 @@
         <dependency>
             <groupId>io.smallrye</groupId>
             <artifactId>smallrye-health</artifactId>
-            <exclusions>
```
geoand: If the exclusion is no longer needed, does that mean that `smallrye-health` now ships with `jakarta.inject`? If so, then the dependency below should be removed

extensions/smallrye-health/deployment/src/main/java/io/quarkus/smallrye/health/deployment/SmallRyeHealthProcessor.java
geoand: I don't think mentioning `servlet` is correct, right?

gsmet:
@antoinesd @kenfinnigan is it necessary for MP 3.3 implementation? In other words: do we need to backport this to 1.3?

kenfinnigan:
It's not part of the spec so not required.

Having for 1.4 is fine

extensions/smallrye-health/runtime/pom.xml
antoinesd: > If the exclusion is no longer needed, does that mean that smallrye-health now ships with jakarta.inject?

Yes it is.

> If so, then the dependency below should be removed

You're right! Thanks

geoand: Excellent, thanks!

gsmet (dismissed):
Could you squash the 2 commits?

Thanks!

antoinesd:
Done

geoand (approved):

PR #7983, opened 2020-03-19
Remove now obsolete guide

geoand:


gastaldi (approved):

gsmet:
@geoand my understanding is that somehow the `kubernetes.adoc` is now replacing this one? If true maybe we should rename it to the name we had here to be consistent with the other deploying- guides.

Maybe better to have a quick chat on Zulip.

geoand:
I'm out for an hour or so, so if you need to do something now, go ahead 😎

gsmet:
No it can wait.

geoand:
@gsmet pretty much yeah - `kubernetes.adoc` now does it all :)

Issue #7955, opened 2020-03-19
CombinedIndexBuildItem to receive ClassInfo on inherited interfaces


oztimpower:
**Description**
For example, per Amazon Lambda extension documentation updates:

> Writing Skills with the Amazon Alexa ASK SDK v2
> 
>  The starting point for an Alexa Skill on AWS Lambda is typically via subclassing `SkillStreamHandler`, which implements the `RequestStreamHandler::requestHandler` method.  If you examine the source code for this class it is very simple, and can be implemented directly without this helper base class.
> 
>  NOTE: Quarkus requires a direct implementation of `RequestStreamHandler`, as the build process will scan only directly implemented interfaces, inherited interfaces will not be seen.

**Implementation ideas**

Generated ClassInfo, as part of the CombinedIndexBuildItem, to include all inherited interfaces. It currently only contains directly implemented interfaces.

For example in Amazon Lambda, AmazonLambdaProcesser, the code below looking for a specific interface implementation, would see the inherited interfaces.

`      Collection<ClassInfo> allKnownImplementors = combinedIndexBuildItem.getIndex().getAllKnownImplementors(REQUEST_HANDLER);
`
```
    @BuildStep
    List<AmazonLambdaBuildItem> discover(CombinedIndexBuildItem combinedIndexBuildItem,
            Optional<ProvidedAmazonLambdaHandlerBuildItem> providedLambda,
            BuildProducer<AdditionalBeanBuildItem> additionalBeanBuildItemBuildProducer,
            BuildProducer<ReflectiveHierarchyBuildItem> reflectiveHierarchy,
            BuildProducer<ReflectiveClassBuildItem> reflectiveClassBuildItemBuildProducer) throws BuildException {
  
      Collection<ClassInfo> allKnownImplementors = combinedIndexBuildItem.getIndex().getAllKnownImplementors(REQUEST_HANDLER);

      Collection<ClassInfo> allKnownStreamImplementors = combinedIndexBuildItem.getIndex()
                .getAllKnownImplementors(REQUEST_STREAM_HANDLER);


```
/cc @gsmet 
/cc @patriot1burke

geoand:
I think this request is essentially this issue: https://github.com/wildfly/jandex/issues/65

gsmet:
For now, I would just also get the classes implementing `SkillStreamHandler` and merge them with the others.

Not ideal and we should definitely pursue this on the Jandex side but good enough for this particular case.

[...]

PR #7985, opened 2020-03-19
Alexa SkillStreamHandler subclass check for RequestStreamHandler in AmazonLambdaProcessor

oztimpower:
Per suggestion of @gsmet on #7955, I have added a simple 2 line check for subclasses of the Alexa SkillStreamHandler in the Amazon Lambda extension deployment module, AmazonLambdaProcessor.  

It worked with my sample project.

runtime bom updated with the relevant Alexa SDK dependency, which also excludes conflicting dependencies of sl4j.  Checked via mvn dependency:tree

/cc @gsmet 
/cc @geoand 
/cc @patriot1burke

### DIFF
--- a/extensions/amazon-lambda/deployment/src/main/java/io/quarkus/amazon/lambda/deployment/AmazonLambdaProcessor.java
+++ b/extensions/amazon-lambda/deployment/src/main/java/io/quarkus/amazon/lambda/deployment/AmazonLambdaProcessor.java
@@ -50,6 +50,7 @@ public final class AmazonLambdaProcessor {
 
     private static final DotName REQUEST_HANDLER = DotName.createSimple(RequestHandler.class.getName());
     private static final DotName REQUEST_STREAM_HANDLER = DotName.createSimple(RequestStreamHandler.class.getName());
+    private static final DotName SKILL_STREAM_HANDLER = DotName.createSimple("com.amazon.ask.SkillStreamHandler");
 
     private static final DotName NAMED = DotName.createSimple(Named.class.getName());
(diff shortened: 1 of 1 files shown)


extensions/amazon-lambda/deployment/src/main/java/io/quarkus/amazon/lambda/deployment/AmazonLambdaProcessor.java
```diff
@@ -50,6 +51,7 @@
 
     private static final DotName REQUEST_HANDLER = DotName.createSimple(RequestHandler.class.getName());
     private static final DotName REQUEST_STREAM_HANDLER = DotName.createSimple(RequestStreamHandler.class.getName());
+    private static final DotName SKILL_STREAM_HANDLER = DotName.createSimple(SkillStreamHandler.class.getName());
```
gsmet: You should use the class name as a string instead of a class. This way, you won't need the dependency if not required.

gsmet (changes requested):
Cool! I added one comment inline, could you adjust your PR?

extensions/amazon-lambda/deployment/src/main/java/io/quarkus/amazon/lambda/deployment/AmazonLambdaProcessor.java
oztimpower: Ok, yes, a far simpler way to do it.

That's done.  Validated against my sample Alexa project.

geoand (approved):
:+1:

gsmet (approved):

gsmet:
This is very cool, thanks!

PR #7987, opened 2020-03-19
fix OpenTracing trace context propagation with Fault Tolerance @Asynchronous methods

### LINKED ISSUE #7981: test JDBC tracing with OpenTracing
**Description**
This is documented (https://quarkus.io/guides/opentracing#additional-instrumentation), but not tested. Should be easy to do, as the OpenTracing unit tests are already prepared for it, someone just forgot to write the actual test.

### LINKED ISSUE #7982: test OpenTracing trace context propagation to @Asynchronous Fault Tolerance methods
**Description**
There already seems to be a test for this in the OpenTracing unit test, but it doesn't use `@Asynchronous`, so there's nothing being propagated really. Should be easy to adjust.

### LINKED ISSUE #7984: OpenTracing trace context is not propagated to Fault Tolerance @Asynchronous methods
**Describe the bug**
If I use OpenTracing and invoke an `@Asynchronous` Fault Tolerance method, the trace context should be propagated. Any spans created in the `@Asynchronous` method should be part of the same trace.

**Expected behavior**
Tracing context is propagated.

**Actual behavior**
Tracing context is not propagated.

**To Reproduce**
1. Add `quarkus-smallrye-opentracing` and `quarkus-smallrye-fault-tolerance`.
2. Invoke an `@Asynchronous` method that does some tracing from a `@Traced` method.
3. Observe that spans are not part of the same trace.

Ladicek:
There used to be a test for that, but didn't involve `@Asynchronous`
methods, so was in fact useless. This commit changes the test to
actually call an `@Asynchronous` method, which revealed an issue.
The `smallrye-fault-tolerance-tracing-propagation` dependency was
missing, so trace context propagation didn't actually work.

Then, the test also uncovered an issue in SmallRye Fault Tolerance
(see https://github.com/smallrye/smallrye-fault-tolerance/pull/208),
which is why this commit also brings a new version of it.

In addition to that, this commit also fixes a devmode issue where
on hot reload, the Jaeger extension tried to register a gauge
that was already present. That ended up with an exception.

And finally, this commit also adds a test for tracing JDBC.

Resolves #7981.
Resolves #7982.
Resolves #7984.

### DIFF
--- a/extensions/smallrye-opentracing/deployment/pom.xml
+++ b/extensions/smallrye-opentracing/deployment/pom.xml
@@ -41,6 +41,11 @@
 
         <dependency>
             <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-resteasy-jackson</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
             <artifactId>quarkus-rest-client-deployment</artifactId>
             <scope>test</scope>
         </dependency>
--- a/bom/runtime/pom.xml
+++ b/bom/runtime/pom.xml
@@ -37,7 +37,7 @@
         <smallrye-metrics.version>2.4.1</smallrye-metrics.version>
         <smallrye-open-api.version>1.2.1</smallrye-open-api.version>
         <smallrye-opentracing.version>1.3.4</smallrye-opentracing.version>
-        <smallrye-fault-tolerance.version>4.1.1</smallrye-fault-tolerance.version>
+        <smallrye-fault-tolerance.version>4.2.0</smallrye-fault-tolerance.version>
         <smallrye-jwt.version>2.1.1</smallrye-jwt.version>
         <smallrye-context-propagation.version>1.0.11</smallrye-context-propagation.version>
         <smallrye-reactive-streams-operators.version>1.0.10</smallrye-reactive-streams-operators.version>
@@ -1445,6 +1445,11 @@
             <dependency>
                 <groupId>io.smallrye</groupId>
                 <artifactId>smallrye-fault-tolerance-context-propagation</artifactId>
+                <version>${smallrye-fault-tolerance.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>io.smallrye</groupId>
+                <artifactId>smallrye-fault-tolerance-tracing-propagation</artifactId>
                 <version>${smallrye-fault-tolerance.version}</version>
             </dependency>
             <dependency>
--- a/extensions/jaeger/deployment/src/main/java/io/quarkus/jaeger/deployment/JaegerProcessor.java
+++ b/extensions/jaeger/deployment/src/main/java/io/quarkus/jaeger/deployment/JaegerProcessor.java
@@ -2,6 +2,7 @@ package io.quarkus.jaeger.deployment;
 
 import javax.inject.Inject;
 
+import io.quarkus.deployment.Capabilities;
 import io.quarkus.deployment.annotations.BuildProducer;
 import io.quarkus.deployment.annotations.BuildStep;
 import io.quarkus.deployment.annotations.ExecutionTime;
@@ -21,13 +22,15 @@ public class JaegerProcessor {
     @BuildStep
     @Record(ExecutionTime.RUNTIME_INIT)
     void setupTracer(JaegerDeploymentRecorder jdr, JaegerBuildTimeConfig buildTimeConfig, JaegerConfig jaeger,
-            ApplicationConfig appConfig) {
+            ApplicationConfig appConfig, Capabilities capabilities) {
 
         // Indicates that this extension would like the SSL support to be enabled
         extensionSslNativeSupport.produce(new ExtensionSslNativeSupportBuildItem(FeatureBuildItem.JAEGER));
 
         if (buildTimeConfig.enabled) {
-            jdr.registerTracer(jaeger, appConfig);
+            boolean metricsEnabled = capabilities.isCapabilityPresent(Capabilities.METRICS);
+
+            jdr.registerTracer(jaeger, appConfig, metricsEnabled);
         }
     }
 
--- a/extensions/jaeger/runtime/pom.xml
+++ b/extensions/jaeger/runtime/pom.xml
@@ -18,7 +18,7 @@
             <groupId>io.quarkus</groupId>
             <artifactId>quarkus-core</artifactId>
         </dependency>
-       <dependency>
+        <dependency>
             <groupId>io.jaegertracing</groupId>
             <artifactId>jaeger-core</artifactId>
         </dependency>
@@ -31,12 +31,9 @@
             <artifactId>svm</artifactId>
         </dependency>
         <dependency>
-            <groupId>io.smallrye</groupId>
-            <artifactId>smallrye-metrics</artifactId>
-        </dependency>
-        <dependency>
-            <groupId>org.eclipse.microprofile.metrics</groupId>
-            <artifactId>microprofile-metrics-api</artifactId>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-smallrye-metrics</artifactId>
+            <optional>true</optional>
         </dependency>
     </dependencies>
 
--- a/extensions/jaeger/runtime/src/main/java/io/quarkus/jaeger/runtime/JaegerDeploymentRecorder.java
+++ b/extensions/jaeger/runtime/src/main/java/io/quarkus/jaeger/runtime/JaegerDeploymentRecorder.java
@@ -15,7 +15,7 @@ public class JaegerDeploymentRecorder {
     private static final Optional UNKNOWN_SERVICE_NAME = Optional.of("quarkus/unknown");
     private static final QuarkusJaegerTracer quarkusTracer = new QuarkusJaegerTracer();
 
-    synchronized public void registerTracer(JaegerConfig jaeger, ApplicationConfig appConfig) {
+    synchronized public void registerTracer(JaegerConfig jaeger, ApplicationConfig appConfig, boolean metricsEnabled) {
         if (!jaeger.serviceName.isPresent()) {
             if (appConfig.name.isPresent()) {
                 jaeger.serviceName = appConfig.name;
@@ -24,6 +24,7 @@ public class JaegerDeploymentRecorder {
             }
         }
         initTracerConfig(jaeger);
+        quarkusTracer.setMetricsEnabled(metricsEnabled);
         quarkusTracer.reset();
         // register Quarkus tracer to GlobalTracer.
         // Usually the tracer will be registered only here, although consumers
--- a/extensions/jaeger/runtime/src/main/java/io/quarkus/jaeger/runtime/QuarkusJaegerTracer.java
+++ b/extensions/jaeger/runtime/src/main/java/io/quarkus/jaeger/runtime/QuarkusJaegerTracer.java
@@ -2,6 +2,8 @@ package io.quarkus.jaeger.runtime;
 
 import io.jaegertracing.Configuration;
 import io.jaegertracing.internal.JaegerTracer;
+import io.jaegertracing.internal.metrics.NoopMetricsFactory;
+import io.jaegertracing.spi.MetricsFactory;
 import io.opentracing.ScopeManager;
 import io.opentracing.Span;
 import io.opentracing.SpanContext;
@@ -14,9 +16,14 @@ public class QuarkusJaegerTracer implements Tracer {
     private volatile JaegerTracer tracer;
 
     private boolean logTraceContext;
+    private boolean metricsEnabled;
 
     void setLogTraceContext(boolean logTraceContext) {
         this.logTraceContext = logTraceContext;
+    }
+
+    void setMetricsEnabled(boolean metricsEnabled) {
+        this.metricsEnabled = metricsEnabled;
     }
 
     @Override
@@ -35,8 +42,11 @@ public class QuarkusJaegerTracer implements Tracer {
         if (tracer == null) {
             synchronized (this) {
                 if (tracer == null) {
+                    MetricsFactory metricsFactory = metricsEnabled
+                            ? new QuarkusJaegerMetricsFactory()
+                            : new NoopMetricsFactory();
                     tracer = Configuration.fromEnv()
-                            .withMetricsFactory(new QuarkusJaegerMetricsFactory())
+                            .withMetricsFactory(metricsFactory)
                             .getTracerBuilder()
                             .withScopeManager(getScopeManager())
                             .build();
--- a/extensions/smallrye-opentracing/deployment/src/test/java/io/quarkus/smallrye/opentracing/deployment/RestService.java
+++ b/extensions/smallrye-opentracing/deployment/src/test/java/io/quarkus/smallrye/opentracing/deployment/RestService.java
@@ -1,10 +1,13 @@
 package io.quarkus.smallrye.opentracing.deployment;
 
+import java.util.List;
+import java.util.concurrent.CompletionStage;
+
 import javax.ws.rs.GET;
 import javax.ws.rs.Path;
+import javax.ws.rs.Produces;
+import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
-
-import org.eclipse.microprofile.opentracing.Traced;
 
 @Path("/")
 public interface RestService {
@@ -22,7 +25,11 @@ public interface RestService {
     Response restClient();
 
     @GET
-    @Traced(false)
     @Path("/faultTolerance")
-    Response faultTolerance();
+    CompletionStage<String> faultTolerance();
+
+    @GET
+    @Path("/jpa")
+    @Produces(MediaType.APPLICATION_JSON)
+    List<Fruit> jpa();
 }
--- a/extensions/smallrye-opentracing/deployment/src/test/java/io/quarkus/smallrye/opentracing/deployment/Service.java
+++ b/extensions/smallrye-opentracing/deployment/src/test/java/io/quarkus/smallrye/opentracing/deployment/Service.java
@@ -1,10 +1,15 @@
 package io.quarkus.smallrye.opentracing.deployment;
 
 import java.time.temporal.ChronoUnit;
+import java.util.List;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionStage;
 
 import javax.enterprise.context.ApplicationScoped;
 import javax.inject.Inject;
+import javax.persistence.EntityManager;
 
+import org.eclipse.microprofile.faulttolerance.Asynchronous;
 import org.eclipse.microprofile.faulttolerance.Fallback;
 import org.eclipse.microprofile.faulttolerance.Retry;
 import org.eclipse.microprofile.faulttolerance.Timeout;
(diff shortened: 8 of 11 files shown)


extensions/smallrye-opentracing/deployment/pom.xml
```diff
@@ -30,6 +30,20 @@
             <groupId>io.quarkus</groupId>
             <artifactId>quarkus-smallrye-opentracing</artifactId>
         </dependency>
+        <dependency>
+            <groupId>io.smallrye</groupId>
+            <artifactId>smallrye-fault-tolerance-tracing-propagation</artifactId>
```
Ladicek: So this is what I'd like to ask about. I'm not exactly sure where to add this dependency so that it also works in user applications (it obviously works in the unit test, which is in the same module :-) ).

The intent is that this module is available when both `quarkus-smallrye-fault-tolerance` and `quarkus-smallrye-opentracing` are present.

gsmet: /cc @aloubyansky @stuartwdouglas 

I don't think we have anything for that. Not sure how we could solve this.

Ladicek: Some extensions already bring other extensions transitively (e.g. I think Fault Tolerance brings Context Propagation). We can do the same here, but not sure from which side :-)

aloubyansky: @Ladicek by work in user applications you mean as a runtime dependency?

To this day, runtime dependencies remain in user's hands, expressed in traditional maven/gradle way. And e.g. `dependency:tree` (and the gradle equivalents) is still showing the actual runtime dependency tree. From this perspective, it's not like we don't have anything for that, it's the driving build system.
We could definitely come up with an enhancement to be able to stick other dependencies into the runtime classpath, it will break what's mentioned above though.

I think today, you'd have to either add it as a dependency of the runtime artifact or ask users to add it explicitly as an application dependency.

aloubyansky: > Some extensions already bring other extensions transitively (e.g. I think Fault Tolerance brings Context Propagation). We can do the same here, but not sure from which side :-)

The least popular one :)

Ladicek: Hi @michalszynkiewicz @pavolloffay @loicmathieu, do you have an opinion here? Would it be better for Fault Tolerance to automatically bring OpenTracing, or the other way around?

michalszynkiewicz: @Ladicek isn't a mistake (my mistake :D) that `smallrye-fault-tolerance-tracing-propagation` depends on `smallrye-fault-tolerance`?
If we could drop this dependency, we don't need to bring the whole fault tolerance with opentracing, right?

Ladicek: Yea, correct. In fact, just moving the dependency on `smallrye-fault-tolerance-tracing-propagation` to the `runtime` module should do, no? It brings some unnecessary dependencies on SmallRye FT, but those should basically be ignored, because the Quarkus extension isn't present. Or is that a complete misunderstanding?

stuartwdouglas: This is actually a relatively common use case, so I think we should add something to handle it in our resolution/classloading infrasturucture.

There are lots of cases like this where you only want an integration module if two different dependencies are present, I am not 100% sure what the best way to express/handle this would be though.

loicmathieu: > It brings some unnecessary dependencies on SmallRye FT, but those should basically be ignored, because the Quarkus extension isn't present.

In JVM mode the JAR will be included inside the final artifact by Maven. We do nothing about it. I'm not sure we can prune some dependencies out of the build easily.

You should use one of these two Maven tricks to avoid pulling all dependencies :
- Add your dependency as optional: it's what we do for health and metrics, not sure it can be used here as you want to include this one if both are present ...
- Exclude all transitive dependencies that will be included if the other libary (fault tolerance) is included.

Then use the `Capabilities` facility to check if both FT and context-propagation are present and add the functionality inside the Quarkus build. But apparently the fonctionality is automatically added thanks to the presence of the dependency so this can be tricky.

I think the best is what @michalszynkiewicz suggest: make `smallrye-fault-tolerance-tracing-propagation` not depending on `smallrye-fault-tolerance` if possible.

Ladicek: Oh, dependency exclusions should work. Basically, `quarkus-smallrye-opentracing` would depend on `smallrye-fault-tolerance-tracing-propagation` (but that dependency would exclude all transitive dependencies). It only contains 1 class and a `ServiceLoader` file for MP Context Propagation. Hmm, I wonder what would happen if this would be present, ConProp would be present, but FT would be missing... Well, since `smallrye-fault-tolerance-tracing-propagation` doesn't really depend on anything FT, we would probably do just the right thing -- have tracing context propagated. I need to test this.

(Rant: OpenTracing needs to integrate with ConProp, then we could remove this from FT entirely. We need ConProp in MP Platform!)

Ladicek: OK, that works, I've updated the PR.

Ladicek: I mean: I have made `smallrye-fault-tolerance-tracing-propagation` a dependency of `quarkus-smallrye-opentracing`, but all transitive dependencies are excluded. This works both when Fault Tolerance is present and missing. This is because `smallrye-fault-tolerance-tracing-propagation` in fact integrates MP OpenTracing with MP Context Propagation, and it only lives in SRye Fault Tolerance by accident.

I've tested that in an external test suite; adding tests to Quarkus would require adding 1 or 2 Maven modules, which is something I'd like to avoid.

Ladicek:
Hmm, the failing test passes for me... I'll dig into it.

Ladicek:
Hmm, it seems the test is quite flaky :-(

Ladicek:
I can reproduce the test failure locally when I let the test repeat 1000 times. It typically takes cca 400 attempts for the failure to occur. Not terribly fun to debug and I suspect there's some thread safety issue buried deep in ... I don't know where :-)

gsmet:
Not sure if it makes sense to backport that one, given the tests are flaky?

Ladicek:
I would _love_ to have this backported, but the test issue indeed needs to be resolved first.

Ladicek:
I tracked this down to a SmallRye Fault Tolerance threading issue: https://github.com/smallrye/smallrye-fault-tolerance/pull/208 Will update this PR when new SmallRye Fault Tolerance is released.

Ladicek:
This is now ready.

The failing test is unrelated: `io.quarkus.mongodb.reactive.legacy.AxleReactiveMongoClientTest`.

The test that was previously failing (`TracingTest`) is now OK. I've ran it many thousands of times locally while figuring out what's wrong (there's was a concurrency issue in SmallRye Fault Tolerance which is fixed in new version).

gsmet:
@jmartisk could you have a look? Thanks!

extensions/jaeger/runtime/src/main/java/io/quarkus/jaeger/runtime/QuarkusJaegerMetricsFactory.java
```diff
@@ -45,12 +46,17 @@ public void durationMicros(long time) {
 
     @Override
     public Gauge createGauge(final String name, final Map<String, String> tags) {
-        JaegerGauge gauge = registry.register(meta(name, MetricType.GAUGE), new JaegerGauge(), toTagArray(tags));
+        Tag[] tagArray = toTagArray(tags);
+        JaegerGauge gauge = (JaegerGauge) registry.getGauges().get(new MetricID(name, tagArray));
+        if (gauge == null) {
```
jmartisk, line 51: After https://github.com/quarkusio/quarkus/issues/7018 we automatically drop and re-create all metric registries during shutdown before a hot reload, so it should not happen anymore that the gauge is still registered. Are you sure we still need this? That would mean the registry dropping is wrong?!

Ladicek: I've seen an exception during hot reload on this. I'll try again.

Ladicek: No, this is still required. When I revert this change, I get this on hot reload:

```
org.jboss.resteasy.spi.UnhandledException: java.lang.RuntimeException: No reflection exceptions should be thrown unless there is a fundamental error in your code set up.
        at org.jboss.resteasy.core.ExceptionHandler.handleException(ExceptionHandler.java:381)
        at org.jboss.resteasy.core.SynchronousDispatcher.writeException(SynchronousDispatcher.java:216)
        at org.jboss.resteasy.core.SynchronousDispatcher.invoke(SynchronousDispatcher.java:515)
        at org.jboss.resteasy.core.SynchronousDispatcher.lambda$invoke$4(SynchronousDispatcher.java:259)
        at org.jboss.resteasy.core.SynchronousDispatcher.lambda$preprocess$0(SynchronousDispatcher.java:160)
        at org.jboss.resteasy.core.interception.jaxrs.PreMatchContainerRequestContext.filter(PreMatchContainerRequestContext.java:362)
        at org.jboss.resteasy.core.SynchronousDispatcher.preprocess(SynchronousDispatcher.java:163)
        at org.jboss.resteasy.core.SynchronousDispatcher.invoke(SynchronousDispatcher.java:245)
        at io.quarkus.resteasy.runtime.standalone.RequestDispatcher.service(RequestDispatcher.java:73)
        at io.quarkus.resteasy.runtime.standalone.VertxRequestHandler.dispatch(VertxRequestHandler.java:122)
        at io.quarkus.resteasy.runtime.standalone.VertxRequestHandler.access$000(VertxRequestHandler.java:36)
        at io.quarkus.resteasy.runtime.standalone.VertxRequestHandler$1.run(VertxRequestHandler.java:87)
        at io.quarkus.runtime.CleanableExecutor$CleaningRunnable.run(CleanableExecutor.java:231)
        at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
        at java.util.concurrent.FutureTask.run(FutureTask.java:266)
        at org.jboss.threads.ContextClassLoaderSavingRunnable.run(ContextClassLoaderSavingRunnable.java:35)
        at org.jboss.threads.EnhancedQueueExecutor.safeRun(EnhancedQueueExecutor.java:2027)
        at org.jboss.threads.EnhancedQueueExecutor$ThreadBody.doRunTask(EnhancedQueueExecutor.java:1551)
        at org.jboss.threads.EnhancedQueueExecutor$ThreadBody.run(EnhancedQueueExecutor.java:1442)
        at org.jboss.threads.DelegatingRunnable.run(DelegatingRunnable.java:29)
        at org.jboss.threads.ThreadLocalResettingRunnable.run(ThreadLocalResettingRunnable.java:29)
        at java.lang.Thread.run(Thread.java:748)
        at org.jboss.threads.JBossThread.run(JBossThread.java:479)
Caused by: java.lang.RuntimeException: No reflection exceptions should be thrown unless there is a fundamental error in your code set up.
        at io.jaegertracing.internal.metrics.Metrics.createMetrics(Metrics.java:76)
        at io.jaegertracing.internal.metrics.Metrics.<init>(Metrics.java:32)
        at io.jaegertracing.internal.metrics.Metrics.<init>(Metrics.java:28)
        at io.jaegertracing.Configuration.getTracerBuilder(Configuration.java:221)
        at io.quarkus.jaeger.runtime.QuarkusJaegerTracer.tracer(QuarkusJaegerTracer.java:40)
        at io.quarkus.jaeger.runtime.QuarkusJaegerTracer.buildSpan(QuarkusJaegerTracer.java:59)
        at io.opentracing.util.GlobalTracer.buildSpan(GlobalTracer.java:133)
        at io.quarkus.smallrye.opentracing.runtime.TracerProducer_ProducerMethod_tracer_96dadb3d6afa0cccadfe742c3e06ad433737c844_ClientProxy.buildSpan(TracerProducer_ProducerMethod_tracer_96dadb3d6afa0cccadfe742c3e06ad433737c844_ClientProxy.zig:238)
        at io.opentracing.contrib.jaxrs2.server.ServerTracingFilter.filter(ServerTracingFilter.java:69)
        at org.jboss.resteasy.core.interception.jaxrs.PreMatchContainerRequestContext.filter(PreMatchContainerRequestContext.java:310)
        at org.jboss.resteasy.core.ResourceMethodInvoker.invokeOnTarget(ResourceMethodInvoker.java:439)
        at org.jboss.resteasy.core.ResourceMethodInvoker.invoke(ResourceMethodInvoker.java:400)
        at org.jboss.resteasy.core.ResourceMethodInvoker.invoke(ResourceMethodInvoker.java:374)
        at org.jboss.resteasy.core.ResourceMethodInvoker.invoke(ResourceMethodInvoker.java:67)
        at org.jboss.resteasy.core.SynchronousDispatcher.invoke(SynchronousDispatcher.java:488)
        ... 20 more
Caused by: java.lang.IllegalArgumentException: A metric with metricID MetricID{name='jaeger_tracer_reporter_queue_length', tags=[]} already exists
        at io.smallrye.metrics.MetricsRegistryImpl.register(MetricsRegistryImpl.java:131)
        at io.quarkus.jaeger.runtime.QuarkusJaegerMetricsFactory.createGauge(QuarkusJaegerMetricsFactory.java:48)
        at io.jaegertracing.internal.metrics.Metrics.createMetrics(Metrics.java:66)
        ... 34 more
```

Ladicek: @jmartisk Any idea? Thanks! :-)

jmartisk: That's fishy, I'll try to reproduce this and let you know

jmartisk: @Ladicek I can't reproduce that - I created a simple app which uses tracing and metrics, called one traced method, triggered a reload, called that traced method again, the `jaeger_tracer_reporter_queue_length` gauge was re-registered correctly and no errors occurred.
I used Quarkus built from this PR branch, but with this one change reverted. Any other specific thing that I should try? :)

Ladicek: Oh well. I've just verified this still occurs, even if I rebase this PR on top of current `master`. I'll put together a detailed reproducer.

Ladicek: @jmartisk Here's a reproducer with detailed instructions: https://github.com/Ladicek/quarkus-jaeger-metrics-hot-reload-issue

jmartisk: @Ladicek  Ha! Thanks for the reproducer. I think I understand why this happens now and why it didn't happen with my reproducer: your application does **not** have a dependency on the `quarkus-smallrye-metrics` module, so the metrics processor doesn't run and doesn't register the task to clean up registries during reload.

But what now? Is it correct that Jaeger registers metrics even if the application does not declare a dependency on `quarkus-smallrye-metrics`? In other extensions, we generally try to make their dependency on metrics optional, so metrics are activated only if the user explicitly adds the  `quarkus-smallrye-metrics` dependency.

jmartisk: And if we can't make changes to the Jaeger extension such that metrics become optional (and we can run without them completely), then I suppose we should make the Jaeger module depend on `quarkus-smallrye-metrics-deployment` in order to make sure that the metrics processor is always applied and can do proper cleanup

jmartisk: It also happens that if the Metrics extension is not present (as in your reproducer), then metrics are not being exposed on the `/metrics` endpoint, so they are quite useless.

Ladicek: Now that is ... interesting. `quarkus-jaeger` depends on `smallrye-metrics` and `microprofile-metrics-api`, but not on `quarkus-smallrye-metrics`. That is ... wrong, for sure.

Ladicek: Or perhaps that was done to make metrics semi-optional...

Ladicek: @jmartisk What's the way to detect if metrics are enabled? I guess `capabilities.isCapabilityPresent(Capabilities.METRICS)`? If so, then we could decide on that and add either `NoopMetricsFactory` or `QuarkusJaegerMetricsFactory`. And probably make the dependencies on `smallrye-metrics` and `microprofile-metrics-api` `<scope>provided</scope>`, because they are only present to be able to compile `QuarkusJaegerMetricsFactory`. Does that sound about right? If so, I can check if that works and submit a PR.

jmartisk: @Ladicek yes, checking for the capability is ok.
I think the correct way is have an `optional` dependency on `quarkus-smallrye-metrics`, like Agroal does it:
https://github.com/quarkusio/quarkus/blob/master/extensions/agroal/runtime/pom.xml#L44

jmartisk: And what would be absolutely optimal and consistent with most other extensions, would be that the metrics from the extension are disabled by default (**even if** the metrics dependency is present) and have to be enabled by a config property `quarkus.jaeger.metrics.enabled`. See https://quarkus.io/guides/writing-extensions#extension-metrics

Even more optimal would be using the Metrics SPI (mentioned in the linked document) to register metrics eagerly, but if you feel that's too much for now, I guess we can leave that for later :)

Ladicek: Seriously? :open_mouth: :grin:

Ladicek: I mean, I would probably be able to do all that, but that would make this PR pretty large. I just wanted to fix the trace context propagation, found a dev mode issue, included a fix which was tiny, and now this :-/

Here's my proposal: I already have finished the conditional inclusion/exclusion of metrics based on the capability. I have also changed `quarkus-jaeger` to no longer depend on `smallrye-metrics` and `microprofile-metrics-api`; instead, it uses an `optional` dependency on `quarkus-smallrye-metrics` as you suggest. It seems to work (didn't try native :grin:). What's is _not_ done is addressing your 2nd comment, the one starting with "And what would be absolutely optimal" -- I'd leave that for future. Agree?

jmartisk: Sure, that's clearly not the goal of this PR. 
I created https://github.com/quarkusio/quarkus/issues/8502 for that

Ladicek: Nice, thanks! I'll update this PR in a few minutes.

Ladicek: PR rebased and updated per discussion above.

jmartisk (approved):
LGTM!

gsmet (approved):
Approving, per Jan.

PR #7991, opened 2020-03-19
Add note about class reading in class loading guide

sarxos:
Hello Quarkus Team!

I'm sending this PR due to the problem with class reading I experienced when migrating from 1.2.1.Final to 1.3.0.Final. After some digging I found this document, i.e. the [Class Loading Reference guide](https://quarkus.io/guides/class-loading-reference), but it didn't help much. I went back to the code to check how other extensions are doing class reading. I nailed my problem down to the following points:

- A `@BuildStep` which reads classes needs to have `loadsApplicationClasses` set to `true`.
- A `ClassLoader` should be taken from `Thread.currentThread().getContextClassLoader()`.

Adding parameter itself did not help. Using appropriate class loader was also important.

Before I migrated to 1.3.0 I haven't had these issues. My build step did not use `loadsApplicationClasses` parameter, I got `ClassLoader` from `getClass().getClassLoader()`, and everything worked just fine. 

I hope the change I made in the guide will help extension writers in the future.

This concludes this PR.

### DIFF
--- a/docs/src/main/asciidoc/class-loading-reference.adoc
+++ b/docs/src/main/asciidoc/class-loading-reference.adoc
@@ -19,6 +19,28 @@ a normal production Quarkus application.
 For all other use cases (e.g. tests, dev mode, and building the application) Quarkus
 uses the class loading architecture outlined here.
 
+== Reading Class Bytecode
+
+Whenever class bytecode needs to be read, a `@BuildStep` where class reading is done must have the `loadsApplicationClasses` parameter set to `true`. Otherwise the class will not be found.
+
+It's also important to use the correct `ClassLoader`. The recommended approach is to get it by calling the
+`Thread.currentThread().getContextClassLoader()` method.
+
+Example:
+
+
+[source,java,subs=attributes+]
+----
+@BuildStep(loadsApplicationClasses = true)
+GeneratedClassBuildItem instrument(final CombinedIndexBuildItem index) {
+    final String classname = "com.example.SomeClass";
+    final ClassLoader cl = Thread.currentThread().getContextClassLoader();
+    final byte[] originalBytecode = IoUtil.readClassAsBytes(cl, classname);
+    final byte[] enhancedBytecode = ... // class instrumentation from originalBytecode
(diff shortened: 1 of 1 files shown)


docs/src/main/asciidoc/class-loading-reference.adoc
```diff
@@ -19,6 +19,30 @@ a normal production Quarkus application.
 For all other use cases (e.g. tests, dev mode, and building the application) Quarkus
 uses the class loading architecture outlined here.
 
+== Reading Class Bytecode
+
+Whenever class bytecode needs to be read, a `@BuildStep` where class reading is done have
+to have `loadsApplicationClasses` parameter set to `true`. Otherwise the class will not be
+found.
```
gastaldi:
```suggestion
Whenever class bytecode needs to be read, a `@BuildStep` where class reading is done must have the `loadsApplicationClasses` parameter set to `true`. Otherwise the class will not be found.
```

docs/src/main/asciidoc/class-loading-reference.adoc
```diff
@@ -19,6 +19,30 @@ a normal production Quarkus application.
...
+
+It's also important to use correct `ClassLoader`. Best is to get it by calling the
```
gastaldi:
```suggestion
It's also important to use the correct `ClassLoader`. The recommended approach is to get it by calling the
```

docs/src/main/asciidoc/class-loading-reference.adoc
```diff
@@ -19,6 +19,30 @@ a normal production Quarkus application.
...
+`Thread.currentThread().getContextClassLoader()` method.
+
+Example `@BuildStep` may be similar to the following:
```
gastaldi:
```suggestion
Example:
```

gastaldi (commented):
Added some minor comments

stuartwdouglas (approved):

stuartwdouglas:
Can this be squashed into one commit?

gsmet:
I squashed the 2 commits. Let's merge! Thanks!

sarxos:
Thank you!

PR #7993, opened 2020-03-19
Bump jib version

geoand:


gsmet:
@geoand should we include it in dependabot?

Is this something we should backport?

geoand:
I am a little hesitent to to include it in dependabot because Jib core doesn't have a 1.x version yet so I am afraid of incompatibilities.

I do we think we should backport, since this jib version is strictly a bug fix version.

gastaldi (approved):

machi1990 (approved):

PR #8000, opened 2020-03-19
Fixed logging messages in the Gradle plugin

gastaldi:


geoand (approved):

gastaldi:
I added a new commit that would be really nice to backport, the first commit not that much

geoand:
I think you will need to do 2 different PRs if you want one and not the other to be backported, to make it easier for @gsmet

aloubyansky (approved):

gastaldi:
OTOH it's better to backport both commits since the log message is borked otherwise

PR #8005, opened 2020-03-20
Don't let the thread interrupt state interfere with class loading

stuartwdouglas:


geoand:
@evacchi would you be able to test this PR in your (hitherto) failing CI pipeline?

evacchi:
not sure how 🤔  @cristianonicolai

independent-projects/bootstrap/core/src/main/java/io/quarkus/bootstrap/classloading/QuarkusClassLoader.java
```diff
@@ -253,44 +253,55 @@ public InputStream getResourceAsStream(String nm) {
...
-            }
-            if (parentFirst) {
-                try {
-                    return parent.loadClass(name);
-                } catch (ClassNotFoundException ignore) {
-                    log.tracef("Class %s not found in parent first load from %s", name, parent);
+        //even if the thread is interrupted we still want to be able to load classes
+        //if the interrupt bit is set then we clear it and restore it at the end
+        boolean interrupted = Thread.interrupted();
+        try {
```
dmlloyd: In general I think this makes sense.  It's not perfect in that the thread can be interrupted *during* the class load, but I don't really have any better solution for that at bootstrap (`JBossThread` does have support for blocking interrupts but I'm assuming that it isn't available at bootstrap).

dmlloyd (approved):

independent-projects/bootstrap/core/src/main/java/io/quarkus/bootstrap/classloading/QuarkusClassLoader.java
stuartwdouglas: I guess we could catch InterruptedIOException and retry if it fails?

dmlloyd: Yeah that could work.  That exception has an `int` field which contains the number of bytes transferred thus far.  You'll have to clear the interrupt status manually though because `InterruptedIOException` (unlike `InterruptedException`) does not imply this.

evacchi:
just a heads up, we are observing this behavior on CI here https://github.com/kiegroup/kogito-examples/pull/149

but there is another weird thing (to me): a file is being read _at all_. The file that causes the error should be read from a jar in the dependencies (`kogito-runtimes` project dependency `org.kie.kogito:jbpm-flow`)

```
Unable to read /home/jenkins/workspace/IE_kogito_kogito-runtimes_PR-372/jbpm/jbpm-flow/target/classes/org/jbpm/workflow/instance/node/EndNodeInstance.class
```

instead it seems like it's picking it from a target folder (which _is_ being built on our CI too, but I still don't see why it should read it from there instead of the local .m2)

stuartwdouglas:
It reads from the project so that if you are just running mvn test and not mvn package or mvn install it will still run with the up to date version of depenedencies.

evacchi:
but how is it doing so? there is no parent pom pointing to `kogito-runtimes` in `kogito-examples` https://github.com/kiegroup/kogito-examples/blob/stable/pom.xml

gsmet:
I'm not sure about the status of this one. Please let me know when it's ready and I can merge/backport it.

Thanks!

stuartwdouglas:
This should be ready to go

PR #8006, opened 2020-03-20
Authentication fixes

### LINKED ISSUE #5284: Default realm is always returned with HTTP 401
**Describe the bug**
Andrey has reported that a default `Quarkus` realm is returned with 401 even when a property such as `quarkus.security.users.file.realm-name=MyRealm` is set.

**Expected behavior**
`WebAuthenticate` needs to have a custom realm name set.

### LINKED ISSUE #7768: Enabling both basic auth and form auth in quarkus does not work
**Description**
I tried enabling both basic and form authentication in quarkus 1.2.1.Final. According to the [documentation](https://quarkus.io/guides/all-config#quarkus-vertx-http_quarkus.http.auth.basic) when enabling both form and basic auth, basic auth should work in silent mode. As far as i understand this means that basic auth is used when an authorization header is sent.

**Expected behavior**
When enabling both form and basic auth and sending a GET request with valid Authorization header to a rest endpoint, basic auth should be used, the user should be authenticated successfully and the rest resource should be called.

**Actual behavior**
The request is forwarded to the form login page (default=login.html). It does not matter whether Authorization header is set and if credentials are valid. Form authentication will always be used instead of basic auth. When I disable form authentication and only enable basic auth, then basic auth works as expected.

**To Reproduce**
1. Start with the code.quarkus.io project setup enabling "Properties File based Security"
2. Add to the application.properties file:
```application.properties
quarkus.http.auth.form.enabled=true
quarkus.http.auth.basic=true

quarkus.security.users.file.enabled=true
quarkus.security.users.file.users=test-users.properties
quarkus.security.users.file.roles=test-roles.properties
quarkus.security.users.file.realm-name=MyRealm
quarkus.security.users.file.plain-text=true
```
Add test-users.properties with content:
```test-users.properties
admin=admin
```

and test-roles.properties with content:
```test-roles.properties
admin=adminrole
```

3. Add annotation @RolesAllowed("adminrole") to the example rest resource at /hello
4. Send a GET request to /hello with Authorization header for user admin, password admin and observe that you get a 302 response instead of a 200.

**Environment:**
 - Output of `uname -a` or `ver`: Linux N302 5.3.0-40-generic #32-Ubuntu SMP Fri Jan 31 20:24:34 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux
 - Output of `java -version`: openjdk version "1.8.0_242"
OpenJDK Runtime Environment (build 1.8.0_242-8u242-b08-0ubuntu3~19.10-b08)
OpenJDK 64-Bit Server VM (build 25.242-b08, mixed mode)
 - Quarkus version or git rev: 1.2.1.Final

### LINKED ISSUE #8138: Exception for multiple HTTP authentication mechanisms is never thrown.
**Describe the bug**
When multiple HTTP authentication mechanisms are present (not implemented yet) a IllegalStateException should be thrown but is not.

**Expected behavior**
The IllegalStateException to be thrown.

**Actual behavior**

**To Reproduce**
1. Create new project and add 'io.quarkus:quarkus-smallrye-jwt' and 'io.quarkus:quarkus-oidc'.
2. Set `quarkus.oidc.tenant-enabled=false` to avoid startup error. 
3. Quarkus starts up without IllegalStateException

**Configuration**
```properties
quarkus.oidc.tenant-enabled=false
```

**Additional context**
In HttpAuthenticator.java lift `if (instance.isAmbiguous())` out of ` if (instance.isResolvable())`.

```
@Inject
    public HttpAuthenticator(Instance<HttpAuthenticationMechanism> instance,
            Instance<IdentityProvider<UsernamePasswordAuthenticationRequest>> usernamePassword) {
        if (instance.isResolvable()) { 
            **if (instance.isAmbiguous()) {** // Not reachable due because isResolvable checks !isAmbiguous
                throw new IllegalStateException("Multiple HTTP authentication mechanisms are not implemented yet, discovered "
                        + instance.stream().collect(Collectors.toList()));
            }
            mechanism = instance.get();
        } else {
            if (!usernamePassword.isUnsatisfied()) {
                //TODO: config
                mechanism = new BasicAuthenticationMechanism("Quarkus");
            } else {
                mechanism = new NoAuthenticationMechanism();
            }
        }
    }
```

stuartwdouglas:
- Use synthetic beans to configure form and basic auth
- Allow multiple authentication mechanisms
- Better default behaviour based on what is configured

Also includes synthetic bean fixes from @mkouba 

Fixes #7768
Fixes #5284
Fixes #8138

### DIFF
--- a/extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/OidcAuthenticationMechanism.java
+++ b/extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/OidcAuthenticationMechanism.java
@@ -1,5 +1,7 @@
 package io.quarkus.oidc.runtime;
 
+import java.util.Collections;
+import java.util.Set;
 import java.util.concurrent.CompletionStage;
 
 import javax.enterprise.context.ApplicationScoped;
@@ -8,8 +10,11 @@ import javax.inject.Inject;
 import io.quarkus.oidc.OIDCException;
 import io.quarkus.security.identity.IdentityProviderManager;
 import io.quarkus.security.identity.SecurityIdentity;
+import io.quarkus.security.identity.request.AuthenticationRequest;
+import io.quarkus.security.identity.request.TokenAuthenticationRequest;
 import io.quarkus.vertx.http.runtime.security.ChallengeData;
 import io.quarkus.vertx.http.runtime.security.HttpAuthenticationMechanism;
+import io.quarkus.vertx.http.runtime.security.HttpCredentialTransport;
 import io.vertx.ext.web.RoutingContext;
 
 @ApplicationScoped
@@ -41,4 +46,15 @@ public class OidcAuthenticationMechanism implements HttpAuthenticationMechanism 
         return OidcTenantConfig.ApplicationType.WEB_APP == tenantContext.oidcConfig.applicationType;
     }
 
+    @Override
+    public Set<Class<? extends AuthenticationRequest>> getCredentialTypes() {
+        return Collections.singleton(TokenAuthenticationRequest.class);
+    }
+
+    @Override
+    public HttpCredentialTransport getCredentialTransport() {
+        //not 100% correct, but enough for now
+        //if OIDC is present we don't really want another bearer mechanism
+        return new HttpCredentialTransport(HttpCredentialTransport.Type.AUTHORIZATION, "bearer");
+    }
 }
--- a/extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/security/HttpAuthenticator.java
+++ b/extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/security/HttpAuthenticator.java
@@ -1,8 +1,14 @@
 package io.quarkus.vertx.http.runtime.security;
 
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
-import java.util.stream.Collectors;
+import java.util.function.Function;
 
 import javax.enterprise.context.ApplicationScoped;
 import javax.enterprise.inject.Instance;
@@ -12,7 +18,8 @@ import io.netty.handler.codec.http.HttpResponseStatus;
 import io.quarkus.security.identity.IdentityProvider;
 import io.quarkus.security.identity.IdentityProviderManager;
 import io.quarkus.security.identity.SecurityIdentity;
-import io.quarkus.security.identity.request.UsernamePasswordAuthenticationRequest;
+import io.quarkus.security.identity.request.AnonymousAuthenticationRequest;
+import io.quarkus.security.identity.request.AuthenticationRequest;
 import io.vertx.ext.web.RoutingContext;
 
 /**
@@ -24,33 +31,55 @@ public class HttpAuthenticator {
     @Inject
     IdentityProviderManager identityProviderManager;
 
-    final HttpAuthenticationMechanism mechanism;
+    final HttpAuthenticationMechanism[] mechanisms;
 
     public HttpAuthenticator() {
-        mechanism = null;
+        mechanisms = null;
     }
 
     @Inject
     public HttpAuthenticator(Instance<HttpAuthenticationMechanism> instance,
-            Instance<IdentityProvider<UsernamePasswordAuthenticationRequest>> usernamePassword) {
-        if (instance.isResolvable()) {
-            if (instance.isAmbiguous()) {
-                throw new IllegalStateException("Multiple HTTP authentication mechanisms are not implemented yet, discovered "
-                        + instance.stream().collect(Collectors.toList()));
+            Instance<IdentityProvider<?>> providers) {
+        List<HttpAuthenticationMechanism> mechanisms = new ArrayList<>();
+        for (HttpAuthenticationMechanism mechanism : instance) {
+            boolean notFound = false;
+            for (Class<? extends AuthenticationRequest> mechType : mechanism.getCredentialTypes()) {
+                boolean found = false;
+                for (IdentityProvider<?> i : providers) {
+                    if (i.getRequestType().equals(mechType)) {
+                        found = true;
+                        break;
+                    }
+                }
+                if (!found) {
+                    notFound = true;
+                    break;
+                }
             }
-            mechanism = instance.get();
-        } else {
-            if (!usernamePassword.isUnsatisfied()) {
-                //TODO: config
-                mechanism = new BasicAuthenticationMechanism("Quarkus");
-            } else {
-                mechanism = new NoAuthenticationMechanism();
+            if (!notFound) {
+                mechanisms.add(mechanism);
             }
         }
-    }
+        if (mechanisms.isEmpty()) {
+            this.mechanisms = new HttpAuthenticationMechanism[] { new NoAuthenticationMechanism() };
+        } else {
+            this.mechanisms = mechanisms.toArray(new HttpAuthenticationMechanism[mechanisms.size()]);
+            //validate that we don't have multiple incompatible mechanisms
+            Map<HttpCredentialTransport, HttpAuthenticationMechanism> map = new HashMap<>();
+            for (HttpAuthenticationMechanism i : mechanisms) {
+                HttpCredentialTransport credentialTransport = i.getCredentialTransport();
+                if (credentialTransport == null) {
+                    continue;
+                }
+                HttpAuthenticationMechanism existing = map.get(credentialTransport);
+                if (existing != null) {
+                    throw new RuntimeException("Multiple mechanisms present that use the same credential transport "
+                            + credentialTransport + ". Mechanisms are " + i + " and " + existing);
+                }
+                map.put(credentialTransport, i);
+            }
 
-    public HttpAuthenticator(HttpAuthenticationMechanism mechanism) {
-        this.mechanism = mechanism == null ? new NoAuthenticationMechanism() : mechanism;
+        }
     }
 
     /**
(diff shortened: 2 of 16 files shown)


sberyozkin:
Hi @stuartwdouglas What is going to happen now when `quarkus-oidc` and `quarkus-security-oauth2` and `quarkus-smallrye-jwt` are all included ? For example, what happens when `quarkus-oidc` detects an invalid token and fails the authentication ?

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/security/HttpAuthenticator.java
```diff
@@ -24,33 +29,40 @@
...
-                throw new IllegalStateException("Multiple HTTP authentication mechanisms are not implemented yet, discovered "
-                        + instance.stream().collect(Collectors.toList()));
+            Instance<IdentityProvider<?>> providers) {
+        List<HttpAuthenticationMechanism> mechanisms = new ArrayList<>();
+        for (HttpAuthenticationMechanism mechanism : instance) {
+            boolean notFound = false;
+            for (Class<? extends AuthenticationRequest> mechType : mechanism.getCredentialTypes()) {
+                boolean found = false;
+                for (IdentityProvider<?> i : providers) {
+                    if (i.getRequestType().equals(mechType)) {
```
sberyozkin: In case of those 3 extensions on the class path, can we get a case where `quarkus-oidc` will an identity provider with the other matching extensions ?
IMHO for a given credential type only a single authentication mechanism should be supported. So if the users add all 3 of those extensions plus some custom one for the same credential type then only one is selected which means they will need to choose what exactly is needed by adding only one of these extensions :-)

stuartwdouglas:
authenticate explicitly differentiates between invalid and missing. If you return null the next one is tried, if it fails then no further action is taken.

In general you should only mix mechanisms that use different credentials, e.g. form and basic.

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/security/HttpAuthenticator.java
stuartwdouglas: It's not the credential type that is the issue, it is the mechanism that is used to provide it. FORM and BASIC are both username/password, however there is no confusion as one is a POST and the other is a header. If two mechanims are both using Bearer auth then that is a problem.

sberyozkin:
@stuartwdouglas Sorry, I glanced and assumed the iteration would continue even if one of the Bearer authentication mechanism fails.

extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/security/HttpAuthenticator.java
sberyozkin: @stuartwdouglas how the unexpected side-effects can be prevented then ?
Here is one case: `quarkus-smallrye-jwt` is configured to accept tokens as the cookies, `quarkus-oidc` `web-app` application is getting an `IdToken`, but  `quarkus-smallrye-jwt` `MpJwtValidator` is picked up to validate this token, which can either be unexpectedly accepted or rejected (say due to the different expiry limits which apply to OIDC access and id tokens). This is a user error I guess but it feels we need to tighten something here:

 May be do something as basic as restricting a number of `HttpAuthenticationMechanism` per credential type ?

 Or should users who need such combinations provide a custom mechanism which would delegate to either Form or Basic ? May be we can just ship an extension providing this composite mechanism. Or introduce `CompositeHttpAuthenticationMechanism` which if injected, takes priority over everything else and will return a set of `HttpAuthenticationMechanism`s. So if the user needs a composition, they'd only need to provide:
```
@ApplicationScoped
class FormBasicHttpAuthMechanism {
    Set<HttpAuthenticationMechanism> getMechanisms() {
         // return as set
         new FormHttpAuthenticationMechanism()
         new BasicHttpAuthenticationMechanism()
    }
}
```

etc

gsmet:
@sberyozkin what's the status of this one? I can't infer it from your last comment.

sberyozkin:
@gsmet Sorry, I initially assumed that multiple mechanisms will be iterated over even if the current one returns a fault, so Stuart clarified it was not the case, which is good.
I have though another question for Stuart, in the other thread [above](https://github.com/quarkusio/quarkus/pull/8006#discussion_r396385169), it appears to me there could be a risk of one mechanism picking the identity provider from the other one.

gsmet:
@stuartwdouglas I let you check @sberyozkin 's comment and give me the status of this one?

If some of the changes are too risky, maybe it warrants extracting the safe fixes so that I can only backport them.

sberyozkin:
@gsmet @stuartwdouglas I've been thinking that in a way we can't prevent the users adding `quarkus-oidc`, `quarkus-smallrye-jwt` and `quarkus-elytron-security-oauth2` at the same time and perhaps I'm over dramatizing there with my comment :-), as the same kind of scenario is possible even with the current code if this mix happens.
IMHO it would indeed be better though to have a dedicated issue for supporting multiple authentication mechanisms at runtime, I'd rather let the users compose the mechanisms easily depending on their requirements (may be with form+basic being the only exception) as opposed to trying all the matching HttpAuthenticationMechanisms - as it does not give a guarantee this is what a user intended :-)

sberyozkin:
Or just mark `quarkus-oidc`, `quarkus-smallrye-jwt` and `quarkus-elytron-security-oauth2` mechanisms as mutually exclusive with some annotation ? @stuartwdouglas - that would avoid all the ambiguities with respect to those 3 extensions ? It feels some more thinking is needed around it :-)

stuartwdouglas:
They are currently mutually exclusive, in that they will fail if all of them are added. I think this is something we can improve later.

One option would be to have some way to express a credential transport type, and check compatibility that way, it's kinda hard to express in a generic way though. Stacking auth mechanisms is already an advanced use case, and I think it should be fairly obvious that you can't mix two that work the same way.

sberyozkin:
Hi Stuart, @stuartwdouglas, #8138 has just been opened, I haven't looked into it yet, but I think it is related

sberyozkin:
@stuartwdouglas My understanding is that a user expects an exception when more than one mechanism is present. It makes me think if the user-driven composition is the way to go - otherwise some users will be happy (form+basic case), and some - not

sberyozkin:
@stuartwdouglas Or just let the users set a property
`quarkus.security.support-multiple-mechanisms=true/false`, `false` (default ?)
If someone wants to have more than one, they just set that property to true (or vice versa)
That would be enough ?

gsmet:
Guys, I will need a conclusion for this one if you expect it to be part of 1.3.1.

sberyozkin:
IMHO it would be better to extract the authentication related updates and merge the rest, as it is obvious now different users have different expectations about what should happen when more than one mechanism is available. Though having a property for the users to choose if multiple mechanisms are OK or not seems like an easy way to resolve the ambiguity :-). Not sure if Stuart will agree

KeesKoffeman:
Well I stumbled upon #8138 while debugging why multiple mechanisms aren’t supported (my use case is multiple). Now it’s very blackbox if you add multiple and it just doesn’t work without any feedback. TLDR; I’d be very happy if the next release includes this PR!

sberyozkin:
@KeesKoffeman OK thanks, I thought you did not want that...
@stuartwdouglas What is your opinion about the property (qualified as needed) I suggested ? It would not change much in your PR except make it a bit more predictable. Example, if the property is disabled, then in case of #8138, a user would see "Please enable this property for the multiple authentication mechanisms to work" and once the user does it, all starts working.
If you are not convinced, then lets go ahead as this PR has other fixes. @gsmet Guillaume, please approve/merge once Stuart confirms.

stuartwdouglas:
@sberyozkin if that property is false the app just won't work at all. I don't see any reason to have a config flag that serves no purpose other than to stop the users application from working.

Basically there are two cases:
- There is only one mechanism, things work the same as they do today
- There are multiple mechanisms, today this will just break (well for form+basic it won't work properly, for any other combination the app just won't work). If you add this flag then this situation will continue (i.e. the app won't work), so I don't really see the point of a flag thats only purpose is to stop the app from booting.

stuartwdouglas:
I have added a credential transport notion to make sure that incompatible mechanisms are not enabled together.

gsmet:
@sberyozkin can you check that PR in priority?

sberyozkin:
Hi Stuart @stuartwdouglas 
Thanks for the earlier comment,

> so I don't really see the point of a flag thats only purpose is to stop the app from booting

I was just hoping to avoid some unpredictability. At the moment multiple mechanisms don't work, and now they will start working, and it may be worth to give an option to and prevent opening up the authentication space by accident.
Thanks for introducing this incompatibility notion, IMHO it would be better to start in a more restrictive mode and if it proves too restricting we can just drop that altogether :-)

extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/OidcAuthenticationMechanism.java
```diff
@@ -41,4 +46,15 @@ private boolean isWebApp(RoutingContext context) {
...
+    @Override
+    public Set<Class<? extends AuthenticationRequest>> getCredentialTypes() {
+        return Collections.singleton(TokenAuthenticationRequest.class);
+    }
+
+    @Override
+    public HttpCredentialTransport getCredentialTransport() {
+        //not 100% correct, but enough for now
+        //if OIDC is present we don't really want another bearer mechanism
+        return new HttpCredentialTransport(HttpCredentialTransport.Type.AUTHORIZATION, "bearer");
```
sberyozkin: That is fine, we can support `bearer,code-flow` in the future when a given mechanism supports different types

sberyozkin (approved):
@stuartwdouglas Thanks a lot, and thanks for the patience. The authentication part looks perfect now :-) The rest looks good for sure too

sberyozkin:
@gsmet HI Guillaume sorry for a delay. Hope everyone is happy with the latest update from Stuart, I certainly am :-).

PR #8014, opened 2020-03-20
Implement Vault Health Check integration

### LINKED ISSUE #7007: Add health check for Vault
**Description**

Add health check for Vault.

**Implementation ideas**

Create a readiness health check that checks if Vault is up and running.

(I can provide a PR as I have implemented for my project)

lordofthejars:
Fixes #7007

### DIFF
--- a/docs/src/main/asciidoc/vault.adoc
+++ b/docs/src/main/asciidoc/vault.adoc
@@ -497,6 +497,27 @@ You should see:
 {password=connor}
 ----
 
+== Vault Health Check
+
+If you are using the `quarkus-smallrye-health` extension, `quarkus-vault` can add a readiness health check
+to validate the connection to the Vault server. This is disabled by default.
+
+If enabled, when you access the `/health/ready` endpoint of your application you will have information about the connection validation status.
+
+This behavior can be enabled by setting the `quarkus.vault.health.enabled` property to `true` in your `application.properties`.
+
+Only if Vault is initialized, unsealed and active, the health endpoint returns that Vault is ready to serve requests.
+
+You can change a bit this behaviour by using `quarkus.vault.health.stand-by-ok` and `quarkus.vault.health.performance-stand-by-ok` to `true` in your `application.properties`.
+
+stand-by-ok:: Specifies if being a standby should still return the active status code instead of the standby status code.
+
+performance-stand-by-ok:: Specifies if being a performance standby should still return the active status code instead of the performance standby status code.
+
+You can inject `io.quarkus.vault.VaultSystemBackendEngine` to run system operations programmatically.
+
+WARNING: When the readiness probe is failing in Kubernetes, then the application is not reachable. This means that if Vault is failing, all services depending on Vault will become unreachable and maybe this is not the desired state, so use this flag according to your requirements.
+
 == TLS
 
 In production mode, TLS should be activated between the Quarkus application and Vault to prevent _man-in-the-middle_ attacks.
--- a/extensions/vault/runtime/src/main/java/io/quarkus/vault/VaultSystemBackendEngine.java
+++ b/extensions/vault/runtime/src/main/java/io/quarkus/vault/VaultSystemBackendEngine.java
@@ -1,0 +1,50 @@
+package io.quarkus.vault;
+
+import io.quarkus.vault.runtime.config.VaultRuntimeConfig;
+import io.quarkus.vault.sys.VaultHealth;
+import io.quarkus.vault.sys.VaultHealthStatus;
+import io.quarkus.vault.sys.VaultInit;
+import io.quarkus.vault.sys.VaultSealStatus;
+
+/**
+ * This service provides access to the system backend.
+ *
+ * @see VaultRuntimeConfig
+ */
+public interface VaultSystemBackendEngine {
+
+    /**
+     * Initializes a new Vault.
+     * 
+     * @param secretShares specifies the number of shares to split the master key into.
+     * @param secretThreshold specifies the number of shares required to reconstruct the master key.
+     * @return Vault Init.
+     */
+    VaultInit init(int secretShares, int secretThreshold);
+
+    /**
+     * Check the health status of Vault.
+     * Returns Vault health status code only by using HTTP HEAD requests.
+     * It is faster than calling {@link #healthStatus() healthStatus()} method which uses HTTP GET to return a complete
+     * VaultHealthStatus state.
+     *
+     * @return Vault Health Status.
+     */
+    VaultHealth health();
+
+    /**
+     * Check and return the health status of Vault.
+     * Returns a complete VaultHealthStatus state.
+     * This method uses HTTP GET to return a complete state.
+     *
+     * @return Complete Vault Health Status.
+     */
+    VaultHealthStatus healthStatus();
+
+    /**
+     * Check the seal status of a Vault.
+     *
+     * @return Vault Seal Status.
+     */
+    VaultSealStatus sealStatus();
+}
--- a/extensions/vault/runtime/src/main/java/io/quarkus/vault/runtime/VaultSystemBackendManager.java
+++ b/extensions/vault/runtime/src/main/java/io/quarkus/vault/runtime/VaultSystemBackendManager.java
@@ -1,0 +1,106 @@
+package io.quarkus.vault.runtime;
+
+import io.quarkus.vault.VaultSystemBackendEngine;
+import io.quarkus.vault.runtime.client.VaultClient;
+import io.quarkus.vault.runtime.client.dto.sys.VaultHealthResult;
+import io.quarkus.vault.runtime.client.dto.sys.VaultInitResponse;
+import io.quarkus.vault.runtime.client.dto.sys.VaultSealStatusResult;
+import io.quarkus.vault.runtime.config.VaultBuildTimeConfig;
+import io.quarkus.vault.sys.VaultHealth;
+import io.quarkus.vault.sys.VaultHealthStatus;
+import io.quarkus.vault.sys.VaultInit;
+import io.quarkus.vault.sys.VaultSealStatus;
+
+public class VaultSystemBackendManager implements VaultSystemBackendEngine {
+
+    private VaultClient vaultClient;
+    private VaultBuildTimeConfig buildTimeConfig;
+
+    public VaultSystemBackendManager(VaultBuildTimeConfig buildTimeConfig, VaultClient vaultClient) {
+        this.vaultClient = vaultClient;
+        this.buildTimeConfig = buildTimeConfig;
+    }
+
+    @Override
+    public VaultInit init(int secretShares, int secretThreshold) {
+        final VaultInitResponse init = this.vaultClient.init(secretShares, secretThreshold);
+
+        final VaultInit vaultInit = new VaultInit(init.keys, init.keysBase64, init.rootToken);
+        return vaultInit;
+    }
+
+    @Override
+    public VaultHealth health() {
+
+        boolean isStandByOk = false;
+        if (this.buildTimeConfig.health.standByOk) {
+            isStandByOk = true;
+        }
+
+        boolean isPerfStandByOk = false;
+        if (this.buildTimeConfig.health.performanceStandByOk) {
+            isPerfStandByOk = true;
+        }
+
+        return this.health(isStandByOk, isPerfStandByOk);
+    }
+
+    @Override
+    public VaultHealthStatus healthStatus() {
+        boolean isStandByOk = false;
+        if (this.buildTimeConfig.health.standByOk) {
+            isStandByOk = true;
+        }
+
+        boolean isPerfStandByOk = false;
+        if (this.buildTimeConfig.health.performanceStandByOk) {
+            isPerfStandByOk = true;
+        }
+
+        return this.healthStatus(isStandByOk, isPerfStandByOk);
+    }
+
+    @Override
+    public VaultSealStatus sealStatus() {
+        final VaultSealStatusResult vaultSealStatusResult = this.vaultClient.systemSealStatus();
+
+        final VaultSealStatus vaultSealStatus = new VaultSealStatus();
+        vaultSealStatus.setClusterId(vaultSealStatusResult.clusterId);
+        vaultSealStatus.setClusterName(vaultSealStatusResult.clusterName);
+        vaultSealStatus.setInitialized(vaultSealStatusResult.initialized);
+        vaultSealStatus.setMigration(vaultSealStatusResult.migration);
+        vaultSealStatus.setN(vaultSealStatusResult.n);
+        vaultSealStatus.setNonce(vaultSealStatusResult.nonce);
+        vaultSealStatus.setProgress(vaultSealStatusResult.progress);
+        vaultSealStatus.setRecoverySeal(vaultSealStatusResult.recoverySeal);
+        vaultSealStatus.setSealed(vaultSealStatusResult.sealed);
+        vaultSealStatus.setT(vaultSealStatusResult.t);
+        vaultSealStatus.setType(vaultSealStatusResult.type);
+        vaultSealStatus.setVersion(vaultSealStatusResult.version);
+
+        return vaultSealStatus;
+    }
+
+    private VaultHealthStatus healthStatus(boolean isStandByOk, boolean isPerfStandByOk) {
+        final VaultHealthResult vaultHealthResult = this.vaultClient.systemHealthStatus(isStandByOk, isPerfStandByOk);
+
+        final VaultHealthStatus vaultHealthStatus = new VaultHealthStatus();
+        vaultHealthStatus.setClusterId(vaultHealthResult.clusterId);
+        vaultHealthStatus.setClusterName(vaultHealthResult.clusterName);
+        vaultHealthStatus.setInitialized(vaultHealthResult.initialized);
+        vaultHealthStatus.setPerformanceStandby(vaultHealthResult.performanceStandby);
+        vaultHealthStatus.setReplicationDrMode(vaultHealthResult.replicationDrMode);
+        vaultHealthStatus.setReplicationPerfMode(vaultHealthResult.replicationPerfMode);
+        vaultHealthStatus.setSealed(vaultHealthResult.sealed);
+        vaultHealthStatus.setServerTimeUtc(vaultHealthResult.serverTimeUtc);
+        vaultHealthStatus.setStandby(vaultHealthResult.standby);
+        vaultHealthStatus.setVersion(vaultHealthResult.version);
+
+        return vaultHealthStatus;
+    }
+
+    private VaultHealth health(boolean isStandByOk, boolean isPerfStandByOk) {
+        final int systemHealthStatusCode = this.vaultClient.systemHealth(isStandByOk, isPerfStandByOk);
+        return new VaultHealth(systemHealthStatusCode);
+    }
+}
--- a/extensions/vault/runtime/src/main/java/io/quarkus/vault/runtime/client/OkHttpVaultClient.java
+++ b/extensions/vault/runtime/src/main/java/io/quarkus/vault/runtime/client/OkHttpVaultClient.java
@@ -6,6 +6,8 @@ import static io.quarkus.vault.runtime.client.OkHttpClientFactory.createHttpClie
 import java.io.IOException;
 import java.net.MalformedURLException;
 import java.net.URL;
+import java.util.Collections;
+import java.util.HashMap;
 import java.util.Map;
 
 import org.jboss.logging.Logger;
@@ -29,9 +31,13 @@ import io.quarkus.vault.runtime.client.dto.kv.VaultKvSecretV1;
 import io.quarkus.vault.runtime.client.dto.kv.VaultKvSecretV2;
 import io.quarkus.vault.runtime.client.dto.kv.VaultKvSecretV2Write;
 import io.quarkus.vault.runtime.client.dto.kv.VaultKvSecretV2WriteBody;
+import io.quarkus.vault.runtime.client.dto.sys.VaultHealthResult;
+import io.quarkus.vault.runtime.client.dto.sys.VaultInitBody;
+import io.quarkus.vault.runtime.client.dto.sys.VaultInitResponse;
 import io.quarkus.vault.runtime.client.dto.sys.VaultLeasesBody;
 import io.quarkus.vault.runtime.client.dto.sys.VaultLeasesLookup;
 import io.quarkus.vault.runtime.client.dto.sys.VaultRenewLease;
+import io.quarkus.vault.runtime.client.dto.sys.VaultSealStatusResult;
 import io.quarkus.vault.runtime.client.dto.totp.VaultTOTPCreateKeyBody;
 import io.quarkus.vault.runtime.client.dto.totp.VaultTOTPCreateKeyResult;
 import io.quarkus.vault.runtime.client.dto.totp.VaultTOTPGenerateCodeResult;
@@ -49,6 +55,7 @@ import io.quarkus.vault.runtime.client.dto.transit.VaultTransitSignBody;
 import io.quarkus.vault.runtime.client.dto.transit.VaultTransitVerify;
 import io.quarkus.vault.runtime.client.dto.transit.VaultTransitVerifyBody;
 import io.quarkus.vault.runtime.config.VaultRuntimeConfig;
+import okhttp3.HttpUrl;
 import okhttp3.MediaType;
 import okhttp3.OkHttpClient;
 import okhttp3.Request;
@@ -221,6 +228,30 @@ public class OkHttpVaultClient implements VaultClient {
         return post(path, token, body, VaultTOTPValidateCodeResult.class);
     }
 
+    @Override
+    public int systemHealth(boolean isStandByOk, boolean isPerfStandByOk) {
+        Map<String, String> queryParams = getHealthParams(isStandByOk, isPerfStandByOk);
+
+        return head("sys/health", queryParams);
+    }
+
+    @Override
+    public VaultHealthResult systemHealthStatus(boolean isStandByOk, boolean isPerfStandByOk) {
+        Map<String, String> queryParams = getHealthParams(isStandByOk, isPerfStandByOk);
+        return get("sys/health", queryParams, VaultHealthResult.class);
+    }
+
+    @Override
+    public VaultSealStatusResult systemSealStatus() {
+        return get("sys/seal-status", Collections.emptyMap(), VaultSealStatusResult.class);
+    }
+
+    @Override
+    public VaultInitResponse init(int secretShares, int secretThreshold) {
+        VaultInitBody body = new VaultInitBody(secretShares, secretThreshold);
+        return put("sys/init", body, VaultInitResponse.class);
+    }
+
     // ---
 
     protected <T> T list(String path, String token, Class<T> resultClass) {
@@ -253,9 +284,29 @@ public class OkHttpVaultClient implements VaultClient {
         return exec(request, resultClass);
     }
 
+    protected <T> T put(String path, Object body, Class<T> resultClass) {
+        Request request = builder(path).put(requestBody(body)).build();
+        return exec(request, resultClass);
+    }
+
     protected <T> T get(String path, String token, Class<T> resultClass) {
         Request request = builder(path, token).get().build();
         return exec(request, resultClass);
+    }
+
+    protected <T> T get(String path, Map<String, String> queryParams, Class<T> resultClass) {
+        final Request request = builder(path, queryParams).get().build();
+        return exec(request, resultClass);
+    }
+
+    protected int head(String path) {
+        final Request request = builder(path).head().build();
+        return exec(request);
+    }
+
+    protected int head(String path, Map<String, String> queryParams) {
+        final Request request = builder(path, queryParams).head().build();
+        return exec(request);
     }
 
     private <T> T exec(Request request, Class<T> resultClass) {
@@ -273,6 +324,14 @@ public class OkHttpVaultClient implements VaultClient {
             }
             String jsonBody = response.body().string();
             return resultClass == null ? null : mapper.readValue(jsonBody, resultClass);
+        } catch (IOException e) {
+            throw new VaultException(e);
+        }
+    }
+
+    private int exec(Request request) {
+        try (Response response = client.newCall(request).execute()) {
+            return response.code();
         } catch (IOException e) {
             throw new VaultException(e);
         }
@@ -296,6 +355,20 @@ public class OkHttpVaultClient implements VaultClient {
         return builder;
     }
 
+    private Request.Builder builder(String path) {
+        Request.Builder builder = new Request.Builder().url(getUrl(path));
+        return builder;
+    }
+
+    private Request.Builder builder(String path, Map<String, String> queryParams) {
+        HttpUrl.Builder httpBuilder = HttpUrl.parse(getUrl(path).toExternalForm()).newBuilder();
+        if (queryParams != null) {
+            queryParams.forEach((name, value) -> httpBuilder.addQueryParameter(name, value));
+        }
+        Request.Builder builder = new Request.Builder().url(httpBuilder.build());
+        return builder;
+    }
+
     private RequestBody requestBody(Object body) {
         try {
             return RequestBody.create(JSON, mapper.writeValueAsString(body));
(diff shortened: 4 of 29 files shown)


lordofthejars:
@sberyozkin @vsevel This is the new PR where I fixed all the Git history problem I did by error. I think that now we are fixing all the things we talked in the previous conversation and @gsmet can do its final review.

lordofthejars:
The build is canceled but it seems that everything is working.

sberyozkin:
Copying here the last comments from Vincent and Alex.
Vincent:
my main comment revolves around some exchanges we had earlier about this feature's applicability in #7007 (comment)
```
It is disabled by default and it is a good thing. I think we should stress in the documentation when to activate it, and/or the downsides of activating it.
```
People will have to realize that if they activate that feature, and vault goes rogue, the quarkus application will become not ready, will start not responding to external invocations, and most likely will trigger some monitoring alerts. If you have 200 microservices fetching secrets from vault, and vault gets down, all 200 microservices will turn red, whereas there is nothing wrong with those microservices. and it might take some time for ops to figure out that 200 incidents are just one.
To me when vault gets down, it should be red in the monitoring console, and all services depending on it should be orange to indicate that they are not running at 100% capacity.
So the vault healthcheck is really useful in situations where your microservice is dependent upon vault for 100% of its use cases (which I think is going to be rare). And still, making the service not ready is probably not going to make any difference for my clients since they are going to get a 503 instead of a 500.
That said I am seeing other healthcheck impl for external systems (neo4j, mongo, agroal - but not for ES). So may be this is the norm?

To wrap up on this issue I would add some doc saying:
```
    activate this healthcheck in situations where you do not want your application to be reachable if vault is not available (for instance in situations where all your microservice's features are dependent on vault).
    understand that if vault goes down, or switches to one of the statuses picked up by the healthcheck, all microservice replicas will become unreachable, as if they are not running.
```
Am I too dramatic? ;-)

other minor comments:
```
    there is a bit of copy/pasting between getBuildConfig() and getConfig() in VaultConfigSource
    I had already implemented the sys/health endpoint, but it was only used for testing purposes. so we located it into io.quarkus.vault.test.client.dto and used only from TestVaultClient, with other use cases such as sys/init, sys/seal-status, ... My version had a complete representation of the response, but is lacking the 2 optional arguments (standbyok and perfstandbyok). it would be nice to fully implement the sys/health api (including the other arguments such as activecode, standbycode, ...), plus remove my old code from TestVaultClient and use the new code in VaultTestExtension.initVault()
```

and Alex:

```
Ok, thins I am going to do:

Resolve conflicts
Update the sys API with the code that was developed for testing purposes
Update docs with a warning
Remove the Test client for doing sys calls and use the new one
```

lordofthejars:
@sberyozkin Thanks, I think everything is addressed now.

extensions/vault/runtime/src/main/java/io/quarkus/vault/runtime/sys/seal/VaultInit.java
```diff
@@ -0,0 +1,36 @@
...
+    public List<String> getKeysBase64() {
+        return keysBase64;
+    }
+
+    public String getRootToken() {
+        return rootToken;
+    }
+
+    @Override
+    public String toString() {
```
vsevel: it is too dangerous to include sensitive data into a `toString()`. it is way too easy for somebody to do a `log.info("init result = " + vaultInit)`. if we want to have a meaningful `toString()`, let's replace it with: 
```
return "VaultInit{"+keys.size()+" keys)"
```

extensions/vault/runtime/src/main/java/io/quarkus/vault/runtime/sys/seal/VaultInit.java
```diff
@@ -0,0 +1,36 @@
+package io.quarkus.vault.runtime.sys.seal;
```
vsevel: this is part of the api. it should be in package `io.quarkus.vault.sys`

extensions/vault/runtime/src/main/java/io/quarkus/vault/runtime/sys/seal/VaultSealStatus.java
```diff
@@ -0,0 +1,122 @@
+package io.quarkus.vault.runtime.sys.seal;
```
vsevel: this is part of the api. it should be in package `io.quarkus.vault.sys`

extensions/vault/runtime/src/main/java/io/quarkus/vault/runtime/sys/health/VaultHealth.java
```diff
@@ -0,0 +1,61 @@
+package io.quarkus.vault.runtime.sys.health;
```
vsevel: this is part of the api. it should be in package `io.quarkus.vault.sys`

extensions/vault/runtime/src/main/java/io/quarkus/vault/runtime/sys/health/VaultHealthStatus.java
```diff
@@ -0,0 +1,100 @@
+package io.quarkus.vault.runtime.sys.health;
```
vsevel: this is part of the api. it should be in package `io.quarkus.vault.sys`

extensions/vault/runtime/src/main/java/io/quarkus/vault/runtime/config/VaultConfigSource.java
```diff
@@ -148,16 +150,29 @@ private void fetchSecrets(List<String> paths, String prefix, Map<String, String>
...
         // init at most once
         if (init.compareAndSet(false, true)) {
-            VaultManager.init(serverConfig);
+            VaultManager.init(buildTimeConfig, serverConfig);
         }
 
         return VaultManager.getInstance();
     }
 
+    private VaultBuildTimeConfig getBuildConfig() {
```
vsevel:
```
    private VaultRuntimeConfig getRuntimeConfig() {
        return getConfig(runtimeConfig, () -> loadRuntimeConfig(), "runtime");
    }

    private VaultBuildimeConfig getBuildtimeConfig() {
        return getConfig(buildtimeConfig, () -> loadBuildtimeConfig(), "buildtime");
    }

    private <T> T getConfig(AtomicReference<T> ref, Supplier<T> supplier, String name) {
        T config = ref.get();
        if (config != null) {
            return config;
        } else {
            config = supplier.get();
            log.debug("loaded vault " + name + " config " + config);
            ref.set(config);
            return ref.get();
        }
    }
```

extensions/vault/runtime/src/main/java/io/quarkus/vault/runtime/VaultSystemBackendManager.java
```diff
@@ -0,0 +1,106 @@
...
+        vaultSealStatus.setRecoverySeal(vaultSealStatusResult.recoverySeal);
+        vaultSealStatus.setSealed(vaultSealStatusResult.sealed);
+        vaultSealStatus.setT(vaultSealStatusResult.t);
+        vaultSealStatus.setType(vaultSealStatusResult.type);
+        vaultSealStatus.setVersion(vaultSealStatusResult.version);
+
+        return vaultSealStatus;
+    }
+
+    private VaultHealthStatus healthStatus(boolean isStandByOk, boolean isPerfStandByOk) {
```
vsevel: why did you distinguished `VaultHealthStatus` from `VaultHealth`? we could have added the status code into the `VaultHealthStatus`. did you think that for performance reasons , in the context of healthchecks, it was important to do a `head` instead of a full `get`?

vsevel (changes requested):
hello, I have added a few more comments (and a question):
 - api packages
 - `toString` with sensitive information

extensions/vault/runtime/src/main/java/io/quarkus/vault/runtime/sys/health/VaultHealthStatus.java
lordofthejars: I was thinking the same but, there are a lot of sys operations so I decided to split but we can put together for sure.

extensions/vault/runtime/src/main/java/io/quarkus/vault/runtime/VaultSystemBackendManager.java
lordofthejars: Well two reasons, the first is that as you said head is more performant and in case of readiness checks it is important to be as performant as possible and the second is that Vault offers both options in the API. For this reason, I implemented both, so if someone wants to use programmatically is free to use it but in case of health checks use the more performant one.

extensions/vault/runtime/src/main/java/io/quarkus/vault/runtime/sys/health/VaultHealthStatus.java
vsevel: not sure I got you. everything that is either a param or a return type in the system backend engine interface needs to be in a non runtime package. that's the rule I got from Guillaume when I initially pushed the extension. this applies to:
 - VaultInit
 - VaultHealth
 - VaultHealthStatus 
 - VaultSealStatus
also I saw that you decided to not allow passing params to `VaultHealth health();` whereas the vault api allows to pass a bunch of args: standbyok, perfstandbyok, activecode, standbycode, drsecondarycode, performancestandbycode, sealedcode 
is that on purpose?

lordofthejars:
Ok then I move to sys package only

extensions/vault/runtime/src/main/java/io/quarkus/vault/runtime/sys/health/VaultHealthStatus.java
lordofthejars: Yes it is on purpose. As for me, it sounds strange that someone might want to change the status code definition and the status flags can be configured in `application.properties`. I was thinking more about simplicity/common cases and if someone requests it we can always extend it easily.

lordofthejars:
@vsevel Your comments are fixed.

lordofthejars:
Vault integration tests are passing, the failure on Java11 is on OIDC integration tests that runs after Vault tests.

vsevel:
hi @lordofthejars  I do not see the change in vault.adoc.did you miss it when bringing the changes from the old PR?

```
== Vault Health Check

If you are using the `quarkus-smallrye-health` extension, `quarkus-vault` can add a readiness health check
to validate the connection to the Vault server. This is disabled by default.

If enabled, when you access the `/health/ready` endpoint of your application you will have information about the connection validation status.

This behavior can be enabled by setting the `quarkus.vault.health.enabled` property to `true` in your `application.properties`.

Only if Vault is initialized, unsealed and active, the health endpoint returns that Vault is ready to serve requests.

You can change a bit this behaviour by using `quarkus.vault.health.standbyok` and `quarkus.vault.health.perfstandbyok` to `true` in your `application.properties`.

standbyok:: Specifies if being a standby should still return the active status code instead of the standby status code.

perfstandbyok:: Specifies if being a performance standby should still return the active status code instead of the performance standby status code.

You can inject `io.quarkus.vault.VaultSystemBackendEngine` to run system operations programmatically.

WARNING: When the readiness probe is failing in Kubernetes, then the application is not reachable. This means that if Vault is failing, all services depending on Vault will become unreachable and maybe this is not the desired state, so use this flag accordingly to your requirements.
```

extensions/vault/runtime/src/main/java/io/quarkus/vault/runtime/sys/health/VaultHealthStatus.java
vsevel: > As for me, it sounds strange that someone might want to change the status code definition

some time ago I had to integrate vault into k8s. the situation was different because I was implementing the liveness/readiness probes of vault itself (not a client of vault). but still it shows a situation where I had to change the status codes. basically I turned everything to 200 for liveness, otherwise vault would have been killed and restarted anytime its state would have bee something other than the ok state. which would have been counter productive because no restart would have helped. and for the readiness I did the same except for the sealedcode because when vault is sealed I did not want it to be part of the service. but I still wanted it to be reachable when it is not initialized. because if it is, all vaults are in the same situation, and making it not ready will just make it harder to access it.
see [the probe definitions.](https://github.com/vsevel/vault/blob/master/demoit/4-vault/vault-deployment.yaml#L64-L69)
you have also to take into account that people sometime use the same endpoint for liveness and readiness, which is a very bad idea when integrating external systems into the app health checks (this could lead to unwanted app restarts, and cascading crash loop backoffs).

are you expecting your healthchecks to be used for readiness only?

vsevel:
> Ok then I move to sys package only

the point was not to moving them from `io.quarkus.vault.runtime.sys.health` to `io.quarkus.vault.runtime.sys`, but to `io.quarkus.vault.sys.health` or `io.quarkus.vault.sys`.
since it is part of the api, it can't be in a `io.quarkus.vault.runtime` package.
so we did not understand each other.

extensions/vault/runtime/src/main/java/io/quarkus/vault/runtime/sys/health/VaultHealthStatus.java
lordofthejars: I understand but as you said this is different, we are not integrating Vault directly but the app which we are already taking care of the status code (200) so one might not need to tune this parameter.

By default yes, if people wants to use it as liveness they can implement by themselves using the programmatic way.

lordofthejars:
@vsevel Cool now I moved the classes to the correct package and updated the docs. I think that we are ready for merging and in the future update if it is necessary then let's add more specific flags. Now I think we are providing the same capabilities as other automatic Health Checks provided by Quarkus.

vsevel (approved):

vsevel:
great. thanks.

lordofthejars:
A pleasure, always a pleasure to meet another guy who is using Vault :)

extensions/vault/runtime/src/main/java/io/quarkus/vault/runtime/client/OkHttpVaultClient.java
```diff
@@ -312,4 +385,17 @@ private URL getUrl(String path) {
...
 
+    private Map<String, String> getHealthParams(boolean isStandByOk, boolean isPerfStandByOk) {
+        Map<String, String> queryParams = new HashMap<>();
+        if (isStandByOk) {
+            queryParams.put("standbyok", "true");
+        }
+
+        if (isPerfStandByOk) {
+            queryParams.put("perfstandbyok", "true");
+        }
```
machi1990:
```suggestion
        queryParams.put("standbyok", String.valueOf(isStandByOk));
        queryParams.put("perfstandbyok", String.valueOf(isPerfStandByOk));
```

extensions/vault/runtime/src/main/java/io/quarkus/vault/runtime/config/VaultConfigSource.java
```diff
@@ -148,30 +151,55 @@ private void fetchSecrets(List<String> paths, String prefix, Map<String, String>
...
     }
 
     // need to recode config loading since we are at the config source level
-    private VaultRuntimeConfig loadConfig() {
+    private VaultBuildTimeConfig loadBuildtimeConfig() {
+        VaultBuildTimeConfig vaultBuildTimeConfig = new VaultBuildTimeConfig();
+        vaultBuildTimeConfig.health = new HealthConfig();
+
+        vaultBuildTimeConfig.health.enabled = parseBoolean(
+                getVaultProperty("health.enable", "false"));
```
machi1990:
```suggestion
                getVaultProperty("health. enabled", "false"));
```

extensions/vault/runtime/src/main/java/io/quarkus/vault/runtime/client/OkHttpVaultClient.java
lordofthejars: I have not done in this way to not create objects + parse things when it is false. Also notice that this map is iterated in another piece of code, so a readiness check is called several times + usually user will leave these with defaults (`false`) I thought that it will be a minor improvement (that's true) but at least an improvement.

docs/src/main/asciidoc/vault.adoc
```diff
@@ -497,6 +497,27 @@ You should see:
...
+
+You can change a bit this behaviour by using `quarkus.vault.health.standbyok` and `quarkus.vault.health.perfstandbyok` to `true` in your `application.properties`.
+
+standbyok:: Specifies if being a standby should still return the active status code instead of the standby status code.
+
+perfstandbyok:: Specifies if being a performance standby should still return the active status code instead of the performance standby status code.
+
+You can inject `io.quarkus.vault.VaultSystemBackendEngine` to run system operations programmatically.
+
+WARNING: When the readiness probe is failing in Kubernetes, then the application is not reachable. This means that if Vault is failing, all services depending on Vault will become unreachable and maybe this is not the desired state, so use this flag accordingly to your requirements.
```
machi1990:
```suggestion
WARNING: When the readiness probe is failing in Kubernetes, then the application is not reachable. This means that if Vault is failing, all services depending on Vault will become unreachable and maybe this is not the desired state, so use this flag according to your requirements.
```

machi1990 (dismissed):
Added some minor suggestions.

extensions/vault/runtime/src/main/java/io/quarkus/vault/runtime/client/OkHttpVaultClient.java
machi1990: It was a minor nitpick. We can leave the code as is.

extensions/vault/runtime/src/main/java/io/quarkus/vault/runtime/config/VaultConfigSource.java
lordofthejars: Good catch.

lordofthejars:
@machi1990 Changes done.

sberyozkin:
Sorry, I see Vincent approved, it did look good to me the last tine I looked, I'll have a quick look tomorrow morning, so as not to do it in a rush :-), thanks

extensions/vault/runtime/src/main/java/io/quarkus/vault/VaultSystemBackendEngine.java
```diff
@@ -0,0 +1,45 @@
...
+    /**
+     * Initializes a new Vault.
+     * 
+     * @param secretShares specifies the number of shares to split the master key into.
+     * @param secretThreshold specifies the number of shares required to reconstruct the master key.
+     * @return Vault Init.
+     */
+    VaultInit init(int secretShares, int secretThreshold);
+
+    /**
```
sberyozkin: @lordofthejars Given what you've discussed with Vincent below, IMHO the java docs have to updated to clarify the difference between `health` and `healthStatus` methods and `VaultHealth` and `VaultHealthStatus` need to be have their names `exchanged` as it is quite confusing IMHO. 
Right now `healthStatus` returns `VaultHealthStatus` which is filled with the real information, while `health` returns `VaultHealth` which only has all that real information available in `VaultHealthStatus` encapsulated  in a status code :-).
Irrespectively of the way the lower level Vault API represents it all, it would be clearer IMHO if we had:
```
// Returns Vault health status code only by using HTTP HEAD requests.
// It is faster than a health() method which uses HTTP GET  to return a complete VaultHealth state
VaultHealthStatus healthStatus(); 

// Returns a complete VaultHealth state.
VaultHealth health(); 
```
Thanks

CC @vsevel

extensions/vault/runtime/src/main/java/io/quarkus/vault/VaultSystemBackendEngine.java
```diff
@@ -0,0 +1,45 @@
...
+public interface VaultSystemBackendEngine {
+
+    /**
+     * Initializes a new Vault.
+     * 
+     * @param secretShares specifies the number of shares to split the master key into.
+     * @param secretThreshold specifies the number of shares required to reconstruct the master key.
+     * @return Vault Init.
+     */
+    VaultInit init(int secretShares, int secretThreshold);
```
sberyozkin: We don't have `init` methods exposed anywhere else in the public Vault interfaces, this is an admin operation which should not be visible to the JAX-RS service code having `VaultSystemBackendEngine` injected

lordofthejars: We agreed with @vsevel to move the sys operations here.

extensions/vault/runtime/src/main/java/io/quarkus/vault/VaultSystemBackendEngine.java
lordofthejars: I'll update the JavaDoc.

extensions/vault/runtime/src/main/java/io/quarkus/vault/VaultSystemBackendEngine.java
sberyozkin: @lordofthejars OK, @vsevel Hi Vincent, do you feel it would be useful to have some users doing the Vault Initialization in scope of the current call ?

extensions/vault/runtime/src/main/java/io/quarkus/vault/VaultSystemBackendEngine.java
```diff
@@ -0,0 +1,50 @@
...
+
+    /**
+     * Check the health status of Vault.
+     * Returns Vault health status code only by using HTTP HEAD requests.
+     * It is faster than calling {@link #healthStatus() healthStatus()} method which uses HTTP GET to return a complete
+     * VaultHealthStatus state.
+     *
+     * @return Vault Health Status.
+     */
+    VaultHealth health();
```
sberyozkin: @lordofthejars So do you think that `VaultHealth` is more appropriate than `VaultHealthStatus` given that is only has a status code ? Wouldn;t `VaultHealthStatusCode` be more appropriate ?

lordofthejars:
Hi guys, I have no more time to invest in the PR sorry, I have the feeling that we are walking in a circle and any review will deal with another review and another review of things that has been there from the beginning. I was doing this in my spare time, but now I need this time for other things, sorry, feel free to fork my repo and do any fix.

Thanks.

extensions/vault/runtime/src/main/java/io/quarkus/vault/VaultSystemBackendEngine.java
sberyozkin: @vsevel I'm asking because we've tried hard to have all Vault interfaces very user centric in a sense that each of those operations can be do something meaningful in scope of a current HTTP request, and `init` does not fit. However if you do feel it is worth it then I'll be good with it, thanks

sberyozkin:
@lordofthejars I'm sorry to hear it. I'm not trying to make things difficult for you. I approach my own work the same way, and of other of my colleagues. We have worked a long time with Vincent too on his PRs. So I encourage you to stick to it and complete it. I'm definitely not going to block it if you won't respond. 
@gsmet Guillaume please merge it whenever you prefer, cheers

lordofthejars:
@sberyozkin @vsevel Init is used for testing purposes now, if you think we can remove I'll do it as it is really quick to do.

sberyozkin:
@lordofthejars Happy to see you back :-). I've actually removed myself as a reviewer to avoid interfering, but wanted to say that I've supported this enhancement from the get go which can be easily traced in the issue, in the closed PR too.
Re `init`- I know we've talked a lot about it with Vincent hence I asked. I'll now have no problems with whatever the final decision Vincent and yourself will make about it. 
Thanks all

lordofthejars:
@sberyozkin don't remove from reviewer as I said now I am getting out of time, so I need to be really performant on PRs, and I had the feeling we were just on circles, so removing the init method (if necessary) would be my last one.

sberyozkin:
@lordofthejars Sure, will be happy to complete the process. May be I'm indeed circling in our wrapped token discussion :-) but here I thought I only added a few comments :-). Our friend Vincent @vsevel can probably share some stories about my reviews all related to circling around, lol :-)

OK, I've resolved the comments to do with the bean names, the docs have helped to clarify.
re `init`:  I think Vincent used an internal `TestVaultClient` to do `init` in his tests.
IMHO having an `init` exposed to the user code is not ideal, but if both Vincent and yourself would prefer to keep it, I'll be good as per my earlier comment.

Please have a quick look at `TestVaultClient` and lets finalize tomorrow at the latest.
Cheers

extensions/vault/runtime/src/main/java/io/quarkus/vault/VaultSystemBackendEngine.java
vsevel: when I talked about moving some methods,I had specifically the `sys/health` in mind, but did not make it clear. I know @sberyozkin we had long discussion about this :) I think there is value in having those operations too on the `SystemBackendEngine`. and sometimes it confuses people to not find operations that we categorize as `admin` (although it is a valid criteria when priorizing what we want to work on first).
to make it short, now that @lordofthejars has moved all of them, I would not move them back.

lordofthejars:
@vsevel gotcha now sorry, well I agree with you that having them there is a good thing. Also if you check Spring Vault support you'll see that his Vault client just supports the full API, so every user is free to use any endpoint they require, so they just add support for Vault and the final user decides. So we are not totally unbiased providing `init` method too.

lordofthejars:
@gsmet I have squashed all the commits

sberyozkin:
@lordofthejars Just for the record, the fact that some other Vault API puts everything into API and thus overwhelming the users is not convincing IMHO, and we are not just following someone else's design decisions. My own approach, as a reviewer, was to to ensure that every method makes sense in scope of the JAX-RS applications. So far I haven't seen the users complaining. I'd rather have someone come up and say, where is `init` instead of putting everything into interfaces facing the users, As such letting someone do Vault init in the running application is not ideal at all IMHO, but I'll support Vincent's decision - may be someone will find it useful
Thanks

sberyozkin (approved):

lordofthejars:
@sberyozkin I understand your point and I tend to agree, but what I see is that any potential user might inspect our Vault code and then think that it is not mature at all as it only supports a minimal set of operations, so it is interesting to find the good balance.

vsevel:
hello @sberyozkin 

> So far I haven't seen the users complaining

I would not say complaining. but there could be some confusion. I was thinking about this https://github.com/quarkusio/quarkus/issues/7155#issuecomment-585555254 specifically

as a vault quarkus newbie I have to admit that I could be surprised too to see `sys/health` and not `sys/init`. and since anybody doing some IT will need to initialize a vault, I think providing them with the service seems reasonable. if not, they would have to create it in their own project, or depend on `quarkus-vault-test`, which is not documented anywhere, contains stuff that should not be exposed to user projects (so ideally we would need to split it in a `quarkus-vault-test-internal` and `quarkus-vault-test-for-user-projects`), and requires to use a low level `OkHttpVaultClient` (aka `TestVaultClient`). All in all a lot of restrictions with little gain. but like I said we discussed that at length some time ago, and I am ok either way, with some bias toward providing whatever is useful to the customer.

May be that is the notion of administrative services vs user services that we need to challenge. I would definitely use this as a criteria for priorization (assuming user services will provide the most value initially), but not use it as a hard line.

For the transit engine, we supported all services (encrypt, decrypt, ...) excluding:
 - the one that did not seem to make sense in quarkus. e.g. random number generator
 - the administrative tasks, such as creating an encryption key

but what if somebody said he really needed the `createEncryptionKey` either for:
 - a business use case => if it is valid, then the answer is obvious
 - write some tests => but what about this one? what should we say and do?
 - a technical use case where the quarkus app is used as a provisioning tool => and this one?

I would say:
 - start with business cases (ie. user type services)
 - stay open to other use cases (tests, provisioning use case)

In that respect the `init` service falls into the second category, which is the reason why I am fine on having it on the `SystemBackendEngine`. Long message for a short conclusion ;)

sberyozkin:
Hey Alex, @lordofthejars, if a user would just work out without even engaging with us then I'd not worry about losing such a user :-). But with `init` I personally can't explain a flow to any user who would ask when to use it given that the current JAX-RS application is up and supposedly Vault is up too :-). I'm not a Vault user but I can explain to anyone, in high level terms, how they can use `readSecrets` or `encrypt` etc - I'll tell them, get Vault started before starting your application, and then use `readSecrets` etc :-).
For example, we definitely not going to provide to OIDC users something like `Keycloak.init()` :-)

But Vault is special and perhaps with us opening up some of the sys operations some Vault power users will find it useful. This is how we discussed awhile back with Vincent: we can always add these ops into Vault interfaces when needed, and yeah, the time for `init` has come with this PR :-)

Thanks

sberyozkin:
Hi @vsevel Sure, in that case the user clarified and we added this method :-), there was a valid case there.
I support yourself and Alex with getting `init` in `SystemBackendEngine` - you are all right it is a dedicated interface and perhaps this is where the admin users can do some administrative tasks, perhaps as part of a management JAX-RS application, with UI letting the users to restart Vault, etc.
It sounds good
Cheers

sberyozkin:
@gsmet Hi Guillaume, it is good to go IMHO.

docs/src/main/asciidoc/vault.adoc
```diff
@@ -497,6 +497,27 @@ You should see:
...
+
+If enabled, when you access the `/health/ready` endpoint of your application you will have information about the connection validation status.
+
+This behavior can be enabled by setting the `quarkus.vault.health.enabled` property to `true` in your `application.properties`.
+
+Only if Vault is initialized, unsealed and active, the health endpoint returns that Vault is ready to serve requests.
+
+You can change a bit this behaviour by using `quarkus.vault.health.standbyok` and `quarkus.vault.health.perfstandbyok` to `true` in your `application.properties`.
+
+standbyok:: Specifies if being a standby should still return the active status code instead of the standby status code.
```
gsmet:
```suggestion
stand-by-ok:: Specifies if being a standby should still return the active status code instead of the standby status code.
```
would be more readable maybe?

docs/src/main/asciidoc/vault.adoc
```diff
@@ -497,6 +497,27 @@ You should see:
...
+
+perfstandbyok:: Specifies if being a performance standby should still return the active status code instead of the performance standby status code.
```
gsmet:
```suggestion
performance-stand-by-ok:: Specifies if being a performance standby should still return the active status code instead of the performance standby status code.
```
same here.

extensions/vault/runtime/src/main/java/io/quarkus/vault/health/VaultHealthCheck.java
```diff
@@ -0,0 +1,59 @@
...
+import org.eclipse.microprofile.health.HealthCheckResponse;
+import org.eclipse.microprofile.health.HealthCheckResponseBuilder;
+import org.eclipse.microprofile.health.Readiness;
+
+import io.quarkus.vault.VaultSystemBackendEngine;
+import io.quarkus.vault.sys.VaultHealth;
+
+@Readiness
+@ApplicationScoped
+public class VaultHealthCheck implements HealthCheck {
```
gsmet: This is not API, it should be in the `vault.runtime.health` package.

gsmet (dismissed):
I added some comments about the config.

We try to have readable config as much as possible.

Obviously, this needs to be applied everywhere in the patch.

docs/src/main/asciidoc/vault.adoc
lordofthejars: I understand your point and I thought the same but in Vault, the property is named in this way `standbyok` so  I decided to leave it in the same name as in Vault so anyone familiar with Vault properties finds the same property. Do you think it has sense or better to adopt to the Quarkus way?

docs/src/main/asciidoc/vault.adoc
lordofthejars: I understand your point and I thought the same but in Vault, the property is named in this way `perfstandbyok` so  I decided to leave it in the same name as in Vault so anyone familiar with Vault properties finds the same property. Do you think it has sense or better to adopt to the Quarkus way?

sberyozkin: @lordofthejars It should be fine IMHO, Quarkus users who know about Vault will associate `performance-stand-by-ok` with a lower level Vault `perfstandbyok` without any problems :-).

Lets tweak the properties across the PR and I believe Guillaume should be fine with the PR going in

sberyozkin: @lordofthejars FYI, I'm holding merging this PR, #8143, so as to not affect this PR (extra rebase)

lordofthejars:
@sberyozkin Done. Let's see if build passes. Then a Merge + Squash can be used to merge it and squash all these commits right?

sberyozkin:
@lordofthejars Thanks; AFAIK the manual squashing is needed, CC @gsmet

lordofthejars:
@sberyozkin Then I wait until the final approval and then I squash it manually.

docs/src/main/asciidoc/vault.adoc
sberyozkin: @lordofthejars Nearly there :-), can you please update `vault.adoc` too to match the renamed properties (s) ? Just press `Commit-Suggestion` or do it locally

sberyozkin:
@lordofthejars Please do [this](https://github.com/quarkusio/quarkus/pull/8014#discussion_r398541772) and I will be able to merge by lifting Guillaume's @gsmet change request because you will have addressed all of the comments

docs/src/main/asciidoc/vault.adoc
lordofthejars: Oh yes sorry I changed everything and I miss this

lordofthejars:
@sberyozkin ok pushed.

machi1990:
Merged. Thanks all and mostly to you @lordofthejars for the work and patience on this PR.

PR #8022, opened 2020-03-20
validate column names of sort properties

### LINKED ISSUE #1120: Panache - Sort properties injected in the query should be escaped somehow
Currently, they are not and we could imagine a user passing the sort property from the UI to the backend leading to a potential SQL injection.

glefloch:
close #1120 

Throws an `IllegalArgumenException` if column names used in `Sort` expression are not valid.

### DIFF
--- a/extensions/panache/panache-common/runtime/src/main/java/io/quarkus/panache/common/Sort.java
+++ b/extensions/panache/panache-common/runtime/src/main/java/io/quarkus/panache/common/Sort.java
@@ -2,7 +2,6 @@ package io.quarkus.panache.common;
 
 import java.util.ArrayList;
 import java.util.List;
-import java.util.regex.Pattern;
 
 /**
  * <p>
@@ -43,7 +42,6 @@ public class Sort {
 
     public class Column {
 
-        private final Pattern COLUMN_PATTERN = Pattern.compile("^[a-zA-Z_][a-zA-Z0-9_]*$");
         private String name;
         private Direction direction;
 
@@ -52,10 +50,7 @@ public class Sort {
         }
 
         public Column(String name, Direction direction) {
-            if (COLUMN_PATTERN.asPredicate().negate().test(name)) {
-                throw new IllegalArgumentException("Column name must match pattern: " + COLUMN_PATTERN.pattern());
-            }
-            this.name = name;
+            this.name = escape(name);
             this.direction = direction;
         }
 
@@ -73,6 +68,10 @@ public class Sort {
 
         public void setDirection(Direction direction) {
             this.direction = direction;
+        }
+
+        private String escape(String column) {
+            return "'" + column.replace("\\", "\\\\").replace("'", "\\'") + "'";
         }
     }
 
--- a/extensions/panache/panache-common/runtime/src/main/java/io/quarkus/panache/common/Sort.java
+++ b/extensions/panache/panache-common/runtime/src/main/java/io/quarkus/panache/common/Sort.java
@@ -2,6 +2,7 @@ package io.quarkus.panache.common;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.regex.Pattern;
 
 /**
  * <p>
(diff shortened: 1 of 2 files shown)


extensions/panache/panache-common/runtime/src/main/java/io/quarkus/panache/common/Sort.java
```diff
@@ -41,6 +42,8 @@
     }
 
     public class Column {
+
+        private final Pattern COLUMN_PATTERN = Pattern.compile("^[a-zA-Z_][a-zA-Z0-9_]*$");
```
gastaldi:
```suggestion
        private static final Pattern COLUMN_PATTERN = Pattern.compile("^[a-zA-Z_][a-zA-Z0-9_]*$");
```

extensions/panache/panache-common/runtime/src/main/java/io/quarkus/panache/common/Sort.java
```diff
@@ -49,6 +52,9 @@ public Column(String name) {
         }
 
         public Column(String name, Direction direction) {
+            if (COLUMN_PATTERN.asPredicate().negate().test(name)) {
```
gastaldi:
```suggestion
            if (!COLUMN_PATTERN.matcher(name).matches()) {
```

gastaldi:
I think the original issue also asks to escape (quote) the properties used in the `SORT BY` statement?

extensions/panache/panache-common/runtime/src/main/java/io/quarkus/panache/common/Sort.java
gsmet: Are we sure this is OK and covering all the possibilities? Where does this pattern come from?

gsmet: At the beginning, I was more thinking about escaping the column names. And the hard part is probably to get the dialect at some point to get it to properly escape things.

glefloch: @gsmet, as you said, this pattern may not be valid for all databases. 
Can we do a simple quote escape here ? Getting the dialect here seems complicated.

gsmet: While I agree it's probably be a bit difficult, this can cause all sorts of security issues so I would prefer we get this right.

@Sanne any advice on how we could get Hibernate ORM to properly escape this?

Sanne: This looks suspicious indeed. Why does it need to be validated, couldn't we pass the sort intruction to ORM directly?

Some background: in ORM 5.x there have been various attempts in the Dialets to automagically escape / veto column names, they had all to be rolled back as they would break in various ways - I don't remember the details but yes it's dangeour to impose such a rule without a thoutough analysis.

The proper solution is included in the design of ORM6, so I would suggest to just let the user be responsible for using the right names.

gsmet: The issue is that the sort column is directly appended to the generated HQL so you basically end up with an HQL injection if you're not cautious about it.

It somehow needs to be properly quoted/escaped/whatever makes sense.

Sanne: Ok, but is "HQL injection" actually a dangeous thing? We're not sending it to the DB unparsed.

gsmet: Hmmm, unparsed no, sure. But can't you completely change the meaning of the query just with carefully forged sort column?

Sanne: I suppose this could be explored - never seen it done. Also I just realized that these columns names are appended into HQL, didn't expect that

gsmet:
I'm closing this one as out of date but the issue remains.

PR #8031, opened 2020-03-20
Fix AmazonLambdaProcessor SkillStreamHandler subclass check

oztimpower:
Apologies I have to re-do #7985 due to an error in the commit. 

SkillStreamHandler to check to for known subclasses of SkillStreamHandler (fix). 

Plus, the methods() check does not find the requestHandler(), only directly implemented methods, so I have to force this with an additional check.

Output from execution on AWS:
```
END RequestId: f7b66f2a-7951-4da9-9860-36b82c417a93
REPORT RequestId: f7b66f2a-7951-4da9-9860-36b82c417a93  Duration: 4380.92 ms    Billed Duration: 4400 ms        Memory Size: 256 MB     Max Memory Used: 125 MB Init Duration: 1341.26 ms       
XRAY TraceId: 1-5e74edf6-cbd07c56fb45500a5179c402       SegmentId: 0d6d366a14777dff     Sampled: true   
{"version":"1.0","userAgent":"ask-java/2.28.0 Java/11.0.6 templateResolver","response":{"outputSpeech":{"type":"SSML","ssml":"<speak>Hi <alexa:name type=\"first\" \"/></speak>"}}}%   

```
/cc @gsmet

### DIFF
--- a/extensions/amazon-lambda/deployment/src/main/java/io/quarkus/amazon/lambda/deployment/AmazonLambdaProcessor.java
+++ b/extensions/amazon-lambda/deployment/src/main/java/io/quarkus/amazon/lambda/deployment/AmazonLambdaProcessor.java
@@ -76,7 +76,7 @@ public final class AmazonLambdaProcessor {
         allKnownImplementors.addAll(combinedIndexBuildItem.getIndex()
                 .getAllKnownImplementors(REQUEST_STREAM_HANDLER));
         allKnownImplementors.addAll(combinedIndexBuildItem.getIndex()
-                .getAllKnownImplementors(SKILL_STREAM_HANDLER));
+                .getAllKnownSubclasses(SKILL_STREAM_HANDLER));
 
         if (allKnownImplementors.size() > 0 && providedLambda.isPresent()) {
             throw new BuildException(
@@ -102,7 +102,7 @@ public final class AmazonLambdaProcessor {
 
             ClassInfo current = info;
             boolean done = false;
-            boolean streamHandler = false;
+            boolean streamHandler = info.superName().equals(SKILL_STREAM_HANDLER) ? true : false;
             while (current != null && !done) {
                 for (MethodInfo method : current.methods()) {
                     if (method.name().equals("handleRequest")) {


extensions/amazon-lambda/deployment/src/main/java/io/quarkus/amazon/lambda/deployment/AmazonLambdaProcessor.java
```diff
@@ -102,7 +102,7 @@ AdditionalApplicationArchiveMarkerBuildItem marker() {
 
             ClassInfo current = info;
             boolean done = false;
-            boolean streamHandler = false;
+            boolean streamHandler = info.superName().equals(SKILL_STREAM_HANDLER) ? true : false;
             while (current != null && !done) {
```
gsmet: Do we need to execute this if streamHandler is true?

extensions/amazon-lambda/deployment/src/main/java/io/quarkus/amazon/lambda/deployment/AmazonLambdaProcessor.java
```diff
@@ -102,7 +102,7 @@ AdditionalApplicationArchiveMarkerBuildItem marker() {
 
             ClassInfo current = info;
             boolean done = false;
-            boolean streamHandler = false;
+            boolean streamHandler = info.superName().equals(SKILL_STREAM_HANDLER) ? true : false;
```
gsmet: That won't work if you have an intermediary subclass.

gsmet: Ah OK I see we couldn't arrive here with an intermediary class.

oztimpower: Yes, in the while loop it doesn't enter the handleRequest() check, due to methods() only returning methods of the subclass.

```
            while (current != null && !done) {
                for (MethodInfo method : current.methods()) {
                    if (method.name().equals("handleRequest")) {

```

It does skip around the while loop a couple of times, though I though for readability I didn't want to add another condition to the while statement -> && !streamHandler

I debated doing this, and decided against in the end to minimise changes.
`            while (current != null && !done &&!streamHandler) {
`


I also though about doing it a more verbose way for readability and to be consistent on the place in the code where the decision is made:

```
            boolean streamHandler = false;
            while (current != null && !done) {
                for (MethodInfo method : current.methods()) {

                    if (current.superName().equals(SKILL_STREAM_HANDLER)) {
                        streamHandler = true;
                        done = true;
                        break;
                    }
                    
                    if (method.name().equals("handleRequest")) {
                        ...

```

extensions/amazon-lambda/deployment/src/main/java/io/quarkus/amazon/lambda/deployment/AmazonLambdaProcessor.java
oztimpower: Discussed below.

gsmet (approved):

gsmet:
I suppose it's good enough for now. We could refine later if someone complains.

gsmet:
Thanks!

PR #7992, opened 2020-03-19
Rename kubernetes doc to be more consistent

geoand:

PR #8040, opened 2020-03-21
Fix links following Kubernetes guide renaming

gsmet:
Follow up of #7992


machi1990 (approved):
Thanks for catching these.

PR #8043, opened 2020-03-21
Support Maven sha1 & changelist version properties

### LINKED ISSUE #7852: Support for "Maven CI Friendly Versions" properties sha1 and changelist
**Description**
In response to #7747, @aloubyansky added support for `${revision}` in the `version` element via PR #7786.
It was [discussed](https://github.com/quarkusio/quarkus/pull/7786#discussion_r391295590) that the two other ["Maven CI Friendly Versions"](https://maven.apache.org/maven-ci-friendly.html) properties `${sha1}` and `${changelist}` should also be supported.

This would/should then also allow the following scenario from the Maven documentation:
> grafik

PS: My personal use case would be a common version prefix in `pom.xml` that is used for both local development and on Jenkins but a flexible version suffix on Jenkins. Something like:
```xml
<version>${revision}${sha1}</version>
<properties>
    <revision>1.0</revision>
    <sha1>-SNAPSHOT</sha1>
</properties>
```
- local: no `-D`
- Jenkins: `-Dsha1=-build-123`

This way I would not need to duplicate the `1.0` in `Jenkinsfile` (or would not have to parse it somehow from the `pom.xml`).

famod:
Fixes #7852

cc @aloubyansky


independent-projects/bootstrap/core/src/main/java/io/quarkus/bootstrap/resolver/maven/workspace/LocalProject.java
```diff
@@ -30,8 +32,9 @@
 
     private static final String PROJECT_BASEDIR = "${project.basedir}";
     private static final String POM_XML = "pom.xml";
-    private static final String REVISION = "revision";
-    static final String REVISION_EXPR = "${revision}";
+
+    private static final Pattern UNRESOLVED_VERSION_PATTERN = Pattern
+            .compile(Pattern.quote("${") + "(revision|sha1|changelist)" + Pattern.quote("}"));
```
gastaldi: I believe you can use other property names as the version. Shouldn't we just test if the version matches `\$\{.*\}`?

famod: That's what I first thought, too. But that is _not_ the case as @aloubyansky pointed out here: https://github.com/quarkusio/quarkus/pull/7786#discussion_r391700587

See also the following bit from https://maven.apache.org/maven-ci-friendly.html:
grafik 

In fact I had a very strange problem when I used a custom property in another project and back then I did not know that custom properties are not allowed.

gastaldi: Interesting. Didn't know about that

gastaldi (approved):
LGTM

famod:
I suppose it comes too late and is not critical enough for 1.3.1, right?

gastaldi:
> I suppose it comes too late and is not critical enough for 1.3.1, right?

1.3.1.Final is coming on March 27th, but yeah, I don't think it's critical enough. 1.4.0.Final is just a few weeks later :wink:

aloubyansky:
Thanks @famod

independent-projects/bootstrap/core/src/main/java/io/quarkus/bootstrap/resolver/maven/workspace/LocalProject.java
famod: Yeah, me neither. I should have added a small comment to this field but I forgot about that...

gastaldi: There is still time, the PR wasn't merged yet :wink:

famod: Ok, will add a comment once the last check has finished (Windows JDK 11 JVM Tests).

PR #8055, opened 2020-03-21
Stop calling Jandex deprecated methods

gwenneg:
This is a follow-up of https://github.com/wildfly/jandex/pull/70


machi1990 (approved):
Thanks!

PR #8056, opened 2020-03-21
Replace obsolete OS-dependent compilation condition

gwenneg:
As discussed in https://github.com/oracle/graal/pull/2236, we're currently using an obsolete OS-dependent compilation condition that will be removed soon from GraalVM.

This PR fixes that problem by using the OS-dependent compilation condition recommended by the GraalVM team.

It was successfully tested with Windows 10 and GraalVM 20.0.0 (the condition was introduced because of a Windows issue).


gwenneg:
cc @dmlloyd

dmlloyd (approved):

PR #8065, opened 2020-03-22
fix(flyway): multiple datasource migration does not work on native

### LINKED ISSUE #7685: Flyway multiple datasource migration does not work on native
The below is the application.properties

it works on jdk 8.
not on graalvm 19.3.1
the error is 
```
org.flywaydb.core.api.flywayexception: found more than one migration with version 1.0.0
```

if i change the sql files with different versions.
then all the sql files will be executed for all the data sources

```
quarkus.datasource.db1.driver=com.mysql.cj.jdbc.Driver
quarkus.datasource.db1.url=jdbc:mysql://mysql:3306/db1
quarkus.datasource.db1.username=user
quarkus.datasource.db1.password=p@ssw0rd
quarkus.datasource.db1.min-size=1
quarkus.datasource.db1.max-size=10

quarkus.datasource.db2.driver=com.mysql.cj.jdbc.Driver
quarkus.datasource.db2.url=jdbc:mysql://mysql:3306/db2
quarkus.datasource.db2.username=user
quarkus.datasource.db2.password=p@ssw0rd
quarkus.datasource.db2.min-size=1
quarkus.datasource.db2.max-size=10


quarkus.flyway.migrate-at-start=false

quarkus.flyway.db1.schemas=MyTracker_PATIENTS
quarkus.flyway.db1.locations=db/db1
quarkus.flyway.db1.migrate-at-start=true

quarkus.flyway.db2.schemas=db2
quarkus.flyway.db2.locations=db/db2
quarkus.flyway.db2.migrate-at-start=true
```

machi1990:
The QuarkusPathLocationScanner returned all migrations locations by default, this risked to cause
collision when we have a same filename for a migration script which is handled by different
datasources - a behaviour encountered on the issue #7685. Let's avoid this case by making sure that
the scanner only returns migration that it can recognize.

Fixes #7685

/cc @yanxuehe this should fix the issue for you.

### DIFF
--- a/extensions/flyway/deployment/src/main/java/io/quarkus/flyway/FlywayProcessor.java
+++ b/extensions/flyway/deployment/src/main/java/io/quarkus/flyway/FlywayProcessor.java
@@ -52,9 +52,7 @@ class FlywayProcessor {
     private static final String FILE_APPLICATION_MIGRATIONS_PROTOCOL = "file";
 
     private static final Logger LOGGER = Logger.getLogger(FlywayProcessor.class);
-    /**
-     * Flyway build config
-     */
+
     FlywayBuildTimeConfig flywayBuildConfig;
 
     @BuildStep
@@ -90,15 +88,8 @@ class FlywayProcessor {
         containerListenerProducer.produce(new BeanContainerListenerBuildItem(recorder.setFlywayBuildConfig(flywayBuildConfig)));
     }
 
-    /**
-     * Handles all the operations that can be recorded in the RUNTIME_INIT execution time phase
-     *
-     * @param recorder Used to set the runtime config
-     * @param flywayRuntimeConfig The Flyway configuration
-     * @param jdbcDataSourceBuildItems Added this dependency to be sure that Agroal is initialized first
-     */
-    @Record(ExecutionTime.RUNTIME_INIT)
     @BuildStep
+    @Record(ExecutionTime.RUNTIME_INIT)
     ServiceStartBuildItem configureRuntimeProperties(FlywayRecorder recorder,
             FlywayRuntimeConfig flywayRuntimeConfig,
             BeanContainerBuildItem beanContainer,
@@ -117,10 +108,6 @@ class FlywayProcessor {
 
     /**
      * Collects the configured migration locations for the default and all named DataSources.
-     * <p>
-     * A {@link LinkedHashSet} is used to avoid duplications.
-     *
-     * @return {@link Collection} of {@link String}
      */
     private Collection<String> getMigrationLocations(Collection<String> dataSourceNames) {
         Collection<String> migrationLocations = dataSourceNames.stream()
--- a/extensions/flyway/deployment/src/main/java/io/quarkus/flyway/FlywayProcessor.java
+++ b/extensions/flyway/deployment/src/main/java/io/quarkus/flyway/FlywayProcessor.java
@@ -6,7 +6,6 @@ import java.io.IOException;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.net.URL;
-import java.nio.charset.StandardCharsets;
 import java.nio.file.FileSystem;
 import java.nio.file.FileSystems;
 import java.nio.file.Files;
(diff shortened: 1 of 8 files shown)


extensions/flyway/deployment/src/main/java/io/quarkus/flyway/FlywayProcessor.java
```diff
@@ -88,19 +82,20 @@ void build(BuildProducer<AdditionalBeanBuildItem> additionalBeanProducer,
...
+        containerListenerProducer.produce(new BeanContainerListenerBuildItem(recorder.setFlywayBuildConfig(flywayBuildConfig)));
     }
 
     /**
      * Handles all the operations that can be recorded in the RUNTIME_INIT execution time phase
      *
      * @param recorder Used to set the runtime config
      * @param flywayRuntimeConfig The Flyway configuration
-     * @param dataSourceInitializedBuildItem Added this dependency to be sure that Agroal is initialized first
+     * @param jdbcDataSourceBuildItems Added this dependency to be sure that Agroal is initialized first
```
gsmet: Could you clean up all these useless comments? I did it in the Liquibase extension that was based on it. It should be done in a separate commit.

extensions/flyway/deployment/src/main/java/io/quarkus/flyway/FlywayProcessor.java
```diff
@@ -88,19 +82,20 @@ void build(BuildProducer<AdditionalBeanBuildItem> additionalBeanProducer,
                 .collect(Collectors.toSet());
         new FlywayDatasourceBeanGenerator(dataSourceNames, generatedBeanBuildItem).createFlywayProducerBean();
 
-        registerNativeImageResources(resourceProducer, generatedResourceProducer,
-                discoverApplicationMigrations(getMigrationLocations(dataSourceNames)));
+        List<String> applicationMigrations = discoverApplicationMigrations(getMigrationLocations(dataSourceNames));
+        recorder.setApplicationMigrationFiles(applicationMigrations);
```
gsmet: It might be a silly idea but shouldn't we have a `Map<Datasource name, Migration files>`?

extensions/flyway/runtime/src/main/java/io/quarkus/flyway/runtime/graal/QuarkusPathLocationScanner.java
```diff
@@ -21,32 +15,46 @@
...
+
+    private final Collection<LoadableResource> scannedResources;
+
+    public QuarkusPathLocationScanner(Collection<Location> locations) {
+        this.scannedResources = new ArrayList<>();
+        ClassLoader classLoader = Thread.currentThread().getContextClassLoader();
+
+        for (String migrationFile : applicationMigrationFiles) {
+            if (canHandleMigrationFile(locations, migrationFile)) {
+                LOG.debug("Loading " + migrationFile);
```
gsmet: It was preexisiting bug we would need to use `debugf` here. I fixed a few in the Liquibase extension.

gsmet (changes requested):
I added a couple of suggestions, not all related to your change though.

extensions/flyway/deployment/src/main/java/io/quarkus/flyway/FlywayProcessor.java
machi1990: I see the idea but then we do not have the `Datasource name` during scanning so we can't exploit the Map as we'd have liked.

extensions/flyway/deployment/src/main/java/io/quarkus/flyway/FlywayProcessor.java
machi1990: Done.

extensions/flyway/runtime/src/main/java/io/quarkus/flyway/runtime/graal/QuarkusPathLocationScanner.java
machi1990: Done

extensions/flyway/deployment/src/main/java/io/quarkus/flyway/FlywayProcessor.java
```diff
@@ -117,10 +108,6 @@ ServiceStartBuildItem configureRuntimeProperties(FlywayRecorder recorder,
 
     /**
      * Collects the configured migration locations for the default and all named DataSources.
-     * <p>
-     * A {@link LinkedHashSet} is used to avoid duplications.
```
gastaldi: Dunno, this doesn't seem useless to me :)

gsmet: Well, I don't think we should add a comment every time we choose a data structure. Otherwise, we will have a loooot of comments.

If you use a Set, that's usually to avoid duplications :). I think just using it is enough documentation.

gsmet (approved):

gsmet:
Thanks @machi1990 !

gsmet:
@machi1990 btw, we have the same weird pattern of writing files in the Liquibase extension. Maybe we could get rid of it too?

machi1990:
> @machi1990 btw, we have the same weird pattern of writing files in the Liquibase extension. Maybe we could get rid of it too?

Sure, I'll look at it. Thanks for the review 👍

PR #8071, opened 2020-03-23
JPA named query with Hibernate with Panache

### LINKED ISSUE #3483: Support named queries in Panache
Named query invocations are missing in Panache. 

Adding named queries can make #3482 less of a priority, since hints can be directly defined in the annotated named query.

loicmathieu:
This PR adds support for JPA named queries inside Hibernate with Panache.

Fixes #3483

I only added tests for the `find()` method but it should works for `count()`, `delete()` and `update()` as the code is common to all query constructions.

### DIFF
--- a/docs/src/main/asciidoc/hibernate-orm-panache.adoc
+++ b/docs/src/main/asciidoc/hibernate-orm-panache.adoc
@@ -518,6 +518,25 @@ Order.find("select distinct o from Order o left join fetch o.lineItems");
 Order.update("update from Person set name = 'Mortal' where status = ?", Status.Alive);
 ----
 
+=== Named queries
+
+You can reference a named query instead of a (simplified) HQL query by prefixing its name with the '#' character.
+
+[source,java]
+----
+@Entity
+@NamedQuery(name = "Person.getByName", query = "from Person where name = :name")
+public class Person extends PanacheEntity {
+    public String name;
+    public LocalDate birth;
+    public Status status;
+
+    public static Person findByName(String name){
+        return find("#Person.getByName", name).firstResult();
+    }
+}
+----
+
 === Query parameters
 
 You can pass query parameters by index (1-based) as shown below:
--- a/docs/src/main/asciidoc/hibernate-orm-panache.adoc
+++ b/docs/src/main/asciidoc/hibernate-orm-panache.adoc
@@ -537,6 +537,12 @@ public class Person extends PanacheEntity {
 }
 ----
 
+[WARNING]
+====
+Named queries can only be defined inside your JPA entity classes (being the Panache entity class, or the repository parameterized type),
+or on one of its super classes.
+====
+
 === Query parameters
 
 You can pass query parameters by index (1-based) as shown below:
--- a/extensions/panache/hibernate-orm-panache/runtime/src/main/java/io/quarkus/hibernate/orm/panache/runtime/JpaOperations.java
+++ b/extensions/panache/hibernate-orm-panache/runtime/src/main/java/io/quarkus/hibernate/orm/panache/runtime/JpaOperations.java
@@ -265,7 +265,9 @@ public class JpaOperations {
         // FIXME: check for duplicate ORDER BY clause?
         Query jpaQuery;
         if (isNamedQuery(query)) {
-            jpaQuery = em.createNamedQuery(query.substring(1));
+            String namedQuery = query.substring(1);
+            NamedQueryUtil.checkNamedQuery(entityClass, namedQuery);
+            jpaQuery = em.createNamedQuery(namedQuery);
         } else {
             jpaQuery = em.createQuery(sort != null ? findQuery + toOrderBy(sort) : findQuery);
         }
@@ -284,7 +286,9 @@ public class JpaOperations {
         // FIXME: check for duplicate ORDER BY clause?
         Query jpaQuery;
         if (isNamedQuery(query)) {
-            jpaQuery = em.createNamedQuery(query.substring(1));
+            String namedQuery = query.substring(1);
+            NamedQueryUtil.checkNamedQuery(entityClass, namedQuery);
+            jpaQuery = em.createNamedQuery(namedQuery);
         } else {
             jpaQuery = em.createQuery(sort != null ? findQuery + toOrderBy(sort) : findQuery);
         }
--- a/extensions/panache/hibernate-orm-panache/runtime/src/main/java/io/quarkus/hibernate/orm/panache/runtime/JpaOperations.java
+++ b/extensions/panache/hibernate-orm-panache/runtime/src/main/java/io/quarkus/hibernate/orm/panache/runtime/JpaOperations.java
@@ -115,12 +115,19 @@ public class JpaOperations {
     }
 
     static String createFindQuery(Class<?> entityClass, String query, int paramCount) {
-        if (query == null)
+        if (query == null) {
             return "FROM " + getEntityName(entityClass);
+        }
 
         String trimmed = query.trim();
-        if (trimmed.isEmpty())
+        if (trimmed.isEmpty()) {
             return "FROM " + getEntityName(entityClass);
+        }
+
+        if (isNamedQuery(query)) {
+            // we return named query as is
+            return query;
+        }
 
         String trimmedLc = trimmed.toLowerCase();
         if (trimmedLc.startsWith("from ") || trimmedLc.startsWith("select ")) {
@@ -133,6 +140,13 @@ public class JpaOperations {
             query += " = ?1";
         }
         return "FROM " + getEntityName(entityClass) + " WHERE " + query;
+    }
+
+    static boolean isNamedQuery(String query) {
+        if (query == null || query.isEmpty()) {
+            return false;
+        }
+        return query.charAt(0) == '#';
     }
 
     private static String createCountQuery(Class<?> entityClass, String query, int paramCount) {
@@ -249,7 +263,12 @@ public class JpaOperations {
         String findQuery = createFindQuery(entityClass, query, paramCount(params));
         EntityManager em = getEntityManager();
         // FIXME: check for duplicate ORDER BY clause?
-        Query jpaQuery = em.createQuery(sort != null ? findQuery + toOrderBy(sort) : findQuery);
+        Query jpaQuery;
+        if (isNamedQuery(query)) {
+            jpaQuery = em.createNamedQuery(query.substring(1));
+        } else {
+            jpaQuery = em.createQuery(sort != null ? findQuery + toOrderBy(sort) : findQuery);
+        }
         bindParameters(jpaQuery, params);
         return new PanacheQueryImpl(em, jpaQuery, findQuery, params);
     }
@@ -263,7 +282,12 @@ public class JpaOperations {
         String findQuery = createFindQuery(entityClass, query, paramCount(params));
         EntityManager em = getEntityManager();
         // FIXME: check for duplicate ORDER BY clause?
-        Query jpaQuery = em.createQuery(sort != null ? findQuery + toOrderBy(sort) : findQuery);
+        Query jpaQuery;
+        if (isNamedQuery(query)) {
+            jpaQuery = em.createNamedQuery(query.substring(1));
+        } else {
+            jpaQuery = em.createQuery(sort != null ? findQuery + toOrderBy(sort) : findQuery);
+        }
         bindParameters(jpaQuery, params);
         return new PanacheQueryImpl(em, jpaQuery, findQuery, params);
     }
--- a/extensions/panache/hibernate-orm-panache/deployment/src/main/java/io/quarkus/hibernate/orm/panache/deployment/NamedQueryEntityClassBuildStep.java
+++ b/extensions/panache/hibernate-orm-panache/deployment/src/main/java/io/quarkus/hibernate/orm/panache/deployment/NamedQueryEntityClassBuildStep.java
@@ -1,0 +1,23 @@
+package io.quarkus.hibernate.orm.panache.deployment;
+
+import java.util.Set;
+
+import io.quarkus.builder.item.MultiBuildItem;
+
+final class NamedQueryEntityClassBuildStep extends MultiBuildItem {
+    private String className;
+    private Set<String> namedQueries;
+
+    public NamedQueryEntityClassBuildStep(String className, Set<String> namedQueries) {
+        this.className = className;
+        this.namedQueries = namedQueries;
+    }
+
+    public String getClassName() {
+        return this.className;
+    }
+
+    public Set<String> getNamedQueries() {
+        return namedQueries;
+    }
+}
--- a/extensions/panache/hibernate-orm-panache/deployment/src/main/java/io/quarkus/hibernate/orm/panache/deployment/PanacheResourceProcessor.java
+++ b/extensions/panache/hibernate-orm-panache/deployment/src/main/java/io/quarkus/hibernate/orm/panache/deployment/PanacheResourceProcessor.java
@@ -1,25 +1,36 @@
 package io.quarkus.hibernate.orm.panache.deployment;
 
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 
 import javax.persistence.EntityManager;
+import javax.persistence.NamedQueries;
+import javax.persistence.NamedQuery;
 
+import org.jboss.jandex.AnnotationInstance;
+import org.jboss.jandex.AnnotationValue;
 import org.jboss.jandex.ClassInfo;
 import org.jboss.jandex.DotName;
+import org.jboss.jandex.Type;
 
 import io.quarkus.arc.deployment.UnremovableBeanBuildItem;
 import io.quarkus.deployment.annotations.BuildProducer;
 import io.quarkus.deployment.annotations.BuildStep;
+import io.quarkus.deployment.annotations.ExecutionTime;
+import io.quarkus.deployment.annotations.Record;
 import io.quarkus.deployment.builditem.BytecodeTransformerBuildItem;
 import io.quarkus.deployment.builditem.CombinedIndexBuildItem;
 import io.quarkus.deployment.builditem.FeatureBuildItem;
+import io.quarkus.deployment.util.JandexUtil;
 import io.quarkus.hibernate.orm.deployment.AdditionalJpaModelBuildItem;
 import io.quarkus.hibernate.orm.deployment.HibernateEnhancersRegisteredBuildItem;
 import io.quarkus.hibernate.orm.panache.PanacheEntity;
 import io.quarkus.hibernate.orm.panache.PanacheEntityBase;
+import io.quarkus.hibernate.orm.panache.PanacheHibernateRecorder;
 import io.quarkus.hibernate.orm.panache.PanacheRepository;
 import io.quarkus.hibernate.orm.panache.PanacheRepositoryBase;
 import io.quarkus.panache.common.deployment.EntityField;
@@ -37,6 +48,10 @@ public final class PanacheResourceProcessor {
     private static final DotName DOTNAME_PANACHE_ENTITY = DotName.createSimple(PanacheEntity.class.getName());
 
     private static final DotName DOTNAME_ENTITY_MANAGER = DotName.createSimple(EntityManager.class.getName());
+
+    private static final DotName DOTNAME_NAMED_QUERY = DotName.createSimple(NamedQuery.class.getName());
+    private static final DotName DOTNAME_NAMED_QUERIES = DotName.createSimple(NamedQueries.class.getName());
+    private static final DotName DOTNAME_OBJECT = DotName.createSimple(Object.class.getName());
 
     @BuildStep
     FeatureBuildItem featureBuildItem() {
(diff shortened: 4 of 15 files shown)


FroMage:
Cool, I wanted this for a while. I do wonder if we _must_ prefix named queries. We could just collect all named queries at compile time and know that `findByName` is a named query, so we could refer to it with `Person.find("findByName", name)` uses that named query.

It would conflict with attribute names (if we have an attribute named `findByName`), but we could probably fix this with convention by telling people they must use named queries with a capital letter? Or all caps?

WDYT?

Probably need @Sanne and @emmanuelbernard for advice on ORM usage.

emmanuelbernard:
Careful, Stef and I thjink Loic assume that nbamed queries are contextual tot he entity. This is not the case.
Named queries in JPA are global and not contextual. So in JPA parleys `Person.find("findByName", name)` could refer to the `findByName` query returning `Country` if somewhere the `@NamedQuery` is defined as from `Country c where c.name = ?1`

We could
- embrace the global scope and keep it as is
- decide that by default named queries name on Panache Entity are to be prefixed automatically 
    - with which prefix?
- decide that named query calls from a panache entity, auto prefix (`find("findByName")` converted to `find("Person.findByName")`)

I am not sure I'm enamored by the auto prefix solution but could be convinced otherwise.

loicmathieu:
@emmanuelbernard 

> Named queries in JPA are global and not contextual. 

It's the developers's responsability to defines the named query and named it globally (that why it is usually prefixed by the name of the entity, but nothing enforce this). As named queries are defined using an annotation on the entity class by the developer I don't think we could (and should) do anything to prefix it.

My PR is a very simple one, a user can reference a named query when using the `find()` method by prefixing the query with `#`. If a user uses a named query from an other entity we cannot do anything about it and Hibernate will generates an exception.

emmanuelbernard:
So what don't you like about `#` @FroMage ? It gives a visual cue differentiator so I think it's a good idea.

FroMage:
Well, I'm thinking that it's unnecessary, but it's probably fine for a convention. And it avoids the ambiguity. So fine, let's go that way then.

loicmathieu:
Cool :)

@emmanuelbernard do you want us to do something to enforce that the named query should be defined on the entity (or the repository parameterized type) to avoid mis-use ? Or do we stick with the current implementation ? My feeling is that it's a developer's responsibility to avoid mis-use ... And we could do it later if we want ...

emmanuelbernard:
> @emmanuelbernard do you want us to do something to enforce that the named query should be defined on the entity (or the repository parameterized type) to avoid mis-use ? Or do we stick with the current implementation ? My feeling is that it's a developer's responsibility to avoid mis-use ... And we could do it later if we want ...

I don't think we should. It will make porting JPA apps harder.

loicmathieu:
@FromMage do you validate this PR? Is it ready to be merged ?

extensions/panache/hibernate-orm-panache/runtime/src/main/java/io/quarkus/hibernate/orm/panache/runtime/JpaOperations.java
```diff
@@ -135,6 +142,10 @@ static String createFindQuery(Class<?> entityClass, String query, int paramCount
         return "FROM " + getEntityName(entityClass) + " WHERE " + query;
     }
 
+    static boolean isNamedQuery(String query) {
+        return query.charAt(0) == '#';
```
FroMage: Given that we're testing for this before we check for `isEmpty` this can throw.

FroMage (changes requested):
Spotted one issue, otherwise LGTM.

Sanne:
One thing that has always bothered me is that a Named Query in Hibernate isn't actually parsed & validated at bootstrap: it's merely an API convenience, we never took advantage of the option to pre-parse in advance, but a lot more options will be available to us in Hibernate ORM 6.

So I wonder if we could have the "scoping" problem that @emmanuelbernard mentioned solved by simply not delegating this to the existing Named Query engine in ORM and then evolve this in multiple small steps:

1#
have this exposed just as a Panache convenience : store these queries in a Map or constants, pass them to ORM as normal Queries.
That means that at this stage they would behave like they do in Hibernate ORM: just syntactic sugar.

2#
perfom build-time JPQL validation of those named queries

3#
actually pre-parse the JPQL and generate the AST at build time during augmentation. This would be the most efficient solution, but I'd recommend waiting to implement this step to have upgraded to Hibernate ORM 6 - as this could would both be easier to do, and totally different anyways.

There's another reason for this: currently to allow parsing of any "freestyle" JPQL in native-image we need to include a lot of things; not just the parser but a lot of its components via reflection.
Would be great to be able to detect that all parsers are only necessary during augmentation, and then get rid of the overhead.

So I'd do only 1# for now, and hopefully that's not too complex? Just a Map and store them in there, you can also control the scope.

FroMage:
What's the gain of implementing 1# ourselves as opposed to keeping them as hibernate named queries as we do ATM? The behaviour would be the same from the user's PoV, no?

Sanne:
You could implement better scoping.

FroMage:
Ah, yeah, OK. But we'd have to duplicate later the efforts that ORM will be doing to pre-parse/validate them at build-time since we won't be able to implement them as `@NamedQuery` later because Hibernate ORM will not check scoping, will it?

Actually we can check scoping right now (for all users of Panache APIs) while still using `@NamedQuery`.

Sanne:
'm not suggesting you _do_ the optimisations of pre-parse & validate at this point.. Just consider that ORM6 will make it fairly easy, so we'll be able to tie into it.

I'd go for the better scoping now - it's an orthogonal problem and not something that "vanilla ORM" will be able to solve as we follow the JPA spec.

extensions/panache/hibernate-orm-panache/runtime/src/main/java/io/quarkus/hibernate/orm/panache/runtime/JpaOperations.java
loicmathieu: Fixed

extensions/panache/hibernate-orm-panache/runtime/src/main/java/io/quarkus/hibernate/orm/panache/runtime/JpaOperations.java
```diff
@@ -249,7 +260,12 @@ public static Object findById(Class<?> entityClass, Object id, LockModeType lock
         String findQuery = createFindQuery(entityClass, query, paramCount(params));
         EntityManager em = getEntityManager();
         // FIXME: check for duplicate ORDER BY clause?
-        Query jpaQuery = em.createQuery(sort != null ? findQuery + toOrderBy(sort) : findQuery);
+        Query jpaQuery;
+        if (isNamedQuery(query)) {
```
FroMage: Well, it can also happen here. You could have fixed `isNamedQuery` instead ;)

extensions/panache/hibernate-orm-panache/runtime/src/main/java/io/quarkus/hibernate/orm/panache/runtime/JpaOperations.java
```diff
@@ -263,7 +279,12 @@ public static Object findById(Class<?> entityClass, Object id, LockModeType lock
...
```
FroMage: And here.

FroMage (changes requested):

extensions/panache/hibernate-orm-panache/runtime/src/main/java/io/quarkus/hibernate/orm/panache/runtime/JpaOperations.java
loicmathieu: No, here this is the query computed by `createFindQuery` and it's never null (if null it will be `from EntityClass`.

extensions/panache/hibernate-orm-panache/runtime/src/main/java/io/quarkus/hibernate/orm/panache/runtime/JpaOperations.java
loicmathieu: same

extensions/panache/hibernate-orm-panache/runtime/src/main/java/io/quarkus/hibernate/orm/panache/runtime/JpaOperations.java
FroMage: it looks like the query created by `createFindQuery` is stored in `findQuery` and not in the `query` used, which is a parameter.

FroMage: Nope, this comes straight from the user.

extensions/panache/hibernate-orm-panache/runtime/src/main/java/io/quarkus/hibernate/orm/panache/runtime/JpaOperations.java
FroMage: Same here, it comes straight from the user. Now, I'll admit it looks like we're missing a null check too…

extensions/panache/hibernate-orm-panache/runtime/src/main/java/io/quarkus/hibernate/orm/panache/runtime/JpaOperations.java
loicmathieu: Damned it! I need to clean my glasses ;)
I will add a test with null and emtpy query so we didn't risk such bugs later

loicmathieu: @FroMage fixed.
I also add tests for empty and null query.

extensions/panache/hibernate-orm-panache/runtime/src/main/java/io/quarkus/hibernate/orm/panache/runtime/JpaOperations.java
loicmathieu: same, fixed

emmanuelbernard:
To be honest, I am not super happy with my scoping definition at it stands (especially because it is at odd with teh JPA spec (thus complicate things). So we could go with no scope and add natural overriding rules down the line based on conventions.

loicmathieu:
@Sanne you suggest to pass named Query to JPA as normal Query. Users can configure hints and locks on named queries, if we do this we will also need to send hints and locks to the entityManager and this becomes to be a lot of works with no direct benefits ... We can do it later when we have the facility to validate the query.

For scoping, I can think of this implementation :
- At build time, list all `@NamedQuery` of an entity and put them on a Map with the entity class as key
- At run time (when building the query), check that the name query is in the cache, if not LOG an error (so we tell the user he should not do this) or throw an exception.

WDYT ?

loicmathieu:
Still not convinced for scoping, will need at least to lookup for `@NamedQuery` on the hierarchy of types.

@emmanuelbernard just saw your comments after my proposal on scoping implementation. We can  decided to ship this functionality without it (I'll open a followup issue to not loose the discussion we had) or just LOG a warning (or info) for some time and decide later if we transform it to an exception.

emmanuelbernard:
Sure. But can you write down precisely the rules of when you would raise this warning / error? I'm not sure of what you have in mind loic.

Sanne:
> I am not super happy with my scoping definition at it stands (especially because it is at odd with teh JPA spec (thus complicate things). So we could go with no scope and add natural overriding rules down the line based on conventions

Agreed we need to avoid any confusion, but I think there is no big risk of cufusion mainly because people will get a clear exception stating t hat the query doesn't exist.

We can further mitigate any source of confusion by:

- not using the same annotation as JPA
- build-time validation

AFAIR @FroMage explained during one of our design meetings that it would be possible to recognize invocation points of a specific method, and validate the input as long as it was a constant.

We said that would be challenging for JPQL queries, as sometimes they are not constant and possibly built via some string appender / builder helper, but the Name of a named query should be a constant, or we could require it to be: that would be a reasonable limitation IMO, especially if it gets us build-time validation hooks.

@loicmathieu yes you'll have to copy over any additional flags and options too. I'd not try to "recycle" the `@NamedQuery` annotation as-is though - mainly to avoid confusion, but also you could take advantage of that to start small and expose more options later.

My _personal_ idea of ideal scoping is rather simple: limit the use of the defined queries to the same class they are defined in (so either the specific single entity, or the specific repository). I think that's how I've always used them - although in our teams we'd enforce such scoping by convention only.

I have no strong opinion about subclasses; again I'd propose for KISS and disallow, so you know exactly which type you can target: it's easy for the user to just expose the actual business query methods as public or protected as needed, they don't need direct access to the query definition in subclasses.

loicmathieu:
For scoping, I agree with @Sanne: if we do it, let's do it the scriter wa,y we can still relax it later if needed. So lookup on the named query only inside the entity class.

loicmathieu:
@emmanuelbernard do you want scoping on the first implementation ?

emmanuelbernard:
In the scoping rules I first drafted, I was not even convinced by myself. @loicmathieu what would be the rules you enforce? And would you create a separate @Query annotation?

emmanuelbernard:
Ah sorry so seeing Sanne's proposal it would be:

1. you must define your named query on the `PanacheEntity` for it to be used in a `find("#")` method.
2. what about repository pattern? Should the query be on the entity or ont he repository?
3. can I use `repo.find("#someQuery")` or do I have to encapsulate it `repo.findSomeQuery() { ... }`

I would imagine the rules are

1. put them on either the entity or the repo, for the repo both entity hosted and repo hosted would be accepted
2. you can use `repo.find("#...")`

Sure we can start with these rules. I'm a little chagrined by the fact that "legacy" named queries can't be executed with these but I can accept that to the hotel of less surprising results.

Note that this does not fix the query overlapping issue (two named queries with the same name).

FroMage:
@emmanuelbernard that behaviour seems the most intuitive yes.

What the "legacy" queries?

emmanuelbernard:
If I have imported a set of entities from an older JPA app, the named queries would not adhere to these rules necessarilya nd I would not be able to execute them

FroMage:
Well, if you imported from non-Panache ORM you would not have any calls to `MyEntity.find("#query")` anyway, so it will keep on working fine. It only kicks in once you start using the Panache API, in which case you can edit your entities and move the named queries around.

IMO the biggest issue might be for queries that belong between two entities if you want to use it in the two entities. But pfff.

Sanne:
> IMO the biggest issue might be for queries that belong between two entities if you want to use it in the two entities. But pfff.

I agree. It's always easy to make the method which actually refers to the named query public, then you can use it from wherever.

loicmathieu:
> I would imagine the rules are
> 
> 1. put them on either the entity or the repo, for the repo both entity hosted and repo hosted would be accepted
> 2. you can use repo.find("#...")

I'm OK with this and I will propose an implementation soon.

Regarding "legacy" named queries on old JPA entities, if you wan to use them with Panache, I guess you will use the repository pattern, so we are fine.

I found it strange to define a query on a repository class, I would restrict it to only the repository parameter type (the entity), in fact this is what the spec for named query ask (from the JavaDoc) : 

> The NamedQuery annotation can be applied to an entity or mapped superclass.

But reading the JavaDoc of the `entityManager.createNamedQuery()` method, it didn't specify any exception in case the named query is not define inside the entity: https://docs.oracle.com/javaee/7/api/javax/persistence/EntityManager.html#createNamedQuery-java.lang.String-

As it can also be applied to a mapped supeclass, I assume we need to search for named queries on the hierarchy of the entity class to be spec compliant.

loicmathieu:
I implement the scoping functionality as following:

- Named queries must be defined on the _JPA entity_ or one of it's parent classes.
- _JPA entity means Panache entity or the parameterized type of a Panache repository.
- I didn't include the Panache repository class as it didn't comply with the spec (see the JavaDoc of NamedQuery) and we didn't have a reference of this class inside `JpaOperations` so the implementation will be more complex.
- Named queries defined on a class not in the Jandex index are not included.

Please give me feedback, I'll squash the two commits when we agree on the implementation.

emmanuelbernard:
> I found it strange to define a query on a repository class, I would restrict it to only the repository parameter type (the entity), in fact this is what the spec for named query ask (from the JavaDoc) :
> 
> > The NamedQuery annotation can be applied to an entity or mapped superclass.
> 
> But reading the JavaDoc of the `entityManager.createNamedQuery()` method, it didn't specify any exception in case the named query is not define inside the entity: https://docs.oracle.com/javaee/7/api/javax/persistence/EntityManager.html#createNamedQuery-java.lang.String-
> 
> As it can also be applied to a mapped supeclass, I assume we need to search for named queries on the hierarchy of the entity class to be spec compliant.

The reason the JPA specs mentions this restriction is because it was an implementation limitation. Impl would only scan for `@Entity` classes and look up `@NamedQuery` on them. But here we also lookup repos. Personally I think the repo is a more natural place to put your query definitions. After all only there will you be able to use them. But this might not be picked up by Hibernate ORM proper so we can start with forcing named queries to be on entities for now. I would love the feature though

emmanuelbernard:
> I implement the scoping functionality as following:
> 
>     * Named queries must be defined on the **_JPA entity_** or one of it's parent classes.
> 
>     * **_JPA entity_** means Panache entity or the parameterized type of a Panache repository.
> 
>     * I didn't include the Panache repository class as it didn't comply with the spec (see the JavaDoc of NamedQuery) and we didn't have a reference of this class inside `JpaOperations` so the implementation will be more complex.
> 
>     * Named queries defined on a class not in the Jandex index are not included.

I  suppose that's the same limitaiton as a PanacheEntity? A PanacheEntity cannot be on a non jandexed class?

loicmathieu:
> I suppose that's the same limitaiton as a PanacheEntity? A PanacheEntity cannot be on a non jandexed class?

Yes, in fact it's the same limiation as JPA entities as Hibernate ORM lookup for them only on jandexed classes.

docs/src/main/asciidoc/hibernate-orm-panache.adoc
```diff
@@ -518,6 +518,31 @@ Order.find("select distinct o from Order o left join fetch o.lineItems");
...
+    public static Person findByName(String name){
+        return find("#Person.getByName", name).firstResult();
+    }
+}
+----
+
+[WARNING]
+====
+Named queries can only be defined inside your JPA entity classes (being the Panache entity class, or the repository parameterized type),
+or on one of it's super classes.
```
FroMage, line 543:
```suggestion
or on one of its super classes.
```

FroMage (changes requested):
LGTM, but could you add a test with an entity that defines two named queries? I'm unsure of the behaviour of jandex with annotation nesting.

loicmathieu:
>  I'm unsure of the behaviour of jandex with annotation nesting.

In fact when I repeat the named queries it is not working. I think I need to support `@NamedQueries` for this ...

loicmathieu:
@FroMage I added a test with two Named queries and add support for `@NamedQueries`.

FroMage (approved):
OK great, thanks!

nimo23:
>Personally I think the repo is a more natural place to put your query definitions. After all only there will you be able to use them. But this might not be picked up by Hibernate ORM proper so we can start with forcing named queries to be on entities for now. I would love the feature though

Good point! 

@loicmathieu I also want to put all my `@NamedQuery` and `@NamedEntityGraph` within the repository object itself instead of polluting my entity with it. For example:

```
@NamedEntityGraph(..)
@NamedQuery(..)
UserRepository{

}
```
I have a "repository object" and I would like to put my `@NamedQuery` and `@NamedEntityGraph` within the "repository object" instead of the "entity":

The "repository object" is responsible for CRUD-operations and the `@NamedQuery` is nothing more than a CRUD-operation, so the _more natural place_ to put those query definitions is indeed the "repository object" instead of the "entity".

I have thought that JPA does not care at which class the `@NamedQuery` or `@NamedEntityGraph`were placed into because on start JPA/Hibernate will scan all those annotations and registers them independent in which class they are actually put. Is this not the case with Quarkus?

FroMage:
> Personally I think the repo is a more natural place to put your query definitions. After all only there will you be able to use them. But this might not be picked up by Hibernate ORM proper so we can start with forcing named queries to be on entities for now

Only if you're using the repository pattern. If using the Active-Record pattern it's the only place that makes sense ;)

nimo23:
> Only if you're using the repository pattern. If using the Active-Record pattern it's the only place that makes sense ;)

Yes, of course. In the former, I also used the Active-Record pattern but I decided to go with the Repository Pattern because:
- code within the entity is grown after time 
- maintenance can be easier with repository pattern
- I wanted to **decouple** database operations from my entity and make CRUD operations more generic (for example, `User.class` can be stored by JPA/SQL with `UserJpaRepository.class` and can also be stored with `UserInfinispanRepository.class`).

nimo23:
However, in general, putting `@NamedQuery` outside of its entity will not work (I tried it also with Wildfly). So I think it would be better to leave it as it is..maybe in future JPA will support it..

loicmathieu:
@FroMage can we merge this one so it could be inside 1.4 ?

FroMage:
Sure.

PR #8082, opened 2020-03-23
fix security checks to make @RolesAllowed("**") work

### LINKED ISSUE #7994: @Authenticated works, but @RolesAllowed("*") doesn't
**Describe the bug**
I'm used to writing `@RolesAllowed("*")` when I want to let all authenticated users have access. (That is, all authenticated users are authorized to that particular method/class.)

The documentation (https://quarkus.io/guides/security#standard-security-annotations) specifically says:

> Quarkus also provides the `io.quarkus.security.Authenticated` annotation that will permit any authenticated user to access the resource (equivalent to `@RolesAllowed("*")`).

I interpret that as: `@Authenticated` is a quality of life improvement, but I can use both `@Authenticated` and `@RolesAllowed("*")` and they have the same meaning.

The javadoc of `@Authenticated` says:

> This is equivalent to the Jakarta EE8 {@code RolesAllowed("**")} construct.

OK, that's says `"**"` instead of `"*"`, but I'd interpret that as, again, I can use both interchangeably.

However, neither `@RolesAllowed("*")` nor `@RolesAllowed("**")` actually work. That is, authenticated users are _not_ authorized. The only way that actually works is `@Authenticated`.

**Expected behavior**
`@RolesAllowed("*")` works.

**Actual behavior**
`@RolesAllowed("*")` doesn't work, I have to use `@Authenticated`.

**To Reproduce**
1. Add a JAX-RS resource, such as:
```java
@Path("/everyone")
@RolesAllowed("*")
public class EveryoneResource {
    @GET
    public String get(@Context SecurityContext security) {
        return "Hello, " + security.getUserPrincipal().getName();
    }
}
```
2. Set up HTTP auth (see below for config file).
3. Perform an authenticated request against `/everyone` and observe authorization failure.

**Configuration**
```properties
quarkus.http.auth.basic=true

quarkus.security.users.embedded.enabled=true
quarkus.security.users.embedded.plain-text=true
quarkus.security.users.embedded.users.albert=E!nst3iN
quarkus.security.users.embedded.roles.albert=admin
```

**Additional context**
Per the EJB spec (chapter 12.2.5.2 in EJB 3.2 Core spec), `**` means all roles.

Per the Servlet spec (chapter 13.3 in Servlet 4.0 spec), `**` means all roles and `*` should not be used (`HttpServletRequest.isUserInRole("*")` must always return `false`).

The JAX-RS specification doesn't seem to define this, but the `QuarkusResteasySecurityContext.isUserInRole` method already checks for `**` and returns `true` for all authenticated users.

On the other hand, SmallRye JWT uses `*` to mean "all roles" when processing `@RolesAllowed`.

Quarkus documentation is ambiguous in this regard, as explained above.

I personally would prefer if we could just use `*`, but it seems we really should use `**` (and fix the documentation and SmallRye JWT).

Ladicek:
The correct wildcard to use seems to be `**`, because that's what
the Jakarta EE specifications use. More specifically:

- the EJB spec says that `**` means all roles
  (chapter 12.2.5.2 in EJB 3.2 Core spec)
- the Servlet spec says that `**` means all roles and `*` should
  never be used (chapter 13.3 in Servlet 4.0 spec)
- the JAX-RS specification doesn't seem to define this, but
  `QuarkusResteasySecurityContext.isUserInRole` already uses `**`

Resolves #7994.

### DIFF
--- a/extensions/security/runtime/src/main/java/io/quarkus/security/runtime/interceptor/check/RolesAllowedCheck.java
+++ b/extensions/security/runtime/src/main/java/io/quarkus/security/runtime/interceptor/check/RolesAllowedCheck.java
@@ -54,7 +54,7 @@ public class RolesAllowedCheck implements SecurityCheck {
         Set<String> roles = identity.getRoles();
         if (roles != null) {
             for (String role : allowedRoles) {
-                if (roles.contains(role)) {
+                if (roles.contains(role) || ("**".equals(role) && !identity.isAnonymous())) {
                     return;
                 }
             }
--- a/docs/src/main/asciidoc/security.adoc
+++ b/docs/src/main/asciidoc/security.adoc
@@ -78,7 +78,7 @@ Quarkus comes with built-in security to allow for Role-Based Access Control (lin
 based on the common security annotations `@RolesAllowed`, `@DenyAll`, `@PermitAll` on REST endpoints and CDI beans.
 An example of an endpoint that makes use of both JAX-RS and Common Security annotations to describe and secure its endpoints is given in <<subject-example>>. Quarkus also provides
 the `io.quarkus.security.Authenticated` annotation that will permit any authenticated user to access the resource
-(equivalent to `@RolesAllowed("*")`).
+(equivalent to `@RolesAllowed("**")`).
 
 [#subject-example]
 .SubjectExposingResource Example
--- a/integration-tests/main/src/main/java/io/quarkus/it/rest/RBACBean.java
+++ b/integration-tests/main/src/main/java/io/quarkus/it/rest/RBACBean.java
@@ -18,6 +18,11 @@ public class RBACBean {
         return "callingAuthenticated";
     }
 
+    @RolesAllowed("**")
+    public String allRoles() {
+        return "callingAllRoles";
+    }
+
     @PermitAll
     public String permitted() {
         return "callingPermitted";
--- a/integration-tests/main/src/main/java/io/quarkus/it/rest/RBACSecuredResource.java
+++ b/integration-tests/main/src/main/java/io/quarkus/it/rest/RBACSecuredResource.java
@@ -59,9 +59,22 @@ public class RBACSecuredResource {
     }
 
     @GET
+    @RolesAllowed("**")
+    @Path("allRoles")
+    public String allRoles() {
+        return "allRoles";
+    }
+
+    @GET
     @Path("callingAuthenticated")
     public String callingAuthenticated() {
         return bean.authenticated();
+    }
+
+    @GET
+    @Path("callingAllRoles")
+    public String callingAllRoles() {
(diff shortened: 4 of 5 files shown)


gastaldi (approved):

extensions/security/runtime/src/main/java/io/quarkus/security/runtime/interceptor/check/RolesAllowedCheck.java
```diff
@@ -54,7 +54,7 @@ public void apply(SecurityIdentity identity, Method method, Object[] parameters)
         Set<String> roles = identity.getRoles();
         if (roles != null) {
             for (String role : allowedRoles) {
-                if (roles.contains(role)) {
+                if (roles.contains(role) || ("**".equals(role) && !identity.isAnonymous())) {
```
gsmet: So do we support `*`?

If it's what we documented before, we should probably support it.

gsmet (dismissed):
Asked a question.

extensions/security/runtime/src/main/java/io/quarkus/security/runtime/interceptor/check/RolesAllowedCheck.java
Ladicek: It was in the documentation, but didn't work. `**` is what used to work (and still works) on other places (such as `isUserInRole`), this commit just brings `@RolesAllowed` on par.

gsmet: OK, let's have @sberyozkin check that.

sberyozkin: @Ladicek That looks fine, but do we also have to tweak `smallrye-jwt` code somewhere as you said it supports `*` ?

Ladicek: Yea: https://github.com/smallrye/smallrye-jwt/blob/master/implementation/src/main/java/io/smallrye/jwt/auth/jaxrs/RolesAllowedFilter.java#L45

I'll submit a PR to SmallRye JWT once I verify that the SmallRye JWT code is actually used. The security stuff in Quarkus is hard to navigate :-)

sberyozkin: @Ladicek Cool, thanks for spotting it, that JAX-RS code is not used by `quarkus-smallrye-jwt`

sberyozkin (approved):

sberyozkin:
@gsmet Hi Guillaume, thanks for asking to review, LGTM

PR #8086, opened 2020-03-23
ArC - fix post construct invocation for hierarchical cases with beans from different packages

### LINKED ISSUE #8085: ArC doesn't allow package private @PostConstruct in hierarchy of beans with different packages
Similarly to #8020 for observers, I've tested initializers (which work) and `@PostConstruct` which doesn't work and throws `IllegalAccess`.

Scenario was the same - bean `Foo` and bean `@Alternative Bar extends Foo`, both in different packages. `Foo` declares a `@PostConstruct` which should be inherited by `Bar` but this blows up upon invocation.

We should be able to workaround this with reflection as we do for the other two cases.

FYI @mkouba

manovotn:
… from different packages.

Fixes #8085

### DIFF
--- a/independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/BeanGenerator.java
+++ b/independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/BeanGenerator.java
@@ -1431,10 +1431,12 @@ public class BeanGenerator extends AbstractGenerator {
                     DotNames.POST_CONSTRUCT,
                     bean.getDeployment().getIndex());
             for (MethodInfo callback : postConstructCallbacks) {
-                if (Modifier.isPrivate(callback.flags())) {
-                    privateMembers.add(isApplicationClass,
-                            String.format("@PostConstruct callback %s#%s()", callback.declaringClass().name(),
-                                    callback.name()));
+                if (isReflectionFallbackNeeded(callback, targetPackage)) {
+                    if (Modifier.isPrivate(callback.flags())) {
+                        privateMembers.add(isApplicationClass,
+                                String.format("@PostConstruct callback %s#%s()", callback.declaringClass().name(),
+                                        callback.name()));
+                    }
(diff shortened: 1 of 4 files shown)


independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/BeanGenerator.java
```diff
@@ -1431,10 +1431,12 @@ void implementCreateForClassBean(ClassOutput classOutput, ClassCreator beanCreat
                     DotNames.POST_CONSTRUCT,
                     bean.getDeployment().getIndex());
             for (MethodInfo callback : postConstructCallbacks) {
-                if (Modifier.isPrivate(callback.flags())) {
-                    privateMembers.add(isApplicationClass,
-                            String.format("@PostConstruct callback %s#%s()", callback.declaringClass().name(),
-                                    callback.name()));
+                if (isReflectionFallbackNeeded(callback, targetPackage)) {
+                    if (Modifier.isPrivate(callback.flags())) {
+                        privateMembers.add(isApplicationClass,
```
manovotn: This only logs private members in the code. Do we want some logging for the pack-private plus package split case too? Because effectively both scenarios force us into reflection which is something we want to avoid, right?

mkouba: I'm not sure. We could log something but the message would have to be quite "elaborate".

mkouba (approved):

independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/BeanGenerator.java
manovotn: Yeah and the detection too (since this can happen on multiple places), not sure it's worth it.

gsmet:
We decided to not backport that one for now.

Issue #8063, opened 2020-03-22
Jackson deserialization not supported for REST client without including RESTEasy server


PieterjanDeconinck:
**Describe the bug**
Jackson JSON deserialization is not supported with Quarkus REST client without including dependencies for the RESTEasy server.

**Expected behavior**
One should be able to use the Quarkus REST client with Jackson, independent from the server.
The following dependencies should be sufficient to support Jackson JSON deserialization on the REST client:
```
<dependency>
    <groupId>io.quarkus</groupId>
    <artifactId>quarkus-rest-client</artifactId>
</dependency>
<dependency>
    <groupId>io.quarkus</groupId>
    <artifactId>quarkus-jackson</artifactId>
</dependency>
```

**Actual behavior**
Running with just these 2 dependencies produces the following error:
```
WARN  [io.qua.res.com.dep.ResteasyCommonProcessor] (build-11) Quarkus detected the need of REST JSON support but you have not provided the necessary JSON extension for this. You can visit https://quarkus.io/guides/rest-json for more information on how to set one.
...
ERROR [...] ProcessingException while creating reply for journey details request: RESTEASY003145: Unable to find a MessageBodyReader of content-type application/json and type class X.
```
The deserialization issue is solved when adding the quarkus-resteasy-jackson dependency. However this also starts the web server, which might not always be desired and adds some unused overhead.
```
<dependency>
    <groupId>io.quarkus</groupId>
    <artifactId>quarkus-rest-client</artifactId>
</dependency>
<dependency>
    <groupId>io.quarkus</groupId>
    <artifactId>quarkus-resteasy-jackson</artifactId>
</dependency>
```
In my (basic) test case, the issue could be solved by adding an additional dependency. But this might not be an appropriate solution (or might not work in all cases - especially in native image).
```
<dependency>
    <groupId>io.quarkus</groupId>
    <artifactId>quarkus-rest-client</artifactId>
</dependency>
<dependency>
    <groupId>io.quarkus</groupId>
    <artifactId>quarkus-jackson</artifactId>
</dependency>
<dependency>
    <groupId>org.jboss.resteasy</groupId>
    <artifactId>resteasy-jackson2-provider</artifactId>
</dependency>
```

**Environment:**
 - Quarkus 1.3.0.Final
 - Maven 3.6.3 with option -Dquarkus.native.container-build=true

**Additional context**
Original question on StackOverflow: https://stackoverflow.com/q/60787746/12355118

[...]

PR #8089, opened 2020-03-23
Provide rest-client-<serialization> extensions

gsmet:
Relates to: #8063


geoand:
I also tested the `rest-client-quickstart` with the `quarkus-rest-client-jsonb` and it worked like a charm!

The only thing I am not excited about is that for the tests in the new modules, the meaningful assertions are done in the application code, making it hard to see the real error if a test fails. 
But that shouldn't be a problem since the applications are very small, so nailing down the cause of any future test failure should be easy.

geoand (approved):

geoand (approved):
This is very useful for users, thanks!

PR #8092, opened 2020-03-23
Fixed doc compilation warnings

gastaldi:


docs/src/main/asciidoc/amazon-lambda.adoc
```diff
@@ -480,7 +480,7 @@ when using synchronous mode, due to issues in the GraalVM compilation (at presen
 Add `quarkus-jaxb` as a dependency in your Maven `pom.xml`, or Gradle `build.gradle` file.
 
 You must also force your AWS service client for SQS, SNS, S3 et al, to use the URL Connection client,
-which connects to AWS services over HTTPS, hence the inclusion of the SSL enabled property, as described in the <<Using HTTPS>> section above.
+which connects to AWS services over HTTPS, hence the inclusion of the SSL enabled property, as described in the <<Using HTTPS or SSL/TLS>> section above.
```
gsmet: We need to use an id here to avoid that kind of issues.

gsmet (changes requested):
Good catch, comment inline.

docs/src/main/asciidoc/amazon-lambda.adoc
gastaldi: done

gsmet (approved):

gsmet:
Ok, I will have to backport only part of it as I'm not sure the lambda stuff is in the 1.3 branch.

Will let Git do the work for me :).

gsmet:
Apparently, these issues were only in master.

PR #8114, opened 2020-03-24
Add more test to MongoDB with Panache

### LINKED ISSUE #7698: Add more tests to MongoDB with Panache
**Description**
MongoDB with Panache provides _functional_ integration tests, that test the main functionality of MongoDB with Panache based on functional REST endpoints.

This is not exhaustive and the extension test coverage is not good.

**Implementation ideas**
Add a technical endpoint that test the various methods of MongoDB with Panache API that are not covered by the current integration test, and return a single OK. 
This is inspired by the way the current Hibernate with Panache integration test works.

loicmathieu:
This PR adds a bunch of tests to MongoDB with Panache.

Fixes #7698

I don't know if it's easily possible to have the test coverage of a specific extension before and after a PR but this could be a good way to spot missing tests.

I found two bugs (so worth it to add the tests) and fix them in the same PR.

### DIFF
--- a/integration-tests/mongodb-panache/src/main/java/io/quarkus/it/mongodb/panache/test/TestImperativeEntity.java
+++ b/integration-tests/mongodb-panache/src/main/java/io/quarkus/it/mongodb/panache/test/TestImperativeEntity.java
@@ -1,0 +1,20 @@
+package io.quarkus.it.mongodb.panache.test;
+
+import io.quarkus.mongodb.panache.PanacheMongoEntity;
+import io.quarkus.runtime.annotations.RegisterForReflection;
+
+@RegisterForReflection
+public class TestImperativeEntity extends PanacheMongoEntity {
+    public String title;
+    public String category;
+    public String description;
+
+    public TestImperativeEntity() {
+    }
+
+    public TestImperativeEntity(String title, String category, String description) {
+        this.title = title;
+        this.category = category;
+        this.description = description;
+    }
+}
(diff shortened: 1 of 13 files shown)


loicmathieu:
@cescoffier I found a tricky bug on Reactive MongoDB with Panache, I previously compose Uni with `and()` in a not-working way. 
I found a correct way using a flatMap but I wonder if there is an easier way to compose multiple Uni. If not maybe something is missing in Mutiny ;)
So, can you check this changes : https://github.com/quarkusio/quarkus/pull/8114/files#diff-6f82b3b3188b379b72862a3cf8e202c2

loicmathieu:
The reactive changes in the following commit https://github.com/quarkusio/quarkus/pull/8114/commits/140259fb4249900f52fd1efadcf5cbb139bf14f2 has been reviewed yesterday with @cescoffier

integration-tests/mongodb-panache/src/main/java/io/quarkus/it/mongodb/panache/test/TestImperativeEntity.java
```diff
@@ -0,0 +1,20 @@
+package io.quarkus.it.mongodb.panache.test;
+
+import io.quarkus.mongodb.panache.PanacheMongoEntity;
+import io.quarkus.runtime.annotations.RegisterForReflection;
+
+@RegisterForReflection
```
FroMage: Isn't that automatic? Shouldn't it be?

FroMage (approved):
LGTM, except the question around reflection registering, but it's orthogonal to this issue.

integration-tests/mongodb-panache/src/main/java/io/quarkus/it/mongodb/panache/test/TestImperativeEntity.java
loicmathieu: This should be fixed by #8211 I'll test to remove it as the PR is fixed and send another PR

FroMage: Right, great.

PR #8115, opened 2020-03-24
Use CustomScopeAnnotationsBuildItem where needed

mkouba:
- follows up on #7076

### DIFF
--- a/extensions/scheduler/deployment/src/main/java/io/quarkus/scheduler/deployment/SchedulerProcessor.java
+++ b/extensions/scheduler/deployment/src/main/java/io/quarkus/scheduler/deployment/SchedulerProcessor.java
@@ -104,14 +104,14 @@ public class SchedulerProcessor {
 
             @Override
             public void transform(TransformationContext context) {
-                if (context.isClass() && !scopes.isScopeDeclaredOn(context.getTarget().asClass())) {
+                ClassInfo target = context.getTarget().asClass();
+                if (!scopes.isScopeIn(context.getAnnotations())
+                        && (target.annotations().containsKey(SCHEDULED_NAME)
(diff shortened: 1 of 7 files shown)


extensions/scheduler/deployment/src/main/java/io/quarkus/scheduler/deployment/SchedulerProcessor.java
```diff
@@ -104,14 +104,14 @@ public boolean appliesTo(org.jboss.jandex.AnnotationTarget.Kind kind) {
...
+                        && (target.annotations().containsKey(SCHEDULED_NAME)
+                                || target.annotations().containsKey(SCHEDULES_NAME))) {
                     // Class with no scope annotation but with @Scheduled method
-                    if (context.getTarget().asClass().annotations().containsKey(SCHEDULED_NAME)
-                            || context.getTarget().asClass().annotations().containsKey(SCHEDULES_NAME)) {
-                        LOGGER.debugf("Found scheduled business methods on a class %s with no annotations - adding @Singleton",
-                                context.getTarget());
-                        context.transform().add(Singleton.class).done();
-                    }
+                    LOGGER.debugf("Found scheduled business methods on a class %s with no annotations - adding @Singleton",
```
jmartisk: would be better to say "with no scope defined", there could be some other unrelated annotations

jmartisk (approved):
Experimented with it locally and got it working. Just left a minor note about a debug message.
A test would be nice too, but I'll give a +1 already :)

mkouba:
> A test would be nice too..

@jmartisk I don't think we need a new test. It's just improving the existing code...

manovotn (approved):
Nice catch! Apparently in my original PR I missed a lot of cases which should have been updated.
You should have bullied me harder :-D

PR #8116, opened 2020-03-24
Liquibase - Record service classes implementation instead of  generating a native resource file for each service interface

machi1990:
This cleans up the extension processing part a little bit as we can record these info in a registry
and do a map lookup during runtime instead of reading the generated resource file for each service
class

Follows up this https://github.com/quarkusio/quarkus/pull/8065#issuecomment-603217470 comment.

### DIFF
--- a/extensions/liquibase/deployment/src/main/java/io/quarkus/liquibase/LiquibaseProcessor.java
+++ b/extensions/liquibase/deployment/src/main/java/io/quarkus/liquibase/LiquibaseProcessor.java
@@ -1,15 +1,14 @@
 package io.quarkus.liquibase;
 
 import static io.quarkus.deployment.annotations.ExecutionTime.STATIC_INIT;
-import static io.quarkus.liquibase.runtime.graal.LiquibaseServiceLoader.serviceResourceFile;
 
-import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.LinkedHashSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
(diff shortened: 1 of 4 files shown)


extensions/liquibase/deployment/src/main/java/io/quarkus/liquibase/LiquibaseProcessor.java
```diff
@@ -96,6 +97,7 @@ void nativeImageConfiguration(
                 liquibase.sql.visitor.SqlVisitor.class);
 
         // load the liquibase services
+        Map<String, List<String>> serviceClassesImplementationRegistry = new HashMap<>();
```
gastaldi: Should we care about the order here (`LinkedHashMap`) ?

gastaldi: > Should we care about the order here (`LinkedHashMap`) ?

Probably yes, to avoid uncertainty while loading each service, WDYT?

machi1990: > > Should we care about the order here (`LinkedHashMap`) ?
> 
> Probably yes, to avoid uncertainty while loading all services, WDYT?

I do not think so. I think we just need a lookup table here, order is irrelevant. What uncertainty are you thinking of?

gastaldi: Ah nevermind, I thought you were iterating entries. I should stop reviewing code late at night :stuck_out_tongue:

gastaldi (approved):

gastaldi:
I'll let @gsmet review and merge

machi1990:
Thanks @gastaldi

gsmet:
I rebased it, let's wait for CI.

gsmet (approved):

Issue #8088, opened 2020-03-23
Add `Proxy` configuration group to `OidcTenantConfig`


sberyozkin:
**Description**
Proxy options may need to be set for the OIDC connection to be created.

**Implementation ideas**
Paulo explained that Vertx client will use [ProxyOptions](https://vertx.io/docs/apidocs/io/vertx/core/net/ProxyOptions.html) if they are set.

Proposal: add `OidcTenantConfig` `Proxy` configuration group and have 4 optional properties there for a start: `host`, `port`, `username`, `password`:
```
quarkus.oidc.proxy.host=123.345.456.789
quarkus.oidc.proxy.port=5678
```

devisions:
@sberyozkin As discussed on the chat, as we have this need on our case, I'll try to come back with a PR. Thanks again for the feedback.

sberyozkin:
@devisions Thanks, perhaps it can make it to 1.3.1.Final then

devisions:
@sberyozkin [PR 8120](https://github.com/quarkusio/quarkus/pull/8120) contains the additions.
I am waiting for any review and feedback. As being my first contribution to this project, maybe I am missing something.

After this, I would like to extend the existing blog entry to showcase this addition also. But that's another story that we can discuss right after we complete this topic. 😊

devisions:
It would be great to include it in the upcoming 1.3.1 whose milestone has due date this Friday.

devisions:
@sberyozkin Sergey, just a small update that reflects our final decision is:
- to just provide the "host" (as `quarkus.oidc.proxy.host` config item)
- "port" has the default value of 80 (if not explicitly configured)
- of course, "username" and "password" are optional, since not all proxies require authn.

sberyozkin:
@devisions Sorry for a delay, I actually missed this comment. I commented earlier at the PR that it was considered to be a new feature and we could only apply the bug fixes to 1.3.1.Final. 
Sorry, `1.4.0` won't be in the works for long :-)

PR #8120, opened 2020-03-24
Add quarkus.oidc.proxy config

dxps:
One addition to `oidc` extension: ability to use a proxy for talking with OIDC provider.
The following configuration items were added and can be used for this purpose:
- `quarkus.oidc.proxy.host`
- `quarkus.oidc.proxy.port`
- `quarkus.oidc.proxy.username`
- `quarkus.oidc.proxy.password`
- ~`quarkus.oidc.proxy.type`~ (update: ignored for now)

This refers to [issue 8088](https://github.com/quarkusio/quarkus/issues/8088).

Waiting for Quarkus team review and feedback.
Thanks.

### DIFF
--- a/extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/OidcRecorder.java
+++ b/extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/OidcRecorder.java
@@ -2,6 +2,7 @@ package io.quarkus.oidc.runtime;
 
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionException;
 import java.util.function.Function;
@@ -19,6 +20,8 @@ import io.quarkus.runtime.configuration.ConfigurationException;
 import io.vertx.core.AsyncResult;
 import io.vertx.core.Handler;
 import io.vertx.core.Vertx;
+import io.vertx.core.json.JsonObject;
+import io.vertx.core.net.ProxyOptions;
 import io.vertx.ext.auth.PubSecKeyOptions;
 import io.vertx.ext.auth.oauth2.OAuth2Auth;
 import io.vertx.ext.auth.oauth2.OAuth2ClientOptions;
@@ -132,6 +135,11 @@ public class OidcRecorder {
             options.setClientSecretParameterName(null);
         }
 
+        Optional<ProxyOptions> proxyOpt = toProxyOptions(oidcConfig.getProxy());
+        if (proxyOpt.isPresent()) {
+            options.setProxyOptions(proxyOpt.get());
+        }
+
         final long connectionDelayInSecs = oidcConfig.getConnectionDelay().isPresent()
                 ? oidcConfig.getConnectionDelay().get().toMillis() / 1000
                 : 0;
@@ -185,6 +193,23 @@ public class OidcRecorder {
                 + "Please make sure it is correct. Note it has to end with a realm value if you work with Keycloak, for example:"
                 + " 'https://localhost:8180/auth/realms/quarkus'";
         return new OIDCException(message, cause);
+    }
+
+    protected static Optional<ProxyOptions> toProxyOptions(OidcTenantConfig.Proxy proxyConfig) {
+        // Proxy is enabled if (at least) "host" is configured.
+        if (!proxyConfig.host.isPresent()) {
+            return Optional.empty();
+        }
+        JsonObject jsonOptions = new JsonObject();
+        jsonOptions.put("host", proxyConfig.host.get());
+        jsonOptions.put("port", proxyConfig.port);
+        if (proxyConfig.username.isPresent()) {
+            jsonOptions.put("username", proxyConfig.username.get());
+        }
+        if (proxyConfig.password.isPresent()) {
+            jsonOptions.put("password", proxyConfig.password.get());
+        }
+        return Optional.of(new ProxyOptions(jsonOptions));
     }
 
 }
--- a/extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/OidcTenantConfig.java
+++ b/extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/OidcTenantConfig.java
@@ -82,6 +82,11 @@ public class OidcTenantConfig {
     @ConfigItem
     Credentials credentials = new Credentials();
     /**
+     * Options to configure a proxy that OIDC adapter will use for talking with OIDC server.
+     */
+    @ConfigItem
+    Proxy proxy = new Proxy();
+    /**
      * Different options to configure authorization requests
      */
     Authentication authentication = new Authentication();
@@ -172,6 +177,14 @@ public class OidcTenantConfig {
 
     public void setTenantId(String tenantId) {
         this.tenantId = Optional.of(tenantId);
+    }
+
+    public Proxy getProxy() {
+        return proxy;
+    }
+
+    public void setProxy(Proxy proxy) {
+        this.proxy = proxy;
     }
 
     @ConfigGroup
@@ -465,6 +478,37 @@ public class OidcTenantConfig {
         public void setPrincipalClaim(String principalClaim) {
             this.principalClaim = Optional.of(principalClaim);
         }
+    }
+
+    @ConfigGroup
+    public static class Proxy {
+
+        /**
+         * The host (name or IP address) of the Proxy.<br/>
+         * Note: If OIDC adapter needs to use a Proxy to talk with OIDC server (Provider),
+         * then at least the "host" config item must be configured to enable the usage of a Proxy.
+         */
+        @ConfigItem
+        public Optional<String> host = Optional.empty();
+
+        /**
+         * The port number of the Proxy. Default value is 80.
+         */
+        @ConfigItem(defaultValue = "80")
+        public int port = 80;
+
+        /**
+         * The username, if Proxy needs authentication.
+         */
+        @ConfigItem
+        public Optional<String> username = Optional.empty();
+
+        /**
+         * The password, if Proxy needs authentication.
+         */
+        @ConfigItem
+        public Optional<String> password = Optional.empty();
+
     }
 
     public static enum ApplicationType {
--- a/extensions/oidc/runtime/src/test/java/io/quarkus/oidc/runtime/OidcRecorderTest.java
+++ b/extensions/oidc/runtime/src/test/java/io/quarkus/oidc/runtime/OidcRecorderTest.java
@@ -1,0 +1,24 @@
+package io.quarkus.oidc.runtime;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+import java.util.Optional;
+
+import org.junit.jupiter.api.Test;
+
+public class OidcRecorderTest {
+
+    @Test
+    public void testtoProxyOptionsWithHostCheckPresent() {
+        OidcTenantConfig.Proxy proxy = new OidcTenantConfig.Proxy();
+        proxy.host = Optional.of("server.example.com");
+        assertTrue(OidcRecorder.toProxyOptions(proxy).isPresent());
+    }
+
+    @Test
+    public void testtoProxyOptionsWithoutHostCheckNonPresent() {
+        OidcTenantConfig.Proxy proxy = new OidcTenantConfig.Proxy();
+        assertFalse(OidcRecorder.toProxyOptions(proxy).isPresent());
+    }
+
+}


extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/OidcTenantConfig.java
```diff
@@ -467,6 +483,59 @@ public void setPrincipalClaim(String principalClaim) {
...
+         */
+        @ConfigItem
+        public Optional<String> password = Optional.empty();
+
+        /**
+         * The proxy type. Default value is "HTTP".<br/>
+         * Allowed values (as defined in {@link io.vertx.core.net.ProxyType}) are: "HTTP", "SOCKS4", and "SOCKS5".
+         */
+        @ConfigItem(defaultValue = "HTTP")
+        public Optional<String> type = Optional.empty();
```
sberyozkin: Please drop `type` for now, we can add later if someone will prefer non-`HTTP`

extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/OidcTenantConfig.java
```diff
@@ -467,6 +483,59 @@ public void setPrincipalClaim(String principalClaim) {
...
+
+        /**
+         * Get this proxy config as {@link ProxyOptions}
+         * 
+         * @return a ProxyOptions objects built based on the included config items
+         */
+        ProxyOptions asProxyOptions() {
```
sberyozkin: Please move this to the internal `OidcRecorder` helper, we don't really put `Vertx` dependencies in `OidcTenantConfig`; instead please add getters/setters (as is done alsewhere), getters are not really needed, but setters would help to `TenantConfigResolver` multi-tenancy implementations which create `OidcTenantConfig` :-)

extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/OidcRecorder.java
```diff
@@ -132,6 +132,15 @@ private TenantConfigContext createTenantContext(Vertx vertx, OidcTenantConfig oi
...
 
+        // Proxy config
+        OidcTenantConfig.Proxy proxy = oidcConfig.getProxy();
+        if (proxy.host.isPresent() && proxy.port.isPresent()) {
+            options.setProxyOptions(proxy.asProxyOptions());
+        } else if (proxy.host.isPresent() || proxy.port.isPresent()) {
+            throw new ConfigurationException(
+                    "Both 'proxy.host' and 'proxy.port' must be configured to enable the proxy usage.");
+        }
+
```
sberyozkin: Yes, it would help to have optional configuration groups :-). But we can simplify a bit and ignore everything unless `proxy.host` is set, and only check the port inside of that condition
```
        if (proxy.host.isPresent()) {
           if (!proxy.port.isPresent()) {
               throw new ConfigurationException("Both 'proxy.host' and 'proxy.port' must be configured to enable the proxy usage.");
        }
        options.setProxyOptions(proxy.asProxyOptions());
   } 
```
How does it look :-) ?

sberyozkin:
@devisions Many thanks for this PR, minor updates are suggested, cheers

extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/OidcTenantConfig.java
devisions: You are right, indeed. 👍

extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/OidcRecorder.java
devisions: Oh, yeah. True also! And it looks more compact and nice. 😊

devisions:
@sberyozkin Many thanks for the feedback, Sergey! Changes submitted. I hope it's ok now.

extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/OidcRecorder.java
```diff
@@ -132,6 +134,14 @@ private TenantConfigContext createTenantContext(Vertx vertx, OidcTenantConfig oi
             options.setClientSecretParameterName(null);
         }
 
+        OidcTenantConfig.Proxy proxyConfig = oidcConfig.getProxy();
+        if (proxyConfig.host.isPresent()) {
+            if (!proxyConfig.port.isPresent())
```
sberyozkin: This also requires `{` and `}` per the checkstyle rules AFAIK.

sberyozkin:
@devisions It all looks perfect to me now, as far as I recall the curly braces should be used all the time, so please tweak that and it should be ready to go.
There is some strange JDK11 compilation error which is not related to your PR.
Can you please do another minor update re those curly braces, squash the commits, and also fetch and `rebase  upstream/master` and force-push ?
Thanks

extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/OidcTenantConfig.java
```diff
@@ -467,6 +480,37 @@ public void setPrincipalClaim(String principalClaim) {
...
+         * then at least both "host" and "port" items must be configured to enable the usage of a Proxy.
+         */
+        @ConfigItem
+        public Optional<String> host = Optional.empty();
+
+        /**
+         * The port number of the Proxy.
+         */
+        @ConfigItem
+        public Optional<Integer> port = Optional.empty();
```
gastaldi: Use `OptionalInt` instead

devisions: Sure, done it. Thanks.

extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/OidcRecorder.java
devisions: Yeah, I have it. Probably it existed before formatting was run. Now it exists.

devisions:
@sberyozkin We should be good now, right? 😊 Thanks

gastaldi:
Please squash your commits into a single one

sberyozkin:
Hi @devisions It looks nice and ready to go :-), one more thing, @stuartwdouglas recommends avoiding using lambdas in the recorder in general due to them being a bit suboptimal. Can you please give me a favour and update that `toProxyOptions` with simple `if proxy.username.isPresent()` ? You can just use `.get()` for `host` and `port` because the check that both of these props are set has already been done.

And squash as George has said.
Cheers

devisions:
Hi @sberyozkin , Sure. You mean like this:
```java
private ProxyOptions toProxyOptions(OidcTenantConfig.Proxy proxyConfig) {
        JsonObject jsonOptions = new JsonObject();
        jsonOptions.put("host", proxyConfig.host.get());
        jsonOptions.put("host", proxyConfig.port.getAsInt());
        if (proxyConfig.username.isPresent()) {
            jsonOptions.put("username", proxyConfig.username.get());
        }
        if (proxyConfig.password.isPresent()) {
            jsonOptions.put("password", proxyConfig.password.get());
        }
        return new ProxyOptions(jsonOptions);
    }
```
And I'll try to do the squashing now...

gastaldi:
Be aware that host and port are optional too

gastaldi:
Perhaps it's better to move the optional checks from the Recorder class to this method. I forgot you were performing the optional test there

gastaldi:
It woukd be nice to also have a small unit test to ensure that this method works

devisions:
@gastaldi Roger both, let me do these now ...

devisions:
> Be aware that host and port are optional too
@gastaldi George, if both host and port are optional, therefore we "enable" the proxy if at least "host" is being defined in the config?

gastaldi:
> if both host and port are optional, therefore we "enable" the proxy if at least "host" is being defined in the config?

Yes, that makes sense, and make sure to set a default value of `80` for the port

sberyozkin:
@devisions @gastaldi so if we move the optional checks out of OidcRecorder to this method then if then setting a host is enough (with a default `80` port) to return `ProxyOptions` and if `host` is not set then `null` is returned ? That is good

devisions:
@sberyozkin Actually, the enablement is done in `.createTenantContext()` as:
```java
        Optional<ProxyOptions> proxyOpt = toProxyOptions(oidcConfig.getProxy());
        if (proxyOpt.isPresent()) {
            options.setProxyOptions(proxyOpt.get());
        }
```
And the rules are now moved into `toProxyOptions(Proxy)`.

extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/OidcTenantConfig.java
```diff
@@ -467,6 +477,37 @@ public void setPrincipalClaim(String principalClaim) {
...
+         * Note: If OIDC adapter needs to use a Proxy to talk with OIDC server (Provider),
+         * then at least both "host" and "port" items must be configured to enable the usage of a Proxy.
+         */
+        @ConfigItem
+        public Optional<String> host = Optional.empty();
+
+        /**
+         * The port number of the Proxy.
+         */
+        @ConfigItem
```
gastaldi: You can set the defaultValue of `80` here instead

devisions: 👍 Updated submitted.

extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/OidcTenantConfig.java
```diff
@@ -467,6 +477,37 @@ public void setPrincipalClaim(String principalClaim) {
...
+         * then at least both "host" and "port" items must be configured to enable the usage of a Proxy.
+         */
+        @ConfigItem
+        public Optional<String> host = Optional.empty();
+
+        /**
+         * The port number of the Proxy.
+         */
+        @ConfigItem(defaultValue = "80")
+        public OptionalInt port = OptionalInt.empty();
```
gastaldi: Not being picky here, but you can also use `OptionalInt.of(80)` for cases where config injection is not available :wink:

gastaldi: Or maybe replace the OptionalInt with an `int`, since we're assuming a default value now

extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/OidcTenantConfig.java
```diff
@@ -467,6 +477,37 @@ public void setPrincipalClaim(String principalClaim) {
         }
     }
 
+    @ConfigGroup
+    public static class Proxy {
+
+        /**
+         * The host (name or IP address) of the Proxy.<br/>
+         * Note: If OIDC adapter needs to use a Proxy to talk with OIDC server (Provider),
+         * then at least both "host" and "port" items must be configured to enable the usage of a Proxy.
```
gastaldi: `port` is no longer required now as it assumes 80 as the default value

extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/OidcTenantConfig.java
devisions: No, it's fine. 😊 You're right.

extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/OidcTenantConfig.java
```diff
@@ -467,6 +477,37 @@ public void setPrincipalClaim(String principalClaim) {
...
+         */
+        @ConfigItem
+        public Optional<String> host = Optional.empty();
+
+        /**
+         * The port number of the Proxy.
+         */
+        @ConfigItem
+        public OptionalInt port = OptionalInt.of(80);
```
gastaldi: Ah sorry, I didn't mean to remove the `defaultValue` from @ConfigItem. I was just suggesting to make it non-injection-environments friendly :)

extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/OidcTenantConfig.java
devisions: You're right! Again! 😊  Sorry, forgot about it.
I was in a hurry to make sure this small scope is covered.

extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/OidcTenantConfig.java
```diff
@@ -1,10 +1,7 @@
 package io.quarkus.oidc.runtime;
 
 import java.time.Duration;
-import java.util.Arrays;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
+import java.util.*;
```
gastaldi: Also we don't use star imports. Can you double-check if your IDE is using the Quarkus formatter?

extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/OidcTenantConfig.java
devisions: Np, but you lost me. Can you provide the concrete expectiation, please?

Sorry that I keep coming with non standard things in the code.
But eventually, after this small journey I will definitely want to contribute later to this wonderful project.

extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/OidcTenantConfig.java
devisions: So, woud this work:
```java
@ConfigItem
public int port = 80;
```

extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/OidcTenantConfig.java
gastaldi: No problem. I mean something like: 

```java 
@ConfigItem(defaultValue="80")
public int port = 80;
```

extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/OidcTenantConfig.java
devisions: Indeed. It was an IDE issue.

extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/OidcTenantConfig.java
devisions: So, would this work? 
```java
@ConfigItem
public int port = 80;
```
I'm a bit lost now ... Please, a last help here.

gastaldi: Yes, it should work, but you need to add the `defaultValue` attribute in `@ConfigItem` AFAIK, as I showed in https://github.com/quarkusio/quarkus/pull/8120#discussion_r397866426

extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/OidcTenantConfig.java
devisions: Can we clarify it here on the conversation below, please?

extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/OidcTenantConfig.java
sberyozkin:
```
@ConfigItem(defaultValue="80")
public int port;
```
? - we do it for booleans this way

sberyozkin: Sorry @gastaldi, @devisions, it should probably be
```
@ConfigItem(defaultValue="80")
public int port = 80;
```
as suggested by George if we want to make it non-injection env friendly (for TenantConfigResolver, etc)

sberyozkin:
`boring-cyborg` got over-excited and added lots of labels :-)

gastaldi (approved):

sberyozkin:
@devisions, nearly there :-), can you please do `mvn process-sources` and and squash again ?

devisions:
@sberyozkin Done! Are we there? ☺️

sberyozkin:
@devisions Hope so :-), it is already approved

sberyozkin:
@devisions thanks for your PR. It will only be available in `1.4.0` though because only critical go to `1.3.1.Final`. `1.4.0` will be out soon enough too I guess :-)

devisions:
@sberyozkin @gastaldi Thank you both for the support and quick feedback!

PR #8127, opened 2020-03-25
Load Graal SDK Parent first

### LINKED ISSUE #8035: Quarkus 1.3.0.Final class loader can not load Graal's polyglot framework
New class loader of Quarkus 1.3.0 can not load polyglot framework of GraalVM,
Previous version (1.2.1) works fine.

**Expected behavior**
JavaScript code works when executed inside of REST resource

**Actual behavior**
Throws exception
org.jboss.resteasy.spi.UnhandledException: java.lang.IllegalStateException: No language and polyglot implementation was found on the classpath. Make sure the truffle-api.jar is on the classpath.

**To Reproduce**
Create and run following simple resource:
<pre>
import javax.ws.rs.GET;
import javax.ws.rs.Path;
import javax.ws.rs.Produces;
import javax.ws.rs.core.MediaType;
import org.graalvm.polyglot.*;
import org.graalvm.polyglot.proxy.*;

@Path("/hello")
public class ExampleResource {

    @GET
    public String hello() {
        String out = "From JS:";
        try (Context context = Context.create()) {
            Value function = context.eval("js", "x => x+1");
            assert function.canExecute();
            int x = function.execute(41).asInt();
            out=out+x;
        }
        return out;
    }
}
</pre>
**Configuration**
Simplest possible

**Environment**
Linux alhp.alhome 5.5.9-200.fc31.x86_64
openjdk version "11.0.6" 2020-01-14
OpenJDK Runtime Environment GraalVM CE 20.0.0 (build 11.0.6+9-jvmci-20.0-b02)
OpenJDK 64-Bit Server VM GraalVM CE 20.0.0 (build 11.0.6+9-jvmci-20.0-b02, mixed mode, sharing)

stuartwdouglas:
We always want to load this from the actual
GraalVM installation if present, this will only
be loaded from the artifact if running on a
normal JVM.

Fixes #8035

### DIFF
--- a/core/runtime/pom.xml
+++ b/core/runtime/pom.xml
@@ -104,6 +104,8 @@
                 <artifactId>quarkus-bootstrap-maven-plugin</artifactId>
                 <configuration>
                     <parentFirstArtifacts>
+                        <parentFirstArtifact>org.graalvm.sdk:graal-sdk</parentFirstArtifact>
+                        <parentFirstArtifact>org.graalvm.nativeimage:svm</parentFirstArtifact>
                         <parentFirstArtifact>io.quarkus:quarkus-bootstrap-core</parentFirstArtifact>
                         <parentFirstArtifact>io.quarkus:quarkus-development-mode-spi</parentFirstArtifact>
                         <parentFirstArtifact>org.jboss.logmanager:jboss-logmanager-embedded</parentFirstArtifact>


gastaldi (approved):

gastaldi:
How about `org.graalvm.nativeimage:svm` ?

stuartwdouglas:
I guess that should also be parent first, but is much less likely to cause problems.

PR #8128, opened 2020-03-25
Remove MS SQL Server JDBC driver from Dependabot

gastaldi:
Dependabot always suggests the JDK13 driver, which is always incompatible.
Revert this commit when Dependabot supports Regular expression patterns in the dependency gavs


gastaldi:
See https://github.com/quarkusio/quarkus/pull/8125 for an example

geoand (approved):
Good idea

PR #8132, opened 2020-03-25
[docker][jib] Bump baseJvmImage from JRE8 to JRE11

OneCricketeer:
JDK11 has been around over a year. Time to upgrade

### DIFF
--- a/devtools/platform-descriptor-json/src/main/resources/templates/dockerfile-jvm.ftl
+++ b/devtools/platform-descriptor-json/src/main/resources/templates/dockerfile-jvm.ftl
@@ -16,7 +16,7 @@
 ###
 FROM registry.access.redhat.com/ubi8/ubi-minimal:8.1
 
-ARG JAVA_PACKAGE=java-1.8.0-openjdk-headless
+ARG JAVA_PACKAGE=java-11-openjdk-headless
 ARG RUN_JAVA_VERSION=1.3.5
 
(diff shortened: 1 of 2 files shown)


geoand:
Thanks a lot!

You would also updating https://github.com/quarkusio/quarkus/blob/master/devtools/platform-descriptor-json/src/main/resources/templates/dockerfile-jvm.ftl#L19 as part of the PR?
Since this template is used to create the Dockerfile that the `container-image-docker` extension uses, it would be nice to have both extensions updated.

cricket007:
Done 😸 

Thanks! 😷

geoand (approved):
Excellent, thanks!

gsmet:
@geoand I think it should be backported, right?

geoand:
@gsmet I don't think so.

We only start warning people against using Java 8 in `1.4`, so I don't think we need to make the change in `1.3`.

geoand:
After discussing with @gsmet I see the merit in this, let's do it.

cricket007:
So, this'll go in 1.4, then? 

Roadmap one can watch for release cycles?

gsmet:
@cricket007 it will be in 1.3.1 that should be released tomorrow.

cricket007:
So, I saw Jib things in here. Is this ticket still valid? https://github.com/GoogleContainerTools/jib/issues/1555

geoand:
I added a comment there.

PR #8149, opened 2020-03-25
Make enum values generated for documentation consistent with converter

gsmet:
At some point, we should probably share that code somehow but we need
something easy to backport for now.


geoand:
Thanks! 

After an intense debugging session I was able to arrive to the same part of code. At least now I know something about doc generation.

geoand (approved):

gsmet:
You should read your emails :).

geoand:
I do, doesn't mean I remember what they say :)

machi1990:
Thanks @gsmet @geoand

PR #8150, opened 2020-03-25
Vault documentation improvement secret-config-kv-path vs kv-secret-en…

### LINKED ISSUE #6062: Vault documentation improvement secret-config-kv-path vs kv-secret-engine-mount-path
**Description**
improve vault documentation with respect to how `https://quarkus.io/guides/vault#quarkus-vault_quarkus.vault.secret-config-kv-path` and `https://quarkus.io/guides/vault#quarkus-vault_quarkus.vault.kv-secret-engine-mount-path`

see [SO issue](https://stackoverflow.com/questions/59198181/cannot-get-vault-kv-v2-secrets-in-quarkus-using-token-auth) and [zulip thread](https://quarkusio.zulipchat.com/#narrow/stream/187030-users/topic/vault.20integration)

vsevel:
Improve documentation with respect to how the the full url gets built from params and config properties.
Fixes https://github.com/quarkusio/quarkus/issues/6062

### DIFF
--- a/extensions/vault/runtime/src/main/java/io/quarkus/vault/runtime/config/VaultRuntimeConfig.java
+++ b/extensions/vault/runtime/src/main/java/io/quarkus/vault/runtime/config/VaultRuntimeConfig.java
@@ -38,6 +38,9 @@ public class VaultRuntimeConfig {
      *
      * Example: https://localhost:8200
      *
+     * See also the documentation for the `kv-secret-engine-mount-path` property for some insights on how
+     * the full Vault url gets built.
+     *
      * @asciidoclet
      */
     @ConfigItem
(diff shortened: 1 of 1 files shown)


sberyozkin:
@vsevel That looks much more informative, where does the formatter complains ? May be it can tweaked a bit ? Thanks

vsevel:
@sberyozkin 
> where does the formatter complain

where do you see that the formatter complains? it fails on:
```
Cache Maven Repository55s
##[error]The operation was canceled.
Run n1hility/cache@v2
##[error]The operation was canceled.
```

I do not understand the error. is there a way to relaunch only that one check?

sberyozkin:
@vsevel sorry, I meant `@formatter:off` in the file, what was the reason for it ?

vsevel:
@sberyozkin 

> sorry, I meant @formatter:off in the file, what was the reason for it ?

not in this case. but in some other comments it was the case where the formatter would change the text of the comment, which would affect how it is rendered in the doc. and since it is asciidoclet, there is no need for the formatter to deal with that kind of comment. I do not remember the exact situation.

but nevermind. I removed it and squashed the commits. I think it is good to go.

sberyozkin:
@vsevel Thanks for the explanation, and showing how to disable the formatter too :-), I was just curious what was that about.

sberyozkin (approved):

PR #8154, opened 2020-03-25
fix: Service type is no longer ignored

iocanel:
This pull request deals with the kubernetes extension completely ignoring the service type, when provided using `quarkus.kubernetes.service-type`.


geoand:
Can we also add an assertion to the default test to make sure it has the default value?

geoand (approved):

PR #8155, opened 2020-03-25
Ensure ordering of Config items when an extension contains multiple Config classes

### LINKED ISSUE #8147: https://quarkus.io/guides/all-config should be in order
the keys are not in order.

You start with  : 

quarkus.knative
quarkus.kubernetes
quarkus.openshift

and go back to : quarkus.knative   ...

geoand:
…onfig classes

Fixes: #8147

### DIFF
--- a/core/processor/src/main/java/io/quarkus/annotation/processor/generate_doc/ConfigDocKey.java
+++ b/core/processor/src/main/java/io/quarkus/annotation/processor/generate_doc/ConfigDocKey.java
@@ -22,6 +22,8 @@ final public class ConfigDocKey implements ConfigDocElement, Comparable<ConfigDo
     private boolean list;
     private boolean passThroughMap;
     private boolean withinAConfigGroup;
+    // if a key is "quarkus.kubernetes.part-of", then the value of this would be "kubernetes"
+    private String topLevelGrouping;
 
     public ConfigDocKey() {
     }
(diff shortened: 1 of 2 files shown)


geoand:
This attempt is by no means perfect, but I should be good enough hopefully to make it into `1.3.1.Final`

core/processor/src/main/java/io/quarkus/annotation/processor/generate_doc/ConfigDocKey.java
```diff
@@ -56,6 +58,19 @@ public String getKey() {
...
+            this.topLevelGrouping = key;
+            return;
+        }
+        String withoutFirstDotIndex = key.substring(firstDotIndex + 1);
+        int secondDotIndex = withoutFirstDotIndex.indexOf('.');
+        if (secondDotIndex == -1) {
+            this.topLevelGrouping = withoutFirstDotIndex;
+            return;
+        }
+        this.topLevelGrouping = withoutFirstDotIndex.substring(0, secondDotIndex);
```
machi1990: We have this info here - https://github.com/quarkusio/quarkus/blob/c8b4f89b8fe4a097e698f2f18d52adc7ab3e1ebe/core/processor/src/main/java/io/quarkus/annotation/processor/generate_doc/ConfigDoItemFinder.java#L58 with `configRootInfo.getName()`, maybe we can pass/set (possibly recursively) this from there instead of computing this?

machi1990 (dismissed):
Just did a quick review , added one suggestion.  I'll need to check the output  later.

core/processor/src/main/java/io/quarkus/annotation/processor/generate_doc/ConfigDocKey.java
geoand: Sure, that makes sense. But I would really like to get this in for `1.3.1.Final` if the output is correct that is.
We can improve in a follow up when we don't have a release date breathing down our necks :)

machi1990: Ok to merge as is and I'll prepare a followup PR.

geoand: Cool, thanks for checking.

If you can open a PR tomorrow, that would be great. Otherwise I might have time to do it

gastaldi (approved):
LGTM

PR #8162, opened 2020-03-25
Remove unneeded paging when using PanacheQuery

### LINKED ISSUE #8098: Panache queries are always issuing SQL statements with paging information
**Describe the bug**
When executing a simple Panache query, the generated SQL statement includes the keywords and parameters in order to do paging, no matter the Panache Query have never been configured with a Page, for example:

```sql
-- PostgreSQL
select
    mytable0_.id as id1_0_,
    mytable0_.name as name2_0_ 
from
    mytable mytable0_ 
where
    mytable0_.name=? limit ?
```

**Expected behavior**
If no paging information has been setup on the PanacheQuery instance, SQL statements should not include, for example: `limit ?`

**Actual behavior**
Paging restrictions are being added to the SQL statement for all queries generated by Panache, even if not paging the PanacheQuery.

**To Reproduce**
1. Enable SQL logging with `quarkus.hibernate-orm.log.sql=true`
2. Execute any simple Panache query like: `MyEntity.list("name", "My Name")`;
3. Check the log and notice the SQL statement includes paging related Information, `limit ?` in the PostgreSQL example.

**Configuration**
```properties
quarkus.hibernate-orm.log.sql=true 
```

Linux localhost.localdomain 5.5.9-200.fc31.x86_64 #1 SMP Thu Mar 12 13:55:19 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux
 - Output of `java -version`:
openjdk version "11.0.6" 2020-01-14
OpenJDK Runtime Environment 18.9 (build 11.0.6+10)
OpenJDK 64-Bit Server VM 18.9 (build 11.0.6+10, mixed mode, sharing)
 - Quarkus version or git rev: 
1.3.0.Final
Apache Maven 3.6.3

**Additional context**

The problem is generated because `PanacheQuery` has a default paging of `Integer.MAX_VALUE` rows at [PanacheQueryImpl](https://github.com/quarkusio/quarkus/blob/251eaab8cd0b95b76a9002dcf16d18f6e46e080e/extensions/panache/hibernate-orm-panache/runtime/src/main/java/io/quarkus/hibernate/orm/panache/runtime/PanacheQueryImpl.java#L38)

The default page should be `null` and every page related action should check if a page size has been set or throw an exception. `setMaxResults(int)` should not be called if the page size has not been provided. Another option is to skip invoking `setMaxResults(int)` if the page size is `Integer.MAX_VALUE`, but this could cause problems with code that really want to stream that quantity of rows only (not something that should be done with JPA, but I stil don't like magic special numbers)

Note: some DBM implementations of paging are not invasive like the PostgreSQL one, but the Oracle ones from the Oracle*Dialect classes are:

```sql
select * from ( <your query> ) where rownum <= ?
```
and it is not nice to have all your queries being logged with DBM auditing tools, with that when it isn't needed

robmv:
Avoid setting paging information on the native query implementation when no paging has been set on PanacheQuery.
 
Closes #8098

### DIFF
--- a/extensions/panache/hibernate-orm-panache/runtime/src/main/java/io/quarkus/hibernate/orm/panache/runtime/PanacheQueryImpl.java
+++ b/extensions/panache/hibernate-orm-panache/runtime/src/main/java/io/quarkus/hibernate/orm/panache/runtime/PanacheQueryImpl.java
@@ -12,6 +12,8 @@ import javax.persistence.LockModeType;
 import javax.persistence.NonUniqueResultException;
 import javax.persistence.Query;
 
+import org.hibernate.engine.spi.RowSelection;
+
 import io.quarkus.hibernate.orm.panache.PanacheQuery;
 import io.quarkus.panache.common.Page;
 import io.quarkus.panache.common.Range;
@@ -27,10 +29,6 @@ public class PanacheQueryImpl<Entity> implements PanacheQuery<Entity> {
     private String query;
     private EntityManager em;
 
-    /*
-     * We store the pageSize and apply it for each request because getFirstResult()
-     * sets the page size to 1
-     */
     private Page page;
     private Long count;
 
@@ -41,7 +39,6 @@ public class PanacheQueryImpl<Entity> implements PanacheQuery<Entity> {
         this.jpaQuery = jpaQuery;
         this.query = query;
         this.paramsArrayOrMap = paramsArrayOrMap;
-        page = new Page(0, Integer.MAX_VALUE);
     }
 
     // Builder
@@ -61,43 +58,43 @@ public class PanacheQueryImpl<Entity> implements PanacheQuery<Entity> {
 
     @Override
     public <T extends Entity> PanacheQuery<T> nextPage() {
-        checkNotInRange();
+        checkPagination();
         return page(page.next());
     }
 
     @Override
     public <T extends Entity> PanacheQuery<T> previousPage() {
-        checkNotInRange();
+        checkPagination();
         return page(page.previous());
     }
 
     @Override
     public <T extends Entity> PanacheQuery<T> firstPage() {
-        checkNotInRange();
+        checkPagination();
         return page(page.first());
     }
 
     @Override
     public <T extends Entity> PanacheQuery<T> lastPage() {
-        checkNotInRange();
+        checkPagination();
         return page(page.index(pageCount() - 1));
     }
 
     @Override
     public boolean hasNextPage() {
-        checkNotInRange();
+        checkPagination();
         return page.index < (pageCount() - 1);
     }
 
     @Override
     public boolean hasPreviousPage() {
-        checkNotInRange();
+        checkPagination();
         return page.index > 0;
     }
 
     @Override
     public int pageCount() {
-        checkNotInRange();
+        checkPagination();
         long count = count();
         if (count == 0)
             return 1; // a single page of zero results
@@ -106,11 +103,15 @@ public class PanacheQueryImpl<Entity> implements PanacheQuery<Entity> {
 
     @Override
     public Page page() {
-        checkNotInRange();
+        checkPagination();
         return page;
     }
 
-    private void checkNotInRange() {
+    private void checkPagination() {
+        if (page == null) {
+            throw new UnsupportedOperationException("Cannot call a page related method, " +
+                    "call page(Page) or page(int, int) to initiate pagination first");
+        }
         if (range != null) {
             throw new UnsupportedOperationException("Cannot call a page related method in a ranged query, " +
                     "call page(Page) or page(int, int) to initiate pagination first");
@@ -121,7 +122,7 @@ public class PanacheQueryImpl<Entity> implements PanacheQuery<Entity> {
     public <T extends Entity> PanacheQuery<T> range(int startIndex, int lastIndex) {
         this.range = Range.of(startIndex, lastIndex);
         // reset the page to its default to be able to switch from page to range
-        this.page = new Page(0, Integer.MAX_VALUE);
+        this.page = null;
         return (PanacheQuery<T>) this;
     }
 
@@ -208,7 +209,7 @@ public class PanacheQueryImpl<Entity> implements PanacheQuery<Entity> {
     public <T extends Entity> Optional<T> singleResultOptional() {
         manageOffsets(2);
         List<T> list = jpaQuery.getResultList();
-        if (list.size() == 2) {
+        if (list.size() > 1) {
             throw new NonUniqueResultException();
         }
 
@@ -220,19 +221,27 @@ public class PanacheQueryImpl<Entity> implements PanacheQuery<Entity> {
             jpaQuery.setFirstResult(range.getStartIndex());
             // range is 0 based, so we add 1
             jpaQuery.setMaxResults(range.getLastIndex() - range.getStartIndex() + 1);
-        } else {
+        } else if (page != null) {
             jpaQuery.setFirstResult(page.index * page.size);
             jpaQuery.setMaxResults(page.size);
+        } else {
+            // Use deprecated API in org.hibernate.Query that will be moved to org.hibernate.query.Query on Hibernate 6.0
+            RowSelection options = jpaQuery.unwrap(org.hibernate.query.Query.class).getQueryOptions();
+            options.setFirstRow(null);
+            options.setMaxRows(null);
         }
     }
 
     private void manageOffsets(int maxResults) {
         if (range != null) {
             jpaQuery.setFirstResult(range.getStartIndex());
-            jpaQuery.setMaxResults(maxResults);
-        } else {
+        } else if (page != null) {
             jpaQuery.setFirstResult(page.index * page.size);
-            jpaQuery.setMaxResults(maxResults);
+        } else {
+            // Use deprecated API in org.hibernate.Query that will be moved to org.hibernate.query.Query on Hibernate 6.0
+            RowSelection options = jpaQuery.unwrap(org.hibernate.query.Query.class).getQueryOptions();
+            options.setFirstRow(null);
         }
+        jpaQuery.setMaxResults(maxResults);
     }
 }
--- a/integration-tests/hibernate-orm-panache/pom.xml
+++ b/integration-tests/hibernate-orm-panache/pom.xml
@@ -64,6 +64,11 @@
             <scope>test</scope>
         </dependency>
         <dependency>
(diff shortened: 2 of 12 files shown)


robmv:
The comment for the `page` field was removed because it doesn't apply any more, it is stored now because there are two states, by page and range, so both are stored.

I don't have an installation of MongoDB for testing but the same changes where done for it, and optimizing the query to limit 1 or 2 entries when using `firstResult*()` and `singleResult*()`

One extra question, currently you can switch between paging and range states, but it isn't possible to switch to non paging and non range states, unless you set null the page and the range, but that is an implementation leftover. Should an API for this be added, or document that page and range can be set to null, or add null checks to these APIs? (for another patch)

robmv:
Force updated because mixed accounts for the author of the commit

extensions/panache/hibernate-orm-panache/runtime/src/main/java/io/quarkus/hibernate/orm/panache/runtime/PanacheQueryImpl.java
```diff
@@ -100,11 +97,15 @@ public int pageCount() {
...
+        checkPagination();
         return page;
     }
 
-    private void checkNotInRange() {
+    private void checkPagination() {
+        if (page == null) {
+            throw new UnsupportedOperationException("Cannot call a page related method, " +
+                    "call page(Page) or page(int, int) to initiate pagination first");
+        }
```
FroMage: Well that's a change in behaviour: we used to have a single page encompassing all results. Perhaps it's fine, but at the very least, this has to be mentioned in the javadocs for the `PanacheQuery` interface.

FroMage (changes requested):
Thanks. We need some javadoc to indicate why the page methods can throw now that we don't have a default page.
We also need some tests to make sure we don't set paging info by default, though I'm not quite sure how to test that. Perhaps it's not possible.

extensions/panache/hibernate-orm-panache/runtime/src/main/java/io/quarkus/hibernate/orm/panache/runtime/PanacheQueryImpl.java
robmv: Javadoc updated, Noticed too that there is another interface `ReactivePanacheQuery` that has the same behaviour, so I updated it too so all three of them match.

robmv:
Just a of topic question, why these interfaces are defined this way?

```java
public interface PanacheQuery<Entity> {
    public <T extends Entity> PanacheQuery<T> page(Page page);
    ...
}
```

instead of:

```java
public interface PanacheQuery<Entity> {
    public PanacheQuery<Entity> page(Page page);
    ...
}
```

Why these methods have to change the returned generic from `PanacheQuery<Entity>` to `PanacheQuery<T extends Entity>`?. The only method that should do that is `project()`, all others do not modify the target entity. Am I missing something?

robmv:
> We also need some tests to make sure we don't set paging info by default, though I'm not quite sure how to test that. Perhaps it's not possible.

I agree that testing it will be difficult without going inside the wrapped APIs (JPA, MongoDB). The only way I think it is possible externally is to have a table with more than 2^31 rows, not very practical.

FroMage:
In RESTEasy they do some testing by examining the logs. Do you know if we can write a test that checks some logs @geoand @stuartwdouglas @dmlloyd ?

geoand:
The `QuarkusProdModeTest` does provide that ability:

For build logs see: https://github.com/quarkusio/quarkus/blob/master/integration-tests/spring-data-jpa/src/test/java/io/quarkus/it/spring/data/jpa/PropertyWarningsPMT.java#L32

For logs of the running application: https://github.com/quarkusio/quarkus/blob/master/integration-tests/kubernetes/standard/src/test/java/io/quarkus/it/kubernetes/KubernetesWithHealthTest.java#L51

FroMage:
Indeed in `QuarkusProdModeTest` you can get the `ProdModeTestResults.retainedBuildLogRecords` (you can inject `ProdModeTestResults`) and you can set `QuarkusProdModeTest.setLogRecordPredicate` to mark the hibernate SQL logs as being recorded so you can later obtain them.

FroMage:
Thanks @geoand !

geoand:
You are welcome @FroMage :)

robmv:
Thanks @geoand. @FroMage: I will check the suggestions and add test in the following days.

robmv:
Rebased in order to pull changes to ReactivePanacheQueryImpl on master.

I added a test based on the example `KubernetesWithHealthTest`. As the ProdModeTest run on another JVM I have to add the `prod-mode` surefire plugin, because there is already running and H2 server on normal tests and both conflicts.

I have to add  temporary workaround on `NoPagingPMT` to start the H2 server, because not matter what I do with the `QuarkusProdModeTest` instance, It doesn't start. I tried adding `H2DatabaseTestResource.class` with `addClasses()`  (Note that `TestResources.class` already has the `H2DatabaseTestResource` annotation and is added). Tried adding the artifact `quarkus-test-h2` with `setForcedDependencies()`. No luck. Any ideas?, maybe @geoand could help.

gsmet:
@robmv could you squash your commits into one?

@FroMage @geoand what's the status of this?

geoand:
It slipped through the cracks for me... Sorry.
I'll see if I can answer the question tomorrow

geoand:
@robmv I am not sure I understand the issue you are facing. Can you please elaborate?

integration-tests/hibernate-orm-panache/src/test/java/io/quarkus/it/panache/NoPagingPMT.java
```diff
@@ -0,0 +1,92 @@
...
+/**
+ * Test if PanacheQuery is using unnecessary SQL for limiting the number of output rows, the log output is tested for the
+ * presence of <code>offset</code> or <code>limit</code> in the SQL.
+ */
+public class NoPagingPMT {
+
+    static {
+        // WORKAROUND
+        try {
+            Server.createTcpServer().start();
```
FroMage, line 34: This worries me a bit, why is it needed and where is it stopped?

integration-tests/hibernate-orm-panache/pom.xml
```diff
@@ -109,6 +119,30 @@
...
                 </executions>
             </plugin>
+            <plugin>
+                <artifactId>maven-surefire-plugin</artifactId>
+                <executions>
+                    <!--
+                    The prod mode tests need to be part of a different execution to ensure that they don't mess with the standard tests.
+                    By adding this configuration we ensure that the maven surefire plugin will execute twice, one for the regular **/*Test.java
+                    tests (using the 'default-test' execution), and one for the prod mode tests (this 'prod-mode' execution)
+                    -->
```
FroMage, line 129: Are we sure we need to split the test runs in two parts?

FroMage (changes requested):
Sorry I took so long to review. I added some questions.

integration-tests/hibernate-orm-panache/pom.xml
geoand: Yes, this is needed because of classpath issues - I've encountered it before.

integration-tests/hibernate-orm-panache/src/test/java/io/quarkus/it/panache/NoPagingPMT.java
robmv: As ProdModeTest build a small JAR and run the test on a different JVM, There is a need to start the H2 Database server, My question to @geoand that I will detail later is about this, The workaround to start it manually is because the ProdModeTest isn't starting it even when I am adding `TestResources` that has the `@QuarkusTestResource(H2DatabaseTestResource.class)` annotation

As a temporary workaround didn't closed it myself because it is closed when that specific test JVM is closed.

geoand: Yeah, the prod mode test doesn't handle resources...

integration-tests/hibernate-orm-panache/pom.xml
robmv: and in this case it wasn't for the classpath but because the regular tests already start an H2 database instance, and the *PMT test run on its own JVM, so there are TCP port conflicts.

geoand: Right, good find

integration-tests/hibernate-orm-panache/src/test/java/io/quarkus/it/panache/NoPagingPMT.java
geoand: We could make it handle them for sure, but currently it doesn't.

robmv:
> @robmv I am not sure I understand the issue you are facing. Can you please elaborate?

@geoand The test needs an H2 Database running. I am packing everything that is needed to run the test:

```java
.setArchiveProducer(() -> ShrinkWrap.create(JavaArchive.class)
        .addClasses(PageItem.class, TestResources.class, NoPagingTestEndpoint.class))
```

`TestResources` has the `@QuarkusTestResource(H2DatabaseTestResource.class)` annotation but it doesn't automatically start the database, so I added the workaround to start it manually before `QuarkusProdModeTest` is instantiated. I want to fix the need for the workaround to start the H2 database explicitly.

geoand:
Thanks for the explanation @robmv.

I'll try and get `@QuarkusTestResource` support for the ProdModeTest soon (hopefully this week).

robmv:
> Thanks for the explanation @robmv.
> 
> I'll try and get `@QuarkusTestResource` support for the ProdModeTest soon (hopefully this week).

Great @geoand, I will wait for it, meanwhile I will squash everything and rebase and push that when that lands. Thanks

geoand:
#8597 should take care of the `@QuarkusTestResource` use.

robmv:
Squashed and workaround removed

geoand:

FroMage (approved):
LGTM, thanks!

FroMage:
@gsmet this will make it in 1.5, right, not 1.4?

gsmet:
Yeah, I would prefer it being part of a CR so more in 1.5.

I could be convinced though if someone had a strong opinion about it being needed for 1.4.

FroMage:
I think it's a cosmetic issue, only for Oracle, so it can probably wait.

gsmet:
In the end I decided to backport it as we have another PR that needs it.

PR #8169, opened 2020-03-25
Add a WebJar Locator extension that duplicates webjar locator behavior from Spring Boot

Christopher-Chianelli:
Currently, if you use webjars in your application, you need to specify the version in the URL. Spring-Boot allows you to not specify the webjar version in the URL by adding the `webjar-locator` dependency. I propose we replicate this functionality with a Quarkus extension.

 Side note: is it WebJar Locator or Webjars Locator? The artifactId and name from Maven Central is inconsistent (see https://mvnrepository.com/artifact/org.webjars/webjars-locator/0.39 and https://github.com/webjars/webjars-locator-core/blob/master/pom.xml).

To add: tests (which I guess would be testing the handler for given root paths and webjar name to version map)

### DIFF
--- a/docs/src/main/asciidoc/http-reference.adoc
+++ b/docs/src/main/asciidoc/http-reference.adoc
@@ -29,6 +29,29 @@ was chosen as it is the standard location for resources in `jar` files as define
 Quarkus can be used without Servlet following this convention allows existing code that places its resources in this
 location to function correctly.
 
+=== WebJar Locator Support
+
+If you are using webjars, like the following JQuery one
+[source, xml]
+----
+<dependency>
+    <groupId>org.webjars</groupId>
+    <artifactId>jquery</artifactId>
+    <version>3.1.1</version>
+</dependency>
+----
+and rather write `/webjars/jquery/jquery.min.js` instead of `/webjars/jquery/3.1.1/jquery.min.js`
+in your HTML files, you can add the `quarkus-webjars-locator` extension to your project.
+To use it, add the following to your project's dependencies:
+
+[source, xml]
+----
+<dependency>
+    <groupId>io.quarkus</groupId>
+    <artifactId>quarkus-webjars-locator</artifactId>
+</dependency>
+----
+
 == Configuring the Context path
 
 By default Quarkus will serve content from under the root context. If you want to change this you can use the
--- a/extensions/webjars-locator/deployment/pom.xml
+++ b/extensions/webjars-locator/deployment/pom.xml
@@ -15,10 +15,6 @@
 
     <dependencies>
         <dependency>
-            <groupId>org.webjars</groupId>
-            <artifactId>webjars-locator</artifactId>
-        </dependency>
-        <dependency>
             <groupId>io.quarkus</groupId>
             <artifactId>quarkus-core-deployment</artifactId>
         </dependency>
--- a/extensions/webjars-locator/deployment/pom.xml
+++ b/extensions/webjars-locator/deployment/pom.xml
@@ -1,0 +1,84 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <artifactId>quarkus-webjars-locator-parent</artifactId>
+        <groupId>io.quarkus</groupId>
+        <version>999-SNAPSHOT</version>
+        <relativePath>../</relativePath>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>quarkus-webjars-locator-deployment</artifactId>
+    <name>Quarkus - WebJar Locator - Deployment</name>
+
+    <dependencies>
+        <dependency>
+            <groupId>org.webjars</groupId>
+            <artifactId>webjars-locator</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-core-deployment</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-vertx-http-deployment</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-webjars-locator</artifactId>
+        </dependency>
+        
+        <!-- Tests -->
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-junit5-internal</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>io.rest-assured</groupId>
+            <artifactId>rest-assured</artifactId>
+            <scope>test</scope>
+        </dependency>
+        
+        <!-- Using setForcedDependencies only works if the dependency is in the pom's test scope. -->
+        <dependency>
+          <groupId>org.webjars</groupId>
+          <artifactId>jquery</artifactId>
+          <version>3.4.1</version>
+          <scope>test</scope>
+        </dependency>
+        <dependency>
+          <groupId>org.webjars</groupId>
+          <artifactId>momentjs</artifactId>
+          <version>2.24.0</version>
+          <scope>test</scope>
+        </dependency>
+        
+        <!-- DevMode has no setForcedDependencies, so need to add resteasy extension here -->
+        <dependency>
+          <groupId>io.quarkus</groupId>
+          <artifactId>quarkus-resteasy</artifactId>
+        </dependency>
+    </dependencies>
+
+    <build>
+        <plugins>
+            <plugin>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <configuration>
+                    <annotationProcessorPaths>
+                        <path>
+                            <groupId>io.quarkus</groupId>
+                            <artifactId>quarkus-extension-processor</artifactId>
+                            <version>${project.version}</version>
+                        </path>
+                    </annotationProcessorPaths>
+                </configuration>
+            </plugin>
+        </plugins>
+    </build>
+
+</project>
--- a/extensions/webjars-locator/deployment/pom.xml
+++ b/extensions/webjars-locator/deployment/pom.xml
@@ -11,7 +11,7 @@
     <modelVersion>4.0.0</modelVersion>
 
     <artifactId>quarkus-webjars-locator-deployment</artifactId>
-    <name>Quarkus - WebJar Locator - Deployment</name>
+    <name>Quarkus - WebJars Locator - Deployment</name>
 
     <dependencies>
         <dependency>
@@ -30,7 +30,7 @@
             <groupId>io.quarkus</groupId>
             <artifactId>quarkus-webjars-locator</artifactId>
         </dependency>
-        
+
         <!-- Tests -->
         <dependency>
             <groupId>io.quarkus</groupId>
@@ -42,25 +42,24 @@
             <artifactId>rest-assured</artifactId>
             <scope>test</scope>
         </dependency>
-        
-        <!-- Using setForcedDependencies only works if the dependency is in the pom's test scope. -->
+
         <dependency>
-          <groupId>org.webjars</groupId>
-          <artifactId>jquery</artifactId>
-          <version>3.4.1</version>
-          <scope>test</scope>
+            <groupId>org.webjars</groupId>
+            <artifactId>jquery</artifactId>
+            <version>3.4.1</version>
+            <scope>test</scope>
         </dependency>
         <dependency>
-          <groupId>org.webjars</groupId>
-          <artifactId>momentjs</artifactId>
-          <version>2.24.0</version>
-          <scope>test</scope>
+            <groupId>org.webjars</groupId>
+            <artifactId>momentjs</artifactId>
+            <version>2.24.0</version>
+            <scope>test</scope>
         </dependency>
-        
-        <!-- DevMode has no setForcedDependencies, so need to add resteasy extension here -->
+
         <dependency>
-          <groupId>io.quarkus</groupId>
-          <artifactId>quarkus-resteasy</artifactId>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-resteasy-deployment</artifactId>
+            <scope>test</scope>
         </dependency>
     </dependencies>
 
--- a/extensions/webjars-locator/deployment/src/main/java/io/quarkus/webjar/locator/deployment/WebJarLocatorStandaloneBuildStep.java
+++ b/extensions/webjars-locator/deployment/src/main/java/io/quarkus/webjar/locator/deployment/WebJarLocatorStandaloneBuildStep.java
@@ -1,0 +1,42 @@
+package io.quarkus.webjar.locator.deployment;
+
+import java.util.Map;
+
+import org.webjars.WebJarAssetLocator;
+
+import io.quarkus.deployment.annotations.BuildProducer;
+import io.quarkus.deployment.annotations.BuildStep;
+import io.quarkus.deployment.annotations.ExecutionTime;
+import io.quarkus.deployment.annotations.Record;
+import io.quarkus.deployment.builditem.FeatureBuildItem;
+import io.quarkus.vertx.http.deployment.RouteBuildItem;
+import io.quarkus.vertx.http.runtime.HttpBuildTimeConfig;
+import io.quarkus.webjar.locator.runtime.WebJarLocatorRecorder;
+
+public class WebJarLocatorStandaloneBuildStep {
+
+    @BuildStep
+    @Record(ExecutionTime.RUNTIME_INIT)
+    public void findWebjarsAndCreateHandler(
+            HttpBuildTimeConfig httpConfig,
+            BuildProducer<FeatureBuildItem> feature,
+            BuildProducer<RouteBuildItem> routes,
+            WebJarLocatorRecorder recorder) throws Exception {
+
+        WebJarAssetLocator webJarLocator = new WebJarAssetLocator();
+        Map<String, String> webjarNameToVersionMap = webJarLocator.getWebJars();
+
+        if (!webjarNameToVersionMap.isEmpty()) {
+            // The context path + the resources path
+            String rootPath = httpConfig.rootPath;
+            String webjarRootPath = (rootPath.endsWith("/")) ? rootPath + "webjars/" : rootPath + "/webjars/";
+            feature.produce(new FeatureBuildItem(FeatureBuildItem.WEBJARS_LOCATOR));
+            routes.produce(
+                    new RouteBuildItem(webjarRootPath + "*",
+                            recorder.getHandler(webjarRootPath, webjarNameToVersionMap),
+                            false));
+        }
+
+    }
+
+}
--- a/extensions/webjars-locator/deployment/src/test/java/io/quarkus/webjar/locator/test/WebJarLocatorDevModeTest.java
+++ b/extensions/webjars-locator/deployment/src/test/java/io/quarkus/webjar/locator/test/WebJarLocatorDevModeTest.java
@@ -1,0 +1,109 @@
+package io.quarkus.webjar.locator.test;
+
+import static org.hamcrest.core.Is.is;
+
+import org.hamcrest.Matchers;
+import org.jboss.shrinkwrap.api.ShrinkWrap;
+import org.jboss.shrinkwrap.api.asset.StringAsset;
+import org.jboss.shrinkwrap.api.spec.JavaArchive;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.RegisterExtension;
+
+import io.quarkus.test.QuarkusDevModeTest;
+import io.restassured.RestAssured;
+
+public class WebJarLocatorDevModeTest {
+    private static final String META_INF_RESOURCES = "META-INF/resources/";
+
+    @RegisterExtension
+    static QuarkusDevModeTest test = new QuarkusDevModeTest()
+            .setArchiveProducer(() -> ShrinkWrap.create(JavaArchive.class)
+                    .addClass(PostResource.class)
+                    .addAsResource(new StringAsset("<html>Hello!<html>"), META_INF_RESOURCES + "/index.html")
+                    .addAsResource(new StringAsset("Test"), META_INF_RESOURCES + "/some/path/test.txt"));
+
+    @Test
+    public void testDevMode() {
+        // Test Endpoint
+        RestAssured.given().body("Stuart")
+                .when()
+                .post("/post")
+                .then()
+                .body(Matchers.equalTo("Hello: Stuart"));
+        // Test normal files
+        RestAssured.get("/").then()
+                .statusCode(200)
+                .body(is("<html>Hello!<html>"));
+
+        RestAssured.get("/index.html").then()
+                .statusCode(200)
+                .body(is("<html>Hello!<html>"));
+
+        RestAssured.get("/some/path/test.txt").then()
+                .statusCode(200)
+                .body(is("Test"));
+
+        // Test Existing Web Jars
+        RestAssured.get("/webjars/jquery/jquery.min.js").then()
+                .statusCode(200);
+        RestAssured.get("/webjars/momentjs/min/moment.min.js").then()
+                .statusCode(200);
+
+        // Test using version in url of existing Web Jar
+        RestAssured.get("/webjars/jquery/3.4.1/jquery.min.js").then()
+                .statusCode(200);
+        RestAssured.get("/webjars/momentjs/2.24.0/min/moment.min.js").then()
+                .statusCode(200);
+
+        // Test non-existing Web Jar
+        RestAssured.get("/webjars/bootstrap/js/bootstrap.min.js").then()
+                .statusCode(404);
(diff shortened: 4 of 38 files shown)


extensions/webjars-locator/deployment/pom.xml
```diff
@@ -0,0 +1,59 @@
...
+        </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-junit5-internal</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-webjars-locator</artifactId>
+            <version>999-SNAPSHOT</version>
```
stuartwdouglas: This should be managed from the runtime bom (and be project.version).

extensions/webjars-locator/deployment/pom.xml
```diff
@@ -0,0 +1,59 @@
...
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>quarkus-webjars-locator-deployment</artifactId>
+    <name>Quarkus - WebJar Locator - Deployment</name>
+
+    <dependencies>
+        <dependency>
+            <groupId>org.webjars</groupId>
+            <artifactId>webjars-locator</artifactId>
+            <version>0.39</version>
```
stuartwdouglas: This should be managed from the deployment bom.

extensions/webjars-locator/runtime/src/main/resources/META-INF/quarkus-extension.yaml
```diff
@@ -0,0 +1,11 @@
+---
+name: "WebJar Locator"
+metadata:
+  short-name: "webjars-locator"
+  keywords:
+  - "web"
+  - "webjar"
+  guide: "https://quarkus.io/guides/rest-json"
```
stuartwdouglas: This guide should be empty if there is no guide for this, although it would be great if you could add a note about this to the HTTP reference instead.

extensions/webjars-locator/deployment/src/main/java/io/quarkus/webjar/locator/deployment/WebJarLocatorStandaloneBuildStep.java
```diff
@@ -0,0 +1,58 @@
...
+import io.quarkus.vertx.http.runtime.HttpBuildTimeConfig;
+import io.quarkus.webjar.locator.runtime.WebJarLocatorHandler;
+
+public class WebJarLocatorStandaloneBuildStep {
+
+    @BuildStep()
+    public void staticInit(Capabilities capabilities,
+            HttpBuildTimeConfig httpConfig,
+            BuildProducer<WebJarLocatorConfigBuildItem> standalone) throws Exception {
+        if (capabilities.isCapabilityPresent(Capabilities.SERVLET)) {
```
stuartwdouglas: Why is this not needed for Servlet?

extensions/webjars-locator/deployment/src/main/java/io/quarkus/webjar/locator/deployment/WebJarLocatorStandaloneBuildStep.java
```diff
@@ -0,0 +1,58 @@
...
+            // The context path + the resources path
+            String rootPath = httpConfig.rootPath;
+            String webjarRootPath = (rootPath.endsWith("/")) ? rootPath + "webjars/" : rootPath + "/webjars/";
+            standalone.produce(new WebJarLocatorConfigBuildItem(webjarRootPath, webjarNameToVersionMap));
+        }
+    }
+
+    @BuildStep
+    @Record(ExecutionTime.RUNTIME_INIT)
+    public void boot(
```
stuartwdouglas: I don't think you actually need WebJarLocatorConfigBuildItem, it looks like this could just be a single build step?

stuartwdouglas (changes requested):

extensions/webjars-locator/deployment/src/main/java/io/quarkus/webjar/locator/deployment/WebJarLocatorStandaloneBuildStep.java
geoand: Can you please elaborate on this? I am sure you have some good technical reason in mind, but I am also sure that Quarkus has a lot up it's sleeve to lift any such restrictions

extensions/webjars-locator/deployment/src/main/java/io/quarkus/webjar/locator/deployment/WebJarLocatorStandaloneBuildStep.java
```diff
@@ -0,0 +1,58 @@
...
+import io.quarkus.deployment.annotations.Record;
+import io.quarkus.deployment.builditem.FeatureBuildItem;
+import io.quarkus.vertx.http.deployment.RouteBuildItem;
+import io.quarkus.vertx.http.runtime.HttpBuildTimeConfig;
+import io.quarkus.webjar.locator.runtime.WebJarLocatorHandler;
+
+public class WebJarLocatorStandaloneBuildStep {
+
+    @BuildStep()
+    public void staticInit(Capabilities capabilities,
```
geoand: The name doesn't seem really fitting :)

geoand:
@Christopher-Chianelli Thanks for the contribution!

When adding tests for this, it is important to keep in mind that we will need tests for native mode and also for dev mode to ensure that all our bases are covered.

geoand:
It seems that webjars uses https://github.com/classgraph/classgraph, so I am thinking that we definitely need to see how it's usage would affect the time of dev-mode restart.

extensions/webjars-locator/deployment/pom.xml
ge0ffrey: see https://github.com/quarkusio/quarkus/blob/master/bom/runtime/pom.xml

extensions/webjars-locator/deployment/src/main/java/io/quarkus/webjar/locator/deployment/WebJarLocatorStandaloneBuildStep.java
ge0ffrey: Isn't the reverse needed? if servlet compability is _not_ present, return?

extensions/webjars-locator/runtime/src/main/resources/META-INF/quarkus-extension.yaml
ge0ffrey: @Christopher-Chianelli How about adding a really short guide for webjars here: https://github.com/quarkusio/quarkus/tree/master/docs/src/main/asciidoc ?

Also, an integration test here to prove it works natively would be great:
https://github.com/quarkusio/quarkus/tree/master/integration-tests

extensions/webjars-locator/deployment/src/main/java/io/quarkus/webjar/locator/deployment/WebJarLocatorStandaloneBuildStep.java
Christopher-Chianelli: I based this extension on the resteasy-extension: https://github.com/quarkusio/quarkus/blob/0d29a66fe75b43dc9914c738e4ceb876272d0d74/extensions/resteasy/deployment/src/main/java/io/quarkus/resteasy/deployment/ResteasyStandaloneBuildStep.java#L62-L103. I figured if `RESTeasy` has it for a reason, this extension will probably have it for the same reason. It may be entirely unneeded.

extensions/webjars-locator/deployment/src/main/java/io/quarkus/webjar/locator/deployment/WebJarLocatorStandaloneBuildStep.java
Christopher-Chianelli: Will there be any performance downside though? Or are all BuildStep (regardless if recorded or not) rerun at runtime? (the idea behind having it in two separate methods is to put the possibly computational expensive/classpath scanning at buildtime, and use the saved Map when we start the application)

extensions/webjars-locator/deployment/src/main/java/io/quarkus/webjar/locator/deployment/WebJarLocatorStandaloneBuildStep.java
Christopher-Chianelli: The name was copied from resteasy extension. How about `findWebjarsAndRootPath`?

extensions/webjars-locator/deployment/src/main/java/io/quarkus/webjar/locator/deployment/WebJarLocatorStandaloneBuildStep.java
geoand: Build Steps are build-time only. What gets run at runtime are the recorders (that are associated with a build step).
So there is no reason to have different build steps for what you are doing.

extensions/webjars-locator/deployment/src/main/java/io/quarkus/webjar/locator/deployment/WebJarLocatorStandaloneBuildStep.java
geoand: Sure :)

Christopher-Chianelli:
Note: Test fails with
```
Caused by: java.lang.Error: Unresolved compilation problem: 
	The type sun.misc.Unsafe cannot be resolved. It is indirectly referenced from required .class files

	at io.quarkus.runtime.ExecutorRecorder.<init>(ExecutorRecorder.java:1)
	at io.quarkus.deployment.steps.ThreadPoolSetup$createExecutor7.deploy_0(ThreadPoolSetup$createExecutor7.zig:59)
	at io.quarkus.deployment.steps.ThreadPoolSetup$createExecutor7.deploy(ThreadPoolSetup$createExecutor7.zig:36)
	at io.quarkus.runner.ApplicationImpl.doStart(ApplicationImpl.zig:106)
	... 44 more
```
For me. Maybe `webjar-locator` uses it somewhere (meaning we'll need to implement it main functionality without it)

geoand:
> Note: Test fails with
> 
> ```
> Caused by: java.lang.Error: Unresolved compilation problem: 
> 	The type sun.misc.Unsafe cannot be resolved. It is indirectly referenced from required .class files
> 
> 	at io.quarkus.runtime.ExecutorRecorder.<init>(ExecutorRecorder.java:1)
> 	at io.quarkus.deployment.steps.ThreadPoolSetup$createExecutor7.deploy_0(ThreadPoolSetup$createExecutor7.zig:59)
> 	at io.quarkus.deployment.steps.ThreadPoolSetup$createExecutor7.deploy(ThreadPoolSetup$createExecutor7.zig:36)
> 	at io.quarkus.runner.ApplicationImpl.doStart(ApplicationImpl.zig:106)
> 	... 44 more
> ```
> 
> For me. Maybe `webjar-locator` uses it somewhere (meaning we'll need to implement it main functionality without it)

Did it fail with in a JVM test or native test?

Christopher-Chianelli:
@geoand I think it failed with a JVM test (ran from maven, fails on both Java 11 and Java 8); the test use `QuarkusUnitTest`.

geoand:
Thanks for the information @Christopher-Chianelli. Yes that would be a JVM mode test then that failed.

Christopher-Chianelli:
Hmm... I think my IDE (Flatpak Eclipse) might be doing weird magic behind the scenes. If I build Quarkus from scratch using maven, I get an expectation error (which means it was able to start Quarkus). If I open my IDE (without doing any changes), I get `java.lang.RuntimeException: java.lang.RuntimeException: java.lang.RuntimeException: Failed to load steps from class io.quarkus.vertx.http.deployment.HttpSecurityProcessor`. So look like Eclipse's build seems break Quarkus and corrupts my .m2?

Christopher-Chianelli:
Update: `quarkus-webjar-locator` unit test now actually works. Turns out you need to add forced dependencies into `pom.xml` test's scope for webjars. Also Eclipse automatic build can be troublesome (as in, breaks Quarkus). Working on native test now.

Christopher-Chianelli:
Update: Add integration and native test for `quarkus-webjar-locator`.

Christopher-Chianelli:
Update: Add `dev-mode` test for the `webjars-locator` extension.
Notes:

- The extension does not provide web server capabilities, and is reliant on there being
an extension that add web server capabilities. I use resteasy in the tests, but it *should*
work with any extension that adds webserver capabilities.

Christopher-Chianelli:
Java 8 build failed at `quarkus-smallrye-reactive-messaging-amqp-deployment`:
`AmqpDevModeTest.testCodeUpdate » Composite 2 exceptions occurred.` 

This appears to be the root cause:
```
2020-04-14 19:05:22,446 INFO  [io.sma.rea.mes.ext.MediatorManager] (Quarkus Main Thread) Cancel subscriptions
2020-04-14 19:05:22,449 INFO  [io.quarkus] (Quarkus Main Thread) Quarkus stopped in 0.053s
java.lang.IllegalStateException: Arc context propagation was attempted but the container is not running.
	at io.quarkus.arc.runtime.context.ArcContextProvider.lambda$currentContext$2(ArcContextProvider.java:35)
	at io.smallrye.context.impl.ActiveContextState.<init>(ActiveContextState.java:18)
	at io.smallrye.context.impl.CapturedContextState.begin(CapturedContextState.java:35)
	at io.smallrye.context.SmallRyeThreadContext.lambda$withContext$0(SmallRyeThreadContext.java:213)
	at io.smallrye.context.propagators.rxjava2.ContextPropagatorOnFlowableCreateAction$ContextCapturerFlowable.onNext(ContextPropagatorOnFlowableCreateAction.java:50)
	at io.reactivex.internal.operators.flowable.FlowableFlatMap$MergeSubscriber.drainLoop(FlowableFlatMap.java:506)
```

Doesn't appear related to the PR.

geoand:
Can you do me a favor and rebase onto master?

Christopher-Chianelli:
Sure, (I prefer merging master back into the branch since rebases need to be forced-push, cause comments to go missing, amd mess up history). One sec

Christopher-Chianelli:
@geoand Done

geoand:
@ia3andy would you like to review this as well?

extensions/webjars-locator/deployment/src/test/java/io/quarkus/webjar/locator/test/WebJarLocatorDevModeTest.java
```diff
@@ -0,0 +1,58 @@
...
+import org.jboss.shrinkwrap.api.ShrinkWrap;
+import org.jboss.shrinkwrap.api.asset.StringAsset;
+import org.jboss.shrinkwrap.api.spec.JavaArchive;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.RegisterExtension;
+
+import io.quarkus.test.QuarkusDevModeTest;
+import io.restassured.RestAssured;
+
+public class WebJarLocatorDevModeTest {
```
geoand, line 14: Actually I think this test doesn't actually test dev mode changes (unless I am missing something). The point of the `QuarkusDevModeTest` is to make it easy to change application code and/or configuration and make sure the application works properly with the new changes.

Does that make sense?

docs/src/main/asciidoc/http-reference.adoc
```diff
@@ -29,6 +29,20 @@ was chosen as it is the standard location for resources in `jar` files as define
...
+If you add the `quarkus-webjars-locator` extension to your project, you do not need to specify version
+for webjars. That is, a URL like `/webjars/jquery/3.1.1/jquery.min.js` becomes `/webjars/jquery/jquery.min.js` in your
+HTML files. To use it, add the following to your project's dependencies:
+
+[source, xml]
+----
+<dependency>
+    <groupId>io.quarkus</groupId>
+    <artifactId>quarkus-webjars-locator</artifactId>
+</dependency>
```
ge0ffrey, line 43: I'd also add the jquery webjar itself here in the pom example too, for illustration:

```
    <dependency>
      <groupId>org.webjars</groupId>
      <artifactId>jquery</artifactId>
    </dependency>
```

ge0ffrey (approved):

extensions/webjars-locator/deployment/src/test/java/io/quarkus/webjar/locator/test/WebJarLocatorDevModeTest.java
Christopher-Chianelli: True, issue is you cannot change dependencies for `QuarkusDevModeTest` (or if there is I couldn't find it), which is basically the only way of modifying the behaviour of this extension (ex: adding the jquery webjar dependency, removing the moment webjar dependency, etc. would be the ideal dev mode test).

Christopher-Chianelli: This test as it currently is basically make sure nothing odd happens when you run in dev mode.

geoand: I understand that, but we need to make sure that in an application that contains the extension, restart works properly.

So I propose you do the following: add a dummy JAX-RS resource (like other dev mode tests do), and change a string value in it.
Then hit the endpoints again to ensure that everything still works after the restart.

geoand:
Thanks a lot for this work @Christopher-Chianelli!

Mind squashing the commits into a single one?

Christopher-Chianelli:
@geoand Done!

geoand:
@stuartwdouglas your concerns seem to be addressed, right?

extensions/webjars-locator/runtime/src/main/java/io/quarkus/webjar/locator/runtime/WebJarLocatorHandler.java
```diff
@@ -0,0 +1,55 @@
...
+import java.util.Map;
+
+import io.quarkus.runtime.annotations.Recorder;
+import io.vertx.core.Handler;
+import io.vertx.ext.web.RoutingContext;
+
+@Recorder
+public class WebJarLocatorHandler {
+
+    private String webjarsRootUrl;
```
stuartwdouglas, line 12: These should be parameters on the getHandler() method, recorders are not really supposed to be stateful.

extensions/webjars-locator/runtime/src/main/java/io/quarkus/webjar/locator/runtime/WebJarLocatorHandler.java
```diff
@@ -0,0 +1,55 @@
+package io.quarkus.webjar.locator.runtime;
+
+import java.util.Map;
+
+import io.quarkus.runtime.annotations.Recorder;
+import io.vertx.core.Handler;
+import io.vertx.ext.web.RoutingContext;
+
+@Recorder
+public class WebJarLocatorHandler {
```
stuartwdouglas, line 10: This should be WebJarLocatorRecorder to be in keeping with the rest of the code base

stuartwdouglas (changes requested):
There is a little bit of minor stuff that I missed the first time around, otherwise looks good.

Christopher-Chianelli:
@stuartwdouglas Merged your suggested changes.

stuartwdouglas (approved):

geoand:
Sorry, can you please rebase the Pull Request onto master and fix the conflict? Once that is done we can merge it.

geoand (approved):

gsmet:
Found a few minor infelicities (mostly formatting issues). I will amend the commit, rebase, fix the conflict and get this in.

Please don't merge :).

geoand:
Seems like it needs a rebase now.

Christopher-Chianelli:
@gsmet are you going to rebase it or should I?

emmanuelbernard:
@Christopher-Chianelli If you don't, it's likely Guillaume will but it would be much appreciated if you do :)

Christopher-Chianelli:
@geoand , @emmanuelbernard rebased

geoand:
@gsmet did you end up making the changes you wanted to make?

gsmet:
I made all the necessary adjustments in a separate commit and also rebased everything.

In passing, I found out that there was no native tests for this feature and adding one make it fail. I get a 404. I tried adding the resources with `quarkus.native.resources.includes=META-INF/resources/webjars/**` in the application.properties but it didn't work (we would need a proper build item for that anyway but that was to test if it was just resources missing).

I don't have the time to debug that right now and see what's necessary so I let someone else take the lead on this.

Christopher-Chianelli:
@gsmet I will investigate

gsmet:
@stuartwdouglas so we have an issue with this classgraph library. Looks like it doesn't play well with our class loader infrastructure. While things work well when in tests, it doesn't work as soon as you build a real runner jar: it doesn't find the webjar.

I tried to push the TCCL to it but didn't work well: at least the behavior was consistent, it doesn't find anything in all cases :).

We can push the resources to Webjars Locator instead of using classgraph if we have it but I don't know if we have an efficient way to find the artifacts in which we have `META-INF/resources/webjars/` and then scan all the resources here without going through the jar entries. We don't have any classes, just plain resources.

Any idea pops up?

Wondering if we could mimic something similar to what is done with `AdditionalApplicationArchiveMarkerBuildItem`? Was also wondering if it could be something useful in general e.g. have the hability to declare a marker and a tag and be able to get all the archives with a given tag easily.

Note that I might miss something obvious and maybe we already have everything in place to do that!

Christopher-Chianelli:
I got it to work with the runner Jar but not the native image using this (and remove the version map):
```
 event.vertx().fileSystem().readDir("META-INF/resources/webjars/" + webjar.toString(), readDir -> {
    // Find version in file list, and use it to resolve path
});
```

When it is run as a native image, the directory is empty. When run as a runner jar, the directory has the version of the webjar.

stuartwdouglas:
I have changed this to just resolve the versions at build time.

Christopher-Chianelli:
@stuartwdouglas your changes LGTM; All tests passed in native mode.

gsmet:
The test failures are related. The deployment tests are failing with 404 with JDK 8. And on Windows too.

Christopher-Chianelli:
I added a commit that use File.separator and the Path API in order to get the Windows directory to see if that changes anything

Christopher-Chianelli:
Found out why Windows was failing: we were using
```
String resolvedPath = webjarsRootUrl + webjar.toString() + "/" +
                            version + "/" + webjar.relativize(rest).toString();`
```
for redirection, but `webjar.relativize(rest).toString()` use `\` instead of `/` on Windows. To fix this, I added this small method to do the `webjar.relativize(rest).toString()` bit:

```
    private static String convertToRelativeUrl(Path root, Path rest) {
        Path relativePath = root.relativize(rest);
        StringBuilder url = new StringBuilder();
        for (Path part : relativePath) {
            url.append("/");
            try {
                url.append(URLEncoder.encode(part.toString(), "UTF-8"));
            } catch (UnsupportedEncodingException e) {
                // Shouldn't happen, "UTF-8 should be available in all JVM's
                throw new IllegalStateException(e);
            }
        }
        return url.toString();
    }
```

Christopher-Chianelli:
@gsmet Summary of failed checks:

- Windows: Fails due to
```
[ERROR] Errors: 
[ERROR]   JarRunnerIT.testThatFastJarFormatWorks:106 � ConditionTimeout Condition with i...
```
WebJar Locator extensions tests passed; WebJar Locator IT was not run due to the above failure.

- Native Tests - Meassaging: Fails due to
```
 [INFO] Running io.quarkus.it.kafka.streams.KafkaStreamsITCase
[ERROR] Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 0.019 s <<< FAILURE! - in io.quarkus.it.kafka.streams.KafkaStreamsITCase
[ERROR] io.quarkus.it.kafka.streams.KafkaStreamsITCase  Time elapsed: 0.017 s  <<< ERROR!
org.junit.platform.commons.JUnitException: @Inject is not supported in NativeImageTest tests. Offending field is io.quarkus.it.kafka.streams.KafkaStreamsTest.stateListener

[INFO] 
[INFO] Results:
[INFO] 
[ERROR] Errors: 
[ERROR]   KafkaStreamsITCase » JUnit @Inject is not supported in NativeImageTest tests. ...
[INFO] 
[ERROR] Tests run: 1, Failures: 0, Errors: 1, Skipped: 0
```

- Native Tests - Misc 4 (which contains WebJar Locator native tests) was cancelled during the build phase of the executable (it passes on my machine; guess: OOM or something of the like) (the JAR version of the test ran and passed, native untested).

Situlation this leaves us in regards for this PR: non-native tests pass on JDK 8 and higher on Linux and Windows; native tests are untested (pass on my machine).

Christopher-Chianelli:
There might be an issue if JDK 8 can also create native images, as it uses a workaround that does not work with native images, but works with an executable JAR. If native images can only be created on JDK 11 and above, there should be no issues.

stuartwdouglas:
oops, I tried to fix the test failures but pushed the results to my branch out of habit, instead of this PR. I have created a new PR with my changes for testing: https://github.com/quarkusio/quarkus/pull/9392

Note that the windows path stuff should not matter, / is still a valid separator on windows.

geoand:
So I assume we should close this in favor of #9392 right?

stuartwdouglas:
Lets see what CI has to say, I did not notice there was a conflict so it did not run.

Christopher-Chianelli:
It does matter in URL's: what was happening was the url '/webjar/path/to/file' was being converted to '/webjar\path\to\file', which is a problem

Christopher-Chianelli:
If CI passes for #9392, I am in favour of closing this PR for  #9392

gsmet:
Superseded by https://github.com/quarkusio/quarkus/pull/9392 .

PR #8172, opened 2020-03-25
Remove unused classes

stuartwdouglas:


stuartwdouglas:
These should never have been added, they were part of the original new classloading branch, but should have been deleted before it was merged. They have never been documented and have no effect, so can be safely deleted.

geoand (approved):
I actually that those had been deleted :)

PR #8175, opened 2020-03-26
Support for mocking CDI beans

stuartwdouglas:
This allows all CDI normal scoped beans to be mocked at
test time by adding a form of redirection into the client
proxy.

### DIFF
--- a/docs/src/main/asciidoc/getting-started-testing.adoc
+++ b/docs/src/main/asciidoc/getting-started-testing.adoc
@@ -302,7 +302,11 @@ public class TestStereotypeTestCase {
 
 == Mock Support
 
-Quarkus supports the use of mock objects using the CDI `@Alternative` mechanism.
+Quarkus supports the use of mock objects using two different approaches. You can either use CDI alternatives to
+mock out a bean for all test classes, or use `QuarkusMock` to mock out beans on a per test basis.
+
+=== CDI `@Alternative` mechanism.
+
 To use this simply override the bean you wish to mock with a class in the `src/test/java` directory, and put the `@Alternative` and `@Priority(1)` annotations on the bean. 
 Alternatively, a convenient `io.quarkus.test.Mock` stereotype annotation could be used.
 This built-in stereotype declares `@Alternative`, `@Priority(1)` and `@Dependent`. 
@@ -341,6 +345,64 @@ it will take effect all the time, not just when testing.
 
 Note that at present this approach does not work with native image testing, as this would required the test alternatives
 to be baked into the native image.
+
+=== Mocking using QuarkusMock
+
+The `io.quarkus.test.junit.QuarkusMock` class can be used to temporarily mock out any normal scoped
+bean. If you use this method in a `@BeforeAll` method the mock will take effect for all tests on the current class,
+while if you use this in a test method the mock will only take effect for the duration of the current test.
+
+This method can be used for any normal scoped CDI bean (e.g. `@ApplicationScoped`, `@RequestScoped` etc, basically
+every scope except `@Singleton` and `@Dependent`).
+
+An example usage could look like:
+
+[source,java]
+----
+@QuarkusTest
+public class MockTestCase {
+
+    @Inject
+    MockableBean1 mockableBean1;
+
+    @Inject
+    MockableBean2 mockableBean2;
+
+    @BeforeAll
+    public static void setup() {
+        MockableBean1 mock = Mockito.mock(MockableBean1.class);
+        Mockito.when(mock.greet("Stuart")).thenReturn("A mock for Stuart");
+        QuarkusMock.installMockForType(mock, MockableBean1.class);  <1>
+    }
+
+    @Test
+    public void testBeforeAll() {
+        Assertions.assertEquals("A mock for Stuart", mockableBean1.greet("Stuart"));
+        Assertions.assertEquals("Hello Stuart", mockableBean2.greet("Stuart"));
+    }
+
+    @Test
+    public void testPerTestMock() {
+        QuarkusMock.installMockForInstance(new BonourGreeter(), mockableBean2); <2>
+        Assertions.assertEquals("A mock for Stuart", mockableBean1.greet("Stuart"));
+        Assertions.assertEquals("Bonjour Stuart", mockableBean2.greet("Stuart"));
+    }
+
+    public static class BonourGreeter extends MockableBean2 {
+        @Override
+        public String greet(String name) {
+            return "Bonjour " + name;
+        }
+    }
+
+}
+----
+<1> As the injected instance is not available here we use `installMockForType`, this mock is used for both test methods
+<2> We use `installMockForInstance` to replace the injected bean, this takes effect for the duration of the test method.
+
+Note that there is no dependency on Mockito, you can use any mocking library you like, or even manually override the
+objects to provide the behaviour you require.
+
 
 == Test Bootstrap Configuration Options
 
--- a/integration-tests/main/src/test/java/io/quarkus/it/main/MockTestCase.java
+++ b/integration-tests/main/src/test/java/io/quarkus/it/main/MockTestCase.java
@@ -1,0 +1,53 @@
+package io.quarkus.it.main;
+
+import javax.inject.Inject;
+
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+import org.mockito.Mockito;
+
+import io.quarkus.it.mock.MockableBean1;
+import io.quarkus.it.mock.MockableBean2;
+import io.quarkus.test.junit.QuarkusMock;
+import io.quarkus.test.junit.QuarkusTest;
+
+@QuarkusTest
+public class MockTestCase {
+
+    protected static final String BONJOUR = "Bonjour ";
+
+    @Inject
+    MockableBean1 mockableBean1;
+
+    @Inject
+    MockableBean2 mockableBean2;
+
+    @BeforeAll
+    public static void setup() {
+        MockableBean1 mock = Mockito.mock(MockableBean1.class);
+        Mockito.when(mock.greet("Stuart")).thenReturn("A mock for Stuart");
+        QuarkusMock.installMockForType(mock, MockableBean1.class);
+    }
+
+    @Test
+    public void testBeforeAll() {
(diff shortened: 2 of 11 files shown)


geoand:
This is +100 from me

integration-tests/main/src/test/java/io/quarkus/it/main/MockTestCase.java
```diff
@@ -0,0 +1,60 @@
...
+
+    @Inject
+    MockableBean1 mockableBean1;
+
+    @Inject
+    MockableBean2 mockableBean2;
+
+    @BeforeAll
+    public static void setup() {
+        QuarkusMock.installMock(CDI.current().select(MockableBean1.class).get(), new HiGreeter());
```
mkouba: You could just use `QuarkusMock.installMock(mockableBean1, new HiGreeter())`, right?

integration-tests/main/src/test/java/io/quarkus/it/main/MockTestCase.java
```diff
@@ -0,0 +1,60 @@
...
+
+    @Test
+    public void testBeforeAll() {
+        Assertions.assertEquals("Hi Stuart", mockableBean1.greet("Stuart"));
+        Assertions.assertEquals("Hello Stuart", mockableBean2.greet("Stuart"));
+    }
+
+    @Test
+    public void testPerTestMock() {
+        MockSupport.installMock(mockableBean2, new BonourGreeter());
```
mkouba: `MockSupport` is not part of the user API, or? In fact, it could be package-private...

integration-tests/main/src/test/java/io/quarkus/it/main/MockTestCase.java
stuartwdouglas: It's a static method

mkouba: I see. That's not very nice though. How about specifying the required type/qualifiers, i.e. in this case something like `QuarkusMock.installMock(MockableBean1.class, new HiGreeter())`?

mkouba: And do the lookup under the hood.

mkouba:
Looks really good! We should be careful though when propagating this feature that it really only works for normal scoped beans (so no `@Dependent` and `@Singleton`).

geoand:
Yeah, we need to be very explicit in docs.

I want to try and reboot my `@MockBean` PR I did a while back based on this and see if it can make things easier for users.

geoand:
We also need to tell users that they won't be able to use parallel test execution when this is being used.

integration-tests/main/src/test/java/io/quarkus/it/main/MockTestCase.java
stuartwdouglas: Yea, it should be. Originally this was in the common module and I did not update it.

stuartwdouglas:
Good point, also i need to make that field volatile

FroMage:
Hah, this is funny because it's very relevant to what I'm working on ATM with mocking for ORM with Panache.

I also inject mocking bytecode into the target methods to be able to reroute traffic to mock infrastructure.

Except users seemed to want abitlity to do like in Mockito and specify results with something like:

```java
Mockito.mock(Entity.count()).thenReturn(2);
Mockito.mock(Entity.findById(12)).thenReturn(myEntity);
```

And verify calls happened with:

```java
Mockito.verify(Entity.count()).called();
```

So it's not really providing an entire bean to replace one, but specifying which methods get re-routed and veryfying that they were called.

So I've been starting to reimplement something like that, but now I wonder if I shouldn't rather inject hooks that can be used by Mockito and let it handle the implementation.

geoand:
I think that @stuartwdouglas's proposal covers the most common use case, which is to have a mock per-test.
The next step IMHO will be to make it dead simple to use ( 1) automatically provide the mock, 2) no need to call any Quarkus APIs to wire things up ) which is why I want to redo my old `@MockBean` proposal based on this since I think that zero-effort mocking from a users perspective can be achieved.

geoand:
I have my prototype of `@MockBean` working against this PR.

Here is what it looks like: https://github.com/geoand/quarkus/blob/mockbean2/integration-tests/mockbean/src/test/java/io/quarkus/it/mockbean/GreetingResourceTest.java#L13

I would love to get this PR in so I can then open mine :)

test-framework/junit5/src/main/java/io/quarkus/test/junit/QuarkusTestExtension.java
```diff
@@ -176,6 +177,7 @@ public void beforeEach(ExtensionContext context) throws Exception {
         if (isNativeTest(context)) {
             return;
         }
+        pushMockContext();
```
geoand: This method should probably be called after the `failedBoot` because if we do fail to boot, `pushMockContext` (and `popMockContext` will throw a NPE, thus masking the real problem)

mkouba:
@geoand the `@MockBean` proposal looks nice. However, it seems that it does not support CDI qualifiers... 

Another thing that comes to my mind is that users might expect CDI interceptors to work which would not be the case because the contextual instance is a mock and we don't implement interceptors in client proxies. And the same applies to this PR. On the other hand, it's probably OK if we explicitly mention this in the docs.

geoand:
> @geoand the `@MockBean` proposal looks nice. However, it seems that it does not support CDI qualifiers...
> 

Good catch! Should be easy to fix :)

famod:
Watching with great interest...
@geoand 
> We also need to tell users that they won't be able to use parallel test execution when this is being used.

You mean within one test JVM, not multiple forked test JVMs in the same module, right?
See also: https://maven.apache.org/surefire/maven-surefire-plugin/examples/fork-options-and-parallel-execution.html "Forked Test Execution"

geoand:
@famod yes, within the same JVM you won't be able to run parallel tests when this mocking is used

geoand (approved):

docs/src/main/asciidoc/getting-started-testing.adoc
```diff
@@ -302,7 +302,11 @@ public class TestStereotypeTestCase {
 
 == Mock Support
 
-Quarkus supports the use of mock objects using the CDI `@Alternative` mechanism.
+Quarkus supports the use of mock objects using two different apporoaches. You can either use CDI alternatives to
```
paraita: typo

geoand:
@mkouba can you take a look and make sure it's good for you?

test-framework/junit5/src/main/java/io/quarkus/test/junit/QuarkusMock.java
```diff
@@ -0,0 +1,53 @@
...
+    /**
+     * Installs a mock for a CDI normal scoped bean
+     *
+     * @param mock The mock object
+     * @param instance The CDI normal scoped bean that was injected into your test
+     * @param <T> The bean type
+     */
+    public static <T> void installMockForInstance(T mock, T instance) {
+        //mock support does the actual work, but exposes other methods that are not part of the user API
+        MockSupport.installMock(instance, mock);
```
mkouba: We should probably add a check that the `instance` is an instance of `io.quarkus.arc.ClientProxy` and throw an IAE if it's not? I know that we throw a `RuntimeException` inside the `MockSupport` but it would be probably better to fail fast.

mkouba (commented):
I've added one minor comment. Otherwise looks good.

asodja:
I am also interested in this feature.
geoand 
> @famod yes, within the same JVM you won't be able to run parallel tests when this mocking is used

Are there any plans to make this work running parallel tests within the same JVM, so per test based, or is that not easy to achieve due to Quarkus way of injecting things?

I believe a lot of people are using just maven `-T` option and not forks (not sure how gradle handles this).

geoand:
With the current way this are doing this, it would be quite difficult to allow multiple tests to run in parallel on the same JVM

stuartwdouglas:
We could add a thread local version, but it would not work for requests over HTTP.

test-framework/junit5/src/main/java/io/quarkus/test/junit/QuarkusMock.java
stuartwdouglas: It does not make any real difference, and this is significantly more complex than it sounds, as MockSupport is inside the application class loader. This would mean a lot of extra logic to duplicate a check that is done first thing in the method call.

asodja:
Oh, I see. So for http one would probably have to install mocks and run a separate server for every thread.

famod:
@asodja 

> So for http one would probably have to install mocks for every thread and run a separate server for every thread.

Or use Surefire forks. 😉 Although I haven't tried this QuarkusTests myself, yet. But I have used Surefire forks heavily in an "legacy" project (JBoss EAP 6.4, Arquillian).

test-framework/junit5/src/main/java/io/quarkus/test/junit/QuarkusMock.java
mkouba: Ok.

mkouba (approved):

geoand:
I am thinking that if push comes to shove, we could probably implement per test mocks with something like the following:

* Have a global storage per (test) JVM that is something like a map from a thread name to a map of configured mocks per bean.
* Configure RESTAssured to send an HTTP header that indicate the originating test thread name
* Capture the aforementioned HTTP header very early and place it in a ThreadLocal.
* Change the generated CDI proxy code to check the ThreadLocal for the originating test thread, then look at the global storage to see if there is any mock configured for it.
* Unset the previously set ThreadLocal when the response has been sent back.

Thoughts?

stuartwdouglas:
I had through about doing this (for transaction propagation as well), but it is not quite that simple as we provide lots of reactive stuff that does not always run on the same thread. In theory we could propagate it with MP Context Propagation, but it is not as simple as it looks at first glance.

geoand:
Oh, right I remember you mentioned transactions in some other issue.

Yeah I figured reactive stuff would be an issue. And indeed it would be super confusing if some of the stuff worked and and some didn't, so I guess it's best to not do anything or do it all together.

vsevel:
excellent feature.
and it brings a closure to a year old [inquiry](https://groups.google.com/d/topic/quarkus-dev/S9l4LHCBE54/discussion).

PR #8182, opened 2020-03-26
Switch from -Dno-format to -Dformat.skip

galderz:
Speed up builds

### DIFF
--- a/.github/workflows/ci-actions.yml
+++ b/.github/workflows/ci-actions.yml
@@ -34,8 +34,8 @@ on:
 env:
   # Workaround testsuite locale issue
   LANG: en_US.UTF-8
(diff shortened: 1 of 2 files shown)


.github/workflows/ci-actions.yml
```diff
@@ -80,7 +80,7 @@ jobs:
           restore-only: ${{ github.event_name == 'pull_request' }}
       - name: Build
         run: |
-          mvn -e -B  -DskipTests=true -Dno-format -DskipDocs clean install
+          mvn -e -B  -DskipTests=true -Dformat.skip -DskipDocs clean install
```
gsmet: I don't think we want to switch that particular one because I suppose `format.skip` disables the formatter entirely?

We need the first build to test the format and report the errors.

gsmet (changes requested):
Added a question.

.github/workflows/ci-actions.yml
```diff
@@ -269,7 +269,7 @@ jobs:
         shell: bash
         run: tar -xzvf maven-repo.tgz -C ~
       - name: Build with Maven
-        run: mvn -B --settings .github/mvn-settings.xml -Dno-native -Dno-format -DskipTests -Dtcks install
+        run: mvn -B --settings .github/mvn-settings.xml -Dno-native -Dformat.skip -DskipTests -Dtcks install
```
gsmet: Also, I think we need to keep that one as the TCK tests are only compiled here.

galderz:
Fixed comments.

stuartwdouglas (approved):

gsmet (approved):

PR #8183, opened 2020-03-26
config doc generator  top level grouping refactor

machi1990:
Follow up https://github.com/quarkusio/quarkus/pull/8155 and most precisely this comment https://github.com/quarkusio/quarkus/pull/8155#discussion_r398112922

/cc @geoand thanks for the initial work and  sorry, I did not have the time to check it yesterday, but here it is. Let me know what you think.

### DIFF
--- a/core/processor/src/main/java/io/quarkus/annotation/processor/generate_doc/ConfigDoItemFinder.java
+++ b/core/processor/src/main/java/io/quarkus/annotation/processor/generate_doc/ConfigDoItemFinder.java
@@ -55,8 +55,10 @@ class ConfigDoItemFinder {
              * ('='* (N + 1)) where N is section level.
              */
             final int sectionLevel = 2;
-            final List<ConfigDocItem> configDocItems = recursivelyFindConfigItems(element, configRootInfo.getName(),
-                    configRootInfo.getConfigPhase(), false, sectionLevel);
+            String rootName = configRootInfo.getName();
+            ConfigPhase configPhase = configRootInfo.getConfigPhase();
+            final List<ConfigDocItem> configDocItems = recursivelyFindConfigItems(element, rootName, rootName, configPhase,
+                    false, sectionLevel);
             holder.addConfigRootItems(configRootInfo.getClazz().getQualifiedName().toString(), configDocItems);
         }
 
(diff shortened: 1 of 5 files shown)


core/processor/src/main/java/io/quarkus/annotation/processor/generate_doc/ConfigDoItemFinder.java
```diff
@@ -67,12 +69,13 @@ ScannedConfigDocsItemHolder findInMemoryConfigurationItems() {
...
      * @param element - root element
-     * @param parentName - root name
+     * @param rootName - root name
+     * @param parentName - parent name
      * @param configPhase - configuration phase see {@link ConfigPhase}
      * @param withinAMap - indicates if a a key is within a map or is a map configuration key
      * @param sectionLevel - section sectionLevel
      */
-    private List<ConfigDocItem> recursivelyFindConfigItems(Element element, String parentName,
+    private List<ConfigDocItem> recursivelyFindConfigItems(Element element, String rootName, String parentName,
```
geoand: When would `rootName` be different than `parentName`?

geoand:
Thanks a lot of the follow up @machi1990! 

I've added a comment.

core/processor/src/main/java/io/quarkus/annotation/processor/generate_doc/ConfigDoItemFinder.java
machi1990: When scanning group configs items -see  https://github.com/quarkusio/quarkus/pull/8183/files#diff-5b3aaa451926ba3ee4ee4207f00433f1R289

So taking the `liveness-probe` config group for example, we'll have `kubernetes` as root name,  and `kubernetes.liveness-probe` as parent name.

machi1990: That's to say, the `rootName` does not change with depth while the `parentName` changes.

geoand: Gotcha, thanks!

geoand (approved):
Thanks a lot!

Issue #8081, opened 2020-03-23
Multiple NoClassDefFoundError with SVM built with jdk11u


galderz:
[jdk11u](https://github.com/openjdk/jdk11u) now contains the necessary bits to build a JDK with static libraries, which is required to build substratevm. While testing a snapshot JDK built out of there the following error has appeared:

```
Caused by: java.lang.NoClassDefFoundError: [Lorg/springframework/http/HttpMethod;
        at java.base/java.lang.Class.getDeclaredMethods0(Native Method)
        at java.base/java.lang.Class.privateGetDeclaredMethods(Class.java:3166)
        at java.base/java.lang.Class.getDeclaredMethods(Class.java:2309)
        at jdk.internal.vm.ci/jdk.vm.ci.hotspot.HotSpotResolvedObjectTypeImpl.getDeclaredMethods(HotSpotResolvedObjectTypeImpl.java:932)
        at com.oracle.svm.hosted.classinitialization.ClassInitializationFeature.declaresDefaultMethods(ClassInitializationFeature.java:375)
        at com.oracle.svm.hosted.classinitialization.ConfigurableClassInitialization.processInterfaces(ConfigurableClassInitialization.java:580)
        at com.oracle.svm.hosted.classinitialization.ConfigurableClassInitialization.computeInitKindAndMaybeInitializeClass(ConfigurableClassInitialization.java:556)
        at com.oracle.svm.hosted.classinitialization.ConfigurableClassInitialization.computeInitKindAndMaybeInitializeClass(ConfigurableClassInitialization.java:130)
        at com.oracle.svm.hosted.classinitialization.ConfigurableClassInitialization.maybeInitializeHosted(ConfigurableClassInitialization.java:158)
        at com.oracle.svm.hosted.SVMHost.registerType(SVMHost.java:200)
```

Longer output can be found [here](https://gist.github.com/galderz/d739dcebd6872c503c634fcfeca6872f).

We've had a little Zulip conversation [here](https://quarkusio.zulipchat.com/#narrow/stream/187038-dev/topic/CI/near/191449925), and we've discovered that `HttpMethod` is not included in the jar. The question is, why does it get into the pointsto analysis with this particular JDK? Further analysis is required.

galderz:
[...]

PR #7198, opened 2020-02-14
#6549 - Apache Tika can not parse Microsoft Docx format in native mode

tpenakov:
Fixes: #6549
A few things to points out:

- xalan dependency is included - probably on wrong place: extensions/arc/runtime/pom.xml
- additional config property is added for supported file types - we have support for pptx ans xlsx file types as well, however if we include all of them - the native build ends with OOM error.
- Trying to use `@ConfigProperty` in `io.quarkus.it.tika.TikaEmbeddedContentTest` leads to NPE for native build. This one (#2061) claims that is fixed, but I am receiving it.

PR #8185, opened 2020-03-26
Add missing ooxml-schemas dependency #8081

galderz:
~This is a draft PR~.

Required to avoid [issue](https://github.com/quarkusio/quarkus/issues/8081#issuecomment-604014138).

Note that this change happens to be part of the bigger #7198 PR. I came to it separately following the advice [here](https://stackoverflow.com/questions/49483042/apache-poi-java-lang-noclassdeffounderror). Talking to @sberyozkin he pointed out it was part of the other PR.

### DIFF
--- a/extensions/tika/runtime/pom.xml
+++ b/extensions/tika/runtime/pom.xml
@@ -44,6 +44,16 @@
             </exclusions>
         </dependency>
         <dependency>
+            <groupId>org.apache.poi</groupId>
+            <artifactId>ooxml-schemas</artifactId>
+            <exclusions>
+                <exclusion>
+                    <groupId>org.apache.xmlbeans</groupId>
+                    <artifactId>xmlbeans</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+        <dependency>
             <groupId>jakarta.annotation</groupId>
             <artifactId>jakarta.annotation-api</artifactId>
         </dependency>
--- a/bom/runtime/pom.xml
+++ b/bom/runtime/pom.xml
@@ -135,6 +135,7 @@
         <!-- Scala is used by Kafka so we need to choose a compatible version -->
         <scala.version>2.12.9</scala.version>
         <tika.version>1.22</tika.version>
+        <ooxml-schemas.version>1.4</ooxml-schemas.version>
         <aws-lambda-java.version>1.1.0</aws-lambda-java.version>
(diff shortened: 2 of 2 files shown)


galderz:
Ups, too many commits there, let me revert the others.

galderz:
Oh crap, messed up commits, one sec...

galderz:
Fixed now 😅

extensions/tika/runtime/pom.xml
```diff
@@ -43,6 +43,16 @@
                 </exclusion>
             </exclusions>
         </dependency>
+        <dependency>
```
sberyozkin: @galderz Thanks for this PR. My only reservation is that with `quarkus-tika`, if a user configures a PDF parser only for example (`quarkus.tika.parser=pdf`)), then so far none of the non-PDF dependencies are expected to make it to the native image. 
I wonder, if this dependency were optional, then would making a build step which would check if `OOXML` is enabled (can be true by default to start with - I can add an `ooxml` shortcut later) and if yes - add a native resource build item which would make this `ooxml-schemas` included ?
Is it realistic ?

galderz: This dependency is only needed during building of the image itself. It's only for class resolution while trying to discover which methods the class has, nothing else. None of the classes that were missing would actually make it to the final image.

galderz: IOW, it has 0 impact on the pointsto analysis that svm does. So, unless the end application uses something there indirectly, nothing will be included in the native image.

sberyozkin: Hi @galderz, so for example, `org/openxmlformats/schemas/wordprocessingml/x2006/main/STDisplacedByCustomXml` is missing, and few more I guess, so adding  this dependency also contributes to resolving #7198 (see for example https://github.com/quarkusio/quarkus/pull/7198/files#diff-cefda81cecd3360b76139d0044ecb53aR25). So unless `quarkus-tika` processor references these classes directly or indirectly, no native image size will be affected ?

sberyozkin: @galderz cool, your last comment makes it quite clear, cheers

sberyozkin (approved):

galderz:
This is no longer in draft. The change had a positive effect.

sberyozkin:
@galderz Something failed in the native build but I'm pretty sure it is not related, have restarted the build...

PR #8189, opened 2020-03-26
Indexing null streams shows Stream closed errors in the logs

gastaldi:


geoand:
Can you provide more details on how / when this occurs?

independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/BeanArchives.java
```diff
@@ -36,7 +36,7 @@
 
 public final class BeanArchives {
 
-    private static final Logger LOGGER = Logger.getLogger(BeanArchives.class);
+    private static final Logger logger = Logger.getLogger(BeanArchives.class);
```
mkouba: I prefer to use upper case for `static final` constants...

mkouba:
@geoand it could happen when a bean class links a type that is not part of the index, e.g. as a return type of the producer method, and we're not able to find the class.

@gastaldi I'd prefer to keep the same warning for both the `IOException` and `null`.

independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/BeanArchives.java
gastaldi: The logger reference is not a constant, but a final reference, and should NOT be in uppercase. A constant VALUE should be in uppercase. :wink: 

See https://stackoverflow.com/questions/1417190/should-a-static-final-logger-be-declared-in-upper-case

geoand:
> @geoand it could happen when a bean class links a type that is not part of the index, e.g. as a return type of the producer method, and we're not able to find the class.
> 

Gotcha, thanks!

geoand:
I'll let you guys figure out the proper naming for the logger :P

independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/BeanArchives.java
gastaldi: I had the same discussion with the launcher team back then: https://github.com/fabric8-launcher/launcher-application/pull/263

gastaldi:
@mkouba I changed the `null` debug to a warning instead, let me know if that works for you

gastaldi:
I have also created https://github.com/wildfly/jandex/pull/77 to check if the argument is a null stream

independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/BeanArchives.java
mkouba: Fair enough. For me anything `static final` feels like a constant and I prefer to use upper case because it's very clear in the code (I'm not going to dive into the immutability discussions ;-). However, I don't think it's a good idea to change this at random. It should be either forced by the formatter or just leave this to the author...

mkouba:
> I changed the null debug to a warning instead, let me know if that works for you

I think that we should use the same message, i.e. `logger.warnf(e, "Failed to index %s: %s", className, e.getMessage());` for both cases... 
```java
if (!result) {
 logger.warnf(e, "Failed to index %s: %s", className, e.getMessage());
}
```

independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/BeanArchives.java
gastaldi: I agree it should be forced by the formatter at some point. 

PS: Doing a quick grep I found that `logger` is more common in the codebase than `LOGGER` :)

gastaldi:
@mkouba hm, it would be a bit hard to keep the same message because we're not throwing an exception here :thinking:

independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/BeanArchives.java
manovotn: You could consider it a constant so long as the reference is immutable. Which I am not sure about with `Logger`. Either way, if we are already using upper-case in various places throughout the code, I'd just keep it that way ;-)

mkouba:
@gastaldi Ok. You could at least share the same message...

manovotn:
> @mkouba hm, it would be a bit hard to keep the same message because we're not throwing an exception here

Why is that a problem? We swallow the exception anyway and then log something and have access to CL and class variables. It can be the very same message in both cases.

mkouba:
> Why is that a problem?

Because the is no exception and we use `logger.warnf(e,...`?

gastaldi:
Okay, I've reverted the `LOGGER`x`logger` change and made the message similar. See if that helps

mkouba (approved):

manovotn:
> Because the is no exception and we use logger.warnf(e,...?

I really shouldn't try to review PRs while attending a meeting :-)

manovotn (approved):

PR #8203, opened 2020-03-26
Render non built-in database access configuration more explicit

emmanuelbernard:
Use oracle as example
Make dedicated section
Also simply the flow for some sentences

### DIFF
--- a/docs/src/main/asciidoc/datasource.adoc
+++ b/docs/src/main/asciidoc/datasource.adoc
@@ -96,16 +96,18 @@ We currently include these built-in database kinds:
 * MySQL: `mysql`
 * PostgreSQL: `postgresql`, `pgsql` or `pg`
 
-Note that these database kinds only facilitate things.
-For instance, when we have a JDBC driver extension for a given kind,
-we resolve the JDBC driver to be used automatically, so you don't need to configure it yourself.
-If you want to use a database kind that is not part of this list, just use `other` or whatever name you prefer.
+Giving Quarkus the database kind you are targeting will facilitate configuration.
+By using a JDBC driver extension and setting the kind in the configuration,
+Quarkus resolves the JDBC driver automatically,
+so you don't need to configure it yourself.
+If you want to use a database kind that is not part of the built-in ones, use `other` and define the JDBC driver explicitly.
 
 [NOTE]
 ====
-Although it is possible to use any JDBC driver in JVM mode, it is unlikely that it will work when compiling your application to a native executable.
+You can use any JDBC driver in a Quarkus app in JVM mode (see <<other-databases,Using other databases>>).
+It is unlikely that it will work when compiling your application to a native executable though.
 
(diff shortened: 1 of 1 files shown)


emmanuelbernard:
@gsmet I'll let you be the judge if you want to backport to 1.3

emmanuelbernard:
@tqvarnst PR for oracle DB usage ^

docs/src/main/asciidoc/datasource.adoc
```diff
@@ -229,16 +233,34 @@ As previously stated, most of the time, this automatic resolution will suit you
...
+If you plan to make a native executable, we recommend you use the existing JDBC Quarkus extensions (or contribute one for your driver).
+
+Here is how you would use the OpenTracing driver.
 
 [source, properties]
 ----
 quarkus.datasource.jdbc.driver=io.opentracing.contrib.jdbc.TracingDriver
 ----
 
+Here is how you would define access to a database with no built-in support (in JVM mode)
```
machi1990:
```suggestion
Here is how you would define access to an Oracle database with no built-in support (in JVM mode)
```

docs/src/main/asciidoc/datasource.adoc
```diff
@@ -229,16 +233,34 @@ As previously stated, most of the time, this automatic resolution will suit you
...
-or if you want to use a database for which we don't have a built-in JDBC driver extension.
+[[other-databases]]
+==== Use a database with no built-in extension or with a different driver
+
+You can use a specific driver if you need to (for instance for using the OpenTracing driver)
+or if you want to use a database for which Quarkus does not have a built-in JDBC driver extension.
 
-Defining a specific JDBC driver is as simple as:
+Without an extension, the driver will work fine in any Quarkus app running in JVM mode.
+It is unlikely that it will work when compiling your application to a native executable though.
```
machi1990:
```suggestion
It is unlikely that it will work when compiling your application to a native executable.
```

docs/src/main/asciidoc/datasource.adoc
emmanuelbernard: Not convinced. Keeping it as is

docs/src/main/asciidoc/datasource.adoc
emmanuelbernard: Why do you want to remove it? @machi1990 ? I understand my general rule is not to add such frivolous words in doc but here it's to balance the previous sentence

emmanuelbernard:
Thanks @machi1990

docs/src/main/asciidoc/datasource.adoc
machi1990: Okay, I though we could explicit the target database. I guess it does not add that much considering we have the driver right after in config properties.

machi1990:
> Thanks @machi1990

You are welcome. Did a quick check though, and would not call it a thorough review :-)

gsmet (approved):

PR #8211, opened 2020-03-27
Enlist reactive entity for reflection

loicmathieu:
MongoDB with panache enlist for reflection entities (and repository parameterized types) automatically.

This PR adds the same facility for the reactive implementation.

This has been discuss on Zuilip: https://quarkusio.zulipchat.com/#narrow/stream/187030-users/topic/Bson.20codec.20error.20when.20using.20reactive.20mongo.20panache.20in.20native

### DIFF
--- a/extensions/panache/mongodb-panache/deployment/src/main/java/io/quarkus/mongodb/panache/deployment/PanacheResourceProcessor.java
+++ b/extensions/panache/mongodb-panache/deployment/src/main/java/io/quarkus/mongodb/panache/deployment/PanacheResourceProcessor.java
@@ -301,10 +301,12 @@ public class PanacheResourceProcessor {
     @BuildStep
     void buildMutiny(CombinedIndexBuildItem index,
             ApplicationIndexBuildItem applicationIndex,
+            BuildProducer<ReflectiveClassBuildItem> reflectiveClass,
             BuildProducer<BytecodeTransformerBuildItem> transformers) throws Exception {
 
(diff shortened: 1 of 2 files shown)


extensions/panache/mongodb-panache/deployment/src/main/java/io/quarkus/mongodb/panache/deployment/PanacheResourceProcessor.java
```diff
@@ -336,8 +347,13 @@ void buildMutiny(CombinedIndexBuildItem index,
             if (modelClasses.add(classInfo.name().toString()))
                 modelEnhancer.collectFields(classInfo);
         }
+
+        // iterate over all the entity classes
```
machi1990: do we need these comments?

loicmathieu: I copied them for the implementation of the impertive version.
this method is really big and complex so the comments help me to maintain it yes ...

machi1990: I personally do not see if they add any value.. upon arriving here the loop tells me that an iteration is going on..same for the one below about reflection registration.

FroMage: While I also don't think this particular comment adds much value, I think it's way overkill to ask people to put less comments. In general comments are good and should be encouraged.

FroMage (approved):
LGTM.

PR #8219, opened 2020-03-27
Qute - fix parser and computed section helpers used during validation

### LINKED ISSUE #8217: Qute - computed section helpers used during validation don't play well with {#include}
As a result section blocks  in an `{#include}` section must also define an end tag which is normally optional.

mkouba:
- resolves #8217


gastaldi (approved):
Nice!

PR #8229, opened 2020-03-28
Upgrade zookeeper to 3.5.7

### LINKED ISSUE #8212: Kafka not starting with JDK 14
I'm still trying to get CI on JDK 14:
- [x] REST Assured
- [x] Gradle
- [ ] Now Kafka is not starting properly apparently:

```
Caused by: java.lang.RuntimeException: Unable to start Quarkus test resource io.quarkus.it.kafka.KafkaTestResource@c11332b
	at io.quarkus.test.common.TestResourceManager.start(TestResourceManager.java:51)
	... 44 more
Caused by: java.lang.RuntimeException: kafka.zookeeper.ZooKeeperClientTimeoutException: Timed out waiting for connection while in state: CONNECTING
	at io.quarkus.it.kafka.KafkaTestResource.start(KafkaTestResource.java:30)
	at io.quarkus.test.common.TestResourceManager.start(TestResourceManager.java:46)
	... 44 more
Caused by: kafka.zookeeper.ZooKeeperClientTimeoutException: Timed out waiting for connection while in state: CONNECTING
	at kafka.zookeeper.ZooKeeperClient.$anonfun$waitUntilConnected$3(ZooKeeperClient.scala:259)
	at scala.runtime.java8.JFunction0$mcV$sp.apply(JFunction0$mcV$sp.java:23)
	at kafka.utils.CoreUtils$.inLock(CoreUtils.scala:253)
	at kafka.zookeeper.ZooKeeperClient.waitUntilConnected(ZooKeeperClient.scala:255)
	at kafka.zookeeper.ZooKeeperClient.<init>(ZooKeeperClient.scala:113)
	at kafka.zk.KafkaZkClient$.apply(KafkaZkClient.scala:1858)
	at kafka.server.KafkaServer.createZkClient$1(KafkaServer.scala:375)
	at kafka.server.KafkaServer.initZkClient(KafkaServer.scala:399)
	at kafka.server.KafkaServer.startup(KafkaServer.scala:207)
	at io.debezium.kafka.KafkaServer.startup(KafkaServer.java:220)
	at java.base/java.util.concurrent.ConcurrentHashMap$ValuesView.forEach(ConcurrentHashMap.java:4780)
	at io.debezium.kafka.KafkaCluster.startup(KafkaCluster.java:250)
	at io.quarkus.it.kafka.KafkaTestResource.start(KafkaTestResource.java:28)
	... 45 more
```

jaikiran:
Fixes https://github.com/quarkusio/quarkus/issues/8212

### DIFF
--- a/bom/runtime/pom.xml
+++ b/bom/runtime/pom.xml
@@ -131,7 +131,7 @@
         <mutiny-client.version>0.0.13</mutiny-client.version>
         <kafka2.version>2.4.1</kafka2.version>
         <debezium.version>1.1.0.Final</debezium.version>
-        <zookeeper.version>3.4.14</zookeeper.version>
+        <zookeeper.version>3.5.7</zookeeper.version>
         <!-- Scala is used by Kafka so we need to choose a compatible version -->
         <scala.version>2.12.9</scala.version>
         <tika.version>1.22</tika.version>


geoand:
Let's let the expert @gunnarmorling decide on this one 😉

gunnarmorling (approved):
LGTM, 3.5.7 is the right ZK version matching AK 2.4.1. Note that Debezium 1.1.0 still is on 2.4.0, but I don't think it makes any difference.

I'd only suggest to add a comment before the `kafka2.version` expressing that the next three lines need to be upgraded in lockstep, so to prevent similar issues from occuring again down the road.

Thanks for the thorough analysis and fix, @jaikiran!

machi1990:
> I'd only suggest to add a comment before the `kafka2.version` expressing that the next three lines need to be upgraded in lockstep, so to prevent similar issues from occuring again down the road.
> 

I noticed that we have dependabot in charge of the upgrade - https://github.com/quarkusio/quarkus/blob/master/.dependabot/config.yml#L65. 

@gastaldi Is there a way to configure this with dependabot? Or it is something that we'll have to keep in mind when the bot creates PRs upgrading the kafka, debezium and zookeeper?

gastaldi:
@machi1990 I don't think that's possible. In this case I'd recommend to remove this dependency from Dependabot and updating manually

gsmet:
So I think it would be worth adding a comment and keeping dependabot.

We should notice it when the upgrades are done as we have a bit of context in the diff.

Otherwise, we will never upgrade those and they will end up totally outdated.

gastaldi:
@gsmet aren't those dependencies managed by Debezium? I was under the
impression they were

Em Sáb, 28 de mar de 2020 08:45, Guillaume Smet <notifications@github.com>
escreveu:

gsmet:
Not sure about Zookeeper but we definitely need the Kafka version to be defined.

The Zookeeper version might have been added to ensure dependency convergence.

Given Debezium is only used as a test dependency, I would prefer us to keep the liberty to upgrade Kafka when we want to do it.

gsmet (approved):

gsmet:
Maybe we should add Zookeeper to Dependabot too. We are not forced to accept the updates but at least it will be kept updated.

PR #8237, opened 2020-03-28
Increase kafka client test coverage

### LINKED ISSUE #8025: Increase test coverage of the Kafka client
**Description**
there are a few use cases that should be covered by tests.

**Implementation ideas**

* Testing secure connection, and Kafka security - especially in native
* Testing Avro schema support
* Testing hot reload (code change)

cescoffier:
Fixes #8025 

add a few tests related to the Kafka client.

Unfortunately, Avro does not work in native mode (fields written using unsafe). 
I will create an issue to track this.


cescoffier:
Forgot to disable the test if docker is not enabled. Will fix that.

integration-tests/kafka/pom.xml
```diff
@@ -54,6 +54,26 @@
...
+        <!-- Avro -->
+        <dependency>
+            <groupId>org.apache.avro</groupId>
+            <artifactId>avro</artifactId>
+            <version>1.9.1</version>
+        </dependency>
+        <dependency>
+            <!-- Not in Maven Central -->
+            <groupId>io.confluent</groupId>
+            <artifactId>kafka-avro-serializer</artifactId>
```
gunnarmorling, line 66: Have you considered using the serializer from the Apicurio project?

integration-tests/kafka/src/main/java/io/quarkus/it/kafka/avro/AvroEndpoint.java
```diff
@@ -0,0 +1,90 @@
...
+        return consumer;
+    }
+
+    public static KafkaProducer<Integer, Pet> createProducer(String registry) {
+        Properties props = new Properties();
+        props.put(ProducerConfig.BOOTSTRAP_SERVERS_CONFIG, "localhost:19092");
+        props.put(ProducerConfig.CLIENT_ID_CONFIG, "test-avro");
+        props.put(ProducerConfig.KEY_SERIALIZER_CLASS_CONFIG, IntegerSerializer.class.getName());
+        props.put(ProducerConfig.VALUE_SERIALIZER_CLASS_CONFIG, KafkaAvroSerializer.class.getName());
+        props.put(AbstractKafkaAvroSerDeConfig.SCHEMA_REGISTRY_URL_CONFIG, registry);
```
gunnarmorling, line 87: Same thought as above: use the Apicurio SerDe?

integration-tests/kafka/src/main/java/io/quarkus/it/kafka/sasl/SaslKafkaEndpoint.java
```diff
@@ -0,0 +1,66 @@
...
+    @GET
+    public String get() {
+        final ConsumerRecords<Integer, String> records = consumer.poll(Duration.ofMillis(60000));
+        if (records.isEmpty()) {
+            return null;
+        }
+        return records.iterator().next().value();
+    }
+
+    private static void addSSL(Properties props) {
```
gunnarmorling, line 43: `addSsl`, to be consistent with other abbreviation casing in this test?

integration-tests/kafka/src/test/java/io/quarkus/it/kafka/SaslKafkaConsumerTest.java
```diff
@@ -0,0 +1,60 @@
...
+
+import io.quarkus.test.common.QuarkusTestResource;
+import io.quarkus.test.junit.QuarkusTest;
+import io.restassured.RestAssured;
+
+@QuarkusTest
+@QuarkusTestResource(KafkaSASLTestResource.class)
+public class SaslKafkaConsumerTest {
+
+    private static void addSSL(Properties props) {
```
gunnarmorling, line 27:
```suggestion
    private static void addSsl(Properties props) {
```

gunnarmorling (approved):
LGTM, a few minor comments inline.

gsmet:
@cescoffier can you check Gunnar's comments?

cescoffier:
@gunnarmorling I need to look at the Apicurio version of the Avro serde. So far the need was the "original" Avro. I can see an opportunity here if it works flawlessly in native. 
I have created https://github.com/quarkusio/quarkus/issues/8541.

gsmet (changes requested):
I really don't like the fact that we have versions in the pom files. They should either be in the BOM if it makes sense or in the build-parent.

cescoffier:
As these artifacts are not available from Maven Central it's tricky to add them to the dependency management from the parent. That's why I put the version there on purpose.

Someone not aware of this would add the dependency that won't get resolved. 
Note that we should NOT declare the repository in the parent either.

gsmet (approved):

gsmet:
I rebased and force pushed as it was an old PR. I'm a bit worried about the external repository. We'll see if it causes problems and reevaluate.

Issue #7046, opened 2020-02-06
Add equals/ hashcode to PanacheMongoEntity


GregJohnStewart:
**Description**

I would like it if the `PanacheMongoEntity` would have its `equals` and `hashcode` methods implemented, in order to more easily compare Mongo entity objects that utilize this class. Without the methods defined by the panache entity class, it makes using lombok to simplify the code written for extending classes more difficult.

**Implementation ideas**

I wrote my own wrapper class for `PanacheMongoEntity` as a stopgap, and the following seems to work: (the equals implementation is pulled from a delombok, and I now realize is missing a null check on the id, but it's essentially what I would say we want)

```
public abstract class OurMongoEntity extends PanacheMongoEntity {
    @Override
    public int hashCode() {
        return Objects.hash(this.id);
    }

    @Override
    public boolean equals(Object o) {
        if (o == this) return true;
        if (!(o instanceof User)) return false;
        final User other = (User) o;
        if (!other.canEqual((Object) this)) return false;
        if (!this.id.equals(other.id)) return false;
        return true;
    }
}
```

It seems like a fairly trivial thing to add to me, having not actually looked at the panache entity class. I can probably add it in if it is determined to be something we want to add.

loicmathieu:
@FroMage WDYT ?
If we do it, we will do it also for Hibernate with Panache ...

According to be, providing default hashcode/equals for `PanacheEntity` and `PanacheMongoEntity` is easy but as the ID are only provided after the entity was persisted to the database, this means the hashcode and equals method will always return false when the entity is created but not yet persisted. This can be very surprising in some cases ...

FroMage:
I think @Sanne had an opinion about hashCode/equals for entities, but I don't remember it. I've never ever needed it myself because I've never had to deal with collections of unpersisted entities, or multiple instances of the same persisted entity.

Sanne:
indeed, the question of how to properly implement entity equality is a subject of strong debates.

Personally I strongly believe that entities managed by Hibernate ORM are best compared via object identity:  having no overrides for equals/hashcode is the only solution that gives no strong headaches to users - provided they have no "special needs".

The problem with those users who do have "special needs" most often will also require a customization of such equality contracts, so I'd say it's best to not generate any at all.

Identity is correctly managed automatically for managed entities; it also so happens that the default identity is the most efficient.

[...]

PR #8241, opened 2020-03-28
Adding equals and hashcode methods to PanacheMongoEntity

rhuan080:
Signed-off-by: rhuan080 <rhuan080@gmail.com>

### DIFF
--- a/extensions/panache/mongodb-panache/runtime/src/main/java/io/quarkus/mongodb/panache/PanacheMongoEntity.java
+++ b/extensions/panache/mongodb-panache/runtime/src/main/java/io/quarkus/mongodb/panache/PanacheMongoEntity.java
@@ -35,7 +35,7 @@ public abstract class PanacheMongoEntity extends PanacheMongoEntityBase {
         if (this == o) {
             return true;
         }
-        if (Objects.isNull(o) || getClass() != o.getClass()) {
+        if (o == null || getClass() != o.getClass()) {
             return false;
         }
         PanacheMongoEntity that = (PanacheMongoEntity) o;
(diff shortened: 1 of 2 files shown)


extensions/panache/mongodb-panache/runtime/src/main/java/io/quarkus/mongodb/panache/PanacheMongoEntity.java
```diff
@@ -27,4 +29,21 @@
     public String toString() {
         return this.getClass().getSimpleName() + "<" + id + ">";
     }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) {
+            return true;
+        }
+        if (Objects.isNull(o) || getClass() != o.getClass()) {
```
nimo23: use `object == null` instead of `Objects.isNull(object)`.

rhuan080: I'm so sorry, but why? Objects.isNull(object) executes it:
`public static boolean isNull(Object obj) {
        return obj == null;
    }`

Objects.isNull(object) is more semantic.

nimo23: There is a slight performance difference and why the indirection? `Objects.isNull` exists to be used as a Predicate, filter(Objects::isNull), lamda style. You do not need it here, so why the extra method call? https://stackoverflow.com/questions/37972859/java-util-objects-isnull-vs-object-null

nimo23: And `==` is more readability than `Objects.isNull(obj)`..

loicmathieu:
Adding equals and hash code to Panache has been discussed in #7046 and dismissed.

rhuan080:
Hi @loicmathieu 

I didn't know about the issue. I have commented there.

Sanne:
Please see https://github.com/quarkusio/quarkus/issues/7046#issuecomment-605973954 : I really don't think this is a good idea.

PR #8248, opened 2020-03-29
WIP: fix(#7900): Adds Cubbyhole Response Wrapping

lordofthejars:
The PR is not ready to be merged, need tests, docs, ... but it is the first iteration. As I said in the issue, I have used this method in the past and this is how I used it. But basically this method implements the next part: https://learn.hashicorp.com/vault/secrets-management/sm-cubbyhole#step-2-unwrap-the-secret which is how Vault creators shows that this should be used.

Things I want you to note:

1) It is really simple solution ad the code looks easy to follow.
2) It is not invasive at all, we are not changing the authentication method at all, what was working before should work now without any problem, this PR just add a new logic after this.
3) To access to unwrap token, the user can use any auth method supported.
4) an `application.properties` that implements the steps showed at https://learn.hashicorp.com/vault/secrets-management/sm-cubbyhole#step-2-unwrap-the-secret  would look like (notice that the token with `default` role is something that the Vault operator must create we cannot do this as an operator might decide to use another role):

```
quarkus.vault.authentication.client-token=s.VR3mwdk9miJ6VRLFbQe0Duwq
quarkus.vault.authentication.unwrap-token=s.3Kf3Xfn58Asr3bSDkRXATHrw
```

What is happening here is that we are logging to Vault using the `client-token` (notice that could be any other auth method), when logged unwraps the `unwrap-token` and use the returned token as logged token all the time.

So if you read the link these are the two steps implemented here:

1. Authenticate with Vault using this default token (less privileged token)
2. Unwrap the secret to obtain more privileged token (apps persona token)
3. Read secret/dev using the appstoken (unwrapped)

If you read the link I provided they say as a note:

`If a client has been expecting delivery of a response-wrapping token and none arrives, this may be due to an attacker intercepting the token and then preventing it from traveling further. This should cause an alert to trigger an immediate investigation.`

This is something I plan to implement when we agree with this first iteration.

Also they say:

`Verify that this token has apps policy attached.`

This is something that I'll add after this first iteration, if user for example adds a property with the roles this unwrapped token should contain, then I can automatically check them as well.

And that's all, I have no much time to spend on this so let's try to make things easy, and at the end the only thing I am implementing here is the lessons provided by Vault team.

cc/ @vsevel @sberyozkin @machi1990

### DIFF
--- a/extensions/vault/runtime/src/main/java/io/quarkus/vault/runtime/client/dto/sys/VaultUnwrapAuth.java
+++ b/extensions/vault/runtime/src/main/java/io/quarkus/vault/runtime/client/dto/sys/VaultUnwrapAuth.java
@@ -1,0 +1,6 @@
+package io.quarkus.vault.runtime.client.dto.sys;
+
+import io.quarkus.vault.runtime.client.dto.auth.AbstractVaultAuthAuth;
+
+public class VaultUnwrapAuth extends AbstractVaultAuthAuth<Object> {
+}
--- a/extensions/vault/runtime/src/main/java/io/quarkus/vault/runtime/client/dto/sys/VaultUnwrapResult.java
+++ b/extensions/vault/runtime/src/main/java/io/quarkus/vault/runtime/client/dto/sys/VaultUnwrapResult.java
@@ -1,0 +1,34 @@
+package io.quarkus.vault.runtime.client.dto.sys;
+
+import io.quarkus.vault.runtime.client.dto.AbstractVaultDTO;
+
+/**
+ * {
+ *   "request_id": "5f654232-b0de-c27f-1fae-70c9d323c458",
+ *   "lease_id": "",
+ *   "lease_duration": 0,
+ *   "renewable": false,
+ *   "data": null,
+ *   "warnings": null,
+ *   "auth": {
+ *     "client_token": "s.6tyZnRgG7TpDMPg135CEVMHk",
+ *     "accessor": "Z00Qd1lU9apbt8eBIm1XUQHE",
+ *     "policies": [
+ *       "apps",
+ *       "default"
+ *     ],
+ *     "token_policies": [
+ *       "apps",
+ *       "default"
+ *     ],
+ *     "identity_policies": null,
+ *     "metadata": null,
+ *     "orphan": false,
+ *     "entity_id": "",
+ *     "lease_duration": 2764800,
+ *     "renewable": true
+ *   }
+ * }
+ */
+public class VaultUnwrapResult extends AbstractVaultDTO<Object, VaultUnwrapAuth> {
+}
--- a/extensions/vault/runtime/src/main/java/io/quarkus/vault/runtime/client/dto/sys/VaultUnwrapResult.java
+++ b/extensions/vault/runtime/src/main/java/io/quarkus/vault/runtime/client/dto/sys/VaultUnwrapResult.java
@@ -4,30 +4,30 @@ import io.quarkus.vault.runtime.client.dto.AbstractVaultDTO;
 
 /**
  * {
- *   "request_id": "5f654232-b0de-c27f-1fae-70c9d323c458",
- *   "lease_id": "",
- *   "lease_duration": 0,
- *   "renewable": false,
- *   "data": null,
- *   "warnings": null,
- *   "auth": {
- *     "client_token": "s.6tyZnRgG7TpDMPg135CEVMHk",
- *     "accessor": "Z00Qd1lU9apbt8eBIm1XUQHE",
- *     "policies": [
- *       "apps",
- *       "default"
- *     ],
- *     "token_policies": [
- *       "apps",
- *       "default"
- *     ],
- *     "identity_policies": null,
- *     "metadata": null,
- *     "orphan": false,
- *     "entity_id": "",
- *     "lease_duration": 2764800,
- *     "renewable": true
- *   }
+ * "request_id": "5f654232-b0de-c27f-1fae-70c9d323c458",
+ * "lease_id": "",
+ * "lease_duration": 0,
+ * "renewable": false,
+ * "data": null,
+ * "warnings": null,
+ * "auth": {
+ * "client_token": "s.6tyZnRgG7TpDMPg135CEVMHk",
+ * "accessor": "Z00Qd1lU9apbt8eBIm1XUQHE",
+ * "policies": [
+ * "apps",
+ * "default"
+ * ],
+ * "token_policies": [
+ * "apps",
+ * "default"
+ * ],
+ * "identity_policies": null,
+ * "metadata": null,
+ * "orphan": false,
+ * "entity_id": "",
+ * "lease_duration": 2764800,
+ * "renewable": true
+ * }
  * }
  */
 public class VaultUnwrapResult extends AbstractVaultDTO<Object, VaultUnwrapAuth> {
--- a/extensions/vault/runtime/src/main/java/io/quarkus/vault/runtime/config/VaultAppRoleAuthenticationConfig.java
+++ b/extensions/vault/runtime/src/main/java/io/quarkus/vault/runtime/config/VaultAppRoleAuthenticationConfig.java
@@ -15,9 +15,17 @@ public class VaultAppRoleAuthenticationConfig {
     public Optional<String> roleId;
 
     /**
-     * Secret Id for AppRole auth method. This property is required when selecting the app-role authentication type.
+     * Secret Id for AppRole auth method. This property is required when selecting the app-role authentication type and not setting the secret id as wrapped.
+     *
+     * {@link VaultAppRoleAuthenticationConfig#secretIdWrapped}
      */
     @ConfigItem
     public Optional<String> secretId;
+
+    /**
+     * Wrapped token containing the secret Id for AppRole auth method. This flag is not mandatory, only required if secret id is wrapped.
+     */
+    @ConfigItem
+    public Optional<String> secretIdWrapped;
 
 }
--- a/extensions/vault/runtime/src/main/java/io/quarkus/vault/runtime/config/VaultConfigSource.java
+++ b/extensions/vault/runtime/src/main/java/io/quarkus/vault/runtime/config/VaultConfigSource.java
@@ -215,9 +215,10 @@ public class VaultConfigSource implements ConfigSource {
                 DEFAULT_KUBERNETES_JWT_TOKEN_PATH);
         serverConfig.authentication.userpass.username = getOptionalVaultProperty("authentication.userpass.username");
         serverConfig.authentication.userpass.password = getOptionalVaultProperty("authentication.userpass.password");
+        serverConfig.authentication.userpass.passwordWrapped = getOptionalVaultProperty("authentication.userpass.password-wrapped");
         serverConfig.authentication.appRole.roleId = getOptionalVaultProperty("authentication.app-role.role-id");
         serverConfig.authentication.appRole.secretId = getOptionalVaultProperty("authentication.app-role.secret-id");
-        serverConfig.authentication.wrapToken = getOptionalVaultProperty("authentication.wrap-token");
+        serverConfig.authentication.appRole.secretIdWrapped = getOptionalVaultProperty("authentication.app-role.secret-id-wrapped");
         serverConfig.renewGracePeriod = getVaultDuration("renew-grace-period", DEFAULT_RENEW_GRACE_PERIOD);
         serverConfig.secretConfigCachePeriod = getVaultDuration("secret-config-cache-period",
                 DEFAULT_SECRET_CONFIG_CACHE_PERIOD);
--- a/extensions/vault/runtime/src/main/java/io/quarkus/vault/runtime/config/VaultConfigSource.java
+++ b/extensions/vault/runtime/src/main/java/io/quarkus/vault/runtime/config/VaultConfigSource.java
@@ -217,6 +217,7 @@ public class VaultConfigSource implements ConfigSource {
         serverConfig.authentication.userpass.password = getOptionalVaultProperty("authentication.userpass.password");
         serverConfig.authentication.appRole.roleId = getOptionalVaultProperty("authentication.app-role.role-id");
         serverConfig.authentication.appRole.secretId = getOptionalVaultProperty("authentication.app-role.secret-id");
+        serverConfig.authentication.wrapToken = getOptionalVaultProperty("authentication.wrap-token");
         serverConfig.renewGracePeriod = getVaultDuration("renew-grace-period", DEFAULT_RENEW_GRACE_PERIOD);
         serverConfig.secretConfigCachePeriod = getVaultDuration("secret-config-cache-period",
                 DEFAULT_SECRET_CONFIG_CACHE_PERIOD);
--- a/extensions/vault/runtime/src/main/java/io/quarkus/vault/runtime/config/VaultUserpassAuthenticationConfig.java
+++ b/extensions/vault/runtime/src/main/java/io/quarkus/vault/runtime/config/VaultUserpassAuthenticationConfig.java
@@ -15,9 +15,17 @@ public class VaultUserpassAuthenticationConfig {
     public Optional<String> username;
 
     /**
-     * Password for userpass auth method. This property is required when selecting the userpass authentication type.
+     * Password for userpass auth method. This property is required when selecting the userpass authentication type and not setting the password as wrapped.
+     *
+     * {@link VaultUserpassAuthenticationConfig#passwordWrapped}
      */
     @ConfigItem
     public Optional<String> password;
+
+    /**
+     * Wrapped token containing the password for userpass auth method. This flag is not mandatory, only required if password is wrapped.
+     */
+    @ConfigItem
+    public Optional<String> passwordWrapped;
 
 }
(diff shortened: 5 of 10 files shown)


vsevel:
hello, I have a different interpretation of the workflow. please see my comments https://github.com/quarkusio/quarkus/issues/7900#issuecomment-605608332 and https://github.com/quarkusio/quarkus/issues/7900#issuecomment-605615651

lordofthejars:
Yes it has sense too, I used the way I implemented. How about supporting
both?

vsevel:
but you do not need a client token to unwrap the wrapping token, do you?

lordofthejars:
yes you need

vsevel:
well, not in my tests. see my https://github.com/quarkusio/quarkus/issues/7900#issuecomment-605615651, I can unwrap with the CLI or the rest api, with no initial client token. 
Have I missed something?

lordofthejars:
didn't you set the token environment var?

vsevel:
no

lordofthejars:
how did you insert the wrapped token then?

vsevel:
```
VAULT_TOKEN=<wrapping_token> vault unwrap
# => <client_token>

VAULT_TOKEN=<client_token> vault kv get secret/dev
# => my-long-password
```

that is all is needed on the client side

lordofthejars:
yes but did you do that in the same terminal where you added the secret/dev or in a different terminal?

vsevel:
different of course. for the client side I am doing another: `docker exec -it dev-vault sh`
and to make double sure I did not have any previously context stored somewhere, I curled it from outside the container: `curl --header "X-Vault-Token: <wraping_token>" --request POST http://127.0.0.1:8200/v1/sys/wrapping/unwrap | jq`

vsevel:
I think some of the difficulty comes mixing 2 features:
 - response wrapping (which might use cubbyhole underneath, I do not know)
 - cubbyhole as a private store for your own token where you can put whatever sensitive data you have (this could be the age of your mother in law, or a token that gives you access to some information that is even more sensitive)

let's work on the scenario first before going into the impl.
my scenarii described in https://github.com/quarkusio/quarkus/issues/7900# are plain and entirely executable.
let's agree on the workflow of actions, and then the impl should be simple.

lordofthejars:
Yes, I'lll check it out when I have some time

vsevel:
some interesting insights to understand why it is working
```
> vault token create -policy=apps -wrap-ttl=120
# => <wrapping_token>

> vault token lookup <wrapping_token>

Key                 Value
---                 -----
accessor            nTyFvPUWyfhG8p8Hs9O7avut
creation_time       1585485701
creation_ttl        2m
display_name        n/a
entity_id           n/a
expire_time         2020-03-29T12:43:41.8518303Z
explicit_max_ttl    2m
id                  s.oiUYc5vOWAVuVP9PDg39t7SL
issue_time          2020-03-29T12:41:41.8518134Z
meta                <nil>
num_uses            1
orphan              true
path                auth/token/create
policies            [response-wrapping]
renewable           false
ttl                 1m46s
type                service

> vault policy read response-wrapping

path "cubbyhole/response" {
    capabilities = ["create", "read"]
}

path "sys/wrapping/unwrap" {
    capabilities = ["update"]
}
```

that is why you can always unwrap a wrapping token with itself.

the [unwrap](https://www.vaultproject.io/api-docs/system/wrapping-unwrap/) rest offers additional insights:

> This endpoint can be used by using a wrapping token as the client token in the API call, in which case the token parameter is not required; or, a different token with permissions to access this endpoint can make the call and pass in the wrapping token in the token parameter.

this means that somebody could also unwrap a wrapping token with its own token, granting it had the capabilities `update` on `sys/wrapping/unwrap`. 
I believe this is what you were trying to do:
- do an initial login => `lesser_privileged_client_token`
 - unwrap a wrapping token with that `lesser_privileged_client_token` => get whatever was wrapped in the wrapping token (this could be a client token, an `approle` `secret-id`, ...)

but this seems pointless since the wrapping token can unwrap itself.
may be they left that as a facility if your client is already logged in, and he wants to do all operations with his own token even when unwrapping, which seems confirmed by:

```
> vault policy read response-wrapping

...
# Allow a token to unwrap a response-wrapping token. This is a convenience to
# avoid client token swapping since this is also part of the response wrapping
# policy.
path "sys/wrapping/unwrap" {
    capabilities = ["update"]
}
...
```

vsevel:
may be you have had some experience with the [cubbyhole auth](https://www.hashicorp.com/blog/cubbyhole-authentication-principles/), but this approach seems outdated as described at the beginning of the article, and the [Cubbyhole Response Wrapping lesson](https://learn.hashicorp.com/vault/secrets-management/sm-cubbyhole) might be just outdated as well with respect to client operations steps 1 and 2, which very much look like the `perm` and `temp` tokens of the old `cubbyhole auth`:
> Create a token with default policy
> Authenticate with Vault using this default token (less privileged token)

from a quarkus perspective, there might be actually 3 use cases worth supporting:
 - [cubbyhole](https://www.vaultproject.io/docs/secrets/cubbyhole/index.html#cubbyhole-secrets-engine) programmatic api (because why not)
 - response wrapping applied to secret auth params (.i.e what I have described  in https://github.com/quarkusio/quarkus/issues/7900: wrap `secret-id` for `approle`, `password` for `userpass`, ...)
 - response wrapping in general, which means the [unwrap](https://www.vaultproject.io/api-docs/system/wrapping-unwrap/) operation, plus the ability to wrap responses. For instance vincent wants to share a secret with alex. vincent reads the secret using response wrapping and puts the wrapping token on disk somewhere for alex to find it. alex reads it from disk and unwrap it to get the secret.

vincent:
```
/ # vault kv get -wrap-ttl=60s secret/myapp 
Key                              Value
---                              -----
wrapping_token:                  s.q91VfGbfUdI6sKeviGHqrG5M
wrapping_accessor:               7G5a13HJwTLxHR2qEbj9idlz
wrapping_token_ttl:              1m
wrapping_token_creation_time:    2020-03-29 16:30:27.011161051 +0000 UTC
wrapping_token_creation_path:    secret/data/myapp
```

alex:
```
/ # VAULT_TOKEN=s.q91VfGbfUdI6sKeviGHqrG5M vault unwrap
Key         Value
---         -----
data        map[password:pa$$w0rd]
metadata    map[created_time:2020-03-29T16:29:53.576739463Z deletion_time: destroyed:false version:1]
```

lordofthejars:
Ok for this issue, let's implement the point number 1

lordofthejars:
@vsevel  Ok, implemented. Just review and if you agree, then I move forward to write tests and docs. As I said previously, I think you are right and this is how it should be implemented.

extensions/vault/runtime/src/main/java/io/quarkus/vault/runtime/config/VaultConfigSource.java
```diff
@@ -215,8 +215,10 @@ private VaultRuntimeConfig loadRuntimeConfig() {
                 DEFAULT_KUBERNETES_JWT_TOKEN_PATH);
         serverConfig.authentication.userpass.username = getOptionalVaultProperty("authentication.userpass.username");
         serverConfig.authentication.userpass.password = getOptionalVaultProperty("authentication.userpass.password");
+        serverConfig.authentication.userpass.passwordWrapped = getOptionalVaultProperty("authentication.userpass.password-wrapped");
```
vsevel: `userpass.password-wrapping-token` but let's review this later

extensions/vault/runtime/src/main/java/io/quarkus/vault/runtime/config/VaultConfigSource.java
```diff
@@ -215,8 +215,10 @@ private VaultRuntimeConfig loadRuntimeConfig() {
...
         serverConfig.authentication.appRole.roleId = getOptionalVaultProperty("authentication.app-role.role-id");
         serverConfig.authentication.appRole.secretId = getOptionalVaultProperty("authentication.app-role.secret-id");
+        serverConfig.authentication.appRole.secretIdWrapped = getOptionalVaultProperty("authentication.app-role.secret-id-wrapped");
```
vsevel: `secret-id-wrapping-token`

extensions/vault/runtime/src/main/java/io/quarkus/vault/runtime/config/VaultUserpassAuthenticationConfig.java
```diff
@@ -15,9 +15,17 @@
...
+     * {@link VaultUserpassAuthenticationConfig#passwordWrapped}
      */
     @ConfigItem
     public Optional<String> password;
 
+    /**
+     * Wrapped token containing the password for userpass auth method. This flag is not mandatory, only required if password is wrapped.
+     */
+    @ConfigItem
+    public Optional<String> passwordWrapped;
```
vsevel: `password-wrapping-token`, but let^s review this later

extensions/vault/runtime/src/main/java/io/quarkus/vault/runtime/config/VaultAppRoleAuthenticationConfig.java
```diff
@@ -15,9 +15,17 @@
...
+     * {@link VaultAppRoleAuthenticationConfig#secretIdWrapped}
      */
     @ConfigItem
     public Optional<String> secretId;
 
+    /**
+     * Wrapped token containing the secret Id for AppRole auth method. This flag is not mandatory, only required if secret id is wrapped.
+     */
+    @ConfigItem
+    public Optional<String> secretIdWrapped;
```
vsevel: `secret-id-wrapping-token`

extensions/vault/runtime/src/main/java/io/quarkus/vault/runtime/client/dto/sys/VaultUnwrapAuth.java
```diff
@@ -0,0 +1,6 @@
+package io.quarkus.vault.runtime.client.dto.sys;
+
+import io.quarkus.vault.runtime.client.dto.auth.AbstractVaultAuthAuth;
+
+public class VaultUnwrapAuth extends AbstractVaultAuthAuth<Object> {
```
vsevel: the response is going to be generic

extensions/vault/runtime/src/main/java/io/quarkus/vault/runtime/client/dto/sys/VaultUnwrapResult.java
```diff
@@ -0,0 +1,34 @@
...
+ * "identity_policies": null,
+ * "metadata": null,
+ * "orphan": false,
+ * "entity_id": "",
+ * "lease_duration": 2764800,
+ * "renewable": true
+ * }
+ * }
+ */
+public class VaultUnwrapResult extends AbstractVaultDTO<Object, VaultUnwrapAuth> {
```
vsevel: a `VaultUnwrapResult` cannot be made of a `VaultUnwrapAuth`. it has to be generic, because response could be wrapped.

vsevel (changes requested):

extensions/vault/runtime/src/main/java/io/quarkus/vault/runtime/client/dto/sys/VaultUnwrapResult.java
lordofthejars: Ok I understand in general what you are mention here and I agree, what is not clear to me and I need to play is reproduce the example you provided before of `approle` with wrapped token and see all the outputs to see how are different between them and how we can implement it. I don't know when I will have time to do it.

About renames I agree with you to follow the Vault terminology, although my guess is that i terms of Docs, Vault documentation uses different concepts for the same, but let's use what you suggest here which is more clear.

sberyozkin: @lordofthejars +1, we've touched upon the same issue in the vault health PR, we don't need to match the low level Vault API property names at the Quarkus level, users will appreciate it.

lordofthejars: I am not sure if I will be able to work on this in the near time, I mean let's collaborate all of us in this.

lordofthejars: @vsevel do you think we could arrange a call (1 hour call) and do pair programming and fix everything?

sberyozkin:
@lordofthejars If you could find few mins and do `mvn process-sources` in `extensions/vault`  and push it then it will help to see if the tests pass, the build fails with the formatting error.  Thanks

lordofthejars:
@sberyozkin this is a work in progress and I have not added any test in fact, this PR is more something for discussing.

sberyozkin:
@lordofthejars That is fine. Just FYI anyway, @stuartwdouglas's PR has just been merged, so without it  the native tests would likely fail in the unrelated OIDC extension, so whenever you are happy with the status, please rebase.
Let me also add `WIP:` to the title for now, please remove it when ready. And thanks for making a go and pushing this feature closer to Quarkus Vault :-). I'm sure some Vault users in the Quarkus community will appreciate it in due time.

extensions/vault/runtime/src/main/java/io/quarkus/vault/runtime/client/dto/sys/VaultUnwrapResult.java
sberyozkin: @lordofthejars I'm also quite stretched but will be happy to contribute with some comments :-). There is no rush, unless Vincent prefers it merged asap. Thanks

sberyozkin:
It is really great we now have 2 Vault experts contributing :-)

extensions/vault/runtime/src/main/java/io/quarkus/vault/runtime/client/dto/sys/VaultUnwrapResult.java
lordofthejars: Sure no worries, as I said to @vsevel if we can do a pair programming session together will be the best way to proceed as with one hour we can have everything done and reviewed at the same time.

vsevel: @lordofthejars I doubt we can get close to merge in 1 hour, considering even the early code needs some changes, and adding on top of that testing and documentation.
by experience it is going to take some effort. if you do not have time now, may be this is something that you can put aside for a while. I think there is no rush. I have given my input on the applicability, and no user has asked for it.
If you decided to pursue it, one thing I did to help you out is providing you with example responses fo the 2 main use cases: approle and client-token:

approle wrapping token creation:
```
/ # vault write -wrap-ttl=60s -f auth/approle/role/myapp/secret-id
Key                              Value
---                              -----
wrapping_token:                  s.QdjujN8PGGN8EICldXpFnSYO
wrapping_accessor:               NCk8gkK90xLNzFMcnTq68nY8
wrapping_token_ttl:              1m
wrapping_token_creation_time:    2020-03-31 16:29:05.5958584 +0000 UTC
wrapping_token_creation_path:    auth/approle/role/myapp/secret-id
```

approle unwrap response:
```
curl --header "X-Vault-Token: s.QdjujN8PGGN8EICldXpFnSYO" --request POST http://127.0.0.1:8200/v1/sys/wrapping/unwrap | jq
{
  "request_id": "5829832d-113a-48ed-7881-739d470fdbdf",
  "lease_id": "",
  "renewable": false,
  "lease_duration": 0,
  "data": {
    "secret_id": "ef8e1ebc-3be1-7bb7-3166-8b63b6289449",
    "secret_id_accessor": "e70b94da-e1ab-03c1-0fbc-8e73218f83b1"
  },
  "wrap_info": null,
  "warnings": null,
  "auth": null
}
```

client-token wrapping token creation:
```
/ # vault token create -wrap-ttl=60s -policy=myapp
Key                              Value
---                              -----
wrapping_token:                  s.xSr3IHx9RXxpSSwCctTd6n02
wrapping_accessor:               8H0BUXHPFv6LVIrXa9H9XNWb
wrapping_token_ttl:              1m
wrapping_token_creation_time:    2020-03-31 16:26:49.0405113 +0000 UTC
wrapping_token_creation_path:    auth/token/create
wrapped_accessor:                xoFYP9EgPi9I3xa7cDB3bm5v
```

client-token unwrap response:
```
curl --header "X-Vault-Token: s.xSr3IHx9RXxpSSwCctTd6n02" --request POST http://127.0.0.1:8200/v1/sys/wrapping/unwrap | jq
{
  "request_id": "7dd87b91-0c6b-c837-7918-4c377f7e635a",
  "lease_id": "",
  "renewable": false,
  "lease_duration": 0,
  "data": null,
  "wrap_info": null,
  "warnings": [
    "Policy \"myapp\" does not exist"
  ],
  "auth": {
    "client_token": "s.DiaodHOYFPxdsWbKtKX6qXr6",
    "accessor": "xoFYP9EgPi9I3xa7cDB3bm5v",
    "policies": [
      "default",
      "myapp"
    ],
    "token_policies": [
      "default",
      "myapp"
    ],
    "metadata": null,
    "lease_duration": 2764800,
    "renewable": true,
    "entity_id": "",
    "token_type": "service",
    "orphan": false
  }
}
```

so we can see that for the approle, the output is the original response, if it had not been wrapped. so in effect it is the response for the [Generate New Secret ID](https://www.vaultproject.io/api-docs/auth/approle/#sample-response-4)

since this operation has not been coded in `VaultClient` the data structure does not exist.
it is the same for the token creation.

however I do not think there is any challenge. just create any data structure you might need when unwrapping, then create a VaultClient `unwrap` operation that returns an `Object` with no unmarshalling (or just as jsonp objects), and depending on the use case unmarshall it on the right datastructure.

I do not think it is a huge piece of work, but it is definitely going to take some effort.

sberyozkin:
#8411 is now available so closing this one, thanks

PR #8252, opened 2020-03-29
Load Hamcrest parent first

### LINKED ISSUE #8251: ClassCastException in @ParameterizedTest: Base Class from Different Class Loader
**Describe the bug**
Running a `@QuarkusTest` fails with a `java.lang.ClassCastException`, see [MWE](https://github.com/schulzp/quarkus-classcast-mwe).

**Expected behavior**
The tests passes.

**Actual behavior**
When using `@ParameterizedTest` with `@MethodSource`, see [MWE](https://github.com/schulzp/quarkus-classcast-mwe), the test fails with the following exception:
`java.lang.ClassCastException: class org.hamcrest.core.Is cannot be cast to class org.hamcrest.Matcher (org.hamcrest.core.Is is in unnamed module of loader 'app'; org.hamcrest.Matcher is in unnamed module of loader io.quarkus.bootstrap.classloading.QuarkusClassLoader @7188af83)`

However, if the test parameters are [mapped manually](https://github.com/schulzp/quarkus-classcast-mwe/blob/master/src/test/java/com/example/ExampleResourceTest.java#L28) (using a regular `@Test`) it passes.

**To Reproduce**
1. Check out the [MWE](https://github.com/schulzp/quarkus-classcast-mwe)
2. run `./mvnw verify`

**Configuration**
```properties
# no properties are set explicitly
```

   ```
   Darwin Kernel Version 19.3.0; root:xnu-6153.81.5~1/RELEASE_X86_64 x86_64
   ```
 - Output of `java -version`: `11.0.1, vendor: Oracle Corporation`
 - Quarkus version or git rev: `1.3.1.Final`
   ```
   Apache Maven 3.6.3 (cecedd343002696d0abb50b32b541b8a6ba2883f)
   Java version: 11.0.1, vendor: Oracle Corporation, runtime: [...]
   Default locale: de_DE, platform encoding: UTF-8
   OS name: "mac os x", version: "10.15.3", arch: "x86_64", family: "mac"
   ```
**Additional context**
The behavior occurred when I upgraded quarkus from 1.2.1.Final to 1.3.0.Final/1.3.1.Final.

geoand:
Fixes: #8251

### DIFF
--- a/core/runtime/pom.xml
+++ b/core/runtime/pom.xml
@@ -127,6 +127,7 @@
                         <parentFirstArtifact>org.junit.jupiter:junit-jupiter-params</parentFirstArtifact>
                         <parentFirstArtifact>org.junit.platform:junit-platform-commons</parentFirstArtifact>
                         <parentFirstArtifact>org.junit.platform:junit-platform-engine</parentFirstArtifact>
+                        <parentFirstArtifact>org.hamcrest:hamcrest</parentFirstArtifact>
                     </parentFirstArtifacts>
                     <excludedArtifacts>
                         <excludedArtifact>io.smallrye:smallrye-config</excludedArtifact>
--- a/integration-tests/main/src/test/java/io/quarkus/it/main/ParameterizedHamcrestTestCase.java
+++ b/integration-tests/main/src/test/java/io/quarkus/it/main/ParameterizedHamcrestTestCase.java
@@ -1,0 +1,46 @@
+package io.quarkus.it.main;
+
+import static io.restassured.RestAssured.given;
+import static org.hamcrest.CoreMatchers.is;
+
+import java.util.Collections;
+import java.util.Map;
+import java.util.stream.Stream;
+
+import org.hamcrest.Matcher;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+
+import io.quarkus.test.junit.QuarkusTest;
+
+@QuarkusTest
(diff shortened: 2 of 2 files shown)


stuartwdouglas:
This does not fix the underlying problem, which is that the 'args' method is invoked on the instance in the original CL.

geoand:
Indeed, `args` does get invoked on the original CL instead of the TCCL.

After going through the JUnit source code, I didn't see a way we could use to intercept the call.
The method lookup code is at https://github.com/junit-team/junit5/blob/r5.6.1/junit-jupiter-params/src/main/java/org/junit/jupiter/params/provider/MethodArgumentsProvider.java#L58 (`MethodArgumentsProvider` is determined by the fact that `@MethodSource` is annotated with `@ArgumentsSource(MethodArgumentsProvider.class)`) and as you see it uses `context.getRequiredTestClass()` to obtain the class. 
So short of us providing the functionality in `TestTemplateInvocationContextProvider`, I don't see how we can change this behavior.

@stuartwdouglas what do you propose we do?

stuartwdouglas:
I don't know. I also looked and could not see any obvious way of changing how this works (it may be possible to do some reflections hacks to change the values in the extension context, but I am not actually sure if this would work).

Longer term I think we should try and push to have some way to intercept these calls, I have asked on the JUnit gitter about it.

geoand:
I'll get on gitter and check out the conversation

stuartwdouglas:
It's not so much a conversation, as an unanswered question

geoand:
I'll keep an eye out nonetheless :wink:

gastaldi:
For the record, in the chat it was mentioned that https://github.com/junit-team/junit5/issues/201 could fix this

geoand:
Superseded by #9348

PR #8264, opened 2020-03-30
Document the flags to make the full build faster

cescoffier:
As this is a recurring question, I changed the contributing.md file.


geoand (approved):
Makes perfect sense!

Issue #8277, opened 2020-03-30
Support java 14 records in quarkus:dev


arvidvillen:
**Description**
Support the usage of Java 14 records (preview feature) in quarkus:dev.

I’m trying to get java 14 records to work in Quakrus and I’m (quite easily) able to get the tests to pass and the final jar to run but DevMode fails with the following error:
```
Exception in thread "main" java.lang.RuntimeException: java.lang.RuntimeException: java.lang.RuntimeException: java.lang.UnsupportedOperationException: This feature requires ASM8_EXPERIMENTAL
	at io.quarkus.dev.DevModeMain.start(DevModeMain.java:117)
	at io.quarkus.dev.DevModeMain.main(DevModeMain.java:54)
Caused by: java.lang.RuntimeException: java.lang.RuntimeException: java.lang.UnsupportedOperationException: This feature requires ASM8_EXPERIMENTAL
	at io.quarkus.bootstrap.app.CuratedApplication.runInCl(CuratedApplication.java:134)
	at io.quarkus.bootstrap.app.CuratedApplication.runInAugmentClassLoader(CuratedApplication.java:84)
	at io.quarkus.dev.DevModeMain.start(DevModeMain.java:113)
	... 1 more
Caused by: java.lang.RuntimeException: java.lang.UnsupportedOperationException: This feature requires ASM8_EXPERIMENTAL
	at io.quarkus.dev.IsolatedDevModeMain.accept(IsolatedDevModeMain.java:259)
	at io.quarkus.dev.IsolatedDevModeMain.accept(IsolatedDevModeMain.java:39)
	at io.quarkus.bootstrap.app.CuratedApplication.runInCl(CuratedApplication.java:131)
	... 3 more
Caused by: java.lang.UnsupportedOperationException: This feature requires ASM8_EXPERIMENTAL
	at org.objectweb.asm.ClassVisitor.visitRecordComponentExperimental(ClassVisitor.java:304)
	at org.objectweb.asm.ClassReader.readRecordComponent(ClassReader.java:929)
	at org.objectweb.asm.ClassReader.accept(ClassReader.java:704)
	at org.objectweb.asm.ClassReader.accept(ClassReader.java:401)
	at io.quarkus.dev.JavaCompilationProvider.getSourcePath(JavaCompilationProvider.java:78)
	at io.quarkus.dev.ClassLoaderCompiler.findSourcePath(ClassLoaderCompiler.java:183)
[...]

PR #8279, opened 2020-03-30
Allow dev-mode to use preview Java features

geoand:
Relates to: #8277


gastaldi (approved):

gastaldi:
Make sure the Gradle plugin also works with this setting

geoand:
I know it doesn't 😉.

We'll need to follow up for that since I don't think we have the compiler
flag grabbing mechanism.

Issue #8093, opened 2020-03-23
Generating a fieldRef env var with the Kubernetes Extension is not working


sebastienblanc:
**Describe the bug**
In my properties I put : 
```
kubernetes.env-vars[0].name=GREETING_MYADRESS
kubernetes.env-vars[0].field=metadata.name
```

This should generate : 

```
- name: "GREETING_MYADRESS"
          valueFrom:
            fieldRef:
              fieldPath: "metadata.name"
```

But nothing is generated in `target/kubernetes/kubernetes.yaml`
**Expected behavior**
in  `target/kubernetes/kubernetes.yaml`
```
- name: "GREETING_MYADRESS"
          valueFrom:
            fieldRef:
              fieldPath: "metadata.name"
```

**Actual behavior**
It's not generated in the yaml file

**To Reproduce**
1. add properties as mentioned in the beginning of the report 
2. `mvn clean package`
3. inspect the content of  `target/kubernetes/kubernetes.yaml`

**Environment (please complete the following information):**
Quarkus 1.3.0.Final

@iocanel @geoand

PR #8293, opened 2020-03-31
test (#8093): Add test for env from field.

iocanel:
I should have added this as part of the #8093 fix. 

It's never too late to add tests.


geoand (approved):
Thanks!

PR #8296, opened 2020-03-31
ArC - select alternatives through the MicroProfile Config

### LINKED ISSUE #8135: Allow users to select alternatives though the MicroProfile Config
Currently, an alternative must be annotated with `@Priority` or `@io.quarkus.arc.AlternativePriority`. It should be possible to use MP Config to select an alternative as well.

The simplest approach would be to list all selected alternatives, e.g. `quarkus.arc.selected-alternatives=org.acme.Foo`. It's not possible to specify the priority but could take the spec approach where components annotated with `@Priority` take precedence.

mkouba:
- resolves #8135


mkouba:
This PR is a breaking change. BEFORE: a bean class annotated with `@Alternative` but not selected by `@Priority` resulted in a deployment problem. AFTER: such a bean class is ignored and INFO message is logged.

stuartwdouglas:
This is not really a breaking change, as there are no valid deployments that are now broken.

mkouba:
> This is not really a breaking change, as there are no valid deployments that are now broken.

Ok. It's breaking from the behavior POV - the deployment that failed before does not fail anymore. So it does not break existing deployments but in theory a user could rely on the fact that a similar config fails during build and now it's ignored... so it's a "very minor breaking change" ;-).

docs/src/main/asciidoc/cdi-reference.adoc
```diff
@@ -408,6 +426,33 @@ public class CustomTracerConfiguration {
...
+In CDI, an alternative bean may be selected either globally for an application by means of `@Priority`, or for a bean archive using a `beans.xml` descriptor.
+Quarkus has a simplified bean discovery and the content of `beans.xml` is ignored.
+
+The disadvantage of `@Priority` is that it has `@Target({ TYPE, PARAMETER })` and so it cannot be used for producer methods and fields.
+To address this problem and to simplify the code Quarkus provides the `io.quarkus.arc.AlternativePriority` annotation.
+It's basically a shortcut for `@Alternative` plus `@Priority`.
+Additionaly, it can be used for producers.
+
+However, it is also possible to select alternatives for an application using the unified configuration.
+The `quarkus.arc.selected-alternatives` property accepts a list of string values.
```
stuartwdouglas: How does this work for producer methods? You talk about them in the paragraph above but this section only mentions classes.

stuartwdouglas (approved):
LGTM

docs/src/main/asciidoc/cdi-reference.adoc
mkouba: You should state the declaring class for producers - it's the same as in `beans.xml`. I'll update the docs.

mkouba: Is it better now?

manovotn (commented):
Looks good. I have three notes:
* You don't seem to document the fact that this config approach overrides any other applied means of defining priority
* When you do ^^, you log `info`, I think this could be lower. It is intended and expected and might even be a use case in which case you maybe don't want to see this repeatedly?
* Shouldn't we blow up when you have incorrect config? Like empty string for instance.

mkouba:
> You don't seem to document the fact that this config approach overrides any other applied means of defining priority

Good catch!

> When you do ^^, you log info, I think this could be lower. It is intended and expected and might even be a use case in which case you maybe don't want to see this repeatedly?

I'm not so sure. I think it's good to know  that a `@Priority` is overriden and I also don't expect that many alternatives ;-).

> Shouldn't we blow up when you have incorrect config? Like empty string for instance.

I don't think it's worth the effort. A value that does not contain dot, `.*` or `.**` is considered a simple name and thus would not match anything.

manovotn (approved):

PR #8302, opened 2020-03-31
Properly handle Optional<List<String>> in @ConfigProperties

### LINKED ISSUE #8294: Using @ConfigProperty with Optional<List<String>> throw an exception
**Describe the bug**
Using:
```
    @ConfigProperty(name = "items3")
    Optional<List<String>> getItems3();
```

**Expected behavior**
Return an empty list when unset

**Actual behavior**
Throws an exception:
```
java.lang.IllegalArgumentException: No Converter registered for interface java.util.List

	at io.smallrye.config.SmallRyeConfig.lambda$getConverter$1(SmallRyeConfig.java:202)
	at java.base/java.util.concurrent.ConcurrentHashMap.computeIfAbsent(ConcurrentHashMap.java:1705)
	at io.smallrye.config.SmallRyeConfig.getConverter(SmallRyeConfig.java:199)
```

**To Reproduce**
Reproducer:
https://github.com/ia3andy/check-list-config-empty

 - Quarkus version or git rev: 1.3.1.Final

geoand:
Fixes #8294

### DIFF
--- a/extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/configproperties/InterfaceConfigPropertiesUtil.java
+++ b/extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/configproperties/InterfaceConfigPropertiesUtil.java
@@ -1,6 +1,7 @@
 package io.quarkus.arc.deployment.configproperties;
 
 import static io.quarkus.arc.deployment.configproperties.ConfigPropertiesUtil.createReadMandatoryValueAndConvertIfNeeded;
+import static io.quarkus.arc.deployment.configproperties.ConfigPropertiesUtil.createReadOptionalValueAndConvertIfNeeded;
 import static io.quarkus.arc.deployment.configproperties.ConfigPropertiesUtil.determineSingleGenericType;
 
 import java.lang.reflect.Modifier;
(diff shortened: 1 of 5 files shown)


extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/configproperties/InterfaceConfigPropertiesUtil.java
```diff
@@ -134,12 +135,32 @@ static String generateImplementationForInterfaceConfigProperties(ClassInfo origi
 
                             Type genericType = determineSingleGenericType(returnType,
                                     method.declaringClass().name());
-                            ResultHandle result = methodCreator.invokeInterfaceMethod(
-                                    MethodDescriptor.ofMethod(Config.class, "getOptionalValue", Optional.class, String.class,
-                                            Class.class),
-                                    config, methodCreator.load(fullConfigName),
-                                    methodCreator.loadClass(genericType.name().toString()));
-                            methodCreator.returnValue(result);
+
```
ia3andy: @geoand no way to factorize this with the `ClassConfigPropertiesUtil`?

ia3andy (approved):
Beside the duplication of code which I think you would have avoided in the first place if it was possible, I tested it in my reproducer and everything is fine!!

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/configproperties/InterfaceConfigPropertiesUtil.java
geoand: Sure, do you want to do that as an exercise to get your hands dirty with this code?

ia3andy: It looks like a trap 🤔

ia3andy: Sure

ia3andy:
Regarding the "duplication of code", we discussed a possible refactor with @geoand and reach that conclusion:
The code is different enough to justify a bit of duplication, at the same time it sucks to risk a diff in behaviors. Here are the reasons not to refactor:
- A refactor would not be really trivial and without the result is not guaranteed to be way better
- This code is not supposed to change much once stable
- Having some more testing could reduce the "risk of diff in behavior"

PR #8303, opened 2020-03-31
Parameterised docker image for DBs

rsvoboda:
Parameterised docker image for DBs

### DIFF
--- a/integration-tests/jpa-postgresql/pom.xml
+++ b/integration-tests/jpa-postgresql/pom.xml
@@ -13,6 +13,11 @@
     <artifactId>quarkus-integration-test-jpa-postgresql</artifactId>
     <name>Quarkus - Integration Tests - JPA - PostgreSQL</name>
     <description>Module that contains JPA related tests running with the PostgreSQL database</description>
+
+    <properties>
+        <postgres.image>postgres:10.5</postgres.image>
+    </properties>
+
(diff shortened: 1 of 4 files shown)


geoand (approved):
LGTM

integration-tests/jpa-postgresql/pom.xml
```diff
@@ -169,7 +174,7 @@
                         <configuration>
                             <images>
                                 <image>
-                                    <name>postgres:10.5</name>
+                                    <name>${postgresql.image}</name>
```
gsmet: It should be consistent with `postgres.url` a few lines above. So either we fix the existing one (and the CI jobs) or we stick to `postgres` as the prefix.

gsmet (dismissed):

integration-tests/jpa-postgresql/pom.xml
rsvoboda: Updated and pulled up the url property so all 4 modules are in sync

rsvoboda:
@gsmet PR updated, but CI seems to be having hard time

geoand:
Yeah, I saw that in other PRs as well...

rsvoboda:
@gsmet added triage/backport-1.3? label, this is needed for checks against different versions of databases

PR #8307, opened 2020-03-31
set groups attribute name as configurable

adrianfiedler:


extensions/elytron-security/runtime/src/main/java/io/quarkus/elytron/security/runtime/DefaultRoleDecoder.java
```diff
@@ -21,7 +22,8 @@
  */
 public class DefaultRoleDecoder {
 
-    private static final String DEFAULT_ATTRIBUTE_NAME = "groups";
+    @ConfigProperty(name = "quarkus.security.default-role-decoder-name", defaultValue="groups")
```
sberyozkin: It might make sense to make this property name more neutral I guess, so that the custom mappers can also rely on it.
How about `quarkus.security.roles-claim-name`  ?

sberyozkin:
@adrianfiedler please also do `mvn process-sources`. Having a simple unit test would also be fine, you can make the affected method package private

extensions/elytron-security/runtime/src/main/java/io/quarkus/elytron/security/runtime/DefaultRoleDecoder.java
adrianfiedler: Thanks for your feedback, I changed it now to `quarkus.security.roles-claim-name`

sberyozkin:
@adrianfiedler thanks, can you please now add a small unit test for `DefaultRoleDecoder` ? I guess `AuthorizationIdentity` can be mocked.
CC @gastaldi

sberyozkin (approved):
Thanks

PR #8312, opened 2020-03-31
Bug/kubernetes extension replicas fix

### LINKED ISSUE #8273: Kubernetes extension replicas configuration
**Describe the bug**

Im trying out the Kubernetes extension have stumbled upon a bug. I have the following dependencies

     <dependency>
      <groupId>io.quarkus</groupId>
      <artifactId>quarkus-kubernetes</artifactId>
    </dependency>
    <dependency>
      <groupId>io.quarkus</groupId>
      <artifactId>quarkus-container-image-jib</artifactId>
    </dependency>
    <dependency>
      <groupId>io.quarkus</groupId>
      <artifactId>quarkus-config-yaml</artifactId>
    </dependency>
I preferred the yaml config option and have the following configuration.

```yaml
quarkus:
  kubernetes:
    part-of: hr-app
    name: employee-service
    version: 1.0.0
    replicas: 3
```

Yet the generated Kubernetes manifest file does not set the replicas as 3

```yaml
apiVersion: "apps/v1"
kind: "Deployment"
metadata:
  annotations:
    app.quarkus.io/build-timestamp: "2020-03-27 - 21:43:23 +0000"
    app.quarkus.io/commit-id: "809f76f1b476ea1dc07cf311d079baceeb414d21"
  labels:
    app.kubernetes.io/name: "employee-service"
    app.kubernetes.io/part-of: "hr-app"
    app.kubernetes.io/version: "1.0.0"
  name: "employee-service"
spec:
  replicas: 1
  selector:
    matchLabels:
      app.kubernetes.io/name: "employee-service"
      app.kubernetes.io/part-of: "hr-app"
      app.kubernetes.io/version: "1.0.0"
```

**Expected behavior**
Generated Kubernetes manifest file replicas are set correctly.

**Actual behavior**
Generated Kubernetes manifest file replicas are not set correctly. 

It seems as quarkus.kubernetes.replicas is ignored.

**To Reproduce**
1. Generate project

 mvn io.quarkus:quarkus-maven-plugin:1.3.1.Final:create \
    -DprojectGroupId=com.acme \
    -DprojectArtifactId=getting-started \
    -DprojectVersion=1.0 \
    -DclassName="org.my.group.MyResource"

2. Add extensions
./mvnw quarkus:add-extension -Dextensions="quarkus-kubernetes,quarkus-container-image-jib,quarkus-config-yaml"

3. Add the following configuration to application.yaml to resources
```yaml
quarkus:
  kubernetes:
    part-of: getting-started
    name: getting-started
    version: 1.0.0
    replicas: 3
```
3. Package
./mvnw package
4. Observe generated Kubernetes Manifest 
target/kubernetes/kubernetes.yml

**Configuration**
```yaml
quarkus:
  kubernetes:
    part-of: getting-started
    name: getting-started
    version: 1.0.0
    replicas: 3
```

**Environment:**
 Linux localhost.localdomain 5.0.9-301.fc30.x86_64 #1 SMP Tue Apr 23 23:57:35 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux
 - Output of `java -version`:
 N/A
 - Quarkus version or git rev:
 1.3.1.Final
 Apache Maven 3.6.3

cemnura:
Fixes #8273

### DIFF
--- a/extensions/kubernetes/vanilla/deployment/src/main/java/io/quarkus/kubernetes/deployment/KubernetesApplyReplicasDecorator.java
+++ b/extensions/kubernetes/vanilla/deployment/src/main/java/io/quarkus/kubernetes/deployment/KubernetesApplyReplicasDecorator.java
@@ -1,0 +1,21 @@
+package io.quarkus.kubernetes.deployment;
+
+import io.dekorate.kubernetes.decorator.ApplyReplicasDecorator;
+import io.dekorate.kubernetes.decorator.Decorator;
+
+public class KubernetesApplyReplicasDecorator extends ApplyReplicasDecorator {
+
+    public KubernetesApplyReplicasDecorator(int replicas) {
+        super(replicas);
+    }
+
+    public KubernetesApplyReplicasDecorator(String deploymentName, int replicas) {
+        super(deploymentName, replicas);
+    }
+
+    @Override
+    public Class<? extends Decorator>[] after() {
+        return new Class[] { ApplyReplicasDecorator.class };
+    }
+
+}
--- a/extensions/kubernetes/vanilla/deployment/src/main/java/io/quarkus/kubernetes/deployment/KubernetesConfig.java
+++ b/extensions/kubernetes/vanilla/deployment/src/main/java/io/quarkus/kubernetes/deployment/KubernetesConfig.java
@@ -96,6 +96,12 @@ public class KubernetesConfig implements PlatformConfiguration {
     Map<String, PortConfig> ports;
 
     /**
+     * The number of desired pods
+     */
+    @ConfigItem(defaultValue = "1")
+    Integer replicas;
+
+    /**
      * The type of service that will be generated for the application
      */
     @ConfigItem(defaultValue = "ClusterIP")
@@ -250,6 +256,10 @@ public class KubernetesConfig implements PlatformConfiguration {
 
     public Map<String, PortConfig> getPorts() {
         return ports;
+    }
+
+    public Integer getReplicas() {
+        return replicas;
     }
 
     public ServiceType getServiceType() {
--- a/extensions/kubernetes/vanilla/deployment/src/main/java/io/quarkus/kubernetes/deployment/KubernetesProcessor.java
+++ b/extensions/kubernetes/vanilla/deployment/src/main/java/io/quarkus/kubernetes/deployment/KubernetesProcessor.java
@@ -403,6 +403,10 @@ class KubernetesProcessor {
         configMap.put(OPENSHIFT, openshiftConfig);
         configMap.put(KNATIVE, knativeConfig);
 
+        //Replicas
+        session.resources().decorate(new KubernetesApplyReplicasDecorator(kubernetesName, kubernetesConfig.getReplicas()));
+        session.resources().decorate(new OpenshiftApplyReplicasDecorator(openshiftConfig.getReplicas()));
+
         kubernetesAnnotations.forEach(a -> {
             session.resources().decorate(a.getTarget(), new AddAnnotationDecorator(new Annotation(a.getKey(), a.getValue())));
         });
--- a/extensions/kubernetes/vanilla/deployment/src/main/java/io/quarkus/kubernetes/deployment/OpenshiftApplyReplicasDecorator.java
+++ b/extensions/kubernetes/vanilla/deployment/src/main/java/io/quarkus/kubernetes/deployment/OpenshiftApplyReplicasDecorator.java
@@ -1,0 +1,17 @@
+package io.quarkus.kubernetes.deployment;
+
+import io.dekorate.kubernetes.decorator.Decorator;
+import io.dekorate.openshift.decorator.ApplyReplicasDecorator;
+
+public class OpenshiftApplyReplicasDecorator extends ApplyReplicasDecorator {
+
+    public OpenshiftApplyReplicasDecorator(int replicas) {
+        super(replicas);
+    }
+
+    @Override
+    public Class<? extends Decorator>[] after() {
+        return new Class[] { ApplyReplicasDecorator.class };
+    }
+
+}
--- a/extensions/kubernetes/vanilla/deployment/src/main/java/io/quarkus/kubernetes/deployment/OpenshiftConfig.java
+++ b/extensions/kubernetes/vanilla/deployment/src/main/java/io/quarkus/kubernetes/deployment/OpenshiftConfig.java
@@ -97,6 +97,12 @@ public class OpenshiftConfig implements PlatformConfiguration {
     Map<String, PortConfig> ports;
 
     /**
+     * The number of desired pods
+     */
+    @ConfigItem(defaultValue = "1")
+    Integer replicas;
+
+    /**
      * The type of service that will be generated for the application
      */
     @ConfigItem(defaultValue = "ClusterIP")
@@ -239,6 +245,10 @@ public class OpenshiftConfig implements PlatformConfiguration {
 
     public Optional<String> getHost() {
         return host;
+    }
+
+    public Integer getReplicas() {
+        return replicas;
     }
 
     public Map<String, PortConfig> getPorts() {
--- a/integration-tests/kubernetes/standard/src/test/java/io/quarkus/it/kubernetes/BasicOpenshiftTest.java
+++ b/integration-tests/kubernetes/standard/src/test/java/io/quarkus/it/kubernetes/BasicOpenshiftTest.java
@@ -44,7 +44,7 @@ public class BasicOpenshiftTest {
                 assertThat(m.getName()).isEqualTo("basic-openshift");
                 assertThat(m.getLabels().get("app.openshift.io/runtime")).isEqualTo("quarkus");
             });
-
+            assertThat(h).extracting("spec").extracting("replicas").isEqualTo(1);
             assertThat(h).extracting("spec").extracting("template").extracting("spec").isInstanceOfSatisfying(PodSpec.class,
                     podSpec -> {
                         assertThat(podSpec.getContainers()).hasOnlyOneElementSatisfying(container -> {
--- a/integration-tests/kubernetes/standard/src/test/java/io/quarkus/it/kubernetes/KubernetesWithApplicationPropertiesTest.java
+++ b/integration-tests/kubernetes/standard/src/test/java/io/quarkus/it/kubernetes/KubernetesWithApplicationPropertiesTest.java
@@ -55,6 +55,7 @@ public class KubernetesWithApplicationPropertiesTest {
                 });
 
                 assertThat(d.getSpec()).satisfies(deploymentSpec -> {
+                    assertThat(deploymentSpec.getReplicas()).isEqualTo(3);
                     assertThat(deploymentSpec.getTemplate()).satisfies(t -> {
                         assertThat(t.getSpec()).satisfies(podSpec -> {
                             assertThat(podSpec.getContainers()).hasOnlyOneElementSatisfying(container -> {
--- a/integration-tests/kubernetes/standard/src/test/java/io/quarkus/it/kubernetes/KubernetesWithDefaultsTest.java
+++ b/integration-tests/kubernetes/standard/src/test/java/io/quarkus/it/kubernetes/KubernetesWithDefaultsTest.java
@@ -43,6 +43,7 @@ public class KubernetesWithDefaultsTest {
             assertThat(i).isInstanceOfSatisfying(Deployment.class, d -> {
 
                 assertThat(d.getSpec()).satisfies(deploymentSpec -> {
+                    assertThat(deploymentSpec.getReplicas()).isEqualTo(1);
                     assertThat(deploymentSpec.getTemplate()).satisfies(t -> {
                         assertThat(t.getSpec()).satisfies(podSpec -> {
                             assertThat(podSpec.getContainers()).hasOnlyOneElementSatisfying(container -> {
--- a/integration-tests/kubernetes/standard/src/test/java/io/quarkus/it/kubernetes/OpenshiftWithApplicationPropertiesTest.java
+++ b/integration-tests/kubernetes/standard/src/test/java/io/quarkus/it/kubernetes/OpenshiftWithApplicationPropertiesTest.java
@@ -46,6 +46,7 @@ public class OpenshiftWithApplicationPropertiesTest {
                 assertThat(m.getName()).isEqualTo("test-it");
                 assertThat(m.getLabels()).contains(entry("foo", "bar"));
             });
+            assertThat(h).extracting("spec").extracting("replicas").isEqualTo(3);
             assertThat(h).extracting("spec").extracting("template").extracting("spec").isInstanceOfSatisfying(PodSpec.class,
                     podSpec -> {
                         assertThat(podSpec.getContainers()).hasOnlyOneElementSatisfying(container -> {
--- a/integration-tests/kubernetes/standard/src/test/resources/kubernetes-with-application.properties
+++ b/integration-tests/kubernetes/standard/src/test/resources/kubernetes-with-application.properties
@@ -9,3 +9,4 @@ quarkus.kubernetes.expose=true
 quarkus.kubernetes.host=example.com
 quarkus.kubernetes.service-type=NodePort
 quarkus.kubernetes.image-pull-policy=IfNotPresent
+quarkus.kubernetes.replicas=3
--- a/integration-tests/kubernetes/standard/src/test/resources/openshift-with-application.properties
+++ b/integration-tests/kubernetes/standard/src/test/resources/openshift-with-application.properties
@@ -6,3 +6,4 @@ quarkus.openshift.env-vars.my-env-var.value=SOMEVALUE
 quarkus.openshift.group=grp
 quarkus.openshift.expose=true
 quarkus.s2i.registry=quay.io
+quarkus.openshift.replicas=3


extensions/kubernetes/vanilla/deployment/src/main/java/io/quarkus/kubernetes/deployment/KubernetesConfigUtil.java
```diff
@@ -63,6 +63,7 @@
             p.getPartOf().ifPresent(g -> quarkusPrefixed.put(DEKORATE_PREFIX + p.getConfigName() + ".part-of", g));
             p.getName().ifPresent(n -> quarkusPrefixed.put(DEKORATE_PREFIX + p.getConfigName() + ".name", n));
             p.getVersion().ifPresent(v -> quarkusPrefixed.put(DEKORATE_PREFIX + p.getConfigName() + ".version", v));
+            p.getReplicas().ifPresent(r -> quarkusPrefixed.put(DEKORATE_PREFIX + p.getConfigName() + ".replicas", r));
```
geoand, line 66: This is the "legacy" way of doing things and we would really like to avoid it

geoand (changes requested):
Very good first attempt, but I would rather we use the Dekorator approach that I outlined in my comment on the issue.

geoand:
Also keep in kind that after the comment has been addressed, we would also like the commits to be squashed into a single one.

Thanks

geoand:
@iocanel do Knative resource have replicas? If not, then the replicas should be removed from `PlarformConfiguration` and added only to `KubernetesConfig` and `OpenshiftConfig`.

geoand:
@iocanel as mentioned (in chat) by @cemnura the `ApplyReplicasDecorator` doesn't seem to work. 
I tried it as well and can confirm the problem. Sounds like a bug in Dekorate? If so, I propose we merge this PR and then you can come back and apply the proper fix when you have a new Dekorate release.

cemnura:
I was able to successfully  read the **quarkus.kubernetes.replicas** configuration and added the io.dekorate.kubernetes.decorator.ApplyReplicasDecorator as the following

```java
            config.getReplicas().ifPresent(i -> session.resources().decorate(target, new ApplyReplicasDecorator(name, i)));
```

However,  the ApplyReplicasDecorator value is overrided during dekorate generate phase. By default the dekorate adds a ApplyReplicasDecorator with the value 1. I observed this from debugging.

Would it be useful to show that in debugging? If so how would you recommend be to show in debug?

geoand:
> I was able to successfully read the **quarkus.kubernetes.replicas** configuration and added the io.dekorate.kubernetes.decorator.ApplyReplicasDecorator as the following
> 
> ```java
>             config.getReplicas().ifPresent(i -> session.resources().decorate(target, new ApplyReplicasDecorator(name, i)));
> ```
> 
> However, the ApplyReplicasDecorator value is overrided during dekorate generate phase.

Yeah, exactly what I saw as well.

> Would it be useful to show that in debugging? If so how would you recommend be to show in debug?

I don't think this would help at all.

My proposal is to use the approach that this PR uses for now, and then @iocanel can update the code when we have a new Dekorate release.

extensions/kubernetes/vanilla/deployment/src/main/java/io/quarkus/kubernetes/deployment/KubernetesConfig.java
```diff
@@ -95,6 +96,12 @@
     @ConfigItem
     Map<String, PortConfig> ports;
 
+    /**
+     * The number of desired pods
+     */
+    @ConfigItem
+    OptionalInt replicas;
```
geoand, line 103: I think we should probably use:

```
    @ConfigItem(defaultValue = "1")
    Integer replicas;
```

It will make the default more visible.

extensions/kubernetes/vanilla/deployment/src/main/java/io/quarkus/kubernetes/deployment/KnativeConfig.java
```diff
@@ -95,6 +96,12 @@
     @ConfigItem
     Map<String, PortConfig> ports;
 
+    /**
+     * The number of desired pods
+     */
+    @ConfigItem
+    OptionalInt replicas;
```
geoand, line 103: I guess we need to drop this since Knative doesn't seem to have replicas.

extensions/kubernetes/vanilla/deployment/src/main/java/io/quarkus/kubernetes/deployment/OpenshiftConfig.java
```diff
@@ -96,6 +97,12 @@
     @ConfigItem
     Map<String, PortConfig> ports;
 
+    /**
+     * The number of desired pods
+     */
+    @ConfigItem
+    OptionalInt replicas;
```
geoand, line 104: Same comment as the `KubernetesConfig`

extensions/kubernetes/vanilla/deployment/src/main/java/io/quarkus/kubernetes/deployment/KubernetesConfig.java
cemnura: Okey sounds like a good idea. I will update the pull request

iocanel:
IMHO, the easiest approach would be to copy/paste the `ApplyReplicasDecorator` in quarkus and override the `after` method so that the quarkus decorator is applied after the dekorate one.

Once, we address the issue of decorators applying default values conflicting with the ones applied via quarkus (in the dekorate side), we will remove the copied class and the code will remain pretty much the same.

geoand:
Sounds like a good idea to me.

extensions/kubernetes/vanilla/deployment/src/main/java/io/quarkus/kubernetes/deployment/KnativeConfig.java
cemnura: How would you recommend this? Should I delete the method from the PlatformConfiguration as well? 

Deleting the method would mean that I wouldn't be able use PlatformConfiguration in KubernetesConfigUtil#toMap

geoand: Yeah, we'll have to remove it from `PlatformConfiguration`.

If we follow the latest suggestion from @iocanel then what you mention won't be a problem since the field won't be used in `toMap`

cemnura:
Sounds like a better fix then legacy method. That way the code would not have to change. Would extending the ApplyReplicasDecorator and override the after method be a good idea?

geoand:
I don't think it would make much of a difference, but I don't see why not :)

cemnura:
Thought it would be simpler. Then when delete it and reimport Dekorate when the issue is resolved.

iocanel:
> Sounds like a better fix then legacy method. That way the code would not have to change. Would extending the `ApplyReplicasDecorator` and override the after method be a good idea?

I like this idea a lot and I think that this `should be the way`. I am not 100% sure that it will work though. So, I'd say try it, but if it doesn't work don't sweat it and please let me know so that it can be supported in the future.

cemnura:
> > Sounds like a better fix then legacy method. That way the code would not have to change. Would extending the `ApplyReplicasDecorator` and override the after method be a good idea?
> 
> I like this idea a lot and I think that this `should be the way`. I am not 100% sure that it will work though. So, I'd say try it, but if it doesn't work don't sweat it and please let me know so that it can be supported in the future.

@iocanel Okey I will give it a try

extensions/kubernetes/vanilla/deployment/src/main/java/io/quarkus/kubernetes/deployment/KnativeConfig.java
cemnura: Okey I will remove that form PlatformConfiguration

geoand:
@cemnura there is also a conflict now due to a PR we merged yesterday.

cemnura:
Hello @geoand @iocanel 

I was able to extend ApplyReplicasDecorator and override after method to enable the quarkus decorator is applied `after` the dekorate one as @iocanel suggested and that allowed the generated manifest to have the value indicated in the `application.properties` file.

However, this time the Openshift manifest configuration failed. I later understand that the `ApplyReplicasDecorator` for kubernetes and openshift where different due to the different kinds of generated manifest Deployment & DeploymentConfig. 

Therefore, when I add the `KubernetesApplyReplicasDecortor` for each targeted environment (kubernetes, openshift) the extended ApplyReplicasDecorator only applied for the Kubernetes target. 

Thus, I created a `OpenshiftApplyReplicasDecorator` to be able to distinguish the decorator that will be applied for desired target.

The two extended `ApplyReplicasDecorator` classes lead to the following 

```java
// Replicas
if (KUBERNETES.equals(target)) {
    config.getReplicas().ifPresent(i -> session.resources().decorate(new KubernetesApplyReplicasDecorator(name, i)));
} else if (OPENSHIFT.equals(target)) {
    config.getReplicas().ifPresent(i -> session.resources().decorate(new OpenshiftApplyReplicasDecorator(i)));
}
```

Is there a common `ApplyReplicasDecorator` for both Kubernetes and Openshift targets?


Also I have a thought;

Would it be beneficial for Kubernetes and Openshift to be different extensions? The code would be similar but the differences can be distinguished and defined separately. Maybe that way the complexity of generating either manifest in the same extension can be avoided.

geoand:
> Is there a common `ApplyReplicasDecorator` for both Kubernetes and Openshift targets?
> 

I don't think there is.

> Also I have a thought;
> 
> Would it be beneficial for Kubernetes and Openshift to be different extensions? The code would be similar but the differences can be distinguished and defined separately. Maybe that way the complexity of generating either manifest in the same extension can be avoided.

Currently we have an openshift extension whose purpose is only to make it easier for users to configure openshift stuff.
But some better code organization would be nice as well.
I'll let @iocanel comment on that.

extensions/kubernetes/vanilla/deployment/src/main/java/io/quarkus/kubernetes/deployment/OpenshiftApplyReplicasDecorator.java
```diff
@@ -0,0 +1,18 @@
...
+    public OpenshiftApplyReplicasDecorator(int replicas) {
+        super(replicas);
+    }
+
+    @Override
+    public Class<? extends Decorator>[] after() {
+        return new Class[] { NamedResourceDecorator.class };
+    }
+
+}
```
cemnura, line 18: @iocanel what do you think about this? Would you have a different suggestion on how to override after() method

extensions/kubernetes/vanilla/deployment/src/main/java/io/quarkus/kubernetes/deployment/KubernetesProcessor.java
```diff
@@ -403,6 +403,11 @@ private void applyBuildItems(Session session,
         configMap.put(OPENSHIFT, openshiftConfig);
         configMap.put(KNATIVE, knativeConfig);
 
+        //Replicas
+        kubernetesConfig.getReplicas()
+                .ifPresent(i -> session.resources().decorate(new KubernetesApplyReplicasDecorator(kubernetesName, i)));
+        openshiftConfig.getReplicas().ifPresent(i -> session.resources().decorate(new OpenshiftApplyReplicasDecorator(i)));
+
```
cemnura, line 410: @geoand I deleted the getReplicas() method from PlatformConfiguration and applied the decator addition for kubernetes and openshift like this. Would you have any suggestions?

cemnura (commented):
> @cemnura there is also a conflict now due to a PR we merged yesterday.

YEs @geoand I realized that and updated my branch

extensions/kubernetes/vanilla/deployment/src/main/java/io/quarkus/kubernetes/deployment/KubernetesProcessor.java
geoand: I looks good to me, :+1:

extensions/kubernetes/vanilla/deployment/src/main/java/io/quarkus/kubernetes/deployment/KubernetesApplyReplicasDecorator.java
```diff
@@ -0,0 +1,20 @@
+package io.quarkus.kubernetes.deployment;
+
+import io.dekorate.kubernetes.decorator.*;
```
geoand, line 3: Please don't use "star" imports in Quarkus :)

geoand: Other than this, I don't have any more objections and we should be good to go.

cemnura: Okey will fix this.

extensions/kubernetes/vanilla/deployment/src/main/java/io/quarkus/kubernetes/deployment/KubernetesProcessor.java
cemnura: Glad to hear @geoand  :)

extensions/kubernetes/vanilla/deployment/src/main/java/io/quarkus/kubernetes/deployment/KubernetesApplyReplicasDecorator.java
cemnura: Removed "star" imports

extensions/kubernetes/vanilla/deployment/src/main/java/io/quarkus/kubernetes/deployment/KubernetesConfig.java
geoand: Oh sorry, I missed this. Any reason to not use `Integer` instead of `OptionalInt`?

geoand: Once this is addressed, we can merge

geoand (approved):
Thanks a lot for this!

geoand:
@iocanel mind taking a look as well?

cemnura (commented):
> Thanks a lot for this!

I thank you for you help and reviews :)

extensions/kubernetes/vanilla/deployment/src/main/java/io/quarkus/kubernetes/deployment/KubernetesConfig.java
geoand: Yeah, it's never empty now, if the user doesn't specify a value `1` will be used by Quarkus so there is no need to be defensive.

extensions/kubernetes/vanilla/deployment/src/main/java/io/quarkus/kubernetes/deployment/KubernetesApplyReplicasDecorator.java
```diff
@@ -0,0 +1,22 @@
...
+        super(replicas);
+    }
+
+    public KubernetesApplyReplicasDecorator(String deploymentName, int replicas) {
+        super(deploymentName, replicas);
+    }
+
+    @Override
+    public Class<? extends Decorator>[] after() {
+        return new Class[] { NamedResourceDecorator.class };
```
iocanel, line 19: Just to be on the safe side of things, I would explicitly express that I want to it to be called after the ApplyReplicasDecorator. Why?

1. It justifies the purpose of the class
2. It makes it easy to understand which dekorator is applied first. (e.g. atm I can't understand which will be applied first without checking the sources of AppliedReplicasDecorator).

extensions/kubernetes/vanilla/deployment/src/main/java/io/quarkus/kubernetes/deployment/OpenshiftApplyReplicasDecorator.java
iocanel: I would prefer if it also included `ApplyReplicasDecorator` (same reasoning as the kubernetes variant).

iocanel (changes requested):
Great work, I would just like a tiny change, so that its more clear, what the decorator does and when its going to be applied!

extensions/kubernetes/vanilla/deployment/src/main/java/io/quarkus/kubernetes/deployment/KubernetesApplyReplicasDecorator.java
cemnura: Hi @iocanel this was my initial intension and I had implemented it but failed. I now changed it as your said and it seems to be working. I think I had a package problem.

My intension was to enable this decorator to work after all NamedResourceDecorators. However, the method you suggested is much more understandable.

cemnura: I change the after call to ApplyReplicasDecorator

iocanel (approved):
LGTM

geoand:
Thanks a lot for this!

PR #8316, opened 2020-04-01
Override the host from X-Forwarded-Host

### LINKED ISSUE #4922: Handle forwarded headers

stuartwdouglas:
Fixes #4922

### DIFF
--- a/extensions/vertx-http/deployment/src/test/java/io/quarkus/vertx/http/ForwardedHandlerInitializer.java
+++ b/extensions/vertx-http/deployment/src/test/java/io/quarkus/vertx/http/ForwardedHandlerInitializer.java
@@ -3,6 +3,7 @@ package io.quarkus.vertx.http;
 import javax.enterprise.context.ApplicationScoped;
 import javax.enterprise.event.Observes;
 
+import io.vertx.core.http.HttpHeaders;
 import io.vertx.ext.web.Router;
 
 @ApplicationScoped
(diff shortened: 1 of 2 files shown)


gastaldi (approved):
LGTM

geoand (approved):

gsmet:
@stuartwdouglas any chance we could have someone working on test coverage for that feature in general? Maybe someone from the Vert.x team who is familiar with all that could help pointing to what we should test?

astanik:
@stuartwdouglas and @gsmet I wrote a test, because I ran into the issue that the x-Forwarded-For Header is not handled correctly. You can find a very simple test in my repository: https://github.com/astanik/X-Forwarded-Host-Issue

Can you please take a look at my repository and tell me if I've made a mistake?

Issue #8297, opened 2020-03-31
Unable to copy .env file in dev mode


Legion2:
**Describe the bug**
The `.env` file for environment variables at project root can't be used on Windows 10.
When creating this file and running `mvnw quarkus:dev` it produces a warning in the log and the environment variables are not used by quarkus.
**Expected behavior**
I can define environment variables in `.env` file and use it in `application.properties` like this:
`greeting.message = ${MESSAGE}`

**Actual behavior**
A warning and exception is printed to the console:
```
Listening for transport dt_socket at address: 5005
__  ____  __  _____   ___  __ ____  ______ 
 --/ __ \/ / / / _ | / _ \/ //_/ / / / __/
 -/ /_/ / /_/ / __ |/ , _/ ,< / /_/ /\ \
--\___\_\____/_/ |_/_/|_/_/|_|\____/___/
2020-03-31 12:12:54,098 WARN  [io.qua.dev.DevModeMain] (main) Unable to copy .env file
2020-03-31 12:12:57,080 ERROR [io.qua.application] (main) Failed to start application: java.util.NoSuchElementException: Property MESSAGE not found
        at io.smallrye.config.SmallRyeConfig.propertyNotFound(SmallRyeConfig.java:209)
        at io.smallrye.config.SmallRyeConfig.getValue(SmallRyeConfig.java:96)
        at io.smallrye.config.SmallRyeConfig.getValue(SmallRyeConfig.java:80)
        at io.quarkus.runtime.configuration.ConfigExpander.accept(ConfigExpander.java:55)
        at io.quarkus.runtime.configuration.ConfigExpander.accept(ConfigExpander.java:15)
        at org.wildfly.common.expression.ExpressionNode.emit(ExpressionNode.java:42)
        at org.wildfly.common.expression.Expression.evaluateException(Expression.java:75)
        at org.wildfly.common.expression.Expression.evaluate(Expression.java:89)
        at io.quarkus.runtime.configuration.ExpandingConfigSource.expandValue(ExpandingConfigSource.java:101)
        at io.quarkus.runtime.configuration.ExpandingConfigSource.expand(ExpandingConfigSource.java:69)   
        at io.quarkus.runtime.configuration.ExpandingConfigSource.getValue(ExpandingConfigSource.java:47) 
        at io.quarkus.runtime.configuration.DeploymentProfileConfigSource.getValue(DeploymentProfileConfigSource.java:80)
[...]

PR #8319, opened 2020-04-01
Include the original exception that caused .env file copy to fail

jaikiran:
Relates to https://github.com/quarkusio/quarkus/issues/8297

I initially thought of having an additional `logger.debug` which would include the exception that caused this to fail. But thinking a bit more, I think this warn message itself should include the original exception since a failure to copy an application artifact (which this `.env` is) can cause issues in the application (like shown in that linked issue), so IMO, the root cause should be available easily instead of just the warn message.


geoand (approved):
Makes sense, thanks!

PR #8320, opened 2020-04-01
Remove Docker bridge mode for jpa-mssql and jpa-mariadb tests

rsvoboda:
Remove Docker bridge mode for jpa-mssql and jpa-mariadb tests

Bridge mode is not available on non-linux systems

jpa-mysql and jpa-postgresql have it already solved
jpa-mariadb uses the same approach as jpa-mysql
jpa-mssql has log based check

Can be check via:
```
cd integration-tests
mvn clean install -fae -pl jpa-mariadb,jpa-mssql,jpa-mysql,jpa-postgresql -Ddocker \
  -Dtest-mariadb -Dtest-mssql -Dtest-mysql -Dtest-postgresql
```

### DIFF
--- a/integration-tests/jpa-mariadb/pom.xml
+++ b/integration-tests/jpa-mariadb/pom.xml
@@ -178,9 +178,6 @@
                                     <name>${mariadb.image}</name>
                                     <alias>quarkus-test-mariadb</alias>
                                     <run>
-                                        <network>
-                                            <mode>bridge</mode>
-                                        </network>
                                         <ports>
                                             <port>3308:3306</port>
                                         </ports>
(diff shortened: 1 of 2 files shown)


integration-tests/jpa-mariadb/pom.xml
```diff
@@ -198,15 +195,10 @@
...
-                                                <mode>direct</mode>
-                                                <ports>
-                                                    <port>3306</port>
-                                                </ports>
-                                            </tcp>
-                                            <!-- Unfortunately booting MariaDB is slow, needs to set a generous timeout: -->
-                                            <time>40000</time>
+                                            <time>20000</time>
+                                            <exec>
+                                                <postStart>mysqladmin ping -h localhost -u hibernate_orm_test -phibernate_orm_test</postStart>
```
geoand, line 200: So we assume that `mysqladmin` will need to be installed to run this test?

rsvoboda: jpa-mariadb uses the same approach as jpa-mysql

https://github.com/quarkusio/quarkus/blob/master/integration-tests/jpa-mysql/pom.xml#L201

geoand: I see we already do that for `jpa-mysql`, so I guess it's OK :)

geoand (approved):

integration-tests/jpa-mariadb/pom.xml
rsvoboda: I don't have mysqladmin on my path and it works for me.
`exec` command runs inside the container. 
looking at https://dmp.fabric8.io/#start-wait

geoand: Oh, that's cool. Thanks for checking that out!

PR #8325, opened 2020-04-01
deprecate buildNative gradle task

### LINKED ISSUE #7104: Deprecate QuarkusNative
As discussed in #7051 and in [Zulip](https://quarkusio.zulipchat.com/#narrow/stream/187030-users/topic/configure.20builderImage.20via.20gradle/near/187596485), the `QuarkusNative` task from our Gradle plugin should be deprecated and kept around for backwards compatibility just like `NativeImageMojo` from the Maven plugin.

The `gradle-tooling` guide will have to be updated, but there's also some work to do in the Gradle plugin since the `QuarkusTestNative` task currently depends on the soon-to-be deprecated `QuarkusNative` task.

glefloch:
this pull request deprecates gradle buildNative task. 

close #7104


gastaldi (approved):
LGTM

gastaldi (changes requested):
You need to update the `io.quarkus.gradle.QuarkusPluginFunctionalTest` too I suppose

gastaldi:
Also make sure the Gradle tests pass in your environment by running `gradle test functionalTest` in the plugin folder

glefloch:
Thank you for the review, I'm checking why the build fails.

machi1990:
Thanks for the contribution @glefloch, can we squash the commits into one?

glefloch:
Yes, for sure, I squashed commits.

devtools/gradle/src/main/java/io/quarkus/gradle/tasks/QuarkusNative.java
```diff
@@ -353,9 +361,9 @@ public void setReportExceptionStackTraces(boolean reportExceptionStackTraces) {
         this.reportExceptionStackTraces = reportExceptionStackTraces;
     }
 
-    @TaskAction
+    @TaskAction()
```
gastaldi, line 364: No need for parenthesis here

devtools/gradle/src/main/java/io/quarkus/gradle/tasks/QuarkusNative.java
```diff
@@ -22,6 +22,14 @@
 import io.quarkus.bootstrap.resolver.AppModelResolver;
 import io.quarkus.bootstrap.resolver.AppModelResolverException;
 
+/**
+ * Legacy task for backwards compatibility reasons. This should not be used in new projects
+ *
+ * This has been replaced by setting quarkus.package.type=native in the configuration.
```
gastaldi, line 28: Can you put this sentence next to the `@deprecated` below?

devtools/gradle/src/main/java/io/quarkus/gradle/tasks/QuarkusNative.java
```diff
@@ -22,6 +22,14 @@
...
+ *
+ * @deprecated
+ */
+@Deprecated
 public class QuarkusNative extends QuarkusTask {
```
gastaldi, line 33: Can we remove this class and declare the `buildNative` task to use `QuarkusBuild` with the extra parameter instead?

gastaldi (changes requested):
Added some comments below. Can you verify them before we merge it?

devtools/gradle/src/main/java/io/quarkus/gradle/tasks/QuarkusNative.java
glefloch: Yes, we could delete this, and declare the `buildNative` as a `DefaultTask`, something like this in `QuarkusPlugin`:

    Task buildNative = tasks.create(BUILD_NATIVE_TASK_NAME, DefaultTask.class);

But this will be a breaking change instead of a deprecation. Is that ok? or should we first release the `@Deprecate` and then this update?

gastaldi: If the behavior is kept the same, it's not exactly a breaking change, since no other module depends on classes of this plugin, I think it should be fine.  We should hopefully remove this task in the future, so it would be nice to show a warning if `buildNative` is invoked directly.

docs/src/main/asciidoc/gradle-tooling.adoc
```diff
@@ -261,7 +261,7 @@ Native executables make Quarkus applications ideal for containers and serverless
 
 Make sure to have `GRAALVM_HOME` configured and pointing to GraalVM version {graalvm-version}.
 
-Create a native executable using: `./gradlew buildNative`.
+Create a native executable using: `./gradlew quarkusBuild -Dquarkus.package.type=native`.
```
gastaldi, line 264: No biggie, but I wonder if we shouldn't promote using `./gradlew build` instead. Seems easier to remember and well-understood by Gradle users I think

glefloch: yes, you right, I'm updating it.

glefloch: Should we also make `buildNative` depends on `build` instead of `quarkusBuild` ?

gastaldi: Hm, I think in this case depending on `quarkusBuild` may be better, but maybe it doesn't matter, WDYT?

glefloch: Well, yes, as we aim to remove `buildNative` task, this may be better to keep it as an "alias" of `quarkusBuild -Dquarkus.package.type=native` and not make it depends on something else which actually depends on `quarkusBuild`

devtools/gradle/src/main/java/io/quarkus/gradle/QuarkusPlugin.java
```diff
@@ -111,9 +115,16 @@ private void registerTasks(Project project) {
                             .extendsFrom(configurations.findByName(JavaPlugin.TEST_RUNTIME_ONLY_CONFIGURATION_NAME));
 
                     Task testNative = tasks.create(TEST_NATIVE_TASK_NAME, QuarkusTestNative.class);
-                    testNative.dependsOn(buildNative);
+                    testNative.dependsOn(quarkusBuild);
                     testNative.setShouldRunAfter(Collections.singletonList(tasks.findByName(JavaPlugin.TEST_TASK_NAME)));
 
+                    project.getGradle().getTaskGraph().whenReady(taskGraph -> {
```
gastaldi, line 121: Isn't this already done in `configureBuildNativeTask(...)`?

glefloch: well, yes, you right, I moved out the logger and used task path instead of task reference.

gastaldi:
It seems that the warning shows only after `quarkusBuild` executes. Is there any way to show that warning BEFORE `quarkusBuild` kicks in?

gastaldi:
Also, don't forget to update the READMEs generated when a project is created: 
https://github.com/quarkusio/quarkus/blob/master/devtools/platform-descriptor-json/src/main/resources/templates/README.gradle.ftl#L27-L31

glefloch:
@gastaldi, I updated `buildNative` task to be `finalizedBy` by `quarkusBuild`. The output looks like: 

    > Task :buildNative
    buildNative task has been deprecated in favor of quarkusBuild -Dquarkus.package.type=native

    > Task :quarkusBuild
    building quarkus runner

    BUILD SUCCESSFUL in 1m 20s

WDYT?

devtools/gradle/src/main/java/io/quarkus/gradle/QuarkusPlugin.java
```diff
@@ -69,9 +72,15 @@ private void registerTasks(Project project) {
 
         Task quarkusBuild = tasks.create(QUARKUS_BUILD_TASK_NAME, QuarkusBuild.class);
         Task quarkusDev = tasks.create(QUARKUS_DEV_TASK_NAME, QuarkusDev.class);
-        Task buildNative = tasks.create(BUILD_NATIVE_TASK_NAME, QuarkusNative.class);
         Task quarkusTestConfig = tasks.create(QUARKUS_TEST_CONFIG_TASK_NAME, QuarkusTestConfig.class);
 
+        Task buildNative = tasks.create(BUILD_NATIVE_TASK_NAME, DefaultTask.class);
+        buildNative.finalizedBy(quarkusBuild);
+        buildNative.doFirst(t -> project.getLogger()
+                .warn("buildNative task has been deprecated in favor of quarkusBuild -Dquarkus.package.type=native"));
```
gastaldi, line 80:
```suggestion
                .warn("The 'buildNative' task has been deprecated in favor of 'build -Dquarkus.package.type=native'"));
```

glefloch: Yes, this is better, I updated the warning

docs/src/main/asciidoc/gradle-tooling.adoc
```diff
@@ -300,35 +289,6 @@ The list of the available Docker images can be found on https://quay.io/reposito
...
-[source,kotlin,subs=attributes+]
-----
-tasks {
-    named<QuarkusNative>("buildNative") {
-        setEnableHttpUrlHandler(true)
-    }
-}
-----
-
-
```
gastaldi, line 323: @geoand this PR removes the possibility of doing this, WDYT?

geoand: Had we deprecated it before removing it?

geoand: If not, I think we should first deprecate it (while also removing from the docs)

gastaldi: I agree with @geoand 

@glefloch do you think it would be feasible if we had a PR deprecating this feature first?

glefloch: yes for sure, I rollback the code and keep the docs as this?

gastaldi: I was thinking of keeping this PR as is and having another PR updating the docs saying that this feature is deprecated and will be removed in the next release

glefloch: Oh ok. I misunderstood. I updated the code. I will open an other pull request saying the task is deprecated.

gastaldi (approved):

aloubyansky (approved):

PR #8327, opened 2020-04-01
Add support for @InjectMock

geoand:
This PR builds on #8175 and allows users to easily 
combine Mockito with Quarkus mocks without having 
to use anything more than the `@MockBean` annotation

### DIFF
--- a/test-framework/junit5-mockito/src/main/java/io/quarkus/test/junit/mockito/internal/CreateMockitoMocksCallback.java
+++ b/test-framework/junit5-mockito/src/main/java/io/quarkus/test/junit/mockito/internal/CreateMockitoMocksCallback.java
@@ -1,0 +1,70 @@
+package io.quarkus.test.junit.mockito.internal;
+
+import java.lang.annotation.Annotation;
+import java.lang.reflect.Field;
+import java.util.ArrayList;
+import java.util.List;
+
+import javax.inject.Qualifier;
+
+import org.mockito.Mockito;
+
+import io.quarkus.arc.Arc;
+import io.quarkus.arc.InstanceHandle;
+import io.quarkus.test.junit.callback.QuarkusTestBeforeAllCallback;
+import io.quarkus.test.junit.mockito.InjectMock;
+
+public class CreateMockitoMocksCallback implements QuarkusTestBeforeAllCallback {
+
+    @Override
+    public void beforeAll(Object testInstance) {
+        Class<?> current = testInstance.getClass();
+        while (current.getSuperclass() != null) {
+            for (Field field : current.getDeclaredFields()) {
+                InjectMock injectMockAnnotation = field.getAnnotation(InjectMock.class);
+                if (injectMockAnnotation != null) {
+                    Object beanInstance = getBeanInstance(testInstance, field);
+                    Object mock = createMockAndSetTestField(testInstance, field, beanInstance);
+                    MockitoMocksTracker.track(testInstance, mock, beanInstance);
+                }
+            }
+            current = current.getSuperclass();
+        }
+    }
+
+    private Object createMockAndSetTestField(Object testInstance, Field field, Object beanInstance) {
+        Object mock = Mockito.mock(beanInstance.getClass());
+        field.setAccessible(true);
+        try {
+            field.set(testInstance, mock);
+        } catch (IllegalAccessException e) {
+            throw new RuntimeException(e);
+        }
+        return mock;
+    }
+
+    private Object getBeanInstance(Object testInstance, Field field) {
+        Class<?> fieldClass = field.getType();
+        InstanceHandle<?> instance = Arc.container().instance(fieldClass, getQualifiers(field));
+        if (!instance.isAvailable()) {
+            throw new IllegalStateException("Invalid use of @InjectMock - could not determine bean of type: "
+                    + fieldClass + ". Offending field is " + field.getName() + " of test class "
+                    + testInstance.getClass());
+        }
+        return instance.get();
+    }
+
+    private Annotation[] getQualifiers(Field fieldToMock) {
+        List<Annotation> qualifiers = new ArrayList<>();
+        Annotation[] fieldAnnotations = fieldToMock.getDeclaredAnnotations();
+        for (Annotation fieldAnnotation : fieldAnnotations) {
+            for (Annotation annotationOfFieldAnnotation : fieldAnnotation.annotationType().getAnnotations()) {
+                if (annotationOfFieldAnnotation.annotationType().equals(Qualifier.class)) {
+                    qualifiers.add(fieldAnnotation);
+                    break;
+                }
+            }
+        }
+        return qualifiers.toArray(new Annotation[0]);
+    }
+}
--- a/bom/deployment/pom.xml
+++ b/bom/deployment/pom.xml
@@ -720,6 +720,11 @@
             </dependency>
             <dependency>
                 <groupId>io.quarkus</groupId>
+                <artifactId>quarkus-junit5</artifactId>
+                <version>${project.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>io.quarkus</groupId>
                 <artifactId>quarkus-integration-test-class-transformer</artifactId>
                 <version>${project.version}</version>
             </dependency>
--- a/bom/runtime/pom.xml
+++ b/bom/runtime/pom.xml
@@ -904,6 +904,11 @@
             </dependency>
             <dependency>
                 <groupId>io.quarkus</groupId>
+                <artifactId>quarkus-junit5-mockito</artifactId>
+                <version>${project.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>io.quarkus</groupId>
                 <artifactId>quarkus-arquillian</artifactId>
                 <version>${project.version}</version>
             </dependency>
--- a/docs/src/main/asciidoc/getting-started-testing.adoc
+++ b/docs/src/main/asciidoc/getting-started-testing.adoc
@@ -388,13 +388,28 @@ public class MockTestCase {
         Assertions.assertEquals("Bonjour Stuart", mockableBean2.greet("Stuart"));
     }
 
+    @ApplicationScoped
+    public static class MockableBean1 {
+
+        public String greet(String name) {
+            return "Hello " + name;
+        }
+    }
+
+    @ApplicationScoped
+    public static class MockableBean2 {
+
+        public String greet(String name) {
+            return "Hello " + name;
+        }
+    }
+
     public static class BonourGreeter extends MockableBean2 {
         @Override
         public String greet(String name) {
             return "Bonjour " + name;
         }
     }
-
 }
 ----
 <1> As the injected instance is not available here we use `installMockForType`, this mock is used for both test methods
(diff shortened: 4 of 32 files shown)


test-framework/junit5-mockito/src/main/java/io/quarkus/test/junit/mockito/internal/CreateMockitoMocksCallback.java
```diff
@@ -0,0 +1,70 @@
...
+            field.set(testInstance, mock);
+        } catch (IllegalAccessException e) {
+            throw new RuntimeException(e);
+        }
+        return mock;
+    }
+
+    private Object getBeanInstance(Object testInstance, Field field) {
+        Class<?> fieldClass = field.getType();
+        InstanceHandle<?> instance = Arc.container().instance(fieldClass, getQualifiers(field));
```
mkouba, line 48: If you store the `InstanceHandle` you can use it later to destroy a `@Dependent` bean correctly. See `io.quarkus.arc.InstanceHandle.close()`.

integration-tests/mockbean/src/test/java/io/quarkus/it/mockbean/GreetingResourceTest.java
```diff
@@ -0,0 +1,44 @@
...
+import org.mockito.Mockito;
+
+import io.quarkus.test.junit.QuarkusTest;
+import io.quarkus.test.junit.mockito.MockBean;
+
+@QuarkusTest
+class GreetingResourceTest {
+
+    @MockBean
+    MessageService messageService;
```
mkouba, line 16: Since these are not CDI injection points it could happen that ArC will consider the resolved beans removable. Maybe we should handle this use case and mark them unremovable?

integration-tests/mockbean/src/test/java/io/quarkus/it/mockbean/GreetingResourceTest.java
```diff
@@ -0,0 +1,44 @@
...
+
+    @MockBean
+    SuffixService suffixService;
+
+    @Test
+    public void testGreet() {
+        Mockito.when(messageService.getMessage()).thenReturn("hi");
```
mkouba, line 23: So I wonder if we could just get rid of the injection part and provide an API like this:
```java
@Test
public void testGreet() {
   Mockito.when(MockBean.of(MessageService.class).getMessage()).thenReturn("hi");
   ...
}
```
... where `MockBean.of()` would basically do `QuarkusMock.installMockForType(Mockito.mock(MessageService.class), MessageService.class)` and return the mock instance (and possibly cache the instance or something like that).

mkouba: Hm, maybe it's not a good idea. The injection points can be reused and allow us to mark the beans unremovable. Another idea - if we do `@Inject @MockBean` the client proxies will be injected first and we could wrap it afterwards. `@MockBean` is ignored by ArC but allows `CreateMockitoMocksCallback` to wrap the stuff. The advantage of this approach would be (1) no need to mark the beans unremovable, (2) no need to hold the instance handles and destroy them correctly.

geoand: Let me think about it and get back to you tomorrow :)

mkouba: Feel free to ping me to discuss the API ;-)

manovotn:
I didn't see Stuart's PR until today, so I am trying to catch up here...
So the idea is that we always require ArC to supply the contextual instance anyway and then pass that instance to Mockito to let them create their mock on top of that? Isn't this going to interfere with beans that have interceptors? Since we already have a subclass there and the proxy now returns the new mock as delegate.

Also, all of this is still has to be on top of normal scoped beans, right?

integration-tests/mockbean/src/test/java/io/quarkus/it/mockbean/GreetingResourceTest.java
stuartwdouglas: If a bean is removable then then mocking it would have no effect anyway.

test-framework/junit5-mockito/src/main/java/io/quarkus/test/junit/mockito/internal/CreateMockitoMocksCallback.java
stuartwdouglas: We can't really mock @Dependent beans anyway.

stuartwdouglas (approved):
LGTM

integration-tests/mockbean/src/test/java/io/quarkus/it/mockbean/GreetingResourceTest.java
geoand: @mkouba if we do `@Inject @MockBean` in the test, then how would the user control the mocks? 
My idea about `@MockBean` is that it should just be a Mockito mock that users already know how to handle without having to do anything else.  If we add `@Inject` in the mix, won't that complicate the mock usage on the user side?

test-framework/junit5-mockito/src/main/java/io/quarkus/test/junit/mockito/internal/CreateMockitoMocksCallback.java
geoand: Hm... Here we aren't storing the references, just using them to check and then obtain the bean. So we don't need to close them anywhere, no?

integration-tests/mockbean/src/test/java/io/quarkus/it/mockbean/GreetingResourceTest.java
geoand: Like @stuartwdouglas said, if the bean has been removed, mocking won't change anything, so how would the user see the results of the their test?
I don't think this scenario is actually possible. 
It's likely I think that if normally the bean would be removed in the prod app, the user would be injecting some other bean into the test to check the effect of the mock, thus making the mock unremovable.

geoand:
@manovotn yeah, the idea is that a normal scoped bean's proxy can be used to set a mock. This will only work with normal scoped beans indeed.
Dunno about interceptors TBH.

integration-tests/mockbean/src/test/java/io/quarkus/it/mockbean/GreetingResourceTest.java
mkouba: > If a bean is removable then then mocking it would have no effect anyway.

If a bean is removable then a null would be injected or an exception would be thrown - depends on the `CreateMockitoMocksCallback` impl. My point is that right now you can test a bean in a `@QuarkusTest` even if it would be normally removed in production because test class is considered a bean.

> if we do @Inject @MockBean in the test, then how would the user control the mocks?

Like I said - the client proxies could be injected by ArC and then wrapped in the `CreateMockitoMocksCallback` and the wrapped mocks could be set back to the fields. So you would do `Mockito.when(messageService.getMessage()).thenReturn("hi")` like you do now.

mkouba:
I added a comment about interceptors to [Stuart's PR](https://github.com/quarkusio/quarkus/pull/8175#issuecomment-604956731) - interceptors are implemented on the subclass and thus will not be invoked. 

> we always require ArC to supply the **contextual instance** anyway

@manovotn Not really. The mocks work with **contextual references** -> client proxies which delegate all calls to the mock instance instead of an intercepted subclass instance. In other words, the intercepted subclass instance may not be instantiated at all (unless the bean declares an observer or producer).

test-framework/junit5-mockito/src/main/java/io/quarkus/test/junit/mockito/internal/CreateMockitoMocksCallback.java
mkouba: > We can't really mock @dependent beans anyway.

Ah, that's correct. For `@Dependent` it would fail later when we try to install the mock by means of `QuarkusMock`.

> So we don't need to close them anywhere, no?

It's ok for normal scoped beans and `@Singleton`. Nok for `@Dependent`.

integration-tests/mockbean/src/test/java/io/quarkus/it/mockbean/GreetingResourceTest.java
geoand: @mkouba I agree completely about testing beans injected into a class - they are made unremovable because you are actually testing those beans.
In the mock case however, you aren't testing the injected mocks, you are testing whatever component uses the mocks - so making them unremovable should have no effect.
Am I missing something?

test-framework/junit5-mockito/src/main/java/io/quarkus/test/junit/mockito/internal/CreateMockitoMocksCallback.java
geoand: Ah, now I see what your concern was! So yeah, no `@Dependent` - no problem :)

integration-tests/mockbean/src/test/java/io/quarkus/it/mockbean/GreetingResourceTest.java
mkouba: You're probably right. It's a bit confusing though. We should probably add some docs and explain the intention...

geoand: Yes of course, I haven't added any docs yet but I absolutely plan to do so.

geoand: I added some documentation. WDYT @mkouba ?

mkouba (approved):
Looks good now. The only thing I'm thinking about is whether it would sense to rename `@MockBean` to `@InjectMock` because that's what this annotation is used for, or?

geoand:
@mkouba sure, that's probably a better name. Let me rename everything.

manovotn:
> I added a comment about interceptors to Stuart's PR - interceptors are implemented on the subclass and thus will not be invoked.

I see, thanks for the reference.

> In other words, the intercepted subclass instance may not be instantiated at all

Yeah, I figured that later yesterday when reading the code more closely. However, we still do validation so users need to have all deps of that given bean available even if they then never instantiate it.

manovotn (approved):
I think we're about to see quite some users trying unexpected/unsupported things with this.
But hey, it's still better then what we had before, so LGTM!

geoand:
> I think we're about to see quite some users trying unexpected/unsupported things with this.
> But hey, it's still better then what we had before, so LGTM!

I have no doubt that will be the case, however with `@InjectMock` we'll be able to work around some and/or provide good error messages to guide users, as opposed to the low-level mode where users have to figure out things on their own.

lordofthejars:
Thanks @geoand for this. I tested and it is awesome and even it is test aware, so it is not like previous `@Mock` support that affects all tests.

geoand:
Glad you like it 😁

lordofthejars:
What I love is that it is isolated by test class. It was something really basic that was missing.

PR #8328, opened 2020-04-01
Upgrade to Hibernate Search 6.0.0.Beta6

yrodiere:
The last commit is not really necessary, but I made this change as I was working on adding more unit tests related to configuration properties... and I then gave up on that.

### DIFF
--- a/docs/src/main/asciidoc/hibernate-search-elasticsearch.adoc
+++ b/docs/src/main/asciidoc/hibernate-search-elasticsearch.adoc
@@ -537,8 +537,8 @@ quarkus.hibernate-orm.sql-load-script=import.sql <4>
 
 quarkus.hibernate-search.elasticsearch.version=7 <5>
 quarkus.hibernate-search.elasticsearch.analysis.configurer=org.acme.hibernate.search.elasticsearch.config.AnalysisConfigurer <6>
-quarkus.hibernate-search.elasticsearch.index-defaults.lifecycle.strategy=drop-and-create <7>
-quarkus.hibernate-search.elasticsearch.index-defaults.lifecycle.required-status=yellow <8>
+quarkus.hibernate-search.schema-management.strategy=drop-and-create <7>
+quarkus.hibernate-search.elasticsearch.index-defaults.schema-management.required-status=yellow <8>
 quarkus.hibernate-search.automatic-indexing.synchronization.strategy=sync <9>
 ----
 <1> We won't use SSL so we disable it to have a more compact native executable.
@@ -550,7 +550,7 @@ It is important because there are significant differences between Elasticsearch 
 Since the mapping is created at build time to reduce startup time, Hibernate Search cannot connect to the cluster to automatically detect the version.
 <6> We point to the custom `AnalysisConfigurer` which defines the configuration of our analyzers and normalizers.
 <7> Obviously, this is not for production: we drop and recreate the index every time we start the application.
-<8> We consider the `yellow` status is sufficient to connect to the Elasticsearch cluster.
+<8> We consider the `yellow` status is sufficient to proceed after an index is created.
 This is for testing purposes with the Elasticsearch Docker container.
 It should not be used in production.
 <9> This means that we wait for the entities to be searchable before considering a write complete.
@@ -601,7 +601,7 @@ Let's use Docker to start one of each:
 
 [source, shell]
 ----
-docker run -it --rm=true --name elasticsearch_quarkus_test -p 9200:9200 -p 9300:9300 -e "discovery.type=single-node" docker.elastic.co/elasticsearch/elasticsearch:7.5.0
+docker run -it --rm=true --name elasticsearch_quarkus_test -p 9200:9200 -p 9300:9300 -e "discovery.type=single-node" docker.elastic.co/elasticsearch/elasticsearch:7.6.1
 ----
 
 [source, shell]
--- a/extensions/hibernate-search-elasticsearch/runtime/src/main/java/io/quarkus/hibernate/search/elasticsearch/runtime/HibernateSearchElasticsearchRuntimeConfig.java
+++ b/extensions/hibernate-search-elasticsearch/runtime/src/main/java/io/quarkus/hibernate/search/elasticsearch/runtime/HibernateSearchElasticsearchRuntimeConfig.java
@@ -479,6 +479,13 @@ public class HibernateSearchElasticsearchRuntimeConfig {
     public static class ElasticsearchIndexIndexingConfig {
         /**
          * The number of indexing queues assigned to each index.
+         * <p>
+         * Higher values will lead to more connections being used in parallel,
+         * which may lead to higher indexing throughput,
+         * but incurs a risk of overloading Elasticsearch,
+         * i.e. of overflowing its HTTP request buffers and tripping
+         * <a href="https://www.elastic.co/guide/en/elasticsearch/reference/7.6/circuit-breaker.html">circuit breakers</a>,
+         * leading to Elasticsearch giving up on some request and resulting in indexing failures.
          */
         // We can't set an actual default value here: see comment on this class.
         @ConfigItem(defaultValueDocumentation = "10")
@@ -486,6 +493,11 @@ public class HibernateSearchElasticsearchRuntimeConfig {
 
         /**
          * The size of indexing queues.
+         * <p>
+         * Lower values may lead to lower memory usage, especially if there are many queues,
+         * but values that are too low will reduce the likeliness of reaching the max bulk size
+         * and increase the likeliness of application threads blocking because the queue is full,
+         * which may lead to lower indexing throughput.
          */
         // We can't set an actual default value here: see comment on this class.
         @ConfigItem(defaultValueDocumentation = "1000")
@@ -493,6 +505,16 @@ public class HibernateSearchElasticsearchRuntimeConfig {
 
         /**
          * The maximum size of bulk requests created when processing indexing queues.
+         * <p>
+         * Higher values will lead to more documents being sent in each HTTP request sent to Elasticsearch,
+         * which may lead to higher indexing throughput,
+         * but incurs a risk of overloading Elasticsearch,
+         * i.e. of overflowing its HTTP request buffers and tripping
+         * <a href="https://www.elastic.co/guide/en/elasticsearch/reference/7.6/circuit-breaker.html">circuit breakers</a>,
+         * leading to Elasticsearch giving up on some request and resulting in indexing failures.
+         * <p>
+         * Note that raising this number above the queue size has no effect,
+         * as bulks cannot include more requests than are contained in the queue.
          */
         // We can't set an actual default value here: see comment on this class.
         @ConfigItem(defaultValueDocumentation = "100")
--- a/extensions/hibernate-search-elasticsearch/runtime/src/main/java/io/quarkus/hibernate/search/elasticsearch/runtime/HibernateSearchElasticsearchRuntimeConfig.java
+++ b/extensions/hibernate-search-elasticsearch/runtime/src/main/java/io/quarkus/hibernate/search/elasticsearch/runtime/HibernateSearchElasticsearchRuntimeConfig.java
@@ -180,6 +180,12 @@ public class HibernateSearchElasticsearchRuntimeConfig {
          */
         @ConfigItem
         ElasticsearchIndexSchemaManagementConfig schemaManagement;
+
+        /**
+         * Configuration for the indexing process that creates, updates and deletes documents.
+         */
+        @ConfigItem
+        ElasticsearchIndexIndexingConfig indexing;
     }
 
     @ConfigGroup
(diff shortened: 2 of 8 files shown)


extensions/hibernate-search-elasticsearch/runtime/src/main/java/io/quarkus/hibernate/search/elasticsearch/runtime/HibernateSearchElasticsearchRuntimeConfig.java
```diff
@@ -118,6 +119,12 @@
         @ConfigItem
         DiscoveryConfig discovery;
 
+        /**
+         * Configuration for the thread pool assigned to the backend.
```
Sanne, line 123: Could you add a short paragraph to describe when one might want to tune this?

extensions/hibernate-search-elasticsearch/runtime/src/main/java/io/quarkus/hibernate/search/elasticsearch/runtime/HibernateSearchElasticsearchRuntimeConfig.java
```diff
@@ -455,4 +461,30 @@ public String getHibernateSearchString() {
...
+
+    // We can't set actual default values in this section,
+    // otherwise "quarkus.hibernate-search.elasticsearch.index-defaults" will be ignored.
+    @ConfigGroup
+    public static class ElasticsearchIndexIndexingConfig {
+        /**
+         * The number of indexing queues assigned to each index.
+         */
+        // We can't set an actual default value here: see comment on this class.
+        @ConfigItem(defaultValueDocumentation = "10")
```
Sanne, line 473: I'm not entirely sure what a reasonable default would be here - so bear with me as I only know the high level picture.

I wonder if this is a reasonable value for an high-performance index. But maybe we should default to small and lean for a Quarkus default?

Sanne (commented):
Looks good, added a couple minor questions. I'll let @gsmet approve as he was assigned.

extensions/hibernate-search-elasticsearch/runtime/src/main/java/io/quarkus/hibernate/search/elasticsearch/runtime/HibernateSearchElasticsearchRuntimeConfig.java
yrodiere: Just so you know, queues don't have an assigned thread. Threads will only be used when we process the queues and to orchestrate the works, but after the first work is sent, everything happens in the HTTP client threads.

So you can raise this to 256, if you have a thread pool of size 8, you'll still have only 8 threads. Since queue processing is non-blocking, that shouldn't be a problem.

The number of queues defines the level of parallelism, i.e. with 10 queues you will have 10 indexing requests sent in parallel. Crucially, this means we'll use at most 10 connections in parallel for indexing for this index.

The "best" value will likely depend on the size of your cluster and the number of available connections. I haven't been able to pinpoint a "best" value here, as the setting will have a different effect depending on the number of indexes being written to in parallel, the number of available connections, the topology of your Elasticsearch cluster, ...

Bottom line: I don't know what's best for the user either. Which is why I expose this configuration option: so that they can set it themselves. I can change the default to 5, but that'd be as much guesswork as it is now.

Maybe after some more performance testing we'll get a better idea of what the default should be.

extensions/hibernate-search-elasticsearch/runtime/src/main/java/io/quarkus/hibernate/search/elasticsearch/runtime/HibernateSearchElasticsearchRuntimeConfig.java
yrodiere: Honestly, for the Elasticsearch backend, I'm not sure. Everything happening in these threads is non-blocking, except maybe some logging than can happen on timeouts/errors. So, theoretically, having more threads than the number of available cores is pointless.

I suppose one could want to set it to less than the number of cores, just to have less threads around?

Like all these performance settings, they are mostly here so that users are free to experiment. I don't have the answer to all performance problems, unfortunately.

yrodiere:
I added some more documentation to the configuration properties. It's literally copy-pasted from the reference documentation, but I guess that's the way to go.

extensions/hibernate-search-elasticsearch/runtime/src/main/java/io/quarkus/hibernate/search/elasticsearch/runtime/HibernateSearchElasticsearchRuntimeConfig.java
Sanne: In that case it might be wise to not expose anything?  Remember in Quarkus we're trying to minimise the configuration needs.

I would suggest to defer exposing this to when we have a better understanding for it. For example, I suppose it's possible such an exploration concludes there's only need for some "profile" enums: "light" (default) vs "throughput" (matching num cores).

extensions/hibernate-search-elasticsearch/runtime/src/main/java/io/quarkus/hibernate/search/elasticsearch/runtime/HibernateSearchElasticsearchRuntimeConfig.java
Sanne: Thanks for the explanation.  Yes 10 might be a good starting point then but it feels like a random number - since you mentioned they are non-blocking perhaps the best default is 1? One connection should be able to make a quite efficient use of the available network resources - similar systems will use one by default, and maybe 2 or 3 when they really need to push for the maximum.

This also makes me wonder why we expose this. Would you agree that what matters for end users is to understand that this is about the number of connections? Maybe it should be named and documented like that.

Sanne: (again: when in doubt or when you believe it won't matter much, make a choice and don't expose it)

extensions/hibernate-search-elasticsearch/runtime/src/main/java/io/quarkus/hibernate/search/elasticsearch/runtime/HibernateSearchElasticsearchRuntimeConfig.java
yrodiere: I came up with a reason to use this setting a few minutes after I wrote this comment, and I documented it in a following commit: you may want to set this to a low number if your machine has many cores and you only have one index with one queue, for example.

While I'm all for this approach of "not exposing until we're sure" for feature-related settings, I'm not sure it makes sense when it comes to performance settings which will likely be different for every machine, and for which we don't have a silver bullet. In you example, I'd have a really hard time to find the number of core to assign to the "light" profile.

extensions/hibernate-search-elasticsearch/runtime/src/main/java/io/quarkus/hibernate/search/elasticsearch/runtime/HibernateSearchElasticsearchRuntimeConfig.java
```diff
@@ -399,18 +419,45 @@ public String getHibernateSearchString() {
...
+         * Note that number is <em>per backend</em>, not per index.
+         * Adding more indexes will not add more threads.
+         * <p>
+         * As all operations happening in this thread-pool are non-blocking,
+         * raising its size above the number of processor cores available to the JVM will not bring noticeable performance
+         * benefit.
+         * The only reason to alter this setting would be to reduce the number of threads;
+         * for example, in an application with a single index with a single indexing queue,
+         * running on a machine with 64 processor cores,
+         * you will likely want to bring the number of threads to a more reasonable number like 1 or 4.
```
Sanne, line 448: s/you will likely want/you might want/

Sanne: s/to a more reasonable number like 1 or 4/to a lower number/

extensions/hibernate-search-elasticsearch/runtime/src/main/java/io/quarkus/hibernate/search/elasticsearch/runtime/HibernateSearchElasticsearchRuntimeConfig.java
Sanne: yes I'm skeptical about applying such strategy to performance levers as well - but we're asked to try hard, so here I am having you do this exercise of defending your flags :)

I suppose one thing we could try to improve on is ergonomic strategies: rather than exposing such options, ideally we'd study our own statistics at runtime and automate what you'd do as a smart human who knows the project well. Essentially: ask yourself which stats would you need to know to make this perform better? Then what would you do? And automate that - sometimes this still needs some configuration flags, but they are more likely to look like "profiles" and strategy names, so that people can at least express what they are optimising for.

I'm sure that won't work with all options, but it makes sense when there's a possibility of automation to avoid exposing options.

(That being sad - beyond the exercise - I'm fine to keep this configuration property, thanks for your explanations)

extensions/hibernate-search-elasticsearch/runtime/src/main/java/io/quarkus/hibernate/search/elasticsearch/runtime/HibernateSearchElasticsearchRuntimeConfig.java
yrodiere: **TL;DR**: we don't have great defaults for the most common setups yet, and we will never have defaults that fit every setup. Hence the newly exposed settings.

> One connection should be able to make a quite efficient use of the available network resources - similar systems will use one by default, and maybe 2 or 3 when they really need to push for the maximum.

I'm not sure, but you have more experience about network performance than I do. If you're certain about that, I will definitely change this setting.

From what I've seen with Elasticsearch... in raw throughput tests, where we just push everything we can to the index, and there is only one Elasticsearch instance, you're absolutely right. Less connections is even better, probably because it means larger bulks and those are better handled by Elasticsearch.

However, when I spin up a complete application with database access, with 2 (local) Elasticsearch nodes, and I start mass indexing, I witnessed the opposite: jumping from 2 queues to 10 queues gave a +50% throughput increase in a small local test (~25000 wikipedia pages) I performed. Granted there's nothing scientific to that test, but it's at least enough to give pause about whether "more connections is pointless".

There are other things that come into play, I suppose; maybe optimizing the global indexing time while allowing high latency from time to time for some documents is not what works best in all situations.

> This also makes me wonder why we expose this.

Because I can't come up with defaults that will work for the most common setup (which is it, by the way?), much less for everyone. At least, not yet.

> when in doubt or when you believe it won't matter much, make a choice and don't expose it

I don't believe it won't matter, I just believe I don't have a way to come up with the best set of settings for everyone, especially with varying environments: local Elasticsearch instance vs remote, single ES instance vs 100+ nodes, 1 index vs 50, ...

I can postpone exposing these settings until we spend a few weeks doing reliable, scientific performance tests, but who knows when that will happen... At least if I expose these settings, users have a way to improve performance while we work on implementing missing features. And who knows, maybe someone tinkering with these settings will give us useful feedback :)

Sanne: That's great that you tested it. Ok then keep it :)

> And who knows, maybe someone tinkering with these settings will give us useful feedback :)

+1 that's a very good point. And a good reason to document some of your suggestions, so that others power users can actually do that.

docs/src/main/asciidoc/hibernate-search-elasticsearch.adoc
```diff
@@ -537,8 +537,8 @@ quarkus.hibernate-orm.sql-load-script=import.sql <4>
 
 quarkus.hibernate-search.elasticsearch.version=7 <5>
 quarkus.hibernate-search.elasticsearch.analysis.configurer=org.acme.hibernate.search.elasticsearch.config.AnalysisConfigurer <6>
-quarkus.hibernate-search.elasticsearch.index-defaults.lifecycle.strategy=drop-and-create <7>
-quarkus.hibernate-search.elasticsearch.index-defaults.lifecycle.required-status=yellow <8>
+quarkus.hibernate-search.schema-management.strategy=drop-and-create <7>
+quarkus.hibernate-search.elasticsearch.index-defaults.schema-management.required-status=yellow <8>
```
gsmet, line 541: Could you write something I would include in the migration guide?

gsmet (approved):

docs/src/main/asciidoc/hibernate-search-elasticsearch.adoc
yrodiere: @gsmet Here you go:

``````
## Hibernate Search + Elasticsearch (Preview)

Configuration properties relative to Elasticsearch index lifecycle management have changed.
See [this section of the reference documentation](https://docs.jboss.org/hibernate/search/6.0/reference/en-US/html_single/#mapper-orm-schema-management) for detailed documentation about schema management.

In short:

* `quarkus.hibernate-search.elasticsearch.index_defaults.lifecycle.strategy` is now `quarkus.hibernate-search.schema-management.strategy` and accepts the same values as before: `none`, `create`, `validate`, `update`, ...
* A new schema management strategy on startup was introduced: `create-or-validate`. It create indexes if they don't exist, validates their schema if they already exist.
* The default schema management strategy on startup is now `create-or-validate`.
*  `quarkus.hibernate-search.elasticsearch.index_defaults.lifecycle.required_status` is now `quarkus.hibernate-search.elasticsearch.index_defaults.schema_management.required_status`.
*  `quarkus.hibernate-search.elasticsearch.index_defaults.lifecycle.required_status_wait_timeout` is now `quarkus.hibernate-search.elasticsearch.index_defaults.schema_management.required_status_wait_timeout`.

Also related: you can now use [`SearchSchemaManager` APIs](https://docs.jboss.org/hibernate/search/6.0/reference/en-US/html_single/#mapper-orm-schema-management-manager) to trigger schema management operations at the time of your choosing (not just on startup).
``````

gsmet:
Thanks!

Issue #4394, opened 2019-10-05
Kotlin based PanacheEntity throws runtime exception


dodalovic:
Described at [stackoverflow](https://stackoverflow.com/questions/58249773/kotlin-based-panacheentity-throws-runtime-exception) , as well as on [zulipchat](https://quarkusio.zulipchat.com/#narrow/stream/187030-users/topic/kotlin-panacheentity-issue/near/177423729)

geoand:
The problem is essentially that the Kotlin compiler when generating an `invokeStatic` bytecode instruction is not using the entity as the target, but the `PanacheEntity`.

It looks something like:

 ```
 13: invokestatic  #xy                 // Method io/quarkus/hibernate/orm/panache/PanacheEntity.find:(Ljava/lang/String;[Ljava/lang/Object;)Lio/quarkus/hibernate/orm/panache/PanacheQuery;
```

whereas similar examples in Java have bytecode like this:

```
5: invokestatic  #xy                  // Method io/quarkus/hibernate/orm/panache/Fruit.find:(Ljava/lang/String;[Ljava/lang/Object;)Lio/quarkus/hibernate/orm/panache/PanacheQuery;
```

dodalovic:
@geoand 
[ordinacija-api.zip](https://github.com/quarkusio/quarkus/files/3694744/ordinacija-api.zip) contains code and `readme.md` in the projet root describing the issue. 

@FroMage @emmanuelbernard FYI

FroMage:
Well that's very bad luck. Does this also happen on static calls outside of the entity class or only in the companion object?

geoand:
It seems like one can't even do `Fruit.findAll` (where `Fruit` extends `PanacheEntity`) from Kotlin code... I wonder if that is by design or not...

FroMage:
You mean it doesn't compile (different Kotlin syntax for static calls), or it doesn't run (due to invalid bytecode being produced by the Kotlin compiler)?

geoand:
Doesn't compile :(

FroMage:
Hah. Well then we're screwed. Probably an alternative solution/API has to be made especially for Kotlin. I don't know the language enough to think of something, though.

geoand:
For now I think that using the repository approach is pretty much the only option for people wanting to stick to pure Kotlin.

dodalovic:
@geoand I'd agree with you. If we go this way - we need to clearly state that in the docs. 
[...]

PR #8333, opened 2020-04-01
support for using panache with kotlin

evanchooly:
I know there's a fair bit here to discuss so I'm viewing this as an initial spike and we can revise and rethink approaches as those more panache-experienced weigh in.

To start, kotlin doesn't have the notion of static.  So the static helper methods on `PanacheEntity` have been moved to `PanacheCompanion`.  Two separate visitors are then used to manage the changes need for `PanacheEntity` and `PanacheCompanion` functionality.  `PanacheRepository` is all interface driven so no such distinctions were necessary there.

Now one of the fun wrinkles I learned in all of this is that kotlin compiles all the default methods on interfaces (though kotlin doesn't really use that term) actually get compiled in to the implementing type.  This means that we can't make simple assumptions about the method being present meaning that it's been overridden by the developer.  It took me some time to figure out how asm works but finally it all came together.

You'll notice a `JpaOperations.java` in there.  It is basically just a copy of what's in the existing panache module.  However, I did not want to depend directly on that as it would lead to confusing autocompletes at dev time having mirrored types available.  So I added some testing to be sure that as the "main" panache repo evolves, we can try to catch an API drift that needs to be cleaned up on the kotlin side.

There are differences in the `PanacheXBase` types between the java and kotlin sides of the world.  Because kotlin supports nullable types, I've been able to eliminate some methods from the kotlin variant.  Namely, anything having to do with `Optional` I've removed as having a `Entity?` type is much more idiomatic.

I'm sure I'm missing a few things but it's all up for discussion.  This PR is mostly intended to start the discussion.  It can ship as is but i'm sure i'm missing some nuances wrt panache compatibility and usage.


extensions/panache/panache-common/deployment/src/main/java/io/quarkus/panache/common/deployment/PanacheEntityEnhancer.java
```diff
@@ -189,13 +190,15 @@ private void generateMethod(MethodInfo method, AnnotationValue targetReturnTypeE
             mv.visitEnd();
         }
 
-        protected abstract String getModelDescriptor();
+        protected String getModelDescriptor() {
```
FroMage, line 193: This is not a correct default for all implementations.

integration-tests/hibernate-orm-panache-kotlin/src/main/kotlin/io/quarkus/it/panache/kotlin/Dog.kt
```diff
@@ -0,0 +1,21 @@
+package io.quarkus.it.panache.kotlin
+
+import io.quarkus.hibernate.orm.panache.kotlin.PanacheCompanion
+import io.quarkus.hibernate.orm.panache.kotlin.PanacheEntity
+import javax.persistence.Entity
+import javax.persistence.ManyToOne
+
+@Entity
+open class Dog() : PanacheEntity() {
+    companion object : PanacheCompanion<Dog>()
```
FroMage, line 10: Is there really no way to declare that in the `PanacheEntityBase` supertype? By making it self-parameterised with `PanacheEntity<Dog>`?

FroMage (commented):
Looks pretty cool.

FroMage:
I understand why you have your own `PanacheEntity` and `PanacheEntityBase` (and repos) but it's very unfortunate as we need to keep their methods synched. Your test helps. I'm not sure how to do better.

Do you have such signature changes for `PanacheQuery` too?

For `JpaOperations` we should probably add a `quarkus-panache-jpa-common` module and stick it there, no?

FroMage:
Initial issue was #4394.

extensions/panache/panache-common/deployment/src/main/java/io/quarkus/panache/common/deployment/PanacheEntityEnhancer.java
evanchooly: when I looked at the 3ish current implementations they all returned the same value.  I can change it back but it just seemed redundant.

FroMage: The `panache-rx` impl (in a branch ATM) doesn't, that's why.

evanchooly: ah.  does it *hurt* having that as the default that can be optionally overridden, though?  Seems like pointless work to push downstream if most are always going to be the same.  I'm happy to revert that if you'd like but it seems cleaner this way.

integration-tests/hibernate-orm-panache-kotlin/src/main/kotlin/io/quarkus/it/panache/kotlin/Dog.kt
evanchooly: companion objects aren't inherited so we can't go that route so far as I can tell.  Kotlin doesn't currently (and probably never will) support self types so I can't make that any less verbose.  At least, not that i've been able to determine.  It's not *bad*, though.  Just a touch redundant but it doesn't appear there's much to be done there.

evanchooly:
> I understand why you have your own `PanacheEntity` and `PanacheEntityBase` (and repos) but it's very unfortunate as we need to keep their methods synched. Your test helps. I'm not sure how to do better.
> 
> Do you have such signature changes for `PanacheQuery` too?
> 
> For `JpaOperations` we should probably add a `quarkus-panache-jpa-common` module and stick it there, no?

I just added a test for `PanacheQuery`.  (good catch.  i missed that one.)  It would be helpful to have that in a common area but it's not a huge burden as it is now.

geoand:
One comment I have (I haven't looked through the PR, hopefully tomorrow) is
that I would like to see comments on the actual transformers about what
they do and it's needed for Kotlin.

evanchooly:
@geoand that's a good idea.  i'll add some comments/javadoc in the relevant areas.

loicmathieu:
Very good that someone take a deep look at our Panache Kotlin support: great job!

As @FroMage, I'm very concerned with the fact that you duplicate a lot of code for Kotlin: JpaOperations, PanacheEntityBase, PanacheRepositoryBase, ...

This means that each time we add functionalityt to Panache (and it happens each week), we will need to adds in two places (in fact 6 places as we also have MongoDB implementations for imperative and reactive Panache that will need Kotlin variant).

We really need a way to lower the maintenance burden, this could be a nightmare if someone merge a PR that miss the Kotlin variant ... and it will happens !

If we keep all these implementations, we will need interfaces at least between Java and Kotlin version to be sure that they are 100% compliant API wise ... Or some build process ...

geoand:
> @geoand that's a good idea. i'll add some comments/javadoc in the relevant areas.

One reason I would like these comments is because it's entirely possible that we will need similar transformation in other domains.

evanchooly:
@loicmathieu I'm not a huge fan of the duplication but there are two requirements to this:

1.  Make kotlin work properly with quarkus
2.  Provide idiomatic APIs so that kotlin devs can write "proper" kotlin code.  

#2 is why you see things like a having a kotlin version of, say, `PanacheEntityBase`.  There are unnecessary elements in the Java API that completely unnecessary and regressive in kotlin.  On the other hand, `JpaOperations` is just a straight up copy.  That particular API will rarely if ever make it's way in to user code since it's used for behind the scenes default implementations.  As such, it can just be a 100% match.  We just can't rely on the Java module for this because autocomplete becomes a confusing nightmare of near-identical types making it hard to find the correct types for the kotlin implementation.

extensions/panache/hibernate-orm-panache-kotlin/deployment/src/main/java/io/quarkus/hibernate/orm/panache/kotlin/deployment/KotlinPanacheCompanionEnhancer.java
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    @Override
+    public ClassVisitor apply(String className, ClassVisitor outputClassVisitor) {
+        return new KotlinPanacheCompanionClassVisitor(className, outputClassVisitor, modelInfo, panacheEntityBaseClassInfo,
+                panacheCompanionClassInfo, indexView.getClassByName(DotName.createSimple(className)));
+    }
+
+    @Override
+    public void collectFields(ClassInfo classInfo) {
```
geoand, line 39: I think this is exactly the same as `KotlinPanacheEntityEnhancer#collectFields`

geoand:
Looking at the tests, it seems that entities and repositories needs to be open. I am pretty sure we can relax that constraint by applying a transformation like this: https://github.com/quarkusio/quarkus/blob/master/independent-projects/arc/processor/src/main/java/io/quarkus/arc/processor/Beans.java#L844

That file also contains code for generating a no-args constructor which is I see in some `TODO` comments.
I wonder if we could reuse those transformation instead of having each extension re-implement them.

geoand:
Also it seems that the new test module has not been added to the native tests CI run in `.github/workflows/ci-actions.yml`

extensions/panache/hibernate-orm-panache-kotlin/deployment/src/main/java/io/quarkus/hibernate/orm/panache/kotlin/deployment/KotlinPanacheCompanionEnhancer.java
evanchooly: It is, actually, i and I was thinking of removing it.  it's vestigial here, tbh, as kotlinc takes care of the field access for us.

geoand: :+1:

loicmathieu:
@evanchooly 

> On the other hand, JpaOperations is just a straight up copy. That particular API will rarely if ever make it's way in to user code since it's used for behind the scenes default implementations. 

If it's a straight up copy, and not used by the user, why do you need to copy it ? It contains 80% of the Panache implementation so I really don't want to copy it.

You also duplicate PanacheQuery only to support null-safe `firstResult` and `singleResult`, again I think it's too much maintenance burden to duplicate all API. Maybe (not a Kotlin expert) we can use extension methods to provides these two Kotlin methods and avoid duplicating.

I'm not a Kotlin expert but Spring provides Kotlin specialized class inside the same artifact as the rest of their implementation. Would it be possible to integrate the Kotlin code inside the `hibernate-orm-panache` extension instead of creating a new one? Like this, there will be less possibilitites that someone modifying Panache Java version will forget the Kotlin one.

Sorry if I ask a lot, your work is great but I'm very concern about the maintenance burden (as I said, I open new PR on Panache each week) and the risk that the Kotlin version will not be aligned with the Java one.

Maybe we should integrate a first implementation with only the companion object that will fix most of our current issues, and decide later if we need to provide a full new API for Kotlin ?

KaenDagger:
@evanchooly 
we already had `hibernate-orm-panache` extension, but for the kotlin  support you had to add the another extension `hibernate-orm-panache-kotlin`. So do we have to make specific kotlin extensions for other extensions for adding kotlin support ?
This PR gives a lot of information for transformation on kotlin, so thanks for that

FroMage:
@evanchooly you didn't say if `PanacheQuery` was also specialised for Kotlin?

I understand it is the case for `PanacheEntityBase` and `PanacheRepository`. Can you tell me what exactly was specialised?

And what alternative we have for specialising it in Java? Perhaps it's just some Kotlin type annotations we can put on the specialised methods?

IMO It's fine if we must specialise `PanacheEntityBase` and `PanacheRepository` because they don't contain code at all. But we should move `JpaOperations` to a module you can include if you didn't modify it. Same for `PanacheQuery` I guess.

integration-tests/hibernate-orm-panache-kotlin/src/main/kotlin/io/quarkus/it/panache/kotlin/Dog.kt
FroMage: Ah right, they're not inherited. OK fine.

extensions/panache/panache-common/deployment/src/main/java/io/quarkus/panache/common/deployment/PanacheEntityEnhancer.java
FroMage: No, I suppose it won't hurt. Fine.

evanchooly:
Honestly, everything there was copied from the original module.  the idea was to make whatever changes were needed to make the kotlin side work and feel idiomatic without having to worry about breaking the java implementation.  Now that that's done we can start looking at how we minimize those changes so we won't have to maintain wildly divergent variations.  Looking at the diffs on `PanacheQuery` the only thing that leaps out at me is the removal of the methods that use `Optional`.  This was done partly for aesthetic reasons but i'm not married to the idea.  It feels a little dirty to expose that method to kotlin users but we can happily ignore it if we want.  I have a slight allergy for parts of the API that get marked with "well, just ignore that part" but I can live with it for the simpler maintenance.  Probably both of the query classes could go in a common module.  Users will never see the `Impl` class so who cares what's in *there* for the most part.

For the `PanacheEntityBase` and `PanacheRepositoryBase`, i was able to slim down the API a bit more and make use of nullable types to eliminate some "fluff" on the kotlin side.  These changes, IMO, have bit more value and go a long way toward more idiomatic code.  Furthermore, the use of static methods make it ... complicated ... at best to properly model the entity/repository types.  We kinda really do need to make that Entity/Companion distinction to make compilation work.  With the tests in place catching any drift shouldn't be overly difficult though admittedly a touch more work than would be ideal.

As for merging the two modules, from an end-user/developer perspective that might make some sense in terms of what APIs are visible (we could still push any kotlin specific APIs to their own package, e.g.,) but on the deployment side, determining which Processor and ClassVisitors to use would get more complicated.  Now, I'm working on refining how to determine which methods to generate and when to rely on the class local definitions and so we would begin to consolidate there.  We would still have a bit of work to do to determine how to handle the entities as that's where things do get quite different.

e.g., since kotlinc already generates getters and setters, we don't need to generate those on the kotlin side unlike on the java side where we do.  Not a big deal, we simply filter out existing methods and only generate the missing.  On the Java side, that "missing set" would comprise all of the fields and things would work as they do today.  We would simply have to check that those methods are not marked as final and clear those flags if they are.  Also, very easy.

Now as for the static methods on `PanacheEntity` that part is harder to work around but I'm definitely open to suggestion there.  It might be that there's something clever we can do there and it might turn out that we just need the separate interfaces.  The repository interfaces *might* be mergeable but if the entity bits need separate interfaces, i would push for both entity and repository types to be pulled from the same packages for consistency's sake.

One of the things to keep in mind with all this, though, is that what we need here is something that feels reasonably idiomatic for kotlin developers.  That is part of my charge with this kotlin work coming from several rungs up the management chain.  While acknowledging my own proclivities toward nitpicking my way to "perfect" APIs, this is not just me fiddling for the sake of changing things.  It's an explicit goal as an org.  We just have to find the proper balance between those goals and the burden of maintenance.

evanchooly:
> @evanchooly
> we already had `hibernate-orm-panache` extension, but for the kotlin support you had to add the another extension `hibernate-orm-panache-kotlin`. So my question is do we have to make specific kotlin extensions for other extensions for adding kotlin support ?
> This PR gives a lot of information for transformation for kotlin, so thanks for that :)

I'm hoping that we won't need to do that.  But I have a feeling that we'll have to play with a few other areas and see what commonalities fall out and what can be generalized.  If it weren't for staticness of certain elements of the API here, i think we could like come to a really nice, slim accommodation between the java and kotlin sides of the world.  It'll come down to more experimentation/research and, honestly, user feedback.

evanchooly:
> @evanchooly
> 
> > On the other hand, JpaOperations is just a straight up copy. That particular API will rarely if ever make it's way in to user code since it's used for behind the scenes default implementations.
> 
> If it's a straight up copy, and not used by the user, why do you need to copy it ? It contains 80% of the Panache implementation so I really don't want to copy it.

It's a copy because the only other way to get to it would be to depend on that module and then that would result in multiple versions of, say, `PanacheEntity` and `PanacheRepository` showing up on the classpath and autocompletion in IDEs.  While we might end up there eventually, it's not my ideal scenario as it should be clear what to use and what's available.

> 
> You also duplicate PanacheQuery only to support null-safe `firstResult` and `singleResult`, again I think it's too much maintenance burden to duplicate all API. Maybe (not a Kotlin expert) we can use extension methods to provides these two Kotlin methods and avoid duplicating.

You may well be right there.  This module was "cloned" to be able to play with the API and see what's ultimately necessary and what's not.  This one is likely to not make that cut.  We just need to a common (API?) module to able to share these definitions without the problems mentioned above.

> 
> I'm not a Kotlin expert but Spring provides Kotlin specialized class inside the same artifact as the rest of their implementation. Would it be possible to integrate the Kotlin code inside the `hibernate-orm-panache` extension instead of creating a new one? Like this, there will be less possibilitites that someone modifying Panache Java version will forget the Kotlin one.

Yeah.  it's why i added some tests to try and catch as much of that automatically as possible.  Even in a shared module, though, there's the risk of changing the java version of the API and not the kotlin version so having one module is not a 100% fix either.

> 
> Sorry if I ask a lot, your work is great but I'm very concern about the maintenance burden (as I said, I open new PR on Panache each week) and the risk that the Kotlin version will not be aligned with the Java one.

No worries.  I invited questions.  :)  This is the kind of thing that will inform all our kotlin interop so we need to make sure we suss out all the questions and concerns as we can so we don't perpetuate poor design decisions elsewhere.

> 
> Maybe we should integrate a first implementation with only the companion object that will fix most of our current issues, and decide later if we need to provide a full new API for Kotlin ?

I think we can go a long way with `@Nullable` and `@NotNull` (kotlinc honors those annotations as if they were kotlin native nullable types).  The Entity/Companion divide is a harder gulf to cross but hopefully something clever will present itself.

evanchooly:
> Also it seems that the new test module has not been added to the native tests CI run in `.github/workflows/ci-actions.yml` (in the `native-tests` section)

I was unaware of this file, @geoand , but one of my questions coming up was about native.  I had thought that the IT tests covered that but then I remembered we turned that off because a full build was already taking a month as it was.

geoand:
Yeah, we need to include each test in a specific group to achieve test parallelization

FroMage:
@evanchooly I agree you can't copy `PanacheEntity` and `PanacheEntityBase` without modifications for kotlin. Probably not so true with `PanacheRepository` and `PanacheRepositoryBase` but they're just shells, so I'm honestly fine with copying them.

`JpaOperations` we should be able to share. Same for `PanacheQueryImpl` (I'm fine with copying the `PanacheQuery` interface).

I have one question regarding how kotlin stores the model information which is lost in compilation. In Ceylon we used to have a lot of model annotations to indicate the unerased type information. Including `@Ignore` to make a method invisible to Ceylon. Is there an equivalent in Kotlin? There must be a way to preserve mapping info. In Scala it used to be a compile-time annotation unique on the class (not on methods/fields) but even they moved away from this IIRC.

If there is an equivalent, that would give you a way to hide any `Optional` method we define in a shared module.

FroMage:
OK, so given that `PanacheEntityBase` for Kotlin looks very different due to the companion object, let's not merge the ORM/Kotlin modules. We can still create a shared module for common stuff like `JpaOperations` to ease maintenance.

evanchooly:
OK.  I found a bug or two when testing for how overloaded methods are handled so i took some time to clean it up.  I also found a *much* simpler way to achieve what I needed for the kotlinc output so what you'll find here is vastly simplified from the original PR.  Have at it, folks.  :D

extensions/panache/hibernate-orm-panache-kotlin/deployment/src/main/java/io/quarkus/hibernate/orm/panache/kotlin/deployment/KotlinPanacheResourceProcessor.java
```diff
@@ -0,0 +1,153 @@
...
+import io.quarkus.hibernate.orm.panache.kotlin.runtime.JpaOperations;
+import io.quarkus.panache.common.deployment.EntityField;
+import io.quarkus.panache.common.deployment.EntityModel;
+import io.quarkus.panache.common.deployment.MetamodelInfo;
+import io.quarkus.panache.common.deployment.PanacheFieldAccessEnhancer;
+import io.quarkus.panache.common.deployment.PanacheRepositoryEnhancer;
+
+public final class KotlinPanacheResourceProcessor {
+
+    private static final Set<DotName> UNREMOVABLE_BEANS = Collections.singleton(
```
geoand, line 42: This is odd. Why does the bean need to be unremovable? I mean what special thing does Kotlin Panache do that needs to make the bean unremovable when it previously wasn't required?

extensions/panache/hibernate-orm-panache-kotlin/runtime/src/test/kotlin/io/quarkus/hibernate/orm/panache/kotlin/runtime/TestAnalogs.kt
```diff
@@ -0,0 +1,90 @@
...
+        compare(map(io.quarkus.hibernate.orm.panache.PanacheRepositoryBase::class.java), map(PanacheRepositoryBase::class.java))
+    }
+
+/*
+    @Test
+    fun testPanacheEntity() {
+        compare(map(io.quarkus.hibernate.orm.panache.PanacheEntity::class.java), map(PanacheEntity::class))
+    }
+
+    private fun map(type: KClass<PanacheEntity>): MutableList<String> {
```
geoand, line 43: To be removed I assume?

extensions/panache/hibernate-orm-panache-kotlin/runtime/src/test/kotlin/io/quarkus/hibernate/orm/panache/kotlin/runtime/TestAnalogs.kt
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    @Test
+    fun testPanacheRepositoryBase() {
+        compare(map(io.quarkus.hibernate.orm.panache.PanacheRepositoryBase::class.java), map(PanacheRepositoryBase::class.java))
+    }
+
+/*
+    @Test
+    fun testPanacheEntity() {
```
geoand, line 39: Is this test disabled for a specific reason?

evanchooly: Only that I'd forgotten I'd disabled it because it was leading to so much debug output at the time.  Reenabled, though.

evanchooly:
rebased against master last night

extensions/panache/hibernate-orm-panache-kotlin/runtime/src/test/kotlin/io/quarkus/hibernate/orm/panache/kotlin/runtime/TestAnalogs.kt
evanchooly: The scope might be reduced.  We're debating how much needs to be cloned in to the kotlin side of the world and how much should be shared from a common module.

extensions/panache/hibernate-orm-panache-kotlin/deployment/src/main/java/io/quarkus/hibernate/orm/panache/kotlin/deployment/KotlinPanacheResourceProcessor.java
evanchooly: That was copied from the java side.  I honestly don't know what it's there for but assumed @FroMage knew what he was doing when that went in.

extensions/panache/hibernate-orm-panache-kotlin/runtime/src/test/kotlin/io/quarkus/hibernate/orm/panache/kotlin/runtime/TestAnalogs.kt
geoand: Gotcha

extensions/panache/hibernate-orm-panache-kotlin/deployment/src/main/java/io/quarkus/hibernate/orm/panache/kotlin/deployment/KotlinPanacheResourceProcessor.java
geoand: If it's in the Java side, then OK, let's keep it :)

geoand:
Seems like a rebase is needed since we since removed kscript from the cron job.

geoand:

evanchooly:
rebased for 5th time.  a review soon would be lovely.

geoand:
I think that @FroMage and @loicmathieu have the final say since this is all Panache stuff and they are the ones that will be doing all the heavy lifting in that area

FroMage:
Sorry I missed the cue that I had to re-review, as I was waiting for some answers from you. I will review tomorrow.

extensions/panache/hibernate-orm-panache-kotlin/deployment/src/main/java/io/quarkus/hibernate/orm/panache/kotlin/deployment/KotlinPanacheRepositoryClassVisitor.java
```diff
@@ -0,0 +1,97 @@
...
+
+    public KotlinPanacheRepositoryClassVisitor(String className, ClassVisitor outputClassVisitor, IndexView indexView) {
+        super(Gizmo.ASM_API_VERSION, outputClassVisitor);
+        this.indexView = indexView;
+        indexView.getClassByName(createSimple(className))
+                .methods()
+                .forEach(method -> {
+                    if (method.hasAnnotation(JandexUtil.DOTNAME_GENERATE_BRIDGE)) {
+                        bridgeMethods.put(method.name() + JandexUtil.getDescriptor(method,
+                                m -> m.equals("Entity") ? entitySignature : null), method);
```
FroMage, line 46: Don't repositories have an `Id` type parameter too?

extensions/panache/hibernate-orm-panache-kotlin/runtime/src/test/kotlin/io/quarkus/hibernate/orm/panache/kotlin/runtime/TestAnalogs.kt
FroMage: Mmm, it still appears disabled here. Commented out.

extensions/panache/panache-common/deployment/src/main/java/io/quarkus/panache/common/deployment/JandexUtil.java
```diff
@@ -346,6 +347,23 @@ private static void unbox(MethodVisitor mv, String owner, String methodName, Str
         mv.visitMethodInsn(Opcodes.INVOKEVIRTUAL, owner, methodName, "()" + returnTypeSignature, false);
     }
 
+    public static MethodInfo findMethod(ClassInfo classInfo, MethodInfo methodInfo) {
```
FroMage, line 350: This can't work, as it ignores type arguments. I've already removed this method two times because it's incorrect and lures people into thinking it will work when it does not. In both cases I didn't need that method.

FroMage (changes requested):
I've added some question. Overall we're still missing docs, and I'm very uncertain about the maintanance burden of not sharing the query implementation and `JpaOperations`. Case in point we've just added a change in paging for the query and support for named queries and I'm not sure this is supported here.

The plumbing is all done, so that's fine, but I feel really uncomfortable with this duplication.

Also, you haven't answered my question regarding the type signatures and whether it would be possible to reuse the `PanacheQuery` interface with a few kotlin type annotations.

FroMage:
OTOH, moving `PanacheQuery` to another module could work, but not if we rename its package, as that would not be binary compatible.
We could extract its methods into a new superinterface, but we'd have to find its new name and still override everymethod with covariant return types because they all return the current type. So that's vastly impractical.

Perhaps the less headache-inducing way to proceed is to give up on duplication for Kotlin, remove the special query/repository (and leave them for the Java processor, as they can be handled fine there) and only keep the `PanacheKotlinEntity`/`PanacheKotlinEntityBase`/`PanacheKotlinEntityCompanion` trio in the Kotlin module, which would delegate to the Java module's `JpaOperations`.

It would be less idiomatic, and would introduce a possible mishap where Kotlin users would extend `PanacheEntity` instead of `PanacheKotlinEntity` but we can document it and catch that error at build time and complain.

I tried to look for how Kotlin stores metadata info to be able to use an annotation to hide methods, but only came up with https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-metadata/ which seems to indicate a single annotation per class (Scala old style) which means very hard to support.

extensions/panache/panache-common/deployment/src/main/java/io/quarkus/panache/common/deployment/JandexUtil.java
evanchooly: No longer used.  removed.

evanchooly:
I've reworked things to delegate to the Java impls and smartened up the tests to ensure that the kotlin interfaces have all the same methods.  (I even found a missing one already!)  There's a comment above about documenting some internals from @geoand but since things have changed a fair bit I'd love for folks to have another look and see what needs documenting then I can doc what's needed.

There is a failing mutiny test but it fails on master, too, so is unrelated.

geoand:
> I've reworked things to delegate to the Java impls and smartened up the tests to ensure that the kotlin interfaces have all the same methods. (I even found a missing one already!) There's a comment above about documenting some internals from @geoand but since things have changed a fair bit I'd love for folks to have another look and see what needs documenting then I can doc what's needed.
> 
> There is a failing mutiny test but it fails on master, too, so is unrelated.

Sure thing, I'll check tomorrow morning.

extensions/panache/hibernate-orm-panache-kotlin/deployment/src/main/java/io/quarkus/hibernate/orm/panache/kotlin/deployment/KotlinPanacheCompanionClassVisitor.java
```diff
@@ -0,0 +1,98 @@
...
+import java.util.TreeMap;
+
+import org.jboss.jandex.ClassInfo;
+import org.jboss.jandex.MethodInfo;
+import org.objectweb.asm.ClassVisitor;
+import org.objectweb.asm.MethodVisitor;
+
+import io.quarkus.panache.common.deployment.JandexUtil;
+
+class KotlinPanacheCompanionClassVisitor extends ClassVisitor {
```
geoand, line 23: I think this needs to be have some comments on what it does

extensions/panache/hibernate-orm-panache-kotlin/deployment/src/main/java/io/quarkus/hibernate/orm/panache/kotlin/deployment/KotlinPanacheEntityClassVisitor.java
```diff
@@ -0,0 +1,84 @@
...
+import org.objectweb.asm.MethodVisitor;
+import org.objectweb.asm.Opcodes;
+
+import io.quarkus.panache.common.deployment.EntityField;
+import io.quarkus.panache.common.deployment.EntityModel;
+import io.quarkus.panache.common.deployment.MetamodelInfo;
+import io.quarkus.panache.common.deployment.PanacheEntityEnhancer;
+import io.quarkus.panache.common.deployment.PanacheMethodCustomizer;
+
+class KotlinPanacheEntityClassVisitor extends PanacheEntityEnhancer.PanacheEntityClassVisitor<EntityField> {
```
geoand, line 17: Looks like this removes the `final` modifier among other things, but it would good to have some comments

integration-tests/pom.xml
```diff
@@ -17,7 +17,7 @@
 
     <properties>
         <maven.deploy.skip>true</maven.deploy.skip>
-        <quarkus.build.skip>${skipTests}</quarkus.build.skip>
+<!--        <quarkus.build.skip>${skipTests}</quarkus.build.skip>-->
```
geoand, line 20: I think we need to get rid of this :wink:

evanchooly: oops.  yeah.  usually i catch that.  ;)

extensions/panache/hibernate-orm-panache-kotlin/deployment/src/main/java/io/quarkus/hibernate/orm/panache/kotlin/deployment/KotlinPanacheCompanionClassVisitor.java
geoand: Great, thanks!

core/runtime/src/main/java/io/quarkus/runtime/ApplicationLifecycleManager.java
```diff
@@ -27,7 +27,7 @@
  * The {@link Application} class is responsible for starting and stopping the application,
  * but nothing else. This class can be used to run both persistent applications that will run
  * till they receive a signal, and command mode applications that will run until the main method
- * returns. This class registers a shutdown hook to properly shut down the application, and handles
+ * returns. This class registers a shutªdown hook to properly shut down the application, and handles
```
FroMage, line 30: WTF happened here? ;)

extensions/panache/hibernate-orm-panache-kotlin/deployment/src/main/java/io/quarkus/hibernate/orm/panache/kotlin/deployment/KotlinPanacheRepositoryClassVisitor.java
FroMage: Up.

FroMage (dismissed):
Two comments still. But I still think we should get rid of the extra JpaOperations and QueryImpl code.

Do you want me to do it on your PR?

core/runtime/src/main/java/io/quarkus/runtime/ApplicationLifecycleManager.java
evanchooly: oops.  i switch between windows and mac regularly and the muscle sometimes lags.  :)

evanchooly:
@FroMage i removed that filter as it is ultimately unnecessary for how the kotlin visitors work.  It was a carryover from the java versions but are ultimately vestigial for the kotlin side.

evanchooly:
as for the removal of the duplicated JpaOperations and QueryImpl if you want to do it in my PR directly or in a follow up PR, either way works for me.

FroMage:
I can work on your PR, but can you rebase and make sure all tests pass first? If I can't rely on CI to tell me if I messed up, it's going to be really hard.

evanchooly:
@FroMage done

FroMage:
@evanchooly please allow me to push my commit to your PR

evanchooly:
@FroMage done

FroMage:
Apparently not
```
To github.com:evanchooly/quarkus.git
 ! [remote rejected]       kotlin-panache -> kotlin-panache (permission denied)
error: failed to push some refs to 'git@github.com:evanchooly/quarkus.git'
```

evanchooly:
I've clicked the "allow edits by maintainers" button 8 times now...

FroMage:
> I've clicked the "allow edits by maintainers" button 8 times now...

just grant me push rights personally

FroMage:
I've put it up at https://github.com/quarkusio/quarkus/pull/9145 meanwhile, to see if CI passes.

FroMage:
I broke some tests lemme work on fixing them before I push here.

FroMage:
CI passed on my branch, so pushing here.

Note that because I'm moving `PanacheQueryImpl` and `JpaOperations` to a new module, pretty much all "ORM with Panache" bugs/features should be on hold until this gets merged, because the target classes will have moved and be super hard to rebase.

So I'd like to get this merged on the fast lane, if it's OK with @gsmet ?

AFAICT this is ready to be merged, since the new extension is experimental, and the affected extension (ORM with Panache) is well tested.

The only thing we need to check is that I've created a new extension `quarkus-hibernate-orm-panache-common` which has a common build step to handle named queries, so I had to make it an extension, but I didn't add a `quarkus-extension.yaml` to not make it visible, nor add a feature for it.

Is this the right way to do it @aloubyansky or should I still add that file?

FroMage:
Ah, damnit this has conflicts. @evanchooly I can fix them but I'll have to force-push which means you'll have to force-pull. OK?

FroMage:
OK rebased, let's see what CI says.

FroMage:
Fixed the damn imports sort order.

FroMage:
Well, CI passed. That's one thing.

aloubyansky:
> The only thing we need to check is that I've created a new extension `quarkus-hibernate-orm-panache-common` which has a common build step to handle named queries, so I had to make it an extension, but I didn't add a `quarkus-extension.yaml` to not make it visible, nor add a feature for it.

If you don't add `quarkus-extension.yaml` the basic one will be generated. If you want it to be unlisted you should just add `unlisted: true`. We should probably be generating `unlisted: true` by default to force people provide a meaningful descriptor. You can find a lot of unlisted extensions in the Quarkus repo (just search for YAML containing `unlisted`.
I am not sure I understand why you __had to__ make that an extension though.

aloubyansky:
Ok, I saw your comment on zulip.

geoand:
I'm in favor of getting this in so you guys can continue working on Panache

evanchooly:
So can this get merged now?  What's the status after @FroMage's refactor?

geoand:
Well I think we all agree it needs to go in. I'll approve and you guys figure out the rest :wink:

geoand:
Actually, was the `unlisted` thing added?

FroMage:
OK thanks, I've added an unlisted extension descriptor. Should be good, then?

geoand (approved):

FroMage:
Yay, great!

Now we need docs!

@loicmathieu please note that the impl of PanacheQueryImpl and JpaOperations have moved to the new quarkus-hibernate-orm-panache-common module.

loicmathieu:
> Now we need docs!

+1 ;)

And we will need the same for MongoDB, @evanchooly I cretaed https://github.com/quarkusio/quarkus/issues/9211 to track this. If you plan to work on this don't hesitate to assign it to yourself.

evanchooly:
oof.  ok.  I'll take a look.  probably easier for me to tackle it but we'll see how it goes.

PR #8334, opened 2020-04-01
New panache-mock module for mocking Panache static methods

FroMage:
Tied to `Mockito`, usage:

```java
// start mocking this entity's static methods
        PanacheMock.mock(Person.class);

// mocked method returns the Mockito defaults: 0 in this case
        Assertions.assertEquals(0, Person.count());

// specify a new return value
        Mockito.when(Person.count()).thenReturn(23l);
        Assertions.assertEquals(23, Person.count());

// specify a new return value
        Mockito.when(Person.count()).thenReturn(42l);
        Assertions.assertEquals(42, Person.count());

// now delegate to the original method
        Mockito.when(Person.count()).thenCallRealMethod();
// Here H2 is running so we do get a result
        Assertions.assertEquals(0, Person.count());

// Verify that our mock method was called
        Mockito.verify(PanacheMock.getMock(Person.class), Mockito.times(4)).count();

// Same with params
        Mockito.when(Person.findById(12l)).thenReturn(p);
        Assertions.assertSame(p, Person.findById(12l));
        Assertions.assertNull(Person.findById(42l));

        Mockito.when(Person.findById(12l)).thenThrow(new WebApplicationException());
        try {
            Person.findById(12l);
            Assertions.fail();
        } catch (WebApplicationException x) {
        }
        
// We can even mock the user's public static methods
        Mockito.when(Person.findOrdered()).thenReturn(Collections.emptyList());
        Assertions.assertTrue(Person.findOrdered().isEmpty());
```

### DIFF
--- a/bom/runtime/pom.xml
+++ b/bom/runtime/pom.xml
@@ -429,6 +429,11 @@
             </dependency>
             <dependency>
                 <groupId>io.quarkus</groupId>
+                <artifactId>quarkus-panache-mock</artifactId>
+                <version>${project.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>io.quarkus</groupId>
                 <artifactId>quarkus-panacheql</artifactId>
                 <version>${project.version}</version>
             </dependency>
--- a/docs/src/main/asciidoc/getting-started-testing.adoc
+++ b/docs/src/main/asciidoc/getting-started-testing.adoc
@@ -569,6 +569,10 @@ public class GreetingResourceTest {
 ----
 <1> Indicate that this injection point is meant to use an instance of `RestClient`.
 
+=== Mocking with Panache
+
+If you are using the `quarkus-hibernate-orm-panache` or `quarkus-mongodb-panache` extensions, check out the link:hibernate-orm-panache#mocking[Hibernate ORM with Panache Mocking] and link:mongodb-panache#mocking[MongoDB with Panache Mocking] documentation for the easiest way to mock your data access.
+
 == Test Bootstrap Configuration Options
 
 There are a few system properties that can be used to tune the bootstrap of the test, specifically its classpath.
--- a/docs/src/main/asciidoc/hibernate-orm-panache.adoc
+++ b/docs/src/main/asciidoc/hibernate-orm-panache.adoc
@@ -611,7 +611,7 @@ public void create(Parameter parameter){
 
 Panache provides direct support for database locking with your entity/repository, using `findById(Object, LockModeType)` or `find().withLock(LockModeType)`.
 
-The following examples are for the entity pattern, but the same can be used with repositories.
+The following examples are for the active record pattern, but the same can be used with repositories.
 
 === First: Locking using findById().
 
@@ -683,6 +683,184 @@ and specify your ID type as an extra type parameter:
 @ApplicationScoped
 public class PersonRepository implements PanacheRepositoryBase<Person,Integer> {
     //...
+}
+----
+
+== Mocking
+
+=== Using the active record pattern
+
+If you are using the active record pattern you cannot use Mockito directly as it does not support mocking static methods,
+but you can use the `quarkus-panache-mock` module which allows you to use Mockito to mock all provided static
+methods, including your own.
+
+Add this dependency to your `pom.xml`:
+
+[source,xml]
+----
+<dependency>
+    <groupId>io.quarkus</groupId>
+    <artifactId>quarkus-panache-mock</artifactId>
+    <scope>test</scope>
+</dependency>
+----
+
+
+Given this simple entity:
+
+[source,java]
+----
+@Entity
+public class Person extends PanacheEntity {
+
+    public String name;
+
+    public static List<Person> findOrdered() {
+        return find("ORDER BY name").list();
+    }
+}
+----
+
+You can write your mocking test like this:
+
+[source,java]
+----
+@QuarkusTest
+public class PanacheFunctionalityTest {
+
+    @Test
+    public void testPanacheMocking() {
+        PanacheMock.mock(Person.class);
+
+        // Mocked classes always return a default value
+        Assertions.assertEquals(0, Person.count());
+
+        // Now let's specify the return value
+        Mockito.when(Person.count()).thenReturn(23l);
+        Assertions.assertEquals(23, Person.count());
+
+        // Now let's change the return value
+        Mockito.when(Person.count()).thenReturn(42l);
+        Assertions.assertEquals(42, Person.count());
+
+        // Now let's call the original method
+        Mockito.when(Person.count()).thenCallRealMethod();
+        Assertions.assertEquals(0, Person.count());
+
+        // Check that we called it 4 times
+        PanacheMock.verify(Person.class, Mockito.times(4)).count();// <1>
+
+        // Mock only with specific parameters
+        Person p = new Person();
+        Mockito.when(Person.findById(12l)).thenReturn(p);
+        Assertions.assertSame(p, Person.findById(12l));
+        Assertions.assertNull(Person.findById(42l));
+
+        // Mock throwing
+        Mockito.when(Person.findById(12l)).thenThrow(new WebApplicationException());
+        Assertions.assertThrows(WebApplicationException.class, () -> Person.findById(12l));
+
+        // We can even mock your custom methods
+        Mockito.when(Person.findOrdered()).thenReturn(Collections.emptyList());
+        Assertions.assertTrue(Person.findOrdered().isEmpty());
+
+        PanacheMock.verify(Person.class).findOrdered();
+        PanacheMock.verify(Person.class, Mockito.atLeastOnce()).findById(Mockito.any());
+        PanacheMock.verifyNoMoreInteractions(Person.class);
+    }
+}
+----
+<1> Be sure to call your `verify` methods on `PanacheMock` rather than `Mockito`, otherwise you won't know
+what mock object to pass.
+
+=== Using the repository pattern
+
+If you are using the repository pattern you can use Mockito directly, using the `quarkus-junit5-mockito` module,
+which makes mocking beans much easier:
+
+[source,java]
+----
+<dependency>
+    <groupId>io.quarkus</groupId>
+    <artifactId>quarkus-junit5-mockito</artifactId>
+    <scope>test</scope>
+</dependency>
+----
+
+Given this simple entity:
+
+[source,java]
+----
+@Entity
+public class Person {
+
+    @Id
+    @GeneratedValue
+    public Long id;
+
+    public String name;
+}
+----
+
+And this repository:
+
+[source,java]
+----
+@ApplicationScoped
+public class PersonRepository implements PanacheRepository<Person> {
+    public List<Person> findOrdered() {
+        return find("ORDER BY name").list();
+    }
+}
+----
+
+You can write your mocking test like this:
+
+[source,java]
+----
+@QuarkusTest
+public class PanacheFunctionalityTest {
+    @InjectMock
+    PersonRepository personRepository;
+
+    @Test
+    public void testPanacheRepositoryMocking() throws Throwable {
+        // Mocked classes always return a default value
+        Assertions.assertEquals(0, personRepository.count());
+
+        // Now let's specify the return value
+        Mockito.when(personRepository.count()).thenReturn(23l);
+        Assertions.assertEquals(23, personRepository.count());
+
+        // Now let's change the return value
+        Mockito.when(personRepository.count()).thenReturn(42l);
+        Assertions.assertEquals(42, personRepository.count());
+
+        // Now let's call the original method
+        Mockito.when(personRepository.count()).thenCallRealMethod();
+        Assertions.assertEquals(0, personRepository.count());
+
+        // Check that we called it 4 times
+        Mockito.verify(personRepository, Mockito.times(4)).count();
+
+        // Mock only with specific parameters
+        Person p = new Person();
+        Mockito.when(personRepository.findById(12l)).thenReturn(p);
+        Assertions.assertSame(p, personRepository.findById(12l));
+        Assertions.assertNull(personRepository.findById(42l));
+
+        // Mock throwing
+        Mockito.when(personRepository.findById(12l)).thenThrow(new WebApplicationException());
+        Assertions.assertThrows(WebApplicationException.class, () -> personRepository.findById(12l));
+
+        Mockito.when(personRepository.findOrdered()).thenReturn(Collections.emptyList());
+        Assertions.assertTrue(personRepository.findOrdered().isEmpty());
+
+        // We can even mock your custom methods
+        Mockito.verify(personRepository).findOrdered();
+        Mockito.verify(personRepository, Mockito.atLeastOnce()).findById(Mockito.any());
+        Mockito.verifyNoMoreInteractions(personRepository);
+    }
 }
 ----
 
--- a/docs/src/main/asciidoc/mongodb-panache.adoc
+++ b/docs/src/main/asciidoc/mongodb-panache.adoc
@@ -867,6 +867,181 @@ public Multi<ReactivePerson> streamPersons() {
 
 TIP: `@SseElementType(MediaType.APPLICATION_JSON)` tells RESTEasy to serialize the object in JSON.
 
+== Mocking
+
+=== Using the active-record pattern
+
+If you are using the active-record pattern you cannot use Mockito directly as it does not support mocking static methods,
+but you can use the `quarkus-panache-mock` module which allows you to use Mockito to mock all provided static
+methods, including your own.
+
+Add this dependency to your `pom.xml`:
+
+[source,xml]
+----
+<dependency>
+    <groupId>io.quarkus</groupId>
+    <artifactId>quarkus-panache-mock</artifactId>
+    <scope>test</scope>
+</dependency>
+----
+
+Given this simple entity:
+
+[source,java]
+----
+public class Person extends PanacheMongoEntity {
+
+    public String name;
+
+    public static List<Person> findOrdered() {
+        return findAll(Sort.by("lastname", "firstname")).list();
+    }
+}
+----
+
+You can write your mocking test like this:
+
+[source,java]
+----
+@QuarkusTest
+public class PanacheFunctionalityTest {
+
+    @Test
+    public void testPanacheMocking() {
+        PanacheMock.mock(Person.class);
+
+        // Mocked classes always return a default value
+        Assertions.assertEquals(0, Person.count());
+
+        // Now let's specify the return value
+        Mockito.when(Person.count()).thenReturn(23l);
+        Assertions.assertEquals(23, Person.count());
+
+        // Now let's change the return value
+        Mockito.when(Person.count()).thenReturn(42l);
+        Assertions.assertEquals(42, Person.count());
+
+        // Now let's call the original method
+        Mockito.when(Person.count()).thenCallRealMethod();
+        Assertions.assertEquals(0, Person.count());
+
+        // Check that we called it 4 times
+        PanacheMock.verify(Person.class, Mockito.times(4)).count();// <1>
+
+        // Mock only with specific parameters
+        Person p = new Person();
+        Mockito.when(Person.findById(12l)).thenReturn(p);
+        Assertions.assertSame(p, Person.findById(12l));
+        Assertions.assertNull(Person.findById(42l));
+
+        // Mock throwing
+        Mockito.when(Person.findById(12l)).thenThrow(new WebApplicationException());
+        Assertions.assertThrows(WebApplicationException.class, () -> Person.findById(12l));
+
+        // We can even mock your custom methods
+        Mockito.when(Person.findOrdered()).thenReturn(Collections.emptyList());
+        Assertions.assertTrue(Person.findOrdered().isEmpty());
+
+        PanacheMock.verify(Person.class).findOrdered();
+        PanacheMock.verify(Person.class, Mockito.atLeastOnce()).findById(Mockito.any());
+        PanacheMock.verifyNoMoreInteractions(Person.class);
+    }
+}
+----
+<1> Be sure to call your `verify` methods on `PanacheMock` rather than `Mockito`, otherwise you won't know
+what mock object to pass.
+
+=== Using the repository pattern
+
+If you are using the repository pattern you can use Mockito directly, using the `quarkus-junit5-mockito` module,
+which makes mocking beans much easier:
+
+[source,java]
+----
+<dependency>
+    <groupId>io.quarkus</groupId>
+    <artifactId>quarkus-junit5-mockito</artifactId>
+    <scope>test</scope>
+</dependency>
+----
+
+Given this simple entity:
+
+[source,java]
+----
+public class Person {
+
+    @BsonId
+    public Long id;
+
+    public String name;
+}
+----
+
+And this repository:
+
+[source,java]
+----
+@ApplicationScoped
+public class PersonRepository implements PanacheMongoRepository<Person> {
+    public List<Person> findOrdered() {
+        return findAll(Sort.by("lastname", "firstname")).list();
+    }
+}
+----
+
+You can write your mocking test like this:
+
+[source,java]
+----
+@QuarkusTest
+public class PanacheFunctionalityTest {
+    @InjectMock
+    PersonRepository personRepository;
+
+    @Test
+    public void testPanacheRepositoryMocking() throws Throwable {
+        // Mocked classes always return a default value
+        Assertions.assertEquals(0, personRepository.count());
+
+        // Now let's specify the return value
+        Mockito.when(personRepository.count()).thenReturn(23l);
+        Assertions.assertEquals(23, personRepository.count());
+
+        // Now let's change the return value
+        Mockito.when(personRepository.count()).thenReturn(42l);
+        Assertions.assertEquals(42, personRepository.count());
+
+        // Now let's call the original method
+        Mockito.when(personRepository.count()).thenCallRealMethod();
+        Assertions.assertEquals(0, personRepository.count());
+
+        // Check that we called it 4 times
+        Mockito.verify(personRepository, Mockito.times(4)).count();
+
+        // Mock only with specific parameters
+        Person p = new Person();
+        Mockito.when(personRepository.findById(12l)).thenReturn(p);
+        Assertions.assertSame(p, personRepository.findById(12l));
+        Assertions.assertNull(personRepository.findById(42l));
+
+        // Mock throwing
+        Mockito.when(personRepository.findById(12l)).thenThrow(new WebApplicationException());
+        Assertions.assertThrows(WebApplicationException.class, () -> personRepository.findById(12l));
+
+        Mockito.when(personRepository.findOrdered()).thenReturn(Collections.emptyList());
+        Assertions.assertTrue(personRepository.findOrdered().isEmpty());
+
+        // We can even mock your custom methods
+        Mockito.verify(personRepository).findOrdered();
+        Mockito.verify(personRepository, Mockito.atLeastOnce()).findById(Mockito.any());
+        Mockito.verifyNoMoreInteractions(personRepository);
+    }
+}
+----
+
+
 == How and why we simplify MongoDB API
 
 When it comes to writing MongoDB entities, there are a number of annoying things that users have grown used to
--- a/extensions/panache/panache-common/deployment/src/main/java/io/quarkus/panache/common/deployment/PanacheRepositoryEnhancer.java
+++ b/extensions/panache/panache-common/deployment/src/main/java/io/quarkus/panache/common/deployment/PanacheRepositoryEnhancer.java
@@ -1,7 +1,11 @@
 package io.quarkus.panache.common.deployment;
 
 import java.lang.reflect.Modifier;
+import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
+import java.util.Set;
 import java.util.function.BiFunction;
 
 import org.jboss.jandex.AnnotationInstance;
@@ -37,10 +41,15 @@ public abstract class PanacheRepositoryEnhancer implements BiFunction<String, Cl
         protected Type entityType;
         protected String entitySignature;
         protected String entityBinaryType;
+        protected String idSignature;
+        protected String idBinaryType;
         protected String daoBinaryName;
         protected ClassInfo daoClassInfo;
         protected ClassInfo panacheRepositoryBaseClassInfo;
         protected IndexView indexView;
+        protected Map<String, String> typeArguments = new HashMap<>();
+        // set of name + "/" + descriptor
+        protected Set<String> userMethods = new HashSet<>();
 
         public PanacheRepositoryClassVisitor(String className, ClassVisitor outputClassVisitor,
                 ClassInfo panacheRepositoryBaseClassInfo, IndexView indexView) {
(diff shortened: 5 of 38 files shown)


FroMage:
So the injected implementation looks like this at the start of every static public method:

```java
if(PanacheMock.IsMockEnabled && PanacheMock.isMocked(Person.class)) {
  try {
    return (int)PanacheMock.mockMethod(Person.class, "count", new Class<?>[] {}, new Object[] {});
  } catch (PanacheMock.InvokeRealMethodException e) {
    // fall-through
  }
}
```

I still have to make this conditional to the TEST mode.

Now the question is how do I tie this with the `quarkus-test` module?

Users probably don't want to write tests like this:

```java
    @AfterEach
    public void afterEach() {
// make sure one test's mocking doesn't leak to other tests
        PanacheMock.reset();
    }

    @Test
    public void testPanacheMocking() {
// specify all mocked classes
        PanacheMock.mock(Person.class, Dog.class);
    }
```

So perhaps I should make `PanacheMock.reset()` called by default after each test? How? I don't want to make the `quarkus-test` module depend on `panache-mock`.

Is there value in this alternative form to declare mocked classes?

```java
    @Test
    @PanacheMockClasses({Person.class, Dog.class})
    public void testPanacheMocking() {
    }
```

Doesn't look very justified to me.

loicmathieu:
This looks promising !

> So perhaps I should make PanacheMock.reset() called by default after each test? How? I don't want to make the quarkus-test module depend on panache-mock.

You need to turn PanacheMock in a Mockito extension or a JUnit extension.

For Mockito  extension point, I read quickly the documentation and you can provide you own `MockMaker` https://javadoc.io/static/org.mockito/mockito-core/3.3.3/org/mockito/plugins/MockMaker.html this seems to be close to what you want to achieve or you can use one of the integration point of Mockito: https://javadoc.io/static/org.mockito/mockito-core/3.3.3/org/mockito/Mockito.html#41

For JUnit extension, the better is to implements `AfterTestExecutionCallback` https://junit.org/junit5/docs/current/user-guide/#extensions-lifecycle-callbacks-before-after-execution then create a JUnit Extension that extends the `QuarkusTest` extension and adds mocking capabilities. So the user can just use `@PanacheTest` instead of `@QuarkusTest`.

If you don't need to integrate too deep into Mockito the JUnit extension is the easiest.

Some early feedback on your PR:

- I don't see the `PanacheMock` class in it.
- `Mockito.verify(PanacheMock.getMock(Person.class), Mockito.times(4)).count();` it didn't seems very convenient to need to call `getMock` to verify the mock behaviour. Maybe the `PanacheMock.mock(Person.class)` can return a mock instance that we can use with the verify method, or maybe you can provides a `PanacheMock.verify(Person.class, Mockito.times(4))` method that allow verifycation.
- As I understand the PR, you add the hability to create method customizer to mock them. I'm not fund of using the standard build mechanism for mocking stuff, I would expect mocking happening at test execution time only, but I can understand that it's the easiest (and maybe the only) way to do it.

bom/runtime/pom.xml
```diff
@@ -409,6 +409,11 @@
                 <artifactId>quarkus-panache-common</artifactId>
                 <version>${project.version}</version>
             </dependency>
+            <dependency>
+                <groupId>io.quarkus</groupId>
+                <artifactId>quarkus-panache-mock</artifactId>
```
gsmet, line 414: I don't see a panache-mock module in what you pushed. Is that expected?

gsmet (dismissed):
Looks like there's some code missing.

Also, big question: should this really be an extension or should it be part of our test framework modules? I would favor the latter if there's no good reason to make it an extension.

FroMage:
> Looks like there's some code missing.

Yeah, I was missing the extension, sorry. Updated.

FroMage:
> Also, big question: should this really be an extension or should it be part of our test framework modules?

So ATM, the extension depends on `PanacheMethodCustomizer` and `PanacheMethodCustomizerBuildItem` which are in `panache-common`, and I didn't want to make `quarkus-junit5` depend on `panache-common`.

This works because `hibernate-orm-panache` depends on `panache-common`. It didn't feel right to make it depend on `quarkus-junit5` either.

We could place the customizer interface and build item somewhere else, because they don't really depend on anything Panache, but where?

extensions/panache/panache-mock/runtime/src/main/java/io/quarkus/panache/mock/PanacheMock.java
```diff
@@ -0,0 +1,81 @@
...
+import org.mockito.internal.invocation.RealMethod;
+import org.mockito.internal.util.MockUtil;
+import org.mockito.invocation.MockHandler;
+import org.mockito.mock.MockCreationSettings;
+
+public class PanacheMock {
+
+    public static boolean IsMockEnabled = false;
+
+    private final static Map<Class<?>, Object> mocks = new HashMap<>();
```
loicmathieu, line 19: Maybe use a ConcurentHashMap in case tests are run in parallel ?

FroMage:
> Mockito.verify(PanacheMock.getMock(Person.class), Mockito.times(4)).count(); it didn't seems very convenient to need to call getMock to verify the mock behaviour. Maybe the PanacheMock.mock(Person.class) can return a mock instance that we can use with the verify method, or maybe you can provides a PanacheMock.verify(Person.class, Mockito.times(4)) method that allow verifycation.

I could add that, yes.

> As I understand the PR, you add the hability to create method customizer to mock them. I'm not fund of using the standard build mechanism for mocking stuff, I would expect mocking happening at test execution time only, but I can understand that it's the easiest (and maybe the only) way to do it.

Well we have to inject _some_ code in the methods at build time so they can be mocked. We should only do it in the TEST mode, though, so I think it's OK.

extensions/panache/panache-mock/runtime/src/main/java/io/quarkus/panache/mock/PanacheMock.java
FroMage: That would not help: tests can't run in parallel with that setup, because the set of classes to mock and the methods mocked is test-specific. So we'd have to make this a thread-local, which would mean that reactive code would likely fail or require context propagation.

loicmathieu:
Just read the PR again with the panache-mock extension and it is clearer now ;)

> Well we have to inject some code in the methods at build time so they can be mocked. We should only do it in the TEST mode, though, so I think it's OK.

I agree, in fact I found now the approach OK, we can reuse it if needed (for example to fix some Kotlin issue ?, or to provides specialization for some databases ?) so it's great.

> Also, big question: should this really be an extension or should it be part of our test framework modules?

I think that being a panache-mock extension is OK, Panache users will tipically add it with the test scope if they need it (not everybody uses panache entity).
If some other areas need it, you can then move some of it's facility inside a global Quarkus mock extension (for example to mock messaging system ?) ...

Of course, if it could be only a JUnit or a Mockito extension, this would be better as everything will be a test runtime and scoped to a single test instead of at build time and scopped at the entire test suite. But I'm OK with this anyway.

extensions/panache/panache-mock/runtime/src/main/java/io/quarkus/panache/mock/PanacheMock.java
loicmathieu: I think it's OK for a first implementation as soon as it is clearly documented that test with panache mock should not be run in parallel.

FroMage:
> Of course, if it could be only a JUnit or a Mockito extension, this would be better as everything will be a test runtime and scoped to a single test instead of at build time and scopped at the entire test suite. But I'm OK with this anyway.

As I said, there _has_ to be something at build-time to inject a hook into the static methods so they can delegate to mocking. I don't know how to do otherwise.

FroMage:
- Rebased on master
- Now depend on the new `quarkus-junit5-mockito`
- Uses that to reset all mocks between each test method
- Added mongodb-panache test
- Added `PanacheMock.verify(Class)` methods
- Only generate the mock calls in TEST mode

FroMage:
So now the only thing left IMO is where this extension sits.

ATM it's an extension `panache-mock` because it must generate a `BuildItem` that `*-panache` will collect and call.

That build item is coupled with an interface `PanacheMethodCustomiser`. Those are currently defined in `panache-common` which is not an extension.

I could move their implementations to the `panache-junit5-mockito` module, except it would have to depend on `panache-common` (perhaps fine, WDYT @geoand?), and it would have to be able to inject a build item into the build. I've no idea how to do that, if it's even possible when you're not a Quarkus extension.

Perhaps @geoand or @stuartwdouglas have ideas?

geoand:
Is having a specific panache testing module not a good option? 
It is certainly easier (and perhaps cleaner?) than doing anything else.

FroMage:
No idea. @gsmet asked the question, and it's a good question.

It does make sense that `quarkus-junit5-mockito` would give us mocks for everything, including Panache stuff: it makes it easier for users because they get all quarkus/mockito in one place.

OTOH it's not that annoying to have a dedicated extension IMO.

So I'm really open to opinions here.

geoand:
I think that the panache testing stuff should be a dedicated module (that perhaps includes `quarkus-junit5-mockito`).

I don't think it will be a problem for users to add a test dependency. We can also make it clear in the release notes and the docs. Plus we can augment our project generation to add that dependency when the Panache extension is added.

loicmathieu:
I'm OK with `panache-mock` being an extension, as long as you don't create a `FeatureBuildItem` it will be an hidden one and the user will not see it.

A few questions:

- Users will ask if they can use PanacheMock to mock a repository. I don't think it provides any value to do this compared to raw Mockito so this should be documented (I don't know what will happens if some do `PanacheMock.mock(PersonRepository.class)`).
- Can you add an example of mocking a PanacheQuery ? Maybe we will need support for this also as it will be very common use case: `PanacheMock.mockPanacheQuery(results)` could return a mock of PanacheQuery with the list of results inside, then a user can paginate it, get the first result, ... Maybe it's a different story ;)
- I usually end my tests with `Mockito.verifyNoMoreInterration()` to check that all calls to my mocks has been verify, will this methods works ? Can  you have a test for it ?

FroMage:
> I'm OK with panache-mock being an extension, as long as you don't create a FeatureBuildItem it will be an hidden one and the user will not see it.

Yeah, I should definitely remove it.

> Users will ask if they can use PanacheMock to mock a repository. I don't think it provides any value to do this compared to raw Mockito so this should be documented (I don't know what will happens if some do PanacheMock.mock(PersonRepository.class)).

Well Mockito can mock repos already so there's no need for that. It will not work if you pass a repo. I could add a type constraint for `PanacheEntityBase` but it would add a dependency to it, and we don't want that, unless we add a marker super interface to both orm/mongo entities. `MockablePanacheEntity` with no method. That should work.

> Can you add an example of mocking a PanacheQuery ? Maybe we will need support for this also as it will be very common use case: PanacheMock.mockPanacheQuery(results) could return a mock of PanacheQuery with the list of results inside, then a user can paginate it, get the first result, ... Maybe it's a different story ;)

I can't because it would add a dependency to orm/mongo-panache. Unless we create a common interface, which we didn't want to do. Not sure how to proceed here.

> I usually end my tests with Mockito.verifyNoMoreInterration() to check that all calls to my mocks has been verify, will this methods works ? Can you have a test for it ?

It should work, I can try it.

loicmathieu:
> Well Mockito can mock repos already so there's no need for that. It will not work if you pass a repo. I could add a type constraint for PanacheEntityBase but it would add a dependency to it, and we don't want that, unless we add a marker super interface to both orm/mongo entities. MockablePanacheEntity with no method. That should work.

Just add a sentence inside the documentation this should be OK for now.

Regarding mocking of PanacheQuery, we can add it later if needed.  As we don't have a supertype for both orm/mongo it will need to be created in specific extension ... not sure we want that now.

FroMage:
Just got this:

```
org.mockito.exceptions.verification.NoInteractionsWanted: 
No interactions wanted here:
-> at io.quarkus.it.mongodb.panache.MongodbPanacheResourceTest.testPanacheMocking(MongodbPanacheResourceTest.java:401)
But found this interaction on mock 'personEntity':
null
***
For your reference, here is the list of all invocations ([?] - means unverified).
1. null
2. null
3. null
4. null
5. [?]null
6. [?]null
7. [?]null
8. null
```

So I'm guessing my mocking is lacking something :)

FroMage:
OK that's fixed.

FroMage:
All working and with docs now, ready for review.

geoand:
Seems like imports need to be fixed:

```
2020-04-08T14:50:12.9163657Z [ERROR] Failed to execute goal net.revelc.code:impsort-maven-plugin:1.3.2:check (check-imports) on project quarkus-panache-common-deployment: Imports are not sorted in /home/runner/work/quarkus/quarkus/extensions/panache/panache-common/deployment/src/main/java/io/quarkus/panache/common/deployment/JandexUtil.java -> [Help 1]
2020-04-08T14:50:12.9183434Z org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal net.revelc.code:impsort-maven-plugin:1.3.2:check (check-imports) on project quarkus-panache-common-deployment: Imports are not sorted in /home/runner/work/quarkus/quarkus/extensions/panache/panache-common/deployment/src/main/java/io/quarkus/panache/common/deployment/JandexUtil.java
2020-04-08T14:50:12.9191363Z     at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:215)
```

FroMage:
Let's try again.

extensions/panache/panache-mock/deployment/src/main/java/io/quarkus/panache/mock/deployment/PanacheMockMethodCustomizer.java
```diff
@@ -0,0 +1,123 @@
...
+            i++;
+        }
+
+        mv.visitLdcInsn(method.parameters().size());
+        mv.visitTypeInsn(Opcodes.ANEWARRAY, "java/lang/Object");
+
+        i = 0;
+        for (org.jboss.jandex.Type paramType : method.parameters()) {
+            mv.visitInsn(Opcodes.DUP);
+            mv.visitLdcInsn(i);
```
stuartwdouglas, line 101: Double and long take up two local variable slots. Have you tested this methods methods that take double and long as params? Whenever you are doing anything with bytecode that involves primitives it is always a good idea to write explicit checks for long and double as they are a pain.

extensions/panache/panache-mock/runtime/src/main/java/io/quarkus/panache/mock/PanacheMock.java
stuartwdouglas: Is it possible that running user code will access this? e.g. If I mock a request then send a HTTP request will the HTTP request handler code access this map? If so it should be a CHM anyway to make sure that any modifications are visible to the other thread.

integration-tests/hibernate-orm-panache/src/test/java/io/quarkus/it/panache/PanacheMockingTest.java
```diff
@@ -0,0 +1,134 @@
...
+        Mockito.when(Person.count()).thenReturn(42l);
+        Assertions.assertEquals(42, Person.count());
+
+        Mockito.when(Person.count()).thenCallRealMethod();
+        Assertions.assertEquals(0, Person.count());
+
+        PanacheMock.verify(Person.class, Mockito.times(4)).count();
+
+        Person p = new Person();
+        Mockito.when(Person.findById(12l)).thenReturn(p);
```
stuartwdouglas, line 42: Can you add some tests with multiple long/double method parameters? It should show up the issue I brought up above. These tests don't show it because there is only a single parameter.

stuartwdouglas (changes requested):

extensions/panache/panache-common/deployment/src/main/java/io/quarkus/panache/common/deployment/PanacheRepositoryEnhancer.java
```diff
@@ -101,26 +117,102 @@ private String recursivelyFindEntityTypeFromClass(DotName clazz, DotName reposit
...
+                for (int i = 0; i < parameters.size(); i++) {
+                    mv.visitParameter(method.parameterName(i), 0 /* modifiers */);
+                }
+                mv.visitCode();
+                // this
+                mv.visitIntInsn(Opcodes.ALOAD, 0);
+                // each param
+                for (int i = 0; i < parameters.size(); i++) {
+                    org.jboss.jandex.Type paramType = parameters.get(i);
+                    mv.visitIntInsn(Opcodes.ALOAD, i + 1);
```
stuartwdouglas, line 183: I don't really understand the context here, but are you sure there will never be primitive params?

extensions/panache/panache-mock/deployment/src/main/java/io/quarkus/panache/mock/deployment/PanacheMockResourceProcessor.java
```diff
@@ -0,0 +1,17 @@
+package io.quarkus.panache.mock.deployment;
+
+import io.quarkus.deployment.annotations.BuildProducer;
+import io.quarkus.deployment.annotations.BuildStep;
+import io.quarkus.deployment.builditem.LaunchModeBuildItem;
+import io.quarkus.panache.common.deployment.PanacheMethodCustomizerBuildItem;
+import io.quarkus.runtime.LaunchMode;
+
+public final class PanacheMockResourceProcessor {
```
geoand, line 9: I wonder if we could / should ditch the full blown extension approach here, and just introduce a test module that utilizes the `TestBuildChainCustomizerProducer` we are introducing in #8491

docs/src/main/asciidoc/hibernate-orm-panache.adoc
```diff
@@ -661,6 +661,169 @@ public class PersonRepository implements PanacheRepositoryBase<Person,Integer> {
 }
 ----
 
+== Mocking
+
+=== Using the “entity pattern”
```
loicmathieu, line 666: The documentation now refers to it as the "active record pattern". And I didn't put quote in it so if you want your quote please update the other titles ;)

docs/src/main/asciidoc/hibernate-orm-panache.adoc
```diff
@@ -661,6 +661,169 @@ public class PersonRepository implements PanacheRepositoryBase<Person,Integer> {
...
+        return find("ORDER BY name").list();
+    }
+}
+----
+
+You can write your mocking test like this:
+
+[source,java]
+----
+@QuarkusTest
```
loicmathieu, line 691: We usually include the imports inside the doc.

docs/src/main/asciidoc/hibernate-orm-panache.adoc
```diff
@@ -661,6 +661,169 @@ public class PersonRepository implements PanacheRepositoryBase<Person,Integer> {
...
+        // Now let's change the return value
+        Mockito.when(Person.count()).thenReturn(42l);
+        Assertions.assertEquals(42, Person.count());
+
+        // Now let's call the original method
+        Mockito.when(Person.count()).thenCallRealMethod();
+        Assertions.assertEquals(0, Person.count());
+
+        // Check that we called it 4 times
+        PanacheMock.verify(Person.class, Mockito.times(4)).count();
```
loicmathieu, line 714: Maybe add a not that explain that `Mockito.verify()` will not works and all verification must be done via `PanacheMock`.

docs/src/main/asciidoc/hibernate-orm-panache.adoc
```diff
@@ -661,6 +661,169 @@ public class PersonRepository implements PanacheRepositoryBase<Person,Integer> {
...
+        PanacheMock.verify(Person.class).findOrdered();
+        PanacheMock.verify(Person.class, Mockito.atLeastOnce()).findById(Mockito.any());
+        PanacheMock.verifyNoMoreInteractions(Person.class);
+    }
+}
+----
+
+=== Using the “repository pattern”
+
+If you are using the “repository pattern” you can use Mockito directly, using the `quarkus-junit5-mockito` module.
```
loicmathieu, line 743: Is using `quarkus-junit5-mockito` mandatory ?
If so, maybe explain why in a few words.

docs/src/main/asciidoc/hibernate-orm-panache.adoc
```diff
@@ -661,6 +661,169 @@ public class PersonRepository implements PanacheRepositoryBase<Person,Integer> {
...
 ----
 
+== Mocking
+
+=== Using the “entity pattern”
+
+If you are using the “entity pattern” you cannot use Mockito directly as it does not support mocking static methods,
+but you can use the `quarkus-panache-mock` module which allows you to use Mockito to mock all provided static
+methods, including your own.
+
```
loicmathieu, line 671: Maybe add the needed `pom.xml` lines.

docs/src/main/asciidoc/hibernate-orm-panache.adoc
loicmathieu: And add the `pom.xml` snippet.

docs/src/main/asciidoc/hibernate-orm-panache.adoc
```diff
@@ -661,6 +661,169 @@ public class PersonRepository implements PanacheRepositoryBase<Person,Integer> {
...
+        return find("ORDER BY name").list();
+    }
+}
+----
+
+You can write your mocking test like this:
+
+[source,java]
+----
+@QuarkusTest
```
loicmathieu, line 776: We usually include the imports inside the doc.

docs/src/main/asciidoc/hibernate-orm-panache.adoc
```diff
@@ -661,6 +661,169 @@ public class PersonRepository implements PanacheRepositoryBase<Person,Integer> {
...
+public class PanacheFunctionalityTest {
+
+    @Test
```
loicmathieu, line 694: This one is not disabled on native but the repository variant is ?
As injection didn't works in native mode you should add a note that explain it

docs/src/main/asciidoc/hibernate-orm-panache.adoc
```diff
@@ -661,6 +661,169 @@ public class PersonRepository implements PanacheRepositoryBase<Person,Integer> {
...
+        Assertions.assertSame(p, Person.findById(12l));
+        Assertions.assertNull(Person.findById(42l));
+
+        // Mock throwing
+        Mockito.when(Person.findById(12l)).thenThrow(new WebApplicationException());
+        try {
+            Person.findById(12l);
+            Assertions.fail();
+        } catch (WebApplicationException x) {
+        }
```
loicmathieu, line 728: I usually use this construct that is clearer :
```
assertThrows(WebApplicationException.class, () -> Person.findById(12l));
```

loicmathieu (changes requested):
Overall looks like a very good addition to Panache ;)

I only review the documentation parts, I didn't duplicate my comments to the mongodb-panache guide but they are the same.

I let other more knowledgeable than me on the implementation details as they are already on it ;)

extensions/panache/panache-mock/deployment/src/main/java/io/quarkus/panache/mock/deployment/PanacheMockMethodCustomizer.java
FroMage: So, we don't have any method in the superclasses in question that we generate bridges for (`PanacheEntityBase` and `PanacheRepositoryBase`) so it doesn't happen. And I can't test this by adding a method to these classes, and no other classes have bridge methods generated.

So in general, sure, but I couldn't test this if I supported it. I could add a guard to make sure we never add such methods.

integration-tests/hibernate-orm-panache/src/test/java/io/quarkus/it/panache/PanacheMockingTest.java
FroMage: I can't, see above.

FroMage: Actually I can make user methods do this. Sorry.

docs/src/main/asciidoc/hibernate-orm-panache.adoc
FroMage: We don't in any other example in this guide.

docs/src/main/asciidoc/hibernate-orm-panache.adoc
FroMage: Grr, I much prefer "entity pattern" over "active record pattern" because those are not active records.

docs/src/main/asciidoc/hibernate-orm-panache.adoc
FroMage: ok

docs/src/main/asciidoc/hibernate-orm-panache.adoc
FroMage: ok

docs/src/main/asciidoc/hibernate-orm-panache.adoc
FroMage: ok

docs/src/main/asciidoc/hibernate-orm-panache.adoc
FroMage: this is not a native test, I've removed that annotation

docs/src/main/asciidoc/hibernate-orm-panache.adoc
FroMage: good point.

FroMage:
All fixed, rebased, tested primitive params, updated docs, turned from extension into test module with custom build chain extension.

extensions/panache/panache-mock/deployment/src/main/java/io/quarkus/panache/mock/deployment/PanacheMockResourceProcessor.java
FroMage: Good idea, done.

extensions/panache/panache-common/deployment/src/main/java/io/quarkus/panache/common/deployment/PanacheRepositoryEnhancer.java
FroMage: ATM there are none. I've added a check if it ever happens.

extensions/panache/panache-mock/deployment/src/main/java/io/quarkus/panache/mock/deployment/PanacheMockMethodCustomizer.java
FroMage: You were right, I've implemented and tested it.

docs/src/main/asciidoc/hibernate-orm-panache.adoc
loicmathieu: Yes, but other guides do.
And I was advised to always include the imports inside the guide.
And here it is helpful to see what comes from Mockito and what comes from PanacheMock

docs/src/main/asciidoc/getting-started-testing.adoc
```diff
@@ -525,6 +525,10 @@ public class MockGreetingServiceTest {
 ----
 <1> Since we configured `greetingService` as a mock, the `GreetingResource` which uses the `GreetingService` bean, we get the mocked response instead of the response of the regular `GreetingService` bean
 
+=== Mocking with Panache
+
+If you are using the `quarkus-hibernate-orm-panache` or `quarkus-mongodb-panache` extensions with the “entity pattern” you can use the `quarkus-panache-mock` extension as documented in the link:hibernate-orm-panache#mocking[Hibernate ORM with Panache Mocking] and link:mongodb-panache#mocking[MongoDB with Panache Mocking] documentation.
```
loicmathieu, line 530: Sorry, I missed this one:
entity pattern -> active record pattern

docs/src/main/asciidoc/hibernate-orm-panache.adoc
```diff
@@ -686,6 +686,184 @@ public class PersonRepository implements PanacheRepositoryBase<Person,Integer> {
...
+
+=== Using the active record pattern
+
+If you are using the active record pattern you cannot use Mockito directly as it does not support mocking static methods,
+but you can use the `quarkus-panache-mock` module which allows you to use Mockito to mock all provided static
+methods, including your own.
+
+Add this dependency to your `pom.xml`:
+
+[source,java]
```
loicmathieu, line 699: Should be `[source,xml]`

docs/src/main/asciidoc/mongodb-panache.adoc
```diff
@@ -867,6 +867,181 @@ public Multi<ReactivePerson> streamPersons() {
...
+
+=== Using the active-record pattern
+
+If you are using the active-record pattern you cannot use Mockito directly as it does not support mocking static methods,
+but you can use the `quarkus-panache-mock` module which allows you to use Mockito to mock all provided static
+methods, including your own.
+
+Add this dependency to your `pom.xml`:
+
+[source,java]
```
loicmathieu, line 880: Should be `[source,xml]`

extensions/panache/panache-mock/runtime/src/main/java/io/quarkus/panache/mock/PanacheMock.java
FroMage: Fixed

FroMage:
Rebased and fixed @loicmathieu 's comments. @stuartwdouglas and @gsmet can you re-review?

extensions/panache/panache-mock/deployment/src/main/java/io/quarkus/panache/mock/deployment/PanacheMockResourceProcessor.java
geoand: Excellent

extensions/panache/panache-mock/pom.xml
```diff
@@ -0,0 +1,37 @@
...
+        <artifactId>quarkus-build-parent</artifactId>
+        <groupId>io.quarkus</groupId>
+        <version>999-SNAPSHOT</version>
+        <relativePath>../../../build-parent/pom.xml</relativePath>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>quarkus-panache-mock</artifactId>
+    <name>Quarkus - Panache - Mock - Runtime</name>
+    <description>An opinionated approach to make Hibernate as easy as possible</description>
```
geoand, line 14: This probably need to be updated :)

geoand (approved):
Awesome work!

Testing part looks good, can't comment about the rest since I don't know much about Panache

stuartwdouglas (approved):
My concerns have been addressed

geoand:
I believe this is something we should include in 1.4. @gsmet WDYT?

loicmathieu (approved):
LGTM

extensions/panache/panache-mock/pom.xml
FroMage: Good point, fixed, thanks!

FroMage:
So that leaves @gsmet

FroMage:
The tests fail for vertx-web, but I don't see how that can be related to this PR. @cescoffier are you aware of current/intermitent test failures in vertx-web?

cescoffier:
Looks like a race condition or something clearly wrong unrelated to this PR:

```
2020-04-16T13:06:58.3356832Z 2020-04-16 13:06:58,109 ERROR [io.qua.ver.htt.run.QuarkusErrorHandler] (vert.x-eventloop-thread-0) HTTP Request to /hello failed, error id: faca1667-b810-4e7e-88b6-8e522e4f8169-1: java.lang.NullPointerException
2020-04-16T13:06:58.3437894Z 	at io.quarkus.vertx.web.base.RouteBaseTest_RouteBaseTest$SimpleBean_RouteHandler_hello_254a2a71f98e407c82cb72702a943a981ffe2f58.invoke(RouteBaseTest_RouteBaseTest$SimpleBean_RouteHandler_hello_254a2a71f98e407c82cb72702a943a981ffe2f58.zig:84)
2020-04-16T13:06:58.3465221Z 	at io.quarkus.vertx.web.runtime.RouteHandler.handle(RouteHandler.java:44)
2020-04-16T13:06:58.3553001Z 	at io.quarkus.vertx.web.runtime.RouteHandler.handle(RouteHandler.java:15)
2020-04-16T13:06:58.3609712Z 	at io.vertx.ext.web.impl.RouteState.handleContext(RouteState.java:1034)
```

cescoffier:
Ah ah, I believe it's the race condition addressed by @mkouba !

FroMage:
Already fixed in master you mean?

mkouba:
AFAIK it's not merged yet

mkouba:
@FroMage @cescoffier yes, this should be addressed by https://github.com/quarkusio/quarkus/pull/8579. Unfortunately, I can't "force merge" anymore.

FroMage:
Rebased on master to see if the test failures are gone, they're definitely not related to this PR.

FroMage:
```
Caused by: java.lang.ClassNotFoundException: org.eclipse.jetty.alpn.ALPN$Provider
```

I declare CI to be dead.

geoand:

FroMage:
@gsmet just merged it. Not sure if it's too late to make 1.4 or wait for 1.5?

gsmet:
If it changed only the test code, I would have said 1.4 but I can see it changes quite a lot of things in the regular code too so I'm leaning towards 1.5.

FroMage:
Yes, and 1.5 is not far off and let's be honest, nobody actually cares about mocking, as we've uncovered in this PR that nobody is mocking panache repositories ;)

loicmathieu:
>  we've uncovered in this PR that nobody is mocking panache repositories ;)

For the record, I mock panache repositories without any issue since the beginning !

FroMage:
That can't be true: this PR uncovered so many bugs in the generated code for repositories that prevented them from being mocked!

Or you've never mocked any method with generics or primitives in their signature ;) `findById` didn't work, neither did `count()`.

FroMage:
When I tried mocking repos, Mockito just threw exceptions when it loaded the generated classes with bytebuddy.

loicmathieu:
I mocked all my repos on raw JUnit tests with Mockito without any issue, even the findById methods. I think Daniel Petisme also (he report it during it's effort on JHipster).
But we both uses raw JUnit tests without QuarkusTest it should be why it worked.

Issue #7996, opened 2020-03-19
Quarkus 1.3.0.Final gradle does not start with Undertow Servlet


tk303:
**Describe the bug**
When I include Undertow Servlet in a Gradle project, quarkus application does not start.

**Expected behavior**
Quarkus app should start without throwing an error

**Actual behavior**
(Describe the actual behavior clearly and concisely.)

**To Reproduce**
1.  Go to https://code.quarkus.io/ 
2.  Change Build Tool to: Gradle
3.  Add Undertow Servlet as a dependency.
4. Generate project
5. Try running with quarkusDev build task.

Project throws java.lang.ExceptionInInitializerError

The same configuration when used with maven starts fine.

**Environment (please complete the following information):**
 - Output of `uname -a` or `ver`: 
 - Output of `java -version`: 
 - GraalVM version (if different from Java): 
 - Quarkus version or git rev: 
 - Build tool (ie. output of `mvnw --version` or `gradlew --version`): 

**Additional context**
(Add any other context about the problem here.)

gastaldi:
I see this stacktrace, is the same you see? 

```
2020-03-19 17:04:51,293 INFO  [org.jbo.threads] (main) JBoss Threads version 3.0.1.Final
2020-03-19 17:04:51,934 ERROR [io.qua.dev.DevModeMain] (main) Failed to start Quarkus: java.lang.ExceptionInInitializerError
        at java.base/java.lang.Class.forName0(Native Method)
        at java.base/java.lang.Class.forName(Class.java:398)
        at io.quarkus.runner.bootstrap.StartupActionImpl.run(StartupActionImpl.java:80)
        at io.quarkus.dev.IsolatedDevModeMain.firstStart(IsolatedDevModeMain.java:60)
        at io.quarkus.dev.IsolatedDevModeMain.accept(IsolatedDevModeMain.java:236)
        at io.quarkus.dev.IsolatedDevModeMain.accept(IsolatedDevModeMain.java:39)
        at io.quarkus.bootstrap.app.CuratedApplication.runInCl(CuratedApplication.java:131)
        at io.quarkus.bootstrap.app.CuratedApplication.runInAugmentClassLoader(CuratedApplication.java:84)
        at io.quarkus.dev.DevModeMain.start(DevModeMain.java:108)
        at io.quarkus.dev.DevModeMain.main(DevModeMain.java:47)
Caused by: java.lang.RuntimeException: Failed to start quarkus
[...]

PR #8337, opened 2020-04-01
Gradle: Refining the fix for #7996

roguexz:
The original fix for #7996 skipped a service provider if it showed up as part of the current classloader. This was done on the assumption that the parent classloader would contain the resource. But what happens if the parent classloader does not contain the reference? This commit attempts to refine the fix further by checking with the parent classloader, before banning the service provider implementation.


roguexz:
@gastaldi what do you think?

gastaldi:
I'll let Stuart Douglas review this one ;)

Em Qua, 1 de abr de 2020 15:55, Harsha Ramesh <notifications@github.com>
escreveu:

stuartwdouglas:
What is the advantage of this? If the current class loader defines the service then even if the parent does not contain a copy of the Service no good can come of attempting to load from it. Best case is a noop, worst case the parent has an impl but not the service itself so it fails with NCDFE?

roguexz:
Ok. I guess I will need to understand more about the actual issue, because right now, I am seeing issues with services being loaded from other projects. Could be a configuration issue, but not yet sure.

Validating the original test case has been challenging, not sure why, but yeah.

roguexz:
Closing this. Will come back to this one later. Found another issue with the Gradle plugin. Will raise a fix for that shortly.

PR #8340, opened 2020-04-02
Gradle: Search for build script dependencies in the parent projects as well

roguexz:
It is possible to construct a multi-module project where the build script dependencies are defined in a parent project. This avoids having to include them in every project explicitly.

This is applicable when you are applying plugin via the legacy methodology.

Not sure how many teams run in to this problem, but I found this to be the reason why I wasn't able to use `startQuarkusDev`.


gastaldi:
I wouldn't expect the legacy method to longer work with the latest changes to the plugin, but it looks OK. 

I'll let @aloubyansky  review before merging this

aloubyansky (approved):

PR #8341, opened 2020-04-02
Log downloads to the console

### LINKED ISSUE #4460: Bootstrap resolver should log if it is accessing the network
When running tests or building for the first time the bootstrap resolver might need to download -deployment artifacts and their dependencies. If the network is slow this can appear that tests or the build are hanging. If we need to hit the network we should log a message and update the user on our progress in a similar way to a normal maven build,

stuartwdouglas:
Fixes #4460

### DIFF
--- a/independent-projects/bootstrap/core/src/main/java/io/quarkus/bootstrap/resolver/maven/MavenArtifactResolver.java
+++ b/independent-projects/bootstrap/core/src/main/java/io/quarkus/bootstrap/resolver/maven/MavenArtifactResolver.java
@@ -17,6 +17,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import org.apache.maven.cli.transfer.ConsoleMavenTransferListener;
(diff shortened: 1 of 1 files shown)


independent-projects/bootstrap/core/src/main/java/io/quarkus/bootstrap/resolver/maven/MavenArtifactResolver.java
```diff
@@ -145,6 +146,7 @@ private MavenArtifactResolver(Builder builder) throws AppModelResolverException
         if (builder.offline != null) {
             newSession.setOffline(builder.offline);
         }
+        newSession.setTransferListener(new ConsoleMavenTransferListener(System.out, true));
```
aloubyansky, line 149: `newSession` here might be wrapping one that is initialized by the actual Maven process, in which case it most probably already has the transfer listener in place. Perhaps it'd be better to check whether it's already present and if not initialize it that way.

aloubyansky (approved):

Issue #6814, opened 2020-01-27
Support the Concept of @ActiveProfile from Spring CDI


sherl0cks:
**Description**
[Spring Profiles](https://www.baeldung.com/spring-profiles) are very different than [Quarkus Profiles](https://quarkus.io/guides/config#configuration-profiles). The former is focused on decomposing your DI config into subsets, and the later is focused on environment specific config properties. 

The most useful part of Spring Profiles is the ability to annotate a test with `@ActiveProfile("SomeProfile")`, which will instruct the DI system to only spin up beans with that profile. This is very helpful for tests written against an API, e.g. JPA. I may have two profiles "postgres" and "hsql", and depending on the test, I will want to spin up one or the other, but not both. Today's `@QuarkusTest` will spin up the entire CDI config and there is no way to spin up a subset. User must use `@ApplicationScoped` for lazy evaluation.

You can kind of assemble this "Profile" concept using CDI qualifiers like `@Named` and `@Typed` and `@Alternative`, but it would be really nice to have `@Profile` and `@ActiveProfile`as first order concepts.

See discussion in https://quarkusio.zulipchat.com/#narrow/stream/187030-users/topic/configuring.20Mocks.20in.20.40QuarkusTest

**Implementation ideas**
It would be nice to have this added as an ArC extensions like the CDI oriented convenience features in ArC. The Spring DI extension would work as well.

geoand:
cc @mkouba @manovotn

mkouba:
I like the idea but there are some problems we would need to resolve first:

1. Our DI is build-time oriented, ie. **injection points are resolved/validate at build time**. And the applications is only built/started once for the whole test suite. In other words, currently we can't activate a different "DI profile" per `@QuarkusTest`.
2. We would have to (1) merge this feature with "Configuration Profiles", otherwise users might be confused or (2) use a different name.

@dmlloyd @emmanuelbernard Any ideas?

manovotn:
I've see the Zulip discussion yesterday.
In pure CDI you usually deal with this via alternatives enabled only in test environments (which basically allows you to mock whichever beans you want to). Here, you would need a combination of `@DefaultBean` and then another impl in the test itself. For smaller apps that would work but if you have something huge, it can see how this becomes a pain.

> Our DI is build-time oriented, ie. injection points are resolved/validate at build time.
I agree this is the problematic part.

[...]

PR #8343, opened 2020-04-02
Support having a bean enabled only when a build time profile is enabled or disabled

geoand:


geoand:
@manovotn @mkouba I would like to hear your thoughts on this one... 

What I am basically going for here is to allow users to configure different beans on different profiles.
I am also thinking that a similar mechanism could be used for build time properties.

I haven't written any docs or javadoc yet as I wanted to get your opinion first.

mkouba:
Hm, I think that this feature could be implemented by alternatives + `quarkus.arc.selected-alternatives` + configuration profiles, i.e. `%dev.quarkus.arc.selected-alternatives=MyBean`. TBH I'm not super excited about adding redundant stuff to the ArC core.

geoand:
@mkouba yeah, but wouldn't you have to do that for every bean you with to configure that way?
Say I have 10 beans that I only want in `dev`, I would have to add those 10 properties.

Another thing I have in mind with this feature is that users could write modules that provide auto-configuration for libraries, that user applications would then include. In that model, but just including such a module, everything gets auto-configured without the user having to do anything.

mkouba:
>  I would have to add those 10 properties.

No. You would have to mark those beans `@Alternative` and add one property to list all beans, or the matching the package, etc.

> Another thing I have in mind with this feature is that users could write modules that provide auto-configuration for libraries...

I'm not sure I understand. Could you try to describe a simple example?

geoand:
> > I would have to add those 10 properties.
> 
> No. You would have to mark those beans `@Alternative` and add one property to list all beans, or the matching the package, etc.
> 

But this is much harder to manage isn't it? Also, the annotation has a straighforward meaning to the user.

> > Another thing I have in mind with this feature is that users could write modules that provide auto-configuration for libraries...
> 
> I'm not sure I understand. Could you try to describe a simple example?

Let's say some team wants to configure Tracing (or maybe Jackson, or Datasources, or...) in the same way for all the Quarkus applications (they don't want to repeat the same setup for all the time and they have found a setup that matches the team's needs).
So they create a jar that contains a bunch of CDI producers that configure the proper beans for each environment.

Does that make sense?

mkouba:
I think I get the point. On the other hand, `@WhenBuildTimeProfile("dev")` could be translated to `@Alternative` + enable the class if in the dev mode. So in theory it could be implemented by means of  `AnnotationTransformer` + modify the existing `io.quarkus.arc.processor.AlternativePriorities`. I.e. simply add `@Alternative` to a bean annotated with `@WhenBuildTimeProfile("dev")` and in the `AlternativePriorities.compute()` method return some value if `LaunchMode.DEVELOPMENT`. WDYT?

geoand:
@mkouba thanks for the tip. I will look into it and come back with concrete info :)

manovotn:
@geoand Is this same as https://github.com/quarkusio/quarkus/issues/6814?

> On the other hand, @WhenBuildTimeProfile("dev") could be translated to @Alternative + enable the class if in the dev mode. 

That's smart. But I am not sure it fits the scenario due to original beans not being ruled out (producers still count, validation still happens)?

geoand:
> @geoand Is this same as #6814?
> 
> > On the other hand, @WhenBuildTimeProfile("dev") could be translated to @alternative + enable the class if in the dev mode.
> 
> That's smart. But I am not sure it fits the scenario due to original beans not being ruled out (producers still count, validation still happens)?
> 
> EDIT: actually nevermind, even this solution has to keep original beans

I have a slightly different idea, along the same lines (the suggestion is great and as I have said in the past, the `AnnotationTransformer` is a stroke of genius!)

manovotn:
Looking more closely at your tests, I see what you mean.
Your use case if more about your application needing a bean `Foo` and then you have two or more `@WhenBuildTimeProfile` beans for it, each covering different profile.

But that doesn't cover a case where you actually already have `Foo` declared in your app and then you declare `Foo` producer with `@WhenBuildTimeProfile("test")` - if you run tests, you then get ambiguous resolution exception (unless I am missing something). This can also be solved via the alternative + priority approach but with the drawback of original bean still being discovered and validated. That would be fair enough I guess,

+1 for this feature, I like it. I think it will also cover the issue I linked since it allows pretty much that (with little tweaks).

geoand:
> But that doesn't cover a case where you actually already have `Foo` declared in your app and then you declare `Foo` producer with `@WhenBuildTimeProfile("test")` - if you run tests, you then get ambiguous resolution exception (unless I am missing something). This can also be solved via the alternative + priority approach but with the drawback of original bean still being discovered and validated. That would be fair enough I guess,
> 

Yeah, I don't want this annotation to deal with ambiguity - that's up to the user to fix.

geoand:
New implementation is up!

I takes the suggestion to use the annotation transformer but turns it on it head :). 
Basically annotations are now removed when the profile doesn't match.

Also I am basing the commit on #8359 (for the simple reason that it has a method I want)

mkouba:
> That's smart. But I am not sure it fits the scenario due to original beans not being ruled out (producers still count, validation still happens)?

Yes, but this is only the case when an alternative "overrides" another bean. However, alternatives do not have to "override" anything, i.e. you can have `@Alternative Foo` that is the only bean that satisfies `@Inject Foo`... and if it's not selected -> unsatisfied dependency. Just for the record ;-)

manovotn:
I am aware of that. I was just considering the case where "override" happens.

geoand:
@mkouba, @manovotn  what do you think of the current approach compared to the `@Alternative` approach you suggested?

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/WhenBuildTimeProfileProcessor.java
```diff
@@ -0,0 +1,90 @@
...
+
+                String profileOnInstance = whenBuildTimeProfileInstance.value().asString();
+                if (profileOnInstance.equals(ProfileManager.getActiveProfile())) {
+                    return;
+                }
+
+                if (ctx.isMethod() || ctx.isField()) {
+                    ctx.transform().remove(ann -> DotNames.PRODUCES.equals(ann.name())).done();
+                } else if (ctx.isClass()) {
+                    ctx.transform().remove(ann -> scopes.isScopeDeclaredOn(ann.target().asClass())).done();
```
mkouba, line 67: Hm, this should remove any annotation on a class that is annotated with a scope. Is that intentional? If so, then this could be replaced with something like:
```java
if (scopes.isScopeIn(ctx.getAnnotations())) {
  ctx.transform().removeAll().done();
}
```

geoand: The intention was to remove only scope annotations, but now I see that I pulled the trigger a little to quickly on it :)

mkouba: Actually, you should also remove stereotypes. And additional bean defining annotations may declare a default scope too...

geoand: So should we just go with `removeAll`? I wanted to avoid that if possible.

manovotn: I don't think just stereotypes will do.
If you don't want a class picked up as a bean, you'd need to remove any bean defining annotation from it. That can be scope, qualifier, stereotype, interceptor binding. See [spec](https://docs.jboss.org/cdi/spec/2.0/cdi-spec.html#bean_defining_annotations). Note that in Quarkus we can also have [additional bean defining annotation](https://quarkus.io/guides/cdi-reference#additional-bean-defining-annotations).

geoand: So it seems that removing all the annotations is innevitable with this approach, right?

manovotn: Probably... but at that point it might be easier to instead just temper with Alternative and priority.
BTW how many build time profiles are there? This approach will make a bean exclusively available in one of them, what if you want the bean to be available in two or more profiles but not in third?

geoand: For now I just pushed `removeAll`. I'll give the Alternatives approach a go later on.

Very good question about the priorities! There is always a single active build profile, so maybe we should also introduce the `WhenNotBuildTimeProfile`. How does that sound?

manovotn: @geoand sounds good. Or you could make `WhenBuildTimeProfile` a repeating annotation.

geoand: What is the usual CDI strategy in "similar" cases? 
Best try and be consistent if possible.

manovotn: I don't see similarities that we could stick to here, might just be lack of caffeine though :)
Maybe @mkouba has some ideas?

geoand:
@mkouba @manovotn looking at what you propose with `@Alternative` I don't see how the following scenario could be implemented:

```
    @WhenBuildTimeProfile("dev")
    @Singleton
    static class PongBean {

        String pong() {
            return "pong from dev";
        }
    }

   @Singleton
    static class Producer {

        @Produces
        @DefaultBean
        PongBean defaultPongBean() {
            return new PongBean() {
                @Override
                String pong() {
                    return "pong";
                }
            };
        }
    }

    @WhenBuildTimeProfile("other")
    static class OtherProducer {

        @Produces
        PongBean otherPongBean() {
            return new PongBean() {
                @Override
                String pong() {
                    return "pong from other";
                }
            };
        }
    }
```

With the code above, I expect that:
*  `dev` profile is active, the bean will come from the `PongBean CLASS bean`
*  `other` profile is active, the bean will come from `OtherProducer#otherPongBean PRODUCER METHOD bean`
* In any other case, the bean will come from `Producer#defaultPongBean PRODUCER METHOD bean`

Am I missing something?

P.S. The approach I currently have in this PR (which removes annotations) does cover this case.

mkouba:
@geoand This should work just fine if you add `@Alternative` to beans annotated with `@WhenBuildTimeProfile` that should not be activated. E.g. in case of the `dev` profile there will be 3 beans matching the `PongBean` type:
1. PongBean class with added `@Alternative` and `@Priority` -> enabled
2. `Producer.defaultPongBean()` is default, no change
3. `OtherProducer.pong()`  with added `@Alternative`, no priority -> disabled

And when resolving ambiguity the default beans are removed and so 1. would win.

geoand:
Aha! Now I understand how `@Alternative` is meant to work with `@Priority`.

Thanks a lot @mkouba

geoand:
@mkouba @manovotn I updated the PR to use the `@Alternative` approach and also added support for `@WhenNotBuildProfile` annotation.

WDYT?

mkouba (commented):
Looks really good! One last thing - the name of `@WhenBuildTimeProfile` is quite long. What about `@BuildProfile`?

geoand:
As you probably remember, I am horrible at naming, so I'll go with whatever you and @manovotn thing is best :)

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/BuildTimeProfileProcessor.java
```diff
@@ -0,0 +1,115 @@
...
+                AnnotationInstance whenBuildTimeProfileInstance = find(ctx.getAnnotations(), WHEN_BUILD_TIME_PROFILE);
+                if (whenBuildTimeProfileInstance == null) {
+                    return;
+                }
+
+                Transformation transform = ctx.transform();
+
+                String profileOnInstance = whenBuildTimeProfileInstance.value().asString();
+                if (profileOnInstance.equals(ProfileManager.getActiveProfile())) {
+                    transform.add(DotNames.ALTERNATIVE_PRIORITY,
```
manovotn, line 58: Recently we added the ability to enable alternatives via config file and those also use `Integer.MAX_VALUE`.
So if a user has an alternative bean that is enabled that way and then adds another bean just for one given profile, he will actually end up with two alternatives with maximum priority (which I think should blow up). Not sure if this corner case is something we want to address right away, just writing it down :)

docs/src/main/asciidoc/cdi-reference.adoc
```diff
@@ -426,6 +426,57 @@ public class CustomTracerConfiguration {
 `@DefaultBean` allows extensions (or any other code for that matter) to provide defaults while backing off if beans of that type are supplied in any
 way Quarkus supports.
 
+=== Enabling Beans for Quarkus Build Profile
+
+Quarkus adds a capability that CDI currently does not support which is to conditionally enable a bean when a Quarkus build-time profile is enabled,
+via the `@WhenBuildTimeProfile` and ``@WhenNotBuildTimeProfile`` annotations.
```
manovotn, line 432: Accidental double grave accent(`) around `WhenNotBuildTimeProfile`?

manovotn (commented):
Looks good and simple!
As for names, I don't even mind these but if you want some suggestion, I can think of `EnableInProfile`/`DisableInProfile` or more verbosely `EnableBeanInProfile`.

geoand:
@manovotn although I like your suggestions, I really want to have `build` in the name, so uses will be under no illusion that they can change the behavior at runtime.

How about `@EnableOnBuildProfile` and `@DisableOnBuildProfile`? @mkouba WDYT?

docs/src/main/asciidoc/cdi-reference.adoc
geoand: Good catch, thanks!

mkouba:
I agree that the annotation name should contain `build` and `profile`. I'm +0 for Enable/Disable. Another options could be `@ActiveOnBuildProfile("dev")` or `@IfBuildProfile("dev")/@UnlessBuildProfile("dev")`...

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/BuildTimeProfileProcessor.java
mkouba: Hm, yes that would blow up with ambiguous dependency. But I'd say that it's more like a corner case and maybe we could add a note to the docs? Or even use `Intereg.MAX_VALUE - 1` ;-).

geoand: How about goign with `Intereg.MAX_VALUE - 1` for now and revisiting if this every becomes a problem?
Seems like an extreme corner case to me.

geoand:
> I agree that the annotation name should contain `build` and `profile`. I'm +0 for Enable/Disable. Another options could be `@ActiveOnBuildProfile("dev")` or `@IfBuildProfile("dev")/@UnlessBuildProfile("dev")`...

Oh, now these I think really hit the sweetspot!

extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/BuildTimeProfileProcessor.java
manovotn: +1 to `Intereg.MAX_VALUE - 1`, it just occured to me so I thought I'd mention it :)

manovotn:
> @IfBuildProfile("dev")/@UnlessBuildProfile("dev")...

Those are good!

geoand:
PR updated with the new names and the commits squashed.

manovotn (approved):
Nice work, I really like this feature! :)

mkouba (approved):
Hm, what happens if a users adds both annotations? I know it would be a super weird use case but maybe we should detect this problem? 

Another idea that comes to my mind is that `@IfBuildProfile` could be repeatable so that it's possible to specify multiple profiles. Again, it's a corner case because users probably don't use custom profiles very often and the built-in profiles are covered.

So just some ideas for later ;-).

geoand:
Valid points @mkouba! @manovotn had already mentioned the repeatable annotations part and I think it does make sense to add it later.

As for having both annotations together... I honestly dunno :). If people start doing that we should probably start being more strict on what annotations we accept. 

Thanks for the help both of you :)

PR #8355, opened 2020-04-02
Fix a few things in the new Keycloak Admin Client extension

gsmet:


bom/deployment/pom.xml
```diff
@@ -198,6 +198,16 @@
...
+            <dependency>
+                <groupId>io.quarkus</groupId>
+                <artifactId>quarkus-keycloak-admin-client-deployment</artifactId>
+                <version>${project.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>io.quarkus</groupId>
+                <artifactId>quarkus-keycloak-authorization-deployment</artifactId>
+                <version>${project.version}</version>
+            </dependency>
```
sberyozkin, line 210: Thanks, my fault I missed it

extensions/keycloak-admin-client/runtime/src/main/resources/META-INF/quarkus-extension.yaml
```diff
@@ -2,10 +2,9 @@
 name: "Keycloak Admin Client"
 metadata:
   keywords:
-  - "oauth2"
-  - "openid-connect"
```
sberyozkin, line 6: IMHO keeping `openid-connect` makes sense. We only this extension to help Keycloak users control Keycloak dynamically which is what they do with this admin client - but Keycloak is part of the OIDC setups. So if a KC user enters `openid-connect` this extension should be listed too

gsmet: @sberyozkin done

sberyozkin: @gsmet thanks

sberyozkin (approved):

PR #8356, opened 2020-04-02
getting-started guide - small fix in TestHTTPResource related info

rsvoboda:
Small fix in TestHTTPResource related info

`inject the URL of the test into a field` doesn't sound right

It's URL of the tested application or resource, depending on the value of the annotation

### DIFF
--- a/docs/src/main/asciidoc/getting-started.adoc
+++ b/docs/src/main/asciidoc/getting-started.adoc
@@ -411,7 +411,7 @@ You can also run the test from your IDE directly (be sure you stopped the applic
 
 By default, tests will run on port `8081` so as not to conflict with the running application. We automatically
 configure RestAssured to use this port. If you want to use a different client you should use the `@TestHTTPResource`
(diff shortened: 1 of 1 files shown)


docs/src/main/asciidoc/getting-started.adoc
```diff
@@ -411,7 +411,7 @@ You can also run the test from your IDE directly (be sure you stopped the applic
 
 By default, tests will run on port `8081` so as not to conflict with the running application. We automatically
 configure RestAssured to use this port. If you want to use a different client you should use the `@TestHTTPResource`
-annotation to directly inject the URL of the test into a field on the test class. This field can be of the type
+annotation to directly inject the URL into a field on the test class. This field can be of the type
```
gastaldi, line 414: Still sounds like something is missing to me  Maybe URL of the tested application into a field ?

rsvoboda: PR updated, resolving conversation

gastaldi (approved):

PR #8359, opened 2020-04-02
Automatically add @Produces to methods annotated with scope or qualifier

### LINKED ISSUE #4831: HealthChecks and CDI Produces
**Description**

Hi I have developed a Health Check using CDI so the code looks like:

```
@ApplicationScoped
public class HealthChecks {
    @Readiness
    @Produces
    public HealthCheck healthy() {
        return HealthStatus.up("name");
    }
}
```
I have read in this document https://github.com/eclipse/microprofile-health/blob/51c0421ee9ab68742126e98d52ff20542273d2fe/spec/src/main/asciidoc/java-api.adoc that  you can define health checks using CDI producers. 

I don't know exactly if this is mandatory or not regarding the MicroProfile Health Checks spec to have to use @Produces or not in case of Health Checks and CDI, but maybe we could skip this annotation as we are skipping for `@Inject`.

**Implementation ideas**

I think that skipping the `@Produces` annotation should only affect in case of HealthCheck is the return type.

mkouba:
- resolves #4831

### DIFF
--- a/docs/src/main/asciidoc/cdi-reference.adoc
+++ b/docs/src/main/asciidoc/cdi-reference.adoc
@@ -282,7 +282,7 @@ NOTE: An event with qualifier `@Initialized(RequestScoped.class)` is fired when 
 
 === Qualified Injected Fields
 
-In CDI, if you declare a field injection point you need to use `@Inject` and optionally a set of qualifiers:
+In CDI, if you declare a field injection point you need to use `@Inject` and optionally a set of qualifiers.
  
 [source,java]
 ----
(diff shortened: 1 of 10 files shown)


geoand (approved):
I really like anything that reduces boilerplate, so this is +1 from me!

manovotn (commented):
@mkouba looks good, I am just thinking - could you use stereotype to define a scope for producer method?

mkouba:
@manovotn Another good point!

manovotn (approved):

geoand:
@mkouba mind force pushing the PR? I would like to have CI run properly (some infrastructure issue seems to have prevented the JDK tests from running).

Issue #8357, opened 2020-04-02
Missing Liquibase properties in native image


Rupert-RR:
**Describe the bug**
When using Quarkus with a Postgre database (but this may apply to other database types also) and Liquibase, if an (as yet unapplied) database migration uses the `autoIncrement` property on a column the application fails to start with an exception ` java.lang.IllegalStateException: liquibase.exception.UnexpectedLiquibaseException: Property [autoIncrement] was not found for object type [liquibase.datatype.core.IntType]`.

**Expected behavior**
The application should apply the migration and startup without error.

**Actual behavior**
The application crashes while trying to apply the migration.

**To Reproduce**
1. Create and download a new application from the [Start coding](https://code.quarkus.io) page of the Quarkus web site, including the JDBC Driver - PostgreSQL and the Liquibase extensions;
2. Configure the given `application.properties` with the a Postgre datasource and Liquibase migration applied at startup;
3. Add the simple migration given as an [example](https://quarkus.io/guides/liquibase), adding the `autoIncrement="true"` attribute on the `ID` column to `resources/db/changeLog.xml`;
4. Run the application using `./mvnw compile quarkus:dev` and notice that it launches correctly, applying the migration;
5. Stop the application and delete all the tables it created from your Postgre instance;
6. Build the application as a native image using `./mvnw package -Pnative`;
7. Run the native image application by `cd ./target` and ` ./code-with-quarkus-1.0.0-SNAPSHOT-runner`;
8. Notice that the application crashes with the error messages given below.

**Configuration**
```properties
quarkus.datasource.url=jdbc:postgresql://localhost:5432/mypad
quarkus.datasource.driver=org.postgresql.Driver
quarkus.datasource.username=[your username]
quarkus.datasource.password=[your password]
quarkus.liquibase.change-log=db/changeLog.xml
quarkus.liquibase.migrate-at-start=true
```

**Screenshots**
```
__  ____  __  _____   ___  __ ____  ______ 
 --/ __ \/ / / / _ | / _ \/ //_/ / / / __/ 
 -/ /_/ / /_/ / __ |/ , _/ ,< / /_/ /\ \   
--\___\_\____/_/ |_/_/|_/_/|_|\____/___/   
2020-04-02 15:27:12,596 INFO  [liq.exe.jvm.JdbcExecutor] (main) SELECT COUNT(*) FROM biz.databasechangelog
[...]

PR #8368, opened 2020-04-02
Register all methods of Liquibase services sub-classes for reflection

andrejpetras:
Issue: #8357

### DIFF
--- a/extensions/liquibase/deployment/src/main/java/io/quarkus/liquibase/LiquibaseProcessor.java
+++ b/extensions/liquibase/deployment/src/main/java/io/quarkus/liquibase/LiquibaseProcessor.java
@@ -96,8 +96,7 @@ class LiquibaseProcessor {
                 liquibase.sql.visitor.SqlVisitor.class);
 
         // load the liquibase services
-        Stream.of(liquibase.license.LicenseService.class,
-                liquibase.diff.compare.DatabaseObjectComparator.class,
+        Stream.of(liquibase.diff.compare.DatabaseObjectComparator.class,
                 liquibase.parser.NamespaceDetails.class,
                 liquibase.precondition.Precondition.class,
                 liquibase.database.Database.class,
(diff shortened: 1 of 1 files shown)


extensions/liquibase/deployment/src/main/java/io/quarkus/liquibase/LiquibaseProcessor.java
```diff
@@ -187,11 +188,11 @@ ServiceStartBuildItem configureRuntimeProperties(LiquibaseRecorder recorder,
     private void addService(BuildProducer<ReflectiveClassBuildItem> reflective,
             BuildProducer<GeneratedResourceBuildItem> generatedResourceProducer,
             BuildProducer<NativeImageResourceBuildItem> resourceProducer,
-            Class<?> className) {
+            Class<?> className, boolean methods) {
 
         Class<?>[] impl = ServiceLocator.getInstance().findClasses(className);
         if (impl != null && impl.length > 0) {
-            reflective.produce(new ReflectiveClassBuildItem(true, false, false, impl));
+            reflective.produce(new ReflectiveClassBuildItem(true, methods, false, impl));
```
gsmet, line 195: I think I would make it true for all services.

gsmet (changes requested):

extensions/liquibase/deployment/src/main/java/io/quarkus/liquibase/LiquibaseProcessor.java
andrejpetras: It was my first solution, but I'm getting errors for some classes at native compile time.
I'll try again and create two class lists.

Does it make big different for the native image (compile time) if we register all the classes?

andrejpetras: Done, only the `LicenseService` class needs to be `methods=false`

gsmet: So doing that might include more methods than strictly needed in the native image but I think it's safer.

As for the `LicenseService` it's probably related to the substitutions you had to make there so looks OK to me.

gsmet (approved):

gsmet:
Nice work, thanks!

PR #8370, opened 2020-04-02
Funqy azure and lambda http support

patriot1burke:
* Allow Funqy http to use virtual http so that it can run with quarkus-azure-functions-http and quarkus-amazon-lambda-http extensions


patriot1burke:
@stuartwdouglas Funqy code is a WIP, but there was some improvement in lambda code.  Would be nice to get a review.  Is there anybody else I can bug to review things besides gsmet and you?

oztimpower:
The NPE for the test-framework LambdaResourceManager upon stop is fixed by this.  Cheers!

extensions/amazon-lambda/common-runtime/src/main/java/io/quarkus/amazon/lambda/runtime/AbstractLambdaPollLoop.java
```diff
@@ -73,22 +75,35 @@ public void run() {
                                     postResponse(url, output);
                                 }
                             } catch (Exception e) {
-                                log.error("Failed to run lambda", e);
+                                if (e instanceof SocketException) {
+                                    log.warn("Failed to run lambda.  Connection issues: " + e.getMessage());
```
stuartwdouglas, line 79: Shouldn't this still be an error? Also with this failure now you won't get the stack trace.

stuartwdouglas (commented):
Other than the question about log levels LGTM

extensions/amazon-lambda/common-runtime/src/main/java/io/quarkus/amazon/lambda/runtime/AbstractLambdaPollLoop.java
patriot1burke: Exception trace would pop up every single test run with the test harness.  Since we're now making the test harness public, this is kind of an issue.

You don't need a stack trace for SocketException as the error message would be good enough and you could infer where the error was.

stuartwdouglas: Maybe you should check the launch mode and just don't log anything if it is TEST, as it sounds like you will still get the WARN in tests now?

patriot1burke: I just reverted this file.  Don't feel like arguing.

stuartwdouglas: So there is always going to be a big stacktrace in the logs when running tests now?

patriot1burke: For now.  I don't have time to update this PR as I'm working on something else at the moment. A contributor is working on Funqy so I'd like to get this PR in.

stuartwdouglas (approved):

PR #8376, opened 2020-04-03
Support Kubernetes ConfigMap and Secrets as configuration/value sources 

### LINKED ISSUE #6745: Support Kubernetes ConfigMap and Secrets as configuration/value sources
**Description**
Is https://github.com/quarkusio/quarkus/issues/305#issuecomment-516830160 about mapping config map to a "magic destination" what we expect users to do for kubernetes deployments ? 

Would it be better that quarkus-kubernetes extension or maybe a separate quarkus-kubernetes-config extension would have a configuration source that could be added in with a fairly high precedence order by default.

And then kubernetes-secrets would be available somehow to resolve values in any config source ?

geoand:
Still to do:

- [x]  Add Secret support
- [x]  Add documentation (added in #8739)
- [ ]  Figure out whether we should support namespaced ConfigMaps (and Secrets) - the current implementation only looks for them in the current namespace
- [x]  Figure out whether this should be a separate extension or not (perhaps it should because then that would open up opportunities to use a slimmer version of the Kubernetes Model in the future)

Fixes: #6745


geoand:
cc @maxandersen @iocanel @gastaldi @Ladicek

extensions/kubernetes-client/runtime/src/main/java/io/quarkus/kubernetes/client/runtime/KubernetesConfigSourceConfig.java
```diff
@@ -0,0 +1,31 @@
...
+public class KubernetesConfigSourceConfig {
+
+    /**
+     * If set to true, the application will attempt to look up the configuration from the API server
+     */
+    @ConfigItem(defaultValue = "false")
+    public boolean enabled;
+
+    /**
+     * If set to true, the application will not stand up if it cannot obtain configuration from the Config Server
```
Ladicek, line 20: Please avoid the "stand up" idiom :-) What's wrong with "start"?

Also, "Config Server" -> "API server"?

Ladicek:
IMHO, this totally needs to be a separate extension.

Also, I'm leaving for your consideration whether this should or shouldn't be part of SmallRye Config. For example, here's a PR that adds a Kubernetes config source in SRye Config: https://github.com/smallrye/smallrye-config/pull/238 Except it's a fancy filesystem config source :-/ This one is what I'd expect from a Kubernetes config source.

geoand:
We can certainly start out having this here and move the core parts into the SmallryeConfig if that makes sense

gastaldi:
i agree with @Ladicek ,that should be a separate extension (like the config YAML extension). Preferably having SM Config support should be the first step (to avoid API inconsistencies, etc)

geoand:
I don't think that SR Config support should be a prerequisite for this, but a useful future avenue to explore.

geoand:
I moved the code into its own extension.

Also, I think that item no.2 can be addressed later on if people request it, while I would like to do the documentation in a subsequent PR.

Ladicek:
Don't be lazy, documentation is as important as tests :stuck_out_tongue: :grin:

geoand:
Sure, but I don't want the PR to be help up by typos or my very frequent habit of forgetting to type words as I think about that I need to right 😂

dmlloyd (approved):
The config source stuff looks OK to me, I don't know anything about the Kubernetes part though.  It might be a good idea to explain (in the commit message) why (if I understand correctly) the "volume" approach which `io.smallrye.config.source.file.FileSystemConfigSource` exploits was not chosen.

geoand:
@dmlloyd the idea is to provide users the ability to read `ConfigMap` or `Secret` via the Kubernetes API _instead_ of the file system.
The advantage of this approach is that the user doesn't need to alter the Kubernetes manifest to configure it to mount a `ConfigMap` or `Secret`.

So it's basically complementary to what users can already do with regular config sources.

geoand:
A follow-up for @iocanel:

When this extension is active, generate the proper RoleBinding for reading ConfigMap, and Secret

geoand:
@maxandersen any comments on this?

maxandersen:
> @maxandersen any comments on this?

discussed here https://quarkusio.zulipchat.com/#narrow/stream/187038-dev/topic/k8s.20config/near/193046063

but in short: looks good, its default off which is good as its set to also  fails if config/secrets not reachable.  I haven't tried to run it in practice (had build issues today ;/ ) but looking at code and expected semantics  looks good imo.

geoand:
Thanks for checking @maxandersen.

I'll write some documentation within the week.

@iocanel mind looking at the `RoleBinding` enhancement I mentioned above?

PR #8377, opened 2020-04-03
feat: add spring datasource data property + refactor

aureamunoz:
I added the new property and did a refactor to go back to the first implementation with a map because we are adding more properties....I let you take a look @geoand

### DIFF
--- a/extensions/spring-data-jpa/deployment/src/main/java/io/quarkus/spring/data/deployment/SpringDataJPAProcessor.java
+++ b/extensions/spring-data-jpa/deployment/src/main/java/io/quarkus/spring/data/deployment/SpringDataJPAProcessor.java
@@ -1,5 +1,6 @@
 package io.quarkus.spring.data.deployment;
 
+import static java.util.stream.Collectors.*;
 import static java.util.stream.Collectors.toList;
 
 import java.lang.reflect.Modifier;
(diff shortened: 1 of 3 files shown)


extensions/spring-data-jpa/deployment/src/main/java/io/quarkus/spring/data/deployment/SpringDataJPAProcessor.java
```diff
@@ -1,13 +1,16 @@
 package io.quarkus.spring.data.deployment;
 
+import static java.util.stream.Collectors.*;
```
geoand, line 3: No start imports please :)

geoand:
I am not sure about the refactor, because Quarkus's `quarkus.hibernate-orm.sql-load-script` property isn't a direct replacement for `spring.datasource.data` because it does not support the prefix (like `classpath:` or `file:`), so we can't just tell users to replace the property. 
We should probably just use the term "adapt" in this case so the user knows it's not a straight replace.

geoand:
Can we also link to the issue that sparked this PR?

extensions/spring-data-jpa/deployment/src/main/java/io/quarkus/spring/data/deployment/SpringDataJPAProcessor.java
aureamunoz: I did it manually :woman_facepalming:

aureamunoz:
I tried a more adapted message @geoand

geoand (approved):
Nice, thanks!

geoand:
@aureamunoz please squash the 2 commits. Thanks

aureamunoz:
Done!

PR #8380, opened 2020-04-03
Document how to use a random test port

famod:
Support was added in #4140 but no documentaion seems to exist to this day.

### DIFF
--- a/docs/src/main/asciidoc/getting-started-testing.adoc
+++ b/docs/src/main/asciidoc/getting-started-testing.adoc
@@ -150,6 +150,7 @@ You can configure the port used by tests by configuring `quarkus.http.test-port`
 ----
 quarkus.http.test-port=8083
 ----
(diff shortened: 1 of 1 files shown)


docs/src/main/asciidoc/getting-started-testing.adoc
```diff
@@ -150,6 +150,7 @@ You can configure the port used by tests by configuring `quarkus.http.test-port`
 ----
 quarkus.http.test-port=8083
 ----
+`0` will yield a random port (assigned by the operating system).
```
geoand, line 153:
```suggestion
`0` will result in the use of a random port (assigned by the operating system).
```

famod: Done.

geoand (approved):
Thanks a lot!

PR #8383, opened 2020-04-03
Add config to override chunk sizes

### LINKED ISSUE #8379: Add config to override DEFAULT_MAX_CHUNK_SIZE from HttpServerOptions
**Description**
When using SSE you are limited to 8192 bytes per chunk by default from Vert.x:
https://vertx.io/docs/apidocs/io/vertx/core/http/HttpServerOptions.html#DEFAULT_MAX_CHUNK_SIZE

In order to send larger payload sizes per chunk (think Base64 encoded images that might be multiple megabytes), this should be configurable.

I am not sure if there are downsides to large chunk sizes for memory usage or performance. That being said, I don't want to have to revert to a more complex Websockets solution for such a use case.

**Implementation ideas**
Make a Quarkus HTTP config option for DEFAULT_MAX_CHUNK_SIZE

gastaldi:
This commit introduces a new config property `quarkus.http.limits.max-chunk-size`

Fixes #8379

### DIFF
--- a/extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/ServerLimitsConfig.java
+++ b/extensions/vertx-http/runtime/src/main/java/io/quarkus/vertx/http/runtime/ServerLimitsConfig.java
@@ -9,7 +9,7 @@ import io.quarkus.runtime.configuration.MemorySize;
 @ConfigGroup
 public class ServerLimitsConfig {
     /**
-     * The the maximum length of all headers.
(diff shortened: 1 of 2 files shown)


gastaldi:
@murphye can you give it a try with this PR?

geoand (approved):
:+1:

machi1990 (approved):
LGTM

stuartwdouglas (approved):

PR #8388, opened 2020-04-04
Fix reference to Vert.x API used in examples

apeteri:
The text said that examples in this guide are using the Axle API, while this seems to be not the case; Mutiny classes are used across all code snippets, with the exception of a single part that shows how the four APIs differ.


machi1990 (approved):

machi1990:
LGTM

Thanks @apeteri for the contribution.

PR #8389, opened 2020-04-04
Remove WireMock from BOM

### LINKED ISSUE #8378: (Outdated) Wiremock should not be defined in BOM?
**Describe the bug**
A fellow developer has just introduced Wiremock in our current project and I was surprised that he did not have to define a version for it. I then discovered that suddenly Junit4 popped up on the test classpath (Eclipse asked me to choose between two `@Test` annotations).
This is when I stumbled upon https://github.com/tomakehurst/wiremock/issues/684#issuecomment-600127042:

> the JUnit 4 dependency is already optional

which is not (yet) true for 2.24.1 that is defined in the Quarkus BOM. The dependency was removed in 2.25.0, latest version is 2.26.3.

To make a long story short: Why does the "user-facing" BOM bring in Wiremock anyway?

**Expected behavior**
_No_ Wiremock in the Quarkus BOM _at all_ or at least a recent version that does not pull in Junit4.

**Actual behavior**
Outdated Wiremock with JUnit4 dependency is present in BOM.

**To Reproduce**
1. Just look at the BOM 😉 and the Dependencies: https://mvnrepository.com/artifact/com.github.tomakehurst/wiremock-jre8/2.24.1

**Configuration**
n/a

**Screenshots**
n/a

 - Quarkus version or git rev: 1.3.1.Final
 - Build tool (ie. output of `mvnw --version` or `gradlew --version`): Maven 3.6.3

**Additional context**
Wiremock was introduced here, but just in a test of an extension: https://github.com/quarkusio/quarkus/commit/b958ca1022b64c2698701706312230df7749a779

famod:
Fixes #8378

### DIFF
--- a/bom/runtime/pom.xml
+++ b/bom/runtime/pom.xml
@@ -172,7 +172,6 @@
         <optaplanner.version>7.34.0.Final</optaplanner.version>
         <sundr.version>0.19.1</sundr.version> <!-- this is to avoid annoying pop-up in eclipse about failure to init Velocity logging -->
         <flapdoodle.mongo.version>2.2.0</flapdoodle.mongo.version>
-        <wiremock.version>2.24.1</wiremock.version>
         <quarkus-spring-api.version>5.2.SP1</quarkus-spring-api.version>
         <quarkus-spring-data-api.version>2.1.Final</quarkus-spring-data-api.version>
         <quarkus-spring-security-api.version>5.2.Final</quarkus-spring-security-api.version>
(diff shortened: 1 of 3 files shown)


geoand (approved):

gsmet:
Do we really want to do that?

First, this is a breaking change as it might break some applications using wiremock. Second, we don't want versions in separate pom files. And third, we have all the versions centralized, this one should be too.

If we don't want it in the bom, it should be in the build-parent so that we are sure it's consistent across the project.

Also maybe we should add it to dependabot?

geoand:
We can probably add it to `dependanbot` and be fine. It's just a test dependency for some of the integration tests, nothing more.

gsmet:
we still need it to be consistent because of dependency convergence. Thus it would be better to have it centralized.

geoand:
OK

famod:
If you choose to centralize it (which makes sense IMHO), please beware that there are more WireMock dependencies that just `wiremock-jre8`.
In some other modules there is also the vanilla (non-JRE8) `wiremock` dependency and there is also `wiremock-maven-plugin`, with mixed versions.

PR #8390, opened 2020-04-04
do not quote url so that it is rendered nicely

machi1990:


docs/src/main/asciidoc/vertx.adoc
```diff
@@ -239,7 +239,7 @@ public class GreetingResource {
 ./mvnw compile quarkus:dev
 ----
 
-Then, open your browser to 'http://localhost:8080/lorem' and you should get the message.
+Then, open your browser to `http://localhost:8080/lorem` and you should get the message.
```
gsmet, line 242: Dunno. I think it would be better to not use anything and have a proper link that is clickable?

gsmet (changes requested):

gsmet (approved):

PR #8397, opened 2020-04-05
Increase timeout for Data3 group of native tests

geoand:
Done because I've seen this group exceeding the current time limit on some PRs


machi1990 (approved):
Thanks, I have been meaning to do this too.

PR #8403, opened 2020-04-06
Warn if resteasy-mutiny is missing

stuartwdouglas:


stuartwdouglas:
I ran into this with the reactive security PR in the keycloak tests, I think we really need to warn users about this.

gsmet (approved):
Fixed the case of `Mutiny` and force pushed. Let's wait for CI.

PR #8408, opened 2020-04-06
add list jvm list arguments

### LINKED ISSUE #6464: JVM arguments should be an array instead (under Gradle's QuarkusDev task)
## Description

Currently, `jvmArgs` is just a `String` member; I think it's more appropriate to "convert it" into an array/list of `String`s.

```groovy
tasks.withType(QuarkusDev) {
  final def walletHome = System.getenv("WALLET_HOME")

  // Sausage-like String; way better as "tokens"
  jvmArgs = "-Djavax.net.ssl.keyStore=/path/to/keystores/keystore.jks -Djavax.net.ssl.keyStorePassword=store -Djavax.net.ssl.trustStore=/path/to/keystores/truststore.jks -Djavax.net.ssl.trustStorePassword=trusted -Djdk.tls.allowUnsafeServerCertChange=true -Djdk.tls.client.protocols=TLSv1.2 -Doracle.net.tns_admin=${walletHome} -Doracle.net.wallet_location=(SOURCE=(METHOD=FILE)(METHOD_DATA=(DIRECTORY=${walletHome})))"
}
```

glefloch:
Add `jvmArgsList` property on `quarkusDev` task and deprecate `jvmArgs`. 

close #6464

### DIFF
--- a/devtools/gradle/src/main/java/io/quarkus/gradle/tasks/QuarkusDev.java
+++ b/devtools/gradle/src/main/java/io/quarkus/gradle/tasks/QuarkusDev.java
@@ -19,7 +19,6 @@ import java.net.URL;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Path;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
@@ -75,7 +74,7 @@ public class QuarkusDev extends QuarkusTask {
 
     private String workingDir;
 
-    private String jvmArgs;
+    private List<String> jvmArgs;
 
     private boolean preventnoverify = false;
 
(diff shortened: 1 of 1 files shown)


gastaldi (changes requested):
@glefloch hey! Thanks for your PR! 

Can you simply change the `jvmArgs` to a `List` instead? We know it's a breaking change (we'll make sure it is communicated properly in the release this is in) and the team feels it is a better option

glefloch:
@gastaldi, Thanks for the review. I agree with you, this is cleaner to only have one property for JVM arguments. I pushed an update. Is there a `changelog` to track breaking changes ?

gastaldi:
The changelog is created when the release is performed. We maintain a Migration Guide here: https://github.com/quarkusio/quarkus/wiki/Migration-Guide-1.4

It would be nice if you could mention this change there

gastaldi (approved):
Looks great to me. Good job!

PR #8409, opened 2020-04-06
Remove RESTEasy exclusion

### LINKED ISSUE #7809: New exception on 1.3.0.CR2 with kafka and without reasteasy
**Describe the bug**

On a project using kafka, undertow and vertx web (and not reasteasy), when upgrading from `1.3.0.Alpha1` to `1.3.0.CR2`, in jar mode, this exception is now popping (no problem before):
```
        at io.quarkus.runner.ApplicationImpl.doStart(ApplicationImpl.zig:248)
        at io.quarkus.runtime.Application.start(Application.java:90)
        at io.quarkus.runtime.Application.run(Application.java:228)
Caused by: java.lang.NoClassDefFoundError: org/jboss/resteasy/core/ResteasyContext
        at io.smallrye.context.SmallRyeThreadContext.currentContextExecutor(SmallRyeThreadContext.java:208)
        at io.smallrye.context.propagators.rxjava2.ContextPropagatorOnFlowableAssemblyAction.apply(ContextPropagatorOnFlowableAssemblyAction.java:23)
        at io.smallrye.context.propagators.rxjava2.ContextPropagatorOnFlowableAssemblyAction.apply(ContextPropagatorOnFlowableAssemblyAction.java:11)
        at io.reactivex.plugins.RxJavaPlugins.apply(RxJavaPlugins.java:1268)
        at io.reactivex.plugins.RxJavaPlugins.onAssembly(RxJavaPlugins.java:1022)
        at io.vertx.reactivex.FlowableHelper.toFlowable(FlowableHelper.java:36)
        at io.vertx.reactivex.kafka.client.consumer.KafkaConsumer.toFlowable(KafkaConsumer.java:107)
        at io.smallrye.reactive.messaging.kafka.KafkaSource.<init>(KafkaSource.java:56)
        at io.smallrye.reactive.messaging.kafka.KafkaConnector.getPublisherBuilder(KafkaConnector.java:65)
        at io.smallrye.reactive.messaging.kafka.KafkaConnector_ClientProxy.getPublisherBuilder(KafkaConnector_ClientProxy.zig:203)
        at io.smallrye.reactive.messaging.impl.ConfiguredChannelFactory.createPublisherBuilder(ConfiguredChannelFactory.java:171)
        at io.smallrye.reactive.messaging.impl.ConfiguredChannelFactory.lambda$register$4(ConfiguredChannelFactory.java:149)
        at java.util.HashMap.forEach(HashMap.java:1289)
        at io.smallrye.reactive.messaging.impl.ConfiguredChannelFactory.register(ConfiguredChannelFactory.java:149)
        at io.smallrye.reactive.messaging.impl.ConfiguredChannelFactory.initialize(ConfiguredChannelFactory.java:123)
        at io.smallrye.reactive.messaging.impl.ConfiguredChannelFactory_ClientProxy.initialize(ConfiguredChannelFactory_ClientProxy.zig:195)
        at java.util.Iterator.forEachRemaining(Iterator.java:116)
        at java.util.Spliterators$IteratorSpliterator.forEachRemaining(Spliterators.java:1801)
        at java.util.stream.ReferencePipeline$Head.forEach(ReferencePipeline.java:647)
        at io.smallrye.reactive.messaging.extension.MediatorManager.initializeAndRun(MediatorManager.java:156)
        at io.smallrye.reactive.messaging.extension.MediatorManager_ClientProxy.initializeAndRun(MediatorManager_ClientProxy.zig:100)
        at io.quarkus.smallrye.reactivemessaging.runtime.SmallRyeReactiveMessagingLifecycle.onApplicationStart(SmallRyeReactiveMessagingLifecycle.java:20)
        at io.quarkus.smallrye.reactivemessaging.runtime.SmallRyeReactiveMessagingLifecycle_Observer_onApplicationStart_4e8937813d9e8faff65c3c07f88fa96615b70e70.notify(SmallRyeReactiveMessagingLifecycle_Observer_onApplicationStart_4e8937813d9e8faff65c3c07f88fa96615b70e70.zig:51)
        at io.quarkus.arc.impl.EventImpl$Notifier.notifyObservers(EventImpl.java:282)
        at io.quarkus.arc.impl.EventImpl$Notifier.notify(EventImpl.java:267)
        at io.quarkus.arc.impl.EventImpl.fire(EventImpl.java:69)
        at io.quarkus.arc.runtime.LifecycleEventRunner.fireStartupEvent(LifecycleEventRunner.java:23)
        at io.quarkus.arc.runtime.ArcRecorder.handleLifecycleEvents(ArcRecorder.java:109)
        at io.quarkus.deployment.steps.LifecycleEventsBuildStep$startupEvent39.deploy_0(LifecycleEventsBuildStep$startupEvent39.zig:58)
        at io.quarkus.deployment.steps.LifecycleEventsBuildStep$startupEvent39.deploy(LifecycleEventsBuildStep$startupEvent39.zig:77)
        at io.quarkus.runner.ApplicationImpl.doStart(ApplicationImpl.zig:210)
        ... 3 more
Caused by: java.lang.ClassNotFoundException: org.jboss.resteasy.core.ResteasyContext
        at java.net.URLClassLoader.findClass(URLClassLoader.java:382)
        at java.lang.ClassLoader.loadClass(ClassLoader.java:418)
        at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:352)
        at java.lang.ClassLoader.loadClass(ClassLoader.java:351)
        ... 37 more

```

CC @cescoffier

stuartwdouglas:
This instead adds the dependency in RESTEasy Common,
with an exclusion on the actual CP implementation.

This approach is much less problematic, as RESTEasy
has no knowledge of CP, so there is no chance these
classes can be accidently loaded unless CP is on the
classpath.

The current approach does not have this advatage so
problematic hacks have been used to try and stop it
being loaded, but this approach is much simpler.
Fixes #7809

### DIFF
--- a/build-parent/pom.xml
+++ b/build-parent/pom.xml
@@ -298,7 +298,7 @@
                                             <!-- We use our own impl here, including this one causes problems-->
                                             <exclude>org.jboss.resteasy:resteasy-context-propagation</exclude>
                                         </excludes>
-                                        <includes>a
+                                        <includes>
                                             <!-- this is for REST Assured -->
                                             <include>jakarta.xml.bind:jakarta.xml.bind-api:*:*:test</include>
                                         </includes>
--- a/build-parent/pom.xml
+++ b/build-parent/pom.xml
@@ -295,8 +295,10 @@
                                             <!-- Ban checker-qual, we don't use Checker Framework -->
                                             <exclude>org.checkerframework:checker-qual</exclude>
                                             <exclude>com.google.code.findbugs:jsr305</exclude>
+                                            <!-- We use our own impl here, including this one causes problems-->
+                                            <exclude>org.jboss.resteasy:resteasy-context-propagation</exclude>
                                         </excludes>
-                                        <includes>
+                                        <includes>a
                                             <!-- this is for REST Assured -->
                                             <include>jakarta.xml.bind:jakarta.xml.bind-api:*:*:test</include>
                                         </includes>
--- a/bom/runtime/pom.xml
+++ b/bom/runtime/pom.xml
@@ -2014,6 +2014,12 @@
                 <groupId>org.jboss.resteasy</groupId>
                 <artifactId>resteasy-rxjava2</artifactId>
                 <version>${resteasy.version}</version>
+                <exclusions>
+                    <exclusion>
+                        <groupId>org.jboss.resteasy</groupId>
+                        <artifactId>resteasy-context-propagation</artifactId>
+                    </exclusion>
+                </exclusions>
             </dependency>
             <dependency>
                 <groupId>org.jboss.resteasy</groupId>
--- a/extensions/mutiny/runtime/pom.xml
+++ b/extensions/mutiny/runtime/pom.xml
@@ -26,12 +26,6 @@
         <dependency>
             <groupId>io.quarkus</groupId>
             <artifactId>quarkus-smallrye-context-propagation</artifactId>
-            <exclusions>
-                <exclusion>
-                    <groupId>org.jboss.resteasy</groupId>
-                    <artifactId>resteasy-context-propagation</artifactId>
-                </exclusion>
-            </exclusions>
         </dependency>
         <dependency>
             <groupId>io.smallrye.reactive</groupId>
--- a/extensions/resteasy-common/runtime/pom.xml
+++ b/extensions/resteasy-common/runtime/pom.xml
@@ -26,7 +26,6 @@
             <groupId>org.jboss.resteasy</groupId>
             <artifactId>resteasy-core</artifactId>
         </dependency>
-        <!--
         <dependency>
             <groupId>org.jboss.resteasy</groupId>
             <artifactId>resteasy-context-propagation</artifactId>
@@ -37,7 +36,6 @@
                 </exclusion>
             </exclusions>
         </dependency>
-        -->
         <dependency>
             <groupId>io.quarkus</groupId>
             <artifactId>quarkus-arc</artifactId>
--- a/extensions/resteasy-common/runtime/pom.xml
+++ b/extensions/resteasy-common/runtime/pom.xml
@@ -57,6 +57,7 @@
                         <excludedArtifact>javax.activation:javax.activation-api</excludedArtifact>
                         <excludedArtifact>javax.activation:activation</excludedArtifact>
                         <excludedArtifact>jakarta.ws.rs:jakarta.ws.rs-api</excludedArtifact>
+                        <excludedArtifact>org.jboss.resteasy:resteasy-context-propagation</excludedArtifact>
                     </excludedArtifacts>
                 </configuration>
             </plugin>
--- a/extensions/resteasy-common/runtime/pom.xml
+++ b/extensions/resteasy-common/runtime/pom.xml
@@ -27,16 +27,6 @@
             <artifactId>resteasy-core</artifactId>
         </dependency>
         <dependency>
-            <groupId>org.jboss.resteasy</groupId>
-            <artifactId>resteasy-context-propagation</artifactId>
-            <exclusions>
-                <exclusion>
-                    <groupId>io.smallrye</groupId>
-                    <artifactId>smallrye-context-propagation</artifactId>
-                </exclusion>
-            </exclusions>
-        </dependency>
-        <dependency>
             <groupId>io.quarkus</groupId>
             <artifactId>quarkus-arc</artifactId>
         </dependency>
--- a/extensions/resteasy-common/runtime/src/main/java/io/quarkus/resteasy/common/runtime/ResteasyContextProvider.java
+++ b/extensions/resteasy-common/runtime/src/main/java/io/quarkus/resteasy/common/runtime/ResteasyContextProvider.java
@@ -1,0 +1,40 @@
+package io.quarkus.resteasy.common.runtime;
+
+import java.util.Collections;
+import java.util.Map;
+
+import org.eclipse.microprofile.context.spi.ThreadContextProvider;
+import org.eclipse.microprofile.context.spi.ThreadContextSnapshot;
+import org.jboss.resteasy.core.ResteasyContext;
+
+public class ResteasyContextProvider implements ThreadContextProvider {
+
+    private static final String JAXRS_CONTEXT = "JAX-RS";
+
+    @Override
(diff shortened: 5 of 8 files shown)


gsmet (dismissed):
But doesn't that mean that you will bring `resteasy-core` as soon as you use Mutiny in your app?

Looks a bit counter productive?

cescoffier:
Mutiny must NOT depend on Resteasy.

stuartwdouglas:
There is already an exclusion: https://github.com/quarkusio/quarkus/blob/77ce7dc8547b4279805956d8c886c5915a44c185/extensions/smallrye-context-propagation/runtime/pom.xml#L32

The bit that you end up depending on is resteasy-context-propagation which has 2 classes in it

FroMage (commented):
This looks like the right thing to do:

- resteasy-cp was only made to depend on `sr-cp` due to WildFly's multi-tenancy and forcing lazy initialisation of `sr-cp` for every new modular class loader
- it should only really depend on `mp-cp`, so this exclusion does that
- resteasy should come with its context provider, like arc does, so it should indeed depend on `resteasy-cp`
- those providers will only be used if we have context propagation anyway to load them, which should be the case with Mutiny present.

But I fear this is incomplete, because the problematic class in `resteasy-context-propagation` I mentioned (for WildFly) is:

```java
@Provider
public class ContextFeature implements Feature
{
   @Override
   public boolean configure(FeatureContext context)
   {
      // this is tied to the deployment, which is what we want for the reactive context
      if(context.getConfiguration().getRuntimeType() == RuntimeType.CLIENT)
         return false;
      Dispatcher dispatcher = ResteasyContext.getContextData(Dispatcher.class);
      if(dispatcher == null) {
         // this can happen, but it means we're not able to find a deployment
         return false;
      }
      // Make sure we have context propagation for this class loader
      ContextManagerProvider.instance().getContextManager();
      return true;
   }

}
```

And so I fear that having this class present in Quarkus IFF we don't have SR-CP will lead to another type of CNFE.

With this patch, this can happen if we have a dep on `resteasy-common` which pulls in `resteasy-context-propagation` (and this class) and no dep on SR-CP.

We should either disable this class (can we?) in Quarkus, or I should change that module in RESTEasy (not sure how because this is required in WildFly).

patriot1burke:
I do not agree with this PR and wish I had seen it earlier. 

@FroMage Why not just remove ContextFeature?  I'm not sure WTF that class does other than some "weird" defensive coding.  Then I think its ok to add resteasy-cp as to quarkus-resteasy so long as smallrye dependencies are excluded.

stuartwdouglas:
I did not notice that ContextFeature was also these, so this is a problem either way, if this is present it will break both unless both features are present. 

@aloubyansky and I were talking about a conditional dependencies feature the other day, which would be really useful here.

aloubyansky:
I think I have two concerns about introducing conditional deps in our bootstrap/build. I'm not necessarily against the feature, just having doubts.
1) so far the runtime classpath was completely defined by the user and the utilities like `mvn dependency:tree` would show the actual dep tree.
2) dependency version convergence. New transitive dependencies may completely change the resulting tree (by that mean the transitive dep version and even the whole tree branches may disappear/appear).
I do recognize this is a valid use-case though. In fact, I already implemented something similar in Galleon provisioning (although the concerns mentioned above aren't relevant in that context).
If we decide we absolutely need this kind of feature, I think I'd consider adding these conditional deps to the __deployment__ deps but adding some info about what they are to the extension descriptor. I'm thinking the way we assemble the full deployment dependency graph will make it easier/safer to perform the version convergence of the whole thing. This would need some testing of course.

FroMage:
> Why not just remove ContextFeature? I'm not sure WTF that class does other than some "weird" defensive coding

I explained it: it's the WildFly integration. Without it, CP doesn't work in WildFly, which is the basis for the RESTEasy test suite.

We can't remove it, but we can copy the `ResteasyContextProvider` to quarkus and blacklist `resteasy-context-propagation`.

But this won't solve the issue of how we get users to have the dependency on SR-CP automatically when they do reactive rest endpoints. We need to find a solution for that too.

FroMage:
TBH I think it's fine if we tell our users that any reactive programming on resteasy should use the `quarkus-resteasy-mutiny` which does depend on SR-CP.

FroMage:
I've updated the PR to specify our own context propagator for resteasy in `quarkus-resteasy-common`, this should fix this issue but perhaps we want to blacklist `resteasy-context-propagation`?

cescoffier (approved):
LGTM!

FroMage:
Tests fail because `org.jboss.resteasy:resteasy-rxjava2` depends on `resteasy-cp` which we don't want.

Is it acceptable to tell users to do an exclusion every time they import this module? I don't think we can define the exclusion in our BOM, can we? I think exclusions are not transitive. @gsmet do you know?

FroMage:
Just tried it and it seems transitive in the BOM.

stuartwdouglas:
This is pretty broken ATM, either I can change back to the earlier version that just removed the exclusion from quarkus-smallrye-mutitny, or we can close this and look at the idea of conditional dependencies.

FroMage:
Why do you say that? The changes are pushed should be good for a fix, no?

stuartwdouglas:
I had not noticed that you had pushed to the PR. Is context propagation always pulled in with RESTeasy now?

FroMage:
No, but RESTEasy will always have a CP service provider, so if CP is imported (via mutiny for example) then RESTEasy will have its context propagated correctly.

build-parent/pom.xml
```diff
@@ -295,8 +295,10 @@
                                             <!-- Ban checker-qual, we don't use Checker Framework -->
                                             <exclude>org.checkerframework:checker-qual</exclude>
                                             <exclude>com.google.code.findbugs:jsr305</exclude>
+                                            <!-- We use our own impl here, including this one causes problems-->
+                                            <exclude>org.jboss.resteasy:resteasy-context-propagation</exclude>
                                         </excludes>
-                                        <includes>
+                                        <includes>a
```
FroMage, line 301:
```suggestion
                                        <includes>
```

FroMage:
@gsmet can you re-review please?

gsmet:
@stuartwdouglas considering it's your PR, could you have a look at the final result and merge it if it seems appropriate?

PR #8420, opened 2020-04-06
Revert how readers and writers are registered for REST clients

### LINKED ISSUE #8223: Using the REST Client with Multipart not working in all cases
**Describe the bug**
(Describe the problem clearly and concisely.)
Proxying multipart call using `@RestClient` is not working with only GET resource, you have to have at least one POST.

**Expected behavior**
(Describe the expected behavior clearly and concisely.)
Expected behavior is successful multipart call toward another service even when there is only GET resource.

**Actual behavior**
Results is:
RESTEASY003215: could not find writer for content-type multipart/form-data type: org.acme.rest.client.multipart.MultipartBody

**To Reproduce**
1.  Follow this guide till end: https://quarkus.io/guides/rest-client-multipart
2. At this point everything should work so run test to confirm everything is working as intended.
3. Replace only `@POST` with `@GET` at
Class: `MultipartClientResource `
Method: `public String sendFile() throws Exception`

So, from:
```
@POST
@Path("/multipart")
public String sendFile() throws Exception {
    MultipartBody body = new MultipartBody();
    body.fileName = "greeting.txt";
    body.file = new ByteArrayInputStream("HELLO WORLD".getBytes(StandardCharsets.UTF_8));
    return service.sendMultipartData(body);
}
```
to:
```
@GET
@Path("/multipart")
public String sendFile() throws Exception {
    MultipartBody body = new MultipartBody();
    body.fileName = "greeting.txt";
    body.file = new ByteArrayInputStream("HELLO WORLD".getBytes(StandardCharsets.UTF_8));
    return service.sendMultipartData(body);
}
```

Now if you run it, it should fail with:

> RESTEASY003215: could not find writer for content-type multipart/form-data type: org.acme.rest.client.multipart.MultipartBody

To make it work, leave sendFile as GET and add another simple POST method in same class. Such as:
```
@POST
@Path("/temp")
public String simplePost() throws Exception {
    return "something";
}
```
Now it should work with GET.

 - Output of `uname -a` or `ver`: Microsoft Windows [Version 10.0.18363.720]
 - Output of `java -version`: openjdk version "1.8.0_202"
 - Quarkus version or git rev: 1.3.0
 - Build tool (ie. output of `mvnw --version` or `gradlew --version`): Apache Maven 3.6.3 (cecedd343002696d0abb50b32b541b8a6ba2883f)

gsmet:
They used to be registered as for the JAX-RS resources but they should
be registered the opposite way.

Fixes #8223

### DIFF
--- a/extensions/resteasy-common/deployment/src/main/java/io/quarkus/resteasy/common/deployment/RestClientBuildItem.java
+++ b/extensions/resteasy-common/deployment/src/main/java/io/quarkus/resteasy/common/deployment/RestClientBuildItem.java
@@ -1,0 +1,21 @@
+package io.quarkus.resteasy.common.deployment;
+
+import io.quarkus.builder.item.MultiBuildItem;
+
+/**
+ * Used to mark a class as a potential REST client interface consumed by the MicroProfile REST client.
+ * <p>
+ * Useful when you want to apply different behaviors to REST resources and REST clients.
+ */
+public final class RestClientBuildItem extends MultiBuildItem {
+
+    private String interfaceName;
+
+    public RestClientBuildItem(String interfaceName) {
+        this.interfaceName = interfaceName;
+    }
+
+    public String getInterfaceName() {
+        return interfaceName;
+    }
+}
--- a/extensions/rest-client/deployment/src/main/java/io/quarkus/restclient/deployment/RestClientProcessor.java
+++ b/extensions/rest-client/deployment/src/main/java/io/quarkus/restclient/deployment/RestClientProcessor.java
@@ -68,6 +68,7 @@ import io.quarkus.gizmo.ResultHandle;
 import io.quarkus.restclient.runtime.RestClientBase;
 import io.quarkus.restclient.runtime.RestClientRecorder;
 import io.quarkus.resteasy.common.deployment.JaxrsProvidersToRegisterBuildItem;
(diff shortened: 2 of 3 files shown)


gsmet:
I discussed with @irenakezic , she will provide an integration test as another PR.

extensions/resteasy-common/deployment/src/main/java/io/quarkus/resteasy/common/deployment/RestClientBuildItem.java
```diff
@@ -0,0 +1,16 @@
+package io.quarkus.resteasy.common.deployment;
+
+import io.quarkus.builder.item.MultiBuildItem;
+
+public final class RestClientBuildItem extends MultiBuildItem {
```
geoand, line 5: Can we have some Javadoc on this one?

It seems pretty generic to be able to understand what it's used for

geoand (approved):

extensions/resteasy-common/deployment/src/main/java/io/quarkus/resteasy/common/deployment/RestClientBuildItem.java
gsmet: Done.

PR #8424, opened 2020-04-06
Set a path on the invalidated cookie

### LINKED ISSUE #8417: Quarkus OIDC with cookie-path does not remove cookie
**Describe the bug**
When using Quarkus OIDC with type web-app and a cookie-path set, the whole authentication process only works the first time. When the tokens expire, a new challenge is sent out but the old session cookie is never removed. The call to remove cookie at https://github.com/quarkusio/quarkus/blob/7e969004866b195a4ee20b5e16ff005b9e3c4d84/extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/CodeAuthenticationMechanism.java#L283

will not work because the initial cookie was set with a cookiepath, which means it can only be expired using the same cookiepath. 

**Expected behavior**
1. user logs in
2. tokens expire
3. new challenge is sent, old session cookie is removed, redirect to OIDC provider
4. redirect back to app, new session cookie is created

**Actual behavior**
1. user logs in
2. tokens expire
3. new challenge is sent, old session cookie is not removed, redirect to OIDC provider
4. redirect back to app, new challenge is sent, old session cookie is not removed, redirect to OIDC provider
5. infinite loop

**Configuration**
```properties
quarkus.oidc.authentication.cookie-path=/api
```
 - Quarkus version or git rev: 91248634585a3b75ccfc8addf49902bd04b973cd

sberyozkin:
Fixes #8417
CC @pmlopes, Hi Paulo, I've copied this code from `CookieImpl` and updated a path.

### DIFF
--- a/extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/CodeAuthenticationMechanism.java
+++ b/extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/CodeAuthenticationMechanism.java
@@ -27,6 +27,7 @@ import io.quarkus.vertx.http.runtime.security.AuthenticationRedirectException;
 import io.quarkus.vertx.http.runtime.security.ChallengeData;
 import io.vertx.core.http.Cookie;
 import io.vertx.core.http.HttpHeaders;
+import io.vertx.core.http.impl.ServerCookie;
 import io.vertx.core.json.JsonArray;
 import io.vertx.core.json.JsonObject;
 import io.vertx.ext.auth.oauth2.AccessToken;
@@ -85,9 +86,8 @@ public class CodeAuthenticationMechanism extends AbstractOidcAuthenticationMecha
     }
 
     public CompletionStage<ChallengeData> getChallenge(RoutingContext context, DefaultTenantConfigResolver resolver) {
-        removeCookie(context, SESSION_COOKIE_NAME);
-
         TenantConfigContext configContext = resolver.resolve(context, false);
+        removeCookie(context, configContext, SESSION_COOKIE_NAME);
 
         ChallengeData challenge;
         JsonObject params = new JsonObject();
@@ -170,10 +170,10 @@ public class CodeAuthenticationMechanism extends AbstractOidcAuthenticationMecha
                     return cf;
                 }
                 // The original request path does not have to be restored, the state cookie is no longer needed
-                removeCookie(context, STATE_COOKIE_NAME);
+                removeCookie(context, configContext, STATE_COOKIE_NAME);
             } else {
                 // Local redirect restoring the original request path, the state cookie is no longer needed
-                removeCookie(context, STATE_COOKIE_NAME);
+                removeCookie(context, configContext, STATE_COOKIE_NAME);
             }
         } else {
             // State cookie must be available to minimize the risk of CSRF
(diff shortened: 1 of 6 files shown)


stuartwdouglas:
Any chance of a test for this?

sberyozkin:
@stuartwdouglas Sure, let me think how to do it

extensions/oidc/runtime/src/main/java/io/quarkus/oidc/runtime/CodeAuthenticationMechanism.java
```diff
@@ -279,7 +279,14 @@ private String buildRedirectUri(RoutingContext context, URI absoluteUri, String
                 .toString();
     }
 
-    private Cookie removeCookie(RoutingContext context, String cookieName) {
-        return context.response().removeCookie(cookieName, true);
+    private void removeCookie(RoutingContext context, TenantConfigContext configContext, String cookieName) {
+        ServerCookie cookie = (ServerCookie) context.cookieMap().get(cookieName);
+        if (cookie != null) {
+            cookie.setMaxAge(0);
```
gsmet, line 285: If you want the cookie to be invalidated, why don't you also reset the value?

stuartwdouglas: Because to clear a cookie you set max age 0. The value has no effect in this case.

gsmet: yes, I know. Just a bit useless to send the value over the wire IMHO.

sberyozkin: Hi @gsmet @stuartwdouglas yes, I've seen it being suggested, for the moment I've just reused the code from Vertx and added a path, but I suppose I can set its value as empty. @lodrantl mentioned it is less obvious the cookie has been invalidated when tracing the network requests if the name remains

pedroigor:
@sberyozkin LGTM. Thanks for fixing this, we indeed missed the cookie removal after adding the `cookie-path` option.

sberyozkin:
Hi @pedroigor, All

I've updated the session cookie `max-age` to the `exp` - `iat`, though this in itself does not fix any problems, as Pedro reminded we don't align the cookie itself with the login validity but the content of the cookie (idtoken), but this update should be ok anyway as it will recycle the cookie better and might even be marginally more performing if the browser proactively removes the cookie given that the Vertx token verification involves the signature recheck plus the `exp` vs `iat` check.

The only problem is that I can't get the updated test failing, the one which checks the cookie path is `web-app2` - there, after a logout, the session cookie is asserted to be null - and I've verified that even without setting the path on the removed cookie (in `CodeAuthenticationMechansim`) it still passes (thanks Pedro for asking to verify it).

So, while the fix seems good, it would be have an independent confirmation. It may well be that setting the `max-age` to the value of the id token lifespan fixes the problem in its own way, irrespectively of the cookie path property :-)

@lodrantl can you please experiment with this PR and check if it fixes the problem for you ? IMHO we can go ahead and merge once the review is complete but if you can confirm as well then it would be good, thanks

pedroigor (approved):
I managed to reproduce, and your changes seem to work.

But it would be nice if the reporter could try it out and make sure it works for him too.

sberyozkin:
Hi @pedroigor Thanks for spending the time on it. Sure lets wait for the reporter to check, though I'm happy enough that you have confirmed independently.
By the way there is one more thing I forgot here, when Vetx checks if the token has expired it can take a leeway into consideration so that is the only difference that remains to be aligned.

lodrantl:
Thanks @pedroigor  and @sberyozkin. Tested it with the PR branch and it works as expected.

gsmet:
What's the status of this one?

sberyozkin:
@lodrantl Thanks
@gsmet IMHO it is good to go as we've had @pedroigor and @lodrantl verifying this code independently.
We need 1 approval I guess, please approve or we'll need to wait for Stuart

sberyozkin:
@gsmet actually I see Pedro @pedroigor has already approved so we have 1 approval

stuartwdouglas (approved):

sberyozkin:
@stuartwdouglas, all, thanks, let me merge now

Issue #8425, opened 2020-04-06
Creates Documentation for Vault TOTP


lordofthejars:
**Description**

Creates Documentation for Vault TOTP

PR #8426, opened 2020-04-06
docs(#8425): Adds Documentation for Vault TOTP secret engine

lordofthejars:


docs/src/main/asciidoc/vault.adoc
```diff
@@ -492,6 +492,45 @@ You should see:
 {password=connor}
 ----
 
+[[totp]]
+== TOTP Secrets Engine
+
+The TOTP secrets engine generates time-based credentials according to the TOTP standard.
+
+Vault supports both the generator scenario (like Google Authenticator) and the provider scenario (like the Google.com sign in).
```
sberyozkin, line 500: And replace `Vault` with a `Vault TOTP` link as well

docs/src/main/asciidoc/vault.adoc
```diff
@@ -492,6 +492,45 @@ You should see:
...
+You can read more about Vault TOTP at https://www.vaultproject.io/docs/secrets/totp/.
+
+The Vault extension also supports TOTP by integrating with TOTP secret engine by providing an `io.quarkus.vault.VaultTOTPSecretEngine` class.
```
sberyozkin, line 503: `also` can be dropped: `The Vault extension integrates with the http://...[Vault TOTP] secret engine by providing...`

sberyozkin:
Hi @lordofthejars This looks informative, thanks, I've proposed few minor updates
CC @gsmet, Hi Guillaume, please verify if what I've proposed makes sense or if it is already ready to go, etc, thanks

docs/src/main/asciidoc/vault.adoc
```diff
@@ -492,6 +492,45 @@ You should see:
 {password=connor}
 ----
 
+[[totp]]
+== TOTP Secrets Engine
+
+The TOTP secrets engine generates time-based credentials according to the TOTP standard.
```
sberyozkin, line 498: Hmm, I've lost a comment, proposal was to replace `The TOTP secrets engine` with a link, https://www.vaultproject.io/docs/secrets/totp/[VaultTOTP] and drop "You can read more about..."

lordofthejars:
@sberyozkin I think I have addressed all your comments. I have added the link in just one place as if not we were repeating the same link in three consecutive paragraphs and I think it is not necessary.

@gsmet When @sberyozkin is ready we can merge this so it can be shown in 1.4.0.CR1. Notice that the logic is already there.

sberyozkin (approved):
Thanks @lordofthejars, IMHO we can merge it now, I can do it later today unless Guillaume or Vincent will add something.

PR #8429, opened 2020-04-06
Hibernate ORM upgrade to 5.4.14.Final

Sanne:
Upgrading to Hibernate ORM `5.4.14.Final`: this version is now available in Maven Central.

Fixes #8323 

Also including a couple minor optimisations based on the classloading flamegraphs discussed on zulip - didn't separate them as they require this version of ORM.

PR #8432, opened 2020-04-06
Upgrade Hibernate ORM to 5.4.14.Final and generate proxies at build time

stuartwdouglas:
Also implements a fix for runtime proxy generation by moving it to build time.


gastaldi:
Supersedes #8429

gastaldi:
@stuartwdouglas I think it would be easier if you just pushed your commit on top of Sanne's PR. Just saying :wink:

gsmet:
@Sanne I let you review Stuart's commit?

extensions/hibernate-orm/deployment/src/main/java/io/quarkus/hibernate/orm/deployment/DeploymentProxyDefinitions.java
```diff
@@ -0,0 +1,149 @@
...
+ * an additional round trip to the database in some circumstances.
+ * Most notably we'll fail to generate such a proxy when the entity has a "final" modifier;
+ * we'll also need a default constructor.
+ * Default constructors are required beyond proxy generation, so a lack of such a constructor
+ * will have us abort the bootstrap process with a critical error.
+ * On the other hand, having the entities marked as "final" is handled gracefully, as we
+ * can simply fallback to not use the enhanced proxy for the specific entity, and because
+ * it's a common case when writing entities in Kotlin.
+ */
+public final class DeploymentProxyDefinitions {
```
stuartwdouglas, line 43: Oops, this class should not have made it into the PR, fixing now.

integration-tests/jpa-h2/src/main/java/io/quarkus/it/jpa/h2/proxy/Dog.java
```diff
@@ -0,0 +1,21 @@
...
+import org.hibernate.annotations.Proxy;
+
+@Entity
+@Proxy(proxyClass = DogProxy.class)
+@DiscriminatorValue("DOG")
+public class Dog extends Pet implements DogProxy {
+    @Override
+    public String makeNoise() {
+        return bark();
+    }
```
emmanuelbernard, line 15: maybe add a dog specific property (getFavoriteToy())

emmanuelbernard (dismissed):
I've asked for minor changes and clarifications. The rest looks good to me.

extensions/hibernate-orm/runtime/src/main/java/io/quarkus/hibernate/orm/runtime/proxies/ProxyDefinitions.java
```diff
@@ -106,8 +108,45 @@ private static Class generateProxyClass(PersistentClass persistentClass, ByteBud
...
+                } catch (ClassNotFoundException e) {
+                    //should never happen
+                    throw new RuntimeException("Unable to load proxy class", e);
+                }
+            }
+        }
+
+        if (preGeneratedProxy == null) {
+            if (match) {
+                LOGGER.warnf("Unable to find a build time generated proxy for entity %s",
```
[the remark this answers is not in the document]
emmanuelbernard: meaning add that to the message.

extensions/hibernate-orm/runtime/src/main/java/io/quarkus/hibernate/orm/runtime/proxies/ProxyDefinitions.java
```diff
@@ -106,8 +108,45 @@ private static Class generateProxyClass(PersistentClass persistentClass, ByteBud
...
+                        persistentClass.getClassName());
+            } else {
+                LOGGER.errorf(
+                        "Unable to use a build time generated proxy for entity %s, as the build time proxy interfaces are different to the runtime ones. This should not happen.",
```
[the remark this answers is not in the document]
emmanuelbernard: meaning add that to the message.

extensions/hibernate-orm/deployment/src/main/java/io/quarkus/hibernate/orm/deployment/HibernateOrmProcessor.java
```diff
@@ -256,9 +269,89 @@ public void build(RecorderContext recorderContext, HibernateOrmRecorder recorder
...
+
+            for (String entity : entityClassNames) {
+                Set<Class<?>> proxyInterfaces = new HashSet<>();
+                proxyInterfaces.add(HibernateProxy.class); //always added
+                Class<?> mappedClass = Class.forName(entity, false, Thread.currentThread().getContextClassLoader());
+                String proxy = proxyAnnotations.get(entity);
+                if (proxy != null) {
+                    proxyInterfaces.add(Class.forName(proxy, false, Thread.currentThread().getContextClassLoader()));
+                } else {
+                    //if there is no @Proxy we need to make sure the actual class is proxiable
```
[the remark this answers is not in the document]
Sanne: it's correct, as it mimicks the current semantics: we won't generate them for final classes or entities lacking a default ctor. Appropriate warnings / failures are generated later.

extensions/hibernate-orm/deployment/src/main/java/io/quarkus/hibernate/orm/deployment/HibernateOrmProcessor.java
```diff
@@ -256,9 +269,89 @@ public void build(RecorderContext recorderContext, HibernateOrmRecorder recorder
                     .add((Class<? extends ServiceContributor>) recorderContext.classProxy(serviceContributorClassName));
         }
 
+        Set<String> entitiesToGenerateProxiesFor = new HashSet<>(domainObjects.getEntityClassNames());
```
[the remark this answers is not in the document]
Sanne: I don't see why we'd need it? The class is hopefully already extending its mappedSuperClass - or it's an user error.

extensions/hibernate-orm/deployment/src/main/java/io/quarkus/hibernate/orm/deployment/HibernateOrmProcessor.java
```diff
@@ -256,9 +269,89 @@ public void build(RecorderContext recorderContext, HibernateOrmRecorder recorder
...
+        for (ParsedPersistenceXmlDescriptor unit : allDescriptors) {
+            entitiesToGenerateProxiesFor.addAll(unit.getManagedClassNames());
+        }
+
+        PreGeneratedProxies proxyDefinitions = generatedProxies(entitiesToGenerateProxiesFor, compositeIndex,
```
Sanne, line 277: I know it's horrible, but given the context I think I'd prefer this method to be surrounded by a catch Runtime exception / log warning ?

That's to ensure that any unforeseen issue with those proxy generation will indeed get us the "falllback the current state" behaviour we discussed.

(Especially as the method intentionally throws RuntimeException on any CNFE)

extensions/hibernate-orm/runtime/src/main/java/io/quarkus/hibernate/orm/runtime/proxies/ProxyDefinitions.java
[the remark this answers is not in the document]
Sanne: +1

Also could we include the details of the mismatch, if any. I think we're assuming that this will never actually happen, but if it does it would be super useful to know the class which was expected, if we had that class at all registered in any `PreGeneratedProxies` , and if we had it what's the interfaces mismatch.

extensions/hibernate-orm/runtime/src/main/java/io/quarkus/hibernate/orm/runtime/proxies/ProxyDefinitions.java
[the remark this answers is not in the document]
Sanne: (same content as above : we'll need to know the details of the mismatch)

Sanne (changes requested):

extensions/hibernate-orm/deployment/src/main/java/io/quarkus/hibernate/orm/deployment/HibernateOrmProcessor.java
Sanne: p.s. as a fallback of course register an empty `PreGeneratedProxies`

extensions/hibernate-orm/runtime/src/main/java/io/quarkus/hibernate/orm/runtime/proxies/ProxyDefinitions.java
```diff
@@ -49,7 +49,7 @@ private ProxyDefinitions(Map<Class<?>, ProxyClassDetailsHolder> proxyDefinitionM
         this.proxyDefinitionMap = proxyDefinitionMap;
     }
 
-    public static ProxyDefinitions createFromMetadata(Metadata storeableMetadata) {
+    public static ProxyDefinitions createFromMetadata(Metadata storeableMetadata, PreGeneratedProxies preGeneratedProxies) {
         //Check upfront for any need across all metadata: would be nice to avoid initializing the Bytecode provider.
```
Sanne, line 53: There's an important optimisation missing here: even if all proxies are generated upfront, we still initialize the Bytecode Provider!

The line `final BytecodeProviderImpl bytecodeProvider = new BytecodeProviderImpl();` needs to become lazily evaulates if we aim to skip loading all those classes.

extensions/hibernate-orm/deployment/src/main/java/io/quarkus/hibernate/orm/deployment/HibernateOrmProcessor.java
```diff
@@ -256,9 +269,89 @@ public void build(RecorderContext recorderContext, HibernateOrmRecorder recorder
...
+                    if (Modifier.isFinal(mappedClass.getModifiers())) {
+                        continue;
+                    }
+                    try {
+                        mappedClass.getConstructor();
+                    } catch (NoSuchMethodException e) {
+                        continue;
+                    }
+                }
+                for (ClassInfo subclass : combinedIndex.getAllKnownSubclasses(DotName.createSimple(entity))) {
```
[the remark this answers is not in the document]
stuartwdouglas: You can't enlist the subclasses themselves, as you can't inherit from multiple classes. From what I could see in the hibernate code this is all that is registered.

extensions/hibernate-orm/deployment/src/main/java/io/quarkus/hibernate/orm/deployment/HibernateOrmProcessor.java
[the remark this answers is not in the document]
stuartwdouglas: yea, this would have no effect.

extensions/hibernate-orm/deployment/src/main/java/io/quarkus/hibernate/orm/deployment/HibernateOrmProcessor.java
stuartwdouglas: Why does this get such special treatment? Any PR that normally included such a try/catch because the author was not sure if would work would get rejected. If there is an NPE or some other problem then IMHO this is no different to any other bug.

stuartwdouglas: @Sanne maybe we should ask for broader review from the hibernate team if you are worried there are cases I have missed?

extensions/hibernate-orm/runtime/src/main/java/io/quarkus/hibernate/orm/runtime/proxies/ProxyDefinitions.java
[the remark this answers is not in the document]
stuartwdouglas: I thought this could happen with Envars? This particular WARN would only happen if an entity was registered we don't know about.

extensions/hibernate-orm/deployment/src/main/java/io/quarkus/hibernate/orm/deployment/HibernateOrmProcessor.java
Sanne: yes I would generally react like you say. but we decided to go very cautious here and promised that _any problem_ this would fallback to the previous, slower behaviour gracefully.

As you prefer though, I don't consider it blocking.

Sanne: > @Sanne maybe we should ask for broader review from the hibernate team if you are worried there are cases I have missed?

because we either will get a timely answer but not entirely sure, or it will take much longer.

stuartwdouglas: Against my better judgement I have added it.

stuartwdouglas:
@Sanne @emmanuelbernard I have addressed all the comments, except the 'open an issue' request in the WARN message. Sanne suggested last night that this code path would happen for envars users, so if we have known use cases where it will happen then I don't think there is any point in asking users to open an issue.

gsmet:
@Sanne @stuartwdouglas please ping me on Zulip when this one is merged as I could consider starting the release process tonight my time.

Sanne:
> @Sanne @emmanuelbernard I have addressed all the comments, except the 'open an issue' request in the WARN message. Sanne suggested last night that this code path would happen for envars users, so if we have known use cases where it will happen then I don't think there is any point in asking users to open an issue.

Just to clarify, my example with Envers was probably confusing. Envers is an example of an ORM extension for which the metadata gets modified beyond the annotations we find in Jandex - but also it happenso to be a special case for which we actually don't use these particular proxies (for Emmanuel: envers auditing relations are mapped as non-pojo mappers)

So we wouldn't end up logging this for Envers users I believe, but fair enought about not being that explicit.

Sanne:
Thanks all, looks great on surface. I'll do some runtime tests and if I have time some profiling, then merge.

rsvoboda:
https://github.com/hibernate/hibernate-orm/compare/5.4.13...5.4.14 .. .there are quite a few changes ... @gsmet / @Sanne anything risky from your PoV ?

gsmet:
You have to thank Yoann for that. AFAICS most of the commits are related to the Gradle upgrade. It's just that Yoann likes his commits in small pieces.

I'll let @Sanne  answer to you though as I haven't followed everything.

extensions/hibernate-orm/deployment/src/main/java/io/quarkus/hibernate/orm/deployment/HibernateOrmProcessor.java
emmanuelbernard: Created https://github.com/quarkusio/quarkus/issues/8447 to remove it in 1.5

Sanne: thank you both!

Sanne: > Created #8447 to remove it in 1.5

why not in 1.4 ?

Sanne:
@rsvoboda like Guillaume mentioned, most changes relate to Gradle being updated. Then most other changes are the actual fixes we need, with the exception of 2 commits:
 - an Envers fix (does not affect ORM)
 - a contributed fix for HHH-13889 - seems good to me

Sanne (approved):

PR #8433, opened 2020-04-07
Pass args for command-mode application to dev-mode.

### LINKED ISSUE #8428: Command-mode applications in Gradle quarkusDev mode should allow passing commandline args
**Description**
This is in regards to command-mode application-support implemented by https://github.com/quarkusio/quarkus/pull/7681.
I have been already playing around with the feature based on a Gradle-project.
Running "gradle quarkusDev" allows to run the command-mode application in dev-mode; however, for command-mode applications it would be useful to pass command-line args to the application.
For the standard gradle-application-plugin, the approach is to pass args as "gradle run --args='--arg=1'". I think it would be good to have the exact same parameter to make it easy to comprehend.

Obviously, something similar needs to be provided for Maven, though my interest current is on Gradle.

38leinaD:
Fixes #8428
For Maven and Gradle.

Usage:
gardle quarkusDev --quarkus-args='arg1=1 arg2=2'
mvn quarkus:dev -Dquarkus.args='arg1=1 arg2=2'

### DIFF
--- a/core/runtime/src/main/java/io/quarkus/runtime/CommandLineRuntimeConfig.java
+++ b/core/runtime/src/main/java/io/quarkus/runtime/CommandLineRuntimeConfig.java
@@ -1,0 +1,23 @@
+package io.quarkus.runtime;
+
+import java.util.Optional;
+
+import io.quarkus.runtime.annotations.ConfigItem;
+import io.quarkus.runtime.annotations.ConfigPhase;
+import io.quarkus.runtime.annotations.ConfigRoot;
+
+/**
+ * This configuration class is here to avoid warnings when using {@code -Dquarkus.args=...}.
+ */
+@ConfigRoot(name = ConfigItem.PARENT, phase = ConfigPhase.RUN_TIME)
+public class CommandLineRuntimeConfig {
+
+    /**
+     * The arguments passed to the command line.
+     * <p>
+     * We don't make it a list as the args are separated by a space, not a comma.
+     */
+    @ConfigItem
+    public Optional<String> args;
+
+}
--- a/devtools/maven/src/main/java/io/quarkus/maven/DevMojo.java
+++ b/devtools/maven/src/main/java/io/quarkus/maven/DevMojo.java
@@ -49,6 +49,7 @@ import org.apache.maven.plugins.annotations.Mojo;
 import org.apache.maven.plugins.annotations.Parameter;
 import org.apache.maven.plugins.annotations.ResolutionScope;
 import org.apache.maven.project.MavenProject;
+import org.apache.maven.shared.utils.cli.CommandLineUtils;
 import org.codehaus.plexus.util.xml.Xpp3Dom;
 import org.eclipse.aether.RepositorySystem;
 import org.eclipse.aether.RepositorySystemSession;
@@ -183,6 +184,9 @@ public class DevMojo extends AbstractMojo {
 
     @Parameter(defaultValue = "${jvm.args}")
     private String jvmArgs;
+
+    @Parameter(defaultValue = "${quarkus.args}")
+    private String argsString;
 
     @Parameter(defaultValue = "${session}")
     private MavenSession session;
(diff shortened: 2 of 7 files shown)


maxandersen:
if this changes arguments for gradle can you include updates to the guide ? 

should also see if "args" is a good name to use here - should match what we would do for maven so its not too different.

38leinaD:
my intention for using "--args" was that this is the way i would expect it to work with gradle as the application-plugin does it the same way: https://docs.gradle.org/current/userguide/application_plugin.html
On the other hand, it might be different than the expected way in Maven.
So, I am not sure which view is more important and confuses people the least.
Thoughts are welcome; maybe someone can provide input what what mechanism would be expected on the Maven side...

maxandersen:
In java world the corresponding would be `-Dexec.args="argument1"` according to java exec plugin https://www.mojohaus.org/exec-maven-plugin/usage.html

then it would be `-Dquarkus.args` ? not sure if that conflicts with anything. @aloubyansky are doing anything with passing -Dquarkus.*  forward to quarkus apps ?

38leinaD:
I have added integration to Maven and documentation.
Taken your suggestion to use -Dquarkus.args for Maven and --quarkus.args for Gradle.
This seems to be the convention for other arguments as well. E.g. -Djvm.args for Maven and --jvm-args for Gradle.

devtools/maven/src/main/java/io/quarkus/maven/DevMojo.java
```diff
@@ -640,7 +643,8 @@ void prepare() throws Exception {
             }
             args.add("-jar");
             args.add(tempFile.getAbsolutePath());
-
+            if (argsString != null)
+                args.addAll(Arrays.asList(argsString.split(" ")));
```
stuartwdouglas, line 647: This does not really deal with escaping, but I guess we can handle that later.

stuartwdouglas (approved):
Even though the maven version does not handle escaping IMHO it is better to get this in now and improve it in a later PR.

devtools/maven/src/main/java/io/quarkus/maven/DevMojo.java
38leinaD: I have dug up the correct way to escape from the JavaExecMojo.
Should be properly escaped now.

38leinaD:
Feel free to merge.

stuartwdouglas:
Does the maven version give a warning on startup? If the system property is propagated to the running process then we will log a warning about an unrecognised configuration key.

38leinaD:
Yes, you are right. What would be the best way to fix this?
2020-04-09 20:00:20,610 WARN  [io.qua.config] (Quarkus Main Thread) Unrecognized configuration key "quarkus.args" was provided; it will be ignored

gsmet:
@38leinaD I pushed an additional commit to your PR. Could you pull and test it?

38leinaD:
@gsmet nice. it works!

core/runtime/src/main/java/io/quarkus/runtime/CommandLineRuntimeConfig.java
```diff
@@ -0,0 +1,23 @@
+package io.quarkus.runtime;
+
+import java.util.Optional;
+
+import io.quarkus.runtime.annotations.ConfigItem;
+import io.quarkus.runtime.annotations.ConfigPhase;
+import io.quarkus.runtime.annotations.ConfigRoot;
+
+/**
+ * This configuration class is here to avoid warnings when using {@code -Dquarkus.args=...}.
```
maxandersen, line 10: is this really the right thing to do ?  

maybe we should rename it to something different at maven level so there is no collision ? 

having args as just a string seems counter to how we otherwise treat args in rest of quarkus (as a String[] or eventually a List<String>). ?

maxandersen (changes requested):
having 'quarkus.args' added as just long string in our config model seems wrong to me.

is there not a better way to handle that ?

core/runtime/src/main/java/io/quarkus/runtime/CommandLineRuntimeConfig.java
gsmet: We do that a lot.

As for string vs list, see the comment, I can't be more explicit.

As for all the other knobs we introduced to avoid the warnings, they are not supposed to be used. That's what we decided a long time ago rather than having an ignore list (which we had at some point).

We could introduce a special converter to split the things in the same way Maven is doing things but I don't think it's worth it. And we would probably have to copy Maven code to be exactly on par.

BTW, I think we need that PR for 1.4 so blocking the PR just for that seems counterproductive.

PR #8352, opened 2020-04-02
Move the Security API to Mutiny

stuartwdouglas:

PR #8434, opened 2020-04-07
Update to Mutiny 0.4.4

cescoffier:


cescoffier:
@stuartwdouglas With this PR, you should be able to simplify a bit the security PR (#8352). 
You can create a Uni from `null` using:

```
Uni<Void> uni = Uni.createFrom().nullItem();
Uni<String> noString = Uni.createFrom().nullItem();
```

stuartwdouglas:
I already updated the PR to use that. I think it was missing from the version that I used when I started the PR, but after I rebased it was available.

gsmet (approved):

PR #8449, opened 2020-04-07
Remove openssl from docker file and enforce image name

cescoffier:


gastaldi (approved):

cescoffier:
I had to remove the last commit because of the platform.

This last commit was fixing a real bug as image names with uppercase are not accepted (and the error message is very obscure). 

We cannot modify the template and add a new variable because of old versions that still use the new templates (and so would fail as this variable would not be known). IMHO it's a big limitation of the platform and a blocker as basically we cannot fix the template or improve them.

gsmet (approved):

PR #8453, opened 2020-04-07
javax.imageio is now supported in GraalVM 19.3+

gastaldi:
As reported in https://quarkusio.zulipchat.com/#narrow/stream/187030-users/topic/ImageIO.20is.20not.20supported.20when.20using.20native.20image/near/193211218


gsmet:
But we still support 19.3 AFAIK so I'm not sure what we should do with it.

gastaldi:
I can confirm it works in GraalVM 19.3.1 too

gsmet:
Woot! Thanks for revisiting this!

dmlloyd (approved):

murphye:
I tested this with 1.4.0.CR1 and GraalVM 20.

ImageIO is unblocked, but many things will still not work for ImageIO. For example, JPEGImageReader plugin is not linked. This will likely be supported in the future.

So for my use case, using ImageIO for reading in image doesn't work now.

PR #8458, opened 2020-04-07
Add deprecate warning for buildNative task

glefloch:
@gastaldi, here is a pull request for the documentation. I was thinking of adding all available native properties. Something like https://quarkus.io/guides/building-native-image#configuration-reference

WDYT? Should I update the other pull request with something like that?

### DIFF
--- a/docs/src/main/asciidoc/gradle-tooling.adoc
+++ b/docs/src/main/asciidoc/gradle-tooling.adoc
@@ -264,6 +264,13 @@ Make sure to have `GRAALVM_HOME` configured and pointing to GraalVM version {gra
 Create a native executable using: `./gradlew buildNative`.
 A native executable will be present in `build/`.
 
+[WARNING]
+====
+The `buildNative` task has been deprecated in favor of
+`build -Dquarkus.package.type=native`.
+Native related properties can either be added in `application.properties` file or as command line arguments.
+====
+
 === Build a container friendly executable
 
 The native executable will be specific to your operating system.


docs/src/main/asciidoc/gradle-tooling.adoc
```diff
@@ -264,6 +264,13 @@ Make sure to have `GRAALVM_HOME` configured and pointing to GraalVM version {gra
 Create a native executable using: `./gradlew buildNative`.
 A native executable will be present in `build/`.
 
+[WARNING]
+====
+The `buildNative` task has been deprecated in favor of task +
```
gastaldi, line 269:
```suggestion
The `buildNative` task has been deprecated in favor of
```

gastaldi (changes requested):
I added a minor suggestion, see if that makes sense. 

You can add the following snippet to the documentation if you want all available native properties as a table: 

```asciidoc
include::{generated-dir}/config/quarkus-native-pkg-native-config.adoc[opts=optional]
```

glefloch:
@gastaldi thanks for the review, it totally makes sense. I updated the doc. 
Thanks for the snippet, I will add it in the other pull request.

gastaldi (approved):
LGTM

geoand (approved):

andreas-eberle:
What was the reasoning behind the deprecation of the buildNative task? If I have a gradle project with many microservices, I was able to use the buildNative task to configure common properties for all of them in the parent gradle file. Now I'll have to it for every module/every microservice.

PR #8459, opened 2020-04-07
Scheduler - add configurable delay

### LINKED ISSUE #8314: Scheduler - Configurable Init Delay
**Description**
In the @Scheduled annotation, it is possible to specify a delay attribute, which will cause the scheduled method to not initially run until that delay has expired. It is defined as follows:

```java
    /**
     * Delays the time the trigger should start at. The value is rounded to full second.
     * <p>
     * By default, the trigger starts when registered.
     *
     * @return the initial delay
     */
    long delay() default 0;

    /**
     * 
     * @return the unit of initial delay
     */
    TimeUnit delayUnit() default TimeUnit.MINUTES;
```

This feature is working as is, but requires that the delay be defined in a hardcoded fashion when called. 

It seems like a reasonable ask to request that this variable be parsed similarly to the **cronExpr** or **every** attributes such that it can be specified as a configurable value. 

**Implementation ideas**
If we treat the delay attribute as a String, it should be possible to use an existing check in SimpleScheduler to see if it is a configuration or not, and then parse that value as a Long. 

Proposed changes:
```java
// In @Scheduled
String delay() default "0";

// In SimpleScheduler
SimpleTrigger createTrigger(String invokerClass, CronParser parser, Scheduled scheduled) {
    String id = triggerNameSequence.getAndIncrement() + "_" + invokerClass;
    ZonedDateTime start = ZonedDateTime.now().truncatedTo(ChronoUnit.SECONDS);

    long delay = getDelay(scheduled);
    if (delay > 0) {
        start = start.toInstant().plusMillis(scheduled.delayUnit().toMillis(delay).atZone(start.getZone());
    }
    ....
}

/**
 * Retrieve the delay number from the Scheduled annotation. If it is a configuration, retrieve the configuration value
 * @param scheduled The annotation defined on the method, and its attributes
 * @return delay as a long
 * @throws NumberFormatException if the resoved delay value is not a number
 */
long getDelay(Scheduled scheduled) throws NumberFormatException {
    String delay = scheduled.delay().trim();
    if (SchedulerSupport.isConfigValue(delay)) {
        delay = ConfigProviderResolver.instance().getConfig().getValue(SchedulerSupport.getConfigProperty(delay)), String.class);
    }
    return Long.parseLong(delay);
}
```

mkouba:
- resolves #8314
- also introduce the SchedulerContext synthetic bean and refactor SchedulerProcessor

### DIFF
--- a/extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/ValidationPhaseBuildItem.java
+++ b/extensions/arc/deployment/src/main/java/io/quarkus/arc/deployment/ValidationPhaseBuildItem.java
@@ -45,6 +45,10 @@ public final class ValidationPhaseBuildItem extends SimpleBuildItem {
             this.values = Arrays.asList(values);
         }
 
+        public ValidationErrorBuildItem(List<Throwable> values) {
+            this.values = values;
+        }
+
         public List<Throwable> getValues() {
             return values;
         }
--- a/extensions/quartz/deployment/src/test/java/io/quarkus/quartz/test/InvalidDelayedExpressionTest.java
+++ b/extensions/quartz/deployment/src/test/java/io/quarkus/quartz/test/InvalidDelayedExpressionTest.java
@@ -1,0 +1,33 @@
+package io.quarkus.quartz.test;
+
+import javax.enterprise.inject.spi.DeploymentException;
+
+import org.jboss.shrinkwrap.api.ShrinkWrap;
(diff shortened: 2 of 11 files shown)


mkouba:
FYI @samjdacanay

samjdacanay:
Awesome!

gastaldi (approved):
LGTM

PR #8468, opened 2020-04-08
Add ReflectionValueResolver to DefaultValueResolvers.

xiaodong-xie:
The issue I run into was like this:

```
public class SimpleTest {
    public static void main(String[] args) {
        final Engine engine = Engine.builder().addDefaults().addValueResolver(new ReflectionValueResolver()).build();
        final Template template =
            engine.parse("{single} -> {single.name}");
        final String result = template
            .data("single", new Single("some name"))
            .render();
        System.out.println(result);
    }

    public static class Single {
        public String name;

        public Single(final String name) {
            this.name = name;
        }

        @Override
        public String toString() {
            return "Single{" +
                "name='" + name + '\'' +
                '}';
        }
    }
}
```

Without `.addValueResolver(new ReflectionValueResolver())`, the result
would be `Single{name='some name'} -> NOT_FOUND`.

Or we could change the documentation, mentioning that when a Java bean is
used, one needs to add `.addValueResolver(new ReflectionValueResolver())`.

Please let me know how I could continue with this. Thanks.

### DIFF
--- a/independent-projects/qute/core/src/main/java/io/quarkus/qute/EngineBuilder.java
+++ b/independent-projects/qute/core/src/main/java/io/quarkus/qute/EngineBuilder.java
@@ -85,7 +85,7 @@ public final class EngineBuilder {
      */
     public EngineBuilder addDefaultValueResolvers() {
         return addValueResolvers(mapResolver(), mapperResolver(), mapEntryResolver(), collectionResolver(),
-                thisResolver(), orResolver(), trueResolver());
+                thisResolver(), orResolver(), trueResolver(), new ReflectionValueResolver());
     }
 
     public EngineBuilder addDefaults() {


mkouba:
So the `ReflectionValueResolver` is really just a _fallback_ that is automatically registered in quarkus and it would only work in the JVM mode (you would need `@RegisterForReflection` for native). You should not need it when using type-safe templates, `@TemplateData` and template extension methods (i.e. idiomatic constructs). 

I don't think we should make it part of the defaults or even mention it in the docs ;-).

@xiaodong-xie are you tryting to use qute standalone?

xiaodong-xie:
Thanks for the info and comments, @mkouba.

I assume in the description of this PR, I'm using `qute` standalone?

mkouba:
> I assume in the description of this PR, I'm using qute standalone?

Ah, sorry I missed the test source. Yes, you're using it standalone which currently something we don't expect the users will do ;-).

xiaodong-xie:
Ah, OK. Thanks for the info, @mkouba 🙏 👍 

Then it's clear. Sorry for the trouble, I'll close. 😄

mkouba:
Np, thanks for your interest in quarkus and qute! ;-)

PR #8471, opened 2020-04-08
remove call to oc version

### LINKED ISSUE #7637: the `container-image-s2i` extension runs `oc version` without reason
**Describe the bug**

I added the `container-image-s2i` extension so that I can customize the base S2I image in generated OpenShift resources. (When the extension is not present, the base S2I image is always `fabric8/s2i-java`, even though I added the `quarkus.s2i.base-jvm-image` config property. I personally consider the `fabric8/s2i-java` image long deprecated and dangerous to use, as it clearly hasn't been kept up to date.)

That works fine, if the `oc` binary is present. (If the `oc` binary is not present, I still get all the resources, but again, I always get `fabric8/s2i-java`, even though I added the `quarkus.s2i.base-jvm-image` config property.)

However, there's a noticeable delay as it's running `oc version` to determine OpenShift version. This is because `oc version` by default talks to the OpenShift API server, which takes time.

(The process of determining server version is actually tailored to `oc` from OpenShift 3.x. The `oc` binary from OpenShift 4.x no longer prints a line starting with `kubernetes`. Also the server version is not even used!)

**Expected behavior**
I don't know why I even need the `container-image-s2i` extension to get S2I configuration applied, because the old deprecated `s2i.builder-image` config property works without the extension!

But even if I add the `container-image-s2i` extension, I expect the S2I configuration to be applied when I don't have the `oc` binary present, and I certainly don't expect the build to talk to OpenShift if I don't ask it to. (If I ask to push the container image to the OpenShift registry, or start a build in OpenShift, then sure. But not for creating the YAML files locally!)

**Actual behavior**
The build process talks to OpenShift server needlessly.

**To Reproduce**
1. Add the `container-image-s2i` extension.
2. Run `mvn clean package`.
3. If `oc` is present: observe noticeable delay. If `oc` is not present: observe that S2I configuration is not taken into account.

**Configuration**
```properties
quarkus.kubernetes.deployment-target=openshift
quarkus.openshift.expose=true
quarkus.s2i.base-jvm-image=registry.access.redhat.com/openjdk/openjdk-11-rhel7
```

glefloch:
This pull-request removes the call to `oc version` . 
@geoand @iocanel @Ladicek can you review it?

close #7637

### DIFF
--- a/extensions/container-image/container-image-s2i/deployment/src/main/java/io/quarkus/container/image/s2i/deployment/S2iBuild.java
+++ b/extensions/container-image/container-image-s2i/deployment/src/main/java/io/quarkus/container/image/s2i/deployment/S2iBuild.java
@@ -1,24 +1,10 @@
 package io.quarkus.container.image.s2i.deployment;
 
-import java.io.BufferedReader;
-import java.io.File;
-import java.io.IOException;
-import java.io.InputStream;
-import java.io.InputStreamReader;
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Optional;
 import java.util.function.BooleanSupplier;
-import java.util.function.Function;
-
-import org.jboss.logging.Logger;
 
 import io.quarkus.container.image.deployment.ContainerImageConfig;
-import io.quarkus.deployment.util.ExecUtil;
 
 public class S2iBuild implements BooleanSupplier {
-
-    private static final Logger LOGGER = Logger.getLogger(S2iBuild.class.getName());
 
     private ContainerImageConfig containerImageConfig;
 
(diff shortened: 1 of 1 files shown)


Ladicek (approved):
I personally think this is exactly how it should look like :-)

geoand (approved):
+1 from me.

@iocanel has the final say

gsmet:
I would suggest we replace more complicated with `return true` in Quarkus. Looks like good simplification and optimization :).

geoand:
Would do wonders for both performance and maintenance :P

Ladicek:
That's the zen of programming: each program can be simplified down to `return true`. Obviously each program has a bug, so...

iocanel (approved):
+1 for removing the `oc version` thing.

Not 100% sure if just replacing it with true is the way to go.

As, its a clear improvement on what we already have and lacking a better alternative I'll say `LGTM`!

geoand:
OK, let's get it in now and we can improve if needed later.

PR #8480, opened 2020-04-08
Do not trigger CI for assigned and edited PRs

gastaldi:


gsmet:
The edited one was about having WIP in the title and removing it. But I would be in favor of dropping the WIP thing and using draft request always now that we can make them draft afterwards.

gsmet (approved):

PR #8484, opened 2020-04-08
Include native resources via application.properties

### LINKED ISSUE #8437: Include native resources via application.properties
As discussed on the mailing list [1] We should introduce a new build time application.properties config that would allow users to add resources to the native image. Something like

```
quarkus.native.resources.includes = my-resources/**
quarkus.native.resources.excludes = my-resources/**/*.ignored
```

Let me prepare a PoC.

[1] https://groups.google.com/forum/#!topic/quarkus-dev/SLuS7DiDMVA

ppalaga:
…cation.properties

Fix #8437

### DIFF
--- a/integration-tests/main/src/main/resources/application.properties
+++ b/integration-tests/main/src/main/resources/application.properties
@@ -37,3 +37,5 @@ configproperties.message=Hello
 configproperties.other=1
 
 quarkus.swagger-ui.always-include=true
+
+quarkus.native.resources.includes = test-resources/**.txt
--- a/core/deployment/src/main/java/io/quarkus/deployment/pkg/NativeConfig.java
+++ b/core/deployment/src/main/java/io/quarkus/deployment/pkg/NativeConfig.java
@@ -4,6 +4,7 @@ import java.io.File;
 import java.util.List;
 import java.util.Optional;
 
+import io.quarkus.runtime.annotations.ConfigGroup;
 import io.quarkus.runtime.annotations.ConfigItem;
 import io.quarkus.runtime.annotations.ConfigPhase;
 import io.quarkus.runtime.annotations.ConfigRoot;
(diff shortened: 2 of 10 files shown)


ppalaga:
I tried to implement the resource scanning as efficiently as I could but I still wonder whether there is a way to avoid full tree traversal for includes starting with `**`?

Second, a potential problem with the current proposal is that it is not excluding class files by default. People wanting to include all resources might think that `**` is all they need but `**` would include also all class files. So they should include `**` and exclude `**.class`. 

I wonder whether we should exclude `**.class` by default and I wonder whether we should always add the `**.class` exclude to the excludes defined by the user. I am open to change the default behavior. 

@stuartwdouglas mentioned that we should maybe include all resources by default. I am open to that too if there is an agreement.

dmlloyd:
I'm no longer very familiar with these areas of the codebase, so I'm handing it off to @stuartwdouglas to have a look.

stuartwdouglas:
Do we actually need this includes/excludes approach vs just a straight pass through of include regex to the native image command (https://github.com/oracle/graal/blob/master/substratevm/RESOURCES.md)? 

It would not allow for excludes, but because it is regex based I don't know if this is that much of an issue. 

This PR looks really good for this type of approach, I am just not sure if maybe the simpler approach is better in this case.

ppalaga:
> Do we actually need this includes/excludes approach vs just a straight pass through of include regex to the native image command (https://github.com/oracle/graal/blob/master/substratevm/RESOURCES.md)?

Indeed, this would be much simpler to implement and maintain. 

I still think having excludes would be a big advantage from the user point of view. Maybe we could propose `-H:ExcludeResources` to native-image to have both simple implementation and flexibility?

stuartwdouglas:
If you really feel that the excludes this is necessary then we can go with this approach, but my gut feeling is that the extra complexity (plus the requirement for indexing) makes it not worth it.

ppalaga:
1893cdf implements the proposal of @stuartwdouglas

ppalaga:
I have filed https://github.com/oracle/graal/issues/2347 for `-H:ExcludeResources`.

stuartwdouglas (approved):

ppalaga:
Do I need to care for the JDK 8 JVM failure? I do not think it is related to this change because I can reproduce it locally with the base master commit.

ppalaga:
e8a9f3c rebased to see if the CI failure persists.

ppalaga:
This time a different error in different test, still looks unrelated. Do I need to care?

integration-tests/main/src/main/resources/application.properties
```diff
@@ -37,3 +37,5 @@ configproperties.message=Hello
 configproperties.other=1
 
 quarkus.swagger-ui.always-include=true
+
+quarkus.native.resources.includes = test-resources/.*\\\\.txt
```
gsmet, line 41: Oh my, this is very ugly, especially since it will be very common :/

@dmlloyd any idea?

gsmet:
@ppalaga we have some CI instability issues, sorry about that.

integration-tests/main/src/main/resources/application.properties
ppalaga: It is hard to do anything as long as we keep using regular expressions and MP config list converter.

One way out could be to use globs instead of REs. Globs are a bit less expressive than REs, but much easier to use for the 99% of common cases. With globs, the above would be `quarkus.native.resources.includes = test-resources/**.txt`.

Using a custom list converter with some different escaping mechanism could eliminate two of the four backslashes - not sure it is worth the effort.

stuartwdouglas: Maybe globs are better, what do you think @gsmet ?

ppalaga:
ff7ed60 rebased and added a commit to use globs instead of REs to define the includes.

I chose to call `sun.nio.fs.Globs.toUnixRegexPatternMethod(String) via reflection (option (a)), because I saw it as the best from the bad options. 

Alternatives I have considered:

(b) Copy the code from `sun.nio.fs.Globs` here is not an option, because the class is GPL licensed and we are AL.
(c) Depend on `com.madgag:globs-for-java` or `com.github.protobufel:globs` which both re-package `sun.nio.fs.Globs` making the relevant methods public and keeping the GPL license. While this is clean legally, I find it an overkill from the technical PoV, as long as we are not on module path. I see (c) as a fallback option once we hit issues with (a).

WDYT?

dmlloyd:
It's fairly simple to use `Pattern` to write a regular expression that can transform a glob into a regular expression IIRC.  That's how I usually tackle the issue.  Definitely don't use `sun.*` classes.

ppalaga:
9a4b341 implements a proper glob-to-regex transformer

ppalaga:
601a3f8 is addressing the comments by @dmlloyd

ppalaga:
a6fe657 rebased to trigger the CI anew.

ppalaga:
Is this good to merge now?

stuartwdouglas (approved):

stuartwdouglas:
Looks good, thanks for your patience around this.

PR #8486, opened 2020-04-09
Deprecate loadsApplicationClasses

stuartwdouglas:
After the ClassLoader change it is no longer required.

### DIFF
--- a/core/builder/src/main/java/io/quarkus/builder/BuildContext.java
+++ b/core/builder/src/main/java/io/quarkus/builder/BuildContext.java
@@ -24,12 +24,14 @@ import io.quarkus.builder.location.Location;
  * @author <a href="mailto:david.lloyd@redhat.com">David M. Lloyd</a>
  */
 public final class BuildContext {
+    private final ClassLoader classLoader;
     private final StepInfo stepInfo;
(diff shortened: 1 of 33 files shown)


core/builder/src/main/java/io/quarkus/builder/BuildContext.java
```diff
@@ -24,12 +24,14 @@
...
 public final class BuildContext {
+    private final ClassLoader classLoader;
     private final StepInfo stepInfo;
     private final Execution execution;
     private final AtomicInteger dependencies;
     private volatile boolean running;
 
-    BuildContext(final StepInfo stepInfo, final Execution execution) {
+    BuildContext(ClassLoader classLoader, final StepInfo stepInfo, final Execution execution) {
+        this.classLoader = classLoader;
```
gastaldi, line 34: Better add an assertion here to make sure this is never null, or weird errors might appear

gastaldi (approved):
LGTM, let's see what @dmlloyd says

gsmet:
@dmlloyd any chance you could have a look to that one?

dmlloyd (approved):
Seems OK to me.

PR #8284, opened 2020-03-30
Allow AppArtifact to have more than one path

aloubyansky:
Refactoring to allow AppArtifact to provide more than one path to represent its content. This is necessary, for example, to represent not yet packaged Gradle projects.

Fixes #7426 
Fixes #7675 
Fixes #8118 
Fixes #8657 
Fixes #7156 

Among other things, I also introduced `QuarkusBuildCloseablesBuildItem` to easily collect `Closeables` that should be closed at the end of the augmentation. It looked like some app archives haven't actually been closed.

PR #8489, opened 2020-04-09
Gradle: Use the Jar task output instead of folders in a multi-module project

roguexz:
When working with multimodule projects, the testRuntimeClasspath already includes the "project"-dependencies as jar files. When running tests, the Quarkus plugin adds those project outputs as folders instead of re-using the jar task. This leads to a scenario where the builder sees two identical classes coming from different locations and throws ambiguous dependency injection errors.


aloubyansky:
This is something that should be fixed in https://github.com/quarkusio/quarkus/pull/8284
I'm not closing this one until #8284 has been merged though.
Thanks for your contribution though @roguexz

gsmet:
OK, so this one should be fixed now and will be available in the upcoming 1.4. Please come back to us if your issue is not solved after the upgrade.

Thanks anyway!

PR #8516, opened 2020-04-10
Type-safe message bundles powered by Qute

mkouba:
NOTE: The docs are updated too: https://github.com/quarkusio/quarkus/pull/8516/files#diff-2cca36cb4d8997f164012dd9f0bd2d6c

...

The basic idea is that each message is potentially a very simple template.

### Example

First you need to define the default message bundle interface:

```java
@MessageBundle // the name is defaulted to "msg" and is used as a namespace in templates expressions
public interface Messages {

    @Message("Hello {name}!") // value is a qute template
    String hello_name(String name); // name is a param that can be used in the template
}
```
Quarkus generates the implementations during build and you can use the bundles:

1. directly in your code `MessageBundles.get(Messages.class).hello_name("Jachym")`
2. `@Inject Messages messages` and call the bundle methods
3. in the templates: `{msg:hello_name('Jachym')}`

### Message Keys

Keys are used in qute templates. The `@MessageBundle` can be used to define the default strategy used to generate message keys for methods. By default, the annotated element's name is used as-is. Other possibilities are (1) de-camel-cased and hyphenated; `helloName()` -> `hello-name`, (2) de-camel-cased and parts separated by underscores; `helloName()` -> `hello_name`. `@Message` can override this strategy and even define a custom key.

### Localization

You can define a localized bundle interface that extends the default interface:

```java
@Localized("cs")
public interface CzechMessages extends Messages {

    @Message("Ahoj {name}!") // localized template
    @Override
    String hello_name(String name);
}
```

Alternatively, you can create an UTF-8 encoded file located in `src/main/resources/messages`; e.g. `msg_de.properties`. The file name consists of the relevant bundle name (`msg`) and underscore followed by the locate tag (IETF).

```properties
hello_name=Hallo {name}!
```

### Validation

Message templates are validated:
1. all expressions without a namespace must map to a param; e.g. `Hello {foo}` -> the method must have a param of name `foo`
2. all expressions are validated against the types of the parameters; e.g. `Hello {foo.bar}` where the parem `foo` is of type `org.acme.Foo` -> `org.acme.Foo` must have a property of name `bar`
3. a warning is logged for each unused param

Expressions that reference a message bundle method, such as `{msg:hello(name)}`, are validated too.

### TODO

- [x] Finish working prototype + validation
- [x] Update docs
- [x] Support localized properties file
- [x] Generate qute resolvers for all parameter types
- [x] Verify the limits of the generated implementations (each impl is also a registered as a namespace resolver)
  - ~I've verified that the current impl is able to handle ~ 1000 messages; if we split the generated `resolve()` method and group messages (a strategy already used in `io.quarkus.arc.processor.ComponentsProviderGenerator`) we could handle much more~
  - after the split 5000 messages is not a problem (except that we generate a huge class of course)

### DIFF
--- a/docs/src/main/asciidoc/qute-reference.adoc
+++ b/docs/src/main/asciidoc/qute-reference.adoc
@@ -986,10 +986,101 @@ class DetailResource {
 
 In the development mode, all files located in `src/main/resources/templates` are watched for changes and modifications are immediately visible.
 
+=== Type-safe Message Bundles
+
+==== Basic Concepts
+
+The basic idea is that every message is potentially a very simple template.
+In order to prevent type errors a message is defined as an annotated method of a *message bundle interface*.
+Quarkus generates the *message bundle implementation* at build time.
+Subsequently, the bundles can be used at runtime:
+
+1. Directly in your code via `io.quarkus.qute.i18n.MessageBundles#get()`; e.g. `MessageBundles.get(AppMessages.class).hello_name("Lucie")`
+2. Injected in your beans via `@Inject`; e.g. `@Inject AppMessages`
+3. Referenced in the templates via the message bundle name; e.g. `{msg:hello_name('Lucie')}`
+
+.Message Bundle Interface Example
+[source,java]
+----
+import io.quarkus.qute.i18n.Message;
+import io.quarkus.qute.i18n.MessageBundle;
+
+@MessageBundle <1>
+public interface AppMessages {
+
+    @Message("Hello {name}!") <2>
+    String hello_name(String name); <3>
+}
+----
+<1> Denotes a message bundle interface. The bundle name is defaulted to `msg` and is used as a namespace in templates expressions, e.g. `{msg:hello_name}`.
+<2> Each method must be annotated with `@Message`. The value is a qute template.
+<3> The method parameters can be used in the template.
+
+==== Bundle Name and Message Keys
+
+Keys are used directly in templates.
+The bundle name is used as a namespace in template expressions. 
+The `@MessageBundle` can be used to define the default strategy used to generate message keys from method names. 
+However, the `@Message` can override this strategy and even define a custom key.
+By default, the annotated element's name is used as-is. 
+Other possibilities are:
+
+1. De-camel-cased and hyphenated; e.g. `helloName()` -> `hello-name`
+2. De-camel-cased and parts separated by underscores; e.g. `helloName()` -> `hello_name`. 
+
+==== Validation
+
+* All message bundle templates are validated:
+** All expressions without a namespace must map to a parameter; e.g. `Hello {foo}` -> the method must have a param of name `foo`
+** All expressions are validated against the types of the parameters; e.g. `Hello {foo.bar}` where the parameter `foo` is of type `org.acme.Foo` -> `org.acme.Foo` must have a property of name `bar`
++
+NOTE: A warning message is logged for each _unused_ parameter.
+* Expressions that reference a message bundle method, such as `{msg:hello(item.name)}`, are validated too.
+
+==== Localization
+
+The default locale of the Java Virtual Machine used to *build the application* is used for the `@MessageBundle` interface by default.
+However, the `io.quarkus.qute.i18n.MessageBundle#locale()` can be used to specify a custom locale.
+Additionaly, there are two ways to define a localized bundle:
+
+1. Create an interface that extends the default interface that is annotated with `@Localized`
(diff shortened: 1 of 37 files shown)


extensions/qute/runtime/src/main/java/io/quarkus/qute/i18n/MessageBundle.java
```diff
@@ -0,0 +1,49 @@
...
+import java.lang.annotation.Target;
+
+/**
+ * Marks a message bundle interface.
+ * 
+ * Each method of such interface represents a single message.
+ * 
+ * <pre>
+ * &#64;MessageBundle("msg")
+ * class MyBundle {
```
FroMage, line 16: should be `interface` here.

mkouba: Definitely! ;-)

FroMage:
That looks pretty cool.

Given that when we write internationalised templates we will have many many localised strings, I'd suggest to use a shorter syntax such as `{&message}` and `{&message(foo, bar)}` (doesn't matter if it's `&` or `@` or `%` or whatever is free.

mkouba:
> That looks pretty cool.

Thanks :D

> Given that when we write internationalised templates ...

The point of using namespaces here is that you can have several bundles in your app, e.g. one for web, one for emails, etc. Also the localized messages are not a core feature of qute. We could use `m` (or `msg`) as the default namespace (i.e. when you use `@MessageBundle` with no explicit value), so that you could use `{m:message(foo,bar)}` by default. That would mean one more char compared to your proposal (which is hopefully acceptable).

docs/src/main/asciidoc/qute-reference.adoc
```diff
@@ -986,10 +986,92 @@ class DetailResource {
 
 In the development mode, all files located in `src/main/resources/templates` are watched for changes and modifications are immediately visible.
 
-=== Configuration Reference
+=== Type-safe Message Bundles
+
+==== Basic Concepts
+
+The basic idea is that every message is potentially a very simple template.
+In order to prevent type-safe errors a message is defined as an annotated method of a *message bundle interface*.
```
Ladicek, line 994: "prevent type-safe errors" sounds weird :-) You probably meant just "prevent errors".

mkouba: Well, I meant "prevent type errors". Thanks for spotting this.

extensions/qute/runtime/src/main/java/io/quarkus/qute/i18n/MessageBundle.java
```diff
@@ -0,0 +1,67 @@
...
+ * }
+ * </pre>
+ */
+@Retention(RUNTIME)
+@Target(TYPE)
+public @interface MessageBundle {
+
+    /**
+     * Constant value for {@link #locale()} indicating that the default locale of the Java Virtual Machine used to build the
+     * application should be used.
```
Ladicek, line 29: I [hope that I] understand why this is, but it seems quite counter-intuitive. If it can't be the default locale of the _runtime_ JVM, then it should probably be highlighted in the user documentation more prominently.

Ladicek:
I don't feel qualified to comment on the code, but the general design seems fine. Good to see the option for externalizing localizations, as translators are really not going to edit Java code :-)

extensions/qute/runtime/src/main/java/io/quarkus/qute/i18n/MessageBundle.java
mkouba: Good point...

mkouba:
> EDIT: actually even the .properties format is full of surprises

In fact, we don't use the properties format as defined by `java.util.Properties`. The file suffix can be anything. Our format is simple: `key=value` on  each line.

mkouba:
> I don't feel qualified to comment on the code, but the general design seems fine.

@Ladicek I believe that you're more than qualified. It's no rocket science. Just shout out if there's something you don't like or don't understand ;-).

Ladicek:
> In fact, we don't use the properties format as defined by `java.util.Properties`. The file suffix can be anything. Our format is simple: `key=value` on each line. I used `.properties` because it's supported by many IDEs. I'll update the docs.

Very nice!

mkouba:
@FroMage @Ladicek @maxandersen @gastaldi @kenfinnigan If there are no objections I'd like to get this merged.

gastaldi (approved):
LGTM, do it! :+1:

PR #8521, opened 2020-04-10
Support for reusing existing kubernetes resources

iocanel:
This pull request allows the kubernetes extension to reuse existing kubernetes/openshift/knative manifests and dekorate them based on the available build items and kubernetes configuration.


geoand:
This feature is super cool!

I think we really need some tests for it though.

docs/src/main/asciidoc/deploying-to-kubernetes.adoc
```diff
@@ -735,3 +736,87 @@ This assumes that a `.kube/config` is available in your user directory that poin
...
+        imagePullPolicy: "Always"
+        name: "kubernetes-quickstart"
+        ports:
+        - containerPort: 8080
+          name: "http"
+          protocol: "TCP"
+      serviceAccount: "kubernetes-quickstart"
+---
+
+The provided, replicas, environment variables and labels where retained. However, the image and container port where modified. Also the standard annotations have been added.
```
geoand, line 819:
```suggestion
The provided replicas, environment variables and labels where retained. However, the image and container port where modified. Also the standard annotations have been added.
```

docs/src/main/asciidoc/deploying-to-kubernetes.adoc
```diff
@@ -735,3 +736,87 @@ This assumes that a `.kube/config` is available in your user directory that poin
...
 
 At the moment no additional options are provided for further customization.
+
+== Using existing resources
+
+Some times it's desirable to either provide additional resources (e.g. a ConfigMap, a Secret, a Deployment for a database etc) or provide custom ones that will be used as a `base` for the generation process.
+Those resources can be added under `src/main/kubernetes` and can be named after the target environment (e.g. kubernetes.json, openshift.json, knative.json, or the yml equivalents). Each of these files may contain one or more Kubernetes resources.
+
+Any resource found, will be added in the generated manifests. Global modifications (e.g. labels, annotations etc) will also be applied to those resources.
+If one of the provided resources has the same name with one of the generated resources, then the generated resource will be created on top of the provided, respecting existing content when possible (e.g. existing labels, annotations, environment variables, mounts, replicas etc).
```
geoand, line 746:
```suggestion
If one of the provided resources has the same name as one of the generated resources, then the generated resource will be created on top of the provided, respecting existing content when possible (e.g. existing labels, annotations, environment variables, mounts, replicas etc).
```

docs/src/main/asciidoc/deploying-to-kubernetes.adoc
```diff
@@ -735,3 +736,87 @@ This assumes that a `.kube/config` is available in your user directory that poin
...
+
+The name of the resource is determined by the application name, and may be overridden by `quarkus.kubernetes.name`, `quarkus.openshift.name` and `quarkus.knative.name`.
+
+For example, in the kubernetes-quickstart application, we can add a `kuberenetes.yml` file in the `src/main/kubernetes` that looks like:
```
geoand, line 750:
```suggestion
For example, in the kubernetes-quickstart application, we can add a `kubernetes.yml` file in the `src/main/kubernetes` that looks like:
```

docs/src/main/asciidoc/deploying-to-kubernetes.adoc
```diff
@@ -735,3 +736,87 @@ This assumes that a `.kube/config` is available in your user directory that poin
...
+        ports:
+        - containerPort: 8080
+          name: "http"
+          protocol: "TCP"
+      serviceAccount: "kubernetes-quickstart"
+---
+
+The provided, replicas, environment variables and labels where retained. However, the image and container port where modified. Also the standard annotations have been added.
+
+Note: If the resource name does not match the application name (or the overridden name) instead of reusing the resource a new one will be added. Similar is the case of the container.
```
geoand, line 821:
```suggestion
Note: If the resource name does not match the application name (or the overridden name) instead of reusing the resource a new one will be added. Same goes for the container.
```

iocanel:
@maxandersen @lordofthejars @Ladicek We have discussed about this feature in the mailing lists, so here's a first draft.j 

> This feature is super cool!
> 
> I think we really need some tests for it though.

Agree! Do the prodTests allow for specifying additional resources like `src/main/kubernetes/kubernetes.yml`?

geoand:
Unfortunately no... Although it could be extended to do so.
I would say for now just create the old style tests

docs/src/main/asciidoc/deploying-to-kubernetes.adoc
```diff
@@ -735,3 +736,87 @@ This assumes that a `.kube/config` is available in your user directory that poin
...
+        imagePullPolicy: "Always"
+        name: "kubernetes-quickstart"
+        ports:
+        - containerPort: 8080
+          name: "http"
+          protocol: "TCP"
+      serviceAccount: "kubernetes-quickstart"
+---
+
+The provided replicas, environment variables and labels where retained. However, the image and container port where modified. Moreover, the standard annotations have been added.
```
maxandersen, line 819: I assume you meant it to be "were retained" and "port were modified" ? might actually be clearer if we simplify and just writes "...where used as is. However, the image and container port where replaced and standard annotations were added"

Also by "standard" I assume you mean the default annotations dekorate/quarkus extension adds ? I would use "default" instead of standard.

docs/src/main/asciidoc/deploying-to-kubernetes.adoc
```diff
@@ -735,3 +736,87 @@ This assumes that a `.kube/config` is available in your user directory that poin
...
+---
+
+The generated `kubernetes.yml` will look like:
+
+[source,yaml]
+---
+apiVersion: "apps/v1"
+kind: "Deployment"
+metadata:
+  annotations:
```
maxandersen, line 787: suggestion, add asciidoctor callouts ( `## <>`) to mark the changed/unedited places to make it easier for the reader to spot the differences.

maxandersen (changes requested):
this is great.

Left a few comments on clarifying the docs.

and a question: what attributes/rules are replaced and what are contained ? is it just the specific known image references or is there additional "magic guessing" in place ?

geoand:
@iocanel I also think we need another piece:

While looking into handling roles (for the new `kubernetes-config` extension), I saw that currently we are assigning the `view` `ClusterRole` to any application using the `kubernetes-client` extension (which is also used by `kubernetes-config`).
This is fine for local development and demos, but it's definitely not good for production. Moreover this role is not enough for viewing `Secret` objects.
So I am thinking that we probably need to allow users to bring their own `Role` object (in `src/main/kubernetes`) which will then be copied into the output `kubernetes` directory and used in the generated `RoleBinding`.

WDYT?

gsmet (changes requested):
We will need a squash before commit. Please dismiss this review once done.

docs/src/main/asciidoc/deploying-to-kubernetes.adoc
iocanel: Great idea!

docs/src/main/asciidoc/deploying-to-kubernetes.adoc
iocanel: Makes sense!

docs/src/main/asciidoc/deploying-to-kubernetes.adoc
```diff
@@ -735,3 +736,87 @@ This assumes that a `.kube/config` is available in your user directory that poin
...
+    app.quarkus.io/build-timestamp: "2020-04-10 - 12:54:37 +0000"
+  labels:
+    app: "greeting"
```
Ladicek, line 790: In the previous file, we have `app: quickstart`, and here, we have `app: "greeting"`. Is that correct, or a typo? If correct, it deserves explanation.

Ladicek:
Very nice, thanks!

docs/src/main/asciidoc/deploying-to-kubernetes.adoc
iocanel: Good point!

iocanel:
> This is fine for local development and demos, but it's definitely not good for production.

I don't agree. Production or not if an application needs to talk to the api master, it will at least need the view role (I can't think of a scenario using anything less).

> So I am thinking that we probably need to allow users to bring their own `Role` object (in `src/main/kubernetes`) which will then be copied into the output `kubernetes` directory and used in the generated `RoleBinding`.
> 
> WDYT?

Users are already able to bring their own `Role` object and its expected to work. 
We might need to handle naming collisions (what if the user adds a role binding with the same name as the one the kubernetes extension creates).

geoand:
> > This is fine for local development and demos, but it's definitely not good for production.
> 
> I don't agree. Production or not, if an application needs to talk to the api master, it will at least need the view role (I can't think of a scenario using anything less).
> 

I don't disagree on that. But why does the it have to be a `ClusterRole`? 

> > So I am thinking that we probably need to allow users to bring their own `Role` object (in `src/main/kubernetes`) which will then be copied into the output `kubernetes` directory and used in the generated `RoleBinding`.
> > WDYT?
> 
> Users are already able to bring their own `Role` object and its expected to work.
> We might need to handle naming collisions (what if the user adds a role binding with the same name as the one the kubernetes extension creates).

You mean with this PR?

iocanel:
> You mean with this PR?

Yes! The user can bring in ANY kind of resource, that is supported by the kubernetes-client.

geoand:
OK, I guess we should just get this in when it's done and go from there. Perhaps for the kubernetes-config extension all we will need is appropriate docs.

Ladicek:
I took a quick look, because I was also wondering "why does it have to be a `ClusterRole`?", and it turns out that itself is harmless. That's simply where the role is defined: https://kubernetes.io/docs/reference/access-authn-authz/rbac/#user-facing-roles What is important is that we generate a `RoleBinding` (which only applies in given namespace) and _not_ a `ClusterRoleBinding` (which would apply across all namespaces). So I think the way we generate the `ServiceAccount` and `RoleBinding` is fine even for production usage.

geoand:
Thanks for checking @Ladicek.

iocanel:
@maxandersen: your feedback has been applied, can you have another look?

maxandersen:
looks good @iocanel 

I still have this question: what attributes/rules are replaced and what are contained ? is it just the specific known image references or is there additional "magic guessing" in place ?

maxandersen (approved):

docs/src/main/asciidoc/deploying-to-kubernetes.adoc
```diff
@@ -767,3 +768,87 @@ This assumes that a `.kube/config` is available in your user directory that poin
 In other words the extension will use whatever cluster `kubectl` uses. The same applies to credentials.
 
 At the moment no additional options are provided for further customization.
+
+== Using existing resources
+
+Some times it's desirable to either provide additional resources (e.g. a ConfigMap, a Secret, a Deployment for a database etc) or provide custom ones that will be used as a `base` for the generation process.
```
gsmet, line 774:
```suggestion
Sometimes it is desirable to either provide additional resources (e.g. a ConfigMap, a Secret, a Deployment for a database etc) or provide custom ones that will be used as a `base` for the generation process.
```

docs/src/main/asciidoc/deploying-to-kubernetes.adoc
```diff
@@ -767,3 +768,87 @@ This assumes that a `.kube/config` is available in your user directory that poin
...
+Those resources can be added under `src/main/kubernetes` and can be named after the target environment (e.g. kubernetes.json, openshift.json, knative.json, or the yml equivalents). Each of these files may contain one or more Kubernetes resources.
+
+Any resource found, will be added in the generated manifests. Global modifications (e.g. labels, annotations etc) will also be applied to those resources.
```
gsmet, line 777:
```suggestion
Any resource found will be added in the generated manifests. Global modifications (e.g. labels, annotations etc) will also be applied to those resources.
```

docs/src/main/asciidoc/deploying-to-kubernetes.adoc
```diff
@@ -767,3 +768,87 @@ This assumes that a `.kube/config` is available in your user directory that poin
...
+If one of the provided resources has the same name as one of the generated resources, then the generated resource will be created on top of the provided, respecting existing content when possible (e.g. existing labels, annotations, environment variables, mounts, replicas etc).
```
gsmet, line 778:
```suggestion
If one of the provided resources has the same name as one of the generated resources, then the generated resource will be created on top of the provided resource, respecting existing content when possible (e.g. existing labels, annotations, environment variables, mounts, replicas etc).
```

docs/src/main/asciidoc/deploying-to-kubernetes.adoc
```diff
@@ -767,3 +768,87 @@ This assumes that a `.kube/config` is available in your user directory that poin
...
+
+The name of the resource is determined by the application name, and may be overridden by `quarkus.kubernetes.name`, `quarkus.openshift.name` and `quarkus.knative.name`.
+
+For example, in the `kubernetes-quickstart` application, we can add a `kubernetes.yml` file in the `src/main/kubernetes` that looks like:
```
gsmet, line 782:
```suggestion
For example, in the `kubernetes-quickstart` application, we can add a `kubernetes.yml` file in the `src/main/kubernetes` directory that looks like:
```

docs/src/main/asciidoc/deploying-to-kubernetes.adoc
```diff
@@ -767,3 +768,87 @@ This assumes that a `.kube/config` is available in your user directory that poin
...
+        - containerPort: 8080 <5>
+          name: "http"
+          protocol: "TCP"
+      serviceAccount: "kubernetes-quickstart"
+---
+
+The provided replicas <1>, labels <2> and environment variables <3>  were retained. However, the image <4> and container port <5> were modified. Moreover, the default annotations have been added.
+
+Note: If the resource name does not match the application name (or the overridden name) instead of reusing the resource a new one will be added. Same goes for the container.
+If the name of the container does not match the application name (or the overridden name), container specific configuration will be ignored.
```
gsmet, line 854:
```suggestion
[NOTE]
====
If the resource name does not match the application name (or the overridden name) instead of reusing the resource a new one will be added. Same goes for the container.
If the name of the container does not match the application name (or the overridden name), the container specific configuration will be ignored.
====
```

gsmet (changes requested):
Added a couple of doc suggestions.

gsmet:
@iocanel could you rebase and apply my doc suggestions?

@maxandersen also had a few questions.

Let's drive to get this one in soon.

iocanel:
@gsmet: I'll try to get that tommorow.

PR #8539, opened 2020-04-12
Write quarkus-core this extension depends on in quarkus-extension.yaml

gastaldi:
The quarkus-extension.yaml now has the `quarkus-core` version used when the extension is built. This is useful for tooling when dealing with third-party extensions to avoid traversing the dependency tree

### DIFF
--- a/independent-projects/bootstrap/maven-plugin/src/main/java/io/quarkus/maven/ExtensionDescriptorMojo.java
+++ b/independent-projects/bootstrap/maven-plugin/src/main/java/io/quarkus/maven/ExtensionDescriptorMojo.java
@@ -10,14 +10,18 @@ import com.fasterxml.jackson.databind.SerializationFeature;
 import com.fasterxml.jackson.databind.node.ObjectNode;
 import com.fasterxml.jackson.dataformat.yaml.YAMLFactory;
 import io.quarkus.bootstrap.BootstrapConstants;
+import io.quarkus.bootstrap.resolver.AppModelResolverException;
+import io.quarkus.bootstrap.resolver.maven.MavenArtifactResolver;
 import java.io.BufferedWriter;
 import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.util.Collections;
 import java.util.List;
 import java.util.Properties;
+import java.util.concurrent.atomic.AtomicReference;
 import org.apache.maven.plugin.AbstractMojo;
 import org.apache.maven.plugin.MojoExecutionException;
 import org.apache.maven.plugins.annotations.Component;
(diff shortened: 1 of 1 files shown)


independent-projects/bootstrap/maven-plugin/src/main/java/io/quarkus/maven/ExtensionDescriptorMojo.java
```diff
@@ -274,16 +283,50 @@ private void transformLegacyToNew(final Path output, ObjectNode extObject, Objec
...
+        }
         extObject.set("metadata", metadata);
+    }
 
-        //   updateSourceFiles(output, extObject, mapper);
+    private String resolveQuarkusCoreVersion() throws MojoExecutionException {
+        MavenArtifactResolver mvn = getMavenResolver();
+        DefaultArtifact artifact = new DefaultArtifact(project.getGroupId(),
+                project.getArtifactId(),
+                project.getPackaging(),
```
aloubyansky, line 298: 'packaging' is not the same as the type. `projectArtifact.getArtifactHandler().getExtension()` will be more appropriate.

independent-projects/bootstrap/maven-plugin/src/main/java/io/quarkus/maven/ExtensionDescriptorMojo.java
```diff
@@ -274,16 +283,50 @@ private void transformLegacyToNew(final Path output, ObjectNode extObject, Objec
...
+            public boolean visitEnter(DependencyNode node) {
+                return version.get() == null;
+            }
 
+            @Override
+            public boolean visitLeave(DependencyNode node) {
+                final Dependency dep = node.getDependency();
+                if (dep != null
+                        && dep.getArtifact().getArtifactId().equals("quarkus-core")
+                        && dep.getArtifact().getGroupId().equals("io.quarkus")) {
```
aloubyansky, line 318: There are extensions that don't depend on `quarkus-core` and it's not necessarily a mistake. However every deployment artifact has `quarkus-core-deployment` among its deps. So, that's what you should be looking for instead.

aloubyansky (changes requested):

independent-projects/bootstrap/maven-plugin/src/main/java/io/quarkus/maven/ExtensionDescriptorMojo.java
gastaldi: The problem is that the `quarkus-extension.yaml` is in the runtime module, and the deployment module is a dependency of of this module

aloubyansky: Is there a reason to keep it in the runtime artifact?

gastaldi: Dunno, it has always been like this AFAIK.

gastaldi:
Closing for now

PR #8544, opened 2020-04-13
Fix issue with native cron build

geoand:
Attempt to get around:

```
No space left on device
```

error encountered during the `Quarkus CI - JDK 8 Native Build` run

### DIFF
--- a/.github/workflows/native-cron-build.yml
+++ b/.github/workflows/native-cron-build.yml
@@ -35,9 +35,9 @@ jobs:
       - uses: actions/cache@v1
         with:
           path: ~/.m2/repository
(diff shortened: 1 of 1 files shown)


gsmet:
I'm not sure I see how it fixes the issue?

geoand:
It hopefully fixes it by removing the existing cached file. It might not be the right fix, but it can't hurt.

gsmet:
But AFAICS the cache file is removed every time we update a pom file. Which means it's done quite often and in any case each time we .

geoand:
If you think it's not correct, feel free to close. This was a zero-thought attempt in any case

gsmet (approved):

gsmet:
I suppose it doesn't hurt.

Issue #8525, opened 2020-04-10
Enhance JavaDoc for "quarkus-hibernate-orm" and "quarkus-hibernate-orm-deployment"


michael-schnell:
Currently there is almost no JavaDoc in "quarkus-hibernate-orm" or "quarkus-hibernate-orm-deployment". This makes it really hard, even for experienced developers, to work on this code.

More information needs to be added to all types & methods explaining the purpose.

Issue #2835, opened 2019-06-13
Support multiple persistence units


FroMage:
As shown in #2759 comments, while we support defining multiple datasources, we don't actually support defining multiple persistence units:

- we don't support them via `application.properties`
- we can't use the datasources in `persistence.xml`
- we can't seem to be able to inject `EntityManager` for multiple persistence units.

radcortez:
+1

gsmet:
This one should still be open.

JfEstrela:
this approach is necessary

stefanorg:
+1
we really need this

famod:
Very much needed for modular monoliths.

ionutbalosin:
+1

gpietrek:
+1

GuangjieZHANG:
Yes, we really want it, please ... lol

yuhaibohotmail:
+1

dragondiver:
+1

madprogger:
We really need that.

chrisgleissner:
This is currently a blocker for us. Is any work planned on this?

Sanne:
yes, of course we'll implement this. Sorry it's taking some time: several other issues are pre-requisites to be able to do this.

dev-whiteflag:
Porting a project to Quarkus, really need this. +1

AmsterdamFilho:
Currently migrating 40 Microservices from Thorntail to Quarkus. Really need this too. +1

chrisgleissner:
We plan on breaking up a JBoss 4 monolith into microservice and would be really keen to use Quarkus instead of Spring Boot. This feature is the only problem that would currently percent us from using Quarkus. Really hope it can be implemented soon, or maybe there is a workaround to manually create EntityManager instances from Datasource instances as Quarkus already supports multiple Datasources?

mlsdeveloper:
Currently migranting my projects from spring boot to Quarkus. Realy need this too. +1

sekaijin:
I have hundreds of camel routes, some of which have three or even four different data sources.
Going from servicemix where these datasources are defined once for all routes, to microservices where they have to be defined as many times as there is use is already a major operating constraint. when a connection parameter changes for a source it is necessary to review all the routes which use it. But if you cannot read data in two databases to aggregate them in a third, it is on that the microservice architecture is not the solution for data exchanges.
Realy need this too. +5

famod:
[...]

PR #8545, opened 2020-04-13
Hibernate ORM Multitenancy

### LINKED ISSUE #5681: Hibernate ORM - Multitenancy support
### Description
Hibernate ORM is able to support multi-tenancy in various ways.

Some of these imply "plugging" custom strategy implementations during the bootstrap of the ORM, and/or setting the tenant ID when the `Session` is opened.

Both aspects could see some level of automation and "just work" OOB without needing to implement such SPIs - most importantly, since Quarkus needs to control the boostrap proces we would need to ensure such details are available during augmentation.

### Analysis
TBD

First user request is here: https://quarkusio.zulipchat.com/#narrow/stream/187030-users/topic/Quarkus.20ORM.20multitenancy

### Tasks
- [ ]

michael-schnell:
Fixes #5681


michael-schnell:
One of the checks failed, but it seems to be a build server problem

geoand:
Yeah, one of the steps timed out, it has nothing to do with this PR.

michael-schnell:
> Yeah, one of the steps timed out, it has nothing to do with this PR.

Will someone restart that step or how is the procedure?

geoand:
I restarted it

michael-schnell:
@machi1990 Can you invite the reviewers again? (First pull request had problems and I started this new one)

FroMage:
I'm not sure I'm qualified to review this PR. But I thought Hibernate could not use multiple entity managers for multiple data sources yet? This seems to be based around the fact that it can. Did we already add support for that, @Sanne ?

michael-schnell:
> I'm not sure I'm qualified to review this PR. But I thought Hibernate could not use multiple entity managers for multiple data sources yet? This seems to be based around the fact that it can. Did we already add support for that, @Sanne ?

Multiple data sources is already supported by Quarkus:
https://quarkus.io/guides/datasource#multiple-datasources

FroMage:
Ah, it's multiple persistence units that it doesn't support: https://github.com/quarkusio/quarkus/issues/2835

machi1990:
> Ah, it's multiple persistence units that it doesn't support: #2835

@FroMage That's why (#2835) I asked a review from you too  :-)

michael-schnell:
> Ah, it's multiple persistence units that it doesn't support: #2835

Yes, but when it is finally supported, it would be easily possible to enhance the multitenancy to use persistence units in addition to multiple data sources.

michael-schnell:
Any chance that we get this PR into 1.4.0? 
I would like to use it in my code, but without a release this is difficult.

geoand:
I'm not the one that decides, but I think that would be difficult. We already have a CR for 1.4 and we usually only tend to add fixes into the Final.

But like I said, it's not up to me 😎

Sanne:
sorry for the delay @michael-schnell , finally getting to try this out

michael-schnell:
> finally getting to try this out

@Sanne How is it going?

michael-schnell:
Any chance to get this some day done? It is now two weeks since the pull request was made and there is no real feedback til now. I know the situation is special, but this is not a thousands-of-lines codes review. It's just a few files, easy to test and should be a matter of half an hour or hour.  Sorry folks, but this is really annoying.

Sanne:
bummer, sorry I had some early feedback but it seems I didn't publish it. Back to it now..

Sanne (commented):
I didn't get to review the implementation details - however that's less important, it can always be refined in a second time.

I have some questions about the API we want to expose though. Added them in the review.

docs/src/main/asciidoc/hibernate-orm.adoc
```diff
@@ -542,3 +542,216 @@ Datasource configuration is extremely simple, but is covered in a different guid
...
+@Consumes("application/json")
+@Path("/{tenant}")
+public class FruitResource {
+
+    @Inject
+    EntityManager entityManager;
+
+    @GET
+    @Path("fruits")
+    public Fruit[] getFruits() {
```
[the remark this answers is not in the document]
michael-schnell: I don't think it is necessary for the reads. The description is besically just a copy of the quickstart example [FruitResource.java](https://github.com/michael-schnell/quarkus-quickstarts/blob/master/hibernate-orm-quickstart/src/main/java/org/acme/hibernate/orm/FruitResource.java) that has been copied from the orginal [Hibernate ORM Quickstart](https://github.com/michael-schnell/quarkus-quickstarts/blob/master/hibernate-orm-quickstart/src/main/java/org/acme/hibernate/orm/FruitResource.java).

docs/src/main/asciidoc/hibernate-orm.adoc
```diff
@@ -542,3 +542,216 @@ Datasource configuration is extremely simple, but is covered in a different guid
...
+@Unremovable
+public class CustomTenantResolver implements TenantResolver {
+
+    @Override
+    public String getDefaultHibernateOrmTenantId() {
+        return "base";
+    }
+    
+    @Override
+    public String resolveHibernateOrmTenantId(RoutingContext context) {
```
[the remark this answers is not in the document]
michael-schnell: I think the routing context is a good choice. That's why I have taken the idea from the ODIC Multitenant implementation: [CustomTenantResolver.java](https://github.com/quarkusio/quarkus-quickstarts/blob/master/security-openid-connect-multi-tenancy/src/main/java/org/acme/quickstart/oidc/CustomTenantResolver.java).

You would normally need to get some information from the incoming request to decide which tenant is targeted with the request. The information could be taken from path (as in the example) or maybe also from the request header. 

It is also very likely that the two parts Hibernate & ODIC multtenancy will be used in combination. You can simply create a single class that shares some code between the two. Unfortunately it was not really possible to use exactly the same signature. The ODIC multtenancy relies on the fact that NULL is considered as the value for the default tenant. Hibernate does not allow a NULL value for the tenant id.

docs/src/main/asciidoc/hibernate-orm.adoc
```diff
@@ -542,3 +542,216 @@ Datasource configuration is extremely simple, but is covered in a different guid
...
+
+== Multitenancy
+
+"The term multitenancy, in general, is applied to software development to indicate an architecture in which a single running instance of an application simultaneously serves multiple clients (tenants). This is highly common in SaaS solutions. Isolating information (data, customizations, etc.) pertaining to the various tenants is a particular challenge in these systems. This includes the data owned by each tenant stored in the database" (link:https://docs.jboss.org/hibernate/orm/5.4/userguide/html_single/Hibernate_User_Guide.html#multitenacy[Hibernate User Guide]).
+
+Quarkus currently supports the link:https://docs.jboss.org/hibernate/orm/5.4/userguide/html_single/Hibernate_User_Guide.html#multitenacy-separate-database[separate database] and the link:https://docs.jboss.org/hibernate/orm/5.4/userguide/html_single/Hibernate_User_Guide.html#multitenacy-separate-schema[separate schema] approach. 
+
+=== Writing the application
+
+Let's start by implementing the `/{tenant}` endpoint. As you can see from the source code below it is just a regular JAX-RS resource:
```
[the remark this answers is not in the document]
michael-schnell: Tnx. The enhanced doc is combined with a new quickstart that can be merged right after this pull request: https://github.com/quarkusio/quarkus-quickstarts/pull/525

It has an additional dropdown to select the tenant for every request. It shows in detail how to use this feature and it works quite well.

docs/src/main/asciidoc/hibernate-orm.adoc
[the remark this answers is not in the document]
Sanne: I'm not eager to _require_ a dependency on the vertx module: it needs to be possible to use the Hibernate ORM extension even in an application which doesn't have any web endpoint.

I think there's some options:

1. Expose a different type, something like `TenantId`, or `MultitenancyContext`, and leave it to the developer to have a bean which performs the conversion - possibly making it easy such as having a CDI Factory.
2. Move your `TenantResolver` approach into a dedicated module / extension
3. Make sure that `TenantResolver` is one way out of several options - making the binding to the VertX module optional.

I suppose we can mix these approaches; for example my favourite is 1# in terms of clean design, but you could have a default factory (2#) provided for user's convenience. This could initially like in this same module (3#) and later be broken out as a future improvement.

Or maybe you have a better idea?

Sanne (changes requested):

extensions/hibernate-orm/runtime/src/main/java/io/quarkus/hibernate/orm/runtime/tenant/HibernateCurrentTenantIdentifierResolver.java
```diff
@@ -0,0 +1,89 @@
...
+        return routingContextInstance.get();
+    }
+
+    /**
+     * Retrieves the tenant resolver or fails if it is not available.
+     * 
+     * @return Current tenant resolver.
+     */
+    private static TenantResolver tenantResolver() {
+        InstanceHandle<TenantResolver> resolverInstance = Arc.container().instance(TenantResolver.class);
```
Sanne, line 81: Resolving the reference to the right `TenantResolver` via Arc is not necessary and actually a problem as we want to be able to start multiple instances of Hibernate ORM, which could be configured differently.

You should be able to simply inject the `HibernateCurrentTenantIdentifierResolver` instance into the ORM configuration; for example this should work:

    configuration.put( org.hibernate.cfg.AvailableSettings.MULTI_TENANT_IDENTIFIER_RESOLVER, new HibernateCurrentTenantIdentifierResolver( tenantResolver ) );

This way you can store the tenant into a `final` field, properly and simply initialized in its constructor.

michael-schnell: Which " ORM configuration" do you exactly refer to?

Sanne: It's the `cfg` Map variable you can see in `io.quarkus.hibernate.orm.runtime.boot.FastBootMetadataBuilder#mergeSettings`, were we set most other automatically inferred configuration properties.

docs/src/main/asciidoc/hibernate-orm.adoc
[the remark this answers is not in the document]
michael-schnell: OK, I think this is not a big problem. I will just rewrite the TenantResolver like this:
```java
public interface TenantResolver {
    String getDefaultTenantId();
    String resolveTenantId();
}
```
The implementation by the user then is:
```java
@RequestScoped
public class CustomTenantResolver implements TenantResolver {
    
    @Inject
    RoutingContext context;
    
    public String getDefaultTenantId() {
        return "base";
    }
    public String resolveTenantId() {
        ...
    }
}
```
So there is no longer a dependency in the Hibernate ORM extension to vertx, only in the user's implementation.

extensions/hibernate-orm/runtime/src/main/java/io/quarkus/hibernate/orm/runtime/tenant/HibernateCurrentTenantIdentifierResolver.java
michael-schnell: OK, I guess it is then not the "FastBootMetadataBuilder", but the [FastBootEntityManagerFactoryBuilder](https://github.com/michael-schnell/quarkus/blob/a24a32087d6f107eb5fe4fff8f757a1baf928850/extensions/hibernate-orm/runtime/src/main/java/io/quarkus/hibernate/orm/runtime/boot/FastBootEntityManagerFactoryBuilder.java#L166) as the instance of HibernateCurrentTenantIdentifierResolver is created there.

But I assume this will not work as the "FastBootEntityManagerFactoryBuilder" will be created at startup (maybe application scope), but the "TenantResolver" needs to be resolved within a request scope.

Is there really a need to support multiple instances of Hibernate ORM with multiple different implementations of the "TenantResolver"? This will get much more complicated. I guess the only way then would be to configure the class name of the actual "TenantResolver" per Hibernate ORM instance. But how do I configure multiple instances of Hibernate ORM? I haven't seen such a feature anywhere in the docs.

docs/src/main/asciidoc/hibernate-orm.adoc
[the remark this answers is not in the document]
Sanne: Thanks!

The new `TenantResolver` interface looks good; question: wouldn't it be better to not have a defaultTenantId ?
If you don't force one, I would expect implementors can still choose to return a default - should other options fail. But it might also be clearer to them when the default would be used (since it's their own code), which I would find more reassuring.

Regarding the implementaiton, I do like this example but I believe it should be possible as well to not need Arc. Shouldn't the `CustomTenantResolver` rather have the option to lookup an Arc component?

extensions/hibernate-orm/runtime/src/main/java/io/quarkus/hibernate/orm/runtime/tenant/HibernateCurrentTenantIdentifierResolver.java
Sanne: > Is there really a need to support multiple instances of Hibernate ORM with multiple different implementations of the "TenantResolver"?

I'm afraid so: if we take such shortcuts it would all get confusing to use.

> But how do I configure multiple instances of Hibernate ORM? I haven't seen such a feature anywhere in the docs.

We don't have the capability yet but it is in high demand; the problem is that as lots of people keep adding features and changes that make it conceptually harder we'll never be able to complete that.. which is why I please need some help in making sure all newly merged code at least take the requirement into account.

I do understand it gets a little bit more complicated but I just want to make sure the design will make it possible at a conceptual level - not asking you to actually implement it.

Thanks!

michael-schnell: OK, let's assume we have multiple instances of Hibernate ORM. Then we will need to distinguish multiple instances of the tenant resolver and the resolvers may have different types. This could be achieved using CDI qualifiers and explicitly naming the resolver class.

Then the solution would have two parts:
1. Define the type of the resolver by having it's fully qualified class name in the config
2. Define a qualifier in the config that maps somehow to the Hibernate ORM instance

Example:
```
quarkus.hibernate-orm.one.multitenant=SCHEMA
quarkus.hibernate-orm.one.multitenancy-resolver=a.b.c.MyResolver // Implements TenantResolver
quarkus.hibernate-orm.one.multitenancy-qualifier=one // Qualifier
quarkus.hibernate-orm.two.multitenant=DATABASE
quarkus.hibernate-orm.two.multitenancy-resolver=a.b.c.OtherResolver // Implements TenantResolver
quarkus.hibernate-orm.two.multitenancy-qualifier=two // Qualifier
```
Currently we don't have multiple instances and we don't want to bother the user with such edge case configurations. The "95% case" will simply be one instance. Therefore we need a way to keep it still simple. This is exactly how it is implemented now and it is open to above future design.

In the FastBootEntityManagerFactoryBuilder the HibernateCurrentTenantIdentifierResolver instance is created and might use some configuration options later on, that are not needed now:
```java
if (strategy != null && strategy != MultiTenancyStrategy.NONE) {
    options.applyMultiTenancyStrategy(strategy);
    options.applyCurrentTenantIdentifierResolver(new HibernateCurrentTenantIdentifierResolver());
}
```
In the HibernateCurrentTenantIdentifierResolver class we still resolve the TenantResolver dynamically (using Arc), because it is RequestScoped and we cannot create it at boot time in the FastBootEntityManagerFactoryBuilder:
```java
private static TenantResolver tenantResolver() {
    InstanceHandle<TenantResolver> resolverInstance = Arc.container().instance(TenantResolver.class);
    ....
}
```
Later on, when multiple Hibernate ORM instances are available you could use the qualifier and class from the configuration similar to this one:
```java
private static TenantResolver tenantResolver(Class<TenantResolver> resolverClass, String qualifierName) {
    InstanceHandle<TenantResolver> resolverInstance = Arc.container().instance(resolverClass,
            new AnnotationLiteral() {
                public String value() {
                    return qualifierName;
                };
            });
      ...
```

**Summary**
If we do the changes you proposed in regard to the vertx module dependency and make the TenantResolver request scoped (as suggested above), the HibernateCurrentTenantIdentifierResolver can stay as it is now.

docs/src/main/asciidoc/hibernate-orm.adoc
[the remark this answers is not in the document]
michael-schnell: > The new TenantResolver interface looks good; question: wouldn't it be better to not have a defaultTenantId ?
> If you don't force one, I would expect implementors can still choose to return a default - should other options fail. But it might also be clearer to them when the default would be used (since it's their own code), which I would find more reassuring.

I did that first, but unfortunately Hibernate does not allow a null value for a tenantId. This means we would still have to create an artifical default value internally to use it with the Hibernate settings. Then we have to translate that back to the null value for the TenantConnectionResolver, because the user expects now also a null tenantId over there:
```java
public interface TenantConnectionResolver {
    ConnectionProvider resolve(String tenantId);
}
```
At the end I think it is easier to understand if you say "You need to define a default value for the tenantId and it cannot be null". That is what is expressed in the interface. That also matches the Hibernate philosophy which does not allow null.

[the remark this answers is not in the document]
Sanne: Regarding the default tenantId : ok, if you prefer so that's fine. Just making sure we're understanding the implications: don't you think it's generally a possible design error from the user's part to ever return null?
In such cases we normally prefer to throw an explict exception rather than silently and sneakily change semantics; I suspect that just requiring a tenantId would make for more robust end user systems.

Regarding the Arc puzzle:

I meant it could be the user's implementation of `TenantConnectionResolver` to explicitly have some code like 

    Arc.container().instance(RoutingContext.class);

should he/she want to base the decision on the routing - or anything else.

It's not a suggestion though, I'm not an expert of CDI nor ArC - just wondering if that's not conceptually a clean solution so to not need users to specify multiple configuration properties.

extensions/hibernate-orm/runtime/src/main/java/io/quarkus/hibernate/orm/runtime/tenant/HibernateCurrentTenantIdentifierResolver.java
Sanne: > Currently we don't have multiple instances and we don't want to bother the user with such edge case configurations. The "95% case" will simply be one instance. Therefore we need a way to keep it still simple. This is exactly how it is implemented now and it is open to above future design.

I like what you're saying here and totally agree that we should keep it simple for the 95% of users.

Referring to your example:

I might be missing something since I've never used CDI, but wouldn't we conceptually need the user to just specify _either_ the qualifier _or_ the class? I would expect it to be possible to only need to express a "single choice".

Further more, consider that via the Jandex index we have a full analysis of the application; this implies that we can enumerate the implementations of, for example, a TenantResolver implementation and choose it as default when only a single implementation is found.
(should we go with a single TenantResolver implementation and use qualifiers instead - assuming it's possible to use only one approach as mention in the previous paragraph - then resolving the default bean rather than using qualifiers should come even more natural?)

That would be sweet as it means the components could be bound correctly without any configuration at all; requiring configuration only for multiple persistence units which require to differentiate their settings.

@manovotn / @mkouba would you have some time to review and correct my ramblings about ArC? I'd really hope we could keep this both flexible and straight forward to use :)

Thanks!

manovotn: I have quite limited understanding of what you're trying to achieve and might need some more context, but broadly speaking - if you want to have multiple implementations of, say, `TenantResolver`, then you will end up with ambiguity if you try to just inject it based on type `TenantResolver`. You'd have to either use qualifiers, or resolve it based on the implementing class.

As for simple cases and having one `TenantResolver` by default, that's doable, you just have to have a singular existing bean with that type and without CDI qualifiers. From the above comments I understood that this is something you already have.

If the impl of `TenantResolver` is something that user defines, then they can define the qualifiers used with it as well. And if you then need to retrieve all impls of `TenantResolver` via CDI (regardless of qualifiers) you can do that too with (for instance) `Arc.container().instance(TenantResolver.class, Any.Literal.INSTANCE)` and then iterate over them.

manovotn: If I am blabbering about something else then what you wanted to know, then please give me some more info and I'll try to help some more :-)

michael-schnell: At the end I just wanted to show that using "Arc.container().instance(..)" is not "evil" here. It is required, because we need to resolve a request scoped bean and therefore we cannot do it at boot time. In addition to that it allows us in a later stage to change this as needed in HibernateCurrentTenantIdentifierResolver. But no need to do it now as we don't have multiple Hibernate ORM instances yet. So no need for qualifiers and other stuff now, but a way to go in the future.

I already tested the new classes with the new Hibernate ORM multitenancy quickstart and it still works finr with the changes above. I will now commit them to continue here with the next review steps.

michael-schnell:
@Sanne Can you check the changes?

michael-schnell:
@Sanne PING

michael-schnell:
@gsmet @FroMage @emmanuelbernard @Sanne Can someone please continue with the review - I started working on this more than a month ago. The pull request is now also already three weeks old. I changed several things on demand of @Sanne and now again no response since 7 days. Looks like you are not really interested in people doing some work. Please let's continue to close this never ending story!

FroMage:
We are always very interested in contributions. But we're also all very busy on a lot of other things, not to mention disrupted by the current pandemic. So please be patient. It's not unusual for PRs to take a few weeks or months. What matters is that at the end they get merged and are much better than before reviews took place.
Don't lose hope :)

emmanuelbernard:
> We are always very interested in contributions. But we're also all very busy on a lot of other things, not to mention disrupted by the current pandemic. So please be patient. It's not unusual for PRs to take a few weeks or months. What matters is that at the end they get merged and are much better than before reviews took place.
> Don't lose hope :)

Right, the specific difficulty here is that you are touching on an area with lots of implications and ramifications. So a casual check / optimistic push is out of the equation. This low level detail is only know intimately enough by few people. That's why we prefer to wait for Sanne's feedback to get a good solution out.

michael-schnell:
Hope we will get it done at the end...

> ... you are touching on an area with lots of implications and ramifications. So a casual check / optimistic push is out of the equation. This low level detail is only know intimately enough by few people. That's why we prefer to wait for Sanne's feedback to get a good solution out.

Hmmm.... This is not what I see in the code. It is currently pretty "hacky" at many places and has almost no documentation. Definetly a long way to "Clean Code". I guess this is also one of the reasons why there are so few developers who can take care of a PR. 

I already suggested in another issue (#8525) to setup a SonarCloud instance, which is free for OS projects. This will be a huge step toward a better maintainability together with some JavaDoc (#8525)

michael-schnell:
@Sanne 14 more days passed since the last commit - Any chance to do the review?

Sanne:
hi @michael-schnell , sure I'm back on this this afternoon. checking it out..

extensions/hibernate-orm/runtime/src/main/java/io/quarkus/hibernate/orm/runtime/tenant/HibernateMultiTenantConnectionProvider.java
```diff
@@ -0,0 +1,78 @@
...
+                    + "This violates the contract of the interface!");
+        }
+        return selectConnectionProvider(tenantId);
+    }
+
+    @Override
+    protected ConnectionProvider selectConnectionProvider(String tenantIdentifier) {
+        LOG.debugv("selectConnectionProvider({0})", tenantIdentifier);
+
+        return PROVIDER_MAP.computeIfAbsent(tenantIdentifier, tid -> {
```
Sanne, line 41: I'd prefer if you could use an "old fashioned get" here first, only falling back to a `computeIfAbsent` operation if your local cache doesn't have it so to avoid contention in the implementation of `ConcurrentHashMap#computeIfAbsent` since there is no need to have contention.

extensions/hibernate-orm/deployment/src/main/java/io/quarkus/hibernate/orm/deployment/HibernateOrmConfig.java
```diff
@@ -101,6 +101,33 @@
...
+    /**
+     * Defines the name of the data source to use in case of SCHEMA approach. The default data source will be used if not set.
+     *
+     * @asciidoclet
+     */
+    @ConfigItem
+    public Optional<String> multitenantSchemaDatasource;
+
+    /**
+     * Defines if the tenant identifier of the "current sessions" should be validated every time the tenant identifier is
```
Sanne, line 123: I'm wondering if it wouldn't be better to refrain from exposing this option. AFAIR Quarkus we don't expose `org.hibernate.context.spi.CurrentSessionContext`, which makes this check useless?

I'd prefer to omit this, as less is more - especially when it's easy to add things, but never easy to remove something.

Also, removing this configuration option would resolve a couple more comments I had :)

extensions/hibernate-orm/deployment/src/main/java/io/quarkus/hibernate/orm/deployment/HibernateOrmConfig.java
```diff
@@ -101,6 +101,33 @@
     @ConfigItem
     Optional<String> implicitNamingStrategy;
 
+    /**
+     * Defines the method for multi-tenancy (DATABASE, NONE, SCHEMA). The complete list of allowed values is available in the
+     * https://docs.jboss.org/hibernate/stable/orm/javadocs/org/hibernate/MultiTenancyStrategy.html[Hibernate ORM JavaDoc].
+     * The type DISCRIMINATOR is currently not supported.
```
Sanne, line 107: Could you also clarify that `NONE` is the default?

extensions/hibernate-orm/deployment/src/main/java/io/quarkus/hibernate/orm/deployment/HibernateOrmProcessor.java
```diff
@@ -405,7 +417,7 @@ void registerBeans(BuildProducer<AdditionalBeanBuildItem> additionalBeans, Combi
 
         additionalBeans.produce(AdditionalBeanBuildItem.builder().setUnremovable()
                 .addBeanClasses(JPAConfig.class, TransactionEntityManagers.class,
-                        RequestScopedEntityManagerHolder.class)
+                        RequestScopedEntityManagerHolder.class, DataSourceTenantConnectionResolver.class)
```
Sanne, line 420: Could you avoid registering this bean as "unremovable" when multitenancy is not being used?

extensions/hibernate-orm/deployment/src/main/java/io/quarkus/hibernate/orm/deployment/HibernateOrmConfig.java
```diff
@@ -101,6 +101,33 @@
...
+    @ConfigItem
+    public Optional<String> multitenant;
+
+    /**
+     * Defines the name of the data source to use in case of SCHEMA approach. The default data source will be used if not set.
+     *
+     * @asciidoclet
+     */
+    @ConfigItem
+    public Optional<String> multitenantSchemaDatasource;
```
Sanne, line 120: So assuming I set it to "schema" approach, why would I ever want to set this to a different value than what I'm configuring Hiberante ORM to use in the `database.defaultSchema` setting?

I would prefer to imply and enforce a reasonable default rather than exposing another configuration option. Unless I'm missing something of course.

docs/src/main/asciidoc/hibernate-orm.adoc
```diff
@@ -542,3 +542,219 @@ Datasource configuration is extremely simple, but is covered in a different guid
...
+INSERT INTO known_fruits(id, name) VALUES (2, 'Apricots');
+INSERT INTO known_fruits(id, name) VALUES (3, 'Blackberries');
+----
+
+=== Programmatically Resolving Tenants Connections
+
+If you need a more dynamic configuration for the different tenants you want to support and don't want to end up with multiple
+entries in your configuration file, you can use the `io.quarkus.hibernate.orm.runtime.tenant.TenantConnectionResolver`.
+
+This interface allows you to dynamically create tenant connections at runtime.
```
Sanne, line 759: you're suggesting that the interface exists, but not how one creates one. Should it suggest also that we'll look for a CDI bean having such an interface?

Sanne (changes requested):
Great progress! I think it's almost ready, so started to look into more tiny details of the implementation - hence more fine grained feedback this time.

The bigger elefant in the room is the lack of any integration tests. Do you think you could add at least one?
Just covering one of the use cases from your documentation would give reasonable confidence IMO.

And thanks for the awesome docs!

Thanks

michael-schnell:
> Great progress! I think it's almost ready, so started to look into more tiny details of the implementation - hence more fine grained feedback this time.
> 
> The bigger elefant in the room is the lack of any integration tests. Do you think you could add at least one?
> Just covering one of the use cases from your documentation would give reasonable confidence IMO.

I will work on it over the weekend.

michael-schnell:
@gsmet @FroMage @emmanuelbernard @Sanne Is there something special about the integration test setup? 

I created a new project based on one of the existing integration test projects. The same setup works fine for the quickstart example, but it does not for the integration project. The @PathParam is somehow not passed correctly.

If I run the project manually using `./mvnw clean quarkus:dev` and I try a simple `GET http://localhost:8080/fruits/1` using Postman, no matter what integer value I pass, it is always zero.

Example:
```
@GET
@Path("fruits/{id}")
public Fruit getFruit(@PathParam int id) {
  // id is always 0, no matter what
}
```
The same works fine in the quickstart example.

Any idea what this could be?

docs/src/main/asciidoc/hibernate-orm.adoc
michael-schnell: > you're suggesting that the interface exists, but not how one creates one. Should it suggest also that we'll look for a CDI bean having such an interface?

I enhanced the doc for the custom tenant resolver.

extensions/hibernate-orm/runtime/src/main/java/io/quarkus/hibernate/orm/runtime/tenant/HibernateCurrentTenantIdentifierResolver.java
```diff
@@ -0,0 +1,66 @@
...
+        super();
+        this.jpaConfig = jpaConfig;
+    }
+
+    @Override
+    public String resolveCurrentTenantIdentifier() {
+
+        // Make sure that we're in a request
+        if (!Arc.container().requestContext().isActive()) {
+            return null;
```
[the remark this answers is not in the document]
michael-schnell: It is actually OK here. I checked this with the Hibernate source and it is not required to have a non-null value. The case is caused by calls to this method during early startup phase. At that time there is no request context so the only chance is to return null. For Hibernate there is in this case simply no tenant information available, which is fine.

extensions/hibernate-orm/runtime/src/main/java/io/quarkus/hibernate/orm/runtime/tenant/HibernateMultiTenantConnectionProvider.java
```diff
@@ -0,0 +1,78 @@
...
+
+/**
+ * Maps from the Quarkus {@link TenantConnectionResolver} to the {@link HibernateMultiTenantConnectionProvider} model.
+ * 
+ * @author Michael Schnell
+ *
+ */
+public class HibernateMultiTenantConnectionProvider extends AbstractMultiTenantConnectionProvider {
+
+    private static final long serialVersionUID = 1L;
```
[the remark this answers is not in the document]
michael-schnell: I removed it

extensions/hibernate-orm/runtime/src/main/java/io/quarkus/hibernate/orm/runtime/tenant/HibernateMultiTenantConnectionProvider.java
```diff
@@ -0,0 +1,78 @@
...
+
+    private static final Logger LOG = Logger.getLogger(HibernateMultiTenantConnectionProvider.class);
+
+    private static final Map<String, ConnectionProvider> PROVIDER_MAP = new ConcurrentHashMap<>();
```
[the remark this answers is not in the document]
michael-schnell: You're right, I changed it This was a leftover from setting the provider statically as a kind of singleton.

extensions/hibernate-orm/runtime/src/main/java/io/quarkus/hibernate/orm/runtime/tenant/HibernateMultiTenantConnectionProvider.java
michael-schnell: OK, I changed it

extensions/hibernate-orm/deployment/src/main/java/io/quarkus/hibernate/orm/deployment/HibernateOrmConfig.java
michael-schnell: OK, I removed it

extensions/hibernate-orm/deployment/src/main/java/io/quarkus/hibernate/orm/deployment/HibernateOrmConfig.java
michael-schnell: OK, I added a hint for the default.

extensions/hibernate-orm/deployment/src/main/java/io/quarkus/hibernate/orm/deployment/HibernateOrmProcessor.java
michael-schnell: OK, I changed it.

extensions/hibernate-orm/deployment/src/main/java/io/quarkus/hibernate/orm/deployment/HibernateOrmConfig.java
michael-schnell: This could be helpful if you for example want to separate the default database from the one used for tenant schemas. I would prefer to leave it as it is, to allow separation in case this is needed. We also have  a good default if the user does not set it. But there is now still the option to separate the tenant stuff from the default data source.

extensions/hibernate-orm/runtime/src/main/java/io/quarkus/hibernate/orm/runtime/tenant/HibernateCurrentTenantIdentifierResolver.java
```diff
@@ -0,0 +1,66 @@
...
+                    + "Please use a non-null value!");
+        }
+        LOG.debugv("resolveCurrentTenantIdentifier(): {0}", tenantId);
+        return tenantId;
+
+    }
+
+    @Override
+    public boolean validateExistingCurrentSessions() {
+        return jpaConfig.isValidateTenantInCurrentSessions();
```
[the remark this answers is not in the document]
michael-schnell: Already removed when removing config 'ValidateTenantInCurrentSessions'.

michael-schnell:
@Sanne All changes are now commited, including the new 'hibernate-tenancy' integration test. 
I used the 'jpa-mariadb' as template for the test. It states "By default, the tests of this module are disabled." - I hope this is also OK for the new 'hibernate-tenancy' integration test.

Sanne:
Hi @michael-schnell , regarding a question you had regarding the "The @PathParam is somehow not passed correctly." you mentioned in a comment (which I can't find on the github UI now) - did you solve it?

In case, it might relate to the compiler settings; you need to have the `parameters=true` configuration property on the compiler pluugin, like we have it in most quickstarts:

    <plugin>
                <groupId>org.apache.maven.plugins</groupId>
                <artifactId>maven-compiler-plugin</artifactId>
                <version>${compiler-plugin.version}</version>
                <configuration>
                    <source>1.8</source>
                    <target>1.8</target>
                    
                    <parameters>true</parameters>
                </configuration>
            </plugin>

Awesome progress on the other commits, I'll check it out in more detail tonight.

michael-schnell:
> Hi @michael-schnell , regarding a question you had regarding the "The @PathParam is somehow not passed correctly." you mentioned in a comment (which I can't find on the github UI now) - did you solve it?

Tnx, I solved it 5 minutes after I posted the comment (That's why I deleted it). It was a strange behaviour that was different between the integration test project and the quickstart example. I forgot to add the name of the parameter in the "@PathParam" annotation. So it was something like `@PathParam String myVar` instead of `@PathParam("myVar") String myVar`. It worked for the quickstart project (that's were I copied it from), but failed in the integration test. I was searching for hours for what was wrong... Fortunately finally I got it.

Sanne:
Haven't fully finished the review, but since it looks great so far:

could you amend the expected MariaDB configuration? The used database password is not consistent with other modules, it would be great if you could use exactly the same configuration so that the CI infrastructure can provide a single DB instance for all testing needs.

See https://github.com/quarkusio/quarkus/blob/df66ad45da4cac4aa1d4e761550117eae42f8757/.github/workflows/ci-actions.yml#L141-L149

michael-schnell:
@geoand  Can you restart the build job? Seems as if something timed out that is not related to the changes.

geoand:
Done

michael-schnell:
> could you amend the expected MariaDB configuration? The used database password is not consistent with other modules, it would be great if you could use exactly the same configuration so that the CI infrastructure can provide a single DB instance for all testing needs.
 
Which one do you mean exactly? The test needs to create two new databases for the tenants in the MariaDB instance. So I need a user that is allowed to do that. Only the root user normally has the necessary rights. That's why the default datasource has a different user than the two other tenant data sources.

Must be root to be able to create new DBs:
```
# Default DB Configuration
quarkus.datasource.db-kind=mariadb
quarkus.datasource.username=root
quarkus.datasource.password=secret
```
The other two are created by `src/main/resources/database/default/V1.0.0__init_databases.sql`:
```
# DATABASE Tenant 'base' Configuration
quarkus.datasource.base.db-kind=mariadb
quarkus.datasource.base.username=jane
quarkus.datasource.base.password=abc

# DATABASE Tenant 'mycompany' Configuration
quarkus.datasource.mycompany.db-kind=mariadb
quarkus.datasource.mycompany.username=john
quarkus.datasource.mycompany.password=def
```
Unfortunately only one database (the "hibernate_orm_test") will get created automatically. The other ones are therefore created by the script including the users and rights.

michael-schnell:
> Done

Tnx. Unfortunately now the another one fails that was fine before... :-(

michael-schnell:
@geoand Another try? Seems to be unstable...

integration-tests/hibernate-tenancy/src/main/resources/application.properties
```diff
@@ -0,0 +1,38 @@
+# Hibernate ORM settings 
+quarkus.hibernate-orm.database.generation=none
+quarkus.hibernate-orm.multitenant=DATABASE
+quarkus.hibernate-orm.validate-tenant-in-current-sessions=false
```
Sanne, line 4: this property no longer exists

Sanne:
> Which one do you mean exactly? The test needs to create two new databases for the tenants in the MariaDB instance. So I need a user that is allowed to do that. Only the root user normally has the necessary rights. That's why the default datasource has a different user than the two other tenant data sources.

ah right, thanks! I hadn't seen the integration test yet - now I understand and agree.

So we'll need to adapt CI for this - but it's ok we can follow up on that as a separate issue if you prefer. I've opened #9389 - we recently changed CI system so I'm not overly familiar with it yet so I might try volunteering for that :)

Sanne (approved):

Sanne:
Merged! Many thanks @michael-schnell !

integration-tests/hibernate-tenancy/src/main/resources/application.properties
michael-schnell: Fixed with https://github.com/quarkusio/quarkus/pull/9423

